PR #594, opened 2018-11-10
new demo script

ericxtang:
Transcoding demo script update to accommodate for cli/media port change.

### DIFF
--- a/transcode_demo.sh
+++ b/transcode_demo.sh
@@ -1,10 +1,4 @@
-strmID=$(curl localhost:8935/streamID 2> /dev/null)
-echo $strmID
-curl "http://localhost:8935/transcode?strmID=$strmID"
-sleep 2
-
-manifestID=$(curl localhost:8935/manifestID 2> /dev/null)
-manifest=$(curl localhost:8935/stream/$manifestID.m3u8 2> /dev/null)
+manifest=$(curl -s 'localhost:7935/status' | jq -r '.[][]' 2> /dev/null)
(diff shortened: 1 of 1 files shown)


transcode_demo.sh
```diff
@@ -1,13 +1,7 @@
...
-manifest=$(curl localhost:8935/stream/$manifestID.m3u8 2> /dev/null)
+manifest=$(curl -s 'localhost:7935/status' | jq -r '.[][]' 2> /dev/null)
 sid1=$(echo $manifest| cut -d' ' -f 4)
 sid2=$(echo $manifest| cut -d' ' -f 6)
 sid3=$(echo $manifest| cut -d' ' -f 8)
 ffplay "http://localhost:8935/stream/$sid1" &
 ffplay "http://localhost:8935/stream/$sid2" &
-ffplay "http://localhost:8935/stream/$sid3" &
\ No newline at end of file
+ffplay "http://localhost:8935/stream/$sid3" &
```
j0sh, line 7: Actually, now they can just `ffplay "http://localhost:8935/stream/current.m3u8` without having to query for the manifest.

It seems like there are several streamids here in this script -- I don't think that's supported anymore, at least until we put multiple-stream support back into the node.

ericxtang: Here the demo tries to play the 3 different renditions after transcoding.

j0sh: Ah got it!

j0sh (approved):
LGTM 🚢

PR #601, opened 2018-11-21
Self check related changes.

### LINKED ISSUE #600: Orchestrator doesn't start
**Describe the bug**
Orchestrator, at start, fails to pass `CheckTranscoderAvailability` check and exits.

**To Reproduce**
Try to run orchestrator.

**Additional context**
here
https://github.com/livepeer/go-livepeer/blob/3a435f5e8c3a8a0b2dcfa366587d2127a04d7907/server/rpc.go#L120-L122
connection gets closed right after being open

also this function doesn't work as expected - it fails at `verifyMsgSig`
https://github.com/livepeer/go-livepeer/blob/3a435f5e8c3a8a0b2dcfa366587d2127a04d7907/server/rpc.go#L76-L100

j0sh:
* Fix verification param transposition. This wasn't caught by unit testing because of a typo in the test name.
* Use manually specified ServiceURI if not set. Not strictly necessary at the moment, but useful for quick testing. Will become necessary for offchain mode. Perhaps we should also rename this to "orchAddr" following the O/T split.
* rpc: Return the underlying orchestrator connection so we don't close the cxn immediately after creating it.

Fixes https://github.com/livepeer/go-livepeer/issues/600

### DIFF
--- a/server/rpc.go
+++ b/server/rpc.go
@@ -82,10 +82,11 @@ func CheckTranscoderAvailability(orch Orchestrator) bool {
 
 	ping := crypto.Keccak256(ts_signature)
 
-	orch_client, err := startOrchestratorClient(orch.ServiceURI().String())
+	orch_client, conn, err := startOrchestratorClient(orch.ServiceURI().String())
 	if err != nil {
 		return false
 	}
+	defer conn.Close()
 
 	ctx, cancel := context.WithTimeout(context.Background(), GRPCTimeout)
 	defer cancel()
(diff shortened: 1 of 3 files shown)


darkdarkdragon (approved):
LGTM

server/rpc_test.go
```diff
@@ -292,7 +292,7 @@ func TestRPCSeg(t *testing.T) {
 	}
 }
 
-func testPing(t *testing.T) {
+func TestPing(t *testing.T) {
```
angyangie, line 295: Didn't realize this had to be capitalized. Noted!

server/rpc.go
```diff
@@ -82,10 +82,11 @@ func CheckTranscoderAvailability(orch Orchestrator) bool {
 
 	ping := crypto.Keccak256(ts_signature)
 
-	orch_client, err := startOrchestratorClient(orch.ServiceURI().String())
+	orch_client, conn, err := startOrchestratorClient(orch.ServiceURI().String())
 	if err != nil {
 		return false
 	}
+	defer conn.Close()
```
angyangie, line 89: I see. We want the orchestrator connection to close after this function exits .... i.e. once we've checked that the orchestrator ("transcoder") is available. Noted!

server/rpc.go
```diff
@@ -96,15 +97,15 @@ func CheckTranscoderAvailability(orch Orchestrator) bool {
 		return false
 	}
 
-	return verifyMsgSig(orch.Address(), string(pong.Value), ping)
+	return verifyMsgSig(orch.Address(), string(ping), pong.Value)
```
angyangie, line 100: `Sig` is the return value. `Msg` is the input ping value. Noted!

angyangie (approved):
Thanks for handling this @j0sh !

PR #604, opened 2018-11-26
Orchestrator Discovery Changes

angyangie:
# O Discovery Changes

## CLI Flags

Broadcaster needs to be started with an `-orchAddr address:port` option in order to specify the orchestrator to use. This is an interim solution while the full negotiation piece is being developed.

If using a standalone transcoder, needs to be started with `-standaloneTranscoder`, along with the existing `-orchAddr` and `-orchSecret` options. This was needed to disambiguate broadcaster mode from standalone transcoder mode, since we opted to not require any more flags for the broadcaster with its use of `-orchAddr`. Discuss this here: https://github.com/livepeer/go-livepeer/issues/605

## Segment Signing

Segment signature needs to be updated to incorporate the desired profiles from the orchestrator.

**Old signature:**

```
sig = broadcaster.Sign(streamId | seqNo | segment.dataHash)
```

This data was taken from the `ethTypes.Segment` struct which reflects how the information is processed on-chain during verification. Since we're not updating the on-chain processing right now, there is a new struct `core.SegmentMetadata` from which the signature is constructed. The old `eth/types.Segment` struct is left unmodified.

**New signature:**

```
type SegmentMetadata struct {
	ManifestID ManifestID
	Seq        int64
	Hash       ethcommon.Hash
	Profiles   []ffmpeg.VideoProfile
	OS         *net.OSInfo
}

sig = broadcaster.sign(manifestId | seqNo | profiles | segment.dataHash)
```

The ManifestID is used here rather than the StreamID since it is a little bit shorter, and the additional (often incorrect) information carried by the appended rendition string is not needed.

The composition of the signature and the representation of the profile is subject to change once we have flexible transcoding options via non-deterministic verification and price menus, but for now we continue to use the existing `[]ffmpeg.VideoProfile` bytes-representation created by concatenating strings from a lookup table.

## Networking

* Renamed `GetTranscoder : TranscoderRequest -> TranscoderInfo` to `GetOrchestrator : OrchestratorRequest -> OrchestratorInfo`

The `GetOrchestrator` request is the discovery mechanism.

```
// This request is sent by the broadcaster in `GetOrchestrator` to request
// information on which transcoder to use.
message OrchestratorRequest {

  // Ethereum address of the broadcaster
  bytes address = 1;

  // Broadcaster's signature over its address
  bytes sig   = 2;
}

// The orchestrator sends this in response to `GetTranscoder`, containing the
// transcoder URI, associated credentials authorizing the broadcaster to
// use the transcoder, and miscellaneous data related to the job.
message OrchestratorInfo {

  // URI of the transcoder to use for submitting segments.
  string transcoder  = 1;

  // Signals the authentication method to expect within `credentials`. This
  // field is opaque to the broadcaster, and should be passed to the transcoder.
  string authType    = 2;

  // Credentials to verify the request has been authorized by an orchestrator.
  // This field is opaque to the broadcaster.
  string credentials = 3;

  // Orchestrator returns info about own input object storage, if it wants it to be used.
  repeated OSInfo storage = 32;
}
```

Note it is likely `GetOrchestrator` will need to be invoked twice: once during the periodic O discovery, and again before a broadcast starts for more up-to-information such as storage credentials and PM parameters.

In that respect, the `OrchestratorInfo.Storage` field may seem redundant for O Discovery alone, but otherwise the differences are not enough to warrant a separate RPC type for the two `GetOrchestrator` invocations.

## Broadcaster interface session split

Separate out transient per-session data from the more permanent interface exposing the broadcaster's address and its signing capability. Previously, they were mixed, which led to unclear semantics around the data required for O Discovery.

O Discovery and segment submission only requires the address and its signature as give by `GetOrchestrator`.

```
type Broadcaster interface {
	Address() ethcommon.Address
	Sign([]byte) ([]byte, error)
}

// Session-specific state for broadcasters
type BroadcastSession struct {
	Broadcaster      Broadcaster
	ManifestID       core.ManifestID
	Profiles         []ffmpeg.VideoProfile
	OrchestratorInfo *net.OrchestratorInfo
	OrchestratorOS   drivers.OSSession
	BroadcasterOS    drivers.OSSession
}

```

## Orchestrator S3 Path Layout

The locations of the source and transcoded results may differ in their path prefixes. See [this writeup](https://hackmd.io/0C2Zi0VVSyGEgN_4lEu_WA) for more details on the issue. We generate different prefixes each time the session is initialized (corresponding to the first solution listed).


cmd/livepeer/livepeer.go
```diff
@@ -154,6 +155,10 @@ func main() {
 		n.NodeType = core.Broadcaster
 	}
 
+	// if *orchAddr != "" {
+	// 	offchainOrch := n.OrchestratorSelector.NewOffchainOrchestrator(*orchAddr)
```
angyangie: With the way things are now, we'd have to use offchainOrch as an input to `StartMediaServer`, then `gotRTMPStreamHandler`, and `startBroadcast` ... unless we added an address field to `NewLivepeerNode` in `livepeernode.go`. What do you think @j0sh ?

core/streamdata.go
```diff
@@ -21,9 +23,33 @@ const (
 	HashLength = 32
 )
 
+type SegmentMetadata struct {
```
ericxtang: Can we leave a comment here explaining why we are tracking ManifestID instead of StreamID?  (Already explained in the PR notes, but want to make sure it stays with the code)

README.md
```diff
@@ -101,7 +101,21 @@ We'll walk through the steps of becoming a transcoder on the test network.  To l
...
-- If running on Rinkeby or mainnet, ensure your transcoder is *publicly accessible* in order to receive jobs from broadcasters. The only port that is required to be public is the one that was set during the transcoder registration step (default 8935).
+- If running on Rinkeby or mainnet, ensure your orchestrator is *publicly accessible* in order to receive jobs from broadcasters. The only port that is required to be public is the one that was set during the transcoder registration step (default 8935).
+
+### Standalone Orchestrators
+
+Orchestrators can be run in standalone mode without an attached transcoder. Standalone transcoders will need to connect to this orchestrator in order for the orchestrator to process jobs.
+
+- `livepeer --rinkeby --orchestrator -orchSecret asdf`
+
+The orchSecret is a shared secret used to authenticate remote transcoders.
```
ericxtang: Looks like secret can be any arbitrary string?  If so, let's put it in the comment.

cmd/livepeer/livepeer.go
```diff
@@ -187,6 +189,17 @@ func main() {
 		n.NodeType = core.BroadcasterNode
 	}
 
+	if n.NodeType == core.BroadcasterNode {
+		if *orchAddr == "" {
+			glog.Info("No orchestrator specified; transcoding will not happen")
```
ericxtang, line 194: Can we exit here for now?  In the next iteration when we implement O discovery via onchain info, we can take it out.

common/util.go
```diff
@@ -90,6 +91,31 @@ func TxDataToVideoProfile(txData string) ([]ffmpeg.VideoProfile, error) {
 	return profiles, nil
 }
 
+func BytesToVideoProfile(txData []byte) ([]ffmpeg.VideoProfile, error) {
```
ericxtang, line 94: `BytesToVideoProfile` and `VideoProfileToBytes` have different expectation of what "Bytes" should be.  For symmetry sake, can we make them the same?

common/util_test.go
```diff
@@ -21,3 +23,42 @@ func TestTxDataToVideoProfile(t *testing.T) {
...
+func TestProfilesToHex(t *testing.T) {
+	// Sanity checking against an existing eth impl that we know works
+	compare := func(profiles []ffmpeg.VideoProfile) {
+		b1 := ethcommon.ToHex(ProfilesToTranscodeOpts(profiles))[2:]
+		b2 := ProfilesToHex(profiles)
+		if b1 != b2 {
+			t.Error("Unequal profile hex ")
+		}
+	}
+	// XXX double check which one is wrong! ethcommon method produces "0" zero string
```
ericxtang, line 59: Did we resolve this?

common/videoprofile_ids.go
```diff
@@ -13,3 +20,19 @@ var VideoProfileNameLookup = map[string]string{
...
+			continue
+		}
+		var key [VideoProfileIDBytes]byte
+		copy(key[:], b)
+		ret[key] = v
+	}
+	return ret
+}
+
+var VideoProfileByteLookup = videoProfileToBytes()
```
ericxtang: What's the point of this var?

core/livepeernode_test.go
```diff
@@ -3,45 +3,28 @@ package core
 import (
 	"fmt"
 	"io/ioutil"
-	"math/big"
 	"os"
 	"testing"
-	"time"
+	//"time"
```
ericxtang: remove dead code

core/livepeernode_test.go
```diff
@@ -3,45 +3,28 @@ package core
...
-	lpTypes "github.com/livepeer/go-livepeer/eth/types"
-	ffmpeg "github.com/livepeer/lpms/ffmpeg"
-)
-
-type StubClaimManager struct {
-	verifyCalled         bool
-	distributeFeesCalled bool
-	receiptAdded         bool
-}
+	//ethcommon "github.com/ethereum/go-ethereum/common"
```
ericxtang: remove dead code

core/livepeernode_test.go
```diff
@@ -84,90 +59,38 @@ func TestTranscodeAndBroadcast(t *testing.T) {
...
-	if len(tr.InputData) != 1 {
-		t.Errorf("Expecting 1 segment to be transcoded, got %v", tr.InputData)
+	//Should have 1 transcoded segment into 2 different profiles
+	if tr.SegCount != 1 {
+		t.Errorf("Expecting 1 segment to be transcoded, got %v", tr.SegCount)
 	}
 
 	// Check playlist was updated
-	for _, v := range strmIds {
+	// XXX Fix this
```
ericxtang, line 65: Are we fixing this or removing it?

core/orchestrator.go
```diff
@@ -97,23 +60,8 @@ func (orch *orchestrator) TranscoderSecret() string {
...
-			return []StreamID{}, err
-		}
-		streamIds[i] = strmId
-	}
-	return streamIds, nil
-}
-
-func (orch *orchestrator) TranscodeSeg(job *ethTypes.Job, ss *SignedSegment) (*TranscodeResult, error) {
-	return orch.node.TranscodeSegment(job, ss)
+func (orch *orchestrator) TranscodeSeg(md *SegmentMetadata, seg *stream.HLSSegment) (*TranscodeResult, error) {
```
ericxtang: A little nitpicky, but I find it confusing to have `TranscodeSeg`, `transcodeSeg`, and `TranscodeSegment` that seem to have no real functional relationships (Normally, a `TranscodeSeg` would call `transcodeSeg` because of public/private function).  Can we come up with more appropriate function names?

core/livepeernode_test.go
j0sh: Should be fixed, but the test needs to be redone significantly to achieve the same level of coverage. The note is just so we don't forget.

common/util.go
j0sh: Can you elaborate on this? There isn't a `VideoProfileToBytes` anywhere. However, there is a `TxDataToVideoProfile` which takes a hex-encoded string.

We can express `TxDataToVideoProfile` in terms of `BytesToVideoProfile` by decoding the hex input into bytes, but I think we'll be removing `TxDataToVideoProfile` entirely once we do a second pass of cleanups -- there's code in the ethclient and DB that still depends on it.

common/videoprofile_ids.go
j0sh: Creates a lookup table translating each profile to its four-byte equivalents, rather than its hex-string equivalent as given by `VideoProfileLookup`. Used in `BytesToVideoProfile`.

This way we can look up video profiles based on its bytes without doing the int-to-hexstring conversion.

server/mediaserver.go
```diff
@@ -349,9 +292,9 @@ func gotRTMPStreamHandler(s *LivepeerServer) func(url *url.URL, rtmpStrm stream.
 					monitor.LogSegmentEmerged(nonce, seg.SeqNo)
 				}
 
-				if jobId != nil {
+				/*if jobId != "" { // ANGIE - DATABASE.SETSEGMENTCOUNT NEEDS TO BE UPDATED TO TAKE A STRING
```
ericxtang: Remove?

server/mediaserver.go
```diff
@@ -466,10 +409,10 @@ func gotRTMPStreamHandler(s *LivepeerServer) func(url *url.URL, rtmpStrm stream.
 						}
 						cond.L.Unlock()
 
-						if !eth.VerifySig(rpcBcast.Job().TranscoderAddress, crypto.Keccak256(segHashes...), res.Sig) {
-							glog.Error("Sig check failed for segment ", seg.SeqNo)
-							return
-						}
+						// if !eth.VerifySig(transcoderAddress, crypto.Keccak256(segHashes...), res.Sig) { // ANGIE - GET TRANSCODER ADDRESS HERE
```
ericxtang: Remove?

common/util.go
ericxtang: `VideoProfileToBytes` is in `common/videoprofile_ids.go`.  Just realized they have different semantics (this is just an unfortunate coincidence of private function naming)

core/orchestrator.go
j0sh: I do agree that the `TranscodeSegment` here is perhaps misattributed as public when it's really a private function.

* `TranscodeSeg` The "friendly" name of the external interface that's exposed to RPC, testing, etc.
* `TranscodeSegment` Agreed that this is misattributed as public when it's really a private function. Perhaps `sendToTranscodeLoop"` ?
* `transcodeSeg` This is **the** function that calls the transcoder. The naming seems OK here, but open to other ideas.

common/videoprofile_ids.go
ericxtang: Oh I see.  That explains my confusion with `VideoProfileToBytes` earlier.

server/mediaserver.go
j0sh: We really should be verifying the segment sig once we have the transcoder address, so I wouldn't remove this part entirely.

core/livepeernode_test.go
ericxtang: Ok.  Are we waiting for the next iteration to write this test?  Or should we do it now?

common/util.go
j0sh: Oh, `videoProfileToBytes`  -- caps.

I can rename the `videoProfileToBytes` function to `makeVideoProfileByteMap` if that's clearer.

core/streamdata.go
j0sh: How should we explain it? The PR notes work for reviewers who understand the previous state of the codebase and need to relate to how it's changed, but it seems a little strange to refer to "previous state" for comments in an updated codebase.

Presumably the PR note is this one:

> The ManifestID is used here rather than the StreamID since it is a little bit shorter, and the additional (often incorrect) information carried by the appended rendition string is not needed.

core/livepeernode_test.go
j0sh: Next iteration probably, since the playlist (OS actually) is updated in the RPC layer (`ServeSegment`) and that particular area is hard to test right now. There needs to be some refactoring for testability.

Most of these tests have been disabled since OS. We're gaining them back a little at a time though.

common/util_test.go
j0sh: Nope, it's still a bit of an open question. To clarify,

```
ethcommon.ToHex([]byte{}) == "0x0"
---
hex.EncodeToString([]byte{}) == ""
---
"0x0" != ""
```

I'm inclined to say that the `ethcommon` behavior is incorrect because an empty string does not equal zero, but what do I know.

j0sh: And the reason for using `hex.EncodeToString` over `ethcommon.ToHex`within `ProfilesToHex` is because the latter appends a leading `0x`, which we end up removing nearly every time anyway, even among our other uses of `ethcommon.Hex`.

So here we avoid the extra manipulation and use a function that gives us the results we want.

cmd/livepeer/livepeer.go
j0sh: We can do that, but it'd mean stopping support for view-only source streams.

Although I suppose we might also want to figure out how to support view-only mode once we use onchain info.

core/streamdata.go
ericxtang: I think we want to explain the semantics of what it means to have a ManifestID and multiple VideoProfile in SegmentMetadata.  Theoretically, a video segment can ONLY have 1 VideoProfile.  It looks like we are only use `SegmentMetadata` to signal transcoding preference.  I don't want someone to misuse this for another purpose (for example, representing metadata for a specific segment)

cmd/livepeer/livepeer.go
ericxtang: Ahh I see.  Ok I'm on board for keeping things this way then.  view-only feels useful.

common/util_test.go
ericxtang: Ahh I see.  I guess technically `ethcommon.Hex` is in context of Ethereum, and in that world `0x` might be fine, but I agree it's confusing.  I'm fine with either as long as we are wrapping that so programmers don't have to think about it.

core/orchestrator.go
ericxtang: I'm good with these names.

core/streamdata.go
j0sh: The `SegmentMetadata` does represent metadata for a specific segment, since it also holds the segment `Hash` and `Seq`.

In theory the `Profiles` and `OS` could also change per-segment as well.

The (only?) reason `ManifestID`  is here is to act as an index to look up the transcoding loop. We could mitigate the need for a ManifestID in the metadata if we had kept the AuthToken.

j0sh: (The ManifestID is also part of the segment signature, but I don't think it's strictly necessary for Streamflow.)

common/util_test.go
angyangie: `ethcommon.ToHex` appends a leading `0x`, which when removed, which equals the output of `hex.EncodeToString`? What do you mean by wrapping? Do programmers have to "think about it" the way it's written here? Just trying to understand what's going on here!

j0sh: > ethcommon.ToHex appends a leading 0x, which when removed, which equals the output of hex.EncodeToString?

Correct, in most cases. What's under discussion here is the cases where the outputs are *not* equal between the two hex-encoding functions.

> What do you mean by wrapping? Do programmers have to "think about it" the way it's written here?

Many of the uses of `ToHex` throughout the codebase involve stripping the leading "0x" prefix with `ethcommon.ToHex(string)[2:]`

Wrapping means the programmer doesn't have "think about" doing that stripping, regardless of the hex-encoding method that's being used. It is a form of encapsulation. Here, we use `hex.EncodeToString` in `ProfilesToHex` just so we don't have to strip at all, but even if we used `ethcommon.ToHex` instead, its use would be wrapped by the `ProfilesToHex` function, so the programmer doesn't have to be aware of the differences in hex-encoding methods.

angyangie: That makes sense. I do see where it's being used this way `ethcommon.ToHex(string)[2:]`. I understand now, thank you Josh!

core/streamdata.go
ericxtang: @j0sh and I talked about renaming `SegmentMetadata` to `SegTranscodingMetadata`.  Can we do that?

j0sh: Done

core/orchestrator.go
j0sh: Done

ericxtang (approved):
LGTM.  Let's 🚢

server/rpc.go
```diff
@@ -455,7 +454,7 @@ func SubmitSegment(bcast Broadcaster, seg *stream.HLSSegment, nonce uint64) (*ne
 		data = []byte(seg.Name)
 	}
 
-	ti := bcast.GetTranscoderInfo()
+	ti := sess.OrchestratorInfo
```
eladmallel, line 463: rename `ti` to `oi`?

more generally, can you share the ideas behind using very short variable names? (for example in this case I'm inclined to name this variable at least `oInfo` if not `orchInfo` or even longer :) )

discovery/discovery.go
```diff
@@ -0,0 +1,32 @@
...
+	uri   *url.URL
+	bcast server.Broadcaster
+}
+
+func NewOffchainOrchestrator(node *core.LivepeerNode, address string) *offchainOrchestrator {
+	addr := "https://" + address
+	uri, err := url.ParseRequestURI(addr)
+	if err != nil {
+		glog.Error("Could not parse orchestrator URI: ", err)
+		return nil
```
eladmallel, line 23: why not follow the pattern of returning two return values (o, err) like we do in other places, e.g. `NewLivepeerNode`?

eladmallel:
Making sure I fully understand this - why do we need B to share their address and signature in `OrchestratorRequest`?

server/rpc.go
j0sh: > more generally, can you share the ideas behind using very short variable names?

Just a particular style for short-lived local variables. The meaning can be clearly seen from the variable declaration, although you are correct here it would've been better with `oi`.

This is mostly my particular style of programming, but it seems to be shared by the golang community. See https://talks.golang.org/2014/names.slide

discovery/discovery.go
j0sh: We can do that in the next round of updates.

angyangie:
@eladmallel we use both in [verifyOrchestratorReq](https://github.com/livepeer/go-livepeer/pull/604/files#diff-44bb03d888e11d6429223a7b0e55220dR149) to verify the Orchestrator request! But I think you might be alluding to the fact that we can [make the Sig from the address](https://github.com/livepeer/go-livepeer/pull/604/files#diff-44bb03d888e11d6429223a7b0e55220dR81), so why store both?

j0sh:
> why do we need B to share their address and signature in OrchestratorRequest?

The idea is to give O an opportunity to distinguish who's sending the request and authenticate the sender; that can't (currently) be inferred from the TLS connection alone.

angyangie:
@j0sh now that Elad brought it up .. why can't we just store the address and just generate the Sig when we need it?

j0sh:
> why can't we just store the address and just generate the Sig when we need it?

Perhaps I'm missing something, but that's pretty much what we do now.

angyangie:
@j0sh I mean, instead of storing both in the protobuf, we could just do `sig, err := b.Sign([]byte(fmt.Sprintf("%v", b.Address().Hex())))` in `verifyOrchestratorReq` and only have to store address in the protobuf. Not a big deal at all, but was just wondering if there was a reason we were storing the sig and the address in the protobuf! I haven't worked with protobufs for a long time, but it seems to me thus far that we use the protobuf to store variables that will be used all over the codebase (and save ourselves the trouble of having to pass these variables through tons of functions before they're actually used anywhere, etc.)

j0sh:
> we could just do sig, err := b.Sign([]byte(fmt.Sprintf("%v", b.Address().Hex()))) in verifyOrchestratorReq

The orchestrator can't generate a signature on behalf of B; it can only check a sig given an address, so we need to supply both here.

> we use the protobuf to store variables that will be used all over the codebase

The only one I can think of is `OrchestratorInfo`. Are there others?

It's debatable whether we should continue using the `net.OrchestratorInfo` struct to propagate orchestrator state throughout the broadcast flow, because what we need to transmit on-the-wire (via protobufs) may not necessarily correspond directly to the state that we need to track internally on the node. But for now, there isn't much of a discrepancy, so it's OK until things change.

>  having to pass these variables through tons of functions before they're actually used anywhere, etc

Can you elaborate on where this is an issue?

eladmallel:
My question is specifically around what does O gain from being able to authenticate the sender of the initial `OrchestratorRequest`? 

Authentication is obviously key when receiving a request to do work. However, in this simple request for information, I'm not sure I understand the importance of authentication.

Does this clarify the question better? @j0sh @angyangie

j0sh:
@eladmallel  O can customize the response based on the sender. Preferential pricing or parameterization, etc.

server/rpc.go
eladmallel: thanks for sharing your general school of thought! I like the content in the link you shared :)

PR #608, opened 2018-12-04
ServiceURI and language changes

j0sh:
Refactoring of ServiceURI and renaming CheckTranscoderAvailability

- Extract ServiceURI processing code into its own function
- Rename CheckTranscoderAvailability to CheckOrchestratorAvailability
- Misc language changes around transcoders / orchestrators 

* Start a node with `-serviceAddr` . Ensure that given address is used
* Start a node without `-serviceAddr` . Ensure that the address stored on blockchain is used.

Note that the availability self check will fail if the ServiceAddr is invalid.

N/A

**Checklist:**

### DIFF
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -370,11 +370,12 @@ func main() {
 			tc <- struct{}{}
 		}()
 
-		// check whether or not the transcoder is available
+		// check whether or not the orchestrator is available
 		time.Sleep(2 * time.Second)
-		transcoderAvailable := server.CheckTranscoderAvailability(orch)
-		if !transcoderAvailable {
-			// shut down transcoder
+		orchAvail := server.CheckOrchestratorAvailability(orch)
+		if !orchAvail {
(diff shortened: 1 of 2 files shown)


j0sh:
Should we re-purpose `-orchAddr` option to override the on-chain ServiceURI, rather than the old `-serviceAddr` ?

The `-orchAddr` option is already being used for broadcasters (as of #604) and standalone transcoders. Reusing it for the orchestrator would give us a bit of consistency in flag naming, and allow for one less option overall. However the semantics of the flag would be different depending on the mode the node is in, which may be confusing for some users.

ericxtang:
I think `-orchAddr` and `-serviceAddr` are 2 different flags that exists on 2 different types of nodes (one is for a broadcaster, another is for an orchestrator).  If we move toward having different node types in the future, then I think we should have different sets of flags for these nodes.

ericxtang (approved):
The changes here look good 🚢

j0sh:
Fixed a typo in one of the log lines.

PR #610, opened 2018-12-05
Fix offchain mode.

j0sh:
**What does this pull request do? Explain your changes. (required)**
Makes offchain mode work for broadcasters and orchestrators.

There are two dependencies for this PR: https://github.com/livepeer/go-livepeer/pull/608 and https://github.com/livepeer/go-livepeer/pull/604. Once #608 is merged, we can pull this PR into #604.

**Specific updates (required)**
- Return default values instead of errors when trying to sign or verify sigs
- Add a VerifySig to the RPC Orchestrator interface. Needed to determine whether we're offchain.

**How did you test each of these updates (required)**
Run B and O with an `-offchain` flag on top of the O Discovery PR. https://github.com/livepeer/go-livepeer/pull/604

* B needs `-orchAddr` specified to select the O to connect to for transcoding.
* O needs `-serviceAddr` specified to substitute for the on-chain ServiceURI. See https://github.com/livepeer/go-livepeer/pull/608#issuecomment-444363685 about renaming this option.

**Does this pull request close any open issues?**

https://github.com/livepeer/go-livepeer/issues/609

**Checklist:**

- [ ] README and other documentation updated
- [x] Node runs in OSX and devenv

PR #612, opened 2018-12-06
It/latency metrics

darkdarkdragon:
Improve metrics code, adds metrics events needed to track end-2-end latency

Ran on nodes on EC2

**Checklist:**


core/transcoder.go
```diff
@@ -15,9 +20,31 @@ type LocalTranscoder struct {
 
 func (lt *LocalTranscoder) Transcode(fname string, profiles []ffmpeg.VideoProfile) ([][]byte, error) {
 	tr := transcoder.NewFFMpegSegmentTranscoder(profiles, lt.workDir)
-	return tr.Transcode(fname)
+	mid, profile, seqNo, err := parseURI(fname)
```
j0sh: The URI format will change with O Discovery. It'll just be:

```
<mid>/<seq>.ts
```

Note that the profile doesn't matter here, since it's always the source stream.

core/transcoder.go
```diff
@@ -15,9 +20,31 @@ type LocalTranscoder struct {
...
+	if monitor.Enabled && err == nil {
+		monitor.LogSegmentTranscodeStarting(seqNo, mid, profile)
+	}
+	data, err := tr.Transcode(fname)
+	if monitor.Enabled && err == nil {
+		monitor.LogSegmentTranscodeEnded(seqNo, mid, profile)
```
j0sh: What happens if we have an error? Seems like the analytics would "dangle" where there's a `TranscodeStarting` event but no corresponding ending event.

core/orchestrator.go
```diff
@@ -291,8 +294,8 @@ func (n *LivepeerNode) transcodeAndCacheSeg(config transcodeConfig, ss *SignedSe
 	}
 
 	transcodeStart := time.Now().UTC()
-	// Ensure length matches expectations. 4 second + 25% wiggle factor, 60fps
-	if err := ffmpeg.CheckMediaLen(fname, 4*1.25*1000, 60*4*1.25); err != nil {
+	// Ensure length matches expectations. SegLen second + 25% wiggle factor, 60fps
+	if err := ffmpeg.CheckMediaLen(fname, SegLen*1.25*1000, 60*SegLen*1.25); err != nil {
```
j0sh: This is going away with O Discovery

core/orchestrator.go
```diff
@@ -325,6 +329,11 @@ func (n *LivepeerNode) transcodeAndCacheSeg(config transcodeConfig, ss *SignedSe
 		}
 		seg.Name = url
 	}
+	if isLocal && monitor.Enabled {
+		sid := StreamID(config.StrmID)
+		mid, _ := sid.ManifestIDFromStreamID()
+		monitor.LogSegmentTranscodeStarting(seg.SeqNo, mid.String(), sid.GetRendition())
```
j0sh: How we get the manifest ID and renditions will change with O Discovery.

eth/eventservices/jobservice.go
```diff
@@ -138,14 +138,14 @@ func (s *JobService) firstClaim(ctx context.Context, job *lpTypes.Job) error {
 				if lastSeenBlock.Cmp(firstClaimBlock) >= 0 {
 					canClaim, err := cm.CanClaim(lastSeenBlock, job)
 					if err != nil {
-						glog.Errorf("Error checking if the transcoder can claim ", err)
+						glog.Errorf("Error checking if the transcoder can claim %v", err)
```
j0sh: We won't be hitting this path after O Discovery

cmd/livepeer/livepeer.go
```diff
@@ -180,11 +180,32 @@ func main() {
...
 	} else if transcoder {
 		n.NodeType = core.TranscoderNode
+	} else {
+		n.NodeType = core.BroadcasterNode
+	}
+
+	if *monitor {
+		glog.Infof("Monitoring endpoint: %s", *monhost)
+		lpmon.Enabled = true
+		nodeID := *ethAcctAddr
```
j0sh, line 197: Do we want to be sending the Eth address?

j0sh (commented):
This looks pretty good, but will cause merge conflicts with O Discovery in a number of places -- I've tried to point them out. Do you think you can extract the relevant parts atop OD?

core/transcoder.go
darkdarkdragon: my mistake - here I meant to check error that is returned from `parseURI`, will fix

core/orchestrator.go
darkdarkdragon: That's very good! Eric wanted to change segment's length to 2s second, but I wasn't able to create new `test.ts` video to satisfy tests...

eth/eventservices/jobservice.go
darkdarkdragon: go 1.11 gives error here, and that makes unit tests to fail, hence the change

cmd/livepeer/livepeer.go
darkdarkdragon: it is public anyway, so why not? I want some identifier of node, to be able in future measure how many streams goes through node concurrently

darkdarkdragon:
@j0sh is OD PR in useable state? If so I'll rebase my PR on top of it.

j0sh:
> is OD PR in useable state? 

Yes, it's usable in conjunction with the offchain PR https://github.com/livepeer/go-livepeer/pull/610.

Some things might change depending on how testing/review goes, but in general it works pretty well.

core/transcoder.go
darkdarkdragon: fixed

darkdarkdragon:
I've rebased it

j0sh:
This is looking good, did a few things in preparation for the merge:
* Rebased onto master
* Squashed the fixup commits into the originals
* Moved the `SegLen` declaration back into `mediaserver.go` since it's not used anywhere else. The 2-second segment change is still there. https://github.com/livepeer/go-livepeer/pull/612/files#diff-236217e7f473dfbde9134c6a391f3408L49

j0sh (approved):

ericxtang (approved):
LGTM 🚢

PR #613, opened 2018-12-07
mediaserver: Loosen getHLSSegment path requirements.

j0sh:
We don't need fully formed StreamIDs anymore, so look up the OS session from the path prefix.

Fixes an issue exposed by O Discovery with standalone Ts where a complete StreamID was not provided to OS, making the `getHLSSegmentHandler` function break. This loosens up the StreamID requirement to use arbitrary prefixes instead.

- Split the path and use the first part as a session lookup key
- Remove newly unused funciton `parseStreamIDFromSeg` 
- Add more comments and logging

Run the nodes in:
* offchain mode
* Against the O Discovery branch
* With standalone T talking to an O

**Checklist:**

### DIFF
--- a/server/mediaserver.go
+++ b/server/mediaserver.go
@@ -12,7 +12,6 @@ import (
 	"math/rand"
 	"net/http"
 	"net/url"
-	"path"
 	"regexp"
 	"strings"
 	"sync"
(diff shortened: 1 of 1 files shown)


ericxtang (approved):
LGTM 🚢

PR #617, opened 2018-12-10
cmd/livepeer: Don't show testing flags in usage string

### LINKED ISSUE #615: Gitcoin Bounty: Remove flags from "testing" package in golang
**Describe the bug**
The Livepeer node contains a list of flags for the command line.  However, if you run `livepeer -h`, there are a lot of flags introduced by the `testing` package.  

This is because the `testing` package introduces testing-specific flags for things like benchmarking, profiling, etc.  These flags are introduced when the `init()` function is called in the `testing` package.  (More information can be found [here](https://stackoverflow.com/questions/37240065/some-how-all-sorts-of-test-argument-flags-showing-up-in-go-program-usage)

For this bounty, your task is to re-structure the code so that `testing` is not imported in non `_test.go` files, and that the test flags will no longer be shown when you run `livepeer -h`.

**To Reproduce**
1. Build the Livepeer node from the master branch by first running `./install_ffmpeg.sh`, and then `go build cmd/livepeer/livepeer.go`.  This should give you an executable.
2. Run `./livepeer -h`
3. You should see a lot of testing-related flags like "test.bench", "test.benchtime", etc.

**Expected behavior**
Testing-related flags should not be available when you run `livepeer -h`

**Additional context**
Please join the [Discord](https://discord.gg/q6XrfwN) and ask your questions in the #dev channel if you have more questions about this bounty.

speunz:

1. <s>Minor refactor to move out the `testing` package import from non `*_test.go` files. This was not enough to fix the issue though.</s> (removed - not necessary)
2. Reset the default FlagSet in main() because the `go-ipfs` package depends on `go-testutil` which  links the `testing` package, which in turn adds the testing flags. This fixes the issue, though I believe that the `go-ipfs` or `go-testutil` should fix the root cause.

All updates related to `cmd/livepeer`.

1. <s>Remove the `common.WaitAssert()` function because it's not used.</s>
2. <s>Move test utilities from `common/testutil.go` to `common/db_test.go` and remove `common/testutil.go`</s>
3. In the `main()` function, reset the default FlagSet before adding go-livepeer's command line flags. This removes the testing flags indirectly added by the `go-ipfs` dependency on `go-testutil` (eg. [go-testutil/gen.go](https://github.com/livepeer/go-livepeer/blob/master/vendor/gx/ipfs/QmVvkK7s5imCiq3JVbL3pGfnhcCnf3LrFJPF4GE2sAoGZf/go-testutil/gen.go#L9))

1. Built `cmd/livepeer` and ran `livepeer -h` to verify that there were no testing flags in the output
2. Ran tests and compared with tests in master. All tests that passed in master passed in this branch.

Fixes #615 

**Checklist:**

<s>- [ ] README and other documentation updated</s> (not applicable)
- [ ] Node runs in OSX and devenv *Edit* Is this necessary since the changes only affect test cases? I don't have access to OSX unfortunately.

### DIFF
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -31,6 +31,7 @@ import (
 	"github.com/livepeer/go-livepeer/drivers"
 	"github.com/livepeer/go-livepeer/eth"
 	"github.com/livepeer/go-livepeer/eth/eventservices"
+
 	//"github.com/livepeer/go-livepeer/ipfs" until we re-enable IPFS
 	lpmon "github.com/livepeer/go-livepeer/monitor"
 	"github.com/livepeer/go-livepeer/server"
@@ -49,6 +50,10 @@ const RpcPort = "8935"
 const CliPort = "7935"
 
 func main() {
+	// Override the default flag set since there are dependencies that
+	// incorrectly add their own flags (specifically, due to the 'testing'
+	// package being linked)
+	flag.CommandLine = flag.NewFlagSet(os.Args[0], flag.ExitOnError)
 	flag.Set("logtostderr", "true")
 
 	usr, err := user.Current()


ericxtang:
Hi @speunz - the change currently breaks the test suite - you can see it by running `./test.sh`.  I think we can revert the changes in the `common` package for now.  Thanks for doing the bounty!

speunz:
@ericxtang apologies, yes that indeed caused the tests to fail.  I was running the tests with `go test ./...` which resulted in the same build failures in the same packages in both master and the feature branch during tests and it didn't run the tests that actually broke.

I've removed the commit to the `common` package changes and force pushed the branch. The `test.sh` script now runs successfully.

Thanks for the review!

ericxtang (approved):
🚢

ericxtang:
This change is due to the fact that some of the dependencies for Livepeer import `testing` into its main runtime.

j0sh:
Is it just IPFS? Can we just not import IPFS anymore since it's disabled for now?

ericxtang:
Still imported for the IPFS OS driver...

j0sh:
Which can't really be used at the moment? We could comment out everything?

ericxtang:
Hmm... I'd rather not remove a feature because of this bug. Maybe future ipfs releases will get rid of the this dependency.

Issue #616, opened 2018-12-10
On-Chain Orchestrator Discovery


j0sh:
**Is your feature request related to a problem? Please describe.**
The next task in orchestrator discovery is to use on-chain information to discover active orchestrators. This can be broken up into a few steps.

**Describe the solution you'd like**
- [ ] Modify [`offchainOrchestrator`](https://github.com/livepeer/go-livepeer/blob/a8362ac3655062328d95be07976d5f90ccbc136f/discovery/discovery.go#L13) to store a list of URIs, rather than a single URI
  * Perhaps `offchainOrchestrator` should be renamed `offchainOrchestrators` or something else entirely.
- [ ] Modify CLI arguments to accept a list of orchestrators.
  * Perhaps via multiple `-orchAddrs` flags, or a single comma-separated `-orchAddr` flag.
  * Long term, it'd be good to clean up the CLI flags processing and/or implement a config file with a 1-to-1 correspondence to the CLI flags -- similar to Bitcoin and Bitcoin-dervied nodes
  * This is an "addition" to the Orchestrator Discovery RP but still seems useful.
- [ ] In the `discovery.GetOrchestrators` function (invoked at the beginning of the broadcast): parallel invoke `rpc.GetOrchestrator` for all orchestrators (or first N, perhaps N=10). Gather replies. If there are still pending requests after M seconds (say, M=3), time out the remaining requests and return all valid responses.
  * The job will use the first valid entry in the list returned from `discovery.GetOrchestrators`. Ensure list is constructed in the order that responses are received.
  * Modulo goroutine scheduling, this should generally approximate the lowest latency connection.
  * Pricing information is not really relevant for MVS so we won't incorporate it for now.
- [ ] If no orchestrator(s) specified via flags, invoke `eth.RegisteredTranscoders` at node startup time, and populate `offchainOrchestrator` list with resulting transcoders. This could be implemented as another constructor that returns an `offchainOrchestrator` instance. Perhaps it can be called  `NewOffchainOrchestratorFromOnchainList` but I'm open to anything less verbose.

**Describe alternatives you've considered**
* Omit specifying a list of orchestrators via CLI and **only** take the list from the blockchain.
* Define a separate struct rather than reusing/modifying `offchainOrchestrator`
* Only return the first response rather than waiting up to M seconds.
[...]

PR #619, opened 2018-12-11
Add Multiple URIS to Offchain Orchestrators & edit discovery.GetOrchestrators

angyangie:

See here: https://github.com/livepeer/go-livepeer/issues/616

Running orchestrator and broadcaster

### DIFF
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -31,6 +31,7 @@ import (
 	"github.com/livepeer/go-livepeer/drivers"
 	"github.com/livepeer/go-livepeer/eth"
 	"github.com/livepeer/go-livepeer/eth/eventservices"
+
 	//"github.com/livepeer/go-livepeer/ipfs" until we re-enable IPFS
 	lpmon "github.com/livepeer/go-livepeer/monitor"
 	"github.com/livepeer/go-livepeer/server"
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -201,17 +201,6 @@ func main() {
 		n.NodeType = core.BroadcasterNode
 	}
 
-	if n.NodeType == core.BroadcasterNode {
-		if len(orchAddresses) <= 0 {
-			n.OrchestratorPool = discovery.NewOffchainOrchestratorFromOnchainList(n)
-		} else {
-			n.OrchestratorPool = discovery.NewOffchainOrchestrator(n, orchAddresses)
-		}
-		if n.OrchestratorPool == nil {
-			glog.Errorf("No orchestrator specified; transcoding will not happen")
-		}
-	}
-
 	if *offchain {
 		glog.Infof("***Livepeer is in off-chain mode***")
 	} else {
@@ -335,6 +324,17 @@ func main() {
 			base = n.ServiceURI.String()
 		}
 		drivers.NodeStorage = drivers.NewMemoryDriver(base)
+	}
+
+	if n.NodeType == core.BroadcasterNode {
+		if len(orchAddresses) > 0 {
+			n.OrchestratorPool = discovery.NewOrchestratorPool(n, orchAddresses)
+		} else if !*offchain {
+			n.OrchestratorPool = discovery.NewOnchainOrchestratorPool(n)
+		}
+		if n.OrchestratorPool == nil {
+			glog.Errorf("No orchestrator specified; transcoding will not happen")
+		}
 	}
 
 	//Create Livepeer Node
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -98,6 +98,8 @@ func main() {
 		return
 	}
 
+	var orchAddresses []string
+
 	if *orchAddr != "" {
 		if *standaloneTranscoder && *orchSecret == "" {
 			glog.Error("Running a standalone transcoder requires both -orchAddr and -orchSecret")
@@ -105,7 +107,12 @@ func main() {
 		} else if *standaloneTranscoder {
 			transcoder = true
 		}
-		*orchAddr = defaultAddr(*orchAddr, "127.0.0.1", RpcPort)
+
+		orchAddresses = strings.Split(*orchAddr, ",")
+		for i := range orchAddresses {
+			orchAddresses[i] = strings.TrimSpace(orchAddresses[i])
+			orchAddresses[i] = defaultAddr(orchAddresses[i], "127.0.0.1", RpcPort)
+		}
 	}
 
 	if *rinkeby {
@@ -184,19 +191,23 @@ func main() {
 	} else if transcoder {
 		n.NodeType = core.TranscoderNode
 		glog.Info("***Livepeer is in transcoder mode ***")
-		server.RunTranscoder(n, *orchAddr)
+		server.RunTranscoder(n, orchAddresses[0])
 		return
 	} else {
 		n.NodeType = core.BroadcasterNode
 	}
 
 	if n.NodeType == core.BroadcasterNode {
-		if *orchAddr == "" {
+		if orchAddresses[0] == "" {
 			glog.Info("No orchestrator specified; transcoding will not happen")
 		} else {
-			n.OrchestratorSelector = discovery.NewOffchainOrchestrator(n, *orchAddr)
-			if n.OrchestratorSelector == nil {
-				return
+			if len(orchAddresses) > 0 {
+				n.OrchestratorSelector = discovery.NewOffchainOrchestrator(n, orchAddresses[0])
+				if n.OrchestratorSelector == nil {
+					return
+				}
+			} else {
+				n.OrchestratorSelector = discovery.NewOffchainOrchestratorFromOnchainList(n)
 			}
 		}
 	}
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -100,7 +100,7 @@ func main() {
 
 	var orchAddresses []string
 
-	if *orchAddr != "" {
+	if len(*orchAddr) > 0 {
 		if *standaloneTranscoder && *orchSecret == "" {
 			glog.Error("Running a standalone transcoder requires both -orchAddr and -orchSecret")
 			return
@@ -191,24 +191,24 @@ func main() {
 	} else if transcoder {
 		n.NodeType = core.TranscoderNode
 		glog.Info("***Livepeer is in transcoder mode ***")
-		server.RunTranscoder(n, orchAddresses[0])
+		if len(orchAddresses) > 0 {
+			server.RunTranscoder(n, orchAddresses[0])
+		} else {
+			glog.Errorf("No orchestrator specified; transcoding will not happen")
+		}
 		return
 	} else {
 		n.NodeType = core.BroadcasterNode
 	}
 
 	if n.NodeType == core.BroadcasterNode {
-		if orchAddresses[0] == "" {
-			glog.Info("No orchestrator specified; transcoding will not happen")
+		if len(orchAddresses) <= 0 {
+			n.OrchestratorPool = discovery.NewOffchainOrchestratorFromOnchainList(n)
 		} else {
-			if len(orchAddresses) > 0 {
-				n.OrchestratorSelector = discovery.NewOffchainOrchestrator(n, orchAddresses[0])
-				if n.OrchestratorSelector == nil {
-					return
-				}
-			} else {
-				n.OrchestratorSelector = discovery.NewOffchainOrchestratorFromOnchainList(n)
-			}
+			n.OrchestratorPool = discovery.NewOffchainOrchestrator(n, orchAddresses)
+		}
+		if n.OrchestratorPool == nil {
+			glog.Errorf("No orchestrator specified; transcoding will not happen")
 		}
 	}
 
--- a/discovery/discovery.go
+++ b/discovery/discovery.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"net/url"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/livepeer/go-livepeer/core"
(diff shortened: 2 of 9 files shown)


net/interface.go
```diff
@@ -9,7 +9,7 @@ import (
 )
 
 type OrchestratorSelector interface {
-	GetOrchestrators(int) ([]*OrchestratorInfo, error)
+	GetOrchestrators(int) ([][]*OrchestratorInfo, error)
```
j0sh: The interface should stay the same. The other changes to `offchainOrchestrators` is what's enabling us to return more than one `OrchestratorInfo` here.

cmd/livepeer/livepeer.go
```diff
@@ -31,6 +31,7 @@ import (
 	"github.com/livepeer/go-livepeer/drivers"
 	"github.com/livepeer/go-livepeer/eth"
 	"github.com/livepeer/go-livepeer/eth/eventservices"
+
```
j0sh, line 34: stray?

angyangie: Siii :)

net/interface.go
angyangie: Interesting. How can we return the info for multiple Orchestrators here (assuming that's what we want)?

j0sh:
We already return an array of OrchestratorInfos

cmd/livepeer/livepeer.go
angyangie: Actually, that's a space automatically added by Visual Code. Must be because of the commented out import

cmd/livepeer/livepeer.go
```diff
@@ -183,19 +191,23 @@ func main() {
...
-		server.RunTranscoder(n, *orchAddr)
+		server.RunTranscoder(n, orchAddresses[0])
 		return
 	} else {
 		n.NodeType = core.BroadcasterNode
 	}
 
 	if n.NodeType == core.BroadcasterNode {
-		if *orchAddr == "" {
+		if orchAddresses[0] == "" {
```
j0sh: Better to check `len(orchAddresses) <= 0` ; I think this will crash with an empty array

cmd/livepeer/livepeer.go
```diff
@@ -183,19 +191,23 @@ func main() {
...
 			glog.Info("No orchestrator specified; transcoding will not happen")
```
j0sh: This is probably where you want to create `NewOffchainOrchestratorFromOnchainList`

cmd/livepeer/livepeer.go
```diff
@@ -183,19 +191,23 @@ func main() {
...
 		} else {
-			n.OrchestratorSelector = discovery.NewOffchainOrchestrator(n, *orchAddr)
-			if n.OrchestratorSelector == nil {
-				return
+			if len(orchAddresses) > 0 {
+				n.OrchestratorSelector = discovery.NewOffchainOrchestrator(n, orchAddresses[0])
```
j0sh: This probably just hasn't been implemented yet, but we're going to pass in the entire `orchAddresses` list here, right?

discovery/discovery.go
```diff
@@ -10,23 +13,80 @@ import (
...
+
+	for _, orch := range orchestrators {
+		serviceURI := orch.ServiceURI
+		if !strings.HasPrefix(serviceURI, "http") {
+			serviceURI = "https://" + serviceURI
+		}
+		uri, err := url.ParseRequestURI(serviceURI)
+		if err != nil {
+			glog.Error("Could not parse orchestrator URI: ", err)
+			return nil
```
j0sh: Perhaps just `continue` here rather than return from the function empty-handed

discovery/discovery.go
```diff
@@ -10,23 +13,80 @@ import (
...
+	orchChan := make(chan struct{})
+
+	getOrchInfo := func(uri *url.URL, bcast server.Broadcaster, i int) {
+		tinfo, err := server.GetOrchestratorInfo(bcast, uri)
+		if err != nil {
+			glog.Errorf("Error getting OrchestratorInfo for uri: %v", uri)
+			return
+		}
+		orchInfo = append(orchInfo, tinfo)
+		if numOrchestrators <= i {
```
j0sh: This probably won't work as expected; it'd signal the chan for the `i`th request, regardless of whether the other requests have completed. I think something like this will work for `GetOrchestrators`:

```golang
numResp := 0
numSuccessResp := 0
respLock := sync.Mutex{}
orchInfos := []*net.OrchestratorInfo{}
getOrchInfo := func(uri *url.URL) {
   // should we also modify this to take a cancelable context?
  info, err = server.GetOrchestratorInfo(o.bcast, uri)
  respLock.Lock()
  defer respLock.Unlock()
  numResp++
  if err != nil {
    orchInfos = append(orchInfos, info)
    numSuccessResp++
  }
  // Signal the chan if we've received the expected number of successful responses
  // or if we've exhausted the list of URIs
  // If we ever look at other metrics for selecting O, then we'll need to gather
  // *all* responses, rather than stop at numOrchestrators. But this is OK for now.
  if numSuccessResp >= numOrchestrators || numResp >= len(o.uris) {
    orchChan <- struct{}{}
  }
}
for _, uri := range o.uris {
  go getOrchInfo(uri)
}
select {
  case <- ctx.Done():
    ...
  case <- orchChan:
    ...
}
```

cmd/livepeer/livepeer.go
angyangie: That's right. I'm working on that now :)

discovery/discovery.go
```diff
@@ -10,23 +14,97 @@ import (
...
+
+func NewOffchainOrchestratorFromOnchainList(node *core.LivepeerNode) *offchainOrchestrators {
+	orchestrators, err := node.Eth.RegisteredTranscoders()
 	if err != nil {
-		glog.Error("Could not parse orchestrator URI: ", err)
+		glog.Error(err)
 		return nil
 	}
+
+	var uris []*url.URL
```
j0sh: Rather than repeat most of the same code below, could we invoke `NewOffchainOrchestrator` with the orchestrators list from `node.Eth.RegisteredTranscoders` ?

```golang
func NewOffchainOrchestratorFromOnchainList(node *core.LivepeerNode) *offchainOrchestrators {
	orchestrators, err := node.Eth.RegisteredTranscoders()
	if err != nil {
		glog.Error(err)
		return nil
	}
        return NewOffchainOrchestrators(node, orchestrators)
}
```

cmd/livepeer/livepeer.go
```diff
@@ -183,17 +191,17 @@ func main() {
 	} else if transcoder {
 		n.NodeType = core.TranscoderNode
 		glog.Info("***Livepeer is in transcoder mode ***")
-		server.RunTranscoder(n, *orchAddr)
+		server.RunTranscoder(n, orchAddresses[0])
```
j0sh: We probably need a length check here

discovery/discovery.go
```diff
@@ -10,23 +14,97 @@ import (
...
+		uri, err := url.ParseRequestURI(addr)
+		if err != nil {
+			glog.Error("Could not parse orchestrator URI: ", err)
+			continue
+		}
+		uris = append(uris, uri)
+	}
+
+	bcast := core.NewBroadcaster(node)
+	return &offchainOrchestrators{bcast: bcast, uri: uris}
```
j0sh: I wonder if we should log something, somewhere, if `uris` is empty, just to notify the user.

cmd/livepeer/livepeer.go
```diff
@@ -183,17 +191,17 @@ func main() {
...
 		return
 	} else {
 		n.NodeType = core.BroadcasterNode
 	}
 
 	if n.NodeType == core.BroadcasterNode {
-		if *orchAddr == "" {
-			glog.Info("No orchestrator specified; transcoding will not happen")
+		if len(orchAddresses) <= 0 {
+			n.OrchestratorSelector = discovery.NewOffchainOrchestratorFromOnchainList(n)
```
j0sh: We probably need a couple more checks around this area:

* Check for offchain mode somewhere (indicated by `n.Eth = nil`). Can either be done right here, or within `NewOffchainOrchestratorFromOnchainList`. I'm inclined to do it within the latter function, and return nil in offchain mode.
* Handle the case where `n.OrchestratorSelector` is nil, and accordingly log "No orchestrator specified; transcoding will not happen". This check should cover the results from either `NewOffchainOrchestrator` or `NewOffchainOrchestratorFromOnchainList`

cmd/livepeer/livepeer.go
angyangie: Since we're checking for `transcoder` on line 191, and transcoder can only be true if [`if len(*orchAddr) > 0`](https://github.com/livepeer/go-livepeer/blob/dd5dc14ed2976655b299c2c9f7fffbe9c1e0f365/cmd/livepeer/livepeer.go#L103) (<-- I'll change the line linked here to check for the len of the array), then is there a danger to not checking len here?

discovery/discovery.go
angyangie: Agreed! No need for repeated code here

discovery/discovery.go
angyangie: Sure, that sounds reasonable!

cmd/livepeer/livepeer.go
j0sh: Good point. One has to be careful to follow the logic -- `transcoder` can be set with either the `-transcoder` flag or the `-standaloneTranscoder` flag, and the `-transcoder` flag results in line 189 being checked, whereas `-standaloneTranscoder` can only enable `transcoder`  mode if `orchAddr` is non-empty.

Overall it's a bit fragile and we probably should refactor this soon for robustness. The good thing is, if this flow ever breaks, it'll likely crash at start-up time, making the issue easier to diagnose.

cmd/livepeer/livepeer.go
angyangie: Is there a danger to `n.Eth = client` not being set? We're returning an error when calling `eth.NewClient(ethcommon.HexToAddress(*ethAcctAddr), keystoreDir, backend, ethcommon.HexToAddress(*controllerAddr), EthTxTimeout)`, which should cause the function to return. What do you think?

j0sh: Just checked: it does crash at this point if `-offchain` mode is set on the broadcaster.

cmd/livepeer/livepeer.go
angyangie: Ah yes, we'll have to refactor soon! I'll do a len check here until we can figure out a better system for these flags. Meanwhile, that's right, an immediate-crash is easy to diagnose!

discovery/discovery.go
```diff
@@ -10,23 +14,97 @@ import (
...
+	orchChan := make(chan struct{})
+	numResp := 0
+	numSuccessResp := 0
+	respLock := sync.Mutex{}
+	getOrchInfo := func(uri *url.URL) {
+
+		externalCtx, cancelCtx := context.WithCancel(context.Background())
+		defer cancelCtx()
+
+		info, err := server.GetOrchestratorInfo(externalCtx, o.bcast, uri)
```
j0sh: We should use `ctx`here rather than `externalCtx` (which can be removed)

cmd/livepeer/livepeer.go
```diff
@@ -183,20 +191,23 @@ func main() {
 	} else if transcoder {
 		n.NodeType = core.TranscoderNode
 		glog.Info("***Livepeer is in transcoder mode ***")
-		server.RunTranscoder(n, *orchAddr)
+		if len(orchAddresses) > 0 {
+			server.RunTranscoder(n, orchAddresses[0])
+		}
```
j0sh, line 198: Log if `orchAddresses <= 0` ?

cmd/livepeer/livepeer.go
```diff
@@ -183,20 +191,23 @@ func main() {
...
 		} else {
-			n.OrchestratorSelector = discovery.NewOffchainOrchestrator(n, *orchAddr)
-			if n.OrchestratorSelector == nil {
-				return
-			}
+			n.OrchestratorSelector = discovery.NewOffchainOrchestrator(n, orchAddresses)
+		}
+		if n.OrchestratorSelector == nil {
+			glog.Errorf("No orchestrator specified; transcoding will not happen")
+			return
```
j0sh: This will quit early for offchain mode; we can just continue running instead of returning for view-only streams.

discovery/discovery.go
```diff
@@ -10,23 +14,92 @@ import (
...
+
+		info, err := server.GetOrchestratorInfo(ctx, o.bcast, uri)
+		if err != nil {
+			glog.Errorf("Error getting OrchestratorInfo for uri: %v", uri)
+		}
+		respLock.Lock()
+		defer respLock.Unlock()
+		numResp++
+		orchInfos = append(orchInfos, info)
+		numSuccessResp++
```
j0sh: orchInfo should only be appended to and `numSucessResp` incremented if the response was actually successful (`err` nil)

discovery/discovery.go
j0sh: Just realized there's a typo here -- the `err  != nil` check should be `err == nil` !

discovery/discovery.go
```diff
@@ -10,23 +14,92 @@ import (
...
+
+type offchainOrchestrators struct {
+	uri   []*url.URL
 	bcast server.Broadcaster
 }
 
-func NewOffchainOrchestrator(node *core.LivepeerNode, address string) *offchainOrchestrator {
-	addr := "https://" + address
-	uri, err := url.ParseRequestURI(addr)
+func NewOffchainOrchestrator(node *core.LivepeerNode, addresses []string) *offchainOrchestrators {
```
eladmallel: When we use this function we assign the return value into `n.OrchestratorSelector`. Perhaps this function should be named `NewOffchainOrchestratorSelector`?

j0sh: Agreed that the naming is a bit imprecise. Generally, things are getting to be a mouthful, such as with `NewOffchainOrchestratorFromOnchainList`.

Trying to think of more concise options. Perhaps `NewOrchestratorSelector` for this base case since the offchain part is a given anyway, and `NewOnchainOrchestratorSelector` for orchestrators registered on-chain? Is there a better name for the interface than `OrchestratorSelector` ?

eladmallel: Perhaps `OrchestratorPool`, since it's a bit shorter? :)

j0sh: I love it 👍

discovery/discovery.go
angyangie: ah yessss. I'll follow your format from your example above!

discovery/discovery.go
angyangie: Agree that `OrchestratorPool` conveys meaning better than `OrchestratorSelector`! I'll change

cmd/livepeer/livepeer.go
```diff
@@ -183,20 +191,24 @@ func main() {
...
-		if *orchAddr == "" {
-			glog.Info("No orchestrator specified; transcoding will not happen")
+		if len(orchAddresses) <= 0 {
+			n.OrchestratorPool = discovery.NewOffchainOrchestratorFromOnchainList(n)
 		} else {
-			n.OrchestratorSelector = discovery.NewOffchainOrchestrator(n, *orchAddr)
-			if n.OrchestratorSelector == nil {
-				return
-			}
+			n.OrchestratorPool = discovery.NewOffchainOrchestrator(n, orchAddresses)
```
j0sh: `discovery.NewOrchestratorPool` ?

cmd/livepeer/livepeer.go
```diff
@@ -183,20 +191,24 @@ func main() {
...
 		return
 	} else {
 		n.NodeType = core.BroadcasterNode
 	}
 
 	if n.NodeType == core.BroadcasterNode {
-		if *orchAddr == "" {
-			glog.Info("No orchestrator specified; transcoding will not happen")
+		if len(orchAddresses) <= 0 {
+			n.OrchestratorPool = discovery.NewOffchainOrchestratorFromOnchainList(n)
```
j0sh: `discovery.NewOnchainOrchestratorPool` ?

eth/stubclient.go
```diff
@@ -98,8 +102,18 @@ func (e *StubClient) GetDelegatorUnbondingLock(addr common.Address, unbondingLoc
...
+func (e *StubClient) RegisteredTranscoders() ([]*lpTypes.Transcoder, error) {
+	var addresses = []string{"https://127.0.0.1:8936", "https://127.0.0.1:8937", "https://127.0.0.1:8938"}
+	var transcoders []*lpTypes.Transcoder
+
+	for _, addr := range addresses {
+		transc := &lpTypes.Transcoder{ServiceURI: addr}
+		transcoders = append(transcoders, transc)
+	}
+
+	return transcoders, nil
```
j0sh: This is OK for now. The general pattern we follow for stubbing is:

Define `StubClient` with a list of `Transcoders` (technically `Orchestrators`), eg

```
type StubClient struct {
   ...
   Orchestrators []*lpTypes.Transcoder,
   ...
}
```

As necessary, initialize the `Orchestrators` list as necessary during testing time. Eg, 
```
func TestOnchainOrchestratorPool() {
  var orchestrators []*lpTypes.Transcoder
  // populate `orchestrators` list
  stubclient := eth.StubClient{Orchestrators: orchestrators}
  n := NewLivepeerNode(stubclient, "", nil)
... 
}
```

The rest of the time, the Orchestrators list can be left unset as it doesn't affect the remainder of the test cases using `StubClient`. This way we can adjust the list of orchestrators as needed to match the expectations of the test (eg, perhaps checking the case where we have zero onchain orchestrators.)

test.sh
```diff
@@ -30,6 +30,11 @@ go test -logtostderr=true
 t6=$?
 cd ..
 
+cd discovery
+go test -logtostderr=true
+t7=$?
+cd ..
+
 if (($t1!=0||$t2!=0||$t3!=0||$t4!=0||$t5!=0||$t6!=0))
```
j0sh: Missing check for $t7 here

cmd/livepeer/livepeer.go
```diff
@@ -183,20 +191,24 @@ func main() {
...
+			n.OrchestratorPool = discovery.NewOnchainOrchestratorPool(n)
 		} else {
-			n.OrchestratorSelector = discovery.NewOffchainOrchestrator(n, *orchAddr)
-			if n.OrchestratorSelector == nil {
-				return
-			}
+			n.OrchestratorPool = discovery.NewOrchestratorPool(n, orchAddresses)
+		}
+		if n.OrchestratorPool == nil {
+			n.OrchestratorPool = discovery.NewOrchestratorPool(n, []string{})
```
j0sh: It should be safe to leave `OrchestratorPool` empty since we check for that specific case and return an `ErrDiscovery` if the pool is empty. That feels like a more consistent way of differentiating between the states of the program (eg, no orchestrators specified here) rather than mixing in with the possibility that some orchestrators *were* specified but none were available in a timely fashion during the `GetOrchestrators` call.

test.sh
angyangie: No wonder when tests failed correct error message wasn't showing up ...

cmd/livepeer/livepeer.go
angyangie: I see why the distinction is important. We return an ErrDiscovery in `startBroadcast`, only place we actually need to call `GetOrchestrators` from the OrchestratorPool.

eth/stubclient.go
angyangie: That makes a lot of sense. I didn't feel very comfortable adding a hard-coded value to `RegisteredTranscoders` in `StubClient`. This is a better way of doing it. I've changed it!

j0sh (approved):
Looks great! :ship:

discovery/discovery.go
```diff
@@ -10,23 +14,90 @@ import (
 	"github.com/golang/glog"
 )
 
-type offchainOrchestrator struct {
-	uri   *url.URL
+const GetOrchestratorsTimeoutLoop = 3 * time.Second
+
+type offchainOrchestrators struct {
```
ericxtang: Should we find another name for this?  `offchainOrchestrators` seems less appropriate than something like `orchestratorUrls` (it's especially strange to return `offchainOrchestrators` for `NewOnchainOrchestratorPool`)

angyangie: Ah yes, I was looking at that and thought it strange too, but hadn't done anything about it. Would orchestratorPool make more sense? What do you think @j0sh

j0sh: Sure, either `orchestratorPool` or `orchestratorUrls` works

angyangie: Awesome! I like `orchestratorPool`, since we already have `NewOnchainOrchestratorPool`, and `NewOrchestratorPool`

PR #625, opened 2018-12-14
StreamID Refactoring

### LINKED ISSUE #622: StreamID Refactoring

StreamID handling should be refactored to increase internal consistency, increase flexibility in StreamID selection and improve readability of the StreamID itself.

**Background Context**

Currently, the StreamID construction looks like:
```<ManifestID><Rendition>```
Where the ManifestID is a 256-bit random number encoded in hexadecimal, and the Rendition follows the stringified form of the [preset profiles](https://github.com/livepeer/lpms/blob/master/ffmpeg/videoprofile.go) from LPMS.

Concretely, a StreamID looks like this:

```
a22d1c5916b02bf5f18e79de3eb017c32e021deb13153276c4472b5760993c34P720p30fps16x9
```

Where `a22d1c5916b02bf5f18e79de3eb017c32e021deb13153276c4472b5760993c34` is the ManifestID and `P720p30fps16x9` is the rendition.

Under the old protocol, the StreamID needed to be globally unique. This is no longer the case for streamflow, where the StreamID is only used to group related segments together locally for playback [1].

**Current Status**

Object storage introduced further changes with segment indexing during storage and playback.  StreamID references within storage are defined using a slash to delimit the different parts of a segment name: 

```
<manifestID>/<rendition>/<seq>.ts
a22d1c5916b02bf5f18e79de3eb017c32e021deb13153276c4472b5760993c34/P720p30fps16x9/9.ts
```

Additionally, we are shifting away from the requirement to use ["complete" StreamIDs](https://github.com/livepeer/go-livepeer/pull/613) in the playback path by loosening up what's considered "acceptable" for object lookup.

[1] There is currently some spillover whereas the orchestrator currently uses the B-provided ManifestID to index into the transcoding loop. This should be changed to eliminate O's dependency on the ManifestID. Eliminating this dependency has a few positive implications for O that can be elaborated on separately.

Let's double down on this gradual shift by loosening up how StreamIDs are defined. Not only does this increase consistency with StreamID handling in OS and playback, this should also give us more friendly streamIDs: they can be shorter, and perhaps even human readable.

- [ ] Define a StreamID struct that stores a ManifestID and Rendition separately. Allows for faster access to each part, rather than slicing a string.
- [ ] Use slashes as ManifestID/Rendition separators when expressing the StreamID as a string. We need a delimiter for arbitrary-length or human readable strings. This was not required earlier with the ManifestID as a fixed length string.
- [ ] Remove references to `VideoID`. This is a leftover from when the ManifestID was composed of the libp2p node ID and a random video ID.
- [ ] Make the randomized ManifestID shorter, such as a 32-bit number. This is OK since it is only needed on B, and collisions with active streams are unlikely.

Leave things as-is.

### PR DESCRIPTION
**Pull request by j0sh**
## What does this pull request do?
Refactors the StreamID and fixes miscellaneous issues found along the way.

### Randomize segmenter output filename

Avoid passing in the HLS Stream ID, since that could be user
supplied, and we don't want to open ourselves up to directory
traversal attacks.

### Return  root-relative URLs from memory store

Fixes an issue that happens when we store playlists hierarchically. That is, given a playlist at `<ManifestID>/<Rendition>.m3u8`, the segment at `<ManifestID>/<Rendition>/<Seq>.ts` would have resulted in the HLS player doing a lookup for
    
    <ManifestID>/<ManifestID>/<Rendition>/<Seq>.ts
    
due to the relative path given by the playlist. Fix this by making the returned path unconditionally root-relative; this results in the correct prefix being applied to the request. Local URLs now become
    
    /source/<ManifestID>/<Rendition>/<Seq>.ts

### Refactor StreamIDs

Loosen up the definition of a StreamID to "any valid string" . This greatly simplifies error handling. While a StreamID can technically be empty, the media server will generate a random Stream ID if an empty ID is passed in via the `hlsStreamID` RTMP query parameter.

By convention, we split StreamIDs into two components: the "manifest ID" and the "rendition" . Either can be empty, but non-empty renditions will generally have a non-empty manifest ID.

Components are separated by path-like slashes. When converting from a string into a StreamID, we use the first component as the ManifestID. The rest is considered part of the rendition. For example:

```
String: abc/def/ghi`
ManifestID: abc
Rendition: def/ghi
```

The test cases cover this extensively, including some "odd" scenarios with multiple leading slashes, whitespace, etc. At some point we may want to consider adding a sanitization/canonicalization step, but it doesn't seem to be required at the moment.

* Unit testing
* Manual testing

Fixes https://github.com/livepeer/go-livepeer/issues/622

**Checklist:**

### DIFF
--- a/core/stream_test.go
+++ b/core/stream_test.go
@@ -22,6 +22,7 @@ func TestSegmentFlatten(t *testing.T) {
 	}
 }
 
+/*
 func TestStreamID(t *testing.T) {
 	vid := RandomVideoID()
 	id, err := MakeStreamID(vid, ffmpeg.P144p30fps16x9.Name)
@@ -95,3 +96,4 @@ func TestManifestID(t *testing.T) {
 		t.Error("did not expect manifestd to be valid")
 	}
 }
+*/
--- a/core/stream_test.go
+++ b/core/stream_test.go
@@ -2,6 +2,8 @@ package core
 
 import (
 	"bytes"
+	"fmt"
+	"math/rand"
 	"testing"
 
 	ethcommon "github.com/ethereum/go-ethereum/common"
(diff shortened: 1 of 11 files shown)

### CODE REVIEW

File: core/streamdata.go [hunk 1]
```diff
@@ -42,88 +43,47 @@ func (md *SegTranscodingMetadata) Flatten() []byte {
...
-//StreamID is VideoID|Rendition
-type StreamID string
-
-func MakeStreamID(id []byte, rendition string) (StreamID, error) {
-	if len(id) != HashLength || rendition == "" {
-		return "", ErrStreamID
-	}
-	return StreamID(fmt.Sprintf("%x%v", id, rendition)), nil
-}
+type ManifestID string
```
yondonfu, line 54: After this PR, do we have any methods that use the `ManifestID` type as a receiver? Do we need a dedicated type here or would just using a regular `string` suffice which would avoid any additional type casting between `string` and `ManifestID`?

core/streamdata.go
```diff
@@ -42,88 +43,47 @@ func (md *SegTranscodingMetadata) Flatten() []byte {
 	return buf
 }
 
-func RandomVideoID() []byte {
-	return RandomIdGenerator(HashLength)
-}
-
-func RandomIdGenerator(length uint) []byte {
+var RandomIdGenerator = func(length uint) []byte {
```
yondonfu, line 46: Perhaps a more idiomatic Go alternative to allowing `RandomIdGenerator` to be monkey patched would be to pass around a `RandomIdGenerator` interface with the default implementation reflecting this function's logic and then using a mock implementation in tests.

core/stream_test.go
```diff
@@ -23,75 +25,91 @@ func TestSegmentFlatten(t *testing.T) {
 }
 
 func TestStreamID(t *testing.T) {
-	vid := RandomVideoID()
-	id, err := MakeStreamID(vid, ffmpeg.P144p30fps16x9.Name)
-	if err != nil {
-		t.Error("Error making StreamID ", err)
+	rand.Seed(123)
```
yondonfu, line 28: Should `RandomIdGenerator` be monkey patched so we don't have to deal with seeding at all?

core/stream_test.go
```diff
@@ -23,75 +25,91 @@ func TestSegmentFlatten(t *testing.T) {
...
+	mid := RandomManifestID()
+	profile := ffmpeg.P144p30fps16x9
+
+	// Test random manifest ID generation
+	if string(mid) != "17b336b6" {
+		t.Error("Unexxpected ManifestID ", mid)
```
yondonfu: "Unexxpected" -> "Unexpected"?

core/streamdata.go
```diff
@@ -42,88 +43,47 @@ func (md *SegTranscodingMetadata) Flatten() []byte {
...
-func (id *StreamID) GetVideoID() []byte {
-	if len(*id) < 2*HashLength {
-		return nil
-	}
-	vid, err := hex.DecodeString(string((*id)[:2*HashLength]))
-	if err != nil {
-		return nil
-	}
-	return vid
+type StreamID struct {
```
yondonfu, line 57: Could we add some comments for this struct describing how each of the fields are used? In particular, I'm thinking of some of the changes introduced in #618 i.e. using the `Rendition` field to hold a RTMP stream key

server/mediaserver.go
```diff
@@ -144,24 +145,12 @@ func createRTMPStreamIDHandler(s *LivepeerServer) func(url *url.URL) (strmID str
 		//Create a StreamID
```
yondonfu: Update comment to reflect that we're creating a ManifestID here and not a StreamID?

yondonfu (changes requested):

core/streamdata.go
j0sh: We could do that, but the boilerplate (defining an interface, defining default + test structs and implementations, etc) doesn't seem justified for one-off cases like this.

If we end up needing to monkey patch several things at once, then we could probably extract those out into a more general interface. Here, "the function is the interface".

core/stream_test.go
j0sh: Sure, we probably should still test the bytes generator itself with a predictable seed. Will add a separate test for that.

core/streamdata.go
j0sh: > After this PR, do we have any methods that use the ManifestID type as a receiver

I don't believe so. We do use ManifestID to index into transcode-loops on the orchestrator, but using the ManifestID in the orchestrator is something I'd really like to get rid of sooner rather than later.

Regardless, I sort of like having a dedicated type for ManifestID. If we ever need to impose stricter constraints on the format of the string, then we'll have stronger assurances that a given ManifestID type is well-formed.

Can we still declare the type but hide the implementation to prevent type casting? Also, I'd say that casting from `ManifestID -> string` is okay, but `string -> ManifestID` is a red flag. I think the only place we do a lot of this is during testing, which seems OK, but let me know if you think otherwise.

Totally acknowledge that this is a bit inconsistent right now since the Rendition is a string. Perhaps a new type for Rendition as well? 😅

core/stream_test.go
j0sh: The RandomIdGenerator actually isn't monkey patched at this point yet; the change will land in the failover PR.

core/streamdata.go
yondonfu: Re: the future flexibility of having a dedicated type for ManifestID if we impose formatting restrictions - yeah that is fair. The type casting isn't thaaat big of a deal (was just wondering if it was worth eliminating it if we could) though I do agree we should avoid any instances of `string -> ManifestID` outside of tests.

> Can we still declare the type but hide the implementation to prevent type casting?

What did you have in mind?

server/mediaserver.go
j0sh: In the rendition PR, we actually create a full StreamID in this function again, as part of the stream key for playback protection.

j0sh: Two seconds after posting, I see where the context may have been confusing. Sorry about that -- fixing the comment, thanks!

core/streamdata.go
j0sh: > What did you have in mind?

I'm actually not sure if it's possible in golang with primitive types; the goal would be to prevent type casting entirely. It It'd be similar in spirit to hiding non-exported opaque structs behind a public interface. If we used an empty interface here, I think we'd also lose the ability to use the type as keys in maps.

```
foo.h:
struct foo_t;

foo.c:
struct foo_t {
  int a;
};
```

Here, only `foo.c` has access to the definition of foo_t. All other files see it as a opaque type of `struct foo_t` with a hidden definition. Even in C, this doesn't work for primitive types, but it is possible to make opaque primitive types with OCaml modules:

```
foo.mli:
type foo

foo.ml:
type foo = string
```

server/mediaserver.go
j0sh: Fixed

core/streamdata.go
j0sh: Fixed

core/stream_test.go
j0sh: Rearranged things a bit and it's in the streamid PR now; fixed here https://github.com/livepeer/go-livepeer/pull/625/commits/5f857b2d9c896f2cfc252fa85fd968749638a208

core/stream_test.go
j0sh: Fixed in https://github.com/livepeer/go-livepeer/pull/625/commits/5f857b2d9c896f2cfc252fa85fd968749638a208

core/streamdata.go
j0sh: Left the ManifestID type as-is for now, and the Rendition is still a string. We can revisit later.

yondonfu (approved):
LGTM pending update with master

Issue #630, opened 2018-12-14
PM Recipient


eladmallel:
**Is your feature request related to a problem? Please describe.**
Support O's state management and ticket handling for PM interactions.

**Describe the solution you'd like**
A `Recipient` struct and logic under the pm package, including:

- Random numbers state management
- Using Ticket and TicketBroker to validate B's funds, validate and redeem tickets

This implementation will be accompanied by unit tests while stubbing out dependencies such at Eth.

**Describe alternatives you've considered**
Nothing to add.

**Additional context**
Nothing to add.

PR #638, opened 2018-12-17
Ticket struct and related interfaces

### LINKED ISSUE #626: TicketBroker Interface
As part of the PM implementation, both B and O need to communicate with the on-chain TicketBroker contract to manage and read deposits and penalty escrows, redeem tickets, etc.

Create an `TicketBroker` interface that exposes the following functionality:

- fundDeposit
- fundPenanltyEscrow
- approveSigners
- requestSignersRevocation
- redeemWinningTicket
- unlock & cancelUnlock
- withdraw

This interface will be provided as a dependency to both the `Sender` and `Receiver` structs that represent both sides of a PM exchange. The real implementation use our Eth client. For testing purposes we can provide a mocked implementation of `TicketBroker`.

We considered just using Eth client, but felt the interface is a minor addition that improves testability.

**Additional context**
For additional context on TicketBroker, check out the smart contract [here](https://github.com/livepeer/protocol/blob/pm/contracts/pm/TicketBroker.sol).

### LINKED ISSUE #628: Ticket Struct
We need a simple way to represent a PM ticket to encapsulate all relevant parameters, as well as convenience of packaging the data for signing.

A `Ticket` struct with all the fields a ticket requires:

- recipient
- sender
- faceValue
- winProb
- senderNonce
- recipientRandHash
- senderSig
- additional data for ticket expiration

It will also have a method for
- Getting the hash representation of the ticket to be used for signing.
- Ticket validation
- Check for winning ticket

None really :)

**Additional context**
No additional context at the moment.

yondonfu:

Adds common PM ticket related code that will be used by both a sender and a recipient in the PM protocol.

- Adds the `pm` package
- Adds a `Ticket` struct that reflects the data included in on-chain redemption transactions. Note: The auxiliary data for expiration validation is omitted for the time being until a concrete decision is made on our desired expiration policy
- Adds a `Broker` interface that serves as an abstraction over the on-chain smart contract and its functions
- Adds a `Validator` interface with functions for validating a ticket and checking for a winning ticket. The `BrokerValidator` implementation uses an implementation of the `Broker` interface to check for used tickets (replay prevention) and an implementation of the `SigVerifier` interface to validate signatures over ticket hashes
- Adds a `SigVerifier` interface with functions for verifying signatures. The `BrokerSigVerifier` implementation uses an implementation of the `Broker` interface to check for approved signers when verifying a signature.

- Unit tests for each new component with stubbed out dependencies

- Fixes #626 
- Fixes #628

**Checklist:**

### DIFF
--- a/pm/sigverifier.go
+++ b/pm/sigverifier.go
@@ -12,7 +12,7 @@ import (
 type SigVerifier interface {
 	// Verify checks if a provided signature over a message
 	// is valid for a given ETH address
-	Verify(addr ethcommon.Address, sig []byte, msg []byte) bool
+	Verify(addr ethcommon.Address, msg, sig []byte) bool
 }
 
 // ApprovedSigVerifier is an implementation of the SigVerifier interface
@@ -33,26 +33,26 @@ func NewApprovedSigVerifier(broker Broker) *ApprovedSigVerifier {
 
 // Verify checks if a provided signature over a message
 // is valid for a given ETH address
-func (sv *ApprovedSigVerifier) Verify(addr ethcommon.Address, sig []byte, msg []byte) (bool, error) {
+func (sv *ApprovedSigVerifier) Verify(addr ethcommon.Address, msg, sig []byte) bool {
 	personalMsg := fmt.Sprintf("\x19Ethereum Signed Message:\n%d%s", 32, msg)
 	personalHash := crypto.Keccak256([]byte(personalMsg))
 
 	pubkey, err := crypto.SigToPub(personalHash, sig)
 	if err != nil {
-		return false, err
+		return false
 	}
 
 	rec := crypto.PubkeyToAddress(*pubkey)
 
 	if addr == rec {
 		// If recovered address matches, return early
-		return true, nil
+		return true
 	}
 
 	approved, err := sv.broker.IsApprovedSigner(addr, rec)
 	if err != nil {
-		return false, err
+		return false
 	}
 
-	return approved, nil
+	return approved
 }
--- a/pm/sigverifier.go
+++ b/pm/sigverifier.go
@@ -1,0 +1,58 @@
+package pm
+
+import (
+	"fmt"
+
+	ethcommon "github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/crypto"
+)
+
+// SigVerifier is an interface which describes an object capable
+// of verification of ECDSA signatures produced by ETH addresses
+type SigVerifier interface {
+	// Verify checks if a provided signature over a message
+	// is valid for a given ETH address
+	Verify(addr ethcommon.Address, sig []byte, msg []byte) bool
+}
+
+// ApprovedSigVerifier is an implementation of the SigVerifier interface
+// that relies on an implementation of the Broker interface to provide a registry
+// mapping ETH addresses to approved signer sets. This implementation will
+// recover a ETH address from a signature and check if the recovered address
+// is approved
+type ApprovedSigVerifier struct {
+	broker Broker
+}
+
+// NewApprovedSigVerifier returns an instance of an approved signature verifier
+func NewApprovedSigVerifier(broker Broker) *ApprovedSigVerifier {
+	return &ApprovedSigVerifier{
+		broker: broker,
+	}
+}
+
+// Verify checks if a provided signature over a message
+// is valid for a given ETH address
+func (sv *ApprovedSigVerifier) Verify(addr ethcommon.Address, sig []byte, msg []byte) (bool, error) {
+	personalMsg := fmt.Sprintf("\x19Ethereum Signed Message:\n%d%s", 32, msg)
+	personalHash := crypto.Keccak256([]byte(personalMsg))
+
+	pubkey, err := crypto.SigToPub(personalHash, sig)
+	if err != nil {
+		return false, err
+	}
+
+	rec := crypto.PubkeyToAddress(*pubkey)
+
+	if addr == rec {
+		// If recovered address matches, return early
+		return true, nil
+	}
+
+	approved, err := sv.broker.IsApprovedSigner(addr, rec)
+	if err != nil {
+		return false, err
+	}
+
+	return approved, nil
+}
--- a/pm/sigverifier_test.go
+++ b/pm/sigverifier_test.go
@@ -25,52 +25,44 @@ func TestVerify(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	signerPrivKey, err := crypto.GenerateKey()
+	unapprovedSignerPrivKey, err := crypto.GenerateKey()
 	if err != nil {
 		t.Fatal(err)
 	}
 
-	signer := crypto.PubkeyToAddress(signerPrivKey.PublicKey)
+	unapprovedSignerSig, err := crypto.Sign(personalHash, unapprovedSignerPrivKey)
 
-	signerSig, err := crypto.Sign(personalHash, signerPrivKey)
-
-	b := &stubBroker{
-		usedTickets:     make(map[ethcommon.Hash]bool),
-		approvedSigners: make(map[ethcommon.Address]bool),
-	}
+	b := newStubBroker()
 
 	sv := NewApprovedSigVerifier(b)
 
 	// Test invalid signature for non-approved signer
-	valid, err := sv.Verify(sender, signerSig, msg)
-	if err != nil {
-		t.Fatal(err)
-	}
-
-	if valid {
-		t.Fatal("expected invalid signature for non-approved signer")
+	if valid := sv.Verify(sender, msg, unapprovedSignerSig); valid {
+		t.Error("expected invalid signature for non-approved signer")
 	}
 
 	// Test valid signature for approved signer
-	// Approve signer
-	b.ApproveSigners([]ethcommon.Address{signer})
-
-	valid, err = sv.Verify(sender, signerSig, msg)
+	approvedSignerPrivKey, err := crypto.GenerateKey()
 	if err != nil {
 		t.Fatal(err)
 	}
 
-	if !valid {
-		t.Fatal("expected valid signature for approved signer")
+	approvedSigner := crypto.PubkeyToAddress(approvedSignerPrivKey.PublicKey)
+
+	approvedSignerSig, err := crypto.Sign(personalHash, approvedSignerPrivKey)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Approve signer
+	b.ApproveSigners([]ethcommon.Address{approvedSigner})
+
+	if valid := sv.Verify(sender, msg, approvedSignerSig); !valid {
+		t.Error("expected valid signature for approved signer")
 	}
 
 	// Test valid signature for sender
-	valid, err = sv.Verify(sender, senderSig, msg)
-	if err != nil {
-		t.Fatal(err)
-	}
-
-	if !valid {
-		t.Fatal("expected valid signature for sender")
+	if valid := sv.Verify(sender, msg, senderSig); !valid {
+		t.Error("expected valid signature for sender")
 	}
 }
--- a/pm/sigverifier_test.go
+++ b/pm/sigverifier_test.go
@@ -1,0 +1,76 @@
+package pm
+
+import (
+	"fmt"
+	"testing"
+
+	ethcommon "github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/crypto"
+)
+
+func TestVerify(t *testing.T) {
+	msg := []byte("foo")
+	personalMsg := fmt.Sprintf("\x19Ethereum Signed Message:\n%d%s", 32, msg)
+	personalHash := crypto.Keccak256([]byte(personalMsg))
+
+	senderPrivKey, err := crypto.GenerateKey()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	sender := crypto.PubkeyToAddress(senderPrivKey.PublicKey)
+
+	senderSig, err := crypto.Sign(personalHash, senderPrivKey)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	signerPrivKey, err := crypto.GenerateKey()
+	if err != nil {
+		t.Fatal(err)
+	}
+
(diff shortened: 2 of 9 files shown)


pm/ticket.go
```diff
@@ -0,0 +1,109 @@
...
+
+	// Sender is the ETH address of sender
+	Sender ethcommon.Address
+
+	// FaceValue represents the pay out to
+	// the recipient if the ticket wins
+	FaceValue *big.Int
+
+	// WinProb represents how likely a ticket will win
+	WinProb *big.Int
```
j0sh, line 32: Just wondering, how were you planning on representing a probability as an integer?

yondonfu: `WinProb` would just be an integer in the range 0 through 2^256 - 1 with the winning ticket check defined as `H(SIG(H(T)), recipientRand) < T.WinProb)` (the hash result is also in the range 0 through 2^256 - 1). Since `MaxWinProb = 2^256 - 1`, each `WinProb` value is can be thought of as a probability represented as the fraction `WinProb / MaxWinProb`.

pm/sigverifier.go
```diff
@@ -0,0 +1,58 @@
...
+	}
+
+	rec := crypto.PubkeyToAddress(*pubkey)
+
+	if addr == rec {
+		// If recovered address matches, return early
+		return true, nil
+	}
+
+	approved, err := sv.broker.IsApprovedSigner(addr, rec)
```
eladmallel, line 52: I'm wondering about the performance implications this will have. Under the following conditions I think this might be problematic:

- If we assume a naive implementation, this would result in a function call to an ETH node every time, which could take a while
- If ticket validation is a blocker for processing a segment

Will these conditions be true when we implement `Broker` and integrate ticket validation in the `ServeSegment` flow?

pm/sigverifier_test.go
```diff
@@ -0,0 +1,76 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	signerPrivKey, err := crypto.GenerateKey()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	signer := crypto.PubkeyToAddress(signerPrivKey.PublicKey)
```
eladmallel: Perhaps rename `signer` to `unapprovedSigner` to increase readability? - I noticed we later approve `signer`, but perhaps then we can just create a new signer instance and name that one `approvedSigner`?

pm/sigverifier_test.go
```diff
@@ -0,0 +1,76 @@
...
+
+	signerSig, err := crypto.Sign(personalHash, signerPrivKey)
+
+	b := &stubBroker{
+		usedTickets:     make(map[ethcommon.Hash]bool),
```
eladmallel: Why do we need to instantiate `usedTickets` for this test?

pm/sigverifier_test.go
```diff
@@ -0,0 +1,76 @@
...
+		approvedSigners: make(map[ethcommon.Address]bool),
+	}
+
+	sv := NewApprovedSigVerifier(b)
+
+	// Test invalid signature for non-approved signer
+	valid, err := sv.Verify(sender, signerSig, msg)
+	if err != nil {
+		t.Fatal(err)
```
eladmallel: I think that starting from this line forward you want to switch from `t.Fatal` to `t.Fail`, since I believe `Fatal` stops the test function on that line and does not proceed, versus `Fail` that allows the test to continue, therefore not masking the rest of the checks in this function.

wdyt?

j0sh: We also use `t.Error` elsewhere in the codebase to indicate "fail but continue"

pm/ticket.go
```diff
@@ -0,0 +1,109 @@
...
+const (
+	addressSize = 20
+	uint256Size = 32
+	bytes32Size = 32
+)
+
+// Broker is an interface which serves as an abstraction over an on-chain
+// smart contract that handles the administrative tasks in a probabilistic micropayment protocol
+// including processing deposits and pay outs
+type Broker interface {
```
eladmallel: What are the pros/cons of having the `Broker` interface in the `ticket.go` file, versus creating a new file like `broker.go` ?

pm/validator.go
```diff
@@ -0,0 +1,79 @@
...
+func NewValidator(addr ethcommon.Address, broker Broker, sigVerifier SigVerifier) Validator {
+	return &validator{
+		addr:        addr,
+		broker:      broker,
+		sigVerifier: sigVerifier,
+	}
+}
+
+// ValidateTicket checks if a ticket is valid
+func (v *validator) ValidateTicket(ticket *Ticket, sig []byte, recipientRand *big.Int) error {
```
eladmallel, line 46: Wondering when/where should we introduce the HMAC-based validation?

pm/validator.go
```diff
@@ -0,0 +1,79 @@
...
+
+	if (ticket.Sender == ethcommon.Address{}) {
+		return fmt.Errorf("invalid ticket sender")
+	}
+
+	if crypto.Keccak256Hash(ethcommon.LeftPadBytes(recipientRand.Bytes(), uint256Size)) != ticket.RecipientRandHash {
+		return fmt.Errorf("invalid preimage provided for hash commitment recipientRandHash")
+	}
+
+	used, err := v.broker.IsUsedTicket(ticket)
```
eladmallel: Is `IsUsedTicket` focused on checking for random number replays, winning ticket replays, or both?

Intuitively seems to me that `Broker` should not be aware of local replay protection like `senderNonce` state, so based on your answer to the question above, would love to get a clearer sense of how we're organizing all the validations we need to run on each incoming ticket.

wdyt?

pm/validator.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+
+	return nil
+}
+
+// IsWinningTicket checks if a ticket won
+// Note: This method does not check if a ticket is valid which is done using IsValidTicket
+// A ticket wins if:
+// H(SIG(H(T)), T.RecipientRand) < T.WinProb
+func (v *validator) IsWinningTicket(ticket *Ticket, sig []byte, recipientRand *big.Int) bool {
```
eladmallel, line 70: Just want to double check our approach regarding winning ticket replay protection: seems like that's not something we should worry about in the client because we invalidate `recipiendRand` values every time we submit a ticket, so there's no way an attempt to replay a winning ticket on O will actually work. It will be blocked because it will have a no-longer-valid `recipientRandHash` value.

Correct?

pm/validator_test.go
```diff
@@ -0,0 +1,167 @@
...
+		Recipient:         sender,
+		Sender:            ethcommon.Address{},
+		FaceValue:         big.NewInt(0),
+		WinProb:           big.NewInt(0),
+		SenderNonce:       0,
+		RecipientRandHash: recipientRandHash,
+	}
+
+	if err := v.ValidateTicket(ticket, sig, recipientRand); err == nil {
+		t.Fatal("expected invalid recipient (non-null address)")
```
eladmallel: Change to `t.Fail` or `t.Error`? (writing this comment once for all occurrences in this test file)

pm/validator_test.go
```diff
@@ -0,0 +1,167 @@
...
+	// Test used ticket
+	ticket = &Ticket{
+		Recipient:         recipient,
+		Sender:            sender,
+		FaceValue:         big.NewInt(0),
+		WinProb:           big.NewInt(0),
+		SenderNonce:       0,
+		RecipientRandHash: recipientRandHash,
+	}
+	// Set ticket as used in stub broker
```
eladmallel: So what makes this ticket used? - sounds like there should be a call to some stub function setting this ticket to used?

pm/validator_test.go
```diff
@@ -0,0 +1,167 @@
...
+	b.RedeemWinningTicket(ticket, sig, recipientRand)
+
+	if err := v.ValidateTicket(ticket, sig, recipientRand); err == nil {
```
eladmallel: General concern: shouldn't we validate that the error we're getting is the specific error we expect?

pm/stub.go
```diff
@@ -0,0 +1,137 @@
...
+	"math/big"
+
+	ethcommon "github.com/ethereum/go-ethereum/common"
+)
+
+type stubSigVerifier struct {
+	shouldVerify bool
+}
+
+func (sv *stubSigVerifier) SetShouldVerify(shouldVerify bool) {
```
eladmallel: I think this function name is a bit confusing. Perhaps changing it to `SetVerifyResult` is a bit clearer?

pm/sigverifier.go
yondonfu: Great point. Here is a summary of an offline discussion about this:

An ETH call to check for approved signers to validate every ticket would create a performance hit for ticket validation which affects segment processing performance. In the short term for MVS, we can remove the approved signer feature client side and instead the recipient will just check that the signature corresponds to the sender address in the ticket (only requires client side cryptographic operations). In the longer term, we could implement an caching layer that replicates state from the on-chain contract so that a recipient can access a sender's approved signers without an ETH call.

pm/validator.go
yondonfu: `IsUsedTicket` checks if the broker has already processed this particular winning ticket (i.e. query the smart contract). Here is a summary of an offline discussion about this:

The broker keeps track of used winning tickets in order to protect a sender from a malicious recipient that tries to replay a winning ticket over and over again. Since the validator is the recipient, we do not need to protect against this scenario. However, we do need to protect against the scenario in which a sender purposefully sends used winning tickets to the recipient which results in the recipient being unable to redeem them. 

If we assume that the recipient also:

- Handles replay prevention for invalid already winning `recipientRand` values
- Uses an HMAC based on a secret generated on start up and the sender's address to validate `recipientRand` values

then the recipient does not need to call `IsUsedTicket` when validating a ticket.

The recipient will not accept tickets with an old winning `recipientRand` which means no old winning tickets created by the current sender will be accepted. Sender A can try to use winning tickets sent by sender B, but these tickets will not pass the HMAC validation step since the tickets include a different sender. If the recipient restarts, thereby clearing its current set of invalid already winning `recipientRand` values, it will generate a new secret so if a sender tries to replay an old `recipientRand` value, it will fail the HMAC validation which uses the recipient's new secret as the key.

Thus, the recipient will reject old winning tickets obviating the need for the call to `IsUsedTicket` which is a plus because it removes an ETH call from the ticket validation process which could've been a performance bottleneck.

pm/validator.go
yondonfu: Yep see my comment https://github.com/livepeer/go-livepeer/pull/638#discussion_r summarizing this conclusion.

pm/validator.go
yondonfu: My plan was to introduce HMAC validation as a part of https://github.com/livepeer/go-livepeer/issues/630. The motivation behind this was to have the `Validator` object be responsible for validating a ticket given a provided `recipientRand` value - this value could be randomly generated, could be generated using a HMAC based construction, etc. The HMAC validation logic would then be held in the recipient which passes the `recipientRand` value generated into the `Validator` object. Let me know if you have any thoughts/concerns on this separation of concerns!

pm/ticket.go
yondonfu: I didn't have a strong reason for keeping the `Broker` interface in `pm/ticket.go`. Since there is not an concrete implementation for this interface yet, I didn't bother putting it into it's own file. I can move this interface into it's own file thought - it seems cleaner and will probably be helpful later on for the concrete implementation

pm/sigverifier.go
j0sh: >  In the longer term, we could implement either an caching layer

One way is to to lookup the list when the first segment of a stream comes in. That list only needs to be retained for the duration of the stream (eg, transcode loop). There'd be a hit for the very first segment, but subsequent lookups can remain local.

I wonder if borrowing the Explorer's GraphQL caching layer would also help with the initial lookup.

pm/validator.go
j0sh: At this point we assume that the given `recipientRand` is valid, and that it's been checked earlier in the flow? Makes sense.

Technically we could also keep the validation contained by making some `ReceipentRandValidator` interface accessible to the validator object, perhaps by passing it in during construction of the `validator` struct. No real preference though; incorporating it in #630 works too.

pm/sigverifier_test.go
yondonfu: I've replaced declarations of `stubBroker` struct literals with a call to the constructor `newStubBroker` which creates a `stubBroker` with all fields initialized and ready for use.

My thought was to try to avoid any accidental usage of `stubBroker` that leads to accessing a nil map.

pm/validator_test.go
yondonfu: This particular test has been removed in de637a5 since we no longer check the broker for used tickets during ticket validation

pm/sigverifier.go
eladmallel: The Graph is definitely worth exploring for this use case. We just need to make sure it has sufficiently efficient response times and no hangups, since we don't want The Graph to be the source of segment processing delays :)

pm/sigverifier_test.go
eladmallel: So what should be the consistent way to do it? `t.Error` or `t.Fail`?

pm/sigverifier.go
yondonfu: Interesting approach. O would likely also want to monitor revocation events for the duration of the stream in order to determine when certain previously approved signers will be considered revoked (and thus their signatures deemed invalid if submitted on-chain). 

Tracking this and other possible areas of exploration in #645

pm/sigverifier_test.go
yondonfu: `t.Error` logs the provided message and then triggers a `t.Fail` https://golang.org/pkg/testing/#T.Error while `t.Fail` just marks the function as having failed without a message logged

eladmallel: Sounds like `t.Error` is the choice then!

eladmallel (approved):
🚢

j0sh (approved):
LGTM!

j0sh:
If this is being blocked because of CI, feel free to cherry pick https://github.com/livepeer/go-livepeer/pull/625/commits/85227b54c92d9202ad8e979532614ef88fcdd2cb and push

Issue #641, opened 2018-12-18
Out-of-band Orchestrator Discovery and Caching


angyangie:
**Is your feature request related to a problem? Please describe.**
This last phase of Orchestrator discovery is to use out-of-band processes to gather information from all Os in the service registry, so broadcasters don’t need to do discovery in real-time when a new stream is created. This can be broken up into a few steps:

**Describe the solution you'd like**
- [x] Build DB schema for caching with the following fields, to start with (Include DB indexes for more efficient lookups ). We'll add pricing information once it's available, if necessary. 

> id INTEGER PRIMARY KEY
> createdAt STRING DEFAULT CURRENT_TIMESTAMP
> updatedAt STRING DEFAULT CURRENT_TIMESTAMP
> serviceURI string
> ethereumAddr string (or varbinary)
> PRIMARY KEY(id, ethereumAddress)

- [x] Cache refreshing:  check for updates from the blockchain every hour. Scan all the rows in that table, and compare them to the refresh. If any rows are missing, add them (no rows will be deleted). Once list is refreshed from the blockchain, we can combine updating the table and inputing new rows for new orchestrators on the list as an "upsert", that is, check for the existence of a row, update it if necessary, otherwise insert

- [x] Integrate GetOrchestrator with DB cache

- [x] Add unit test cases for DB-backed orchestrator selection

**Describe alternatives you've considered**
When thinking about refreshing the table, we could drop in the full table, but not something we want to do in the long run, because it could be that we have good work history with specific orchestrator that's not on the list anymore. Broadcaster should still have access to them.

Future Optimizations/Questions to think about (from @j0sh):
- Only rows that haven't been updated in +24 hours will be updated. 
- At some point we'll also want to query GetOrchestratorInfo to get things like pricing information, but since we aren't storing that in the DB right now it can wait. 
- We could update records in the DB one-by-one and check the results in the DB. We'd need to fetch all Os from the blockchain, then filter out Os that already exist in the DB but aren't due for an update, then call GetOrchestrator for each and populate the DB with the values received from OrchestratorInfo. It's more of a concern when the list of orchestrators grows really large, and we don't need to be pinging them so frequently for data (faceValue/winProb) that changes less frequently. 

**Additional context**
See the O Discovery RP, https://github.com/livepeer/go-livepeer/pull/604, and https://github.com/livepeer/go-livepeer/pull/619 for further details.

j0sh:
[...]

PR #642, opened 2018-12-19
Out of band orchestrator discovery

angyangie:

All changes explained here: https://github.com/livepeer/go-livepeer/issues/641

This one: https://github.com/livepeer/go-livepeer/issues/641

**Checklist:**

### DIFF
--- a/common/db.go
+++ b/common/db.go
@@ -20,7 +20,7 @@ type DB struct {
 
 	// prepared statements
 	selectOrchs                *sql.Stmt
-	updateOrchs                *sql.Stmt
+	updateOrch                 *sql.Stmt
 	updateKV                   *sql.Stmt
 	insertUnbondingLock        *sql.Stmt
 	useUnbondingLock           *sql.Stmt
@@ -56,8 +56,8 @@ var schema = `
 
 	CREATE TABLE IF NOT EXISTS orchestrators (
 		ethereumAddr STRING PRIMARY KEY,
-		createdAt STRING DEFAULT CURRENT_TIMESTAMP,
-		updatedAt STRING DEFAULT CURRENT_TIMESTAMP,
+		createdAt STRING DEFAULT CURRENT_TIMESTAMP NOT NULL,
+		updatedAt STRING DEFAULT CURRENT_TIMESTAMP NOT NULL,
 		serviceURI STRING
 	);
 
@@ -145,13 +145,13 @@ func InitDB(dbPath string) (*DB, error) {
 	d.selectOrchs = stmt
 
 	// updateOrchestrators statement
-	stmt, err = db.Prepare("INSERT OR REPLACE INTO orchestrators(updatedAt, serviceURI, ethereumAddr) VALUES(datetime(), ?, ?)")
+	stmt, err = db.Prepare("INSERT OR REPLACE INTO orchestrators(updatedAt, serviceURI, ethereumAddr, createdAt) VALUES(datetime(), ?1, ?2, (SELECT createdAt FROM orchestrators WHERE ethereumAddr = ?2))")
 	if err != nil {
 		glog.Error("Unable to prepare updateOrchestrators stmt ", err)
 		d.Close()
 		return nil, err
 	}
-	d.updateOrchs = stmt
+	d.updateOrch = stmt
 
 	// updateKV prepared statement
 	stmt, err = db.Prepare("UPDATE kv SET value=?, updatedAt = datetime() WHERE key=?")
@@ -207,8 +207,8 @@ func InitDB(dbPath string) (*DB, error) {
 
 func (db *DB) Close() {
 	glog.V(DEBUG).Info("Closing DB")
-	if db.updateOrchs != nil {
-		db.updateOrchs.Close()
+	if db.updateOrch != nil {
+		db.updateOrch.Close()
 	}
 	if db.selectOrchs != nil {
 		db.selectOrchs.Close()
@@ -265,11 +265,12 @@ func (db *DB) LastSeenBlock() (*big.Int, error) {
 	return big.NewInt(lastSeenBlock), nil
 }
 
-func (db *DB) UpdateOrchs(orch *DBOrch) error {
-	if db == nil {
+func (db *DB) UpdateOrch(orch *DBOrch) error {
+	if db == nil || orch == nil || orch.ServiceURI == "" || orch.EthereumAddr == "" {
 		return nil
 	}
-	_, err := db.updateOrchs.Exec(orch.ServiceURI, orch.EthereumAddr)
+
+	_, err := db.updateOrch.Exec(orch.ServiceURI, orch.EthereumAddr)
 	if err != nil {
 		glog.Error("db: Unable to update orchestrator ", err)
 	}
--- a/common/db.go
+++ b/common/db.go
@@ -19,12 +19,19 @@ type DB struct {
 	dbh *sql.DB
 
 	// prepared statements
+	selectOrchs                *sql.Stmt
+	updateOrchs                *sql.Stmt
 	updateKV                   *sql.Stmt
 	insertUnbondingLock        *sql.Stmt
 	useUnbondingLock           *sql.Stmt
 	unbondingLocks             *sql.Stmt
 	withdrawableUnbondingLocks *sql.Stmt
 	insertWinningTicket        *sql.Stmt
+}
+
+type DBOrch struct {
+	ServiceURI   string
+	EthereumAddr string
 }
 
 type DBUnbondingLock struct {
@@ -46,6 +53,13 @@ var schema = `
 	);
 	INSERT OR IGNORE INTO kv(key, value) VALUES('dbVersion', '{{ . }}');
 	INSERT OR IGNORE INTO kv(key, value) VALUES('lastBlock', '0');
+
+	CREATE TABLE IF NOT EXISTS orchestrators (
+		ethereumAddr STRING PRIMARY KEY,
+		createdAt STRING DEFAULT CURRENT_TIMESTAMP,
+		updatedAt STRING DEFAULT CURRENT_TIMESTAMP,
+		serviceURI STRING
+	);
 
 	CREATE TABLE IF NOT EXISTS unbondingLocks (
 		id INTEGER NOT NULL,
@@ -72,6 +86,10 @@ var schema = `
 
 	CREATE INDEX IF NOT EXISTS idx_winningtickets_sessionid ON winningTickets(sessionID);
 `
+
+func NewDBOrch(serviceURI string, orchAddr string) *DBOrch {
+	return &DBOrch{ServiceURI: serviceURI, EthereumAddr: orchAddr}
+}
 
 func InitDB(dbPath string) (*DB, error) {
 	// XXX need a way to ensure (via unit tests?) that all DB{} fields are
@@ -117,8 +135,26 @@ func InitDB(dbPath string) (*DB, error) {
 		// all good; nothing to do
 	}
 
+	// select all orchestrators updated in the last 24 hours
+	stmt, err := db.Prepare("SELECT serviceURI, ethereumAddr FROM orchestrators WHERE updatedAt >= datetime('now','-1 day')")
+	if err != nil {
+		glog.Error("Unable to select orchestrators updated in the past 24 hours", err)
+		d.Close()
+		return nil, err
+	}
+	d.selectOrchs = stmt
+
+	// updateOrchestrators statement
+	stmt, err = db.Prepare("INSERT OR REPLACE INTO orchestrators(updatedAt, serviceURI, ethereumAddr) VALUES(datetime(), ?, ?)")
+	if err != nil {
+		glog.Error("Unable to prepare updateOrchestrators stmt ", err)
+		d.Close()
+		return nil, err
+	}
+	d.updateOrchs = stmt
+
 	// updateKV prepared statement
-	stmt, err := db.Prepare("UPDATE kv SET value=?, updatedAt = datetime() WHERE key=?")
+	stmt, err = db.Prepare("UPDATE kv SET value=?, updatedAt = datetime() WHERE key=?")
 	if err != nil {
 		glog.Error("Unable to prepare updatekv stmt ", err)
 		d.Close()
@@ -171,6 +207,12 @@ func InitDB(dbPath string) (*DB, error) {
 
 func (db *DB) Close() {
 	glog.V(DEBUG).Info("Closing DB")
+	if db.updateOrchs != nil {
+		db.updateOrchs.Close()
+	}
+	if db.selectOrchs != nil {
+		db.selectOrchs.Close()
+	}
 	if db.updateKV != nil {
 		db.updateKV.Close()
 	}
@@ -221,6 +263,45 @@ func (db *DB) LastSeenBlock() (*big.Int, error) {
 	}
 
 	return big.NewInt(lastSeenBlock), nil
+}
+
+func (db *DB) UpdateOrchs(orch *DBOrch) error {
+	if db == nil {
+		return nil
+	}
+	_, err := db.updateOrchs.Exec(orch.ServiceURI, orch.EthereumAddr)
+	if err != nil {
+		glog.Error("db: Unable to update orchestrator ", err)
+	}
+
+	return err
+}
+
+func (db *DB) SelectOrchs() ([]*DBOrch, error) {
+	if db == nil {
+		return nil, nil
+	}
+
+	rows, err := db.selectOrchs.Query()
+	defer rows.Close()
+	if err != nil {
+		glog.Error("db: Unable to get orchestrators updated in the last 24 hours", err)
+		return nil, err
+	}
+	orchs := []*DBOrch{}
+	for rows.Next() {
+		var orch DBOrch
+		var serviceURI string
+		var ethereumAddr string
+		if err := rows.Scan(&serviceURI, &ethereumAddr); err != nil {
+			glog.Error("db: Unable to fetch orchestrator ", err)
+			continue
+		}
+		orch.ServiceURI = serviceURI
+		orch.EthereumAddr = ethereumAddr
+		orchs = append(orchs, &orch)
+	}
+	return orchs, nil
 }
 
 func (db *DB) InsertUnbondingLock(id *big.Int, delegator ethcommon.Address, amount, withdrawRound *big.Int) error {
--- a/common/db_test.go
+++ b/common/db_test.go
@@ -106,6 +106,13 @@ func TestDBVersion(t *testing.T) {
 	}
 }
 
+func NewStubOrch(serviceURI string, ethereumAddr string) *DBOrch {
+	return &DBOrch{
+		ServiceURI:   serviceURI,
+		EthereumAddr: ethereumAddr,
+	}
+}
+
 func profilesMatch(j1 []ffmpeg.VideoProfile, j2 []ffmpeg.VideoProfile) bool {
 	if len(j1) != len(j2) {
 		return false
@@ -116,6 +123,47 @@ func profilesMatch(j1 []ffmpeg.VideoProfile, j2 []ffmpeg.VideoProfile) bool {
 		}
 	}
 	return true
+}
+
+func TestDBOrchestrators(t *testing.T) {
+	dbh, dbraw, err := TempDB(t)
+	defer dbh.Close()
+	defer dbraw.Close()
+
+	orch := NewStubOrch("127.0.0.1:8936", "0xde0B295669a9FD93d5F28D9Ec85E40f4cb697BAe")
+	err = dbh.UpdateOrchs(orch)
+
+	if err != nil {
+		t.Error("Unexpected error in update orchs: ", err)
+	}
+
+	orchs, err := dbh.SelectOrchs()
+	if err != nil || len(orchs) != 1 || orchs[0].ServiceURI != orch.ServiceURI {
+		t.Error("Unexpected error in update orchs: ", err)
+	}
+
+	orchUpdate := NewStubOrch("127.0.0.1:8937", "0xde0B295669a9FD93d5F28D9Ec85E40f4cb697BAe")
+	err = dbh.UpdateOrchs(orchUpdate)
+
+	if err != nil {
+		t.Error("Unexpected error in update orchs: ", err)
+	}
+
+	orchs, err = dbh.SelectOrchs()
+	if err != nil || len(orchs) != 1 || orchs[0].ServiceURI != orchUpdate.ServiceURI {
+		t.Error("Unexpected error in update orchs: ", err)
+	}
+
+	orchAdd := NewStubOrch("127.0.0.1:8938", "0x5eD8Cee6b63b1c6AFce3AD7c92f4fD7E1B8fAd9F")
+	err = dbh.UpdateOrchs(orchAdd)
+
+	if err != nil {
+		t.Error("Unexpected error in update orchs: ", err)
+	}
+	orchs, err = dbh.SelectOrchs()
+	if err != nil || len(orchs) != 2 || orchs[1].ServiceURI != orchAdd.ServiceURI {
+		t.Error("Unexpected error in update orchs: ", err)
+	}
 }
 
 func TestDBUnbondingLocks(t *testing.T) {
--- a/common/db_test.go
+++ b/common/db_test.go
@@ -106,13 +106,6 @@ func TestDBVersion(t *testing.T) {
 	}
 }
 
-func NewStubOrch(serviceURI string, ethereumAddr string) *DBOrch {
-	return &DBOrch{
-		ServiceURI:   serviceURI,
-		EthereumAddr: ethereumAddr,
-	}
-}
-
 func profilesMatch(j1 []ffmpeg.VideoProfile, j2 []ffmpeg.VideoProfile) bool {
 	if len(j1) != len(j2) {
 		return false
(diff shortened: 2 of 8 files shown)


common/db.go
```diff
@@ -223,6 +265,45 @@ func (db *DB) LastSeenBlock() (*big.Int, error) {
...
+	if err != nil {
+		glog.Error("db: Unable to get orchestrators updated in the last 24 hours", err)
+		return nil, err
+	}
+	orchs := []*DBOrch{}
+	for rows.Next() {
+		var orch DBOrch
+		var serviceURI string
+		var ethereumAddr string
+		if err := rows.Scan(&serviceURI, &ethereumAddr); err != nil {
```
j0sh, line 297: We can avoid using the serviceURI and ethereumAddr variables here, eg with `rows.Scan(&orch.ServiceURI, &orch.EthereumAddr)`

discovery/db_discovery.go
```diff
@@ -0,0 +1,92 @@
...
+	node *core.LivepeerNode
+}
+
+func NewDBOrchestratorPoolCache(node *core.LivepeerNode) *DBOrchestratorPoolCache {
+	if node.Eth == nil {
+		glog.Error("Could not refresh DB list of orchestrators: LivepeerNode nil")
+		return nil
+	}
+
+	ticker := time.NewTicker(1 * time.Hour)
```
j0sh: Maybe make `1 * time.Hour` a constant?

Also, what do you think of making this overridable so we can explicitly emit ticks during testing? See a general example of this pattern: https://github.com/livepeer/go-livepeer/pull/657#discussion_r244917469

discovery/db_discovery.go
```diff
@@ -0,0 +1,92 @@
...
+func CacheDBOrchs(node *core.LivepeerNode, orchs []*lpTypes.Transcoder) ([]*common.DBOrch, error) {
+	var dbOrchs []*common.DBOrch
+	if orchs == nil {
+		glog.Error("No new DB orchestrators created: no orchestrators found onchain")
+		return dbOrchs, nil
+	}
+
+	for _, orch := range orchs {
+		dbOrch := EthOrchToDBOrch(orch)
+		if err := node.Database.UpdateOrchs(dbOrch); err != nil {
```
j0sh: Just a note for later. At some point we'll probably want to do one of a few things (or even a combination of both):

1. Call `GetOrchestratorInfo` prior to inserting to DB to receive and insert ticket parameters. We need to figure out how to schedule this for large O sets (sequentially? asynchronously? in batches?)
2. Consolidate multiple INSERTs into a single statement, eg `INSERT INTO orchestrators (updatedAt, serviceURI, ethereumAddr) VALUES(datetime(), abc, def), (datetime(), ghi, jkl), (datetime(), mno, pqr)`. We want to minimize the amount of time SQLite spends writing to disk. Unfortunately, it seems we can't use prepared statements for this, so we might need to construct the query at runtime and take special care to escape all the input prior to execution. See https://stackoverflow.com/questions/49156991/sqlite-range-bind-or-column-out-of-range-for-insert-statement

common/db.go
```diff
@@ -223,6 +265,45 @@ func (db *DB) LastSeenBlock() (*big.Int, error) {
 	return big.NewInt(lastSeenBlock), nil
 }
 
+func (db *DB) UpdateOrchs(orch *DBOrch) error {
```
j0sh: This function (and the associated prepared statement) only updates a single orchestrator; perhaps it should be named `UpdateOrch` instead?

common/db_test.go
```diff
@@ -118,6 +125,47 @@ func profilesMatch(j1 []ffmpeg.VideoProfile, j2 []ffmpeg.VideoProfile) bool {
...
+	defer dbraw.Close()
+
+	orch := NewStubOrch("127.0.0.1:8936", "0xde0B295669a9FD93d5F28D9Ec85E40f4cb697BAe")
+	err = dbh.UpdateOrchs(orch)
+
+	if err != nil {
+		t.Error("Unexpected error in update orchs: ", err)
+	}
+
+	orchs, err := dbh.SelectOrchs()
```
j0sh, line 154: Perhaps also add a test for selecting from an empty table?

common/db.go
j0sh: Also probably needs an `orch == nil` check here

common/db_test.go
```diff
@@ -118,6 +125,47 @@ func profilesMatch(j1 []ffmpeg.VideoProfile, j2 []ffmpeg.VideoProfile) bool {
 	return true
 }
 
+func TestDBOrchestrators(t *testing.T) {
+	dbh, dbraw, err := TempDB(t)
+	defer dbh.Close()
+	defer dbraw.Close()
+
+	orch := NewStubOrch("127.0.0.1:8936", "0xde0B295669a9FD93d5F28D9Ec85E40f4cb697BAe")
+	err = dbh.UpdateOrchs(orch)
```
j0sh: Perhaps a test for a `nil` orch since that'll crash `UpdateOrchs` at the moment.

discovery/db_discovery.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+func CacheDBOrchs(node *core.LivepeerNode, orchs []*lpTypes.Transcoder) ([]*common.DBOrch, error) {
+	var dbOrchs []*common.DBOrch
+	if orchs == nil {
+		glog.Error("No new DB orchestrators created: no orchestrators found onchain")
+		return dbOrchs, nil
+	}
+
+	for _, orch := range orchs {
```
j0sh, line 77: Do we need a `if orch == nil { continue }` check here? Only reason for asking is because EthOchToDBOrch has one, so not sure if there's a case where the `orchs` list could contain a nil entry.

common/db_test.go
```diff
@@ -118,6 +125,47 @@ func profilesMatch(j1 []ffmpeg.VideoProfile, j2 []ffmpeg.VideoProfile) bool {
 	return true
 }
 
+func TestDBOrchestrators(t *testing.T) {
```
eladmallel: I'd like to suggest breaking this down into more succinct or "surgical" test functions, where we state more clearly the scenario and the expected outcome. IMO taking the approach of smaller functions and the specific naming paradigm nudges us to think more carefully about all the cases we want to cover. See this recent example: https://github.com/livepeer/go-livepeer/pull/655/files#diff-c20584ae20b069ed08dc6769a1c19e77.

In general I like structuring the names of test functions in 3 parts:

- The code under test (usually a function name)
- The scenario
- The expected result

While it's not necessarily how we normally name functions in Go, for the sake of readability I also like to separate these parts with an underscore.

Of course one of the downsides could be repetitive setup code, but that can be mitigated with some solid helper functions.

wdyt?

common/db.go
```diff
@@ -117,8 +135,26 @@ func InitDB(dbPath string) (*DB, error) {
...
+	stmt, err := db.Prepare("SELECT serviceURI, ethereumAddr FROM orchestrators WHERE updatedAt >= datetime('now','-1 day')")
+	if err != nil {
+		glog.Error("Unable to select orchestrators updated in the past 24 hours", err)
+		d.Close()
+		return nil, err
+	}
+	d.selectOrchs = stmt
+
+	// updateOrchestrators statement
+	stmt, err = db.Prepare("INSERT OR REPLACE INTO orchestrators(updatedAt, serviceURI, ethereumAddr) VALUES(datetime(), ?, ?)")
```
eladmallel: According to [this tutorial](http://www.sqlitetutorial.net/sqlite-replace-statement/) we can shorten the `INSERT OR REPLACE` statement to simply `REPLACE`. wdyt?

eladmallel: Plus, I didn't know about the `REPLACE` feature until just now - thanks for this new learning!

common/db_test.go
```diff
@@ -106,6 +106,13 @@ func TestDBVersion(t *testing.T) {
 	}
 }
 
+func NewStubOrch(serviceURI string, ethereumAddr string) *DBOrch {
```
eladmallel: Why do we need this function? - seems like we can just use `NewDBOrch`

common/db_test.go
```diff
@@ -118,6 +125,47 @@ func profilesMatch(j1 []ffmpeg.VideoProfile, j2 []ffmpeg.VideoProfile) bool {
...
+	if err != nil {
+		t.Error("Unexpected error in update orchs: ", err)
+	}
+
+	orchs, err := dbh.SelectOrchs()
+	if err != nil || len(orchs) != 1 || orchs[0].ServiceURI != orch.ServiceURI {
+		t.Error("Unexpected error in update orchs: ", err)
+	}
+
+	orchUpdate := NewStubOrch("127.0.0.1:8937", "0xde0B295669a9FD93d5F28D9Ec85E40f4cb697BAe")
```
eladmallel: Does the ETH address need to be this specific value? - if not I personally like the idea of using random values, primarily because it clarifies to the test reader that we don't really care what value we use here ([see example](https://github.com/livepeer/go-livepeer/pull/655/files#diff-c20584ae20b069ed08dc6769a1c19e77R803)).

wdyt?

common/db_test.go
```diff
@@ -118,6 +125,47 @@ func profilesMatch(j1 []ffmpeg.VideoProfile, j2 []ffmpeg.VideoProfile) bool {
...
+	err = dbh.UpdateOrchs(orchUpdate)
+
+	if err != nil {
```
eladmallel: Would you like to use the new `testify` library we've been using? - it should allow you to replace all these if blocks with a simple `assert.Nil(err)` ([example here](https://github.com/livepeer/go-livepeer/pull/655/files#diff-c20584ae20b069ed08dc6769a1c19e77R756)). wdyt?

discovery/db_discovery.go
```diff
@@ -0,0 +1,92 @@
...
+			glog.Error("Error updating Orchestrator in DB: ", err)
+			return nil, err
+		}
+		dbOrchs = append(dbOrchs, dbOrch)
+	}
+	return dbOrchs, nil
+
+}
+
+func EthOrchToDBOrch(orch *lpTypes.Transcoder) *common.DBOrch {
```
eladmallel: Do we need to export this function, or can we make it private?

discovery/db_discovery.go
```diff
@@ -0,0 +1,92 @@
...
+	if orch == nil {
+		return nil
+	}
+	return common.NewDBOrch(orch.ServiceURI, orch.Address.String())
+}
+
+func (dbo *DBOrchestratorPoolCache) GetOrchestrators(numOrchestrators int) ([]*net.OrchestratorInfo, error) {
```
eladmallel, line 48: My 2 cents on file layout: I like keeping all methods (i.e. functions with a receiver) close together, usually having a section below all the methods for the non-receiver functions, so it ends up looking something like:

```

func NewSomething(...) {

}

// Something methods

// functions Something uses (and don't have a receiver)
```

wdyt?

discovery/db_discovery.go
eladmallel: Thinking about how to simplify these concerns, seems to me that the simplest approach would be:

- Update the DB with `TicketParam`s using an asynchronous process: possibly a goroutine for each O we're talking to
- I think in this case it's more valuable to prioritize sql injection protection over disk performance optimization, particularly because these operations are not on the critical path for video processing. This operation seems to be storing external inputs directly into the DB, which poses a real injection risk. I think it's best to use prepared statements in this case, and it's fine to write them one by one as they come in from the async goroutines.

wdyt?

eladmallel: Another question I had - where are we capturing the need to augment this caching process with `TicketParam`s?

discovery/db_discovery.go
angyangie: Definitely on the constant! 

I'm down to try the pattern. Is what you're suggesting adding a context timeout here, like so? `NewDBOrchestratorPoolCache(node *core.LivepeerNode, ctx context.Context)`. In that example, I'm not sure where the `transcodeLoopContext = stubTranscodeLoopContext` gets used. Is `transcodeLoopContext` a global variable that is then overwritten in the tests (in the example)?

discovery/db_discovery.go
angyangie: Good to starting thinking about `GetOrchestratorInfo` retrieval. For when this happens in the background, it seems like it'd be fine to do async updating. However, when we're setting the cache for the first time on [startup](https://github.com/livepeer/go-livepeer/pull/642/files#diff-205b550ef0ac99b368da16cfb883960bR376), if the number of orchs is very large, wouldn't we experience a significant lag? Could we theoretically have two approaches?

Great idea on multiple INSERTs into a single statement @j0sh. I see the limit is 500 for that, so we'd be able to for now insert all at once. What do you mean by "escape all the input" @j0sh? SQL injections is a valid concern

discovery/db_discovery.go
angyangie: I added the check to `EthOchToDBOrch` to be safe. I'm not sure orchs from `orchestrators, err := node.Eth.RegisteredTranscoders()` can come back nil (i.e. if an error is thrown when `tAddr, err := c.GetFirstTranscoderInPool()` or `tAddr, err = c.GetNextTranscoderInPool(tAddr)` is nil). We can leave both checks in to be safe, though technically as things are written now, if we're checking for a nil orch in this for loop above, we don't need to check in `EthOrchToDBOrch`.

common/db_test.go
angyangie: Repetitive code would be the only downside, as you mentioned, but helper functions sound good. I like this new approach!

common/db.go
angyangie: Sounds good to me!

common/db_test.go
angyangie: True!

common/db_test.go
angyangie: I like the improvement - you're right in that it does not need to be a specific value

common/db_test.go
angyangie: Yessss!

discovery/db_discovery.go
angyangie: We don't use this function anywhere else atm, so it can be kept private!

discovery/db_discovery.go
angyangie: Is that a best practice in go? I am fine reorganizing the functions. It does seem like it would improve readability

common/db.go
j0sh: REPLACE alone does a delete-and-insert which might not necessarily always be what we want. For example, we store a creation timestamp here, which would be lost on a replace. It's not critical to keep that right now, but I can see us wanting to retain other types of fields in the future.

This brings up the fact that the query doesn't seem quite well formed, so thanks for bringing this up. To retain upsert-type behavior we probably need a nested select for the insertion time (and any other fields we wish to preserve). See https://stackoverflow.com/a/4330694

The latest SQLIte does seem to have native upsert as a brand-new feature, but I'm not sure what that means as far as distributing an up-to-date sqlite implementation (IIRC SQLite is typically a system lib but it could well be bundled somewhere in the go-sqlite implementation; can dig into this later). https://www.sqlite.org/draft/lang_UPSERT.html

discovery/db_discovery.go
j0sh: We don't necessarily need to pass in a context right now.

Correct that transcodeLoopContext is a global, and the stub is set during the testing function that needs it.

In this case, stubbing the ticker out cleanly seems a bit tricky because because the Ticker is a concrete type rather than an interface. I'm hoping we can explicitly set the tick, eg 

```
// default impl
getTicker = func() {
  return time.Ticker(1 * time.Hour)
}
...

// during testing
st := time.NewTicker(1 * time.Hour)
stubTicker = func() {
  return st
}
getTicker = stubTicker
.... do things that depend on the ticker ...
st.C <- time.Now() // generate a tick
... check things that depend on the ticker ...
```
Let me know if that works.

discovery/db_discovery.go
j0sh: While this would likely be a "low-priority" background task, SQLite writes incur a global lock which blocks all other operations such as reads. We don't want to stall elsewhere in the critical path because of a blocked DB read, so we really should minimize writes as much as possible.

common/db_test.go
j0sh: 😓 I'll go out on a limb here and say that I don't find the "surgical" approach to help readability or understanding the code flow for a few reasons:

* The method names are really long, which takes conscious effort to mentally parse and understand. Even if retaining the surgical approach, I'd really prefer more concise method names, and elaborating on the scenario/expected result as a comment.

* It generates a large amount of boilerplate, which we'd still have with helper functions.

* Separating each test case into its own function, we lose the benefit of code locality. It's harder to quickly scan the tests and reason about the various conditions that are covered (or might not be covered). Here, I was able to quickly see that we were missing nil/empty checks. Also, the progression here can logically be followed: testing empty case, testing one case, testing multiple cases. Splitting that up doesn't improve the readability.

* Related to code locality, one pattern I've used is to check for various conditions following the implementation. This makes it easier to scan the implementation and identify the test case that's covering it. For example, if a function checks and early-exits for conditions A, B, C, we'll have a unit-test checking for A, B, C.  The RPC tests have a bunch of these, eg
https://github.com/livepeer/go-livepeer/blob/master/server/rpc.go#L176 and https://github.com/livepeer/go-livepeer/blob/master/server/rpc_test.go#L147 (admittedly the checks here aren't quite in the proper order; that could be improved.)

angyangie: I'll try a blended approach and see what y'all think! I see what you mean on making each individual test its own function @j0sh  - that would mean a lot of repeated code. It just doesn't make sense for some of our tests. I am going to try push something that adds comments (which helps with readability), and only breaks up the tests that could stand on their own, minimizing code repetition. The method names are definitely long ... I'll take a stab at using the format @eladmallel suggested, and Josh tell me what you think! Perhaps to make them more readable we can take out the results portion of the name after the last "_". For some of the tests I'm working on, doesn't make too much sense. For now, I'll push something up and open the floor for discussion. We can always edit/adjust after.

It might be good to come up with a "standard" way to write tests, that everyone can be happy with. It would be great to have consistency across all of our test files (though we'd have old tests that wouldn't comply to the new test writing style until we make a deliberate attempt to transition them)

common/db.go
angyangie: Update - doing some research on how we can make that work. Getting an error that I'm trying to fix locally... 

```db: Unable to fetch orchestrator sql: Scan error on column index 1, name "ethereumAddr": unsupported Scan, storing driver.Value type <nil> into type *string```

File: discovery/db_discovery.go (hunk 2 above)
Comment (angyangie) [reply discovery/db_discovery.go]: Update - Josh and I are working on testing the ticker! `st.C <- time.Now()` does not work the way we thought it did

File: common/db.go (hunk 9 above)
Comment (j0sh) [reply common/db.go]: These changes did the trick for me

```diff
diff --git a/common/db.go b/common/db.go
index 3d7b6b66..9c0a3c91 100644
--- a/common/db.go
+++ b/common/db.go
@@ -56,8 +56,8 @@ var schema = `
 
 	CREATE TABLE IF NOT EXISTS orchestrators (
 		ethereumAddr STRING PRIMARY KEY,
-		createdAt STRING DEFAULT CURRENT_TIMESTAMP,
-		updatedAt STRING DEFAULT CURRENT_TIMESTAMP,
+		createdAt STRING DEFAULT CURRENT_TIMESTAMP NOT NULL,
+		updatedAt STRING DEFAULT CURRENT_TIMESTAMP NOT NULL,
 		serviceURI STRING
 	);
 
@@ -145,8 +145,7 @@ func InitDB(dbPath string) (*DB, error) {
 	d.selectOrchs = stmt
 
 	// updateOrchestrators statement
-	stmt, err = db.Prepare("INSERT OR REPLACE INTO orchestrators(updatedAt, serviceURI, ethereumAddr) VALUES(datetime(), ?, ?)")
-	// stmt, err = db.Prepare("INSERT OR REPLACE INTO orchestrators(updatedAt, serviceURI, ethereumAddr) VALUES(datetime(), ?, (SELECT ethereumAddr FROM orchestrators WHERE ethereumAddr = ?))")
+	stmt, err = db.Prepare("INSERT OR REPLACE INTO orchestrators(updatedAt, serviceURI, ethereumAddr, createdAt) VALUES(datetime(), ?1, ?2, (SELECT createdAt FROM orchestrators WHERE ethereumAddr = ?2))")
 	if err != nil {
 		glog.Error("Unable to prepare updateOrchestrators stmt ", err)
 		d.Close()
```

j0sh: >  I'm not sure what that means as far as distributing an up-to-date sqlite implementation 

Looks like we're in luck and go-sqlite **does** incorporate [its own copy of sqlite](https://github.com/mattn/go-sqlite3/commit/554bc0fb2e0ee7ec0105f4296b7fa5e01f85aa32), and the repo looks to be fairly up-to-date; we just need to refresh the goclient's vendored copy. This also partially explains why the goclient takes so long to build/link...

Doesn't to be done in this PR though; only if we really want to use the new upsert syntax here.

discovery/db_discovery.go
```diff
@@ -0,0 +1,100 @@
...
+
+	ticker := getTicker()
+	go func(node *core.LivepeerNode) {
+		for {
+			select {
+			case <-ticker.C:
+				orchestrators, err := node.Eth.RegisteredTranscoders()
+				if err != nil {
+					glog.Error("Could not refresh DB list of orchestrators: ", err)
+					return
```
j0sh: Probably best to `continue` here; the issue may have resolved itself by the next tick

discovery/db_discovery.go
```diff
@@ -0,0 +1,100 @@
...
+				}
+				_, dbOrchErr := cacheDBOrchs(node, orchestrators)
+				if dbOrchErr != nil {
+					glog.Error("Could not refresh DB list of orchestrators: cacheDBOrchs err")
+					return
```
j0sh: Ditto on `continue`

discovery/db_discovery.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	for _, orch := range orchs {
+		if orch == nil {
+			continue
+		}
+		dbOrch := ethOrchToDBOrch(orch)
+		if err := node.Database.UpdateOrch(dbOrch); err != nil {
+			glog.Error("Error updating Orchestrator in DB: ", err)
+			return nil, err
```
j0sh: `continue`

common/db.go
angyangie: 🙀 thank you @j0sh! Adding `createdAt` and numbering the ?'s is genius. Why is it better to select on`createdAt` here instead of `ethereumAddr`? 
```sql
(SELECT createdAt FROM orchestrators WHERE ethereumAddr = ?2))")

Comment (j0sh) [reply common/db.go]: Because REPLACE is a delete-and-insert, we need to preserve columns that aren't explicitly inserted. The only column we don't insert here is `createdAt`, so we select that based on the `ethereumAddr`.

If the row doesn't exist (eg, this is an insert) then the select returns NULL. With NULL, the default value (CURRENT_TIMESTAMP) ultimately gets set for the column. This is why the `NOT NULL` constraint had to be set above; it was also added to the updatedAt for completeness.

File: discovery/db_discovery.go (hunk 3 above)
Comment (eladmallel) [reply discovery/db_discovery.go]: Indeed writing to the DB involves locking, which is a certain cost. There is also the cost of writing more complex INSERT code, testing it and maintaining it.

What's your framework for deciding which performance hits (such as DB locks) are worth having suboptimal code, versus hits that aren't worth it?

File: common/db.go (hunk 9 above)
Comment (angyangie) [reply common/db.go]: Makes sense. Thank you!

File: common/db_test.go (hunk 8 above)
Comment (eladmallel) [reply common/db_test.go]: Seems like this is a harder point to achieve consensus around, since I actually find the long test functions with comments along the way harder to parse than the shorter functions I've been trying out.

Since I find consistency very important, I would love for us to land on a specific style all of us commit to.

@j0sh as the leader of this repo I believe you should drive this decision. I realize it is probably going to be long test functions, which will give me the opportunity to improve my code parsing skills :D

File: discovery/db_discovery.go (hunk 14 above)
Comment (eladmallel) [reply discovery/db_discovery.go:48]: I honestly don't know the idiomatic Go file structure. Found [this thread](https://www.reddit.com/r/golang/comments/2xn1ws/idiomatic_go_ordering_of_functions_in_source_files/) with some thoughts, in which no "official" position was shared.

I very much connect with doing our best to make code read coherently from top to bottom, which is why I like placing public functions before private ones, so that when you reach the private function you already know in what context it's used. I also find it very helpful to make sure all of a type's functions are batched together (they can be in separate files if it makes sense, but they shouldn't be scattered across a file with other functions intermingled). 

Perhaps @j0sh should weigh in here and suggest a standard we should stick to?

Comment (j0sh) [reply discovery/db_discovery.go:48]: I don't really have a strong preference about function placement. The proposed approach sounds good.

Personally, I typically start off with keeping function definitions as close as possible to its use to make quick scanning (or even editing) easy, even if this means some interleaving of methods and functions. Of course, there are lots of exceptions to this, and code tends to drift anyway without a conscious effort to keep things organized.

The C programmer in me has always been slightly bemused about defining a function *after* it's used (without a forward declaration) -- it feels a bit like using a variable before it's declared. Lots of languages have this feature though, so why not.

File: discovery/db_discovery.go [hunk 18]
```diff
@@ -0,0 +1,100 @@
...
+
+func NewDBOrchestratorPoolCache(node *core.LivepeerNode) *DBOrchestratorPoolCache {
+	if node.Eth == nil {
+		glog.Error("Could not refresh DB list of orchestrators: LivepeerNode nil")
+		return nil
+	}
+
+	ticker := getTicker()
+	go func(node *core.LivepeerNode) {
+		for {
```
yondonfu: I believe an alternative to nesting the single case select statement within this for loop is to just use a for loop and range over the `ticket.C` channel:

```
for _ := range ticker.C {
    ...
}
```

discovery/db_discovery.go
```diff
@@ -0,0 +1,100 @@
...
+
+	var uris []string
+	for _, orch := range orchs {
+		uri := orch.ServiceURI
+		uris = append(uris, uri)
+	}
+
+	orchPool := NewOrchestratorPool(dbo.node, uris)
+
+	orchInfo, err := orchPool.GetOrchestrators(numOrchestrators)
```
yondonfu: Should `orchInfo` be plural if it is an array?

angyangie: Yes! I also changed `uri` in the `orchestratorPool` struct to `uris`, which makes more sense as well

angyangie:
@j0sh @eladmallel @yondonfu, what do you think of the changes? Any other comments?

j0sh (approved):
LGTM as-is.

Once rebased on top of master, please give it another run-through of manual tests and ensure that e2e transcoding works (well, up to [segment submission](https://github.com/livepeer/go-livepeer/issues/666)), and that the DB is updated periodically.

angyangie:
Thanks @j0sh. Will do!

yondonfu (approved):

PR #643, opened 2018-12-19
Rendition handling updates

j0sh:
Some more refactoring on top of Stream ID related to profile/rendition handling. This is open for discussion: we can refine this further. See https://github.com/livepeer/go-livepeer/pull/625

1. Use the source stream dimensions to construct the VideoProfile rather than selecting the profile from a preset list which is often incorrect and will lead to collisions if there's a transcoding preference matching the profile.

As part of this, the source stream rendition name is set to "source". Most of the work here involves making the playlist accept non-standard VideoProfiles.

2. Change the `hlsStreamID` RTMP query parameter to `manifestID` . This removes the ability to set the source stream rendition but seems safer since there's less potential for collisions with transcoding preferences. This also has the benefit that we only need to check the RTMP query string once, when we generate the StreamID. The RTMP handler itself can use the ManifestID that's extracted.

Note that StreamID parsing rules still apply here: if an ID is in the form `abc/def` then only `abc` will be used as the ManifestID. Once we have RTMP ingest authentication, users can take advantage of this to append auth info as part of the `manifestID` parameter.

Example playlist :
```
$ curl http://localhost:8935/stream/current.m3u8
#EXTM3U
#EXT-X-VERSION:3
#EXT-X-STREAM-INF:PROGRAM-ID=0,BANDWIDTH=4000000,RESOLUTION=854x480
1b1c8eed/source.m3u8
#EXT-X-STREAM-INF:PROGRAM-ID=0,BANDWIDTH=1200000,RESOLUTION=640x360
1b1c8eed/P360p30fps16x9.m3u8
#EXT-X-STREAM-INF:PROGRAM-ID=0,BANDWIDTH=600000,RESOLUTION=320x240
1b1c8eed/P240p30fps4x3.m3u8

...

$ curl http://localhost:8935/stream/1b1c8eed/source.m3u8
#EXTM3U
#EXT-X-VERSION:3
#EXT-X-MEDIA-SEQUENCE:46
#EXT-X-TARGETDURATION:2
#EXTINF:2.000,
/stream/1b1c8eed/source/46.ts
#EXTINF:2.000,
/stream/1b1c8eed/source/47.ts
#EXTINF:2.000,
/stream/1b1c8eed/source/48.ts
#EXTINF:2.000,
/stream/1b1c8eed/source/49.ts
#EXTINF:2.000,
/stream/1b1c8eed/source/50.ts
#EXTINF:2.000,
/stream/1b1c8eed/source/51.ts
```

Unit testing, manual testing

**Checklist:**

### DIFF
--- a/core/playlistmanager.go
+++ b/core/playlistmanager.go
@@ -18,11 +18,11 @@ type PlaylistManager interface {
 	ManifestID() ManifestID
 	// Implicitly creates master and media playlists
 	// Inserts in media playlist given a link to a segment
-	InsertHLSSegment(streamID StreamID, seqNo uint64, uri string, duration float64) error
+	InsertHLSSegment(profile *ffmpeg.VideoProfile, seqNo uint64, uri string, duration float64) error
 
 	GetHLSMasterPlaylist() *m3u8.MasterPlaylist
 
-	GetHLSMediaPlaylist(streamID StreamID) *m3u8.MediaPlaylist
+	GetHLSMediaPlaylist(rendition string) *m3u8.MediaPlaylist
 
 	GetOSSession() drivers.OSSession
 
@@ -34,7 +34,7 @@ type BasicPlaylistManager struct {
 	manifestID     ManifestID
 	// Live playlist used for broadcasting
 	masterPList *m3u8.MasterPlaylist
-	mediaLists  map[StreamID]*m3u8.MediaPlaylist
+	mediaLists  map[string]*m3u8.MediaPlaylist
 	mapSync     *sync.RWMutex
 }
 
@@ -46,7 +46,7 @@ func NewBasicPlaylistManager(manifestID ManifestID,
 		storageSession: storageSession,
 		manifestID:     manifestID,
 		masterPList:    m3u8.NewMasterPlaylist(),
-		mediaLists:     make(map[StreamID]*m3u8.MediaPlaylist),
+		mediaLists:     make(map[string]*m3u8.MediaPlaylist),
 		mapSync:        &sync.RWMutex{},
 	}
 	return bplm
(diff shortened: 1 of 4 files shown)


server/mediaserver.go
```diff
@@ -227,18 +227,12 @@ func gotRTMPStreamHandler(s *LivepeerServer) func(url *url.URL, rtmpStrm stream.
...
-				break
-			}
-		}
-		if vProfile.Name == "" {
-			vProfile = ffmpeg.P720p30fps16x9
-			glog.V(common.SHORT).Infof("Cannot automatically detect the video profile - setting it to %v", vProfile)
+		vProfile := ffmpeg.VideoProfile{
+			Name:       "source",
+			Resolution: resolution,
+			Bitrate:    "4000k", // Fix this
```
angyangie, line 254: Where does 400k come from? Judging by the `\\ Fix this` comment, I assume it might be there as a placeholder, until we build a function to autodetect bitrate?

j0sh: Exactly, yes. It was taken from the "old" 720p default, https://github.com/livepeer/lpms/blob/master/ffmpeg/videoprofile.go#L31 . Mostly inaccurate until we have a better way of returning the estimated bitrate.

angyangie: Awesome! Would we use something like ffprobe for that?

j0sh: The bandwidth indicator in HLS indicates the peak (maximum) bitrate; I suppose we'll need to modify LPMS to return a few metrics which we can use to derive the information (min, max, total, duration, etc). https://tools.ietf.org/html/rfc8216#section-4.3.4.2

angyangie: OooOO cool!

angyangie:
This looks great! Question on https://github.com/livepeer/go-livepeer/pull/618:

It says, "stream key only protects playback (eg, via the segmenter); it doesn't prevent unauthorized streams from being ingested". Why is that? I know that the stream key is randomly generated. If someone can ingest a stream, wouldn't they then know the stream key, and be able to hijack the stream?

angyangie (approved):

core/playlistmanager.go
```diff
@@ -64,55 +64,34 @@ func (mgr *BasicPlaylistManager) GetOSSession() drivers.OSSession {
...
 	duration float64) error {
 
-	if err := mgr.isRightStream(streamID); err != nil {
-		return err
-	}
-	mpl := mgr.getPL(streamID)
+	mpl := mgr.getPL(profile.Name)
 	if mpl == nil {
-		mpl = mgr.createPL(streamID)
+		mpl = mgr.createPL(profile)
```
yondonfu, line 94: Not sure how likely a race condition is, but should `getPL` and `createPL` be combined into `getOrCreatePL` that reads and writes (if a media playlist for the rendition does not already exist) to `mgr.mediaLists` atomically?

core/playlistmanager_test.go
```diff
@@ -39,18 +38,19 @@ func TestGetMasterPlaylist(t *testing.T) {
 }
 
 func TestForWrongStream(t *testing.T) {
+	/* Mismatched streams don't really happen anymore
```
yondonfu, line 41: Should we remove this test function? In it's place should we add a few test case for the new semantics of `InsertHLSSegment`? i.e. making sure that if called with the same rendition twice, the second call returns the first media playlist and does not create a new media playlist that gets appended to the master playlist

yondonfu (changes requested):
> Use the source stream dimensions to construct the VideoProfile rather than selecting the profile from a preset list which is often incorrect and will lead to collisions if there's a transcoding preference matching the profile.

Just for clarification - when you mention collisions between the source profile and the transcoding preference do you mean the following:

1. We currently try to use the source stream resolution to match with the resolution for a preset video profile. Since the source stream resolution could not match the resolution of any of the preset video profiles, the source stream would be assigned the default `P720p30fps16x9` profile.
2. The user selects `P720p30fps16x9` as one of the transcoding preferences
3. When B receives a transcoded result for `P720p30fps16x9` it will insert the HLS segment into the playlist manager. However, since the source stream is using the same rendition name, there already is a media playlist in the playlist manager. Thus, we have a collision.

server/mediaserver_test.go
```diff
@@ -158,7 +158,9 @@ func TestCreateRTMPStreamHandler(t *testing.T) {
 	// (Would be nice to stub out parseManifestID to receive stronger
 	//  transitive assurance via existing parseManifestID tests)
 	testHlsQueryParam := func(inp string) {
```
yondonfu: Should this be renamed to `testManifestIDQueryParam`?

core/playlistmanager.go
j0sh: Sure thing. Any irregularities stemming from a race condition is likely to be transient, but doesn't hurt.

j0sh:
> when you mention collisions between the source profile and the transcoding preference do you mean the following:

Yep, that is correct.

server/mediaserver_test.go
j0sh: Fixed

core/playlistmanager_test.go
j0sh: Sure -- in the interests of keeping this PR focused on failover, I did this in a separate branch along with the getPL / createPL refactor which I will create a PR for shortly. See: https://github.com/livepeer/go-livepeer/compare/ja/failover...ja/playlist

core/playlistmanager.go
j0sh: Fixed this in a separate branch since it's unrelated rendition handling. PR coming shortly. https://github.com/livepeer/go-livepeer/compare/ja/failover...ja/playlist

yondonfu (approved):
LGTM

j0sh:
Mistakenly merged into the streamid branch. Need to approve this one instead: https://github.com/livepeer/go-livepeer/pull/669

PR #646, opened 2018-12-20
Rudimentary failover

j0sh:
Still needs a lot more testing before I'm really comfortable with this, but wanted to get it out here for a quick look-over in case there are concerns about the general approach.

## What does this pull request do?
Implements rudimentary failover:

* Doesn't retry the current segment, but subsequent segments will select a new O.
* If a segment fails, lazily goes through the entire O selection process again rather than caching and using the list that was previously selected. This is actually pretty quick as long as at least one O responds, and is probably a better approach for long-running streams.
* We compare error strings to determine whether we should stop the session and failover. Not pretty, but works. Curious if there's a better approach. The list of error strings are compiled into a single regex for that extra oomph.

### Orchestrator Selector Listener

The major bit that is introduced here is the mechanism for setting the `rtmpConnection.sess` field within `startSessionListener`. A worker loop listens to the `rtmpConnection.needOrch` channel and kicks off the orchestrator-selection process. The orchestrator-selection process is mostly unchanged, but it's been extracted into its own function, `startSession`, with a few modifications for thread safety.

Some of this machinery may seem overwrought but it needs to be thread safe under a number of cases. Happy to walk through some of these if folks have questions, or if anyone has ideas for improvement.

The idea is to ensure exclusion of the `startSession` function: only one invocation can run at a time. If multiple requests for `startSession` come in while one is in-flight, they should be ignored. This prevents a pileup or unnecessary resetting of the session. This is accomplished with a traditional setup of a mutex and a condvar, but I'm also interested in exploring other approaches.

Implementation: https://github.com/livepeer/go-livepeer/commit/bc53ce29a2df4b7c3eb7eecc2fb166d43e7c2c56

An alternative is perhaps to do this with two channels, including a limited-length chan that rejects any attempts to write to it during an in-flight request. The transcode loop uses a similar method to indicate the busy state.

```
cxn.needOrch = make(chan struct{}, 1) // length limit of 1
cxn.getOrch = make(chan struct{}) // unlimited length?
cxn.eof = make(chan struct{})
...
if shouldStopSession(err) {
  select {
  case cxn.needOrch <- struct{}{}:
    cxn.getOrch <- struct{}{}
  case <-cxn.eof
     // rtmp stream ended; exit
     return
  case default:
    // already have an in flight request; exit
    return
  }
}
...
go func() {
  for {
    select {
    case <-cxn.getOrch:
      cxn.sess = nil
      cxn.sess = startSession() // select an O and get a session out of it
      <- cxn.needOrch // drain
    }
  }
}()
```

I'm still not convinced that we don't need locking somewhere. Putting on my old-school hat, I'm also not quite comfortable yet with the "golanginess"  of the approach. Explaining how this works would be extra thinking for folks that aren't familiar with golang concurrency constructs and the specific semantics thereof, such as the behavior of `select` when writing to a full chan. However, most developers should be comfortable with standard synchronization primitives.

## Specific updates

- Split out much of the logic in `gotRTMPStreamHandler` in standalone functions. This is finally straightforward thanks to the latest round of refactors that made the `rtmpConnection` struct available. Primarily `processSegment` and `startSession`.
- Place some code in `server/broadcast.go` to make `mediaserver.go` a bit less heavy. What got moved was pretty arbitrary, but in general, functions that aren't dependent on `LivepeerServer` go into `broadcast.go`

## Testing

1. Run a couple orchestrators in offchain mode. You'll want to set `-offchain` , `-serviceAddr`, `-cliAddr` and probably `-datadir` as well.
2. Run a broadcaster in `-offchain` mode with the above orchestrators specified in `-orchAddr orch1_ip:orch1_port,orch2_ip:orch2_port`
3. Start streaming into the broadcaster
4. Kill the active orchestrator that's working the stream
5. Ensure the second orchestrator picks it up
6. Restart the killed orchestrator
7. GOTO 4

Unit tests forthcoming.

**Checklist:**

### DIFF
--- a/core/orchestrator.go
+++ b/core/orchestrator.go
@@ -2,6 +2,7 @@ package core
 
 import (
 	"context"
+	ogErrors "errors"
 	"fmt"
 	"io/ioutil"
 	"math/big"
@@ -136,6 +137,8 @@ func NewOrchestrator(n *LivepeerNode) *orchestrator {
 
 // LivepeerNode transcode methods
 
+var ErrOrchBusy = ogErrors.New("OrchestratorBusy")
+
 type TranscodeResult struct {
 	Err  error
 	Sig  []byte
@@ -223,7 +226,7 @@ func (n *LivepeerNode) sendToTranscodeLoop(md *SegTranscodingMetadata, seg *stre
 	default:
 		// sending segChan should not block; if it does, the channel is busy
 		glog.Error("Transcoder was busy with a previous segment!")
-		return nil, fmt.Errorf("TranscoderBusy")
+		return nil, ErrOrchBusy
 	}
 	res := <-segChanData.res
 	return res, res.Err
--- a/server/broadcast.go
+++ b/server/broadcast.go
@@ -78,7 +78,7 @@ func processSegment(cxn *rtmpConnection, seg *stream.HLSSegment) {
 	nonce := cxn.nonce
 	rtmpStrm := cxn.stream
 	cpl := cxn.pl
-	mid := rtmpManifestID(rtmpStrm)
+	mid := cxn.mid
 	vProfile := cxn.profile
 
 	cxn.lock.RLock()
--- a/server/broadcast.go
+++ b/server/broadcast.go
@@ -2,18 +2,74 @@ package server
 
 import (
 	"fmt"
+	"math/big"
 	"sync"
 
 	"github.com/golang/glog"
 
+	ethcommon "github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/crypto"
 
 	"github.com/livepeer/go-livepeer/common"
+	"github.com/livepeer/go-livepeer/core"
 	"github.com/livepeer/go-livepeer/drivers"
 	"github.com/livepeer/go-livepeer/monitor"
+	"github.com/livepeer/go-livepeer/pm"
 
 	"github.com/livepeer/lpms/stream"
 )
+
+func selectOrchestrator(n *core.LivepeerNode, cpl core.PlaylistManager) (*BroadcastSession, error) {
+
+	if n.OrchestratorPool == nil {
+		glog.Info("No orchestrators specified; not transcoding")
+		return nil, ErrDiscovery
+	}
+
+	rpcBcast := core.NewBroadcaster(n)
+
+	tinfos, err := n.OrchestratorPool.GetOrchestrators(1)
+	if len(tinfos) <= 0 {
+		glog.Info("No orchestrators found; not transcoding. Error: ", err)
+		return nil, ErrNoOrchs
+	}
+	if err != nil {
+		return nil, err
+	}
+	tinfo := tinfos[0]
+
+	var sessionID string
+
+	if n.Sender != nil {
+		protoParams := tinfo.TicketParams
+		params := pm.TicketParams{
+			Recipient:         ethcommon.BytesToAddress(protoParams.Recipient),
+			FaceValue:         new(big.Int).SetBytes(protoParams.FaceValue),
+			WinProb:           new(big.Int).SetBytes(protoParams.WinProb),
+			RecipientRandHash: ethcommon.BytesToHash(protoParams.RecipientRandHash),
+			Seed:              new(big.Int).SetBytes(protoParams.Seed),
+		}
+
+		sessionID = n.Sender.StartSession(params)
+	}
+
+	// set OSes
+	var orchOS drivers.OSSession
+	if len(tinfo.Storage) > 0 {
+		orchOS = drivers.NewSession(tinfo.Storage[0])
+	}
+
+	return &BroadcastSession{
+		Broadcaster:      rpcBcast,
+		ManifestID:       cpl.ManifestID(),
+		Profiles:         BroadcastJobVideoProfiles,
+		OrchestratorInfo: tinfo,
+		OrchestratorOS:   orchOS,
+		BroadcasterOS:    cpl.GetOSSession(),
+		Sender:           n.Sender,
+		PMSessionID:      sessionID,
+	}, nil
+}
 
 func processSegment(cxn *rtmpConnection, seg *stream.HLSSegment) {
 
--- a/server/broadcast.go
+++ b/server/broadcast.go
@@ -1,0 +1,162 @@
+package server
+
+import (
+	"fmt"
+	"sync"
+
+	"github.com/golang/glog"
+
+	"github.com/ethereum/go-ethereum/crypto"
+
+	"github.com/livepeer/go-livepeer/common"
+	"github.com/livepeer/go-livepeer/drivers"
+	"github.com/livepeer/go-livepeer/monitor"
+
+	"github.com/livepeer/lpms/stream"
+)
+
+func processSegment(cxn *rtmpConnection, seg *stream.HLSSegment) {
+
+	nonce := cxn.nonce
+	rtmpStrm := cxn.stream
+	sess := cxn.sess
+	cpl := cxn.pl
+	mid := rtmpManifestID(rtmpStrm)
+	vProfile := cxn.profile
+
+	if monitor.Enabled {
+		monitor.LogSegmentEmerged(nonce, seg.SeqNo)
+	}
+
+	seg.Name = "" // hijack seg.Name to convey the uploaded URI
+	name := fmt.Sprintf("%s/%d.ts", vProfile.Name, seg.SeqNo)
+	uri, err := cpl.GetOSSession().SaveData(name, seg.Data)
+	if err != nil {
+		glog.Errorf("Error saving segment %d: %v", seg.SeqNo, err)
+		if monitor.Enabled {
+			monitor.LogSegmentUploadFailed(nonce, seg.SeqNo, err.Error())
+		}
+		return
+	}
+	if cpl.GetOSSession().IsExternal() {
+		seg.Name = uri // hijack seg.Name to convey the uploaded URI
+	}
+	err = cpl.InsertHLSSegment(vProfile, seg.SeqNo, uri, seg.Duration)
+	if monitor.Enabled {
+		monitor.LogSourceSegmentAppeared(nonce, seg.SeqNo, string(mid), vProfile.Name)
+		glog.V(6).Infof("Appeared segment %d", seg.SeqNo)
+	}
+	if err != nil {
+		glog.Errorf("Error inserting segment %d: %v", seg.SeqNo, err)
+		if monitor.Enabled {
+			monitor.LogSegmentUploadFailed(nonce, seg.SeqNo, err.Error())
+		}
+	}
+
+	// Return early under a few circumstances:
+	// View-only (non-transcoded) streams or mid-failover
+	if sess == nil {
+		return
+	}
+
+	// Process the rest of the segment asynchronously - transcode
+	go func() {
+		// storage the orchestrator prefers
+		if ios := sess.OrchestratorOS; ios != nil {
+			// XXX handle case when orch expects direct upload
+			uri, err := ios.SaveData(name, seg.Data)
+			if err != nil {
+				glog.Error("Error saving segment to OS ", err)
+				if monitor.Enabled {
+					monitor.LogSegmentUploadFailed(nonce, seg.SeqNo, err.Error())
+				}
+				return
+			}
+			seg.Name = uri // hijack seg.Name to convey the uploaded URI
+		}
+
+		// send segment to the orchestrator
+		glog.V(common.DEBUG).Infof("Submitting segment %d", seg.SeqNo)
+
+		res, err := SubmitSegment(sess, seg, nonce)
+		if err != nil {
+			if shouldStopStream(err) {
+				glog.Warningf("Stopping current stream due to: %v", err)
+				rtmpStrm.Close()
+			}
+			return
+		}
+
+		// download transcoded segments from the transcoder
+		gotErr := false // only send one error msg per segment list
+		errFunc := func(subType, url string, err error) {
+			glog.Errorf("%v error with segment %v: %v (URL: %v)", subType, seg.SeqNo, err, url)
+			if monitor.Enabled && !gotErr {
+				monitor.LogSegmentTranscodeFailed(subType, nonce, seg.SeqNo, err)
+				gotErr = true
+			}
+		}
+
+		segHashes := make([][]byte, len(res.Segments))
+		n := len(res.Segments)
+		segHashLock := &sync.Mutex{}
+		cond := sync.NewCond(segHashLock)
+
+		dlFunc := func(url string, i int) {
+			defer func() {
+				cond.L.Lock()
+				n--
+				if n == 0 {
+					cond.Signal()
+				}
+				cond.L.Unlock()
+			}()
+
+			if bos := sess.BroadcasterOS; bos != nil && !drivers.IsOwnExternal(url) {
+				data, err := drivers.GetSegmentData(url)
+				if err != nil {
+					errFunc("Download", url, err)
+					return
+				}
+				name := fmt.Sprintf("%s/%d.ts", sess.Profiles[i].Name, seg.SeqNo)
+				newUrl, err := bos.SaveData(name, data)
+				if err != nil {
+					errFunc("SaveData", url, err)
+					return
+				}
+				url = newUrl
+
+				hash := crypto.Keccak256(data)
+				segHashLock.Lock()
+				segHashes[i] = hash
+				segHashLock.Unlock()
+			}
+
+			if monitor.Enabled {
+				monitor.LogTranscodedSegmentAppeared(nonce, seg.SeqNo, sess.Profiles[i].Name)
+			}
+			err = cpl.InsertHLSSegment(&sess.Profiles[i], seg.SeqNo, url, seg.Duration)
+			if err != nil {
+				errFunc("Playlist", url, err)
+				return
+			}
+		}
+
+		for i, v := range res.Segments {
+			go dlFunc(v.Url, i)
+		}
+
+		cond.L.Lock()
+		for n != 0 {
+			cond.Wait()
+		}
+		cond.L.Unlock()
+
+		// if !eth.VerifySig(transcoderAddress, crypto.Keccak256(segHashes...), res.Sig) { // need transcoder address here
+		// 	glog.Error("Sig check failed for segment ", seg.SeqNo)
+		// 	return
+		// }
+
+		glog.V(common.DEBUG).Info("Successfully validated segment ", seg.SeqNo)
+	}()
+}
--- a/server/broadcast.go
+++ b/server/broadcast.go
@@ -226,11 +226,12 @@ func processSegment(cxn *rtmpConnection, seg *stream.HLSSegment) {
 	}()
 }
 
+var sessionErrStrings = []string{"dial tcp", "unexpected EOF", core.ErrOrchBusy.Error()}
+
 func generateSessionErrors() *regexp.Regexp {
 	// Given a list [err1, err2, err3] generates a regexp `(err1)|(err2)|(err3)`
-	errStrings := []string{"dial tcp", "unexpected EOF", "TranscoderBusy"}
 	groups := []string{}
-	for _, v := range errStrings {
+	for _, v := range sessionErrStrings {
 		groups = append(groups, fmt.Sprintf("(%v)", v))
 	}
 	return regexp.MustCompile(strings.Join(groups, "|"))
--- a/server/broadcast.go
+++ b/server/broadcast.go
@@ -3,6 +3,8 @@ package server
 import (
 	"fmt"
 	"math/big"
+	"regexp"
+	"strings"
 	"sync"
 
 	"github.com/golang/glog"
@@ -75,10 +77,13 @@ func processSegment(cxn *rtmpConnection, seg *stream.HLSSegment) {
 
 	nonce := cxn.nonce
 	rtmpStrm := cxn.stream
-	sess := cxn.sess
 	cpl := cxn.pl
 	mid := rtmpManifestID(rtmpStrm)
 	vProfile := cxn.profile
+
+	cxn.lock.RLock()
+	sess := cxn.sess
+	cxn.lock.RUnlock()
 
 	if monitor.Enabled {
 		monitor.LogSegmentEmerged(nonce, seg.SeqNo)
@@ -139,6 +144,10 @@ func processSegment(cxn *rtmpConnection, seg *stream.HLSSegment) {
 			if shouldStopStream(err) {
 				glog.Warningf("Stopping current stream due to: %v", err)
 				rtmpStrm.Close()
+				return
+			}
+			if shouldStopSession(err) {
+				cxn.needOrch <- struct{}{}
 			}
 			return
 		}
@@ -215,4 +224,20 @@ func processSegment(cxn *rtmpConnection, seg *stream.HLSSegment) {
 
 		glog.V(common.DEBUG).Info("Successfully validated segment ", seg.SeqNo)
 	}()
+}
+
+func generateSessionErrors() *regexp.Regexp {
+	// Given a list [err1, err2, err3] generates a regexp `(err1)|(err2)|(err3)`
+	errStrings := []string{"dial tcp", "unexpected EOF", "TranscoderBusy"}
+	groups := []string{}
+	for _, v := range errStrings {
+		groups = append(groups, fmt.Sprintf("(%v)", v))
+	}
+	return regexp.MustCompile(strings.Join(groups, "|"))
+}
+
+var sessionErrRegex = generateSessionErrors()
+
+func shouldStopSession(err error) bool {
+	return sessionErrRegex.MatchString(err.Error())
 }
--- a/server/broadcast_test.go
+++ b/server/broadcast_test.go
@@ -1,0 +1,40 @@
+package server
+
+import (
+	"errors"
+	"testing"
+
+	"github.com/livepeer/go-livepeer/core"
+)
+
+func TestStopSessionErrors(t *testing.T) {
+
+	// check error cases
+	errs := []string{
+		"Unable to read response body for segment 4 : unexpected EOF",
+		"Unable to submit segment 5 Post https://127.0.0.1:8936/segment: dial tcp 127.0.0.1:8936: getsockopt: connection refused",
+		core.ErrOrchBusy.Error(),
+	}
+
+	// Sanity check that we're checking each failure case
+	if len(errs) != len(sessionErrStrings) {
+		t.Error("Mismatched error cases for stop session")
+	}
+	for _, v := range errs {
+		if !shouldStopSession(errors.New(v)) {
+			t.Error("Should have stopped session but didn't: ", v)
+		}
+	}
+
+	// check non-error cases
+	errs = []string{
+		"",
+		"not really an error",
+	}
+	for _, v := range errs {
+		if shouldStopSession(errors.New(v)) {
+			t.Error("Should not have stopped session but did: ", v)
+		}
+	}
+
+}
--- a/server/mediaserver.go
+++ b/server/mediaserver.go
@@ -18,8 +18,6 @@ import (
 	"sync"
 	"time"
 
-	ethcommon "github.com/ethereum/go-ethereum/common"
-	"github.com/ethereum/go-ethereum/crypto"
 	"github.com/livepeer/go-livepeer/drivers"
 	"github.com/livepeer/go-livepeer/monitor"
 	"github.com/livepeer/go-livepeer/net"
@@ -63,10 +61,11 @@ var LastHLSStreamID core.StreamID
 var LastManifestID core.ManifestID
 
 type rtmpConnection struct {
-	nonce  uint64
-	stream stream.RTMPVideoStream
-	pl     core.PlaylistManager
-	sess   *BroadcastSession
+	nonce   uint64
+	stream  stream.RTMPVideoStream
+	pl      core.PlaylistManager
+	sess    *BroadcastSession
+	profile *ffmpeg.VideoProfile
 }
 
 type LivepeerServer struct {
@@ -230,6 +229,13 @@ func gotRTMPStreamHandler(s *LivepeerServer) func(url *url.URL, rtmpStrm stream.
 			return ErrStorage
 		}
 		storage := drivers.NodeStorage.NewSession(string(mid))
+		// Build the source video profile from the RTMP stream.
+		resolution := fmt.Sprintf("%vx%v", rtmpStrm.Width(), rtmpStrm.Height())
+		vProfile := ffmpeg.VideoProfile{
+			Name:       "source",
+			Resolution: resolution,
+			Bitrate:    "4000k", // Fix this
+		}
 		s.connectionLock.Lock()
 		_, exists := s.rtmpConnections[mid]
 		if exists {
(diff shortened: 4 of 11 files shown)


core/orchestrator.go
```diff
@@ -2,6 +2,7 @@ package core
 
 import (
 	"context"
+	"errors"
```
eladmallel: @yondonfu recently found an awesome errors package we're planning to use in our PM PRs ([link](https://github.com/pkg/errors)).

The advantage we get from using this errors package is that it captures the stack trace, which adds some very useful data to our logs. It also connects with the author's general error handling paradigm where you don't log errors as they bubble up, but instead treat them as opaque errors and only log them at the top level of handling that error (i.e. where you stop bubbling it up and actually do something with it).

Here's [one of the links](https://www.sohamkamani.com/blog/2018/05/15/golang-error-stack-trace/) Yondon shared, from the author.

@j0sh what do you think about starting to use this approach for errors?

j0sh: That seems pretty nice, I'll start incorporating the package as we go along. We've been doing an ad-hoc version of this by concatenating error strings and returning those, but something more structured could be helpful.

Note, I don't think that this would completely mitigate the type of of "ugly" checking we need to do, eg https://github.com/livepeer/go-livepeer/pull/646/files#diff-cecc27bfe2b6b19181ce443733dfba57R213 

Without actually trying this out (which I will!), the only other thing I can think of is, we don't necessarily need/want a stack trace for every single error. Some types of errors are "expected" and it's nice not to clutter the logs, which are noisy enough. We can probably handle that on a case-by-case basis and get a feel for recurring patterns.

server/mediaserver.go
```diff
@@ -489,12 +310,84 @@ func endRTMPStreamHandler(s *LivepeerServer) func(url *url.URL, rtmpStrm stream.
...
+}
+
+func (s *LivepeerServer) startSessionListener(cxn *rtmpConnection) {
+	mut := sync.Mutex{}
+	cv := sync.NewCond(&mut)
+	pending := false
+	runStartSession := func() {
+		sess := s.startSession(cxn)
+		mut.Lock()
+		defer mut.Unlock()
```
angyangie, line 352: For my own info - when we `defer` here, does the mutex Unlock at the end of runStartSession? Or at the end of startSessionListener? The purpose of this mutex is to make sure only this function accesses cxn at a time? i.e. so that only one function is setting session at a time?

server/mediaserver.go
```diff
@@ -489,12 +310,84 @@ func endRTMPStreamHandler(s *LivepeerServer) func(url *url.URL, rtmpStrm stream.
...
+				defer mut.Unlock()
+				if pending {
+					// Skip if we already have a request inflight
+					return
+				}
+				cxn.sess = nil
+				pending = true
+				go runStartSession()
+				for pending {
+					cv.Wait()
```
angyangie: Hmmm I need some help to really understand CondVars  ... why do we need to do cond.Wait() here? From what I'm reading in the docs (https://golang.org/pkg/sync/#Cond.Wait), it unlocks the Mutex and suspends the execution of the goroutine (which in this case starts on line 369). But, aren't we unlocking the Mutex in `runStartSession` too? Why wouldn't this conflict with the Locking happening on line 358? How about what's happening on line 370 and 371? I see that the mutex is used multiple times, but maybe what I don't understand is why that wouldn't conflict. Is the `defer mut.Unlock()` just a safety precaution, and therefore stated at the end of every function that is using the Mutex just in care the mutex isn't unlocked?

server/mediaserver.go
```diff
@@ -210,6 +183,13 @@ func gotRTMPStreamHandler(s *LivepeerServer) func(url *url.URL, rtmpStrm stream.
 			return ErrStorage
 		}
 		storage := drivers.NodeStorage.NewSession(string(mid))
+		// Build the source video profile from the RTMP stream.
+		resolution := fmt.Sprintf("%vx%v", rtmpStrm.Width(), rtmpStrm.Height())
+		vProfile := ffmpeg.VideoProfile{
+			Name:       "source",
+			Resolution: resolution,
+			Bitrate:    "4000k", // Fix this
```
angyangie: Is this just a placeholder?

server/mediaserver.go
j0sh: The purpose of a condvar is to block execution of the thread (`wait`). The thread is unblocked once the condvar is `signal`'d. So here, we basically `wait` until the `runStartSession` goroutine completes.

The mutex within the condvar is to ensure atomicity of the wait/signal operations; it prevents things like multiple threads calling `signal` at the same time, interleaving of `wait`/`signal` across multiple threads, etc. (The docs say that holding the lock isn't required to call `signal` but I'm sure the lock is still acquired as part of the implementation.)

I'm not sure what you mean by locking conflicts. The locking in `runStartSession` is not strictly necessary for the condvar itself but is required to protect the assignment to the `cxn` and `pending` objects. 

The sequence is:
```
* Goroutine 1: Acquire lock
* Goroutine 1: Wait, release lock (via condvar)
* Goroutine 2: Acquire lock
* Goroutine 2: Signal
* Goroutine 2: Release lock
* Goroutine 1: Acquire lock (via condvar), resume execution
```

server/mediaserver.go
j0sh: Yes, the mutex unlocks at the end of `runStartSession` ; defers are function scoped. Defer actually has a lot of gotchas which is unfortunate because it's a neat construct that minimizes some common errors such as missing a clean-up step (such as unlocking a mutex) in a error-return path.

See https://blog.learngoprogramming.com/gotchas-of-defer-in-go-1-8d070894cb01

server/mediaserver.go
j0sh: Yeah, it was taken from the "old" 720p default, https://github.com/livepeer/lpms/blob/master/ffmpeg/videoprofile.go#L31 . Mostly inaccurate until we have a better way of returning the estimated bitrate.

server/mediaserver.go
angyangie: What a great article! Thank you Josh

server/mediaserver.go
angyangie: Ah I see what you mean! What I meant by "locking conflicts" had to do with mutex being unlocked in `runStartSession` then, unlocked again by line 371. I just found a really good tutorial on condVars and mutexes that makes this all make more sense! : https://computing.llnl.gov/tutorials/pthreads/#ConditionVariables.

server/mediaserver.go
```diff
@@ -283,7 +289,8 @@ func gotRTMPStreamHandler(s *LivepeerServer) func(url *url.URL, rtmpStrm stream.
 		glog.V(common.SHORT).Infof("\n\nhlsStrmID: %v\n\n", hlsStrmID)
 
 		//Create Transcode Job Onchain
-		go s.startSession(cxn)
+		go s.startSessionListener(cxn)
+		cxn.needOrch <- struct{}{}
```
angyangie, line 234: What ensures that the select statement in `s.startSessionListener` is actually running/listening when line 293 gets called? Trying to again figure out concurrency in go!

server/mediaserver.go
```diff
@@ -285,143 +259,13 @@ func gotRTMPStreamHandler(s *LivepeerServer) func(url *url.URL, rtmpStrm stream.
...
-
-						// if !eth.VerifySig(transcoderAddress, crypto.Keccak256(segHashes...), res.Sig) { // need transcoder address here
-						// 	glog.Error("Sig check failed for segment ", seg.SeqNo)
-						// 	return
-						// }
-
-						glog.V(common.DEBUG).Info("Successfully validated segment ", seg.SeqNo)
-					}()
-				}
+				processSegment(cxn, seg)
```
angyangie, line 210: How do we know `cxn.sess` will be set here? I see that we're setting it on line 292, through `go s.startSessionListener(cxn)`, but will that for sure execute before this line?

server/mediaserver.go
j0sh: > What ensures that the select statement in s.startSessionListener is actually running/listening when line 293 gets called

Nothing actually. If the RTMP stream stops, `eof` is called and the select statement exits, we'll still be feeding `cxn.needOrch` for each segment that is in-flight at the time of `eof`.

This seems OK. We could `close` the needOrch chanel instead of using an `eof` chan, but that panics (crashes) if we ever inadvertently write to a closed channel. It seems safer to use an `eof` chan and accept the small hit of writing to a channel that will never be drained. The GC ensures that we eventually reclaim any empty `struct{}`s that may have been buffered in the meantime.

server/mediaserver.go
j0sh: We don't; `processSegment` returns early if `cxn.sess` is nil. There are situations where this is normal, eg view-only streams or while we are in the middle of the O re-selection process for failover.  https://github.com/livepeer/go-livepeer/pull/646/files#diff-cecc27bfe2b6b19181ce443733dfba57R93

Where we set `sess = cxn.sess` in `processSegment` should probably be wrapped in a mutex though.

angyangie: That makes sense! `go s.startSessionListener(cxn)` is running in a for loop, and line 292 could very well run after `processSegment(cxn, seg)`

server/mediaserver.go
angyangie: Interesting! Didn't know about the crash if we write to a closed channel. Agreed with your thinking here. I'm wondering though if `s.startSessionListener` will be listening the first time `cxn.needOrch <- struct{}{}` gets called on line 293? i.e. do we know with certainty that the`for{}` statement on line 365 would get executed before line 293 every time?

j0sh: Ah I see. If the scheduling is such that `cxn.needOrch` is written to before `startSessionListener` completes the `for/select` setup, then the `needOrch` chan will have its contents buffered until the `for/select` drains it (or `needOrch` is GC'd at stream exit).

server/broadcast.go
```diff
@@ -0,0 +1,214 @@
...
+		// 	return
+		// }
+
+		glog.V(common.DEBUG).Info("Successfully validated segment ", seg.SeqNo)
+	}()
+}
+
+func generateSessionErrors() *regexp.Regexp {
+	// Given a list [err1, err2, err3] generates a regexp `(err1)|(err2)|(err3)`
+	errStrings := []string{"dial tcp", "unexpected EOF", "OrchestratorBusy"}
```
eladmallel: I'm curious to learn:

- How did you compile this list?
- How do we know it's exhaustive?

j0sh: Manually tested the following conditions:
* Orchestrator down at initial connection / segment submission time yields an error with the substring `dial tcp`
* Orchestrator down mid-transcode yields an error with the substring `unexpected EOF`
* Orchestrator with a busy transcode loop yields an error with `OrchestratorBusy`

This list is by no means exhaustive, it's just what I've been able to capture in practice. If you can think of more scenarios that'd merit an immediate failover, or see additional cases during testing, this list can certainly be expanded.

server/broadcast.go
```diff
@@ -0,0 +1,214 @@
...
+			glog.Errorf("%v error with segment %v: %v (URL: %v)", subType, seg.SeqNo, err, url)
+			if monitor.Enabled && !gotErr {
+				monitor.LogSegmentTranscodeFailed(subType, nonce, seg.SeqNo, err)
+				gotErr = true
+			}
+		}
+
+		segHashes := make([][]byte, len(res.Segments))
+		n := len(res.Segments)
+		SegHashLock := &sync.Mutex{}
```
yondonfu: `SegHashLock` -> `segHashLock`?

server/broadcast.go
```diff
@@ -0,0 +1,214 @@
...
+		}
+		return
+	}
+	if cpl.GetOSSession().IsExternal() {
+		seg.Name = uri // hijack seg.Name to convey the uploaded URI
+	}
+	err = cpl.InsertHLSSegment(vProfile, seg.SeqNo, uri, seg.Duration)
+	if monitor.Enabled {
+		monitor.LogSourceSegmentAppeared(nonce, seg.SeqNo, string(mid), vProfile.Name)
+		glog.V(6).Infof("Appeared segment %d", seg.SeqNo)
```
yondonfu, line 108: Should this log statement be outside of the `if monitor.Enabled...` conditional?

server/broadcast_test.go
```diff
@@ -0,0 +1,21 @@
...
+import (
+	"errors"
+	"testing"
+
+	"github.com/livepeer/go-livepeer/core"
+)
+
+func TestStopSessionErrors(t *testing.T) {
+	errs := []string{
+		"Unable to read response body for segment 4 : unexpected EOF",
```
yondonfu, line 14: Should we add an error message that should not stop a session and test that `shouldStopSession` returns false as expected?

server/broadcast.go
```diff
@@ -0,0 +1,214 @@
...
+		}
+
+		// send segment to the orchestrator
+		glog.V(common.DEBUG).Infof("Submitting segment %d", seg.SeqNo)
+
+		res, err := SubmitSegment(sess, seg, nonce)
+		if err != nil {
+			if shouldStopStream(err) {
+				glog.Warningf("Stopping current stream due to: %v", err)
+				rtmpStrm.Close()
```
yondonfu, line 146: What is the expected behavior when an RTMP stream is closed? If we proceed to also stop the session and find a new orchestrator when is a new RTMP stream created?

server/broadcast.go
```diff
@@ -0,0 +1,214 @@
...
+			}
+			if shouldStopSession(err) {
+				cxn.needOrch <- struct{}{}
+			}
+			return
+		}
+
+		// download transcoded segments from the transcoder
+		gotErr := false // only send one error msg per segment list
```
yondonfu, line 156: If different errors occur while downloading each segment in a list, wouldn't we want to know about all of them?

server/broadcast.go
```diff
@@ -0,0 +1,214 @@
...
+				return
+			}
+		}
+
+		for i, v := range res.Segments {
+			go dlFunc(v.Url, i)
+		}
+
+		cond.L.Lock()
+		for n != 0 {
```
yondonfu, line 215: Curious what your thoughts are on [sync.WaitGroup](https://golang.org/pkg/sync/#WaitGroup)?

```
var wg sync.WaitGroup

    wg.Add(1)
    go func(i int, url string) {
        defer wg.Done()
        dlFunc(url, i)
    }(i, v.Url)
}

wg.Wait()
```

server/mediaserver.go
```diff
@@ -489,12 +310,84 @@ func endRTMPStreamHandler(s *LivepeerServer) func(url *url.URL, rtmpStrm stream.
...
 
+func (s *LivepeerServer) startSession(cxn *rtmpConnection) *BroadcastSession {
+
+	mid := rtmpManifestID(cxn.stream)
+	cpl := cxn.pl
+	var sess *BroadcastSession
+
+	// Connect to the orchestrator. If it fails, retry for as long
+	// as the RTMP stream is alive; maybe the orchestrator hasn't
+	// received the block containing the job yet
```
yondonfu: I think we can update the last part of this comment since orchestrators no longer wait for blocks containing jobs

server/mediaserver.go
```diff
@@ -489,12 +310,84 @@ func endRTMPStreamHandler(s *LivepeerServer) func(url *url.URL, rtmpStrm stream.
...
+	runStartSession := func() {
+		sess := s.startSession(cxn)
+		mut.Lock()
+		defer mut.Unlock()
+		cxn.sess = sess
+		pending = false
+		cv.Signal()
+	}
+	glog.V(common.DEBUG).Info("Starting broadcast listener for ", cxn.pl.ManifestID())
+	for {
```
yondonfu, line 357: I think this could be a buffered (size = 1) channel based alternative to using a cond var:

```
sem := make(chan struct{}, 1)

for {
    select {
        case <-cxn.needOrch:
            select {
                // If this send succeeds then no other request will be accepted until we unblock sem
                case sem <- struct{}{}: 
                    cxn.sess = s.startSession()
                    <-sem // Unblock sem
                default:
                    return // The above case is blocking so we already have an inflight request
            }
            ...
    }
}
```

Thoughts? Also, do we generally need a mutex for `cxn.sess` since it is read in a few other places?

server/mediaserver.go
angyangie: Ooooo I see, that makes sense. +1 for the creators of Golang!

j0sh: Heh. I'd argue that the opaqueness of most of channel semantics is a drawback here -- someone made a choice for a lot of these behaviors; they have to be "memorized" and can't be derived or reasoned about as with traditional concurrency primitives. That might be justified if chans were implemented as a library atop the base, but it's somewhat questionable as a general-purpose, first-class language construct.

server/broadcast.go
j0sh: Sure, the reason wait groups are not used here is that I didn't know about them when this was implemented. The refactored code was just copy-pasted from the existing mediaserver code; can fix up the style issues separately.

server/mediaserver.go
j0sh: I like this, will give it a test. Thanks!

> Also, do we generally need a mutex for cxn.sess since it is read in a few other places?

Yes, we generally should use a mutex to protect accesses to `ctx.sess`. Needs to be added to the cxn struct though, see https://github.com/livepeer/go-livepeer/pull/646#discussion_r244882689

server/broadcast.go
angyangie: Cool! Didn't know about it either

server/broadcast.go
eladmallel: Perhaps it's interesting to test the case of O taking a very long time to respond to the HTTP request?

j0sh: The error on timeouts will look approximately like this:

``` rpc.go:459] Unable to submit segment 12 Post https://127.0.0.1:8936/segment: net/http: request canceled (Client.Timeout exceeded while awaiting headers)```

From our [earlier discussion](https://github.com/livepeer/go-livepeer/issues/636) it seemed like we settled on using a counter for timeouts. We can probably match on `Client.Timeout` . The current implementation doesn't incorporate a counter yet though; it only handles hard failovers.

Comment (eladmallel) [reply server/broadcast.go]: Agreed that timeouts fall under the issue you linked to. Do I understand correctly that in the current implementation we give full grace to timeouts? i.e. just keep sending segments regardless of how many timeouts occur?

File: server/broadcast.go (hunk 8 above)
Comment (j0sh) [reply server/broadcast.go]: Fixed

File: server/broadcast_test.go (hunk 10 above)
Comment (j0sh) [reply server/broadcast_test.go:14]: Fixed

File: server/broadcast.go (hunk 11 above)
Comment (j0sh) [reply server/broadcast.go:146]: A `BroadcastSession` would still be assigned to `rtmpConnection.sess`, but should eventually be GC'd along with the `rtmpConnection`, so it's most likely harmless -- we don't have anything to manually clean up.

This is probably not the most robust solution as it could lead to a resource leak if we ever need to do a manual clean-up around `BroadcastSession` in the future.

File: server/broadcast.go (hunk 12 above)
Comment (j0sh) [reply server/broadcast.go:156]: IIRC it was done this way to minimize noise in metrics-reporting. Most errors are likely to be connectivity related, so if one fails, others are likely to too.

File: server/mediaserver.go (hunk 14 above)
Comment (j0sh) [reply server/mediaserver.go]: Fixed

File: server/mediaserver.go (hunk 15 above)
Comment (j0sh) [reply server/mediaserver.go:357]: Protective mutex around `cxn.sess` has been added, and tests added to this area of code. Things look clean, but will refactor around the suggested solution shortly.

File: server/broadcast.go (hunk 7 above)
Comment (j0sh) [reply server/broadcast.go]: Yes, that's correct.

File: server/mediaserver.go (hunk 15 above)
Comment (j0sh) [reply server/mediaserver.go:357]: Fixed in https://github.com/livepeer/go-livepeer/pull/646/commits/6940c486d82e6d8b0444fa5b2a7461066161ca97

File: server/broadcast.go (hunk 11 above)
Comment (j0sh) [reply server/broadcast.go:146]: Improved this behavior; now we check whether the RTMP connection is still active prior to assigning `cxn.sess = sess`. If inactive, `sess` is set to nil. This should be safe since all accesses to the `rtmpConnections` map need to be protected by a mutex. See https://github.com/livepeer/go-livepeer/pull/646/commits/43d97d78b92131c41b7f66b1e5433bdb58d34d70#diff-236217e7f473dfbde9134c6a391f3408R336

Comment (j0sh) [reply server/broadcast.go:146]: Circling back to the original question -- added a `return` after the `rtmpStrm.Close()` so we aren't doing unnecessary work after the close. https://github.com/livepeer/go-livepeer/pull/646/commits/43d97d78b92131c41b7f66b1e5433bdb58d34d70#diff-cecc27bfe2b6b19181ce443733dfba57R147

**Comment by j0sh** [comment]
Removed WIP label

**Comment by eladmallel** [comment]
@j0sh is this PR basically ready for review and then merge?

**Comment by j0sh** [comment]
Yes, it is ready, thanks!

File: server/mediaserver.go [hunk 16]
```diff
@@ -481,12 +287,88 @@ func endRTMPStreamHandler(s *LivepeerServer) func(url *url.URL, rtmpStrm stream.
...
+	expb.MaxElapsedTime = 0
+	backoff.Retry(broadcastFunc, expb)
+	return sess
+}
+
+func (s *LivepeerServer) startSessionListener(cxn *rtmpConnection) {
+	mut := cxn.lock
+	sem := make(chan struct{}, 1)
+	runStartSession := func() {
+		mut.Lock()
```
yondonfu, line 338: Why do we lock/unlock `mut` here on L332-334 and then again on L341-342 instead of locking `mut` once for the duration of this function? Is it just to minimize the amount of time that`cxn.sess` is locked?

server/mediaserver.go
```diff
@@ -481,12 +287,88 @@ func endRTMPStreamHandler(s *LivepeerServer) func(url *url.URL, rtmpStrm stream.
...
+		cxn.sess = nil
+		mut.Unlock()
+		sess := s.startSession(cxn)
+		s.connectionLock.RLock()
+		defer s.connectionLock.RUnlock()
+		if _, active := s.rtmpConnections[cxn.pl.ManifestID()]; !active {
```
darkdarkdragon: Why here we are using `cxn.pl.ManifestID()` and not `cxn.mid`?

server/mediaserver_test.go
```diff
@@ -410,6 +427,189 @@ func TestGetHLSMasterPlaylistHandler(t *testing.T) {
...
+	s.LivepeerNode.OrchestratorPool = nil
+	assert.Nil(s.startSession(cxn))
+	assert.Equal(1, sd.getOrchCalls)
+	s.LivepeerNode.OrchestratorPool = sd
+
+	// ensure a single try works
+	s.connectionLock.Lock()
+	s.rtmpConnections[mid] = cxn
+	s.connectionLock.Unlock()
+	sd.infos = []*net.OrchestratorInfo{&net.OrchestratorInfo{}}
```
yondonfu: Isn't `sd.infos` already set to this on L443?

server/mediaserver.go
```diff
@@ -481,12 +287,88 @@ func endRTMPStreamHandler(s *LivepeerServer) func(url *url.URL, rtmpStrm stream.
...
+		cxn.eof <- struct{}{}
 		delete(s.rtmpConnections, mid)
 
 		return nil
 	}
 }
 
+func (s *LivepeerServer) startSession(cxn *rtmpConnection) *BroadcastSession {
+
+	mid := rtmpManifestID(cxn.stream)
```
darkdarkdragon: we already have `ManifestID` in connection - `cxn.mid` - why we're parsing it again from RtmpStream?

server/broadcast.go
```diff
@@ -0,0 +1,244 @@
...
+	err = cpl.InsertHLSSegment(vProfile, seg.SeqNo, uri, seg.Duration)
+	if monitor.Enabled {
+		monitor.LogSourceSegmentAppeared(nonce, seg.SeqNo, string(mid), vProfile.Name)
+		glog.V(6).Infof("Appeared segment %d", seg.SeqNo)
+	}
+	if err != nil {
+		glog.Errorf("Error inserting segment %d: %v", seg.SeqNo, err)
+		if monitor.Enabled {
+			monitor.LogSegmentUploadFailed(nonce, seg.SeqNo, err.Error())
+		}
```
yondonfu, line 114: Just for clarification - since we continue here even if there is an error inserting the HLS segment into the playlist manager that means we are ok with HLS segments potentially being missing for the source stream playlist yeah?

server/broadcast.go
```diff
@@ -0,0 +1,244 @@
...
+	}
+	err = cpl.InsertHLSSegment(vProfile, seg.SeqNo, uri, seg.Duration)
+	if monitor.Enabled {
+		monitor.LogSourceSegmentAppeared(nonce, seg.SeqNo, string(mid), vProfile.Name)
+		glog.V(6).Infof("Appeared segment %d", seg.SeqNo)
+	}
+	if err != nil {
+		glog.Errorf("Error inserting segment %d: %v", seg.SeqNo, err)
+		if monitor.Enabled {
+			monitor.LogSegmentUploadFailed(nonce, seg.SeqNo, err.Error())
```
yondonfu, line 113: This monitor event seems like it might a bit off - the error we encounter here is for inserting an HLS segment into the playlist manager and not due to failure to upload a segment (which happens [here](https://github.com/livepeer/go-livepeer/pull/646/files#diff-cecc27bfe2b6b19181ce443733dfba57R98))

server/broadcast.go
```diff
@@ -0,0 +1,244 @@
...
+		PMSessionID:      sessionID,
+	}, nil
+}
+
+func processSegment(cxn *rtmpConnection, seg *stream.HLSSegment) {
+
+	nonce := cxn.nonce
+	rtmpStrm := cxn.stream
+	cpl := cxn.pl
+	mid := rtmpManifestID(rtmpStrm)
```
darkdarkdragon: why not using `cxn.mid`?

server/mediaserver.go
```diff
@@ -481,12 +287,88 @@ func endRTMPStreamHandler(s *LivepeerServer) func(url *url.URL, rtmpStrm stream.
...
+		_, active := s.rtmpConnections[mid]
+		if active {
+			return err
+		}
+		return nil // stop if inactive
+	}
+	expb := backoff.NewExponentialBackOff()
+	expb.MaxInterval = BroadcastRetry
+	expb.MaxElapsedTime = 0
+	backoff.Retry(broadcastFunc, expb)
```
darkdarkdragon, line 329: What is the point of using backoff here? If it doesn't succeed on the first run, results of any next run will be just lost. Probably it is just left from old code?

darkdarkdragon (commented):
> If a segment fails, lazily goes through the entire O selection process again rather than caching and using the list that was previously selected. 

But right now it still uses Os list from database, so it is uses cached data?

darkdarkdragon:
@j0sh Correct me if I wrong, but in current implementation, if O properly responds to ping request, but respond with error on `SubmitSegment`, same O will be chosen over and over, and no transcoding will happen?

server/mediaserver.go
j0sh: `cxn.mid` was introduced in a more recent PR, https://github.com/livepeer/go-livepeer/pull/680 , I can fix that up separately within that PR. (Same goes for the rest of the feedback on the cxn.mid)

server/mediaserver.go
j0sh: Not sure that I understand the question. If acquiring a session doesn't succeed, it retries for as long as the RTMP client is connected.

server/broadcast.go
j0sh: Probably could be improved, but most of this bit of code was refactored (moved around) without functional changes.

server/mediaserver.go
darkdarkdragon: Yes, it retries it, and puts result in local var on line 306, and that result doesn't go anywhere - `startSession` is synchronous function that should return result (session) immediately.

server/broadcast.go
j0sh: Correct that we continue even if there is an error inserting into the playlist, which could result in some playlists missing segments and gaps during playback.

We can certainly discuss this separately if the behavior should be modified, but as-is just copied from existing code.

server/mediaserver.go
j0sh: Just want to be sure we're looking at the same place -- the `sess` is declared outside of the `broadcastFunc` function that actually calls `selectOrchestrator` (so `sess` is non-local in this case).

https://github.com/livepeer/go-livepeer/pull/646/files/1a5f92d1c5902044bc33933cc9bf7cb8eeb07b72#diff-236217e7f473dfbde9134c6a391f3408R307

This is one of the small functional changes that was made during the refactoring; rather than assigning the `sess` to the `cxn` here as was done earlier, we do the assignment during `startSessionListener` in order to keep the state mutation somewhat self-contained to startSessionListener. Exact commit where this was done: https://github.com/livepeer/go-livepeer/pull/646/commits/bed60552e05f416c193a8d1f6d653a198cdb44b3

darkdarkdragon: Before the mentioned commit (bed6055)  `sess` variable was global to `gotRTMPStreamHandler` function, but that doesn't matter, it wasn't used in that function. It was working because previously it was
```
				if active {
					cxn.sess = sess
					return err
				}
```

but now it is
```
				if active {
					return err
				}
```

j0sh: Note that `sess` is still returned from the top level function as long as the inner `broadcastFunc` returns non-nil. The `sess` is still assigned and returned.

See the related changes here: https://github.com/livepeer/go-livepeer/pull/646/commits/372afc25194ec13f78bc27ea0db6d44d69b46bf5

Here's some pseudocode of how the function is supposed to work, let me know if you have concerns about the overall approach or whether the current implementation doesn't match the pseudocode:

```
fun GetSession() *session {
  var sess *session // this is local to this top-level "GetSession"
  var tryGetttingSession = func () error {
    var err error
    sess, err = Session() // sess is non-local to `tryGettingSession`
     return err // will retry if `err` is non-nil
  }
  retryUntilSuccess(tryGettingSession)
  return sess
}
```

This approach is definitely quirky. It exposes two things I'm not a fan of:
* The odd semantics of golang `:=` declaration syntax and how it does not compose well esp around error handling. We have to be careful here to not (re-)declare the `sess` and `err` vars; this has caused bugs for us in the past.
* The somewhat strange usage for the retry module in using closures and mutation in order to extract any useful state from the function under retry. Until golang gets proper parametric polymorphism (generics), this is the best we have.

j0sh:
> > If a segment fails, lazily goes through the entire O selection process again rather than caching and using the list that was previously selected.

> But right now it still uses Os list from database, so it is uses cached data?

Yeah, that refers to "caching" in the sense that it doesn't remember the original list that was used for O selection, and doesn't exclude the previously selected orchestrator when attempting a failover. So it is certainly possible that the same orchestrator is selected twice for failover.

> if O properly responds to ping request, but respond with error on SubmitSegment, same O will be chosen over and over, and no transcoding will happen?

That's why this is rudimentary :) To be precise though, the failover only happens under conditions where the O is down hard from B's perspective (see the discussion with Elad on this https://github.com/livepeer/go-livepeer/pull/646#discussion_r244981475 ).

If the error is a transient network issue then it's likely OK to select the same orchestrator again (any issue would have to resolve itself quite quickly). Otherwise, it's likely that `GetOrchestrator` would also fail, rather than flip-flopping indefinitely.

server/mediaserver_test.go
j0sh: Fixed

server/mediaserver.go
j0sh: Yes, because the following `s.startSession` could take some time (it ends up invoking GetOrchestrator RPC, potentially multiple times). We don't want to block the `cxn` in other places that acquire the lock, such as during `processSegment` which acquires the lock for each segment, so every 2 seconds.

Added some comments clarifying this area of the code, including the connectionLock

server/mediaserver.go
j0sh: Fixed most of the `cxn.mid` occurrences in https://github.com/livepeer/go-livepeer/pull/680/commits/0873879a7246327635e1cee55ab16a76aca46a20

server/mediaserver.go
darkdarkdragon: Sorry, my mistake - I, after reading documentation on `backoff.Retry`, somehow thought that it operates asynchronously.

darkdarkdragon:
> That's why this is rudimentary :) 

OK, this will do for now :) 
But one more scenario we should check - is when O is standalone and T goes down.

j0sh:
> But one more scenario we should check - is when O is standalone and T goes down.

That's a good question. What will happen is that an "ErrTranscoderUnavailable" is returned immediately and no segment is transcoded. That error isn't considered a hard fail right now and the broadcaster will continue to send segments to the orchestrator. It's not ideal but it also won't flap on failover, so not it's not a new issue exposed by this PR.

https://github.com/livepeer/go-livepeer/blob/a2c3f5156cbb97dd6816a035629b96c1504326be/core/orchestrator.go#L262-L270

In order to make the "ErrTranscoderUnavailable" a hard fail and to avoid flapping on failovers, we'll likely need to modify `GetOrchestrator` to return an error under the scenario where there are no transcoders available in standalone mode. There's a bit of additional plumbing required for this (the transcoders aren't currently accessible via rpc), so we can wait until later for this; my sense is that we'll introduce a "transcoder management layer" with multi-T which will likely lead to some refactoring around this area of the code.

yondonfu (approved):
LGTM

PR #648, opened 2018-12-20
PM: TicketParams struct

eladmallel:
Adding this struct in a standalone commit since we have two parallel
branches of pm recipient and sender in progress, and both need this
struct.

### DIFF
--- a/pm/ticket.go
+++ b/pm/ticket.go
@@ -14,6 +14,16 @@ const (
 	bytes32Size = 32
 )
 
+type TicketParams struct {
+	FaceValue *big.Int
+
+	WinProb *big.Int
+
+	RecipientRandHash ethcommon.Hash
+
+	Seed *big.Int
+}
+
 // Ticket is lottery ticket payment in a probabilistic micropayment protocol
 // The expected value of the ticket constitutes the payment and can be
 // calculated using the ticket's face value and winning probability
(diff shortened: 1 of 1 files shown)


pm/ticket.go
```diff
@@ -14,6 +14,16 @@ const (
 	bytes32Size = 32
 )
 
+type TicketParams struct {
```
yondonfu, line 19: I've been trying to add comments for all exported functions/data (i.e. capitalized variables) so that godoc can then generate nice documentation for everything (see: https://golang.org/doc/effective_go.html#commentary) so I thought we can make this an ongoing practice - thoughts?

yondonfu (changes requested):

yondonfu (approved):
LGTM

pm/ticket.go
j0sh: In favor of this practice :+1:  Knowing golang, is there a linter or something we can use to tell us if something is missing godoc documentation?

yondonfu: [golint](https://github.com/golang/lint) seems to do the trick. I use VSCode which runs golint automatically in the editor. 

On the topic of linters, there seems to be a fairly large amount of static analysis/linting tools for Go mentioned in the README for [gometalinter](https://github.com/alecthomas/gometalinter) which looks like a tool that can run a bunch of them simultaneously. We're probably fine with just adding golint to the CI process initially (though we'll have to fix a bunch of its warnings around things like missing godoc), but perhaps we can explore using gometalinter later on.

Issue #621, opened 2018-12-11
pm: Store winning tickets on-disk with corresponding recipientRand value


yondonfu:
**Is your feature request related to a problem? Please describe.**

While in MVS, O will submit an on-chain transaction to redeem a winning ticket immediately after receiving the ticket, it is possible that O will require the ticket data even after submitting the transaction

- The original transaction might not have been successfully broadcasted due to issues with O's Ethereum connection
- O might crash prior to the original transaction being successfully broadcasted

Beyond MVS, O might not submit an on-chain transaction to redeem a winning ticket immediately after receiving a ticket and instead choose to hold on to winning tickets to be redeemed later on.

Thus, O would benefit from being able to store winning tickets on-disk.

**Describe the solution you'd like**

O has a `winningTickets` DB table. Each table row contains all the data necessary to redeem a winning ticket on-chain: the fields of a ticket, B's signature and `recipientRand` (which should correspond to `ticket.recipientRandHash`).

When O receives a winning ticket, it stores the winning ticket data in its DB. After a winning ticket redemption transaction confirms on-chain, O can clear the relevant entry for the winning ticket in the DB.

**Describe alternatives you've considered**

- For MVS, we could skip DB persistence since O will always redeem winning tickets immediately, but this would leave O vulnerable to loss of funds in cases where the transaction does not confirm on-chain for some reason

PR #649, opened 2018-12-21
PM recipient

### LINKED ISSUE #630: PM Recipient

yondonfu:

Adds recipient specific code to the `pm` package.

- `pm/recipient.go`: Recipient specific state management (i.e. HMAC construction for generating `recipientRand`, replay prevention of invalid `recipientRand` and `senderNonce` values, etc.) 
- `pm/ticketstore.go`: The TicketStore interface allows a recipient to persist winning tickets and fetch them when they need to be redeemed on-chain. The implementation will rely on the node's existing DB and will be done as a part of #621
- Introduce the [pkg/errors](https://github.com/pkg/errors) package for error handling with stack trace capture support. The hope is that using this package will assist with debugging errors and keeping logging at the top level of the program when possible. For more reading on this see: https://dave.cheney.net/2016/06/12/stack-traces-and-the-errors-package and http://dave.cheney.net/paste/gocon-spring-2016.pdf

A few design notes for `recipient.go`:
- `recipient.invalidRands` is a [sync.Map](https://golang.org/pkg/sync/#Map) because as an append only cache for detecting replays of invalidated `recipientRand` values it fits into one of the two recommend use cases for `sync.Map`. Using a dedicated data structure from the the Go standard library for this allows us to avoid writing additional unit tests for the correctness of manually implemented locks.
- `recipient.senderNonces` is a plain Go map with read/write access protected by a `sync.RWMutex`. In the happy case (i.e. no one is attacking you), `recipient.senderNonces` fits into one of the two recommend use cases for `sync.Map` in that it usually supports concurrent operations on a disjoint set of keys. However, for every update to a value in `recipient.senderNonces`, we want to check that the new nonce > the current highest nonce. Otherwise, we want to return an error because the new nonce is invalid. `sync.Map` does not support this type of atomic operation - we would have to first load the current highest nonce and check it and then store the new nonce. These two operations would not be atomic thereby still leaving the possibility of a race condition. Instead of using `sync.Map`, we just use `sync.RWMutex` to implement manually locking - we lock, atomically load the current highest nonce, check it, store the new nonce and then we unlock. An alternative to manually locking is to use an open source [concurrent-map](https://github.com/orcaman/concurrent-map) package which allows a caller to pass in a callback to include additional logic when storing a value in the map. The limited code for manual locking seems easier to understand at the moment, but definitely up for other thoughts (with the downside of having to test the manual locking).

- Unit tests for each of the new data structures with stubbed out dependencies
- Used `go test -race` to find race conditions for manually implemented lock code

- Fixes #630 

**Checklist:**

Remaining TODOs:
- [x] Write a unit test for the manual locking used to protect `recipient.senderNonces`

### DIFF
--- a/pm/recipient.go
+++ b/pm/recipient.go
@@ -29,8 +29,9 @@ type Recipient interface {
 	// ReceiveTicket validates and processes a received ticket
 	ReceiveTicket(ticket *Ticket, sig []byte, seed *big.Int) (won bool, err error)
 
-	// RedeemWinningTicket redeems a winning ticket with the broker
-	RedeemWinningTicket(ticketID ethcommon.Hash) error
+	// RedeemWinningTicket redeems all winning tickets with the broker
+	// for a session ID
+	RedeemWinningTickets(sessionID string) error
 
 	// TicketParams returns the recipient's currently accepted ticket parameters
 	// for a provided sender ETH adddress
@@ -97,7 +98,7 @@ func (r *recipient) ReceiveTicket(ticket *Ticket, sig []byte, seed *big.Int) (bo
 	}
 
 	if r.val.IsWinningTicket(ticket, sig, recipientRand) {
-		if err := r.store.Store(ticket, sig, recipientRand); err != nil {
+		if err := r.store.Store(ticket.RecipientRandHash.Hex(), ticket, sig, recipientRand); err != nil {
 			return true, err
 		}
 
@@ -107,13 +108,43 @@ func (r *recipient) ReceiveTicket(ticket *Ticket, sig []byte, seed *big.Int) (bo
 	return false, nil
 }
 
-// RedeemWinningTicket redeems a winning ticket with the broker
-func (r *recipient) RedeemWinningTicket(ticketID ethcommon.Hash) error {
-	ticket, sig, recipientRand, err := r.store.Load(ticketID)
+// RedeemWinningTicket redeems all winning tickets with the broker
+// for a session ID
+func (r *recipient) RedeemWinningTickets(sessionID string) error {
+	tickets, sigs, recipientRands, err := r.store.Load(sessionID)
 	if err != nil {
 		return err
 	}
 
+	for i := 0; i < len(tickets); i++ {
+		if err := r.redeemWinningTicket(tickets[i], sigs[i], recipientRands[i]); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// TicketParams returns the recipient's currently accepted ticket parameters
+func (r *recipient) TicketParams(sender ethcommon.Address) (*TicketParams, error) {
+	randBytes := make([]byte, 32)
+	if _, err := rand.Read(randBytes); err != nil {
+		return nil, err
+	}
+
+	seed := new(big.Int).SetBytes(randBytes)
+	recipientRand := r.rand(seed, sender)
+	recipientRandHash := crypto.Keccak256Hash(ethcommon.LeftPadBytes(recipientRand.Bytes(), uint256Size))
+
+	return &TicketParams{
+		FaceValue:         r.faceValue,
+		WinProb:           r.winProb,
+		RecipientRandHash: recipientRandHash,
+		Seed:              seed,
+	}, nil
+}
+
+func (r *recipient) redeemWinningTicket(ticket *Ticket, sig []byte, recipientRand *big.Int) error {
 	deposit, err := r.broker.GetDeposit(ticket.Sender)
 	if err != nil {
 		return err
@@ -142,25 +173,6 @@ func (r *recipient) RedeemWinningTicket(ticketID ethcommon.Hash) error {
 	}
 
 	return nil
-}
-
-// TicketParams returns the recipient's currently accepted ticket parameters
-func (r *recipient) TicketParams(sender ethcommon.Address) (*TicketParams, error) {
-	randBytes := make([]byte, 32)
-	if _, err := rand.Read(randBytes); err != nil {
-		return nil, err
-	}
-
-	seed := new(big.Int).SetBytes(randBytes)
-	recipientRand := r.rand(seed, sender)
-	recipientRandHash := crypto.Keccak256Hash(ethcommon.LeftPadBytes(recipientRand.Bytes(), uint256Size))
-
-	return &TicketParams{
-		FaceValue:         r.faceValue,
-		WinProb:           r.winProb,
-		RecipientRandHash: recipientRandHash,
-		Seed:              seed,
-	}, nil
 }
 
 func (r *recipient) rand(seed *big.Int, sender ethcommon.Address) *big.Int {
--- a/pm/recipient.go
+++ b/pm/recipient.go
@@ -1,0 +1,207 @@
+package pm
+
+import (
+	"crypto/hmac"
+	"crypto/rand"
+	"crypto/sha256"
+	"math/big"
+	"sync"
+
+	ethcommon "github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/pkg/errors"
+)
+
+var (
+	errInvalidRecipientRandFromSeed = errors.New("invalid recipientRand generated from seed")
+	errInvalidRecipientRandRevealed = errors.New("invalid already revealed recipientRand")
+
+	errInvalidTicketFaceValue   = errors.New("invalid ticket faceValue")
+	errInvalidTicketWinProb     = errors.New("invalid ticket winProb")
+	errInvalidTicketSenderNonce = errors.New("invalid ticket senderNonce")
+
+	errZeroDepositAndPenaltyEscrow = errors.New("sender has zero deposit and penalty escrow")
+)
+
+// Recipient is an interface which describes an object capable
+// of receiving tickets
+type Recipient interface {
+	// ReceiveTicket validates and processes a received ticket
+	ReceiveTicket(ticket *Ticket, sig []byte, seed *big.Int) (won bool, err error)
+
+	// RedeemWinningTicket redeems a winning ticket with the broker
+	RedeemWinningTicket(ticketID ethcommon.Hash) error
+
+	// TicketParams returns the recipient's currently accepted ticket parameters
+	// for a provided sender ETH adddress
+	TicketParams(sender ethcommon.Address) (*TicketParams, error)
+}
+
+// recipient is an implementation of the Recipient interface that
+// receives tickets and redeems winning tickets
+type recipient struct {
+	val    Validator
+	broker Broker
+	store  TicketStore
+
+	secret [32]byte
+
+	invalidRands sync.Map
+
+	senderNonces     map[string]uint64
+	senderNoncesLock sync.Mutex
+
+	faceValue *big.Int
+	winProb   *big.Int
+}
+
+// NewRecipient creates an instance of a recipient
+func NewRecipient(broker Broker, val Validator, store TicketStore, secret [32]byte, faceValue *big.Int, winProb *big.Int) Recipient {
+	return &recipient{
+		broker:       broker,
+		val:          val,
+		store:        store,
+		secret:       secret,
+		faceValue:    faceValue,
+		senderNonces: make(map[string]uint64),
+		winProb:      winProb,
+	}
+}
+
+// ReceiveTicket validates and processes a received ticket
+func (r *recipient) ReceiveTicket(ticket *Ticket, sig []byte, seed *big.Int) (bool, error) {
+	recipientRand := r.rand(seed, ticket.Sender)
+
+	if crypto.Keccak256Hash(ethcommon.LeftPadBytes(recipientRand.Bytes(), uint256Size)) != ticket.RecipientRandHash {
+		return false, errInvalidRecipientRandFromSeed
+	}
+
+	if ticket.FaceValue.Cmp(r.faceValue) != 0 {
+		return false, errInvalidTicketFaceValue
+	}
+
+	if ticket.WinProb.Cmp(r.winProb) != 0 {
+		return false, errInvalidTicketWinProb
+	}
+
+	if err := r.val.ValidateTicket(ticket, sig, recipientRand); err != nil {
+		return false, err
+	}
+
+	if !r.validRand(recipientRand) {
+		return false, errInvalidRecipientRandRevealed
+	}
+
+	if err := r.updateSenderNonce(recipientRand, ticket.SenderNonce); err != nil {
+		return false, err
+	}
+
+	if r.val.IsWinningTicket(ticket, sig, recipientRand) {
+		if err := r.store.Store(ticket, sig, recipientRand); err != nil {
+			return true, err
+		}
+
+		return true, nil
+	}
+
+	return false, nil
+}
+
+// RedeemWinningTicket redeems a winning ticket with the broker
+func (r *recipient) RedeemWinningTicket(ticketID ethcommon.Hash) error {
+	ticket, sig, recipientRand, err := r.store.Load(ticketID)
+	if err != nil {
+		return err
+	}
+
+	deposit, err := r.broker.GetDeposit(ticket.Sender)
+	if err != nil {
+		return err
+	}
+
+	penaltyEscrow, err := r.broker.GetPenaltyEscrow(ticket.Sender)
+	if err != nil {
+		return err
+	}
+
+	if deposit.Cmp(big.NewInt(0)) == 0 && penaltyEscrow.Cmp(big.NewInt(0)) == 0 {
+		return errZeroDepositAndPenaltyEscrow
+	}
+
+	// We are about to reveal recipientRand on-chain so invalidate it locally
+	if err := r.updateInvalidRands(recipientRand); err != nil {
+		return err
+	}
+
+	// After we invalidate recipientRand we can clear the memory used to track
+	// its latest senderNonce
+	r.clearSenderNonce(recipientRand)
+
+	if err := r.broker.RedeemWinningTicket(ticket, sig, recipientRand); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// TicketParams returns the recipient's currently accepted ticket parameters
+func (r *recipient) TicketParams(sender ethcommon.Address) (*TicketParams, error) {
+	randBytes := make([]byte, 32)
+	if _, err := rand.Read(randBytes); err != nil {
+		return nil, err
+	}
+
+	seed := new(big.Int).SetBytes(randBytes)
+	recipientRand := r.rand(seed, sender)
+	recipientRandHash := crypto.Keccak256Hash(ethcommon.LeftPadBytes(recipientRand.Bytes(), uint256Size))
+
+	return &TicketParams{
+		FaceValue:         r.faceValue,
+		WinProb:           r.winProb,
+		RecipientRandHash: recipientRandHash,
+		Seed:              seed,
+	}, nil
+}
+
+func (r *recipient) rand(seed *big.Int, sender ethcommon.Address) *big.Int {
+	h := hmac.New(sha256.New, r.secret[:])
+	h.Write(append(seed.Bytes(), sender.Bytes()...))
+
+	return new(big.Int).SetBytes(h.Sum(nil))
+}
+
+func (r *recipient) validRand(rand *big.Int) bool {
+	_, ok := r.invalidRands.Load(rand.String())
+	return !ok
+}
+
+func (r *recipient) updateInvalidRands(rand *big.Int) error {
+	_, loaded := r.invalidRands.LoadOrStore(rand.String(), true)
+	if loaded {
+		return errInvalidRecipientRandRevealed
+	}
+
+	return nil
+}
+
+func (r *recipient) updateSenderNonce(rand *big.Int, senderNonce uint64) error {
+	r.senderNoncesLock.Lock()
+	defer r.senderNoncesLock.Unlock()
+
+	randStr := rand.String()
+	nonce, ok := r.senderNonces[randStr]
+	if ok && senderNonce <= nonce {
+		return errInvalidTicketSenderNonce
+	}
+
+	r.senderNonces[randStr] = senderNonce
+
+	return nil
+}
+
+func (r *recipient) clearSenderNonce(rand *big.Int) {
+	r.senderNoncesLock.Lock()
+	defer r.senderNoncesLock.Unlock()
+
+	delete(r.senderNonces, rand.String())
+}
--- a/pm/recipient.go
+++ b/pm/recipient.go
@@ -16,7 +16,7 @@ import (
 // of receiving tickets
 type Recipient interface {
 	// ReceiveTicket validates and processes a received ticket
-	ReceiveTicket(ticket *Ticket, sig []byte, seed *big.Int) (won bool, err error)
+	ReceiveTicket(ticket *Ticket, sig []byte, seed *big.Int) (sessionID string, won bool, err error)
 
 	// RedeemWinningTicket redeems all winning tickets with the broker
 	// for a session ID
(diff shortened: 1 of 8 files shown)


pm/recipient_test.go
```diff
@@ -0,0 +1,436 @@
...
+	"encoding/hex"
+	"math/big"
+	"testing"
+
+	ethcommon "github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/crypto"
+)
+
+func TestReceiveTicket(t *testing.T) {
+	sender := ethcommon.HexToAddress("A69cdA26600c155cF2c150964Bdb5371ac3f606F")
```
eladmallel: Is this specific address important, or would any random address work here?

If a random address is just as good, I'd like to suggest using the approach I took [here](https://github.com/livepeer/go-livepeer/pull/651/files#diff-0800fac1c4a865f1fd74c8aeebcd8e60R231), basically generating a random address every time the test runs.

The advantage is a more readable test - because I don't have to wonder about the specific significance of this address.

wdyt?

pm/recipient_test.go
```diff
@@ -0,0 +1,436 @@
...
+		Recipient:         ethcommon.Address{},
+		Sender:            sender,
+		FaceValue:         faceValue,
+		WinProb:           winProb,
+		SenderNonce:       0,
+		RecipientRandHash: params.RecipientRandHash,
+	}
+
+	// Using invalid seed
+	_, err = r.ReceiveTicket(ticket, sig, big.NewInt(12312))
```
eladmallel: How do we know for sure that this seed value is invalid?

I would feel more comfortable if the invalid seed was derived from the valid seed, e.g. `invalidSeed := params.Seed + 1234`.

wdyt?

pm/recipient_test.go
```diff
@@ -0,0 +1,436 @@
...
+	if err == nil {
+		t.Error("expected invalid recipientRand from seed error")
+	}
+	if err != nil && err != errInvalidRecipientRandFromSeed {
+		t.Errorf("expected invalid recipientRand from seed error, got %v", err)
```
eladmallel, line 97: I was thinking that it might be more helpful for real-life debugging to include variable values in the error strings we generate. In this case, I would suggest we format the error return value as: `invalid recipientRand generated from seed. Sender: <senderAddress>, Seed: <seed value>`, perhaps with a few more data points.

Wdyt of this approach versus always returning the same error string?

pm/recipient_test.go
```diff
@@ -0,0 +1,436 @@
...
+	}
+
+	// Test invalid recipientRand from seed (invalid sender)
+	ticket = &Ticket{
+		Recipient:         ethcommon.Address{},
+		Sender:            ethcommon.Address{}, // Using invalid sender
```
eladmallel: This is confusing - can you help me understand how adding an invalid sender tests for the use case of an invalid recipientRand?

I thought testing for an invalid recipientRand would look more like providing an invalid value for `RecipientRandHash`. Perhaps the invalid sender scenario should have a test of its own with its own error?

pm/recipient_test.go
```diff
@@ -0,0 +1,436 @@
...
+	sender := ethcommon.HexToAddress("A69cdA26600c155cF2c150964Bdb5371ac3f606F")
+	b := newStubBroker()
+	v := &stubValidator{}
+	ts := newStubTicketStore()
+	secret := [32]byte{3}
+	faceValue := big.NewInt(100)
+	winProb := big.NewInt(100)
+	sig := []byte("foo")
+
+	r := NewRecipient(b, v, ts, secret, faceValue, winProb)
```
eladmallel: Perhaps extract this line and the lines above it into a helper function that creates a recipient with some default setup? ([example here](https://github.com/livepeer/go-livepeer/pull/651/files#diff-0800fac1c4a865f1fd74c8aeebcd8e60R199)).

pm/recipient_test.go
```diff
@@ -0,0 +1,436 @@
...
+		SenderNonce:       0,
+		RecipientRandHash: params1.RecipientRandHash,
+	}
+
+	won, err := r.ReceiveTicket(ticket, sig, params1.Seed)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if !won {
+		t.Fail()
```
eladmallel: I'm not sure why this should be a different kind of failure compared to `err != nil`. I think both cases should be the same, probably using `t.Fatal` to stop execution, since in this specific test we're assuming that `ReceiveTicket` should work as expected, i.e. `won` should be true and `err` should be nil and anything different means we can't reliably test `RedeemWinningTicket`.

wdyt?

pm/recipient_test.go
```diff
@@ -0,0 +1,436 @@
...
+	}
+
+	err = r.RedeemWinningTicket(ticket.Hash())
+	if err == nil {
+		t.Error("expected zero deposit and penalty escrow error")
+	}
+	if err != nil && err != errZeroDepositAndPenaltyEscrow {
```
eladmallel: Thinking about the state of zero deposit and PE (penalty escrow) - perhaps zeros is only a specific case of a more general idea? Thoughts that seem to make sense to me:

- Deposit needs to be above some threshold, such that it's worthwhile for me to even pay for gas. This sounds like a fairly low number, but perhaps higher than zero.
- PE should also be above some threshold, but it's less of a client concern thanks to the minimal PE we will configure in the Broker contract.

I don't think we it's urgent to solve this concern right now, but perhaps it's good to capture the idea of refactoring this zero comparison to a configurable-threshold comparison.

wdyt?

pm/recipient_test.go
```diff
@@ -0,0 +1,436 @@
...
+	// already invalidated
+	err = r.RedeemWinningTicket(ticket.Hash())
+	if err == nil {
+		t.Error("expected invalid recipientRand revealed error")
+	}
+	if err != nil && err != errInvalidRecipientRandRevealed {
+		t.Errorf("expected invalid recipientRand revealed error, got %v", err)
+	}
+
+	// TODO: Unit test for internal clearing of senderNonce memory?
```
eladmallel: I agree with adding tests for memory cleanup. I also think we should test for when `r.broker.RedeemWinningTicket()` return an error.

In case of error from `broker.RedeemWinningTicket` I wonder if it makes sense that we still invalidate the `recipientRand` in question, versus perhaps only invalidating upon successful txn submission?

Another question is around retries. Should we expand what we're doing upon failure to redeem a winning ticket?

pm/validator.go
```diff
@@ -29,15 +29,13 @@ type Validator interface {
 // validator is an implementation of the Validator interface
 type validator struct {
 	addr        ethcommon.Address
-	broker      Broker
 	sigVerifier SigVerifier
 }
 
 // NewValidator returns an instance of a validator
-func NewValidator(addr ethcommon.Address, broker Broker, sigVerifier SigVerifier) Validator {
+func NewValidator(addr ethcommon.Address, sigVerifier SigVerifier) Validator {
```
eladmallel, line 36: 👍

pm/recipient.go
```diff
@@ -0,0 +1,207 @@
...
+	winProb   *big.Int
+}
+
+// NewRecipient creates an instance of a recipient
+func NewRecipient(broker Broker, val Validator, store TicketStore, secret [32]byte, faceValue *big.Int, winProb *big.Int) Recipient {
+	return &recipient{
+		broker:       broker,
+		val:          val,
+		store:        store,
+		secret:       secret,
```
eladmallel, line 70: Perhaps we can generate `secret` inside of `Recipient`? Wondering why it's better to provide it externally?

eladmallel:
@yondonfu overall looks good, just had a few comments/questions. On a more general point, I'd like to open up the possibility of writing shorter test functions, that test a smaller scope. I personally find the long test functions harder to read since I need to remember a bunch of setup stuff from high up in the beginning, etc. Would love to get your take on this matter, especially after you review my recent attempt to demonstrate what shorter functions might look like ([here](https://github.com/livepeer/go-livepeer/pull/651/files#diff-0800fac1c4a865f1fd74c8aeebcd8e60R16)).

Thanks!

pm/recipient_test.go
yondonfu: Yeah I agree that the condition for redemption based on a sender's deposit and penalty escrow can be more sophisticated in the future. I think the transaction cost based threshold for a deposit check is a good start! I can add a TODO comment and we can track this in an issue as well.

pm/recipient_test.go
yondonfu: Yeah in addition to maintaining consistent with method usage, I think it makes sense to add a log message with `Fatal()` for this case so there is at least a clear reason why the test is terminating execution instead of proceeding.

pm/recipient.go
yondonfu: I though generating `secret` outside of `Recipient` would make testing easier especially in the future if we add a method for updating `secret`. Plus, the caller can choose how it wants to initialize `secret`.

pm/recipient_test.go
yondonfu: I think the `recipientRand` should be invalidated as soon as the transaction is submitted (but the recipient should not wait until the transaction is confirmed since `recipientRand` is already visible in the mempool).

I imagined `broker.RedeemWinningTicket` to handle retries when the original transaction has an insufficient gas price (likely with a threshold at which it will stop trying to bump the gas price). It can kickoff a goroutine for transaction retries and return an error indicating whether the original transaction was successfully submitted or not. Then, in `Recipient`, we would check the error returned and invalidate `recipientRand` if the error is nil.

pm/recipient_test.go
yondonfu: Sure, we can format the error string. The downside with formatting the error string is that we will no longer be able to do simple constant error checks while testing and instead we'll have to type out error strings or do string matching.

pm/recipient_test.go
yondonfu: While the seed is valid since it was received from the recipient, the sender address included in the ticket is invalid such that a invalid `recipientRand` is generated since `HMAC(recipient.Secret, seed | T.Sender) != T.RecipientRandHash`.

I could add an additional test where `T.RecipientRandHash` is some nonsense value as opposed to the value included in the `TicketParams` returned by a recipient.

yondonfu:
I'm on board with shorter test functions - I'll follow up with some updates

pm/recipient_test.go
eladmallel: We're having the same conversation on both our PRs :) Let's resolve this dilemma in one place and apply our conclusions in both places. Where would you like our main thread about const error strings vs dynamic error string params to take place?

pm/recipient_test.go
eladmallel: Ok, that makes sense to me now! I think the additional test you suggested could be nice!

pm/recipient_test.go
eladmallel: Sounds reasonable! IIUC the current implementation assumes that the txn would be submitted in the same thread as `recipient. RedeemWinningTicket`, right? - otherwise the `err` return value has no meaning. 

Sounds like you're suggesting to change this paradigm to actually submitting the txn in a separate goroutine?

pm/recipient.go
eladmallel: What concerns me is that in theory this design does not prevent the user of `Recipient` from reusing the same secret across multiple sessions. That is of course very dangerous because it exposes O to replays.

How do you suggest we think about this potential risk?

pm/recipient_test.go
yondonfu: Yeah the original assumption was that `broker.RedeemWinningTicket` would run in the same thread and would block until the transaction was mined (and it would be ok to block here because `recipient.RedeemWinningTicket` would likely be run in its own thread). The error returned in this paradigm would inform the caller if the transaction was successfully mined or if the transaction retry process was unsuccessful (i.e. time out or exceeded gas price threshold while replacing transactions).

In the new paradigm that I am suggesting, `broker.RedeemWinningTicket` would:

1. Submit the transaction
2. Start a separate goroutine to wait for the transaction to be mined (and retry if necessary)
3. Return

If there is a problem with step 1, `broker.RedeemWinningTicket` would return an error immediately - since the transaction never made it to the network we wouldn't invalidate `recipientRand` here. If there is no problem with step 1, then nil is returned in step 3 so we invalidate `recipientRand` because we assume that the transaction made it to the network.

eladmallel: SGTM!

pm/recipient.go
yondonfu: So this is a question of whether the user should have to understand how to securely create a secret for this package. In the interest of making the package as mistake proof as possible I think creating the secret within `Recipient` makes sense. I guess I can always give `Recipient` a SecretGenerator interface that can facilitate stubbing in tests...

eladmallel: Right. Then the risk moves to the `SecretGenerator`, where a pm package user might implement their own generator that also brings about potential security risks :)

I believe having at least one constructor function that does not take a `SecretGenerator` argument, but rather uses some default implementation is a good mitigation.

In addition, perhaps on the second constructor that does take in a generator, we should add very clear documentation that supplying your own generator adds risk.

wdyt?

yondonfu: I like having multiple constructor functions. We can omit `SecretGenerator` altogether and just have one function that accepts a user defined secret and one function that generates the secret internally. The documentation would strongly suggest using the latter. The former would just be handy for testing.

pm/recipient_test.go
yondonfu: Switched to dynamic error strings as of c0506a9

pm/recipient_test.go
yondonfu: Added the additional test case: https://github.com/livepeer/go-livepeer/pull/649/files#diff-bd67cf9d99c17eb1ffc2c3d08ea455b0R99

pm/recipient_test.go
yondonfu: Added a TODO here: https://github.com/livepeer/go-livepeer/pull/649/files#diff-bd67cf9d99c17eb1ffc2c3d08ea455b0R99

Also tracking in #653

pm/recipient.go
yondonfu: Updated with two constructors:

[NewRecipientWithSecret](https://github.com/livepeer/go-livepeer/pull/649/files#diff-3a62c7d0189200e6024095a62a267eaaR65) accepts a user provided secret.

[NewRecipient](https://github.com/livepeer/go-livepeer/pull/649/files#diff-3a62c7d0189200e6024095a62a267eaaR50) automatically generates a random secret and then uses NewRecipientWithSecret.

pm/recipient.go
```diff
@@ -0,0 +1,223 @@
...
+	ethcommon "github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/pkg/errors"
+)
+
+// Recipient is an interface which describes an object capable
+// of receiving tickets
+type Recipient interface {
+	// ReceiveTicket validates and processes a received ticket
+	ReceiveTicket(ticket *Ticket, sig []byte, seed *big.Int) (won bool, err error)
```
eladmallel: I think we want to add a return value here - the sessionID the user of `Recipient` needs to keep track of in order to later redeem tickets. Otherwise how will O know what value to provide as the input to `RedeemWinningTickets`?

pm/recipient.go
```diff
@@ -0,0 +1,223 @@
...
+func NewRecipient(broker Broker, val Validator, store TicketStore, faceValue *big.Int, winProb *big.Int) (Recipient, error) {
+	randBytes := make([]byte, 32)
+	if _, err := rand.Read(randBytes); err != nil {
+		return nil, err
+	}
+
+	var secret [32]byte
+	copy(secret[:], randBytes[:32])
+
+	return NewRecipientWithSecret(broker, val, store, secret, faceValue, winProb), nil
```
eladmallel, line 59: Great constructor reuse 👍

pm/recipient.go
```diff
@@ -0,0 +1,223 @@
...
+	if !r.validRand(recipientRand) {
+		return false, errors.Errorf("invalid already revealed recipientRand %v", recipientRand)
+	}
+
+	if err := r.updateSenderNonce(recipientRand, ticket.SenderNonce); err != nil {
+		return false, err
+	}
+
+	if r.val.IsWinningTicket(ticket, sig, recipientRand) {
+		if err := r.store.Store(ticket.RecipientRandHash.Hex(), ticket, sig, recipientRand); err != nil {
```
eladmallel: Was thinking a bit about how we generate sessionID from `ticket.RecipientRandHash.Hex()` - that's something we now do both in `Sender` and `Recipient` and actually do in a bunch of places.

What do you think about extracting this to a helper function we would use across the board?

I think it makes the code a bit more readable and definitely will be helpful if we ever change the way we generate sessionIDs.

wdyt?

pm/recipient.go
```diff
@@ -0,0 +1,223 @@
...
+// for a session ID
+func (r *recipient) RedeemWinningTickets(sessionID string) error {
+	tickets, sigs, recipientRands, err := r.store.Load(sessionID)
+	if err != nil {
+		return err
+	}
+
+	for i := 0; i < len(tickets); i++ {
+		if err := r.redeemWinningTicket(tickets[i], sigs[i], recipientRands[i]); err != nil {
+			return err
```
eladmallel, line 128: Wondering - perhaps it's better to not break the loop upon a single ticket failure, but instead allow it to run through all tickets in case some of them do succeed?

pm/recipient.go
```diff
@@ -0,0 +1,223 @@
...
+	// After we invalidate recipientRand we can clear the memory used to track
+	// its latest senderNonce
+	r.clearSenderNonce(recipientRand)
+
+	return nil
+}
+
+func (r *recipient) rand(seed *big.Int, sender ethcommon.Address) *big.Int {
+	h := hmac.New(sha256.New, r.secret[:])
+	h.Write(append(seed.Bytes(), sender.Bytes()...))
```
eladmallel, line 191: Question of ignorance :) what does the `...` at the end of the `append` call do?

pm/recipient.go
yondonfu: What do you imagine this helper to look like?

At the moment, the sender generates a sessionID in `StartSession()` using the provided `TicketParams` struct:

```
sessionID := ticketParams.RecipientRandHash.Hex()
```

Meanwhile, the recipient generates a sessionID in `ReceiveTicket()` using the provided `Ticket` struct:

```
sessionID := ticket.RecipientRandHash.Hex()
```

Since the sender and recipient use different data structures in order to access `RecipientRandHash`, the common input for both of them would just be `RecipientRandHash` and at that point the sessionID helper would just return the hex string representation of the input as the output.

pm/recipient_test.go
```diff
@@ -0,0 +1,851 @@
...
+	return new(big.Int).SetBytes(h.Sum(nil))
+}
+
+func TestReceiveTicket_InvalidRecipientRand_InvalidSeed(t *testing.T) {
+	sender, b, v, ts, faceValue, winProb, sig := newRecipientFixture(t)
+
+	r, err := NewRecipient(b, v, ts, faceValue, winProb)
+	if err != nil {
+		t.Fatal(err)
+	}
```
eladmallel: These few lines of creating a new recipient repeat in several tests - perhaps it's nice to extract it into a helper function like `newRecipientOrFatal(t)` and have it just return the recipient?

eladmallel: Actually, you might be able to just add this logic to `newRecipientFixture`

pm/recipient_test.go
```diff
@@ -0,0 +1,851 @@
...
+	sender, b, v, ts, faceValue, winProb, sig := newRecipientFixture(t)
+	secret := [32]byte{3}
+
+	r := NewRecipientWithSecret(b, v, ts, secret, faceValue, winProb)
+
+	// Config stub validator with valid winning tickets
+	v.SetIsValidTicket(true)
+	v.SetIsWinningTicket(true)
+	// Config stub ticket store to fail store
+	ts.storeShouldFail = true
```
eladmallel, line 276: IMO if you move this configuration to the line before calling the "code under test", the test becomes more readable, because this line is really the most different setup line compared to previous tests.

pm/recipient_test.go
```diff
@@ -0,0 +1,851 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Test valid winning ticket
+	newSenderNonce := uint64(3)
+	ticket := newTicket(sender, params, newSenderNonce)
+
+	_, err = r.ReceiveTicket(ticket, sig, params.Seed)
+	if err == nil {
```
eladmallel, line 279: Is there perhaps something in the error content you can assert here to make sure we're indeed failing because of the TicketStore setup and not something else?

pm/recipient.go
yondonfu: I initially considered not breaking the loop upon a single ticket failure, but ended up holding off because it would complicate error handling. The caller would likely want to know the errors for each ticket failure so this function would need to either aggregate the errors (perhaps using something like [go-multierror](https://github.com/hashicorp/go-multierror)) or log individual errors as they occur.

If we think that it reasonable to continue trying to redeem even after a single failure we can make that change, but my initial thought is that if one fails there is a high likelihood that the rest will fail as well because:

- If we assume that each ticket redemption is for the same sender then the recipient will be checking the same deposit and penalty escrow
- If we assume that each ticket redemption will return immediately if there is an error submitting the transaction then there is likely a deeper ETH connection issue that could cause the subsequent transactions to fail during submission as well

We could handle retries for error relating to a bad ETH connection at a layer below this so that if the recipient gets an error during a ETH related action it knows that something is definitely wrong.

pm/recipient_test.go
```diff
@@ -0,0 +1,851 @@
...
+	r, err := NewRecipient(b, v, ts, faceValue, winProb)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Config stub validator with invalid non-winning tickets
+	v.SetIsValidTicket(true)
+	v.SetIsWinningTicket(true)
+	// Config stub broker with zero deposit and non-zero penalty escrow
+	b.SetDeposit(sender, big.NewInt(0))
```
eladmallel: Why set deposit to zero in this test case? - for max readability I'd expect any test that isn't testing for different deposit/PE cases to have both values greater than zero.

wdyt?

pm/recipient_test.go
```diff
@@ -0,0 +1,851 @@
...
+	b.SetPenaltyEscrow(sender, big.NewInt(500))
+
+	params, err := r.TicketParams(sender)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Test invalid recipientRand revealed
```
eladmallel, line 314: Is this comment accurately describing the next line? - thought we're revealing the reused rand only later in the test

pm/recipient.go
yondonfu: `sender.Bytes()...` destructures `[]byte` from a slice into a list of arguments i.e. individual bytes which is necessary or else `append` will add `sender.Bytes()` as a single element to the slice `seed.Bytes()` when we really want `append` to add each byte of `sender.Bytes()` as its own element.

pm/recipient_test.go
```diff
@@ -0,0 +1,851 @@
...
+		t.Fatal("expected valid winning ticket")
+	}
+
+	err = r.RedeemWinningTickets(ticket.RecipientRandHash.Hex())
+	if err == nil {
+		t.Error("expected get deposit error")
+	}
+}
+
+func TestRedeemWinningTickets_SingleTicket_GetPenaltyEscrowError(t *testing.T) {
```
eladmallel, line 460: I'm slightly confused here because in the test `TestRedeemWinningTickets_SingleTicket_GetDepositError` I expected the deposit to be set to zero when in fact PE is set to zero, and in this test `TestRedeemWinningTickets_SingleTicket_GetPenaltyEscrowError` I expected PE to be set to zero, when in fact deposit is set to zero.
In both tests I also expected the other value to be explicitly set to some non-zero value.

wdyt?

pm/recipient_test.go
```diff
@@ -0,0 +1,851 @@
...
+	won, err := r.ReceiveTicket(ticket, sig, params.Seed)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if !won {
+		t.Fatal("expected valid winning ticket")
+	}
+
+	err = r.RedeemWinningTickets(ticket.RecipientRandHash.Hex())
+	if err == nil {
```
eladmallel, line 483: Same question for this test and the one above - is there a way we can make this check more specific to know that we're indeed failing for the reason we expect to be failing?

pm/recipient_test.go
yondonfu: To clarify: by moving this before the code under test do you mean `r.ReceiveTicket` in this case?

pm/recipient_test.go
yondonfu: No particular reason - will make the change

pm/recipient_test.go
yondonfu: Yeah this comment would be better placed when we actually call `r.ReceiveTicket` expecting a failure

pm/recipient_test.go
yondonfu: These tests are to test if `TicketBroker.GetDeposit()` and `TicketBroker.GetPenaltyEscrow()` return an error (i.e. if the RPC call to fetch the deposit/penalty escrow failed). In order to stub this error, `stubBroker` is configured such that if `SetDeposit()`/`SetPenaltyEscrow()` is never called for a given address, then it will return an error. Thus, for the `TestRedeemWinningTickets_SingleTicket_GetDepositError`, the deposit is never set, but the PE is set to zero so an error would be returned by `stubError` since the deposit is never set.

I can make a few updates to make this clearer:

- Use an explicit boolean flag to trigger errors for `TicketBroker.GetDeposit()`/`TicketBroker.GetPenaltyEscrow`
- Create `stubBroker` with default non-zero values for deposit and PE and explicitly set them to zero in tests if needed

pm/recipient_test.go
```diff
@@ -0,0 +1,851 @@
...
+	params, err := r.TicketParams(sender)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Config stub validator with valid winning tickets
+	v.SetIsValidTicket(true)
+	v.SetIsWinningTicket(true)
+	// Config stub broker with zero deposit and penalty escrow and to fail redemption
+	b.SetDeposit(sender, big.NewInt(0))
```
eladmallel: Why should deposit be zero here?

pm/recipient.go
eladmallel: You're right - I didn't keep in mind that they take different data structures as inputs. Nevermind :)

pm/recipient.go
eladmallel: Yeah, I think this line of reasoning makes sense. My concern was indeed mainly around silly temporary errors that could be resolved with a retry, so I think having retries at a lower level could make sense. As long as ticket redemption isn't part of the critical path of video processing I think retries are a great idea.

Perhaps we should capture this thought as a SS issue?

pm/recipient_test.go
eladmallel: Yes :)

pm/recipient_test.go
eladmallel: Awesome, I think both your suggestions would make these tests easier to understand!

pm/recipient_test.go
yondonfu: Updated in a72a3af to set up stub broker with non-zero deposit and penalty escrow for the test sender so that we only have to explicitly set deposit and penalty escrow to zero when testing for errors

eladmallel (approved):
🚢

pm/recipient.go
yondonfu: Tracking in #654

PR #651, opened 2018-12-21
PM Sender

### LINKED ISSUE #629: PM Sender
Support B's state management and ticket creation for PM interactions.

A `Sender` struct and logic under the `pm` package, including:

- Random numbers state management
- Using `Ticket` and `TicketBroker` to manage B's funds and create tickets

This implementation will be accompanied by unit tests while stubbing out dependencies such at Eth.

Nothing to add.

**Additional context**
Nothing to add.

eladmallel:
Adds the PM `Sender` interface and implementation, allowing Bs to create PM tickets based on params received from O (B/O PM param exchange has yet to be implemented).

- `pm.Sender` with support for `StartSession` and `CreateTicket`
- `StartSession` begins a PM session with specific params, e.g. `recipientRandHash`, `winProb` and a few more
- `CreateTicket` creates the next valid ticket given a specific session
- Refactored `AccountManager` to be an interface so it can be stubbed out in `sender_test.go`

- Unit tests for `defaultSender`, which is the only `Sender` implementation as of now
- Used `go test -race` and made sure the race condition detector did not find any race conditions

- Fixes #629 

**Checklist:**

### DIFF
--- a/eth/accountmanager.go
+++ b/eth/accountmanager.go
@@ -22,14 +22,23 @@ var (
 	ErrPassphraseMismatch = fmt.Errorf("passphrases do not match")
 )
 
-type AccountManager struct {
-	Account accounts.Account
+type AccountManager interface {
+	Unlock(passphrase string) error
+	Lock() error
+	CreateTransactOpts(gasLimit uint64, gasPrice *big.Int) (*bind.TransactOpts, error)
+	SignTx(signer types.Signer, tx *types.Transaction) (*types.Transaction, error)
+	Sign(msg []byte) ([]byte, error)
+	Account() accounts.Account
+}
+
+type DefaultAccountManager struct {
+	account accounts.Account
 
 	unlocked bool
 	keyStore *keystore.KeyStore
 }
 
-func NewAccountManager(accountAddr ethcommon.Address, keystoreDir string) (*AccountManager, error) {
+func NewAccountManager(accountAddr ethcommon.Address, keystoreDir string) (AccountManager, error) {
 	keyStore := keystore.NewKeyStore(keystoreDir, keystore.StandardScryptN, keystore.StandardScryptP)
 
 	acctExists := keyStore.HasAddress(accountAddr)
@@ -43,7 +52,7 @@ func NewAccountManager(accountAddr ethcommon.Address, keystoreDir string) (*Acco
 		glog.Infof("Please enter a passphrase to encrypt the Private Keystore file for the Ethereum account.")
 		glog.Infof("This process will ask for this passphrase every time it is launched")
 		glog.Infof("(no characters will appear in Terminal when the passphrase is entered)")
-		
+
 		// Account does not exist yet, set it up
 		acct, err = createAccount(keyStore)
 		if err != nil {
@@ -61,26 +70,26 @@ func NewAccountManager(accountAddr ethcommon.Address, keystoreDir string) (*Acco
 
 	glog.Infof("Using Ethereum account: %v", acct.Address.Hex())
 
-	return &AccountManager{
-		Account:  acct,
+	return &DefaultAccountManager{
+		account:  acct,
 		unlocked: false,
 		keyStore: keyStore,
 	}, nil
 }
 
 // Unlock account indefinitely using underlying keystore
-func (am *AccountManager) Unlock(passphrase string) error {
+func (am *DefaultAccountManager) Unlock(passphrase string) error {
 	var err error
 
-	err = am.keyStore.Unlock(am.Account, passphrase)
+	err = am.keyStore.Unlock(am.account, passphrase)
 	if err != nil {
 		if passphrase != "" {
 			return err
 		}
-		glog.Infof("Please enter the passphrase to unlock Ethereum account %v", am.Account.Address.Hex())
+		glog.Infof("Please enter the passphrase to unlock Ethereum account %v", am.account.Address.Hex())
 
 		passphrase, err = getPassphrase(false)
-		err = am.keyStore.Unlock(am.Account, passphrase)
+		err = am.keyStore.Unlock(am.account, passphrase)
 		if err != nil {
 			return err
 		}
@@ -88,14 +97,14 @@ func (am *AccountManager) Unlock(passphrase string) error {
 
 	am.unlocked = true
 
-	glog.Infof("Unlocked ETH account: %v", am.Account.Address.Hex())
+	glog.Infof("Unlocked ETH account: %v", am.account.Address.Hex())
 
 	return nil
 }
 
 // Lock account using underlying keystore and remove associated private key from memory
-func (am *AccountManager) Lock() error {
-	err := am.keyStore.Lock(am.Account.Address)
+func (am *DefaultAccountManager) Lock() error {
+	err := am.keyStore.Lock(am.account.Address)
 	if err != nil {
 		return err
 	}
@@ -107,17 +116,17 @@ func (am *AccountManager) Lock() error {
 
 // Create transact opts for client use - account must be unlocked
 // Can optionally set gas limit and gas price used
-func (am *AccountManager) CreateTransactOpts(gasLimit uint64, gasPrice *big.Int) (*bind.TransactOpts, error) {
+func (am *DefaultAccountManager) CreateTransactOpts(gasLimit uint64, gasPrice *big.Int) (*bind.TransactOpts, error) {
 	if !am.unlocked {
 		return nil, ErrLocked
 	}
 
 	return &bind.TransactOpts{
-		From:     am.Account.Address,
+		From:     am.account.Address,
 		GasLimit: gasLimit,
 		GasPrice: gasPrice,
 		Signer: func(signer types.Signer, address ethcommon.Address, tx *types.Transaction) (*types.Transaction, error) {
-			if address != am.Account.Address {
+			if address != am.account.Address {
 				return nil, errors.New("not authorized to sign this account")
 			}
 
@@ -127,8 +136,8 @@ func (am *AccountManager) CreateTransactOpts(gasLimit uint64, gasPrice *big.Int)
 }
 
 // Sign a transaction. Account must be unlocked
-func (am *AccountManager) SignTx(signer types.Signer, tx *types.Transaction) (*types.Transaction, error) {
-	signature, err := am.keyStore.SignHash(am.Account, signer.Hash(tx).Bytes())
+func (am *DefaultAccountManager) SignTx(signer types.Signer, tx *types.Transaction) (*types.Transaction, error) {
+	signature, err := am.keyStore.SignHash(am.account, signer.Hash(tx).Bytes())
 	if err != nil {
 		return nil, err
 	}
@@ -137,11 +146,15 @@ func (am *AccountManager) SignTx(signer types.Signer, tx *types.Transaction) (*t
 }
 
 // Sign byte array message. Account must be unlocked
-func (am *AccountManager) Sign(msg []byte) ([]byte, error) {
+func (am *DefaultAccountManager) Sign(msg []byte) ([]byte, error) {
 	personalMsg := fmt.Sprintf("\x19Ethereum Signed Message:\n%d%s", 32, msg)
 	personalHash := crypto.Keccak256([]byte(personalMsg))
 
-	return am.keyStore.SignHash(am.Account, personalHash)
+	return am.keyStore.SignHash(am.account, personalHash)
+}
+
+func (am *DefaultAccountManager) Account() accounts.Account {
+	return am.account
 }
 
 // Get account from keystore using hex address
--- a/eth/accountmanager.go
+++ b/eth/accountmanager.go
@@ -31,7 +31,7 @@ type AccountManager interface {
 	Account() accounts.Account
 }
 
-type DefaultAccountManager struct {
+type accountManager struct {
 	account accounts.Account
 
 	unlocked bool
@@ -70,7 +70,7 @@ func NewAccountManager(accountAddr ethcommon.Address, keystoreDir string) (Accou
 
 	glog.Infof("Using Ethereum account: %v", acct.Address.Hex())
 
-	return &DefaultAccountManager{
+	return &accountManager{
 		account:  acct,
 		unlocked: false,
 		keyStore: keyStore,
@@ -78,7 +78,7 @@ func NewAccountManager(accountAddr ethcommon.Address, keystoreDir string) (Accou
 }
 
 // Unlock account indefinitely using underlying keystore
-func (am *DefaultAccountManager) Unlock(passphrase string) error {
+func (am *accountManager) Unlock(passphrase string) error {
 	var err error
 
 	err = am.keyStore.Unlock(am.account, passphrase)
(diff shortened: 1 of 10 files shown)


pm/helpers.go
```diff
@@ -0,0 +1,11 @@
+package pm
+
+import ethcommon "github.com/ethereum/go-ethereum/common"
+
+func hexToHash(in string) ethcommon.Hash {
+	return ethcommon.BytesToHash(ethcommon.FromHex(in))
+}
+
+func hashToHex(hash ethcommon.Hash) string {
+	return ethcommon.ToHex(hash.Bytes())
```
yondonfu: Looks like the `Hash` type has a [Hex()](https://github.com/ethereum/go-ethereum/blob/master/common/types.go#L72) method that converts a `Hash` type to a hex string which we can use instead of defining an additional helper here.

pm/helpers.go
```diff
@@ -0,0 +1,11 @@
+package pm
+
+import ethcommon "github.com/ethereum/go-ethereum/common"
+
+func hexToHash(in string) ethcommon.Hash {
+	return ethcommon.BytesToHash(ethcommon.FromHex(in))
```
yondonfu: Looks like there is already a [HexToHash()](https://github.com/ethereum/go-ethereum/blob/master/common/types.go#L63) in `go-ethereum/common` that converts a hex string into a `Hash` type which we can use instead of defining an additional helper here.

pm/sender.go
```diff
@@ -0,0 +1,82 @@
...
+func (s *defaultSender) CreateTicket(sessionID string) (*Ticket, *big.Int, []byte, error) {
+	recipientRandHash := hexToHash(sessionID)
+
+	tempSession, ok := s.sessions.Load(sessionID)
+	if !ok {
+		return nil, nil, nil, errors.Errorf("cannot create a ticket for an unknown session: %+v", sessionID)
+	}
+	session := tempSession.(*session)
+
+	senderNonce := atomic.AddUint64(&session.senderNonce, 1)
```
yondonfu, line 65: Since `atomic.AddUint64` mutates `session.senderNonce` and returns the updated value, I think the first ticket for a sender would always have `senderNonce = 1`. Should we change this so that the first ticket for a sender always has `senderNonce = 0`?

pm/sender_test.go
```diff
@@ -0,0 +1,258 @@
...
+				RecipientRandHash: ticketParams.RecipientRandHash,
+			})
+			currentCount := atomic.AddInt32(&sessionCount, 1)
+			if currentCount >= 100 {
+				ch <- struct{}{}
+			}
+		}()
+	}
+	// waiting for all session to be created
+	<-ch
```
yondonfu: Instead of waiting for a completion signal using a channel here and a shared counter, perhaps using [sync.WaitGroup](https://golang.org/pkg/sync/#WaitGroup) to wait for a set of goroutines to complete would be simpler and easier to read?

pm/sender.go
```diff
@@ -0,0 +1,82 @@
...
+
+	return sessionID
+}
+
+func (s *defaultSender) CreateTicket(sessionID string) (*Ticket, *big.Int, []byte, error) {
+	recipientRandHash := hexToHash(sessionID)
+
+	tempSession, ok := s.sessions.Load(sessionID)
+	if !ok {
+		return nil, nil, nil, errors.Errorf("cannot create a ticket for an unknown session: %+v", sessionID)
```
yondonfu: Should we use an error constant here for ease of testing?

pm/sender_test.go
```diff
@@ -0,0 +1,258 @@
...
+		t.Errorf("expected an error for nonexistent session")
+	}
+	if !strings.Contains(err.Error(), "unknown session") {
+		t.Errorf("expected error to contain 'unknown session' but instead got: %v", err.Error())
+	}
+}
+
+func TestCreateTicket_GivenValidSessionId_UsesSessionParamsInTicket(t *testing.T) {
+	sender := defaultNewSender(t)
+	am := sender.accountManager.(*stubAccountManager)
```
yondonfu, line 80: Could we have the `stubAccountManager` instance available here and add exported methods (or export the relevant fields) for `stubAccountManager` for setting these boolean flags to avoid using a type assertion?

pm/sender_test.go
```diff
@@ -0,0 +1,258 @@
...
+		WinProb:           big.NewInt(2222),
+		Seed:              big.NewInt(3333),
+		RecipientRandHash: recipientRandHash,
+	}
+	sessionID := sender.StartSession(recipient, ticketParams)
+
+	ticket, actualSeed, actualSig, err := sender.CreateTicket(sessionID)
+
+	if err != nil {
+		t.Errorf("error tryint to create a ticket: %v", err)
```
yondonfu: "tryint" -> "trying"?

pm/sender_test.go
```diff
@@ -0,0 +1,258 @@
...
+	if ticket.WinProb != ticketParams.WinProb {
+		t.Errorf("expected ticket WinProb %v to be %v", ticket.WinProb, ticketParams.WinProb)
+	}
+	if ticket.SenderNonce != 1 {
+		t.Errorf("expected ticket SenderNonce %d to be 1", ticket.SenderNonce)
+	}
+	if actualSeed != ticketParams.Seed {
+		t.Errorf("expected actual seed %d to be %d", actualSeed, ticketParams.Seed)
+	}
+	if !reflect.DeepEqual(actualSig, am.signResponse) {
```
yondonfu: Thoughts on using [bytes.Equal](https://golang.org/pkg/bytes/#Equal) here for clarity on types being compared?

pm/sender_test.go
```diff
@@ -0,0 +1,258 @@
...
+			lock.Unlock()
+
+			currentCount := atomic.AddInt32(&ticketCount, 1)
+			if currentCount >= int32(totalTickets) {
+				ch <- struct{}{}
+			}
+		}()
+	}
+	// waiting for all session to be created
+	<-ch
```
yondonfu: Echoing previous question/comment re: using sync.WaitGroup

yondonfu (changes requested):

pm/helpers.go
eladmallel: Great point, fixing.

pm/helpers.go
eladmallel: Great, fixing!

pm/sender.go
eladmallel: We could do that by assigning `senderNonce - 1` to the ticket, but I thought that would be more confusing, and didn't really see a problem with nonce starting from 1 and not 0.

Another solution could be to assign `-1` as the first value of `session.senderNonce` but that also felt clunky and confusing.

Do you have a solution in mind that is easy to read and still uses the `atomic` functionality?

pm/sender_test.go
eladmallel: Beautiful, thanks for teaching me this awesome new way.

pm/sender_test.go
eladmallel: Do you mean for example, instead of setting `saveSignRequest`, have a function like `ShouldSaveSignRequest(bool)`?

For the case of a stub I don't think we gain much by adding a function that simply wraps a boolean, that's why I chose to stick with this approach.

What are the benefits you see us having from adding these wrapper methods?

As for the type assertion, are you suggesting that `defaultNewSender` should return both sender and accountManager instead of just the sender? - I think that can definitely work. I avoided this at first because it kept all the other tests that don't need the accountManager a bit cleaner.
Do you think it's worth adding the extra return value?

pm/sender.go
eladmallel: That could work. However, before I go for it, I'm curious as to how you would resolve the string const with the idea of adding runtime variable values to the error string?

For example, if the const string were `"cannot create a ticket for an unknown session: %+v"`, it would not really help us with testing.

An alternative I can think of is to concatenate the const error message and the variable in question, but I was having resistance to this approach because it feels unidiomatic.

wdyt?

eladmallel: Now that we've discussed the idea of erring on the side of detailed errors for now - would love to get your take on if and how you would change this line of code?

pm/sender.go
yondonfu: I guess my initial reaction was why wouldn't `senderNonce` start at 0 (i.e. is there some reason for it not to). However, I don't feel particularly strongly about this and I think starting a nonce at 0 or 1 is ultimately a matter of preference (as you mentioned, I don't think there are positive/negative implications for either relative to the other) so feel free to go with what you're the most comfortable with.

If you did want to start `senderNonce` at 0, could you create the ticket prior to updating `session.senderNonce`, so the ticket created always uses the current nonce value prior to its update?

eladmallel: I believe the flow you have in mind could result in a race condition, since more than one goroutine could run the same lines of code in parallel that take the current value of `senderNonce`, say 0, and then when the increment it, one will increment to 1, the other to 2, because we're using atomic increment, but both tickets would have `senderNonce` of zero.

Does this make sense?

Basically my thinking is that if we use this atomic increment approach incrementing before ticket creation is the best approach, and if we used a lock then we could definitely increment after ticket creation and start with `senderNonce = 0`.

pm/sender.go
yondonfu: I think if we're using dynamic error strings, we'll have to use string matching while testing anyway so I think the way this error is returned seems fine now.

However, I don't think we need to use `%+v` here - IIUC that verb is only necessary when we need to log an error along with its stack trace of wrapped errors according to https://dave.cheney.net/2016/06/12/stack-traces-and-the-errors-package

eladmallel: yeah, I think you're right. getting rid of the `%+v`.

pm/sender_test.go
yondonfu: > For the case of a stub I don't think we gain much by adding a function that simply wraps a boolean, that's why I chose to stick with this approach.

I actually just realized that since we're writing tests internal to the package, we can just directly set the unexported fields of `stubAccountManager` so ignore my comment on using an exported wrapper function!

> As for the type assertion, are you suggesting that defaultNewSender should return both sender and accountManager instead of just the sender?

I was actually thinking that for this test you could manually instantiate `stubAccountManager` and `sender` so that you can directly use `stubAccountManager` within the test

pm/sender.go
```diff
@@ -0,0 +1,82 @@
...
+
+type session struct {
+	senderNonce uint64
+
+	recipient ethcommon.Address
+
+	ticketParams TicketParams
+}
+
+type defaultSender struct {
```
yondonfu: If we're not exporting this struct right now i.e. consumers of the `pm` package wouldn't know anything about it, why not just call it `sender`?

pm/sender_test.go
eladmallel: I prefer the cost of the type assertion for the benefit of less pain if/when we need to change the way we instantiate Sender (e.g. add more constructor params).

Does this consideration make the current approach reasonable, or do you still think it's better to manually instantiate sender and accountManager in this test function?

eth/accountmanager.go
```diff
@@ -22,14 +22,23 @@ var (
...
+type AccountManager interface {
+	Unlock(passphrase string) error
+	Lock() error
+	CreateTransactOpts(gasLimit uint64, gasPrice *big.Int) (*bind.TransactOpts, error)
+	SignTx(signer types.Signer, tx *types.Transaction) (*types.Transaction, error)
+	Sign(msg []byte) ([]byte, error)
+	Account() accounts.Account
+}
+
+type DefaultAccountManager struct {
```
yondonfu: I see that we're using the naming convention of prepending default to the interface name for the main implementation of an interface. Why is `DefaultAccountManager` exported, but `defaultSender` in `sender.go` not exported?

pm/sender.go
yondonfu: Ah yeah I think you're right about the race condition.

I'm ok with starting `senderNonce = 1` if that means we can avoid implementing an additional lock.

pm/sender_test.go
yondonfu: Hm yeah that's a good consideration. With that in mind I think keeping the type assertion + the current approach for instantiation is reasonable for now

yondonfu (approved):
LGTM 🚢

pm/sender_test.go
```diff
@@ -0,0 +1,249 @@
...
+
+	if err != nil {
+		t.Fatalf("failed generating random bytes: %v", err)
+		return nil
+	}
+
+	return res
+}
+
+func randBytes(size int) ([]byte, error) {
```
j0sh, line 244: Curious why this implementation was taken rather than the one here; this is nice because it doesn't return an error, so there's less boilerplate to handle in the above helpers.

https://github.com/livepeer/go-livepeer/blob/94c0238ddb5f73c6142d5a622d1de1a6fe1dd730/core/streamdata.go#L49-L55

pm/helpers.go
j0sh: The Ethereum `Hex` and related methods have some unexpected behavior: zero-values produce "0" and it appends "0x" regardless of the value. The latter might be defensible as a question of style (note we frequently stripped the leading "0x" in most cases for the v1 protocol), but I don't believe the former is.

I've been starting to use golang's built-in `encoding/hex` instead, which should be fine for situations where we don't need compatibility with the Ethereum implementation. See the earlier discussion around this: https://github.com/livepeer/go-livepeer/pull/604/files#r239601925

pm/sender_test.go
eladmallel: No reason except my not being aware of this implementation. Practically speaking if I really wanted to reuse this code I would have to refactor it to a super basic package that all packages can import, because in the current state if I just used it in `pm` it would result in a dependency cycle.

What do you think about creating such a package, and how would you like us to capture all these little refactoring ideas that come up here?

pm/helpers.go
eladmallel: Thanks for this reference - I agree this is confusing. My followup on this is similar to your other note on this PR - I'd love to have a barebones helpers package that depends on nothing but golang and vendor packages such that all of our packages can import it, and there we can gradually refactor all these heavily reused functions, and resolve the "ToHex" implementation discussion once and for all.

wdyt?

pm/sender_test.go
j0sh: Ah I was referring more to the construct of generating the bytes via iterated `rand.Uint32` rather than `rand.Read` since the former doesn't throw an error. That seems like it would simplify a lot of the invocations of `randBytes`. Agree that trying to use `core.RandomIDGenerator` as-is would be complex. We actually have three or four random-bytes helpers scattered through the code 🙃 

I like the idea of a package for helpers, but it seems like `common` should mostly already be "that" package. The DB stuff is an interesting exception there, but I have some further thoughts which I'll elaborate in the [other thread](https://github.com/livepeer/go-livepeer/pull/655#discussion_r244243772).

eladmallel: I'm still of the opinion that we would benefit from separating DB code out of `common`, and would still love to keep the `pm` package as decoupled as possible from any Livepeer-specific assumptions, so this implementation could potentially be reused as an open source package later.

To summarize I am taking the conversation here and in the other thread as potential refactoring TODOs between MVS and the next milestone, since I'd love to discuss it a bit further before we land on a specific action plan.

Issue #632, opened 2018-12-14
pm: recipient redeems old valid winning tickets on start up


yondonfu:
**Is your feature request related to a problem? Please describe.**

If O crashes prior to a redeeming a winning ticket on-chain, it will lose out on the pay out from the ticket unless O has a way to persist the ticket data such that it can try to redeem the ticket at a later point in time.

**Describe the solution you'd like**

Once #621 is implemented, O will always store winning ticket data in its DB at least until the ticket is successfully redeemed on-chain.

On start up, O can do the following:

- Query its DB for all winning tickets [1]
- For each winning ticket check that it is still valid (given expiration requirements) and that it has not already been redeemed on-chain. We perform the last check because it might have been the case that O submitted a tx, crashed and then the tx confirmed.
- Redeem each valid winning ticket on-chain

[1] After observing a successful on-chain tx we can either clear the winning ticket from the DB or mark the winning ticket as redeemed by storing the confirmation block number in the relevant row of the winning tickets table. If we do the former, we can just query for all rows in the table here. If we do the latter, we can query for rows that do not have a confirmation block number. 

**Describe alternatives you've considered**

O does not necessarily need to redeem the old valid winning tickets on start up, but it is unclear if there is a better time to do so at least in the MVS setting where we are not yet implementing any smarter strategy ticket redemption timing (i.e. O could only redeem tickets when the average gas price is at some level and if that level is not met by `X` blocks prior to the ticket's expiration O will redeem the ticket on-chain). Since we are leaving alternative ticket redemption strategies for the future, redeeming old valid winning tickets on start up seems like the simplest solution for now.

j0sh:
Keeping a historical list of winning tickets seems useful; I like using the confirmation block number to indicate rows that have completed processing.

Just curious: are we implementing ticket DB persistence for MVS? I was under the impression that would come in during SS or later. Cool if that can get in sooner though.

yondonfu:
> Keeping a historical list of winning tickets seems useful; I like using the confirmation block number to indicate rows that have completed processing.

Agreed.

> Just curious: are we implementing ticket DB persistence for MVS?

The thought was that it would be nice to add it sooner rather than later especially once we start testing with more realistic blockchain conditions (i.e. dealing with on-chain transactions that do not confirm within pre-defined period of time). Although adding persistence is lower on the MVS priority list with the higher priority items being getting the happy ticket flow working (i.e. very quick on-chain transaction confirmation)

[...]

PR #655, opened 2018-12-27
Persist Winning Tickets

### LINKED ISSUE #621: pm: Store winning tickets on-disk with corresponding recipientRand value

eladmallel:
Supports the storage of winning PM tickets into the node's local database.

- A new `winningTickets` table in the local database
- `db.go` implements the `TicketStore` interface
- A new testing dependency for succinct assertions (and more!)
- Had to change `SenderNonce` to `uint32` since the SQL library doesn't support very high values of `uint64`
- Refactor to abstract away the `AccountManager` dependency such that the `pm` package no longer imports anything from the `eth` package - this was critical to resolve a cyclical dependency issue (since `common` now imports `pm` for the `pm.Ticket` type in the new function signatures)

- Unit tests that interact with an actual test DB instance

- Fixes https://github.com/livepeer/go-livepeer/issues/621

**Checklist:**

### DIFF
--- a/common/db.go
+++ b/common/db.go
@@ -165,7 +165,8 @@ var schema = `
 		winProb BLOB,
 		senderNonce INTEGER,
 		recipientRand BLOB,
-		sig BLOB
+		sig BLOB,
+		sessionID STRING
 	);
 `
 
@@ -409,7 +410,7 @@ func InitDB(dbPath string) (*DB, error) {
 	d.withdrawableUnbondingLocks = stmt
 
 	// Winning tickets prepared statements
-	stmt, err = db.Prepare("INSERT INTO winningTickets(sender, recipient, faceValue, winProb, senderNonce, recipientRand, sig) VALUES(?, ?, ?, ?, ?, ?, ?)")
+	stmt, err = db.Prepare("INSERT INTO winningTickets(sender, recipient, faceValue, winProb, senderNonce, recipientRand, sig, sessionID) VALUES(?, ?, ?, ?, ?, ?, ?, ?)")
 	if err != nil {
 		glog.Error("Unable to prepare insertWinningTicket ", err)
 		d.Close()
@@ -899,10 +900,10 @@ func (db *DB) UnbondingLocks(currentRound *big.Int) ([]*DBUnbondingLock, error) 
 	return unbondingLocks, nil
 }
 
-func (db *DB) InsertWinningTicket(sender ethcommon.Address, recipient ethcommon.Address, faceValue *big.Int, winProb *big.Int, senderNonce uint64, recipientRand *big.Int, sig []byte) error {
+func (db *DB) InsertWinningTicket(sender ethcommon.Address, recipient ethcommon.Address, faceValue *big.Int, winProb *big.Int, senderNonce uint64, recipientRand *big.Int, sig []byte, sessionID string) error {
 	glog.V(DEBUG).Infof("db: Inserting winning ticket from %v, recipientRand %d, senderNonce %d", sender.Hex(), recipientRand, senderNonce)
 
-	_, err := db.insertWinningTicket.Exec(sender.Hex(), recipient.Hex(), faceValue.Bytes(), winProb.Bytes(), senderNonce, recipientRand.Bytes(), sig)
+	_, err := db.insertWinningTicket.Exec(sender.Hex(), recipient.Hex(), faceValue.Bytes(), winProb.Bytes(), senderNonce, recipientRand.Bytes(), sig, sessionID)
 
 	if err != nil {
 		// TODO wrap with custom error
--- a/common/db.go
+++ b/common/db.go
@@ -163,7 +163,7 @@ var schema = `
 		recipient STRING,
 		faceValue BLOB,
 		winProb BLOB,
-		senderNonce UNSIGNED BIG INT,
+		senderNonce INTEGER,
 		recipientRand BLOB,
 		sig BLOB
 	);
--- a/common/db.go
+++ b/common/db.go
@@ -39,6 +39,7 @@ type DB struct {
 	useUnbondingLock           *sql.Stmt
 	unbondingLocks             *sql.Stmt
 	withdrawableUnbondingLocks *sql.Stmt
+	insertWinningTicket        *sql.Stmt
 }
 
 type DBJob struct {
@@ -156,6 +157,16 @@ var schema = `
 	);
 	-- Index to only retrieve unbonding locks that have not been used
 	CREATE INDEX IF NOT EXISTS idx_unbondinglocks_usedblock ON unbondingLocks(usedBlock);
+
+	CREATE TABLE IF NOT EXISTS winningTickets (
+		sender STRING,
+		recipient STRING,
+		faceValue BLOB,
+		winProb BLOB,
+		senderNonce UNSIGNED BIG INT,
+		recipientRand BLOB,
+		sig BLOB
+	);
 `
 
 func NewDBJob(id *big.Int, streamID string,
@@ -397,6 +408,15 @@ func InitDB(dbPath string) (*DB, error) {
 	}
 	d.withdrawableUnbondingLocks = stmt
 
+	// Winning tickets prepared statements
+	stmt, err = db.Prepare("INSERT INTO winningTickets(sender, recipient, faceValue, winProb, senderNonce, recipientRand, sig) VALUES(?, ?, ?, ?, ?, ?, ?)")
+	if err != nil {
+		glog.Error("Unable to prepare insertWinningTicket ", err)
+		d.Close()
+		return nil, err
+	}
+	d.insertWinningTicket = stmt
+
 	glog.V(DEBUG).Info("Initialized DB node")
 	return &d, nil
 }
@@ -459,6 +479,9 @@ func (db *DB) Close() {
 	}
 	if db.withdrawableUnbondingLocks != nil {
 		db.withdrawableUnbondingLocks.Close()
+	}
+	if db.insertWinningTicket != nil {
+		db.insertWinningTicket.Close()
 	}
 	if db.dbh != nil {
 		db.dbh.Close()
@@ -874,4 +897,16 @@ func (db *DB) UnbondingLocks(currentRound *big.Int) ([]*DBUnbondingLock, error) 
 		unbondingLocks = append(unbondingLocks, &unbondingLock)
 	}
 	return unbondingLocks, nil
+}
+
+func (db *DB) InsertWinningTicket(sender ethcommon.Address, recipient ethcommon.Address, faceValue *big.Int, winProb *big.Int, senderNonce uint64, recipientRand *big.Int, sig []byte) error {
+	glog.V(DEBUG).Infof("db: Inserting winning ticket from %v, recipientRand %d, senderNonce %d", sender.Hex(), recipientRand, senderNonce)
+
+	_, err := db.insertWinningTicket.Exec(sender.Hex(), recipient.Hex(), faceValue.Bytes(), winProb.Bytes(), senderNonce, recipientRand.Bytes(), sig)
+
+	if err != nil {
+		// TODO wrap with custom error
+		return err
+	}
+	return nil
 }
--- a/common/db.go
+++ b/common/db.go
@@ -929,7 +929,7 @@ func (db *DB) StoreWinningTicket(sessionID string, ticket *pm.Ticket, sig []byte
 	_, err := db.insertWinningTicket.Exec(ticket.Sender.Hex(), ticket.Recipient.Hex(), ticket.FaceValue.Bytes(), ticket.WinProb.Bytes(), ticket.SenderNonce, recipientRand.Bytes(), ticket.RecipientRandHash.Hex(), sig, sessionID)
 
 	if err != nil {
-		return errors.Wrapf(err, "failed inserting winning ticket. sessionID: %v, ticket: %v", sessionID, ticket)
+		return errors.Wrapf(err, "failed inserting winning ticket for sessionID: %v, ticket: %v", sessionID, ticket)
 	}
 	return nil
 }
--- a/common/db.go
+++ b/common/db.go
@@ -937,7 +937,7 @@ func (db *DB) LoadWinningTickets(sessionID string) (tickets []*pm.Ticket, sigs [
 	for rows.Next() {
 		var sender, recipient, sessionID string
 		var faceValue, winProb, recipientRandBytes, sig []byte
-		var senderNonce uint64
+		var senderNonce uint32
 
 		err = rows.Scan(&sender, &recipient, &faceValue, &winProb, &senderNonce, &recipientRandBytes, &sig, &sessionID)
 		if err != nil {
--- a/common/db.go
+++ b/common/db.go
@@ -3,7 +3,6 @@ package common
 import (
 	"bytes"
 	"database/sql"
-	"errors"
 	"math/big"
 	"text/template"
 	"time"
@@ -14,6 +13,7 @@ import (
 	"github.com/livepeer/go-livepeer/pm"
 	"github.com/livepeer/lpms/ffmpeg"
 	_ "github.com/mattn/go-sqlite3"
+	"github.com/pkg/errors"
 )
 
 type DB struct {
@@ -920,8 +920,7 @@ func (db *DB) StoreWinningTicket(sessionID string, ticket *pm.Ticket, sig []byte
 	_, err := db.insertWinningTicket.Exec(ticket.Sender.Hex(), ticket.Recipient.Hex(), ticket.FaceValue.Bytes(), ticket.WinProb.Bytes(), ticket.SenderNonce, recipientRand.Bytes(), ticket.RecipientRandHash.Hex(), sig, sessionID)
 
 	if err != nil {
-		// TODO wrap with custom error
-		return err
+		return errors.Wrapf(err, "failed inserting winning ticket. sessionID: %v, ticket: %v", sessionID, ticket)
 	}
 	return nil
 }
@@ -931,8 +930,8 @@ func (db *DB) LoadWinningTickets(sessionID string) (tickets []*pm.Ticket, sigs [
 	defer rows.Close()
 
 	if err != nil {
-		// TODO wrap with custom error
-		return nil, nil, nil, err
+		err = errors.Wrapf(err, "failed loading winning tickets for sessionID %v", sessionID)
+		return
 	}
 
 	for rows.Next() {
@@ -942,7 +941,7 @@ func (db *DB) LoadWinningTickets(sessionID string) (tickets []*pm.Ticket, sigs [
 
 		err = rows.Scan(&sender, &recipient, &faceValue, &winProb, &senderNonce, &recipientRandBytes, &recipientRandHash, &sig, &sessionID)
 		if err != nil {
-			// TODO
+			err = errors.Wrapf(err, "failed scanning a winning ticket row for sessionID %v", sessionID)
 			return
 		}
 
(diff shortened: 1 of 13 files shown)


eladmallel:
@j0sh would love your eyes on this one in particular since we're trying out a new package for testing!

common/db.go
```diff
@@ -875,3 +913,61 @@ func (db *DB) UnbondingLocks(currentRound *big.Int) ([]*DBUnbondingLock, error)
...
+	}
+	if recipientRand == nil {
+		return errors.New("cannot store nil recipientRand")
+	}
+	glog.V(DEBUG).Infof("db: Inserting winning ticket from %v, recipientRand %d, senderNonce %d", ticket.Sender.Hex(), recipientRand, ticket.SenderNonce)
+
+	_, err := db.insertWinningTicket.Exec(ticket.Sender.Hex(), ticket.Recipient.Hex(), ticket.FaceValue.Bytes(), ticket.WinProb.Bytes(), ticket.SenderNonce, recipientRand.Bytes(), ticket.RecipientRandHash.Hex(), sig, sessionID)
+
+	if err != nil {
+		return errors.Wrapf(err, "failed inserting winning ticket. sessionID: %v, ticket: %v", sessionID, ticket)
```
yondonfu: Should we match the error message format used in `LoadWinningTickets()`?

i.e. "failed inserting winning ticket. sessionID ... " -> "failed inserting winning ticket for sessionID ... "

common/db_test.go
```diff
@@ -578,3 +582,311 @@ func TestDBUnbondingLocks(t *testing.T) {
...
 	}
 }
+
+func TestInsertWinningTicket_GivenValidInputs_InsertsOneRowCorrectly(t *testing.T) {
+	dbh, dbraw, err := TempDB(t)
+	defer dbh.Close()
+	defer dbraw.Close()
+	if err != nil {
+		t.Error(err)
+		return
```
yondonfu: Why do we use an explicit return here? Does `t.Fatal()` not return cleanly for this function thereby messing with the defer statements closing the DB handler (although I do see that we are still using `t.Fatal()` elsewhere in this test file)?

common/db_test.go
```diff
@@ -578,3 +582,311 @@ func TestDBUnbondingLocks(t *testing.T) {
...
+	}
+	sessionID, ticket, sig, recipientRand := defaultWinningTicket(t)
+	err = dbh.StoreWinningTicket(sessionID, ticket, sig, recipientRand)
+	if err != nil {
+		t.Fatalf("unexpected errro storing ticket: %v", err)
```
yondonfu: "unexpected errro" -> "unexpected error"?

common/db_test.go
```diff
@@ -578,3 +582,311 @@ func TestDBUnbondingLocks(t *testing.T) {
...
+
+	if err != nil {
+		t.Fatalf("error getting db table count, cannot proceed: %v", err)
+		return -1
+	}
+
+	return count
+}
+
+func randAddressOrFatal(t *testing.T) ethcommon.Address {
```
yondonfu: Should we move some of these test helpers i.e. `randBytesOrFatal`, `randBytes`, etc. into a single place to avoid duplication since `pm/helpers.go` already contains these functions?

common/db_test.go
```diff
@@ -578,3 +582,311 @@ func TestDBUnbondingLocks(t *testing.T) {
...
+	defer dbraw.Close()
+	if err != nil {
+		t.Error(err)
+		return
+	}
+	sessionID, ticket, sig, recipientRand := defaultWinningTicket(t)
+
+	err = dbh.StoreWinningTicket(sessionID, ticket, sig, recipientRand)
+
+	if err != nil {
```
yondonfu: Could we do the following:

```
if assert.Nilf(err, "failed to insert winning ticket with error: %v", err) {
    ...
    assertions assuming that winning ticket insertion succeeded
    ....
}
```

common/db_test.go
```diff
@@ -578,3 +582,311 @@ func TestDBUnbondingLocks(t *testing.T) {
...
+		t.Error(err)
+		return
+	}
+	sig := randBytesOrFatal(42, t)
+	recipientRand := new(big.Int).SetInt64(1234)
+
+	err = dbh.StoreWinningTicket("sessionID", nil, sig, recipientRand)
+
+	assert := assert.New(t)
+	assert.NotNil(err)
```
yondonfu, line 668: IIUC `assert` will not stop test execution so in cases like this if this line fails, we would continue executing the next line which would always fail since `err = nil`. Should we try to avoid the additional assertion errors by structuring the assertions like this:

```
if assert.NotNil(err) {
    assert.Contains(err.Error(), "nil ticket")
}
```

common/db_test.go
```diff
@@ -578,3 +582,311 @@ func TestDBUnbondingLocks(t *testing.T) {
...
+		return
+	}
+	sessionID, ticket, sig, recipientRand := defaultWinningTicket(t)
+	err = dbh.StoreWinningTicket(sessionID, ticket, sig, recipientRand)
+	if err != nil {
+		t.Fatalf("unexpected errro storing ticket: %v", err)
+	}
+
+	tickets, sigs, recipientRands, err := dbh.LoadWinningTickets(sessionID)
+	if err != nil {
```
yondonfu: `require.Nil(...)`?

common/db_test.go
```diff
@@ -578,3 +582,311 @@ func TestDBUnbondingLocks(t *testing.T) {
...
+		t.Fatalf("unexpected error loading tickets: %v", err)
+	}
+	if len(tickets) != 1 || len(sigs) != 1 || len(recipientRands) != 1 {
```
yondonfu: Should we use `assert` here?

common/db_test.go
```diff
@@ -578,3 +582,311 @@ func TestDBUnbondingLocks(t *testing.T) {
...
+		t.Error(err)
+		return
+	}
+
+	// Two tickets in the first session
+	firstSessionID := "first session"
+	_, ticket0, sig0, recipientRand0 := defaultWinningTicket(t)
+	err = dbh.StoreWinningTicket(firstSessionID, ticket0, sig0, recipientRand0)
+	if err != nil {
+		t.Fatalf("unexpected errro storing ticket: %v", err)
```
yondonfu: "errro" -> "error"? Same question applies for the other error messages in this test function

common/db_test.go
```diff
@@ -578,3 +582,311 @@ func TestDBUnbondingLocks(t *testing.T) {
...
+	sessionID, ticket, sig, _ := defaultWinningTicket(t)
+
+	err = dbh.StoreWinningTicket(sessionID, ticket, sig, nil)
+
+	assert := assert.New(t)
+	assert.NotNil(err)
+	assert.Contains(err.Error(), "nil recipientRand")
+}
+
+func TestLoadtWinningTicket_GivenStoredTicket_LoadsItCorrectly(t *testing.T) {
```
yondonfu: "TestLoadtWinningTicket_" -> "TestLoadWinningTicket"? Same question applies for the rest of the test function names with this beginning

common/db_test.go
```diff
@@ -578,3 +582,311 @@ func TestDBUnbondingLocks(t *testing.T) {
...
+	tickets, sigs, recipientRands, err := dbh.LoadWinningTickets("some sessionID")
+
+	assert := assert.New(t)
+	assert.Nil(err)
+	assert.Len(tickets, 0)
+	assert.Len(sigs, 0)
+	assert.Len(recipientRands, 0)
+}
+
+func TestLoadtWinningTicket_GivenEmptySessionID_ReturnsEmptySlicesNoError(t *testing.T) {
```
yondonfu: If we ignore that fact that a `pm` recipient will use `recipientRandHash` as the sessionID, technically the DB could accept an empty string as the sessionID and return non-empty slices right?

common/db_test.go
```diff
@@ -578,3 +582,311 @@ func TestDBUnbondingLocks(t *testing.T) {
...
+}
+
+func getRowCountOrFatal(query string, dbraw *sql.DB, t *testing.T) int {
+	var count int
+	row := dbraw.QueryRow(query)
+	err := row.Scan(&count)
+
+	if err != nil {
+		t.Fatalf("error getting db table count, cannot proceed: %v", err)
+		return -1
```
yondonfu: I don't think this return statement would ever be reached?

common/db_test.go
```diff
@@ -578,3 +582,311 @@ func TestDBUnbondingLocks(t *testing.T) {
...
+
+func randBytes(size int) ([]byte, error) {
+	key := make([]byte, size)
+	_, err := rand.Read(key)
+
+	return key, err
+}
+
+func maxUint256OrFatal(t *testing.T) *big.Int {
+	n, ok := new(big.Int).SetString("FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF", 16)
```
yondonfu: An alternative that would not require error checking could be:

```
new(big.Int).Sub(new(big.Int).Exp(big.NewInt(2), big.NewInt(256)), big.NewInt(1))
```

However, the statement is a little less comprehensible so it might not be that much of a win...

pm/stub.go
```diff
@@ -188,40 +187,31 @@ func (v *stubValidator) IsWinningTicket(ticket *Ticket, sig []byte, recipientRan
...
-	return nil
-}
-
-func (am *stubAccountManager) CreateTransactOpts(gasLimit uint64, gasPrice *big.Int) (*bind.TransactOpts, error) {
-	return nil, nil
-}
-
-func (am *stubAccountManager) SignTx(signer types.Signer, tx *types.Transaction) (*types.Transaction, error) {
+// TODO remove this function
+// NOTE: Keeping this function for now because removing it causes the tests to fail when run with the
```
yondonfu, line 199: 😭

yondonfu (changes requested):
Looking good. Left a few minor comments. My main question is regarding the usage of the new `testify` package in tests - should we be consistent with our usage of `assert` and `require` over their vanilla Go test counterparts i.e. `t.Error` and `t.Fatal`?

common/db_test.go
eladmallel: I will change this to use the new `require` package - hope that will do the trick!

common/db_test.go
eladmallel: I'm just switching to `require.Nil(err)` for all the instances where we would use `t.Fatalf`

common/db_test.go
eladmallel: My approach so far is:

- Use `require` or `Fatal` for all errors that are part of the code the test function is relying on, i.e. assumes it works perfectly.
- Use `assert` for any errors that arise from the "code under test", since in this case it's also interesting to learn if we fail due to other errors.

Does this line of thinking make sense to you, or do you think we should think of this differently?

common/db_test.go
eladmallel: oh totally my bad!!

common/db_test.go
eladmallel: Just went ahead and deleted the error messages since the behavior of `require` and `assert` is to print out the failure line, and I think that's good enough for debugging!

common/db_test.go
eladmallel: Yeah, changing that

common/db_test.go
eladmallel: You're right. In this case I was aiming at making sure that we just get an empty result set rather than an error. I didn't find it interesting to add a test where there would be tickets stored with `sessionID = ""`, since I believe the combination of the tests we already have covers it. But please let me know if you think we need this additional test case and I'll add it :)

common/db_test.go
yondonfu: That makes sense! Using `require` seems cleaner than nested asserts

common/db_test.go
yondonfu: 👍 I don't think an additional test case is necessary - I just wanted to clarify that an empty sessionID should return an empty slice in this test case, but does not necessarily always return an empty slice

pm/helpers.go
```diff
@@ -7,7 +7,7 @@ import (
 	ethcommon "github.com/ethereum/go-ethereum/common"
 )
 
-func randHashOrFatal(t *testing.T) ethcommon.Hash {
+func RandHashOrFatal(t *testing.T) ethcommon.Hash {
```
yondonfu, line 10: Feels like these functions should actually live in a common test utils file since they aren't really specific to `pm`. Unfortunately, I think moving these functions to the `common` package might introduce a cycle since the DB code is also in the `common` package? Perhaps if we move the DB code into its own package then we can move these functions into the `common` package - I don't think we have to do this in this PR though

eladmallel: Yes, that was exactly my thinking - that all these test helpers should go under `common` but to do that we must solve the cyclical dependency issue.

Another approach we could take is have a package dedicated to test helpers we could name `testutils` or something similar.

Wdyt?

yondonfu (approved):
LGTM 🚢 Loving the concise tests using testify

pm/helpers.go
j0sh: I've also addressed this issue in the past by wrapping types into DB-specific structs, with external helpers to convert between the DB struct and the original. eg, `pm.Ticket -> common.DBTicket` and so forth.

This has certain benefits; we can incorporate "extra" information that exists in the DB that may not necessarily belong in the PM ticket struct, such as its submission timestamp, tx hash and other transaction status info, block confirmation number, any other fields that may have a slightly different representation, etc.

common/db.go
```diff
@@ -156,6 +159,18 @@ var schema = `
...
+	CREATE TABLE IF NOT EXISTS winningTickets (
+		sender STRING,
+		recipient STRING,
+		faceValue BLOB,
+		winProb BLOB,
+		senderNonce INTEGER,
+		recipientRand BLOB,
+		recipientRandHash STRING,
+		sig BLOB,
+		sessionID STRING
```
j0sh, line 172: Generally I like to incorporate a creation timestamp here as well. Useful for forensic purposes and doesn't affect any queries with a `DEFAULT CURRENT_TIMESTAMP`. Also, some constraints might have been nice such as a composite key consisting of`(recipientRand, senderNonce)` which would have ensured uniqueness of the pair.

Haven't fully read through the approach here yet, but I've used DB constraints as one form of checking to ensure subsequent code-paths are only executed once (eg, preventing double-submits to the blockchain).

pm/helpers.go
eladmallel: @j0sh how does a `DBTicket` relate to the conversation about refactoring these helper functions to a more common package? - I think I get the future benefits of having such a type, just trying to connect the dots to this specific thread

common/db.go
eladmallel: I'm more than happy to add a creation timestamp! Created this issue https://github.com/livepeer/go-livepeer/issues/661.

As for the constraints - the reason I did not add them is because they weren't necessary as of yet, and I don't like adding "future code" that isn't being used when it's added. I really can't see in what scenario we would benefit from this constraint since I don't see a code path that would lead to our saving the same winning ticket twice (since in-memory protections are in place to reject a repeat incoming ticket). So on this one I'm still leaning more towards not adding anything that isn't really needed, but would love to hear your further thoughts to learn more about why you think it's good to add?

As for double-submit prevention - I agree it's a good idea to defend against that, and think we can do that well by adding some ETH-related columns when we get to spending more time on a robust redemption mechanism. I think we will add these DB columns as we solve for this issue https://github.com/livepeer/go-livepeer/issues/632. wdyt?

pm/helpers.go
j0sh: Rather than have `common` import `pm`, we can have `common` export a DBTicket which is converted into a `pm.Ticket` somewhere else, eg in the PM code. That would allow us to keep common helpers in `common` and avoid an import cycle.

yondonfu: In the case of tickets, there is an additional annoyance in that `pm.Recipient.ReceiveTicket()` will use the method `StoreWinningTicket(string, *Ticket, []byte, *big.Int)` defined in the `TicketStore` interface to persist winning tickets. `common.DB` currently implements `TicketStore`. We can avoid importing `pm` in `common` by using `common.DBTicket` in the `TicketStore` interface: `StoreWinningTicket(string, *common.DBTicket, []byte, *big.Int)`. But, now we're importing code specific to `common.DB` into `pm` when it would be preferable to keep `pm` agnostic to how persistence is handled. 

One solution is to destructure the ticket fields in the `TicketStore` method call:

`StoreWinningTicket(string, ethcommon.Address, ethcommon.Address ... rest of ticket fields ..., []byte, *big.Int)`

Neither package would have to import the either. But this leads to a very unwieldy method call with many parameters - not great.

Alternatively, we could use an anonymous struct - the method signature would look ugly, but the method calls would be cleaner.

Another solution is to move the `Ticket` struct into the `common` package (or whatever the name is for the package we use to store common types used throughout go-livepeer). But this creates more external dependencies for the `pm` package when it would preferable to keep it self contained.

Extracting `DB` into its own package that imports `pm` and having the `common` package specifically focused on just helper functions seems cleaner than the above solutions.

j0sh: > But, now we're importing code specific to common.DB into pm when it would be preferable to keep pm agnostic to how persistence is handled.

The PM code right now is quite nicely cleaved with minimal glue needed between the DB and PM itself. However, I do wonder how hard we'll need to work to maintain this clean separation in the future.

One difficulty with the current approach is that nearly all modules could have a legitimate reason to use the DB [1], but if the DB module imports everything, none of those modules can actually invoke the DB itself. That means we need a third module to glue together the DB and the module, or to implement more of the "glue code" within the DB module itself, which bleeds over the concerns.

We could define a generic `Ticket` as an interface to avoid embedding any DB-specific constructs within the `TicketStore` interface (perhaps with a single PMTicketToGenericTicket function for the `Ticket` interface). However, I don't think there is any harm in putting DB-aware code within the PM implementation itself, as long as the interface remains storage-agnostic. We may eventually find it difficult to keep the concrete implementation completely agnostic to the DB.

[1] That most modules could make use of the DB is one reason I like to consider the DB common code, and by keeping it narrowly scoped to processing queries (eg, minimal glue), there is less reason to move the DB into its own module.

> Another solution is to move the Ticket struct into the common package (or whatever the name is for the package we use to store common types used throughout go-livepeer). But this creates more external dependencies for the pm package when it would preferable to keep it self contained.

Presumably `pm` would still be importing common helpers so it wouldn't be a technical obstacle to defining the Ticket struct within `common`. However, I understand the incongruity of separating the definition from the actual use.

common/db.go
j0sh: > why you think it's good to add?

The ability to declare constraints among fields and key relationships is probably my favorite feature of relational DBs. It's a type system that helps ensure the "shape" (consistency and integrity) of the data during storage and retrieval.

>  I really can't see in what scenario we would benefit from this constraint since I don't see a code path that would lead to our saving the same winning ticket twice (since in-memory protections are in place to reject a repeat incoming ticket)

We may have in-memory protections now, but any constraints act as a safety net in case these are missed somehow. Actually, I have been even lazier in the past: I've used the DB in lieu of in-memory protection, because why write additional code when your logic can be enforced declaratively?

Anyway, just my two cents! This is still fine as-is.

PR #656, opened 2018-12-27
eth: update contract bindings and code cleanup

### LINKED ISSUE #652: pm: generate Go contract bindings for LivepeerETHTicketBroker contract

The `pm` package includes a Broker interface that represents an object capable of processing deposits and ticket redemptions. In practice, the main implementation of this interface will be a Go contract binding for [LivepeerETHTicketBroker](https://github.com/livepeer/protocol/blob/pm/contracts/pm/LivepeerETHTicketBroker.sol). We must generate this Go contract binding.

We can generate the contract binding using the `abigen` tool that is a part of the go-ethereum toolchain. We already use `abigen` to generate the rest of our contract bindings.

However, we should investigate the ability of `abigen` to generate a contract binding for a contract that is uses the experimental `ABIEncoderV2` (i.e. LivepeerETHTicketBroker). I made a very quick attempt at using `abigen` to generate the binding for LivepeerETHTicketBroker, but encountered the following error:

```
Failed to generate ABI binding: unsupported arg type: tuple
client.go:14: running "abigen": exit status 255
```

We should investigate this error and if `abigen`, does not currently support contracts that use the experimental `ABIEncoderV2`, we should file a feature request and maybe consider implementing the feature ourselves in order to generate our required contract binding.

We can choose to not use the experimental `ABIEncoderV2` in LivepeerETHTicketBroker, but this would force us to list out all parameters in any function signature where we want to provide a struct which hurts code readability and the usability of the contract API.

yondonfu:

Update contract bindings to reflect the latest changes in the protocol contracts [pm](https://github.com/livepeer/protocol/tree/pm) branch. In preparation for integrating the `pm` package with the rest of the go-livepeer codebase and for interactions with the updated contracts, this PR seeks to remove all remaining V1 protocol job related code.

- Introduces LivepeerETHTicketBroker contract binding
- Updates the rest of the contract bindings
- Remove job related code

I built the node and ran it in broadcaster/orchestrator mode in devenv

Fixes #652 

**Checklist:**


server/mediaserver.go
```diff
@@ -277,23 +277,8 @@ func gotRTMPStreamHandler(s *LivepeerServer) func(url *url.URL, rtmpStrm stream.
...
 
-			minDeposit := big.NewInt(0).Mul(BroadcastPrice, big.NewInt(MinDepositSegmentCount))
-			if deposit.Cmp(minDeposit) < 0 {
-				glog.Errorf("Low deposit (%v) - cannot start broadcast session.  Need at least %v", deposit, minDeposit)
-				if monitor.Enabled {
-					monitor.LogStreamCreateFailed(nonce, "LowDeposit")
-				}
-				return ErrBroadcast
-			}
+			// TODO: Check broadcaster's deposit with TicketBroker
```
eladmallel, line 281: Perhaps worth capturing as an issue? - I'm concerned we might miss this. wdyt?

eladmallel (approved):
Besides the one little comment I added, LGTM. However, I'm probably not the best person to review this kind of change.. wondering if perhaps another set of eyes should review this PR?

yondonfu:
Since master is known to be unstable and in the interest of avoiding further merge conflicts with other people working off of master I'd be ok with getting this merged sooner rather than later. I think it is also worth noting that master as is also won't work with the `pm` branch of the protocol repo.

Putting the merge on hold because I'm going to push an update to this PR to make sure LP client adheres to the `pm/Broker` interface

yondonfu:
661f2c9: client implements Broker

Tested by adding the following line in `client.go` which should raise a compiler error if the client did not implement Broker:

```
var _ pm.Broker = client{}
```

cmd/livepeer_cli/wizard_stats.go
```diff
@@ -325,14 +315,6 @@ func (w *wizard) getEthBalance() string {
 	return e
 }
 
-func (w *wizard) getDeposit() string {
-	e := httpGet(fmt.Sprintf("http://%v:%v/broadcasterDeposit", w.host, w.httpPort))
```
j0sh, line 329: Don't we still need a deposit?

yondonfu: Yeah, but it is with a different contract so this PR cleaned up the old code and the new code for the new contract is included in #658 which fetches all of a PM sender's information in a single request instead of individually fetching the sender's deposit and penalty escrow.

Issue #635, opened 2018-12-14
pm: use recipient in RPC


yondonfu:
**Is your feature request related to a problem? Please describe.**

This issue assumes we've implemented `pm.Recipient` functionality, but haven't yet connected it to the video transcoding flow.

For this issue we want to incorporate `pm.Recipient` into the video flow.

**Describe the solution you'd like**

- On start up, create a `pm.Recipient` for O
- Use the recipient to create random number commitments to send to B
- Validate and check for winning tickets received over RPC (in the header of segments). Replay prevention is included as a part of this.
- Redeem winning tickets when received

PR #657, opened 2018-12-28
PM: Redeem winning tickets upon stream end

eladmallel:
Adds a PM `Recipient` instance to O nodes, and enables O to redeem winning tickets upon video stream end.

- Loading winning tickets from DB in bulk, supporting many `sessionID`s as input
- O tries to redeem winning tickets upon transcoding loop timeout
- Upon O node boot we create a `Recipient` instance that's connected to ETH and DB

- Through unit / integration tests

- Partial contribution to https://github.com/livepeer/go-livepeer/issues/635

**Checklist:**


eth/client_ticketbroker.go
```diff
@@ -69,3 +69,8 @@ func (c *client) IsUsedTicket(ticket *pm.Ticket) (bool, error) {
 
 	return c.LivepeerETHTicketBrokerSession.UsedTickets(ticketHash)
 }
+
+// ApproveSigners approves a set of ETH addresses to sign on behalf of the sender
+func (c *client) ApproveSigners(signers []ethcommon.Address) (*types.Transaction, error) {
```
yondonfu: I don't think we need to define this method since all it does is wrap `c.LivepeerETHTicketBrokerSession.ApproveSigners()` - `client` embeds `LivepeerETHTicketBrokerSession`, so `client` is already a receiver for `ApproveSigners()`

pm/stub.go
```diff
@@ -216,3 +236,24 @@ func (s *stubSigner) Sign(msg []byte) ([]byte, error) {
...
 }
+
+// MockRecipient is useful for testing components that depend on pm.Recipient, such as
+// LivepeerNode
+type MockRecipient struct {
+	mock.Mock
+}
+
+func (m *MockRecipient) ReceiveTicket(ticket *Ticket, sig []byte, seed *big.Int) (sessionID string, won bool, err error) {
+	args := m.MethodCalled("ReceiveTicket", ticket, sig, seed)
```
yondonfu: Could we use [Called()](https://godoc.org/github.com/stretchr/testify/mock#Mock.Called) here which automatically uses the calling function's name instead of `MethodCalled()` which requires explicitly providing the function name?

The same question applies for all usage of `MethodCalled()` for `MockRecipient`.

pm/stub.go
```diff
@@ -216,3 +236,24 @@ func (s *stubSigner) Sign(msg []byte) ([]byte, error) {
 func (s *stubSigner) Account() accounts.Account {
 	return s.account
 }
+
+// MockRecipient is useful for testing components that depend on pm.Recipient, such as
+// LivepeerNode
+type MockRecipient struct {
```
yondonfu, line 242: Since we've been using the "stub" naming convention i.e. "stubSigner", should we remain consistent and name this "StubRecipient"? I don't mind using one word over the other, but for the sake of consistency it would be nice to pick one and stick to it.

pm/stub.go
```diff
@@ -216,3 +236,24 @@ func (s *stubSigner) Sign(msg []byte) ([]byte, error) {
...
+}
+
+func (m *MockRecipient) RedeemWinningTickets(sessionIDs []string) error {
+	args := m.MethodCalled("RedeemWinningTickets", sessionIDs)
+	return args.Error(0)
+}
+
+func (m *MockRecipient) TicketParams(sender ethcommon.Address) (*TicketParams, error) {
+	args := m.MethodCalled("TicketParams", sender)
+	return args.Get(0).(*TicketParams), args.Error(1)
```
yondonfu: Just noting that this type assertion will panic if `args.Get(0)` is ever nil so if we expect to ever want the mock return value to be nil we should first check for nil before doing the type assertion.

pm/recipient_test.go
```diff
@@ -648,6 +651,53 @@ func TestRedeemWinningTickets_MultipleTickets(t *testing.T) {
...
+	sender, b, v, ts, faceValue, winProb, sig := newRecipientFixtureOrFatal(t)
+	secret := [32]byte{3}
+	r := NewRecipientWithSecret(b, v, ts, secret, faceValue, winProb)
+	// Config stub validator with valid winning tickets
+	v.SetIsWinningTicket(true)
+	require := require.New(t)
+
+	params0 := ticketParamsOrFatal(t, r, sender)
+	ticket0 := newTicket(sender, params0, 1)
+	sessionID1, won, err := r.ReceiveTicket(ticket0, sig, params0.Seed)
```
yondonfu: Should this be `sessionID0` to match the naming convention of `params0` and `ticket0`?

pm/recipient_test.go
```diff
@@ -648,6 +651,53 @@ func TestRedeemWinningTickets_MultipleTickets(t *testing.T) {
...
+	require.Nil(err)
+	require.True(won)
+
+	params1 := ticketParamsOrFatal(t, r, sender)
+	ticket1 := newTicket(sender, params1, 1)
+	sessionID2, won, err := r.ReceiveTicket(ticket1, sig, params1.Seed)
```
yondonfu: Should this be `sessionID1` to match the naming convention of `params1` and `ticket1`?

pm/recipient.go
```diff
@@ -116,8 +116,8 @@ func (r *recipient) ReceiveTicket(ticket *Ticket, sig []byte, seed *big.Int) (st
 
 // RedeemWinningTicket redeems all winning tickets with the broker
```
yondonfu, line 117: Update comment to reflect the update from a single sessionID to a list of sessionIDs?

core/orchestrator.go
```diff
@@ -318,6 +318,15 @@ func (n *LivepeerNode) transcodeSegmentLoop(md *SegTranscodingMetadata, segChan
 					delete(n.SegmentChans, md.ManifestID)
 				}
 				n.segmentMutex.Unlock()
+				if n.Recipient != nil {
+					sessionIDs := getPMSessionIDsByManifestID(n, md.ManifestID)
+					if len(sessionIDs) > 0 {
+						err := n.Recipient.RedeemWinningTickets(sessionIDs)
```
yondonfu, line 324: Should we clear the sessionIDs for the manifestID associated with this loop after we finish redeeming?

core/livepeernode.go
```diff
@@ -51,33 +53,38 @@ type LivepeerNode struct {
 
 	// Transcoder public fields
 	SegmentChans     map[ManifestID]SegmentChan
+	Recipient        pm.Recipient
+	PMSessions       map[ManifestID]map[string]bool
```
yondonfu: Why not use `map[ManifestID][]string`?

yondonfu: Feels like this data structure is just used for internal state management - should this be private?

core/core_test.go
```diff
@@ -75,6 +82,190 @@ func TestTranscode(t *testing.T) {
...
+	segChan := getSegChan(n, md.ManifestID)
+	require.NotNil(segChan)
+
+	waitForTranscoderLoopTimeout(n, md.ManifestID)
+
+	segChan = getSegChan(n, md.ManifestID)
+	assert.Nil(segChan)
+}
+
+func TestTranscodeLoop_GivenOnePMSession_RedeemsOneSession(t *testing.T) {
```
yondonfu: Should we include `_GivenNoSegmentsPastTimeout_` to the test function name?

Same question applies to the rest of the test function names.

core/core_test.go
```diff
@@ -75,6 +82,190 @@ func TestTranscode(t *testing.T) {
...
+	n.pmSessionsMutex.Unlock()
+	recipient.On("RedeemWinningTickets", mock.Anything).Return(fmt.Errorf("some error"))
+
+	errorLogsBefore := glog.Stats.Error.Lines()
+	_, err := n.sendToTranscodeLoop(md, ss)
+	require.Nil(err)
+	waitForTranscoderLoopTimeout(n, md.ManifestID)
+	errorLogsAfter := glog.Stats.Error.Lines()
+
+	assert.Equal(t, int64(1), errorLogsAfter-errorLogsBefore)
```
yondonfu, line 285: This test seems fragile - if someone ever adds an additional log to `sendToTranscodeLoop` this test would fail. Another question: is it worthwhile to test the presence of log messages in this manner?

core/core_test.go
```diff
@@ -75,6 +82,190 @@ func TestTranscode(t *testing.T) {
...
+}
+
+func waitForTranscoderLoopTimeout(n *LivepeerNode, m ManifestID) {
+	for i := 0; i < 3; i++ {
+		time.Sleep(transcodeLoopTimeout * 2)
+		segChan := getSegChan(n, m)
+		if segChan == nil {
```
yondonfu, line 326: Should this function *just* wait for the duration of `transcodeLoopTimeout`? At the moment, this function also defines an additional break condition for when the transcode loop is cleaned up (i.e. `segChan == nil`). I feel like this function shouldn't have to know anything about that condition and instead the test should wait for the duration of `transcodeLoopTimeout` and then confirm that the clean up happened. In that case, the rest of this function could be omitted and we could just use `time.Sleep(transcodeLoopTimeout)`

cmd/livepeer/livepeer.go
```diff
@@ -302,6 +306,16 @@ func main() {
 				glog.Errorf("Error setting up orchestrator: %v", err)
 				return
 			}
+
+			sigVerifier := pm.NewApprovedSigVerifier(n.Eth)
```
yondonfu: I don't think we want to use `ApprovedSigVerifier` right now since it requires ETH calls to check for approved signers and instead we want to use an implementation of `SigVerifier` that only does client side signature verification - there is already a helper function here https://github.com/livepeer/go-livepeer/blob/master/eth/helpers.go#L18

eth/helpers_test.go
```diff
@@ -0,0 +1,39 @@
...
+func TestFromPerc_DefaultDenominator(t *testing.T) {
+	assert.Equal(t, big.NewInt(1000000), FromPerc(100.0))
+
+	assert.Equal(t, big.NewInt(500000), FromPerc(50.0))
+
+	assert.Equal(t, big.NewInt(0), FromPerc(0.0))
+}
+
+func TestFromPercOfUint256_Given100Percent_ResultWithinEpsilon(t *testing.T) {
+	maxBigInt := common.MaxUint256OrFatal(t)
```
yondonfu: I think we can use `maxUint256` defined in `helpers.go`

yondonfu (changes requested):

cmd/livepeer/livepeer.go
eladmallel: @yondonfu so what specific type should we provide to `NewValidator`? - do you mean that I should create a new concrete type that implements the `SigVerifier` interface and implements the `Verify` function from the interface the same way `VerifySig` is implemented in the link you attached?

core/core_test.go
eladmallel: I thought so initially, but in practice waiting that long isn't sufficient - the test would pass/fail inconsistently. It even failed for me when I waited `transcodeLoopTimeout * 2`. I prefer this approach to avoid any chance of flaky tests, while paying the cost of some extra coupling between the test and the implementation.

Wdyt?

core/core_test.go
eladmallel: Excellent question. I was thinking what's the best way to handle a failure in redeeming winning tickets, and currently logging an error was the best I could come up with (perhaps we want to throw some event to some monitoring solution, but I think that's for a separate discussion about instrumentation).

If we agree that logging upon redemption error is the best course of action, I felt like erring on the side of having a test for it rather than not, because without a test, for all I know we are doing nothing in case of an error.

I agree this test is somewhat fragile, but thought we might want to keep it as long as it's not causing us too much trouble, for the benefit of assurance that we're at least doing something about failed redemptions.

Wdyt?

core/core_test.go
eladmallel: Maybe change the middle piece to say `GivenOnePMSessionAtVideoSessionTimeout` ?

pm/stub.go
eladmallel: fixing that in an upcoming commit

pm/stub.go
eladmallel: I think that perhaps it's ok to not be only using one word for all of our "test doubles"? If you check out [this discussion](https://stackoverflow.com/questions/3459287/whats-the-difference-between-a-mock-stub), they talk about the difference between stub and mock. 

I honestly don't care very deeply about what wording we use, but chose to call this one mock simply because it relies on a mocking library, and does therefore provide the functionality of setting specific usage expectations that stubs are less suited for.

If after you read the SO link I shared you still want this to be named `StubRecipient`, I will gladly change it :)

pm/stub.go
eladmallel: Forgive my ignorance, but where in the documentation does it say that `Called` definitely takes the called function name automatically?

Of course I'm thrilled to simplify this and use `Called` if I know it works as we say it does!

eth/client_ticketbroker.go
eladmallel: For some reason during the coding process the compiler yelled at me about needing to implement this function, but now I tried removing it and it seems ok, so I suppose I misunderstood something..

Removing it.

pm/recipient.go
eladmallel: I believe I did in the next line, no?

core/orchestrator.go
eladmallel: I think you're right. I was having some thoughts around retries in case of an error, but any retry mechanism should probably not rely on this mapping because this ManifestID is no longer relevant anyway.

Will make this change.

core/livepeernode.go
eladmallel: Because when we populate it I wanted to be able to just insert sessionIDs without having to check if they already exist in the string slice.

Wdyt?

eladmallel: I was simply following what seemed like a convention of making all these fields public by default. I think you have a good point - looking into making it private.

pm/stub.go
yondonfu: The documentation doesn't seem to explicitly state it, but I think it does automatically take the called function name based upon the example usage in the package [README](https://github.com/stretchr/testify#mock-package) and the [source code](https://github.com/stretchr/testify/blob/master/mock/mock.go#L317) for `Called` looks like it calls `MethodCalled` internally after it gets the called function name

pm/stub.go
yondonfu: Interesting, I didn't know about that distinction - thanks for the link! I think your naming choice here makes sense then - perhaps we make it a convention that test doubles that use the `testify/mock` package to enable setting of expectations during a test use the "mock" naming convention.

pm/recipient.go
yondonfu: Hm I see the comment is updated for `RedeemWinningTickets` in the interface, but not here (for the implementation)

core/livepeernode.go
yondonfu: Makes sense!

core/livepeernode.go
```diff
@@ -54,7 +54,7 @@ type LivepeerNode struct {
 	// Transcoder public fields
 	SegmentChans     map[ManifestID]SegmentChan
 	Recipient        pm.Recipient
-	PMSessions       map[ManifestID]map[string]bool
+	pmSessions       map[ManifestID]map[string]bool
```
yondonfu: Group with the rest of the private fields of this struct?

core/core_test.go
yondonfu: That makes sense and I think emitting some type of event for a monitoring solution would be a nice future addition with logging being the short term solution.

core/core_test.go
yondonfu: SGTM as long as the test function name conveys that we're specifically testing for the scenario after the transcode loop times out (since we already have a test function named `TestTranscodeLoop_GivenNoSegmentsPastTimeout_CleansSegmentChan` explicitly indicating we're testing after the transcode loop times out)

cmd/livepeer/livepeer.go
yondonfu: Yeah maybe implement a `DefaultSigVerifier` type that just calls `VerifySig`?

core/core_test.go
yondonfu: Would it still be flaky if we waited `transcodeLoopTimeout` multiplied by a larger constant like 5?  I do agree that avoiding flaky tests would be preferable

eladmallel: Even if you wait 5x the wait time, you're still running the risk of the test failing rarely. I think it's best to have a slight fragile test in terms of its assertion, for the benefit of not making our tests significantly longer than they need to be + we get less flakiness. We just need to be ok with the compromise on the assertion.

I still feel like the current approach is the lesser evil.

Wdyt?

cmd/livepeer/livepeer.go
eladmallel: Sure, just pushed up a change with that

pm/stub.go
eladmallel: yeah, that sounds good to me :)

core/core_test.go
eladmallel: Yes! Would love for us to take monitoring to the next level!

yondonfu (approved):
LGTM 🚢

core/core_test.go
j0sh: Later on we might want to consider stubbing out the context creation itself (rather than overriding the `transcodeLoopTimeout`) in order to gain more control of behavior during testing. We can use a plain cancellable context for testing. Something like:

```golang
// default implementation
var transcodeLoopContext = func() (ctx context.Context, cancel context.CancelFunc) {
  return context.WithTimeout(context.Background(), transcodeLoopTimeout) 
}

// during testing

stubCtx, stubCancel = context.WithCancel(context.Background()) 
stubTranscodeLoopContext = func() {
  return stubCtx, stubCancel
}
transcodeLoopContext = stubTranscodeLoopContext
...
sendToTranscodeLoop(md, ss)
stubCancel()
assert(getSegChan(n, m) == nil)
```

This would give us the opportunity to test other "hard-to-reach" behaviors such as ensuring the `cancel` function is called after each segment.

This mitigates the timing sensitivity from most places in the testing code; about the only place we'd need to be timing-sensitive is when testing the default `transcodeLoopContext` function. This pattern isn't used anywhere else in the codebase yet, but there are a few other places where it'd be useful as well, eg during tests for O discovery.

core/livepeernode.go
```diff
@@ -51,33 +53,38 @@ type LivepeerNode struct {
...
 	OrchSecret       string
 	Transcoder       Transcoder
 
 	// Transcoder private fields
-	segmentMutex *sync.Mutex
-	tcoderMutex  *sync.RWMutex
-	taskMutex    *sync.RWMutex
-	taskChans    map[int64]TranscoderChan
-	taskCount    int64
+	pmSessions      map[ManifestID]map[string]bool
```
j0sh, line 64: The full use of `pmSessions` seems to still be a WIP but is there a reason we can't store the sessions locally within the transcode loop rather than keeping it as essentially global state here? The transcode loop is single-threaded which would mitigate the need for locking and auxiliary helpers such as `getAndClearPMSessionIDsByManifestID`. The sessions list could go into the `transcodeConfig` (perhaps with an appropriate rename), or a new struct for mutable state within the transcode loop.

eladmallel: It's definitely the initial intent I had, and I'm definitely open to this refactor. One of the challenges I noticed is in the process writing pm sessionIDs to this memory cache. That moment needs to be when we receive a winning ticket - which we will probably unmarshal and process in `ServeSegment`. To store the sessionID then in the transcode loop, I believe we would need to pass the sessionID somehow through some function calls, and eventually add it to the struct we send over the segment channel, in order to get this data point all the way through to `transcodeSegmentLoop`. From a quick look at `SegChanData` it doesn't look like the best place to add payment-related information. Therefore I'm still wondering whether it makes sense to keep this as a global variable such that we can write to it from `ServeSegment` whenever we receive a winning ticket, and then read those sessionIDs in the transcode loop when it's time to redeem.

Would love hear more on how you think we should overcome some of the challenges I mentioned above, and any other challenges you are thinking of!

core/core_test.go
eladmallel: I love the idea! I'll add a note to introduce this refactor, ideally in the time between MVS and SS.

eladmallel: @j0sh added an issue: https://github.com/livepeer/go-livepeer/issues/660

PR #658, opened 2019-01-03
Broadcaster PM CLI Flow

### LINKED ISSUE #631: PM Broadcaster CLI Flows
B needs to be able to perform the following operations:

- Fund deposit and penalty escrow
- Withdraw deposit and penalty escrow
- Delegate signing permissions to other ETH accounts

B's admin should be able to perform these operations from the CLI.

- Add the above functionality to the CLI wizard
- Expose this functionality to the CLI from the node's HTTP interface

Nothing to add.

**Additional context**
Nothing to add.

yondonfu:

Enables the broadcaster PM CLI flow.

- HTTP handlers for depositing, unlocking, cancelling unlocks, withdrawing, querying sender info and querying TicketBroker params
- Attached HTTP handlers to webserver endpoints
- Added support for hitting the HTTP endpoints in the CLI

- Wrote unit tests for the HTTP handlers
- Tested the broadcaster CLI actions (and their ETH interactions) within devenv (on the [pm branch](https://github.com/livepeer/devenv/tree/pm))

Fixes #631 

This PR does not implement signer approval which will not be included in MVS. Furthermore, this PR seeks to provide the most basic CLI functionality and we can certainly improve on the UX going forward.

**Checklist:**


cmd/livepeer_cli/wizard_ticketbroker.go
```diff
@@ -0,0 +1,217 @@
...
+	}
+	fmt.Println(httpPostWithParams(fmt.Sprintf("http://%v:%v/fundAndApproveSigners", w.host, w.httpPort), form))
+}
+
+func (w *wizard) fundDeposit(currDeposit *big.Int) {
+	fmt.Printf("Current Deposit: %v\n", currDeposit)
+
+	fmt.Printf("Enter deposit amount in ETH - ")
+
+	amount := w.readFloat()
```
eladmallel: I think this means we're allowing users to input a negative number?

cmd/livepeer_cli/wizard_ticketbroker.go
```diff
@@ -0,0 +1,217 @@
...
+
+func (w *wizard) unlock() {
+	sender, err := w.senderInfo()
+	if err != nil {
+		glog.Errorf("Error getting sender info: %v", err)
+		return
+	}
+
+	if sender.WithdrawBlock.Cmp(big.NewInt(0)) > 0 {
+		fmt.Println("Already in the unlock period")
```
eladmallel: I think this might be slightly confusing, since the state of `withdrawBlock > 0` can mean two things:

1. If you still have funds in your account, you're in the "unlocking" state
2. If you have no funds in your account, it's a different state we could name "empty"

For case 2 I think it's helpful to print something else to the user, maybe something like "Funds already withdrawn, the account is empty".

wdyt?

cmd/livepeer_cli/wizard_ticketbroker.go
```diff
@@ -0,0 +1,217 @@
...
+	}
+
+	fmt.Printf("Current Deposit: %v\n", eth.FormatUnits(sender.Deposit, "ETH"))
+	fmt.Printf("Current Penalty Escrow: %v\n", eth.FormatUnits(sender.PenaltyEscrow, "ETH"))
+
+	fmt.Printf("Would you like initiate the unlock period? (y/n) - ")
+
+	input := w.readStringAndValidate(func(in string) (string, error) {
+		if in != "y" && in != "n" {
+			return "", errors.New("Enter (y)es or (n)o")
```
eladmallel: I feel like this is slightly confusing.. perhaps we can change the instructions or the values we accept to have a more straightforward relationship between instruction and parsing?

for example we could ask the user only for `y / n` in the print out OR we could accept "yes" / "no" strings.

wdyt?

eladmallel: another note - I think it's a worthwhile refactor to just extract the multiple copies of the y/n validation implementation to a single func and just provide a pointer to that func here, to promote reuse.

wdyt?

cmd/livepeer_cli/wizard_ticketbroker.go
```diff
@@ -0,0 +1,217 @@
...
+}
+
+func (w *wizard) withdraw() {
+	sender, err := w.senderInfo()
+	if err != nil {
+		glog.Errorf("Error getting sender info: %v", err)
+		return
+	}
+
+	if sender.WithdrawBlock.Cmp(big.NewInt(0)) <= 0 {
```
eladmallel: Just to make sure - are we making an explicit decision to not defend users from trying to withdraw before the withdraw block has passed? IIUC this code will result in gas spend even before that block has passed.

wdyt?

eth/stubclient.go
```diff
@@ -12,8 +12,110 @@ import (
...
+	arg := args.Get(idx)
+
+	if arg == nil {
+		return nil
+	}
+
+	return arg.(*big.Int)
+}
+
+type MockClient struct {
```
eladmallel, line 38: 🎉

server/handlers.go
```diff
@@ -0,0 +1,283 @@
...
+	"net/http"
+
+	ethcommon "github.com/ethereum/go-ethereum/common"
+	"github.com/golang/glog"
+	"github.com/livepeer/go-livepeer/common"
+	"github.com/livepeer/go-livepeer/eth"
+)
+
+func logAndRespondWithError(w http.ResponseWriter, errMsg string, code int) {
+	glog.Error(errMsg)
```
eladmallel: Could be cool to include the HTTP code along with the msg, wdyt?

server/handlers.go
```diff
@@ -0,0 +1,283 @@
...
+	http.Error(w, errMsg, code)
+}
+
+func mustHaveFormParams(h http.Handler, params ...string) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if err := r.ParseForm(); err != nil {
+			glog.Error(err)
```
eladmallel: redundant since we're logging in the next line, right?

eladmallel: oh, I see we're logging a different error here. why not wrap this error with our string and log just once with both data points?

eladmallel: I notice this pattern below as well, so let's consider this thread relevant for all instances :)

eladmallel:
I'm generally wondering what are the pros and cons of coding the CLI user input validations in the CLI wizard vs in the HTTP handlers. @yondonfu what's your take on this?

.gitignore
```diff
@@ -18,4 +18,4 @@ eth/protocol
 
 # Our compiled binary
 livepeer
-livepeer_cli
+/livepeer_cli
```
j0sh, line 21: What does this do?

server/handlers.go
yondonfu: [http.Error](https://golang.org/pkg/net/http/#Error) on the next line accepts `code` as a parameter and the handlers should specify a status code using one of the [constants](https://golang.org/pkg/net/http/#pkg-constants) defined in the `http` package.

server/handlers.go
yondonfu: I wasn't sure if we wanted to return the same exact logged error message as a part of the HTTP response. However, after thinking about it, if an adversary is already able to send requests to these HTTP endpoints (i.e. via the CLI, curl, etc.) then returning informative HTTP errors is the least of the worries here...I think these HTTP endpoints are either secured against external parties by only accepting requests via localhost or they are not so, I think it makes sense to just wrap these errors so that they are returned in the HTTP response

cmd/livepeer_cli/wizard_ticketbroker.go
yondonfu: Good point - I'll make the change

cmd/livepeer_cli/wizard_ticketbroker.go
yondonfu: > for example we could ask the user only for y / n in the print out OR we could accept "yes" / "no" strings.

For the error message, I was trying to be super obvious by indicating that y = yes and n = no. We could also just say `Enter y or n`

cmd/livepeer_cli/wizard_ticketbroker.go
yondonfu: Good point - I'll make an update to only accept positive numbers

server/handlers.go
```diff
@@ -0,0 +1,283 @@
...
+			logAndRespondWithError(w, "parse form error", http.StatusInternalServerError)
```
j0sh: Perhaps we could have some more concise helpers, eg

```
func respondWith500(w, err) {
  logAndRespondWithError(w, err, http.StatusInternalServerError)
}
func respondWith400(w, err) {
  logAndRespondWithError(w, err, http.StatusBadRequest)
}
```
(The "logAnd" prefix doesn't really seem necessary for the helpers, but that's just my personal preference.)

j0sh:
Sharing my preliminary experience testing this PR with a protocol deployed from the devenv `pm` branch.

1. Setting up the broadcaster (via devenv `lpdev`) works fine. However, setting the transcoder failed.

2. Attempted to register a transcoder manually via `livepeer_cli` and this branch.

3. Would not get past the `bond` tx; consistently received the error on submission "Transaction Failed: failed to estimate gas needed: gas required exceeds allowance or always failing transaction".

4. Set the `-gasLimit=8000000` on the node and the transaction is submitted. However, it quickly fails with this log on the goclient: `webserver.go:284] tx 0x8d024a1d50a31f90ed9720a2336994de760f7a83cdb0f9cafd8356a9876bcb14 failed` The geth log does not show any errors:

```INFO [01-04|01:13:37] Commit new mining work                   number=59817 txs=0 uncles=0 elapsed=2.118ms
INFO [01-04|01:13:37] Submitted transaction                    fullhash=0x8d024a1d50a31f90ed9720a2336994de760f7a83cdb0f9cafd8356a9876bcb14 recipient=0x475ed5f14B09A3E63ED811C7862b110d4fA56131
INFO [01-04|01:13:38] Successfully sealed new block            number=59817 hash=243224…a2d118
INFO [01-04|01:13:38] 🔗 block reached canonical chain          number=59812 hash=337be8…50c79a
```

5. The round is then initialized, but the information still does not appear on the blockchain during subsequent requests.

6. Happens consistently, even with varying settings (eg, gas price 100).

Haven't diagnosed further beyond this point yet.

server/handlers_test.go
```diff
@@ -0,0 +1,644 @@
...
+
+	form := url.Values{
+		"amount": {"100"},
+	}
+	resp := httpResp(handler, "POST", strings.NewReader(form.Encode()))
+	body, _ := ioutil.ReadAll(resp.Body)
+
+	assert := assert.New(t)
+	assert.Equal(http.StatusBadRequest, resp.StatusCode)
+	assert.Equal("amount is not sufficient for minimum penalty escrow", strings.TrimSpace(string(body)))
```
eladmallel: Perhaps it's valuable to include the minimum amount from the contract in this response (for better UX)?

server/handlers_test.go
```diff
@@ -0,0 +1,644 @@
...
+	}
+	resp := httpResp(handler, "POST", strings.NewReader(form.Encode()))
+	body, _ := ioutil.ReadAll(resp.Body)
+
+	assert := assert.New(t)
+	assert.Equal(http.StatusInternalServerError, resp.StatusCode)
+	assert.Equal("could not execute fundAndApproveSigners", strings.TrimSpace(string(body)))
+}
+
+func TestFundAndApproveSignersHandler_Success(t *testing.T) {
```
eladmallel, line 227: Love how exhaustive the tests are so far!

server/handlers.go
eladmallel: Yeah but we don't write the status code to the log, right? - I was hoping we could log the status code. wdyt?

server/handlers.go
eladmallel: Yeah good thinking about potential attackers, but I also agree that it's probably wise for node operators to not expose this interface to the internet.

.gitignore
yondonfu: Without the slash, Git was ignoring new files added to `cmd/livepeer_cli`. I believe the slash just tells Git to only ignore files/directories with the name `livepeer_cli` in the base directory.

cmd/livepeer_cli/wizard_ticketbroker.go
yondonfu: Updated to use the [readPositiveFloat()](https://github.com/livepeer/go-livepeer/pull/658/files#diff-5d52c4c551172dd0fedcbb693fdda67eR193) and [readPositiveFloatAndValidate()](https://github.com/livepeer/go-livepeer/pull/658/files#diff-5d52c4c551172dd0fedcbb693fdda67eR203) helpers

cmd/livepeer_cli/wizard_ticketbroker.go
yondonfu: Updated to print out the sender's current status if it is unable to unlock [here](https://github.com/livepeer/go-livepeer/pull/658/files#diff-282e3c4a46af4e072fc859cc195b88eeR127) using this [helper](https://github.com/livepeer/go-livepeer/pull/658/files#diff-282e3c4a46af4e072fc859cc195b88eeR269)

cmd/livepeer_cli/wizard_ticketbroker.go
yondonfu: Updated to use the [readStringYesOrNo()](https://github.com/livepeer/go-livepeer/pull/658/files#diff-5d52c4c551172dd0fedcbb693fdda67eR64) helper which has an error message of "Enter y or n"

cmd/livepeer_cli/wizard_ticketbroker.go
yondonfu: Updated to prevent users from withdrawing if the sender is not unlocked [here](https://github.com/livepeer/go-livepeer/pull/658/files#diff-282e3c4a46af4e072fc859cc195b88eeR203). 

Note: If the node is using gas estimation (instead of selecting fixed gas limit), technically the user wouldn't have wasted gas because gas estimation would fail and the tx would not be submitted.

server/handlers.go
yondonfu: Added the status code [here](https://github.com/livepeer/go-livepeer/pull/658/files#diff-105fedf355a9fcad82982745cef24f7aR24)

server/handlers.go
yondonfu: Updated the handlers to just wrap any errors when creating the HTTP error response

server/handlers.go
yondonfu: Updated [here](https://github.com/livepeer/go-livepeer/pull/658/files#diff-105fedf355a9fcad82982745cef24f7aR15)

server/handlers_test.go
yondonfu: In the latest updates, the server no longer checks if the amount is sufficient for the minimum penalty escrow - the wizard client does these checks instead when prompting the user for an amount to deposit

yondonfu:
> I'm generally wondering what are the pros and cons of coding the CLI user input validations in the CLI wizard vs in the HTTP handlers.

In the latest updates, I removed the bits of input validation in the HTTP handlers and instead leaned into doing all the input validations in the CLI wizard. I chose to do this for now because the CLI wizard already has to guide the user to provide the correct inputs and it felt logical to be able to see all the validation in a single place with the server endpoints just being used to control the node. Don't feel that strongly about this though, so open to hearing other thoughts. If the HTTP handlers implemented all the validation, then we would just need to make sure to return informative error messages that the CLI wizard can use to guide user input.

yondonfu:
> Sharing my preliminary experience testing this PR with a protocol deployed from the devenv pm branch.

I didn't have any trouble using the devenv wizard when registering an orchestrator. Just to eliminate the obvious, if you were manually registering using `livepeer_cli` did you make sure to request LPT first?

Related to manually testing this PR with devenv, here is a check list of things worth trying out listed by the CLI option number used to take an action:

First deposit means this is the first time B is making a deposit and its deposit/penaltyEscrow are both currently 0.

- 12 (first deposit): rejects negative input
- 12 (first deposit): rejects input greater than current ETH balance
- 12 (first deposit): rejects input less than minimum penalty escrow
- 12 (first deposit): upon valid input, should set depositAmount = input - minPenaltyEscrow and penaltyEscrowAmount = minPenaltyEscrow. Should see parameters in node logs and see the correct values under “Broadcaster Stats"
- 12 (not first deposit): rejects negative input
- 12 (not first deposit): rejects input greater than current ETH balance
- 12 (not first deposit): Should see correct updated deposit under “Broadcaster Stats"
- 13 (locked): Should see updated withdraw block under “Broadcaster Stats"
- 13 (when not locked): reject
- 14 (when not in unlock period): reject
- 14 (when in unlock period): Should not see withdraw block under “Broadcaster Stats"
- 15 (when not unlocked): reject
- 15 (when unlocked): Should see updated deposit and penalty escrow under “Broadcaster Stats” and ETH balance

j0sh:
> if you were manually registering using livepeer_cli did you make sure to request LPT first?

Yep, the account already had LPT in it as indicated by `livepeer_cli`, and was in the unbonded state.

> here is a check list of things worth trying out listed by the CLI option number used to take an action:

Thanks for this!

eladmallel:
> In the latest updates, I removed the bits of input validation in the HTTP handlers and instead leaned into doing all the input validations in the CLI wizard. I chose to do this for now because the CLI wizard already has to guide the user to provide the correct inputs and it felt logical to be able to see all the validation in a single place with the server endpoints just being used to control the node. Don't feel that strongly about this though, so open to hearing other thoughts. If the HTTP handlers implemented all the validation, then we would just need to make sure to return informative error messages that the CLI wizard can use to guide user input.

I don't have a very strong opinion here.. my natural inclination was to actually put all the logic in the HTTP server side, because of best practices I'm used to from building classic client-server models, where the client can be a web/iOS/Android app, and you want to minimize the amount of data validation code you must write, etc.

If for example we decide to have a web interface that speaks to this same HTTP server for admin control, I think we would prefer having as much validation as possible in a shared implementation, rather than having the web client copy the CLI behavior.

wdyt?

eladmallel (approved):
Aside from the more strategic discussion of where validations should be, I think this PR is good to go!

j0sh:
Agreed with Elad. Typically, client-side validation is done for UX, but the server should always be checking the data as well ("never trust your input"). That shouldn't be a blocker here though, this PR looks good!

yondonfu:
@j0sh @eladmallel I agree with both your comments! We can take a pass to update the HTTP handler validation logic during our refactoring period post-MVS (not just the handlers included in this PR, but all of our HTTP handlers). 

EDIT: Tracking in #665

Issue #633, opened 2018-12-14
Use PM Sender in RPC


eladmallel:
**Is your feature request related to a problem? Please describe.**
This issue assumes we've implemented `Sender`'s functionality, but haven't yet connected it to the video transcoding flow.

For this issue we want to incorporate `Sender` into the video flow.

**Describe the solution you'd like**
- Create a `Sender` per broadcasting session with the appropriate params.
- Use it to construct PM tickets.
- Attach a ticket to each outgoing segment.

_Note:_ this issue does not handle any failure conditions B might have relating to PM, such as low deposit, etc.

**Describe alternatives you've considered**
Nothing to add.

**Additional context**
Nothing to add.

j0sh:
`server/rpc.BroadcastSession` would be a good place to put the `Sender`

PR #659, opened 2019-01-03
net: add TicketParams, Ticket and Payment messages

yondonfu:

Add the TicketParams, Ticket and Payment messages and their Go code generated by the protocol buffer compiler.

- Add TicketParams, Ticket and Payment messages to `net/lp_rpc.proto`
- Nest a TicketParams field in the existing OrchestratorInfo message
- Generated the Go code for the updated messages using the protocol buffer with the updates in `net/lp_rpc.pb.go`

N/A

N/A

This PR establishes the common messages that are required for both #635 and #633 

**Checklist:**


yondonfu:
As a PM sender, B will need to know O's ETH address in order to construct tickets. At the moment, it looks like the on-chain orchestrator pool constructed during discovery does not store the ETH address of each O (even though the ETH addresses are returned from the contract call): https://github.com/livepeer/go-livepeer/blob/d9bae50378e6136212eff29bb416a2d0eea01d74/discovery/discovery.go#L53

There are a few options here:

1. The on-chain orchestrator pool constructed during discovery could store ETH addresses and the struct included in the array returned by [GetOrchestrators()](https://github.com/livepeer/go-livepeer/blob/d9bae50378e6136212eff29bb416a2d0eea01d74/discovery/discovery.go#L67) could be modified to include each O's ETH address (i.e. create a Go struct that has a superset of the fields of `net.OrchestratorInfo` which includes the O's ETH address)
2. Update the `OrchestratorInfo` message to include O's ETH address.  If we add O designed signers post-MVS (i.e. O registers a set of signing addresses on-chain that can sign transcoded results - signatures from any of these addresses could be used by B for verification challenges), then it might make sense to include an ETH address in the `OrchestratorInfo` message so B can accept signatures from the address after checking that the address is an approved signer for O.

Option 2 feels more flexible.

@j0sh @angyangie @eladmallel WDYT?

j0sh:
We'll probably end up with a combination of both approaches at some point. Regarding Option 1, the use of the raw  `net.OrchestratorInfo` struct isn't really meant to be permanent, but it gets things working with the least amount of fuss. In particular, we'll probably want to convert on-the-wire fields into more appropriate types (eg, bytes -> addresses). This follows the pattern outlined here https://github.com/livepeer/go-livepeer/pull/655/files#r244896355

Option 2 is probably the quickest approach for now (in addition to having the mentioned flexibility). The `TicketPrams` struct seems the most appropriate place to hold the ETH address, rather than directly within `OrchestratorInfo` itself.

net/lp_rpc.proto
```diff
@@ -154,3 +157,56 @@ message NotifySegment {
...
+    bytes seed = 4;
+}
+
+// Probabilistic micropayment ticket
+message Ticket {
+    // ETH address of the recipient
+    string recipient = 1;
+
+    // ETH address of the sender
+    string sender = 2;
```
j0sh: Curious, why `string` and not `bytes` for ETH addresses?  Note that we already use `bytes` within the `OrchestratorRequest` struct itself.

https://github.com/livepeer/go-livepeer/blob/59dd08af477d95bce7869a4d8636759038f86e97/net/lp_rpc.proto#L29-L32

net/lp_rpc.proto
```diff
@@ -154,3 +157,56 @@ message NotifySegment {
...
+
+    // Pay out to the recipient if the ticket wins
+    bytes face_value = 3;
```
j0sh, line 190: In wei, right? Would a uint64 [1] work here to avoid a post-processing step, and carrying around a bignum? I guess this technically limits the recipient to ~18 ETH per winning ticket [2] but that seems OK.

[1] https://developers.google.com/protocol-buffers/docs/proto3#scalar
[2] 2^(64-lg(1e18))

eladmallel: IMO it seems more cost-effective to pay the price of additional processing (converting `[]byte` to `*big.Int`) for the benefit of avoiding any bugs with people trying to use higher values of `face_value`. It seems to me like an awkward limitation to introduce by design. I believe the performance cost should be negligible.

wdyt?

eladmallel:
> As a PM sender, B will need to know O's ETH address in order to construct tickets. At the moment, it looks like the on-chain orchestrator pool constructed during discovery does not store the ETH address of each O (even though the ETH addresses are returned from the contract call):
> 
> [go-livepeer/discovery/discovery.go](https://github.com/livepeer/go-livepeer/blob/d9bae50378e6136212eff29bb416a2d0eea01d74/discovery/discovery.go#L53)
> 
> Line 53 in [d9bae50](/livepeer/go-livepeer/commit/d9bae50378e6136212eff29bb416a2d0eea01d74)
> 
>  orchestrators, err := node.Eth.RegisteredTranscoders() 
> There are a few options here:
> 
> 1. The on-chain orchestrator pool constructed during discovery could store ETH addresses and the struct included in the array returned by [GetOrchestrators()](https://github.com/livepeer/go-livepeer/blob/d9bae50378e6136212eff29bb416a2d0eea01d74/discovery/discovery.go#L67) could be modified to include each O's ETH address (i.e. create a Go struct that has a superset of the fields of `net.OrchestratorInfo` which includes the O's ETH address)
> 2. Update the `OrchestratorInfo` message to include O's ETH address.  If we add O designed signers post-MVS (i.e. O registers a set of signing addresses on-chain that can sign transcoded results - signatures from any of these addresses could be used by B for verification challenges), then it might make sense to include an ETH address in the `OrchestratorInfo` message so B can accept signatures from the address after checking that the address is an approved signer for O.
> 
> Option 2 feels more flexible.
> 
> @j0sh @angyangie @eladmallel WDYT?

I like option 2 much more than option 1, for the reasons mentioned already. I think we should do that for MVS, and later consider also adding option 1 for SS.

wdyt?

net/lp_rpc.proto
```diff
@@ -154,3 +157,56 @@ message NotifySegment {
     int64 taskId   = 16;
     bytes profiles = 17;
 }
+
+// Required parameters for probabilistic micropayment tickets
+message TicketParams {
```
eladmallel, line 162: I think @j0sh's note about adding the recipient's ETH address in here makes sense.

@yondonfu wdyt?

net/lp_rpc.proto
yondonfu: For some reason I thought we were using `string` for ETH addresses - must've misread. I'll make the change.

net/lp_rpc.proto
yondonfu: Yeah makes sense to me. Might be worth updating the Go [TicketParams](https://github.com/livepeer/go-livepeer/blob/master/pm/ticket.go#L19) type to include the recipient's ETH address as well for consistency, but we can address that separately from this PR.

yondonfu:
So, if we eventually implement option 1 then B always know's O's recipient address for payments. O's recipient address can always be its registered on-chain address and it probably should be so that the value of winning tickets can be sent to O's fee pool (to later distribute to delegators). Note that anyone can redeem winning tickets and their value will be sent to the ticket's recipient address so this doesn't affect O's ability to have a multi-node setup. There doesn't seem to be a reason for O to provide a recipient address in the `TicketParams` message that is different from its on-chain address. However, I do see an argument for defining these messages without making any assumptions about how B originally fetches information to connect with O (i.e. B already has O's recipient ETH address).

We might still want to add another ETH address field to `TicketParams` in the future so that O can specify the ETH address it will be using to sign transcoded results (distinct from O's recipient ETH address). We can address this when O can actually specify different signing addresses though - for now, B can just use the recipient address included in `TicketParams` for checking signatures over transcoded results.

eladmallel:
> So, if we eventually implement option 1 then B always know's O's recipient address for payments. O's recipient address can always be its registered on-chain address and it probably should be so that the value of winning tickets can be sent to O's fee pool (to later distribute to delegators). Note that anyone can redeem winning tickets and their value will be sent to the ticket's recipient address so this doesn't affect O's ability to have a multi-node setup. There doesn't seem to be a reason for O to provide a recipient address in the `TicketParams` message that is different from its on-chain address. However, I do see an argument for defining these messages without making any assumptions about how B originally fetches information to connect with O (i.e. B already has O's recipient ETH address).
> 
> We might still want to add another ETH address field to `TicketParams` in the future so that O can specify the ETH address it will be using to sign transcoded results (distinct from O's recipient ETH address). We can address this when O can actually specify different signing addresses though - for now, B can just use the recipient address included in `TicketParams` for checking signatures over transcoded results.

Yeah, I really like the idea of implementing PM interactions such that they make as few assumptions as possible on what happens before/after the specific PM interaction. But I also see your point on how it should probably always be the same address registered on-chain.

So @yondonfu, just to confirm, how do you see us moving forward? Do we start with option 2 and later do 1, or are you considering a different approach?

yondonfu:
I’m in favor of starting with option 2 (made an update to reflect that)

eladmallel (approved):
🚢

net/lp_rpc.proto
j0sh: Yeah, I also find it slightly annoying to carry around a `*big.Int` everywhere to indicate the price, but it's not that big of a deal.

Also wonder if we should still have reasonable limits for safety, especially since we're dealing in wei -- I can see someone accidentally tacking on a couple too many zeros and having an "oh crap" moment when the ticket is submitted. But that's probably besides this particular discussion (and still somewhat unlikely since O sets the price; B would simply not use an O with a too-high price and O doesn't have visibility into B's preferences).

yondonfu: > Yeah, I also find it slightly annoying to carry around a *big.Int everywhere to indicate the price, but it's not that big of a deal.

Yeah I agree that moving around a `big.Int` is not ideal, but at the moment it feels simpler/more flexible relative to using a type that places a limitation on what values can be used.

> Also wonder if we should still have reasonable limits for safety, especially since we're dealing in wei -- I can see someone accidentally tacking on a couple too many zeros and having an "oh crap" moment when the ticket is submitted.

From an O operator's point of view we could support inputting prices in gwei or whatever unit as long as it is easier to input than wei. Then, under the hood, we can convert the user inputted value into wei.

j0sh (approved):

PR #662, opened 2019-01-04
PM sender RPC handling

### LINKED ISSUE #633: Use PM Sender in RPC

yondonfu:

Implements PM sender RPC message handling for a broadcaster.

- Add a PM sender to the LivepeerNode
- Use TicketParams included in OrchestratorInfo messages to start local PM sessions
- Create Ticket payments which are attached to the header of segment messages sent to orchestrators

I wrote additional unit tests.

Fixes #633 

**Checklist:**

### DIFF
--- a/server/mediaserver.go
+++ b/server/mediaserver.go
@@ -202,13 +202,14 @@ func (s *LivepeerServer) startBroadcast(cpl core.PlaylistManager) (*BroadcastSes
 	if s.LivepeerNode.Sender != nil {
 		protoParams := tinfo.TicketParams
 		params := pm.TicketParams{
+			Recipient:         ethcommon.BytesToAddress(protoParams.Recipient),
 			FaceValue:         new(big.Int).SetBytes(protoParams.FaceValue),
 			WinProb:           new(big.Int).SetBytes(protoParams.WinProb),
 			RecipientRandHash: ethcommon.BytesToHash(protoParams.RecipientRandHash),
 			Seed:              new(big.Int).SetBytes(protoParams.Seed),
 		}
 
-		sessionID = s.LivepeerNode.Sender.StartSession(ethcommon.BytesToAddress(protoParams.Recipient), params)
+		sessionID = s.LivepeerNode.Sender.StartSession(params)
 	}
 
 	// set OSes
--- a/server/mediaserver.go
+++ b/server/mediaserver.go
@@ -17,10 +17,12 @@ import (
 	"sync"
 	"time"
 
+	ethcommon "github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/crypto"
 	"github.com/livepeer/go-livepeer/drivers"
 	"github.com/livepeer/go-livepeer/monitor"
 	"github.com/livepeer/go-livepeer/net"
+	"github.com/livepeer/go-livepeer/pm"
 
 	"github.com/cenkalti/backoff"
 	"github.com/ericxtang/m3u8"
@@ -195,6 +197,20 @@ func (s *LivepeerServer) startBroadcast(cpl core.PlaylistManager) (*BroadcastSes
 	}
 	tinfo := tinfos[0]
 
+	var sessionID string
+
+	if s.LivepeerNode.Sender != nil {
+		protoParams := tinfo.TicketParams
+		params := pm.TicketParams{
+			FaceValue:         new(big.Int).SetBytes(protoParams.FaceValue),
+			WinProb:           new(big.Int).SetBytes(protoParams.WinProb),
+			RecipientRandHash: ethcommon.BytesToHash(protoParams.RecipientRandHash),
+			Seed:              new(big.Int).SetBytes(protoParams.Seed),
+		}
+
+		sessionID = s.LivepeerNode.Sender.StartSession(ethcommon.BytesToAddress(protoParams.Recipient), params)
+	}
+
 	// set OSes
 	var orchOS drivers.OSSession
 	if len(tinfo.Storage) > 0 {
@@ -208,6 +224,8 @@ func (s *LivepeerServer) startBroadcast(cpl core.PlaylistManager) (*BroadcastSes
 		OrchestratorInfo: tinfo,
 		OrchestratorOS:   orchOS,
 		BroadcasterOS:    cpl.GetOSSession(),
+		Sender:           s.LivepeerNode.Sender,
+		PMSessionID:      sessionID,
 	}, nil
 }
 
--- a/server/rpc.go
+++ b/server/rpc.go
@@ -22,6 +22,7 @@ import (
 	"github.com/livepeer/go-livepeer/drivers"
 	"github.com/livepeer/go-livepeer/monitor"
 	"github.com/livepeer/go-livepeer/net"
+	"github.com/livepeer/go-livepeer/pm"
 	ffmpeg "github.com/livepeer/lpms/ffmpeg"
 	"github.com/livepeer/lpms/stream"
 
@@ -77,6 +78,8 @@ type BroadcastSession struct {
 	OrchestratorInfo *net.OrchestratorInfo
 	OrchestratorOS   drivers.OSSession
 	BroadcasterOS    drivers.OSSession
+	Sender           pm.Sender
+	PMSessionID      string
 }
 
 func genOrchestratorReq(b Broadcaster) (*net.OrchestratorRequest, error) {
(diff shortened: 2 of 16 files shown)


yondonfu:
Ready for review

server/mediaserver.go
```diff
@@ -195,6 +197,20 @@ func (s *LivepeerServer) startBroadcast(cpl core.PlaylistManager) (*BroadcastSes
 	}
 	tinfo := tinfos[0]
 
+	var sessionID string
+
+	if s.LivepeerNode.Sender != nil {
+		protoParams := tinfo.TicketParams
```
eladmallel, line 203: what's the meaning behind the name `protoParams`? why not stick with `ticketParams`?

eladmallel: Oh, I see that `proto` is referring to protobuf. What's the value in defining this variable versus just using `tinfo.TicketParams` ?

server/mediaserver.go
```diff
@@ -195,6 +197,20 @@ func (s *LivepeerServer) startBroadcast(cpl core.PlaylistManager) (*BroadcastSes
...
+		params := pm.TicketParams{
+			FaceValue:         new(big.Int).SetBytes(protoParams.FaceValue),
+			WinProb:           new(big.Int).SetBytes(protoParams.WinProb),
+			RecipientRandHash: ethcommon.BytesToHash(protoParams.RecipientRandHash),
+			Seed:              new(big.Int).SetBytes(protoParams.Seed),
+		}
+
+		sessionID = s.LivepeerNode.Sender.StartSession(ethcommon.BytesToAddress(protoParams.Recipient), params)
```
eladmallel: Perhaps it's simpler to just add `Recipient` into the `pm.TicketParams` struct?

server/rpc.go
```diff
@@ -217,6 +220,39 @@ func verifySegCreds(orch Orchestrator, segCreds string) (*core.SegTranscodingMet
...
+	}
+	protoPayment := &net.Payment{
+		Ticket: protoTicket,
+		Sig:    sig,
+		Seed:   seed.Bytes(),
+	}
+
+	data, err := proto.Marshal(protoPayment)
+	if err != nil {
+		glog.Error("Unable to marshal ", err)
```
eladmallel: why do we log this error specifically whereas with other errors we return with no logging?

server/mediaserver.go
```diff
@@ -208,6 +224,8 @@ func (s *LivepeerServer) startBroadcast(cpl core.PlaylistManager) (*BroadcastSes
 		OrchestratorInfo: tinfo,
 		OrchestratorOS:   orchOS,
 		BroadcasterOS:    cpl.GetOSSession(),
+		Sender:           s.LivepeerNode.Sender,
+		PMSessionID:      sessionID,
```
j0sh, line 229: Curious why it's done this way -- is there a reason for returning an ID to the PM session itself, rather than the PM session itself (or an interface thereof with a `CreateTicket` function)? That way the PM session doesn't have to be stored externally, and we only have to add one field to the connection context. Is there a reason to retain the session forever on the sender side?

server/mediaserver.go
yondonfu: I defined the additional variable to avoid having multiple instances of `tinfo.TicketParams.<FieldName>` when converting into a `pm.Ticket` struct. However, I'm thinking of just using a helper here:

```
params := net.FromProtoTicketParams(tinfo.TicketParams)
```

WDYT?

server/rpc.go
yondonfu: I forgot to remove this log - I don't think it is necessary because we wrap the error returned by `genPayment()` and log it [here](https://github.com/livepeer/go-livepeer/pull/662/files#diff-44bb03d888e11d6429223a7b0e55220dR477)

server/mediaserver.go
yondonfu: The intent was to to keep the `pm` package unaware of how it would eventually be used and for as much of the PM related logic to be handled out of the box by the `pm` package. We initially considered storing a `pm.Sender` per `BroadcastSession`, however, we realized we would still eventually need some way to synchronize all these `pm.Sender` instances so that they all have an up to date view of the remaining account deposit so that no `pm.Sender` instance overspends because another instance sent out many winning tickets. So, rather than having the `pm` package user handling deposit view synchronization we can use a single `pm.Sender` instance across each `BroadcastSession`. In this case, we still need a way the single `pm.Sender` instance to keep track of the current `senderNonce` for a set of ticket params, so we use an internal session within the `pm.Sender` that is referenced by its sessionID (which is just `recipientRandHash`).

I realize that the design probably could be simplified since the user of the `pm` package already has a notion of session, but we were interested in keeping the `pm` package as reusable as possible without any assumption that the user already has a notion of a session. The current design of the `pm` package would allow users to use it to send one-to-many out of the box without the user having to implement its own notion of a session. Definitely interested to hear if you have any thoughts on how this design could be improved though!

j0sh: Ah I see. The synchronization could still happen under the hood within `pm` as tickets are created. How will the account synchronization happen? I'm assuming via `TicketBroker` -- which probably needs to be threadsafe anyway.

I'm not sure what's meant by the "user having its own notion of a session". The user needs to store a session ID (or many); it could also store a PM session (or many) that's defined as part of the `pm` package. Storing sessions on the user eliminates that duplicate layer of bookkeeping within the `Sender` and makes it less explicitly stateful. Either way, there is some tracking required from the user.

Also, by storing sessions proper, we don't need to retain session information indefinitely (do we?) as sessions would be GC'd once the connection terminates. Right now, we accumulate sessions in memory without cleaning them up, which is not ideal for a long-running service process. Eventually this will exhaust the available memory.

The one reason I can see the current setup being useful is if the `Sender` needs a way to scan over all the (active) sessions. But I don't really see a clear need for this at the moment.

Perhaps it'd help me to understand how the fully-formed `Sender` interface is intended to work? The comments still indicate there are some TODOs.

To accommodate the current setup, I was thinking something like:

```
interface PMSession {
  CreateTicket(...) // syncs deposit view under the hood via broker?
}
func NewPMSession(broker TicketBroker, params TicketParams) PMSession
...
struct BroadcastSession {
  PMSession session
  ...
}
```

yondonfu: > How will the account synchronization happen?

One possibility is for `Sender` to maintain a running view of the account's deposit (which might be different from the on-chain deposit) if the recipient notifies the sender of winning tickets prior to redemption. This would also be compatible with your suggestion of creating a `PMSession` object per each set of ticket params - the `PMSession` object could hold a reference to the `Sender` object and would query its running view of the account's deposit when creating a ticket.

> I'm not sure what's meant by the "user having its own notion of a session". 

I meant that in our case there is already a `BroadcastSession` object used for streams so one approach would be to eliminate the notion of a PM session and simply add additional fields to the `BroadcastSession` object to track PM related information. However, it would be nice for users of the package that do not have the equivalent of a `BroadcastSession` object to not have to worry about defining their own session objects unless they absolutely need to.

> Also, by storing sessions proper, we don't need to retain session information indefinitely (do we?) as sessions would be GC'd once the connection terminates. Right now, we accumulate sessions in memory without cleaning them up, which is not ideal for a long-running service process. Eventually this will exhaust the available memory.

Great point - I really like this benefit of your suggestion.

> Perhaps it'd help me to understand how the fully-formed Sender interface is intended to work?

The original intent was for the global `Sender` object to handle all the state related to sending tickets including tracking internal sessions and `BroadcastSession` would just use `sessionID` to tell the `Sender` object which session it cares about without it knowing anything about the details of the session. Later on, we could add support for receiving new ticket params mid stream and we could replace the existing sessionID in `BroadcastSession` with the new sessionID when this happens.

Overall, I like your suggestion - what do you think about tackling it as a part of the refactor period post-MVS in the interest of not increasing the scope of this PR beyond its core intent (which is just integrating the PM sender into the node and handling RPC messages)?

server/mediaserver.go
j0sh: The helper is fine, but I wouldn't put it in the `net` package since that's become the de facto place for shared type/interface declarations to avoid import cycles.

My feeling is it's best to define methods closest to where they're used, so putting the helper in `mediaserver.go` (or  somewhere in `server`) is probably better.

j0sh:
> However, it would be nice for users of the package that do not have the equivalent of a BroadcastSession object to not have to worry about defining their own session objects unless they absolutely need to.

I don't think there is a requirement for a user-supplied "broadcast session" . Users still have to track PM sessions regardless. The question here is whether they track session IDs, or a pm-provided `PMSession`.

> what do you think about tackling it as a part of the refactor period post-MVS in the interest of not increasing the scope of this PR

Sure, but if anything, it'd actually reduce the scope of the PR, since there won't be a need for a `sync.Map` or Sender-side session tracking, and one fewer field in the BroadcastSession (just the `PMSession`  rather than both the `Sender` + `SessionID`).

server/mediaserver.go
angyangie: When would a user not have the a `BroadcastSession` (or the equivalent, though not sure what that refers to)?

eladmallel (approved):
LGTM pending merging master into this branch and resolving conflicts!

j0sh:
Just a PSA, a `git rebase master` rather than a `git merge master` will avoid those merge messages.

yondonfu:
> Just a PSA, a git rebase master rather than a git merge master will avoid those merge messages.

Woops...yeah forgot to rebase instead of merge.

> Sure, but if anything, it'd actually reduce the scope of the PR, since there won't be a need for a sync.Map or Sender-side session tracking, and one fewer field in the BroadcastSession (just the PMSession rather than both the Sender + SessionID).

Good points. Though I think there will still be a bit of refactoring and code moved around so my preference is still to do all that outside of this PR if that is alright with you as this feels like a code maintainability issue that doesn't need to block this PR

j0sh:
> Woops...yeah forgot to rebase instead of merge

It's not too late to rebase! :)

> my preference is still to do all that outside of this PR if that is alright with you.

No problem :+1:

j0sh (approved):

server/mediaserver.go
yondonfu: I was referring to the case where the `pm` package is being used outside of the Livepeer project. Though as Josh has mentioned, even if there is an session created within the `pm` package the user would still need to be aware of it somehow whether it be via a sessionID or via the returned session object itself.

PR #664, opened 2019-01-06
PM Recipient Network Integration

### LINKED ISSUE #635: pm: use recipient in RPC

eladmallel:
Makes use of the PM Recipient side in the Livepeer networking, allowing Bs to receive Os' payment parameters, and for O to process incoming PM tickets from B.

- O responds to `GetOrchestratorInfo` with `TicketParams` which B must use to send valid PM tickets
- O processes incoming PM tickets, starting with parsing them from the HTTP header and passing them down to `Recipient.ReceiveTicket`
- Some refactoring in support of this change

- Added unit tests for the new code and no previous tests have regressed

- Fixes https://github.com/livepeer/go-livepeer/issues/635

**Checklist:**


pm/recipient.go
```diff
@@ -24,7 +24,7 @@ type Recipient interface {
 
 	// TicketParams returns the recipient's currently accepted ticket parameters
 	// for a provided sender ETH adddress
-	TicketParams(sender ethcommon.Address) (*TicketParams, error)
```
yondonfu, line 27: 🎉

core/orchestrator.go
```diff
@@ -79,6 +80,40 @@ func (orch *orchestrator) TranscoderResults(tcId int64, res *RemoteTranscoderRes
 	orch.node.transcoderResults(tcId, res)
 }
 
+func (orch *orchestrator) ProcessPayment(payment net.Payment, manifestID ManifestID) error {
```
yondonfu, line 83: I think we'll get a nil pointer exception on L94 if the node is running in off-chain mode since `orch.node.Recipient` will be nil. The rest of the orchestrator methods that depend on the presence of a ETH client first check that both the node and the ETH client are not nil. Perhaps we can add a similar check here:

```
if orch.node == nil || orch.node.Recipient == nil {
    return nil
}
```

core/orchestrator.go
```diff
@@ -79,6 +80,40 @@ func (orch *orchestrator) TranscoderResults(tcId int64, res *RemoteTranscoderRes
...
+	}
+
+	if won {
+		cachePMSessionID(orch.node, manifestID, sessionID)
+	}
+
+	return nil
+}
+
+func (orch *orchestrator) TicketParams(sender ethcommon.Address) *net.TicketParams {
```
yondonfu, line 110: Echoing my comment for `ProcessPayment` - I think we'll get a nil pointer exception on L107 if the node is running in off-chain mode since `orch.node.Recipient` will be nil. The rest of the orchestrator methods that depend on the presence of a ETH client first check that both the node and the ETH client are not nil. Perhaps we can add a similar check here:

```
if orch.node == nil || orch.node.Recipient == nil {
    return nil
}
```

server/rpc_test.go
```diff
@@ -205,3 +222,207 @@ func TestPing(t *testing.T) {
...
+	assert.Equal(expectedParams, oInfo.TicketParams)
+}
+
+type mockOrchestrator struct {
+	mock.Mock
+}
+
+func (o *mockOrchestrator) ServiceURI() *url.URL {
+	args := o.Called()
+	return args.Get(0).(*url.URL)
```
yondonfu: Might be worth being cautious with type asserting on nil here and checking for that in case this is ever used in the future to return a nil value

yondonfu (changes requested):

core/orchestrator.go
eladmallel: Awesome tip, thank you! I had that thought for a moment but then felt like it might be too defensive, forgetting all about off-chain mode! My bad!

core/orchestrator.go
eladmallel: Thank you, fixed!

server/rpc_test.go
eladmallel: haha I was being lazy here thinking if someone ran into that in the future they will notice (tests will panic) and they will fix it :) but since you mind it, I will add the quick fix :)

server/rpc.go
```diff
@@ -242,6 +245,9 @@ func getOrchestrator(context context.Context, orch Orchestrator, req *net.Orches
 	if os != nil && os.IsExternal() {
 		tr.Storage = []*net.OSInfo{os.GetInfo()}
 	}
+
+	tr.TicketParams = orch.TicketParams(addr)
```
j0sh: Since TicketParams is unconditionally assigned, can we add this where `tr` is defined?

server/rpc.go
```diff
@@ -264,6 +284,11 @@ func (h *lphttp) ServeSegment(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	err = processPayment(orch, r.Header.Get(PaymentHeader), segData.ManifestID)
+	if err != nil {
+		glog.Errorf("Error processing payment: %v", err)
```
j0sh, line 288: We continue to process the segment even if the payment fails?

server/rpc_test.go
```diff
@@ -205,3 +222,207 @@ func TestPing(t *testing.T) {
...
+
+	oInfo, err := getOrchestrator(orch, &net.OrchestratorRequest{})
+
+	assert := assert.New(t)
+	assert.Nil(err)
+	assert.Equal(expectedParams, oInfo.TicketParams)
+}
+
+type mockOrchestrator struct {
+	mock.Mock
```
j0sh, line 361: We already have a `stubOrchestrator`, could the `mock.Mock` be worked into that rather than re-defining a separate set of stubs?

pm/helpers.go
```diff
@@ -22,42 +21,18 @@ func VerifySig(addr ethcommon.Address, msg, sig []byte) bool {
...
-	key, err := randBytes(32)
-
-	if err != nil {
-		t.Fatalf("failed generating random hash: %v", err)
-		return ethcommon.Hash{}
-	}
-
-	return ethcommon.BytesToHash(key[:])
+func RandHash() ethcommon.Hash {
+	return ethcommon.BytesToHash(RandBytes(32))
```
j0sh, line 25: Beautiful set of cleanups.

server/rpc.go
eladmallel: Thank you for bringing this up - I wasn't sure what's the best move here, specifically for MVS. Do you think we should just not process the segment if there's an error here?

server/rpc_test.go
eladmallel: I believe it's possible, but since other tests are using `stubOrchestrator` I preferred to avoid that refactor for now, for the sake of getting us faster to the point of integration tests. When do you think is the best time to tackle this?

yondonfu: @j0sh curious to hear your take on this related [discussion](https://github.com/livepeer/go-livepeer/pull/657#discussion_r244420843) about the distinction between stubs and mocks and the possibility of having both.

core/orch_test.go
```diff
@@ -286,65 +285,53 @@ func TestGetSegmentChan(t *testing.T) {
...
-	require.Nil(t, err)
-	defer os.RemoveAll(tmpdir)
+	n, _ := NewLivepeerNode(nil, "", nil)
 	recipient := new(pm.MockRecipient)
 	n.Recipient = recipient
 	orch := NewOrchestrator(n)
 	recipient.On("ReceiveTicket", mock.Anything, mock.Anything, mock.Anything).Return("some sessionID", true, errors.New("mock error"))
 
-	err = orch.ProcessPayment(defaultPayment(t), ManifestID("some manifest"))
+	orch.ProcessPayment(defaultPayment(t), ManifestID("some manifest"))
```
yondonfu: Check error here?

server/rpc.go
j0sh: I have to dig up the discussion(s?), but IIRC we had a few scenarios that would justify a "grace period" where we'd continue execution even if the payment fails, and a few other scenarios that justified a hard stop.

I believe we were able to whittle down the MVS scope such that we could avoid implementing a grace period, leaving us with just hard stops. If that's still true, then I guess the appropriate thing to do here would be to send an error response and exit early.

server/rpc_test.go
j0sh: Naming conventions aside (which I also don't really have a preference for), I was mostly wondering if the Mock could be incorporated into `stubSegment` without affecting existing tests. If that can't be accomplished cleanly, then it makes sense to have separate mocked stubs.

Are we set on using mocks instead of manual stubs? I can see the helpers saving us some work (eg, counting method calls), but the practice of setting expectations in advance prior to execution is fairly different from the post-execution assertions that we use everywhere else. My initial feeling is that I'd prefer to have a uniform way of checking those conditions, but I'm open if the consensus is that the mixed approach saves us time/complexity/etc.

server/rpc.go
yondonfu: I agree with @j0sh point about sending an error response and exiting if the node is in *on-chain* mode. That being said, could there be a problem if we exit early as is when the node is in *off-chain* mode when the broadcaster tries to unmarshal an empty buffer in `processPayment`? If so, then the orchestrator will always exit early. Perhaps there should also be a check in `processPayment` to return nil if the node is in off-chain mode

j0sh: IIRC, the deserialized `payment` that we receive from protobufs will be empty in off-chain mode -- probufs won't return an error on missing fields since all fields are optional in protobufs v3. The `orch.ProcessPayment` method does incorporate an off-chain check, so we should be all set here.

Would be good to test in `-offchain` mode just to be sure, though.

server/rpc_test.go
yondonfu: > I can see the helpers saving us some work (eg, counting method calls)

For this specific example, the `testify/mock` package seems to have a [Times()](https://godoc.org/github.com/stretchr/testify/mock#Call.Times) method that can be used to set the number of times that a mocked method should return a value and then the test would assert that the method was called the expected number of times using [AssertExpectations()](https://godoc.org/github.com/stretchr/testify/mock#Mock.AssertExpectations). I *think* this should do the trick? I haven't tested this in practice yet though.

> My initial feeling is that I'd prefer to have a uniform way of checking those conditions, but I'm open if the consensus is that the mixed approach saves us time/complexity/etc.

I personally have found the ability to set expectations prior to execution to be useful and feels cleaner than defining custom state variables in stubs. For that reason, I feel ok about a mixed approach right now.

j0sh: The ability to count invocations of a method without writing boilerplate for each occurrence is indeed nice, and I'd like it even better if that could be checked after the fact, eg `assert.Equals(1, mockObj.MethodCallCount("foo"))` . If `AssertExpectations` can give us that, all the better.

yondonfu: Ah there is actually an [AssertNumberOfCalls()](https://godoc.org/github.com/stretchr/testify/mock#Mock.AssertNumberOfCalls) method which does the trick too:

```
m.AssertNumberOfCalls(t, "Bar", 3)
```

server/rpc.go
yondonfu: Ah nice. Might be worthwhile to add an extra test with an empty payment string header too

server/rpc_test.go
j0sh: I suppose that works. Would be more consistent to just return the count somehow, which composes better with the existing `assert` functionality -- passing in `t` every time is ugly.

server/rpc.go
eladmallel: Awesome feedback - changed the code accordingly!

server/rpc.go
```diff
@@ -286,6 +286,8 @@ func (h *lphttp) ServeSegment(w http.ResponseWriter, r *http.Request) {
 	err = processPayment(orch, r.Header.Get(PaymentHeader), segData.ManifestID)
 	if err != nil {
 		glog.Errorf("Error processing payment: %v", err)
+		http.Error(w, err.Error(), http.StatusPaymentRequired)
```
yondonfu, line 289: ❤️ that this is an HTTP status code

yondonfu (approved):
LGTM

PR #670, opened 2019-01-08
server: fix O sig check

### LINKED ISSUE #666: Fix O verification of B seg sig
Currently the B address is hardcoded in O. As a result, segment signing fails and we can only really test e2e in offchain mode.

Now that we mostly have end-to-end PM tickets in, we can use the broadcaster address in the ticket to verify the seg sig.

Use the ticket-provided B address to verify the seg sig.

yondonfu:

Use the sender ETH address in PM tickets for verification of a
signature over seg creds.

- Create a separate helper `getPayment()` for parsing the payment header
- Create a separate helper `getPaymentSender()` for parsing the payment sender (handles cases where the payment header is an empty string, the ticket is missing from the payment or the ticket sender is missing - the helper will make sure to return the null address)
- Within `ServeSegment`, parse the payment header first and try getting the broadcaster address to be used in `verifySegCreds()`. When running in off-chain mode, the null address would be used as the broadcaster address
- Remove initialized round check in the mediaserver when starting a stream

- Unit tests
- Manual tests: transcoding with B & O in off-chain mode, transcoding with B & O in on-chain mode

Fixes #666 

**Checklist:**

### DIFF
--- a/server/mediaserver.go
+++ b/server/mediaserver.go
@@ -260,19 +260,6 @@ func gotRTMPStreamHandler(s *LivepeerServer) func(url *url.URL, rtmpStrm stream.
 		var sess *BroadcastSession
 
 		if s.LivepeerNode.Eth != nil {
-
-			//Check if round is initialized
-			initialized, err := s.LivepeerNode.Eth.CurrentRoundInitialized()
-			if err != nil {
-				glog.Errorf("Could not check whether round was initialized: %v", err)
-				return err
-			}
-			if !initialized {
-				glog.Infof("Round was uninitialized. Please try again in a few blocks.")
-				// todo send to metrics ?
-				return ErrRoundInit
-			}
-
 			// TODO: Check broadcaster's deposit with TicketBroker
 		}
 
--- a/server/rpc.go
+++ b/server/rpc.go
@@ -34,8 +34,6 @@ import (
 	"github.com/pkg/errors"
 )
 
-var broadcasterAddress = ethcommon.BytesToAddress([]byte("111 Transcoder Address 1")) // need to remove hard-coded broadcaster address
-
 const HTTPTimeout = 8 * time.Second
 const GRPCTimeout = 8 * time.Second
 const GRPCConnectTimeout = 3 * time.Second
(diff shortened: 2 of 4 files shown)


eladmallel (approved):
🚢

server/mediaserver.go
```diff
@@ -260,19 +260,6 @@ func gotRTMPStreamHandler(s *LivepeerServer) func(url *url.URL, rtmpStrm stream.
 		var sess *BroadcastSession
 
 		if s.LivepeerNode.Eth != nil {
-
-			//Check if round is initialized
-			initialized, err := s.LivepeerNode.Eth.CurrentRoundInitialized()
```
eladmallel, line 265: 👏

server/rpc.go
```diff
@@ -210,7 +208,7 @@ func verifySegCreds(orch Orchestrator, segCreds string) (*core.SegTranscodingMet
 		OS:         os,
 	}
 
-	if !orch.VerifySig(broadcasterAddress, string(md.Flatten()), segData.Sig) {
```
eladmallel, line 213: @yondonfu I couldn't find where in the old code we used `broadcasterAddress` to sign anything, so how did this address come up as valid until now?

yondonfu: It wasn't coming up as valid - an orchestrator would not have been able to transcode in on-chain mode because every segment received from a broadcaster would fail the signature verification up until now. However, in off-chain mode `orch.VerifySig` would always return `true` so an orchestrator could have still transcoded in off-chain mode.

angyangie: When testing in on-chain mode, I just commented out `orch.VerifySig` for the time being, knowing we'd eventually remove the broadcasterAddress stub

server/mediaserver.go
angyangie: Why does round no longer need to be initialized?

angyangie (approved):

server/mediaserver.go
yondonfu: In the V1 protocol, there was a restriction that on-chain jobs could not be created if the current round was not initialized. With the removal of on-chain jobs, there is no longer an initialized round requirement for the broadcaster.

Note: Given our current planned approach of using round initialization to store block hashes which are attached to tickets to prove that a creation round number existed at the time of ticket creation, we will eventually have some dependence on round initialization. However, the current thought is that a broadcaster still does not have to be blocked by round initialization when sending segments to an orchestrator - it can use the block hash of the last initialized round. In this scenario, the broadcaster can always create tickets/segments regardless of whether the current round is initialized. However, the effective validity period of ticket is reduced if the previous round is used instead of the current round. An orchestrator *might* have an incentive to initialize the round in this case so that the broadcaster can use the current round as the creation round to extend the validity period of tickets. We'll ultimately address this in SS, but I think for the time being eliminating this check suffices for MVS.

angyangie: Makes sense, thanks @yondonfu! I see how "extending" the effective validity period by initializing the round would be a strong incentive to do so.

j0sh (approved):
🎉 🎉 🎉

PR #687, opened 2019-01-16
eth: use NonceManager to manage transaction nonces

### LINKED ISSUE #684: Incorrect tx nonce used if previous tx submission failed
**Describe the bug**

If the node fails to submit a tx, the node's next tx will not be processed properly because an invalid nonce is used.

**To Reproduce**
1. Start a livepeer node with the `-transcoder` flag. Make that the `-gasLimit` flag is NOT set.
2. Use the livepeer_cli to try to register on-chain when the round is not initialized.
3. Observe the gas estimation error (due to an always failing tx) in the node's logs
4. Use the livepeer_cli to try to initialize the round.
5. Observe that the livepeer_cli hangs and the initialization tx is never confirmed on-chain.

**Expected behavior**

If the node fails to submit a tx, the node's next tx will be processed properly with a valid nonce.

 - OS: Ubuntu
 - Version: 1b65ae0

**Additional context**

The node currently supports very basic tx nonce management - before the first tx submission it will fetch the account's current pending nonce and afterwords it will increment the nonce value for every tx it submits. This is to allow for concurrent transaction submission. The node will also check that the incremented nonce is not less than the account's current pending nonce. However, if a transaction is not submitted, the node will still use the incremented nonce N + 1 for the next transaction. The next transaction will not be processed since there is now a nonce gap because a transaction with the previous nonce N was never processed.

This is a problem whenever transaction submission fails. Example scenarios include:
- The node is using the gas estimator which will return an error if a transaction will always fail
- The node encounters some network error when submitting the transaction

yondonfu:

Add a global NonceManager to assign ETH transaction nonces. The included implementation also supports multiple ETH addresses - this feature isn't used at the moment, but might be useful in the future if a node submits transactions with multiple ETH addresses.

- Introduce the NonceManager interface in the vendored go-ethereum
package. The interface is used to get and update the next nonce
- Added a NonceManager implementation in the eth package

Wrote unit tests for the NonceManager implementation. Confirmed that transaction submission works in devenv and that a failed transaction submission would not disrupt subsequent transactions.

Fixes #684 

**Checklist:**

### DIFF
--- a/eth/client.go
+++ b/eth/client.go
@@ -190,6 +190,8 @@ func (c *client) SetGasInfo(gasLimit uint64, gasPrice *big.Int) error {
 		return err
 	}
 
+	opts.NonceManager = NewNonceManager(c.backend)
+
 	if err := c.setContracts(opts); err != nil {
 		return err
 	} else {
--- a/eth/client.go
+++ b/eth/client.go
@@ -19,7 +19,6 @@ import (
 	"fmt"
 	"math/big"
 	"strings"
-	"sync"
 	"time"
 
 	"github.com/cenkalti/backoff"
@@ -156,11 +155,8 @@ type client struct {
 	*contracts.LivepeerVerifierSession
 	*contracts.LivepeerTokenFaucetSession
 
-	nonceInitialized bool
-	nextNonce        uint64
-	nonceLock        sync.Mutex
-	gasLimit         uint64
-	gasPrice         *big.Int
+	gasLimit uint64
+	gasPrice *big.Int
 
 	txTimeout time.Duration
 }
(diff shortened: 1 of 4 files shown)


ericxtang (approved):
This makes sense to me for now.  If we ever run into a situation with a lot of concurrent transactions, we can change the "manager-level" lock to a RWLock to increase performance.  But I don't want to do any pre maturely optimization here.

🚢

j0sh (changes requested):
Could the changes to the go-ethereum ABI bindings first be pushed to our [geth branch](https://github.com/livepeer/go-ethereum) and the git short sha referenced in a commit message here?

Preferably the vendored updates would also go in as a separate commit if it can be done without breaking the build, (example -- 5ec79820fba43e817180dc94625295e4c02479a4 and 35bbae8955a59e0328487ceef2a8bb42d1aa82b4) but separating the commit is not as important as having a reference to the sha that's used.

yondonfu:
@j0sh Updated the commit history to include one referencing the livepeer/go-ethereum vendor update and a follow up commit that includes changes in the eth package

j0sh (approved):

j0sh:
Thanks! 🚢

PR #688, opened 2019-01-16
cmd: reject invalid values for -faceValue & -winProb

### LINKED ISSUE #678: -faceValue and -winProb flags accept negative values
**Describe the bug**

The Livepeer node accepts negative values for the -faceValue and -winProb flags (which are both supposed to be type float64). For faceValue, the user provided float64 is converted into a big.Int which is meant to represent an amount of ETH in wei. For winProb, the user provided input float64 is converted into a big.Int which is meant to represent the numerator of a fraction over `2^256 - 1`.

When the user provides negative values for either of these parameters, the values sent to a B in TicketParams are positive (we use [SetBytes()](https://golang.org/pkg/math/big/#Int.SetBytes) to convert the faceValue/winProb in the protobuf into a Go `big.Int` type [here](https://github.com/livepeer/go-livepeer/blob/master/server/mediaserver.go#L192) which interprets the bytes as an unsigned integer). When B sends a ticket created using those TicketParams, O rejects every ticket because the faceValue/winProb is positive while O's internally required faceValue/winProb is negative.

**To Reproduce**
1. Start an O with a negative value for either the -faceValue or -winProb flags
2. Start streaming from a B

**Expected behavior**

O will reject every ticket/segment from B.

 - OS: Ubuntu
 - Browser: N/A
 - Version: db1cbd2

**Additional context**

We should prevent negative user input when configuring an O.

yondonfu:

Checks the -faceValue and -winProb flags when starting a node - if invalid values are passed in, we log an error message and stop the node.

- Add error message when -faceValue or -winProb are given invalid values (values > 100 are also treated as invalid for -winProb)
- If one of the values for these flags is invalid, the node stops.

Confirmed in devenv that passing in invalid values for the flags will result in logged error messages and the node stopping.

Fixes #678 

**Checklist:**

### DIFF
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -307,6 +307,16 @@ func main() {
 				return
 			}
 
+			if *faceValue < float64(0) {
+				glog.Errorf("-faceValue must be greater than 0, but %v provided. Restart the node with a different valid value for -faceValue", *faceValue)
+				return
+			}
+
+			if *winProb < float64(0) || *winProb > float64(100) {
+				glog.Errorf("-winProb must be between 0 and 100, but %v provided. Restart the node with a different valid value for -winProb", *winProb)
+				return
+			}
+
 			sigVerifier := &pm.DefaultSigVerifier{}
 			validator := pm.NewValidator(sigVerifier)
 			faceValueInWei := eth.ToBaseUnit(big.NewFloat(*faceValue))


j0sh:
Is there an advantage to continuing with a default value of 0, rather than just stopping the node and avoiding some the potential for surprising behavior around zero values?

yondonfu:
Hm yeah actually probably best to avoid the possibility of unexpected behavior. Made the change.

j0sh (approved):
:ship:

ericxtang (approved):
🚢

PR #690, opened 2019-01-17
Small ux improvements

ericxtang:

Small improvements:
* An error message improvement
* CLI improvement, so that we are printing Eth amount when it's more than 0.01 Eth (instead of 1 Eth)


j0sh (approved):
LGTM 🚢

PR #692, opened 2019-01-17
Google Storage driver

darkdarkdragon:
OS driver for Google Storage

Manually

**Checklist:**


ericxtang:
Initial feedback is that `s3Session.isGoogle` feels a little strange.  What happens when we want to add Microsoft Storage, Digital Ocean Storage, etc?  Might be better to have a `cloudStorageSession` and an `iota` for {S3|Google|MS|DO|etc}

darkdarkdragon:
@ericxtang Google Storage's API is basically same, with only slightly different authentication. I do not want right now engineer "for the future", I wanted minimal changes. If new drivers will have different API - they will have completely separate implementation, if they will be like Google's ones - same API with just small differences - then we will find best way to abstract those differences.

j0sh:
I'm in agreement with Eric. I also think we can avoid explicitly enumerating the storage type altogether, since that can be determined from the `host` field, and the on-the-wire `OSInfo` struct already has the `StorageType` field. We should also have separate functions for `s3Host` and `gsHost` and so forth.

I know the goal was to minimize code changes, but it would be nice to make more use of interfaces rather than special-casing the code around a specific storage driver. This may mean a little more redundancy in the implementation, but the changes within the S3 code itself would be less invasive, and hopefully easier to extend from here for other S3-compatible drivers.

For example, we could have a google-specific structs that implement `OSDriver` and `OSSession`, perhaps something like

```golang
struct gsOS type {
  s3OS
  gsSigner *gsSigner
}
struct gsSession {
  s3Session
}
func  NewGoogleDriver(path string) OSDriver {
  return &gsOS {
    ...
  }
}
func (os *gsOS) NewSession  {
  policy, signature = gsCreatePolicy(os.gsSigner, os.bucket, os.region, path)
  credential = os.gsSigner.clientEmail()
  return &gsSession{
    s3Session {
		driver:         os, // is this actually used anywhere?
		host:           gsHost(os.s3.bucket),
		key:            path + "/",
		policy:         policy,
		signature:      signature,
		credential: credential,
	}
  }
}
```

Likewise, the `postData` method could also take the fields as a parameter rather than embedding the if-check directly into the function, perhaps by generalizing `s3.SaveData` a bit by making most of it a shared helper between the s3 and google functions.

```golang
func (os *gsSession) SaveData(name string, []data byte) (string, error) {
  fields:= map[string]string {
    "GoogleAccessId" = os.credential,
    "signature"  = os.signature,
  }
  gs.s3Session.saveData(name, data, fields)
}
```

This should allow for some more code reuse without intermixing too much driver-specific logic. This approach is not perfect and could certainly be stretched past its breaking point in the future, but it seems preferable to the current approach.

darkdarkdragon:
That's is exactly the thing I wanted to avoid. You started to engineering things for the future, but that future may never happen. And when it will happen, probably we anyway will be re-architecting whole thing again, because we don't know what new provider's API will bring to us...
Plus golang doesn't give us tools to nicely abstract differences between this two driver. What you're suggesting is basically copy-paste of code, which is never good (and that's not all, there is `newS3Session` that's need to be made aware of new structures). 
So I think we should left it as is - right now it doesn't have code duplication and it has clear (and small) separation between two drivers. Anything other will just bloat the code without adding anything to the table. If, and when, we will be adding third S3-like driver, then we will have chance to create more elegant architecture, but considering three different drivers, not two.

j0sh:
> You started to engineering things for the future, but that future may never happen.

There's certainly a balance between speculatively over-engineering, and leaving just enough room for expansion. However, this is less about "engineering for the future" and more about cohesion within the code -- intermixing logic for two different drivers in one.

That being said, I'd fine with leaving this mostly as-is if others are also OK with the approach. There are still a few changes that should be made regardless; I'll note those inline within review comments.

> Plus golang doesn't give us tools to nicely abstract differences between this two driver. What you're suggesting is basically copy-paste of code, which is never good 

Agreed that the abstractions available here aren't quite as nice. My preference is for a little more redundancy or code duplication rather than strictly following DRY at the expense of maintainability or logical flow.

> there is newS3Session that's need to be made aware of new structures

Given that the intent is to make changes to S3 itself less invasive, would something like this work?

```golang
func newGSSession(info *net.S3OSInfo) OSSession {
  return &gsSession{ *(newS3Session(info)).(*s3Session) }
}
```

Or if we're preserving the changes to the `s3Session` object:

```golang
func newGSSession(info *net.S3OSInfo) OSSession {
   sess := newS3Session(info).(*s3Session)
   sess.isGoogle = true
   return sess
}
```

net/lp_rpc.proto
```diff
@@ -68,10 +68,12 @@ message S3OSInfo {
   string signature = 4;
 
   // Needed for POST policy.
-  string xAmzCredential = 5;
+  string credential = 5;
 
   // Needed for POST policy.
   string xAmzDate = 6;
+
+  bool isGoogle = 7;
```
j0sh: Let's add GOOGLE to the `StorageType` enum rather than adding a field here

drivers/s3.go
```diff
@@ -38,47 +38,51 @@ type s3OS struct {
...
 }
 
 // S3BUCKET s3 bucket owned by this node
 var S3BUCKET string
 
-// S3REGION region of s3 bucket owned by this node
-var S3REGION string
-
-func s3Host(bucket string) string {
+func s3Host(bucket string, isGoogle bool) string {
```
j0sh: We probably should just have a `googleHost` or `gsHost` function instead of overloading `s3Host`.

drivers/s3.go
```diff
@@ -38,47 +38,51 @@ type s3OS struct {
...
-		signature:      info.Signature,
-		xAmzDate:       info.XAmzDate,
-		xAmzCredential: info.XAmzCredential,
+		host:       info.Host,
+		key:        info.Key,
+		policy:     info.Policy,
+		signature:  info.Signature,
+		xAmzDate:   info.XAmzDate,
+		credential: info.Credential,
+		isGoogle:   info.IsGoogle,
```
j0sh: Adding Google to the `StorageType` field would allow us to use a separate constructor for Google and avoid setting this field for S3.

```golang
func newGSSession(info *net.S3OSInfo) OSSession {
   sess := newS3Session(info).(*s3Session)
   sess.isGoogle = true
   return sess
}
```

j0sh (changes requested):

darkdarkdragon:
@j0sh I've made the changes, now all the things nicely decoupled :)

drivers/gs.go
```diff
@@ -0,0 +1,162 @@
...
+		ClientX509CertURL       string `json:"client_x509_cert_url,omitempty"`
+	}
+
+	gsSigner struct {
+		jsKey     *gsKeyJSON
+		parsedKey *rsa.PrivateKey
+	}
+
+	gsOS struct {
+		s3OS
```
ericxtang, line 41: This might feel nitpicky, but I think we can do better at naming here.  Should we call this `cloudOS` instead of `s3OS`, move it into `drivers.go` instead of keeping it in `s3.go`, and do the same for s3Session?  Semantically it feels weird to have a `s3OS` inside a `gsOS`, and I think it might be a point of confusion for a new developer who is learning the code.

drivers/gs.go
```diff
@@ -0,0 +1,162 @@
...
+		jsKey:     &gsKey,
+		parsedKey: parsedKey,
+	}
+	return os, nil
+}
+
+func (os *gsOS) NewSession(path string) OSSession {
+	var policy, signature = gsCreatePolicy(os.gsSigner, os.bucket, os.region, path)
+
+	sess := os.s3OS.NewSession(path).(*s3Session)
```
j0sh: Could we initialize the `s3Session` struct directly for this function?

It looks like the only field that's carried over from `s3OS.NewSession` is the `key` field. This is one place where that extra bit of redundancy to re-define `key` is probably OK [1].

This would also avoid us having that conditional around the S3 `createPolicy` which would allow us to keep the flow straight-through without special casing.

```golang
sess := s3Session {
   host: gsHost(os.bucket),
   key: path + "/",
   ... 
}
return &gsSession{s3Session: &sess}
```

[1] To avoid this extra bit of "common computation"  we could modify the code that uses `OSSession.key` to also insert `+ "/" +` wherever it's used. This is sort of repetitive and ugly, but it also seems that `OSSession.key` is only used in concatenations to create the full file path, so it feels a little less bad that way. I don't really have a preference whether to actually do this though; your call.

drivers/gs.go
```diff
@@ -0,0 +1,162 @@
...
+	fields := map[string]string{
+		"GoogleAccessId": os.credential,
+		"signature":      os.signature,
+	}
+	return os.saveData(name, data, fields)
+}
+
+func (os *gsSession) GetInfo() *net.OSInfo {
+	oi := os.s3Session.GetInfo()
+	oi.StorageType = net.OSInfo_GOOGLE
```
j0sh: This works out nicely (along with the use of `newS3Session`) because these functions don't require changes within the s3 driver to accommodate google storage. Looks great.

drivers/gs.go
j0sh: I don't really have a preference about naming here, but `s3OS` could be justified considering Google is using a "S3-compatible API" which is what many object storage providers advertise themselves as supporting.

j0sh (changes requested):
Looking great! Just a couple more small requests around NewSession.

j0sh:
Tinkering with the abstraction a bit; this might also work (untested).

It allows us to reuse the s3Session struct entirely without defining a separate gsSession, and avoids the need to add an extra parameter to `saveData/postData`. This comes at the expense of storing a little more data -- the POST fields and the type.

Up to you what to use.

```diff
diff --git a/drivers/gs.go b/drivers/gs.go
index a38c4d56..f547f52a 100644
--- a/drivers/gs.go
+++ b/drivers/gs.go
@@ -40,10 +40,6 @@ type (
 		s3OS
 		gsSigner *gsSigner
 	}
-
-	gsSession struct {
-		s3Session
-	}
 )
 
 func gsHost(bucket string) string {
@@ -97,31 +93,24 @@ func NewGoogleDriver(bucket, keyFileName string) (OSDriver, error) {
 func (os *gsOS) NewSession(path string) OSSession {
 	var policy, signature = gsCreatePolicy(os.gsSigner, os.bucket, os.region, path)
 
-	sess := os.s3OS.NewSession(path).(*s3Session)
-	sess.credential = os.gsSigner.clientEmail()
-	sess.host = gsHost(os.bucket)
-	sess.policy = policy
-	sess.signature = signature
-	return &gsSession{s3Session: *sess}
-}
-
-func (os *gsSession) SaveData(name string, data []byte) (string, error) {
-	fields := map[string]string{
-		"GoogleAccessId": os.credential,
-		"signature":      os.signature,
+	return &s3Session{
+		typ:        net.OSInfo_GOOGLE,
+		host:       gsHost(os.bucket),
+		key:        path + "/",
+		credential: os.gsSigner.clientEmail(),
+		policy:     policy,
+		signature:  signature,
+		fields: map[string]string{
+			"GoogleAccessId": os.gsSigner.clientEmail(),
+			"signature":      signature,
+		},
 	}
-	return os.saveData(name, data, fields)
-}
-
-func (os *gsSession) GetInfo() *net.OSInfo {
-	oi := os.s3Session.GetInfo()
-	oi.StorageType = net.OSInfo_GOOGLE
-	return oi
 }
 
 func newGSSession(info *net.S3OSInfo) OSSession {
 	sess := newS3Session(info).(*s3Session)
-	return &gsSession{s3Session: *sess}
+	sess.typ = net.OSInfo_GOOGLE
+	return sess
 }
 
 // gsCreatePolicy returns policy, signature
diff --git a/drivers/s3.go b/drivers/s3.go
index 6c34f030..dc8f3eb6 100644
--- a/drivers/s3.go
+++ b/drivers/s3.go
@@ -47,6 +47,8 @@ type s3Session struct {
 	signature  string
 	credential string
 	xAmzDate   string
+	fields     map[string]string
+	typ        net.OSInfo_StorageType
 }
 
 // S3BUCKET s3 bucket owned by this node
@@ -64,6 +66,7 @@ func IsOwnStorageS3(uri string) bool {
 
 func newS3Session(info *net.S3OSInfo) OSSession {
 	return &s3Session{
+		typ:        net.OSInfo_S3,
 		host:       info.Host,
 		key:        info.Key,
 		policy:     info.Policy,
@@ -90,19 +93,22 @@ func NewS3Driver(region, bucket, accessKey, accessKeySecret string) OSDriver {
 }
 
 func (os *s3OS) NewSession(path string) OSSession {
-	var policy, signature, credential, xAmzDate, host string
-	if os.awsAccessKeyID != "" {
-		policy, signature, credential, xAmzDate = createPolicy(os.awsAccessKeyID,
-			os.bucket, os.region, os.awsSecretAccessKey, path)
-		host = s3Host(os.bucket)
-	}
+	policy, signature, credential, xAmzDate := createPolicy(os.awsAccessKeyID,
+		os.bucket, os.region, os.awsSecretAccessKey, path)
 	return &s3Session{
-		host:       host,
+		typ:        net.OSInfo_S3,
+		host:       s3Host(os.bucket),
 		key:        path + "/",
 		policy:     policy,
 		signature:  signature,
 		credential: credential,
 		xAmzDate:   xAmzDate,
+		fields: map[string]string{
+			"x-amz-algorithm":  "AWS4-HMAC-SHA256",
+			"x-amz-credential": credential,
+			"x-amz-date":       xAmzDate,
+			"x-amz-signature":  signature,
+		},
 	}
 }
 
@@ -114,20 +120,10 @@ func (os *s3Session) EndSession() {
 }
 
 func (os *s3Session) SaveData(name string, data []byte) (string, error) {
-	fields := map[string]string{
-		"x-amz-algorithm":  "AWS4-HMAC-SHA256",
-		"x-amz-credential": os.credential,
-		"x-amz-date":       os.xAmzDate,
-		"x-amz-signature":  os.signature,
-	}
-	return os.saveData(name, data, fields)
-}
-
-func (os *s3Session) saveData(name string, data []byte, specificFields map[string]string) (string, error) {
 	// tentativeUrl just used for logging
 	tentativeURL := os.host + "/" + os.key + name
 	glog.V(common.VERBOSE).Infof("Saving to S3 %s", tentativeURL)
-	path, err := os.postData(name, data, specificFields)
+	path, err := os.postData(name, data)
 	if err != nil {
 		// handle error
 		glog.Errorf("Save S3 error: %v", err)
@@ -145,7 +141,8 @@ func (os *s3Session) getAbsURL(path string) string {
 }
 
 func (os *s3Session) GetInfo() *net.OSInfo {
-	oi := &net.OSInfo{
+	return &net.OSInfo{
+		StorageType: os.typ,
 		S3Info: &net.S3OSInfo{
 			Host:       os.host,
 			Key:        os.key,
@@ -154,13 +151,11 @@ func (os *s3Session) GetInfo() *net.OSInfo {
 			Credential: os.credential,
 			XAmzDate:   os.xAmzDate,
 		},
-		StorageType: net.OSInfo_S3,
 	}
-	return oi
 }
 
 // if s3 storage is not our own, we are saving data into it using POST request
-func (os *s3Session) postData(fileName string, buffer []byte, specificFields map[string]string) (string, error) {
+func (os *s3Session) postData(fileName string, buffer []byte) (string, error) {
 	fileBytes := bytes.NewReader(buffer)
 	fileType := http.DetectContentType(buffer)
 	path, fileName := path.Split(os.key + fileName)
@@ -170,7 +165,7 @@ func (os *s3Session) postData(fileName string, buffer []byte, specificFields map
 		"key":          path + "${filename}",
 		"policy":       os.policy,
 	}
-	for k, v := range specificFields {
+	for k, v := range os.fields {
 		fields[k] = v
 	}
 	req, err := newfileUploadRequest(os.host, fields, fileBytes, fileName)
```

drivers/gs.go
darkdarkdragon: I think is `s3` will be better understood by developers - because it is essentially S3 API, and everyone knows what it is.

darkdarkdragon:
I've pushed new version - removed `gsSession` struct

cmd/livepeer/livepeer.go
```diff
@@ -346,9 +348,12 @@ func main() {
 	}
 	if *s3bucket != "" {
 		s3bp := strings.Split(*s3bucket, "/")
-		drivers.S3REGION = s3bp[0]
 		drivers.S3BUCKET = s3bp[1]
 	}
+	if *gsBucket != "" && *gsKey == "" || *gsBucket == "" && *gsKey != "" {
```
eladmallel, line 353: Couldn't this be a bit shorter like `if *gsKey == "" || *gsBucket == "" {` ?

cmd/livepeer/livepeer.go
```diff
@@ -357,6 +362,14 @@ func main() {
 		drivers.NodeStorage = drivers.NewS3Driver(br[0], br[1], cr[0], cr[1])
 	}
 
+	if *gsBucket != "" && *gsKey != "" {
```
eladmallel, line 365: I think we don't need this condition anymore, since the above check returns if either of these variables is empty, right?

cmd/livepeer/livepeer.go
darkdarkdragon: no, here I'm checking if only one of flags defined, and not both

cmd/livepeer/livepeer.go
darkdarkdragon: no, above check only returns if one of the variables empty, and not the other

cmd/livepeer/livepeer.go
eladmallel: Thank you, I now understand what we're doing here. This condition is checking that if the node operator wants to use GS, they must provide both variables, and if they don't want to use GS, they should not provide either of them.

drivers/gs.go
j0sh: Another option is to add the `host` and `bucket` fields in the `gsOS` struct instead of embedding `s3OS`. We only use two out of six fields in `s3OS`. The redundancy is minimal, and we wouldn't have to rename anything.

darkdarkdragon: For me it would be illogical - we still sharing `s3session` and much code around that, so having `s3OS` and `gsOS` seems logical.

drivers/s3.go
```diff
@@ -41,39 +41,41 @@ type s3OS struct {
...
-
 func s3Host(bucket string) string {
 	return fmt.Sprintf("https://%s.s3.amazonaws.com", bucket)
 }
 
 // IsOwnStorageS3 returns true if turi points to S3 bucket owned by this node
 func IsOwnStorageS3(uri string) bool {
-	return strings.HasPrefix(uri, s3Host(S3BUCKET))
+	return strings.HasPrefix(uri, s3Host(S3BUCKET)) ||
+		strings.HasPrefix(uri, gsHost(S3BUCKET))
```
j0sh: Some preliminary testing results:

The `S3BUCKET` variable is never set with google storage, so this will always returns false for Google. Probably need to set the variable (or an equivalent) in livepeer.go.

Also, now that I'm re-reading this, should the google check be put into `drivers.IsOwnExternal` instead of S3?

https://github.com/livepeer/go-livepeer/blob/61071f711fe219daa20615381444f591f21b37b9/drivers/drivers.go#L49-L51

darkdarkdragon: good point, I've separated them

j0sh: S3BUCKET still needs to be set in `livepeer.go` (or a google specific var to hold the bucket)

j0sh:
Testing with B-owned storage seems OK pending the resolution of the S3BUCKET issue.

When trying to test with O-owned storage, I keep getting this:

```
ERROR: 2019/01/22 15:51:39 grpc: server failed to encode response:  rpc error: code = Internal desc = grpc: error while marshaling: proto: string field "net.S3OSInfo.Key" contains invalid UTF-8
```

Haven't diagnosed the issue further beyond this point.

drivers/s3.go
darkdarkdragon: my mistake, fixed

darkdarkdragon:
@j0sh 
it is because of
https://github.com/livepeer/go-livepeer/blob/61071f711fe219daa20615381444f591f21b37b9/server/rpc.go#L293-L294

added in this commit 1fe1479a93e7628539a5fb52da72b2d601756ac4

`RandomIdGenerator` returns slice of arbitrary bytes, and it is being casted into string

j0sh:
How about using `core.RandomManifestID()` rather than `RandomIdGenerator` now that we have a shorter default Manifest ID? Then casting that to a string, of course. A better long term fix is resolving https://github.com/livepeer/go-livepeer/issues/695

darkdarkdragon:
@j0sh I've made the change, now O with Google Storage works

j0sh (approved):
LGTM 🚢

PR #693, opened 2019-01-17
server: set transcoding options in offchain mode

### LINKED ISSUE #681: Set -transcodingOptions in offchain mode
**Describe the bug**
The `BroadcastJobVideoProfiles` field in `mediaserver.go` is set via the `-transcodingOptions` CLI flag during `StartMediaServer`. However, this only happens in onchain mode. This should work in offchain mode as well.

**To Reproduce**
* Run a broadcast node with `-offchain -orchAddr ... -transcodingOptions P240p30fps4x3`
* Start a stream and check the transcoded output
* The default transcoding options are used instead rather than the given `-transcodingOptions` field.

**Expected behavior**
The `-transcodingOptions` should be used in offchain mode, rather than the defaults.

yondonfu:

Allow a broadcaster to use the `-transcodingOptions` flag to set requested video profiles for transcoding even when in off-chain mode.

- Set `BroadcastJobVideoProfiles` in `mediaserver.go` even when in off-chain mode

Ran the node in devenv with the `-offchain` and `-transcodingOptions` flags.

Fixes #681 

**Checklist:**

### DIFF
--- a/server/mediaserver.go
+++ b/server/mediaserver.go
@@ -102,18 +102,16 @@ func NewLivepeerServer(rtmpAddr string, httpAddr string, lpNode *core.LivepeerNo
 
 //StartServer starts the LPMS server
 func (s *LivepeerServer) StartMediaServer(ctx context.Context, maxPricePerSegment *big.Int, transcodingOptions string) error {
-	if s.LivepeerNode.Eth != nil {
-		BroadcastPrice = maxPricePerSegment
-		bProfiles := make([]ffmpeg.VideoProfile, 0)
-		for _, opt := range strings.Split(transcodingOptions, ",") {
-			p, ok := ffmpeg.VideoProfileLookup[strings.TrimSpace(opt)]
-			if ok {
-				bProfiles = append(bProfiles, p)
-			}
+	bProfiles := make([]ffmpeg.VideoProfile, 0)
+	for _, opt := range strings.Split(transcodingOptions, ",") {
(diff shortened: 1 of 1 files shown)


server/mediaserver.go
```diff
@@ -102,18 +102,16 @@ func NewLivepeerServer(rtmpAddr string, httpAddr string, lpNode *core.LivepeerNo
...
+		p, ok := ffmpeg.VideoProfileLookup[strings.TrimSpace(opt)]
+		if ok {
+			bProfiles = append(bProfiles, p)
 		}
-		BroadcastJobVideoProfiles = bProfiles
-		glog.V(common.SHORT).Infof("Transcode Job Price: %v, Transcode Job Type: %v", BroadcastPrice, BroadcastJobVideoProfiles)
 	}
+	BroadcastJobVideoProfiles = bProfiles
+
+	glog.V(common.SHORT).Infof("Transcode Video Profiles: %v", BroadcastJobVideoProfiles)
```
j0sh: The broadcast price was printed before. Shouldn't we still print it now?

j0sh (approved):
Just a small comment regarding logging, otherwise LGTM 🚢

server/mediaserver.go
yondonfu: Broadcast price isn't used right now, but perhaps it will used in the future when we add support for broadcasters to set max prices to use for payments - added it back to the log message!

PR #696, opened 2019-01-18
server: add check for 0 deposit in registerConnection

### LINKED ISSUE #694: Broadcaster Doesn't Halt on Nonzero Deposit Balance
**Describe the bug**
The node used to halt the RTMP stream if the broadcaster had an insufficient deposit to cover a certain length of the stream. Currently, the broadcast node no longer checks the deposit balance and allows the stream to be transcoded even if the balance is zero.

While a properly configured orchestrator should never accept a job from a broadcaster with an insufficient balance, there are still a few reasons for halting the RTMP stream on the B side if the balance is unlikely to be sufficient to be accepted by any orchestrators.

* Fail fast gives immediate feedback for broadcast users around an issue that needs to be addressed. The alternative is continuing the stream, and forcing the broadcaster to "work backwards" : the broadcaster has to realize their stream isn't being transcoded, and further diagnose the reason for that as being an insufficient balance.
* Prevents accidental over-spending, which a malicious orchestrator could capitalize on to slash a broadcaster with a low balance.
* Helps prevent silly mistakes during manual testing 😄 

The old behavior ensured the balance was enough to cover a fixed 5 minutes worth of video given the current price. We could re-add in this type of check using the `BroadcastPrice` or simply ensure the balance is nonzero pending the inclusion of Streamflow-specific pricing options.

**To Reproduce**
1. Start a broadcast node with a deposited balance of zero
2. Start an orchestrator node with a high winProb
3. Stream between B/O, producing a winning ticket
4. Stop the stream

**Expected behavior**
When a B starts a RTMP stream with a low balance (or zero balance), disconnect the stream.

**Additional context**
Related discussion here, with the opposite suggestion: https://github.com/livepeer/go-livepeer/issues/584

yondonfu:

Add a check for 0 deposit when broadcaster is in on-chain mode and registering a connection.

I chose to do the most basic implementation of this check in the interest of not over-engineering until we finalize Streamflow specific pricing options for a broadcaster. At that point, we can implement a more sophisticated check than just checking for 0.

- Add a check for 0 deposit in `registerConnection()`

Added unit tests in `mediaserver_test.go`. Ran a broadcaster node in on-chain mode in devenv and confirmed that 1) RTMP stream is disconnected if the broadcaster's deposit is 0 2) RTMP stream is not disconnected as long as the broadcaster's deposit > 0.

Fixes #694 

**Checklist:**

### DIFF
--- a/server/mediaserver_test.go
+++ b/server/mediaserver_test.go
@@ -3,6 +3,7 @@ package server
 import (
 	"context"
 	"encoding/hex"
+	"errors"
 	"fmt"
 	"math/big"
 	"math/rand"
@@ -11,6 +12,9 @@ import (
 	"testing"
 	"time"
 
+	"github.com/ethereum/go-ethereum/accounts"
+	ethcommon "github.com/ethereum/go-ethereum/common"
+	"github.com/livepeer/go-livepeer/eth"
 	"github.com/livepeer/go-livepeer/pm"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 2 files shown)


server/mediaserver_test.go
```diff
@@ -435,9 +439,38 @@ func TestRegisterConnection(t *testing.T) {
...
-	// Should return an error if missing node storage
-	drivers.NodeStorage = nil
+	// Switch to on-chain mode
+	c := &eth.MockClient{}
+	addr := ethcommon.Address{}
+	s.LivepeerNode.Eth = c
+
+	// Should return an error if in on-chain mode and fail to get sender deposit
+	c.On("Account").Return(accounts.Account{Address: addr})
+	c.On("Senders", addr).Return(nil, nil, nil, errors.New("Senders error")).Once()
```
j0sh, line 449: Just a note that I don't find this particular mocking API too nice. The style is a bit too indirect for me. Much prefer post-test assertions (rather than inverting the flow of logic with expects) and all those `Return` parameters look unwieldy.

I know we're still experimenting with different testing styles though -- and I really haven't looked at what else might be available so I'm just complaining without providing constructive alternatives ATM.

This isn't a change request, just noting some personal impressions around this mocking API.

j0sh (approved):
LGTM 🚢

server/mediaserver_test.go
yondonfu: Yeah I hear you. Definitely up for finding a solid mocking style that everyone is happy with - let's add it as a discussion topic to refactoring week!

PR #698, opened 2019-01-18
Fix ServiceURI Reset Bug

### LINKED ISSUE #691: Transcoding Result Contains Incorrect URL
**Describe the bug**
The transcoding result contains incorrect URL when using a newly registered O. This seems to be because the OS.SaveData in `rpc.go->ServeSegment` is giving back the old URI.

**To Reproduce**
1. Start a new O.
2. Register the O onchain (make sure it's the only O)
3. Start streaming to B
4. You'll see errors described below
5. Stop Stream, restart O
6. The error goes away.

**Expected behavior**
O should automatically update the ServiceURI with the new URI without needed a restart.

**Screenshots**
```
E0117 21:24:22.813874       1 player.go:163] Error getting segment /stream/85eb7484/P240p30fps16x9/0.ts: ErrNotFound
E0117 21:24:22.813901       1 player.go:163] Error getting segment /stream/85eb7484/P360p30fps16x9/0.ts: ErrNotFound
E0117 21:24:22.814074       1 drivers.go:75] Non-200 response 500 Internal Server Error
E0117 21:24:22.814102       1 broadcast.go:158] Download error with segment 0: 500 Internal Server Error (URL: http://127.0.0.1:8935/stream/85eb7484/P360p30fps16x9/0.ts)
E0117 21:24:22.814150       1 drivers.go:75] Non-200 response 500 Internal Server Error
E0117 21:24:22.814179       1 broadcast.go:158] Download error with segment 0: 500 Internal Server Error (URL: http://127.0.0.1:8935/stream/85eb7484/P240p30fps16x9/0.ts)
I0117 21:24:24.166130       1 rpc.go:530] Submitting segment 1 : 567008 bytes
I0117 21:24:24.179402       1 rpc.go:553] Uploaded segment 1
```

ericxtang:

`baseURI` is not updated in the memory driver because it's set at startup time.  So we now change it to a pointer, so it'll always load the latest ServiceURI value.  Thanks to @angyangie for pairing with me on this.

- Update `baseURI` to a pointer
- Create getters and setters to `ServiceURI` so when it's updated, we set the value instead of the pointer value
- Add a test

With `TestServiceURIChange`

Closes https://github.com/livepeer/go-livepeer/issues/691

**Checklist:**

### DIFF
--- a/core/livepeernode.go
+++ b/core/livepeernode.go
@@ -54,7 +54,6 @@ type LivepeerNode struct {
 	Recipient        pm.Recipient
 	OrchestratorPool net.OrchestratorPool
 	Ipfs             ipfs.IpfsApi
-	ServiceURI       *url.URL
 	OrchSecret       string
 	Transcoder       Transcoder
 
@@ -62,6 +61,7 @@ type LivepeerNode struct {
 	Sender pm.Sender
 
 	// Transcoder private fields
+	serviceURI      *url.URL
 	pmSessions      map[ManifestID]map[string]bool
 	pmSessionsMutex *sync.Mutex
 	segmentMutex    *sync.RWMutex
@@ -124,4 +124,15 @@ func (n *LivepeerNode) StopEthServices() error {
 	}
 
 	return nil
+}
+
+func (n *LivepeerNode) GetServiceURI() *url.URL {
+	return n.serviceURI
+}
+
+func (n *LivepeerNode) SetServiceURI(newUrl *url.URL) {
+	if newUrl != nil && n.serviceURI != nil {
+		*n.serviceURI = *newUrl
+	}
+	n.serviceURI = newUrl //for nil cases
 }
--- a/core/livepeernode.go
+++ b/core/livepeernode.go
@@ -61,7 +61,7 @@ type LivepeerNode struct {
 	Sender pm.Sender
 
 	// Transcoder private fields
-	serviceURI      *url.URL
+	serviceURI      url.URL
 	pmSessions      map[ManifestID]map[string]bool
 	pmSessionsMutex *sync.Mutex
 	segmentMutex    *sync.RWMutex
@@ -127,12 +127,9 @@ func (n *LivepeerNode) StopEthServices() error {
 }
 
 func (n *LivepeerNode) GetServiceURI() *url.URL {
-	return n.serviceURI
+	return &n.serviceURI
 }
 
 func (n *LivepeerNode) SetServiceURI(newUrl *url.URL) {
-	if newUrl != nil && n.serviceURI != nil {
-		*n.serviceURI = *newUrl
-	}
-	n.serviceURI = newUrl //for nil cases
+	n.serviceURI = *newUrl
 }
--- a/core/livepeernode_test.go
+++ b/core/livepeernode_test.go
@@ -3,11 +3,15 @@ package core
 import (
 	"fmt"
 	"io/ioutil"
+	"net/url"
 	"os"
 	"testing"
 
+	"github.com/golang/glog"
 	"github.com/livepeer/go-livepeer/drivers"
 	"github.com/livepeer/lpms/ffmpeg"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 type StubTranscoder struct {
@@ -34,7 +38,7 @@ func TestTranscodeAndBroadcast(t *testing.T) {
 	ffmpeg.InitFFmpeg()
 	p := []ffmpeg.VideoProfile{ffmpeg.P720p60fps16x9, ffmpeg.P144p30fps16x9}
 	tr := &StubTranscoder{Profiles: p}
-	storage := drivers.NewMemoryDriver("").NewSession("")
+	storage := drivers.NewMemoryDriver(nil).NewSession("")
 	config := transcodeConfig{LocalOS: storage, OS: storage}
 
 	tmpdir, _ := ioutil.TempDir("", "")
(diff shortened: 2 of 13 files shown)


angyangie (approved):
Thanks for taking me on this bug hunting journey with you @ericxtang ! I learned a lot

drivers/local.go
```diff
@@ -124,8 +124,8 @@ func (ostore *MemorySession) getAbsolutePath(name string) string {
 
 func (ostore *MemorySession) getAbsoluteURI(name string) string {
 	name = "/stream/" + ostore.getAbsolutePath(name)
-	if ostore.os.baseURI != "" {
-		return ostore.os.baseURI + name
+	if ostore.os.baseURI != nil {
+		return ostore.os.baseURI.String() + name
```
j0sh, line 128: Just a note that this won't be as "cheap" as it used to be, since serializing a string from a URL object does have a cost, see:  https://golang.org/src/net/url/url.go?s=21757:21786#L741 . Not a huge deal though as we don't call this function too often, just once per save.

j0sh (approved):
LGTM! 🚢

core/livepeernode_test.go
```diff
@@ -91,3 +95,29 @@ func TestTranscodeAndBroadcast(t *testing.T) {
...
+	require.Nil(err)
+	assert.Equal("test://testurl.com/stream/testpath/testdata1", savedUrl)
+
+	glog.Infof("Setting service URL to newurl")
+	newUrl, err := url.Parse("test://newurl.com")
+	n.SetServiceURI(newUrl)
+	require.Nil(err)
+	url, err := sesh.SaveData("testdata2", []byte{0, 0, 0})
+	require.Nil(err)
+	assert.Equal("test://newurl.com/stream/testpath/testdata2", url)
```
darkdarkdragon: adding this 
```
	glog.Infof("Setting service URL to secondurl")
	secondUrl, err := url.Parse("test://secondurl.com")
	n.SetServiceURI(secondUrl)
	require.Nil(err)
	surl, err = sesh.SaveData("testdata3", []byte{0, 0, 0})
	require.Nil(err)
	assert.Equal("test://secondurl.com/stream/testpath/testdata3", surl)
```
to the end of test function will make it break

core/livepeernode.go
```diff
@@ -125,3 +125,14 @@ func (n *LivepeerNode) StopEthServices() error {
...
+
+func (n *LivepeerNode) GetServiceURI() *url.URL {
+	return n.serviceURI
+}
+
+func (n *LivepeerNode) SetServiceURI(newUrl *url.URL) {
+	if newUrl != nil && n.serviceURI != nil {
+		*n.serviceURI = *newUrl
+	}
+	n.serviceURI = newUrl //for nil cases
```
darkdarkdragon: this should be in `else` branch - right now you're overwriting pointer and it breaks on second call to `SetServiceURI`.
I've added test case in comments.

darkdarkdragon (changes requested):

core/livepeernode.go
j0sh: Part of why this reads a bit strangely is because we're both 1) setting a possibly-nil pointer, and 2) setting the contents of that pointer. Both are done under different conditions.

If we can remove the first case then the function becomes much more straightforward. We can do this by storing the `LivepeerNode.serviceURI` as a value rather than a pointer. Eg,

```
type LivepeerNode struct {
   ...
	// Transcoder private fields
	serviceURI      url.URL // not a pointer
   ...
}

func (n *LivepeerNode) SetServiceURI(newUrl *url.URL) {
    // Is there anything here we need to be concerned about deep copying?
    // Most of the fields are straightforward but there's the UserInfo pointer
    // which I'm unsure about: https://golang.org/pkg/net/url/#URL
    n.serviceURI = *newURL
}
func (n *LivepeerNode) GetServiceURI() *url.URL {
    return &n.serviceURI
}
```

We might even be able to get rid of the accessor functions that way, but keeping them in is fine too.

Also, if we accept that the ServiceURI can be (re-)set at any time (and I suppose there's no reason it can't) then the race detector will probably complain that accesses to `n.serviceURI` are not protected, since the node may be transcoding at the moment that the service URI is set. Properly fixing that is more trouble than it's worth at this point though.

darkdarkdragon: > Part of why this reads a bit strangely is because we're both 1) setting a possibly-nil pointer, and 2) setting the contents of that pointer. Both are done under different conditions.

3) it is always overwriting pointer. So we are not achieving goal of `so it'll always load the latest ServiceURI value`- because on second invocation of `SetServiceURI` we're not setting content of `baseURI` - `LivepeerNode` doesn't have that pointer anymore. 
And we can't just simply set `serviceURI` to be value instead of pointer - `MemoryDriver` still need a way to check if it's empty.

Probably ideal solution here will be 
```golang
	if newUrl != nil && n.serviceURI != nil {
		*n.serviceURI = *newUrl
	} else if n.serviceURI == nil && newUrl != nil {
		n.serviceURI = &url.URL{}
		*n.serviceURI = *newUrl
	}
```

j0sh: > And we can't just simply set serviceURI to be value instead of pointer - MemoryDriver still need a way to check if it's empty.

The MemoryDriver only needs to check that the ServiceURI is non-nil here (which this PR does in `getAbsoluteURI`). Calling `.String()` on an empty URL yields the empty string which is fine.

Storing a value rather than a pointer should still work, unless I'm missing something else.

darkdarkdragon: If `.String()` returns empty string for empty `URL` structure then storing value is a way to go

j0sh (approved):

darkdarkdragon (approved):
LGTM

Issue #685, opened 2019-01-16
Broadcaster does not verify orchestrator's signature


yondonfu:
**Describe the bug**

A broadcaster currently does not verify an orchestrator's signature over the hash of the concatenated transcoded segment hashes. 

See: https://github.com/livepeer/go-livepeer/blob/c45d7e33e6ae22ac08260085e379cfd6bb56e7d3/server/mediaserver.go#L409

**To Reproduce**
1. See commented out code in above link.

**Expected behavior**

A broadcaster should verify the orchestrator's signature over transcoded results. 

**Desktop (please complete the following information):**
 - OS: Ubuntu
 - Version: 1b65ae0

j0sh:
Note that the sig + hash `Sig_O(H(H(Rendition_1) | ... | H(Rendition_N)))` also needs to be generated on the O side first.

The hash was formerly generated by the claimmanager but was never re-implemented after the claimmanager was removed.

j0sh:
Addressed by https://github.com/livepeer/go-livepeer/pull/702

PR #702, opened 2019-01-22
Add Broadcaster Verification of Orchestrator Signature

angyangie:

Adds verification of an orchestrator's signature over the hash of the concatenated transcoded segment hashes.

- Adds sig + `hash Sig_O(H(H(Rendition_1) | ... | H(Rendition_N)))` on orchestrator.go
- Adds `pm.VerifySig` to `processSegment` in `broadcast.go`

- I ran and testes using test harness
- Ran `test.sh`

https://github.com/livepeer/go-livepeer/issues/685

**Checklist:**

### DIFF
--- a/core/core_test.go
+++ b/core/core_test.go
@@ -85,6 +85,70 @@ func TestTranscode(t *testing.T) {
 	// TODO check transcode loop expiry, storage, sig construction, etc
 }
 
+func StubTranscodeConfig() transcodeConfig {
+	return transcodeConfig{
+		OS:      nil,
+		LocalOS: nil,
+	}
+}
+
+func TestTranscodeSeg_OffchainMode_ReturnsEmptySignature(t *testing.T) {
+	//Set up the node
+	drivers.NodeStorage = drivers.NewMemoryDriver("")
+	db, err := common.InitDB("file:TestTranscode?mode=memory&cache=shared")
+	require := require.New(t)
+	require.Nil(err)
+	defer db.Close()
+
+	tmp, err := ioutil.TempDir("", "")
+	require.Nil(err)
+	n, err := NewLivepeerNode(nil, tmp, db)
+	require.Nil(err)
+
+	// mock transcoder
+	n.Transcoder = NewLocalTranscoder(tmp)
+	ffmpeg.InitFFmpeg()
+	defer os.RemoveAll(tmp)
+
+	// mock transcodeSeg
+	transcodeConfig := StubTranscodeConfig()
+	segment := StubSegment()
+	segmentMetadata := &SegTranscodingMetadata{Profiles: videoProfiles}
+	transcodeResult := n.transcodeSeg(transcodeConfig, segment, segmentMetadata)
+
+	assert := assert.New(t)
+	assert.Nil(transcodeResult.Sig)
+}
+
+func TestTranscodeSeg_OnchainMode_ReturnsSignature(t *testing.T) {
+	//Set up the node
+	drivers.NodeStorage = drivers.NewMemoryDriver("")
+	db, err := common.InitDB("file:TestTranscode?mode=memory&cache=shared")
+	require := require.New(t)
+	require.Nil(err)
+	defer db.Close()
+
+	tmp, err := ioutil.TempDir("", "")
+	require.Nil(err)
+	seth := &eth.StubClient{}
+	n, err := NewLivepeerNode(seth, tmp, db)
+	require.Nil(err)
+
+	// mock transcoder
+	n.Transcoder = NewLocalTranscoder(tmp)
+	ffmpeg.InitFFmpeg()
+	defer os.RemoveAll(tmp)
+
+	// mock transcodeSeg
+	transcodeConfig := StubTranscodeConfig()
+	segment := StubSegment()
+	segmentMetadata := &SegTranscodingMetadata{Profiles: videoProfiles}
+	transcodeResult := n.transcodeSeg(transcodeConfig, segment, segmentMetadata)
+
+	assert := assert.New(t)
+	assert.NotNil(transcodeResult.Sig)
+}
+
 func TestTranscodeLoop_GivenNoSegmentsPastTimeout_CleansSegmentChan(t *testing.T) {
 	//Set up the node
 	drivers.NodeStorage = drivers.NewMemoryDriver(nil)
--- a/core/core_test.go
+++ b/core/core_test.go
@@ -17,7 +17,6 @@ import (
 	"github.com/stretchr/testify/mock"
 	"github.com/stretchr/testify/require"
 
-	"github.com/livepeer/go-livepeer/common"
 	"github.com/livepeer/go-livepeer/drivers"
 	"github.com/livepeer/go-livepeer/eth"
 	"github.com/livepeer/go-livepeer/pm"
@@ -43,15 +42,10 @@ var videoProfiles = []ffmpeg.VideoProfile{ffmpeg.P144p30fps16x9, ffmpeg.P240p30f
 func TestTranscode(t *testing.T) {
 	//Set up the node
 	drivers.NodeStorage = drivers.NewMemoryDriver(nil)
-	db, err := common.InitDB("file:TestTranscode?mode=memory&cache=shared")
-	if err != nil {
-		t.Error("Error initializing DB: ", err)
-	}
-	defer db.Close()
 	seth := &eth.StubClient{}
 	tmp, _ := ioutil.TempDir("", "")
-	n, _ := NewLivepeerNode(seth, tmp, db)
 	defer os.RemoveAll(tmp)
+	n, _ := NewLivepeerNode(seth, tmp, nil)
 	ffmpeg.InitFFmpeg()
 
 	ss := StubSegment()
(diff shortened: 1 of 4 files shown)


core/orchestrator.go
```diff
@@ -315,16 +316,34 @@ func (n *LivepeerNode) transcodeSeg(config transcodeConfig, seg *stream.HLSSegme
...
 			glog.Errorf("Cannot find transcoded segment for %v", seg.SeqNo)
 			continue
 		}
 		tProfileData[md.Profiles[i]] = tData[i]
 		tr.Data = append(tr.Data, tData[i])
+
+		hash := crypto.Keccak256(tData[i])
+		segHashLock.Lock()
+		segHashes[i] = hash
+		segHashLock.Unlock()
```
j0sh: The iteration over `md.Profiles` is done synchronously, so the locking is not necessary here

server/broadcast.go
```diff
@@ -217,10 +217,11 @@ func processSegment(cxn *rtmpConnection, seg *stream.HLSSegment) {
 		}
 		cond.L.Unlock()
 
-		// if !eth.VerifySig(transcoderAddress, crypto.Keccak256(segHashes...), res.Sig) { // need transcoder address here
-		// 	glog.Error("Sig check failed for segment ", seg.SeqNo)
-		// 	return
-		// }
+		ticketParams := cxn.sess.OrchestratorInfo.GetTicketParams()
+		if !pm.VerifySig(ethcommon.BytesToAddress(ticketParams.Recipient), crypto.Keccak256(segHashes...), res.Sig) {
```
j0sh, line 221: It feels a bit weird to be invoking `pm` here but I see this avoids having to bring the `eth` package in so this is fine

core/orchestrator.go
```diff
@@ -315,16 +316,34 @@ func (n *LivepeerNode) transcodeSeg(config transcodeConfig, seg *stream.HLSSegme
...
 	}
 	os.Remove(fname)
 	tr.OS = config.OS
+
+	if n == nil || n.Eth == nil {
+		return &tr
+	}
+
+	segHash := crypto.Keccak256(segHashes...)
+	tr.Sig, err = n.Eth.Sign(segHash)
```
j0sh: At some point it might be nice to use the `Sign` method via the `orchestrator` struct to avoid doing the on-chain checks ourselves, but we don't have access to such a struct at this point, so this is OK.

j0sh (changes requested):

server/broadcast.go
eladmallel: @j0sh @yondonfu I think we would all enjoy further refactoring around ETH stuff; basically breaking it down into: (a) generic ETH helpers, and (b) Livepeer-specific stuff.

I think that `VerifySig` could live in package (a), such that `pm` would depend on (a) as well.

wdyt?

core/orchestrator.go
angyangie: Agreed that this is a bit awkward, and we could work to streamline in the future!

server/broadcast.go
angyangie: Where did you envision the generic ETH helpers would live @eladmallel ?

eladmallel: @angyangie in a new package. I think it's best to have this conversation with a graph of the current dependencies, and from there try to create a new graph of what things should look like.

We've had to do some back bending due to cyclic import problems.

I don't know of any official tools for this, but bumped into [this one](https://github.com/davecheney/graphpkg) that might be useful.

@j0sh do you have a go-to tool for visualizaing dependencies in Go?

angyangie: That's a great idea @eladmallel. I've never used one of these packages, but seems like it could indeed help us in organizing our dependencies/imports

core/orchestrator.go
```diff
@@ -315,16 +316,34 @@ func (n *LivepeerNode) transcodeSeg(config transcodeConfig, seg *stream.HLSSegme
...
+
+		hash := crypto.Keccak256(tData[i])
+		segHashLock.Lock()
+		segHashes[i] = hash
+		segHashLock.Unlock()
 	}
 	os.Remove(fname)
 	tr.OS = config.OS
+
+	if n == nil || n.Eth == nil {
```
yondonfu, line 352: Could we add unit tests for the following cases?

- Node in off-chain mode should return `tr` with `tr.Sig` empty
- Node in on-chain mode should return `tr` with the correct `tr.Sig`
- Node in on-chain mode that fails to generate a signature on L343 should return `tr` with `tr.Sig = nil` (should be possible to use a mock implementation for the `LivepeerEthClient` interface field on `n`)

[core_test.go](https://github.com/livepeer/go-livepeer/blob/master/core/core_test.go#L85) seems to be the current file for unit tests related to the transcode loop.

yondonfu (changes requested):

server/broadcast.go
j0sh: We could split out the helpers. Note that I don't really have objections to the redundancy within PM since that's a bit of a special case. Ethereum-related code isn't the only area where we technically could refactor; we also have a number of helpers to generate random bytes, and so forth. Do we pull `pm.RandBytes` into a special non-Ethereum, non-Livepeer helper?

This line of thinking extends further anytime we might have a new module that might be a good candidate for open sourcing [1], especially if the helpers required by the new module are disjoint from others such as `pm`. Each new "standalone" module potentially means another attached helper module.

We should be mindful of taking DRY too far and splintering the codebase or making the project structure unnatural to satisfy a very specific constraint (eg, minimal dependencies in PM just so it may be generalized or open sourced one day [2]). We already have eth/helpers.go for ETH-specific helpers, `common` for more general helpers, and so forth. Then we'll have questions about which new code is Livepeer-specific and what "might" be useful outside the context of Livepeer. It's very easy to burn cycles on issues such as code organization when the project structure becomes complex.

I'm more inclined to accept small peccadilloes as-is rather than aiming for perfection, and refactoring as-we-go when the pieces naturally fall into place to enable that. In other words, reorganization and refactoring is a continuous process, rather than a single end goal.

[1]  Eg, we extract a 'transcoder management layer' into a general worker management mechanism incorporating reputation, redundancy, failover, etc. I have no intentions of doing this right now, but just putting this out by way of example.

[2] We probably should formalize this separation and extract out `pm` sooner rather than later, lest we lose sight of the original intention, and later developers start to introduce changes that increase the coupling with the Livepeer code.

yondonfu: @j0sh regarding [2], do you think it would suffice to leave the `pm` package and write a separate README indicating that it can be standalone package (despite being nested in go-livepeer) and highlight that Livepeer specific dependencies should not be introduced in the package?

eladmallel: @j0sh I completely agree with your iterative approach, and that we should not strive for "perfect DRY".

Having said that, I also think that in the upcoming refactoring/cleanup session, it could be beneficial to re-org some of our packages.

From your note I'm not sure if I understand your POV for the very near term: do you think we should leave the current architecture as-is, or are there refactoring opportunities that you're in favor of? (e.g. splitting out DB from `common`? moving `pm` out, maybe a-la `lpms`?)

j0sh: > do you think it would suffice to leave the pm package and write a separate README indicating that it can be standalone package

We can certainly try to enforce it socially for now. Long-term, splitting it into a standalone repo also means the split is enforced by the compiler, so people would have to 'think before they push'. It would also be more beneficial as far as encouraging visibility, adoption and community participation around probabilistic micropayments as a concept.

> are there refactoring opportunities that you're in favor of? (e.g. splitting out DB from common? moving pm out, maybe a-la lpms?)

Those two sound good, yes.

If we take the view that the `common` package is actually for *helpers* rather than common code that's used by most modules, then the DB does feel out of place there.

We'll might still end up with issues if we're not consistent with how the DB is used. Thus far, all code has used the DB directly because there hasn't been a need to abstract it out. PM inverts the flow of control by making the DB implement a PM-specific storage interface. The DB then becomes pm-aware, and it's not really possible to export data or features that are specific to the storage layer, at least not without yet another central module for `types` or so forth (the `net` module in goclient has sort of taken on this role elsewhere). We also end up with more glue code to make modules interoperate (eg, livepeer.go).

Anyway, I'm really OK with the split, just highlighting some potential issues that we should be aware of wrt the inversion of control.

angyangie: Seems like we can continue discussion here: https://github.com/livepeer/go-livepeer/issues/717

core/core_test.go
```diff
@@ -85,6 +85,70 @@ func TestTranscode(t *testing.T) {
...
+	defer os.RemoveAll(tmp)
+
+	// mock transcodeSeg
+	transcodeConfig := StubTranscodeConfig()
+	segment := StubSegment()
+	segmentMetadata := &SegTranscodingMetadata{Profiles: videoProfiles}
+	transcodeResult := n.transcodeSeg(transcodeConfig, segment, segmentMetadata)
+
+	assert := assert.New(t)
+	assert.NotNil(transcodeResult.Sig)
```
j0sh: We should also check the signature itself to ensure the sig construction was done properly. There's already a StubTranscoder which might help with this, which can be invoked on its own without initializing FFmpeg and the LocalTranscoder.

core/core_test.go
```diff
@@ -85,6 +85,70 @@ func TestTranscode(t *testing.T) {
...
+	defer db.Close()
+
+	tmp, err := ioutil.TempDir("", "")
+	require.Nil(err)
+	n, err := NewLivepeerNode(nil, tmp, db)
+	require.Nil(err)
+
+	// mock transcoder
+	n.Transcoder = NewLocalTranscoder(tmp)
+	ffmpeg.InitFFmpeg()
```
j0sh: Using the existing StubTranscoder would avoid ffmpeg initialization and lead to faster / more predictable test results. (And technically we'd no longer be writing to disk either). Note that this option would require us to set up a `transcodeConfig.LocalOS` storage session.

core/core_test.go
```diff
@@ -85,6 +85,70 @@ func TestTranscode(t *testing.T) {
...
+func TestTranscodeSeg_OffchainMode_ReturnsEmptySignature(t *testing.T) {
+	//Set up the node
+	drivers.NodeStorage = drivers.NewMemoryDriver("")
+	db, err := common.InitDB("file:TestTranscode?mode=memory&cache=shared")
+	require := require.New(t)
+	require.Nil(err)
+	defer db.Close()
+
+	tmp, err := ioutil.TempDir("", "")
+	require.Nil(err)
```
j0sh: We should `defer os.RemoveAll(tmp)` immediately here, because if any other `require`s fail before calling `os.RemoveAll` then the dir will dangle.

core/orchestrator.go
j0sh: Should we propagate any signing errors? That is, `tr.Sig, tr.Err = n.Eth.Sign(segHash)`

core/core_test.go
```diff
@@ -85,6 +85,70 @@ func TestTranscode(t *testing.T) {
...
+	return transcodeConfig{
+		OS:      nil,
+		LocalOS: nil,
+	}
+}
+
+func TestTranscodeSeg_OffchainMode_ReturnsEmptySignature(t *testing.T) {
+	//Set up the node
+	drivers.NodeStorage = drivers.NewMemoryDriver("")
+	db, err := common.InitDB("file:TestTranscode?mode=memory&cache=shared")
```
j0sh: I wonder if we can live without the database for these tests. I sort of like the minimal approach for a few reasons:

* Reduces boilerplate
* Makes it explicit what parts are required to make the set of tests pass
* If we ever do start using the DB around transcoding, then it's an opportunity to reassess whether it should be mandatory (we used to check for DB presence in a few places). Although I can see the argument for "future proofing" this to mitigate additions down the road.

core/core_test.go
```diff
@@ -85,6 +85,70 @@ func TestTranscode(t *testing.T) {
...
+	transcodeConfig := StubTranscodeConfig()
+	segment := StubSegment()
+	segmentMetadata := &SegTranscodingMetadata{Profiles: videoProfiles}
+	transcodeResult := n.transcodeSeg(transcodeConfig, segment, segmentMetadata)
+
+	assert := assert.New(t)
+	assert.Nil(transcodeResult.Sig)
+}
+
+func TestTranscodeSeg_OnchainMode_ReturnsSignature(t *testing.T) {
```
j0sh: To me, this seems like a good example of where the style of breaking cases into single, specific tests does not facilitate overall understanding of the cases under test. Too much boilerplate, and the high-level understanding of the test coverage is inhibited by the loss of locality.

How does this look in terms of a halfway point between "whole function testing" and "single case testing"? This also incorporates most of the feedback from elsewhere in this review.

```diff
diff --git a/core/core_test.go b/core/core_test.go
index 7cc06889..01ca7d0e 100644
--- a/core/core_test.go
+++ b/core/core_test.go
@@ -82,7 +82,45 @@ func TestTranscode(t *testing.T) {
                t.Error("Unexpected transcode result ", len(tr.Data[1]))
        }
 
-       // TODO check transcode loop expiry, storage, sig construction, etc
+       // TODO check transcode loop expiry, storage, etc
+}
+
+func TestTranscodeSig(t *testing.T) {
+       tmp, _ := ioutil.TempDir("", "")
+       defer os.RemoveAll(tmp)
+
+       p := []ffmpeg.VideoProfile{ffmpeg.P720p60fps16x9, ffmpeg.P144p30fps16x9}
+       n, _ := NewLivepeerNode(nil, tmp, nil)
+       n.Transcoder = &StubTranscoder{Profiles: p}
+       conf := transcodeConfig{LocalOS: (drivers.NewMemoryDriver("")).NewSession("")}
+       md := &SegTranscodingMetadata{Profiles: p}
+       seg := StubSegment()
+
+       assert := assert.New(t)
+       require := require.New(t)
+
+       // Test offchain mode
+       require.Nil(n.Eth) // sanity check the offchain precondition of a nil eth
+       res := n.transcodeSeg(conf, seg, md)
+       assert.Nil(res.Err)
+       assert.Nil(res.Sig)
+       // sanity check results
+       resBytes, _ := n.Transcoder.Transcode("", p)
+       assert.Equal(resBytes, res.Data)
+
+       // Test onchain mode
+       n.Eth = &eth.StubClient{}
+       res = n.transcodeSeg(conf, seg, md)
+       assert.Nil(res.Err)
+       assert.NotNil(res.Sig)
+       // check sig
+       resHashes := make([][]byte, len(p))
+       for i, v := range resBytes {
+               resHashes[i] = ethCrypto.Keccak256(v)
+       }
+       resHash := ethCrypto.Keccak256(resHashes...)
+       assert.Equal(resHash, res.Sig)
 }
 
 func TestTranscodeLoop_GivenNoSegmentsPastTimeout_CleansSegmentChan(t *testing.T) {

diff --git a/eth/stubclient.go b/eth/stubclient.go
index 18cba6de..5e6f2ad2 100644
--- a/eth/stubclient.go
+++ b/eth/stubclient.go
@@ -260,7 +260,7 @@ func (c *StubClient) CheckTx(tx *types.Transaction) error {
 func (c *StubClient) ReplaceTransaction(tx *types.Transaction, method string, gasPrice *big.Int) (*types.Transaction, error) {
        return nil, nil
 }
-func (c *StubClient) Sign(msg []byte) ([]byte, error)   { return nil, nil }
+func (c *StubClient) Sign(msg []byte) ([]byte, error)   { return msg, nil }
```

j0sh (changes requested):

core/core_test.go
angyangie: It's not mandatory to use, so we can remove for now. However, you're right that in the future, we might need to add back in if we start using the DB in transcoding. But we can cross that bridge when we get there!

core/core_test.go
angyangie: Good point!

core/core_test.go
angyangie: Thank you for the example Josh! Really helpful.

core/core_test.go
angyangie: Interesting. I remove `ffmpeg.InitFFmpeg()` from all test files, without adding anything else, and all tests still pass. Why could that be?

core/orchestrator.go
```diff
@@ -315,16 +315,31 @@ func (n *LivepeerNode) transcodeSeg(config transcodeConfig, seg *stream.HLSSegme
...
+		return &tr
+	}
+
+	segHash := crypto.Keccak256(segHashes...)
+	tr.Sig, tr.Err = n.Eth.Sign(segHash)
+	if err != nil {
```
j0sh: `tr.Err != nil`

core/orchestrator.go
```diff
@@ -315,16 +315,31 @@ func (n *LivepeerNode) transcodeSeg(config transcodeConfig, seg *stream.HLSSegme
...
+		glog.Error("Unable to sign hash of transcoded segment hashes")
```
yondonfu: Might be helpful to log the signing error here as well

j0sh (approved):
LGTM

yondonfu (approved):
LGTM 🚢

Issue #715, opened 2019-01-24
Split `rpc.go` into client and server parts


darkdarkdragon:
**Is your feature request related to a problem? Please describe.**
`rpc.go` grown to big and it's hard to understand program flow in it.

**Describe the solution you'd like**
Split `rpc.go` into `rpc_server.go` and `roc_client.go`

**Describe alternatives you've considered**
Do nothing.

angyangie:
Perhaps we can tackle this task along with these two issues, since they're all related:

1. Adding unit tests for ServeSegment and SubmitSegment: https://github.com/livepeer/go-livepeer/issues/675

2. Cleaning up lengthy ServeSegment and SubmitSegment functions: https://github.com/livepeer/go-livepeer/issues/714

darkdarkdragon:
PR #732  is merged

PR #723, opened 2019-01-29
Split server and client code from rpc file

angyangie:
******NOTE: HAD STRANGE GIT ISSUES, SO OPENED NEW BRANCH:** #732

********************

https://github.com/livepeer/go-livepeer/issues/715

- Split server and client code from rpc file
- Organize functions by order of where they're called in the code. Functions called by another function within the same file are placed below caller function

https://github.com/livepeer/go-livepeer/issues/715

**Checklist:**

### DIFF
--- a/core/orch_test.go
+++ b/core/orch_test.go
@@ -261,6 +261,18 @@ func TestGetSegmentChan(t *testing.T) {
 		t.Error("SegmentChans mapping did not include channel")
 	}
 
+	// Test max sessions
+	oldTranscodeSessions := MaxTranscodeSessions
+	MaxTranscodeSessions = 0
+	if _, err := n.getSegmentChan(segData); err != nil {
+		t.Error("Existing mid should continue processing even when O is at capacity: ", err)
+	}
+	segData.ManifestID = ManifestID(t.Name())
+	if _, err := n.getSegmentChan(segData); err != ErrOrchCap {
+		t.Error("Didn't fail when orch cap hit: ", err)
+	}
+	MaxTranscodeSessions = oldTranscodeSessions
+
 	// Test what happens when invoking the transcode loop fails
 	drivers.NodeStorage = nil // will make the transcode loop fail
 	node, _ := NewLivepeerNode(nil, "", nil)
@@ -286,6 +298,29 @@ func TestGetSegmentChan(t *testing.T) {
 		t.Error("SegmentChans mapping included new channel when expected to return an err/nil")
 	}
 
+}
+
+func TestOrchCheckCapacity(t *testing.T) {
+	drivers.NodeStorage = drivers.NewMemoryDriver("")
+	n, _ := NewLivepeerNode(nil, "", nil)
+	o := NewOrchestrator(n)
+	md := StubSegTranscodingMetadata()
+	cap := MaxTranscodeSessions
+	assert := assert.New(t)
+
+	// happy case
+	assert.Nil(o.CheckCapacity(md.ManifestID))
+
+	// capped case
+	MaxTranscodeSessions = 0
+	assert.Equal(ErrOrchCap, o.CheckCapacity(md.ManifestID))
+
+	// ensure existing segment chans pass while cap is active
+	MaxTranscodeSessions = cap
+	_, err := n.getSegmentChan(md) // store md into segment chans
+	assert.Nil(err)
+	MaxTranscodeSessions = 0
+	assert.Nil(o.CheckCapacity(md.ManifestID))
 }
 
 func TestProcessPayment_GivenRecipientError_ReturnsError(t *testing.T) {
--- a/server/rpc_test.go
+++ b/server/rpc_test.go
@@ -29,9 +29,10 @@ import (
 )
 
 type stubOrchestrator struct {
-	priv    *ecdsa.PrivateKey
-	block   *big.Int
-	signErr error
+	priv       *ecdsa.PrivateKey
+	block      *big.Int
+	signErr    error
+	sessCapErr error
 }
 
 func (r *stubOrchestrator) ServiceURI() *url.URL {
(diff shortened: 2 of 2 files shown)


j0sh (changes requested):
Should the files be named `rpc_orchestrator.go` and `rpc_broadcaster.go` instead of `rpc_server` and `rpc_client`? Client/server feels too generic here.

`StartTranscodeServer` is in rpc_client; this probably should be rpc_server.

The following functions are used in `rpc_client` but are defined in `rpc_server`. Let's move these over:
* genSegCreds
* genOrchestratorReq
* startOrchestratorClient
* genPayment
* and possibly more?

angyangie:
There might be a difference in ideology here. What @darkdarkdragon had in mind, was more along the lines of dividing these two files into code that implements remote procedure calls, and code that responds to RPC calls. So, by that logic, calling functions go into one file and responding, into another.

From one of our discord conversations:
> (@angyangie ) When we talk about client side, we mean code used by the Broadcaster? Server side is code used by the transcoders/orchs?
> (@darkdarkdragon ) I wanted to split all 'calling' code from code that responding to that calls. Say, we have `CheckOrchestratorAvailability` function - it is used by orchestrator in livepeer.go, but it is making call, and there is `func (h *lphttp) Ping` that responds to the call. (for example) `startOrchestratorClient` should be on client side, and `StartTranscodeServer` - on server

Originally, I thought splitting by `broadcaster` and `orchestrator` code might make sense as well. Anyone would like to make a case for that here? @yondonfu @j0sh

j0sh:
The majority of the O code is server side. The only place where this isn't the case is when the O is calling itself during the `ping` check. I'm OK with keeping `startOrchestratorClient` along with the rest of the broadcaster code and accepting that it's not 100% consistent naming-wise.

If the goal of this refactor is to increase clarity then I think that should come with a corresponding increase in specificity, so I'm against the server/client naming.

Just mentioning again that I prefer keeping functionally related code together (eg, requests with responses) to make it easier to understand the round-trip flow. Scattering code between files does not really facilitate this. How about this:

* SubmitSegment and ServeSegment and related code (genPayment, genSegCreds, etc) in its own file, say `segment_rpc.go`.
* Rest of the RPC methods in another file (perhaps just leave them in `rpc.go`). There won't be too much left after the segment-specific code is split. If we feel that `rpc.go` starts becoming too big again, then we can further split out specific cases later.

yondonfu:
> SubmitSegment and ServeSegment and related code (genPayment, genSegCreds, etc) in its own file, say `segment_rpc.go`.

I am on board with keeping the segment metadata/payment generation/verification code together for clarity. But, perhaps the file name should be `rpc_segment.go` so that the file name gets more specific after each underscore which seems to be common elsewhere?

I do think separating RPC methods into `rpc_broadcaster.go` and `rpc_orchestrator.go` could be helpful for clarify. In my own personal experience, encountering `GetOrchestrator` and `GetOrchestratorInfo` in the same file led to some initial confusion around which one was meant to be called by the orchestrator vs. the broadcaster.

j0sh:
> But, perhaps the file name should be `rpc_segment.go` so that the file name gets more specific after each underscore which seems to be common elsewhere?

`segment_rpc.go` autocompletes better though. We also have `ot_rpc.go` , `lp_rpc.pb.go` and `db_discovery.go` / `discovery.go` so at least that feels more internally consistent.

> I do think separating RPC methods into `rpc_broadcaster.go` and `rpc_orchestrator.go` could be helpful for clarify

Is this in addition to splitting the segment related code in its own file? Seems odd to have both request/response in the segment file but split elsewhere. We can do either, but I'd like to choose one style rather than have a mix.

> encountering `GetOrchestrator` and `GetOrchestratorInfo` in the same file led to some initial confusion

I can see that. Ideally we could rename them, especially `GetOrchestrator` which is looks odd on the O side, but part of that oddity is due to how grpc generates code. Godoc style documentation for public methods would probably help a lot too (eg, "The broadcaster calls GetOrchestratorInfo which invokes GetOrchestrator on the orchestrator").

My preference is still to keep them together (since the network is the boundary between B/O/T and this code joins them together) but we can go with whatever style folks prefer as long as it's consistent and specific.

yondonfu:
> `segment_rpc.go` autocompletes better though. We also have `ot_rpc.go` , `lp_rpc.pb.go` and `db_discovery.go` / `discovery.go` so at least that feels more internally consistent.

Fair point - I didn't realize how prevalent that naming convention already is. In that case, `segment_rpc.go` seems fine.

> Seems odd to have both request/response in the segment file but split elsewhere. We can do either, but I'd like to choose one style rather than have a mix.

Yeah on second thought I agree that we should just be consistent with a single style.

> Godoc style documentation for public methods would probably help a lot too (eg, "The broadcaster calls GetOrchestratorInfo which invokes GetOrchestrator on the orchestrator").

I'm on board with better Godoc style comments for the exported methods! 

> My preference is still to keep them together (since the network is the boundary between B/O/T and this code joins them together) but we can go with whatever style folks prefer as long as it's consistent and specific.

If we separate out the segment related RPC methods and make the usage of RPC methods clearer via better comments, I am ok with not separating based on O vs. B/client vs. server.

angyangie:
Thanks all for the feedback! I made a second attempt at splitting the files based on consensus. I agree that we should be consistent with a single style, and am okay with new naming conventions. I tried to place functions that were called by others in the file, directly below where they were called. That order makes sense to me. Thoughts?

**segment_rpc.go**
StartTranscodeServer
serveSegment
getPayment
getPaymentSender
verifySegCreds

SubmitSegment
genSegCreds
genPayment
genOrchestratorReq

**rpc.go**
_// grpc methods_
ServeHTTP
GetOrchestrator
Ping

CheckOrchestratorAvailability
ping
GetOrchestratorInfo
startOrchestratorClient
getOrchestrator
verifyOrchestratorReq

j0sh:
Looking pretty good, some comments after a quick glance.

`StartTranscodeServer` -- let's put this in `rpc.go` since that's the glue which ties together all the grpc and HTTP calls, including the O/T networking calls.

`genOrchestratorReq` into `rpc.go` since that's only used in `GetOrchestratorInfo`

angyangie:
******NOTE: HAD STRANGE GIT ISSUES, SO OPENED NEW BRANCH:** https://github.com/livepeer/go-livepeer/pull/732

PR #724, opened 2019-02-01
Randomizing orch selection

angyangie:
We are running into some issues with the orchestrator selection algorithm when testing. For example, in a situation with 15 Os and 15Bs, if each B gets a stream at the same time, based on our simple round-robin selection algorithm, they will all likely pick the same O. 

The onchain DB orch list is populated when in `StartMediaServer`, then `gotRTMPStreamHandler`, which ends up calling `selectOrchestrator` and `n.OrchestratorPool.GetOrchestrators(1)`, which grabs orch info asynchronously (`server.GetOrchestratorInfo(ctx, o.bcast, uri)`) and returns  it in the order it was received: https://github.com/livepeer/go-livepeer/blob/master/discovery/discovery.go#L76-L88. Though the first one on that list could more often than not be the same, that's not necessarily the case every time (which is why you might be seeing some variability in test results).

This PR randomizes which orch gets selected, for testing purposes, until we come up with a more complex way of selecting orchs. We have hard-coded the number of orchestrators that get selected (1), and this solution is ONLY works when that hard-code is kept in place: https://github.com/livepeer/go-livepeer/blob/master/server/broadcast.go#L33, and is meant to be a TEMPORARY solution.

**Checklist:**

### DIFF
--- a/discovery/discovery.go
+++ b/discovery/discovery.go
@@ -2,6 +2,7 @@ package discovery
 
 import (
 	"context"
+	"math/rand"
 	"net/url"
 	"strings"
 	"sync"
@@ -20,6 +21,8 @@ type orchestratorPool struct {
 	uris  []*url.URL
 	bcast server.Broadcaster
 }
+
+var perm = func(len int) []int { return rand.Perm(len) }
 
 func NewOrchestratorPool(node *core.LivepeerNode, addresses []string) *orchestratorPool {
 	var uris []*url.URL
@@ -40,8 +43,14 @@ func NewOrchestratorPool(node *core.LivepeerNode, addresses []string) *orchestra
 		glog.Error("Could not parse orchAddresses given - no URIs returned ")
 	}
 
+	var randomizedUris []*url.URL
+	for _, i := range perm(len(uris)) {
+		uri := uris[i]
+		randomizedUris = append(randomizedUris, uri)
+	}
+
 	bcast := core.NewBroadcaster(node)
-	return &orchestratorPool{bcast: bcast, uris: uris}
+	return &orchestratorPool{bcast: bcast, uris: randomizedUris}
 }
 
 func NewOnchainOrchestratorPool(node *core.LivepeerNode) *orchestratorPool {
@@ -93,12 +102,18 @@ func (o *orchestratorPool) GetOrchestrators(numOrchestrators int) ([]*net.Orches
 
 	select {
 	case <-ctx.Done():
-		glog.Info("Done fetching orch info for orchestrators, context timeout: ", orchInfos)
+		respLock.Lock()
+		returnOrchs := orchInfos[:numOrchestrators]
+		respLock.Unlock()
+		glog.Info("Done fetching orch info for orchestrators, context timeout: ", returnOrchs)
 		cancel()
-		return orchInfos, nil
+		return returnOrchs, nil
 	case <-orchChan:
-		glog.Info("Done fetching orch info for orchestrators, numResponses fetched: ", orchInfos)
+		respLock.Lock()
+		returnOrchs := orchInfos[:numOrchestrators]
+		respLock.Unlock()
+		glog.Info("Done fetching orch info for orchestrators, numResponses fetched: ", returnOrchs)
 		cancel()
-		return orchInfos, nil
+		return returnOrchs, nil
 	}
 }
--- a/discovery/discovery_test.go
+++ b/discovery/discovery_test.go
@@ -1,6 +1,7 @@
 package discovery
 
 import (
+	"math/rand"
 	"net/url"
 	"testing"
 
@@ -99,30 +100,27 @@ func TestNewDBOrchestratorPoolCache_GivenListOfOrchs_CreatesPoolCacheCorrectly(t
 func TestNewOrchestratorPoolCache_GivenListOfOrchs_CreatesPoolCacheCorrectly(t *testing.T) {
 	node, _ := core.NewLivepeerNode(nil, "", nil)
 	addresses := []string{"https://127.0.0.1:8936", "https://127.0.0.1:8937", "https://127.0.0.1:8938"}
-	expectedOffchainOrch := StubOrchestratorPool(addresses)
+	expected := []string{"https://127.0.0.1:8938", "https://127.0.0.1:8937", "https://127.0.0.1:8936"}
 	assert := assert.New(t)
-	require := require.New(t)
 
 	// creating NewOrchestratorPool with orch addresses
-	offchainOrch := NewOrchestratorPool(node, addresses)
-	for i, uri := range offchainOrch.uris {
-		assert.Equal(uri.String(), expectedOffchainOrch.uris[i].String())
-	}
+	rand.Seed(321)
+	perm = func(len int) []int { return rand.Perm(3) }
 
-	// creating new OrchestratorPool with different first value
-	addresses[0] = "https://127.0.0.1:89"
-	expectedOffchainOrch = StubOrchestratorPool(addresses)
-	assert.NotEqual(offchainOrch.uris[0].String(), expectedOffchainOrch.uris[0].String())
+	offchainOrch := NewOrchestratorPool(node, addresses)
+
+	for i, uri := range offchainOrch.uris {
+		assert.Equal(uri.String(), expected[i])
+	}
 
 	orchestrators := StubOrchestrators(addresses)
 	node.Eth = &eth.StubClient{Orchestrators: orchestrators}
 
-	expectedRegisteredTranscoders, err := node.Eth.RegisteredTranscoders()
-	require.Nil(err)
-
 	// testing NewOnchainOrchestratorPool
+	rand.Seed(321)
+	perm = func(len int) []int { return rand.Perm(3) }
 	offchainOrchFromOnchainList := NewOnchainOrchestratorPool(node)
 	for i, uri := range offchainOrchFromOnchainList.uris {
-		assert.Equal(uri.String(), expectedRegisteredTranscoders[i].ServiceURI)
+		assert.Equal(uri.String(), expected[i])
 	}
 }


discovery/discovery.go
```diff
@@ -82,7 +83,10 @@ func (o *orchestratorPool) GetOrchestrators(numOrchestrators int) ([]*net.Orches
 			orchInfos = append(orchInfos, info)
 			numSuccessResp++
 		}
-		if numSuccessResp >= numOrchestrators || numResp >= len(o.uris) {
+		if numResp >= len(o.uris) {
+			rand.Seed(time.Now().Unix())
```
j0sh: Don't think we need this here. We might want to control the RNG elsewhere, eg during testing, and this would prevent that.

discovery/discovery.go
```diff
@@ -82,7 +83,10 @@ func (o *orchestratorPool) GetOrchestrators(numOrchestrators int) ([]*net.Orches
 			orchInfos = append(orchInfos, info)
 			numSuccessResp++
 		}
-		if numSuccessResp >= numOrchestrators || numResp >= len(o.uris) {
```
j0sh: Long term, removing the `numSuccessResp >= numOrchestrators` check is problematic because not all orchestrators on the network will respond in a timely manner, which means we have to wait for a time-out before returning from O selection. This impacts stream start-up time and failover behavior.

j0sh (changes requested):
Note that we anticipated this type of issue when first discussing O selection during the RP. The O's are also selected "first come first serve" rather than round-robin.

Does randomizing the initial list before sending out the `GetOrchestrator` request make a difference? Right now, I think they are sent out in the same order across all Bs. If that approach helps, it's preferable.

Another tool to address this would be a capacity cap on O. This allows the network to self-balance which is a much more robust solution over the long term.

discovery/discovery.go
angyangie: That's a good point. I'll add randomization to the initial list. Seems like an easier solution for now

discovery/discovery.go
```diff
@@ -93,12 +110,15 @@ func (o *orchestratorPool) GetOrchestrators(numOrchestrators int) ([]*net.Orches
 
 	select {
 	case <-ctx.Done():
-		glog.Info("Done fetching orch info for orchestrators, context timeout: ", orchInfos)
+		glog.Error("Done fetching orch info for orchestrators, context timeout: ", orchInfos)
 		cancel()
-		return orchInfos, nil
+		return orchInfos[:numOrchestrators], nil
```
yondonfu: Do we need to index into `orchInfos` here? Seems like it would suffice to just return `orchInfos` since `getOrchInfo()` will send a signal into `orchChan` once `numOrchestrators` responses have been received (and `orchInfos` would just be empty if we timed out)

discovery/discovery_test.go
```diff
@@ -96,18 +98,41 @@ func TestNewDBOrchestratorPoolCache_GivenListOfOrchs_CreatesPoolCacheCorrectly(t
...
 }
 
+type stubRpc struct {
+}
+
+func (s *stubRpc) getOrchestrator(ctx context.Context, orchestratorServer *url.URL) (*net.OrchestratorInfo, error) {
+	return &net.OrchestratorInfo{Transcoder: orchestratorServer.String()}, nil
+}
+
+func TestTempGetOrchestrators_GivenOneOrchestratorRequested_ReturnsRandomOrch(t *testing.T) {
```
yondonfu: I don't think this test currently actually checks that a random O is returned since it just checks that an O was returned

discovery/discovery.go
```diff
@@ -40,8 +50,14 @@ func NewOrchestratorPool(node *core.LivepeerNode, addresses []string) *orchestra
...
 
+	var randomizedUris []*url.URL
+	r := rand.New(rand.NewSource(time.Now().Unix()))
+	for _, i := range r.Perm(len(uris)) {
+		uri := uris[i]
+		randomizedUris = append(randomizedUris, uri)
+	}
 	bcast := core.NewBroadcaster(node)
-	return &orchestratorPool{bcast: bcast, uris: uris}
+	return &orchestratorPool{bcast: bcast, uris: randomizedUris}
```
yondonfu, line 53: I wonder how much randomizing the list will help in practice (for example, in the situation where a O is just a little bit faster at responding, it might still end being the selected O every time regardless of the order with which the goroutines for sending requests are created - I do suspect we'll have to implement the O concurrency cap to mitigate this as suggested by @j0sh). I guess trying it out as a quick fix on the test harness will answer that!

discovery/discovery.go
angyangie: The reason we're indexing into orchInfos here, is because often, even if `numOrchestrators` is set to one, multiple orchs are returned. This is what's happening:

> append orch1
> orchChan <- {}{}
> append orch2
> print orch list // now contains 2 orchs even though we only needed 1 

That said, since we do have a `numOrchestrators` expected, wanted to keep the returned # consistent. This is the issue I spent a while trying to figure out on Friday, trying to fix the weirdness with scheduling here. It's not a perfect solution, but ultimately, this solution gives us the outcome we want.

j0sh: Note that it's not really an issue to return **more** orchestrators than what we asked for. However, the reason that we're locking + copying the O list is because the list may still be modified after the fact. With concurrent writes+reads, golang will crash.

That could still happen here even as the context time-out fires, so this clause should also have the same locking+copying treatment, using the original mutex.

discovery/discovery_test.go
angyangie: Yes, this is true. I did give this a lot of thought on Friday.

My initial thinking was that I could just call `NewOrchestratorPool` and `GetOrchestrators(1)` multiple times in a row, compare outcomes, and assure that at least one of the resulting `orchInfo[0].Transcoder` addresses from those calls would be different. But I had some concerns:

- Calling those two functions, three times, in the same test, was yielding the same ordering of URIs for me, always. I have since figured out that `r := rand.New(rand.NewSource(time.Now().Unix()))` could be swapped out for `r := rand.New(rand.NewSource(time.Now().UnixNano()))`, and the resulting`randomizedUris` do indeed come back in a different order.
- Remaining concern: Test might fail sometimes, not because the underlying (non-test) code is faulty. The probability of all three `orchInfo[0].Transcoder` addresses being the same (with a sample size of 3 orchAddr) is real. I could increase the number of `orchAddr` I include here. Is this a valid concern? Any other ideas?

discovery/discovery.go
angyangie: Agree that this is a concern, esp when there aren't that many O's available. Will be interesting to see how often that might happen in practice with the `test-harness`

discovery/discovery.go
```diff
@@ -40,8 +50,14 @@ func NewOrchestratorPool(node *core.LivepeerNode, addresses []string) *orchestra
 		glog.Error("Could not parse orchAddresses given - no URIs returned ")
 	}
 
+	var randomizedUris []*url.URL
+	r := rand.New(rand.NewSource(time.Now().Unix()))
```
j0sh: Let's remove this and just use the default state instead, eg `rand.Perm` . The LivepeerNode initializes the state at startup, and we can control it during testing.

discovery/discovery_test.go
j0sh: We really want tests to be deterministic, hence my concerns earlier in the PR about resetting the RNG state. When we control the seed, that gives us better odds of producing deterministic outputs (as long as other code using the same RNG state is doing so in a predictable manner).

As long as output is consistently in an order somewhat different from the original input then I think it's okay. Yes, this approach is somewhat reminiscent of https://xkcd.com/221/  😄

angyangie: I see what you mean now. Didn't really understand the importance of `rand.Perm` before. I'm pushing that fix in a couple of min!

discovery/discovery.go
```diff
@@ -93,12 +109,18 @@ func (o *orchestratorPool) GetOrchestrators(numOrchestrators int) ([]*net.Orches
...
+		glog.Error("Done fetching orch info for orchestrators, context timeout: ", orchInfos)
 		cancel()
-		return orchInfos, nil
+		return returnOrchs, nil
 	case <-orchChan:
-		glog.Info("Done fetching orch info for orchestrators, numResponses fetched: ", orchInfos)
+		orchChanLock.Lock()
+		returnOrchs := orchInfos[:numOrchestrators]
+		orchChanLock.Unlock()
+		glog.Error("Done fetching orch info for orchestrators, numResponses fetched: ", orchInfos)
```
j0sh: This doesn't seem like an error?

discovery/discovery.go
```diff
@@ -93,12 +109,18 @@ func (o *orchestratorPool) GetOrchestrators(numOrchestrators int) ([]*net.Orches
...
+		respLock.Unlock()
+		glog.Error("Done fetching orch info for orchestrators, context timeout: ", orchInfos)
 		cancel()
-		return orchInfos, nil
+		return returnOrchs, nil
 	case <-orchChan:
-		glog.Info("Done fetching orch info for orchestrators, numResponses fetched: ", orchInfos)
+		orchChanLock.Lock()
+		returnOrchs := orchInfos[:numOrchestrators]
+		orchChanLock.Unlock()
```
j0sh: Should be `respLock`

discovery/discovery_test.go
```diff
@@ -96,18 +98,59 @@ func TestNewDBOrchestratorPoolCache_GivenListOfOrchs_CreatesPoolCacheCorrectly(t
...
+	dbh, dbraw, err := common.TempDB(t)
+	defer dbh.Close()
+	defer dbraw.Close()
```
j0sh: The DB stuff isn't used here; can be removed

discovery/discovery.go
```diff
@@ -40,8 +50,13 @@ func NewOrchestratorPool(node *core.LivepeerNode, addresses []string) *orchestra
...
 	}
 
+	var randomizedUris []*url.URL
+	for _, i := range rand.Perm(len(uris)) {
+		uri := uris[i]
+		randomizedUris = append(randomizedUris, uri)
+	}
 	bcast := core.NewBroadcaster(node)
-	return &orchestratorPool{bcast: bcast, uris: uris}
+	return &orchestratorPool{bcast: bcast, uris: randomizedUris}
```
j0sh, line 53: Need to assign to `rpc` here

discovery/discovery_test.go
```diff
@@ -96,18 +98,59 @@ func TestNewDBOrchestratorPoolCache_GivenListOfOrchs_CreatesPoolCacheCorrectly(t
...
 	expectedOffchainOrch := StubOrchestratorPool(addresses)
 	assert := assert.New(t)
-	require := require.New(t)
 
 	// creating NewOrchestratorPool with orch addresses
 	offchainOrch := NewOrchestratorPool(node, addresses)
-	for i, uri := range offchainOrch.uris {
-		assert.Equal(uri.String(), expectedOffchainOrch.uris[i].String())
-	}
+	assert.Len(offchainOrch.uris, 3)
```
j0sh: Seed the default RNG with a constant value before this point (but after LivepeerNode is initialized). Then see what order the O list is initialized. Update the expected offchain orch list to match. That preserves the character of this test -- just checking for lengths doesn't offer as strong of an assurance.

```diff
diff --git a/discovery/discovery_test.go b/discovery/discovery_test.go
index c5c5d979..78f4aebf 100644
--- a/discovery/discovery_test.go
+++ b/discovery/discovery_test.go
@@ -2,6 +2,7 @@ package discovery
 
 import (
 	"context"
+	"math/rand"
 	"net/url"
 	"testing"
 
@@ -145,29 +146,25 @@ func TestTempGetOrchestrators_GivenOneOrchestratorRequested_ReturnsRandomOrch(t
 func TestNewOrchestratorPoolCache_GivenListOfOrchs_CreatesPoolCacheCorrectly(t *testing.T) {
 	node, _ := core.NewLivepeerNode(nil, "", nil)
 	addresses := []string{"https://127.0.0.1:8936", "https://127.0.0.1:8937", "https://127.0.0.1:8938"}
-	expectedOffchainOrch := StubOrchestratorPool(addresses)
+	expected := []string{"https://127.0.0.1:8938", "https://127.0.0.1:8937", "https://127.0.0.1:8936"}
 	assert := assert.New(t)
 
 	// creating NewOrchestratorPool with orch addresses
+	rand.Seed(321) // reset rng
 	offchainOrch := NewOrchestratorPool(node, addresses)
-	assert.Len(offchainOrch.uris, 3)
 
 	// creating new OrchestratorPool with different first value
-	addresses[0] = "https://127.0.0.1:89"
-	expectedOffchainOrch = StubOrchestratorPool(addresses)
-	assert.NotEqual(offchainOrch.uris[0].String(), expectedOffchainOrch.uris[0].String())
+	for i, uri := range offchainOrch.uris {
+		assert.Equal(uri.String(), expected[i])
+	}
 
 	orchestrators := StubOrchestrators(addresses)
 	node.Eth = &eth.StubClient{Orchestrators: orchestrators}
 
 	// testing NewOnchainOrchestratorPool
+	rand.Seed(321) // reset rng
 	offchainOrchFromOnchainList := NewOnchainOrchestratorPool(node)
-	newAddressExists := false
-	for _, uri := range offchainOrchFromOnchainList.uris {
-		if uri.String() == addresses[0] {
-			newAddressExists = true
-		}
+	for i, uri := range offchainOrchFromOnchainList.uris {
+		assert.Equal(uri.String(), expected[i])
 	}
-
-	assert.True(newAddressExists)
 }
``` 

With this, I don't believe `TestTempGetOrchestrators_GivenOneOrchestratorRequested_ReturnsRandomOrch` is needed anymore since we're also testing a randomized list as part this function.

Perhaps a slightly saner method would be to make the permutation monkey-patchable. Have tests use a monkey-patched function returning a fixed sequence (say the original input without any shuffling) and the actual RNG can be tested in isolation, in a manner similar to what was outlined above.

j0sh (changes requested):

discovery/discovery_test.go
angyangie: Interesting! Thank you for the suggestion. Setting RNG with a constant value makes the test truly deterministic, and allows for checking of randomness simultaneously. No need to monkey-patch for the other test.

yondonfu (approved):
LGTM

discovery/discovery.go
```diff
@@ -93,12 +99,18 @@ func (o *orchestratorPool) GetOrchestrators(numOrchestrators int) ([]*net.Orches
...
+		respLock.Unlock()
 		glog.Info("Done fetching orch info for orchestrators, context timeout: ", orchInfos)
 		cancel()
-		return orchInfos, nil
+		return returnOrchs, nil
 	case <-orchChan:
+		respLock.Lock()
+		returnOrchs := orchInfos[:numOrchestrators]
+		respLock.Unlock()
 		glog.Info("Done fetching orch info for orchestrators, numResponses fetched: ", orchInfos)
```
j0sh: There's still the possibility of `orchInfos` mutation here so the log should print `returnOrchs`

j0sh (approved):
LGTM 🚢

Issue #695, opened 2019-01-18
Transcode loop is indexed by ManifestID


j0sh:
**Describe the bug**

The ManifestID is broadcaster generated (and may be user supplied). The ManifestID is currently used for two reasons:

* To index streams on the broadcaster for playback
* To index into the orchestrator's transcoding loop

The ManifestID should not be used by the orchestrator in any capacity. There are a few reasons for this:

* Unfavorable behavior within the S3 storage layer with the ManifestID at the root of it all. See [this writeup](https://hackmd.io/0C2Zi0VVSyGEgN_4lEu_WA) for mind-numbing detail on the subject. In short: for O-owned storage, we store source segments in a different location from transcoded results, which generally hinders UX / debugging / forensics / other data collation.
* Minimizes the space of "funny values" supplied by the broadcaster causing issues on the orchestrator (eg, injection). There doesn't seem to be any risk of this now, but doesn't mean there never will be if ManifestID usage proliferates on the orchestrator.
* Orchestrators may be working with multiple broadcasters, increasing the risk of a transcode-loop collision where multiple streams share the same ManifestID. Collisions on the transcoding loop are problematic for a few reasons:
  * The transcode loop itself functions as a backpressure mechanism. If an orchestrator is processing a segment while another comes in, it's likely that the O is unable to keep up with live video. Returning a busy signal here makes the broadcaster fail over elsewhere, depriving the O of legitimate work.
  * Extending the transcode loop beyond its natural expiration also lengthens the time before PM tickets are redeemed.
  * Violating the invariant of "one stream to one transcode loop" may complicate things such as pixel accounting unless we are extra careful.

**Desired Solution**

Collisions and funny values could be mitigated by making any relevant bookkeeping per-broadcaster and hashing the ManifestID. For example, making the transcode loop index `H(<broadcasterAddress> | <manifestID>)` . Even so, this does not address the lingering issues around S3 storage.

Restore the "authentication token" to preserve stateless continuity between GetOrchestrator and segment transcoding. This mitigates all the issues we have on the orchestrator stemming from the Manifest ID.

Here is the commit removing the AuthToken feature: https://github.com/livepeer/go-livepeer/commit/fae70f384ab151b2f2c50495f3c406c706ef0412

The auth token works like this:

```
type AuthToken struct {
  LoopIndex int
  Sig bytes
}
token = AuthToken{ LoopIndex: loopIndex, Sig: Sign(H(loopIndex)) }
```

[...]

PR #729, opened 2019-02-05
Orchestrator capacity cap

### LINKED ISSUE #603: O should respond with an InsufficientCapacity when B attempts a new job/segment negotiation and O is at max capacity
As we are thinking about off-chain negotiation messages between B and O, in a naive implementation we would not be able to tell apart Os that are offline/unresponsive vs responsive Os that are aware they are at max capacity and cannot take on additional work.

It seems important to tell these scenarios apart, to allow the good Os to maintain a strong reputation and avoid seeming offline when they are not.

B and O should both support the handling of an `InsufficientCapacity` response O can send B when they are negotiating work. Perhaps in future version this message could also include some heuristic prediction by O of when some capacity should free up.

Suggested naive implementation: the node operator can set a max concurrent streams flag, as mentioned in #706 .

The naive alternative is described above - O simply does not respond and B needs to have some timeout and then try other Os.

**Additional context**
None to add.

j0sh:
Implements an orchestrator capacity cap. Most of the interesting changes are here: https://github.com/livepeer/go-livepeer/commit/70342afc9cdd8599241271aa2b0066c23577ddd7

- Adds a `-maxSession` CLI parameter
- Checks the capacity during GetOrchInfo (so B doesn't send segments over unnecessarily), during seg creds checking (prior to download) and prior to actually transcoding (also atomic to creating a new transcode loop).
- Reduces transcode loop timeout from 10 minutes to 1 minute. The previous value of 10 minutes was set in order to give plenty of time for retries before submitting an on-chain transaction to claim payment. Here, there is no such concern, and we want to release resources as soon as possible if a transcode session shows signs of being idle.
- Add tests around the capacity cap
- Fixes an older, separate test case that wasn't really checking the right thing.

Unit testing, manual testing.

Fixes https://github.com/livepeer/go-livepeer/issues/603

**Checklist:**

### DIFF
--- a/server/broadcast.go
+++ b/server/broadcast.go
@@ -226,7 +226,7 @@ func processSegment(cxn *rtmpConnection, seg *stream.HLSSegment) {
 	}()
 }
 
-var sessionErrStrings = []string{"dial tcp", "unexpected EOF", core.ErrOrchBusy.Error()}
+var sessionErrStrings = []string{"dial tcp", "unexpected EOF", core.ErrOrchBusy.Error(), core.ErrOrchCap.Error()}
 
 func generateSessionErrors() *regexp.Regexp {
 	// Given a list [err1, err2, err3] generates a regexp `(err1)|(err2)|(err3)`
--- a/server/rpc.go
+++ b/server/rpc.go
@@ -56,6 +56,7 @@ type Orchestrator interface {
 	Sign([]byte) ([]byte, error)
 	VerifySig(ethcommon.Address, string, []byte) bool
 	CurrentBlock() *big.Int
+	CheckCapacity(core.ManifestID) error
 	TranscodeSeg(*core.SegTranscodingMetadata, *stream.HLSSegment) (*core.TranscodeResult, error)
 	ServeTranscoder(stream net.Transcoder_RegisterTranscoderServer)
 	TranscoderResults(job int64, res *core.RemoteTranscoderResult)
@@ -135,7 +136,7 @@ func verifyOrchestratorReq(orch Orchestrator, addr ethcommon.Address, sig []byte
 		glog.Error("orchestrator req sig check failed")
 		return fmt.Errorf("orchestrator req sig check failed")
 	}
-	return nil
+	return orch.CheckCapacity("")
 }
 
 func genSegCreds(sess *BroadcastSession, seg *stream.HLSSegment) (string, error) {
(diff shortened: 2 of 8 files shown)


server/rpc.go
```diff
@@ -135,7 +136,7 @@ func verifyOrchestratorReq(orch Orchestrator, addr ethcommon.Address, sig []byte
 		glog.Error("orchestrator req sig check failed")
 		return fmt.Errorf("orchestrator req sig check failed")
 	}
-	return nil
+	return orch.CheckCapacity("")
```
yondonfu, line 139: If any B uses an empty manifestID which is used to index into `orch.node.SegmentChans`, this capacity check would never return an error as long as the transcode loop for the empty manifestID is active right? Though even if this is true, I don't think you can do much else besides bypass the capacity check during `GetOrchestratorInfo` since the capacity checks during seg creds verification and prior to transcoding should still return errors

server/broadcast.go
```diff
@@ -226,7 +226,7 @@ func processSegment(cxn *rtmpConnection, seg *stream.HLSSegment) {
 	}()
 }
 
-var sessionErrStrings = []string{"dial tcp", "unexpected EOF", core.ErrOrchBusy.Error()}
+var sessionErrStrings = []string{"dial tcp", "unexpected EOF", core.ErrOrchBusy.Error(), core.ErrOrchCap.Error()}
```
yondonfu, line 229: Should we update `TestStopSessionErrors()` in `broadcast_test.go` to check for this new error?

server/rpc.go
j0sh: > this capacity check would never return an error as long as the transcode loop for the empty manifestID is active right

Correct, just as it would not return an error if the transcode loop was active for any non-empty ManifestID. So it's not bypassing any checks beyond the possibility of hijacking the loop for a somehow ill-formed ManifestID.

Although this is mostly harmless, it's still another example of why I'd prefer not to use the Manifest ID as an index to the transcoding loop as mentioned in https://github.com/livepeer/go-livepeer/issues/695 .

server/broadcast.go
j0sh: It's actually already updated. The test is designed such that it'll fail if we add a new error value but forget to update the test 😄 See:


https://github.com/livepeer/go-livepeer/pull/729/commits/70342afc9cdd8599241271aa2b0066c23577ddd7#diff-f986f75ea9a0ee09429050246648e0b7R17

yondonfu: Oh weird I don't know how I missed that...oops 😓

yondonfu (approved):
LGTM 🚢

darkdarkdragon (approved):
LGTM

PR #731, opened 2019-02-05
eth: add context with timeout to eth_getLogs queries

yondonfu:

Added a context with timeout to methods that create an eth_getLogs RPC query. Previously, we did not attach a timeout to these queries. The default timeout is currently 20 seconds.

- Added a context with timeout to `client.ProcessHistoricalUnbond`, `client.ProcessHistoricalRebond` and `client.ProcessHistoricalWithdrawStake`.
- Also updated unbondingservice.go to proceed even if processing of
historical events fails.

I built the node in the devenv VM and adjusted the timeout to 2 nanoseconds and observed a log message indicating failure to process historical events and the node continued booting up.

**Checklist:**

### DIFF
--- a/eth/client.go
+++ b/eth/client.go
@@ -40,6 +40,8 @@ var (
 	ErrReplacingMinedTx   = fmt.Errorf("trying to replace already mined tx")
 	ErrCurrentRoundLocked = fmt.Errorf("current round locked")
 	ErrMissingBackend     = fmt.Errorf("missing Ethereum client backend")
+
+	RPCTimeout = 20 * time.Second
 )
 
 type LivepeerEthClient interface {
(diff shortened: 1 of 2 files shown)


eth/client.go
```diff
@@ -876,7 +878,13 @@ func (c *client) LatestBlockNum() (*big.Int, error) {
...
 	// Retrieve historical logs starting from startBlock
 	// WatchForUnbond() will not emit past logs
-	filterOpts := &bind.FilterOpts{Start: startBlock.Uint64()}
+	ctx, cancel := context.WithTimeout(context.Background(), RPCTimeout)
+	defer cancel()
+
+	filterOpts := &bind.FilterOpts{
+		Start:   startBlock.Uint64(),
+		Context: ctx,
+	}
```
j0sh: Could these be factored out into a helper with the sig `func (startBlock *big.Int) *bind.FilterOpts` ?

j0sh (approved):
Other than possibly incorporating a helper for the filter options, LGTM

eth/client.go
yondonfu: Updated to use a `filterOptsWithTimeout()` helper

PR #732, opened 2019-02-06
adding segment_rpc and rpc

angyangie:
******NOTE: HAD STRANGE GIT ISSUES, SO OPENED NEW BRANCH. THIS IS OLD BRANCH:********* https://github.com/livepeer/go-livepeer/pull/723

*************

https://github.com/livepeer/go-livepeer/issues/715

- Split server and client code from rpc file
- Organize functions by order of where they're called in the code. Functions called by another function within the same file are placed below caller function

https://github.com/livepeer/go-livepeer/issues/715

**Checklist:**

### DIFF
--- a/server/rpc.go
+++ b/server/rpc.go
@@ -1,26 +1,19 @@
 package server
 
 import (
-	"bytes"
 	"context"
-	"crypto/tls"
-	"encoding/base64"
 	"fmt"
-	"io/ioutil"
 	"math/big"
 	"net/http"
 	"net/url"
 	"strings"
 	"time"
 
-	"golang.org/x/net/http2"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
 
-	"github.com/livepeer/go-livepeer/common"
 	"github.com/livepeer/go-livepeer/core"
 	"github.com/livepeer/go-livepeer/drivers"
-	"github.com/livepeer/go-livepeer/monitor"
 	"github.com/livepeer/go-livepeer/net"
 	"github.com/livepeer/go-livepeer/pm"
 	ffmpeg "github.com/livepeer/lpms/ffmpeg"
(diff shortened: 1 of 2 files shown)


j0sh (commented):
The changes themselves LGTM. Might be nice to add godoc style comments around major entry points to alleviate some of the confusion mentioned by @yondonfu

darkdarkdragon (commented):
For me it is no better than it was originally. Keeping code that implements both sides of rpc call together is really confusing and doesn't help to understand control flow.

angyangie:
Thanks for voicing your opinion @darkdarkdragon. My thinking was to keep some of the segment-specific processing code in `segment_rpc.go`. So, in that file, we have `ServeSegment`, which calls `getPayment`, `getPaymentSender`, and `verifySegCreds`. We then have `SubmitSegment` which calls `genSegCreds` and `genPayment`. I tried to organize the functions in order of when they were called. 

All other functions went in `rpc.go`,  but there again I tried to in general keep functions that called other functions together. At the top is `ServeHTTP` and all the `grpc` methods,  `GetOrchestrator`, `Ping`. `StartTranscodeServer` doesn't call any other function in that file (though it does call `ServeSegment` in `segment_rpc.go`... and I'd be fine moving it there). Then we have `CheckOrchestratorAvailability` which calls `ping. Then, `GetOrchestratorInfo` and all the methods it calls, in order: `GetOrchestratorInfo`, `startOrchestratorClient`, `genOrchestratorReq`, `getOrchestrator` and `verifyOrchestratorReq`.

Is there anything that is particularly worrisome to you about the change, or anything you think might make you feel more comfortable with it? How strongly do you feel about this?

darkdarkdragon:
For me, it just doesn't fixes orginal issue - I brought it up to split code specifically by the network boundary.
When I'm working with file that have code from both sides of rpc, I need to make special mental effort to keep in mind overall code flow. For me it would be much easier to work with code when it is separated by files by network boundary.

j0sh:
Funny, because I actually like how it joins the components across the network boundary. Network roles are important to be aware of, but dividing code strictly along network boundaries makes it less clear what is related from a functional perspective: at what stage does this bit of code come into play? Which request produces which response?

@darkdarkdragon What makes it difficult to keep track of the overall flow right now, other than the file name? Is it because of the function placement in the file? Unclear naming? Lack of comments / documentation? Are there any other steps we can take with this approach to make things easier?

darkdarkdragon:
@j0sh I usually working on application level logic, say, broadcaster's one, and thus I don't need to follow control flow across network boundary - for the broadcaster's logic it is not important what happens inside orchestrator. And when I want to follow across network boundary - I would prefer to have other side of network inside other file - it is easier to switch tabs in the editor then to jump back and forth through one long file.

And another thing - this pattern:
```golang
func (h *lphttp) GetOrchestrator(context context.Context, req *net.OrchestratorRequest) (*net.OrchestratorInfo, error) {
	return getOrchestrator(h.orchestrator, req)
}

func getOrchestrator(orch Orchestrator, req *net.OrchestratorRequest) (*net.OrchestratorInfo, error) {
```
makes code harder to understand - every time I see `getOrchestrator` function, I need to go and check if it is used somewhere else or not.

j0sh:
> for the broadcaster's logic it is not important what happens inside orchestrator

At this point, when you're working on one side of the network or another, does it matter which file the code is in? They are still functionally related.

> it is easier to switch tabs in the editor then to jump back and forth through one long file.

Does your editor allow to have multiple views of the same file? In vim, this is actually less work; just split the window and get to work.

image

> And another thing - this pattern:

This particular pattern was introduced to minimize the boilerplate around testing. Send a PR if you'd prefer another way.

@darkdarkdragon Anyway this is as much a matter of style as anything. If there are no concrete suggestions for improvement around this PR then perhaps @angyangie can add godocs and we'll move on from here. Perhaps the experience for you will improve with some better documentation. This type of maintenance should be viewed as an ongoing process, not a final goal. As features are added, removed or moved around then we should periodically take the opportunity to review how things are organized.

darkdarkdragon (approved):
LGTM

PR #737, opened 2019-02-07
Debian based docker image for using in test harness

darkdarkdragon:
Debian based docker image for using in test harness
Also updated nasm to 2.14.02, x264 to current master and ffmpeg to 4.1 release

New docker image

Manually inside docker

**Checklist:**

### DIFF
--- a/install_ffmpeg.sh
+++ b/install_ffmpeg.sh
@@ -5,7 +5,7 @@ export PKG_CONFIG_PATH=$HOME/compiled/lib/pkgconfig
 
 if [ ! -e "$HOME/nasm/nasm" ]; then
   # sudo apt-get -y install asciidoc xmlto # this fails :(
-  git clone -b nasm-2.13.02 http://repo.or.cz/nasm.git "$HOME/nasm"
+  git clone -b nasm-2.14.02 https://repo.or.cz/nasm.git "$HOME/nasm"
   cd "$HOME/nasm"
   ./autogen.sh
   ./configure --prefix="$HOME/compiled"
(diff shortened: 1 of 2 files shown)


install_ffmpeg.sh
```diff
@@ -16,15 +16,13 @@ fi
 if [ ! -e "$HOME/x264/x264" ]; then
   git clone http://git.videolan.org/git/x264.git "$HOME/x264"
   cd "$HOME/x264"
-  # git master as of this writing
-  git checkout 7d0ff22e8c96de126be9d3de4952edd6d1b75a8c
```
j0sh, line 20: For now, let's keep checking out a consistent hash because of the bit-exact verification requirement. Once we're confident in non-determinisic verification then we work directly from master.

install_ffmpeg.sh
```diff
@@ -16,15 +16,13 @@ fi
...
   ./configure --prefix="$HOME/compiled" --enable-pic --enable-static
   make
   make install-lib-static
 fi
 
 if [ ! -e "$HOME/ffmpeg/libavcodec/libavcodec.a" ]; then
-  git clone -b n4.0 https://git.ffmpeg.org/ffmpeg.git "$HOME/ffmpeg" || echo "FFmpeg dir already exists"
+  git clone -b release/4.1 https://git.ffmpeg.org/ffmpeg.git "$HOME/ffmpeg" || echo "FFmpeg dir already exists"
```
j0sh: Same comment as above; let's track tags (which should be immutable) rather than release branches (which can change). There's a `n4.1` tag that we can use.

j0sh (changes requested):
Would be good to have the changes to install_ffmpeg also reflected in LPMS and devenv.

install_ffmpeg.sh
darkdarkdragon: fixed

install_ffmpeg.sh
darkdarkdragon: fixed

j0sh (approved):
LGTM

Issue #710, opened 2019-01-24
Clean Up livepeer.go Initialization


j0sh:
**Describe the bug**
Generally, how we handle flags needs a lot of cleanup. Rather than flag parsing or the number of flags itself, this issue is more around how we check and use those flags in `cmd/livepeer` to initialize the node.

The code interleaves between initialization for orchestrators, broadcasters and both, often switching back-and-forth. I am not suggesting to split `livepeer` into separate executables, but perhaps the organization within `livepeer.go` could be improved. This would also make it easier to reason about the behavior of the node under various combinations of flags.

Related issues:

* The code setting the network config is difficult to follow and could be improved.  https://github.com/livepeer/go-livepeer/issues/580

* Might be nice to incorporate a config file along with this cleanup, if the work breakdown ends up making sense. Related: https://github.com/livepeer/go-livepeer/issues/101

* The flags themselves could use a lot of love too. https://github.com/livepeer/go-livepeer/issues/708 for docs and https://github.com/livepeer/go-livepeer/issues/709 for improving consistency.

Issue #709, opened 2019-01-24
CLI Flags Cleanup


j0sh:
**Is your feature request related to a problem? Please describe.**
With all the changes to the node for Streamflow it might be a good opportunity to make breaking changes to the command line flags.

The first question is whether we should do this. If we decide to try and maintain backwards compatibility with flags, we can close this issue.

There are a few issues with the flags, which would be nice to resolve without being concerned with backwards compatibility:
    * Unused flags: maxPricePerSegment, ipfsPath, noipfsLogFiles
    * Some flags that might make sense to combine: ethIpcPath and ethUrl? 
    * Flags that should be renamed for consistency with the other flag names: monhost -> monAddr?
    * Flags that should be renamed for behavioral consistency: `-standaloneTranscoder` -> `-transcoder`
    * Adding a`-broadcaster` flag?
    * Any other miscellaneous changes that might be good to make sooner rather than later? Eg, if someone objects to the `*Addr` naming scheme then now is the time to change it.

Similar review is probably needed for livepeer_cli flags handling.

j0sh:
Responding to the non-network part of https://github.com/livepeer/go-livepeer/issues/580#issuecomment-458283965

> Could we autogenerate an orchSecret for a user, instead of having them pass in a random phrase?

We can autogenerate, but we should still allow users to specify their own secret. If autogenerate, we need an easy way for the user to put the secret on transcoder nodes, eg by copy-pasting. That seems like a separate feature, so perhaps hold that for now.

> Could we just check the presence of transcoder and orchestrator flags set to true to determine whether user is running in both transc and orch mode

Yep, the idea of flag composition is good.
  * Run with `-transcoder` : standalone transcoder
  * Run with `-orchestrator` : standalone orchestrator
  * Run with `-orchestrator -transcoder` : combined O+T

> Could we initialize round based on the presence of a transcoder or orchestrator flag and a testNet flag

We should still keep the `-initializeRound` flag since it's important to have that be configurable, especially for mainnet. We can certainly set round initialization automatically in test environments (as long as it can still be disabled), perhaps that can be added in after the cleanup is done.

One thing we want to be mindful of is having too many options implicitly toggle depending on the values of other flags. This is somewhat understandable for the network config (such as host names), but I can see it becoming confusing with other types of options (eg, `-initializeRound` being false by default but true in devenv or rinkeby).

[...]

Issue #580, opened 2018-10-22
Clean up Network Config


j0sh:
**Is your feature request related to a problem? Please describe.**
The network configuration code is getting unwieldy. There are many hard-coded values and uncertain interactions between the options. For example, offchain mode is mutually exclusive with mainnet and rinkeby, but this is not obvious from reading the code.

https://github.com/livepeer/go-livepeer/blob/bbb568ffdc1c4dfd0f92a4f3c9e31e81aab5c285/cmd/livepeer/livepeer.go#L91-L118

**Describe the solution you'd like**
More structure around our network configuration options. This could even be split out into a separate configuration file, for example:

```ini
  name = mainnet
  datadir = /path/to/custom/datadir # this could be optional; default being ~/.lpData/<network-name>
  controllerAddr = 0xabcdef
  ethUrl = wss://mainnet.infura.io/ws
  metricsUrl = http://mainnet-metrics.livepeer.org
```

**Describe alternatives you've considered**
Keep things hard-coded but use a typed data structure for it, eg
```go
type NetworkConfig struct {
  name string
  datadir string
  controllerAddr string
  ...
}
networkConfigs["mainnet"] = { name: "mainnet", ... }
```

eladmallel:
@j0sh I think we should close this issue and merge its specs into this one: https://github.com/livepeer/go-livepeer/issues/537

wdyt?

j0sh:
These are different issues; #537 is solved by putting all network specific data into its own folder (https://github.com/livepeer/go-livepeer/pull/578), while this one is about cleaning up the code around network configuration.

j0sh:
Before anything is implemented for a config file (if that's the direction we go with), it's probably worth discussing the format that should be used. TOML / ini-style looks good to me, for what it's worth. This library seems nice: https://github.com/BurntSushi/toml

angyangie:
It'd be great to get everyone's thoughts on the following flag clean-up. As suggested, we'll focus on cleaning up the flags before introducing a config file + a new flag parsing library (though this one seems like a great candidate: https://github.com/octago/sflags). To make analysis easier to digest, I've grouped flags by function, and have added notes below, along with questions (forgive my ignorance on some of these):

[...]

PR #738, opened 2019-02-07
Clean up flag values in livepeer.go

angyangie:
This is the first step in this issue: https://github.com/livepeer/go-livepeer/issues/710. It is a halfway step to adding a config file and different flag parsing code to `livepeer.go`.

Generally, how we handle flags needs a lot of cleanup. Rather than flag parsing or the number of flags itself, this issue is more around how we check and use those flags in `cmd/livepeer` to initialize the node. The code interleaves between initialization for orchestrators, broadcasters and both, often switching back-and-forth.

Discussion on flag clean-up is here:
https://github.com/livepeer/go-livepeer/issues/709
https://github.com/livepeer/go-livepeer/issues/580

- Grouped flags together based on general function (makes the file more readable)
- Added a `NetworkConfig` struct to hold network config options like `ethUrl`, `ethController`, and `monHost`. If more network specific options are added in the future, we could add them here.
- Removed `maxPricePerSegment` (unused flag)
- Combined `ethIpcPath` and `ethUrl`
- Renamed `noipfsLogFiles` to `logIpfs`
- Renamed `monhost` to `monHost`
- Made `transcoder` var into a flag,  removed `standaloneTranscoder`
- Removed `orchAndTrans`. Temporarily setting orchs and transc to be both, here: 
```golang
if *transcoder || *orchestrator {
		*transcoder = true
		*orchestrator = true
	}
```
- Added `-network` flag. Removed `-rinkeby`, `devenv`, `offchain` flags. If network isn't set, `-network offchain` is assumed, regardless of what other options might have been specified. If network is set, but not `rinkeby` or `mainnet`, network remains as was specified. (REVIEW)
- Edited devtool script to work with all changes

**To do:**
- [x] (TBD) Add a `-broadcaster` flag for consistency 
- [x] (Must review) No way for user to operate in `offchain` mode and to name their own network
- [x] (TBD) Allow users to set their own `datadir` name, without network subfolder?
- [ ] More testing on different networks

**Will not do:**
Remove `ipfsPath`, `noipfsLogFiles`, unused flags - a lot of code was already written in anticipation of these flags. Since it seems like these flags will be used in the near future, I vote for keeping them commented out. 

I can successfully transcode using devtool. All tests passed. 

https://github.com/livepeer/go-livepeer/issues/709
https://github.com/livepeer/go-livepeer/issues/580

**Checklist:**

### DIFF
--- a/cmd/devtool/devtool.go
+++ b/cmd/devtool/devtool.go
@@ -32,8 +32,8 @@ var (
 	endpoint                  = "ws://localhost:8546/"
 	gethMiningAccount         = "87da6a8c6e9eff15d703fc2773e32f6af8dbe301"
 	gethMiningAccountOverride = false
-	controllerAddr            = "0x04B9De88c81cda06165CF65a908e5f1EFBB9493B"
-	controllerAddrOverride    = false
+	ethController             = "0x04B9De88c81cda06165CF65a908e5f1EFBB9493B"
+	ethControllerOverride     = false
 )
 
 func main() {
@@ -41,7 +41,7 @@ func main() {
 	baseDataDir := flag.String("datadir", ".lpdev2", "default data directory")
 	endpointAddr := flag.String("endpoint", "", "Geth endpoint to connect to")
 	miningAccountFlag := flag.String("miningaccount", "", "Override geth mining account (usually not needed)")
-	controllerAddrFlag := flag.String("controller", "", "Override controller address (usually not needed)")
+	ethControllerFlag := flag.String("controller", "", "Override controller address (usually not needed)")
 
 	flag.Parse()
 	if *endpointAddr != "" {
@@ -51,9 +51,9 @@ func main() {
 		gethMiningAccount = *miningAccountFlag
 		gethMiningAccountOverride = true
 	}
-	if *controllerAddrFlag != "" {
-		controllerAddr = *controllerAddrFlag
-		controllerAddrOverride = true
+	if *ethControllerFlag != "" {
+		ethController = *ethControllerFlag
+		ethControllerOverride = true
 	}
 	args := flag.Args()
 	goodToGo := false
@@ -120,10 +120,10 @@ func ethSetup(ethAcctAddr, keystoreDir string, isBroadcaster bool) {
 		glog.Errorf("Failed to connect to Ethereum client: %v", err)
 		return
 	}
-	glog.Infof("Using controller address %s", controllerAddr)
+	glog.Infof("Using controller address %s", ethController)
 
 	client, err := eth.NewClient(ethcommon.HexToAddress(ethAcctAddr), keystoreDir, backend,
-		ethcommon.HexToAddress(controllerAddr), ethTxTimeout)
+		ethcommon.HexToAddress(ethController), ethTxTimeout)
 	if err != nil {
 		glog.Errorf("Failed to create client: %v", err)
 		return
@@ -253,13 +253,13 @@ func ethSetup(ethAcctAddr, keystoreDir string, isBroadcaster bool) {
 
 func createRunScript(ethAcctAddr, dataDir string, isBroadcaster bool) {
 	script := "#!/bin/bash\n"
-	script += fmt.Sprintf(`./livepeer -v 99 -controllerAddr %s -datadir ./%s \
+	script += fmt.Sprintf(`./livepeer -v 99 -ethController %s -datadir ./%s \
     -ethAcctAddr %s \
     -ethUrl %s \
     -ethPassword "" \
-    -gasPrice 200 -gasLimit 2000000 -devenv=true \
+    -gasPrice 200 -gasLimit 2000000 -network=devenv \
     -monitor=false -currentManifest=true `,
-		controllerAddr, dataDir, ethAcctAddr, endpoint)
+		ethController, dataDir, ethAcctAddr, endpoint)
 
 	if !isBroadcaster {
 		script += fmt.Sprintf(` -initializeRound=true \
@@ -331,9 +331,9 @@ func remoteConsole(destAccountAddr string) error {
 	}
 	defer console.Stop(false)
 
-	if !controllerAddrOverride {
+	if !ethControllerOverride {
 		// f9a6cf519167d81bc5cb3d26c60c0c9a19704aa908c148e82a861b570f4cd2d7 - SetContractInfo event
-		getControllerAddressScript := `
+		getethControllerScript := `
 		var logs = [];
 		var filter = web3.eth.filter({ fromBlock: 0, toBlock: "latest",
 			topics: ["0xf9a6cf519167d81bc5cb3d26c60c0c9a19704aa908c148e82a861b570f4cd2d7"]});
@@ -342,17 +342,17 @@ func remoteConsole(destAccountAddr string) error {
 		});
 		console.log(logs[0][0].address);''
 	`
-		glog.Infof("Running eth script: %s", getControllerAddressScript)
-		err = console.Evaluate(getControllerAddressScript)
+		glog.Infof("Running eth script: %s", getethControllerScript)
+		err = console.Evaluate(getethControllerScript)
 		if err != nil {
 			glog.Error(err)
 		}
 		if printer.Len() == 0 {
 			glog.Fatal("Can't find deployed controller")
 		}
-		controllerAddr = strings.Split(printer.String(), "\n")[0]
+		ethController = strings.Split(printer.String(), "\n")[0]
 
-		glog.Infof("Found controller address: %s", controllerAddr)
+		glog.Infof("Found controller address: %s", ethController)
 	}
 
 	script := fmt.Sprintf("eth.sendTransaction({from: \"%s\", to: \"%s\", value: web3.toWei(834, \"ether\")})",
--- a/cmd/devtool/devtool.go
+++ b/cmd/devtool/devtool.go
@@ -87,13 +87,12 @@ func main() {
 	acc := createKey(tempKeystoreDir)
 	glog.Infof("Using account %s", acc)
 	dataDir := filepath.Join(*baseDataDir, t+"_"+acc)
-	dataDirToCreate := filepath.Join(dataDir, "devenv")
-	err = os.MkdirAll(dataDirToCreate, 0755)
+	err = os.MkdirAll(dataDir, 0755)
 	if err != nil {
 		glog.Fatalf("Can't create directory %v", err)
 	}
 
-	keystoreDir := filepath.Join(dataDirToCreate, "keystore")
+	keystoreDir := filepath.Join(dataDir, "keystore")
 	err = os.Rename(tempKeystoreDir, keystoreDir)
 	if err != nil {
 		glog.Fatal(err)
@@ -263,9 +262,13 @@ func createRunScript(ethAcctAddr, dataDir string, isBroadcaster bool) {
 
 	if !isBroadcaster {
 		script += fmt.Sprintf(` -initializeRound=true \
-    -serviceAddr 127.0.0.1:8936 -httpAddr 127.0.0.1:8936  -transcoder \
+    -serviceAddr 127.0.0.1:8936 -httpAddr 127.0.0.1:8936  -transcoder=true -orchestrator=true \
     -cliAddr 127.0.0.1:7936 -ipfsPath ./%s/trans
     `, dataDir)
+	}
+
+	if isBroadcaster {
+		script += fmt.Sprint(` -broadcaster=true`)
 	}
 
 	glog.Info(script)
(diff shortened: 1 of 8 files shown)


cmd/livepeer/livepeer.go
```diff
@@ -130,47 +153,30 @@ func main() {
...
-			}
-			glog.Infof("***Livepeer is running on the Ethereum main network: %v***", MainnetControllerAddr)
-			*datadir = *datadir + "/mainnet"
-		} else {
-			*datadir = *datadir + "/offchain"
-		}
-		if *monitor && *monhost == "" {
-			*monhost = "http://metrics-mainnet.livepeer.org/api/events"
-		}
+		*network = "offchain"
```
angyangie: Note: need to add ability for users to name their own network, and still be "offchain"

cmd/livepeer/livepeer.go
```diff
@@ -60,67 +58,92 @@ func main() {
...
+			monHost:       "http://metrics-rinkeby.livepeer.org/api/events",
+		},
+		"mainnet": {
+			ethUrl:        "wss://mainnet.infura.io/ws",
+			ethController: "0xf96d54e490317c557a967abfa5d6e33006be69b3",
+			monHost:       "http://metrics-mainnet.livepeer.org/api/events",
+		},
+	}
+
+	// temporary until we fully separate transcoder and orch nodes
```
j0sh: Each mode can already be run standalone so I don't think this is needed

cmd/livepeer/livepeer.go
```diff
@@ -130,47 +153,30 @@ func main() {
...
-		}
+		*network = "offchain"
+		glog.Infof("***Livepeer is in off-chain mode***")
 	}
 
-	if *orchAndTrans {
-		transcoder = true
-		*orchestrator = true
-	}
+	*datadir = *datadir + "/" + *network
```
j0sh:
```
import "path/filepath"
if *datadir == "" {
  *datadir = filepath.Join(usr.HomeDir, ".lpData", *network)
}
```

cmd/livepeer/livepeer.go
j0sh: Can we just make `offchain` the default in `network := flag.String("network", "", "Network to connect to")` ?

That being said, I think we can live without the "feature" of giving another name to an offchain network. The only thing that changes is the default datadir, which the user can still set to whatever.

cmd/livepeer/livepeer.go
```diff
@@ -130,47 +153,30 @@ func main() {
 		}
 	}
 
-	if *rinkeby {
-		if !*offchain && !transcoder {
-			if *ethUrl == "" {
-				*ethUrl = "wss://rinkeby.infura.io/ws"
+	/// Setting config options based on specified network
+	if *network != "" {
```
j0sh: Don't think this `if` is necessary since configOptions[*network] will just return empty if the network is empty

cmd/livepeer/livepeer.go
```diff
@@ -130,47 +153,30 @@ func main() {
...
+		if netw, ok := configOptions[*network]; ok {
+			if *orchestrator {
```
j0sh: We'd want to also set these values for broadcasters. Seems OK to just set the values regardless of the node type.

cmd/livepeer/livepeer.go
```diff
@@ -60,67 +58,92 @@ func main() {
...
+		ethUrl        string
+		ethController string
+		monHost       string
+	}
 
+	configOptions := map[string]*NetworkConfig{
+		"rinkeby": {
+			ethUrl:        "wss://rinkeby.infura.io/ws",
+			ethController: "0x37dc71366ec655093b9930bc816e16e6b587f968",
+			monHost:       "http://metrics-rinkeby.livepeer.org/api/events",
```
j0sh: Presented like this, the naming feels a bit inconsistent in having `ethUrl` and `monHost` . How about `monUrl` ?

cmd/livepeer/livepeer.go
```diff
@@ -254,14 +260,14 @@ func main() {
 
 		//Get the Eth client connection information
 		gethUrl := ""
-		if *ethIpcPath != "" {
+		if *ethUrl != "" {
 			//Connect to specified IPC file
-			gethUrl = *ethIpcPath
+			gethUrl = *ethUrl
 		} else if *ethUrl != "" {
```
j0sh: This part isn't needed anymore

cmd/livepeer/livepeer.go
```diff
@@ -254,14 +260,14 @@ func main() {
 
 		//Get the Eth client connection information
 		gethUrl := ""
-		if *ethIpcPath != "" {
+		if *ethUrl != "" {
 			//Connect to specified IPC file
```
j0sh: "Connect to the specified Ethereum node"

cmd/livepeer/livepeer.go
```diff
@@ -254,14 +260,14 @@ func main() {
...
-			gethUrl = *ethIpcPath
+			gethUrl = *ethUrl
 		} else if *ethUrl != "" {
 			//Connect to specified websocket
 			gethUrl = *ethUrl
 		} else {
-			glog.Errorf("Need to specify ethIpcPath or ethWsUrl")
+			glog.Errorf("Need to specify ethUrl or ethWsUrl")
```
j0sh: `glog.Error("Need to specify ethUrl")`

j0sh (changes requested):

cmd/livepeer/livepeer.go
angyangie: True! We can indeed live without that feature. I'll make the default `*network = "offchain"`

cmd/livepeer/livepeer.go
angyangie: True. I placed it there to mirror previous functionality with `* orchAndTrans` - when transcoder was set to true, `* orchAndTrans` also became true. My concern was keeping `devtool` script working. But, to that end, I'll just add an `-orchestrator` flag to devtool script and achieve the same result.

cmd/livepeer/livepeer.go
angyangie: That's right - now that we've set a default network of "offchain", no need to do this check.

cmd/livepeer/livepeer.go
angyangie: Sounds good. In the original file, we first checked for `if !*offchain && !transcoder {`. Do we need transcoders to also have these values? I could edit that line to read: `if *orchestrator || *broadcaster {`, and add a broadcaster flag, which I intended to do today?

cmd/livepeer/livepeer.go
angyangie: Ah yes, I had that change locally. Was going to push today!

cmd/livepeer/livepeer.go
j0sh: Transcoders don't really need to specify a `-network` flag. If they do, it seems harmless, so I wouldn't special case this to exclude transcoders in favor of keeping the flow straight-through.

j0sh:
Pushed some changes.

* Node type (-broadcaster / -transcoder / -orchestrator) is now **required** and will fail with a message if not specified. Earlier versions of this PR did not set a type if none was omitted. Whether we should have a default (broadcaster?) is certainly up for discussion. 356f87a83b9e1334ac7d9d7f7cd4d8ae4b05da36

* Improved handling of required O/T flags. 9436176d8187f7e599a2f3aa81d0f6f470177dea

* Look at the $HOME environment variable prior to asking the OS for the user's homedir. This is better practice anyway, and facilitates testing. Since it's technically non-portable, we fall back to the old method if $HOME is not set. 6a7807615973f9621e8dbda31ad63bbe568b6255

* Some other code consolidation around B's flag handling 453741dfae1ce90264f4f28ebc27199a725fd5fb

* Fix B's result sig checking in offchain mode. Separate bug but including it here anyway. 21bcb43aba89e4a5a9fb01fdfdbb63f0c134e0e1

* Added a test script that checks our handling of CLI flags. 4a167d3f9770d628aff65c86cbcf15b72c60665f

Once this is merged, we need to make the corresponding changes to devenv and the test harness.

j0sh:
Updated the README and rebased onto master. 94b30b595be8ada3f0e83a2d4f1863ba0b542070

Since there are some behavioral changes (2530428a312244b8d56ecb8fea65f71a22a0a129 , f34a80d23e6adf84208f152048c5d7358ac901a9) it would be nice to get some eyes on this before merging. See the [earlier comment](https://github.com/livepeer/go-livepeer/pull/738#issuecomment-466238486)  for details.

cc @angyangie @darkdarkdragon @yondonfu @ya7ya @iameli @ericxtang

cmd/livepeer/livepeer.go
```diff
@@ -254,14 +254,11 @@ func main() {
 
 		//Get the Eth client connection information
 		gethUrl := ""
-		if *ethIpcPath != "" {
-			//Connect to specified IPC file
-			gethUrl = *ethIpcPath
-		} else if *ethUrl != "" {
-			//Connect to specified websocket
+		if *ethUrl != "" {
```
yondonfu: Could we simplify this if/else into:

```
if *ethUrl == "" {
    glog.Error("Need to specify ethUrl")
    return
}
```

since `*ethUrl` would either be equal to the user specified value or one of the network defaults at this point? We could also remove the extra `gethUrl` variable.

cmd/devtool/devtool.go
```diff
@@ -331,9 +334,9 @@ func remoteConsole(destAccountAddr string) error {
 	}
 	defer console.Stop(false)
 
-	if !controllerAddrOverride {
+	if !ethControllerOverride {
 		// f9a6cf519167d81bc5cb3d26c60c0c9a19704aa908c148e82a861b570f4cd2d7 - SetContractInfo event
-		getControllerAddressScript := `
+		getethControllerScript := `
```
yondonfu: `getethControllerScript` -> `getEthControllerScript`? Same comment applies for the other appearances of this variable in this file.

cmd/livepeer/livepeer.go
```diff
@@ -397,21 +410,10 @@ func main() {
...
-		if n.OrchestratorPool == nil {
-			glog.Errorf("No orchestrator specified; transcoding will not happen")
-		}
-	}
-
 	//Create Livepeer Node
 
 	// Set up logging
-	if !*noIPFSLogFiles {
+	if !*logIPFS {
```
yondonfu: Perhaps it would be clearer to default `logIPFS` to true and check for `if *logIPFS {}` here?

test_args.sh
```diff
@@ -0,0 +1,93 @@
...
+[ -d "$CUSTOM_DATADIR" ]
+[ ! -d  "$CUSTOM_DATADIR"/offchain ] # sanity check that network isn't included
+kill $pid
+
+# check custom datadir with a network
+run_lp -broadcaster -datadir "$CUSTOM_DATADIR" -network rinkeby $ETH_ARGS
+[ ! -d  "$CUSTOM_DATADIR"/rinkeby ] # sanity check that network isn't included
+kill $pid
+
+# check invalid / missing service address.
```
yondonfu: More specifically, we'd want to test for a mismatch between the specified service address (either declared on-chain or user provided) and the inferred service address right?

yondonfu (changes requested):

test_args.sh
j0sh: It's meant to ensure an unset serviceURI makes the availability check fail, since the inferred address will (most likely!) be non-public. Updated the comment to reflect that.

One day we should spend time getting this type of test to work reliably. Then there are various items we should test related to this specific item:
* Missing `serviceAddr` (the current test, but commented out)
* Malformed `serviceAddr` (added this test)
* Well-formed `serviceAddr` pointing somewhere else
* On-chain but empty `serviceURI`
* On-chain but malformed `serviceURI`
* On-chain, well-formed `serviceURI` pointing somewhere else

cmd/livepeer/livepeer.go
j0sh: Ah this actually should have been `if *logIPFS {`. Fixed.

Should we set the default to `true` even with IPFS being unused?

cmd/livepeer/livepeer.go
j0sh: Fixed

cmd/devtool/devtool.go
j0sh: Fixed

cmd/livepeer/livepeer.go
yondonfu: Hm right - I think it makes sense to default to `false` right now since IPFS is unused, but if/when we start using IFPS again then it might make sense to default it to `true` to avoid IPFS log messages adding noise to the Livepeer node log output.

yondonfu (approved):
LGTM

j0sh (approved):

cmd/livepeer/livepeer.go
```diff
@@ -229,10 +224,13 @@ func main() {
 
 	if n.NodeType == core.TranscoderNode {
 		glog.Info("***Livepeer is in transcoder mode ***")
+		if n.OrchSecret == "" {
```
angyangie, line 229: Good thinking here

test_args.sh
```diff
@@ -0,0 +1,97 @@
+#!/usr/bin/env bash
```
angyangie, line 1: Nice addition (whole file)!

PR #740, opened 2019-02-08
Improve docker build time

darkdarkdragon:
Improves re-build time for Debian based docker image.

Changes to Dockerfiles.debian

manually

**Checklist:**

- [] Node runs in OSX and devenv


darkdarkdragon:
@j0sh please check if this will improve your re-build time - for time it makes re-builds 3 times faster.

.deps
```diff
@@ -0,0 +1,586 @@
+errors
```
j0sh: How was this file generated? When would it have to be updated? Maybe add a comment somewhere explaining that.

Is it Docker specific? If so, perhaps the file name could be adjusted to reflect that it's for the benefit of Docker, since otherwise it could appear as if it's part of some official golang dependency management solution.

Dockerfile.debian
```diff
@@ -30,7 +30,33 @@ RUN go get -u google.golang.org/grpc
...
 
-COPY . .
+COPY vendor vendor
+COPY .deps .deps
+
+# this will pre-build and cache files in /vendor dir
+RUN go list -v -export $(cat .deps)
+
+COPY ipfs ipfs
+RUN go list -v -export github.com/livepeer/go-livepeer/ipfs
```
j0sh, line 49: Just curious why these dirs `ipfs`, `monitor` and `pm` have a `go list` here when the rest aren't. Perhaps add a comment explaining why?

j0sh (changes requested):

Dockerfile.debian
darkdarkdragon: added some comments

.deps
darkdarkdragon: Added comment inside docker file:
```
# .dockerbuild.deps contains list of packages used by go-client
# needed to build them before building our code so it will be cached
# by docker and reused on next builds.
# it was generated manually by running `go build -v` and capturing output
# should be updated only if we remove some dependency from vendor dir
```
It was best way I was able to find to make docker to cache out build. I can't just precompile `vendor` dir because it contains many things that just do not compile.

j0sh: > I can't just precompile `vendor` dir because it contains many things that just do not compile.

Should we remove those? The vendor dir really needs some cleanup anyway. Preferably we'd get rid of the whole thing with golang modules...

j0sh (approved):
:ship:

.deps
darkdarkdragon: The way to precompile whole dir is like this ``
But this leads to errors like these:
```
# github.com/livepeer/go-livepeer/vendor/github.com/ethereum/go-ethereum/crypto/bn256
vendor/github.com/ethereum/go-ethereum/crypto/bn256/bn256_fast.go:26:6: G1 redeclared in this block
	previous declaration at vendor/github.com/ethereum/go-ethereum/crypto/bn256/bn256.go:30:6
```
Probably it's because it's tries to compile whole package, and I don't know how to get over it without manually removing files one-by-one from inside package - and this is obviously not good idea.

PR #741, opened 2019-02-12
set livepeer binary version using `git describe` command

### LINKED ISSUE #574: Better Versioning
Release versioning should have git commit hashes included.

Proposal: `v0.3.1-cd4fe0d`

darkdarkdragon:
Set version show by Livepeer binary to `git describe` output. This includes git's commit hash, that should help with troubleshooting.

Added `livepeer` target to makefile which uses `-ldflags` to set content of `core.LivepeerVersion` to output of `git describe` command.

manually

Fixes #574 

**Checklist:**


j0sh:
We can still use tags as part of the release process but IMO it would also be nice to have the version number somewhere visible, rather than sorta hidden in git.

This human-visible version number could also be used to generate the git tag, eg checking whether the version has been changed since the last release with a `make release` target.

darkdarkdragon:
I wanted to have single source for out version information. Plus I want to introduce as little extra steps as possible. And anyway, to find current version I was needed to trace it through source files, don't see much problem if that search will be one step longer.

darkdarkdragon:
Added explanatory comment around `LivepeerVersion` constant in code

j0sh:
The issue is it's non-obvious where the version number lives or how it's generated. Git tags are useful as a supplement to snapshot releases, but they aren't a replacement for version numbers by themselves.

There can still be a single source of truth for the version information: in the file that we define the version number. The git tag etc can be generated from that.

darkdarkdragon:
@j0sh 
> The issue is it's non-obvious where the version number lives or how it's generated.

It's not obvious in any case. When you first time opens source code for any project, you need to trace through source code to find out how this project generates version number. It doesn't matter where it will be - you still need to make an effort to find it. And when you're working with project everyday - then you'll just now where it is, so it would be seems obvious.
Keeping tag in separate file and generating git tags from that adds extra step, which needs to be maintained and can potentially be source of problems.

j0sh:
@darkdarkdragon Git tags are not a substitute for version numbers. There are several problems with the approach of using git tags to substitute for version numbers:

* Not searchable (eg, grep).
* The git tree is not always available along with the source code (eg, `git archive` or github source releases). This makes it impossible to determine approximately what version a given copy of the source corresponds to.
* Makes the release manager's job harder. They have to:
  * Find the previous version number
  * Ensure the new version is tagged correctly in git. Do they append a `v` ?
* Git tags are actually not a "single source" of truth; there can be many tags. Did the release manager intend to create a new release, or fix an earlier build? There's no way to "compile and test" without making relatively permanent changes to the git tree. (Yes, tags can be fixed locally but it's finicky, and again, not obvious because they are rather 'invisible' most times.) With the version encoded separately, this is a complete non-issue: there is only one canonical place where the version is defined, and tags can be created and checked accordingly based on that version, after the release build has been verified.

darkdarkdragon:
@j0sh 
> * There's no way to "compile and test" without making relatively permanent changes to the git tree.

What do you mean by that?

OK, having `VERSION` file in the root of repo with version inside sounds good?

j0sh:
> > * There's no way to "compile and test" without making relatively permanent changes to the git tree.

> What do you mean by that?

A tag has to first be created in git *before* a build is made. If there's something wrong with the build, you have to go in, remove the tag, etc.

> OK, having `VERSION` file in the root of repo with version inside sounds good?

:+1:

darkdarkdragon:
@j0sh I've made changes

j0sh (approved):
LGTM, thanks! 🚢

PR #752, opened 2019-02-26
Webhook URL

### LINKED ISSUE #703: Webhooks: RTMP Approval
In the context of demoing Livepeer to potential customers, we want to have a simple stream authentication mechanism.

We should be able to configure B nodes with a webhook URL, which they will call for every incoming stream with that stream's user-defined ManifestID, and approve or deny the processing of that stream based on the webhook response.

- Add a CLI flag, such as `webhookURL` of type `string`
- If `webhookURL` is set, B should **block on the webhook response** before approving/denying the incoming stream
- The request to `webhookURL` should include the `ManifestID` value
- The webhook should respond with HTTP 200 if the stream is approved, and HTTP 403 if the stream is denied

We discussed a few alternatives:
- REST API
- Forcing `ManifestID` values to include a pre-approved prefix, which would in effect act like an API key (being random, long, secret, and unique per customer)

We prefer the webhook approach because it adds less specific business considerations (e.g. prefixes), and requires less ops overhead from the node operator (e.g. with REST they need a TLS cert setup).

**Additional context**
None at the moment.

darkdarkdragon:
Webhook URL option - allows authenticate incoming streams through call to webhook

Adds `webhookUrl` option to Livepeer node - this URL get called upon receiving RTMP connection and manifest id being passed to it. If this request returns non 200 response, then RTMP connection being denied.

Fixes #703

**Checklist:**

### DIFF
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -106,6 +106,9 @@ func main() {
 	gsBucket := flag.String("gsbucket", "", "Google storage bucket")
 	gsKey := flag.String("gskey", "", "Google Storage private key file name (in json format)")
 
+	// API
+	authWebhookURL := flag.String("authWebhookUrl", "", "RTMP authentication webhook URL")
+
 	flag.Parse()
 	vFlag.Value.Set(*verbosity)
 
@@ -383,7 +386,10 @@ func main() {
 			// Not a fatal error; may continue operating in segment-only mode
 			glog.Error("No orchestrator specified; transcoding will not happen")
 		}
-
+		var err error
+		if server.AuthWebhookURL, err = getAuthWebhookURL(*authWebhookURL); err != nil {
+			glog.Fatal("Error setting auth webhook URL ", err)
+		}
 	} else if n.NodeType == core.OrchestratorNode {
 		suri, err := getServiceURI(n, *serviceAddr)
 		if err != nil {
@@ -497,6 +503,21 @@ func main() {
 		time.Sleep(time.Millisecond * 500) //Give time for other processes to shut down completely
 		return
 	}
+}
+
+func getAuthWebhookURL(u string) (string, error) {
+	if u == "" {
+		return "", nil
+	}
+	p, err := url.ParseRequestURI(u)
+	if err != nil {
+		return "", err
+	}
+	if p.Scheme != "http" && p.Scheme != "https" {
+		return "", errors.New("Webhook URL should be HTTP or HTTP")
+	}
+	glog.Infof("Using webhook url %s", u)
+	return u, nil
 }
 
 // ServiceURI checking steps:
--- a/server/mediaserver.go
+++ b/server/mediaserver.go
@@ -6,8 +6,10 @@ package server
 import (
 	"context"
 	"encoding/hex"
+	"encoding/json"
 	"errors"
 	"fmt"
+	"io/ioutil"
 	"math/big"
 	"math/rand"
 	"net/http"
@@ -53,6 +55,8 @@ const BroadcastRetry = 15 * time.Second
 var BroadcastPrice = big.NewInt(1)
 var BroadcastJobVideoProfiles = []ffmpeg.VideoProfile{ffmpeg.P240p30fps4x3, ffmpeg.P360p30fps16x9}
 
+var AuthWebhookURL string
+
 type rtmpConnection struct {
 	mid     core.ManifestID
 	nonce   uint64
@@ -83,6 +87,10 @@ type LivepeerServer struct {
 	lastHLSStreamID core.StreamID
 	lastManifestID  core.ManifestID
 	connectionLock  *sync.RWMutex
+}
+
+type authWebhookResponse struct {
+	ManifestID string `json:"manifestID"`
 }
 
 func NewLivepeerServer(rtmpAddr string, httpAddr string, lpNode *core.LivepeerNode) *LivepeerServer {
@@ -146,33 +154,71 @@ func (s *LivepeerServer) StartMediaServer(ctx context.Context, transcodingOption
 //RTMP Publish Handlers
 func createRTMPStreamIDHandler(s *LivepeerServer) func(url *url.URL) (strmID string) {
 	return func(url *url.URL) (strmID string) {
-		//Create a ManifestID
-		//If manifestID is passed in, use that one
+		//Check webhook for ManifestID
+		//If ManifestID is returned from webhook, use it
+		//Else check URL for ManifestID
+		//If ManifestID is passed in URL, use that one
 		//Else create one
-		mid := parseManifestID(url.Query().Get("manifestID"))
+		var mid core.ManifestID
+		var err error
+		if mid, err = authenticateStream(url.String()); err != nil {
+			glog.Error("Authentication denied for ", err)
+			return ""
+		}
+
+		if mid == "" {
+			mid = parseManifestID(url.Query().Get("manifestID"))
+		}
 		if mid == "" {
 			mid = core.RandomManifestID()
 		}
 
 		// Ensure there's no concurrent StreamID with the same name
 		s.connectionLock.RLock()
+		defer s.connectionLock.RUnlock()
 		if core.MaxSessions > 0 && len(s.rtmpConnections) >= core.MaxSessions {
 			glog.Error("Too many connections")
-			s.connectionLock.RUnlock()
 			return ""
 		}
 		if _, exists := s.rtmpConnections[mid]; exists {
 			glog.Error("Manifest already exists ", mid)
-			s.connectionLock.RUnlock()
 			return ""
 		}
-		s.connectionLock.RUnlock()
 
 		// Generate RTMP part of StreamID
 		key := hex.EncodeToString(core.RandomIdGenerator(StreamKeyBytes))
 		return core.MakeStreamIDFromString(string(mid), key).String()
 	}
+}
 
+func authenticateStream(url string) (core.ManifestID, error) {
+	if AuthWebhookURL == "" {
+		return "", nil
+	}
+	payload := fmt.Sprintf(`{"url":"%s"}"`, url)
+	body := strings.NewReader(payload)
+
+	resp, err := http.Post(AuthWebhookURL, "application/json", body)
+	if err != nil {
+		return "", err
+	}
+	rbody, err := ioutil.ReadAll(resp.Body)
+	resp.Body.Close()
+	if resp.StatusCode != 200 {
+		return "", errors.New(resp.Status)
+	}
+	if len(rbody) == 0 {
+		return "", nil
+	}
+	var authResp authWebhookResponse
+	err = json.Unmarshal(rbody, &authResp)
+	if err != nil {
+		return "", err
+	}
+	if authResp.ManifestID == "" {
+		return "", errors.New("Empty manifest id not allowed")
+	}
+	return core.ManifestID(authResp.ManifestID), nil
 }
 
 func rtmpManifestID(rtmpStrm stream.RTMPVideoStream) core.ManifestID {
--- a/server/mediaserver_test.go
+++ b/server/mediaserver_test.go
@@ -8,6 +8,8 @@ import (
 	"fmt"
 	"math/big"
 	"math/rand"
+	"net/http"
+	"net/http/httptest"
 	"net/url"
 	"sync"
 	"testing"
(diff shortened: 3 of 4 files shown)


darkdarkdragon:
@j0sh If this looks ok, then I will (at least try) add tests for it.

darkdarkdragon:
added tests

cmd/livepeer/livepeer.go
```diff
@@ -102,6 +102,7 @@ func main() {
 	faceValue := flag.Float64("faceValue", 0, "The faceValue to expect in PM tickets, denominated in ETH (e.g. 0.3)")
 	winProb := flag.Float64("winProb", 0, "The win probability to expect in PM tickets, as a percent float between 0 and 100 (e.g. 5.3)")
 	maxSessions := flag.Int("maxSessions", 10, "Orchestrator only. Maximum number of concurrent transcoding sessions")
+	webhookURL := flag.String("webhookUrl", "", "Authentication webhook URL")
```
j0sh: Maybe the description could be "RTMP authentication webhook URL". Should we also call it something like `authUrl` so we don't have to rename things if/when we add more hooks later on?

cmd/livepeer/livepeer.go
```diff
@@ -379,6 +380,19 @@ func main() {
 		// TODO provide an option to disable this?
 		*rtmpAddr = defaultAddr(*rtmpAddr, "127.0.0.1", RtmpPort)
 		*httpAddr = defaultAddr(*httpAddr, "127.0.0.1", RpcPort)
+		if *webhookURL != "" {
```
j0sh: Perhaps factor this out into a `getWebhookURL` helper? Eg, `server.WebhookURL = getWebhookURL(*webhookURL)`

server/mediaserver.go
```diff
@@ -162,9 +164,15 @@ func createRTMPStreamIDHandler(s *LivepeerServer) func(url *url.URL) (strmID str
...
 		s.connectionLock.RLock()
-		defer s.connectionLock.RUnlock()
 		if _, exists := s.rtmpConnections[mid]; exists {
 			glog.Error("Manifest already exists ", mid)
+			s.connectionLock.RUnlock()
+			return ""
+		}
+		s.connectionLock.RUnlock()
+
+		if !authenticateStream(string(mid)) {
```
j0sh: Let's authenticate first thing, before acquiring the lock.

I'm thinking we should pass in the full URL rather than the ManifestID. That way the auth endpoint can inspect the entire host + path, and enforce their own rules about path construction.

The thought was to have the webhook URL return the manifest ID. We should also structure the response to leave room for further customization, eg transcoding options for the stream.

```
GET https://host/webhook/endpoint
    Request Body: { "url" : <url> }
    Response Body: { "manifestID" : <manifestID> }
```
We could also do a POST here since a GET with a body is a bit unusual.

````golang
func createRTMPStreamIDHandler(s *LivepeerServer) func(url *url.URL) (strmID string) {     
     return func(url *url.URL) (strmID string) {     
        // Create a ManifestID
        // If an auth webhook is defined, get the mid from the webhook
        // Else if manifestID is passed in, use that one
        // Else create one
        var mid string 
        if WebhookURL != "" {
          mid, err = authenticateStream(url)
          if err != nil {
            return ""
          }
        } else {
          mid = parseManifestID(...)
        }
```

darkdarkdragon: > Let's authenticate first thing, before acquiring the lock.

Previous check will allow us to quit early in case of same Manifest ids, so why not?

> The thought was to have the webhook URL return the manifest ID

I really don't remember discussing such thing. Can you point me to such discussion?

>We should also structure the response to leave room for further customization,

This is starting to be something other from what we're discussed.

j0sh: > Previous check will allow us to quit early in case of same Manifest ids

If the manifest ID is defined by the webhook then the webhook needs to come before.  Also avoids acquiring the lock which also helps.

> I really don't remember discussing such thing. Can you point me to such discussion?

Yeah, we had a lot of discussions around this and some details may have been lost. This was mentioned here https://hackmd.io/n0Lq1l6_SJCe0VnVHf4G-A#Webhook-Flow which was outlined as part of the discussion here : 
https://docs.google.com/a/livepeer.org/document/d/1ThTyFeTjTfnWWj-HeNCMnkglD7Vf-9YvA71fB9dKVQs/edit?disco=AAAACet6oNM

> This is starting to be something other from what we're discussed.

How so? The end result is the same, we aren't changing the scope of the work here. But tweaking the webhook format makes the construct much more versatile.

Once we merge this, it effectively becomes part of the public API, and it's harder to make breaking changes (even internally), so it helps a lot to leave a little room for expansion.

darkdarkdragon: > https://hackmd.io/n0Lq1l6_SJCe0VnVHf4G-A#Webhook-Flow

missed this one

> was outlined as part of the discussion here :
https://docs.google.com/a/livepeer.org/document/d/1ThTyFeTjTfnWWj-HeNCMnkglD7Vf-9YvA71fB9dKVQs/edit?disco=AAAACet6oNM

Here is no mention of webhook defining manifest ids. 

> How so? The end result is the same, we aren't changing the scope of the work here.

Webhook's server should be more complicated now, plus our logic gets more complicated - what to do if webhook returned manifest id that we already have? Or if it is invalid manifest id?

j0sh: > Here is no mention of webhook defining manifest ids.

There seems to have been quite a bit lost "in translation" when migrating from the [old doc](https://docs.google.com/document/d/16Z0xvIYe8FNcKJ76HCVZ8JJ8xfHytr9te5GRNrE3XhU/edit) to the [new one](https://docs.google.com/document/d/1ThTyFeTjTfnWWj-HeNCMnkglD7Vf-9YvA71fB9dKVQs/edit#).

> Alternative Implementation - Webhook:
> Start up node with a `-streamAuthWebhook http://host/streamauth`
> RTMP stream comes in
> Node calls the streamAuthWebhook with the RTMP URL, and  other parameters?
> Node receives manifest ID in response (disconnects incoming rtmp stream if webhook  response indicates an invalid stream)
> Stream is available for playback via the manifest ID

The behavior spec in the old doc also details how it would be desirable to have the input URL to be decoupled from the output URL to prevent input hijacking or guessing playback URLs given an input. Having the webhook return the manifest ID enables that.

Do you think that is not useful?

> Webhook's server should be more complicated now

Reading a JSON request and replying with JSON is standard, as far as APIs go.

> plus our logic gets more complicated

Our logic stays the same. The only thing that changes is the webhook gets called first to set the manifest id, if necessary.

A webhook is a large feature as far as API surface area goes, and we want to extract as much utility from it as possible. Right now its utility is limited.

cmd/livepeer/livepeer.go
darkdarkdragon: Renamed

cmd/livepeer/livepeer.go
darkdarkdragon: done

server/mediaserver.go
darkdarkdragon: I've added POST API, with only one addition - if webhook returns 200 with empty body - it is enough to pass authentication. This will simplify creation of simple webhooks.

cmd/livepeer/livepeer.go
```diff
@@ -499,6 +505,23 @@ func main() {
...
+	if err != nil {
+		glog.Errorf("Error in webhook URL: %v", err)
+		return "", true
+	}
+	if p.Scheme != "http" && p.Scheme != "https" {
+		glog.Error("Webhook URL should be HTTP or HTTP")
+		return "", true
+	}
+	glog.Infof("Using webhook url %s", u)
+	return "", false
```
j0sh: `return u` ?

cmd/livepeer/livepeer.go
```diff
@@ -499,6 +505,23 @@ func main() {
 	}
 }
 
+func getAuthWebhookURL(u string) (string, bool) {
```
j0sh: Perhaps returning the error here rather than a bool would be more informative

cmd/livepeer/livepeer.go
```diff
@@ -383,7 +386,10 @@ func main() {
 			// Not a fatal error; may continue operating in segment-only mode
 			glog.Error("No orchestrator specified; transcoding will not happen")
 		}
-
+		var err bool
+		if server.AuthWebhookURL, err = getAuthWebhookURL(*authWebhookURL); err {
+			return
```
j0sh: The logging + exit can be done once here by returning an error:

```
if server.AuthWebhookURL, err = getAuthWebookURL(*authWebhookURL); err != nil {
  glog.Fatal("Error setting auth webhook URL ", err)
}
```

I've been trying to use `glog.Fatal` around `livepeer.go` to log and exit.

Then maybe some test cases can be added in the new script checking cli args - https://github.com/livepeer/go-livepeer/blob/master/test_args.sh

server/mediaserver.go
```diff
@@ -154,19 +162,59 @@ func createRTMPStreamIDHandler(s *LivepeerServer) func(url *url.URL) (strmID str
...
 		key := hex.EncodeToString(core.RandomIdGenerator(StreamKeyBytes))
 		return core.MakeStreamIDFromString(string(mid), key).String()
 	}
+}
 
+func authenticateStream(url, mid string) (core.ManifestID, bool) {
+	if AuthWebhookURL == "" {
+		return core.ManifestID(mid), true
+	}
+	payload := "{\"manifestID\":\"" + mid + "\",\"url\":\"" + url + "\"}"
```
j0sh: I think you can use backticks to avoid escaping, like `` `{"url":"`+url.String()+`"}` ``. Using `fmt.Sprintf` might also help readability around concatenation.

Don't think passing in the manifestID is necessary since any query strings can be extracted from the URL itself.

server/mediaserver.go
```diff
@@ -154,19 +162,59 @@ func createRTMPStreamIDHandler(s *LivepeerServer) func(url *url.URL) (strmID str
...
 		}
+		s.connectionLock.RUnlock()
 
 		// Generate RTMP part of StreamID
 		key := hex.EncodeToString(core.RandomIdGenerator(StreamKeyBytes))
 		return core.MakeStreamIDFromString(string(mid), key).String()
 	}
+}
 
+func authenticateStream(url, mid string) (core.ManifestID, bool) {
```
j0sh: Returning an error rather than a bool would help upstream logging and avoid the need for logs within this function

server/mediaserver.go
```diff
@@ -154,19 +162,59 @@ func createRTMPStreamIDHandler(s *LivepeerServer) func(url *url.URL) (strmID str
 			mid = core.RandomManifestID()
 		}
 
+		var auth bool
+		if mid, auth = authenticateStream(url.String(), string(mid)); !auth {
```
j0sh: I think the flow clearer if we called `authenticateStream` first, before the manifestID checks. This also avoids having to pass in the manifest ID itself, which makes `authenticateStream` easier to understand (single responsibility). Also, returning an error rather than bool is more helpful.

```golang
mid, err := authenticateStream(url)
if err != nil {
    glog.Error("Could not authenticate stream: ", err)
    return ""
}
if mid == "" {
    mid = parseManifestID(url.Query().Get("manifestID"))
}
if mid == "" {
  mid = core.RandomManifestID()
}
```

server/mediaserver.go
```diff
@@ -154,19 +162,59 @@ func createRTMPStreamIDHandler(s *LivepeerServer) func(url *url.URL) (strmID str
...
+			glog.Error("Authentication denied for ", url)
+			return ""
+		}
+
 		// Ensure there's no concurrent StreamID with the same name
 		s.connectionLock.RLock()
-		defer s.connectionLock.RUnlock()
```
j0sh: We can probably leave the `defer` here since subsequent operations are not expensive. Less error prone

server/mediaserver_test.go
```diff
@@ -42,6 +43,29 @@ func setupServer() *LivepeerServer {
...
+	mux.HandleFunc("/auth/emptymanifest", func(w http.ResponseWriter, r *http.Request) {
+		w.Write([]byte("{\"manifestID\":\"\"}"))
+	})
+	mux.HandleFunc("/auth/xy", func(w http.ResponseWriter, r *http.Request) {
+		w.Write([]byte("{\"manifestID\":\"xy\"}"))
+	})
+	mux.HandleFunc("/auth/invalidjson", func(w http.ResponseWriter, r *http.Request) {
+		w.Write([]byte("{manifestID\":}"))
+	})
+	go srv.ListenAndServe()
```
j0sh: Spinning up a live HTTP server for testing can be unpredictable / problematic so it's best to avoid it if possible. It's not always possible, but here I think we can avoid it here we follow the format that's used in https://github.com/livepeer/go-livepeer/blob/90c10193860999497cef5f1031fec4890bb9f19e/server/handlers_test.go

cmd/livepeer/livepeer.go
```diff
@@ -106,6 +106,9 @@ func main() {
 	gsBucket := flag.String("gsbucket", "", "Google storage bucket")
 	gsKey := flag.String("gskey", "", "Google Storage private key file name (in json format)")
 
+	// API
+	authWebhookURL := flag.String("authWebhookUrl", "", "Authentication webhook URL")
```
j0sh: Can the message mention that it's a "RTMP authentication webhook URL" ? That makes it clearer what it's used for.

cmd/livepeer/livepeer.go
```diff
@@ -499,6 +505,23 @@ func main() {
...
+		return "", false
+	}
+	p, err := url.ParseRequestURI(u)
+	if err != nil {
+		glog.Errorf("Error in webhook URL: %v", err)
+		return "", true
+	}
+	if p.Scheme != "http" && p.Scheme != "https" {
+		glog.Error("Webhook URL should be HTTP or HTTP")
+		return "", true
```
j0sh: `return "", fmt.Errorf("Webhook URL should be HTTP or HTTPS")` then we can avoid the logging.

j0sh (changes requested):

cmd/livepeer/livepeer.go

cmd/livepeer/livepeer.go
darkdarkdragon: done

server/mediaserver.go
darkdarkdragon: Originally it was done to pass randomly generated manifest id, but you're right, there is no point in that. Fixed.

server/mediaserver.go
darkdarkdragon: done

server/mediaserver.go
darkdarkdragon: done

server/mediaserver.go
darkdarkdragon: done

server/mediaserver_test.go
darkdarkdragon: done

cmd/livepeer/livepeer.go
darkdarkdragon: done

cmd/livepeer/livepeer.go
darkdarkdragon: done

darkdarkdragon:
@j0sh Looks like I've addressed all the comments

server/mediaserver.go
```diff
@@ -146,33 +154,71 @@ func (s *LivepeerServer) StartMediaServer(ctx context.Context, transcodingOption
...
-		//If manifestID is passed in, use that one
+		//Check webhook for ManifestID
+		//If ManifestID is returned from webhook, use it
+		//Else check URL for ManifestID
+		//If ManifestID is passed in URL, use that one
 		//Else create one
-		mid := parseManifestID(url.Query().Get("manifestID"))
+		var mid core.ManifestID
+		var err error
+		if mid, err = authenticateStream(url.String()); err != nil {
```
j0sh, line 164: Just a tiny style nit, but we don't need to predeclare the vars here (and would save one line overall):

```golang
mid, err := authenticateStream(url.String())
if err != nil {
```

j0sh (approved):
Tested with a local auth server and works nicely!! Just one style nit, otherwise LGTM :ship:

PR #753, opened 2019-02-26
Capacity Cap for B

### LINKED ISSUE #706: Capacity Cap for B
To improve customer experience, B should respond with a clear error message whenever it's at its max capacity of concurrent streams and a new RTMP stream comes in.

The user/customer should receive clear feedback from B that it is now at max capacity and therefore cannot handle new streams.

We probably want to add yet another CLI param, such as `maxConcurrentStreams` of type `int`. B should reject any incoming stream once its concurrent stream count reaches `maxConcurrentStreams`.

Not sure yet how to best achieve this goal?
- Does RTMP support this kind of response?
- Should we rely on webhooks for this kind of response?
- Any other approaches?

Discussed above.

**Additional context**
None.

darkdarkdragon:
Add capacity cap for Broadcaster

Reuse `-maxSessions` option for Broadcaster, on new RTMP connection check if number of current connections (`rtmpConnections`) is less than `-maxSessions`. By default it is not specified and not checked.

Added unit tests

Fixes #706 

**Checklist:**

### DIFF
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -101,7 +101,7 @@ func main() {
 	verbosity := flag.String("v", "", "Log verbosity.  {4|5|6}")
 	faceValue := flag.Float64("faceValue", 0, "The faceValue to expect in PM tickets, denominated in ETH (e.g. 0.3)")
 	winProb := flag.Float64("winProb", 0, "The win probability to expect in PM tickets, as a percent float between 0 and 100 (e.g. 5.3)")
-	maxSessions := flag.Int("maxSessions", 10, "Orchestrator only. Maximum number of concurrent transcoding sessions")
+	maxSessions := flag.Int("maxSessions", 10, "Maximum number of concurrent transcoding sessions for Orchestrator or maximum number or RTMP streams for Broadcaster")
 	webhookURL := flag.String("webhookUrl", "", "Authentication webhook URL")
 
 	flag.Parse()
@@ -373,7 +373,7 @@ func main() {
 		}
 	}
 
-	core.MaxTranscodeSessions = *maxSessions
+	core.MaxSessions = *maxSessions
 
 	if n.NodeType == core.BroadcasterNode {
 		// default lpms listener for broadcaster; same as default rpc port
--- a/core/orchestrator.go
+++ b/core/orchestrator.go
@@ -27,8 +27,6 @@ import (
 	"github.com/livepeer/lpms/stream"
 )
 
-var MaxTranscodeSessions = 10
-
 var transcodeLoopTimeout = 1 * time.Minute
 
 // Transcoder / orchestrator RPC interface implementation
@@ -77,7 +75,7 @@ func (orch *orchestrator) CheckCapacity(mid ManifestID) error {
 	if _, ok := orch.node.SegmentChans[mid]; ok {
 		return nil
 	}
-	if len(orch.node.SegmentChans) >= MaxTranscodeSessions {
+	if len(orch.node.SegmentChans) >= MaxSessions {
 		return ErrOrchCap
 	}
 	return nil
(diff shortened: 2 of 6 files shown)


cmd/livepeer/livepeer.go
```diff
@@ -101,7 +101,7 @@ func main() {
 	verbosity := flag.String("v", "", "Log verbosity.  {4|5|6}")
 	faceValue := flag.Float64("faceValue", 0, "The faceValue to expect in PM tickets, denominated in ETH (e.g. 0.3)")
 	winProb := flag.Float64("winProb", 0, "The win probability to expect in PM tickets, as a percent float between 0 and 100 (e.g. 5.3)")
-	maxSessions := flag.Int("maxSessions", 10, "Orchestrator only. Maximum number of concurrent transcoding sessions")
+	maxSessions := flag.Int("maxSessions", 0, "Maximum number of concurrent transcoding sessions for Orchestrator (default is 10) or maximum number or RTMP streams for Broadcaster (not set by default)")
```
j0sh: Does it seem reasonable to have a default limit for B? We could keep the cap for 10 Bs, which would simplify a number of the other default cases here. If a B wants unlimited sessions it can just set `-maxSessions 0`.

server/mediaserver_test.go
```diff
@@ -177,6 +177,27 @@ func TestSelectOrchestrator(t *testing.T) {
 	assert.Equal(expSessionID, sess.PMSessionID)
 }
 
+func TestCreateRTMPStreamHandlerCap(t *testing.T) {
+	s := &LivepeerServer{
+		connectionLock:  &sync.RWMutex{},
+		rtmpConnections: make(map[core.ManifestID]*rtmpConnection),
+	}
+	createSid := createRTMPStreamIDHandler(s)
```
j0sh, line 185: Maybe add some comments indicating the cases that are being tested, similar to what we have for the orchestrator?

core/orchestrator.go
```diff
@@ -27,7 +27,7 @@ import (
 	"github.com/livepeer/lpms/stream"
 )
 
-var MaxTranscodeSessions = 10
+var MaxSessions = 0
```
j0sh: Should this be in `livepeernode.go` since it's shared between B/O now?

j0sh (changes requested):

cmd/livepeer/livepeer.go
darkdarkdragon: Good that you've pointed that out - right now we have different behavior for O and B - zero for B means unlimited, but for O it really means zero - this is not very intuitive.
I think we should make it same - say, default is 10 for both, but zero is unlimited for O too.
wdyt?

j0sh: Unlimited O seems fraught. I'm OK with the minor inconsistency, or perhaps we should just not allow unlimited sessions for B?

darkdarkdragon: Set default 10 to B and zero as unlimited for B

server/mediaserver_test.go
darkdarkdragon: added some comments

core/orchestrator.go
darkdarkdragon: moved

cmd/livepeer/livepeer.go
```diff
@@ -394,6 +394,9 @@ func main() {
 			server.WebhookURL = *webhookURL
 		}
 	} else if n.NodeType == core.OrchestratorNode {
+		if core.MaxSessions == 0 {
+			core.MaxSessions = 10
```
j0sh: Not needed anymore

server/mediaserver_test.go
```diff
@@ -358,6 +382,7 @@ func TestGotRTMPStreamHandler(t *testing.T) {
 }
 
 func TestMultiStream(t *testing.T) {
+	core.MaxSessions = 0
```
j0sh, line 367: Is this still needed?

j0sh: NVM, I see now. Perhaps add a comment saying it's needed to set unlimited sessions.

darkdarkdragon: yes, this test creates 500 streams, so it fails if capped to 10

cmd/livepeer/livepeer.go
darkdarkdragon: removed

server/mediaserver_test.go
darkdarkdragon: done

j0sh (approved):
🚢

PR #754, opened 2019-02-26
Add unit tests for ServeSegment and SubmitSegment

### LINKED ISSUE #675: Add unit tests for ServeSegment and SubmitSegment

The [ServeSegment](https://github.com/livepeer/go-livepeer/blob/master/server/rpc.go#L309) and [SubmitSegment](https://github.com/livepeer/go-livepeer/blob/master/server/rpc.go#L489) have limited test coverage. We should add unit tests for these functions so that we can achieve greater test coverage in our unit tests starting from submission/receipt of a RPC message through the processing of a RPC message (which we already have unit tests for via lower level components).

Add unit tests for both these functions.

The [httptest](https://golang.org/pkg/net/http/httptest/) package from the Go stdlib might be useful. See the test structure of some of the latest HTTP handler code [here](https://github.com/livepeer/go-livepeer/blob/master/server/handlers_test.go#L611).

We could choose not to add unit tests for these functions. However, adding these unit tests would increase our confidence about the correctness of the full RPC message passing workflow starting from submission, to receipt through processing.

yondonfu:

Add unit tests for the ServeSegment and SubmitSegment.

- Refactored ServeSegment to return http.Handler to make it easier to test
- Added unit tests for ServeSegment and SubmitSegment - used the [httptest](https://golang.org/pkg/net/http/httptest/) package

I ran the additional unit tests introduced by this PR.

Fixes #675 

**Checklist:**

### DIFF
--- a/server/segment_rpc.go
+++ b/server/segment_rpc.go
@@ -27,6 +27,7 @@ import (
 )
 
 const PaymentHeader = "Livepeer-Payment"
+const SegmentHeader = "Livepeer-Segment"
 
 var ErrSegEncoding = errors.New("ErrorSegEncoding")
 var ErrSegSig = errors.New("ErrSegSig")
@@ -48,7 +49,7 @@ func (h *lphttp) ServeSegment(w http.ResponseWriter, r *http.Request) {
 	}
 
 	// check the segment sig from the broadcaster
-	seg := r.Header.Get("Livepeer-Segment")
+	seg := r.Header.Get(SegmentHeader)
 
 	segData, err := verifySegCreds(orch, seg, getPaymentSender(payment))
 	if err != nil {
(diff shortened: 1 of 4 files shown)


server/segment_rpc.go
```diff
@@ -37,120 +38,120 @@ var httpClient = &http.Client{
...
-	orch := h.orchestrator
-
-	payment, err := getPayment(r.Header.Get(PaymentHeader))
-	if err != nil {
-		glog.Error("Could not parse payment")
-		http.Error(w, err.Error(), http.StatusPaymentRequired)
-		return
-	}
+func ServeSegment(orch Orchestrator) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
j0sh: Is there a way we can avoid the extra level of indentation here? It's partially stylistic, but it's a little easier to follow the context and deeper indentation is typically one sign of code smell. For reviewability here, the diff is also a bit hard to follow but I know that's harder to control (but avoiding the indentation would make the diff much cleaner overall).

Although we use this pattern of handler-within-a-closure elsewhere (eg, mediaserver.go) it's still something I'd prefer not to promulgate if we can help it.

yondonfu: @j0sh Updated so `ServeSegment` no longer returns `http.Handler` and instead added a test helper that wraps `ServeSegment` in 0b1fda7761235f6ba7227afb8082220953f93eee. WDYT?

j0sh (approved):
LGTM -- when merging, can you squash to the commits to exclude the intermediate changes to the diffs such as the large re-indents of ServeSegment that was undone?

PR #797, opened 2019-03-20
Send `NoOrchestrators` event

darkdarkdragon:
Send `NoOrchestrators` event in case source segment appeared
but there is not orchestrators selected

- 
- 
- 

**Checklist:**

### DIFF
--- a/monitor/events.go
+++ b/monitor/events.go
@@ -129,7 +129,7 @@ func LogSegmentUploadFailed(nonce, seqNo uint64, reason string) {
 }
 
 func LogTranscodedSegmentAppeared(nonce, seqNo uint64, profile string) {
-	glog.Infof("Logging LogTranscodedSegmentAppeared...")
+	glog.Infof("Logging LogTranscodedSegmentAppeared... ", nonce, seqNo, profile)
 	props := map[string]interface{}{
 		"seqNo":   seqNo,
 		"profile": profile,
@@ -218,7 +218,7 @@ func LogSegmentTranscoded(nonce, seqNo uint64, transcodeDur, totalDur time.Durat
 }
 
 func LogSegmentTranscodeFailed(subType string, nonce, seqNo uint64, err error) {
-	glog.Info("Logging LogSegmentTranscodeFailed", subType)
+	glog.Info("Logging LogSegmentTranscodeFailed ", subType, nonce, seqNo)
 
 	if metrics.lastSegmentNonce == nonce {
 		metrics.segmentsInFlight--
--- a/server/broadcast.go
+++ b/server/broadcast.go
@@ -1,6 +1,7 @@
 package server
 
 import (
+	"errors"
 	"fmt"
 	"math/big"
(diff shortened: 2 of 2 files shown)


monitor/events.go
```diff
@@ -234,6 +235,16 @@ func LogSegmentTranscodeFailed(subType string, nonce, seqNo uint64, err error) {
...
 	sendPost("SegmentTranscodeFailed", nonce, props)
+
+	if subType == "NoOrchestrators" {
+		// this will allow us to find out at what stages of stream "NoOrchestrators" errors occur
+		props2 := map[string]interface{}{
+			"subType": fmt.Sprintf("NoOrchestrators-seq-%04d", seqNo),
+			"reason":  err.Error(),
+			"seqNo":   seqNo,
+		}
+		sendPost("SegmentTranscodeFailed", nonce, props2)
```
j0sh: It seems like we'd be sending SegmentTranscodeFailed twice? Once with `props` and again with `props2` ?

darkdarkdragon: Yes, but they're get sent with different `subType` - second one differentiate events by SeqNo: `NoOrchestrators-seq-%04d`. They will be shown in metrics separately, so we will be able to see when in the stream these errors took place.

j0sh: Don't we also send the sequence number in the `seqNo` field along with the subtype?

Is this temporary until Prometheus? Will we have a better story for drilling down into this type of event as opposed to double-counting errors?

Can you confirm this is the source of the discrepancy in upload behavior by running the test harness with this patch?

darkdarkdragon: > Don't we also send the sequence number in the seqNo field along with the subtype?

Yes, we do. But sending it two times allow me to show separate entry containing `seqNo` in metrics UI without changing metrics server code.

image



> Is this temporary until Prometheus? 

All code related to own metrics should be removed eventually, when we will be satisfied with Prometheus based metrics. 

> Will we have a better story for drilling down into this type of event as opposed to double-counting errors?

I'm working on that.

> Can you confirm this is the source of the discrepancy in upload behavior by running the test harness with this patch?

Wasn't able to reproduced, but I haven't tried too hard.

j0sh: > Wasn't able to reproduced, but I haven't tried too hard.

My guess is that it'd be easier to track down this specific behavior by:

* Drilling down on errors for a specific stream (maybe we can avoid this if load is light)
* Viewing seqNo for all errors, in addition to NoOrchestrators.
* Seeing if we're losing the O in relation to some other error (seqNo should closely follow between errors).

The approach here seems a little ad-hoc, especially if it hasn't already helped to give a sense of *why* the errors are happening. If it's not worth modifying the metrics server code then maybe we should focus on getting this type of visibility through Prometheus, especially if all this is going to be removed soon anyway.

darkdarkdragon:
@j0sh I've removed error duplication.

j0sh (approved):
🚢

Issue #708, opened 2019-01-24
Document CLI Flags


j0sh:
**Is your feature request related to a problem? Please describe.**
The command line flags are under-documented.

**Describe the solution you'd like**
Document the flags more thoroughly, including with examples. Ideally we would keep the documentation "close to the code", eg somewhere in `docs` where they can be easily referenced offline and updated as the code changes.

**Describe alternatives you've considered**
We could place the docs somewhere in readthedocs. However, a link from readthedocs to the in-node docs might suffice.

yondonfu:
@j0sh Do you think it would be sufficient to just update the descriptions that are passed in as flag parameters in [livepeer.go](https://github.com/livepeer/go-livepeer/blob/master/cmd/livepeer/livepeer.go#L90)? Then, the documentation could just live with the flag code itself and we could just dump the output of `livepeer -help` into a file in `docs` if needed.

j0sh:
@yondonfu The inline CLI documentation is typically terse. It would be nice to have expanded descriptions for most options - context around the option itself, the B/O/T modes they apply to, any interactions with other options (some are mutually exclusive), examples of usage, etc etc.

PR #798, opened 2019-03-22
Structured Ingest URLs

### LINKED ISSUE #764: Custom Stream Keys via Auth Webhook
Playback of RTMP ingest streams is protected by a stream key, but there is no programmatic way to discover this key for playback from other elements on the network, or to control the generation of the key.

Currently, the RTMP auth webhook has a limitation that any backslashes from the returned manifestID will be incorrectly parsed during playback as being part of the rendition, based on the playback path convention of `manifestID/rendition/seqNo`.

Looking at the first backslash in the manifestID returned from the webhook would be a useful way to incorporate the RTMP stream key, which is conditionally autogenerated. So, if the manifestID contains a slash, we consider it a manifestid+streamkey (eg, fully formed RTMP StreamID), and this can be used for RTMP playback from other elements of the broadcaster's infrastructure.

Note that subsequent slashes after the first can be considered part of the stream key, without any issue.

An alternative is to set this as separate fields in the response, eg: `{"manifestID": mid, "streamKey": key}`. However, we would likely still have to scan the manifestID and terminate the stream if any invalid characters (currently just a backslash) are found.

We could also implement an API for explicit RTMP stream aliases, but that seems unwieldy compared to the webhooks approach.

### LINKED ISSUE #791: Structured Ingest URL Paths
Currently, our ingest URLs are relatively unstructured. The manifestID is passed in as an optional query string parameter, with the path ignored:

```
rtmp://ingest-host/path/is/ignored?manifestID=mid
```

There is some concern about the use of "special characters" to form the query string. Additionally, it would be nice to make the RTMP URL more consistent with the HLS playback URL, which is of the form:

```
http://ingest-host/stream/<manifestID>.m3u8
http://ingest-host/stream/<manifestID>/<rendition>.m3u8
http://ingest-host/stream/<manifestID>/<rendition>/<num>.ts
```

We can use the first part of the RTMP path as the manifestID and the rest as the stream key. For example:

```
# manifest ID = "streamName"
# stream key = "secretStreamKey"
rtmp://host/streamName/secretStreamKey

# manifest ID = "streamName"
# stream key = "secret/stream/key"
rtmp://host/streamName/secret/stream/key

# manifest ID = "streamName"
# Stream key = randomized
rtmp://host/streamName

# manifest ID = randomized
# Stream key = randomized
rtmp://host
```

**Additional context**

Because of how we parse URLs for HLS playback [1], the manifestID cannot have slashes in it. This is the reason the streamKey could end up being `secret/stream/key` in the URL `rtmp://host/streamName/secret/stream/key` .

[1] http://host/stream/<manifestID>/<rendition>/<num>.ts

j0sh:
Uses RTMP URL by default to construct manifest ID and stream key.

- https://github.com/livepeer/go-livepeer/commit/bcd4202eb34912f4527ad66ccf444cad600b64f9 Extract ManifestID from URL path. Note that this follows the scheme outlined in https://github.com/livepeer/go-livepeer/issues/791#issuecomment-472572559 where the user can incorporate an optional `/stream/` prefix into the path.

- https://github.com/livepeer/go-livepeer/commit/4bd3167d2f1d11a1ff34dd794540c907ee6d58e9 Extract stream key from URL path. Also return stream keys from the auth webhook. If no key is provided, then a random key is generated (following the old behavior).

- Removed the support for looking up the manifestID via the query parameter. This is technically a breaking change; the test harness and any client scripts that use the query parameter will have to change. cc @ya7ya @darkdarkdragon This does *not* affect webhooks, which are still free to use any URL structure they please, including query parameters, as shown in our [simple auth server](https://github.com/livepeer/go-livepeer/blob/master/cmd/simple_auth_server/simple_auth_server.go) example.

Added unit testing, performed manual testing.

Fixes https://github.com/livepeer/go-livepeer/issues/791
Fixes https://github.com/livepeer/go-livepeer/issues/764

**Checklist:**

### DIFF
--- a/doc/ingest.md
+++ b/doc/ingest.md
@@ -1,0 +1,86 @@
+# RTMP Ingest
+
+Livepeer starts a RTMP server on the default port of 1935 as the ingest point
+into the Livepeer network. Upon ingest, the segmenter pulls the RTMP stream
+prior to transcoding.
+
+### Ingest Configuration
+
+By default, the RTMP server listens to localhost on the default RTMP port of 1935.
+
+This can be set at node start-up time with the `-rtmpAddr` flag, which takes an
+`interface:port` pair, such as `-rtmpAddr 0.0.0.0:1936` which would make the
+Livepeer node listen to all interfaces on port 1936.
+
+The node has a default maximum of 10 concurrent RTMP sessions. To change this, run the node with the `-maxSessions` flag indicating the limit, for example `-maxSessions 100` to raise the limit to 100 concurrent sessions.
+
+### Stream Naming and Addressing
+
+The stream name is taken to be the first part of the RTMP URL path. The stream name may optionally be prefixed with `/stream/` to match the HLS output address.
+
+```
+# Canonical Form
+rtmp://localhost/movie1
+rtmp://localhost/movie2
+
+# Alternate Form
+rtmp://localhost/stream/movie1
+rtmp://localhost/stream/movie2
+
+# Output URL
+http://localhost:8935/stream/movie1.m3u8
+http://localhost:8935/stream/movie2.m3u8
+```
+
+If no name is provided, then a stream name is randomly generated. For example:
+
+```
+# Ingest URL
+rtmp://localhost
+rtmp://localhost/stream # alternate form; /stream gets stripped
+
+# RTMP Playback URL
+rtmp://localhost/<randomStreamName>/<randomStreamKey>
+
+# HLS Playback URL
+http://localhost:8935/stream/<randomStreamName>.m3u8
+```
+
+There are two options for using randomly generated stream names.
+
+If the node is started with the `-currentManifest` flag, then the latest stream can be
+accessed via the HLS `current.m3u8` endpoint, regardless of its name, eg
+
+`curl http://localhost:8935/stream/current.m3u8`
+
+Alternatively, a list of active streams can be found by querying the CLI API:
+
+`curl http://localhost:7935/status`
+
+
+### Stream Authentication
+
+Streams can be authenticated through a webhook. See the documentation on the
+[RTMP Authentication Webhook](rtmpauthwebhook.md) for more details.
+
+### RTMP Playback Protection
+
+The RTMP stream can be played back, or pulled from Livepeer by another part of
+the ingest infrastructure. To prevent unauthorized RTMP playback of streams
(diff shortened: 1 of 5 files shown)


doc/rtmpwebhookauth.md
```diff
@@ -22,11 +22,15 @@ Webhook can respond with zero body - in that case `ManifestID` for the stream wi
 
 ```json
 {
-    "manifestID": "StrigThatIsIDofManifest"
+    "manifestID": "ManifestIDString"
```
darkdarkdragon: should be comma in the end

darkdarkdragon (changes requested):
I think we should at least state in the docs that feeding RTMP stream int
`rtmp://host/stream` will not yield stream to be available at
`http://host/stream/stream.m3u8` - this can be confusing.

doc/rtmpwebhookauth.md
j0sh: Fixed

j0sh:
Expanded the documentation with a new page on ingest handling since explaining the `/stream/` handling seemed like too much detail for the README: https://github.com/livepeer/go-livepeer/pull/798/commits/19b045de0844e3cc758a9cb7962f215227662757

The doc also expands on the concept of protecting RTMP playback via stream keys, and documents a few other CLI flags that modify ingest behavior. (Although we really should have a single page that comprehensively documents all our CLI flags: https://github.com/livepeer/go-livepeer/issues/708 )

darkdarkdragon:
LGTM 🚢

doc/ingest.md
```diff
@@ -0,0 +1,86 @@
...
+Livepeer node listen to all interfaces on port 1936.
+
+The node has a default maximum of 10 concurrent RTMP sessions. To change this, run the node with the `-maxSessions` flag indicating the limit, for example `-maxSessions 100` to raise the limit to 100 concurrent sessions.
+
+### Stream Naming and Addressing
+
+The stream name is taken to be the first part of the RTMP URL path. The stream name may optionally be prefixed with `/stream/` to match the HLS output address.
+
+```
+# Canonical Form
```
angyangie, line 22: Really useful examples here!!

angyangie (approved):
LGTM!

j0sh:
@darkdarkdragon Github won't let me merge until you approve the changes, can that be done?

darkdarkdragon (approved):

PR #799, opened 2019-03-25
BroadcastSessionManager struct + methods

angyangie:

This PR builds the `BroadcastSessionManager` struct and associated methods. The struct manages the list of BroadcastSessions, and handles BroadcastSession selection/additions/list refreshes. 

For background, we've chosen the round-robin strategy as the most optimal method at this point in time, and get us to 99%+ reliability. A B will “take turns” assigning segments to be transcoded to different Os. eg. send seg1 to O1, seg2 to O2, seg3 to O3, seg4 to O1, seg5 to O2, etc. (in the case of three O's). Alternating sending segments to a selection of Os gives individual Os more time to process segments - the more O’s involved, the more time for segment processing. Os will not have more than one segment per stream in flight, which mitigates back logging.

- BroadcastSessionManager {}
- addToList()
- selectFromList()
- checkStatusOfList()

For more info, see here: https://docs.google.com/document/d/1KfEWm0z3svw10TrFH9hiMImkXocl3ba_S-8vmX94HNc/edit#

**Checklist:**

### DIFF
--- a/server/rpc.go
+++ b/server/rpc.go
@@ -56,50 +56,42 @@ type BroadcastSessionsManager struct {
 	sessLock          *sync.Mutex
 }
 
-// func (bsm *BroadcastSessionsManager) selectFromList(node *core.LivepeerNode, pl core.PlaylistManager) *BroadcastSession {
 func (bsm *BroadcastSessionsManager) selectFromList() *BroadcastSession {
-	// bsm.checkStatusOfList(numOrchs, node, pl)
 	bsm.checkStatusOfList()
 
 	bsm.sessLock.Lock()
 	defer bsm.sessLock.Unlock()
 
-	sess, sessions := bsm.broadcastSessions[0], bsm.broadcastSessions[1:]
-	bsm.broadcastSessions = sessions
-
-	return sess
-}
-
-func (bsm *BroadcastSessionsManager) removeFromList(session *BroadcastSession) {
-	bsm.sessLock.Lock()
-	defer bsm.sessLock.Unlock()
-	var sessions []*BroadcastSession
-	for i, s := range bsm.broadcastSessions {
-		if s == session {
-			sessions = append(bsm.broadcastSessions[:i], bsm.broadcastSessions[i+1:]...)
-			break
-		}
+	if len(bsm.broadcastSessions) > 0 {
+		last := len(bsm.broadcastSessions) - 1
+		sess, sessions := bsm.broadcastSessions[last], bsm.broadcastSessions[:last]
+		bsm.broadcastSessions = sessions
+		return sess
 	}
-	bsm.broadcastSessions = sessions
+	return nil
 }
 
 func (bsm *BroadcastSessionsManager) addToList(sess *BroadcastSession) {
 	bsm.sessLock.Lock()
+	defer bsm.sessLock.Unlock()
+
+	for _, currentSess := range bsm.broadcastSessions {
+		if currentSess == sess {
+			return
+		}
+	}
 	bsm.broadcastSessions = append(bsm.broadcastSessions, sess)
-	bsm.sessLock.Unlock()
 }
 
-// func (bsm *BroadcastSessionsManager) checkStatusOfList(numOrchs int, node *core.LivepeerNode, pl core.PlaylistManager) {
 func (bsm *BroadcastSessionsManager) checkStatusOfList() {
 	var newBroadcastSessions []*BroadcastSession
-	defaultNumOrchs := int(HTTPTimeout / SegLen)
-	reqLengthOfList := defaultNumOrchs * 3
+	// minOrchs is maximum number of inflight sessions
+	minOrchs := int(HTTPTimeout / SegLen)
+	reqLengthOfList := minOrchs * 2
 
-	bsm.sessLock.Lock()
-	if len(bsm.broadcastSessions) < defaultNumOrchs {
-		bsm.sessLock.Unlock()
-		// newBroadcastSessions, err := selectOrchestrator(node, pl)
+	if len(bsm.broadcastSessions) < minOrchs {
 		newBroadcastSessions = []*BroadcastSession{}
+		// newBroadcastSessions, err := selectOrchestrator(node, pl)
 		// if err != nil {
 		// 	return
 		// }
@@ -110,19 +102,31 @@ func (bsm *BroadcastSessionsManager) checkStatusOfList() {
 
 	bsm.sessLock.Lock()
 	defer bsm.sessLock.Unlock()
+
 	// include all O's being used successfully in refreshed list
-	for _, currentSess := range bsm.broadcastSessions {
+	var newSessionsMap map[*BroadcastSession]bool
+	if len(newBroadcastSessions) > 0 {
+
 		for _, newSess := range newBroadcastSessions {
-			if len(bsm.broadcastSessions) < reqLengthOfList {
-				if newSess != currentSess {
-					bsm.broadcastSessions = append(bsm.broadcastSessions, newSess)
+			newSessionsMap[newSess] = true
+		}
+
+		if len(bsm.broadcastSessions) > 0 {
+			for _, currentSess := range bsm.broadcastSessions {
+				if _, ok := newSessionsMap[currentSess]; ok {
+					delete(newSessionsMap, currentSess)
 				}
-			} else {
-				return
+			}
+		}
+
+		if len(newSessionsMap) > 0 {
+			for sess, _ := range newSessionsMap {
+				if len(bsm.broadcastSessions) < reqLengthOfList {
+					bsm.broadcastSessions = append(bsm.broadcastSessions, sess)
+				}
 			}
 		}
 	}
-	return
 }
 
 // Session-specific state for broadcasters
--- a/server/rpc.go
+++ b/server/rpc.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 	"net/url"
 	"strings"
+	"sync"
 	"time"
 
 	"google.golang.org/grpc"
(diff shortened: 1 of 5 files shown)


server/rpc.go
```diff
@@ -50,6 +51,81 @@ type Broadcaster interface {
...
 }
 
+type BroadcastSessionsManager struct {
+	broadcastSessions []*BroadcastSession
+	sessLock          *sync.Mutex
+}
+
+// func (bsm *BroadcastSessionsManager) selectFromList(node *core.LivepeerNode, pl core.PlaylistManager) *BroadcastSession {
+func (bsm *BroadcastSessionsManager) selectFromList() *BroadcastSession {
+	numOrchs := 3
```
angyangie: This hardcode will be removed once we determine how to generate this number for B's, keeping in mind available capacity on the network (which will be important especially in our current state, when the network is still growing).

server/rpc.go
```diff
@@ -50,6 +51,81 @@ type Broadcaster interface {
...
+// func (bsm *BroadcastSessionsManager) checkStatusOfList(numOrchs int, node *core.LivepeerNode, pl core.PlaylistManager) {
+func (bsm *BroadcastSessionsManager) checkStatusOfList(numOrchs int) {
+	reqLengthOfList := numOrchs * 3
+
+	var newBroadcastSessions []*BroadcastSession
+
+	bsm.sessLock.Lock()
+	if len(bsm.broadcastSessions) < numOrchs {
+		bsm.sessLock.Unlock()
+		// newBroadcastSessions, err := selectOrchestrator(node, pl)
```
angyangie: Once we integrate with selectOrchestrator, this function will become more relevant and will be properly tested.

server/rpc.go
j0sh: Technically `numOrchs` can be determined as a function of the segment length (2 secs now) and the request timeout (8 seconds now). This would give us at most 4 segments in-flight before earlier segments start to time out. Network capacity would be a concern if the number of available orchestrators is less than `numOrchs` as it would lead to list refreshes on every segment.

We can try to minimize this at some point with something like `numOrchs = min(numAvailableOrchs, defaultNumOrchs)`

angyangie: That sounds like a fine way to calculate a universal `numOrchs`, until we find a use case for allowing Orchs to set their own

server/rpc.go
```diff
@@ -50,6 +51,80 @@ type Broadcaster interface {
...
+		// 	return
+		// }
+		glog.Info(newBroadcastSessions)
+	} else {
+		return
+	}
+
+	bsm.sessLock.Lock()
+	defer bsm.sessLock.Unlock()
+	// include all O's being used successfully in refreshed list
```
j0sh, line 106: How do we prevent double-including Os here?

server/rpc.go
```diff
@@ -50,6 +51,80 @@ type Broadcaster interface {
...
+	if len(bsm.broadcastSessions) < defaultNumOrchs {
+		bsm.sessLock.Unlock()
+		// newBroadcastSessions, err := selectOrchestrator(node, pl)
+		newBroadcastSessions = []*BroadcastSession{}
+		// if err != nil {
+		// 	return
+		// }
+		glog.Info(newBroadcastSessions)
+	} else {
+		return
```
j0sh, line 100: the `bsm.sessLock.Lock()` would dangle here

server/rpc.go
```diff
@@ -50,6 +51,80 @@ type Broadcaster interface {
...
+	bsm.sessLock.Unlock()
+}
+
+// func (bsm *BroadcastSessionsManager) checkStatusOfList(numOrchs int, node *core.LivepeerNode, pl core.PlaylistManager) {
+func (bsm *BroadcastSessionsManager) checkStatusOfList() {
+	var newBroadcastSessions []*BroadcastSession
+	defaultNumOrchs := int(HTTPTimeout / SegLen)
+	reqLengthOfList := defaultNumOrchs * 3
+
+	bsm.sessLock.Lock()
```
j0sh: Unclear what this lock is doing

server/rpc.go
```diff
@@ -50,6 +51,80 @@ type Broadcaster interface {
...
+	bsm.broadcastSessions = append(bsm.broadcastSessions, sess)
+	bsm.sessLock.Unlock()
+}
+
+// func (bsm *BroadcastSessionsManager) checkStatusOfList(numOrchs int, node *core.LivepeerNode, pl core.PlaylistManager) {
+func (bsm *BroadcastSessionsManager) checkStatusOfList() {
+	var newBroadcastSessions []*BroadcastSession
+	defaultNumOrchs := int(HTTPTimeout / SegLen)
+	reqLengthOfList := defaultNumOrchs * 3
```
j0sh: Just curious how a multiplier of 3 was chosen here

server/rpc.go
```diff
@@ -50,6 +51,80 @@ type Broadcaster interface {
...
+func (bsm *BroadcastSessionsManager) addToList(sess *BroadcastSession) {
+	bsm.sessLock.Lock()
+	bsm.broadcastSessions = append(bsm.broadcastSessions, sess)
+	bsm.sessLock.Unlock()
+}
+
+// func (bsm *BroadcastSessionsManager) checkStatusOfList(numOrchs int, node *core.LivepeerNode, pl core.PlaylistManager) {
+func (bsm *BroadcastSessionsManager) checkStatusOfList() {
+	var newBroadcastSessions []*BroadcastSession
+	defaultNumOrchs := int(HTTPTimeout / SegLen)
```
j0sh: maybe a note this is the maximum number of inflight sessions.

Since `defaultNumOrchs`, is probably better described as "the minimum number of orchestrators we need before we'll try to refresh the orchestrator list", maybe call it `minOrchs` or something

server/rpc.go
```diff
@@ -50,6 +51,80 @@ type Broadcaster interface {
...
+
+// func (bsm *BroadcastSessionsManager) selectFromList(node *core.LivepeerNode, pl core.PlaylistManager) *BroadcastSession {
+func (bsm *BroadcastSessionsManager) selectFromList() *BroadcastSession {
+	// bsm.checkStatusOfList(numOrchs, node, pl)
+	bsm.checkStatusOfList()
+
+	bsm.sessLock.Lock()
+	defer bsm.sessLock.Unlock()
+
+	sess, sessions := bsm.broadcastSessions[0], bsm.broadcastSessions[1:]
```
j0sh: Needs length checks

server/rpc.go
```diff
@@ -50,6 +51,80 @@ type Broadcaster interface {
...
+	for i, s := range bsm.broadcastSessions {
+		if s == session {
+			sessions = append(bsm.broadcastSessions[:i], bsm.broadcastSessions[i+1:]...)
+			break
+		}
+	}
+	bsm.broadcastSessions = sessions
+}
+
+func (bsm *BroadcastSessionsManager) addToList(sess *BroadcastSession) {
```
j0sh: Here, we would be cycling continuously through the list like a queue, selecting from the head and pushing (appending) to the tail. This is probably OK if we want to give everyone a chance at processing a segment, but for now we probably want to give more weight to quick respondents.

To reward quick responders with more segments, we want stack-like behavior: select from the head, and push back to the head (prepend rather than append).

The asymptotics of prepend in golang don't appear great (involves copying a slice which is not ideal under a lock), but I wonder if treating the end of the list as the head to would help here. So:

```golang
selectFromList() {
  last = len(broadcastSessions) - 1
  // ...do bounds checking...
  // select from end of list
  sess, sessions = broadacastSessions[last], broadcastSessions[:last]
}
addToList(session) {
  // append to end of list
  sessions = append(broadcastSessions, session)
}
```

Slice indexing behavior might totally negate this, but it looks a little better than an explicit copy.

server/rpc.go
j0sh: And how do we accommodate Os that might have segments in flight and hence not reflected in the list?

server/rpc_test.go
```diff
@@ -532,3 +534,59 @@ func defaultTicket(t *testing.T) *net.Ticket {
...
+		ManifestID:  core.RandomManifestID(),
+	}
+	sess2 := &BroadcastSession{
+		Broadcaster: StubBroadcaster2(),
+		ManifestID:  core.RandomManifestID(),
+	}
+	return &BroadcastSessionsManager{broadcastSessions: []*BroadcastSession{sess1, sess2}, sessLock: &sync.Mutex{}}
+}
+
+func TestSelectFromList_Successt(t *testing.T) {
```
j0sh: The `_Success` seems redundant in the name of these functions.

server/rpc_test.go
```diff
@@ -532,3 +534,59 @@ func defaultTicket(t *testing.T) *net.Ticket {
...
+	bsm := StubBroadcastSessionsManager()
+	expectedSess := bsm.broadcastSessions[0]
+	sess := bsm.selectFromList()
```
j0sh: Also test the empty-list case, and check that the list lengths are as expected before-and-after the function call.

server/rpc_test.go
```diff
@@ -532,3 +534,59 @@ func defaultTicket(t *testing.T) *net.Ticket {
...
+}
+
+func TestAddToList_Success(t *testing.T) {
+	bsm := StubBroadcastSessionsManager()
+	sess := &BroadcastSession{
+		Broadcaster: StubBroadcaster2(),
+		ManifestID:  core.RandomManifestID(),
+	}
+
+	bsm.addToList(sess)
```
j0sh, line 575: Sanity check the "before" case as well, eg that the list length is as expected.

Also test concurrent executions  of `AddToList` and `SelectFromList` and run with the `-race` flag.

j0sh (changes requested):

server/rpc.go
angyangie: Yeaaaa I found that out the hard way when sketching out integrating all of this into `processSegment`. I'll add that!

server/rpc.go
angyangie: Good points! For your first one, how about this? 
```golang 
// include all O's being used successfully in refreshed list
var newSessionsMap map[*BroadcastSession]bool
if len(newBroadcastSessions) > 0 {

	for _, newSess := range newBroadcastSessions {
		newSessionsMap[newSess] = true
	}

	if len(bsm.broadcastSessions) > 0 {
		for _, currentSess := range bsm.broadcastSessions {
			if _, ok := newSessionsMap[currentSess]; ok {
				delete(newSessionsMap, currentSess)
			}
		}
	}

	if len(newSessionsMap) > 0 {
		for sess, _ := range newSessionsMap {
			if len(bsm.broadcastSessions) < reqLengthOfList {
				bsm.broadcastSessions = append(bsm.broadcastSessions, sess)
			}
		}
	}
}
```

As for the second, how about this? 
```golang
func (bsm *BroadcastSessionsManager) addToList(sess *BroadcastSession) {
	bsm.sessLock.Lock()
	defer bsm.sessLock.Unlock()

	for _, currentSess := range bsm.broadcastSessions {
		if currentSess == sess {
			return
		}
	}
	bsm.broadcastSessions = append(bsm.broadcastSessions, sess)
}
```

server/rpc.go
angyangie: I'm fine with doing it this way. Works as long as I add a `len(bsm.broadcastSessions) > 0` to `selectFromList`. Good to know on prepending copying a slice under the hood!

server/rpc.go
angyangie: My thinking was that using a multiple of 2 could lead to the following scenario: if `minOrchs` = 4 and `reqLengthOfList` = 8, and there are 4 segments in flight, `bsm.BroadcastSessionsManager` = 4. Say one of those Orchs fails and is removed from the list. We now we have 7 orchs. If there are 4 segments in flight, and 3 orchs left on the list, `bsm.BroadcastSessionsManager` will be refreshed. If this scenario happens often, the list will be refreshed very often. However, in this same scenario, theoretically, we'd be adding 5 orchs to the list with the listrefresh, since we'd be locking the list until it was once again at `len = reqLengthOfList`. So perhaps it wouldn't be a problem (a list longer than `reqLengthOfList` would be fine as well). What do you think?

server/rpc.go
angyangie: Ah yes, perhaps I should just add `bsm.sessLock.Lock() .... defer bsm.sessLock.Unlock()` at the top of the function

server/rpc.go
angyangie: I'll remove this and lock/defer unlock at the top of the function!

server/rpc_test.go
angyangie: Do you mean this? 
```golang
bsm := StubBroadcastSessionsManager()
expectedSess := bsm.broadcastSessions[1]

assert := assert.New(t)
assert.Len(bsm.broadcastSessions, 2)

sess := bsm.selectFromList()

assert.Equal(expectedSess, sess)
assert.Len(bsm.broadcastSessions, 1)
```

As for testing concurrent executions.... do you mean on my local environment?

PR #800, opened 2019-03-25
Export metrics using OpenCensus library

### LINKED ISSUE #760: Integrate OpenCensus client library
Make it available at `/metrics` endpoint on CLI port
Send to Prometheus following events (from Broadcaster) as simple counters:
- RTMP stream started
- RTMP stream ended
- Source segment appeared in manifest
- Segment uploaded for transcoding
- Transcoded segment appeared in manifest

Prometheus client library
https://github.com/prometheus/client_golang

darkdarkdragon:
Export internal metrics using OpenCensus library. 

Added OpenCensus library dependency. Export events that goes into old metrics server as metrics in Prometheus. 
Calc transcode time on transcode/orchestrator and export as metrics.
Calc time between source segment appears from segmenter and till transcoded segment put into resulting manifest and export this time as latency metric.

Manually.

Fixes #760

**Checklist:**

### DIFF
--- a/monitor/events.go
+++ b/monitor/events.go
@@ -10,6 +10,7 @@ import (
 	"github.com/golang/glog"
 )
 
+// Enabled true if metrics was enabled in command line
 var Enabled bool
 
 type metricsVars struct {
@@ -35,11 +36,12 @@ type event struct {
 
 var eventsURL string
 
-func Init(url string, nodeType string, nodeID string) {
+func Init(url, nodeType, nodeID, version string) {
 	eventsURL = url
 	metrics.nodeID = nodeID
 	metrics.nodeType = nodeType
 	go sendLoop(metrics.ch)
+	initCensus(nodeType, nodeID, version)
 }
 
 func sendLoop(inCh chan *event) {
@@ -64,7 +66,8 @@ func sendLoop(inCh chan *event) {
 }
 
 func LogSegmentTranscodeStarting(seqNo uint64, manifestID string) {
-	glog.Infof("Logging SegmentTranscodeStarting...")
+	glog.Infof("Logging SegmentTranscodeStarting... seqNo=%d manifestID=%s",
+		seqNo, manifestID)
 
 	props := map[string]interface{}{
 		"seqNo":      seqNo,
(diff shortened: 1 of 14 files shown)


server/mediaserver.go
```diff
@@ -72,6 +72,12 @@ type rtmpConnection struct {
 	// following fields should be protected by `lock`
 	sess *BroadcastSession
 	lock *sync.RWMutex
+
+	// Stuff for analytics
```
j0sh: Is there a way we can keep this self-contained within analytics? Eg, defining some struct within `monitor` and using the nonce to index into a map inside the `monitor` package. Then we can adjust counters as needed within specific events, eg `LogSegmentEmerged` and `LogSegmentTranscodeFailed`. I'd really like to avoid sprawling analytics code around various parts of the codebase if possible.

Also, I think most of these would need to be protected by a lock to make the race detector happy.

monitor/events.go
```diff
@@ -234,8 +260,19 @@ func LogSegmentTranscodeFailed(subType string, nonce, seqNo uint64, err error) {
...
+		props2 := map[string]interface{}{
+			"subType": fmt.Sprintf("NoOrchestrators-seq-%04d", seqNo),
+			"reason":  err.Error(),
+			"seqNo":   seqNo,
+		}
+		sendPost("SegmentTranscodeFailed", nonce, props2)
+	}
 }
 
+/*
```
j0sh: Do you see a path towards using this code? If not, should we remove it?

monitor/events.go
```diff
@@ -234,8 +260,19 @@ func LogSegmentTranscodeFailed(subType string, nonce, seqNo uint64, err error) {
 	detectSeqDif(props, nonce, seqNo)
 
 	sendPost("SegmentTranscodeFailed", nonce, props)
+
+	if subType == SegmentTranscodeErrorNoOrchestrators {
+		// this will allow us to find out at what stages of stream "NoOrchestrators" errors occur
+		props2 := map[string]interface{}{
+			"subType": fmt.Sprintf("NoOrchestrators-seq-%04d", seqNo),
```
j0sh: I'm still not totally sold on double-counting the errors in this way. Can you address the concerns that were raised here? https://github.com/livepeer/go-livepeer/pull/797#discussion_r267575117

j0sh (changes requested):
Looks good but would be nice to have the tracking be more self-contained within the `monitor` package; see notes around that.

Are we going to maintain the mongo metrics along with Prometheus? If we don't see the mongo stuff being useful in the long term, I'd like to also remove it to minimize further maintenance.

darkdarkdragon:
@j0sh I've moved all the implementation into own module and added some more metrics.

server/broadcast.go
```diff
@@ -220,6 +231,9 @@ func processSegment(cxn *rtmpConnection, seg *stream.HLSSegment) {
 			cond.Wait()
 		}
 		cond.L.Unlock()
+		if monitor.Enabled {
+			monitor.SegmentFullyTranscoded(nonce, seg.SeqNo, common.ProfilesNames(sess.Profiles), int(successes) == len(res.Segments))
```
j0sh: If the comparison is `len(segHashes) == len(res.Segments)` then that avoids the additional variable

j0sh (changes requested):
Looks very nice! Just a small request around how we check whether the segment was fully transcoded.

server/broadcast.go
darkdarkdragon: Fixed.

j0sh (approved):
LGTM! :ship:

PR #801, opened 2019-03-26
Ar/select orchestrator returns broadcast sessions

angyangie:

** **NOTE** : includes these changes: https://github.com/livepeer/go-livepeer/pull/799. Will rebase once #799 is merged. **

Edit `selectOrchestrator` to build and return []*BroadcastSessions, not just a BroadcastSession. This will also involve changing how a single BroadcastSession will be stored/created in startSession and in rtmpConnection struct (BroadcastSessionManager will be stored instead of a single BroadcastSession . 

Unit tests

See here for more info: https://docs.google.com/document/d/1KfEWm0z3svw10TrFH9hiMImkXocl3ba_S-8vmX94HNc/edit#

**Checklist:**

### DIFF
--- a/server/mediaserver.go
+++ b/server/mediaserver.go
@@ -59,11 +59,12 @@ var BroadcastJobVideoProfiles = []ffmpeg.VideoProfile{ffmpeg.P240p30fps4x3, ffmp
 var AuthWebhookURL string
 
 type rtmpConnection struct {
-	mid     core.ManifestID
-	nonce   uint64
-	stream  stream.RTMPVideoStream
-	pl      core.PlaylistManager
-	profile *ffmpeg.VideoProfile
+	mid         core.ManifestID
+	nonce       uint64
+	stream      stream.RTMPVideoStream
+	pl          core.PlaylistManager
+	profile     *ffmpeg.VideoProfile
+	sessManager *BroadcastSessionsManager
 
 	needOrch chan struct{}
 	eof      chan struct{}
@@ -365,15 +366,15 @@ func (s *LivepeerServer) registerConnection(rtmpStrm stream.RTMPVideoStream) (*r
 		return nil, ErrAlreadyExists
 	}
 	cxn := &rtmpConnection{
-		mid:     mid,
-		nonce:   nonce,
-		stream:  rtmpStrm,
-		pl:      core.NewBasicPlaylistManager(mid, storage),
-		profile: &vProfile,
-		lock:    &sync.RWMutex{},
-
-		needOrch: make(chan struct{}),
-		eof:      make(chan struct{}),
+		mid:         mid,
+		nonce:       nonce,
+		stream:      rtmpStrm,
+		pl:          core.NewBasicPlaylistManager(mid, storage),
+		profile:     &vProfile,
+		lock:        &sync.RWMutex{},
+		sessManager: &BroadcastSessionsManager{broadcastSessions: []*BroadcastSession{}, sessLock: &sync.Mutex{}},
+		needOrch:    make(chan struct{}),
+		eof:         make(chan struct{}),
 	}
 	s.rtmpConnections[mid] = cxn
 	s.lastManifestID = mid
--- a/server/mediaserver.go
+++ b/server/mediaserver.go
@@ -390,7 +390,7 @@ func (s *LivepeerServer) startSession(cxn *rtmpConnection) *BroadcastSession {
 
 	mid := cxn.mid
 	cpl := cxn.pl
-	var sess *BroadcastSession
+	var sess []*BroadcastSession
 
 	// Connect to the orchestrator. If it fails, retry for as long
 	// as the RTMP stream is alive
@@ -414,7 +414,10 @@ func (s *LivepeerServer) startSession(cxn *rtmpConnection) *BroadcastSession {
 	expb.MaxInterval = BroadcastRetry
 	expb.MaxElapsedTime = 0
 	backoff.Retry(broadcastFunc, expb)
-	return sess
+	if len(sess) > 0 {
+		return sess[0]
+	}
+	return nil
 }
 
 func (s *LivepeerServer) startSessionListener(cxn *rtmpConnection) {
--- a/server/mediaserver_test.go
+++ b/server/mediaserver_test.go
@@ -125,23 +125,29 @@ func TestSelectOrchestrator(t *testing.T) {
 		&net.OrchestratorInfo{},
 	}
 	sess, _ := selectOrchestrator(s.LivepeerNode, pl)
-	if sess == nil {
-		t.Error("Expected nil session")
+
+	if len(sess) != len(sd.infos) {
+		t.Error("Expected session length of 2")
 	}
+
+	if sess == nil {
+		t.Error("Expected non-nil session")
+	}
+
 	// Sanity check a few easy fields
-	if sess.ManifestID != mid {
+	if sess[0].ManifestID != mid {
 		t.Error("Expected manifest id")
 	}
-	if sess.BroadcasterOS != storage {
+	if sess[0].BroadcasterOS != storage {
 		t.Error("Unexpected broadcaster OS")
 	}
-	if sess.OrchestratorInfo != sd.infos[0] || sd.infos[0] == sd.infos[1] {
+	if sess[0].OrchestratorInfo != sd.infos[0] || sd.infos[0] == sd.infos[1] {
 		t.Error("Unexpected orchestrator info")
 	}
-	if sess.Sender != nil {
+	if sess[0].Sender != nil {
 		t.Error("Unexpected sender")
 	}
-	if sess.PMSessionID != "" {
+	if sess[0].PMSessionID != "" {
 		t.Error("Unexpected PM sessionID")
 	}
 
@@ -164,21 +170,46 @@ func TestSelectOrchestrator(t *testing.T) {
 		Seed:              params.Seed.Bytes(),
 	}
 
+	params2 := pm.TicketParams{
+		Recipient:         pm.RandAddress(),
+		FaceValue:         big.NewInt(1234),
+		WinProb:           big.NewInt(5678),
+		RecipientRandHash: pm.RandHash(),
+		Seed:              big.NewInt(7777),
+	}
+	protoParams2 := &net.TicketParams{
+		Recipient:         params2.Recipient.Bytes(),
+		FaceValue:         params2.FaceValue.Bytes(),
+		WinProb:           params2.WinProb.Bytes(),
+		RecipientRandHash: params2.RecipientRandHash.Bytes(),
+		Seed:              params2.Seed.Bytes(),
+	}
+
 	sd.infos = []*net.OrchestratorInfo{
 		&net.OrchestratorInfo{
 			TicketParams: protoParams,
+		},
+		&net.OrchestratorInfo{
+			TicketParams: protoParams2,
 		},
 	}
 
 	expSessionID := "foo"
 	sender.On("StartSession", params).Return(expSessionID)
 
+	expSessionID2 := "fool"
+	sender.On("StartSession", params2).Return(expSessionID2)
+
 	sess, err := selectOrchestrator(s.LivepeerNode, pl)
 	require.Nil(t, err)
 
 	assert := assert.New(t)
-	assert.Equal(sender, sess.Sender)
-	assert.Equal(expSessionID, sess.PMSessionID)
+	assert.Len(sess, 2)
+	assert.Equal(sender, sess[0].Sender)
+	assert.Equal(expSessionID, sess[0].PMSessionID)
+	assert.Equal(expSessionID2, sess[1].PMSessionID)
+	assert.Equal(sess[0].OrchestratorInfo, &net.OrchestratorInfo{TicketParams: protoParams})
+	assert.Equal(sess[1].OrchestratorInfo, &net.OrchestratorInfo{TicketParams: protoParams2})
 }
 
 func TestCreateRTMPStreamHandlerCap(t *testing.T) {
--- a/server/rpc.go
+++ b/server/rpc.go
@@ -53,80 +53,66 @@ type Broadcaster interface {
 
 type BroadcastSessionsManager struct {
 	broadcastSessions []*BroadcastSession
+	broadcastSessMap  map[string]*BroadcastSession
 	sessLock          *sync.Mutex
 }
 
-func (bsm *BroadcastSessionsManager) selectFromList() *BroadcastSession {
-	bsm.checkStatusOfList()
-
-	bsm.sessLock.Lock()
-	defer bsm.sessLock.Unlock()
-
-	if len(bsm.broadcastSessions) > 0 {
-		last := len(bsm.broadcastSessions) - 1
-		sess, sessions := bsm.broadcastSessions[last], bsm.broadcastSessions[:last]
-		bsm.broadcastSessions = sessions
-		return sess
-	}
-	return nil
-}
-
-func (bsm *BroadcastSessionsManager) addToList(sess *BroadcastSession) {
-	bsm.sessLock.Lock()
-	defer bsm.sessLock.Unlock()
-
-	for _, currentSess := range bsm.broadcastSessions {
-		if currentSess == sess {
-			return
-		}
-	}
-	bsm.broadcastSessions = append(bsm.broadcastSessions, sess)
-}
-
-func (bsm *BroadcastSessionsManager) checkStatusOfList() {
-	var newBroadcastSessions []*BroadcastSession
+func (bsm *BroadcastSessionsManager) selectSession() *BroadcastSession {
 	// minOrchs is maximum number of inflight sessions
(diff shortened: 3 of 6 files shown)


server/mediaserver.go
```diff
@@ -399,7 +399,10 @@ func (s *LivepeerServer) startSession(cxn *rtmpConnection) *BroadcastSession {
 	expb.MaxInterval = BroadcastRetry
 	expb.MaxElapsedTime = 0
 	backoff.Retry(broadcastFunc, expb)
-	return sess
+	if len(sess) > 0 {
```
angyangie, line 417: Temporary, until we integrate `BroadcastSessionManager` methods into `processSegment`

server/mediaserver_test.go
```diff
@@ -128,20 +128,21 @@ func TestSelectOrchestrator(t *testing.T) {
 	if sess == nil {
 		t.Error("Expected nil session")
 	}
+
```
j0sh, line 136: Might be good to assert on the expected length here.

server/mediaserver_test.go
```diff
@@ -164,21 +165,46 @@ func TestSelectOrchestrator(t *testing.T) {
...
 
 	assert := assert.New(t)
-	assert.Equal(sender, sess.Sender)
-	assert.Equal(expSessionID, sess.PMSessionID)
+	assert.Equal(sender, sess[0].Sender)
+	assert.Equal(expSessionID, sess[0].PMSessionID)
+	assert.Equal(expSessionID2, sess[1].PMSessionID)
+	assert.Equal(sess[0].OrchestratorInfo, &net.OrchestratorInfo{TicketParams: protoParams})
+	assert.Equal(sess[1].OrchestratorInfo, &net.OrchestratorInfo{TicketParams: protoParams2})
+	assert.Len(sess, 2)
```
j0sh: This check for `assert.Len` probably should be first, after the `assert.New`

j0sh (approved):

server/rpc.go
```diff
@@ -50,6 +51,84 @@ type Broadcaster interface {
...
+	sessLock          *sync.Mutex
+}
+
+func (bsm *BroadcastSessionsManager) selectFromList() *BroadcastSession {
+	bsm.checkStatusOfList()
+
+	bsm.sessLock.Lock()
+	defer bsm.sessLock.Unlock()
+
+	if len(bsm.broadcastSessions) > 0 {
```
j0sh: The flow is nicer if the logic is inverted:

if len(bsm.broadcastSessions) <= 0 {
  return nil
}
...

server/rpc.go
```diff
@@ -50,6 +51,84 @@ type Broadcaster interface {
...
+	bsm.broadcastSessions = append(bsm.broadcastSessions, sess)
+}
+
+func (bsm *BroadcastSessionsManager) checkStatusOfList() {
+	var newBroadcastSessions []*BroadcastSession
+	// minOrchs is maximum number of inflight sessions
+	minOrchs := int(HTTPTimeout / SegLen)
+	reqLengthOfList := minOrchs * 2
+
+	if len(bsm.broadcastSessions) < minOrchs {
```
j0sh: This should still be protected by the lock

server/rpc.go
```diff
@@ -50,6 +51,84 @@ type Broadcaster interface {
...
+		return sess
+	}
+	return nil
+}
+
+func (bsm *BroadcastSessionsManager) addToList(sess *BroadcastSession) {
+	bsm.sessLock.Lock()
+	defer bsm.sessLock.Unlock()
+
+	for _, currentSess := range bsm.broadcastSessions {
```
j0sh: If I'm understanding this correctly, the intent is to prevent double-inserts of the same O [1] . However, during refreshing, the new and old sessions would be different objects and wouldn't match using a simple equality comparison. The comparison probably should  be against something like OrchestratorInfo.Transcoder 

https://github.com/livepeer/go-livepeer/blob/7e1dde56f602c3ca5b8e160b11d20bcc0ae46a85/net/lp_rpc.proto#L80-L83

Also, scanning through the list here while under a lock feels expensive. Could a map be used here in conjunction with the list?

[1] Segments in flight to an O, list gets refreshed and the O gets a new session, then after the in-flight segment returns, the O gets the old session inserted again. At this point O has two sessions.

server/rpc.go
```diff
@@ -50,6 +51,84 @@ type Broadcaster interface {
...
+		// }
+		glog.Info(newBroadcastSessions)
+	} else {
+		return
+	}
+
+	bsm.sessLock.Lock()
+	defer bsm.sessLock.Unlock()
+
+	// include all O's being used successfully in refreshed list
```
j0sh, line 99: I'm afraid this still doesn't account for Os that have segments in flight.

Referring back to the notes around `addToList`, we could maintain a map of segments for quick lookup, and this map would not be modified when pushing / popping from the list. (Very similar, if not identical to what you have planned for multi-T.) This brings up the question of how to distinguish between Os that have failed, and Os that are in flight. There are two options here:

* Reset the map unconditionally when the list is refreshed. This effectively resets the entire list. Simplest, but we may lose "good" Os that we're working with if those Os are already at capacity.

* Restore the `removeFromList` method and make it only remove from the map [1]. With this, the `checkStatusOfList` function becomes something like:

```
newSessions := getSessions(numReqOrchs)
newSessions = newSessions.filter( /* exclude Os that already exist */ )
bsm.broadcastSessions = append(newSessions, bsm.broadcastSessions...) // append newSessions to *front* of existing list
// add newSessions to map
```

[1] We probably should rename these *List methods to reflect the fact that we're manipulating a "working set" of orchestrators; maybe "select" for selectFromList / "complete" for addToList / "remove" for removeFromList and "checkStatus"

server/rpc.go
```diff
@@ -50,6 +51,84 @@ type Broadcaster interface {
...
+			for _, currentSess := range bsm.broadcastSessions {
+				if _, ok := newSessionsMap[currentSess]; ok {
+					delete(newSessionsMap, currentSess)
+				}
+			}
+		}
+
+		if len(newSessionsMap) > 0 {
+			for sess, _ := range newSessionsMap {
+				if len(bsm.broadcastSessions) < reqLengthOfList {
```
j0sh: FWIW, I don't think it does much harm if the `broadcastSessions` list happens to be longer than `reqLengthOfList` as long as there are no duplicates.

server/rpc.go
```diff
@@ -50,6 +51,84 @@ type Broadcaster interface {
...
+	} else {
+		return
+	}
+
+	bsm.sessLock.Lock()
+	defer bsm.sessLock.Unlock()
+
+	// include all O's being used successfully in refreshed list
+	var newSessionsMap map[*BroadcastSession]bool
+	if len(newBroadcastSessions) > 0 {
```
j0sh: We can combine the check for `<=0` with `if len(bsm.broadcastSessions) < minOrchs {` and save one level of indentation here.

server/rpc_test.go
```diff
@@ -532,3 +533,56 @@ func defaultTicket(t *testing.T) *net.Ticket {
...
+	expectedSess := bsm.broadcastSessions[1]
+
+	assert := assert.New(t)
+	assert.Len(bsm.broadcastSessions, 2)
+
+	sess := bsm.selectFromList()
+	assert.Equal(expectedSess, sess)
+	assert.Len(bsm.broadcastSessions, 1)
+
+	bsm = &BroadcastSessionsManager{broadcastSessions: []*BroadcastSession{}, sessLock: &sync.Mutex{}}
```
j0sh: Curious why `bsm` is being reset here? A comment explaining as much would help.

server/rpc_test.go
```diff
@@ -532,3 +533,56 @@ func defaultTicket(t *testing.T) *net.Ticket {
...
+}
+
+func TestAddToList(t *testing.T) {
+	bsm := StubBroadcastSessionsManager()
+	sess := &BroadcastSession{
+		Broadcaster: StubBroadcaster2(),
+		ManifestID:  core.RandomManifestID(),
+	}
+
+	bsm.addToList(sess)
```
j0sh: Would be good to sanity check the previous length

j0sh (changes requested):

server/rpc.go
angyangie: It does indeed. Thanks for pointing it out!

server/rpc.go
angyangie: You're right that the BroadcastSession wouldn't be the same object. I'll use `sess.OrchestratorInfo.Transcoder` instead. 

From our Discord discussion: if we choose to go with the second option you mention in a later comment, restoring the `removeFromList` method to only remove from the map, it would solve our `addToList` issue of having to loop through `bsm.broadcastSessions`. Since we just don't add back an O to `bsm.broadcastSessions` if a segment fails,  the `removeFromList` would remove from the map.

server/rpc.go
angyangie: [From Discord]: Conclusion: if we do this check: `newSessions = newSessions.filter( /* exclude Os that already exist */ )` on the new map of O's, we'd be accounting for all O's, whether or not they had any segments in flight. So, we'd just want to calculate how many O's we'd need based on the mapped list of Os. That list only gets shorter when an O fails. All segments in-flight could theoretically also fail, but at that point, we could refresh the list again.

server/rpc.go
angyangie: True. [From Discord]: With our new method of adding and taking individual Os we're working with from the back of the list, and adding new O's fetched to the front of the list, we're essentially getting to work with the same O's all the time, which is what we want. The length of that list doesn't seem too relevant anymore with this new method

server/rpc.go
angyangie: Good idea!

server/rpc_test.go
angyangie: Come to think of it, I can just add a round of `.selectFromList()` and avoid having to do this! 
```
sess = bsm.selectFromList()
assert.Equal(expectedSess2, sess)
assert.Len(bsm.broadcastSessions, 0)
```

server/rpc.go
```diff
@@ -50,6 +51,85 @@ type Broadcaster interface {
...
+func (bsm *BroadcastSessionsManager) checkStatus() {
+	// newBroadcastSessions, err := selectOrchs(node, pl)
+	// if err != nil {
+	// 	return
+	// }
+
+	newBroadcastSessions := []*BroadcastSession{}
+
+	// include all O's being used successfully in refreshed list
+	if len(newBroadcastSessions) > 0 {
```
j0sh: I think the flow for the rest of this function can be simplified to:

```golang
uniqueSessions := make([]*BroadcastSession, len(newBroadcastSessions)) // minimize allocations
for _, sess := range newBroadcastSessions {
  if _, ok := bsm.broadcastSessMap[sess.OrchestratorInfo.Transcoder]; ok {
    continue
  }
  uniqueSessions = append(uniqueSessions, sess)
  bsm.broadcastSessMap[sess.OrchestratorInfo.Transcoder] = sess
}
bsm.broadcastSessions = append(uniqueSessions, bsm.broadcastSessions...)
```

server/rpc_test.go
```diff
@@ -532,3 +533,104 @@ func defaultTicket(t *testing.T) *net.Ticket {
...
+	sess1 := bsm.broadcastSessions[0]
+	sess2 := bsm.broadcastSessions[1]
+
+	assert := assert.New(t)
+	assert.Len(bsm.broadcastSessMap, 2)
+
+	bsm.removeSession(sess1)
+	assert.Len(bsm.broadcastSessMap, 1)
+
+	bsm.removeSession(sess1)
```
j0sh, line 599: Looking great -- maybe a comments along the lines of `// remove nonexistent session` would help here. Also it might generally be good to do pre/post checks that the sessions are actually in the map, eg

```
assert.NotNil(bsm.broadcastSessMap[sess])
bsm.removeSession(sess1)
assert.Nil(bsm.broadcastSessMap[sess])
```

server/rpc.go
```diff
@@ -50,6 +51,85 @@ type Broadcaster interface {
...
+	return sess
+}
+
+func (bsm *BroadcastSessionsManager) removeSession(session *BroadcastSession) {
+	bsm.sessLock.Lock()
+	delete(bsm.broadcastSessMap, session.OrchestratorInfo.Transcoder)
+	bsm.sessLock.Unlock()
+}
+
+func (bsm *BroadcastSessionsManager) completeSessions(sess *BroadcastSession) {
```
j0sh: Perhaps `completeSession` would be a more accurate name

angyangie: I don't have a preference between these two. `completeSession` sounds good to me!

server/rpc.go
angyangie: Much more concise! This works

server/rpc.go
```diff
@@ -50,6 +51,64 @@ type Broadcaster interface {
...
+	if _, ok := bsm.broadcastSessMap[sess.OrchestratorInfo.Transcoder]; ok {
+		bsm.broadcastSessions = append(bsm.broadcastSessions, sess)
+	}
+}
+
+func (bsm *BroadcastSessionsManager) checkStatus() {
+	newBroadcastSessions := []*BroadcastSession{}
+
+	// include all O's being used successfully in refreshed list
+	if len(newBroadcastSessions) > 0 {
```
j0sh: I don't think this length check is really needed anymore. If we keep it, the check should exit early rather than introduce in another level of indentation, eg

```golang
if len(newBroadcastSessions) <= 0 {
  return
}
uniqueSessions := ...
```

server/rpc.go
```diff
@@ -50,6 +51,64 @@ type Broadcaster interface {
...
+}
+
+func (bsm *BroadcastSessionsManager) selectSession() *BroadcastSession {
+	bsm.sessLock.Lock()
+	defer bsm.sessLock.Unlock()
+
+	// minOrchs is maximum number of inflight sessions
+	minOrchs := int(HTTPTimeout / SegLen)
+	if len(bsm.broadcastSessions) < minOrchs {
+		bsm.checkStatus()
```
j0sh: Note a small complication here.

* We really want the checkStatus call to be async because it could take a while.
* We need to ensure proper locking within checkStatus, especially where we access and mutate the existing list of sessions.
* Lastly, "checkStatus" is starting to seem like an odd name now. Perhaps a more accurate naming would be "refreshSessions" or some such.

This would work:
```golang
  // minOrchs is maximum number of inflight sessions
  minOrchs := int(HTTPTimeout / SegLen)
  bsm.sessLock.Lock()
  defer bsm.sessLock.Unlock()
  numSess = len(bsm.broadcastSessions) // we use this a bunch of other places
  if numSess < minOrchs {
    go bsm.refreshSessions()
  }
  ... rest of function here ...
}

func (bsm *BroadcastSessionsManager) refreshSessions() {
  newBroadcastSessions := ...
  if (len(newBroadcastSessions) <= 0) {
    return
  }
  uniqueSessions := make([]*BroadcastSession, 0, len(newBroadcastSessions))
  bsm.sessLock.Lock()
  defer bsm.sessLock.Unlock()
  ... rest of function here ... 
}
```

j0sh (changes requested):

server/rpc.go
angyangie: I placed that check in there in case `newBroadcastSessions := []*BroadcastSession{}` which will become `newBroadcastSessions, err := selectOrchestrator(node, pl)` comes back empty, so as to not go through the following loop. I'll have it exit early and not intro more indentation! Is less indentation generally seen as more "golang-y"?

server/rpc.go
angyangie: Ah I see what you're doing there. I misunderstood locks a bit here - I thought they were "carried along" to every function that was called within another function. I see now that we need to use a lock again in `refreshSessions`. `refreshSessions` works for me!

server/rpc.go
j0sh: If `len(newBroadcastSessions) == 0` then the loop wouldn't have anything to execute through, but it doesn't hurt to exit early here. See here for an example: https://play.golang.org/p/5JbDbGAFkjp

> Is less indentation generally seen as more "golang-y"?

It's generally better to exit early if possible; it makes the flow more "straight through" more and easier to follow. Deep nesting is also a bit of a code smell.

server/rpc.go
j0sh: >  I misunderstood locks a bit here - I thought they were "carried along" to every function that was called within another function

That is correct -- locks are "carried through" until unlocked somewhere. (The unlocking doesn't necessarily have to be in the same part of the call stack or even the same thread, but this naturally makes things difficult to reason about, so it's preferable to keep locking/unlocking localized where possible.)

The reason we need to lock within refreshSessions is because it's being called asynchronously, and we can't count on the lock from `selectSession` to fully "carry through". For example:

```
selectSession : Acquire lock
refreshSessions : Start execution
selectSession : Release lock
refreshSessions : Stop execution
```

So here we need to lock in both selectSessions and refreshSessions in order to serialize accesses to the set of sessions.

angyangie: Oh I see. That makes more sense! Thanks for the explanation Josh

PR #804, opened 2019-03-28
Small Error Fix

ericxtang:

Small fix - we were returning erorrs when a sender is not found.  This happens naturally when a new node is starting and a new Eth addr is used.

Just return an empty struct instead of an error to the cli.

By hand.

### DIFF
--- a/server/handlers.go
+++ b/server/handlers.go
@@ -215,8 +215,20 @@ func senderInfoHandler(client eth.LivepeerEthClient) http.Handler {
 
 		sender, err := client.Senders(client.Account().Address)
 		if err != nil {
-			respondWith500(w, fmt.Sprintf("could not query sender info: %v", err))
-			return
+			if err.Error() == "ErrNoResult" {
+				type Sender struct {
+					Deposit       *big.Int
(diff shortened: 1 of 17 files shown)


server/handlers.go
```diff
@@ -215,8 +215,20 @@ func senderInfoHandler(client eth.LivepeerEthClient) http.Handler {
 
 		sender, err := client.Senders(client.Account().Address)
 		if err != nil {
-			respondWith500(w, fmt.Sprintf("could not query sender info: %v", err))
-			return
+			if err.Error() == "ErrNoResult" {
+				type Sender struct {
```
j0sh, line 219: Just curious, if the upstream Sender struct ever changes, will the compiler catch it here and prompt for an update?

j0sh (approved):
LGTM

server/handlers.go
ericxtang: I think it does
Screen Shot 2019-04-01 at 7 15 18 PM

j0sh (approved):
Looks good - might want to rebase before merging this because there'll be an extraneous merge commit here. :ship:

PR #805, opened 2019-03-28
Improve simple auth server

ericxtang:

Simple auth server improvement for demo purposes (match new RTMP stream ID format)

### DIFF
--- a/cmd/simple_auth_server/simple_auth_server.go
+++ b/cmd/simple_auth_server/simple_auth_server.go
@@ -3,44 +3,57 @@ package main
 import (
 	"encoding/json"
 	"fmt"
-	"io/ioutil"
 	"log"
 	"net/http"
 	"net/url"
+	"path"
+	"regexp"
+	"strings"
+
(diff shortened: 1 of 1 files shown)


j0sh:
Note that the auth endpoint isn't required to match the default interpretation of the RTMP path. One of our strengths here is that developers can interpret the RTMP URL however best suits their application.

In that case, I don't mind a server which does something "different" just to exhibit the versatility of the system. Eg, a fizzbuzz authenticator:

```
// Only allow RTMP streams with "fizz" as the path.
//   Eg, `rtmp://localhost/fizz`
// Produces a HLS stream with of the name "buzz"
//    Eg, `http://localhost:8935/stream/buzz.m3u8`
// Disallow all other streams

if "fizz" == url.Path
  manifestID = "buzz"
else return 404
```

This seems like it's a nicer representation of the intent and capabilities of the webhooks system. What do you think?

ericxtang:
@j0sh made the change to update manifestID when string matches "fizz"

j0sh (approved):
:ship:

Issue #667, opened 2019-01-07
Use InsertSegment in HLS playlist


j0sh:
**Is your feature request related to a problem? Please describe.**
Currently the playlist manager uses `AppendSegment` to append a new segment to the end of the playlist. This works OK for most purposes.

There is a function in the m3u8 package, [InsertSegment](https://github.com/ericxtang/m3u8/blob/575aeb9f754a5dabcc03d4aa0ed05ecaee26499e/writer.go#L329), that attempts to incorporate some nice properties.

* Thread safety on inserts
* Sliding windows (reimplemented by playlist manager)
* Rearrange out of order insertions
* Rejects duplicate sequence numbers
* Sets playlist SeqNo for `EXT-X-MEDIA-SEQUENCE` (special cased for the first segment by the playlist manager)

Using `InsertSegment` instead has the potential to save quite a bit of code within the PlaylistManager and better compartmentalize the above concerns [1].

However, the `InsertSegment` implementation needs some more work before it's ready for prime time:

*  Handling of sliding windows needs to be improved  -- doesn't handle wraparounds of the preallocated array, and crashes
* Not yet fully thread safe around accesses of the segment list
* Potentially re-sorting the segment list on every insert appears to make this problematic for VOD -- complexity becomes O(n^2 * logn) under pathological conditions where every segment is incorrectly numbered. We probably could optimize this later.

[1] Potential PlaylistManager simplifications from using InsertSegment
```diff
commit 3c4ddd3f13537f5d77e945a03afc7a84f04bb89b
Author: Josh Allmann <joshua.allmann@gmail.com>
Date:   Fri Jan 4 22:12:43 2019 -0800

    playlist: Use InsertSegment rather than AppendSegment.
    
    InsertSegment gives us the following features:
    
    * Thread safe playlists
    * Sliding window playlists
    * Rearranges out-of-order insertions
    * Rejects duplicate segments
    * Sets playlist SeqNo for EXT-X-MEDIA-SEQUENCE

diff --git a/core/playlistmanager.go b/core/playlistmanager.go
index b9a233cf..a717e0e7 100644
--- a/core/playlistmanager.go
+++ b/core/playlistmanager.go
@@ -95,21 +95,8 @@ func (mgr *BasicPlaylistManager) InsertHLSSegment(profile *ffmpeg.VideoProfile,
 	if err != nil {
 		return err
 	}
-	return mgr.addToMediaPlaylist(uri, seqNo, duration, mpl)
-}
-
[...]

PR #806, opened 2019-03-29
Adding broadcast session manager methods to process segment

angyangie:

**NOTE**: Still need to test in test-harness

Integrate `BroadcastSessionManager` and associated methods into our existing codebase - this is a rough MVP  implementation of a round-robin strategy for multi-O.

- [x] Add `sessManager` to `cxn := &rtmpConnection{}` in `registerConnection`
- [x] Add sessions coming back from `sess := s.startSession(cxn)` in `startSessionListener` into `cxn.sessManager.broadcastSessions = sess`
- [x] Integrate `cxn.sessManager.removeSession(sess)` and `cxn.sessManager.completeSession(sess)` into `processSegment`.
- [x] Add more robust testing around `bsm.checkStatus`, monkey-patching `selectOrchestrator`
- [x] Add to mediaserver_test.go

- [x] Running nodes locally
- [x] Running nodes in test-harness

More details under the `Task Breakdown: Engineering` section of this document, under `Multi-O breakdown`, step # 4  : 

https://docs.google.com/document/d/1KfEWm0z3svw10TrFH9hiMImkXocl3ba_S-8vmX94HNc/edit#heading=h.1mjk0mji9b8w

**Checklist:**


server/mediaserver.go
```diff
@@ -368,7 +369,11 @@ func (s *LivepeerServer) registerConnection(rtmpStrm stream.RTMPVideoStream) (*r
 		pl:      core.NewBasicPlaylistManager(mid, storage),
 		profile: &vProfile,
 		lock:    &sync.RWMutex{},
-
+		sessManager: &BroadcastSessionsManager{
+			broadcastSessions: []*BroadcastSession{},
+			sessLock:          &sync.Mutex{},
+			broadcastSessMap:  make(map[string]*BroadcastSession),
```
j0sh: I think we can refine how we do this. In particular, let's:

* Clean  up initialization of the BroadcastSessionManager struct
* Eliminate passing around the LivepeerNode and the playlist, which is only there to propagate down to `selectOrchs` which makes it a bit smelly.
* Double down on the concept of abstracting away the `selectOrchs` function for testability / reusability
* Maybe eliminate the need for the code around startSession / startSessionListener.

```golang
type BroadcastSessionsManager  struct {
  sessions []*BroadcastSession{}
  sessionMap map[string]*BroadcastSession
  lock sync.Mutex
  createSessions func() []*BroadcastSession{}
}

func NewSessionManager(n *core.LivepeerNode, pl *core.PlaylistManager) *BroadcastSessionManager {
  m := &BroadcastSessionManager {
    // we really don't have to explicitly define additional fields here; zero values are fine
    // given the field definitions in BroadcastSessionManager above
    sessionMap: make(map[string]*BroadcastSession),
    createSessions: func() []*BroadcastSessionManager {
      return selectOrchestrator(node, pl)
    },
  }
  m.refreshSessions()
  return m
}

...

cxn := &rtmpConnection {
  ...
  sessManager = NewSessionManager(node, pl)
}
```

j0sh (changes requested):

server/broadcast.go
```diff
@@ -82,7 +88,7 @@ func processSegment(cxn *rtmpConnection, seg *stream.HLSSegment) {
 	vProfile := cxn.profile
 
 	cxn.lock.RLock()
```
j0sh: We won't need to lock here anymore since `selectSession` gives us atomicity

server/broadcast.go
```diff
@@ -82,7 +88,7 @@ func processSegment(cxn *rtmpConnection, seg *stream.HLSSegment) {
...
-	sess := cxn.sess
+	sess := cxn.sessManager.selectSession(node, cxn.pl)
```
j0sh: This call can probably be moved down to line 117. That allows us to avoid selecting a session until we've completed any uploads to our local OS.

server/broadcast.go
```diff
@@ -150,6 +157,8 @@ func processSegment(cxn *rtmpConnection, seg *stream.HLSSegment) {
 				cxn.needOrch <- struct{}{}
 			}
 			return
+		} else {
```
j0sh: The `else` is not really needed since there is a `return` above. This goes back to the earlier notes about minimizing code under conditionals and maximizing the straight-through code flow.

server/broadcast.go
angyangie: True! This prompted me to notice something else - even if we move to line 117, if `uri, err := ios.SaveData(name, seg.Data)` fails (line 134), the session wouldn't be added back to our list (`cxn.sessManager.completeSession(sess)`), to no fault of the O. I'll add `cxn.sessManager.completeSession(sess)` to line 140 right before the return statement.

server/mediaserver.go
angyangie: This is very helpful feedback Josh! Always making me a better engineer

server/rpc.go
```diff
@@ -50,6 +51,88 @@ type Broadcaster interface {
...
+func (bsm *BroadcastSessionsManager) completeSession(sess *BroadcastSession) {
+	bsm.sessLock.Lock()
+	defer bsm.sessLock.Unlock()
+
+	if _, ok := bsm.broadcastSessMap[sess.OrchestratorInfo.Transcoder]; ok {
+		bsm.broadcastSessions = append(bsm.broadcastSessions, sess)
+	}
+}
+
+var selectOrchs = func(n *core.LivepeerNode, pl core.PlaylistManager) ([]*BroadcastSession, error) {
```
j0sh: I don't think this is used anywhere?

server/mediaserver.go
```diff
@@ -59,11 +59,12 @@ var BroadcastJobVideoProfiles = []ffmpeg.VideoProfile{ffmpeg.P240p30fps4x3, ffmp
 var AuthWebhookURL string
 
 type rtmpConnection struct {
```
j0sh, line 60: Do we still need the following fields within `rtmpConnection`? I think we can remove these and the associated code:

* needOrch
* eof
* sess
* lock

We can remove startSession/startSessionListener as well. All this has been subsumed by the session manager.

Note that the startSessionListener has an important property that the session manager doesn't currently have: it only allows for one refresh in-flight at a given time.

server/mediaserver.go
```diff
@@ -436,7 +439,10 @@ func (s *LivepeerServer) startSessionListener(cxn *rtmpConnection) {
 		}
 		mut.Lock()
 		defer mut.Unlock()
-		cxn.sess = sess
+		cxn.sessManager.broadcastSessions = sess
+		for _, s := range sess {
+			cxn.sessManager.broadcastSessMap[s.OrchestratorInfo.Transcoder] = s
+		}
```
j0sh: This part has been subsumed by the `refreshSession` call in NewSessionManager so this isn't needed anymore.

server/rpc.go
```diff
@@ -50,6 +51,88 @@ type Broadcaster interface {
 	Sign([]byte) ([]byte, error)
 }
 
+type BroadcastSessionsManager struct {
```
j0sh: We don't currently use this struct within `rpc.go` , should we put it somewhere else? I'm thinking either broadcast.go, mediaserver.go or its own file.

server/rpc.go
```diff
@@ -50,6 +51,88 @@ type Broadcaster interface {
...
+	broadcastSessions []*BroadcastSession
+	broadcastSessMap  map[string]*BroadcastSession
```
j0sh: Naming is a little inconsistent throughout here.

* sessions / sess
* BroadcastSessionManager.broadcastSessions feels a little long and redundant.

FWIW, I prefer brevity in my naming but I'll let you call the scheme as long as it's consistent. Some suggestions: list / table / lock / createSessions

server/rpc.go
```diff
@@ -50,6 +51,88 @@ type Broadcaster interface {
...
+	return selectOrchestrator(n, pl)
+}
+
+func (bsm *BroadcastSessionsManager) refreshSessions() {
+	newBroadcastSessions, err := bsm.createSessions()
+	if err != nil {
+		return
+	}
+
+	// include all O's being used successfully in refreshed list
```
j0sh: This comment is unclear; I think here it's just exiting early if we receive an empty list from `createSessions` ?

server/mediaserver_test.go
```diff
@@ -770,24 +813,25 @@ func TestSessionAssigment_WithTerminatedConnection(t *testing.T) {
...
+		eof:         make(chan struct{}),
+		pl:          core.NewBasicPlaylistManager(mid, storage),
+		stream:      strm,
+		sessManager: &BroadcastSessionsManager{broadcastSessions: []*BroadcastSession{}, sessLock: &sync.Mutex{}},
 	}
 
 	// sanity: nil session
 	cxn.lock.Lock()
-	assert.Nil(cxn.sess) // nil session
+	assert.Len(cxn.sessManager.broadcastSessions, 0) // nil session
```
j0sh: Following the sessManager API might be cleaner here:

```golang
sess := cxn.sessManager.selectSession()
assert.Nil(sess)
```

server/rpc.go
angyangie: hmm don't we use it in `NewSessionManager`, and have associated `BroadcastSessionsManager` methods in the same file - `selectSession`, `removeSession`, etc? Or, do you mean something else?

server/mediaserver.go
angyangie: Really great observations! Helping me be more vigilant about looking at existing code and reviewing which has become superfluous

server/broadcast_test.go
```diff
@@ -1,41 +0,0 @@
-package server
```
j0sh: I'm not sure why this file was removed. Fixed locally; force pushing.

server/rpc.go
```diff
@@ -63,19 +68,18 @@ func NewSessionManager(node *core.LivepeerNode, pl core.PlaylistManager) *Broadc
...
 
 func (bsm *BroadcastSessionsManager) selectSession() *BroadcastSession {
-	// minOrchs is maximum number of inflight sessions
-	minOrchs := int(HTTPTimeout / SegLen)
 	bsm.sessLock.Lock()
 	defer bsm.sessLock.Unlock()
 	numSess := len(bsm.sessList)
 
-	if numSess < minOrchs {
+	if numSess < bsm.minOrchs {
```
j0sh: This makes testing a bit easier (cleans up some race conditions since refreshSessions will no longer be triggered by default) and also gives us an avenue to handle the case where the session list is naturally shorter than the default minOrchs.

server/rpc.go
```diff
@@ -104,11 +108,16 @@ func (bsm *BroadcastSessionsManager) completeSession(sess *BroadcastSession) {
 	}
 }
 
-var selectOrchs = func(n *core.LivepeerNode, pl core.PlaylistManager) ([]*BroadcastSession, error) {
```
j0sh: This was unused

server/rpc_test.go
```diff
@@ -559,14 +559,14 @@ func StubBroadcastSessionsManager() *BroadcastSessionsManager {
 	}
 }
 
-func TestSelect(t *testing.T) {
+func TestSelectSession(t *testing.T) {
```
j0sh: We already have a `TestSelectOrchestrator` which has issues with the match done if running `go test -run Select`. This was mostly problematic because SelectOrchestrator is currently not concurrent safe under `-race`.

server/rpc_test.go
```diff
@@ -559,14 +559,14 @@ func StubBroadcastSessionsManager() *BroadcastSessionsManager {
...
 	bsm := StubBroadcastSessionsManager()
-	expectedSess1 := bsm.sessList[1]
-	expectedSess2 := bsm.sessList[0]
 
 	// assert that initial lengths are as expected
 	assert := assert.New(t)
 	assert.Len(bsm.sessList, 2)
+	expectedSess1 := bsm.sessList[1]
+	expectedSess2 := bsm.sessList[0]
```
j0sh: Moved this because if the session list happens to be less than length 2 then we'll segfault before the assert.Len

server/mediaserver_test.go
```diff
@@ -668,6 +668,11 @@ func TestRegisterConnection(t *testing.T) {
 
 }
 
+// XXX check that broadcast session manager is:
```
j0sh: Just a reminder for current gaps in test coverage

server/rpc.go
j0sh: We don't actually call any of this in `rpc.go` though, so I'm not sure that this really needs to live there.

server/rpc.go
angyangie: True! Makes testing `refreshSessions` much easier given differing #s of registered Orchs

server/rpc.go
angyangie: Oh I see what you mean. Most of these methods are used in `Broadcast.go` ... perhaps we put them there?

angyangie: `NewSessionManager` could go in mediaserver.go

server/broadcast.go
```diff
@@ -22,56 +22,155 @@ import (
...
+
+	createSessions func() ([]*BroadcastSession, error)
+}
+
+func (bsm *BroadcastSessionsManager) selectSession() *BroadcastSession {
+	bsm.sessLock.Lock()
+	defer bsm.sessLock.Unlock()
+	numSess := len(bsm.sessList)
+
+	if numSess < bsm.minOrchs {
```
j0sh: How we handle minOrchs is a bit problematic right now. I think this would cause a refresh to be fired every time an O fails or if the number of Os in the pool is less than `minOrchs` (4 right now?) - this would be common during offchain scenarios and testing.

Also, since 4 is the maximum number of inflight sessions, we want to give ourselves some more overhead there, so let's say 8.

We should fix this with something like `minOrchs=min(OrchestratorPool.Size(), HTTPTimeout / SegLen * 2)`, and set the refresh threshold to something like `refreshThreshold=ceil(minOrchs/2)`. This would allow the refresh threshold to scale along with the pool size up to 8 orchestrators.

O Pool Size | Quantity to Refresh (minOrchs) | Refresh Below (numSessions < refreshThreshold)
-- | -- | --
1 | 1 | 1
2 | 2 | 1
3 | 3 | 2
4 | 4 | 2
5 | 5 | 3
6 | 6 | 3
7 | 7 | 4
8 | 8 | 4
9 | 8 | 4
10 | 8 | 4
11 | 8 | 4
12 | 8 | 4

j0sh:
* Rebased onto master and fixed a merge conflict
* Added a few commits addressing https://github.com/livepeer/go-livepeer/pull/806#discussion_r274187906
* Also moved some more code into broadcast.go / broadcast_test.go

server/broadcast.go
```diff
@@ -22,63 +23,177 @@ import (
 	"github.com/livepeer/lpms/stream"
 )
 
-func selectOrchestrator(n *core.LivepeerNode, cpl core.PlaylistManager) (*BroadcastSession, error) {
+type BroadcastSessionsManager struct {
+	sessList []*BroadcastSession
+	sessMap  map[string]*BroadcastSession
+	sessLock *sync.Mutex
+	numOrchs int // how many orchs to request at once
```
j0sh, line 32: Renamed this from `minOrchs` for clarity.

server/broadcast.go
```diff
@@ -147,6 +144,7 @@ func processSegment(cxn *rtmpConnection, seg *stream.HLSSegment, node *core.Live
 				if monitor.Enabled {
 					monitor.LogSegmentUploadFailed(nonce, seg.SeqNo, monitor.SegmentUploadErrorOS, err.Error())
 				}
+				cxn.sessManager.completeSession(sess)
```
j0sh: Pushing the O back on the list after an upload failure may be an issue. Consider the following:

```
O1 uses S3
O2 uses direct upload
S3 goes down
Segment 1 comes in
B selects O1
S3 upload fails
B adds O1 back to beginning of list
Segment 2 comes in
B selects O1
S3 upload fails
B adds O1 back to beginning of list
...
```
By removing O1 from the list outright, we can fall back to O2.

angyangie: You're right - `cxn.sessManager.completeSession(sess)` should be removed from here in the event of consistent OS failures

angyangie: Can be replaced with `cxn.sessManager.removeSession(sess)`

server/broadcast.go
```diff
@@ -233,18 +232,33 @@ func processSegment(cxn *rtmpConnection, seg *stream.HLSSegment) {
...
-	// View-only (non-transcoded) streams or mid-failover
+	// View-only (non-transcoded) streams or no sessions available
 	if sess == nil {
 		if monitor.Enabled {
 			monitor.LogSegmentTranscodeFailed(monitor.SegmentTranscodeErrorNoOrchestrators, nonce, seg.SeqNo, errors.New("No Orchestrators Error"))
 		}
-		return
+		return nil
 	}
+	{
```
j0sh, line 282: This is to save the indentation from gofmt in order to make the actual differences in this commit easier to eyeball. We can do a separate commit for cosmetics later.

j0sh:
Added rudimentary retries. Missing tests.

j0sh:
Note that with retries, we are more prone to receiving segments out of order. Fixing https://github.com/livepeer/go-livepeer/issues/667 is more important now.

angyangie:
I'll work on out of order segments now!

j0sh:
* Rebased on top of latest master.

There were some odd merge conflicts, so fixed those up. This rewrote most of the commit hashes, so delete the local branch prior to fetching the updated remote.

* https://github.com/livepeer/go-livepeer/pull/806/commits/511fdfa24550f0e68280b12be342dd734ab4df86 Thread safe error handling during refreshes.

Also rearranged+commented the fields a bit in the
BroadcastSessionManager struct to hopefully make it more
explicit that basically everything in that struct needs to be
protected by a lock.

* https://github.com/livepeer/go-livepeer/pull/806/commits/7c30eff2dd84b8c0bc70464cc57df89def13e347 Don't select sessions that are no longer in the sessions map.

Retry if the first selected session has been removed from the map.
This may occur if the session is removed while still in the list.
To avoid a runtime search of the session list under lock, simply
fixup the session list at selection time by retrying the selection.

This only becomes an issue in the next commit, where we may be
removing sessions from the map *after* we 'provisionally' add
them back to the list after a successful transcode.

* https://github.com/livepeer/go-livepeer/pull/806/commits/2dcf826485969be9415cc0b93b2237d867ecc751 Retry transcode on failed downloads.

Since the download is out of our control, we also remove the
session from the manager lest it fail again on the next retry.

Also recognize that errors saving data to our OS may cause issues
with sig hash checking due to the early exit prior to hashing.
However, such errors do not result in a retry, since they are
most likely local errors, rather than errors from the transcoder.

j0sh:
Rebased onto master.

Added some more logging, fixed up into the "nearest" commits:
* Segment refresh begin / end (debug)
* When a segment begins processing (debug)
* When we fail to select an O (eg, O list is empty) and we drop the segment (normal / info).

j0sh:
Reworked the commits to make things presentable for merging. The history now reflects more of the finished product with a clearer evolution from start-to-finish.

j0sh (approved):
OK to :ship: if folks are comfortable with the results from testing this.

angyangie:
Just tested again locally, killing O's and T's in rapid succession. No errors in transcoding. I'm comfortable merging this. @ericxtang @darkdarkdragon thoughts?

ericxtang (approved):
I've tested a bunch yesterday.  I'm good with this.  🚢

PR #807, opened 2019-04-03
Option to forward all logs to Grafana’s Loki log aggregator

darkdarkdragon:
Adds forwarding of logs to [Loki](https://github.com/grafana/loki).

Currently our logs gets sent to stderr. This update takes intercepts data flow into stderr and send it to [Loki](https://github.com/grafana/loki) server, while also sending it to original stderr. There is no official separate client to Loki, only one that is in main Loki repository itself, but it can't be used as lib - it panics in runtime, so I've cut out minimal client implementation and put it to separate [repo](https://github.com/livepeer/loki-client).

Manually

**Checklist:**

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -30,6 +30,9 @@ jobs:
       - run: go get -u -v go.opencensus.io/stats
       - run: go get -u -v go.opencensus.io/tag
       - run: go get -u -v go.opencensus.io/exporter/prometheus
+      - run: go get -u -v github.com/gogo/protobuf/proto
+      - run: go get -u -v github.com/golang/snappy
+      - run: go get -u -v github.com/livepeer/loki-client/client
 
       - run:
           name: Lint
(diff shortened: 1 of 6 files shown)


j0sh:
What's the advantage to integrating Loki here rather than forwarding the logs externally via the system/container, like a typical log aggregator?

The Loki project also seems very new, would be good to be certain that we'll get a lot of mileage out of this integration. Client crashes aside, is the API stable?

darkdarkdragon:
@j0sh I've outlined my reasoning [here](https://github.com/livepeer/go-livepeer/issues/808)
Yes, it's very new, but it's integration with Prometheus and Grafana makes it already very useful. 
Client is pretty simple (and I'll simplify it even more - need to get rid of protobuf protocol and switch it to json), so I don't see much risk of crashing in it.
Not, API is not stable - but in my implementation logs gets sent both to Loki and Stackdriver, so if something goes wrong and Loki will stop working - logs still be in Google's Cloud.

darkdarkdragon:
Closing this in favor of external logging forwarding [solution](https://github.com/livepeer/test-harness/commit/0d93a3074bace8c8e95005e739da9a3810d652b4)

PR #813, opened 2019-04-09
Ar/add unique os session per orchestrator

angyangie:

When B-supplied OS is in the picture: Since segment paths are predictable, competing Os could attempt to "front run" or overwrite other Os by inserting bogus data for segments they aren't assigned to. This is possible because B currently creates a single OS session with just the manifestID as the path. To fix this, we are generating a fresh external OS session for each O using a prefix such as "manifestID/$randomPrefixPerOrch" to avoid these types of issues. 

Unit tests

Step 5 in Multi-O: https://docs.google.com/document/d/1KfEWm0z3svw10TrFH9hiMImkXocl3ba_S-8vmX94HNc/edit#

**Checklist:**
- [ X ] All tests in `./test.sh` pass

### DIFF
--- a/server/broadcast.go
+++ b/server/broadcast.go
@@ -62,13 +62,20 @@ func selectOrchestrator(n *core.LivepeerNode, cpl core.PlaylistManager) (*Broadc
 		orchOS = drivers.NewSession(tinfo.Storage[0])
 	}
 
+	bcastOS := cpl.GetOSSession()
+	if bcastOS.IsExternal() {
+		// Give each O its own OS session to prevent front running uploads
+		pfx := fmt.Sprintf("%v/%v", cpl.ManifestID(), core.RandomManifestID())
+		bcastOS = drivers.NodeStorage.NewSession(pfx)
+	}
+
 	return &BroadcastSession{
 		Broadcaster:      rpcBcast,
 		ManifestID:       cpl.ManifestID(),
 		Profiles:         BroadcastJobVideoProfiles,
 		OrchestratorInfo: tinfo,
 		OrchestratorOS:   orchOS,
-		BroadcasterOS:    cpl.GetOSSession(),
+		BroadcasterOS:    bcastOS,
 		Sender:           n.Sender,
 		PMSessionID:      sessionID,
 	}, nil


server/rpc.go
```diff
@@ -60,6 +61,7 @@ type BroadcastSession struct {
 	BroadcasterOS    drivers.OSSession
 	Sender           pm.Sender
 	PMSessionID      string
+	Storage          drivers.OSSession
```
j0sh: This is essentially the same as `BroadcasterOS`

server/broadcast.go
```diff
@@ -187,17 +193,25 @@ func processSegment(cxn *rtmpConnection, seg *stream.HLSSegment) {
 
 	seg.Name = "" // hijack seg.Name to convey the uploaded URI
 	name := fmt.Sprintf("%s/%d.ts", vProfile.Name, seg.SeqNo)
-	uri, err := cpl.GetOSSession().SaveData(name, seg.Data)
+	sess := cxn.sessManager.selectSession()
```
j0sh: The downside of moving `selectSession` here is we're making the O unavailable while we upload the segment into our own OS.

We can still maintain different OS prefixes for our uploads and O uploads. This would mean an upload structure of:

```
# For source segments
<source-mid>/source/<num>.ts

# For transcoded renditions
<source-mid>/<orch-mid>/<rendition>/<num>.ts
```
This also mitigates the possibility of an O front-running our own uploads. The structure is not as nice but I'm not sure of an easy alternative. We could add in support for signed upload URLs one day.

j0sh (changes requested):
I think we can achieve essentially the same effect with a more focused change in place of https://github.com/livepeer/go-livepeer/pull/813/commits/748625a02f8a417c32667daef8ee5ca478f42001:

```diff
diff --git a/server/broadcast.go b/server/broadcast.go
index f5bceeaf..ef0ec679 100644
--- a/server/broadcast.go
+++ b/server/broadcast.go
@@ -63,13 +63,20 @@ func selectOrchestrator(n *core.LivepeerNode, cpl core.PlaylistManager) ([]*Broa
 			orchOS = drivers.NewSession(tinfo.Storage[0])
 		}
 
+		bcastOS := cpl.GetOSSession()
+		if bcastOS.IsExternal() {
+			// Give each O its own OS session to prevent front running uploads
+			pfx := fmt.Sprintf("%v/%v", cpl.ManifestID(), core.RandomManifestID())
+			bcastOS = drivers.NodeStorage.NewSession(pfx)
+		}
+
 		session := &BroadcastSession{
 			Broadcaster:      rpcBcast,
-			ManifestID:       cpl.ManifestID(),
+			ManifestID:       cpl.ManifestID(), // XXX this might also be problematic
 			Profiles:         BroadcastJobVideoProfiles,
 			OrchestratorInfo: tinfo,
 			OrchestratorOS:   orchOS,
-			BroadcasterOS:    cpl.GetOSSession(),
+			BroadcasterOS:    bcastOS,
 			Sender:           n.Sender,
 			PMSessionID:      sessionID,
 		}
```

Reusing the ManifestID between Os is an issue because of https://github.com/livepeer/go-livepeer/issues/695 . Looked at it a bit and it seems that we *should* be OK if we just give each O a fresh ManifestID, or we can fix #695.

j0sh (approved):
LGTM 🚢

PR #815, opened 2019-04-11
lpms fix by Josh

ericxtang:
LPMS update from https://github.com/livepeer/lpms/pull/113 by @j0sh


j0sh:
* Updated the commit message to reflect the LPMS commit hash
* Fixed tests that changed as a result  - LPMS emits more data as a result of flushing, so some size checks needed to be updated

Although as noted in https://github.com/livepeer/lpms/pull/113 , I don't think that the backward-timestamp issue is completely resolved yet.

ericxtang:
Thanks for fixing up the tests @j0sh.  Do you think this can be merged even though the backward-timestamp issue isn't completely solved?

j0sh:
I'm not convinced that the fix for the backward timestamp is really correct, but if it seems to help then we can merge it and see if any further issues come up in this area.

ericxtang:
Sounds good.  In that case, can someone approve so we can merge?

j0sh (approved):

PR #816, opened 2019-04-15
report 100% success rate in case there is no new data

darkdarkdragon:
report 100% success rate in case there is no running streams
(this is needed because Prometheus will continue to show last value
if there is no new data)

- 
- 
- 

Manually

**Checklist:**

### DIFF
--- a/server/broadcast.go
+++ b/server/broadcast.go
@@ -165,11 +165,13 @@ func processSegment(cxn *rtmpConnection, seg *stream.HLSSegment) {
 
 		// download transcoded segments from the transcoder
 		gotErr := false // only send one error msg per segment list
+		var errCode monitor.SegmentTranscodeError
 		errFunc := func(subType monitor.SegmentTranscodeError, url string, err error) {
 			glog.Errorf("%v error with segment %v: %v (URL: %v)", subType, seg.SeqNo, err, url)
 			if monitor.Enabled && !gotErr {
 				monitor.LogSegmentTranscodeFailed(subType, nonce, seg.SeqNo, err)
 				gotErr = true
+				errCode = subType
 			}
 		}
 		if res == nil {
@@ -236,7 +238,7 @@ func processSegment(cxn *rtmpConnection, seg *stream.HLSSegment) {
 		}
 		cond.L.Unlock()
 		if monitor.Enabled {
-			monitor.SegmentFullyTranscoded(nonce, seg.SeqNo, common.ProfilesNames(sess.Profiles), len(segHashes) == len(res.Segments))
+			monitor.SegmentFullyTranscoded(nonce, seg.SeqNo, common.ProfilesNames(sess.Profiles), errCode)
 		}
 
 		ticketParams := sess.OrchestratorInfo.GetTicketParams()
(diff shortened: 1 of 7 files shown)


j0sh:
Looks OK. Reporting "100% success" feels kind of strange if there isn't actual work being done. Is there some sort of intermediate state such as "inactive" that we could report? I'm concerned that this may distort the metrics over time, eg

T1. Success (active)
T2. Success (inactive)
T3. Fail (active)

Here the metrics would indicate a 66% success rate from T1-T3 when it's actually 50%.

darkdarkdragon:
This is needed because Prometheus client library doesn't have ability to remove metric from being reported. 

> Here the metrics would indicate a 66% success rate from T1-T3 when it's actually 50%.

With current algo this situation isn't possible - once stream ended, it gets removed from success rate calculation immediately. But this means we could loose failures of up to four last segments (if they not time fail immediately, but, say, fail because of timeout).
Do you think I should improve algo to account for this possibility?

j0sh:
> Do you think I should improve algo to account for this possibility?

Maybe - perhaps I misunderstand but how does the success rate metric / algorithm work right now?

darkdarkdragon:
for every stream there is ring buffer (`segmentsAverager`) which keeps last 30 segments emerged. For each segment I'm recording `seqNo`, time in which it appeared and emerged/transcoded flags. Overall success rate is average of success rate of every stream. Stream's success rate is number of successfully transcoded segments divided by number of emerged segments (excluding segments that was emerged less than 8.5s ago and hasn't received success or failure event) - so success rate for stream is essentially is moving average with 60s window, and timeout errors gets delayed by 8.5 seconds.
But after stream ends, whole `segmentsAverager` structure with list of segments gets removed, so any possible timeout errors for last 8.5s gets lost.
After second thought, I'll change algorithm to wait for last segments completion or timeout, will do this with another PR.

darkdarkdragon:
@j0sh Now if last couple of segments timeout, it will be reflected in success rate.
Also I've added unit tests.

server/broadcast.go
```diff
@@ -236,7 +238,7 @@ func processSegment(cxn *rtmpConnection, seg *stream.HLSSegment) {
 		}
 		cond.L.Unlock()
 		if monitor.Enabled {
-			monitor.SegmentFullyTranscoded(nonce, seg.SeqNo, common.ProfilesNames(sess.Profiles), len(segHashes) == len(res.Segments))
+			monitor.SegmentFullyTranscoded(nonce, seg.SeqNo, common.ProfilesNames(sess.Profiles), len(segHashes) == len(res.Segments), errCode)
```
j0sh: Just noting that we'll exit early if there's an error with the download (in order to retry), so I hope we aren't using the SegmentFullyTranscoded call to wrap up any stateful tracking:  https://github.com/livepeer/go-livepeer/commit/2dcf826485969be9415cc0b93b2237d867ecc751

The params here are getting sort of long, would just passing in the error suffice to determine whether `len(segHashes) == len(res.Segments)` ? Does the `allSuccess` parameter convey any additional data?

Additionally, we might want to incorporate any errors from a failed sig check.

darkdarkdragon: > I hope we aren't using the SegmentFullyTranscoded call to wrap up any stateful tracking

We are. Adjust metrics to retries is my next task.

> would just passing in the error suffice to determine whether len(segHashes) == len(res.Segments)

Changed it to use `gotErr` flag.

> Additionally, we might want to incorporate any errors from a failed sig check.

Will do this later, when we will start doing with this information something else than just printing it to log.

j0sh (approved):
LGTM! 🚢

Issue #817, opened 2019-04-16
Step 1: Track remote transcoder capacity


ericxtang:
1. Add an option for Ts to specify their capacity via CLI, through maxSessions. As a rule of thumb, each T will be able to handle one segment per core (though capacity advertised is up to each T to determine).
* add `maxSessions` to `server.RunTranscoder` (in livepeer.go), `runTranscoder` (ot_rpc.go), `RegisterTranscoder`, `orch.node.serveTranscoder` (in orchestrator.go), `orch.ServeTranscoder`, `serveTranscoder`, `NewRemoteTranscoder`
* add `MaxSessions` to `TranscoderClient.RegisterTranscoder` in `lp_rpc.proto`

2. Build the RemoteTranscoder struct, which will keep track of remote transcoder capacity, and implements the existing Transcoder interface:
```
type RT struct {
	capacity int
}

func NewRemoteTranscoder(capacity int) *RT {
  return &RT{ capacity: capacity }
}

func (r *RT) Transcode() {
}

func (r *RT) Capacity() int {
	return r.capacity
}
```

j0sh:
Fixed by https://github.com/livepeer/go-livepeer/pull/833/commits/378e9f4ee5b7d046fb97562e9c92023c10ea834b

PR #833, opened 2019-04-17
Multi-T Allocation and Management (steps broken down by commit)

angyangie:
**What does this pull request do? Explain your changes**

Step 1: 91a992f958cfe39aaa38a494ca24bd309c02cd74
Add an option for Ts to specify their capacity via CLI, through maxSessions. As a rule of thumb, each T will be able to handle one segment per core (though capacity advertised is up to each T to determine). MaxSessions is stored in the RemoteTranscoder struct.
- add maxSessions to server.RunTranscoder (in livepeer.go), runTranscoder (ot_rpc.go), RegisterTranscoder, orch.node.serveTranscoder (in orchestrator.go), orch.ServeTranscoder, serveTranscoder, NewRemoteTranscoder
- add MaxSessions to TranscoderClient.RegisterTranscoder in lp_rpc.proto

Step 2: 3260895d09960294f5b89f8901d28e99f323b26e
Implement the RemoteTranscoderManager struct and its corresponding Register (not to be confused with the lphttp.Register method), Unregister and Transcode methods.

Step 3: 3260895d09960294f5b89f8901d28e99f323b26e & 30b998b988d2855e1b4a96ecdebb37bed205339f
- Add RemoteTranscoderManager to LivepeerNode (Transcoder interface is already a part of it).
- Add n.RTmanager, n.Transcoder, RTManager.Register and RTManager.Unregister to livepeer.go

Step 4:  5280e9e66de0a1c9d453b60172681f94d602b247
- T can currently only handle one job. Make the relevant parts in ot_rpc.go async

https://github.com/livepeer/go-livepeer/issues/817

**Checklist:**


server/ot_rpc.go
```diff
@@ -83,57 +83,57 @@ func runTranscoder(n *core.LivepeerNode, orchAddr string, capacity int) error {
...
 		}
-		profiles, err := common.TxDataToVideoProfile(hex.EncodeToString(notify.Profiles))
-		if err != nil {
-			glog.Info("Unable to deserialize profiles ", err)
-		}
+		go runTranscode(n, orchAddr, httpc, notify)
+	}
+}
+
+func runTranscode(n *core.LivepeerNode, orchAddr string, httpc *http.Client, notify *net.NotifySegment) {
```
j0sh, line 91: For code splits that result in a large amount of re-indentation such as this, I often find it helpful to insert a dummy set of braces to maintain the original indentation. This allows reviewers to clearly see what has changed in the commit. Afterwards, a cosmetic patch can be applied that only re-idents code without functional changes.

So this would become something like:

```
func runTranscode(n *core.LivepeerNode, orchAddr string, httpc *http.Client, notify *net.NotifySegment) {
    { // dummy indentation
        ... original code here ...
    }
}
```

angyangie: Interesting! That certainly does make it easier for the reviewer. Thanks for the tip!

core/orchestrator.go
```diff
@@ -517,49 +514,58 @@ type RemoteTranscoderManager struct {
...
 	defer rtm.RTmutex.Unlock()
+
 	rtm.liveTranscoders[transcoder.stream] = transcoder
 
 	for i := 0; i < transcoder.Capacity(); i++ {
-		rtm.AppendTranscoder(transcoder)
-	}
-}
-
-func (rtm *RemoteTranscoderManager) AppendTranscoder(transcoder *RemoteTranscoder) {
```
j0sh: These changes to the RemoteTranscoderManager can probably be fixed up into https://github.com/livepeer/go-livepeer/pull/833/commits/207e18a84a0fc4603ebab83bcb477a469eddcad8 ?

core/orchestrator.go
```diff
@@ -426,18 +426,15 @@ func (n *LivepeerNode) transcodeSegmentLoop(md *SegTranscodingMetadata, segChan
 }
 
 func (n *LivepeerNode) serveTranscoder(stream net.Transcoder_RegisterTranscoderServer, capacity int) {
-	transcoder := NewRemoteTranscoder(n, stream, capacity)
-
 	n.tcoderMutex.Lock()
```
j0sh: Do we still need this particular mutex?

core/orchestrator.go
```diff
@@ -517,49 +514,58 @@ type RemoteTranscoderManager struct {
...
+	checkTranscoders := func(rtm *RemoteTranscoderManager) bool {
+		return len(rtm.remoteTranscoders) > 0
 	}
 
-	var remoteTranscoders []*RemoteTranscoder
-	for _, v := range rtm.remoteTranscoders {
-		if v.stream == t.stream {
-			continue
+	for checkTranscoders(rtm) {
+		currentTranscoder, remoteTranscoders := rtm.remoteTranscoders[0], rtm.remoteTranscoders[1:]
```
j0sh: Should we take the same "stack based" approach as multi-O to reduce some of the load here, especially with appends? That is, select from the end rather than the beginning (and append back on the end) so the backing store doesn't need to be continuously reallocated.

j0sh (changes requested):

core/orchestrator.go
angyangie: Sure we can do that. But is it a concern if some of the transcoders end up never getting any work, though they are comparably good Ts? Sure, if one T is getting all the work, that T probably has a higher capacity, can do transcoding faster, and in a free market, should get as many segments as it can handle ... and sure, T's could move onto working with a different O if they're not getting too much work. But, is that going to promote large players from taking all work, and smaller players to flounder? I imagine a lot of O's might run their own transcoders, so it wouldn't matter for them.

core/orchestrator.go
angyangie: We do not!

core/orchestrator.go
j0sh: At this point, the economics / incentives of Ts isn't really a consideration here.

There is something to be said about spreading the load evenly among Ts - currently, the way we construct lists on registration leads to Ts being "clumped" together. Whether we use a stack or a queue wouldn't change this clumping behavior. There are ways we can optimize this, but I don't think it's something we need to address immediately.

angyangie: Got it, we can address later then!

core/orchestrator.go
```diff
@@ -485,14 +486,89 @@ func (rt *RemoteTranscoder) Transcode(fname string, profiles []ffmpeg.VideoProfi
...
+	rtm.liveTranscoders[transcoder.stream] = transcoder
+
+	for i := 0; i < transcoder.Capacity(); i++ {
+		rtm.remoteTranscoders = append(rtm.remoteTranscoders, transcoder)
+	}
+}
+
+func (rtm *RemoteTranscoderManager) Unregister(t *RemoteTranscoder) {
+	rtm.RTmutex.Lock()
+	delete(rtm.liveTranscoders, t.stream)
```
darkdarkdragon: @j0sh @angyangie 
Should we delete from `remoteTranscoders` too? If we not, then O will start hitting dead transcoder again, even if it known that it is dead. Or maybe we could add flag `dead` to `RemoteTranscoder`, so O will not try It again if it was already unregistered?

j0sh:
Reworked the commit structure a bit to more cleanly separate the TranscoderManager implementation (https://github.com/livepeer/go-livepeer/pull/833/commits/4194ef244cfa28006221e79335b3727d9fbbc298) from the integration of that manager (https://github.com/livepeer/go-livepeer/pull/833/commits/82493a301dcd61abbb05813e547e568777f1bb8a).

Tested a bit and it's working nicely. Still need to review the unit tests and do more testing but things are looking good!

core/orchestrator.go
j0sh: In `SelectTranscoder` any transcoders not in the `liveTranscoders` table are permanently removed from `remoteTranscoders`, so we wouldn't actually hit any dead transcoders. That bit of logic was borrowed from multi-O where it was important to release the lock quickly. In this case, it is less important to release the lock quickly since unregisters are probably not in the critical path. `SelectTranscoder` is still in the critical path.

However I'm not sure what would be better: to hold the lock longer here in a (probably) non-critical path, and iterate through the entire list of `remoteTranscoders`, or do shorter iterations as-needed during `SelectTranscoder` ?

core/orchestrator.go
```diff
@@ -485,12 +482,86 @@ func (rt *RemoteTranscoder) Transcode(fname string, profiles []ffmpeg.VideoProfi
...
+	defer rtm.RTmutex.Unlock()
```
darkdarkdragon: removing `defer` here will spare couple of CPU cycles

core/orchestrator.go
darkdarkdragon: Yeah, my mistake - somehow looking into logs I thought it hits dead transcoders.

>However I'm not sure what would be better

I think it is best how it is now - now, for dead transcoder, it will quickly cut out that transcoder from end of the list. If we will move it into `Unregister` - then we will be actually copying parts of the list, and it will probably will be slower.

core/orchestrator.go
angyangie: Good catch! I'll remove that

j0sh:
Generally LGTM. Some very small changes that have been fixed up (diff below):

* Returned a pointer value from NewRemoteTranscoderManager, because all method receivers are pointers.

* Made `selectTranscoder` a private method, since it's not really meant to be used outside the transcoder manager.

Our error handling was making me nervous (turns out there was a bug there), so I did a separate PR to fix that and a few other things: https://github.com/livepeer/go-livepeer/pull/854 

Diff of changes to this branch:
```diff
diff --git a/cmd/livepeer/livepeer.go b/cmd/livepeer/livepeer.go
index 833564e7..4c3257f3 100644
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -203,7 +203,7 @@ func main() {
 		n.NodeType = core.OrchestratorNode
 		if !*transcoder {
 			n.TranscoderManager = core.NewRemoteTranscoderManager()
-			n.Transcoder = &n.TranscoderManager
+			n.Transcoder = n.TranscoderManager
 		}
 	} else if *transcoder {
 		n.NodeType = core.TranscoderNode
diff --git a/core/livepeernode.go b/core/livepeernode.go
index 465c4fd6..27ae5b11 100644
--- a/core/livepeernode.go
+++ b/core/livepeernode.go
@@ -63,7 +63,7 @@ type LivepeerNode struct {
 	Ipfs              ipfs.IpfsApi
 	OrchSecret        string
 	Transcoder        Transcoder
-	TranscoderManager RemoteTranscoderManager
+	TranscoderManager *RemoteTranscoderManager
 
 	// Broadcaster public fields
 	Sender pm.Sender
diff --git a/core/orch_test.go b/core/orch_test.go
index b8d8b0fb..da8bdbdc 100644
--- a/core/orch_test.go
+++ b/core/orch_test.go
@@ -196,8 +196,8 @@ func TestSelectTranscoder(t *testing.T) {
 	assert.NotNil(n.TranscoderManager.liveTranscoders[strm2])
 	assert.Len(n.TranscoderManager.remoteTranscoders, 9)
 
-	// assert transcoder is returned from SelectTranscoder and removed from list
-	currentTranscoder := n.TranscoderManager.SelectTranscoder()
+	// assert transcoder is returned from selectTranscoder and removed from list
+	currentTranscoder := n.TranscoderManager.selectTranscoder()
 	assert.Equal(currentTranscoder, transcoder2)
 	assert.NotNil(n.TranscoderManager.liveTranscoders[strm])
 	assert.Len(n.TranscoderManager.remoteTranscoders, 8)
@@ -207,8 +207,8 @@ func TestSelectTranscoder(t *testing.T) {
 	assert.NotNil(n.TranscoderManager.liveTranscoders[strm2])
 	assert.Nil(n.TranscoderManager.liveTranscoders[strm])
 
-	// assert transcoder is returned from SelectTranscoder and removed from list
-	currentTranscoder = n.TranscoderManager.SelectTranscoder()
+	// assert transcoder is returned from selectTranscoder and removed from list
+	currentTranscoder = n.TranscoderManager.selectTranscoder()
 	assert.Equal(currentTranscoder, transcoder2)
 	assert.NotNil(n.TranscoderManager.liveTranscoders[strm2])
 	assert.Len(n.TranscoderManager.remoteTranscoders, 7)
diff --git a/core/orchestrator.go b/core/orchestrator.go
index 14b99323..0b4038e7 100644
--- a/core/orchestrator.go
+++ b/core/orchestrator.go
@@ -491,8 +491,8 @@ func NewRemoteTranscoder(n *LivepeerNode, stream net.Transcoder_RegisterTranscod
 	}
 }
 
-func NewRemoteTranscoderManager() RemoteTranscoderManager {
-	return RemoteTranscoderManager{
+func NewRemoteTranscoderManager() *RemoteTranscoderManager {
+	return &RemoteTranscoderManager{
 		remoteTranscoders: []*RemoteTranscoder{},
 		liveTranscoders:   map[net.Transcoder_RegisterTranscoderServer]*RemoteTranscoder{},
 		RTmutex:           &sync.Mutex{},
@@ -521,7 +521,7 @@ func (rtm *RemoteTranscoderManager) Unregister(t *RemoteTranscoder) {
 	rtm.RTmutex.Unlock()
 }
 
-func (rtm *RemoteTranscoderManager) SelectTranscoder() *RemoteTranscoder {
+func (rtm *RemoteTranscoderManager) selectTranscoder() *RemoteTranscoder {
 	rtm.RTmutex.Lock()
 	defer rtm.RTmutex.Unlock()
 
@@ -552,7 +552,7 @@ func (rtm *RemoteTranscoderManager) completeTranscoders(trans *RemoteTranscoder)
 }
 
 func (rtm *RemoteTranscoderManager) Transcode(fname string, profiles []ffmpeg.VideoProfile) ([][]byte, error) {
-	currentTranscoder := rtm.SelectTranscoder()
+	currentTranscoder := rtm.selectTranscoder()
 	if currentTranscoder == nil {
 		return nil, errors.New("No transcoders available")
 	}
```

j0sh:
Rebased against master - although I'm not sure what's up with AppVeyor making CI red. But tests pass otherwise. We can merge this early next week once https://github.com/livepeer/go-livepeer/pull/854 is in.

🏁 Finish line in sight! 🏁

core/orchestrator.go
```diff
@@ -316,37 +316,41 @@ func (n *LivepeerNode) transcodeSeg(config transcodeConfig, seg *stream.HLSSegme
...
 	// Prepare the result object
 	var tr TranscodeResult
 	segHashes := make([][]byte, len(tData))
 
-	for i, _ := range md.Profiles {
-		if tData[i] == nil {
-			glog.Errorf("Cannot find transcoded segment for %v", seg.SeqNo)
-			continue
+	for i := range md.Profiles {
+		if tData[i] == nil || len(tData[i]) < 25 {
```
angyangie, line 339: Where does the 25 come from here?

core/orchestrator.go
```diff
@@ -474,21 +478,27 @@ func (rt *RemoteTranscoder) Transcode(fname string, profiles []ffmpeg.VideoProfi
...
 func NewRemoteTranscoder(n *LivepeerNode, stream net.Transcoder_RegisterTranscoderServer, capacity int) *RemoteTranscoder {
-	return &RemoteTranscoder{
+	t := &RemoteTranscoder{
 		node:     n,
 		stream:   stream,
 		eof:      make(chan struct{}, 1),
 		capacity: capacity,
+		addr:     "unknown",
+	}
+	if p, ok := peer.FromContext(stream.Context()); ok {
```
angyangie: Didn't know about `peer.FromContext()` . Cool!

core/orchestrator.go
darkdarkdragon: Size of empty frame used in unit tests. If you increase this this constant here, you'll see tests fails.

core/orchestrator.go
darkdarkdragon: I dug it up from StackOverflow and documentation :)

j0sh (approved):
LGTM! 🚢 🚢 🚢

PR #844, opened 2019-05-01
Success rate fix

darkdarkdragon:
Fixes success rate calculation to account for segment retries.
Better logged messages.
Change everywhere to use Golang 1.12

- 
- 
- 

Manually

**Checklist:**


server/broadcast.go
```diff
@@ -396,23 +402,23 @@ func transcodeSegment(cxn *rtmpConnection, seg *stream.HLSSegment, name string)
...
 		}
 		cond.L.Unlock()
-		if dlErr != nil {
-			return dlErr
-		}
 		if monitor.Enabled {
 			monitor.SegmentFullyTranscoded(nonce, seg.SeqNo, common.ProfilesNames(sess.Profiles), errCode)
 		}
+		if dlErr != nil {
+			return dlErr
```
j0sh, line 404: If download fails and retry succeeds, does it mean we get SegmentFullyTranscoded twice which means the success rate could be >100% ?

PM seems to have the same issue. Perhaps rather than moving this chunk of code, we should be moving the `SegmentFullyTranscoded` check to after PM and before we exit.

j0sh (changes requested):
Can you describe how adding a `permanent` field helps account for retries? I'm not sure if I understand how that affects things.

server/broadcast.go
darkdarkdragon: Fixed

darkdarkdragon:
`LogSegmentUploadFailed` gets called from places before `transcodeSegment` - and that means that segment will not be retried, so error is permanent and should be counted into success rate. Also it gets called from inside `transcodeSegment` - and then segment will be retried, so this shouldn't be counted towards success rate. Same thing with `LogSegmentTranscodeFailed`.

j0sh:
Can the success rate be simply something like `segmentsFullyTranscoded / segmentsAppeared` rather than doing stateful tracking of each retry?

darkdarkdragon:
No, because at the moment you add segment to `segmentsAppeared` counter and to the moment it is added into `segmentsFullyTranscoded` - your success rate is below 100%. Hence all the stateful tracking - we should increase `segmentsAppeared` only when we have successfully transcoded segment or we're certain it is failed and will not be tried again.
Plus, just counting segments means that metric will be absolute from beginning of stream - this is useful for our tests, but not for production - that's why I'm calculating success rate as moving average for one minute.

j0sh:
> No, because at the moment you add segment to segmentsAppeared counter and to the moment it is added into segmentsFullyTranscoded - your success rate is below 100%

This seems OK since it should be only off by one or two while a stream is running. That seems acceptable to simplify the implementation. Right now the implementation is complicated; I'm concerned this may be prone to introducing errors in our statistics.

j0sh:
Another option is to have explicit start / end events for each stateful block we're trying to track, and doing that state tracking within the metrics system itself. That would avoid the need for a lot of bookkeeping outside the metrics system, eg tracking the `try` counter.

darkdarkdragon:
> This seems OK since it should be only off by one or two while a stream is running.

It is not - Prometheus do momentary scrapes of metrics from client - so it will be constantly showing inconsistencies between `emerged` and `transcoded` counters, which is not accurate. And it can be off not by one or two, but for number of orchestrators used.

> Another option is to have explicit start / end events for each stateful block we're trying to track, and doing that state tracking within the metrics system itself.

We're doing it right now. What else state tracking beside `try` counter do you see outside of metrics system? And for that one - it is same argument as for having external server reconstructing what's going on in app by looking into events it emits vs. tracking some state inside app itself, where it is already have all that state. We surely can move `try` counter inside metrics code, but that will complicate metrics code even more.

>  Right now the implementation is complicated; I'm concerned 

It does bothers me too, but I really don't see simpler way. One my mistake I see now - I should've been thrown out old metrics code and have started from defining interfaces to new metrics system. Probably I'll be able to make current implementation slightly clearer when will be throwing out old metrics code/refactoring what we have.

j0sh:
> And it can be off not by one or two, but for number of orchestrators used.

How does the number of orchestrators affect this? Approximately how much does the difference grow as the number of orchestrators grows? Does it eventually converge as the streams end?

If we know approximately how much drift to expect then maybe we can factor that in to our calculations? If we're just looking at the total numbers (as most of the current testing is doing) then is accommodating that drift still as important?

> We surely can move try counter inside metrics code, but that will complicate metrics code even more.

Given the choice of keeping metrics counters outside the metrics code (eg, `try`), I'd prefer to keep it self-contained within the metrics code. The locality should help in overall understanding and reducing code sprawl; right now, the `try` counter is being generated and passed around only for a single call to metrics.

darkdarkdragon:
> How does the number of orchestrators affect this? 

Theoretically there can be `maxInflight*2` segments transcoded simultaneously. But, more importantly, this should be multiplied to number of streams. Yes, it will converge at the stream end, but what's the point of having metrics that will be accurate eventually?

> If we're just looking at the total numbers

This metrics is meant to be used for monitoring, not for testing. Is gives answer on the question 'what success rate is for the last minute', not the 'what success rate of complete stream, after it ended'.

> right now, the try counter is being generated and passed around only for a single call to metrics.

Fixed that.

j0sh:
>  this should be multiplied to number of streams

Wouldn't the drift percentage should stay steady as the number of streams increases? And shrink over time as more segments are processed? My guess is the total drift will be negligible after a few hours?

> what's the point of having metrics that will be accurate eventually?

To do a fair comparison during benchmarking, we really need to be running uniform and reproducible tests, which means using the same inputs each time, producing the same number of segments, etc. "In progress" metrics might be helpful later on [1], but right now the final results are more important to us.

> This metrics is meant to be used for monitoring, not for testing.

We have been using these numbers for testing and benchmarking. If you believe this is fundamentally flawed then we need to come up with another way to get the numbers we need.

[1] And I think it's totally OK to have some drift between the numbers as long as we understand where the drift is coming from. This seems like a good tradeoff to for a simpler implementation and more confidence in our numbers.

darkdarkdragon:
> We have been using these numbers for testing and benchmarking. 

I've added two new metrics - `segment_source_emerged_unprocessed_total` (number of segments emerged from segmenter) and `segment_transcoded_unprocessed_total`) - they are just raw counters without any conditions around them. They can be used for our benchmarking purposes. Once merged, I'll made dashboard to show these numbers.

darkdarkdragon:
@j0sh I have also remove old metrics code.

core/orchestrator.go
```diff
@@ -329,10 +326,9 @@ func (n *LivepeerNode) transcodeSeg(config transcodeConfig, seg *stream.HLSSegme
 	}
 	took := time.Since(start)
 	tProfileData := make(map[ffmpeg.VideoProfile][]byte, 0)
-	glog.V(common.DEBUG).Infof("Transcoding of segment %v took %v", seg.SeqNo, took)
+	glog.V(common.DEBUG).Infof("Transcoding of segment manifestID=%s seqNo=%d took %v", string(md.ManifestID), seg.SeqNo, took)
 	if isLocal && monitor.Enabled {
-		monitor.LogSegmentTranscodeEnded(seg.SeqNo, string(md.ManifestID), took,
-			common.ProfilesNames(md.Profiles))
+		monitor.SegmentTranscoded(0, seg.SeqNo, took, 0, common.ProfilesNames(md.Profiles))
```
j0sh: Would a separate function be more appropriate here and in `transcoder.go`? Also just want to double check that this won't lead to double counting successes if the B or any attached Ts also report success

j0sh (changes requested):
Just one more note but otherwise this is looking much better! Really like the event monitoring code cleanup.

core/orchestrator.go
darkdarkdragon: Removed one unused argument and added some explanation of calling logic 540ff3d586ae6f5aa706be8c4ab2c24c7f0176eb 
It will not lead to double counting, because each metric gets tagged with node name and node type, and in query we can aggregate by any tag.
If O is transcoding by itself, only call in `orchestrator.go` will report success, code in `transcoder.go` only reports when it sees real http urls, not local filenames (and that means it works as separate transcoder).

darkdarkdragon:
@j0sh I've rebase this on master

core/orchestrator.go
```diff
@@ -245,10 +245,10 @@ func (n *LivepeerNode) sendToTranscodeLoop(md *SegTranscodingMetadata, seg *stre
 	segChanData := &SegChanData{seg: seg, md: md, res: make(chan *TranscodeResult, 1)}
 	select {
 	case ch <- segChanData:
-		glog.V(common.DEBUG).Infof("Submitted segment to transcode loop manifes=%s seqNo=%d", string(md.ManifestID), md.Seq)
+		glog.V(common.DEBUG).Infof("Submitted segment to transcode loop manifesID=%s seqNo=%d", md.ManifestID, md.Seq)
```
j0sh: typo - manifestID (missing the 't'

core/orchestrator.go
```diff
@@ -245,10 +245,10 @@ func (n *LivepeerNode) sendToTranscodeLoop(md *SegTranscodingMetadata, seg *stre
...
 	default:
 		// sending segChan should not block; if it does, the channel is busy
-		glog.Error("Transcoder was busy with a previous segment!")
+		glog.Errorf("Transcoder was busy with a previous segment manifesID=%s seqNo=%d", md.ManifestID, md.Seq)
```
j0sh: same typo

core/transcoder.go
```diff
@@ -31,14 +31,15 @@ func (lt *LocalTranscoder) Transcode(fname string, profiles []ffmpeg.VideoProfil
...
 	start := time.Now()
-	if monitor.Enabled && parseErr == nil {
-		monitor.LogSegmentTranscodeStarting(seqNo, mid)
-	}
 	data, err := tr.Transcode(fname)
 	if monitor.Enabled && parseErr == nil {
-		monitor.LogSegmentTranscodeEnded(seqNo, mid, time.Since(start), common.ProfilesNames(profiles))
+		// This will run only when fname is actual URL and contains seqNo in it.
+		// When orchestrator works as transcoder, `fname` will be relative path to file in local
+		// filesystem and will not contain seqNo in it. For that case `SegmentTranscoded` will
```
j0sh, line 40: Are these assumptions robust / durable? If the fname structure ever changes, then this metrics code needs to change? Eg, https://github.com/livepeer/go-livepeer/issues/846

core/transcoder.go
```diff
@@ -31,14 +31,15 @@ func (lt *LocalTranscoder) Transcode(fname string, profiles []ffmpeg.VideoProfil
...
+		// be called in orchestrator.go
+		monitor.SegmentTranscoded(0, seqNo, time.Since(start), common.ProfilesNames(profiles))
```
j0sh, line 42: It seems like we're tracking three events under this same metric:

* Orchestrator transcode complete
* Transcoder transcoded
* Broadcaster transcode complete

Distinguishing between those seems non-obvious. Would it make sense to track these separately?

cmd/devtool/devtool.go
```diff
@@ -69,11 +73,25 @@ func main() {
 	}
 	if !goodToGo {
 		fmt.Println(`
-    Usage: go run cmd/devtool/devtool.go setup broadcaster|transcoder
+    Usage: go run cmd/devtool/devtool.go setup broadcaster|transcoder [nodeIndex]
         It will create initilize eth account (on private testnet) to be used for broadcaster or transcoder
-        and will create shell script (run_broadcaster_ETHACC.sh or run_transcoder_ETHACC.sh) to run it.`)
+				and will create shell script (run_broadcaster_ETHACC.sh or run_transcoder_ETHACC.sh) to run it.
+				Node index indicates how much to offset node's port. Orchestrator node's index by default is 1.`)
```
j0sh: Not quite clear on what an "node index" is supposed to do, maybe an example would help or some deeper documentation of each of devtool's options elsewhere (eg, in `/docs/devtool.md`).

j0sh (changes requested):

core/orchestrator.go
darkdarkdragon: fixed

core/orchestrator.go
darkdarkdragon: fixed

core/transcoder.go
darkdarkdragon: Not of course, and will need to be changed in case of changes like in #846, but it's best what can be done now - because for local files here we only have random name and no other information.

core/transcoder.go
darkdarkdragon: These are same events, but happening on different nodes. It is matter of query to distinguish between them:

`livepeer_segment_transcoded_total{node_type="orch"}`
or `livepeer_segment_transcoded_total{node_type="bctr"}`

Like [this](http://34.83.171.222:9090/graph?g0.range_input=1h&g0.expr=livepeer_segment_transcoded_total%7Bnode_type%3D%22orch%22%7D&g0.tab=1)

cmd/devtool/devtool.go
darkdarkdragon: Added example

j0sh (approved):

Issue #845, opened 2019-05-02
Remote transcoder can't transcode more than one stream


darkdarkdragon:
Remote transcoder feeds to `LPMS.Transcode` function URL of source segment (pointing to orchestrator).
This URL looks like this: `https://orchestrator:8936/stream/manifestid/1.ts`
LPMS, calling ffmpeg lib, sets output destination for transcoded data as this:
`fmt.Sprintf("out%v%v", i, filepath.Base(url))`, where `I` is transcoding profile number.
This means that requests to transcoded segments from different streams but with same seqNos will overwrite each other (and transcoding for one of the streams will fail).

Possible solutions:
1. Randomize output name in LPMS itself.
2. In transcoder, for every segment create random dir and pass it too LPMS.

darkdarkdragon:
Suggested [fix](https://github.com/livepeer/lpms/pull/116)

j0sh:
Good catch!

The transcoding API (which is admittedly not great) is getting an overhaul with the GPU support. I've already incorporated the [first suggested fix](https://github.com/livepeer/go-livepeer/compare/master...ja/nvidia#diff-418dc657e12875665c93dffe96a0e464R73) into the goclient's nvidia integration.

Once nvidia integration lands, we can retrofit the software transcoding to use the new API. If the fix for this is more urgent, then the second option would be less invasive - we can put a sub-directory in here (and `defer os.RemoveAll(subdir)`). Not pretty but should do the trick for now.

https://github.com/livepeer/go-livepeer/blob/960084b4864927a61f3dc77fb6d6e0cda29c6e63/core/transcoder.go#L23-L24

darkdarkdragon:
[Here](https://github.com/livepeer/go-livepeer/pull/848/commits/9cdddf8b28d58a5ec7731f319fb70c3cdd0bc880) implementation.

j0sh:
Fixed by https://github.com/livepeer/go-livepeer/pull/848/commits/76a5dd8449094d45b26232ba3b9b02f9b4976512

PR #848, opened 2019-05-02
Improvements

darkdarkdragon:
Some improvements

- Do not start to listen for remote transcoders if orchestrator serves as transcoder itself
- More logging
- Added fields to `/status` endpoint

Manually

**Checklist:**

### DIFF
--- a/core/orchestrator.go
+++ b/core/orchestrator.go
@@ -13,6 +13,8 @@ import (
 	"sync"
 	"time"
 
+	"google.golang.org/grpc/peer"
+
 	ethcommon "github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/crypto"
 	"github.com/golang/glog"
@@ -449,6 +451,7 @@ type RemoteTranscoder struct {
 	node     *LivepeerNode
 	stream   net.Transcoder_RegisterTranscoderServer
 	eof      chan struct{}
+	addr     string
 	capacity int
 }
 
@@ -488,17 +491,17 @@ func (rt *RemoteTranscoder) Transcode(fname string, profiles []ffmpeg.VideoProfi
 }
 
 func NewRemoteTranscoder(n *LivepeerNode, stream net.Transcoder_RegisterTranscoderServer, capacity int) *RemoteTranscoder {
-	t := &RemoteTranscoder{
+	addr := "unknown"
+	if p, ok := peer.FromContext(stream.Context()); ok {
+		addr = p.Addr.String()
+	}
+	return &RemoteTranscoder{
 		node:     n,
 		stream:   stream,
 		eof:      make(chan struct{}, 1),
 		capacity: capacity,
-		addr:     "unknown",
+		addr:     addr,
 	}
-	if p, ok := peer.FromContext(stream.Context()); ok {
-		t.addr = p.Addr.String()
-	}
-	return t
 }
 
 func NewRemoteTranscoderManager() *RemoteTranscoderManager {
@@ -515,6 +518,22 @@ type RemoteTranscoderManager struct {
 	RTmutex           *sync.Mutex
 }
 
+func (rtm *RemoteTranscoderManager) RegisteredTranscodersCount() int {
+	rtm.RTmutex.Lock()
+	defer rtm.RTmutex.Unlock()
+	return len(rtm.liveTranscoders)
+}
+
+func (rtm *RemoteTranscoderManager) RegisteredTranscodersInfo() []net.RemoteTranscoderInfo {
+	rtm.RTmutex.Lock()
+	res := make([]net.RemoteTranscoderInfo, 0, len(rtm.liveTranscoders))
+	for _, transcoder := range rtm.liveTranscoders {
+		res = append(res, net.RemoteTranscoderInfo{Address: transcoder.addr, Capacity: transcoder.capacity})
+	}
+	rtm.RTmutex.Unlock()
+	return res
+}
+
 func (rtm *RemoteTranscoderManager) Register(transcoder *RemoteTranscoder) {
 	rtm.RTmutex.Lock()
 	defer rtm.RTmutex.Unlock()
@@ -527,7 +546,6 @@ func (rtm *RemoteTranscoderManager) Register(transcoder *RemoteTranscoder) {
 
 func (rtm *RemoteTranscoderManager) Unregister(t *RemoteTranscoder) {
 	rtm.RTmutex.Lock()
-	// Probably should remove from remoteTranscoders too?
 	delete(rtm.liveTranscoders, t.stream)
 	rtm.RTmutex.Unlock()
 }
--- a/core/orchestrator.go
+++ b/core/orchestrator.go
@@ -517,6 +517,7 @@ func (rtm *RemoteTranscoderManager) Register(transcoder *RemoteTranscoder) {
 
 func (rtm *RemoteTranscoderManager) Unregister(t *RemoteTranscoder) {
 	rtm.RTmutex.Lock()
+	// Probably should remove from remoteTranscoders too?
 	delete(rtm.liveTranscoders, t.stream)
 	rtm.RTmutex.Unlock()
 }
--- a/core/orchestrator.go
+++ b/core/orchestrator.go
@@ -234,7 +234,7 @@ func (n *LivepeerNode) getSegmentChan(md *SegTranscodingMetadata) (SegmentChan, 
 }
 
 func (n *LivepeerNode) sendToTranscodeLoop(md *SegTranscodingMetadata, seg *stream.HLSSegment) (*TranscodeResult, error) {
-	glog.V(common.DEBUG).Infof("Starting to transcode segment %v", md.Seq)
+	glog.V(common.DEBUG).Infof("Starting to transcode segment manifest=%s seqNo=%d", string(md.ManifestID), md.Seq)
 	ch, err := n.getSegmentChan(md)
 	if err != nil {
 		glog.Error("Could not find segment chan ", err)
@@ -243,7 +243,7 @@ func (n *LivepeerNode) sendToTranscodeLoop(md *SegTranscodingMetadata, seg *stre
 	segChanData := &SegChanData{seg: seg, md: md, res: make(chan *TranscodeResult, 1)}
 	select {
 	case ch <- segChanData:
-		glog.V(common.DEBUG).Info("Submitted segment to transcode loop ", md.Seq)
+		glog.V(common.DEBUG).Infof("Submitted segment to transcode loop manifes=%s seqNo=%d", string(md.ManifestID), md.Seq)
 	default:
 		// sending segChan should not block; if it does, the channel is busy
 		glog.Error("Transcoder was busy with a previous segment!")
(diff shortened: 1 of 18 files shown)


server/mediaserver.go
```diff
@@ -520,7 +520,34 @@ func (s *LivepeerServer) GetNodeStatus() *net.NodeStatus {
...
+		OrchestratorPool:      []net.StatusOrchestratorInfo{},
+		RegisteredTranscoders: []net.RemoteTranscoderInfo{},
+		LocalTranscoding:      s.LivepeerNode.TranscoderManager == nil,
+	}
+	if s.LivepeerNode.TranscoderManager != nil {
+		res.RegisteredTranscodersNumber = s.LivepeerNode.TranscoderManager.RegisteredTranscodersCount()
+		res.RegisteredTranscoders = s.LivepeerNode.TranscoderManager.RegisteredTranscodersInfo()
+	}
+	if s.LivepeerNode.OrchestratorPool != nil {
+		orchs, err := s.LivepeerNode.OrchestratorPool.GetOrchestrators(s.LivepeerNode.OrchestratorPool.Size())
```
j0sh: This will mean a RPC call to all O in the pool which seems slow and potentially inconsistent. Do we need all this information within this endpoint? Would another endpoint for `/orchestrators` be more appropriate?

core/orchestrator.go
```diff
@@ -522,6 +551,7 @@ func (rtm *RemoteTranscoderManager) Register(transcoder *RemoteTranscoder) {
 
 func (rtm *RemoteTranscoderManager) Unregister(t *RemoteTranscoder) {
 	rtm.RTmutex.Lock()
+	// Probably should remove from remoteTranscoders too?
```
j0sh: We fix up the list during selection time to avoid iterating under the lock. (Also, since the multi-T PR is still in review, welcome to submit comments there)

server/ot_rpc.go
```diff
@@ -141,7 +142,11 @@ func runTranscode(n *core.LivepeerNode, orchAddr string, httpc *http.Client, not
 // Orchestrator gRPC
 
 func (h *lphttp) RegisterTranscoder(req *net.RegisterRequest, stream net.Transcoder_RegisterTranscoderServer) error {
-	glog.Info("Got a RegisterTranscoder request for ", req.Secret)
+	from := "unknown"
+	if p, ok := peer.FromContext(stream.Context()); ok {
```
j0sh, line 146: Very nice! Looking at the `peer` docs in [1] , what does it mean if the stream does not exist in the context? Is the stream disconnected? If so, can we just exit early and not register?

[1] https://godoc.org/google.golang.org/grpc/peer#FromContext

core/orchestrator.go
```diff
@@ -509,6 +526,18 @@ type RemoteTranscoderManager struct {
 	RTmutex           *sync.Mutex
 }
 
+func (rtm *RemoteTranscoderManager) RegisteredTranscodersCount() int {
+	return len(rtm.liveTranscoders)
+}
+
+func (rtm *RemoteTranscoderManager) RegisteredTranscodersInfo() []net.RemoteTranscoderInfo {
+	res := make([]net.RemoteTranscoderInfo, 0, len(rtm.liveTranscoders))
```
j0sh, line 529: Needs locking

core/orchestrator.go
```diff
@@ -478,25 +490,30 @@ func (rt *RemoteTranscoder) Transcode(fname string, profiles []ffmpeg.VideoProfi
...
 func NewRemoteTranscoder(n *LivepeerNode, stream net.Transcoder_RegisterTranscoderServer, capacity int) *RemoteTranscoder {
-	return &RemoteTranscoder{
+	t := &RemoteTranscoder{
 		node:     n,
 		stream:   stream,
 		eof:      make(chan struct{}, 1),
 		capacity: capacity,
+		addr:     "unknown",
+	}
+	if p, ok := peer.FromContext(stream.Context()); ok {
```
j0sh: Same question as earlier; under what circumstances is `peer.FromContext` not-`ok`?

core/orchestrator.go
```diff
@@ -478,25 +490,30 @@ func (rt *RemoteTranscoder) Transcode(fname string, profiles []ffmpeg.VideoProfi
...
 }
 
 func NewRemoteTranscoder(n *LivepeerNode, stream net.Transcoder_RegisterTranscoderServer, capacity int) *RemoteTranscoder {
-	return &RemoteTranscoder{
+	t := &RemoteTranscoder{
 		node:     n,
 		stream:   stream,
 		eof:      make(chan struct{}, 1),
 		capacity: capacity,
+		addr:     "unknown",
```
j0sh: This seems nicer than mutating the struct:

```golang
addr := "unknown"
if p, ok := peer.FromContext(stream.Context()); ok {
   addr = p.Addr.String()
}
return &RemoteTranscoder {
  ...
  addr: addr,
}
```

server/ot_rpc.go
```diff
@@ -141,7 +142,11 @@ func runTranscode(n *core.LivepeerNode, orchAddr string, httpc *http.Client, not
...
+		from = p.Addr.String()
+	}
+	glog.Infof("Got a RegisterTranscoder request for %s from %s", req.Secret, from)
```
j0sh: I think printing `req.Secret` might have been a mistake in the original code here.

Should we also tag the log, eg  `glog.Infof("Got a RegisterTranscoder request tc=%s", from)` and use the `tc`  tag (or something similar) throughout when logging transcoder info?

server/segment_rpc.go
```diff
@@ -348,7 +348,7 @@ func SubmitSegment(sess *BroadcastSession, seg *stream.HLSSegment, nonce uint64)
 		monitor.LogSegmentTranscoded(nonce, seg.SeqNo, transcodeDur, tookAllDur, common.ProfilesNames(sess.Profiles))
 	}
 
-	glog.Info("Successfully transcoded segment ", seg.SeqNo)
+	glog.Infof("Successfully transcoded segment segName=%s seqNo=%d", seg.Name, seg.SeqNo)
```
j0sh: Should we use `mid=%s seq=%s", sess.ManifestID, sess.SeqNo` here instead? (Or the mid in addition to the segName which is basically a URL if I recall.)

As an aside, it'd probably be good to settle on conventions for what and how to tag fields in logs.

j0sh (changes requested):

core/orchestrator.go
```diff
@@ -320,17 +322,25 @@ func (n *LivepeerNode) transcodeSeg(config transcodeConfig, seg *stream.HLSSegme
...
 	}
 	// transcodeEnd := time.Now().UTC()
 	if len(tData) != len(md.Profiles) {
-		glog.Errorf("Did not receive the correct number of transcoded segments; got %v expected %v", len(tData), len(md.Profiles))
+		glog.Errorf("Did not receive the correct number of transcoded segments; got %v expected %v manifest=%s seqNo=%d", len(tData),
+			len(md.Profiles), string(md.ManifestID), seg.SeqNo)
 		return terr(fmt.Errorf("MismatchedSegments"))
 	}
+	for i, segData := range tData {
+		if len(segData) == 0 {
```
j0sh: When does this happen? If we do keep this type of check, maybe better to do it when we also iterate over `md.Profiles`

darkdarkdragon: Originally I've hit this in #845. I've spent some time figuring this out, so I've thought we should keep this check.
Why it's better to check `md.Profiles`? It's just checking that there is some data, not that it's aligned with `profiles`.

j0sh: Yeah, it doesn't really hurt to have the check, but I guess that brings up the question why is a length of 0 special as opposed to some other reasonably small minimum number.

> Why it's better to check md.Profiles? It's just checking that there is some data, not that it's aligned with profiles.

So we don't iterate over the results multiple times. We already check whether `tData[i] == nil` so might as well add a ` || len(tData[i]) < minimum` here. And probably returning an error instead of continuing.

darkdarkdragon: OK, what minimum size it could be? What the minimum size of valid but empty segment?

server/segment_rpc.go
darkdarkdragon: Changed.
Yes, it would be good.

server/ot_rpc.go
darkdarkdragon: If I'm not mistaken, stream and context created [here](https://github.com/grpc/grpc-go/blob/master/internal/transport/http2_server.go#L330), so stream should always have context with peer.

server/mediaserver.go
darkdarkdragon: Yes, it was me being lazy here - I actually wanted just list of urls contained inside `orchestratorPool`, but this wasn't available in interface so I've took what was available. 
What do you think about adding `GetURLs` like this:
```golang
type OrchestratorPool interface {
       GetURLs() []*url.URL
	GetOrchestrators(int) ([]*OrchestratorInfo, error)
	Size() int
}
```
?

Also, I think separate `/orchestrators` endpoint makes sense if deliberately want it to hit orchestrators and check what alive (like it's happening now), but for `/status` I just wanted to some sneak peek into nodes state.

core/orchestrator.go
darkdarkdragon: Fixed

core/orchestrator.go
darkdarkdragon: Fixed

core/orchestrator.go
darkdarkdragon: Made fix

darkdarkdragon: In unit tests transcoder returns segment with length 25 - is it ok?

server/mediaserver.go
j0sh: The suggested `GetURLs` method sounds good!

core/orchestrator.go
j0sh: Sure, I can't think of a better number either.

server/ot_rpc.go
j0sh: Kinda funny how grpc goes against golang's own guidelines not to store contexts in a struct: https://golang.org/pkg/context/

j0sh:
Rebased the multi-T PR, so took the liberty of rebasing this one too. No code changes except for a small merge conflict around orch_test.

core/orchestrator.go
```diff
@@ -474,25 +481,30 @@ func (rt *RemoteTranscoder) Transcode(fname string, profiles []ffmpeg.VideoProfi
 	case <-ctx.Done():
 		// XXX remove transcoder from streams
 		rt.eof <- struct{}{}
-		return [][]byte{}, fmt.Errorf("Remote transcoder took too long")
+		return [][]byte{}, fmt.Errorf("Remote transcoder %s took too long", rt.addr)
```
j0sh: Maybe also add `transcoder=%s taskId=%s` here?

darkdarkdragon:
@j0sh 
> Rebased the multi-T PR, so took the liberty of rebasing this one too.

Looks like I don't need do anything here? From what I see in git, my PR is already rebased with Multi-T PR together?

core/orchestrator.go
darkdarkdragon: done

server/ot_rpc.go

server/mediaserver.go
darkdarkdragon: Added `GetURLs` method

darkdarkdragon:
@j0sh rebased this one

discovery/db_discovery.go
```diff
@@ -41,6 +42,20 @@ func NewDBOrchestratorPoolCache(node *core.LivepeerNode) *DBOrchestratorPoolCach
...
 
+func (dbo *DBOrchestratorPoolCache) GetURLs() []*url.URL {
+	orchs, err := dbo.node.Database.SelectOrchs()
+	if err != nil || len(orchs) <= 0 {
+		return nil
+	}
+
+	var uris []*url.URL
+	for _, orch := range orchs {
+		uri, _ := url.Parse(orch.ServiceURI)
```
j0sh: probably should check for an error and `continue` if there's something wrong; otherwise the url will be empty

core/orchestrator.go
```diff
@@ -515,6 +518,22 @@ type RemoteTranscoderManager struct {
 	RTmutex           *sync.Mutex
 }
 
+func (rtm *RemoteTranscoderManager) RegisteredTranscodersCount() int {
```
j0sh, line 521: Needs tests

core/orchestrator.go
```diff
@@ -515,6 +518,22 @@ type RemoteTranscoderManager struct {
...
+	rtm.RTmutex.Lock()
+	defer rtm.RTmutex.Unlock()
+	return len(rtm.liveTranscoders)
+}
+
+func (rtm *RemoteTranscoderManager) RegisteredTranscodersInfo() []net.RemoteTranscoderInfo {
```
j0sh, line 527: Needs tests

discovery/db_discovery.go
```diff
@@ -41,6 +42,20 @@ func NewDBOrchestratorPoolCache(node *core.LivepeerNode) *DBOrchestratorPoolCach
 	return &DBOrchestratorPoolCache{node: node}
 }
 
+func (dbo *DBOrchestratorPoolCache) GetURLs() []*url.URL {
```
j0sh, line 45: Needs tests; be sure to cover the error cases as well

server/webserver.go
```diff
@@ -740,23 +740,13 @@ func (s *LivepeerServer) StartWebserver(bindAddr string) {
...
-			}
-			d := struct {
-				Manifests map[string]string
-				Version   string
-			}{
-				Manifests: mstrs,
-				Version:   core.LivepeerVersion,
-			}
-			if data, err := json.Marshal(d); err == nil {
+			if data, err := json.Marshal(status); err == nil {
```
j0sh, line 749: Should we start doing expects-style testing for this?

j0sh (changes requested):

server/webserver.go
darkdarkdragon: What do you mean by that?

j0sh: Mostly checking the JSON output of the `/status` call. Roughly paraphrased:

```
w *ResponseWriter
r *Request
err = test_call('/status', w, r)
expected := `{"manifests": ..., "version": ... }` 
body := w.Body.ReadAll()
if body != expected {
  t.Error("Body did not match; expected '%s 'got '%s'", expected, body)
}  
```

Especially since the `GetNodeStatus` call seems to have a lot of possible variations (orchestrators, registered transcoders, etc). Alternatively, we could have more involved tests for `GetNodeStatus` and a simpler test for this `/status` endpoint.

(The CLI API is lacking test coverage; improving that as we make CLI changes seems like a good opportunity.)

j0sh:
Fixed to line up with the latest multi-T branch

discovery/db_discovery.go
darkdarkdragon: Fixed

core/orchestrator.go
darkdarkdragon: Added

core/orchestrator.go
darkdarkdragon: Added

discovery/db_discovery.go
darkdarkdragon: Added tests

server/webserver.go
darkdarkdragon: Added one simple test for `/status` endpoint

j0sh (approved):
🎉 🎉 First CLI tests!! 🎉 🎉 

Feel free to rebase this into the multi-T PR

common/testing_stubs.go
```diff
@@ -0,0 +1,47 @@
+package common
```
j0sh: Actually maybe this could incorporated into our existing `testutil.go` ? I also wonder if it'd be good to rename that file `util_test.go` so we aren't compiling that code in by default (or rename this one to stub_test.go`).

darkdarkdragon: I'll move it into `testutil.go`, but it can't be ranamed to `util_test.go` - or else I could have been able just use `core.StubServerStream` from my new test.

j0sh: Ah, test files from other modules aren't incorporated? Bummer, alright.

darkdarkdragon:
Merged this into Multi-T branch

PR #852, opened 2019-05-08
Add buffer to the channel to avoid deadlock

### LINKED ISSUE #843: Test-harness instances stop receiving streams after certain # of streams sent to it
**Describe the bug**

When streaming 5 streams into the test-harness (3 from twitch, one from OBS, one from test-harness BBB video), using [this](https://gist.github.com/angyangie/7bedeffd53e1b50e99b24b4a5fb9c00a) `multigroup.js` configuration (20 Ts, 20 Os, 1 B), all orchestrators seem to go offline periodically, for 3-5 minutes at a time. Transcoding rate goes down to 0%, we get `NoOrchestrators` and `ReadBody` errors from all O's, and ultimately discovery errors of  `caller=rpc.go:181 msg="Did not connect: context deadline exceeded"` and `"Discovery error=Could not get orchestrator: rpc error: code = Unknown desc = Invalid orchestrator request (OrchestratorCapped)"`. 

See dashboard here: http://34.74.10.88:3001/d/ZsZv-B3iz/livepeer-overview?orgId=1&from=1556313277185&to=1556314656952

**To Reproduce**

1. Use [this](https://gist.github.com/angyangie/7bedeffd53e1b50e99b24b4a5fb9c00a) `multigroup.js` to set up test-harness (run `node examples/multigroup.js`)
2. Stream using OBS to whatever RTMP port is output at the end of that file ^. Ex: rtmp://34.74.248.47:2548
3. Use the steps below and repeat 3 times to run 3 twitch live streams through our system. 

* Go to twitch, select any top channel
* Use youtube-dl, run youtube-dl -g <twitch_url>
* You should get a manifest URL
* Use ffmpeg: ffmpeg -re -i <manifest_url> -c:v copy -c:a aac -f flv <ingest_url>

4. Use ` ./test-harness stream angie-multigroup` with your own deployed test-harness name, to stream BBB.
5. Watch one of the streams using `ffplay http://34.74.248.47:2548/stream/current.m3u8` (insert your own IP address and port number)
6. Watch all transcoding cease

**Expected behavior**
Transcoding should remain at 100%.

**Screenshots**

Screen Shot 2019-04-26 at 5 23 24 PM

Screen Shot 2019-04-26 at 5 28 42 PM

Screen Shot 2019-04-26 at 5 29 01 PM

Screen Shot 2019-04-26 at 5 30 27 PM

darkdarkdragon:
Add buffer to the channel to avoid deadlock.

- 
- 
- 

Manually

Fixes #843

**Checklist:**

### DIFF
--- a/discovery/discovery_test.go
+++ b/discovery/discovery_test.go
@@ -1,15 +1,20 @@
 package discovery
 
 import (
+	"context"
 	"math/rand"
 	"net/url"
+	"runtime"
+	"sync"
 	"testing"
+	"time"
 
 	ethcommon "github.com/ethereum/go-ethereum/common"
 	"github.com/livepeer/go-livepeer/common"
 	"github.com/livepeer/go-livepeer/core"
 	"github.com/livepeer/go-livepeer/eth"
 	lpTypes "github.com/livepeer/go-livepeer/eth/types"
+	"github.com/livepeer/go-livepeer/net"
 	"github.com/livepeer/go-livepeer/server"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
@@ -45,6 +50,33 @@ func StubOrchestrators(addresses []string) []*lpTypes.Transcoder {
 	}
 
 	return orchestrators
+}
+
+func TestDeadLock(t *testing.T) {
+	gmp := runtime.GOMAXPROCS(50)
+	defer runtime.GOMAXPROCS(gmp)
+	var mu sync.Mutex
+	first := true
+	serverGetOrchInfo = func(ctx context.Context, bcast server.Broadcaster, orchestratorServer *url.URL) (*net.OrchestratorInfo, error) {
+		mu.Lock()
+		if first {
+			time.Sleep(100 * time.Millisecond)
+			first = false
+		}
+		mu.Unlock()
+		return &net.OrchestratorInfo{Transcoder: "transcoderfromtestserver"}, nil
+	}
+	addresses := []string{}
+	for i := 0; i < 50; i++ {
+		addresses = append(addresses, "https://127.0.0.1:8936")
+	}
+
+	assert := assert.New(t)
+	pool := NewOrchestratorPool(nil, addresses)
+	infos, err := pool.GetOrchestrators(1)
+	assert.Nil(err, "Should not be error")
+	assert.Len(infos, 1, "Should return one orchestrator")
+	assert.Equal("transcoderfromtestserver", infos[0].Transcoder)
 }
 
 func TestPoolSize(t *testing.T) {
--- a/discovery/discovery.go
+++ b/discovery/discovery.go
@@ -17,7 +17,9 @@ import (
 	"github.com/golang/glog"
 )
 
-const GetOrchestratorsTimeoutLoop = 1 * time.Hour
+const getOrchestratorsTimeoutLoop = 1 * time.Hour
+
+var serverGetOrchInfo = server.GetOrchestratorInfo
 
 type orchestratorPool struct {
 	uris  []*url.URL
@@ -79,15 +81,15 @@ func NewOnchainOrchestratorPool(node *core.LivepeerNode) *orchestratorPool {
 func (o *orchestratorPool) GetOrchestrators(numOrchestrators int) ([]*net.OrchestratorInfo, error) {
 	numAvailableOrchs := len(o.uris)
 	numOrchestrators = int(math.Min(float64(numAvailableOrchs), float64(numOrchestrators)))
-	ctx, cancel := context.WithTimeout(context.Background(), GetOrchestratorsTimeoutLoop)
+	ctx, cancel := context.WithTimeout(context.Background(), getOrchestratorsTimeoutLoop)
 	orchInfos := []*net.OrchestratorInfo{}
-	orchChan := make(chan struct{})
+	orchChan := make(chan struct{}, len(o.uris))
 	numResp := 0
 	numSuccessResp := 0
 	respLock := sync.Mutex{}
 
 	getOrchInfo := func(uri *url.URL) {
-		info, err := server.GetOrchestratorInfo(ctx, o.bcast, uri)
+		info, err := serverGetOrchInfo(ctx, o.bcast, uri)
 		respLock.Lock()
 		defer respLock.Unlock()
 		numResp++
--- a/server/mediaserver.go
+++ b/server/mediaserver.go
@@ -346,8 +346,8 @@ func (s *LivepeerServer) registerConnection(rtmpStrm stream.RTMPVideoStream) (*r
 	}
 	hlsStrmID := core.MakeStreamID(mid, &vProfile)
 	s.connectionLock.Lock()
-	defer s.connectionLock.Unlock()
 	_, exists := s.rtmpConnections[mid]
+	s.connectionLock.Unlock()
 	if exists {
 		// We can only have one concurrent stream per ManifestID
 		return nil, ErrAlreadyExists
@@ -362,11 +362,14 @@ func (s *LivepeerServer) registerConnection(rtmpStrm stream.RTMPVideoStream) (*r
 		profile:     &vProfile,
 		sessManager: NewSessionManager(s.LivepeerNode, playlist),
 	}
+	s.connectionLock.Lock()
 	s.rtmpConnections[mid] = cxn
 	s.lastManifestID = mid
 	s.lastHLSStreamID = hlsStrmID
+	sessionsNumber := len(s.rtmpConnections)
+	s.connectionLock.Unlock()
 	if monitor.Enabled {
-		monitor.CurrentSessions(len(s.rtmpConnections))
+		monitor.CurrentSessions(sessionsNumber)
 	}
 
 	return cxn, nil


j0sh:
Nice! Can you elaborate on how this leads to a deadlock? And maybe add a test that fails with the fix disabled?

darkdarkdragon:
Pseudo code:

```golang
func main() {
   	respLock := sync.Mutex{}
   	orchChan := make(chan struct{})

    func worker() {
        respLock.Lock()
        defer respLock.Unlock()
        orchChan <- struct{}{}
    }
    go worker() // instance1
    go worker() // instance2

	select {
	case <-orchChan:
		respLock.Lock()
        
		respLock.Unlock()
	}
}
```

Flow:
`instance1` takes the lock
`instance2` stops at `respLock.Lock()` and waits for the lock to be released
`instance1` writes to `orchChan`
`case <-orchChan` branch (inside `select` statement) starts executing
`instance1` releases lock
`instance2` takes the lock
`case` branch stops on `respLock.Lock()` and waits for the lock
`instance2` writes to `orchChan`, but stops and waits for somebody to read from the channel
 result: both `instance2` and `case` branch waiting indefinitely - deadlock happened

darkdarkdragon:
@j0sh added unit test

j0sh:
Thanks for the explanation! Could that also be added to the commit message for posterity? (I'd like to get into the habit of writing more informative commit messages for the historical record.)

This deadlock aside, I wonder if another contributing factor is [this blocking call](https://github.com/livepeer/go-livepeer/blob/5bfc71a47afe50c45394678453410aeabad512f9/server/broadcast.go#L153) in the middle of `NewSessionManager`, which is called under `connectionLock`. Should we also make that call async [1], or increase the [locking granularity](https://github.com/livepeer/go-livepeer/blob/5bfc71a47afe50c45394678453410aeabad512f9/server/mediaserver.go#L348-L372) in `registerConnection` so we aren't holding the lock for so long?

[1] The drawback to making the `refreshSessions` async seems to be that we aren't guaranteed to have O available when the first segment comes in. Which might be OK.

discovery/discovery_test.go
```diff
@@ -47,6 +59,57 @@ func StubOrchestrators(addresses []string) []*lpTypes.Transcoder {
...
+		}
+	}()
+	return lis
+}
+
+func testDialer(string, time.Duration) (gnet.Conn, error) {
+	return lis.Dial()
+}
+
+func TestDeadLock(t *testing.T) {
```
j0sh, line 55: Maybe we could stub out the `server.GetOrchestrator` part rather than building out a bunch of scaffolding around a real network etc? Something like:

```golang
type orchestratorPool struct {
 ...
orchInfo: func(context.Context, *url.URL) (*net.OrchestratorInfo, error)
}

func NewOrchestratorPool(...) *orchestratorPool {
...
// bcast is only used for server.GetOrchestratorInfo so can prob be removed from the struct at this point
bcast := core.NewBroadcaster(node)
orchInfo := func(context.Context, *url.URL) (*net.OrchestratorInfo, error) { return server.GetOrchestratorInfo(ctx, bcast,url)
return &orchestratorPool{orchInfo}
}
...
// testing code
pool := NewOrchestratorPool(nil, addresses)
pool.orchInfo = func ... // avoids the whole network dance!
```

@angyangie and I have encountered similar issues a couple times in the past and came close to implementing something like this a few times, but have always managed to work around it... might be unavoidable now.

What do you think?

darkdarkdragon:
Added explanation to commit message

> he drawback to making the refreshSessions async seems to be that we aren't guaranteed to have O available when the first segment comes in. 

With that we will not be able to achieve 100% success, and probably it make our target of 99.999% unachievable too.

>  increase the locking granularity in registerConnection so we aren't holding the lock for so long?

Does call to `NewSessionManager` really needs to be under `connectionLock`?

discovery/discovery_test.go
darkdarkdragon: I prefer not to have dynamic code - it is harder to follow. But it is unavoidable of course, and should be done for new code.
For that particular unit test I would prefer to left it as is - it shows example how to test `grpc` services/mock `grpc` service's server side. 

For the future - maybe it's better not to change `orchestratorPool`, but change `server.GetOrchestratorInfo` to `server.globalVar.GetOrchestratorInfo` - that way we will be able to replace `server.globalVar` for testing - this will give us ability to mock it for all the code that uses `server.GetOrchestratorInfo`?

j0sh:
> With that we will not be able to achieve 100% success, and probably it make our target of 99.999% unachievable too.

Note that we already have "missing" segments towards the end ("EndSession") so having a couple missing at the beginning *might* be reasonable for live if we stipulate there are boundary conditions. The 99.999 is amortized over time.

> Does call to NewSessionManager really needs to be under connectionLock?

Only these fields need to be protected by the lock:

https://github.com/livepeer/go-livepeer/blob/5bfc71a47afe50c45394678453410aeabad512f9/server/mediaserver.go#L77-L82

discovery/discovery_test.go
j0sh: > For that particular unit test I would prefer to left it as is

Disagree. gRPC is not required here, and really clutters up the testing; it's difficult to distinguish what's essential to the test and what's (avoidable) scaffolding.

Agree that using function pointers is a hackish, but it seems to be the least worst option.

> For the future - maybe it's better not to change orchestratorPool, but change server.GetOrchestratorInfo to server.globalVar.GetOrchestratorInfo

If you really want to use a global rather than a member var, the global can be within `discovery.go` : `var getOrchInfo = server.getOrchestratorInfo`. Within the testing procedure, we can reset `getOrchInfo = <stub>`.

darkdarkdragon:
> Note that we already have "missing" segments towards the end ("EndSession")

I not counting it in `success rate` metric.

> The 99.999 is amortized over time.

Per stream - yes, for all streams - they add up - every new stream will decrease success rate of broadcaster.

darkdarkdragon:
> Only these fields need to be protected by the lock:

Added new commit that changes locking in `registerConnection`

j0sh:
> I not counting it in success rate metric.

If we make that kind of exception towards the end, then why not towards the beginning also?

In general I agree it'd be better not to drop segments. But our numbers are already wiggly. we really need to come to a consensus as a team about how we define and understand the success rate, but that's a separate discussion. Will make an issue for it shortly.

One concern I have is that any further blockage here prevents the B itself from getting any more work done. The locking fixes help so this doesn't affect *other* B's, but the current B will still be blocked, IIRC. Do you know the extent of this? What's the behavior under extreme cases such as a `sleep(10 minutes)` within `refreshSessions`?

discovery/discovery_test.go
darkdarkdragon: Made the change

darkdarkdragon:
>  What's the behavior under extreme cases such as a sleep(10 minutes) within refreshSessions?

Probably `rtmp` connection will timeout. But `refreshSessions` shouldn't take more than one hour! 
Curious, why do we have `const getOrchestratorsTimeoutLoop = 1 * time.Hour`?
Probably we can safely reduce this timeout to four seconds - anyway, if orchestrator hasn't answered in that period - then it won't be able to do timely transcoding.

discovery/discovery_test.go
```diff
@@ -47,6 +51,30 @@ func StubOrchestrators(addresses []string) []*lpTypes.Transcoder {
 	return orchestrators
 }
 
+func TestDeadLock(t *testing.T) {
+	gmp := runtime.GOMAXPROCS(50)
+	defer runtime.GOMAXPROCS(gmp)
```
j0sh, line 57: Do we still need to set GOMAXPROCS?

darkdarkdragon: Yes. This test not always fails even with `GOMAXPROCS`, without it it fails with much less probability. (This test can't be made to fail reliably - it depends on go scheduler (or OS?) which decides who take lock first).

j0sh:
> Curious, why do we have const getOrchestratorsTimeoutLoop = 1 * time.Hour?

... I'm not sure. The timeout was originally 3 seconds but it got changed at some point.

> Probably we can safely reduce this timeout to four seconds

Yeah, anything from 2-4 seconds would be OK.

discovery/discovery_test.go
j0sh: I'd take a closer look to see if there's another way we can control the scheduling, but my brain is tired right now. LGTM 😄

j0sh (approved):
🚢 🚢 if the discovery tests are clean under `go test -race`

Great work narrowing this down @angyangie and @darkdarkdragon !

server/rpc.go
```diff
@@ -171,12 +172,20 @@ func GetOrchestratorInfo(ctx context.Context, bcast Broadcaster, orchestratorSer
...
+var TestDialer func(string, time.Duration) (gnet.Conn, error)
+
 func startOrchestratorClient(uri *url.URL) (net.OrchestratorClient, *grpc.ClientConn, error) {
 	glog.Infof("Connecting RPC to %v", uri)
-	conn, err := grpc.Dial(uri.Host,
-		grpc.WithTransportCredentials(credentials.NewTLS(tlsConfig)),
-		grpc.WithBlock(),
-		grpc.WithTimeout(GRPCConnectTimeout))
+	opts := []grpc.DialOption{grpc.WithBlock(), grpc.WithTimeout(GRPCConnectTimeout)}
+	if TestDialer != nil {
```
j0sh: It seems confusing to leave this in without a corresponding usage example. Is there another `rpc.go` function that could use improved test coverage? It shouldn't block this PR though.

darkdarkdragon: I actually just forgot to remove it. Will remove it - it's not needed right now.

PR #854, opened 2019-05-10
Misc TranscoderManager Changes

j0sh:
A few changes on top of the multi-T branch. The most important functional change is adding the notion of "fatal" errors (description in the following commit message), T retries, and API shrinkage.

Initially made the changes for fatal errors because that area of the code was making me nervous, and the feeling turns out to have been warranted: we had a bug where *any* error, including errors within the segment itself, would cause a T to not be added back on the `remoteTranscoders` list, eventually leading to dangling entries in the `liveTranscoders` table.

Following this, the fatal / non-fatal distinction offered a good opportunity to retry transcoding under certain conditions. We now retry the transcoding for fatal, non-timeout errors. Non-fatal errors are likely to mean an issue with the segment or something external to the transcoder, so we don't retry those.

Unit tests were also missing in this area so added some. Notably, the API felt a bit clunky when writing these tests, since a bit of the responsibility for dealing with remote transcoders was shared between the TranscoderManager and the LivepeerNode.  The notes in the [test code](https://github.com/livepeer/go-livepeer/commit/dc7eabc8d2f105d5a6791211d6cb113585488230#diff-4eb362572b306b3d740814ad1eeb41f5R260) should give an idea of the clunkiness. Fatal errors would lead to permanent removal from the `remoteTranscoders` list as expected, but Unregister would not be automatically called to clean up the `liveTranscoders` table unless we also listen for the EOF within `LivepeerNode.serveTranscoder`. This also exposed a latent blocking bug similar to https://github.com/livepeer/go-livepeer/pull/852 ...

All that being said, we fix a couple things, get a tighter TranscoderManager API, and have an overall reduced footprint for the LivepeerNode.

If all this looks good then I can fix up the rest of the tests that are commented out.

https://github.com/livepeer/go-livepeer/commit/dc7eabc8d2f105d5a6791211d6cb113585488230 tmanager: Distinguish between fatal and non-fatal errors.
    
    Fatal errors likely indicate something wrong with the transcoder
    itself and will lead to a disconnect. Non-fatal errors are more
    likely to indicate an issue with the source stream or something
    other than the transcoder, so maintain the transcoder within the
    manager.
    
    This also enables safer retries of failed transcodes; only fatal
    errors are retries.

https://github.com/livepeer/go-livepeer/commit/ebd9a2241270912abdc075a1386f8d4f0085a6c2 tmanager: Retry on most fatal errors.

https://github.com/livepeer/go-livepeer/commit/082da834ef4600cb61b214158d57e7ad8d016e77 
tmanager: Collapse Register/Unregister into Manage.  …
This shrinks the API surface area and takes some of the logic
out of the LivepeerNode, which (hopefully) makes the manager
workflow more straightforward.

https://github.com/livepeer/go-livepeer/commit/d8b00a8f9cc44f45355d653667e2b4013abb77a3 tmanager: Move task tracking into manager.  …
Further reduces the footprint of the LivepeerNode.

Unit testing, manual testing

**Checklist:**

- [] README and other documentation updated

### DIFF
--- a/core/orch_test.go
+++ b/core/orch_test.go
@@ -88,10 +88,10 @@ func TestServeTranscoder(t *testing.T) {
 }
 
 func TestRemoteTranscoder(t *testing.T) {
-	n, _ := NewLivepeerNode(nil, "", nil)
+	m := NewRemoteTranscoderManager()
 	initTranscoder := func() (*RemoteTranscoder, *StubTranscoderServer) {
-		strm := &StubTranscoderServer{node: n}
-		tc := NewRemoteTranscoder(n, strm, 5)
+		strm := &StubTranscoderServer{manager: m}
+		tc := NewRemoteTranscoder(m, strm, 5)
 		return tc, strm
 	}
 
@@ -123,7 +123,7 @@ func TestRemoteTranscoder(t *testing.T) {
 	// simulate timeout
 	tc, strm = initTranscoder()
 	strm.WithholdResults = true
-	n.taskCount = 1001
+	m.taskCount = 1001
 	RemoteTranscoderTimeout = 1 * time.Millisecond
 	_, err = tc.Transcode("fileName", nil)
 	if err.Error() != "Remote transcoder took too long" {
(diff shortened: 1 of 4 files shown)


core/orch_test.go
```diff
@@ -219,6 +220,47 @@ func TestSelectTranscoder(t *testing.T) {
...
+	assert.NotNil(m.liveTranscoders[s])
+
+	// happy path
+	res, err := m.Transcode("", nil)
+	assert.Nil(err)
+	assert.Len(res, 1)
+	assert.Equal(string(res[0]), "asdf")
+
+	// non-fatal error should not remove from list
+	s.TranscodeError = fmt.Errorf("TranscodeError")
```
angyangie, line 278: So here, when we do successfully receive results from remote transcoder, but there's a transcodeError, that doesn't warrant us stopping working with the transcoder? Is it because the error could've been with the segment itself, or just an innocent mistake?

core/orchestrator.go
```diff
@@ -453,12 +453,22 @@ type RemoteTranscoderFatalError struct {
...
+var ErrRemoteTranscoderTimeout = errors.New("Remote transcoder took too long")
 
 func (r *RemoteTranscoder) Capacity() int {
 	return r.capacity
 }
 
 func (rt *RemoteTranscoder) Transcode(fname string, profiles []ffmpeg.VideoProfile) ([][]byte, error) {
+	signalEof := func(err error) ([][]byte, error) {
+		// select so we don't block indefinitely if there's no listener
+		select {
```
angyangie, line 463: I'm confused on the syntax here. We're not executing anything in the case of `rt.eof <- struct{}{}:` or in the case of `default:`. Why is it important to have that statement there in the first place?

core/orchestrator.go
```diff
@@ -563,9 +568,15 @@ func (rtm *RemoteTranscoderManager) Transcode(fname string, profiles []ffmpeg.Vi
 	}
 	res, err := currentTranscoder.Transcode(fname, profiles)
 	_, fatal := err.(RemoteTranscoderFatalError)
-	if !fatal {
-		rtm.completeTranscoders(currentTranscoder)
+	if fatal {
+		// Don't retry if we've timed out; broadcaster likely to have moved on
```
angyangie, line 598: Interesting! So we're not retrying here because the B probably has already sent the segment that's taking too long to another O / T pair for transcoding, so a retry would result in double transcoding? Is that kind of redundancy here a big issue?

core/orchestrator.go
```diff
@@ -515,19 +509,18 @@ type RemoteTranscoderManager struct {
...
-	defer rtm.RTmutex.Unlock()
 	rtm.liveTranscoders[transcoder.stream] = transcoder
-
 	for i := 0; i < transcoder.Capacity(); i++ {
 		rtm.remoteTranscoders = append(rtm.remoteTranscoders, transcoder)
 	}
-}
+	rtm.RTmutex.Unlock()
+
+	<-transcoder.eof
```
angyangie, line 553: Much more succinct way of building the `Register` and `Unregister` functions. I didn't think about using `<-transcoder.eof` as a blocker during this function (and have the rest of the function not execute), but I see how that works now.

core/orch_test.go
j0sh: Yes, exactly - we don't want to stop working with the transcoder if the response is "clean", even if that response indicates the transcoder encountered an error while transcoding. We only stop working if it seems that there's a connectivity issue with the transcoder.

There might well still be certain types of errors from the transcoder that would warrant dropping the T entirely, but we'd need to to take the time to analyze what those cases might be.

core/orchestrator.go
j0sh: Yeah, this is a golang-ism that stems from the behavior of go channels.

Basically, writing to a channel blocks if the channel is full (or unbuffered without a corresponding listener, as we discovered in https://github.com/livepeer/go-livepeer/pull/852). We can avoid that blocking by putting the channel write (`rt.eof <- struct{}{}`) into a `select` , which will fall through to the default case if the write will block.

Encountered this block while unit-testing. Basically we can't assume the `eof` channel will always have an external listener attached to it. Eg, with two concurrent transcodes, if the first one fails, the second will hang since the first `eof` has been consumed.

core/orchestrator.go
j0sh: I didn't think of it either until noticed that the API felt clunky while writing the transcoding tests for the manager. That prompted a closer review, and here we are.

core/orchestrator.go
j0sh: Correct, it's a "selfish" optimization to save the O from doing unnecessary work if it's likely the B isn't depending on its results any longer. Don't think it's a big issue, but it was easy to add in. We might have to address the issue for VOD though, depending on how we do it.

core/orch_test.go
angyangie: I see. I agree that we should analyze what kind of errors from the transcoder itself would warrant dropping the T entirely. I imagine those cases do exist, but sounds like an optimization for later.

core/orchestrator.go
angyangie: I see what you're saying. If we have a transcoder with capacity of 5, for example, and have two transcodes happening at once, the second won't "end". Would another solution be to just make the eof channel a length of the transcoder's capacity? Or is this solution better since we don't have to specify that?

j0sh: Yeah, that approach of increasing the buffer size would probably work (and is what was used in https://github.com/livepeer/go-livepeer/pull/852)- note that `eof` itself is buffered with size 1.

Since a buffered / bounded channel can still block if the buffer was full for any reason, I think I still prefer this approach because there's no possibility of blocking.

angyangie: Yea that makes sense to me. But, I do have to say I'm proud of myself for thinking of an alternate solution relevant to some of the fixes we've seen in the last couple of days 😄

core/orchestrator.go
angyangie: Makes sense. I see how multi-T/O architecture wouldn't be very important for VOD, since we don't need real-time transcoding and can retry using the same O / T as many times as we wanted (unless there's a fatal mistake, O is down, etc.)

angyangie (approved):
This seems good to me!

j0sh:
Rebased against the latest multi-T and fixed up the tests.

Will leave this open over the weekend / Monday if anyone else wants to take a look at this, then merge this into multi-T and the whole thing by EOD Monday or early Tuesday.

PR #861, opened 2019-05-14
Graceful shutdown of transcoder

### LINKED ISSUE #847: Transcoder should notify orchestrator on exit
We don't have timeout on connection between O and T, so right now O becomes aware that T is no longer working is when it sends request to transcode segment.
Transcoder, upon exit, should sending notification to orchestrator, so it can remove it from list of alive transcoders.

darkdarkdragon:
On transcoder's shutdown, wait for currently transcoded segments to finish transcoding.
Notify orchestrator about shutdown, so it doesn't send any more tasks.

Slightly changes `Transcoder` interface to allow monitoring of connection state.

Manually

fixes #847

**Checklist:**


net/lp_rpc.proto
```diff
@@ -14,7 +14,7 @@ service Transcoder {
 
   // Called by the transcoder to register to an orchestrator. The orchestrator
   // notifies registered transcoders of segments as they come in.
-  rpc RegisterTranscoder(RegisterRequest) returns (stream NotifySegment);
+  rpc RegisterTranscoder(stream RegisterRequest) returns (stream NotifySegment);
```
j0sh, line 17: Curious, what's the advantage of making the RegisterTranscoder request a stream, rather than just having an explicit UnregisterTranscoder RPC call?

darkdarkdragon: It allows to also to catch situation when connection was lost but notification for it was not sent.

core/orchestrator.go
```diff
@@ -536,8 +536,38 @@ func (rtm *RemoteTranscoderManager) RegisteredTranscodersInfo() []net.RemoteTran
...
+					// remove it from map so no tasks assigned
+					// and wait till all current transcoding tasks will finish
+					rtm.RTmutex.Lock()
+					delete(rtm.liveTranscoders, transcoder.stream)
+					rtm.RTmutex.Unlock()
+					for {
+						if atomic.LoadInt64(&transcoder.transcodingSegments) == 0 {
+							break
+						}
+						time.Sleep(100 * time.Millisecond)
```
j0sh, line 560: This feels really fragile. If we are to handle this case at all, it seems better to do it on the client side: upon a graceful disconnect, reject new incoming segments, then close the stream once all segments are complete.

net/lp_rpc.proto
j0sh: The approach of turning the Register request into a stream feels weird. Will cancelling the context that we pass into `RegisterTranscoder` work? We'd probably have to propagate it from `RunTranscder / runTranscoder` . See https://github.com/grpc/grpc-go/issues/990

core/orchestrator.go
j0sh: Actually maybe we can close the stream on the client by cancelling the context (rather than having an explicit reject), then holding off on the shutdown until all segments are complete. T can continue segment processing as normal. This should work since the request to transcode the segment is separate from the segment notification request.

My guess is that a `sync.WaitGroup` for each transcode should do the trick, and doing a `wg.Wait()` after the `Recv` for-loop  within `runTranscoder`.

darkdarkdragon:
Closing this in favor of #864 - addressed comments there.

PR #864, opened 2019-05-16
Graceful shutdown of transcoder (second try)

### LINKED ISSUE #847: Transcoder should notify orchestrator on exit

darkdarkdragon:
On transcoder's shutdown, wait for currently transcoded segments to finish transcoding.
Notify orchestrator about shutdown, so it doesn't send any more tasks.

Manually

Fixes #847

**Checklist:**

### DIFF
--- a/server/ot_rpc.go
+++ b/server/ot_rpc.go
@@ -125,56 +125,54 @@ func runTranscoder(n *core.LivepeerNode, orchAddr string, capacity int) error {
 }
 
 func runTranscode(n *core.LivepeerNode, orchAddr string, httpc *http.Client, notify *net.NotifySegment) {
-	{ // dummy indentation to maintain the original indentation which allows reviewers to clearly see what has changed in the commit
-		profiles, err := common.TxDataToVideoProfile(hex.EncodeToString(notify.Profiles))
-		if err != nil {
-			glog.Info("Unable to deserialize profiles ", err)
-		}
-
-		glog.Infof("Transcoding taskId=%d url=%s", notify.TaskId, notify.Url)
-		var contentType string
-		var body bytes.Buffer
-
-		tData, err := n.Transcoder.Transcode(notify.Url, profiles)
-		glog.V(common.VERBOSE).Infof("Transcoding done for taskId=%d url=%s err=%v", notify.TaskId, notify.Url, err)
-		if err != nil {
-			glog.Error("Unable to transcode ", err)
-			body.Write([]byte(err.Error()))
-			contentType = TranscodingErrorMimeType
-		} else {
-			boundary := randName()
-			w := multipart.NewWriter(&body)
-			for _, v := range tData {
-				w.SetBoundary(boundary)
-				hdrs := textproto.MIMEHeader{
-					"Content-Type":   {"video/MP2T"},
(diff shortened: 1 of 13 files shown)


server/ot_rpc.go
```diff
@@ -73,23 +77,40 @@ func runTranscoder(n *core.LivepeerNode, orchAddr string, capacity int) error {
 		}
 		return err
 	}
+	exitc := make(chan os.Signal)
```
j0sh, line 97: Maybe add a comment for this part, "Catch interrupt signal to shut down transcoder"

server/ot_rpc.go
```diff
@@ -73,23 +77,40 @@ func runTranscoder(n *core.LivepeerNode, orchAddr string, capacity int) error {
...
 			if status.Message() == SecretErr.Error() { // consider this unrecoverable
 				return SecretErr
 			}
+			glog.Infof("End of stream recieve cylcle because err=%v, waiting for running transcode jobs to complete", err)
+			wg.Wait()
 			return err
 		}
-		go runTranscode(n, orchAddr, httpc, notify)
+		wg.Add(1)
+		go runTranscode(n, orchAddr, httpc, notify, &wg)
```
j0sh: Could probably do this to keep `wg` local and avoid the need to propagate it:

```
wg.Add(1)
go func() {
  runTranscode(n, orchAddr, httpc, notify)
  wg.Done()
}()
```

server/ot_rpc.go
```diff
@@ -73,23 +77,40 @@ func runTranscoder(n *core.LivepeerNode, orchAddr string, capacity int) error {
...
 	httpc := &http.Client{Transport: &http.Transport{TLSClientConfig: &tls.Config{InsecureSkipVerify: true}}}
+	var wg sync.WaitGroup
 	for {
 		notify, err := r.Recv()
 		if err != nil {
 			status := status.Convert(err)
 			if status.Message() == SecretErr.Error() { // consider this unrecoverable
 				return SecretErr
 			}
+			glog.Infof("End of stream recieve cylcle because err=%v, waiting for running transcode jobs to complete", err)
```
j0sh: Would it make sense to check for the `codes.Canceled` here and return a `TranscoderShutdownErr` to get a more readable message in `RunTranscoder`?

Then it might also make sense to adapt the error-wrapping approach that O uses to signal fatal/non-fatal errors so we only need to check the error type once in `RunTranscoder`?

j0sh (changes requested):
Looking MUCH better!! Just a few small changes.

server/ot_rpc.go
darkdarkdragon: Fixed

server/ot_rpc.go

server/ot_rpc.go

darkdarkdragon:
@j0sh this is ready for review. Also added commit with changes to reduce number of linter errors

j0sh (approved):
Fabulous! 🚢 🚢 🚢

PR #842, opened 2019-04-26
Dockerfile for building livepeer node on Windows.

darkdarkdragon:
**What does this pull request do? Explain your changes. (required)**
New Dockerfile for building on Windows. Based on previous Eli work.
Here I've split out ffmpeg building part into separate Dockerfile (it is in docker-livepeer repo).
Also there is Powershell script (windows build.ps1) which automates building Livepeer executable and getting it out of Docker container.

**Specific updates (required)**
Added Dockerfile and Powershell script.

**How did you test each of these updates (required)**
Manually.

**Does this pull request close any open issues?**

**Checklist:**

- [x] README and other documentation updated
- [ ] Node runs in OSX and devenv
- [ ] All tests in `./test.sh` pass

PR #865, opened 2019-05-16
Automatically build and upload binaries for Linux, Mac, Windows

### LINKED ISSUE #112: Windows support for go-livepeer
We probably need to update some special logic to support windows paths and os calls.

iameli:
Adds scripts to build binaries for our three supported platforms, upload them to GCP, and share a notification in Discord when the new binaries are ready.

I was having trouble testing cross-platform changes; this should make it easier for everybody.

- Added appveyor.yml for Windows builds on AppVeyor CI
- Added .travis.yml for Mac builds on Travis CI (cc @ericxtang - need one more integration set up :) )
- Added a binary upload step to .circleci/config.yml for Linux builds

Once these are merged in, all three integrations will need `GCLOUD_TOKEN` and `DISCORD_URL` environment variables set.

- Downloaded and run the binaries on each platform.

Fixes #112 
Builds on @darkdarkdragon's work in #842

**Checklist:**


install_ffmpeg.sh
```diff
@@ -5,12 +5,22 @@ set -ex
...
+  fi
+fi
+
 if [ ! -e "$HOME/nasm/nasm" ]; then
   # sudo apt-get -y install asciidoc xmlto # this fails :(
   git clone -b nasm-2.14.02 https://repo.or.cz/nasm.git "$HOME/nasm"
   cd "$HOME/nasm"
   ./autogen.sh
-  ./configure --prefix="$HOME/compiled"
+  ./configure --prefix="$HOME/compiled" --enable-pic --enable-static ${HOST_OS:-} --disable-cli
```
j0sh: Are these options actually valid? Tried those here on OSX and got warnings for each of them (and I think we need to preserve the nasm CLI anyhow since that's how it's invoked; it's a compile-time utility, not a goclient runtime dependency.)

Also when/where is HOST_OS set and what would it be? I don't see it in the env vars on any of my machines.

install_ffmpeg.sh
```diff
@@ -20,15 +30,20 @@ if [ ! -e "$HOME/x264/x264" ]; then
...
+if [ $(uname) != "Darwin" ]; then
+  EXTRA_FFMPEG_FLAGS="--enable-cuda --enable-cuvid --enable-nvenc"
+fi
+
 if [ ! -e "$HOME/ffmpeg/libavcodec/libavcodec.a" ]; then
-  git clone -b n4.1 https://git.ffmpeg.org/ffmpeg.git "$HOME/ffmpeg" || echo "FFmpeg dir already exists"
+  git clone --depth 1 -b n4.1 https://git.ffmpeg.org/ffmpeg.git "$HOME/ffmpeg" || echo "FFmpeg dir already exists"
   cd "$HOME/ffmpeg"
-  ./configure --disable-programs --disable-doc --disable-sdl2 --disable-iconv \
+  ./configure ${TARGET_OS:-} --disable-programs --disable-doc --disable-sdl2 --disable-iconv \
```
j0sh, line 60: Is TARGET_OS needed for Windows? (Are we cross-compiling?)

install_ffmpeg.sh
```diff
@@ -41,6 +56,7 @@ if [ ! -e "$HOME/ffmpeg/libavcodec/libavcodec.a" ]; then
     --enable-filter=aresample,asetnsamples,fps,scale \
     --enable-encoder=aac,libx264 \
```
j0sh, line 71: That reminds me, I wonder if there are additional flags we'll need to add for a minimal build:

* h264_nvenc encoder
* scale_npp, hwupload_cuda and hwdownload filters
* Anything for the decoder hwaccel?

Rather than just add these in blindly, let's test this build in practice and see what the results look like, so we aren't cargo-culting unnecessary flags. I've been testing with a fully loaded ffmpeg and forgot that we might need to make updates to the minimal build.

.circleci/config.yml
```diff
@@ -19,34 +19,28 @@ jobs:
...
-      - run: go get github.com/aws/aws-sdk-go/aws
-      - run: go get -u google.golang.org/grpc
-      - run: go get github.com/jstemmer/go-junit-report
-      - run: go get github.com/pkg/errors
-      - run: go get github.com/stretchr/testify
-      - run: go get github.com/alecthomas/gometalinter && gometalinter --install
-      - run: go get -u -v go.opencensus.io/stats
-      - run: go get -u -v go.opencensus.io/tag
-      - run: go get -u -v contrib.go.opencensus.io/exporter/prometheus
+      - run: ./install_dependencies.sh
```
j0sh, line 22: If we're refactoring / cleaning up this part, I'd really much rather just `go get ./...` and fetch everything at once, since enumerating deps in this way feels brittle (eg, it's easy to lose track of what's actually required, so fresh installs could end up pulling down unnecessary packages).

I believe the results of `go get` are cached by golang after the first run, but it's tricky to determine when the cache needs to be updated. @darkdarkdragon mentioned that `go get -u ./...` is slow (since it then tries to fetch everything), but right now the packages that are `-u`'d are totally arbitrary (and it gradually slows things down the more we add).

One approach is to check in a file with a nonce or timestamp and cache that within CircleCI [1][2]. If we check in dependency changes, we can update the file. This script can pick that up and run `go get -u ./...` to update the deps. This step could be integrated as part of the Makefile build process as well.

[1] Just checked whether we could just `touch` and `stat` an empty file, but git doesn't seem to pick those up...

[2] In an ideal packaging world, this file would be a list of dependencies and a list of commit hashes corresponding to the version of the dependency we want. If anything on the list changes, the system would synchronize the local copy to the correct hashes. Probably too much trouble for us right now though?

.circleci/config.yml
```diff
@@ -19,34 +19,28 @@ jobs:
...
           name: Lint
           command: |
             gometalinter --disable-all --enable=gofmt --enable=vet --enable=golint --deadline=4m pm
 
       - run:
-          name: Compile CLI binary
+          name: Compile binaries
           command: |
-            PKG_CONFIG_PATH=$HOME/compiled/lib/pkgconfig go build -v cmd/livepeer_cli/*
+            PKG_CONFIG_PATH=$HOME/compiled/lib/pkgconfig make
```
j0sh: Using the default `make` target may generate a different version of the `net/lp.rpc.pb.go` protobufs file which could cause unexpected behaviors if different versions of the protobuf/grpc compiler happen to produce different code (eg, what's installed on Circle vs what we have checked into the codebase). We can either move the `net/lp_rpc.pb.go` target outside the `all` default target, or explicitly name each target, eg `make livepeer livepeer_cli`. FWIW, I prefer the latter approach.

appveyor.yml
```diff
@@ -0,0 +1,22 @@
+# This is the one with Docker
```
j0sh, line 1: Does this fix the Appveyor test breakage that github has been complaining about?

install_ffmpeg.sh
iameli: Whoops, updated with the latest version of `install_ffmpeg.sh` from docker-livepeer - HOST_OS is something we need for the MSYS2 build. Should see it now.

Deleted those parameters - looks like I accidentally copied them from the x264 build, where they do exist.

install_ffmpeg.sh
iameli: Correct. We're not cross-compiling but the configure scripts for x264 and ffmpeg have trouble autodetecting that we're doing an mingw64 build. There might be some way for me to better specify this in the Docker environment, but this seems to work too.

.circleci/config.yml
iameli: Good catch, updated to do exactly that.

appveyor.yml
iameli: It will once I work with @ericxtang to get a few more CI parameters tweaked :)

.circleci/config.yml
iameli: I seem to recall `go get -u ./...` not working at all; let me run a test and see if I'm remembering that properly.

darkdarkdragon: @j0sh @iameli I would rather not put any effort into optimizing this right now - I think it better to just to wait for the next `golang` release and switch to the modules.

> In an ideal packaging world, this file would be a list of dependencies and a list of commit hashes

Modules do just that

install_ffmpeg.sh
j0sh: Looked at this a bit more and this is what I came up with.

The `--extra-cflags` and `--extra-ldflags` are needed here because CUDA installs itself in a nonstandard location but may not be required for Appveyor builds.

I also have to build the goclient with the `LIBRARY_PATH=/usr/local/cuda/lib64` environment variable when ffmpeg is linked statically (as is here).

I'm also unsure sure if we should actually add these options by default to the base build, since a lot of users won't actually be GPU enabled and we don't want to foist unnecessary dependencies on them. But we should definitely enable it for CI. Maybe toggle via environment variable or checking the presence of `/usr/local/cuda` ?

```diff
diff --git a/install_ffmpeg.sh b/install_ffmpeg.sh
index f47b5508..59ec8c3e 100755
--- a/install_ffmpeg.sh
+++ b/install_ffmpeg.sh
@@ -32,15 +32,18 @@ if [ ! -e "$HOME/ffmpeg/libavcodec/libavcodec.a" ]; then
     --disable-muxers --disable-demuxers --disable-parsers --disable-protocols \
     --disable-encoders --disable-decoders --disable-filters --disable-bsfs \
     --disable-postproc --disable-lzma \
-    --enable-gnutls --enable-libx264 --enable-gpl \
+    --enable-gnutls --enable-libx264 --enable-gpl --enable-nonfree \
     --enable-protocol=https,rtmp,file \
     --enable-muxer=mpegts,hls,segment --enable-demuxer=flv,mpegts \
     --enable-bsf=h264_mp4toannexb,aac_adtstoasc,h264_metadata,h264_redundant_pps \
     --enable-parser=aac,aac_latm,h264 \
     --enable-filter=abuffer,buffer,abuffersink,buffersink,afifo,fifo,aformat \
-    --enable-filter=aresample,asetnsamples,fps,scale \
-    --enable-encoder=aac,libx264 \
+    --enable-filter=aresample,asetnsamples,fps,scale,scale_npp \
+    --enable-encoder=aac,libx264,h264_nvenc \
     --enable-decoder=aac,h264 \
+    --enable-cuda --enable-cuvid --enable-nvenc --enable-libnpp \
+    --extra-cflags=-I/usr/local/cuda/include \
+    --extra-ldflags=-L/usr/local/cuda/lib64 \
     --prefix="$HOME/compiled"
   make
   make install
```

iameli:
Closing in favor of #887 which does the same thing way better and brings GPU support to the table.

PR #879, opened 2019-05-22
Update LPMS API and webhook presets

### LINKED ISSUE #793: Return Structured Data From `createRTMPStreamIDHandler`
We invoke the auth webhook within `makeStreamID`. Using this webhook has the potential to be a versatile method of controlling the behavior of the stream. We currently only specify the manifestID via the webhook, but we could control a number of other per-stream properties, such as:

* Stream keys - https://github.com/livepeer/go-livepeer/issues/764
* Transcode options
* Segment length
* Playlist window size
* Per-stream storage configuration
* VOD config

However, in order to support these features, we need to persist additional data from the `createRTMPStreamIDHandler` function.

Currently, the `createRTMPStreamIDHandler` function has this signature:

```golang
 func createRTMPStreamIDHandler(url *url.URL) (strmID string)
```

where the strmID is a (stringified) streamID composed of the `manifestID/streamKey` . We could change this to expect an interface following:

```golang
type RTMPStreamData interface {
  func StreamID() string
}
 func createRTMPStreamIDHandler(url *url.URL) (data RTMPStreamData)
```

In the concrete implementation we can return a struct that satisfies the interface.  Within this struct, we can store any necessary data generated by the stream ID handler or the webhook. This would also allow us to store the streamID itself in a structured fashion, avoiding a string-parse step each time we extract the streamID from the existing `stream.RTMPVideoStream` object.
  
We can then modify `stream.RTMPVideoStream` to make the  `RTMPStreamData` available as an accessor, eg
```
type RTMPVideoStream interface {
    ... existing methods here ...
     func StreamData() RTMPStreamData
}
```

Within the mediaserver, we can reify the StreamData accessor to the concrete struct as needed.

1. Keep the current streamID construction, and store the webhook-returned data within the `rtmpConnection` struct. This means we would have to register the connection earlier within `createRTMPStreamIDHandler` . This is not ideal, as `createRTMPStreamIDHandler` is blocking, and registering the connection involves some heavy lifting.

2. An alternative to above to this would be to split the initialization of the `rtmpConnection` struct between the stream ID creation and the actual stream handler, but this is also non-ideal from a code flow and maintainability perspective.

3. Store the generated per-stream options as a global map within `LivepeerServer`. Write to that within `createRTMPStreamIDHandler` and access it within `gotRTMPStreamHandler`. However, additional per-stream fields within `LivepeerServer` are undesirable, as this increases the bookkeeping.

**Additional context**

Builds off this older issue on refactoring the internal mediaserver state: https://github.com/livepeer/go-livepeer/issues/480

j0sh:

Two primary changes here:

1. Changes the `createRTMPStreamIDHandler` to return structured data via the new `streamParameters` struct. This allows us to propagate additional information from the auth webhook. See https://github.com/livepeer/go-livepeer/issues/793 for more details.

This also has the effect of laying groundwork to more cleanly support additional ingest methods beyond RTMP (eg, API-initiated HLS). We can continue to refactor things into the `streamParameters` struct as necessary without being tied to RTMP (eg, populated via API call).

2. To make this change more immediately useful by way of example, added support for presets to the webhook callback. Hopefully this makes extending the webhook to support other features more straightforward (eg, storage).

- Refactors the `createRTMPStreamID` to match the updated LPMS API.
- Introduces a `streamParameters` struct that is returned via the handler and used downstream.
- Adds preset support to the webhook callback.
- Cleans up some webhook tests (and adds more tests in support of presets)

* Unit testing
* Manual testing via modifying the `simple_auth_server` sample code

Fixes https://github.com/livepeer/go-livepeer/issues/793

**Checklist:**

### DIFF
--- a/server/mediaserver.go
+++ b/server/mediaserver.go
@@ -60,8 +60,9 @@ var BroadcastJobVideoProfiles = []ffmpeg.VideoProfile{ffmpeg.P240p30fps4x3, ffmp
 var AuthWebhookURL string
 
 type streamParameters struct {
-	mid     core.ManifestID
-	rtmpKey string
+	mid      core.ManifestID
+	rtmpKey  string
+	profiles []ffmpeg.VideoProfile
 }
 
 func (s *streamParameters) StreamID() string {
@@ -74,6 +75,7 @@ type rtmpConnection struct {
 	stream      stream.RTMPVideoStream
 	pl          core.PlaylistManager
 	profile     *ffmpeg.VideoProfile
+	params      *streamParameters
 	sessManager *BroadcastSessionsManager
 }
 
@@ -94,8 +96,9 @@ type LivepeerServer struct {
 }
 
 type authWebhookResponse struct {
-	ManifestID string `json:"manifestID"`
-	StreamKey  string `json:"streamKey"`
+	ManifestID string   `json:"manifestID"`
+	StreamKey  string   `json:"streamKey"`
+	Presets    []string `json:"presets"`
 }
 
 func NewLivepeerServer(rtmpAddr string, httpAddr string, lpNode *core.LivepeerNode) *LivepeerServer {
@@ -116,14 +119,7 @@ func NewLivepeerServer(rtmpAddr string, httpAddr string, lpNode *core.LivepeerNo
 
 //StartServer starts the LPMS server
 func (s *LivepeerServer) StartMediaServer(ctx context.Context, transcodingOptions string) error {
-	bProfiles := make([]ffmpeg.VideoProfile, 0)
-	for _, opt := range strings.Split(transcodingOptions, ",") {
-		p, ok := ffmpeg.VideoProfileLookup[strings.TrimSpace(opt)]
-		if ok {
-			bProfiles = append(bProfiles, p)
-		}
-	}
-	BroadcastJobVideoProfiles = bProfiles
+	BroadcastJobVideoProfiles = parsePresets(strings.Split(transcodingOptions, ","))
 
 	glog.V(common.SHORT).Infof("Transcode Job Type: %v", BroadcastJobVideoProfiles)
 
@@ -168,12 +164,17 @@ func createRTMPStreamIDHandler(s *LivepeerServer) func(url *url.URL) (strmID str
 		var mid core.ManifestID
 		var err error
 		var key string
+		presets := BroadcastJobVideoProfiles
 		if resp, err = authenticateStream(url.String()); err != nil {
 			glog.Error("Authentication denied for ", err)
 			return nil
 		}
 		if resp != nil {
 			mid, key = parseStreamID(resp.ManifestID).ManifestID, resp.StreamKey
+			// Process transcoding options presets
+			if len(resp.Presets) > 0 {
+				presets = parsePresets(resp.Presets)
+			}
 		}
 
 		if mid == "" {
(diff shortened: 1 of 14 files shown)


server/mediaserver.go
```diff
@@ -116,14 +119,7 @@ func NewLivepeerServer(rtmpAddr string, httpAddr string, lpNode *core.LivepeerNo
...
 func (s *LivepeerServer) StartMediaServer(ctx context.Context, transcodingOptions string) error {
-	bProfiles := make([]ffmpeg.VideoProfile, 0)
-	for _, opt := range strings.Split(transcodingOptions, ",") {
-		p, ok := ffmpeg.VideoProfileLookup[strings.TrimSpace(opt)]
-		if ok {
-			bProfiles = append(bProfiles, p)
-		}
-	}
-	BroadcastJobVideoProfiles = bProfiles
+	BroadcastJobVideoProfiles = parsePresets(strings.Split(transcodingOptions, ","))
```
angyangie, line 122: When would we opt to use a global variable, as we do here, as opposed to putting `BroadcastJobVideoProfiles` within the LivepeerServer` or `LivepeerNode` itself? I know we add `BroadcastJobVideoProfiles` to each `BroadcastSession`, and the value doesn't change, so perhaps that value doesn't need to live in a struct?

j0sh: Yeah, it's a node-wide default. Globals aren't great, putting it in a struct is another option [1] but this is how it's been done for a while, and it's a (mostly) read-only value.

The behavior of `BroadcastJobVideoProfiles` isn't really being changed here, only choosing whether we actually use that default or the set of options returned via webhook.

[1] This is why folks sometimes say that "singleton structs" stuffed with random values are no different than globals, eg LivepeerServer or LivepeerNode.

angyangie: > [1] This is why folks sometimes say that "singleton structs" stuffed with random values are no different than globals, eg LivepeerServer or LivepeerNode.

 Good point. I can see why that'd be the case. That's what I was wondering! 

> The behavior of `BroadcastJobVideoProfiles` isn't really being changed here, only choosing whether we actually use that default or the set of options returned via webhook.

Makes sense. Wouldn't really make a difference if it were included in `LivepeerNode` or `LivepeerServer`. The value is only read from a couple of times.

server/mediaserver_test.go
```diff
@@ -278,48 +310,63 @@ func TestCreateRTMPStreamHandlerWebhook(t *testing.T) {
...
-	sid = createSid(u)
-	mid = parseManifestID(sid)
-	if mid != "xyz" || sid != "xyz/zyx" {
-		t.Error("Should set manifest / streamkey  to one provided by webhook")
-	}
-	AuthWebhookURL = ""
+	params = createSid(u).(*streamParameters)
+	mid = params.mid
+	assert.Equal(core.ManifestID("xyz"), mid, "Should set manifest to one provided by webhook")
+	assert.Equal("xyz/zyx", params.StreamID(), "Should set streamkey to one provided by webhook")
```
yondonfu, line 355: Do we want to check `params.rtmpKey` as well?

j0sh: Sure. This was just a conversion of an older set of tests, rather than adding new ones, but added a new one here.

yondonfu (approved):
LGTM!

j0sh:
Updated to point to merged LPMS

PR #880, opened 2019-05-22
[Don't Merge] docker ubuntu build

ya7ya:
this creates an `ubuntu:18.04` base that can be used alongside the `ffmpeg-base:nvidia` to build a GPU enabled livepeer without needing access to actual GPU to build. 

- 
- 
- 

**Checklist:**

### DIFF
--- a/docker/Dockerfile.ubuntu
+++ b/docker/Dockerfile.ubuntu
@@ -1,0 +1,92 @@
+# This file aimed for developers and optimized for quick re-builds
+# Shouild be used by running `make localdocker`
+FROM ffmpeg-base:nvidia as builder
+
+FROM golang:1-stretch as builder2
+ENV PKG_CONFIG_PATH /root/compiled/lib/pkgconfig
+WORKDIR /root
+RUN apt update \
+    && apt install -y \
+    git gcc g++ gnutls-dev 
+    # git gcc g++ gnutls-dev linux-headers
+COPY --from=builder /root/compiled /root/compiled/
+COPY --from=builder /usr/local/cuda /usr/local/cuda
+
+ENV PKG_CONFIG_PATH /root/compiled/lib/pkgconfig
+ENV CPATH /usr/local/cuda/include 
+ENV LIBRARY_PATH /usr/local/cuda/lib64
+
+WORKDIR /go/src/github.com/livepeer/go-livepeer
+COPY . .
+RUN go get -u -v ./...
+
+#RUN go get github.com/golang/glog
+#RUN go get -u -v github.com/ericxtang/m3u8
+#RUN go get github.com/aws/aws-sdk-go/aws
+#RUN go get -u google.golang.org/grpc
+#RUN go get github.com/pkg/errors
+#RUN go get github.com/stretchr/testify/mock
+#RUN go get -u -v go.opencensus.io/stats
+#RUN go get -u -v go.opencensus.io/tag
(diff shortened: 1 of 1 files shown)


docker/Dockerfile.ubuntu
```diff
@@ -0,0 +1,92 @@
...
+
+
+COPY cmd cmd
+COPY common common
+COPY core core
+COPY server server
+COPY discovery discovery
+COPY drivers drivers
+COPY net net
+COPY eth eth
```
angyangie, line 66: What's the purpose of copying to and from the same folder here?

docker/Dockerfile.ubuntu
```diff
@@ -0,0 +1,92 @@
+# This file aimed for developers and optimized for quick re-builds
+# Shouild be used by running `make localdocker`
+FROM ffmpeg-base:nvidia as builder
+
+FROM golang:1-stretch as builder2
```
angyangie, line 5: I'm interested in really learning how this Dockerfile works! To that end ...

Is `golang:1-stretch` another Docker imagine that's been built somewhere else?

docker/Dockerfile.ubuntu
```diff
@@ -0,0 +1,92 @@
...
+WORKDIR /root
+RUN apt update \
+    && apt install -y \
+    git gcc g++ gnutls-dev 
+    # git gcc g++ gnutls-dev linux-headers
+COPY --from=builder /root/compiled /root/compiled/
+#COPY --from=builder /usr/local/cuda /usr/local/cuda
+
+ENV PKG_CONFIG_PATH /root/compiled/lib/pkgconfig
+#ENV CPATH /usr/local/cuda/include
```
angyangie, line 16: Is this commented out for when we use CUDA drivers?

docker/Dockerfile.ubuntu
ya7ya: The `COPY` command in docker copies a folder during the build to the image being created, so for example when we run 
```bash
sudo docker build --rm -t livepeerbinary:latest -f docker/Dockerfile.debian .
```

the `.` is our source folder, so `COPY net net` would copy the folder  `go-livepeer/net` to the image working directory.

docker/Dockerfile.ubuntu
ya7ya: Yes, docker image can use other images as base layers , so `golang:1-stretch` is a public image available in dockerhub , which comes with golang installed.

this structure allows us to have smaller final image too, since you can have a large base where the binary is being compiled, then you'd create a fresh, much smaller layer that will be used at runtime, and just copy the binary from the previous layer. since most of the dependencies at compile won't be needed at runtime.

docker/Dockerfile.ubuntu
ya7ya: this is still a work in progress, the flag wasn't needed at that point. I'll remove it 👍

docker/Dockerfile.ubuntu
angyangie: Ah I see. It makes sense that it's being copied to the docker image! 💯

docker/Dockerfile.ubuntu
angyangie: Very cool! That makes a lot of sense. That certainly makes deployment much easier/faster. Thanks for the explanation Yahya!

iameli:
Incorporated into #887, thanks!

PR #882, opened 2019-05-23
vendor: Update go-ethereum to livepeer/go-ethereum@806420c

yondonfu:

Update vendored go-ethereum.

- Update vendored go-ethereum to livepeer/go-ethereum@806420c
- Update receipt status check in `eth/client.go`

**Checklist:**


darkdarkdragon:
LGTM

iameli:
IIRC go-ethereum is the thing that didn't work in Go 1.12 - worth checking if we can make that upgrade now.

j0sh (approved):

yondonfu:
@iameli Confirmed that the node builds with go-ethereum when using `go version go1.12.4 darwin/amd64`

yondonfu:
@iameli The node seems to build fine with go-ethereum on `go version go1.12.4 darwin/amd64` as well as within the debian Docker container using `golang:1-stretch` (which I believe corresponds to 1.12).

@darkdarkdragon I updated `.dockerbuild.deps` to only contain packages in the `vendor` directory in 7e06253. My understanding from [this comment](https://github.com/livepeer/go-livepeer/pull/740#discussion_r258477915) is that `.dockerbuild.deps` is primarily used for caching vendored packages when running a Docker build so it seems that we only needed to list vendored packages in `.dockerbuild.deps`. Let me know if this approach is flawed though...For reference, [this](https://github.com/livepeer/go-livepeer/pull/882/commits/7e062534fa0ed87409acdc74c838d9abc87b0125#diff-488b4b4c72cc4277418ab9fd645fe02bR33) is the command I used to generate the new `.dockerbuild.deps` file.

darkdarkdragon:
@yondonfu `.dockerbuild.deps` used to cache not only packages in `vendor` dir, but all the dependencies, including ones in `GOPATH` dir. Caching only `vendor` dir we're letting out big chunk of code out of caching.

yondonfu:
@darkdarkdragon Added other dependencies back in to `.dockerbuild.deps` and the Docker image build seems to work with the latest vendored go-ethereum updates

darkdarkdragon:
@yondonfu LGTM

PR #883, opened 2019-05-23
cli: Update wizard based on TicketBroker contract changes

### LINKED ISSUE #825: pm: Broadcaster can create a deposit and reserve via the node's CLI wizard
The node CLI wizard currently allows a broadcaster to create a deposit and penalty escrow (with the minimum penalty escrow defined by the contract). The `TicketBroker` contract will be updated to use a reserve instead of a penalty escrow and there will no longer be a contract defined minimum. So, the node CLI wizard workflow will need to be updated to support calling the `fundReserve()` and `fundDepositAndReserve()` contract functions.

yondonfu:

Update CLI wizard so that the funds deposit flow works again based on the latest TicketBroker contract changes.

- Update CLI wizard to work with latest TicketBroker contract changes
- 
- 

Ran `bash test.sh`.

Manually tested the CLI wizard on a private test network.

Fixes #825 

**Checklist:**

### DIFF
--- a/cmd/livepeer_cli/wizard_ticketbroker.go
+++ b/cmd/livepeer_cli/wizard_ticketbroker.go
@@ -10,16 +10,17 @@ import (
 
 	"github.com/golang/glog"
 	"github.com/livepeer/go-livepeer/eth"
+	"github.com/livepeer/go-livepeer/pm"
 )
 
 // SenderStatus represents a sender's current status
 type SenderStatus int
 
 const (
-	// Empty is a sender's status when it has no deposit or penalty escrow
+	// Empty is a sender's status when it has no deposit or reserve
 	Empty SenderStatus = iota
 
-	// Locked is a sender's status when it has a deposit or penalty escrow and has not initiated
+	// Locked is a sender's status when it has a deposit or reserve and has not initiated
 	// the unlock period
 	Locked
 
@@ -27,15 +28,15 @@ const (
 	Unlocking
 
 	// Unlocked is a sender's status when it is no longer in the unlocking period, but it still
-	// has a deposit or penalty escrow
+	// has a deposit or reserve
 	Unlocked
 )
 
 const (
-	senderEmptyStatusMsg     = "sender's deposit and penalty escrow are zero"
-	senderLockedStatusMsg    = "sender's deposit and penalty escrow are locked"
+	senderEmptyStatusMsg     = "sender's deposit and reserve are zero"
+	senderLockedStatusMsg    = "sender's deposit and reserve are locked"
 	senderUnlockingStatusMsg = "sender is in the unlocking period"
-	senderUnlockedStatusMsg  = "sender's deposit and penalty escrow are unlocked"
+	senderUnlockedStatusMsg  = "sender's deposit and reserve are unlocked"
 )
 
 func (w *wizard) deposit() {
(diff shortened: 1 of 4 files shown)


cmd/livepeer_cli/wizard_ticketbroker.go
```diff
@@ -45,64 +46,34 @@ func (w *wizard) deposit() {
...
-	fmt.Printf("Enter deposit amount in ETH (%v will be used cover the minimum penalty escrow) - ", eth.FormatUnits(reqPenaltyEscrowAmount, "ETH"))
-
-	amount := w.readPositiveFloatAndValidate(func(in float64) (float64, error) {
-		amountWei := eth.ToBaseUnit(big.NewFloat(in))
+	fmt.Printf("Enter deposit amount in ETH - ")
 
-		if amountWei.Cmp(reqPenaltyEscrowAmount) < 0 {
-			return 0, fmt.Errorf("must deposit at least %v to cover the minimum penalty escrow", eth.FormatUnits(reqPenaltyEscrowAmount, "ETH"))
-		}
+	depositAmount := w.readPositiveFloat()
```
j0sh: Are we OK with not validating this anymore?

cmd/livepeer_cli/wizard_ticketbroker.go
```diff
@@ -45,64 +46,34 @@ func (w *wizard) deposit() {
...
+	fmt.Printf("Enter reserve amount in ETH - ")
 
-	amountWei := eth.ToBaseUnit(big.NewFloat(amount))
-	depositAmount := new(big.Int).Sub(amountWei, reqPenaltyEscrowAmount)
+	reserveAmount := w.readPositiveFloat()
 
 	form := url.Values{
-		"depositAmount":       {depositAmount.String()},
-		"penaltyEscrowAmount": {reqPenaltyEscrowAmount.String()},
+		"depositAmount": {eth.ToBaseUnit(big.NewFloat(depositAmount)).String()},
```
j0sh, line 71: Just curious, we don't have precision issues when handling potentially 18 decimal places of float here?

cmd/livepeer_cli/wizard_ticketbroker.go
```diff
@@ -119,12 +90,23 @@ func (w *wizard) unlock() {
...
-	fmt.Printf("Current Penalty Escrow: %v\n", eth.FormatUnits(sender.PenaltyEscrow, "ETH"))
+	fmt.Printf("Current Reserve: %v\n", eth.FormatUnits(sender.Reserve, "ETH"))
+
+	if sender.ReserveState == pm.Frozen {
+		currRound := w.currentRound()
+
+		fmt.Printf("Current Round: %v\n", currRound)
+		fmt.Printf("Thaw Round: %v\n", sender.ThawRound)
+
+		fmt.Printf("Cannot deposit because sender's reserve is frozen and not yet thawed")
```
j0sh, line 101: Refresh my memory : we're only allowing one reserve at a time? So they can't deposit a fresh reserve if their old one has become frozen?

j0sh (approved):

cmd/livepeer_cli/wizard_ticketbroker.go
yondonfu: Yeah this is the case at the moment - if a sender's reserve is frozen the sender will be unable to do anything until the reserve is thawed. I'm still thinking about whether to change this in the contracts, but I think we can revisit this separately (tracking in https://github.com/livepeer/protocol/issues/292)

cmd/livepeer_cli/wizard_ticketbroker.go
yondonfu: There could be precision issues - I decided to stick with the current float64 helper usage in this PR for now and was thinking of coming back to re-consider helper usage separately (maybe switch to converting string -> `big.Float` instead of string -> float64)

PR #888, opened 2019-05-28
Update install.md

kyriediculous:

Add an extra step to building from source to make sure all necessary libraries are installed. 

- Added a step in install.md 
- Incremented subsequential steps

**Checklist:**
- [x ] README and other documentation updated
- [ x] Node runs in OSX and devenv
- [ x] All tests in `./test.sh` pass

### DIFF
--- a/doc/install.md
+++ b/doc/install.md
@@ -8,15 +8,21 @@ The easiest way to install Livepeer is by downloading the `livepeer` and `livepe
 ### Option 2: Build from source
 You can also build the executables from scratch.
 
-1. If you have never set up your Go programming environment, do so according to Go's [Getting Started Guide](https://golang.org/doc/install).
+&ensp; 1\. If you have never set up your Go programming environment, do so according to Go's [Getting Started Guide](https://golang.org/doc/install).
 
-2. You can fetch the code running `go get github.com/livepeer/go-livepeer/cmd/livepeer` in terminal.
+&ensp; 2\. You can fetch the code running `go get github.com/livepeer/go-livepeer/cmd/livepeer` in terminal.
 
-3. You need to install `ffmpeg` as a dependency.  Run `./install_ffmpeg.sh`.  This will install the dependencies in `~/compiled`.  You need to have `pkg-config` installed.
+&ensp; 3\. Make sure you have the necessary libraries installed:
 
-4. You can now run `PKG_CONFIG_PATH=~/compiled/lib/pkgconfig go build ./cmd/livepeer/livepeer.go` from the project root directory. To get latest version, `git pull` from the project root directory.
+* Linux: `apt-get update && apt-get -y install build-essential pkg-config autoconf nodejs gnutls-dev`
 
-5. To run tests in locall run `./test.sh`, to run in docker container run `./test_docker.sh`
+ * OSX: `brew update && brew install pkg-config autoconf node gnutls`
+
+&ensp; 4\. You need to install `ffmpeg` as a dependency.  Run `./install_ffmpeg.sh`.  This will install the dependencies in `~/compiled`.  You need to have `pkg-config` installed.
+
+&ensp; 5\. You can now run `PKG_CONFIG_PATH=~/compiled/lib/pkgconfig go build ./cmd/livepeer/livepeer.go` from the project root directory. To get latest version, `git pull` from the project root directory.
+
+&ensp; 6\. To run tests locally `./test.sh`, to run in docker container run `./test_docker.sh`
 
 
 ### Option 3: Using Docker


doc/install.md
```diff
@@ -12,11 +12,13 @@ You can also build the executables from scratch.
...
-3. You need to install `ffmpeg` as a dependency.  Run `./install_ffmpeg.sh`.  This will install the dependencies in `~/compiled`.  You need to have `pkg-config` installed.
+3. Make sure you have the necessary libraries installed `apt-get update && apt-get -y install build-essential pkg-config autoconf nodejs gnutls-dev`
 
-4. You can now run `PKG_CONFIG_PATH=~/compiled/lib/pkgconfig go build ./cmd/livepeer/livepeer.go` from the project root directory. To get latest version, `git pull` from the project root directory.
+4. You need to install `ffmpeg` as a dependency.  Run `./install_ffmpeg.sh`.  This will install the dependencies in `~/compiled`.  You need to have `pkg-config` installed.
 
-5. To run tests in locall run `./test.sh`, to run in docker container run `./test_docker.sh`
+5. You can now run `PKG_CONFIG_PATH=~/compiled/lib/pkgconfig go build ./cmd/livepeer/livepeer.go` from the project root directory. To get latest version, `git pull` from the project root directory.
+
+6. To run tests in locall run `./test.sh`, to run in docker container run `./test_docker.sh`
```
yondonfu: Small wording fix: "To run tests in locall" -> "To run tests locally"

yondonfu (changes requested):

yondonfu:
@kyriediculous We usually try to maintain a clean Git commit history when possible (benefits future code readers so they can better understand the historical flow of changes). For this PR, we can drop the commit that introduces a permission change (1cb3bb2) and the commit that reverts the permission change (72c692a). I usually use `git rebase` for this:

```
# Start an interactive rebase 3 commits back from the current HEAD
git rebase -i HEAD~3
```

The prompt will offer various commands to use for each of the listed commits:

```
pick <short-commit-hash> foo
pick <short-commit-hash> bar
pick <short-commit-hash> baz

# Commands:
# p, pick ...
# ...
# d, drop ..
```

We can use `drop` to remove the commits we no longer want to use and then push the changes on the relevant branch.

doc/install.md
```diff
@@ -12,11 +12,13 @@ You can also build the executables from scratch.
 
 2. You can fetch the code running `go get github.com/livepeer/go-livepeer/cmd/livepeer` in terminal.
 
-3. You need to install `ffmpeg` as a dependency.  Run `./install_ffmpeg.sh`.  This will install the dependencies in `~/compiled`.  You need to have `pkg-config` installed.
+3. Make sure you have the necessary libraries installed `apt-get update && apt-get -y install build-essential pkg-config autoconf nodejs gnutls-dev`
```
yondonfu: This command is only relevant for a Linux user using the `apt` package manager. Perhaps we should also provide the command for an OSX user using the `brew` package manager?

Ex.

Using `apt`

```
apt-get update && apt-get -y install build-essential pkg-config autoconf gnutls-dev
```

Using `brew`

```
brew update && brew install pkg-config autoconf node gnutls
```

I'm not sure what the equivalent for Windows is...maybe @iameli knows since he's been working on Windows builds?

kyriediculous: Is the windows build stable though? Otherwise we should probably not include it for now.
I imagine it involves some obnoxious steps like using Mingw or Cygwin and static builds for ffmpeg.

I would happily try it out though since I have dual boot on this machine.

yondonfu: > Is the windows build stable though?

Not sure, but I thought I'd note it.

We can include the OSX/Linux commands in this PR and come back and add Windows commands later on.

yondonfu (approved):
LGTM

doc/install.md
```diff
@@ -8,15 +8,21 @@ The easiest way to install Livepeer is by downloading the `livepeer` and `livepe
...
+&ensp; 1\. If you have never set up your Go programming environment, do so according to Go's [Getting Started Guide](https://golang.org/doc/install).
 
-2. You can fetch the code running `go get github.com/livepeer/go-livepeer/cmd/livepeer` in terminal.
+&ensp; 2\. You can fetch the code running `go get github.com/livepeer/go-livepeer/cmd/livepeer` in terminal.
 
-3. You need to install `ffmpeg` as a dependency.  Run `./install_ffmpeg.sh`.  This will install the dependencies in `~/compiled`.  You need to have `pkg-config` installed.
+&ensp; 3\. Make sure you have the necessary libraries installed:
 
-4. You can now run `PKG_CONFIG_PATH=~/compiled/lib/pkgconfig go build ./cmd/livepeer/livepeer.go` from the project root directory. To get latest version, `git pull` from the project root directory.
+* Linux: `apt-get update && apt-get -y install build-essential pkg-config autoconf nodejs gnutls-dev`
```
j0sh, line 17: Why does the goclient need nodejs?

kyriediculous: Ah good point, I think we copied this from an issue in another repo not related to livepeer.

yondonfu: I think this was copied from the [Dockerfile](https://github.com/livepeer/go-livepeer/blob/master/docker/Dockerfile#L7) in the repo - the nodejs dependency might've been leftover from another Dockerfile that was being used as a base image for this [nodejs program that used to be used for verification](https://github.com/livepeer/verification-computation-archive/blob/master/Dockerfile). We can probably do some cleanup and remove it.

darkdarkdragon: I don't think that mentioned Dockerfile is being used anywhere, so probably it could be just removed.

kyriediculous: Shall I fix this up in my next PR? feels kind of silly to open a seperate PR for it.

yondonfu: IMO there is no such thing as a PR that is too small to open 😃

Issue #492, opened 2018-07-26
More Robust Block/Event Watcher


yondonfu:
**Is your feature request related to a problem? Please describe.**

There are a few problems with the structure of our event subscription code at the moment:

- Does not handle chain re-orgs
- Does not handle block propagation delays when the Livepeer node connects to the Ethereum network via a remote service backed by pools of ETH nodes such as Infura. We have observed in testing that in some cases, the Livepeer node can receive an event from Infura, but not be able to retrieve the corresponding block immediately. I suspect the cause of this issue is that some nodes in Infura's pool might lag behind sometimes such that they do not all have a uniform view of the head of the chain.
- The Livepeer node stores a last seen block, but this is not synchronized with other event services because the last seen block is tracked via its own subscription and each event is processed using its own separate subscription. As a result, we can miss events if a particular event subscription breaks before the Livepeer node can process all the events it received. These events might not be recoverable using the approach in #315 in conjunction with the last seen block because there is no guarantee that the last seen block corresponds to the last set of events received by an event service.

**Describe the solution you'd like**

## Overview

Create a single block watcher service that pulls blocks from the head of the chain. Event services consume blocks from the block watcher and implement their own processing logic for specific events and are registered with the block watcher. The block watcher keeps track of the last seen block. For each block that it receives, it will pass it to each of the registered event services. 

Upon receiving a new block, an event service will use the block's bloom filter to test if the event topics it cares about are present in the block (for example, a specific broadcaster address that is used as an indexed argument in the `NewJob` event). If so, the event service will use the `eth_getLogs` RPC call with the appropriate log filter specifying the relevant event topics to retrieve the events for that block. The event service can then run its own custom processing (i.e update local DB).

If the block watcher stops and resumes after a period of time, it can pick up where it left off using its last seen block to back fill all the blocks in between the last seen block and the current head of the chain.

## Implementation

[...]

PR #890, opened 2019-05-29
pm: Round based ticket auxilary data creation & validation

### LINKED ISSUE #826: pm: Broadcaster can create tickets with creationRound & creationRoundHash
A broadcaster will need to fetch the current round and the block hash for the current round when creating a ticket. A naive way to do this would be to query the `RoundsManager` contract every time a ticket is created. A better way to do this could be to run a background goroutine that updates an in-memory cache containing the current round and the block hash for the current round and the broadcaster would fetch the current round and the block hash from the in-memory cache every time it creates a ticket.

### LINKED ISSUE #827: pm: Orchestrators can validate tickets with creationRound & creationRoundHash
An orchestrator will need to fetch the current round and the block hash for the current round when validating received tickets. We can re-use the in-memory cache solution described in #826 - the orchestrator would get the current round and the block hash from the in-memory cache every time it needs to validate a received ticket.

yondonfu:

Implements round based ticket auxilary data creation & validation. PM tickets include an `auxData` byte array field that can be used to include additional data that is used for validation. The implementation for this PR encodes the last initialized round and the block hash for that round in the `auxData` field. The recipient only considers `auxData` valid if the last initialized round and block hash encoded in `auxData` correspond to the current last initialized round and its block hash.

- Add `pm.AuxDataCreator` and `pm.AuxDataValidator` interfaces which are used by the `pm.Sender` and `pm.Validator` interfaces respectively
- Add `pm.RoundAuxDataCreator` which implements `pm.AuxDataCreator`
- Add `pm.RoundAuxDataValidator` which implements `pm.AuxDataValidator`
- Both `pm.RoundAuxDataCreator` and `pm.RoundAuxDataValidator` accept the `RoundsManager` interface. At the moment, the client from the `eth` package implements the `RoundsManager` interface. However, this is sub-optimal because the client from the `eth` package submits a network request for each call to `LastInitializedRound()` and `BlockHashForRound()` which increases latency in an already latency sensitive workflow (ticket creation is necessary for sending a segment and ticket validation is necessary for transcoding a segment). A more optimal solution would be to use an implementation of the `RoundsManager` interface that reads values for `LastInitializedRound()` and `BlockHashForRound()` from a cache (either in-memory or from the DB) with the cache being updated whenever a new round is initialized. The cache update logic could be implemented by watching for `NewRound` events emitted from the RoundsManager contract. My current preference is to use a more robust event watching architecture (as described in #492) rather than create another one off event watching service to implement the cache update logic. So, I excluded the more optimal implementation of the `RoundsManager` interface from this PR and plan on including the more optimal implementation after #492 is implemented.

Added unit tests and did some manual testing with the nodes running in on-chain mode.

Fixes #826 
Fixes #827

**Checklist:**


j0sh:
> PM tickets include an auxData byte array field that can be used to include additional data that is used for validation

Just curious, is there an advantage to having the auxData as a byte array rather than a more structured representation, eg `message ValidatorParams { creationRound, creationRoundBlockHash }` ? (This could be nested within `PMPayments`, and maybe wrapped in a `oneof` to accommodate the possibility of multiple validator protocols, but we can wrap fields in a oneof later without losing back-compat.)

Note that if we ever update the validation protocol, then both server and client need to be updated, so also updating the wire protocol doesn't seem like an issue, and is preferable since it avoids manual de/serialization, bakes in versioning, etc.

One reason I can think for doing it the current way is if auxData was intended to be opaque data that the goclient passes through to the validator without interpretation, but that doesn't seem to be the case here.

yondonfu:
> Just curious, is there an advantage to having the auxData as a byte array rather than a more structured representation

I included the `bytes aux_data` as a field in `message Ticket` to match the on-chain representation of a ticket - the TicketBroker contract expects the ticket provided in a function call to include an `auxData` byte array that encodes `creationRound` and `creationRoundBlockHash` (the original motivation behind using an `auxData` byte array in the on-chain representation of a ticket was to maintain flexibility for other contract implementations to include different auxilary data for validation rather than each of them defining custom ticket data structure with custom fields for validation i.e. an alternative implementation could encode the block timestamp in the auxilary data).

We could use a more structured representation of `creationRound` and `creationRoundBlockHash` in the wire protocol, but we would still need to encode the `creationRound` and `creationRoundBlockHash` into an `auxData` byte array prior to passing a ticket in a TicketBroker contract function call. So, whether we use the structured representation or not, the manual de/serialization step between `auxData` <-> `creationRound`/`creationRoundBlockHash` is still necessary - the only difference is when the step occurs. We could change the TicketBroker contract ticket definition to explicitly include `creationRound` and `creationRoundHash` as fields instead of using the `auxData` byte array, but I have a slight preference for keeping the `auxData` byte array there because if we ever have to change this extra validation step in the contract, we can just [modify a single function](https://github.com/livepeer/protocol/blob/pm/contracts/pm/mixins/MixinTicketProcessor.sol#L55).

I don't have a strong preference for using `bytes aux_data` in the wire protocol though so I'm happy to switch to a structured representation of `creationRound` and `creationRoundBlockHash` - baked in versioning is nice.

kyriediculous:
I personally have a minor preference for this to be structured data for the reasons Josh mentions, and I also agree with Yondon that the contract should remain to take a byte array.

j0sh:
Now that the smart contract is mentioned, a couple questions there:

How does this particular ticket validity check justify a separate encoding method for `auxData`? To extrapolate in either direction, we could store the entire ticket as a flat byte array (which is obviously undesirable, so why is it being done here?), or add the two new fields to the ticket?

Also, would we get more static assurances by parameterizing the check appropriately? This gives us an extra layer of safety, especially if the function signature changes. Type errors from the compiler can guide us through the refactoring at call sites. We'd have to make these changes anyway; this way, we're guaranteed not to miss any.

Note that I didn't review the protocol changes and I'm not that familiar with Solidity, so may be totally off base here.

yondonfu:
> How does this particular ticket validity check justify a separate encoding method for auxData?

The original thought was that a contract developer would use all the "standard" ticket validity checks (i.e. non-replay, valid signature, valid `recipientRand`, etc.), but might also wish to add an extra validity check for expiration (or perhaps another purpose). There are a number of ways to implement a validity check for expiration depending on the application. Currently, we choose to use `creationRound` and `creationRoundBlockHash`. Another application might choose to use a block timestamp. Encoding this data used for a validity check such as expiration in the `auxData` byte array allows a contract developer to easily extend/modify the PM contract implementation by swapping out the implementation for the function that interprets and checks `auxData`.

> Also, would we get more static assurances by parameterizing the check appropriately?

Yeah we do lose static assurances by encoding `creationRound` and `creationRoundBlockHash` in a `bytes` instead of in a pair of `uint256`. This is a downside.

> We'd have to make these changes anyway; this way, we're guaranteed not to miss any.

I agree though I see the primary purpose of using the `auxData` byte array as reusability/extensability for other developers besides ourselves.

yondonfu:
Pushed a few updates to this PR:

- Rebased using `master` (pulls in protobuf message from #903)
- Adds the `TicketExpirationParams` message which is a structured representation of the `creationRound` and `creationRoundBlockHash` which are used for ticket expiration related validation
- Adds the `TicketExpirationParams` message to the `Payment` message - this field does not need to be repeated since all tickets that are a part of a payment should have the same `creationRound` and `creationRoundBlockHash`
- Removed the `AuxData_` interfaces and instead moved the expiration related validation into `pm.Validator` and the expiration params fetching into `pm.Sender`

cmd/livepeer/livepeer.go
```diff
@@ -331,7 +333,9 @@ func main() {
 		}
 
 		if n.NodeType == core.BroadcasterNode {
-			n.Sender = pm.NewSender(n.Eth)
+			// TODO: Initialize Sender with an implementation
+			// of RoundsManager that reads from a cache
+			n.Sender = pm.NewSender(n.Eth, n.Eth)
```
kyriediculous, line 368: Why are we passing in the Eth client twice as an argument to `pm.NewSender` ? 

I understand that it implements both `Signer` and `RoundsManager` interfaces so why can't we just pass it once? 

Is this a design flaw in our code or is this a design flaw in go? 

Because as I see it the only option to "solve" this would be to have an additional package that defines the `pm.Ticket` and `pm.SenderInfo` types to be then used in both the `pm` and `eth` packages. Which is less clean/more code than just passing the same argument twice. But then again not allowing cyclical imports is considered to be enforcing good layered design. Seems quite paradoxal to me right now.

core/orch_test.go
```diff
@@ -754,10 +754,15 @@ func defaultPaymentWithTickets(t *testing.T, senderParams []*net.TicketSenderPar
 	}
 
 	sender := pm.RandBytes(123)
+	expirationParams := &net.TicketExpirationParams{
```
kyriediculous, line 771: Considering `ExpirationParams` aren't repeated values (similar to why we have seperated `TicketParams` from `TicketSenderParams`), I seem to have a preference to just define these within the `TicketParams` protobuf definition rather than creating an additional message that will never be passed over the wire by itself. Does this seem reasonable? 

This would be in line with the adjusted `Ticket` message as well, which I think we could also remove if we change the implementation for Creating Tickets around a bit to make optimal use of the new `Payment` construction, as we briefly discussed early last week. 

Or is this a decision resulting from wanting to have the pm package as a reference that allows for different implementations rather than for the pure and sole purpose of being used within the livepeer client?

cmd/livepeer/livepeer.go
yondonfu: > I understand that it implements both Signer and RoundsManager interfaces so why can't we just pass it once?

`pm.NewSender` accepts the `pm.Signer` and `pm.RoundsManager` interfaces separately to indicate that each of these interfaces could be implemented by a different struct. In this case, `LivepeerEthClient` happens to implement both the `pm.Signer` and `pm.RoundsManager` interfaces, but this need not always be the case. In fact, as mentioned in the TODO comment, the `pm.RoundsManager` interface will likely be implemented by a different struct later on i.e. a struct that can return values cached from the RoundsManager contract instead of requesting the values with each method invocation.

> Is this a design flaw in our code or is this a design flaw in go?

I think this looks strange because I chose not to include the caching implementation of `pm.RoundsManager` in this PR (ideally we can implement #492 first instead of adding more one off subscription code). Once we add the caching implementation of `pm.RoundsManager`, we won't be passing `n.Eth` as a parameter twice.

core/orch_test.go
yondonfu: > I seem to have a preference to just define these within the TicketParams protobuf definition rather than creating an additional message that will never be passed over the wire by itself. Does this seem reasonable?

The `TicketParams` message contains fields that the recipient provides to the sender that would be fixed across all tickets in a payment. `creation_round` and `creation_round_block_hash` are set by the sender when creating tickets so I don't think the `TicketParams` message is the right place for those fields.

I defined an additional `TicketExpirationParams` message to group together `creation_round` and `creation_round_block_hash` since these fields come as a pair i.e. we cannot have one without the other for expiration validation. I thought the nested message better conveyed this relationship. Furthermore, a smaller additional benefit is that if we ever support multiple types of expiration params then we could wrap `TicketExpirationParams` in a `oneof` as mentioned in [this comment](https://github.com/livepeer/go-livepeer/pull/890#issuecomment-499135690).

> if we change the implementation for Creating Tickets around a bit to make optimal use of the new Payment construction

I agree that we should think about how to optimize the current ticket creation process which is now inefficient.

> Or is this a decision resulting from wanting to have the pm package as a reference that allows for different implementations rather than for the pure and sole purpose of being used within the livepeer client?

This wasn't a big motivator for this particular decision (see my response earlier in this comment). At this point, the `pm` package is becoming increasingly coupled with Livepeer specific logic (since the reserve logic is heavily based on how the Livepeer protocol works) and I think that's ok since we're not too focused on generic reserve logic at the moment.

kyriediculous: > creation_round and creation_round_block_hash are set by the sender when creating tickets

Good point, I didn't consider this when making the comment. :100: 

>  I thought the nested message better conveyed this relationship.

Sounds reasonable. 

>  Furthermore, a smaller additional benefit is that if we ever support multiple types of expiration params then we could wrap TicketExpirationParams in a oneof

Since expiration is so tightly coupled to the protocol implementation I don't see this happening any time soon and the only  alternative I can think of is future block timestamps. Then again more flexibilty from the get-go sounds good. 

Thanks for claryfing ! approved.

kyriediculous (approved):
lgtm :+1:

Issue #868, opened 2019-05-16
Document O selection and reliability strategy


j0sh:
**Is your feature request related to a problem? Please describe.**
We're starting to accumulate quite a bit of "oral knowledge" on how the O selection mechanism works, the overall reliability strategy, and the justifications for that strategy. It would be good to document those.

There are two possible audiences here that the documentation should be addressed towards: developers working on the goclient, and users of the goclient (B/O/T operators). 

This probably means separate sections or separate pages; it would be good to clearly identify the target audience when writing.

**Describe the solution you'd like**

A `docs/reliability.md` page describing our accumulated knowledge and strategy.

**Describe alternatives you've considered**

Separate pages, one for developers and another for users. Whichever is more appropriate probably depends on the depth of each.

It may be enough to begin with a high-level description suitable for users, then drill down into details suitable for developers.

**Additional context**

Some things to describe, off the top of my head:

* Reliability and load control: round robin, retries
* Backpressure: One segment in flight
* Load balancing: `-maxSessions`
* BroadcasterManager list refresh intervals https://github.com/livepeer/go-livepeer/pull/806#discussion_r274187906
* Front running prevention: External storage prefix for multi-O
* Justification for each of these mechanisms, the problem(s) they solve and how

(this list is not exhaustive)

PR #901, opened 2019-05-30
Orchestrator Selection and Reliability Strategy Documentation

angyangie:

Documentation for how the O selection mechanism works, the overall reliability strategy, and the justifications for that strategy.

The document describes:
- Reliability and load control: round robin, retries
- Backpressure: One segment in flight
- BroadcasterManager list refresh intervals #806 (comment)
- Front running prevention: External storage prefix for multi-O
- Justification for each of these mechanisms, the problem(s) they solve and how

https://github.com/livepeer/go-livepeer/issues/868

**Checklist:**

### DIFF
--- a/doc/reliability.md
+++ b/doc/reliability.md
@@ -1,0 +1,27 @@
+# Livepeer Orchestrator Selection and Reliability Strategy
+
+## Overview
+
+To achieve greater network scalability, a broadcaster works with multiple orchestrators at once. The Broadcaster stops working with any orchestrator that has gone offline or does not return transcoded segments, and "refreshes" the list of orchestrators it works with if "enough" orchestrators on its original list are unresponsive (see `Orchestrator List Refresh`). The Broadcaster distributes segments to Orchestrators using a "round robin" strategy (see `Orchestrator Selection`), given that Orchestrator cannot have more than one segment per stream in flight (this mitigates back logging). Therefore, a Broadcaster sends segments to "free" Orchestrators on their "saved list". The ability to send segments to a selection of Orchestrators gives individual Orchestrators more time to process segments, and prevents `OrchestratorBusy` errors.
+
+## BroadcastSessionsManager
+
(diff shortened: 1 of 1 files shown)


doc/reliability.md
```diff
@@ -0,0 +1,27 @@
...
+
+## Orchestrator Selection
+
+To give preference to O's that respond with transcoded segments quickly, instead of selecting an Orchestrator from the beginning of `sessList` when needed, and placing new Orchestrators that are finished processing a segment at the end, `selectSession` takes Orchestrators from the end of `sessList`. If transcoding is successful, it adds them back to the end of `sessList`. 
+
+## Transcoding Errors & Retries
+
+If there is an error uploading segment to an Orchestrator's OS, submitting the segment to an Orchestrator, downloading transcoded segments, or the segment signature check fails, the Orchestrator is removed from the `sessMap`. The segment is retried with a different Orchestrator. When `selectSession` is called in this retry scenario, though the removed session might still exist in `sessList`, only a session that still exists in `sessMap` will be selected.  If there is no error in segment transcoding, `completeSession` adds session back to `sessList`.
```
darkdarkdragon: I think it will be good to add that retries stops when `sessMap` is empty.

darkdarkdragon (changes requested):

doc/reliability.md
angyangie: I'll add that!

darkdarkdragon (approved):
LGTM

j0sh:
This is a nice start, but it feels a bit heavy on the details of the implementation. 
Referring to particular variables in external documentation is a bit too brittle and specific (eg, `sessMap`, `sessList`); but it's a good indicator of where we can improve the in-code comments.

A higher level description of the behavior would be more accessible, and we can point interested readers to certain areas of the code for further follow-up on their own. Developers should be able to follow the code from the high-level description, while the current description will be confusing to non-developers (users) who might just want to understand the O selection mechanic.

angyangie:
Thanks for the feedback @j0sh. So you think it would be more readable if I replaced `sessList` with sessions list, `HTTPTimeout` with http timeout, `sessMap` with sessions map, `BroadcastSession` with broadcast session? Or you're saying I should re-write the content of a lot of this to simplify?

angyangie:
I will merge this as is, and make edits once @j0sh is back!

Issue #897, opened 2019-05-30
B generates a payment that will ensure that minimum credit requirement is met to include with segment


yondonfu:
**PaymentAmount()**

Should return the amount represented by a payment. When using PM, a payment is a ticket and the payment amount is the ticket EV.

**GenPayment(sess, transOpts)**

Should create a payment to be included with a segment to be sent to the orchestrator.

```
minCredit = MinCredit()
currCredit = CurrCredit(self)

paymentValue = 0
if currCredit < minCredit {
    // This will ensure that the orchestrator will at least
    // process the next segment, but the broadcaster technically
    // could send more
    // Note: currCredit can be negative here indicating that
    // the broadcaster needs to pay the orchestrator an amount
    // that covers the owed amount and the minimum credit requirement
    creditGap = minCredit - currCredit
    payAmount = PaymentAmount()
    paymentValue = ceil(creditGap / payAmount) * payAmount
}

// Create a payment for paymentValue
...

UpdateCredit(self, minCredit)

```

yondonfu:
This task will be composed of the following changes:

1. Track B's balance for a stream corresponding to a specific manifestID
2. Update `genPayment()` to compute B's credit gap based on its current balance and O's minimum credit (ticket EV)
3. Update `genPayment()` to use B's credit gap to create N tickets to be included in a payment

We can accomplish 1 by adding a `Balance` field to the `BroadcastSession` struct which is already associated with a specific manifestID and O. `genPayment()` can use this field since it already takes a `BroadcastSession` param. I'm leaning toward also returning the payment amount from `genPayment()` i.e.

```
func genPayment(sess *BroadcastSession) (string, *big.Rat, error)
```

The motivation behind returning the payment amount from `genPayment()` is that it can be used to compute B's updated balance after debiting the fee for results later on in `SubmitSegment()`. This updated balance can then be included in the return value for `SubmitSegment()` (would need to nest `net.TranscodeData` in a wrapper struct). Then, the caller of `SubmitSegment()` can use the updated balance when creating a new `BroadcastSession` to be appended to the session list. This follows the approach described in #961 which seeks to avoid mutating the `BroadcastSession` struct - so, instead of mutating the `Balance` field within `SubmitSegment()`, we can just compute the updated balance and include it in the return value for `SubmitSegment()`. We don't need to complete this described workflow for this task though - returning the payment amount in `genPayment()` will suffice.

We can accomplish 2 with something like this piece of pseudocode:

```
[...]

PR #903, opened 2019-05-31
net, server, core: Enable sending multiple PM tickets with a segment

### LINKED ISSUE #893: Enable sending of multiple PM tickets as a payment with each segment
B might need to send multiple tickets with each segment in order to meet the minimum credit requirement.

Ex: B's credit balance = 5 ETH, the minimum credit requirement = 10 ETH and O's ticket EV = 1 ETH. In this case, B would need to send 5 tickets to meet the minimum credit requirement

At the moment, B sends a single [Payment](https://github.com/livepeer/go-livepeer/blob/master/net/lp_rpc.proto#L209) with each segment by encoding the `Payment` message as a base64 encoded string and storing it in the "Livepeer-Payment" header of the segment message. Instead, B can encode an array of `Payment` messages as a base64 encoded string and store it in the "Livepeer-Payment" header of the segment message. O would also need to be updated to decode an array of `Payment` messages from the "Livepeer-Payment" header data.

kyriediculous:
This pull request enables sending multiple PM tickets as one `Payment` with each segment.
Currently still only one ticket is being wrapped, but the functionality to process multiple is there.

- changed `Payment` protobuf message to take in arrays of `Ticket`, `Sig` and `Seed` so that a **"Payment" can now consist out of multiple tickets**. [ not backwards compatible with Payment messages from older versions]  
- Adjusted `orchestrator.ProcessPayment` to batch process `Tickets`. It still takes in the `Payment` messages as argument, but that can now consist of multiple Tickets, Seeds and Signatures
- Adjusted helpers `GetPayment` and `GenPayment` to adhere to the edited `Payment` protobuf message
- Adjusted GetPaymentSender to retrieve the sender from the first ticket in the batch. This is not ideal but since we're currently only batching up 1 ticket for now, it will suffice. 
- Adjusted test files to test batch processing of `Tickets`
- Added helper to `orch_test.go` to have the ability to generate a payment with multiple tickets
- Added test for GetPaymentSender to be zero value if there's no tickets in a payment
- Added test for GetPayment to have zero values for the fields in payment if there's no tickets

Ran the test files seperately and `./test.sh` passes 
_Ubuntu 18.04 LTS Bionic Beaver_ 
_go1.12.2 linux/amd64_

fixes #893 

**Checklist:**
- [ ] Node runs in OSX and devenv _(I don't have OSX)_

** Extra / Questions ** 
* Might be forgetting edge cases, all feedback welcome
* There seems to be a styling issue in orch.ProcessPayment according to github, but not for me locally. Does the repo include a linter?


kyriediculous:
@yondonfu Should we add tests for the `GenPayment` helper analogous to `GetPayment` ?

core/orchestrator.go
```diff
@@ -100,27 +100,37 @@ func (orch *orchestrator) ProcessPayment(payment net.Payment, manifestID Manifes
...
-		RecipientRandHash: ethcommon.BytesToHash(payment.Ticket.RecipientRandHash),
-	}
-	seed := new(big.Int).SetBytes(payment.Seed)
+	for i, t := range payment.Tickets {
+		ticket := & pm.Ticket{
+			Recipient: 			ethcommon.BytesToAddress(t.Recipient),
+			Sender:				ethcommon.BytesToAddress(t.Sender),
+			FaceValue: 			new(big.Int).SetBytes(t.FaceValue),
+			WinProb:			new(big.Int).SetBytes(t.WinProb),
+			SenderNonce:		t.SenderNonce,
```
yondonfu: > There seems to be a styling issue in orch.ProcessPayment according to github, but not for me locally. Does the repo include a linter

We do use a linter [as a part of CI](https://github.com/livepeer/go-livepeer/blob/master/.circleci/config.yml#L37), but at the moment it is only enabled for the `pm` package because we only recently added the linter and there are a bunch of areas of the codebase that do not adhere to the standard linter rules. So, we only enabled the linter for new packages with the plan to gradually update the codebase to adhere to the standard linter rules.

Regarding the style issue here: do you have the Go plugin for VSCode installed? If you do, normally the plugin should auto format the file according to standard style rules upon saving.

core/orchestrator.go
```diff
@@ -100,27 +100,37 @@ func (orch *orchestrator) ProcessPayment(payment net.Payment, manifestID Manifes
...
+		}
+
+		if won {
+			glog.V(common.DEBUG).Info("Received winning ticket")
+			cachePMSessionID(orch.node, manifestID, sessionID)
+		}
 	}
 
 	return nil
+
```
yondonfu: Can kill this extra new line?

net/lp_rpc.pb.go
```diff
@@ -22,7 +43,7 @@ var _ = math.Inf
 // is compatible with the proto package it is being compiled against.
 // A compilation error at this line likely means your copy of the
 // proto package needs to be updated.
-const _ = proto.ProtoPackageIsVersion3 // please upgrade the proto package
+const _ = proto.ProtoPackageIsVersion2 // please upgrade the proto package
```
yondonfu: Hm looks like a lower proto package version was used to generate this file - can we maintain the same package version by using the version 3 for the locally installed package?

server/segment_rpc.go
```diff
@@ -169,11 +173,11 @@ func getPayment(header string) (net.Payment, error) {
 }
 
 func getPaymentSender(payment net.Payment) ethcommon.Address {
-	if payment.Ticket == nil || payment.Ticket.Sender == nil {
+	if len(payment.Tickets) == 0 || payment.Tickets[0] == nil || payment.Tickets[0].Sender == nil {
```
yondonfu: Let's add a test for the new case i.e. when `len(payment.Tickets) == 0`?

core/orchestrator.go
```diff
@@ -100,27 +100,37 @@ func (orch *orchestrator) ProcessPayment(payment net.Payment, manifestID Manifes
...
-	ticket := &pm.Ticket{
-		Recipient:         ethcommon.BytesToAddress(payment.Ticket.Recipient),
-		Sender:            ethcommon.BytesToAddress(payment.Ticket.Sender),
-		FaceValue:         new(big.Int).SetBytes(payment.Ticket.FaceValue),
-		WinProb:           new(big.Int).SetBytes(payment.Ticket.WinProb),
-		SenderNonce:       payment.Ticket.SenderNonce,
-		RecipientRandHash: ethcommon.BytesToHash(payment.Ticket.RecipientRandHash),
-	}
-	seed := new(big.Int).SetBytes(payment.Seed)
+	for i, t := range payment.Tickets {
```
yondonfu: Let's add a test for a payment with multiple tickets?

yondonfu (changes requested):
Left a few comments. Could we also exclude 6a64a1b from this PR since that commit is already included in https://github.com/livepeer/go-livepeer/pull/888?

yondonfu:
> Should we add tests for the GenPayment helper analogous to GetPayment ?

Since `genPayment()` just returns a payment with a single ticket and there is already a test that handles that case I think it is ok to leave it for now. Once we modify `genPayment()` to return a payment with multiple tickets we can add a test to handle the multiple ticket case too.

j0sh:
> changed Payment protobuf message to take in arrays of Ticket, Sig and Seed so that a "Payment" can now consist out of multiple tickets. [ not backwards compatible with Payment messages from older versions]

Just a quick note, because these types of encoding changes are hard to reverse once deployed. 

Most of the fields in `Ticket` will be redundantly specified, as well as the `Seed` from the Payment struct; these fields don't vary even if multiple tickets are sent.

From a quick glance at the protobuf definitions, the only fields that would be unique among tickets are the `sender_nonce` and the `sig` fields. So it'd probably make sense to parameterize just these two, and combine the relevant fields of `Ticket` and `Payment`, eg:

```protobuf
// come up with a better name here?
message PerPMParams {
  bytes sender_nonce = 1;
  bytes sig = 2;
}

// Be a little more specific that it's a payment using PM?
message PMPayment {
  // take most of the fields from Ticket here except for `sender_nonce`
  bytes recipient = 1;
  ...
  bytes seed;
  repeated PerPMParams params;
}
```

From there we can construct and validate each ticket as needed.

I suspect the intent behind the original construction was to have the wire protocol somewhat follow the on-chain representation, but that's not really required (and leads to a bloated representation on the wire when transmitting multiple tickets). If we really want to preserve this property anyway, maybe we can make a new message `PMPayments` with the `PMPayment{ticket, sig}`  as a repeated field, and move `seed` into the new `PMPayments` message.

yondonfu:
Good point @j0sh. I agree that we should avoid redundantly specified fields to minimize message bloat in the wire protocol.

> From a quick glance at the protobuf definitions, the only fields that would be unique among tickets are the sender_nonce and the sig fields

Right - since B will be using the same ticket params received from O for each of the tickets included in a payment sent to O the following fields should remain fixed for each ticket:

- `recipient`
- `sender`
- `face_value`
- `win_prob`
- `recipient_rand_hash`
- `aux_data` (included in the [round based ticket expiration PR](https://github.com/livepeer/go-livepeer/pull/890/files#diff-afa2cdc3d0285595d3ca53db8b21226fR208) which is currently under review)
- `seed`

What if we name the message containing `sender_nonce` and `sig` "TicketSenderParams"? These are the only parameters unique to each ticket provided by the sender in the PM protocol - at the moment, `sender` should just be B's address and fixed across all tickets (I wonder if there could be a use case for B using multiple addresses with multiple deposits/reserves in the future which *might* warrant having `sender` be a part of `TicketSenderParams` as well, but I don't think we need to accommodate for this scenario just yet) and the other fields of a ticket are provided by the recipient.

kyriediculous:
Good point @j0sh . 

Part of the reasoning was indeed to resemble the on-chain chain representation. 

Another one was that we didn't need to introduce an additional message definition, and therefore the required changes to work with the adjusted protobuf messages/definitions to accept multiple tickets were rather minimal as most function signatures remained the same, and just some of the logic needed to change. 

having `repeated sender_nonce` would also allow that field to benefit from the default packed encoding for scalar numeric types, further reducing payload size over the wire, even though by a tiny amount. 

> In proto3, repeated fields of scalar numeric types are packed by default. These function like repeated fields, but are encoded differently. A packed repeated field containing zero elements does not appear in the encoded message.

As you said though these changes are hard (or nearly impossible) to revert once deployed, so I think we should settle on some final definitions rather soon as it is a blocker for #897 _B generates a payment that will ensure that minimum credit requirement is met to include with segment_

Here is my suggestion. Since we already have a `TicketParams` message defined, why not just leverage that and change `Payment` to use that instead of `repeated Ticket`.

That way we do not need to introduce a new message of `Payments` which constitutes from a `Payment`.  

of course could still use a seperate message definition for the `TicketSenderParams` and put that in `Payment`, instead of defining the nonces, sigs and sender directly in `Payment`

```
// Payment for transcoding video segments
// A payment can constitute of multiple tickets
// A broadcaster might need to send multiple tickets to top up his credit with an Orchestrator
message Payment {
  // Probabilistic micropayment ticket parameters
  // These remain the same even when sending multiple tickets
  TicketParams ticket_params = 1;

  // ETH address of the sender
  bytes sender = 2;

  // Monotonically increasing counter that makes the ticket
  // unique relative to a particular hash commitment to a recipient's random number
  repeated uint32 sender_nonce = 3; // [encode=packed] by default

  // Sender signatures over the tickets
  repeated bytes sigs = 4;
}
```

We could then change `orchestrator.ProcessPayment` to something like (just typed out without checking if it would work) 

```
func (orch *orchestrator) ProcessPayment(payment *net.Payment, manifestID ManifestID) error {
	if orch.node == nil || orch.node.Recipient == nil {
		return nil
	}

	seed := new(big.Int).SetBytes(payment.TicketParams.seed)

	ticket := &pm.Ticket{
		Recipient:         ethcommon.BytesToAddress(payment.TicketParams.recipient),
		Sender:            ethcommon.BytesToAddress(payment.Sender),
		FaceValue:         new(big.Int).SetBytes(payment.TicketParams.FaceValue),
		WinProb:           new(big.Int).SetBytes(payment.TicketParams.WinProb),
		SenderNonce:       0,
		RecipientRandHash: ethcommon.BytesToHash(payment.TicketParams.RecipientRandHash),
	}

	for i, s := range payment.Sigs {

		ticket.SenderNonce = payment.SenderNonce[i]

		sessionID, won, err := orch.node.Recipient.ReceiveTicket(
			ticket,
			s,
			seed,
		)
		if err != nil {
			return errors.Wrapf(err, "Error receiving ticket %v for manifest %v", ticket, manifestID)
		}

		if won {
			glog.V(common.DEBUG).Info("Received winning ticket")
			cachePMSessionID(orch.node, manifestID, sessionID)
		}
	}

	return nil
}
```

altough I'm probably missing something on why this wouldn't work in practice.

j0sh:
> having repeated sender_nonce would also allow that field to benefit from the default packed encoding for scalar numeric types

I have a slight preference for keeping the sender_nonce / sigs grouped into a repeated `TicketSenderParams ` message for a couple reasons (a 'row major' rather than a 'column major' representation):

* Stronger structural invariants, eg we don't have to check the length of each field matches.
* Easier to add new per-ticket fields as necessary and maintain the above invariants. We won't know if such fields can take advantage of packed encoding  - `sigs` can't.

It might end up being a bit larger on the wire by a couple bytes per PM ticket, which seems OK given the gains in logical consistency and correctness. Feel free to benchmark though.

kyriediculous:
Good point again. 

What about the following definitions (I'm currently ommiting aux_data because I haven't looked into that PR yet): 

I'm also wondering if we even still need the `Ticket` message then. I don't think so?

```
message TicketAuxData {
  // Last initliazed round the ticket was created in
  uint32 creationRound = 1;
  // Eth Blockhash of the block in which the last round was initialized
  bytes  creationRoundBlockHash = 2;
}

// Required parameters for probabilistic micropayment tickets
message TicketParams {
    // ETH address of the recipient
    bytes recipient = 1;

    // Pay out (in Wei) to the recipient if the ticket wins
    bytes face_value = 2;

    // Probability that the ticket wins
    bytes win_prob = 3;

    // 32 byte keccak-256 hash commitment to a random number provided
    // by the recipient
    bytes recipient_rand_hash = 4;

    // Value generated by recipient that the recipient can use
    // to derive the random number corresponding to the recipient's hash commitment
    bytes seed = 5;

    // Auxilary data for a ticket used for validation
    TicketAuxData aux_data = 6;
}

// Sender Params per ticket
message TicketSenderParams {
    // Monotonically increasing counter that makes the ticket
    // unique relative to a particular hash commitment to a recipient's random number
    uint32 sender_nonce = 1;

    // Sender signature over the ticket
    bytes sig = 2;
}

// Payment for transcoding video segments
// A payment can constitute of multiple tickets
// A broadcaster might need to send multiple tickets to top up his credit with an Orchestrator
message Payment {
  // Probabilistic micropayment ticket parameters
  // These remain the same even when sending multiple tickets
  TicketParams ticket_params = 1;

  // ETH address of the sender
  bytes sender = 2;

  //Sender parameters (nonce and signature) for each ticket
  repeated TicketSenderParams ticket_sender_params = 3;
}

```

server/segment_rpc.go
```diff
@@ -393,24 +396,36 @@ func genPayment(sess *BroadcastSession) (string, error) {
...
+	 * See how many tickets to make according to needed credit and ticket EV
+	 * Invoke CreateTicket that many times and add to protoPayment
+	 */
 	ticket, seed, sig, err := sess.Sender.CreateTicket(sess.PMSessionID)
 	if err != nil {
 		return "", err
 	}
 
-	protoTicket := &net.Ticket{
+	// &net.TicketParams
```
yondonfu: Can we remove these lines?

core/orchestrator.go
```diff
@@ -102,26 +102,43 @@ func (orch *orchestrator) ProcessPayment(payment net.Payment, manifestID Manifes
...
-		glog.V(common.DEBUG).Info("Received winning ticket")
-		cachePMSessionID(orch.node, manifestID, sessionID)
+	seed := new(big.Int).SetBytes(payment.TicketParams.Seed)
+
+	ticket := &pm.Ticket{
+		Recipient:         ethcommon.BytesToAddress(payment.TicketParams.Recipient),
+		Sender:            ethcommon.BytesToAddress(payment.Sender),
+		FaceValue:         new(big.Int).SetBytes(payment.TicketParams.FaceValue),
+		WinProb:           new(big.Int).SetBytes(payment.TicketParams.WinProb),
+		SenderNonce:       0,
```
yondonfu: We can probably omit this line since it defaults to 0 anyway and by leaving it out it might be clearer that we are not explicitly setting the value at the moment (but we'll set the value when we range over `payment.TicketSenderParams`)

core/orchestrator.go
```diff
@@ -102,26 +102,43 @@ func (orch *orchestrator) ProcessPayment(payment net.Payment, manifestID Manifes
...
 	}
 
-	ticket := &pm.Ticket{
-		Recipient:         ethcommon.BytesToAddress(payment.Ticket.Recipient),
-		Sender:            ethcommon.BytesToAddress(payment.Ticket.Sender),
-		FaceValue:         new(big.Int).SetBytes(payment.Ticket.FaceValue),
-		WinProb:           new(big.Int).SetBytes(payment.Ticket.WinProb),
-		SenderNonce:       payment.Ticket.SenderNonce,
-		RecipientRandHash: ethcommon.BytesToHash(payment.Ticket.RecipientRandHash),
+	if payment.TicketParams == nil {
```
yondonfu, line 105: Test for this case?

core/orchestrator.go
```diff
@@ -102,26 +102,43 @@ func (orch *orchestrator) ProcessPayment(payment net.Payment, manifestID Manifes
...
+		return errors.Wrapf(nil, "Payment TicketParams is nil")
 	}
-	seed := new(big.Int).SetBytes(payment.Seed)
 
-	sessionID, won, err := orch.node.Recipient.ReceiveTicket(ticket, payment.Sig, seed)
-	if err != nil {
-		return errors.Wrapf(err, "error receiving ticket for payment %v for manifest %v", payment, manifestID)
+	if payment.Sender == nil {
```
yondonfu: Test for this case?

core/orchestrator.go
```diff
@@ -102,26 +102,43 @@ func (orch *orchestrator) ProcessPayment(payment net.Payment, manifestID Manifes
...
+	ticket := &pm.Ticket{
+		Recipient:         ethcommon.BytesToAddress(payment.TicketParams.Recipient),
+		Sender:            ethcommon.BytesToAddress(payment.Sender),
+		FaceValue:         new(big.Int).SetBytes(payment.TicketParams.FaceValue),
+		WinProb:           new(big.Int).SetBytes(payment.TicketParams.WinProb),
+		SenderNonce:       0,
+		RecipientRandHash: ethcommon.BytesToHash(payment.TicketParams.RecipientRandHash),
 	}
 
+	for _, tsp := range payment.TicketSenderParams {
```
yondonfu, line 123: Test for a payment with multiple tickets? We could add a test in `orch_test.go` that checks if `orch.node.Recipient.ReceiveTicket()` is called N times if there are N `TicketSenderParams` in the payment

core/orchestrator.go
```diff
@@ -102,26 +102,43 @@ func (orch *orchestrator) ProcessPayment(payment net.Payment, manifestID Manifes
...
+
+		ticket.SenderNonce = tsp.SenderNonce
+
+		sessionID, won, err := orch.node.Recipient.ReceiveTicket(
+			ticket,
+			tsp.Sig,
+			seed,
+		)
+		if err != nil {
+			return errors.Wrapf(err, "Error receiving ticket %v for manifest %v", ticket, manifestID)
```
yondonfu: If a single ticket is results in an error, the rest of the tickets would be skipped over. I think it would be preferable to log the error and continue with the rest of the tickets in this case. WDYT?

yondonfu (changes requested):

server/segment_rpc.go
kyriediculous: Absolutely, I forgot. My bad.

core/orchestrator.go
kyriediculous: Yes, you're probably right. Done.

core/orchestrator.go
kyriediculous: Was already there with bad (the original) naming, but just passed because I used `errors.Wrapf` in the wrong way. Apparently when passing `nil` as the first argument, errors.Wrapf will not return an error. Changed that to take in a new error instead.

core/orchestrator.go
kyriediculous: Done

core/orchestrator.go
kyriediculous: Definitely ! Thanks for the thorough review. Learning a lot.

core/orchestrator.go
kyriediculous: Added according to the docs you sent me.

core/orchestrator.go
```diff
@@ -98,30 +98,48 @@ func (orch *orchestrator) TranscoderResults(tcId int64, res *RemoteTranscoderRes
 }
 
 func (orch *orchestrator) ProcessPayment(payment net.Payment, manifestID ManifestID) error {
+
 	if orch.node == nil || orch.node.Recipient == nil {
 		return nil
```
kyriediculous, line 102: @yondonfu Should we return an error here instead of nil or not (analogous to the two checks below) ?

core/orch_test.go
```diff
@@ -485,29 +486,68 @@ func TestOrchCheckCapacity(t *testing.T) {
...
+	recipient.On("ReceiveTicket", mock.Anything, mock.Anything, mock.Anything).Return(sessionID, true, errors.New("mock error"))
 
-	err := orch.ProcessPayment(defaultPayment(t), ManifestID("some manifest"))
+	err := orch.ProcessPayment(defaultPayment(t), manifestID)
+	errorLogsAfter := glog.Stats.Error.Lines()
 
 	assert := assert.New(t)
-	assert.NotNil(err)
+	assert.Nil(err)
+	assert.NotZero(errorLogsAfter)
```
yondonfu: If `glog` logged > 0 errors prior to this test (i.e. due to errors logged during other tests in this file), I believe this assertion would always pass? I think `assert.NotZero(errorLogsAfter-errorsLogsBefore)` would be more robust here.

core/orchestrator.go
yondonfu: We should return `nil` here because the orchestrator could be running in off-chain mode (i.e. broadcaster is directly connecting to an orchestrator without any interactions with the blockchain). `orch.node.Recipient = nil` when the orchestrator is running in off-chain mode and we wouldn't want to return an error in this scenario - returning nil allows the orchestrator to proceed with the transcoding by ignoring the PM related logic.

core/orchestrator.go
```diff
@@ -102,26 +102,43 @@ func (orch *orchestrator) ProcessPayment(payment net.Payment, manifestID Manifes
...
 
-	ticket := &pm.Ticket{
-		Recipient:         ethcommon.BytesToAddress(payment.Ticket.Recipient),
-		Sender:            ethcommon.BytesToAddress(payment.Ticket.Sender),
-		FaceValue:         new(big.Int).SetBytes(payment.Ticket.FaceValue),
-		WinProb:           new(big.Int).SetBytes(payment.Ticket.WinProb),
-		SenderNonce:       payment.Ticket.SenderNonce,
-		RecipientRandHash: ethcommon.BytesToHash(payment.Ticket.RecipientRandHash),
+	if payment.TicketParams == nil {
+		return errors.Wrapf(errors.New("No TicketParams in payment"), "Could not find TicketParams for payment %v", payment)
```
yondonfu: `errors.Wrapf` is typically used when:

- Function A calls function B calls function C
- C returns an error to B
- B wraps C's error with its own error
- A receives the wrapped error which includes context from both function B and C

We're not receiving an error from a function call here so we can just return a single error instead of wrapping an error with another error.

core/orchestrator.go
```diff
@@ -102,26 +102,43 @@ func (orch *orchestrator) ProcessPayment(payment net.Payment, manifestID Manifes
...
 	}
-	seed := new(big.Int).SetBytes(payment.Seed)
 
-	sessionID, won, err := orch.node.Recipient.ReceiveTicket(ticket, payment.Sig, seed)
-	if err != nil {
-		return errors.Wrapf(err, "error receiving ticket for payment %v for manifest %v", payment, manifestID)
+	if payment.Sender == nil || len(payment.Sender) == 0 {
+		return errors.Wrapf(errors.New("No sender in payment"), "Payment %v does not contain a sender", payment)
```
yondonfu: Same comment as above re: `errors.Wrapf` usage.

core/orchestrator.go
```diff
@@ -102,26 +102,43 @@ func (orch *orchestrator) ProcessPayment(payment net.Payment, manifestID Manifes
...
+
+		ticket.SenderNonce = tsp.SenderNonce
+
+		sessionID, won, err := orch.node.Recipient.ReceiveTicket(
+			ticket,
+			tsp.Sig,
+			seed,
+		)
+		if err != nil {
+			err = errors.Wrapf(err, "Error receiving ticket %v for manifest %v", ticket, manifestID)
```
yondonfu: Since we are just logging `err` here, we can log the error as-is instead of wrapping it with another error message since we are not propagating it to a caller. `errors.Wrapf` is primarily useful for annotating an error so that you can maintain an error stack trace as an error is propagated up the caller stack.

yondonfu (changes requested):

core/orch_test.go
kyriediculous: Wouldn't the state of `glog.Stats` be wiped out for every test considering they are unit tests? Regardless, I'll follow your advice and adjust to use `after-before`

core/orchestrator.go
kyriediculous: Considering that `ReceiveTicket` errors don't return useful information for which ticket errored in a multi-ticket setup, it might be useful to at least add that information to the error message through formatting or wrapping? 

eg. 

`glog.Errorf("Error receiving ticket %v for manifest %v: %v\n", ticket, manifestID, err)`

running into an invalid `WinProb` would log 

`Error receiving ticket <ticket> for manifest <manifest>: invalid ticket winProb <winProb>`

yondonfu: Yeah we could include the extra information in the logged error message as you describe in the first `glog` code snippet you listed.

I slightly misspoke in my original comment - I intended to say we don't need to use `errors.Wrapf` here to wrap `err` i.e. we don't need L133 which modifies `err` with `errors.Wrapf`. However, I *do* think it is useful to add extra contextual information when logging `err` with `glog` via a formatted error message.

core/orch_test.go
yondonfu: I don't think `glog.Stats` is reset. I just tried out this code snippet to confirm that the line count is preserved:

```
package logger

import "github.com/golang/glog"

func LogError(msg string) {
	glog.Errorf("ERROR: %v", msg)
}
```

```
package logger

import (
	"fmt"
	"testing"

	"github.com/golang/glog"
)

func TestLogError1(t *testing.T) {
	LogError("foo")
	LogError("foo")

	logCount := glog.Stats.Error.Lines()

	fmt.Printf("%v\n", logCount)
}

func TestLogError2(t *testing.T) {
	logCount := glog.Stats.Error.Lines()

	fmt.Printf("%v\n", logCount)
}
```

`TestLogError1` should print out 2 and `TestLogError2` should also print out 2.

yondonfu (approved):
LGTM 🚢

Issue #875, opened 2019-05-17
pm: Fetch OrchestratorInfo updates for a broadcast session in response


yondonfu:
When a broadcaster creates a broadcast session for a particular orchestrator, it should subscribe to OrchestratorInfo updates from the orchestrator so that the broadcaster can be notified whenever the orchestrator updates the required ticket params (when we implement payment accounting this subscription can also be used to receive notifications of updated price).

The subscription would be implemented as a service side streaming RPC call

```
service Orchestrator {
    ...

    // Called by the broadcaster to receive streaming orchestrator info updates
    rpc SubscribeOrchestratorUpdates(OrchestratorRequest) returns (stream OrchestratorInfo)
}
```

```
func (h *lphttp) SubscribeOrchestratorUpdates(..., stream net.Orchestrator_SubscribeOrchestratorUpdatesServer) error {
    // gasPriceNotifier is a global instance of an object that continuous watches for network
    // gas price updates. The object can be used to subscribe to updates via a channel
    sub := h.gasPriceNotifier.Subscribe()

    defer sub.Cancel()
    
    for {
        case gasPrice := <-sub.Updates():
            // Update ticket params
            ticketParams := TicketParams(gasPrice)
            // With payment accounting, update price if needed as well
            info := &net.OrchestratorInfo{
                ...
                ticketParams,
            }

            if err := stream.Send(info); err != nil {
                return err
            }
        
        // Case to detect server shut down
    }
}
```

j0sh:
Subscribing via a separate RPC call is one way to do it.

Another is to return the updated pricing info as part of a segment submission request - submit a segment, receive an error response indicating a price update, and go from there. That makes use of the mechanisms we already have in place, rather than adding another RPC call which expands the surface area of the networking protocol.

yondonfu:
Yeah I see the benefit in minimizing networking surface area. If O responds with an `OrchestratorInfo` message (or whatever message includes ticket params + pricing) in the error I guess B could check the HTTP status code (`http.StatusPaymentRequired`?) to determine whether it needs to parse the error for the ticket params + pricing and any other relevant orchestrator info?

j0sh:
> I guess B could check the HTTP status code (http.StatusPaymentRequired?) to determine whether it needs to parse the error for the ticket params + pricing

We could also put in additional structured error types to the protobuf `TranscodeResult` 

[...]

Issue #829, opened 2019-04-16
pm: Implement orchestrator strategy for winning ticket redemption


yondonfu:
Without taking into account O's strategy for generating ticket parameters, O will have to redeem a winning ticket when we are within X blocks of the ticket's expiration.

Taking into account O's strategy for generating ticket parameters, if O is setting the ticket `faceValue` to B's `maxFloat` (or `maxFloat / 2`), then O will have to redeem a winning ticket as soon as it receives it if O wants to ensure that the `faceValue` of the ticket is guaranteed by O's reserve and if O wants to accept tickets again from B ASAP (though O could use B's current deposit to determine whether to keep accepting tickets past B's `maxFloat` for a brief period of time i.e. the time it takes for a winning ticket redemption to confirm on-chain). I think this would basically eliminate O's ability to submit batch ticket redemptions since O is always redeeming tickets as soon as they are received.

yondonfu:
Since the #828 will have an orchestrator either use a default faceValue or a faceValue that matches a broadcaster's maxFloat (if default faceValue > maxFloat) there are two cases to handle:

1) If the orchestrator is using a default faceValue that is at least x2 the maxFloat, then the orchestrator can hold on to winning tickets until it has received maxFloat worth of winning tickets
2) If the orchestrator is matches faceValue to maxFloat, then the orchestrator needs to redeem the winning ticket when it receives it

For simplicity we can handle 2 in a first pass and come back to handle 1

Issue #866, opened 2019-05-16
Payment Accounting v0


yondonfu:
# Payment Accounting v0

## Messages

Introduce a new `PriceInfo` message.

```
message PriceInfo {

    // Price in wei
    int64 pricePerUnit = 1;
    
    // Number of pixels that the price covers
    int64 pixelsPerUnit = 2;
}
```

`PriceInfo` specifies the price for N pixels. By setting the price for a variable number of pixels, we maximize pricing flexibility. If an O wants to specify a price per pixel less than 1 wei then it can set `price` to 1 wei and set `pixels` > 1.

Add `price_info` field to the existing `OrchestratorInfo` message.

```
message OrchestratorInfo {
    
    // URI of the transcoder to use for submitting segments
    string transcoder = 1;
    
    // Parameters for probabilistic micropayment tickets
    TicketParams ticket_params = 2;
    
    // Price info
    PriceInfo price_info = 3;
    
    // Orchestrator returns info about own input object storage, if it wants it to be used
    repeated OSInfo storage = 32;
}
```

Introduce a new `TranscodeReceipt` message.

```
message TranscodeReceipt {
    
    // Number of pixels processed to produce the transcoded data
    int64 pixels_processed = 1;
    
    // Price info used during transcoding
    PriceInfo price_info = 2;
}
```

In the short term, `pixels_processed` will just track output pixels, but in the future it could track input pixels as well.

Add a `receipt` field to the existing `TranscodedSegmentData` message.

```
message TranscodedSegmentData {

    // URL where the transcoded data can be downloaded from.
    string url = 1;
    
    // Transcoding receipt
    TranscodeReceipt receipt = 2;
}
```

## Errors

`ERROR_INSUFFICIENT_CREDIT` indicates that B does not have sufficient credit for O to process the next segment. The most likely reason for this error is that O's price has changed and B generated its latest payment based on an outdated view of O's price. O could include its latest price in the error message. One option would be to return the `OrchestratorInfo` message with the updated price.

## Helpers

Credit balances can be tracked locally on a per stream basis with relevant state flushed when the session associated with a stream times out (approx. 1 minute without receiving a segment at the moment). Both the broadcaster and orchestrator would keep track of their own view of the credit relationship.

**CurrCredit(addr)**

Should return the current credit balance for an ETH address.

**UpdateCredit(addr, updatedCredit)**

Should update the current credit balance for an ETH address.

**MinCredit()**

[...]

PR #904, opened 2019-06-01
Dynamic ticket param generation

### LINKED ISSUE #828: pm: Implement orchestrator strategy for generating ticket parameters

### LINKED ISSUE #830: pm: Implement orchestrator strategy for detecting invalid tickets and responding with error mesages
Upon receiving a ticket, an orchestrator should use the ticket parameter generation algorithm described in https://github.com/livepeer/go-livepeer/issues/828#issuecomment-493498594 in order to determine the required ticket parameters. The orchestrator should compare the parameters of the ticket received with the required parameters. If the parameters are different, then the orchestrator should return an error indicating invalid parameters. 

Note: Tackling this issue can just involve implementing the check and returning an error message. #875 will handle returning updated orchestrator information to notify a broadcaster of new ticket parameters

yondonfu:

This PR implements dynamic ticket parameter generation for an orchestrator which is also used by the orchestrator to check for a valid faceValue and winProb when receiving a ticket.

- The orchestrator will use a `GasPriceMonitor` to poll for gas price updates and cache the latest fetched gas price in memory. Fetching the gas price from an in-memory cache is preferable to sending a network request for an updated gas price since the orchestrator will need to use the gas price for every ticket that it receives. `GasPriceMonitor` uses a `GasPriceOracle` interface which is currently implemented by the `ethclient` package from `go-ethereum` which uses the `eth_gasPrice` JSON-RPC call to fetch a suggested gas price from an ETH node. We can consider other implementations as well such as using the ethgasstation API and picking a safe low, standard or fast gas price (there could even be an option exposed to users to choose a preferred transaction confirmation speed).
- The orchestrator will use a `FloatMonitor` to cache the claimable reserve for a broadcaster which is used to compute the current max float (i.e. the highest possible faceValue to use) for a broadcaster. The `FloatMonitor` also exposes methods for updating a broadcaster's current max float which are meant to be used during winning ticket redemption - an orchestrator would decrease the broadcaster's max float by the ticket faceValue after submitting the redemption transaction and increase the broadcaster's max float by the ticket faceValue after the redemption transaction confirms. However, this feature is not implemented in this PR and will be implemented in the PR for #829. The `FloatMonitor` also runs a cleanup worker in the background to clear the max float cache if a max float value has not been used in awhile
- Updated the TicketBroker contract binding based on livepeer/protocol@57f400a9175a6ebdc1d46e14e98586c9a32cfa8e so the orchestrator can ask the contract for a broadcaster's claimable reserve
- Implemented the ticket parameter generation strategy described in [this comment](https://github.com/livepeer/go-livepeer/issues/828#issuecomment-493498594). The one addition in this PR is that the orchestrator will set the faceValue to the EV if the default faceValue computed from the current redemption transaction cost < the EV. See the in-line code docs for more information on potential future changes here as well.
- The orchestrator runs the ticket param generation computation when it returns a `OrchestratorInfo` message and when it receives a ticket. If the faceValue and winProb included in a ticket does not match the values produced by the ticket param generation computation, then the orchestrator will return an error. A PR for #875 will allow the orchestrator to return updated params instead.
- An orchestrator node can now be started with an optional `-ticketEV` flag (by default it is set to 1 GWEI)

Wrote additional unit tests and did some manual testing.

Fixes #828 
Fixes #830

**Checklist:**


yondonfu:
Ready for review!

eth/gaspricemonitor_test.go
```diff
@@ -0,0 +1,177 @@
...
+
+func (s *stubGasPriceOracle) SetErr(err error) {
+	s.err = err
+}
+
+func (s *stubGasPriceOracle) SuggestGasPrice(ctx context.Context) (*big.Int, error) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	if s.err != nil {
```
angyangie: In what scenarios would there be an `s.err` here?

eth/gaspricemonitor.go
```diff
@@ -0,0 +1,116 @@
...
+}
+
+// Start starts polling for gas price updates
+func (gpm *GasPriceMonitor) Start(ctx context.Context) error {
+	if gpm.polling {
+		return errors.New("already polling")
+	}
+
+	// Initialize gasPrice before starting to poll
+	if err := gpm.fetchAndUpdateGasPrice(ctx); err != nil {
```
angyangie, line 58: In what scenarios would there be an `err` here?

pm/floatmonitor.go
```diff
@@ -0,0 +1,132 @@
...
+		if err != nil {
+			return nil, err
+		}
+
+		fm.maxFloats[addr] = &item{
+			val: maxFloat,
+		}
+	}
+
+	fm.maxFloats[addr].lastAccess = unixNow()
```
angyangie, line 118: Could we run into concurrency issues here without the use of a lock?

cmd/livepeer/livepeer.go
```diff
@@ -44,6 +44,17 @@ import (
 var (
 	ErrKeygen    = errors.New("ErrKeygen")
 	EthTxTimeout = 600 * time.Second
+
+	// redeemGas is the gas required to redeem a PM ticket
+	redeemGas = 100000
+	// txCostMultiplier is the multipler on the transaction cost to use for PM ticket faceValue
```
angyangie, line 50: Why do we need a multiplier here?

pm/recipient.go
```diff
@@ -134,20 +165,80 @@ func (r *recipient) RedeemWinningTickets(sessionIDs []string) error {
...
+	txCost := new(big.Int).Mul(big.NewInt(int64(r.cfg.RedeemGas)), gasPrice)
+
+	// faceValue = txCost * txCostMultiplier
+	faceValue := new(big.Int).Mul(txCost, big.NewInt(int64(r.cfg.TxCostMultiplier)))
+
+	// TODO: Consider setting faceValue to some value higher than
+	// EV in this case where the default faceValue < the desired EV.
+	// At the moment, for simplicity we just adjust faceValue to the
+	// desired EV in this case (which would result in winProb = 100%).
+	// In practice, EV should be smaller than the default faceValue
```
angyangie, line 201: Could you help me understand why EV should be smaller then the default `faceValue`? Perhaps point me to some documentation that might be helpful?

pm/floatmonitor.go
kyriediculous: We're using a lock in the exported wrapper function (`MaxFloat`) and defering unlocking it until `maxFloat` finishes (the helper). 

This is suboptimal however since we're locking the entire map and not the key/value pair for a certain address.

eth/gaspricemonitor.go
kyriediculous: When the GasPriceOracle is a service that implements network requests. 

Eg.: an ethereum node or ethgasstation

cmd/livepeer/livepeer.go
kyriediculous: In the original outline there is mention of a transaction cost overhead, represented by a float. This represents the overhead cost for redeeming a ticket. 

For reasons such as rounding errors with floats and having to use the big.Float library (for more details you should ask @yondonfu ) this is now represented by a proportion between two integers (`redeemGas` and `txCostMultiplier`)

redeemGas is currently at `100 000`
txCostMultiplier is `100`

This would mean a transaction cost overhead of 1% 

If we look at the `faceValue` method in `recipient.go` we'll see: 

```
gasPrice := r.gpm.GasPrice()
	// txCost = redeemGas * gasPrice
	txCost := new(big.Int).Mul(big.NewInt(int64(r.cfg.RedeemGas)), gasPrice)

	// faceValue = txCost * txCostMultiplier
	faceValue := new(big.Int).Mul(txCost, big.NewInt(int64(r.cfg.TxCostMultiplier)))
```

So let's say we have a gasPrice of 40 wei (not realistic but just for exampel sake), we'd get a transaction cost of `4 million wei` 

This would put the faceValue at 100x that , or `400 million wei` 

Which remains a 1% transaction cost overhead.

tl;dr redeemGas with txMultiplier replaces a float representation for transaction cost overhead.

pm/recipient.go
kyriediculous: Essentially `EV = faceValue * winProb`

So a ticket with a faceValue of `500 wei` and a winProb of `10%` would have an expected payout per ticket of `50 wei` 

Now let's say our EV is 500 wei and our defaultFaceValue is 250 wei we'd get 
`500 = 250 * winProb`
which results in a winProbability of 200%. 

That's of course not possible which is why if `defaultFaceValue < EV` we adjust the faceValue to the EV. Resulting in a winning probability of 100%, which is still not optimal but from talking to @yondonfu he considers that an edge case currently.

I'm probably missing quite a bit of essential context here that @yondonfu could provide better than me I think. But I think it essentially boils down to:

> _No expectation of long lived relationships with some minimum value transacted:_ An orchestrator is not affected by broadcasters that send a few non-winning tickets and then disconnect because it knows that it will eventually receive a sufficient number of tickets from the numerous broadcasters it works for such that it will be paid fairly. Furthermore, since orchestrators set the face value of tickets, they can always ensure that it will be profitable to redeem received winning tickets

From the [livepeer blog on PM](https://medium.com/livepeer-blog/streamflow-probabilistic-micropayments-f3a647672462)

eth/gaspricemonitor.go
angyangie: Ah I see. So that part is something that will be built out later?

cmd/livepeer/livepeer.go
angyangie: Thanks for the really detailed explanation @kyriediculous !! Makes sense to me

pm/recipient.go
angyangie: Again, makes a lot of sense! Thanks @kyriediculous. Basically, the only time winning probability could be 100% is when an O doesn't understand the concept of PM? No one would rationally choose to have every ticket be a winning ticket, which would almost always mean a higher transaction cost than ticket value

eth/gaspricemonitor.go
yondonfu: At the moment, the `GasPriceOracle` interface is implemented by the `ethclient` from go-ethereum which [implements SuggestGasPrice by querying an ETH node for a gas price](https://github.com/livepeer/go-livepeer/blob/master/vendor/github.com/ethereum/go-ethereum/ethclient/ethclient.go#L476). 

Under the hood this implementation uses the [eth_gasPrice JSON-RPC request](https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_gasprice) which is implemented by ETH nodes. The algorithm used by ETH nodes to compute a suggested gas price can vary. geth will return a gas price [based on the content of recent blocks](https://github.com/ethereum/go-ethereum/blob/master/eth/gasprice/gasprice.go#L40). 

This `GasPriceOracle` implementation gets things working for now, but the interface gives us flexibility to implement more optimal gas price fetching mechanisms in the future (i.e. using ethgasstation or a custom service).

angyangie: Super cool! The `go-ethereum` links are great in helping me understand how we're fetching this information. Good thinking on the interface. Thanks @yondonfu

pm/recipient.go
kyriediculous: Still trying to wrap my mind around it sometimes as well. But yes that would mean an Orchestrator doesn't really understand the concept. I think most Orchestrators would be using the default value anyway and changing defaults would be for advanced users only. 

There wouldn't directly be a higher tx cost than ticket value, but high winning probabilities could add to network congestion leads to having to use a higher gasPrice in order to keep the same confirmation time for transactions. 

A higher winning probability would also require the orchestrator to redeem his tickets more frequently to increase the chance of actually getting paid in full based on the outstanding float from a broadcaster to an orchestrator. 

Eg. 
Round 1: 10 O's
B's Reserve: 5 eth (0.5 eth per Orch allocated) 
B's Deposit: 1 eth

- In Round 1 O receives a winning ticket (100% winProb) for 1.5 eth but he waits to redeem it until the next round because the gasPrice is currently high

Now let's say in the next round there are 20 O's , then the allocation for each O from B would drop to 0.25eth (for simplicity sake, no one has redeemed a ticket from B in the meantime, so deposit remains at 1 eth) 

B's deposit + allocation from B to O = 1.25 eth now, which is not enough to cover the full payment of O's winning ticket of 1.5eth sent in the previous round. 

From the [PM spec](https://github.com/livepeer/wiki/blob/master/spec/streamflow/pm.md#reserve): 

> An orchestrator can accept or reject work from a broadcaster based upon the broadcaster's reserve and the value of the allocation from the reserve committed to the orchestrator. For example, since the value of the allocation affects the maximum face value that can be used for tickets, lower reserves would require lower ticket face values and higher winning probabilities - if the winning probability is too high such that an orchestrator would need to redeem winning tickets too frequently (potentially contributing to network congestion), then an orchestrator might reject work from the broadaster.

yondonfu: In the context of payment accounting, an orchestrator's EV is also the minimum credit requirement imposed on a broadcaster (as described [here](https://github.com/livepeer/go-livepeer/issues/866)) - given an orchestrator EV of X, an orchestrator will only transcode segments if the broadcaster's current credit balance is at least X. 

The minimum credit requirement feature allows an orchestrator to control how much value it is guaranteed when it transcodes segments. A higher EV = a higher minimum credit requirement which is "safer" for an orchestrator (in case transcoding a segment is costly). However, at the same time, a higher EV is more "risky" for a broadcaster because if a broadcaster has to send X value to an orchestrator up front, the broadcaster stands to lose X if the orchestrator does not respond. So, in practice, the EV should be high enough to cover some reasonable amount of transcoding work for the orchestrator and low enough so the broadcaster doesn't put a lot of value at risk.

If the EV is higher than the default faceValue (which by default is 100x the current transaction cost of ticket redemption), then that would represent a pretty high minimum credit requirement:

Ex. 

```
redeemGas = 100000
gasPrice = 5 GWEI
txCost = 0.0005 ETH
faceValue = 100 * txCost = .05 ETH -> $12.5 @250 ETH/USD
```

So, if the EV is equal to the default faceValue here, then the minimum credit requirement would be $12.5 and the broadcaster would be putting $12.5 at risk up front which is pretty unreasonable and shouldn't happen in practice as long as the broadcaster software is configured properly to reject orchestrators that set an absurdly high EV. In practice, the EV should be on the magnitude of cents or fractions of cents. Consider the following purely illustrative example using per segment pricing (we're planning on using per pixel pricing instead), AWS Elastic Transcoder pricing results in something like $.003 per 2 second segment. So, a reasonable EV/minimum credit requirement covering 3 segments would be 3 x $.003 = $.009 -> .9 cents. I think a reasonable EV will generally be closer to this range of values than it would be to the `faceValue`. Higher EV values that approach the default faceValue would be more likely to be rejected by broadcasters. Thus, the case where EV > default faceValue feels like an edge case that we can handle in a simplistic manner for this PR and consider more optimal logic later (i.e. if the EV > default faceValue, set a new default faceValue that is some multiple of the EV).

kyriediculous (approved):
:+1:

pm/floatmonitor.go
yondonfu: While I initially considered a key/value pair lock, I just remembered that even with a key/value pair lock, the global map lock is unavoidable if you want to eliminate data races because the underlying implementation for a map can involve internal data updates (i.e. rebalancing to maintain even value distribution across buckets to ensure O(1) lookups in a hash map implementation - see: https://dave.cheney.net/2015/12/07/are-go-maps-sensitive-to-data-races) that impact the entirety of the map when updating a single key.

The one improvement that could be considered here is to use sync.RWMutex instead of sync.Mutex because the former will allow for multiple non-blocking concurrent readers - in *most* cases, `FloatMonitor` is probably going to be used for reading values and writing values will only occur when a winning ticket is being redeemed so it might benefit from using a sync.RWMutex instead of a sync.Mutex.

yondonfu: Now using an [sync.RWMutex](https://github.com/livepeer/go-livepeer/pull/904/files#diff-2bfb8ebb26cc77da0bad94931ddc85f0R38).

Issue #634, opened 2018-12-14
pm: responses to acceptable and unacceptable invalid tickets and 


yondonfu:
**Is your feature request related to a problem? Please describe.**

There is a distinction between "acceptable invalid tickets" and "unacceptable invalid tickets" using the following criteria: could the invalid data in the ticket be a result of information that B has not yet received?

O needs to be able to differentiate between acceptable and unacceptable invalid tickets using the above criteria. 

O also should provide a grace period for acceptable invalid tickets for the sake of quality of service.

**Describe the solution you'd like**

Acceptable invalid tickets:

- Invalid faceValue and/or winProb since these values might have been acceptable at one point, but perhaps O now no longer accepts these values and B does not know about the new values yet
- Invalid recipientRand + randSeed since B might not have received the new pair after sending out a winning ticket yet due to async messaging

Unacceptable invalid tickets:

- Invalid senderNonce i.e. a replay
- Invalid randSeed s.t. `H(randSeed, O.randSecret) != T.recipientRandHash`
- Other invalid fields in the ticket. Ex. invalid recipient address (i.e. not O's address), invalid timestamp (or whatever time indicator is being used), invalid signature from B, etc.

Furthermore, O should respond different to these two types of invalid tickets.

- When O receives an unacceptable invalid ticket, O should not process the segment from B, respond with `ErrUnacceptableInvalidTicket` and disconnect.
- When O receives an acceptable invalid ticket, O will respond with `ErrAcceptableInvalidTicket`, provide B with a grace period and after this grace period, O will not process the next segment from B and disconnect. See below for grace period details.

We can implement the grace period by having O maintaining a counter `acceptableInvalidTickets` for each B that is incremented every time O receives an acceptable invalid ticket from B. This counter is reset every time O receives a winning ticket from B. O also defines a threshold `maxAcceptableInvalidTickets`. If `acceptableInvalidTickets > maxAcceptableInvalidTickets`, the grace period ends.

**Describe alternatives you've considered**

- O allows B to send invalidTicketsInARowAllowed invalid tickets in a row before disconnecting, to allow B time to receive and use the new recipientRand value (default setting invalidTicketsInARowAllowed=3).
- O maintains (on disk) an ongoing invalid tickets ratio per B as well as a invalidTicketRatioThreshold above which O rejects further video/tickets from B (invalidTicketRatioThreshold should be at least invalidTicketsInARowAllowed/ticketWinningProbability)

[...]

PR #905, opened 2019-06-01
Update OrchestratorInfo for acceptable payment errors

### LINKED ISSUE #875: pm: Fetch OrchestratorInfo updates for a broadcast session in response

yondonfu:

This PR adds support for an orchestrator returning updated OrchestratorInfo messages with a TranscodeResult message. At the moment, the primary purpose of returning updated OrchestratorInfo message is to inform a broadcaster of updated ticket parameters. The OrchestratorInfo message can also be used to notify of other [updated orchestrator information](https://github.com/livepeer/go-livepeer/issues/875#issuecomment-493597221).

- Add the `OrchestratorInfo` message to the `TranscodeResult` message. We use the field number 16 for the `OrchestratorInfo` message to reserve smaller field numbers for future information that might be used more often.
- If the orchestrator encounters an "acceptable payment error" it will return an updated OrchestratorInfo message with the TranscodeResult message instead of returning an error. We currently consider invalid ticket faceValue and winProb errors as acceptable payment errors. The orchestrator should only accept a fixed number of these types of errors - the implementation of such as grace period is left for a separate PR targeting #634.
- If a broadcaster receives an updated OrchestratorInfo message it will update its current broadcast session with it so that the next segment submitted using the broadcast session will use the updated orchestrator information

Added unit tests.

Fixes #875 

**Checklist:**


yondonfu:
Ready for review!

server/segment_rpc.go
```diff
@@ -58,10 +58,24 @@ func (h *lphttp) ServeSegment(w http.ResponseWriter, r *http.Request) {
...
 		glog.Errorf("Error processing payment: %v", err)
-		http.Error(w, err.Error(), http.StatusPaymentRequired)
-		return
+
+		if !acceptablePaymentError(err) {
+			http.Error(w, err.Error(), http.StatusPaymentRequired)
+			return
+		}
+
+		oInfo, err = orchestratorInfo(orch, getPaymentSender(payment), orch.ServiceURI().String(), segData.ManifestID)
```
angyangie: Does this still preserve the `core.RandomManifestID()` external storage path prefix used to create unique OS sessions per Orchestrator (to prevent segment front-running)?

server/segment_rpc.go
```diff
@@ -330,6 +332,11 @@ func SubmitSegment(sess *BroadcastSession, seg *stream.HLSSegment, nonce uint64)
 		return nil, err
 	}
 
+	// update OrchestratorInfo if necessary
+	if tr.Info != nil {
```
angyangie, line 335: Will this field only be populated when OrchestratorInfo needs to be updated?

yondonfu: Yep. So, under normal circumstances `tr.Info` will be nil and it will only be non-nil when updated information needs to be returned. This is also why `OrchestratorInfo` is assigned a higher field number > 15 within the `TranscodeResult` message since [field numbers 1-15 generally should be reserved for very frequently occurring message elements](https://developers.google.com/protocol-buffers/docs/proto#assigning-field-numbers)

server/segment_rpc.go
yondonfu: Hm upon further inspection it seems that the current orchestrator behavior is to generate a fresh random manifestID as the storage prefix for its OS session every time a broadcaster calls `GetOrchestrator()` on the orchestrator. So, in order to preserve this behavior, we could also generate a fresh random manifestID here instead of using `segData.ManifestID` which is the broadcaster supplied manifestID. However, this would mean that the orchestrator OS storage prefix can change mid-stream every time the orchestrator sends an updated `OrchestratorInfo` message - not sure if this has any negative implications, but it seems like it might be ok

yondonfu: Updated to always [generate a random manifestID for the orchestrator OS session](https://github.com/livepeer/go-livepeer/pull/905/files#diff-44bb03d888e11d6429223a7b0e55220dR220).

server/segment_rpc.go
angyangie: That makes sense. Great attention to detail with `OrchestratorInfo` being higher than 15 here

server/segment_rpc.go
angyangie: I can't think of a reason why generating a new OS session mid-stream might cause issues either. That looks like a good solution to me!

angyangie (approved):

PR #916, opened 2019-06-13
cmd, core, pm, server: O creates Price info for B based on txCost for payment redemption

### LINKED ISSUE #894: O creates price info for B based on a base price and payment redemption tx cost overhead
**TxCostOverhead(addr)**

Should return the overhead (as a float) of payment redemption based on the transaction cost relative to the value of the payment. The overhead will be based on the current gas price. Additionally, when using PM, the overhead will be based on the `faceValue` used for tickets. Since the `faceValue` used for tickets can depend on a broadcaster's reserve, the value returned by this function can vary depending on the broadcaster which is why the function accepts the broadcaster's address as a paramter.

**PriceInfo(addr)**

Should return the price info for a broadcaster. An orchestrator will call this function in order to populate the `price_info` field in the `OrchestratorInfo` message returned to a broadcaster during discovery.

```
// BasePriceInfo returns the orchestrator's configured base price info
basePriceInfo = BasePriceInfo()
txCostOverhead = TxCostOverhead(addr)

basePricePerPixel = priceInfo.pricePerUnit / priceInfo.pixelsPerUnit
pricePerPixel = basePricePerPixel * (1 + txCostOverhead)

if pricePerPixel < 1 {
    // If price per pixel is less than 1 wei then set
    // the price per unit to 1 wei and pay for multiple pixels
    priceInfo.pricePerUnit = 1
    priceInfo.pixelsPerUnit = floor(1 / pricePerPixel)
} else {
    // If price per pixel is >= 1 wei then set
    // the price per unit to the price per pixel and
    // pay for a single pixel
    priceInfo.pricePerUnit = pricePerPixel
    priceInfo.pixelsPerUnit = 1
}

return priceInfo
```

The returned `PriceInfo` can be included in the `OrchestratorInfo` message.

kyriediculous:
Allows an `Orchestator` to define a `pixelsPerUnit` and `pricePerUnit` flag which results as an O's base `Price Per Pixel`, represented as a `big.Rat`. Defining both the amount of pixels in a unit and the price for it allows smaller price granularity than `1 wei / pixel`. This baseInfo is then used to create `PriceInfo` for a `BroadCaster` based on the transaction cost to redeem a payment on-chain (see #904). This PriceInfo changes depending on how the `MaxFloat` from a B to an O affects the transaction cost for payment redemption. 

When a B has adequate `MaxFloat` to an O to cover for its `faceValue` the overhead will always be 1% 
-> `basePricePerPixel * 1,01`

When a B has only half of the `MaxFloat` to cover for an O's `faceValue` the overhead becomes double of that , or 2%. 
-> `basePricePerPixel * 1,02`

- Added Flags in `cmd` allowing an O to set `pixelsPerUnit` and `pricePerUnit`. Added error handling if either is set to 0. (Defaults : `PricePerUnit = 0` , `PixelsPerUnit = 1`. This means an Orchestrator will have to explicitly set a price, as it defaults to 0 to prevent the node starting up with 1 wei/pixel unknowingly.
- Added a `priceInfo` field to the `LivepeerNode` struct to set `PriceInfo` for an `Orchestrator` after calling `NewOrchestrator(node)`
- Created a new method in `recipient.go` , `r.txCost()`, which calculates the transaction cost based on the current `gasPrice` (through `gasPriceMonitor`) and `r.cfg.RedeemGas`. This method is introduced to keep the code DRY and is now used in both `r.faceValue()` and `r.TxCostMultiplier`
- Created a new method in `recipient.go` , `r.TxMultiplier(sender)` which dynamically calculates and returns the `TxCostMultiplier` based upon the sender's `MaxFloat`
- added `r.TxMultiplier(sender)` to the `Recipient` interface
- Created a new method in `orchestrator.go`, `orch.PriceInfo(sender)` which returns an orchestrator's priceInfo depending on the sender based on the `PriceInfo` set by an orchestrator, scaled by multiplying `1+ 1/TxCostMultiplier`.
- Added `orch.PriceInfo` to the `Orchestrator` interface in `rpc.go`
- Added unit tests for `orch.PriceInfo` and `r.TxCostMultiplier()`
- Added tests in `test_args.sh` for testing the new flags and its fatal errors. 

- Ran unit tests 
- Ran argument tests 
- Manually started a rinkeby node setting `PricePerUnit` 
_Ubuntu 18.04 LTS Bionic Beaver
go1.12.2 linux/amd64_

Fixes #894 

**Checklist:**
- [x] Node runs in ~~OSX~~ _Ubuntu_ and devenv

### DIFF
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -103,6 +103,10 @@ func main() {
 	initializeRound := flag.Bool("initializeRound", false, "Set to true if running as a transcoder and the node should automatically initialize new rounds")
 	ticketEV := flag.String("ticketEV", "1000000000", "The expected value for PM tickets")
 
+	// Price Info
+	pricePerUnit := flag.Int("pricePerUnit", 0, "The price per 'pixelsPerUnit' amount pixels")
+	pixelsPerUnit := flag.Int("pixelsPerUnit", 1, "Amount of pixels per unit. Set to '> 1' to have smaller price granularity than 1 wei / pixel")
+
 	// Metrics & logging:
 	monitor := flag.Bool("monitor", false, "Set to true to send performance metrics")
 	version := flag.Bool("version", false, "Print out the version")
@@ -311,6 +315,21 @@ func main() {
 		n.EthServices["UnbondingService"] = eventservices.NewUnbondingService(n.Eth, dbh)
 
 		if *orchestrator {
+
+			// Set price per pixel base info
+			if *pixelsPerUnit <= 0 {
+				// Can't divide by 0
+				glog.Fatalf("The amount of pixels per unit must be greater than 0, provided %d instead\n", *pixelsPerUnit)
+				return
+			}
+			if *pricePerUnit <= 0 {
+				// Prevent orchestrator from unknowingly provide free transcoding
+				glog.Fatalf("Price per unit of pixels must be greater than 0, provided %d instead\n", *pricePerUnit)
+				return
+			}
+			n.PriceInfo = big.NewRat(int64(*pricePerUnit), int64(*pixelsPerUnit))
+			glog.Infof("Price: %d wei for %d pixels\n ", *pricePerUnit, *pixelsPerUnit)
+
 			ctx, cancel := context.WithCancel(context.Background())
 			defer cancel()
 
@@ -496,6 +515,7 @@ func main() {
 		if core.OrchestratorNode != n.NodeType {
 			return
 		}
+
 		orch := core.NewOrchestrator(s.LivepeerNode)
 
 		go func() {
--- a/core/orch_test.go
+++ b/core/orch_test.go
@@ -749,6 +749,131 @@ func TestTicketParams_Error(t *testing.T) {
 	assert.EqualError(t, err, expErr.Error())
 }
 
+func TestPriceInfo_ReturnsBigRat(t *testing.T) {
+	// basePrice = 1/1, txMultiplier = 100/1 => expPricePerPixel = 101/100
+	basePrice := big.NewRat(1, 1)
+	txMultiplier := big.NewRat(100, 1)
+	expPricePerPixel := big.NewRat(101, 100)
+
+	n, _ := NewLivepeerNode(nil, "", nil)
+	n.PriceInfo = basePrice
+
+	recipient := new(pm.MockRecipient)
+	n.Recipient = recipient
+	recipient.On("TxCostMultiplier", mock.Anything).Return(txMultiplier, nil)
+	orch := NewOrchestrator(n)
+
+	priceInfo, err := orch.PriceInfo(ethcommon.Address{})
+	assert.Nil(t, err)
+	assert.Zero(t, expPricePerPixel.Cmp(priceInfo))
+
+	// basePrice = 10/1, txMultiplier = 100/1 => expPricePerPixel = 1010/100
+	basePrice = big.NewRat(10, 1)
+	n.PriceInfo = basePrice
+	orch = NewOrchestrator(n)
+	expPricePerPixel = big.NewRat(1010, 100)
+
+	priceInfo, err = orch.PriceInfo(ethcommon.Address{})
+	assert.Nil(t, err)
+	assert.Zero(t, expPricePerPixel.Cmp(priceInfo))
+
+	// basePrice = 1/10, txMultiplier = 100 => expPricePerPixel = 101/1000
+	basePrice = big.NewRat(1, 10)
+	n.PriceInfo = basePrice
+	orch = NewOrchestrator(n)
+	expPricePerPixel = big.NewRat(101, 1000)
+
+	priceInfo, err = orch.PriceInfo(ethcommon.Address{})
+	assert.Nil(t, err)
+	assert.Zero(t, expPricePerPixel.Cmp(priceInfo))
+
+	// basePrice = 25/10 , txMultiplier = 100 => expPricePerPixel = 2525/1000
+	basePrice = big.NewRat(25, 10)
+	n.PriceInfo = basePrice
+	orch = NewOrchestrator(n)
+	expPricePerPixel = big.NewRat(2525, 1000)
+
+	priceInfo, err = orch.PriceInfo(ethcommon.Address{})
+	assert.Nil(t, err)
+	assert.Zero(t, expPricePerPixel.Cmp(priceInfo))
+
+	// basePrice = 10/1 , txMultiplier = 100/10 => expPricePerPixel = 11
+	basePrice = big.NewRat(10, 1)
+	txMultiplier = big.NewRat(100, 10)
+	n.PriceInfo = basePrice
+	recipient = new(pm.MockRecipient)
+	n.Recipient = recipient
+	recipient.On("TxCostMultiplier", mock.Anything).Return(txMultiplier, nil)
+	orch = NewOrchestrator(n)
+	expPricePerPixel = big.NewRat(11, 1)
+
+	priceInfo, err = orch.PriceInfo(ethcommon.Address{})
+	assert.Nil(t, err)
+	assert.Zero(t, expPricePerPixel.Cmp(priceInfo))
+
+	// basePrice = 10/1 , txMultiplier = 1/10 => expPricePerPixel = 110
+	basePrice = big.NewRat(10, 1)
+	txMultiplier = big.NewRat(1, 10)
+	n.PriceInfo = basePrice
+	recipient = new(pm.MockRecipient)
+	n.Recipient = recipient
+	recipient.On("TxCostMultiplier", mock.Anything).Return(txMultiplier, nil)
+	orch = NewOrchestrator(n)
+	expPricePerPixel = big.NewRat(1100, 10)
+
+	priceInfo, err = orch.PriceInfo(ethcommon.Address{})
+	assert.Nil(t, err)
+	assert.Zero(t, expPricePerPixel.Cmp(priceInfo))
+
+	// basePrice = 10, txMultiplier = 1 => expPricePerPixel = 20
+	basePrice = big.NewRat(10, 1)
+	txMultiplier = big.NewRat(1, 1)
+	n.PriceInfo = basePrice
+	recipient = new(pm.MockRecipient)
+	n.Recipient = recipient
+	recipient.On("TxCostMultiplier", mock.Anything).Return(txMultiplier, nil)
+	orch = NewOrchestrator(n)
+	expPricePerPixel = big.NewRat(20, 1)
+
+	priceInfo, err = orch.PriceInfo(ethcommon.Address{})
+	assert.Nil(t, err)
+	assert.Zero(t, expPricePerPixel.Cmp(priceInfo))
+}
+
+func TestPriceInfo_GivenNilNode_ReturnsNilError(t *testing.T) {
+	n, _ := NewLivepeerNode(nil, "", nil)
+	orch := NewOrchestrator(n)
+	orch.node = nil
+
+	priceInfo, err := orch.PriceInfo(ethcommon.Address{})
+	assert.Nil(t, err)
+	assert.Nil(t, priceInfo)
+}
+
+func TestPriceInfo_GivenNilRecipient_ReturnsNilError(t *testing.T) {
+	n, _ := NewLivepeerNode(nil, "", nil)
+	orch := NewOrchestrator(n)
+	n.Recipient = nil
+
+	priceInfo, err := orch.PriceInfo(ethcommon.Address{})
+	assert.Nil(t, err)
+	assert.Nil(t, priceInfo)
+}
+
+func TestPriceInfo_TxMultiplierError_ReturnsError(t *testing.T) {
+	expError := errors.New("TxMultiplier Error")
+
+	n, _ := NewLivepeerNode(nil, "", nil)
+	recipient := new(pm.MockRecipient)
+	n.Recipient = recipient
+	recipient.On("TxCostMultiplier", mock.Anything).Return(nil, expError)
+	orch := NewOrchestrator(n)
+
+	priceInfo, err := orch.PriceInfo(ethcommon.Address{})
+	assert.Nil(t, priceInfo)
+	assert.EqualError(t, err, expError.Error())
+}
+
 func defaultPayment(t *testing.T) net.Payment {
 	ticketSenderParams := &net.TicketSenderParams{
 		SenderNonce: 456,
--- a/core/orchestrator.go
+++ b/core/orchestrator.go
@@ -160,6 +160,20 @@ func (orch *orchestrator) TicketParams(sender ethcommon.Address) (*net.TicketPar
 	}, nil
 }
 
+func (orch *orchestrator) PriceInfo(sender ethcommon.Address) (*big.Rat, error) {
+	if orch.node == nil || orch.node.Recipient == nil {
+		return nil, nil
+	}
+
+	txCostMultiplier, err := orch.node.Recipient.TxCostMultiplier(sender)
+	if err != nil {
+		return nil, err
+	}
+	// pricePerPixel = basePrice * (1 + 1/ txCostMultiplier)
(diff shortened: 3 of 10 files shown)


cmd/livepeer/livepeer.go
```diff
@@ -103,6 +103,10 @@ func main() {
 	initializeRound := flag.Bool("initializeRound", false, "Set to true if running as a transcoder and the node should automatically initialize new rounds")
 	ticketEV := flag.String("ticketEV", "1000000000", "The expected value for PM tickets")
 
+	// Price Info for Ticket Params
```
yondonfu: Technically these price related values don't actually affect how ticket params are determined so I suggest leaving out "for Ticket Params" to avoid any confusion for future readers

cmd/livepeer/livepeer.go
```diff
@@ -103,6 +103,10 @@ func main() {
...
+	pricePerUnit := flag.Int("pricePerUnit", 0, "The price per X amount pixels")
```
yondonfu: What do you think about replacing "X" with "pixelsPerUnit" to emphasize the relationship between the `pricePerUnit` and `pixelsPerUnit` flags?

cmd/livepeer/livepeer.go
```diff
@@ -311,6 +318,21 @@ func main() {
 		n.EthServices["UnbondingService"] = eventservices.NewUnbondingService(n.Eth, dbh)
 
 		if *orchestrator {
+
+			// Set price per pixel base info
+			if *pixelsPerUnit == 0 {
+				// Can't divide by 0
+				glog.Fatalln("The amount of pixels per unit must be greater than 0")
```
yondonfu: It could be helpful for the logged message to also indicate to the user the provided value

Ex. `glog.Fatalf("-pixelsPerUnit must be greater than 0, but %d provided", *pixelsPerUnit)`

Same comment for the `pricePerUnit` check

cmd/livepeer/livepeer.go
```diff
@@ -311,6 +318,21 @@ func main() {
...
+				return
+			}
+			if *pricePerUnit == 0 {
+				glog.Fatalln("Price per unit of pixels is set to 0, please set a price")
+				return
+			} else {
```
yondonfu: I think we can drop this `else` block since the code inside will only run if the `pixelsPerUnit` and `pricePerUnit` checks pass

cmd/livepeer/livepeer.go
```diff
@@ -311,6 +318,21 @@ func main() {
...
+				n.PriceInfo = big.NewRat(int64(*pricePerUnit), int64(*pixelsPerUnit))
+				glog.Infof("Price per pixel set to %v wei", pricePerPixel)
```
yondonfu: This will print out a `pricePerPixel` as a fraction - I wonder if the presentation of this value could be a bit clearer.

One possibility:

```
glog.Infof("Price: %d wei for %d pixels", pricePerPixel.Num(), pricePerPixel.Denom())
```

What do you think?

pm/recipient.go
```diff
@@ -186,13 +189,17 @@ func (r *recipient) TicketParams(sender ethcommon.Address) (*TicketParams, error
 	}, nil
 }
 
-func (r *recipient) faceValue(sender ethcommon.Address) (*big.Int, error) {
+func (r *recipient) txCost() *big.Int {
+	// Fetch current gasprice from cache through gasPrice monitor
 	gasPrice := r.gpm.GasPrice()
 	// txCost = redeemGas * gasPrice
 	txCost := new(big.Int).Mul(big.NewInt(int64(r.cfg.RedeemGas)), gasPrice)
+	return txCost
```
yondonfu: Instead of declaring the `txCost` variable, we could directly return the computed value in the previous line

i.e.

```
return new(big.Int).Mul(big.NewInt(int64(r.cfg.RedeemGas)), gasPrice)
```

kyriediculous:
force-pushed: Linter check failed because I didn't add a comment to exported method `mockRecipient.TxMultiplier` even though there's plenty of other exported methods that are uncommented. Is not every file added to the linter?

core/orchestrator.go
```diff
@@ -33,8 +33,9 @@ var transcodeLoopTimeout = 1 * time.Minute
 
 // Transcoder / orchestrator RPC interface implementation
 type orchestrator struct {
-	address ethcommon.Address
-	node    *LivepeerNode
+	address       ethcommon.Address
+	node          *LivepeerNode
+	basePriceInfo *big.Rat
```
yondonfu: Since `LivepeerNode` is already holding the price, I don't think we need to also hold the price here. Then, in `PriceInfo` we can use `orch.node.PriceInfo`.

cmd/livepeer/livepeer.go
```diff
@@ -155,6 +159,9 @@ func main() {
 		}
 	}
 
+	// Initialize pricePerPixel
+	var pricePerPixel *big.Rat
```
yondonfu: Perhaps we should call this `priceInfo` to be consistent about naming (since we're already using the "PriceInfo" naming convention elsewhere)

core/orch_test.go
```diff
@@ -749,6 +749,51 @@ func TestTicketParams_Error(t *testing.T) {
...
+	recipient.On("TxCostMultiplier", mock.Anything).Return(txMultiplier, nil)
+	orch := NewOrchestrator(n)
+
+	priceInfo, err := orch.PriceInfo(ethcommon.Address{})
+
+	assert.Nil(t, err)
+	assert.Equal(t, expPricePerPixel, priceInfo)
+}
+
+func TestPriceInfo_GivenNilRecipient_ReturnsNilError(t *testing.T) {
```
yondonfu, line 853: Test for when `orch.node == nil` as well?

test_args.sh
```diff
@@ -89,12 +89,21 @@ res=0
 ./livepeer -orchestrator -serviceAddr 127.0.0.1:8935 || res=$?
 [ $res -ne 0 ]
 
+# Orchestrator needs PricePerUnit > 0 
+res=0
+./livepeer -orchestrator -serviceAddr 127.0.0.1:8935 -transcoder -pricePerUnit 0 -network rinkeby $ETH_ARGS || res=$?
```
yondonfu, line 98: I think its worthwhile to add a test for when `-pricePerUnit` flag is not set at all since we expect the default value to be 0 resulting in the node shutting down

test_args.sh
```diff
@@ -89,12 +89,21 @@ res=0
...
+[ $res -ne 0 ]
+# Orchestrator needs PixelsPerUnit > 0
+res=0
+./livepeer-orchestrator -serviceAddr 127.0.0.1:8935 -transcoder -pixelsPerUnit 0 -pricePerUnit 5 -network rinkeby $ETH_ARGS || res=$?
```
yondonfu: Space between `./livepeer` and `-orchestrator`?

core/orch_test.go
```diff
@@ -749,6 +749,51 @@ func TestTicketParams_Error(t *testing.T) {
 	assert.EqualError(t, err, expErr.Error())
 }
 
+func TestPriceInfo_ReturnsBigRat(t *testing.T) {
+	basePrice := big.NewRat(1, 1)
+	txMultiplier := big.NewRat(100, 1)
```
yondonfu, line 755: Would be nice to have some additional tests using different values for `basePrice` and `txMultiplier`.

i.e.

- `basePrice.Num() > 1`
- `basePrice.Denom() > 1`
- `basePrice.Num() > 1 && basePrice.Denom() > 1`
- `txMultiplier.Denom() > 1`
- `txMultiplier.Num() = 1 && txMultiplier.Denom() > 1`
- `txMultiplier = 1/1`

yondonfu (changes requested):

yondonfu:
> Is not every file added to the linter?

The linter run by CI only checks the `pm` package at the moment (see [.circleci/config.yml](https://github.com/livepeer/go-livepeer/blob/master/.circleci/config.yml#L79)). The linter was only recently added to the CI around the time when the `pm` package was first introduced. Since a linter was not used previously, many of the existing packages did not adhere to all the standard linting rules. So, rather than fix all the style issues in a single pass, the thought was to apply the linter to all new packages and gradually fix the style issues in existing packages over time.

cmd/livepeer/livepeer.go
kyriediculous: Wouldn't this always result in the underlying ? I guess it could be a negative integer as well though, I just didn't ever see a user entering a negative number there that's why I omitted the provided value. It did cross my mind. 

> pixelsPerUnit must be greater than 0, but 0 provided

cmd/livepeer/livepeer.go
kyriediculous: Yeah I guess you're right. I'm just personally not a big fan of omitting `else` blocks for structure, but it does result in shorter code. Will try to set my personal preferences aside for the sake of the codebase :smile: . And even If I were to apply my personal preference I should have probably made the second one an `else if` as well.

cmd/livepeer/livepeer.go
kyriediculous: I guess it's this or a rounded float, where the seperation of numerator and denominator results in a clearer and more accurate message. So I like it ! 

We can actually use the values set by the flags directly here , no need to get numerator and denominator from the `big.Rat` value I think , it will also result in a `big.Int` rather than an `int` (I'm not sure if base10 formatting with `%d` resolves this ? )

`glog.Infof("Price: %d wei for %d pixels", *pricePerUnit, *pixelsPerUnit)
`

pm/recipient.go
kyriediculous: You're right. Just a personal thing again, I tend to do this a lot but will try not to and clarify the return value with a comment instead.

cmd/livepeer/livepeer.go
kyriediculous: Actually we can remove this, this is a remenant of using `NewOrchestrator` with an extra argument, but we're setting it on `LivepeerNode.PriceInfo` now

cmd/livepeer/livepeer.go
yondonfu: Yeah something like:

```
if a > 0 {
    return 1
} else {
    return 2
}
```

vs.

```
if a > 0 {
    return 1
}

return 2
```

is usually a matter of preference. But, I think the latter is considered more [idiomatic Go](https://golang.org/doc/effective_go.html#if) so I often just default to the latter (I tend to stick to the idiomatic approach in these cases).

cmd/livepeer/livepeer.go
yondonfu: > I guess it could be a negative integer as well though

Right and in fact we should probably ensure that `*pricePerUnit` and `*pixelsPerUnit` are not negative integers (we're just checking that they are not 0 at the moment).

While users typically wouldn't enter negative numbers, we should still make sure that our code is able to handle atypical input.

cmd/livepeer/livepeer.go
kyriediculous: You're right, I should complete effective Go and make notes on it ASAP.

cmd/livepeer/livepeer.go
kyriediculous: Entirely correct and thus changed the checks to check for `<= 0` and will add tests for negative ints as well.

core/orch_test.go
```diff
@@ -765,9 +764,90 @@ func TestPriceInfo_ReturnsBigRat(t *testing.T) {
...
+	recipient.On("TxCostMultiplier", mock.Anything).Return(txMultiplier, nil)
+	orch = NewOrchestrator(n)
+	expPricePerPixel = new(big.Rat).Mul(basePrice, new(big.Rat).Add(big.NewRat(1, 1), new(big.Rat).Inv(txMultiplier)))
+
+	priceInfo, err = orch.PriceInfo(ethcommon.Address{})
+	assert.Nil(t, err)
+	assert.Equal(t, expPricePerPixel, priceInfo)
+}
+
+func TestPriceInfo_GivenNilNode_ReturnsNilError(t *testing.T) {
```
yondonfu, line 843: Doesn't look like this test is actually testing for when `orch.node == nil` (only `orch.node.Recipient = nil`

core/orch_test.go
```diff
@@ -765,9 +764,90 @@ func TestPriceInfo_ReturnsBigRat(t *testing.T) {
...
+	n.Recipient = recipient
+	recipient.On("TxCostMultiplier", mock.Anything).Return(txMultiplier, nil)
+	orch = NewOrchestrator(n)
+	expPricePerPixel = new(big.Rat).Mul(basePrice, new(big.Rat).Add(big.NewRat(1, 1), new(big.Rat).Inv(txMultiplier)))
+
+	priceInfo, err = orch.PriceInfo(ethcommon.Address{})
+	assert.Nil(t, err)
+	assert.Equal(t, expPricePerPixel, priceInfo)
+
+	// basePrice = 10/1 , txMultiplier = 1/10 => expPricePerPixel = 100
```
yondonfu: Should `expPricePerPixel = 110`?

`(10 / 1) * (1 + inv(1/ 10)) = (10 / 1) * (1 + 10) = 10 * 11 = 110`

core/orch_test.go
```diff
@@ -765,9 +764,90 @@ func TestPriceInfo_ReturnsBigRat(t *testing.T) {
...
 
 	priceInfo, err := orch.PriceInfo(ethcommon.Address{})
+	assert.Nil(t, err)
+	assert.Equal(t, expPricePerPixel, priceInfo)
+
+	// basePrice = 10/1, txMultiplier = 100/1 => expPricePerPixel = 1010/100
+	basePrice = big.NewRat(10, 1)
+	n.PriceInfo = basePrice
+	orch = NewOrchestrator(n)
+	expPricePerPixel = new(big.Rat).Mul(basePrice, new(big.Rat).Add(big.NewRat(1, 1), new(big.Rat).Inv(txMultiplier)))
```
yondonfu: We're repeating this programmatic calculation for `expPricePerPixel` multiple times in this test which is the same logic implemented in the actual `PriceInfo()` method. Perhaps a more robust way to set expectations for these tests is to to hardcode the computed values (which you've already included in comments) for `expPricePerPixel` to ensure that the computation results in the correct value given a set of inputs.

yondonfu (changes requested):

core/orch_test.go
kyriediculous: Woops, you're right , I forgot doing +1 in my notebook for that one.

core/orch_test.go
kyriediculous: Perhaps you're right.
Creating those comments was actually the last thing I did.

Although a human would probably end up calculating those values according to the same formula in `PriceInfo()`, so whether hardcoded or not it doesn't really check the correctness of the formula in the context of creating appropriate price info. Just that the formula returns the results it should return. But I can see the benefit.

core/orch_test.go
kyriediculous: I just copy pasted from the recipient nil test, forgot to change node to nil instead.
My bad.

core/orch_test.go
yondonfu: Human error is certainly possible as well. Though, IMO it feels like we're just duplicating the code snippet from `PriceInfo()` in the test as well so if the code snippet was incorrect the tests would still pass since the expected value is calculated with the same code as the actual value.

kyriediculous: For reference: 

Hardcoding the values resulted in absolute comparison errors using `Assert.Equal` from Go's testing package. Therefore we are now using `big.Rat.Cmp()` wrapped in `Assert.Zero` from go's testing package, which seems to work.

For some reason the formatting of the denominators is different in the return value from `PriceInfo()` compared to hardcoding values with `big.NewRat()` : `big.nat{} vs big.nat(nil)`

This only happened for two tests though and might be related to a denominator set to `1` even though another test with the same context (denominator of 1) passed without issue. 

```
expected: &big.Rat{a:big.Int{neg:false, abs:big.nat{0x6e}}, b:big.Int{neg:false, abs:big.nat{}}}
actual  : &big.Rat{a:big.Int{neg:false, abs:big.nat{0x6e}}, b:big.Int{neg:false, abs:big.nat(nil)}}
```

yondonfu (approved):
LGTM :shipit: 

Needs a `git rebase --interactive --autosquash master` and then good to merge!

PR #920, opened 2019-06-17
doc/rtmpwebhookauth.md: English & grammar, minor corrections

### LINKED ISSUE #909: rtmpwebhookauth.md is difficult to understand due to grammar and typos
liverpeer should be livepeer, sentences should have a subject and a predicate.  If we make this more widely understood, we'd probably instill higher confidence levels from implementors and investors

This must be re-written and reviewed

mk-livepeer:
minor documentation corrections & cleanup.

closes #909

### DIFF
--- a/doc/rtmpwebhookauth.md
+++ b/doc/rtmpwebhookauth.md
@@ -22,7 +22,7 @@ For example, if the incoming RTMP request was made to `rtmp://livepeer.node:1935
 
 The webhook server should respond with HTTP status code `200` in order to authenticate / authorize the RTMP stream. A response with a HTTP status code other than `200` will cause the Livepeer node to drop the RTMP stream.
 
-The webhook may respond with an empty body.  In this case, the `manifestID` property of the stream will taken from the URL.  If the URL does not specify a manifest id, then it will be generated at random.  Otherwise, the webhook endpoint should respond with a JSON object in the following format:
+The webhook may respond with an empty body.  In this case, the `manifestID` property of the stream will be taken from the RTMP URL.  If the RTMP URL does not specify a manifest id, then it will be generated at random.  Otherwise, the webhook endpoint should respond with a JSON object in the following format:
 
 ```json
 {
--- a/doc/rtmpwebhookauth.md
+++ b/doc/rtmpwebhookauth.md
@@ -20,7 +20,7 @@ For example, if the incoming RTMP request was made to `rtmp://livepeer.node:1935
 }
 ```
 
-The webhook server should respond with HTTP status code `200` in order to authenticate / allow the RTMP stream. A response with a HTTP status code other than `200` will cause the Livepeer node to drop the RTMP stream.
+The webhook server should respond with HTTP status code `200` in order to authenticate / authorize the RTMP stream. A response with a HTTP status code other than `200` will cause the Livepeer node to drop the RTMP stream.
 
 The webhook may respond with an empty body.  In this case, the `manifestID` property of the stream will taken from the URL.  If the URL does not specify a manifest id, then it will be generated at random.  Otherwise, the webhook endpoint should respond with a JSON object in the following format:
 
--- a/doc/rtmpwebhookauth.md
+++ b/doc/rtmpwebhookauth.md
@@ -20,7 +20,7 @@ For example, if the incoming RTMP request was made to `rtmp://livepeer.node:1935
 }
 ```
 
-The webhook server should respond with HTTP status code `200` in order to authenticate / authorize the RTMP stream. A response with a HTTP status code other than `200` will cause the Livepeer node to drop the RTMP stream.
+The webhook server should respond with HTTP status code `200` in order to authenticate / authorize the RTMP stream. A response with a HTTP status code other than `200` will cause the Livepeer node to disconnect the RTMP stream.
 
 The webhook may respond with an empty body.  In this case, the `manifestID` property of the stream will be taken from the RTMP URL.  If the RTMP URL does not specify a manifest id, then it will be generated at random.  Otherwise, the webhook endpoint should respond with a JSON object in the following format:
 
--- a/doc/rtmpwebhookauth.md
+++ b/doc/rtmpwebhookauth.md
@@ -12,11 +12,11 @@ livepeer -authWebhookUrl http://ownserver/auth
 
 For each incoming RTMP stream, the Livepeer node will make a `POST` request to the `http://ownserver/auth` endpoint, passing the URL of the RTMP request as JSON object.
 
-For example, if the incoming RTMP request was made to `rtmp://livepeer.node:1935/something?manifestID=manifest`, the Liverpeer node will provide the following object as a request to the webhook endpoint:
+For example, if the incoming RTMP request was made to `rtmp://livepeer.node/manifest`, the Liverpeer node will provide the following object as a request to the webhook endpoint:
 
 ```json
 {
-    "url": "rtmp://livepeer.node:1935/manifest"
+    "url": "rtmp://livepeer.node/manifest"
 }
 ```
 
--- a/doc/rtmpwebhookauth.md
+++ b/doc/rtmpwebhookauth.md
@@ -1,14 +1,18 @@
 # RTMP Webhook Authentication
 
-Incoming RTMP streams can be authenticated using Webhooks. To use it you need to implement own web server which will be called by Livepeer node when new RTMP stream received to check if that stream is allowed.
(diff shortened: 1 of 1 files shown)


doc/rtmpwebhookauth.md
```diff
@@ -1,24 +1,28 @@
 # RTMP Webhook Authentication
 
-Incoming RTMP streams can be authenticated using Webhooks. To use it you need to implement own web server which will be called by Livepeer node when new RTMP stream received to check if that stream is allowed.
+Incoming RTMP streams can be authenticated using Webhooks. To provide these webhooks, node operators must implement their own web service / endpoint to be accessed only by the Livepeer node. As new RTMP streams appear, the Livepeer node will hit this endpoint to determine whether the given stream is allowed.
```
j0sh: > To provide these webhooks,

"To use these webhooks," ?

> the Livepeer node will hit this endpoint

"the Livepeer node will call this endpoint"

doc/rtmpwebhookauth.md
```diff
@@ -1,24 +1,28 @@
...
 
-To enable Webhook authentication functionality Livepeer node should be started with flag `-authWebhookUrl`, like this:
+To enable webhook authentication functionality, the Livepeer node should be started with the flag, `-authWebhookUrl` along with the webhook endpoint URL.
```
j0sh: > with the flag, `-authWebhookUrl` 

with the `-authWebhookUrl` flag

doc/rtmpwebhookauth.md
```diff
@@ -1,24 +1,28 @@
...
 
-For every incoming RTMP stream Livepeer node will make `POST` request to `http://ownserver/auth` endpoint passing url to which RTMP request was made as JSON object.
+```console
+livepeer -authWebhookUrl http://ownserver/auth
+```
+
+For each incoming RTMP stream, the Livepeer node will make a `POST` request to the `http://ownserver/auth` endpoint, passing the URL of the RTMP request as JSON object.
 
-For example, if incoming RTMP request was made to `rtmp://livepeer.node:1935/something?manifestID=manifest`, Liverpeer node will pass this object in request to webhook:
+For example, if the incoming RTMP request was made to `rtmp://livepeer.node:1935/something?manifestID=manifest`, the Liverpeer node will provide the following object as a request to the webhook endpoint:
```
j0sh: The RTMP URL can be changed to `rtmp://livepeer.node:1935/manifest` to match the json body. Also, the `:1935` can probably be omitted since it's implied as the default RTMP port (json body example should be updated as well).

doc/rtmpwebhookauth.md
```diff
@@ -1,24 +1,28 @@
...
 
 ```json
 {
     "url": "rtmp://livepeer.node:1935/manifest"
 }
 ```
 
-Webhook server should respond with code 200 if it wants to allow RTMP stream. Response with any other code will lead to Livepeer node dropping RTMP stream.
+The webhook server should respond with HTTP status code `200` in order to authenticate / allow the RTMP stream. A response with a HTTP status code other than `200` will cause the Livepeer node to drop the RTMP stream.
```
j0sh: > authenticate / allow 

authenticate / authorize ?

> cause the Livepeer node to drop the RTMP stream.

"cause the Livepeer node to terminate the RTMP stream"

doc/rtmpwebhookauth.md
```diff
@@ -1,24 +1,28 @@
...
 
-Webhook can respond with zero body - in that case `ManifestID` for the stream will be taken from url or generated randomly in case manifest id is not specified in url. Or webhook server can respond with JSON object in this format:
+The webhook may respond with an empty body.  In this case, the `manifestID` property of the stream will taken from the URL.  If the URL does not specify a manifest id, then it will be generated at random.  Otherwise, the webhook endpoint should respond with a JSON object in the following format:
```
j0sh: > will taken from the URL

"will be taken from the RTMP URL" (added 'be' and specified RTMP)

>  If the URL does not specify a manifest id

"If the RTMP URL does not specify a manifest id"

And maybe give the RTMP URLs that are considered to be missing manifest IDs: `rtmp://host` and `rtmp://host/stream` (I don't recall if we document the `stream`-stripping behavior anywhere)

j0sh (changes requested):
Thanks for making these updates! Just a few nits.

doc/rtmpwebhookauth.md
mkrufky: thanks.  addressed in cbe12e2e45fe29193b9385c06451e9af5b0176c7

doc/rtmpwebhookauth.md
mkrufky: thanks.  addressed in 76c1c1f02776a600a29b4cca3af20e3d89b22ee8

doc/rtmpwebhookauth.md
mkrufky: `allow` changed to `authorize` in 65d6f015b7aa875e6d606161da115e1559610b8a

mkrufky: I disagree with usage of the word "terminate" here.  No matter what `go-livepeer` wants, it will not succeed in terminating the incoming stream if it arrives via UDP.   Even if it arrives via TCP, many third party RTMP streamers will attempt reconnection in the event of a disconnect.  I think the word, "drop" is most appropriate here, but if you reiterate your preference fir the word, "terminate" then I will change it.

mkrufky: I disagree with usage of the word "terminate" here.  No matter what `go-livepeer` wants, it will not succeed in terminating the incoming stream if it arrives via UDP.   Even if it arrives via TCP, many third party RTMP streamers will attempt reconnection in the event of a disconnect.  I think the word, "drop" is most appropriate here, but if you reiterate your preference for the word, "terminate" then I will change it.

doc/rtmpwebhookauth.md
mkrufky: partially addressed in 2d7244324bdade66d0e440cc46273801e52d29bf

doc/rtmpwebhookauth.md
j0sh: > it will not succeed in terminating the incoming stream if it arrives via UDP

We don't use UDP for ingest though?

> many third party RTMP streamers will attempt reconnection in the event of a disconnect

Could use the term "disconnect" if that works better

>  think the word, "drop" is most appropriate here

To me, "drop" has a more transitory connotation, eg dropped packets, dropped segments - a temporary hiccup that may occur without affecting the rest of the processing. Here, we're stopping / terminating the processing entirely, at least from the node's perspective. Also, since "dropping segments" is a thing we do, I want to stay as consistent and tight with terminology as possible.

The term "disconnect" works as well, but also open to other suggestions.

mkrufky: ok ... done in 07c6ab17ff7cadad02544cb112ed2aa4052c97fd

doc/rtmpwebhookauth.md
mkrufky: done in f69f993fffbd8bc2d3c6eb5e0a600e64ad420e5e

j0sh:
LGTM after fixing up the commits 🚢 

Since this is my first encounter with the new review workflow and I guess we're still iterating on it, some notes on that:

I don't know how others feel about this but I'd be totally fine with force-pushing updates here. With more complex changesets where it's easier to review things on a per-commit basis (eg, https://github.com/livepeer/go-livepeer/pull/926), it seems that many tiny git changes actually would make things harder to follow and reconcile on a per-commit basis. Or maybe I'm just not understanding the intended review workflow.

Didn't have any issues reviewing this PR anyway - truthfully, I just ignored the incremental commits and looked at the full diff since the changes were straightforward. It does feel like it'd be a bit cumbersome to be doing separate commits (and cross-referencing) for each note, but I won't object if others really like the workflow. @mk-livepeer curious how this was for you?

j0sh (approved):

mkrufky:
I like to reference the commits that resolve a given comment, but I do not ask others to do it.

There was nothing special about this workflow.   I made a PR, you made some comments, I address them, you approve, we squash, then merge.

I am 100% against a force-push in mid-review.  It removes the frame of reference from the reviewer.

Please never do that -- force push is an obfuscation - it hides the previous work and prevents the reviewer from being able to see the incremental difference.

If the reviewer does not care to see the incremental difference, the reviewer should just look at the full diff as you described.  A pull requestor should not force this choice on the reviewer -- force push is not a nice way to treat somebody reviewing your work.

NEVER force-push if somebody has already begun looking at your PR.  Instead, just squash it at the end.

PR #923, opened 2019-06-17
ci: build binaries with Ubuntu 16.04

iameli:
Some of our GPU testing servers have Ubuntu 16.04 installed, which was incompatible with the binaries we were using. There are more sophisticated ways to fix this but tweaking the Docker base fixes it for the CI builds at least.


j0sh (approved):

j0sh:
Are builds still compatible with Ubuntu 18.04 with this change?

iameli:
Yep! glibc bumps like this appear to be backwards-compatible but not forwards-compatible.

PR #926, opened 2019-06-18
Update ticket redemption strategy

### LINKED ISSUE #829: pm: Implement orchestrator strategy for winning ticket redemption

yondonfu:

This PR updates the winning ticket redemption strategy for an orchestrator to the following:

- If the broadcaster’s max float is sufficient to cover the ticket face value, the orchestrator will redeem the ticket immediately upon receiving it
- If the broadcaster’s max float is insufficient to cover the ticket face value, the orchestrator will add the ticket to a queue associated with the broadcaster and redeem the ticket when the broadcaster’s max float is sufficient to cover the ticket face value

Previously, an orchestrator would hold on to winning tickets and only redeem them when the transcode loop (the goroutine that transcodes incoming segments) for a particular manifestID timed out. Redeeming tickets upon transcode loop timeout is fine in a trusted environment with an honesty assumption for broadcasters, but an orchestrator should not use this strategy in an untrusted environment because a broadcaster’s deposit could be insufficient to cover all outstanding tickets by the time the orchestrator redeems the tickets after the transcode loop times out.

Furthermore, the latest PM protocol changes introduce a max float derived from a broadcaster’s reserve which represents the maximum amount of value a broadcaster can guarantee an orchestrator. Note: We make a distinction between a broadcaster’s reserve allocation for an orchestrator and the broadcaster’s max float for an orchestrator. The former is the current funds in the broadcaster’s reserve available to be claimed by an orchestrator in case the broadcaster’s deposit is insufficient. The latter is computed from the former and the value of winning tickets received by an orchestrator that have not yet been redeemed on-chain - prior to receiving any winning tickets an orchestrator defines the max float for a broadcaster to be equal to the reserve allocation and when it receives/redeems winning tickets it will decrease/increase the max float.

When an orchestrator receives a winning ticket, the broadcaster’s max float is decreased by ticket.FaceValue because until the ticket is redeemed on-chain, the orchestrator does not know if redemption will draw ticket.FaceValue from the broadcaster’s reserve or deposit. When an orchestrator successfully redeems a ticket on-chain, the broadcaster’s max float is increased because the orchestrator knows that ticket.FaceValue was not drawn from the broadcaster’s reserve so those funds are still intact.

Based on the changes in #904, there are two cases to consider:

ticket.FaceValue < MaxFloat(sender)
ticket.FaceValue = MaxFloat(sender)

The updates in this PR seek to handle both of these cases.

When an orchestrator receives a winning ticket, it will decrease the max float for a broadcaster such that subsequent ticket params will be based on this updated max float (see #905 for the changes that allow an orchestrator to notify a broadcaster of new ticket params).

If the max float is sufficient to cover the ticket.FaceValue, the orchestrator will redeem the ticket. Once the redemption confirms on-chain, the orchestrator will increase the max float for the broadcaster (once again subsequent ticket params will be based on this updated value). However, if the max float is insufficient to cover the ticket.FaceValue (the orchestrator could have received successive winning tickets), the orchestrator will add the ticket to a queue. Then, when there is sufficient max float available, the orchestrator will pull the ticket from the queue and try to redeem it again.

Here is a high level summary of some of the biggest changes:

- 13db566: Add support for redeeming a single winning ticket with ticket data that is already in-memory (previously, the orchestrator would read multiple tickets from disk)
- 7fe7598: Add `pm/queue.go` which includes a ticket queue implementation to handle winning tickets received when a broadcaster's max float is insufficient to cover them (i.e. O is handling N streams for a single B and happens to receive > 1 winning tickets successively)
- 6bc826f and 7d0f41f: Swap the existing `pm.FloatMonitor` for `pm.SenderMonitor` which not only monitors the current max float for active broadcasters, but also maintains ticket queues for each active broadcaster
- 0dee6d1 and 4f75c42: Add a redeem manager helper goroutine to `pm.Recipient` which handles tickets popped from each of the ticket queues associated with each active broadcaster
- a7f1df5: Fixes the OrchestratorInfo update logic - previously, we were not returning updated OrchestratorInfo messages if the orchestrator received a ticket with an already revealed recipientRand (which is always the case for the next ticket after a winning ticket is redeemed)

See the commit log for more details on other changes.

Apologies for this fairly large PR, I got a little carried away. I tried to break up each logical change into a standalone commit though so I would suggest reviewing commit by commit.

Added unit tests. Did some manual testing in on-chain mode using the devtool script.

Fixes #829

**Checklist:**


kyriediculous:
My comment might be a little premature here (I haven't checked the commits themselves yet, just PR overview). 

Sorry if I'm rambling over the top here as it's not really part of the code review but rather the process in general and I presume what I mentioned are just extreme edge cases that will likely never occur. 

> Then, when there is sufficient max float available, the orchestrator will pull the ticket from the queue and try to redeem it again.

Is there a policy in place for when a broadcaster doesn't add to its reserve so that winning tickets sent to an O can eventually be redeemed even though not in full, such as a period of X time (e.g. duration of `freezePeriod`) during which we will wait for the reserve to increase until tickets can be redeemed in full otherwise just take whatever is left in the reserve for compensation ? This is probably an extreme edge case though and only happens when a B never returns to the network again (and the transcoding job is at its end presumably). I'm thinking if there would be cases where it would be economically advantageous for the broadcaster to do this and sybil for the time of the `freezePeriod`, after which B can unlock and withdraw. 

The second thing I was wondering about is if there's ever a scenario possible in which the reserve does not get frozen AND is insufficient. We're counting on the fact that the last redeemed ticket before a deposit runs out would also claim from the reserve, freezing it. But it could always happen that the deposit becomes exactly `0` and the reserve (or maxFloat to an O rather) becomes insufficient in the round after. Although probably an extreme edge case again. But it might be worth splitting up the `else` in an `else if` for when `ticket.faceValue == sender.deposit` perhaps? I'm not sure. 

```
 if (_ticket.faceValue > sender.deposit) {
            // If ticket face value > sender's deposit then claim from
            // the sender's reserve

            amountToTransfer = sender.deposit.add(claimFromReserve(
                _ticket.sender,
                _ticket.recipient,
                _ticket.faceValue.sub(sender.deposit)
            ));

            sender.deposit = 0;
        } else {
            // If ticket face value <= sender's deposit then only deduct
            // from sender's deposit

            amountToTransfer = _ticket.faceValue;
            sender.deposit = sender.deposit.sub(_ticket.faceValue);
        }

```

yondonfu:
> Is there a policy in place for when a broadcaster doesn't add to its reserve so that winning tickets sent to an O can eventually be redeemed even though not in full, such as a period of X time (e.g. duration of freezePeriod) during which we will wait for the reserve to increase until tickets can be redeemed in full otherwise just take whatever is left in the reserve for compensation ?

Just to make sure we're on the same page, I *think* your description is captured by the following example scenario (let me know if you were thinking about something else):

1. B has reserve = X and its max float for O is Y
2. O sends B ticket params with faceValue = Y (assume that O's default faceValue is higher than B's max float so it adjusts its faceValue down to match B's max float)
3. B starts 2 streams with O and sends tickets with each stream
4. O receives 2 winning tickets simultaneously (1 for each stream)
5. O submits a redemption tx for the first ticket which sets B's max float to 0 until the tx confirms
6. O queues the second ticket since B's max float is insufficient to cover the ticket faceValue
7. The redemption tx for the first ticket confirms which should increase B's max float back to Y. However, the redemption tx drew from B's reserve which results in B's max float being less than Y.
8. The second ticket is never popped from the queue because B's max float cannot cover the second ticket's faceValue

We currently do not have a policy for O to automatically watch for B's reserve top offs. Furthermore, in the above scenario, B's reserve would be frozen and no funds could go in/out for the duration of the `freezePeriod` under the current contract rules. Considering whether this is desirable or not in livepeer/protocol#292.

> otherwise just take whatever is left in the reserve for compensation

At the moment, a ticket is only popped from the queue if B's max float >= ticket.faceValue. We could consider relaxing this condition to B's max float > redemption txCost (so that O at least comes out positive in the worst case)?

> I'm thinking if there would be cases where it would be economically advantageous for the broadcaster to do this and sybil for the time of the freezePeriod, after which B can unlock and withdraw.

Can you elaborate on "sybil for the time of the freezePeriod"?

> We're counting on the fact that the last redeemed ticket before a deposit runs out would also claim from the reserve, freezing it. But it could always happen that the deposit becomes exactly 0 and the reserve (or maxFloat to an O rather) becomes insufficient in the round after.

I don't think the reserve needs to be frozen if B's deposit is set exactly to 0 because B is only overspending if a ticket redemption draws from B's reserve. If a ticket redemption sets B's deposit to 0 without touching its reserve, then the reserve allocation for the Os in the current round remains the same. So, the reserve could only become insufficient after an O redeems an additional ticket that lowers its own reserve allocation from B's reserve.

kyriediculous:
> Just to make sure we're on the same page

Yes we are.

In addition: 
> B's reserve would be frozen and no funds could go in/out for the duration of the freezePeriod under the current contract rules.

While this is true, I think an O should still pop the ticket off the queue before it expires  (since `ticketValidityPeriod` < `freezePeriod` + `unlockPeriod` ) and be happy with taking less `faceValue` than the winning ticket has, rather than not redeeming it and not being compensated at all.

So as you describe (but only when maxFloat is insufficient), because currently a B will not be able to top off the reserve before the ticket expires considering `ticketValidityPeriod == freezePeriod`  : 

`We could consider relaxing this condition to B's max float > redemption txCost (so that O at least comes out positive in the worst case)?`


> If a ticket redemption sets B's deposit to 0 without touching its reserve, then the reserve allocation for the Os in the current round remains the same.

But for the next round it can become insufficient if there are more active O's, considering ticket validity period is 2 rounds if I'm not mistaken? 
But again likely an edge case that will never occur since the chances of it happening are rather small. 

> Can you elaborate on "sybil for the time of the freezePeriod"?


A B can create a different address during the time the reserve allocated for his original address is frozen and start pushing work to O's again. Maybe even the same O without knowing the same B (with a different address) still owes them winning ticket redemptions. When `freezePeriod` ends for the original address for B, he can still withdraw his (likely still insufficient) allocated reserve. Although after second thought I don't think there is any economical advantage to be gained for B, it would rather be an extreme case of griefing. 

______

But I don't think this discussion adds to the PR review process in any way so maybe we should have it somewhere else considering it is more protocol related and those situations would likely never occur. This is just me spitting out thoughts before thinking them over twice.

yondonfu:
> happy with taking less faceValue than the winning ticket has, rather than not redeeming it and not being compensated at all.

With this in mind, I think a better design is the following:

Let `pendingAmount` be the sum of face values of tickets that have been submitted for redemption, but the transactions have not confirmed on-chain yet.

`maxFloat = reserveAlloc - pendingAmount` where `reserveAlloc` is the portion of B's reserve that is guaranteed to O in the current round.

If `pendingAmount > 0`, then we only pop a ticket from the queue if `maxFloat >= ticket.FaceValue`. Since `pendingAmount > 0`, there is a ticket redemption pending confirmation, so we can wait for the confirmation to update `maxFloat` by updating `pendingAmount`. After the update, we can once again check if `maxFloat >= ticket.FaceValue`.

If `pendingAmount = 0`, then we only pop a ticket from the queue if `maxFloat >= txCost`. Since `pendingAmount = 0`, there is no ticket redemption pending confirmation, so `maxFloat = reserveAlloc` in this case. At this point, O knows that there definitely is not enough funds to back the full face value of its ticket. But, O would still like to be compensated as long as the pay out exceeds the transaction cost it incurs to get the pay out.

> But for the next round it can become insufficient if there are more active O's, considering ticket validity period is 2 rounds if I'm not mistaken?

Right. This is indeed an edge case: if O receives a winning ticket that is covered by B's max float at the end of round N and if O's redemption tx does not confirm on-chain until round N + 1, B's max float might not be sufficient to cover the winning ticket in round N + 1.

One way to address this is to check if the end of a round is imminent and to try to force a redemption tx on-chain ASAP by attaching a higher gas price (there would be an economic trade-off between being guaranteed B's max float vs. paying a higher tx cost relative to the estimated tx cost originally used to compute the ticket params). Given the low frequency of this scenario, I think we can probably address this later as a lower priority issue.

yondonfu:
Added a few commits.

Fixups:

- 1b80526: Start and stop the pm.SenderMonitor goroutines
- e7aa291: Start and stop the pm.Recipient goroutines
- a2ab469: Update a few debugging statements
- 332bb00: Fix data race in tests discovered via `go test -race`

Non-fixups:

- 4e46637: Always try to redeem winning tickets as long as they are valid based on the contract validation rules. Previously, if O received a winning ticket with an already revealed recipientRand or an invalid faceValue/winProb, O would not store/redeem the ticket

pm/sendermonitor.go
```diff
@@ -0,0 +1,264 @@
...
+type SenderMonitor interface {
+	// Start initiates the helper goroutines for the monitor
+	Start()
+
+	// Stop signals the monitor to exit gracefully
+	Stop()
+
+	// Redeemable returns a channel that a consumer can use to receive tickets that
+	// should be redeemed
+	Redeemable() chan *SignedTicket
```
kyriediculous: Can't we embed the `RedeemableEmitter` interface here instead?

cmd/livepeer/livepeer.go
```diff
@@ -52,9 +52,9 @@ var (
 	// gmPollingInterval is the interval at which to poll for gas price updates
 	gpmPollingInterval = 1 * time.Minute
 	// fmCleanupInterval is the interval at which to clean up cached max float values for PM senders
-	fmCleanupInterval = 1 * time.Minute
+	smCleanupInterval = 1 * time.Minute
 	// fmTTL is the time to live for cached max float values for PM senders (else they will be cleaned up)
```
j0sh: Comments referencing `fmCleanupInterval` and `fmTTL` need to be updated. It's probably OK to not mention the variable name here since that's more brittle.

pm/sendermonitor.go
```diff
@@ -0,0 +1,270 @@
...
+		return err
+	}
+
+	sm.senders[addr].queue.Add(ticket)
+
+	return nil
+}
+
+// maxFloat is a helper that returns the sender's max float as:
+// reserveAlloc - pendingAmount
```
j0sh, line 172: Maybe also mention in comments that for {maxFloat, ensureCache, cache} (and anything else?) that it's a precondition for the caller to hold the lock

pm/queue.go
```diff
@@ -0,0 +1,158 @@
...
+func (q *ticketQueue) Length() int32 {
+	return atomic.LoadInt32(&q.queueLen)
+}
+
+func (q *ticketQueue) startQueueLoop() {
+	for {
+		// Lock and wait until the queue is non-empty
+		q.cond.L.Lock()
+		for len(q.queue) == 0 {
+			q.cond.Wait()
```
j0sh, line 128: As much of a fan as I am of traditional synchronization primitives, I'm not really clear on why condvars and channels are mixed here. Since channels themselves block, could a buffered channel be used that blocks while the queue is empty? See how the [O transcode loop](https://github.com/livepeer/go-livepeer/blob/19526554b54ea7291692a6b7ad2b18cc7ff9b4d9/core/orchestrator.go#L422-L457) works for an example of this type of behavior.

j0sh (changes requested):
Looking pretty good! Just a few small comments.

core/orchestrator.go
```diff
@@ -120,6 +120,8 @@ func (orch *orchestrator) ProcessPayment(payment net.Payment, manifestID Manifes
 		CreationRoundBlockHash: ethcommon.BytesToHash(payment.ExpirationParams.CreationRoundBlockHash),
 	}
 
+	var firstErr error
```
kyriediculous, line 123: I'd have a slight preference for using a named return variable here, since error is the only returned value.
`(err error)` or `(firstError error)`. Named return variables are scoped to the function body and is a more idiomatic go approach ([according to effective go](https://golang.org/doc/effective_go.html#named-results))

darkdarkdragon: Personally I would prefer not to use name return variables at all - they makes code harder to read.

yondonfu: `firstErr` is actually removed in this [commit](https://github.com/livepeer/go-livepeer/pull/937/commits/d7dcbd3bddc86ddda37f269dfb0c5b26956abe6f) from a follow up PR. So, perhaps we can leave this as-is and continue any additional discussion on how the error is returned from `ProcessPayment()` based on the referenced commit in #937?

kyriediculous: Sure, didn't get to that one yet.

j0sh: As a general note on style, agreed with Ivan in that I prefer not to use named return variables since it's less explicit and requires holding more mental context for the function.

Given the emphasis that golang places on explicitness, I'm surprised this is even a feature in the language.

kyriediculous: Actually I find it to be more clear most times especially when using the go plugin for vscode that uses `godef` when browsing through code it actually helps as a form of documentation rather than having to go into the call chain to see what the context of the return value is. 

I'm pretty used to this in solidity as well. But perhaps a discussion to be had elsewhere 😄 .

pm/queue.go
yondonfu: Are you suggesting to use a channel to maintain the state of the queue instead of maintaining the state of the queue in a slice and using a condvar for concurrency safety? Whenever a max float update is signalled, we peek at the head of the queue and if the ticket at the head of the queue cannot be covered by the max float, we leave the queue as-is. I guess if we wanted to use a channel we could try to replicate this behavior by reading the ticket from the channel and if it cannot be covered by the max float we pass the ticket back into the channel. The one difference is that at the moment the queue remains unchanged when a max float is insufficient to cover the head of the queue while the described channel approach would pop the head of the queue and add it to the tail in the same scenario.

cmd/livepeer/livepeer.go
yondonfu: Fixed in d4fe922

pm/sendermonitor.go
yondonfu: Fixed in f1234f5

pm/sendermonitor.go
yondonfu: Fixed in 64eb640

pm/queue.go
yondonfu: @j0sh 

I ended up taking a different approach in f46945f than what is described in the above comment. I swapped the condvar for an unbuffered channel which the `(*queue).Add()` sends tickets into. The queue loop uses a select statement to receive on the unbuffered channel and if there is nothing to receive it will proceed to the rest of the loop. Let me know what you think!

j0sh: That looks closer to what I was thinking of. Will add some notes inline.

pm/queue.go
```diff
@@ -0,0 +1,148 @@
...
+
+// ticketQueue is a queue of winning tickets that are in line for redemption on-chain.
+// A recipient will have a ticketQueue per sender that it is actively receiving tickets from.
+// If a sender's max float is insufficient to cover the face value of a ticket it is added to the queue.
+// A ticket is pulled from the queue by the recipient when a sender has sufficient max float to cover
+// the next ticket in the queue
+//
+// Based off of: https://github.com/lightningnetwork/lnd/blob/master/htlcswitch/queue.go
+type ticketQueue struct {
+	queue []*SignedTicket
```
j0sh, line 25: Looks like the `queue` var can be made local to the for-loop now, but might not be necessary at all (see later comments). This goes back to my aversion around storing mutable state in long-lived structs.

pm/queue.go
```diff
@@ -0,0 +1,148 @@
...
+
+// Stop signals the ticketQueue to gracefully shutdown
+func (q *ticketQueue) Stop() {
+	close(q.quit)
+}
+
+// Add adds a ticket to the queue
+func (q *ticketQueue) Add(ticket *SignedTicket) {
+	select {
+	case q.in <- ticket:
```
j0sh: If there is no corresponding listener on the `in` channel (eg, waiting for the ticket to be redeemable), then the ticket would be dropped here. Is that the intended behavior? The `in` channel should be buffered to avoid this situation. But I may be mis-interpreting the intent of the code here.

pm/queue_test.go
```diff
@@ -0,0 +1,145 @@
...
+	redeemable []*SignedTicket
+	mu         sync.Mutex
+}
+
+// Redeemable returns the consumed redeemable tickets from a ticket queue
+func (qc *queueConsumer) Redeemable() []*SignedTicket {
+	qc.mu.Lock()
+	defer qc.mu.Unlock()
+
+	return qc.redeemable
```
j0sh, line 30: (NB: Just noticed this is a test file so the concern is not that important anymore, but just wanted to leave the original note below,  since I still wonder about returning slices from threadsafe functions.)

Can you confirm that this is indeed thread safe? Slice operations such as `append` can mutate underlying values (eg, outside the lock if the `Redeemable` caller chose to do so), but I'm unsure if returning a slice from an otherwise threadsafe function insulates us from that possibility (eg, by returning a copy).

pm/queue.go
```diff
@@ -0,0 +1,148 @@
...
+	return q.redeemable
+}
+
+// Length returns the current length of the queue
+func (q *ticketQueue) Length() int32 {
+	return atomic.LoadInt32(&q.queueLen)
+}
+
+func (q *ticketQueue) startQueueLoop() {
+	for {
```
j0sh, line 124: Isn't this basically a busy loop since nothing blocks? Trying to unpack my understanding here. It looks like there are two things happening within this for-loop:

1. Draining the `in` channel and adding to the queue
2. Checking for max float updates, and if there is one, sending the next ticket (head of queue) to the `redeemable` channel for redemption (`redeemable` is unbuffered; are we confident it will always have a listener?)

The queue may grow if there isn't a max float update at the time the queue is added.

Should these be separated into goroutines running their own blocking selects? This would mean locking around a (function-local) queue. Or maybe even the logic could be combined into one, where we wait for max float updates, then drain `q.in` directly into `q.redeemable` as necessary? Haven't thought about the approach too closely, so I might be off base. Something like this:

```golang
		select {
		case maxFloat := <-q.availableMaxFloat:
			if q.sufficientMaxFloat(maxFloat, nextTicket) {
				select {
				case q.redeemable <- q.in: // `q.in` should be buffered
					continue
				case <-q.quit:
					return
				default:
					continue
				}
			}
		case <-q.quit:
			return
```

BTW if my understanding is inaccurate (or even if it is accurate), it'd probably be helpful to have a brief description in comments of how the queue-loop works, to help folks reviewing the code in the future.

pm/queue.go
```diff
@@ -0,0 +1,148 @@
...
+		}
+
+		if len(q.queue) == 0 {
+			continue
+		}
+
+		nextTicket := q.queue[0]
+
+		select {
+		case maxFloat := <-q.availableMaxFloat:
```
j0sh: Semantically, it feels a little strange reading this: we're waiting on a channel, but `availableMaxFloat` implies the value is immediately available (when it may not be). Would renaming the channel to something like `maxFloatUpdate` work?

pm/queue.go
j0sh: BTW if it seems more straightforward to maintain the old approach with condvars then by all means stay with that.

(Offside rant, but this is one of the things I dislike about golang's concurrency constructs: the language doesn't give you a lot of tools to safely use those constructs. The experience of reasoning through their use is still a very raw process and not a huge step up from traditional primitives for non-trivial cases. It doesn't help that most of the behaviors are 'design choices' so they have to be learned and aren't as amenable to deduction from first principles.)

pm/queue.go
j0sh: Ignore this comment; I thought there was a `default` fall through here but was mistaken. This may block if there's no listener immediately though...

pm/queue.go
yondonfu: > Isn't this basically a busy loop since nothing blocks?

Yeah you're right - this is a busy loop. Previously, the condvar based architecture would have the for loop block until the queue is non-empty and only then would the loop continuously check for max float updates by receiving on `availableMaxFloat`. Now, the loop never blocks - we don't want to block when receiving on `q.in` because this would prevent the loop from receiving on `availableMaxFloat` when the queue is non-empty. We only want to block prior to receiving on `availableMaxFloat` if the queue is empty. I think this is a case where the condvar based architecture feels cleaner (not sure if there is a clean way to exactly replicate the described behavior with channels at the moment).

> The queue may grow if there isn't a max float update at the time the queue is added.

Yep or if winning tickets are received in quick succession such that the sender's max float is not sufficient to cover all of them immediately so they are instead put on the queue and redeemed one by one with the recipient making sure that the sender's max float is sufficient whenever it pops a ticket from the queue.

> redeemable is unbuffered; are we confident it will always have a listener?

The SenderMonitor will [create a queue](https://github.com/livepeer/go-livepeer/pull/926/commits/6bc826f45492a26edae26c61842c3377a875b052#diff-bc4d3ce81466bdbc2bcf8ecbe06525b8R198) for each tracked sender and start a consumer loop that listens on the queue's `redeemable` channel, let's call this `qRedeemable`, for tickets which are then passed into the SenderMonitor's single `redeemable` channel (tickets received from all queues managed by SenderMonitor are fanned into this single channel), let's call this `smRedeemable`, which the Recipient is listening on. The Recipient will redeem each ticket it receives and block until the transaction confirms on-chain - then, the Recipient will receive additional tickets on `smRedeemable`.

I think there is a minor problem here - the send for `qRedeemable` will block if there is a blocking send on `smRedeemable` which will be the case when the Recipient is waiting for a redemption transaction to confirm. This would be fine in a setting where the Recipient is interacting with a single sender. However, in this case, ticket redemptions for *other* senders would slow down the processing of a queue for a particular sender. For example, if the Recipient is waiting for the redemption for sender A and there are tickets waiting to be received on `smRedeemable` for senders B and C, then it takes longer to process the tickets in the queue for senders B and C. An improvement would be for the Recipient not to block while waiting for a redemption to confirm. 

> Should these be separated into goroutines running their own blocking selects? This would mean locking around a (function-local) queue.

If we're locking, I think I'd prefer just sticking with the condvar based architecture.

> Or maybe even the logic could be combined into one, where we wait for max float updates, then drain q.in directly into q.redeemable as necessary?

This was actually what I was thinking of in [this comment](https://github.com/livepeer/go-livepeer/pull/926#discussion_r297364022) describing the use of a buffered channel to maintain the state of the queue instead of relying on a slice. The one thing I don't like about it is that whenever the max float is *not* sufficient to cover the ticket at the head of the queue, we are left with two choices: drop the ticket or pass the ticket back into `q.in`. I don't think we want to drop tickets since its possible that the next max float update is sufficient to cover the ticket at the head of the queue. Passing the ticket back into `q.in` would not only require another select statement that receives on `q.quit` (to exit cleanly if needed), but the send would also need to be in a goroutine because we don't want to block or else the loop won't get to receive on `q.in`. The latter problem could be mitigated somewhat by a buffered channel. Maintaining the state of the queue in a slice that remains unchanged if the max float is insufficient to cover the ticket at the head of the queue feels simpler.

> BTW if my understanding is inaccurate (or even if it is accurate), it'd probably be helpful to have a brief description in comments of how the queue-loop works, to help folks reviewing the code in the future.

👍

pm/queue_test.go
yondonfu: > Can you confirm that this is indeed thread safe?

Good call. Looked into this a bit - this is *not* actually thread-safe for the reason that you mention.

See the following example: https://play.golang.org/p/M-CFYO4H3wu

EDIT: Updated the Go playground link to include comments on what is going on for those that are interested (new to me - I was not that familiar with the actual internals of slices before).

pm/queue.go
yondonfu: Yeah I think this is a case where after looking at a few other concurrency safe queues for inspiration the use of condvar felt a bit more intuitive than trying to use channels. Though I guess trying to figure out how to cleanly replace condvar here with channels is a useful thought exercise...

pm/queue.go
yondonfu: Yeah in the latest commit this would block if the queue loop is blocked while waiting on `availableMaxFloat` or blocked while sending into `redeemable`. Not that big of a problem if its called within a goroutine, but this could also be mitigated somewhat if `q.in` is a buffered channel or if we switch back to using a condvar as mentioned in other comments

pm/queue.go
j0sh: > The one thing I don't like about it is that whenever the max float is not sufficient to cover the ticket at the head of the queue, we are left with two choices: drop the ticket or pass the ticket back into q.in

Would checking `sufficientMaxFloat` prior to checking `q.in` address that? As in the example above, `q.in` is nonblocking (via the `default` case), so if empty, we'd continue waiting for the next max float update.

Also, reviewing [recipient.redeemWinningTicket](https://github.com/livepeer/go-livepeer/blob/f46945f0864e52f6ed87c20a1ac668a918a834c7/pm/recipient.go#L323-L333), it does seem that we have the potential to re-queue if the `redeemWinningTicket` function detects the float is insufficient.

Noticed this when I was trying to figure out the "chicken-egg" problem of actually submitting the first ticket from an empty queue, since there'd be no pending transaction waiting to be signaled, so theoretically tickets could be stuck in the queue indefinitely. I see how we're handling that situation now, but it seems like something that could be accidentally triggered by another developer that's unfamiliar with the code. I'd have to think some more about how to future-proof that sort of behavior.

It'd be good to note in the comments that the queue has a precondition of expecting a pending blockchain transaction per element, and only moves forward as the transaction is confirmed.

A more robust long-term solution would be to update the DB with the ticket submission status, and have the queue check the DB for the oldest non-submitted ticket, rather than using a channel. That gives us persistence across reboots and some more options around failed transactions (where tickets in the queue would be stuck indefinitely).

yondonfu: > Would checking sufficientMaxFloat prior to checking q.in address that?

I don't think so because checking `sufficientMaxFloat` requires receiving the ticket from `q.in` first (or else where does `nextTicket` come from). So, the code snippet you provided where `q.in` is drained directly into `q.redeemable` would have to look something like this:

```golang
		select {
		case maxFloat := <-q.availableMaxFloat:
			select {
			case nextTicket := <-q.in:
				if q.sufficientMaxFloat(maxFloat, nextTicket) {
					select {
					case q.redeemable <- nextTicket:
					case <-q.quit:
						return
					}
				} else {
					// Send `nextTicket` back into `q.in`
					// to add it back to the queue?
				}
			case <-q.quit:
				return
			default:
			}
		case <-q.quit:
			return
		default:
		}
```

> It'd be good to note in the comments that the queue has a precondition of expecting a pending blockchain transaction per element, and only moves forward as the transaction is confirmed.

👍 

> A more robust long-term solution would be to update the DB with the ticket submission status, and have the queue check the DB for the oldest non-submitted ticket, rather than using a channel. That gives us persistence across reboots and some more options around failed transactions (where tickets in the queue would be stuck indefinitely).

Ah yes, excellent point. We're already storing each winning ticket in the DB before we redeem it, so we're already well set up to do this. I'll create a GH issue to track this and we can follow up in a separate PR if that works for you.

j0sh: > checking sufficientMaxFloat requires receiving the ticket from q.in first (or else where does nextTicket come from)

You're right, I totally missed that.

Could we skip the sufficiency check here and rely on the check that's [within `redeemWinningTicket`](https://github.com/livepeer/go-livepeer/blob/f46945f0864e52f6ed87c20a1ac668a918a834c7/pm/recipient.go#L318-L333) ? Seems like that would also lead to re-queuing if necessary. That maybe comes at the expense of less "defense in depth" but it seems the code would be cleaner overall. If the flow is too hard to reason about right now, then whatever approach you're thinking of is fine (condvars?).

yondonfu: > Could we skip the sufficiency check here and rely on the check that's within redeemWinningTicket ?

That check's purpose is a bit different. The check in `redeemWinningTicket` is to determine if we should queue a ticket when the max float is insufficient. The check here is to determine if we should pop a ticket i.e. if the most up to date max float is sufficient to cover the next ticket from the queue. We need some way to determine when a ticket should come off the queue.

> If the flow is too hard to reason about right now, then whatever approach you're thinking of is fine (condvars?).

My current preference is to revert back to using a condvar in this PR for now and then revisit this workflow separately using the DB backed approach that pulls the oldest non-confirmed winning ticket that you [mentioned](https://github.com/livepeer/go-livepeer/pull/926#discussion_r297903557) which could make a lot of this design discussion around channels vs. condvars moot.

pm/queue.go
yondonfu: Fixed in 1d23297

pm/queue.go
yondonfu: Rolled back the commit that replaced condvar with `q.in` so we're using a condvar again. Also, added some more clarifying comments in 1d23297

j0sh (approved):
🚢

PR #928, opened 2019-06-20
Return go compiler version in `/status` endpoint

darkdarkdragon:
Return go compiler version used to compile node
in `/status` endpoint.
Will show this info in `test-harness test` command.

- 
- 
- 

manually

**Checklist:**

- [] All tests in `./test.sh` pass

### DIFF
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -130,6 +130,8 @@ func main() {
 	if *version {
 		fmt.Println("Livepeer Node Version: " + core.LivepeerVersion)
 		fmt.Printf("Compiler version: %s %s\n", runtime.Compiler, runtime.Version())
+		fmt.Printf("Architecture: %s\n", runtime.GOARCH)
+		fmt.Printf("Operating system: %s\n", runtime.GOOS)
 		return
 	}
 
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -129,7 +129,7 @@ func main() {
 
 	if *version {
 		fmt.Println("Livepeer Node Version: " + core.LivepeerVersion)
-		fmt.Printf("Compiler version: %s %s\n", runtime.Compiler, runtime.Version())
+		fmt.Printf("Golang runtime version: %s %s\n", runtime.Compiler, runtime.Version())
 		fmt.Printf("Architecture: %s\n", runtime.GOARCH)
 		fmt.Printf("Operating system: %s\n", runtime.GOOS)
 		return
--- a/server/cliserver_test.go
+++ b/server/cliserver_test.go
@@ -40,8 +40,8 @@ func TestGetStatus(t *testing.T) {
 	defer res.Body.Close()
 	body, err := ioutil.ReadAll(res.Body)
 	req.Nil(err)
-	expected := fmt.Sprintf(`{"Manifests":{},"OrchestratorPool":[],"Version":"undefined","GORuntimeVersion":"%s","RegisteredTranscodersNumber":1,"RegisteredTranscoders":[{"Address":"TestAddress","Capacity":5}],"LocalTranscoding":false}`,
-		runtime.Version())
+	expected := fmt.Sprintf(`{"Manifests":{},"OrchestratorPool":[],"Version":"undefined","GORuntimeVersion":"%s","GOArch":"%s","GOOS":"%s","RegisteredTranscodersNumber":1,"RegisteredTranscoders":[{"Address":"TestAddress","Capacity":5}],"LocalTranscoding":false}`,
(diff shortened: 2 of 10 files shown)


server/cliserver_test.go
```diff
@@ -34,6 +35,6 @@ func TestGetStatus(t *testing.T) {
 	defer res.Body.Close()
 	body, err := ioutil.ReadAll(res.Body)
 	req.Nil(err)
-	assert.Equal(`{"Manifests":{},"OrchestratorPool":[],"Version":"undefined","RegisteredTranscodersNumber":1,"RegisteredTranscoders":[{"Address":"TestAddress","Capacity":5}],"LocalTranscoding":false}`,
+	assert.Equal(`{"Manifests":{},"OrchestratorPool":[],"Version":"undefined","CompilerVersion":"`+runtime.Version()+`","RegisteredTranscodersNumber":1,"RegisteredTranscoders":[{"Address":"TestAddress","Capacity":5}],"LocalTranscoding":false}`,
```
mkrufky: I would suggest using `fmt.Sprintf` here rather than the nested backticks.  It would expand to more lines of code, but might be easier to read and edit later on.  <-- this is just a suggestion.  I think this is OK as-is for now, since it's a single JSON block with only one generated value within a single test.

I think this should be called `RuntimeVersion` instead of `CompilerVersion`.

Golang uses the `gcc` compiler, and it might be any version.  I'm not sure if you're looking to track that, or if its easily discoverable.

Do you think we should also include `sys.GOARCH` and `sys.GOOS` for architecture target and running OS?

darkdarkdragon: > I would suggest using fmt.Sprintf 

fixed

> I think this should be called RuntimeVersion instead of CompilerVersion

renamed to `GORuntimeVersion`

> Golang uses the gcc compiler, and it might be any version. I'm not sure if you're looking to track that, or if its easily discoverable.

> Golang uses the gcc compiler, and it might be any version. I'm not sure if you're looking to track that, or if its easily discoverable.

`runtime.Compiler` returns just `gc` - not very useful. 

> Do you think we should also include sys.GOARCH and sys.GOOS for architecture target and running OS?

Not sure if Livepeer node will every be running on something other than `amd64` (on phones? RaspberryPi?), but to be future proof, added GOARCH and GOOS to output.

There was two reasons to adding all this:
1. For future bug-reports from users.
2. For own testing - in test harness I often use image built in Docker Hub, and I want to check with what version of `golang` it was built - we already once hit bug that was manifested itself only in `go1.11`, but not in `go1.10` and not in `go1.12`.

darkdarkdragon:
@mk-livepeer @mkrufky I've made changes, it can be reviewed again.

cmd/livepeer/livepeer.go
```diff
@@ -130,6 +130,8 @@ func main() {
 	if *version {
 		fmt.Println("Livepeer Node Version: " + core.LivepeerVersion)
 		fmt.Printf("Compiler version: %s %s\n", runtime.Compiler, runtime.Version())
```
mkrufky: Maybe we should change the logging here, too.   "Compiler version" --> "Golang runtime version"

mkrufky (approved):
I approve, after changing the `fmt.Printf` from "Compiler version" to "GO runtime version"

cmd/livepeer/livepeer.go
j0sh: s/go/golang ?

server/cliserver_test.go
```diff
@@ -34,6 +40,49 @@ func TestGetStatus(t *testing.T) {
 	defer res.Body.Close()
 	body, err := ioutil.ReadAll(res.Body)
 	req.Nil(err)
-	assert.Equal(`{"Manifests":{},"OrchestratorPool":[],"Version":"undefined","RegisteredTranscodersNumber":1,"RegisteredTranscoders":[{"Address":"TestAddress","Capacity":5}],"LocalTranscoding":false}`,
-		string(body))
+	expected := fmt.Sprintf(`{"Manifests":{},"OrchestratorPool":[],"Version":"undefined","GORuntimeVersion":"%s","GOArch":"%s","GOOS":"%s","RegisteredTranscodersNumber":1,"RegisteredTranscoders":[{"Address":"TestAddress","Capacity":5}],"LocalTranscoding":false}`,
```
mkrufky: As per @j0sh 's suggestion, we should also change "GORuntimeVersion" to "GolangRuntimeVersion"

darkdarkdragon: fixed

cmd/livepeer/livepeer.go

darkdarkdragon:
@j0sh @mkrufky done

j0sh (approved):

PR #933, opened 2019-06-21
cmd, discovery, server: allow B to set MaxBroadcastPrice to filter acceptable O's stored in cache

### LINKED ISSUE #906: MVP B price per pixel preference setting
B should be able to set a maximum price per pixel so that it will refuse to work with Os that advertise a higher price per pixel. The node already defines a [BroadcastPrice](https://github.com/livepeer/go-livepeer/blob/master/server/mediaserver.go#L57) which is configurable via the CLI wizard. For this MVP price preference setting we can just update that variable to indicate a price per pixel in same format as described in #891.

Then, when [creating broadcast sessions](https://github.com/livepeer/go-livepeer/blob/master/server/broadcast.go#L176) for a list of orchestrators, we can exclude orchestrators that advertise an unacceptable price per pixel.

kyriediculous:
Allows a broadcaster to set a maximum price preference `BroadcastPrice *MaxBroadcastPrice` (not required) in the same format as #916 either by specifying the flags `maxPricePerUnit` and `pixelsPerUnit` on node startup, or through the CLI. 

When `orchestratorPool.getOrchestrators` is called on the orchestrators cached in the DBcache, it will filter orchestrators based on this max price setting (if set). 

- Added `maxPricePerUnit` flag to `cmd/livepeer.go`
- Added setting `maxPricePerUnit` and `pixelsPerUnit` through broadcaster CLI config 
- Added a predicate function to `orchestratorPool` struct
- Added a new constructor to create an `orchestratorPool` with a predicate function `NewOrchestratorPoolWithPred`
- Adjusted `DBOrchestratorPoolCache.getOrchestrators` to initiate `orchestratorPool` with a predicate function through the new constructor `NewOrchestratorPoolWithPred`. The predicate function returns true when B's `MaxBroadcastPrice` isn't set, or is greater than Orchestrators `PriceInfo`
- Adjusted exported variable in `server` package `BroadcastPrice` to be a struct of type `MaxBroadcastPrice` consisting out of a `*big.Rat` and `sync.RWMutex` 

Added unit tests (80.3 % coverage for `discovery` package)
Added test to `./test_args.sh` for when flags are set , but incorrectly provided 

Fixes #906 

**Checklist:**

### DIFF
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -103,8 +103,11 @@ func main() {
 	initializeRound := flag.Bool("initializeRound", false, "Set to true if running as a transcoder and the node should automatically initialize new rounds")
 	ticketEV := flag.String("ticketEV", "1000000000", "The expected value for PM tickets")
 
-	// Price Info
+	// Orchestrator base pricing info
 	pricePerUnit := flag.Int("pricePerUnit", 0, "The price per 'pixelsPerUnit' amount pixels")
+	// Broadcaster max acceptable price
+	maxPricePerUnit := flag.Int("maxPricePerUnit", 0, "The maximum transcoding price (in wei) per 'pixelsPerUnit' a broadcaster is willing to accept. If not set explicitly, broadcaster is willing to accept ANY price")
+	// Unit of pixels for both O's basePriceInfo and B's MaxBroadcastPrice
 	pixelsPerUnit := flag.Int("pixelsPerUnit", 1, "Amount of pixels per unit. Set to '> 1' to have smaller price granularity than 1 wei / pixel")
 
 	// Metrics & logging:
@@ -321,13 +324,11 @@ func main() {
 			// Set price per pixel base info
 			if *pixelsPerUnit <= 0 {
 				// Can't divide by 0
-				glog.Fatalf("The amount of pixels per unit must be greater than 0, provided %d instead\n", *pixelsPerUnit)
-				return
+				panic(fmt.Errorf("The amount of pixels per unit must be greater than 0, provided %d instead\n", *pixelsPerUnit))
 			}
 			if *pricePerUnit <= 0 {
 				// Prevent orchestrator from unknowingly provide free transcoding
-				glog.Fatalf("Price per unit of pixels must be greater than 0, provided %d instead\n", *pricePerUnit)
-				return
+				panic(fmt.Errorf("Price per unit of pixels must be greater than 0, provided %d instead\n", *pricePerUnit))
 			}
 			n.PriceInfo = big.NewRat(int64(*pricePerUnit), int64(*pixelsPerUnit))
 			glog.Infof("Price: %d wei for %d pixels\n ", *pricePerUnit, *pixelsPerUnit)
@@ -387,6 +388,17 @@ func main() {
 			// TODO: Initialize Sender with an implementation
 			// of RoundsManager that reads from a cache
 			n.Sender = pm.NewSender(n.Eth, n.Eth)
+
+			if *pixelsPerUnit <= 0 {
+				// Can't divide by 0
+				panic(fmt.Errorf("The amount of pixels per unit must be greater than 0, provided %d instead\n", *pixelsPerUnit))
+			}
+			if *maxPricePerUnit > 0 {
+				server.BroadcastCfg.SetMaxPrice(big.NewRat(int64(*maxPricePerUnit), int64(*pixelsPerUnit)))
+			} else {
+				glog.Infof("Maximum transcoding price per pixel is not greater than 0: %v, broadcaster is currently set to accept ANY price.\n", *maxPricePerUnit)
+				glog.Infoln("To update the broadcaster's maximum acceptable transcoding price per pixel, use the CLI or restart the broadcaster with the appropriate 'maxPricePerUnit' and 'pixelsPerUnit' values")
+			}
 		}
 
 		// Start services
--- a/cmd/livepeer_cli/wizard_broadcast.go
+++ b/cmd/livepeer_cli/wizard_broadcast.go
@@ -51,8 +51,14 @@ func (w *wizard) allTranscodingOptions() map[int]string {
 }
 
 func (w *wizard) setBroadcastConfig() {
-	fmt.Printf("Enter broadcast max price per segment in Wei - ")
-	maxPricePerSegment := w.readBigInt()
+	fmt.Printf("Enter a maximum transcoding price per pixel, in wei per pixels (pricePerUnit / pixelsPerUnit).\n")
+	fmt.Printf("eg. 1 wei / 10 pixels = 0,1 wei per pixel \n")
+	fmt.Printf("\n")
+	fmt.Printf("Enter amount of pixels that make up a single unit (default: 1 pixel) - ")
+	pixelsPerUnit := w.readDefaultInt(1)
+	fmt.Printf("\n")
+	fmt.Printf("Enter the maximum price to pay for %d pixels in Wei (required) - ", pixelsPerUnit)
+	maxPricePerUnit := w.readDefaultInt(0)
 
 	opts := w.allTranscodingOptions()
 	if opts == nil {
@@ -69,7 +75,8 @@ func (w *wizard) setBroadcastConfig() {
 	}
 
 	val := url.Values{
-		"maxPricePerSegment": {fmt.Sprintf("%v", maxPricePerSegment.String())},
+		"pixelsPerUnit":      {fmt.Sprintf("%v", strconv.Itoa(pixelsPerUnit))},
+		"maxPricePerUnit":    {fmt.Sprintf("%v", strconv.Itoa(maxPricePerUnit))},
 		"transcodingOptions": {fmt.Sprintf("%v", transOpts)},
 	}
 
--- a/discovery/discovery.go
+++ b/discovery/discovery.go
@@ -24,6 +24,7 @@ var serverGetOrchInfo = server.GetOrchestratorInfo
 type orchestratorPool struct {
 	uris  []*url.URL
 	bcast server.Broadcaster
+	pred  func(info *net.OrchestratorInfo) bool
 }
 
 var perm = func(len int) []int { return rand.Perm(len) }
@@ -55,6 +56,17 @@ func NewOrchestratorPool(node *core.LivepeerNode, addresses []string) *orchestra
 
 	bcast := core.NewBroadcaster(node)
 	return &orchestratorPool{bcast: bcast, uris: randomizedUris}
+}
+
+func NewOrchestratorPoolWithPred(node *core.LivepeerNode, addresses []string, pred func(*net.OrchestratorInfo) bool) *orchestratorPool {
+	// if livepeer running in offchain mode, return nil
+	if node.Eth == nil {
+		glog.Error("Could not refresh DB list of orchestrators: LivepeerNode nil")
+		return nil
+	}
+	pool := NewOrchestratorPool(node, addresses)
+	pool.pred = pred
+	return pool
 }
 
 func NewOnchainOrchestratorPool(node *core.LivepeerNode) *orchestratorPool {
@@ -97,7 +109,7 @@ func (o *orchestratorPool) GetOrchestrators(numOrchestrators int) ([]*net.Orches
 		respLock.Lock()
 		defer respLock.Unlock()
 		numResp++
-		if err == nil {
+		if err == nil && (o.pred == nil || o.pred(info)) {
 			orchInfos = append(orchInfos, info)
 			numSuccessResp++
 		} else if monitor.Enabled {
--- a/discovery/discovery_test.go
+++ b/discovery/discovery_test.go
@@ -2,14 +2,17 @@ package discovery
 
 import (
 	"context"
+	"math/big"
 	"math/rand"
 	"net/url"
 	"runtime"
+	"strconv"
 	"sync"
 	"testing"
 	"time"
 
 	ethcommon "github.com/ethereum/go-ethereum/common"
+	"github.com/golang/glog"
 	"github.com/livepeer/go-livepeer/common"
 	"github.com/livepeer/go-livepeer/core"
 	"github.com/livepeer/go-livepeer/eth"
@@ -52,6 +55,33 @@ func StubOrchestrators(addresses []string) []*lpTypes.Transcoder {
 	return orchestrators
 }
 
+func TestNewOrchestratorPoolWithPred_NilEthClient_ReturnsNil_LogsError(t *testing.T) {
+	node, _ := core.NewLivepeerNode(nil, "", nil)
+	errorLogsBefore := glog.Stats.Error.Lines()
+	pool := NewOrchestratorPoolWithPred(node, nil, nil)
+	errorLogsAfter := glog.Stats.Error.Lines()
+	assert.NotZero(t, errorLogsAfter-errorLogsBefore)
+	assert.Nil(t, pool)
+}
+
+func TestNewOnchainOrchestratorPool_NilEthClient_ReturnsNil_LogsError(t *testing.T) {
+	node, _ := core.NewLivepeerNode(nil, "", nil)
+	errorLogsBefore := glog.Stats.Error.Lines()
+	pool := NewOnchainOrchestratorPool(node)
+	errorLogsAfter := glog.Stats.Error.Lines()
+	assert.NotZero(t, errorLogsAfter-errorLogsBefore)
+	assert.Nil(t, pool)
+}
+
+func TestNewDBOrchestratorPoolCache_NilEthClient_ReturnsNil_LogsError(t *testing.T) {
+	node, _ := core.NewLivepeerNode(nil, "", nil)
+	errorLogsBefore := glog.Stats.Error.Lines()
+	poolCache := NewDBOrchestratorPoolCache(node)
+	errorLogsAfter := glog.Stats.Error.Lines()
+	assert.NotZero(t, errorLogsAfter-errorLogsBefore)
+	assert.Nil(t, poolCache)
+}
+
 func TestDeadLock(t *testing.T) {
 	gmp := runtime.GOMAXPROCS(50)
 	defer runtime.GOMAXPROCS(gmp)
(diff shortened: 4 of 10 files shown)


cmd/livepeer/livepeer.go
```diff
@@ -385,6 +388,18 @@ func main() {
...
 			// of RoundsManager that reads from a cache
 			n.Sender = pm.NewSender(n.Eth, n.Eth)
+
+			if *pixelsPerUnit <= 0 {
+				// Can't divide by 0
+				glog.Fatalf("The amount of pixels per unit must be greater than 0, provided %d instead\n", *pixelsPerUnit)
+				return
+			}
+			if *maxPricePerUnit > 0 {
+				server.BroadcastPrice.Price = big.NewRat(int64(*maxPricePerUnit), int64(*pixelsPerUnit))
```
yondonfu: Lock/unlock here while setting `BroadcastPrice`?

cmd/livepeer/livepeer.go
```diff
@@ -385,6 +388,18 @@ func main() {
...
+			} else {
+				glog.Infof("maxPricePerUnit is not greater than 0: %v, Broadcaster is currently set to accept ANY price.\n", *maxPricePerUnit)
+				glog.Infoln("To update MaxBroadcastPrice use the CLI or restart the broadcaster with the appropriate maxPricePerUnit and pixelsPerUnit values")
```
yondonfu: Since the variable `MaxBroadcastPrice` isn't actually exposed to the user, I suggest describing the variable here instead of explicitly referencing the variable i.e. "To update the max broadcast price..."

cmd/livepeer/livepeer.go
```diff
@@ -385,6 +388,18 @@ func main() {
...
+
+			if *pixelsPerUnit <= 0 {
+				// Can't divide by 0
+				glog.Fatalf("The amount of pixels per unit must be greater than 0, provided %d instead\n", *pixelsPerUnit)
+				return
+			}
+			if *maxPricePerUnit > 0 {
+				server.BroadcastPrice.Price = big.NewRat(int64(*maxPricePerUnit), int64(*pixelsPerUnit))
+			} else {
+				glog.Infof("maxPricePerUnit is not greater than 0: %v, Broadcaster is currently set to accept ANY price.\n", *maxPricePerUnit)
```
yondonfu: Style nit: I think we've mainly used "broadcaster"/"orchestrator" instead of "Broadcaster"/"Orchestrator" within in-line comments.

discovery/discovery.go
```diff
@@ -97,7 +109,10 @@ func (o *orchestratorPool) GetOrchestrators(numOrchestrators int) ([]*net.Orches
 		respLock.Lock()
 		defer respLock.Unlock()
 		numResp++
-		if err == nil {
+		if err == nil && (o.pred != nil && o.pred(info)) {
```
yondonfu: If we change this conditional to:

```
if err == nil && (o.pred == nil || o.pred(info))
```

we can remove the `else if` block that follows.

server/mediaserver.go
```diff
@@ -54,7 +54,15 @@ const StreamKeyBytes = 6
 const SegLen = 2 * time.Second
 const BroadcastRetry = 15 * time.Second
 
-var BroadcastPrice = big.NewInt(1)
+type MaxBroadcastPrice struct {
```
yondonfu: Since we'll eventually need to protect `BroadcastJobVideoProfiles` with a lock as well (it is modifiable from the CLI), what do you think about using a `BroadcastConfig` struct here instead?

i.e.

```
type BroadcastConfig struct {
    maxPrice *big.Rat
    
    mu sync.RWMutex
}
```

Then, we can attach methods to `BroadcastConfig` that handle locking so that we don't need to remember to lock/unlock when we modify the config values elsewhere:

```
func (bc *BroadcastConfig) MaxPrice() *big.Rat {
    bc.mu.RLock()
    defer bc.mu.RUnlock()
    
    return bc.MaxPrice
}

func (bc *BroadcastConfig) SetMaxPrice(price *big.Rat) {
    bc.mu.Lock()
    defer bc.mu.Unlock()
    
    return bc.MaxPrice
}
```

We could skip bringing `BroadcastJobVideoProfiles` into the `BroadcastConfig` struct for now and do that in a separate PR.

server/webserver.go
```diff
@@ -105,10 +117,13 @@ func (s *LivepeerServer) cliWebServerHandlers(bindAddr string) *http.ServeMux {
 			return
 		}
 
-		BroadcastPrice = price
-		BroadcastJobVideoProfiles = profiles
-
-		glog.Infof("Transcode Job Price: %v, Transcode Job Type: %v", BroadcastPrice, BroadcastJobVideoProfiles)
+		BroadcastPrice.Price = price
```
yondonfu: Need to lock/unlock here

server/webserver.go
```diff
@@ -105,10 +117,13 @@ func (s *LivepeerServer) cliWebServerHandlers(bindAddr string) *http.ServeMux {
 			return
 		}
 
-		BroadcastPrice = price
-		BroadcastJobVideoProfiles = profiles
```
yondonfu: We should keep this line since it sets the desired video profiles to be used for transcoding.

server/webserver.go
```diff
@@ -105,10 +117,13 @@ func (s *LivepeerServer) cliWebServerHandlers(bindAddr string) *http.ServeMux {
...
-
-		glog.Infof("Transcode Job Price: %v, Transcode Job Type: %v", BroadcastPrice, BroadcastJobVideoProfiles)
+		BroadcastPrice.Price = price
+		if BroadcastPrice.Price != nil {
+			glog.Infof("Max Transcode Price: %d per %q pixels\n", pr, px)
```
yondonfu: I think we should try to stay consistent with language i.e. we use "max broadcast price" elsewhere, but we use "max transcode price" here. Since we've already using "max broadcast price" elsewhere, I'd lean towards using that for now and if we decide later on that we prefer alternate terminology we can return and do a find and replace.

cmd/livepeer/livepeer.go
```diff
@@ -385,6 +388,18 @@ func main() {
 			// TODO: Initialize Sender with an implementation
 			// of RoundsManager that reads from a cache
 			n.Sender = pm.NewSender(n.Eth, n.Eth)
+
+			if *pixelsPerUnit <= 0 {
+				// Can't divide by 0
+				glog.Fatalf("The amount of pixels per unit must be greater than 0, provided %d instead\n", *pixelsPerUnit)
```
yondonfu: I recommend using `glog.Errorf()` and a `return` here instead of using `glog.Fatalf()`.

Since `glog.Fatalf()` exits the program immediately (I believe it is equivalent to `os.Exit(1)`), if there are any cleanup statements that were previously wrapped with a `defer`, they will not be run upon program exit.

server/webserver.go
```diff
@@ -76,17 +76,29 @@ func (s *LivepeerServer) cliWebServerHandlers(bindAddr string) *http.ServeMux {
...
+
+		pixelsPerUnit := r.FormValue("pixelsPerUnit")
+		px, err := strconv.ParseInt(pixelsPerUnit, 10, 64)
 		if err != nil {
-			glog.Error(err)
+			glog.Errorf("Error converting string to int64: %v\n", err)
+			return
+		}
+		if px <= 0 {
+			glog.Errorf("pixels per unit must be greater than 0, provided %d,\n", px)
```
yondonfu: Extraneous trailing comma at the end?

cmd/livepeer_cli/wizard_broadcast.go
```diff
@@ -51,8 +51,15 @@ func (w *wizard) allTranscodingOptions() map[int]string {
 }
 
 func (w *wizard) setBroadcastConfig() {
-	fmt.Printf("Enter broadcast max price per segment in Wei - ")
-	maxPricePerSegment := w.readBigInt()
+	fmt.Printf("Enter a broadcast max price per pixels transcoded in Wei.\n")
```
yondonfu: Should we use "wei" instead of "Wei" since we use "wei" elsewhere (i.e. later in this prompt)?

discovery/discovery_test.go
```diff
@@ -222,3 +328,187 @@ func TestNewOrchestratorPoolCache_GivenListOfOrchs_CreatesPoolCacheCorrectly(t *
...
+	// Set MaxBroadcastPrice higher than PriceInfo , should return true
+	server.BroadcastPrice.Price = big.NewRat(10, 1)
+	assert.True(t, pool.pred(oInfo))
+
+	// Set MaxBroadcastPrice lower than PriceInfo, should return false
+	server.BroadcastPrice.Price = big.NewRat(1, 1)
+	assert.False(t, pool.pred(oInfo))
+}
+
+func TestCachedPool_AllOrchestratorsTooExpensive_ReturnsEmptyList(t *testing.T) {
```
yondonfu, line 370: So, we have a test for the following:

- All Os too expensive
- Max broadcast price is not set

It would be nice to also have a test that checks for N out of 50 Os meeting the max price requirements.

yondonfu (changes requested):

discovery/discovery.go
kyriediculous: I tried that and ran into a dead lock on the `TestDeadLock` test when pred is nil 🤔 . let me double check if I missed something there.

Oh right, I'm checking for not nil. My bad.

discovery/discovery_test.go
kyriediculous: Yeah I realized that but had no immeadiate idea on how to implement it cleanly. Any suggestions?

cmd/livepeer/livepeer.go
kyriediculous: Is that needed here considering this is on node startup? (I'll add it, I'm just wondering)

server/webserver.go
kyriediculous: Perhaps I should fix this now instead of later and use "Max transcode job price" because it conveys better what it means?

cmd/livepeer/livepeer.go
kyriediculous: Good point, I did not realize that. It is indeed equivalent to `os.Exit(n)`. I will have to add this to my hackMD document on deferring. 

I will have to make this change for orchestrator setting base price as well.

cmd/livepeer/livepeer.go
kyriediculous: Entirely correct, I'll pay more attention to consistency in comments and logging messages.

cmd/livepeer/livepeer.go
kyriediculous: Entirely correct. I should replace fatals for orchestrator setting basePrice as well then probably?

server/mediaserver.go
kyriediculous: I like this.

discovery/discovery_test.go
yondonfu: The stubbed [serverGetOrchInfo()](https://github.com/livepeer/go-livepeer/pull/933/files#diff-13f8a7f541eee06b276d455b486bb2efR381) function in the tests accepts a `orchestratorServer *url.URL` argument. We could modify the stubbed function to return an unacceptable `PriceInfo` message for different `orchestratorServer` values.

Ex.

```
// Initialize with URLs of Os that should return an unacceptable PriceInfo message
unacceptablePriceURLs := ...

serverGetOrchInfo = func(ctx context.Context, bcast server.Broadcaster, orchestratorServer *url.URL) (*net.OrchestratorInfo, error) {
    ...

    if _, ok := unacceptablePriceURLs[orchestratorServer]; ok {
        // Return an unacceptable PriceInfo message
    }

    // Return an acceptable PriceInfo message
}
```

server/webserver.go
yondonfu: I'm on board with "Max transcode price"

cmd/livepeer/livepeer.go
yondonfu: Not absolutely necessary, but if we end up using helper methods attached to a `BroadcastConfig` struct, we'll end up doing so anyway

server/mediaserver.go
kyriediculous: I moved this to `server/broadcast.go` , it makes more sense to me there filename wise than in mediaserver.

cmd/livepeer/livepeer.go
kyriediculous: Now in `./test_args.sh` our exit code checking statements don't work anymore, e.g. : 

```
# Orchestrator needs to explicitly set PricePerUnit otherwise it will default to 0 resulting in a fatal error
res=0
./livepeer -orchestrator -serviceAddr 127.0.0.1:8935 -transcoder -network rinkeby $ETH_ARGS || res=$?
[ $res -ne 0 ]
```
Checking for $res -eq 0 seems trivial if we are not checking the actual error being logged. 

It seems like `panic` actually executes the deferred functions normally 

> Panic is a built-in function that stops the ordinary flow of control and begins panicking. When the function F calls panic, execution of F stops, any deferred functions in F are executed normally, and then F returns to its caller.

So I'm using that for now, please let me know if that's not a good idea.

server/broadcast.go
```diff
@@ -23,6 +23,27 @@ import (
 	"github.com/livepeer/lpms/stream"
 )
 
+var BroadcastCfg = &BroadcastConfig{
+	mu: &sync.RWMutex{},
```
yondonfu: The zero value of `sync.RWMutex` is an [unlocked mutex](https://golang.org/pkg/sync/#RWMutex) so we can shorten this line to:

```
```

server/webserver.go
```diff
@@ -132,10 +132,10 @@ func (s *LivepeerServer) cliWebServerHandlers(bindAddr string) *http.ServeMux {
 			pNames = append(pNames, p.Name)
 		}
 		config := struct {
-			MaxBroadcastPrice  *MaxBroadcastPrice
+			BroadcastConfig    *BroadcastConfig
```
yondonfu: This will return the entire `BroadcastConfig` struct nested within the `config` struct, but we want to return the `config` struct just with the max price and transcoding options fields.

Later on, when we bring `BroadcastJobVideoProfiles` into the `BroadcastConfig` struct as well, we can directly respond with the `BroadcastConfig` struct here instead of creating the anonymous `config` struct.

server/webserver.go
```diff
@@ -116,12 +116,12 @@ func (s *LivepeerServer) cliWebServerHandlers(bindAddr string) *http.ServeMux {
 			glog.Errorf("Invalid transcoding options: %v", transcodingOptions)
 			return
 		}
-
-		BroadcastPrice.Price = price
-		if BroadcastPrice.Price != nil {
-			glog.Infof("Max Transcode Price: %d per %q pixels\n", pr, px)
+		BroadcastCfg.SetMaxPrice(price)
+		BroadcastJobVideoProfiles = profiles
+		if p := BroadcastCfg.MaxPrice(); p != nil {
```
yondonfu: I think we can avoid this `BroadcastCfg.MaxPrice()` call and just check `if price != nil`

cmd/livepeer/livepeer.go
```diff
@@ -389,7 +389,8 @@ func main() {
 
 			if *pixelsPerUnit <= 0 {
 				// Can't divide by 0
-				panic(errors.New(fmt.Sprintf("The amount of pixels per unit must be greater than 0, provided %d instead\n", *pixelsPerUnit)))
+
```
yondonfu: Can remove this extra new line

cmd/livepeer/livepeer.go
yondonfu: While `panic` does get the job done here, I'm not a big fan of the output displayed to the user:

Ex.

```
panic: Price per unit of pixels must be greater than 0, provided -5 instead


goroutine 1 [running]:
main.main()
	/Users/yondonfu/Development/go/src/github.com/livepeer/go-livepeer/cmd/livepeer/livepeer.go:329 +0x559f
```

IMO the goroutine stack trace makes it seem like something went horribly wrong in the program execution in a similar manner as panics due to nil pointer references when in reality the source of the issue is user input as opposed to programmer error (anecdotally I am not accustomed to seeing this type of error format displayed when running CLI based programs).

An alternative could be to wrap everything that is currently in the `main()` function in its own helper function which is then invoked by the `main()` function. The `main()` function could check the error returned by the helper function and if it is non-nil, the `main()` function will call `os.Exit(1)`. 

Ex.

```
func lpMain() error {
    ...
}

func main() {
    if err := lpMain(); err != nil {
        os.Exit(1)
    }
}
```

kyriediculous: The current construction of the main function is something I wanted to discuss after researching good packet/code structure and will be part of my writeup which I'll hopefully get to this weekend. In short all startup of services should be wrapped in helper methods so the clutter within the main itself is rather small. 

I'm just not sure if it would be ideal to include such a refactor into this PR or if it warrants a seperate issue down the line (optimization). 

Perhaps we should compromise here for the time being and either go with `panic` (testable based on error code),  `return (status code 0)` (not testable on error code, but perhaps in a different way I'm not familiar with), or wrapping the entire current `main` body as you decribe.

kyriediculous: Just on a sidenote coming back to the original use of `glog.Fatal`,  we are also using `glog.Fatal()` for many other checks such as setting service URI, auth webhook URL, etc.

So I will be updating those as well to use `glog.Errorf + return`

yondonfu: Yeah I agree that we can keep the refactor of the `main()` function out of this PR.

I'm ok with using `panic` for now as wrapping the current body of `main()` will require a bunch of changes to return proper error values, but let's make sure to come back and address this particular issue in a separate PR.

kyriediculous: > Just on a sidenote coming back to the original use of glog.Fatal, we are also using glog.Fatal() for many other checks such as setting service URI, auth webhook URL, etc.

Coming back to my previous comment. I should update these to use panic as well then for proper execution of deferred functions.

server/webserver.go
```diff
@@ -132,10 +132,10 @@ func (s *LivepeerServer) cliWebServerHandlers(bindAddr string) *http.ServeMux {
 			pNames = append(pNames, p.Name)
 		}
 		config := struct {
-			BroadcastConfig    *BroadcastConfig
+			MaxPrice           *big.Rat
```
yondonfu, line 135: We'll need to update the CLI wizard as well because it currently tries to unmarshal the JSON response into a struct with a `big.Int` (https://github.com/livepeer/go-livepeer/blob/master/cmd/livepeer_cli/wizard_stats.go#L363) which would fail at the moment.

server/webserver.go
```diff
@@ -132,10 +132,10 @@ func (s *LivepeerServer) cliWebServerHandlers(bindAddr string) *http.ServeMux {
...
 			TranscodingOptions string
 		}{
-			BroadcastCfg,
+			BroadcastCfg.MaxPrice(),
```
yondonfu, line 138: This could just be `price`?

cmd/livepeer/livepeer.go
yondonfu: If we're planning on switching away from using `panic` in a separate PR, I think its ok to leave the existing usage of `glog.Fatal` for now just as long as we don't introduce any additional usage.

server/webserver.go
kyriediculous: Not the same scope. Could define price at the top of that function rather than in the mux block for `setBroadcastConfig `but seems less clean to do it that way imo.

yondonfu: Oh I see, my mistake - I was reading the collapsed diff on GH and didn't realize that this was in its own handler 😓

server/broadcast.go
kyriediculous: The zero value for `sync.RWMutex `is an unlocked mutex but the zero value for `*sync.RWMutex` is nil so we're getting a nil pointer panic when calling `BroadcastCfg.SetMaxBroadcastPrice()`

Am I correct in saying that? 
https://play.golang.org/p/b6CfEjnSuyw

Meaning we actually got to keep that line? See failing test in Circle as well.

yondonfu: Ah yeah should be fixed by using `sync.RWMutex` instead of `*sync.RWMutex` in `BroadcastConfig`

j0sh: The annoying thing would be having to dereference the mutex each time it's used. I've run up against this exact issue as well.

kyriediculous: Actually, after checking my notes again go seems to implicitly dereference this when using pointer receivers in the containing struct, if I'm correct? 

https://hackmd.io/A7ti7n0gQPC6p8OF_PMAzQ#syncMutex

yondonfu (approved):
LGTM after a rebase!

Issue #895, opened 2019-05-30
O processes payment from B and checks that minimum credit requirement is met


yondonfu:
**ProcessPayment(payment, transOpts, manifestID)**

Should check that a broadcaster meets the minimum credit requirement.

Note: In practice, `ProcessPayment` will include additional logic. That logic is ommitted in favor of focusing exclusively on payment accounting related logic.

```
currCredit = CurrCredit(payment.sender)
updatedCredit = currCredit + payment.value
UpdateCredit(payment.sender, updatedCredit)

if updatedCredit < MinCredit() {
    return ERROR_INSUFFICIENT_CREDIT
}

...

```

yondonfu:
We're currently planning on tracking credit balances locally on a per stream basis with relevant state flushed when the session with a stream times out (at the moment, approximately [1 minute](https://github.com/livepeer/go-livepeer/blob/master/core/orchestrator.go#L32) without receiving a segment). Orchestrators are already tracking session state using the [SegmentChans](https://github.com/livepeer/go-livepeer/blob/master/core/livepeernode.go#L61) map. We could make some modifications to this map and the [transcode segment loop](https://github.com/livepeer/go-livepeer/blob/master/core/orchestrator.go#L422) associated with each stream tracked by the map to support credit balance maintenance.

At the moment, `(*LivepeerNode).SegmentChans` maps the manifestID for a stream to a `SegmentChan` which is a type alias for a channel accepting [SegChanData](https://github.com/livepeer/go-livepeer/blob/master/core/orchestrator.go#L200) that holds input segment data as well as an output channel for sending transcoded results. Each `SegmentChan` can already be considered a "session" since an orchestrator maintains a single `SegmentChan` per manifestID (in fact, the code already checks the size of the `SegmentChans` map to [enforce an orchestrator session cap](https://github.com/livepeer/go-livepeer/blob/master/core/orchestrator.go#L80)).

In order to hold additional credit balance state per session, we can introduce a `SegSess` struct:

```
type SegSess struct {
    segChan SegmentChan
    credit *big.Rat
}
```

`SegSess` just wraps `SegmentChan` and also includes the `credit` field which is used to keep track of the broadcaster's current credit balance for the stream. We can maintain a `SegSess` per manifestID which still results in a `SegmentChan` per manifestID as well as a credit balance counter.

[...]

PR #937, opened 2019-06-25
Orchestrator acceptable PM ticket error handling

### LINKED ISSUE #634: pm: responses to acceptable and unacceptable invalid tickets and

yondonfu:

This PR adds logic for handling "acceptable" PM ticket errors. See [this comment](https://github.com/livepeer/go-livepeer/issues/634#issuecomment-503741952) for additional details on what constitutes an acceptable PM ticket error.

The approach we take in this PR is as follows:

- Count the number of acceptable PM ticket errors for each tracked sender
- Define a custom `pm.Error` interface that exposes an `Acceptable()` method. Callers that are interested in checking if an error returned from the `pm` package can use a type assertion and check the output of `Acceptable()`
- `pm.Recipient` will return `pm.Error` for "invalid ticket faceValue", "invalid ticket winProb" and "invalid already revealed recipientRand" errors (see the comment linked at the beginning of this post for the rationale behind focusing on these errors)
- `pm.SenderMonitor` will handle error count resets for max float changes and gas price changes

Note that `(*lphttp).ServeSegment` will always send back an updated OrchestratorInfo message when an error is returned by `(*orchestrator).ProcessPayment()`. The thought here is that when a #895 is implemented, the minimum credit requirement check will prevent an orchestrator from transcoding a segment if the tickets received were unacceptable such that the broadcaster's credit balance is insufficient - `(*orchestrator).ProcessPayment()` would only add to the credit amount if there was not an error in receiving the ticket or if the error receiving the ticket was acceptable.

- dee67e9: Count acceptable errors for senders and compare with a max error count
- 6c2e38b: Reset a sender's error count when their max float changes. Since the recipient will update the sender's max float when redeeming a winning ticket, resetting the error count for max float changes should eliminate the need to also reset the error count when redeeming a winning ticket.
- 0ae12d6: Listen for changes to the cached gas price in GasPriceMonitor
- cf35fbd: Fixed a few data races in the GasPriceMonitor tests
- 6d7a970: Reset the error count for all senders when the SenderMonitor receives a cached gas price change notification from the GasPriceMonitor
- abdeb9a d7dcbd3 and 040bd39: Isolate the acceptable error logic in the `core` and `pm` packages. `(*lphttp).ServeSegment()` will send back an updated OrchestratorInfo message whenever there is an error from `(*orchestrator).ProcessPayment()`

Wrote additional unit tests. Still need to do some manual testing.

Fixes #634 

**Checklist:**

### DIFF
--- a/eth/gaspricemonitor.go
+++ b/eth/gaspricemonitor.go
@@ -30,6 +30,10 @@ type GasPriceMonitor struct {
 	// gasPrice is the current gas price to be returned
 	// to users
 	gasPrice *big.Int
+
+	// update is a channel used to send notifications to a listener
+	// when the gas price is updated
+	update chan struct{}
 }
 
 // NewGasPriceMonitor returns a GasPriceMonitor
@@ -49,16 +53,19 @@ func (gpm *GasPriceMonitor) GasPrice() *big.Int {
 	return gpm.gasPrice
 }
 
-// Start starts polling for gas price updates
-func (gpm *GasPriceMonitor) Start(ctx context.Context) error {
+// Start starts polling for gas price updates and returns a channel to receive
+// notifications of gas price changes
+func (gpm *GasPriceMonitor) Start(ctx context.Context) (chan struct{}, error) {
 	if gpm.polling {
-		return errors.New("already polling")
+		return nil, errors.New("already polling")
 	}
 
 	// Initialize gasPrice before starting to poll
 	if err := gpm.fetchAndUpdateGasPrice(ctx); err != nil {
-		return err
+		return nil, err
 	}
+
+	gpm.update = make(chan struct{})
 
 	cctx, cancel := context.WithCancel(ctx)
 	gpm.cancel = cancel
@@ -73,7 +80,6 @@ func (gpm *GasPriceMonitor) Start(ctx context.Context) error {
 					glog.Errorf("error getting gas price: %v", err)
 				}
 			case <-ctx.Done():
-				gpm.cancel = nil
 				gpm.polling = false
 				return
 			}
@@ -82,7 +88,7 @@ func (gpm *GasPriceMonitor) Start(ctx context.Context) error {
 
 	gpm.polling = true
 
-	return nil
+	return gpm.update, nil
 }
 
 // Stop stops polling for gas price updates
@@ -104,16 +110,27 @@ func (gpm *GasPriceMonitor) fetchAndUpdateGasPrice(ctx context.Context) error {
 		return err
 	}
 
-	gpm.updateGasPrice(gasPrice)
+	gpm.updateGasPrice(ctx, gasPrice)
 
 	glog.V(common.DEBUG).Infof("Cached gas price: %v", gasPrice)
 
 	return nil
 }
 
-func (gpm *GasPriceMonitor) updateGasPrice(gasPrice *big.Int) {
+func (gpm *GasPriceMonitor) updateGasPrice(ctx context.Context, gasPrice *big.Int) {
 	gpm.mu.Lock()
 	defer gpm.mu.Unlock()
+
+	if gasPrice.Cmp(gpm.gasPrice) != 0 && gpm.update != nil {
+		// Notify listener that the new gas price
+		// is different from the current cached gas price
+		// Note: If there is no listener, this call will block
+		select {
+		case gpm.update <- struct{}{}:
+		case <-ctx.Done():
+			return
+		}
+	}
 
 	gpm.gasPrice = gasPrice
 }
--- a/eth/gaspricemonitor.go
+++ b/eth/gaspricemonitor.go
@@ -20,15 +20,17 @@ type GasPriceOracle interface {
 // GasPriceMonitor polls for gas price updates and updates its
 // own view of the current gas price that can be used by others
 type GasPriceMonitor struct {
-	gpo             GasPriceOracle
+	gpo GasPriceOracle
+	// The following fields should be protected by `pollingMu`
 	polling         bool
 	pollingInterval time.Duration
 	cancel          context.CancelFunc
+	// pollingMu protects access to polling related fields
+	pollingMu sync.Mutex
 
-	// mu is a mutex that protects access to gasPrice
-	mu sync.Mutex
-	// gasPrice is the current gas price to be returned
-	// to users
+	// gasPriceMu protects access to gasPrice
+	gasPriceMu sync.RWMutex
+	// gasPrice is the current gas price to be returned to users
 	gasPrice *big.Int
 
 	// update is a channel used to send notifications to a listener
@@ -47,8 +49,8 @@ func NewGasPriceMonitor(gpo GasPriceOracle, pollingInterval time.Duration) *GasP
 
 // GasPrice returns the current gas price
 func (gpm *GasPriceMonitor) GasPrice() *big.Int {
-	gpm.mu.Lock()
-	defer gpm.mu.Unlock()
+	gpm.gasPriceMu.RLock()
+	defer gpm.gasPriceMu.RUnlock()
 
 	return gpm.gasPrice
 }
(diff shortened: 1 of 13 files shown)


eth/gaspricemonitor_test.go
```diff
@@ -59,23 +59,26 @@ func TestStart(t *testing.T) {
...
 	gpo.SetErr(expErr)
-	err := gpm.Start(context.Background())
+	update, err := gpm.Start(context.Background())
+	assert.Nil(update)
 	assert.EqualError(err, expErr.Error())
 
 	// Switch back to no errors for SuggestGasPrice
 	gpo.SetErr(nil)
 
 	// Test success
```
kyriediculous, line 77: blank line can be removed I guess?

eth/gaspricemonitor_test.go
```diff
@@ -59,23 +59,26 @@ func TestStart(t *testing.T) {
...
 
-	err = gpm.Start(context.Background())
+	update, err = gpm.Start(context.Background())
+	assert.NotNil(update)
 	assert.Nil(err)
 	defer gpm.Stop()
 
 	assert.Equal(gasPrice, gpm.GasPrice())
 
 	// Test error when already polling
```
kyriediculous, line 86: also blank line here that can be removed I think.

eth/gaspricemonitor.go
```diff
@@ -24,9 +24,12 @@ type GasPriceMonitor struct {
 	polling         bool
 	pollingInterval time.Duration
 	cancel          context.CancelFunc
+	// pollingMu is a mutex that protects access to polling related
+	// fields
```
kyriediculous: seems sort of trivial to have one word on a seperate line 😅

eth/gaspricemonitor.go
```diff
@@ -24,9 +24,12 @@ type GasPriceMonitor struct {
...
+	pollingMu sync.Mutex
 
-	// mu is a mutex that protects access to gasPrice
-	mu sync.Mutex
+	// gasPriceMu is a mutex that protects access to gasPrice
+	gasPriceMu sync.RWMutex
 	// gasPrice is the current gas price to be returned
```
kyriediculous: Same here, but maybe my screen is just too wide.

pm/sendermonitor.go
```diff
@@ -82,11 +82,15 @@ type senderMonitor struct {
 	// each of currently active remote senders
 	redeemable chan *SignedTicket
 
+	// gasPriceUpdate is a channel used to receive notifications
+	// of gas price changes
```
kyriediculous: Could be one line comment as well but I'm really just nitpicking when it comes to that

pm/recipient.go
```diff
@@ -274,10 +274,10 @@ func (r *recipient) acceptTicket(ticket *Ticket, sig []byte, recipientRand *big.
 		// When a winning ticket is redeemed, the ticket's recipientRand is invalidated
 		// and the sender must send tickets with a new seed, but there could be a delay
 		// before the sender is notified of the new seed.
-		//
-		// TODO(yondonfu): Track the number of these types of errors to determine
-		// whether we should accept more of these types of errors
-		return errors.Errorf("invalid already revealed recipientRand %v", recipientRand)
+		return newReceiveError(
+			errors.Errorf("invalid already revealed recipientRand %v", recipientRand),
```
kyriediculous, line 278: I think there's a word missing here ? 
eg. 
`"Invalid seed, already revealed reciepientRand %v"`
or
`"Invalid ticket, already revealed reciepientRand %v"`

eth/gaspricemonitor.go
j0sh: Comment golf is fun: how concise can we make the comment while preserving clarity of the meaning? In this case, we can probably omit the first five words and say "Protects access to polling related fields"

Although the convention I like is to place this note preceding the fields that need to be protected; eg "The following fields need to be protected by the mutex `pollingMu`" and so forth; see an example here

https://github.com/livepeer/go-livepeer/blob/17701ca9a7bfb8b84de08da9fae1223ae2053b29/server/mediaserver.go#L90-L95

yondonfu: Fixed in 2cec5ef

eth/gaspricemonitor.go
yondonfu: Fixed in 2cec5ef

pm/recipient.go
yondonfu: I feel like this error message does accurately describe what went wrong while keeping it concise, but happy to change it if you feel strongly about it

eth/gaspricemonitor_test.go
yondonfu: There is a lot of variability when it comes to spacing of comments in the codebase - I believe some people prefer having the extra space for readability while some might prefer omitting the extra space to minimize LOC. Personally I don't really mind either style and don't think it has any negative impact as long as the spacing isn't extraneous (i.e. if nothing else changed besides the addition of a new line then I would suggest removing it in most cases) so I would suggest excluding those style nits during review if the extra space is already there (i.e. not added as a part of the PR in an extraneous way) - what do you think?

pm/sendermonitor.go
yondonfu: Fixed in 0130b20

pm/recipient.go
kyriediculous: No not at all. I just reads a bit weird that's all, perhaps a colon or comma might help but up to you.

eth/gaspricemonitor_test.go
kyriediculous: Sounds good !

kyriediculous (approved):
LGTM but perhaps you want to check whether @j0sh approves if he has the bandwith for it.

j0sh (approved):
🚢

PR #939, opened 2019-06-26
common, discovery: add price info to DB cache

### LINKED ISSUE #934: Update DB cache to store Pricing information
#906 allows broadcasters to optionally set a `MaxBroadcastPrice` to filter the orchestrators in the cached orchestratorpool based on an O's pricing info. 

This results`getOrchestratorInfo` call for every orchestrator in the cached `orchestratorPool` , which might result in a lot of requests being sent when there hundreds of orchestrators on the network. 

Rather than querying price for all orchestrators in the cached pool, we should do a quick first pass on orchestrators we know passed the pricing requirements for B before. 

Pricing information for all orchestrators will still be updated when the cache is refreshed 

0. Refresh DB cache, insert orchestrator pricing info. 
    * update `cacheRegisteredTranscoders` in `db_discover.go` to include a flow/logic similar to initiating  `newOnchainOrchestratorPool` and calling `orchestratorPool.getOrchestrators` with the addition of including the ethereum address.
    * **CAVEAT**: `NewOnchainOrchestratorPool`/ `orchestratorPool.getOrchestrators` does not return the ethereum address for the orchestrator. So it seems it might be better to largely duplicate the logic from these two methods into `cacheRegisteredTranscoders` rather than changing `NewOrchestratorPool` to take in a struct of `{ServiceURI, EthAddress}` which implies reworking many tests as well. 

2. Stream comes in 

3. Filter eligible orchestrators from DB cache
    * Update` db.SelectOrchs` to take in a filter struct as argument if `filter == nil , ignore filter`
    * Update `selectOrchs query` or create a new SQL query in `common/db` to filter based on filter struct

5. Send `OrchestratorInfo` request to eligible orchestrators

6. Filter out eligble orchestrators from response
    * See #906

8. Use top-N eligible responses for orchestrator selection

A slightly different flow, that only stores eligble O's in DBcache to begin with. Omitting the need to alter `db.SelectOrchs` which would need access to `server.BroadcastPrice` resulting in a cyclical import. 

0. Refresh DB cache
  0.1 Fetch orchestrator info/pricing info for O's 
  0.2 Insert only eligible O's into DBcache 
2. Get O's from DBcache 
3. Send Orchestrator info request to O's in cache
4. Filter out eligble O's from response
5 Use top-N eligble responses for O selection

**Additional context**
For more context, please see the discussion and comment from Josh in #906 .

kyriediculous:
Adds pricing info for orchestrators to the DBcache after fetching them on-chain. This cache refreshes just like before every hour currently. 

Broadcasters can now use the pricing info stored in the cache to pre-filter orchestrators that don't match their pricing preferences (see #906 ), rather than getting all orchestrators from the cache and sending `OrchestratorInfo` messages to all of them. 

- Added a column to the orchestrator table containing priceInfo as a fixed point
- Added an SQL query `db.FilterOrchs(maxPrice *big.Rat)` to filter stored price info based on the `maxPrice` passed to this function, which is fetched from `BroadcastCfg.MaxPrice()`
- Added a helper function in `util`to convert `big.Rat` to fixed point (`int64`)
- Changed `cacheDBOrchs`in `db_discovery.go` to concurrently fetch `OrchestratorInfo` messages for all orchestrators found on-chain. We are now discarding orchestrators with bad URLs as well (I'm not sure if this is behaviour that we want; but we can't query those URL's anyway)

Added unit tests in `db_test.go` , `util_test.go` and slightly altered unit tests in `db_discovery.go`
tests in the latter aren't really pretty and `serverGetOrchInfo` is defined repeadetly

Fixes #934 

**Checklist:**

### DIFF
--- a/common/db.go
+++ b/common/db.go
@@ -20,6 +20,7 @@ type DB struct {
 
 	// prepared statements
 	selectOrchs                *sql.Stmt
+	filterOrchs                *sql.Stmt
 	updateOrch                 *sql.Stmt
 	updateKV                   *sql.Stmt
 	insertUnbondingLock        *sql.Stmt
@@ -30,8 +31,9 @@ type DB struct {
 }
 
 type DBOrch struct {
-	ServiceURI   string
-	EthereumAddr string
+	ServiceURI    string
+	EthereumAddr  string
+	PricePerPixel int64
 }
 
 type DBUnbondingLock struct {
@@ -39,6 +41,10 @@ type DBUnbondingLock struct {
 	Delegator     ethcommon.Address
 	Amount        *big.Int
 	WithdrawRound int64
+}
+
+type DBOrchFilter struct {
+	MaxPrice *big.Rat
 }
 
 var LivepeerDBVersion = 1
@@ -58,7 +64,8 @@ var schema = `
 		ethereumAddr STRING PRIMARY KEY,
 		createdAt STRING DEFAULT CURRENT_TIMESTAMP NOT NULL,
 		updatedAt STRING DEFAULT CURRENT_TIMESTAMP NOT NULL,
-		serviceURI STRING
+		serviceURI STRING,
+		pricePerPixel int64
 	);
 
 	CREATE TABLE IF NOT EXISTS unbondingLocks (
@@ -137,17 +144,8 @@ func InitDB(dbPath string) (*DB, error) {
 		// all good; nothing to do
 	}
 
-	// select all orchestrators updated in the last 24 hours
-	stmt, err := db.Prepare("SELECT serviceURI, ethereumAddr FROM orchestrators WHERE updatedAt >= datetime('now','-1 day')")
-	if err != nil {
-		glog.Error("Unable to select orchestrators updated in the past 24 hours", err)
-		d.Close()
-		return nil, err
-	}
-	d.selectOrchs = stmt
-
 	// updateOrchestrators statement
-	stmt, err = db.Prepare("INSERT OR REPLACE INTO orchestrators(updatedAt, serviceURI, ethereumAddr, createdAt) VALUES(datetime(), ?1, ?2, (SELECT createdAt FROM orchestrators WHERE ethereumAddr = ?2))")
+	stmt, err := db.Prepare("INSERT OR REPLACE INTO orchestrators(updatedAt, serviceURI, ethereumAddr, pricePerPixel, createdAt) VALUES(datetime(), ?1, ?2, ?3, (SELECT createdAt FROM orchestrators WHERE ethereumAddr = ?2))")
 	if err != nil {
 		glog.Error("Unable to prepare updateOrchestrators stmt ", err)
 		d.Close()
@@ -215,6 +213,9 @@ func (db *DB) Close() {
 	if db.selectOrchs != nil {
 		db.selectOrchs.Close()
 	}
+	if db.filterOrchs != nil {
+		db.filterOrchs.Close()
+	}
 	if db.updateKV != nil {
 		db.updateKV.Close()
 	}
@@ -272,7 +273,7 @@ func (db *DB) UpdateOrch(orch *DBOrch) error {
 		return nil
 	}
 
-	_, err := db.updateOrch.Exec(orch.ServiceURI, orch.EthereumAddr)
+	_, err := db.updateOrch.Exec(orch.ServiceURI, orch.EthereumAddr, orch.PricePerPixel)
 	if err != nil {
 		glog.Error("db: Unable to update orchestrator ", err)
 	}
@@ -280,15 +281,15 @@ func (db *DB) UpdateOrch(orch *DBOrch) error {
 	return err
 }
 
-func (db *DB) SelectOrchs() ([]*DBOrch, error) {
+func (db *DB) SelectOrchs(filter *DBOrchFilter) ([]*DBOrch, error) {
 	if db == nil {
 		return nil, nil
 	}
 
-	rows, err := db.selectOrchs.Query()
+	rows, err := db.dbh.Query(buildSelectOrchsQuery(filter))
 	defer rows.Close()
 	if err != nil {
-		glog.Error("db: Unable to get orchestrators updated in the last 24 hours", err)
+		glog.Error("db: Unable to get orchestrators updated in the last 24 hours: ", err)
 		return nil, err
 	}
 	orchs := []*DBOrch{}
@@ -461,4 +462,16 @@ func buildWinningTicketsQuery(sessionIDs []string) string {
 		sessionIDs[i] = strconv.Quote(sessionIDs[i])
 	}
 	return "SELECT sender, recipient, faceValue, winProb, senderNonce, recipientRand, recipientRandHash, sig, sessionID FROM winningTickets WHERE sessionID IN (" + strings.Join(sessionIDs, ", ") + ")"
+}
+
+func buildSelectOrchsQuery(filter *DBOrchFilter) (string, error) {
+	query := "SELECT serviceURI, ethereumAddr FROM orchestrators WHERE updatedAt >= datetime('now','-1 day')"
+	if filter != nil && filter.MaxPrice != nil {
+		fixedPrice, err := PriceToFixed(filter.MaxPrice)
+		if err != nil {
+			return "", err
+		}
+		query = query + " AND pricePerPixel <= " + strconv.FormatInt(fixedPrice, 10)
+	}
+	return query, nil
 }
--- a/common/db_test.go
+++ b/common/db_test.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"math"
 	"math/big"
+	"strconv"
 	"testing"
 
 	ethcommon "github.com/ethereum/go-ethereum/common"
@@ -127,7 +128,7 @@ func TestSelectUpdateOrchs_EmptyOrNilInputs_NoError(t *testing.T) {
 	require.Nil(err)
 
 	// selecting empty set of orchs
-	orchs, err := dbh.SelectOrchs()
+	orchs, err := dbh.SelectOrchs(nil)
 	require.Nil(err)
 	assert.Empty(orchs)
 
@@ -151,7 +152,7 @@ func TestSelectUpdateOrchs_AddingUpdatingRow_NoError(t *testing.T) {
 	err = dbh.UpdateOrch(orch)
 	require.Nil(err)
 
-	orchs, err := dbh.SelectOrchs()
+	orchs, err := dbh.SelectOrchs(nil)
 	require.Nil(err)
 	assert.Len(orchs, 1)
 	assert.Equal(orchs[0].ServiceURI, orch.ServiceURI)
@@ -161,7 +162,7 @@ func TestSelectUpdateOrchs_AddingUpdatingRow_NoError(t *testing.T) {
 	err = dbh.UpdateOrch(orchUpdate)
 	require.Nil(err)
 
-	updatedOrch, err := dbh.SelectOrchs()
+	updatedOrch, err := dbh.SelectOrchs(nil)
 	assert.Len(updatedOrch, 1)
 	assert.Equal(updatedOrch[0].ServiceURI, orchUpdate.ServiceURI)
 }
@@ -180,7 +181,7 @@ func TestSelectUpdateOrchs_AddingMultipleRows_NoError(t *testing.T) {
 	err = dbh.UpdateOrch(orch)
 	require.Nil(err)
 
-	orchs, err := dbh.SelectOrchs()
+	orchs, err := dbh.SelectOrchs(nil)
 	require.Nil(err)
 	assert.Len(orchs, 1)
 	assert.Equal(orchs[0].ServiceURI, orch.ServiceURI)
@@ -191,10 +192,56 @@ func TestSelectUpdateOrchs_AddingMultipleRows_NoError(t *testing.T) {
 	err = dbh.UpdateOrch(orchAdd)
 	require.Nil(err)
 
-	orchsUpdated, err := dbh.SelectOrchs()
+	orchsUpdated, err := dbh.SelectOrchs(nil)
 	require.Nil(err)
 	assert.Len(orchsUpdated, 2)
 	assert.Equal(orchsUpdated[1].ServiceURI, orchAdd.ServiceURI)
+}
+
+func TestDBFilterOrchs(t *testing.T) {
+	var nilDb *DB
+	nilOrchs, nilErr := nilDb.SelectOrchs(&DBOrchFilter{big.NewRat(1, 1)})
+	assert.Nil(t, nilOrchs)
+	assert.Nil(t, nilErr)
+
+	dbh, dbraw, err := TempDB(t)
+	defer dbh.Close()
+	defer dbraw.Close()
+	require := require.New(t)
+	assert := assert.New(t)
+	require.Nil(err)
+
+	for i := 0; i < 10; i++ {
+		orch := NewDBOrch("https://127.0.0.1:"+strconv.Itoa(8936+i), string(pm.RandBytes(32)))
+		orch.PricePerPixel, err = PriceToFixed(big.NewRat(1, int64(5+i)))
+		require.Nil(err)
+		err = dbh.UpdateOrch(orch)
+		require.Nil(err)
+	}
+
+	orchsUpdated, err := dbh.SelectOrchs(nil)
+	require.Nil(err)
+	assert.Len(orchsUpdated, 10)
+
+	// Passing in nil max price to filterOrchs returns a query for selectOrchs
+	orchsFiltered, err := dbh.SelectOrchs(nil)
+	require.Nil(err)
+	assert.Len(orchsFiltered, 10)
+
+	// Passing in a higher maxPrice than all orchs to filterOrchs returns all orchs
+	orchsFiltered, err = dbh.SelectOrchs(&DBOrchFilter{big.NewRat(10, 1)})
+	require.Nil(err)
+	assert.Len(orchsFiltered, 10)
+
+	// Passing in a lower price than all orchs returns no orchs
+	orchsFiltered, err = dbh.SelectOrchs(&DBOrchFilter{big.NewRat(1, 15)})
+	require.Nil(err)
+	assert.Len(orchsFiltered, 0)
+
+	// Passing in 1/10 returns 5 orchs
+	orchsFiltered, err = dbh.SelectOrchs(&DBOrchFilter{big.NewRat(1, 10)})
+	require.Nil(err)
+	assert.Len(orchsFiltered, 5)
 }
 
 func TestDBUnbondingLocks(t *testing.T) {
--- a/common/util.go
+++ b/common/util.go
@@ -160,3 +160,15 @@ func GenErrRegex(errStrings []string) *regexp.Regexp {
 	}
 	return regexp.MustCompile(strings.Join(groups, "|"))
 }
+
+// PriceToFixed converts a big.Rat into a fixed point number represented as int64
+// using a scaleFactor of 1000 resulting in max decimal places of 3
+func PriceToFixed(price *big.Rat) (int64, error) {
+	scalingFactor := int64(1000)
+	if price == nil {
+		return 0, fmt.Errorf("reference to rat is nil")
+	}
+	scaled := new(big.Rat).Mul(price, big.NewRat(scalingFactor, 1))
+	fp, _ := new(big.Float).SetRat(scaled).Int64()
+	return fp, nil
+}
--- a/common/util_test.go
+++ b/common/util_test.go
@@ -2,10 +2,12 @@ package common
 
 import (
 	"encoding/hex"
+	"math/big"
 	"testing"
 
 	ethcommon "github.com/ethereum/go-ethereum/common"
 	"github.com/livepeer/lpms/ffmpeg"
+	"github.com/stretchr/testify/assert"
 )
 
 func TestTxDataToVideoProfile(t *testing.T) {
@@ -61,4 +63,33 @@ func TestProfilesToHex(t *testing.T) {
 	// compare([]ffmpeg.VideoProfile{})
 	compare([]ffmpeg.VideoProfile{ffmpeg.P240p30fps16x9})
 	compare([]ffmpeg.VideoProfile{ffmpeg.P240p30fps16x9, ffmpeg.P360p30fps16x9})
+}
+
+func TestPriceToFixed(t *testing.T) {
+	assert := assert.New(t)
+
+	// if rat is nil returns 0 and error
+	fp, err := PriceToFixed(nil)
+	assert.Zero(fp)
+	assert.Error(err)
+
+	// 1/10 rat returns 100
+	fp, err = PriceToFixed(big.NewRat(1, 10))
+	assert.Nil(err)
+	assert.Equal(fp, int64(100))
+
+	// 500/1 returns 500000 with
+	fp, err = PriceToFixed(big.NewRat(500, 1))
+	assert.Nil(err)
+	assert.Equal(fp, int64(500000))
+
+	// 125/100 returns 1250
+	fp, err = PriceToFixed(big.NewRat(125, 100))
+	assert.Nil(err)
+	assert.Equal(fp, int64(1250))
+
+	// rat smaller than 1/1000 returns 0
+	fp, err = PriceToFixed(big.NewRat(1, 5000))
+	assert.Nil(err)
+	assert.Zero(fp)
 }
--- a/discovery/db_discovery.go
+++ b/discovery/db_discovery.go
@@ -1,8 +1,11 @@
 package discovery
 
 import (
+	"context"
+	"fmt"
 	"math/big"
 	"net/url"
+	"strings"
 	"time"
 
 	"github.com/livepeer/go-livepeer/common"
@@ -45,7 +48,7 @@ func NewDBOrchestratorPoolCache(node *core.LivepeerNode) *DBOrchestratorPoolCach
 }
 
 func (dbo *DBOrchestratorPoolCache) GetURLs() []*url.URL {
-	orchs, err := dbo.node.Database.SelectOrchs()
+	orchs, err := dbo.node.Database.SelectOrchs(&common.DBOrchFilter{MaxPrice: server.BroadcastCfg.MaxPrice()})
 	if err != nil || len(orchs) <= 0 {
 		return nil
 	}
@@ -60,7 +63,7 @@ func (dbo *DBOrchestratorPoolCache) GetURLs() []*url.URL {
 }
 
 func (dbo *DBOrchestratorPoolCache) GetOrchestrators(numOrchestrators int) ([]*net.OrchestratorInfo, error) {
-	orchs, err := dbo.node.Database.SelectOrchs()
+	orchs, err := dbo.node.Database.SelectOrchs(&common.DBOrchFilter{MaxPrice: server.BroadcastCfg.MaxPrice()})
 	if err != nil || len(orchs) <= 0 {
 		return nil, err
 	}
@@ -90,7 +93,7 @@ func (dbo *DBOrchestratorPoolCache) GetOrchestrators(numOrchestrators int) ([]*n
 }
 
 func (dbo *DBOrchestratorPoolCache) Size() int {
-	orchs, err := dbo.node.Database.SelectOrchs()
+	orchs, err := dbo.node.Database.SelectOrchs(&common.DBOrchFilter{MaxPrice: server.BroadcastCfg.MaxPrice()})
 	if err != nil {
 		return 0
 	}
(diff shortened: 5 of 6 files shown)


discovery/db_discovery.go
```diff
@@ -119,19 +123,84 @@ func cacheDBOrchs(node *core.LivepeerNode, orchs []*lpTypes.Transcoder) ([]*comm
...
+	ctx, cancel := context.WithTimeout(context.Background(), getOrchestratorsTimeoutLoop)
+	orchChan := make(chan struct{}, numOrchestrators)
+	numResp := 0
+	numSuccessResp := 0
+	respLock := sync.Mutex{}
+
+	getOrchInfo := func(dbOrch *common.DBOrch) {
+		uri := parseURI(dbOrch.ServiceURI)
+		respLock.Lock()
+		defer respLock.Unlock()
```
j0sh: This holds the lock for the duration of the `serverGetOrchInfo` request which basically serializes all the requests and would make it very slow. I think you want to:

1. Parse the URI
2. Invoke serverGetOrchInfo
3. Acquire the lock (and defer unlock)
4. Do the rest

This would also maintain the serialization of DB writes which is probably a good thing given sqlite is single-writer.

discovery/db_discovery.go
```diff
@@ -119,19 +123,84 @@ func cacheDBOrchs(node *core.LivepeerNode, orchs []*lpTypes.Transcoder) ([]*comm
...
-		if err := node.Database.UpdateOrch(dbOrch); err != nil {
-			glog.Error("Error updating Orchestrator in DB: ", err)
-			continue
+		go getOrchInfo(dbOrch)
+
+	}
+
+	select {
+	case <-ctx.Done():
+		respLock.Lock()
```
j0sh: I'd be OK wrapping this up into a local function to minimize code duplication across the two cases (which is also an issue in the other file)

discovery/db_discovery.go
```diff
@@ -119,19 +123,84 @@ func cacheDBOrchs(node *core.LivepeerNode, orchs []*lpTypes.Transcoder) ([]*comm
...
+		if len(dbOrchs) < numOrchestrators {
+			numOrchestrators = len(dbOrchs)
```
j0sh: I'd probably rather unconditionally set a new variable here (numOrchestrators is tracking a different thing at this point, which is potentially confusing), or use `dbOrchs[:len(dbOrchs)]` which seems clear enough on its own.

discovery/db_discovery.go
```diff
@@ -119,19 +123,84 @@ func cacheDBOrchs(node *core.LivepeerNode, orchs []*lpTypes.Transcoder) ([]*comm
...
+		glog.Info("Done fetching orch info for orchestrators, numResponses fetched: ", returnDbOrchs)
+		cancel()
+		return returnDbOrchs, nil
 	}
-	return dbOrchs, nil
+}
 
+func parseURI(addr string) *url.URL {
+	if !strings.HasPrefix(addr, "http") {
+		addr = "https://" + addr
```
j0sh, line 180: I'm not sure if we should be forcing the prefix at this point; seems like we should just use the string we've received.

common/db.go
```diff
@@ -288,7 +303,47 @@ func (db *DB) SelectOrchs() ([]*DBOrch, error) {
...
+	)
+
+	if maxPrice == nil {
+		rows, err = db.selectOrchs.Query()
+	} else {
+		fpMaxPrice, err := RatToFixed(maxPrice, 1000, 3)
+		if err != nil {
+			return nil, err
+		}
+		rows, err = db.filterOrchs.Query(fpMaxPrice)
```
j0sh: This `err` wouldn't be actually be checked past the `else` clause, because it's re-declared with `fpMaxPrice`. See https://play.golang.org/p/reE2q-dZI_9

(i really dislike golang's style of variable initialization because of easy-to-make mistakes with shadowing like this)

discovery/db_discovery.go
```diff
@@ -60,7 +64,7 @@ func (dbo *DBOrchestratorPoolCache) GetURLs() []*url.URL {
 }
 
 func (dbo *DBOrchestratorPoolCache) GetOrchestrators(numOrchestrators int) ([]*net.OrchestratorInfo, error) {
-	orchs, err := dbo.node.Database.SelectOrchs()
+	orchs, err := dbo.node.Database.FilterOrchs(server.BroadcastCfg.MaxPrice())
```
j0sh: We probably should update the DB call in the `Size` function as well - given that it already filters on date, my interpretation of `Size` is "size of the orchestrator set we can use given our current criteria" and this should include pricing

discovery/db_discovery.go
```diff
@@ -119,19 +123,84 @@ func cacheDBOrchs(node *core.LivepeerNode, orchs []*lpTypes.Transcoder) ([]*comm
...
+			info, err := serverGetOrchInfo(ctx, core.NewBroadcaster(node), uri)
+			if err == nil {
+				dbOrch.PricePerPixel, err = common.RatToFixed(big.NewRat(info.PriceInfo.GetPixelsPerUnit(), info.PriceInfo.GetPricePerUnit()), 1000, 3)
+				if err != nil {
+					glog.Error("Error converting orchestrator price per pixel to a fixed point number")
+				} else if err := node.Database.UpdateOrch(dbOrch); err != nil {
+					glog.Error("Error updating Orchestrator in DB: ", err)
+				} else {
+					dbOrchs = append(dbOrchs, dbOrch)
+					numSuccessResp++
```
j0sh: The numSuccessResp counter looks unnecessary - it serves a different purpose in the `discovery.go` code where `numOrchestrators < len(orchs)`; here they are the same.

common/db.go
```diff
@@ -288,7 +303,47 @@ func (db *DB) SelectOrchs() ([]*DBOrch, error) {
...
+			continue
+		}
+		orch.ServiceURI = serviceURI
+		orch.EthereumAddr = ethereumAddr
+		orchs = append(orchs, &orch)
+	}
+	return orchs, nil
+}
+
+func (db *DB) FilterOrchs(maxPrice *big.Rat) ([]*DBOrch, error) {
```
j0sh: Curious why we don't modify SelectOrchs to take a `maxPrice` since passing in `nil` here will achieve the same effect (keeping two separate SQL queries is fine). That would minimize the code duplication and this file is already boilerplate-heavy.

common/util.go
```diff
@@ -149,3 +150,18 @@ func GetConnectionAddr(ctx context.Context) string {
...
+
+// RatToFixed converts a big.Rat into a fixed point number represented as int64
+// according to supplied scale factor and maximum decimalplaces
+func RatToFixed(rat *big.Rat, scalingFactor int64, maxDecimalPlaces int) (int64, error) {
+	if rat == nil {
+		return 0, fmt.Errorf("reference to rat is nil")
+	}
+	scaled := new(big.Rat).Mul(rat, big.NewRat(scalingFactor, 1))
+	s := scaled.FloatString(maxDecimalPlaces)
+	intPart, err := strconv.ParseInt(strings.Split(s, ".")[0], 10, 64)
```
j0sh: Wow, that is painful, but I see why it's done this way.

common/util.go
```diff
@@ -149,3 +150,18 @@ func GetConnectionAddr(ctx context.Context) string {
 	}
 	return from
 }
+
+// RatToFixed converts a big.Rat into a fixed point number represented as int64
+// according to supplied scale factor and maximum decimalplaces
+func RatToFixed(rat *big.Rat, scalingFactor int64, maxDecimalPlaces int) (int64, error) {
```
j0sh: Do we anticipate variation in how this is used, or could we hard-code the last two params? The advantage I see is there's less room for error when converting pricing info (eg, someone down the road uses a scaling factor of 10000 rather than 1000).

j0sh (changes requested):
I realize most of the code was copied over from `discovery.go` but that in itself could use some TLC, and not everything that's in there applies here.

Here's how I would do it today:

```golang
	numOrchs := 0
	resc, errc := make(chan *int), make(chan error)
	ctx, cancel := context.WithTimeout(context.Background(), getOrchestratorsTimeoutLoop)
	defer cancel()
	
	// no need for the numResp / orchChan / respLock

	getOrchInfo := func(dbOrch *common.DBOrch) {
		uri, err := url.ParseRequestURI(dbOrch.ServiceURI)
		if err != nil {
                         // we can wrap the following into a defer if we don't shadow `err` with each declaration
                         errc <- err
			return
		}
		info, err := serverGetOrchInfo(ctx, core.NewBroadcaster(node), uri)
		if err != nil {
                         errc <- err
			return
		}
		// ... do the rest of the important stuff ...
		resc <- dbOrch // return single DB info here
	}

	for _, dbOrch := range orchs {
		if v == nil {
			continue
		}
		numOrchs++
		go getOrchInfo(dbOrch)
	}

	// collect as many responses as we can and append to slice here 
	dbOrchs := []*common.DBOrch{}
	for j := 0; j < numOrchs; j++ {
		select {
		case res := <-resc:
			dbOrchs = append(dbOrchs, res)
		case <-errc:
			fmt.Println("Error getting O response")
		case <-ctx.Done():
			fmt.Println("Timed out")
			break
		}
	}
	return dbOrchs, nil
```

(The pattern for waiting multiple goroutines as shown [here](https://nathanleclaire.com/blog/2014/02/21/how-to-wait-for-all-goroutines-to-finish-executing-before-continuing-part-two-fixing-my-ooops/) seems to fit much better)

common/util.go
kyriediculous: Could you elaborate on what exactly is painful 😆 .
The scaleFactor and decimals are arbitrary and we can still play with this but the workflow is what it is. sqlite-go doesn't have DECIMALS so we have to use int64.

discovery/db_discovery.go
kyriediculous: I think so, because we're fetching these from on-chain and bad URL's can thus happen.

discovery/db_discovery.go
kyriediculous: My interpretation was that it's used to return the number of O's in the dbCache regardelss of filters, that's how it's used in tests at least.

common/db.go
kyriediculous: We could have. I think semantically FilterOrchs represents better what it does, since it now takes in a Filter struct rather than just a price (for possible flexibility towards the future). I thought this would have been cleaner / clearer than changing every selectOrchs statement to take in nil/empty filter struct, I haven't figured out yet if that's me being lazy or not 😅 .

common/util.go
kyriediculous: I don't know going forward, but I constructed it this way out of possible flexibility in the future. 

I did some playing with the scaling factor (I was a proponent of using fixed point 52.12) and with the current setup we run into "collisions" starting at a 1/36 big.Rat. Whether that's sufficient will be dependant on real market prices set by users.  

Just for reference, here's an excerpt of the discussion I had with Yondon: 

**N to Y:**
> We can use a higher scaling factor which would allow for more decimal places, but we'd have to decide how much precision is enough precision which feels like an arbitrary decision until people actually start setting their own prices in an open marketplace.

I agree that this is arbitrary. I just noticed that accuracy is rather bad using 1000 & 3 since the ultimate goal for this is to compare prices.  We are running into "collisions" starting at 1/36. But then again that accuracy might be sufficient , I guess we don't know and I guess that will also depend on the future dollar value of a wei.

**Y to N:**
>  I just noticed that accuracy is rather bad using 1000 & 3 since the ultimate goal for this is to compare prices

Sure, but my point is that we should consider the setting we're operating in and that we'll always be limited in precision whether the scaling factor is 1000 or 1000000. If ETH/USD skyrockets, then perhaps there could become a greater need to price more granularly. There's also a possibility that we switch to a more stable underlying currency for payment in the future - in that scenario, the exchange rate would have less of an impact on the need to price more granularly. 

If not 1000, we could bump it to something like 100000 which provides 5 decimal places of precision, but I wouldn't spend too much time optimizing at this moment past that

common/db.go
kyriediculous: Yep, totally right. Wasn't paying attention. 
Would you be ok with declaring fpMaxPrice in the `var` block above, despite it not having to be in the outer scope necessarily? I mean we could also just initiate it in the else statement itself. Not sure what's better.

discovery/db_discovery.go
kyriediculous: I adopted the design using channels you proposed. I think it is way better, I just didn't think of better ways when writing up the code during an all-nighter and I still need some practice with concurrency patterns in go.

common/util.go
yondonfu: I think declaring `scalingFactor` and `maxDecimalPlaces` as constants is just as flexible because if the values do need to change for some reason in the future, we just change the values of the constants in a single place (though any fixed point numbers using the old scaling factor stored in the DB would need to be cleaned out). Plus, this is actually safer, because we want to avoid the possibility of accidentally using different values for `scalingFactor` and `maxDecimalsPlaces` in different areas of the codebase.

common/db.go
yondonfu: An alternative is to declare `fpMaxPrice` within the `else` block prior to calling `RatToFixed` - see the code snippet in [this comment](https://github.com/livepeer/go-livepeer/pull/939#discussion_r298225478).

common/util_test.go
```diff
@@ -62,3 +64,22 @@ func TestProfilesToHex(t *testing.T) {
 	compare([]ffmpeg.VideoProfile{ffmpeg.P240p30fps16x9})
 	compare([]ffmpeg.VideoProfile{ffmpeg.P240p30fps16x9, ffmpeg.P360p30fps16x9})
 }
+
+func TestRatToFixed(t *testing.T) {
```
yondonfu: Let's add tests for the following cases:

- The float representation of `rat` has > 3 decimal places
- The float representation of `rat` has a non-zero integer part and a non-zero fractional part i.e. 1.79, 3.78

common/db.go
yondonfu: > I thought this would have been cleaner / clearer than changing every selectOrchs statement to take in nil/empty filter struct,

We could avoid this while also minimizing code duplication a bit by extracting the boilerplate code into a helper function.

Ex.

```golang
func (db *DB) SelectOrchs() ([]*DBOrch, error) {
    // Other logic

    return db.selectOrchs(DBOrchFilter{})
}

func (db *DB) FilterOrchs(filter DBOrchFilter) ([]*DBOrch, error) {
    // Other logic

    return db.selectOrchs(filter)
}

func (db *DB) selectOrchs(filter DBOrchFilter) ([]*DBOrch, error) {
    var (
        rows *sql.Rows
        err error
    )

    if filter == DBOrchFilter{} || filter.MaxPrice == nil {
        rows, err = db.selectOrchs.Query()
    } else {
        var fpMaxPrice int64
        // Convert maxPrice -> fpMaxPrice
        rows, err = db.filterOrchs.Query(fpMaxPrice)
    }
    
    // Error handling, row parsing, etc.
    // Return orchs
}
```

common/db_test.go
```diff
@@ -197,6 +198,54 @@ func TestSelectUpdateOrchs_AddingMultipleRows_NoError(t *testing.T) {
...
+	defer dbh.Close()
+	defer dbraw.Close()
+	require := require.New(t)
+	assert := assert.New(t)
+	require.Nil(err)
+
+	for i := 0; i < 10; i++ {
+		orch := NewDBOrch("https://127.0.0.1:"+strconv.Itoa(8936+i), string(pm.RandBytes(32)))
+		orch.PricePerPixel, err = RatToFixed(big.NewRat(1, int64(5+i)), 1000, 3)
+		t.Logf("index %v:  %v\n", i, orch)
```
yondonfu: Do we need this log here?

common/db_test.go
```diff
@@ -197,6 +198,54 @@ func TestSelectUpdateOrchs_AddingMultipleRows_NoError(t *testing.T) {
...
+	// Passing in a lower price than all orchs returns no orchs
+	orchsFiltered, err = dbh.FilterOrchs(DBOrchFilter{big.NewRat(1, 15)})
+	require.Nil(err)
+	assert.Len(orchsFiltered, 0)
+
+	// Passing in 1/10 returns 5 orchs
+	orchsFiltered, err = dbh.FilterOrchs(DBOrchFilter{big.NewRat(1, 10)})
+	require.Nil(err)
+	assert.Len(orchsFiltered, 5)
+	t.Log(orchsFiltered)
```
yondonfu: Do we need this log here?

discovery/discovery_test.go
```diff
@@ -234,13 +252,11 @@ func TestNewDBOrchestratorPoolCache_GivenListOfOrchs_CreatesPoolCacheCorrectly(t
 	cachedOrchs, err := cacheDBOrchs(node, orchestrators)
 	require.Nil(err)
 	assert.Len(cachedOrchs, 3)
-	assert.Equal(cachedOrchs[1].ServiceURI, addresses[1])
```
yondonfu, line 237: Why are these tests going away?

discovery/discovery_test.go
```diff
@@ -416,13 +453,11 @@ func TestCachedPool_AllOrchestratorsTooExpensive_ReturnsEmptyList(t *testing.T)
 	cachedOrchs, err := cacheDBOrchs(node, orchestrators)
 	require.Nil(err)
 	assert.Len(cachedOrchs, 50)
-	assert.Equal(cachedOrchs[1].ServiceURI, addresses[1])
```
yondonfu, line 419: Why are these tests going away?

discovery/db_discovery.go
yondonfu: I agree with @j0sh - I think it makes sense to interpret `Size()` as the size of the orchestrator pool that can be selected. In this change, while all Os are cached, we're only selecting from Os that pass the filter.

common/util.go
j0sh: The values of scalingFactor/maxDecimalPlaces are fine to me, the idea is more towards making the function itself misuse-resistant. (Also, as @yondonfu mentioned, changing this in a backwards-compatible manner would be difficult.) Renaming the function may also help clarify its intended usage. In particular, this is what I was thinking:

```
func PriceToFixed(price *big.Rat) {
  scalingFactor := 1000
  maxDecimalPlaces := 3
  ... rest of the code ...
}
```

Later, if we have a need for further specialization on those params, we can split into helpers as needed - exposing those right now seems premature, and in particular the `maxDecimalPlaces` param is really just the implementation leaking through the API.

common/util.go
j0sh: The intermediate string representation and parsing that is what's painful:
rat -> float -> string -> parse string -> int

Glancing through the docs again, could we do this instead:

ret, _ := new(big.Float).SetRat(scaled).Int64()

https://golang.org/pkg/math/big/#Float.SetRat

discovery/db_discovery.go
j0sh: The old `SelectOrch` *does* filter out rows based on date; it doesn't return all rows unconditionally.

If you look at the [actual usage](https://github.com/livepeer/go-livepeer/blob/master/server/broadcast.go#L162) of the Size function then it becomes clear we should be maintaining the same parameters, since the pool size is used to determine how many Os to select (and we don't want to over-select for Os that we shouldn't be using).

common/util.go
kyriediculous: Correct me if I'm wrong but `FloatString` despite its name is not actually a float conversion. 

Rather rat -> buffer -> string -> int64

common/db.go
kyriediculous: Oh yeah sorry I meant to say allocate. But I was trying to say the same as in that comment😅.

common/db.go
j0sh: In practice, we won't be using `SelectOrchs()` anywhere else (other than testing) so I have a preference for keeping things lean and removing unused code. Using `SelectOrchs(filter)` is clear enough to me, but I'll leave that up to you.

common/util.go
kyriediculous: In my opinion utility functions should be named according to the functionality they provide, regardless of context , in this case converting a big.Rat to a fixed point. 

RatToFixed can be used to convert any big.Rat to a fixed point representation, not limited to only pricing information represented as a big.Rat. 

If we think we'll only be converting those price specific big.Rats to fixed points ever, then I would agree hardcoding the scalingfactor/maxDecimals and renaming it to `PriceToFixed`
That's more the thought process that was on my mind when writing this the way I did.

common/util.go
j0sh: Sure, just that the stringification / parsing step could be removed.

discovery/discovery_test.go
kyriediculous: If an `OrchInfo `request to` addresses[2]` returns before the one to` addresses[1]` then `addresses[2]` will be inserted into the database before `addresses[1]`

So the tests don't pass. Beforehand we were just looping over an array sequentially to correct them to the right format. Now we have an http request on a goroutine in between.

discovery/db_discovery.go
j0sh: I'm a bit uneasy about this for a few (admittedly abstract) reasons.

If there's bad data on the blockchain, then there's bad data on the blockchain. If the O registers itself normally (eg, via CLI) then the URL will always have the https prefix. I'm nervous about things interacting unexpectedly [1] or downstream ecosystem impacts [2] from special-cases like these if we take the default behavior of "fixing" URLs. The brevity of not needing a helper function is also nice... Anyway, I'll leave this one up to you.

[1] Can't think of a concrete case right now, but what makes me nervous is this doesn't prove the absence of such cases.

[2] An example: third-party transcoder registration software starts omitting the prefix, which may break other implementations of the Livepeer node (even though there aren't any right now), and forces subsequent implementations to accommodate this special case; implicitly becoming a required part of the "contract" for the ServiceURI

kyriediculous: Hmm I guess the same functionality here 

https://github.com/livepeer/go-livepeer/blob/17701ca9a7bfb8b84de08da9fae1223ae2053b29/discovery/discovery.go#L35

is then rather to force CLI supplied orchestrator URIs to have the prefix in offchain mode, not per se URIs fetched on chain ?

common/util.go
yondonfu: > Later, if we have a need for further specialization on those params, we can split into helpers as needed - exposing those right now seems premature

Agreed. Even if we do end up needing to use different values for `scalingFactor` (i.e. if we need different fixed point number types with difference precision in other areas of the codebase) in the future, we would probably want to abstract `scalingFactor` away from the caller to avoid programmer error (i.e. passing in an incorrect value for `scalingFactor`).

This is premature, but in the future it could look something like:

```
func PriceToFixed(price *big.Rat) {
    scalingFactor := 1000
    return ratToFixed(price, scalingFactor)
}
```

In this future scenario, where there might be a use case for different values of `scalingFactor`, we could refactor and move the current logic of `RatToFixed()` down one level on the abstraction ladder into the `ratToFixed()` helper which *does* expose `scalingFactor` as a parameter. However, `ratToFixed()` would only ever be used internally within the package, and package users would only use the exported functions that abstract `scalingFactor` away.

For now, in the interest of avoiding premature optimization, IMO just having `PriceToFixed(rat *big.Rat)` while omitting `scalingFactor` from the parameters seems fine.

discovery/discovery_test.go
yondonfu: Ah I see. Should we check that the orchs stored in the DB are actually the ones we expect then (i.e. all the fields match the ones we define in this test)?

discovery/db_discovery.go
j0sh: Ah yeah, you're right - IIRC it was meant as a UX affordance and probably belongs better where we handle the CLI flags in [cmd/livepeer.go](https://github.com/livepeer/go-livepeer/blob/17701ca9a7bfb8b84de08da9fae1223ae2053b29/cmd/livepeer/livepeer.go#L157-L165)

Still up to you - especially since we already have this fixup another place, I'm slightly less uneasy (but still a little bit).

common/db.go
kyriediculous: Just out of curiosity. What's the idiomatic way?

A. Using a value arg and passing in an empty struct when no filter is used 
B. Using a pointer and passing in nil when no filter is used 

I'm thinking passing a value is okay since we're not modifying the underlying struct and the struct is not going to be very large in the foreseeable future? But passing in an empty struct as argument seems 'dirtier' than passing in nil although that might just be visually?

j0sh: Structs should typically be passed in as a pointer (option B) so we don't incur the cost of copying the entire struct each time the parameter is called.

At least, that's how it is in most languages with some control over addressing; not sure if golang's semantics are different in that regard.

kyriediculous: Let me repost my comment after second thought. 

What is the idiomatic go way? 

A. Pointer and passing in nil when filter is not being used 
B. Value and empty struct when filter is not used 

I often read if it's a struct pass a pointer but that's just not a good argument I think. 
I mean we're not modifying the struct and it isn't very large either so a passing it as a value is okay and using the stack means it doesn't have to be garbage collected. 
It's just that visually being able to pass `nil` is a bit neater? 

Preferences?

j0sh: Even if the struct is small now, we don't know if it'll always be small. In the interests of API stability, I'd go with option A.

But I'd also be totally fine with reverting to the original approach of passing in a `big.Rat` directly, because it's still unclear how we'd compose various config options, and there are some aspects that could be "configurable" on paper but aren't right now (filtering by date).

kyriediculous: Gotcha, thanks.

discovery/discovery_test.go
kyriediculous: Instead of checking 1 element we are now checking all elements are within the returned slices.  

```
	cachedOrchs, err := cacheDBOrchs(node, orchestrators)
	require.Nil(err)
	assert.Len(cachedOrchs, 3)
	for _, o := range orchestrators {
		dbO := ethOrchToDBOrch(o)
		dbO.PricePerPixel, _ = common.PriceToFixed(big.NewRat(999, 1))

		assert.Contains(cachedOrchs, dbO)
	}

	// ensuring orchs exist in DB
	orchs, err := node.Database.SelectOrchs(nil)
	require.Nil(err)
	assert.Len(orchs, 3)
	for _, o := range orchestrators {
		dbO := ethOrchToDBOrch(o)

		assert.Contains(orchs, dbO)
	}
```

discovery/db_discovery.go
kyriediculous: I left them in place for now. But could still remove this after re-review.

common/util.go
```diff
@@ -151,17 +150,17 @@ func GetConnectionAddr(ctx context.Context) string {
...
 	}
-	scaled := new(big.Rat).Mul(rat, big.NewRat(scalingFactor, 1))
-	s := scaled.FloatString(maxDecimalPlaces)
-	intPart, err := strconv.ParseInt(strings.Split(s, ".")[0], 10, 64)
-	if err != nil {
-		return 0, err
+	scaled := new(big.Rat).Mul(price, big.NewRat(scalingFactor, 1))
+	fp, _ := new(big.Float).SetRat(scaled).Int64()
+	if fp == 0 {
+		return 0, fmt.Errorf("Too many decimal places")
```
yondonfu: I don't think we should return an error here since 0 is a valid fixed point number. Furthermore, the error string indicates that the decimal representation of the price passed in has "too many decimal places", but there could be other scenarios besides when `fp == 0` where the decimal representation of the price passed in has "too many decimal places" i.e. 1.666666666

discovery/db_discovery.go
```diff
@@ -93,7 +93,7 @@ func (dbo *DBOrchestratorPoolCache) GetOrchestrators(numOrchestrators int) ([]*n
 }
 
 func (dbo *DBOrchestratorPoolCache) Size() int {
-	orchs, err := dbo.node.Database.SelectOrchs()
+	orchs, err := dbo.node.Database.SelectOrchs(nil)
```
yondonfu: We should pass in a filter with the max price here

yondonfu (changes requested):

discovery/db_discovery.go
kyriediculous: Right, was thinking that after the discussion earlier but wasn't sure. 
fixed in https://github.com/livepeer/go-livepeer/pull/939/commits/91536e4e33b71ae4cc5bcf1078ddda29ec399bff

common/util.go
kyriediculous: Fixed in https://github.com/livepeer/go-livepeer/pull/939/commits/91536e4e33b71ae4cc5bcf1078ddda29ec399bff

discovery/db_discovery.go
```diff
@@ -45,7 +48,7 @@ func NewDBOrchestratorPoolCache(node *core.LivepeerNode) *DBOrchestratorPoolCach
 }
 
 func (dbo *DBOrchestratorPoolCache) GetURLs() []*url.URL {
-	orchs, err := dbo.node.Database.SelectOrchs()
+	orchs, err := dbo.node.Database.SelectOrchs(nil)
```
j0sh: This should have a filter as well (missed this case during our earlier discussions)

kyriediculous: Fixed in https://github.com/livepeer/go-livepeer/pull/939/commits/34be3ca41fbaed6a94a78b807eada7fd2508e370

kyriediculous:
Resolved conflicts in `common/util.go` (accepted both changes)

common/util_test.go
```diff
@@ -90,6 +90,6 @@ func TestRatToFixed(t *testing.T) {
 
 	// more than 3 decimals returns an error and 0
```
yondonfu: We should update this comment too since we're no longer returning an error for this case

yondonfu (commented):
We can avoid the extra merge commit by doing `git rebase master` (and we need to do a rebase anyway to squash the fixup commits) instead of `git merge master`. Avoiding the extra merge commit is not the biggest deal, but its always nice to maintain a cleaner Git commit history when possible.

common/util_test.go
kyriediculous: adjusted in https://github.com/livepeer/go-livepeer/pull/939/commits/c4631f7cd9d9706f31bdfb76981c380796b98ec0

kyriediculous:
> We can avoid the extra merge commit by doing `git rebase master` (and we need to do a rebase anyway to squash the fixup commits) instead of `git merge master`. Avoiding the extra merge commit is not the biggest deal, but its always nice to maintain a cleaner Git commit history when possible.

Ok I didn't know , my eagerness got ahead of me. I think the merge commit will be left out though when we rebase?

yondonfu:
> Ok I didn't know , my eagerness got ahead of me.

No worries! Thanks to Git's flexibility we can always make history corrections if we want to.

> I think the merge commit will be left out though when we rebase?

Yep `git rebase -i master` gets us most of the way there. However, you'll probably run into a bunch of conflicts during the rebase (the ones in `common/util.go` are to be expected since #926 made changes there). I took a look at these locally and they're a bit hairy because your fixup commits target d319665 but they include code based on the filter struct that is only introduced in the subsequent commit de00fb6. For future reference, the rebase ends up being cleaner if fixup commits don't include code dependencies that are not introduced until commits that follow the commit referenced by the fixup. For this PR, the easiest approach is probably just to squash everything into a single commit - when you run an interactive rebase via `git rebase -i master` you can move all the fixup commits to be after de00fb6 and you can squash de00fb6 into d319665 as well so you end up with a single commit.

yondonfu (approved):
LGTM after the rebase mentioned in [this comment](https://github.com/livepeer/go-livepeer/pull/939#issuecomment-506777624). Before merging, I'd check with @j0sh in case he has any additional thoughts since he's provided a bunch of feedback already.

j0sh (approved):
🚢

PR #944, opened 2019-06-28
Broadcaster ticket param validation

### LINKED ISSUE #831: pm: Implement broadcaster strategy for determining whether ticket parameters are acceptable

### LINKED ISSUE #927: pm: B ticket param validation mid-stream
While #831 will implement B validation of ticket params during O discovery, B should also validate new ticket params received when O sends B an updated OrchestratorInfo message mid-stream. If the ticket param check for the updated OrchestratorInfo message fails, then B should switch away from the current O.

yondonfu:

This PR implements ticket param validation for a broadcaster. The ticket params that are most important for a broadcaster to validate are the ticket `faceValue` and the ticket `EV`.

See below for a summary of a description from from #831

The ticket EV is important to validate because the higher the `EV`, the more value B must put at risk prior to O performing any transcoding. This is because we are planning on treating the ticket `EV` as a minimum credit requirement enforced by O. B has to send a ticket worth `EV` to top off its balance with O before receiving transcoded results.

The ticket `faceValue` is important to validate because B wants to avoid having its reserve frozen (inaccessible for a fixed period of time) by overspending from its deposit. The higher the ticket `faceValue`, the higher the chance that B will overspend from its deposit by sending out successive winning tickets that not only draw from B's deposit, but also draw from its reserve (thereby freezing it). While we cannot completely guarantee that B will not overspend from its deposit, we can try to reduce the probability by checking for an acceptable `faceValue` that is not too high.

One approach could involve B defining a `depositMultiplier` config option. `depositMultiplier can be used to derive `maxFaceValue = deposit / depositMultiplier`. 

B would validate ticket params in the following scenarios:

- During O discovery upon receiving an `OrchestratorInfo` message containing ticket params
- When O sends an updated `OrchestratorInfo` message containing new ticket params
- When creating a ticket. Since B's deposit will be consistently decreasing, the `maxFaceValue` will be consistently decreasing as well. So, whenever B creates a ticket it should check that the ticket params it is currently using are still acceptable given its current deposit - in other words, B should check that the face value of a new ticket is less than its most up to date `maxFaceValue`

- 9c8384c: Implements ticket param validation on `pm.Sender`. When a broadcaster node is initialized, the `-maxTicketEV` and `-depositMultiplier` flags can be used to pass in values other than the defaults.
- 81ea445: Implements ticket param validation whenever B receives an updated `OrchestratorInfo` message
- 0db1672: Implements ticket param validation during O discovery (Os that fail validation will be filtered out)

Added additional unit tests.

Fixes #831 
Fixes #927

**Checklist:**

### DIFF
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -104,6 +104,10 @@ func main() {
 	gasPrice := flag.Int("gasPrice", 0, "Gas price for ETH transactions")
 	initializeRound := flag.Bool("initializeRound", false, "Set to true if running as a transcoder and the node should automatically initialize new rounds")
 	ticketEV := flag.String("ticketEV", "1000000000", "The expected value for PM tickets")
+	// Broadcaster max acceptable ticket EV
+	maxTicketEV := flag.String("maxTicketEV", "10000000000", "The maximum acceptable expected value for PM tickets")
+	// Broadcaster deposit multiplier to determine max acceptable ticket faceValue
+	depositMultiplier := flag.Int("depositMultiplier", 1000, "The deposit multiplier used to determine max acceptable faceValue for PM tickets")
 
 	// Orchestrator base pricing info
 	pricePerUnit := flag.Int("pricePerUnit", 0, "The price per 'pixelsPerUnit' amount pixels")
@@ -396,9 +400,24 @@ func main() {
 		}
 
 		if n.NodeType == core.BroadcasterNode {
+			ev, _ := new(big.Rat).SetString(*maxTicketEV)
+			if ev == nil {
+				panic(fmt.Errorf("-maxTicketEV must be a valid rational number, but %v provided. Restart the node with a valid value for -maxTicketEV", *maxTicketEV))
+			}
+
+			if ev.Cmp(big.NewRat(0, 1)) < 0 {
+				panic(fmt.Errorf("-maxTicketEV must not be negative, but %v provided. Restart the node with a valid value for -maxTicketEV", *maxTicketEV))
+			}
+
+			if *depositMultiplier <= 0 {
+				panic(fmt.Errorf("-depositMultiplier must be greater than 0, but %v provided. Restart the node with a valid value for -depositMultiplier", *depositMultiplier))
+			}
+
 			// TODO: Initialize Sender with an implementation
 			// of RoundsManager that reads from a cache
-			n.Sender = pm.NewSender(n.Eth, n.Eth)
+			// TODO: Initialize Sender with an implementation
+			// of SenderManager that reads from a cache
+			n.Sender = pm.NewSender(n.Eth, n.Eth, n.Eth, ev, *depositMultiplier)
 
 			if *pixelsPerUnit <= 0 {
 				// Can't divide by 0
--- a/discovery/db_discovery.go
+++ b/discovery/db_discovery.go
@@ -8,10 +8,12 @@ import (
 	"strings"
 	"time"
 
+	ethcommon "github.com/ethereum/go-ethereum/common"
 	"github.com/livepeer/go-livepeer/common"
 	"github.com/livepeer/go-livepeer/core"
 	lpTypes "github.com/livepeer/go-livepeer/eth/types"
 	"github.com/livepeer/go-livepeer/net"
+	"github.com/livepeer/go-livepeer/pm"
 	"github.com/livepeer/go-livepeer/server"
 
 	"github.com/golang/glog"
@@ -75,6 +77,12 @@ func (dbo *DBOrchestratorPoolCache) GetOrchestrators(numOrchestrators int) ([]*n
 	}
 
 	pred := func(info *net.OrchestratorInfo) bool {
+		if dbo.node.Sender != nil {
+			if err := dbo.node.Sender.ValidateTicketParams(pmTicketParams(info.TicketParams)); err != nil {
+				return false
+			}
+		}
+
 		price := server.BroadcastCfg.MaxPrice()
 		if price != nil {
 			return big.NewRat(info.PriceInfo.PricePerUnit, info.PriceInfo.PixelsPerUnit).Cmp(price) <= 0
@@ -191,4 +199,18 @@ func ethOrchToDBOrch(orch *lpTypes.Transcoder) *common.DBOrch {
 		return nil
 	}
 	return common.NewDBOrch(orch.ServiceURI, orch.Address.String())
+}
+
+func pmTicketParams(params *net.TicketParams) *pm.TicketParams {
+	if params == nil {
+		return nil
+	}
+
+	return &pm.TicketParams{
+		Recipient:         ethcommon.BytesToAddress(params.Recipient),
+		FaceValue:         new(big.Int).SetBytes(params.FaceValue),
+		WinProb:           new(big.Int).SetBytes(params.WinProb),
+		RecipientRandHash: ethcommon.BytesToHash(params.RecipientRandHash),
+		Seed:              new(big.Int).SetBytes(params.Seed),
+	}
 }
--- a/pm/sender.go
+++ b/pm/sender.go
@@ -20,7 +20,8 @@ type Sender interface {
 	// and signature over the new ticket for a given session ID
 	CreateTicket(sessionID string) (*Ticket, *big.Int, []byte, error)
 
-	// Later: support receiving new recipientRandHash values mid-stream
+	// ValidateTicketParams checks if ticket params are acceptable
+	ValidateTicketParams(ticketParams *TicketParams) error
 }
 
 type session struct {
@@ -32,15 +33,22 @@ type session struct {
 type sender struct {
 	signer        Signer
 	roundsManager RoundsManager
+	senderManager SenderManager
+
+	maxEV             *big.Rat
+	depositMultiplier int
 
 	sessions sync.Map
 }
 
 // NewSender creates a new Sender instance.
-func NewSender(signer Signer, roundsManager RoundsManager) Sender {
+func NewSender(signer Signer, roundsManager RoundsManager, senderManager SenderManager, maxEV *big.Rat, depositMultiplier int) Sender {
 	return &sender{
-		signer:        signer,
-		roundsManager: roundsManager,
+		signer:            signer,
+		roundsManager:     roundsManager,
+		senderManager:     senderManager,
+		maxEV:             maxEV,
+		depositMultiplier: depositMultiplier,
 	}
 }
 
@@ -63,6 +71,10 @@ func (s *sender) CreateTicket(sessionID string) (*Ticket, *big.Int, []byte, erro
 		return nil, nil, nil, errors.Errorf("cannot create a ticket for an unknown session: %v", sessionID)
 	}
 	session := tempSession.(*session)
+
+	if err := s.ValidateTicketParams(&session.ticketParams); err != nil {
+		return nil, nil, nil, err
+	}
 
 	senderNonce := atomic.AddUint32(&session.senderNonce, 1)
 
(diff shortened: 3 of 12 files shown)


discovery/db_discovery.go
```diff
@@ -75,6 +77,22 @@ func (dbo *DBOrchestratorPoolCache) GetOrchestrators(numOrchestrators int) ([]*n
 	}
 
 	pred := func(info *net.OrchestratorInfo) bool {
+		protoParams := info.TicketParams
```
kyriediculous: Since this package imports the server package and this exact same workflow of conversion between protobuf `net.TicketParams` and `pm.TicketParams` types happens there in two files as well (`broadcast.go` and `segment_rpc.go`) , can we avoid code duplication and move this into a helper? 

```
getPMTicketParams(orchInfo *net.OrchestratorInfo) *pm.TicketParams {
  if orchInfo.TicketParams != nil {
    return &pm.TicketParams{
	Recipient:         ethcommon.BytesToAddress(orchInfo.TicketParams.Recipient),
	FaceValue:         new(big.Int).SetBytes(orchInfo.TicketParams.FaceValue),
	WinProb:           new(big.Int).SetBytes(orchInfo.TicketParams.WinProb),
	RecipientRandHash: ethcommon.BytesToHash(orchInfo.TicketParams.RecipientRandHash),
	Seed:              new(big.Int).SetBytes(orchInfo.TicketParams.Seed),
    }
  }
  return nil
}
```

cmd/livepeer/livepeer.go
```diff
@@ -393,9 +397,24 @@ func main() {
...
+			if *depositMultiplier <= 0 {
+				panic(fmt.Errorf("-depositMultiplier must be greater than 0, but %v provided. Restart the node with a valid value for -depositMultiplier", *depositMultiplier))
+			}
+
 			// TODO: Initialize Sender with an implementation
 			// of RoundsManager that reads from a cache
-			n.Sender = pm.NewSender(n.Eth, n.Eth)
+			// TODO: Initialize Sender with an implementation
+			// of SenderManager that reads from a cache
+			n.Sender = pm.NewSender(n.Eth, n.Eth, n.Eth, ev, *depositMultiplier)
```
kyriediculous, line 420: This argument list is starting to look weirder and weirder by the day to me 😅 . But I can see why it's done this way, the EthClient implements the respective interfaces in the constructor parameter list. I'm just wondering if there isn't a better way to do this such as dividing up the EthClient into those structs that implement those respective interfaces. Rather than having their implementation defined on the client struct directly. Altough introducing another level of wrappers might not be a good idea here. 

I'm not sure, it just looks odd. This structure isn't new to this PR either so my comment is slightly out of place here. This is just me thinking out loud.

pm/sender_test.go
```diff
@@ -78,6 +78,51 @@ func TestCreateTicket_GivenNonexistentSession_ReturnsError(t *testing.T) {
...
+
+	ticketParams := TicketParams{
+		Recipient:         RandAddress(),
+		FaceValue:         big.NewInt(202),
+		WinProb:           new(big.Int).Div(maxWinProb, big.NewInt(2)),
+		Seed:              big.NewInt(3333),
+		RecipientRandHash: RandHash(),
+	}
+	sessionID := sender.StartSession(ticketParams)
+	_, _, _, err := sender.CreateTicket(sessionID)
```
kyriediculous, line 103: Should we check that the returned ticket pointer is in fact `nil` or would that be assertion overkill?

pm/sender.go
```diff
@@ -32,15 +33,22 @@ type session struct {
...
+
+	maxEV             *big.Rat
+	depositMultiplier int
 
 	sessions sync.Map
 }
 
 // NewSender creates a new Sender instance.
-func NewSender(signer Signer, roundsManager RoundsManager) Sender {
+func NewSender(signer Signer, roundsManager RoundsManager, senderManager SenderManager, maxEV *big.Rat, depositMultiplier int) Sender {
```
kyriediculous, line 45: Again a slightly out of place comment since this functionality was already in place before the PR (consider it a question), but shouldn't we be returning a concrete struct or pointer here, not an interface value, according to the [go wiki](https://github.com/golang/go/wiki/CodeReviewComments#interfaces) ?

> Go interfaces generally belong in the package that uses values of the interface type, not the package that implements those values. The implementing package should return concrete (usually pointer or struct) types

discovery/db_discovery.go
yondonfu: Fixed:

- 8e6c3f5: Use a helper in `discovery/db_discovery.go`
- 7fcae8c: Use a helper in `server/segment_rpc.go`
- c6b010e: Use helper in `server/broadcast.go`

I ended up defining the same `pmTicketParams()` helper in both the `discovery` and `server` packages for now. I know this introduces a bit of duplication across packages, but I was a little hesitant about defining the helper in the `server` package and then importing it in the `discovery` package. While the `discovery` package does already depend on the `server` package, it felt a bit weird to import a function from the `server` package that uses the `pm` and `net` packages when those packages are already imported in the `discovery` package. Plus, we should probably avoid turning `server` into a package that is depended on for exporting these types of helpers (i.e. converting `net` types into other types). So, I ended up prioritizing minimizing the amount of stuff that needs to be imported at the expense of a bit of code duplication.

pm/sender.go
yondonfu: Yeah returning a concrete type here instead of an interface defined by the `pm` package probably would be better. Another annoyance that arises from the current approach is that in unit tests we have to do a type assertion on the interface if we want to run checks on unexported fields.

pm/sender_test.go
yondonfu: Seems ok not to check the ticket pointer right now since the expectation is that if the error is non-nil then none of the other values are safe to use for the `CreateTicket()` method. WDYT?

cmd/livepeer/livepeer.go
yondonfu: Yeah I hear you. This should eventually change though as each of those interfaces should be implemented by structs that cache values instead of executing a RPC request per method call.

discovery/db_discovery.go
kyriediculous: >  it felt a bit weird to import a function from the server package that uses the pm and net packages when those packages are already imported in the discovery package. Plus, we should probably avoid turning server into a package that is depended on for exporting these types of helpers

Great point and I should have thought of that considering my packet structure research.

pm/sender_test.go
kyriediculous: Seems reasonable to me 👍

cmd/livepeer/livepeer.go
kyriediculous: I can live with that 👍

kyriediculous (approved):
LGTM 🚀 

Maybe that Josh has some thoughts , so your call of course if you want to let the PR sit until tomorrow.

j0sh:
This looks pretty good at a first glance.

What happens if validation fails on the broadcaster?

~~From a quick look it seems that `genPayment` will just print an error and continue even if there was a failure (genPayment -> CreateTicket -> ValidateTicketParams) . If an error is received from `genPayment`, it should be safe to return the error immediately; the broadcaster will discard the O from the working set and retry the segment with another O. This saves time over waiting for the O's grace period to expire, which would eventually lead to a hard fail from the O side.~~ This is exactly what we do; missed this

With the `updateOrchestratorInfo` call in `SubmitSegment`  - given a grace period on O, could we get valid transcode results along with updated (but invalid) pricing info?

If we update the PM params regardless of validity, we could receive the transcode results, then lazy-unload the session next time the O is selected - genPayment would fail and remove the O from the working set immediately.

We could also try and do this unloading eagerly within SubmitSegment (and the SubmitSegment caller), by returning the PM error along with the transcode results: we can use the results, but stop working with the O for further segments. This might seem strange, but I suspect this won't be the only situation where we want this type of behavior.

I do have some concerns about mutating the session object within updateOrchestratorInfo - up until recently, the session had been immutable - but as of now I think we're only using it on a single thread, so it's probably OK for now. We can't be assured that'll always be the case though.

yondonfu:
> With the updateOrchestratorInfo call in SubmitSegment - given a grace period on O, could we get valid transcode results along with updated (but invalid) pricing info?

Yeah that's possible. O could receive a ticket with outdated params, credit to B's balance as a part of the grace period and return a valid transcode result with the updated params. The updated params *could* then be unacceptable to B based on its configured `maxEV` and `maxFaceValue`.

> If we update the PM params regardless of validity, we could receive the transcode results, then lazy-unload the session next time the O is selected - genPayment would fail and remove the O from the working set immediately.

I like this - I'm all for removing the extra check in `updateOrchestratorInfo` if we can.

> We could also try and do this unloading eagerly within SubmitSegment (and the SubmitSegment caller), by returning the PM error along with the transcode results: we can use the results, but stop working with the O for further segments. 

By PM error do you mean the error returned by `genPayment()` if the params for the session are invalid?

> I do have some concerns about mutating the session object within updateOrchestratorInfo - up until recently, the session had been immutable - but as of now I think we're only using it on a single thread, so it's probably OK for now. We can't be assured that'll always be the case though.

What if we did the following:

- Change `SubmitSegment()` to return `(*net.TranscodeResult, error)` instead of `(*net.TranscodeData, error)`. If O sent back an updated `OrchestratorInfo` message (could contain updated PM params), then this data will be carried in the result returned by `SubmitSegment()`
- Add a `addSession()` method to `BroadcastSessionsManager` that accepts `*net.OrchestratorInfo` as a parameter. This method will create a new `*BroadcastSession` and add it to the session list
- In [transcodeSeg()](https://github.com/livepeer/go-livepeer/blob/master/server/broadcast.go#L339), if `SubmitSegment()` does not return an error and the `OrchestratorInfo` field in the result is non-nil, then call `addSession()` with the `OrchestratorInfo` field to create a new session based on the updated O info. If the field is nil, then call `completeSession()` which will add the existing session

j0sh:
> By PM error do you mean the error returned by genPayment() if the params for the session are invalid?

That note related to "eagerly unloading" the O from B's working set was referring to returning a PM error from `updateOrchestratorInfo` while still returning the transcode results from SubmitSegment (if any).

We can just stay with the lazy-unloading approach for now, although it does make reasoning about the behavior a little less straightforward. Maybe add a note in `updateOrchestratorInfo` describing the behavior, in lieu of the validity check?

> Add a addSession() method to BroadcastSessionsManager that accepts *net.OrchestratorInfo as a parameter. This method will create a new *BroadcastSession and add it to the session list

I think we should be able to use the following trick without enlarging the BroadcastSessionManager API:

* Create an entirely new BroadcastSession with updated fields rather than mutating an existing session (would need to copy other fields as necessary such as the OrchestratorInfo.Transcoder which is the primary lookup key within the session manager)
* Use this new BroadcastSession in place of the old session
* New BroadcastSession is passed in to `completeSession` (at the same place it is now; no changes)
* Within `completeSession`, look up the session in the map and check for physical equality (compare pointer addresses).
* If the sessions differ, replace the existing session in the map
* Append new session to list as we do now.

We could also do something similar via  `replace(newSession, oldSession)` to subsume `completeSession` for a more explicit mechanism. This may give us some advantages if the lookup key changes (eg, Transcoder address changes due to O's internal load balancing). For robustness we probably would need to make `completeSession` append using the session as it exists in the map, rather than the passed-in session.

This approach isn't perfect - it doesn't really *prevent* mutations - but at least it should make such mutations harder to do concurrently on the same session.

We also don't need to implement any of this right now, but tracking it via an issue would be good.

yondonfu:
> Maybe add a note in updateOrchestratorInfo describing the behavior, in lieu of the validity check?

Removed params validation in `updateOrchestratorInfo` and added a comment describing the expected behavior in 91ed8d6

> We also don't need to implement any of this right now, but tracking it via an issue would be good.

Tracking in #961

j0sh (approved):
🚢

Issue #919, opened 2019-06-17
devtool script should run in CI; required for PR approval


mkrufky:
The devtool script is very helpful for development and testing.  We should always be sure that any changes to procedure are accounted for in this script.

kyriediculous:
The only thing I'm not certain about is , should we run the generated  bash scripts in `devtool.go` as part of this additional check.
Considering flags and their respective arguments should already be covered by `test_args.sh`

j0sh:
The tricky part here is that devtool expects a geth running somewhere with the Livepeer contracts deployed. If we can figure out how to get that working for CI then that'd be great.

kyriediculous:
I believe @mkrufky ran this against a docker image Ivan made (#918) . We should be able to run the script against the image, no?

yondonfu:
We should be able pull down the [geth-with-livepeer-protocol](https://github.com/livepeer/docker-livepeer/tree/master/geth-with-protocol) Docker image that is now hosted on the Livepeer DockerHub.

There is also an open [issue](https://github.com/livepeer/protocol/issues/295) to auto-build and push a new image whenever a new contracts PR lands on the `pm` branch.

ericxtang:
@mk-livepeer can we close this issue now that the PR is merged?

mkrufky:
No, the PR that was merged mentions this issue but does not close it.

PR #949, opened 2019-06-30
devtool: allow passing service host as optional arg, `svchost`

mk-livepeer:
* allow passing service host other than the default, `127.0.0.1` as optional arg, `svchost`
* necessary for container orchestration tools such as `docker-compose` and `kubernetes`

relates-to: #919


darkdarkdragon (approved):
LGTM 🚢

PR #956, opened 2019-07-02
core, pm: cmd: O processes payments and checks credit

### LINKED ISSUE #895: O processes payment from B and checks that minimum credit requirement is met

kyriediculous:
This is a work in progress PR to receive feedback on a first implementation attempt as outlined in #895 

This PR allows an orchestrator to add processed tickets (received and acceptable errors) to a stream's credit balance and checks whether the minimum credit requirements on a per stream basis are met in `ProcessPayment` 

- Added a new type alias `Accounts` with `Credit`, `Debit`, `Balance` and `Cleanup` methods 
- Added a new method to `pm.Recipient` to retrieve the desired EV 
- Added a new `Accounts` field to the `LivepeerNode` and initialized the `Accounts.Balances` mapping in the `NewLivepeerNode` constructor
- Added logic in `ProcessPayment` (TicketSenderParams loop) to add a received ticket's `EV` to the credit balance for a `ManifestID`
- Added logic at the end of `ProcessPayment` , after all tickets are processed to check if the credit requirement for a `ManifestID` meets the orchestrators configured `EV` 
- Start a cleanup routine for `Accounts` on node startup (orchestrator mode only).

Wrote unit tests for the new accounting type alias
Mocked out MaxEV for `ProcessPayment` tests. 

Fixes #895 

**Checklist:**

### DIFF
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -51,8 +51,8 @@ var (
 	txCostMultiplier = 100
 	// The interval at which to poll for gas price updates
 	gpmPollingInterval = 1 * time.Minute
-	// The interval at which to clean up cached max float values for PM senders
-	smCleanupInterval = 1 * time.Minute
+	// The interval at which to clean up cached max float values for PM senders and balances per stream
+	cleanupInterval = 1 * time.Minute
 	// The time to live for cached max float values for PM senders (else they will be cleaned up)
 	smTTL = 3600 // 1 minute
 	// smMaxErrCount is the maximum number of acceptable errors tolerated by a PM recipient for a sender
@@ -371,7 +371,7 @@ func main() {
 			}
 			defer gpm.Stop()
 
-			sm := pm.NewSenderMonitor(n.Eth.Account().Address, n.Eth, gasPriceUpdate, smCleanupInterval, smTTL, smMaxErrCount)
+			sm := pm.NewSenderMonitor(n.Eth.Account().Address, n.Eth, gasPriceUpdate, cleanupInterval, smTTL, smMaxErrCount)
 			// Start sender monitor
 			sm.Start()
 			defer sm.Stop()
@@ -397,6 +397,12 @@ func main() {
 
 			n.Recipient.Start()
 			defer n.Recipient.Stop()
+
+			n.Balances = core.NewBalances(cleanupInterval)
+			// Run n.Accounts.Cleanup() Routine
+			go n.Balances.StartCleanup()
+			// Stop the cleanup routine on program exit
+			defer n.Balances.StopCleanup()
 		}
 
 		if n.NodeType == core.BroadcasterNode {
--- a/core/accounting.go
+++ b/core/accounting.go
@@ -1,0 +1,89 @@
+package core
+
+import (
+	"math/big"
+	"sync"
+	"time"
+)
+
+// Balances holds credit balances on a per-stream basis
+type Balances struct {
+	balances map[ManifestID]*balance
+	mtx      sync.RWMutex
+	ttl      time.Duration
+	quit     chan struct{}
+}
+
+type balance struct {
+	lastUpdate time.Time // Unix time since last update
+	amount     *big.Rat  // Balance represented as a big.Rat
+}
+
+// NewBalances creates a Balances instance with the given ttl
+func NewBalances(ttl time.Duration) *Balances {
+	return &Balances{
+		balances: make(map[ManifestID]*balance),
+		ttl:      ttl,
+		quit:     make(chan struct{}),
+	}
+}
+
+// Credit adds an an amount to the balance for a ManifestID
+func (b *Balances) Credit(id ManifestID, amount *big.Rat) {
+	b.mtx.Lock()
+	defer b.mtx.Unlock()
+	if b.balances[id] == nil {
+		b.balances[id] = &balance{amount: big.NewRat(0, 1)}
+	}
+	b.balances[id].amount.Add(b.balances[id].amount, amount)
+	b.balances[id].lastUpdate = time.Now()
+}
+
+// Debit substracts an amount from the balance for a ManifestID
+func (b *Balances) Debit(id ManifestID, amount *big.Rat) {
+	b.mtx.Lock()
+	defer b.mtx.Unlock()
+	if b.balances[id] == nil {
+		b.balances[id] = &balance{amount: big.NewRat(0, 1)}
+	}
+	b.balances[id].amount.Sub(b.balances[id].amount, amount)
+	b.balances[id].lastUpdate = time.Now()
+}
+
+// Balance retrieves the current balance for a ManifestID
+func (b *Balances) Balance(id ManifestID) *big.Rat {
+	b.mtx.RLock()
+	defer b.mtx.RUnlock()
+	if b.balances[id] == nil {
+		return nil
+	}
+	return b.balances[id].amount
+}
+
+func (b *Balances) cleanup() {
+	for id, balance := range b.balances {
+		b.mtx.Lock()
+		if int64(time.Since(balance.lastUpdate)) > int64(b.ttl) {
+			delete(b.balances, id)
+		}
+		b.mtx.Unlock()
+	}
+}
+
+// StartCleanup is a state flushing method to clean up the balances mapping
+func (b *Balances) StartCleanup() {
+	ticker := time.NewTicker(b.ttl)
+	for {
+		select {
+		case <-ticker.C:
+			b.cleanup()
+		case <-b.quit:
+			return
+		}
+	}
+}
+
+// StopCleanup stops the cleanup loop for Balances
+func (b *Balances) StopCleanup() {
+	close(b.quit)
+}
--- a/core/accounting_test.go
+++ b/core/accounting_test.go
@@ -1,0 +1,92 @@
+package core
+
+import (
+	"math/big"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestEmptyBalances_ReturnsZeroedValues(t *testing.T) {
+	mid := ManifestID("some manifest id")
+	b := NewBalances(5 * time.Second)
+	assert := assert.New(t)
+
+	assert.Nil(b.Balance(mid))
+	assert.Nil(b.balances[mid])
+}
+
+func TestCredit_ReturnsNewCreditBalance(t *testing.T) {
+	mid := ManifestID("some manifest id")
+	b := NewBalances(5 * time.Second)
+	assert := assert.New(t)
+	amount := big.NewRat(100, 1)
+
+	b.Credit(mid, amount)
+	assert.Zero(b.Balance(mid).Cmp(amount))
+}
+
+func TestDebitAfterCredit_SameAmount_ReturnsZero(t *testing.T) {
+	mid := ManifestID("some manifest id")
+	b := NewBalances(5 * time.Second)
+	assert := assert.New(t)
+	amount := big.NewRat(100, 1)
+
+	b.Credit(mid, amount)
+	assert.Zero(b.Balance(mid).Cmp(amount))
+
+	b.Debit(mid, amount)
+	assert.Zero(b.Balance(mid).Cmp(big.NewRat(0, 1)))
+}
+
+func TestDebitHalfOfCredit_ReturnsHalfOfCredit(t *testing.T) {
+	mid := ManifestID("some manifest id")
+	b := NewBalances(5 * time.Second)
+	assert := assert.New(t)
+	credit := big.NewRat(100, 1)
+	debit := big.NewRat(50, 1)
+	b.Credit(mid, credit)
+	assert.Zero(b.Balance(mid).Cmp(credit))
+
+	b.Debit(mid, debit)
+	assert.Zero(b.Balance(mid).Cmp(debit))
+}
+
+func TestBalancesCleanup(t *testing.T) {
+	b := NewBalances(5 * time.Second)
+	assert := assert.New(t)
+
+	// Set up two mids
+	// One we will update after 2*time.Seconds
+	// The other one we will not update before timeout
+	// This should run clean only the second
+	mid1 := ManifestID("First MID")
+	mid2 := ManifestID("Second MID")
+	// Start cleanup loop
+	go b.StartCleanup()
+	defer b.StopCleanup()
+
+	// Fund balances
+	credit := big.NewRat(100, 1)
+	b.Credit(mid1, credit)
+	b.Credit(mid2, credit)
+	assert.Zero(b.Balance(mid1).Cmp(credit))
+	assert.Zero(b.Balance(mid2).Cmp(credit))
+
+	time.Sleep(2 * time.Second)
+	b.Credit(mid1, credit)
+	assert.Zero(b.Balance(mid1).Cmp(big.NewRat(200, 1)))
+
+	time.Sleep(4 * time.Second)
+
+	// Balance for mid1 should still be 200/1
+	assert.NotNil(b.Balance(mid1))
+	assert.Zero(b.Balance(mid1).Cmp(big.NewRat(200, 1)))
+	// Balance for mid2 should be cleaned
+	assert.Nil(b.Balance(mid2))
+
+	time.Sleep(5 * time.Second)
+	// Now balance for mid1 should be cleaned as well
+	assert.Nil(b.Balance(mid1))
+}
--- a/core/livepeernode.go
+++ b/core/livepeernode.go
@@ -66,6 +66,7 @@ type LivepeerNode struct {
 	Transcoder        Transcoder
 	TranscoderManager *RemoteTranscoderManager
 	PriceInfo         *big.Rat
+	Balances          *Balances
 
 	// Broadcaster public fields
 	Sender pm.Sender
--- a/core/orch_test.go
+++ b/core/orch_test.go
@@ -488,12 +488,12 @@ func TestOrchCheckCapacity(t *testing.T) {
 
 func TestProcessPayment_GivenRecipientError_ReturnsNil(t *testing.T) {
 	n, _ := NewLivepeerNode(nil, "", nil)
+	n.Balances = NewBalances(5 * time.Second)
 	recipient := new(pm.MockRecipient)
 	n.Recipient = recipient
 	orch := NewOrchestrator(n)
 
 	recipient.On("ReceiveTicket", mock.Anything, mock.Anything, mock.Anything).Return("", false, nil)
-
 	err := orch.ProcessPayment(defaultPayment(t), ManifestID("some manifest"))
 
 	assert := assert.New(t)
@@ -537,6 +537,7 @@ func TestProcessPayment_GivenNilRecipient_ReturnsNilError(t *testing.T) {
 
 func TestProcessPayment_GivenLosingTicket_DoesNotRedeem(t *testing.T) {
 	n, _ := NewLivepeerNode(nil, "", nil)
+	n.Balances = NewBalances(5 * time.Second)
 	recipient := new(pm.MockRecipient)
 	n.Recipient = recipient
 	orch := NewOrchestrator(n)
@@ -552,12 +553,14 @@ func TestProcessPayment_GivenLosingTicket_DoesNotRedeem(t *testing.T) {
 
 func TestProcessPayment_GivenWinningTicket_RedeemError(t *testing.T) {
 	n, _ := NewLivepeerNode(nil, "", nil)
+	n.Balances = NewBalances(5 * time.Second)
 	recipient := new(pm.MockRecipient)
 	n.Recipient = recipient
 	orch := NewOrchestrator(n)
 	manifestID := ManifestID("some manifest")
 	sessionID := "some sessionID"
 	recipient.On("ReceiveTicket", mock.Anything, mock.Anything, mock.Anything).Return(sessionID, true, nil)
+
 	recipient.On("RedeemWinningTicket", mock.Anything, mock.Anything, mock.Anything).Return(errors.New("RedeemWinningTicket error"))
 
 	errorLogsBefore := glog.Stats.Error.Lines()
@@ -574,12 +577,14 @@ func TestProcessPayment_GivenWinningTicket_RedeemError(t *testing.T) {
 
 func TestProcessPayment_GivenWinningTicket_Redeems(t *testing.T) {
 	n, _ := NewLivepeerNode(nil, "", nil)
+	n.Balances = NewBalances(5 * time.Second)
 	recipient := new(pm.MockRecipient)
 	n.Recipient = recipient
 	orch := NewOrchestrator(n)
 	manifestID := ManifestID("some manifest")
 	sessionID := "some sessionID"
 	recipient.On("ReceiveTicket", mock.Anything, mock.Anything, mock.Anything).Return(sessionID, true, nil)
+
 	recipient.On("RedeemWinningTicket", mock.Anything, mock.Anything, mock.Anything).Return(nil)
 
 	errorLogsBefore := glog.Stats.Error.Lines()
@@ -596,6 +601,7 @@ func TestProcessPayment_GivenWinningTicket_Redeems(t *testing.T) {
 
 func TestProcessPayment_GivenMultipleWinningTickets_RedeemsAll(t *testing.T) {
 	n, _ := NewLivepeerNode(nil, "", nil)
+	n.Balances = NewBalances(5 * time.Second)
 	recipient := new(pm.MockRecipient)
 	n.Recipient = recipient
 	orch := NewOrchestrator(n)
@@ -624,6 +630,7 @@ func TestProcessPayment_GivenMultipleWinningTickets_RedeemsAll(t *testing.T) {
 
 func TestProcessPayment_GivenConcurrentWinningTickets_RedeemsAll(t *testing.T) {
 	n, _ := NewLivepeerNode(nil, "", nil)
+	n.Balances = NewBalances(5 * time.Second)
 	recipient := new(pm.MockRecipient)
 	n.Recipient = recipient
 	orch := NewOrchestrator(n)
@@ -664,6 +671,7 @@ func TestProcessPayment_GivenConcurrentWinningTickets_RedeemsAll(t *testing.T) {
 
 func TestProcessPayment_GivenReceiveTicketError_ReturnsError(t *testing.T) {
 	n, _ := NewLivepeerNode(nil, "", nil)
+	n.Balances = NewBalances(5 * time.Second)
 	recipient := new(pm.MockRecipient)
 	n.Recipient = recipient
 	orch := NewOrchestrator(n)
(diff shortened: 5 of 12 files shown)


core/accounting.go
```diff
@@ -0,0 +1,73 @@
...
+	if a.Balances[id].Amount == nil {
+		a.Balances[id] = balance{Amount: big.NewRat(0, 1)}
+	}
+	a.Balances[id] = balance{
+		LastUpdate: time.Now(),
+		Amount:     new(big.Rat).Add(a.Balances[id].Amount, amount),
+	}
+}
+
+// Debit substracts an mount from the balance for a ManifestID
```
kyriediculous: should be "an amount" , rather than "an mount" 🏇

kyriediculous: fixed in https://github.com/livepeer/go-livepeer/pull/956/commits/9a31f394cc058ad90f4f1566c6f5880490ef861d

core/accounting.go
```diff
@@ -0,0 +1,72 @@
...
+	"sync"
+	"time"
+)
+
+// Use transcodeLoopTimeout as TTL for a ManifestID's credit balance
+// We should prevent the use of global state if they are not constants
+
+// Accounts holds credit balances on a per-stream basis
+type Accounts struct {
+	Balances map[ManifestID]balance
```
yondonfu: Can we make these fields unexported to reduce the API surface area?

core/accounting.go
```diff
@@ -0,0 +1,72 @@
...
+	Mtx      sync.RWMutex
+}
+
+type balance struct {
+	LastUpdate time.Time // Unix time since last update
```
yondonfu: Can we make these fields unexported to reduce the API surface area?

core/accounting.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+// Credit adds an an amount to the balance for a ManifestID
+func (a *Accounts) Credit(id ManifestID, amount *big.Rat) {
+	a.Mtx.Lock()
+	defer a.Mtx.Unlock()
+	if a.Balances[id].Amount == nil {
+		a.Balances[id] = balance{Amount: big.NewRat(0, 1)}
+	}
+	a.Balances[id] = balance{
```
yondonfu: Since L27 will ensure that the map entry is initialized, we could also directly set the fields on the `balance` struct instead of creating a new `balance` struct for each update (applies for both this method and the `Debit()` method):

Ex.

```
a.Balances[id].LastUpdate = time.Now()
// Since we're updating the current value of `a.Balances[id].Amount` we can just mutate the existing
// `big.Rat` instead of allocating a new one
a.Balances[id].Amount.Add(a.Balances[id].Amount, amount)
```

core/accounting.go
```diff
@@ -0,0 +1,72 @@
...
+
+// Balance retrieves the current balance for a ManifestID
+func (a *Accounts) Balance(id ManifestID) *big.Rat {
+	a.Mtx.RLock()
+	defer a.Mtx.RUnlock()
+	return a.Balances[id].Amount
+}
+
+// Cleanup is a state flushing method to clean up the balances mapping
+func (a *Accounts) Cleanup() {
```
yondonfu: The caller of this method doesn't have a way to signal the loop to stop if the method is run within a goroutine. Generally, its a good idea to always know how goroutines will exit when starting them. One way that we currently do this in the existing codebase is to allow the caller to call a `Stop()` method on the struct that is managing the goroutine which will use a channel to signal to the goroutine to exit (this is what the [SenderMonitor](https://github.com/livepeer/go-livepeer/blob/master/pm/sendermonitor.go#L294) does).

core/livepeernode.go
```diff
@@ -78,14 +79,16 @@ type LivepeerNode struct {
...
+	lpNode := &LivepeerNode{
 		Eth:          e,
 		WorkDir:      wd,
 		Database:     dbh,
 		EthServices:  make(map[string]eth.EventService),
 		SegmentChans: make(map[ManifestID]SegmentChan),
 		segmentMutex: &sync.RWMutex{},
-	}, nil
+	}
+	lpNode.Accounts.Balances = make(map[ManifestID]balance)
```
yondonfu: Can we create a constructor for `Accounts` that handles initialization of any internal fields instead of directly initializing the fields of `Accounts` in this manner?

pm/recipient.go
```diff
@@ -43,6 +43,9 @@ type Recipient interface {
 
 	// TxCostMultiplier returns the multiplier -
 	TxCostMultiplier(sender ethcommon.Address) (*big.Rat, error)
+
+	// MinEV returns the recipients EV requirement as configured on startup
+	MinEV() *big.Int
```
yondonfu: Should this be called `EV()` instead? The EV configured on node start up is the fixed value required for tickets. It just happens to also be the value for O's minimum credit requirement.

Could we also return a `big.Rat` here (we'd have to convert the config value from a `big.Int`) so that callers don't need to worry about the conversion?

core/orchestrator.go
```diff
@@ -141,17 +141,28 @@ func (orch *orchestrator) ProcessPayment(payment net.Payment, manifestID Manifes
 			}
 		}
 
+		// Add ticket EV to credit
+		orch.node.Accounts.Credit(manifestID, new(big.Rat).Mul(new(big.Rat).SetFrac(ticket.FaceValue, big.NewInt(1)), ticket.WinProbRat()))
```
yondonfu: We can use [(*Ticket).EV()](https://github.com/livepeer/go-livepeer/blob/master/pm/ticket.go#L80) here.

core/orchestrator.go
```diff
@@ -141,17 +141,28 @@ func (orch *orchestrator) ProcessPayment(payment net.Payment, manifestID Manifes
...
+
 		if won {
 			glog.V(common.DEBUG).Infof("Received winning ticket manifestID=%v recipientRandHash=%x senderNonce=%v", manifestID, ticket.RecipientRandHash, ticket.SenderNonce)
 
 			go func(ticket *pm.Ticket, sig []byte, seed *big.Int) {
 				if err := orch.node.Recipient.RedeemWinningTicket(ticket, sig, seed); err != nil {
 					glog.Errorf("error redeeming ticket manifestID=%v recipientRandHash=%x senderNonce=%v: %v", manifestID, ticket.RecipientRandHash, ticket.SenderNonce, err)
 				}
+				// Update credit balance
```
yondonfu: Don't think we need to update the credit balance here

core/accounting.go
```diff
@@ -0,0 +1,72 @@
+package core
+
+import (
+	"math/big"
+	"sync"
+	"time"
+)
+
+// Use transcodeLoopTimeout as TTL for a ManifestID's credit balance
+// We should prevent the use of global state if they are not constants
```
yondonfu: What's the relevance of this line to the previous line? Or is this a general comment

core/accounting.go
```diff
@@ -0,0 +1,72 @@
+package core
+
+import (
+	"math/big"
+	"sync"
+	"time"
+)
+
+// Use transcodeLoopTimeout as TTL for a ManifestID's credit balance
```
yondonfu: I think its fine to use `transcodeLoopTimeout`, but curious what your motivation for using it was?

core/accounting.go
```diff
@@ -0,0 +1,72 @@
...
+// We should prevent the use of global state if they are not constants
+
+// Accounts holds credit balances on a per-stream basis
+type Accounts struct {
```
yondonfu: Should this be called `Balances`? IMO `Accounts` seems to imply that the struct is will hold ETH account specific data, but in reality the underlying map is just holding balances keyed by a manifestID.

core/orchestrator.go
```diff
@@ -141,17 +141,28 @@ func (orch *orchestrator) ProcessPayment(payment net.Payment, manifestID Manifes
...
 			}(ticket, tsp.Sig, seed)
 		}
 	}
 
+	// Check if there is enough credit for the manifestID to proceed
+	currBalance := orch.node.Accounts.Balance(manifestID)
+	reqBalance := new(big.Rat).SetFrac(orch.node.Recipient.MinEV(), big.NewInt(1))
+	if currBalance.Cmp(reqBalance) < 0 {
```
yondonfu: Add test that this error is returned properly?

core/orchestrator.go
yondonfu: Worth rebasing with master to pull in the changes from #937 - we should only credit the ticket EV to B's balance if there was either no error receiving the ticket or if the error receiving the error was acceptable (see this [TODO](https://github.com/livepeer/go-livepeer/pull/937/files#diff-b76fbd82bbf5b7a4d2fdf59ecd8d4ca6R147)).

yondonfu (changes requested):

core/orchestrator.go
kyriediculous: No indeed leftover comment from a stupid thought I had to add credit after ticket is redeemed which didnt' make sense anyway.

core/accounting.go
kyriediculous: good point, I'll look into it.

core/accounting.go
kyriediculous: I wanted to avoid having `Balances.balances` but we can, yes.

core/accounting.go
kyriediculous: Just a general comment, Can be removed. It's also not correct and should read "We should prevent the use of package level variables if they are not constants"

core/accounting.go
kyriediculous: Because it didn't make sense to me to keep track of a balance after the transcode loop times out. 
We're not bound to use this setting however and could do it another way if you prefer that.

yondonfu: > it didn't make sense to me to keep track of a balance after the transcode loop times out.

Gotcha. Just wanted to point out that even if we use `transcodeLoopTimeout` here, the clean up loop will not necessarily run at the same time as the timeout for a transcode loop. I think the just the name `transcodeLoopTimeout` here might be a bit misleading for some readers. What if we passed the value as a parameter similar to what is done for the [pm.SenderMonitor](https://github.com/livepeer/go-livepeer/blob/master/pm/sendermonitor.go#L94)? We could even use the same top level [value](https://github.com/livepeer/go-livepeer/blob/master/cmd/livepeer/livepeer.go#L57).

core/accounting.go
yondonfu: Or maybe `BalanceSheet`?

kyriediculous: I guess Balances is fine considering the underlying map will be (after fixup) unexported, so `Balances.balances` would only be internal actually internal to that file. 

What's offputting still is the actually exported getter `Balances.Balance(mid)` , or isn't it ?

core/accounting.go
kyriediculous: Crossed my mind to define it on the struct itself and after second thought that might be better. 

I'd like to use https://github.com/livepeer/go-livepeer/blob/a7ae0133c97ed05d77c130dcc9faf0b5392f7976/cmd/livepeer/livepeer.go#L55 though just because it's a `time.Duration` type as well. Not that it matters that much because we are converting it to int64 , but at least we're checking two `int64` casted `time.Duration `types that way.

core/accounting.go
kyriediculous: `cannot assign to struct field in map` compiler error, regardless whether it is initialized or not. 

Only way to circumvent is use a pointer as value to a key in the map, rather than a value itself it seems? I initially had it as a pointer, but it seemed excessive use of pointers. 

`balances map[ManifestID]*balance`

yondonfu: Ah right the map is tracking values instead of pointers right now. I think we should track pointers since we'll be doing frequent balance updates so we should avoid creating a struct each time and instead just do an in-place update on a pointer to an existing struct.

core/accounting.go
yondonfu: Either using `BalanceSheet` as the name of the struct or keeping the name `Balances` is fine by me so I'll leave that up to you

kyriediculous: I will go with Balances for brevity.

cmd/livepeer/livepeer.go
```diff
@@ -391,8 +391,11 @@ func main() {
 			n.Recipient.Start()
 			defer n.Recipient.Stop()
 
+			n.Balances = core.NewBalances(smCleanupInterval)
```
yondonfu: If we're going to be using the value of `smCleanupInterval` in multiple places, mind changing the variable name to make it clear that it isn't specifically for `pm.SenderMonitor`? Ex. `cleanupInterval`

pm/recipient.go
```diff
@@ -45,7 +45,7 @@ type Recipient interface {
 	TxCostMultiplier(sender ethcommon.Address) (*big.Rat, error)
 
 	// MinEV returns the recipients EV requirement as configured on startup
```
yondonfu: Comment needs to be updated to reflect new method name

pm/recipient.go
```diff
@@ -443,6 +443,6 @@ func (r *recipient) redeemManager() {
 	}
 }
```
yondonfu, line 445: Add method comment here?

core/accounting.go
```diff
@@ -6,67 +6,85 @@ import (
...
+	amount     *big.Rat  // Balance represented as a big.Rat
 }
 
-// Credit adds an an amount to the balance for a ManifestID
-func (a *Accounts) Credit(id ManifestID, amount *big.Rat) {
-	a.Mtx.Lock()
-	defer a.Mtx.Unlock()
-	if a.Balances[id].Amount == nil {
-		a.Balances[id] = balance{Amount: big.NewRat(0, 1)}
+// NewBalances returns a pointer to a Balance type with an initialized map
```
yondonfu: I think its clear from the function signature that we're returning a `*Balances` type so maybe we can just say something like: "NewBalances creates a Balances instance with the given ttl`

cmd/livepeer/livepeer.go
```diff
@@ -390,6 +390,12 @@ func main() {
 
 			n.Recipient.Start()
 			defer n.Recipient.Stop()
+
+			n.Balances = core.NewBalances(smCleanupInterval)
+			// Run n.Accounts.Cleanup() Routine
+			go n.Balances.Cleanup()
+			// Stop the cleanup routine on program exit
+			defer n.Balances.Stop()
```
yondonfu: Should be `StopCleanup()`

kyriediculous: Yup, This is addressed in the commit that includes the rebase
https://github.com/livepeer/go-livepeer/pull/956/commits/32c2ccb4c36e850fa79e065f2566f43b6721550b

core/accounting.go
kyriediculous: Adopted your comment in https://github.com/livepeer/go-livepeer/pull/956/commits/789195926a3d4628b3bc3e6acec20f4b78035e68

pm/recipient.go
kyriediculous: Added in https://github.com/livepeer/go-livepeer/pull/956/commits/789195926a3d4628b3bc3e6acec20f4b78035e68

pm/recipient.go
kyriediculous: Fixed in https://github.com/livepeer/go-livepeer/pull/956/commits/789195926a3d4628b3bc3e6acec20f4b78035e68

cmd/livepeer/livepeer.go
kyriediculous: fixed in https://github.com/livepeer/go-livepeer/pull/956/commits/789195926a3d4628b3bc3e6acec20f4b78035e68

core/orchestrator.go
kyriediculous: Added `TestProcesspayment_GivenInsufficientCredit_ReturnsErr` in https://github.com/livepeer/go-livepeer/pull/956/commits/789195926a3d4628b3bc3e6acec20f4b78035e68

core/orchestrator.go
kyriediculous: Fixed in https://github.com/livepeer/go-livepeer/pull/956/commits/32c2ccb4c36e850fa79e065f2566f43b6721550b

core/accounting_test.go
```diff
@@ -0,0 +1,92 @@
...
+	assert := assert.New(t)
+
+	// Set up two mids
+	// One we will update after 2*time.Seconds
+	// The other one we will not update before timeout
+	// This should run clean only the second
+	mid1 := ManifestID("First MID")
+	mid2 := ManifestID("Second MID")
+	// Start cleanup loop
+	go b.Cleanup()
```
yondonfu: `defer b.StopCleanup()` here?

core/accounting_test.go
```diff
@@ -0,0 +1,92 @@
...
+	credit := big.NewRat(100, 1)
+	b.Credit(mid1, credit)
+	b.Credit(mid2, credit)
+	assert.Zero(b.Balance(mid1).Cmp(credit))
+	assert.Zero(b.Balance(mid2).Cmp(credit))
+
+	time.Sleep(2 * time.Second)
+	b.Credit(mid1, credit)
+	assert.Zero(b.Balance(mid1).Cmp(big.NewRat(200, 1)))
+	// TODO: Assert LastUpdateTime
```
yondonfu: Were you planning on addressing this TODO? I think its fine not to check for lastUpdateTime equality as long as we're testing that the clean up loop works properly

core/accounting.go
```diff
@@ -0,0 +1,90 @@
...
+	defer b.mtx.RUnlock()
+	if b.balances[id] == nil {
+		return nil
+	}
+	return b.balances[id].amount
+}
+
+// Cleanup is a state flushing method to clean up the balances mapping
+func (b *Balances) Cleanup() {
+	clean := func() {
```
yondonfu: FWIW an easier way to test the clean up could be to define this function as an unexported method on `*Balances` which can be invoked within the tests. This way you could test the cleanup logic without depending on the ticker.

Ex.

```
func (b *Balances) cleanup() {
    ...
}

func (b *Balances) StartCleanup() {
    ...
    for {
        select {
            case <-ticker.C:
                b.cleanup()
        }
    }
}
```

core/orch_test.go
```diff
@@ -692,6 +710,49 @@ func TestProcessPayment_GivenReceiveTicketError_ReturnsError(t *testing.T) {
 	recipient.AssertNumberOfCalls(t, "RedeemWinningTicket", 2)
 }
 
+func TestProcesspayment_GivenInsufficientCredit_ReturnsErr(t *testing.T) {
```
yondonfu: "TestProcesspayment_" -> "TestProcessPayment"?

core/orchestrator.go
```diff
@@ -139,14 +139,11 @@ func (orch *orchestrator) ProcessPayment(payment net.Payment, manifestID Manifes
...
-		// TODO(yondonfu):
-		// Add ticket EV to the credit amount if any of the following conditions hold:
-		// - No error receiving the ticket
-		// - There is an error receiving the ticket, the type assertion on the error succeeds and the error is acceptable
-		// pmErr, ok := err.(pm.Error)
-		// if err == nil || (ok && pmErr.Acceptable()) {
-
-		// }
+		pmErr, ok := err.(pm.Error)
+		if err == nil || (ok && pmErr.Acceptable()) {
```
yondonfu, line 143: Could we add tests for the following cases:

- When there is an error, but it is acceptable we should still credit B's balance
- When there is an error, but it is not acceptable we should NOT credit B's balance

core/accounting_test.go
```diff
@@ -0,0 +1,92 @@
...
+
+import (
+	"math/big"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestEmptyb_ReturnsZeroedValues(t *testing.T) {
```
yondonfu: "Emptyb" -> "EmptyBalances"?

core/orch_test.go
```diff
@@ -502,10 +503,12 @@ func TestProcessPayment_GivenRecipientError_ReturnsNil(t *testing.T) {
 
 func TestProcessPayment_GivenNoSender_ReturnsError(t *testing.T) {
 	n, _ := NewLivepeerNode(nil, "", nil)
+	n.Balances = NewBalances(5 * time.Second)
 	recipient := new(pm.MockRecipient)
 	n.Recipient = recipient
 	orch := NewOrchestrator(n)
 	recipient.On("ReceiveTicket", mock.Anything, mock.Anything, mock.Anything).Return("some sessionID", false, nil)
+	recipient.On("EV").Return(big.NewRat(1000000000, 1))
```
yondonfu: Do we need this mock? `ProcessPayment()` should return before getting to the `EV()` call

core/orch_test.go
```diff
@@ -527,6 +530,7 @@ func TestProcessPayment_GivenNilNode_ReturnsNilError(t *testing.T) {
 
 func TestProcessPayment_GivenNilRecipient_ReturnsNilError(t *testing.T) {
 	n, _ := NewLivepeerNode(nil, "", nil)
+	n.Balances = NewBalances(5 * time.Second)
```
yondonfu: Do we to initialize this here? `ProcessPayment()` won't access this before returning here

core/orch_test.go
kyriediculous: Correct, `n.Balances = NewBalances(5 * time.Second)` can be removed as well.

core/accounting_test.go
kyriediculous: I was contemplating it for completeness but I can't do an EXACT test for `time.Now()` I think, so it's probably not as valuable to test it as long as the cleanup loop works as you say.

kyriediculous: Removed comment in https://github.com/livepeer/go-livepeer/pull/956/commits/794632acb80531d8222ca2a12334d3ddcd321427

core/accounting_test.go
kyriediculous: Fixed in https://github.com/livepeer/go-livepeer/pull/956/commits/794632acb80531d8222ca2a12334d3ddcd321427

core/accounting_test.go
kyriediculous: fixed in https://github.com/livepeer/go-livepeer/pull/956/commits/794632acb80531d8222ca2a12334d3ddcd321427

core/orch_test.go
kyriediculous: fixed in https://github.com/livepeer/go-livepeer/pull/956/commits/794632acb80531d8222ca2a12334d3ddcd321427

core/orch_test.go
kyriediculous: fixed in https://github.com/livepeer/go-livepeer/pull/956/commits/794632acb80531d8222ca2a12334d3ddcd321427

core/accounting.go
kyriediculous: fixed in https://github.com/livepeer/go-livepeer/pull/956/commits/794632acb80531d8222ca2a12334d3ddcd321427 
removed blank lines in https://github.com/livepeer/go-livepeer/pull/956/commits/e5fe59bb76eee1d407644a3c936cc0dc6329e497

core/accounting.go
```diff
@@ -60,23 +60,24 @@ func (b *Balances) Balance(id ManifestID) *big.Rat {
...
+	for id, balance := range b.balances {
+		b.mtx.Lock()
+		if int64(time.Since(balance.lastUpdate)) > int64(b.ttl) {
+			delete(b.balances, id)
 		}
+		b.mtx.Unlock()
 	}
+}
+
+// Cleanup is a state flushing method to clean up the balances mapping
```
yondonfu: Need to update method name in comment

core/orch_test.go
yondonfu: Looks like the `n.Balances` initialization on L506 still needs to be removed?

core/orch_test.go
```diff
@@ -708,27 +700,88 @@ func TestProcessPayment_GivenReceiveTicketError_ReturnsError(t *testing.T) {
 	recipient.AssertNumberOfCalls(t, "RedeemWinningTicket", 2)
 }
 
-func TestProcessPayment_GivenInsufficientCredit_ReturnsErr(t *testing.T) {
+// Check that an Acceptable error increases the credit
+
```
yondonfu: Kill space?

core/orch_test.go
```diff
@@ -708,27 +700,88 @@ func TestProcessPayment_GivenReceiveTicketError_ReturnsError(t *testing.T) {
...
 	manifestID := ManifestID("some manifest")
-	sessionID := "some sessionID"
-	recipient.On("ReceiveTicket", mock.Anything, mock.Anything, mock.Anything).Return(sessionID, true, nil)
-	recipient.On("EV").Return(big.NewRat(1000000000, 1))
+	acceptableError := pm.NewMockReceiveError(errors.New("Acceptable ReceiveTicket error"), true)
+	recipient.On("ReceiveTicket", mock.Anything, mock.Anything, mock.Anything).Return("", false, acceptableError).Once()
+	assert := assert.New(t)
+	// Create a ticket where faceValue = EV instead of using 'defaultTicket'
+	// FaceValue is the expected balance increase
+	faceVal := big.NewInt(100)
```
yondonfu: Why do we need to do all this boilerplate setup for generating a payment? Couldn't we just use `defaultPayment()` and then calculate the EV based on the ticket params included in the returned payment? These questions apply to the other tests in this commit that do a similar setup for generating a payment

core/orch_test.go
```diff
@@ -708,27 +700,88 @@ func TestProcessPayment_GivenReceiveTicketError_ReturnsError(t *testing.T) {
...
+
+// Check that an unacceptable error does NOT increase the credit
+func TestProcessPayment_UnacceptablePaymentError_DoesNotIncreaseCreditBalance(t *testing.T) {
+	n, _ := NewLivepeerNode(nil, "", nil)
+	n.Balances = NewBalances(5 * time.Second)
+	recipient := new(pm.MockRecipient)
+	n.Recipient = recipient
+	orch := NewOrchestrator(n)
+	manifestID := ManifestID("some manifest")
+	UnacceptableError := pm.NewMockReceiveError(errors.New("Unacceptable ReceiveTicket error"), false)
```
yondonfu: "UnacceptableError" -> "unacceptableError"

core/orch_test.go
```diff
@@ -708,27 +700,88 @@ func TestProcessPayment_GivenReceiveTicketError_ReturnsError(t *testing.T) {
...
+	recipient.On("ReceiveTicket", mock.Anything, mock.Anything, mock.Anything).Return("", false, UnacceptableError).Once()
+	assert := assert.New(t)
+
+	err := orch.ProcessPayment(defaultPayment(t), manifestID)
+	assert.Error(err)
+	assert.Nil(orch.node.Balances.Balance(manifestID))
+}
 
+func TestSufficientPayment_IsSufficient_ReturnsTrue(t *testing.T) {
```
yondonfu: I think the test prefix should be "TestSufficienceBalance" to match the name of the method being tested

core/orchestrator.go
```diff
@@ -156,17 +156,11 @@ func (orch *orchestrator) ProcessPayment(payment net.Payment, manifestID Manifes
 		}
 	}
 
+	// Return errors if there were acceptable payment errors but credit is sufficient
```
yondonfu: I think this comment is a bit inaccurate because `ProcessPayment()` does not know anything about the credit balance being sufficient/insufficient - it will return an error as long as it encountered *any* error (regardless of whether it was acceptable or not) when invoking `ReceiveTicket()`

yondonfu (commented):
Looking good! Left a few additional small comments

core/orch_test.go
kyriediculous: You're saying calling the mock for "EV" with the faceValue/winProb that is generated by `defaultPayment`?

Could do that, yes. I didn't know it it would be robust enough since these fields are randomly generated byte slices by using `defaultpayment` and I'd rather rely on hardcoded values than calculated values. 

What do you think about generating a payment with defaultPayment but explicitly setting faceVal and winProb after generation? This way we can remove the boilerplate but still have fixed values to expect. It seems to be the best of both worlds? See https://github.com/livepeer/go-livepeer/pull/956/commits/f7b0485b60d135b363e3b5d78b6b0b7dadf1e4e9 , WDYT?

```
	// Create a ticket where faceVal = EV
	// making faceVal the expected balance increase
	faceVal := big.NewInt(100)
	payment := defaultPayment(t)
	payment.TicketParams.FaceValue = faceVal.Bytes()
	payment.TicketParams.WinProb = new(big.Int).Sub(new(big.Int).Lsh(big.NewInt(1), 256), big.NewInt(1)).Bytes()
```

core/orchestrator.go
kyriediculous: Leftover note/comment from the discussion we had that I forgot to remove again. I looked over this when doing a final checkup of working trees (initially the logic to check balances was here).

Fixed in https://github.com/livepeer/go-livepeer/pull/956/commits/f7b0485b60d135b363e3b5d78b6b0b7dadf1e4e9

core/orch_test.go
kyriediculous: Fixed in https://github.com/livepeer/go-livepeer/pull/956/commits/f7b0485b60d135b363e3b5d78b6b0b7dadf1e4e9

core/orch_test.go
kyriediculous: Fixed in https://github.com/livepeer/go-livepeer/pull/956/commits/f7b0485b60d135b363e3b5d78b6b0b7dadf1e4e9

core/orch_test.go
kyriediculous: Fixed in https://github.com/livepeer/go-livepeer/pull/956/commits/f7b0485b60d135b363e3b5d78b6b0b7dadf1e4e9

core/orchestrator.go
yondonfu: Looks like this comment is still here?

kyriediculous: Weird.. Perhaps pressed the wrong key combo again to save before committing. Should be fixed now.

yondonfu:
Needs a rebase (squash fixups and resolve conflicts)!

core/orchestrator.go
```diff
@@ -199,6 +196,15 @@ func (orch *orchestrator) PriceInfo(sender ethcommon.Address) (*big.Rat, error)
 	return new(big.Rat).Mul(orch.node.PriceInfo, overhead), nil
 }
 
+// SufficientBalance checks whether the credit balance for a stream is sufficient
+// to proceed with downloading and transcoding
+func (orch *orchestrator) SufficientBalance(manifestID ManifestID) bool {
```
yondonfu, line 201: Just realized I missed this in the original code review - currently, this method will break off-chain mode because `orch.node.Balances` will be be nil in off-chain mode. We should add a guard clause in this method that returns `true` if `orch.node`, `orch.node.Balances` or `orch.node.Recipient` are nil similar to the guard clause in [ProcessPayment()](https://github.com/livepeer/go-livepeer/blob/master/core/orchestrator.go#L99).

yondonfu (changes requested):

core/orchestrator.go
kyriediculous: You're right, I noticed this when firing up nodes, but forgot this over the weekend.

kyriediculous: Should be dealt with in https://github.com/livepeer/go-livepeer/pull/956/commits/59957ab90df4921cb86674348b1c78374098c0b0

core/orch_test.go
```diff
@@ -785,6 +785,16 @@ func TestSufficientBalance_IsNotSufficient_ReturnsFalse(t *testing.T) {
 	assert.False(orch.SufficientBalance(manifestID))
 }
 
+func TestSufficientBalance_OffChainMode_ReturnsTrue(t *testing.T) {
+	n, _ := NewLivepeerNode(nil, "", nil)
+	manifestID := ManifestID("some manifest")
+	orch := NewOrchestrator(n)
+	assert.True(t, orch.SufficientBalance(manifestID))
+
+	orch.node = nil
```
yondonfu, line 801: Can we add tests for the other cases included in the guard clause as well?

kyriediculous: orch.node.Recipients and orch.node.Balances should be nil in the first assert check, the wrapping node should be nil in the second assert check. You want to explicitly set a recipient with Balances nil and vice versa as well?

yondonfu: Yeah I think it would be good to include to cover all the cases introduced by the guard clause which would include:

- `orch.node != nil && orch.node.Recipient = nil`
- `orch.node != nil && orch.node.Recipient != nil && orch.node.Balances = nil`

kyriediculous: Fair enough, added the test cases.

yondonfu (commented):
Needs another rebase!

kyriediculous:
> Needs another rebase!

Done !

yondonfu (approved):
LGTM :shipit:

PR #958, opened 2019-07-03
net, server: include expected price in payment message

### LINKED ISSUE #954: Include pricePerPixel in Payment message

Currently, B expects to pay O's last known price when it submits a segment - B expects the fee debited from its balance for a transcoded result to be the # of pixels processed * O's last known price. However, this expectation can be left unfulfilled if O's `pricePerPixel` changes from the time that B receives the price to the time that O computes the fee to debit from B's balance. B should be able to indicate the `pricePerPixel` it expects to pay and if O chooses to transcode the segment then the fee debited from B's balance should be based on the `pricePerPixel` provided by B.

B can include O's last known price with the `Payment` message passed over the wire.

We'll need to:

- Update the [Payment](https://github.com/livepeer/go-livepeer/blob/master/net/lp_rpc.proto#L245) RPC message, which is passed over the wire, to include a nested `PriceInfo` message
- Update the [genPayment()](https://github.com/livepeer/go-livepeer/blob/master/server/segment_rpc.go#L452) to include O's last known price in the `Payment` message to be serialized and passed over the wire

Changes to [(*orchestrator).ProcessPayment()](https://github.com/livepeer/go-livepeer/blob/master/core/orchestrator.go#L98) (i.e. actually validating the provided price) can be implemented separately.

B could skip providing any indication of the `pricePerPixel` that it expects to pay and O could just transcode a segment based on whatever its current `pricePerPixel` is. Then, when O returns a transcode receipt, B could use O's last known `pricePerPixel` to verify the balance. In this case, O's `pricePerPixel` could have legitimately changed resulting in the balance in the receipt appearing to be incorrect. As a result, B disconnects from O because it had no way to receive notifications of an updated `pricePerPixel` prior to receiving the transcoded result. Alternatively, B could derive the `pricePerPixel` used by computing the difference between its own view of its balance and the balance in the receipt - if the `pricePerPixel` is still below B's `maxPricePerPixel`, B would consider the balance in the receipt valid. However, the problem here is that if O knows B's `maxPricePerPixel`, it could advertise low `pricePerPixel` values, but when returning receipts always use B's `maxPricePerPixel - 1`.

Another design involves O including the `pricePerPixel` used in the receipt, but I think the implications would be the same as ones for the above design.

The proposed solution would allow B expect notifications for `pricePerPixel` updates earlier in the workflow allowing B to just disconnect if a transcode receipt includes a balance that does not reflect the `pricePerPixel` provided with the segment.

**See below for another approach for B to provide a pricePerPixel with segments**

B already includes segment metadata in the header of a segment so we can just have B add O's known price to this metadata payload.

We'll need to:

- Update [core.SegTranscodingMetadata](https://github.com/livepeer/go-livepeer/blob/master/core/streamdata.go#L25), which is flattened and signed by B, to include O's last known price represented by a `*big.Rat`
- Update the RPC message [SegData](https://github.com/livepeer/go-livepeer/blob/master/net/lp_rpc.proto#L105), which is passed over the wire, to include a nested `PriceInfo` message
- Update [genSegCreds()](https://github.com/livepeer/go-livepeer/blob/master/server/segment_rpc.go#L414) to include O's last known price in B's signature and to include O's last known price in the `SegData` message to be serialized and passed over the wire

Changes to [verifySegCreds()](https://github.com/livepeer/go-livepeer/blob/master/server/segment_rpc.go#L209) (i.e. actually validating the provided price) can be implemented separately.

**Additional context**

kyriediculous:
B can include O's last known price with the `Payment` message passed over the wire. This is the price that B is expecting to pay for the transcoding job. 

- Added `PriceInfo `to `Payment` message (https://github.com/livepeer/go-livepeer/pull/958/commits/bd4e8208c0a4c89c67a54c7cbf4fd6616f689064) 
- Updated `genPayment` in `server/segment_rpc.go` to include O's last known price in the serialized `Payment `message returned from this function. (https://github.com/livepeer/go-livepeer/pull/958/commits/bd4e8208c0a4c89c67a54c7cbf4fd6616f689064)
- Added comments for the added fields in the proto definitions (https://github.com/livepeer/go-livepeer/pull/958/commits/7ce34f50c590cdbdf516a060134ea60910bd08c8)

- Added an assertion and mock `OrchestratorInfo` in `TestGenpayment`
- Updated `defaultPayment()` test helper to include an `OrchestratorInfo` field
- Added assertions in relevant `GetPayment` tests that make use of `defaultPayment()` to generate a `Payment` message

Fixes #954 

**Checklist:**


server/rpc_test.go
```diff
@@ -561,11 +572,17 @@ func defaultPayment(t *testing.T) *net.Payment {
 
 func defaultPaymentWithTickets(t *testing.T, senderParams []*net.TicketSenderParams) *net.Payment {
 	sender := pm.RandBytes(123)
-
+	oinfo := &net.OrchestratorInfo{
```
yondonfu: Doesn't look like we need to create `*net.OrchestratorInfo` here since we just need to provide `*net.PriceInfo` for the `ExpectedPrice` field when creating `*net.Payment`.

yondonfu (changes requested):
Looking good, just one tiny comment

server/rpc_test.go
kyriediculous: Ah yes true, I was being lazy and copy pasted that from the `genPayment` test.

Fixed in https://github.com/livepeer/go-livepeer/pull/958/commits/ee0955b26e5234bee08257d1a055f607be90a51a

yondonfu (approved):
LGTM :shipit:

PR #962, opened 2019-07-04
Improve caching of segment's data in memory object storage.

darkdarkdragon:
More efficient cache in local (memory) object storage.
Previously for removing item from cache was used slicing operation like this:
```golang
		dc.cache = dc.cache[1:]
```
This doesn't free memory immediately - in my tests, underlying array can grow twice the size before it get garbage collected.

Instead of slicing/appending array using fixed size array.

Manually. In my tests this change increased number of simultaneous stream single broadcaster can handle (on `n1-highcpu-4` instance (3.6G memory)) from 16 to 24.

**Checklist:**

### DIFF
--- a/drivers/local.go
+++ b/drivers/local.go
@@ -132,7 +132,8 @@ func (ostore *MemorySession) getAbsoluteURI(name string) string {
 
 type dataCache struct {
 	cacheLen int
-	cache    []*dataCacheItem
+	nextFree int
+	cache    []dataCacheItem
 }
 
 type dataCacheItem struct {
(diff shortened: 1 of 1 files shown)


j0sh (approved):
LGTM! Curious, is RAM still the bottleneck after this fix?

j0sh:
Also - the OS stuff is sorely lacking in unit tests. If we could start introducing some basic tests here (eg, for inserts and rollovers) then that would be really helpful.

darkdarkdragon:
@j0sh 
Yes, at least for small instance (get's killed by OOM if more then 24 streams on 3.6G machine) - roughly it needs 150Mb per stream.
Will test bigger instance over weekend (was seeing ` Possible SYN flooding on port 8935. Sending cookies.` on `n1-highcpu-32` 28.8G mem instance with 160 streams).

darkdarkdragon:
@j0sh Added tests - #964. Also caught bug in previous fix...

PR #964, opened 2019-07-06
Fix for memory OS (was not replacing properly item with the same name)

darkdarkdragon:
Structure in array can't be modified - whole structure needs to be replaced.

Added fix

Unit test added.

**Checklist:**

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -12,6 +12,11 @@ go test -logtostderr=true
 t2=$?
 cd ..
 
+cd drivers
+go test -logtostderr=true
+t2=$?
+cd ..
+
 cd monitor
 go test -logtostderr=true
 t3=$?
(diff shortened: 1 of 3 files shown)


mk-livepeer (approved):
approved 👍

test.sh
```diff
@@ -12,6 +12,11 @@ go test -logtostderr=true
 t2=$?
 cd ..
 
+cd drivers
+go test -logtostderr=true
+t2=$?
```
j0sh, line 17: This reuses `t2` ; it shouldn't. If the previous assignment fails and this one succeeds, then CI will consider the tests to have passed when that is clearly not the case. Needs to be fixed.

darkdarkdragon:
@j0sh Made fix: #968

PR #967, opened 2019-07-08
WIP Y/lp joy4 fix

ya7ya:
upgrades lpms to the latest version which includes new `joy4` links

- removed old `joy4` dependencies
- added new `joy4` livepeer fork.

**Checklist:**


j0sh:
Noticed that this replaces the vendored dir with a Git submodule. Are submodules the typical way for `go mod` to handle Git dependencies?

My experience is that Git submodules have poor UX. Getting the goclient built is hard enough; my feeling is that submodules would add yet more complexity to process. This may be OK if that's something that `go mod` handles for us transparently. Curious what others think.

ya7ya:
@j0sh you're right, this isn't the standard way of handling dependencies, I was just testing the fixes and accidently opened the PR tbh 😅 .

I do think we should use `go mod` to handle dependencies.

PR #968, opened 2019-07-08
Fix tests

darkdarkdragon:

- 
- 
- 

**Checklist:**

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -12,11 +12,6 @@ go test -logtostderr=true
 t2=$?
 cd ..
 
-cd drivers
-go test -logtostderr=true
-t2=$?
-cd ..
-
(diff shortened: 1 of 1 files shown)


j0sh (approved):
LGTM.

I do wonder if we can simplify this some day. Eg, the `set -e` Bash [builtin](https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html) is really nice, but wouldn't allow other tests to continue running if an earlier test fails (if we want to maintain that property at all).

darkdarkdragon:
@j0sh Why we're not doing `go test ./...` ?

j0sh:
Didn't know about that feature - nice!

Issue #953, opened 2019-07-01
Implement HTTP POST/PUT ingest of video segments


iameli:
This is designed to be a simple API for backing external API services such as the Wowza plugin.

### Request: `POST /:streamKey/:segment.ts`
Required Headers: 
* `Content-Type: video/MP2T`
* `Accept: multipart/form-data`

Example: `POST /stream/eli/1.ts`

This posts in an MPEG-TS video segment to a go-livepeer broadcaster. The URL structure uses the same patterns we use for RTMP URLs — the "stream key" would be the same and the webhook can be triggered in the same way to access the bitrate ladder and whatnot.

Behind the scenes, go-livepeer should begin its main transcoding loop. It should send the segment off to an orchestrator and be ready to await new sequential video requests following the same URL pattern.

### Response:

Server sends a `200 OK` response with a multipart/form-data blob containing all of the transcoded outputs.

mkrufky:
I was under an impression that the `m3u8` playlist (or a url pointing to one) is what we'd POST to the endpoint.  This would be enough info for the process to pull down segments on its own.

Is POST'ing segments directly to the endpoint a requirement?  If so, can you explain why?  Might it make sense to support both playlist POST and ts direct POST?

iameli:
Interesting — that could certainly work also, direct POST isn't a requirement. I might upload the segments to GCP buckets or something to allow us to handle the case where the Wowza server isn't on the public internet but that's workable. I suppose I'd POST the manifest every time a new segment is available? The only downside off the top of my head would be that in a failover/deployment case we'd do a bit of unnecessary transcoding:

Wowza:
```
POST /stream/eli.m3u8

https://example.com/1.ts
https://example.com/2.ts
https://example.com/3.ts
```

Livepeer Broadcaster A: Okay, I'll transcode 1.ts, 2.ts, and 3.ts

Wowza:
```
POST /stream/eli.m3u8

https://example.com/1.ts
https://example.com/2.ts
https://example.com/3.ts
https://example.com/4.ts
```

Livepeer Broadcaster A: Cool, already did most of those, but I'll transcode 4.ts.

Livepeer Broadcaster A: Oh no, I'm dead! 💀

Livepeer Broadcaster B: I'll take over.

Wowza:
```
POST /stream/eli.m3u8

https://example.com/2.ts
https://example.com/3.ts
https://example.com/4.ts
https://example.com/5.ts
```

[...]

Issue #938, opened 2019-06-25
Latency Based O Selection


j0sh:
**Is your feature request related to a problem? Please describe.**

The broadcaster's orchestrator selection algorithm currently works by popping [O off a stack](https://github.com/livepeer/go-livepeer/blob/master/server/broadcast.go#L52-L65). The [attached diagram](https://sequencediagram.org/index.html#initialData=C4S2BsFMAICECcD2BDAJgY2QZ2Je0B5edAC2gGVIp1REA7AKAYAdl5R0RW7hoAiBCgzZc8PtGzQsWJgwBGiAB7QoAM16JVUrAC5oAKn2VpIehWDJ0Aa0MAdOomIkAgvcelYsrK3QwAtADMTNJ+AHxEpM56lADmAIzySiqQ6tCa2noRJJ4M3pb+QblYYVmw0ZAx0ABMTArKahpa0jrBPgXBWNAAPH5+hE6w0OWVVdAASpDe9FiQtUkNaU26-R6t+dCBTFnOG2HaQxQV0HHjk8zTs3XJqenNKy72WXC5bRtB0huh9zvD0EFXC1uy1KL3WfgALEA) shows how this is suboptimal: here, orchB has a faster response time, but orchA is selected for the next segment anyway due to coming in later and being on top of the stack.

**Describe the solution you'd like**

We can fine tune this to select O based on a latency score. As a first approximation, the latency score for a given segment can be calculated as the transcode duration over the segment duration. We might also want to apply some form of hysteresis on this score over the life of the stream (eg, EWMA) but the appropriate parameters should be modeled first. Implementation can consist of replacing the O stack in the session manager with a heap.

Initial latency values for new Os should be 1. Os that tend to process faster than realtime will have scores <0 , while those slower than realtime will have scores > 1. This allows us to cycle through a set of orchestrators that may not be able to process segments faster than realtime, and settle on the top fastest few.

Selecting based on latency also gives orchestrators more incentive to complete the work quickly. Slow or over-subscribed Os may not receive additional segments beyond the first one, regardless of their initial place in the selection stack.

![image](https://user-images.githubusercontent.com/292510/60117565-21edcb00-972f-11e9-8526-c51c9bb7a470.png)
[...]

Issue #846, opened 2019-05-02
Do not save source segments to disk


darkdarkdragon:
Right now orchestrator always [saves](https://github.com/livepeer/go-livepeer/blob/master/core/orchestrator.go#L280) source segment to disk.
But in case remote transcoders used, this saved files isn't used, and just [deleted](https://github.com/livepeer/go-livepeer/blob/master/core/orchestrator.go#L353) moment after.

j0sh:
Yes, there are a number of leftovers here from the old protocol that ought to be cleaned up. In particular, I don't think that saving to disk is necessary anymore. We used to save to disk for claiming / verification (and since we were doing that, it was a small optimization to just reuse the disk store for local transcoding), but that flow isn't here anymore.

Here, we can probably just save directly to local OS if we aren't using external OS.

(I'm really curious whether the localhost network overhead is less than the disk IO, but I don't think it's critical to answer or optimize at this moment. And I have a feeling we might end up with a disk-backed OS at some point soon.)

j0sh:
On my local machine, some perhaps unscientific testing [1] shows that the network-only approach is about 20% slower [2]. Will have to benchmark a bit in different environments where we may be more IOPS constrained or have faster CPUs.

[1]
* Start a B / O both on the local machine. 2014-era Macbook Air, with SSD and a 1.7GHz i7.
* Run a software transcode session for BBB with a single O with/without the following patch. Let it process at least 51 segments.
* Extract timings: `head -51 orchestrator.log | grep "Transcoding of" orchestrator.log | grep -Eo '[0-9]+\.[0-9]+m?s' | grep -Eo '[0-9]+\.[0-9]+'` | pbcopy
* Copy results into the linked spreadsheet.
* Manually fix up any rows that are single (or small)-digit decimals by moving the decimal place over three places (converting seconds to milliseconds)

[2]
Results: https://docs.google.com/spreadsheets/d/1L0kxi3Hy9pTaVRrGOLZKkDEcTSVH7Utu7gq484NLK0w/edit#gid=28917058

FFmpeg Input Config:
```
ffmpeg -re -i big_buck_bunny_480p_h264.mov -c:a copy -c:v copy -f flv rtmp://localhost
```

Broadcaster Config:
```
      /Users/josh/Livepeer/node/src/github.com/livepeer/go-livepeer/livepeer \
              -datadir /Users/josh/.lpData/broadcaster \
              -orchAddr 127.0.0.1:8936 \
[...]

PR #972, opened 2019-07-09
HTTP Segment Ingest

j0sh:
### What is it?

A sketch of a push-based HTTP segment ingest. Refer to #953 for details.

This doesn't attempt to address the push vs pull debate. This is only a sketch of one possible implementation of HTTP segment push.

Not really meant for review / merging, but more as an overview of what might need to be addressed if we take this approach all the way to production.

### Usage

Check out the [`ja/liveingest`](https://github.com/livepeer/go-livepeer/compare/33fa9b2...ja/liveingest?expand=1) branch of go-livepeer

Input : `ffmpeg -re -i <input> -c:a copy -c:v copy -f hls http://localhost:7935/live/<streamname>/` Wowza might work as well.

Output: Same as existing HLS playback. `ffplay http://localhost:8935/stream/<streamname>.m3u8` 

### Path Prefix

Currently `/live/` , as opposed to the proposed `/stream/`.  LPMS hijacks the `/stream/` prefix so it's easier to just use a new prefix. Transcoded HLS is still available for playback under `/stream/` as before.

### Port numbering

Currently use port 7935, mostly out of expediency - the B doesn't pass in a custom mux that can be shared between LPMS and the goclient.

We have an internal CLI  API running on port 7935 that is mostly used for node management and diagnostics. An "external" port lives on 8935. The broadcaster uses the external port for HLS playback.  (O uses 8935 for gRPC and segment handling as well).

We may want to switch the `/live/` endpoint to port 8935. One bit of concern here is relating to "safe defaults". Historically, we have attempted to keep all potentially Eth-draining operations segregated from the port on 8935, which is considered public. Since there is no authentication by default, a B operator may unintentionally open itself up to drive-by transcoding.

One mitigating factor is that the broadcaster listens to 8935 on localhost by default. (We do the same for RTMP.) Another layer of safety would be to have a flag to toggle the push API. For an improved UX, this flag could be enabled by default. It would be disabled by default if 8935 is listening to non-localhost *and* webhook auth is not enabled. Enabling the flag under those conditions would be an explicit signal - "I know this is potentially an expensive choice, do it anyway."

### Durations

The duration of the source segment is currently not passed in. The duration is necessary to insert valid timings for the resulting HLS playlist. Other features such as [latency-based O selection](https://github.com/livepeer/go-livepeer/issues/938) need the duration.

To solve this, it would be nice to send a HTTP header, eg `Content-Duration` with the value in seconds. We could also probe each segment but that's not desirable on the broadcaster: it's format-specific and another step in the process.

### Other Missing Metadata

Might be nice to incorporate the width/height of the source segment as a header. We use this information in the master playlist to describe the dimensions of the source stream.

### State Timeouts

Need to monitor liveness of a given broadcast. There is a bunch of state associated with each job such as the working set of orchestrators, payments, etc. This state needs to be cleaned up after a certain period of idleness.

One way to do this is to look a the O's [transcode loop](https://github.com/livepeer/go-livepeer/blob/3b6a1918a646383f4413b8a3a8f3469d9fa27818/core/orchestrator.go#L456-L481). Uses a context timeout to clean itself up. Nice because it serializes work for a given job onto the same thread. Not so nice because the async dispatching of request/response makes the overall flow hard to follow. We don't really need the serialization either.

Another way is to have a separate goroutine sweep live sessions periodically. Even a naive linear scan should be OK with a few hundred streams.

### Rate Limiting

If auth fails or there is an unrecoverable error, we should stop processing requests for that stream for a while (at least until the state expires).

### Reconstruct Target URL

The golang `request.URL` field is simply extracted from the Request-URI part of the HTTP request, and is usually missing the scheme, host, port, etc. Having the full target URL would be useful for the auth webhook. The URL needs to be reconstructed from the data available in the [Request](https://golang.org/pkg/net/http/#Request) object.

### Responses

Currently, nothing is returned in response. A [`multipart/mixed` implementation](https://github.com/livepeer/go-livepeer/blob/3b6a1918a646383f4413b8a3a8f3469d9fa27818/server/ot_rpc.go#L145-L160) can be borrowed from the standalone transcoder.

### Non-Segmented Transcodes

There is an easy path towards standalone (non-segmented) transcoding. This would enable a semblance of JIT VOD.

If there is no stream name (eg, the ingest path is `http://localhost:7935/live`) then we can create a standalone "rtmp connection" (without registering), use that for the duration of the transcode, and [clean it up](https://github.com/livepeer/go-livepeer/blob/3b6a1918a646383f4413b8a3a8f3469d9fa27818/server/mediaserver.go#L325-L326) after `processSegment` completes.

### RTMP Segmentation

This type of push ingest can also be used to receive segments via HTTP as they are produced from a RTMP stream [2]. This would remove a lot of complexity from the LPMS side: it would avoid disk writes [1], polling the disk for segments and manifests, and so forth. This would also give us a mechanism to accommodate RTMP reconnects for the same broadcast session, rather than starting fresh as we do now.

However, the RTMP workflow is still just different enough that this should be looked at a bit more closely. Some issues:

* We authenticate the RTMP stream *prior* to starting segmentation. With push, we authenticate *after* segmentation.
* A naive application of may unintentionally block the segmenter. ffmpeg will shut down the segmeter if there is a backlog in HTTP.

[1] https://github.com/livepeer/go-livepeer/issues/846
[2] https://github.com/livepeer/lpms/issues/128#issuecomment-506794144

### DIFF
--- a/server/mediaserver.go
+++ b/server/mediaserver.go
@@ -17,6 +17,7 @@ import (
 	"path"
 	"regexp"
 	"runtime"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
@@ -53,15 +54,17 @@ const StreamKeyBytes = 6
 
 const SegLen = 2 * time.Second
 const BroadcastRetry = 15 * time.Second
+const RefreshIntervalHttpPush = 1 * time.Minute
 
 var BroadcastJobVideoProfiles = []ffmpeg.VideoProfile{ffmpeg.P240p30fps4x3, ffmpeg.P360p30fps16x9}
 
 var AuthWebhookURL string
 
 type streamParameters struct {
-	mid      core.ManifestID
-	rtmpKey  string
-	profiles []ffmpeg.VideoProfile
+	mid        core.ManifestID
+	rtmpKey    string
+	profiles   []ffmpeg.VideoProfile
+	resolution string
 }
 
 func (s *streamParameters) StreamID() string {
@@ -76,14 +79,14 @@ type rtmpConnection struct {
 	profile     *ffmpeg.VideoProfile
 	params      *streamParameters
 	sessManager *BroadcastSessionsManager
+	lastUsed    time.Time
 }
 
 type LivepeerServer struct {
-	RTMPSegmenter lpmscore.RTMPSegmenter
-	LPMS          *lpmscore.LPMS
-	LivepeerNode  *core.LivepeerNode
-	HTTPMux       *http.ServeMux
-
+	RTMPSegmenter         lpmscore.RTMPSegmenter
+	LPMS                  *lpmscore.LPMS
+	LivepeerNode          *core.LivepeerNode
+	HTTPMux               *http.ServeMux
 	ExposeCurrentManifest bool
 
 	// Thread sensitive fields. All accesses to the
@@ -100,24 +103,29 @@ type authWebhookResponse struct {
 	Presets    []string `json:"presets"`
 }
 
-func NewLivepeerServer(rtmpAddr string, httpAddr string, lpNode *core.LivepeerNode) *LivepeerServer {
+func NewLivepeerServer(rtmpAddr string, lpNode *core.LivepeerNode) *LivepeerServer {
 	opts := lpmscore.LPMSOpts{
-		RtmpAddr: rtmpAddr, RtmpDisabled: true,
-		HttpAddr: httpAddr,
-		WorkDir:  lpNode.WorkDir,
+		RtmpAddr:     rtmpAddr,
+		RtmpDisabled: true,
+		WorkDir:      lpNode.WorkDir,
+		HttpMux:      http.NewServeMux(),
 	}
 	switch lpNode.NodeType {
 	case core.BroadcasterNode:
 		opts.RtmpDisabled = false
-	case core.OrchestratorNode:
-		opts.HttpMux = http.NewServeMux()
 	}
 	server := lpmscore.New(&opts)
-	return &LivepeerServer{RTMPSegmenter: server, LPMS: server, LivepeerNode: lpNode, HTTPMux: opts.HttpMux, connectionLock: &sync.RWMutex{}, rtmpConnections: make(map[core.ManifestID]*rtmpConnection)}
+	ls := &LivepeerServer{RTMPSegmenter: server, LPMS: server, LivepeerNode: lpNode, HTTPMux: opts.HttpMux, connectionLock: &sync.RWMutex{},
+		rtmpConnections: make(map[core.ManifestID]*rtmpConnection),
+	}
+	if lpNode.NodeType == core.BroadcasterNode {
+		opts.HttpMux.HandleFunc("/live/", ls.HandlePush)
+	}
+	return ls
 }
 
-//StartServer starts the LPMS server
-func (s *LivepeerServer) StartMediaServer(ctx context.Context, transcodingOptions string) error {
+//StartMediaServer starts the LPMS server
+func (s *LivepeerServer) StartMediaServer(ctx context.Context, transcodingOptions string, httpAddr string) error {
 	BroadcastJobVideoProfiles = parsePresets(strings.Split(transcodingOptions, ","))
 
 	glog.V(common.SHORT).Infof("Transcode Job Type: %v", BroadcastJobVideoProfiles)
@@ -131,7 +139,8 @@ func (s *LivepeerServer) StartMediaServer(ctx context.Context, transcodingOption
 
 	//Start the LPMS server
 	lpmsCtx, cancel := context.WithCancel(context.Background())
-	ec := make(chan error, 1)
+
+	ec := make(chan error, 2)
 	go func() {
 		if err := s.LPMS.Start(lpmsCtx); err != nil {
 			// typically triggered if there's an error with broadcaster LPMS
@@ -139,6 +148,12 @@ func (s *LivepeerServer) StartMediaServer(ctx context.Context, transcodingOption
 			ec <- s.LPMS.Start(lpmsCtx)
 		}
 	}()
+	if s.LivepeerNode.NodeType == core.BroadcasterNode {
+		go func() {
+			glog.V(4).Infof("HTTP Server listening on http://%v", httpAddr)
+			ec <- http.ListenAndServe(httpAddr, s.HTTPMux)
+		}()
+	}
 
 	select {
 	case err := <-ec:
@@ -169,7 +184,7 @@ func createRTMPStreamIDHandler(s *LivepeerServer) func(url *url.URL) (strmID str
 			return nil
 		}
 		if resp != nil {
-			mid, key = parseStreamID(resp.ManifestID).ManifestID, resp.StreamKey
+			mid, key = parseManifestID(resp.ManifestID), resp.StreamKey
 			// Process transcoding options presets
 			if len(resp.Presets) > 0 {
 				presets = parsePresets(resp.Presets)
@@ -312,24 +327,12 @@ func endRTMPStreamHandler(s *LivepeerServer) func(url *url.URL, rtmpStrm stream.
 		if params == nil {
 			return errMismatchedParams
 		}
-		mid := params.mid
-		//Remove RTMP stream
-		s.connectionLock.Lock()
-		defer s.connectionLock.Unlock()
-		cxn, ok := s.rtmpConnections[mid]
-		if !ok || cxn.pl == nil {
-			glog.Error("Attempted to end unknown stream with manifest ID ", mid)
-			return errUnknownStream
-		}
-		cxn.sessManager.cleanup()
-		cxn.pl.Cleanup()
-		glog.Infof("Ended stream with id=%s", mid)
-		delete(s.rtmpConnections, mid)
-		if monitor.Enabled {
-			monitor.StreamEnded(cxn.nonce)
-			monitor.CurrentSessions(len(s.rtmpConnections))
-		}
 
+		//Remove RTMP stream
+		err := removeRTMPStream(s, params.mid)
+		if err != nil {
+			return err
+		}
 		return nil
 	}
 }
@@ -367,10 +370,12 @@ func (s *LivepeerServer) registerConnection(rtmpStrm stream.RTMPVideoStream) (*r
 	}
 	storage := drivers.NodeStorage.NewSession(string(mid))
 	// Build the source video profile from the RTMP stream.
-	resolution := fmt.Sprintf("%vx%v", rtmpStrm.Width(), rtmpStrm.Height())
+	if params.resolution == "" {
+		params.resolution = fmt.Sprintf("%vx%v", rtmpStrm.Width(), rtmpStrm.Height())
+	}
 	vProfile := ffmpeg.VideoProfile{
 		Name:       "source",
-		Resolution: resolution,
+		Resolution: params.resolution,
 		Bitrate:    "4000k", // Fix this
 	}
 	hlsStrmID := core.MakeStreamID(mid, &vProfile)
@@ -391,18 +396,42 @@ func (s *LivepeerServer) registerConnection(rtmpStrm stream.RTMPVideoStream) (*r
 		profile:     &vProfile,
 		params:      params,
 		sessManager: NewSessionManager(s.LivepeerNode, params, playlist),
+		lastUsed:    time.Now(),
 	}
+
 	s.connectionLock.Lock()
 	s.rtmpConnections[mid] = cxn
 	s.lastManifestID = mid
 	s.lastHLSStreamID = hlsStrmID
 	sessionsNumber := len(s.rtmpConnections)
 	s.connectionLock.Unlock()
+
 	if monitor.Enabled {
 		monitor.CurrentSessions(sessionsNumber)
 	}
 
 	return cxn, nil
+}
+
+func removeRTMPStream(s *LivepeerServer, mid core.ManifestID) error {
+	s.connectionLock.Lock()
+	defer s.connectionLock.Unlock()
+	cxn, ok := s.rtmpConnections[mid]
+	if !ok || cxn.pl == nil {
+		glog.Error("Attempted to end unknown stream with manifest ID ", mid)
+		return errUnknownStream
+	}
+	cxn.sessManager.cleanup()
+	cxn.pl.Cleanup()
+	glog.Infof("Ended stream with id=%s", mid)
+	delete(s.rtmpConnections, mid)
+
+	if monitor.Enabled {
+		monitor.StreamEnded(cxn.nonce)
+		monitor.CurrentSessions(len(s.rtmpConnections))
+	}
+
+	return nil
 }
 
 //End RTMP Publish Handlers
@@ -509,10 +538,104 @@ func getRTMPStreamHandler(s *LivepeerServer) func(url *url.URL) (stream.RTMPVide
 
 //End RTMP Handlers
 
+func (s *LivepeerServer) HandlePush(w http.ResponseWriter, r *http.Request) {
+	// we read this unconditionally, mostly for ffmpeg
+	body, err := ioutil.ReadAll(r.Body)
+
+	if err != nil {
+		httpErr := fmt.Sprintf(`Error reading http request body: %s`, err.Error())
+		glog.Error(httpErr)
+		http.Error(w, httpErr, http.StatusInternalServerError)
+		return
+	}
+	r.Body.Close()
+	r.URL = &url.URL{Scheme: "http", Host: r.Host, Path: r.URL.Path}
+
+	if ".ts" != path.Ext(r.URL.Path) {
+		// ffmpeg sends us a m3u8 as well, so ignore
+		// Alternatively, reject m3u8s explicitly and take any other type
+		// TODO also look at use content-type
+		http.Error(w, fmt.Sprintf(`ignoring file extension: %s`, path.Ext(r.URL.Path)), http.StatusBadRequest)
+		return
+	}
+
+	now := time.Now()
+	mid := parseManifestID(r.URL.Path)
+	s.connectionLock.Lock()
+	cxn, exists := s.rtmpConnections[mid]
+	if exists && cxn != nil {
+		cxn.lastUsed = now
+	}
+	s.connectionLock.Unlock()
+
+	// Check for presence and register if a fresh cxn
+	if !exists {
+		appData := (createRTMPStreamIDHandler(s))(r.URL)
+		if appData == nil {
+			http.Error(w, "Could not create stream ID: ", http.StatusInternalServerError)
+			return
+		}
+		st := stream.NewBasicRTMPVideoStream(appData)
+		params := streamParams(st)
+		params.resolution = r.Header.Get("Content-Resolution")
+
+		cxn, err = s.registerConnection(st)
+		if err != nil {
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+			return
+		}
+
+		ticker := time.NewTicker(RefreshIntervalHttpPush)
+
+		go func(s *LivepeerServer, mid core.ManifestID) {
+			for range ticker.C {
+				s.connectionLock.RLock()
+				lastUsed := s.rtmpConnections[mid].lastUsed
+				s.connectionLock.RUnlock()
+
+				if time.Since(lastUsed) > RefreshIntervalHttpPush {
+					_ = removeRTMPStream(s, mid)
+					ticker.Stop()
+					return
+				}
+			}
+		}(s, mid)
+	}
+
+	fname := path.Base(r.URL.Path)
+	seq, err := strconv.ParseUint(strings.TrimSuffix(fname, path.Ext(fname)), 10, 64)
+	if err != nil {
+		seq = 0
+	}
+
+	duration, err := strconv.Atoi(r.Header.Get("Content-Duration"))
+	if err != nil {
+		duration = 2000
+		glog.Info("Missing duration; filling in a default of 2000ms")
+	}
+
+	seg := &stream.HLSSegment{
+		Data:     body,
+		Name:     fname,
+		SeqNo:    seq,
+		Duration: float64(duration) / 1000.0,
+	}
+
+	// Do the transcoding!
+	err = processSegment(cxn, seg)
+	if err != nil {
+		// TODO return error
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	w.WriteHeader(http.StatusOK)
+}
+
 //Helper Methods Begin
 
 // Match all leading spaces, slashes and optionally `stream/`
-var StreamPrefix = regexp.MustCompile(`^[ /]*(stream/)?`)
+var StreamPrefix = regexp.MustCompile(`^[ /]*(stream/)?|(live/)?`) // test carefully!
 
 func cleanStreamPrefix(reqPath string) string {
 	return StreamPrefix.ReplaceAllString(reqPath, "")
--- a/server/mediaserver_test.go
+++ b/server/mediaserver_test.go
@@ -39,8 +39,8 @@ func setupServer() *LivepeerServer {
 	drivers.NodeStorage = drivers.NewMemoryDriver(nil)
 	if S == nil {
 		n, _ := core.NewLivepeerNode(nil, "./tmp", nil)
-		S = NewLivepeerServer("127.0.0.1:1938", "127.0.0.1:8080", n)
-		go S.StartMediaServer(context.Background(), "")
+		S = NewLivepeerServer("127.0.0.1:1938", n)
+		go S.StartMediaServer(context.Background(), "", "127.0.0.1:8080")
 		go S.StartCliWebserver("127.0.0.1:8938")
 	}
 	return S
@@ -873,4 +873,5 @@ func TestParsePresets(t *testing.T) {
 
 	p = parsePresets(presets)
 	assert.Equal([]ffmpeg.VideoProfile{ffmpeg.P240p30fps16x9, ffmpeg.P720p30fps16x9}, p)
+
 }
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -633,7 +633,7 @@ func main() {
 	}
 
 	//Set up the media server
-	s := server.NewLivepeerServer(*rtmpAddr, *httpAddr, n)
+	s := server.NewLivepeerServer(*rtmpAddr, n)
 	ec := make(chan error)
 	tc := make(chan struct{})
 	wc := make(chan struct{})
@@ -655,7 +655,7 @@ func main() {
 		close(wc)
 	}()
 	go func() {
-		ec <- s.StartMediaServer(msCtx, *transcodingOptions)
+		ec <- s.StartMediaServer(msCtx, *transcodingOptions, *httpAddr)
 	}()
 
 	go func() {
--- a/doc/ingest.md
+++ b/doc/ingest.md
@@ -84,3 +84,36 @@ Here, the stream name is `movie` and the stream key is `Secret/Stream/Key`.
 The RTMP stream can then be played back with this complete RTMP URL. The key is
 optional; if one is not supplied, then a random key will be generated. The key
 may also be specified via webhook.
+
+### HTTP Push
+
+Livepeer starts an HTTP server on the default port of 8935, as another ingest point
+into the Livepeer network. Upon ingest, HTTP stream is pushed to the segmenter
+prior to transcoding. The stream can be pushed via a PUT or POST HTTP request to the
+`/live/` endpoint. HTTP request timeout is 8 seconds.
+
+The following are expected in the request:
+
+* MPEG TS segments with segment numbers
+
+* `Content-Resolution` and `Content-Duration` headers
+
+* Stream name must be provided. In the examples below, the stream name is `movie`
+
+
+Sample URLs and requests:
+
+```
+# Push URL
+http://localhost:8935/live/movie/
+
+# HLS Playback URL
+http://localhost:8935/stream/movie.m3u8
+
+# Curl request
+curl -X PUT -H "Content-Duration: 2000" -H "Content-Resolution: 1920x1080"  --data-binary
+"@bbb0.ts" http://localhost:8935/live/movie/bbb0.ts
+
+# FFMPEG request
+ffmpeg -re -i movie.mp4 -c:a copy -c:v copy -f hls http://localhost:8935/live/movie/
+```
--- a/server/cliserver_test.go
+++ b/server/cliserver_test.go
@@ -23,7 +23,7 @@ func newMockServer() *httptest.Server {
 	go func() { n.TranscoderManager.Manage(strm, 5) }()
 	time.Sleep(1 * time.Millisecond)
 	n.Transcoder = n.TranscoderManager
-	s := NewLivepeerServer("127.0.0.1:1938", "127.0.0.1:8080", n)
+	s := NewLivepeerServer("127.0.0.1:1938", n)
 	mux := s.cliWebServerHandlers("addr")
 	srv := httptest.NewServer(mux)
 	return srv
--- a/server/push_test.go
+++ b/server/push_test.go
@@ -1,0 +1,190 @@
+package server
+
+import (
+	"encoding/json"
+	"io/ioutil"
+	"net/http"
+	"net/http/httptest"
+	"os"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/golang/glog"
+	"github.com/livepeer/go-livepeer/core"
+	"github.com/livepeer/go-livepeer/drivers"
+)
+
+func requestSetup(s *LivepeerServer) (http.Handler, *strings.Reader, *httptest.ResponseRecorder) {
+	handler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		s.HandlePush(w, r)
+	})
+	reader := strings.NewReader("")
+	writer := httptest.NewRecorder()
+	return handler, reader, writer
+}
+
+func TestNoErrors(t *testing.T) {
+	assert := assert.New(t)
+	handler, reader, w := requestSetup(setupServer())
+	req := httptest.NewRequest("POST", "/live/seg.ts", reader)
+
+	handler.ServeHTTP(w, req)
+	resp := w.Result()
+	defer resp.Body.Close()
+
+	body, err := ioutil.ReadAll(resp.Body)
+	require.Nil(t, err)
+	assert.Equal(200, resp.StatusCode)
+	assert.Equal(strings.TrimSpace(string(body)), "")
+}
+
+func TestMemoryRequestError(t *testing.T) {
+	// assert http request body error returned
+	assert := assert.New(t)
+	handler, _, w := requestSetup(setupServer())
+	f, err := os.Open(`doesn't exist`)
+	req := httptest.NewRequest("POST", "/live/seg.ts", f)
+
+	handler.ServeHTTP(w, req)
+	resp := w.Result()
+	defer resp.Body.Close()
+
+	body, err := ioutil.ReadAll(resp.Body)
+	require.Nil(t, err)
+	assert.Equal(http.StatusInternalServerError, resp.StatusCode)
+	assert.Contains(strings.TrimSpace(string(body)), "Error reading http request body")
+}
+
+func TestFileExtensionError(t *testing.T) {
+	// assert file extension error returned
+	assert := assert.New(t)
+	handler, reader, w := requestSetup(setupServer())
+	req := httptest.NewRequest("POST", "/live/seg.m3u8", reader)
+
+	handler.ServeHTTP(w, req)
+	resp := w.Result()
+	defer resp.Body.Close()
+
+	body, err := ioutil.ReadAll(resp.Body)
+	require.Nil(t, err)
+	assert.Equal(http.StatusBadRequest, resp.StatusCode)
+	assert.Contains(strings.TrimSpace(string(body)), "ignoring file extension")
+}
+
+func TestStorageError(t *testing.T) {
+	// assert storage error
+	assert := assert.New(t)
+	s := setupServer()
+	handler, reader, w := requestSetup(s)
+
+	tempStorage := drivers.NodeStorage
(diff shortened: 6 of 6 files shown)


server/mediaserver.go
```diff
@@ -510,10 +511,88 @@ func getRTMPStreamHandler(s *LivepeerServer) func(url *url.URL) (stream.RTMPVide
...
+	}
+
+	mid := parseManifestID(r.URL.Path)
+	s.connectionLock.RLock()
+	cxn, exists := s.rtmpConnections[mid]
+	s.connectionLock.RUnlock()
+
+	// Check for presence and register if a fresh cxn
+	if !exists {
+		appData := (createRTMPStreamIDHandler(s))(r.URL)
```
iameli, line 573: This (I think it's here) leads to a slight inconsistency in the webhook handling between HTTP and RTMP — the RTMP URLs come in with the full URL protocol like `rtmp://example.com/streamkey` but HTTP hooks come in with just the path `/live/streamkey/0.ts`. It'd probably be good to include the protocol in all cases in case the webhook server wants to disallow a protocol or something.

j0sh: Ah yeah looks like the `r.URL` here is simply extracted from the R-URI field and that's usually missing the scheme, host, port, etc. Looks like that'll need to be reconstructed from the golang [Request](https://golang.org/pkg/net/http/#Request) object.

j0sh: Updated the writeup to mention this should be addressed. Thanks!

iameli:
Thanks for throwing this together! Worked really well in the demo context.

> Path Prefix/Port numbering

I think your defaults are reasonable, but I kind of also want the ability to run the HTTP ingest on a separate port entirely. I know it seems like we're getting up to an unreasonable number of HTTP servers, but I like the idea of RTMP and HTTP ingest having the same security story — listening on localhost by default, optionally enhanced with a webhook. There are lots of current hobbyist workflows that involve setting up a local server, exposing 8935 publicly, and streaming OBS to localhost over RTMP — it'd be a shame not to be able to implement that same pattern using HTTP without a webhook.

> Durations / Other Missing Metadata
> The duration of the source segment is currently not passed in. The duration is necessary to insert valid timings for the resulting HLS playlist. [...] Might be nice to incorporate the width/height of the source segment as a header. We use this information in the master playlist to describe the dimensions of the source stream.

Confirmed this issue — the durations show up as zero in the M3U8, which stymies a variety of HLS players including our own. ffplay shrugs it off but polls for new segments from the manifest relentlessly, many times per second.

Something about passing duration/resolution as HTTP headers rubs me the wrong way, feels like a very ad-hoc protocol, but I don't really have a better solution at the moment. The data is actually present in the incoming manifests from both Wowza and ffmpeg, but I'm not wild about requiring that. Messy.

> Currently, nothing is returned in response. A multipart/mixed implementation can be borrowed from the standalone transcoder.

Let's allow for both no-return and `multipart/mixed` implementations as determined by the presence or absence of the `Accept: multipart/mixed` header? I think they both have valid use cases.

server/broadcast.go
```diff
@@ -261,15 +261,13 @@ func processSegment(cxn *rtmpConnection, seg *stream.HLSSegment) {
...
 	}
 
-	// Process the rest of the segment asynchronously - transcode
-	go func() {
-		for true {
-			// if fails, retry; rudimentary
-			if err := transcodeSegment(cxn, seg, name); err == nil {
-				return
-			}
+	for true {
```
mkrufky: `for (;;)`

mkrufky: Please remove `true` and change this to:
```golang
for {
   [...]
}
```

mkrufky: Do we really want to try this repeatedly forever?  What if it errors consistently?  Should there be a sanity limit here, rather than loop forever?

server/broadcast.go
```diff
@@ -285,6 +283,9 @@ func transcodeSegment(cxn *rtmpConnection, seg *stream.HLSSegment, name string)
 			monitor.SegmentTranscodeFailed(monitor.SegmentTranscodeErrorNoOrchestrators, nonce, seg.SeqNo, errNoOrchs, true)
 		}
 		glog.Infof("No sessions available for segment nonce=%d seqNo=%d", nonce, seg.SeqNo)
+		// We may want to introduce a "non-retryable" error type here
+		// would help error propagation for live ingest.
+		// similar to the orchestrator's RemoteTranscoderFatalError
```
mkrufky: perhaps move to using `github.com/cenkalti/backoff`  ....   this can be used to handle retries and retry limits and time between retries, etc

server/mediaserver.go
```diff
@@ -281,7 +282,7 @@ func gotRTMPStreamHandler(s *LivepeerServer) func(url *url.URL, rtmpStrm stream.
 						monitor.StreamStarted(nonce)
 					}
 				}
-				processSegment(cxn, seg)
+				go processSegment(cxn, seg)
```
mkrufky: I see that the entire function is run in a go routine now, as opposed to the old `for (true)` loop.   That seems fine.  So my comment here does not apply to this change, but it applies to the general idea of running the transcode function in a go routine at all -- 

Is it safe to spawn an arbitrary number of go routines here?  If lots of segments come in simultaneously, will it cause contention?

This looks like a job for a workpool.   I wrote a generic workpool module that might save the day, here.

https://github.com/mkrufky/genepool

server/mediaserver.go
mkrufky: ```r.URL.RequestURI()``` ?

server/mediaserver.go
```diff
@@ -510,10 +511,88 @@ func getRTMPStreamHandler(s *LivepeerServer) func(url *url.URL) (stream.RTMPVide
...
+
+	// Check for presence and register if a fresh cxn
+	if !exists {
+		appData := (createRTMPStreamIDHandler(s))(r.URL)
+		if appData == nil {
+			return
+		}
+		st := stream.NewBasicRTMPVideoStream(appData)
+		cxn, err = s.registerConnection(st)
+		// TODO monitor for activity and clean up
```
mkrufky: activity monitor and cleanup should block this PR

server/mediaserver.go
```diff
@@ -510,10 +511,88 @@ func getRTMPStreamHandler(s *LivepeerServer) func(url *url.URL) (stream.RTMPVide
...
+	if sess == nil {
+		return
+	}
+	params := streamParams(cxn.stream)
+	if params == nil {
+		return
+	}
+	// fetch each rendition from local storage
+	for _, p := range params.profiles {
+		name := fmt.Sprintf("%s/%s/%s", mid, p.Name, fname)
```
mkrufky: does this do anything?   is it just a TO DO bookmark?  If so, can we have some more detailed comments illustrating what the plan is, here?

server/webserver.go
```diff
@@ -69,6 +69,10 @@ func (s *LivepeerServer) StartCliWebserver(bindAddr string) {
 func (s *LivepeerServer) cliWebServerHandlers(bindAddr string) *http.ServeMux {
 	mux := http.NewServeMux()
 
+	mux.HandleFunc("/live/", func(w http.ResponseWriter, r *http.Request) {
```
mkrufky: is `live` the most appropriate name for this endpoint?

how about `livesegs` ? `livesegment` ? `seglive` ?I think `live` is going to get confusing later on

server/broadcast.go
j0sh: We don't really want to back off here (retry immediately with another O), although whether we should stop after a certain number of retries is something that can be discussed separately.

server/broadcast.go
j0sh: We retry only for as long as there is an orchestrator available in the working set. If the working set empties, `nil` is returned as a signal to exit. See the additional note within this PR about perhaps being a little more explicit about "non-retryable errors".

But as the comment following this line mentions, this is rudimentary :) How to improve the retry strategy is something that can be discussed / implemented separately outside this PR.

server/mediaserver.go
j0sh: > If lots of segments come in simultaneously, will it cause contention?

Normally they should not under the "segmented live stream" assumption: a properly configured single stream should produce single segments at a comfortable cadence. Pushing RTMP streams through at wire speed (eg, via ffmpeg without `-re`) is not really a use case we should put much effort into supporting at the moment, IMO. At the very least, we'll run out of orchestrators and drop segments if that happens.

We do have the capability to absorb occasional fluctuations though; I believe the default is 8 orchestrators in the working set (one O per inflight segment).

If/when we start looking at other types of videos (eg, VOD) then we can look at building workflows specialized around that. Trying to accommodate disparate use cases under the same workflow leads to mediocrity.

> This looks like a job for a workpool.

Genepool ooks pretty nice. The B session manager is essentially a worker pool as well, albeit with some additional logic around managing Os.

server/mediaserver.go
j0sh: Yes, it's mentioned under "State Cleanup"

server/mediaserver.go
j0sh: It's necessary for constructing the multipart response. The data for each segment is fetched from local OS using that name.

server/webserver.go
j0sh: Open to other names. What's confusing about `/live/` ?

I like the brevity of `/live/` though, and can see a path towards us supporting non-segmented transcoding under this API; see the "Non-Segmented Transcodes" section.

mk-livepeer: Personally, I believe that we will discover, in the future, that this is not how we want to transmit live streams.   At that point, we probably won't remove this endpoint, but we'll add another live endpoint, and what will it be called?  As such, the name `live` is very confusing.

server/broadcast.go
mk-livepeer: I'm suggesting the fact that the cenktali package includes a solution for non-retryable errors.  Especially if we already have that package linked to this project, it makes sense to use its non-retryable error flag.

server/broadcast.go
mk-livepeer: I reiterate my request: Please remove `true` and change this to:
```golang
for {
   [...]
}
```

j0sh:
> run the HTTP ingest on a separate port entirely

Yeah, that's a potential use case. I'm torn, because I also like minimizing the exposed surface area on the node. Curious what others think.

> The data is actually present in the incoming manifests from both Wowza and ffmpeg

Is the data available programmatically if you're able to access the source segments within Wowza? I have a little patch adding this to ffmpeg that should be cleaned up and submitted...

> Let's allow for both no-return and multipart/mixed implementations as determined by the presence or absence of the Accept: multipart/mixed header? I think they both have valid use cases.

Haven't checked yet but is a response like 200 OK even returned right now? If not, should it be?

I like using the Accept header to signal the format of the response body. Part of me wonders if we should make the default a bit more user-friendly, like a list of transcoded URLs or even a zip/tar file with the results (would be a super cool feature in conjunction with "Non-Segmented Transcodes") but that's something we can figure out later.

server/webserver.go
angyangie: @mk-livepeer are you referring to your comments on using the object store as a transport mechanism for the segment? Or an entirely different approach to transmitting live streams?

server/broadcast.go
angyangie: Why does this matter? Just stylistic preference?

server/mediaserver.go
angyangie: > Normally they should not under the "segmented live stream" assumption

To confirm, the issue we had before was that multiple frames were being sent at a time, not multiple segments? I believe this was the PR that fixed it: https://github.com/livepeer/lpms/pull/131

j0sh: Yes, the LPMS PR fixes an unrelated issue with multiple frames being made available from within the transcoder. @mkrufky is referring to the B receiving a deluge of segments from a stream simultaneously.

server/broadcast.go
mkrufky: We should write idiomatic code in golang.  Generally, there is no place for "style" in golang.  There is the idiomatic way, and there is everything else.   The idiomatic way to loop endlessly until internal interruption is as I described above.

server/webserver.go
mkrufky: I'm not talking about any one solution.  I am saying that "live" is too generic a name for this endpoint, as we are likely to end up having multiple ingest methods.

iameli:
LivepeerWowza's [PushPublishHTTPCupertinoLivepeerHandler](https://github.com/livepeer/livepeer-wowza/blob/master/src/org/livepeer/LivepeerWowza/PushPublishHTTPCupertinoLivepeerHandler.java) is getting intermittent timeouts hitting this:

```
ERROR server comment - PushPublishHTTPCupertinoLivepeerHandler.sendMediaSegment([null] live/_definst_/bbb->bbb) Failed to send media segment data to http://wowza-tester.default.svc.cluster.local:7935/live/7d5ec710-3cec-40ad-9246-6be12d929c48/bbb/yl5ez1f9/6.ts: java.net.SocketTimeoutException: Read timed out|at java.net.SocketInputStream.socketRead0(Native Method)|at java.net.SocketInputStream.socketRead(SocketInputStream.java:116)|at java.net.SocketInputStream.read(SocketInputStream.java:170)|at java.net.SocketInputStream.read(SocketInputStream.java:141)|at java.io.BufferedInputStream.fill(BufferedInputStream.java:246)|
ERROR server comment - PushPublishHTTPCupertinoLivepeerHandler.outputSend([null] live/_definst_/bbb->bbb) Send media segment. rendition: AUDIOVIDEO chunkId:6 uri:yl5ez1f9/media_6.ts result:FAILURE
```

I'm in the process of digging in, but @j0sh you mentioned that there might not be a 200 OK coming from the server? Is it possible we're currently just letting the connections sit there open until the client times out, even though everything else is working? @angyangie something you could look into maybe?

(I tried [a slight tweak](https://github.com/livepeer/go-livepeer/commit/11c33cc645ad88d0042317940c071a59fa995b83) last night but it gave me errors about redundant transmission of headers so I suspect I don't quite understand how out HTTP response pipeline functions.

server/webserver.go
iameli: ...reflecting on this, I actually really like that video ingress is at `/live` and video egress is at `/stream`. Feels consistent.

angyangie:
Hi all! Seems like there is a lot left here to decide. To move forward with this PR, we need consensus on a couple of points below. I also included some of my own questions. @j0sh, @iameli, would be great to hear your thoughts on this.

1. Why do we need something to be returned in responses here, to require a multipart/mixed implementation? Which response are we talking about? I see a comment on returning "a list of transcoded URLs or even a zip/tar file with the results" ... do we agree what should be returned?

2. `Reconstruct target URL` section: I see "having the full target URL would be useful for the auth webhook" .... where are we connecting to the auth webhook?

3. `If auth fails or there is an unrecoverable error, we should stop processing requests for that stream` ... what kind of other errors are we talking about here? Where are the errors being returned?

4. Re: missing metadata - are we suggesting adding content-duration to `seg := &stream.HLSSegment` & width/height? Where exactly would these be used? I see we need the former for `valid timings for the resulting HLS playlist` ... is that something that happens in LPMS? What about height/width? I also see arguments against requiring content-duration/resolution.

5. HTTP ingest - arguments were made for port 7935 (current setup), for 8935 and using a flag for push API, disabling it automatically if there is no webhook auth present/non-localhost, and setting it up on a different port entirely. We should get consensus around this! Thoughts?

6. A `monitor for activity` is mentioned here: https://github.com/livepeer/go-livepeer/pull/972/files/4ea9a62304384adba7367fb890404f5eae946472#diff-236217e7f473dfbde9134c6a391f3408R544. What would that look like? Is this something we want to include in this PR?

7. Path prefix ... `/live/` v. others. What will it be?

8. Non-Segmented Transcodes: to enable (JIT VOD) .... seems like this should be handled in a separate issue and could be done at a later point?

9. Use this push ingest to receive segments via HTTP from an RTMP stream ... same as #8 above? Seems like there are a lot of things to think through before tackling this one. Not critical for the Wowza plugin.

iameli:
Thanks for driving this forward @angyangie!

> 1. Why do we need something to be returned in responses here, to require a multipart/mixed implementation? Which response are we talking about? I see a comment on returning "a list of transcoded URLs or even a zip/tar file with the results" ... do we agree what should be returned?

I think there is not consensus on the point. I would be entirely comfortable shipping this PR without any return type at all — [as discussed here](https://github.com/livepeer/go-livepeer/pull/972#issuecomment-510196290), any future return types can be added through the `Accept:` header.

> 2. `Reconstruct target URL` section: I see "having the full target URL would be useful for the auth webhook" .... where are we connecting to the auth webhook?

I totally forgot that I made a branch to test @mk-livepeer's proposed fix for this! https://github.com/livepeer/go-livepeer/tree/eli/liveingest-request-uri <-- if that works, my concern is fully addressed.

> 4. Re: missing metadata - are we suggesting adding content-duration to `seg := &stream.HLSSegment` & width/height? Where exactly would these be used? I see we need the former for `valid timings for the resulting HLS playlist` ... is that something that happens in LPMS? What about height/width? I also see arguments against requiring content-duration/resolution.

Yeah... this is my biggest area of ambivalence in this PR. I see a few possible paths forward:

A. Ship as-is, with the segment lengths for the source stream being listed as zero in the HTTP output.

B. Provide bitrate/duration data in HTTP headers. I like that this is low-overhead and could be implemented quickly. I dislike that it's an ad-hoc, Livepeer-specific protocol that won't be followed by any "HLS Push" except for LivepeerWowza's implementation.

C. Implement `ffprobe`-style discovery of bitrate/duration. This would definitely work but is really unfortunate overhead for the broadcaster to incur. If we go this route, it should probably enabled/disabled through a CLI flag.

D. Consume the manifests that are being provided by the LivepeerWowza plugin and ffmpeg, get the bitrate data from there. I like that this feels closest to the HLS spec. I also like that the work to implement this would likely overlap with the work necessary to implement an "HLS Pull" feature in go-livepeer.

Loosely held ranking of these: D, B, A, and C.

> 5. HTTP ingest - arguments were made for port 7935 (current setup), for 8935 and using a flag for push API, disabling it automatically if there is no webhook auth present/non-localhost, and setting it up on a different port entirely. We should get consensus around this!

I'm leaning toward - it should be at `http://0.0.0.0:8935/live` but the webhook should be required unless explicitly disabled through a CLI flag. `--allow-anonymous-http-push` or some such.

> 6. A `monitor for activity` is mentioned here: https://github.com/livepeer/go-livepeer/pull/972/files/4ea9a62304384adba7367fb890404f5eae946472#diff-236217e7f473dfbde9134c6a391f3408R544. What would that look like?

I'm pretty sure that the API server could get any information it needs from the `/status` endpoint.

> 7. Path prefix ... `/live/` v. others. What will it be?

I'm rather enamored with HTTP video ingress being at `http://0.0.0.0:8935/live` and HTTP video egress being at `http://0.0.0.0:8935/stream`. (When I say "live", you say "stream"!) But I'm totally fine deferring to consensus for an alternative name.

EDIT: I'm unopinionated on 8 and 9. I think the phrase "JIT VOD" is fun to say, though.

angyangie:
Thanks @iameli for the VERY speedy response!

> I think there is not consensus on the point. I would be entirely comfortable shipping this PR without any return type at all 

Which response are we talking about? So return nothing then?

> I totally forgot that I made a branch to test @mk-livepeer's proposed fix for this! 

Awesome!

> Yeah... this is my biggest area of ambivalence in this PR

Option D to get bitrate data from manifests that are being provided by the LivepeerWowza plugin and ffmpeg sounds reasonable. Where do we use the bitrate/duration data?

> I'm pretty sure that the API server could get any information it needs from the /status endpoint.

True! Didn't think about that. 

> I'm leaning toward - it should be at `http://0.0.0.0:8935/live` but the webhook should be required unless explicitly disabled through a CLI flag. `--allow-anonymous-http-push` or some such.

Isn't authentication already required in `createRTMPStreamIDHandler`? So what we'd be building would be a way to NOT do `authenticateStream` if flag was provided?

> I'm rather enamored with HTTP video ingress being at http://0.0.0.0:8935/live and HTTP video egress being at http://0.0.0.0:8935/stream. (When I say "live", you say "stream"!) But I'm totally fine deferring to consensus for an alternative name.

I don't have a preference here. I'm fine with that.

> EDIT: I'm unopinionated on 8 and 9. I think the phrase "JIT VOD" is fun to say, though.

Cooool. My vote is to handle in a separate issue and do at a later point.

j0sh:
> Option D to get bitrate data from manifests that are being provided by the LivepeerWowza plugin and ffmpeg sounds reasonable. Where do we use the bitrate/duration data?

The problem with doing this is that we have to handle it across separate requests: one request to receive the M3U8, and another request to receive the mpegts file. (And I don't think we can always depend on receiving one before or after the other.) That's increased bookkeeping we really could do without.

> I like that this feels closest to the HLS spec. I also like that the work to implement this would likely overlap with the work necessary to implement an "HLS Pull" feature in go-livepeer.

Note that this is much closer to "HLS Push" rather than "HLS Pull". A true "HLS Pull" feature wouldn't have the client also pushing segments or manifests.

Another downside here is that we'd be baking in the expectation of a corresponding manifest, when there is otherwise no reason to need one. The less we can make this HLS-specific, the better.

>  Option B ... I dislike that it's an ad-hoc, Livepeer-specific protocol that won't be followed by any "HLS Push" except for LivepeerWowza's implementation.

Which is totally OK, because nearly all APIs are ad-hoc like that. "HLS Push" isn't a standard in any meaningful sense.

>> A monitor for activity is mentioned here: What would that look like?

> I'm pretty sure that the API server could get any information it needs from the /status endpoint.

Doesn't feel appropriate to depend on an external service to track and clean up idle sessions. That can be done internally to goclient.

> I'm leaning toward - it should be at http://0.0.0.0:8935/live but the webhook should be required unless explicitly disabled through a CLI flag. --allow-anonymous-http-push or some such.

Note that the broadcaster listens to localhost by default. I don't think we should change it.

So your proposal is that we make `--allow-anonymous-http-push` a required flag *unless* `-authWebhookUrl` is also specified?

I'm unsure because it basically means we're introducing a required CLI flag on the broadcaster. Personally, I think I would rather toggle the push API based on:

* Whether we're running on localhost - default enabled
* Non-localhost - default disabled unless explicitly enabled

> what kind of other errors are we talking about here? Where are the errors being returned?

Anywhere we have an error that is ignored right now. We currently don't return any errors to the client; they are all ignored.

> I totally forgot that I made a branch to test @mk-livepeer's proposed fix for this! https://github.com/livepeer/go-livepeer/tree/eli/liveingest-request-uri <-- if that works, my concern is fully addressed.

I don't think that will actually be any different from what we're currently sending, since golang already populates the URL field with the request URI [1]. If this is an issue that needs to be resolved, we need to manually rebuild the address with the scheme / host as the writeup describes.

[1] https://golang.org/pkg/net/http/#Request
```golang
        // For server requests, the URL is parsed from the URI
        // supplied on the Request-Line as stored in RequestURI.  For
        // most requests, fields other than Path and RawQuery will be
        // empty. (See RFC 7230, Section 5.3)
```

server/mediaserver.go
```diff
@@ -393,20 +387,61 @@ func (s *LivepeerServer) registerConnection(rtmpStrm stream.RTMPVideoStream) (*r
...
 	}
+
 	s.connectionLock.Lock()
 	s.rtmpConnections[mid] = cxn
 	s.lastManifestID = mid
 	s.lastHLSStreamID = hlsStrmID
 	sessionsNumber := len(s.rtmpConnections)
 	s.connectionLock.Unlock()
+
+	ticker := getConnectionTicker()
```
j0sh: Not sure we want to also start the cleanup ticker for RTMP as well. We can probably move the following bits into the `HandlePush` function, within the `if !exists` check.

server/mediaserver.go
```diff
@@ -393,20 +387,61 @@ func (s *LivepeerServer) registerConnection(rtmpStrm stream.RTMPVideoStream) (*r
...
+	go func(s *LivepeerServer, mid core.ManifestID) {
+		for _ = range ticker.C {
+			lastUsed := s.rtmpConnections[mid].lastUsed
```
j0sh: Needs locking around any usage of `rtmpConnections`; see

https://github.com/livepeer/go-livepeer/blob/d41cf5d7682c3deff44a32bad6b764d350886717/server/mediaserver.go#L90-L95

server/mediaserver.go
```diff
@@ -393,20 +387,61 @@ func (s *LivepeerServer) registerConnection(rtmpStrm stream.RTMPVideoStream) (*r
...
+			now := time.Now()
+			diff := now.Sub(lastUsed)
+			if diff.Hours() >= 1 || diff.Minutes() >= 1 {
```
j0sh: One hour timeout is a really long time, given that segments are just a few seconds long. One minute seems enough.

Assuming we reuse `connectionRefreshInterval` as the 1-minute timeout, this approach is more terse:

```golang
if time.Since(lastUsed) > connectionRefreshInterval {
 // ... remove RTMP stream ...
}
```

server/mediaserver.go
```diff
@@ -393,20 +387,61 @@ func (s *LivepeerServer) registerConnection(rtmpStrm stream.RTMPVideoStream) (*r
...
+				err := removeRTMPStream(s, mid)
+				if err != nil {
+					continue
```
j0sh: Let's just stop the ticker here regardless of whether an error occurs.

Otherwise, the the ticker loops indefinitely if an error occurs. Since we're not sure such errors are actually recoverable, this is tantamount to a memory leak.

This would currently happen with the RTMP ingest workflow:

* RTMP comes in, cxn registers, ticker is kicked off
* RTMP disconnects, cxn is cleaned up
* Ticker ticks, finds no RTMP connection, returns error
* Ticker ticks, finds no RTMP connection, returns error
* ...

(As noted earlier, we probably don't need to actually start the cleanup goroutine for RTMP ingest. Doing so could be useful one day to resume a HLS session after a RTMP disconnect, but that's outside the scope of this PR.)

j0sh (commented):
@angyangie The `lastUsed` needs to be reset on each incoming request. Also, all accesses to `cxn.lastUsed` need to be done under a lock ; we can reuse `connectionLock` for the purpose. Something like this:

```golang
        now := time.Now()
	mid := parseManifestID(r.URL.Path)
	s.connectionLock.Lock()
	cxn, exists := s.rtmpConnections[mid]
        if exists && cxn != nil {
            cxn.lastUsed = now
        }
	s.connectionLock.Unlock()

        if !exists {
            ...
        }
```

Not totally ideal since we're also write-locking all connections. We could also introduce another lock on the `rtmpConnection` itself, and maintain the read-lock for connections.

server/mediaserver.go
angyangie: I misunderstood and assumed we wanted it for both. Moving to `HandlePush` makes sense

server/mediaserver.go
angyangie: True, a minute timeout would negate the need to check for an hr difference

server/mediaserver.go
angyangie: Thanks for the detailed explanation! Is this what you're suggesting? If `removeRTMPStream` errors and cxn isn't cleaned up, wouldn't that be a memory leak if ticker is stopped and cleanup isn't retried?

```golang
if time.Since(lastUsed) > connectionRefreshInterval {
     _ = removeRTMPStream(s, mid)
    ticker.Stop()
    return
}
```

j0sh: >  If removeRTMPStream errors and cxn isn't cleaned up, wouldn't that be a memory leak if ticker is stopped and cleanup isn't retried?

That would be true if the error was somehow recoverable by the next tick. Don't think we have any cases like that.

The code sample looks good. Could probably omit the `_ =` from `removeRTMPStream`.

angyangie:
Since there seems to be a lot of discussion on several issues here. I've made separate issues for some that warrant more discussion to make things easier to follow (wish the github UI was more capable...). I tried to separate out what'd been discussed on each. 

1. Add multipart responses: https://github.com/livepeer/go-livepeer/issues/1003
2. Add duration and other missing metadata to source segment: https://github.com/livepeer/go-livepeer/issues/1002
3. Path prefix and port numbering: https://github.com/livepeer/go-livepeer/issues/1004
4. Non-segmented transcodes & RTMP segmentation: https://github.com/livepeer/go-livepeer/issues/1005
5. Reconstruct Target URL: https://github.com/livepeer/go-livepeer/issues/1006

@iameli and @j0sh .... do we think we should tackle any of the 4 issues above in this PR? Or could we work on separate PR's for each of these once we come to consensus? Did I miss anything? Trying to push us forward to continue discussing & start to merge parts of this incrementally. 

There is one minor point on a monitor for activity I didn't include in issues above. Do we want to see an issue created for this one as well? Discussion summary is below:

(Angie) A monitor for activity is mentioned [here](https://github.com/livepeer/go-livepeer/pull/972/files/4ea9a62304384adba7367fb890404f5eae946472#diff-236217e7f473dfbde9134c6a391f3408R544). What would that look like? Is this something we want to include in this PR?

(Eli) I'm pretty sure that the API server could get any information it needs from the /status endpoint.

(Josh) Doesn't feel appropriate to depend on an external service to track and clean up idle sessions. That can be done internally to goclient.

j0sh:
Thanks for the issue breakdown @angyangie . I'll leave whether to address any of these up to you and @iameli based on the requirements of the plugin.

Re: the activity monitor / state cleanup, that is already being addressed by https://github.com/livepeer/go-livepeer/pull/972/commits/8b75cc2fa44aaf7f0715600469f4ff0e1effbd01 and https://github.com/livepeer/go-livepeer/pull/972/commits/b762d30bd5514c8ad3f265458b7c788b5a80ad1c

server/mediaserver.go
```diff
@@ -576,7 +562,24 @@ func (s *LivepeerServer) HandlePush(w http.ResponseWriter, r *http.Request) {
 		st := stream.NewBasicRTMPVideoStream(appData)
 		cxn, err = s.registerConnection(st)
 		// TODO monitor for activity and clean up
+
+		ticker := getConnectionTicker()
```
j0sh: The ticker could be scoped within the goroutine

server/mediaserver.go
```diff
@@ -576,7 +562,24 @@ func (s *LivepeerServer) HandlePush(w http.ResponseWriter, r *http.Request) {
...
+
+		go func(s *LivepeerServer, mid core.ManifestID) {
```
j0sh, line 590: Minimizing free variables isn't strictly necessary here, but I like it anyway :+1:

server/mediaserver_test.go
```diff
@@ -874,3 +875,24 @@ func TestParsePresets(t *testing.T) {
 	p = parsePresets(presets)
 	assert.Equal([]ffmpeg.VideoProfile{ffmpeg.P240p30fps16x9, ffmpeg.P720p30fps16x9}, p)
 }
+
+func TestHandlePush(t *testing.T) {
```
j0sh: Tests need to be added covering all code paths within HandlePush, including errors and timeouts.

This was missing from the original writeup, but we might also want to return error responses in case an error does occur.

server/mediaserver.go
```diff
@@ -509,10 +523,114 @@ func getRTMPStreamHandler(s *LivepeerServer) func(url *url.URL) (stream.RTMPVide
...
+		}
+		st := stream.NewBasicRTMPVideoStream(appData)
+		cxn, err = s.registerConnection(st)
+		// TODO monitor for activity and clean up
+
+		ticker := getConnectionTicker()
+
+		go func(s *LivepeerServer, mid core.ManifestID) {
+			for _ = range ticker.C {
+				s.connectionLock.Lock()
```
j0sh: This can be a RLock/RUnlock

server/mediaserver_test.go
```diff
@@ -877,58 +877,131 @@ func TestParsePresets(t *testing.T) {
...
+	assert.Equal(http.StatusInternalServerError, resp.StatusCode)
+	assert.Contains(strings.TrimSpace(string(body)), "ErrStorage")
+}
+
+func TestHandlePushAppDataEmptyError(t *testing.T) {
+	// assert app data error
+	assert := assert.New(t)
+	s := setupServer()
+	handler, reader, w := handlePushRequestSetup(s)
+	AuthWebhookURL = "notaurl"
```
j0sh: Needs to reset the URL to the original value after running this case

server/mediaserver_test.go
```diff
@@ -877,58 +877,131 @@ func TestParsePresets(t *testing.T) {
...
 	// assert storage error
-	mid := parseManifestID(request.URL.Path)
-	err = removeRTMPStream(s, mid)
-	assert.Nil(err)
-	err = s.HandlePush(writer, request)
-	assert.Equal("ErrStorage", err.Error())
+	assert := assert.New(t)
+	s := setupServer()
+	handler, reader, w := handlePushRequestSetup(s)
+	drivers.NodeStorage = nil
```
j0sh: Need to reset `drivers.NodeStorage` to the original value after running this case

server/mediaserver_test.go
```diff
@@ -877,58 +877,131 @@ func TestParsePresets(t *testing.T) {
...
+	resp := w.Result()
+	defer resp.Body.Close()
+
+	body, err := ioutil.ReadAll(resp.Body)
+	require.Nil(t, err)
+	assert.Equal(http.StatusInternalServerError, resp.StatusCode)
+	assert.Contains(strings.TrimSpace(string(body)), "ErrStorage")
+}
+
+func TestHandlePushAppDataEmptyError(t *testing.T) {
```
j0sh: Maybe a better way to describe it is "test for auth webhook failure" rather than app data

server/mediaserver_test.go
```diff
@@ -877,58 +877,131 @@ func TestParsePresets(t *testing.T) {
...
+}
+
+func TestHandlePushFileExtensionError(t *testing.T) {
 	// assert file extension error returned
-	request = httptest.NewRequest("GET", "/live/angie.m3u8", reader)
-	err = s.HandlePush(writer, request)
-	assert.Equal("ignoring file extension: .m3u8", err.Error())
+	assert := assert.New(t)
+	handler, reader, w := handlePushRequestSetup(setupServer())
+	req := httptest.NewRequest("GET", "/live/seg.m3u8", reader)
```
j0sh: This might actually break ffmpeg with `ffmpeg -re -i <input> -c:a copy -c:v copy -f hls http://localhost:7935/live/<streamname>/` Can you check?

server/mediaserver_test.go
```diff
@@ -877,58 +877,131 @@ func TestParsePresets(t *testing.T) {
...
 	defer resp.Body.Close()
 
+	body, err := ioutil.ReadAll(resp.Body)
+	require.Nil(t, err)
 	glog.Error(resp.StatusCode)
 	assert.Equal(200, resp.StatusCode)
+	assert.Equal(strings.TrimSpace(string(body)), "")
 }
 
-func TestHandlePushErrors(t *testing.T) {
```
j0sh: The earlier setup of most simple tests cases the same function was concise, is there a reason they were split up?

server/mediaserver.go
```diff
@@ -626,7 +632,6 @@ func (s *LivepeerServer) HandlePush(w http.ResponseWriter, r *http.Request) erro
 		glog.Infof("%s - %d bytes", name, len(data))
 	}
 	w.WriteHeader(200)
```
j0sh: As @mk-livepeer  mentioned, `http.StatusOK` is more idiomatic here

server/mediaserver.go
```diff
@@ -555,12 +555,14 @@ func (s *LivepeerServer) HandlePush(w http.ResponseWriter, r *http.Request) erro
...
+			return
 		}
 		st := stream.NewBasicRTMPVideoStream(appData)
 		cxn, err = s.registerConnection(st)
 		if err != nil {
-			return err
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+			return
 		}
 		// TODO monitor for activity and clean up
```
j0sh: This comment can probably be removed

server/mediaserver.go
```diff
@@ -524,22 +524,22 @@ func getRTMPStreamHandler(s *LivepeerServer) func(url *url.URL) (stream.RTMPVide
...
 	}
 	r.Body.Close()
 
 	if ".ts" != path.Ext(r.URL.Path) {
 		// ffmpeg sends us a m3u8 as well, so ignore
 		// Alternatively, reject m3u8s explicitly and take any other type
 		// TODO also look at use content-type
-		err := fmt.Sprintf(`ignoring file extension: %s`, path.Ext(r.URL.Path))
-		return errors.New(err)
+		http.Error(w, fmt.Sprintf(`ignoring file extension: %s`, path.Ext(r.URL.Path)), http.StatusInternalServerError)
```
j0sh: If we return an error here, it should be a `StatusBadRequest` . But first check whether this breaks the ffmpeg testing workflow.

server/mediaserver.go
```diff
@@ -555,12 +555,14 @@ func (s *LivepeerServer) HandlePush(w http.ResponseWriter, r *http.Request) erro
 	if !exists {
 		appData := (createRTMPStreamIDHandler(s))(r.URL)
 		if appData == nil {
-			return errors.New("stream.AppData is empty")
+			http.Error(w, "stream.AppData is empty", http.StatusInternalServerError)
```
j0sh: "Could not create stream ID" is a better way of describing it

server/mediaserver.go
```diff
@@ -597,23 +599,27 @@ func (s *LivepeerServer) HandlePush(w http.ResponseWriter, r *http.Request) erro
 	err = processSegment(cxn, seg)
 	if err != nil {
 		// TODO return error
-		return err
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
 	}
 
 	// Return the results (blobs) using local OS lookup.
```
j0sh: If we're not actually returning data then this whole part can probably be tossed

angyangie: Which part do you mean?

server/mediaserver_test.go
angyangie: It does break ffmpeg!

server/mediaserver.go
j0sh: The rest of the code in this function was setting things up to return the transcoded contents as binary data, but if that's not being done, all the code after this point can all be removed.

angyangie:
@j0sh @darkdarkdragon any more thoughts here? I'd love to be able to merge this soon.

darkdarkdragon (approved):
LGTM

j0sh:
Since other PRs depend on this branch, let's merge those first, then we can pull everything into master.

Would be good to add some documentation in `/doc` for HTTP push as well.

Issue #955, opened 2019-07-01
O should validate pricePerPixel sent by B


yondonfu:
**Is your feature request related to a problem? Please describe.**

(To be filled in)

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

kyriediculous:
We could validate the `ExpectedPrice` B sends along with `Payment` in `Orchestrator.ProcessPayment` in a similar fashion to how we handle acceptable payment errors. 

Potentially we could also distinguish between an acceptable price and a non-acceptable one by implementing a maximum Delta. 

If the Delta is too large we would throw an error from `ProcessPayment` which makes B refetch `OrchestratorInfo` 

If the Delta is acceptable we would call `Recipient.SenderMonitor.AcceptErr()` to check whether any more acceptable errors are allowed for `sender` 

The drawback of this approach is that `PriceInfo` is a concern of `LivepeerNode` and not `LivepeerNode.Recipient` and thus not specific to the `pm` package. This implies we should not be using `pm.Error` here but simply something like the pseudocode below, which might not be the most beautiful solution. We'd also have to expose the `SenderMonitor.AcceptErr(sender)` method on the `Recipient` interface. Alternatively we could only check against a Delta and leave an acceptable error count out of the picture. 

```
...
if payment.Sender == nil || len(payment.Sender) == 0 {
	return fmt.Errorf("Could not find Sender for payment: %v", payment)
}

ok, err := orch.acceptablePrice(payment.Sender, payment.ExpectedPrice)
if  !ok { 
    return err
} else {
 glog.Errorf("%v", err)
}

seed := new(big.Int).SetBytes(payment.TicketParams.Seed)
...
```

```
func (orch *orchestrator) acceptablePrice(sender ethcommon.Address, ep *net.PriceInfo) (bool, error) {
  // convert ep to big.Rat 'epRat'
  orchPrice := orch.PriceInfo(sender)
  // convert orchPrice to big.Rat 'orchPriceRat'

 // happy case, price is good
  if eprat.Cmp(orchPriceRat) >= 0 {
 return true, nil
}
  
  // If price delta is not acceptable return an error
  // we are only concerned when orchPrice-Expected Price is positive
  if epRat.Sub(orchPriceRat, epRat).Cmp(orch.priceDelta) == 1 {
    return false, fmt.Errorf("unacceptable price")
 } else {
  // If price delta is acceptable see if we are not exceeding the max Error count. 
[...]

PR #979, opened 2019-07-11
core, net, server: debit fees for transcoded result & add pixel count to TranscodedSegmentData

### LINKED ISSUE #892: Add a TranscodeReceipt message for reporting pixels processed in TranscodedSegmentData message
Introduce a new TranscodeReceipt message.

```
message TranscodeReceipt {
    
    // Number of pixels processed to produce the transcoded data
    int64 pixels_processed = 1;
    
    // Price info used during transcoding
    PriceInfo price_info = 2;
}
```

In the short term, pixels_processed will just track output pixels, but in the future it could track input pixels as well.

Add a receipt field to the existing TranscodedSegmentData message.

```
message TranscodedSegmentData {

    // URL where the transcoded data can be downloaded from.
    string url = 1;
    
    // Transcoding receipt
    TranscodeReceipt receipt = 2;
}
```

### LINKED ISSUE #896: O processes transcoded result and debits from B's credit balance
**TranscodeSeg(segTranscodingMetadata, seg)**

Should return the transcoded results for `seg` based on the requested renditions included in `segTranscodingMetadata` and additional metadata for each of the transcoded results. The transcoded results will be returned as an array `transRes` and the associated metadata will be returned as an array `transResMetadata` such that `transResMetadata[i]` is the metadata associated with `transRes[i]`. In the context of payment accounting, the particular field in the metadata that we care about is the output pixel count for a particular transcoded result.

The number of output pixels processed should be included in a `TranscodeReceipt` message that is returned to the broadcaster.

**ProcessTranscodedResults(addr, transResMetadata)**

Should debit a broadcaster's credit balance based on the number of output pixels included in the transcoded results. 

This will be called after `TranscodeSeg` and before the results are sent to the broadcaster.

```
priceInfo = PriceInfo(addr)
pricePerPixel = priceInfo.pricePerUnit / priceInfo.pixelsPerUnit

debit = 0

for meta in transResMetadata {
    debit += pricePerPixel * meta.outPixels
}

currCredit = CurrCredit(payment.sender)
UpdateCredit(addr, updatedCredit)
```

In the future, if transcoding is short-circuited due to insufficient broadcaster credit balance, the orchestrator would likely need to withhold the result because the produced result would considered an invalid output. However, in the short term without short-circuiting, it might be fine to return the result even if the broadcaster credit balance drops below 0 to give the broadcaster an opportunity to make up for the amount owed with a subsequent payment. If the next segment the broadcaster sends doesn't include a payment that brings the broadcaster's credit balance back up to the minimum credit requirement then the orchestrator will return an error.

kyriediculous:
Debits a broadcaster's credit balance for a manifestID based on the number of output pixels included in the transcoded results, based on the requested renditions, and returns the number of output pixels for each rendition to the broadcaster.

- Added output pixels field `pixels` to `TranscodedSegmentData`protobuf message
- Added an orchestrator interface method `orch.DebitFees` that debits the balance for a manifestID based on O's price per pixel and amount of processed pixels 
- Changed `TranscodeResult` in core/orchestrator to contain a slice of `TranscodeData`, containing a byte slice of the transcoded segment as well as the number of output pixels instead of a two-dimensional byte slice of all transcoded segments. 

Added unit tests for `orch.DebitFees` logic
Added unit tests for `ServeSegment` to make sure `orch.DebitFees` is properly called

Fixes #892 
Fixes #896

**Checklist:**


core/orchestrator.go
```diff
@@ -208,6 +208,15 @@ func (orch *orchestrator) SufficientBalance(manifestID ManifestID) bool {
 	return true
 }
 
+func (orch *orchestrator) DebitFees(manifestID ManifestID, price *net.PriceInfo, pixels int64) {
+	// Don't debit in offchain mode
+	if orch.node == nil || orch.node.Recipient == nil || orch.node.Balances == nil {
```
yondonfu: We should be able to remove the `orch.node.Recipient == nil` check since we don't use `orch.node.Recipient` anywhere within this method

core/orchestrator.go
```diff
@@ -208,6 +208,15 @@ func (orch *orchestrator) SufficientBalance(manifestID ManifestID) bool {
...
+		return
+	}
+	amount := new(big.Rat).Mul(big.NewRat(price.GetPricePerUnit(), price.GetPixelsPerUnit()), big.NewRat(pixels, 1))
```
yondonfu: I think its worth adding an additional guard clause that checks that `price.GetPixelsPerUnit() > 0` because if someone sends a `Payment` that fulfills any of the below conditions:

- `payment.GetExpectedPrice() = nil`
- `payment.GetExpectedPrice().GetPixelsPerUnit() = 0`

the node will panic with a division by zero error.

yondonfu: I think we can reduce the number of `big.Rat` allocations by reusing the `big.Rat` that stores the price and just updating it (since we don't need to reference it outside of the fee calculation):

```
price := big.NewRat(price.GetPricePerUnit(), price.GetPixelsPerUnit())
orch.node.Balances.Debit(manifestID, price.Mul(price, big.NewRat(pixels, 1))
```

core/orch_test.go
```diff
@@ -983,6 +983,68 @@ func TestPriceInfo_TxMultiplierError_ReturnsError(t *testing.T) {
 	assert.EqualError(t, err, expError.Error())
 }
 
+func TestDebitFees(t *testing.T) {
+	n, _ := NewLivepeerNode(nil, "", nil)
+	n.Balances = NewBalances(5 * time.Second)
+	recipient := new(pm.MockRecipient)
```
yondonfu: Shouldn't need a mock recipient for this test

server/segment_rpc_test.go
```diff
@@ -469,6 +477,261 @@ func TestServeSegment_InsufficientBalanceError(t *testing.T) {
...
+		Data:   []byte("foo"),
+		Pixels: int64(110592000),
+	}
+	tRes := &core.TranscodeResult{
+		TranscodeData: []*core.TranscodeData{tData},
+		Sig:           []byte("foo"),
+		OS:            drivers.NewMemoryDriver(nil).NewSession(""),
+	}
+	orch.On("TranscodeSeg", md, seg).Return(tRes, nil)
+	orch.On("DebitFees", md.ManifestID, mock.Anything, mock.Anything)
```
yondonfu: I think we can replace the `mock.Anything` arguments with the actual values that we know the method will be called with which would ensure that we're checking that `DebitFees()` is called with the correct values for `price net.PriceInfo` and `pixels int64`

server/segment_rpc_test.go
```diff
@@ -469,6 +477,261 @@ func TestServeSegment_InsufficientBalanceError(t *testing.T) {
...
+
+	assert := assert.New(t)
+	assert.Equal(http.StatusOK, resp.StatusCode)
+
+	res, ok := tr.Result.(*net.TranscodeResult_Data)
+	assert.True(ok)
+	assert.Equal([]byte("foo"), res.Data.Sig)
+	assert.Equal(1, len(res.Data.Segments))
+	assert.Equal(res.Data.Segments[0].Pixels, tData.Pixels)
+	orch.AssertCalled(t, "DebitFees", md.ManifestID, mock.Anything, tData.Pixels)
```
yondonfu, line 545: I think we can replace `mock.Anything` with the actual `net.PriceInfo` we expect `DebitFees()` to be called with

yondonfu (changes requested):

core/orchestrator.go
kyriediculous: I'm wondering under what conditions this would be possible (once we actually populate PriceInfo in OrchestratorInfo 😄 ) , considering we are enforcing the checks on O's BasePrice to not be those values when starting the node? 

It would require the `txCostOverhead` to be 0, which is impossible as well since that is calculated as `1+ 1/txCostMultiplier`  ? 
`new(big.Rat).Mul(orchBasePrice, txCostOverhead)`

And I don't think `txCostMultiplier` can be 0 either to cause a panic calculating the overhead unless the `gasPrice` is 0 AND the `ticketEV` is set to 0 ?

yondonfu: > I'm wondering under what conditions this would be possible

Ah yeah I think we can skip the guard clause here if we implement checks for `payment.GetExpectedPrice()` in #955 which will likely involve rejection of the payment if the expected price is invalid. Will be good to keep note of this while implementing that task though

server/segment_rpc_test.go
kyriediculous: You're right and I tried that, but it was not working and I didn't know why. 

Now I do,  this is because of #980 not being implemented.

```
--- Expected
+++ Actual
@@ -1,2 +1,2 @@
-(*net.PriceInfo)(pricePerUnit:2 pixelsPerUnit:3 )
+(*net.PriceInfo)(<nil>)
```

So we'll have to merge that and then I can rebase this PR of off master and implement actual values. 

Or can I rebase from that remote branch and not necessarily master?

core/orchestrator.go
kyriediculous: Actually price is a `*net.PriceInfo` type so we have to assign a new variable or pass it inline to the `Debit` call. 

I prefer not having excessive line lengths.

server/segment_rpc_test.go
yondonfu: Either should be fine

core/orchestrator.go
yondonfu:
```
priceRat := big.NewRat(price.GetPricePerUnit(), price.GetPixelsPerUnit())
orch.node.Balances.Debit(manifestID, priceRat.Mul(priceRat, big.NewRat(pixels, 1))
```

should do the trick then.

yondonfu (commented):
Needs a rebase (to squash fixup commits and resolve conflicts with master)!

yondonfu (approved):
LGTM pending all green CI

PR #981, opened 2019-07-12
Broadcaster multi-ticket payment generation

### LINKED ISSUE #897: B generates a payment that will ensure that minimum credit requirement is met to include with segment

### LINKED ISSUE #898: B synchronously processes transcoded results to determine its provisional current credit balance
**ProcessTranscodedResults(sess, transRes)**

Should compute the output pixels in each transcoded result from the array `transRes` in order to calculate the amount that should be debited from the broadcaster's credit balance. A broadcaster will call this function synchronously after receiving the transcoded results for a segment.

```
debit = 0

for res in transRes {
    pricePerPixel = res.receipt.pricePerUnit / res.receipt.pixelsPerUnit
    debit += pricePerPixel * res.receipt.pixelsProcessed
}

currCredit = CurrCredit(self)
UpdateCredit(self, updatedCredit)
```

yondonfu:

This PR implements multi-ticket payment generation for Bs based on their available credit and the minimum credit requirement of O.

- ea5be6a: Add method for creating a batch of tickets. `pm.TicketBatch` mostly follows the same structure as the `net.Payment` protobuf message which already encodes multiple tickets
- cb8f592: Removed the existing `CreateTicket()` method from `pm.Sender` because `CreateTicketBatch()` can be used in its place with `size = 1`
- e11e809: Added a helper `EV()` method to `pm.Sender` to make it easier to calculate the required credit when creating a payment during segment submission
- d1f6333: Wrap the concurrency safe `core.Balances` struct in a `core.Balance` struct that is specific to a session and also exposes a method to reserve existing credit and compute the required # of tickets
- fdd7487: Glue everything together!

The workflow implemented in this PR largely follows [this comment](https://github.com/livepeer/go-livepeer/issues/897#issuecomment-510327432) and [this comment](https://github.com/livepeer/go-livepeer/issues/897#issuecomment-510556017).

Note: There are some commented out code snippets in `server/segment_rpc.go`. I started sketching out what the remainder of the balance update workflow will look like, but I elected to leave out the rest of the implementation/tests from this PR. The completion of the balance update workflow will be a part of the PR to fix #898 

Added unit tests.

Fixes #897 

**Checklist:**


core/accounting.go
```diff
@@ -6,6 +6,51 @@ import (
 	"time"
 )
 
+// Balance holds the credit balance for a stream
```
kyriediculous: Since we are using this part of the accounting logic exclusively for B's I'd perhaps use "for a broadcast session" rather than "for a stream". 

A B could also send segments to multiple O's and then the semantics of "broadcast session" reflect the meaning of a manifestID and a specific O.

core/accounting.go
```diff
@@ -6,6 +6,51 @@ import (
...
+	// and the total payment value is 0
+	if existingCredit.Cmp(minCredit) >= 0 {
+		return 0, big.NewRat(0, 1), existingCredit
+	}
+
+	creditGap := new(big.Rat).Sub(minCredit, existingCredit)
+	sizeRat := creditGap.Quo(creditGap, ev)
+	sizeRatNum := sizeRat.Num()
+	res := sizeRatNum.Div(sizeRatNum, sizeRat.Denom())
+	if !sizeRat.IsInt() {
```
kyriediculous, line 42: If I'm not mistaken `big.Rat.Quo()` already resolves to the smallest possible denominator, checking for `big.rat.IsInt()` would be the same as `big.Rat.Denom()==1 ` . So when true we could save ourselves the division operation. 

The current logic is fine however so I'd also be ok with keeping it as is. 

```
	sizeRat := creditGap.Quo(creditGap, ev)
	res := sizeRat.Num()
	if !sizeRat.IsInt() {
		res = res.Add(res.Div(res, sizeRat.Denom()), big.NewInt(1))
	}
	size := res.Int64()
```

core/accounting.go
```diff
@@ -50,6 +95,20 @@ func (b *Balances) Debit(id ManifestID, amount *big.Rat) {
...
+// Reserve zeros the balance for a ManifestID and returns the current balance
+func (b *Balances) Reserve(id ManifestID) *big.Rat {
+	b.mtx.Lock()
+	defer b.mtx.Unlock()
+	if b.balances[id] == nil {
+		b.balances[id] = &balance{amount: big.NewRat(0, 1)}
+	}
+
+	amount := b.balances[id].amount
+	b.balances[id].amount = big.NewRat(0, 1)
```
kyriediculous, line 111: Since `Balances.Credit` and `Balances.Debit `will initiate non-existing keys I feel like we could probably use `delete` here as well?

core/accounting.go
```diff
@@ -6,6 +6,51 @@ import (
...
+}
+
+// Credit adds an amount to the balance
+func (b *Balance) Credit(amount *big.Rat) {
+	b.balances.Credit(b.manifestID, amount)
+}
+
+// StageUpdate prepares a balance update by reserving the current balance and returning the number of tickets
+// to send with a payment, the new credit represented by the payment and the existing credit (i.e reserved balance)
+func (b *Balance) StageUpdate(minCredit *big.Rat, ev *big.Rat) (int, *big.Rat, *big.Rat) {
```
kyriediculous: Just a minor detail but we can shorten this to `StageUpdate(minCredit, ev *big.Rat)`

core/accounting_test.go
```diff
@@ -8,6 +8,78 @@ import (
...
+func TestBalance_StageUpdate(t *testing.T) {
+	mid := ManifestID("some manifestID")
+	balances := NewBalances(5 * time.Second)
+	b := NewBalance(mid, balances)
+
+	assert := assert.New(t)
+
+	// Test existing credit > minimum credit
+	b.Credit(big.NewRat(2, 1))
+	numTickets, newCredit, existingCredit := b.StageUpdate(big.NewRat(1, 1), nil)
```
kyriediculous, line 37: I'm surprised that setting the `ev` argument to `nil` doesn't cause `panic`? 
https://play.golang.org/p/OMWo4KyPZKu

server/rpc.go
```diff
@@ -52,6 +52,45 @@ type Broadcaster interface {
 	Sign([]byte) ([]byte, error)
 }
 
+// Balance describes methods for a session's balance maintenance
+type Balance interface {
```
kyriediculous, line 57: While in theory there is nothing wrong with this, it does seem like excessive use of interfaces to me?

I think there's nothing wrong with `server` having a dependency on `core`, the other way around however would be problematic. 

the `core` package contains core business/application logic whereas `server` is more of an 'outer' infrastructure package. This can potentially be swapped out for another server implementation although unlikely.

A better example would be if we were to use something to store our balances in a database. We would not want the `core/accounting` logic to be dependent on a particular storage implementation so we would use an interface to generalize this functionality. The storage implementations themselves however can import the core/accounting package without issue if they need to. 

Long story short, I think it's perfectly fine to use `*core.Balance` as a type for the Balance field on the `BoradcastSession`. Analogous to that we could also define the other structs below in `core/accounting` instead, what do you think?

pm/ticket.go
```diff
@@ -126,3 +185,7 @@ func (t *Ticket) flatten() []byte {
 
 	return buf
 }
+
+func ticketEV(faceValue *big.Int, winProb *big.Int) *big.Rat {
```
kyriediculous, line 189: We could use this in `sender.ValidateTicketParams` as well

server/broadcast.go
```diff
@@ -192,11 +192,16 @@ func selectOrchestrator(n *core.LivepeerNode, params *streamParameters, cpl core
 
 	for _, tinfo := range tinfos {
 		var sessionID string
+		var balance *core.Balance
 
 		if n.Sender != nil {
 			sessionID = n.Sender.StartSession(*pmTicketParams(tinfo.TicketParams))
 		}
 
+		if n.Balances != nil {
```
kyriediculous, line 206: I don't think we're currently initiating balances on the node if it's started up as a broadcaster, unless I missed that change in this PR. Given that fact, `n.Balances` will always be `nil`. We'll have to place the initalization of the balances outside of the check for `orchestrator` flag. 

https://github.com/livepeer/go-livepeer/blob/fdd74879cbe11d792dcadcfbec217ae1bcf2e680/cmd/livepeer/livepeer.go#L401

kyriediculous (changes requested):

core/accounting.go
yondonfu: We're setting the balance to zero here with the expectation that it might become non-zero later on (i.e. crediting positive change back to the balance) so I don't think we want to delete the key. Deleting the key would indicate that the entry is not present in the map at all (since `delete` would set `balance = nil`) as opposed to indicating that the entry is present with a balance of zero. Another example of setting the balance to zero without wanting to delete the key is when a caller invokes `Debit()` such that after subtracting the provided amount the balance becomes zero.

core/accounting_test.go
yondonfu: In this case, `StageUpdate()` is set up to return an error before referencing the `ev` argument (see: https://github.com/livepeer/go-livepeer/pull/981/files#diff-541c0849ae46cbbfdb65ee01dcec77b5R35)

server/rpc.go
yondonfu: I prefer to use an interface here because it makes testing here easier by abstracting the implementation details for balance maintenance away from `BroadcastSession`. If we use a concrete type here, then `BroadcastSession` is made aware of how balances are actually maintained under the hood and these details seep into how we write our tests for `SubmitSegment()` as well.

core/accounting.go
yondonfu: Fixed in bb5b890

core/accounting.go
yondonfu: Fixed in bb5b890

core/accounting.go
yondonfu: Fixed in bb5b890

pm/ticket.go
yondonfu: Fixed in 8cdd39b

server/broadcast.go
yondonfu: Fixed in 19bec65

server/rpc.go
yondonfu: I also considered defining `BalanceUpdate` in `core/accounting.go`, but the type didn't seem to have much relevance to the implementation as it would just be wrapping values returned by `StageUpdate()`. The real utility of the type after being created by `StageUpdate()` is to keep track of the status of an update which changes during the execution of `SubmitSegment()`. So, given that the status of a balance update is defined within `SubmitSegment()`, I felt that it would be ok defining `BalanceUpdate` in `server/rpc.go`

core/accounting.go
kyriediculous: Yeah I entirely agree with that reasoning. 

It's just that both `Credit` and `Debit` make no distinction between `balance` nil and `balance.amount = 0` since they zero the balance for a non-existing key before executing any logic.

So I guess depending on the frequency of expected non-zero updates later on there is a tradeoff here between less `Cleanup` work and having to do a re-allocation when updating deleted key.  Not that it matters that much and it logically checks out more to have a zeroed balance in this case rather than a deleted key.

server/broadcast.go
kyriediculous: One more concern that just came up in my mind.

`cleanupInterval` is orchestrator specific and used to be the SenderMonitor and Balance cleanup interval. 

I'm wondering if there's any danger that if we re-use the same value we are cleaning up balances for `BroadcastSessions` when they should in fact be maintained. Sure removing a session happens when the transcodeLoop times out, but I feel like we should only clean after receiving the httpError since we need to account for latency sending back the response. 

I'm also wondering if the `Cleanup` loop is really needed on the `Broadcaster` side because we could also be removing a `Balance` for a session from the map when the session itself is removed or the stream ends.

yondonfu: Updated the PR so that B will not use the cleanup routine - see the following commits:

- 6b0968c: Move cleanup loop start/stop so that the routine is only run for orchestrators
- 5ef2995: Add a `Clear()` method to `core.Balance`
- b30af1a: Clear a session's balance upon removal

server/broadcast.go
```diff
@@ -192,7 +197,7 @@ func selectOrchestrator(n *core.LivepeerNode, params *streamParameters, cpl core
 
 	for _, tinfo := range tinfos {
 		var sessionID string
-		var balance *core.Balance
+		var balance Balance
```
yondonfu, line 200: For anyone that is wondering why this type declaration changed: when `balance` was declared as the concrete type `*core.Balance` its nil value is different than the nil value for the `Balance` interface. As a result, even if `balance` was nil, you would get false when comparing it to the explicit `nil`. See https://golang.org/doc/faq#nil_error for more details!

kyriediculous: This is only true because we're also using the interface on the `BroadcastSession` type definition right?

kyriediculous:
Changes look good, needs a rebase !

yondonfu:
Rebased!

server/broadcast.go
yondonfu: Yep so if we declared `balance` as `*core.Balance` then the comparison `b.Balance != nil` would evaluate to true if `balance` is nil when we set `b.Balance = balance`

kyriediculous (approved):
Ship it ! 🚀

PR #985, opened 2019-07-16
server: Complete balance update when SubmitSegment() returns

### LINKED ISSUE #898: B synchronously processes transcoded results to determine its provisional current credit balance

yondonfu:

This PR implements balance update completion within `SubmitSegment()` with logic that differs depending on the update's status at the time of completion. There are three possible update statuses:

- `Staged`: This is the initial status of the `BalanceUpdate` type. The update remains in this status until the HTTP request for a segment (and its payment header) is submitted.
- `CreditSpent`: This is the status of the `BalanceUpdate` after the HTTP request for a segment is submitted and before a response from O is received. Note: there is no distinction between a HTTP request with a success code and a failure code - as long as the HTTP request was submitted without an error in the Go code, the `BalanceUpdate` status is `CreditSpent`
- `ReceivedChange`: This is the status of the `BalanceUpdate` after a response with transcoded results is received from O

- Add the `completeBalanceUpdate()` helper in `server/segment_rpc.go` which executes the logic for completing a balance update depending on the update's status
- Defer the execution of `completeBalanceUpdate()` in `SubmitSegment()` to guarantee its execution before the function returns
- Change the status of `BalanceUpdate` created when preparing a payment at different points of the execution of `SubmitSegment`

Added unit tests.

Fixes #898 

**Checklist:**

### DIFF
--- a/server/segment_rpc.go
+++ b/server/segment_rpc.go
@@ -349,7 +349,7 @@ func SubmitSegment(sess *BroadcastSession, seg *stream.HLSSegment, nonce uint64)
 
 	// update OrchestratorInfo if necessary
 	if tr.Info != nil {
-		updateOrchestratorInfo(sess, tr.Info)
+		defer updateOrchestratorInfo(sess, tr.Info)
 	}
 
 	// check for errors and exit early if there's anything unusual
--- a/server/segment_rpc.go
+++ b/server/segment_rpc.go
@@ -264,11 +264,10 @@ func SubmitSegment(sess *BroadcastSession, seg *stream.HLSSegment, nonce uint64)
 		return nil, err
 	}
 
-	// TODO(yondonfu):
-	// // The balance update should be completed when this function returns
-	// // The logic of balance update completion depends on the status of the update
-	// // at the time of completion
-	// defer completeBalanceUpdate(sess, balUpdate)
+	// The balance update should be completed when this function returns
+	// The logic of balance update completion depends on the status of the update
+	// at the time of completion
+	defer completeBalanceUpdate(sess, balUpdate)
 
 	payment, err := genPayment(sess, balUpdate.NumTickets)
 	if err != nil {
@@ -307,10 +306,9 @@ func SubmitSegment(sess *BroadcastSession, seg *stream.HLSSegment, nonce uint64)
 	}
 	defer resp.Body.Close()
 
-	// TODO(yondonfu):
 	// If the segment was submitted then we assume that any payment included was
 	// submitted as well so we consider the update's credit as spent
-	// balUpdate.Status = CreditSpent
+	balUpdate.Status = CreditSpent
 
 	if resp.StatusCode != 200 {
 		data, _ := ioutil.ReadAll(resp.Body)
@@ -386,9 +384,19 @@ func SubmitSegment(sess *BroadcastSession, seg *stream.HLSSegment, nonce uint64)
 		return nil, err
 	}
 
-	// TODO(yondonfu):
-	// Set balUpdate.Debit = fee computed based on returned results
-	// Set balUpdate.Status = ReceivedChange
+	// We treat a response as "receiving change" where the change is the difference between the credit and debit for the update
+	balUpdate.Status = ReceivedChange
+	priceInfo := sess.OrchestratorInfo.PriceInfo
+	if priceInfo != nil {
+		// The update's debit is the transcoding fee which is computed as the total number of pixels processed
+		// for all results returned multiplied by the orchestrator's price
+		var pixelCount int64
+		for _, res := range tdata.Segments {
+			pixelCount += res.Pixels
+		}
+
+		balUpdate.Debit.Mul(new(big.Rat).SetInt64(pixelCount), big.NewRat(priceInfo.PricePerUnit, priceInfo.PixelsPerUnit))
+	}
 
 	// transcode succeeded; continue processing response
 	if monitor.Enabled {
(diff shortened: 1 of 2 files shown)


server/segment_rpc.go
```diff
@@ -386,9 +384,19 @@ func SubmitSegment(sess *BroadcastSession, seg *stream.HLSSegment, nonce uint64)
 		return nil, err
 	}
 
-	// TODO(yondonfu):
-	// Set balUpdate.Debit = fee computed based on returned results
-	// Set balUpdate.Status = ReceivedChange
+	// We treat a response as "receiving change" where the change is the difference between the credit and debit for the update
+	balUpdate.Status = ReceivedChange
+	priceInfo := sess.OrchestratorInfo.PriceInfo
```
kyriediculous, line 389: Since we're fetching the priceInfo from the session after calling `updateOrchestratorInfo`, we'll be overwriting any old PriceInfo on which B's expected price was based and we're using B's expected price (if acceptable) to Debit the balance from the orchestrator side. 

This could result in a discrepancy between the actual debit amounts. 

https://github.com/livepeer/go-livepeer/blob/40c37f7264f4f8ac980cd24e17d15b3a98571fd9/server/segment_rpc.go#L152

server/segment_rpc.go
```diff
@@ -386,9 +384,19 @@ func SubmitSegment(sess *BroadcastSession, seg *stream.HLSSegment, nonce uint64)
...
+	if priceInfo != nil {
```
kyriediculous, line 390: I don't think priceInfo can ever be `nil` here, can it? 

1. It would mean expected price is nil , which would trigger an update from the orchestrator and we're using the value after update here (which I don't think we should, see previous comment).
2. it's not possible for O to set a nil priceInfo on node startup, meaning when O is picked up from orchestratorPool and its info is requested it is never nil on the session?

server/segment_rpc.go
```diff
@@ -477,31 +485,31 @@ func newBalanceUpdate(sess *BroadcastSession) (*BalanceUpdate, error) {
...
+	// If the update's credit has not been spent then add the existing credit
+	// back to the balance
+	if update.Status == Staged {
+		sess.Balance.Credit(update.ExistingCredit)
+		return
+	}
+
+	// If the update did not include a processed debit then no change was received
+	// so we exit without updating the balance because the credit was spent already
+	if update.Status != ReceivedChange {
```
kyriediculous, line 502: Doesn't this essentially zero the entire balance ? Why is there no residual balance possible at this point? Thanks.

kyriediculous (changes requested):

server/segment_rpc.go
yondonfu: Ah good catch. I think we can fix this by deferring execution of `updateOrchestratorInfo()` until the very end up of `SubmitSegment()` since the only current requirement is that the session's `OrchestratorInfo` is updated when `SubmitSegment()` returns

server/segment_rpc.go
yondonfu: `priceInfo` would be `nil` when the node is running in off-chain mode because Os don't include a price in the `OrchestratorInfo` message in off-chain mode.

kyriediculous: Right, I tend to always forget about offchain mode existing😅 . Thanks !

server/segment_rpc.go
yondonfu: Yeah this would zero the entire balance. The motivation behind this is that if the segment is submitted (i.e. the `update.Status = CreditSpent`), but B does not receive a successful response before `SubmitSegment()` errors out then B does not know with certainty whether O successfully received the payment and applied credit/debit to its view of B's balance. If we instead add `existingCredit` back to B's balance then it is possible that O did receive the payment and applied credit/debit such that its view of B's balance is `< existingCredit`. Then, if B reuses the session with O, B would receive an insufficient balance error from O because B generates a payment based on an out of sync view of its balance. Considering the existing credit as "spent" in this scenario makes the workflow between B & O less dependent on the two parties having the same exact view of B's balance.

I actually just realized that technically we already zero the balance when `SubmitSegment()` encounters an error because we call [removeSession()](https://github.com/livepeer/go-livepeer/blob/master/server/broadcast.go#L92) which will clear the balance for the session. However, `removeSession()` will actually delete the manifestID key in the underlying `core.Balances` type. Meanwhile, if `update.Status = CreditSpent` when we complete the balance update in `SubmitSegment()`, the balance stored for the manifestID in the underlying `core.Balances` type will be left as `big.NewRat(0, 1)` (and then the entire map entry would be deleted when we call `removeSession()`). I'm ok with leaving this behavior as-is since `removeSession()` actually deletes the keys from `core.Balances`

server/segment_rpc.go
yondonfu: Fixed in 9979bf4

kyriediculous:
Changes look good to me. 
Needs a rebase !

yondonfu:
Rebased - decided to retain two distinct commits for this PR since I think having the second commit as a standalone change might be potentially useful to have in the history (i.e. this change ensures that price and other O info updates for a session happen at the very end of `SubmitSegment()`). I did fix a typo in the commit message though.

kyriediculous (approved):
That sounds reasonable to me, Good to go 🚀 . 
The gaspricemonitor polling test is flaky again on circle though. 
Can we re-run it for good measure before merging?

PR #988, opened 2019-07-16
pm: Validate params for multiple tickets in CreateTicketBatch()

### LINKED ISSUE #982: B should not send a payment if its total value is too high

After #898 is implemented, B could have a negative balance at the time of payment generation which could result in B generating a multi-ticket payment in order to meet O's minimum credit requirement (ticket EV). Since B updates its balance provisionally based on the pixel count reported by O, if O misreports a high pixel count, the higher the pixel count, the lower B's balance and the more tickets that need to be included in a payment.

B should be able to cap the value represented by a payment. B already checks whether the ticket EV advertised by O is acceptable based on a configured maxEV. However, this value check is based on just a single ticket. We can make this value check based on N tickets by having B check that for a batch of N tickets, the total value `N * EV` is less than or equal to `maxEV`.

yondonfu:
Refactor ValidateTicketParams() to use a helper that accepts a
numTickets params

This PR implements the ability for a `pm.Sender` to validate ticket params for sending multiple tickets in `CreateTicketBatch()`. Previously, ticket param validation consisted of checking if the EV and faceValue for a single ticket were too high. Now, ticket param validation can check if the total EV and faceValue for multiple tickets are too high.

- Refactor `ValidateTicketParams()` to use a helper that accepts a
numTickets param
- Use the aforementioned helper in `CreateTicketBatch()` with the size of the batch as the value for the `numTickets` param

Added unit tests.

Fixes #982 

**Checklist:**

### DIFF
--- a/pm/sender.go
+++ b/pm/sender.go
@@ -81,7 +81,7 @@ func (s *sender) CreateTicketBatch(sessionID string, size int) (*TicketBatch, er
 		return nil, err
 	}
 
-	if err := s.ValidateTicketParams(&session.ticketParams); err != nil {
+	if err := s.validateTicketParams(&session.ticketParams, size); err != nil {
 		return nil, err
 	}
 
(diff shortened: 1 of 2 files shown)


kyriediculous (approved):
Looks good to me 💯 

Just one thought (just so that I can say something at least 😄 )

I'm okay with keeping the exported Validate method that does NOT take a size argument as long as we do not have to export the one that does. 

From the moment we do have to export the new validation function that takes size as a parameter I am in favor of removing the old exported method because that is only used in the `DBOrchestratorPool.GetOrchestrators` predicate function anyway.

yondonfu:
> From the moment we do have to export the new validation function that takes size as a parameter I am in favor of removing the old exported method because that is only used in the DBOrchestratorPool.GetOrchestrators predicate function anyway.

Yep if we ever have to validate ticket params for sending multiple tickets anywhere else besides `CreateTicketBatch()`, then we can consider removing `ValidateTicketParams(ticketParams)` in favor of `ValidateTicketParams(ticketParams, numTIckets)`.

PR #991, opened 2019-07-17
cmd, core, pm: O validates B's expected price for a payment

### LINKED ISSUE #955: O should validate pricePerPixel sent by B

kyriediculous:
This PR allows an orchestrator to check whether a broadcaster's `payment.ExpectedPrice` matches the dynamically generated `PriceInfo` broadcaster based on the orchestrator's `basePrice` and transaction cost overhead. 

If the `payment.ExpectedPrice` is outdated (doesn't match) this will count as an acceptable error , similar to acceptable errors for pm tickets, as long as the broadcaster is still within the grace period (hasn't reached the maximum error count). 

The orchestrator uses a shared  error counter defined on the `LivepeerNode`  for both acceptable price errors and acceptable pm ticket errors. 

- Created (extracted out of `SenderMonitor`)  an `ErrorMonitor` that tracks both acceptable pm ticket errors as well as acceptable price errors.
- Added a price check in `orch.ProcessPayment`
- Added an unexported helper method on `orchestrator` to check expected price vs the orchestrator's price for a sender
- Added an `ErrorMonitor` interface in the `pm` package
- Added a `stubErrorMonitor` to `pm` stubs/mocks

Added unit tests

Fixes #955 

**Checklist:**

### DIFF
--- a/core/errormonitor.go
+++ b/core/errormonitor.go
@@ -59,3 +59,34 @@ func (em *errorMonitor) StartGasPriceUpdateLoop() {
 		em.resetErrCounts()
 	}
 }
+
+// AcceptableError is an interface that describes methods for a payment related error that
+// may be acceptable depending on the type of underlying error
+type AcceptableError interface {
+	error
+
+	// Acceptable returns whether the error is acceptable
+	Acceptable() bool
+}
+
+type acceptableError struct {
+	err        error
+	acceptable bool
+}
+
+func newAcceptableError(err error, acceptable bool) *acceptableError {
+	return &acceptableError{
+		err:        err,
+		acceptable: acceptable,
+	}
+}
+
+// Error returns the underlying error as a string
+func (re *acceptableError) Error() string {
+	return re.err.Error()
+}
+
+// Acceptable returns whether the error is acceptable
+func (re *acceptableError) Acceptable() bool {
+	return re.acceptable
+}
--- a/core/errormonitor.go
+++ b/core/errormonitor.go
@@ -7,18 +7,24 @@ import (
 )
 
 type errorMonitor struct {
-	mu          sync.RWMutex
-	maxErrCount int
-	errCount    map[ethcommon.Address]int
+	mu             sync.Mutex
+	maxErrCount    int
+	errCount       map[ethcommon.Address]int
+	gasPriceUpdate chan struct{}
 }
 
-func NewErrorMonitor(maxErrCount int) *errorMonitor {
+// NewErrorMonitor returns a new errorMonitor instance
+func NewErrorMonitor(maxErrCount int, gasPriceUpdate chan struct{}) *errorMonitor {
 	return &errorMonitor{
-		maxErrCount: maxErrCount,
-		errCount:    make(map[ethcommon.Address]int),
+		maxErrCount:    maxErrCount,
+		errCount:       make(map[ethcommon.Address]int),
+		gasPriceUpdate: gasPriceUpdate,
 	}
 }
 
+// AcceptErr checks if a sender has reached the max error count
+// returns false if no more errors can be accepted
+// returns true and increments the error count when smaller than the max error count
 func (em *errorMonitor) AcceptErr(sender ethcommon.Address) bool {
 	em.mu.Lock()
 	defer em.mu.Unlock()
@@ -30,14 +36,26 @@ func (em *errorMonitor) AcceptErr(sender ethcommon.Address) bool {
 	return true
 }
 
+// ClearErrCount zeroes the error count for a sender
 func (em *errorMonitor) ClearErrCount(sender ethcommon.Address) {
 	em.mu.Lock()
 	defer em.mu.Unlock()
 	em.errCount[sender] = 0
 }
 
-func (em *errorMonitor) ResetErrCounts() {
-	for s := range em.errCount {
-		em.ClearErrCount(s)
+// ResetErrCounts clears error counts for all senders
+func (em *errorMonitor) resetErrCounts() {
+	em.mu.Lock()
+	defer em.mu.Unlock()
+	// Init a fresh map
+	em.errCount = make(map[ethcommon.Address]int)
+}
+
+// StartGasPriceUpdateLoop initiates a loop that runs a worker
+// to reset the errCount for senders every time a gas price change
+// notification is received
+func (em *errorMonitor) StartGasPriceUpdateLoop() {
+	for range em.gasPriceUpdate {
+		em.resetErrCounts()
 	}
 }
--- a/core/errormonitor.go
+++ b/core/errormonitor.go
@@ -1,0 +1,43 @@
+package core
+
+import (
+	"sync"
+
+	ethcommon "github.com/ethereum/go-ethereum/common"
+)
+
+type errorMonitor struct {
+	mu          sync.RWMutex
+	maxErrCount int
+	errCount    map[ethcommon.Address]int
+}
+
+func NewErrorMonitor(maxErrCount int) *errorMonitor {
+	return &errorMonitor{
+		maxErrCount: maxErrCount,
+		errCount:    make(map[ethcommon.Address]int),
+	}
+}
+
+func (em *errorMonitor) AcceptErr(sender ethcommon.Address) bool {
+	em.mu.Lock()
+	defer em.mu.Unlock()
+
+	if em.errCount[sender] >= em.maxErrCount {
+		return false
+	}
+	em.errCount[sender]++
+	return true
+}
+
+func (em *errorMonitor) ClearErrCount(sender ethcommon.Address) {
+	em.mu.Lock()
+	defer em.mu.Unlock()
+	em.errCount[sender] = 0
+}
+
+func (em *errorMonitor) ResetErrCounts() {
+	for s := range em.errCount {
+		em.ClearErrCount(s)
+	}
+}
--- a/core/orchestrator.go
+++ b/core/orchestrator.go
@@ -108,13 +108,21 @@ func (orch *orchestrator) ProcessPayment(payment net.Payment, manifestID Manifes
 	}
 
 	var (
-		acceptablePriceErr error
-		okPrice            bool
-		didReceiveErr      bool
+		didPriceErr            bool
+		acceptablePrice        bool
+		unacceptableReceiveErr bool
+		didReceiveErr          bool
 	)
 
-	if okPrice, acceptablePriceErr = orch.acceptablePrice(ethcommon.BytesToAddress(payment.Sender), payment.GetExpectedPrice()); !okPrice {
-		return acceptablePriceErr
+	err := orch.acceptablePrice(ethcommon.BytesToAddress(payment.Sender), payment.GetExpectedPrice())
+	if err != nil {
+		glog.Error(err)
+		didPriceErr = true
+	}
+
+	acceptablePriceErr, ok := err.(AcceptableError)
+	if err == nil || (ok && acceptablePriceErr.Acceptable()) {
+		acceptablePrice = true
 	}
 
 	seed := new(big.Int).SetBytes(payment.TicketParams.Seed)
@@ -146,10 +154,12 @@ func (orch *orchestrator) ProcessPayment(payment net.Payment, manifestID Manifes
 			didReceiveErr = true
 		}
 
-		pmErr, ok := err.(pm.Error)
-		if err == nil || (ok && pmErr.Acceptable()) {
+		pmErr, ok := err.(AcceptableError)
+		if acceptablePrice && err == nil || (ok && pmErr.Acceptable()) {
 			// Add ticket EV to credit
 			orch.node.Balances.Credit(manifestID, ticket.EV())
+		} else {
+			unacceptableReceiveErr = true
 		}
 
 		if won {
@@ -163,12 +173,18 @@ func (orch *orchestrator) ProcessPayment(payment net.Payment, manifestID Manifes
 		}
 	}
 
-	if acceptablePriceErr != nil {
-		return acceptablePriceErr
+	if didPriceErr {
+		return newAcceptableError(
+			fmt.Errorf("expected price did not match orchestrator price"),
+			acceptablePrice,
+		)
 	}
 
 	if didReceiveErr {
-		return fmt.Errorf("error receiving tickets with payment")
+		return newAcceptableError(
+			fmt.Errorf("error receiving tickets with payment"),
+			!unacceptableReceiveErr,
+		)
 	}
 
 	return nil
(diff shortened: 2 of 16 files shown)


core/errormonitor.go
```diff
@@ -0,0 +1,72 @@
...
+)
+
+type errorMonitor struct {
+	mu             sync.RWMutex
+	maxErrCount    int
+	errCount       map[ethcommon.Address]int
+	gasPriceUpdate chan struct{}
+}
+
+// NewErrorMonitor returns a enw errorMonitor instance
```
yondonfu: "enw" -> "new"

core/errormonitor.go
```diff
@@ -0,0 +1,72 @@
...
+
+// ClearErrCount zeroes the error count for a sender
+func (em *errorMonitor) ClearErrCount(sender ethcommon.Address) {
+	em.mu.Lock()
+	defer em.mu.Unlock()
+	em.errCount[sender] = 0
+}
+
+// ResetErrCounts clears error counts for all senders
+func (em *errorMonitor) ResetErrCounts() {
```
yondonfu: I don't think this needs to be exported

core/errormonitor.go
```diff
@@ -0,0 +1,72 @@
+package core
+
+import (
+	"sync"
+
+	ethcommon "github.com/ethereum/go-ethereum/common"
+)
+
+type errorMonitor struct {
+	mu             sync.RWMutex
```
yondonfu: Doesn't look we exclusively read lock/unlock anywhere so I think this can be `sync.Mutex` instead

core/errormonitor.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+// Start creates a new worker to watch for gasPrice updates
+func (em *errorMonitor) Start() {
+	go em.startGasPriceUpdateLoop()
+}
+
+// Stop clears all errors and stops the gasPrice update watcher
+func (em *errorMonitor) Stop() {
+	close(em.gasPriceUpdate)
```
yondonfu: Currently, the `gasPriceUpdate` channel is actually created by `eth.GasPriceMonitor` and then passed into `errorMonitor` so if `errorMonitor` closes the channel and for some reason `eth.GasPriceMonitor` is still running a send into the closed `gasPriceUpdate` channel would panic. Furthermore, generally, the sender of a channel should be the one to close the channel to signal to a receiver that there are no more values to receive. So, I don't think `errorMonitor` should close the channel and instead we should update `eth.GasPriceMonitor` to close the channel when it is stopped.

core/errormonitor.go
```diff
@@ -0,0 +1,72 @@
...
+	defer em.mu.Unlock()
+	// Init a fresh map
+	em.errCount = make(map[ethcommon.Address]int)
+}
+
+// startGasPriceUpdateLoop initiates a loop that runs a worker
+// to reset the errCount for senders every time a gas price change
+// notification is received
+func (em *errorMonitor) startGasPriceUpdateLoop() {
+	defer em.ResetErrCounts()
```
yondonfu: Why do we need this deferred `em.ResetErrCounts()`? IIUC, this call would reset the error counts when `em.gasPriceUpdate` is closed, but we only need to reset the error counts when we receive an actual update from the channel. If `em.gasPriceUpdate` is closed because `em.GasPriceMonitor` is stopped (likely because the program is exiting), then any remaining values in `em.errCount` won't be retained anyway.

core/orchestrator.go
```diff
@@ -222,6 +234,23 @@ func (orch *orchestrator) DebitFees(manifestID ManifestID, price *net.PriceInfo,
 	orch.node.Balances.Debit(manifestID, priceRat.Mul(priceRat, big.NewRat(pixels, 1)))
 }
 
+// Acceptable price checks whether the payment sender's expected price sent with a payment is acceptable
+func (orch *orchestrator) acceptablePrice(sender ethcommon.Address, ep *net.PriceInfo) (bool, error) {
+	epRat := big.NewRat(ep.GetPricePerUnit(), ep.GetPixelsPerUnit())
```
yondonfu, line 257: Should check that `ep.GetPixelsPerUnit() > 0` since `ep` is supplied by B and it could be the case that `ep = nil` or `ep.PixelsPerUnit = 0`

yondonfu (changes requested):
Left a few comments as a part of a first pass review. Before taking a second pass I wanted to highlight the following:

Currently, this PR implements the acceptable price check in `ProcessPayment()` - if the error is acceptable we proceed and process the tickets in the payment, else we return an error from `ProcessPayment()` and do not process the tickets in the payment. The problem is that even if `ProcessPayment()` returns an error, at the moment O will continue execution of `ServeSegment()`. An initial thought might be that this is okay because if an invalid price triggers a non-acceptable error then `ProcessPayment()` would not credit B and eventually B's balance will become insufficient preventing any further transcoding from taking place. However, the following attack is still possible:

1. B sends a payment worth EV with a valid price
2. B sends a payment worth 0 (no tickets) with price = 0
3. Repeat step 2 until B exceeds the max error count
4. B sends a payment worth 0 (no tickets) with price = 0

After step 1, at no point does B's balance drop below EV because starting at step 2, B sends a payment with a price = 0 which means no fees are ever debited from B's balance. In step 4, `ProcessPayment()` will return an error since B has exceeded the max error count. However, O will still transcode B's video because B's balance is still sufficient at this point because it never dropped below EV after step 1.

If B exceeds the max error count and the price it includes in the payment is invalid, O should not download/transcode the segment. A few initial thoughts on how to do this:

1. Add an additional `Orchestrator` method for checking the price and if an error is returned, `ServeSegment()` will halt execution
2. Rename `SufficientBalance()` to something like `ValidatePriceAndBalance()` and update the method to not only check for a sufficient balance, but also to check for a valid price. If this method returns an error then `ServeSegment()` will halt execution
3. Add a way to signal from `ProcessPayment()` whether to halt execution or not. This could be accomplished by adding an additional boolean return value or by wrapping the error returned to include a boolean (checking it would be similar to what we do with the `pm.Error` type assertion in `ProcessPayment())`.

My preferred approach of the above options is 2, but perhaps there are better approaches.

kyriediculous:
I favor pulling all acceptable error related types and logic into their own package

**errors:** 
* acceptable.go
* errormonitor.go

We can then check whether the error returned from `ProcessPayment` is actually acceptable because at that point it doesn't matter whether is a receiveError or an expectedPriceError, just whether it is acceptable.

We can also use the error package in core, pm and server without facing cyclic imports. 

Quick prototype: https://github.com/livepeer/go-livepeer/commit/a09aa8387e7cdec0aaccfbe75c3ad2a0150ecaa1

j0sh:
> I favor pulling all acceptable error related types and logic into their own package

Not in favor of this. Not only do we probably have too many packages already - I'd actually prefer fewer, since then it's that much easier to settle the increasingly common question of "where to put this new helper function?". Moreover, having a "PM helper that's not a Livepeer helper, but still used by Livepeer"  really sets an undesirable precedent in fracturing package responsibilities. Some thoughts when we had an earlier discussion about this:

https://github.com/livepeer/go-livepeer/pull/702#discussion_r250301294

In general, I'd prefer a bit of repetition over bending bendwards to stay DRY and accommodate the golang cyclic checker. Or bite the bullet, unwind the strained inversion of control, and have `pm` import `common`.

kyriediculous:
> "PM helper that's not a Livepeer helper, but still used by Livepeer"

Acceptable errors are not specific to the `pm` package anymore though, acceptable ticket receive errors are, but pricing errors can also be acceptable which is pertained to the accounting logic.

While I originally intended for this to go into the `common` package rather than its own, this isn't possible as it imports the `pm` package solely for the ticket struct for db storage. Then again, the SQL specific database logic shouldn't be in the `common` package as a whole either, in my opinion, now that we are going down the package structure rabit hole 🐰. (edit: you mention this further in the discussion as well, my bad didn't get to that far yet when writing the reply)

The original idea of making the `pm` package extractable is a noble one. It is however a client implementation and different projects might have different client-side requirements for the pm protocol, I'm not sure about this. As I understood from discussions with Yondon about this, I thought we had abandoned that route for now. However I would like to keep it as extractable as possible so how about an alternative: we could place `AcceptableErrors` in `core` and keep `ReceiveErrors` in the `pm` package so we do not create that dependency. But since they adhere to the same interface we can still check if a receiveError is acceptable. This is similar to what @yondonfu described as proposal three, albeit a little more formal code-wise. In my opinion the other two proposed solutions circumvent a larger problem at hand, but require less work.

core/orchestrator.go
```diff
@@ -236,7 +236,7 @@ func (orch *orchestrator) DebitFees(manifestID ManifestID, price *net.PriceInfo,
 
 // Acceptable price checks whether the payment sender's expected price sent with a payment is acceptable
 func (orch *orchestrator) acceptablePrice(sender ethcommon.Address, ep *net.PriceInfo) (bool, error) {
-	if ep == nil || ep.GetPixelsPerUnit() < 0 {
+	if ep == nil || ep.GetPixelsPerUnit() <= 0 {
```
yondonfu, line 254: Since we're checking that `ep.GetPixelsPerUnit() <= 0` I think we should also have a test case for when the value is negative (looks like we only have one for when the value is 0 right now)

eth/gaspricemonitor.go
```diff
@@ -112,6 +112,9 @@ func (gpm *GasPriceMonitor) Stop() error {
 	gpm.cancel = nil
 	gpm.polling = false
 
+	// Close the update channel when gpm is stopped
+	close(gpm.update)
```
yondonfu, line 116: Can we add a test that checks for this close?

yondonfu:
I don't think we need to export any error struct types.

> we could place AcceptableErrors in core and keep ReceiveErrors in the pm package so we do not create that dependency. But since they adhere to the same interface we can still check if a receiveError is acceptable.

Yeah we could remove the `pm.Error` interface and define a `core.AcceptableError` interface and have acceptable errors returned by both the `pm` package and the `core` package adhere to this interface.

eth/gaspricemonitor.go
kyriediculous: added in https://github.com/livepeer/go-livepeer/pull/991/commits/b9d2fae0b21dd245bd209c43f55a989f8dd36f34

core/orchestrator.go
kyriediculous: added in https://github.com/livepeer/go-livepeer/pull/991/commits/b9d2fae0b21dd245bd209c43f55a989f8dd36f34

kyriediculous:
So the error handling has been addressed in a few subsequent commits.

In the rebase process I would possibly like to squash following commits but keep the squashed version as a seperate commit in the history:

```
pm, core: server: AcceptableError interface
server, core: process payment acceptable errors
```

core/orchestrator.go
```diff
@@ -109,13 +109,21 @@ func (orch *orchestrator) ProcessPayment(payment net.Payment, manifestID Manifes
...
+		acceptablePrice        bool
+		unacceptableReceiveErr bool
+		didReceiveErr          bool
 	)
 
-	if okPrice, acceptablePriceErr = orch.acceptablePrice(ethcommon.BytesToAddress(payment.Sender), payment.GetExpectedPrice()); !okPrice {
-		return acceptablePriceErr
+	err := orch.acceptablePrice(ethcommon.BytesToAddress(payment.Sender), payment.GetExpectedPrice())
+	if err != nil {
+		glog.Errorf("%s", err.Error())
```
yondonfu: Can shorten to `glog.Error(err)`

core/orchestrator.go
```diff
@@ -109,13 +109,21 @@ func (orch *orchestrator) ProcessPayment(payment net.Payment, manifestID Manifes
...
-		didReceiveErr      bool
+		didPriceErr            bool
+		acceptablePrice        bool
+		unacceptableReceiveErr bool
+		didReceiveErr          bool
 	)
 
-	if okPrice, acceptablePriceErr = orch.acceptablePrice(ethcommon.BytesToAddress(payment.Sender), payment.GetExpectedPrice()); !okPrice {
-		return acceptablePriceErr
+	err := orch.acceptablePrice(ethcommon.BytesToAddress(payment.Sender), payment.GetExpectedPrice())
```
yondonfu, line 117: Can use `if err := ... ; err != nil {}` syntax to make this more concise

pm/stub.go
```diff
@@ -444,6 +444,13 @@ type MockReceiveError struct {
 	acceptable bool
 }
 
+type mockReceiveError interface {
```
yondonfu: Since we usually mock/stub the struct that implements an interface (as opposed to creating a mock/stub interface) and since we already have a type called `MockReceiveError`, could we name this something else? Maybe just `acceptableError`?

server/segment_rpc.go
```diff
@@ -59,24 +59,30 @@ func (h *lphttp) ServeSegment(w http.ResponseWriter, r *http.Request) {
...
-	// to the broadcaster
+	// oInfo will be non-nil if we need to send an updated net.OrchestratorInfo to the broadcaster
 	var oInfo *net.OrchestratorInfo
 
-	if err := orch.ProcessPayment(payment, segData.ManifestID); err != nil {
-		glog.Errorf("Error processing payment: %v", err)
+	if paymentError := orch.ProcessPayment(payment, segData.ManifestID); paymentError != nil {
 
+		acceptableErr, ok := paymentError.(core.AcceptableError)
+		if !ok || !acceptableErr.Acceptable() {
```
yondonfu: Noting that this will result in a HTTP error when `payment.Sender = nil || len(payment.Sender) == 0` and when `payment.TicketParams = nil` which would be the case for empty payment headers. While we don't send empty payments right now (since B usually doesn't credit its balance past the minimum), it might be useful to allow empty payments as long as the balance is sufficient and the price is acceptable (small change to `if ok && !acceptableErr.Acceptable()`)

server/segment_rpc.go
```diff
@@ -59,24 +59,30 @@ func (h *lphttp) ServeSegment(w http.ResponseWriter, r *http.Request) {
...
+			return
+		}
 		oInfo, err = orchestratorInfo(orch, getPaymentSender(payment), orch.ServiceURI().String())
 		if err != nil {
 			glog.Errorf("Error updating orchestrator info: %v", err)
 			http.Error(w, "Internal Server Error", http.StatusInternalServerError)
 			return
 		}
+
+		glog.Errorf("Acceptable error occured when processing payment: %v", paymentError.Error())
```
yondonfu: Can pass in `paymentError` instead of `paymentError.Error()`

server/segment_rpc.go
```diff
@@ -59,24 +59,30 @@ func (h *lphttp) ServeSegment(w http.ResponseWriter, r *http.Request) {
...
 	var oInfo *net.OrchestratorInfo
 
-	if err := orch.ProcessPayment(payment, segData.ManifestID); err != nil {
-		glog.Errorf("Error processing payment: %v", err)
+	if paymentError := orch.ProcessPayment(payment, segData.ManifestID); paymentError != nil {
 
+		acceptableErr, ok := paymentError.(core.AcceptableError)
+		if !ok || !acceptableErr.Acceptable() {
+			glog.Errorf("Unacceptable error occured processing payment: %v", paymentError.Error())
+			http.Error(w, fmt.Sprintf("Internal Server Error: %s", paymentError.Error()), http.StatusInternalServerError)
```
yondonfu: `http.StatusBadRequest` is properly a better fit here (if we make the switch we can also remove the "Internal Server Error" prefix)

server/segment_rpc.go
```diff
@@ -59,24 +59,30 @@ func (h *lphttp) ServeSegment(w http.ResponseWriter, r *http.Request) {
...
 			return
 		}
+
+		glog.Errorf("Acceptable error occured when processing payment: %v", paymentError.Error())
 	}
 
 	if !orch.SufficientBalance(segData.ManifestID) {
 		glog.Errorf("Insufficient credit balance for stream with manifestID %v\n", segData.ManifestID)
-		http.Error(w, "Internal Server Error", http.StatusInternalServerError)
+		http.Error(w, "Internal Server Error: insufficient balance", http.StatusInternalServerError)
```
yondonfu: `http.StatusBadRequest` is properly a better fit here (if we make the switch we can also remove the "Internal Server Error" prefix)

core/orchestrator.go
kyriediculous: I think that makes `err` scoped to that `if` block if I'm not mistaken and we need to access it at line 124, outside of that particular if block.

server/segment_rpc.go
kyriediculous: Fixed in aceab05

server/segment_rpc.go
kyriediculous: Fixed in aceab05

server/segment_rpc.go
kyriediculous: Fixed in aceab05

server/segment_rpc.go
kyriediculous: Fixed in aceab05

pm/stub.go
kyriediculous: Fixed in aceab05

core/orchestrator.go
kyriediculous: Fixed in aceab05

core/orchestrator.go
yondonfu: Ah yes you're right - forgot that we need `err` for the type assertion on L124

yondonfu:
Needs a rebase!

kyriediculous:
Rebased, are you okay with keeping the 4 seperate commits?

yondonfu (approved):
> are you okay with keeping the 4 seperate commits?

👍 

LGTM :shipit:

PR #992, opened 2019-07-17
server: B validates PriceInfo from O

### LINKED ISSUE #929: B should be able to react to PriceInfo updates from O mid-stream and adjust according to its MaxBroadcastPrice
#924 would allow an O to adjust the `PriceInfo` he sends along with `OrchestratorInfo` mid-stream through the CLI. A B would be notified of such an update when a transcoded segment is returned along with `OrchestratorInfo`

B should then be able to check that the updated `PriceInfo` from O, it is still below B's `MaxBroadcastPrice`. 

If the updated price from O exceeds B's preferrence, B should be able to fluently switch to a new O with a time to transcode for the next segment as little as possible. 

Since `PriceInfo` is part of `OrchestratorInfo` which is sent along with `TranscodeResult` we can check for updated pricing upon receiving a transcoded segment.

If `OrchestratorInfo && Orchestrator.PriceInfo` is not `nil` we should update the `PriceInfo` for O in `func updateOrchestratorInfo(sess *BroadcastSession, oInfo *net.OrchestratorInfo)` (`segment_rpc.go`)

If the NEW price exceeds B's `MaxBroadcastPrice` a new O with acceptable pricing should be selected. 

Selection of a new O should be fluent to reduce latency. Therefore we could go through the previously selected O's for B in the `DBCache` rather than using `orchestratorPool.GetOrchestrators` to refresh `OrchestratorInfo` for the entire active transcoder pool as described by Josh's comment in #906. 

Refresh the entire orchestratorPool to get the absolute best price. 

**Additional context**
This issue depends on/is blocked by the completion of #906 and #924

kyriediculous:
Checks whether an Orchestrator's price for a `BroadcastSession` isn't greater than the `MaxBroadcastPrice` when generating a payment in `genPayment`. If this value isn't set then it will be ignored and the broadcaster will accept any price. 

If an orchestrator's price is not acceptable `genPayment` and subsequently `submitSegment` will error out, causing the `BroadcastSession` to be removed from the session list. The broadcaster will then retry the current segment using the next `BroadcastSession` in the session list or select new O's from the `DBcache` if the list has become too short. 

- Created a `validatePrice` helper in `segment_rpc.go`
- Added a price validation check using `validatePrice` in `genPayment`

added & adjusted unit tests

Fixes #929

**Checklist:**

### DIFF
--- a/server/rpc_test.go
+++ b/server/rpc_test.go
@@ -338,6 +338,14 @@ func TestGenPayment(t *testing.T) {
 	sender := &pm.MockSender{}
 	s.Sender = sender
 
+	// Test invalid price
+	BroadcastCfg.SetMaxPrice(big.NewRat(1, 5))
+	payment, err = genPayment(s, 1)
+	assert.Equal("", payment)
+	assert.Errorf(err, err.Error(), "Orchestrator price higher than the set maximum price of %v wei per %v pixels", int64(1), int64(5))
+
+	BroadcastCfg.SetMaxPrice(nil)
+
 	// Test CreateTicketBatch error
 	sender.On("CreateTicketBatch", mock.Anything, mock.Anything).Return(nil, errors.New("CreateTicketBatch error")).Once()
 
@@ -437,6 +445,55 @@ func TestPing(t *testing.T) {
 	if !verified {
 		t.Error("Unable to verify response from ping request")
 	}
+}
+
+func TestValidatePrice(t *testing.T) {
+	assert := assert.New(t)
+	mid := core.RandomManifestID()
+	b := stubBroadcaster2()
+	oinfo := &net.OrchestratorInfo{
+		PriceInfo: &net.PriceInfo{
+			PricePerUnit:  1,
+			PixelsPerUnit: 3,
+		},
+	}
+
+	s := &BroadcastSession{
+		Broadcaster:      b,
+		ManifestID:       mid,
+		OrchestratorInfo: oinfo,
+		PMSessionID:      "foo",
+	}
+
+	// B's MaxPrice is nil
+	err := validatePrice(s)
+	assert.Nil(err)
+
+	// B MaxPrice > O Price
+	BroadcastCfg.SetMaxPrice(big.NewRat(5, 1))
+	err = validatePrice(s)
+	assert.Nil(err)
+
+	// B MaxPrice == O Price
+	BroadcastCfg.SetMaxPrice(big.NewRat(1, 3))
+	err = validatePrice(s)
+	assert.Nil(err)
+
+	// B MaxPrice < O Price
+	BroadcastCfg.SetMaxPrice(big.NewRat(1, 5))
+	err = validatePrice(s)
+	assert.Errorf(err, err.Error(), "Orchestrator price higher than the set maximum price of %v wei per %v pixels", int64(1), int64(5))
+
+	// O.PriceInfo is nil
+	s.OrchestratorInfo.PriceInfo = nil
+	err = validatePrice(s)
+	assert.EqualError(err, err.Error(), "Invalid orchestrator price")
+
+	// O.PriceInfo.PixelsPerUnit is 0
+	s.OrchestratorInfo.PriceInfo = &net.PriceInfo{PricePerUnit: 1, PixelsPerUnit: 0}
+	err = validatePrice(s)
+	assert.EqualError(err, err.Error(), "Invalid orchestrator price")
+
 }
 
 func TestGetPayment_GivenInvalidBase64_ReturnsError(t *testing.T) {
--- a/server/segment_rpc.go
+++ b/server/segment_rpc.go
@@ -516,6 +516,11 @@ func genPayment(sess *BroadcastSession, numTickets int) (string, error) {
 		return "", nil
 	}
 
+	// Compare Orchestrator Price against BroadcastConfig.MaxPrice
+	if err := validatePrice(sess); err != nil {
+		return "", err
+	}
+
 	batch, err := sess.Sender.CreateTicketBatch(sess.PMSessionID, numTickets)
 	if err != nil {
 		return "", err
@@ -556,4 +561,16 @@ func genPayment(sess *BroadcastSession, numTickets int) (string, error) {
 	}
 
 	return base64.StdEncoding.EncodeToString(data), nil
+}
+
+func validatePrice(sess *BroadcastSession) error {
+	if sess.OrchestratorInfo.PriceInfo.GetPixelsPerUnit() == 0 {
+		return fmt.Errorf("Invalid orchestrator price")
+	}
+	oPrice := big.NewRat(sess.OrchestratorInfo.PriceInfo.GetPricePerUnit(), sess.OrchestratorInfo.PriceInfo.GetPixelsPerUnit())
+	maxPrice := BroadcastCfg.MaxPrice()
+	if maxPrice != nil && oPrice.Cmp(maxPrice) == 1 {
+		return fmt.Errorf("Orchestrator price higher than the set maximum price of %v wei per %v pixels", maxPrice.Num().Int64(), maxPrice.Denom().Int64())
+	}
+	return nil
 }
--- a/server/segment_rpc_test.go
+++ b/server/segment_rpc_test.go
@@ -770,7 +770,7 @@ func TestSubmitSegment_NewBalanceUpdateError(t *testing.T) {
 	assert.EqualError(t, err, expErr.Error())
 }
 
-func TestSubmitSegment_GenPaymentError(t *testing.T) {
+func TestSubmitSegment_GenPaymentError_CreateTicketBatchError(t *testing.T) {
 	b := stubBroadcaster2()
 	existingCredit := big.NewRat(5, 1)
 	balance := &mockBalance{}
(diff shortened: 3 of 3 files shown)


server/segment_rpc.go
```diff
@@ -508,6 +508,12 @@ func genPayment(sess *BroadcastSession, numTickets int) (string, error) {
 		return "", nil
 	}
 
+	// Compare Orchestrator Price against BroadcastConfig.MaxPrice
+	ok := validatePrice(sess)
```
yondonfu: We can omit the `ok` variable here and instead do:

```
if !validatePrice(sess) {
    ...
}
```

Alternatively, if `validatePrice()` returns an error, as suggested in my other comment, instead of a boolean we can do:

```
if err := validatePrice(sess); err != nil {
    return "", err
}
```

server/segment_rpc.go
```diff
@@ -549,3 +555,11 @@ func genPayment(sess *BroadcastSession, numTickets int) (string, error) {
 
 	return base64.StdEncoding.EncodeToString(data), nil
 }
+
+func validatePrice(sess *BroadcastSession) bool {
```
yondonfu: What do you think about returning an error instead of a boolean here so all the details about price validation can be encapsulated in this function (rather than having `genPayment()` decide on an appropriate error message for an invalid price)?

yondonfu: What do you think about return an error instead of a boolean here so all the details about price validation can be encapsulated in this function (rather than having `genPayment()` decide on an appropriate error message for an invalid price)?

server/segment_rpc.go
```diff
@@ -549,3 +555,11 @@ func genPayment(sess *BroadcastSession, numTickets int) (string, error) {
...
+	oPrice := big.NewRat(sess.OrchestratorInfo.PriceInfo.GetPricePerUnit(), sess.OrchestratorInfo.PriceInfo.GetPixelsPerUnit())
```
yondonfu, line 570: I think we should make sure we don't trigger a division by zero error here - `sess.OrchestratorInfo.PriceInfo` could be anything here since we do not apply any checks on a new `OrchestratorInfo` message when calling `updateOrchestratorInfo()` in `SubmitSegment()`

server/segment_rpc.go
```diff
@@ -549,3 +555,11 @@ func genPayment(sess *BroadcastSession, numTickets int) (string, error) {
...
+	if BroadcastCfg.MaxPrice() != nil && oPrice.Cmp(BroadcastCfg.MaxPrice()) == 1 {
```
yondonfu: We can avoid double locking by storing the result of `BroadcastCfg.MaxPrice()` in a variable instead of making 2 calls

server/rpc_test.go
```diff
@@ -338,6 +338,14 @@ func TestGenPayment(t *testing.T) {
 	sender := &pm.MockSender{}
 	s.Sender = sender
 
+	// Test invalid price
+	BroadcastCfg.SetMaxPrice(big.NewRat(1, 5))
+	payment, err = genPayment(s, 1)
+	assert.Equal("", payment)
+	assert.Error(err)
```
yondonfu: Check that this is the error that we expect?

server/rpc_test.go
```diff
@@ -439,6 +447,44 @@ func TestPing(t *testing.T) {
...
+		ManifestID:       mid,
+		OrchestratorInfo: oinfo,
+		PMSessionID:      "foo",
+	}
+
+	// B's MaxPrice is nil
+	ok := validatePrice(s)
+	assert.True(ok)
+
+	// B MaxPrice > O MaxPrice
```
yondonfu: Technically, the price in `OrchestratorInfo` is O's min price and not its max price so I suggest just saying "O price" in the comments for these test cases.

server/segment_rpc_test.go
```diff
@@ -779,18 +779,55 @@ func TestSubmitSegment_GenPaymentError(t *testing.T) {
...
+		Sender:           sender,
+		Balance:          balance,
+		OrchestratorInfo: oinfo,
+	}
+
+	BroadcastCfg.SetMaxPrice(big.NewRat(1, 5))
+
+	_, err := SubmitSegment(s, &stream.HLSSegment{}, 0)
+
+	assert.Error(t, err)
```
yondonfu: Check that this is the error we expect?

yondonfu (changes requested):

server/rpc_test.go
kyriediculous: Oh yeah totally right, not intended, my bad.

server/segment_rpc.go
kyriediculous: While that is true we only call `updateOrchestratorInfo` when `TranscodeResult.OrchestratorInfo` is not `nil`. 

Since `PriceInfo` is always set on the Orchestrator (it can never be nil) there should always be a  `PriceInfo` field included in `TranscodeResult.OrchestratorInfo`. If that's not the case something went horribly wrong in creating the HTTP response.

We could add a guard clause here but if the assumption above doesn't hold up it means code elsewhere isn't working as intended. So I felt like a panic there in such a case would be "OK" as it is unexpected behaviour. 

Wdyt?

server/segment_rpc.go
kyriediculous: That's actually a good idea and I don't know why I didn't think of that 🤦‍♂

server/segment_rpc.go
kyriediculous: for completeness since I tend to forget about offchain mode: 

`PriceInfo` can be nil on the orchestrator if it's in offchain mode, but then we will never be calling `genPayment` anyway.

server/segment_rpc_test.go
kyriediculous: fixed in https://github.com/livepeer/go-livepeer/pull/992/commits/18648d78c0d532e7e1cc26d8bc1e374a560cee34

server/rpc_test.go
kyriediculous: fixed in https://github.com/livepeer/go-livepeer/pull/992/commits/18648d78c0d532e7e1cc26d8bc1e374a560cee34

server/rpc_test.go
kyriediculous: fixed in https://github.com/livepeer/go-livepeer/pull/992/commits/18648d78c0d532e7e1cc26d8bc1e374a560cee34

server/segment_rpc.go
kyriediculous: fixed in https://github.com/livepeer/go-livepeer/pull/992/commits/18648d78c0d532e7e1cc26d8bc1e374a560cee34

server/segment_rpc.go
kyriediculous: implemented in https://github.com/livepeer/go-livepeer/pull/992/commits/18648d78c0d532e7e1cc26d8bc1e374a560cee34

server/segment_rpc.go
kyriediculous: addressed in https://github.com/livepeer/go-livepeer/pull/992/commits/18648d78c0d532e7e1cc26d8bc1e374a560cee34

server/segment_rpc.go
yondonfu: > Since PriceInfo is always set on the Orchestrator (it can never be nil) there should always be a PriceInfo field included in TranscodeResult.OrchestratorInfo. If that's not the case something went horribly wrong in creating the HTTP response.

While we know what the default reference O implementation will be, we shouldn't assume that all Os will be using the same implementation. B shouldn't crash just because an O it is working with is using a faulty implementation (purposeful or not). We should expect the possibility of bad data coming in and program defensively when necessary.

kyriediculous: Fair argument !

kyriediculous: Fixed in https://github.com/livepeer/go-livepeer/pull/992/commits/d828e873a90affa1a5b8f4544df6b4f0103d68e7

Tests added in https://github.com/livepeer/go-livepeer/pull/992/commits/f7cc31f5c8f88b98fe62df5ad649f5be3a37ed98

kyriediculous:
You're right, fixed !

yondonfu:
Needs a rebase!

kyriediculous:
rebased! 
Edit: seems we have a failing test after rebase. Let me check.

Oh yeah we now need to mock credit because we are doing the `completeBalanceUpdate` after returning an error and the balance remains `Staged` so we need to credit it back.

server/segment_rpc_test.go
```diff
@@ -809,6 +809,7 @@ func TestSubmitSegment_GenPaymentError_ValidatePriceError(t *testing.T) {
 	balance.On("StageUpdate", mock.Anything, mock.Anything).Return(1, nil, nil)
 	sender := &pm.MockSender{}
 	sender.On("EV", mock.Anything).Return(big.NewRat(1, 1), nil)
+	balance.On("Credit", mock.Anything)
```
yondonfu, line 813: Can we check that `Credit()` is called with the existing credit (i.e. B's reserved balance) as we do in `TestSubmitSegment_GenPaymentError_CreateTicketBatchError()`?

kyriediculous: Fixed, will await your comment for another rebase.

yondonfu:
Needs another rebase

kyriediculous:
rebased & checks should be fine this time 😅

yondonfu (approved):
LGTM 🚢

PR #993, opened 2019-07-18
cmd, server: O can set/get priceInfo in CLI

### LINKED ISSUE #924: O should be able to modify priceInfo through CLI when node is already running
Orchestrators can now set `pricePerUnit`and `pixelsPerUnit` (required) flags when starting their node to advertise their pricing preferences on a per pixel basis. This change was introduced with #916. There is no way for O's to change this information through the CLI when the node is already running. 

1. add `pricePerUnit` and `pixelsPerUnit` to `w.promtOrchestratorConfig` using `w.readDefaultInt`
2. add newly obtained prompted values to be set in `w.setOrchestratorConfig` and `w.activateOrchestrator`
3. Parse the added form values in the handler for `/setOrchestratorConfig` and `/activateOrchestrator` and set PriceInfo on the node
4. Return PriceInfo from `/OrchestratorInfo`  handler

For (4) we could modify `wizard.getOrchestratorInfo` to return `*lpTypes.Transcoder, *big.Rat, error`
We'd also encode a wrapper struct to JSON in the handler rather than solely `lpTypes.Transcoder`
```
var orchestratorInfo struct {
 Transcoder *lpTypes.Transcoder,
 PriceInfo *big.rat
}
```

kyriediculous:
Enables orchestrators to set new base price info and query its current base price through the CLI

_* This update contains some code duplication but the wizard is due for a rework so we limit the amount of time spent working on it currently_ 

IMPORTANT! 
Breaking change introduced in CLI: `pricePerUnit` is now a required field

- Added user input prompts for `pricePerUnit` and `pixelsPerUnit`
- Added these new fields to the cli server http handlers
- Return `PriceInfo` as a `big.Rat` from `/getOrchestratorInfo`

**TODO:** run the CLI 

Fixes #924

**Checklist:**

### DIFF
--- a/cmd/livepeer_cli/wizard_stats.go
+++ b/cmd/livepeer_cli/wizard_stats.go
@@ -176,7 +176,7 @@ func (w *wizard) orchestratorStats() {
 	if w.offchain {
 		return
 	}
-	t, err := w.getOrchestratorInfo()
+	t, priceInfo, err := w.getOrchestratorInfo()
 	if err != nil {
 		glog.Errorf("Error getting orchestrator info: %v", err)
 		return
@@ -194,11 +194,10 @@ func (w *wizard) orchestratorStats() {
 		[]string{"Delegated Stake", eth.FormatUnits(t.DelegatedStake, "LPT")},
 		[]string{"Reward Cut (%)", eth.FormatPerc(t.RewardCut)},
 		[]string{"Fee Share (%)", eth.FormatPerc(t.FeeShare)},
-		[]string{"Price Per Segment", eth.FormatUnits(t.PricePerSegment, "ETH")},
 		[]string{"Pending Reward Cut (%)", eth.FormatPerc(t.PendingRewardCut)},
 		[]string{"Pending Fee Share (%)", eth.FormatPerc(t.PendingFeeShare)},
-		[]string{"Pending Price Per Segment", eth.FormatUnits(t.PendingPricePerSegment, "ETH")},
 		[]string{"Last Reward Round", t.LastRewardRound.String()},
+		[]string{"Base price per pixel", fmt.Sprintf("%v wei / %v pixels", priceInfo.Num(), priceInfo.Denom())},
 	}
 
 	for _, v := range data {
@@ -387,26 +386,29 @@ func (w *wizard) getBroadcastConfig() (*big.Rat, string) {
 	return config.MaxPrice, config.TranscodingOptions
 }
 
-func (w *wizard) getOrchestratorInfo() (lpTypes.Transcoder, error) {
+func (w *wizard) getOrchestratorInfo() (*lpTypes.Transcoder, *big.Rat, error) {
 	resp, err := http.Get(fmt.Sprintf("http://%v:%v/orchestratorInfo", w.host, w.httpPort))
 	if err != nil {
-		return lpTypes.Transcoder{}, err
+		return nil, nil, err
 	}
 
 	defer resp.Body.Close()
 
 	result, err := ioutil.ReadAll(resp.Body)
 	if err != nil {
-		return lpTypes.Transcoder{}, err
+		return nil, nil, err
 	}
 
-	var tInfo lpTypes.Transcoder
-	err = json.Unmarshal(result, &tInfo)
+	var config struct {
+		Transcoder *lpTypes.Transcoder
+		PriceInfo  *big.Rat
+	}
+	err = json.Unmarshal(result, &config)
 	if err != nil {
-		return lpTypes.Transcoder{}, err
+		return nil, nil, err
 	}
 
-	return tInfo, nil
+	return config.Transcoder, config.PriceInfo, nil
 }
 
 func (w *wizard) getOrchestratorEventSubscriptions() (map[string]bool, error) {
--- a/cmd/livepeer_cli/wizard_transcoder.go
+++ b/cmd/livepeer_cli/wizard_transcoder.go
@@ -28,11 +28,10 @@ func myHostPort() string {
 	return ip + ":" + defaultRPCPort
 }
 
-func (w *wizard) promptOrchestratorConfig() (float64, float64, *big.Int, string) {
+func (w *wizard) promptOrchestratorConfig() (float64, float64, int, int, string) {
 	var (
-		blockRewardCut  float64
-		feeShare        float64
-		pricePerSegment *big.Int
+		blockRewardCut float64
+		feeShare       float64
 	)
 
 	fmt.Printf("Enter block reward cut percentage (default: 10) - ")
@@ -41,8 +40,13 @@ func (w *wizard) promptOrchestratorConfig() (float64, float64, *big.Int, string)
 	fmt.Printf("Enter fee share percentage (default: 5) - ")
 	feeShare = w.readDefaultFloat(5.0)
 
-	fmt.Printf("Enter price per segment in Wei (default: 1) - ")
-	pricePerSegment = w.readDefaultBigInt(big.NewInt(1))
+	fmt.Println("Enter a transcoding base price in wei per pixels")
+	fmt.Println("eg. 1 wei / 10 pixels = 0,1 wei per pixel")
+	fmt.Println()
+	fmt.Printf("Enter amount of pixels that make up a single unit (default: 1 pixel) ")
+	pixelsPerUnit := w.readDefaultInt(1)
+	fmt.Printf("Enter the price for %d pixels in Wei (required) ", pixelsPerUnit)
+	pricePerUnit := w.readDefaultInt(0)
 
 	addr := myHostPort()
 	fmt.Printf("Enter the public host:port of node (default: %v)", addr)
@@ -61,7 +65,7 @@ func (w *wizard) promptOrchestratorConfig() (float64, float64, *big.Int, string)
 		return in, nil
 	})
 
-	return blockRewardCut, feeShare, pricePerSegment, serviceURI
+	return blockRewardCut, feeShare, pricePerUnit, pixelsPerUnit, serviceURI
 }
 
 func (w *wizard) activateOrchestrator() {
@@ -74,14 +78,7 @@ func (w *wizard) activateOrchestrator() {
 	fmt.Printf("Current token balance: %v\n", w.getTokenBalance())
 	fmt.Printf("Current bonded amount: %v\n", d.BondedAmount.String())
 
-	blockRewardCut, feeShare, pricePerSegment, serviceURI := w.promptOrchestratorConfig()
-
-	val := url.Values{
-		"blockRewardCut":  {fmt.Sprintf("%v", blockRewardCut)},
-		"feeShare":        {fmt.Sprintf("%v", feeShare)},
-		"pricePerSegment": {fmt.Sprintf("%v", pricePerSegment.String())},
-		"serviceURI":      {fmt.Sprintf("%v", serviceURI)},
-	}
+	val := w.getOrchestratorConfigFormValues()
 
 	if d.BondedAmount.Cmp(big.NewInt(0)) <= 0 || d.DelegateAddress != d.Address {
 		fmt.Printf("You must bond to yourself in order to become a orchestrator\n")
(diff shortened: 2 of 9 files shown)


cmd/livepeer_cli/wizard_transcoder.go
```diff
@@ -41,9 +41,19 @@ func (w *wizard) promptOrchestratorConfig() (float64, float64, *big.Int, string)
 	fmt.Printf("Enter fee share percentage (default: 5) - ")
 	feeShare = w.readDefaultFloat(5.0)
 
+	// TODO (nico): Remove pricePerSegment
```
yondonfu: Perhaps in the interim before we update the BondingManager contract, to avoid confusion we can:

- Remove all references to `pricePerSegment` in the CLI
- Update the `/activateOrchestrator` endpoint in `server/webserver.go` to *not* check for a `pricePerSegment` in the request and when it makes the `BondingManager.Transcoder()` contract call it hardcodes `pricePerSegment = 0`.

cmd/livepeer_cli/wizard_stats.go
```diff
@@ -199,6 +199,7 @@ func (w *wizard) orchestratorStats() {
 		[]string{"Pending Fee Share (%)", eth.FormatPerc(t.PendingFeeShare)},
 		[]string{"Pending Price Per Segment", eth.FormatUnits(t.PendingPricePerSegment, "ETH")},
 		[]string{"Last Reward Round", t.LastRewardRound.String()},
+		[]string{"base price per pixel", priceInfo.Num().String(), "wei /", priceInfo.Denom().String(), "pixels"},
```
yondonfu: Maybe capitalize to match style of other fields

server/webserver.go
```diff
@@ -253,6 +253,30 @@ func (s *LivepeerServer) cliWebServerHandlers(bindAddr string) *http.ServeMux {
 			return
 		}
 
+		pricePerUnitStr := r.FormValue("pricePerUnit")
```
yondonfu: This code diff block looks identical to the next code diff block. Maybe we can pull this into a helper? We could then unit test the helper.

Ex.

```
func setPrice(node *LivepeerNode, r *http.Request) error {
    // Add code here
}
```

yondonfu (changes requested):

server/webserver.go
kyriediculous: I didn't want to spent too much time on this since we're looking to overhaul the wizard anyway. But if you want I can do this no problem. 

We can do the same for setting the priceInfo on the node in the handlers

server/webserver.go
```diff
@@ -378,6 +402,30 @@ func (s *LivepeerServer) cliWebServerHandlers(bindAddr string) *http.ServeMux {
...
+		pixelsPerUnit, err := strconv.ParseInt(pixelsPerUnitStr, 10, 64)
+		if err != nil {
+			glog.Errorf("Error converting pixelsPerUnit string to int64: %v\n", err)
+			return
+		}
+		if pixelsPerUnit <= 0 {
+			glog.Errorf("pixels per unit must be greater than 0, provided %d\n", pixelsPerUnit)
+			return
+		}
+		s.LivepeerNode.PriceInfo = big.NewRat(pricePerUnit, pixelsPerUnit)
```
j0sh: If we do this here, it should be under a lock, and all other accesses to PriceInfo need to be at least read-locked as well.

kyriediculous: fair point, but then we would need a wrapper `orchestratorConfig` struct, analogous to how we have `BroadcastCfg` in `server/broadcast.go`. This wouldn't be a big deal but then I feel like we should define it on `LivpeerNode` but on `orchestrator` itself. WDYT?

server/webserver.go
yondonfu: Since defining a helper should be pretty straightforward I think its worth moving this request parsing + price info setting code into the helper and just calling the helper in the two required areas

server/webserver.go
j0sh: We could just have a mutex in the livepeer node itself rather than introduce yet another wrapper struct. There's some other data that could be protected around the same mutex as well - `BroadcastJobVideoProfiles` comes to mind, although IIRC it lives in `server`. It could also just be a global: https://github.com/livepeer/go-livepeer/pull/879#discussion_r286709412 (which does not obviate the need for protection, but... one RWMutex to rule them all.)

Writes / reads could also be meditated by a helper function, although IMO golang does not give us convenient misuse resistance due to the lack of const pointer return values.

kyriediculous: I'd have a preference to not use globals or "package level vars" rather unless they are constants.[ This seems to be an idiom widely spread amongst the go community.
](https://twitter.com/davecheney/status/871939730761547776)

If we use one RWMutex on the node itself, wouldn't we also be locking all values ? What are the side-effects of doing so? 

Another reason I would have no problem with an `orchConfig` wrapper here in this particular case is that it becomes easily extendable and all orchestrater related config lives on that `orchestrator` struct rather than on the universal `LivepeerNode` struct. 

Because currently broadcaster has no use for `node.PriceInfo` as he defines his maxPriceInfo in `server.Broadcastcfg`

cmd/livepeer_cli/wizard_transcoder.go
kyriediculous: So we don't have to worry about breaking changes here because in general people won't be using master with the current protocol implementation, or what is the assumption here?

yondonfu: `master` isn't compatible with the protocol live on mainnet right now so its fine to remove references to price per segment except for when the node calls `BondingManager.Transcoder()` - we can change that when we update the contract.

EDIT: On second thought, It might be easier for now to leave the `/activateOrchestrator` endpoint as-is and instead hardcode the CLI initiated request with `pricePerSegment = 0` so that we don't need to update the test-harness as well (which uses the `/activateOrchestrator` endpoint).

kyriediculous: Thanks for clarifying !

server/webserver.go
yondonfu: > If we use one RWMutex on the node itself, wouldn't we also be locking all values ? What are the side-effects of doing so?

We would only lock/unlock the mutex when accessing fields that need to be concurrency safe i.e. `PriceInfo`. I think `serviceURI` also needs to be protected since it can be changed in the CLI.

I would consider eventually introducing a config struct with a single mutex for all fields that are initialized via flags (or parsing a config file) and that can be modified while the node is running if we end up with more of these fields, but I don't think we need to introduce that now since `PriceInfo` already lives within the `LivepeerNode` struct - an additional mutex and helper methods would suffice.

While I agree that this field is O specific and an O specific struct would be a better place for it, `LivepeerNode` holds other O specific state already as well. So, if we wanted to move the field to `orchestrator` we would probably want to move the other O specific state there as well. This feels like a change that doesn't need to happen right now so I would be fine with keeping the field in `LivepeerNode` and then considering a bigger move of O specific state into `orchestrator` separately.

kyriediculous: Can we use segmentMutex , which is already defined on the node for the time being? 
Creating another general RWMutex on the node seems trivial? 

Edit: I turned segmentMutex into a general mutex for the entire `LivepeerNode`

yondonfu: `segmentMutex` protects access to `n.SegmentChans` which is specific to the transcoding workflow. I think we should use a separate mutex for protecting config fields like `PriceInfo` - locking/unlocking for a config field shouldn't impact the transcoding workflow.

kyriediculous: Reinstated segmentMutex for transcoding specific locking/unlocking and created a seperate mutex for other LivepeerNode stuff. Also added locks/unlocks to Set/Get serviceURI 
https://github.com/livepeer/go-livepeer/pull/993/commits/f5edd9c989f6ca0f466973cbc4ad9cf91b166a8c

j0sh: > I would consider eventually introducing a config struct with a single mutex for all fields that are initialized via flags (or parsing a config file) and that can be modified while the node is running

💯 💯 💯 

(The progress here looks good so far; I'll return to the sidelines and leave the rest up to you both.)

cmd/livepeer_cli/wizard_transcoder.go
```diff
@@ -61,7 +66,8 @@ func (w *wizard) promptOrchestratorConfig() (float64, float64, *big.Int, string)
 		return in, nil
 	})
 
-	return blockRewardCut, feeShare, pricePerSegment, serviceURI
+	// TODO (nico): Remove pricePerSegment after contract updates
```
yondonfu: Can remove this comment?

core/livepeernode.go
```diff
@@ -67,7 +67,7 @@ type LivepeerNode struct {
 	TranscoderManager *RemoteTranscoderManager
 	PriceInfo         *big.Rat
 	Balances          *Balances
-
+	mu                sync.RWMutex
```
yondonfu: Since we have the `SetBasePrice()` and `GetBasePrice()` methods now, can we unexport `PriceInfo` and group it with the rest of the unexported fields i.e. `serviceURI`?

It could also be helpful to add a comment noting that `mu` protects config fields

server/webserver_test.go
```diff
@@ -0,0 +1,19 @@
...
+
+import (
+	"math/big"
+	"testing"
+
+	"github.com/livepeer/go-livepeer/core"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestSetOrchestratorPriceInfo(t *testing.T) {
```
yondonfu, line 12: Could we add test cases for an invalid price?

cmd/livepeer_cli/wizard_stats.go
```diff
@@ -197,8 +197,8 @@ func (w *wizard) orchestratorStats() {
 		[]string{"Price Per Segment", eth.FormatUnits(t.PricePerSegment, "ETH")},
```
yondonfu: Can remove this now?

yondonfu (changes requested):

cmd/livepeer_cli/wizard_stats.go
kyriediculous: fixed in https://github.com/livepeer/go-livepeer/pull/993/commits/54863dce0b699a81e16eb27e09bf64b6b28e79a8

server/webserver_test.go
kyriediculous: fixed in https://github.com/livepeer/go-livepeer/pull/993/commits/54863dce0b699a81e16eb27e09bf64b6b28e79a8

cmd/livepeer_cli/wizard_transcoder.go
kyriediculous: fixed in https://github.com/livepeer/go-livepeer/pull/993/commits/54863dce0b699a81e16eb27e09bf64b6b28e79a8

core/livepeernode.go
kyriediculous: fixed in https://github.com/livepeer/go-livepeer/pull/993/commits/54863dce0b699a81e16eb27e09bf64b6b28e79a8

server/webserver.go
kyriediculous: Thanks for your input Josh, as always !

core/livepeernode.go
kyriediculous: fixed in https://github.com/livepeer/go-livepeer/pull/993/commits/54863dce0b699a81e16eb27e09bf64b6b28e79a8

server/webserver_test.go
```diff
@@ -0,0 +1,31 @@
...
```
yondonfu, line 12: Can we add tests for when `pricePerUnitStr` and/or `pixelsPerUnitStr` are not integers i.e. if `strconv.ParseInt()` returns an error?

core/livepeernode.go
```diff
@@ -126,9 +128,27 @@ func (n *LivepeerNode) StopEthServices() error {
...
 }
+
+// SetBasePrice sets the base price for an orchestrator on the node
+func (n *LivepeerNode) SetBasePrice(price *big.Rat) {
+	n.mu.Lock()
+	defer n.mu.Unlock()
+	n.priceInfo = price
+}
+
+// GetBasePrice gets teh base price for an orchestrator
```
yondonfu: "teh" -> "the"

cmd/livepeer_cli/wizard_stats.go
```diff
@@ -194,11 +194,10 @@ func (w *wizard) orchestratorStats() {
 		[]string{"Delegated Stake", eth.FormatUnits(t.DelegatedStake, "LPT")},
 		[]string{"Reward Cut (%)", eth.FormatPerc(t.RewardCut)},
 		[]string{"Fee Share (%)", eth.FormatPerc(t.FeeShare)},
-		[]string{"Price Per Segment", eth.FormatUnits(t.PricePerSegment, "ETH")},
 		[]string{"Pending Reward Cut (%)", eth.FormatPerc(t.PendingRewardCut)},
 		[]string{"Pending Fee Share (%)", eth.FormatPerc(t.PendingFeeShare)},
-		[]string{"Pending Price Per Segment", eth.FormatUnits(t.PendingPricePerSegment, "ETH")},
 		[]string{"Last Reward Round", t.LastRewardRound.String()},
+		[]string{"Base price per pixel", priceInfo.Num().String(), "wei /", priceInfo.Denom().String(), "pixels"},
```
yondonfu: This ends up rendering like this:

Screen Shot 2019-07-22 at 3 46 35 PM

We really want a single concatenated string table entry instead of multiple strings with multiple table entries. Something like this should suffice:

```
fmt.Sprintf("%v wei / %v pixels", priceInfo.Num(), priceInfo.Denom())
```

cmd/livepeer_cli/wizard_transcoder.go
```diff
@@ -41,8 +40,14 @@ func (w *wizard) promptOrchestratorConfig() (float64, float64, *big.Int, string)
 	fmt.Printf("Enter fee share percentage (default: 5) - ")
 	feeShare = w.readDefaultFloat(5.0)
 
-	fmt.Printf("Enter price per segment in Wei (default: 1) - ")
-	pricePerSegment = w.readDefaultBigInt(big.NewInt(1))
+	fmt.Printf("Enter a transcoding base price in wei per pixels")
```
yondonfu: Missing new line here - we should be able to just use `fmt.Println()`

cmd/livepeer_cli/wizard_transcoder.go
```diff
@@ -41,8 +40,14 @@ func (w *wizard) promptOrchestratorConfig() (float64, float64, *big.Int, string)
...
+	fmt.Printf("eg. 1 wei / 10 pixels = 0,1 wei per pixel \n")
```
yondonfu: Can just use `fmt.Println()`

kyriediculous: fixed in 894375e

cmd/livepeer_cli/wizard_transcoder.go
kyriediculous: fixed in 894375e

cmd/livepeer_cli/wizard_stats.go
kyriediculous: fixed in 894375e

core/livepeernode.go
kyriediculous: fixed in 894375e

server/webserver_test.go
kyriediculous: fixed in 894375e

yondonfu (commented):
Needs a rebase!

kyriediculous:
Rebased 👍

yondonfu (approved):
LGTM 🚢

PR #994, opened 2019-07-18
fix `randName` function

darkdarkdragon:
Fixes `randName` function. Sometimes it was returning same name at the same time.

- do `rand.Seed` call only once, at module initialization, instead of calling at every `randName` invocation
- increased returned string length from 10 to 20
- removed copy-paste and moved function to util.go

manually 

**Checklist:**

### DIFF
--- a/core/transcoder.go
+++ b/core/transcoder.go
@@ -27,7 +27,7 @@ type LocalTranscoder struct {
 }
 
 func (lt *LocalTranscoder) Transcode(fname string, profiles []ffmpeg.VideoProfile) ([][]byte, error) {
-	dirName := randName()
+	dirName := common.RandName()
 	fullDirName := filepath.Join(lt.workDir, dirName)
 	err := os.MkdirAll(fullDirName, 0755)
 	if err != nil {
@@ -78,7 +78,7 @@ func (nv *NvidiaTranscoder) Transcode(fname string, profiles []ffmpeg.VideoProfi
 	opts := make([]ffmpeg.TranscodeOptions, len(profiles), len(profiles))
 	for i := range profiles {
 		o := ffmpeg.TranscodeOptions{
-			Oname:   fmt.Sprintf("%s/out_%s", nv.workDir, randName()),
+			Oname:   fmt.Sprintf("%s/out_%s.ts", nv.workDir, common.RandName()),
 			Profile: profiles[i],
 			Accel:   ffmpeg.Nvidia,
 		}
--- a/server/mediaserver_test.go
+++ b/server/mediaserver_test.go
@@ -2,7 +2,6 @@ package server
 
 import (
 	"context"
-	"encoding/hex"
 	"encoding/json"
 	"errors"
 	"flag"
@@ -19,6 +18,7 @@ import (
 
 	"github.com/ethereum/go-ethereum/accounts"
 	ethcommon "github.com/ethereum/go-ethereum/common"
+	"github.com/livepeer/go-livepeer/common"
 	"github.com/livepeer/go-livepeer/eth"
 	"github.com/livepeer/go-livepeer/pm"
 	"github.com/stretchr/testify/assert"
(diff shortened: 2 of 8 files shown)


core/transcoder.go
```diff
@@ -78,7 +78,7 @@ func (nv *NvidiaTranscoder) Transcode(fname string, profiles []ffmpeg.VideoProfi
 	opts := make([]ffmpeg.TranscodeOptions, len(profiles), len(profiles))
 	for i := range profiles {
 		o := ffmpeg.TranscodeOptions{
-			Oname:   fmt.Sprintf("%s/out_%s", nv.workDir, randName()),
+			Oname:   fmt.Sprintf("%s/out_%s", nv.workDir, common.RandName()),
```
j0sh: Probably should add a ".ts" here, because otherwise I suspect the wrong muxer is used. `Oname:   fmt.Sprintf("%s/out_%s.ts", nv.workDir, common.RandName()),`

j0sh:
Is the entropy increase from 80 -> 160 bits still necessary with the seeding fix? Just asking because it's still somewhat nice to have a not-too-long string when looking at file names on disk, etc.

There's also `RandomIdGenerator` within `core/streamdata.go` which is a bit more versatile; maybe we can express `RandomName` in terms of `RandomIdGenerator`?

Maybe this sequence (each can be a separate commit for clarity):
* Move RandomIdGenerator into `common` (or can we keep it in `core` ?)
* Modify `RandomIdGenerator` to return a hex string  - I don't think we use the returned bytes anywhere?
* Define `func RandName() string = RandomIdGenerator(10)`

core/transcoder.go
darkdarkdragon: fixed

darkdarkdragon:
@j0sh fixed, you can look at it again

server/mediaserver_test.go
```diff
@@ -389,24 +381,28 @@ func TestCreateRTMPStreamHandler(t *testing.T) {
...
-		t.Error("Unexpected streamid")
+		t.Error("Unexpected streamid", sid.StreamID())
 	}
 	u, _ = url.Parse("rtmp://localhost/stream/" + expectedSid.String()) // with stream
 	if sid := createSid(u); sid.StreamID() != expectedSid.String() {
 		t.Error("Unexpected streamid")
 	}
 	expectedMid := "mnopq"
-	key := hex.EncodeToString(core.RandomIdGenerator(StreamKeyBytes))
+	rand.Seed(123)
```
j0sh: It's a bit unfortunate we can't override the RNG anymore, because the side effects from the RNG make thinking about the tests harder if we have to take care to reset the internal state. What do you think about reinstating the old monkey-patching behavior?

darkdarkdragon: I would really prefer not to keep functions in the variables if it is avoidable - it makes reasoning about code harder. Here it makes tests slightly harder, but I think it's not complicated enough to resort to monkey-patching.

j0sh: How does it make reasoning about the code harder, since any changes are local changes? And it does make the tests easier  - because they are local changes.

darkdarkdragon: Where it is local? It will be global variable containing function, which can be changed anywhere in the code.
Tests gets read much more rarely than main code.

j0sh: Overriding the RNG is local to the test function. The test function is also the only place we'd do that, which makes the test code cleaner. What is the concern with making reasoning more difficult? Do we have other options?

darkdarkdragon: When I look at `var RandomIdGenerator = func(length uint) []byte {` - I don't know where it is overwritten, I forced to search through all the code to find out that it is done only in test function.

j0sh: We can add a comment to the RandomIdGenerator definition to say that it should only be overwritten during testing.

Overriding the RNG was done for a reason. The original behavior matched your change - we'd reset the seed for each test invocation - but it made tests harder to maintain, and we encountered problems with the RNG being called in *other* parts of the code, leading to hard-to-trace test breakage. See this commit: https://github.com/livepeer/go-livepeer/commit/c50c7b7995594b36ee64295e40970981b8970a6a

darkdarkdragon: @j0sh OK, I've put `RandomIdGenerator` back to variable, you can check it again.

j0sh (approved):
LGTM 🚢

PR #1000, opened 2019-07-23
server: Populate Sender and ExpectedPrice for 0 ticket payments

yondonfu:

This PR populates the `Sender` and `ExpectedPrice` fields of a payment with 0 tickets. Currently, a B will not generate a payment if it already meets O's minimum credit requirement. Generating a 0 ticket payment in this scenario allows B to top off its balance past the minimum credit requirement (O would just debit fees from the balance for each segment transcoded). This change also helps in testing scenarios where O's price is 0 or if O is not counting any pixels in results.

- Always create a payment with `Sender` and `ExpectedPrice` populated in `genPayment()`
- Embed tickets in a payment if the number of tickets required is greater than 0

Added unit tests.

**Checklist:**


kyriediculous (approved):
LGTM 🕵

PR #1007, opened 2019-07-24
Add O Discovery Webhook

### LINKED ISSUE #851: Feature: Webhook on B for managing O connection Info in Offchain mode
An operator currently cannot control the O connection information on a B node dynamically (it can only be done on the command line).

Add an API on B so the operator of a private offchain Livepeer deployment has the ability to:
* Query the orchAddr array on a B
* Dynamically add / remove orchAddr on a B

j0sh:
Adds webhook support for O Discovery and various other cleanups around the discovery code.

Still needs a few things to move this out of WIP status:

* Caching responses. Currently, each method of the OrchestratorPool interface calls out to the webhook. This really is not necessary, since we typically invoke `Size()` then `GetOrchestrators` in quick succession within [`NewSessionManager`](https://github.com/livepeer/go-livepeer/blob/5c17d9d119255d4d4567cb8223fd5dd0ac8a07f3/server/broadcast.go#L162-L177). We really should cache the resulting OrchestratorPool and return its URIs immediately if the O list is requested within some interval, say 1 minute.

* Refresh on empty responses - currently, if an empty response is received from the webhook, then the session manager will not refresh, leading to no Os for the duration of the session. Should be fixed - need to check whether this is an issue within the session manager, or O Discovery.

* Unit tests, including CLI tests

- https://github.com/livepeer/go-livepeer/commit/c92c0a36d66f8927c42b11b5135c11ad14eba8c7 : discovery: Remove OnchainOrchestratorPool.  …
No longer used.

- https://github.com/livepeer/go-livepeer/commit/98e4350358bcbd4ef9ea36d55657debb2a29734d : db_discovery: Use GetURLs as the basis for Size()  …
Code reuse avoids any potential discrepancy between GetURLs / Sizes.

Fixes an issue that was first flagged here : https://github.com/livepeer/go-livepeer/pull/939#discussion_r298403855 and https://github.com/livepeer/go-livepeer/pull/939#discussion_r298407762

- https://github.com/livepeer/go-livepeer/commit/fcc8d7f6958974f7bd0d1635369fedffb7a6073f : discovery: Take well-formed URLs within OrchestratorPool.  …
Reduces the amount of sanitzation the OrchestratorPool performs.
Makes it more explicit that any address cleanup done within the CLI
is a UX affordance.

Corrects an issue first flagged here: https://github.com/livepeer/go-livepeer/pull/939#discussion_r298327624

- https://github.com/livepeer/go-livepeer/commit/4f3696eb62f50b59c0beb0144076546873597ac0 : discovery: Add orchestrator discovery via webhooks.

Manual testing. (Unit testing needed!)

Fixes https://github.com/livepeer/go-livepeer/issues/851

**Checklist:**

### DIFF
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -140,6 +140,7 @@ func main() {
 
 	// API
 	authWebhookURL := flag.String("authWebhookUrl", "", "RTMP authentication webhook URL")
+	orchWebhookURL := flag.String("orchWebhookUrl", "", "Orchestrator discovery callback URL")
 
 	flag.Parse()
 	vFlag.Value.Set(*verbosity)
@@ -585,7 +586,13 @@ func main() {
 		*httpAddr = defaultAddr(*httpAddr, "127.0.0.1", RpcPort)
 
 		// Set up orchestrator discovery
-		if len(orchURLs) > 0 {
+		if *orchWebhookURL != "" {
+			whurl, err := getOrchWebhook(*orchWebhookURL)
+			if err != nil {
+				glog.Fatal("Error setting orch webhook URL ", err)
+			}
+			n.OrchestratorPool = discovery.NewWebhookPool(n, whurl)
+		} else if len(orchURLs) > 0 {
 			n.OrchestratorPool = discovery.NewOrchestratorPool(n, orchURLs)
 		} else if *network != "offchain" {
 			n.OrchestratorPool = discovery.NewDBOrchestratorPoolCache(n)
@@ -704,6 +711,21 @@ func main() {
 	}
 }
 
+func getOrchWebhook(u string) (*url.URL, error) {
+	if u == "" {
+		return nil, nil
+	}
+	p, err := url.ParseRequestURI(u)
+	if err != nil {
+		return nil, err
+	}
+	if p.Scheme != "http" && p.Scheme != "https" {
+		return nil, errors.New("Webhook URL should be HTTP or HTTP")
+	}
+	glog.Infof("Using orchestrator webhook url %s", u)
+	return p, nil
+}
+
 func getAuthWebhookURL(u string) (string, error) {
 	if u == "" {
 		return "", nil
@@ -715,7 +737,7 @@ func getAuthWebhookURL(u string) (string, error) {
 	if p.Scheme != "http" && p.Scheme != "https" {
 		return "", errors.New("Webhook URL should be HTTP or HTTP")
 	}
-	glog.Infof("Using webhook url %s", u)
+	glog.Infof("Using auth webhook url %s", u)
 	return u, nil
 }
 
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -720,7 +720,7 @@ func getOrchWebhook(u string) (*url.URL, error) {
 		return nil, err
 	}
 	if p.Scheme != "http" && p.Scheme != "https" {
-		return nil, errors.New("Webhook URL should be HTTP or HTTP")
+		return nil, errors.New("Webhook URL should be HTTP or HTTPS")
 	}
 	glog.Infof("Using orchestrator webhook url %s", u)
 	return p, nil
@@ -735,7 +735,7 @@ func getAuthWebhookURL(u string) (string, error) {
 		return "", err
 	}
 	if p.Scheme != "http" && p.Scheme != "https" {
-		return "", errors.New("Webhook URL should be HTTP or HTTP")
+		return "", errors.New("Webhook URL should be HTTP or HTTPS")
 	}
 	glog.Infof("Using auth webhook url %s", u)
 	return u, nil
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -173,13 +173,21 @@ func main() {
 		},
 	}
 
-	// If multiple orchAddresses specified, ensure other necessary flags present and clean up list
-	var orchAddresses []string
+	// If multiple orchAddr specified, ensure other necessary flags present and clean up list
+	var orchURLs []*url.URL
 	if len(*orchAddr) > 0 {
-		orchAddresses = strings.Split(*orchAddr, ",")
-		for i := range orchAddresses {
-			orchAddresses[i] = strings.TrimSpace(orchAddresses[i])
-			orchAddresses[i] = defaultAddr(orchAddresses[i], "127.0.0.1", RpcPort)
+		for _, addr := range strings.Split(*orchAddr, ",") {
+			addr = strings.TrimSpace(addr)
+			addr = defaultAddr(addr, "127.0.0.1", RpcPort)
+			if !strings.HasPrefix(addr, "http") {
+				addr = "https://" + addr
+			}
+			uri, err := url.ParseRequestURI(addr)
+			if err != nil {
+				glog.Error("Could not parse orchestrator URI: ", err)
+				continue
+			}
+			orchURLs = append(orchURLs, uri)
 		}
 	}
 
@@ -273,8 +281,8 @@ func main() {
 		if n.OrchSecret == "" {
 			glog.Fatal("Missing -orchSecret")
 		}
-		if len(orchAddresses) > 0 {
-			server.RunTranscoder(n, orchAddresses[0], *maxSessions)
+		if len(orchURLs) > 0 {
+			server.RunTranscoder(n, orchURLs[0].Host, *maxSessions)
 		} else {
 			glog.Fatal("Missing -orchAddr")
 		}
@@ -577,8 +585,8 @@ func main() {
 		*httpAddr = defaultAddr(*httpAddr, "127.0.0.1", RpcPort)
 
 		// Set up orchestrator discovery
-		if len(orchAddresses) > 0 {
-			n.OrchestratorPool = discovery.NewOrchestratorPool(n, orchAddresses)
+		if len(orchURLs) > 0 {
+			n.OrchestratorPool = discovery.NewOrchestratorPool(n, orchURLs)
 		} else if *network != "offchain" {
 			n.OrchestratorPool = discovery.NewDBOrchestratorPoolCache(n)
 		}
--- a/discovery/discovery_test.go
+++ b/discovery/discovery_test.go
@@ -2,6 +2,7 @@ package discovery
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"math/big"
 	"math/rand"
@@ -743,4 +744,138 @@ func TestCachedPool_GetOrchestrators_TicketParamsValidation(t *testing.T) {
 	assert.Nil(err)
 	assert.Len(infos, 0)
 	sender.AssertNumberOfCalls(t, "ValidateTicketParams", 50)
+}
+
+func TestNewWHOrchestratorPoolCache(t *testing.T) {
+	require := require.New(t)
+	assert := assert.New(t)
+
+	// mock webhook and orchestrator info request
+	addresses := []string{"https://127.0.0.1:8936", "https://127.0.0.1:8937", "https://127.0.0.1:8938"}
+
+	getURLsfromWebhook = func(cbUrl *url.URL) ([]byte, error) {
+		var wh []webhookResponse
+		for _, addr := range addresses {
+			wh = append(wh, webhookResponse{Address: addr})
+		}
+		return json.Marshal(&wh)
+	}
+
+	serverGetOrchInfo = func(c context.Context, b server.Broadcaster, s *url.URL) (*net.OrchestratorInfo, error) {
+		return &net.OrchestratorInfo{Transcoder: "transcoder"}, nil
+	}
+
+	// mock livepeer node
+	node, _ := core.NewLivepeerNode(nil, "", nil)
+	perm = func(len int) []int { return rand.Perm(3) }
+
+	// assert created webhook pool is correct length
+	whURL, _ := url.ParseRequestURI("https://livepeer.live/api/orchestrator")
+	whpool := NewWebhookPool(node, whURL)
+	assert.Equal(3, whpool.Size())
+
+	// assert that list is not refreshed if lastRequest is less than 1 min ago and hash is the same
+	lastReq := whpool.lastRequest
+	orchInfo, err := whpool.GetOrchestrators(2)
+	require.Nil(err)
+	assert.Len(orchInfo, 2)
+	assert.Equal(3, whpool.Size())
+
+	urls := whpool.pool.GetURLs()
+	assert.Len(urls, 3)
+
+	for _, addr := range addresses {
+		uri, _ := url.ParseRequestURI(addr)
+		assert.Contains(urls, uri)
+	}
+
+	//  assert that list is not refreshed if lastRequest is more than 1 min ago and hash is the same
+	lastReq = time.Now().Add(-2 * time.Minute)
+	whpool.lastRequest = lastReq
+	orchInfo, err = whpool.GetOrchestrators(2)
+	require.Nil(err)
+	assert.Len(orchInfo, 2)
+	assert.Equal(3, whpool.Size())
+	assert.NotEqual(lastReq, whpool.lastRequest)
+
+	urls = whpool.pool.GetURLs()
+	assert.Len(urls, 3)
+
+	for _, addr := range addresses {
+		uri, _ := url.ParseRequestURI(addr)
+		assert.Contains(urls, uri)
+	}
+
+	// mock a change in webhook addresses
+	addresses = []string{"https://127.0.0.1:8932", "https://127.0.0.1:8933", "https://127.0.0.1:8934"}
+
+	//  assert that list is not refreshed if lastRequest is less than 1 min ago and hash is not the same
+	lastReq = time.Now()
+	whpool.lastRequest = lastReq
+	orchInfo, err = whpool.GetOrchestrators(2)
+	require.Nil(err)
+	assert.Len(orchInfo, 2)
+	assert.Equal(3, whpool.Size())
+	assert.Equal(lastReq, whpool.lastRequest)
+
+	urls = whpool.pool.GetURLs()
+	assert.Len(urls, 3)
+
+	for _, addr := range addresses {
+		uri, _ := url.ParseRequestURI(addr)
+		assert.NotContains(urls, uri)
+	}
+
+	//  assert that list is refreshed if lastRequest is longer than 1 min ago and hash is not the same
+	lastReq = time.Now().Add(-2 * time.Minute)
+	whpool.lastRequest = lastReq
+	orchInfo, err = whpool.GetOrchestrators(2)
+	require.Nil(err)
+	assert.Len(orchInfo, 2)
+	assert.Equal(3, whpool.Size())
+	assert.NotEqual(lastReq, whpool.lastRequest)
+
+	urls = whpool.pool.GetURLs()
+	assert.Len(urls, 3)
+
+	for _, addr := range addresses {
+		uri, _ := url.ParseRequestURI(addr)
+		assert.Contains(urls, uri)
+	}
+}
+
+func TestDeserializeWebhookJSON(t *testing.T) {
+	assert := assert.New(t)
+
+	// assert input of webhookResponse address object returns correct address
+	resp, _ := json.Marshal(&[]webhookResponse{webhookResponse{Address: "https://127.0.0.1:8936"}})
+	urls, err := deserializeWebhookJSON(resp)
+	assert.Nil(err)
+	assert.Equal("https://127.0.0.1:8936", urls[0].String())
+
+	// assert input of empty byte array returns JSON error
+	urls, err = deserializeWebhookJSON([]byte{})
+	assert.Contains(err.Error(), "unexpected end of JSON input")
+	assert.Nil(urls)
+
+	// assert input of empty byte array returns empty object
+	resp, _ = json.Marshal(&[]webhookResponse{webhookResponse{}})
+	urls, err = deserializeWebhookJSON(resp)
+	assert.Nil(err)
+	assert.Empty(urls)
+
+	// assert input of invalid addresses returns invalid JSON error
+	urls, err = deserializeWebhookJSON(make([]byte, 64))
+	assert.Contains(err.Error(), "invalid character")
+	assert.Empty(urls)
+
+	// assert input of invalid JSON returns JSON unmarshal object error
+	urls, err = deserializeWebhookJSON([]byte(`{"name":false}`))
+	assert.Contains(err.Error(), "cannot unmarshal object")
+	assert.Empty(urls)
+
+	// assert input of invalid JSON returns JSON unmarshal number error
+	urls, err = deserializeWebhookJSON([]byte(`1112`))
+	assert.Contains(err.Error(), "cannot unmarshal number")
+	assert.Empty(urls)
 }
--- a/discovery/discovery_test.go
+++ b/discovery/discovery_test.go
@@ -67,15 +67,6 @@ func TestNewOrchestratorPoolWithPred_NilEthClient_ReturnsNil_LogsError(t *testin
 	assert.Nil(t, pool)
 }
 
-func TestNewOnchainOrchestratorPool_NilEthClient_ReturnsNil_LogsError(t *testing.T) {
-	node, _ := core.NewLivepeerNode(nil, "", nil)
-	errorLogsBefore := glog.Stats.Error.Lines()
-	pool := NewOnchainOrchestratorPool(node)
-	errorLogsAfter := glog.Stats.Error.Lines()
-	assert.NotZero(t, errorLogsAfter-errorLogsBefore)
-	assert.Nil(t, pool)
-}
-
 func TestNewDBOrchestratorPoolCache_NilEthClient_ReturnsNil_LogsError(t *testing.T) {
 	node, _ := core.NewLivepeerNode(nil, "", nil)
 	errorLogsBefore := glog.Stats.Error.Lines()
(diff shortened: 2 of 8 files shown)


cmd/livepeer/livepeer.go
```diff
@@ -629,6 +636,21 @@ func main() {
...
+func getOrchWebhook(u string) (*url.URL, error) {
+	if u == "" {
+		return nil, nil
+	}
+	p, err := url.ParseRequestURI(u)
+	if err != nil {
+		return nil, err
+	}
+	if p.Scheme != "http" && p.Scheme != "https" {
+		return nil, errors.New("Webhook URL should be HTTP or HTTP")
```
iameli: typo: `HTTP or HTTPS`

discovery/wh_discovery.go
```diff
@@ -3,80 +3,148 @@ package discovery
...
+		orchs = append(orchs, &orch)
+	}
+	_, err = cacheDBOrchs(node, orchs)
+	return addrs, hash, err
+
 }
 
-func getURLs(cbUrl *url.URL) ([]*url.URL, error) {
-	// TODO cache results for some time to avoid repeated callouts
+var getURLsfromWebhook = func(cbUrl *url.URL, hashPreviousResponse ethcommon.Hash) ([]*url.URL, ethcommon.Hash, error) {
```
j0sh: The `hashPreviousResponse` argument here is not used

discovery/wh_discovery.go
```diff
@@ -3,80 +3,148 @@ package discovery
...
+	if len(addrs) == 0 || hash == hashPreviousResponse || err != nil {
+		return nil, hash, err
+	}
+
+	// cache addrs retrieved from webhook
+	var orchs []*lpTypes.Transcoder
+	for _, addr := range addrs {
+		orch := lpTypes.Transcoder{
+			// TODO: What do we store here??? Must be unique for each transcoder stored
+			Address:    common.HexToAddress(randString()),
```
j0sh: This would insert a new DB row for every refresh, for every orchestrator. For now, would simply using the `addr` itself work? Then we don't need the additional `randString` function anymore.

discovery/wh_discovery.go
```diff
@@ -3,80 +3,148 @@ package discovery
...
+		hashPreviousResponse: hash,
+	}
+}
+
+func (w *webhookPool) GetURLs() []*url.URL {
+	return w.db.GetURLs()
+}
+
+func (w *webhookPool) Size() int {
+	return len(w.db.GetURLs())
```
j0sh: Recall that the `Size()` is called *first* before GetOrchestrators within NewSessionManager. That would lead to a mismatch between `Size` and `GetOrchestrators` under the following conditions:

1. Node starts up, populates cache.
2. Webhook server updates orchestrator list, changing its size
3. Some time passes
4. New session is registered
5. `Size` is called via NewSessionManager. This pulls from DB
6. GetOrchestrators invoked. This returns a different set of orchestrators, with a different size

To mitigate this, I'd recommend following the original approach of defining a helper method to get to get a list of addresses in the same way that GetOrchestrators does, without actually creating the OrchestratorPool which incurs a bunch of network calls. For the DB call-outs in this part, you'll need to use `db.getURLs()` that was introduced earlier in this PR. Basically, the only thing that should really change about this PR is the addition of a `getURLs` helper method for the webhook pool itself.

```golang
func (w *webhookPool) getURLs() ([]*url.URL, error) {
   if time.Since(w.lastRequested) < whRefreshInterval {
    return w.db.getURLs()
  }
  ... remaining code here from GetOrchestrators, except for OrchestratorPool stuff ...
}

func (w *webhookPool) GetURLs() []*url.URL {
	uris, _ := w.getURLs()
	return uris
}

func (w *webhookPool) Size() int {
	return len(w.GetURLs())
}

func (w *webhookPool) GetOrchestrators(numOrchestrators int) ([]*net.OrchestratorInfo, error) {
	addrs, err := w.getURLs()
	if err != nil {
		return nil, err
	}
	pool := NewOrchestratorPool(w.node, addrs)
	if pool == nil {
		return nil, errors.New("Nil orchestrator pool")
	}
	return pool.GetOrchestrators(numOrchestrators)
}
```

discovery/wh_discovery.go
```diff
@@ -3,80 +3,148 @@ package discovery
...
+
+	if err != nil || len(addresses) == 0 {
+		glog.Error("Error getting urls from webhook: ", err)
+		return nil
+	}
+
+	return &webhookPool{
+		db:                   NewDBOrchestratorPoolCache(node),
+		callback:             callback,
+		lastRequested:        time.Now(),
```
j0sh: This should only be set when the callback succeeds

discovery/wh_discovery.go
```diff
@@ -3,80 +3,148 @@ package discovery
...
+func (w *webhookPool) Size() int {
+	return len(w.db.GetURLs())
+}
+
+func (w *webhookPool) GetOrchestrators(numOrchestrators int) ([]*net.OrchestratorInfo, error) {
+	var addrs []*url.URL
+
+	if time.Since(w.lastRequested) < whRefreshInterval {
+		// retrive addrs from DB if time since lastRequested is less than the refresh interval
+		glog.Error("IN TIME SINCE")
```
j0sh: stray

discovery/wh_discovery.go
```diff
@@ -3,80 +3,148 @@ package discovery
...
+		return w.db.GetOrchestrators(numOrchestrators)
+	}
+
+	// retrive addrs from webhook if time since lastRequested is more than the refresh interval
+	addresses, hash, err := cacheWebhookTranscoders(w.db.node, w.callback, w.hashPreviousResponse)
```
j0sh: The division of responsibilities between `cacheWebhookTranscoders` and `GetOrchestrators` is a little unclear. One way to improve the flow is to simplify `cacheWebhookTranscoders` so it only does the actual caching, eg the DB insertion, and extract out the actual webhook call into the body here. Following up on the earlier suggestion of a `getURLs` helper method, something like:

```golang
func (w *webhookPool) getURLs() ([]*url.URL, error) {
  w.mu.RLock() // see explanation below
  lastReq := w.lastRequested
  w.mu.RUnlock()
  if time.Since(lastReq) < whRefreshInterval {
    return w.db.getURLs()
  }
  addrs, hash, err := getURLs(w.callback)
  if err != nil {
    return w.db.getURLs()
  }
  // Since these instance vars could be modified by different threads
  // they probably need to be locked as well. Remember to to read-lock
  // accesses as necessary.
  w.mu.Lock()
  if hash != w.hashPreviousResponse {
    cacheAddressesInDB(w.db, addrs)
  }
  w.lastRequested = time.Now() // this needs to be set as well
  w.hashPreviousResponse = hash
  w.mu.Unlock()
  return addrs, nil
}
```

(This example isn't as tight as it could be, especially because of the lack of mutual exclusion around the webhook call and the blocking DB write. If you want to see a more robust method of implementing this, look through git history for the code controlling access around selectOrchestrator, prior to us adding the broadcast session manager.)

discovery/wh_discovery.go
```diff
@@ -3,80 +3,148 @@ package discovery
...
 type webhookResponse struct {
 	Address string
 }
 
 type webhookPool struct {
-	node     *core.LivepeerNode
-	callback *url.URL
+	db                   *DBOrchestratorPoolCache
+	callback             *url.URL
+	hashPreviousResponse ethcommon.Hash
```
j0sh: `responseHash` ?

discovery/wh_discovery.go
```diff
@@ -3,80 +3,148 @@ package discovery
...
+	lastRequested        time.Time
```
j0sh: `lastRequest` ?

j0sh (commented):
@angyangie Left some notes on the DB caching implementation -- haven't looked through the tests yet

discovery/wh_discovery.go
angyangie: Ah yes, that's what it was originally, but I was using the wrong string to bytes conversion. Initially I was using `common.HexToAddress(addr.String())` which was giving me the same value for all transcoder addresses of the same length. Just switched to: `common.BytesToAddress([]byte(addr.String())),`. The `randString` was a placeholder. Thank you for answering my question!

discovery/wh_discovery.go
angyangie: Does line 35 and the following check for an error & existence of addresses not satisfy this requirement? I'm not sure what you mean.

discovery/wh_discovery.go
angyangie: Ah really great points! Didn't look at `NewSessionManager`, but see what you mean re: Size(), I see the benefits of calling `getURLsfromWebhook` only from `NewWebhookPool` and `getURLs` ....  `getURLsfromWebhook` would be called by `Size()`, `GetURLs()`, and `GetOrchestrators()`, which would offer consistency. I'll change to look like your example!

I do wonder what would happen with this line in `NewSessionManager` though ....
`createSessions: func() ([]*BroadcastSession, error) { return selectOrchestrator(node, params, pl, numOrchs) }` ... `createSessions` is called when `refreshSessions` is called. Wouldn't we also have a problem with `m.numOrchs` being derived from the number present in `NewSessionManager`? That number wouldn't necessarily reflect the number of O's currently in the DB.

discovery/wh_discovery.go
angyangie: I like the separation of `cacheAddressesInDB` and `getURLsfromWebhook`. 

What do you mean by `especially because of the lack of mutual exclusion around the webhook call `? i.e. only one call to the webhook should be happening at a time? 

I'm going to push something up, but will earmark this particular comment as a TODO, for now. Looking for that code you suggested and found this, when the function was renamed, but not sure if what you mentioned is included here: https://github.com/livepeer/go-livepeer/commit/f7c57ed0f43dfeee16e018a25fc9c5b147bd5f09

go.mod
```diff
@@ -17,6 +17,7 @@ require (
 	github.com/fjl/memsize v0.0.0-20190710130421-bcb5799ab5e5 // indirect
 	// github.com/gballet/go-libpcsclite v0.0.0-20190607065134-2772fd86a8ff // indirect
 	github.com/gballet/go-libpcsclite v0.0.0-20190403181518-312b5175032f // indirect
+	github.com/go-stack/stack v1.8.0 // indirect
```
j0sh: Hmm, where are these coming from? Not that I know much about go modules...

discovery/wh_discovery.go
j0sh: How about this:

```golang
func NewWebhookPool(node *core.LivepeerNode, callback *url.URL) *webhookPool {
  p := &webhookPool{
    db: NewDBOrchestratorPoolCache(node),
    callback: callback,
    mu: &sync.RWMutex{}
  }
  go p.getURLs() // OK if this fails; that's really the whole reason we have a persistent cache
  return p
}
```

discovery/wh_discovery.go
j0sh: > only one call to the webhook should be happening at a time?

Yes, exactly. We also don't want to block other requests that may be requesting the webhook at the same time; they should return immediately with the cached values that we do have.

I was thinking of startSessionListener which was removed here, although I see now that the requirements there were slightly different: 52163927850356e1027017426024d1f0ea783b8c

The desired flow here works something like:

1. Check a busy flag (under a lock)
1. If busy, return cached values and exit
1. If not busy, set busy flag.
1. Make a webhook request
1. Update DB if needed
1. Unset busy flag

In fact, now that I'm reviewing things again, this flow is similar to how the current `refreshSession` works, so that can be used as a reference as well. Note that maintaining mutual exclusion in `refreshSessions` doesn't negate the need for mutual exclusion in the webhook, since the broadcast manager is per-stream, while the webhook is global.

discovery/wh_discovery.go
j0sh: Yes, good point - we could still get a size mismatch there. Since it won't result in a discrepancy at the very beginning of a session, I'll leave it up to you whether to fix this. One approach is to derive `count` within `selectOrchestrator` itself, but I'm not excited about potentially hitting the DB twice with each `refreshSessions` call. This might be more palatable if we had a dedicated `COUNT` query to retrieve `Size()`.

This brings up two more issues with using the DB that I'm seeing now:

* It's unclear when a DB entry should be removed or not. Ideally we would only be using the set of Os that the webhook returns; the current upsert behavior makes more sense for onchain Os but probably less so here. Otherwise we'll have an ever-growing set of Os in DB that doesn't actually reflect the set that the webhook wants us to use.

* Now that I'm reviewing things yet again, we're going to be doing redundant `GetOrchInfo` calls to all the orchestrators: once within `cacheDBOrchs`, then again in `GetOrchestrators`. This will most often be done in quick succession, which may be excessive. (At least we have the hash to mitigate this somewhat.)

What do you think?

iameli:
Currently getting a division-by-zero error after the webhook runs:

```
I0913 04:53:51.621238       1 livepeer.go:205] ***Livepeer is running on the offchain*** network
I0913 04:53:51.621322       1 livepeer.go:218] Creating data dir: /root/.lpData/offchain
I0913 04:53:51.654367       1 livepeer.go:295] ***Livepeer is in off-chain mode***
I0913 04:53:51.654398       1 livepeer.go:725] Using orchestrator webhook url http://api.default.svc.cluster.local/api/orchestrator
I0913 04:53:51.676179       1 rpc.go:220] Connecting RPC to https://10.40.2.45:8935
I0913 04:53:51.676234       1 rpc.go:220] Connecting RPC to https://10.40.0.34:8935
I0913 04:53:51.676253       1 rpc.go:220] Connecting RPC to https://10.40.1.36:8935
panic: division by zero

goroutine 39 [running]:
math/big.(*Rat).SetFrac64(0xc00024df88, 0x0, 0x0, 0xc00014a110)
	/usr/lib/go-1.13/src/math/big/rat.go:318 +0xfa
math/big.NewRat(...)
	/usr/lib/go-1.13/src/math/big/rat.go:32
github.com/livepeer/go-livepeer/discovery.cacheDBOrchs.func1(0xc0001149f0)
	/go/src/github.com/livepeer/go-livepeer/discovery/db_discovery.go:143 +0x194
created by github.com/livepeer/go-livepeer/discovery.cacheDBOrchs
	/go/src/github.com/livepeer/go-livepeer/discovery/db_discovery.go:158 +0x1fe
```

j0sh:
@iameli Looks like it's because the DB caching code hasn't been tuned for offchain mode where there's no pricing information available.

https://github.com/livepeer/go-livepeer/blob/988f2e8b639d0e60108fee36867079129b7b7d0d/discovery/db_discovery.go#L143

Coupled with the DB-related issues mentioned in https://github.com/livepeer/go-livepeer/pull/1007#discussion_r323834062 , I'm thinking it'd be best to forgo DB caching for this PR and rely on storing a list in-memory.

Having a persistent cache offers an advantage when the webhook endpoint is offline, but the node itself hasn't yet received a set of orchestrators (eg, during node startup). That would be a nice-to-have but given the issues we are having fitting the DB cache into this use case, we can forgo it for now.

discovery/wh_discovery.go
angyangie: We chatted about this on our last one on one for a bit, but to summarize what I think we concluded:
- The size mismatch is okay for now
- We should remove all DB entries that do not exist in the new set of Os pulled from the Webhook. @j0sh I'm thinking we can just loop through the saved DB Os, compare with the new O list and remove those that don't exist in the new O list, then save new O's that aren't in the DB. Does that make sense as a list of steps?

As for redundant `GetOrchInfo` calls, do you mean when we first create the webhook pool in `NewWebhookPool` and when we call `n.OrchestratorPool.GetOrchestrators(count)` in `selectOrchestrator`? Do you think this should be addressed in this PR?

(though this might be a mute point since it seems we concluded that DB caching isn't necessary for now)

discovery/wh_discovery.go
angyangie: Oh I see what you mean -  `refreshSession` is a good guide. Again, might be something for when we do do caching.

go.mod
angyangie: Apparently these are changes that need to be added, given my addition of some new libraries in this PR

angyangie:
@j0sh if we're not doing DB caching, should I remove the last two commits here, add some unit tests, and call it a day?

j0sh:
> if we're not doing DB caching, should I remove the last two commits here, add some unit tests, and call it a day?

@angyangie We should remove the DB part, but keep the other non-DB changes such as the refresh interval check and the hash check. Take note of the comments around the structure of the code around the usage of `getURLs` because I believe most of it still applies.

We do still need a cache. That cache can just be non-persistent, eg us storing a reference to an `OrchestratorPool` within the `webhookPool` struct. If we're within the refresh interval or the response hash matches, retrieve the URLs from that cached pool. Otherwise, recreate the OrchestratorPool field if the hash has changed.

Something like this:

```golang
type webhookPool struct {
 ...
 OrchestratorPool pool
...
}

func (w *webhookPool) getURLs() ([]*url.URL, error) { {
  // Various things simplified here around particular APIs, error handling, locking, etc
  // All these are still important!
  if (time.Now() < w.lastRefresh) return w.pool.GetURLs(), nil
  body, err := callWebhook(w.callback)
  if (err != nil) return nil, err
  w.lastRefresh = time.Now()
  // note shortcut here: exit early prior to deserialization 
  if (w.hash == hash(body)) return w.pool.GetURLs(), nil
  urls, err = deserializeWebhookJSON(body)
  if (err != nil) return nil, err // should we also reset w.lastRefresh ??
  w.hash = hash(body)
  w.pool = NewOrchestratorPool(w.pool.node, urls)
  return urls, nil
}
```

discovery/wh_discovery.go
j0sh: > The size mismatch is okay for now

:+1:

> I'm thinking we can just loop through the saved DB Os, compare with the new O list and remove those that don't exist in the new O list, then save new O's that aren't in the DB

Or just DROP the table / DELETE all entries from the DB unconditionally, and repopulate.

> As for redundant GetOrchInfo calls,

That would happen every time we call out to `GetOrchestrators(count)` and receive a response that leads to a DB update. This is because `GetOrchestrators` and `cacheDBOrchs` both invoke `GetOrchInfo`. But yeah, not a concern if we omit the DB caching for now.

discovery/wh_discovery.go
j0sh: We still need caching. We can just live without the persistence that the DB offers, since that only helps us at the very beginning, during start-up time if the webhook endpoint happens to be down and we don't already have a populated in-memory cache. For other scenarios, an in-memory cache of orchestrator URLs is fine.

angyangie:
@j0sh ahhh I see, I understand now. Thank you for the explanation and example :)

discovery/wh_discovery.go
angyangie: > Or just DROP the table / DELETE all entries from the DB unconditionally, and repopulate.

Ah yes, that's easier!

> not a concern if we omit the DB caching for now.

That makes sense, great :)

discovery/wh_discovery.go
```diff
@@ -0,0 +1,162 @@
...
+func (w *webhookPool) Size() int {
+	return len(w.GetURLs())
+}
+
+func (w *webhookPool) GetOrchestrators(numOrchestrators int) ([]*net.OrchestratorInfo, error) {
+	addrs, err := w.getURLs()
+	if err != nil {
+		return nil, err
+	}
+	pool := NewOrchestratorPool(w.node, addrs)
```
j0sh: Do we need this part since it's already done within `w.getURLs()`?

discovery/wh_discovery.go
```diff
@@ -0,0 +1,162 @@
...
+type webhookPool struct {
+	node         *core.LivepeerNode
+	pool         *orchestratorPool
+	callback     *url.URL
+	responseHash ethcommon.Hash
+	lastRequest  time.Time
+	mu           *sync.RWMutex
+}
+
+func NewWebhookPool(node *core.LivepeerNode, callback *url.URL) *webhookPool {
```
j0sh, line 34: The repetition here can probably be avoided, would this work?

```golang
func NewWebhookPool(node *core.LivepeerNode, callback *url.URL) *webhookPool {
  p := &webhookPool {
    node: node,
    callback: callback
    mu: &sync.RWMutex{}
  }
  go p.getURLs()
  return p
}
```

discovery/wh_discovery.go
```diff
@@ -0,0 +1,162 @@
...
+		pool:         pool,
+		node:         node,
+		callback:     callback,
+		lastRequest:  time.Now(),
+		responseHash: ethcommon.BytesToHash(crypto.Keccak256(body)),
+		mu:           &sync.RWMutex{},
+	}
+}
+
+func (w *webhookPool) getURLs() ([]*url.URL, error) {
```
j0sh, line 44: This function doesn't really assure mutual exclusivity of the webhook call - multiple calls could be occurring at once, with "last response wins" - but it's probably OK for now.

j0sh (commented):
@angyangie Glanced through it quickly, getting close! Haven't looked through the tests yet, but left some comments on the code itself.

discovery/wh_discovery.go
angyangie: You're right - we don't!

discovery/wh_discovery.go
angyangie: Much more concise - thank you

discovery/wh_discovery.go
angyangie: oh yes, that's a part I didn't add. If it's okay for now, I'd rather leave it as is and make note of that as an optimization for the future

discovery/wh_discovery.go
```diff
@@ -0,0 +1,138 @@
...
+	return len(w.GetURLs())
+}
+
+func (w *webhookPool) GetOrchestrators(numOrchestrators int) ([]*net.OrchestratorInfo, error) {
+	_, err := w.getURLs()
+	if err != nil {
+		return nil, err
+	}
+
+	return w.pool.GetOrchestrators(numOrchestrators)
```
j0sh, line 102: Needs to be RWLocked before accessing `w.pool`

cmd/livepeer/livepeer.go
j0sh: @angyangie This typo is still here

discovery/wh_discovery.go
```diff
@@ -0,0 +1,138 @@
...
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		glog.Error("Unable to read response body ", err)
+		return nil, err
+	}
+
+	return body, nil
+}
+
+var deserializeWebhookJSON = func(body []byte) ([]*url.URL, error) {
```
j0sh: Can standalone tests be added for this function? Test things like empty bodies, invalid addresses, unexpected JSON (eg, an object rather than an address), invalid/mangled JSON, etc.

discovery/discovery_test.go
```diff
@@ -756,3 +744,110 @@ func TestCachedPool_GetOrchestrators_TicketParamsValidation(t *testing.T) {
...
 }
+
+func TestNewWHOrchestratorPoolCache(t *testing.T) {
+	require := require.New(t)
+	assert := assert.New(t)
+
+	// mock webhook and orchestrator info request
+	addresses := []string{"https://127.0.0.1:8936", "https://127.0.0.1:8937", "https://127.0.0.1:8938"}
+
+	getURLsfromWebhook = func(cbUrl *url.URL) ([]byte, error) { return []byte{}, nil }
```
j0sh: If we returned actual json here, then wouldn't need to stub out the `deserialzeWebhookJSON` part. That's somewhat preferable since it'll exercise more of the actual code paths. Something like:

```golang
getURLSfromWebhook = func ... {
  a := []
  for _, v := range addresses {
    append(a, webhookResponse{Address: a})
  }
  return json.Marshal(&a)
}
```

angyangie: Good point!! Thanks for pointing out this optimization

discovery/wh_discovery.go
angyangie: Fo sho (i.e. for sure)!

discovery/wh_discovery.go
```diff
@@ -96,7 +96,11 @@ func (w *webhookPool) GetOrchestrators(numOrchestrators int) ([]*net.Orchestrato
 		return nil, err
 	}
 
-	return w.pool.GetOrchestrators(numOrchestrators)
+	w.mu.Lock()
```
j0sh: I think a RLock would be okay here. Also, a `defer` might be cleaner, since there's no need for an intermediate variable:

```
w.mu.RLock()
defer w.mu.RUnLock()
return w.pool.GetOrchestrators
```

discovery/wh_discovery.go
j0sh: Can `deserializeWebhookJSON` be a regular func now, rather than a variable?

discovery/wh_discovery.go
angyangie: Good idea!

j0sh:
@angyangie Did some local testing and a last pass through the PR [1] and things generally look great. There's a few small issues then we can merge:

[1] Including going through the merge checklist and reviewing the original writeup

* The typo mentioned by @iameli is still present: https://github.com/livepeer/go-livepeer/pull/1007/files#r306931947

* Would be good to add a page in `/doc` on the webhook itself : the CLI args to use and the format of the response to expect.

* The CLI flags still need tests in https://github.com/livepeer/go-livepeer/blob/966f215b16ee7e4ec4e7edada5a99fd6f3b2fbe3/test_args.sh . (This was part of the request in the original writeup ; search for  "including CLI tests"

* Fixup your commits from  into one, since we don't need the earlier DB changes in tree anymore: from https://github.com/livepeer/go-livepeer/pull/1007/commits/29b140649a4d01811ad2d2d0701da69947fefc03 to https://github.com/livepeer/go-livepeer/pull/1007/commits/82b27f7f60ac03f18aea3b2ecbb0a2e659d7dd0b . (Feel free to split the commits along functional lines, eg caching vs testing, but it's not required.) Be sure to leave the earlier commits from me intact, eg from https://github.com/livepeer/go-livepeer/pull/1007/commits/6d0256877a123e2fde0d2c7b2aa08c6e8071ce22 to https://github.com/livepeer/go-livepeer/pull/1007/commits/34337825a3cfda22abba8b534c8f204524da68a4 since those are already functionally separate.

* Did some testing locally and the issue in https://github.com/livepeer/go-livepeer/pull/1007#discussion_r323540172 is still present, and is actually the root of an issue that originally mentioned in the writeup: search the page for "Refresh on empty responses " . I'm OK with just filing a ticket so we keep track of this, since I'm not yet entirely comfortable with some of the implications of fixing this, eg around DB lookups.

j0sh (commented):
LGTM! 🚢 

Can't approve PRs that I opened, so feel free to approve and merge @angyangie

angyangie (approved):

PR #1008, opened 2019-07-24
Load-based T selection

darkdarkdragon:
Selects remote transcoder based on load.

For every transcoder we keep how much load it has currently, and select transcoder with least load for the next task.

With unit tests.

**Checklist:**

### DIFF
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -144,6 +144,11 @@ func main() {
 		return
 	}
 
+	if *maxSessions <= 0 {
+		glog.Error("-maxSessions must be greater than zero")
+		return
+	}
+
 	type NetworkConfig struct {
 		ethUrl        string
 		ethController string
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -145,7 +145,7 @@ func main() {
 	}
 
 	if *maxSessions <= 0 {
-		glog.Error("-maxSessions must be greater than zero")
+		glog.Fatal("-maxSessions must be greater than zero")
 		return
 	}
 
--- a/core/orch_test.go
+++ b/core/orch_test.go
@@ -169,7 +169,7 @@ func TestManageTranscoders(t *testing.T) {
 
 	assert.NotNil(m.liveTranscoders[strm])
 	assert.Len(m.liveTranscoders, 1)
-	assert.Len(m.remoteTranscoders, 5)
+	assert.Len(m.remoteTranscoders, 1)
 	assert.Equal(1, m.RegisteredTranscodersCount())
 	ti := m.RegisteredTranscodersInfo()
 	assert.Len(ti, 1)
@@ -184,7 +184,7 @@ func TestManageTranscoders(t *testing.T) {
 	assert.NotNil(m.liveTranscoders[strm])
 	assert.NotNil(m.liveTranscoders[strm2])
 	assert.Len(m.liveTranscoders, 2)
-	assert.Len(m.remoteTranscoders, 9)
+	assert.Len(m.remoteTranscoders, 2)
 	assert.Equal(2, m.RegisteredTranscodersCount())
 
 	// test that transcoders are removed from liveTranscoders and remoteTranscoders
@@ -193,7 +193,7 @@ func TestManageTranscoders(t *testing.T) {
 	assert.Nil(m.liveTranscoders[strm])
 	assert.NotNil(m.liveTranscoders[strm2])
 	assert.Len(m.liveTranscoders, 1)
-	assert.Len(m.remoteTranscoders, 9)
+	assert.Len(m.remoteTranscoders, 2)
 	assert.Equal(1, m.RegisteredTranscodersCount())
 
 	m.liveTranscoders[strm2].eof <- struct{}{}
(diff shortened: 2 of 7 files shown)


server/ot_rpc.go
```diff
@@ -179,12 +183,16 @@ func runTranscode(n *core.LivepeerNode, orchAddr string, httpc *http.Client, not
...
 func (h *lphttp) RegisterTranscoder(req *net.RegisterRequest, stream net.Transcoder_RegisterTranscoderServer) error {
 	from := common.GetConnectionAddr(stream.Context())
-	glog.Infof("Got a RegisterTranscoder request from transcoder=%s", from)
+	glog.Infof("Got a RegisterTranscoder request from transcoder=%s capacity=%d", from, req.Capacity)
 
 	if req.Secret != h.orchestrator.TranscoderSecret() {
 		glog.Info(errSecret.Error())
 		return errSecret
 	}
+	if req.Capacity == 0 {
```
j0sh: ` req.Capacity <= 0` ?

cmd/livepeer/livepeer.go
```diff
@@ -144,6 +144,11 @@ func main() {
 		return
 	}
 
+	if *maxSessions <= 0 {
+		glog.Error("-maxSessions must be greater than zero")
```
j0sh: Would be good to add a CLI test to `test_args.sh` for this

core/orch_test.go
```diff
@@ -224,15 +224,23 @@ func TestSelectTranscoder(t *testing.T) {
 
 	assert.NotNil(m.liveTranscoders[strm])
 	assert.NotNil(m.liveTranscoders[strm2])
-	assert.Len(m.remoteTranscoders, 9)
+	assert.Len(m.remoteTranscoders, 2)
 
 	// assert transcoder is returned from selectTranscoder and removed from list
```
j0sh: Comment could probably be updated, since the "removed from list" part isn't relevant anymore

core/orch_test.go
```diff
@@ -224,15 +224,23 @@ func TestSelectTranscoder(t *testing.T) {
...
 	assert.NotNil(m.liveTranscoders[strm])
-	assert.Len(m.remoteTranscoders, 8)
+	assert.Len(m.remoteTranscoders, 2)
+
+	// assert transcoder with less load selected
+	currentTranscoder2 := m.selectTranscoder()
+	assert.Equal(t1, currentTranscoder2)
+	assert.Equal(1, t1.load)
+	m.completeTranscoders(t1)
+	assert.Equal(0, t1.load)
```
j0sh, line 253: Would be good to add another test case that attempts to select transcoders when all are at capacity - `selectTranscoder` should return nil.

core/orch_test.go
```diff
@@ -242,14 +250,18 @@ func TestSelectTranscoder(t *testing.T) {
...
 	_, err := m.Transcode("", nil)
 	assert.Nil(err)
-	assert.Len(m.remoteTranscoders, 4)
+	assert.Len(m.remoteTranscoders, 2)
+	assert.Equal(1, t1.load)
+	m.completeTranscoders(t1)
+	assert.Equal(0, t1.load)
 }
 
 func TestTranscoderManagerTranscoding(t *testing.T) {
```
j0sh, line 277: I think this one is also missing a simple test at the beginning where it attempts to transcode with an empty set of remote transcoders - should return a "No transcoders available" error. (Should have been here earlier before this PR, but would be good to add it anyway.)

core/orchestrator.go
```diff
@@ -90,8 +91,8 @@ func (orch *orchestrator) ServeTranscoder(stream net.Transcoder_RegisterTranscod
 	orch.node.serveTranscoder(stream, capacity)
 }
 
-func (orch *orchestrator) TranscoderResults(tcId int64, res *RemoteTranscoderResult) {
-	orch.node.TranscoderManager.transcoderResults(tcId, res)
+func (orch *orchestrator) TranscoderResults(tcID int64, res *RemoteTranscoderResult) {
```
j0sh, line 94: With these changes, does the `core` package pass the CircleCI linter? Is there a way to lint specific files so we don't regress on this?

j0sh (changes requested):
Fantastic! Just a few small changes.

test_args.sh
```diff
@@ -141,4 +141,13 @@ res=0
 ./livepeer -broadcaster -authWebhookUrl http\\://host/ || res=$?
 [ $res -ne 0 ]
 
+# exit early if maxSessions less or equal to zero
+res=0
+./livepeer -maxSessions -1 || res=$?
```
j0sh: I know that the `-maxSessions` check exits early, but this would fail anyway since the node type is not specified. Maybe add a `-broadcaster` to this and the following test?

j0sh (approved):
Can the commits related to unit testing be fixed up / squashed into one? `git rebase -i` works really well for that. Other than that, LGTM!

core/orchestrator.go
darkdarkdragon: We could lint whole package or separate file. `orchestrator.go` still has many linter errors. I've tried to fix them, but stopped at one that will lead to big changes:
`func NewOrchestrator(n *LivepeerNode) *orchestrator` returns private struct `orchestrator`.
It should be either changed to be public (not good), or we should return interface here - but this interface (`type Orchestrator interface`) is defined in `server` package, and to use it in `core` we should move it to some other package, probably to introduce `model` package.

PR #1010, opened 2019-07-26
Add basic payment metrics reporting

### LINKED ISSUE #936: Add PM ticket information to metrics

Both broadcasters and orchestrators would benefit from aggregated data about non-winning and winning tickets received. 

Add support for sending non-winning and winning ticket data to a metrics server. 

Node operators could just parse logs outputted by the node, but this is cumbersome. Furthermore, aggregating these metrics will be useful for internal network testing.

**Additional context**

yondonfu:

This PR adds basic payment metrics reporting. The following metrics are included:

- `ticket_value_sent`
- `tickets_sent`
- `ticket_value_recv`
- `tickets_recv`
- `winning_tickets_recv`
- `value_redeemed`
- `suggested_gas_price`

A broadcaster will record `ticket_value_sent` and `tickets_sent` after submitting a segment with a payment.

An orchestrator will record `ticket_value_recv` and `tickets_recv` for each valid ticket in a payment received with a segment. It will also record `winning_tickets_recv` for each winning ticket received, `value_redeemed` when it successfully redeems a winning ticket and `suggested_gas_price` whenever it caches a new gas price to use for ticket param/price calculations.

Note: This is not a complete list of payment metrics. We'll likely also want to report error types, but I excluded those errors for now because the metrics included in this PR should already get us started with payment workflow testing.

- Added new metrics to `monitor`
- Recorded metrics when a segment with payment is submitted, a segment with payment is received, a winning ticket is redeemed and when a new gas price is cached. Each of these additions is included in its own separate commit

Added some tests for helpers introduced in `monitor`. The current layout of the `monitor` package makes it a bit difficult to unit test the addition of new metrics - perhaps we can consider options to make unit testing easier separately.

Also tried using PromQL queries within the Grafana dashboard created in a local test-harness deployment to fetch the metrics reported. Graph templates still need to be added though.

Fixes #936 

**Checklist:**

### DIFF
--- a/pm/recipient.go
+++ b/pm/recipient.go
@@ -10,6 +10,7 @@ import (
 	ethcommon "github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/crypto"
 	"github.com/golang/glog"
+	"github.com/livepeer/go-livepeer/monitor"
 	"github.com/pkg/errors"
 )
 
@@ -388,6 +389,12 @@ func (r *recipient) redeemWinningTicket(ticket *Ticket, sig []byte, recipientRan
 	// Wait for transaction to confirm
 	if err := r.broker.CheckTx(tx); err != nil {
 		return err
+	}
+
+	if monitor.Enabled {
+		// TODO(yondonfu): Handle case where < ticket.FaceValue is actually
+		// redeemed i.e. if sender reserve cannot cover the full ticket.FaceValue
+		monitor.ValueRedeemed(ticket.Sender.String(), ticket.FaceValue)
 	}
 
 	return nil
--- a/core/orchestrator.go
+++ b/core/orchestrator.go
@@ -2,7 +2,6 @@ package core
 
 import (
 	"context"
-	"encoding/hex"
 	ogErrors "errors"
 	"fmt"
 	"io/ioutil"
@@ -185,7 +184,7 @@ func (orch *orchestrator) ProcessPayment(payment net.Payment, manifestID Manifes
 	}
 
 	if monitor.Enabled {
-		sender := "0x" + hex.EncodeToString(payment.Sender)
+		sender := ethcommon.BytesToAddress(payment.Sender).String()
 		mid := string(manifestID)
 
 		monitor.TicketValueRecv(sender, mid, totalEV)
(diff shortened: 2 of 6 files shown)


pm/recipient.go
```diff
@@ -390,6 +391,12 @@ func (r *recipient) redeemWinningTicket(ticket *Ticket, sig []byte, recipientRan
 		return err
 	}
 
+	if monitor.Enabled {
+		// TODO(yondonfu): Handle case where < ticket.FaceValue is actually
+		// redeemed i.e. if sender reserve cannot cover the full ticket.FaceValue
+		monitor.ValueRedeemed(ticket.Sender.String(), ticket.FaceValue)
```
darkdarkdragon, line 397: `ticket.Sender.String()` will return value in standard ETH address representation - with different case for each letter. This will not match for other addresses that's formatted using `hex.EncodeToString`.
I think this should be lowercased.

darkdarkdragon (changes requested):
Only one comment, otherwise looks good.

pm/recipient.go
yondonfu: Good point about being consistent about the address format. I actually think we should use the checksummed (mixed case) ETH address since it is commonly used in applications to check for typos in the address - not particularly important this moment since we're just displaying them in a metrics dashboard, but I think it would be nice to stick with standard convention and it could also be helpful if we're ever copy and pasting addresses from the dashboard into an application that performs checksum validation.

The following commits remove usage of `hex.EncodeToString` in favor of a byte slice -> address -> string conversion:

- 633259a
- 69dbec3

darkdarkdragon (approved):
LGTM 🚢

j0sh:
Note that the Ethereum hex converter has some unexpected behavior, which is one of the reasons that the built-in golang hex converter has been used elsewhere in the codebase. See https://github.com/livepeer/go-livepeer/pull/651#discussion_r244902551

yondonfu:
> Note that the Ethereum hex converter has some unexpected behavior, which is one of the reasons that the built-in golang hex converter has been used elsewhere in the codebase.

Right, good reminder. Perhaps a rule of thumb we can follow is to use go-ethereum's address to hex converter when the data could be displayed in a UI to show the checksummed address and for all other cases we can just use the standard Go decoder/encoder in `encoding/hex`?

Also, ugh I forgot to rebase to squash the fixup commits before merging this PR - apologies.

PR #1011, opened 2019-07-27
Fix remote transcoder

darkdarkdragon:
Fix remote transcoder not properly checking error when submitting results
(it was trying to read from nil response and panicked with segment violation)

- 
- 
- 

Added unit tests

**Checklist:**

### DIFF
--- a/server/ot_rpc_test.go
+++ b/server/ot_rpc_test.go
@@ -1,0 +1,93 @@
+package server
+
+import (
+	"bytes"
+	"crypto/tls"
+	"io"
+	"io/ioutil"
+	"math/rand"
+	"mime"
+	"mime/multipart"
+	"net/http"
+	"net/http/httptest"
+	"net/url"
+	"testing"
+
+	"github.com/livepeer/go-livepeer/common"
+	"github.com/livepeer/go-livepeer/core"
+	"github.com/livepeer/go-livepeer/net"
+	"github.com/livepeer/lpms/ffmpeg"
+	"github.com/stretchr/testify/assert"
+)
+
+type stubTranscoder struct {
+	called int
+	fname  string
+}
+
+var testRemoteTranscoderResults = [][]byte{[]byte("body1"), []byte("body2")}
+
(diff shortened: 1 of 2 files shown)


server/ot_rpc_test.go
```diff
@@ -0,0 +1,98 @@
...
+
+var testRemoteTranscoderResults = [][]byte{[]byte("body1"), []byte("body2")}
+
+func (st *stubTranscoder) Transcode(fname string, profiles []ffmpeg.VideoProfile) ([][]byte, error) {
+	st.called++
+	st.fname = fname
+	return testRemoteTranscoderResults, nil
+}
+
+func TestRemoteTranscoder(t *testing.T) {
```
j0sh, line 38: Can some comments be added throughout here describing what is being tested? It's kind of hard to tell by just scanning through the code.

server/ot_rpc_test.go
```diff
@@ -0,0 +1,98 @@
...
+
+	assert := assert.New(t)
+	assert.Nil(nil)
+	notify := &net.NotifySegment{
+		TaskId:   742,
+		Profiles: common.ProfilesToTranscodeOpts(profiles),
+		Url:      "linktomanifest",
+	}
+	tr := &stubTranscoder{}
+	node := &core.LivepeerNode{
```
j0sh: Probably better to use the NewLivepeerNode, eg `node = NewLivepeerNode(nil, tmpdir, nil)` then set any required fields, eg `node.OrchSecret  = ...`

Also, for the tmpdir, I'd suggest actually making a TempDir so it can be cleaned up properly after the test exists; see the various examples around the other test code, eg

https://github.com/livepeer/go-livepeer/blob/c20c0f699cf598093f9ef05b75a7c5c1c28ec088/core/core_test.go#L78-L82

j0sh (changes requested):

j0sh:
Added some change requests, cc @ya7ya

server/ot_rpc_test.go
darkdarkdragon: Fixed
`WorkDir` is not used in this test

j0sh (approved):
LGTM to merge after fixing up the two commits 🚢

PR #1012, opened 2019-07-28
Set timeout for downloading segments

darkdarkdragon:
There was no timeout set for downloading segments. Without timeout, under the load, sometimes TCP connection just stalled, and breaked after 15 minutes with 'connection reset by peer' error. This prevented B to retry segment in time.

Added timeout

Manually

**Checklist:**

### DIFF
--- a/common/util.go
+++ b/common/util.go
@@ -18,6 +18,9 @@ import (
 	"google.golang.org/grpc/peer"
 )
 
+// HTTPTimeout timeout used in HTTP connections between nodes
+const HTTPTimeout = 8 * time.Second
+
 var (
 	ErrParseBigInt = fmt.Errorf("failed to parse big integer")
 	ErrProfile     = fmt.Errorf("failed to parse profile")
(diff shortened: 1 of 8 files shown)


common/model.go
```diff
@@ -0,0 +1,8 @@
+package common
+
+import (
+	"time"
+)
+
+// HTTPTimeout timeout used in HTTP connections between nodes
+const HTTPTimeout = 8 * time.Second
```
j0sh: Can we add this to `common/util.go` rather than define a new file with just this one constant?

j0sh (changes requested):

common/model.go
j0sh: cc @ya7ya

darkdarkdragon: I've added `model.go` to be place to move interfaces like `Orchestrator` into it (in the future), but if it's too early I can move `HTTPTimeout` into `util.go`

darkdarkdragon: Fixed

j0sh (approved):
LGTM after fixing up the two commits together 🚢

PR #1017, opened 2019-08-02
core: allow empty ticketparams

kyriediculous:
In case of an empty payment (`numTickets = 0`, B does not need to send tickets because his balance with O is still sufficient) we always populate the `Sender` and `ExpectedPrice` fields for a payment but `TicketParams` are `nil` this leads to an early error in `ProcessPayment`, while the payment might actually be valid (just without value). 

Since we are now populating the Sender in case of empty payments in https://github.com/livepeer/go-livepeer/pull/1000 and empty TicketParams return `nil` instead of an error we should also revert [this requested change](https://github.com/livepeer/go-livepeer/pull/991#discussion_r306030736) back to its original state. 

`if ok && !acceptableErr.Acceptable())` -> `if !ok || !acceptableErr.Acceptable()`

- in `ProcessPayment` return `nil` error when `TicketParams` are `nil`
- Changed `TestProcessPayment_GivenNoTicketParams_ReturnsError` test to `TestProcessPayment_GivenNoTicketParams_ReturnsNoError_LogsMessage``
- reverted the acceptable error check for processed payments in `serveSegment` back to the original form

Ran unit tests
Ran test-harness

**Checklist:**

### DIFF
--- a/core/orchestrator.go
+++ b/core/orchestrator.go
@@ -101,7 +101,7 @@ func (orch *orchestrator) ProcessPayment(payment net.Payment, manifestID Manifes
 	}
 
 	if payment.TicketParams == nil {
-		return fmt.Errorf("Could not find TicketParams for payment %v", payment)
+		return nil
 	}
 
 	if payment.Sender == nil || len(payment.Sender) == 0 {
--- a/server/segment_rpc.go
+++ b/server/segment_rpc.go
@@ -65,7 +65,7 @@ func (h *lphttp) ServeSegment(w http.ResponseWriter, r *http.Request) {
 	if paymentError := orch.ProcessPayment(payment, segData.ManifestID); paymentError != nil {
 
 		acceptableErr, ok := paymentError.(core.AcceptableError)
-		if ok && !acceptableErr.Acceptable() {
+		if !ok || !acceptableErr.Acceptable() {
 			glog.Errorf("Unacceptable error occured processing payment: %v", paymentError)
 			http.Error(w, paymentError.Error(), http.StatusBadRequest)
 			return
--- a/core/orch_test.go
+++ b/core/orch_test.go
@@ -553,7 +553,7 @@ func TestProcessPayment_GivenNoSender_ReturnsError(t *testing.T) {
 	assert.False(ok)
 }
 
-func TestProcessPayment_GivenNoTicketParams_ReturnsError(t *testing.T) {
+func TestProcessPayment_GivenNoTicketParams_ReturnsNoError(t *testing.T) {
 	n, _ := NewLivepeerNode(nil, "", nil)
 	recipient := new(pm.MockRecipient)
 	n.Recipient = recipient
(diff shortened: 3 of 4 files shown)


core/orchestrator.go
```diff
@@ -101,7 +101,8 @@ func (orch *orchestrator) ProcessPayment(payment net.Payment, manifestID Manifes
 	}
 
 	if payment.TicketParams == nil {
-		return fmt.Errorf("Could not find TicketParams for payment %v", payment)
+		glog.Infof("Could not find TicketParams for payment %v", payment)
```
yondonfu: I feel like we can omit this log to minimize log clutter - WDYT?

server/segment_rpc.go
```diff
@@ -65,7 +65,7 @@ func (h *lphttp) ServeSegment(w http.ResponseWriter, r *http.Request) {
 	if paymentError := orch.ProcessPayment(payment, segData.ManifestID); paymentError != nil {
 
 		acceptableErr, ok := paymentError.(core.AcceptableError)
-		if ok && !acceptableErr.Acceptable() {
+		if !ok || !acceptableErr.Acceptable() {
```
yondonfu, line 68: Do the `ProcessPayment()` unit tests need to be updated to cover this change?

yondonfu (changes requested):

core/orchestrator.go
kyriediculous: Sure!  Do you think it's useful to log when we receive an empty payment though or is that information not that relevant?

server/segment_rpc.go
kyriediculous: I don't think so as the output of this only changes for one error (TicketParams being nil), for that one `TestProcessPayment_GivenNoTicketParams_ReturnsNoError_LogsMessage` has been updated.

core/orchestrator.go
yondonfu: I think it is ok to omit that log right now as we can also infer an empty payment by the absence of any logs for receiving tickets.

server/segment_rpc.go
yondonfu: Misspoke in my first comment - I was actually thinking about the `ServeSegment()` unit tests and how they handle errors returned from `ProcessPayment()`.

At the moment, the [unacceptable ProcessPayment() error](https://github.com/livepeer/go-livepeer/blob/master/server/segment_rpc_test.go#L356) unit test mocks an error that adheres to the `core.AcceptableError` interface and returns false for `Acceptable()`. A useful addition to that test could be to mock an error that *does not* adhere to the `core.AcceptableError` interface.

core/orchestrator.go
kyriediculous: Fixed in 18b4192

server/segment_rpc.go
kyriediculous: Fixed in 18b4192

yondonfu (commented):
Changes look good. Needs a rebase!

kyriediculous:
Rebased!

yondonfu:
Since the single commit in this PR already indicates a package in its message prefix, but now it touches both the `core` and `server` packages could we update the prefix for the commit message to either `core+server` or `core,server`?

kyriediculous:
yeah you're totally right, done !

yondonfu (approved):
LGTM :shipit:

Issue #1002, opened 2019-07-23
HTTP Segment Ingest: Adding duration & other missing metadata to source segment 


angyangie:
Issue originally brought up here: https://github.com/livepeer/go-livepeer/pull/972

**Summary of discussion:**

(Josh) Durations: The duration of the source segment is currently not passed in. The duration is necessary to insert valid timings for the resulting HLS playlist. Other features such as latency-based O selection need the duration.

To solve this, it would be nice to send a HTTP header, eg Content-Duration with the value in seconds. We could also probe each segment but that's not desirable on the broadcaster: it's format-specific and another step in the process.

(Josh) Other Missing Metadata: Might be nice to incorporate the width/height of the source segment as a header. We use this information in the master playlist to describe the dimensions of the source stream.

> Durations / Other Missing Metadata

(Eli) Confirmed this issue — the durations show up as zero in the M3U8, which stymies a variety of HLS players including our own. ffplay shrugs it off but polls for new segments from the manifest relentlessly, many times per second.

Something about passing duration/resolution as HTTP headers rubs me the wrong way, feels like a very ad-hoc protocol, but I don't really have a better solution at the moment. The data is actually present in the incoming manifests from both Wowza and ffmpeg, but I'm not wild about requiring that. Messy.

> The data is actually present in the incoming manifests from both Wowza and ffmpeg

(Josh) Is the data available programmatically if you're able to access the source segments within Wowza? I have a little patch adding this to ffmpeg that should be cleaned up and submitted...

(Angie) Re: missing metadata - are we suggesting adding content-duration to seg := &stream.HLSSegment & width/height? Where exactly would these be used? I see we need the former for valid timings for the resulting HLS playlist ... is that something that happens in LPMS? What about height/width? I also see arguments against requiring content-duration/resolution.

(Eli) Yeah... this is my biggest area of ambivalence in this PR. I see a few possible paths forward:

A. Ship as-is, with the segment lengths for the source stream being listed as zero in the HTTP output.

B. Provide bitrate/duration data in HTTP headers. I like that this is low-overhead and could be implemented quickly. I dislike that it's an ad-hoc, Livepeer-specific protocol that won't be followed by any "HLS Push" except for LivepeerWowza's implementation.

C. Implement ffprobe-style discovery of bitrate/duration. This would definitely work but is really unfortunate overhead for the broadcaster to incur. If we go this route, it should probably enabled/disabled through a CLI flag.

[...]

PR #1019, opened 2019-08-05
Ar/add duration resolution to hls segment metadata

angyangie:
https://github.com/livepeer/go-livepeer/issues/1002

Devtool script
Tests

https://github.com/livepeer/go-livepeer/issues/1002

**Checklist:**

### DIFF
--- a/server/mediaserver.go
+++ b/server/mediaserver.go
@@ -10,7 +10,6 @@ import (
 	"errors"
 	"fmt"
 	"io/ioutil"
-	"math"
 	"math/big"
 	"math/rand"
 	"net/http"
--- a/server/mediaserver.go
+++ b/server/mediaserver.go
@@ -596,7 +596,7 @@ func (s *LivepeerServer) HandlePush(w http.ResponseWriter, r *http.Request) {
 
 	duration, err := strconv.Atoi(r.Header.Get("Content-Duration"))
 	if err != nil {
-		duration = 2000 // maybe do some reasonable default rather than zero?
+		duration = 2000
 		glog.Info("Missing duration; filling in a default of 2000ms")
 	}
 
--- a/server/mediaserver.go
+++ b/server/mediaserver.go
@@ -10,6 +10,7 @@ import (
 	"errors"
 	"fmt"
 	"io/ioutil"
+	"math"
 	"math/big"
 	"math/rand"
 	"net/http"
@@ -258,7 +259,7 @@ func streamParams(rtmpStrm stream.RTMPVideoStream) *streamParameters {
 func gotRTMPStreamHandler(s *LivepeerServer) func(url *url.URL, rtmpStrm stream.RTMPVideoStream) (err error) {
 	return func(url *url.URL, rtmpStrm stream.RTMPVideoStream) (err error) {
 
-		cxn, err := s.registerConnection(rtmpStrm)
+		cxn, err := s.registerConnection(rtmpStrm, "")
 		if err != nil {
 			return err
 		}
@@ -325,7 +326,7 @@ func endRTMPStreamHandler(s *LivepeerServer) func(url *url.URL, rtmpStrm stream.
 	}
 }
 
-func (s *LivepeerServer) registerConnection(rtmpStrm stream.RTMPVideoStream) (*rtmpConnection, error) {
+func (s *LivepeerServer) registerConnection(rtmpStrm stream.RTMPVideoStream, resolution string) (*rtmpConnection, error) {
 	nonce := rand.Uint64()
 
 	// If running in on-chain mode, check for a reasonable deposit
@@ -358,7 +359,10 @@ func (s *LivepeerServer) registerConnection(rtmpStrm stream.RTMPVideoStream) (*r
 	}
 	storage := drivers.NodeStorage.NewSession(string(mid))
 	// Build the source video profile from the RTMP stream.
-	resolution := fmt.Sprintf("%vx%v", rtmpStrm.Width(), rtmpStrm.Height())
+	// Build the source video profile from the RTMP stream.
+	if resolution == "" {
+		resolution = fmt.Sprintf("%vx%v", rtmpStrm.Width(), rtmpStrm.Height())
+	}
 	vProfile := ffmpeg.VideoProfile{
 		Name:       "source",
 		Resolution: resolution,
@@ -559,7 +563,8 @@ func (s *LivepeerServer) HandlePush(w http.ResponseWriter, r *http.Request) {
 			return
 		}
 		st := stream.NewBasicRTMPVideoStream(appData)
-		cxn, err = s.registerConnection(st)
+		resolution := r.Header.Get("Content-Resolution")
+		cxn, err = s.registerConnection(st, resolution)
 		if err != nil {
 			http.Error(w, err.Error(), http.StatusInternalServerError)
 			return
@@ -587,11 +592,18 @@ func (s *LivepeerServer) HandlePush(w http.ResponseWriter, r *http.Request) {
 	if err != nil {
 		seq = 0
 	}
+
+	duration, err := strconv.ParseFloat(r.Header.Get("Content-Duration"), 64)
+	duration = math.Floor(duration * 1000)
+	if err != nil {
+		duration = 0
+	}
+
 	seg := &stream.HLSSegment{
-		Data:  body,
-		Name:  fname,
-		SeqNo: seq,
-		// TODO duration
+		Data:     body,
+		Name:     fname,
+		SeqNo:    seq,
+		Duration: duration,
 	}
 
 	// Do the transcoding!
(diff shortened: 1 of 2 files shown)


angyangie:
Haha I assumed when we said we wanted milliseconds, assumed the input comes in in seconds, so multiplying made sense (1 s = 1000 ms). What's the input coming in as? We could just not convert to ms if that doesn't make sense?

server/mediaserver.go
```diff
@@ -587,11 +592,18 @@ func (s *LivepeerServer) HandlePush(w http.ResponseWriter, r *http.Request) {
 	if err != nil {
 		seq = 0
 	}
+
+	duration, err := strconv.ParseFloat(r.Header.Get("Content-Duration"), 64)
```
j0sh: The idea of was to pass in an integer number of milliseconds, then convert to float as necessary at the last minute.  Something like:

```golang
duration, err := strconv.Atoi(r.Header.Get("Content-Duration"))
if err != nil {
  duration = 2000 // maybe do some reasonable default rather than zero?
  glog.Info("MIssing duration; filling in a default of 2000ms")
}
seg := &stream.HLSSegment {
   ....
   Duration: float64(duration)/1000.0,
}

File: server/mediaserver.go [hunk 2]
```diff
@@ -559,7 +563,8 @@ func (s *LivepeerServer) HandlePush(w http.ResponseWriter, r *http.Request) {
 			return
 		}
 		st := stream.NewBasicRTMPVideoStream(appData)
-		cxn, err = s.registerConnection(st)
+		resolution := r.Header.Get("Content-Resolution")
+		cxn, err = s.registerConnection(st, resolution)
```
j0sh: A less invasive way to do this might be to add a `resolution` field within the streamParameters struct and using that instead: Something like:

1. Add a `resolution` field to the `streamParameters` struct
2. Set the resolution unconditionally within `createRTMPStreamIDHandler` . This might end up being `0x0` for HLS push. That's OK.
3. Within the HLS push function, re-assign `appData.resolution = r.Header.Get("Content-Resolution")` after calling `createRTMPStreamIDHandler`
4. Within `registerConnection`, set `Resolution: appData.Resolution`

How does that sound?

The LPMS stream API is starting to feel a bit clunky (ideally we'd be able to slot in different stream types, and hence supply our own `Width` / `Height` accessors) but this should be OK for now.

server/mediaserver.go
angyangie: I'm fine with that. Why does it make a difference to convert at the last minute, rather than when first extracted from `Content-Duration`?

server/mediaserver.go
angyangie: That does sound a lot cleaner! However, wouldn't we need something like `appData.SetResolution(r.Header.Get("Content-Resolution"))` to set the resolution on `streamParameters`? The `AppData` interface is defined in `lmps > stream > interface.go`, so we'd have to add a `SetResolution(string)` method there to access the `streamParameters` property, which would require LPMS editing. Is that something we'd want to do? 
```golang
func (s *streamParameters) SetResolution(resolution string) {
	s.resolution = resolution
}
```

server/mediaserver.go
j0sh: If we ever need to use the value somewhere else, it's more helpful to maintain in integers rather than float.

server/mediaserver.go
angyangie: We'd also need this: 
```golang
func (s *streamParameters) Resolution() string {
	return s.resolution
}
```

j0sh: Reify the appData into a concrete type via `streamParams` and use that:

https://github.com/livepeer/go-livepeer/blob/f5106de234ff99f26916f6965a0f8fe41e789f25/server/mediaserver.go#L245-L253

```golang
appData := (createRTMPStreamIDHandler(s))(r.URL)
params := streamParams(st)
params.resolution = r.Header.Get("Content-Resolution")
cxn, err := s.registerConnection(st)
...
```

angyangie: OH I see. That's a much cleaner way of doing it. Thanks @j0sh !

server/mediaserver.go
```diff
@@ -358,10 +360,10 @@ func (s *LivepeerServer) registerConnection(rtmpStrm stream.RTMPVideoStream) (*r
 	}
 	storage := drivers.NodeStorage.NewSession(string(mid))
 	// Build the source video profile from the RTMP stream.
-	resolution := fmt.Sprintf("%vx%v", rtmpStrm.Width(), rtmpStrm.Height())
+	// Build the source video profile from the RTMP stream.
```
j0sh: I think this will break with RTMP ingest

server/mediaserver.go
```diff
@@ -204,9 +205,10 @@ func createRTMPStreamIDHandler(s *LivepeerServer) func(url *url.URL) (strmID str
 			key = common.RandomIDGenerator(StreamKeyBytes)
 		}
 		return &streamParameters{
-			mid:      mid,
-			rtmpKey:  key,
-			profiles: presets,
+			mid:        mid,
+			rtmpKey:    key,
+			profiles:   presets,
+			resolution: "0x0", // re-assigned to an actual value for RTMP Push only
```
j0sh: This can probably be omitted

server/mediaserver.go
angyangie: What part? I'm testing locally and things seem to work fine.

j0sh: Is the resolution in the master playlist actually populated with RTMP?

angyangie: I'll check. Is the resolution we're looking for populated here?
```golang
playlist := core.NewBasicPlaylistManager(mid, storage)
masterPL := playlist.GetHLSMasterPlaylist()
variants := masterPL.Variants
for i, v := range variants {
       resolution := v.Resolution
       glog.Error("resolution: ", resolution)
}
```

j0sh: Yeah, or `curl http://localhost:8935/stream/mid.m3u8` to see the resolution as part of the m3u8 header while you're streaming.

angyangie: I've added the resolution back like this and all works:
``` golang
if params.resolution == "" {
       params.resolution = fmt.Sprintf("%vx%v", rtmpStrm.Width(), rtmpStrm.Height())
}
```

darkdarkdragon:
@j0sh @angyangie Instead of passing resolution in headers, we could parse segment using `joy4` and get that data from segment itself. What do you think?

server/mediaserver.go
j0sh: Yep. Thought we would have access to the `rtmpStrm` from within the `createStreamID` callback but that's not the case, so this line is needed. 👍

darkdarkdragon:
And probably we could get segment's duration using joy4 too.

j0sh:
> Instead of passing resolution in headers, we could parse segment using joy4 and get that data from segment itself. What do you think?

@darkdarkdragon That would work if Wowza consistently includes the sps/pps extradata at the beginning of each segment, or at least for the very first segment that it transmits to the broadcaster. Might be worth checking.

Edit: The only thing that I don't like about this is that it bakes in an assumption that H.264 is coming in. Any time we want to extend support for additional codecs, we need to update this part.

> And probably we could get segment's duration using joy4 too.

Does this involve taking a full pass over the data? If so, I'd avoid doing that.

darkdarkdragon:
@j0sh 
> Does this involve taking a full pass over the data? If so, I'd avoid doing that.

Yes, it requires to through all the packets. So no go for now. But if we would be able to enhance joy4 parser to go through the h264 data and count pixels in it, then we could get both pixels count and duration in one iteration.

angyangie:
@darkdarkdragon does that mean you're okay with us incorporating the `joy4` segment parsing at a later time?

darkdarkdragon:
@angyangie I'm even ok with adding it now :) For parsing resolution.
@j0sh 
> The only thing that I don't like about this is that it bakes in an assumption that H.264 is coming in. Any time we want to extend support for additional codecs, we need to update this part.

I agree with you here, but I would still prefer to be less reliant on external data (resolution in headers) and get it by ourselves (through parsing).

j0sh (approved):
LGTM if the commits are squashed / fixed up.

j0sh:
Can unit tests be added that checks the duration / resolution?

j0sh:
> prefer to be less reliant on external data (resolution in headers) and get it by ourselves (through parsing).

@darkdarkdragon We could parse if the headers are not present. But if there's a way for the caller to supply the information such that we don't need to do additional processing on the data, then that's preferable IMO. (That can be added later; don't think it should block this PR.)

darkdarkdragon:
@j0sh 
> We could parse if the headers are not present.

Totally agree here.

server/mediaserver_test.go
```diff
@@ -980,3 +980,23 @@ func TestHandlePushForAuthWebhookFailure(t *testing.T) {
 	// reset AuthWebhookURL to original value
 	AuthWebhookURL = ""
 }
+
+func TestResolutionInRegisterConnection(t *testing.T) {
```
j0sh, line 984: Missed this bit because it was merged right away without review. Can this test be adjusted to pass in the resolution via HTTP headers?

PR #1021, opened 2019-08-06
core: fix sendernonce race condition

### LINKED ISSUE #1014: Periodic winning ticket redemption transaction failure
**Describe the bug**

I observed periodic winning ticket redemption transaction failure when doing e2e testing with a test-harness deployment. Some of the redemption transactions succeeded while others failed.

**To Reproduce**
1. Clone the latest master branch for the test-harness
2. In the go-livepeer codebase, change [this line](https://github.com/livepeer/go-livepeer/blob/master/core/orchestrator.go#L506) to 900000000.
3. Run `make localdocker` in the go-livepeer directory
4. In the test-harness directory, run `node examples/local.js` and wait for the setup to complete
5. Make sure you have a `geth` binary installed locally
6. Run `geth attach http://localhost:8545`
7. In the interactive console, run:

```
web3.eth.sendTransaction({from: web3.eth.accounts[0], to: web3.eth.accounts[0], value: 0, gasPrice: 400})
```

multiple times.
8. Watch the logs of the orchestrator nodes until you observe a message that indicates that the new gas price of 400 has been cached
9. Start a stream by running `./test-harness stream test123`
10. Watch the logs of the orchestrators nodes and observe some winning ticket redemption transactions succeeding and some failing

**Expected behavior**

Winning ticket redemption transactions should not fail.

**Screenshots**

Screen Shot 2019-07-30 at 7 01 51 PM

Screen Shot 2019-07-30 at 7 02 03 PM

 - OS: OSX
 - Version: 0c1a6d0 (this is the starting HEAD commit, but see the reproduction steps for a note on the modification made to the code before building)

kyriediculous:
This PR fixes a race condition in `ProcessPayment` where the ticket pointer in the redemption for-loop was overwritten with an already incremented `SenderNonce` prematurely. This resulted in a signature validation failure in the TicketBroker contracts when trying to redeem the ticket on chain.

- Allocated a new instance of `Ticket` for each iteration of the for loop

ran unit tests and test-harness
image

Fixes #1014 

**Checklist:**

### DIFF
--- a/core/orchestrator.go
+++ b/core/orchestrator.go
@@ -134,12 +134,15 @@ func (orch *orchestrator) ProcessPayment(payment net.Payment, manifestID Manifes
 
 	seed := new(big.Int).SetBytes(payment.TicketParams.Seed)
 
-	ticket := &pm.Ticket{
-		Recipient:              ethcommon.BytesToAddress(payment.TicketParams.Recipient),
-		Sender:                 ethcommon.BytesToAddress(payment.Sender),
-		FaceValue:              new(big.Int).SetBytes(payment.TicketParams.FaceValue),
-		WinProb:                new(big.Int).SetBytes(payment.TicketParams.WinProb),
-		RecipientRandHash:      ethcommon.BytesToHash(payment.TicketParams.RecipientRandHash),
+	ticketParams := &pm.TicketParams{
+		Recipient:         ethcommon.BytesToAddress(payment.TicketParams.Recipient),
+		FaceValue:         new(big.Int).SetBytes(payment.TicketParams.FaceValue),
+		WinProb:           new(big.Int).SetBytes(payment.TicketParams.WinProb),
+		RecipientRandHash: ethcommon.BytesToHash(payment.TicketParams.RecipientRandHash),
+		Seed:              seed,
+	}
+
+	ticketExpirationParams := &pm.TicketExpirationParams{
 		CreationRound:          payment.ExpirationParams.CreationRound,
 		CreationRoundBlockHash: ethcommon.BytesToHash(payment.ExpirationParams.CreationRoundBlockHash),
 	}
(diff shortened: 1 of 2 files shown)


core/orchestrator.go
```diff
@@ -150,12 +150,14 @@ func (orch *orchestrator) ProcessPayment(payment net.Payment, manifestID Manifes
 
 	for _, tsp := range payment.TicketSenderParams {
 
+		// shadow ticket declaration & allocate new instance to prevent SenderNonce race conditions
+		ticket := *ticket
```
yondonfu: A more explicit (and IMO a more readable) approach that doesn't require copying all the ticket fields (though technically, only the struct fields that are values are copied and not the struct fields that are pointers)  in each iteration is to use the `pm.NewTicket()` constructor within the loop.

Ex.

```
ticketParams := &pm.TicketParams{
    // Initialize using payment.TicketParams
}

expirationParams := &pm.TicketExpirationParams{
    // Initialize using payment.ExpirationParams
}

for _, tsp := range payment.TicketSenderParams {
    ticket := pm.NewTicket(ticketParams, expirationParams, sender, tsp.SenderNonce)
    ...
}
```

In the above code snippet, we can create the `ticketParams` and `expirationParams` pointers once and then each iteration of the loop will create a ticket that references the same `ticketParams` and `expirationParams`, but uses a different `senderNonce`.

Something to make note of is that if multiple goroutines are created by the loop then a data race could be possible if we ever modify the `ticketParams` and/or `expirationParams` pointers. There is a similar concern if we took the approach of copying the ticket in each iteration of the loop except the data race could be around `ticket.FaceValue` and `ticket.WinProb` since those are the pointers shared by each ticket copy. One way to completely eliminate the possibility of data races is to create a fresh ticket in each iteration of the loop that does not share *any* pointers with other tickets. But, at the moment since we're never modifying `ticketParams` or `expirationParams` after creating them, we should be ok.

yondonfu (changes requested):
Can we add a test demonstrating that the fix in this PR works (and the test should fail if we omit the fix)?

We could update [this test](https://github.com/livepeer/go-livepeer/blob/master/core/orch_test.go#L667) to also check that `RedeemWinningTicket()` is always called with the correct `senderNonce`

core/orchestrator.go
kyriediculous: Is a9075d4 this something you were thinking of in terms of testing or how to better approach this case?

yondonfu: I was just highlighting something that I think is worth being aware of. I don't think we need to take further action for now.

kyriediculous:
Fixed in ee7f4de

Seems like circle is failing but both `go test ./...`  and `./test.sh` pass locally.

yondonfu (commented):
Seems to be a flaky test - rerunning the workflow. Changes look good though. Needs a rebase.

kyriediculous:
rebased !

yondonfu (approved):
LGTM 🚢

PR #1022, opened 2019-08-06
Fix insufficient balance on first segment

### LINKED ISSUE #1001: Insufficient balance error on first segment
**Describe the bug**

When B sends the first segment (with a payment) to O, it receives an error and O logs an insufficient balance error.

**To Reproduce**
1. Run a local on-chain test-harness deployment with 1 O and 1 B
2. Make sure that B has giant deposit & reserve - let's say 1000 ETH 
3. Stream into B
4. Observe insufficient balance error for the first segment in O's logs

**Expected behavior**

B should not encounter an insufficient balance error for the first segment because it should have calculated the necessary payment to top off its balance to meet O's minimum credit requirement.

**Screenshots**

Screen Shot 2019-07-23 at 5 04 51 PM

I added some additional log messages to display B's balance and O's minimum credit requirement when the insufficient balance error is logged.

 - OS: OSX (running test-harness deployment)

kyriediculous:
This PR fixes the insufficient balance error for a broadcaster on the first segment. This is because the calculated winning probability used for creating tickets to send to O is rounded slightly down resulting in a  ticket EV that is 1 unit of value short to be equal to which it is compared. 

- Round up winprob if `maxWinProb % faceValue` is greater than 0 
- Adjusted unit tests 

Adjusted unit tests, ran test harness

Fixes #1001 

**Alternative considered**
Since the ticket EV always seemed to be short by 1 wei, the easy alternative was to increment each ticket's EV by 1 wei by adding `1` to the value returned by the underlying function in `ticket.go`

```
func ticketEV(faceValue *big.Int, winProb *big.Int) *big.Rat {
	return new(big.Rat).Mul(new(big.Rat).SetInt(faceValue), new(big.Rat).SetFrac(winProb, maxWinProb))
}
````

**Checklist:**

### DIFF
--- a/pm/recipient.go
+++ b/pm/recipient.go
@@ -254,8 +254,11 @@ func (r *recipient) winProb(faceValue *big.Int) *big.Int {
 		return maxWinProb
 	}
 
-	x := new(big.Int).Div(maxWinProb, faceValue)
-
+	m := new(big.Int)
+	x, m := new(big.Int).DivMod(maxWinProb, faceValue, m)
+	if m.Int64() != 0 {
+		return new(big.Int).Mul(r.cfg.EV, x.Add(x, big.NewInt(1)))
+	}
 	// Compute winProb as the numerator of a fraction over maxWinProb
 	return new(big.Int).Mul(r.cfg.EV, x)
 }
--- a/pm/recipient_test.go
+++ b/pm/recipient_test.go
@@ -1142,7 +1142,7 @@ func TestTicketParams(t *testing.T) {
 		t.Errorf("expected faceValue %d got %d", faceValue, params1.FaceValue)
 	}
 
-	winProb, _ := new(big.Int).SetString("5789604461865809771178549250434395392663499233282028201972879200395655", 10)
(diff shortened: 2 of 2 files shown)


yondonfu (commented):
Changes look good to me.

Small comment re: commit structure

Not sure if you were planning on rebasing before merging, but when possible, I like to include the unit tests for a change in the same commit as the change so that anyone that is inspecting the commit history in the future can easily look at the relevant tests when looking at a change. This also helps with trying to make sure that commits result in a passing build/test suite. Let me know what you think about this convention.

kyriediculous:
Yeah you're absolutely right, I just didn't realize I already made a first commit (you can tell by the similarity of the messages). I was planning on squashing these regardless.

yondonfu:
> Yeah you're absolutely right, I just didn't realize I already made a first commit (you can tell by the similarity of the messages). I was planning on squashing these regardless.

👍 

In that case, needs a rebase and then this PR should be good to go.

kyriediculous:
Rebased !

yondonfu (approved):
LGTM :shipit:

PR #1030, opened 2019-08-06
blockwatch: Add stack

### LINKED ISSUE #1024: blockwatch: Stack implementation

A block watcher will maintain an internal stack of block headers with the latest latest header at the top of the stack and the oldest header at the bottom of the stack. As the block watcher polls for new blocks, new headers will be pushed to the stack and when the stack is full the oldest header will be evicted. If an re-org occurs, the block watcher can pop headers off the stack until it finds the header of the start block for the re-org - then, the block watcher can push the headers of the new blocks included as a part of the re-org onto the stack.

The implementation can be largely based off of: https://github.com/0xProject/0x-mesh/blob/development/ethereum/blockwatch/stack.go

A few changes will need to be made:

In order to remove the `meshdb` dependency, we can define the `MiniHeader` type in the package and swap the `*meshdb.MeshDB` type in favor of a `MiniHeaderStore` interface:

```
type MiniHeaderStore interface {
    FindLatestMiniHeader() (*MiniHeader, error)
    FindAllMiniHeadersSortedbyNumber() ([]*MiniHeader, error)
    Insert(header *MiniHeader) error
    Delete(hash common.Hash) error
}
```

yondonfu:

This PR introduces the `blockwatch` package which will largely be based off of the [0x blockwatch package](https://github.com/0xProject/0x-mesh/tree/development/ethereum/blockwatch) and also adds a stack implementation to track block headers along with additional unit tests.

- Port https://github.com/0xProject/0x-mesh/blob/development/ethereum/blockwatch/stack.go
- Introduce the `MiniHeaderStore` interface in the package which allows the stack to be agnostic to how the block headers are stored (i.e. could be any DB, in-memory cache, etc.)
- Move the `MiniHeader` type definition into the package
- Added unit test coverage for `Stack`
- Added concurrency tests to confirm that the mutex locking/unlocking in `Push()` and `Pop()` are working correctly

Added unit tests. Ran `go test -race` in `eth/blockwatch`.

Fixes #1024

**Checklist:**

### DIFF
--- a/eth/blockwatch/stack_test.go
+++ b/eth/blockwatch/stack_test.go
@@ -1,0 +1,214 @@
+package blockwatch
+
+import (
+	"errors"
+	"testing"
+
+	ethcommon "github.com/ethereum/go-ethereum/common"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+type stubMiniHeaderStore struct {
+	headers           []*MiniHeader
+	latestErr         error
+	sortedByNumberErr error
+	insertErr         error
+	deleteErr         error
+}
+
+func (s *stubMiniHeaderStore) FindLatestMiniHeader() (*MiniHeader, error) {
+	if s.latestErr != nil {
+		return nil, s.latestErr
+	}
+
+	if len(s.headers) == 0 {
+		return nil, nil
+	}
+
+	return s.headers[len(s.headers)-1], nil
+}
+
+func (s *stubMiniHeaderStore) FindAllMiniHeadersSortedByNumber() ([]*MiniHeader, error) {
+	if s.sortedByNumberErr != nil {
+		return nil, s.sortedByNumberErr
+	}
+
+	return s.headers, nil
+}
+
+func (s *stubMiniHeaderStore) InsertMiniHeader(header *MiniHeader) error {
+	if s.insertErr != nil {
+		return s.insertErr
+	}
+
+	s.headers = append(s.headers, header)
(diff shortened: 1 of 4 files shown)


eth/blockwatch/stack_test.go
```diff
@@ -0,0 +1,317 @@
...
+	_, err := stack.Pop()
+	assert.EqualError(err, store.latestErr.Error())
+
+	// Test when store returns nil MiniHeader
+	store.latestErr = nil
+	header, err := stack.Pop()
+	assert.Nil(header)
+	assert.Nil(err)
+
+	// Test when store.DeleteMiniHeader() returns error
```
kyriediculous: Just a detail but since `DeleteMiniHeader` is a stub where we error simply if `store.deleteErr` is set on the `store` struct I would place this comment at line 90 where we are actually setting `store.deleteErr`. The current placement of the comment where we are simply adding a first header to the store had me confused for a second as that doesn't lead to an error being returned.

kyriediculous (changes requested):
Looks good! I like how the tests are done. Just one minor comment.

eth/blockwatch/stack_test.go
yondonfu: Fixed in 9b746c0

kyriediculous:
Looks good, needs a rebase before merging

yondonfu:
Rebased!

kyriediculous (approved):
🚀 🚀 🚀

Issue #1004, opened 2019-07-23
HTTP segment ingest: path prefix and port numbering


angyangie:
Original discussion here: https://github.com/livepeer/go-livepeer/pull/972

(Josh) Path Prefix: Currently /live/ , as opposed to the proposed /stream/. LPMS hijacks the /stream/ prefix so it's easier to just use a new prefix. Transcoded HLS is still available for playback under /stream/ as before.

(Josh) Port numbering: Currently use port 7935, mostly out of expediency - the B doesn't pass in a custom mux that can be shared between LPMS and the goclient.

We have an internal CLI API running on port 7935 that is mostly used for node management and diagnostics. An "external" port lives on 8935. The broadcaster uses the external port for HLS playback. (O uses 8935 for gRPC and segment handling as well).

We may want to switch the /live/ endpoint to port 8935. One bit of concern here is relating to "safe defaults". Historically, we have attempted to keep all potentially Eth-draining operations segregated from the port on 8935, which is considered public. Since there is no authentication by default, a B operator may unintentionally open itself up to drive-by transcoding.

One mitigating factor is that the broadcaster listens to 8935 on localhost by default. (We do the same for RTMP.) Another layer of safety would be to have a flag to toggle the push API. For an improved UX, this flag could be enabled by default. It would be disabled by default if 8935 is listening to non-localhost and webhook auth is not enabled. Enabling the flag under those conditions would be an explicit signal - "I know this is potentially an expensive choice, do it anyway."

(Eli) I think your defaults are reasonable, but I kind of also want the ability to run the HTTP ingest on a separate port entirely. I know it seems like we're getting up to an unreasonable number of HTTP servers, but I like the idea of RTMP and HTTP ingest having the same security story — listening on localhost by default, optionally enhanced with a webhook. There are lots of current hobbyist workflows that involve setting up a local server, exposing 8935 publicly, and streaming OBS to localhost over RTMP — it'd be a shame not to be able to implement that same pattern using HTTP without a webhook.

> run the HTTP ingest on a separate port entirely

(Josh) Yeah, that's a potential use case. I'm torn, because I also like minimizing the exposed surface area on the node. Curious what others think.

(Angie) HTTP ingest - arguments were made for port 7935 (current setup), for 8935 and using a flag for push API, disabling it automatically if there is no webhook auth present/non-localhost, and setting it up on a different port entirely. We should get consensus around this!

[...]

PR #1036, opened 2019-08-08
Ar/http push path prefix and numbering

angyangie:
https://github.com/livepeer/go-livepeer/issues/1004

### DIFF
--- a/server/mediaserver.go
+++ b/server/mediaserver.go
@@ -540,6 +540,7 @@ func (s *LivepeerServer) HandlePush(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	r.Body.Close()
+	r.URL = &url.URL{Scheme: "http", Host: r.Host, Path: r.URL.Path}
 
 	if ".ts" != path.Ext(r.URL.Path) {
 		// ffmpeg sends us a m3u8 as well, so ignore
--- a/server/mediaserver.go
+++ b/server/mediaserver.go
@@ -82,12 +82,12 @@ type rtmpConnection struct {
 }
 
 type LivepeerServer struct {
-	RTMPSegmenter lpmscore.RTMPSegmenter
-	LPMS          *lpmscore.LPMS
-	LivepeerNode  *core.LivepeerNode
-	HTTPMux       *http.ServeMux
-
+	RTMPSegmenter         lpmscore.RTMPSegmenter
+	LPMS                  *lpmscore.LPMS
+	LivepeerNode          *core.LivepeerNode
+	HTTPMux               *http.ServeMux
 	ExposeCurrentManifest bool
+	httpAddr              string
 
 	// Thread sensitive fields. All accesses to the
 	// following fields should be protected by `connectionLock`
@@ -105,21 +105,27 @@ type authWebhookResponse struct {
 
 func NewLivepeerServer(rtmpAddr string, httpAddr string, lpNode *core.LivepeerNode) *LivepeerServer {
 	opts := lpmscore.LPMSOpts{
-		RtmpAddr: rtmpAddr, RtmpDisabled: true,
-		HttpAddr: httpAddr,
-		WorkDir:  lpNode.WorkDir,
+		RtmpAddr:     rtmpAddr,
+		RtmpDisabled: true,
+		HttpAddr:     httpAddr,
+		WorkDir:      lpNode.WorkDir,
+		HttpMux:      http.NewServeMux(),
 	}
 	switch lpNode.NodeType {
 	case core.BroadcasterNode:
 		opts.RtmpDisabled = false
-	case core.OrchestratorNode:
-		opts.HttpMux = http.NewServeMux()
 	}
 	server := lpmscore.New(&opts)
-	return &LivepeerServer{RTMPSegmenter: server, LPMS: server, LivepeerNode: lpNode, HTTPMux: opts.HttpMux, connectionLock: &sync.RWMutex{}, rtmpConnections: make(map[core.ManifestID]*rtmpConnection)}
+	ls := &LivepeerServer{RTMPSegmenter: server, LPMS: server, LivepeerNode: lpNode, HTTPMux: opts.HttpMux, connectionLock: &sync.RWMutex{},
+		rtmpConnections: make(map[core.ManifestID]*rtmpConnection), httpAddr: httpAddr,
+	}
+	if lpNode.NodeType == core.BroadcasterNode {
+		opts.HttpMux.HandleFunc("/live/", ls.HandlePush)
+	}
+	return ls
 }
 
-//StartServer starts the LPMS server
+//StartMediaServer starts the LPMS server
 func (s *LivepeerServer) StartMediaServer(ctx context.Context, transcodingOptions string) error {
 	BroadcastJobVideoProfiles = parsePresets(strings.Split(transcodingOptions, ","))
 
@@ -130,12 +136,12 @@ func (s *LivepeerServer) StartMediaServer(ctx context.Context, transcodingOption
 	s.LPMS.HandleRTMPPlay(getRTMPStreamHandler(s))
 
 	//LPMS hanlder for handling HLS video play
-	s.LPMS.HandleHLSPlay(getHLSMasterPlaylistHandler(s), getHLSMediaPlaylistHandler(s), getHLSSegmentHandler(s), getSegmentPushHandler(s))
+	s.LPMS.HandleHLSPlay(getHLSMasterPlaylistHandler(s), getHLSMediaPlaylistHandler(s), getHLSSegmentHandler(s))
 
 	//Start the LPMS server
 	lpmsCtx, cancel := context.WithCancel(context.Background())
 
-	ec := make(chan error, 1)
+	ec := make(chan error, 2)
 	go func() {
 		if err := s.LPMS.Start(lpmsCtx); err != nil {
 			// typically triggered if there's an error with broadcaster LPMS
@@ -143,6 +149,12 @@ func (s *LivepeerServer) StartMediaServer(ctx context.Context, transcodingOption
 			ec <- s.LPMS.Start(lpmsCtx)
 		}
 	}()
+	if s.LivepeerNode.NodeType == core.BroadcasterNode {
+		go func() {
+			glog.V(4).Infof("HTTP Server listening on http://%v", s.httpAddr)
+			ec <- http.ListenAndServe(s.httpAddr, s.HTTPMux)
+		}()
+	}
 
 	select {
 	case err := <-ec:
@@ -152,12 +164,6 @@ func (s *LivepeerServer) StartMediaServer(ctx context.Context, transcodingOption
 	case <-ctx.Done():
 		cancel()
 		return ctx.Err()
-	}
-}
-
-func getSegmentPushHandler(s *LivepeerServer) func(w http.ResponseWriter, r *http.Request) {
-	return func(w http.ResponseWriter, r *http.Request) {
-		s.HandlePush(w, r)
 	}
 }
 
--- a/server/mediaserver.go
+++ b/server/mediaserver.go
@@ -102,11 +102,10 @@ type authWebhookResponse struct {
 	Presets    []string `json:"presets"`
 }
 
-func NewLivepeerServer(rtmpAddr string, httpAddr string, lpNode *core.LivepeerNode) *LivepeerServer {
+func NewLivepeerServer(rtmpAddr string, lpNode *core.LivepeerNode) *LivepeerServer {
 	opts := lpmscore.LPMSOpts{
 		RtmpAddr:     rtmpAddr,
 		RtmpDisabled: true,
-		HttpAddr:     httpAddr,
 		WorkDir:      lpNode.WorkDir,
 		HttpMux:      http.NewServeMux(),
 	}
--- a/server/mediaserver.go
+++ b/server/mediaserver.go
@@ -130,10 +130,11 @@ func (s *LivepeerServer) StartMediaServer(ctx context.Context, transcodingOption
 	s.LPMS.HandleRTMPPlay(getRTMPStreamHandler(s))
 
 	//LPMS hanlder for handling HLS video play
-	s.LPMS.HandleHLSPlay(getHLSMasterPlaylistHandler(s), getHLSMediaPlaylistHandler(s), getHLSSegmentHandler(s))
+	s.LPMS.HandleHLSPlay(getHLSMasterPlaylistHandler(s), getHLSMediaPlaylistHandler(s), getHLSSegmentHandler(s), getSegmentPushHandler(s))
 
 	//Start the LPMS server
 	lpmsCtx, cancel := context.WithCancel(context.Background())
+
 	ec := make(chan error, 1)
 	go func() {
 		if err := s.LPMS.Start(lpmsCtx); err != nil {
@@ -151,6 +152,12 @@ func (s *LivepeerServer) StartMediaServer(ctx context.Context, transcodingOption
 	case <-ctx.Done():
 		cancel()
 		return ctx.Err()
+	}
+}
+
+func getSegmentPushHandler(s *LivepeerServer) func(w http.ResponseWriter, r *http.Request) {
+	return func(w http.ResponseWriter, r *http.Request) {
+		s.HandlePush(w, r)
 	}
 }
 
--- a/server/mediaserver.go
+++ b/server/mediaserver.go
@@ -87,7 +87,6 @@ type LivepeerServer struct {
 	LivepeerNode          *core.LivepeerNode
 	HTTPMux               *http.ServeMux
 	ExposeCurrentManifest bool
-	httpAddr              string
 
 	// Thread sensitive fields. All accesses to the
 	// following fields should be protected by `connectionLock`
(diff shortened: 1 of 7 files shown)


server/mediaserver.go
```diff
@@ -130,10 +130,11 @@ func (s *LivepeerServer) StartMediaServer(ctx context.Context, transcodingOption
 	s.LPMS.HandleRTMPPlay(getRTMPStreamHandler(s))
 
 	//LPMS hanlder for handling HLS video play
-	s.LPMS.HandleHLSPlay(getHLSMasterPlaylistHandler(s), getHLSMediaPlaylistHandler(s), getHLSSegmentHandler(s))
+	s.LPMS.HandleHLSPlay(getHLSMasterPlaylistHandler(s), getHLSMediaPlaylistHandler(s), getHLSSegmentHandler(s), getSegmentPushHandler(s))
```
j0sh: I'd steer away from making changes to LPMS here.

LPMS can be parameterized with a custom HTTP mux, and this can be used to extend HTTP with additional endpoints. The orchestrator already does this:

https://github.com/livepeer/go-livepeer/blob/f5106de234ff99f26916f6965a0f8fe41e789f25/server/mediaserver.go#L103-L117

https://github.com/livepeer/go-livepeer/blob/f5106de234ff99f26916f6965a0f8fe41e789f25/server/rpc.go#L133-L145

server/mediaserver.go
```diff
@@ -540,6 +540,7 @@ func (s *LivepeerServer) HandlePush(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	r.Body.Close()
+	r.URL = &url.URL{Scheme: "http", Host: r.Host, Path: r.URL.Path}
```
j0sh, line 549: Naturally we won't always have plain `http` but this is probably OK for now. Just double checking, does the `r.Host` field actually contain the hostname data we need?

server/mediaserver.go
angyangie: Would that mean including something like this in mediaserver.go, in under `StartMediaServer`? 
```golang
if s.LivepeerNode.NodeType == core.BroadcasterNode {
		s.HTTPMux.HandleFunc("/live/", func(w http.ResponseWriter, r *http.Request) {
			s.HandlePush(w, r)
		})
	}
```

If so, how would we then add the following, which currently live in player.go, to that same `HTTMux`, so all routes can be handled on that port? 
``` golang
        s.mux.HandleFunc("/stream/", func(w http.ResponseWriter, r *http.Request) {
		handleLive(w, r, getMasterPlaylist, getMediaPlaylist, getSegment)
	})

	s.mux.HandleFunc("/vod/", func(w http.ResponseWriter, r *http.Request) {
		handleVOD(r.URL, s.VodPath, w)
	})
```

`func (s *VidPlayer) HandleHLSPlay` in lpms/vidplayer/player.go does not have access to the Livepeer server, meaning no access to it's `HTTPMux` field.

server/mediaserver.go
angyangie: It does indeed! I tested it locally under a number of different scenarios and it works

server/mediaserver.go
j0sh: > Would that mean including something like this in mediaserver.go, in under StartMediaServer?

Yep, that looks about right. Things could likely be simplified to:

```golang
 opts := lpmscore.LPMSOpts{ 
 	RtmpAddr: rtmpAddr, RtmpDisabled: true, 
 	HttpAddr: httpAddr, 
 	WorkDir:  lpNode.WorkDir, 
 	HttpMux: http.NewServeMux(), // does it make sense to pass this in via NewLivepeerNode param?
 } 
switch lpNode.NodeType {
case core.BroadcasterNode:
  opts.RtmpDisabled = false;
  opts.HttpMux.HandleFunc("/live/", s.HandlePush)
}
```

> func (s *VidPlayer) HandleHLSPlay in lpms/vidplayer/player.go does not have access to the Livepeer server, meaning no access to it's HTTPMux field.

The LPMS vidplayer code remains untouched, since the `s.mux` would be the same as the `opts.HTTPMux` that's passed in to `lpmscore.New`. See the implementation:

https://github.com/livepeer/lpms/blob/4b2845787110924d95aff019d16174232b29d88c/core/lpms.go#L81

https://github.com/livepeer/lpms/blob/4b2845787110924d95aff019d16174232b29d88c/vidplayer/player.go#L43-L47

One day, I'd be happy to refactor LPMS to remove the vidplayer HTTP handling logic because it's barely used, but this will be fine for now.

angyangie: Hmmm but how would we have access to `s.HandlePush` in `NewLivepeerServer`, if we're creating the LivepeerServer at the end of that function? Even if we do create the LivepeerServer prior to that, we have to create the LPMS server prior to creating the LivepeerServer, which is the only time we can change/add to `lpmscore.LPMSOpts`, and therefore ... we don't have access to those properties from `go-livepeer`. 

We could do this: 
``` golang
func NewLivepeerServer(rtmpAddr string, httpAddr string, lpNode *core.LivepeerNode) *LivepeerServer {
	opts := lpmscore.LPMSOpts{
		RtmpAddr: rtmpAddr, RtmpDisabled: true,
		HttpAddr: httpAddr,
		WorkDir:  lpNode.WorkDir,
		HttpMux:  http.NewServeMux(),
	}
	switch lpNode.NodeType {
	case core.BroadcasterNode:
		opts.RtmpDisabled = false
	}
	server := lpmscore.New(&opts)
	s := LivepeerServer{RTMPSegmenter: server, LPMS: server, LivepeerNode: lpNode, HTTPMux: opts.HttpMux, connectionLock: &sync.RWMutex{}, rtmpConnections: make(map[core.ManifestID]*rtmpConnection)}
	s.HTTPMux.HandleFunc("/live/", func(w http.ResponseWriter, r *http.Request) { //out
		s.HandlePush(w, r)
	})
	return &s
}
```

However, that doesn't work in practice. 8935 still does not serve `/live` in that scenario. Adding a route to  `s.HTTPMux` does not seem to carry over to `opts.HttpMux`. I'm not sure I understand how we can avoid changing LPMS code to make this work :(

j0sh: > We could do this:

Yes, that way will work. I'd still avoid wrapping it in an intermediate trampoline-like function and do it like this:

```
s.HTTPMux.HandleFunc("/live/", s.HandlePush)
```

> 8935 still does not serve /live in that scenario.

Does it serve `/stream/` at all? If you look at the [LPMS implementation](https://github.com/livepeer/lpms/blob/master/core/lpms.go#L96-L100) [1],  LPMS also starts the HTTP server if a default mux is used. If a mux is passed in, then the HTTP server attached to that needs to be started by someone else. So I suspect right now there isn't even a running server on 8935.

In that case, the server will need to be started somewhere on the broadcaster side, eg `s.HttpMux.ListenAndServe`

darkdarkdragon: @angyangie it should look like this: 41535c6947a22d5349ceb92e94f27e5e08f9162e

angyangie: THANK YOU @darkdarkdragon and @j0sh! No server was running on 8935, so your solution works well @darkdarkdragon

server/mediaserver.go
```diff
@@ -509,10 +537,121 @@ func getRTMPStreamHandler(s *LivepeerServer) func(url *url.URL) (stream.RTMPVide
...
+		cxn, err = s.registerConnection(st)
+		if err != nil {
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+			return
+		}
+
+		ticker := time.NewTicker(RefreshIntervalHttpPush)
+
+		go func(s *LivepeerServer, mid core.ManifestID) {
+			for _ = range ticker.C {
```
darkdarkdragon: This should be changed to
`for range ticker.C {`

server/mediaserver.go
```diff
@@ -509,10 +537,121 @@ func getRTMPStreamHandler(s *LivepeerServer) func(url *url.URL) (stream.RTMPVide
 
 //End RTMP Handlers
 
+func (s *LivepeerServer) HandlePush(w http.ResponseWriter, r *http.Request) {
+	// we read this unconditionally, mostly for ffmpeg
+	body, err := ioutil.ReadAll(r.Body)
+
+	if err != nil {
+		http.Error(w, fmt.Sprintf(`Error reading http request body: %s`, err.Error()), http.StatusInternalServerError)
```
darkdarkdragon: It would be good to also print this error with `glog`

server/mediaserver.go
```diff
@@ -509,10 +537,121 @@ func getRTMPStreamHandler(s *LivepeerServer) func(url *url.URL) (stream.RTMPVide
...
+					return
+				}
+			}
+		}(s, mid)
+	}
+
+	fname := path.Base(r.URL.Path)
+	seq, err := strconv.ParseUint(strings.TrimSuffix(fname, path.Ext(fname)), 10, 64)
+	if err != nil {
+		seq = 0
```
darkdarkdragon: So, if names in incoming segments would be not parseable, every segment would be having `SeqNo` set to zero, and playlist will always contains just one segment. Maybe shouldn't accept segments for which we couldn't find `SeqNo`?

darkdarkdragon (changes requested):
Have couple of comments, otherwise looks good.

server/mediaserver.go
```diff
@@ -105,21 +105,27 @@ type authWebhookResponse struct {
 
 func NewLivepeerServer(rtmpAddr string, httpAddr string, lpNode *core.LivepeerNode) *LivepeerServer {
 	opts := lpmscore.LPMSOpts{
-		RtmpAddr: rtmpAddr, RtmpDisabled: true,
-		HttpAddr: httpAddr,
-		WorkDir:  lpNode.WorkDir,
+		RtmpAddr:     rtmpAddr,
+		RtmpDisabled: true,
+		HttpAddr:     httpAddr,
```
j0sh: We can probably omit `HttpAddr` here (and from the `LivepeerServer` as well), since the `httpAddr` value from CLI should still be in scope when `http.ListenAndServe` is called.

So the function signature here becomes:

```golang
func NewLivepeerServer(rtmpAddr string, lpNode *core.LivepeerNode) *LivepeerServer
```

server/mediaserver.go
angyangie: That's a good point @darkdarkdragon. @j0sh do you have any color on this one?

server/mediaserver.go
angyangie: Interesting. So how would we access `httpAddr` here? 

```golang
ec <- http.ListenAndServe(s.httpAddr, s.HTTPMux)
```

j0sh: Does `ec <- http.ListenAndServe(*httpAddr, s.HTTPMux)` work?

angyangie: `Mediaserver.go` doesn't seem to have access to `*httpAddr`. It's undefined

server/mediaserver.go
j0sh: Probably OK. I can see cases where we might want to be more lenient wrt naming but we don't have to cover those in this PR.

server/mediaserver.go
darkdarkdragon: Probably @josh mean to change `StartMediaServer` to be `StartMediaServer(ctx context.Context, transcodingOptions, httpAddre string)`.
Anyway `httpAddr` is not used in `NewLivepeerServer` directly.

angyangie: OH yes, we could do that @darkdarkdragon !

server/mediaserver.go
darkdarkdragon: Probably OK as it is now, or OK to to change it to not accept segments with un-parsable names?

server/mediaserver.go
```diff
@@ -105,22 +104,28 @@ type authWebhookResponse struct {
 
 func NewLivepeerServer(rtmpAddr string, httpAddr string, lpNode *core.LivepeerNode) *LivepeerServer {
```
darkdarkdragon: I think there is no need to pass `httpAddr` to `NewLivepeerServer`, because it is not used in it.

darkdarkdragon (changes requested):

server/mediaserver.go
angyangie: Do you think we should change `lpmscore.LPMSOpts` to not include `httpAddr`?

darkdarkdragon: Yes. If you look at `LPMS` struct, it uses `HttpAddr` only in case if `HttpMux` is null. But now we always setting `HttpMux` to new `Mux`, so `HttpAddr` is not used by `LPMS`.

angyangie: I see that now. Thanks @darkdarkdragon !

vendor/github.com/livepeer/lpms/core/lpms.go
```diff
@@ -30,7 +30,6 @@ type LPMS struct {
 	vidPlayer   *vidplayer.VidPlayer
 	vidListener *vidlistener.VidListener
 	workDir     string
-	httpAddr    string
```
darkdarkdragon: I don't think we should change anything in LPMS itself. It's good as it is, we just don't need to pass our `httpAddr` to it.

darkdarkdragon (changes requested):

vendor/github.com/livepeer/lpms/core/lpms.go
angyangie: I've tested and everything works as expected. Do we not want to change LPMS in case we in the future add something that does need the `httpAddr` field? Or is that a field that might be used for something else? 

So the only changes we make will be to go-livepeer?

j0sh: Agreed with @darkdarkdragon . The changes here should be for go-livepeer only. Any changes to LPMS should be proposed and implemented in a separate PR there. Even then, this type of change is not really necessary IMO, as it removes a feature that is still useful to LPMS by itself.

edit: There is an argument to be made that LPMS would be better off without HTTP handling support at all (or rather, the goclient does not necessarily have to use LPMS HTTP handling), but that's a discussion that's outside the scope of this PR.

server/mediaserver.go
j0sh: Either works for me

server/mediaserver.go
j0sh: Seems like there was a misunderstanding. 

> Do you think we should change lpmscore.LPMSOpts to not include httpAddr?

I understood this question to mean that the `HttpAddr: httpAddr` should be removed from the specific instantiation of LPMSOpts within `mediaserver.go`, as in

https://github.com/livepeer/go-livepeer/blob/81414c252679a8af38b0c81757071bbbb9533ac9/server/mediaserver.go#L106-L112

Not that feature of configuring the HTTP binding address should be removed from LPMS completely, eg https://github.com/livepeer/go-livepeer/pull/1036#discussion_r315778381

vendor/github.com/livepeer/lpms/core/lpms.go
angyangie: Ah I see. I misunderstood. I've reverted the LPMS changes!

darkdarkdragon (approved):
LGTM

server/mediaserver.go
```diff
@@ -529,10 +540,13 @@ func (s *LivepeerServer) HandlePush(w http.ResponseWriter, r *http.Request) {
...
 
 	if err != nil {
-		http.Error(w, fmt.Sprintf(`Error reading http request body: %s`, err.Error()), http.StatusInternalServerError)
+		httpErr := fmt.Sprintf(`Error reading http request body: %s`, err.Error())
+		glog.Error(httpErr)
+		http.Error(w, httpErr, http.StatusInternalServerError)
 		return
 	}
 	r.Body.Close()
+	r.URL = &url.URL{Scheme: "http", Host: r.Host, Path: r.URL.Path}
```
j0sh, line 549: Can unit tests be added to check this bit works as expected?

angyangie: I'm not sure how to test that - whether I include this or not: `r.URL = &url.URL{Scheme: "http", Host: r.Host, Path: r.URL.Path}` .... both tests will either pass or not. Any suggestions?

angyangie: I've tried editing the request URL here, `req := httptest.NewRequest("GET", "/live/seg.ts", reader)`, trying different permutations, and with or without the line in question, the result is the same (either passing or not passing)

angyangie: These give me the same values: 
```
glog.Error("R URL Scheme: ", r.URL.Scheme)
	glog.Error("R URL Host: ", r.URL.Host)
	glog.Error("R URL Host222: ", r.Host)
	glog.Error("R URL Port: ", r.URL.Port())
	glog.Error("R URL Path: ", r.URL.Path)
	glog.Error("R URL Query: ", r.URL.Query())
	glog.Error("R requestURI", r.RequestURI)
	glog.Error("R header HOST: ", r.Header.Get("Host"))
	glog.Error("R header forwarded proto: ", r.Header.Get("X-Forwarded-Proto"))
```

j0sh: > both tests will either pass or not

Which tests?

The original concern around this line was that the URL in the webhook request did not have the full scheme or host. So I'd build the test along those lines - stub out the webhook callback, invoke the push, check the URL in the callback request matches.

If you look at https://golang.org/pkg/net/http/httptest/#NewRequest then the documentation states that if the host is not given, the `example.com` is used. So for your example of `req := httptest.NewRequest("GET", "/live/seg.ts", reader)` , the URL in the webhook should look something like `http://example.com/live/seg.ts`

angyangie: OH that was the missing piece for me ... didn't realize that I could check `req.URL` after calling `handler.ServeHTTP(w, req)` and `resp := w.Result()` in our test file, and that it would reflect the URL changes on line 549. I'll write up a couple of tests for that now!

Issue #1025, opened 2019-08-06
common: Add DB methods adhering to blockwatch.MiniHeaderStore


yondonfu:
**Is your feature request related to a problem? Please describe.**

#1024 will define the `MiniHeaderStore` interface which the stack can use to store block headers. We need an implementation of this interface.

**Describe the solution you'd like**

We can use the SQLite DB to store block headers and update `common.DB` with methods so that it adheres to the `MiniHeaderStore` interface.

PR #1037, opened 2019-08-08
eth: Port blockwatcher and tests

### LINKED ISSUE #1026: blockwatch: Add block polling

A block watcher will need to poll for new blocks at a regular interval. It should also be able to poll for blocks starting at a specific block number.

This implementation can be largely based off of: https://github.com/0xProject/0x-mesh/blob/development/ethereum/blockwatch/block_watcher.go

### LINKED ISSUE #1027: blockwatch: Handle re-orgs

A block watcher will need to be able to handle blockchain re-orgs and notify any consumer that is subscribed to new blocks of a re-org so that the consumer can react to the re-orged blocks.

This implementation can largely be based off of: https://github.com/0xProject/0x-mesh/blob/development/ethereum/blockwatch/block_watcher.go

### LINKED ISSUE #1028: blockwatch: Backfill logic

The block watcher should be able to back fill events/blocks from the past if it is missing them. One thing to consider here is that there may be query limits depending on the ETH node provider. We've run into this issue in the past trying to backfill with our current event watching logic - since it was low priority at the time we just gave up on backfilling if the block was too far in the past, but a better approach is to break up the blocks that we need to query for into smaller chunks.

This implementation can largely be based off of https://github.com/0xProject/0x-mesh/blob/development/ethereum/blockwatch/block_watcher.go

### LINKED ISSUE #1029: blockwatch: Subscription logic

A block watcher should expose an subscription API that allows external consumers to subscribe to new blocks/events. The block watcher shouldn't need to know about the actual event definitions - only the consumers need to know the event definitions. The architecture that we want is for the block watcher to be the single source of truth in the node when it comes to new blocks and events which are then passed to consumers that can use the new blocks and events in whatever way they see fit.

This implementation can largely be based off of https://github.com/0xProject/0x-mesh/blob/development/ethereum/blockwatch/block_watcher.go

yondonfu:

This PR ports the remainder of the block watcher from https://github.com/0xProject/0x-mesh/tree/development/ethereum/blockwatch

- In the interest of not adding an additional package dependency, I swapped [logrus](https://github.com/sirupsen/logrus) for `glog`. I actually would be interested to switching to a structured logger for go-livepeer, but that's something we can decide on separately and we should discuss since we've been using `glog` exclusively up until this point
- Use `*MiniHeader` instead of `*meshdb.MiniHeader` and use the `MiniHeaderStore` interface instead of the `*meshdb.MeshDB` concrete type
- Fixed a few typos and outdated comments

After this PR we will be done porting the `blockwatch` package into go-livepeer. However, we will still need to implement a `*common.DB` backed `MiniHeaderStore` implementation (see #1025).

You can run the following commands in order to generate diff files that compare the files included in this PR with the files in the 0x package:

```
git diff --color $GOPATH/src/github.com/0xProject/0x-mesh/ethereum/blockwatch/block_watcher.go block_watcher.go > block_watcher.diff
git diff --color $GOPATH/src/github.com/0xProject/0x-mesh/ethereum/blockwatch/block_watcher_test.go block_watcher_test.go > block_watcher_test.diff
git diff --color $GOPATH/src/github.com/0xProject/0x-mesh/ethereum/blockwatch/fake_client.go fake_client.go > fake_client.diff
git diff --color $GOPATH/src/github.com/0xProject/0x-mesh/ethereum/blockwatch/fake_log_client.go fake_log_client.go > fake_log_client.diff
git diff --color $GOPATH/src/github.com/0xProject/0x-mesh/ethereum/blockwatch/testdata testdata > testdata.diff
```

Ran the unit tests that come with the block watcher to ensure that none of the minor modifications I made cause test breakage.

Fixes #1026 
Fixes #1027 
Fixes #1028 
Fixes #1029

**Checklist:**

### DIFF
--- a/eth/blockwatch/block_watcher.go
+++ b/eth/blockwatch/block_watcher.go
@@ -1,0 +1,581 @@
+package blockwatch
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"math/big"
+	"sync"
+	"time"
+
+	"github.com/ethereum/go-ethereum"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/event"
+	"github.com/ethereum/go-ethereum/rpc"
+	"github.com/golang/glog"
+)
+
+// maxBlocksInGetLogsQuery is the max number of blocks to fetch logs for in a single query. There is
+// a hard limit of 10,000 logs returned by a single `eth_getLogs` query by Infura's Ethereum nodes so
+// we need to try and stay below it. Parity, Geth and Alchemy all have much higher limits (if any) on
+// the number of logs returned so Infura is by far the limiting factor.
+var maxBlocksInGetLogsQuery = 60
+
+// EventType describes the types of events emitted by blockwatch.Watcher. A block can be discovered
+// and added to our representation of the chain. During a block re-org, a block previously stored
+// can be removed from the list.
+type EventType int
+
+const (
+	// Added indicates that an event was added to our representation of the chain
+	Added EventType = iota
+	// Removed indicates that an event was removed from our representation of the chain
+	Removed
+)
+
+// Event describes a block event emitted by a Watcher
+type Event struct {
+	Type        EventType
+	BlockHeader *MiniHeader
+}
+
+// Config holds some configuration options for an instance of BlockWatcher.
+type Config struct {
+	Store               MiniHeaderStore
+	PollingInterval     time.Duration
+	StartBlockDepth     rpc.BlockNumber
+	BlockRetentionLimit int
+	WithLogs            bool
+	Topics              []common.Hash
+	Client              Client
+}
+
+// Watcher maintains a consistent representation of the latest `blockRetentionLimit` blocks,
+// handling block re-orgs and network disruptions gracefully. It can be started from any arbitrary
+// block height, and will emit both block added and removed events.
+type Watcher struct {
+	blockRetentionLimit int
+	startBlockDepth     rpc.BlockNumber
+	stack               *Stack
+	client              Client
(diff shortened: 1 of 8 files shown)


yondonfu:
1690c98 uses the `hexutil.EncodeBig()` helper to encode the block param when constructing the JSON-RPC request in `HeaderByNumber()` which makes sure that the hex string does not have any leading zeros after the `0x` prefix. See the [ETH JSON-RPC spec](https://github.com/ethereum/wiki/wiki/JSON-RPC#hex-value-encoding) for a note on disallowing leading zeros in hex strings representing quantities.

Example code:

```go
num := big.NewInt(271)
fmt.Printf("%v\n", fmt.Sprintf("0x%s", common.Bytes2Hex(num.Bytes())))
fmt.Printf("%v\n", fmt.Sprintf("0x%s", hex.EncodeToString(num.Bytes())))
fmt.Printf("%v\n", hexutil.EncodeBig(num))
fmt.Printf("%v\n", fmt.Sprintf("%#x", num))

num = big.NewInt(6000000)
fmt.Printf("%v\n", fmt.Sprintf("0x%s", common.Bytes2Hex(num.Bytes())))
fmt.Printf("%v\n", fmt.Sprintf("0x%s", hex.EncodeToString(num.Bytes())))
fmt.Printf("%v\n", hexutil.EncodeBig(num))
fmt.Printf("%v\n", fmt.Sprintf("%#x", num))
```

The output:

```
0x010f
0x010f
0x10f
0x10f
0x5b8d80
0x5b8d80
0x5b8d80
0x5b8d80
```

Note the difference in hex string output and the presence of leading zeros when using `common.Bytes2Hex()` and `hex.EncodeToString()`.

kyriediculous (approved):
Looks good to me, will give it a second pass in the morning. 🚀

PR #1038, opened 2019-08-08
common: DB adapter for blockwatch

### LINKED ISSUE #1025: common: Add DB methods adhering to blockwatch.MiniHeaderStore

kyriediculous:
This PR implements the necessary methods for the sqlite DB in the go-livepeer `common` package to adhere to the `MiniHeaderStore` interface in the `blockwatch` package. 

These methods allow us to store the stack that the `blockwatch` package maintains in the DB.

- block numbers are stored as `int64`
- blockhash and parenthash are stored as hex encoded `string` 
- a block's event logs are stored as JSON encoded `string` 

- Added a new table `blockheaders` to the schema 
- Added 4 new queries:  `inserMiniHeader`, `findLatestMiniHeader`, `findAllMiniHeadersSortedByNumber` and `deleteMiniHeader`
- Added methods that adhere to the `MiniHeaderStore` interface that implement the new queries
- Added helpers to encode and decode event logs to JSON for storage.
- Added unit tests `db_test.go` to test the new methods

ran unit tests

Fixes #1025 

**Checklist:**

### DIFF
--- a/common/db.go
+++ b/common/db.go
@@ -3,13 +3,17 @@ package common
 import (
 	"bytes"
 	"database/sql"
+	"encoding/json"
+	"fmt"
 	"math/big"
 	"strconv"
 	"strings"
 	"text/template"
 
 	ethcommon "github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/golang/glog"
+	"github.com/livepeer/go-livepeer/eth/blockwatch"
 	"github.com/livepeer/go-livepeer/pm"
 	_ "github.com/mattn/go-sqlite3"
 	"github.com/pkg/errors"
@@ -27,6 +31,10 @@ type DB struct {
 	unbondingLocks                   *sql.Stmt
 	withdrawableUnbondingLocks       *sql.Stmt
 	insertWinningTicket              *sql.Stmt
+	insertMiniHeader                 *sql.Stmt
+	findLatestMiniHeader             *sql.Stmt
+	findAllMiniHeadersSortedByNumber *sql.Stmt
+	deleteMiniHeader                 *sql.Stmt
 }
 
 type DBOrch struct {
@@ -93,6 +101,15 @@ var schema = `
 	);
 
 	CREATE INDEX IF NOT EXISTS idx_winningtickets_sessionid ON winningTickets(sessionID);
+
+	CREATE TABLE IF NOT EXISTS blockheaders (
+		number int64,
+		parent STRING,
+		hash STRING PRIMARY KEY,
+		logs BLOB
+	);
+
+	CREATE INDEX IF NOT EXISTS idx_blockheaders_number ON blockheaders(number);
 `
 
 func NewDBOrch(serviceURI string, orchAddr string) *DBOrch {
@@ -200,6 +217,42 @@ func InitDB(dbPath string) (*DB, error) {
 	}
 	d.insertWinningTicket = stmt
 
+	// Insert block header
+	stmt, err = db.Prepare("INSERT INTO blockheaders(number, parent, hash, logs) VALUES(?, ?, ?, ?)")
+	if err != nil {
+		glog.Error("Unable to prepare insertMiniHeader", err)
+		d.Close()
+		return nil, err
+	}
+	d.insertMiniHeader = stmt
+
+	// Find the latest block header
+	stmt, err = db.Prepare("SELECT * FROM blockheaders ORDER BY number DESC LIMIT 1")
+	if err != nil {
+		glog.Error("Unable to prepare findLatestMiniHeader", err)
+		d.Close()
+		return nil, err
+	}
+	d.findLatestMiniHeader = stmt
+
+	// Find all block headers sorted by number
+	stmt, err = db.Prepare("SELECT * FROM blockheaders ORDER BY number DESC")
+	if err != nil {
+		glog.Error("Unable to prepare findAllMiniHeadersSortedByNumber", err)
+		d.Close()
+		return nil, err
+	}
+	d.findAllMiniHeadersSortedByNumber = stmt
+
+	// Delete block header
+	stmt, err = db.Prepare("DELETE FROM blockheaders WHERE hash=?")
+	if err != nil {
+		glog.Error("Unable to prepare deleteMiniHeader", err)
+		d.Close()
+		return nil, err
+	}
+	d.deleteMiniHeader = stmt
+
 	glog.V(DEBUG).Info("Initialized DB node")
 	return &d, nil
 }
@@ -229,6 +282,18 @@ func (db *DB) Close() {
 	}
 	if db.insertWinningTicket != nil {
 		db.insertWinningTicket.Close()
+	}
+	if db.insertMiniHeader != nil {
+		db.insertMiniHeader.Close()
+	}
+	if db.findLatestMiniHeader != nil {
+		db.findLatestMiniHeader.Close()
+	}
+	if db.findAllMiniHeadersSortedByNumber != nil {
+		db.findAllMiniHeadersSortedByNumber.Close()
+	}
+	if db.deleteMiniHeader != nil {
+		db.deleteMiniHeader.Close()
 	}
 	if db.dbh != nil {
 		db.dbh.Close()
@@ -470,4 +535,110 @@ func buildSelectOrchsQuery(filter *DBOrchFilter) (string, error) {
 		query = query + " AND pricePerPixel <= " + strconv.FormatInt(fixedPrice, 10)
 	}
 	return query, nil
+}
+
+// FindLatestMiniHeader returns the MiniHeader with the highest blocknumber in the DB
+func (db *DB) FindLatestMiniHeader() (*blockwatch.MiniHeader, error) {
+	row := db.findLatestMiniHeader.QueryRow()
+	var (
+		number  int64
+		parent  string
+		hash    string
+		logsEnc []byte
+	)
+	if err := row.Scan(&number, &parent, &hash, &logsEnc); err != nil {
+		if err.Error() != "sql: no rows in result set" {
+			return nil, fmt.Errorf("could not retrieve latest header: %v", err)
+		}
+		// If there is no result return no error, just nil value
+		return nil, nil
+	}
+
+	logs, err := decodeLogsJSON(logsEnc)
+	if err != nil {
+		return nil, err
+	}
+	return &blockwatch.MiniHeader{
+		Number: big.NewInt(number),
+		Parent: ethcommon.HexToHash(parent),
+		Hash:   ethcommon.HexToHash(hash),
+		Logs:   logs,
+	}, nil
+}
+
+// FindAllMiniHeadersSortedByNumber returns all MiniHeaders in the DB sorting in descending order by block number
+func (db *DB) FindAllMiniHeadersSortedByNumber() ([]*blockwatch.MiniHeader, error) {
+	var headers []*blockwatch.MiniHeader
+	rows, err := db.findAllMiniHeadersSortedByNumber.Query()
+	defer rows.Close()
+	if err != nil {
+		return nil, err
+	}
+	for rows.Next() {
+		var (
+			number  int64
+			parent  string
+			hash    string
+			logsEnc []byte
+		)
+		if err := rows.Scan(&number, &parent, &hash, &logsEnc); err != nil {
+			return nil, err
+		}
+		logs, err := decodeLogsJSON(logsEnc)
+		if err != nil {
+			return nil, err
+		}
+		headers = append(headers, &blockwatch.MiniHeader{
+			Number: big.NewInt(number),
+			Parent: ethcommon.HexToHash(parent),
+			Hash:   ethcommon.HexToHash(hash),
+			Logs:   logs,
+		})
+	}
+	return headers, nil
+}
+
+// InsertMiniHeader inserts a MiniHeader into the database
+func (db *DB) InsertMiniHeader(header *blockwatch.MiniHeader) error {
+	if header == nil {
+		return errors.New("must provide a MiniHeader")
+	}
+	if header.Number == nil {
+		return errors.New("no block number found")
+	}
+	logsEnc, err := encodeLogsJSON(header.Logs)
+	if err != nil {
+		return err
+	}
+	_, err = db.insertMiniHeader.Exec(header.Number.Int64(), header.Parent.Hex(), header.Hash.Hex(), logsEnc)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// DeleteMiniHeader deletes a MiniHeader from the DB and takes in the blockhash of the block to be deleted as an argument
+func (db *DB) DeleteMiniHeader(hash ethcommon.Hash) error {
+	_, err := db.deleteMiniHeader.Exec(hash.Hex())
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func encodeLogsJSON(logs []types.Log) ([]byte, error) {
+	logsEnc, err := json.Marshal(logs)
+	if err != nil {
+		return []byte{}, err
+	}
+	return logsEnc, nil
+}
+
+func decodeLogsJSON(logsEnc []byte) ([]types.Log, error) {
+	var logs []types.Log
+	err := json.Unmarshal(logsEnc, &logs)
+	if err != nil {
+		return []types.Log{}, err
+	}
+	return logs, nil
 }
--- a/common/db.go
+++ b/common/db.go
@@ -19,15 +19,14 @@ type DB struct {
 	dbh *sql.DB
 
 	// prepared statements
-	selectOrchs                *sql.Stmt
-	filterOrchs                *sql.Stmt
-	updateOrch                 *sql.Stmt
-	updateKV                   *sql.Stmt
-	insertUnbondingLock        *sql.Stmt
-	useUnbondingLock           *sql.Stmt
-	unbondingLocks             *sql.Stmt
-	withdrawableUnbondingLocks *sql.Stmt
-	insertWinningTicket        *sql.Stmt
+	selectOrchs                      *sql.Stmt
+	updateOrch                       *sql.Stmt
+	updateKV                         *sql.Stmt
+	insertUnbondingLock              *sql.Stmt
+	useUnbondingLock                 *sql.Stmt
+	unbondingLocks                   *sql.Stmt
+	withdrawableUnbondingLocks       *sql.Stmt
+	insertWinningTicket              *sql.Stmt
 }
 
 type DBOrch struct {
@@ -212,9 +211,6 @@ func (db *DB) Close() {
 	}
 	if db.selectOrchs != nil {
 		db.selectOrchs.Close()
-	}
-	if db.filterOrchs != nil {
-		db.filterOrchs.Close()
 	}
 	if db.updateKV != nil {
 		db.updateKV.Close()
--- a/common/db_test.go
+++ b/common/db_test.go
@@ -8,7 +8,10 @@ import (
 	"strconv"
 	"testing"
 
+	"github.com/ethereum/go-ethereum/common"
 	ethcommon "github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/livepeer/go-livepeer/eth/blockwatch"
 	"github.com/livepeer/go-livepeer/pm"
 	"github.com/livepeer/lpms/ffmpeg"
 	_ "github.com/mattn/go-sqlite3"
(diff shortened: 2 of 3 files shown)


common/db.go
```diff
@@ -213,9 +260,6 @@ func (db *DB) Close() {
 	if db.selectOrchs != nil {
 		db.selectOrchs.Close()
 	}
-	if db.filterOrchs != nil {
```
yondonfu, line 216: I guess this is a leftover statement that we forgot to remove after deciding to use `selectOrchs` exclusively. Since this removal is not related to the addition of `MiniHeader` persistence in the DB, it could be a good idea to push this removal into its own distinct commit (we can also remove `filterOrchs` from the `DB` type definition as well).

common/db.go
```diff
@@ -475,3 +531,113 @@ func buildSelectOrchsQuery(filter *DBOrchFilter) (string, error) {
 	}
 	return query, nil
 }
+
+// FindLatestMiniHeader eturns the MiniHeader with the highest blocknumber in the DB
```
yondonfu: "eturns" -> "returns"

common/db.go
```diff
@@ -94,6 +101,13 @@ var schema = `
 	);
 
 	CREATE INDEX IF NOT EXISTS idx_winningtickets_sessionid ON winningTickets(sessionID);
+
+	CREATE TABLE IF NOT EXISTS blockheaders (
+		number int64 PRIMARY KEY,
+		parent STRING,
+		hash STRING,
```
yondonfu: I think the primary key should actually be `hash` since each block header is uniquely identified by its hash (there could be 2 block headers with the same `number` but different hashes).

common/db.go
```diff
@@ -94,6 +101,13 @@ var schema = `
 	);
 
 	CREATE INDEX IF NOT EXISTS idx_winningtickets_sessionid ON winningTickets(sessionID);
+
+	CREATE TABLE IF NOT EXISTS blockheaders (
+		number int64 PRIMARY KEY,
```
yondonfu: If we make `hash` the primary key (see my comment below), then we should add an index for `number` to speed up the `findLatestMiniHeader` and `findAllMiniHeadersSortedByNumber` queries.

common/db.go
```diff
@@ -201,6 +215,39 @@ func InitDB(dbPath string) (*DB, error) {
...
+	if err != nil {
+		glog.Error("Unable to prepare findAllMiniHeadersSortedByNumber", err)
+		return nil, err
+	}
+	d.findAllMiniHeadersSortedByNumber = stmt
+
+	// Delete block header
+	stmt, err = db.Prepare("DELETE FROM blockheaders WHERE hash=?")
+	if err != nil {
+		glog.Error("Unable to prepare DeleteMiniHeader", err)
```
yondonfu: Should be `deleteMiniHeader` to match the style used in the other error messages?

common/db.go
```diff
@@ -201,6 +215,39 @@ func InitDB(dbPath string) (*DB, error) {
 	}
 	d.insertWinningTicket = stmt
 
+	// Insert block header
+	stmt, err = db.Prepare("INSERT INTO blockheaders(number, parent, hash, logs) VALUES(?, ?, ?, ?)")
+	if err != nil {
+		glog.Error("Unable to prepare insertMiniHeader", err)
```
yondonfu, line 223: Missing `d.Close()` in the error handlers for all of the statement preparations except for `deleteMiniHeader`

common/db.go
```diff
@@ -475,3 +531,113 @@ func buildSelectOrchsQuery(filter *DBOrchFilter) (string, error) {
...
+		return nil, err
+	}
+	for rows.Next() {
+		var (
+			number  int64
+			parent  string
+			hash    string
+			logsEnc string
+		)
+		if err := rows.Scan(&number, &parent, &hash); err != nil {
```
yondonfu: Need to pass `&logsEnc` into `rows.Scan()` or else it will always be the empty string

common/db.go
```diff
@@ -94,6 +101,13 @@ var schema = `
 	);
 
 	CREATE INDEX IF NOT EXISTS idx_winningtickets_sessionid ON winningTickets(sessionID);
+
+	CREATE TABLE IF NOT EXISTS blockheaders (
+		number int64 PRIMARY KEY,
+		parent STRING,
+		hash STRING,
+		logs TEXT
```
yondonfu: Since `encodeLogsJSON()` returns a `[]byte` and `decodeLogsJSON()` accepts a `[]byte`, I think storing `logs` as a `BLOB` would allow us to skip the extra `string <-> []byte` conversion step.

common/db.go
```diff
@@ -475,3 +531,113 @@ func buildSelectOrchsQuery(filter *DBOrchFilter) (string, error) {
...
+		Parent: ethcommon.HexToHash(parent),
+		Hash:   ethcommon.HexToHash(hash),
+		Logs:   logs,
+	}, nil
+}
+
+// FindAllMiniHeadersSortedByNumber returns all MiniHeaders in the DB sorting in descending order by block number
+func (db *DB) FindAllMiniHeadersSortedByNumber() ([]*blockwatch.MiniHeader, error) {
+	var headers []*blockwatch.MiniHeader
+	rows, err := db.findAllMiniHeadersSortedByNumber.Query()
```
yondonfu, line 572: We should add a `defer rows.Close()` here.

common/db.go
```diff
@@ -475,3 +531,113 @@ func buildSelectOrchsQuery(filter *DBOrchFilter) (string, error) {
...
+			Logs:   logs,
+		})
+	}
+	return headers, nil
+}
+
+// InsertMiniHeader inserts a MiniHeader into the database
+func (db *DB) InsertMiniHeader(header *blockwatch.MiniHeader) error {
+	if header == nil {
+		return errors.New("must provide a Miniheader")
```
yondonfu: "Miniheader" -> "MiniHeader"

common/db.go
```diff
@@ -475,3 +531,113 @@ func buildSelectOrchsQuery(filter *DBOrchFilter) (string, error) {
...
+	}
+	if len(header.Hash) != 32 {
```
yondonfu: I think we could be ok without the length check here since `common.Hash` is just a [type alias](https://github.com/ethereum/go-ethereum/blob/master/common/types.go#L47) for a `[]byte` of length 32.

common/db.go
```diff
@@ -475,3 +531,113 @@ func buildSelectOrchsQuery(filter *DBOrchFilter) (string, error) {
...
+	_, err = db.insertMiniHeader.Exec(header.Number.Int64(), header.Parent.Hex(), header.Hash.Hex(), string(logsEnc))
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// DeleteMiniHeader deletes a MiniHeader from the DB and takes in the blockhash of the block to be deleted as an argument
+func (db *DB) DeleteMiniHeader(hash ethcommon.Hash) error {
+	if len(hash) != 32 {
```
yondonfu: I think we could be ok without the length check here since common.Hash is just a type alias for a []byte of length 32.

common/db_test.go
```diff
@@ -679,3 +759,19 @@ func getRowCountOrFatal(query string, dbraw *sql.DB, t *testing.T) int {
 
 	return count
 }
+
+func defaultMiniHeader(t *testing.T) *blockwatch.MiniHeader {
+
```
yondonfu: Can kill space?

common/db_test.go
```diff
@@ -679,3 +759,19 @@ func getRowCountOrFatal(query string, dbraw *sql.DB, t *testing.T) int {
 
 	return count
 }
+
+func defaultMiniHeader(t *testing.T) *blockwatch.MiniHeader {
```
yondonfu: I don't think this function needs to accept `*testing.T` as an argument.

common/db_test.go
```diff
@@ -656,6 +659,83 @@ func TestLoadWinningTickets_GivenTwoSessionsWithTickets_ReturnsAllTickets(t *tes
...
+func TestInsertMiniHeader_ReturnsFindLatestMiniHeader(t *testing.T) {
+	dbh, dbraw, err := TempDB(t)
+	defer dbh.Close()
+	defer dbraw.Close()
+	assert := assert.New(t)
+	require := require.New(t)
+	require.Nil(err)
+
+	h0 := defaultMiniHeader(t)
+	err = dbh.InsertMiniHeader(h0)
```
yondonfu, line 671: Test the error paths for `InsertMiniHeader()`?

common/db_test.go
```diff
@@ -656,6 +659,83 @@ func TestLoadWinningTickets_GivenTwoSessionsWithTickets_ReturnsAllTickets(t *tes
...
+	require.Nil(err)
+
+	h0db, err := dbh.FindLatestMiniHeader()
```
yondonfu, line 674: Test the error paths for `FindLatestMiniHeader()`?

common/db_test.go
```diff
@@ -656,6 +659,83 @@ func TestLoadWinningTickets_GivenTwoSessionsWithTickets_ReturnsAllTickets(t *tes
...
+	require.Nil(err)
+	assert.Equal(h0, h0db)
+
+	err = dbh.DeleteMiniHeader(h0.Hash)
```
yondonfu, line 756: Could add a few additional tests:

- When the header doesn't exist
- When there are > 1 headers already in the DB

common/db_test.go
yondonfu, line 674: I think its also worth checking that the `MiniHeader` returned by `FindLatestMiniHeader()` or `FindAllMiniHeadersSortedByNumber()` can contain multiple logs (at the moment looks like we're always inserting `defaultMiniHeader()` which only has a single log).

yondonfu (changes requested):

common/db.go
kyriediculous: How can I extract that into it's seperate commit since it's already part of this commit? thanks !

common/db.go
kyriediculous: Will the queries automatically make use of the index?

Is the efficiency gain here for sorting the blocknumber in descending order on those queries?

common/db_test.go
kyriediculous: This is being checked in the Delete test, where after there are no more remaining entries `FindLatestMiniHeader` returns an error. I have added a comment there for clarity.

I didn't add a seperate test for `FindLatestMiniHeader` because its functionality is being tested in the other tests and it's correct working is crucial for those tests to pass.

common/db.go
kyriediculous: Found an answer

> When you use ORDER BY without an index on the column to be sorted, SQLite needs to build up a temporary data structure that contains the sorted results each time the query is executed. That data structure will use up space in memory (or on disk, depending on the circumstances) and will be torn down after the query is executed. This can be a lot of extra work that might noticeably degrade performance.

[Squeezeing performance from SQLite indexes: Indexes? Indexes!](https://medium.com/@JasonWyatt/squeezing-performance-from-sqlite-indexes-indexes-c4e175f3c346)

common/db_test.go
kyriediculous: Fixed in ced063e by adding a second log for uneven indices in tests and asserting the length of the logs contained in the MiniHeader

common/db_test.go
kyriediculous: Fixed in ced063e

common/db_test.go
kyriediculous: Fixed in ced063e

common/db_test.go
kyriediculous: Fixed in ced063e

common/db_test.go
kyriediculous: Fixed in ced063e

common/db.go
kyriediculous: Fixed in ced063e

common/db.go
kyriediculous: Fixed in ced063e

common/db.go
kyriediculous: Fixed in ced063e

common/db.go
kyriediculous: Fixed in ced063e

common/db.go
kyriediculous: Fixed in ced063e

common/db.go
kyriediculous: Fixed in ced063e

common/db.go
kyriediculous: Fixed in ced063e

common/db.go
kyriediculous: Fixed in ced063e

common/db.go
kyriediculous: Fixed in ced063e

common/db.go
kyriediculous: Fixed in ced063e

common/db.go
yondonfu: You can use `git add -p` which triggers git's patch mode (see this [blog post](https://johnkary.net/blog/git-add-p-the-most-powerful-git-feature-youre-not-using-yet/) for a more in-depth treatment of the command).

Suppose 8f4ac86 is the latest HEAD commit and the PR is approved already. You can do the following:

```
# This will unstage all the changes for latest HEAD commit
git reset HEAD~1

# Run git add in patch mode
git add -p
```

If the commit that you care about is further back then you can either use `HEAD~N` where N is the # of commits back from the HEAD or you can reference the commit hash.

The last command will open an interactive console that allows you to select "hunks" (basically chunks of code changes that Git detects) to stage for a commit. You can stage the hunk that corresponds to this change and then add a specific commit for it.

common/db.go
yondonfu: > Will the queries automatically make use of the index?

Yep.

> Is the efficiency gain here for sorting the blocknumber in descending order on those queries?


Yaeh - adding an index can help avoid the sort because the index can allow the [rows to be delivered in the required order](https://use-the-index-luke.com/sql/sorting-grouping/indexed-order-by).

common/db_test.go
```diff
@@ -656,6 +659,129 @@ func TestLoadWinningTickets_GivenTwoSessionsWithTickets_ReturnsAllTickets(t *tes
...
+	err = dbh.InsertMiniHeader(h1)
+	require.Nil(err)
+	latest, err := dbh.FindLatestMiniHeader()
+	require.Nil(err)
+	assert.Equal(h1, latest)
+	assert.Equal(len(h1.Logs), 2)
+
+	// test MiniHeader = nil error
+	err = dbh.InsertMiniHeader(nil)
+	assert.EqualError(err, err.Error(), "must provide a MiniHeader")
```
yondonfu: Looks like `err` is compared against itself here when we want to compare it with the expected error string

common/db_test.go
```diff
@@ -656,6 +659,129 @@ func TestLoadWinningTickets_GivenTwoSessionsWithTickets_ReturnsAllTickets(t *tes
...
+
+	// test blocknumber = nil error
+	h1.Number = nil
+	err = dbh.InsertMiniHeader(h1)
+	assert.EqualError(err, err.Error(), "no block number found")
```
yondonfu: Looks like `err` is compared against itself here when we want to compare it with the expected error string

common/db_test.go
yondonfu: Can we add a sanity check that the logs are correctly encoded/decoded as well?

yondonfu (changes requested):

common/db_test.go
```diff
@@ -656,6 +659,129 @@ func TestLoadWinningTickets_GivenTwoSessionsWithTickets_ReturnsAllTickets(t *tes
...
+		err = dbh.InsertMiniHeader(h)
+		require.Nil(err)
+		added[9-i] = h
+	}
+
+	headers, err := dbh.FindAllMiniHeadersSortedByNumber()
+	for i, h := range headers {
+		assert.Equal(h.Number.Int64(), int64(len(headers)-1-i))
+		assert.Equal(h, added[i])
+		// Since MiniHeaders are inserted into 'added' in reverse order we need to check for uneven here instead of even
```
yondonfu: This comment is a bit confusing due to the `i%2 == 0` - might be clearer to say that even index = 1 log & odd index = 2 logs

common/db_test.go
kyriediculous: Aren't we checking that at line 676 where we compare that the inserted value equals the retrieved value from the DB? Or are we talking about different things?

common/db_test.go
kyriediculous: fixed in 72b6c06

common/db_test.go
kyriediculous: fixed in 72b6c06

common/db_test.go
kyriediculous: fixed in 72b6c06

kyriediculous:
In 0ca92c0 `FindLatestMiniHeader` no longer returns an error, but nil instead, when it can't find a result in DB.

If we'd return an error `stack.Peek` would error out on startup with an empty stack.

common/db_test.go
yondonfu: I wasn't sure if `testify/assert` handled the equality comparison for a nested array of structs properly, but after trying it out locally, it doesn't seem to be a problem. Never mind!

common/db_test.go
```diff
@@ -656,6 +659,83 @@ func TestLoadWinningTickets_GivenTwoSessionsWithTickets_ReturnsAllTickets(t *tes
...
+	require := require.New(t)
+	require.Nil(err)
+
+	h0 := defaultMiniHeader(t)
+	err = dbh.InsertMiniHeader(h0)
+	require.Nil(err)
+
+	h0db, err := dbh.FindLatestMiniHeader()
+	require.Nil(err)
+	assert.Equal(*h0, *h0db)
```
yondonfu: Just a heads up - [assert.Equal()](https://godoc.org/github.com/stretchr/testify/assert#Equal) will compare the referenced values when comparing 2 pointers so we don't need to explicitly deference pointers in this test file.

common/db_test.go
```diff
@@ -656,6 +659,128 @@ func TestLoadWinningTickets_GivenTwoSessionsWithTickets_ReturnsAllTickets(t *tes
...
+		err = dbh.InsertMiniHeader(h)
+		require.Nil(err)
+		added[9-i] = h
+	}
+
+	headers, err := dbh.FindAllMiniHeadersSortedByNumber()
+	for i, h := range headers {
+		assert.Equal(h.Number.Int64(), int64(len(headers)-1-i))
+		assert.Equal(h, added[i])
+		// even = 2 logs , uneven = 1 log
```
yondonfu: Isn't it the opposite? i.e. `i % 2 == 0` (even) -> 1 log, `i % 2 != 0` (odd) -> 2 logs

common/db_test.go
kyriediculous: Cool, I thought so but wasn't entirely sure!

common/db.go
```diff
@@ -19,15 +23,18 @@ type DB struct {
...
+	updateOrch                       *sql.Stmt
+	updateKV                         *sql.Stmt
+	insertUnbondingLock              *sql.Stmt
+	useUnbondingLock                 *sql.Stmt
+	unbondingLocks                   *sql.Stmt
+	withdrawableUnbondingLocks       *sql.Stmt
+	insertWinningTicket              *sql.Stmt
+	insertMiniHeader                 *sql.Stmt
+	findLatestMiniHeader             *sql.Stmt
+	findAllMiniHeadersSortedByNumber *sql.Stmt
```
j0sh, line 36: Would `getMiniHeaders` suffice here as a variable name? That way we aren't re-indenting a large block, and it is more concise while not being *too* specific. Same goes for the function name, `GetMiniHeaders`.

Just curious how reorgs are planned to be handled here?

kyriediculous: I thought it would be nice for the queries to be named according to the exported method that implements those queries. While we could rename the queries, we can't rename the exported methods as they need to adhere to the `MiniHeaderStore` interface introduced in https://github.com/livepeer/go-livepeer/pull/1030/files

> Just curious how reorgs are planned to be handled here?

Re-orgs are handles by the blockwatch package itself (`blockwatch.buildCanonicalChain()`), which maintains a stack of blocks. in this PR we implement only a persistant storage for that stack. 
If we find a re-org we will pop off blocks off the stack until the point of the re-org and then start Pushing blocks again. 

eg. `Stack.Pop` uses `DB.FindLatestMiniHeader()` and then `DB.DeleteMiniHeader(latestMiniHeader.Hash)` but DB doesn't need to know about any of the reorg logic.

j0sh: Sounds like the interface could also be renamed if we're just importing everything :) But that's me - just renaming the query handle itself is fine here too.

Thanks for the explanation on reorgs! I would have thought it would build all chains as a tree and use the longest / deepest as the canonical chain, but I guess it makes sense to prune the DB occasionally if a reorg is detected.

kyriediculous: Yeah we could have, but we decided to move fast and just port blockwatch package over from 0x-mesh and remove dependencies on their codebase. The latter included moving away from a specific database implementation they used to an interface that defines the same methods so we can plug in our own DB. 

0x will make their blockwatch package standalone in the coming weeks which we'll then contribute back to, this is more of an intermediate solution that will behave very similar to the eventual API.  I also had a few optimisations in mind but in today's streamflow design meeting we decided to hold off on those to not diverge from the original API too much and focus in implementing the livepeer specific event watching logic first.

common/db_test.go
kyriediculous: Ah yes you're right because we are reversing the order that headers are appended to `added`. 

Fixed in cb0ca5e

common/db_test.go
kyriediculous: Fixed in cb0ca5e

yondonfu:
Changes look good to me.

Re: [this comment thread](https://github.com/livepeer/go-livepeer/pull/1038#discussion_r313003414) - I would be ok keeping the current query name, but I'll leave that up to you.

When you're ready to merge, I recommend trying out `git add -p` to separate the code change mentioned in this [comment](https://github.com/livepeer/go-livepeer/pull/1038#discussion_r312597104) in addition to squashing the fixup commits as it is a useful tool to be familiar with IMO.

kyriediculous:
rebased and created a seperate commit for the cleanup of `filterorchs` references

yondonfu (approved):
LGTM! 🚢 

I pushed 9523478 to fix an import issue in `eth/blockwatch/client.go`

PR #1044, opened 2019-08-12
Fix for #1007 , http or https only

ya7ya:
this fixes the [typo](https://github.com/livepeer/go-livepeer/pull/1007#discussion_r306931947) mentioned by @iameli on #1007 

- 
- 
- 

**Checklist:**

### DIFF
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -644,11 +644,13 @@ func getOrchWebhook(u string) (*url.URL, error) {
 	if err != nil {
 		return nil, err
 	}
-	if p.Scheme != "http" && p.Scheme != "https" {
-		return nil, errors.New("Webhook URL should be HTTP or HTTP")
+
+	if p.Scheme == "http" || p.Scheme == "https" {
+		glog.Infof("Using orchestrator webhook url %s", u)
+		return p, nil
 	}
-	glog.Infof("Using orchestrator webhook url %s", u)
-	return p, nil
+
+	return nil, errors.New("Webhook URL should be HTTP or HTTP")
 }
 
 func getAuthWebhookURL(u string) (string, error) {


cmd/livepeer/livepeer.go
```diff
@@ -644,11 +644,13 @@ func getOrchWebhook(u string) (*url.URL, error) {
...
-		return nil, errors.New("Webhook URL should be HTTP or HTTP")
+
+	if p.Scheme == "http" || p.Scheme == "https" {
+		glog.Infof("Using orchestrator webhook url %s", u)
+		return p, nil
 	}
-	glog.Infof("Using orchestrator webhook url %s", u)
-	return p, nil
+
+	return nil, errors.New("Webhook URL should be HTTP or HTTP")
```
j0sh, line 653: It's generally a bit more idiomatic in golang to exit early on errors, and then proceed with the normal result after errors are handled - so the old flow should be retained.

Also, the original typo is still present here. The error message should say: `"Webhook URL should be HTTP or HTTPS"`

Essentially, this PR should be a one-line commit fixing that error message.

j0sh (changes requested):

cmd/livepeer/livepeer.go
ya7ya: > Essentially, this PR should be a one-line commit fixing that error message.

Yes, but when i tried `if p.Scheme != "http" || p.Scheme != "https" {` , `go vet` complained about `suspect or` because this apparently will always return true. 

Also, I didn't wanna commit to your branch, but feel free to do the changes over there and close this one

ya7ya: wait a minute, nvm I thought the issue was the condition , not the error msg 🤦‍♂

ya7ya:
I misunderstood the issue here, I thought it was the `if` statement while the original issue was the error msg itself. 

Sorry about that 🤦‍♂

j0sh:
@ya7ya No worries, feel free to commit directly to the branch fixing the error message!

wohlner:

PR #1045, opened 2019-08-14
Initial proof of concept for orchestrator webhook

angyangie:

In `offchain` mode, O's passed in through the `-orchAddr` flag are added to `OrchestratorPool`. Once the node is started, there is no way to update that list currently. This PR populates `n.OrchestratorPool` from `--orchestrator-webhook-url=https://livepeer.live/api/orchestrator` when flag is supplied. The cached DB list is refreshed every minute. When list is too short to handle all incoming work, `OrchestratorPool` is refreshed from DB as works now.

_Looking for initial thoughts on approach! Tests have still not been included, nor have I done extensive review._

Read more here: https://github.com/livepeer/go-livepeer/issues/851

TBD

https://github.com/livepeer/go-livepeer/issues/851

**Checklist:**

### DIFF
--- a/discovery/webhook_discovery.go
+++ b/discovery/webhook_discovery.go
@@ -1,0 +1,137 @@
+package discovery
+
+import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"net/url"
+	"time"
+
+	"github.com/livepeer/go-livepeer/common"
+	"github.com/livepeer/go-livepeer/core"
+	"github.com/livepeer/go-livepeer/net"
+	"github.com/livepeer/go-livepeer/server"
+
+	"github.com/golang/glog"
+)
+
+var webhookCacheRefreshInterval = 1 * time.Minute
+var getWebhookTicker = func() *time.Ticker {
+	return time.NewTicker(webhookCacheRefreshInterval)
+}
+
+type WHOrchestratorPoolCache struct {
+	node       *core.LivepeerNode
+	webhookURL string
+}
+
+type orchAddress struct {
+	address string `json:address`
+}
+
+func NewWebhookOrchestratorPoolCache(node *core.LivepeerNode, orchWebhookUrl string) *WHOrchestratorPoolCache {
+	_ = cacheWebhookTranscoders(node, orchWebhookUrl)
+
+	ticker := getWebhookTicker()
+	go func(node *core.LivepeerNode) {
+		for _ = range ticker.C {
+			err := cacheWebhookTranscoders(node, orchWebhookUrl)
+			if err != nil {
+				continue
+			}
+		}
+	}(node)
+
+	return &WHOrchestratorPoolCache{node: node, webhookURL: orchWebhookUrl}
+}
+
+func getOrchAddresses(orchWebhookUrl string) ([]string, error) {
+	// Get data from the Orchestrator Webhook URL
+	resp, err := http.Get(orchWebhookUrl)
+	if err != nil || resp.StatusCode != 200 {
+		return nil, fmt.Errorf("Http.Get error: %v ", err)
+	}
+	defer resp.Body.Close()
+
+	// Read json info on available orchestrators
+	orchAddrBytes, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return nil, fmt.Errorf("File ReadAll error: %v", err)
+	}
+
+	var orchAddresses []orchAddress
+	err = json.Unmarshal(orchAddrBytes, &orchAddresses)
+	if err != nil {
+		return nil, fmt.Errorf("Json unmarshal error: %v", err)
+	}
+
+	var stringAddr []string
+	for _, addr := range orchAddresses {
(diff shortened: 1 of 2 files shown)


darkdarkdragon:
@angyangie I thought this was done here: 4f3696eb62f50b59c0beb0144076546873597ac0?
Or I'm missing something?

angyangie:
I just figured that out @darkdarkdragon :( . I assumed this ticket hadn't been done yet https://github.com/livepeer/go-livepeer/issues/851. Thanks for pointing it out

darkdarkdragon:
@angyangie Sorry for not mentioning it early. I've saw this, and just assumed you saw it too :( Need to be more verbose...

angyangie:
No worries @darkdarkdragon. This is actually a great teaching technique.... implement a solution on my own first, then learn from seeing another's solution, taking note of things I missed, etc.

discovery/webhook_discovery.go
```diff
@@ -0,0 +1,137 @@
...
+type orchAddress struct {
+	address string `json:address`
+}
+
+func NewWebhookOrchestratorPoolCache(node *core.LivepeerNode, orchWebhookUrl string) *WHOrchestratorPoolCache {
+	_ = cacheWebhookTranscoders(node, orchWebhookUrl)
+
+	ticker := getWebhookTicker()
+	go func(node *core.LivepeerNode) {
+		for _ = range ticker.C {
```
j0sh, line 38: I'm not so sure about periodic polling via a ticker for a few reasons:

* The current 1 minute update frequency is very high for data that will hardly ever change. We could decrease this frequency, but...
* The vast majority of webhook calls will be unnecessary, especially if the broadcaster is idle. This puts avoidable strain on the webhook endpoint if there are many broadcasters in play. This is especially undesirable if the webhook itself is a pay-per-call endpoint such as a file in S3.
* Updates won't be picked up until the next cache refresh interval.

My preferred approach is to make the call out only when needed, eg during session manager refreshes. This might add a bit of latency to the call itself, but as https://github.com/livepeer/go-livepeer/pull/1007 mentions, that can be mitigated for subsequent call-outs by a cache (and 1 minute would be a reasonable timeout for this cache), and the original cache can be reused if something goes wrong with the webhook request. (Nice work here reusing the DB as a cache! I was just thinking of storing the list in memory.)

discovery/webhook_discovery.go
```diff
@@ -0,0 +1,137 @@
...
+func cacheWebhookTranscoders(node *core.LivepeerNode, webhookURL string) error {
+	orchAddresses, err := getOrchAddresses(webhookURL)
+	if err != nil {
+		glog.Error("Could not refresh webhook list of orchestrators: ", err)
+		return err
+	}
+
+	for _, addr := range orchAddresses {
+		DBOrch := common.NewDBOrch(addr, addr)
+		if err := node.Database.UpdateOrch(DBOrch); err != nil {
```
j0sh, line 108: Given the refresh frequency, re-populating the DB each time seems excessive. (We really want to minimize writes as those are slow with SQLite.) Increasing the refresh interval has tradeoffs as mentioned earlier.

If the approach in https://github.com/livepeer/go-livepeer/pull/1007/ is taken, then the DB can be updated only if the contents change. For example, we can store a hash of the response along with the timestamp of the last webhook call.

Also, can we reuse the preexisting `cacheDBOrchs` function within DB? I'm unsure of the extent we might have to make accommodations there for offchain use, but I'd also like to minimize the number of DB entry points in this area.

discovery/webhook_discovery.go
```diff
@@ -0,0 +1,137 @@
...
+			glog.Error("Error updating Orchestrator in webhook: ", err)
+		}
+	}
+
+	return nil
+}
+
+func (who *WHOrchestratorPoolCache) GetURLs() []*url.URL {
+	orchs, err := who.node.Database.SelectOrchs(&common.DBOrchFilter{MaxPrice: server.BroadcastCfg.MaxPrice()})
```
j0sh, line 117: Just checking, does this still work OK in offchain mode when there's no price?

j0sh (commented):
Interesting approach! Storing in the DB may be a helpful resiliency measure in case the webhook API itself is down when the node starts up. Some notes along those lines:

I'd suggest building on top of https://github.com/livepeer/go-livepeer/pull/1007 since that incorporates some changes, especially around the DB pool, that would make this approach cleaner (eg, the use of `getURLS()` to minimize the repeated calls to db.SelectOrchs which is error-prone given its parameterization)

How much of the existing DB pool code can we reuse? Something like:

```golang
type webhookPool struct {
  db *DBOrchestratorPoolCache // except don't fetch the Os from blockchain!
}
func (w *webhookPool) GetURLs() []*url.URL {
  return w.db.GetURLs()
}
```

This pattern can be extended to other methods in the interface as well. As noted in earlier, I'm unsure how much it would take to accommodate offchain mode with the existing DB pool, but I think it's desirable to minimize DB entry points.

j0sh:
Resolved by https://github.com/livepeer/go-livepeer/pull/1007

Issue #1041, opened 2019-08-09
Refactor eth.RoundService to only poll without creating a subscription


yondonfu:
**Is your feature request related to a problem? Please describe.**

`eth.RoundService` currently subscribes to a remote ETH node for new blocks and will check if the current round is initialized before trying to initialize the round. Since we are moving away from pub-sub subscriptions with remote ETH nodes we want to remove the event monitor from `eth.RoundService`.

**Describe the solution you'd like**

The only utility of the event monitor in `eth.RoundService` is to determine the frequency at which it should check if the round is initialized. `eth.RoundService` doesn't really need the event monitor to do this - it can set a polling frequency that is close to the average block time of the network and then check if the round is initialized at that frequency.

PR #1046, opened 2019-08-15
Query for last seen block using blockheaders table

### LINKED ISSUE #1039: Use blockwatch to track last seen block

At the moment, the node uses `eth/eventservices/blockservice.go` to subscribe to new blocks and for each new block the service writes the block number to the DB. After #1038 is merged, we can use the `blockwatch` package instead.

The `blockwatch` package can offer the same functionality as `eth/eventservices/blockservice.go` out of the box - the initiated block watcher will write the latest N block headers (N is a configured parameter) to the DB which means that the latest block number of the latest header written to the DB will be the latest block number the node has seen.

We can do the following:

- Remove `eth/eventservices/blockservice.go`
- Update `(*db).LastSeenBlock()` to call `(*db).FindLatestMiniHeader()` and return `miniHeader.number`

yondonfu:

This PR updates `db.LastSeenBlock()` to query the last seen block number from the blockheaders table which is updated by `blockwatch.Watcher`. 

- Update `db.LastSeenBlock()` to return the number for the latest block header stored in the blockheaders table
- Remove `db.SetLastSeenBlock()` because `blockwatch.Watcher` takes care of updating the blockheaders table with new headers (which the last seen block number can be derived from). In order to remove this method, I also removed the test file for `UnbondingService` since it will no longer be needed after we refactor that file
- Remove `BlockService`

Updated unit tests.

Fixes #1039

**Checklist:**


common/db_test.go
```diff
@@ -27,53 +27,42 @@ func TestDBLastSeenBlock(t *testing.T) {
...
-		return
-	}
+	assert.Nil(err)
+	assert.Nil(blk)
+
+	// When there is a single header, return its number
+	h0 := defaultMiniHeader()
+	h0.Number = big.NewInt(100)
+	err = dbh.InsertMiniHeader(h0)
+	require.Nil(err)
```
kyriediculous, line 42: What's the difference here between using assert and require?

core/livepeernode.go
```diff
@@ -94,25 +94,13 @@ func NewLivepeerNode(e eth.LivepeerEthClient, wd string, dbh *common.DB) (*Livep
 
 func (n *LivepeerNode) StartEthServices() error {
 	var err error
-	for k, s := range n.EthServices {
-		// Skip BlockService until the end
-		if k == "BlockService" {
-			continue
-		}
+	for _, s := range n.EthServices {
 		err = s.Start(context.Background())
```
kyriediculous, line 98: What service are we starting here? aren't we removing the existing EthServices completely?

common/db_test.go
yondonfu: `require` will call `t.FailNow()` if the check evaluates to false (see the [require](https://godoc.org/github.com/stretchr/testify/require) package). Using `require` is similar to using `t.Fatal()` - if the check fails the test exits immediately rather than continuing to run which would be the case when using `assert` or `t.Error()`.

`require` is generally used in scenarios where a failing check would cause the rest of the test to provide little to no additional information (i.e. if the rest of the checks would always fail if the precondition specified by the `require` does not hold). In this case, we have already written unit tests for `InsertMiniHeader()` so we are assuming that it works correctly. So, the precondition that should hold is that `InsertMiniHeader()` does not return an error, otherwise the rest of the checks in the test will likely fail.

core/livepeernode.go
yondonfu: We want to eventually remove the existing `EthServices` completely once we finish refactoring the existing services. I was planning on refactoring each existing service in its own PR and then once all of them are refactored completely remove `EthServices`.

common/db_test.go
kyriediculous: Thanks for explaining ! 🥇

core/livepeernode.go
kyriediculous: Oh yeah that sounds very reasonable !

mkrufky, line 98: It would be much more efficient to declare `ctxt := context.Background()` before this range loop and pass `ctxt` into `Start()` instead, rather than fetching `context.Background()` each time.

mkrufky (changes requested):

core/livepeernode.go
kyriediculous: This code is related to the old event monitor and will eventually be removed once we refactor the two services, but you're totally right if we were to keep this!  Check out this comment: https://github.com/livepeer/go-livepeer/pull/1046#discussion_r317320941

yondonfu:
Resolved conflicts and ready for another review.

kyriediculous (approved):
LGTM :rocket: 

I suppose we can now safely remove `LivepeerNode.StartEthServices()` in a subsequent PR after merging this one?

yondonfu:
> I suppose we can now safely remove LivepeerNode.StartEthServices() in a subsequent PR after merging this one?

Yep we just need to finish implementing #1041

PR #1047, opened 2019-08-16
eth: EventDecoder for event watchers

### LINKED ISSUE #1043: Filter events by contract addresses when using blockwatch.Watcher

At the moment, `blockwatch.Watcher` accepts a slice of topics in the [Config](https://github.com/livepeer/go-livepeer/blob/master/eth/blockwatch/block_watcher.go#L50) struct that it is constructed with. These topics are then included in the filter used to query for logs. Since the filter does not specify contract addresses, an event signature collision between 2 contracts that use the same signature for an event (i.e. the `Transfer()` event in 2 ERC20 contracts) is possible.

See [this comment](https://github.com/livepeer/go-livepeer/issues/1040#issuecomment-520484177) for the discussion thread that this issue originated from.

Here are 2 possible solutions:

1. When decoding logs in a block, check if the contract address for the log is known (this is what the [0x event decoder](https://github.com/0xProject/0x-mesh/blob/development/zeroex/orderwatch/event_decoder.go) does). The event decoder would keep track of "known" contract addresses (i.e. contract addresses that the event service cares about)

2. Add `ContractAddresses` field to the `Config` struct that `blockwatch.Watcher` is constructed with. Then, these contract addresses could be included in the filter used to query for logs

kyriediculous:
Implement a generic event decoder according to the EventDecoder implemented in [this branch](https://github.com/livepeer/go-livepeer/compare/yf/refactor-unbondingservice#diff-3d0bf718180c420a3a3a7629e5e3bbf8R1).

The event decoder decodes
- Introduces a new EventDecoder type
- Introduces logic to decode events based on the provided contract ABI's events

unit tests

Fixes #1043 

**Checklist:**

### DIFF
--- a/eth/watchers/eventdecoder.go
+++ b/eth/watchers/eventdecoder.go
@@ -1,0 +1,64 @@
+package watchers
+
+import (
+	"errors"
+	"fmt"
+	"strings"
+
+	"github.com/ethereum/go-ethereum/accounts/abi"
+	"github.com/ethereum/go-ethereum/accounts/abi/bind"
+	ethcommon "github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+)
+
+// EventDecoder decodes logs into events for known contracts
+type EventDecoder struct {
+	addr             ethcommon.Address
+	contract         *bind.BoundContract
+	topicToEventName map[ethcommon.Hash]string
+}
+
+// NewEventDecoder returns a new instance of EventDecoder with a contract binding to the provided ABI string
+func NewEventDecoder(addr ethcommon.Address, abiJSON string) (*EventDecoder, error) {
+	abi, err := abi.JSON(strings.NewReader(abiJSON))
+	if err != nil {
+		return nil, err
+	}
+
+	topicToEventName := make(map[ethcommon.Hash]string)
+	for _, event := range abi.Events {
+		topicToEventName[event.Id()] = event.Name
+	}
+
+	return &EventDecoder{
+		addr: addr,
+		// Create BoundContract without a backend because we just need to access
+		// log unpacking without contract interaction
+		contract:         bind.NewBoundContract(addr, abi, nil, nil, nil),
+		topicToEventName: topicToEventName,
+	}, nil
+}
+
+// FindEventName returns the event name for a log. An error will be returned if the log is not emitted
+// from a known contract or if it does not map to a known event
+func (e *EventDecoder) FindEventName(log types.Log) (string, error) {
+	if log.Address != e.addr {
+		return "", errors.New("log not from known contract")
+	}
+	eventName, ok := e.topicToEventName[log.Topics[0]]
+	if !ok {
+		return "", fmt.Errorf("unknown event for %v", e.addr.Hex())
+	}
+
+	return eventName, nil
+}
+
+// Decode decodes a log into an event struct. An error will be returned if the log is not emitted
+// from a known contract or if it does not map to a known event
+func (e *EventDecoder) Decode(eventName string, log types.Log, decodedLog interface{}) error {
+	if log.Address != e.addr {
+		return errors.New("log not from known contract")
+
+	}
+	return e.contract.UnpackLog(decodedLog, eventName, log)
+}
--- a/eth/watchers/eventdecoder_test.go
+++ b/eth/watchers/eventdecoder_test.go
@@ -1,0 +1,75 @@
+package watchers
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/ethereum/go-ethereum/common"
+	ethcommon "github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/livepeer/go-livepeer/eth/contracts"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestTopicToEventName_AllEventsIncluded(t *testing.T) {
+	abi := "[{\"anonymous\": false,\"inputs\": [],\"name\": \"First\",\"type\": \"event\"},{\"anonymous\": false,\"inputs\": [],\"name\": \"Second\",\"type\": \"event\"},{\"anonymous\": false,\"inputs\": [],\"name\": \"Third\",\"type\": \"event\"}]"
+	assert := assert.New(t)
+
+	addr := ethcommon.HexToAddress("0x692a70d2e424a56d2c6c27aa97d1a86395877b3a")
+	dec, err := NewEventDecoder(addr, abi)
+	assert.Nil(err)
+	assert.Equal(dec.topicToEventName[crypto.Keccak256Hash([]byte("First()"))], "First")
+	assert.Equal(dec.topicToEventName[crypto.Keccak256Hash([]byte("Second()"))], "Second")
+	assert.Equal(dec.topicToEventName[crypto.Keccak256Hash([]byte("Third()"))], "Third")
+
+}
+
+func TestEventDecoder_FindEventType(t *testing.T) {
+	dec, err := NewEventDecoder(stubBondingManagerAddr, contracts.BondingManagerABI)
+	require.Nil(t, err)
+
+	assert := assert.New(t)
+
+	// Test unknown contract address
+	log := newStubBaseLog()
+	log.Address = common.BytesToAddress([]byte("foo"))
+	_, err = dec.FindEventName(log)
+	assert.EqualError(err, "log not from known contract")
+
+	// Test unknown contract event
+	log.Address = stubBondingManagerAddr
+	log.Topics = []common.Hash{common.BytesToHash([]byte("foo"))}
+	_, err = dec.FindEventName(log)
+	assert.EqualError(err, fmt.Sprintf("unknown event for %v", stubBondingManagerAddr.Hex()))
+
+	// Test known contract address
+	log = newStubUnbondLog()
+	eventName, err := dec.FindEventName(log)
+	assert.Nil(err)
+	assert.Equal("Unbond", eventName)
+}
+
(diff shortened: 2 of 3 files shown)


eth/watchers/eventdecoder.go
```diff
@@ -0,0 +1,73 @@
...
+type EventDecoder struct {
+	addr             ethcommon.Address
+	contract         *bind.BoundContract
+	topicToEventName map[ethcommon.Hash]string
+}
+
+// FindEventName returns the event name for a log. An error will be returned if the log is not emitted
+// from a known contract or if it does not map to a known event
+func (e *EventDecoder) FindEventName(log types.Log) (string, error) {
+	switch log.Address {
```
yondonfu: Since `EventDecoder` only cares about a single address can we use an `if` statement instead of a `switch` since there is only ever 2 cases (I originally intended for `EventDecoder` to handle multiple addresses which is why I used a `switch` statement, but I actually like keeping `EventDecoder` dedicated to a single address so the caller doesn't need to pass in multiple addresses if it only cares about one)?

Ex.

```
if log.Address != e.addr {
    // Return error
}
// Return event name
```

eth/watchers/eventdecoder.go
```diff
@@ -0,0 +1,73 @@
...
+		return e.findEventName(log)
+	default:
+		return "", errors.New("log not from known contract")
+	}
+}
+
+// Decode decodes a log into an event struct. An error will be returned if the log is not emitted
+// from a known contract or if it does not map to a known event
+func (e *EventDecoder) Decode(eventName string, log types.Log, decodedLog interface{}) error {
+	switch log.Address {
```
yondonfu: Same comment as above about using an `if` statement instead of a `switch` statement.

eth/watchers/eventdecoder.go
```diff
@@ -0,0 +1,73 @@
...
+	return &EventDecoder{
+		addr: addr,
+		// Create BoundContract without a backend because we just need to access
+		// log unpacking without contract interaction
+		contract:         bind.NewBoundContract(addr, abi, nil, nil, nil),
+		topicToEventName: topicToEventName,
+	}, nil
+}
+
+func (e *EventDecoder) decode(eventName string, log types.Log, decodedLog interface{}) error {
```
yondonfu: Since this is a one line method, I think we can in-line it into `Decode()`. WDYT?

eth/watchers/eventdecoder.go
```diff
@@ -0,0 +1,73 @@
...
+	return e.contract.UnpackLog(decodedLog, eventName, log)
+}
+
+func (e *EventDecoder) findEventName(log types.Log) (string, error) {
```
yondonfu: Perhaps we can in-line this into `FindEventName()` as well.

eth/watchers/eventdecoder.go
```diff
@@ -0,0 +1,73 @@
...
+	eventName, ok := e.topicToEventName[log.Topics[0]]
+	if !ok {
+		return "", errors.New("unknown event for BondingManager")
```
yondonfu: We should update this error message since `EventDecoder` is no longer specific to the BondingManager contract. Instead of providing an explicit name, since `EventDecoder` already has access to the contract address, we could change the error to something like this:

```
return "", fmt.Errorf("unknown event for %v", e.addr.Hex())
```

eth/watchers/eventdecoder_test.go
```diff
@@ -0,0 +1,74 @@
...
+	assert := assert.New(t)
+
+	// Test unknown contract address
+	log := newStubBaseLog()
+	log.Address = common.BytesToAddress([]byte("foo"))
+
+	_, err = dec.FindEventName(log)
+	assert.EqualError(err, "log not from known contract")
+
+	// Check that bondingManagerDecoder is used
```
yondonfu: We should update this comment. Perhaps something like "Check for known contract address"

eth/watchers/eventdecoder_test.go
```diff
@@ -0,0 +1,74 @@
...
+	log := newStubBaseLog()
+	log.Address = common.BytesToAddress([]byte("foo"))
+
+	var dummyEvent struct {
+		foo string
+	}
+	err = dec.Decode("foo", log, dummyEvent)
+	assert.EqualError(err, "log not from known contract")
+
+	// Check that bondingManagerDecoder is used
```
yondonfu: As mentioned above, we should update this comment.

eth/watchers/eventdecoder_test.go
```diff
@@ -0,0 +1,74 @@
...
+	log = newStubUnbondLog()
+
+	var unbondEvent contracts.BondingManagerUnbond
+	err = dec.Decode("Unbond", log, &unbondEvent)
+	assert.Nil(err)
+}
+
+func TestBondingManagerDecoder_FindEventType(t *testing.T) {
```
yondonfu: If we in-line the code block in `findEventType()` into `FindEventType()` then we can move these tests into `TestEventDecoder_FindEventType`

eth/watchers/eventdecoder_test.go
```diff
@@ -0,0 +1,74 @@
...
+	dec, err := NewEventDecoder(stubBondingManagerAddr, contracts.BondingManagerABI)
+	require.Nil(t, err)
+
+	assert := assert.New(t)
+
+	// Test unknown event for BondingManager
```
yondonfu: Comment needs to be updated, maybe something like "Test unknown event for contract"

eth/watchers/eventdecoder.go
```diff
@@ -0,0 +1,73 @@
...
+func (e *EventDecoder) Decode(eventName string, log types.Log, decodedLog interface{}) error {
+	switch log.Address {
+	case e.addr:
+		return e.decode(eventName, log, decodedLog)
+	default:
+		return errors.New("log not from known contract")
+	}
+}
+
+func NewEventDecoder(addr ethcommon.Address, abiJSON string) (*EventDecoder, error) {
```
yondonfu, line 22: Small style nit: could we move this constructor up after the type definition so that the layout is:

```
type definition -> constructor -> methods
```

yondonfu (changes requested):

eth/watchers/eventdecoder_test.go
kyriediculous: wdyt about having a seperate test like this for every contract ABI that we use with the event decoder or would that be overkill? 
This would mean renaming it to something like `TestEventDecoder_BondingManager_FindEventType`

Example in ca5b5af

eth/watchers/eventdecoder.go
kyriediculous: fixed in ca5b5af

eth/watchers/eventdecoder_test.go
kyriediculous: fixed in ca5b5af

eth/watchers/eventdecoder_test.go
kyriediculous: fixed in ca5b5af

eth/watchers/eventdecoder.go
kyriediculous: fixed in ca5b5af

eth/watchers/eventdecoder.go
kyriediculous: fixed in ca5b5af

eth/watchers/eventdecoder.go
kyriediculous: fixed in ca5b5af

eth/watchers/eventdecoder.go
kyriediculous: fixed in ca5b5af

eth/watchers/eventdecoder.go
kyriediculous: fixed in ca5b5af

eth/watchers/eventdecoder.go
```diff
@@ -0,0 +1,65 @@
...
+
+// Decode decodes a log into an event struct. An error will be returned if the log is not emitted
+// from a known contract or if it does not map to a known event
+func (e *EventDecoder) Decode(eventName string, log types.Log, decodedLog interface{}) error {
+	if log.Address != e.addr {
+		return errors.New("log not from known contract")
+
+	}
+	return e.contract.UnpackLog(decodedLog, eventName, log)
+
```
yondonfu: Can remove this new line

eth/watchers/eventdecoder_test.go
yondonfu: I'd be ok with not testing for every contract ABI right now because `FindEventType()` just does a mapping lookup (which has no direct dependence on the contract ABI used) and we're already checking that behavior (for known vs. unknown events). What might be worth testing is the logic that populates `e.topicToEventName` in `NewEventDecoder()` - the constructor should populate the mapping with *all* the events included in a contract ABI. A small test that might be worth adding is:

1. Create a custom ABI string with a few events (i.e. 2 or 3)
2. Pass the custom ABI string into `NewEventDecoder()`
3. Check if the values for `e.topicToEventName` are correct for the custom ABI string

eth/watchers/eventdecoder.go
kyriediculous: fixed in 7ce08e7

eth/watchers/eventdecoder_test.go
kyriediculous: fixed in 7ce08e7

eth/watchers/eventdecoder_test.go
kyriediculous: fixed and added test in 7ce08e7

yondonfu:
Changes look good! Needs a rebase.

kyriediculous:
Rebased !

yondonfu (approved):
LGTM 🚢

PR #1048, opened 2019-08-16
Use UnbondingWatcher instead of UnbondingService

### LINKED ISSUE #1040: Refactor eth.UnbondingService to use blockwatch

`eth.UnbondingService` currently creates a subscription with a remote ETH node in order to watch for unbond, rebond and withdraw events. Instead, it can subscribe to `blockwatch.Watcher` (all such services will rely on the same reliable stream of blocks from the watcher).

Refactor `eth/eventservices/unbondingservice.go` to use `blockwatch.Watcher`.

This will involve:

- Passing an instance of `blockwatch.Watcher` (that has already started watching for blocks) as an argument when creating `UnbondingService`
- `UnbondingService` will create a block subscription when it is started
- Add an event decoder implementation that uses the JSON ABI definitions for the unbond, rebond and withdraw events to decode logs that are emitted from the block subscription

yondonfu:

This PR implements `watchers.UnbondingWatcher` which replicates the existing functionality provided by `eventservices.UnbondingService`, but it uses logs emitted from `blockwatch.Watcher` and also handles removed logs caused by block re-orgs.

- Adds an `EventDecoder` type in 3fa085c. I suggest we merge #1047 first (which should be the same as this commit) because other event watchers besides the `UnbondingWatcher` will rely on this new type. If we merge #1047 first then this commit can be discarded prior to merging this PR
- Introduces the `eth/watchers` and adds `UnbondingWatcher`
- Updates the DB with methods to be used when a re-org occurs (i.e. delete a previously added unbonding lock and reverting a previously used unbonding lock)
- Removes the existing `eventservices.UnbondingService`
- Initialize the `blockwatch.Watcher` and `watchers.UnbondingWatcher` when the node starts in on-chain mode

Added unit tests and tested the unbonding/rebonding/withdrawal flow using the CLI with a node created using the devtool.

Fixes #1040

**Checklist:**


cmd/livepeer/livepeer.go
```diff
@@ -45,6 +48,14 @@ var (
 	ErrKeygen    = errors.New("ErrKeygen")
 	EthTxTimeout = 600 * time.Second
 
+	// The timeout for ETH RPC calls
+	ethRPCTimeout = 20 * time.Second
+
+	// The interval at which the block watcher polls for new blocks
+	blockWatcherPollingInterval = 1 * time.Second
```
kyriediculous: Wouldn't it perhaps be better to have a polling interval closer to the mainnet blocktime? We can always adjust this downward for testing locally. Wdyt?

common/db.go
```diff
@@ -379,9 +390,29 @@ func (db *DB) InsertUnbondingLock(id *big.Int, delegator ethcommon.Address, amou
 	return nil
 }
 
+// DeleteUnbondingLock deletes an unbonding lock from the DB with the given ID and delegator address. This method will
```
kyriediculous: Perhaps it's better to start the new sentence on a new line
```
// DeleteUnbondingLock deletes an unbonding lock from the DB with the given ID and delegator address. 
// This method will return nil for non-existent unbonding lock
```

common/db.go
```diff
@@ -379,9 +390,29 @@ func (db *DB) InsertUnbondingLock(id *big.Int, delegator ethcommon.Address, amou
...
+	glog.V(DEBUG).Infof("db: Deleting unbonding lock %v for delegator %v", id, delegator.Hex())
+	_, err := db.deleteUnbondingLock.Exec(id.Int64(), delegator.Hex())
+	if err != nil {
+		glog.Errorf("db: Error deleting unbonding lock %v for delegator %v: %v", id, delegator.Hex(), err)
+		return err
+	}
+	return nil
+}
+
+// UseUnbondingLock sets an unbonding lock in the DB as used by setting the lock's used block. If usedBlock is nil
```
kyriediculous: Perhaps it's better to start a the new sentence on a new line 

```
// UseUnbondingLock sets an unbonding lock in the DB as used by setting the lock's used block.
// If usedBlock is nil this method will set the lock's used block to NULL
```

eth/watchers/unbondingwatcher.go
```diff
@@ -0,0 +1,162 @@
...
+// Start kicks off a loop that handles events from a block subscription
+func (w *UnbondingWatcher) Start() {
+	blockEvents := make(chan []*blockwatch.Event, 10)
+	w.sub = w.bw.Subscribe(blockEvents)
+
+	for {
+		select {
+		case <-w.quit:
+			w.sub.Unsubscribe()
+			close(blockEvents)
```
kyriediculous: I don't think we should be closing the sink channel from the receiving side. I think unsubscribe will close the producer channel. This could lead to unwanted panicking by closing an already closed channel. 

```
// NewSubscription runs a producer function as a subscription in a new goroutine. The
// channel given to the producer is closed when Unsubscribe is called. If fn returns an
// error, it is sent on the subscription's error channel.
```

eth/watchers/unbondingwatcher.go
```diff
@@ -0,0 +1,162 @@
...
+
+// Start kicks off a loop that handles events from a block subscription
+func (w *UnbondingWatcher) Start() {
+	blockEvents := make(chan []*blockwatch.Event, 10)
+	w.sub = w.bw.Subscribe(blockEvents)
+
+	for {
+		select {
+		case <-w.quit:
+			w.sub.Unsubscribe()
```
kyriediculous: I have a slight preference for deferring the unsubscribe right after subscribing and simply returning from `Start()` when we close the `quit` channel so that we make sure we unsubscribe for all cases when the `Start()` function returns. WDYT?

eth/watchers/unbondingwatcher.go
```diff
@@ -0,0 +1,162 @@
...
+		addr:  addr,
+		bw:    bw,
+		store: store,
+		dec:   dec,
+		quit:  make(chan struct{}),
+	}, nil
+}
+
+// Start kicks off a loop that handles events from a block subscription
+func (w *UnbondingWatcher) Start() {
```
kyriediculous: I named this method `Watch` in my PR's. Perhaps it might be good to adopt ubiquitous naming here. I have no preference on naming here, so I'll leave it up to you.

kyriediculous (changes requested):

common/db.go
yondonfu: Fixed in 7ef1c97

common/db.go
yondonfu: Fixed in da84f63

eth/watchers/unbondingwatcher.go
yondonfu: Fixed in 3ba3b20

eth/watchers/unbondingwatcher.go
yondonfu: Fixed in 3ba3b20 and 68d75d7

eth/watchers/unbondingwatcher.go
yondonfu: The subscription here is actually the result of calling [(*Feed).Subscribe()](https://github.com/ethereum/go-ethereum/blob/master/event/feed.go) which accepts an `interface{}` channel which allows our sink channel to be of any event type. Under the hood, the subscription returned is implemented by `feedSub` and we can see [here](https://github.com/ethereum/go-ethereum/blob/master/event/feed.go#L199) that `Unsubscribe()` *does not* close the sink channel. Instead, `Unsubscribe()` closes the subscription's *error channel* and tells `*Feed` to [remove](https://github.com/ethereum/go-ethereum/blob/master/event/feed.go#L104) the subscription from its known set of subscriptions. When removing a subscription, `*Feed` does not touch the original sink channel - it just removes the subscription so that whenever the feed receives a new event, it will not send to the sink channel for that subscription. 

Re: the `NewSubscription` comment - note that the channel accepted by the producer function is a `struct{}` type. This channel is not meant to be used as the sink channel (which we know needs to be able to accept any event type). The channel accepted by the producer is actually used to signal to the producer function that `Unsubscribe()` was called and is not used to carry values to be emitted for the subscription.

cmd/livepeer/livepeer.go
yondonfu: I agree that we should eventually change the polling interval to be closer to mainnet block time, but since we're still at a stage where we're mainly running nodes in environments with fast block times (and potentially even eventually auto-deploying builds to these environments as our devops setup improves) I think its ok to keep the polling interval as is for now and once we move to a stage where most of our testing is in environments with more realistic block times then we can switch. WDYT?

yondonfu:
There are a bunch of conflicts that will need to be resolved, but I'll wait to resolve the comments from the initial code review first. After the comments have been resolved, I'll fix the conflicts and we can do a follow up review based on the fixed up branch.

cmd/livepeer/livepeer.go
kyriediculous: Sounds good !

eth/watchers/unbondingwatcher.go
kyriediculous: I see what you are saying. But wouldn't closing the sink channel from the receiver side lead to potential unwanted behaviour? 

Isn't it an unwritten rule to only close channels from the sender side? Not sure anymore where I got that from. 

Altough I guess if we remove the subscription the sink channel the sender will not be sending values into the channel anymore so we can safely close it.

yondonfu: > Isn't it an unwritten rule to only close channels from the sender side?

After a second thought, I do agree with you. It is generally a good convention to only have senders close channels.

My original desire to make sure to close the channel after it is done being used was misplaced - there actually isn't any need to close a channel because if it is not being used GC will clean it up. The primary reason for closing a channel is to signal to a consumer that no more values will be emitted. In this case, since the consumer is the one that decides it doesn't care about values anymore, it also doesn't need a signal from a closed channel.

> Although I guess if we remove the subscription the sink channel the sender will not be sending values into the channel anymore so we can safely close it.

While this is true, I think you make a good point about following the convention of only having senders close channels. Thanks for bringing it up.

Fixed in b680940

kyriediculous:
Changes look good to me, let's resolve the conflicts !

yondonfu:
Conflicts have been resolved via a rebase.

Added 2 additional commits:

b487980: Return nil for unknown events in the UnbondingWatcher. Otherwise, anytime a `NewRound` event (or any event that the UnbondingWatcher doesn't care about) is emitted, the node will log an unknown event error. In this scenario, an error isn't actually being observed - the UnbondingWatcher just doesn't care about the `NewRound` event

b74b762: Return nil for unknown events in the RoundsWatcher. Otherwise, anytime a unbonding event (or any event that the RoundsWatcher doesn't care about) is emitted, the node will log an unknown event error. In this scenario, an error isn't actually being observed - the RoundsWatcher just doesn't care about any other events besides `NewRound`

kyriediculous (commented):
Resolved conflicts look good, there's one fixup commit (b487980d9ba931daf89b3f64badc121ed36609c4)  left that needs rebasing it seems. After that should be good to merge !

yondonfu:
Rebased

kyriediculous (approved):
LGTM 🚀 !

PR #1052, opened 2019-08-19
eth, cmd: roundswatcher with cache

### LINKED ISSUE #946: pm: RoundsManager implementation that maintains a cache

At the moment, both [pm.Sender](https://github.com/livepeer/go-livepeer/blob/master/pm/sender.go#L34) and [pm.Validator](https://github.com/livepeer/go-livepeer/blob/master/pm/validator.go#L35) rely on the `RoundsManager` interface in order to query the latest round and its associated block hash. This interface is currently implemented by [eth/client.go](https://github.com/livepeer/go-livepeer/blob/master/eth/client.go#L129) which will execute a RPC call against an ETH node in order to fetch the values. This is fine for preliminary testing, but is sub-optimal in a production setting because both senders and recipients of PM tickets will need to execute a RPC call for each ticket sent and received.

We can add an implementation of the `RoundsManager` interface that maintains cached values for the current round and its associated block hash. The implementation could initialize the cache and then monitor contract events for new round and block hash values that would be used to update its cache.

We could run a Graph node in parallel with the Livepeer node which would handle the event monitoring. However, this adds additional devops overhead for a node operator.

**Additional context**

kyriediculous:
This PR adds an implementation of the `RoundsManager` interface that watches for on-chain event logs using the `blockwatch` package introduced in #1037. 

When a new block is added that contains event logs for the `NewRound` event, we update the cached values accordingly. When a re-org occurs we make an RPC call to a remote ethereum node to fetch the most up-to-date contract data. 

- Created a type alias `RoundsWatcher` adhering to the `RoundsManager` interface that represents a concurrency safe cache 
- Initialize a `RoundsWatcher` on node startup and pass it into the constructor for a `Sender`
- Add unit tests for the new `RoundsWatcher` type 
- added the `NewRound` event signature to `watchers/topics.go`

Added & ran unit tests 

Fixes #946 

**Checklist:**

### DIFF
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -21,6 +21,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/ethereum/go-ethereum/rpc"
 	"github.com/livepeer/go-livepeer/pm"
 
 	ipfslogging "gx/ipfs/QmSpJByNKFX1sCsHBEp3R73FL4NF6FnQTEGyNAXHm2GS52/go-log"
@@ -33,7 +34,9 @@ import (
 	"github.com/livepeer/go-livepeer/discovery"
 	"github.com/livepeer/go-livepeer/drivers"
 	"github.com/livepeer/go-livepeer/eth"
+	"github.com/livepeer/go-livepeer/eth/blockwatch"
 	"github.com/livepeer/go-livepeer/eth/eventservices"
+	"github.com/livepeer/go-livepeer/eth/watchers"
 
 	//"github.com/livepeer/go-livepeer/ipfs" until we re-enable IPFS
 	lpmon "github.com/livepeer/go-livepeer/monitor"
@@ -44,6 +47,14 @@ import (
 var (
 	ErrKeygen    = errors.New("ErrKeygen")
 	EthTxTimeout = 600 * time.Second
+
+	// The timeout for ETH RPC calls
+	ethRPCTimeout = 20 * time.Second
+
+	// The interval at which the block watcher polls for new blocks
+	blockWatcherPollingInterval = 1 * time.Second
+	// The maximum block sfor the block watcher to retain
+	blockWatcherRetentionLimit = 20
 
 	// The gas required to redeem a PM ticket
 	redeemGas = 100000
@@ -330,6 +341,49 @@ func main() {
 		// Setup unbonding service to manage unbonding locks
 		n.EthServices["UnbondingService"] = eventservices.NewUnbondingService(n.Eth, dbh)
 
+		// Initialize block watcher that will emit logs used by event watchers
+		blockWatcherClient, err := blockwatch.NewRPCClient(*ethUrl, ethRPCTimeout)
+		if err != nil {
+			glog.Errorf("Failed to setup blockwatch client: %v", err)
+			return
+		}
+		topics := watchers.FilterTopics()
+		blockWatcherCfg := blockwatch.Config{
+			Store:               n.Database,
+			PollingInterval:     blockWatcherPollingInterval,
+			StartBlockDepth:     rpc.LatestBlockNumber,
+			BlockRetentionLimit: blockWatcherRetentionLimit,
+			WithLogs:            true,
+			Topics:              topics,
+			Client:              blockWatcherClient,
+		}
+		// Wait until all event watchers have been initialized before starting the block watcher
+		blockWatcher := blockwatch.New(blockWatcherCfg)
+
+		roundsWatcher, err := watchers.NewRoundsWatcher(addrMap["RoundsManager"], blockWatcher, n.Eth)
+		if err != nil {
+			glog.Errorf("Failed to setup roundswatcher: %v", err)
+			return
+		}
+		go roundsWatcher.Watch()
+		defer roundsWatcher.Stop()
+
+		blockWatchCtx, cancel := context.WithCancel(context.Background())
+		defer cancel()
+
+		// Backfill events that the node has missed since its last seen block. This method will block
+		// and the node will not continue setup until it finishes
+		if err := blockWatcher.BackfillEventsIfNeeded(blockWatchCtx); err != nil {
+			glog.Errorf("Failed to backfill events: %v", err)
+			return
+		}
+
+		go func() {
+			if err := blockWatcher.Watch(blockWatchCtx); err != nil {
+				glog.Errorf("block watcher error: %v", err)
+			}
+		}()
+
 		n.Balances = core.NewBalances(cleanupInterval)
 
 		if *orchestrator {
@@ -368,7 +422,7 @@ func main() {
 			sigVerifier := &pm.DefaultSigVerifier{}
 			// TODO: Initialize Validator with an implementation
 			// of RoundsManager that reads from a cache
-			validator := pm.NewValidator(sigVerifier, n.Eth)
+			validator := pm.NewValidator(sigVerifier, roundsWatcher)
 			gpm := eth.NewGasPriceMonitor(backend, gpmPollingInterval)
 			// Start gas price monitor
 			gasPriceUpdate, err := gpm.Start(context.Background())
@@ -434,7 +488,7 @@ func main() {
 			// of RoundsManager that reads from a cache
 			// TODO: Initialize Sender with an implementation
 			// of SenderManager that reads from a cache
-			n.Sender = pm.NewSender(n.Eth, n.Eth, n.Eth, ev, *depositMultiplier)
+			n.Sender = pm.NewSender(n.Eth, roundsWatcher, n.Eth, ev, *depositMultiplier)
 
 			if *pixelsPerUnit <= 0 {
 				// Can't divide by 0
--- a/eth/watchers/roundswatcher.go
+++ b/eth/watchers/roundswatcher.go
@@ -1,0 +1,151 @@
+package watchers
+
+import (
+	"fmt"
+	"log"
+	"math/big"
+	"sync"
+
+	ethcommon "github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/livepeer/go-livepeer/eth/contracts"
+
+	"github.com/golang/glog"
+	"github.com/livepeer/go-livepeer/eth"
+	"github.com/livepeer/go-livepeer/eth/blockwatch"
+)
+
+// RoundsWatcher is a type for a thread safe in-memory cache that watches on-chain events for new initialized rounds
+type RoundsWatcher struct {
+	mu                       sync.RWMutex
+	lastInitializedRound     *big.Int
+	lastInitializedBlockHash [32]byte
+
+	quit chan struct{}
+
+	watcher BlockWatcher
+	lpEth   eth.LivepeerEthClient
+	dec     *EventDecoder
+}
+
+// NewRoundsWatcher creates a new instance of RoundsWatcher and sets the initial cache through an RPC call to an ethereum node
+func NewRoundsWatcher(roundsManagerAddr ethcommon.Address, watcher BlockWatcher, lpEth eth.LivepeerEthClient) (*RoundsWatcher, error) {
+	dec, err := NewEventDecoder(roundsManagerAddr, contracts.RoundsManagerABI)
+	if err != nil {
+		return nil, err
+	}
+	lr, err := lpEth.LastInitializedRound()
+	if err != nil {
+		return nil, err
+	}
+	bh, err := lpEth.BlockHashForRound(lr)
+	if err != nil {
+		return nil, err
+	}
+
+	return &RoundsWatcher{
+		quit:                     make(chan struct{}),
+		watcher:                  watcher,
+		lpEth:                    lpEth,
+		lastInitializedRound:     lr,
+		lastInitializedBlockHash: bh,
+		dec: dec,
+	}, nil
+}
+
+// LastInitializedRound gets the last initialized round from cache
+func (rw *RoundsWatcher) LastInitializedRound() *big.Int {
+	rw.mu.RLock()
+	defer rw.mu.RUnlock()
+	return rw.lastInitializedRound
+}
+
+// LastInitializedBlockHash returns the blockhash of the block the last round was initiated in
+func (rw *RoundsWatcher) LastInitializedBlockHash() [32]byte {
+	rw.mu.RLock()
+	defer rw.mu.RUnlock()
+	return rw.lastInitializedBlockHash
+}
+
+func (rw *RoundsWatcher) setLastInitializedRound(round *big.Int, hash [32]byte) {
+	rw.mu.Lock()
+	defer rw.mu.Unlock()
+	rw.lastInitializedRound = round
+	rw.lastInitializedBlockHash = hash
+}
+
+// Watch the blockwatch subscription for NewRound events
+func (rw *RoundsWatcher) Watch() {
+	lr, err := rw.lpEth.LastInitializedRound()
+	if err != nil {
+		glog.Errorf("error fetching initial lastInitializedRound value: %v", err)
+	}
+
+	bh, err := rw.lpEth.BlockHashForRound(lr)
+	if err != nil {
+		glog.Errorf("error fetching initial lastInitializedBlockHash value: %v", err)
+	}
+	rw.setLastInitializedRound(lr, bh)
+
+	events := make(chan []*blockwatch.Event, 10)
+	sub := rw.watcher.Subscribe(events)
+	defer sub.Unsubscribe()
+	for {
+		select {
+		case <-rw.quit:
+			return
+		case err := <-sub.Err():
+			log.Fatal(err)
+		case events := <-events:
+			rw.handleBlockEvents(events)
+		}
+	}
+}
+
+// Stop watching for NewRound events
+func (rw *RoundsWatcher) Stop() {
+	close(rw.quit)
+}
+
+func (rw *RoundsWatcher) handleBlockEvents(events []*blockwatch.Event) {
+	for _, event := range events {
+		for _, log := range event.BlockHeader.Logs {
+			if event.Type == blockwatch.Removed {
+				log.Removed = true
+			}
+			if err := rw.handleLog(log); err != nil {
+				glog.Error(err)
+			}
+		}
+	}
+}
+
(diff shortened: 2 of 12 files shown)


cmd/livepeer/livepeer.go
```diff
@@ -330,6 +341,49 @@ func main() {
...
+			StartBlockDepth:     rpc.LatestBlockNumber,
+			BlockRetentionLimit: blockWatcherRetentionLimit,
+			WithLogs:            true,
+			Topics:              topics,
+			Client:              blockWatcherClient,
+		}
+		// Wait until all event watchers have been initialized before starting the block watcher
+		blockWatcher := blockwatch.New(blockWatcherCfg)
+
+		blockWatchCtx, cancel := context.WithCancel(context.Background())
```
yondonfu, line 371: Let's move the logic for backfilling and starting the block watcher to the end of the if clause for on-chain mode so that we can ensure that we always start the event watchers before the block watcher even if certain event watchers in the future are specific to a particular node type (i.e. orchestrator or broadcaster only). See the [UnbondingWatcher PR](https://github.com/livepeer/go-livepeer/pull/1048/files#diff-205b550ef0ac99b368da16cfb883960bR497).

cmd/livepeer/livepeer.go
```diff
@@ -330,6 +341,49 @@ func main() {
...
+				glog.Errorf("block watcher error: %v", err)
+			}
+		}()
+
+		roundsWatcher, err := watchers.NewRoundsWatcher(blockWatcher, n.Eth, addrMap["RoundsManager"])
+		if err != nil {
+			glog.Errorf("Failed to setup roundswatcher: %v", err)
+			return
+		}
+		go roundsWatcher.Watch()
```
yondonfu: Let's start the round watcher before starting the block watcher (and generally we can start event watchers before the block watcher) to ensure that the round watcher is ready when the block watcher starts emitting blocks.

eth/watchers/roundswatcher.go
```diff
@@ -0,0 +1,160 @@
...
+	"github.com/golang/glog"
+	"github.com/livepeer/go-livepeer/eth"
+	"github.com/livepeer/go-livepeer/eth/blockwatch"
+)
+
+// RoundsWatcher is a type alias for a thread safe in-memory cache that watches on-chain events for new initialized rounds
+type RoundsWatcher struct {
+	mu                   sync.RWMutex
+	lastInitializedRound *big.Int
+	blockHashForRound    [32]byte
```
yondonfu: Let's call this `lastInitializedBlockHash`?

eth/watchers/roundswatcher.go
```diff
@@ -0,0 +1,160 @@
...
+
+	return &RoundsWatcher{
+		quit:    make(chan struct{}),
+		watcher: watcher,
+		lpEth:   lpEth,
+		dec:     dec,
+	}, nil
+}
+
+// LastInitializedRound gets the last initialized from cache
```
yondonfu: "last initialized" -> "last initialized round"

eth/watchers/topics.go
```diff
@@ -0,0 +1,23 @@
...
+import (
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/crypto"
+)
+
+var eventSignatures = []string{
+	"Unbond(address,address,uint256,uint256,uint256)",
+	"Rebond(address,address,uint256,uint256)",
+	"WithdrawStake(address,uint256,uint256,uint256)",
+	"NewRound(uint256 round, bytes32 blockHash)",
```
yondonfu: The event topic hash omits parameter names and spaces so the event signature for `NewRound` should be:

```
NewRound(uint256,bytes32)
```

eth/watchers/roundswatcher.go
```diff
@@ -0,0 +1,160 @@
...
+	blockHashForRound    [32]byte
+
+	quit chan struct{}
+
+	watcher Watcher
+	lpEth   eth.LivepeerEthClient
+	dec     *EventDecoder
+}
+
+type Watcher interface {
```
yondonfu: Can we move this to a separate file since it is used by multiple event watcher implementations and call it `BlockWatcher`? Maybe a `types.go` file.

eth/watchers/roundswatcher.go
```diff
@@ -0,0 +1,160 @@
...
+	Subscribe(sink chan<- []*blockwatch.Event) event.Subscription
+}
+
+// NewRoundsWatcher creates a new instance of RoundsWatcher and sets the initial cache through an RPC call to an ethereum node
+func NewRoundsWatcher(watcher Watcher, lpEth eth.LivepeerEthClient, addr ethcommon.Address) (*RoundsWatcher, error) {
```
yondonfu: Can we rename `addr` to something like `roundsManagerAddr` to make it clear that the address is supposed to represent a deployed RoundsManager contract? Could we also have the address before the interface dependencies just to be consistent with the constructors of [existing implementations](https://github.com/livepeer/go-livepeer/pull/1048/files#diff-fd62975d7c6fadd8b06b097faddf0074R33)?

eth/watchers/roundswatcher.go
```diff
@@ -0,0 +1,160 @@
...
+}
+
+func (rw *RoundsWatcher) handleBlockEvents(events []*blockwatch.Event) {
+	for _, event := range events {
+		for _, log := range event.BlockHeader.Logs {
+			nr, err := rw.handleLog(log)
+			if err != nil {
+				glog.Error(err)
+			}
+			if event.Type == blockwatch.Added && nr.Round != nil {
```
yondonfu: Can we move all of this logic into `handleLog` since it is all related to processing a log? We can set `log.Removed` based on `event.Type == blockwatch.Removed` so that we just need to pass the log into `handleLog()`

Another benefit of this is that the unit tests will be easier to write because you can invoke `handleLog()` and check all the relevant error paths and any you can just check the returned error.

eth/watchers/roundswatcher.go
```diff
@@ -0,0 +1,160 @@
...
+
+// LastInitializedRound gets the last initialized from cache
+func (rw *RoundsWatcher) LastInitializedRound() (*big.Int, error) {
+	rw.mu.RLock()
+	defer rw.mu.RUnlock()
+	return rw.lastInitializedRound, nil
+}
+
+// BlockHashForRound gets the blockhash for the last initialized round from cache
+func (rw *RoundsWatcher) BlockHashForRound(round *big.Int) ([32]byte, error) {
```
yondonfu: I suggest that we change this to `LastInitializedBlockHash()` and omit any arguments so that we don't have to worry about argument checks. Then, separately we can change the `pm.RoundsManager` interface method (the interface was originally written so that `eth.LivepeerEthClient` could be passed in directly, but now we can pass in `RoundsWatcher` so I think it is reasonable to also update to a cleaner API).

eth/watchers/roundswatcher.go
```diff
@@ -0,0 +1,160 @@
...
+		return rw.lpEth.BlockHashForRound(round)
+	}
+	rw.mu.RLock()
+	defer rw.mu.RUnlock()
+	return rw.blockHashForRound, nil
+}
+
+// setLastInitializedRound sets the last initialized round in cache
+func (rw *RoundsWatcher) setLastInitializedRound(round *big.Int) {
+	rw.mu.RLock()
```
yondonfu: Should be a write lock

eth/watchers/roundswatcher.go
```diff
@@ -0,0 +1,160 @@
...
+	defer rw.mu.RUnlock()
+	rw.lastInitializedRound = round
+}
+
+//setBlochashForRound sets the blockhash for the last initialized round in cache
+func (rw *RoundsWatcher) setBlockHashForRound(hash [32]byte) {
+	rw.mu.RLock()
```
yondonfu: Should be a write lock

eth/watchers/roundswatcher.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+	rw.setLastInitializedRound(lr)
+
+	bh, err := rw.lpEth.BlockHashForRound(lr)
+	if err != nil {
+		glog.Errorf("error fetching initial blockHashForRound value: %v", err)
+	}
+	rw.setBlockHashForRound(bh)
+
+	events := make(chan []*blockwatch.Event)
```
yondonfu: We should make this a buffered channel so that additional block events can be queued up even if the rounds watcher is still processing existing block events. The UnbondingWatcher currently uses a [buffer size of 10](https://github.com/livepeer/go-livepeer/pull/1048/files#diff-fd62975d7c6fadd8b06b097faddf0074R50).

eth/watchers/roundswatcher.go
```diff
@@ -0,0 +1,160 @@
...
+	for _, event := range events {
+		for _, log := range event.BlockHeader.Logs {
+			nr, err := rw.handleLog(log)
+			if err != nil {
+				glog.Error(err)
+			}
+			if event.Type == blockwatch.Added && nr.Round != nil {
+				rw.setLastInitializedRound(nr.Round)
+				rw.setBlockHashForRound(nr.BlockHash)
+			} else if event.Type == blockwatch.Removed && nr.Round != nil {
```
yondonfu: Is `nr.Round` ever nil besides when there is an error unpacking a log?

eth/watchers/roundswatcher_test.go
```diff
@@ -0,0 +1,135 @@
...
+
+func TestWatchAndStop(t *testing.T) {
+	assert := assert.New(t)
+	lpEth := &eth.StubClient{}
+	watcher := &stubBlockWatcher{}
+	rw, err := NewRoundsWatcher(watcher, lpEth, stubRoundsManagerAddr)
+	assert.Nil(err)
+
+	header := defaultMiniHeader()
+	round := [32]byte{0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 8}
```
yondonfu: For consistency can we use a `newStubNewRoundLog()` method to return the log here? See [examples](https://github.com/livepeer/go-livepeer/pull/1048/files#diff-c5c00511e3b96fae6be6a993099cb207R43) of other stub logs that are pulled from mainnet

eth/watchers/roundswatcher.go
```diff
@@ -0,0 +1,160 @@
...
+	if lastRound.Cmp(round) != 0 {
+		return rw.lpEth.BlockHashForRound(round)
+	}
+	rw.mu.RLock()
+	defer rw.mu.RUnlock()
+	return rw.blockHashForRound, nil
+}
+
+// setLastInitializedRound sets the last initialized round in cache
+func (rw *RoundsWatcher) setLastInitializedRound(round *big.Int) {
```
yondonfu: I think we should use a single setter that locks -> updates the round and block hash -> unlocks. Otherwise, it could be possible for a reader to read an updated round, but stale block hash:

Ex.

Writer updates old round to new round
Reader fetches new round
Reader fetches old block hash
Writer updates old block hash to new block hash

yondonfu (changes requested):

eth/watchers/roundswatcher.go
kyriediculous: No, this is a leftover from when I was using abi.Unpack rather than the event decoder. Can be deleted.

eth/watchers/roundswatcher.go
kyriediculous: I have reworked both methods then to not return an error either, since we're just reading values and don't have to make external calls.

I addressed the `pm` package specific changes in a seperate commit (and fixup): c32dba0 & 315ca2b

eth/watchers/roundswatcher.go
kyriediculous: fixed in 80d3e25

eth/watchers/roundswatcher_test.go
kyriediculous: fixed in 80d3e25

eth/watchers/roundswatcher.go
kyriediculous: fixed in 80d3e25

eth/watchers/roundswatcher.go
kyriediculous: fixed in 80d3e25

eth/watchers/roundswatcher.go
kyriediculous: now using a single setter in 80d3e25

eth/watchers/roundswatcher.go
kyriediculous: fixed in 80d3e25

eth/watchers/roundswatcher.go
kyriediculous: fixed in 80d3e25

eth/watchers/roundswatcher.go
kyriediculous: fixed in 80d3e25

eth/watchers/topics.go
kyriediculous: woops, totally right. fixed in 80d3e25

eth/watchers/roundswatcher.go
kyriediculous: fixed in 80d3e25

eth/watchers/roundswatcher.go
kyriediculous: fixed in 80d3e25

cmd/livepeer/livepeer.go
kyriediculous: fixed in 5399d9d

cmd/livepeer/livepeer.go
kyriediculous: fixed in 5399d9d

eth/watchers/roundswatcher.go
```diff
@@ -0,0 +1,140 @@
...
+	ethcommon "github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/livepeer/go-livepeer/eth/contracts"
+
+	"github.com/golang/glog"
+	"github.com/livepeer/go-livepeer/eth"
+	"github.com/livepeer/go-livepeer/eth/blockwatch"
+)
+
+// RoundsWatcher is a type alias for a thread safe in-memory cache that watches on-chain events for new initialized rounds
```
yondonfu: I don't think using the term "type alias" here is quite right - in Go (see Go 1.9 [release notes](https://golang.org/doc/go1.9#language) for the introduction of type aliases), type alias refers to a name `T1` that is an alternate spelling of the concrete type `T2` which you would declare as `type T1 = T2`. In this scenario, we're just defining a new type.

eth/watchers/stub.go
```diff
@@ -35,3 +39,37 @@ func newStubUnbondLog() types.Log {
 
 	return log
 }
+
+func newStubNewRoundLog() types.Log {
```
yondonfu, line 46: Can we structure this function in a similar manner as the other stub log methods? i.e. Use hex strings and add comments for what the non-event signature topics + the log data decodes to

cmd/livepeer/livepeer.go
```diff
@@ -330,6 +341,49 @@ func main() {
...
+
+		// Backfill events that the node has missed since its last seen block. This method will block
+		// and the node will not continue setup until it finishes
+		if err := blockWatcher.BackfillEventsIfNeeded(blockWatchCtx); err != nil {
+			glog.Errorf("Failed to backfill events: %v", err)
+			return
+		}
+
+		go func() {
+			if err := blockWatcher.Watch(blockWatchCtx); err != nil {
```
yondonfu, line 382: If there is an error starting the block watcher in on-chain mode we should probably just exit - in this scenario, nodes will not receive blocks/logs. I think it's ok to address that in a separate PR though as we'll need to handle errors from any of the other watchers similarly.

eth/watchers/roundswatcher.go
```diff
@@ -0,0 +1,140 @@
...
+	rw.mu.Lock()
+	defer rw.mu.Unlock()
+	rw.lastInitializedRound = round
+	rw.lastInitializedBlockHash = hash
+}
+
+// Watch the blockwatch subscription for NewRound events
+func (rw *RoundsWatcher) Watch() {
+	lr, err := rw.lpEth.LastInitializedRound()
+	if err != nil {
```
yondonfu, line 69: If any of the initial requests to populate the cache fail then we should probably propagate the error to the caller because there is a good chance that there is a problem with the remote ETH node and because otherwise if we continue there will be a period of time where any `RoundWatcher` user will get nil back for the last initialized round which can trigger a nil pointer error.

If we return an error from `Watch()` then we can handle any errors returned. But, similar to my comment about handling errors from the block watcher, I think we can address that in a separate PR.

yondonfu (changes requested):

eth/watchers/stub.go
kyriediculous: I'm not using real-world logs but arbitrary values, in which case it seemed less tedious and more robust to encode the topics and data from scratch. 

What I can do is not hardcode the roundnumber as a byte slice but do something like

`common.LeftPadBytes(big.NewInt(8).Bytes(), 32)`

Or I could figure out the hex string representations first, it's as you wish. 

I was actually doing something similar for the senderWatcher:

```
func newStubDepositFundedLog() types.Log {
	log := newStubBaseLog()
	log.Address = stubTicketBrokerAddr
	amount, _ := new(big.Int).SetString("5000000000000000000", 10)
	amountData := common.LeftPadBytes(amount.Bytes(), 32)
	sender := common.LeftPadBytes(stubSender.Bytes(), 32)
	var senderTopic common.Hash
	copy(senderTopic[:], sender[:])
	log.Topics = []common.Hash{
		crypto.Keccak256Hash([]byte("DepositFunded(address,uint256)")),
		senderTopic,
	}
	log.Data = amountData
	return log
}
```

yondonfu:
The style of the code example provided is fine by me. Just prefer not to
use hard coded byte slices as they are difficult to read (and copy/paste) -
hex strings or code generated values makes tests easier to review IMO.

yondonfu (commented):
Change look good. Needs a rebase.

yondonfu (approved):
LGTM 🚢 (I rebased and pushed)

PR #1054, opened 2019-08-21
eth, cmd: SenderWatcher with cache

### LINKED ISSUE #947: pm: SenderManager implementation that maintains a cache

#944 introduces the `SenderManager` interface which exposes a method for fetching on-chain information about sender including its deposit and reserve. `pm.Sender` uses the `SenderManager` interface and in the PR the interface is implemented by [eth/client.go](https://github.com/livepeer/go-livepeer/blob/master/eth/client.go#L129) which will execute a RPC call against an ETH node in order to fetch the sender information. This is fine for preliminary testing, but is sub-optimal in a production setting because the sender will need to execute a RPC call for each ticket sent and received.

We can add an implementation of the `SenderManager` interface that maintains cached values for the sender's deposit and reserve. The implementation could initialize the cache and then monitor contract events that contain values for the sender's deposit and/or reserve after each on-chain update that would be used to update the cache.

We could run a Graph node in parallel with the Livepeer node which would handle the event monitoring. However, this adds additional devops overhead for a node operator.

**Additional context**

kyriediculous:
This PR implements a senderwatcher that adheres to the `SenderManager` interface with a concurrency safe map that keeps track of `pm.SenderInfo` for an ethereum address. 

In this case the map will keep track of a single entry for the broadcaster.

- Created a `SenderWatcher` type that adheres to the `SenderManager` interfaces and used it to initiate `pm.Sender` 
- Init a `SenderWatcher` instance watching for on-chain events log on node startup 
- Added unit tests for `SenderWatcher`
- Added event signatures to `watchers/topics.go`
- added a `*pm.SenderInfo` field to the `eth.stubClient` so that we can test for values returned by stubbed out RPC calls made by `SenderWatcher`. Now returning the new value from `stubClient.GetSenderInfo()` as well

created unit tests

Fixes #947 

**Checklist:**

### DIFF
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -380,6 +380,14 @@ func main() {
 		go unbondingWatcher.Watch()
 		defer unbondingWatcher.Stop()
 
+		senderWatcher, err := watchers.NewSenderWatcher(addrMap["TicketBroker"], blockWatcher, n.Eth)
+		if err != nil {
+			glog.Errorf("Failed to setup senderwatcher: %v", err)
+			return
+		}
+		go senderWatcher.Watch()
+		defer senderWatcher.Stop()
+
 		blockWatchCtx, cancel := context.WithCancel(context.Background())
 		defer cancel()
 
@@ -519,11 +527,7 @@ func main() {
 				panic(fmt.Errorf("-depositMultiplier must be greater than 0, but %v provided. Restart the node with a valid value for -depositMultiplier", *depositMultiplier))
 			}
 
-			// TODO: Initialize Sender with an implementation
-			// of RoundsManager that reads from a cache
-			// TODO: Initialize Sender with an implementation
-			// of SenderManager that reads from a cache
-			n.Sender = pm.NewSender(n.Eth, roundsWatcher, n.Eth, ev, *depositMultiplier)
+			n.Sender = pm.NewSender(n.Eth, roundsWatcher, senderWatcher, ev, *depositMultiplier)
 
 			if *pixelsPerUnit <= 0 {
 				// Can't divide by 0
--- a/eth/watchers/senderwatcher.go
+++ b/eth/watchers/senderwatcher.go
@@ -1,0 +1,218 @@
+package watchers
+
+import (
+	"fmt"
+	"math/big"
+	"sync"
+
+	ethcommon "github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/golang/glog"
+	"github.com/livepeer/go-livepeer/eth"
+	"github.com/livepeer/go-livepeer/eth/blockwatch"
+	"github.com/livepeer/go-livepeer/eth/contracts"
+	"github.com/livepeer/go-livepeer/pm"
+)
+
+// SenderWatcher maintains a concurrency-safe map with SenderInfo
+type SenderWatcher struct {
+	senders map[ethcommon.Address]*pm.SenderInfo
+	mu      sync.RWMutex
+	quit    chan struct{}
+	watcher BlockWatcher
+	lpEth   eth.LivepeerEthClient
+	dec     *EventDecoder
+}
+
+// NewSenderWatcher initiates a new SenderWatcher
+func NewSenderWatcher(ticketBrokerAddr ethcommon.Address, watcher BlockWatcher, lpEth eth.LivepeerEthClient) (*SenderWatcher, error) {
+	dec, err := NewEventDecoder(ticketBrokerAddr, contracts.TicketBrokerABI)
+	if err != nil {
+		return nil, err
+	}
+	return &SenderWatcher{
+		quit:    make(chan struct{}),
+		watcher: watcher,
+		lpEth:   lpEth,
+		senders: make(map[ethcommon.Address]*pm.SenderInfo),
+		dec:     dec,
+	}, nil
+}
+
+// GetSenderInfo returns the senderInfo for a sender
+// if values for a sender are not cached an RPC call to a remote ethereum node will be made to initialize the cache
+func (sw *SenderWatcher) GetSenderInfo(addr ethcommon.Address) (*pm.SenderInfo, error) {
+	sw.mu.RLock()
+	cache := sw.senders[addr]
+	sw.mu.RUnlock()
+	if cache == nil {
+		info, err := sw.lpEth.GetSenderInfo(addr)
+		if err != nil {
+			return nil, fmt.Errorf("GetSenderInfo RPC call to remote node failed: %v", err)
+		}
+		sw.setSenderInfo(addr, info)
+		return info, nil
+	}
+	return cache, nil
+}
+
+func (sw *SenderWatcher) setSenderInfo(addr ethcommon.Address, info *pm.SenderInfo) {
+	sw.mu.Lock()
+	defer sw.mu.Unlock()
+	sw.senders[addr] = info
+}
+
+// Watch starts the event watching loop
+func (sw *SenderWatcher) Watch() {
+	events := make(chan []*blockwatch.Event, 10)
+	sub := sw.watcher.Subscribe(events)
+	defer sub.Unsubscribe()
+	for {
+		select {
+		case <-sw.quit:
+			return
+		case err := <-sub.Err():
+			glog.Error(err)
+		case events := <-events:
+			sw.handleBlockEvents(events)
+		}
+	}
+}
+
+// Stop watching for events
+func (sw *SenderWatcher) Stop() {
+	close(sw.quit)
+}
+
+// Clear removes a key-value pair from the map
+func (sw *SenderWatcher) Clear(addr ethcommon.Address) {
+	sw.mu.Lock()
+	defer sw.mu.Unlock()
+	if _, ok := sw.senders[addr]; ok {
+		delete(sw.senders, addr)
+	}
+}
+
+func (sw *SenderWatcher) handleBlockEvents(events []*blockwatch.Event) {
+	for _, event := range events {
+		for _, log := range event.BlockHeader.Logs {
+			if event.Type == blockwatch.Removed {
+				log.Removed = true
+			}
+			if err := sw.handleLog(log); err != nil {
+				glog.Error(err)
+			}
+		}
+	}
+}
+
+func (sw *SenderWatcher) handleLog(log types.Log) error {
+	eventName, err := sw.dec.FindEventName(log)
+	if err != nil {
+		// Noop if we cannot find the event name
+		return nil
+	}
+
+	sw.mu.Lock()
+	defer sw.mu.Unlock()
+
+	var sender ethcommon.Address
+	switch eventName {
+	case "DepositFunded":
+		var depositFunded contracts.TicketBrokerDepositFunded
+		if err := sw.dec.Decode("DepositFunded", log, &depositFunded); err != nil {
+			return fmt.Errorf("failed to decode DepositFunded event: %v", err)
+		}
+		sender = depositFunded.Sender
+		if info, ok := sw.senders[sender]; ok && !log.Removed {
+			info.Deposit.Add(info.Deposit, depositFunded.Amount)
+		}
+	case "ReserveFunded":
+		var reserveFunded contracts.TicketBrokerReserveFunded
+		if err := sw.dec.Decode("ReserveFunded", log, &reserveFunded); err != nil {
+			return fmt.Errorf("failed to decode ReserveFunded event: %v", err)
+		}
+		sender = reserveFunded.ReserveHolder
+		if info, ok := sw.senders[sender]; ok && !log.Removed {
+			info.Reserve.Add(info.Reserve, reserveFunded.Amount)
+		}
+	case "Withdrawal":
+		var withdrawal contracts.TicketBrokerWithdrawal
+		if err := sw.dec.Decode("Withdrawal", log, &withdrawal); err != nil {
+			return fmt.Errorf("failed to decode Withdrawal event: %v", err)
+		}
+		sender = withdrawal.Sender
+		if info, ok := sw.senders[sender]; ok && !log.Removed {
+			info.Deposit = big.NewInt(0)
+			info.Reserve = big.NewInt(0)
+		}
+	case "WinningTicketTransfer":
+		var winningTicketTransfer contracts.TicketBrokerWinningTicketTransfer
+		if err := sw.dec.Decode("WinningTicketTransfer", log, &winningTicketTransfer); err != nil {
+			return fmt.Errorf("failed to decode WinningTicketTransfer event: %v", err)
+		}
+		amount := winningTicketTransfer.Amount
+		sender = winningTicketTransfer.Sender
+
+		if info, ok := sw.senders[sender]; ok && !log.Removed {
+			// See if amount > deposit
+			if info.Deposit.Cmp(amount) < 0 {
+				// Draw from reserve
+				// ReserveFrozen will be handled in it's own event log handler
+				diff := new(big.Int).Sub(amount, info.Deposit)
+				info.Deposit = big.NewInt(0)
+				info.Reserve.Sub(info.Reserve, diff)
+			} else {
+				// Draw from deposit
+				info.Deposit.Sub(info.Deposit, amount)
+			}
+		}
+	case "ReserveFrozen":
+		// Set reserveStatus and thawround
+		var reserveFrozen contracts.TicketBrokerReserveFrozen
+		if err := sw.dec.Decode("ReserveFrozen", log, &reserveFrozen); err != nil {
+			return fmt.Errorf("failed to decode ReserveFrozen event: %v", err)
+		}
+		sender = reserveFrozen.ReserveHolder
+		if info, ok := sw.senders[sender]; ok && !log.Removed {
+			info.ReserveState = pm.Frozen
+			// TODO: fetch freezePeriod instead of hardcoding or use a const
+			// TODO: how to handle unthaw
+			// frozen checks will have to be against ReserveFrozen and ThawRound < CurrentRound
+			// But how do we handle state updates when the reserve thaws?
+			info.ThawRound.Add(reserveFrozen.FreezeRound, big.NewInt(2))
+		}
+	case "Unlock":
+		// Set withdraw block
+		var unlock contracts.TicketBrokerUnlock
+		if err := sw.dec.Decode("Unlock", log, &unlock); err != nil {
+			return fmt.Errorf("failed to decode Unlock event: %v", err)
+		}
+		sender = unlock.Sender
+		if info, ok := sw.senders[sender]; ok && !log.Removed {
+			info.WithdrawBlock = unlock.EndBlock
+		}
+	case "UnlockCancelled":
+		// Unset withdrawblock
+		var unlockCancelled contracts.TicketBrokerUnlockCancelled
+		if err := sw.dec.Decode("UnlockCancelled", log, &unlockCancelled); err != nil {
+			return fmt.Errorf("failed to decode UnlockCancelled event: %v", err)
+		}
+		sender = unlockCancelled.Sender
+		if info, ok := sw.senders[sender]; ok && !log.Removed {
+			info.WithdrawBlock = big.NewInt(0)
+		}
+	default:
+		return nil
+	}
+
+	if _, ok := sw.senders[sender]; ok && log.Removed {
+		info, err := sw.lpEth.GetSenderInfo(sender)
+		if err != nil {
+			return fmt.Errorf("GetSenderInfo RPC call to remote node failed: %v", err)
+		}
+		sw.senders[sender] = info
+	}
+
+	return nil
+}
--- a/eth/watchers/senderwatcher_test.go
+++ b/eth/watchers/senderwatcher_test.go
@@ -1,0 +1,576 @@
+package watchers
+
+import (
+	"math/big"
+	"testing"
+	"time"
+
+	ethcommon "github.com/ethereum/go-ethereum/common"
+	"github.com/livepeer/go-livepeer/eth"
+	"github.com/livepeer/go-livepeer/eth/blockwatch"
+	"github.com/livepeer/go-livepeer/pm"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestGetAndSetSenderInfo(t *testing.T) {
+	assert := assert.New(t)
+	sw := &SenderWatcher{
+		senders: make(map[ethcommon.Address]*pm.SenderInfo),
+		lpEth: &eth.StubClient{
+			SenderInfo: &pm.SenderInfo{
+				Deposit: big.NewInt(10),
+				Reserve: big.NewInt(5),
+			},
+		},
+	}
+
+	info := &pm.SenderInfo{
+		Deposit:      big.NewInt(1000000000000000000),
+		Reserve:      big.NewInt(1000000000000000000),
+		ReserveState: pm.NotFrozen,
+	}
+	sender := pm.RandAddress()
+
+	sw.setSenderInfo(sender, info)
+	entry, err := sw.GetSenderInfo(sender)
+	assert.Nil(err)
+	assert.Equal(sw.senders[sender], info)
+	assert.Equal(info, entry)
+
+	// If map entry is nil and we call get, make RPC call and set the returned data
+	sw.Clear(sender)
+	info, err = sw.GetSenderInfo(sender)
+	assert.Nil(err)
+	assert.Zero(info.Deposit.Cmp(big.NewInt(10)))
+	assert.Zero(info.Reserve.Cmp(big.NewInt(5)))
+	info = sw.senders[sender]
+	assert.Zero(info.Deposit.Cmp(big.NewInt(10)))
+	assert.Zero(info.Reserve.Cmp(big.NewInt(5)))
+}
+
+func TestSenderWatcher_Clear(t *testing.T) {
+	assert := assert.New(t)
+	stubClientSenderInfo := &pm.SenderInfo{
+		Deposit: big.NewInt(10),
+		Reserve: big.NewInt(5),
+	}
+	sw := &SenderWatcher{
+		senders: make(map[ethcommon.Address]*pm.SenderInfo),
+		lpEth: &eth.StubClient{
+			SenderInfo: stubClientSenderInfo,
+		},
+	}
+	info := &pm.SenderInfo{
+		Deposit:      big.NewInt(1000000000000000000),
+		Reserve:      big.NewInt(1000000000000000000),
+		ReserveState: pm.NotFrozen,
+	}
+	sender := pm.RandAddress()
+
+	sw.setSenderInfo(sender, info)
+
+	getInfo, err := sw.GetSenderInfo(sender)
+	assert.Nil(err)
+	assert.Equal(info, getInfo)
+
+	sw.Clear(sender)
+
+	getInfo = sw.senders[sender]
+	assert.Nil(getInfo)
+	// on calling GetSenderInfo should be back to RPC stub values
+	getInfo, err = sw.GetSenderInfo(sender)
+	assert.Nil(err)
+	assert.Equal(getInfo, stubClientSenderInfo)
+}
+
+func TestSenderWatcher_WatchAndStop(t *testing.T) {
+	assert := assert.New(t)
+	lpEth := &eth.StubClient{}
+	watcher := &stubBlockWatcher{}
+	sw, err := NewSenderWatcher(stubTicketBrokerAddr, watcher, lpEth)
+	assert.Nil(err)
+
+	go sw.Watch()
+	time.Sleep(2 * time.Millisecond)
+	sw.Stop()
+	time.Sleep(2 * time.Millisecond)
+	assert.True(watcher.sub.unsubscribed)
+}
+
+func TestSenderWatcher_HandleLog(t *testing.T) {
+	lpEth := &eth.StubClient{}
+	watcher := &stubBlockWatcher{}
+	rw, err := NewSenderWatcher(stubTicketBrokerAddr, watcher, lpEth)
+	require.Nil(t, err)
+
+	assert := assert.New(t)
+
+	// Test unknown event
+	log := newStubBaseLog()
+	log.Topics = []ethcommon.Hash{ethcommon.BytesToHash([]byte("foo"))}
+
+	err = rw.handleLog(log)
+	assert.Nil(err)
+}
+
+func TestFundDepositEvent(t *testing.T) {
+	assert := assert.New(t)
+	startDeposit := big.NewInt(10)
+	lpEth := &eth.StubClient{
+		SenderInfo: &pm.SenderInfo{
+			Deposit: big.NewInt(10),
+			Reserve: big.NewInt(5),
+		},
+	}
+	watcher := &stubBlockWatcher{}
+	sw, err := NewSenderWatcher(stubTicketBrokerAddr, watcher, lpEth)
+	assert.Nil(err)
+
+	header := defaultMiniHeader()
+	newDepositEvent := newStubDepositFundedLog()
+	header.Logs = append(header.Logs, newDepositEvent)
+
+	blockEvent := &blockwatch.Event{
+		Type:        blockwatch.Added,
+		BlockHeader: header,
+	}
+
+	go sw.Watch()
+	defer sw.Stop()
+	time.Sleep(2 * time.Millisecond)
+
+	// If map entry is nil don't handle event
+	watcher.sink <- []*blockwatch.Event{blockEvent}
+	time.Sleep(2 * time.Millisecond)
+	_, ok := sw.senders[stubSender]
+	assert.False(ok)
+
+	// if block removed, make RPC call
+	blockEvent.Type = blockwatch.Removed
+	sw.senders[stubSender] = &pm.SenderInfo{}
+	watcher.sink <- []*blockwatch.Event{blockEvent}
+	time.Sleep(2 * time.Millisecond)
+	info, ok := sw.senders[stubSender]
+	assert.True(ok)
+	assert.Zero(info.Deposit.Cmp(startDeposit))
+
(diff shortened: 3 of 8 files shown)


eth/watchers/senderwatcher.go
```diff
@@ -0,0 +1,248 @@
...
+	senders map[ethcommon.Address]*pm.SenderInfo
+	mu      sync.RWMutex
+	quit    chan struct{}
+	watcher BlockWatcher
+	lpEth   eth.LivepeerEthClient
+	dec     *EventDecoder
+}
+
+// NewSenderWatcher initiates a new SenderWatcher
+func NewSenderWatcher(addr ethcommon.Address, watcher BlockWatcher, lpEth eth.LivepeerEthClient) (*SenderWatcher, error) {
```
yondonfu: Let's rename `addr` to `ticketBrokerAddr` to make it clear that it is the TicketBroker contract address

eth/watchers/senderwatcher.go
```diff
@@ -0,0 +1,248 @@
...
+
+func (sw *SenderWatcher) setSenderInfo(addr ethcommon.Address, info *pm.SenderInfo) {
+	sw.mu.Lock()
+	defer sw.mu.Unlock()
+	sw.senders[addr] = info
+}
+
+// Watch starts the event watching loop
+func (sw *SenderWatcher) Watch() {
+	events := make(chan []*blockwatch.Event)
```
yondonfu: Let's make this a buffered channel. We can use a buffer size of 10 for now since that is the value we use for other event watchers.

eth/watchers/senderwatcher.go
```diff
@@ -0,0 +1,248 @@
...
+	sub := sw.watcher.Subscribe(events)
+	defer sub.Unsubscribe()
+	for {
+		select {
+		case <-sw.quit:
+			return
+		case err := <-sub.Err():
+			log.Fatal(err)
```
yondonfu: Why are we using `log.Fatal()` here? We are just logging the error if an error is received from the subscription channel in other watchers so we should be able to do the same here.

eth/watchers/senderwatcher.go
```diff
@@ -0,0 +1,248 @@
...
+	}
+}
+
+// Stop watching for events
+func (sw *SenderWatcher) Stop() {
+	close(sw.quit)
+}
+
+// Clear removes a key-value pair from the map
+// this can be invoked by the SenderMonitor cleanup loop
```
yondonfu: I don't think we need to mention the SenderMonitor in this comment as that type is external to this package.

cmd/livepeer/livepeer.go
```diff
@@ -434,7 +488,7 @@ func main() {
 			// of RoundsManager that reads from a cache
 			// TODO: Initialize Sender with an implementation
 			// of SenderManager that reads from a cache
-			n.Sender = pm.NewSender(n.Eth, n.Eth, n.Eth, ev, *depositMultiplier)
+			n.Sender = pm.NewSender(n.Eth, n.Eth, senderWatcher, ev, *depositMultiplier)
```
j0sh: Should the TODO note before this line be updated?

🎉 for starting to differentiate the NewSender parameters :)

kyriediculous: Yes it should, missed that ! Fixed it up in 40d04c9.

eth/watchers/senderwatcher.go
```diff
@@ -0,0 +1,248 @@
...
+	ethcommon "github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/golang/glog"
+	"github.com/livepeer/go-livepeer/eth"
+	"github.com/livepeer/go-livepeer/eth/blockwatch"
+	"github.com/livepeer/go-livepeer/eth/contracts"
+	"github.com/livepeer/go-livepeer/pm"
+)
+
+// SenderWatcher maintains a concurrency-safe map with SenderInfo
```
yondonfu, line 17: I don't think we need to mention that the `SenderManager` interface here since it is external to this package.

eth/watchers/senderwatcher.go
```diff
@@ -0,0 +1,248 @@
...
+				glog.Error(err)
+			}
+		}
+	}
+}
+
+func (sw *SenderWatcher) handleLog(log types.Log) error {
+	eventName, err := sw.dec.FindEventName(log)
+	if err != nil {
+		return fmt.Errorf("could not find event name: %v", err)
```
yondonfu: Let's return nil here? See [this comment](https://github.com/livepeer/go-livepeer/pull/1048#issuecomment-525027670)

eth/watchers/senderwatcher.go
```diff
@@ -0,0 +1,248 @@
...
+	switch eventName {
+	case "DepositFunded":
+		var depositFunded contracts.TicketBrokerDepositFunded
+		if err := sw.dec.Decode("DepositFunded", log, &depositFunded); err != nil {
+			return fmt.Errorf("failed to decode DepositFunded event: %v", err)
+		}
+		sender := depositFunded.Sender
+		if sw.senders[sender] != nil {
+			sw.senders[sender].Deposit = sw.senders[sender].Deposit.Add(sw.senders[sender].Deposit, depositFunded.Amount)
+		} else if sw.senders[sender] == nil || log.Removed {
```
yondonfu: If we only start watching for events for a sender address after the sender has already been cached via an initial call to `sw.GetSenderInfo()` (see my comment further up in this review), then I don't think we need to do these nil checks for `sw.senders[sender]` here. Same comment applies for the rest of the `sw.senders[sender]` nil checks in `handleLog()`

eth/watchers/senderwatcher.go
```diff
@@ -0,0 +1,248 @@
...
+	defer sw.mu.RUnlock()
+	switch eventName {
+	case "DepositFunded":
+		var depositFunded contracts.TicketBrokerDepositFunded
+		if err := sw.dec.Decode("DepositFunded", log, &depositFunded); err != nil {
+			return fmt.Errorf("failed to decode DepositFunded event: %v", err)
+		}
+		sender := depositFunded.Sender
+		if sw.senders[sender] != nil {
+			sw.senders[sender].Deposit = sw.senders[sender].Deposit.Add(sw.senders[sender].Deposit, depositFunded.Amount)
```
yondonfu: This could be shortened to `sw.senders[sender].Deposit.Add(sw.senders[sender].Deposit, depositFunded.Amount)`

eth/watchers/senderwatcher.go
```diff
@@ -0,0 +1,248 @@
...
+	}
+}
+
+func (sw *SenderWatcher) handleLog(log types.Log) error {
+	eventName, err := sw.dec.FindEventName(log)
+	if err != nil {
+		return fmt.Errorf("could not find event name: %v", err)
+	}
+
+	sw.mu.RLock()
```
yondonfu: Needs to be a write lock if we update the cache

eth/watchers/senderwatcher.go
```diff
@@ -0,0 +1,248 @@
...
+		watcher: watcher,
+		lpEth:   lpEth,
+		senders: make(map[ethcommon.Address]*pm.SenderInfo),
+		dec:     dec,
+	}, nil
+}
+
+// GetSenderInfo returns the senderInfo for a sender
+// if values for a sender are not cached an RPC call to a remote ethereum node will be made to initialize the cache
+func (sw *SenderWatcher) GetSenderInfo(addr ethcommon.Address) (*pm.SenderInfo, error) {
```
yondonfu, line 44: Right now we're going to cache every single sender that is involved in one of the on-chain events that we're watching for even if we don't care about them. We could instead maintain a set of known addresses for senders that we care about and only cache those senders. One way to do this is: during the first call to `sw.GetSenderInfo()` add the sender's address to an internal map of known addresses. Then, when we're handling logs, we check if the sender address in the log is known and skip the logs that have unknown sender addresses.

eth/watchers/senderwatcher.go
```diff
@@ -0,0 +1,248 @@
...
+			sw.senders[sender] = info
+		}
+	case "ReserveFunded":
+		var reserveFunded contracts.TicketBrokerReserveFunded
+		if err := sw.dec.Decode("ReserveFunded", log, &reserveFunded); err != nil {
+			return fmt.Errorf("failed to decode ReserveFunded event: %v", err)
+		}
+		sender := reserveFunded.ReserveHolder
+		if sw.senders[sender] != nil {
+			sw.senders[sender].Reserve = sw.senders[sender].Reserve.Add(sw.senders[sender].Reserve, reserveFunded.Amount)
```
yondonfu: This could be shortened to `sw.senders[sender].Reserve.Add(sw.senders[sender].Reserve, reserveFunded.Amount)`

eth/watchers/senderwatcher.go
```diff
@@ -0,0 +1,248 @@
...
+		var withdrawal contracts.TicketBrokerWithdrawal
+		if err := sw.dec.Decode("Withdrawal", log, &withdrawal); err != nil {
+			return fmt.Errorf("failed to decode Withdrawal event: %v", err)
+		}
+		sender := withdrawal.Sender
+		if _, ok := sw.senders[sender]; ok && !log.Removed {
+			sw.senders[sender].Deposit = big.NewInt(0)
+			sw.senders[sender].Reserve = big.NewInt(0)
+		}
+		if log.Removed {
```
yondonfu: It seems that we have the same logic for `log.Removed` no matter what event we are dealing with (i.e. fetching the sender info and updating the cache). Perhaps we can extract the logic into a helper. The helper could also be used by `sw.GetSenderInfo()`

yondonfu (changes requested):

eth/watchers/senderwatcher.go
kyriediculous: Good catch, remainder from when I was prototyping this outside of go-livepeer scope. Fixed in 4e646b8

eth/watchers/senderwatcher.go
kyriediculous: fixed in 4e646b8

Did the same for other similar expressions in this PR

eth/watchers/senderwatcher.go
kyriediculous: fixed in 4e646b8

eth/watchers/senderwatcher.go
kyriediculous: fixed in 4e646b8

eth/watchers/senderwatcher.go
kyriediculous: fixed in 4e646b8

eth/watchers/senderwatcher.go
kyriediculous: fixed in 4e646b8

eth/watchers/senderwatcher.go
kyriediculous: Fixed in 4e646b8

eth/watchers/senderwatcher.go
kyriediculous: fixed in 4e646b8

eth/watchers/senderwatcher.go
kyriediculous: fixed in 4e646b8

eth/watchers/senderwatcher.go
kyriediculous: Instead of creating an additional map, could we instead check if the key exists in `SenderWatcher.senders` when handling logs? 

My thought process being: 

For use of this watcher on the broadcaster side, there will always only be one entry (B's address) present in the map after the first `SenderWatcher.GetSenderInfo()` call. 

For use of this watcher on the orchestrator side we will initialize a map entry for a new sender upon our first call of `SenderWatcher.GetSenderInfo()` at which point we'll start handling events pertaining to that sender. When that map entry gets removed by the SenderMonitor cleanup loop (TODO!) we will no longer process events for that sender. 

Keep in mind that we'd still need to decode the event log to be able to get the necessary key to check if that entry exists on the `SenderWatcher.senders` map.

eth/watchers/senderwatcher.go
kyriediculous: We can do that but If we ever add additional events from `TicketBroker` to another watcher we will make RPC calls for removed logs even if they do not pertain to the `SenderWatcher` since the eventDecoder is bound to an ABI. 

Altough one way to circumvent this I think would be to add a default case that simply returns `nil` and doesn't do anything else ? 

We'd also have to add a check now to the expressions in the switch statements to not do state updates when the log is removed (something that should have perhaps been included in the initial PR as well) otherwise we'd be making a state update and then overwriting it with an RPC call.  We could also encapsulate the switch in an if else block , but go idiomatics seem to argue against too many indentation levels.

We do however still need to decode the event log , whether it is removed or not to be able to get the address from the sender to use as a key when setting state on `SenderWatcher.senders` 

We could extract the RPC call and state update into a helper, but doing so would only save us one line of code (setting the state) everytime we do this operation compared to just using a bit of replication. After the change to handling log removals outside of the switch statement we'd only be duplicating this logic twice and we'd have to call an additional getter in `SenderWatcher.GetSenderInfo`  So in the end using a helper here would (for now) result in actually more code than the current way. WDYT?

eth/watchers/senderwatcher.go
yondonfu: Checking if an address already exists in the `SenderWatcher.senders` map sounds good to me.

eth/watchers/senderwatcher.go
yondonfu: I actually mistyped in my original comment - I meant to only suggest extracting the `log.Removed` logic into a helper and *not* handling `log.Removed` outside of the switch statement. Updated my original comment to reflect that. If we extract the logic into a helper then we can reduce some of the repetition in all of the `log.Removed` clauses within the switch statement as well as for the caching logic in `sw.GetSenderInfo()`.

kyriediculous: I took a slightly different approach in bc0fff2 and interested to know what you think. 

I ended up extracting the logic for `log.Removed` out of the switch statement.

Now we will decode the event logs to get `sender` but we'll only parse state updates when `sender` already exists in the `SenderWatcher.senders` map and `log.Removed == false`.

Since we scope `sender` outside of the switch block now, we can handle `log.Removed == true` outside of the switch statement without duplicating the logic.

We can still extract this logic into a helper `fetchAndSetSenderInfo()` but it seems that would require some juggling with the locks and unlocks in `GetSenderInfo()` which might end up being uglier than just duplicating a few lines.

eth/watchers/senderwatcher.go
```diff
@@ -0,0 +1,216 @@
...
+		}
+	case "UnlockCancelled":
+		// Unset withdrawblock
+		var unlockCancelled contracts.TicketBrokerUnlockCancelled
+		if err := sw.dec.Decode("UnlockCancelled", log, &unlockCancelled); err != nil {
+			return fmt.Errorf("failed to decode UnlockCancelled event: %v", err)
+		}
+		sender = unlockCancelled.Sender
+		if _, ok := sw.senders[sender]; ok && !log.Removed {
+			sw.senders[sender].WithdrawBlock = nil
```
yondonfu: Should we be setting `WithdrawBlock` to 0 instead since its default value in the EVM will be 0?

eth/watchers/senderwatcher.go
```diff
@@ -0,0 +1,216 @@
...
+		}
+	default:
+		return nil
```
yondonfu, line 206: Return an error here? [Example](https://github.com/livepeer/go-livepeer/blob/master/eth/watchers/unbondingwatcher.go#L149)

eth/watchers/senderwatcher.go
```diff
@@ -0,0 +1,216 @@
...
+
+	var sender ethcommon.Address
+	switch eventName {
+	case "DepositFunded":
+		var depositFunded contracts.TicketBrokerDepositFunded
+		if err := sw.dec.Decode("DepositFunded", log, &depositFunded); err != nil {
+			return fmt.Errorf("failed to decode DepositFunded event: %v", err)
+		}
+		sender = depositFunded.Sender
+		if _, ok := sw.senders[sender]; ok && !log.Removed {
```
yondonfu: If you set both return value variables i.e. `info, ok := sw.senders[sender]`, then the statement within this if clause could be shortened since you can reference `info.Deposit` instead of reading `sw.senders[sender]` again.

Same comment applies for the rest of the events in this switch clause.

yondonfu (changes requested):
A few tests are failing for the `eth/watchers` package. I just realized that the `bash test.sh` command used in CI is not running the tests for the `eth/watchers` package right now. We can replace the command for running the tests for `eth/eventservices` with a command for running the tests for `eth/watchers` [here](https://github.com/livepeer/go-livepeer/blob/master/test.sh#L25).

eth/watchers/senderwatcher.go
kyriediculous: Shouldn't this be a no-op instead similar to when the eventName can't be found? 

The default case here essentially means that the event exists on the ABI but we don't care about it. It's not that we don't want to decode it as the error message in the example implies but rather that we have no need for it. So to me a no-op seems to be a better fit here?

eth/watchers/senderwatcher.go
kyriediculous: right, pointer 😅 fixed in f06a0ec

eth/watchers/senderwatcher.go
kyriediculous: Yeah that seems like a good idea, fixed in f06a0ec

kyriediculous:
replaced `eth/eventServices` with `eth/watchers` in `test.sh` in cf2f512

Fixed up the tests in f06a0ec by calling `SenderWatcher.GetSenderInfo()` when setting up the tests for each log to init an entry for `stubSender` on the `SenderWatcher.senders` map now that we only do state updates for addresses that are already present in the map.

eth/watchers/senderwatcher_test.go
```diff
@@ -0,0 +1,469 @@
...
+	blockEvent := &blockwatch.Event{
+		Type:        blockwatch.Added,
+		BlockHeader: header,
+	}
+
+	go sw.Watch()
+	defer sw.Stop()
+	time.Sleep(2 * time.Millisecond)
+
+	// If map entry is nil , make RPC call and init the map entry
```
yondonfu: Outdated comment?

Same comment applies to the similar comments in the rest of the tests in this file - if the map entry is nil we shouldn't be initializing the map entry in `handleLog()` since we're only initializing map entries when `GetSenderInfo()` is called.

eth/watchers/senderwatcher_test.go
```diff
@@ -0,0 +1,469 @@
...
+	assert.Equal(info, lpEth.SenderInfo)
+
+	// If map entry exists parse event log
+	// ticket amount is 200 gwei
+	faceValue := big.NewInt(200000000000)
+	watcher.sink <- []*blockwatch.Event{blockEvent}
+	time.Sleep(2 * time.Millisecond)
+	info, err = sw.GetSenderInfo(stubSender)
+	assert.Nil(err)
+	t.Log(info)
```
yondonfu: Can remove this log statement

eth/watchers/senderwatcher_test.go
```diff
@@ -0,0 +1,469 @@
...
+	header := defaultMiniHeader()
+	newWinningTicketEvent := newStubWinningTicketLog()
+	header.Logs = append(header.Logs, newWinningTicketEvent)
+
+	blockEvent := &blockwatch.Event{
+		Type:        blockwatch.Added,
+		BlockHeader: header,
+	}
+
+	t.Log(newWinningTicketEvent.Address.Hex(), stubTicketBrokerAddr.Hex(), sw.dec.addr.Hex())
```
yondonfu: Can remove this log statement

eth/watchers/senderwatcher_test.go
```diff
@@ -0,0 +1,469 @@
...
+
+	go sw.Watch()
+	defer sw.Stop()
+	time.Sleep(2 * time.Millisecond)
+
+	// If map entry is nil , make RPC call and init the map entry
+	watcher.sink <- []*blockwatch.Event{blockEvent}
+	time.Sleep(2 * time.Millisecond)
+	info := sw.senders[stubSender]
+	assert.Equal(info, senderInfo)
```
yondonfu: I think there is a problem with this test - `senderInfo` is being mutated since it is pointer returned by the stub client. So, when `SenderWatcher` caches that pointer and then updates it internally, the `senderInfo` will reflect those updates. As a result, any check such as:

`assert.Equal(info, senderInfo)`

will always evaluate to true because `senderInfo` is updated every time the cached pointer in `SenderWatcher` is updated. 

Same comment applies to the similar checks used in the rest of the tests in this file.

eth/watchers/senderwatcher_test.go
```diff
@@ -0,0 +1,469 @@
...
+
+	// If map entry exists parse event log
```
yondonfu, line 159: Let's also add a test case for an event log that references an address we don't care about?

Same comment applies to the similar checks used in the rest of the tests in this file.

eth/watchers/senderwatcher_test.go
```diff
@@ -0,0 +1,469 @@
...
+	blockEvent := &blockwatch.Event{
+		Type:        blockwatch.Added,
+		BlockHeader: header,
+	}
+
+	go sw.Watch()
+	defer sw.Stop()
+	time.Sleep(2 * time.Millisecond)
+
+	// If map entry is nil it is left nil since reserve and deposit are withdrawn
```
yondonfu: I think a more accurate comment is probably "If map entry is nil it is left nil since we do not care about the address in the log"

yondonfu (changes requested):

eth/watchers/senderwatcher_test.go
kyriediculous: fixed in a18ffc1

eth/watchers/senderwatcher_test.go
kyriediculous: fixed in a18ffc1

eth/watchers/senderwatcher_test.go
kyriediculous: fixed in a18ffc1

eth/watchers/senderwatcher_test.go
kyriediculous: fixed in a18ffc1

eth/watchers/senderwatcher_test.go
kyriediculous: fixed in a18ffc1

eth/watchers/senderwatcher_test.go
kyriediculous: I'm not sure what you mean here, could you please elaborate?

 In the cases we parse the event logs we are making checks against the event log data itself. 
When we make an RPC call we check against SenderInfo, yes the cached pointer will be updated with the stubbed out RPC call values, but isn't that the goal here?

eth/watchers/senderwatcher.go
```diff
@@ -0,0 +1,216 @@
...
+		}
+		amount := winningTicketTransfer.Amount
+		sender = winningTicketTransfer.Sender
+
+		if info, ok := sw.senders[sender]; ok && !log.Removed {
+			// See if amount > deposit
+			if info.Deposit.Cmp(amount) < 0 {
+				// Draw from reserve
+				// ReserveFrozen will be handled in it's own event log handler
+				info.Reserve.Sub(info.Reserve, amount)
```
yondonfu: We should first drain the deposit to 0 and then subtract the remaining amount from the reserve for the case where a deposit can only cover a portion of a ticket's face value.

eth/watchers/senderwatcher_test.go
```diff
@@ -0,0 +1,571 @@
...
+	sender := pm.RandAddress()
+
+	sw.setSenderInfo(sender, info)
+
+	getInfo, err := sw.GetSenderInfo(sender)
+	assert.Nil(err)
+	assert.Equal(info, getInfo)
+
+	sw.Clear(sender)
+	time.Sleep(2 * time.Millisecond)
```
yondonfu: Doesn't seem like we need to sleep here

eth/watchers/senderwatcher_test.go
```diff
@@ -0,0 +1,571 @@
...
+
+	// If map entry exists parse event log
+	// ticket amount is 200 gwei
+	// init the map entry
+	info, err = sw.GetSenderInfo(stubSender)
+	require.Nil(err)
+	require.Equal(info, senderInfo)
+	assert.Zero(info.Deposit.Cmp(senderInfo.Deposit))
+	assert.Zero(info.Reserve.Cmp(senderInfo.Reserve))
+	faceValue := big.NewInt(200000000000)
```
yondonfu, line 363: We should also have a test case for when the ticket face value exceeds the deposit so that the redemption drains the deposit to 0 and subtracts the remaining face value from the reserve

eth/watchers/senderwatcher_test.go
```diff
@@ -0,0 +1,571 @@
...
+	// If map entry exists parse event log
+	info, err := sw.GetSenderInfo(stubSender)
+	require.Nil(err)
+	require.Equal(info, senderInfo)
+	watcher.sink <- []*blockwatch.Event{blockEvent}
+	time.Sleep(2 * time.Millisecond)
+	expectedDeposit := senderInfo.Deposit.Add(senderInfo.Deposit, new(big.Int).SetBytes(newDepositEvent.Data))
+	info, err = sw.GetSenderInfo(stubSender)
+	assert.Nil(err)
+	assert.Zero(info.Deposit.Cmp(expectedDeposit))
```
yondonfu: While this test is supposed to check that a sender's deposit increases by the amount indicated by the `DepositFunded` event, in reality it is checking that a sender's deposit increased by the amount indicated by the `DepositFunded` event multiplied by 2.

See below for the test output when we add print statements in the following locations:
- Print `senderInfo.Deposit` right before L156
- Print `senderInfo.Deposit` right before L158
- Print `expectedDeposit` right after L159
- Print `info.Deposit` and `expectedDeposit` right after L161

```
=== RUN   TestFundDepositEvent
senderInfo.Deposit before sending to sink: 10
senderInfo.Deposit after sending to sink: 5000000000000000010
expected deposit: 10000000000000000010
asserted that info.Deposit = expectedDeposit 10000000000000000010 = 10000000000000000010
```

Notice how `senderInfo.Deposit` increases by the `DepositFunded` amount after the block event is sent into the sink. But, then on L158, we mutate `senderInfo.Deposit` when calculating `expectedDeposit` such that it `senderInfo.Deposit` ends up equaling 10 + (the `DepositFunded` amount * 2). The assertion on L161 still passes because any updates to `senderInfo.Deposit` are reflected in the `info` variable returned on L159 since both `info` and `senderInfo` reference the same pointer.

eth/watchers/senderwatcher_test.go
yondonfu: My original comment was a little off-base - let me rephrase. The test was not technically incorrect - it checked that the values referenced by a cached pointer changed whenever the values referenced by the pointer changed. But, I found reading the test a bit confusing.

Consider this code snippet when f06a0ec was the tip of the branch:

```
// We only keep cache for addresses that exist in the map
_, err = sw.GetSenderInfo(stubSender)
assert.Nil(err)

// Code omitted

// If map entry is nil , make RPC call and init the map entry
watcher.sink <- []*blockwatch.Event{blockEvent}
time.Sleep(2 * time.Millisecond)
info := sw.senders[stubSender]
assert.Equal(info, senderInfo)
```

- `sw.GetSenderInfo()` will set the value for `stubSender` in the cache to `senderInfo`
- `watcher.sink <- []*blockwatch.Event{blockEvent}` will send a `DepositFunded` event to be processed by the SenderWatcher. Since, the address in the event is `stubSender`, the SenderWatcher should increase the `Deposit` field for `stubSender` in the cache
- My original reaction when reading the `assert.Equal(info, senderInfo)` was: "why isn't this failing if `info.Deposit` should be increased?". The answer to my question was that `info` and `senderInfo` reference the same pointer so they will *always* be equal in an assertion. `info.Deposit` is being increased here, but IMO the assertion makes it look like we're checking that `info.Deposit` did not increase

The latest commit makes the test a bit clearer, but I just wanted to point this out because I think it can lead to confusing test semantics. See [this comment](https://github.com/livepeer/go-livepeer/pull/1054/files#r320357953) for an example where it might lead to confusion.

eth/watchers/senderwatcher_test.go
kyriediculous: fixed in 6e29515

eth/watchers/senderwatcher_test.go
kyriediculous: fixed in 6e29515

eth/watchers/senderwatcher.go
kyriediculous: fixed in 6e29515

eth/watchers/senderwatcher_test.go
kyriediculous: fixed in 6e29515

eth/watchers/senderwatcher.go
```diff
@@ -0,0 +1,218 @@
...
+		}
+		amount := winningTicketTransfer.Amount
+		sender = winningTicketTransfer.Sender
+
+		if info, ok := sw.senders[sender]; ok && !log.Removed {
+			// See if amount > deposit
+			if info.Deposit.Cmp(amount) < 0 {
+				// Draw from reserve
+				// ReserveFrozen will be handled in it's own event log handler
+				diff := new(big.Int).Abs(new(big.Int).Sub(info.Deposit, amount))
```
yondonfu: I don't think we need to take the absolute value here since we know that `info.Deposit` is less than `amount`. We could just write `diff := new(big.Int).Sub(amount, info.Deposit)`

eth/watchers/senderwatcher_test.go
```diff
@@ -0,0 +1,595 @@
...
+	blockEvent.Type = blockwatch.Removed
+	watcher.sink <- []*blockwatch.Event{blockEvent}
+	time.Sleep(2 * time.Millisecond)
+	info, err := sw.GetSenderInfo(stubSender)
+	assert.Nil(err)
+	assert.Zero(info.Deposit.Cmp(startDeposit))
+
+	// If map entry exists parse event log
+	blockEvent.Type = blockwatch.Added
+	info, err = sw.GetSenderInfo(stubSender)
```
yondonfu: Do we need this `GetSenderInfo()` call for the test setup if we're already caching the sender info for `stubSender` when we test the event removal above? Same comment applies to the rest of the tests in this file where we call `GetSenderInfo()` as a part of test setup after already caching the sender info previously.

eth/watchers/senderwatcher_test.go
```diff
@@ -0,0 +1,595 @@
...
+	_, ok := sw.senders[stubSender]
+	assert.False(ok)
+
+	// if block removed, make RPC call
+	blockEvent.Type = blockwatch.Removed
+	watcher.sink <- []*blockwatch.Event{blockEvent}
+	time.Sleep(2 * time.Millisecond)
+	info, err := sw.GetSenderInfo(stubSender)
+	assert.Nil(err)
+	assert.Zero(info.Deposit.Cmp(startDeposit))
```
yondonfu, line 157: This test would still pass even if we remove L153 since `GetSenderInfo()` will cache the sender info for `stubSender` if the current internal map entry for `stubSender` is nil. Perhaps we could access the internal map directly to make sure we're not populating the cache via a call to `GetSenderInfo()`? Same comment applies to the rest of the tests in this file where we do this.

eth/watchers/senderwatcher_test.go
```diff
@@ -0,0 +1,595 @@
...
+
+	// Test facevalue > deposit
+	senderInfo.Deposit = big.NewInt(100000000000)
+	info, err = sw.GetSenderInfo(stubSender)
+	watcher.sink <- []*blockwatch.Event{blockEvent}
+	time.Sleep(2 * time.Millisecond)
+	info, err = sw.GetSenderInfo(stubSender)
+	assert.Nil(err)
+	assert.Zero(info.Deposit.Int64())
+	diff := new(big.Int).Abs(new(big.Int).Sub(big.NewInt(100000000000), faceValue))
```
yondonfu: We don't need to take the absolute value here and could instead just write: `new(big.Int).Sub(faceValue, big.NewInt(100000000000))`

eth/watchers/senderwatcher_test.go
```diff
@@ -0,0 +1,595 @@
...
+	_, ok = sw.senders[s]
+	assert.False(ok)
+}
+
+func TestUnlockCancelledEvent(t *testing.T) {
+	assert := assert.New(t)
+	require := require.New(t)
+	lpEth := &eth.StubClient{
+		SenderInfo: &pm.SenderInfo{
+			WithdrawBlock: big.NewInt(0),
```
yondonfu: I suggest making this a non-zero value to differentiate between the case where we use an RPC call to refresh the cached `WithdrawBlock` and the case where `WithdrawBlock = 0` because we processed an `UnlockCancelled` event. At the moment, we wouldn't be able to differentiate between the two cases because `WithdrawBlock` is always 0.

kyriediculous: fixed in 4530eff

eth/watchers/senderwatcher_test.go
kyriediculous: fixed in 4530eff

eth/watchers/senderwatcher_test.go
kyriediculous: fixed in 4530eff

eth/watchers/senderwatcher_test.go
kyriediculous: fixed in 4530eff

eth/watchers/senderwatcher.go
kyriediculous: fixed in 4530eff

yondonfu (commented):
Changes look good. Let's rebase to resolve the conflicts with master!

kyriediculous:
Rebased !

yondonfu (approved):
LGTM 🚢

Issue #940, opened 2019-06-26
cmd: Refactor main() through helpers.


kyriediculous:
**Is your feature request related to a problem? Please describe.**

- We should prevent the main function from terminating before deferred functions can be executed. Currently we are mainly using `glog.Fatal` which is equivalent to calling `os.Exit(1)`, this exits the program not allowing any deferred functions to run. 

- In some statements `panic` is used, which executes deferred functions after throwing but it is not idiomatic to panic in such contexts. 
```
panic: The amount of pixels per unit must be greater than 0, provided 0 instead
goroutine 1 [running]:
main.main()
	/Users/nico/go/src/github.com/livepeer/go-livepeer/cmd/livepeer/livepeer.go:327 +0x56f6
```

-  Idiomatically the main function should be kept short and purely used for startup of the program. Startup of services should be initiated through helpers functions.

- We could refactor the cluttered main to simply `glog.Error + return` instead of `glog.Fatal / os.Exit(1)` but this makes testing the returned error codes in `test_args.sh` impossible

Additionally refactoring main makes it easier to maintain the main function and add functionality for the startup of services as needed.  
 
**Describe the solution you'd like**
Extract the startup of various services into helpers. Such as 
`parseFlags`
`startDBservice`
`startEthClient`
`setPriceInfo`
`startMediaServer`
`initS3storage`
`startLivepeerNode`

All error statements should be refactored to match the following pattern 

```
glog.Errorf("...")
return
```
If the helper returns it's deferred functions will execute. We can then safely os.Exit() in the main after encountering an error 

eg.
```
if err := parseFlags(); err != nil {
   os.Exit(1)
}
```

This is the same pattern used by [Cobra](https://github.com/spf13/cobra): 
eg. 
```
// Execute adds all child commands to the root command and sets flags appropriately.
// This is called by main.main(). It only needs to happen once to the rootCmd.
func Execute() {
	if err := rootCmd.Execute(); err != nil {
		fmt.Println(err)
		os.Exit(1)
	}
}
```

**Additional context**
See [this discussion](https://github.com/livepeer/go-livepeer/pull/933#discussion_r297212045) in #933

j0sh:
Generally I agree that the `livepeer.go` could use some cleanup ; see https://github.com/livepeer/go-livepeer/issues/710

> Idiomaticly the main function should be kept short.

[...]

PR #1058, opened 2019-08-22
eth, cmd: return error from roundswatcher.Watch if init fails

kyriediculous:
This PR makes `RoundsWatcher.Watch()` throw an error when the initial RPC requests to populate the cache fail, this will in turn exit the node. 

See https://github.com/livepeer/go-livepeer/pull/1052#discussion_r315754345 

- Added an error as return value from `RoundsWatcher.Watch()`
- Added a test to check for this error, also added a stub error for this case to the `stubEthClient`
- Now exiting the node when `RoundsWatcher `fails to initialize

added and ran unit tests

**Checklist:**

### DIFF
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -284,6 +284,7 @@ func main() {
 		return
 	}
 
+	watcherErr := make(chan error)
 	if *network == "offchain" {
 		glog.Infof("***Livepeer is in off-chain mode***")
 	} else {
@@ -360,7 +361,13 @@ func main() {
 			glog.Errorf("Failed to setup roundswatcher: %v", err)
 			return
 		}
-		go roundsWatcher.Watch()
+
+		roundsWatcherErr := make(chan error, 1)
+		go func() {
+			if err := roundsWatcher.Watch(); err != nil {
+				roundsWatcherErr <- fmt.Errorf("roundswatcher failed to start watching for events: %v", err)
+			}
+		}()
 		defer roundsWatcher.Stop()
 
 		// Initialize unbonding watcher to update the DB with latest state of the node's unbonding locks
@@ -383,10 +390,21 @@ func main() {
 			return
 		}
 
+		blockWatcherErr := make(chan error, 1)
 		go func() {
 			if err := blockWatcher.Watch(blockWatchCtx); err != nil {
-				glog.Errorf("block watcher error: %v", err)
+				blockWatcherErr <- fmt.Errorf("block watcher error: %v", err)
 			}
+		}()
+
+		go func() {
+			var err error
+			select {
+			case err = <-roundsWatcherErr:
+			case err = <-blockWatcherErr:
+			}
+
+			watcherErr <- err
 		}()
 
 		n.Balances = core.NewBalances(cleanupInterval)
@@ -667,6 +685,9 @@ func main() {
 	c := make(chan os.Signal)
 	signal.Notify(c, os.Interrupt)
 	select {
+	case err := <-watcherErr:
+		glog.Error(err)
+		return
 	case err := <-ec:
 		glog.Infof("Error from media server: %v", err)
 		return
--- a/eth/stubclient.go
+++ b/eth/stubclient.go
@@ -127,6 +127,7 @@ type StubClient struct {
 	LatestBlockError             error
 	ProcessHistoricalUnbondError error
 	Orchestrators                []*lpTypes.Transcoder
+	RoundsErr                    error
 }
 
 type stubTranscoder struct {
@@ -140,8 +141,8 @@ func (e *StubClient) Backend() (*ethclient.Client, error)                       
 // Rounds
 
 func (e *StubClient) InitializeRound() (*types.Transaction, error)       { return nil, nil }
-func (e *StubClient) CurrentRound() (*big.Int, error)                    { return big.NewInt(0), nil }
-func (e *StubClient) LastInitializedRound() (*big.Int, error)            { return big.NewInt(0), nil }
+func (e *StubClient) CurrentRound() (*big.Int, error)                    { return big.NewInt(0), e.RoundsErr }
+func (e *StubClient) LastInitializedRound() (*big.Int, error)            { return big.NewInt(0), e.RoundsErr }
 func (e *StubClient) BlockHashForRound(round *big.Int) ([32]byte, error) { return [32]byte{}, nil }
 func (e *StubClient) CurrentRoundInitialized() (bool, error)             { return false, nil }
 func (e *StubClient) CurrentRoundLocked() (bool, error)                  { return false, nil }
--- a/eth/watchers/roundswatcher.go
+++ b/eth/watchers/roundswatcher.go
@@ -64,14 +64,14 @@ func (rw *RoundsWatcher) setLastInitializedRound(round *big.Int, hash [32]byte) 
 }
 
 // Watch the blockwatch subscription for NewRound events
-func (rw *RoundsWatcher) Watch() {
+func (rw *RoundsWatcher) Watch() error {
 	lr, err := rw.lpEth.LastInitializedRound()
 	if err != nil {
-		glog.Errorf("error fetching initial lastInitializedRound value: %v", err)
+		return fmt.Errorf("error fetching initial lastInitializedRound value: %v", err)
 	}
 	bh, err := rw.lpEth.BlockHashForRound(lr)
 	if err != nil {
-		glog.Errorf("error fetching initial lastInitializedBlockHash value: %v", err)
+		return fmt.Errorf("error fetching initial lastInitializedBlockHash value: %v", err)
 	}
 	rw.setLastInitializedRound(lr, bh)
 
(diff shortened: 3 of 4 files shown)


yondonfu (commented):
As is, this PR updates the RoundsWatcher to return an error from `Watch()`, but since we still use the RoundsWatcher as a dependency when initializing `pm.Sender` and `pm.Validator`, ticket creation and validation will result in a nil pointer error. Even if we add nil checks in the `pm` package, the lack of a functioning RoundsWatcher would mean that no tickets could be created.

I actually think we should exit completely if any of the block/event watchers encounter an error when calling `Watch()` since any error invoking `Watch()` prior to entering the for loop likely means that the block/event watcher cannot be relied upon for the node's operation. 

Here is an example that uses channels to [listen for errors](https://github.com/0xProject/0x-mesh/blob/development/core/core.go#L396) from types that start a watching loop in a goroutine and that [exits the program](https://github.com/0xProject/0x-mesh/blob/development/cmd/mesh/main.go#L80) if any of the channels returns an error.

kyriediculous:
Would the changes in 290c14c do the job? 

Would we also want to exit if we encounter a [subscribtion error](https://github.com/livepeer/go-livepeer/blob/f834af95f30f958f15e7a0712c20c462d52bc6de/eth/watchers/roundswatcher.go#L85) ? 

I would perhaps prefer to start all watching services within a helper and do something similar as in the example you provided. But perhaps that's technical debt we can add to https://github.com/livepeer/go-livepeer/issues/940 ?

cmd/livepeer/livepeer.go
```diff
@@ -284,6 +284,7 @@ func main() {
 		return
 	}
 
+	watcherErr := make(chan error)
```
yondonfu, line 287: A small note on channel convention: Since we use an unbuffered channel here it is possible that if multiple watcher goroutines send errors around the same time, then at least one goroutine will be blocked after the consumer of `watcherErr` processes the first error which causes the program to exit. After the program exits there is clearly no consumer of `watcherErr` so we would have exited the program without properly terminating at least one goroutine. This technically isn't that big of a deal since we're exiting the program anyway, but it does violate the common convention of [never starting a goroutine without knowing how it will stop](https://dave.cheney.net/2016/12/22/never-start-a-goroutine-without-knowing-how-it-will-stop) since there isn't a clear mechanism for the blocked watcher goroutines to stop besides relying on the program to exit.

I think we can make a small change (which is largely based on the example [here](https://github.com/0xProject/0x-mesh/blob/development/core/core.go#L407)) to follow the convention of never starting a goroutine without knowing how it will stop. We can create a separate error channel with a buffer size of 1 for each watcher that we start. Then, after we have started all the watchers, we can create `watcherErr` as an error channel with a buffer size 1 and start an additional goroutine that uses a select statement to receive the first error sent by one of the individual watcher error channels - the first error that is received is sent into `watcherErr`. It could look something like this:

```
watcherErr := make(chan error 1)
go func() {
    var err error
    select {
        case err = <-roundsWatcherErr:
        case err = <-blockWatcherErr:
        // Other cases for other watchers
    }

    watcherErr <- err
}()
```

I think the benefit of this approach is that since we're always using buffered channels for each of the goroutines, even if multiple watchers send errors around the same time none of the goroutines will be blocked.

As you mentioned in your PR review comment, eventually all of this could be extracted into a helper if needed.

cmd/livepeer/livepeer.go
```diff
@@ -362,10 +363,14 @@ func main() {
 
 		roundsWatcher, err := watchers.NewRoundsWatcher(addrMap["RoundsManager"], blockWatcher, n.Eth)
 		if err != nil {
-			glog.Errorf("Failed to setup roundswatcher: %v", err)
+			glog.Errorf("failed to setup roundswatcher: %v", err)
```
yondonfu: Looks like the f in failed is capitalized in other error messages - any particular reason to change?

yondonfu (commented):
The updates basically get the job done, but I left a small comment with a suggestion to follow a goroutine convention.

> Would we also want to exit if we encounter a subscribtion error ?

I think its ok not to exit here because I *think* the subscription could recover.

> I would perhaps prefer to start all watching services within a helper and do something similar as in the example you provided. But perhaps that's technical debt we can add to #940 ?

Agreed.

cmd/livepeer/livepeer.go
kyriediculous: I think I understand, but just to make sure: 

Considering the context that we currently have two watchers that potentially return an error when watch is invoked (but not during the for loop itself) , if both error at around the same time we're not correctly terminating whichever of the two goroutines passed an error into the channel last? 

Does that mean that we could resolve this by simply having a seperate channel for the blockwatcher as well, without necessarily wrapping it in a seperate goRoutine that sends the first error encountered into a more general channel? 

Would another approach also work, such as setting a buffer of `2` considering we have two watchers that could return an error into this channel? I know this wouldn't be the most elegant approach but would it work? I think so because the value is sent into the buffer, but the goroutine doesn't care whether it is handled since because there is room in the buffer it doesn't block? 

To me the additional goroutine seemed like a useful approach if we were to use a helper, a little less so if we keep all the logic inside of `main()`, is that an incorrect statement to make? 

That being said I'll introduce the changes as per the suggestion you made. Above questions are more for the purpose of personal  clarification and insight.

cmd/livepeer/livepeer.go
kyriediculous: There's a convention to not start error messages with capital letters, but on second reading of that it does not apply to logging, which is what we're doing here so I'll undo this.

Reference: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

d0e6987

cmd/livepeer/livepeer.go
kyriediculous: I think this should do the trick then ? 0929eee

yondonfu: > Considering the context that we currently have two watchers that potentially return an error when watch is invoked (but not during the for loop itself) , if both error at around the same time we're not correctly terminating whichever of the two goroutines passed an error into the channel last?

Yeah. It is possible that the program exits while one of the goroutines is still blocked because it is trying to send into an unbuffered channel without a receiver. Technically, this is ok since the program is exiting anyway, but IMO it is a code smell because the blocked goroutine never exits cleanly.

> Does that mean that we could resolve this by simply having a seperate channel for the blockwatcher as well, without necessarily wrapping it in a seperate goRoutine that sends the first error encountered into a more general channel?

Yeah that could work, but you would then have to receive on multiple channels to determine if the program should exit. IMO using a separate goroutine that sends the first error encountered from any of the watchers is a bit cleaner because the main thread doesn't need to know which watcher encountered an error (which might be helpful if we end up having some watchers that are only initiated if the node is a broadcaster or an orchestrator) - knowledge about which watcher encountered an error is isolated in a separate select statement and the main thread just needs to know that it received *some* watcher error. The isolation of the select statement over all the individual watcher error channels also makes it easier to extract into a helper in the future if needed.

> Would another approach also work, such as setting a buffer of 2 considering we have two watchers that could return an error into this channel? I know this wouldn't be the most elegant approach but would it work? I think so because the value is sent into the buffer, but the goroutine doesn't care whether it is handled since because there is room in the buffer it doesn't block?

Yeah that would also work. One thing I like about having separate error channels for each watcher and then a select statement over those channels that passes the first error into a dedicated channel is that we don't have to worry about an incorrect buffer size if we add additional watchers over time.

> To me the additional goroutine seemed like a useful approach if we were to use a helper, a little less so if we keep all the logic inside of main(), is that an incorrect statement to make?

An alternative would be to move the select statement from that goroutine [here](https://github.com/livepeer/go-livepeer/pull/1058/files#diff-205b550ef0ac99b368da16cfb883960bR691). I think isolating the select statement can be nice for future extraction and also adding additional watchers if needed though (see my thoughts further up in this comment).

yondonfu (commented):
Changes look good to me - needs a rebase.

Would also be good to do a sanity check that the node actually exits if any of these watchers returns an error when being started (for example by modifying either watcher to return an error when `Watch()` is called).

kyriediculous:
Rebased ! 

Apparently the sanity check results in an error without explicitly returning one because on the current rinkeby deployment there is no notion of `blockHashForRound`.

`E0829 22:07:49.033804   82524 livepeer.go:689] roundswatcher failed to start watching for events: error fetching initial lastInitializedBlockHash value: abi: unmarshalling empty output`

Is that a concern for merging this into master? Internally it shouldn't be an issue because the test-harness uses the `protocol/pm` branch although we should still internally communicate the breaking change, but if someone externally builds from master and uses the node with mainnet or rinkeby, that person might run into trouble.

yondonfu (approved):
LGTM!

yondonfu:
> Is that a concern for merging this into master? Internally it shouldn't be an issue because the test-harness uses the protocol/pm branch although we should still internally communicate the breaking change, but if someone externally builds from master and uses the node with mainnet or rinkeby, that person might run into trouble.

That's ok - the current `master` branch isn't compatible with the mainnet and Rinkeby contract deployments right now anyway.

PR #1061, opened 2019-08-22
server: Verify pixel count after receiving result

### LINKED ISSUE #899: B asynchronously verifies its current credit balance using transcode receipts from O
**CheckTranscodeReceiptsInBackground(sess, seg, transRes)**

Should check that the information in receipts for transcoded results (i.e. pixels processed) is valid. A broadcaster will call this function *asynchronously* after receiving the transcoded results for a segment. This function should be called asynchronously because we want to avoid as much computation as possible on the critical path for streaming and reduce timing pressure. The broadcaster can provisionally use the receipt information returned by an orchestrator in order to update its credit balance in order to send a subsequent payment and the actual verification of the receipt information can occur in the background.

```
for res in transRes {
    // Check that the pixels processed in the receipt matches
    // the number of output pixels in the result
    // In the future, we can take into account input pixels as well
    if res.receipt.pixelsProcessed != outPixels(res) {
        // Send an error to disconnect from the orchestrator
    }
    
    receiptPricePerPixel = res.receipt.priceInfo.pricePerUnit / res.receipt.priceInfo.pixelsPerUnit
    pricePerPixel = sess.priceInfo.pricePerUnit / sess.priceInfo.pixelsPerUnit
    
    // EPISLON is some constant which is used to determine
    // if the price indicated in the receipt is too far off
    // of the latest price that the broadcaster is aware of
    if abs(receiptPricePerPixel - pricePerPixel) > EPISLON {
        // Send an error to disconnect from the orchestrator
    }
}
```

yondonfu:

This PR implements async pixel count verification after a broadcaster receives a transcoded result. 

- Run pixel count verification in a goroutine so that pixel count verification does not impact the stream latency (by slowing down segment insertion into the playlist)
- If the pixel count verification fails, then the broadcaster removes the orchestrator's session from its current session list

Note: An orchestrator that fails the pixel count verification could still be added back to the broadcaster's session list after a refresh - we would need to implement a block list feature in order to exclude the orchestrator from future session lists.

Added unit tests.

Fixes #899

**Checklist:**

### DIFF
--- a/server/broadcast_test.go
+++ b/server/broadcast_test.go
@@ -3,16 +3,22 @@ package server
 import (
 	"errors"
 	"fmt"
+	"net/http"
 	"sync"
 	"testing"
 	"time"
 
+	"github.com/golang/protobuf/proto"
 	"github.com/livepeer/go-livepeer/common"
 	"github.com/livepeer/go-livepeer/core"
 	"github.com/livepeer/go-livepeer/drivers"
 	"github.com/livepeer/go-livepeer/net"
+	"github.com/livepeer/lpms/ffmpeg"
+	"github.com/livepeer/lpms/stream"
+	"github.com/livepeer/m3u8"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func StubBroadcastSession(transcoder string) *BroadcastSession {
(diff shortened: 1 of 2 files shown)


server/broadcast_test.go
```diff
@@ -338,3 +376,112 @@ func TestCleanupSessions(t *testing.T) {
...
+		pl:          &stubPlaylistManager{core.ManifestID("foo")},
+		profile:     &ffmpeg.P144p30fps16x9,
+		sessManager: bsm,
+	}
+
+	err = transcodeSegment(cxn, &stream.HLSSegment{Data: []byte("dummy")}, "dummy")
+	assert.Nil(err)
+
+	// Wait for async pixels verification to finish
+	time.Sleep(500 * time.Millisecond)
```
j0sh, line 426: I've been wondering if there's a better way to test against this type of pattern since sleeps tend to be racy, but can't think of anything that isn't unwieldy (eg, callback that's only used for testing).

j0sh (approved):
LGTM 🚢

server/broadcast_test.go
yondonfu: One thought is to make `cxn.sessManager` a interface instead of a concrete type. Then, during a test the stub for `cxn.sessManager` could implement `removeSession()` as:

```
type stubSessManager struct {
    removed chan *BroadcastSession
}

func (s *stubSessManager) removeSession(sess *BroadcastSession) {
    s.removed <- sess
}
```

In the test, we could do:

```
bsm := &stubSessManager{}

// Other test code

err := transcodeSeg(...)

sess := <-sessManager.removed

// Assert that sess is the BroadcastSession that we expect to be removed i.e. by checking
// sess.OrchestratorInfo.Transcoder
```

This would require a bit of refactoring though since there are parts of the codebase that directly access fields on `cxn.sessManager` right now

j0sh: That sounds good. Seems like there's still the possibility of `removeSession` being called for a separate / unrelated failure, so we can't always be assured that would be testing what we want it to test

PR #1062, opened 2019-08-23
cmd: Set default storage base URI for broadcaster

yondonfu:

This PR sets a default storage base URI for a broadcaster's memory driver when the broadcaster's storage is not configured (i.e. using a S3 or GS driver).

At the moment, if the broadcaster's storage is not configured, it will use a memory driver with an empty base URI. This is a problem because the [absolute URI](https://github.com/livepeer/go-livepeer/blob/master/drivers/local.go#L128) returned by `drivers.SaveData()` for a memory driver will not be resolvable. An example where the lack of a base URI can lead to undesirable behavior is during [pixel count verification](https://github.com/livepeer/go-livepeer/blob/master/server/broadcast.go#L405) - after receiving a transcoded result, the broadcaster will pass the result URI with the arguments to [ffmpeg.Transcode3()](https://github.com/livepeer/go-livepeer/blob/master/server/broadcast.go#L471), but the ffmpeg transcoder will be unable to resolve the URI.

- When `drivers.NodeStorage = nil` and the node is a broadcaster, set the base URI for the broadcaster's memory driver to `"http://" + *httpAddr` where `*httpAddr` is configured by the broadcaster operator

I observed the e2e transcoding workflow with B + O as well as B + O/T in on-chain mode using the devtool.

**Checklist:**

### DIFF
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -581,8 +581,16 @@ func main() {
 	*cliAddr = defaultAddr(*cliAddr, "127.0.0.1", CliPort)
 
 	if drivers.NodeStorage == nil {
-		// base URI will be empty for broadcasters; that's OK
-		drivers.NodeStorage = drivers.NewMemoryDriver(n.GetServiceURI())
+		if n.NodeType == core.OrchestratorNode {
+			drivers.NodeStorage = drivers.NewMemoryDriver(n.GetServiceURI())
+		} else {
+			uri, err := url.ParseRequestURI("http://" + *httpAddr)
+			if err != nil {
+				glog.Errorf("error parsing broadcaster default storage base URI: %v", err)
+				return
+			}
+			drivers.NodeStorage = drivers.NewMemoryDriver(uri)
+		}
 	}
 
 	//Create Livepeer Node


j0sh (changes requested):
> "http://" + *httpAddr where *httpAddr is configured by the broadcaster operator

The httpAddr indicates the binding address. The broadcaster's default of `127.0.0.1` will work here. But `0.0.0.0` is also a valid binding address, and `http://0.0.0.0` usually won't resolve to anything.

There's a few ways to address this I can think of:

* Manually construct the URL for pixel verification if the scheme is missing (it always will be unless external OS is used), and special-case the httpAddr of 0.0.0.0 to 127.0.0.1 . This probably means keeping track of httpAddr somewhere accessible to the pixel verification function or its caller.
* If the scheme is missing,  write the data a tempfile just for pixel verification.

Might also be good to do this for onchain mode only, and skip the pixel check offchain. (Missed this during the original review.)

The old method of using paths also had the benefit of seamless interoperability with CDNs without having to specify the full CDN URL as part of manifest, but it's also possible that's doesn't work anymore; it's been a while since we tested with CDNs.

yondonfu:
What if we check for `0.0.0.0` when setting the base URI for `drivers.NewMemoryDriver()` and use `127.0.0.1` in those cases? This would avoid tracking `*httpAddr` in the pixels verification function or its caller.

> The old method of using paths also had the benefit of seamless interoperability with CDNs without having to specify the full CDN URL as part of manifest, but it's also possible that's doesn't work anymore; it's been a while since we tested with CDNs.

So, this might be a downside of actually setting the base URI for the memory driver. Could we strip the schema when we're inserting the segment into the playlist if the segment is being served from local storage?

> Might also be good to do this for onchain mode only, and skip the pixel check offchain. (Missed this during the original review.)

Good point. I'll make the change.

j0sh:
> What if we check for 0.0.0.0 when setting the base URI for drivers.NewMemoryDriver() and use 127.0.0.1 in those cases?

You mean use 127.0.0.1 as the base URI if 0.0.0.0 is passed in to NewMemoryDriver? I believe that would restrict playback to localhost only - which would be incorrect, since 0.0.0.0 usually indicates the user wants the server to be listening to an external (non-localhost) interface. This would break B's ability to serve segments externally to players, Wowza, etc.

> Could we strip the schema when we're inserting the segment into the playlist if the segment is being served from local storage?

Possibly. The knock-on effect on playlist insertion just to accommodate this special case doesn't really sit well with me though, especially since we'd be getting quite far from the actual problem itself, which is pixel counting being unable to access segments. (And presumably this step would also become necessary if we special-cased a 0.0.0.0 baseURI as 127.0.0.1  within the memory driver, which just further highlights the knock-on effect here.)

That being said, as this PR suggests, there is probably room to improve URL handling in around OS generally and the memory driver specifically, so things compose better under a variety of conditions. But I don't have any good solutions right now without looking at things a bit more closely.

yondonfu:
> You mean use 127.0.0.1 as the base URI if 0.0.0.0 is passed in to NewMemoryDriver? I believe that would restrict playback to localhost only

Ah right...

> That being said, as this PR suggests, there is probably room to improve URL handling in around OS generally and the memory driver specifically, so things compose better under a variety of conditions.

I'll take a look, but I'm leaning towards just writing a temp file for this particular situation right now.

yondonfu:
Closing in favor of #1064

Issue #20, opened 2017-07-06
PPSPP Integration


ericxtang:
Integrate [go-ppspp](https://github.com/livepeer/go-PPSPP).

Tasks:
- Separate Livepeer network p2p protocol from video delivery (ppspp) protocol - this means each will have its own handler
- Implement the VideoNetwork interface

PR #1063, opened 2019-08-26
Connor/3box profiles

S1nus:
This pull request adds an option to `livepeer_cli` which allows you to sign messages using your Ethereum account. It is for use with the 3Box integration.

- a CLI option is added to livepeer_cli (#20) which prompts you for a message to sign
- A Cli Webserver Handler is added which signs arbitrary data and returns the result

- I am unsure of the testing requirements and could use guidance on making it compliant

**Checklist:**

### DIFF
--- a/cmd/livepeer_cli/livepeer_cli.go
+++ b/cmd/livepeer_cli/livepeer_cli.go
@@ -57,6 +57,25 @@ func main() {
 
 		return nil
 	}
+	app.Commands = []cli.Command{
+		{
+			Name: "link-profile",
+			Usage: "link a livepeer / 3Box profile to your transcoder address",
+			Action: func(c *cli.Context) error {
+				w := &wizard{
+					endpoint: fmt.Sprintf("http://%v:%v/status", c.String("host"), c.String("http")),
+					httpPort: "7395",
+					host:     "localhost",
+					in:       bufio.NewReader(os.Stdin),
+				}
+				w.orchestrator = w.isOrchestrator()
+				w.checkNet()
+				w.run()
+				w.signDid()
+				return nil
+			},
+		},
+	}
 	app.Version = core.LivepeerVersion
 	// flag.Parse()
 	app.Run(os.Args)
@@ -107,6 +126,7 @@ func (w *wizard) initializeOptions() []wizardOpt {
 			fmt.Print("For Rinkeby Eth, go to the Rinkeby faucet (https://faucet.rinkeby.io/).")
 			w.read()
 		}, testnet: true},
+		{desc: "Link Transcoder Account to Profile", invoke: w.signDid},
 	}
 	return options
 }
--- a/cmd/livepeer_cli/livepeer_cli.go
+++ b/cmd/livepeer_cli/livepeer_cli.go
@@ -57,25 +57,6 @@ func main() {
 
 		return nil
 	}
-	app.Commands = []cli.Command{
-		{
-			Name: "link-profile",
-			Usage: "link a livepeer / 3Box profile to your transcoder address",
-			Action: func(c *cli.Context) error {
-				w := &wizard{
-					endpoint: fmt.Sprintf("http://%v:%v/status", c.String("host"), c.String("http")),
-					httpPort: "7395",
-					host:     "localhost",
-					in:       bufio.NewReader(os.Stdin),
-				}
-				w.orchestrator = w.isOrchestrator()
-				w.checkNet()
-				w.run()
-				w.signDid()
-				return nil
-			},
-		},
-	}
 	app.Version = core.LivepeerVersion
 	// flag.Parse()
 	app.Run(os.Args)
@@ -126,7 +107,7 @@ func (w *wizard) initializeOptions() []wizardOpt {
 			fmt.Print("For Rinkeby Eth, go to the Rinkeby faucet (https://faucet.rinkeby.io/).")
 			w.read()
 		}, testnet: true},
-		{desc: "Link Transcoder Account to Profile", invoke: w.signDid},
+		{desc: "Sign a message", invoke: w.signMessage},
 	}
 	return options
 }
--- a/cmd/livepeer_cli/wizard.go
+++ b/cmd/livepeer_cli/wizard.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"bufio"
 	"bytes"
 	"errors"
 	"fmt"
@@ -8,6 +9,7 @@ import (
 	"math/big"
 	"net/http"
 	"net/url"
+	"os"
 	"strconv"
 	"strings"
 
@@ -39,6 +41,18 @@ func (w *wizard) readString() string {
 			return text
 		}
 	}
+}
+
+func (w *wizard) readMultilineString() string {
+	fmt.Printf("(press enter followed by CTRL+D when done) > ")
+
+	var buf strings.Builder
+	scanner := bufio.NewScanner(os.Stdin)
+	for scanner.Scan() {
+		buf.WriteString(fmt.Sprintln(scanner.Text()))
+	}
+
+	return strings.TrimSuffix(buf.String(), "\n")
 }
 
 // readStringAndValidate reads a single line from stdin, trims spaces and
--- a/cmd/livepeer_cli/wizard_eth.go
+++ b/cmd/livepeer_cli/wizard_eth.go
@@ -16,3 +16,13 @@ func (w *wizard) setGasPrice() {
 
 	httpPostWithParams(fmt.Sprintf("http://%v:%v/setGasPrice", w.host, w.httpPort), val)
 }
+
+func (w *wizard) signMessage() {
+	fmt.Printf("Enter or paste the message to sign: \n")
+	msg := w.readMultilineString()
+	val := url.Values{
+		"message": {msg},
+	}
+	sig := httpPostWithParams(fmt.Sprintf("http://%v:%v/signMessage", w.host, w.httpPort), val)
+	fmt.Println(fmt.Sprintf("\n\nSignature:\n0x%x", sig))
+}
--- a/eth/stubclient.go
+++ b/eth/stubclient.go
@@ -326,7 +326,7 @@ func (c *StubClient) CheckTx(tx *types.Transaction) error {
 func (c *StubClient) ReplaceTransaction(tx *types.Transaction, method string, gasPrice *big.Int) (*types.Transaction, error) {
 	return nil, nil
 }
-func (c *StubClient) Sign(msg []byte) ([]byte, error)   { return msg, nil }
+func (c *StubClient) Sign(msg []byte) ([]byte, error)   { return msg, c.Err }
 func (c *StubClient) GetGasInfo() (uint64, *big.Int)    { return 0, nil }
 func (c *StubClient) SetGasInfo(uint64, *big.Int) error { return nil }
 
--- a/server/handlers.go
+++ b/server/handlers.go
@@ -290,3 +290,22 @@ func ticketBrokerParamsHandler(client eth.LivepeerEthClient) http.Handler {
 		w.Write(data)
 	})
 }
+
+func signMessageHandler(client eth.LivepeerEthClient) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if client == nil {
+			respondWith500(w, "missing ETH client")
+			return
+		}
+
+		message := r.FormValue("message")
+		signed, err := client.Sign([]byte(message))
+		if err != nil {
+			respondWith500(w, fmt.Sprintf("could not sign message - err=%v", err))
+			return
+		}
+
+		w.WriteHeader(http.StatusOK)
+		w.Write(signed)
+	})
+}
--- a/server/handlers_test.go
+++ b/server/handlers_test.go
@@ -3,6 +3,7 @@ package server
 import (
 	"encoding/json"
 	"errors"
+	"fmt"
 	"io"
 	"io/ioutil"
 	"math/big"
@@ -645,6 +646,43 @@ func TestTicketBrokerParamsHandler_Success(t *testing.T) {
 	assert := assert.New(t)
 	assert.Equal(http.StatusOK, resp.StatusCode)
 	assert.Equal(unlockPeriod, params.UnlockPeriod)
+}
+
+func TestSignMessageHandler(t *testing.T) {
+	assert := assert.New(t)
+
+	// Test missing client
+	handler := signMessageHandler(nil)
+	resp := httpPostFormResp(handler, nil)
+	defer resp.Body.Close()
+	body, _ := ioutil.ReadAll(resp.Body)
+
+	assert.Equal(http.StatusInternalServerError, resp.StatusCode)
+	assert.Equal("missing ETH client", strings.TrimSpace(string(body)))
+
+	// Test signing error
+	err := errors.New("signing error")
+	client := &eth.StubClient{Err: err}
+	handler = signMessageHandler(client)
+	resp = httpPostFormResp(handler, nil)
+	defer resp.Body.Close()
+	body, _ = ioutil.ReadAll(resp.Body)
+
+	assert.Equal(http.StatusInternalServerError, resp.StatusCode)
+	assert.Equal(fmt.Sprintf("could not sign message - err=%v", err), strings.TrimSpace(string(body)))
+
+	// Test signing success
+	client.Err = nil
+	msg := "foo"
+	form := url.Values{
+		"message": {msg},
+	}
+	handler = signMessageHandler(client)
+	resp = httpPostFormResp(handler, strings.NewReader(form.Encode()))
+	defer resp.Body.Close()
+	body, _ = ioutil.ReadAll(resp.Body)
+	assert.Equal(http.StatusOK, resp.StatusCode)
+	assert.Equal([]byte(msg), body)
 }
 
 func httpPostFormResp(handler http.Handler, body io.Reader) *http.Response {
--- a/server/webserver.go
+++ b/server/webserver.go
@@ -76,6 +76,15 @@ func (s *LivepeerServer) cliWebServerHandlers(bindAddr string) *http.ServeMux {
 	mux := http.DefaultServeMux
 	http.DefaultServeMux = http.NewServeMux()
 
+	mux.HandleFunc("/signDid", func(w http.ResponseWriter, r *http.Request) {
+		signed, err := s.LivepeerNode.Eth.Sign([]byte("coolio"))
+		fmt.Println(string(signed))
+		fmt.Println("error:")
+		fmt.Println(err)
+		fmt.Println("cliWebserverHandler hit!")
+		w.Write(signed)
+	})
+
 	//Set the broadcast config for creating onchain jobs.
 	mux.HandleFunc("/setBroadcastConfig", func(w http.ResponseWriter, r *http.Request) {
 		if err := r.ParseForm(); err != nil {
--- a/server/webserver.go
+++ b/server/webserver.go
@@ -76,14 +76,7 @@ func (s *LivepeerServer) cliWebServerHandlers(bindAddr string) *http.ServeMux {
 	mux := http.DefaultServeMux
 	http.DefaultServeMux = http.NewServeMux()
 
-	mux.HandleFunc("/signDid", func(w http.ResponseWriter, r *http.Request) {
-		signed, err := s.LivepeerNode.Eth.Sign([]byte("coolio"))
-		fmt.Println(string(signed))
-		fmt.Println("error:")
-		fmt.Println(err)
-		fmt.Println("cliWebserverHandler hit!")
-		w.Write(signed)
-	})
+	mux.Handle("/signMessage", mustHaveFormParams(signMessageHandler(s.LivepeerNode.Eth), "message"))
 
 	//Set the broadcast config for creating onchain jobs.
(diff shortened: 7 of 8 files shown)


cmd/livepeer_cli/wizard_eth.go
```diff
@@ -16,3 +16,17 @@ func (w *wizard) setGasPrice() {
...
+	fmt.Printf("Enter the message to sign:\n")
+	text, err := w.in.ReadString('\n')
+	if err != nil {
+		fmt.Printf("Failed to read user input", "err", err)
+	}
+	//fmt.Printf(text)
+	val := url.Values{
+		"message": {fmt.Sprintf("%v", text)},
+	}
+	r := httpPostWithParams(fmt.Sprintf("http://%v:%v/signDid", w.host, w.httpPort), val)
```
j0sh: Is `signDid` the best name for this endpoint and function? What about something like `signMessage` ?

cmd/livepeer_cli/livepeer_cli.go
```diff
@@ -107,6 +107,7 @@ func (w *wizard) initializeOptions() []wizardOpt {
 			fmt.Print("For Rinkeby Eth, go to the Rinkeby faucet (https://faucet.rinkeby.io/).")
 			w.read()
 		}, testnet: true},
+		{desc: "Link Transcoder Account to Profile", invoke: w.signDid},
```
j0sh: Would "Sign a message" be a better / more general description?

server/webserver.go
```diff
@@ -69,6 +69,22 @@ func (s *LivepeerServer) StartCliWebserver(bindAddr string) {
 func (s *LivepeerServer) cliWebServerHandlers(bindAddr string) *http.ServeMux {
 	mux := http.NewServeMux()
 
+	mux.HandleFunc("/signDid", func(w http.ResponseWriter, r *http.Request) {
+		//signed, err := s.LivepeerNode.Eth.Sign([]byte("coolio"))
```
j0sh: Needs to be cleaned up, or even better, moved into a unit test. See https://github.com/livepeer/go-livepeer/blob/master/cmd/livepeer_cli/wizard_ticketbroker_test.go for an example.

cmd/livepeer_cli/wizard_eth.go
```diff
@@ -16,3 +16,17 @@ func (w *wizard) setGasPrice() {
...
 	httpPostWithParams(fmt.Sprintf("http://%v:%v/setGasPrice", w.host, w.httpPort), val)
 }
+
+func (w *wizard) signDid() {
+	fmt.Printf("Enter the message to sign:\n")
+	text, err := w.in.ReadString('\n')
+	if err != nil {
+		fmt.Printf("Failed to read user input", "err", err)
+	}
+	//fmt.Printf(text)
```
j0sh: stray?

j0sh (dismissed):
Thanks for this! Left some notes.

cmd/livepeer_cli/livepeer_cli.go
S1nus: Yes, it would.

cmd/livepeer_cli/wizard_eth.go
S1nus: changed.

server/webserver.go
S1nus: Is checking `err == nil` a good enough way of unit testing this?

j0sh: Something like this should work

```
// create a LivepeersServer with a stubbed Eth client. See 
// https://github.com/livepeer/go-livepeer/blob/master/server/mediaserver_test.go#L38-L47
// https://github.com/livepeer/go-livepeer/blob/master/eth/stubclient.go

cli := server.cliWebServerHandlers("")
ts := httptest.NewServer(cli)
client := ts.Client()
resp, err := client.PostForm(ts.URL+"/signMessage", url.Values{ "message": "abcdef" })
body, _ := ioutil.ReadAll(resp.Body)
// now verify that the body was signed .
// The stubbed Eth client doesn't look like it does a good job of actually signing anything;
// maybe modify it to append something to the input like []byte(string(msg)+"signed") .
// Or ask @yondonfu about invoking sig verification from PM ; there's probably a stubbed version for testing.
```

(sample was not tested at all and missing a bunch of things but hopefully the gist is clear)

There may be goclient helpers to accommodate this already ; @yondonfu has worked more with HTTP test cases so might know better.

yondonfu: The way that I have tested HTTP handlers in the past is to define the HTTP handler itself as a function that is passed in as an argument to `mux.Handle()`. See [this example](https://github.com/livepeer/go-livepeer/blob/master/server/webserver.go#L1079). Then I use the [httptest](https://golang.org/pkg/net/http/httptest/) package to simulate HTTP requests. See [this example](https://github.com/livepeer/go-livepeer/blob/master/server/handlers_test.go#L175).

server/webserver.go
```diff
@@ -76,6 +76,18 @@ func (s *LivepeerServer) cliWebServerHandlers(bindAddr string) *http.ServeMux {
 	mux := http.DefaultServeMux
 	http.DefaultServeMux = http.NewServeMux()
 
+	mux.HandleFunc("/signMessage", func(w http.ResponseWriter, r *http.Request) {
+		if err := r.ParseForm(); err != nil {
+			glog.Errorf("Parse Form Error: %v", err)
```
yondonfu: Should return if we hit an error here.

[mustHaveFormParams](https://github.com/livepeer/go-livepeer/blob/a9e6dd764ee179db69309618917d176b5cba26fc/server/handlers.go#L28) might be helpful here - see this [example usage](https://github.com/livepeer/go-livepeer/blob/master/server/webserver.go#L1100)).

server/webserver.go
```diff
@@ -76,6 +76,18 @@ func (s *LivepeerServer) cliWebServerHandlers(bindAddr string) *http.ServeMux {
...
+		}
+		message := r.FormValue("message")
+		signed, err := s.LivepeerNode.Eth.Sign([]byte(message))
+		if err != nil {
+			glog.Errorf("error: %v", err)
```
yondonfu: Let's return an HTTP error here - [example](https://github.com/livepeer/go-livepeer/blob/a9e6dd764ee179db69309618917d176b5cba26fc/server/handlers.go#L110)

server/webserver.go
```diff
@@ -76,6 +76,18 @@ func (s *LivepeerServer) cliWebServerHandlers(bindAddr string) *http.ServeMux {
...
+		}
+		w.Write([]byte(signed))
```
yondonfu: Let's write back an OK status here - [example](https://github.com/livepeer/go-livepeer/blob/a9e6dd764ee179db69309618917d176b5cba26fc/server/handlers.go#L150)

cmd/livepeer_cli/wizard_eth.go
```diff
@@ -16,3 +19,25 @@ func (w *wizard) setGasPrice() {
...
+			mystr = append(mystr, text)
+		}
+	}
+	message := strings.Join(mystr, "")
+	message = strings.TrimSpace(message)
+	val := url.Values{
+		"message": {fmt.Sprintf("%v", message)},
+	}
+	signedMessage := httpPostWithParams(fmt.Sprintf("http://%v:%v/signMessage", w.host, w.httpPort), val)
+	clipboard.WriteAll(fmt.Sprintf("0x%x", signedMessage))
```
yondonfu: Personally not a huge fan of automatic clipboard hijacking (although I do see why it might be convenient). I'd be ok with having the user copy the signed message themselves.

cmd/livepeer_cli/wizard_eth.go
```diff
@@ -16,3 +19,25 @@ func (w *wizard) setGasPrice() {
 
 	httpPostWithParams(fmt.Sprintf("http://%v:%v/setGasPrice", w.host, w.httpPort), val)
 }
+
+func (w *wizard) signMessage() {
+	fmt.Printf("Enter or paste the message to sign, then press ctrl + d:\n")
+	mystr := []string{""}
+	text, err := w.in.ReadString('\n')
```
yondonfu: Minor: If we declare `text` and `err` before the loop, we could avoid the extra `ReadString()` call outside of the loop. Something like:

```
var text string
var err error
for err != io.EOF {
    if text, err = w.in.ReadString('\n'); err != io.EOF {
        ...
    }
}
```

yondonfu (changes requested):

kyriediculous:
made the necessary fixups and reworked the commit history.

Can I run `go mod tidy` to clean up old deps (clipboard stuff) ?

server/handlers.go
```diff
@@ -290,3 +290,23 @@ func ticketBrokerParamsHandler(client eth.LivepeerEthClient) http.Handler {
...
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if client == nil {
+			respondWith500(w, "missing ETH client")
+			return
+		}
+
+		message := r.FormValue("message")
+		signed, err := client.Sign([]byte(message))
+		if err != nil {
+			glog.Errorf("error: %v", err)
```
yondonfu: `respondWith500` will already log the error so we don't need to log the error here.

server/handlers.go
```diff
@@ -290,3 +290,23 @@ func ticketBrokerParamsHandler(client eth.LivepeerEthClient) http.Handler {
...
+			respondWith500(w, fmt.Sprintf("could not sign message - err=%v", err))
+			return
+		}
+
+		w.WriteHeader(http.StatusOK)
+		w.Write([]byte(signed))
```
yondonfu: `signed` is already a `[]byte` so we don't need the type cast here.

server/handlers_test.go
```diff
@@ -647,6 +649,43 @@ func TestTicketBrokerParamsHandler_Success(t *testing.T) {
 	assert.Equal(unlockPeriod, params.UnlockPeriod)
 }
 
+func TestSignMessageHandler(t *testing.T) {
+	assert := assert.New(t)
+
+	// Test missing client
+	handler := signMessageHandler(nil)
+	resp := httpGetResp(handler)
```
yondonfu: Doesn't make a huge difference right now since the server doesn't actually restrict the allowed methods (i.e. only allow POST), but might be nice for consistency to use `httpPostFormResp(handler, nil)` here since we use it later on in the test to actually submit a request with form params (this is also what is done in other tests).

eth/stubclient.go
```diff
@@ -326,7 +327,7 @@ func (c *StubClient) CheckTx(tx *types.Transaction) error {
 func (c *StubClient) ReplaceTransaction(tx *types.Transaction, method string, gasPrice *big.Int) (*types.Transaction, error) {
 	return nil, nil
 }
-func (c *StubClient) Sign(msg []byte) ([]byte, error)   { return msg, nil }
+func (c *StubClient) Sign(msg []byte) ([]byte, error)   { return crypto.Keccak256(msg), c.Err }
```
yondonfu: Why return a hash instead of just returning `msg`?

cmd/livepeer_cli/wizard_eth.go
```diff
@@ -16,3 +16,13 @@ func (w *wizard) setGasPrice() {
...
 }
+
+func (w *wizard) signMessage() {
+	fmt.Printf("Enter or paste the message to sign, then press ctrl + d:\n")
+	msg := w.readString()
+	val := url.Values{
+		"message": {msg},
+	}
+	sig := httpPostWithParams(fmt.Sprintf("http://%v:%v/signMessage", w.host, w.httpPort), val)
+	fmt.Println(fmt.Sprintf("\n\nSignature:\n%s", sig))
```
yondonfu: We should print the hex representation of the sig - could use "0x%x" here.

cmd/livepeer_cli/wizard_eth.go
```diff
@@ -16,3 +16,13 @@ func (w *wizard) setGasPrice() {
 
 	httpPostWithParams(fmt.Sprintf("http://%v:%v/setGasPrice", w.host, w.httpPort), val)
 }
+
+func (w *wizard) signMessage() {
+	fmt.Printf("Enter or paste the message to sign, then press ctrl + d:\n")
+	msg := w.readString()
```
yondonfu: This won't be able to handle a multi-line string (which the previous impl in this PR did). An example of a multi-line string is the 3Box auth message:

```
Create a new 3Box profile

-
Your unique profile ID is did:muport:<DID>
Timestamp: <TIMESTAMP>
```

Would be good to double check that the CLI can accept a message with the above format that is copy + pasted from the consent message pop up in the explorer when setting up a 3Box profile:

Screen Shot 2020-01-29 at 6 06 22 PM

yondonfu (changes requested):

yondonfu:
> Can I run go mod tidy to clean up old deps (clipboard stuff) ?

👍for removing the no longer needed clipboard dep.

cmd/livepeer_cli/wizard_eth.go
kyriediculous: fixed in f811dbd

cmd/livepeer_cli/wizard_eth.go
kyriediculous: fixed in f811dbd

eth/stubclient.go
kyriediculous: fixed in b63fb25

server/handlers_test.go
kyriediculous: fixed in b63fb25

server/handlers.go
kyriediculous: fixed in b63fb25

server/handlers.go
kyriediculous: fixed in b63fb25

cmd/livepeer_cli/wizard.go
```diff
@@ -41,6 +42,24 @@ func (w *wizard) readString() string {
 	}
 }
 
+func (w *wizard) readMultilineString() string {
+	ss := make([]string, 1)
```
yondonfu: Any reason for creating a slice of length 1 here instead of just creating an empty slice without specifying a length i.e. `[]string{}`?

Side note: The `strings` package also has a [Builder](https://golang.org/pkg/strings/#Builder) type that is also handy for situations such as this one where we need to iteratively build a string.

cmd/livepeer_cli/wizard.go
```diff
@@ -41,6 +42,23 @@ func (w *wizard) readString() string {
...
+		if err != nil {
+			if err != io.EOF {
+				log.Crit("Failed to read user input", "err", err)
+				continue
+			}
+			break
+		}
+		ss = append(ss, text)
+	}
+	return strings.TrimSpace(strings.Join(ss, ""))
```
yondonfu: Some messages end in whitespace (either horizontal or vertical i.e. `\n`) - we shouldn't mutate the provided message.

cmd/livepeer_cli/wizard.go
```diff
@@ -41,6 +42,23 @@ func (w *wizard) readString() string {
...
+	ss := make([]string, 1)
+	fmt.Printf("(press enter, then ctrl+d when done) > ")
+	for {
+		text, err := w.in.ReadString('\n')
+		if err != nil {
+			if err != io.EOF {
+				log.Crit("Failed to read user input", "err", err)
+				continue
+			}
+			break
```
yondonfu: If the user presses ctrl+d, we should still read the current line even if there is no newline character at the end. Otherwise, we will always omit the last line of a multi-line message if the last line does not contain a newline.

For example, consider the 3Box auth message:

"Create a new 3Box profile\n\n-\nYour unique profile ID is did:muport:QmPXNzBdKMrHcnEaCwCAGcxGCmyiNqoHYgSn34z5uJFvwh \nTimestamp: 1580338635"

There is no newline at the end of the message. So, when the message is copy + pasted, if the user presses ctrl+d, the parsed message will just be:

"Create a new 3Box profile\n\n-\nYour unique profile ID is did:muport:QmPXNzBdKMrHcnEaCwCAGcxGCmyiNqoHYgSn34z5uJFvwh \n"

And if the user presses enter and *then* ctrl+d, the parsed message will be:

""Create a new 3Box profile\n\n-\nYour unique profile ID is did:muport:QmPXNzBdKMrHcnEaCwCAGcxGCmyiNqoHYgSn34z5uJFvwh \nTimestamp: 1580338635\n"

which is also incorrect because there is an extra newline at the end of the message.

cmd/livepeer_cli/wizard.go
```diff
@@ -41,6 +42,23 @@ func (w *wizard) readString() string {
 	}
 }
 
+func (w *wizard) readMultilineString() string {
+	ss := make([]string, 1)
+	fmt.Printf("(press enter, then ctrl+d when done) > ")
```
yondonfu: We probably shouldn't tell the user to press enter and then ctrl+d if this would add an additional newline to the message that was not originally there.

yondonfu (changes requested):
Left a few comments on message parsing. 

Could we also double check that the steps in https://forum.livepeer.org/t/how-to-securely-link-your-3box-profile-to-your-orchestrator-account-using-mycrypto/986 work when signing via the CLI (instead of via MyCrypto)?

cmd/livepeer_cli/wizard.go
kyriediculous: 24be2a7 should do the trick without having to join strings together or anything, however in my case I would have to press ctrl+d twice to send an EOF signal without having to create a new line. 

output using `%q` formatter 
```
Enter or paste the message to sign:
(press ctrl+d when done) > Create a new 3Box profile

-
Your unique profile ID is did:muport:<DID>
Timestamp: <TIMESTAMP>^D
"Create a new 3Box profile\n\n-\nYour unique profile ID is did:muport:<DID>\nTimestamp: <TIMESTAMP>"


Signature:
0x4571136f7d2a0ab82fb6da4d6206ff3a8027e3328593b805b0c640e2f35bdbab58121bb2ab0589221c02f665058f396a4bddc5d551629bd3d2c557bf419afc651c
```

cmd/livepeer_cli/wizard.go
```diff
@@ -41,6 +42,20 @@ func (w *wizard) readString() string {
 	}
 }
 
+func (w *wizard) readMultilineString() string {
+	fmt.Printf("(press ctrl+d when done) > ")
+	for {
+		text, err := w.in.ReadString(';')
```
yondonfu: This won't be able to handle input that contains ";". For example, if the input is "foobar;hello\nbaz", the parsed message will be "hello\nbaz".

What about something like:

```
var buf strings.Builder
scanner := bufio.NewScanner(os.Stdin)
for scanner.Scan() {
	buf.WriteString(fmt.Sprintln(scanner.Text()))
}

return strings.TrimSuffix(buf.String(), "\n")
```

Since the scanner strips the new line at the end of each token, `fmt.Sprintln` adds the newline back and at the end we strip the last newline to make sure we don't add an extra one when there shouldn't be.

If we use this approach then we can tell the user to press enter and then ctrl+d instead of ctrl+d twice (which might be confusing).

yondonfu (changes requested):

cmd/livepeer_cli/wizard_eth.go
kyriediculous: Double checked that the CLI is able to sign the message format provided by 3Box.

Can't really find where I could enter the message and its signature manually to access my profile as a sanity check though. cc @adamsoffer

cmd/livepeer_cli/wizard.go
kyriediculous: fixed in eed6d0b0

cmd/livepeer_cli/wizard_eth.go
adamsoffer: @kyriediculous after you set up your profile, you should see an option to an "add account" underneath the "edit profile" modal. Once you click that you'll see two input fields at the bottom of the form: one to enter the signature and one to ender the address you signed the message with to verify it was signed correctly. If it's valid, you'll see a green message underneath the input letting you know it passed, or red if it failed.

image

kyriediculous: Thanks @adamsoffer 

Looks like it was succesful

image

yondonfu: Steps 6-12 in https://forum.livepeer.org/t/how-to-securely-link-your-3box-profile-to-your-orchestrator-account-using-mycrypto/986 describe how to edit an existing profile link an external account by providing a signature and the account that produced the signature.

yondonfu (commented):
Changes look good! Let's rebase.

kyriediculous:
rebased !

yondonfu:
Can we also squash 8f0b3bd into d099739 instead of having an additional commit that removes a dependency that was introduced in d099739? This would completely clean up the dependency from the history.

kyriediculous:
Sure ! Done.

go.mod
```diff
@@ -7,6 +7,7 @@ require (
 	github.com/allegro/bigcache v1.2.1 // indirect
 	github.com/apilayer/freegeoip v3.5.0+incompatible // indirect
 	github.com/aristanetworks/goarista v0.0.0-20190909155222-05df9ecbb0dc // indirect
+	github.com/atotto/clipboard v0.1.2
```
yondonfu: Looks like this dependency was re-introduced

kyriediculous: hmm ah could it be that since I had to re-order the commits and squashed it into the first commit that it was re-introduced in the second commit?

kyriediculous: should be removed now

yondonfu (approved):
LGTM pending green CI!

S1nus:
😀

PR #1064, opened 2019-08-27
Temp file for pixels verification

yondonfu:

This PR writes a temp file for pixels verification if the transcoded result url is a relative URI and the broadcaster is using local memory storage.

- 4da8ac9 updates `(*MemorySession).GetData()` to handle input names with a prefix (i.e. if the name is an absolute URI or prepended with "/stream/"). This change makes it easier to use the return value of `(*MemorySession).SaveData()` to then fetch the saved data using `(*MemorySession).GetData()`
- 138494b updates pixels verification to write a temp file if the transcoded result is a relative URI and the broadcaster is using local memory storage. This commit also updates pixels verification to only run when the node is in on-chain mode

Added unit tests.

**Checklist:**

### DIFF
--- a/server/broadcast_test.go
+++ b/server/broadcast_test.go
@@ -3,6 +3,7 @@ package server
 import (
 	"errors"
 	"fmt"
+	"io/ioutil"
 	"net/http"
 	"sync"
 	"testing"
@@ -13,6 +14,7 @@ import (
 	"github.com/livepeer/go-livepeer/core"
 	"github.com/livepeer/go-livepeer/drivers"
 	"github.com/livepeer/go-livepeer/net"
+	"github.com/livepeer/go-livepeer/pm"
 	"github.com/livepeer/lpms/ffmpeg"
 	"github.com/livepeer/lpms/stream"
 	"github.com/livepeer/m3u8"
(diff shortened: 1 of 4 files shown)


server/broadcast_test.go
```diff
@@ -471,17 +489,37 @@ func TestPixels(t *testing.T) {
...
-	err := verifyPixels("foo", 50)
+	// Create memory session and save a test file
+	bos := drivers.NewMemoryDriver(nil).NewSession("foo")
+	data, err := ioutil.ReadFile("test.flv")
+	require.Nil(err)
+	fname, err := bos.SaveData("test.ts", data)
+	require.Nil(err)
+
+	// Test error for relative URI and no memory storage
+	// Will not be able to directly read file for a relative URI
```
j0sh: Nit: technically a relative URI would work if it points to a file that lives on disk. While this might not ever be the case in non-test code,  that is essentially what's happening here when "test.flv" is passed in without a corresponding OS.

The comment can probably be updated to reflect something like, "Test error for nonexistent file". (But if we do test for that, then should we also add a test for a nonexistent file in the OS itself? But that's veering into testing OS rather than the pixel verification function.)

j0sh:
Generally looks good, modulo the small nit around the comment in test code. Can you sanity check that everything still works as expected with:

* local OS, standalone Os and remote Ts (since we tweaked local OS - just to be sure that flow isn't affected)
* An external OS such as S3 on B (checks that the verification logic still handles non-local OS)

server/broadcast_test.go
yondonfu: Fixed in 74a3414

yondonfu:
> Can you sanity check that everything still works as expected with:
> - local OS, standalone Os and remote Ts (since we tweaked local OS - just to be sure that flow isn't affected)
> - An external OS such as S3 on B (checks that the verification logic still handles non-local OS)

Verified the first scenario. Discovered a problem with the second scenario - when running in on-chain mode, the signature verification fails because the results are directly uploaded to B's OS so B does not have access to any of the result hashes needed for signature verification. The signature verification failure in this scenario is fixed in b5e1d2d and I added a TODO to consider downloading the data in that scenario to properly generate the result hashes for signature verification instead of skipping the check altogether.

I also verified that the transcoding workflow works properly when nodes are in on-chain mode for these scenarios:

- O uses an external OS
- Both B & O use an external OS

Tested using a S3 bucket as the external OS.

j0sh (approved):
Thanks for checking these cases! LGTM after a rebase.

> added a TODO to consider downloading the data in that scenario to properly generate the result hashes for signature verification instead of skipping the check altogether.

Yeah - the original motivation for this was to avoid the additional download. But we're doing the download now anyway.

Since the final construction of the sig check might change as the verification protocol settles, it's fine to leave as a TODO. However, if/when were to implement the sig check, the signature of the `verifyPixels` function should be simplified to:

```golang
verifyPixels(data []byte, reportedPixels int64) error
```

and then `data` can be written to a tempfile unconditionally. If we don't do this, we'd essentially be downloading the file twice with external OS: once for sig checking, and once for pixel verification.

PR #1066, opened 2019-08-28
[READY] gpu enabled image

ya7ya:
allows the test-harness to run gpu enabled livepeer nodes

- 
- 
- 

**Checklist:**


docker/Dockerfile.gpu
```diff
@@ -0,0 +1,149 @@
...
+    apt-get install software-properties-common -y && \
+    add-apt-repository ppa:longsleep/golang-backports && \
+    apt-get update && apt-get dist-upgrade -y && \
+    apt-get install -y build-essential pkg-config autoconf gnutls-dev \
+            git curl golang-go && \
+   ## curl -O http://developer.download.nvidia.com/compute/cuda/repos/ubuntu1804/x86_64/cuda-repo-ubuntu1804_10.0.130-1_amd64.deb && \
+   # curl -O http://developer.download.nvidia.com/compute/cuda/repos/ubuntu1804/x86_64/cuda-repo-ubuntu1804_10.1.243-1_amd64.deb && \
+   # dpkg --force-confdef --force-confnew -i ./cuda-repo-ubuntu1804_10.1.243-1_amd64.deb && \
+    ## apt-key adv --fetch-keys http://developer.download.nvidia.com/compute/cuda/repos/ubuntu1604/x86_64/7fa2af80.pub && \
+   # apt-key adv --fetch-keys http://developer.download.nvidia.com/compute/cuda/repos/ubuntu1804/x86_64/7fa2af80.pub && \
```
mkrufky: does it serve any purpose to keep these commented lines here?

docker/Dockerfile.gpu
```diff
@@ -0,0 +1,149 @@
...
+RUN export PATH=$HOME/compiled/bin:$PATH && \
+    export PATH=/usr/local/cuda/bin:$PATH && \
+    export PKG_CONFIG_PATH=$HOME/compiled/lib/pkgconfig && \
+    cd "$HOME/ffmpeg" && \
+    make && make install
+
+RUN export PATH=$HOME/compiled/bin:$PATH && \
+    export PATH=/usr/local/cuda/bin:$PATH && \
+    export PKG_CONFIG_PATH=$HOME/compiled/lib/pkgconfig && \
+    go get github.com/livepeer/go-livepeer/cmd/livepeer && \
```
mkrufky: does this still work?   Now that we've moved to go.mod, it might make more sense to `git clone` the repo rather than working within the GOPATH.   For instance:
```
RUN export PATH=$HOME/compiled/bin:$PATH && \
￼	    export PATH=/usr/local/cuda/bin:$PATH && \
￼	    export PKG_CONFIG_PATH=$HOME/compiled/lib/pkgconfig && \
￼	    git clone https://github.com/livepeer/go-livepeer "$HOME/go-livepeer" && \
￼	    cd "$HOME/go-livepeer" && \
￼	    go mod download && \
￼	    go build ./cmd/livepeer/livepeer.go && \
￼	    go build ./cmd/devtool/devtool.go

Comment (darkdarkdragon) [reply docker/Dockerfile.gpu]: it doesn't work.

File: docker/Dockerfile.gpu [hunk 3]
```diff
@@ -0,0 +1,149 @@
...
+RUN apt-get update && apt-get dist-upgrade -y && \
+    apt-get install software-properties-common -y && \
+    add-apt-repository ppa:longsleep/golang-backports && \
+    apt-get update && apt-get dist-upgrade -y && \
+    apt-get install -y build-essential pkg-config autoconf gnutls-dev \
+            git curl golang-go && \
+    ##apt-get install -y libnvidia-encode-430 && \
+    ##curl -O http://developer.download.nvidia.com/compute/cuda/repos/ubuntu1804/x86_64/cuda-repo-ubuntu1804_10.0.130-1_amd64.deb && \
+    ##dpkg --force-confdef --force-confnew -i ./cuda-repo-ubuntu1804_10.0.130-1_amd64.deb && \
+    ##apt-key adv --fetch-keys http://developer.download.nvidia.com/compute/cuda/repos/ubuntu1604/x86_64/7fa2af80.pub && \
```
mkrufky: Again, if this commented code isn't needed, then it should be removed.

Another thing to point out - as I believe this was based on one of my Dockerfiles, the first few commands in the "final stage" must be exactly identical to the first few commands of the "build stage", in order to take advantage of cache optimization.

docker/Dockerfile.gpu
```diff
@@ -0,0 +1,149 @@
...
+
+    curl -O http://ftp.ubuntu.com/ubuntu/ubuntu/pool/restricted/n/nvidia-graphics-drivers-418/libnvidia-decode-418_418.56-0ubuntu1_amd64.deb && \
+    dpkg --force-confdef --force-confnew -i ./libnvidia-decode-418_418.56-0ubuntu1_amd64.deb && \
+
+    curl -O http://ftp.ubuntu.com/ubuntu/ubuntu/pool/restricted/n/nvidia-graphics-drivers-418/libnvidia-encode-418_418.56-0ubuntu1_amd64.deb && \
+    dpkg --force-confdef --force-confnew -i ./libnvidia-encode-418_418.56-0ubuntu1_amd64.deb
+
+# everything above this line has already been cached in the build stage -
+# we save time and disk space by keeping the common parts of each stage
+# identical at the start of each image build
```
mkrufky: This is what I was talking about in my previous comment.    I'm not sure this comment is in the correct location -- it should be moved up, but only if the first few commands are identical to those in the build stage, otherwise remove the comment.

docker/Dockerfile.gpu
```diff
@@ -0,0 +1,149 @@
...
+#COPY docker/start.sh .
+#RUN chmod +x start.sh
+
+EXPOSE 7935/tcp
+EXPOSE 8935/tcp
+EXPOSE 1935/tcp
+
+ENTRYPOINT ["/usr/local/bin/gpu-start.sh"]
+CMD ["--help"]
+
```
mkrufky, line 119: I think this this should all be done in yet another stage.   Perhaps the "final" stage should be called "intermediate" and this would be the "final" stage.   I don't feel strongly about this, but it does feel like a lot of changes related to configuration rather than image building.

install_dependencies.sh
```diff
@@ -7,7 +7,7 @@ go get -u -v github.com/livepeer/m3u8
 go get -u -v go.opencensus.io/stats
 go get -u -v go.opencensus.io/tag
 go get -u google.golang.org/grpc
-go get github.com/alecthomas/gometalinter && gometalinter --install
+go get github.com/alecthomas/gometalinter #&& gometalinter --install
```
mkrufky: did something change with `gometalinter`?

darkdarkdragon:
@j0sh @ya7ya do we really want to have keylase patch in our repo?

darkdarkdragon:
@ya7ya what is the purpose of this file? CI process that Eli created already creates docker image with GPU support (but I can't make it work, getting `Cannot load libcuda.so.1 Could not dynamically load CUDA`). Maybe we should just fix our CI build?

j0sh:
> CI process that Eli created already creates docker image with GPU support 

@darkdarkdragon My understanding is that the current build works for extracting GPU-enabled binaries from Docker and running them standalone. However, to actually run GPUs within docker (eg, for test harness), we need to use a special base image provided by nvidia -- `nvidia/cuda:10.1-devel-ubuntu18.04` from the looks of it. So this PR introduces a new build meant for running GPUs within Docker, using that nvidia base image.

@ya7ya please correct me if that's not the intent here.

darkdarkdragon:
@j0sh image that is being built and pushed to Docker Hub is based on `nvidia/cuda:10.1-base`.

docker/Dockerfile.gpu
ya7ya: not really, I'm just removed them.

install_dependencies.sh
ya7ya: nop, I don't remember why I commented this out. sorry 🙏

ya7ya:
> do we really want to have keylase patch in our repo?

@darkdarkdragon No, we don't, I removed it and used `git clone` to grab it instead

docker/Dockerfile.gpu
ya7ya: I fixed this to work with `go mod`, thanks @mkrufky , that snippet worked 💯

ya7ya:
> what is the purpose of this file? CI process that Eli created already creates docker image with GPU support (but I can't make it work, getting Cannot load libcuda.so.1 Could not dynamically load CUDA). Maybe we should just fix our CI build?

I couldn't get it to work either, and after a week of debugging along with the team i opted to using @mkrufky build with few modifications .

also, the `go-livepeer:master` doesn't have the keylase patch included or the JSON key initiation required for the `test-harness`

j0sh:
Is this PR still needed now that https://github.com/livepeer/go-livepeer/pull/1101 is merged?

ya7ya:
nop

wohlner:

PR #1067, opened 2019-08-29
Refactor eventservices.RoundService

### LINKED ISSUE #1041: Refactor eth.RoundService to only poll without creating a subscription

yondonfu:

This PR refactors eventservices.RoundService and extracts it into `RoundInitializer` which is defined in the base `eth` package. The new `RoundInitializer` type polls at regular intervals to determine when to initialize the round which allows us to remove the legacy event monitor/subscription code that is no longer needed since we are now using the `blockwatch` package and the types implemented in `eth/watchers` for event subscriptions.

- f92002b Removes the `LatestBlockNum()` method from `LivepeerEthClient` since it isn't needed anymore
- c48b049 adds the RoundInitializer type
- ee1404e and 185f7f5 remove all code that references the legacy event monitor/subscription code
- 070ab3d Initializes the RoundInitializer type on startup if the node is in on-chain mode as an orchestrator with the `-initializeRound` flag set to true

Note: The RoundInitializer type largely uses the same logic for determining when to initialize the round as an orchestrator with a few minor tweaks - see the comments in `eth/roundinitializer.go` for more details. There is probably room for improvement on the strategy for round initialization that can be worked on later on.

Added unit tests. Ran multiple orchestrators in on-chain mode using the devtool and observed them taking turns initializing the round.

Fixes #1041

**Checklist:**


eth/roundinitializer.go
```diff
@@ -0,0 +1,188 @@
...
+	return nil
+}
+
+func (r *RoundInitializer) shouldInitialize(epochSeed *big.Int) (bool, error) {
+	transcoders, err := r.client.RegisteredTranscoders()
+	if err != nil {
+		return false, err
+	}
+
+	numActive, err := r.client.NumActiveTranscoders()
```
kyriediculous, line 130: In #1068 the `RoundsWatcher` will have a cached value for `transcoderPoolSize`. Since we already use it we could use the cached value instead.

We would have to change `BlockHashReader` for another interface that exposes `GetTranscoderPoolSize`.

This can be done in a seperate PR though, just putting it out there in case we'd want to do that.

kyriediculous (commented):
LGTM  🕵 

Just one comment for future reference

kyriediculous (approved):

PR #1068, opened 2019-09-03
cmd, eth, pm: O monitors B's reserve allocation towards O from cached values

### LINKED ISSUE #932: O should monitor B's reserve after it is cached

At the moment, O caches B's reserve alloc (i.e. the portion B's reserve guaranteed to O). There are a few cases where B's reserve alloc for O could change:

- B tops off its reserve
- The round changes and the reserve alloc changes because there are more or less registered Os (the reserve alloc = reserve / # of registered Os)

O might also want to stop working with B if:

- B initiates an unlock of its deposit/reserve
- B's reserve is frozen

O doesn't necessary need to stop working with B in these last two cases - this could be the simple MVP.

After O caches the reserve alloc for a B, it should watch for on-chain events and update the cached reserve alloc for B accordingly in the background.

If O does not watch for on-chain events and update the cached reserve alloc for B accordingly, O will compute ticket params for B based on inaccurate information i.e. it could set the required faceValue based on a maxFloat that is higher than what it actually is.

Another alternative is to perform a contract call to re-fetch B's reserve alloc every time O needs it (i.e. when providing ticket params and when receiving a ticket), but this would introduce many network calls to an already latency sensitive workflow which feels undesirable.

**Additional context**

kyriediculous:
This PR allows an O to keep track of his reserve allocation from a B based on its reserve, how much O has already claimed from the reserve (for a `reserveNonce`) and the transcoderPoolSize. 

When the reserve allocation is needed it will be calculated from the cached `reserve`, `claimedReserve` and `transcoderPoolSize` values. When a cache is not present for B, an RPC call will be made. 

Note: commit history needs some cleanup after rebasing to master to include the requested changes in #1054 

- Added a `claimedReserve` map to `SenderWatcher` to keep track of the claimed reserve for a `claimant` from a `reserveHolder`. This method has the same signature as it's contract counterpart. `claimedReserve` for a `reserveHolder` is cleared when the `reserveNonce` is incremented. 
- Expose the `GetTranscoderPoolSize` method from the `LivepeerEthClient` and add a cache for `transcoderPoolSize` to the `RoundsWatcher`
- Created a new method on `SenderMonitor`, `reserveAlloc(reserveHolder, claimant)` that calculates the reserveAllocation for a B to O , based on the cached `reserve`, `claimedReserve` and `transcoderPoolSize` values. If these values are not present for a sender, RPC calls will be made to initialize these values on the cache. 
- Removed the embedded `SenderManager` interface from the `Broker` interface since the implementation for `SenderManager` is no longer a `LivepeerEthClient` instance but `SenderWatcher`. Additionally we are now initializing `SenderMonitor` with both a `SenderWatcher` (for `reserve` and `claimedReserve`) and a `RoundsWatcher` (for `transcoderPoolSize`) 

Adjusted & ran unit tests

Fixes #932 

**Checklist:**


yondonfu:
Let's do a rebase to resolve conflicts prior to review since #1054 was just merged?

kyriediculous:
Done !

eth/watchers/roundswatcher.go
```diff
@@ -75,6 +88,10 @@ func (rw *RoundsWatcher) Watch() error {
 	}
 	rw.setLastInitializedRound(lr, bh)
 
+	if err := rw.fetchAndSetTranscoderPoolSize(); err != nil {
+		glog.Errorf("error fetching initial transcoderPoolSize: %v", err)
```
yondonfu: Since we're returning an error on L83 and L87 let's also return an error here.

eth/watchers/senderwatcher.go
```diff
@@ -62,6 +64,24 @@ func (sw *SenderWatcher) setSenderInfo(addr ethcommon.Address, info *pm.SenderIn
 	sw.senders[addr] = info
 }
 
+// ClaimedReserve returns the amount claimed from a sender's reserve
+func (sw *SenderWatcher) ClaimedReserve(sender ethcommon.Address, claimant ethcommon.Address) (*big.Int, error) {
```
yondonfu: Since the `ClaimedReserve()` method in the `LivepeerEthClient` interface uses `reserveHolder` and `claimant` as the parameter names, can we use the same parameter names here for consistency?

eth/watchers/senderwatcher.go
```diff
@@ -62,6 +64,24 @@ func (sw *SenderWatcher) setSenderInfo(addr ethcommon.Address, info *pm.SenderIn
...
+	sw.mu.RLock()
+	claimed := sw.claimedReserve[sender]
+	sw.mu.RUnlock()
+	if claimed != nil {
+		return claimed, nil
+	}
+	claimed, err := sw.lpEth.ClaimedReserve(claimant, sender)
```
yondonfu: Looks like the parameter order is reversed - should be `sender` followed by `claimant`

pm/recipient.go
```diff
@@ -332,16 +334,12 @@ func (r *recipient) redeemWinningTicket(ticket *Ticket, sig []byte, recipientRan
...
-		}
-
+		r.sm.QueueTicket(ticket.Sender, &SignedTicket{ticket, sig, recipientRand})
 		glog.Infof("Queued ticket sender=%x recipientRandHash=%x senderNonce=%v", ticket.Sender, ticket.RecipientRandHash, ticket.SenderNonce)
-
 		return nil
 	}
 
-	info, err := r.broker.GetSenderInfo(ticket.Sender)
+	info, err := r.smgr.GetSenderInfo(ticket.Sender)
```
yondonfu: I think we can actually get rid of this and the check on L349 and instead before we check if `maxFloat < ticket.FaceValue` we can check if `maxFloat = 0`. If `maxFloat = 0` then we return an error. The reasoning is that if `maxFloat > 0` then the sender definitely has a non-zero reserve so the current ticket could be eligible for redemption. And if `maxFloat = 0` then either the sender's reserve is 0 OR the current ticket is not eligible for redemption because the recipient has already claimed the max amount from the sender's reserve. In either case, we can just return an error and not try to submit the ticket on-chain.

Removing this check would also remove the `SenderManager` dependency in `Recipient` which is an added bonus.

eth/client.go
```diff
@@ -95,6 +96,7 @@ type LivepeerEthClient interface {
 	GetSenderInfo(addr ethcommon.Address) (*pm.SenderInfo, error)
 	ClaimableReserve(reserveHolder, claimant ethcommon.Address) (*big.Int, error)
```
yondonfu: Think we can remove this method from the interface if we're no longer implementing it anywhere?

pm/broker.go
```diff
@@ -90,10 +88,16 @@ type RoundsManager interface {
 	LastInitializedRound() *big.Int
 	// LastInitializedBlockHash returns the blockhash of the block the last round was initiated in
 	LastInitializedBlockHash() [32]byte
+	// GetTranscoderPoolSize returns the size active transcoder set for a round
```
yondonfu: "size of the"

pm/sendermonitor.go
```diff
@@ -157,82 +158,75 @@ func (sm *senderMonitor) SubFloat(addr ethcommon.Address, amount *big.Int) error
...
+	return reserveAlloc.Sub(reserveAlloc, sm.senders[addr].pendingAmount), nil
+}
+
+func (sm *senderMonitor) reserveAlloc(addr ethcommon.Address) (*big.Int, error) {
+	info, err := sm.smgr.GetSenderInfo(addr)
+	if err != nil {
+		return nil, err
+	}
+	claimed, err := sm.smgr.ClaimedReserve(addr, sm.claimant)
+	return new(big.Int).Sub(new(big.Int).Div(info.Reserve, sm.rm.GetTranscoderPoolSize()), claimed), nil
```
yondonfu: While the case is unlikely, let's return 0 if `sm.rm.GetTranscoderPoolSize()` is 0 just to make sure that a divide by zero is never possible.

pm/sendermonitor.go
```diff
@@ -157,82 +158,75 @@ func (sm *senderMonitor) SubFloat(addr ethcommon.Address, amount *big.Int) error
...
 // reserveAlloc - pendingAmount
 // Caller should hold the lock for senderMonitor
-func (sm *senderMonitor) maxFloat(addr ethcommon.Address) *big.Int {
-	return new(big.Int).Sub(sm.senders[addr].reserveAlloc, sm.senders[addr].pendingAmount)
+func (sm *senderMonitor) maxFloat(addr ethcommon.Address) (*big.Int, error) {
+	reserveAlloc, err := sm.reserveAlloc(addr)
+	if err != nil {
+		return nil, err
+	}
+	return reserveAlloc.Sub(reserveAlloc, sm.senders[addr].pendingAmount), nil
```
yondonfu: Might be worth allocating a new big.Int here instead of mutating `reserveAlloc` to make sure we never accidentally mutate a value we're not supposed to (not a problem now since `reserveAlloc()` returns a fresh big.Int

yondonfu (changes requested):

eth/watchers/roundswatcher.go
```diff
@@ -75,6 +88,10 @@ func (rw *RoundsWatcher) Watch() error {
 	}
 	rw.setLastInitializedRound(lr, bh)
 
+	if err := rw.fetchAndSetTranscoderPoolSize(); err != nil {
+		fmt.Errorf("error fetching initial transcoderPoolSize: %v", err)
```
yondonfu: Missing a return here

pm/recipient_test.go
```diff
@@ -910,6 +873,7 @@ func TestRedeemWinningTicket_MaxFloatError(t *testing.T) {
 	assert.EqualError(err, sm.maxFloatErr.Error())
 }
 
+/*
```
yondonfu: Can we just remove this test instead of commenting it out?

pm/stub.go
```diff
@@ -478,3 +467,32 @@ func (em *stubErrorMonitor) AcceptErr(sender ethcommon.Address) bool {
...
+
+func (sw *stubSenderWatcher) GetSenderInfo(addr ethcommon.Address) *SenderInfo {
+	return sw.info
+}
+
+func (sw *stubSenderWatcher) ClaimedReserve(sender ethcommon.Address) *big.Int {
+	return sw.claimedReserve
+}
+
+type stubRoundsWatcher struct {
```
yondonfu: Do we need this stub type? Can we just use `stubRoundsManager`?

pm/stub.go
```diff
@@ -478,3 +467,32 @@ func (em *stubErrorMonitor) AcceptErr(sender ethcommon.Address) bool {
 func (em *stubErrorMonitor) ClearErrCount(sender ethcommon.Address) {
 	em.acceptable = true
 }
+
+type stubSenderWatcher struct {
```
yondonfu: Do we need this stub type? Can we just use `stubSenderManager`?

pm/sendermonitor_test.go
```diff
@@ -380,6 +424,51 @@ func TestCleanup(t *testing.T) {
...
+		WithdrawBlock: big.NewInt(0),
+		ReserveState:  NotFrozen,
+		ThawRound:     big.NewInt(0),
+	}
+	smgr.claimedReserve[addr] = big.NewInt(100)
+	rm := &stubRoundsManager{
+		transcoderPoolSize: big.NewInt(5),
+	}
+	em := &stubErrorMonitor{}
+	sm := &senderMonitor{
```
yondonfu: Can we just use the `NewSenderMonitor` constructor here?

pm/sendermonitor_test.go
```diff
@@ -380,6 +424,51 @@ func TestCleanup(t *testing.T) {
...
+}
+
+func TestReserveAlloc(t *testing.T) {
+	assert := assert.New(t)
+
+	claimant := RandAddress()
+	b := newStubBroker()
+	addr := RandAddress()
+	smgr := newStubSenderManager()
+	smgr.info[addr] = &SenderInfo{
```
yondonfu, line 408: Looks like this initialization code is used a lot in this file. Could we extract it into a helper?

pm/sendermonitor_test.go
kyriediculous: We can't because NewSenderMonitor returns the interface, not the concrete type. Therefore we wouldn't have access to the unexported `reserveAlloc` method which we are testing in this test.

yondonfu: Ah right. A shortcut here could be to use `NewSenderMonitor` and then use a type assertion to convert the interface into the `senderMonitor` concrete type.

yondonfu (commented):
Let's rebase against master to fix the failing local Docker build in CI. I also left a small comment about a shortcut in the `reserveAlloc()` test - if you use that approach feel free to make the change and then rebase and I'll just look for the change in the end diff post-rebase.

kyriediculous:
Rebased and used a type assertion alongside the `NewSenderMonitor` constructor in the `reserveAlloc` test

yondonfu:
Changes look good. Blocked by failing CI due to #1071 though

yondonfu:
CI should be fixed since #1080 was merged. Let's rebase with master to fix CI!

kyriediculous:
Rebased & passing now

yondonfu (approved):
LGTM! :shipit:

Issue #1072, opened 2019-09-05
Server tests take a long time to run on macOS


j0sh:
**Describe the bug**
While the unit tests in `server` all pass, they take a long time to run on macOS. In particular, it seems like we wait 2+ minutes for a particular test to (successfully) fail.

```
I0905 01:36:10.751574    1718 ot_rpc.go:137] Transcoding taskId=742 url=linktomanifest
E0905 01:38:41.396707    1718 ot_rpc.go:177] Error submitting results Post https://badaddress/transcodeResults: dial tcp 23.202.231.166:443: connect: operation timed out
```

**To Reproduce**
1. Go to `server` dir
2. Run `go test -logtostderr=true`
3. Wait
4. Should be stuck around this line: `Transcoding taskId=742 url=linktomanifest`
4. Tests should eventually pass after something times out

**Expected behavior**
Timeouts to elapse quicker.

**Desktop (please complete the following information):**
 - OS: macOS Mojave 10.14
- Golang: go version go1.12.7 darwin/amd64

Also tested on:
- OS: macOS High Sierra 10.13
-Golang: go version go1.12 darwin/amd64 (presumably this is 12.0)

**Additional context**
This test seems to have completed quickly on Linux, so may be a platform specific issue:
```
$ cat /etc/lsb-release 
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=18.04
DISTRIB_CODENAME=bionic
DISTRIB_DESCRIPTION="Ubuntu 18.04.2 LTS"
$ go version
go version go1.12.1 linux/amd64
```

j0sh:
Seems OK now; may have been some weirdness around the version of the golang compiler used. This works now:

```
$ go version
go version go1.13.3 darwin/amd64
```

PR #1073, opened 2019-09-05
Copy audio streams

j0sh:

Converts go-livepeer to use stream copy for audio, rather than transcoding audio to AAC.

* This allows us to verify audio for Streamflow, which can be bitexact.
* Sidesteps an issue with the LPMS transcoder producing a clicking sound : https://github.com/livepeer/lpms/issues/152

- https://github.com/livepeer/go-livepeer/pull/1073/commits/59a9c112369c6a2f574443dfca405feb25116030 mod: Update LPMS to livepeer/lpms@7f095b3
- https://github.com/livepeer/go-livepeer/pull/1073/commits/7df3e331e47c4e822a4997982551ea194fc94349 mod: go mod tidy
- https://github.com/livepeer/go-livepeer/pull/1073/commits/f5259456cbe2d1bd4f7be33fd10894dd1ef33587 core: Stream copy audio rather than transcoding.

* Added unit tests, including tests for bitexactness
* Manual testing. Visually verified that the spectrogram of transcoded audio is smooth.

* Unit tests pass within `core` but didn't run the full test suite because of https://github.com/livepeer/go-livepeer/issues/1072

```
I0905 01:36:10.751574    1718 ot_rpc.go:137] Transcoding taskId=742 url=linktomanifest
```

**Checklist:**


iameli:
I'ma merge this into the new `media-integration` branch, test it out in the Wowza workflow.

j0sh:
Rebased to master, changed from a vendored LPMS update to using `go mod` via  `go get -u github.com/livepeer/lpms`. Let me know if it's done incorrectly.

I'd like to hold off on merging this until the fix for https://github.com/livepeer/lpms/issues/154 is in.

darkdarkdragon:
@j0sh it's better to add version tag to `lams` repo, this will allow go.mod to use it as reference, so instead of 
`github.com/livepeer/lpms v0.0.0-20190913184151-0246c39344c6` it will be 
`github.com/livepeer/lpms v0.1.0` (if we will add `v0.1.0` as tag to current `lpms` master)

j0sh:
@darkdarkdragon I guess this means we'd have to update the version every time we do a sync with go-livepeer which seems mildly annoying? (And that presumes that go-livepeer is the only user of LPMS, which is probably accurate for now, but it does make me question why we have a separate library at all.)

darkdarkdragon:
@j0sh 
>  I guess this means we'd have to update the version every time we do a sync with go-livepeer

yes

> which seems mildly annoying

why? for me it's seems logical - it will clearly distinct versions we intend to be used in other projects (yes, in `go-livepeer` :)

> And that presumes that go-livepeer is the only user of LPMS

why?

> but it does make me question why we have a separate library at all

at least for the sake of decoupling our components :)

mkrufky:
separation of components into logical parts is a good thing - version dependency changes are always tough to adjust to at first, but in the longrun, it will be better for the organization of the project.   Having LPMS as a separate module is immensely important, because we want to be able to enable developers to build upon our technology.

j0sh:
>>And that presumes that go-livepeer is the only user of LPMS
>
> why?

Suppose we had a dozen projects using LPMS, then the workflow of bumping upstream LPMS version each time a downstream dependency needs a particular change doesn't make sense. (If we ever hit the point of having dozens of downstream dependencies, we may want to commit to formal releases with an actual compatibility and versioning policy, but that step seems premature right now.)

Basically I'm a little skeptical of introducing a mandatory workflow change if we're not going to do actual, formal releases of LPMS. Is there any reason other than aesthetics to do this? (Which would be a somewhat strange reason for a file that's essentially machine generated and machine readable.)

Hashes seem to give us everything we need and are immediately comprehensible (at least to me). I'm not sure where/how the version number works - within a `go mod` config? A special file? Is it a `git tag` convention? What if more commits are added on top after defining a particular version - how does `go mod` resolve that? (Yeah, I could google all this and figure it out, but the basic point is the semantics of a "version number" here aren't immediately obvious, while the hash unambiguously points to a very specific snapshot of the code, and that seems preferable... at least, preferable to me.)

j0sh:
Rebased to update LPMS to https://github.com/livepeer/lpms/commit/7f095b315d48969f313aaa1152ac1b9ab6ad1509

Not sure why there are those extra lines in `go.sum` though, they seem somewhat redundant.

https://github.com/livepeer/go-livepeer/pull/1073/files#diff-f949e2d81c8076ebbf8af38fcbb72c1f

mkrufky:
> Rebased to update LPMS to [livepeer/lpms@7f095b3](https://github.com/livepeer/lpms/commit/7f095b315d48969f313aaa1152ac1b9ab6ad1509)
> 
> Not sure why there are those extra lines in `go.sum` though, they seem somewhat redundant.
> 
> https://github.com/livepeer/go-livepeer/pull/1073/files#diff-f949e2d81c8076ebbf8af38fcbb72c1f

@j0sh please do `go mod tidy` and then check in the results.

j0sh:
After `go mod tidy` :

```diff
commit 7df3e331e47c4e822a4997982551ea194fc94349
Author: Josh Allmann <joshua.allmann@gmail.com>
Date:   Wed Sep 18 14:41:49 2019 -0700

    mod: go mod tidy

 go.mod |  22 ++++---
 go.sum | 545 ++++++++++------------------------------------------------------------------------------------------------------------------------------------------------------
 2 files changed, 49 insertions(+), 518 deletions(-)
```

mkrufky:
> After `go mod tidy` :
> 
> ```diff
> commit 7df3e331e47c4e822a4997982551ea194fc94349
> Author: Josh Allmann <joshua.allmann@gmail.com>
> Date:   Wed Sep 18 14:41:49 2019 -0700
> 
>     mod: go mod tidy
> 
>  go.mod |  22 ++++---
>  go.sum | 545 ++++++++++------------------------------------------------------------------------------------------------------------------------------------------------------
>  2 files changed, 49 insertions(+), 518 deletions(-)
> ```

Looks like some other dependencies got updated as well, but the duplicated lpms entries are removed :-)

You can also experiment with `go mod update` to update them all or `go get` to fetch specific revisions of any dependency

darkdarkdragon:
@j0sh 
> I'm not sure where/how the version number works - within a go mod config

if your right now tag current lpms HEAD wit tag `v.0.1.0` - then in `go.mod` it will be
`github.com/livepeer/lpms v0.1.0`
if you then add new commit on top of It - it will look like
`github.com/livepeer/lpms v0.1.0-20190820150402-f1b88b60503b`

>  here aren't immediately obvious, 

it's pretty obvious - it's a git's tag == just named commit
for me it looks better because it shows intent to release tagged changes, not just pointing to random commit in the tree.
but yeah, with only one user it's not that important

>  Is there any reason other than aesthetics to do this?

no, aesthetics only

mk-livepeer (approved):

darkdarkdragon (approved):

j0sh:
> Looks like some other dependencies got updated as well

Hmm that could be risky, but this is a good time as any to update those dependencies. We'll need to be a little more careful about this as the release draws closer though.

PR #1081, opened 2019-09-11
switch go `go.mod`

### LINKED ISSUE #1071: CircleCI failing on localdocker build
**Describe the bug**
Recent builds are failing on CircleCI. Examples:

https://circleci.com/gh/livepeer/go-livepeer/2773?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

https://circleci.com/gh/livepeer/go-livepeer/2771?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

**To Reproduce**
1. Push a go-livepeer branch to Github (doesn't have to be a PR)
1.  Wait for CI to run
1. CI fails

**Expected behavior**
1. CI to pass

**Additional context**
Not sure if this is an issue on Circle's end because recent merges seem to have been OK.

darkdarkdragon:
Changes build process to use go modules.

Removed `vendor` dir, removed imports of `ipfs` code.

tests pass 

Fixes #1071 

**Checklist:**


mkrufky (changes requested):
Great stuff!  I'm very happy to finally see this happening-  IPFS was the last thing preventing us from moving to go.mod, thank you for handling this!

One critique:  We aren't being consistent in every Dockerfile.   Some do:
```
COPY . . 
```
While others do:
```
ADD . .
```
I believe that the better way is to use `COPY`.   As per [this post](https://nickjanetakis.com/blog/docker-tip-2-the-difference-between-copy-and-add-in-a-dockerile), If you’re copying in local files to your Docker image, always use COPY because it’s more explicit.

mk-livepeer (commented):
Why do we remove `docker/Dockerfile` and `docker/Dockerfile.alpine` ?

darkdarkdragon:
@mk-livepeer 
> Why do we remove docker/Dockerfile and docker/Dockerfile.alpine ?

not used anywhere. not maintained.

mk-livepeer (approved):

j0sh (changes requested):
The IPFS code has been essentially disabled for a long time, and I don't really see a path for us re-enabling it anytime soon, so let's delete all the IPFS-related code within goclient rather than leaving it commented out.

net/lp_rpc.proto
```diff
@@ -44,7 +44,7 @@ message OSInfo {
   enum StorageType {
     DIRECT     = 0;
     S3         = 1;
-    IPFS       = 2;
+    // IPFS       = 2;
```
mk-livepeer: Do external packages depend on this enum?   Do we need to leave GOOGLE at the value of 3 for any reason?   Should we leave IPFS there even though it may not be used?

The golang idiomatic way of declaring enums is a little different than what we might be used to doing in c / c++ ..   I think this would be better to use `iota` if we're not locked into these integers for any reason.  See here:  https://stackoverflow.com/questions/14426366/what-is-an-idiomatic-way-of-representing-enums-in-go

darkdarkdragon: > Do external packages depend on this enum? 

not yet

>  Do we need to leave GOOGLE at the value of 3 for any reason?

no. I've changed it with next commit

> The golang idiomatic way of declaring enums

protobuf allows us to define enums, so it's better to use it

j0sh: Note that we should really never remove or renumber fields from a protobuf definition, because that breaks compatibility between different client versions. It's probably OK right now because I don't think we've actually made a release with any of this, but something to keep in mind as we go forward. (Might break nodes on the internal testnet though if anyone tries to use GCS...)

darkdarkdragon: > Note that we should really never remove or renumber fields from a protobuf definition, 

Yes, that's why initially I haven't changed it. But, given that it's actually haven't been deployed anywhere outside Livepeer, I think we can change it now.

darkdarkdragon:
@j0sh so, is it good to go?

j0sh (approved):

j0sh:
Good to go! 🚢

j0sh:
Regarding the test2.ts binary file rename. Do we know if that creates a whole new blob in Git? That would make the repo even larger than it already is. Going forward, would love to consider using alternatives like git-lfs to mitigate repo bloat from checking in or modifying binaries.

darkdarkdragon:
> Regarding the test binary file rename. Do we know if that creates a whole new blob in Git?

Don't have any clue. And quick googling doesn't answer this question :(

iameli:
If the file contents haven’t changed, my understanding is that the blob
stays the same with just a different ref pointing to it.

mkrufky:
If it's just a file rename, then it shouldn't create a new blob -- it's just a metadata change

PR #1095, opened 2019-09-20
ci: prevent upload if git state is dirty

iameli:
This will prevent a build from being uploaded and fail CI if a build is dirty. It doesn't fix the problem, presently, but it's a failing test case if anyone is looking to fix the problem.


iameli:
Superceded by #1122

wohlner:

Issue #1093, opened 2019-09-17
Verifier API Task Breakdown


j0sh:
**Is your feature request related to a problem? Please describe.**

The [Epic Labs API](https://github.com/livepeer/verification-classifier/tree/master/api) needs to be integrated into go-livepeer in order to gather real world data on the verification process.

**Describe the solution you'd like**

### Phase 1 : API Integration

Most of these have been implemented within https://github.com/livepeer/go-livepeer/pull/1096

- [x] Add API address CLI flag
  * Will this flag be subject to change once the full policy implementation complete?
  * Do we ship testnet / mainnet with a default verifier API address (eg, Livepeer operated)?
- [x] API type flag ( ? )
  * Built into the API address itself, eg `-epicVerifier http://localhost:port/classifier`
  * Could be built into the address scheme itself, eg `-verifierAddr epic://localhost:port/classifier`
      * May be difficult to distinguish protocol, eg HTTP vs HTTPS
  * Flag for a policy file can also define the type, eg `-verificationPolicy /path/to/policy/file` with CLI flags for each policy option, including the URL
- [x] Define interface for verifier + policy + API parameters
  * Sketch out compatibility with a couple verifier implementations
  * Epic Labs and one other (eg, coersion or an hypothetical SSIM verifier)
- [x] Implement call-out to Epic Labs API
  * Prior to inserting in HLS playlist
  * Only if on-chain *and* a verifier address is specified via CLI
  * Error case from verifier will result in a drop for now - can be configured via policy later.
- [ ] Clean up surrounding code as necessary (eg, remove pixel counting)
- [x] Update documentation on flags / API usage

### Phase 2 : Results Interpretation and Policy Enforcement

Most of these items need further design / spec work but the broad outlines are as following:

- [x] Define policy defaults for verification classifier: https://github.com/livepeer/verification-classifier/issues/61
- [x] Check audio MD5 hash within verifier : https://github.com/livepeer/verification-classifier/issues/60
- [x] Policy implementation in go-livepeer with sequential retries : https://github.com/livepeer/go-livepeer/pull/1096
- [ ] Parallel implementation of retries
- [ ] Ban Policy : https://github.com/livepeer/go-livepeer/issues/1100#issuecomment-558714000

### Phase 3: Dynamic Policies

We might have to consider mid-stream changes to various policy parameters if we detect that the threshold is set too high for a given content, or we may want to adjust the sampling rate based on history or risk tolerance, and so forth.

[...]

PR #1096, opened 2019-09-20
Epic Labs verification classifier integration.

j0sh:

Extremely early integration of the Epic Labs verification classifier API. This should be enough to start running some experiments locally.

Refer to https://github.com/livepeer/go-livepeer/issues/1093 for some details on the following items:

 - Due to https://github.com/livepeer/verification-classifier/issues/64 this works when the verifier is running on localhost only. There is a lot hard-coded in here.
- Results (and errors) from the verification function are ignored completely.
- Adds a new CLI flag, `-verifierUrl` (but this may change in later iterations of the integration)
- Only performs verification if the `-verifierUrl` address is set.
- Currently works in off-chain mode to facilitate quick testing.
- Verifies all segments and renditions; no sampling is done.
- Structure is subject to change as more of the policy is designed and implemented.
- Leaves surrounding code intact, eg pixel counting and sig checking.

- 
- 
- 

**Checklist:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,6 +128,14 @@ A standalone transcoder can be run which connects to a remote orchestrator. The 
 
 GPU transcoding on NVIDIA is supported; see the [GPU documentation](doc/gpu.md) for usage details.
 
+### Verification
+
+Experimental verification of video using the Epic Labs classifier can be enabled with the `-verifierUrl` flag. Pass in the address of the verifier API:
+
+- `livepeer -broadcaster -verifierUrl http://localhost:5000/verify -verifierPath /path/to/verifier`
+
+Refer to the [classifier documentation](https://github.com/livepeer/verification-classifier) for more details on getting the classifier API installed and running.
+
 ## Contribution
 Thank you for your interest in contributing to the core software of Livepeer.
 
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -90,6 +90,8 @@ func main() {
 	orchAddr := flag.String("orchAddr", "", "Orchestrator to connect to as a standalone transcoder")
 	verifierURL := flag.String("verifierUrl", "", "URL of the verifier to use")
 
+	verifierPath := flag.String("verifierPath", "", "Path to verifier shared volume")
+
 	// Transcoding:
 	orchestrator := flag.Bool("orchestrator", false, "Set to true to be an orchestrator")
 	transcoder := flag.Bool("transcoder", false, "Set to true to be a transcoder")
@@ -681,6 +683,13 @@ func main() {
 			server.Policy = &verification.Policy{Retries: 2, Verifier: &verification.EpicClassifier{Addr: *verifierURL}}
 			// TODO Set up a default "empty" verifier-less policy for onchain
 			//      that only checks sigs and pixels?
+
+			// Set the verifier path. Remove once [1] is implemented!
+			// [1] https://github.com/livepeer/verification-classifier/issues/64
+			if drivers.NodeStorage == nil && *verifierPath == "" {
+				glog.Fatal("Requires a path to the verifier shared volume when local storage is in use; use -verifierPath, S3 or GCS")
+			}
+			verification.VerifierPath = *verifierPath
 		}
 	} else if n.NodeType == core.OrchestratorNode {
 		suri, err := getServiceURI(n, *serviceAddr)
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -37,6 +37,7 @@ import (
 	"github.com/livepeer/go-livepeer/eth/blockwatch"
 	"github.com/livepeer/go-livepeer/eth/eventservices"
 	"github.com/livepeer/go-livepeer/eth/watchers"
+	"github.com/livepeer/go-livepeer/verification"
 
 	lpmon "github.com/livepeer/go-livepeer/monitor"
 )
@@ -87,6 +88,7 @@ func main() {
 	httpAddr := flag.String("httpAddr", "", "Address to bind for HTTP commands")
 	serviceAddr := flag.String("serviceAddr", "", "Orchestrator only. Overrides the on-chain serviceURI that broadcasters can use to contact this node; may be an IP or hostname.")
 	orchAddr := flag.String("orchAddr", "", "Orchestrator to connect to as a standalone transcoder")
+	verifierURL := flag.String("verifierUrl", "", "URL of the verifier to use")
 
 	// Transcoding:
 	orchestrator := flag.Bool("orchestrator", false, "Set to true to be an orchestrator")
@@ -667,6 +669,16 @@ func main() {
 			}
 			glog.Info("Using auth webhook URL ", *authWebhookURL)
 			server.AuthWebhookURL = *authWebhookURL
+		}
+
+		// Set up verifier
+		if *verifierURL != "" {
+			_, err := validateURL(*verifierURL)
+			if err != nil {
+				glog.Fatal("Error setting verifier URL ", err)
+			}
+			glog.Info("Using the Epic Labs classifier for verification at ", *verifierURL)
+			server.Verifier = &verification.EpicClassifier{Addr: *verifierURL}
 		}
 	} else if n.NodeType == core.OrchestratorNode {
 		suri, err := getServiceURI(n, *serviceAddr)
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -678,7 +678,9 @@ func main() {
 				glog.Fatal("Error setting verifier URL ", err)
 			}
 			glog.Info("Using the Epic Labs classifier for verification at ", *verifierURL)
-			server.Verifier = &verification.EpicClassifier{Addr: *verifierURL}
+			server.Policy = &verification.Policy{Retries: 2, Verifier: &verification.EpicClassifier{Addr: *verifierURL}}
+			// TODO Set up a default "empty" verifier-less policy for onchain
+			//      that only checks sigs and pixels?
 		}
 	} else if n.NodeType == core.OrchestratorNode {
 		suri, err := getServiceURI(n, *serviceAddr)
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -637,10 +637,11 @@ func main() {
 
 		// Set up orchestrator discovery
 		if *orchWebhookURL != "" {
-			whurl, err := getOrchWebhook(*orchWebhookURL)
+			whurl, err := validateURL(*orchWebhookURL)
 			if err != nil {
 				glog.Fatal("Error setting orch webhook URL ", err)
 			}
+			glog.Info("Using orchestrator webhook URL ", whurl)
 			n.OrchestratorPool = discovery.NewWebhookPool(bcast, whurl)
 		} else if len(orchURLs) > 0 {
 			n.OrchestratorPool = discovery.NewOrchestratorPool(bcast, orchURLs)
@@ -659,9 +660,13 @@ func main() {
 			// Not a fatal error; may continue operating in segment-only mode
 			glog.Error("No orchestrator specified; transcoding will not happen")
 		}
-		var err error
-		if server.AuthWebhookURL, err = getAuthWebhookURL(*authWebhookURL); err != nil {
-			glog.Fatal("Error setting auth webhook URL ", err)
+		if *authWebhookURL != "" {
+			_, err := validateURL(*authWebhookURL)
+			if err != nil {
+				glog.Fatal("Error setting auth webhook URL ", err)
+			}
+			glog.Info("Using auth webhook URL ", *authWebhookURL)
+			server.AuthWebhookURL = *authWebhookURL
 		}
 	} else if n.NodeType == core.OrchestratorNode {
 		suri, err := getServiceURI(n, *serviceAddr)
@@ -769,7 +774,7 @@ func main() {
 	}
 }
 
-func getOrchWebhook(u string) (*url.URL, error) {
+func validateURL(u string) (*url.URL, error) {
 	if u == "" {
 		return nil, nil
 	}
@@ -778,25 +783,9 @@ func getOrchWebhook(u string) (*url.URL, error) {
 		return nil, err
 	}
 	if p.Scheme != "http" && p.Scheme != "https" {
-		return nil, errors.New("Webhook URL should be HTTP or HTTPS")
+		return nil, errors.New("URL should be HTTP or HTTPS")
 	}
-	glog.Infof("Using orchestrator webhook url %s", u)
 	return p, nil
-}
-
-func getAuthWebhookURL(u string) (string, error) {
-	if u == "" {
-		return "", nil
-	}
-	p, err := url.ParseRequestURI(u)
-	if err != nil {
-		return "", err
-	}
-	if p.Scheme != "http" && p.Scheme != "https" {
-		return "", errors.New("Webhook URL should be HTTP or HTTPS")
-	}
-	glog.Infof("Using auth webhook url %s", u)
-	return u, nil
 }
 
 // ServiceURI checking steps:
--- a/server/broadcast.go
+++ b/server/broadcast.go
@@ -440,11 +440,6 @@ func transcodeSegment(cxn *rtmpConnection, seg *stream.HLSSegment, name string,
 			if monitor.Enabled {
 				monitor.TranscodedSegmentAppeared(nonce, seg.SeqNo, sess.Profiles[i].Name)
 			}
-			err = cpl.InsertHLSSegment(&sess.Profiles[i], seg.SeqNo, url, seg.Duration)
-			if err != nil {
-				errFunc(monitor.SegmentTranscodeErrorPlaylist, url, err)
-				return
-			}
 		}
 
 		for i, v := range res.Segments {
@@ -465,6 +460,18 @@ func transcodeSegment(cxn *rtmpConnection, seg *stream.HLSSegment, name string,
 			if err != nil {
 				glog.Errorf("Error verifying nonce=%d manifestID=%s seqNo=%d err=%s", nonce, cxn.mid, seg.SeqNo, err)
 				return err
+			}
+		}
+
+		for i, url := range segURLs {
+			err = cpl.InsertHLSSegment(&sess.Profiles[i], seg.SeqNo, url, seg.Duration)
+			if err != nil {
+				// We assume playlist insertion is *not* recoverable for now
+				glog.Errorf("Playlist insertion error nonce=%d manifestID=%s seqNo=%d err=%s", nonce, cxn.mid, seg.SeqNo, err)
+				if monitor.Enabled {
+					monitor.SegmentTranscodeFailed(monitor.SegmentTranscodeErrorPlaylist, nonce, seg.SeqNo, err, false)
+				}
+				return nil
 			}
 		}
 
--- a/server/broadcast.go
+++ b/server/broadcast.go
@@ -26,6 +26,7 @@ import (
 	"github.com/livepeer/lpms/stream"
 )
 
+var Verifier verification.Verifier
 var BroadcastCfg = &BroadcastConfig{}
 
 type BroadcastConfig struct {
@@ -440,16 +441,17 @@ func transcodeSegment(cxn *rtmpConnection, seg *stream.HLSSegment, name string) 
 			return dlErr
 		}
 
-		// Ignore errors for now
-		params := &verification.Params{
-			ManifestID:   sess.ManifestID,
-			Source:       seg,
-			Profiles:     sess.Profiles,
-			Orchestrator: sess.OrchestratorInfo,
-			Results:      res.TranscodeData,
+		if Verifier != nil {
+			// Ignore errors for now
+			params := &verification.Params{
+				ManifestID:   sess.ManifestID,
+				Source:       seg,
+				Profiles:     sess.Profiles,
+				Orchestrator: sess.OrchestratorInfo,
+				Results:      res.TranscodeData,
+			}
+			Verifier.Verify(params)
 		}
-		verifier := &verification.EpicClassifier{Addr: "http://localhost:5000/verify"}
-		verifier.Verify(params)
 
 		ticketParams := sess.OrchestratorInfo.GetTicketParams()
 		if ticketParams != nil && // may be nil in offchain mode
--- a/server/broadcast.go
+++ b/server/broadcast.go
@@ -8,6 +8,7 @@ import (
 	"math/big"
 	"net/url"
 	"os"
+	"strings"
 	"sync"
 
 	"github.com/golang/glog"
@@ -19,6 +20,7 @@ import (
 	"github.com/livepeer/go-livepeer/core"
 	"github.com/livepeer/go-livepeer/drivers"
 	"github.com/livepeer/go-livepeer/monitor"
+	"github.com/livepeer/go-livepeer/net"
 	"github.com/livepeer/go-livepeer/pm"
 	"github.com/livepeer/go-livepeer/verification"
 
@@ -26,7 +28,7 @@ import (
 	"github.com/livepeer/lpms/stream"
 )
 
-var Verifier verification.Verifier
+var Policy *verification.Policy
 var BroadcastCfg = &BroadcastConfig{}
 
 type BroadcastConfig struct {
@@ -280,15 +282,21 @@ func processSegment(cxn *rtmpConnection, seg *stream.HLSSegment) error {
 		}
 	}
 
+	var sv *verification.SegmentVerifier
+	if Policy != nil {
+		sv = verification.NewSegmentVerifier(Policy)
+	}
+
 	for {
 		// if fails, retry; rudimentary
-		if err := transcodeSegment(cxn, seg, name); err == nil {
+		if err := transcodeSegment(cxn, seg, name, sv); err == nil {
 			return nil
 		}
 	}
 }
 
-func transcodeSegment(cxn *rtmpConnection, seg *stream.HLSSegment, name string) error {
+func transcodeSegment(cxn *rtmpConnection, seg *stream.HLSSegment, name string,
+	verifier *verification.SegmentVerifier) error {
 
 	nonce := cxn.nonce
 	rtmpStrm := cxn.stream
@@ -362,6 +370,7 @@ func transcodeSegment(cxn *rtmpConnection, seg *stream.HLSSegment, name string) 
 
 		var dlErr, saveErr error
 		segHashes := make([][]byte, len(res.Segments))
+		segURLs := make([]string, len(res.Segments))
 		n := len(res.Segments)
 		segHashLock := &sync.Mutex{}
 		cond := sync.NewCond(segHashLock)
@@ -418,6 +427,15 @@ func transcodeSegment(cxn *rtmpConnection, seg *stream.HLSSegment, name string) 
 				}()
 			}
 
+			// Store URLs for the verifier. Be aware that the segment is
+			// already within object storage  at this point, whether local or
+			// external. If a client were to ignore the playlist and
+			// preemptively fetch segments, they could be reading tampered
+			// data. Not an issue if the delivery protocol is being obeyed.
+			segHashLock.Lock()
+			segURLs[i] = url
+			segHashLock.Unlock()
+
 			if monitor.Enabled {
 				monitor.TranscodedSegmentAppeared(nonce, seg.SeqNo, sess.Profiles[i].Name)
 			}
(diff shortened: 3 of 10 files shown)


server/broadcast.go
```diff
@@ -434,6 +444,13 @@ func transcodeSegment(cxn *rtmpConnection, seg *stream.HLSSegment, name string)
 		if dlErr != nil {
 			return dlErr
 		}
+
+		if verifier != nil {
+			if err := verify(verifier, cxn, sess, seg, res); err != nil {
```
yondonfu: If we end up initializing a [verifier-less policy](https://github.com/livepeer/go-livepeer/pull/1096/files#diff-205b550ef0ac99b368da16cfb883960bR664) responsible for sig & pixel count checking when running in on-chain mode, then pixel count checking would become synchronous instead of asynchronous. That might be an ok tradeoff for cleaner separation of verification related concerns - just noting this for now.

server/broadcast.go
```diff
@@ -462,6 +479,35 @@ func shouldStopSession(err error) bool {
...
+		Source:       source,
+		Profiles:     sess.Profiles,
+		Orchestrator: sess.OrchestratorInfo,
+		Results:      res,
+	}
+
+	// Verify. The params we receive, if any, are the *accepted* params
+	// The accepted params are not necessarily the same params we just have!
+	p, err := verifier.Verify(params)
+	if _, retry := err.(verification.Retryable); retry {
```
yondonfu: Is the plan to just check for `verification.Retryable` errors for now and consider distinguishing between fatal and verifier related errors later (as described in #1100)?

verification/verify.go
```diff
@@ -0,0 +1,133 @@
...
+	Profiles []ffmpeg.VideoProfile
+
+	// Information on the orchestrator that performed the transcoding
+	Orchestrator *net.OrchestratorInfo
+
+	// Transcoded result metadata
+	Results *net.TranscodeData
+}
+
+type VerificationResult interface {
```
yondonfu: Why use an interface here? Do you imagine implementations that do more than just return internal score and pixels fields (as is done by the `verificationResult` type)?

verification/epic.go
```diff
@@ -0,0 +1,163 @@
...
+
+func (vr *verificationResult) Pixels() []int64 {
+	return vr.pixels
+}
+func (vr *verificationResult) Score() float64 {
+	return vr.score
+}
+
+func epicResultsToVerificationResults(er *epicResults) (*verificationResult, error) {
+	// find average of scores and build list of pixels
```
yondonfu, line 68: Ah so an implication of our current transcoding workflow where we either receive all renditions requested or we consider the segment transcoding as failed is that if 1 out of N renditions is classified as tampered we retry for all N renditions. I feel like ideally in that case we only retry for the single rendition, but I get why retrying for all N renditions fits better with our current workflow.

server/broadcast.go
j0sh: Good point, I missed that the spec distinguishes between verifier errors and other types of fatal verification-related errors. I had been lumping them all under "retryable" errors (called such because a little weird to be retrying on supposed "fatal" errors).

The spec does imply that verifier errors shouldn't immediately remove the orchestrator from the working set, while fatal errors should - which is probably sensible and the reason for making a distinction between the two. Will adjust.

verification/verify.go
j0sh: Yeah, an interface is not really necessary here. Removed in b70f22f26f926415ae39a887ca1d74eb7646c1cd

j0sh: The linter was complaining about "stuttering" (`verification.VerificationResult`), and suggested just `Result` so I went with that (`Results` actually), the brevity is much better. Also took the liberty of renaming `VerifierParams` to simply `Params` as well for consistency. https://github.com/livepeer/go-livepeer/pull/1096/commits/cbe81d9208d5f65c63e9914c78781156747c9cc0

j0sh:
Functionally complete but missing tests. There are two ways to get data into the verifier:

- External object storage
- Shared volume

The shared volume (assumed to be mounted in the verifier container as `/stream`) actually works rather nicely for remote verifiers as long as the volume is accessible over the network to go-livepeer. I've been using sshfs with success.

Remote shared volumes along with support for broadcaster-external object storage (which also works nicely; tested with GCS) makes https://github.com/livepeer/verification-classifier/issues/64 a lot less urgent now.

@yondonfu I'd really like to clean up the commit log; let me know when you're OK with a rebase.

kyriediculous:
Will give this a look further in the week but I just had two minor questions for now: 

Will this verifier API use a trained model made available to the node so that it only has to run input through an already trained model, meaning that it could be run on lower end machines ? 

What is the default we are aiming for? Running locally or remote?

server/broadcast.go
```diff
@@ -412,14 +423,25 @@ func transcodeSegment(cxn *rtmpConnection, seg *stream.HLSSegment, name string)
...
+			// preemptively fetch segments, they could be reading tampered
+			// data. Not an issue if the delivery protocol is being obeyed.
+			segHashLock.Lock()
+			segURLs[i] = url
+			segHashLock.Unlock()
+
 			if monitor.Enabled {
 				monitor.TranscodedSegmentAppeared(nonce, seg.SeqNo, sess.Profiles[i].Name)
 			}
+			/* Should we do this conditionally in the presence of the verifier? Otherwise playlist insertions are blocked until all segments are available, which seems worse - but is necessary for verification
```
yondonfu: Seems reasonable to insert into the playlist when segments are available if `verifier = nil`

server/broadcast.go
```diff
@@ -434,6 +456,23 @@ func transcodeSegment(cxn *rtmpConnection, seg *stream.HLSSegment, name string)
...
+				return err
+			}
+		}
+
+		for i, url := range segURLs {
+			err = cpl.InsertHLSSegment(&sess.Profiles[i], seg.SeqNo, url, seg.Duration)
+			if err != nil {
+				// We assume playlist insertion is *not* recoverable for now
+				glog.Error("Playlist insertion error; is this recoverable? nonce=%d manifestID=%s seqNo=%d err=%s", nonce, cxn.mid, seg.SeqNo, err)
+				return nil
```
yondonfu, line 474: Should we log this error using the `monitor` package similar to what we did when we [inserted into the playlist as segments became available](https://github.com/livepeer/go-livepeer/pull/1096/files#diff-cecc27bfe2b6b19181ce443733dfba57R441)?

server/broadcast.go
```diff
@@ -462,6 +501,85 @@ func shouldStopSession(err error) bool {
...
+		// Error falls through towards end if necessary
+		cxn.sessManager.removeSession(sess)
+	}
+	if accepted != nil {
+		// The returned set of results has been accepted by the verifier
+
+		// Check if an earlier verification attempt was the one accepted.
+		// If so, reset the local OS if we're using that since it's been
+		// overwritten with this rendition.
+		if accepted != params && !sess.BroadcasterOS.IsExternal() {
```
yondonfu: The renditions wouldn't be overwritten in an external OS for multiple verification attempts?

verification/verify.go
```diff
@@ -0,0 +1,139 @@
...
+type SegmentVerifierResults struct {
+	params *Params
+	res    *Results
+}
+
+type byResScore []SegmentVerifierResults
+
+func (a byResScore) Len() int           { return len(a) }
+func (a byResScore) Swap(i, j int)      { a[i], a[j] = a[j], a[i] }
+func (a byResScore) Less(i, j int) bool { return a[i].res.Score < a[j].res.Score }
```
yondonfu: Won't this result in the first element of the sorted slice having the lowest score when we want the first element of the sorted slice to have the highest score?

verification/verify.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+
+	// Check for max retries
+	// If max hit, return best params so far
+	if sv.count >= sv.policy.Retries {
+		if len(sv.results) <= 0 {
+			return nil, nil
+		}
+		sort.Sort(byResScore(sv.results))
+		return sv.results[0].params, err
```
yondonfu: Given the current definition of [Less()](https://github.com/livepeer/go-livepeer/pull/1096/files#diff-98d66d2dc5372dedc76d5fdde7bedebcR81) won't the first element have the lowest score? Think we would want the last element here to get the result with the highest score

yondonfu:
@j0sh 

> I'd really like to clean up the commit log; let me know when you're OK with a rebase.

👍 to do a rebase. I left a few comments, but feel free to rebase before/after addressing them without a thumbs up. Once you feel like the PR is ready for a full review (I know you mentioned you wanted to add some unit tests and implement the fatal/verifier errors) let me know! In the meantime, I'll do some testing with both the shared Docker volume and an external OS for B.

@kyriediculous 

> Will this verifier API use a trained model made available to the node so that it only has to run input through an already trained model, meaning that it could be run on lower end machines ?

The verifier API would use a trained model.

> What is the default we are aiming for? Running locally or remote?

The user will have to run the verifier API separately from the node so it is up to them whether they want to run the verifier API locally or remotely [1]. For performance, they'll probably want to run it remotely. For ease use of use, we can consider automating the verifier API setup either via a script or via a docker-compose file for Docker users - we can discuss that separately though.

[1] At the moment, this PR enables slightly limited remote verifier API usage - see [this comment](https://github.com/livepeer/go-livepeer/pull/1096#issuecomment-562837768)

server/broadcast.go
j0sh: Thought about this and I'd really prefer not to have two code paths that do the same thing depending on which mode the node is running in. That can get very confusing and error-prone very quickly, and I would really like to avoid setting that precedent here.

While we do have a lot of different operating modes on the node, the composition has been rather clear-cut so far: a bit of additional logic might execute for on-chain mode, or an orchestrator may invoke a remote transcoder in standalone mode, but there aren't really cases AFAIK where we execute the same logic in different places, depending on the mode. So that's a slope I'd like to avoid going down as much as possible.

Removed this bit here: https://github.com/livepeer/go-livepeer/pull/1096/commits/55a8d8b683a0dde032081be9201ddae5113f6aa7

server/broadcast.go
j0sh: If an external OS is used, then each orchestrator writes to a different path. This avoids issues around overwriting, front-running, etc. The actual URL which gets inserted into the playlist is the one that passes verification.

https://github.com/livepeer/go-livepeer/blob/775a76ba16fc71bf901030f994b413f682ff3fdd/server/broadcast.go#L215-L220

verification/verify.go
j0sh: Yep - fixed (and tested!)

server/broadcast.go
j0sh: Fixed in https://github.com/livepeer/go-livepeer/pull/1096/commits/a153b1f8adc0725ef2aad4364960e4f40b32a876

j0sh:
* Added a number of unit tests, bug fixes resulting from those, etc.
* Addressed various bits of feedback
* Only major piece that doesn't have test coverage is the epic classifier integration code
* Still don't distinguish between retryable and fatal errors per https://github.com/livepeer/go-livepeer/pull/1096#discussion_r355035563

j0sh:
>>Will this verifier API use a trained model made available to the node so that it only has to run input through an already trained model, meaning that it could be run on lower end machines ?
>
> The verifier API would use a trained model.

Relatedly, would be good to resolve https://github.com/livepeer/verification-classifier/issues/63 so the node doesn't need to be aware of the model that's being used.

server/broadcast.go
```diff
@@ -432,6 +448,26 @@ func transcodeSegment(cxn *rtmpConnection, seg *stream.HLSSegment, name string)
...
+			if err != nil {
+				return err
+			}
+		}
+
+		for i, url := range segURLs {
+			err = cpl.InsertHLSSegment(&sess.Profiles[i], seg.SeqNo, url, seg.Duration)
+			if err != nil {
+				// We assume playlist insertion is *not* recoverable for now
+				glog.Errorf("Playlist insertion error; is this recoverable? nonce=%d manifestID=%s seqNo=%d err=%s", nonce, cxn.mid, seg.SeqNo, err)
```
yondonfu: Should we omit "is this recoverable?" from the error message?

server/broadcast.go
```diff
@@ -460,6 +496,86 @@ func shouldStopSession(err error) bool {
...
+				// Sanity check that we actually have the rendition data?
+				if len(data) <= 0 {
+					return errors.New("MissingLocalData")
+				}
+				// XXX fix required base path
+				_, err := sess.BroadcasterOS.SaveData(path.Base(accepted.URIs[i]), data)
+				if err != nil {
+					return err
+				}
+				// We *probably* don't need to reset URI[i] here
```
yondonfu: Why would we need a reset here?

server/broadcast.go
```diff
@@ -460,6 +496,86 @@ func shouldStopSession(err error) bool {
...
+		// Check if an earlier verification attempt was the one accepted.
+		// If so, reset the local OS if we're using that since it's been
+		// overwritten with this rendition.
+		if accepted != params && !sess.BroadcasterOS.IsExternal() {
+			for i, data := range accepted.Renditions {
+				// Sanity check that we actually have the rendition data?
+				if len(data) <= 0 {
+					return errors.New("MissingLocalData")
+				}
+				// XXX fix required base path
```
yondonfu: By "required base path", do you mean `<PROFILE_NAME>/<SEG_NO>` ([example](https://github.com/livepeer/go-livepeer/blob/master/server/broadcast.go#L382))?

server/broadcast_test.go
```diff
@@ -519,3 +545,212 @@ func TestVerifyPixels(t *testing.T) {
...
+	sv.err = verification.ErrTampered
+	sv.retries = 10 // Do this to ensure we get a nil result
+	_, retryable := sv.err.(verification.Retryable)
+	assert.True(retryable)
+	verifier = newStubSegmentVerifier(sv)
+	err = verify(verifier, cxn, sess, source, res, URIs)
+	assert.NotNil(err)
+	assert.Equal(2, sv.calls)
+	assert.Equal(sv.err, err)
+	assert.Len(bsm.sessMap, 0) // No effect on map for now
```
yondonfu: Should this comment be removed since the length of `bsm.sessMap` decreases for retryable errors such as `verification.ErrTampered`?

server/broadcast_test.go
```diff
@@ -519,3 +545,212 @@ func TestVerifyPixels(t *testing.T) {
...
+	verifier = newStubSegmentVerifier(sv)
+	assert.Equal(0, sv.calls)  // sanity check initial call count
+	assert.Len(bsm.sessMap, 1) // sanity check initial bsm map
+	err = verify(verifier, cxn, sess, source, res, URIs)
+	assert.NotNil(err)
+	assert.Equal(1, sv.calls)
+	assert.Equal(sv.err, err)
+	assert.Len(bsm.sessMap, 1) // No effect on map for now
+
+	// Check non-retryable errors, esp broadcast session removal from manager
```
yondonfu: Should this say retryable errors?

server/broadcast.go
j0sh: Whenever we use an external OS, the URI changes, so it does need to be reset. Fixed this bit in https://github.com/livepeer/go-livepeer/pull/1096/commits/7c067159f47ca94b032d8bfd15b2f8997e46d68b and added tests in https://github.com/livepeer/go-livepeer/pull/1096/commits/788ed40861b2fbbc7fcf384d33d77c533abf0345

server/broadcast.go
j0sh: The problematic part is actually the `/stream/<manifestID>` part of the base path. The OS `SaveData` method does this:

```
name, err := os.SaveData("<profile>/<segNo>", data)
```

where `name` is a full(er) path - if using a local OS, it'll look something like `/stream/<manifestID>/<profile>/<seqNo>`

However, the returned `name` is not composable with `SaveData()` itself; it can't be passed back into `SaveData` to over-write existing data. Ideally we could do something like this:

```
name1 = os.SaveData("<profile>/<segNo>", "abc")
name2 = os.SaveData(name1, "def")
name1 == name2
os.GetData(name1) == "def"
```

Ideally the OS itself would do the prefix stripping but that needs to be done on a per-OS basis (since the implementation of `SaveData` is per-OS). Right now we only need it for local OS though. We could also strip the prefix within the `verify` function just prior to saving within the OS, but that's an implementation detail leaking out; the driver API is smelly and could use some deeper thought.

Anyway, worked around the issue within `verify` here https://github.com/livepeer/go-livepeer/pull/1096/commits/7c067159f47ca94b032d8bfd15b2f8997e46d68b and tests modified to match / pass in https://github.com/livepeer/go-livepeer/pull/1096/commits/788ed40861b2fbbc7fcf384d33d77c533abf0345

server/broadcast.go
j0sh: Fixed in https://github.com/livepeer/go-livepeer/pull/1096/commits/b2c942db6a235b82a286d0a08b826a494c05e57a

server/broadcast_test.go
j0sh: Fixed in https://github.com/livepeer/go-livepeer/pull/1096/commits/788ed40861b2fbbc7fcf384d33d77c533abf0345

server/broadcast_test.go
j0sh: Fixed in https://github.com/livepeer/go-livepeer/pull/1096/commits/788ed40861b2fbbc7fcf384d33d77c533abf0345

j0sh:
* Updated PR to incorporate some lingering issues that were commented in the code 
* Addressed remaining PR feedback
* Still needs additional tests for epic classifier
* Still needs to distinguish retryable from fatal errors

There is a bug where the pixels are not checked if the error is retryable. Partial mitigation in https://github.com/livepeer/go-livepeer/pull/1096/commits/f061a0f801359209d2c80b653baf610c950ca505 , tests in https://github.com/livepeer/go-livepeer/pull/1096/commits/5724313a8ee39bb30c0164c723ebb606153a9bea . To really address this correctly, I think we need to distinguish retryable from fatal errors. (A fatal error, such as with pixel counts, would still lead to a retry, but the O is removed from the working set and its results do not have the possibility of being selected.)

j0sh:
Added unit tests for the epic classifier. Mostly happy with the state of the tests. There are a few code paths that haven't been exercised, as noted in the comments. The unexercised code paths are mostly around error handling for library functions such as `ioutil.WriteFile`. I'm OK with leaving the tests as-is unless there is a preference to complete those prior to merging. https://github.com/livepeer/go-livepeer/pull/1096/commits/57350e46c577b65d6a5a0d1cb68f06558c7c38a0

Fixes as a result of these tests in https://github.com/livepeer/go-livepeer/pull/1096/commits/f71b4974d2349b3f77068e20166a7f8b8962cc9b and https://github.com/livepeer/go-livepeer/pull/1096/commits/8c030374ecf1d3d839d00154dc681b00ea403027

server/broadcast.go
```diff
@@ -432,6 +448,26 @@ func transcodeSegment(cxn *rtmpConnection, seg *stream.HLSSegment, name string)
 		if dlErr != nil {
 			return dlErr
 		}
+
+		if verifier != nil {
+			err := verify(verifier, cxn, sess, seg, res, segURLs)
+			if err != nil {
+				return err
```
yondonfu, line 462: Can we log the error here? [1] I don't see a log statement to handle some of the errors that would be propagated from `verify()`.

[1] An alternative could be to log the error returned by [transcodeSegment()](https://github.com/livepeer/go-livepeer/pull/1096/files#diff-cecc27bfe2b6b19181ce443733dfba57R286)

server/broadcast_test.go
```diff
@@ -519,3 +547,290 @@ func TestVerifyPixels(t *testing.T) {
...
+	assert.Equal([]byte("attempt1"), mem.GetData(name))
+}
+
+func TestVerifier_HLSInsertion(t *testing.T) {
+	assert := assert.New(t)
+
+	// Ensure that the playlist has the correct URL after verification
+	// Following this sequence:
+	//   1. Verify seg1. Verification fails.
+	//   2. Verify seg2. Verification fails.
```
yondonfu, line 926: Should there be a step 3:

`3. Verify seg3. Verification fails`

since the test runs verification 3 times?

server/broadcast_test.go
```diff
@@ -519,3 +547,290 @@ func TestVerifyPixels(t *testing.T) {
...
+	//   3. Max retries hit, assume false positives
+	//   4. Return seg1 as highest scoring result
```
yondonfu: Should this step and the next step say `seg2` since the test asserts that the second segment is used since it has the highest score?

server/broadcast_test.go
```diff
@@ -519,3 +547,290 @@ func TestVerifyPixels(t *testing.T) {
...
+		}
+
+		buf, err := proto.Marshal(&net.TranscodeResult{
+			Result: &net.TranscodeResult_Data{
+				Data: &net.TranscodeData{Segments: segData},
+			},
+		})
+		assert.Nil(err, fmt.Sprintf("Could not marshal results for %s", url))
+		ts, mux := stubTLSServer()
+		//defer ts.Close()
```
yondonfu: Perhaps we can return `ts` from this function and then call `Close()` on each instance of `httptest.Server` that is created via a call to `genBcastSess()`. Something like:

```
genBcastSess := func(url string) (*BroadcastSession, *httptest.Server) { ... }

sess1, ts1 := genBcastSess(...)
sess2, ts2 := genBcastSess(...)
sess3, ts3 := genBcastSess(...)

defer func() {
    ts1.Close()
    ts2.Close()
    ts3.Close()
}()
```

yondonfu (changes requested):

yondonfu:
> To really address this correctly, I think we need to distinguish retryable from fatal errors. (A fatal error, such as with pixel counts, would still lead to a retry, but the O is removed from the working set and its results do not have the possibility of being selected.)

And right now we're removing the O from the working set for both verifier (i.e. tampered) and fatal errors (i.e. pixel mismatch, audio mismatch) (which are all categorized as `Retryable` at the moment) i.e. `ErrPixelMismatch`, `ErrTampered`, `ErrAudioMismatch`? That seems ok to me for now.

If we distinguish verifier (i.e. `ErrTampered`) from fatal (i.e. `ErrPixelMismatch` and `ErrAudioMismatch`) separately from this PR then we could continue to always remove O from the working set for both types of errors (since we want to make sure not to retry with the same O), but for fatal errors we could apply a more severe penalty based on a [ban policy](https://github.com/livepeer/go-livepeer/issues/1100#issuecomment-558846820).

> I'm OK with leaving the tests as-is

👍

server/broadcast.go
j0sh: Fixed in https://github.com/livepeer/go-livepeer/pull/1096/commits/33856c9130973b277297d0c1032550bebc83d20c

server/broadcast_test.go
j0sh: Fixed in https://github.com/livepeer/go-livepeer/pull/1096/commits/a908c149a95a16d4add837e2a3fe4bb6b0749716

server/broadcast_test.go
j0sh: Fixed in https://github.com/livepeer/go-livepeer/pull/1096/commits/a908c149a95a16d4add837e2a3fe4bb6b0749716

server/broadcast_test.go
j0sh: Took different approach in https://github.com/livepeer/go-livepeer/pull/1096/commits/a908c149a95a16d4add837e2a3fe4bb6b0749716 . It's a bit of a kludge but the API is more satisfying. Do you still prefer the explicit approach?

Do you know how important it is to call `ts.Close()` for httptest servers?  It seems designed to incorporate server shutdowns into testing rather than any critical bookkeeping [1], but I'm having a hard time finding more detailed explanations of its intended usage.

[1] https://golang.org/src/net/http/httptest/server.go?s=5205:5229#L180

yondonfu: > It's a bit of a kludge but the API is more satisfying. Do you still prefer the explicit approach?

I'm fine with the approach in a908c14 - just felt like it would be good practice to be in the habit of closing/shutting things down when they're no longer used even in tests.

> Do you know how important it is to call ts.Close() for httptest servers? It seems designed to incorporate server shutdowns into testing rather than any critical bookkeeping

Not sure exactly how important it is - the [docs for NewServer()](https://golang.org/pkg/net/http/httptest/#NewServer) do mention that the caller should call Close() when finished. Perhaps its just to make sure that there are no idle connections lying around in tests.

yondonfu (commented):
Changes look good to me - let's rebase to clean up the commits and resolve the merge conflict?

cmd/livepeer/livepeer.go
```diff
@@ -663,6 +667,21 @@ func main() {
 		if server.AuthWebhookURL, err = getAuthWebhookURL(*authWebhookURL); err != nil {
 			glog.Fatal("Error setting auth webhook URL ", err)
 		}
+
+		// Set up verifier
+		if *verifierAddr != "" {
```
kyriediculous: For other external URLs entered as flags we prefix the URL with `https://` if it doesn't have a http protocol prefix, do you think it would be a good idea to add it here as well, just to have a uniform approach. Otherwise it might lead to confusion when some flags require the prefix and some don't.

j0sh: This probably needs to be renamed to `verifierUrl` or something [1], including the CLI flag. Most of our other `*Addr` flags technically only take a `host:port` combination, but this is more of a full URL, including the scheme and path such as `http://host:port/path/to/verifier`

[1] For our other external callback URLs (`orchWebhookUrl` and `authWebhookUrl`), we actually fail out if the prefix isn't http/https, rather than trying to correct it. Probably should do similar here - in fact, the below could use some refactoring.

https://github.com/livepeer/go-livepeer/blob/7f441276ac75c235e6578157e462a7ec5d82e275/cmd/livepeer/livepeer.go#L772-L800

kyriediculous: failing when the prefix isn't included sounds good to me !

j0sh: Made the following changes [1]

* Refactored to share webhook URL parsing code (should be no functional changes) 8e9062d393ad69b38b232e3e54179540d992f022

* Updated to use `-verifierUrl` as the flag name, following the scheme (and capitalization 😦) of our external webhook URLs in e62923951858d966623c3e855055039535b18f6d

* Additional CLI tests in b5665a81609ee648de6796680acd76506b1faa6e

[1] Went ahead and fixed up the commits locally rather than pushing a separate fixup commit for review since the timeline here is tight and there would have been merge conflicts due to a few commits changing the same area of code in `livepeer.go`

j0sh:
Rebased on top of master, collapsed all the fixups. There were a number of merge conflicts and build-breaking changes. Most of the commits should still compile OK on their own, but I didn't test every single one of them 🤞

README.md
```diff
@@ -128,6 +128,14 @@ A standalone transcoder can be run which connects to a remote orchestrator. The
 
 GPU transcoding on NVIDIA is supported; see the [GPU documentation](doc/gpu.md) for usage details.
 
+### Verification
+
+Experimental verification of video using the Epic Labs classifier can be enabled with the `-verifierAddr` flag. Pass in the address of the verifier API:
```
yondonfu: `-verifierAddr` -> `-verifierUrl`?

README.md
```diff
@@ -128,6 +128,14 @@ A standalone transcoder can be run which connects to a remote orchestrator. The
...
+
+- `livepeer -broadcaster -verifierAddr http://localhost:5000/verify`
```
yondonfu: Realized that this command would be insufficient for a user to enable verification at the moment since the broadcaster would either need to also use an external storage bucket or provide the `-verifierPath` flag. Should we add a note about that here? Since livepeer/docs#50 adds a verification section to the RTD we can probably go light on details here.

yondonfu (commented):
Left two minor comments about the verification section in the README. Feel free to address them & fixup the relevant commits locally without pushing up separate fixup commits.

README.md
j0sh: Fixed in 796ba8080309dc3e98a955daccbe46fd2f96845c

README.md
j0sh: Fixed in 796ba8080309dc3e98a955daccbe46fd2f96845c

j0sh:
For some reason the verifier / epic tests fell out of git while rebasing, so re-committed those and re-pushed (luckily does not look there was any data loss).

Here's a demonstration of the problem around retryable errors and pixel counts. Basically, what happens is:

1. O returns a legitimate result with a high score
2. The same results have an incorrect pixel count
3. Verifier detects both the high score and the pixel mismatch
4. Because pixel mismatch errors are marked as retryable, the O with the high score may have its results selected even if its reported pixel count is incorrect.

This is problematic because relying on O's self-reported results could lead B to over-count the pixels that were actually processed, resulting in higher effective fees. If the pixel counts don't match, then:

1. The results should not be eligible for selection
2. B should retry the transcoding

I've been referring to errors that should follow these two behaviors "fatal errors" but I think we need a better term for that. Non-fatal errors (ones that are simply retryable) should also re-add any evicted Os back into the pool, especially if we attempt to distinguish between false negatives and true positives via retries (false negatives re-add, true positives should stay evicted and/or be further penalized with a ban policy.). However, re-adding evicted Os isn't strictly needed to resolve the current issue around the fatal/retryable split.

```diff
diff --git a/verification/verify_test.go b/verification/verify_test.go
index b506e713..2d02ce95 100644
--- a/verification/verify_test.go
+++ b/verification/verify_test.go
@@ -115,6 +115,11 @@ func TestVerify(t *testing.T) {
        res, err = sv.Verify(&Params{ManifestID: "stu", Results: data})
        assert.Equal(ErrPixelsAbsent, err)
        assert.Equal("mno", string(res.ManifestID)) // Still return best result
+       // Higher score and incorrect pixel score should prioritize pixel count
+       verifier.results = &Results{Score: 20.0, Pixels: []int64{123, 123}}
+       res, err = sv.Verify(&Params{ManifestID: "vwx", Results: data})
+       assert.Equal(ErrPixelMismatch, err)
+       assert.Equal("mno", string(res.ManifestID)) // This should not fail! But it does.
 ```

cmd/livepeer/livepeer.go
kyriediculous: Thanks ! Changes look good 👍

yondonfu (approved):
LGTM - as discussed in Discord we can address the pixel counting + verification issue in a separate PR.

While this is still experimental and the verifier needs improvement, excited to see this initial integration working! 🎉

PR #1097, opened 2019-09-20
Makefile: `go build -mod=readonly` to avoid dirty git state

mk-livepeer:
This is an attempt to prevent the "dirty git state" that @iameli points out in #1095


j0sh:
This looks good to me, but I don't know much about `go mod`.  The CircleCI failure seems related though, can you look into that?

iameli:
Didn't seem to help, but all these changes look good!

PR #1099, opened 2019-09-20
pm: Add README

yondonfu:

This PR adds:

- A brief README for the `pm` package that references the full PM protocol specification and also highlights a few things that the package implements to ensure fair winning ticket selection.
- A payment doc in `doc/payments.md` that includes links to resources for the PM protocol as well as notes on fee estimation

See above.

N/A

**Checklist:**

### DIFF
--- a/pm/README.md
+++ b/pm/README.md
@@ -1,0 +1,33 @@
+# Probabilistic Micropayments (PM) Client
+
+The `pm` package contains a client implementation of the [Livepeer probabilistic micropayment protocol](https://github.com/livepeer/wiki/blob/master/spec/streamflow/pm.md). The main areas of interest are:
+
+- A `Ticket` type which defines the ticket format that both senders and recipients accept.
+- A `Sender` type which is responsible for creating tickets.
+- A `Recipient` type which is responsible for validating/receiving tickets and redeeming winning tickets using the `TicketBroker` interface (implemented by an Ethereum smart contract). This type will be able to translate a `Ticket` type into the correct format expected by the `TicketBroker`
+
+In practice, the package user will need to define a communication channel (i.e. HTTP connection) and a network message format (i.e. protocol buffer) that peers will use to send and receive tickets. The sending peer will use the `Sender` type and the receiving peer will use the `Recipient` type.
+
+## High Level Flow
+
+![flow diagram](assets/diagram.png)
+
+## RNG
+
+A `Sender` and a `Recipient` execute a collaborative commit-reveal protocol in order to generate the random number used to determine if a ticket won or not. The random number construction is:
+
+```
+keccak256(abi.encodePacked(senderSig, recipientRand))
+```
+
+The `abi.encodePacked()` Solidity built-in is used to match the behavior of the contract implementing the `TicketBroker` interface. `senderSig` is a signature produced by `Sender` over the hash of the ticket and `recipientRand` is the pre-image of the `recipientRandHash` that is included in the ticket.
+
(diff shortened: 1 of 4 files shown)


j0sh:
The first part looks pretty good to me. The RNG part feels a little ad hoc / incomplete without some more context around PM. Even though there is a link to the full PM spec, it might not be immediately clear to a new reader where the RNG part fits in and the reasoning for certain behaviors.

Eg, from a quick glance, the writeup mentions that winning `recipientRand` values are only persisted in memory. A sharp reader might ask why "in memory" is a sufficient defense, and the reader needs to connect a few dots themselves within the full spec (or via reading the code) to figure out it's because `O.secret` is generated fresh at each start-up time.

Not really sure yet how/where to improve this without a closer read of the PM spec.

pm/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+A `Sender` and a `Recipient` execute a collaborative commit-reveal protocol in order to generate the random number used to determine if a ticket won or not. The random number construction is:
+
+```
+keccak256(abi.encodePacked(senderSig, recipientRand))
+```
+
+The `abi.encodePacked()` Solidity built-in is used to match the behavior of the contract implementing the `TicketBroker` interface. `senderSig` is a signature produced by `Sender` over the hash of the ticket and `recipientRand` is the pre-image of the `recipientRandHash` that is included in the ticket.
+
+Before a `Sender` can start creating and sending tickets, it must first fetch ticket parameters from a `Recipient`. When generating ticket parameters, a `Recipient` will generate a random number `recipientRand`, compute the hash of `recipientRand` as `recipientRandHash` and include the hash with the ticket parameters. Once a `Sender` has a set of ticket parameters with a `recipientRandHash`, the `Sender` will use a monontonically increasing `senderNonce` with each new ticket that it creates.
```
kyriediculous, line 25: It might be a good idea to provide some extra context around `TicketParams` here , perhaps we can use the commented protobuf definitions to provide more clarity here?

pm/README.md
```diff
@@ -0,0 +1,37 @@
+# Probabilistic Micropayments (PM) Client
+
+The `pm` package contains a client implementation of the [Livepeer PM protocol](https://github.com/livepeer/wiki/blob/master/spec/streamflow/pm.md). The main areas of interest are:
+
+- A `Ticket` type which defines the ticket format that both senders and recipients accept.
```
kyriediculous, line 5: It might be a good idea to provide some extra context around this type as its fieldnames are used extensively in the RNG part of the README. WDYT?

thomshutt:
@yondonfu Is this documentation still worth getting merged or can we close the PR?

yondonfu:
Reviving this PR! Updated the `pm` package docs with some tweaks based on recent changes (i.e. https://github.com/livepeer/go-livepeer/pull/1383 and https://github.com/livepeer/go-livepeer/pull/2710) as well as an additional payments doc with links to PM resources and some notes on fee estimation. These docs are by no means comprehensive and can certainly be improved, but think it should be a 0 -> 1 upgrade that will be a good start for now!

thomshutt (approved):
This is great, thanks Yondon! I think you also win longest time between PR opening and getting merged 🏆

yondonfu:
Made a few small changes:

- Moved diagram into `pm/assets` folder and included a text file with the sequence diagram in UML so that it can be updated later if needed
- Moved the diagram PNG file into git LFS

The above changes align with the convention used for assets in docs seen in `doc/assets`. They also seem to fix a weird dirty git diff error I was encountering in CI builds...

PR #1105, opened 2019-09-30
Expose number of transcoders as Prometheus metric

darkdarkdragon:
Expose number of transcoders currently connected to orchestrator
as Prometheus metric

- 
- 
- 

Manually

**Checklist:**

### DIFF
--- a/core/orchestrator.go
+++ b/core/orchestrator.go
@@ -753,14 +753,27 @@ func (rtm *RemoteTranscoderManager) Manage(stream net.Transcoder_RegisterTransco
 	rtm.liveTranscoders[transcoder.stream] = transcoder
 	rtm.remoteTranscoders = append(rtm.remoteTranscoders, transcoder)
 	sort.Sort(byLoadFactor(rtm.remoteTranscoders))
+	var totalLoad, totalCapacity, liveTranscodersNum int
+	if monitor.Enabled {
+		totalLoad, totalCapacity, liveTranscodersNum = rtm.totalLoadAndCapacity()
+	}
 	rtm.RTmutex.Unlock()
+	if monitor.Enabled {
+		monitor.SetTranscodersNumberAndLoad(totalLoad, totalCapacity, liveTranscodersNum)
+	}
 
 	<-transcoder.eof
 	glog.Infof("Got transcoder=%s eof, removing from live transcoders map", from)
 
 	rtm.RTmutex.Lock()
 	delete(rtm.liveTranscoders, transcoder.stream)
+	if monitor.Enabled {
+		totalLoad, totalCapacity, liveTranscodersNum = rtm.totalLoadAndCapacity()
+	}
 	rtm.RTmutex.Unlock()
+	if monitor.Enabled {
+		monitor.SetTranscodersNumberAndLoad(totalLoad, totalCapacity, liveTranscodersNum)
+	}
 }
 
 func (rtm *RemoteTranscoderManager) selectTranscoder() *RemoteTranscoder {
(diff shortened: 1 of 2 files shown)


core/orchestrator.go
```diff
@@ -803,6 +811,14 @@ func (rtm *RemoteTranscoderManager) completeTranscoders(trans *RemoteTranscoder)
 	sort.Sort(byLoadFactor(rtm.remoteTranscoders))
 }
 
+func (rtm *RemoteTranscoderManager) totalCapacity() int {
```
j0sh: Maybe add a comment that the caller of this function needs to hold the RTmutex

core/orchestrator.go
```diff
@@ -753,13 +753,21 @@ func (rtm *RemoteTranscoderManager) Manage(stream net.Transcoder_RegisterTransco
...
 	rtm.RTmutex.Unlock()
 
 	<-transcoder.eof
 	glog.Infof("Got transcoder=%s eof, removing from live transcoders map", from)
 
 	rtm.RTmutex.Lock()
 	delete(rtm.liveTranscoders, transcoder.stream)
+	if monitor.Enabled {
+		monitor.SetTranscodersNumber(int64(len(rtm.liveTranscoders)))
+		monitor.SetTranscodersCapacity(int64(rtm.totalCapacity()))
```
j0sh: Is this "expensive" ? If so, would it make more sense to store these in temp variables and add to Prometheus outside the lock?

j0sh: Would it also be useful to expose the current load for the transcoder? Or even the load as a percentage of capacity? That seems like good information to track to get a feel for the overall work distribution on the O.

j0sh (changes requested):
Looking good, just asking for some small changes around comments and a few questions.

core/orchestrator.go
darkdarkdragon: done

core/orchestrator.go
darkdarkdragon: Don't know, but good idea - moved it out.

darkdarkdragon: Good idea. Exposed 'load' metric - just total value of transcoders `load` field. Percentage can be easily calculated in Prometheus itself.

darkdarkdragon:
@j0sh made requested changes

core/orchestrator.go
```diff
@@ -753,14 +753,33 @@ func (rtm *RemoteTranscoderManager) Manage(stream net.Transcoder_RegisterTransco
...
+	var totalLoad, totalCapacity, liveTranscodersNum int
+	if monitor.Enabled {
+		liveTranscodersNum = len(rtm.liveTranscoders)
+		totalLoad, totalCapacity = rtm.totalLoadAndCapacity()
+	}
 	rtm.RTmutex.Unlock()
+	if monitor.Enabled {
+		monitor.SetTranscodersNumber(int64(liveTranscodersNum))
+		monitor.SetTranscodersCapacity(int64(totalCapacity))
+		monitor.SetTranscodersLoad(int64(totalLoad))
```
j0sh: This is starting to feel verbose - would it make sense to have a single function in `monitor` that takes all 3 parameters?

monitor/census.go
```diff
@@ -725,6 +757,24 @@ func TranscodeTry(nonce, seqNo uint64) {
 	}
 }
 
+func SetTranscodersNumber(transcoders int64) {
```
j0sh: Would it be a little cleaner at the call site to take an `int` type as a parameter, and cast to int64 when calling `Record`? (Same goes for the other functions that were added here)

darkdarkdragon: done

core/orchestrator.go
darkdarkdragon: fixed

darkdarkdragon:
@j0sh done changes

j0sh (approved):
LGTM after fixing up all the commits into one

PR #1114, opened 2019-10-04
cmd, eth, server, pm: Streamflow client compatibility

### LINKED ISSUE #1089: Remove frozen reserve handling and recalculate reserve alloc at start of round

Once livepeer/protocol#319 is merged, there will no longer be a concept of a "frozen reserve". The PM client currently has a few pieces of logic that are based off the concept of a "frozen reserve". These areas of the codebase will need to be updated.

- [ ] `eth.SenderWatcher` should refresh the cached reserve for a sender when a new round is initialized in order to update the total funds available for claiming if the sender's deposit depleted (livepeer/protocol#319 will update the reserve funds to be "reset" at the start of a round - active set members can claim up to the total funds available at the start of the round divided by the size of the active set during the round)
- [ ] Remove pieces of code that depend on a frozen reserve. Examples [here](https://github.com/livepeer/go-livepeer/blob/master/eth/watchers/senderwatcher.go#L211) and [here](https://github.com/livepeer/go-livepeer/blob/master/eth/watchers/senderwatcher.go#L164)

Note: This will require up-to-date contract Go bindings.

### LINKED ISSUE #1090: Update TicketBroker unlocking code to use rounds

Once livepeer/protocol#320 is merged, the unlock period for a sender will measured in rounds. There are a few areas of the codebase that assume that the unlock period is measured in blocks. These areas of the codebase will need to be updated.

- [ ] Update `eth.SenderWatcher` to record a withdraw round instead of a withdraw block when an `Unlock()` event is emitted for at tracked sender
- [ ] Update the CLI (both implementation and user facing language) to use a round based unlock period

Note: This will require up-to-date contract Go bindings.

kyriediculous:
This PR addresses some of the compatibility issues the current go-livepeer client has with the new. `protocol/streamflow` branch after regenerating the contract bindings. 

For a detailed list of the updates see _specific upgrades_ 

- e397e0c: Regenerate the contract bindings using `abigen`. Created a wrapper arround the TicketBroker session because `abigen` does not generate client side structs for all contract types. https://github.com/ethereum/go-ethereum/issues/20148

- 3184658: Removes any reference to/usage of `pricePerSegment`, `pendingRewardCut` and `pendingFeeShare` and update the CLI implementation and user facing language. 

- 63fab12: Removes the usage of `NumActiveTranscoders` as the generated function from the `public` variable `numActiveTranscoders` on the contracts is now deprecated. Instead we'll use `GetTranscoderPoolMaxSize`. 

- 7174eb1: update `isActiveTranscoder` usage to reflect its implementation on the `protocol/streamflow` branch, meaning it doesn't take a second `round` argument anymore. Instead `IsActiveTranscoder` will return the active status for a transcoder for the _current_ round

- 34d7c28: Change the definition of `pm.SenderInfo` to  remove `ReserveState` and `ThawRound` references

- 34d7c28: Change the definition of `pm.SenderInfo` to use `WithdrawRound` instead of `WithdrawBlock`

- 34d7c28: Change the definition of `pm.SenderInfo` to use a new `pm.ReserveInfo` type instead of a `SenderInfo.Reserve *big.Int` type. `pm.ReserveInfo` holds `ReserveInfo.FundsRemaining` and `ReserveInfo.ClaimedInCurrentRound` from which a sender's total  available reserve for a round is calculated by adding these two values. 

- 2c9e40f: Adds a dependency to the `RoundsWatcher` to `SenderWatcher` through an `EventWatcher` interface which exposes a single `Subscribe` method. This allows the `SenderWatcher` to describe to `NewRound` events. This is needed because since removing `ReserveState` , amounts claimed from a reserve are reset on a per-round basis rather than when the `reserveNonce` increases due to funding a thawed reserve. 

- c82f82d Removes `LivepeerVerifier` references in the `eth` package since Streamflow will not contain on-chain verification components in it's initial release. The contract itself has also been deleted in the Streamflow branch for the protocol repo.

**TODO: Look into the subscription mechanism used for the blockwatch package.**

Adjusted & ran unit tests

Fixes #1089 
Fixes #1090

**Checklist:**


kyriediculous:
4f0febe refactors the subscription mechanism for `RoundsWatcher` to use the [go-ethereum event package]( https://godoc.org/github.com/ethereum/go-ethereum/event#Feed)

This adds two fields to the `RoundsWatcher` struct
- `subScope`
> SubscriptionScope provides a facility to unsubscribe multiple subscriptions at once.

Subscriptions created and tracked through `subScope` are tied to the lifetime of `RoundsWatcher` (see c6b42186)

- `subFeed`
> Feed implements one-to-many subscriptions where the carrier of events is a channel. Values sent to a Feed are delivered to all subscribed channels simultaneously.
Feeds can only be used with a single type. The type is determined by the first Send or Subscribe operation. Subsequent calls to these methods panic if the type does not match.
The zero value is ready to use.

The sink channel is of type `event.Log` rather than a slice of `blockwatch.Event` because we only care about the NewRound logs emitted by the `RoundsWatcher`, thus the feed can only be used with type `event.Log` for all subscriptions.

cmd/livepeer_cli/wizard_ticketbroker.go
```diff
@@ -116,9 +96,9 @@ func (w *wizard) unlock() {
 		return
 	}
 
-	projWithdrawBlock := new(big.Int).Add(blk, params.UnlockPeriod)
+	projWithdrawRound := new(big.Int).Add(blk, params.UnlockPeriod)
```
yondonfu: We should be adding `params.UnlockPeriod` to the current round and not the current block here.

cmd/livepeer_cli/wizard_ticketbroker.go
```diff
@@ -116,9 +96,9 @@ func (w *wizard) unlock() {
...
 
-	fmt.Printf("If you initiate the unlock period now, you will be able to withdraw at block %v\n", projWithdrawBlock)
+	fmt.Printf("If you initiate the unlock period now, you will be able to withdraw at block %v\n", projWithdrawRound)
```
yondonfu: "withdraw at block" -> "withdraw at round"

cmd/livepeer_cli/wizard_ticketbroker.go
```diff
@@ -143,9 +123,9 @@ func (w *wizard) cancelUnlock() {
 	}
 
 	fmt.Printf("Current Deposit: %v\n", eth.FormatUnits(sender.Deposit, "ETH"))
-	fmt.Printf("Current Reserve: %v\n", eth.FormatUnits(sender.Reserve, "ETH"))
+	fmt.Printf("Current Reserve: %v\n", eth.FormatUnits(sender.Reserve.FundsRemaining, "ETH"))
 	fmt.Printf("Current Block: %v\n", blk)
-	fmt.Printf("Withdraw Block: %v\n", sender.WithdrawBlock)
+	fmt.Printf("Withdraw Block: %v\n", sender.WithdrawRound)
```
yondonfu: "Withdraw Block" -> "Withdraw Round". Same comment applies to other areas of the CLI code that still use the "Withdraw Block" terminology.

cmd/livepeer_cli/wizard_ticketbroker.go
```diff
@@ -90,17 +80,7 @@ func (w *wizard) unlock() {
...
-
-		fmt.Printf("Current Round: %v\n", currRound)
-		fmt.Printf("Thaw Round: %v\n", sender.ThawRound)
-
-		fmt.Printf("Cannot deposit because sender's reserve is frozen and not yet thawed")
-		return
-	}
+	fmt.Printf("Current Reserve: %v\n", eth.FormatUnits(sender.Reserve.FundsRemaining, "ETH"))
 
 	fmt.Printf("Current Block: %v\n", blk)
```
yondonfu: Since the unlock period is denominated in rounds instead of blocks, I think it will be clearer if we show the user the current round instead of the current block whenever the CLI is used to unlock/cancel unlock/withdraw.

cmd/livepeer_cli/wizard_ticketbroker.go
```diff
@@ -283,12 +253,12 @@ func printSenderStatus(msg string, status SenderStatus) {
...
 func senderStatus(sender pm.SenderInfo, currentBlock *big.Int) SenderStatus {
-	if sender.Deposit.Cmp(big.NewInt(0)) == 0 && sender.Reserve.Cmp(big.NewInt(0)) == 0 {
+	if sender.Deposit.Cmp(big.NewInt(0)) == 0 && sender.Reserve.FundsRemaining.Cmp(big.NewInt(0)) == 0 {
 		return Empty
 	}
 
-	if sender.WithdrawBlock.Cmp(big.NewInt(0)) > 0 {
-		if sender.WithdrawBlock.Cmp(currentBlock) <= 0 {
+	if sender.WithdrawRound.Cmp(big.NewInt(0)) > 0 {
+		if sender.WithdrawRound.Cmp(currentBlock) <= 0 {
```
yondonfu: Let's change the `currentBlock` parameter name to `currentRound` and make sure to pass in the right value when `senderStatus()` is called.

cmd/livepeer_cli/wizard_ticketbroker_test.go
```diff
@@ -8,10 +8,13 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-func createSender(deposit *big.Int, reserve *big.Int, withdrawBlock *big.Int) (sender pm.SenderInfo) {
+func createSender(deposit *big.Int, reserve *big.Int, WithdrawRound *big.Int) (sender pm.SenderInfo) {
```
yondonfu: "WithdrawRound" -> "withdrawRound"

cmd/livepeer_cli/wizard_ticketbroker_test.go
```diff
@@ -24,17 +27,17 @@ func TestSenderStatus(t *testing.T) {
...
 	assert.Equal(Empty, ss)
 
-	// Test Empty, but withdrawBlock > 0
+	// Test Empty, but WithdrawRound > 0
 	s = createSender(big.NewInt(0), big.NewInt(0), big.NewInt(5))
 	ss = senderStatus(s, big.NewInt(0))
 	assert.Equal(Empty, ss)
 
-	// Test Unlocked when withdrawBlock = currentBlock
+	// Test Unlocked when WithdrawRound = currentBlock
```
yondonfu: Comments need to be updated to say `currentRound` instead of `currentBlock`.

eth/client_ticketbroker.go
```diff
@@ -62,17 +62,18 @@ func (c *client) RedeemWinningTicket(ticket *pm.Ticket, sig []byte, recipientRan
 
 // GetSenderInfo returns the info for a sender
 func (c *client) GetSenderInfo(addr ethcommon.Address) (*pm.SenderInfo, error) {
-	info, err := c.TicketBrokerSession.GetSenderInfo(addr)
+	info, err := c.WrappedTicketBrokerSession.GetSenderInfo(addr)
```
yondonfu: I realized that since we already wrap `TicketBrokerSession.GetSenderInfo()` here we can just move the logic from `WrappedTicketBrokerSession.GetSenderInfo()` here and remove `WrappedTicketBrokerSession`. This change would allow us to avoid creating an additional struct that would be removed in the future anyway once `abigen` is capable of properly generating contract bindings for solc >= 0.5.11 contracts.

eth/watchers/roundswatcher.go
```diff
@@ -136,6 +149,9 @@ func (rw *RoundsWatcher) handleLog(log types.Log) error {
 		return fmt.Errorf("eventName is not NewRound")
 	}
 
+	fmt.Println(log)
```
yondonfu: Can be removed?

eth/watchers/roundswatcher.go
```diff
@@ -136,6 +149,9 @@ func (rw *RoundsWatcher) handleLog(log types.Log) error {
...
+	rw.subFeed.Send(log)
```
yondonfu: Should we only be sending to the feed if the log decodes to a `NewRound` event?

eth/watchers/roundswatcher.go
```diff
@@ -107,9 +111,18 @@ func (rw *RoundsWatcher) Watch() error {
 	}
 }
 
+// Subscribe allows one to subscribe to the block events emitted by the Watcher.
```
yondonfu, line 114: "round events emitted by the RoundsWatcher"?

eth/watchers/senderwatcher_test.go
```diff
@@ -397,19 +452,18 @@ func TestWinningTicketTransferEvent(t *testing.T) {
 	info, ok = sw.senders[stubSender]
 	assert.True(ok)
 	assert.Equal(info.Deposit, startDeposit)
-	assert.Equal(info.Reserve, startReserve)
+	assert.Equal(new(big.Int).Add(info.Reserve.FundsRemaining, info.Reserve.ClaimedInCurrentRound), startReserve)
```
yondonfu: Can this just be `assert.Zero(startReserve.Cmp(info.Reserve.FundsRemaining))`?

eth/watchers/senderwatcher_test.go
```diff
@@ -537,43 +610,67 @@ func TestUnlockEvent(t *testing.T) {
...
+
+	// Use a seperate stubWatcher here for RoundsWatcher
+	// else the sink channel on the stubBlockWatcher will be overwritten
+	// when calling senderWatcher.Subscribe() and the RoundsWatcher will never receive
+	// Round events to forward to the SenderWatcher
+	watcherForRounds := &stubBlockWatcher{}
 	watcher := &stubBlockWatcher{}
-	sw, err := NewSenderWatcher(stubTicketBrokerAddr, watcher, lpEth)
+
+	rw, err := NewRoundsWatcher(stubRoundsManagerAddr, watcherForRounds, lpEth)
```
yondonfu: Since `SenderWatcher` accepts the `EventWatcher` interface we can slim down + simplify these tests by passing in a stub type that implements `EventWatcher` when creating `SenderWatcher`.

yondonfu (changes requested):

eth/watchers/senderwatcher_test.go
kyriediculous: I considered this but at the time I felt like there was a little bit of added value here in turning this into more of an integration test. 

Would you like me to change this into a pure unit test to be more in line with the other unit test formats used in the remainder of the repo?

yondonfu: Let's go with unit tests here to remain consistent with the type of testing used in this file.

eth/watchers/roundswatcher.go
kyriediculous: I think it makes sense to treat the event the log is sent over to the SenderWatcher as an 'event channel' rather than a 'message channel' even though this is not fully the case. 

By this I mean that we do not use the crucial value pertaining to the emitted event, which would be the round number. We only check whether the event was removed or not to branch into different logic, receiving a message on the channel itself would indicate a new round event already. This seems to be the most effective because otherwise we would be sending _every_  event through the channel and would still have to check the signature in `SenderWatcher` and we might as well use a second `BlockWatcher` using the `RoundsManager` ABI , rather than the `RoundsWatcher` if that were the case.

yondonfu: To clarify, I'm not saying that we shouldn't send the log into the feed. I was suggesting that we send the log to the feed after checking if the log successfully decodes into a `NewRound` event. We already check that the log's first topic hash maps to the `NewRound` event name so this would just be an extra sanity check to make sure that the log actually decodes into the Go struct and that the included data is not malformed.

kyriediculous: Ahh yes sorry, I see what you mean :) !

eth/client_ticketbroker.go
kyriediculous: Not exactly possible because the higher level `sess.Contract.TicketBrokerCaller.contract.Call` method lives on an unexposed object. 

But I can do a low level call instead I think, just a bit longer.

kyriediculous:
removed `ticketBrokerWrapper` in 1cf132e and use a lower level implementation of contract calls in  `client_ticketBroker.GetSenderInfo`

Changed `RoundsWatcher` for a `stubRoundsWatcher` in the `SenderWatcher` tests in 91ea296

CLI wizard & other small fixups in 0f33adc

kyriediculous:
Fixed a small issue whereby the test harness would panic when calling `GetSenderInfo` since the `context` used in `client_ticketBroker.GetSenderInfo` was nil because we use it from `TicketBrokerSession.CallOpts`. 

If `TicketBrokerSession.CallOpts.Context` is nil we will use a new `Context.Background()` instead.

cmd/livepeer_cli/wizard_stats.go
```diff
@@ -76,9 +75,13 @@ func (w *wizard) stats(showOrchestrator bool) {
...
 
-	currentRound := w.currentRound()
+	currentRound, err := w.currentRound()
+	if err != nil {
+		glog.Errorf("Error getting current round: %v", err)
+		return
+	}
 
-	fmt.Printf("CURRENT ROUND: %v\n", currentRound)
+	fmt.Printf("CURRENT ROUND: %v\n", currentRound.Int64())
```
yondonfu: Don't think we need to convert to an `int64` here.

cmd/livepeer_cli/wizard_rounds.go
```diff
@@ -1,11 +1,31 @@
...
 	"fmt"
+	"io/ioutil"
+	"math/big"
+	"net/http"
 )
 
-func (w *wizard) currentRound() string {
-	return httpGet(fmt.Sprintf("http://%v:%v/currentRound", w.host, w.httpPort))
+func (w *wizard) currentRound() (*big.Int, error) {
+	resp, err := http.Get(fmt.Sprintf("http://%v:%v/currentRound", w.host, w.httpPort))
```
yondonfu, line 12: Eventually it will be nice to update the `httpGet()` helper to include all of this boilerplate and then it can return `([]byte, error)` instead of `string` leaving the conversion of the `[]byte` result up to the caller. Just noting this for the future, but no need to do this now.

eth/client_ticketbroker.go
```diff
@@ -62,17 +67,52 @@ func (c *client) RedeemWinningTicket(ticket *pm.Ticket, sig []byte, recipientRan
 
 // GetSenderInfo returns the info for a sender
 func (c *client) GetSenderInfo(addr ethcommon.Address) (*pm.SenderInfo, error) {
-	info, err := c.TicketBrokerSession.GetSenderInfo(addr)
+	info := new(struct {
```
yondonfu, line 68: We might be able to avoid this custom low level call altogether by generating the contract bindings using a version of `abigen` built from this branch https://github.com/ethereum/go-ethereum/pull/20179 which will use the `internalType` specified in a contract's ABI to generate Go structs with meaningful names (another benefit is that we won't end up with struct names like `Struct0` and `Struct1` in the contract binding).

yondonfu (changes requested):

eth/client_ticketbroker.go
kyriediculous: I'm hesitant to include any generated code by that branch in our own codebase for now, seeing as CI checks for that branch do not pass and is likely due for change requests. 

While it is surely a step in the right direction I think it might be good to be a little conservative here and address it in a seperate PR when that branch becomes merged. WDYT?

cmd/livepeer_cli/wizard_stats.go
kyriediculous: fixed in 7dc95c3

eth/client_ticketbroker.go
yondonfu: Agree that we can update the bindings in a separate PR.

I think we can slim down this implementation a bit by instantiating a `bind.BoundContract` type using the contract ABI. Here is a sketch (have not tested this):

```
// GetSenderInfo returns the info for a sender
func (c *client) GetSenderInfo(addr ethcommon.Address) (*pm.SenderInfo, error) {
	out := new(struct {
		Sender struct {
			Deposit       *big.Int
			WithdrawRound *big.Int
		}
		Reserve pm.ReserveInfo
	})

	abi, err := abi.JSON(strings.NewReader(contracts.TicketBrokerABI))
	if err != nil {
		return nil, err
	}

	contract := bind.NewBoundContract(c.ticketBrokerAddr, abi, c.backend, c.backend, c.backend)
	if err := contract.Call(&c.TicketBrokerSession.CallOpts, out, "getSenderInfo", addr); err != nil {
		return nil, err
	}

	return &pm.SenderInfo{
		Deposit:       out.Sender.Deposit,
		WithdrawRound: out.Sender.WithdrawRound,
		Reserve: &pm.ReserveInfo{
			FundsRemaining:        out.Reserve.FundsRemaining,
			ClaimedInCurrentRound: out.Reserve.ClaimedInCurrentRound,
		},
	}, nil
}
```

yondonfu (changes requested):

eth/client_ticketbroker.go
kyriediculous: fixed in 583a3b5

I forgot to drop go modules changes in the commit after updating to new LPMS version but I'll get those in the rebase. 

That being said I didn't run into any issues running this in the test-harness so should be g2g. 

The block header not found issue seems to be occuring on current `master` +  `pm` as well so something we can track in a separate issue (I'll create one?).

`I1021 18:52:22.967576       1 block_watcher.go:167] block header not found blockNumber=946`

yondonfu (changes requested):
Changes look good. Let's rebase!

kyriediculous:
rebase done !

yondonfu (approved):
LGTM 🚢

wohlner:

PR #1115, opened 2019-10-04
Update to latest LPMS

darkdarkdragon:
Update dependency to latest LPMS.
Also did `go mod tidy`

- 
- 
- 

**Checklist:**


j0sh:
The `go mod` stuff somehow feels *less* tidy now (I see a lot more repetition there) but I don't really understand much about how it's supposed to work. Will defer this review to others that understand it better - cc @mk-livepeer

j0sh:
One thing that's a bit concerning here is the number of updated packages. It's probably OK for now while we're doing heavy development / testing but we might want to set a policy about this at some point to minimize unanticipated ripple impacts from some package breaking underneath us.

mk-livepeer (approved):
most of the dependency churn is a result of either:

a) go1.13.1 subtle package management changes (such as the removal of the // indirect references)
or
b) dependency churn within lpms itself

I think it's safe for merge

PR #1124, opened 2019-10-18
Integrate new LPMS API and load balancer.

### LINKED ISSUE #1108: GPU Load Balancing
The current load balancing algorithm for GPUs is a simple, stateless round robin. For each segment:

```
gpu = gpus[i++ % len(gpus)`
```

With the LPMS bottleneck fix, transcode sessions become stateful. A naive approach to maintaining "sticky" sessions is this:

```
if sessions[id].nil ? sessions[id] = gpus[i++ % len(gpus)]
gpu = sessions[id]
```

Can we do better than the naive approach? Probably.

Some challenges here are:

* Segments for a given session may be sporadic
  * O / T may not necessarily be the “primary” for a given stream
* Session assignment on GPUs may grow unbalanced
  * Sessions come and go, some may be long-lived, others short-lived
* Workload varies per session
  * Different transcoding profiles have different performance characteristics, segment lengths, etc
* Hardware capabilities not necessarily known
  * Would be nice not to manually tune / configure

Need to determine:

* What might constitute an improved algorithm
* How to measure / prove its efficacy
* Success criteria for when to wrap this up

TODOs:

- [ ] Finish LPMS Fix #997 and https://github.com/livepeer/lpms/pull/141
- [ ] Test Design / Spec #1107
- [ ] Tooling Development / Improvements #1109
- [ ] Algorithm Refinement #1110
- [ ] Simulation https://github.com/livepeer/go-livepeer/issues/1112
- [ ] Test Execution 
- [ ] Analysis
- [ ] Iteration
- [ ] Reproducible Environment #1111

### LINKED ISSUE #1110: GPU LB : Algorithm Improvements
Can we do better than the following on a per-segment basis?

```
if sessions[id].nil ? sessions[id] = gpus[i++ % len(gpus)]
gpu = sessions[id]
```

j0sh:
**WIP** Tests still need some work but the main implementation should look fairly close to the final revision, subject to feedback.

Summary of the main idea of this PR:

Add a [`LoadBalancingTranscoder`](https://github.com/livepeer/go-livepeer/pull/1124/files#diff-da447e938cf8298a688acc7c4626255aR24-R34) implementation that implements the [`Transcoder`](https://github.com/livepeer/go-livepeer/pull/1124/files#diff-418dc657e12875665c93dffe96a0e464L20-L22) interface. This load balancer is in turn parameterized by a "sessionable" transcoder implementing the [`TranscoderSession`](https://github.com/livepeer/go-livepeer/pull/1124/files#diff-da447e938cf8298a688acc7c4626255aR17-R20) interface. (Happy to take suggestions for better / more dissimilar naming.)  For each new job, we search for the least loaded transcoder (as described [here](https://github.com/livepeer/go-livepeer/issues/1110#issuecomment-551332856)). Upon assigning a transcoder to a new job, we instantiate a transcoding session.

The structure of the transcoding session follows that of the orchestrator: a transcode loop is invoked in order to keep all session-related processing on the same thread, and the transcoder itself is invoked via channels [1]. The session terminates after one minute of inactivity, or a transcoding error.

The nvidia transcoder implementation is modified to be "session-able", following the API changes introduced in https://github.com/livepeer/lpms/pull/141 . See the  LPMS PR for a more detailed writeup of the API changes involved.

[1] Transcodes are "sessionable" now, and a single session can only process one segment at a time. LPMS itself enforces this exclusivity for session processing, but the goclient needs more synchronization around session creation and clean-up. Using a single-threaded loop + channels keeps those synchronization issues to a minimum. Started with a more straightforward approach of "just call the transcoder with a mutex", but this introduced a number of edge cases that could only be mitigated by fragile bookkeeping, or heavy locking. Keeping everything on the same thread, meditated by channels, was the most future-proof way I can think of handling this.

* Uses the new LPMS API for GPU transcoding. This retains a persistent session on the GPU for each stream.  TODO : Describe new API within LPMS PR
* Misc support changes required to support the new LPMS API, such as adding a `job` name parameter to the`Transcoder.Transcode` interface method.
* Incorporates a load balancing algorithm to select GPUs. The algorithm follows the "least loaded once" approach approach as described in https://github.com/livepeer/go-livepeer/issues/1110#issuecomment-551332856 

- Adds a 'job' parameter to Transcoder API
- Support the parameter for standalone T - involves protobuf modification / regeneration
- Update install_ffmpeg.sh to explicitly incorporate h264_cuvid decoder. LPMS fix needs this.
- Update `go.mod` and `go.sum` to incorporate new dependencies - LPMS and rapid
- Implement load balancing algorithm
- Incorporate load balancing algo into Nvidia transcoder and CLI program
- Added property-based testing in various parts. (Still WIP)

* Manual testing
* Unit testing, including property-based tests
* Extensive [stress testing](https://gist.github.com/j0sh/803403b1627945e0bdc2820ddc9e8947#file-rand-sh) (72+ hours)

Fixes https://github.com/livepeer/go-livepeer/issues/1110
Fixes https://github.com/livepeer/go-livepeer/issues/1108

**Checklist:**


j0sh:
Tests still need some work but the main implementation should look fairly close to the final revision, subject to feedback. Opening this up for review to get that feedback in while iterating on the tests.

core/orchestrator.go
```diff
@@ -31,6 +31,11 @@ import (
 
 var transcodeLoopTimeout = 1 * time.Minute
 
+// Gives us more control of "timeout" / cancellation behavior during testing
+var transcodeLoopContext = func() (context.Context, context.CancelFunc) {
+	return context.WithTimeout(context.Background(), transcodeLoopTimeout)
```
ericxtang, line 36: Would it be better if we can set the transcode loop timeout based on the length of the segment?

j0sh: Perhaps - is there a particular case that would be helped by doing it like this? Or a particular formulation you were thinking of?

Some additional notes around depending on the segment length (duration) in the orchestrator.

* Orchestrator currently doesn't know the length (duration) of the segment that it receives. We can add some [signaling](https://github.com/livepeer/go-livepeer/blob/909376fe6ee07144b0bed94f5adead98518d6697/net/lp_rpc.proto#L104-L126) for that data, but I don't believe we can outright trust that bit of data from the B.

* This timer only begins counting *after* a segment completes processing, so there's no possibility of cancelling the loop while a segment is in-flight. As soon as a new segment comes in, the timer is cancelled, and the segment itself can potentially take a very long time to process [1].

[1] If this sounds like a problem: it is! But that's not for this PR to solve. Related: https://github.com/livepeer/lpms/issues/129

* The default timeout is 1 minute, which should be plenty for live segmented streams, where we try to segment on 2-second boundaries. Other types of streams (eg, VOD) are not under consideration here, and the workflow for that would be quite different.

* If the transcode loop timeout is dependent on the segment length, setting a hard cap upper limit is still probably desirable to preserve O resources.

core/lb.go
```diff
@@ -0,0 +1,202 @@
...
+		if err != nil {
+			return nil, err
+		}
+	}
+	return session.Transcode(job, fname, profiles)
+}
+
+func (lb *LoadBalancingTranscoder) createSession(job string, fname string, profiles []ffmpeg.VideoProfile) (*transcoderSession, error) {
+
+	lb.mu.Lock()
```
ericxtang, line 69: if this lock is only protecting the creation of the session, should we unlock after line 83 instead of defer?

j0sh: Do you mean after assigning `lb.idx` on [line 84](https://github.com/livepeer/go-livepeer/pull/1124/files/4f28d093eb2051d2e45b6d4592b0177ff6d7724d#diff-da447e938cf8298a688acc7c4626255aR84)? (That needs to be protected too).

We could explicitly unlock, but after this point the function only starts a goroutine and exits. Feels more future-proof to `defer` the unlock.

Is there a case that could lead to unnecessary lock contention?

core/lb.go
```diff
@@ -0,0 +1,202 @@
...
+
+type LoadBalancedTranscoder interface {
+	Transcoder
+	Stop()
+}
+
+type newTranscoderFn func(device string, workDir string) LoadBalancedTranscoder
+
+type LoadBalancingTranscoder struct {
+	transcoders []string
```
yondonfu: Should we add a comment noting that this needs to be a slice of device IDs?

core/lb.go
```diff
@@ -0,0 +1,202 @@
...
+	}
+}
+
+func (sess *transcoderSession) Transcode(job string, fname string, profiles []ffmpeg.VideoProfile) (*TranscodeData, error) {
+	params := &transcoderParams{job: job, fname: fname, profiles: profiles,
+		res: make(chan struct {
+			*TranscodeData
+			error
+		})}
+	// XXX check transcoding on a looped out channel . May be threading issues...
```
yondonfu: I think we'll always get `ErrTranscoderBusy` in this case since `sess.sender <- params` will always be blocked once `sess.sender` is set to nil for a looped out channel.

core/lb.go
```diff
@@ -0,0 +1,202 @@
...
+
+	"github.com/golang/glog"
+
+	"github.com/livepeer/go-livepeer/common"
+	"github.com/livepeer/lpms/ffmpeg"
+)
+
+var ErrTranscoderBusy = errors.New("TranscoderBusy")
+
+type LoadBalancedTranscoder interface {
```
yondonfu: I know we already have an unexported struct named `transcoderSession`, but I feel like `TranscoderSession` could also be a fitting name for this interface since the types that implement the interface do not necessarily need to be load balanced, but the one thing they all have in common is that they are associated with a single session that a caller can start/stop.

core/lb.go
```diff
@@ -0,0 +1,202 @@
...
+}
+
+func calculateCost(profiles []ffmpeg.VideoProfile) int {
+	cost := 0
+	for _, v := range profiles {
+		w, h, err := ffmpeg.VideoProfileResolution(v)
+		if err != nil {
+			continue
+		}
+		cost += w * h * int(v.Framerate) // TODO incorporate duration
```
yondonfu, line 220: Is the plan to incorporate duration outside of this PR?

core/lb_test.go
```diff
@@ -0,0 +1,198 @@
...
+	lb := NewLoadBalancingTranscoder("0,1,2,3,4", "", newStubTranscoder).(*LoadBalancingTranscoder)
+	sessions := []string{"a", "b", "c", "d", "e"}
+	profiles := []ffmpeg.VideoProfile{}
+	for _, v := range ffmpeg.VideoProfileLookup {
+		profiles = append(profiles, v)
+	}
+
+	shuffleProfiles := func(t *rapid.T) []ffmpeg.VideoProfile {
+		// fisher-yates shuffle. or an approximation thereof. ( should test this)
+		for i := len(profiles) - 1; i >= 1; i-- {
```
yondonfu: I think [rand.Shuffle](https://golang.org/pkg/math/rand/#Shuffle) can be used here - looks like the [implementation](https://golang.org/src/math/rand/rand.go) (should be the result for CMD+F using "Shuffle") uses Fisher-Yates

core/lb_test.go
```diff
@@ -0,0 +1,198 @@
...
+	transcoder := sess.transcoder.(*StubTranscoder)
+	// Now set the fail flag in and check for propagation
+	transcoder.FailTranscode = true
+	_, err = lb.Transcode(sessName, "", []ffmpeg.VideoProfile{ffmpeg.P144p30fps16x9})
+	assert.Equal(err, ErrTranscode)
+
+	// Property: Cancelling a transcoder should remove session.
+	oldContext := transcodeLoopContext
+	defer func() { transcodeLoopContext = oldContext }()
+	//var wg sync.WaitGroup
```
yondonfu: Can remove?

core/lb_test.go
j0sh: That would be ideal. I think the issue with `shuffle` is in reusing the RNG that the `rapid` library uses, do you see a way to do that?

core/lb.go
j0sh: Perhaps eventually - it would allow for more accurate load balancing - but that has the same issues mentioned here https://github.com/livepeer/go-livepeer/pull/1124#discussion_r344978928

core/lb.go
j0sh: Yep, the comment is likely outdated - had an earlier issue in this area and fixed it. Mostly this is a reminder for myself to add this case the test suite. Don't let me merge until that's done!

core/lb.go
j0sh: Sure - fixed

core/lb.go
j0sh: Sure, renamed to `TranscoderSession`

core/lb_test.go
j0sh: Yes - will be reworking the whole thing

iameli:
Just got this from a go-livepeer ffmpeg build downstream of this branch: `WARNING: Disabled scale_cuda_filter because not any dependency is satisfied: cuda_nvcc cuda_llvm`

Looking into it, just wanted to log that was happening. Gonna take this chance to look into compiling with llvm, 'cause we want to do that anyway and it unblocks Windows GPU and stuff.

iameli:
Cool, turned out to be kinda easy? <https://github.com/livepeer/go-livepeer/pull/1207/files>

core/lb.go
```diff
@@ -0,0 +1,202 @@
...
+		load:        make(map[string]int),
+		sessions:    make(map[string]*transcoderSession),
+	}
+}
+
+func (lb *LoadBalancingTranscoder) Transcode(job string, fname string, profiles []ffmpeg.VideoProfile) (*TranscodeData, error) {
+
+	lb.mu.RLock()
+	session, exists := lb.sessions[job]
+	lb.mu.RUnlock()
```
darkdarkdragon, line 54: This will allow creation of two sessions for same job. We should keep the lock until session is created.

core/lb.go
```diff
@@ -0,0 +1,202 @@
...
+	// Local cleanup function
+	cleanupSession := func() {
+		lb.mu.Lock()
+		defer lb.mu.Unlock()
+		sess, exists := lb.sessions[job]
+		if !exists {
+			return
+		}
+		if sess.nonce != session.nonce {
+			// Session cached at `job` has since been recreated
```
darkdarkdragon: How can this happen? Only way I see is the problem in Transcode function I've commented on earlier.
And when it happens, this will leave `lb.load` not decremented?

core/lb.go
```diff
@@ -0,0 +1,202 @@
...
+		})}
+	// XXX check transcoding on a looped out channel . May be threading issues...
+	select {
+	case sess.sender <- params:
+		glog.V(common.DEBUG).Info("LB: Transcode submitted for ", sess.key)
+	default:
+		glog.V(common.DEBUG).Info("LB: Transcoder was busy; exiting ", sess.key)
+		return nil, ErrTranscoderBusy
+	}
+	res := <-params.res
```
darkdarkdragon, line 209: There is memory leak here. `sess.sender` is buffered (of size 1), so two different goroutines can make it to this line simultaneously. And if transcode for the first one will return error, this will make loop to exit, and thus second goroutine will be waiting at this line for eternity.

darkdarkdragon (changes requested):

core/lb.go
j0sh: We'd have to keep a write lock. Not sure if that's desirable when this particular code path should be read-heavy. That's why this check is here as you mentioned: https://github.com/livepeer/go-livepeer/pull/1124#discussion_r357898212

core/lb.go
j0sh: >  this will leave lb.load not decremented?

Yes, good point.

core/lb.go
darkdarkdragon: OK, but then we can in `createSession`, after taking write lock check again if session exists, and return it?

core/lb_test.go
j0sh: Fixed in https://github.com/livepeer/go-livepeer/pull/1124/commits/b9e09ffeac9470c420a0c489cb4df12f4bb2df8d

core/lb.go
j0sh: Fixed in https://github.com/livepeer/go-livepeer/pull/1124/commits/b9e09ffeac9470c420a0c489cb4df12f4bb2df8d

core/lb.go
j0sh: Fixed in https://github.com/livepeer/go-livepeer/pull/1124/commits/b9e09ffeac9470c420a0c489cb4df12f4bb2df8d

core/lb.go
j0sh: Took the brute force route here and made a session-level lock for the whole function in https://github.com/livepeer/go-livepeer/pull/1124/commits/b9e09ffeac9470c420a0c489cb4df12f4bb2df8d

darkdarkdragon: If I understand it right, this still doesn't fix problem:
1. First `Transcode` invocation takes lock
2. Second invocation at the same time stop at lock and waits
3. loop receives error, sends it into `params.res` channel and starts exiting. At exit it stops and lock
4. Now we have two threads waiting at lock. First `Transcode` invocation releases lock, and if the second `Transcode` call takes it, then we have problem again.

Also there is problem in `case <-ctx.Done()` branch - it can lead to memory leak in the same way.
I think proper solution could be like this:
```golang
func (sess *transcoderSession) loop() {
	defer func() {
		sess.transcoder.Stop()
		sess.sender = nil // unblocks any sender + select{ ... }
	}()
	// Run everything on a single loop to mitigate threading issues,
	//   especially around transcoder cleanup
	for {
		ctx, cancel := sess.makeContext()
		select {
		case <-ctx.Done():
			// Terminate the session after a period of inactivity
			glog.V(common.DEBUG).Info("LB: Transcode loop timed out for ", sess.key)
			return
		case params := <-sess.sender:
			cancel()
			res, err :=
				sess.transcoder.Transcode(params.job, params.fname, params.profiles)
			params.res <- struct {
				*TranscodeData
				error
			}{res, err}
			if err != nil {
				glog.V(common.DEBUG).Info("LB: Stopping transcoder due to error for ", sess.key)
				return
			}
		}
	}
}

func (sess *transcoderSession) Transcode(job string, fname string, profiles []ffmpeg.VideoProfile) (*TranscodeData, error) {
	params := &transcoderParams{job: job, fname: fname, profiles: profiles,
		res: make(chan struct {
			*TranscodeData
			error
		})}
	sess.mu.Lock()
	defer sess.mu.Unlock()
	select {
	case sess.sender <- params:
		glog.V(common.DEBUG).Info("LB: Transcode submitted for ", sess.key)
	default:
		glog.V(common.DEBUG).Info("LB: Transcoder was busy; exiting ", sess.key)
		return nil, ErrTranscoderBusy
	}
	select {
	case res := <-params.res:
		return res.TranscodeData, res.error
	case <-time.After(8 * time.Second):
		glog.V(common.DEBUG).Info("LB: Transcoder timeout; exiting ", sess.key)
		return nil, ErrTranscoderBusy
	}
}
```

j0sh: > loop receives error, sends it into params.res channel and starts exiting. At exit it stops and lock
Now we have two threads waiting at lock. First Transcode invocation releases lock, and if the second Transcode call takes it, then we have problem again.

I'm a little hesitant to add yet another layer of timeouts here, just because of how brittle it is (and we already have a number of other request-level timeouts to contend with). What might help this issue is to drain the channel after assigning `sess.sender = nil` in the `loop` cleanup function. Something like:

```golang
sess.mu.Lock()
sender := sess.sender
sess.sender = nil
sess.mu.Unlock()
// Drain the buffered channel
for {
  select {
  params := <-sender:
    params.res <- struct { *TranscodeData, error } { nil, ErrTranscodeEnded }
  default:
    return
  }
}
```

darkdarkdragon: Good idea, this should solve the problem.

j0sh: Fixed in 80d64d5b129d7e7ea6e4828d388ab499cf971dea . Hard to unit-test for directly without making testsuite-specific additions to the code, so I'll just leave things running under a stress test. Tightening this up seems to have removed the need for locking around the stub transcoder, which is a good thing. Note that we run all the load balancing tests via `-race` now so hopefully that helps a bit in terms of future-proofing thread safety here.

core/lb.go
j0sh: Should be mitigated now after addressing related issues @darkdarkdragon brought up. Note that all load balancer tests are run under `-race` now which hopefully will help with future-proofing this.  See: https://github.com/livepeer/go-livepeer/pull/1124/files#diff-1b531c5e405e0bc95890d52cf2f2f5b5

j0sh:
I think most of the addressable feedback has been resolved now.

LPMS and ffmpeg are updated in https://github.com/livepeer/go-livepeer/pull/1124/commits/85bf468ac42879b6651de8b96f6a71842c0e516d and https://github.com/livepeer/go-livepeer/pull/1124/commits/18ff2c6a1f0bfb0d99fd8aaafd07150775fae8be . These are likely the final versions for this PR, but will update if anything changes. CircleCI is green.

darkdarkdragon:
@j0sh I have one more fix for this, I'll commit it once I'll make tests stable

core/lb.go
darkdarkdragon: Added unit test here 6e4d6981f69f349336385a8ba3f0e633379d9951

j0sh:
Hopefully this will fix the deadlocking for once and all: https://github.com/livepeer/go-livepeer/pull/1124/commits/224159423554551dce68afc113f60d210dfbbce2

The basic mechanism is: as we exit the function, attempt to fill up the channel with empty data. Since the channel is buffered with size 1, we know that any successful writes will mean the channel is full, and we can then exit immediately knowing that subsequent write attempts will fail.

This also eliminates any locking within the session; the LB tests are still `-race` clean.

j0sh:
CircleCI is failing because of the following unrelated test. If things look good for everybody, I'll rebase and re-push to retry CI.

```
--- FAIL: TestQueueTicketAndSignalMaxFloat (0.05s)
    sendermonitor_test.go:248: 
        	Error Trace:	sendermonitor_test.go:248
        	Error:      	Not equal: 
        	            	expected: 0x3
        	            	actual  : 0x2
        	Test:       	TestQueueTicketAndSignalMaxFloat
    sendermonitor_test.go:249: 
        	Error Trace:	sendermonitor_test.go:249
        	Error:      	Not equal: 
        	            	expected: 0x2
        	            	actual  : 0x3
        	Test:       	TestQueueTicketAndSignalMaxFloat
=== RUN   TestCleanup
```

core/lb.go
```diff
@@ -0,0 +1,223 @@
...
+		// and we can safely exit immediately knowing that subsequent writes
+		// will be rejected
+		for {
+			select {
+			case params := <-sess.sender:
+				params.res <- struct {
+					*TranscodeData
+					error
+				}{nil, ErrTranscoderStopped}
+			case sess.sender <- &transcoderParams{}:
```
darkdarkdragon, line 163: awesome!

darkdarkdragon (approved):
LGTM 🚢

core/lb.go
yondonfu: 🎉

yondonfu (approved):

j0sh:
Rebased, merging

Issue #1098, opened 2019-09-20
Prevent datadir reuse on different networks


j0sh:
**Is your feature request related to a problem? Please describe.**

There have been issues in the past where the same data directory would be used on different networks (eg, mainnet vs Rinkeby). This leads to odd results due to the state cached in the DB not matching the current network.

While we've made structural changes to datadir handling to make this type of collision less likely, it's still a possibility if a custom `-datadir` is specified, but the `-network` changes between runs.

**Describe the solution you'd like**

We should try to detect this problem at runtime, and inform the user of a mismatch. One approach to doing this is as follows:

* The DB should record the network it's running on when the DB is first created.
* After a node restart, if the current network doesn't match the network recorded in the DB, then the node should exit with an error message.
* The network name can be recorded in the `kv` table.

**Describe alternatives you've considered**

Generated file names (DB, private key, certificate, etc) could all be based on the network name, which would allow data for multiple networks to coexist within the same data directory.

kyriediculous:
As mentioned in the issue outline we already try to mitigate against `datadir` collisions by concatenating the network name to a non-custom `-datadir`. 

Collision when providing `-datadir` flags makes it an end-user responsability however which isn't ideal. 

For this reason I actually like the proposed alternative better than the current structure. I would even go a step further and employ `name+chainID`. 

That being said the simplest and most robust solution seems to store the `chainID` for a network name, on node startup we would then make an RPC call to the provided `ethUrl` (or infura when not provided) and compare that against the ID from the database. 

This should prevent even changing networks but keeping the name the same which is an edge case unhandled by simply storing the network name.

yondonfu:
Agreed that storing the chainID in the DB would be a more robust solution.

If we store the chainID in the DB, we should also define the expected behavior when the node is running in off-chain mode.

One approach is to set the default value for chainID to be 0 which does not correspond to a valid chain. The node would treat chainID = 0 as an indication that the DB was configured in off-chain mode. An alternative is to not set the chainID at all, but this would require error handling and special casing errors when querying for the chainID.

PR #1125, opened 2019-10-19
gomod, eth: fix nonce gap issues

### LINKED ISSUE #1104: Improve nonce handling strategy to mitigate nonce gap issues

There have been reports of nonce gaps that cause txs to sit pending until the gaps are filled. The node's nonce management strategy should be improved to mitigate these issues.

Nodes that automatically submit txs should never run into nonce gap issues.

kyriediculous:
This PR upgrades the version of go-ethereum to include https://github.com/ethereum/go-ethereum/pull/15794 in order to try to mitigate nonce issues that are occuring in the 0.3.3 go-livepeer release

- Now use `go-ethereum@1.9.3` instead of our own `v.1.8.x` fork to include https://github.com/ethereum/go-ethereum/pull/15794
- remove custom `NonceManager` implementation & usage
- changed a function call in `eth/watchers/eventdecoder` to match the API updates for the new `go-ethereum` version
- ran `go mod tidy`

Ran unit tests

Fixes #1104 

**Checklist:**


kyriediculous:
Updated the PR to include a wrapper `Backend` for `ethclient.Client`  that uses a `PendingNonceAt` method which implements the `NonceManager` logic so that this logic no longer has to be maintained in a seperate go-ethereum fork. 

This PR is unlikely however to address nonce gap issues occuring in releases that already included the custom nonce manager.

eth/backend.go
```diff
@@ -0,0 +1,171 @@
...
+
+import (
+	"context"
+	"fmt"
+	"math/big"
+	"strings"
+
+	ethereum "github.com/ethereum/go-ethereum"
+	"github.com/ethereum/go-ethereum/accounts/abi"
+	ethabi "github.com/ethereum/go-ethereum/accounts/abi"
```
yondonfu: Let's remove this package alias and just import the `abi` package

eth/backend.go
```diff
@@ -0,0 +1,171 @@
...
+type Backend interface {
+	ethereum.ChainReader
+	ethereum.ChainStateReader
+	ethereum.ChainSyncReader
+	ethereum.TransactionReader
+	ethereum.TransactionSender
+	ethereum.ContractCaller
+	ethereum.PendingContractCaller
+	ethereum.PendingStateReader
+	// ethereum.PendingStateEventer
```
yondonfu: Can we remove this?

eth/backend.go
```diff
@@ -0,0 +1,171 @@
...
+	contracts.MinterABI,
+	contracts.RoundsManagerABI,
+	contracts.ServiceRegistryABI,
+	contracts.TicketBrokerABI,
+}
+
+type Backend interface {
+	ethereum.ChainReader
+	ethereum.ChainStateReader
+	ethereum.ChainSyncReader
```
yondonfu: Don't think we actually use any of this interface's methods right now

eth/backend.go
```diff
@@ -0,0 +1,171 @@
...
+}
+
+func (b *backend) PendingNonceAt(ctx context.Context, account common.Address) (uint64, error) {
+	var nonce uint64
+	var err error
+	if b.nonceManager != nil {
+		b.nonceManager.Lock(account)
+		defer func() {
+			if err == nil {
+				b.nonceManager.Update(account, nonce)
```
yondonfu: We should be updating the locally tracked nonce in `SendTransaction()` if the tx is successfully submitted

eth/backend.go
```diff
@@ -0,0 +1,171 @@
...
+		client,
+		methods,
+		NewNonceManager(client),
+	}, nil
+}
+
+func (b *backend) PendingNonceAt(ctx context.Context, account common.Address) (uint64, error) {
+	var nonce uint64
+	var err error
+	if b.nonceManager != nil {
```
yondonfu: Don't think we need to check if `b.nonceManager` is nil here since we always instantiate it in the constructor.

eth/backend.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+
+	return &backend{
+		client,
+		methods,
+		NewNonceManager(client),
+	}, nil
+}
+
+func (b *backend) PendingNonceAt(ctx context.Context, account common.Address) (uint64, error) {
```
yondonfu, line 61: I think we can shorten this method to just:

```
func (b *backend) PendingNonceAt(ctx context.Context, account common.Address) (uint64, error) {
    return b.nonceManager.Next(account)
}
```

eth/backend.go
```diff
@@ -0,0 +1,171 @@
...
+}
+
+func (b *backend) SendTransaction(ctx context.Context, tx *types.Transaction) error {
+	err := b.Client.SendTransaction(ctx, tx)
+
+	// decode tx data
+	data := tx.Data()
+	method := b.methods[string(data[:4])]
+	params := make(map[string]interface{})
+	method.abi.UnpackIntoMap(params, method.name, data[4:])
```
yondonfu: Should check error returned here.

eth/backend.go
```diff
@@ -0,0 +1,171 @@
...
+	ethereum.TransactionSender
+	ethereum.ContractCaller
+	ethereum.PendingContractCaller
+	ethereum.PendingStateReader
+	// ethereum.PendingStateEventer
+	ethereum.GasEstimator
+	ethereum.GasPricer
+	ethereum.LogFilterer
+
+	NetworkID(ctx context.Context) (*big.Int, error)
```
yondonfu: I believe this is needed in this interface right now because it is used by the web server `/EthNetworkID` endpoint. However,  I think we should switch over to exclusively using `chainID` instead of the `networkID` (see this [blog post](https://medium.com/@pedrouid/chainid-vs-networkid-how-do-they-differ-on-ethereum-eec2ed41635b) for a good overview of the difference between the two) especially since we'll be using it for #1098 and #1132. The changes necessary would be small: we just need to call [ChainID()](https://github.com/ethereum/go-ethereum/blob/master/ethclient/ethclient.go#L65) instead of `NetworkID()`.

eth/backend.go
```diff
@@ -0,0 +1,171 @@
...
+
+	return nonce, nil
+}
+
+func (b *backend) SendTransaction(ctx context.Context, tx *types.Transaction) error {
+	err := b.Client.SendTransaction(ctx, tx)
+
+	// decode tx data
+	data := tx.Data()
+	method := b.methods[string(data[:4])]
```
yondonfu: Should check that the key actually exists in the map

eth/backend.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+
+	return nonce, nil
+}
+
+func (b *backend) SendTransaction(ctx context.Context, tx *types.Transaction) error {
+	err := b.Client.SendTransaction(ctx, tx)
+
+	// decode tx data
+	data := tx.Data()
```
yondonfu, line 84: I added error checking, ran this locally and I'm seeing a bunch of errors such as:

```
E1024 12:07:58.759626   57619 backend.go:113] abi: improperly encoded boolean value
```

```
E1024 12:07:04.060205   57594 backend.go:113] abi: unmarshalling empty output
```

yondonfu (changes requested):

eth/backend.go
kyriediculous: Actually we are doing it already but it might not be very clear.

`BoundContract.Transact` will call `Backend.PendingNonceAt()` before calling `Backend.SendTransaction()`. 

While we are not checking if `SendTransaction` is succesful, we didn't do that in the `BoundContract.Transact` method for the maintained go-ethereum fork either. The logic used here is 99% identical

yondonfu: Hm so the difference now is that the `defer` statement is going to execute when `Backend.PendingNonceAt()` returns. Previously, the `defer` statement executed when `BoundContract.Transact()` returned and would check the last tracked error (the same [err variable](https://github.com/livepeer/go-ethereum/blob/master/accounts/abi/bind/base.go#L234) is used throughout the method) in the method which happened to be the one returned by `SendTransaction()`. So, the go-ethereum fork *was* checking if `SendTransaction()` was successful although the approach used was definitely confusing.

kyriediculous: Ah yes I see, the order of declaration there is a little confusing.

kyriediculous: The issue is that `SendTransaction` doesn't really have access to the sender address to call `Update` on `NonceManager`

yondonfu: A workaround is to recover the sender address from the signed transaction that is passed into `SendTransaction()`.

We can use the [tx.AsMessage()](https://github.com/ethereum/go-ethereum/blob/master/core/types/transaction.go#L220) method to return a [Message](https://github.com/ethereum/go-ethereum/blob/master/core/types/transaction.go#L412) type and then we can call `.From()` to return the account address that signed the tx. See this [SO post](https://ethereum.stackexchange.com/questions/67449/given-a-raw-transaction-how-to-get-senders-address) for example usage.

For now we'll want to use `types.HomesteadSigner{}` as the provided signer since this is the hardcoded signer used by [BoundContract](https://github.com/ethereum/go-ethereum/blob/master/accounts/abi/bind/base.go#L237). Side note: `BoundContract` should be using the EIP155Signer for signing transactions...there is an [open go-ethereum PR](https://github.com/ethereum/go-ethereum/pull/17046) with this change.

eth/backend.go
```diff
@@ -0,0 +1,151 @@
...
+		}
+		for _, m := range parsedAbi.Methods {
+			methods[string(m.ID())] = m.Name
+		}
+	}
+
+	return methods, nil
+}
+
+func formatParams(params map[string]interface{}) string {
```
yondonfu: Can remove this?

yondonfu (changes requested):

eth/backend.go
kyriediculous: fixed in afb1e0a5f6d9fbbf0c0d4a642a6580281fb5a901

cmd/livepeer_cli/livepeer_cli.go
```diff
@@ -166,11 +166,11 @@ func (w *wizard) doCLIOpt(choice string, options []wizardOpt) {
 	log.Error("That's not something I can do")
 }
 
-var RinkebyNetworkId = "4"
-var DevenvNetworkId = "54321"
+var RinkebychainID = "4"
```
yondonfu: Capitalize the "C" in chainID? Think that would be easier to read i.e. `RinkebyChainID` and `DevenvChainID`

yondonfu (changes requested):

kyriediculous:
fixed in 4eea557f

yondonfu (commented):

kyriediculous:
rebased !

yondonfu (approved):
LGTM!

PR #1126, opened 2019-10-19
add new build notification

iameli:
Just a tiny webhook that fires after the Docker build passes, triggering some other pieces of the infrastructure to check for new images.


j0sh (approved):

iameli:
ty

PR #1127, opened 2019-10-20
eth+pm+server: Generate and verify signatures with v = 27/28

yondonfu:

This PR updates the node's signature format to always use 27 or 28 for the v value. Signatures are of the form `[r || s || v]` where r and s constitute the actual signature. v is the recovery ID that is used for recovering the public key associated with a signature. The [go-ethereum package the node uses for signing](https://github.com/ethereum/go-ethereum/blob/master/crypto/signature_nocgo.go#L59) generates signatures with v = 0/1.

The protocol contracts were recently updated to use the newest version of the OpenZeppelin signature verification library in [this commit](https://github.com/livepeer/protocol/pull/342/commits/325fe884b9c16e380ff1ba3cfb7dd9a000ecdefb). The updated [library](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/cryptography/ECDSA.sol) includes the following additional checks to defend against malleable signatures:

- Checks that the signature's s value is in the lower range of the N (curve order)
- Checks that the signature's v value is either 27/28

In order to pass on-chain signature verification, the node needs to generate signatures that pass these additional checks.

- Update `AccountManager.Sign()` to convert signatures' v value to 27/28
- Add tests for `AccountManager.Sign()`
- Update `pm.VerifySig()` to perform all the checks that are present in the on-chain signature verification library to ensure that the node verifies signatures the same way
- Add tests for `pm.VerifySig()`
- Update tests in the `server` package that rely on signature generation/verification

Note: As already mentioned in a comment, `pm.VerifySig()` should eventually be extracted out of the `pm` package, but I chose not to do so in this PR as it requires some additional package restructuring.

Added unit tests and ran a B/O/T setup in on-chain mode using the test-harness.

**Checklist:**

### DIFF
--- a/eth/accountmanager.go
+++ b/eth/accountmanager.go
@@ -11,7 +11,6 @@ import (
 	ethcommon "github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/console"
 	"github.com/ethereum/go-ethereum/core/types"
-	"github.com/ethereum/go-ethereum/crypto"
 	"github.com/golang/glog"
 	"github.com/livepeer/go-livepeer/common"
 )
@@ -147,10 +146,20 @@ func (am *accountManager) SignTx(signer types.Signer, tx *types.Transaction) (*t
 
 // Sign byte array message. Account must be unlocked
 func (am *accountManager) Sign(msg []byte) ([]byte, error) {
-	personalMsg := fmt.Sprintf("\x19Ethereum Signed Message:\n%d%s", 32, msg)
-	personalHash := crypto.Keccak256([]byte(personalMsg))
+	ethHash := accounts.TextHash(msg)
+	sig, err := am.keyStore.SignHash(am.account, ethHash)
+	if err != nil {
+		return nil, err
+	}
 
-	return am.keyStore.SignHash(am.account, personalHash)
+	// sig is in the [R || S || V] format where V is 0 or 1
+	// Convert the V param to 27 or 28
+	v := sig[64]
+	if v == byte(0) || v == byte(1) {
+		v += 27
+	}
+
+	return append(sig[:64], v), nil
 }
 
 func (am *accountManager) Account() accounts.Account {
--- a/pm/helpers.go
+++ b/pm/helpers.go
@@ -1,11 +1,18 @@
 package pm
 
 import (
-	"fmt"
+	"math/big"
 	"math/rand"
 
+	"github.com/ethereum/go-ethereum/accounts"
 	ethcommon "github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/pkg/errors"
+)
+
+var (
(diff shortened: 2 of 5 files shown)


eth/accountmanager.go
```diff
@@ -147,10 +146,23 @@ func (am *accountManager) SignTx(signer types.Signer, tx *types.Transaction) (*t
...
+	ethHash := accounts.TextHash(msg)
+	sig, err := am.keyStore.SignHash(am.account, ethHash)
+	if err != nil {
+		return nil, err
+	}
+
+	// sig is in the [R || S || V] format where V is 0 or 1
+	// Convert the V param to 27 or 28
+	v := sig[64]
+	if v == byte(0) {
```
kyriediculous: I think we can do something like `if v == byte(0) || v == byte(1) { v = byte(int(v) + 27) )` instead. Haven't tested it.

pm/helpers.go
```diff
@@ -1,27 +1,61 @@
...
+	recovered, err := ecrecover(msg, sig)
 	if err != nil {
 		return false
 	}
 
-	return crypto.PubkeyToAddress(*pubkey) == addr
+	return recovered == addr
+}
+
+func ecrecover(msg, sig []byte) (ethcommon.Address, error) {
```
kyriediculous, line 31: I guess we're using a custom `ecrecover` implementation here because https://godoc.org/github.com/ethereum/go-ethereum/crypto#Ecrecover doesn't check whether the recovery id is `27 || 28` ?

kyriediculous (changes requested):
Looks good, just a minor comment and question

eth/accountmanager.go
yondonfu: Fixed in b309607

pm/helpers.go
yondonfu: `crypto.SigToPub()` expects a signature with v = 0/1. Furthermore, the non-cgo version of `crypto.SigToPub()` does not check that the signature's s-value is in the lower range. So, in order to make sure that local signature verification matches on-chain signature verification I added checks in `pm.ecrecover()` that correspond to the checks in the on-chain signature verification library (i.e. invalid signature length, invalid s-value, invalid v-value)

Chose to use `crypto.SigToPub()` instead of `crypto.Ecrecover()` in this function because it seemed more straightforward to convert the pub key returned by `crypto.SigToPub()` into a `common.Address` using `crypto.PubkeyToAddress()` - `crypto.Ecrecover()` returns a `[]byte` which would need to be converted into a pub key before finally converting into `common.Address`.

kyriediculous (commented):
changes look good let's rebase !

yondonfu:
Rebased

kyriediculous:
LGTM 🚀

yondonfu:
Need an approval!

kyriediculous (approved):

kyriediculous:
Right ! 😅 

Not sure what is going on with the azure pipelines build process though.

PR #1130, opened 2019-10-22
Keep balance when removing session

### LINKED ISSUE #1079: Broadcaster insufficient balance error when reconnecting with negative balance
**Describe the bug**

A broadcaster with a negative credit balance when it disconnects from an orchestrator will receive an insufficient balance error if it reconnects with the same orchestrator and sends a ticket/segment within [1 minute](https://github.com/livepeer/go-livepeer/blob/master/cmd/livepeer/livepeer.go#L69) of the original disconnect (1 minute is the current [cleanup interval](https://github.com/livepeer/go-livepeer/blob/master/core/accounting.go#L126) for cached broadcaster balances).

This behavior is observed because when the broadcaster reconnects with the orchestrator, the broadcaster will view its balance as 0 and the orchestrator will view its balance as N where N is some negative value representing the broadcaster's balance at the time of the original disconnect. The broadcaster will calculate the # of tickets to send to the orchestrator in order to meet the orchestrator's minimum credit requirement (equal to the orchestrators' desired ticket EV). However, since the broadcaster and orchestrator have different views of the broadcaster's balance, the broadcaster will not meet the orchestrator's minimum credit requirement in this case

Example flow:

1. Broadcaster disconnects with balance = -.5 ETH
2. Broadcaster reconnects - it sees its balance = 0 ETH while the orchestrator still sees its balance = -.5 ETH
3. Broadcaster sends tickets with total value = 1 ETH which would typically increase the broadcaster's balance to meet the minimum credit requirement since the orchestrator's ticket EV is 1 ETH
4. Orchestrator sees that the broadcaster's balance = -.5 + 1 = .5 ETH which is less than the orchestrator's minimum credit requirement. Orchestrator responds with an insufficient balance error

**To Reproduce**
1. Start a B in on-chain mode with a `-maxTicketEV N`
2. Start an O in on-chain mode with `-ticketEV N` and `-pricePerUnit N`. We set `ticketEV = pricePerUnit` in order to ensure that the amount debited after transcoding the first segment will cause the broadcaster's balance to go negative
3. Stream into B
4. Observe B disconnect from O after the first segment is transcoded
5. Observe B reconnect with O because it is the only active node
6. Observe B encounter an insufficient balance error

**Expected behavior**

A broadcaster should be able to reconnect with an honest O and send tickets/segments without encountering an insufficient balance error.

**Screenshots**

 - OS: OSX

yondonfu:

This PR updates `BroadcastSessionsManager.removeSession()` to *not* clear the balance associated with the session when the session is removed. This prevents the following scenario:

- B has negative balance with O
- B disconnects because it encounters an error and removes the session
- B refreshes its session list
- B reconnects with O - its session balance is 0, but O's view of B's balance is still negative
- B sends a payment to O with its value calculated based on a session balance of 0
- B encounters an insufficient balance error

- Do not call `Balance.Clear()` in `removeSession()`
- Remove `Balance.Clear()` because it is no longer used

After this change, B will only clear the balance for a manifestID if it has not been accessed for longer than the TTL defined in `core.Balances`.

Checked that B did not encounter insufficient balance errors in a setup (using the devtool script) replicating the scenario described in #1079.

Fixes #1079

**Checklist:**

### DIFF
--- a/core/accounting.go
+++ b/core/accounting.go
@@ -50,11 +50,6 @@ func (b *Balance) StageUpdate(minCredit, ev *big.Rat) (int, *big.Rat, *big.Rat) 
 	return int(size), new(big.Rat).Mul(new(big.Rat).SetInt64(size), ev), existingCredit
 }
 
-// Clear zeros the balance
-func (b *Balance) Clear() {
-	delete(b.balances.balances, b.manifestID)
-}
-
 // Balances holds credit balances on a per-stream basis
 type Balances struct {
 	balances map[ManifestID]*balance
--- a/core/accounting_test.go
+++ b/core/accounting_test.go
@@ -80,23 +80,6 @@ func TestBalance_StageUpdate(t *testing.T) {
 	assert.Zero(big.NewRat(0, 1).Cmp(balances.Balance(mid)))
 }
 
-func TestBalance_Clear(t *testing.T) {
-	mid := ManifestID("some manifestID")
-	balances := NewBalances(5 * time.Second)
-	b := NewBalance(mid, balances)
-
-	assert := assert.New(t)
-
-	// Test non-nil key
-	b.Credit(big.NewRat(5, 1))
-	b.Clear()
-	assert.Nil(balances.balances[mid])
-
-	// Test nil key
-	b.Clear()
-	assert.Nil(balances.balances[mid])
-}
-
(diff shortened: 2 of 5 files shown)


kyriediculous (approved):
LGTM 🚀

PR #1131, opened 2019-10-23
Track stream balances per address

### LINKED ISSUE #1129: Balances for sessions with different orchestrators reference the same value per manifestID
**Describe the bug**

When a broadcaster constructs a list of broadcast sessions, each session will be associated with a different orchestrators. Each session also references a `Balance` which is a wrapper struct that makes updates to an underlying `Balances` struct that keeps track of the current balance for a manifestID. Each `Balance` is initialized with [the same manifestID](https://github.com/livepeer/go-livepeer/blob/master/server/broadcast.go#L211). As a result, each session will be using the same balance despite being associated with a different orchestrator. This is problematic because a payment sent to an orchestrator for one session will update the balance for all sessions that use the same manifestID. 

**Expected behavior**

A payment sent to an orchestrator for a particular manifestID should not affect the balances for any sessions associated with other orchestrators that also use the same manifestID.

**Additional context**

Possible solutions:

- Update the `Balances` struct to store a nested map `map[ethcommon.Address]map[ManifestID]*balance`
- Update the `Balances` struct to store a map `map[string]*balance` where the format of the string value is determined by the caller. Then, the caller can define the format as:

```
keccak256(broadcaster.Address, manifestID)
```

yondonfu:

This PR updates how balances (for payments) are tracked. The `core.AddressBalances` type is introduced which maintains an internal map from ETH addresses to a set of balances associated with different manifestIDs (implemented by the existing `core.Balances` type). Previously, the node was only keeping track of balances per manifestID which was problematic due to the possibility of manifestID collisions (since manifestID is defined by the broadcaster).

- 7994139: Add the `core.AddressBalances` type
- 206dc25: Update packages that were previously using `core.Balances` to now use `core.AddressBalances`

Added unit tests. Ran a test-harness 2O/1B setup 

Fixes #1129

**Checklist:**


kyriediculous (commented):
I'm wondering if we should use `AddressBalances` on the broadcaster as well then where we store 
`orchestrator => manifestID` to avoid collisions there as well when B chooses to operate redundantly?

yondonfu:
> I'm wondering if we should use AddressBalances on the broadcaster as well

We actually are using `AddressBalances` under the hood.

The `Balance` struct now [wraps AddressBalances](https://github.com/livepeer/go-livepeer/pull/1131/files#diff-541c0849ae46cbbfdb65ee01dcec77b5R15). When we create a `Balance` struct for a broadcast session, we [instantiate it](https://github.com/livepeer/go-livepeer/pull/1131/files#diff-cecc27bfe2b6b19181ce443733dfba57R209) with the address of the payment recipient and the manifestID.

kyriediculous:
Ah yes I see it's part of the second commit. 

I'm wondering though could we make it a nested map instead and migrate `Balances` into `AddressBalances` , then rename `Balance` to `Balances` to make the structure of the package a little clearer? 

Currently we have `Balance -> AddressBalances -> Balance -> balance`
`Balances -> AddressBalances -> balance` seems clearer

We could do it in a seperate PR though if you want.

yondonfu:
> I'm wondering though could we make it a nested map instead and migrate Balances into AddressBalances 

I considered this, but wrapping `Balances` in `AddressBalances` seemed to be the cleanest way to implement this since a) it minimizes the amount of refactoring b) it makes use of the existing concurrency safety in `Balances` which we need for each set of balances associated with an address anyway c) it made sense to me that since an address can have a set of balances that `AddressBalances` is just a map from addresses to `Balances` structs. 

If we implemented a nested map, we would still need locking for each of the internal maps - it would end up looking a lot like the `Balances` implementation.

> Currently we have Balance -> AddressBalances -> Balance -> balance
Balances -> AddressBalances -> balance seems clearer

The main reason we have a `Balance` type is to abstract the internal bookkeeping of the `AddressBalances` type away from the caller - in this case the `BroadcastSession`. From the `BroadcastSession`'s POV, it is just interacting with a single balance associated.

kyriediculous (approved):
> From the BroadcastSession's POV, it is just interacting with a single balance associated.

Ok that sounds reasonable ! 

LGTM 👍

kyriediculous (changes requested):
There seems to be a conflict with `go.sum` can we resolve this and make sure it still works with the updated deps?

yondonfu:
Rebased and resolved the conflict with `go.sum`

kyriediculous:
LGTM 👍  was awaiting CI checks

kyriediculous (approved):

wohlner:

PR #1132, opened 2019-10-23
Check for supported network when node starts

### LINKED ISSUE #1092: Check for (un-)supported networks
Currently, the go client is not compatible with mainnet or Rinkeby.

* We should print a friendly error message and exit if the user attempts to connect to an unsupported network
* We should have a process for enabling support for those networks once the contracts have been updated
* We should have a process for *disabling* support for these networks during development, if necessary (subject to our commitment to supporting mainnet/Rinkeby for non-release builds)

* Check if the current on-chain mode is supported
  * How do do this is an open question.
  * This could be a naive hard-coded check against supported network names.
  * This could be the git hash of generated bindings compared against the deployed git hash
  * Is there another way to get ordered versions, eg semver? (Should we even bother since this implies some degree of back compatibility assurance?)
* Print error message and exit if unsupported
* Add a flag to override any compatibility check, or to set the supported version at runtime (eg, for internally deployed test networks)
* Perhaps define a release process that toggles a default compatibility flag at build time, eg via `go build -ldflags="-X SupportedNetworks=..."`, similar as we have for [goclient versioning](https://github.com/livepeer/go-livepeer/blob/82024b4f902f959f09a625a924046a69dee144aa/Makefile#L12). This could be kicked off by a specific build target such as  `make release` .

We can make the check entirely dependent on internal variables such as the contract's git commit hash, which would allow us to avoid a release build step.

We can also hard-code supported networks to avoid a release build step. However, that means that we will be implicitly supporting these networks during subsequent development, unless we revert the hardcoded support. In a sense, this is how our process works right now: the goclient currently assumes itself to be mainnet-compatible, since it was mainnet-compatible at one point, and we never disabled mainnet compatibility. The tradeoff here is that our compatibility story remains subject to our vigilance around actually resetting that compatibility check, or our communication around compatibility of unreleased code.

**Additional questions**

* Is checking the git commit hash of the deployed contracts against the locally generated contracts enough? Is that the only thing we need to check for?

* Seems we could still want some checking for off-chain changes, including those related to off-chain protocol handling. Eg, suppose we make changes to how payment accounting is handled but haven't yet implemented compatibility measures. Naturally we don't want this to run on mainnet until a formal release is made. 

* Do we need to establish a rule on maintaining back-compat with various dependencies (protocol, Wowza, etc) during development? This has the implication that half-finished features (eg, PRs that only implement part of a certain feature) may need to be gated behind feature flags, and we may need to add back-testing for the "current" version of any dependencies to avoid breakage in addition to the "updated" versions we're developing against.

yondonfu:

This PR adds a check when the node starts to determine if the chain to connect to is supported. If the chain is not supported, the node will log an error and exit. Supported chains are configured when the node is built using the `HIGHEST_CHAIN_TAG` build tag. 

The order of supported chains is:

`everything else (i.e. dev chains) -> rinkeby -> mainnet`

All chains less than or equal to `HIGHEST_CHAIN_TAG` in the ordering will be supported by the node.

For example:

A node built using `make livepeer` will only connect to a dev chain. It will not connect with Rinkeby or mainnet.

A node built using `HIGHEST_CHAIN_TAG=rinkeby make livepeer` will connect to Rinkeby or a dev chain. It will not connect with mainnet.

A node build using `HIGHEST_CHAIN_TAG=mainnet make livepeer` will connect to any chain.

Since the highest chain tag is configured at build time, supported chains can be configured for official releases (i.e. testnet only releases, mainnet releases, etc.). Developers can always build the node from scratch and configure that tag by themselves.

- be2ba8b: Adds a `build` package with build specific code. Currently, `chain_mainnet.go`, `chain_rinkeby.go` and `chain_dev.go` use build constraints to determine how the constant `HighestChain` is set based on the build tags used
- cdad028: Use the `HIGHEST_CHAIN_TAG` env variable in the Makefile when building `livepeer` and `livepeer_cli`
- 380730b: Add a check on node startup to determine if the chain is supported

The `build` package could also be used to implement build tag configurable feature/experimental flags in the future.

Built `livepeer` with different values for `HIGHEST_CHAIN_TAG`.

Note: `livepeer_cli` doesn't actually check for supported chains right now despite being built with the `HIGHEST_CHAIN_TAG` because the node is not yet compatible with the contracts on Rinkeby so the node will pass the supported chain check, but exit soon thereafter because it cannot query the contracts. Since `livepeer_cli` relies on the node's web server, it would be a bit tough to test if `livepeer_cli` exits properly when it does not support the chain that the node is connected to. We can add the supported chains check in `livepeer_cli` when the node is actually compatible with Rinkeby since testing will be easier then.

Fixes #1092 

**Checklist:**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,11 +9,11 @@ version=$(shell cat VERSION)
 
 .PHONY: livepeer
 livepeer:
-	GO111MODULE=on go build -ldflags="-X github.com/livepeer/go-livepeer/core.LivepeerVersion=$(version)-$(shell git describe --always --long --dirty --abbrev=8)" cmd/livepeer/*.go
+	GO111MODULE=on go build -tags "$(HIGHEST_CHAIN_TAG)" -ldflags="-X github.com/livepeer/go-livepeer/core.LivepeerVersion=$(version)-$(shell git describe --always --long --dirty --abbrev=8)" cmd/livepeer/*.go
 
 .PHONY: livepeer_cli
 livepeer_cli:
-	GO111MODULE=on go build -ldflags="-X github.com/livepeer/go-livepeer/core.LivepeerVersion=$(version)-$(shell git describe --always --long --dirty --abbrev=8)" cmd/livepeer_cli/*.go
+	GO111MODULE=on go build -tags "$(HIGHEST_CHAIN_TAG)" -ldflags="-X github.com/livepeer/go-livepeer/core.LivepeerVersion=$(version)-$(shell git describe --always --long --dirty --abbrev=8)" cmd/livepeer_cli/*.go
 
 .PHONY: localdocker
 localdocker:
--- a/build/chain.go
+++ b/build/chain.go
@@ -1,0 +1,27 @@
+package build
+
+// SupportedChains is an enum that indicates the chains supported by the node
+// All chains represented by an enum value less than or equal than this value are supported
+// All chains represented by an enum value greater than this value are not supported
+type SupportedChains int
+
+const (
+	// Dev is a development chain
+	Dev SupportedChains = iota
+	// Rinkeby is the Ethereum Rinkeby test network chain
+	Rinkeby
+	// Mainnet is the Ethereum main network chain
+	Mainnet
+)
+
+// ChainSupported returns whether the node can connect to the chain with the given ID
+func ChainSupported(chainID int64) bool {
+	switch chainID {
+	case 4:
+		return Rinkeby <= HighestChain
+	case 1:
+		return Mainnet <= HighestChain
+	default:
+		return Dev <= HighestChain
+	}
+}
--- a/build/chain_dev.go
+++ b/build/chain_dev.go
@@ -1,0 +1,5 @@
+// +build !mainnet,!rinkeby
+
+package build
+
+const HighestChain = Dev
--- a/build/chain_mainnet.go
+++ b/build/chain_mainnet.go
@@ -1,0 +1,5 @@
+// +build mainnet
+
+package build
+
+const HighestChain = Mainnet
--- a/build/chain_rinkeby.go
+++ b/build/chain_rinkeby.go
@@ -1,0 +1,5 @@
+// +build rinkeby
+
+package build
+
+const HighestChain = Rinkeby
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -22,6 +22,7 @@ import (
 	"time"
 
 	"github.com/ethereum/go-ethereum/rpc"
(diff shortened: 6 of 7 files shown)


kyriediculous:
I think it might be interesting to also add a sanity check here in the future that checks if the given network name actually matches the network ID that the node is connected to.

Most often access to remote ethereum nodes would be through the hardcoded infura providers which is sufficient for an MVP, but I imagine quite a few orchestrators would actually likely be running their own infrastructure (since we are targetting mining operators as well). 

I can see potential user errors by specifying a network name as well as an `ethUrl` where the network name actually doesn't match the network `ethUrl` is pointing to. 

Another option I see is using a truffle-like config file with a `networks` field. Which would contain network names, hosts and acceptable chain ID's. (I know we've talked about a general YML file for configuring flags)

While I think being able to have seperate builds for seperate networks is a nice characteristic, I think it would be even nice to have network agnostic builds if possible. Perhaps that CLI rework could use a shove up the priority list after all? 

I'm also bringing this up because I'm working on https://github.com/livepeer/go-livepeer/issues/1098 , where I like the alternative of being able to support multiple networks without having to explicitly change `dataDir` through flags for each seperate network. We'd base generated files such as the DB off of the network name, but in my opinion using the chainID (perhaps in combo with network name) would be more robust.

yondonfu:
> While I think being able to have seperate builds for seperate networks is a nice characteristic, I think it would be even nice to have network agnostic builds if possible

The idea here isn't to have separate builds for separate networks, but rather to define the range of networks supported by a build. Network support in the node is disabled until the node is actually capable of running with it (as mentioned in the OP this same build process can also be useful to implement feature/experimental flags) So, for non-official releases, there is a way to build the node such that it will automatically exit if tries connecting to a network that it is not yet compatible with. The development lifecycle would be:

- Work on development networks
- Make the node compatible with Rinkeby - at this point, the node might not be ready to connect with mainnet yet, so testnet only releases would be built with `HIGHEST_SUPPORTED_NETWORK=rinkeby`
- Make the node compatible with mainnet - at this point, the node could connect to any network 

> We'd base generated files such as the DB off of the network name, but in my opinion using the chainID (perhaps in combo with network name) would be more robust.

Yep agreed on using chainIDs. I can update this PR to do the check based on a chainID. I'll probably branch off of #1125 since that PR updates the go-ethereum version and the `ethclient.ChainID()` method is only introduced in a later go-ethereum version.

yondonfu:
Pushed new commits to use the chainID fetched from a ETH node to determine if the Livepeer node supports the chain. Updated the OP accordingly.

kyriediculous (approved):
LBTM 👍

PR #1133, opened 2019-10-23
cmd, eth: use blockPollingInterval flag

### LINKED ISSUE #1128: Configurable block time polling interval
Currently, there are a few services that the node runs involving polling at regular intervals in order to query the blockchain. Examples include:

- Polling for new blocks added to the head of the chain (https://github.com/livepeer/go-livepeer/blob/master/cmd/livepeer/livepeer.go#L51)
- Polling to check if the current round should be initialized (https://github.com/livepeer/go-livepeer/blob/master/cmd/livepeer/livepeer.go#L56)
- Polling to check if the node should call reward (https://github.com/livepeer/go-livepeer/blob/master/eth/eventservices/rewardservice.go#L17)
- Polling to cache the latest suggested gas price (https://github.com/livepeer/go-livepeer/blob/master/cmd/livepeer/livepeer.go#L63)

In all of the provided examples, it would be sufficient to poll every time a new block is added to the head of the chain instead of having different polling intervals for each service. Since the node could be connected to different networks (i.e. mainnet, Rinkeby, internal testnet, etc.) and each of the networks could have a different average block time, the block time polling interval should be configurable.

We can:

- Add a `-blockPollingInterval` flag with the value denominated in seconds. The default value can be 5s which is middle ground between fast block times (i.e. 1 second) that might be used in local testing and slower block times (i.e. ~15 seconds) that would be used on mainnet or Rinkeby
- Use the value for `-blockPollingInterval` to configure all services that need to query the blockchain at a regular interval

An alternative to adding a `-blockPollingInterval` flag would be to hardcode different values for different networks, but using the flag would be more flexible.

kyriediculous:
This PR removes different hardcoded polling intervals used for different services that query the blockchain periodically. Examples include `gasPriceMonitor`, `blockWatcher`,  `rewardService`, ... 
- added a new flag `blockPollingInterval` that takes in an integer to indicate the polling interval in seconds
- remove all seperate hardcoded polling intervals in livepeer.go
- add a `pollingInterval` field to `RewardService` and set it to the `blockPollingInterval` when invoking the constructor

Ran test-harness
Ran unit tests

Fixes #1128 

**Checklist:**

### DIFF
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -46,21 +46,14 @@ var (
 
 	// The timeout for ETH RPC calls
 	ethRPCTimeout = 20 * time.Second
-
-	// The interval at which the block watcher polls for new blocks
-	blockWatcherPollingInterval = 1 * time.Second
 	// The maximum block sfor the block watcher to retain
 	blockWatcherRetentionLimit = 20
-
-	// The interval at which the round initializer runs
-	roundInitPollingInterval = 10 * time.Second
 
 	// The gas required to redeem a PM ticket
 	redeemGas = 100000
 	// The multiplier on the transaction cost to use for PM ticket faceValue
 	txCostMultiplier = 100
-	// The interval at which to poll for gas price updates
-	gpmPollingInterval = 1 * time.Minute
+
 	// The interval at which to clean up cached max float values for PM senders and balances per stream
 	cleanupInterval = 1 * time.Minute
 	// The time to live for cached max float values for PM senders (else they will be cleaned up) in seconds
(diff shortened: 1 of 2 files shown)


cmd/livepeer/livepeer.go
```diff
@@ -118,14 +111,14 @@ func main() {
...
-
 	// Orchestrator base pricing info
 	pricePerUnit := flag.Int("pricePerUnit", 0, "The price per 'pixelsPerUnit' amount pixels")
 	// Broadcaster max acceptable price
 	maxPricePerUnit := flag.Int("maxPricePerUnit", 0, "The maximum transcoding price (in wei) per 'pixelsPerUnit' a broadcaster is willing to accept. If not set explicitly, broadcaster is willing to accept ANY price")
 	// Unit of pixels for both O's basePriceInfo and B's MaxBroadcastPrice
 	pixelsPerUnit := flag.Int("pixelsPerUnit", 1, "Amount of pixels per unit. Set to '> 1' to have smaller price granularity than 1 wei / pixel")
-
+	// Interval to poll for blocks
+	blockPollTime := flag.Int("blockPollingInterval", 5, "Interval in seconds at which different blockchain event services poll for blocks")
```
yondonfu: Can we use the same name for the variable and the flag to follow the convention used in the rest of the flags?

yondonfu (changes requested):
Looks good! Just one minor variable naming comment.

yondonfu (commented):
Changes look good! Let's rebase.

kyriediculous:
rebased

yondonfu (approved):
LGTM!

wohlner:

PR #1134, opened 2019-10-24
server: Fix /currentRound handler

yondonfu:

This PR fixes the `/currentRound` handler in the webserver. Previously, it was converting a `big.Int` into a `string` and then casting it into a `[]byte`. Then, `livepeer_cli` would use those bytes to create `big.Int`. The act of converting the `big.Int` into a `string` and *then* casting it into a `[]byte` results in a different `[]byte` representation and as a result, when `livepeer_cli` converts the response from the server into a `big.Int`, we end up with an incorrect value for current round.

- Directly return the byte representation of the current round in the `/currentRound` handler
- Refactor the `/currentRound` handler so it is easier to write unit tests for
- Add unit tests for the `currentRoundHandler()`

Added unit tests. Ran the a node using the devtool script to confirm that `livepeer_cli` prints the correct current round.

**Checklist:**

### DIFF
--- a/server/handlers_test.go
+++ b/server/handlers_test.go
@@ -145,6 +145,43 @@ func TestCurrentBlockHandler_Success(t *testing.T) {
 	assert.Equal(http.StatusOK, resp.StatusCode)
 	assert.Equal(big.NewInt(50), new(big.Int).SetBytes(body))
 }
+func TestCurrentRoundHandler(t *testing.T) {
+	assert := assert.New(t)
+
+	// Test missing client
+	handler := currentRoundHandler(nil)
+
+	resp := httpGetResp(handler)
+	defer resp.Body.Close()
+	body, _ := ioutil.ReadAll(resp.Body)
+
+	assert.Equal(http.StatusInternalServerError, resp.StatusCode)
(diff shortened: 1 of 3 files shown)


server/handlers_test.go
```diff
@@ -145,6 +145,47 @@ func TestCurrentBlockHandler_Success(t *testing.T) {
...
+
+	resp = httpGetResp(handler)
+	defer resp.Body.Close()
+	body, _ = ioutil.ReadAll(resp.Body)
+
+	assert.Equal(http.StatusOK, resp.StatusCode)
+	assert.Equal(currentRound, new(big.Int).SetBytes(body))
+}
+
+func TestCurrentRoundHandler_Success(t *testing.T) {
```
kyriediculous: no function body here, can we remove this?

kyriediculous (changes requested):
Looks good, just a minor comment

server/handlers_test.go
yondonfu: Oops forgot to remove that. Fixed in 13e8d32

kyriediculous:
Let's rebase !

yondonfu:
Rebased

kyriediculous (approved):
LGTM :ship:

PR #1135, opened 2019-10-24
common, db: add network id check to prevent datadir collisions

### LINKED ISSUE #1098: Prevent datadir reuse on different networks

kyriediculous:
Prevents further datadir collisions by checking the chainID from the provided remote ethereum node against the chainID in the database when the datadir and node were first initiated. 

This prevents the edge case where a node operator changes network without changing the network name or connecting to a network URL that doesn't fit the name. 

- Introduced db.NetworkID
- Introduced db.SelectKV
- Introduced db.UpdateKV
- save & compare chainID when starting the node

unit tests

Fixes #1098 

**Checklist:**

TODO: still uses `backend.NetworkID` , swap it out for `backend.ChainID`

### DIFF
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -195,9 +195,9 @@ func main() {
 		if *ethController == "" {
 			*ethController = netw.ethController
 		}
-		glog.Infof("***Livepeer is running on the %v*** network: %v***", *network, *ethController)
+		glog.Infof("***Livepeer is running on the %v network: %v***", *network, *ethController)
 	} else {
-		glog.Infof("***Livepeer is running on the %v*** network", *network)
+		glog.Infof("***Livepeer is running on the %v network***", *network)
 	}
 
 	if *datadir == "" {
@@ -288,6 +288,12 @@ func main() {
 	watcherErr := make(chan error)
 	if *network == "offchain" {
 		glog.Infof("***Livepeer is in off-chain mode***")
+
+		if err := checkOrStoreChainID(dbh, big.NewInt(0)); err != nil {
+			glog.Error(err)
+			return
+		}
+
 	} else {
 		var keystoreDir string
 		if _, err := os.Stat(*ethKeystorePath); !os.IsNotExist(err) {
@@ -311,6 +317,16 @@ func main() {
 		backend, err := ethclient.Dial(*ethUrl)
 		if err != nil {
 			glog.Errorf("Failed to connect to Ethereum client: %v", err)
+			return
+		}
+
+		chainID, err := backend.ChainID(context.Background())
+		if err != nil {
+			glog.Errorf("failed to get chain ID from remote ethereum node: %v", err)
+			return
+		}
+		if err := checkOrStoreChainID(dbh, chainID); err != nil {
+			glog.Error(err)
 			return
 		}
 
@@ -814,4 +830,27 @@ func defaultAddr(addr, defaultHost, defaultPort string) string {
 		return addr + ":" + defaultPort
 	}
 	return addr
+}
+
+func checkOrStoreChainID(dbh *common.DB, chainID *big.Int) error {
+	expectedChainID, err := dbh.ChainID()
+	if err != nil {
+		return err
+	}
+
+	if expectedChainID == nil {
+		// No chainID stored yet
+		// Store the provided chainID and skip the check
+		if err := dbh.SetChainID(chainID); err != nil {
+			return err
+		}
+
+		return nil
+	}
+
+	if expectedChainID.Cmp(chainID) != 0 {
+		return fmt.Errorf("expecting chainID of %v, but got %v. Did you change networks without changing network name or datadir?", expectedChainID, chainID)
+	}
+
+	return nil
 }
--- a/common/db.go
+++ b/common/db.go
@@ -25,6 +25,7 @@ type DB struct {
 	// prepared statements
 	selectOrchs                      *sql.Stmt
 	updateOrch                       *sql.Stmt
+	selectKV                         *sql.Stmt
 	updateKV                         *sql.Stmt
 	insertUnbondingLock              *sql.Stmt
 	deleteUnbondingLock              *sql.Stmt
@@ -169,8 +170,17 @@ func InitDB(dbPath string) (*DB, error) {
 	}
 	d.updateOrch = stmt
 
+	// selectKV prepared statement
+	stmt, err = db.Prepare("SELECT value FROM kv WHERE key=?")
+	if err != nil {
+		glog.Error("Unable to prepare selectKV stmt", err)
+		d.Close()
+		return nil, err
+	}
+	d.selectKV = stmt
+
 	// updateKV prepared statement
-	stmt, err = db.Prepare("UPDATE kv SET value=?, updatedAt = datetime() WHERE key=?")
+	stmt, err = db.Prepare("INSERT OR REPLACE INTO kv(key, value, updatedAt) VALUES(?1, ?2, datetime())")
 	if err != nil {
 		glog.Error("Unable to prepare updatekv stmt ", err)
 		d.Close()
@@ -272,6 +282,9 @@ func (db *DB) Close() {
 	if db.selectOrchs != nil {
 		db.selectOrchs.Close()
 	}
+	if db.selectKV != nil {
+		db.selectKV.Close()
+	}
 	if db.updateKV != nil {
 		db.updateKV.Close()
 	}
@@ -321,6 +334,52 @@ func (db *DB) LastSeenBlock() (*big.Int, error) {
 	}
 
 	return header.Number, nil
+}
+
+func (db *DB) ChainID() (*big.Int, error) {
+	idString, err := db.selectKVStore("chainID")
+	if err != nil {
+		return nil, err
+	}
+
+	if idString == "" {
+		return nil, nil
+	}
+
+	id, ok := new(big.Int).SetString(idString, 10)
+	if !ok {
+		return nil, fmt.Errorf("unable to convert chainID string to big.Int")
+	}
+
+	return id, nil
+}
+
+func (db *DB) SetChainID(id *big.Int) error {
+	if err := db.updateKVStore("chainID", id.String()); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (db *DB) selectKVStore(key string) (string, error) {
+	row := db.selectKV.QueryRow(key)
+	var valueString string
+	if err := row.Scan(&valueString); err != nil {
+		if err.Error() != "sql: no rows in result set" {
+			return "", fmt.Errorf("could not retrieve key from database: %v", err)
+		}
+		// If there is no result return no error, just zero value
+		return "", nil
+	}
+	return valueString, nil
+}
+
+func (db *DB) updateKVStore(key, value string) error {
+	_, err := db.updateKV.Exec(key, value)
+	if err != nil {
+		glog.Errorf("db: Unable to update %v in database: %v", key, err)
+	}
+	return err
 }
 
 func (db *DB) UpdateOrch(orch *DBOrch) error {
--- a/common/db_test.go
+++ b/common/db_test.go
@@ -19,6 +19,89 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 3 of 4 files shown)


cmd/livepeer/livepeer.go
```diff
@@ -293,6 +293,20 @@ func main() {
 	watcherErr := make(chan error)
 	if *network == "offchain" {
 		glog.Infof("***Livepeer is in off-chain mode***")
+
+		expectedChainID, err := dbh.NetworkID()
```
yondonfu: I think we can reduce the amount of duplication in the off-chain vs. on-chain case by extracting some logic into a helper function. Perhaps something like this could work:

```
func checkOrStoreChainID(dbh *common.DB, chainID *big.Int) error {
    expectedChainID, err := dbh.ChainID()
    if err != nil {
        return err
    }
    if expectedChainID == nil {
        // No chainID stored yet
        // Store the provided chainID and skip the check
        if err := dbh.SetChainID(chainID); err != nil {
            return err
        }

        return nil
    }

    if expectedChainID.Cmp(chainID) != 0 {
        // Return error
    }

    return nil
}
```

Then in the off-chain case we can simply pass `big.NewInt(0)` into `checkChainID()` as the chainID and in the on-chain case we can fetch the remote chainID and pass it into `checkChainID()`.

What do you think?

cmd/livepeer/livepeer.go
```diff
@@ -293,6 +293,20 @@ func main() {
...
+		if err != nil {
+			glog.Errorf("failed to get chain ID from database: %v", err)
+			return
+		}
+		if expectedChainID == nil {
+			dbh.UpdateKV("network", "0")
+		}
+		if expectedChainID != nil && expectedChainID.Int64() != 0 {
+			glog.Errorf("Livepeer running in offchain mode, expecting chainID of 0 but got %v. Did you change networks withour changing the network name or datadir?", expectedChainID)
```
yondonfu: "withour" -> "without"

Also, I feel like we can shorten this message a bit an omit "Livepeer" - what do you think?

cmd/livepeer/livepeer.go
```diff
@@ -319,6 +333,28 @@ func main() {
...
+		}
+		chainID, err := backend.NetworkID(context.Background())
+		if err != nil {
+			glog.Errorf("failed to get chain ID from remote ethereum node: %v", err)
+			return
+		}
+		if expectedChainID != nil {
+			if expectedChainID.Cmp(chainID) != 0 {
+				glog.Errorf(
+					"ethereum node provided chain ID %v for network '%v', but we expected %v. Did you change networks without changing the network name or dataDir?",
```
yondonfu: "dataDir" -> "datadir" to match error message defined above.

The ordering of the arguments seems off - I think it should be `chainID`, `*network` and then `expectedChainID`

common/db.go
```diff
@@ -169,8 +170,18 @@ func InitDB(dbPath string) (*DB, error) {
...
+	if err != nil {
+		glog.Error("Unable to prepare selectKV stmt", err)
+		d.Close()
+		return nil, err
+	}
+	d.selectKV = stmt
+
 	// updateKV prepared statement
-	stmt, err = db.Prepare("UPDATE kv SET value=?, updatedAt = datetime() WHERE key=?")
+	// stmt, err = db.Prepare("UPDATE kv SET value=?2, updatedAt = datetime() VALUES(?1, ?2) WHERE key=?1")
```
yondonfu: Can remove comment

common/db.go
```diff
@@ -323,6 +337,45 @@ func (db *DB) LastSeenBlock() (*big.Int, error) {
...
+
+	id, ok := new(big.Int).SetString(idString, 10)
+	if !ok {
+		return nil, fmt.Errorf("unable to convert networkID string to big.Int")
+	}
+
+	return id, nil
+}
+
+func (db *DB) SelectKV(key string) (string, error) {
```
yondonfu: Hm while I see why exporting these methods could be useful, I feel like this is leaking details of the DB to the caller (i.e. that we're using a KV table) and am wary of exposing it/encouraging storing/fetching KV pairs with the caller passing in string keys every time they need to call these methods - feels a bit fragile. Isolating the use of string keys to query the underlying KV table in this file feels safer IMO.

yondonfu (changes requested):

common/db.go
kyriediculous: So in other words we'd only export usage specific methods such as `ChainID()` & `SetChainID()` ?

common/db.go
kyriediculous: Yeah I was trying to do an UPDATE statement with VALUES keyword but I couldn't get it to work so I went for INSERT OR REPLACE instead.

cmd/livepeer/livepeer.go
kyriediculous: Yeah sounds good !

cmd/livepeer/livepeer.go
kyriediculous: Yeah caught my eye as well, but restructuring the way we log the error when using a helper as you described above

kyriediculous:
Fixed the requested changes as well as failing test in 5821b09

Also added a test in `./test-args.sh` that checks for an error thrown on datadir collisions when using a custom datadir with a different network. 

Want me to add a test that checks for datadir collissions by using the same network name but a different ID? This would probably require passing in an -ethUrl flag so we can simulate a rinkeby connection with a mainnet chainID

cmd/livepeer/livepeer.go
```diff
@@ -293,7 +293,15 @@ func main() {
...
 	if *network == "offchain" {
 		glog.Infof("***Livepeer is in off-chain mode***")
+
+		if err := checkOrStoreChainID(dbh, big.NewInt(0)); err != nil {
+			glog.Error(err)
+			return
+		}
+
 	} else {
+		glog.Infof("***Livepeer is connected to the %v network***", *network)
```
yondonfu: I don't think we need this log statement since we already log something similar earlier in the file.

common/db.go
yondonfu: UPDATE is for updating existing records so INSERT OR REPLACE is actually a better fit here since we don't have a default value for `chainID`.

Bump on removing comment.

common/db_test.go
```diff
@@ -19,6 +19,87 @@ import (
...
+	row := dbraw.QueryRow("SELECT value FROM kv WHERE key = 'chainID'")
+	err = row.Scan(&networkID)
+	assert.EqualError(err, "sql: no rows in result set")
+	assert.Equal("", networkID)
+
+	dbh.updateKVStore("chainID", expectedChainID)
+	row = dbraw.QueryRow("SELECT value FROM kv WHERE key = 'chainID'")
+	err = row.Scan(&networkID)
+	assert.Nil(err)
+	assert.Equal(expectedChainID, networkID)
```
yondonfu: Let's try to be consistent about using chainID based naming instead of networkID since they technically are different.

common/db_test.go
```diff
@@ -19,6 +19,87 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+func TestUpdateKV_Network(t *testing.T) {
```
yondonfu: Should the name of this test be `TestUpdateKVStore`?

common/db_test.go
```diff
@@ -19,6 +19,87 @@ import (
...
+	assert.Equal("", networkID)
+
+	dbh.updateKVStore("chainID", expectedChainID)
+	row = dbraw.QueryRow("SELECT value FROM kv WHERE key = 'chainID'")
+	err = row.Scan(&networkID)
+	assert.Nil(err)
+	assert.Equal(expectedChainID, networkID)
+}
+
+func TestSelectKV(t *testing.T) {
```
yondonfu: Should the name of this test be `TestSelectKVStore`?

common/db_test.go
```diff
@@ -19,6 +19,87 @@ import (
...
+	assert := assert.New(t)
+	expectedChainID := "1337"
+
+	dbh, dbraw, err := TempDB(t)
+	require.Nil(err)
+
+	defer dbh.Close()
+	defer dbraw.Close()
+
+	dbh.updateKVStore("chainID", expectedChainID)
```
yondonfu: Can use `SetChainID()` here?

cmd/livepeer/livepeer.go
```diff
@@ -820,3 +838,25 @@ func defaultAddr(addr, defaultHost, defaultPort string) string {
...
+		// Store the provided chainID and skip the check
+		if err := dbh.SetChainID(chainID); err != nil {
+			return err
+		}
+
+		return nil
+	}
+
+	if expectedChainID.Cmp(chainID) != 0 {
+		return fmt.Errorf("expecting chainID of %v, but got %v. Did you change networks without changing network name or datadir?")
```
yondonfu: Missing `expectedChainID` and `chainID` as parameters to the error

test_args.sh
```diff
@@ -67,11 +67,17 @@ run_lp -broadcaster -datadir "$CUSTOM_DATADIR"
...
 
+CUSTOM_DATADIR="$TMPDIR"/customDatadir2
+
 # check custom datadir with a network
 run_lp -broadcaster -datadir "$CUSTOM_DATADIR" -network rinkeby $ETH_ARGS
 [ ! -d  "$CUSTOM_DATADIR"/rinkeby ] # sanity check that network isn't included
 kill $pid
 
+# expect error because re-use network name with different chainID
+run_lp -broadcaster -datadir "$CUSTOM_DATADIR" -network mainnet $ETH_ARGS || res=$?
```
yondonfu: I ran this locally and I think you may need to increase SLEEP to something like .5 or else the DB is not initialized in the previous command before the process is killed

yondonfu (changes requested):

common/db.go
yondonfu: Yep

yondonfu:
> Want me to add a test that checks for datadir collissions by using the same network name but a different ID?

I would suggest just adding a test for when the node has an off-chain DB and connects to a ETH network since that is a distinct case from when the node has an on-chain DB and connects to a ETH network (which we have a test for).

common/db_test.go
```diff
@@ -19,6 +19,89 @@ import (
...
+	assert := assert.New(t)
+	expectedChainID := "1337"
+
+	dbh, dbraw, err := TempDB(t)
+	require.Nil(err)
+
+	defer dbh.Close()
+	defer dbraw.Close()
+
+	var networkID string
```
yondonfu: Let's call this chainID

test_args.sh
```diff
@@ -19,7 +19,7 @@ cat > $TMPDIR/key <<ETH_KEY
 ETH_KEY
 
 ETH_ARGS="-ethKeystorePath $TMPDIR/key"
-SLEEP=0.25
+SLEEP=0.5
```
yondonfu: Since we're not adding any additional tests in this file, can we just revert all changes? Are any of these changes necessary?

yondonfu (changes requested):

test_args.sh
kyriediculous: Oh sorry I forgot to change `SLEEP` back

We need the re-assignment of `CUSTOM_DATA_DIR` else we'll get an error saying that we're expecting a different chain ID

yondonfu:
Once #1125 is merged to master, let's rebase so we can pull in the latest go-ethereum version and use `ChainID()` instead of `NetworkID()`

kyriediculous:
rebased !

yondonfu (approved):
LGTM

PR #1136, opened 2019-10-25
cmd: Update Rinkeby network config for Streamflow

yondonfu:

This PR updates the default Controller contract address when connecting to Rinkeby to the new Streamflow address and updates the default RPC URL for Infura to just use HTTP (since we're polling for blocks and not using websockets for event subscriptions).

- Updated default Controller address on Rinkeby
- Updated default RPC URL on Rinkeby

Tried connecting to the Rinkeby contracts after building the node.

**Checklist:**


kyriediculous (approved):
LGTM 🚀

PR #1137, opened 2019-10-28
core: Check for nil balance in Orchestrator.SufficientBalance()

yondonfu:

This PR fixes a bug where a nil pointer error is triggered by `Orchestrator.SufficientBalance()` when an orchestrator does not have a cached balance for a stream and does not receive a payment with a segment so it does not credit the stream's balance (the first credit operation for a stream's balance will create a cached balance if it does not already exist).

The bug can be replicated with the following steps:

1. Build the node using the latest `master`
2. Run `docker run -d -p 8545:8545 -p 8546:8546 --name geth-with-livepeer-protocol livepeer/geth-with-livepeer-protocol:streamflow
3. Setup an O using the devtool
4. Setup a B using the devtool
5. Make sure that B can connect to O
6. Stream into B and observe O start to transcode
7. Restart O
8. Stream into B again and observe O encounter errors while transcoding

After step 7, O will start without any cached balances so O's view of a stream's balance (for a particular manifestID) is out-of-sync with B's view of a stream's balance. As a result, B will not send a payment with the first segment in step 8.

This is a screenshot of the error encountered in step 8:
Screen Shot 2019-10-28 at 10 43 39 AM

This PR does NOT solve the stream balance synchronization issue between B & O when either node restarts and then reconnects with each other shortly afterwords. This PR only fixes the nil pointer error that O encounters in this scenario. In the example scenario mentioned above, in step 8, B will encounter an insufficient balance error and then if reconnects with O again it afterwords it will start off with a stream balance of 0. We might be able to eventually move away from stream balance tracking if we implement short circuited transcoding (which we're already interested in implementing and have already discussed here and there) along with per segment pixel limits (will write up a separate GH issue about this with more details), but we can follow up on that separately.

- Check if the balance for a stream is nil in `Orchestrator.SufficientBalance()`. If the balance is nil, return false

Added a unit test for the update.

**Checklist:**

### DIFF
--- a/core/orch_test.go
+++ b/core/orch_test.go
@@ -1108,8 +1108,7 @@ func TestSufficientBalance_IsSufficient_ReturnsTrue(t *testing.T) {
 	recipient.On("ReceiveTicket", mock.Anything, mock.Anything, mock.Anything).Return("", false, nil).Once()
 	assert := assert.New(t)
 
-	// Create a ticket where faceVal = EV
-	// making faceVal the expected balance increase
+	// Create a ticket where faceVal = EV so that the balance = EV
 	payment := defaultPayment(t)
 	payment.TicketParams.FaceValue = big.NewInt(100).Bytes()
 	payment.TicketParams.WinProb = new(big.Int).Sub(new(big.Int).Lsh(big.NewInt(1), 256), big.NewInt(1)).Bytes()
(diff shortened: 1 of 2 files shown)


kyriediculous (approved):
LGTM 👍

PR #1140, opened 2019-10-28
clean "tagged" release versions

iameli:
This changeset allows us to release "named" versions of go-livepeer where the version string doesn't contain the commit SHA. Made for the upcoming v0.5.0 release.

Also cauterizes the Windows build for now as it's broken and not a priority for anything.

Also adds `HIGHEST_CHAIN_TAG` to the Linux and Mac CI processes to create binaries that support the Rinkeby chain.

### DIFF
--- a/print_version.sh
+++ b/print_version.sh
@@ -14,8 +14,8 @@ DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 currentTag="$(git describe --tags)"
 currentVersion="$(cat "$DIR/VERSION")"
 currentSha="$(git describe --always --long --dirty --abbrev=8)"
-if [[ "$currentTag" == "$currentVersion" ]]; then
-  echo -en "$currentTag"
+if [[ "$currentTag" == "v$currentVersion" ]]; then
+  echo -en "$currentVersion"
 else
   echo -en "$currentVersion-$currentSha"
 fi
--- a/print_version.sh
+++ b/print_version.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+
+# Prints the current build version of go-livepeer. Based on the following considerations:
+# 1. If the VERSION file matches the tag of our current commit, then we're the canonical
+#    x.y.z version, and that will be our version string.
+# 2. Otherwise, our version string is x.y.z-SHA, provided by the VERSION file and
+#    git describe --always --long --abbrev=8 --dirty
+
+set -e
+set -o nounset
+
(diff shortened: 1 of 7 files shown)


yondonfu (approved):
LGTM!

print_version.sh
```diff
@@ -0,0 +1,21 @@
...
+
+set -e
+set -o nounset
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+currentTag="$(git describe --tags)"
+currentVersion="$(cat "$DIR/VERSION")"
+currentSha="$(git describe --always --long --dirty --abbrev=8)"
+if [[ "$currentTag" == "$currentVersion" ]]; then
```
j0sh: I guess there is a question of what our version tagging convention should be. Most projects have the convention of `v<number>`. So far, we have something of a mix:

```
go-livepeer$ git tag -l
0.0.1
0.0.2
0.0.3
0.1.0
0.1.1
0.1.10
0.1.11
0.1.12
0.1.13
0.1.14
0.1.15
0.1.16
0.1.17
0.1.2
0.1.3
0.1.4
0.1.5
0.1.6
0.1.7
0.1.8
0.1.9
0.2.0
0.2.1
0.2.2
0.2.3
0.2.4
0.3.0
0.3.1
0.3.2
0.4.0
v0.1.7
v0.1.7_test
v0.3.3
```

Personally, I have a slight preference for appending a `v` on there, mostly because it gives a bit more context as to what the tag is (following the convention used by other projects). If we do that, then this check should probably have a 'v' prefixed to it, be something like:

```
if [[ "$currentTag" == "v$currentVersion" ]]; then
```

yondonfu: I'm onboard with adding a "v" as a prefix to the version tag.

iameli: I don't care too much one way or another, as long as the `v` is just an artifact on the git tag and the actual version string remains without a v, e.g. `github.com/livepeer/go-livepeer/core.LivepeerVersion=0.3.3-8101daf3)`. I'll add that check.

iameli: changed to

```bash
if [[ "$currentTag" == "v$currentVersion" ]]; then
  echo -en "$currentVersion"
else
  echo -en "$currentVersion-$currentSha"
fi
```

iameli: The other place this should show up soon-ish is in the Docker builds, and those generally follow a vless convention, e.g. `livepeer/go-livepeer:0.5.0`

Issue #811, opened 2019-04-08
Document requirements: Webhooks, Specify Bitrate Ladder


ericxtang:
Deliverable is final requirements

iameli:
Sure!

Upon a stream starting, Livepeer broadcasts nodes currently perform an authentication webhook. Currently, all it receives back is a stream key. Per the work in https://github.com/livepeer/go-livepeer/issues/793, we'd like to return a variety of other structured data, including the desired bitrate ladder for the broadcast.

Per [discussion here](https://hackmd.io/Rb8Hk-sTThm2eR29DXIjzA), that ladder will look something like this.

```json
[
  {
    "width": 1920,
    "height": 1080,
    "bitrate": 6000000,
    "fps": 30
  },
  {
    "width": 1280,
    "height": 720,
    "bitrate": 6000000,
    "fps": 60
  },
  {
    "width": 853,
    "height": 480,
    "bitrate": 2000000,
    "fps": 30
  },
  {
    "width": 426,
    "height": 240,
    "bitrate": 700000,
    "fps": 30
  }
]
```

Or, expressed more formally as a (first draft) JSON-Schema:

```json
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "type": "array",
  "items": {
    "type": "object",
    "requiredProperties": ["width", "height", "bitrate", "fps"],
    "properties": {
      "width": {
        "type": "integer",
        "minimum": 128
      },
      "height": {
        "type": "integer",
        "minimum": 128
      },
      "bitrate": {
        "type": "integer",
        "minimum": 400000
      },
      "fps": {
        "type": "integer",
        "minimum": 1
      }
    }
  }
}
```

For now, these fields will be further locked down to just support [our hardcoded options](https://github.com/livepeer/lpms/blob/master/ffmpeg/videoprofile.go) to limit the surface area of the validation project, but this API will work for arbitrary settings when we get to that point.

The broadcaster must then use this data to interface with its orchestrators.

Open question for discussion: what happens if the API node returns illegal data? We can reduce the chances of this by shipping a cross-language JSON-Schema to ensure all the components are on the same page, but given the structure of a webhook there's not an obvious way for the broadcaster client to communicate an error state to the webhook server.

j0sh:
Looks nice. Presumably this list is put into the webhook response, eg

```
{"manifestID": "abc", "transcodingOptions": <list>}
```

>  these fields will be further locked down to just support our hardcoded options

[...]

PR #1148, opened 2019-10-31
adding 25fps

ericxtang:

Adding 25fps transcoding profile.  In combination with https://github.com/livepeer/lpms/pull/159

- Adding 25fps video profile

Manual testing

**Checklist:**

### DIFF
--- a/common/videoprofile_ids.go
+++ b/common/videoprofile_ids.go
@@ -12,13 +12,18 @@ type VideoProfileByteMap map[[VideoProfileIDBytes]byte]string
 var VideoProfileNameLookup = map[string]string{
 	"a7ac137a": "P720p60fps16x9",
 	"49d54ea9": "P720p30fps16x9",
+	"e4a64019": "P720p25fps16x9",
 	"79332fe7": "P720p30fps4x3",
 	"5ecf4b52": "P576p30fps16x9",
+	"8b1843d6": "P576p25fps16x9",
 	"93c717e7": "P360p30fps16x9",
+	"7cd40fc7": "P360p25fps16x9",
 	"b60382a0": "P360p30fps4x3",
 	"c0a6517a": "P240p30fps16x9",
+	"1301a7d0": "P240p25fps16x9",
 	"d435c53a": "P240p30fps4x3",
 	"fca40bf9": "P144p30fps16x9",
+	"03f01d1f": "P144p25fps16x9",
 }
 
 func makeVideoProfileByteMap() VideoProfileByteMap {


iameli:
Note that any new profiles also need to be added here to be used in the API: https://github.com/livepeer/livepeerjs/blob/master/packages/sdk/src/index.js#L23

j0sh:
To clarify the reason for adding these profiles: is it because we because we observed input coming in at 25fps, and this made the mediastreamvalidator complain about mismatched framerates in the master playlist vis-a-vis our 30fps output options?

This would imply that we need to compute the input frame rate prior to selecting a matching bitrate ladder for transcoding. This may not always be feasible for a number of reasons: 

* The input may be configured as VFR, or adjust mid-stream
* "Constant" fps input may end up being erratic due to drops, eg at the capture side
* The input itself may be an "odd" frame rate
* The actual framerate may not always be signaled and would sometimes need to be approximated.

There are a few other ways we can work around the issue and make mediastreamvalidator happy:

* Omit the source stream from the master playlist. Only make the transcoded renditions available. This would be 30/60fps all around.
* Use a FPS in a "pass-through" configuration that preserves the framerate of the input stream. See https://github.com/livepeer/lpms/issues/120

With all that being said, this type of patch feels quite ad-hoc, because it only addresses the one specific case we've observed, rather than generalizing the solution. That also points to some deeper issues with how we configure out output renditions: it may be time to move away from a hardcoded set of presets and enable full transcoder configurability throughout the network.

iameli:
> To clarify the reason for adding these profiles: is it because we because we observed input coming in at 25fps, and this made the mediastreamvalidator complain about mismatched framerates in the master playlist vis-a-vis our 30fps output options?

It's not about mediastreamvalidator - we have a specific customer ask for 25fps. The downstream client is the Wowza plugin, so we can lean on Wowza to provide us the source framerate information prior to picking a preset in the API.

I agree with everything you're saying about what we should do moving forward though.

> That also points to some deeper issues with how we configure out output renditions: it may be time to move away from a hardcoded set of presets and enable full transcoder configurability throughout the network.

That's https://github.com/livepeer/go-livepeer/issues/811, yeah?

ericxtang:
@j0sh and @iameli - I'm supportive for solving this problem by allowing users to dynamically specify the transcoder configuration.

j0sh:
> That's #811, yeah?

That's part of it - #811 primarily covers the webhook and is one possible interface for supplying the transcoding configuration. The RPC handling between B/O/T would need to be updated as well, including protobuf messages. Hopefully that's it, but there might be a couple other things I'm forgetting.

ericxtang:
@j0sh can you approve?

j0sh (approved):

Issue #1155, opened 2019-11-01
High memory usage


j0sh:
**Is your feature request related to a problem? Please describe.**
The go client allocates excessively, and it often takes a long time for GC to run and RAM usage (RSS) to even out - and RSS often evens out at somewhat concerning levels (eg, 75% of total system memory), even if immediate memory needs are nowhere as high.

This flame graph from a stress test exhibits the issue: `bytes.makeSlice` cumulatively accounts for almost 79% of all allocations, split across two different call stacks. Total allocations for `bytes.makeSlice` amount to over 332GB across a period of roughly 12 hours with 16 concurrent streams on a single OT node with 8 attached GPUs. This eats up over half the working RAM on a 14GB server.

These allocations are because we slurp up video data into memory as soon as it becomes available. The two call stacks where `bytes.MakeSlice` is dominant correspond to the points where we receive source inputs (`ServeSegment`) and produce transcoded outputs (`resToTranscodeData`):

![image](https://user-images.githubusercontent.com/292510/68043983-ca8ad900-fc93-11e9-8589-7db4498c3e17.png)

**Describe the solution you'd like**

Resolving the issue for each call stack will likely have to be done separately, although there are some commonalities.

### ServeSegment

`ServeSegment` is the portion of the call stack that receives and processes segmented inputs from HTTP. As soon as a request is received, the entire body is read into a byte slice. This branch accounts for 55.85% of total allocations. To resolve allocations around `ServeSegment`, there are at least two possible approaches:

1. Use `sync.Pool` but with caveats. Within `ServeSegment`, byte slices can be `Get`'d and the `Put` deferred until the function exits. Since ServeSegment is synchronous and only exits when transcoding is complete, this works nicely and would neatly release the allocation back to the pool, even after errors. However, a problem arises with the in-memory object store, which also takes a reference to the same data and holds it indefinitely outside the `ServeSegment ` call stack. To mitigate this, the local OS would need to be changed to make its own copy of the data, preferably from another `sync.Pool`. There are also reasonable arguments to be made that we shouldn't be explicitly maintaining an in-memory local OS, and should defer to a traditional filesystem for serving segments (which could in turn be backed by a tmpfs).

[...]

PR #1156, opened 2019-11-02
livepeer: Add option to enable pprof.

j0sh:
pprof is a neat profiling tool that's built into golang. In addition to tracking the usual suspects (CPU, allocations, mutexes) this allows us to generate neat flame graphs such as those shown here: https://github.com/livepeer/go-livepeer/issues/1155

- Sets up the pprof web handlers on the CLI port
- Moves the CLI handler to use the default HTTP mux, then reset the pointer to the default HTTP mux. We have to do this because unfortunately the default mux is the only thing that pprof works with, and we don't want to take the risk of some other public listener inadvertently pulling in the default HTTP mux, and pprof endpoints along with it. This is the same reason we run the CLI webserver on its own mux.
- Adds a simple args test

- Manually
- Args unit test

**Checklist:**

### DIFF
--- a/server/webserver.go
+++ b/server/webserver.go
@@ -12,6 +12,9 @@ import (
 	"strings"
 	"time"
 
+	// pprof adds handlers to default mux via `init()`
+	_ "net/http/pprof"
+
 	"github.com/cenkalti/backoff"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/types"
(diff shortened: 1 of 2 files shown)


j0sh:
Reworked the PR: now we make pprof available alongside the CLI port. Seems safe since the CLI port itself is expected to be trusted. This removes the need to run pprof on a separate port or have a special option for it. Note that we need to reset the default HTTP mux and use that for the CLI webserver.

server/webserver.go
```diff
@@ -12,6 +12,8 @@ import (
 	"strings"
 	"time"
 
+	_ "net/http/pprof"
```
darkdarkdragon, line 16: Lint gives this warning:
`a blank import should be only in a main or test package, or have a comment justifying it`
Can we add comment here to satisfy linter?

darkdarkdragon (changes requested):

server/webserver.go
j0sh: Curious, what linter / command shows this message? I'm not seeing it with `golint cmd/livepeer/livepeer.go` .

FWIW, this line is basically a copy-paste from https://golang.org/pkg/net/http/pprof/ . But agreed it'd be good to have an explanatory comment here; just want to make sure I'm satisfying the linter properly.

darkdarkdragon: `golint server/webserver.go`

It is ran by default by VSCode.

It is satisfied by any comment.

j0sh: Doh, ran it `golint` against the wrong file. Fixed.

darkdarkdragon (approved):
LGTM

PR #1157, opened 2019-11-04
eth/blockwatch: BackfillStartBlock option

### LINKED ISSUE #1149: StartBackfillBlockDepth config param for blockwatch package

The `blockwatch` package has a `BackfillEventsIfNeeded()` function which backfills events from the last retained block on the stack maintained by the package. If no last retained block is on the stack (eg. for a new node), backfilling will not happen and the blockwatcher will just start watching for events from the latest block.

With updates to active O set tracking there is a need for a synchronization operation, whereby a node replays all past `TranscoderActivated` and `TranscoderDeactivated` events to establish the current active set. 

To accomodate such synchronization the blockwatcher needs a feature that allows it to start backfilling events starting at block N instead. 

1. Add a `StartBackfillBlockDepth` field to `blockwatch.Config` supplied to the constructor

2. Set `startBlock` to backfill events from based on whether `StartBackfillBlockDepth` is defined

kyriediculous:
This PR updates the `blockwatch` package with a config option to start backfilling missed events from a certain block height, whereas previously backfilling would happen from the last retained block on the stack (in the DB in our case), if present. 

1. If there is a last retained block on the blockwatch stack, start backfilling from there. This should lead to an up to date view for the node
2. If there is no last retained block, start backfilling from the hardcoded `blockwatchBackfillStartBlock` defined in `livepeer.go` this is currently set to `0` and should lead to an up to date view for the node
3. If neither is defined return zero value and no error 

_**The one caveat is that this strategy makes some of the bash tests that use mainnet/rinkeby take quite a while since it starts backfilling from genesis...**_

- Added a `BackFillStartBlock *big.Int` field to `blockwatch.Config`
- Changed `watcher.getMissedEventsToBackfill()` to start backfilling from `watcher.backFillStartBlock` if defined, otherwise continue operating as before the changes this PR makes
- Added a unit test for backfilling when `config.BackfillStartBlock` is defined
- Moved around `SenderWatcher` to not process events when backfilling because it will just lead to no-ops anyway (mapping is empty when starting the node)

Ran unit tests

Fixes #1149 

**Checklist:**

### DIFF
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -47,7 +47,7 @@ var (
 
 	// The timeout for ETH RPC calls
 	ethRPCTimeout = 20 * time.Second
-	// The maximum block sfor the block watcher to retain
+	// The maximum blocks for the block watcher to retain
 	blockWatcherRetentionLimit = 20
 
 	// The gas required to redeem a PM ticket
@@ -365,10 +365,29 @@ func main() {
 			return
 		}
 		topics := watchers.FilterTopics()
+
+		// Determine backfilling start block
+		originalLastSeenBlock, err := dbh.LastSeenBlock()
+		if err != nil {
+			glog.Errorf("db: failed to retrieve latest retained block: %v", err)
+			return
+		}
+		currentRoundStartBlock, err := client.CurrentRoundStartBlock()
+		if err != nil {
+			glog.Errorf("eth: failed to retrieve current round start block: %v", err)
+			return
+		}
+
+		var blockWatcherBackfillStartBlock *big.Int
+		if originalLastSeenBlock == nil || originalLastSeenBlock.Cmp(currentRoundStartBlock) < 0 {
+			blockWatcherBackfillStartBlock = currentRoundStartBlock
+		}
+
 		blockWatcherCfg := blockwatch.Config{
 			Store:               n.Database,
 			PollingInterval:     blockPollingTime,
 			StartBlockDepth:     rpc.LatestBlockNumber,
+			BackfillStartBlock:  blockWatcherBackfillStartBlock,
 			BlockRetentionLimit: blockWatcherRetentionLimit,
 			WithLogs:            true,
 			Topics:              topics,
--- a/eth/blockwatch/block_watcher.go
+++ b/eth/blockwatch/block_watcher.go
@@ -45,6 +45,7 @@ type Config struct {
 	Store               MiniHeaderStore
 	PollingInterval     time.Duration
 	StartBlockDepth     rpc.BlockNumber
+	BackfillStartBlock  *big.Int
 	BlockRetentionLimit int
 	WithLogs            bool
 	Topics              []common.Hash
@@ -57,6 +58,7 @@ type Config struct {
 type Watcher struct {
 	blockRetentionLimit int
 	startBlockDepth     rpc.BlockNumber
+	backfillStartBlock  *big.Int
 	stack               *Stack
 	client              Client
 	blockFeed           event.Feed
@@ -77,6 +79,7 @@ func New(config Config) *Watcher {
 		pollingInterval:     config.PollingInterval,
 		blockRetentionLimit: config.BlockRetentionLimit,
 		startBlockDepth:     config.StartBlockDepth,
+		backfillStartBlock:  config.BackfillStartBlock,
 		stack:               stack,
 		client:              config.Client,
 		withLogs:            config.WithLogs,
@@ -281,28 +284,43 @@ func (w *Watcher) addLogs(header *MiniHeader) (*MiniHeader, error) {
 func (w *Watcher) getMissedEventsToBackfill(ctx context.Context) ([]*Event, error) {
 	events := []*Event{}
 
+	var (
+		startBlockNum          int
+		latestRetainedBlockNum int
+		blocksElapsed          int
+	)
+
 	latestRetainedBlock, err := w.stack.Peek()
 	if err != nil {
 		return events, err
 	}
-	// No blocks stored, nowhere to backfill to
-	if latestRetainedBlock == nil {
-		return events, nil
-	}
+
 	latestBlock, err := w.client.HeaderByNumber(nil)
 	if err != nil {
 		return events, err
 	}
-	blocksElapsed := big.NewInt(0).Sub(latestBlock.Number, latestRetainedBlock.Number)
-	if blocksElapsed.Int64() == 0 {
+	latestBlockNum := int(latestBlock.Number.Int64())
+
+	if latestRetainedBlock != nil {
+		latestRetainedBlockNum = int(latestRetainedBlock.Number.Int64())
+		// Events for latestRetainedBlock already processed, start at latestRetainedBlock + 1
+		startBlockNum = latestRetainedBlockNum + 1
+	} else if w.backfillStartBlock != nil {
+		startBlockNum = int(w.backfillStartBlock.Int64())
+	} else {
 		return events, nil
 	}
 
-	glog.Infof("Some blocks have elapsed since last boot. Backfilling block events (this can take a while)... blocksElapsed=%v", blocksElapsed.Int64())
-	startBlockNum := int(latestRetainedBlock.Number.Int64() + 1)
-	endBlockNum := int(latestRetainedBlock.Number.Int64() + blocksElapsed.Int64())
+	if blocksElapsed = latestBlockNum - startBlockNum; blocksElapsed == 0 {
+		return events, nil
+	}
+
+	glog.Infof("Backfilling block events (this can take a while)...\n")
+	glog.Infof("Start block: %v		 End block: %v		Blocks elapsed: %v\n", startBlockNum, startBlockNum+blocksElapsed, blocksElapsed)
+
+	endBlockNum := startBlockNum + blocksElapsed
 	logs, furthestBlockProcessed := w.getLogsInBlockRange(ctx, startBlockNum, endBlockNum)
-	if int64(furthestBlockProcessed) > latestRetainedBlock.Number.Int64() {
+	if furthestBlockProcessed > latestRetainedBlockNum {
 		// If we have processed blocks further then the latestRetainedBlock in the DB, we
 		// want to remove all blocks from the DB and insert the furthestBlockProcessed
 		// Doing so will cause the BlockWatcher to start from that furthestBlockProcessed.
@@ -337,9 +355,9 @@ func (w *Watcher) getMissedEventsToBackfill(ctx context.Context) ([]*Event, erro
 		for _, log := range logs {
 			blockHeader, ok := hashToBlockHeader[log.BlockHash]
 			if !ok {
-				// TODO(fabio): Find a way to include the parent hash for the block as well.
+				// TODO: Find a way to include the parent hash for the block as well.
 				// It's currently not an issue to omit it since we don't use the parent hash
-				// when processing block events in OrderWatcher.
+				// when processing block events in event watcher services
 				blockHeader = &MiniHeader{
 					Hash:   log.BlockHash,
 					Number: big.NewInt(0).SetUint64(log.BlockNumber),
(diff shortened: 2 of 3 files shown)


kyriediculous:
7b53fe41 adds a unit test for when there is no last retained block and no defined value for `backfillStartBlock`

cmd/livepeer/livepeer.go
```diff
@@ -401,6 +407,13 @@ func main() {
 		go unbondingWatcher.Watch()
 		defer unbondingWatcher.Stop()
 
+		// Backfill events that the node has missed since its last seen block. This method will block
+		// and the node will not continue setup until it finishes
+		if err := blockWatcher.BackfillEventsIfNeeded(blockWatchCtx); err != nil {
```
yondonfu: Do you see any harm in initializing the `SenderWatcher` before starting the backfill if its the case that any events emitted during the backfill will just be noops for the `SenderWatcher` since no senders are tracked yet? If not, it seems fine to always backfill after all event watchers are initialized instead of special casing certain event watchers such that some are initialized before and some are initialized after the backfill [1]. The motivation here would be to remove the need to remember why we're special casing some event watchers and not others.

[1] If we're special casing event watchers that will noop (or not do anything meaningful) for all events emitted during the backfill then the `RoundsWatcher` would also fall into this category since we only care about the *current* `lastInitializedRound` and `lastInitializedBlockHash` and not any values for previous rounds.

eth/blockwatch/block_watcher.go
```diff
@@ -281,28 +284,45 @@ func (w *Watcher) addLogs(header *MiniHeader) (*MiniHeader, error) {
...
 	}
-	blocksElapsed := big.NewInt(0).Sub(latestBlock.Number, latestRetainedBlock.Number)
-	if blocksElapsed.Int64() == 0 {
+	latestBlockNum := int(latestBlock.Number.Int64())
+
+	if latestRetainedBlock != nil {
+		latestRetainedBlockNum = int(latestRetainedBlock.Number.Int64())
+		// Events for latestRetainedBlock already processed, start at latestRetainedBlock + 1
+		startBlockNum = latestRetainedBlockNum + 1
+		blocksElapsed = latestBlockNum - startBlockNum
```
yondonfu: This line is repeated in this clause and the next clause - can we move it after this if/else if/else block?

eth/blockwatch/block_watcher.go
```diff
@@ -281,28 +284,45 @@ func (w *Watcher) addLogs(header *MiniHeader) (*MiniHeader, error) {
...
+		return events, nil
+	}
+
+	glog.Infof("Some blocks have elapsed since last boot. Backfilling block events (this can take a while)...\n")
+	glog.Infof("Start block: %v		 End block: %v		Blocks elapsed: %v\n", startBlockNum, startBlockNum+blocksElapsed, blocksElapsed)
+
+	endBlockNum := startBlockNum + blocksElapsed
 	logs, furthestBlockProcessed := w.getLogsInBlockRange(ctx, startBlockNum, endBlockNum)
-	if int64(furthestBlockProcessed) > latestRetainedBlock.Number.Int64() {
+	if furthestBlockProcessed > latestRetainedBlockNum {
```
yondonfu, line 322: I don't think we need to execute the logic in this if statement in the scenario where there is no last retained block and we're backfilling starting from `backfillStartBlock` since there would be no blocks in the DB yet.

eth/blockwatch/block_watcher.go
```diff
@@ -281,28 +284,45 @@ func (w *Watcher) addLogs(header *MiniHeader) (*MiniHeader, error) {
...
 	}
 
-	glog.Infof("Some blocks have elapsed since last boot. Backfilling block events (this can take a while)... blocksElapsed=%v", blocksElapsed.Int64())
-	startBlockNum := int(latestRetainedBlock.Number.Int64() + 1)
-	endBlockNum := int(latestRetainedBlock.Number.Int64() + blocksElapsed.Int64())
+	if blocksElapsed == 0 {
+		return events, nil
+	}
+
+	glog.Infof("Some blocks have elapsed since last boot. Backfilling block events (this can take a while)...\n")
```
yondonfu: In the case where we're backfilling from `backfillStartBlock` then there was no "last boot". I think we can just remove the first part of this statement and just log "Backfilling block events...<rest of log>"

eth/blockwatch/block_watcher_test.go
```diff
@@ -214,6 +214,34 @@ func TestGetMissedEventsToBackfillSomeMissed(t *testing.T) {
...
+	config.Store = store
+	config.Client = fakeClient
+	config.BackfillStartBlock = big.NewInt(0)
+	watcher := New(config)
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	events, err := watcher.getMissedEventsToBackfill(ctx)
+	require.NoError(t, err)
+	t.Log(events[0].BlockHeader.Number)
```
yondonfu: Can remove?

eth/blockwatch/block_watcher_test.go
```diff
@@ -214,6 +214,34 @@ func TestGetMissedEventsToBackfillSomeMissed(t *testing.T) {
...
+func TestGetMissedEventsToBackfill_BackfillStartBlock(t *testing.T) {
+	// Fixture will return block 30 as the tip of the chain
+	fakeClient, err := newFakeClient("testdata/fake_client_fast_sync_fixture.json")
+	require.NoError(t, err)
+
+	store := &stubMiniHeaderStore{}
+
+	config.Store = store
+	config.Client = fakeClient
+	config.BackfillStartBlock = big.NewInt(0)
```
yondonfu, line 226: Can we also test with a non-zero backfill start block?

eth/blockwatch/block_watcher_test.go
```diff
@@ -246,6 +274,31 @@ func TestGetMissedEventsToBackfillNoneMissed(t *testing.T) {
...
+	config.Client = fakeClient
+	watcher := New(config)
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	events, err := watcher.getMissedEventsToBackfill(ctx)
+	require.NoError(t, err)
+	assert.Len(t, events, 0)
+
+	// Check that block 5 is still in the DB
```
yondonfu: This comment doesn't seem relevant to this test?

yondonfu (changes requested):

cmd/livepeer/livepeer.go
kyriediculous: No I don't see any harm but it provides some very minor optimisation , I can revert the change so we don't have to deal with knowing why some are started before backfilling and some after.

eth/blockwatch/block_watcher.go
kyriediculous: We do need to store the block we backfilled to (last seen block now) in the DB so that when we execute `Watch()` we'll start from that block

So rather than shifting the logic and conditionals around I thought it would be okay to have this statement always execute when there are no blocks in the DB (`lastRetainedBlock = 0`). 

The only operation that is kind of unnecessary is that we fetch the headers from the DB and check the length.

eth/blockwatch/block_watcher_test.go
kyriediculous: fixed in be50d19d

eth/blockwatch/block_watcher_test.go
kyriediculous: fixed in be50d19d

eth/blockwatch/block_watcher_test.go
kyriediculous: fixed in be50d19d

eth/blockwatch/block_watcher.go
kyriediculous: fixed in be50d19d

eth/blockwatch/block_watcher.go
kyriediculous: fixed in be50d19d

cmd/livepeer/livepeer.go
kyriediculous: fixed in be50d19d

cmd/livepeer/livepeer.go
```diff
@@ -377,6 +380,9 @@ func main() {
 		// Wait until all event watchers have been initialized before starting the block watcher
 		blockWatcher := blockwatch.New(blockWatcherCfg)
 
+		blockWatchCtx, cancel := context.WithCancel(context.Background())
```
yondonfu: Can we revert this change?

kyriediculous:
5cba1fe adds support for determining the `blockWatchBackfillStartBlock` based on the current round start block vs the last seen block by the node. 

1. Fetch the node's lastRetainedBlock and store it in a temp variable originalLastRetainedBlock since the value will change after backfilling completes
2. Let backfillStartBlock be the current round start block
3. Start backfilling - see the options below for what block to start from
4. If lastRetainedBlock < backfillStartBlock, start from backfillStartBlock
5. If lastRetainedBlock >= backfillStartBlock, start from lastRetainedBlock
6. Backfilling completes

eth/blockwatch/block_watcher.go
```diff
@@ -281,28 +284,43 @@ func (w *Watcher) addLogs(header *MiniHeader) (*MiniHeader, error) {
...
+
 	latestBlock, err := w.client.HeaderByNumber(nil)
 	if err != nil {
 		return events, err
 	}
-	blocksElapsed := big.NewInt(0).Sub(latestBlock.Number, latestRetainedBlock.Number)
-	if blocksElapsed.Int64() == 0 {
+	latestBlockNum := int(latestBlock.Number.Int64())
+
+	if latestRetainedBlock != nil {
```
yondonfu: This piece of code will currently backfill starting from `lastRetainedBlock` as long as `lastRetainedBlock != nil`, but we actually want to do the following instead:

- If `lastRetainedBlock < currentRoundStartBlock`, backfill from `currentRoundStartBlock`
- Else, backfill from `lastRetainedBlock`

Given that this logic is reflected in how `backfillStartBlock` is initialized in `cmd/livepeer/livepeer.go`, in this piece of code, we could just check if `w.backfillStartBlock != nil` first - in effect, always backfilling from `backfillStartBlock` if it is passed in and if it is not passed in we would backfill from `lastRetainedBlock`

yondonfu (changes requested):

cmd/livepeer/livepeer.go
kyriediculous: Ah yes, this must have happened during rebasing I guess. Fixed in 95f4527f

eth/blockwatch/block_watcher.go
kyriediculous: Fixed in 95f4527f

eth/blockwatch/block_watcher.go
kyriediculous: Any additional thoughts or are you okay with keeping the functionality as is here?

eth/blockwatch/block_watcher.go
```diff
@@ -281,28 +284,43 @@ func (w *Watcher) addLogs(header *MiniHeader) (*MiniHeader, error) {
...
-	startBlockNum := int(latestRetainedBlock.Number.Int64() + 1)
-	endBlockNum := int(latestRetainedBlock.Number.Int64() + blocksElapsed.Int64())
+	if blocksElapsed = latestBlockNum - startBlockNum; blocksElapsed == 0 {
+		return events, nil
+	}
+
+	glog.Infof("Backfilling block events (this can take a while)...\n")
+	glog.Infof("Start block: %v		 End block: %v		Blocks elapsed: %v\n", startBlockNum, startBlockNum+blocksElapsed, blocksElapsed)
+
+	endBlockNum := startBlockNum + blocksElapsed
```
yondonfu: This is just `latestBlockNum`?

eth/blockwatch/block_watcher.go
yondonfu: I think this should be ok

eth/blockwatch/block_watcher.go
kyriediculous: fixed in 32f819a6

yondonfu (commented):

kyriediculous:
rebased !

yondonfu (approved):
LGTM 🚢

PR #1171, opened 2019-11-04
db, discovery: select active Os from DB

### LINKED ISSUE #1152: Select active orchestrators from DB

Update the `orchestrators` table in the DB with `activationRound` and `deactivationRound` columns. These values can then be used to filter active orchestrators for the current round by supplying a `currentRound` field to `DBOrchFilter` that is passed into `db.SelectOrchs()`

### db.DBOrch
Update `db.DBOrch` to include `ActivationRound` and `DeactivationRound` fields

| Field                      | Type    | Description                                                                                                  |
| -------------------------- | ------- | ------------------------------------------------------------------------------------------------------------ |
| **serviceURI**              | string | URI of O to use for submitting segments                                                                         |
| **EthereumAddr**                 | string |  Ethereum address for O                                                                        |
| **PricePerPixel**              | int64 | Price per pixel to transcode represented as a fixed point number with up to 3 decimals                                |
| **ActivationRound**            | int64 | Round in which an O joins the active set               |
| **DeactivationRound**      | int64 | Round in which an O gets evicted from the active set  |

### db.DBOrchFilter

We can extend the filter and `selectOrchs` SQL query to include a `currentRound` field to filter for active orchestrators: 

```
DBOrch.ActivationRound <= currentRound && currentRound < DBOrch.DeactivationRound
``` 

| Field                      | Type    | Description                                                                                                  |
| -------------------------- | ------- | ------------------------------------------------------------------------------------------------------------ |
| **MaxPrice**              | big.Rat | Max Price B is willing to accept from O                                                                        |
| **CurrentRound**                 | big.Int |  Round number for the current round  |

### db.updateOrch query

Update the updateOrch query to include the added fields in `DBOrch`

```
INSERT OR REPLACE INTO orchestrators(updatedAt, serviceURI, ethereumAddr, pricePerPixel, activationRound, deactivationRound, createdAt) VALUES(datetime(), ?1, ?2, ?3, ?4, ?5, (SELECT createdAt FROM orchestrators WHERE ethereumAddr = ?2))
```

Consider using `UPSERT` instead of `INSERT OR REPLACE` : https://www3.sqlite.org/lang_UPSERT.html

### db.selectOrchs query

Update the selectOrchs query to include being able to select based upon the new `DBOrchFilter.CurrentRound` field

**Additional context**

kyriediculous:
 This PR makes the necessary updates to `common/db` for tracking updates to the active orchestrator set. Primarily it adds `activationRound` and `deactivationRound` columns to the `orchestrators` table and allows for a `currentRound` filter to be passed in to the orchestrator selection query. 

- Added `activationRound` and `deactivationRound` columns to the `orchestrators` table
- Added a `currentRound` filter to `DBOrchFilter`
- Changed `updateOrch` query to use `UPSERT` instead of `INSERT OR REPLACE` , the idea behind this is that we don't have to fetch the entire `DBOrch` for an object when updating single fields in event watchers (serviceURI, activation/deactivation rounds)
- Added a filter to `selectOrchs` that filters for active orchestrators whenever `DBOrchFilter.CurrentRound` is defined
- Updated `NewDBOrch` to return a `DBOrch` instance with all values initiated
- Extracted generating the filter part of `selectOrch` query into a seperate helper that can be re-used for the `OrchCount` query
- Changed some tests in discovery to use the same underlying type when using `assert.Contains`

Ran unit tests

Fixes #1152 

**Checklist:**

### DIFF
--- a/common/db.go
+++ b/common/db.go
@@ -458,6 +458,26 @@ func (db *DB) SelectOrchs(filter *DBOrchFilter) ([]*DBOrch, error) {
 	return orchs, nil
 }
 
+func (db *DB) OrchCount(filter *DBOrchFilter) (int, error) {
+	if db == nil {
+		return 0, nil
+	}
+
+	qry, err := buildOrchCountQuery(filter)
+	if err != nil {
+		return 0, err
+	}
+
+	row := db.dbh.QueryRow(qry)
+
+	var count64 int64
+	if err := row.Scan(&count64); err != nil {
+		return 0, err
+	}
+
+	return int(count64), nil
+}
+
 func (db *DB) InsertUnbondingLock(id *big.Int, delegator ethcommon.Address, amount, withdrawRound *big.Int) error {
 	glog.V(DEBUG).Infof("db: Inserting unbonding lock %v for delegator %v", id, delegator.Hex())
 	_, err := db.insertUnbondingLock.Exec(id.Int64(), delegator.Hex(), amount.String(), withdrawRound.Int64())
@@ -636,6 +656,15 @@ func buildWinningTicketsQuery(sessionIDs []string) string {
 
 func buildSelectOrchsQuery(filter *DBOrchFilter) (string, error) {
 	query := "SELECT ethereumAddr, serviceURI, pricePerPixel, activationRound, deactivationRound FROM orchestrators "
+	fil, err := buildFilterOrchsQuery(filter)
+	if err != nil {
+		return "", err
+	}
+	return query + fil, nil
+}
+
+func buildOrchCountQuery(filter *DBOrchFilter) (string, error) {
+	query := "SELECT count(ethereumAddr) FROM orchestrators "
 	fil, err := buildFilterOrchsQuery(filter)
 	if err != nil {
 		return "", err
--- a/common/db.go
+++ b/common/db.go
@@ -19,11 +19,11 @@ import (
 	"github.com/pkg/errors"
 )
 
+// DB is an initialized DB driver with prepared statements
 type DB struct {
 	dbh *sql.DB
 
 	// prepared statements
-	selectOrchs                      *sql.Stmt
 	updateOrch                       *sql.Stmt
 	selectKV                         *sql.Stmt
 	updateKV                         *sql.Stmt
@@ -39,12 +39,16 @@ type DB struct {
 	deleteMiniHeader                 *sql.Stmt
 }
 
+// DBOrch is the type binding for a row result from the orchestrators table
 type DBOrch struct {
-	ServiceURI    string
-	EthereumAddr  string
-	PricePerPixel int64
+	ServiceURI        string
+	EthereumAddr      string
+	PricePerPixel     int64
+	ActivationRound   int64
+	DeactivationRound int64
 }
 
+// DBOrch is the type binding for a row result from the unbondingLocks table
 type DBUnbondingLock struct {
 	ID            int64
 	Delegator     ethcommon.Address
@@ -52,8 +56,10 @@ type DBUnbondingLock struct {
 	WithdrawRound int64
 }
 
+// DBOrchFilter is an object used to attach a filter to a selectOrch query
 type DBOrchFilter struct {
-	MaxPrice *big.Rat
+	MaxPrice     *big.Rat
+	CurrentRound *big.Int
 }
 
 var LivepeerDBVersion = 1
@@ -73,7 +79,9 @@ var schema = `
 		createdAt STRING DEFAULT CURRENT_TIMESTAMP NOT NULL,
 		updatedAt STRING DEFAULT CURRENT_TIMESTAMP NOT NULL,
 		serviceURI STRING,
-		pricePerPixel int64
+		pricePerPixel int64,
+		activationRound int64,
+		deactivationRound int64
 	);
 
 	CREATE TABLE IF NOT EXISTS unbondingLocks (
@@ -113,8 +121,14 @@ var schema = `
 	CREATE INDEX IF NOT EXISTS idx_blockheaders_number ON blockheaders(number);
 `
 
-func NewDBOrch(serviceURI string, orchAddr string) *DBOrch {
-	return &DBOrch{ServiceURI: serviceURI, EthereumAddr: orchAddr}
+func NewDBOrch(ethereumAddr string, serviceURI string, pricePerPixel int64, activationRound int64, deactivationRound int64) *DBOrch {
+	return &DBOrch{
+		ServiceURI:        serviceURI,
+		EthereumAddr:      ethereumAddr,
+		PricePerPixel:     pricePerPixel,
+		ActivationRound:   activationRound,
+		DeactivationRound: deactivationRound,
+	}
 }
 
 func InitDB(dbPath string) (*DB, error) {
@@ -161,17 +175,8 @@ func InitDB(dbPath string) (*DB, error) {
 		// all good; nothing to do
 	}
 
-	// updateOrchestrators statement
-	stmt, err := db.Prepare("INSERT OR REPLACE INTO orchestrators(updatedAt, serviceURI, ethereumAddr, pricePerPixel, createdAt) VALUES(datetime(), ?1, ?2, ?3, (SELECT createdAt FROM orchestrators WHERE ethereumAddr = ?2))")
-	if err != nil {
-		glog.Error("Unable to prepare updateOrchestrators stmt ", err)
-		d.Close()
-		return nil, err
-	}
-	d.updateOrch = stmt
-
 	// selectKV prepared statement
-	stmt, err = db.Prepare("SELECT value FROM kv WHERE key=?")
+	stmt, err := db.Prepare("SELECT value FROM kv WHERE key=?")
 	if err != nil {
 		glog.Error("Unable to prepare selectKV stmt", err)
 		d.Close()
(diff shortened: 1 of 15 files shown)


common/db.go
```diff
@@ -408,16 +407,24 @@ func (db *DB) SelectOrchs(filter *DBOrchFilter) ([]*DBOrch, error) {
...
+			deactivationRound int64
+		)
+		if err := rows.Scan(&serviceURI, &ethereumAddr, &pricePerPixel, &activationRound, &deactivationRound); err != nil {
 			glog.Error("db: Unable to fetch orchestrator ", err)
 			continue
 		}
-		orch.ServiceURI = serviceURI
-		orch.EthereumAddr = ethereumAddr
-		orchs = append(orchs, &orch)
+		orchs = append(orchs, &DBOrch{
```
yondonfu: Could we use `NewDBOrch()` here?

common/db.go
```diff
@@ -599,15 +606,52 @@ func buildWinningTicketsQuery(sessionIDs []string) string {
...
+		if filter.CurrentRound != nil {
+			currentRound := filter.CurrentRound.Int64()
+			qry += fmt.Sprintf(" AND activationRound <= %v AND %v < deactivationRound", currentRound, currentRound)
+		}
+	}
+	return qry, nil
+}
+
+func buildUpsertOrchsQuery(orch *DBOrch) string {
+	qry := "INSERT INTO orchestrators(updatedAt, serviceURI, ethereumAddr, pricePerPixel, activationRound, deactivationRound, createdAt)"
```
yondonfu: Dynamically generating the SQL query seems like the most straightforward way to implement the upsert without requiring the `orch` argument to reflect the current values of the corresponding row in the table if it exists and it seems that SQL injection *should* be a non-issue since the only strings are `serviceURI`/`ethereumAddr` and we validate that `serviceURI` is actually an URI and `ethereumAddr` needs to conform to the ETH address format. I don't see a better alternative at the moment, but maybe @j0sh has some thoughts

common/db_test.go
```diff
@@ -233,13 +233,16 @@ func TestSelectUpdateOrchs_AddingUpdatingRow_NoError(t *testing.T) {
 	assert.Equal(orchs[0].ServiceURI, orch.ServiceURI)
 
 	// updating row with same orchAddress
-	orchUpdate := NewDBOrch("127.0.0.1:8937", orchAddress)
+	orchUpdate := NewDBOrch("127.0.0.1:8937", orchAddress, 1000, 5, 10)
```
yondonfu: Can we add tests for the different cases in `buildUpsertOrchsQuery()` i.e. for an existing row, update a column only if the field that corresponds to the column in the `DBOrch` struct does not equal the zero value?

discovery/discovery_test.go
```diff
@@ -265,10 +265,14 @@ func TestNewDBOrchestratorPoolCache_GivenListOfOrchs_CreatesPoolCacheCorrectly(t
 	orchs, err := node.Database.SelectOrchs(nil)
 	require.Nil(err)
 	assert.Len(orchs, 3)
+	orchsTest := make([]orchTest, 1)
```
yondonfu, line 268: Instead of defining a new `orchTest` struct just for these tests could we:

- Accept `ethAddresses []ethcommon.Address` in `StubOrchestrators()`
- Loop through the members of `orchs` and let `db0` equal the next member in each iteration
- Check `assert.Contains(addresses, db0.ServiceURI)`
- Check `assert.Contains(ethAddresses, db0.EthereumAddr)`

yondonfu (changes requested):

discovery/discovery_test.go
kyriediculous: I think the benefit of having `orchTest` is that we can do an exact object comparison regardless of which we care about rather than comparing seperate fields. 

eg. In `TestNewOrchestratorPoolWithPred_TestPredicate` 
I want to make sure that an orchestrator with address X , ethAddress f(X) actually has a price setting of Y.

common/db.go
kyriediculous: fixed in e8589ee4

common/db_test.go
kyriediculous: fixed in e8589ee4

common/db.go
j0sh: @yondonfu @kyriediculous Looks like SQLite's `CASE` statement can be used to achieve what we want here via a prepared statement. See https://stackoverflow.com/a/4874440 for a usage example.

The query might end up being long, so it wouldn't be terrible to break it across lines. Something like:

```golang
stmt, err := db.Prepare(`
INSERT INTO orchestrators(updatedAt, serviceURI, ethereumAddr, ... )
VALUES (?, ?, ?, ... )
ON CONFLICT(ethereumAddr)  DO UPDATE SET
serviceURI =
  CASE WHEN trim(excluded.serviceURI) == ""
  THEN orchestrators.serviceURI
  ELSE trim(excluded.serviceURI) END,
pricePerPixel =
  CASE WHEN excluded.pricePerPixel == 0 THEN ...
...
`)
```

Since there are a lot of parameters here, I also wonder if binding to [named params](https://www.sqlite.org/c3ref/bind_parameter_name.html) would help readability? Although I haven't looked into how to use those within the go-sqlite3 bindings. Or at some point we could start using a code generator which would cut down on the boilerplate and probably produce a more ergonomic API, but that's not terrible important right now.

BTW this is a nice use case of a the relatively new ON CONFLICT statement - it hadn't been released when we were originally working  on the orchestrator DB stuff.

yondonfu: Thanks for the tip @j0sh!

@kyriediculous I think we should try to stick with prepared statements whenever possible - what do you think about using `CASE` here to setup the prepared statement instead of dynamically generating the statement?

> Since there are a lot of parameters here, I also wonder if binding to named params would help readability?

Seems like it should be possible by passing in a [sql.NamedArg](https://golang.org/pkg/database/sql/#Named) but apparently [go-sqlite3](https://github.com/mattn/go-sqlite3/issues/697) only accepts parameter names of the form `:foo` at the moment.

discovery/discovery_test.go
yondonfu: > I think the benefit of having orchTest is that we can do an exact object comparison regardless of which we care about rather than comparing seperate fields.

👍 

> I want to make sure that an orchestrator with address X , ethAddress f(X) actually has a price setting of Y

I'm a bit confused by this statement - I don't see `orchTest` being used in `TestNewOrchestratorPoolWithPred_TestPredicate`.

common/db.go
kyriediculous: using `CASE` sounds good to me, I'll check it out.

common/db.go
```diff
@@ -113,8 +121,14 @@ var schema = `
 	CREATE INDEX IF NOT EXISTS idx_blockheaders_number ON blockheaders(number);
 `
 
-func NewDBOrch(serviceURI string, orchAddr string) *DBOrch {
-	return &DBOrch{ServiceURI: serviceURI, EthereumAddr: orchAddr}
+func NewDBOrch(serviceURI string, orchAddr string, pricePerPixel int64, activationRound int64, deactivationRound int64) *DBOrch {
```
yondonfu: Since `EthereumAddr` is the primary key in the table and it is always required when constructing a `DBOrch` to be passed into `UpdateOrch()`, it might be helpful to specify `orchAddr` as the first argument here and have the optional arguments follow it.

j0sh: If these args are optional then they should be put into a struct. That would also help make the sig less cumbersome.

yondonfu: Ah right - the above comment made me realize that in the cases where we do not want to set certain values in `DBOrch` since they are optional we can just directly create the struct and only specify the fields we care about (i.e. `&DBOrch{orchAddr: orchAddr, activationRound: 100}`) instead of using this constructor. Any fields not specified will just take on their default value (in this case empty strings or 0 since each field is either a `string` or `int64`).

common/db.go
kyriediculous: It does lead to quite a bit of code duplication whenever we use `DBOrchFilter` though, we currently build that part of the query with a helper than is used in both `selectOrchs` and `orchCount`

common/db.go
kyriediculous: I guess usage of the constructor vs literal is up to the caller of `db.SelectOrchs` and `db.OrchCount` , or do you want to get rid of the `NewDBOrch` function altogether?

discovery/discovery_test.go
kyriediculous: Right, bad example. I did slim down the usage of `orchTest` down in #1182 quite a bit so perhaps we can find a compromise there?

common/db.go
kyriediculous: using `CASE` and named args in 670346d

common/db.go
yondonfu: I'm fine with keeping the `NewDBOrch()` function. The comment about using the literal to avoid specifying optional args was just something that came to mind for the other PRs that build on top of this one, but I see you've already switched to using the literal in those PRs.

common/db_test.go
```diff
@@ -233,13 +233,64 @@ func TestSelectUpdateOrchs_AddingUpdatingRow_NoError(t *testing.T) {
...
 
 	updatedOrch, err := dbh.SelectOrchs(nil)
 	assert.Len(updatedOrch, 1)
 	assert.Equal(updatedOrch[0].ServiceURI, orchUpdate.ServiceURI)
+	assert.Equal(updatedOrch[0].ActivationRound, orchUpdate.ActivationRound)
+	assert.Equal(updatedOrch[0].DeactivationRound, orchUpdate.DeactivationRound)
+	assert.Equal(updatedOrch[0].PricePerPixel, orchUpdate.PricePerPixel)
+
+	// updating only serviceURI
+	serviceURIUpdate := NewDBOrch(orchAddress, "127.0.0.1:8938", 0, 0, 0)
```
yondonfu: Let's use struct literals in this test so we can omit the optional args?

yondonfu (changes requested):

common/db_test.go
kyriediculous: fixed in da29fed7 & efea741f

yondonfu (commented):

kyriediculous:
rebased !

yondonfu (approved):
LGTM - I suggest waiting for approvals of the other PRs built on top of this one so those can be merged into this branch and then this branch can be merged into master.

kyriediculous:
rebased !

PR #1180, opened 2019-11-07
db: orchestrator count sql query

### LINKED ISSUE #1153: activeOrchCount SQL query

Currently the orchestratorPool size for `DBOrchestratorPoolCache` is retrieved by first retrieving all URL's as a slice, then returning the length of the slice. 

Instead, with the updates to the DB in #1146, we can implement a new `activeOrchCount` query, which is then called from `DBOrchestratorPoolCache.Size()`

Instead of fetching all orchestrator URL's from the database and getting the length of the returned slice, we can instead return a direct row count from an SQL query: 

```SELECT count(serviceURI) FROM orchestrators WHERE <filter>``` 

**Additional context**

### PR DESCRIPTION
**Pull request by kyriediculous**
This PR adds an SQL query that gets a count of orchestrators according to the passed in `DBOrchFilter`

- Added a new exported method to DB `db.OrchCount(filter *DBOrchFilter) (int, error)` that returns the number or orchestrators in the db that pass `filter`
- Added a helper that creates the sql query used in `OrchCount`
- added a unit test for the new query

ran unit tests

Fixes #1153

**Checklist:**

### DIFF
--- a/common/db.go
+++ b/common/db.go
@@ -458,6 +458,26 @@ func (db *DB) SelectOrchs(filter *DBOrchFilter) ([]*DBOrch, error) {
 	return orchs, nil
 }
 
+func (db *DB) OrchCount(filter *DBOrchFilter) (int, error) {
+	if db == nil {
+		return 0, nil
+	}
+
+	qry, err := buildOrchCountQuery(filter)
+	if err != nil {
+		return 0, err
+	}
+
+	row := db.dbh.QueryRow(qry)
+
+	var count64 int64
+	if err := row.Scan(&count64); err != nil {
+		return 0, err
+	}
+
+	return int(count64), nil
+}
+
 func (db *DB) InsertUnbondingLock(id *big.Int, delegator ethcommon.Address, amount, withdrawRound *big.Int) error {
 	glog.V(DEBUG).Infof("db: Inserting unbonding lock %v for delegator %v", id, delegator.Hex())
 	_, err := db.insertUnbondingLock.Exec(id.Int64(), delegator.Hex(), amount.String(), withdrawRound.Int64())
@@ -636,6 +656,15 @@ func buildWinningTicketsQuery(sessionIDs []string) string {
 
 func buildSelectOrchsQuery(filter *DBOrchFilter) (string, error) {
 	query := "SELECT ethereumAddr, serviceURI, pricePerPixel, activationRound, deactivationRound FROM orchestrators "
+	fil, err := buildFilterOrchsQuery(filter)
+	if err != nil {
+		return "", err
+	}
+	return query + fil, nil
+}
+
+func buildOrchCountQuery(filter *DBOrchFilter) (string, error) {
+	query := "SELECT count(ethereumAddr) FROM orchestrators "
 	fil, err := buildFilterOrchsQuery(filter)
 	if err != nil {
 		return "", err
--- a/common/db_test.go
+++ b/common/db_test.go
@@ -338,18 +338,73 @@ func TestSelectUpdateOrchs_AddingMultipleRows_NoError(t *testing.T) {
 	assert.Equal(orchsUpdated[1].ServiceURI, orchAdd.ServiceURI)
 }
 
+func TestOrchCount(t *testing.T) {
+	assert := assert.New(t)
+	require := require.New(t)
+	var orchList []string
+	var nilDB *DB
+	zeroOrchs, nilErr := nilDB.OrchCount(&DBOrchFilter{})
+	assert.Zero(zeroOrchs)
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: common/db.go [hunk 1]
```diff
@@ -423,6 +423,25 @@ func (db *DB) SelectOrchs(filter *DBOrchFilter) ([]*DBOrch, error) {
...
+		return 0, nil
+	}
+
+	rows, err := db.dbh.Query(buildOrchCountQuery((filter)))
+	defer rows.Close()
+	if err != nil {
+		return 0, err
+	}
+	var count64 int64
+	for rows.Next() {
```
yondonfu: Could we use [QueryRow()](https://golang.org/pkg/database/sql/#DB.QueryRow) here since we know the query will return at most one row?

common/db_test.go
```diff
@@ -326,18 +326,73 @@ func TestSelectUpdateOrchs_AddingMultipleRows_NoError(t *testing.T) {
...
+	// use maxPrice filter, should return 5 results
+	orchCount, err = dbh.OrchCount(&DBOrchFilter{MaxPrice: big.NewRat(1, 10)})
+	assert.Nil(err)
+	assert.Equal(5, orchCount)
+
+	// use maxPrice and active O filter, should return 3 results
+	orchCount, err = dbh.OrchCount(&DBOrchFilter{MaxPrice: big.NewRat(1, 8), CurrentRound: big.NewInt(5)})
+	assert.Nil(err)
+	assert.Equal(3, orchCount)
+
```
yondonfu: Extra newline here?

common/db_test.go
```diff
@@ -326,18 +326,73 @@ func TestSelectUpdateOrchs_AddingMultipleRows_NoError(t *testing.T) {
...
+	var orchList []string
+	var nilDB *DB
+	zeroOrchs, nilErr := nilDB.OrchCount(&DBOrchFilter{})
+	assert.Zero(zeroOrchs)
+	assert.Nil(nilErr)
+
+	dbh, dbraw, err := TempDB(t)
+	defer dbh.Close()
+	defer dbraw.Close()
+	assert.Nil(err)
```
yondonfu: Should we use `require` here since the rest of the tests would encounter problems if an error is encountered here?

common/db_test.go
```diff
@@ -326,18 +326,73 @@ func TestSelectUpdateOrchs_AddingMultipleRows_NoError(t *testing.T) {
...
+
+	for i := 0; i < 10; i++ {
+		orch := NewDBOrch("https://127.0.0.1:"+strconv.Itoa(8936+i), pm.RandAddress().String(), 1, int64(i), int64(5+i))
+		orch.PricePerPixel, err = PriceToFixed(big.NewRat(1, int64(5+i)))
+		assert.Nil(err)
```
yondonfu: `require` in this setup loop?

yondonfu (changes requested):

common/db.go
kyriediculous: fixed in 2957c859

common/db_test.go
kyriediculous: fixed in 2957c859

common/db_test.go
kyriediculous: fixed in 2957c859

common/db_test.go
kyriediculous: fixed in 2957c859

common/db.go
```diff
@@ -423,6 +423,21 @@ func (db *DB) SelectOrchs(filter *DBOrchFilter) ([]*DBOrch, error) {
 	return orchs, nil
 }
 
+func (db *DB) OrchCount(filter *DBOrchFilter) (int, error) {
+	if db == nil {
+		return 0, nil
+	}
+
+	row := db.dbh.QueryRow(buildOrchCountQuery((filter)))
```
yondonfu: 1.  Since `buildOrchCountQuery()` returns `string, error` we should check if there is an error returned and we should also only pass in the query string to `QueryRow()`.

2. Can we remove the parens around `filter` when passing it into `buildOrchCountQuery()`?

yondonfu (changes requested):

common/db.go
kyriediculous: fixed in dfd3cc53

yondonfu:
Let's rebase

kyriediculous:
rebased !

yondonfu:
3cc883f looks like a duplicate commit (already exists in #1171)? Can it be dropped?

kyriediculous:
Hmm not sure what went wrong there, dropped after rebasing again.

yondonfu (approved):
LGTM 🚢

PR #1181, opened 2019-11-07
cmd, eth/watchers: ServiceRegistry event watcher

kyriediculous:
**What does this pull request do? Explain your changes. (required)**
This PR adds an event watcher that watches the `ServiceRegistry` contract for emitted `ServiceURIUpdate` events and updates orchestrators in the database accordingly.

**Specific updates (required)**
- Added a `ServiceRegistryWatcher` type and start it when the node starts
- Added `ServiceURIUpdate` event to `eth/watchers/topics.go`
- Added a method to generate stubbed out serviceURI update events
- Change the stub eth client method `GetServiceURI` to return a stubbed out value

**How did you test each of these updates (required)**
ran unit tests

**Does this pull request close any open issues?**
Fixes #1151 

**Checklist:**
- [ ] README and other documentation updated
- [x] Node runs in OSX and devenv
- [x] All tests in `./test.sh` pass

PR #1173, opened 2019-11-06
cmd, eth: orchestrator activated/deactivated watcher

kyriediculous:
**What does this pull request do? Explain your changes. (required)**
This PR adds an implementation for an `OrchestratorWatcher` that watches for on-chain emitted `TranscoderActivated` and `TranscoderDeactivated` events and updates orchestrators stored in the database accordingly. 

**Specific updates (required)**
- Added `TranscoderActivated` and `TranscoderDeactivated` to `eth/watchers/topics.go`
- Added an `OrchestratorStore` interface to `eth/watchers/types` to be able to stub out `common.DB` for easier unit testing
- Added an `OrchestratorWatcher` type
- Start an `OrchestratorWatcher`  in `livepeer.go`
- Add a `StubOrchestratorStore` and helpers for creating stubbed out `TranscoderActivated/TranscoderDeactivated` events to `eth/watchers/stub.go`
- Added `ActivationRound` and `DeactivationRound` fields to `types.Transcoder`
- Make sure `LivepeerEthClient.GetTranscoder()` now also returns `ActivationRound` and `DeactivationRound`

**How did you test each of these updates (required)**
Ran unit tests

**Does this pull request close any open issues?**
Fixes #1150 

**Checklist:**
- [ ] README and other documentation updated
- [x] Node runs in OSX and devenv
- [x] All tests in `./test.sh` pass

PR #1182, opened 2019-11-11
cmd, common, discovery, eth, net: orchestrator pool active set tracking

### LINKED ISSUE #1123: Update active set tracking

Note: The contracts still use the term "transcoders". Transcoders within the contracts correspond to orchestrators outside of the contracts.

In the latest version of the contracts, the state of the on-chain active set is tracked in the following way:

- An on-chain pool with a fixed max size is used to keep track of which orchestrators to activate/deactivate
- When an orchestrator is added to the on-chain pool, its `activationRound` is set to `currentRound + 1` and a `TranscoderActivated` event is emitted
- When an orchestrator is removed from the on-chain pool, its `deactivationRound` is set to `currentRound + 1` and a `TranscoderDeactivated` event is emitted
- An orchestrator is active in the `currentRound` if `activationRound <= currentRound < deactivationRound`
- The active set for the `currentRound` consists of all orchestrators that satisfy the above condition in the `currentRound`

Unlike in the old version of the contracts, the on-chain pool in the new version of the contracts does not necessarily reflect the state of the active set in a round. If an orchestrator is removed from the pool in the current round, the orchestrator will remain in the active set until the end of the round. If an orchestrator is added to the pool in the current round, the orchestrator will not be in the active set until the next round.

Given these changes to the way the state of the active set is tracked in the contracts, the node also an updated strategy for tracking the active set.

The node can watch for `TranscoderActivated` and `TranscoderDeactivated` events emitted from the BondingManager contract to updates its view of the active set. A rough sketch of the steps to accomplish this (just a high level overview - implementation details omitted) are:

- [ ] Update the DB schema for the orchestrators table to include `activationRound` and `deactivationRound`
- [ ] Add a `CurrentRound` field to `DBOrchFilter`. If this field is populated, only active orchestrators in `CurrentRound` should be returned
- [ ] Update the `db.SelectOrchs()` query to only fetch active orchestrators in `CurrentRound` if the `DBOrchFilter.CurrentRound` is populated
- [ ] Create an event watcher that watches for `TranscoderActivated` and `TranscoderDeactivated` events from the BondingManager contract and update the corresponding orchestrator table entries accordingly with the address' current `activationRound` and `deactivationRound` values. We likely want to cache the serviceURI for an address when we see the `TranscoderActivated` event
- [ ] Either use the same event watcher as above or create a new event watcher that watches for `ServiceURIUpdate` events and updates serviceURIs stored in the DB. We only care about orchestrators that we are aware of (i.e. orchestrators in DB)
- [ ] Remove the call to `RegisteredTranscoders()` in `DBOrchestratorPoolCache` that fetches the members of the on-chain pool since the event watcher for activation/deactivation events should populate the DB with orchestrator addresses. Instead, we can just query for a list of active orchestrators from the DB. We can also keep the code that calls `GetOrchestrator()` on each of the orchestrators since we need off-chain prices though
- [ ] Update `blockwatch.Watcher` to be able to backfill from a specified block in the past. At the moment, it will only backfill past events if it has seen a block before (i.e. if it has started watching previously). It will not backfill past events if the node is starting for the first time. When the node is starting for the first time, we can have the node backfill past events starting from the block when the contracts were first deployed (on Rinkeby/mainnet this would be the block when the contracts were upgraded)

### LINKED ISSUE #1154: Update DBOrchestratorPoolCache dependencies & methods for active set tracking

Update `DBOrchestratorPoolCache` dependencies and methods to accomdate the changes in #1152 and #1153

We will rework the current `DBOrchestratorPoolCache` type to accomodate event watching and making updates to the DB on the background, while still adhering to the `OrchestratorPool` interface. 

Most exposed methods will remain roughly the same, the main difference will be that the Database will be updated by background services that watch for on-chain events rather than polling a remote ethereum node on a fixed interval. 

```
type OrchestratorPool interface {
	GetURLs() []*url.URL
	GetOrchestrators(int) ([]*OrchestratorInfo, error)
	Size() int
}
```

#### Object

| Field                 | Type                       | Description                                         |
| --------------------- | -------------------------- | --------------------------------------------------- |
| **db**                | common.DB                  | DB instance                                         |
| **sender**            | pm.Sender                  | Sender instance that exposes `ValidateTicketParams` |

** NOTE: We could potentially remove the dependency on `pm.Sender` if we supply a predicate function to the constructor rather than hardcoding it in `GetOrchestrators`, the predicate function would then be an extra field on the struct. 

##### GetOrchestrators

Gets a list of orchestrators filtered for valid ticket params and using the optional `DBOrchFilter`

```
func (pool *orchestratorPool) GetOrchestrators(int) ([]*OrchestratorInfo, error) {
    // 1. Fetch active O set from DB
    // 2. For each O: 
    //     Fetch OrchestratorInfo and filter based on an (optional) predicate function
    //     Under the hood this will use a single-use instance of the `orchestratorPool` type
    // 3. Return []*OrchestratorInfo array
}
```

##### GetURLs

Get a list of orchestrator URLs filtered using the optional `DBOrchFilter`

```
func (dbo *DBOrchestratorPoolCache) GetURLs() []*url.URL {
    // 1. Query O URL's from database
    // 2. Parse to URI's
    // 3. Return array
}
```

##### Size
Get the size of a list of orchestrator URLs filtered using the optional `DBOrchFilter`

```
func (o *orchestratorPool) Size() int {
	// Return DBOrch count query result
}
```

**Additional context**

kyriediculous:
This PR updates the `discovery` package to accomodate for the changes in #1157 #1171 #1173 #1180 and #1181 

From a high level it implements following features:
- Update `DBOrchestratorPoolCache` to filter for active orchestrators as well as a max price
- Remove `LivepeerNode` as a dependency
- Expose a method `DBOrchestratorPoolCache.CacheTranscoderPool()` that updates the database with the transcoders currently in the pool
- Facilitate caching the transcoder pool on node startup when the node's view of the active set is out of date
- Introduce a `common/types.go` file where types commonly used by packages can exist

- 8555283 & 9c5fe43 : adds `Broadcaster` (previously `server.Broadcaster`), `OrchestratorPool` (previously `net.OrchestratorPool` and `OrchestratorStore` (previously unexported from eth/watchers`) to `common/types.go`

- 9f17042  : In `eth/client` renamed `LivepeerEthClient.RegisteredTranscoders()` to `LivepeerEthClient.TranscoderPool()` to better reflect the functionality of the method after the streamflow update

- bf0cbf5 : Adds a `stub.go` file to the `discovery` package for test helpers, stubs and mocks

- 430062d : Adds `common.Broadcaster` (interface for `core.Broadcaster`)  as a function argument to `OrchestratorPool.GetOrchestrators` rather than creating a new instance in the function body since `LivepeerNode` is no longer a dependency of the `discovery` package

- 17c63a5 : Fixes up the different `OrchestratorPool` implementations' constructors to reflect the updated dependencies

- b625830 : Adds the necessary updates to `DBOrchestratorPoolCache` to facilitate filtering active orchestrators 

- 2583ccd : calls `DBOrchestratorPoolCache.CacheTranscoderPool` on node startup when `lastRetainedBlock.Number < currentRoundStartBlock.Number`

ran unit tests

Fixes #1154 
Fixes #1123

**Checklist:**

### DIFF
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -288,6 +288,8 @@ func main() {
 
 	watcherErr := make(chan error)
 	var roundsWatcher *watchers.RoundsWatcher
+	var originalLastSeenBlock *big.Int
+	var currentRoundStartBlock *big.Int
 	if *network == "offchain" {
 		glog.Infof("***Livepeer is in off-chain mode***")
 
@@ -368,12 +370,12 @@ func main() {
 		topics := watchers.FilterTopics()
 
 		// Determine backfilling start block
-		originalLastSeenBlock, err := dbh.LastSeenBlock()
+		originalLastSeenBlock, err = dbh.LastSeenBlock()
 		if err != nil {
 			glog.Errorf("db: failed to retrieve latest retained block: %v", err)
 			return
 		}
-		currentRoundStartBlock, err := client.CurrentRoundStartBlock()
+		currentRoundStartBlock, err = client.CurrentRoundStartBlock()
 		if err != nil {
 			glog.Errorf("eth: failed to retrieve current round start block: %v", err)
 			return
@@ -397,7 +399,7 @@ func main() {
 		// Wait until all event watchers have been initialized before starting the block watcher
 		blockWatcher := blockwatch.New(blockWatcherCfg)
 
-		roundsWatcher, err := watchers.NewRoundsWatcher(addrMap["RoundsManager"], blockWatcher, n.Eth)
+		roundsWatcher, err = watchers.NewRoundsWatcher(addrMap["RoundsManager"], blockWatcher, n.Eth)
 		if err != nil {
 			glog.Errorf("Failed to setup roundswatcher: %v", err)
 			return
@@ -509,8 +511,6 @@ func main() {
 			}
 
 			sigVerifier := &pm.DefaultSigVerifier{}
-			// TODO: Initialize Validator with an implementation
-			// of RoundsManager that reads from a cache
 			validator := pm.NewValidator(sigVerifier, roundsWatcher)
 			gpm := eth.NewGasPriceMonitor(backend, blockPollingTime)
 			// Start gas price monitor
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -287,6 +287,7 @@ func main() {
 	}
 
 	watcherErr := make(chan error)
+	var roundsWatcher *watchers.RoundsWatcher
 	if *network == "offchain" {
 		glog.Infof("***Livepeer is in off-chain mode***")
 
@@ -634,17 +635,27 @@ func main() {
 		*rtmpAddr = defaultAddr(*rtmpAddr, "127.0.0.1", RtmpPort)
 		*httpAddr = defaultAddr(*httpAddr, "127.0.0.1", RpcPort)
 
+		bcast := core.NewBroadcaster(n)
+
 		// Set up orchestrator discovery
 		if *orchWebhookURL != "" {
 			whurl, err := getOrchWebhook(*orchWebhookURL)
 			if err != nil {
 				glog.Fatal("Error setting orch webhook URL ", err)
 			}
-			n.OrchestratorPool = discovery.NewWebhookPool(n, whurl)
+			n.OrchestratorPool = discovery.NewWebhookPool(bcast, whurl)
 		} else if len(orchURLs) > 0 {
-			n.OrchestratorPool = discovery.NewOrchestratorPool(n, orchURLs)
+			n.OrchestratorPool = discovery.NewOrchestratorPool(bcast, orchURLs)
 		} else if *network != "offchain" {
-			n.OrchestratorPool = discovery.NewDBOrchestratorPoolCache(n)
+			ctx, cancel := context.WithCancel(context.Background())
+			defer cancel()
+			dbOrchPoolCache, err := discovery.NewDBOrchestratorPoolCache(ctx, n, roundsWatcher)
+
+			if err != nil {
+				glog.Errorf("Could not create orchestrator pool with DB cache: %v", err)
+			}
+
+			n.OrchestratorPool = dbOrchPoolCache
 		}
 		if n.OrchestratorPool == nil {
 			// Not a fatal error; may continue operating in segment-only mode
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -288,8 +288,6 @@ func main() {
 
 	watcherErr := make(chan error)
 	var roundsWatcher *watchers.RoundsWatcher
-	var originalLastSeenBlock *big.Int
-	var currentRoundStartBlock *big.Int
 	if *network == "offchain" {
 		glog.Infof("***Livepeer is in off-chain mode***")
 
@@ -370,12 +368,12 @@ func main() {
 		topics := watchers.FilterTopics()
 
 		// Determine backfilling start block
-		originalLastSeenBlock, err = dbh.LastSeenBlock()
+		originalLastSeenBlock, err := dbh.LastSeenBlock()
 		if err != nil {
 			glog.Errorf("db: failed to retrieve latest retained block: %v", err)
 			return
 		}
-		currentRoundStartBlock, err = client.CurrentRoundStartBlock()
+		currentRoundStartBlock, err := client.CurrentRoundStartBlock()
 		if err != nil {
 			glog.Errorf("eth: failed to retrieve current round start block: %v", err)
 			return
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -445,33 +445,6 @@ func main() {
 		go serviceRegistryWatcher.Watch()
 		defer serviceRegistryWatcher.Stop()
 
-		blockWatchCtx, cancel := context.WithCancel(context.Background())
-		defer cancel()
-
-		// Backfill events that the node has missed since its last seen block. This method will block
-		// and the node will not continue setup until it finishes
-		if err := blockWatcher.BackfillEventsIfNeeded(blockWatchCtx); err != nil {
-			glog.Errorf("Failed to backfill events: %v", err)
-			return
-		}
-
-		blockWatcherErr := make(chan error, 1)
-		go func() {
-			if err := blockWatcher.Watch(blockWatchCtx); err != nil {
-				blockWatcherErr <- fmt.Errorf("block watcher error: %v", err)
-			}
-		}()
-
-		go func() {
-			var err error
-			select {
-			case err = <-roundsWatcherErr:
-			case err = <-blockWatcherErr:
-			}
-
-			watcherErr <- err
-		}()
-
 		n.Balances = core.NewAddressBalances(cleanupInterval)
 		defer n.Balances.StopCleanup()
 
@@ -591,6 +564,33 @@ func main() {
 				glog.Infoln("To update the broadcaster's maximum acceptable transcoding price per pixel, use the CLI or restart the broadcaster with the appropriate 'maxPricePerUnit' and 'pixelsPerUnit' values")
 			}
 		}
+
+		blockWatchCtx, cancel := context.WithCancel(context.Background())
+		defer cancel()
+
+		// Backfill events that the node has missed since its last seen block. This method will block
+		// and the node will not continue setup until it finishes
+		if err := blockWatcher.BackfillEventsIfNeeded(blockWatchCtx); err != nil {
+			glog.Errorf("Failed to backfill events: %v", err)
+			return
+		}
+
+		blockWatcherErr := make(chan error, 1)
+		go func() {
+			if err := blockWatcher.Watch(blockWatchCtx); err != nil {
+				blockWatcherErr <- fmt.Errorf("block watcher error: %v", err)
+			}
+		}()
+
+		go func() {
+			var err error
+			select {
+			case err = <-roundsWatcherErr:
+			case err = <-blockWatcherErr:
+			}
+
+			watcherErr <- err
+		}()
 	}
 
 	if *s3bucket != "" && *s3creds == "" || *s3bucket == "" && *s3creds != "" {
--- a/common/types.go
+++ b/common/types.go
@@ -14,6 +14,6 @@ type Broadcaster interface {
 
 type OrchestratorPool interface {
 	GetURLs() []*url.URL
-	GetOrchestrators(int, Broadcaster) ([]*net.OrchestratorInfo, error)
+	GetOrchestrators(int) ([]*net.OrchestratorInfo, error)
 	Size() int
 }
--- a/common/types.go
+++ b/common/types.go
@@ -1,0 +1,19 @@
+package common
+
+import (
+	"net/url"
+
+	ethcommon "github.com/ethereum/go-ethereum/common"
+	"github.com/livepeer/go-livepeer/net"
+)
+
+type Broadcaster interface {
+	Address() ethcommon.Address
+	Sign([]byte) ([]byte, error)
+}
+
+type OrchestratorPool interface {
+	GetURLs() []*url.URL
+	GetOrchestrators(int, Broadcaster) ([]*net.OrchestratorInfo, error)
+	Size() int
+}
--- a/common/types.go
+++ b/common/types.go
@@ -17,3 +17,9 @@ type OrchestratorPool interface {
 	GetOrchestrators(int) ([]*net.OrchestratorInfo, error)
 	Size() int
 }
+
+type OrchestratorStore interface {
+	OrchCount(filter *DBOrchFilter) (int, error)
+	SelectOrchs(filter *DBOrchFilter) ([]*DBOrch, error)
+	UpdateOrch(orch *DBOrch) error
+}
--- a/discovery/db_discovery.go
+++ b/discovery/db_discovery.go
@@ -6,6 +6,7 @@ import (
 	"math/big"
 	"net/url"
 	"strings"
+	"time"
 
 	ethcommon "github.com/ethereum/go-ethereum/common"
 	"github.com/livepeer/go-livepeer/common"
@@ -18,6 +19,11 @@ import (
 
 	"github.com/golang/glog"
 )
+
+var cacheRefreshInterval = 1 * time.Hour
+var getTicker = func() *time.Ticker {
+	return time.NewTicker(cacheRefreshInterval)
+}
 
 type ticketParamsValidator interface {
 	ValidateTicketParams(ticketParams *pm.TicketParams) error
@@ -199,13 +205,10 @@ func (dbo *DBOrchestratorPoolCache) cacheDBOrchs() error {
 		if orch == nil {
 			continue
 		}
-		dbOrch := ethOrchToDBOrch(orch)
 		numOrchs++
-		go getOrchInfo(dbOrch)
+		go getOrchInfo(orch)
 
 	}
-
-	var returnDBOrchs []*common.DBOrch
 
 	for i := 0; i < numOrchs; i++ {
 		select {
@@ -213,7 +216,6 @@ func (dbo *DBOrchestratorPoolCache) cacheDBOrchs() error {
 			if err := dbo.store.UpdateOrch(res); err != nil {
 				glog.Error("Error updating Orchestrator in DB: ", err)
 			}
-			returnDBOrchs = append(returnDBOrchs, res)
 		case err := <-errc:
 			glog.Errorln(err)
 		case <-ctx.Done():
@@ -247,13 +249,6 @@ func ethOrchToDBOrch(orch *lpTypes.Transcoder) *common.DBOrch {
 		ActivationRound:   orch.ActivationRound.Int64(),
 		DeactivationRound: orch.DeactivationRound.Int64(),
 	}
-	if orch.ActivationRound != nil {
-		dbO.ActivationRound = orch.ActivationRound.Int64()
-	}
-	if orch.DeactivationRound != nil {
-		dbO.DeactivationRound = orch.DeactivationRound.Int64()
-	}
-	return dbO
 }
 
 func pmTicketParams(params *net.TicketParams) *pm.TicketParams {
--- a/discovery/db_discovery.go
+++ b/discovery/db_discovery.go
@@ -6,11 +6,11 @@ import (
 	"math/big"
 	"net/url"
 	"strings"
-	"time"
 
 	ethcommon "github.com/ethereum/go-ethereum/common"
 	"github.com/livepeer/go-livepeer/common"
 	"github.com/livepeer/go-livepeer/core"
+	"github.com/livepeer/go-livepeer/eth"
 	lpTypes "github.com/livepeer/go-livepeer/eth/types"
 	"github.com/livepeer/go-livepeer/net"
 	"github.com/livepeer/go-livepeer/pm"
@@ -19,38 +19,59 @@ import (
 	"github.com/golang/glog"
 )
 
-var cacheRefreshInterval = 1 * time.Hour
-var getTicker = func() *time.Ticker {
-	return time.NewTicker(cacheRefreshInterval)
+type ticketParamsValidator interface {
+	ValidateTicketParams(ticketParams *pm.TicketParams) error
+}
+
+type roundsManager interface {
+	LastInitializedRound() *big.Int
+}
+
+type orchestratorStore interface {
+	OrchCount(filter *common.DBOrchFilter) (int, error)
+	SelectOrchs(filter *common.DBOrchFilter) ([]*common.DBOrch, error)
+	UpdateOrch(orch *common.DBOrch) error
 }
 
 type DBOrchestratorPoolCache struct {
-	node *core.LivepeerNode
+	store                 orchestratorStore
+	lpEth                 eth.LivepeerEthClient
+	ticketParamsValidator ticketParamsValidator
+	rm                    roundsManager
+	bcast                 common.Broadcaster
 }
 
-func NewDBOrchestratorPoolCache(node *core.LivepeerNode) *DBOrchestratorPoolCache {
+func NewDBOrchestratorPoolCache(ctx context.Context, node *core.LivepeerNode, rm roundsManager) (*DBOrchestratorPoolCache, error) {
 	if node.Eth == nil {
-		glog.Error("Could not refresh DB list of orchestrators: LivepeerNode nil")
-		return nil
+		return nil, fmt.Errorf("could not create DBOrchestratorPoolCache: LivepeerEthClient is nil")
 	}
 
-	_ = cacheRegisteredTranscoders(node)
+	dbo := &DBOrchestratorPoolCache{
+		store:                 node.Database,
+		lpEth:                 node.Eth,
+		ticketParamsValidator: node.Sender,
+		rm:                    rm,
+		bcast:                 core.NewBroadcaster(node),
+	}
 
-	ticker := getTicker()
-	go func(node *core.LivepeerNode) {
-		for _ = range ticker.C {
-			err := cacheRegisteredTranscoders(node)
-			if err != nil {
-				continue
-			}
-		}
-	}(node)
+	if err := dbo.cacheTranscoderPool(); err != nil {
+		return nil, err
+	}
 
-	return &DBOrchestratorPoolCache{node: node}
+	if err := dbo.pollOrchestratorInfo(ctx); err != nil {
+		return nil, err
+	}
+
+	return dbo, nil
 }
 
 func (dbo *DBOrchestratorPoolCache) getURLs() ([]*url.URL, error) {
-	orchs, err := dbo.node.Database.SelectOrchs(&common.DBOrchFilter{MaxPrice: server.BroadcastCfg.MaxPrice()})
+	orchs, err := dbo.store.SelectOrchs(
+		&common.DBOrchFilter{
+			MaxPrice:     server.BroadcastCfg.MaxPrice(),
+			CurrentRound: dbo.rm.LastInitializedRound(),
+		},
+	)
 	if err != nil || len(orchs) <= 0 {
 		return nil, err
 	}
@@ -76,12 +97,12 @@ func (dbo *DBOrchestratorPoolCache) GetOrchestrators(numOrchestrators int) ([]*n
 	}
 
 	pred := func(info *net.OrchestratorInfo) bool {
-		if dbo.node.Sender != nil {
-			if err := dbo.node.Sender.ValidateTicketParams(pmTicketParams(info.TicketParams)); err != nil {
-				return false
-			}
+
+		if err := dbo.ticketParamsValidator.ValidateTicketParams(pmTicketParams(info.TicketParams)); err != nil {
+			return false
 		}
 
+		// check if O's price is below B's max price
 		price := server.BroadcastCfg.MaxPrice()
 		if price != nil {
 			return big.NewRat(info.PriceInfo.PricePerUnit, info.PriceInfo.PixelsPerUnit).Cmp(price) <= 0
@@ -89,7 +110,7 @@ func (dbo *DBOrchestratorPoolCache) GetOrchestrators(numOrchestrators int) ([]*n
 		return true
 	}
 
-	orchPool := NewOrchestratorPoolWithPred(dbo.node, uris, pred)
+	orchPool := NewOrchestratorPoolWithPred(dbo.bcast, uris, pred)
 
 	orchInfos, err := orchPool.GetOrchestrators(numOrchestrators)
 	if err != nil || len(orchInfos) <= 0 {
@@ -100,29 +121,59 @@ func (dbo *DBOrchestratorPoolCache) GetOrchestrators(numOrchestrators int) ([]*n
 }
 
 func (dbo *DBOrchestratorPoolCache) Size() int {
-	return len(dbo.GetURLs())
+	count, _ := dbo.store.OrchCount(
+		&common.DBOrchFilter{
+			MaxPrice: server.BroadcastCfg.MaxPrice(),
+		},
+	)
+	return count
 }
 
-func cacheRegisteredTranscoders(node *core.LivepeerNode) error {
-	orchestrators, err := node.Eth.RegisteredTranscoders()
+func (dbo *DBOrchestratorPoolCache) cacheTranscoderPool() error {
+	orchestrators, err := dbo.lpEth.TranscoderPool()
 	if err != nil {
-		glog.Error("Could not refresh DB list of orchestrators: ", err)
-		return err
+		return fmt.Errorf("Could not refresh DB list of orchestrators: %v", err)
 	}
-	_, dbOrchErr := cacheDBOrchs(node, orchestrators)
-	if dbOrchErr != nil {
-		glog.Error("Could not refresh DB list of orchestrators: cacheDBOrchs err")
-		return err
+
+	for _, o := range orchestrators {
+		if err := dbo.store.UpdateOrch(ethOrchToDBOrch(o)); err != nil {
+			glog.Errorf("Unable to update orchestrator %v in DB: %v", o.Address.Hex(), err)
+		}
 	}
 
 	return nil
 }
 
-func cacheDBOrchs(node *core.LivepeerNode, orchs []*lpTypes.Transcoder) ([]*common.DBOrch, error) {
-	var dbOrchs []*common.DBOrch
-	if orchs == nil {
-		glog.Error("No new DB orchestrators created: no orchestrators found onchain")
-		return dbOrchs, nil
+func (dbo *DBOrchestratorPoolCache) pollOrchestratorInfo(ctx context.Context) error {
+	if err := dbo.cacheDBOrchs(); err != nil {
+		return err
+	}
+
+	ticker := getTicker()
+	go func() {
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case <-ticker.C:
+				if err := dbo.cacheDBOrchs(); err != nil {
+					glog.Errorf("unable to poll orchestrator info: %v", err)
+				}
+			}
+		}
+	}()
+
+	return nil
+}
+
+func (dbo *DBOrchestratorPoolCache) cacheDBOrchs() error {
+	orchs, err := dbo.store.SelectOrchs(
+		&common.DBOrchFilter{
+			CurrentRound: dbo.rm.LastInitializedRound(),
+		},
+	)
+	if err != nil {
+		return fmt.Errorf("could not retrieve orchestrators from DB: %v", err)
 	}
 
 	resc, errc := make(chan *common.DBOrch), make(chan error)
@@ -135,7 +186,7 @@ func cacheDBOrchs(node *core.LivepeerNode, orchs []*lpTypes.Transcoder) ([]*comm
 			errc <- err
 			return
 		}
-		info, err := serverGetOrchInfo(ctx, core.NewBroadcaster(node), uri)
+		info, err := serverGetOrchInfo(ctx, dbo.bcast, uri)
 		if err != nil {
 			errc <- err
 			return
@@ -164,7 +215,7 @@ func cacheDBOrchs(node *core.LivepeerNode, orchs []*lpTypes.Transcoder) ([]*comm
 	for i := 0; i < numOrchs; i++ {
 		select {
 		case res := <-resc:
-			if err := node.Database.UpdateOrch(res); err != nil {
+			if err := dbo.store.UpdateOrch(res); err != nil {
 				glog.Error("Error updating Orchestrator in DB: ", err)
 			}
 			returnDBOrchs = append(returnDBOrchs, res)
@@ -175,7 +226,8 @@ func cacheDBOrchs(node *core.LivepeerNode, orchs []*lpTypes.Transcoder) ([]*comm
 			break
 		}
 	}
-	return returnDBOrchs, nil
+
+	return nil
 }
 
 func parseURI(addr string) (*url.URL, error) {
@@ -193,9 +245,12 @@ func ethOrchToDBOrch(orch *lpTypes.Transcoder) *common.DBOrch {
 	if orch == nil {
 		return nil
 	}
+
 	return &common.DBOrch{
-		ServiceURI:   orch.ServiceURI,
-		EthereumAddr: orch.Address.String(),
+		ServiceURI:        orch.ServiceURI,
+		EthereumAddr:      orch.Address.String(),
+		ActivationRound:   orch.ActivationRound.Int64(),
+		DeactivationRound: orch.DeactivationRound.Int64(),
 	}
 }
 
--- a/discovery/db_discovery.go
+++ b/discovery/db_discovery.go
@@ -27,14 +27,8 @@ type roundsManager interface {
 	LastInitializedRound() *big.Int
 }
 
-type orchestratorStore interface {
-	OrchCount(filter *common.DBOrchFilter) (int, error)
-	SelectOrchs(filter *common.DBOrchFilter) ([]*common.DBOrch, error)
-	UpdateOrch(orch *common.DBOrch) error
-}
-
 type DBOrchestratorPoolCache struct {
-	store                 orchestratorStore
+	store                 common.OrchestratorStore
 	lpEth                 eth.LivepeerEthClient
 	ticketParamsValidator ticketParamsValidator
 	rm                    roundsManager
@@ -123,7 +117,8 @@ func (dbo *DBOrchestratorPoolCache) GetOrchestrators(numOrchestrators int) ([]*n
 func (dbo *DBOrchestratorPoolCache) Size() int {
 	count, _ := dbo.store.OrchCount(
 		&common.DBOrchFilter{
-			MaxPrice: server.BroadcastCfg.MaxPrice(),
+			MaxPrice:     server.BroadcastCfg.MaxPrice(),
+			CurrentRound: dbo.rm.LastInitializedRound(),
 		},
 	)
 	return count
@@ -252,6 +247,13 @@ func ethOrchToDBOrch(orch *lpTypes.Transcoder) *common.DBOrch {
 		ActivationRound:   orch.ActivationRound.Int64(),
 		DeactivationRound: orch.DeactivationRound.Int64(),
 	}
+	if orch.ActivationRound != nil {
+		dbO.ActivationRound = orch.ActivationRound.Int64()
+	}
+	if orch.DeactivationRound != nil {
+		dbO.DeactivationRound = orch.DeactivationRound.Int64()
+	}
+	return dbO
 }
 
 func pmTicketParams(params *net.TicketParams) *pm.TicketParams {
--- a/discovery/discovery_test.go
+++ b/discovery/discovery_test.go
@@ -595,6 +595,7 @@ func TestCachedPool_GetOrchestrators_MaxBroadcastPriceNotSet(t *testing.T) {
 			PriceInfo:  expPriceInfo,
 		}, nil
 	}
+
 	addresses := []string{}
 	for i := 0; i < 50; i++ {
 		addresses = append(addresses, "https://127.0.0.1:"+strconv.Itoa(8936+i))
@@ -965,8 +966,6 @@ func TestNewWHOrchestratorPoolCache(t *testing.T) {
 		return &net.OrchestratorInfo{Transcoder: "transcoder"}, nil
 	}
 
-	// mock livepeer node
-	node, _ := core.NewLivepeerNode(nil, "", nil)
 	perm = func(len int) []int { return rand.Perm(3) }
 
 	// assert created webhook pool is correct length
--- a/discovery/discovery_test.go
+++ b/discovery/discovery_test.go
@@ -1078,13 +1078,3 @@ func TestDeserializeWebhookJSON(t *testing.T) {
 	assert.Contains(err.Error(), "cannot unmarshal number")
 	assert.Empty(urls)
 }
-
-type orchTest struct {
-	EthereumAddr  string
-	ServiceURI    string
-	PricePerPixel int64
-}
-
-func toOrchTest(addr, serviceURI string, pricePerPixel int64) orchTest {
-	return orchTest{EthereumAddr: addr, ServiceURI: serviceURI, PricePerPixel: pricePerPixel}
-}
--- a/discovery/discovery_test.go
+++ b/discovery/discovery_test.go
@@ -13,12 +13,10 @@ import (
 	"testing"
 	"time"
 
-	ethcommon "github.com/ethereum/go-ethereum/common"
 	"github.com/golang/glog"
 	"github.com/livepeer/go-livepeer/common"
 	"github.com/livepeer/go-livepeer/core"
 	"github.com/livepeer/go-livepeer/eth"
-	lpTypes "github.com/livepeer/go-livepeer/eth/types"
 	"github.com/livepeer/go-livepeer/net"
 	"github.com/livepeer/go-livepeer/pm"
 	"github.com/livepeer/go-livepeer/server"
@@ -27,59 +25,23 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-type stubOrchestratorPool struct {
-	uris  []*url.URL
-	bcast server.Broadcaster
-}
-
-func stringsToURIs(addresses []string) []*url.URL {
-	var uris []*url.URL
-
-	for _, addr := range addresses {
-		uri, err := url.ParseRequestURI(addr)
-		if err == nil {
-			uris = append(uris, uri)
-		}
+func TestNewDBOrchestratorPoolCache_NilEthClient_ReturnsError(t *testing.T) {
+	assert := assert.New(t)
+	dbh, dbraw, err := common.TempDB(t)
+	defer dbh.Close()
+	defer dbraw.Close()
+	require := require.New(t)
+	require.Nil(err)
+	node := &core.LivepeerNode{
+		Database: dbh,
+		Eth:      nil,
+		Sender:   &pm.MockSender{},
 	}
-	return uris
-}
-
-func StubOrchestratorPool(addresses []string) *stubOrchestratorPool {
-	uris := stringsToURIs(addresses)
-	node, _ := core.NewLivepeerNode(nil, "", nil)
-	bcast := core.NewBroadcaster(node)
-
-	return &stubOrchestratorPool{bcast: bcast, uris: uris}
-}
-
-func StubOrchestrators(addresses []string) []*lpTypes.Transcoder {
-	var orchestrators []*lpTypes.Transcoder
-
-	for _, addr := range addresses {
-		address := ethcommon.BytesToAddress([]byte(addr))
-		transc := &lpTypes.Transcoder{ServiceURI: addr, Address: address}
-		orchestrators = append(orchestrators, transc)
-	}
-
-	return orchestrators
-}
-
-func TestNewOrchestratorPoolWithPred_NilEthClient_ReturnsNil_LogsError(t *testing.T) {
-	node, _ := core.NewLivepeerNode(nil, "", nil)
-	errorLogsBefore := glog.Stats.Error.Lines()
-	pool := NewOrchestratorPoolWithPred(node, nil, nil)
-	errorLogsAfter := glog.Stats.Error.Lines()
-	assert.NotZero(t, errorLogsAfter-errorLogsBefore)
-	assert.Nil(t, pool)
-}
-
-func TestNewDBOrchestratorPoolCache_NilEthClient_ReturnsNil_LogsError(t *testing.T) {
-	node, _ := core.NewLivepeerNode(nil, "", nil)
-	errorLogsBefore := glog.Stats.Error.Lines()
-	poolCache := NewDBOrchestratorPoolCache(node)
-	errorLogsAfter := glog.Stats.Error.Lines()
-	assert.NotZero(t, errorLogsAfter-errorLogsBefore)
-	assert.Nil(t, poolCache)
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	pool, err := NewDBOrchestratorPoolCache(ctx, node, &stubRoundsManager{})
+	assert.Nil(pool)
+	assert.EqualError(err, "could not create DBOrchestratorPoolCache: LivepeerEthClient is nil")
 }
 
 func TestDeadLock(t *testing.T) {
@@ -87,7 +49,7 @@ func TestDeadLock(t *testing.T) {
 	defer runtime.GOMAXPROCS(gmp)
 	var mu sync.Mutex
 	first := true
-	serverGetOrchInfo = func(ctx context.Context, bcast server.Broadcaster, orchestratorServer *url.URL) (*net.OrchestratorInfo, error) {
+	serverGetOrchInfo = func(ctx context.Context, bcast common.Broadcaster, orchestratorServer *url.URL) (*net.OrchestratorInfo, error) {
 		mu.Lock()
 		if first {
 			time.Sleep(100 * time.Millisecond)
@@ -114,7 +76,7 @@ func TestDeadLock_NewOrchestratorPoolWithPred(t *testing.T) {
 	defer runtime.GOMAXPROCS(gmp)
 	var mu sync.Mutex
 	first := true
-	serverGetOrchInfo = func(ctx context.Context, bcast server.Broadcaster, orchestratorServer *url.URL) (*net.OrchestratorInfo, error) {
+	serverGetOrchInfo = func(ctx context.Context, bcast common.Broadcaster, orchestratorServer *url.URL) (*net.OrchestratorInfo, error) {
 		mu.Lock()
 		if first {
 			time.Sleep(100 * time.Millisecond)
@@ -144,12 +106,7 @@ func TestDeadLock_NewOrchestratorPoolWithPred(t *testing.T) {
 		return true
 	}
 
-	orchestrators := StubOrchestrators(addresses)
-
-	node, _ := core.NewLivepeerNode(nil, "", nil)
-	node.Eth = &eth.StubClient{Orchestrators: orchestrators}
-
-	pool := NewOrchestratorPoolWithPred(node, uris, pred)
+	pool := NewOrchestratorPoolWithPred(nil, uris, pred)
 	infos, err := pool.GetOrchestrators(1)
 
 	assert.Nil(err, "Should not be error")
@@ -172,55 +129,50 @@ func TestPoolSize(t *testing.T) {
 	assert.NotZero(t, errorLogsAfter-errorLogsBefore)
 }
 
-func TestDbOrchestratorPoolCacheSize(t *testing.T) {
+func TestDBOrchestratorPoolCacheSize(t *testing.T) {
+	assert := assert.New(t)
 	dbh, dbraw, err := common.TempDB(t)
 	defer dbh.Close()
 	defer dbraw.Close()
 	require := require.New(t)
 	require.Nil(err)
 
-	node, _ := core.NewLivepeerNode(nil, "", nil)
-	node.Database = dbh
-	node.Eth = &eth.StubClient{}
+	sender := &pm.MockSender{}
+	node := &core.LivepeerNode{
+		Database: dbh,
+		Eth:      &eth.StubClient{},
+		Sender:   sender,
+	}
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
 
-	errorLogsBefore := glog.Stats.Error.Lines()
-	// No orchestrators on eth.Stubclient will result in no registered orchs found on chain
-	emptyPool := NewDBOrchestratorPoolCache(node)
-	errorLogsAfter := glog.Stats.Error.Lines()
+	emptyPool, err := NewDBOrchestratorPoolCache(ctx, node, &stubRoundsManager{})
+	require.NoError(err)
 	require.NotNil(emptyPool)
-	assert.Equal(t, 0, emptyPool.Size())
-	assert.NotZero(t, errorLogsAfter-errorLogsBefore)
+	assert.Equal(0, emptyPool.Size())
 
 	addresses := []string{"https://127.0.0.1:8936", "https://127.0.0.1:8937", "https://127.0.0.1:8938"}
 	orchestrators := StubOrchestrators(addresses)
-	node.Eth = &eth.StubClient{Orchestrators: orchestrators}
-	nonEmptyPool := NewDBOrchestratorPoolCache(node)
+	for _, o := range orchestrators {
+		dbh.UpdateOrch(ethOrchToDBOrch(o))
+	}
+
+	nonEmptyPool, err := NewDBOrchestratorPoolCache(ctx, node, &stubRoundsManager{})
+	require.NoError(err)
 	require.NotNil(nonEmptyPool)
-	assert.Equal(t, len(addresses), emptyPool.Size())
-}
-
-func TestCacheRegisteredTranscoders_GivenListOfOrchs_CreatesPoolCacheCorrectly(t *testing.T) {
-	dbh, dbraw, err := common.TempDB(t)
-	defer dbh.Close()
-	defer dbraw.Close()
-	require := require.New(t)
-	require.Nil(err)
-
-	addresses := []string{"https://127.0.0.1:8936", "https://127.0.0.1:8937", "https://127.0.0.1:8938"}
-	orchestrators := StubOrchestrators(addresses)
-
-	node, _ := core.NewLivepeerNode(nil, "", nil)
-	node.Database = dbh
-	node.Eth = &eth.StubClient{Orchestrators: orchestrators}
-
-	err = cacheRegisteredTranscoders(node)
-	require.Nil(err)
+	assert.Equal(len(addresses), nonEmptyPool.Size())
 }
 
 func TestNewDBOrchestratorPoolCache_GivenListOfOrchs_CreatesPoolCacheCorrectly(t *testing.T) {
+	expPriceInfo := &net.PriceInfo{
+		PricePerUnit:  999,
+		PixelsPerUnit: 1,
+	}
+	expTranscoder := "transcoderFromTest"
+	expPricePerPixel, _ := common.PriceToFixed(big.NewRat(999, 1))
 	var mu sync.Mutex
 	first := true
-	serverGetOrchInfo = func(ctx context.Context, bcast server.Broadcaster, orchestratorServer *url.URL) (*net.OrchestratorInfo, error) {
+	serverGetOrchInfo = func(ctx context.Context, bcast common.Broadcaster, orchestratorServer *url.URL) (*net.OrchestratorInfo, error) {
 		mu.Lock()
 		if first {
 			time.Sleep(100 * time.Millisecond)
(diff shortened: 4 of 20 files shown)


yondonfu:
Let's rebase this on top of master (to slim down the list of commits to be reviewed since many of them are now included in master) and then we can kick off review!

kyriediculous:
rebased

cmd/livepeer/livepeer.go
```diff
@@ -287,6 +287,9 @@ func main() {
 	}
 
 	watcherErr := make(chan error)
+	var roundsWatcher *watchers.RoundsWatcher
```
yondonfu, line 290: These variables can be declared in the else block below since we never need to use these variables when `*network == "offchain"`.

cmd/livepeer/livepeer.go
```diff
@@ -640,11 +641,16 @@ func main() {
...
 			}
-			n.OrchestratorPool = discovery.NewWebhookPool(n, whurl)
+			n.OrchestratorPool = discovery.NewWebhookPool(whurl)
 		} else if len(orchURLs) > 0 {
-			n.OrchestratorPool = discovery.NewOrchestratorPool(n, orchURLs)
+			n.OrchestratorPool = discovery.NewOrchestratorPool(orchURLs)
 		} else if *network != "offchain" {
-			n.OrchestratorPool = discovery.NewDBOrchestratorPoolCache(n)
+			dbOrchPoolCache := discovery.NewDBOrchestratorPoolCache(dbh, n.Eth, roundsWatcher, n.Sender)
+			if originalLastSeenBlock == nil || originalLastSeenBlock.Cmp(currentRoundStartBlock) < 0 {
```
yondonfu: I think making this transcoder pool caching logic post event backfill non-broadcaster specific would be helpful because it would allow an orchestrator to also rely on its DB for the state of the active set.

discovery/db_discovery.go
```diff
@@ -193,10 +212,17 @@ func ethOrchToDBOrch(orch *lpTypes.Transcoder) *common.DBOrch {
 	if orch == nil {
 		return nil
 	}
-	return &common.DBOrch{
+	dbO := &common.DBOrch{
 		ServiceURI:   orch.ServiceURI,
 		EthereumAddr: orch.Address.String(),
 	}
+	if orch.ActivationRound != nil {
```
yondonfu: I don't think `ActivationRound` and `DeactivationRound` will ever be nil since the default value returned from the contract will be 0.

discovery/db_discovery.go
```diff
@@ -19,38 +18,52 @@ import (
...
+	lpEth                 eth.LivepeerEthClient
+	ticketParamsValidator ticketParamsValidator
+	rm                    roundsManager
 }
 
-func NewDBOrchestratorPoolCache(node *core.LivepeerNode) *DBOrchestratorPoolCache {
-	if node.Eth == nil {
-		glog.Error("Could not refresh DB list of orchestrators: LivepeerNode nil")
+func NewDBOrchestratorPoolCache(store common.OrchestratorStore, lpEth eth.LivepeerEthClient, rm roundsManager, ticketParamsValidator ticketParamsValidator) *DBOrchestratorPoolCache {
+	if store == nil {
```
yondonfu: I don't think any of these params could be nil besides `lpEth` (in on-chain mode) so we could just have a single nil check for `lpEth`

discovery/db_discovery.go
```diff
@@ -19,38 +18,52 @@ import (
...
+		glog.Error("could not create DBOrchestratorPoolCache: store is nil")
 		return nil
 	}
 
-	_ = cacheRegisteredTranscoders(node)
+	if lpEth == nil {
+		glog.Error("could not create DBOrchestratorPoolCache: LivepeerNode is nil")
```
yondonfu: Error message should be updated to not say "LivepeerNode"

yondonfu (changes requested):

cmd/livepeer/livepeer.go
kyriediculous: That can be an option but I don't see the purpose of it as O's don't use that data for the time being? 

You could also make an argument for the opposite, saying that we should only backfill and cache the pool when the node is a B because O's currently don't have a purpose for event backfilling (perhaps for unbonding locks maybe).

So I guess the bigger question is to address this now or when we rework livepeer.go ?

cmd/livepeer/livepeer.go
kyriediculous: These variables are also being used when setting up `DBOrchestratorPoolCache` at L648 , which is outside of the scope of the `if *network == "offchain {} else` block. 

We could move the lines that initiate `DBOrchestratorPoolCache` because the conditional for is a duplicate of `if *network == "offchain` but that requires shuffling some of the other conditionals. It also strips away part of the code for starting an `orchestratorPool` from the code block that handles this logic, which wouldn't be very optimal in terms of readability I think?

discovery/db_discovery.go
kyriediculous: fixed in 74f273ef

discovery/db_discovery.go
kyriediculous: fixed in 74f273ef

discovery/db_discovery.go
kyriediculous: fixed in 74f273e

cmd/livepeer/livepeer.go
yondonfu: Oops missed this bit since I didn't unfold the code enough in the review comparison view - gotcha

cmd/livepeer/livepeer.go
yondonfu: > That can be an option but I don't see the purpose of it as O's don't use that data for the time being?

Yes O doesn't use that data at the moment, but I don't think we should rule out the possibility. For example, if O had this data, then if we implement an active status check for O whenever it receives a segment (to make sure that O has access to B's reserve) then O can query its DB instead of using a contract call that involves a network request.

Furthermore, this data could generally be useful regardless of what mode that the node is running in since it could be queryable via the node's API.

discovery/db_discovery.go
```diff
@@ -19,38 +18,37 @@ import (
...
+		glog.Error("could not create DBOrchestratorPoolCache: LivepeerEthClient is nil")
 		return nil
 	}
 
-	_ = cacheRegisteredTranscoders(node)
-
-	ticker := getTicker()
-	go func(node *core.LivepeerNode) {
-		for _ = range ticker.C {
-			err := cacheRegisteredTranscoders(node)
```
yondonfu, line 42: While we no longer need to query for on-chain pool at regular intervals, we still want to call `GetOrchestrator()` on known orchestrators at regular intervals to fetch off-chain information. We no longer do that in the current set of commits since this code was removed.

yondonfu (changes requested):

cmd/livepeer/livepeer.go
kyriediculous: We could leverage this implementation for both O and B, but I don't think that's optimal for O. 

O would only care about `TranscoderActivated` and `TranscoderDeactivated` events pertaining to itself, not the entire set. 

Since `OrchestratorWatcher` is the writer to the DB for these events and because we already run `OrchestratorWatcher` when `NodeType == OrchestratorNode`, I suggest the following:
- Add an address filter option to `OrchestratorWatcher` for events only belonging to certain orchestrators
- Add an `isActive() (bool, error)` method to `core.Orchestrator` that queries `db.SelectOrchs` with a `currentRound` filter., since O will only write events about itself to the database this query will return 1 result when O is active and nil when O is inactive
- Leverage `orch.isActive()` in `server/ServeSegment()`

yondonfu: > O would only care about TranscoderActivated and TranscoderDeactivated events pertaining to itself, not the entire set.

Yep good point.

I do think it could be a useful for a node, regardless of whether it is running in broadcaster mode or orchestrator mode, to be able to compute the active set for the current round. For example, a developer building on top of the node's API could hit an endpoint that responds with the active set which might be used to compare each orchestrator's information. What do you think?

kyriediculous: I don't think we have put in sufficient work and thought into this to make a decision on this right now. 

We still have to throughly think about what we want such API to look like and perhaps even what APIs we want developers to use for what data. 

To illustrate: would it be worth the hassle for an orchestrator to periodically query other orchestrators for their info, just to have that off-chain orchestrator data available for an API endpoint or would an orchestrator list view fetched through eg. the livepeer subgraph suffice ? One should also consider how much this endpoint will actually be hit compared to the continuous computational resources it expends as there are alternatives for it.

yondonfu: I agree with those points.

> would it be worth the hassle for an orchestrator to periodically query other orchestrators for their info, just to have that off-chain orchestrator data available for an API endpoint

To clarify I was *not* suggesting also having the orchestrator fetch off-chain info of other orchestrators. I was only suggesting having the orchestrator compute the state of the active set i.e. the addresses in the set.

> One should also consider how much this endpoint will actually be hit compared to the continuous computational resources it expends as there are alternatives for it.

To clarify I was *not* suggesting we focus on creating that endpoint now. I was only mentioning one potential use case for the node to always be aware of the state of the current active set. 

I hear you on not trying to predict the usefulness of a node always being aware of the current active set - if it ends up being necessary we can make the change when the time comes and it shouldn't be that involved of a change.

We can use your suggestion of adding a filter to `OrchestratorWatcher` when the node is in orchestrator mode, but I recommend doing that outside of this PR.

discovery/db_discovery.go
kyriediculous: I feel like there's several options here

1. Start a simple go routine that queries the DB for all orchestrators that are active (we don't care about inactive ones) , fetch their off-chain info and write the price to the DB. This routine runs seperately from any other functionality (similar to the old caching loop). 

2. Remove MaxPrice filter when querying the DB, solely query for active orchestrators and only worry about filtering for max price in the `predicate` function. This would gives us up-to-date pricing information for active O's on each `OrchestratorPool.GetOrchestrators()`

3. Work out a `OrchestratorInfoPoller` implementation as described in [this doc](https://hackmd.io/SiCcxJXPTTeSJ1HsZEekmA)

We could go for 1 or 2 and implement 3 at a later stage, or bite the bullet and go for 3 now. But since you want to get the release out and `OrchestratorInfoPoller` needs some extra thought I would suggest going for option 2.  WDYT?

yondonfu: I would go with option 1. We ultimately want the orchestrator to fetch off-chain prices at a regular interval so that it can filter by price while constructing its working set for a stream so that it only needs to send requests to orchestrators that meet its price requirements. We also fetch & validate the orchestrator's info in `GetOrchestrators()`, but this is just a sanity check in case the orchestrator's price is no longer valid. Option 1 preserves this behavior and then at a later point we can consider option 3 as a code quality improvement.

kyriediculous: Added in 0be3692

Not exactly the same as before because we need a `common.Broadcaster` instance to pass into `server.GetOrchestratorInfo` to generate the request, but quite similar.

kyriediculous:
114d962 moves `orchTest` and `toOrchtest` to `stub.go`

discovery/db_discovery.go
j0sh: @kyriediculous Just read the writeup for the OrchestratorInfoPoller - good stuff, thanks for sharing! Just one point I wanted to ask about:

> Benefits of caching OrchestratorInfo
> Reduces the amount of requests sent over the wire by caching, rather than fetching the data each time selectOrchestrator is ran

Does this mean the reducing number external calls to reconstruct the state of the current orchestrator set - the current intent of the orch DB polling?

Or is the proposal here more around actually avoiding *all* OrchestratorInfo requests at selection time?

Calling OrchestratorInfo at selection time is still something we should still do; as noted in the hackmd, the data can easily go stale, orchestrators may become unavailable, etc.

While the poller's responsibility is to update the orch DB to help filter out the list as a first pass (minimizes blockchain lookups, excludes clearly unfit orchestrators), we should still be calling OrchestratorInfo at selection time to ensure up-to-date information on pricing and availability.

discovery/db_discovery.go
```diff
@@ -221,3 +227,63 @@ func pmTicketParams(params *net.TicketParams) *pm.TicketParams {
...
+func (dbo *DBOrchestratorPoolCache) PollOrchestratorInfo(bcast common.Broadcaster) {
+	ticker := getTicker()
+	go func() {
+		for range ticker.C {
+			dbo.pollOrchestratorInfo(bcast)
+		}
+	}()
+}
+
+func (dbo *DBOrchestratorPoolCache) pollOrchestratorInfo(bcast common.Broadcaster) {
```
yondonfu: Looks like most of the logic in this method is duplicated from `cacheDBOrchs()`. I think we can reuse some of the existing code - something along the lines of:

- Updating `cacheDBOrchs()` to accept `orchs []*common.DBOrch` instead of `orchs []*lpTypes.Transcoder`. We would also remove the loop that converts each `*lpTypes.Transcoder` into a `*common.DBOrch`
- Updating `CacheTranscoderPool()` to convert the list of orchestrators fetched from the contract into `[]*common.DBOrch` which can be passed to `cacheDBOrchs()`
- Updating this method to pass `orchs` into `cacheDBOrchs()`

What do you think?

kyriediculous: Is 7f8d8795 sorta what you're thinking of here?

discovery/db_discovery.go
```diff
@@ -212,3 +221,21 @@ func pmTicketParams(params *net.TicketParams) *pm.TicketParams {
 		Seed:              new(big.Int).SetBytes(params.Seed),
 	}
 }
+
+func (dbo *DBOrchestratorPoolCache) PollOrchestratorInfo(bcast common.Broadcaster) {
+	ticker := getTicker()
+	go func() {
+		for range ticker.C {
```
yondonfu: Since the first tick will not occur until after `cacheRefreshInterval`, we will not cache any off-chain info until after `cacheRefreshInterval`. The [old code](https://github.com/livepeer/go-livepeer/blob/master/discovery/db_discovery.go#L37) cached off-chain info *before* starting the ticker loop - we could do the same here.

discovery/db_discovery.go
```diff
@@ -212,3 +221,21 @@ func pmTicketParams(params *net.TicketParams) *pm.TicketParams {
...
+			orchs, err := dbo.store.SelectOrchs(
+				&common.DBOrchFilter{
+					CurrentRound: dbo.rm.LastInitializedRound(),
+				},
+			)
+
+			if err != nil {
+				glog.Errorf("could not get orchestrators from database: %v", err)
+			}
+			dbo.cacheDBOrchs(bcast, orchs)
```
yondonfu: Need to handle error here

discovery/db_discovery.go
```diff
@@ -212,3 +221,21 @@ func pmTicketParams(params *net.TicketParams) *pm.TicketParams {
...
+	go func() {
+		for range ticker.C {
+			orchs, err := dbo.store.SelectOrchs(
+				&common.DBOrchFilter{
+					CurrentRound: dbo.rm.LastInitializedRound(),
+				},
+			)
+
+			if err != nil {
+				glog.Errorf("could not get orchestrators from database: %v", err)
```
yondonfu: If we encounter an error here we should `continue`

discovery/db_discovery.go
yondonfu: At the moment, this loop will never exit and thus the goroutine will never exit gracefully. One fix could be to pass in a `context.Context` into `PollOrchestratorInfo()` and then exit the loop when a `select` receives a value on the context's `Done()` channel. The caller of this method could create the context with a cancel function which is called on program exit.

cmd/livepeer/livepeer.go
```diff
@@ -640,11 +641,17 @@ func main() {
...
-			n.OrchestratorPool = discovery.NewWebhookPool(n, whurl)
+			n.OrchestratorPool = discovery.NewWebhookPool(whurl)
 		} else if len(orchURLs) > 0 {
-			n.OrchestratorPool = discovery.NewOrchestratorPool(n, orchURLs)
+			n.OrchestratorPool = discovery.NewOrchestratorPool(orchURLs)
 		} else if *network != "offchain" {
-			n.OrchestratorPool = discovery.NewDBOrchestratorPoolCache(n)
+			dbOrchPoolCache := discovery.NewDBOrchestratorPoolCache(dbh, n.Eth, roundsWatcher, n.Sender)
+			bcast := core.NewBroadcaster(n)
+			dbOrchPoolCache.PollOrchestratorInfo(bcast)
```
yondonfu: If `PollOrchestratorInfo()` caches off-chain info upon being called (see my comment in this review specifically on `PollOrchestratorInfo()` further down), then I think we should call the method after we call `CacheTranscoderPool()` to make sure that `PollOrchestratorInfo()` has access to an up-to-date list of active set members.

cmd/livepeer/livepeer.go
```diff
@@ -640,11 +641,17 @@ func main() {
...
+			if originalLastSeenBlock == nil || originalLastSeenBlock.Cmp(currentRoundStartBlock) < 0 {
+				err = dbOrchPoolCache.CacheTranscoderPool(bcast)
```
yondonfu: Need to handle error here. I think we should just return/exit if we hit an error here since the broadcaster probably shouldn't continue operating in on-chain mode if it is unable to compute the active set. What do you think?

discovery/discovery_test.go
```diff
@@ -335,26 +267,104 @@ func TestNewDBOrchestratorPoolCache_TestURLs_Empty(t *testing.T) {
...
+	require.Equal(pool.Size(), 3)
+	dbOrchs, err := pool.store.SelectOrchs(nil)
+	require.Nil(err)
+	require.Len(dbOrchs, 3)
+	for _, o := range dbOrchs {
+		test := toOrchTest(o.EthereumAddr, o.ServiceURI, o.PricePerPixel)
+		require.Contains(testOrchs, test)
+	}
+
+	cacheRefreshInterval = 1 * time.Second
```
yondonfu: Possible to use a smaller value here and for the subsequent sleep to avoid slowing down the tests too much?

discovery/discovery_test.go
```diff
@@ -335,26 +267,104 @@ func TestNewDBOrchestratorPoolCache_TestURLs_Empty(t *testing.T) {
...
+	returnInfo = polledOrchInfo
+	pool.PollOrchestratorInfo(nil)
+	time.Sleep(3 * time.Second)
+	dbOrchs, err = pool.store.SelectOrchs(nil)
+	require.Nil(err)
+	require.Len(dbOrchs, 3)
+	expPrice, _ = common.PriceToFixed(big.NewRat(1, 1))
+	for _, o := range dbOrchs {
+		t.Log(o)
```
yondonfu: Can remove?

discovery/discovery_test.go
```diff
@@ -335,26 +267,104 @@ func TestNewDBOrchestratorPoolCache_TestURLs_Empty(t *testing.T) {
...
+	require.Nil(err)
+	require.Len(dbOrchs, 3)
+	for _, o := range dbOrchs {
+		test := toOrchTest(o.EthereumAddr, o.ServiceURI, o.PricePerPixel)
+		require.Contains(testOrchs, test)
+	}
+
+	cacheRefreshInterval = 1 * time.Second
+	returnInfo = polledOrchInfo
+	pool.PollOrchestratorInfo(nil)
```
yondonfu: Something that could be nice to test is that based on the `cacheRefreshInterval` value and the sleep time that we actually requested info for a URI a certain # of times.

yondonfu (changes requested):

discovery/db_discovery.go
yondonfu: Yep!

discovery/db_discovery.go
kyriediculous: Thanks @j0sh ! 
My idea initially was to abolish any network requests when selecting orchestrators but when I took a deeper look it's not as easy. In order to not have stale data we would have to (apart from polling at regular intervals) write any `OrchestratorInfo` response included in a call to `SubmitSegment` to the database, injecting the DB as a dependency in `server` doesn't seem quite feasible however. So far the only option to achieve that seems to be adding a method to `core.Broadcaster`, then in submit segment we can call `BroadcastSession.BroadCaster.UpdateOrchestratorInfo`  which under the hood just calls `db.UpdateOrch`.

Alternatively, we can add an `OrchestratorStore` instance to every `BroadcastSession`, which is easy to do because `broadcaster.selectOrchestrator` takes in `LivepeerNode` as an argument on which we can access `db`. 

This structure would prevent stale data due to polling and a feedback loop I think. It would not account for O's that have dropped their connection though. So I guess the tradeoff there is faster "warm up time" vs the risk of running into an O with a dropped connection. 

The O would have had dropped is connection since the last refresh interval though and O's would be incentivized to have as high uptime as possible to not affect their profitability and thus placement in the active orchestrator set. It's not optimal but something to consider ? I think it the probability of a B running into two dropped O's in a row when starting their stream (and thus incur a high warm up time) is fairly insignificant as well.

discovery/db_discovery.go
kyriediculous: I just checked and we're never actually returning an error from `cacheDBOrchs` so instead I opt for removing `error` as return param

cmd/livepeer/livepeer.go
kyriediculous: I don't think so because that will lead to a double operation of fetching off-chain info for the orchestrators in the pool in the case where `originalLastSeenBlock.Cmp(currentRoundStartBlock) < 0` because  `CacheTranscoderPool`will also fetch off-chain info.

If we call it before that block we will only fetch orchestrator info for orchestrators that are already in the database through event backfilling.

discovery/discovery_test.go
kyriediculous: Good idea, would the new test in 9458ef19 suffice?

discovery/discovery_test.go
kyriediculous: fixed in 9458ef19

discovery/discovery_test.go
kyriediculous: Using miliseconds now in 9458ef19

cmd/livepeer/livepeer.go
kyriediculous: fixed in 9458ef19

discovery/db_discovery.go
kyriediculous: fixed in 9458ef19

discovery/db_discovery.go
kyriediculous: fixed in 9458ef19

discovery/db_discovery.go
kyriediculous: fixed in 9458ef19

kyriediculous:
af26991 fixes up `TestNewDBOrchestratorPoolCache_TestURLs` to properly check bad URL's aren't being inserted

discovery/db_discovery.go
j0sh: > My idea initially was to abolish any network requests when selecting orchestrators

Orchestrator selection is usually not time sensitive [1], so I'm not really sure this needs to be a goal? What else are we trying to achieve with this change?

The current approach (before this PR) is a good compromise between the trade-offs of immediacy and freshness. Periodic polling fetches off-chain information, which helps filter out orchestrators as a first pass during selection time. Then actual selection is an opportunity to confirm pricing and availability as needed, without running the risk of having stale data.

[1] The only time O selection blocks a pending segment is for the very first request establishing a session with the HTTP ingest - but there are ways to work around this. RTMP and mid-session refreshes don't typically have issues with refreshes or O selection, as long as the working set of Os is sufficiently large (>2).

discovery/db_discovery.go
```diff
@@ -222,20 +217,43 @@ func pmTicketParams(params *net.TicketParams) *pm.TicketParams {
...
 		for range ticker.C {
-			orchs, err := dbo.store.SelectOrchs(
-				&common.DBOrchFilter{
-					CurrentRound: dbo.rm.LastInitializedRound(),
-				},
-			)
-
-			if err != nil {
-				glog.Errorf("could not get orchestrators from database: %v", err)
+			select {
```
yondonfu: I believe this `select` statement will only run when the ticket ticks. As a result, if the context is cancelled in between ticks, this loop would not receive the cancellation. We could make sure that the loop will always receive the cancellation by doing something like this:

```
for {
    case <-ticker.C:
        // Do something
    case <-ctx.Done():
        return
}
```

discovery/db_discovery.go
```diff
@@ -222,20 +217,43 @@ func pmTicketParams(params *net.TicketParams) *pm.TicketParams {
...
+			case <-ctx.Done():
+				break
+			default:
+				if err := dbo.pollOrchestratorInfo(bcast); err != nil {
+					glog.Error(err)
+					continue
```
yondonfu: If we're not doing anything besides calling `pollOrchestratorInfo()` in this loop we can remove the `continue`

cmd/livepeer/livepeer.go
```diff
@@ -647,10 +647,17 @@ func main() {
...
-			dbOrchPoolCache.PollOrchestratorInfo(bcast)
+
+			ctx, cancel := context.WithCancel(context.Background())
+			defer cancel()
+			dbOrchPoolCache.PollOrchestratorInfo(ctx, bcast)
+
 			if originalLastSeenBlock == nil || originalLastSeenBlock.Cmp(currentRoundStartBlock) < 0 {
-				err = dbOrchPoolCache.CacheTranscoderPool(bcast)
+				if err := dbOrchPoolCache.CacheTranscoderPool(bcast); err != nil {
+					glog.Errorf("Unable to cache transcoder pool: %v", err)
```
yondonfu: Return (aka exit) if we hit an error here?

discovery/db_discovery.go
```diff
@@ -24,33 +24,37 @@ var getTicker = func() *time.Ticker {
...
+	store                 common.OrchestratorStore
+	lpEth                 eth.LivepeerEthClient
+	ticketParamsValidator ticketParamsValidator
+	rm                    roundsManager
 }
 
-func NewDBOrchestratorPoolCache(node *core.LivepeerNode) *DBOrchestratorPoolCache {
-	if node.Eth == nil {
-		glog.Error("Could not refresh DB list of orchestrators: LivepeerNode nil")
+func NewDBOrchestratorPoolCache(store common.OrchestratorStore, lpEth eth.LivepeerEthClient, rm roundsManager, ticketParamsValidator ticketParamsValidator) *DBOrchestratorPoolCache {
```
j0sh: Is this really an improvement? What problem is being solved with these changes, other than the desire not to store the LivepeerNode? (Why?)

* This constructor goes from 1 parameter to 4 (and most of these params can still be derived from the original param), and the internal state goes from 1 value to 4

* We add another parameter to the public OrchestratorPool interface which means changing *all* call sites *with the same parameter that was originally removed*

* We further expand the exposed surface area of DBOrchestratorPoolCache with `CacheTranscoderPool` which drags more of the internal logic outside the DB code (and it still requires the original parameter that we removed...)

Apologies for being blunt, but this doesn't smell right to me. If there are deeper structural issues that are a concern, we can discuss, but this seems to be making things *worse* by fracturing the API all over the place.

j0sh (changes requested):

discovery/discovery_test.go
```diff
@@ -350,20 +350,106 @@ func TestNewDBOrchestorPoolCache_PollOrchestratorInfo(t *testing.T) {
...
+	dbOrchs, err := pool.store.SelectOrchs(nil)
+	require.Nil(err)
+	require.Len(dbOrchs, 3)
+	for _, o := range dbOrchs {
+		test := toOrchTest(o.EthereumAddr, o.ServiceURI, o.PricePerPixel)
+		require.Contains(testOrchs, test)
+	}
+
+	callCount := 0
+	serverGetOrchInfo = func(ctx context.Context, bcast common.Broadcaster, orchestratorServer *url.URL) (*net.OrchestratorInfo, error) {
```
yondonfu: Instead of re-assigning this variable to a tweaked version of the original function defined further up in this test what if we include `callCount` in the original definition and then after the setup phase of this test (i.e. everything before this line) we store the current value of `callCount` and then after we run `PollOrchestratorInfo()` and wait through the sleep time, we compare it against the new value of `callCount`. I think this could help slim down this test a bit.

discovery/discovery_test.go
```diff
@@ -350,20 +350,106 @@ func TestNewDBOrchestorPoolCache_PollOrchestratorInfo(t *testing.T) {
...
 	require.Nil(err)
 	require.Len(dbOrchs, 3)
 	expPrice, _ = common.PriceToFixed(big.NewRat(1, 1))
 	for _, o := range dbOrchs {
-		t.Log(o)
 		assert.Equal(o.PricePerPixel, expPrice)
 	}
 }
 
+func TestNewDBOrchestorPoolCache_PollOrchestratorInfo_CallCount(t *testing.T) {
```
yondonfu: The majority of this test is duplicated from `TestNewDBOrchestorPoolCache_PollOrchestratorInfo` - what do you think about just moving the call count check there?

discovery/discovery_test.go
```diff
@@ -350,20 +350,106 @@ func TestNewDBOrchestorPoolCache_PollOrchestratorInfo(t *testing.T) {
...
+		}, nil
+	}
+
+	cacheRefreshInterval = 200 * time.Millisecond
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	pool.PollOrchestratorInfo(ctx, nil)
+	time.Sleep(1100 * time.Millisecond)
+
+	// called serverGetOrchInfo 3 + 1100 / 200 * 3 = 18 times
```
yondonfu: I'm wary of checking for an exact # of polls here as that feels like it could result in fragile test. What about just checking that it was called >= 2 times?

yondonfu (changes requested):

cmd/livepeer/livepeer.go
yondonfu: At the moment, we will fetch off-chain info for all Os included in the transcoder pool and then when we call `PollOrchestratorInfo()` we will immediately fetch off-chain info for the same Os *again*. Seems simpler to keep the act of fetching off-chain info separate from the act of caching the transcoder pool with the latter just updating addresses in the DB instead of also fetching off-chain info for all the addresses. The former would guarantee that we have an up-to-date view of the active set and *then* we can start polling for off-chain info.

discovery/db_discovery.go
kyriediculous: I see what you'r' saying but we'd still end up with 3 params in the constructor and 3 state fields even when using `LivepeerNode` so one additional param/field doesn't seem that bad to me?

We only use two of the fields on `LivepeerNode` so to me it seems that the only reason for passing in the entire object is to be able to create `Broadcaster` instances, but why create new instances when the caller (`server`) already creates these instances ?  And even if one would create  `Broadcaster` instances, it wouldn't require the entire `LivepeerNode` as a dependency because it would only use `LivepeerEthClient` methods for the time being. 

So why is `LivepeerNode` needed at all here? I see `LivepeerNode` as an aggregate of different services to operate the node, to me it feels really weird to pass in an aggregate to construct one of its the sub-services.

cmd/livepeer/livepeer.go
kyriediculous: right, I see what you're saying. makes sense.

discovery/db_discovery.go
kyriediculous: I assume we still want the select statement though but loop indefinitely rather than over the ticker ? 
```
for {
  select {
    case <- ticker.C: 
       // pollOrchestratorInfo
    case <- ctx.Done: 
      // break
}
}
```

yondonfu: Oops yep my example code snippet is missing the select statement to wrap those cases.

discovery/db_discovery.go
j0sh: Again: what's the actual problem this solves? What do we gain in exchange for fracturing the API like this?

> we'd still end up with 3 params in the constructor and 3 state fields even when using LivepeerNode

How's that? You have: the following fields in LivepeerNode: `Database`, `Eth` and `Sender` . Use those fields. Even the RoundsWatcher could be avoided if the results were written to DB.

> the only reason for passing in the entire object is to be able to create Broadcaster instances

If that's really the root of the concern, then this is much simpler:

```diff
diff --git a/core/broadcaster.go b/core/broadcaster.go
index 401f618e..33fd87f8 100644
--- a/core/broadcaster.go
+++ b/core/broadcaster.go
@@ -7,24 +7,15 @@ import (
 
 // Broadcaster RPC interface implementation
 
-type broadcaster struct {
-	node *LivepeerNode
-}
-
-func (bcast *broadcaster) Sign(msg []byte) ([]byte, error) {
-	if bcast.node == nil || bcast.node.Eth == nil {
+func (node *LivepeerNode) Sign(msg []byte) ([]byte, error) {
+	if node == nil || node.Eth == nil {
 		return []byte{}, nil
 	}
-	return bcast.node.Eth.Sign(crypto.Keccak256(msg))
+	return node.Eth.Sign(crypto.Keccak256(msg))
 }
-func (bcast *broadcaster) Address() ethcommon.Address {
-	if bcast.node == nil || bcast.node.Eth == nil {
+func (node *LivepeerNode) Address() ethcommon.Address {
+	if node == nil || node.Eth == nil {
 		return ethcommon.Address{}
 	}
-	return bcast.node.Eth.Account().Address
-}
-func NewBroadcaster(node *LivepeerNode) *broadcaster {
-	return &broadcaster{
-		node: node,
-	}
+	return node.Eth.Account().Address
 }
diff --git a/discovery/db_discovery.go b/discovery/db_discovery.go
index c009662c..8bba828f 100644
--- a/discovery/db_discovery.go
+++ b/discovery/db_discovery.go
@@ -135,7 +135,7 @@ func cacheDBOrchs(node *core.LivepeerNode, orchs []*lpTypes.Transcoder) ([]*comm
 			errc <- err
 			return
 		}
-		info, err := serverGetOrchInfo(ctx, core.NewBroadcaster(node), uri)
+		info, err := serverGetOrchInfo(ctx, node, uri)
 		if err != nil {
 			errc <- err
 			return
diff --git a/discovery/discovery.go b/discovery/discovery.go
index 6df9dbfc..6046ba21 100644
--- a/discovery/discovery.go
+++ b/discovery/discovery.go
@@ -41,8 +41,7 @@ func NewOrchestratorPool(node *core.LivepeerNode, uris []*url.URL) *orchestrator
 		randomizedUris = append(randomizedUris, uri)
 	}
 
-	bcast := core.NewBroadcaster(node)
-	return &orchestratorPool{bcast: bcast, uris: randomizedUris}
+	return &orchestratorPool{bcast: node, uris: randomizedUris}
 }
 
 func NewOrchestratorPoolWithPred(node *core.LivepeerNode, addresses []*url.URL, pred func(*net.OrchestratorInfo) bool) *orchestratorPool {
diff --git a/discovery/discovery_test.go b/discovery/discovery_test.go
index 998c22a0..c2e2b7e4 100644
--- a/discovery/discovery_test.go
+++ b/discovery/discovery_test.go
@@ -47,9 +47,8 @@ func stringsToURIs(addresses []string) []*url.URL {
 func StubOrchestratorPool(addresses []string) *stubOrchestratorPool {
 	uris := stringsToURIs(addresses)
 	node, _ := core.NewLivepeerNode(nil, "", nil)
-	bcast := core.NewBroadcaster(node)
 
-	return &stubOrchestratorPool{bcast: bcast, uris: uris}
+	return &stubOrchestratorPool{bcast: node, uris: uris}
 }
 
 func StubOrchestrators(addresses []string) []*lpTypes.Transcoder {
diff --git a/server/broadcast.go b/server/broadcast.go
index 6a09977c..739cba91 100644
--- a/server/broadcast.go
+++ b/server/broadcast.go
@@ -182,8 +182,6 @@ func selectOrchestrator(n *core.LivepeerNode, params *streamParameters, cpl core
 		return nil, errDiscovery
 	}
 
-	rpcBcast := core.NewBroadcaster(n)
-
 	tinfos, err := n.OrchestratorPool.GetOrchestrators(count)
 	if len(tinfos) <= 0 {
 		glog.Info("No orchestrators found; not transcoding. Error: ", err)
@@ -222,7 +220,7 @@ func selectOrchestrator(n *core.LivepeerNode, params *streamParameters, cpl core
 		}
 
 		session := &BroadcastSession{
-			Broadcaster:      rpcBcast,
+			Broadcaster:      n,
 			ManifestID:       params.mid,
 			Profiles:         params.profiles,
 			OrchestratorInfo: tinfo,

```

discovery/discovery_test.go
kyriediculous: Ah yes we could just reset `callCount` to 0 as well after calling `CacheTranscoderPool` as well in fact.

discovery/discovery_test.go
kyriediculous: Yeah much better on second thought when incorporating the comment above

discovery/discovery_test.go
kyriediculous: I did run the test quite a few times and didn't run across any anomalies but if it's really a concern we can do a   `16<count<20` range or something like that?

discovery/db_discovery.go
kyriediculous: > How's that?

I forgot about pm.Sender (now ticketParamsValidator), you're right. 

> If that's really the root of the concern

It's not the root of the concern, it's a consequence of the root of the concern. How I see it `Broadcaster` and `Orchestrator` are specific instances of a general `LivepeerNode` instance that aggregates different functionalities from several packages , both with their specific exposed methods through respective interfaces. 

`DBOrchestratorPool` is not a specific instance of `LivepeerNode`, it is a service running on `LivepeerNode`. No other of these 'services'  bound to it use `LivepeerNode` in their constructor or have it as state. Only `DBOrchestratorPoolCache` does, so to me personally that isn't logical. Or is that a wrong perspective? 

There's also an idiom in go that says "pass interfaces, return structs" which I think roughly translates to the "interface segregation principle" described in [solid go design - dave cheney](https://dave.cheney.net/2016/08/20/solid-go-design)

Lastly I don't think a high-level package such as `core` should be imported in a lower level package such as `discovery`  , not having `core.LivepeerNode` as a dependency solves that.

discovery/discovery_test.go
yondonfu: Ok yeah nvm no goroutines are used here so I don't think any racy behavior should occur. 

Side note: While confirming this I noticed that the tests for the `discovery` package will fail if run consecutive times using `go test -count=100 -failfast` because the package variable `perm` is being set in the tests without being re-assigned back to the original implementation. I have a fix for this that I can push up separately though.

kyriediculous: Seems like you're right, ran into a discrepancy although not very often. I assumed the 100ms extra 'buffer' would have been sufficient to avoid this, appears not. 

Using a range in ecacd7d5 (`16<=count=<20`)

kyriediculous:
ecacd7d fixes a fragile test and removes an unused `node` field from `webhookPool` , completely removing `core` as a dependency for `discovery`

cmd/livepeer/livepeer.go
kyriediculous: ec531f3b directly stores O's fetched on-chain in the database when calling `CacheTranscoderPool`, rather than fetching their off-chain info as well.

discovery/db_discovery.go
j0sh: >  their specific exposed methods through respective interfaces.

Would you feel better if we added `Database()` , `Eth()` and `Sender()` methods to LivepeerNode (or Broadcaster) and just defined some interface for `NewDBOrchestratorPoolCache` taking all those? (Not that I actually think it's a good idea; it smells too much like Java. And being too prescriptive about design patterns is exactly why people make fun of Java.)

> No other of these 'services' bound to it use LivepeerNode in their constructor or have it as state. Only DBOrchestratorPoolCache does, so to me personally that isn't logical. Or is that a wrong perspective?

You say it's the only "service" that uses LivepeerNode. OK, so what actual issues has this led to? Why is DBOrchestratorPoolCache using LivepeerNode a problem? (Again, this is the original question.)

I'm also not sure which other services you're specifically referring to, but some of the other stuff is heavily parameterized by interfaces - almost excessively so IMO. Perhaps those could be tidied up by a good helping of LivepeerNode.

> roughly translates to the "interface segregation principle" described in solid go design - dave cheney

It is fine and good to recognize certain classes of design patterns when they're appropriate, but dogmatic adherence leads to... this.

I dislike using design patterns as a cudgel during discussion because that tends to ignore the practical realities of the system in question, so I don't bring up design patterns often, but here is one that's being broken all over the place: favor [deep abstractions over shallow ones](https://web.stanford.edu/~ouster/cgi-bin/cs190-winter18/lecture.php?topic=modularDesign) [1]. This is why I often refer to the 'surface area' of an API - and right now, things are blowing up.

Not only is the OrchestratorPool abstraction being expanded, DBOrchestratorPool has its polling/caching exposed to external callers, and the complexity of initializing everything is hoisted out to `livepeer.go`.

[1] Ousterhout's "A Philosophy of Software Design" has much more on this, and the ideas in there have found their way into a lot of the golang community's recommendations.

> Lastly I don't think a high-level package such as core should be imported in a lower level package such as discovery

Dunno, `core` gives me the impression that it's, well, core - in the middle of everything?

image

kyriediculous: > You say it's the only "service" that uses LivepeerNode. OK, so what actual issues has this led to? 

> This is why I often refer to the 'surface area' of an API - and right now, things are blowing up.

Couldn't one make a similar argument for passing LivepeerNode as a concrete type as an argument? As it is likely to expose a lot of functionality that the package in question does not need. 

> Not only is the OrchestratorPool abstraction being expanded, DBOrchestratorPool has its polling/caching exposed to external callers, and the complexity of initializing everything is hoisted out to livepeer.go.

The interface remains the same (except for the slight signature change), these methods aren't part of the interface so any package that uses `OrchestratorPool` has no knowledge of those methods. Whether we cache or not is depending on whether the last seen block is within the current round or not. But I'd be happy to learn if there is a better pattern here (boolean argument in constructor and not exporting caching/polling mehods?). 

> Dunno, core gives me the impression that it's, well, core - in the middle of everything? 

Yes I agree (altough DB is actually in the middle in your pic), but I  also see it partially different. In my opinion core should not be injected as dependencies anywhere because it is in the middle, it is core business logic. dependencies should be injected into it, otherwise there's a big chance it will lead to cyclical imports somewhere down the line. 
(diagram incomplete and not completely accurate) 
image


Anyway design patterns aside I'm wondering where we can find some common ground in this discussion.  Would you be more okay with the following? 

```
type DBOrchestratorPoolCache struct {
	store                 common.OrchestratorStore
	lpEth                 eth.LivepeerEthClient
	ticketParamsValidator ticketParamsValidator
	rm                    roundsManager
	bcast common.Broadcaster
}

func NewDBOrchestratorPoolCache(n *core.LivepeerNode, rm roundsManager) *DBOrchestratorPoolCache {
	bcast := core.NewBroadcaster(n)
	return &DBOrchestratorPoolCache{
		store: n.Database,
		lpEth: n.Eth,
		ticketParamsValidator: n.Sender,
		rm: rm,
		bcast: bcast,
	}
}

type OrchestratorPool interface {
	GetURLs() []*url.URL
	GetOrchestrators(int) ([]*net.OrchestratorInfo, error)
	Size() int
}
```

discovery/db_discovery.go
kyriediculous: > The only time O selection blocks a pending segment is for the very first request establishing a session with the HTTP ingest 

Ah thanks , I did not realise this. In that case the current approach is as you say a good compromise indeed.

discovery/db_discovery.go
j0sh: > Couldn't one make a similar argument for passing LivepeerNode as a concrete type as an argument? As it is likely to expose a lot of functionality that the package in question does not need.

Having the functionality generally available but only used by other packages is not really a problem. What's problematic is *how* the exposed functionality is expected to be used. As-is right now, too much is being expected of API callers in terms of set-up logic, when most of that logic could be pulled into the DB cache itself. APIs should be misuse-resistant [1] to minimize the risk of "user error" but the current changes substantially increase that risk.

[1] Section 2.3 of the [API Design Manual](https://people.mpi-inf.mpg.de/~jblanche/api-design.pdf), "Hard to misuse"

> The interface remains the same (except for the slight signature change)

The signature change to OrchestratorPool is *significant*, not slight. You're now making it a requirement that the node / broadcaster is available at all call sites of `GetOrchestrator`. Currently that's not really a problem but it may not always be. That is an avoidable issue and the change is unnecessary.

Then there's the change to the usage of `DBOrchestratorPoolCache`, where the caller (`livepeer.go`) needs to do a lot of manual set-up, *and* the particulars of the caching/polling setup are such that it's actually sensitive to initialization order [1]. Both issues are also avoidable, because none of that actually needs to take place outside the `NewDBOrchestratorPoolCache`. And the new functions require essentially the same parameters - the `Broadcaster` - which is yet another indication things should be reworked.

[1] https://github.com/livepeer/go-livepeer/pull/1182#discussion_r349409217
[2] https://people.mpi-inf.mpg.de/~jblanche/api-design.pdf

> Would you be more okay with the following?

My suggestion is to keep overall API changes as small as possible. From what I can tell, the only change that may be required for the functional purpose of this PR is the inclusion of the rounds manager to the `NewDBOrchestratorPoolCache`. So that means:

* Revert the change to OrchestratorPool.GetOrchestrators
* Add the `roundsManager` to `NewDBOrchestratorPoolCache`
* Initialize the cache / poller *within* the `NewDBOrchestratorPoolCache` . This doesn't need to be done within `livepeer.go`.

It's a bit of a sport to keep diffs as minimal as possible while achieving the goals of the task. How small can you go?

yondonfu: IIUC, the following points are being debated:

1. The modification of the public API of the OrchestratorPool interface (change to the signature of GetOrchestrators())
2. The larger API surface area of DBOrchestratorPoolCache because the caller is now responsible for transcoder pool caching and polling via the CacheTranscoderPool() and PollOrchestratorInfo() methods respectively
3. The injection of 3 interface dependencies into DBOrchestratorPoolCache instead of injecting a single concrete LivepeerNode dependency

My take is:

1. +1 for not changing the OrchestratorPool public API since updating the public API by passing an additional arg in the method has a clear cost due to the requirements for call site changes. Furthermore, the caller now needs to be aware of this dependency which might be a worse state of affairs than DBOrchestratorPoolCache being aware of the dependency - we go from a single concrete type being aware of a dependency to every single type that implements the interface being aware of a depdency

2. +1 that the caller shouldn't be responsible for setting up caching and polling logic. We could either move this into the constructor or into a dedicated method i.e. `Start()` that accepts a cancellable context (that can be used to exit the polling goroutine gracefully)

3. I think Josh makes a good point about focusing on how functionality is used instead of just exposed functionality. I also understand the desire to abstract away the implementation details of dependencies by *only* using interfaces. I think avoiding introducing additional package dependencies when we can is a good thing (in this case the `core` package). However, in this case, I think reaching for these abstractions might be premature and we should try to avoid reaching for abstractions until we actually need them rather than trying to predict their need. There is a saying: [duplication is cheaper than the wrong abstraction](https://www.sandimetz.com/blog/2016/1/20/the-wrong-abstraction) - in this case I think accepting the current state of dependencies is cheaper than reaching for what *could* be the wrong abstraction in the absence of more information. For that reason, I'm also +1 on this point

kyriediculous: Let's get this PR merged first a foremost and we can always discuss this later if we want. I'd just want to close off by saying that I can agree with 1 & 2 but not in full with point 3. I listed my objections below just for completeness. But let's get 6061820 reviewed and I hope everyone can find themselves happy with the changes in that commit. 

- Both types are now dependant of eachother, being one step away from cyclical imports
- `LivepeerNode` or most of its functionality is not needed here at all, simply saying "it's there, just don't use it"  doesn't feel right either. 
- Importing an additional package just because it adds a few constructor parameters (what's wrong with composition?) is in my opinion the wrong trade-off to make
- I agree that we shouldn't prematurely reach for abstractions but on the other hand whether an abstraction is wrong can only be known in hindsight, so the statement made by the author there is actually a perfect example of a petitio principii (one can equally say "the right abstraction is cheaper than duplication").

discovery/db_discovery.go
```diff
@@ -24,33 +25,53 @@ var getTicker = func() *time.Ticker {
...
-	node *core.LivepeerNode
+	store                 common.OrchestratorStore
+	lpEth                 eth.LivepeerEthClient
+	ticketParamsValidator ticketParamsValidator
+	rm                    roundsManager
+	bcast                 common.Broadcaster
 }
 
-func NewDBOrchestratorPoolCache(node *core.LivepeerNode) *DBOrchestratorPoolCache {
+func NewDBOrchestratorPoolCache(node *core.LivepeerNode, rm roundsManager, ctx context.Context) (*DBOrchestratorPoolCache, error) {
```
yondonfu: Let's make `ctx context.Context` the first parameter as that is convention (we don't use them right now, but static analysis tools that check for context propagation expect this convention - more details in the [docs](https://golang.org/pkg/context/)).

discovery/db_discovery.go
```diff
@@ -193,10 +236,15 @@ func ethOrchToDBOrch(orch *lpTypes.Transcoder) *common.DBOrch {
 	if orch == nil {
 		return nil
 	}
-	return &common.DBOrch{
-		ServiceURI:   orch.ServiceURI,
-		EthereumAddr: orch.Address.String(),
+
+	dbO := &common.DBOrch{
```
yondonfu: Can we just return the struct literal here?

discovery/db_discovery.go
```diff
@@ -100,31 +121,58 @@ func (dbo *DBOrchestratorPoolCache) GetOrchestrators(numOrchestrators int) ([]*n
...
 	return nil
 }
 
-func cacheDBOrchs(node *core.LivepeerNode, orchs []*lpTypes.Transcoder) ([]*common.DBOrch, error) {
-	var dbOrchs []*common.DBOrch
-	if orchs == nil {
-		glog.Error("No new DB orchestrators created: no orchestrators found onchain")
-		return dbOrchs, nil
-	}
+func (dbo *DBOrchestratorPoolCache) pollOrchestratorInfo(ctx context.Context) error {
```
yondonfu, line 148: The ticker in this loop will not tick until after `cacheRefreshInterval`. While `cacheTranscoderPool()` will fetch off-chain info from any Os in the on-chain pool, it will not fetch off-chain info from any Os that are active in the current round, but that are no longer in the on-chain pool. I recommend the following:

1. `cacheTranscoderPool()` just updates the DB with the on-chain pool and does no off-chain info fetching
2. `pollOrchestratorInfo()` fetches off-chain info of all active Os when it is called *before* it starts the polling loop so that we can make sure that we fetch the off-chain info of all active Os when `DBOrchestratorPoolCache` is created

cmd/livepeer/livepeer.go
```diff
@@ -287,6 +287,9 @@ func main() {
 	}
 
 	watcherErr := make(chan error)
+	var roundsWatcher *watchers.RoundsWatcher
+	var originalLastSeenBlock *big.Int
```
yondonfu: Besides `roundsWatcher`, doesn't look like we need to hoist any of these variables up here anymore.

yondonfu (changes requested):

discovery/db_discovery.go
yondonfu: > so the statement made by the author there is actually a perfect example of a petitio principii (one can equally say "the right abstraction is cheaper than duplication").

I think the author's point is being misinterpreted a bit here...Of course we only know if an abstraction is wrong in hindsight. The point is that *if* the abstraction turns out to be wrong, then dealing with the consequences of that abstraction is almost always more of a headache than dealing with duplication. And "duplication" here can probably be replaced by a variety of not-so-ideal pieces of code including code that exposes more functionality than needed.

discovery/db_discovery.go
kyriediculous: Ah right I didn't consider the O's that are added through backfilling. 

I would prefer option 1 , but would you be okay with duplicating `parseURI` logic here then to prevent inserting O's with a bad URL? We don't actually need the parsed URI, but we need to check whether the `ServiceURI` fetched on-chain is a valid URL. If `parseURI` returns an error we can skip insertion into the DB for that O.

discovery/db_discovery.go
kyriediculous: fixed in d9c1a2d9

discovery/db_discovery.go
kyriediculous: fixed in d9c1a2d9

cmd/livepeer/livepeer.go
kyriediculous: fixed in d9c1a2d9

discovery/db_discovery.go
yondonfu: We would want to do both 1 and 2 - 1 makes sure that our DB has an up-to-date view of the active set and 2 makes sure that given an up-to-date view of the active set that we immediately fetch off-chain info for each active O when we start up.

> but would you be okay with duplicating parseURI logic here then to prevent inserting O's with a bad URL?

Calling `parseURI()` in `cacheTranscoderPool()` should work. Note that the `ServiceRegistryWatcher` does not check the URI before updating the DB so that should probably be updated external to this PR.

kyriediculous: `OrchestratorWatcher` does not update the URI so it just struck me that after abolishing backfilling from a very early block we never save the URI for orchestrators that are inserted through backfilling rather than `CacheTranscoderPool`, so those O's will never have their URI field populated and thus never be selected. When backfilling from a block far in the past (eg protocol genesis) this would have been done by `ServiceURIWatcher`.

We could make an RPC call when a transcoder becomes active to fetch its `ServiceURI` , validate it and insert it if it is valid.

yondonfu: I actually meant `ServiceRegistryWatcher` in the previous post (edited to reflect that), but yeah we should validate & insert the service URI in `OrchestratorWatcher` as well.

j0sh: > We could make an RPC call when a transcoder becomes active to fetch its ServiceURI

Is this the on-chain ServiceURI or the one received from `OrchestratorInfo`[1]? If the latter, is there a possibility of a thundering herd problem? Transcoder becomes active and immediately gets hammered by Bs all trying to fetch its ServiceURI?

[1] I think we should only be caching the on-chain ServiceURI anyway?

kyriediculous: Actually after second thought I'm wondering if we should validate the URI at all (except when fetching off-chain info and building the working set) 

An O with a bad URI will not be selected anyway and we won't try to fetch its `OrchestratorInfo`. The part of the code that does run into trouble here however are the unit tests, because the overwritten `GetOrchestratorInfo` methods do not actually validate the URI, but the "live" method does. Perhaps we should instead add URI validation to the unit tests.

kyriediculous: @j0sh fetch the URI on-chain from `ServiceRegistry` , should have been clearer here, my bad! 

Created an issue to track this: https://github.com/livepeer/go-livepeer/issues/1217

kyriediculous: I'm wondering if we can't simply 'validate' a url string in the sql query to begin with?
Can we use a RegEx here to prevent selecting bad urls from the DB in the first place? 

https://www.sqlite.org/lang_expr.html#regexp

So to summarize my comments

1. Don't validate URI upon insertion
2. Pre-select for bad URL's in SQL query through a regex
3. Use `ParseURI` as an extra safety in `getOrchInfo` to prevent inserting O's for which we can't fetch off-chain data in the working set.

yondonfu: For the purposes of this PR, I think we can just skip URI validation in `cacheTranscoderPool()`. Filtering out bad URIs via a SQL query might be a nice optimization, but I think the `parseURI()` check in `getOrchInfo()` gets the job done for now.

kyriediculous: fixed in d358b9f1

discovery/db_discovery.go
```diff
@@ -100,29 +121,60 @@ func (dbo *DBOrchestratorPoolCache) GetOrchestrators(numOrchestrators int) ([]*n
...
-		return err
+		return fmt.Errorf("Could not refresh DB list of orchestrators: %v", err)
 	}
-	_, dbOrchErr := cacheDBOrchs(node, orchestrators)
-	if dbOrchErr != nil {
-		glog.Error("Could not refresh DB list of orchestrators: cacheDBOrchs err")
+
+	for _, o := range orchestrators {
+		if err := dbo.store.UpdateOrch(ethOrchToDBOrch(o)); err != nil {
+			glog.Errorf("Unable to update orchestrator %v in DB: %v", o.ServiceURI, err)
```
yondonfu: Since an orchestrator is identified by its ETH address in the DB can we use the ETH address instead of the ServiceURI here?

discovery/discovery_test.go
```diff
@@ -172,67 +129,61 @@ func TestPoolSize(t *testing.T) {
...
+	defer cancel()
 
-	addresses := []string{"https://127.0.0.1:8936", "https://127.0.0.1:8937", "https://127.0.0.1:8938"}
-	orchestrators := StubOrchestrators(addresses)
-	node.Eth = &eth.StubClient{Orchestrators: orchestrators}
-	nonEmptyPool := NewDBOrchestratorPoolCache(node)
-	require.NotNil(nonEmptyPool)
-	assert.Equal(t, len(addresses), emptyPool.Size())
-}
+	sender.On("ValidateTicketParams", mock.Anything).Return(nil).Times(3)
```
yondonfu: Why do we need this here? I don't see a call to `GetOrchestrators()`

discovery/discovery_test.go
```diff
@@ -299,31 +273,31 @@ func TestNewDBOrchestratorPoolCache_TestURLs(t *testing.T) {
...
+			Orchestrators: orchestrators,
+		},
+		Sender: sender,
+	}
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
 
-	node, _ := core.NewLivepeerNode(nil, "", nil)
-	node.Database = dbh
+	sender.On("ValidateTicketParams", mock.Anything).Return(nil).Times(2)
```
yondonfu: Why do we need this here? I don't see a call to `GetOrchestrators()`

discovery/discovery_test.go
```diff
@@ -335,26 +309,126 @@ func TestNewDBOrchestratorPoolCache_TestURLs_Empty(t *testing.T) {
...
+			Orchestrators: orchestrators,
+		},
+		Sender: sender,
+	}
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	sender.On("ValidateTicketParams", mock.Anything).Return(nil)
+
+	cacheRefreshInterval = 200 * time.Millisecond
```
yondonfu, line 396: Let's make sure to reset this to the original value after the test finishes in case any other tests depend on the variable value.

Ex.

```
origCacheRefreshInterval = cacheRefreshInterval
cacheRefreshInterval = 200 * time.Millisecond
defer func() { cacheRefreshInterval = origCacheRefreshInterval }()
```

discovery/discovery_test.go
```diff
@@ -335,26 +309,126 @@ func TestNewDBOrchestratorPoolCache_TestURLs_Empty(t *testing.T) {
...
+		Database: dbh,
+		Eth: &eth.StubClient{
+			Orchestrators: orchestrators,
+		},
+		Sender: sender,
+	}
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	sender.On("ValidateTicketParams", mock.Anything).Return(nil)
```
yondonfu: Why do we need this here? I don't see a call to `GetOrchestrators()`

discovery/discovery_test.go
kyriediculous: me being overzealous late at night probably 😄

But indeed, no O's means we never iterate through the for loop that makes that call.

discovery/discovery_test.go
kyriediculous: We do but we overwrite that default function with one that doesn't care about checking ticketparams so we don't need it here indeed.

discovery/discovery_test.go
kyriediculous: fixed in e8acbd03

discovery/discovery_test.go
kyriediculous: fixed in e8acbd03

discovery/db_discovery.go
kyriediculous: fixed in e8acbd03

discovery/discovery_test.go
kyriediculous: fixed in e8acbd03

discovery/discovery_test.go
kyriediculous: fixed in e8acbd03

yondonfu (commented):
Let's rebase

kyriediculous:
rebased !

yondonfu:
Can we reword the messages for commits 1647c20 and 406e27a? They aren't accurate anymore.

yondonfu (approved):
LGTM but let's reword the commit messages I mentioned in the above comment before merging

kyriediculous:
Done, but despite approval merging is blocked for some reason (all conversations are resolved). 

Also one of the bash tests is failing on `E1128 13:06:44.478654    8722 livepeer.go:454] Failed to backfill events: not found` , this doesn't happen locally for me when running the node. Only when running the bash tests.

```
nico@Nicos-MacBook-Pro go-livepeer (nv/poolcache-update) $ ./livepeer -broadcaster -maxTicketEV abcd -network rinkeby
I1128 14:17:26.584366   22904 livepeer.go:199] ***Livepeer is running on the rinkeby network: 0xA268AEa9D048F8d3A592dD7f1821297972D4C8Ea***
I1128 14:17:27.414775   22904 accountmanager.go:70] Using Ethereum account: 0x13d3C3bdDD3531d11F92884bC49cb98255D51370
I1128 14:17:28.077447   22904 accountmanager.go:88] Please enter the passphrase to unlock Ethereum account 0x13d3C3bdDD3531d11F92884bC49cb98255D51370
Passphrase:
I1128 14:17:30.699888   22904 accountmanager.go:99] Unlocked ETH account: 0x13d3C3bdDD3531d11F92884bC49cb98255D51370
I1128 14:17:31.869219   22904 block_watcher.go:318] Backfilling block events (this can take a while)...
I1128 14:17:31.869254   22904 block_watcher.go:319] Start block: 5520513		 End block: 5520524		Blocks elapsed: 11
I1128 14:17:31.873978   22904 block_watcher.go:537] fetching block logs from=5520513 to=5520524
panic: -maxTicketEV must be a valid rational number, but abcd provided. Restart the node with a valid value for -maxTicketEV

goroutine 1 [running]:
main.main()
	/Users/nico/Desktop/livepeer/go-livepeer/cmd/livepeer/livepeer.go:570 +0x74f2
```

Edit: now they pass , results are inconsistent. At first it seemed a bug when there had only been 1 block elapsed, but this isn't the case either: 

```
+ ./livepeer -broadcaster -depositMultiplier 0 -network rinkeby -ethKeystorePath /tmp/livepeer-test-29155/key
I1128 14:30:12.409627   28672 livepeer.go:199] ***Livepeer is running on the rinkeby network: 0xA268AEa9D048F8d3A592dD7f1821297972D4C8Ea***
I1128 14:30:13.030049   28672 accountmanager.go:70] Using Ethereum account: 0x089d8aB6752BaC616a1F17246294eb068EE23d3E
I1128 14:30:13.687394   28672 accountmanager.go:99] Unlocked ETH account: 0x089d8aB6752BaC616a1F17246294eb068EE23d3E
I1128 14:30:14.778222   28672 block_watcher.go:318] Backfilling block events (this can take a while)...
I1128 14:30:14.778259   28672 block_watcher.go:319] Start block: 5520574		 End block: 5520575		Blocks elapsed: 1
I1128 14:30:14.781452   28672 block_watcher.go:537] fetching block logs from=5520574 to=5520575
panic: -depositMultiplier must be greater than 0, but 0 provided. Restart the node with a valid value for -depositMultiplier
```

yondonfu:
> Done, but despite approval merging is blocked for some reason (all conversations are resolved).

Since @j0sh requested changes, he will need to give an approval.

> now they pass , results are inconsistent

I believe the inconsistent results are caused by RPC calls ([example](https://github.com/livepeer/go-livepeer/blob/master/eth/blockwatch/block_watcher.go#L337)) in `getMissedEventsToBackfill()` that sometimes return a `not found` error. I suspect this error is returned sometimes when requests are load balanced across a pool of ETH nodes (i.e. when using a service like Infura) and one of the nodes in the pool is slightly behind. The polling code in `BlockWatcher` handles this possibility by [ignoring the error and retrying after the next interval](https://github.com/livepeer/go-livepeer/blob/master/eth/blockwatch/block_watcher.go#L229).

For the purposes of fixing the inconsistent test behavior in this PR, we can move the calls to `blockWatcher.BackfillEventsIfNeeded()` and `blockWatcher.Watch()` after the rest of the setup code for broadcasters and orchestrators in on-chain mode (which is the code that the `test_args.sh` script is checking):

```diff
diff --git a/cmd/livepeer/livepeer.go b/cmd/livepeer/livepeer.go
index fc460461..77a7faa2 100644
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -445,33 +445,6 @@ func main() {
 		go serviceRegistryWatcher.Watch()
 		defer serviceRegistryWatcher.Stop()
 
-		blockWatchCtx, cancel := context.WithCancel(context.Background())
-		defer cancel()
-
-		// Backfill events that the node has missed since its last seen block. This method will block
-		// and the node will not continue setup until it finishes
-		if err := blockWatcher.BackfillEventsIfNeeded(blockWatchCtx); err != nil {
-			glog.Errorf("Failed to backfill events: %v", err)
-			return
-		}
-
-		blockWatcherErr := make(chan error, 1)
-		go func() {
-			if err := blockWatcher.Watch(blockWatchCtx); err != nil {
-				blockWatcherErr <- fmt.Errorf("block watcher error: %v", err)
-			}
-		}()
-
-		go func() {
-			var err error
-			select {
-			case err = <-roundsWatcherErr:
-			case err = <-blockWatcherErr:
-			}
-
-			watcherErr <- err
-		}()
-
 		n.Balances = core.NewAddressBalances(cleanupInterval)
 		defer n.Balances.StopCleanup()
 
@@ -591,6 +564,33 @@ func main() {
 				glog.Infoln("To update the broadcaster's maximum acceptable transcoding price per pixel, use the CLI or restart the broadcaster with the appropriate 'maxPricePerUnit' and 'pixelsPerUnit' values")
 			}
 		}
+
+		blockWatchCtx, cancel := context.WithCancel(context.Background())
+		defer cancel()
+
+		// Backfill events that the node has missed since its last seen block. This method will block
+		// and the node will not continue setup until it finishes
+		if err := blockWatcher.BackfillEventsIfNeeded(blockWatchCtx); err != nil {
+			glog.Errorf("Failed to backfill events: %v", err)
+			return
+		}
+
+		blockWatcherErr := make(chan error, 1)
+		go func() {
+			if err := blockWatcher.Watch(blockWatchCtx); err != nil {
+				blockWatcherErr <- fmt.Errorf("block watcher error: %v", err)
+			}
+		}()
+
+		go func() {
+			var err error
+			select {
+			case err = <-roundsWatcherErr:
+			case err = <-blockWatcherErr:
+			}
+
+			watcherErr <- err
+		}()
 	}
 
 	if *s3bucket != "" && *s3creds == "" || *s3bucket == "" && *s3creds != "" {
```

For the time being, anyone that encounters this error when starting the node can just restart the node. Outside of this PR we can consider adding retries for RPC calls that return `not found` in `getMissedEventsToBackfill()`.

kyriediculous:
Thanks for the quick reply! moved the logic to end of on-chain setup in 7d0d16f3

common/types.go
```diff
@@ -0,0 +1,25 @@
...
+	ethcommon "github.com/ethereum/go-ethereum/common"
+	"github.com/livepeer/go-livepeer/net"
+)
+
+type Broadcaster interface {
+	Address() ethcommon.Address
+	Sign([]byte) ([]byte, error)
+}
+
+type OrchestratorPool interface {
```
j0sh, line 15: It *feels* like we had earlier problems with the OrchestratorPool in `common`  - hence why it was in `net` - but if it's not leading to any issues right now, then this is probably a better place for it.

j0sh (approved):

wohlner:

PR #1199, opened 2019-11-13
Broadcaster encoding fee estimation

### LINKED ISSUE #1174: Balance synchronization problem when timing out for slow orchestrators
**Describe the bug**

The default HTTP timeout is currently 8 seconds for requests sent by a broadcaster. Thus, if an orchestrator takes longer than 8 seconds to return a transcoded result, the HTTP timeout will be triggered and the broadcaster will retry the segment with a different orchestrator. However, the orchestrator is not aware of the HTTP timeout so it will finish transcoding the segment, debit the transcoding fees for the segment from the broadcaster's balance for the stream and then write the data to the still active connection with the broadcaster. This will result in the orchestrator having a different view of the stream balance than the broadcaster.

**To Reproduce**

This is a problem in the following scenario:

Let `balance_b` be the broadcaster's view of the stream balance.
Let `balance_o` be the orchestrator's view of the stream balance.
Let X be the minimum credit requirement for the orchestrator - it will not transcode a segment unless the stream balance is at least X.

1. Broadcaster submits segment 0 with a payment worth X to the orchestrator. At this point, `balance_b = X`
2. After the orchestrator receives the segment and payment, `balance_o = X`
3. The HTTP timeout is triggered because the broadcaster did not receive a response from the orchestrator within 8 seconds
4. At this point, `balance_b = 0` because the broadcaster considers any existing credit to have been spent when it does not receive a result from the orchestrator
5. Orchestrator finishes transcoding segment 0 and debits the transcoding fees `Y` where `Y > X`. Now, `balance_o < 0`
6. Broadcaster is only aware of this orchestrator so it retries segment 0 with the same orchestrator. It sends segment 0 with a payment worth X to the orchestrator. At this point `balance_b = X`
7. After the orchestrator receives the segment and payment, `0 < balance_o < X`. Since `balance_o < X`, the orchestrator returns an insufficient balance error

**Expected behavior**

The broadcaster should not encounter an insufficient balance error in the above scenario

**Screenshots**

Orchestrator logs:
Screen Shot 2019-11-06 at 11 32 11 AM

Broadcaster logs:
Screen Shot 2019-11-06 at 11 32 23 AM

 - OS: Linux

yondonfu:

This PR implements encoding fee estimation for the broadcaster.

Prior to generating a balance update (which is used to determine the # of tickets to include in a payment), the broadcaster will estimate the transcoding fee for the source segment (at the moment, the broadcaster only estimates the encoding fee, but does not estimate the decoding fee). If the estimated fee is greater than the ticket EV, the broadcaster will use the estimated fee as the "safe minimum credit". If the estimated fee is less than the ticket EV, the broadcaster will use the ticket EV as the "safe minimum credit". We calculate the safe minimum credit in this manner because the orchestrator still requires a minimum credit (EV of a single ticket) - we can remove this safe minimum credit calculation if we implement short circuited transcoding in the future which would allow us to remove minimum credit requirements.

- 7495ea6: Add the `estimateFee()` helper
- 1171e35: Reuse a helper for converting `*net.PriceInfo` into `*big.Rat` 
- 5e31f8d: Add safe minimum credit calculation

Added unit tests. 

Replicated the scenario described in #1174 by streaming into the broadcaster, immediately shutting it down after submitting the first segment and before it receives a response and then reconnecting to the same orchestrator after rebooting the broadcaster.

Fixes #1174

**Checklist:**

### DIFF
--- a/server/segment_rpc.go
+++ b/server/segment_rpc.go
@@ -6,6 +6,7 @@ import (
 	"encoding/base64"
 	"fmt"
 	"io/ioutil"
+	"math"
 	"math/big"
 	"net/http"
 	"strings"
@@ -16,6 +17,7 @@ import (
 	"github.com/livepeer/go-livepeer/drivers"
 	"github.com/livepeer/go-livepeer/monitor"
 	"github.com/livepeer/go-livepeer/net"
+	"github.com/livepeer/lpms/ffmpeg"
 	"github.com/livepeer/lpms/stream"
 	"golang.org/x/net/http2"
 
@@ -29,6 +31,8 @@ import (
 
 const paymentHeader = "Livepeer-Payment"
 const segmentHeader = "Livepeer-Segment"
+
+const pixelEstimateMultiplier = 1.02
 
 var errSegEncoding = errors.New("ErrorSegEncoding")
 var errSegSig = errors.New("ErrSegSig")
@@ -266,8 +270,18 @@ func SubmitSegment(sess *BroadcastSession, seg *stream.HLSSegment, nonce uint64)
 		data = []byte(seg.Name)
 	}
 
+	priceInfo, err := ratPriceInfo(sess.OrchestratorInfo.GetPriceInfo())
+	if err != nil {
+		return nil, err
+	}
+
+	fee, err := estimateFee(seg, sess.Profiles, priceInfo)
+	if err != nil {
+		return nil, err
+	}
+
 	// Create a BalanceUpdate to be completed when this function returns
-	balUpdate, err := newBalanceUpdate(sess)
+	balUpdate, err := newBalanceUpdate(sess, fee)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


server/segment_rpc.go
```diff
@@ -484,15 +497,42 @@ func genSegCreds(sess *BroadcastSession, seg *stream.HLSSegment) (string, error)
...
+			return nil, err
+		}
+
+		// Take the ceiling of the duration to always overestimate
+		outPixels += int64(w*h) * int64(p.Framerate) * int64(math.Ceil(seg.Duration))
+	}
+
+	// feeEstimate = pixels * pixelEstimateMultiplier * priceInfo
+	fee := new(big.Rat).SetInt64(outPixels)
+	// Multiply pixels by pixelEstimateMultiplier to ensure that we never underpay
```
kyriediculous, line 520: Since we overestimate the number output pixels by rounding up the segment duration do we still need `pixelEstimateMultiplier` wouldn't either suffice?

yondonfu: I added `pixelEstimateMultiplier` just to be safe in the scenario where the segment duration is already an integer value (i.e. 2.0). I believe it is possible for the # of frames in the output to not always exactly equal FPS * duration - see [this example](https://github.com/livepeer/lpms/blob/master/ffmpeg/ffmpeg_test.go#L520) where we expect 125 frames for a 1 second segment at 123 FPS

kyriediculous (approved):
This looks good to me 👍

PR #1201, opened 2019-11-15
Small fixes (ticket validation, default flag values, devtool updates, etc.)

### LINKED ISSUE #1176: Improve logging on "Could not create payment: ticket EV higher than max EV"

I'm trying to set up the parameters for PM, and I'm hitting an error `Could not create payment: ticket EV higher than max EV`.

I am aware of [the notes in the documentation about how to resolve this issue](https://livepeer.readthedocs.io/en/latest/streamflow-public-testnet.html#broadcasting-video), but it would be really helpful if the error logging could be more informative - in order to help expose some of the inner workings of the PM setup.

It would really help to log the values of the `ticket EV` and `max EV` that are being compared, so that a user can understand how far off they are.

`Could not create payment: ticket EV = 10000000001, higher than max EV = 10000000000`.

In addition, in general, it would be really useful to expose the parameters being used to "govern" the PM parameters, to help users to know what's going on "under the hood", and thus allow them to understand via their own worked examples.

The `gaspricemonitor.go:134] Cached gas price: 1000000000` is a good start, but I feel like there are other metrics which would be worth logging, to help a user work out what is going on in real time.

I've tried to think about alternatives to this - there is perhaps an option to display a section in the `livepeer_cli` to display some kind of `PAYMENT STATS`, displaying some real time metrics about how the nodes are operating in terms of cached gas price, max ticket EV, ticket EV - although some of those things might be per-stream.

**Additional context**

None, thanks! Keep up the good work... this PM stuff is really interesting, but would help to have some more visibility on how it works in practice, and in real time :)

### LINKED ISSUE #1177: Improve logging on "Could not create payment: ticket faceValue higher than max faceValue"

I'm trying to set up the parameters for PM, and I'm hitting an error `Could not create payment: ticket faceValue higher than max faceValue`.

I am aware of [the notes in the documentation about how to resolve this issue](https://livepeer.readthedocs.io/en/latest/streamflow-public-testnet.html#broadcasting-video), but it would be really helpful if the error logging could be more informative - in order to help expose some of the inner workings of the PM setup.

This issue is related to https://github.com/livepeer/go-livepeer/issues/1176

It would really help to log the values of the `ticket faceValue` and `max faceValue` that are being compared, so that a user can understand how far off they are.

`Could not create payment: ticket faceValue = 123456789 higher than max faceValue = 123456788`.

In addition, in general, it would be really useful to expose the parameters being used to "govern" the PM parameters, to help users to know what's going on "under the hood", and thus allow them to understand via their own worked examples.

Showing some of the static node-specific PM parameters in the `livepeer_cli` could be an effective way of shedding some light on what's going on inside these nodes.

For example, a section which shows the `max faceValue` and other metrics used to calculate the ticket parameters, and to govern what is and is not allowed.

**Additional context**

None.

### LINKED ISSUE #1179: Compare max faceValue against the faceValue of a single ticket in a batch

When a broadcaster sends creates an N ticket batch for a payment, the broadcaster compares its max faceValue (deposit / depositMultiplier) against N * faceValue. If N * faceValue > max faceValue, the broadcaster will not send the payment. This validation presents the following problem: when the ticket EV is much lower than the transcoding fee for a segment, the broadcaster's debt after the first segment will be quite high. As a result, N will be large. Since faceValue will generally be high in order to minimize the tx cost overhead, N * faceValue will be quite large and can easily exceed the max face value for a broadcaster.

Furthermore, given that the broadcaster can already control the maximum payment value it sends out, there isn't that much benefit in controlling the maximum sum of the faceValues of tickets in a batch since the maximum payment value already bounds the risk of the broadcaster sending too large of a payment for a single segment. 

The broadcaster should compare its max faceValue against the faceValue of a single ticket in a batch.

yondonfu:

This PR implements a few small fixes - each fix is contained within a self-contained commit, but I chose to group these commits together.

- cdbc512: Compare max faceValue against the faceValue of a single ticket when validating ticket params even if the broadcaster is sending multiple tickets
- d2e8a44: Include values when logging errors for failed max faceValue and max ticket EV validation to give users a better sense of what is going wrong when these errors occur
- 2fb674c: Log the actual error encountered by `blockwatch.Watcher`. At the moment, we just log that *something* went wrong. This update will help diagnose the details of went wrong
- 17b1798: Do not return an error when `watchers.UnbondingWatcher` tries to process an event that it is not tracking. Since we are now watching for `TranscoderActivated` and `TranscoderDeactivated` events from the BondingManager, it is possible for `watchers.UnbondingWatcher` to receive one of these events which it can just ignore
- e30bb79: Update the defaults for `-ticketEV`, `-maxTicketEV` and `-depositMultiplier` to more reasonable values that make it easier for a user to get started
- fb5fe0a: Log # of responses during discovery instead of an array of all protobuf messages received. This will reduce log clutter - most of the protobuf message is not easily decipherable
- 8557724: Update the devtool script to use HTTP urls for the ETH RPC provider and to set `-blockPollingInterval` to 1 second since the dev chain to be used with the script has 1 second block times

Updated unit tests and ran the devtool script.

Fixes #1179 
Fixes #1176 
Fixes #1177

**Checklist:**

### DIFF
--- a/cmd/devtool/devtool.go
+++ b/cmd/devtool/devtool.go
@@ -30,7 +30,7 @@ const (
 
 var (
 	ethTxTimeout              = 600 * time.Second
-	endpoint                  = "ws://localhost:8546/"
+	endpoint                  = "http://localhost:8545/"
 	gethMiningAccount         = "87da6a8c6e9eff15d703fc2773e32f6af8dbe301"
 	gethMiningAccountOverride = false
 	ethController             = "0x04B9De88c81cda06165CF65a908e5f1EFBB9493B"
@@ -306,6 +306,7 @@ func createRunScript(ethAcctAddr, dataDir, serviceHost string, isBroadcaster boo
     -ethUrl %s \
     -ethPassword "" \
     -network=devenv \
+    -blockPollingInterval 1 \
     -monitor=false -currentManifest=true -cliAddr %s:%d -httpAddr %s:%d `,
 		ethController, dataDir, ethAcctAddr, endpoint, serviceHost, cliPort, serviceHost, mediaPort)
 
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -107,11 +107,11 @@ func main() {
 	gasLimit := flag.Int("gasLimit", 0, "Gas limit for ETH transactions")
 	gasPrice := flag.Int("gasPrice", 0, "Gas price for ETH transactions")
 	initializeRound := flag.Bool("initializeRound", false, "Set to true if running as a transcoder and the node should automatically initialize new rounds")
-	ticketEV := flag.String("ticketEV", "1000000000", "The expected value for PM tickets")
+	ticketEV := flag.String("ticketEV", "1000000000000", "The expected value for PM tickets")
 	// Broadcaster max acceptable ticket EV
-	maxTicketEV := flag.String("maxTicketEV", "10000000000", "The maximum acceptable expected value for PM tickets")
+	maxTicketEV := flag.String("maxTicketEV", "100000000000000", "The maximum acceptable expected value for PM tickets")
 	// Broadcaster deposit multiplier to determine max acceptable ticket faceValue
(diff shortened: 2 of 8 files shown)


kyriediculous (approved):
LGTM 🕵

PR #1206, opened 2019-11-20
support s3-compatible object stores

iameli:

I hacked this together so I could play around with using [Minio](https://min.io/) as an object store, executing go-livepeer like so:

```
./livepeer -s3bucket http://localhost:9000/eli-bucket -s3creds minioadmin/minioadmin -broadcaster -transcodingOptions P240p30fps16x9
```

- Changed some of the logic on the `-s3bucket` parameter to allow it to support both the old `-s3bucket regionname/bucketname` format as well as a fully-specified `-s3bucket https://minio.example.com/bucketname`. 
- Added `--enable-protocol=http` to the ffmpeg build so that the HLS puller can request segments from a non-https store, e.g. for a local Minio server in a secure cluster.

* Tried four test cases with a local Minio server: Broadcaster OS, Orchestrator OS, Both OS, neither OS. I tried to repeat all four cases with an Amazon S3 bucket but got signature errors. Then I tried the same methodology with a build off of master and still got signature errors. Is there a current S3 bug, perhaps?

### DIFF
--- a/drivers/s3.go
+++ b/drivers/s3.go
@@ -10,6 +10,7 @@ import (
 	"io"
 	"mime/multipart"
 	"net/http"
+	"net/url"
 	"path"
 	"strings"
 	"time"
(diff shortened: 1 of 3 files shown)


drivers/s3.go
```diff
@@ -51,16 +52,31 @@ type s3Session struct {
...
+// Returns host, region, bucket
+func s3Host(s3config string) (string, string, string) {
+	u, err := url.Parse(s3config)
+	// This means it's e.g. "us-west-2/my-bucket"
+	if err != nil || u.Scheme == "" {
+		br := strings.Split(s3config, "/")
+		return fmt.Sprintf("https://%s.s3.amazonaws.com", br[0]), br[0], br[1]
+	}
+	// This means it's e.g. "http://minio-server.com:1234/my-bucket"
+	bucket := u.Path[1:]
```
darkdarkdragon, line 67: It can panic here

drivers/s3.go
```diff
@@ -51,16 +52,31 @@ type s3Session struct {
...
 // IsOwnStorageS3 returns true if uri points to S3 bucket owned by this node
 func IsOwnStorageS3(uri string) bool {
-	return strings.HasPrefix(uri, s3Host(S3BUCKET))
+	if S3CONFIG == "" {
+		return false
+	}
+	// We don't care about region here, as it's implicit in host.
+	myHost, _, myBucket := s3Host(S3CONFIG)
+	theirHost, _, theirBucket := s3Host(uri)
+	return (myHost == theirHost) && (myBucket == theirBucket)
```
darkdarkdragon, line 79: `s3Host` will not parse buck name from S3 url, so `theirBucket` will be wrong for own S3 URLs

drivers/s3.go
```diff
@@ -51,16 +52,31 @@ type s3Session struct {
...
 
-func s3Host(bucket string) string {
-	return fmt.Sprintf("https://%s.s3.amazonaws.com", bucket)
+// Returns host, region, bucket
+func s3Host(s3config string) (string, string, string) {
+	u, err := url.Parse(s3config)
+	// This means it's e.g. "us-west-2/my-bucket"
+	if err != nil || u.Scheme == "" {
+		br := strings.Split(s3config, "/")
+		return fmt.Sprintf("https://%s.s3.amazonaws.com", br[0]), br[0], br[1]
```
darkdarkdragon, line 64: should be 
`return fmt.Sprintf("https://%s.s3.amazonaws.com", br[1]), br[0], br[1]`

darkdarkdragon (changes requested):

darkdarkdragon:
> Then I tried the same methodology with a build off of master and still got signature errors. Is there a current S3 bug, perhaps?

Works for me with master with
`livepeer.go -broadcaster  -s3bucket eu-central-1/os-latency-test-1 -s3creds XXX/YYY`

drivers/s3.go
iameli: True. What would be the most appropriate way to handle that? I could add an `err` return value, I suppose.

iameli:
This was supplanted by #1586

PR #1207, opened 2019-11-21
tweak cuda install_ffmpeg parameters, remove build-time cuda dependency

iameli:
just a little build fix, discussion in #1124 

edit: whoops, CUDA wasn't detected at all in this build, please hold...

PR #1208, opened 2019-11-21
build: Update ffmpeg config for CUDA

j0sh:
Splits out the ffmpeg-related updates from https://github.com/livepeer/go-livepeer/pull/1124 in order to make that PR less unwieldy.

- Updates LPMS (and associated joy4) . Requires https://github.com/livepeer/lpms/pull/160 and https://github.com/livepeer/joy4/pull/6
- Update ffmpeg to use Livepeer forked version
- Add h264_cuvid to ffmpeg configure.
- Use LLVM instead of cuda nvcc
- Update unit tests that break as a result of the ffmpeg update
- Misc configure flag cleanup

Existing unit tests

**Checklist:**


darkdarkdragon (commented):
Looks good, but gave `ERROR: cuda_llvm requested but not found` in CI

install_ffmpeg.sh
```diff
@@ -103,7 +103,7 @@ EXTRA_LDFLAGS=""
 if [ $(uname) == "Linux" ]; then
   if [ -e /usr/local/cuda/include ]; then
```
iameli: This isn't a good check anymore, as `clang` doesn't require CUDA as a build-time dependency. I went with

```
```

Really what we want to check for is clang >= 8.0, but I dunno how to do that in shell.

j0sh:
> ERROR: cuda_llvm requested but not found in CI
> This isn't a good check anymore

Both concerns should be addressed with the merge here of https://github.com/livepeer/go-livepeer/pull/1207 - let's see if CI passes.

iameli:
It did! 🎉

iameli (approved):

j0sh:
Rebased and merged, thanks!

PR #1210, opened 2019-11-22
core, pm: validate payment sender

### LINKED ISSUE #1183: Orchestrator should stop working if a broadcaster's funds are unlocked

Currently, an orchestrator will keep working for a broadcaster even after a broadcaster's funds become unlocked. This is a problem because once the broadcaster's funds become unlocked, redemptions for winning tickets sent by the broadcaster will [fail](https://github.com/livepeer/protocol/blob/streamflow/contracts/pm/mixins/MixinTicketBrokerCore.sol#L131). 

The orchestrator already keeps track of whether a broadcaster has [requested an unlock](https://github.com/livepeer/go-livepeer/blob/master/eth/watchers/senderwatcher.go#L223). 

The orchestrator can either:

1. Stop working with the broadcaster if the broadcaster's `withdrawRound` is greater than 0 (meaning that an unlock was requested)
2. Stop working with the broadcaster if the current block is within N blocks of the start block of the broadcaster's `withdrawRound`

Option 1 is simpler and should be fairly quick to implement. Option 2 is a bit more complex, but would allow transcoding to continue even during the unlock period which might be desirable if the unlock period is long.

kyriediculous:
This PR adds functionality that allows an orchestrator to check whether a broadcaster is coming near to the end of its unlock period. If this is the case the orchestrator will not accept segments nor payments from that broadcaster. 

This is accomplished by a `recipient.ValidateSender(sender ethcommon.Address) error` method that is called in `Orchestrator.ProcessPayment()`. This method currently does a single check against an unlock period but can be extended with other functionality if needed. 

This structure allows us to leverage a `RoundsWatcher` and `SenderWatcher` instance that is already active on `SenderMonitor`. The methods for `SenderMonitor` are (for good reason) not directly accessible on the `Recipient` interface that `Orchestrator` uses, therefore the call graph is `core.Orchestrator.ProcessPayment -> Recipient.ValidateSender -> SenderMonitor.IsLocked`.

Alternatively, we could have added a method to `Orchestrator` that accomplishes the same thing, but we would have had to add `SenderWatcher` and `RoundsWatcher` instances to `LivepeerNode` specifically for this. It thus seems a cleaner tradeoff to have a "wrapper/forwarding" function in `Recipient`.

- Added a new method `Recipient.ValidateSender(sender ethcommon.Address) error`
- Added a new method `SenderMonitor.IsLocked(sender ethcommon.Address) error`
- Call `Recipient.ValidateSender` in `Orchestrator.ProcessPayment` , short circuit if the sender is nearing its unlock period
- Added the above methods to stubs

Ran unit tests

Fixes #1183 

**Checklist:**

### DIFF
--- a/pm/recipient.go
+++ b/pm/recipient.go
@@ -138,6 +138,11 @@ func (r *recipient) Stop() {
 func (r *recipient) ReceiveTicket(ticket *Ticket, sig []byte, seed *big.Int) (string, bool, error) {
 	recipientRand := r.rand(seed, ticket.Sender)
 
+	// If sender validation check fails, abort
+	if err := r.sm.ValidateSender(ticket.Sender); err != nil {
+		return "", false, err
+	}
+
 	// If any of the basic ticket validity checks fail, abort
 	if err := r.val.ValidateTicket(r.addr, ticket, sig, recipientRand); err != nil {
 		return "", false, err
--- a/pm/sendermonitor.go
+++ b/pm/sendermonitor.go
@@ -1,6 +1,7 @@
 package pm
 
 import (
+	"fmt"
 	"math/big"
 	"sync"
 	"time"
@@ -37,6 +38,9 @@ type SenderMonitor interface {
 
 	// MaxFloat returns a remote sender's max float
 	MaxFloat(addr ethcommon.Address) (*big.Int, error)
+
+	// ValidateSender checks whether a sender's unlock period ends the round after the next round
+	ValidateSender(addr ethcommon.Address) error
 }
 
 // ErrorMonitor is an interface that describes methods used to monitor acceptable pm ticket errors as well as acceptable price errors
@@ -178,6 +182,19 @@ func (sm *senderMonitor) QueueTicket(addr ethcommon.Address, ticket *SignedTicke
 	sm.ensureCache(addr)
 
 	sm.senders[addr].queue.Add(ticket)
+}
+
+// ValidateSender checks whether a sender's unlock period ends the round after the next round
+func (sm *senderMonitor) ValidateSender(addr ethcommon.Address) error {
+	info, err := sm.smgr.GetSenderInfo(addr)
+	if err != nil {
+		return fmt.Errorf("could not get sender info for %v: %v", addr.Hex(), err)
+	}
+	maxWithdrawRound := new(big.Int).Add(sm.rm.LastInitializedRound(), big.NewInt(1))
+	if info.WithdrawRound.Int64() != 0 && info.WithdrawRound.Cmp(maxWithdrawRound) != 1 {
+		return fmt.Errorf("deposit and reserve for sender %v is set to unlock soon", addr.Hex())
+	}
+	return nil
 }
 
 // maxFloat is a helper that returns the sender's max float as:
--- a/pm/stub.go
+++ b/pm/stub.go
@@ -277,12 +277,13 @@ func (s *stubGasPriceMonitor) GasPrice() *big.Int {
 }
 
 type stubSenderMonitor struct {
-	maxFloat    *big.Int
-	redeemable  chan *SignedTicket
-	queued      []*SignedTicket
-	acceptable  bool
-	addFloatErr error
-	maxFloatErr error
+	maxFloat          *big.Int
+	redeemable        chan *SignedTicket
+	queued            []*SignedTicket
+	acceptable        bool
+	addFloatErr       error
+	maxFloatErr       error
+	validateSenderErr error
 }
 
 func newStubSenderMonitor() *stubSenderMonitor {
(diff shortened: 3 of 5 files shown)


pm/sendermonitor.go
```diff
@@ -180,6 +184,19 @@ func (sm *senderMonitor) QueueTicket(addr ethcommon.Address, ticket *SignedTicke
 	sm.senders[addr].queue.Add(ticket)
 }
 
+// IsLocked checks whether a sender's unlock period ends the round after the next round
+func (sm *senderMonitor) IsLocked(addr ethcommon.Address) error {
+	info, err := sm.smgr.GetSenderInfo(addr)
+	if err != nil {
+		return fmt.Errorf("could not get sender info for %v: %v", addr.Hex(), err)
+	}
+	maxWithdrawRound := new(big.Int).Add(sm.rm.LastInitializedRound(), big.NewInt(1))
```
yondonfu, line 193: I think `nextRound` is a more descriptive variable name here.

pm/sendermonitor.go
```diff
@@ -180,6 +184,19 @@ func (sm *senderMonitor) QueueTicket(addr ethcommon.Address, ticket *SignedTicke
...
+	if info.WithdrawRound != nil && info.WithdrawRound.Int64() != 0 && info.WithdrawRound.Cmp(maxWithdrawRound) != 1 {
```
yondonfu: Don't think `info.WithdrawRound` will ever be nil.

pm/recipient.go
```diff
@@ -134,6 +136,11 @@ func (r *recipient) Stop() {
 	close(r.quit)
 }
 
+// ValidateSender checks whether sender is a valid ticket sender
+func (r *recipient) ValidateSender(sender ethcommon.Address) error {
```
yondonfu: Instead of adding an additional method to the `Recipient` public API we could validate the sender in `ReceiveTicket()`:

```
func (r *recipient) ReceiveTicket(...) (string, bool, error) {
    ...

    if err := r.sm.ValidateSender(ticket.Sender); err != nil {
        return err
    }

    if err := r.val.ValidateTicket(...); err != nil {
        return err
    }

    ...
}
```

I see the benefits of this approach as:

- Reduces the # of public API changes in this PR from 2 to 1 since we would only have to change the `SenderMonitor` public API
- Eliminates the need to modify the `core.Orchestrator` impl and its tests

What do you think?

pm/sendermonitor.go
```diff
@@ -180,6 +184,19 @@ func (sm *senderMonitor) QueueTicket(addr ethcommon.Address, ticket *SignedTicke
 	sm.senders[addr].queue.Add(ticket)
 }
 
+// IsLocked checks whether a sender's unlock period ends the round after the next round
+func (sm *senderMonitor) IsLocked(addr ethcommon.Address) error {
```
yondonfu: This method name is a misnomer - the sender isn't necessarily locked if the method returns an error. What about calling this method `ValidateSender()` (duplicate method names in this type and in `Recipient` wouldn't be a problem if we choose not to modify the `Recipient` public API)?

pm/stub.go
```diff
@@ -283,6 +283,7 @@ type stubSenderMonitor struct {
 	acceptable  bool
 	addFloatErr error
 	maxFloatErr error
+	isLocked    error
```
yondonfu: Would be nice to follow the existing naming convention for error type fields - at the moment, this name would be better suited for a bool type.

yondonfu (changes requested):

pm/recipient.go
kyriediculous: What I don't like about that is that we would be validating the sender for each ticket in the case where a payment constitutes multiple tickets. This seems like an inefficiency to me as we want to validate the sender of a payment as a whole rather than separate tickets. 

We could return an unacceptable payment error from `ReceiveTicket` but this will still try to redeem winning tickets, which I don't see as fair towards B as these errors can be ephemeral and we don't add credit but are redeeming tickets. 

Then again we could do `if won && !unacceptableReceiveErr {...}` or break the loop upon an `unacceptableReceiveError`

pm/sendermonitor.go
kyriediculous: Potentially, I see `maxWithdrawRound` as more general in case we deem the current delta to not be sufficient but in this particular case `nextRound` is more descriptive.

pm/recipient.go
kyriediculous: Fixed in d4b2a96, but I force pushed since the implementation is substantially different and the diffs wouldn't really matter here I guess.

yondonfu: > What I don't like about that is that we would be validating the sender for each ticket in the case where a payment constitutes multiple tickets.

The first ticket should fail validation and we would not validate the rest of the tickets in the batch which seems ok.

> We could return an unacceptable payment error from ReceiveTicket but this will still try to redeem winning tickets

The error returned would already be considered an unacceptable payment error since we are not returning an `AcceptableError` from `ReceiveTicket()` when the sender validation fails.

> which I don't see as fair towards B as these errors can be ephemeral and we don't add credit but are redeeming tickets.

O should use the strategy that best serves its interest - in this case, it should always redeem a valid winning ticket if it receives one. We're treating a sender validation error the same way that we would treat an unacceptable payment error that is triggered due to a sender exceeding its max error count - we will not credit the sender and will not transcode the segment, but if the ticket wins we will still redeem it.

yondonfu (approved):
LGTM 🚢

PR #1212, opened 2019-11-25
pm, server: broadcaster deposit & withdrawround self-check

### LINKED ISSUE #1184: Broadcaster should alert user that stream will not be transcoded if funds are unlocked

Currently, a broadcaster will alert a user that a broadcast session cannot be started if [the deposit is 0](https://github.com/livepeer/go-livepeer/blob/master/server/mediaserver.go#L350). Once #1183 is fixed, a stream will not be transcoded if the broadcaster's funds are unlocked. 

The broadcaster should alert a user that a broadcast session cannot be started if funds are unlocked. If the broadcaster is already mid-stream, it should stop trying to submit segments/payments and notify the user that the broadcast session cannot continue.

kyriediculous:
Currently, a broadcaster in on-chain mode will alert a user that a broadcast session cannot be started if the deposit is 0, we now expand this functionality to also check whether a broadcaster has unlocked it's reserve & deposit and is nearing its `WithdrawRound`

In addition we now favor using cached values (using `SenderWatcher`) over RPC calls to a remote ethereum node to do these checks

- Added a `Validate()` method to `pm.Sender` (perhaps "SelfValidate" is a better name?) 
- Call `pm.Sender.Validate` in `LivepeerServer.registerConnection()` return an error if self validation fails and prevent from sending segments to O's

ran unit tests 

Fixes #1184 

**Checklist:**

**Alternatives Considered**
- Move the validation check to `pm.Sender.StartSession` so we don't have to expand the API surface. We would have to change the signature to return an error when validation logic fails. This will result in an empty sessions list and transcoding will not happen. Harder to propagate the correct error here though. 
- Move the validation check to the constructor `pm.NewSender` and return when we hit an error, this will however shut down the node and prevents a B from re-locking or re-depositing through the CLI

### DIFF
--- a/pm/sender.go
+++ b/pm/sender.go
@@ -75,10 +75,28 @@ func (s *sender) EV(sessionID string) (*big.Rat, error) {
 	return ticketEV(session.ticketParams.FaceValue, session.ticketParams.WinProb), nil
 }
 
+func (s *sender) validateSender() error {
+	info, err := s.senderManager.GetSenderInfo(s.signer.Account().Address)
+	if err != nil {
+		return fmt.Errorf("unable to validate sender: could not get sender info: %v", err)
+	}
+
+	maxWithdrawRound := new(big.Int).Add(s.roundsManager.LastInitializedRound(), big.NewInt(1))
+	if info.WithdrawRound.Int64() != 0 && info.WithdrawRound.Cmp(maxWithdrawRound) != 1 {
+		return fmt.Errorf("unable to validate sender: deposit and reserve is set to unlock soon")
+	}
+
+	return nil
+}
+
 // CreateTicketBatch returns a ticket batch of the specified size
 func (s *sender) CreateTicketBatch(sessionID string, size int) (*TicketBatch, error) {
 	session, err := s.loadSession(sessionID)
 	if err != nil {
+		return nil, err
+	}
+
+	if err := s.validateSender(); err != nil {
 		return nil, err
 	}
 
--- a/pm/sender.go
+++ b/pm/sender.go
@@ -9,7 +9,7 @@ import (
 	"github.com/pkg/errors"
 )
 
-// ErrSenderValidation is emitted when sender.validateSender() encounters an error
+// ErrSenderValidation is returned when the sender cannot send tickets
 type ErrSenderValidation struct {
 	error
 }
--- a/pm/sender.go
+++ b/pm/sender.go
@@ -9,6 +9,11 @@ import (
 	"github.com/pkg/errors"
 )
 
+// ErrSenderValidation is emitted when sender.validateSender() encounters an error
+type ErrSenderValidation struct {
+	error
+}
+
 // Sender enables starting multiple probabilistic micropayment sessions with multiple recipients
 // and create tickets that adhere to each session's params and unique nonce requirements.
 type Sender interface {
@@ -78,12 +83,12 @@ func (s *sender) EV(sessionID string) (*big.Rat, error) {
 func (s *sender) validateSender() error {
 	info, err := s.senderManager.GetSenderInfo(s.signer.Account().Address)
 	if err != nil {
-		return fmt.Errorf("unable to validate sender: could not get sender info: %v", err)
+		return ErrSenderValidation{fmt.Errorf("unable to validate sender: could not get sender info: %v", err)}
 	}
 
 	maxWithdrawRound := new(big.Int).Add(s.roundsManager.LastInitializedRound(), big.NewInt(1))
 	if info.WithdrawRound.Int64() != 0 && info.WithdrawRound.Cmp(maxWithdrawRound) != 1 {
-		return fmt.Errorf("unable to validate sender: deposit and reserve is set to unlock soon")
+		return ErrSenderValidation{fmt.Errorf("unable to validate sender: deposit and reserve is set to unlock soon")}
 	}
 
 	return nil
--- a/pm/sender_test.go
+++ b/pm/sender_test.go
@@ -91,6 +91,54 @@ func TestSenderEV(t *testing.T) {
 	assert.Zero(ticketEV(ticketParams.FaceValue, ticketParams.WinProb).Cmp(ev))
 }
 
+func TestSender_ValidateSender(t *testing.T) {
+	assert := assert.New(t)
+	account := accounts.Account{
+		Address: RandAddress(),
+	}
+	am := &stubSigner{
+		account: account,
+	}
+	rm := &stubRoundsManager{round: big.NewInt(5), blkHash: [32]byte{5}}
+	sm := newStubSenderManager()
+	sm.info[account.Address] = &SenderInfo{
+		Deposit:       big.NewInt(100000),
+		WithdrawRound: big.NewInt(0),
+	}
+	s := &sender{
+		signer:            am,
+		roundsManager:     rm,
+		senderManager:     sm,
+		maxEV:             big.NewRat(100, 1),
+		depositMultiplier: 2,
+	}
+	// GetSenderInfo error
+	sm.err = errors.New("GetSenderInfo error")
+	err := s.validateSender()
+	assert.EqualError(err, "unable to validate sender: could not get sender info: GetSenderInfo error")
+	sm.err = nil
+
+	// Sender's (withdraw round + 1 = current round)
+	sm.info[account.Address].WithdrawRound = big.NewInt(4)
+	err = s.validateSender()
+	assert.EqualError(err, "unable to validate sender: deposit and reserve is set to unlock soon")
+
+	// withdrawround + 1 < current round
+	sm.info[account.Address].WithdrawRound = big.NewInt(2)
+	err = s.validateSender()
+	assert.EqualError(err, "unable to validate sender: deposit and reserve is set to unlock soon")
+
+	// Not unlocked
+	sm.info[account.Address].WithdrawRound = big.NewInt(0)
+	err = s.validateSender()
+	assert.NoError(err)
+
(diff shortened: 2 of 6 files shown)


pm/sender.go
```diff
@@ -74,6 +78,24 @@ func (s *sender) EV(sessionID string) (*big.Rat, error) {
 	return ticketEV(session.ticketParams.FaceValue, session.ticketParams.WinProb), nil
 }
 
+func (s *sender) Validate() error {
```
yondonfu: Instead of adding an additional method to the `Sender` public API we could move these checks into an internal method which is called in `CreateTicketBatch()` (similar to how we do the ticket params checks):

```go
func (s *sender) CreateTicketBatch(...) (*TicketBatch, error) {
    ...

    if err := s.validateSender(); err != nil {
        return err
    }

    if err := s.validateTicketParams(&session.ticketParams, size); err != nil {
        return err
    }

    ...
}
```

I see the benefits of this approach as:

- Eliminates the need to modify the `Sender` public API
- Removes some code from `mediaserver.go`
- Runs sender validation mid-stream with each payment that B prepares for an O for a stream instead of just running sender validation at the beginning of a stream when registering the RTMP connection

The two downsides of this approach are:
- The RTMP connection will not be closed immediately if one of these sender validation checks would fail because the checks are not run until after the RTMP connection is already established
- At the moment, the RTMP connection will not be closed mid-stream if one of these sender validation checks fail because we [always retry segments and do not propagate/handle errors from `transcodeSegment()`](https://github.com/livepeer/go-livepeer/blob/master/server/broadcast.go#L279)

I think the first downside is a reasonable tradeoff - we incur a little extra cost for RTMP connection setup, but we get to isolate the sender validation logic in the `pm` package which will be executed whenever a ticket batch is being prepared via `CreateTicketBatch()`.

As for the second downside, outside of this PR, we could introduce a non-retryable error as suggested by this [comment](https://github.com/livepeer/go-livepeer/blob/master/server/broadcast.go#L298). If a non-retryable error, such as failing sender validation checks), is encountered then the error would be propagated from `transcodeSegment()` up to the caller of `processSegment()` which can then close the RTMP connection to alert the user. Prior to the introduction of a non-retryable error, B would just keep logging errors from sender validation as it continually retries segments (without actually submitting the segments).

pm/sender.go
```diff
@@ -74,6 +78,24 @@ func (s *sender) EV(sessionID string) (*big.Rat, error) {
...
+	info, err := s.senderManager.GetSenderInfo(s.signer.Account().Address)
+	if err != nil {
+		return fmt.Errorf("could not get sender info: %v", err)
+	}
+
+	if info.Deposit.Cmp(big.NewInt((0))) == 0 {
```
yondonfu: I don't think we need this deposit check since `validateTicketParams()` already has a deposit check (based on the `faceValue` required by O).

yondonfu (changes requested):

pm/sender.go
kyriediculous: The reason I included it was because we used to also just close the RTMP connection when B had no deposit.

fixed in 35979be

pm/sender.go
kyriediculous: fixed in 35979be but will need to track the extra PR needed in a seperate issue.

yondonfu: @kyriediculous Can you create an issue to track the additional work needed to close the RTMP connections when the `validateSender()` check fails?

pm/sender.go
```diff
@@ -74,13 +75,31 @@ func (s *sender) EV(sessionID string) (*big.Rat, error) {
...
 }
 
+func (s *sender) validateSender() error {
+	info, err := s.senderManager.GetSenderInfo(s.signer.Account().Address)
+	if err != nil {
+		return fmt.Errorf("unable to validate sender: could not get sender info: %v", err)
+	}
+
+	maxWithdrawRound := new(big.Int).Add(s.roundsManager.LastInitializedRound(), big.NewInt(1))
+	if info.WithdrawRound != nil && info.WithdrawRound.Int64() != 0 && info.WithdrawRound.Cmp(maxWithdrawRound) != 1 {
```
yondonfu: Don't think `info.WithdrawRound` will ever be nil

pm/sender_test.go
```diff
@@ -90,6 +90,49 @@ func TestSenderEV(t *testing.T) {
...
+	assert.EqualError(err, "unable to validate sender: could not get sender info: GetSenderInfo error")
+	sm.err = nil
+
+	// sender's (withdraw round + 1 <= current round)
+	sm.info[account.Address].WithdrawRound = big.NewInt(4)
+	err = s.validateSender()
+	assert.EqualError(err, "unable to validate sender: deposit and reserve is set to unlock soon")
+	sm.info[account.Address].WithdrawRound = big.NewInt(0)
+
+	// Sufficient deposit and not unlocked
```
yondonfu: We aren't checking the deposit any more so this and the other comment below need to be updated

yondonfu (changes requested):

pm/sender_test.go
kyriediculous: fixed in 4fb8b5d2

pm/sender.go
kyriediculous: fixed in 4fb8b5d2

kyriediculous: actually this is nil in most of the tests unless we set WithdrawRound in every test in `sender_test`  instead of simply using `defaultSender` helper so would you be ok with keeping it there (otherwise all tests expect TestValidateSender will have nil pointer errors) ?

kyriediculous: fixed in 915b0de3

pm/sender_test.go
```diff
@@ -90,6 +90,49 @@ func TestSenderEV(t *testing.T) {
...
+	sm.info[account.Address].WithdrawRound = big.NewInt(4)
+	err = s.validateSender()
+	assert.EqualError(err, "unable to validate sender: deposit and reserve is set to unlock soon")
+	sm.info[account.Address].WithdrawRound = big.NewInt(0)
+
+	// Not unlocked
+	err = s.validateSender()
+	assert.NoError(err)
+
+	// Unlocked but (withdrawRound + 1 >= current round)
```
yondonfu: Think this should say "withdraw round + 1 > current round"

kyriediculous: fixed and rebased onto master

pm/sender_test.go
```diff
@@ -91,6 +91,49 @@ func TestSenderEV(t *testing.T) {
...
+		depositMultiplier: 2,
+	}
+	// GetSenderInfo error
+	sm.err = errors.New("GetSenderInfo error")
+	err := s.validateSender()
+	assert.EqualError(err, "unable to validate sender: could not get sender info: GetSenderInfo error")
+	sm.err = nil
+
+	// Sender's (withdraw round + 1 <= current round)
+	sm.info[account.Address].WithdrawRound = big.NewInt(4)
```
yondonfu, line 124: Test for when withdraw round + 1 = current round? (Missed this in the initial review)

pm/sender_test.go
```diff
@@ -91,6 +91,49 @@ func TestSenderEV(t *testing.T) {
 	assert.Zero(ticketEV(ticketParams.FaceValue, ticketParams.WinProb).Cmp(ev))
 }
 
+func TestValidateSender(t *testing.T) {
```
yondonfu: Think this is triggering a CI error since this func name is already declared https://github.com/livepeer/go-livepeer/blob/master/pm/sendermonitor_test.go#L473

pm/sender_test.go
kyriediculous: the current round is set to `5`, so would `4 + 1 ` fulfil that scenario ?  We don't have a test testing `withdrawRound + 1 < current round` though.

pm/sender_test.go
kyriediculous: indeed, fixed here 1cec746

pm/sender_test.go
yondonfu: Ah yes my mistake. We are missing a test for when withdraw round + 1 < current round.

pm/sendermonitor_test.go
```diff
@@ -470,49 +469,6 @@ func TestReserveAlloc(t *testing.T) {
 	assert.Zero(expectedAlloc.Cmp(alloc))
 }
 
-func TestValidateSender(t *testing.T) {
```
yondonfu, line 473: This test is for `SenderMonitor.ValidateSender()` so we don't want to remove it.

kyriediculous: Oh yeah my bad, they looked the same at first glance, fixed acefc0d

pm/sender_test.go
kyriediculous: added test acefc0d

pm/sendermonitor_test.go
```diff
@@ -470,7 +470,7 @@ func TestReserveAlloc(t *testing.T) {
...
+func TestValidateSender_SenderMonitor(t *testing.T) {
```
yondonfu: `<TYPE_NAME>`_`<METHOD_NAME>` feels a bit clearer to me as it gets more specific about what is under test (type under test followed by method under test) as you continue reading the test name. WDYT?

kyriediculous: fixed in 3b14e74 (added it for Both `ValidateSender` tests now)

yondonfu (commented):

kyriediculous:
rebased !

yondonfu (approved):
LGTM

PR #1216, opened 2019-11-26
cmd, common, eth: cache orchestrator stake

### LINKED ISSUE #1213: Cache stake of all active orchestrators at when a new round is initialized

The stake of active orchestrators will be used in selection. Instead, of querying for the stake of an active orchestrator every time we need it, we can cache the stake of all active orchestrators when a new round is initialized because the stake of an active orchestrator is locked in for the duration of a round when it is initialized. 

- [ ] Add a `stake` column to the `orchestrators` table in the DB
- [ ] When a new round is initialized, fetch all the active orchestrators from the DB, iterate through each of them and write their stake for the round to the DB
    - The DB caching logic can be placed in `watchers.OrchestratorWatcher` which is already responsible for updating the DB based on on-chain activation/deactivation events
    - We can subscribe to new round initialization events using `watchers.RoundsWatcher.Subscribe()`
    - For each active orchestrator in the DB, we can call `LivepeerEthClient.GetTranscoderEarningsPoolForRound()` using the orchestrator's address and the current round. The struct returned by this method will include a `totalStake` field that will represent the orchestrator's stake for the round

kyriediculous:
This PR caches the stake for orchestrators in the active set whenever a new round event is emitted. 

This can be both a removed log (re-org) or an added log (new round), in both cases we make an RPC call for the current round number and a subsequent RPC call for each O to fetch its stake for that specific round. We then update the orchestrator in the DB with the new stake. 

- Added a `stake` column to the  `orchestrators` table and updated `UpdateOrch` and `SelectOrchs` methods accordingly 
- Added a `RoundsWatcher` subscription to `OrchestratorWatcher`
- Added internal helper methods `handleRoundEvent` and `cacheOrchestratorStake` to `OrchestratorWatcher`
- updated `StubEthClient.GetTranscoderPoolForRound`
- Cache stake for orchestrators when `DBOrchestratorPoolCache` is created

ran unit tests

Fixes #1213 

**Checklist:**

### DIFF
--- a/discovery/db_discovery.go
+++ b/discovery/db_discovery.go
@@ -58,6 +58,10 @@ func NewDBOrchestratorPoolCache(ctx context.Context, node *core.LivepeerNode, rm
 		return nil, err
 	}
 
+	if err := dbo.cacheOrchestratorStake(); err != nil {
+		return nil, err
+	}
+
 	if err := dbo.pollOrchestratorInfo(ctx); err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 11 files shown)


kyriediculous:
Changed base branch to #1182 for now and added caching stake when starting the node.

yondonfu:
Looks like there are changes from #1182 pulled into this PR (due to the rewording of commit history in #1182). Mind rebasing on top of #1182?

kyriediculous:
dropped the commits and rebased on latest #1182

discovery/db_discovery.go
```diff
@@ -145,6 +149,53 @@ func (dbo *DBOrchestratorPoolCache) cacheTranscoderPool() error {
...
+		if err != nil {
+			errc <- err
+			return
+		}
+		o.Stake = ep.TotalStake.String()
+		resc <- o
+	}
+
+	for _, o := range orchs {
+		go getStake(o)
```
yondonfu, line 179: Creating a goroutine for each query feels like a premature optimization at this point IMO since it introduces a piece of code that needs to be concurrency safe (and ideally we would test for it). Caching the stake for each active O isn't that time sensitive since it just happens on startup. Additionally, a simpler optimization that we could use in the future is to write a wrapper contract that aggregates all the stake values for a list of Os so that we can fetch the results with a single query.

yondonfu (changes requested):

discovery/db_discovery.go
kyriediculous: If I'm not mistaken go-sqlite does use a RWmutex so this should be thread safe regardless, since we're only writing the value to the DB and not memory.

yondonfu: Yeah good point. On second thought, I'm fine with the concurrent requests used here 👍 with the wrapper contract as a nice to have optimization later on that removes the need for any concurrency code.

yondonfu (approved):
LGTM

PR #1219, opened 2019-11-26
eth/watchers: store serviceURI on TranscoderActivated event

### LINKED ISSUE #1217: OrchestratorWatcher should store ServiceURI for orchestrators that become active
**Describe the bug**
Currently `OrchestratorWatcher` only inserts an O's `activationRound` and `deactivationRound` into the DB when we receive a `TranscoderActivated` event. 

Since we're no longer backfilling from a block far in the past, which would synchronize all historical URI updates, we will end up never saving a URI for orchestrators that are added through the DB by emitted `TranscoderActivated` events.

Because there is no URI for these transcoders they will never be included in the working set as we can't retrieve off-chain info for them. 

**Expected behavior**
`OrchestratorWatcher` should fetch an O's `ServiceURI` from `ServiceRegistry` when it receives a `TranscoderActivated` event and store it in DB.

kyriediculous:
This PR fetches and stores an O's `ServiceURI` when handling a `TranscoderActivated` event. 
Since we're no longer backfilling from a block far in the past, which would synchronize all historical URI updates, we will end up never saving a URI for orchestrators that are added through the DB by emitted TranscoderActivated events.

Because there is no URI for these transcoders they will never be included in the working set as we can't retrieve off-chain info for them.

- Added a call to `LivepeerEthClient.GetServiceURI` when a `TranscoderActivated` event is added and populate the `ServiceURI` field on the `DBOrch` struct that's passed into `OrchestratorStore.UpdateOrch`
- In case a `TranscoderActivated` event is removed we already have access to an O's `ServiceURI` because we call `LivepeerEthClient.GetTranscoder` , now we also propagate `ServiceURI` into the BD

ran unit tests

Fixes #1217 

**Checklist:**

### DIFF
--- a/eth/watchers/orchestratorwatcher.go
+++ b/eth/watchers/orchestratorwatcher.go
@@ -97,9 +97,15 @@ func (ow *OrchestratorWatcher) handleTranscoderActivated(log types.Log) error {
 	}
 
 	if !log.Removed {
+		uri, err := ow.lpEth.GetServiceURI(transcoderActivated.Transcoder)
+		if err != nil {
+			return err
+		}
+
 		return ow.store.UpdateOrch(
 			&common.DBOrch{
 				EthereumAddr:      transcoderActivated.Transcoder.String(),
+				ServiceURI:        uri,
 				ActivationRound:   transcoderActivated.ActivationRound.Int64(),
 				DeactivationRound: maxFutureRound,
 			},
@@ -112,6 +118,7 @@ func (ow *OrchestratorWatcher) handleTranscoderActivated(log types.Log) error {
 	return ow.store.UpdateOrch(
 		&common.DBOrch{
 			EthereumAddr:      t.Address.String(),
+			ServiceURI:        t.ServiceURI,
 			ActivationRound:   t.ActivationRound.Int64(),
 			DeactivationRound: t.DeactivationRound.Int64(),
 		},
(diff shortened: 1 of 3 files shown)


eth/watchers/orchestratorwatcher.go
```diff
@@ -97,9 +97,15 @@ func (ow *OrchestratorWatcher) handleTranscoderActivated(log types.Log) error {
 	}
 
 	if !log.Removed {
+		uri, err := ow.lpEth.GetServiceURI(transcoderActivated.Transcoder)
+		if err != nil {
```
yondonfu, line 101: Let's add a test for this error case

yondonfu (changes requested):

eth/watchers/orchestratorwatcher.go
kyriediculous: fixed in 285268c7, but also added a test for `GetTranscoder` error (in case log is removed) and opted to use a single `Err` field on the `StubEthClient` rather than creating a seperate field for each case (b3d0d034). 

Can rework some other tests to use this single `Err` field in a seperate PR but this is not a priority.

yondonfu (commented):

kyriediculous:
rebased !

yondonfu (approved):
LGTM 🚢

PR #1226, opened 2019-11-29
cmd, core, eth/watchers: O checks that it is active in ProcessPayment

### LINKED ISSUE #1202: Orchestrator should check if it is active before accepting work

At the moment, an orchestrator will accept segments/payments even if it is not in the active set during the current round. A non-active orchestrator will not be able to draw from a broadcaster's reserve in the event that the broadcaster's deposit is insufficient to cover all of the winning tickets that it has sent out.

An orchestrator should check if it is active in the current round before accepting segments/payments. If the orchestrator is not active it should reject segments/payments. The orchestrator operator should be able to override this behavior and accept segments/payments from certain broadcasters even if the orchestrator is not active (which is useful for testing). The orchestrator could accept an allow list of broadcaster ETH addresses.

kyriediculous:

This PR adds the necessary functionality for an O to check whether it is active before accepting payments/segments. 

- Added an `AddressFilter` to `OrchestratorWatcher` which allows the `OrchestratorWatcher` implementation to only populate the DB for O itself when running in orchestrator mode
- Now starting `OrchestratorDBPoolCache` in on-chain mode for both O and B, in orchestrator mode the pool will only have one entry , the orchestrator itself. And `OrchestratorPool.Size()` will return 0 if the orchestrator is inactive and 1 if it is active
- Added a helper method `orchestrator.isActive() bool` that checks whether `OrchestratorPool.Size() > 0`
- Added a call to `orchestrator.isActive()` in `orchestrator.ProcessPayment()`, if the orchestrator is inactive we'll return an error. 

Updated & ran unit tests

Fixes #1202 

**Checklist:**

### DIFF
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -745,7 +745,7 @@ func main() {
 			return
 		}
 
-		orch := core.NewOrchestrator(s.LivepeerNode)
+		orch := core.NewOrchestrator(s.LivepeerNode, roundsWatcher)
 
 		go func() {
 			server.StartTranscodeServer(orch, *httpAddr, s.HTTPMux, n.WorkDir, n.TranscoderManager != nil)
--- a/core/orch_test.go
+++ b/core/orch_test.go
@@ -1,6 +1,7 @@
 package core
 
 import (
+	"database/sql"
 	"errors"
 	"fmt"
 	"io/ioutil"
@@ -14,6 +15,7 @@ import (
 	"github.com/golang/glog"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/mock"
+	"github.com/stretchr/testify/require"
 
 	"github.com/livepeer/go-livepeer/pm"
 
@@ -32,7 +34,8 @@ func TestCurrentBlock(t *testing.T) {
 		t.Error(err)
 	}
 	defer os.RemoveAll(tmpdir)
-	orch := NewOrchestrator(n)
+	rm := &stubRoundsManager{}
+	orch := NewOrchestrator(n, rm)
 
 	// test empty db
 	if orch.CurrentBlock() != nil {
@@ -514,10 +517,9 @@ func TestGetSegmentChan(t *testing.T) {
 }
 
 func TestOrchCheckCapacity(t *testing.T) {
-
 	drivers.NodeStorage = drivers.NewMemoryDriver(nil)
 	n, _ := NewLivepeerNode(nil, "", nil)
-	o := NewOrchestrator(n)
+	o := NewOrchestrator(n, nil)
 	md := StubSegTranscodingMetadata()
 	cap := MaxSessions
 	assert := assert.New(t)
@@ -538,13 +540,27 @@ func TestOrchCheckCapacity(t *testing.T) {
 }
 
 func TestProcessPayment_GivenRecipientError_ReturnsNil(t *testing.T) {
-	n, _ := NewLivepeerNode(nil, "", nil)
+	addr := pm.RandAddress()
+	dbh, dbraw := tempDBWithOrch(t, &common.DBOrch{
+		EthereumAddr:      addr.Hex(),
+		ActivationRound:   1,
+		DeactivationRound: 999,
+	})
+	defer dbh.Close()
+	defer dbraw.Close()
+
+	n, _ := NewLivepeerNode(nil, "", dbh)
 	n.Balances = NewAddressBalances(5 * time.Second)
 	recipient := new(pm.MockRecipient)
 	n.Recipient = recipient
-	orch := NewOrchestrator(n)
+	rm := &stubRoundsManager{
+		round: big.NewInt(10),
+	}
+	orch := NewOrchestrator(n, rm)
+	orch.address = addr
 	orch.node.SetBasePrice(big.NewRat(0, 1))
 	orch.node.ErrorMonitor = NewErrorMonitor(0, make(chan struct{}))
+
 	recipient.On("TxCostMultiplier", mock.Anything).Return(big.NewRat(1, 1), nil)
 
 	recipient.On("ReceiveTicket", mock.Anything, mock.Anything, mock.Anything).Return("", false, nil)
@@ -558,8 +574,7 @@ func TestProcessPayment_GivenNoSender_ReturnsError(t *testing.T) {
 	n, _ := NewLivepeerNode(nil, "", nil)
 	recipient := new(pm.MockRecipient)
 	n.Recipient = recipient
-	orch := NewOrchestrator(n)
-	recipient.On("ReceiveTicket", mock.Anything, mock.Anything, mock.Anything).Return("some sessionID", false, nil)
+	orch := NewOrchestrator(n, nil)
 
 	protoPayment := defaultPayment(t)
 
@@ -577,7 +592,7 @@ func TestProcessPayment_GivenNoTicketParams_ReturnsNoError(t *testing.T) {
 	n, _ := NewLivepeerNode(nil, "", nil)
 	recipient := new(pm.MockRecipient)
 	n.Recipient = recipient
-	orch := NewOrchestrator(n)
+	orch := NewOrchestrator(n, nil)
 	recipient.On("ReceiveTicket", mock.Anything, mock.Anything, mock.Anything).Return("some sessionID", false, nil)
 
 	protoPayment := defaultPayment(t)
@@ -600,7 +615,7 @@ func TestProcessPayment_GivenNilNode_ReturnsNilError(t *testing.T) {
 
 func TestProcessPayment_GivenNilRecipient_ReturnsNilError(t *testing.T) {
 	n, _ := NewLivepeerNode(nil, "", nil)
-	orch := NewOrchestrator(n)
+	orch := NewOrchestrator(n, nil)
 	n.Recipient = nil
 
 	err := orch.ProcessPayment(defaultPayment(t), ManifestID("some manifest"))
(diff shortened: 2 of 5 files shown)


core/orchestrator.go
```diff
@@ -316,6 +320,10 @@ func (orch *orchestrator) acceptablePrice(sender ethcommon.Address, ep *net.Pric
 	return nil
 }
 
+func (orch *orchestrator) isActive() bool {
+	return orch.node.OrchestratorPool.Size() > 0
```
yondonfu: Using `OrchestratorPool` for this check feels a bit weird to me because:

1. `OrchestratorPool` is generally used for discovery by B
2. The current changes adds address filtering to `OrchestratorWatcher`, but the DB will still be populated with other addresses due to the call to `cacheTranscoderPool()` in the `DBOrchestratorPoolCache` constructor. We can add address filtering to the `cacheTranscoderPool()` method, but we would be adding address filtering in two places now meaning two separate areas of the codebase that we need to make sure implements address filtering properly or else `orch.node.OrchestratorPool.Size() > 0` will not be an accurate active status check

An alternative would be to query the DB directly for the active status check. #1253 adds an `Addresses` field to `DBOrchFilter` - if the only addresses included in the filter is O's address and the result returned for `SelectOrchs()` has a length > 0 then O is active.

Ex.

```
func (orch *orchestrator) isActive() (bool, error) {
    filter := &DBOrchFilter{
        CurrentRound: orch.roundsManager.LastInitializedRound(),
        Addresses: []ethcommon.Address{orch.address},
    }
    orchs, err := orch.node.Database.SelectOrchs(filter)
    if err != nil {
        return false, err
    }

    return len(orchs) > 0, nil
}
```

The upside of this approach is that it limits the scope of changes to just the `orchestrator` type and it makes use of address filtering which is already implemented at a SQL level (so we don't need to implement it at a Go level).

The above code example assumes that `orchesrator` is initialized with a type that defines the `LastInitializedRound()` method. Another approach to consider is to update `watches.RoundsWatcher` to write directly to the DB so that we simply read the last initialized round from `node.Database` which would eliminate the need to pass around an additional type. Downside is that the value would be read from disk instead of memory.

kyriediculous: Regarding (1) I do remember you said it _could_ be nice to let O have access to `OrchestratorPool` (in on-chain mode) as well in the future. This would be a step in that direction while not introducing unnecessary inefficiencies yet. 

Regarding (2) , didn't think of that, that's true. In hindsight I should have probably covered this with a unit test.

I think it makes sense to write the last initialized round to the DB and use the address filter

kyriediculous: fixed in 53fd42b6

common/db.go
```diff
@@ -394,10 +395,29 @@ func (db *DB) ChainID() (*big.Int, error) {
 }
 
 func (db *DB) SetChainID(id *big.Int) error {
-	if err := db.updateKVStore("chainID", id.String()); err != nil {
-		return err
+	return db.updateKVStore("chainID", id.String())
+}
+
+func (db *DB) CurrentRound() (*big.Int, error) {
```
yondonfu: `LastInitializedRound()` (with the key name as `lastInitializedRound`) so that all types that currently depend on an interface that defines `LastInitializedRound()` can just plug in `DB` instead?

common/db.go
```diff
@@ -394,10 +395,29 @@ func (db *DB) ChainID() (*big.Int, error) {
...
+
+	r, ok := new(big.Int).SetString(round, 10)
+	if !ok {
+		return nil, fmt.Errorf("unable to convert current round string to big.Int")
+	}
+
+	return r, nil
+}
+
+func (db *DB) SetCurrentRound(round *big.Int) error {
```
yondonfu: `SetLastInitializedRound()`?

cmd/livepeer/livepeer.go
```diff
@@ -429,7 +429,12 @@ func main() {
 		go senderWatcher.Watch()
 		defer senderWatcher.Stop()
 
-		orchWatcher, err := watchers.NewOrchestratorWatcher(addrMap["BondingManager"], blockWatcher, dbh, n.Eth, roundsWatcher)
+		// When in orchestrator mode only process TranscoderActivated/TranscoderDeactivated events for O
+		var addressFilter []ethcommon.Address
```
yondonfu: Don't need to filter addresses in `OrchestratorWatcher` anymore in this PR.

core/orchestrator.go
yondonfu: >  do remember you said it could be nice to let O have access to OrchestratorPool (in on-chain mode) as well in the future. This would be a step in that direction while not introducing unnecessary inefficiencies yet.

Hmm IIRC we were talking about letting O have access to the state of the full active set (i.e. keeping O's DB updated with all active set updates) - this would be the case as long as we continue to update the DB for active set updates regardless of what mode the node is running in.

> I think it makes sense to write the last initialized round to the DB

Thinking about this a bit more, I am a bit wary about switching from reading from disk to reading from memory for the last initialized round since the value is read quite often when O receives a segment/payment:

- Used to validate the sender for every ticket in a payment
- Used to check the creation round for every ticket in a payment
- Used to check if O is active

Since these checks are on the critical path for transcoding, I'd actually lean towards continuing to read the value from memory at the cost of passing around an additional type.

cmd/livepeer/livepeer.go
kyriediculous: Mind that orchestratorWatcher is also started for O and will process all activation and deactivation events for no reason. Then again I'm not sure what the inefficiency at runtime will be , it can be minor if active set churn is low.

yondonfu: > Mind that orchestratorWatcher is also started for O and will process all activation and deactivation events for no reason.

This all happens in the background so I don't think this is an immediate concern.

kyriediculous:
> Thinking about this a bit more, I am a bit wary about switching from reading from disk to reading from memory for the last initialized round since the value is read quite often when O receives a segment/payment:
>
> - Used to validate the sender for every ticket in a payment
> - Used to check the creation round for every ticket in a payment
> - Used to check if O is active

Well as is we would only read it from the DB to do the active check, the other packages that use a roundsWatcher still read from memory, perhaps we can make that judgement on a per-package basis. 

For example none of the current pm packages use the DB as a dependency only `RoundInitializer` uses both `RoundsWatcher` and DB. So for the pm packages it would make sense to just import `RoundsWatcher` as a dependency while those that already use the DB to read/write can use the DB to read the last round.

yondonfu:
> Well as is we would only read it from the DB to do the active check, the other packages that use a roundsWatcher still read from memory, perhaps we can make that judgement on a per-package basis.

Reading from the DB for the active check, but reading from memory in the `pm` package seems weird to me unless there is a good reason why the active check should read from the DB as opposed to from memory while the `pm` package always reads from memory. The only reason that I see right now is for the sake of using `LivepeerNode.Database` which `orchestrator` already has access to. While this reason has merit, the upside does not seem to outweigh the downside of incurring a slower DB read operation in a piece of code that is on the performance critical path of transcoding.

core/orch_test.go
```diff
@@ -955,13 +1171,29 @@ func TestProcessPayment_AcceptablePriceError_IncreasesCredit_ReturnsError(t *tes
...
+	rm := &stubRoundsManager{
+		round: big.NewInt(10),
+	}
+	orch := NewOrchestrator(n, rm)
+
+	orch.address = pm.RandAddress()
 	orch.node.SetBasePrice(big.NewRat(5, 1))
 	orch.node.ErrorMonitor = NewErrorMonitor(0, make(chan struct{}))
+
+	dbh.UpdateOrch(&common.DBOrch{
```
yondonfu: Seems that we have to repeatedly initialize a DB with a single O in many of this PR's tests. Worth considering introducing a test helper for initializing a DB that starts off in the right state. Something like:

```
func tempDBWithOrch(t *testing.T, orch *common.DBOrch) (*common.DB, *sql.DB) {
    return tempDBWithOrchs(t, []*common.DBOrch{orch})
}

func tempDBWithOrchs(t *testing.T, orchs []*common.DBOrch) (*common.DB, *sql.DB) {
    dbh, dbraw, err := common.TempDB(t)
    require.Nil(t, err)

    for _, orch := range orchs {
        require.Nil(t, dbh.UpdateOrch(orch))
    }

    return dbh, dbraw
}
```

(The extra `tempDBWithOrchs()` might be overkill)

Then, the helper could be used to reduce the boilerplate in each updated test:

```
addr := pm.RandAddress()
dbh, dbraw := tempDBWithOrch(&common.DBOrch{...})
defer dbh.Close()
defer dbraw.Close()
```

WDYT?

core/orch_test.go
```diff
@@ -589,26 +609,80 @@ func TestProcessPayment_GivenNilNode_ReturnsNilError(t *testing.T) {
...
+		EthereumAddr:      orch.Address().Hex(),
+		ActivationRound:   1,
+		DeactivationRound: 999,
+	})
+
+	recipient.On("TxCostMultiplier", mock.Anything).Return(big.NewRat(1, 1), nil)
+	recipient.On("ReceiveTicket", mock.Anything, mock.Anything, mock.Anything).Return("some sessionID", false, nil)
+	err = orch.ProcessPayment(defaultPayment(t), ManifestID("some manifest"))
+	time.Sleep(time.Millisecond * 20)
```
yondonfu: Why is this sleep needed?

core/orch_test.go
```diff
@@ -1349,7 +1675,10 @@ func TestPriceInfo_TxMultiplierError_ReturnsError(t *testing.T) {
 func TestDebitFees(t *testing.T) {
 	n, _ := NewLivepeerNode(nil, "", nil)
 	n.Balances = NewAddressBalances(5 * time.Second)
-	orch := NewOrchestrator(n)
+	rm := &stubRoundsManager{
```
yondonfu: Can most of these `stubRoundsManager` constructor invocations be empty except for the case where `ProcessPayment()` is invoked in the test?

yondonfu (changes requested):
Looks pretty good - just had a few test related comments

core/orch_test.go
kyriediculous: fixed in 35d8e94a

core/orch_test.go
kyriediculous: fixed in 35d8e94a

core/orch_test.go
kyriediculous: fixed in adaac65b

core/orch_test.go
```diff
@@ -32,7 +34,10 @@ func TestCurrentBlock(t *testing.T) {
 		t.Error(err)
 	}
 	defer os.RemoveAll(tmpdir)
-	orch := NewOrchestrator(n)
+	rm := &stubRoundsManager{
+		round: big.NewInt(10),
```
yondonfu: Empty constructor should be fine here?

core/orch_test.go
```diff
@@ -1094,14 +1304,63 @@ func TestAcceptablePrice_PriceInfoError_ReturnsErr(t *testing.T) {
...
+	}
+	orch := NewOrchestrator(n, rm)
+	orch.address = addr
 	orch.node.SetBasePrice(big.NewRat(0, 1))
 	orch.node.ErrorMonitor = NewErrorMonitor(0, make(chan struct{}))
+
+	dbh.UpdateOrch(&common.DBOrch{
```
yondonfu: Doesn't seem like we need this call

yondonfu (changes requested):

yondonfu:
Addressed the two small code cleanup comments in tests and rebased.

yondonfu (approved):
LGTM

PR #1230, opened 2019-12-02
server: don't update feeshare and rewardcut on-chain if values are unchanged

### LINKED ISSUE #1167: transcoder() is called during "Set orchestrator config" even if reward cut and fee share are unchanged
**Describe the bug**

When calling `Set orchestrator config` using the `livepeer_cli`, the process will call `transcoder()` function even when the `rewardCut` and `feeShare` parameters are not changed.

This is a waste of a transaction.

**To Reproduce**

1. Run livepeer in `-orchestrator` and `-transcoder mode`
2. Run livepeer_cli and `Invoke multi-step "become an orchestrator"`
3. See that it calls `transcoder` (among other functions)
4. Then run `Set orchestrator config`
5. Use the same parameters for `rewardCut` and `feeShare` as in step 2
6. See that it calls `transcoder` again.

**Expected behavior**

I expect that if the `rewardCut` and `feeShare` parameters entered into `Set orchestrator config` are the same as those already configured onchain, then the process shouldn't call `transcoder` again.

```
livepeer-cli version 0.5.0
```
```
Livepeer Node Version: 0.5.0
Golang runtime version: gc go1.13
Architecture: amd64
Operating system: linux
```

kyriediculous:

This PR prevents the webserver from submitting a `BondingManager.transcoder(rewardcut, feeshare)` transaction to set these values on-chain when the values itself aren't changed. This prevents unnecessary gas expenditure. 

- Moved the RPC call to `BondingManager.getTranscoder` above the call to `BondingManager.transcoder`
- Added a check whether feeshare and rewardcut are the same as the on-chain values, only send a tx when either value or both values are changed

Fixes #1167 

**Checklist:**

### DIFF
--- a/server/webserver.go
+++ b/server/webserver.go
@@ -363,28 +363,30 @@ func (s *LivepeerServer) cliWebServerHandlers(bindAddr string) *http.ServeMux {
 			return
 		}
 
-		glog.Infof("Setting orchestrator commission rates %v", s.LivepeerNode.Eth.Account().Address.Hex())
-
-		tx, err := s.LivepeerNode.Eth.Transcoder(eth.FromPerc(blockRewardCut), eth.FromPerc(feeShare))
+		t, err := s.LivepeerNode.Eth.GetTranscoder(s.LivepeerNode.Eth.Account().Address)
 		if err != nil {
 			glog.Error(err)
 			return
 		}
 
-		err = s.LivepeerNode.Eth.CheckTx(tx)
-		if err != nil {
-			glog.Error(err)
-			return
+		if t.RewardCut.Cmp(eth.FromPerc(blockRewardCut)) != 0 || t.FeeShare.Cmp(eth.FromPerc(feeShare)) != 0 {
+			tx, err := s.LivepeerNode.Eth.Transcoder(eth.FromPerc(blockRewardCut), eth.FromPerc(feeShare))
+			if err != nil {
+				glog.Error(err)
+				return
+			}
+
+			glog.Infof("Setting orchestrator commission rates for %v: reward cut=%v feeshare=%v", s.LivepeerNode.Eth.Account().Address.Hex(), blockRewardCut, feeShare)
+
+			err = s.LivepeerNode.Eth.CheckTx(tx)
+			if err != nil {
+				glog.Error(err)
+				return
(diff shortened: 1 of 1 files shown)


server/webserver.go
```diff
@@ -363,32 +363,34 @@ func (s *LivepeerServer) cliWebServerHandlers(bindAddr string) *http.ServeMux {
...
 		if err != nil {
 			glog.Error(err)
 			return
 		}
 
-		err = s.LivepeerNode.Eth.CheckTx(tx)
-		if err != nil {
-			glog.Error(err)
-			return
+		if t.RewardCut.Cmp(eth.FromPerc(blockRewardCut)) != 0 && t.FeeShare.Cmp(eth.FromPerc(feeShare)) != 0 {
```
yondonfu: We still want to submit the tx if one of commission rates is not equal to its on-chain value.

yondonfu (changes requested):

server/webserver.go
kyriediculous: fixed in 98479bae

yondonfu (commented):
Changes look good. Let's rebase

kyriediculous:
rebased !

server/webserver.go
```diff
@@ -363,32 +363,34 @@ func (s *LivepeerServer) cliWebServerHandlers(bindAddr string) *http.ServeMux {
...
+			}
+
+			err = s.LivepeerNode.Eth.CheckTx(tx)
+			if err != nil {
+				glog.Error(err)
+				return
+			}
 		}
 
+		glog.Infof("Setting orchestrator commission rates %v", s.LivepeerNode.Eth.Account().Address.Hex())
```
yondonfu: Could we actually move this log statement into the above conditional before the tx is submitted? As-is, this log statement would imply that the commission rates are in the process of being set *after* the tx is already submitted and confirmed on-chain.

yondonfu (changes requested):

server/webserver.go
kyriediculous: fixed in bb83638d

yondonfu (approved):
LGTM - good to merge after a rebase

PR #1237, opened 2019-12-02
discovery: use maxInt64 deactivationRound in cacheTranscoderPool

kyriediculous:

When an orchestrator becomes active it's on-chain `deactivationRound` gets set to the maximum value for `uint256`. In go-livepeer when we call `Eth.GetTranscoderPool` the value is returned as a `big.Int`, for inserting it into the database we convert it to an `int64` causing an overflow and `-1` is set as `deactivationRound` as a result, causing active O's to show as inactive (in order to be active the following must be true `currentRound < deactivationRound`). 

- Check for each O if the returned `deactivationRound` from `Eth.GetTranscoders` would overflow when converted to `int64` set the max value for `int64` instead. 

Even though that this might cause issues very very very far into the future when we reach a  round where `2^63 − 1 < currentRound < 2^256` , this shouldn't be a concern now and we don't even know if the EVM will still be using 256 bit words when that time rolls around. 

Ran on-chain transcoding workflow

```
I1203 00:15:08.314105   48383 discovery.go:105] Done fetching orch info for orchestrators, numResponses fetched: 5
I1203 00:15:08.314143   48383 broadcast.go:152] Ending session refresh
I1203 00:15:08.670160   48383 broadcast.go:450] Successfully validated segment nonce=8920814800456177905 seqNo=0
I1203 00:15:09.176480   48383 broadcast.go:450] Successfully validated segment nonce=8920814800456177905 seqNo=1
```

**Checklist:**

### DIFF
--- a/discovery/db_discovery.go
+++ b/discovery/db_discovery.go
@@ -3,6 +3,7 @@ package discovery
 import (
 	"context"
 	"fmt"
+	"math"
 	"math/big"
 	"net/url"
 	"strings"
@@ -19,6 +20,8 @@ import (
 
 	"github.com/golang/glog"
 )
+
+const maxInt64 = int64(math.MaxInt64)
 
 var cacheRefreshInterval = 1 * time.Hour
 var getTicker = func() *time.Ticker {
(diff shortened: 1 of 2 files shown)


discovery/db_discovery.go
```diff
@@ -137,6 +138,10 @@ func (dbo *DBOrchestratorPoolCache) cacheTranscoderPool() error {
 	}
 
 	for _, o := range orchestrators {
+		maxInt64 := big.NewInt(int64(math.MaxInt64))
```
yondonfu: What do you think about moving this logic into `ethOrchToDBOrch()` and creating an instance of `DBOrch` with the correct value for `DeactivationRound` instead of mutating each instance of `lpTypes.Transcoder` in this loop? That way `lpTypes.Transcoder` would always represent the on-chain state and `DBOrch` would represent the on-chain state translated into an off-chain DB compatible state.

Then, we could also add a unit test for `ethOrchToDBOrch()` that confirms that we are correctly setting `DeactivationRound` on `DBOrch` to max int64 whenever `DeactivationRound` on `lpTypes.Transcoder` is equal to max uint256.

yondonfu (changes requested):

discovery/db_discovery.go
kyriediculous: fixed in 44a7cf9 

also using a const for `maxInt64` (as  `int64`) so that it is evaluated at compile time rather than runtime.

yondonfu (commented):

kyriediculous:
rebased !

yondonfu (approved):
LGTM 🚢 pending green CI

PR #1241, opened 2019-12-03
[O selection] server: Use BroadcastSessionsSelector in BroadcastSessionsManager

### LINKED ISSUE #1228: Use BroadcastSessionsSelector interface in BroadcastSessionsManager

`server.BroadcastSessionsManager` currently maintains the state of a session list using an [internal field](https://github.com/livepeer/go-livepeer/blob/master/server/broadcast.go#L51) and always uses a LIFO strategy for selecting the next session.

Additional bookkeeping and updated selection strategy logic will be required in order to support #1215. Rather than, prescribing a single strategy `BroadcastSessionsManager` can support multiple strategies.

`BroadcastSessionsManager` can use the following `BroadcastSessionsSelector` interface:

```
type BroadcastSessionsSelector interface {
    Add(sessions []*BroadcastSession)
    Complete(sess *BroadcastSession)
    Select() *BroadcastSession
    Size() int
}
```

Session list state required for selection will be maintained by the type that implements `BroadcastSessionsSelector`. `BroadcastSessionsManager` can still use the `sessMap` field to keep track track of sessions that are still tracked by `BroadcastSessionsSelector`.

The current LIFO selection strategy can be refactored into a `LIFOSelector` type that implements `BroadcastSessionsSelector`.

We could track `sessMap` in the implementation for `BroadcastSessionsSelector`, but the `sessMap` logic would likely be the same for all implementations of `BroadcastSessionsSelector`. 

We could just remove the LIFO selection strategy instead of refactoring it into `LIFOSelector`. Could be useful to have a code example of an alternate strategy even if it is not being used in practice though.

yondonfu:

This PR moves the current LIFO O selection logic from `BroadcastSessionsManager` to `LIFOSelector` which implements the `BroadcastSessionsSelector` interface. The goals of this refactor are:

- Allow new selection logic to be implemented separately from `BroadcastSessionsManager` in follow on PRs. New selection logic can be encapsulated in a new type that implements the `BroadcastSessionsSelector` interface
- Keep the existing LIFO selection logic in `LIFOSelector` to serve as an example of an alternate selection strategy once the new selection strategy is implemented in follow on PRs

- Update `BroadcastSessionsManager` to accept the `BroadcastSessionsSelector` interface as a dependency
- Add `LIFOSelector` which implements `BroadcastSessionsSelector`

Note that `LIFOSelector` is *not* concurrency safe. The caller of `LIFOSelector` should implement locking i.e. `BroadcastSessionsManager`.

The `BroadcastSessionsSelector` API does not contain a method for removing sessions. The caller of the API should keep track of which sessions still exist. A session returned from `Select()` that no longer exists should be discarded and `Select()` should be called again. This design choice matches the existing design in `BroadcastSessionsManager` where `sessMap` is used to keep track of existing sessions and whenever a non-existent session is returned we re-select another session.

Updated unit tests.

Fixes #1228 

**Checklist:**

### DIFF
--- a/common/db.go
+++ b/common/db.go
@@ -61,6 +61,7 @@ type DBUnbondingLock struct {
 type DBOrchFilter struct {
 	MaxPrice     *big.Rat
 	CurrentRound *big.Int
+	Addresses    []ethcommon.Address
 }
 
 var LivepeerDBVersion = 1
@@ -701,6 +702,14 @@ func buildFilterOrchsQuery(filter *DBOrchFilter) (string, error) {
 		if filter.CurrentRound != nil {
 			currentRound := filter.CurrentRound.Int64()
 			qry += fmt.Sprintf(" AND activationRound <= %v AND %v < deactivationRound", currentRound, currentRound)
+		}
+
+		if len(filter.Addresses) > 0 {
+			hexAddrs := make([]string, len(filter.Addresses))
+			for i, addr := range filter.Addresses {
+				hexAddrs[i] = fmt.Sprintf("'%v'", addr.Hex())
+			}
+			qry += fmt.Sprintf(" AND ethereumAddr IN (%v)", strings.Join(hexAddrs, ", "))
 		}
 	}
 	return qry, nil
--- a/common/db.go
+++ b/common/db.go
@@ -46,7 +46,7 @@ type DBOrch struct {
 	PricePerPixel     int64
 	ActivationRound   int64
 	DeactivationRound int64
-	Stake             string
+	Stake             int64 // Stored as a fixed point number
 }
 
 // DBOrch is the type binding for a row result from the unbondingLocks table
@@ -83,7 +83,7 @@ var schema = `
 		pricePerPixel int64,
 		activationRound int64,
 		deactivationRound int64,
-		stake STRING
+		stake int64
 	);
 
 	CREATE TABLE IF NOT EXISTS unbondingLocks (
@@ -123,7 +123,7 @@ var schema = `
 	CREATE INDEX IF NOT EXISTS idx_blockheaders_number ON blockheaders(number);
 `
 
-func NewDBOrch(ethereumAddr string, serviceURI string, pricePerPixel int64, activationRound int64, deactivationRound int64, stake string) *DBOrch {
+func NewDBOrch(ethereumAddr string, serviceURI string, pricePerPixel int64, activationRound int64, deactivationRound int64, stake int64) *DBOrch {
 	return &DBOrch{
 		ServiceURI:        serviceURI,
 		EthereumAddr:      ethereumAddr,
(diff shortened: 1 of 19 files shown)


yondonfu:
Ready for full review.

kyriediculous:
This overall looks good to me. Do you think it's worth the engineering effort to add unit tests for `LIFOSelector` or not , considering it's not actually going to be used once all PRs get merged.

yondonfu:
> Do you think it's worth the engineering effort to add unit tests for LIFOSelector or not , considering it's not actually going to be used once all PRs get merged.

I'm ok with not adding explicit unit tests for `LIFOSelector` right now for the reason that you mentioned and also because the `BroadcastSessionsManager` unit tests already largely test the behavior of `LIFOSelector` since each instance of `BroadcastSessionsManager` created in those tests use `LIFOSelector` and the tests include assertions on state updates for `LIFOSelector`.

j0sh:
This looks pretty good.

On a somewhat related note re: code organization, how feasible would it be to move the `BroadcastSession` into `core` ? The `broadcast.go` file is pretty empty. Then most of the selection code could live within that as well - either directly within `broadcast.go`, in `core` or even its own pacakge, but that's my least favorite choice. We don't actually have to do this though, and I wonder if it actually makes sense for anything other than subjective taste. Keeping everything in `server` isn't really a problem anyway.

yondonfu:
> The broadcast.go file is pretty empty.

Do you mean `core/broadcaster.go`?

Moving `BroadcastSession` and the related selection code (i.e. `BroadcastSessionsManager`, `BroadcastSessionsSelector`, etc.) to `core` seems feasible, but one small downside I see is that if we also move the `NewSessionsManager()` constructor to `core`, then `core` would need to import `server` since `NewSessionsManager()` accepts `streamParameters` as a param which is defined in `server`.

> Keeping everything in server isn't really a problem anyway.

I'm ok with keeping everything in `server` right now.

kyriediculous (approved):

j0sh:
>  Do you mean core/broadcaster.go?

Ah yep, that.

> core would need to import server since NewSessionsManager() accepts `streamParameters` as a param which is defined in server.

`streamParameters` could also be defined within `core/broadcaster.go` as well - that seems OK, although all the fields would need to be made public. Which is also OK.

The only thing I really wish we had was a way to indicate that a struct such as `streamParameters` was strictly read-only, immutable product type without any attached methods. We have a few other data structures following this pattern, such as the `BroadcastSession` struct.

> I'm ok with keeping everything in server right now.

Yeah, that's fine. Just want to keep that possibility in mind the next time someone complains about the length / scope of the files in `server`, and to make us a little more conscious of the issue so we aren't adding things later on which make this type of extraction more difficult. (Another implicit design outcome here is the fact `server/broadcast.go` has no LivepeerServer methods or instance methods outside of the BroadcastSessionsManager.)

I think this also exhibits part of the tension around code placement and package structure - which is pretty arbitrary once you get down to it. Put everything in a single file! (Kidding but only a little bit. Personally, I don't really have an issue with files that are thousands of lines long.)

Anyway all  that's getting off topic for this PR. LGTM!

Issue #1091, opened 2019-09-16
Bind transcoding options with transcoded results in orchestrator signature


yondonfu:
**Is your feature request related to a problem? Please describe.**

At the moment, an orchestrator in on-chain mode will [sign the following payload](https://github.com/livepeer/go-livepeer/blob/master/core/orchestrator.go#L546):

```
H(H(res_1) + H(res_2) + ... + H(res_n))
```

where `res_1, res_2, ..., res_n` is the set of transcoded results for a source segment. The signature over this hash can be used by the broadcaster to authenticate the set of transcoded results received from an orchestrator. However, the signature does not bind a specific set of transcoding options with each result. 

**Describe the solution you'd like**

Bind a set of transcoding options with each result with the orchestrator's signature. We can change the payload signed by the orchestrator to:

```
H(H(res_1 + t_opts_1) + H(res_2 + t_opts_2) + ... + H(res_n + t_opts_n))
```

where `t_opts_1, t_opts_2, ..., t_opts_n` is the set of transcoding options associated with the set of transcoded results where `t_opts_i` corresponds to `res_i`.

Each `t_opt_i` can be the [4 byte identifier](https://github.com/livepeer/go-livepeer/blob/master/common/videoprofile_ids.go) associated with the video profile associated with the result. 

This will require:

- [ ] Updating the orchestrator signing logic
- [ ] Updating the broadcaster's verification logic

j0sh:
> Each t_opt_i can be the 4 byte identifier

I'm unsure how those 4 byte identifiers were generated; is it the first bytes of some hashed construction of the VideoProfile?

At some point we'll need to expand beyond our current fixed set of video profiles, and into arbitrary resolutions (at a minimum).

If we standardize on how to incorporate the [VideoProfile parameters](https://github.com/livepeer/lpms/blob/0246c39344c63ce2f2cd0bbc17d6704d35c998bc/ffmpeg/videoprofile.go#L20-L26) into the signature, then that can be used.

I am a little nervous about forward compatibility because it's possible (likely) that we'll add encoding options that aren't currently reflected in the VideoProfile, and a naive application of this scheme would force B/O/V to upgrade in lockstep. I can see arguments for/against a looser approach towards compatibility; let me think about it a little more.

yondonfu:
> I'm unsure how those 4 byte identifiers were generated; is it the first bytes of some hashed construction of the VideoProfile?

[...]

PR #1242, opened 2019-12-04
Ar/handle custom bitrate ladder

angyangie:
https://github.com/livepeer/go-livepeer/issues/811

- Adding `FullProfiles` to `net.SegData`
- Adding `FullProfiles` to `net.NotifySegment`
- Adding tests and necessary helper methods

Unit tests

https://github.com/livepeer/go-livepeer/issues/811

**Checklist:**


server/mediaserver.go
```diff
@@ -187,7 +194,23 @@ func createRTMPStreamIDHandler(s *LivepeerServer) func(url *url.URL) (strmID str
 			mid, key = parseManifestID(resp.ManifestID), resp.StreamKey
 			// Process transcoding options presets
 			if len(resp.Presets) > 0 {
-				presets = parsePresets(resp.Presets)
+				profiles = parsePresets(resp.Presets)
+			}
+
+			if len(resp.Profiles) > 0 {
+				for _, profile := range resp.Profiles {
+					prof := ffmpeg.VideoProfile{
```
j0sh: Maybe if `profile.Name` is empty we can generate a name, eg `fmt.Sprintf(%dx%d_%d", width, height, bitrate)` . Had this issue when testing and forgot to add a name. Would be good to that behavior to other helpers as well, including `makeFFmpegVideoProfiles` and `FFmpegProfiletoNetProfile` . (And tests for those as well.)

angyangie: good catch, I'll add that + tests!

common/util.go
```diff
@@ -128,9 +130,34 @@ func BytesToVideoProfile(txData []byte) ([]ffmpeg.VideoProfile, error) {
...
+		if err != nil {
+			return nil, err
+		}
+		br := strings.Replace(profile.Bitrate, "k", "", 1)
+		bitrate, err := strconv.Atoi(br)
+		if err != nil {
+			return nil, err
+		}
+		if profile.Name == "" {
+			profile.Name = fmt.Sprintf("%dx%d_%d", width, height, bitrate)
```
j0sh: Should the `fmt.Sprintf` part be factored out into a helper for reuse? That would allow us to catch all instances where we need to change the format for whatever reason. Also I'm a little nervous about mutating the profile itself when it's not strictly necessary. Can an intermediate `name` variable be used instead?

One other trick that might come in handy is to also prefix (or append) the generated name with *where* it came from, so we know at what point the name was missing. Something like this:

```
func defaultProfileName(w, h, b) {
  return fmt.Sprintf(...)
}

name = profile.Name
if name == "" {
 // set default profile name
}

// in createRTMPStreamIDHandler
name = "webhook_"+defaultProfileName(...)

// in FFmpegProfiletoNetProfile
name = "ffmpeg_"+defaultProfileName(...)

// in makeFfmpegVideoProfiles
name = "net_"+defaultProfileName(...)
```

angyangie: How can "mutating the profile" cause issues down the line? Cool, I'll add the prefix!

common/util.go
```diff
@@ -128,9 +130,39 @@ func BytesToVideoProfile(txData []byte) ([]ffmpeg.VideoProfile, error) {
...
+			Fps:     int32(profile.Framerate),
+		}
+		profiles = append(profiles, &fullProfile)
+	}
+	return profiles, nil
+}
+
 func ProfilesToTranscodeOpts(profiles []ffmpeg.VideoProfile) []byte {
-	//Sort profiles first
-	sort.Sort(ffmpeg.ByName(profiles))
```
j0sh, line 133: @yondonfu Note this change - it was leading to some unexpected side effects during our testing. This change was a quick workaround for that, but it does affect how sigs are generated. (See this commit: https://github.com/livepeer/go-livepeer/pull/1242/commits/6ebfbe54363020b1414c9f4bf1e1cd87a3dc5431) Figured it'd be OK since we're reworking the sigs with https://github.com/livepeer/go-livepeer/issues/1091 anyway, but wanted to check in with you first.

server/segment_rpc.go
```diff
@@ -210,6 +210,24 @@ func getPaymentSender(payment net.Payment) ethcommon.Address {
 	return ethcommon.BytesToAddress(payment.Sender)
 }
 
+func makeFfmpegVideoProfiles(protoProfiles []*net.VideoProfile) []ffmpeg.VideoProfile {
+	profiles := []ffmpeg.VideoProfile{}
```
darkdarkdragon: Better to do
`profiles := make([]ffmpeg.VideoProfile, 0, len(protoProfiles))`

common/util.go
```diff
@@ -128,9 +130,39 @@ func BytesToVideoProfile(txData []byte) ([]ffmpeg.VideoProfile, error) {
 	return profiles, nil
 }
 
+func DefaultProfileName(width int, height int, bitrate int) string {
+	return fmt.Sprintf("%dx%d_%d", width, height, bitrate)
+}
+
+func FFmpegProfiletoNetProfile(ffmpegProfiles []ffmpeg.VideoProfile) ([]*net.VideoProfile, error) {
+	profiles := []*net.VideoProfile{}
```
darkdarkdragon: will not make much difference here, but still it's a good practice to do
`profiles := make([]*net.VideoProfile, 0, len(ffmpegProfiles))`

server/mediaserver.go
```diff
@@ -187,7 +194,30 @@ func createRTMPStreamIDHandler(s *LivepeerServer) func(url *url.URL) (strmID str
 			mid, key = parseManifestID(resp.ManifestID), resp.StreamKey
 			// Process transcoding options presets
 			if len(resp.Presets) > 0 {
-				presets = parsePresets(resp.Presets)
+				profiles = parsePresets(resp.Presets)
+			}
+
+			if len(resp.Profiles) > 0 {
```
darkdarkdragon: this check not really neeeded

server/mediaserver.go
```diff
@@ -187,7 +194,30 @@ func createRTMPStreamIDHandler(s *LivepeerServer) func(url *url.URL) (strmID str
...
+				for _, profile := range resp.Profiles {
+					name := profile.Name
+					if name == "" {
+						profile.Name = "webhook_" + common.DefaultProfileName(
+							int(profile.Width),
```
darkdarkdragon: cast not needed

server/mediaserver.go
```diff
@@ -187,7 +194,30 @@ func createRTMPStreamIDHandler(s *LivepeerServer) func(url *url.URL) (strmID str
...
+							int(profile.Height),
```
darkdarkdragon: cast not needed

server/mediaserver.go
```diff
@@ -187,7 +194,30 @@ func createRTMPStreamIDHandler(s *LivepeerServer) func(url *url.URL) (strmID str
...
+							int(profile.Bitrate))
```
darkdarkdragon: cast not needed

server/mediaserver.go
```diff
@@ -187,7 +194,30 @@ func createRTMPStreamIDHandler(s *LivepeerServer) func(url *url.URL) (strmID str
...
+					}
+					prof := ffmpeg.VideoProfile{
+						Name:       profile.Name,
+						Bitrate:    fmt.Sprintf("%dk", profile.Bitrate),
+						Framerate:  uint(profile.FPS),
```
darkdarkdragon: probably we should make it to be `uint` in `authWebhookResponse`?

server/mediaserver.go
```diff
@@ -187,7 +194,30 @@ func createRTMPStreamIDHandler(s *LivepeerServer) func(url *url.URL) (strmID str
...
+						Resolution: fmt.Sprintf("%dx%d", profile.Width, profile.Height),
+					}
+					profiles = append(profiles, prof)
+				}
+			}
+
+			if len(resp.Profiles) <= 0 && len(resp.Presets) <= 0 {
```
darkdarkdragon, line 218: probably we should check `len(profiles)` here? `parsePresets` can skip profile, so len(profiles) can be zero even if len(resp.Presets) is not zero

darkdarkdragon (changes requested):

server/mediaserver.go
```diff
@@ -187,7 +194,28 @@ func createRTMPStreamIDHandler(s *LivepeerServer) func(url *url.URL) (strmID str
...
 			// Process transcoding options presets
 			if len(resp.Presets) > 0 {
-				presets = parsePresets(resp.Presets)
+				profiles = parsePresets(resp.Presets)
+			}
+
+			for _, profile := range resp.Profiles {
+				name := profile.Name
+				if name == "" {
+					profile.Name = "webhook_" + common.DefaultProfileName(
```
j0sh: `name =` rather than `profile.Name =` - we want to avoid the mutation

server/mediaserver.go
```diff
@@ -187,7 +194,28 @@ func createRTMPStreamIDHandler(s *LivepeerServer) func(url *url.URL) (strmID str
...
+						profile.Width,
+						profile.Height,
+						profile.Bitrate)
+				}
+				prof := ffmpeg.VideoProfile{
+					Name:       profile.Name,
```
j0sh: `name` rather than `profile.Name`

server/mediaserver.go
j0sh: Assign to `name` rather than `profile.Name`

server/mediaserver.go
```diff
@@ -187,7 +194,30 @@ func createRTMPStreamIDHandler(s *LivepeerServer) func(url *url.URL) (strmID str
...
+				for _, profile := range resp.Profiles {
+					name := profile.Name
+					if name == "" {
+						profile.Name = "webhook_" + common.DefaultProfileName(
+							int(profile.Width),
+							int(profile.Height),
+							int(profile.Bitrate))
+					}
+					prof := ffmpeg.VideoProfile{
+						Name:       profile.Name,
```
j0sh: `name` rather than `profile.Name`

server/mediaserver.go
j0sh: Actually this was intentional. We only want to set defaults if the user did *not* specify a preset/profile. If the parsing of the preset/profile failed, then setting the default instead is surprising and not what the user intended. In that case, it seems better to continue with an empty set of profiles (so no transcoding happens until the user fixes their configuration). We probably should add a comment here explaining, "Only set defaults if user did not specify a preset/profile"

common/util.go
yondonfu: Thanks for the heads up - think it should be ok since we'll be reworking things in #1091

common/util.go
j0sh: > How can "mutating the profile" cause issues down the line

In this specific instance it is probably mostly OK since the struct being mutated isn't used anywhere else, but in general it's much safer to avoid side effects where possible. Issues include concurrency, reasoning about effects outside the scope of the function, etc.

So I'd keep the entire struct immutable here as good practice, if only to future-proof things. Of course, this only works if future maintainers also uphold the same discipline. For a language of its age, I'm really surprised Golang went with mutable-by-default variables.

server/mediaserver.go
angyangie: That makes sense @j0sh ... I agree that default should only be set if the user does not profile profiles/presets. I'll add a comment to explain

common/util.go
angyangie: Thanks for the explanation! I do remember having issues with that at one point

go.mod
```diff
@@ -1,70 +1,23 @@
 module github.com/livepeer/go-livepeer
 
-go 1.13
+go 1.12
```
j0sh: Is it confirmed that this works on go 1.12 again?

go.mod
```diff
@@ -1,70 +1,23 @@
...
-	github.com/docker/docker v1.13.1 // indirect
-	github.com/edsrzf/mmap-go v1.0.0 // indirect
-	github.com/elastic/gosigar v0.10.5 // indirect
-	github.com/ethereum/go-ethereum v1.9.3
-	github.com/fatih/color v1.7.0 // indirect
-	// replace example.com/some/dependency => example.com/some/dependency-fork v1.2.3
-	github.com/fjl/memsize v0.0.0-20190710130421-bcb5799ab5e5 // indirect
-	github.com/gballet/go-libpcsclite v0.0.0-20190607065134-2772fd86a8ff // indirect
-	// github.com/gballet/go-libpcsclite v0.0.0-20190403181518-312b5175032f // indirect
+	github.com/ethereum/go-ethereum v1.9.9
```
j0sh: I'm nervous about this change; would prefer not to update dependencies if we can avoid it, especially things that we are heavily dependent on such as `go-ethereum`

j0sh:
Did not have an opportunity to closely look at the tests but the changes themselves look good other than the go mod stuff - can that be reset? Works nicely on my machine otherwise, tested with B/OT and B/O/T.

j0sh:
Once the `go.mod` stuff is sorted out, might be good to update LPMS with frame rate passthrough via `go get github.com/livepeer/lpms` . The diff should only show one line changed, something like 

```diff
diff --git a/go.mod b/go.mod
index 99ea115d..bf6c76cf 100644
--- a/go.mod
+++ b/go.mod
@@ -9,7 +9,7 @@ require (
 	github.com/ethereum/go-ethereum v1.9.9
 	github.com/golang/glog v0.0.0-20160126235308-23def4e6c14b
 	github.com/golang/protobuf v1.3.2
-	github.com/livepeer/lpms v0.0.0-20191121223052-fdbf27c7cfbf
+	github.com/livepeer/lpms v0.0.0-20191206210009-e401b07f1fbe
 	github.com/livepeer/m3u8 v0.11.0
 	github.com/mattn/go-sqlite3 v2.0.1+incompatible
 	github.com/olekukonko/tablewriter v0.0.4
```

server/segment_rpc.go
```diff
@@ -210,6 +210,24 @@ func getPaymentSender(payment net.Payment) ethcommon.Address {
...
+	profiles := make([]ffmpeg.VideoProfile, 0, len(protoProfiles))
+	for _, profile := range protoProfiles {
+		name := profile.Name
+		if name == "" {
+			name = "net_" + common.DefaultProfileName(int(profile.Width), int(profile.Height), int(profile.Bitrate))
+		}
+		prof := ffmpeg.VideoProfile{
+			Name:       name,
+			Bitrate:    fmt.Sprintf("%dk", profile.Bitrate),
+			Framerate:  uint(profile.Fps),
```
darkdarkdragon, line 223: I think we should change `profile.Fps` to `uint32` (right now passing -1 in `profile.Fps` will lead `Framerate` to be equal 4294967295).

darkdarkdragon (changes requested):

server/segment_rpc.go
angyangie: Good catch! That's something that would need to also be changed in LPMS. Do you mind if we make a separate ticket for this and tackle it in a separate PR?

angyangie: https://github.com/livepeer/lpms/issues/162

darkdarkdragon (approved):
LGTM 🚢

server/mediaserver.go
```diff
@@ -207,7 +207,7 @@ func createRTMPStreamIDHandler(s *LivepeerServer) func(url *url.URL) (strmID str
 				}
 				prof := ffmpeg.VideoProfile{
 					Name:       name,
-					Bitrate:    fmt.Sprintf("%dk", profile.Bitrate),
+					Bitrate:    fmt.Sprint(profile.Bitrate),
```
j0sh, line 210: The `fmt.Sprint` is unnecessary here (and a number of other places)

angyangie: `profile.Bitrate` is either an `int32` or `int` in various places, but needs to be a string. Seems like the best way of converting consistently!

PR #1243, opened 2019-12-04
cmd, eth: fix floating point math for eth<=>wei conversions

### LINKED ISSUE #1211: livepeer_cli taking more than requested for reserve when calling fundDepositAndReserve
**Describe the bug**

When depositing Broadcasting funds via the livepeer_cli, it seems to take more for the reserve than was requested.

When invoking `12. Invoke "deposit broadcasting funds" (ETH)`, I asked to deposit `0.5 ETH` for deposit and `0.1 ETH` for reserve.

The amount taken for reserve was `0.100000000000000005 ETH`

**To Reproduce**

Steps to reproduce:

1. Run a node in Broadcaster mode
2. Run livepeer_cli
3. Choose the command: `12. Invoke "deposit broadcasting funds" (ETH)`
4. Enter `0.5` for deposit and `0.1` for reserve and wait for transaction to confirm
5. Choose the command: `1. Get node status`
6. See that the amount taken for reserve was `0.100000000000000005 ETH`

**Expected behavior**

I expect it to take exactly the same amount that I instructed it to take.

**Screenshots**

Amount requested for deposit / reserve

Screenshot from 2019-11-22 21-19-08

Amounts actually requested by the process for deposit and reserve:

Screenshot from 2019-11-22 21-19-01

Transaction called by the node (incidentally, the values are not logged in the console - perhaps they should / could be?)

Screenshot from 2019-11-22 21-20-11

Amount shown on Etherscan:

Screenshot from 2019-11-22 21-31-51

And amount requested for reserve, in hexadecimal

Screenshot from 2019-11-22 21-24-29

And amount requested for reserve converted from hex to decimal... showing that the process was calling the function with the incorrect input.

Screenshot from 2019-11-22 21-24-41

 - OS: Ubuntu 19.04
```
chrishobcroft@thinkpad:~$ ./livepeer -version
Livepeer Node Version: 0.5.0
Golang runtime version: gc go1.13
Architecture: amd64
Operating system: linux
```

kyriediculous:

This PR attempts to fix floating point math rounding errors caused when depositing broadcasting funds with the CLI. To fix this we are no longer using floating point math but arbitrary-precision fixed-point decimal numbers instead (up to 2^31 decimals). 

- removed `eth.FromBaseUnit` and `eth.ToBaseUnit`
- added `github.com/shopspring/decimal` third-party library 
- introduce `eth.WeiToEth(interface{}) decimal.Decimal` and `eth.EthToWei(interface{}) *big.Int`

- Added unit tests with different tests cases and ran unit tests 
- Ran #1211 reproduction steps and see that when adding `0.1 ETH` to reserve it no longer results in `0.100000000000000005 ETH` being taken.

Fixes #1211

**Checklist:**

### DIFF
--- a/cmd/livepeer_cli/wizard.go
+++ b/cmd/livepeer_cli/wizard.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/ethereum/go-ethereum/log"
 	lpcommon "github.com/livepeer/go-livepeer/common"
+	"github.com/livepeer/go-livepeer/eth"
 )
 
 // // read reads a single line from stdin, trimming if from spaces.
@@ -83,6 +84,32 @@ func (w *wizard) readDefaultString(def string) string {
 		return text
 	}
 	return def
+}
+
+func (w *wizard) readBaseAmountAndValidate(validate func(in *big.Int) error) *big.Int {
+	for {
+		text := w.readString()
+		val, err := eth.ToBaseAmount(text)
+		if err != nil {
+			log.Error("Error parsing user input", "err", err)
+			continue
+		}
+		if err := validate(val); err != nil {
+			log.Error("Invalid user input", "err", err)
+			continue
+		}
+		return val
+	}
+}
+
+func (w *wizard) readPositiveBaseAmount() *big.Int {
+	return w.readBaseAmountAndValidate(func(in *big.Int) error {
+		if in.Cmp(big.NewInt(0)) < 0 {
+			return errors.New("base amount must be positive")
+		}
+
+		return nil
+	})
 }
 
 // readInt reads a single line from stdin, trimming if from spaces, enforcing it
--- a/cmd/livepeer_cli/wizard_ticketbroker.go
+++ b/cmd/livepeer_cli/wizard_ticketbroker.go
@@ -46,20 +46,18 @@ func (w *wizard) deposit() {
 		return
 	}
 
-	fmt.Printf("Current Deposit: %v\n", sender.Deposit)
-	fmt.Printf("Current Reserve: %v\n", sender.Reserve.FundsRemaining)
+	fmt.Printf("Current Deposit: %v\n", eth.FormatUnits(sender.Deposit, "ETH"))
+	fmt.Printf("Current Reserve: %v\n", eth.FormatUnits(sender.Reserve.FundsRemaining, "ETH"))
 
 	fmt.Printf("Enter deposit amount in ETH - ")
-
-	depositAmount := w.readPositiveFloat()
+	depositAmount := w.readPositiveBaseAmount()
 
 	fmt.Printf("Enter reserve amount in ETH - ")
-
-	reserveAmount := w.readPositiveFloat()
+	reserveAmount := w.readPositiveBaseAmount()
 
 	form := url.Values{
-		"depositAmount": {eth.ToBaseAmount(depositAmount).String()},
-		"reserveAmount": {eth.ToBaseAmount(reserveAmount).String()},
+		"depositAmount": {depositAmount.String()},
+		"reserveAmount": {reserveAmount.String()},
 	}
 	fmt.Println(httpPostWithParams(fmt.Sprintf("http://%v:%v/fundDepositAndReserve", w.host, w.httpPort), form))
 
--- a/cmd/livepeer_cli/wizard_ticketbroker.go
+++ b/cmd/livepeer_cli/wizard_ticketbroker.go
@@ -58,8 +58,8 @@ func (w *wizard) deposit() {
 	reserveAmount := w.readPositiveFloat()
 
 	form := url.Values{
-		"depositAmount": {eth.ToBaseUnit(big.NewFloat(depositAmount)).String()},
-		"reserveAmount": {eth.ToBaseUnit(big.NewFloat(reserveAmount)).String()},
+		"depositAmount": {eth.ToBaseAmount(depositAmount).String()},
+		"reserveAmount": {eth.ToBaseAmount(reserveAmount).String()},
 	}
 	fmt.Println(httpPostWithParams(fmt.Sprintf("http://%v:%v/fundDepositAndReserve", w.host, w.httpPort), form))
 
--- a/eth/helpers.go
+++ b/eth/helpers.go
@@ -3,12 +3,16 @@ package eth
 import (
 	"fmt"
 	"math/big"
+	"regexp"
+	"strings"
 	"time"
 
 	ethcommon "github.com/ethereum/go-ethereum/common"
 	"github.com/golang/glog"
 	"github.com/livepeer/go-livepeer/common"
 )
+
+const maxDecimals = 18
 
 var (
 	BlocksUntilFirstClaimDeadline = big.NewInt(200)
@@ -19,9 +23,8 @@ func FormatUnits(baseAmount *big.Int, name string) string {
 	if baseAmount == nil {
 		return "0 " + name
 	}
-	amount := FromBaseUnit(baseAmount)
 
-	if amount.Cmp(big.NewFloat(0.01)) == -1 {
+	if baseAmount.Cmp(big.NewInt(10000000000000000)) == -1 {
 		switch name {
 		case "ETH":
 			return fmt.Sprintf("%v WEI", baseAmount)
@@ -31,30 +34,11 @@ func FormatUnits(baseAmount *big.Int, name string) string {
 	} else {
 		switch name {
 		case "ETH":
-			return fmt.Sprintf("%v ETH", amount)
+			return fmt.Sprintf("%v ETH", FromBaseAmount(baseAmount))
 		default:
-			return fmt.Sprintf("%v LPT", amount)
+			return fmt.Sprintf("%v LPT", FromBaseAmount(baseAmount))
 		}
 	}
-}
-
-func ToBaseUnit(lptAmount *big.Float) *big.Int {
-	decimals := new(big.Int).Exp(big.NewInt(10), big.NewInt(18), nil)
-	floatDecimals := new(big.Float).SetInt(decimals)
-	floatBaseAmount := new(big.Float).Mul(lptAmount, floatDecimals)
-
-	baseAmount := new(big.Int)
-	floatBaseAmount.Int(baseAmount)
-
-	return baseAmount
-}
-
-func FromBaseUnit(baseAmount *big.Int) *big.Float {
-	decimals := new(big.Int).Exp(big.NewInt(10), big.NewInt(18), nil)
-	floatDecimals := new(big.Float).SetInt(decimals)
-	floatBaseAmount := new(big.Float).SetInt(baseAmount)
-
-	return new(big.Float).Quo(floatBaseAmount, floatDecimals)
 }
 
 func FormatPerc(value *big.Int) string {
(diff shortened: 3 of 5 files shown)


cmd/livepeer_cli/wizard_ticketbroker.go
```diff
@@ -58,8 +58,8 @@ func (w *wizard) deposit() {
 	reserveAmount := w.readPositiveFloat()
 
 	form := url.Values{
-		"depositAmount": {eth.ToBaseUnit(big.NewFloat(depositAmount)).String()},
-		"reserveAmount": {eth.ToBaseUnit(big.NewFloat(reserveAmount)).String()},
+		"depositAmount": {eth.EthToWei(depositAmount).String()},
```
yondonfu: Instead of converting the user's string input into a float can we skip the float conversion and just use the string's integer and decimal components to convert into a `big.Int` which validity checks can be applied against?

Something like:

```
// in -> user string input
parts := strings.Split(in, ".")
// parts[0] -> integer component
// parts[1] -> decimal component

if len(parts[1]) > 18 {
    // Cannot have more than 18 decimal places
}

intPart := parts[0]
// Right pad parts[1] with zeros so that it is of length 18
decPart := ...

val, ok := new(big.Int).SetString(intPart + decPart)
// Check that conversion worked

// val -> string input represented in base units

// val.String() can be used to return a string when creating a request
```

Here is an [example](https://github.com/filecoin-project/go-filecoin/blob/b926dad7ba3d5a6342d73008d465c2a56a37d04e/internal/pkg/types/atto_fil.go#L87) from another codebase. 

We could update the `readPositiveFloat()` helper to return a `big.Int` using the above approach. Might want to update the name of the relevant methods to something more accurate like `readTokenUnitString` and `readPositiveTokenUnitString()`.

yondonfu (changes requested):

cmd/livepeer_cli/wizard_ticketbroker.go
kyriediculous: I would prefer to not introduce a third party library such as is now unfortunately, therefore I like an alternative that just uses the standard library as proposed. Just wondering if depending on a string.Split() here is like a robust enough solution, seems so simplistic (not saying that it's not good).

kyriediculous: There is one thing I don't like though but I don't think it's that important _for now_ as the only usage is comparing the returned value to 0.01 which will always be greater than the point where we start rounding. 

When converting a wei string to eth we will eventually end up converting to a `*big.Float` most likely, whether we return the value as a string or let the the caller handle conversion. This will always end up resulting in rounding errors as this failing test shows.

```
        	Error Trace:	helpers_test.go:78
        	Error:      	Not equal: 
        	            	expected: "0.003333333333"
        	            	actual  : "0.003333333333333333"
```

yondonfu: Not sure I understand. I'm suggesting that we convert all user string input directly into `big.Int` - we would never use floats at all. If the user input is in ETH, then the string input would be parsed and represented in wei using a `big.Int`.

kyriediculous: Right but in `FormatUnits` we convert a wei string to an eth value as `big.Float` (which is the other way around of what's happening with the CLI input for deposits and reserves). 

Converting to a big float there is okay for now because we don't need more precision than 2 decimals, as this is the only place where `FromBaseUnit` or in this PR `WeiToEth` is used.

```
func FormatUnits(baseAmount *big.Int, name string) string {
	if baseAmount == nil {
		return "0 " + name
	}
	
	amount := FromBaseUnit(baseAmount)

	if amount.Cmp(big.NewFloat(0.01)) == -1 {
		switch name {
		case "ETH":
			return fmt.Sprintf("%v WEI", baseAmount)
		default:
			return fmt.Sprintf("%v LPTU", baseAmount)
		}
	} else {
		switch name {
		case "ETH":
			return fmt.Sprintf("%v ETH", amount)
		default:
			return fmt.Sprintf("%v LPT", amount)
		}
	}
}
```

yondonfu: Oh I see what you're saying now.

I think we could do something like the following:

- Update `ToBaseUnit(*big.Float) *big.Int` to `ToBaseUnit(string) *big.Int`
- Update `FromBaseUnit(*big.Int) *big.Float` to `FromBaseUnit(*big.Int) (string, bool)` (the extra `bool` return value indicates whether the string conversion from `big.Int` was successful)
- Update `FormatUnits()` to compare `baseAmount` against a `big.Int` instead of converting `baseAmount` to `big.Float` and comparing it against a `big.Float`

`ToBaseUnit()` could be used by the CLI to handle user inputs denominated in ETH.

kyriediculous: submitted several fixup commits

cmd/livepeer_cli/wizard_ticketbroker.go
```diff
@@ -46,20 +47,41 @@ func (w *wizard) deposit() {
...
-	fmt.Printf("Current Deposit: %v\n", sender.Deposit)
-	fmt.Printf("Current Reserve: %v\n", sender.Reserve.FundsRemaining)
+	fmt.Printf("Current Deposit: %v\n", eth.FormatUnits(sender.Deposit, "ETH"))
+	fmt.Printf("Current Reserve: %v\n", eth.FormatUnits(sender.Reserve.FundsRemaining, "ETH"))
 
 	fmt.Printf("Enter deposit amount in ETH - ")
-
-	depositAmount := w.readPositiveFloat()
+	fmt.Printf("> ")
+	depositAmount, err := w.in.ReadString('\n')
```
yondonfu: Prior to this PR, the `readPositiveFloat()` would keep prompting the user for input if the user provided an invalid input (i.e. negative number or non-number). What do you think about maintaining that behavior but instead of reading a float we would be reading a `big.Int` that represents a "base amount". The helpers would look something like:

```
func (w *wizard) readBaseAmountAndValidate(validate func(in *big.Int) error) *big.Int {
	for {
		fmt.Printf("> ")
		text, err := w.in.ReadString('\n')
		if err != nil {
			log.Crit("Failed to read user input", "err", err)
		}
		if text = strings.TrimSpace(text); text == "" {
			continue
		}
		val, err := eth.ToBaseAmount(text)
		if err != nil {
			log.Error("Error parsing user input", "err", err)
			continue
		}
		if err := validate(val); err != nil {
			log.Error("Invalid user input", "err", err)
			continue
		}
		return val
	}
}

func (w *wizard) readPositiveBaseAmount() *big.Int {
	return w.readBaseAmountAndValidate(func(in *big.Int) error {
		if in.Cmp(big.NewInt(0)) < 0 {
			return errors.New("base amount must be positive")
		}

		return nil
	})
}
```

and the code here would be simplified to

```
// Prompt for deposit
// ...
depositAmount := w.readPositiveBaseAmount()

// Prompt for reserve
// ...
reserveAmount := w.readPositiveBaseAmount()
```

eth/helpers.go
```diff
@@ -121,3 +103,66 @@ func fromPerc(perc float64, multiplier *big.Float) *big.Int {
...
+	}
+
+	// If decimal part is shorter than 18 digits, padd with "0"
+	// The decimal is right padded with 0's if it less than 18 digits long
+	for len(decPart) < 18 {
+		decPart += "0"
+	}
+
+	// Combine intPart and decPart into a big.Int
+	wei, ok := new(big.Int).SetString(intPart+decPart, 10)
```
yondonfu: Let's call this something like `baseAmount` since this helper could be used for LPT in addition to ETH conversions.

eth/helpers.go
```diff
@@ -121,3 +103,66 @@ func fromPerc(perc float64, multiplier *big.Float) *big.Int {
...
+	// If decimal part is longer than 18 decimals, trim the string and round off
+	if len(decPart) > 18 {
+		last := int(decPart[19])
+		if last >= 5 {
+			new := decPart[:18] + string(int(decPart[18])+1)
+			decPart = strings.Replace(decPart, decPart, new, 17)
+		}
+	}
+
+	// If decimal part is shorter than 18 digits, padd with "0"
```
yondonfu: "padd" -> "pad"

eth/helpers.go
```diff
@@ -121,3 +103,66 @@ func fromPerc(perc float64, multiplier *big.Float) *big.Int {
...
+	// The decimal is right padded with 0's if it less than 18 digits long
```
yondonfu: This line doesn't seem necessary given the above line.

eth/helpers.go
```diff
@@ -121,3 +103,66 @@ func fromPerc(perc float64, multiplier *big.Float) *big.Int {
...
+	// If '.' is absent from string, add an empty string to become the decimal part
+	if len(splitNum) == 1 {
+		splitNum = append(splitNum, "")
+	}
+	intPart := splitNum[0]
+	decPart := splitNum[1]
+
+	// If decimal part is longer than 18 decimals, trim the string and round off
+	if len(decPart) > 18 {
+		last := int(decPart[19])
```
yondonfu: A value with greater than 18 decimal places would be valid so I think we can just return an error. Should add a test that checks for this.

go.mod
```diff
@@ -47,6 +47,7 @@ require (
 	github.com/rjeczalik/notify v0.9.2 // indirect
 	github.com/robertkrimen/otto v0.0.0-20180617131154-15f95af6e78d // indirect
 	github.com/rs/cors v1.7.0 // indirect
+	github.com/shopspring/decimal v0.0.0-20191130220710-360f2bc03045
```
yondonfu: Let's revert this change.

go.sum
```diff
@@ -196,6 +192,8 @@ github.com/rs/cors v1.7.0 h1:+88SsELBHx5r+hZ8TCkggzSstaWNbDvThkVK8H6f9ik=
 github.com/rs/cors v1.7.0/go.mod h1:gFx+x8UowdsKA9AchylcLynDq+nNFfI8FkUZdN/jGCU=
 github.com/russross/blackfriday/v2 v2.0.1 h1:lPqVAte+HuHNfhJ/0LC98ESWRz8afy9tM/0RK8m9o+Q=
 github.com/russross/blackfriday/v2 v2.0.1/go.mod h1:+Rmxgy9KzJVeS9/2gXHxylqXiyQDYRxCVz55jmeOWTM=
+github.com/shopspring/decimal v0.0.0-20191130220710-360f2bc03045 h1:8CnFGhoe92Izugjok8nZEGYCNovJwdRFYwrEiLtG6ZQ=
```
yondonfu: Let's revert this change.

eth/helpers.go
```diff
@@ -121,3 +103,66 @@ func fromPerc(perc float64, multiplier *big.Float) *big.Int {
...
+	if len(intPart) == 0 {
+		intPart = "0"
+	}
+
+	// remove right padded 0's from decPart
+	decPart = strings.TrimRightFunc(decPart, func(r rune) bool {
+		return r == '0'
+	})
+
+	return intPart + "." + decPart
```
yondonfu, line 167: If `len(decPart) = 0` after right trimming all 0s then we don't need to include the "."  Should add a test that checks for this.

eth/helpers.go
```diff
@@ -121,3 +103,66 @@ func fromPerc(perc float64, multiplier *big.Float) *big.Int {
...
+	intPart := weiString[:len(weiString)-18]
+	decPart := weiString[len(weiString)-18:]
+
+	// if no int part, add "0"
+	if len(intPart) == 0 {
+		intPart = "0"
+	}
+
+	// remove right padded 0's from decPart
+	decPart = strings.TrimRightFunc(decPart, func(r rune) bool {
```
yondonfu: Could we just do `strings.TrimRight(decPart, "0")`?

eth/helpers.go
```diff
@@ -121,3 +103,66 @@ func fromPerc(perc float64, multiplier *big.Float) *big.Int {
...
+	splitNum := strings.Split(v, ".")
+	// If '.' is absent from string, add an empty string to become the decimal part
+	if len(splitNum) == 1 {
+		splitNum = append(splitNum, "")
+	}
+	intPart := splitNum[0]
+	decPart := splitNum[1]
+
+	// If decimal part is longer than 18 decimals, trim the string and round off
+	if len(decPart) > 18 {
```
yondonfu: Declare 18 as a constant? Maybe `maxDecimals`.

yondonfu (changes requested):

kyriediculous:
submitted fixups, I'll reword the commit messages during rebase

cmd/livepeer_cli/wizard_ticketbroker.go
```diff
@@ -46,20 +46,19 @@ func (w *wizard) deposit() {
...
 
 	fmt.Printf("Enter deposit amount in ETH - ")
-
-	depositAmount := w.readPositiveFloat()
+	depositAmount := w.readPositiveBaseAmount()
 
 	fmt.Printf("Enter reserve amount in ETH - ")
-
-	reserveAmount := w.readPositiveFloat()
+	fmt.Printf("> ")
```
yondonfu: Can remove this

go.sum
```diff
@@ -121,12 +121,8 @@ github.com/kr/pretty v0.1.0/go.mod h1:dAy3ld7l9f0ibDNOQOHHMYYIIbhfbHSm3C4ZsoJORN
 github.com/kr/pty v1.1.1/go.mod h1:pFQYn66WHrOpPYNljwOMqo10TkYh1fy3cYio2l3bCsQ=
 github.com/kr/text v0.1.0 h1:45sCR5RtlFHMR4UwH9sdQ5TC8v0qDQCHnXt+kaKSTVE=
 github.com/kr/text v0.1.0/go.mod h1:4Jbv+DJW3UT/LiOwJeYQe1efqtUx/iVham/4vfdArNI=
-github.com/livepeer/joy4 v0.1.1 h1:Tz7gVcmvpG/nfUKHU+XJn6Qke/k32mTWMiH9qB0bhnM=
```
yondonfu, line 124: Why did this change for this PR? Think it would be best not to pull in unrelated go mod changes.

cmd/livepeer_cli/wizard.go
```diff
@@ -85,6 +86,32 @@ func (w *wizard) readDefaultString(def string) string {
 	return def
 }
 
+func (w *wizard) readBaseAmountAndValidate(validate func(in *big.Int) error) *big.Int {
+	for {
+		text := w.readString()
```
yondonfu, line 91: Good call reusing this existing method!

yondonfu (changes requested):

go.sum
kyriediculous: I ran `go mod tidy` which prunes unused dependencies

cmd/livepeer_cli/wizard_ticketbroker.go
kyriediculous: Looked over it, fixed in 2cda6e56

yondonfu (commented):

kyriediculous:
rebased !

yondonfu (approved):
LGTM 🚢

PR #1248, opened 2019-12-05
Set HIGHEST_CHAIN_TAG=rinkeby during docker image build process

### LINKED ISSUE #1246: we should set `HIGHEST_CHAIN_TAG=rinkeby` during the docker image build process
We should set `HIGHEST_CHAIN_TAG=rinkeby` during the docker image build process

## Abstract

Currently, docker images are built without support for the Rinkeby testnet.

## Motivation

We will be recommending that new miner / transcoder participants try out the docker-compose solution for a quick bring-up for experimenting on the Livepeer testnet.  We need the image to support the Rinkeby testnet in order for this to work.

## Proposed Solution

Pull request will follow shortly

blocks #1223

mk-livepeer:
closes #1246

We will be recommending that new miner / transcoder participants try out the docker-compose solution for a quick bring-up for experimenting on the Livepeer testnet. We need the image to support the Rinkeby testnet in order for this to work.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -36,7 +36,7 @@ jobs:
 
       # Then, build the actual app in a container shared between Linux and Windows
       - run: docker pull livepeerci/build:latest || echo 'no pre-existing cache found'
-      - run: docker build -t livepeerci/build:latest --cache-from=livepeerci/build:latest -f docker/Dockerfile.build .
+      - run: docker build --build-arg HIGHEST_CHAIN_TAG=rinkeby -t livepeerci/build:latest --cache-from=livepeerci/build:latest -f docker/Dockerfile.build .
       - run: docker push livepeerci/build:latest
 
       # Finally, build the minimal go-livepeer distributable
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -36,7 +36,7 @@ jobs:
 
       # Then, build the actual app in a container shared between Linux and Windows
       - run: docker pull livepeerci/build:latest || echo 'no pre-existing cache found'
-      - run: docker build --build-arg HIGHEST_CHAIN_TAG=rinkeby -t livepeerci/build:latest --cache-from=livepeerci/build:latest -f docker/Dockerfile.build .
+      - run: docker build --build-arg HIGHEST_CHAIN_TAG -t livepeerci/build:latest --cache-from=livepeerci/build:latest -f docker/Dockerfile.build .
       - run: docker push livepeerci/build:latest
 
       # Finally, build the minimal go-livepeer distributable
--- a/docker/Dockerfile.build
+++ b/docker/Dockerfile.build
@@ -5,6 +5,9 @@
 
 FROM livepeerci/build-platform:latest
 
+ARG HIGHEST_CHAIN_TAG
+ENV HIGHEST_CHAIN_TAG ${HIGHEST_CHAIN_TAG}
+
 COPY ./install_ffmpeg.sh ./install_ffmpeg.sh
 RUN ./install_ffmpeg.sh
 
@@ -17,6 +20,6 @@ COPY go.sum go.sum
 RUN go mod download
 
 COPY . .
-RUN make livepeer livepeer_cli
+RUN HIGHEST_CHAIN_TAG=${HIGHEST_CHAIN_TAG} make livepeer livepeer_cli
 
 CMD ./upload_build.sh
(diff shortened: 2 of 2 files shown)


.circleci/config.yml
```diff
@@ -36,7 +36,7 @@ jobs:
 
       # Then, build the actual app in a container shared between Linux and Windows
       - run: docker pull livepeerci/build:latest || echo 'no pre-existing cache found'
-      - run: docker build -t livepeerci/build:latest --cache-from=livepeerci/build:latest -f docker/Dockerfile.build .
+      - run: docker build --build-arg HIGHEST_CHAIN_TAG=rinkeby -t livepeerci/build:latest --cache-from=livepeerci/build:latest -f docker/Dockerfile.build .
```
yondonfu: Since we already set the env var [here](https://github.com/livepeer/go-livepeer/blob/master/.circleci/config.yml#L22) can we omit the value and just do:

```
docker build --build-arg HIGHEST_CHAIN_TAG ...
```

The [Set build-time variables section](https://docs.docker.com/engine/reference/commandline/build/#set-build-time-variables---build-arg) of the Docker build docs note that if the value is omitted then the value from the local env will be propagated into the container being built.

docker/Dockerfile.build
```diff
@@ -17,6 +20,6 @@ COPY go.sum go.sum
 RUN go mod download
 
 COPY . .
-RUN make livepeer livepeer_cli
+RUN HIGHEST_CHAIN_TAG=${HIGHEST_CHAIN_TAG} make livepeer livepeer_cli
```
yondonfu: Do we need to set `HIGHEST_CHAIN_TAG` here if the `ENV` statement up top sets `HIGHEST_CHAIN_TAG` already?

yondonfu (changes requested):

.circleci/config.yml
mk-livepeer: I'd advise against it.  The way it works as-is ensures that the env var is defined both during the build and in the running release image.  It will be the only way to tell after the fact what variable was set during the build.

docker/Dockerfile.build
mk-livepeer: yes

.circleci/config.yml
mk-livepeer: OOPS...  this is a typo - will fix it now

yondonfu: Couldn't we tell what variable was set by looking at the [listed env vars](https://github.com/livepeer/go-livepeer/blob/master/.circleci/config.yml#L22)? Seems like the only difference in these two approaches is setting `HIGHEST_CHAIN_TAG` in the `environment` section vs. setting `HIGHEST_CHAIN_TAG` directly when passing the `--build-arg` flag.

Do you prefer setting the value for `HIGHEST_CHAIN_TAG` closer to where it is used as opposed to in a separate `environment` section? Note: We already set `HIGHEST_CHAIN_TAG` in a separate `environment` section when [creating the binary](https://github.com/livepeer/go-livepeer/blob/master/.circleci/config.yml#L63).

mkrufky: On second thought,
```
docker build --build-arg HIGHEST_CHAIN_TAG -t livepeerci/build:latest --cache-from=livepeerci/build:latest -f docker/Dockerfile.build .
```
...might be okay - I will push up the change.

mk-livepeer: done in 003fb82a053d9652243aba0dea69b439234bfb5b - it should be fine.  Thanks for the suggestion @yondonfu :-)

docker/Dockerfile.build
mk-livepeer: I provided detail for this item in the wrong thread :-/

The way it works as-is, with the ENV declared and also passed explicitly to the build process ensures that the env var is defined both during the build and in the running release image. It will be the only way to tell after the fact what variable was set during the build.

.circleci/config.yml
mk-livepeer: our paths crossed here - Yes, I prefer setting the env var closer to where it is actually used, but in this case it makes sense to use the CircleCI environment,

docker/Dockerfile.build
mkrufky: TBH, I'm beginning to question it myself now.  You'd think this would be unnecessary if it's actually in the running build environment.   I looked over some projects from the past, and I've always done it this way, but I think it was actually for the sake of compatibility with older docker versions, or simply force of habit.  If you want to give it a try without setting it explicitly in the RUN line, I'm willing to give that a try :-)

mk-livepeer: 88d9e9ce6324d72a3e97b47dc5c0dda3334984e6

yondonfu (approved):
LGTM

wohlner:

Issue #1215, opened 2019-11-26
Stake/Latency Based O Selection


yondonfu:
## Abstract

This spec outlines a stake/latency based O selection algorithm that Bs can use to select Os to work with.

## Motivation

This design is an extension of #938. It also seeks to address flaws in previous iterations of stake/latency based selection designs which included:

- Increased susceptibility to playback jitter due to a fan out approach that attempted to find the fastest Os in the network. While the fan out approach allowed a B to find Os with lower latency scores than the ones in its current working set, it also increased the variance of latencies since each O used during the fan out process could have very different latencies
- Increased cost overhead incurred by GPU session initialization due to a weighted random selection approach that increased the chance of sporadic segment assignment to Os. Each O incurs a cost overhead for initializing a GPU session for the first segment received in a stream and each session has a fixed timeout. So, if an O does not consistently receive segments, then eventually the session will timeout and the next time the O receives a segment for the same stream it will have incur the cost overhead of GPU session initialization again
- Increased quality variance across segments in a rendition due to a weighted random selection approach that will potentially assign segments of a stream to many different Os. If each O is using different GPUs, then the results produced can vary in quality. In the long term, we may be able to guarantee a certain quality level for each segment in a rendition by [requiring minimum quality scores](https://github.com/livepeer/verification-classifier/issues/71). In the short term, we would benefit from re-using the same GPUs by re-using the same Os

In order to address the flaws in previous iterations, this design's priorities are:

1. Consistent transcode latency. We still want to select Os that produce the fastest result, but more importantly, we want to select Os that *consistently* produce the fastest results
2. Re-use Os as much as possible. Re-using Os allows us to amortize the cost overhead of GPU session initialization across many segments. Additionally, re-using Os can help reduce quality variance in a rendition

## Proposed Solution

The `goodEnoughLatencyScore` parameter is the latency score considered to be "good enough". For example, setting `goodEnoughLatencyScore = 1.0` would mean that B considers real-time transcoding as good enough.

Selection consists of two phases:
- Working set creation
- Working set selection

Working set creation will largely work the same way that it does now - B will call `GetOrchestrator()` on each known O. In on-chain mode B will only send requests to active Os that are not banned (see the section below on dropping and banning). B will use the first N valid responses (where response validity is based on price, PM params, etc.) to form the working set. Using the first N valid responses favors the Os will the fastest initial response times. 

**Dropping and Banning Os**

[...]

Issue #1232, opened 2019-12-02
Investigate using EWMA latency score instead of last known latency score in BroadcastSession


yondonfu:
**Is your feature request related to a problem? Please describe.**

The first pass at stake/latency based O selection will use the last known latency score for each session. We should investigate whether using EWMA latency score would be better.

**Describe the solution you'd like**

Model EWMA latency score [parameters](https://en.wikipedia.org/wiki/EWMA_chart) and see if we would be better off using it instead of last known latency score. If so, we can calculate the EWMA latency score after receiving the transcode duration from `SubmitSegment()` and then attach the result to `BroadcastSession`.

PR #1249, opened 2019-12-05
[O selection] server: Add MinLSSelector

### LINKED ISSUE #1229: Add MinLSSelector that implements BroadcastSessionsSelector interface

We need to implement the stake/latency based selection strategy outlined in #1215.

Add a `MinLSSelector` type that implements the `BroadcastSessionsSelector` interface. The struct definition would look something like:

```
type heapBroadcastSession struct {
    *BroadcastSession
    idx int // For heap bookkeeping
}

type sessHeap []*heapBroadcastSession

type MinLSSelector struct {
    unknownSessList []*BroadcastSession
    knownSessList *sessHeap

    db *common.DB // Used to read stake values for each O

    goodEnoughLS float64
}
```

`sessHeap` will be the min-heap used to select the session with the min latency score. It will need to implement [heap.Interface](https://golang.org/pkg/container/heap/#Interface). The min-heap order should be based on the last known latency score for each session. We will need to add a `LatencyScore` field to `BroadcastSession`.

`MinLSSelector` will run a stake weighted random selection on `unknownSessList` with the stake of each session in `unknownSessionList` read from the DB.

`MinLSSelector` should implement the logic for popping from `unknownSessList` vs. `knownSessList` as described in #1215.

[NewSessionsManager()](https://github.com/livepeer/go-livepeer/blob/master/server/broadcast.go#L169) should instantiate `MinLSSelector`.

yondonfu:

This PR implements the stake/latency based selection strategy described in #1215. 

- Added a `LatencyScore` field to `BroadcastSession` so that slices of `BroadcastSession` can be sorted by `LatencyScore`
- Adds `MinLSSelector` which implements `BroadcastSessionsSelector`. This type contains the stake/latency based selection strategy logic

Added unit tests.

Fixes #1229

**Checklist:**

### DIFF
--- a/server/selection.go
+++ b/server/selection.go
@@ -2,10 +2,12 @@ package server
 
 import (
 	"container/heap"
+	"errors"
 	"math/rand"
 
 	ethcommon "github.com/ethereum/go-ethereum/common"
 	"github.com/golang/glog"
+	"github.com/livepeer/go-livepeer/common"
 )
 
 // BroadcastSessionsSelector selects the next BroadcastSession to use
@@ -62,6 +64,28 @@ func (h *sessHeap) Peek() interface{} {
 
 type stakeReader interface {
 	Stakes(addrs []ethcommon.Address) (map[ethcommon.Address]int64, error)
+}
+
+type storeStakeReader struct {
+	store common.OrchestratorStore
+}
+
+func (r *storeStakeReader) Stakes(addrs []ethcommon.Address) (map[ethcommon.Address]int64, error) {
+	orchs, err := r.store.SelectOrchs(&common.DBOrchFilter{Addresses: addrs})
+	if err != nil {
+		return nil, err
+	}
+
+	if len(orchs) != len(addrs) {
+		return nil, errors.New("could not fetch all stake weights")
+	}
+
+	stakes := make(map[ethcommon.Address]int64)
+	for _, orch := range orchs {
+		stakes[ethcommon.HexToAddress(orch.EthereumAddr)] = orch.Stake
+	}
+
+	return stakes, nil
 }
 
 // MinLSSelector selects the next BroadcastSession with the lowest latency score if it is good enough.
--- a/server/selection.go
+++ b/server/selection.go
@@ -1,5 +1,13 @@
 package server
 
+import (
+	"container/heap"
+	"math/rand"
+
+	ethcommon "github.com/ethereum/go-ethereum/common"
+	"github.com/golang/glog"
+)
+
 // BroadcastSessionsSelector selects the next BroadcastSession to use
 type BroadcastSessionsSelector interface {
 	Add(sessions []*BroadcastSession)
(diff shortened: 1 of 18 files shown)


yondonfu:
Ready for full review.

server/selection_test.go
```diff
@@ -0,0 +1,250 @@
...
+				OrchestratorInfo: &net.OrchestratorInfo{
+					TicketParams: &net.TicketParams{Recipient: []byte("foo")},
+				},
+			},
+		},
+	)
+
+	// Test error when reading stake
+	stakeRdr.err = errors.New("Stakes error")
+	assert.Nil(sel.unknownSessListSelect())
```
kyriediculous: Can we add a check that an error was logged using `glog.Stats.Error.Lines()`?

server/selection_test.go
```diff
@@ -0,0 +1,250 @@
...
+	}
+
+	sel.Add(sessions)
+
+	i := 0
+	// Check that we select sessions based on the order of unknownSessList and that the size of
+	// unknownSessList decreases with each selection
+	for sel.Size() > 0 {
+		sess := sel.unknownSessListSelect()
+		assert.Equal(t, sess, sessions[i])
```
kyriediculous: this will always be equal as we are going to compare values here and not the pointers. To compare the pointers we have to explicitly say so.  So _I think_  in the current state the test will always pass.

kyriediculous (changes requested):

server/selection_test.go
yondonfu: Fixed in 703441e

server/selection_test.go
yondonfu: Fixed in 703441e - using [assert.Same()](https://godoc.org/github.com/stretchr/testify/assert#Same)

server/selection.go
```diff
@@ -1,5 +1,14 @@
 package server
 
+import (
+	"container/heap"
+	"crypto/rand"
```
j0sh: Is there a reason to use `crypto/rand` here rather than `math/rand` ? The latter has a nicer interface IMO - eg, no special handling needed to make sure the function doesn't `panic` on you, better seeding / state support if we need it, etc.

server/selection.go
```diff
@@ -9,6 +18,161 @@ type BroadcastSessionsSelector interface {
...
+
+	return sess
+}
+
+func (h *sessHeap) Peek() interface{} {
+	if h.Len() == 0 {
+		return nil
+	}
+
+	return (*h)[0]
```
j0sh, line 62: We peek from the front, but pop from the back?

server/selection.go
```diff
@@ -9,6 +18,161 @@ type BroadcastSessionsSelector interface {
...
+
+// Select returns the session with the lowest latency score if it is good enough.
+// Otherwise, a session without a latency score yet is returned
+func (s *MinLSSelector) Select() *BroadcastSession {
+	if s.knownSessList.Len() == 0 {
+		return s.unknownSessListSelect()
+	}
+
+	// knownSessList is not empty at this point so we know minSess is not nil
+	minSess := s.knownSessList.Peek().(*BroadcastSession)
```
j0sh: Not sure how much of a concern this is currently, but if there is a small chance that the list could become empty between the previous `Len()` and this `Peek()` (and even the `Pop`) then it might be safer to do something like the following:

```golang
sess = s.knownSessList.Peek()
if sess == nil {
  return s.unknownSessListSelect()
}
minSess := sess.(*BroadcastSession)
...
```

Or the session could be popped unconditionally, and re-pushed in case the latency score is not good enough? This also ensures you're working with a consistent session throughout the function - eg, there's no longer an assumption that `Peek` and `Pop` will always return the same values in consecutive invocations. Something like this:
```golang
sess := heap.Pop(s.knownSessList)
if sess == nil || (sess.*BroadacastSession).LatencyScore < s.goodEnoughLS {
  unknownSess := s.unknownSessListSelect()
  if unknownSess != nil {
   // Use unknownSess instead, so return sess back to list
    s.Complete(sess) // maybe add special handling to Complete to ignore nil arguments?
    return unknownSess
  }
}
return sess // May be nil
```
Note that this also handles the same type of conditions with the unknownSessList being non-empty in one call  and empty in another (length check vs selection), and allows for fall-back to an otherwise valid known session, in the case the `unknownSessListSelect` ever returns empty.

Even though this sort of behavior probably isn't likely right now (are we sure the selector won't be used concurrently right now?), it's safer from a defensive programming standpoint.

server/selection.go
```diff
@@ -9,6 +18,161 @@ type BroadcastSessionsSelector interface {
...
+
+// MinLSSelector selects the next BroadcastSession with the lowest latency score if it is good enough.
+// Otherwise, it selects a session that does not have a latency score yet
+type MinLSSelector struct {
+	unknownSessList []*BroadcastSession
+	knownSessList   *sessHeap
+
+	stakeRdr stakeReader
+
+	goodEnoughLS float64
```
j0sh: Maybe this could be called "minLS"  instead? It's shorter and reads better within the code, eg `if sess.latencyScore < minLS { ... `

server/selection.go
```diff
@@ -9,6 +18,161 @@ type BroadcastSessionsSelector interface {
...
+
+type stakeReader interface {
+	Stakes(addrs []ethcommon.Address) (map[ethcommon.Address]*big.Int, error)
+}
+
+// MinLSSelector selects the next BroadcastSession with the lowest latency score if it is good enough.
+// Otherwise, it selects a session that does not have a latency score yet
+type MinLSSelector struct {
+	unknownSessList []*BroadcastSession
+	knownSessList   *sessHeap
```
j0sh: What do you think about calling these `unknownSessions` and `knownSessions` ?

I realize the old nomenclature used `sessList` (and `sessMap`) but naming based on type always felt a little off. We have somewhat helpful tooling and a compiler that catches typing mistakes for us, so we don't need to muddle that with somewhat inadequate naming (eg, a heap isn't necessarily a list, and what if we ever change the datatype...)

server/selection.go
```diff
@@ -9,6 +18,161 @@ type BroadcastSessionsSelector interface {
...
+
+// Clear resets the selector's state
+func (s *MinLSSelector) Clear() {
+	s.unknownSessList = nil
+	s.knownSessList = &sessHeap{}
+	s.stakeRdr = nil
+}
+
+// Use stake weighted random selection to select from unknownSessList
+func (s *MinLSSelector) unknownSessListSelect() *BroadcastSession {
```
j0sh: This name reads a bit strangely, what about `selectUnknownSession` ?

server/selection.go
```diff
@@ -9,6 +18,161 @@ type BroadcastSessionsSelector interface {
...
+		// If we fail to generate a random stake weight we should not continue with selection
+		if err != nil {
+			glog.Errorf("failed to generate random stake weight for selection: %v", err)
+			return nil
+		}
+
+		r = randStake
+	}
+
+	for i, sess := range s.unknownSessList {
```
j0sh: Can you elaborate on the selection algorithm here? Perhaps as a comment?

server/selection_test.go
```diff
@@ -0,0 +1,254 @@
...
+
+		// Call Add() to add the session back to unknownSessList
+		sel.Add([]*BroadcastSession{sess})
+	}
+
+	sort.Slice(stakes, func(i, j int) bool {
+		return stakes[i].Cmp(stakes[j]) < 0
+	})
+
+	// Check that higher stake weight sessions are selected more often than lower stake weight sessions
```
j0sh, line 254: I guess technically we'd want to check the ratio of selection to ensure the weighing works as expected, but this is probably OK as well.

j0sh (changes requested):

server/selection.go
yondonfu: Fixed in d754c3f

server/selection.go
yondonfu: Fixed in d754c3f

server/selection.go
yondonfu: Fixed in d754c3f

server/selection.go
yondonfu: Fixed in 09df6bc

server/selection_test.go
yondonfu: Fixed in 09df6bc - added a test to check that the selection count ratio matches up with the stake weight ratio.

server/selection.go
yondonfu: Fixed in 09df6bc. Using `(*big.Int).Rand()` with a source initialized from `math/rand` instead of using `crypto/rand`.

server/selection.go
yondonfu: We peek from the front since the minimum element is at the 0th index (see the [container/heap](https://golang.org/pkg/container/heap/) docs). We pop from the back since since [heap.Pop()](https://golang.org/src/container/heap/heap.go?s=2190:2223#L50) swaps the 0th element with the last element before calling the interface's `Pop()` method.

server/selection.go
yondonfu: > Not sure how much of a concern this is currently, but if there is a small chance that the list could become empty between the previous Len() and this Peek() (and even the Pop) then it might be safer to do something like the following

Removed the `Len()` call in favor of just calling `Peek()` and checking for a nil session in 711f9ea.

> Even though this sort of behavior probably isn't likely right now (are we sure the selector won't be used concurrently right now?)

The selector's caller (i.e. `BroadcastSessionsManager`) should implement concurrency safety. Added a comment in 711f9ea noting that `MinLSSelector` is not concurrency safe.

> Or the session could be popped unconditionally, and re-pushed in case the latency score is not good enough?

While I like this approach from a defensive programming standpoint, I don't like that it requires 1 heap pop + 1 heap push (`O(lgn)` cost for each) instead of just 1 heap pop whenever a known session's latency score is not good enough. Since we're offloading the responsibility of concurrency safety to the selector's caller, I'm inclined not to use this defensive programming approach since its slightly more expensive.

kyriediculous (approved):

server/selection.go
j0sh: Maybe add a comment to the code explaining this?

server/selection.go
j0sh: > I don't like that it requires 1 heap pop + 1 heap push (O(lgn) cost for each) instead of just 1 heap pop whenever a known session's latency score is not good enough

Note that N is currently quite small - IIRC, with our current parameters, it maxes out around N=8 orchestrators [1].

For the case where all orchestrators are slow but otherwise performing well enough to establish a steady state, the heap is never pushed more than N times. If the network is performing well, then the additional heap push will never happen [2]. Amortized across all invocations, the cost of the additional heap push should be small.

While I do prefer the defensive approach, will leave it up to you.

[1] HTTPTimeout.Seconds() = 8, SegLen.Seconds() = 2   https://github.com/livepeer/go-livepeer/blob/7fcf97071963657c96176188b398fe9466ba0a6a/server/broadcast.go#L167-L168

[2] The current approach of only considering the most recent latency score all but ensures that well-performing Os will get penalized by transient issues. So right now, the latency selector's view of network health is going to be a bit sensitive for a while: not only will we get additional heap pushes, overall latency may get worse due to mostly-fast Os being mistakenly placed at the back of the queue where it's difficult to recover.

server/selection.go
yondonfu: Added a comment in 5bd2145

server/selection.go
yondonfu: I was originally going to go with the defensive approach, but I realized that the current implementation of `sessHeap` will panic when `heap.Pop()` and `len(sessHeap) = 0` due to an index out of range error triggered by calling `sessHeap.Swap()`.  We could update the implementations of `sessHeap.Swap()` and `sessHeap.Pop()` so that `heap.Pop()` returns nil when `len(sessHeap) = 0` but this would require changing `sessHeap.Swap()` to no-op when `len(sessHeap) = 0` which feels a bit weird to me. Furthermore, this change would introduce the length check into `sessHeap.Swap()` which means we still have the length check it is just executed at a different point in the code. If we're going to need the length check no matter what, I'd prefer to just have the caller of `heap.Pop()` do the length check instead of doing it in `sessHeap.Swap()` as this avoids adding a no-op in `sessHeap.Swap()` when `len(sessHeap) = 0`.

> The current approach of only considering the most recent latency score all but ensures that well-performing Os will get penalized by transient issues.

Good point. Perhaps we can address this with #1232

yondonfu:
11abd0c updates `MinLSSelector` and the `stakeReader` interface to use int64 to represent stake values instead of big.Int since #1253 stores stake values as fixed point numbers with the int64 type.

server/selection.go
j0sh: Ah I see what you mean - it's because the golang implementation is stupid:

https://golang.org/src/container/heap/heap.go?s=2190:2223#L50

OK then.

j0sh (approved):
🔥 🚀 🌕

PR #1250, opened 2019-12-06
eth: log tx params

### LINKED ISSUE #1220: Log tx params

We currently only [log the method name](https://github.com/livepeer/go-livepeer/blob/master/eth/backend.go#L95) for a submitted tx. It would be useful to also log the params used for the tx.

The [backend.SendTransaction()](https://github.com/livepeer/go-livepeer/blob/master/eth/backend.go#L68) method has access to a signed tx that has been submitted. The tx includes:

- The `to` address which indicates the contract that is being called
- The `data` payload which contains the ABI encoded tx data

The `to` address allows us to find the ABI that corresponds to the contract being called. The ABI can then be used to decode the `data` payload and parse the human readable params that can be logged.

kyriediculous:

This PR enables the logging of transaction inputs in the console when submitting eth transactions.

- Create a map of `[ethcommon.Address]abi.ABI` on `eth.Backend` enabling contract address to ABI lookup in `eth.Backend.SendTransaction()`

- Added a helper to `eth/helpers` , `decodeTxParams()`

- Call `decodeTxParams()` in `eth.Backend.SendTransaction()`

- Create a string of the transaction input names and values

- Add the parsed transaction input string to the log messages

- Add a test for `decodeTxParams()`

- Ran unit tests

- Ran `TicketBroker.FundDepositAndReserve()` from a local broadcaster node on rinkeby

```

Invoking transaction: "fundDepositAndReserve". Transaction Inputs: "_depositAmount: 10000000000000000   _reserveAmount: 10000000000000000   "  Hash: "0xefcb6fe0b647b173e2ceeb30715e1d2f2775a6d79ea792394f3cee04c36b287b".

```

Fixes #1220 

**Checklist:**


eth/backend.go
```diff
@@ -42,18 +43,20 @@ type Backend interface {
 type backend struct {
 	*ethclient.Client
 	methods      map[string]string
+	abiMap       map[string]abi.ABI
```
kyriediculous: should we turn this into a pointer ref?

kyriediculous:
added hex encoding for some common types used in our contracts that would be otherwise printed out as byte slices c6c21e07

kyriediculous:
Still pondering on how to deal with tuples and arrays

kyriediculous:
Now succesfully decoding all types ! 

batchRedeemWinningTicket result: 

```
_tickets: [ { Recipient: 0x421e8e76020C725890488D5f45e3b035162B4AC5  Sender: 0x1EB5d124fE565fF06CeF6Bf1B61F4a9952EfFA25  FaceValue: 5000000000  WinProb: 50000000  SenderNonce: 99999  RecipientRandHash: 0x1a259587e227bdc8ffb22a92ea2e512290f5ce56063d7a39d549d2053f4496d0  AuxData: 0xd5ecce4c920ef8683ef2023b1257d32607657db005e6490fb0461bfd834fb7fb8ef464ebfce014537e0c500386f2aeec7cd41a5fc5acbb9db64a49ef6a1df66a }  { Recipient: 0x7064419EB0C7D879819f5834d679908e78B9E31F  Sender: 0xE2F444C4EeFe14741c8E7C21FE811C9C0e51F094  FaceValue: 5000000000  WinProb: 50000000  SenderNonce: 99999  RecipientRandHash: 0xc25ef67e0cd329df611f6cb02fb56ba43ae080e85e959180eb48de13dcd18ff4  AuxData: 0xefa861061e801d59075f9ab9a64b7a5762630d69151071c845eabb58080ef590195f7bdab9111f3bfb92895a4adf8eac408df980a8eb37b6cb8dbceacdc054c0 } ]   _sigs: [ 0x91d65ccba99a4da150389dc6aa0f42346f21e28fa3ef3ebf1bfd66f691e0114bcc994fc4f493fcc7c58507f5168806d2dc0e66856e2199bd8bc959468284ab15  0x0f2c5250cf9a75d4927e4a0030938b99cbb599de68043b76cc3ec28981ef325d29d5bf7592582aadaf801153265c94b926982a354cd6951d362c785aa2dd385a ]   _recipientRands: [ 111111  222222 ]
```

eth/backend.go
```diff
@@ -87,12 +90,23 @@ func (b *backend) SendTransaction(ctx context.Context, tx *types.Transaction) er
 		method = "unknown"
 	}
 
+	txParams := make(map[string]interface{})
+	err = decodeTxParams(b.abiMap[string(data[:4])], txParams, data)
 	if err != nil {
```
mkrufky: since we're only reading `err` within this `if` block, we can optimize as follows:
```go
if err := decodeTxParams(b.abiMap[string(data[:4])], txParams, data); err != nil {
    return err
}
```

eth/backend.go
```diff
@@ -87,12 +90,23 @@ func (b *backend) SendTransaction(ctx context.Context, tx *types.Transaction) er
...
 		return err
 	}
 
-	glog.Infof("\n%vEth Transaction%v\n\nInvoking transaction: \"%v\".  Hash: \"%v\". \n\n%v\n", strings.Repeat("*", 30), strings.Repeat("*", 30), method, tx.Hash().String(), strings.Repeat("*", 75))
+	var txParamsString string
+	for name, val := range txParams {
+		txParamsString += fmt.Sprintf("%v: %v   ", name, val)
+	}
+
+	if err != nil {
```
mkrufky, line 117: We dont need this `nil` check here.  We tested the return value of `decodeTxParams()` just above, and there hasn't been any reassignment to `err` since.  `err` cant be `nil` here because we would have already returned with `err`

mkrufky (changes requested):

eth/backend.go
kyriediculous: Good point. 

Is there a difference from a runtime perspective? As in is declaring a new err var more expensive than reassigning the existing err var or not?

mkrufky: From a run-time perspective, it's pretty much the same.  It's a code readability improvement, mostly.  This scopes the return value of `decodeTxParams` to exist solely within the affirmative if case code block. The expense of allocating this error variable is negligible to non-existent - the compiler knows to optimize it.

eth/backend.go
kyriediculous: The original intention here is to display the transaction failed `glog` statement but this codepath can never be reached as you say which was already the case before this PR. 

Therefore I have renamed the error returned from `SendTransaction` and we delay the check until we have decoded the tx params. d3504b70

eth/backend.go
kyriediculous: That's pretty clear, thanks ! 
fixed in d3504b70

kyriediculous:
I rebased and force pushed. 
Test pass locally, they should do now in CI as well after rebasing.

eth/backend.go
```diff
@@ -89,12 +89,22 @@ func (b *backend) SendTransaction(ctx context.Context, tx *types.Transaction) er
...
-	if err != nil {
-		glog.Infof("\n%vEth Transaction%v\n\nInvoking transaction: \"%v\".  \nTransaction Failed: %v\n\n%v\n", strings.Repeat("*", 30), strings.Repeat("*", 30), method, err, strings.Repeat("*", 75))
+	txParams := make(map[string]interface{})
+	if err := decodeTxParams(b.abiMap[string(data[:4])], txParams, data); err != nil {
 		return err
 	}
 
-	glog.Infof("\n%vEth Transaction%v\n\nInvoking transaction: \"%v\".  Hash: \"%v\". \n\n%v\n", strings.Repeat("*", 30), strings.Repeat("*", 30), method, tx.Hash().String(), strings.Repeat("*", 75))
+	var txParamsString string
+	for name, val := range txParams {
```
yondonfu: We can list the tx params in the order that they are included in tx data by relying on the ordering of the `Inputs` slice stored in an ABI. Something like:

```golang
for _, arg := range abi.Methods[method].Inputs {
    txParamsString += fmt.Sprintf("%v: %v  ", arg.Name, txParams[arg.Name])
}
```

I think listing the tx params in the order that they are included in tx data makes it easier to check param values included in a tx given knowledge of the contract function being called.

eth/backend.go
```diff
@@ -89,12 +89,22 @@ func (b *backend) SendTransaction(ctx context.Context, tx *types.Transaction) er
...
+		txParamsString += fmt.Sprintf("%v: %v   ", name, val)
```
yondonfu: `txParamsString` will always have extra whitespace when printed. We could clean up the extra whitespace calling [strings.TrimSpace()](https://golang.org/pkg/strings/#TrimSpace) after constructing `txParamsString`.

eth/backend.go
```diff
@@ -89,12 +89,22 @@ func (b *backend) SendTransaction(ctx context.Context, tx *types.Transaction) er
...
+	}
+
+	if sendErr != nil {
+		glog.Infof("\n%vEth Transaction%v\n\nInvoking transaction: \"%v\".  Transaction Inputs: \"%v\"   \nTransaction Failed: %v\n\n%v\n", strings.Repeat("*", 30), strings.Repeat("*", 30), method, txParamsString, err, strings.Repeat("*", 75))
```
yondonfu: Nit: Since this statement is already pretty long perhaps we can shorten "Transaction Inputs" to "Inputs" since the rest of the statement should convey that the inputs are for a transaction.

eth/backend.go
```diff
@@ -89,12 +89,22 @@ func (b *backend) SendTransaction(ctx context.Context, tx *types.Transaction) er
...
+		return sendErr
+	}
+
+	glog.Infof("\n%vEth Transaction%v\n\nInvoking transaction: \"%v\". Transaction Inputs: \"%v\"  Hash: \"%v\". \n\n%v\n", strings.Repeat("*", 30), strings.Repeat("*", 30), method, txParamsString, tx.Hash().String(), strings.Repeat("*", 75))
```
yondonfu: Same comment as above regarding "Transaction Inputs" -> "Inputs".

eth/backend.go
yondonfu: `sendErr` should be included in this log statement, but `err` is included right now.

eth/backend.go
```diff
@@ -42,19 +43,21 @@ type Backend interface {
 type backend struct {
 	*ethclient.Client
 	methods      map[string]string
```
yondonfu: Do we still need this map if we're storing a map from method IDs -> ABIs? If we have access to the ABI then we can fetch the method name via `abi.MethodById(methodID).Name`.

eth/backend.go
yondonfu: Yeah seems reasonable for avoiding allocating multiple ABI instances.

yondonfu (changes requested):

eth/backend.go
kyriediculous: fixed in 8a965a6

eth/backend.go
kyriediculous: fixed in e89ba37

eth/backend.go
kyriediculous: fixed in e89ba37

eth/backend.go
kyriediculous: fixed in e89ba37

eth/backend.go
kyriediculous: fixed in e89ba37

eth/backend.go
kyriediculous: fixed in 8a3dda8

eth/backend.go
kyriediculous: good point ! fixed in 71307949

eth/backend.go
```diff
@@ -127,17 +141,18 @@ func (b *backend) retryRemoteCall(remoteCall func() ([]byte, error)) (out []byte
 	return out, err
 }
 
-func makeMethodsMap() (map[string]string, error) {
-	methods := make(map[string]string)
+func makeMethodsMap() (map[string]*abi.ABI, error) {
```
yondonfu: Should this be called `makeABIMap`?

eth/backend.go
```diff
@@ -84,17 +82,33 @@ func (b *backend) SendTransaction(ctx context.Context, tx *types.Transaction) er
...
-	method, ok := b.methods[string(data[:4])]
-	if !ok {
-		method = "unknown"
+	methodSig := string(data[:4])
+	method, err := b.abiMap[methodSig].MethodById(data[:4])
+	if err != nil {
+		return err
+	}
+	var methodName string
+	if method == nil {
```
yondonfu: If we're already checking the error returned by `MethodById()` on L87, I don't think we need this nil check for `method`.

eth/backend.go
```diff
@@ -84,17 +82,33 @@ func (b *backend) SendTransaction(ctx context.Context, tx *types.Transaction) er
...
+	for _, arg := range method.Inputs {
+		txParamsString += fmt.Sprintf("%v: %v  ", arg.Name, txParams[arg.Name])
+	}
+
+	if sendErr != nil {
+		glog.Infof("\n%vEth Transaction%v\n\nInvoking transaction: \"%v\". Inputs: \"%v\"   \nTransaction Failed: %v\n\n%v\n", strings.Repeat("*", 30), strings.Repeat("*", 30), methodName, strings.TrimSpace(txParamsString), sendErr, strings.Repeat("*", 75))
+		return sendErr
+	}
+
+	glog.Infof("\n%vEth Transaction%v\n\nInvoking transaction: \"%v\". Inputs: \"%v\"  Hash: \"%v\". \n\n%v\n", strings.Repeat("*", 30), strings.Repeat("*", 30), methodName, txParamsString, tx.Hash().String(), strings.Repeat("*", 75))
```
yondonfu: Missing a call to `strings.TrimSpace()` for `txParamsString` here. If we trim the whitespace for `txParamsString` once prior to these log statements then we can avoid having to remember to call `strings.TrimSpace` in two different places.

eth/backend.go
```diff
@@ -84,17 +82,33 @@ func (b *backend) SendTransaction(ctx context.Context, tx *types.Transaction) er
 	b.nonceManager.Unlock(sender)
 
 	data := tx.Data()
-	method, ok := b.methods[string(data[:4])]
-	if !ok {
-		method = "unknown"
+	methodSig := string(data[:4])
+	method, err := b.abiMap[methodSig].MethodById(data[:4])
```
yondonfu: If we set `methodSig = data[:4]` then we can reuse it so we only have to index into `data` once (minimizes the chance of accidentally passing in the wrong index value). Then, we can pass `string(methodSig)` to `decodeTxParams()`.

yondonfu (changes requested):

eth/backend.go
kyriediculous: fixed in b47e031

eth/backend.go
kyriediculous: fixed in b47e031

eth/backend.go
kyriediculous: fixed in b47e031

eth/backend.go
kyriediculous: fixed in b47e031

eth/backend.go
```diff
@@ -84,17 +82,33 @@ func (b *backend) SendTransaction(ctx context.Context, tx *types.Transaction) er
...
-		method = "unknown"
+	methodSig := data[:4]
+	method, err := b.abiMap[string(methodSig)].MethodById(methodSig)
+	if err != nil {
+		return err
 	}
 
-	if err != nil {
-		glog.Infof("\n%vEth Transaction%v\n\nInvoking transaction: \"%v\".  \nTransaction Failed: %v\n\n%v\n", strings.Repeat("*", 30), strings.Repeat("*", 30), method, err, strings.Repeat("*", 75))
+	var methodName string
```
yondonfu: Looks like we can just use a short declaration for `methodName` here i.e. `methodName := method.Name`

yondonfu (commented):
Left one comment about a short variable declaration. Feel free to rebase as well.

eth/backend.go
kyriediculous: Actually if method is nil this will be a nil pointer so that's why I orignally added `methodName = "unknown"` if we couldn't find the ABI in the map. 

but instead we could do
```if method != nil { methodName =  method.Name }```

Comment (yondonfu) [reply eth/backend.go]: Ah right - good point about not being able to find the ABI in `abiMap`.

In that case, we should have a check that confirms that we actually have a valid ABI. Something like:

```
var methodName string
var txParamsString string

methodSig := data[:4]
abi, ok := b.abiMap[string(methodSig)]
if !ok {
    methodName = "unknown"
} else {
    methodName = method.Name

    // Decode tx params and populate txParamsString
}

// Log statements
```

**Comment by kyriediculous** [comment]
fixed in most recent fixup

File: eth/backend.go [hunk 14]
```diff
@@ -83,18 +81,38 @@ func (b *backend) SendTransaction(ctx context.Context, tx *types.Transaction) er
...
+		if err == nil {
+			method = m
+			methodName = m.Name
+		}
+		txParams := make(map[string]interface{})
+		if err := decodeTxParams(b.abiMap[string(methodSig)], txParams, data); err != nil {
+			return err
+		}
+
+		for _, arg := range method.Inputs {
```
yondonfu: Ah so it *looks* like `method` could be nil here, but if `method` is nil then `decodeTxParams()` would have already returned an error since `decodeTxParams()` also calls `abi.MethodById()` internally. 

Extracting this logic into a function could make the control flow here easier to follow and would also make this logic easier to test (would be good to add tests for the new function). Something like:

```golang
type txLog struct {
	method string
	params string
}

func (b *backend) newTxLog(tx *types.Transaction) (txLog, error) {
	data := tx.Data()
	methodSig := data[:4]
	abi, ok := b.abiMap[string(methodSig)]
	if !ok {
		return txLog{}, errors.New("unknown ABI")
	}

	m, err := abi.MethodById(methodSig)
	if err != nil {
		return txLog{}, err
	}

	methodName := m.Name
	txParams := make(map[string]interface{})
	if err := decodeTxParams(b.abiMap[string(methodSig)], txParams, data); err != nil {
		return txLog{}, err
	}

	var txParamsString string
	for _, arg := range m.Inputs {
		txParamsString += fmt.Sprintf("%v: %v  ", arg.Name, txParams[arg.Name])
	}

	txParamsString = strings.TrimSpace(txParamsString)

	return txLog{
		method: methodName,
		params: txParamsString,
	}, nil
}
```

Then, we could simplify the code in `SendTransaction()` to something like:

```golang
txLog, err := b.newTxLog(tx)
if err != nil {
    txLog.method = "unknown"
}

// Existing logging logic below
if sendErr != nil {
    ...
}

...
```

kyriediculous: fixed in ac6014f5

yondonfu (commented):
Changes look good. Let's rebase.

kyriediculous:
done !

yondonfu (approved):
LGTM 🚢

PR #1251, opened 2019-12-06
eth, server: faucet rate limit error message

### LINKED ISSUE #1245: Alert user that Rinkeby faucet is rate limited to 1 request per hour

The Rinkeby faucet is rate limited to 1 request per hour. All subsequent requests sent from `livepeer_cli` will fail until the next hour.

`livepeer_cli` should check if the user has sent a request to the faucet in the last hour and alert the user that it has to wait an hour before sending another request.

kyriediculous:
This PR logs an error messages when a user tries to request LPT from the faucet while still rate limited, the error message includes the amount of minutes to wait. 

- exposes `LivepeerEthClient.NextValidRequest`
- add a check if valid request is in the future

Ran broadcaster node and tried requesting from the faucet multiple times on rinkeby

```
E1206 02:43:53.318493   54102 webserver.go:984] Error requesting tokens from faucet: can only request tokens once every hour, please wait 51 more minutes
```

Fixes #1245

**Checklist:**

### DIFF
--- a/server/webserver.go
+++ b/server/webserver.go
@@ -984,6 +984,30 @@ func (s *LivepeerServer) cliWebServerHandlers(bindAddr string) *http.ServeMux {
 
 	mux.HandleFunc("/requestTokens", func(w http.ResponseWriter, r *http.Request) {
 		if s.LivepeerNode.Eth != nil {
+
+			nextValidRequest, err := s.LivepeerNode.Eth.NextValidRequest(s.LivepeerNode.Eth.Account().Address)
+			if err != nil {
+				glog.Errorf("Unable to get the time for the next valid request from faucet: %v", err)
+				return
+			}
+			backend, err := s.LivepeerNode.Eth.Backend()
+			if err != nil {
+				glog.Errorf("Unable to get LivepeerEthClient backend: %v", err)
+				return
+			}
+
+			blk, err := backend.BlockByNumber(context.Background(), nil)
+			if err != nil {
+				glog.Errorf("Unable to get latest block")
+				return
+			}
+
+			now := int64(blk.Time())
+			if nextValidRequest.Int64() != 0 && nextValidRequest.Int64() > now {
+				glog.Errorf("Error requesting tokens from faucet: can only request tokens once every hour, please wait %v more minutes", (nextValidRequest.Int64()-now)/60)
+				return
+			}
+
 			glog.Infof("Requesting tokens from faucet")
 
 			tx, err := s.LivepeerNode.Eth.Request()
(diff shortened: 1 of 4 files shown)


server/webserver.go
```diff
@@ -984,6 +984,14 @@ func (s *LivepeerServer) cliWebServerHandlers(bindAddr string) *http.ServeMux {
 
 	mux.HandleFunc("/requestTokens", func(w http.ResponseWriter, r *http.Request) {
 		if s.LivepeerNode.Eth != nil {
+
+			nextValidRequest, err := s.LivepeerNode.Eth.NextValidRequest(s.LivepeerNode.Eth.Account().Address)
+			now := time.Now().UnixNano() / int64(time.Second)
```
yondonfu: Since the local clock is going to keep progressing even when blockchain time is not progressing it's possible that this check passes, but the request is not yet valid because the block with a timestamp that would make the request valid has not been included yet. Probably safer to just query for the current block timestamp.

Something like:

```
blk, err := s.LivepeerNode.Eth.Backend().GetBlock()
// Handle error
now := blk.Time()
```

yondonfu (changes requested):

server/webserver.go
kyriediculous: fixed in 8f0e80d6 but I had to extend the `Backend` interface with an embedded `ChainReader`

server/webserver.go
```diff
@@ -984,6 +984,32 @@ func (s *LivepeerServer) cliWebServerHandlers(bindAddr string) *http.ServeMux {
...
+			backend, err := s.LivepeerNode.Eth.Backend()
+			if err != nil {
+				glog.Errorf("Unable to get LivepeerEthClient backend: %v", err)
+				return
+			}
+			blkNum, err := s.LivepeerNode.Database.LastSeenBlock()
+			if err != nil {
+				glog.Errorf("Unable to get latest seen block number from database: %v", err)
+			}
+			blk, err := backend.BlockByNumber(context.Background(), blkNum)
```
yondonfu: We can just pass in `nil` for the block number to fetch the latest block instead of reading the latest block number from the DB.

server/webserver.go
```diff
@@ -984,6 +984,32 @@ func (s *LivepeerServer) cliWebServerHandlers(bindAddr string) *http.ServeMux {
...
+			if err != nil {
+				glog.Errorf("Unable to get latest block")
```
yondonfu, line 1001: Need a return when there is an error.

yondonfu (changes requested):

server/webserver.go
kyriediculous: fixed in eb24ec78

server/webserver.go
kyriediculous: fixed in eb24ec78

yondonfu (commented):

kyriediculous:
rebased !

yondonfu (approved):
LGTM

PR #1253, opened 2019-12-06
[O selection] Fetch stake from the DB

yondonfu:

This PR adds support for fetching Os by address from the DB. This feature is a requirement for stake/latency based selection because we need to be able to fetch the stake for specific Os in B's working set.

- 22c8b6a: Use the BLOB type for the stake column in the orchestrators table
- a7fef47: Add an `Addresses` field in `DBOrchFilter` that is used to filter for specific Os by `EthereumAddr` in the SQL query for `SelectOrchs()`

The main motivation behind the BLOB column type update in 22c8b6a was to avoid an additional check when converting from the DB result to a `big.Int` (`string` -> `big.Int` requires an additional check while `[]byte` -> `big.Int` does not). Happy to discuss whether the benefit of change outweighs downsides though!

Added unit tests.

N/A

**Checklist:**

### DIFF
--- a/common/db.go
+++ b/common/db.go
@@ -61,6 +61,7 @@ type DBUnbondingLock struct {
 type DBOrchFilter struct {
 	MaxPrice     *big.Rat
 	CurrentRound *big.Int
+	Addresses    []ethcommon.Address
 }
 
 var LivepeerDBVersion = 1
@@ -701,6 +702,14 @@ func buildFilterOrchsQuery(filter *DBOrchFilter) (string, error) {
 		if filter.CurrentRound != nil {
 			currentRound := filter.CurrentRound.Int64()
 			qry += fmt.Sprintf(" AND activationRound <= %v AND %v < deactivationRound", currentRound, currentRound)
+		}
+
+		if len(filter.Addresses) > 0 {
+			hexAddrs := make([]string, len(filter.Addresses))
+			for i, addr := range filter.Addresses {
+				hexAddrs[i] = fmt.Sprintf("'%v'", addr.Hex())
+			}
+			qry += fmt.Sprintf(" AND ethereumAddr IN (%v)", strings.Join(hexAddrs, ", "))
 		}
 	}
 	return qry, nil
--- a/common/db.go
+++ b/common/db.go
@@ -46,7 +46,7 @@ type DBOrch struct {
 	PricePerPixel     int64
 	ActivationRound   int64
 	DeactivationRound int64
-	Stake             string
+	Stake             int64 // Stored as a fixed point number
 }
 
 // DBOrch is the type binding for a row result from the unbondingLocks table
@@ -83,7 +83,7 @@ var schema = `
 		pricePerPixel int64,
 		activationRound int64,
 		deactivationRound int64,
-		stake STRING
+		stake int64
 	);
 
 	CREATE TABLE IF NOT EXISTS unbondingLocks (
@@ -123,7 +123,7 @@ var schema = `
 	CREATE INDEX IF NOT EXISTS idx_blockheaders_number ON blockheaders(number);
 `
 
-func NewDBOrch(ethereumAddr string, serviceURI string, pricePerPixel int64, activationRound int64, deactivationRound int64, stake string) *DBOrch {
+func NewDBOrch(ethereumAddr string, serviceURI string, pricePerPixel int64, activationRound int64, deactivationRound int64, stake int64) *DBOrch {
 	return &DBOrch{
 		ServiceURI:        serviceURI,
 		EthereumAddr:      ethereumAddr,
@@ -219,7 +219,7 @@ func InitDB(dbPath string) (*DB, error) {
 		THEN orchestrators.deactivationRound
 		ELSE excluded.deactivationRound END,
 	stake = 
-		CASE WHEN excluded.stake == ""
+		CASE WHEN excluded.stake == 0
 		THEN orchestrators.stake
 		ELSE excluded.stake END 
 	`)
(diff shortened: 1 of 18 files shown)


common/db.go
```diff
@@ -46,7 +46,7 @@ type DBOrch struct {
 	PricePerPixel     int64
 	ActivationRound   int64
 	DeactivationRound int64
-	Stake             string
+	Stake             []byte
```
kyriediculous: I see very little to no benefit in changing this to be a blob, I also think strings are more ubiquitous here. 
Any specific reason for the change? 

I'd add that `BLOB` is never actually stored in the table, it's a pointer lookup so I think `STRING` is also more efficient here.

yondonfu: > Any specific reason for the change?

The main reason was to avoid an additional check to determine if the string is successfully converted into a `big.Int`.

Ex.

```
val, ok := new(big.Int).SetString(stake)
if !ok {
    // Handle error here
}
```

vs.

```
val := new(big.Int).SetBytes(stake)
```

I includes these changes here mainly because I wanted to see what the scope would be.

> I also think strings are more ubiquitous here.

What do you mean strings are more ubiquitous? In the context of the Livepeer codebase or in the context of all codebases that need to store a `uint256` in a DB? If the former, its worth noting that we currently do use the BLOB type for the ticket [faceValue and winProb](https://github.com/livepeer/go-livepeer/blob/master/common/db.go#L105) which are `uint256` values. If the latter, could you point me to the right place to see `uint256` -> string conversions for DB storage as a typical convention?

One additional annoyance with using the `string` column type seems to be that it will be mapped to one of SQLite's storage classes so you have to consider how SQLite is going to coerce values - see the section on [type affinity](https://www.sqlite.org/datatype3.html).

yondonfu:
Ready for full review.

kyriediculous:
With address filtering now being a feature for `SelectOrchs` could we add indexing for `orchestrators(ethereumAddress)` ?

common/db.go
kyriediculous: I don't think the coercing matters in this specific case when dealing with big.Ints converted to strings 

> A string might look like a floating-point literal with a decimal point and/or exponent notation but as long as the value can be expressed as an integer, the NUMERIC affinity will convert it into an integer. Hence, the string '3.0e+5' is stored in a column with NUMERIC affinity as the integer 300000, not as the floating point value 300000.0.

Anyway the performance between `TEXT` and `BLOB` does seem the same so I'm fine with using `BLOB` as it does avoid that additional check and having to validate nullstrings.

yondonfu:
> With address filtering now being a feature for `SelectOrchs` could we add indexing for `orchestrators(ethereumAddress)` ? 

The `ethereumAddr` column is already declared as a [primary key](https://github.com/livepeer/go-livepeer/blob/master/common/db.go#L79) so it should be [indexed by default](https://stackoverflow.com/questions/3379292/is-an-index-needed-for-a-primary-key-in-sqlite).

j0sh:
> Happy to discuss whether the benefit of change outweighs downsides though!

One rather large downside which is worth mentioning is that using a blob data type makes it that much harder to run non-trivial SQL queries against the staked value itself, and the data becomes difficult to read if visually inspecting result rows non-programmatically (eg, using sqlite3 CLI).

Having this type of data queryable in the DB is an under-appreciated feature, since it allows for tooling to be built alongside the data without requiring application-level integration. I used this quite often pre-Streamflow to track job and segment arrival rates, claims, etc.

How likely are we to see stakes with values exceeding 2^63 ? If so, given the current use case for staking in the DB, is there any downside to truncating some zeros off the stake value to give a close-enough approximation that fits comfortably as a sqlite numeric type, if we want to avoid the string conversion?

yondonfu:
> How likely are we to see stakes with values exceeding 2^63

If we're storing stakes denominated in LPTU (base LPT units where 1 LPT = 10 ** 18 LPTU) then it is likely that we will see stakes with values exceeding 2^64 since 2^64 = 18.446744073709553 and the highest stake on mainnet today is already > 1m.

> One rather large downside which is worth mentioning is that using a blob data type makes it that much harder to run non-trivial SQL queries against the staked value itself, and the data becomes difficult to read

Good point. I agree that using data types that make SQL queries easier would be preferable if possible. 

I see the following options:

1. Store stake denominated in LPT instead of LPTU and truncate all decimal digits. For example, if the stake is 100.78 LPT we would store the stake as 100 (basically the suggestion mentioned in the above comment)
2. Store the stake ratio (stake / total active stake) as a fixed point integer (with precision limited by whatever we decide to be the scaling factor). The ratio would be used as the weight during selection instead of the stake itself
3. Keep the current approach of storing the exact stake denominated in LPTU

Of options 1-2, I would prefer option 1 as it is the simplest and easiest to reason about IMO (i.e. stake weighted selection is based on the amount of *whole* LPT you have).

That being said, I'm not really a fan of options 1-2 because:

- They all involve a loss of precision which results in 1 additional LPTU no longer influencing selection. I think I think having 1 additional LPTU influence selection is a nice property
- If we ever want the exact stake value we wouldn't be able to rely on the DB

For these reasons, I would still prefer option 4 of keeping the current approach while accepting the tradeoff that stakes will be harder to use in non-trivial SQL queries and will be harder to read non-programatically. WDYT?

kyriediculous (approved):

j0sh:
Agreed that Option 1 is the most preferable approach for the reasons you mentioned. However:

> Store stake denominated in LPT instead of LPTU and truncate all decimal digits.

Not all digits need to be truncated. For example, we could store, for example, 10^5 digits rather than 10^18 (as a fixed point). That still gives us quite a bit of precision available for comparisons, especially since the smallest staking differences on mainnet is in the hundreds of LPT [1] (most differences are in the thousands, and this is likely to continue growing as the network inflates).

[1] As far as I can tell the smallest difference is between [0xa6a9eb2](https://explorer.livepeer.org/accounts/0xa6a9eb29e786b5233bd99c0ba28be882fe954a0e/campaign) and [0x4712e01](https://explorer.livepeer.org/accounts/0x4712e01e944802613de3a0a6d23274e7e0243015/campaign)

IIRC we already use a similar fixed-point truncation scheme to store pricing info, so there is precedent. The final call is up to you, but this kind of decision is difficult to revisit once deployed since it involves a persistent DB change (and inertia). Do we really want to make the universe of future tooling more difficult for the sake of a tiny improvement in precision? In my eyes, that's a disproportionate tradeoff.

yondonfu:
Rebased on top of #1249 

73d4e85 stores stake values as fixed point numbers
- Updates the DB column type for `stake` to int64
- Adds a helper `common.BaseTokenAmountToFixed()` to convert base token amounts (i.e. a stake value fetched from a contract) into a fixed point number. The scaling factor used is 100000 giving us 5 decimal places of precision
- Convert on-chain stake values into fixed point numbers before storing into the DB

Note: The SQL statement for updating an orch in the DB will [ignore the provided stake value](https://github.com/livepeer/go-livepeer/pull/1253/commits/73d4e850d12ba55891f07178bef88529a9c6c2d8#diff-c1db1fb0e701b4e6d43101e8240b1de7R223) if it is 0. As a result, we will be unable to set the stake in the DB to 0. This seems ok right now given that non-zero stake is a requirement to be an active orch on-chain. The only way for the on-chain stake for an orch to be 0 is if it is deactivated - we would already be filtering out these orchs via the `ActivationRound` and `DeactivationRound` values stored in the DB.

j0sh (approved):
Thanks for doing the conversion to fixed point - and a nice side effect is that it really cleans up a lot of other code to avoid passing around `big.Int` everywhere in favor of primitive int64 types. Looks great. 🚢

PR #1254, opened 2019-12-06
[O selection] Calculate latency scores and integrate new selection algo

### LINKED ISSUE #961: Create new BroadcastSession for updated OrchestratorInfo messages

Currently, B mutates the current BroadcastSession whenever it receives an updated OrchestratorInfo message from O. While this is ok for now because we're only using a BroadcastSession in a single thread, it would be preferable to try to keep BroadcastSession immutable if possible.

This is a reposted comment from @j0sh:

> I think we should be able to use the following trick without enlarging the BroadcastSessionManager API:
>
> - Create an entirely new BroadcastSession with updated fields rather than mutating an existing session (would need to copy other fields as necessary such as the OrchestratorInfo.Transcoder which is the primary lookup key within the session manager)
> - Use this new BroadcastSession in place of the old session 
> - New BroadcastSession is passed in to completeSession (at the same place it is now; no changes)
> - Within completeSession, look up the session in the map and check for physical equality (compare pointer addresses).
> - If the sessions differ, replace the existing session in the map
> - Append new session to list as we do now.
> 
> We could also do something similar via replace(newSession, oldSession) to subsume completeSession for a more explicit mechanism. This may give us some advantages if the lookup key changes (eg, Transcoder address changes due to O's internal load balancing). For robustness we probably would need to make completeSession append using the session as it exists in the map, rather than the passed-in session.
>
> This approach isn't perfect - it doesn't really prevent mutations - but at least it should make such mutations harder to do concurrently on the same session.

**Additional context**

### LINKED ISSUE #1231: Calculate latency score for each segment and attach last known latency score to session

We need to attach the last known latency score to each broadcast session so that `MinLSSelector` can select the session with the lowest latency score.

Return the transcode duration from `SubmitSegment()`. One way to do this is to return a wrapper struct around `*net.TranscodeResult` that has an extra transcode duration field.

After `SubmitSegment()` [returns](https://github.com/livepeer/go-livepeer/blob/master/server/broadcast.go#L327), if there is no error, the caller can use `seg.Duration` and the transcode duration returned to calculate the latency score for the segment that was just transcoded. 

We can create a new `BroadcastSession`, copy the existing fields from the current `BroadcastSession` that was just used to transcode a segment and attach the latency score we just calculated to this new session. Then, we can call `BroadcastSessionsManager.complete()` with this new session.

Related #961 

We can just mutate the `LatencyScore` field in `BroadcastSession`, but it would be nice to minimize mutations on `BroadcastSession` (we already mutate it to update the `OrchestratorInfo` message, but as mentioned in #961 we would prefer not to do this).

yondonfu:

This PR attaches the last known latency score to a `BroadcastSession` after each successful transcode and instantiates new `BroadcastSessionsManager` with `MinLSSelector` so that B uses the stake/latency based selection strategy by default.

- 1e2d81d: Copy the current session after each successful transcode, attach the last known latency score and update the session's `OrchestratorInfo` if needed. The copied session is then added back to `BroadcastSessionsManager` via `completeSession()`
- 2169657: Instantiate every new `BroadcastSessionsManager` with `MinLSSelector` which uses `storeStakeReader` (depends on the node's DB) to fetch the stake of Os when running stake weighted random selection

The default good enough latency score for `MinLSSelector` is set to 1.0 (i.e. real time). We may want to allow this to be configurable in the future, but setting this default value and not exposing the config option should be fine for now.

Added unit tests.

Fixes #1231 
Fixes #961 

**Checklist:**

### DIFF
--- a/server/broadcast.go
+++ b/server/broadcast.go
@@ -97,7 +97,13 @@ func (bsm *BroadcastSessionsManager) completeSession(sess *BroadcastSession) {
 	bsm.sessLock.Lock()
 	defer bsm.sessLock.Unlock()
 
-	if _, ok := bsm.sessMap[sess.OrchestratorInfo.Transcoder]; ok {
+	if existingSess, ok := bsm.sessMap[sess.OrchestratorInfo.Transcoder]; ok {
+		// If the new session and the existing session share the same key in sessMap replace
+		// the existing session with the new session
+		if existingSess != sess {
+			bsm.sessMap[sess.OrchestratorInfo.Transcoder] = sess
+		}
+
 		bsm.sel.Complete(sess)
 	}
 }
@@ -338,7 +344,7 @@ func transcodeSegment(cxn *rtmpConnection, seg *stream.HLSSegment, name string) 
 			return err
 		}
 
-		cxn.sessManager.completeSession(sess)
+		cxn.sessManager.completeSession(updateSession(sess, res))
 
 		// download transcoded segments from the transcoder
 		gotErr := false // only send one error msg per segment list
@@ -504,4 +510,35 @@ func pixels(fname string) (int64, error) {
 	}
 
 	return res.Decoded.Pixels, nil
+}
+
+// Return an updated copy of the given session using the received transcode result
+func updateSession(sess *BroadcastSession, res *ReceivedTranscodeResult) *BroadcastSession {
+	// Instead of mutating the existing session we copy it and return an updated copy
+	newSess := &BroadcastSession{}
+	*newSess = *sess
+	newSess.LatencyScore = res.LatencyScore
+
+	if res.Info == nil {
+		// Return newSess early if we do not need to update OrchestratorInfo
+		return newSess
+	}
+
+	oInfo := res.Info
+	newSess.OrchestratorInfo = oInfo
+
+	if len(oInfo.Storage) > 0 {
+		newSess.OrchestratorOS = drivers.NewSession(oInfo.Storage[0])
+	}
+
+	if newSess.Sender != nil && oInfo.TicketParams != nil {
+		// Note: We do not validate the ticket params included in the OrchestratorInfo
+		// message here. Instead, we store the ticket params with the current BroadcastSession
+		// and the next time this BroadcastSession is used, the ticket params will be validated
+		// during ticket creation in genPayment(). If ticket params validation during ticket
+		// creation fails, then this BroadcastSession will be removed
+		newSess.PMSessionID = newSess.Sender.StartSession(*pmTicketParams(oInfo.TicketParams))
+	}
+
+	return newSess
 }
--- a/server/segment_rpc.go
+++ b/server/segment_rpc.go
@@ -255,7 +255,7 @@ func verifySegCreds(orch Orchestrator, segCreds string, broadcaster ethcommon.Ad
 	return md, nil
 }
 
-func SubmitSegment(sess *BroadcastSession, seg *stream.HLSSegment, nonce uint64) (*net.TranscodeData, error) {
+func SubmitSegment(sess *BroadcastSession, seg *stream.HLSSegment, nonce uint64) (*ReceivedTranscodeResult, error) {
 	uploaded := seg.Name != "" // hijack seg.Name to convey the uploaded URI
 
 	segCreds, err := genSegCreds(sess, seg)
@@ -382,11 +382,6 @@ func SubmitSegment(sess *BroadcastSession, seg *stream.HLSSegment, nonce uint64)
 		return nil, err
 	}
 
-	// update OrchestratorInfo if necessary
-	if tr.Info != nil {
-		defer updateOrchestratorInfo(sess, tr.Info)
-	}
-
 	// check for errors and exit early if there's anything unusual
 	var tdata *net.TranscodeData
 	switch res := tr.Result.(type) {
@@ -439,24 +434,11 @@ func SubmitSegment(sess *BroadcastSession, seg *stream.HLSSegment, nonce uint64)
 
 	glog.Infof("Successfully transcoded segment nonce=%d manifestID=%s segName=%s seqNo=%d", nonce, string(sess.ManifestID), seg.Name, seg.SeqNo)
 
-	return tdata, nil
-}
-
-func updateOrchestratorInfo(sess *BroadcastSession, oInfo *net.OrchestratorInfo) {
-	sess.OrchestratorInfo = oInfo
-
-	if len(oInfo.Storage) > 0 {
-		sess.OrchestratorOS = drivers.NewSession(oInfo.Storage[0])
-	}
-
-	if sess.Sender != nil && oInfo.TicketParams != nil {
-		// Note: We do not validate the ticket params included in the OrchestratorInfo
-		// message here. Instead, we store the ticket params with the current BroadcastSession
-		// and the next time this BroadcastSession is used, the ticket params will be validated
-		// during ticket creation in genPayment(). If ticket params validation during ticket
-		// creation fails, then this BroadcastSession will be removed
-		sess.PMSessionID = sess.Sender.StartSession(*pmTicketParams(oInfo.TicketParams))
-	}
+	return &ReceivedTranscodeResult{
+		TranscodeData: tdata,
+		Info:          tr.Info,
+		LatencyScore:  transcodeDur.Seconds() / seg.Duration,
+	}, nil
 }
 
 func genSegCreds(sess *BroadcastSession, seg *stream.HLSSegment) (string, error) {
--- a/server/segment_rpc_test.go
+++ b/server/segment_rpc_test.go
@@ -1129,12 +1129,44 @@ func TestSubmitSegment_Success(t *testing.T) {
 		assert.Equal([]byte("dummy"), data)
 	}
 
-	tdata, err := SubmitSegment(s, &stream.HLSSegment{Data: []byte("dummy")}, 0)
+	noNameSeg := &stream.HLSSegment{Data: []byte("dummy")}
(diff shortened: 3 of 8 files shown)


server/selection.go
```diff
@@ -59,6 +61,28 @@ type stakeReader interface {
 	Stakes(addrs []ethcommon.Address) (map[ethcommon.Address]*big.Int, error)
 }
 
+type dbStakeReader struct {
+	db *common.DB
```
kyriediculous: we can use `common.OrchestratorStore` abstraction here.

From our discussion yesterday I take away that since this is a new type using an abstraction would make writing tests for it easier.

yondonfu: Fixed.

yondonfu:
Ready for full review.

server/selection_test.go
```diff
@@ -10,10 +10,54 @@ import (
 	"testing"
 
 	ethcommon "github.com/ethereum/go-ethereum/common"
+	"github.com/livepeer/go-livepeer/common"
 	"github.com/livepeer/go-livepeer/net"
 	"github.com/stretchr/testify/assert"
 )
 
+type stubOrchestratorStore struct {
```
kyriediculous, line 19: There already is a `stubOrchestratorStore` in `eth/watchers/stub.go` we _could_  move this into common and use the stub from there. Not saying this is the best option necessarily but it would avoid some code duplication perhaps.

server/selection.go
```diff
@@ -59,6 +61,28 @@ type stakeReader interface {
 	Stakes(addrs []ethcommon.Address) (map[ethcommon.Address]*big.Int, error)
 }
 
+type storeStakeReader struct {
+	store common.OrchestratorStore
+}
+
+func (r *storeStakeReader) Stakes(addrs []ethcommon.Address) (map[ethcommon.Address]*big.Int, error) {
+	orchs, err := r.store.SelectOrchs(&common.DBOrchFilter{Addresses: addrs})
```
kyriediculous, line 74: If I'm not mistaken we only care about active O's here so we can add a currentRound filter to `DBOrchFilter`. 

We're only caching stake for active O's anyway

kyriediculous (changes requested):
Not quite done with review yet but just some comments for now

server/selection_test.go
yondonfu: I think moving the stub into the `common` package could make sense since any package that is using the `OrchestratorStore` interface will already be importing the `common` package.

However, many of the tests in the `watchers` package currently contain assertions that directly access the internal state of `stubOrchestratorStore`. So, if we want to use a new `common.StubOrchestratorStore` type instead we would need to refactor the assertions to use a new public API method for checking the store's state. The public API of the stub would further increase in order to support stubbing errors (i.e. `SetSelectErr()` or `SetUpdateErr()`). A better approach that does not require adding custom public API methods could be to define `common.MockOrchestratorStore` that embeds `mock.Mock` from the `testify/mock` package.

I'm personally ok with having a small amount of duplication in this PR for now (we're only at a 2x duplication right now - more than 2 is when I generally think its worth being a bit more wary). The duplication is cheap. If *another* use case for the `common.OrchestratorStore` interface arises in another area of the codebase that requires a stub/mock *then* we can consider removing the duplication and importing a shared stub/mock from the `common` package. What do you think?

server/selection.go
yondonfu: I think the caller of `Stakes()` should receive the stake values for all the addresses that it provides and that it should be up to the caller to decide which addresses that it cares about (i.e. only active addresses). Otherwise, the caller would have to check which addresses it actually received stake values for in the result returned by the method. Currently, the caller of `Stakes()` is `MinLSSelector` which will only be populated with active Os when the node is running in on-chain mode. So, in this case, only active addresses will be provided to this method.

server/selection_test.go
kyriediculous: Sounds good to me !

server/broadcast.go
```diff
@@ -97,7 +97,13 @@ func (bsm *BroadcastSessionsManager) completeSession(sess *BroadcastSession) {
 	bsm.sessLock.Lock()
 	defer bsm.sessLock.Unlock()
 
-	if _, ok := bsm.sessMap[sess.OrchestratorInfo.Transcoder]; ok {
+	if existingSess, ok := bsm.sessMap[sess.OrchestratorInfo.Transcoder]; ok {
+		// If the new session and the existing session share the same key in sessMap replace
+		// the existing session with the new session
+		if existingSess != sess {
+			bsm.sessMap[sess.OrchestratorInfo.Transcoder] = sess
```
kyriediculous, line 104: we can do `*existingSess = *sess` instead

kyriediculous (changes requested):

server/broadcast.go
yondonfu: This would mutate the existing session pointer tracked by `BroadcastSessionsManager`, but the reason we call `completeSession()` with a [copy of the current session](https://github.com/livepeer/go-livepeer/pull/1254/files#diff-cecc27bfe2b6b19181ce443733dfba57R349) is to avoid mutating sessions so I think its actually preferable to keep it as is.

j0sh: As Nico indirectly brought up, it's easy to accidentally mutate, because there's nothing explicitly preventing that. Especially on the caller side.

At some point we might want to make it harder to mutate, perhaps by gating the list via an accessor, and returning a copy. This has the potential to be kind of expensive though.

It's really too bad golang doesn't have built-in immutability; the compiler and tooling could help the ergonomics, and that would facilitate a copy-on-write approach, rather than having to preemptively copy on every access. Possible that we're fighting against the grain here in terms of golang design patterns, also open to other ideas to make this more ergonomic while preserving safety.

server/segment_rpc_test.go
```diff
@@ -1233,181 +1265,56 @@ func TestSubmitSegment_Success(t *testing.T) {
...
-
-	_, err = SubmitSegment(s, &stream.HLSSegment{Data: []byte("dummy")}, 0)
-
-	assert.Nil(err)
-	assert.Equal("http://google.com", s.OrchestratorInfo.Transcoder)
-	assert.Equal(tr.Info.TicketParams.Recipient, s.OrchestratorInfo.TicketParams.Recipient)
-	assert.Equal(tr.Info.TicketParams.FaceValue, s.OrchestratorInfo.TicketParams.FaceValue)
-	assert.Equal(tr.Info.TicketParams.WinProb, s.OrchestratorInfo.TicketParams.WinProb)
-	assert.Equal(tr.Info.TicketParams.RecipientRandHash, s.OrchestratorInfo.TicketParams.RecipientRandHash)
-	assert.Equal(tr.Info.TicketParams.Seed, s.OrchestratorInfo.TicketParams.Seed)
```
j0sh, line 1309: I didn't really look at these test cases too carefully, but just checking that we're still checking the same things, and/or the items that were being checked are no longer relevant here?

Just asking because the diff is significant, but it's not immediately obvious at first glance that the coverage is the same (eg, we checked a lot of TicketParams earlier, but not anymore).

server/segment_rpc.go
```diff
@@ -439,7 +434,11 @@ func SubmitSegment(sess *BroadcastSession, seg *stream.HLSSegment, nonce uint64)
...
 
-	return tdata, nil
+	return &ReceivedTranscodeResult{
+		TranscodeData: tdata,
+		Info:          tr.Info,
+		LatencyScore:  transcodeDur.Seconds() / seg.Duration,
+	}, nil
 }
 
 func updateOrchestratorInfo(sess *BroadcastSession, oInfo *net.OrchestratorInfo) {
```
j0sh: Just noting that this is now only used in `broadcast.go` , but it doesn't really matter because it still compiles.

However, if you were to move this over to broadcast.go, one change I would make is to set the signature like so:

`func updateSession(oldSess *BroadcastSession, res *ReceivedTranscodeResult) (newSess *BroadcastSession)`

And that function can create a new session, copy old data over, update the latency score, update the O info if necessary, and return the new broadcast session. That should simplify some of the boilerplate around the `updateOrchestratorInfo` call site to something like: `cxn.sessManager.completeSession(updateSession(sess, res))`

yondonfu: Replaced `updateOrchestatorInfo()` with `updateSession()` (in `broadcast.go`) in 434f82f

server/segment_rpc_test.go
yondonfu: The `TicketParams` assertions were omitted in this PR in favor of just checking that the `OrchestratorInfo` field of a session was updated (`assert.Equal()` conveniently compares referenced values when pointers are passed in) - see [this test](https://github.com/livepeer/go-livepeer/pull/1254/files#diff-f986f75ea9a0ee09429050246648e0b7R640).

A few of the removed tests were previously included to make sure that the `OrchestratorInfo` field of a session was not updated until the very end of `SubmitSegment()`. This [test](https://github.com/livepeer/go-livepeer/pull/1254/files#diff-6a2971a2e6142892f2da0c440d15a274L1340) is one such example - we checked that the local balance update performed in `SubmitSegment()` is calculated based on the price *before* receiving a new price in an updated `OrchestratorInfo` message. Since we no longer mutate a session within `SubmitSegment()`, we don't need these types of tests anymore.

yondonfu:
424b719  - Updated the stub segment durations in the tests for latency score calculations. I saw a failure when running the test locally (every single other test run passed) - I'm guessing that some slowdown happened that caused the transcode duration for one calculation to be much different than the other calculations. Hopefully the much larger difference between stub segment durations in this commit will help.

kyriediculous (approved):

j0sh (approved):

yondonfu:
Rebased on top of #1253 

ae728a4 updates `storeStakeReader` to use int64 for stake values since #1253 stores stake values as fixed point numbers with the int64 type.

PR #1260, opened 2019-12-09
don't copy over existing binaries when running unit tests

### LINKED ISSUE #1252: SF Rinkeby Linux releases are not failing fast when connected to mainnet
**Describe the bug**

The [0.5.1](https://github.com/livepeer/go-livepeer/releases/tag/v0.5.1) Linux release is not failing fast for the following command:

```
livepeer -network mainnet -broadcaster -datadir blah
```

This was previously a problem for the Docker image as well, but is fixed by #1248 

This is not a problem for the Darwin release.

**To Reproduce**
1. Download and untar the 0.5.1 Linux release
2. Run `livepeer -network mainnet -broadcaster -datadir blah`

**Expected behavior**

All CI builds should currently be Rinkeby compatible, but not mainnet compatible. When one of these binaries is run using the flag `-network mainnet`, the node should fail fast.

kyriediculous:

This PR fixes the CircleCI build process to not overwrite the binaries when running unit tests (`./test_args.sh`). 

- Rename current `livepeer` binary to `.existing_livepeer` if it exists and rename it back to `livepeer` after tests end
- Add `.existing_livepeer` to `.gitignore` just to make the overall experience better when using tools like VSCode with git plugin.  

1. Built from source, checked that mainnet was not supported
2. Ran `./test_args.sh`
3. Checked that `./livepeer -broadcaster -network mainnet` threw the correct error

Fixes #1252 

**Checklist:**

### DIFF
--- a/test_args.sh
+++ b/test_args.sh
@@ -2,9 +2,6 @@
 
 set -eux
 
-# build the binary
-HIGHEST_CHAIN_TAG=mainnet make livepeer
-
 # set a clean slate "home dir" for testing
 TMPDIR=/tmp/livepeer-test-"$RANDOM"
 DEFAULT_DATADIR="$TMPDIR"/.lpData
@@ -12,6 +9,10 @@ CUSTOM_DATADIR="$TMPDIR"/customDatadir
 export HOME=$TMPDIR
 rm -rf "$DEFAULT_DATADIR"
 mkdir -p $TMPDIR  # goclient should make the lpData datadir
+
+# build the binary
+HIGHEST_CHAIN_TAG=mainnet
+go build -tags "$HIGHEST_CHAIN_TAG" -o $TMPDIR/livepeer cmd/livepeer/*.go
 
 # Set up ethereum key
 cat > $TMPDIR/key <<ETH_KEY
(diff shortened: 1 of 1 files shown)


mk-livepeer:
Instead of moving away `livepeer` and then moving it back, why not just use the `-o` flag to output the binary to a different directory?
```
HIGHEST_CHAIN_TAG=mainnet GO111MODULE=on CGO_LDFLAGS="$(cgo_ldflags)" go build -tags "$(HIGHEST_CHAIN_TAG)" -ldflags="$(ldflags)" -o "$(TMPDIR)" cmd/livepeer/*.go
```
...and then execute it like so:
```
$(TMPDIR)/livepeer arg1, arg2, arg3, ...
```

kyriediculous:
That should work indeed if there is no particular reason to use the `make livepeer` command here and I don't think there is as it just executed the `go build` command.

kyriediculous:
using `go build -o $TMPDIR/livepeer` in cf978b6 to output the binary as `livepeer` in `$TMPDIR`

reverted gitignore changes in dc91128

mk-livepeer (approved):
LGTM 👍

PR #1266, opened 2019-12-11
eth: use EIP155Signer

### LINKED ISSUE #1257: Use EIP155 signer in bind.TransactOpts.Signer

Contract bindings generated using `abigen` currently use the [types.HomesteadSigner](https://github.com/ethereum/go-ethereum/blob/master/accounts/abi/bind/base.go#L237) by default when signing txs. go-ethereum#17046 proposes to change the default to `types.EIP155Signer`, but based on the conversation in the thread it sounds like eventually the base code for contract bindings will not prescribe a signer and it will instead be the responsibility of the caller to define the signer used in `bind.TransactOpts.Signer`. Regardless of what decision is made, we should be signing txs based on [EIP155](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-155.md).

Pass in a `types.EIP155Signer` to [am.SignTx()](https://github.com/livepeer/go-livepeer/blob/master/eth/accountmanager.go#L128). We will need access to the current chainID.

kyriediculous:

Force all transactions to be signed with `EIP155Signer` instead of `HomesteadSigner` by overwriting the default signer passed in as an argument by the contract bindings

- Added signer argument to `backend` constructor and type (we need a signer to get the tx as message)
- Added signer argument to `accountManager` constructor and type (we need a signer to sign tx's)
- When instantiating a new `LivepeerEthClient`, fetch the `chainID`, instantiate an `EIP155Signer` and pass it as arguments to `backend` and `accountManager`

Ran unit tests, ran node making a tx

Fixes #1257 

**Checklist:**

### DIFF
--- a/eth/accountmanager.go
+++ b/eth/accountmanager.go
@@ -25,19 +25,19 @@ type AccountManager interface {
 	Unlock(passphrase string) error
 	Lock() error
 	CreateTransactOpts(gasLimit uint64, gasPrice *big.Int) (*bind.TransactOpts, error)
-	SignTx(signer types.Signer, tx *types.Transaction) (*types.Transaction, error)
+	SignTx(tx *types.Transaction) (*types.Transaction, error)
 	Sign(msg []byte) ([]byte, error)
 	Account() accounts.Account
 }
 
 type accountManager struct {
-	account accounts.Account
-
+	account  accounts.Account
+	signer   types.Signer
 	unlocked bool
 	keyStore *keystore.KeyStore
 }
 
-func NewAccountManager(accountAddr ethcommon.Address, keystoreDir string) (AccountManager, error) {
+func NewAccountManager(accountAddr ethcommon.Address, keystoreDir string, signer types.Signer) (AccountManager, error) {
 	keyStore := keystore.NewKeyStore(keystoreDir, keystore.StandardScryptN, keystore.StandardScryptP)
 
 	acctExists := keyStore.HasAddress(accountAddr)
@@ -71,6 +71,7 @@ func NewAccountManager(accountAddr ethcommon.Address, keystoreDir string) (Accou
 
 	return &accountManager{
 		account:  acct,
+		signer:   signer,
 		unlocked: false,
 		keyStore: keyStore,
 	}, nil
@@ -129,19 +130,19 @@ func (am *accountManager) CreateTransactOpts(gasLimit uint64, gasPrice *big.Int)
 				return nil, errors.New("not authorized to sign this account")
 			}
 
-			return am.SignTx(signer, tx)
+			return am.SignTx(tx)
 		},
 	}, nil
 }
 
 // Sign a transaction. Account must be unlocked
-func (am *accountManager) SignTx(signer types.Signer, tx *types.Transaction) (*types.Transaction, error) {
-	signature, err := am.keyStore.SignHash(am.account, signer.Hash(tx).Bytes())
+func (am *accountManager) SignTx(tx *types.Transaction) (*types.Transaction, error) {
+	signature, err := am.keyStore.SignHash(am.account, am.signer.Hash(tx).Bytes())
 	if err != nil {
 		return nil, err
 	}
 
-	return tx.WithSignature(signer, signature)
+	return tx.WithSignature(am.signer, signature)
 }
 
 // Sign byte array message. Account must be unlocked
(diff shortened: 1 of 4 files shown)


eth/accountmanager.go
```diff
@@ -136,12 +137,13 @@ func (am *accountManager) CreateTransactOpts(gasLimit uint64, gasPrice *big.Int)
 
 // Sign a transaction. Account must be unlocked
 func (am *accountManager) SignTx(signer types.Signer, tx *types.Transaction) (*types.Transaction, error) {
```
yondonfu: Omit the `signer` argument?

In the discussion we had earlier this week I mentioned ignoring a `signer` argument, but just to clarify I meant the `signer` argument in [bind.TransactOpts.Signer](https://github.com/livepeer/go-livepeer/blob/master/eth/accountmanager.go#L127).

yondonfu (changes requested):

eth/accountmanager.go
kyriediculous: Since every contract transaction to be signed will use that binding that calls `am.SignTx()` anyhow it seemed reasonable to implement it there instead since the binding is just a pass-through essentially to a method on our own `accountManager` implementation.

One thing however I think I _should_  do after looking through the package is make use of `MakeSigner()` which will automatically instantiate the correct signer if the supplied blockNum is greater or equal than the fork block height.

Edit: Perhaps just using `EIP155Signer` by default will be easier than using `MakeSigner` as that requires `params.ChainConfig` which are hardcoded for rinkeby and mainnet but not for e.g. local dev network.

yondonfu: > Since every contract transaction to be signed will use that binding that calls am.SignTx() anyhow it seemed reasonable to implement it there instead since the binding is just a pass-through essentially to a method on our own accountManager implementation.

We also use `SignTx()` in [ReplaceTransaction()](https://github.com/livepeer/go-livepeer/blob/master/eth/client.go#L799) so having it around still has some use.

kyriediculous: Not sure if we're on the same page , so it's okay as is? 

Or do you want to change `	SignTx(signer types.Signer, tx *types.Transaction) (*types.Transaction, error)` into `	SignTx(tx *types.Transaction) (*types.Transaction, error)` ?

yondonfu: I don't see the utility of having the `signer` param in `SignTx()` if `AccountManager` is always going to be configured with a specific `Signer` so I think omitting the `signer` param would make the API clearer especially since `SignTx()` is also used in `ReplaceTransaction()`.

kyriediculous: gotcha, fixed in f297f49a

eth/accountmanager.go
```diff
@@ -25,19 +25,19 @@ type AccountManager interface {
 	Unlock(passphrase string) error
 	Lock() error
 	CreateTransactOpts(gasLimit uint64, gasPrice *big.Int) (*bind.TransactOpts, error)
-	SignTx(signer types.Signer, tx *types.Transaction) (*types.Transaction, error)
+	SignTx(tx *types.Transaction) (*types.Transaction, error) // SignTx is configured with EIP155Signer by default
```
yondonfu: Don't think we need this comment as the caller can configure `AccountManager` with any `Signer` - we just happen to always configure it with `EIP155Signer`.

eth/accountmanager.go
```diff
@@ -129,19 +130,20 @@ func (am *accountManager) CreateTransactOpts(gasLimit uint64, gasPrice *big.Int)
...
+			return am.SignTx(tx)
 		},
 	}, nil
 }
 
 // Sign a transaction. Account must be unlocked
-func (am *accountManager) SignTx(signer types.Signer, tx *types.Transaction) (*types.Transaction, error) {
-	signature, err := am.keyStore.SignHash(am.account, signer.Hash(tx).Bytes())
+func (am *accountManager) SignTx(tx *types.Transaction) (*types.Transaction, error) {
+	// Ignore signer argument and use EIP155Signer instance instead
```
yondonfu: Don't think we need this comment for the same reason as my comment earlier in this review about the caller configuring the `Signer`.

eth/accountmanager.go
kyriediculous: fixed in aee8f653

eth/accountmanager.go
kyriediculous: fixed in aee8f653

yondonfu (commented):
Changes look good - let's rebase!

kyriediculous:
rebased !

yondonfu (approved):
LGTM! Good to merge pending green CI!

PR #1268, opened 2019-12-13
Accept Ethereum passphrase via file on disk rather than command line

### LINKED ISSUE #1256: Accept Ethereum passphrase via file on disk rather than command line
When bringing up a nodes on an actual network, `-ethPassword` can be specified on the command line to supply the Ethereum account's passphrase and allow unattended startup and operations.  This works for now, but it is very insecure.  A attacker can compromise a system, run `ps aux` and then see all command lines, including the go-livepeer command line containing the Ethereum passphrase in clear text.

We should consider adding alternate solutions such as a text file, for the future.  A text file can have permissions set to only allow the appropriate user access to it.

Both GETH and PARITY give the user a command-line option to specify a file that contains password(s) for one or more Ethereum account.  We should do the same.

I also think we should remove the cleartext passphrase on the command line option, but first thing's first.  We can start by supporting both but from a single command line argument.  If we can fstat the file, then we read it for the passphrase.  If we can't fstat it, then we assume its a cleartext passphrase.

mk-livepeer:
When bringing up nodes on an actual network, `-ethPassword` can be
specified on the command line to supply the Ethereum account's
passphrase and allow unattended startup and operations. This works
for now, but it is very insecure. A attacker can compromise a system,
run `ps aux` and then see all command lines, including the go-livepeer
command line containing the Ethereum passphrase in clear text.

This adds the ability for the user to supply a filepath rather than a
cleartext password to the `-ethPassword` parameter. If the supplied
string is a valid file path, the passphrase will be read from the file,
otherwise the supplied string will be used as-is.

closes #1256

### DIFF
--- a/common/getpass.go
+++ b/common/getpass.go
@@ -8,12 +8,19 @@ import (
 
 // GetPass attempts to read a file for a password at the supplied location.
 // If it fails, then the original supplied string will be returned to the caller.
+// A valid string will always be returned, regardless of whether an error occurred.
 func GetPass(s string, idx int) (string, error) {
 	info, err := os.Stat(s)
-	if os.IsNotExist(err) || info.IsDir() {
-		// If the supplied string is a directory or it is not a
-		// path to a file, assume it is the pass and return it
+	if os.IsNotExist(err) {
+		// If the supplied string is not a path to a file,
+		// assume it is the pass and return it
 		return s, nil
+	}
+	if os.IsNotExist(err) || info.IsDir() {
+		// If the supplied string is a directory,
+		// assume it is the pass and return it
+		// along with an approptiate error.
+		return s, fmt.Errorf("supplied path is a directory")
 	}
 	file, err := os.Open(s)
 	if err != nil {
--- a/common/getpass.go
+++ b/common/getpass.go
@@ -16,7 +16,7 @@ func GetPass(s string) (string, error) {
 		// assume it is the pass and return it
 		return s, nil
 	}
-	if os.IsNotExist(err) || info.IsDir() {
+	if info.IsDir() {
 		// If the supplied string is a directory,
 		// assume it is the pass and return it
 		// along with an approptiate error.
--- a/common/getpass.go
+++ b/common/getpass.go
@@ -9,7 +9,7 @@ import (
 // GetPass attempts to read a file for a password at the supplied location.
 // If it fails, then the original supplied string will be returned to the caller.
 // A valid string will always be returned, regardless of whether an error occurred.
-func GetPass(s string, idx int) (string, error) {
+func GetPass(s string) (string, error) {
 	info, err := os.Stat(s)
 	if os.IsNotExist(err) {
 		// If the supplied string is not a path to a file,
@@ -28,19 +28,14 @@ func GetPass(s string, idx int) (string, error) {
 	}
 	scanner := bufio.NewScanner(file)
 	scanner.Split(bufio.ScanLines)
-	var txtlines []string
 
-	for scanner.Scan() {
-		txtlines = append(txtlines, scanner.Text())
-		if len(txtlines)-idx == 1 {
-			break
-		}
-	}
+	scanner.Scan()
+	txtline := scanner.Text()
 	file.Close()
 
-	if len(txtlines) <= idx {
-		return s, fmt.Errorf("requested pass index not present in file: %s", s)
+	if len(txtline) == 0 {
+		return s, fmt.Errorf("supplied file is empty")
 	}
 
-	return txtlines[idx], nil
+	return txtline, nil
 }
--- a/common/getpass.go
+++ b/common/getpass.go
@@ -6,6 +6,8 @@ import (
 	"os"
 )
 
+// GetPass attempts to read a file for a password at the supplied location.
+// If it fails, then the original supplied string will be returned to the caller.
 func GetPass(s string, idx int) (string, error) {
 	info, err := os.Stat(s)
 	if os.IsNotExist(err) || info.IsDir() {
--- a/common/getpass.go
+++ b/common/getpass.go
@@ -1,0 +1,37 @@
+package common
+
+import (
+	"bufio"
+	"fmt"
+	"os"
+)
+
+func GetPass(s string, idx int) (string, error) {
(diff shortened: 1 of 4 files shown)


common/getpass.go
```diff
@@ -0,0 +1,46 @@
...
+import (
+	"bufio"
+	"fmt"
+	"os"
+)
+
+// GetPass attempts to read a file for a password at the supplied location.
+// If it fails, then the original supplied string will be returned to the caller.
+// A valid string will always be returned, regardless of whether an error occurred.
+func GetPass(s string, idx int) (string, error) {
```
kyriediculous: For now since we don't support unlocking multiple accounts I think we could just omit the `idx` argument and return the `0` index by default

kyriediculous: What do you think of just moving this into `common/util.go` or `eth/helpers` instead of its own file?

kyriediculous (changes requested):

common/getpass.go
mkrufky: If you think that's best, we can do that...  I put it into its own file because I knew I would be writing tests and I prefer to keep things in separate files when each function may have multiple permutations to test.

mkrufky: Since it's a utility function, I thought it would be a good idea to expose this now rather than later.  I could imagine someone running an OT node but want to use separate account for the trans....   Oh, that doesn't make sense :-P

If you don't think we'll ever need the feature, I can remove it.

mk-livepeer: removed in eeb1a3eaf56ff60f903b8884df5f77a0c3602dcc

kyriediculous: Actuallly I thought about this some more and what we should do is the following

In livepeer.go , just pass `*ethPassword` to `client.Setup` https://github.com/livepeer/go-livepeer/blob/dc878b5f938b0c7c7fdf5e69bccf3122c78e78b4/cmd/livepeer/livepeer.go#L352

So no changes have to actually be made here (except perhaps altering the `-ethPassword` flag description). 

In `eth/accountmanager.go` there is an `Unlock()` function, this is where we also prompt for the passphrase in case the `-ethPassword` flag was not provided on node startup.  This is the method that's being called in `client.Setup()` to unlock the account. 

https://github.com/livepeer/go-livepeer/blob/dc878b5f938b0c7c7fdf5e69bccf3122c78e78b4/eth/client.go#L175-L182

My suggestion is that we therefore move this new helper you created  into `eth/helpers` so that we can use it in `eth/accountManager.Unlock()`. 

In addition we can change the current default behaviour of the password prompt to also be able to take in a path (this is where we will change from how geth does is). This means that the user can always supply a password or path regardless of whether the `-ethPassword` flag is used. 

https://github.com/livepeer/go-livepeer/blob/dc878b5f938b0c7c7fdf5e69bccf3122c78e78b4/eth/accountmanager.go#L80-L103

`AccountManager` really is our specific layer on top of what go-ethereum provides so I think it makes sense to implement it there instead of `cmd` or `common`. WDYT?

kyriediculous: I don't see much reason for it in the near future in fact our `AccountManager` implementation only supports a single account as is. I'm not even sure why geth or parity expose the functionality to unlock multiple accounts, do you use multiple unlocks on a single node for mining ?

mkrufky: I use multiple accounts on all of my mining nodes, yes.  There are *many* reasons why a systems integrator may associate multiple accounts to a geth instance.  I'm not arguing it here for our case though - I removed it in eeb1a3eaf56ff60f903b8884df5f77a0c3602dcc

mkrufky: I'm not sure I agree with this -- this same function will be used for orchSecret.

mkrufky: I'm not sure I agree with this -- this same function will be used for `-orchSecret`.  With that in mind, does it still make sense to make the changes you're proposing?

kyriediculous: > I'm not sure I agree with this -- this same function will be used for orchSecret.

Good point, did not think ahead here. We can leave it in `common` then but I'd have a slight preference having it in `util.go` then, but I'll leave that decision up to you.

As for the rest of the changes described I think it's worth to implement them, we can import the helper from the `common` package then. 

We can do something like 

```
func(am *accountManager) Unlock(passphrase string) error {
	if passphrase == "" {
		glog.Infof("Please enter the passphrase or path to password file to unlock Ethereum account %v", am.account.Address.Hex())
		var err error
		passphrase, err = getPassphrase(false)
		if err != nil {
			return err
		}
	}

	pass, err := common.GetPass(passphrase)
	if err != nil {
		return err
	}

	if err := am.keyStore.Unlock(am.account, pass); err != nil  {
		return err
	}

	am.unlocked = true

	glog.Infof("Unlocked ETH account: %v", am.account.Address.Hex())

	return nil
}
```

mkrufky: Done.  Please check :-)

As far as moving `GetPass` into `common/util` , it's an anti-pattern.  It doesn't really go along with the spirit of [Idiomatic Golang](https://about.sourcegraph.com/go/idiomatic-go) ... On the other hand, neither do some of the other parts of this repository.

I hope we'll have a chance to clean up and refactor this sort of thing before too much time goes by.  As an open source project, we should strive to be more idiomatic.

Meanwhile, it's also important to stick with the style of existing code.  ...

It's clear how I feel about this, but I'll move it if that's your preference.

mkrufky: I don't think it's necessary for the interactive prompt for passphrase to be able to read from a file.  The point of this was to be able to keep the cleartext passphrase out of the command line for security purposes.

kyriediculous: > I don't think it's necessary for the interactive prompt for passphrase to be able to read from a file.

You're probably right.

Changes look good, can we clean up the commit history ? 

> As far as moving GetPass into common/util , it's an anti-pattern. It doesn't really go along with the spirit of Idiomatic Golang ... On the other hand, neither do some of the other parts of this repository.

I agree that some parts of the codebases aren't really diehard idomatic go, but as you say

> Meanwhile, it's also important to stick with the style of existing code.

mkrufky: Best practice is to squash on merge.  There is no need to flatten the commits in the PR itself.

kyriediculous (approved):

PR #1275, opened 2019-12-17
Accept orchSecret password via file on disk rather than command line

### LINKED ISSUE #1259: Accept orchSecret password via file on disk rather than command line
When bringing up a livepeer nodes, -orchSecret can be specified on the command line to supply the Orchestrator's password. This works for now, but it is insecure. A attacker can compromise a system, run `ps aux` and then see all command lines, including the go-livepeer command line containing the password in clear text.

We should consider adding (or switching to) alternate solutions such as a text file, for the future. A text file can have permissions set to only allow the appropriate user access to it.

I also think we should remove the cleartext password on the command line option, but first thing's first. We can start by supporting both but from a single command line argument. If we can fstat the file, then we read it for the password. If we can't fstat it, then we assume its a cleartext password.

mk-livepeer:
When bringing up livepeer nodes, -orchSecret can be specified on
the command line to supply the Orchestrator's password. This works
for now, but it is insecure. A attacker can compromise a system,
run ps aux and then see all command lines, including the
go-livepeer command line containing the password in clear text.

This adds the ability to read the secret from a text file. A text
file can have permissions set to only allow the appropriate user
access to it.

closes https://github.com/livepeer/go-livepeer/issues/1259

### DIFF
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -231,7 +231,7 @@ func main() {
 	}
 
 	if *orchSecret != "" {
-		n.OrchSecret = *orchSecret
+		n.OrchSecret, _ = common.GetPass(*orchSecret)
 	}
 
 	if *transcoder {


kyriediculous (approved):
LGTM 👍

Issue #1279, opened 2019-12-20
Spec: HTTP segment ingest return transcoded data in the same request


darkdarkdragon:
## Abstract

Return transcoded data in same POST request.

## Motivation

Currently, handler for HTTP ingest POST endpoint only returns status code, without returning any data. This forces clients to do another HTTP requests to the Broadcaster to receive transcoded data, one request per transcoding profile. This adds unnecessary latency and complicates client's code.

## Proposed Solution

Return transcoded data using `multipart/mixed` encoding. Return it only if client sends `Accept: multipart/mixed header` header.

## Implementation Tasks and Considerations

Borrow implementation from orchestrator. Also, if transcoded segments stored in external object store, instead of actual data return URLs to segments (distinguished by using `application/vnd+livepeer.uri` mime type).

Thing to consider - maybe we should allow client to force to always return URLs instead of returning actual data?  In case client do not want actual data, but, say, want to construct playlist for himself and pass it somewhere?

j0sh:
> Also, if transcoded segments stored in external object store, instead of actual data return URLs to segments (distinguished by using application/vnd+livepeer.uri mime type).

👍  Note that implementing clients would need to be aware of this distinction, but I agree it's a useful feature to have.

>  maybe we should allow client to force to always return URLs instead of returning actual data? 

Yes - maybe this could be signaled by the client requesting a different (more explicit?) content type in the Accept header:
* Maybe something like `multipart/mixed/livepeer.uri`
* Separate JSON response type if the client requests URL-only, again via content-type, eg `application/json`.

For an external OS, the broadcaster will be able to return a full / absolute URL in the response. However, if an internal OS is being used, along with a URL-only return format, the only a relative URL can be returned (since the B doesn't always know the full URL it's accessible at), and clients again need to be prepared for it.

All being said, this particular bit can probably be saved for later.

Issue #1003, opened 2019-07-23
HTTP segment ingest: Add multipart response types


angyangie:
Original discussion here: https://github.com/livepeer/go-livepeer/pull/972

(Josh) Responses: Currently, nothing is returned in response. A [multipart/mixed implementation ](https://github.com/livepeer/go-livepeer/blob/3b6a1918a646383f4413b8a3a8f3469d9fa27818/server/ot_rpc.go#L145-L160) can be borrowed from the standalone transcoder.

(Eli) Let's allow for both no-return and multipart/mixed implementations as determined by the presence or absence of the Accept: multipart/mixed header? I think they both have valid use cases.

(Josh) Haven't checked yet but is a response like 200 OK even returned right now? If not, should it be?

I like using the Accept header to signal the format of the response body. Part of me wonders if we should make the default a bit more user-friendly, like a list of transcoded URLs or even a zip/tar file with the results (would be a super cool feature in conjunction with "Non-Segmented Transcodes") but that's something we can figure out later.

(Angie) Why do we need something to be returned in responses here, to require a multipart/mixed implementation? Which response are we talking about? I see a comment on returning "a list of transcoded URLs or even a zip/tar file with the results" ... do we agree what should be returned?

(Eli) I think there is not consensus on the point. I would be entirely comfortable shipping this PR without any return type at all — as discussed [here](https://github.com/livepeer/go-livepeer/pull/972#issuecomment-510196290), any future return types can be added through the Accept: header.

iameli:
Some feedback from Jaron at MistServer that I 

>I'd say [not returning the HTTP response until the transcoding is complete is] actually a really useful feature: if the endpoint doesn't respond to the POST until the transcode is complete, we can use that fact to not have to continuously poll the playlist for updates, and get the output as soon as it's ready with perfect timing.
> It sounds like you're planning to change the behaviour so the return is immediate; may I suggest making it an option instead? I'd like to be able to rely on this behaviour to optimize timing. 😁
> Or, even better: perhaps have the endpoint respond with either the transcoded data itself, or a redirect towards it? HTTP status code 300 "Multiple choices" sounds like a great candidate for this when encoding multiple profiles at once. (Though we'd be perfectly fine working with either all qualities concatenated after each other or having them in a single TS mux) I'm not sure if all that's feasible for you or not, but it would greatly simplify integration I'd say.

`300 Multiple Choices` is a cool idea for what we could return.

j0sh:
[...]

Issue #1100, opened 2019-09-21
Policy Implementation


j0sh:
Currently, the [transcode loop](https://github.com/livepeer/go-livepeer/blob/78142f04ff58d2aaac765ebb7c18ff497ee5d06f/server/broadcast.go#L279-L304) retries a segment until one of the following "exit conditions" is met:
* It runs out of orchestrators
* A non-error value is returned from the transcode function.

With policy-based enforcement, we may see the set of "exit conditions" be expanded to accommodate various policy parameters, and to have policy-related code commingled with various parts of the transcode logic. While this is in a sense more "straightforward" to read through, the broadcast-related code is already fairly sprawling, and it does not seem desirable to increase that complexity further with increased branchiness around the details of policy implementation in various spots.

A design decision here is to preserve the existing flow with minimal changes. This implies keeping any policy and verification-related bookkeeping entirely self-contained. That way, a policy-driven verifier can influence the transcoding flow only via error codes, with some light handling of particular return values at the verifier call site. This also facilitates better isolation of policy-related code for off-chain use.

The goal of keeping the verifier and policy parts self-contained suggests the following signature for the primary entry point to the verifier:
```

// Call NewSegmentVerifier to instantiate a 
// fresh verifier policy for a given segment.
NewSegmentVerifier(p Policy) Verifier

type Verifier interface {
  Verify(params *VerificationParams) (ret *VerificationParams, err error)
}
```

Since policy enforcement will entail a certain amount of bookkeeping for each segment (tracking retries, sorting results, etc), a fresh verifier should be instantiated for each segment. `NewSegmentVerifier` is the proposed name of this function to instantiate fresh verifiers.

The semantics of the `Verify` function are as following:

* `params` holds all the parameters for the current set of renditions, including information on the source segment.
 
Currently, there are a lot of [parameters](https://github.com/livepeer/go-livepeer/pull/1096/files#diff-98d66d2dc5372dedc76d5fdde7bedebcR11-R26) so it is easier to put them into a struct. Moreover, different verifiers may use different subsets of parameter fields without changing the signature.

* `ret` indicates the parameter set that has been accepted by the verifier, if any. A nil return value indicates the given parameters were not immediately accepted, while a non-nil value indicates an accepted set of parameters.
 
The returned set of parameters will not necessarily match what was just passed into `params`. For example, hitting the retry limit with `params` could result in returning a different set of parameters for `ret` if an earlier set resulted in a better score.

[...]

PR #1280, opened 2019-12-20
Return transcoded segments as multipart/mixed in HTTP ingest

darkdarkdragon:
Returns transcoded data as using `multipart/mixed` encoding.

Changed HTTP push handler to return.

Will write unit test in next commits.

Refs #1003 
Refs #1279 

**Checklist:**

### DIFF
--- a/server/broadcast.go
+++ b/server/broadcast.go
@@ -313,181 +313,179 @@ func transcodeSegment(cxn *rtmpConnection, seg *stream.HLSSegment, name string,
 		// similar to the orchestrator's RemoteTranscoderFatalError
 		return nil, nil
 	}
-	{
-		glog.Infof("Trying to transcode segment nonce=%d seqNo=%d", nonce, seg.SeqNo)
-		if monitor.Enabled {
-			monitor.TranscodeTry(nonce, seg.SeqNo)
-		}
+	glog.Infof("Trying to transcode segment nonce=%d seqNo=%d", nonce, seg.SeqNo)
+	if monitor.Enabled {
+		monitor.TranscodeTry(nonce, seg.SeqNo)
+	}
 
-		// storage the orchestrator prefers
-		if ios := sess.OrchestratorOS; ios != nil {
-			// XXX handle case when orch expects direct upload
-			uri, err := ios.SaveData(name, seg.Data)
-			if err != nil {
-				glog.Errorf("Error saving segment to OS nonce=%d seqNo=%d: %v", nonce, seg.SeqNo, err)
-				if monitor.Enabled {
-					monitor.SegmentUploadFailed(nonce, seg.SeqNo, monitor.SegmentUploadErrorOS, err.Error(), false)
-				}
-				cxn.sessManager.removeSession(sess)
-				return nil, err
+	// storage the orchestrator prefers
+	if ios := sess.OrchestratorOS; ios != nil {
+		// XXX handle case when orch expects direct upload
+		uri, err := ios.SaveData(name, seg.Data)
+		if err != nil {
+			glog.Errorf("Error saving segment to OS nonce=%d seqNo=%d: %v", nonce, seg.SeqNo, err)
+			if monitor.Enabled {
+				monitor.SegmentUploadFailed(nonce, seg.SeqNo, monitor.SegmentUploadErrorOS, err.Error(), false)
 			}
-			seg.Name = uri // hijack seg.Name to convey the uploaded URI
-		}
-
-		// send segment to the orchestrator
-		glog.V(common.DEBUG).Infof("Submitting segment nonce=%d manifestID=%s seqNo=%d orch=%s", nonce, cxn.mid, seg.SeqNo, sess.OrchestratorInfo.Transcoder)
-
-		res, err := SubmitSegment(sess, seg, nonce)
-		if err != nil || res == nil {
 			cxn.sessManager.removeSession(sess)
-			if res == nil && err == nil {
-				return nil, errors.New("Empty response")
-			}
 			return nil, err
 		}
-
-		cxn.sessManager.completeSession(updateSession(sess, res))
-
-		// download transcoded segments from the transcoder
-		gotErr := false // only send one error msg per segment list
-		var errCode monitor.SegmentTranscodeError
-		errFunc := func(subType monitor.SegmentTranscodeError, url string, err error) {
-			glog.Errorf("%v error with segment nonce=%d seqNo=%d: %v (URL: %v)", subType, nonce, seg.SeqNo, err, url)
-			if monitor.Enabled && !gotErr {
-				monitor.SegmentTranscodeFailed(subType, nonce, seg.SeqNo, err, false)
-				gotErr = true
-				errCode = subType
-			}
-		}
-
-		var dlErr, saveErr error
-		segHashes := make([][]byte, len(res.Segments))
-		n := len(res.Segments)
-		segURLs := make([]string, len(res.Segments))
-		segHashLock := &sync.Mutex{}
-		cond := sync.NewCond(segHashLock)
-
-		dlFunc := func(url string, pixels int64, i int) {
-			defer func() {
-				cond.L.Lock()
-				n--
-				if n == 0 {
-					cond.Signal()
-				}
-				cond.L.Unlock()
-			}()
-
-			if bos := sess.BroadcasterOS; bos != nil && !drivers.IsOwnExternal(url) {
-				data, err := drivers.GetSegmentData(url)
-				if err != nil {
-					errFunc(monitor.SegmentTranscodeErrorDownload, url, err)
-					segHashLock.Lock()
-					dlErr = err
-					segHashLock.Unlock()
-					cxn.sessManager.removeSession(sess)
-					return
-				}
-				name := fmt.Sprintf("%s/%d.ts", sess.Profiles[i].Name, seg.SeqNo)
-				newURL, err := bos.SaveData(name, data)
-				if err != nil {
-					segHashLock.Lock()
-					saveErr = err
-					segHashLock.Unlock()
-					switch err.Error() {
-					case "Session ended":
-						errFunc(monitor.SegmentTranscodeErrorSessionEnded, url, err)
-					default:
-						errFunc(monitor.SegmentTranscodeErrorSaveData, url, err)
-					}
-					return
-				}
-				url = newURL
-
-				hash := crypto.Keccak256(data)
-				segHashLock.Lock()
-				segHashes[i] = hash
-				segHashLock.Unlock()
-			}
-
-			// If running in on-chain mode, run pixels verification asynchronously
-			// Only run if a verifier is not being used
-			if sess.Sender != nil && verifier == nil {
-				go func() {
-					if err := verifyPixels(url, sess.BroadcasterOS, pixels); err != nil {
-						glog.Error(err)
-						cxn.sessManager.removeSession(sess)
-					}
-				}()
-			}
-
-			// Store URLs for the verifier. Be aware that the segment is
-			// already within object storage  at this point, whether local or
-			// external. If a client were to ignore the playlist and
-			// preemptively fetch segments, they could be reading tampered
-			// data. Not an issue if the delivery protocol is being obeyed.
-			segHashLock.Lock()
-			segURLs[i] = url
-			segHashLock.Unlock()
-
-			if monitor.Enabled {
-				monitor.TranscodedSegmentAppeared(nonce, seg.SeqNo, sess.Profiles[i].Name)
-			}
-		}
-
-		for i, v := range res.Segments {
-			go dlFunc(v.Url, v.Pixels, i)
-		}
-
-		cond.L.Lock()
-		for n != 0 {
-			cond.Wait()
-		}
-		cond.L.Unlock()
-		if dlErr != nil {
-			return nil, dlErr
-		}
-
-		if verifier != nil {
-			// verify potentially can change content of segURLs
-			err := verify(verifier, cxn, sess, seg, res.TranscodeData, segURLs)
-			if err != nil {
-				glog.Errorf("Error verifying nonce=%d manifestID=%s seqNo=%d err=%s", nonce, cxn.mid, seg.SeqNo, err)
-				return nil, err
-			}
-		}
-
-		for i, url := range segURLs {
-			err = cpl.InsertHLSSegment(&sess.Profiles[i], seg.SeqNo, url, seg.Duration)
-			if err != nil {
-				// InsertHLSSegment only returns ErrSegmentAlreadyExists error
-				// Right now InsertHLSSegment call is atomic regarding transcoded segments - we either inserting
-				// all the transcoded segments or none, so we shouldn't hit this error
-				// But report in case that InsertHLSSegment changed or something wrong is going on in other parts of workflow
-				glog.Errorf("Playlist insertion error nonce=%d manifestID=%s seqNo=%d err=%s", nonce, cxn.mid, seg.SeqNo, err)
-				if monitor.Enabled {
-					monitor.SegmentTranscodeFailed(monitor.SegmentTranscodeErrorPlaylist, nonce, seg.SeqNo, err, false)
-				}
-			}
-		}
-
-		ticketParams := sess.OrchestratorInfo.GetTicketParams()
-		if ticketParams != nil && // may be nil in offchain mode
-			saveErr == nil && // save error leads to early exit before sighash computation
-			// Might not have seg hashes if results are directly uploaded to the broadcaster's OS
-			// TODO: Consider downloading the results to generate seg hashes if results are directly uploaded to the broadcaster's OS
-			len(segHashes) != len(res.Segments) &&
-			!pm.VerifySig(ethcommon.BytesToAddress(ticketParams.Recipient), crypto.Keccak256(segHashes...), res.Sig) {
-			glog.Errorf("Sig check failed for segment nonce=%d seqNo=%d", nonce, seg.SeqNo)
-			cxn.sessManager.removeSession(sess)
-			return nil, errPMCheckFailed
-		}
-		if monitor.Enabled {
-			monitor.SegmentFullyTranscoded(nonce, seg.SeqNo, common.ProfilesNames(sess.Profiles), errCode)
-		}
-
-		glog.V(common.DEBUG).Infof("Successfully validated segment nonce=%d seqNo=%d", nonce, seg.SeqNo)
-		return segURLs, nil
+		seg.Name = uri // hijack seg.Name to convey the uploaded URI
 	}
+
+	// send segment to the orchestrator
+	glog.V(common.DEBUG).Infof("Submitting segment nonce=%d manifestID=%s seqNo=%d orch=%s", nonce, cxn.mid, seg.SeqNo, sess.OrchestratorInfo.Transcoder)
+
+	res, err := SubmitSegment(sess, seg, nonce)
+	if err != nil || res == nil {
+		cxn.sessManager.removeSession(sess)
+		if res == nil && err == nil {
+			return nil, errors.New("Empty response")
+		}
+		return nil, err
+	}
+
+	cxn.sessManager.completeSession(updateSession(sess, res))
+
+	// download transcoded segments from the transcoder
+	gotErr := false // only send one error msg per segment list
+	var errCode monitor.SegmentTranscodeError
+	errFunc := func(subType monitor.SegmentTranscodeError, url string, err error) {
+		glog.Errorf("%v error with segment nonce=%d seqNo=%d: %v (URL: %v)", subType, nonce, seg.SeqNo, err, url)
+		if monitor.Enabled && !gotErr {
+			monitor.SegmentTranscodeFailed(subType, nonce, seg.SeqNo, err, false)
+			gotErr = true
+			errCode = subType
+		}
+	}
+
+	var dlErr, saveErr error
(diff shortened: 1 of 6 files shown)


darkdarkdragon:
@j0sh This one is ready for review

server/mediaserver.go
```diff
@@ -653,19 +657,56 @@ func (s *LivepeerServer) HandlePush(w http.ResponseWriter, r *http.Request) {
...
-	err = processSegment(cxn, seg)
+	tr, err := processSegment(cxn, seg)
 	if err != nil {
 		// TODO return error
 		http.Error(w, err.Error(), http.StatusInternalServerError)
 		return
 	}
 
+	boundary := common.RandName()
+	accept := r.Header.Get("Accept")
```
iameli, line 702: We should probably error out if we see anything other than `multipart/mixed` in this header

server/broadcast.go
```diff
@@ -434,44 +427,42 @@ func transcodeSegment(cxn *rtmpConnection, seg *stream.HLSSegment, name string,
...
 			segHashLock.Unlock()
 
 			if monitor.Enabled {
 				monitor.TranscodedSegmentAppeared(nonce, seg.SeqNo, sess.Profiles[i].Name)
 			}
 		}
 
+		wg.Add(len(res.Segments))
 		for i, v := range res.Segments {
+			tr.url[i] = v.Url
```
j0sh: We probably don't want to do this here, because this is usually going to be the orchestrator's URL, and we don't want to run the risk of returning that if something goes wrong with the download.

server/broadcast.go
```diff
@@ -370,20 +367,15 @@ func transcodeSegment(cxn *rtmpConnection, seg *stream.HLSSegment, name string,
...
 		segHashes := make([][]byte, len(res.Segments))
-		segURLs := make([]string, len(res.Segments))
-		n := len(res.Segments)
+		tr := &transcodeResult{
+			data: make([][]byte, len(res.Segments)),
+			url:  make([]string, len(res.Segments)),
+		}
 		segHashLock := &sync.Mutex{}
-		cond := sync.NewCond(segHashLock)
+		wg := &sync.WaitGroup{}
```
j0sh: This cleanup here is generally good and overdue. But since this change (and a few others in this PR) are bascially unrelated to the main goal here, here is how I would have structured the set of commits:

* Separate commit changing the condvar to use WaitGroup. And ideally, adding tests: IIRC this bit of code doesn't directly have tests (it does have some indirect coverage via other tests), and the friction of adding tests was *the* reason that I haven't made this particular change yet.

Note that if we wanted to *really* improve the code here, we'd be returning a `data` / `url` / `err` via channel, and count the number of channel responses we receive. No need for any mutexes or WaitGroups. Example of this: https://github.com/livepeer/go-livepeer/blob/master/discovery/db_discovery.go#L171-L204

* Separate commit removing the usage of the `shouldStopStream` / `shouldStopSession` functions.

* Separate commit adding a return value to `transcodeSession`

* And finally, a separate commit for the multipart return making use of the `transcodeSession` return value.

Generally if a certain set of changes would make sense independently, it's best to keep them as a separate commit. Having separate commits helps in review, especially historical review, helps bisect should we ever need to, and could even be merged separately if the main PR gets held up for whatever reason. (I'm OK with having a few of those commits within the same PR, since this type of cleanup / refactoring is an encouraged part of the process of adding a new feature.)

server/mediaserver.go
```diff
@@ -673,19 +677,60 @@ func (s *LivepeerServer) HandlePush(w http.ResponseWriter, r *http.Request) {
...
+	if accept == "multipart/mixed" {
+		mw := multipart.NewWriter(w)
+		for i, url := range tr.url {
+			mw.SetBoundary(boundary)
+			data := tr.data[i]
+			var hdrs textproto.MIMEHeader
+			if len(data) > 0 {
+				hdrs = textproto.MIMEHeader{
+					"Content-Type":   {"video/MP2T"},
+					"Content-Length": {strconv.Itoa(len(data))},
```
j0sh: Might be good to add a `Profile-Name` header here as well in case we stop returning things in-order

server/mediaserver.go
```diff
@@ -673,19 +677,60 @@ func (s *LivepeerServer) HandlePush(w http.ResponseWriter, r *http.Request) {
...
+		w.Header().Set("Content-Type", contentType)
+	}
 	w.WriteHeader(http.StatusOK)
+	w.(http.Flusher).Flush()
+	if accept == "multipart/mixed" {
+		mw := multipart.NewWriter(w)
+		for i, url := range tr.url {
+			mw.SetBoundary(boundary)
+			data := tr.data[i]
+			var hdrs textproto.MIMEHeader
```
j0sh: This might read a little better as:

```golang
type, length := "video/MP2T", len(tr.data[i])
if length > 0 {
  type, length = "application/vnd+livepeer.uri", strconv.Itoa(len(tr.url[i]))
}
hdrs := textproto.MIMEHeader{
  "Content-Type":{type},
  "Content-Length":{length},
  "Profile-Name":{tr.profiles[i]},
}
```

server/broadcast.go
```diff
@@ -289,17 +289,21 @@ func processSegment(cxn *rtmpConnection, seg *stream.HLSSegment) error {
...
+		if tr, err := transcodeSegment(cxn, seg, name, sv); err == nil {
+			return tr, nil
 		}
 	}
 }
 
+type transcodeResult struct {
+	data [][]byte
+	url  []string
+}
```
j0sh: Is there a reason for doing a column-major structure here rather than a row-major structure such as the following:

```
type transcodeResult struct {
  data []byte
  url string
}
func transcodeSegment(...)([]transcodeResult, error)
```

server/mediaserver.go
```diff
@@ -673,19 +677,60 @@ func (s *LivepeerServer) HandlePush(w http.ResponseWriter, r *http.Request) {
 	}
 
 	// Do the transcoding!
-	err = processSegment(cxn, seg)
+	tr, err := processSegment(cxn, seg)
 	if err != nil {
 		// TODO return error
```
j0sh: I guess this TODO can be removed now?

server/broadcast.go
```diff
@@ -434,44 +427,42 @@ func transcodeSegment(cxn *rtmpConnection, seg *stream.HLSSegment, name string,
...
+			return nil, dlErr
 		}
 
 		if verifier != nil {
-			err := verify(verifier, cxn, sess, seg, res.TranscodeData, segURLs)
+			err := verify(verifier, cxn, sess, seg, res.TranscodeData, tr.url)
 			if err != nil {
 				glog.Errorf("Error verifying nonce=%d manifestID=%s seqNo=%d err=%s", nonce, cxn.mid, seg.SeqNo, err)
-				return err
+				return tr, err
```
j0sh: Is there a reason we return `tr` here but not for other errors?

server/broadcast.go
```diff
@@ -434,44 +427,42 @@ func transcodeSegment(cxn *rtmpConnection, seg *stream.HLSSegment, name string,
...
+		for i, url := range tr.url {
 			err = cpl.InsertHLSSegment(&sess.Profiles[i], seg.SeqNo, url, seg.Duration)
 			if err != nil {
 				// We assume playlist insertion is *not* recoverable for now
 				glog.Errorf("Playlist insertion error nonce=%d manifestID=%s seqNo=%d err=%s", nonce, cxn.mid, seg.SeqNo, err)
 				if monitor.Enabled {
 					monitor.SegmentTranscodeFailed(monitor.SegmentTranscodeErrorPlaylist, nonce, seg.SeqNo, err, false)
 				}
-				return nil
+				return tr, nil
```
j0sh: Is there a reason we return tr here but not for other errors?

server/mediaserver.go
```diff
@@ -673,19 +677,60 @@ func (s *LivepeerServer) HandlePush(w http.ResponseWriter, r *http.Request) {
...
 
+	boundary := common.RandName()
+	accept := r.Header.Get("Accept")
+	if accept == "multipart/mixed" {
+		contentType := "multipart/mixed; boundary=" + boundary
+		w.Header().Set("Content-Type", contentType)
+	}
 	w.WriteHeader(http.StatusOK)
+	w.(http.Flusher).Flush()
+	if accept == "multipart/mixed" {
```
j0sh: The level of indentation here is getting kind of deep, we could probably save one level of indentation if we return from the function if not multipart, eg

```golang
if accept != "multipart/mixed" {
  return
}
mw := multipart.NewWriter(w)
...
```

(If we ever add another content type then we can factor things out, and the nice thing is that would also lead to a cleaner diff since it likely wouldn't involve re-indentation if the refactoring meant putting each output format into its own function. But the sketch above is OK for now.)

server/push_test.go
```diff
@@ -27,19 +34,154 @@ func requestSetup(s *LivepeerServer) (http.Handler, *strings.Reader, *httptest.R
 	return handler, reader, writer
 }
 
-func TestNoErrors(t *testing.T) {
```
j0sh, line 30: Don't we also want to test non-multipart returns? Isn't that test case being removed here?

j0sh (changes requested):

server/broadcast.go
darkdarkdragon: Initially there was no reason, but now, changing to `[]transcodeResult` will require maintaining separate list of URLs (`segURLs`)

server/mediaserver.go
```diff
@@ -673,19 +677,60 @@ func (s *LivepeerServer) HandlePush(w http.ResponseWriter, r *http.Request) {
...
+		mw := multipart.NewWriter(w)
+		for i, url := range tr.url {
+			mw.SetBoundary(boundary)
+			data := tr.data[i]
```
j0sh: So if verification is used, I think it's possible that `tr.data` could reflect data that was *not* accepted for verification, if all retries result in false negatives and a "higher scoring" result was used. Something like this:

1. Segment_O1 transcoded by O1
2. Verify Segment_O1, verification fails with score -0.5
3. Segment_O2 transcoded by O2
4. Verify Segment_O2, verification fails with score -0.1
5. Segment_O3 transcoded by O3
6. Verify Segment_O3, verification fails with score -0.4
7. Max retries hit, verifier returns Segment_O2 as the "highest scoring" result

However, in this case, `tr.data` will be populated with results from O3. This behavior should be confirmed with a new unit test for the verifier, similar to `TestVerifier_HLSInsertion`. 

Now, there are a few ways to fix this:

1. Pass in the `transcodeResults` into the verifier, and reset `tr.data` in addition to `tr.urls` . This is "easy" but I'm a little skeptical of doing so much bookkeeping on state that is canonically available elsewhere, as in the following...

2. Only return URLs from the `transcodeSegment` function - the verifier already handles resetting those URLs as necessary for HLS segment insertion. Within the multipart writer, we can use the URL and `MemorySession.GetData` if it's not an external URL.

In fact, we don't really need to return anything from the `transcodeSegment` function here; we can reconstruct the URL from the profile name (which is already available) and the segment number. This seems like the safest way to ensure consistency and and requires the fewest changes to `transcodeSegment`, which would be preferable.

server/broadcast.go
darkdarkdragon: Actually, after thinking about this more, I think we should just ignore error returned from `InsertHLSSegment`. Only error it can return is `segment already exists`. Previously `InsertHLSSegment` was called right after downloading segment, independently for each segment. So, in case when one rendition was successfully downloaded and inserted into playlist, but download of second rendition resulted in error and caused to re-try transcoding, first rendition was inserted into playlist again and was returning error. But we was ignoring it anyway.
I'll remove check for this error.

server/push_test.go
darkdarkdragon: This case is handled in `TestMultipartReturn` test under `// Body should be empty if no Accept header specified` comment. And `TestNoErrors`, as it is, was not working anymore - so I've just removed it.

j0sh: > And TestNoErrors, as it is, was not working anymore - so I've just removed it.

What made it stop working? Surely existing behavior shouldn't be changed?

server/broadcast.go
j0sh: Probably OK for now. I wonder if we'll ever have other types of errors, but agreed that how it currently works makes things a bit hard to handle. Maybe add a comment explaining the behavior during retries.

This would be another unrelated change, so best to put that in a separate commit prior to the change to `transcodeSegment`.

server/push_test.go
darkdarkdragon: This addition:
```golang
	if tr == nil {
		http.Error(w, "No sessions available", http.StatusServiceUnavailable)
		return
	}
```
made it to fail. Previously there was no indication for caller that segment actually was not processed. Don't think we should preserve old behaviour.

server/broadcast.go
darkdarkdragon: > Maybe add a comment explaining the behavior during retries.

But after your change there is no special behaviour - insert into playlist is atomic regarding each segment - either every segments are transcoded and inserted, or none - so we really shouldn't hit this error.
But I can can check for this error and report it to log and grafana, but just ignore - this way, if we will hit it, we will know that something is deeply wrong, probably in other part of the stack.

server/mediaserver.go
darkdarkdragon: > we can reconstruct the URL from the profile name (which is already available) 

This dependence on URL structure feels very fragile. All the state manipulation in `verify` already very hacky and hard to follow, adding URL manipulation to `HandlePush` instead of just passing data around will make this worse.
I think, after verifier addition, we should change overall data to look like this:
1. `transcodeSegment` downloads all the transcoded segments.
2. passes downloaded data to verifier
3. and only if it returns success, then saves data to the B's OS. Anyway verification is synchronous, requires all the data to be present and transcoding doesn't considered finished until verification is done.

what do you think?

j0sh: >This dependence on URL structure feels very fragile

We construct the URL in a number of other places though, so it's a pretty safe assumption to make. Could even make it a helper if that makes you feel better. Most importantly to me, it minimizes the scope of changes and hence the reasoning about state that we have to do.

> All the state manipulation in verify already very hacky and hard to follow,

I agree that it's not great, but any changes we make would just mean doing the state manipulation in a slightly different place.

Note that I'm not too happy that HLS insertion is synchronous either - the old approach was nicer in terms of making segments immediately available - but the alternative was to maintain two different code paths for HLS insertion which is worse in terms of maintainability. This was discussed in the verification PR.

> passes downloaded data to verifier and only if it returns success, then saves data to the B's OS..

I feel like we need to be careful about this flow, because it's possible for B and O to both have their own OS, and for O to not upload to B's own OS. Currently, the B handles this situation fine. However, with your approach, we'd end up with a situation where:

* O uploads results to O's OS
* B downloads results from O's OS
* B verifies results (presumably incurring additional transfer through the B)
* B uploads results to B's OS, synchronously

This may have some other unintended side effects as well, such as not using the B's external OS for the verifier. Overall it feels like a fraught change.

Anyway, we can continue to discuss this type of refactoring, but it's really not related to the goals of this PR, so let's keep the changes here focused on what needs to happen for multipart return.

server/broadcast.go
j0sh: That makes sense - would be good to add everything you described in a comment explaining why we don't return from the error, because it may seem odd at first glance.

server/mediaserver.go
darkdarkdragon: > We construct the URL in a number of other places though, so it's a pretty safe assumption to make. Could even make it a helper if that makes you feel better.

Removing reliance on URL composition will make me feel better, not adding more of it.

>  it minimizes the scope of changes

This position makes our worse, with every change.

>  minimizes the scope of changes and hence the reasoning about state that we have to do.

It increases reasoning about state need to be done. 

> I agree that it's not great, but any changes we make would just mean doing the state manipulation in a slightly different place

Right now we're doing unnecessary manipulations - putting data into OS storage (though generic interface), taking it out (using casting!), putting it back - it is unnecessary manipulations, we can decrease number of manipulations and make data flow more straightforward and understandable.

> Currently, the B handles this situation fine. 

How?
- Broadcaster download segment from O's OS
- B uploads segment to own storage 
- B passes URLs to verifier
- Verifier rejects, B transcodes segment again
- B downloads new transcoded segment
- B uploads them to own storage, overwriting old ones
- B passes URLs to verifier 
- Verifier wants to approve segments from the previous round, so it returns older URLs to B
- But that URL are same, data in B's own OS is already overwritten

What I'm missing?


> B verifies results (presumably incurring additional transfer through the B)

Presumably OS will be external - like S3, but verifier will be in internal network, so this would be better?

j0sh: Agree the data flow here could be improved. At this point though, it's a high-risk change in my book because it'll be modifying behavior in scenarios unrelated  to this PR. This is where the "reasoning about state" comes from. Minimizing that risk means avoiding changing anything in how we handle state right now.

The most expedient thing here would be to not change anything about how we handle the `transcodeSegment` function. That keeps this PR scoped strictly to multipart return, which doesn't *need* any other changes outside of the HTTP push handler. We can revisit the data flow when it's not squeezed in with an otherwise unrelated feature.

> B uploads them to own storage, overwriting old ones
> What I'm missing?

The upload goes to a different location precisely to avoid this problem. See [1] . Also, if I understand your proposed approach correctly (eg, downloading the data to a local var rather than pushing directly into an OS immediately) then at the it would not be possible to use remote verifiers, because how else would the verifier receive segments [2]?

Anyway - I agree the state manipulation within the verifier is a bit ugly. But it was done that way for a reason: to keep the verification as self-contained as possible without infiltrating the rest of the broadcaster code, and to minimize the impact to the surrounding code [3]. This is also something we should strive for in general, this feature included.

[1] https://github.com/livepeer/go-livepeer/blob/f41ad9982d6254f1cf217c6834b15c0a02451521/server/broadcast.go#L226-L230

[2]  This admittedly only happens under scenarios where we either have a bug, or the orchestrator doesn't respect the B's upload preferences. But defense in depth says these are scenarios we shouldn't discount if possible. In fact, the initial OS implementation had a bug with exactly this problem.

[3] Policy implementation outline: https://github.com/livepeer/go-livepeer/issues/1100

> Presumably OS will be external - like S3, but verifier will be in internal network, so this would be better?

Don't think we can really expect this assumption to always hold. The least surprising thing to do (and most correct one IMO) is this: if external storage is used, then we should use the external storage URL for all segment handling outside of the node.

j0sh:
In order to merge this:

* Move the WaitGroup / condvar changes into their own commit, and add unit tests around `dlFunc` . The friction around writing tests is *the* reason this otherwise sensible change hasn't been done yet. I'd rather not make changes without improving the test coverage since it's currently inadequate in this corner of the codebase. Since this particular change is also unrelated to this PR, it's also fine to revert if writing tests is going to take too long.

* Move the `shouldStopStream` / `shouldStopSession` cleanup into its own commit, again because it's unrelated to the rest of the changes here. (Also OK to revert.)

* Revert the changes to the `transcodeSegment` function signature. We can revisit the overall data flow in more depth when it's not squeezed into a feature that can otherwise live without it.

* Pull from the local OS if necessary within the HTTP push handler. Feel free to factor out some of the URL path construction into a common helper.

* Address other requested changes around the multipart return code.

server/mediaserver.go
darkdarkdragon: >  Minimizing that risk means avoiding changing anything in how we handle state right now.
> But it was done that way for a reason: to keep the verification as self-contained as possible without infiltrating the rest of the broadcaster code, 

And then I'll add another piece of code which will bit `a bit ugly`. 
And then again. And again. And soon all our code will be `a bit ugly`. Or not a bit.

> The upload goes to a different location precisely to avoid this problem. 

My mistake, I thought about something else which is actually not happening.
I wrote under assumption that B downloads data and then uploads it into own external O.
But that shouldn't happen under normal circumstances, so my other points are irrelevant.

server/mediaserver.go

server/broadcast.go
darkdarkdragon: fixed

server/broadcast.go
darkdarkdragon: fixed

server/mediaserver.go

server/mediaserver.go

server/mediaserver.go
darkdarkdragon: added

darkdarkdragon:
@j0sh please take a look. I've addressed all the comments, except one about returning results.
I've changed `verify`'s signature, but I did not change any logic of data flow. I think that way code is simpler and easier to reason about.

j0sh:
> addressed all the comments, except one about returning results.
I've changed verify's signature, but I did not change any logic of data flow. I think that way code is simpler and easier to reason about.

Thanks for making the changes. I do appreciate you trying to make things easier to understand, and agree that the strategy is generally a good one. However, I do have some additional thoughts about how to improve things. Before I suggest those thoughts, note that the original request was to avoid making any changes in this area entirely, precisely to avoid the additional back-and-forth unrelated to this PR. I'm happy to continue the conversation around the changes in https://github.com/livepeer/go-livepeer/pull/1280/commits/da0e03586c793cda53a24eb4bf9b199506253cb1  in a separate PR that doesn't block the multipart feature from going in.

darkdarkdragon:
@j0sh done

j0sh (approved):
LGTM. As mentioned in PM, please don't create a merge commit for this PR due to the large amount of whitespace-only indentation changes. Rather, rebase via the github UI or do a local fast-forward merge/rebase into master (be sure there is no merge commit), then push into master.

PR #1284, opened 2019-12-23
Fix HTTP push handler accepting bad URL

### LINKED ISSUE #1283: Panic in HTTP push handler
[signal SIGSEGV: segmentation violation code=0x1 addr=0x60 pc=0xf4af36]

goroutine 24705 [running]:
github.com/livepeer/go-livepeer/server.(*LivepeerServer).HandlePush.func1(0xc00016de00, 0xc000231b20, 0xc0001f45a0, 0x24)
        /build/server/mediaserver.go:624 +0xd6
created by github.com/livepeer/go-livepeer/server.(*LivepeerServer).HandlePush
        /build/server/mediaserver.go:621 +0x8a9

darkdarkdragon:
In HTTP push handler, check that manifest ID parsed from incoming URL is not empty.

Unit test added.

Fixes #1283 

**Checklist:**

### DIFF
--- a/server/mediaserver.go
+++ b/server/mediaserver.go
@@ -54,11 +54,12 @@ const StreamKeyBytes = 6
 
 const SegLen = 2 * time.Second
 const BroadcastRetry = 15 * time.Second
-const RefreshIntervalHttpPush = 1 * time.Minute
 
 var BroadcastJobVideoProfiles = []ffmpeg.VideoProfile{ffmpeg.P240p30fps4x3, ffmpeg.P360p30fps16x9}
 
 var AuthWebhookURL string
+
+var refreshIntervalHttpPush = 1 * time.Minute
 
 type streamParameters struct {
 	mid        core.ManifestID
@@ -410,9 +411,10 @@ func (s *LivepeerServer) registerConnection(rtmpStrm stream.RTMPVideoStream) (*r
 		Bitrate:    "4000k", // Fix this
 	}
 	hlsStrmID := core.MakeStreamID(mid, &vProfile)
-	s.connectionLock.Lock()
+	s.connectionLock.RLock()
+	// Fast path - check early if session exists - creating new session can take time
 	_, exists := s.rtmpConnections[mid]
-	s.connectionLock.Unlock()
+	s.connectionLock.RUnlock()
 	if exists {
 		// We can only have one concurrent stream per ManifestID
 		return nil, errAlreadyExists
@@ -431,6 +433,14 @@ func (s *LivepeerServer) registerConnection(rtmpStrm stream.RTMPVideoStream) (*r
 	}
 
 	s.connectionLock.Lock()
+	_, exists = s.rtmpConnections[mid]
+	// Check if session exist again - potentially two sessions can be created simultaneously,
+	// so we don't want to overwrite one that was already created
+	if exists {
+		// We can only have one concurrent stream per ManifestID
+		s.connectionLock.Unlock()
+		return nil, errAlreadyExists
+	}
 	s.rtmpConnections[mid] = cxn
 	s.lastManifestID = mid
 	s.lastHLSStreamID = hlsStrmID
@@ -592,6 +602,10 @@ func (s *LivepeerServer) HandlePush(w http.ResponseWriter, r *http.Request) {
 
 	now := time.Now()
 	mid := parseManifestID(r.URL.Path)
+	if mid == "" {
+		http.Error(w, `Bad URL`, http.StatusBadRequest)
+		return
+	}
 	s.connectionLock.Lock()
 	cxn, exists := s.rtmpConnections[mid]
 	if exists && cxn != nil {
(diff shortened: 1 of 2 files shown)


server/mediaserver.go
```diff
@@ -592,12 +591,16 @@ func (s *LivepeerServer) HandlePush(w http.ResponseWriter, r *http.Request) {
 
 	now := time.Now()
 	mid := parseManifestID(r.URL.Path)
-	s.connectionLock.Lock()
+	if mid == "" {
+		http.Error(w, `Bad URL`, http.StatusBadRequest)
+		return
+	}
+	s.connectionLock.RLock()
```
j0sh: The `cxn.lastUsed = now` assignment is a write, not a read - need to keep the `Lock` here. (Yes, it is not ideal to take a global lock here but it's OK, for now to avoid introducing yet another layer of connection-level locking.)

darkdarkdragon: It doesn't mutate `rtmpConnections` map - which we're protecting by this lock. So why not to use `RLock`?

server/mediaserver.go
```diff
@@ -412,8 +412,8 @@ func (s *LivepeerServer) registerConnection(rtmpStrm stream.RTMPVideoStream) (*r
 	hlsStrmID := core.MakeStreamID(mid, &vProfile)
 	s.connectionLock.Lock()
 	_, exists := s.rtmpConnections[mid]
-	s.connectionLock.Unlock()
```
j0sh, line 415: The following few lines of code can be very expensive, especially in other function call-outs where we don't necessarily always know how long they will take (eg, `NewBasicPlaylistManager` and `NewSessionManager`).

In fact, we all spent a few days debugging a deadlock in this exact same spot a few months ago because we were holding the lock for too long during `NewSessionManager`. We have a few options here:

* Make any external calls outside the lock. This is what the old code does.

* Check again that the session exists before assigning within the lock. We could add this.

* Accept that we might get multiple sessions or requests for a given manifest ID. This usually means the client is behaving inappropriately and shouldn't happen under normal conditions. In our case, the worst that happens is that "old" sessions get overwritten. That seems OK here. What we should ensure is that crash-prone bits are made more robust. If understand the root of the crash in https://github.com/livepeer/go-livepeer/pull/1284 correctly, it was something like this:

1. Multiple pushes come in very quickly such that the several sessions are created for the same manifest ID.
2. Multiple cleanup tickers are started
3. Cleanup ticker didn't check for the existence of the connection
4. Hence, if one connection was cleaned up, the second would lead to a crash

The actual solution for here seems to be to fix the cleanup ticker to check for existence of the connection, rather than making changes around locking here.

server/mediaserver.go
j0sh: Multiple requests to the same MID could still change the same object

darkdarkdragon: It's just one `lastUsed` field. Don't see any problem in that.

server/mediaserver.go
darkdarkdragon: > Check again that the session exists before assigning within the lock. We could add this.
> The actual solution for here seems to be to fix the cleanup ticker to check for existence of the connection

Made both 26a4773e393a7e8159e84dac30145ffe5482ffe7

server/mediaserver.go
```diff
@@ -410,9 +410,9 @@ func (s *LivepeerServer) registerConnection(rtmpStrm stream.RTMPVideoStream) (*r
 		Bitrate:    "4000k", // Fix this
 	}
 	hlsStrmID := core.MakeStreamID(mid, &vProfile)
-	s.connectionLock.Lock()
+	s.connectionLock.RLock()
```
j0sh, line 414: Maybe add a comment explaining that we do an early check for session existence here because it may be a while between initializing a session and assigning the session to the connection map. (Otherwise the code reads a bit strangely, since we do the same check almost immediately after the other.)

darkdarkdragon: Added comments in db802aeba273c686aaa8e0f58d03c65983830fe6

darkdarkdragon:
> Also, could we add a unit test for this case?

Added unit tests covering all branches with `lastUsed` in db802aeba273c686aaa8e0f58d03c65983830fe6

server/mediaserver.go
```diff
@@ -616,19 +630,23 @@ func (s *LivepeerServer) HandlePush(w http.ResponseWriter, r *http.Request) {
...
 
-		ticker := time.NewTicker(RefreshIntervalHttpPush)
+		ticker := time.NewTicker(refreshIntervalHttpPush)
 
 		go func(s *LivepeerServer, mid core.ManifestID) {
+			defer ticker.Stop()
 			for range ticker.C {
 				s.connectionLock.RLock()
-				lastUsed := s.rtmpConnections[mid].lastUsed
-				s.connectionLock.RUnlock()
```
j0sh, line 625: Best to keep lock handling to a minimum since it's so easy to mess up. Right now, the unlock needs to be special-cased in three different places, which makes reasoning about the flow difficult. An improvement would be something like:

```golang
lastUsed := time.Now().Sub(2 * refreshIntervalHttpPush) // default to remove if nonexistent
s.connectionLock.RLock()
cxn, exists := s.rtmpConnections[mid]
if exists {
  lastUsed = cxn.lastUsed
}
s.connectionLock.RULock()

if time.Since(lastUsed) {
  removeRTMPStream(s, mid)
}
```

darkdarkdragon: Fixed

server/mediaserver.go
```diff
@@ -635,18 +635,16 @@ func (s *LivepeerServer) HandlePush(w http.ResponseWriter, r *http.Request) {
...
-				if cxn, exists := s.rtmpConnections[mid]; exists {
-					if time.Since(cxn.lastUsed) > refreshIntervalHttpPush {
-						s.connectionLock.RUnlock()
-						_ = removeRTMPStream(s, mid)
-						return
-					}
-				} else {
-					s.connectionLock.RUnlock()
-					return
+				if cx, exists := s.rtmpConnections[mid]; exists {
```
j0sh: We use `cxn` as a matter of convention elsewhere. Maybe preserve that? Otherwise, these changes look good.

j0sh (approved):
🚢

PR #1287, opened 2020-01-09
Remove unneeded locking in LB transcode function

darkdarkdragon:
Locking in `transcoderSession::Transcode` function kills `one segment in flight` rule.

Removed unneeded locking in LB transcode function.
Add unit tests for "Transcoder Busy" and "Transcoder Stopped" errors

Added unit tests.

**Checklist:**

### DIFF
--- a/core/lb.go
+++ b/core/lb.go
@@ -198,8 +198,11 @@ func (sess *transcoderSession) Transcode(job string, fname string, profiles []ff
 			*TranscodeData
 			error
 		})}
+	sess.mu.Lock()
+	sender := sess.sender
+	sess.mu.Unlock()
 	select {
-	case sess.sender <- params:
+	case sender <- params:
 		glog.V(common.DEBUG).Info("LB: Transcode submitted for ", sess.key, fname)
 	default:
 		glog.V(common.DEBUG).Info("LB: Transcoder was busy; exiting ", sess.key, fname)
--- a/core/lb.go
+++ b/core/lb.go
@@ -198,13 +198,11 @@ func (sess *transcoderSession) Transcode(job string, fname string, profiles []ff
 			*TranscodeData
 			error
 		})}
-	sess.mu.Lock()
-	defer sess.mu.Unlock()
 	select {
 	case sess.sender <- params:
-		glog.V(common.DEBUG).Info("LB: Transcode submitted for ", sess.key)
+		glog.V(common.DEBUG).Info("LB: Transcode submitted for ", sess.key, fname)
 	default:
-		glog.V(common.DEBUG).Info("LB: Transcoder was busy; exiting ", sess.key)
+		glog.V(common.DEBUG).Info("LB: Transcoder was busy; exiting ", sess.key, fname)
 		return nil, ErrTranscoderBusy
 	}
 	res := <-params.res
--- a/core/lb.go
+++ b/core/lb.go
@@ -199,15 +199,15 @@ func (sess *transcoderSession) Transcode(job string, fname string, profiles []ff
 			error
 		})}
 	sess.mu.Lock()
-	sender := sess.sender
-	sess.mu.Unlock()
 	select {
-	case sender <- params:
+	case sess.sender <- params:
 		glog.V(common.DEBUG).Info("LB: Transcode submitted for ", sess.key, fname)
 	default:
 		glog.V(common.DEBUG).Info("LB: Transcoder was busy; exiting ", sess.key, fname)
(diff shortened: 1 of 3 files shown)


j0sh:
This is triggering the data race detector; that's why the lock is there. How does the previous behavior kill the one segment in flight rule?

darkdarkdragon:
@j0sh 
first call to `transcoderSession::Transcode` takes lock.
second call - waits till lock frees, that means till previous transcode completion, so we never will receive `transcoder busy` error.

darkdarkdragon:
I'm at one third of reading Rust book 😄 Want to use Rust for programming ARM microcontrollers - Rust seems ideal for that purpose. Not so sure about using it for big networking apps like ours, but let me finish this book first 😄

j0sh:
I'm still getting deadlocks with the following test case (via `go test -run LB_SessionError` ; it still exhibits in non`-race` mode which is a lot faster). It should succeed with fewer iterations of `j`.

My guess it's something like the sequence.

```
seg1 acquires `sender` channel
seg1 writes to `sender` channel
seg2 acquires `sender` channel
seg1 errors out
loop drains `sender` channel
seg2 writes to a `sender` channel that no longer has any listeners
```

TBH I see a few options here

* Reconsider the whole approach. I've had to rewrite this part a few times already due to concurrency issues, and the difficulty around reasoning here is a strong signal that this approach is probably not the correct one.

* Live with it. For now it's highly unlikely we'll ever get concurrent transcodes anyway, because the O itself enforces the one-segment-at-a-time restriction

```golang
func TestLB_SessionErrors(t *testing.T) {
        for j := 0; j < 100; j++ {
                stubCtx, stubCancel := context.WithCancel(context.Background())
                ctxFunc := func() (context.Context, context.CancelFunc) { return stubCtx, stubCancel }

                tcCtx, tcCancel := context.WithCancel(context.Background())
                transcoder := stubTranscoderWithProfiles(nil)
                transcoder.TranscodeFn = func() { <-tcCtx.Done() }
                sess := &transcoderSession{
                        transcoder:  transcoder,
                        mu:          &sync.Mutex{},
                        sender:      make(chan *transcoderParams, 1),
                        makeContext: ctxFunc,
                }

                go sess.loop()

                iters := 100
                errCh := make(chan int)
                for i := 0; i < iters; i++ {
                        if i == iters/2 {
                                stubCancel()
                        }
                        if i == 2*iters/3 {
                                tcCancel()
                        }
                        go func(ch chan int) {
                                _, err := sess.Transcode("", "", []ffmpeg.VideoProfile{})
                                if err == nil {
                                        ch <- 0
                                } else {
                                        ch <- 1
                                        if err != ErrTranscoderBusy && err != ErrTranscoderStopped {
                                                t.Error("Unexpected error from transcoder ", err)
                                        }
                                }
                        }(errCh)
                }

                errCount := 0
                timeout := time.After(100 * time.Second)
                for i := 0; i < iters; i++ {
                        select {
                        case k := <-errCh:
                                errCount += k
                        case <-timeout:
                                t.Error("Stopped because of timeout")
                                break  
                        }
                }
                assert.Equal(t, iters, errCount+transcoder.SegCount)
                assert.Greater(t, errCount, int(float64(iters)*0.9))
        }
}
```

(requires the following changes to stub transcoder)
```diff
diff --git a/core/livepeernode_test.go b/core/livepeernode_test.go
index d301464a..a1ca12c9 100644
--- a/core/livepeernode_test.go
+++ b/core/livepeernode_test.go
@@ -20,6 +20,7 @@ type StubTranscoder struct {
        SegCount      int
        StoppedCount  int
        FailTranscode bool
+       TranscodeFn   func()
 }
 
 func newStubTranscoder(d string, workDir string) TranscoderSession {
@@ -35,6 +36,10 @@ func (t *StubTranscoder) Transcode(job string, fname string, profiles []ffmpeg.V
                return nil, ErrTranscode
        }
 
+       if t.TranscodeFn != nil {
+               t.TranscodeFn()
+       }
+
        t.SegCount++
 
        segments := make([]*TranscodedSegmentData, 0)
```

darkdarkdragon:
@j0sh our problem was that we was locking entire `Transcode` function, including waiting for result part. I've changed locking to lock only send part, but not wait part. That way we still get `transcoder busy` error but deadlock is eliminated.

j0sh:
As of https://github.com/livepeer/go-livepeer/pull/1124/commits/224159423554551dce68afc113f60d210dfbbce2 the deadlocks are resolved differently via filling up the buffer; see the comments in the patch

PR #1289, opened 2020-01-10
server: Handle missing stakes during selection

yondonfu:

This PR allows `MinLSSelector` to more gracefully handle the scenario where it tries to fetch stake weights from the DB for a set of addresses in order to run weighted random selection from the unknown session list, but certain values are missing. Instead, of logging an error and bailing out of the selection (and returning nil), a session with a missing stake weight will be skipped and removed from the unknown session list.

This PR is motivated by the following scenario observed on Rinkeby:

- One of the orchestrators returns ticket params where the `Recipient` field is *not* equal to the orchestrator's ETH address
- The broadcaster uses the value of the ticket params `Recipient` field to [fetch the stake for the orchestrator from the DB](https://github.com/livepeer/go-livepeer/blob/master/server/selection.go#L168)
- Since the value of the ticket params `Recipient` field is different than the orchestrator's ETH address, no corresponding entry will be in the DB
- As a result of the above, the # of orchestrators fetched from the DB will not match the # of addresses provided - this triggers [an error](https://github.com/livepeer/go-livepeer/blob/master/server/selection.go#L79)
- At the moment, an error triggered when reading stake weights will cause the selector to bail out and return nil
- The above will trigger a nil pointer error [here](https://github.com/livepeer/go-livepeer/blob/master/server/broadcast.go#L74)

The described scenario is atypical and should not occur under most circumstances since the value of the ticket params `Recipient` field *should* be the orchestrator's ETH address. Otherwise, any winning tickets redeemed with a different value for the `Recipient` field would result in non-claimable ETH because only active orchestrators are able to claim ETH from winning tickets on-chain.

One guess as to how this situation came to be is that one of the orchestrator operators on Rinkeby is running two separate processes - one process is responsible for calling reward using the ETH address associated with the orchestrator and the other is the actual orchestrator node using a fresh ETH account. This setup would actually be ok *if* the orchestrator node was configured to specify that the `Recipient` field of its ticket params be equal to the orchestrator's ETH address registered on-chain. Perhaps this should be added as an option...But, before jumping to conclusions I'll try to get more information about the orchestrator operator's setup. In the meantime, this PR should allow a broadcaster to handle this situation gracefully.

- Update `storeStakeReader` to *not* return an error when the # of orchs fetched from the DB does not match the # of addresses provided
- Remove a session that we do not have a stake weight for and proceed to the rest of the unknown sessions during weighted random selection

Added unit tests.

I also built from source and verified that a broadcaster connected to Rinkeby can handle the scenario where it is missing the stake weight for an orchestrator.

N/A

**Checklist:**

### DIFF
--- a/server/selection_test.go
+++ b/server/selection_test.go
@@ -40,16 +40,20 @@ func TestStoreStakeReader(t *testing.T) {
 	_, err := rdr.Stakes(nil)
 	assert.EqualError(err, store.err.Error())
 
+	// Test when we receive results for only some addresses
 	store.err = nil
-	store.orchs = []*common.DBOrch{&common.DBOrch{}}
-	_, err = rdr.Stakes(nil)
-	assert.EqualError(err, "could not fetch all stake weights")
+	store.orchs = []*common.DBOrch{&common.DBOrch{EthereumAddr: "foo", Stake: 77}}
+	stakes, err := rdr.Stakes([]ethcommon.Address{ethcommon.Address{}, ethcommon.Address{}})
+	assert.Nil(err)
+	assert.Len(stakes, 1)
+	assert.Equal(stakes[ethcommon.HexToAddress("foo")], int64(77))
 
+	// Test when we receive results for all addresses
 	store.orchs = []*common.DBOrch{
 		&common.DBOrch{EthereumAddr: "foo", Stake: 77},
 		&common.DBOrch{EthereumAddr: "bar", Stake: 88},
 	}
-	stakes, err := rdr.Stakes([]ethcommon.Address{ethcommon.Address{}, ethcommon.Address{}})
+	stakes, err = rdr.Stakes([]ethcommon.Address{ethcommon.Address{}, ethcommon.Address{}})
 	assert.Nil(err)
 
 	for _, orch := range store.orchs {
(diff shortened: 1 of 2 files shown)


mk-livepeer:
Do you think it might make sense to add a unit test for `removeUnknownSession()`, or do you think it might not be necessary?

yondonfu:
@mk-livepeer 

> Do you think it might make sense to add a unit test for removeUnknownSession(), or do you think it might not be necessary?

Added unit tests for `removeUnknownSession()` in 6d20a8a - always good to make sure edge cases are covered :)

server/selection_test.go
```diff
@@ -355,3 +355,48 @@ func TestMinLSSelector_SelectUnknownSession_NilStakeReader(t *testing.T) {
...
+
+	// Test remove from front of list
+	resetUnknownSessions()
+	sel.removeUnknownSession(0)
+	assert.Len(sel.unknownSessions, 2)
+	assert.Equal("baz", string(sel.unknownSessions[0].ManifestID))
+	assert.Equal("bar", string(sel.unknownSessions[1].ManifestID))
+
+	// Test remove from middle of list
+	resetUnknownSessions()
```
mk-livepeer, line 385: rather than reseting in the middle to initialize the next test, I would write four separate test functions, whose name indicates the specific aspect being tested

yondonfu: Lately I've only been using multiple test functions with names that follow the convention you described when testing larger functions. In this case, given that the function under test is fairly small, I decided to keep all the test cases together instead of scattering them across different test functions. I personally find this easier to read, but this might just be matter of personal preference. If you feel strongly about it I can make the change.

mk-livepeer: I think its ok, since the function is so small and there aren't many moving parts

mk-livepeer (approved):

PR #1291, opened 2020-01-11
cmd, core, eth/watchers: O checks that it is active in ProcessPayment

### LINKED ISSUE #1202: Orchestrator should check if it is active before accepting work

yondonfu:

Reopened version of #1226.

For context, #1226 was merged, but I realized that the I missed something in the review and the feature is incomplete. So, the commit was rolled back.

- 
- 
- 

Fixes #1202 

**Checklist:**

### DIFF
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -745,7 +745,7 @@ func main() {
 			return
 		}
 
-		orch := core.NewOrchestrator(s.LivepeerNode)
+		orch := core.NewOrchestrator(s.LivepeerNode, roundsWatcher)
 
 		go func() {
 			server.StartTranscodeServer(orch, *httpAddr, s.HTTPMux, n.WorkDir, n.TranscoderManager != nil)
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -466,14 +466,6 @@ func main() {
 			n.SetBasePrice(big.NewRat(int64(*pricePerUnit), int64(*pixelsPerUnit)))
 			glog.Infof("Price: %d wei for %d pixels\n ", *pricePerUnit, *pixelsPerUnit)
 
-			ctx, cancel := context.WithCancel(context.Background())
-			defer cancel()
-
-			if err := setupOrchestrator(ctx, n, *initializeRound); err != nil {
-				glog.Errorf("Error setting up orchestrator: %v", err)
-				return
-			}
-
 			ev, _ := new(big.Int).SetString(*ticketEV, 10)
 			if ev == nil {
 				glog.Errorf("-ticketEV must be a valid integer, but %v provided. Restart the node with a different valid value for -ticketEV", *ticketEV)
@@ -482,6 +474,14 @@ func main() {
 
 			if ev.Cmp(big.NewInt(0)) < 0 {
 				glog.Errorf("-ticketEV must be greater than 0, but %v provided. Restart the node with a different valid value for -ticketEV", *ticketEV)
+				return
+			}
+
+			ctx, cancel := context.WithCancel(context.Background())
+			defer cancel()
+
+			if err := setupOrchestrator(ctx, n, *initializeRound); err != nil {
+				glog.Errorf("Error setting up orchestrator: %v", err)
 				return
 			}
 
@@ -864,6 +864,21 @@ func getServiceURI(n *core.LivepeerNode, serviceAddr string) (*url.URL, error) {
 func setupOrchestrator(ctx context.Context, n *core.LivepeerNode, initializeRound bool) error {
 	//Check if orchestrator is active
 	active, err := n.Eth.IsActiveTranscoder()
+	if err != nil {
+		return err
+	}
+
+	// add orchestrator to DB
+	orch, err := n.Eth.GetTranscoder(n.Eth.Account().Address)
+	if err != nil {
+		return err
+	}
+
+	err = n.Database.UpdateOrch(&common.DBOrch{
+		EthereumAddr:      n.Eth.Account().Address.Hex(),
+		ActivationRound:   orch.ActivationRound.Int64(),
+		DeactivationRound: orch.DeactivationRound.Int64(),
+	})
 	if err != nil {
 		return err
 	}
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -876,8 +876,8 @@ func setupOrchestrator(ctx context.Context, n *core.LivepeerNode, initializeRoun
 
 	err = n.Database.UpdateOrch(&common.DBOrch{
 		EthereumAddr:      n.Eth.Account().Address.Hex(),
-		ActivationRound:   orch.ActivationRound.Int64(),
-		DeactivationRound: orch.DeactivationRound.Int64(),
+		ActivationRound:   common.ToInt64(orch.ActivationRound),
+		DeactivationRound: common.ToInt64(orch.DeactivationRound),
 	})
 	if err != nil {
 		return err
--- a/cmd/livepeer/livepeer_test.go
+++ b/cmd/livepeer/livepeer_test.go
@@ -1,0 +1,58 @@
+package main
+
+import (
+	"context"
+	"errors"
+	"math/big"
+	"testing"
+
+	ethcommon "github.com/ethereum/go-ethereum/common"
+	"github.com/livepeer/go-livepeer/common"
+	"github.com/livepeer/go-livepeer/core"
+	"github.com/livepeer/go-livepeer/eth"
+	lpTypes "github.com/livepeer/go-livepeer/eth/types"
+	"github.com/livepeer/go-livepeer/pm"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestSetupOrchestrator(t *testing.T) {
+	require := require.New(t)
+	assert := assert.New(t)
+
+	dbh, dbraw, err := common.TempDB(t)
+	require.Nil(err)
+
+	defer dbh.Close()
+	defer dbraw.Close()
+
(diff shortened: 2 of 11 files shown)


core/orchestrator.go
```diff
@@ -321,14 +331,28 @@ func (orch *orchestrator) acceptablePrice(sender ethcommon.Address, ep *net.Pric
 	return nil
 }
 
-func NewOrchestrator(n *LivepeerNode) *orchestrator {
+func (orch *orchestrator) isActive() (bool, error) {
+	filter := &common.DBOrchFilter{
+		CurrentRound: orch.rm.LastInitializedRound(),
+		Addresses:    []ethcommon.Address{orch.Address()},
+	}
+	orchs, err := orch.node.Database.SelectOrchs(filter)
```
yondonfu, line 339: When running in orchestrator mode, the DB's orchestrator table is currently not populated so this query will always return an empty slice.

kyriediculous: added populating DB to the `setupOrchestrator` helper in livepeer.go , which doesn't seem to do much currently except for logging whether the orchestrator is active. 

the fixup just moves around that helper call to avoid doing those RPC calls when running `./test_args.sh`

cmd/livepeer/livepeer.go
```diff
@@ -868,6 +868,21 @@ func setupOrchestrator(ctx context.Context, n *core.LivepeerNode, initializeRoun
...
+	// add orchestrator to DB
+	orch, err := n.Eth.GetTranscoder(n.Eth.Account().Address)
+	if err != nil {
+		return err
+	}
+
+	err = n.Database.UpdateOrch(&common.DBOrch{
+		EthereumAddr:      n.Eth.Account().Address.Hex(),
+		ActivationRound:   orch.ActivationRound.Int64(),
+		DeactivationRound: orch.DeactivationRound.Int64(),
```
yondonfu: If `DeactivationRound` is greater that `math.MaxInt64` (which it will be for an active orchestrator) then it will overflow a int64. We handle this case in the [ethOrchToDBOrch()](https://github.com/livepeer/go-livepeer/blob/master/discovery/db_discovery.go#L302) helper in the `discovery` package. We could extract that helper into a shared package (i.e. `common`), but I don't really like that approach because we end up pulling the `eth/types` package into `common`. An alternative could be to specifically extract the big.Int -> int64 logic into a helper in `common` - something like:

```
func ToInt64(val *big.Int) int64 {
    maxInt64 := int64(math.MaxInt64)
    if val.Cmp(big.NewInt(maxInt64)) > 0 {
        return maxInt64
    }
    
    return val.Int64()
}
```

Usage would look like:

```
&common.DBOrch{
    ...
    DeactivationRound: common.ToInt64(orch.DeactivationRound)
}
```

Side note: I noticed that we don't handle this case in `OrchestratorWatcher` either. We can fix that in separate PR though.

yondonfu (commented):
Since we don't have a test for updating the DB with the orchestrator's own address/activation round/deactivation round let's make sure to do a manual test that confirms this works.

cmd/livepeer/livepeer.go
kyriediculous: While true, if deactivation round starts overflowing it will break any form of doing an active check anyway and we'll have to start storing these values as `BLOB` and converting them to `big.Int` when retrieving them from the DB. 

Added it for sanity in 893bd1d

Also added unit testing for setupOrchestrator 317ba32

cmd/livepeer/livepeer_test.go
```diff
@@ -0,0 +1,58 @@
...
+	"github.com/livepeer/go-livepeer/common"
+	"github.com/livepeer/go-livepeer/core"
+	"github.com/livepeer/go-livepeer/eth"
+	lpTypes "github.com/livepeer/go-livepeer/eth/types"
+	"github.com/livepeer/go-livepeer/pm"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestSetupOrchestrator(t *testing.T) {
```
yondonfu, line 19: Need to add the `cmd/livepeer` package to `test.sh` in order to run this test in CI.

common/util.go
```diff
@@ -246,3 +249,11 @@ var RandomIDGenerator = func(length uint) string {
 func RandName() string {
 	return RandomIDGenerator(10)
 }
+
+func ToInt64(val *big.Int) int64 {
```
yondonfu, line 253: Small unit test for this?

cmd/livepeer/livepeer.go
yondonfu: > While true, if deactivation round starts overflowing it will break any form of doing an active check anyway

`DeactivationRound.Int64()` for an active orchestrator would always overflow when stored in an int64 since the value stored on-chain is a max uint256. In this case, we would end up storing a large negative integer in the `deactivationRound` column of the DB. As a result, every active orchestrator tracked via the DB would be considered inactive in the current round because the node thinks that the orchestrator's deactivation round is in the past. Using the `common.ToInt64()` helper should ensure that we never end up in such a situation.

kyriediculous: I don't think we're talking about the same thing here. 

That alleviates the problem currently, but whenever the rounds on mainnet become larger than maxInt64 ,the current way of doing this will break anyway, regardless if we deal with overflows or not at this stage. Which makes dealing with overflows currently in this scenario kind of trivial. 

It won't occur currently, and when it starts occuring we'll need to change the codebase anyway to use *big.Int across the board and BLOB for storage of these integers. 

Example: 
1. Orchestrator became active in round `math.MaxInt64`
2. O's deactivation round gets set to `math.MaxUint256` in the contracts, but `math.MaxInt64` on the client
3. Orchestrator shows as inactive because `DeactivationRound == ActivationRound`

cmd/livepeer/livepeer_test.go
kyriediculous: fixed in b2909ee

common/util.go
kyriediculous: fixed in 8bf4b1b

cmd/livepeer/livepeer.go
j0sh: Having a round overflowing MaxInt64 is 5e16 years away... what vector other than a protocol bug would lead us to receiving a value in that neighborhood? Are there other consequences more damaging than marking an orchestrator as inactive? (Which seems relatively minor, all things considered.)

I am in favor of avoiding BLOBs and bignum arithmetic wherever possible.

kyriediculous: @j0sh Don't get me wrong, I agree with you here. 

> what vector other than a protocol bug would lead us to receiving a value in that neighborhood?

None, but that exact same would true for the added `ToInt64` helper here, wouldn't it?  We're not likely to encounter the rounds overflowing here for the forseeable future unless there is some protocol bug. Which is why it's more for sanity than absolute necessity. That's all I was trying to say.

test.sh
```diff
@@ -2,58 +2,63 @@
...
 go test -logtostderr=true -v
-t9=$?
+t10=$?
 cd ..
 
 ./test_args.sh
 t_args=$?
 
-if (($t1!=0||$t1_lb!=0||$t2!=0||$t3!=0||$t4!=0||$t5!=0||$t6!=0||$t7!=0||$t8!=0||$t9!=0||$t_args!=0))
+if (($t1!=0||$t1_lb!=0||$t2!=0||$t3!=0||$t4!=0||$t5!=0||$t6!=0||$t7!=0||$t8!=0||$t9!=0||$t10!=0||$t_args!=0))
```
yondonfu: Need to remove `$t1_lb` and add `$t2_lb`

cmd/livepeer/livepeer.go
yondonfu: > Which is why it's more for sanity than absolute necessity. 

I'm assuming that you're just talking about using `common.ToInt64()`  for `ActivationRound` in the above comment?

It is true that using `common.ToInt64()` for `ActivationRound` would just be for consistency because its value is unlikely to ever overflow an `int64` given the current round length on mainnet.

However, assuming that we want to avoid using `big.Int`/`BLOB` here, using `common.Int64()` *is* an absolute necessity for `DeactivationRound` because its value will overflow an `int64` for every active orchestrator since the value on-chain is a max uint256.

kyriediculous: In the `OrchestratorWatcher` we use `const maxFutureRound` whenever a transcoder becomes activated to mimic the on-chain behaviour: https://github.com/livepeer/go-livepeer/blob/f41ad9982d6254f1cf217c6834b15c0a02451521/eth/watchers/orchestratorwatcher.go#L16

But I do concur that the helper  (`ToInt64()`) comes in handy dealing with this issue in `ethOrchToDBOrch` and `setupOrchestrator`. 

Anyway ignore my previous comments , we need to indeed handle overflows in the case you described when making RPC calls, currently we indeed only do it for handling events.

test.sh
kyriediculous: fixed in 15533a2

yondonfu (commented):

kyriediculous:
rebased !

yondonfu (commented):
LGTM

Since I opened this PR I cannot also approve it...so @kyriediculous feel free to approve the PR as one of the reviewers and then merge.

kyriediculous (approved):

PR #1227, opened 2019-12-01
implement windows gpu build

iameli:
**What does this pull request do? Explain your changes. (required)**

The recent change from `nvcc` to `clang` for the nvidia-flavored parts
of our ffmpeg build chain [1] basically gave us this feature for free.
All that was necessary was to add the mingw64 versions of the clang
compiler and it started happily producing binaries that interface with
CUDA on Windows.

I removed the Windows Docker build entirely. It was terrible, frequently
taking close to 90 minutes per build and frequently crashed. This change
implies that `docker/Dockerfile.build` and `docker/Dockerfile.build-linux`
can be combined back into one Dockerfile.build process; I'll open a tech
debt ticket for that but I don't think it's major enough to block this
merge.

Instead, we have a `.\windows-build.ps1` PowerShell script that takes
care of downloading MSYS2 using [Chocolatey](https://chocolatey.org/),
installing the necessary MSYS2/mingw64 packages, and running through
the build. It unpacks everything into a local .gitignored directory, so
it should be able to handle producing workspace-local binaries without
mucking with any system-level packages installed on the host.

This is the conclusion of a 10-month side project and I'm pretty stoked
😃

[1]: https://github.com/livepeer/go-livepeer/pull/1208

**How did you test each of these updates (required)**
I've tested this locally at my desk with a Windows T, Linux O, and macOS
B just to prove that it's possible. I have not done any kind of extensive
benchmarking work, but I did confirm that the `ja/lb` changes functioned
as expected with the appropriate Nvidia drivers. I think more testing is
definitely necessary, but shipping this PR and getting the binaries onto
more Windows computers is almost a prerequisite. Perhaps if we distribute
any Windows binaries we could label them as "experimental" or some such?

(I do think it's all working quite well, though.)

PR #1295, opened 2020-01-12
new release process for 'rinkeby' and 'mainnet' branches

iameli:
See the new doc/releases.md for an explanation of these changes.

I did a couple dummy commits to test different functionality; [here's a successful build with support for the nonexistent `eli/new-release-process` branch](https://app.circleci.com/jobs/github/livepeer/go-livepeer/4042/parallel-runs/0/steps/0-108) and [here's a failing build illustrating the failsafe that disallows non-semver-tagged mainnet releases](https://app.circleci.com/jobs/github/livepeer/go-livepeer/4048/parallel-runs/0/steps/0-108).

https://github.com/livepeer/go-livepeer/issues/1296 has been causing these builds to fail intermittently.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -44,6 +44,17 @@ jobs:
               export HIGHEST_CHAIN_TAG=$branch
             fi
           done
+
+          # Disallow non-tagged mainnet releases
+          generatedVersion=$(./print_version.sh)
+          definedVersion=$(cat VERSION)
+          if [[ $HIGHEST_CHAIN_TAG == "mainnet" ]]; then
+            if [[ $generatedVersion != $definedVesion ]]; then
+              echo "disallowing mainnet release without semver tag; $generatedVersion != $definedVersion"
+              exit 1
+            fi
+          fi
+
           echo "building with support for network: $HIGHEST_CHAIN_TAG"
           docker build --build-arg HIGHEST_CHAIN_TAG -t livepeerci/build:latest --cache-from=livepeerci/build:latest -f docker/Dockerfile.build .
       - run: docker push livepeerci/build:latest
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -19,8 +19,6 @@ jobs:
       PKG_CONFIG_PATH: /root/compiled/lib/pkgconfig
       GOPATH: /go
       DOCKER_CLI_EXPERIMENTAL: enabled
-      # If we want to build with branch --> network support for any other networks, add them here!
-      NETWORK_BRANCHES: rinkeby mainnet
 
     steps:
       - checkout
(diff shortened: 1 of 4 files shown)


yondonfu (commented):
Should we update Travis with the same build logic for the Darwin binaries?

iameli:
Oh, good point. I'll put it in a script instead, so #1227 can use it too.

iameli:
Updated.

doc/releases.md
```diff
@@ -0,0 +1,42 @@
...
+
+We presently use three branches for go-livepeer releases.
+
+#### master
+
+Built in CI with go build `-tags dev`, releasing binaries that may only connect to private networks. Published to Docker Hub as `livepeer/go-livepeer:master`.
+
+#### rinkeby
+
+Cut off of master periodically when we're comfortable that it's reasonably stable. Built in CI with -tags rinkeby, so it can connect to the Rinkeby test network. Published to Docker Hub as `livepeer/go-livepeer:rinkeby`.
```
yondonfu: Markdown code highlighting for -tags rinkeby (i.e. `-tags rinkeby`) is missing.

yondonfu (approved):
LGTM after a rebase!

.circleci/config.yml
```diff
@@ -36,18 +35,26 @@ jobs:
...
           # Our Docker tag name should be our branch name with just alphanums
-          TAG=$(echo $CIRCLE_BRANCH | tr -cd '[:alnum:]_')
-          docker build -t livepeer/go-livepeer:${TAG}-linux -f docker/Dockerfile.release-linux .
-          docker push livepeer/go-livepeer:${TAG}-linux
-          # Manifest step is optional in case the Windows build hasn't finished yet
-          docker manifest create livepeer/go-livepeer:${TAG} livepeer/go-livepeer:${TAG}-linux livepeer/go-livepeer:${TAG}-windows || true
-          docker manifest push livepeer/go-livepeer:${TAG} || true
+          BRANCH_TAG=$(echo $CIRCLE_BRANCH | sed 's/\//-/g' | tr -cd '[:alnum:]_-')
+          VERSION_TAG=$(./print_version.sh)
+          docker build -t current-build -f docker/Dockerfile.release-linux .
```
mkrufky, line 50: why build the image with the temporary tag, "current-build" ?   How do we ensure that "current-build" tag doesn't persist but the branch tag & version tags do?

mkrufky (commented):
It isn't clear to me now where `HIGHEST_CHAIN_TAG` is set.  With it unset, are we building with mainnet support, only?

iameli:
@mkrufky It's being set in ci_env.sh (https://github.com/livepeer/go-livepeer/pull/1295/files#diff-de4b90f3a3b0b6e0dc35a64cb767cb73) which is sourced by Travis and CircleCI prior to the build (https://github.com/livepeer/go-livepeer/pull/1295/files#diff-1d37e48f9ceff6d8030570cd36286a61R39)

If it's unset, there's no build tag, and we default to building with `dev` support only - no mainnet, no rinkeby.

mkrufky:
OK, so then we just made a native local build more difficult for users.

I think we should error and fail the build with an explanation message if the build tag is missing.

.circleci/config.yml
iameli: I added functionality here to push to both e.g. `livepeer/go-livepeer:mainnet` and `livepeer/go-livepeer:0.5.3` during the build. `current-build` is just a placeholder variable; on line 52 I apply the actual tag then push it. `current-build` doesn't persist anywhere - it's not pushed, so it's just part of the ephemeral local Docker context.

iameli:
Which part became more difficult? The behavior of `make livepeer` is completely unchanged.

.circleci/config.yml
mkrufky: thanks for explaining :+1:

mkrufky:
image resulting from `make livepeer` will work on which networks?  dev only?

iameli:
Correct. I'm not opposed to adding a message/documenting the process of `HIGHEST_CHAIN_TAG=rinkeby make livepeer`, but this change doesn't affect that process at all

mkrufky:
You're right - it's fine, it just wasn't clear to me before.   I think adding such a message would be a good idea.  Anybody trying to build from source could easily build the dev only version and have no idea why things don't work.  ... meanwhile, all makes sense now, thanks for explaining :+1:

j0sh:
> the dev only version and have no idea why things don't work

In that case, we print out an error message and quit. For example:

```
$ ./livepeer -broadcaster -network mainnet
I0113 11:37:08.317882   44812 livepeer.go:203] ***Livepeer is running on the mainnet network: 0xf96d54e490317c557a967abfa5d6e33006be69b3***
I0113 11:37:08.319030   44812 livepeer.go:218] Creating data dir: /Users/josh/.lpData/mainnet
E0113 11:37:08.964984   44812 livepeer.go:336] node does not support chainID = 1 right now
```

The error message could be improved, but at least there is a pretty obvious message for the user that something isn't working correctly, and it fails fast. Feel free to open a ticket / PR to improve the language.

PR #1305, opened 2020-01-14
cmd/cli: ETH & LPT unit formatting for pending fees and pending stake

### LINKED ISSUE #1301: Add unit formatting to all LPT and ETH balances in CLI
Formatting of displayed units of value is not consistent in the CLI. 
All displayed units should use `eth.FormatUnits`

"delegator stats"  under orchestrator node info (option 1 in CLI when running an O) still needs formatting.

**Additional context**

```
+---------------+
|DELEGATOR STATS|
+---------------+
*------------------*--------------------------------------------*
|           Status |                                    Pending |
*------------------*--------------------------------------------*
|            Stake |                       24927400000000000000 |
*------------------*--------------------------------------------*
|   Collected Fees |                                          0 |
*------------------*--------------------------------------------*
|    Pending Stake |                       24927400000000000000 |
*------------------*--------------------------------------------*
|     Pending Fees |                                          0 |
*------------------*--------------------------------------------*
|  Delegated Stake |                     1909738352117206298111 |
*------------------*--------------------------------------------*
```

kyriediculous:
This PR attempts to add small improvements to the CLI by adding unit formatting and double checking if streamflow vocabulary is used ubiquitously everywhere. 

- Since `pendingFees` and `pendingStake` no longer reverts I removed the need to set this to `-1` is the call reverts (empty output)
- Added `eth.FormatUnits` to the delegator stats section
- Renamed the `list registered transcoders` option to `list active transcoders`

ran CLI

Fixes #1301 
**Checklist:**

### DIFF
--- a/cmd/livepeer_cli/wizard_stats.go
+++ b/cmd/livepeer_cli/wizard_stats.go
@@ -224,30 +224,32 @@ func (w *wizard) delegatorStats() {
 		return
 	}
 
-	fmt.Println("+---------------+")
-	fmt.Println("|DELEGATOR STATS|")
-	fmt.Println("+---------------+")
-
 	pendingStake := ""
 	if d.PendingStake.Int64() == -1 {
 		pendingStake = "Please fetch pending stake separately"
 	} else {
-		pendingStake = d.PendingStake.String()
+		pendingStake = eth.FormatUnits(d.PendingStake, "LPT")
 	}
+
 	pendingFees := ""
 	if d.PendingFees.Int64() == -1 {
 		pendingFees = "Please fetch pending fees separately"
 	} else {
-		pendingFees = d.PendingFees.String()
+		pendingFees = eth.FormatUnits(d.PendingFees, "ETH")
 	}
+
+	fmt.Println("+---------------+")
+	fmt.Println("|DELEGATOR STATS|")
+	fmt.Println("+---------------+")
+
 	table := tablewriter.NewWriter(os.Stdout)
 	data := [][]string{
 		[]string{"Status", d.Status},
-		[]string{"Stake", d.BondedAmount.String()},
-		[]string{"Collected Fees", d.Fees.String()},
+		[]string{"Stake", eth.FormatUnits(d.BondedAmount, "LPT")},
+		[]string{"Collected Fees", eth.FormatUnits(d.Fees, "ETH")},
 		[]string{"Pending Stake", pendingStake},
 		[]string{"Pending Fees", pendingFees},
-		[]string{"Delegated Stake", d.DelegatedAmount.String()},
+		[]string{"Delegated Stake", eth.FormatUnits(d.DelegatedAmount, "LPT")},
 		[]string{"Delegate Address", d.DelegateAddress.Hex()},
 		[]string{"Last Claim Round", d.LastClaimRound.String()},
 		[]string{"Start Round", d.StartRound.String()},


cmd/livepeer_cli/livepeer_cli.go
```diff
@@ -84,7 +84,7 @@ func (w *wizard) initializeOptions() []wizardOpt {
 	options := []wizardOpt{
 		{desc: "Get node status", invoke: func() { w.stats(w.orchestrator) }},
 		{desc: "View protocol parameters", invoke: w.protocolStats},
-		{desc: "List registered orchestrators", invoke: func() { w.registeredOrchestratorStats() }},
+		{desc: "List active orchestrators", invoke: func() { w.registeredOrchestratorStats() }},
```
yondonfu: Technically this will list the orchestrators that are *pending* activation in the next round which is not necessarily the active set in the current round. I think "List pending orchestrators" would be more accurate.

cmd/livepeer_cli/wizard_stats.go
```diff
@@ -228,26 +228,14 @@ func (w *wizard) delegatorStats() {
 	fmt.Println("|DELEGATOR STATS|")
 	fmt.Println("+---------------+")
 
-	pendingStake := ""
-	if d.PendingStake.Int64() == -1 {
```
yondonfu, line 232: Hm might be ok to keep this behavior for now [1] - while `pendingStake()` and `pendingFees()` no longer revert, its still possible for calls to run out of gas since remote ETH nodes (i.e. Infura) will usually set a gas cap on RPC calls (for example geth has a `-rpc.gascap` flag) for DDOS protection. According to the [docs](https://infura.io/docs/ethereum/json-rpc/eth_call), Infura currently sets the gas cap to 2x the block gas limit so it is unlikely that `pendingStake()` and `pendingFees()` would run up against this cap on mainnet. That being said, I don't feel great about making assumptions about the gas cap imposed by a remote ETH node and its also easier to run up the gas usage for these calls on a network with faster rounds such as Rinkeby. 

[1] I'm not sure if an out of gas error results in the same empty output error (which `GetDelegator()` currently checks for) that a revert error triggers

yondonfu (changes requested):

cmd/livepeer_cli/livepeer_cli.go
kyriediculous: While you're entirely correct, I felt like active was more intuitive to the user. 

Knowing what pending means would require some deeper understanding of how the protocol keeps track of the working set. 

Have no preference here so I'll leave it up to you.

cmd/livepeer_cli/wizard_stats.go
kyriediculous: If the node operates correctly then it should never run into the gascap, should it? 

So if it does we could assume someone has altered the livepeer node software to manifest such behaviour, no? Then you could argue that if you want to misuse our software that you better handle your own errors as well 😅 . 

Sorry for my possible ignorance here, I hadn't heard of infura imposing that on eth_calls

cmd/livepeer_cli/livepeer_cli.go
yondonfu: The main reason I would avoid saying "active" here is to avoid the following confusing scenarios:
- An orchestrator will become active next round, sees itself in the displayed table and notices that its listed as inactive in the "Active" column
- An orchestrator will become inactive next round, does not see itself in the displayed table and believes that it is inactive when it is still active for the remainder of the round

At the same time, I agree that introducing the pending term might be confusing if there is not also an option to view the active orchestrators.

I think we should standardize the language used to describe the orchestrator set in all tools including the CLI and the explorer. How about we keep the term "registered" in this PR and update the language once we have finalized the language we want to use across all tools?

cmd/livepeer_cli/wizard_stats.go
yondonfu: > If the node operates correctly then it should never run into the gascap, should it?

The node might run into the gas cap is if an account has not claimed earnings in many rounds so `pendingStake()`/`pendingFees()` iterates through many rounds. As mentioned in my previous comment, this is unlikely to happen on mainnet, but just wanted to highlight it as a possibility.

Since this error handling is already here, I don't see any harm in just leaving it as is for now.

cmd/livepeer_cli/livepeer_cli.go
kyriediculous: I'm ok with keeping "registered" for now and formalizing a ubiquitous streamflow language across the board.

cmd/livepeer_cli/wizard_stats.go
kyriediculous: Gotcha.

cmd/livepeer_cli/livepeer_cli.go
kyriediculous: dropped the commit

cmd/livepeer_cli/wizard_stats.go
kyriediculous: Readded the previous behaviour but added unit formatting instead of just string conversion

```
	pendingStake := ""
	if d.PendingStake.Int64() == -1 {
		pendingStake = "Please fetch pending stake separately"
	} else {
		pendingStake = eth.FormatUnits(d.PendingStake, "LPT")
	}

	pendingFees := ""
	if d.PendingFees.Int64() == -1 {
		pendingFees = "Please fetch pending fees separately"
	} else {
		pendingFees = eth.FormatUnits(d.PendingFees, "ETH")
	}
```

yondonfu (approved):
LGTM 🚢

PR #1308, opened 2020-01-15
ci: add failsafes for clobbering mainnet build

iameli:
Two things to prevent what happened yesterday from happening in the future:

1. Semver-tagged builds are disallowed unless they're on the `mainnet` branch
2. We shan't overwrite existing files when uploading to build.livepeer.live.

### DIFF
--- a/ci_env.sh
+++ b/ci_env.sh
@@ -30,4 +30,9 @@ if [[ $HIGHEST_CHAIN_TAG == "mainnet" ]]; then
     echo "disallowing mainnet release without semver tag; $generatedVersion != $definedVersion"
     exit 1
   fi
+else
+  if [[ $generatedVersion == $definedVersion ]]; then
+    echo "disallowing semver tag release $generatedVersion on branch '$branch', should be 'mainnet'"
+    exit 1
+  fi
 fi
--- a/upload_build.sh
+++ b/upload_build.sh
@@ -53,12 +53,20 @@ contentType="application/x-compressed-tar"
 dateValue=`date -R`
 stringToSign="PUT\n\n${contentType}\n${dateValue}\n${resource}"
 signature=`echo -en ${stringToSign} | openssl sha1 -hmac ${GCLOUD_SECRET} -binary | base64`
+fullUrl="https://storage.googleapis.com${resource}"
+
+# Failsafe - don't overwrite existing uploads!
+if curl --head --fail $fullUrl 2>/dev/null; then
+  echo "$fullUrl already exists, not overwriting!"
+  exit 1
+fi
+
 curl -X PUT -T "${FILE}" \
   -H "Host: storage.googleapis.com" \
   -H "Date: ${dateValue}" \
   -H "Content-Type: ${contentType}" \
   -H "Authorization: AWS ${GCLOUD_KEY}:${signature}" \
-  https://storage.googleapis.com${resource}
+  $fullUrl
 
 curl --fail -s -H "Content-Type: application/json" -X POST -d "{\"content\": \"Build succeeded ✅\nBranch: $BRANCH\nPlatform: $ARCH-amd64\nLast commit: $(git log -1 --pretty=format:'%s by %an')\nhttps://build.livepeer.live/$VERSION/${FILE}\"}" $DISCORD_URL 2>/dev/null
 echo "done"
--- a/doc/releases.md
+++ b/doc/releases.md
@@ -18,25 +18,24 @@ This is the only branch that receives proper semver updates; all releases should
 
 ### Cutting a mainnet release of go-livepeer
 
-This process will vary somewhat based on the particular states of the branches, whether we're doing a hotfix, etc. But the overall steps are the same. First, merge changes into the mainnet branch and make the release commit:
+This process will vary somewhat based on the particular states of the branches, whether we're doing a hotfix, etc. But the overall steps are the same. First, make the release commit on a branch:
 
 ```bash
-git checkout mainnet
-git merge --ff-only rinkeby
+git checkout -b v0.5.2 
 echo -n '0.5.2' > VERSION
 git commit -am 'release v0.5.2'
-git push --atomic origin mainnet v0.5.2
+git push --atomic origin v0.5.2 v0.5.2
 ```
 
-Then, merge that version bump back to rinkeby and master:
+Merge the release commit into master via PR. Then, merge the version bump into rinkeby and mainnet:
 
 ```bash
 git checkout rinkeby
-git merge --ff-only mainnet
+git merge --ff-only master 
 git push origin rinkeby
-git checkout master
-git merge --ff-only mainnet
-git push origin master
+git checkout mainnet
+git merge --ff-only rinkeby 
+git push origin mainnet 
 ```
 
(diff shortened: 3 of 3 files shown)


upload_build.sh
```diff
@@ -53,12 +53,20 @@ contentType="application/x-compressed-tar"
 dateValue=`date -R`
 stringToSign="PUT\n\n${contentType}\n${dateValue}\n${resource}"
 signature=`echo -en ${stringToSign} | openssl sha1 -hmac ${GCLOUD_SECRET} -binary | base64`
+fullUrl="https://storage.googleapis.com${resource}"
+
+# Failsafe - don't overwrite existing uploads!
+if curl --head --fail $fullUrl 2>/dev/null; then
```
yondonfu, line 59: This means if we run into a case where there is any problem with the binary when building from `mainnet` like on Monday then we *have* to do a version bump in order to get a new binary and bump for any change and no updates to a binary for a release for any reason?

ci_env.sh
```diff
@@ -30,4 +30,9 @@ if [[ $HIGHEST_CHAIN_TAG == "mainnet" ]]; then
     echo "disallowing mainnet release without semver tag; $generatedVersion != $definedVersion"
     exit 1
   fi
+else
+  if [[ $generatedVersion == $definedVersion ]]; then
```
yondonfu, line 34: So the CircleCI tests won't run in the case when `mainnet` is merged into `rinkeby` (or `rinkeby` into `master`) after a version bump since this would evaluate to true and the `build-builder` job in CircleCI would fail preventing the `build` step (with the tests) from running?

upload_build.sh
iameli: Not through this mechanism, no. But it's just backed by a GCP bucket, it's easy to make changes manually for rare corner cases like that.

ci_env.sh
iameli: Here's what I'm thinking we do for the workflow from now on — and I'll change the doc to reflect this:

1. Make the release commit on `mainnet`, tag, push, same as usual.
1. Open a PR from `mainnet` to `master`. Merge said PR, resulting in a merge commit on master.
1. Fast-forward `rinkeby` to `master`. Assuming that's what we want.

That process will ensure that the case you mention will never happen — no other builds will run with that tagged commit.

j0sh: Is there a particular reason we don't make the release commit on dev? Presumably this is just bumping the versions file.

If we can fast-forward `master` merges into `mainnet` then the resulting SHA will end up the same. (But if not, it's not a big deal.) The tag can then be created on mainnet branch - the tag doesn't really need to be ported back into `dev`.

iameli: Good point, I'll write it up that way instead.

iameli: Wait, I remember why — it's because tags don't point to a particular branch like that. Say I run all these commands at once:

```bash
git checkout master
echo -n '0.5.2' > VERSION
git commit -am 'release v0.5.2'
git push

git checkout mainnet
git merge --ff-only master
git tag v0.5.2
git push --atomic origin mainnet v0.5.2
```

And only after that does the CircleCI/Travis/Azure build start. The build for `master` will start and notice that its current commit has a tag and treat that as the version number.

I suppose I could fix this by having `print_version.sh` only output the tag commit without its SHA if you're on `mainnet`, but that would break building tagged versions locally. So I dunno. Happy to consider alternatives though.

j0sh: Could we incorporate the branch into the version tag? Something like `v0.5.2-mainnet` or `v0.5.2-rinkeby` ? Then builds would only occur for that particular branch?

yondonfu: > Could we incorporate the branch into the version tag? Something like v0.5.2-mainnet or v0.5.2-rinkeby ?

Adding on to this, what if we updated `print_version.sh` to something like this:

- If the `currentTag` = `currentVersion` and the branch is not `mainnet`, print `currentVersion-<BRANCH>`
- If the `currentTag` = `currentVersion` and the branch is `mainnet`, print `currentVersion`
- If the `currentTag` != `currentVersion` and the branch is not `mainnet`, print `currentVersion-<BRANCH>-<SHA>` [1]
- If the `currentTag` != `currentVersion` and the branch is `mainnet`, print `currentVersion-<SHA>` [2]

[1] Might be helpful to differentiate non-release CI builds created from `master` vs. `rinkeby` if we merge `master` into `rinkeby` periodically to do testing prior to a release.

[2] Technically, we would never create a build for this case since we have the rule that requires a semver tag for mainnet releases so we don't really need to handle this case, but adding it for completeness.

After the `v0.5.2` commit is created on `master`, if we merge `master` -> `rinkeby` -> `mainnet` we would end up with the following version tags for each branch:

`master` -> `v0.5.2-master`
`rinkeby` -> `v0.5.2-rinkeby`
`mainnet` -> `v0.5.2`

iameli: @j0sh 

> Could we incorporate the branch into the version tag? Something like v0.5.2-mainnet or v0.5.2-rinkeby ? Then builds would only occur for that particular branch?

As actual git tags? It seems to me that `v0.5.2` should uniquely identify the state of the code at a given time; having both `v0.5.2-rinkeby` and `v0.5.2-mainnet` muddles that reality somewhat.

@yondonfu  I'm not sure `print_version.sh` should be branch-aware. I'm concerned about the use case of a user building themselves a binary locally.

```bash
git checkout v0.5.2
HIGHEST_BUILD_TAG=mainnet make livepeer
```

That should deliver you a mainnet-compatible binary that identifies itself as v0.5.2. That conflicts with your first rule above (If the `currentTag` = `currentVersion` and the branch is not `mainnet`, print `currentVersion-<BRANCH>`).

iameli: But maybe that just means we treat `mainnet` and the tag-but-no-branch case identically. That makes your rules work okay, I think.

yondonfu: > I'm concerned about the use case of a user building themselves a binary locally.
> That should deliver you a mainnet-compatible binary that identifies itself as v0.5.2.

Ah ok I see what you're saying now.

If we're merging a PR with the release commit to `master` and then fast-forward merging `master` into `rinkeby` and `master`, perhaps its ok for CI to just not build anything for `master` and `rinkeby` since the binary built for `mainnet` would be functionally identical to the binaries that would be built for `master` and `rinkeby`. Furthermore, the binary built for `mainnet` would be usable for any use case that one might've had for a binary built for `master` and/or `rinkeby` since the `mainnet` binary will support Rinkeby as well as any dev network. Only building a single binary when there is a semver tag release also feels simpler. 

I would be a proponent of:
- Keeping this check to disallow semver tag release builds on any branch besides `mainnet`
- Making the release process:

1. Make release commit and push semver release tag on a `vX.Y.Z` branch
2. Create PR merging `vX.Y.Z` into `master` and merge PR into `master`. This will not create a binary
3. Fast forward merge `master` into `rinkeby`. This will not create a binary
4. Fast forward merge `master` into `mainnet`. This will create a binary that identifies itself as `vX.Y.Z`

yondonfu: PR with suggested release process #1329

yondonfu (approved):
This process seems reasonable to me and I created #1329 against this branch with updates to the docs about the release process of merging into master first and then fast forwarding `rinkeby` and `mainnet`.

PR #1318, opened 2020-01-22
pm, server: close RTMP connection when sender validation fails

### LINKED ISSUE #1247: Close RTMP connection when sender validation fails

#1212 removes any checks pertaining to an on-chain setup and introduces new sender validation logic, these validations for `pm.Sender` are however not ran anymore until an RTMP connection is already established. 

If these sender validation checks end up failing mid-stream we will currently not terminate the RTMP connection but always retry segments instead  and do not propagate/handle errors from `transcodeSegment()`. To properly propagate errors and close RMPT connections when validation checks fail we can introduce a non-retryable error. If a non-retryable error, such as failing sender validation checks), is encountered then the error would be propagated from `transcodeSegment()` up to the caller of `processSegment()` which can then close the RTMP connection to alert the user. Prior to the introduction of a non-retryable error, B would just keep logging errors from sender validation as it continually retries segments (without actually submitting the segments).

**Additional context**
See the original discussion outlining the problem [here](https://github.com/livepeer/go-livepeer/pull/1212#discussion_r352184948)

**Outline**

1. We can introduce a `NonRetryableTranscodeError` type to wrap possible returned errors inside of `server.transcodeSegment()`

```
type NonRetryableTranscodeError struct {
 error
}
```

2. Wrap errors in `SubmitSegment` and `transcodeSegment` that would cause segments to fail to be transcoded , regardless of the orchestrator used, in a `NonRetryableTranscodeError`

3. In `processSegment` we can do a type assertion for the error returned , if this is of type `NonRetryableTranscodeError` we can break out of the retry loop and call `cxn.Stream.Close()` to close the RTMP connection. 

```
func processSegment(cxn *rtmpConnection, seg *stream.HLSSegment) error {
    ... 

	for {
		// if fails, retry; rudimentary
                err := transcodeSegment(cxn, seg, name);
                 if err == nil { return nil }
		if _, ok := err.(NonRetryableTranscodeError); ok {
                   cxn.stream.Close()
                   return err
                }
	}
}
```
```

### PR DESCRIPTION
**Pull request by kyriediculous**

- Return a top-level error from `sender.validateSender` -> `pm.ErrSenderValidation` so that we can compare a returned error against it in the `server` package
- Introduce a `server.nonRetryableTranscodeError` type that implements the `error` interface
- Wrap `sessions == nil` and failing sender validation in `nonRetryableTranscodeError``
- Modify `server.shouldStopStream(err)` to return whether `err` is a `nonRetryableTranscodeError`
- Close RTMP stream when `server.shouldStopStream(err)` return true and break out of the loop in `server.processSegment`
 
Fixes #1247

**Checklist:**

### DIFF
--- a/pm/sender.go
+++ b/pm/sender.go
@@ -9,7 +9,7 @@ import (
 	"github.com/pkg/errors"
 )
 
-// ErrSenderValidation is emitted when sender.validateSender() encounters an error
+// ErrSenderValidation is returned when the sender cannot send tickets
 type ErrSenderValidation struct {
 	error
 }
--- a/pm/sender.go
+++ b/pm/sender.go
@@ -9,6 +9,11 @@ import (
 	"github.com/pkg/errors"
 )
 
+// ErrSenderValidation is emitted when sender.validateSender() encounters an error
+type ErrSenderValidation struct {
+	error
+}
+
 // Sender enables starting multiple probabilistic micropayment sessions with multiple recipients
 // and create tickets that adhere to each session's params and unique nonce requirements.
 type Sender interface {
@@ -78,12 +83,12 @@ func (s *sender) EV(sessionID string) (*big.Rat, error) {
 func (s *sender) validateSender() error {
 	info, err := s.senderManager.GetSenderInfo(s.signer.Account().Address)
 	if err != nil {
-		return fmt.Errorf("unable to validate sender: could not get sender info: %v", err)
+		return ErrSenderValidation{fmt.Errorf("unable to validate sender: could not get sender info: %v", err)}
 	}
 
 	maxWithdrawRound := new(big.Int).Add(s.roundsManager.LastInitializedRound(), big.NewInt(1))
 	if info.WithdrawRound.Int64() != 0 && info.WithdrawRound.Cmp(maxWithdrawRound) != 1 {
-		return fmt.Errorf("unable to validate sender: deposit and reserve is set to unlock soon")
+		return ErrSenderValidation{fmt.Errorf("unable to validate sender: deposit and reserve is set to unlock soon")}
 	}
 
 	return nil
--- a/server/broadcast.go
+++ b/server/broadcast.go
@@ -250,6 +250,7 @@ func selectOrchestrator(n *core.LivepeerNode, params *streamParameters, cpl core
 
 func processSegment(cxn *rtmpConnection, seg *stream.HLSSegment) ([]string, error) {
 
+	rtmpStrm := cxn.stream
 	nonce := cxn.nonce
 	cpl := cxn.pl
 	mid := cxn.mid
@@ -294,6 +295,14 @@ func processSegment(cxn *rtmpConnection, seg *stream.HLSSegment) ([]string, erro
 		if urls, err := transcodeSegment(cxn, seg, name, sv); err == nil {
 			return urls, nil
 		}
+
+		if shouldStopStream(err) {
+			glog.Warningf("Stopping current stream due to: %v", err)
+			rtmpStrm.Close()
+			return nil, err
+		}
+
+		// recoverable error, retry
 	}
 }
 
--- a/pm/sender_test.go
+++ b/pm/sender_test.go
@@ -131,6 +131,8 @@ func TestSender_ValidateSender(t *testing.T) {
 	sm.info[account.Address].WithdrawRound = big.NewInt(2)
 	err = s.validateSender()
 	assert.EqualError(err, "unable to validate sender: deposit and reserve is set to unlock soon")
+	_, ok = err.(ErrSenderValidation)
+	assert.True(ok)
 
 	// Not unlocked
 	sm.info[account.Address].WithdrawRound = big.NewInt(0)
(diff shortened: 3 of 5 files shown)

### CODE REVIEW

File: pm/sender.go [hunk 1]
```diff
@@ -97,7 +100,7 @@ func (s *sender) CreateTicketBatch(sessionID string, size int) (*TicketBatch, er
 	}
 
 	if err := s.validateSender(); err != nil {
-		return nil, err
+		return nil, ErrSenderValidation
```
yondonfu: We should make sure that the underlying error returned by `s.validateSender()` is not dropped here.

Instead of returning a specific error here, what if we returned a custom error type? Something like:

```
type SenderValidationError struct {
    error
}

...

func (s *sender) CreateTicketBatch(...) (...) {
    ...

    if err := s.validateSender(); err != nil {
        return nil, &SenderValidationError{err}
    }
}
```

If we use a custom error type, then the underlying error message can still be logged by the top level caller. The caller can also do a type assertion to check the error type.

server/broadcast.go
```diff
@@ -276,9 +276,20 @@ func processSegment(cxn *rtmpConnection, seg *stream.HLSSegment) error {
...
 		// if fails, retry; rudimentary
-		if err := transcodeSegment(cxn, seg, name); err == nil {
+		err := transcodeSegment(cxn, seg, name)
+		// no error, transcode succeeded
+		if err == nil {
 			return nil
 		}
+
+		// non retryable error, close stream
+		if shouldStopStream(err) {
```
yondonfu, line 299: The updated impl of `shouldStopStream()` would close the RTMP conn when there are no sessions available for a segment which might not be desirable. I think there is a difference between a non-retryable error that closes the RTMP conn vs. a non-retryable error that allows the stream to continue (but drops the current segment). We currently skip the segment when there are no sessions available, but do not close the RTMP conn.

server/broadcast.go
```diff
@@ -330,11 +337,6 @@ func transcodeSegment(cxn *rtmpConnection, seg *stream.HLSSegment, name string)
 			if res == nil && err == nil {
 				return errors.New("Empty response")
 			}
-			if shouldStopStream(err) {
```
yondonfu: Any particular reason why this code block had to move up to `processSegment()`?

server/errors.go
```diff
@@ -0,0 +1,13 @@
+package server
+
+type nonRetryableTranscodeError struct {
```
yondonfu: Most of this file could be omitted if we embed `error` in the struct i.e.

```
type nonRetryableTranscodeError struct {
    error
}
```

yondonfu (changes requested):

server/broadcast.go
kyriediculous: If I'm not mistaken this code block didn't really surve a purpose before as `shouldStopStream()` would just always return false and so the code block would never be executed. Therefore I just decided to reuse it for the purpose of checking non-retryable transcode errors.

kyriediculous:
Ready for full review

server/errors.go
j0sh: Is this file still needed? I don't see `nonRetryableTranscodeError` used anywhere else in this PR

kyriediculous:
Nope that's right. I thought i had removed it in a previous commit, removed
now.

server/broadcast.go
```diff
@@ -276,16 +277,25 @@ func processSegment(cxn *rtmpConnection, seg *stream.HLSSegment) error {
...
 	for {
 		// if fails, retry; rudimentary
-		if err := transcodeSegment(cxn, seg, name); err == nil {
+		err := transcodeSegment(cxn, seg, name)
+		// no error, transcode succeeded
+		if err == nil {
 			return nil
 		}
+
+		// non retryable error, close stream
```
yondonfu: At the moment, we only close the stream for certain non-retryable errors. We don't close the stream for other non-retryable errors (i.e. "No sessions available"). I'd be ok with omitting the comment here entirely since the name of `shouldStopStream()` seems self-explanatory.

pm/sender.go
```diff
@@ -9,6 +9,11 @@ import (
 	"github.com/pkg/errors"
 )
 
+// ErrSenderValidation is emitted when sender.validateSender() encounters an error
```
yondonfu: In case `ErrSenderValidation` is ever returned from a function that is not called `validateSender()` (i.e. if there was a refactor), we could drop `sender.validateSender()` from this comment and just say something like:

```
ErrSenderValidation is returned when the sender cannot send tickets
```

yondonfu (changes requested):
Ah looks like the function signature of `transcodeSegment()` has changed on `master` (changes from [multi-part return PR](https://github.com/livepeer/go-livepeer/commit/68fdf092377740d34b73f7108fe15e5dc8bbb3d4#diff-cecc27bfe2b6b19181ce443733dfba57)) and `shouldStopStream()` was removed. Think its worth rebasing both `nv/b-deposit-withdrawround-check` on top of `master` to make sure that the updates in this branch are compatible.

kyriediculous:
fixed requested changes and rebased on top of master

yondonfu (approved):
LGTM

PR #1322, opened 2020-01-24
Add VAAPI Transcoding support

### LINKED ISSUE #121: VAAPI Support
Supporting VAAPI accelerated transcoding would allow us to use AMD and  Intel QSV hardware. The support for VAAPI can mostly be laid on top of the current nvidia support. Some of the things that should be done:

* Update ffmpeg.go to add VAAPI accel type, and emit the `h264_vaapi` codec and `scale_vaapi` filters as needed

* Add PIX_FMT_VAAPI to init_video_filters

* Deal with the hwupload issue. For SW decode -> GPU rescale+encode, we avoid "wiring up" the GPU within LPMS by taking advantage of the `hwupload_cuda` filter which allows to specify a particular device to upload to:

https://github.com/livepeer/lpms/blob/77a56e05eba3017885e8a395136cf8726b0c65ee/ffmpeg/ffmpeg.go#L90-L94

However, the `device` option does not seem available within the plain `hwupload` filter that VAAPI uses. Our choices here would be to patch the `hwupload` filter to take a similar `device` option, or wire things together within LPMS (in a manner similar to how we wire in the decoder). The former seems much more preferable.

Note that SW decode -> GPU rescale+encode is a fringe case for us right now, but it works for nvidia and it would be nice to have symmetrical VAAPI features (otherwise it is sure to become tech debt).

* Check the logistics of building + running VAAPI-enabled binaries on non-VAAPI hardware. This works fine for nvidia, and I don't expect VAAPI to be much different, but it's something we should check. Also figure out whether it's better to depend on system shared libs (as we do for cuda) or statically link something like Mesa.

This is just from a quick skim of https://github.com/livepeer/lpms/pull/114/files ; probably missing steps and I'm certain things will break along the way.

AbAb1l:
Adds vaapi transcoding support.

- Updates LPMS to add intel transcoding (see livepeer/lpms#170)
- Adds a new VaapiTranscoder type implementing the Transcoder interface
- Adds a -vaapi command line flag taking a list of GPU device IDs to use for transcoding.
- Updates documentation with a new doc/vaapi.md page with some usage details for the system.

Unit testing, manual testing on Linux with Intel Corporation HD Graphics 5500, and on Philipp's rig

Fixes livepeer/lpms#121

**Checklist:**

- [x ] README and other documentation updated

### DIFF
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -101,6 +101,7 @@ func main() {
 	maxSessions := flag.Int("maxSessions", 10, "Maximum number of concurrent transcoding sessions for Orchestrator, maximum number or RTMP streams for Broadcaster, or maximum capacity for transcoder")
 	currentManifest := flag.Bool("currentManifest", false, "Expose the currently active ManifestID as \"/stream/current.m3u8\"")
 	nvidia := flag.String("nvidia", "", "Comma-separated list of Nvidia GPU device IDs to use for transcoding")
+	vaapi := flag.String("vaapi", "", "Comma-separated list of Intel/AMD GPU device IDs to use for transcoding")
 
 	// Onchain:
 	ethAcctAddr := flag.String("ethAcctAddr", "", "Existing Eth account address")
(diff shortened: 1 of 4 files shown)


AbAb1l:
Depends on github.com/livepeer/lpms v0.0.0-20200124150926-8d16e6ffff4d

thomshutt:

PR #1330, opened 2020-01-28
Stake related selection bug fixes

### LINKED ISSUE #1304: panic: runtime error: invalid memory address or nil pointer dereference
**Describe the bug**

When running livepeer `0.5.3` on rinkeby, I am consistently getting a `panic: runtime error` when streaming into the broadcaster.

**To Reproduce**

I am able to reproduce by running the following processes on the same server (Ubuntu):

For `orchestrator` process:
```
./livepeer-linux-amd64/livepeer -orchestrator -cliAddr 127.0.0.1:7936 -httpAddr 127.0.0.1:8936 -serviceAddr 127.0.0.1:8936 -pricePerUnit 20000000 -pixelsPerUnit 1 -ticketEV 1000000000000000 -orchSecret secret -v 99 -network rinkeby
```

For `transcoder` process
```
./livepeer-linux-amd64/livepeer -transcoder -orchAddr 127.0.0.1:8936 -cliAddr 127.0.0.1:7937 -httpAddr 127.0.0.1:8937 -orchSecret secret -maxSessions 1 -v 99 -network rinkeby
```

For `broadcaster` process
```
./livepeer-linux-amd64/livepeer -broadcaster -cliAddr 127.0.0.1:7935 -rtmpAddr 127.0.0.1:1935 -httpAddr 0.0.0.0:8935 -orchAddr 127.0.0.1:8936 -transcodingOptions P144p30fps16x9 -depositMultiplier 1 -maxPricePerUnit 25000000 -pixelsPerUnit 1 -maxTicketEV 10000000000000000 -v 99 -network rinkeby
```

And then streaming into the `broadcaster` in a `publisher` process:
```
ffmpeg -re -f lavfi -i testsrc=size=256x144:rate=30,format=yuv420p -f lavfi -i sine -threads 1 -c:v libx264 -b:v 1000k -preset ultrafast -x264-params keyint=30 -strict -2 -c:a aac -f flv rtmp://127.0.0.1:1935/hello_world
```

Resulting behaviour:

```
ubuntu@ip-172-31-43-225:~$ ./livepeer-linux-amd64/livepeer -broadcaster -cliAddr 127.0.0.1:7935 -rtmpAddr 127.0.0.1:1935 -httpAddr 0.0.0.0:8935 -orchAddr 127.0.0.1:8936 -transcodingOptions P144p30fps16x9 -depositMultiplier 1 -maxPricePerUnit 25000000 -pixelsPerUnit 1 -maxTicketEV 10000000000000000 -v 99 -network rinkeby
I0114 13:02:52.892025    2699 livepeer.go:203] ***Livepeer is running on the rinkeby network: 0xA268AEa9D048F8d3A592dD7f1821297972D4C8Ea***
I0114 13:02:52.893067    2699 db.go:316] Initialized DB node
I0114 13:02:53.416434    2699 accountmanager.go:61] Found existing ETH account
I0114 13:02:53.416465    2699 accountmanager.go:188] Defaulting to first ETH account in keystore 0xDAC817294c0c87ca4fA1895eF4b972EAde99f2fd
I0114 13:02:53.416552    2699 accountmanager.go:70] Using Ethereum account: 0xDAC817294c0c87ca4fA1895eF4b972EAde99f2fd
I0114 13:02:54.398389    2699 accountmanager.go:93] Please enter the passphrase to unlock Ethereum account 0xDAC817294c0c87ca4fA1895eF4b972EAde99f2fd
Passphrase:
I0114 13:02:59.857319    2699 accountmanager.go:104] Unlocked ETH account: 0xDAC817294c0c87ca4fA1895eF4b972EAde99f2fd
I0114 13:02:59.857623    2699 client.go:215] Controller: 0xA268AEa9D048F8d3A592dD7f1821297972D4C8Ea
I0114 13:02:59.964321    2699 client.go:236] LivepeerToken: 0x23b814a57D53b1a7A860194F53401D0D639abED7
I0114 13:03:00.087567    2699 client.go:257] ServiceRegistry: 0xa7AFBB45eb8C27a26379F9C550B3b9be7f09c3eA
I0114 13:03:00.191952    2699 client.go:278] BondingManager: 0xe75a5DccfFe8939F7f16CC7f63EB252bB542FE95
I0114 13:03:00.297939    2699 client.go:299] TicketBroker: 0x40897Ff6e44Ca2e145A691701897844669Bf7cee
I0114 13:03:00.406250    2699 client.go:320] RoundsManager: 0x572d1591bD41f50130FD0212058eAe34F1B17290
I0114 13:03:00.512472    2699 client.go:341] Minter: 0xe5056d1fE48F562dd8AcB1BBFFa147223d448C1B
I0114 13:03:00.613638    2699 client.go:362] LivepeerTokenFaucet: 0x6b752dA302C80795ACc9347AF8f472D3473b19f2
I0114 13:03:01.052741    2699 livepeer.go:770] ***Livepeer Running in Broadcaster Mode***
I0114 13:03:01.052788    2699 livepeer.go:771] Video Ingest Endpoint - rtmp://127.0.0.1:1935
I0114 13:03:01.052899    2699 webserver.go:68] CLI server listening on 127.0.0.1:7935
I0114 13:03:01.053030    2699 mediaserver.go:139] Transcode Job Type: [{P144p30fps16x9 400k 30 256x144 16:9}]
I0114 13:03:01.053182    2699 mediaserver.go:161] HTTP Server listening on http://0.0.0.0:8935
I0114 13:03:01.053236    2699 lpms.go:92] LPMS Server listening on rtmp://127.0.0.1:1935
I0114 13:03:08.398021    2699 listener.go:36] RTMP server got upstream: rtmp://127.0.0.1:1935/hello_world
I0114 13:03:08.823213    2699 broadcast.go:117] Starting session refresh
I0114 13:03:08.823280    2699 rpc.go:223] Connecting RPC to https://127.0.0.1:8936
I0114 13:03:09.036412    2699 discovery.go:105] Done fetching orch info for orchestrators, numResponses fetched: 1
I0114 13:03:09.036690    2699 broadcast.go:161] Ending session refresh
I0114 13:03:09.036716    2699 mediaserver.go:350]

Video Created With ManifestID: hello_world

I0114 13:03:09.036754    2699 lpms.go:149] Segment RTMP Req: rtmp://127.0.0.1:1935/stream/hello_world/fe478f1cbf1b
I0114 13:03:09.166110    2699 player.go:63] LPMS got RTMP request @ rtmp://127.0.0.1:1935/stream/hello_world/fe478f1cbf1b
I0114 13:03:13.545414    2699 broadcast.go:256] Processing segment nonce=2698611182413602751 manifestID=hello_world seqNo=0 dur=1.98
[signal SIGSEGV: segmentation violation code=0x1 addr=0x38 pc=0xf32e2b]

goroutine 92 [running]:
github.com/livepeer/go-livepeer/server.(*BroadcastSessionsManager).selectSession(0xc0007dcd40, 0x0)
        /build/server/broadcast.go:78 +0xcb
github.com/livepeer/go-livepeer/server.transcodeSegment(0xc00018f7a0, 0xc000755540, 0xc000578c30, 0xb, 0x0, 0x3fffae147ae147ae, 0x0)
        /build/server/broadcast.go:304 +0x68
github.com/livepeer/go-livepeer/server.processSegment(0xc00018f7a0, 0xc000755540, 0x0, 0xc000827b60)
        /build/server/broadcast.go:292 +0x488
created by github.com/livepeer/go-livepeer/server.gotRTMPStreamHandler.func1.1.1
        /build/server/mediaserver.go:330 +0x62
ubuntu@ip-172-31-43-225:~$ 
```

**Expected behaviour**

I expect that it should start to transcode video content via the `orchestrator`.

 - OS: Ubuntu 18.04.1
```
ubuntu@ip-172-31-43-225:~$ ./livepeer-linux-amd64/livepeer -version
Livepeer Node Version: 0.5.3
Architecture: amd64
Operating system: linux
```

**Additional context**

When running the same commands offchain, i.e. without the `-network rinkeby`, everything runs fine.

### LINKED ISSUE #1317: Index out of range during default on-chain selection
**Describe the bug**

Index out of range error for a B when running default on-chain selection. 

I suspect the problem is that if an unknown session is removed due to missing stake and the selected unknown session is the last one in the list then the index will be 1 when the length of the list is also 1. Example:

- `len(unknownSessions) = 2`
- Index is 0
- `unknownSessions[0]` is missing stake so it is removed
- `len(unknownSessions) = 1`
- We continue in the weighted random selection loop and index is 1
- We get an index out of range error - the index used to access the remaining unknown session should be 0 not 1

**To Reproduce**

Will add after reliably reproducing.

**Expected behavior**

No index out of range error for a B when running default on-chain selection.

**Screenshots**

```
broadcaster_1    | panic: runtime error: index out of range [1] with length 1
broadcaster_1    |
broadcaster_1    | goroutine 1261 [running]:
broadcaster_1    | github.com/livepeer/go-livepeer/server.(*MinLSSelector).removeUnknownSession(...)
broadcaster_1    |     /build/server/selection.go:212
broadcaster_1    | github.com/livepeer/go-livepeer/server.(*MinLSSelector).selectUnknownSession(0xc012a65000, 0x1cefad8)
broadcaster_1    |     /build/server/selection.go:202 +0x634
broadcaster_1    | github.com/livepeer/go-livepeer/server.(*MinLSSelector).Select(0xc012a65000, 0x1)
broadcaster_1    |     /build/server/selection.go:126 +0xcf
broadcaster_1    | github.com/livepeer/go-livepeer/server.(*BroadcastSessionsManager).selectSession(0xc012a65040, 0x0)
broadcaster_1    |     /build/server/broadcast.go:77 +0xc1
broadcaster_1    | github.com/livepeer/go-livepeer/server.transcodeSegment(0xc012da4000, 0xc012db3240, 0xc012d8df70, 0xb, 0x0, 0x1f1ab40, 0xc01495e1a0)
broadcaster_1    |     /build/server/broadcast.go:304 +0x68
broadcaster_1    | github.com/livepeer/go-livepeer/server.processSegment(0xc012da4000, 0xc012db3240, 0x0, 0xc013185620)
broadcaster_1    |     /build/server/broadcast.go:292 +0x488
broadcaster_1    | created by github.com/livepeer/go-livepeer/server.gotRTMPStreamHandler.func1.1.1
broadcaster_1    |     /build/server/mediaserver.go:330 +0x62
```

 - OS: [e.g. iOS]: Linux
 - Version [e.g. 22]: 0.5.3

yondonfu:

This PR fixes a nil pointer bug and index out of range bug during selection. Both of these bugs are caused by a failure to read stake weights for sessions during selection.

The nil pointer bug occurs when `Select()` returns nil despite a non-zero session list length. This can happen either when the selector fails to fetch a list of stake weights or when all of the unknown sessions are missing stake weights. #1304 results from the latter scenario.

The index out of range bug occurs when a session is removed from the unknown session list in the middle of the for loop (due to missing stake weight) and then the last session in the list is selected. Suppose the length of the list was `N` at the start of selection. The index used to access the last session in the list will be `N - 1`, but the length of the list will be `N - 2` because a session was removed in the middle of the for loop. Note that removing a session from the unknown list in the middle of the for loop is problematic even if the last session in the list is selected because while an index out of range error will not be triggered, the session returned by `Select()` will be off by one (i.e. we would return the `m + 1`th session when we actually should return the `m`th session).

The solutions implemented in this PR are:

- Instead of removing unknown sessions in the middle of the for loop during selection if they are missing stake weights, we default their stake weight to 0. These sessions can still be selected, but they will always be selected after other unknown sessions that are not missing stake weights. This eliminates the indexing issue that caused the index out of range bug because the only time we mutate the unknown session list is when we are returning the selected session (and exiting the for loop). An alternative would be to remove unknown sessions that are missing stake weights after the for loop completes. I opted not to use this approach because there would be a discrepancy between the state of the `sessMap` maintained by `BroadcastSessionsManager`, which is supposed to track unique sessions, and the internal state of the selector - the session would be removed from the selector, but there would still be an entry in `sessMap`
- Add a nil check in `selectSession()` to make sure we never try to access a field of a session if the session is nil. Usually, this should never happen because we check that the list length is non-zero before calling `Select()`. It can happen if an error is encountered during selection that prevents a session from being selected despite a non-zero list length - currently, this would happen if the selector fails to fetch a list of stake weights (i.e. a DB error) during selection. An alternative would be to make sure that the selector will only ever return nil if the list is empty. I opted not to use this approach because a) when a DB error occurs such that the selector cannot fetch a list of stake weights this feels like a scenario where something is really wrong internally with the node and we should not proceed [1] b) having the caller do a nil check is a safer/more defensive approach that doesn't make assumptions about the implementation of the selector interface.
- Always cache on-chain orchs and poll for updates when the node is in on-chain mode regardless of the discovery mechanism (i.e. hardcoded list, webhook, on-chain active set). This allows a broadcaster to specify a subset of the on-chain active set via a hardcoded list or webhook for stake/latency based selection.  

[1] I think the case where the selector cannot fetch a list of stake weights due to a DB error is different from the case where the stake weights are missing for certain sessions (but the selector still fetched a list with *some* stake weights from the DB). The former is likely the result of some internal node issue. The latter is likely the result of an issue external to the node i.e. a misconfigured orch that is advertising a recipient address that is different from its on-chain address.

- 9fc68e3: Sets the default stake weight of a session to 0 if the selector successfully fetched a list of stake weights for a list of sessions, but some of the stake weights are missing
- 1e62115: Adds a nil check for the session returned by `Select()`
- 9b30acd: Use DBOrchestratorPoolCache constructor to always cache orchs and poll for updates when in on-chain mode

Added unit tests.

Fixes #1304 
Fixes #1317 

**Checklist:**

### DIFF
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -639,6 +639,20 @@ func main() {
 
 		bcast := core.NewBroadcaster(n)
 
+		// When the node is on-chain mode always cache the on-chain orchestrators and poll for updates
+		// Right now we rely on the DBOrchestratorPoolCache constructor to do this. Consider separating the logic
+		// caching/polling from the logic for fetching orchestrators during discovery
+		if *network != "offchain" {
+			ctx, cancel := context.WithCancel(context.Background())
+			defer cancel()
+			dbOrchPoolCache, err := discovery.NewDBOrchestratorPoolCache(ctx, n, roundsWatcher)
+			if err != nil {
+				glog.Errorf("Could not create orchestrator pool with DB cache: %v", err)
+			}
+
+			n.OrchestratorPool = dbOrchPoolCache
+		}
+
 		// Set up orchestrator discovery
 		if *orchWebhookURL != "" {
 			whurl, err := validateURL(*orchWebhookURL)
@@ -649,17 +663,8 @@ func main() {
 			n.OrchestratorPool = discovery.NewWebhookPool(bcast, whurl)
 		} else if len(orchURLs) > 0 {
 			n.OrchestratorPool = discovery.NewOrchestratorPool(bcast, orchURLs)
-		} else if *network != "offchain" {
-			ctx, cancel := context.WithCancel(context.Background())
-			defer cancel()
-			dbOrchPoolCache, err := discovery.NewDBOrchestratorPoolCache(ctx, n, roundsWatcher)
-
-			if err != nil {
-				glog.Errorf("Could not create orchestrator pool with DB cache: %v", err)
-			}
-
-			n.OrchestratorPool = dbOrchPoolCache
 		}
+
 		if n.OrchestratorPool == nil {
 			// Not a fatal error; may continue operating in segment-only mode
 			glog.Error("No orchestrator specified; transcoding will not happen")
--- a/server/broadcast.go
+++ b/server/broadcast.go
@@ -76,6 +76,11 @@ func (bsm *BroadcastSessionsManager) selectSession() *BroadcastSession {
 	}
 	for checkSessions(bsm) {
 		sess := bsm.sel.Select()
+		// Handle the case where there is an error during selection and Select() returns nil when session list length > 0
+		if sess == nil {
+			return nil
+		}
+
 		if _, ok := bsm.sessMap[sess.OrchestratorInfo.Transcoder]; ok {
 			return sess
 		}
--- a/server/selection_test.go
+++ b/server/selection_test.go
@@ -76,7 +76,13 @@ func (r *stubStakeReader) Stakes(addrs []ethcommon.Address) (map[ethcommon.Addre
 	if r.err != nil {
 		return nil, r.err
 	}
-	return r.stakes, nil
+
+	stakes := make(map[ethcommon.Address]int64)
+	for _, addr := range addrs {
+		stakes[addr] = r.stakes[addr]
+	}
+
+	return stakes, nil
 }
 
 func (r *stubStakeReader) SetStakes(stakes map[ethcommon.Address]int64) {
@@ -315,25 +321,48 @@ func TestMinLSSelector_SelectUnknownSession_UniformWeights(t *testing.T) {
 	}
 }
 
-func TestMinLSSelector_SelectUnknownSession_MissingStake(t *testing.T) {
+func TestMinLSSelector_SelectUnknownSession_AllMissingStake(t *testing.T) {
 	assert := assert.New(t)
 
 	stakeRdr := newStubStakeReader()
 	sel := NewMinLSSelector(stakeRdr, 1.0)
 
-	// Initialize stake reader with empty stake map
+	// Initialize stake reader with empty stake map so all sessions are missing stake
 	stakeRdr.SetStakes(make(map[ethcommon.Address]int64))
 
-	sess := &BroadcastSession{
-		OrchestratorInfo: &net.OrchestratorInfo{
-			TicketParams: &net.TicketParams{Recipient: []byte("foo")},
-		},
-	}
-	sel.Add([]*BroadcastSession{sess})
+	sess1 := StubBroadcastSession("")
+	sess1.OrchestratorInfo.TicketParams = &net.TicketParams{Recipient: []byte("foo")}
+	sess2 := StubBroadcastSession("")
+	sess2.OrchestratorInfo.TicketParams = &net.TicketParams{Recipient: []byte("bar")}
 
-	assert.Nil(sel.Select())
-	// Check that the session was removed because its stake was missing
-	assert.Empty(sel.unknownSessions)
+	sel.Add([]*BroadcastSession{sess1, sess2})
+
+	// The stake weight of both sessions defaults to 0 so they should be selected in the order that they were added
+	assert.Same(sess1, sel.Select())
+	assert.Same(sess2, sel.Select())
+}
+
+func TestMinLSSelector_SelectUnknownSession_SomeMissingStake(t *testing.T) {
+	assert := assert.New(t)
+
+	stakeRdr := newStubStakeReader()
+	sel := NewMinLSSelector(stakeRdr, 1.0)
+
+	sess1 := StubBroadcastSession("")
+	sess1.OrchestratorInfo.TicketParams = &net.TicketParams{Recipient: []byte("foo")}
+	addr2 := ethcommon.BytesToAddress([]byte("bar"))
+	sess2 := StubBroadcastSession("")
+	sess2.OrchestratorInfo.TicketParams = &net.TicketParams{Recipient: addr2.Bytes()}
+
+	sel.Add([]*BroadcastSession{sess1, sess2})
+
+	// Initialize stake reader so that some sessions are missing stake
+	stakeMap := map[ethcommon.Address]int64{addr2: 100}
+	stakeRdr.SetStakes(stakeMap)
+
+	// The stake weight of sess1 defaults to 0 so sess2 should always be selected first
+	assert.Same(sess2, sel.Select())
+	assert.Same(sess1, sel.Select())
 }
 
 func TestMinLSSelector_SelectUnknownSession_NilStakeReader(t *testing.T) {
--- a/server/broadcast_test.go
+++ b/server/broadcast_test.go
@@ -126,6 +126,17 @@ func (pm *stubPlaylistManager) GetOSSession() drivers.OSSession {
 
 func (pm *stubPlaylistManager) Cleanup() {}
 
+type stubSelector struct {
+	sess *BroadcastSession
+	size int
+}
+
+func (s *stubSelector) Add(sessions []*BroadcastSession) {}
+func (s *stubSelector) Complete(sess *BroadcastSession)  {}
+func (s *stubSelector) Select() *BroadcastSession        { return s.sess }
+func (s *stubSelector) Size() int                        { return s.size }
+func (s *stubSelector) Clear()                           {}
+
 func TestStopSessionErrors(t *testing.T) {
 
 	// check error cases
@@ -258,6 +269,14 @@ func TestSelectSession(t *testing.T) {
 	assert.Len(bsm.sessMap, 1)
 
 	// XXX check refresh condition more precisely - currently numOrchs / 2
+}
+
+func TestSelectSession_NilSession(t *testing.T) {
+	bsm := StubBroadcastSessionsManager()
+	// Replace selector with stubSelector that will return nil for Select(), but 1 for Size()
+	bsm.sel = &stubSelector{size: 1}
+
+	assert.Nil(t, bsm.selectSession())
 }
 
 func TestRemoveSession(t *testing.T) {
--- a/server/selection.go
+++ b/server/selection.go
@@ -190,12 +190,7 @@ func (s *MinLSSelector) selectUnknownSession() *BroadcastSession {
 	// The greater the stake weight of a session, the more likely that it will be selected because subtracting its stake weight from r
 	// will result in a value <= 0
 	for i, sess := range s.unknownSessions {
-		// If we could not fetch the stake weight for addrs[i] then skip and remove its session
-		if _, ok := stakes[addrs[i]]; !ok {
-			s.removeUnknownSession(i)
-			continue
-		}
-
+		// If we could not fetch the stake weight for addrs[i] then its stake weight defaults to 0
 		r -= stakes[addrs[i]]
 
 		if r <= 0 {


cmd/livepeer/livepeer.go
```diff
@@ -639,6 +639,20 @@ func main() {
 
 		bcast := core.NewBroadcaster(n)
 
+		var dbOrchPoolCache *discovery.DBOrchestratorPoolCache
+		// When the node is on-chain mode always cache the on-chain orchestrators and poll for updates
+		// Right now we rely on the DBOrchestratorPoolCache constructor to do this. Consider separating the logic
+		// caching/polling from the logic for fetching orchestrators during discovery
+		if *network != "offchain" {
+			var err error
+			ctx, cancel := context.WithCancel(context.Background())
```
mkrufky, line 646: There are seven calls to `context.Background()` in this function.  This is wrong.  Please assign
```ctx := context.Background()```
...once at the top of the function, and you can
```
ctx, cancel := context.WithCancel(ctx)
```
...to fork the context as necessary, all seven times.

This covers the `main()` function, but all the others will also need to be modified to take a forked context out of main, rather than creating a new `context.Background()`

```
cmd/livepeer/livepeer.go:		chainID, err := backend.ChainID(context.Background())
cmd/livepeer/livepeer.go:			ctx, cancel := context.WithCancel(context.Background())
cmd/livepeer/livepeer.go:			gasPriceUpdate, err := gpm.Start(context.Background())
cmd/livepeer/livepeer.go:			rs.Start(context.Background())
cmd/livepeer/livepeer.go:		blockWatchCtx, cancel := context.WithCancel(context.Background())
cmd/livepeer/livepeer.go:			ctx, cancel := context.WithCancel(context.Background())
cmd/livepeer/livepeer.go:	msCtx, cancel := context.WithCancel(context.Background())
common/testutil.go:	return peer.NewContext(context.Background(), p)
core/orchestrator.go:	return context.WithTimeout(context.Background(), transcodeLoopTimeout)
core/orchestrator.go:			ctx, cancel := context.WithTimeout(context.Background(), transcodeLoopTimeout)
core/orchestrator.go:	ctx, cancel := context.WithTimeout(context.Background(), RemoteTranscoderTimeout)
discovery/db_discovery.go:	ctx, cancel := context.WithTimeout(context.Background(), getOrchestratorsTimeoutLoop)
discovery/db_discovery.go:	ctx, cancel := context.WithTimeout(context.Background(), getOrchestratorsTimeoutLoop)
discovery/discovery.go:	ctx, cancel := context.WithTimeout(context.Background(), getOrchestratorsTimeoutLoop)
eth/blockwatch/client.go:	ctx, cancel := context.WithTimeout(context.Background(), rc.requestTimeout)
eth/blockwatch/client.go:	ctx, cancel := context.WithTimeout(context.Background(), rc.requestTimeout)
eth/blockwatch/client.go:	ctx, cancel := context.WithTimeout(context.Background(), rc.requestTimeout)
eth/client.go:	chainID, err := eth.ChainID(context.Background())
eth/client.go:	ctx, cancel := context.WithTimeout(context.Background(), c.txTimeout)
eth/client.go:	_, pending, err := c.backend.TransactionByHash(context.Background(), tx.Hash())
eth/client.go:		suggestedGasPrice, err := c.backend.SuggestGasPrice(context.Background())
eth/client.go:	err = c.backend.SendTransaction(context.Background(), newSignedTx)
eth/eventservices/rewardservice.go:	cancelCtx, cancel := context.WithCancel(context.Background())
eth/noncemanager.go:	remoteNonce, err := m.remoteReader.PendingNonceAt(context.Background(), addr)
monitor/census.go:	census.ctx, err = tag.New(context.Background(), tag.Insert(census.kNodeType, nodeType), tag.Insert(census.kNodeID, nodeID))
monitor/census.go:	ctx, err := tag.New(context.Background(), tag.Insert(census.kNodeType, nodeType), tag.Insert(census.kNodeID, nodeID),
server/mediaserver.go:	lpmsCtx, cancel := context.WithCancel(context.Background())
server/mediaserver.go:			err := s.RTMPSegmenter.SegmentRTMPToHLS(context.Background(), rtmpStrm, hlsStrm, segOptions)
server/ot_rpc.go:	ctx := context.Background()
server/rpc.go:	ctx, cancel := context.WithTimeout(context.Background(), GRPCTimeout)
server/webserver.go:			balance, err := b.BalanceAt(context.Background(), s.LivepeerNode.Eth.Account().Address, nil)
server/webserver.go:			blk, err := backend.BlockByNumber(context.Background(), nil)
server/webserver.go:		chainID, err := be.ChainID(context.Background())
```
...We will have to handle these others in a separate PR - no need to muck up this one.

File: server/broadcast.go [hunk 2]
```diff
@@ -76,6 +76,11 @@ func (bsm *BroadcastSessionsManager) selectSession() *BroadcastSession {
 	}
 	for checkSessions(bsm) {
 		sess := bsm.sel.Select()
+		// Handle the case where there is an error during selection and Select() returns nil when session list length > 0
+		if sess == nil {
+			return nil
+		}
+
```
mkrufky, line 83: :+1:

mkrufky (commented):
The PR itself looks fine and straightforward.   Usage of `context.Background()` is incorrect throughout `go-livepeer`, however.   I will create a separate ticket to track that, but we can start by fixing the `main()` function within this PR.

mk-livepeer (approved):
as @mkrufky mentioned, the PR itself looks good.  The issue that I see with the context handling is a problem across the entire repository.  I created this ticket to track that: https://github.com/livepeer/go-livepeer/issues/1332

cmd/livepeer/livepeer.go
```diff
@@ -639,6 +639,20 @@ func main() {
 
 		bcast := core.NewBroadcaster(n)
 
+		var dbOrchPoolCache *discovery.DBOrchestratorPoolCache
+		// When the node is on-chain mode always cache the on-chain orchestrators and poll for updates
+		// Right now we rely on the DBOrchestratorPoolCache constructor to do this. Consider separating the logic
```
kyriediculous, line 643: Keep in mind that even if we seperate this logic (which I'm in favor of) we would still be processing events for all O's even those we don't care about at startup, this is also true when starting the node as an orchestrator currently. Perhaps seperating these concerns depending on the discovery method chosen can be extracted into a seperate issue though.

cmd/livepeer/livepeer.go
```diff
@@ -650,14 +664,6 @@ func main() {
 		} else if len(orchURLs) > 0 {
 			n.OrchestratorPool = discovery.NewOrchestratorPool(bcast, orchURLs)
 		} else if *network != "offchain" {
```
kyriediculous: I think we can move this codeblock up in its entirety ? and overwrite `n.ORchestratorPool` with `n.OrchestratorPool = discovery.NewOrchestratorPool(bcast, orchURLs)` when we have `orchURLs > 0` which will always we be the case when we run offchain. 

In the case we use a webhook , this will also be overwritten if the option is defined. 

Might have looked over something here though

server/selection_test.go
```diff
@@ -315,25 +321,60 @@ func TestMinLSSelector_SelectUnknownSession_UniformWeights(t *testing.T) {
...
 	stakeRdr := newStubStakeReader()
 	sel := NewMinLSSelector(stakeRdr, 1.0)
 
-	// Initialize stake reader with empty stake map
+	// Initialize stake reader with empty stake map so all sessions are missing stake
 	stakeRdr.SetStakes(make(map[ethcommon.Address]int64))
 
-	sess := &BroadcastSession{
+	addr2 := ethcommon.BytesToAddress([]byte("bar"))
+	sess1 := &BroadcastSession{
```
kyriediculous: Can we use the same notation here for creating `BroadcastSession`

If it's the same logic we could potentially extract this into a helper

```
func createTestBcastSess(recipient ethcommon.Address) *BroadcastSession {
 return &BroadcastSession {
  OrchestratorInfo: &net.OrchestratorInfo {
    TicketParams: &net.TicketParams{ Recipient: recipient.Bytes()}
    }
  }
}
```

This would make the code perhaps a little more readable as we're currently starting with an `addr2` declaration but we don't have `addr` declared.

server/selection_test.go
```diff
@@ -315,25 +321,60 @@ func TestMinLSSelector_SelectUnknownSession_UniformWeights(t *testing.T) {
...
+	assert.Same(sess2, sel.Select())
+}
+
+func TestMinLSSelector_SelectUnknownSession_SomeMissingStake(t *testing.T) {
+	assert := assert.New(t)
+
+	stakeRdr := newStubStakeReader()
+	sel := NewMinLSSelector(stakeRdr, 1.0)
+
+	addr2 := ethcommon.BytesToAddress([]byte("bar"))
```
kyriediculous, line 353: See comment above

kyriediculous (changes requested):
Looks good overall, I think the comments added on L642-L644 of livepeer.go are spot on. 

We should track seperating the logic here in a seperate issue.

server/selection_test.go
yondonfu: Fixed in f24fa47

server/selection_test.go
yondonfu: Fixed in f24fa47

cmd/livepeer/livepeer.go
yondonfu: Fixed in 8633ae4

kyriediculous:
Changes look good, let's rebase !

yondonfu:
Rebased

cmd/livepeer/livepeer.go
yondonfu: > Perhaps seperating these concerns depending on the discovery method chosen can be extracted into a seperate issue though.

Something else to consider is validation of the orchs passed in via `-orchAddr` or `-orchWebhookURL` when the node is on-chain mode to ensure that they are active on-chain.

Let's discuss in a separate issue.

kyriediculous:
LGTM pending green CI ! 👍

kyriediculous (approved):

PR #1331, opened 2020-01-28
server: log info about generated payments when using high verbosity

### LINKED ISSUE #1307: Add logline on high verbosity indicating that broadcasters sent PM tickets with given parameters
As a broadcaster, I don't have any visibility into what PM tickets are being sent with what parameters.

An easy first step is to add a log line that I can view at high verbosity. It should indicate the face value and win probability of the ticket as it is sent.

The prometheus integration could capture and aggregate a lot more data for better summaries and full visibility, but I believe this log solution to be a nice minimum viable first step that will help people understand what's going on.

**Additional context**

kyriediculous:
This PR adds a log message for logging information regarding generated ticket batches that gets logged when starting the broadcaster node with `-v 6`. 

While declaring some of the params of the log statement as variables is unnecessary as most return a single value, I do find it improves code readability. 

- Added log statement in `SubmitSegment` that logs the payment generated by `genPayment()`
- Exported `pm.maxWinProb` so that we can calculate the winprob in the log statement as a floatstring

eg. 
```
"Created new payment -  recipient=0xe426ad6DDF3905de9D798f49cb19d6E9A6a3335f  faceValue=0.05 ETH  winProb=0.0000005 %  price=50 wei/pixel  numTickets=500"
```

Fixes #1307 

**Checklist:**

### DIFF
--- a/server/segment_rpc.go
+++ b/server/segment_rpc.go
@@ -15,6 +15,7 @@ import (
 	"github.com/livepeer/go-livepeer/common"
 	"github.com/livepeer/go-livepeer/core"
 	"github.com/livepeer/go-livepeer/drivers"
+	"github.com/livepeer/go-livepeer/eth"
 	"github.com/livepeer/go-livepeer/monitor"
 	"github.com/livepeer/go-livepeer/net"
 	"github.com/livepeer/lpms/ffmpeg"
@@ -634,6 +635,16 @@ func genPayment(sess *BroadcastSession, numTickets int) (string, error) {
 		}
 
 		protoPayment.TicketSenderParams = senderParams
+
+		ratPrice, _ := ratPriceInfo(protoPayment.ExpectedPrice)
+		glog.V(common.VERBOSE).Infof("Created new payment - manifestID=%v recipient=%v faceValue=%v winProb=%v price=%v numTickets=%v",
+			sess.ManifestID,
+			batch.Recipient.Hex(),
+			eth.FormatUnits(batch.FaceValue, "ETH"),
+			batch.WinProbRat().FloatString(10),
+			ratPrice.FloatString(3)+" wei/pixel",
+			numTickets,
+		)
 	}
 
 	data, err := proto.Marshal(protoPayment)
--- a/core/orchestrator.go
+++ b/core/orchestrator.go
@@ -20,6 +20,7 @@ import (
 
 	"github.com/livepeer/go-livepeer/common"
 	"github.com/livepeer/go-livepeer/drivers"
+	"github.com/livepeer/go-livepeer/eth"
 	"github.com/livepeer/go-livepeer/monitor"
 	"github.com/livepeer/go-livepeer/net"
 	"github.com/livepeer/go-livepeer/pm"
@@ -175,7 +176,7 @@ func (orch *orchestrator) ProcessPayment(payment net.Payment, manifestID Manifes
 			tsp.SenderNonce,
 		)
 
-		glog.V(common.DEBUG).Infof("Receiving ticket manifestID=%v faceValue=%v winProb=%v ev=%v", manifestID, ticket.FaceValue, ticket.WinProbRat().FloatString(10), ticket.EV().FloatString(2))
+		glog.V(common.DEBUG).Infof("Receiving ticket manifestID=%v faceValue=%v winProb=%v ev=%v", manifestID, eth.FormatUnits(ticket.FaceValue, "ETH"), ticket.WinProbRat().FloatString(10), ticket.EV().FloatString(2))
 
 		_, won, err := orch.node.Recipient.ReceiveTicket(
 			ticket,
--- a/pm/ticket.go
+++ b/pm/ticket.go
@@ -43,6 +43,11 @@ type TicketParams struct {
 	Seed *big.Int
 }
 
+// WinProbRat returns the ticket WinProb as a percentage represented as a big.Rat
+func (p *TicketParams) WinProbRat() *big.Rat {
+	return winProbRat(p.WinProb)
+}
+
 // TicketExpirationParams indicates when/how a ticket expires
 type TicketExpirationParams struct {
 	CreationRound int64
@@ -142,7 +147,7 @@ func (t *Ticket) EV() *big.Rat {
 
 // WinProbRat returns the ticket WinProb as a percentage represented as a big.Rat
 func (t *Ticket) WinProbRat() *big.Rat {
-	return new(big.Rat).SetFrac(t.WinProb, maxWinProb)
+	return winProbRat(t.WinProb)
 }
 
 // Hash returns the keccak-256 hash of the ticket's fields as tightly packed
@@ -188,4 +193,8 @@ func (t *Ticket) flatten() []byte {
 
 func ticketEV(faceValue *big.Int, winProb *big.Int) *big.Rat {
 	return new(big.Rat).Mul(new(big.Rat).SetInt(faceValue), new(big.Rat).SetFrac(winProb, maxWinProb))
+}
+
+func winProbRat(winProb *big.Int) *big.Rat {
+	return new(big.Rat).SetFrac(winProb, maxWinProb)
 }


server/segment_rpc.go
```diff
@@ -328,6 +330,18 @@ func SubmitSegment(sess *BroadcastSession, seg *stream.HLSSegment, nonce uint64)
 		return nil, err
 	}
 
+	ratPrice, _ := ratPriceInfo(sess.OrchestratorInfo.PriceInfo)
+	faceVal := new(big.Int).SetBytes(sess.OrchestratorInfo.GetTicketParams().GetFaceValue())
+	winProb := new(big.Int).SetBytes(sess.OrchestratorInfo.GetTicketParams().GetWinProb())
+	winProbRat := new(big.Rat).Mul(new(big.Rat).SetFrac(winProb, pm.MaxWinProb), big.NewRat(100, 1)).FloatString(10) + " %"
+	glog.V(common.VERBOSE).Infof("Created new payment - recipient=%v  faceValue=%v  winProb=%v  price=%v  numTickets=%v",
```
yondonfu: What about placing this log statement in `genPayment()` within the code block that executes if `numTickets > 0`? This would also resolve the current the lack of an on-chain mode check here since `genPayment()` already contains an on-chain mode check.

server/segment_rpc.go
```diff
@@ -328,6 +330,18 @@ func SubmitSegment(sess *BroadcastSession, seg *stream.HLSSegment, nonce uint64)
 		return nil, err
 	}
 
+	ratPrice, _ := ratPriceInfo(sess.OrchestratorInfo.PriceInfo)
```
yondonfu: Let's handle the error

server/segment_rpc.go
```diff
@@ -328,6 +330,18 @@ func SubmitSegment(sess *BroadcastSession, seg *stream.HLSSegment, nonce uint64)
...
+	faceVal := new(big.Int).SetBytes(sess.OrchestratorInfo.GetTicketParams().GetFaceValue())
```
yondonfu: If we move the log statement into [this code block](https://github.com/livepeer/go-livepeer/blob/master/server/segment_rpc.go#L609) in `genPayment()` then we can avoid additional conversation steps and exporting `maxWinProb` from the `pm` package.

```
batch.TicketParams.Recipient
batch.TicketParams.FaceValue
batch.Tickets()[0].WinProbRat()
```

`WinProbRat()` is only defined on `pm.Ticket` right now, but might make sense to do something like:

```
type ticketParams struct {
   // All fields of TicketParams besides Seed
}

func (tp *ticketParams) WinProbRat() *big.Rat {}

// Embed ticketParams in TicketParams to expose WinProbRat()
// Embed ticketParams in Ticket to expose WinProbRat()
// Then in the above example we could do batch.TicketParams.WinProbRat()
```

yondonfu (changes requested):

server/segment_rpc.go
kyriediculous: Yup that's true, I guess I didn't like that we would log before assessing of marshalling the protobuf should yield an error but in hindsight this should never really error.

server/segment_rpc.go
kyriediculous: It's impossible for us to error on this call as if that would happen then we would already error out at `validatePrice`

server/segment_rpc.go
kyriediculous: This would get annoying pretty quickly because of not being able to use promoted fields (fields on the embedded struct) in composite literals (eg everytime we call `&Ticket{...} `) so I would be okay of just using 0 based index on Tickets here instead of having to specifically set the fields seperately when creating tickets. 

To me this represents more of a bigger annoyance for me personally in go, embedding works well for methods but not for fields within a type when using composite literals. Perhaps we should default to using constructors only in this case and consider using composite literals with embedded types as an anti-pattern. 

Issue in this specific case is that `NewTicket` takes in `TicketParams` which would also composed of the embedded `ticketParams`. 

WDYT ?

server/segment_rpc.go
kyriediculous: fixed in 5a6081e

server/segment_rpc.go
```diff
@@ -634,6 +635,15 @@ func genPayment(sess *BroadcastSession, numTickets int) (string, error) {
 		}
 
 		protoPayment.TicketSenderParams = senderParams
+
+		ratPrice, _ := ratPriceInfo(protoPayment.ExpectedPrice)
+		glog.V(common.VERBOSE).Infof("Created new payment - recipient=%v  faceValue=%v  winProb=%v  price=%v  numTickets=%v",
```
yondonfu: One space instead of two between each of the fields (to be consistent with the rest of the log messages with fields in this file)?

server/segment_rpc.go
```diff
@@ -634,6 +635,15 @@ func genPayment(sess *BroadcastSession, numTickets int) (string, error) {
...
+			ethcommon.BytesToAddress(sess.OrchestratorInfo.GetTicketParams().GetRecipient()),
```
yondonfu: Can avoid the address conversion by using `batch.Recipient.Hex()`

server/segment_rpc.go
```diff
@@ -634,6 +635,15 @@ func genPayment(sess *BroadcastSession, numTickets int) (string, error) {
...
+			eth.FormatUnits(batch.FaceValue, "ETH"),
+			new(big.Rat).Mul(batch.Tickets()[0].WinProbRat(), big.NewRat(100, 1)).FloatString(10)+" %",
```
yondonfu: Since we're logging the win prob as a value 0.0-1.0 on the orchestrator I think we could just stick with using a value 0.0-1.0 here.

server/segment_rpc.go
yondonfu: Since we include the manifestID when logging on the orchestrator think it could be good to include the manifestID here as well

yondonfu (changes requested):

server/segment_rpc.go
yondonfu: Ah I see what you're saying - yeah updating usage of the `Ticket` constructor literal would be annoying.

Other options to avoid re-constructing the list of tickets (via `batch.Tickets()`) include:
1. Define `WinProbRat()` on `TicketParams` - the impl of `Ticket.WinProbRat()` could be extracted into an internal `winProbRat()` helper that is re-used in these two methods. Then we would just call `batch.WinProbRat()` for the log statement
2. Export a `WinProbRat()` helper that accepts a `big.Int` as a param

IMO Option 1 seems like it would read a bit better (due to the context around the type that the method is being called on).

server/segment_rpc.go
```diff
@@ -634,6 +635,15 @@ func genPayment(sess *BroadcastSession, numTickets int) (string, error) {
 		}
 
 		protoPayment.TicketSenderParams = senderParams
+
+		ratPrice, _ := ratPriceInfo(protoPayment.ExpectedPrice)
+		glog.V(common.VERBOSE).Infof("Created new payment - recipient=%v  faceValue=%v  winProb=%v  price=%v  numTickets=%v",
+			ethcommon.BytesToAddress(sess.OrchestratorInfo.GetTicketParams().GetRecipient()),
+			eth.FormatUnits(batch.FaceValue, "ETH"),
```
yondonfu: If we're using this helper here we should probably use it on the [orchestrator's side](https://github.com/livepeer/go-livepeer/blob/83df0883fd82bec79dae54ef5cd86ab7af0aba10/core/orchestrator.go#L178) as well for consistency

yondonfu, line 643: If we're using this helper here we should probably use it on the [orchestrator's side](https://github.com/livepeer/go-livepeer/blob/83df0883fd82bec79dae54ef5cd86ab7af0aba10/core/orchestrator.go#L178) as well for consistency.

kyriediculous: fixed in  8891b41

server/segment_rpc.go
kyriediculous: fixed in 5549ab8

server/segment_rpc.go
kyriediculous: fixed in 1a02391

server/segment_rpc.go
kyriediculous: fixed in 1a02391

server/segment_rpc.go
kyriediculous: fixed in 1a02391

server/segment_rpc.go
kyriediculous: fixed in 1a02391

yondonfu (commented):
Changes good - let's rebase

kyriediculous:
rebased !

yondonfu (approved):
LGTM

PR #1336, opened 2020-01-30
server: Include rendition name in multipart header.

j0sh:
This saves us from having to make the ordering of rendition
results an implicit part of the multipart API contract.

* Adds a `Rendition-Name` field to the HTTP push multipart return header. 
* Adds a `Content-Disposition: attachment; filename=` field to the return header. The filename follows the format: `<rendition-name>_<seq-num>.<ext>` where the `<ext>` is either `ts` or `txt`, depending on whether a blob or URL is being returned.

Added checking for additional headers to unit tests

**Checklist:**


server/mediaserver.go
```diff
@@ -701,6 +701,7 @@ func (s *LivepeerServer) HandlePush(w http.ResponseWriter, r *http.Request) {
 		hdrs := textproto.MIMEHeader{
 			"Content-Type":   {typ},
 			"Content-Length": {strconv.Itoa(length)},
+			"Rendition-Name": {cxn.params.profiles[i].Name},
```
iameli: Confirming that this'll work with `presets` as well as `profiles` in a webhook response?

iameli (commented):
Any sense in populating the `Content-Disposition` header with a filename? Might make it a bit easier to use with existing multipart libraries; I was using [Apache Commons FileUpload](https://commons.apache.org/proper/commons-fileupload/) and it assigns those names automatically if that field is present. Perhaps something like

```
Content-Disposition: form-data; filename="P144p25fps16x9_173.ts" 
```

Usually we'd call that file `P144p25fps16x9/173.ts` but slashes in a filename field make me nervous.

server/mediaserver.go
j0sh: Yes - if a name isn't specified in the webhook, then we'll create a default name. The client will then have to do a little more work to match up with the rendition, but they should really just set a profile name if they plan to use multipart.

https://github.com/livepeer/go-livepeer/blob/6a485e197c9df6acc0067e55dc80a274fe0cba77/server/mediaserver.go#L205-L211

j0sh:
> Any sense in populating the Content-Disposition header with a filename? 
> Content-Disposition: form-data; filename="P144p25fps16x9_173.ts"
>

Good idea, the proposed scheme sounds good. We'll probably want `Content-Disposition` in addition to `Rendition-Name`. Otherwise the client has to go through another step to split the rendition name from the sequence number / extension. (And the fact we may be returning a `webhook_` prefix only makes it more exciting, because a naive split on underscores wouldn't work.)

iameli:
Yeah, totally agreed that we'd want both.

j0sh:
Added `Content-Disposition: attachment; filename="..."` in 92c00794 after some follow-up on Discord

iameli (approved):
LGTM!

j0sh:
Updated to use a `txt` extension in the filename if we're returning a URL: 9cd1f180

darkdarkdragon (approved):
LGTM

PR #1339, opened 2020-01-31
Log download duration

darkdarkdragon:
Record segment's download time on B and print in to the log

- 
- 
- 

Manually

**Checklist:**

### DIFF
--- a/drivers/drivers.go
+++ b/drivers/drivers.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"time"
 
 	"github.com/golang/glog"
 	"github.com/livepeer/go-livepeer/common"
@@ -60,6 +61,7 @@ var httpc = &http.Client{
 
 func getSegmentDataHTTP(uri string) ([]byte, error) {
 	glog.V(common.VERBOSE).Infof("Downloading uri=%s", uri)
+	started := time.Now()
 	resp, err := httpc.Get(uri)
 	if err != nil {
 		glog.Errorf("Error getting HTTP uri=%s err=%v", uri, err)
@@ -75,6 +77,7 @@ func getSegmentDataHTTP(uri string) ([]byte, error) {
 		glog.Errorf("Error reading body uri=%s err=%v", uri, err)
 		return nil, err
 	}
-	glog.V(common.VERBOSE).Infof("Downloaded uri=%s", uri)
+	took := time.Since(started)
+	glog.V(common.VERBOSE).Infof("Downloaded uri=%s dur=%s", uri, took)
 	return body, nil
 }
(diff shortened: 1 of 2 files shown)


server/segment_rpc.go
```diff
@@ -381,7 +381,7 @@ func SubmitSegment(sess *BroadcastSession, seg *stream.HLSSegment, nonce uint64)
 		}
 		return nil, fmt.Errorf(errorString)
 	}
-	glog.Infof("Uploaded segment nonce=%d seqNo=%d orch=%s", nonce, seg.SeqNo, ti.Transcoder)
+	glog.Infof("Uploaded segment nonce=%d seqNo=%d orch=%s updur=%s", nonce, seg.SeqNo, ti.Transcoder, uploadDur)
```
j0sh: Same note about `dur` . Also, do we happen to have the manifestID accessible that we can put here? `nonce=... manifestID=... seqNo=...`

(I actually was going to submit a PR which adds exactly this but will hold off if it can just be added here)

drivers/drivers.go
```diff
@@ -75,6 +77,7 @@ func getSegmentDataHTTP(uri string) ([]byte, error) {
 		glog.Errorf("Error reading body uri=%s err=%v", uri, err)
 		return nil, err
 	}
-	glog.V(common.VERBOSE).Infof("Downloaded uri=%s", uri)
+	took := time.Since(started)
+	glog.V(common.VERBOSE).Infof("Downloaded uri=%s downdur=%s", uri, took)
```
j0sh: Can we maybe use `dur=` as a consistent key for the various forms of duration? (If someone needs to search specifically for the download duration, they can grep for "Downloaded uri=" and so forth)

server/segment_rpc.go
```diff
@@ -458,7 +458,8 @@ func SubmitSegment(sess *BroadcastSession, seg *stream.HLSSegment, nonce uint64)
 		monitor.SegmentTranscoded(nonce, seg.SeqNo, transcodeDur, common.ProfilesNames(sess.Profiles))
 	}
 
-	glog.Infof("Successfully transcoded segment nonce=%d manifestID=%s segName=%s seqNo=%d orch=%s", nonce, string(sess.ManifestID), seg.Name, seg.SeqNo, ti.Transcoder)
+	glog.Infof("Successfully transcoded segment nonce=%d manifestID=%s segName=%s seqNo=%d orch=%s transdur=%s", nonce,
```
j0sh: Same note about `dur`

darkdarkdragon: changed

drivers/drivers.go
darkdarkdragon: changed

server/segment_rpc.go
darkdarkdragon: added

j0sh (approved):
LGTM! 🚢

PR #1351, opened 2020-02-04
Close LPMS Stream object in `removeRTMPStream`

### LINKED ISSUE #1348: Leaking goroutines
**Describe the bug**
In the long running (three days) broadcaster there was 1300 goroutines in this state:

```
goroutine 488903 [select, 4457 minutes]:
github.com/livepeer/lpms/stream.NewBasicRTMPVideoStream.func1(0xc001070070)
        /go/pkg/mod/github.com/livepeer/lpms@v0.0.0-20200110164555-e34a4737b857/stream/basic_rtmp_videostream.go:41 +0x23d
created by github.com/livepeer/lpms/stream.NewBasicRTMPVideoStream
        /go/pkg/mod/github.com/livepeer/lpms@v0.0.0-20200110164555-e34a4737b857/stream/basic_rtmp_videostream.go:38 +0x12b
```

**To Reproduce**
Connect to the broadcaster that served many stream
http://localhost:7935/debug/pprof/goroutine?debug=2
Count `basic_rtmp_videostream.go:41`.

**Expected behavior**
Shouldn't leak goroutines.

darkdarkdragon:
Close LPMS Stream object in `removeRTMPStream`
Prevents leaking goroutines in LPMS

Added unit test

Fixes #1348

**Checklist:**


server/push_test.go
```diff
@@ -240,11 +246,39 @@ func TestShouldUpdateLastUsed(t *testing.T) {
...
+		"github.com/rjeczalik/notify.(*nonrecursiveTree).internal"}
+
+	res := make([]goleak.Option, 0, len(funcs2ignore))
+	for _, f := range funcs2ignore {
+		res = append(res, goleak.IgnoreTopFunction(f))
+	}
+	return res
+}
+
+func serverCleanup(s *LivepeerServer) {
```
j0sh: Can this be moved into `mediaserver_test.go` along the other LivepeerServer helpers such as `setupServerWithCancel` ?

server/push_test.go
```diff
@@ -240,11 +246,39 @@ func TestShouldUpdateLastUsed(t *testing.T) {
 	assert.True(lu.Before(s.rtmpConnections["mani1"].lastUsed))
 }
 
+func ignoreRoutines() []goleak.Option {
+	funcs2ignore := []string{"github.com/golang/glog.(*loggingT).flushDaemon", "go.opencensus.io/stats/view.(*worker).start",
```
j0sh, line 252: Maybe add a comment here describing what this function does and why we need to exclude these particular functions?

server/push_test.go
j0sh: While this type of cleanup is likely the correct thing to do, curious what the effects are of not cleaning up anything? Our tests seem to have been running OK without this.

j0sh (changes requested):
While I guess we should trust that `goleak` does what it says, did you also manually sanity check that our goroutine usage seems better before / after this fix, the same way that https://github.com/livepeer/go-livepeer/issues/1348 was diagnosed?

server/push_test.go
darkdarkdragon: Yeah, I've started to writing this clean function trying to shutdown everything correctly, but then realised that there is tons of goroutines started over which I don't have control. I'll just remove this function, anyway I was forced to make long list (`funcs2ignore`) of goroutines to ignore.

server/push_test.go
darkdarkdragon: added

server/push_test.go
darkdarkdragon: Actually I was to quick to write previous comment - this function is needed because other tests too starts this goroutine for which we're checking. Added comment.

darkdarkdragon:
@j0sh 
> did you also manually sanity check 

yes. and if my fix is commented, then test will fail.

j0sh (approved):
LGTM

PR #1357, opened 2020-02-05
Raise level of log messages in the census.go

### LINKED ISSUE #1316: Reduce logging level in census.go
No need to log at the "info" level every single time we log a metric; that's what Prometheus is for. Most of this file could be lowered to debug or trace status, I think.

darkdarkdragon:
Raises level of log messages in the census.go

Manually

Fixes #1316

**Checklist:**

### DIFF
--- a/monitor/census.go
+++ b/monitor/census.go
@@ -45,6 +45,9 @@ const (
 
 	numberOfSegmentsToCalcAverage = 30
 	gweiConversionFactor          = 1000000000
+
+	logLevel = 6 // TODO move log levels definitions to separate package
+	// importing `common` package here introduces import cycles
 )
 
 // Enabled true if metrics was enabled in command line
@@ -751,7 +754,7 @@ func TranscodeTry(nonce, seqNo uint64) {
 		} else {
 			av.tries[seqNo] = tryData{tries: 1, first: time.Now()}
 		}
-		glog.Infof("Trying to transcode segment nonce=%d seqNo=%d try=%d", nonce, seqNo, try)
+		glog.V(logLevel).Infof("Trying to transcode segment nonce=%d seqNo=%d try=%d", nonce, seqNo, try)
 	}
 }
 
@@ -764,7 +767,7 @@ func SetTranscodersNumberAndLoad(load, capacity, number int) {
 }
 
 func SegmentEmerged(nonce, seqNo uint64, profilesNum int) {
-	glog.Infof("Logging SegmentEmerged... nonce=%d seqNo=%d", nonce, seqNo)
+	glog.V(logLevel).Infof("Logging SegmentEmerged... nonce=%d seqNo=%d", nonce, seqNo)
 	census.segmentEmerged(nonce, seqNo, profilesNum)
 }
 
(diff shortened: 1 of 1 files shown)


monitor/census.go
```diff
@@ -45,6 +45,9 @@ const (
 
 	numberOfSegmentsToCalcAverage = 30
 	gweiConversionFactor          = 1000000000
+
+	logLevel = 5 // TODO move log levels definitions to separate package
```
j0sh: How about 6 / VERBOSE ?

j0sh (changes requested):

monitor/census.go
darkdarkdragon: changed

monitor/census.go
```diff
@@ -782,8 +789,10 @@ func (cen *censusMetricsCounter) segmentEmerged(nonce, seqNo uint64, profilesNum
 }
 
 func SourceSegmentAppeared(nonce, seqNo uint64, manifestID, profile string) {
-	glog.Infof("Logging SourceSegmentAppeared... nonce=%d seqNo=%d manifestid=%s profile=%s", nonce,
-		seqNo, manifestID, profile)
+	if glog.V(logLevel) {
+		glog.Infof("Logging SourceSegmentAppeared... nonce=%d seqNo=%d manifestid=%s profile=%s", nonce,
```
j0sh: I know this was here before, but since we're in the process of cleaning this up - don't we do `manifestID`  rather than `manifestid` ? Grep is case sensitive :)

Also, I think most of our logging goes in this order `manifestID=... seqNo=...`

darkdarkdragon: changed

monitor/census.go
```diff
@@ -751,7 +754,9 @@ func TranscodeTry(nonce, seqNo uint64) {
 		} else {
 			av.tries[seqNo] = tryData{tries: 1, first: time.Now()}
 		}
-		glog.Infof("Trying to transcode segment nonce=%d seqNo=%d try=%d", nonce, seqNo, try)
+		if glog.V(logLevel) {
+			glog.Infof("Trying to transcode segment nonce=%d seqNo=%d try=%d", nonce, seqNo, try)
```
iameli: Do you think this would be cleaner as `glog.V(loglevel).Info("foo")`? Would save having to write a lot of `if` statements.

iameli (dismissed):

monitor/census.go
j0sh: Yeah, that's the convention we use elsewhere. Presumably @darkdarkdragon  was trying to save a few cycles since this form doesn't evaluate the arguments to `Infof` , but I agree it's very verbose.

I'm indifferent about `census.go` but would rather not see this pattern proliferate to the rest of the codebase.

darkdarkdragon: @iameli @j0sh removed `if V()` pattern

j0sh:
Looks OK, but do we know why CI is failing?

darkdarkdragon:
not because of this PR:
```
livepeer.go:325] Failed to connect to Ethereum client: websocket: bad handshake (HTTP status 403 Forbidden)
```

j0sh:
Let's squash these commits into one

darkdarkdragon:
@j0sh squashed

j0sh (approved):
🚢

Issue #1327, opened 2020-01-27
verification: Distinguish between fatal vs. non-fatal retryable errors


yondonfu:
wrap `Retryable` errors with a new `fatal` error type and add a check for it just like `IsRetryable` called `IsFatal`

Fatal errors should not be included in the results from `verification.Verify()`
Fatal errors should not trigger a `sv.count++` in `verification.Verify()`
Rather than using `sv.count`, we can just rely on the length of the results slice.

Fatal errors include:
* `ErrAudioMismatch`
* `ErrPixelMismatch`

Non-fatal:
* `ErrTampered`

Issue #1325, opened 2020-01-27
Move sig verification into verification.SegmentVerifier


yondonfu:
**Is your feature request related to a problem? Please describe.**

The broadcaster currently runs orchestrator sig verification for transcoded results [here](https://github.com/livepeer/go-livepeer/blob/9cb8a3ddb7492434284cd00509193f53cacd4515/server/broadcast.go#L489).

There are a few downsides with how we handle the orchestrator sig verification for transcoded results right now:

- It is run after the results are inserted into the playlist. We should verify the sig before inserting into the playlist
- The check is outside of `verification.SegmentVerifier` which contains logic for content verification logic. Ideally, all verification logic for transcoded results could live in the same package in order to avoid fragmented verification operations and to improve code readability

**Describe the solution you'd like**

Move the orchestrator sig verification into `verification.SegmentVerifier` (see this [TODO](https://github.com/livepeer/go-livepeer/blob/9cb8a3ddb7492434284cd00509193f53cacd4515/verification/verify.go#L101)).

Sig verification should only be run when the node is on-chain mode. One way to check if the node is in on-chain mode is to check if `params.Orchestrator.TicketParams != nil` (we would only be receiving ticket params from an orchestrator when running in on-chain mode).

A few things to consider:

The [VerifySig()](https://github.com/livepeer/go-livepeer/blob/9cb8a3ddb7492434284cd00509193f53cacd4515/pm/helpers.go#L22) helper is currently defined in the `pm` package. We can avoid importing the `pm` package into the `verification` package by:
- Moving the `VerifySig()` helper into the `eth` package and removing the `DefaultSigVerifier` struct from the `pm` package - the `pm` package doesn't need to be exposed to the details of how sig verification is implemented
- Defining a `SigVerifier` struct that calls `VerifySig()` in the `eth` package. This struct can be passed in as a dependency of [pm.Validator](https://github.com/livepeer/go-livepeer/blob/9cb8a3ddb7492434284cd00509193f53cacd4515/pm/validator.go#L39)

[...]

PR #1360, opened 2020-02-05
Move pixel count verification into verification.SegmentVerifier

### LINKED ISSUE #1326: Move pixel count verification into verification.SegmentVerifier

The broadcaster currently runs pixel count verification for transcoded results [here](https://github.com/livepeer/go-livepeer/blob/9cb8a3ddb7492434284cd00509193f53cacd4515/server/broadcast.go#L426).

A downside with how we handle the pixel count verification for transcoded results right now is that the check is outside of `verification.SegmentVerifier` which contains logic for content verification. Ideally, all verification logic for transcoded results could live in the same package in order to avoid fragmented verification operations and to improve code readability. Furthermore, there is already error handling for failed pixel count verification when it is run by the external verifier used by `verification.SegmentVerifier` (i.e. `EpicClassifier`) - we will want to use the same error handling logic even if pixel count verification is *not* run by the external verifier. This would be easier if related code is grouped together as opposed to scattered across packages.

Move the pixel count verification into `verification.SegmentVerifier`. Pixel count verification should only be run when the node is on-chain mode. One way to check if the node is on-chain mode is to check if `params.Orchestrator.TicketParams != nil` (we would only be receiving ticket params from an orchestrator when running in on-chain mode).

Pixel count verification should be run in two cases within the `Verify()` method of `SegmentVerifier`:

- If the policy's `Verifier` is nil
- If the policy's `Verifier` does not implement pixel count verification (see this [TODO](https://github.com/livepeer/go-livepeer/blob/9cb8a3ddb7492434284cd00509193f53cacd4515/verification/verify.go#L113)).

### LINKED ISSUE #1328: Default verification policy for pixel count & sig verification when in on-chain mode

Once we move sig and pixel count verification into `verification.SegmentVerifier` (related #1325 #1326) we will need to make sure that sig and pixel count verification are still run when the node is in on-chain mode event if the node is not using an external verifier via the `-verifierUrl` flag. 

Currently, we only instantiate a policy if a value is passed in for `-verifierUrl`.

We could instantiate an "empty" policy with a nil verifier even when `-verifierUrl` is not specified. Maybe something like:

```
var verifier *verification.EpicClassifier
if *verifierUrl != "" {
    verifier = &verification.EpicClassifier{Addr: *verifierUrl}
}

server.Policy = &verification.Policy{Retries: 2, Verifier: verifier}
```

This assumes that `verification.SegmentVerifier` will have on-chain mode checks to determine if it should run sig & pixel count verification.

mk-livepeer:
This refactors the code related to the `verifyPixels` function to exist entirely within the `verification` module.  From now on, callers can simply call `Verify()` which will, in turn, call `verifyPixels()`

closes #1326 
closes #1328


verification/verify_test.go
```diff
@@ -135,3 +150,59 @@ func TestVerify(t *testing.T) {
...
+
+func TestVerifyPixels(t *testing.T) {
+	ffmpeg.InitFFmpeg()
+
+	require := require.New(t)
+	assert := assert.New(t)
+
+	// Create memory session and save a test file
+	bos := drivers.NewMemoryDriver(nil).NewSession("foo")
+	data, err := ioutil.ReadFile("test.flv")
```
j0sh: Can we make this a relative path, eg "../server/test.flv" ? This would avoid (re-)adding the large "test.flv" blob, which Git doesn't really handle well.

mk-livepeer: done in 0670fe4097f2d7312bdc818f2aa5d041ec565b80

verification/verify.go
```diff
@@ -112,7 +121,8 @@ func (sv *SegmentVerifier) Verify(params *Params) (*Params, error) {
 		if len(res.Pixels) != len(params.Results.Segments) {
 			// TODO make allowances for the verification algos not doing
```
yondonfu: Let's remove this comment if we're handling this case.

verification/verify.go
```diff
@@ -153,3 +163,80 @@ func IsRetryable(err error) bool {
...
+			return fmt.Errorf("error creating temp file for pixels verification: %v", err)
+		}
+		defer os.Remove(tempfile.Name())
+
+		if _, err := tempfile.Write(data); err != nil {
+			tempfile.Close()
+			return fmt.Errorf("error writing temp file for pixels verification: %v", err)
+		}
+
+		if err = tempfile.Close(); err != nil {
```
yondonfu, line 236: Could we call `Close()` with a `defer` after the `defer` statement that calls `os.Remove()` to avoid calling it in 2 places (this is what was done in this function before this PR)?

verification/verify.go
```diff
@@ -153,3 +163,80 @@ func IsRetryable(err error) bool {
...
+	// check if the node is on-chain mode
+	if params != nil &&
+		params.Orchestrator != nil &&
+		params.Orchestrator.TicketParams != nil {
+
+		var err error
+		for i := 0; err == nil && i < len(params.Results.Segments); i++ {
+			if err := verifyPixels(
+				params.Results.Segments[i].Url,
+				flatten(params.Renditions),
```
yondonfu: We want to pass in `params.Renditions[i]` here so that in the case where in-memory storage is used, `params.Renditions[i]` is the data corresponding to `params.Results.Segments[i].Url`.

verification/verify.go
```diff
@@ -112,7 +121,8 @@ func (sv *SegmentVerifier) Verify(params *Params) (*Params, error) {
 		if len(res.Pixels) != len(params.Results.Segments) {
 			// TODO make allowances for the verification algos not doing
 			//      pixel counts themselves; adapt broadcast.go verifyPixels
-			err = ErrPixelsAbsent
+
+			err = verifyPixelParams(params)
```
yondonfu: Looking at this now I'm realizing that there could be an opportunity to simplify things a bit since the for loop below already performs the comparison of computed pixels vs. reported pixels for each of the rendition segments. At the moment, we're repeating this comparison each time we run `verifyPixelParams()`. It'd be nice if we can just count pixels when we need to and then rely on the below for loop for the comparison of computed pixels vs. reported pixels.

The below for loop relies on the `Pixels` field of the `Results` object returned from `policy.Verifier.Verify()`. We could create a type that implements the `Verifier` interface and its `Verify()` method would be similar to the contents of the current `verifyPixelsParams()` function except instead of doing the comparison against reported pixels it would just populate the `Pixels` field of the returned `Results` object. This would be the "default" implementation of the `Verifier` interface (related to #1328) that would be used unless an external verifier is specified (would still need the on-chain check that we currently have in `verifyPixelsParams()`). WDYT?

yondonfu (changes requested):

verification/verify.go
mkrufky: That was a bug and I fixed it.  It is incorrect, bad practice to defer close() on writeable files
https://www.joeshaw.org/dont-defer-close-on-writable-files

verification/verify.go
mkrufky: I think it all sounds nice, but it is scope creep.  Please write this into a new ticket so that we can close this one and demonstrate progress.

verification/verify.go
mk-livepeer: done in 7fe1d472e80161d03b3bf5643495c8b89d531024

verification/verify.go
yondonfu: I see your point about my suggestion representing additional work that is a bit different from the original suggested implementation approach in #1326. That being said, it represents an alternative approach to accomplishing the same objective that could be cleaner so I think its worthwhile having the discussion about whether we prefer the alternative approach or the current approach.

I brought it up because I think it could be a cleaner approach code-wise that doesn't seem to require a significant amount of additional effort so I wanted to see what you think! If we do prefer the alternative approach then I think it makes sense to address it now rather than committing code that is only then to be immediately changed afterwords.

verification/verify.go
mk-livepeer: done in c2c6653cd125ecae82c816ce49fbd66a9bf75f29

verification/verify.go
yondonfu: I think the previous usage was probably ok given the current implementation of `file.Close()` and the surrounding code, but agreed that always handling errors even in such a situation is better. 

We could use defer here *and* handle the error within the defer, but I can understand a dislike for that pattern, so I'm ok with how the `Close()` calls are currently structured.

verification/verify.go
```diff
@@ -110,9 +119,7 @@ func (sv *SegmentVerifier) Verify(params *Params) (*Params, error) {
 	// Check pixel counts
 	if (err == nil || IsRetryable(err)) && res != nil && params.Results != nil {
 		if len(res.Pixels) != len(params.Results.Segments) {
-			// TODO make allowances for the verification algos not doing
-			//      pixel counts themselves; adapt broadcast.go verifyPixels
-			err = ErrPixelsAbsent
+			err = verifyPixelParams(params)
```
yondonfu: An additional thought about the approach of using a default implementation of the `Verifier` interface - that implementation actually wouldn't even need to do the pixel counting if we're also accommodating for verifiers that don't do pixel counting here.

verification/verify.go
mkrufky: I'm not even sure I understand what you're asking for.   For that reason, alone, this should be a separate ticket.  What you're asking for seems to be a new feature / optimization that depends on this PR to be merged beforehand.  What if this optimization causes some problems in the future?  That doesn't mean the refactoring needs to be reverted - it meant the optimization itself needs debugging.   I don't mind doing the work, but it *is* scope creep and I would prefer to see it on a new ticket.

mkrufky: I'm not even sure I understand what you're asking for. For that reason, alone, this should be a separate ticket. What you're asking for seems to be a new feature / optimization that depends on this PR to be merged beforehand. What if this optimization causes some problems in the future? The entire refactoring should not be put into question - they are separate changes. I don't mind doing the work, but it is scope creep and I would prefer to see it on a new ticket.

mkrufky: I think I see what you're talking about now, but it sounds like some heavy lifting.  I think this should be merged as-is first, if it is correct.  This way, we can introduce this first as a stable change before introducing new optimizations.

verification/verify.go
```diff
@@ -153,3 +160,75 @@ func IsRetryable(err error) bool {
...
+}
+
+func verifyPixels(fname string, data []byte, reportedPixels int64) error {
+	uri, err := url.ParseRequestURI(fname)
+	// If the filename is a relative URI and the broadcaster is using local memory storage
+	// write the data to a temp file
+	if err == nil && !uri.IsAbs() && data != nil {
+		tempfile, err := ioutil.TempFile("", common.RandName())
+		if err != nil {
+			return fmt.Errorf("error creating temp file for pixels verification: %v", err)
```
j0sh: Even though this bit was copy-pasted from the old code... since I think we basically require Go 1.13 now, maybe we could use the %w modifier to make the original error type introspectable? See https://blog.golang.org/go1.13-errors

verification/verify.go
```diff
@@ -153,3 +160,75 @@ func IsRetryable(err error) bool {
...
+			if i < len(params.Renditions) {
+				rendition = params.Renditions[i]
+			}
+			if err := verifyPixels(
+				params.Results.Segments[i].Url,
+				rendition,
+				params.Results.Segments[i].Pixels); err != nil {
+
+				glog.Error(err)
+				return ErrPixelMismatch
```
j0sh: Is there a particular reason for dropping the returned error here and using `ErrPixelMismatch` ?

One side effect that comes to mind is that `ErrPixelMismatch` is marked as retryable (eg, re-transcoding with another O might fix the problem), but most of the other potential errors from `verifyPixels` are IO-related. It's less likely that a retry would fix IO errors. For example, IO errors are not retried in the [epic classifier](https://github.com/livepeer/go-livepeer/blob/0670fe4097f2d7312bdc818f2aa5d041ec565b80/verification/epic.go).

Also if we propagate the error from verifyPixels then we probably wouldn't need to print it or import `glog`. (It was kind of neat not having to import any logging here, but that probably can't last.)

verification/verify.go
mk-livepeer: thanks :-). done in e96c310d6d949b8dd76cc857f3b1b0949a8bc46a

verification/verify.go
mk-livepeer: We have to return `ErrPixelMismatch` here in order to pass the existing tests. We can change it back, along with the tests, but I'm not sure if that's what we want.  I am indifferent about this - maybe @yondonfu can weigh in?

verification/verify.go
```diff
@@ -153,3 +160,75 @@ func IsRetryable(err error) bool {
 	_, retryable := err.(Retryable)
 	return retryable
 }
+
+func verifyPixelParams(params *Params) error {
+	// check if the node is on-chain mode
```
j0sh: Is the onchain check mandatory? I never had an issue testing verification in offchain mode.

If we still need the check, it might be good to exit early to save one level of indentation. Like:

```golang
if <offchain> {
  return
}
... rest of the code
```

verification/verify.go
```diff
@@ -101,6 +107,9 @@ func (sv *SegmentVerifier) Verify(params *Params) (*Params, error) {
 	// TODO sig checking; extract from broadcast.go
 
 	if sv.policy.Verifier == nil {
+		if err := verifyPixelParams(params); err != ErrPixelsAbsent {
```
j0sh: The `ErrPixelsAbsent` is returned if offchain mode is used, right? Assuming we still need the onchain/offchain distinction, could `nil` be returned  for onchain from `verifyPixelParams`, then things simplify to:

```golang
if sv.policy.Verifier == nil {
  return nil, verifyPixelParams(params)
}
```

verification/verify.go
yondonfu: Previously, an off-chain B wouldn't run pixel count verification and since we're moving pixel count verification into `SegmentVerifier` we'll need an on-chain check to preserve that behavior (given the assumption that we still don't want to run pixel count verification for off-chain Bs due to the unnecessary overhead).

>  it might be good to exit early to save one level of indentation

I agree that returning early if the `params.OrchestratorTicketParams == nil` helps with readability.

verification/verify.go
```diff
@@ -153,3 +160,75 @@ func IsRetryable(err error) bool {
...
+	// check if the node is on-chain mode
+	if params != nil &&
+		params.Orchestrator != nil &&
+		params.Orchestrator.TicketParams != nil {
+
+		var err error
+		for i := 0; err == nil && i < len(params.Results.Segments); i++ {
+			rendition := []byte{}
+			if i < len(params.Renditions) {
+				rendition = params.Renditions[i]
```
yondonfu: Would we be able to avoid allocating an extra var here if `i < len(params.Renditions` was a part of the for loop conditional? i.e. `... i < len(params.Results.Segments) && i < len(params.Renditions)`

verification/verify.go
j0sh: Wouldn't we still need a default verifier in order to actually invoke this pixel checking? (Presumably the default verifier would be instantiated in onchain mode if the epic classifier wasn't).

But I'm indifferent as to whether the verification should remain here or moved within the default verifier. @yondon suggestion makes sense ; keeping it here would mitigate the requirement to re-implement pixel checking in future verifiers (although probably still desirable to do as a performance optimization).

verification/verify.go
yondonfu: I don't think we want to retry on I/O errors so I think it makes sense to propagate the error and to adjust the tests.

verification/verify.go
j0sh: I think what may be missing here is there is no default verification policy? So does this bit of code actually get executed at all?

If If we instantiate a default verification policy only in onchain mode, that would mitigate the need for any on-chain checks here, which feels a bit out of place.

verification/verify_test.go
```diff
@@ -135,3 +150,59 @@ func TestVerify(t *testing.T) {
...
+	ffmpeg.InitFFmpeg()
+
+	require := require.New(t)
+	assert := assert.New(t)
+
+	// Create memory session and save a test file
+	bos := drivers.NewMemoryDriver(nil).NewSession("foo")
+	data, err := ioutil.ReadFile("../server/test.flv")
+	require.Nil(err)
+	fname, err := bos.SaveData("test.ts", data)
```
yondonfu: Do we need the memory session at all here? Seems like we can just pass the `data` var initialized above directly into `verifyPixels`?

verification/verify.go
yondonfu: Here is a sketch of what I was thinking.

The default verifier would look something like this:

```
type DefaultVerifier struct {}

func (v *DefaultVerifier) Verify() (*Results, error) {
    return &Results{Score: 1.0}, nil
}
```

this would be passed into the policy set for `server.Policy` when the node is in on-chain mode (based off [this comment](https://github.com/livepeer/go-livepeer/pull/1360/files#r376144499))

`SegmentVerifier.Verify()` would do something like this:

```
res, err := sv.policy.Verifier.Verify(params)

if (...) {
    if len(res.Pixels) != len(params.Results.Segments) {
        res.Pixels, err = countSegmentPixels(params.Results.Segments, params.Renditions)
    }
    // Compare res.Pixels against reported pixels for each rendition here
}
```

The default verifier basically does nothing and the built-in accommodation for verifiers that don't implement pixel counting in `SegmentVerifier` would be executed. `countSegmentPixels()` loops through the list of provided segments and returns a slice of pixel counts.

verification/verify.go
yondonfu: > I think what may be missing here is there is no default verification policy? So does this bit of code actually get executed at all?

Yeah you're right - this wouldn't be executed until #1328 (describes an empty policy with a nil verifier) is implemented. Though your point about setting `server.Policy` only when the node is in on-chain mode makes sense - removing the on-chain check is nice.

verification/verify.go
yondonfu: @mk-livepeer Does the above comment clear things up a bit (might not have explained well the first time around)?

If you prefer to stick with the current approach and explore the approach described in the above comment separately as an optimization then that is fine by me and I'll leave it up to you.

j0sh: Thanks for the explanation @yondonfu - that helped me understand it better. So I guess we'd be talking about incorporating two distinct code paths here:

* The `verifyPixelParams` as implemented here which also does the pixel counting / checking . This would be run if the policy does *not* specify a verifier
* The proposed `countSegmentPixels` which only does the pixel counting. The default policy *could* define a stub verifier to leverage this.

Perhaps one way to unify the two would be to do the following:

```golang
if sv.policy.Verifier == nil {
  res = &Results{ Score: 1.0 }
} else {
  res, err = sv.policy.Verifier.Verify(params)
}

if (...) {
    if len(res.Pixels) != len(params.Results.Segments) {
        res.Pixels, err = countSegmentPixels(params.Results.Segments, params.Renditions)
    }
    // Compare res.Pixels against reported pixels for each rendition here
}
```

This is nice because it avoids having to define a struct for a stub verifier, and keeps the overall flow consistent with the only difference being the pixels are explicitly counted if the verifier doesn't do it.

verification/verify.go
mkrufky: If we do that, then you could end up with a false positive.  If, somehow, the number of segments is not equal, then it should not pass just because the first x segments match - it should fail.

verification/verify.go
mkrufky: done in 458cfcf64197302e40ac39ab4511f4218f8f3740 & 86718ed9956dd3536ee4296aab48c40b05982e98

verification/verify.go
mkrufky: fixed in 10dfcc561387ae04b10531de3abb3dfae4d24b2b for early return when not in on-chain mode

verification/verify.go
mkrufky: done in d1eb9160b7957ec91343d28bd1a8a18bc2c63d68

verification/verify.go
mkrufky: > If you prefer to stick with the current approach and explore the approach described in the above comment separately as an optimization then that is fine by me and I'll leave it up to you.

Yes, @yondonfu , I would greatly prefer to work on this optimization in a separate ticket.  If all else looks good in this PR, please approve.

j0sh: My feeling is this is still in scope - the `verifyPixelParams` is a new function on top of the refactor. Yondon's [suggestion](https://github.com/livepeer/go-livepeer/pull/1360#discussion_r376145833) combined with the [refinement](https://github.com/livepeer/go-livepeer/pull/1360#discussion_r376238249) means we end up with better code that's both simpler and more straightforward: two invocations of `verifyPixelParams` (count + check) can be replaced with a single invocation of `countPixels` (count-only) , and we reuse the existing pixel-checking code. Would rather do this now while we're working in this area; deferring it just increases friction and code churn.

server/broadcast_test.go
```diff
@@ -580,12 +579,8 @@ func TestTranscodeSegment_VerifyPixels(t *testing.T) {
 	// Wait for async pixels verification to finish
 	time.Sleep(1 * time.Second)
 
-	// Check that the session was removed because we are in on-chain mode
-	_, ok = bsm.sessMap[ts.URL]
-	assert.False(ok)
-
 	// Create stub response with correct reported pixels
-	p, err := pixels("test.flv")
+	p, err := verification.Pixels("test.flv")
```
j0sh: I thin this test is basically a series of no-ops right now due to a default policy not being defined, so the verification code never gets executed - see https://github.com/livepeer/go-livepeer/pull/1360#discussion_r376148479 .

I'm thinking that this test can be removed as long as we add something that ensures pixel counting is invoked from the SegmentVerifier.Verify function if a nil verifier is used. There was a test added that does a failure check; we should also check for success as well.

verification/verify_test.go
```diff
@@ -70,6 +74,17 @@ func TestVerify(t *testing.T) {
 	assert.Nil(err)
 	assert.NotNil(res)
 
+	// check pixel count fails w/o external verifier
```
j0sh, line 79: This is a good check for a failure condition. Might be good to also add another check that pixel count succeeds without an external verifier, probably by adapting some of the checks in TestTranscodeSegment_VerifyPixels (which can then be removed as per this [comment](https://github.com/livepeer/go-livepeer/pull/1360/files#r376504083)).

verification/verify.go
mkrufky: It feels like too many moving parts.  I'm attempting to address today's comments, but each small nit results in breaking half the tests.  How about a compromise?  Let's make a `staging` branch, as I've proposed in the past.  We can merge this to `staging` once all points are addressed besides this one, and then it can unblock my work on #1325 which depends on the merge of this PR.  Not to mention the additional work in #1328   ...   So I'm proposing that we fork master to a `staging` (or `dev`) branch and I rebase my branches-in-progress off that.  After the merge, the delta will be greatly reduced, there will be less moving parts, and it will be a lot easier for me to see what Yondon is proposing here.

yondonfu: Thought about this some more and I do think that it would be good idea to update the current implementation given that Josh's [refined approach](https://github.com/livepeer/go-livepeer/pull/1360#discussion_r376238249) should result in less code and less complexity which should make any additional work we have do in this area of the codebase easier.

> It feels like too many moving parts.

I think the update should result in less moving parts in this PR since there will be less code overall and fewer code paths that a reader needs to follow.

> and then it can unblock my work on #1325

If code in this PR is a dependency for #1325 we can always branch off this PR. I see your point about merging PRs to a `staging` branch that contain code changes in an intermediate state (i.e. there will be a follow up PR to bring the code changes to a finished state), but I think that can be a discussion separate from the issue at hand which is the desire to update the implementation in this PR given that a simpler alternative is available.

> Not to mention the additional work in #1328

Josh makes a good point in [this comment](https://github.com/livepeer/go-livepeer/pull/1360#commitcomment-37177513) that it would make sense to address #1328 in this PR given that how we structure this PR will impact how we address that issue.

verification/verify.go
mkrufky: Optimized in 513a646421c01079297b66f40daa5f18040f7b97 to simply exit the function with `ErrPixelMismatch` when `len(params.Results.Segments) != len(params.Renditions)`

server/broadcast_test.go
mkrufky: Test removed & pixels() function unexported in 3bd8bc22820bcbb5061579b1a30f31bd56447349

verification/verify_test.go
mkrufky: Added in c0d756c09e4cb652726d5b4431cea7f9c8684e1d , but I had to add some ugly checks to determine whether to return `nil, err` or `params, err` ...  I'd imagine this might need some adjusting

mkrufky: Actually, something is definitely wrong here.  The test passes, but I populated Params.Renditions with bogus data.

verification/verify.go
mkrufky: how's this?  refactored in 6aa915f5d7256007052c99752b38355cb0f8bab7

mk-livepeer: now integrated the solution for #1328

verification/verify.go
```diff
@@ -153,3 +160,68 @@ func IsRetryable(err error) bool {
...
+
+	if params.Orchestrator == nil {
+		return nil, ErrPixelsAbsent
+	}
+
+	if len(params.Results.Segments) != len(params.Renditions) {
+		return nil, ErrPixelMismatch
+	}
+
+	res := &Results{Pixels: make([]int64, len(params.Results.Segments))}
```
yondonfu: This will overwrite the `*Results` returned by `sv.policy.Verifier.Verify()` on L114. This will be a problem in the scenario where the policy uses a verifier that set the `Score` in the returned `*Results` to X, but the verifier does not populate the `Pixels` field. L114 will replace the existing `*Results` with one that has `Pixels` populated, but a `Score` of 0.

We could avoid this problem by returning a `[]int64` here with the pixel counts and only setting the `Pixels` field of the `*Results` on L114.

verification/verify.go
```diff
@@ -153,3 +160,68 @@ func IsRetryable(err error) bool {
 	_, retryable := err.(Retryable)
 	return retryable
 }
+
+func (sv *SegmentVerifier) countPixelParams(params *Params) (*Results, error) {
+
+	if params.Orchestrator == nil {
```
yondonfu: I think `params.Orchestrator` should always be non-nil if B is verifying results from O (otherwise we wouldn't have any results from O) so I don't think this nil check is needed here. Plus doesn't look like we reference it anywhere in this method.

cmd/livepeer/livepeer.go
```diff
@@ -695,6 +695,8 @@ func main() {
 				glog.Fatal("Requires a path to the verifier shared volume when local storage is in use; use -verifierPath, S3 or GCS")
 			}
 			verification.VerifierPath = *verifierPath
+		} else if *network != "offchain" {
```
yondonfu, line 698: After #1368 is merged into this branch, if the user specifies in `-verifierUrl`, but is running in off-chain mode, [the signature verification](https://github.com/livepeer/go-livepeer/pull/1368/files#diff-98d66d2dc5372dedc76d5fdde7bedebcR130) will always fail because there will be no signature for results returned from an off-chain O. We should be able to handle this case by only setting `server.Policy` if the node is in on-chain mode even if `-verifierUrl` is specified. Maybe something like:

```
if *network != "offchain" {
    // Set default policy
    // There is already an existing check here https://github.com/livepeer/go-livepeer/blob/10873651d9b4a73cc68153fb163fb63750de9543/cmd/livepeer/livepeer.go#L645

    if *verifierUrl != "" {
        // Set up the policy using the verifier with verifierUrl
        // Corresponds to https://github.com/livepeer/go-livepeer/blob/10873651d9b4a73cc68153fb163fb63750de9543/cmd/livepeer/livepeer.go#L683
    } else {
        // Set default policy
    }
}
```

verification/verify.go
```diff
@@ -100,19 +105,21 @@ func (sv *SegmentVerifier) Verify(params *Params) (*Params, error) {
...
-	}
+	var err error
+	res := &Results{}
+
 	// TODO Use policy sampling rate to determine whether to invoke verifier.
 	//      If not, exit early. Seed sample using source data for repeatability!
-	res, err := sv.policy.Verifier.Verify(params)
+	if sv.policy.Verifier != nil {
+		res, err = sv.policy.Verifier.Verify(params)
+	} else if params.Orchestrator == nil {
```
yondonfu: I think `params.Orchestrator` should always be non-nil if B is verifying results from O (otherwise we wouldn't have any results from O) so I don't think this nil check is needed here.

verification/verify_test.go
```diff
@@ -57,19 +61,42 @@ func TestVerify(t *testing.T) {
...
+
+	// Check pixel count succeeds w/o external verifier
+	pixels, err := pixels("../server/test.flv")
+	assert.Nil(err)
+	data = &net.TranscodeData{Segments: []*net.TranscodedSegmentData{
+		{Url: "../server/test.flv", Pixels: pixels},
+		{Url: "../server/test.flv", Pixels: pixels},
+	}}
+	renditions := [][]byte{{123}, {234}}
+	res, err = sv.Verify(&Params{Results: data, Orchestrator: &net.OrchestratorInfo{TicketParams: &net.TicketParams{}}, Renditions: renditions})
```
yondonfu, line 153: Technically we don't need `TicketParams` here in this PR, but we'll need it for once #1368 is merged. Maybe omit in this PR, but introduce it in #1368 for the tests needed there (to test sig verification when running `SegmentVerifier.Verify()`).

yondonfu (changes requested):

verification/verify_test.go
mkrufky: Why not just leave it then, if this PR wont be merged without 1368 ?

verification/verify_test.go
mkrufky: ...ugly checks removed

verification/verify.go
mkrufky: In order to preserve the current behavior, we may need to keep it. Either that, or something else may be wrong.

With it removed by the following patch:
```diff
diff --git a/verification/verify.go b/verification/verify.go
index 6a429060..9a10b82b 100644
--- a/verification/verify.go
+++ b/verification/verify.go
@@ -165,10 +165,6 @@ func IsRetryable(err error) bool {
 
 func (sv *SegmentVerifier) countPixelParams(params *Params) ([]int64, error) {
 
-       if params.Orchestrator == nil {
-               return nil, ErrPixelsAbsent
-       }
-
        if len(params.Results.Segments) != len(params.Renditions) {
                return nil, ErrPixelMismatch
        }
```
many tests fail:
```
$ go test
E0211 16:05:55.057727   31411 epic.go:151] Could not submit request Post : unsupported protocol scheme ""
--- FAIL: TestVerify (1.33s)
    verify_test.go:62: 
        	Error Trace:	verify_test.go:62
        	Error:      	Not equal: 
        	            	expected: *errors.errorString(&errors.errorString{s:"PixelsAbsent"})
        	            	actual  : verification.Retryable(verification.Retryable{error:(*errors.errorString)(0xc00007d620)})
        	Test:       	TestVerify
    verify_test.go:138: 
        	Error Trace:	verify_test.go:138
        	Error:      	Not equal: 
        	            	expected: *errors.errorString(&errors.errorString{s:"PixelsAbsent"})
        	            	actual  : verification.Retryable(verification.Retryable{error:(*errors.errorString)(0xc00007d620)})
        	Test:       	TestVerify
    verify_test.go:139: 
        	Error Trace:	verify_test.go:139
        	Error:      	Not equal: 
        	            	expected: "mno"
        	            	actual  : "pqr"
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-mno
        	            	+pqr
        	Test:       	TestVerify
    verify_test.go:143: 
        	Error Trace:	verify_test.go:143
        	Error:      	Not equal: 
        	            	expected: *errors.errorString(&errors.errorString{s:"PixelsAbsent"})
        	            	actual  : verification.Retryable(verification.Retryable{error:(*errors.errorString)(0xc00007d620)})
        	Test:       	TestVerify
    verify_test.go:144: 
        	Error Trace:	verify_test.go:144
        	Error:      	Not equal: 
        	            	expected: "mno"
        	            	actual  : "stu"
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-mno
        	            	+stu
        	Test:       	TestVerify
demuxer: Unable to open file
Freeing input based on OPEN INPUT error
E0211 16:05:56.387836   31411 ffmpeg.go:292] Transcoder Return : No such file or directory
demuxer: Unable to open file
Freeing input based on OPEN INPUT error
E0211 16:05:56.841673   31411 ffmpeg.go:292] Transcoder Return : No such file or directory
FAIL
exit status 1
FAIL	github.com/livepeer/go-livepeer/verification	4.061s
```

verification/verify.go
mkrufky: In order to preserve the current behavior, we may need to keep it.  Either that, or something else may be wrong.

With it removed by the following patch:
```diff
diff --git a/verification/verify.go b/verification/verify.go
index 6a429060..d00aecb7 100644
--- a/verification/verify.go
+++ b/verification/verify.go
@@ -112,8 +112,6 @@ func (sv *SegmentVerifier) Verify(params *Params) (*Params, error) {
        //      If not, exit early. Seed sample using source data for repeatability!
        if sv.policy.Verifier != nil {
                res, err = sv.policy.Verifier.Verify(params)
-       } else if params.Orchestrator == nil {
-               return nil, nil
        }
 
        pxls := res.Pixels
```
a test fails.  Should we change the test or leave it as-is?  ...or is there something else we should be doing instead?  Here is the failing test:
```
$ go test
E0211 16:08:33.523832   32603 epic.go:151] Could not submit request Post : unsupported protocol scheme ""
--- FAIL: TestVerify (1.34s)
    verify_test.go:40: 
        	Error Trace:	verify_test.go:40
        	Error:      	Expected nil, but got: &verification.Params{ManifestID:"", Source:(*stream.HLSSegment)(nil), Profiles:[]ffmpeg.VideoProfile(nil), Orchestrator:(*net.OrchestratorInfo)(nil), Results:(*net.TranscodeData)(nil), URIs:[]string(nil), Renditions:[][]uint8(nil)}
        	Test:       	TestVerify
demuxer: Unable to open file
Freeing input based on OPEN INPUT error
E0211 16:08:34.866358   32603 ffmpeg.go:292] Transcoder Return : No such file or directory
demuxer: Unable to open file
Freeing input based on OPEN INPUT error
E0211 16:08:35.328943   32603 ffmpeg.go:292] Transcoder Return : No such file or directory
FAIL
exit status 1
FAIL	github.com/livepeer/go-livepeer/verification	4.118s
```

verification/verify.go
mkrufky: done in 078988d7c8a689eb1da0a5806212bcd61473db20

cmd/livepeer/livepeer.go
mkrufky: We already check `*verifierUrl != ""` ~15 lines above that.  I'm not sure what we gain from this.

...or will that be clearer after #1368 is merged?

verification/verify.go
yondonfu: Looks like these tests were originally written to handle a scenario where the verifier does not return pixel counts so they are checking for `ErrPixelsAbsent`. Since that is no longer expected behavior as of this PR we can adjust the tests - a test that expected `ErrPixelsAbsent` should no longer expect an error since pixel counting is run within `SegmentVerifier.Verify()`.

cmd/livepeer/livepeer.go
yondonfu: I'm suggesting to move the `*verifierUrl` check. If you don't nest the `*verifierUrl` check within the on-chain check then when a user specifies a value for the `-verifierUrl` flag *and* runs the node in on-chain mode then the node will hit a problem when it tries to verify a non-existent signature when working with an off-chain O. Furthermore, verification policies should only be necessary when the node is in on-chain mode.

verification/verify.go
yondonfu: Hm so originally the expected behavior was for a segment verifier to exit early using a policy with a nil verifier. In this PR, a segment verifier using a policy with a nil verifier is expected to count pixels and compare with the reported pixels. So, one could argue that it is reasonable for the segment verifier to return the provided `Params` since there are no rendition segments and no reported pixels so pixel count verification automatically passes. In that case, the test would just need to be adjusted to check for that expected behavior. This feels reasonable to me.

verification/verify_test.go
yondonfu: Mainly for commit history clarity/ease of understanding i.e. introducing a change in the commit where it is actually relevant. Could adjust at the very end, but seems easier to do it now.

cmd/livepeer/livepeer.go
j0sh: > Furthermore, verification policies should only be necessary when the node is in on-chain mode.

At least for me, all the verification development happened off-chain. Since we're incorporating the sig check in the other PR, I can see how that could make things a bit complicated: an offchain verifier (say, for development / testing) would run into sig checking errors without further onchain checks within the verifier. But I'm not sure how common that use case will actually be going forward. So simply doing the check as @yondonfu suggested seems OK. Otherwise if we want to preserve the ability to specify the Epic verifier offchain, then the current structure should probably be preserved, such as if there might ever be a desire to isolate verification testing from more general onchain testing.

I'm indifferent though, since I don't really have further plans to work on this area of the code.

verification/verify_test.go
mkrufky: bd4ecfdd3caf31de7c335f883e50ab659646469b

verification/verify.go
mkrufky: What about the tests that expect "mno" but actually get "pqr" or "stu" ?

Also, the comments above the tests that expected `ErrPixelsAbsent` indicate that errors are indeed expected.  So now they expect `ErrPixelMismatch` instead:

```diff
diff --git a/verification/verify.go b/verification/verify.go
index 8809badb..d0b87e7a 100644
--- a/verification/verify.go
+++ b/verification/verify.go
@@ -166,10 +166,6 @@ func IsRetryable(err error) bool {
 
 func (sv *SegmentVerifier) countPixelParams(params *Params) ([]int64, error) {
 
-       if params.Orchestrator == nil {
-               return nil, ErrPixelsAbsent
-       }
-
        if len(params.Results.Segments) != len(params.Renditions) {
                return nil, ErrPixelMismatch
        }
diff --git a/verification/verify_test.go b/verification/verify_test.go
index e23f0c83..c6ad7b00 100644
--- a/verification/verify_test.go
+++ b/verification/verify_test.go
@@ -59,7 +59,7 @@ func TestVerify(t *testing.T) {
        }}
        res, err = sv.Verify(&Params{Results: data})
        assert.Nil(res)
-       assert.Equal(ErrPixelsAbsent, err)
+       assert.Equal(ErrPixelMismatch, err)
 
        // check pixel count fails w/ external verifier
        data.Segments = append(data.Segments, &net.TranscodedSegmentData{Url: "def", Pixels: verifier.results.Pixels[1]})
@@ -135,12 +135,12 @@ func TestVerify(t *testing.T) {
        // Good score but incorrect pixel list should still fail
        verifier.results = &Results{Score: 10.0, Pixels: []int64{789}}
        res, err = sv.Verify(&Params{ManifestID: "pqr", Results: data})
-       assert.Equal(ErrPixelsAbsent, err)
+       assert.Equal(ErrPixelMismatch, err)
        assert.Equal("mno", string(res.ManifestID)) // Still return best result
        // Higher score and incorrect pixel list should prioritize pixel count
        verifier.results = &Results{Score: 20.0, Pixels: []int64{789}}
        res, err = sv.Verify(&Params{ManifestID: "stu", Results: data})
-       assert.Equal(ErrPixelsAbsent, err)
+       assert.Equal(ErrPixelMismatch, err)
        assert.Equal("mno", string(res.ManifestID)) // Still return best result
 
        // Check *not* retryable; should never get a result
```
now yeilds:
```
E0212 06:24:39.765784   11284 epic.go:151] Could not submit request Post : unsupported protocol scheme ""
--- FAIL: TestVerify (1.47s)
    verify_test.go:139: 
        	Error Trace:	verify_test.go:139
        	Error:      	Not equal: 
        	            	expected: "mno"
        	            	actual  : "pqr"
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-mno
        	            	+pqr
        	Test:       	TestVerify
    verify_test.go:144: 
        	Error Trace:	verify_test.go:144
        	Error:      	Not equal: 
        	            	expected: "mno"
        	            	actual  : "stu"
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-mno
        	            	+stu
        	Test:       	TestVerify
demuxer: Unable to open file
Freeing input based on OPEN INPUT error
E0212 06:24:41.236904   11284 ffmpeg.go:292] Transcoder Return : No such file or directory
demuxer: Unable to open file
Freeing input based on OPEN INPUT error
E0212 06:24:41.701484   11284 ffmpeg.go:292] Transcoder Return : No such file or directory
FAIL
exit status 1
FAIL	github.com/livepeer/go-livepeer/verification	4.329s
```

verification/verify.go
mkrufky: done in 1fd94970ddf2bdbe683489e33ba8259a7ef3dc42

cmd/livepeer/livepeer.go
mkrufky: got it.  following yondon's suggestion here https://github.com/livepeer/go-livepeer/pull/1368/commits/e42d610844b20a2771b28c0306d1d076870de63f in #1368

mkrufky: I think this suggested change might be causing the hang in circle CI from PR #1368 - https://circleci.com/gh/livepeer/go-livepeer/4602?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

verification/verify.go
yondonfu: Looks like those tests were checking that if the length of the slice of computed pixels counts is N and the length of the slice of reported pixel counts is M that the segment verifier will *not* return the `Params` object passed in despite it having a higher score than the results already tracked by the segment verifier. However, the segment verifier is returning the `Params` object passed in right now in those tests. The bug is here:

```
if len(pxls) != len(params.Results.Segments) {
    pxls, err = sv.countPixelParams(params)
}
```

In these tests, the verifier returns a slice of computed pixel counts that has an incorrect length. We previously set the error to `ErrPixelsAbsent` in this case, but now we're creating a fresh slice of computed pixel counts. Instead of counting the pixels when `len(pxls) != len(params.Results.Segments)` we can count the pixels when `len(pxls) == 0` and then after that if `len(pxls) != len(params.Results.Segments)` we set the error to `ErrPixelMismatch`.

cmd/livepeer/livepeer.go
yondonfu: > I think this suggested change might be causing the hang in circle CI from PR

Looks like the failure is due to a flaky test. Tracking in #1380 

> Otherwise if we want to preserve the ability to specify the Epic verifier offchain, then the current structure should probably be preserved, such as if there might ever be a desire to isolate verification testing from more general onchain testing.

Ah yeah good point. Ra has mainly been doing verification testing in off-chain mode since it is much easier to setup so I think its a good idea to still allow the Epic verifier to be used in off-chain mode for now. I'm ok keeping this code as-is then and then adding an on-chain check prior to the sig verification in #1368

verification/verify.go
```diff
@@ -100,23 +105,28 @@ func (sv *SegmentVerifier) Verify(params *Params) (*Params, error) {
...
+	res := &Results{}
+
 	// TODO Use policy sampling rate to determine whether to invoke verifier.
 	//      If not, exit early. Seed sample using source data for repeatability!
-	res, err := sv.policy.Verifier.Verify(params)
+	if sv.policy.Verifier != nil {
+		res, err = sv.policy.Verifier.Verify(params)
+	}
+
+	var pxls []int64
```
yondonfu: Doesn't look like this var is used outside of the scope of the if block starting at L123 so maybe we can define it in there which also has the added benefit of avoiding an additional nil check for `res` since the if conditional at L123 already does that check.

verification/verify.go
```diff
@@ -153,3 +163,68 @@ func IsRetryable(err error) bool {
 	_, retryable := err.(Retryable)
 	return retryable
 }
+
+func (sv *SegmentVerifier) countPixelParams(params *Params) ([]int64, error) {
```
yondonfu: Does this need to be a method on `SegmentVerifier`? Doesn't look like it relies on any `SegmentVerifier` state.

verification/verify_test.go
```diff
@@ -57,19 +61,42 @@ func TestVerify(t *testing.T) {
...
 	assert.NotNil(res)
 
+	// check pixel count fails w/o external verifier
+	sv = NewSegmentVerifier(&Policy{Verifier: nil, Retries: 2}) // reset
+	data = &net.TranscodeData{Segments: []*net.TranscodedSegmentData{
+		{Url: "abc", Pixels: verifier.results.Pixels[0] + 1},
+	}}
+	data.Segments = append(data.Segments, &net.TranscodedSegmentData{Url: "def", Pixels: verifier.results.Pixels[1]})
+	assert.Len(data.Segments, len(verifier.results.Pixels)) // sanity check
+	res, err = sv.Verify(&Params{Results: data, Orchestrator: &net.OrchestratorInfo{TicketParams: &net.TicketParams{}}})
```
yondonfu: Looks like there is still an extra `TicketParams` here.

yondonfu (changes requested):

verification/verify.go
mkrufky: Thanks for explaining that.  It makes good sense, but when I apply the following diff I am getting the same exact failure in the tests:
```diff
diff --git a/verification/verify.go b/verification/verify.go
index 8809badb..c630ce81 100644
--- a/verification/verify.go
+++ b/verification/verify.go
@@ -121,9 +121,12 @@ func (sv *SegmentVerifier) Verify(params *Params) (*Params, error) {
 
        // Check pixel counts
        if (err == nil || IsRetryable(err)) && res != nil && params.Results != nil {
-               if len(pxls) != len(params.Results.Segments) {
+               if len(pxls) == 0 {
                        pxls, err = sv.countPixelParams(params)
                }
+               if len(pxls) != len(params.Results.Segments) {
+                       err = ErrPixelMismatch
+               }
                for i := 0; err == nil && i < len(params.Results.Segments) && i < len(pxls); i++ {
                        reportedPixels := params.Results.Segments[i].Pixels
                        verifiedPixels := pxls[i]
@@ -166,10 +169,6 @@ func IsRetryable(err error) bool {
 
 func (sv *SegmentVerifier) countPixelParams(params *Params) ([]int64, error) {
 
-       if params.Orchestrator == nil {
-               return nil, ErrPixelsAbsent
-       }
-
        if len(params.Results.Segments) != len(params.Renditions) {
                return nil, ErrPixelMismatch
        }
diff --git a/verification/verify_test.go b/verification/verify_test.go
index e23f0c83..c6ad7b00 100644
--- a/verification/verify_test.go
+++ b/verification/verify_test.go
@@ -59,7 +59,7 @@ func TestVerify(t *testing.T) {
        }}
        res, err = sv.Verify(&Params{Results: data})
        assert.Nil(res)
-       assert.Equal(ErrPixelsAbsent, err)
+       assert.Equal(ErrPixelMismatch, err)
 
        // check pixel count fails w/ external verifier
        data.Segments = append(data.Segments, &net.TranscodedSegmentData{Url: "def", Pixels: verifier.results.Pixels[1]})
@@ -135,12 +135,12 @@ func TestVerify(t *testing.T) {
        // Good score but incorrect pixel list should still fail
        verifier.results = &Results{Score: 10.0, Pixels: []int64{789}}
        res, err = sv.Verify(&Params{ManifestID: "pqr", Results: data})
-       assert.Equal(ErrPixelsAbsent, err)
+       assert.Equal(ErrPixelMismatch, err)
        assert.Equal("mno", string(res.ManifestID)) // Still return best result
        // Higher score and incorrect pixel list should prioritize pixel count
        verifier.results = &Results{Score: 20.0, Pixels: []int64{789}}
        res, err = sv.Verify(&Params{ManifestID: "stu", Results: data})
-       assert.Equal(ErrPixelsAbsent, err)
+       assert.Equal(ErrPixelMismatch, err)
        assert.Equal("mno", string(res.ManifestID)) // Still return best result
 
        // Check *not* retryable; should never get a result
```

verification/verify.go
mkrufky: fixed in 9bd866f73260f75ba6d7e40aa1d5278ee00ffb6a

verification/verify.go
mkrufky: it did, but it doesn't any more.  fixed in 06542f7887809487b6899749ab3db683dff211cd

verification/verify_test.go
mkrufky: fixed in 6c66a847fbebaeb3eb0c0f5bb1aafd170d31bf4e

verification/verify.go
yondonfu: Ah now the problem is that `ErrPixelMismatch` is a `Retryable` error so this code:

```
if IsRetryable(err) {
    // Append results to sv.results
}
```

will execute and the `Params` passed into the segment verifier will end up being returned because it has the highest score despite having a reported pixels slice with an incorrect length. We could set the error to `ErrPixelsAbsent` when `len(pxls) != len(params.Results.Segments)` for now since it is not `Retryable` and once we implement #1327 we can return `ErrPixelMismatch` since `Params` that trigger a fatal retryable error such as `ErrPixelMismatch` will not be eligible to be returned.

mkrufky: I believe this is resolved by e05a63d37931c4f2c795e5debe0cf71a55fae283

verification/verify_test.go
```diff
@@ -57,19 +61,42 @@ func TestVerify(t *testing.T) {
...
+	// check pixel count fails w/o external verifier
+	sv = NewSegmentVerifier(&Policy{Verifier: nil, Retries: 2}) // reset
+	data = &net.TranscodeData{Segments: []*net.TranscodedSegmentData{
+		{Url: "abc", Pixels: verifier.results.Pixels[0] + 1},
+	}}
+	data.Segments = append(data.Segments, &net.TranscodedSegmentData{Url: "def", Pixels: verifier.results.Pixels[1]})
+	assert.Len(data.Segments, len(verifier.results.Pixels)) // sanity check
+	res, err = sv.Verify(&Params{Results: data, Orchestrator: &net.OrchestratorInfo{}})
+	assert.Nil(res)
+	assert.Equal(ErrPixelsAbsent, err)
```
yondonfu, line 88: Can we also add a test checking for `ErrPixelsMismatch` in the case where `Params.Renditions` is populated and the slice is the same length as the length of `Segments` except the actual pixel counts are incorrect?

yondonfu (changes requested):
Just one comment about adding an additional test when a verifier is not used with the segment verifier.

verification/verify_test.go
mkrufky: added in 381cf5938ce6f66d5a2af293a461282c1d0bff49

yondonfu (approved):
LGTM - let's clean up the commit history before merging.

PR #1364, opened 2020-02-07
transcoder: Use a per-GPU stack for nvidia transcodes. …

j0sh:

transcoder: Use a per-GPU stack for nvidia transcodes. …
This has the effect of serializing transcode jobs, but is generally
20-60% faster depending on the hardware and configuration used.

We use a LIFO stack to hold jobs for the transcoder to minimize latency of newer segments, at the expense of much higher tail latencies. This seems somewhat OK given that: 
* The broadcaster will time out and retry anyway
* Encourages broadcasters to stay with fast transcoders.

Probably needs better buffer management to drop segments that are *too* old, otherwise they could accumulate indefinitely with busy transcoders.

- Initializes each GPU with a stack and a transcode loop popping from the stack
- Moves transcoding logic into the transcode loop
- Rewrite the `Transcode` function to push to the stack, and to receive results
- Add census metrics tracking the GPU backlog
- Linter fixes
- Adds more tests. The tests are run under `-race` to check thread safety. Mostly unused unless the `NV_DEVICE`  env var is set.

Added unit tests. Needs the `NV_DEVICE` env var set to execute nvidia tests, so these won't execute right now with our current CI setup.

WIP - LPMS benchmark report for a more apples-to-apples ffmpeg comparison. 

Brief before / after analysis of logs shows shows that this method is both faster overall and processes more segments.

Test Setup:
* 24 streams of 1080p30fps BBB
* 4 output renditions 720p,576p,360p,240p - 96 concurrent encodes
* B and OT on different machines.

8 OT, 1 GPU each (all same machine):
* Success Rate, Buffered LIFO - 73.5 %
* Success Rate, Unbuffered - 18.0 %
* Success Rate, Buffered FIFO - 1.2% (not a typo)

1 OT, 8 GPUs:
* Success Rate, Buffered LIFO - 80.0 %
* Success Rate, Unbuffered - 76.2 %
* Success Rate, Buffered FIFO - Untested

image

**Checklist:**

### DIFF
--- a/core/transcoder.go
+++ b/core/transcoder.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/livepeer/go-livepeer/common"
@@ -55,40 +56,113 @@ func NewLocalTranscoder(workDir string) Transcoder {
 	return &LocalTranscoder{workDir: workDir}
 }
 
+type nvSegResult struct {
+	*TranscodeData
+	error
+}
+
+type nvSegData struct {
+	session  *ffmpeg.Transcoder
+	fname    string
+	profiles []ffmpeg.VideoProfile
+	res      chan *nvSegResult
+}
+
 type NvidiaTranscoder struct {
-	workDir string
-	device  string
+	device  *segStack
 	session *ffmpeg.Transcoder
 }
 
+type segStack struct {
+	segs []*nvSegData
+	mu   *sync.Mutex
+	cv   *sync.Cond
+	gpu  string
+}
+
+func newSegStack(gpu string) *segStack {
+	mu := &sync.Mutex{}
+	return &segStack{
+		segs: []*nvSegData{},
+		mu:   mu,
+		cv:   sync.NewCond(mu),
+		gpu:  gpu,
+	}
+}
+
+func (ss *segStack) push(seg *nvSegData) {
+	ss.mu.Lock()
+	defer ss.mu.Unlock()
+	ss.segs = append(ss.segs, seg)
+	ss.cv.Broadcast()
+}
+
+func (ss *segStack) pop() *nvSegData {
+	ss.mu.Lock()
+	defer ss.mu.Unlock()
+	for len(ss.segs) <= 0 {
+		ss.cv.Wait()
+	}
+	seg := ss.segs[len(ss.segs)-1]
+	ss.segs = ss.segs[:len(ss.segs)-1]
+	return seg
+}
+
 func (nv *NvidiaTranscoder) Transcode(job string, fname string, profiles []ffmpeg.VideoProfile) (*TranscodeData, error) {
-	// Set up in / out config
-	in := &ffmpeg.TranscodeOptionsIn{
-		Fname:  fname,
-		Accel:  ffmpeg.Nvidia,
-		Device: nv.device,
-	}
-	opts := profilesToTranscodeOptions(nv.workDir, ffmpeg.Nvidia, profiles)
 
-	// Do the Transcoding
-	res, err := nv.session.Transcode(in, opts)
-	if err != nil {
-		return nil, err
+	segData := &nvSegData{
+		session:  nv.session,
+		fname:    fname,
+		profiles: profiles,
+		res:      make(chan *nvSegResult, 1),
 	}
+	nv.device.push(segData)
+	res := <-segData.res
+	return res.TranscodeData, res.error
+}
 
-	return resToTranscodeData(res, opts)
+var nvidiaDevices map[string]*segStack
+
+func StartNvidiaTranscoders(devices string, workDir string) {
+	nvidiaDevices = make(map[string]*segStack)
+	d := strings.Split(devices, ",")
+	for _, n := range d {
+		st := newSegStack(n)
+		nvidiaDevices[n] = st
+		go runTranscodeLoop(st, workDir)
+	}
 }
 
 func NewNvidiaTranscoder(gpu string, workDir string) TranscoderSession {
 	return &NvidiaTranscoder{
-		workDir: workDir,
-		device:  gpu,
+		device:  nvidiaDevices[gpu],
 		session: ffmpeg.NewTranscoder(),
 	}
 }
 
 func (nv *NvidiaTranscoder) Stop() {
 	nv.session.StopTranscoder()
+}
+
+func runTranscodeLoop(stack *segStack, workDir string) {
+	for {
+		seg := stack.pop()
+		// Set up in / out config
+		in := &ffmpeg.TranscodeOptionsIn{
+			Fname:  seg.fname,
+			Accel:  ffmpeg.Nvidia,
+			Device: stack.gpu,
+		}
+		opts := profilesToTranscodeOptions(workDir, ffmpeg.Nvidia, seg.profiles)
+		// Do the Transcoding
+		res, err := seg.session.Transcode(in, opts)
+		if err != nil {
+			seg.res <- &nvSegResult{nil, err}
+			continue
+		}
+		td, err := resToTranscodeData(res, opts)
+		seg.res <- &nvSegResult{td, err}
+	}
 }
 
 func parseURI(uri string) (string, uint64, error) {
--- a/core/transcoder.go
+++ b/core/transcoder.go
@@ -95,6 +95,9 @@ func (ss *segStack) push(seg *nvSegData) {
 	defer ss.mu.Unlock()
 	ss.segs = append(ss.segs, seg)
 	ss.cv.Broadcast()
+	if monitor.Enabled {
+		monitor.GPUBacklog(ss.gpu, len(ss.segs))
+	}
 }
 
 func (ss *segStack) pop() *nvSegData {
@@ -105,6 +108,9 @@ func (ss *segStack) pop() *nvSegData {
 	}
 	seg := ss.segs[len(ss.segs)-1]
 	ss.segs = ss.segs[:len(ss.segs)-1]
+	if monitor.Enabled {
+		monitor.GPUBacklog(ss.gpu, len(ss.segs))
+	}
 	return seg
 }
 
--- a/core/transcoder.go
+++ b/core/transcoder.go
@@ -133,7 +133,7 @@ func StartNvidiaTranscoders(devices string, workDir string) {
 	}
 }
 
-func NewNvidiaTranscoder(gpu string, workDir string) TranscoderSession {
+func NewNvidiaTranscoder(gpu string) TranscoderSession {
 	return &NvidiaTranscoder{
 		device:  nvidiaDevices[gpu],
 		session: ffmpeg.NewTranscoder(),
--- a/core/transcoder_test.go
+++ b/core/transcoder_test.go
@@ -87,7 +87,7 @@ func TestResToTranscodeData(t *testing.T) {
 	assert.EqualError(err, "lengths of results and options different")
 
 	// Test immediate read error
-	opts := []ffmpeg.TranscodeOptions{ffmpeg.TranscodeOptions{Oname: "badfile"}}
+	opts := []ffmpeg.TranscodeOptions{{Oname: "badfile"}}
 	_, err = resToTranscodeData(res, opts)
 	assert.EqualError(err, "open badfile: no such file or directory")
 
@@ -116,7 +116,7 @@ func TestResToTranscodeData(t *testing.T) {
 	res = &ffmpeg.TranscodeResults{Encoded: make([]ffmpeg.MediaInfo, 1)}
 	res.Encoded[0].Pixels = 100
 
-	opts = []ffmpeg.TranscodeOptions{ffmpeg.TranscodeOptions{Oname: file2.Name()}}
+	opts = []ffmpeg.TranscodeOptions{{Oname: file2.Name()}}
 	tData, err := resToTranscodeData(res, opts)
 	assert.Nil(err)
 	assert.Equal(1, len(tData.Segments))
@@ -207,7 +207,7 @@ func TestAudioCopy(t *testing.T) {
 	// LocalTranscoder API and checking that the result is identical.
 	audioSample := dir + "/audio-copy.ts"
 	in := &ffmpeg.TranscodeOptionsIn{Fname: "test.ts"}
-	out := []ffmpeg.TranscodeOptions{ffmpeg.TranscodeOptions{
+	out := []ffmpeg.TranscodeOptions{{
 		Oname:        audioSample,
 		VideoEncoder: ffmpeg.ComponentOptions{Name: "drop"},
 		AudioEncoder: ffmpeg.ComponentOptions{Name: "copy"},
--- a/core/transcoder_test.go
+++ b/core/transcoder_test.go
@@ -39,7 +39,7 @@ func TestNvidia_Transcoder(t *testing.T) {
 	tmp, _ := ioutil.TempDir("", "")
 	defer os.RemoveAll(tmp)
 	StartNvidiaTranscoders("123,456", tmp)
-	tc := NewNvidiaTranscoder("123", tmp)
+	tc := NewNvidiaTranscoder("123")
 	ffmpeg.InitFFmpeg()
 
 	// test.ts sample isn't in a supported pixel format, so use this instead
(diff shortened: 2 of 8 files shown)


iameli:
Do you think there's any sense in exposing the current count of segments sitting on the stack as a Prometheus metric? I'm imagining a case where we're slightly overloaded and we start to build up a backlog of queued segments; is that at all realistic?

j0sh:
> Do you think there's any sense in exposing the current count of segments sitting on the stack as a Prometheus metric

Might be useful if you think it's something we should monitor and alert on. Should this be per-GPU or combined for all GPUs on the machine? Per GPU seems sensible but not sure if that'd be too noisy.

j0sh:
Metrics added in https://github.com/livepeer/go-livepeer/pull/1364/commits/f4d3931879845da019fdad437c813a458cca2566

core/transcoder.go
```diff
@@ -91,6 +150,28 @@ func (nv *NvidiaTranscoder) Stop() {
...
+		opts := profilesToTranscodeOptions(workDir, ffmpeg.Nvidia, seg.profiles)
+		// Do the Transcoding
+		res, err := seg.session.Transcode(in, opts)
+		if err != nil {
+			seg.res <- &nvSegResult{nil, err}
+			continue
+		}
+		td, err := resToTranscodeData(res, opts)
+		seg.res <- &nvSegResult{td, err}
+		continue
```
yondonfu: Looks like this `continue` is unnecessary.

core/transcoder.go
```diff
@@ -55,34 +56,92 @@ func NewLocalTranscoder(workDir string) Transcoder {
...
+}
+
+var nvidiaDevices map[string]*segStack
+
+func StartNvidiaTranscoders(devices string, workDir string) {
+	nvidiaDevices = make(map[string]*segStack)
+	d := strings.Split(devices, ",")
+	for _, n := range d {
+		st := newSegStack()
+		st.gpu = n
```
yondonfu: Could `gpu` be set in `newSegStack()`?

core/transcoder.go
```diff
@@ -91,6 +150,28 @@ func (nv *NvidiaTranscoder) Stop() {
 	nv.session.StopTranscoder()
 }
 
+func runTranscodeLoop(gpu string, stack *segStack, workDir string) {
```
yondonfu: Can we use the `gpu` field in `segStack` and omit the extra `gpu` param here?

core/transcoder.go
```diff
@@ -91,6 +150,28 @@ func (nv *NvidiaTranscoder) Stop() {
...
+	for {
+		seg := stack.pop()
```
yondonfu, line 155: I don't think this loop will exit cleanly when the node stops while the stack is empty. One solution could be to send a cancellation signal via context. The top level context could be passed into `StartNvidiaTranscoders()` which passes it to `runTranscodeLoop()`.

core/transcoder_test.go
```diff
@@ -59,19 +59,86 @@ func TestNvidiaTranscoder(t *testing.T) {
...
+	go func() {
+		ss.pop()
+		wg.Done()
+	}()
+	assert.False(wgWait2(wg, 100*time.Millisecond), "Did not time out on empty stack")
+}
+
+func TestNvidia_ConcurrentStack(t *testing.T) {
+	// Run this under `-race`
+	//assert := assert.New(t)
```
yondonfu: Can remove this.

yondonfu (changes requested):

core/transcoder.go
j0sh: Yes, you're correct - the loop will never exit. This felt like a decent solution that got us 80% of the way with 20% of the effort, because there are only two times when the loop could be terminated: when a node shuts down gracefully, and when we run tests. Since we don't really attempt graceful termination elsewhere (and we have fallback / retry mechanisms for non-graceful termination), this felt OK.

There's a lot more incidental machinery if we bring in a cancellable context into `startTranscodeLoop`. In particular, it means "popping" from the stack actually means reading from a channel so we can `select` on the channel + context - so we still need to maintain the stack behavior internally, but write into a channel when a segment becomes available, rather than actually popping it. We'll also need some sort of feedback mechanism that the channel is available to be written to, since we don't want to queue/buffer segments since this is supposed to be a LIFO stack. All in all, the machinery adds up, and it feels a bit disproportionate to the cases it's meant to accommodate.

It's also more shuffling of data across channel boundaries - and we already send the data through a few channels between the load balancer, transcode loop, etc.

One alternative I've considered is to bring the cancellable context within the stack itself (and returning nil from the `pop` to break), but I haven't fully worked that out; probably by putting the queue in a `terminated` state once the context is signaled. If you still feel that we should cleanly shut down the loop regardless, then will look into it some more.

core/transcoder_test.go
j0sh: Fixed in 26c1f19b46bb9546bdbce22cbe3f890d957e0310

core/transcoder.go
j0sh: Fixed in 26c1f19b46bb9546bdbce22cbe3f890d957e0310

(note to self, this will cause a broken build for the fixed-up version of a15544b314103d1fc56cdcbfc836dd47a84514aa and probably a merge error. will fix when rebasing)

core/transcoder.go
j0sh: Fixed in 26c1f19b46bb9546bdbce22cbe3f890d957e0310

core/transcoder.go
j0sh: Fixed in 26c1f19b46bb9546bdbce22cbe3f890d957e0310

core/transcoder.go
j0sh: Something like this, but it's still non-graceful - any segments that are in the stack when termination is signaled will be dropped. Still needs testing etc. 

```diff
diff --git a/cmd/livepeer/livepeer.go b/cmd/livepeer/livepeer.go
index b1f74219..229bd121 100644
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -240,7 +240,7 @@ func main() {
 
        if *transcoder {
                if *nvidia != "" {
-                       core.StartNvidiaTranscoders(*nvidia, *datadir)
+                       core.StartNvidiaTranscoders(context.Background(), *nvidia, *datadir)
                        n.Transcoder = core.NewLoadBalancingTranscoder(*nvidia, core.NewNvidiaTranscoder)
                } else {
                        n.Transcoder = core.NewLocalTranscoder(*datadir)
diff --git a/core/transcoder.go b/core/transcoder.go
index 5bc891b4..64287f16 100644
--- a/core/transcoder.go
+++ b/core/transcoder.go
@@ -1,6 +1,7 @@
 package core
 
 import (
+       "context"
        "errors"
        "fmt"
        "io/ioutil"
@@ -78,16 +79,28 @@ type segStack struct {
        mu   *sync.Mutex
        cv   *sync.Cond
        gpu  string
+
+       terminated bool
 }
 
-func newSegStack(gpu string) *segStack {
+func newSegStack(ctx context.Context, gpu string) *segStack {
        mu := &sync.Mutex{}
-       return &segStack{
+       ss := &segStack{
                segs: []*nvSegData{},
                mu:   mu,
                cv:   sync.NewCond(mu),
                gpu:  gpu,
        }
+       go func() {
+               select {
+               case <-ctx.Done():
+                       mu.Lock()
+                       defer mu.Unlock()
+                       ss.terminated = true
+                       ss.cv.Broadcast()
+               }
+       }()
+       return ss
 }
 
 func (ss *segStack) push(seg *nvSegData) {
@@ -105,6 +118,9 @@ func (ss *segStack) pop() *nvSegData {
        defer ss.mu.Unlock()
        for len(ss.segs) <= 0 {
                ss.cv.Wait()
+               if ss.terminated {
+                       return nil
+               }
        }
        seg := ss.segs[len(ss.segs)-1]
        ss.segs = ss.segs[:len(ss.segs)-1]
@@ -129,11 +145,11 @@ func (nv *NvidiaTranscoder) Transcode(job string, fname string, profiles []ffmpe
 
 var nvidiaDevices map[string]*segStack
 
-func StartNvidiaTranscoders(devices string, workDir string) {
+func StartNvidiaTranscoders(ctx context.Context, devices string, workDir string) {
        nvidiaDevices = make(map[string]*segStack)
        d := strings.Split(devices, ",")
        for _, n := range d {
-               st := newSegStack(n)
+               st := newSegStack(ctx, n)
                nvidiaDevices[n] = st
                go runTranscodeLoop(st, workDir)
        }
@@ -153,6 +169,9 @@ func (nv *NvidiaTranscoder) Stop() {
 func runTranscodeLoop(stack *segStack, workDir string) {
        for {
                seg := stack.pop()
+               if nil == seg {
+                       return
+               }
                // Set up in / out config
                in := &ffmpeg.TranscodeOptionsIn{
                        Fname:  seg.fname,
```

yondonfu: Ah I see what you mean now.

> All in all, the machinery adds up, and it feels a bit disproportionate to the cases it's meant to accommodate.

While I do think its worthwhile to get in the habit of keeping graceful termination in mind,  I agree that the effort/additional complexity involved with the extra machinery required in this case feels greater than the benefit of ensuring graceful termination. I'm ok with the code as is.

I did think a bit about this a bit though. Maybe something like this could work:

```golang
type segStack struct {
    ...
    quit chan struct{}
}

// defer call this after calling StartNvidiaTranscoders()
func StopNvidiaTranscoders() {
    for _, device := range nvidiaDevices {
        close(device.quit)
        
        // If the stack is waiting, wake it up
        // Might need to do this in a for loop in case the the stack is not yet waiting
        // i.e. Code execution in the transcode loop is at stack.pop()
        device.cv.Signal()
    }
}

func runTranscodeLoop(stack *segStack, workDir string) {
    for {
        seg := stack.pop()

        select {
            case <-stack.quit:
                return
            default:
        }

        ...
    }
}
```

Feel free to look into the above if you'd like - I don't feel strongly one way or another.

j0sh:
```golang
func runTranscodeLoop(stack *segStack, workDir string) {
    for {
        seg := stack.pop()

        select {
            case <-stack.quit:
                return
            default:
        }
```

As-is, isn't that still non-graceful? We could be exiting from the loop while the stack still has elements. Doable if we check the stack size and skip the select if nonzero, just requires some care in testing.

I'm hoping that this whole serialization workaround is temporary anyway, and we can revert as soon as LPMS perf is better.

yondonfu: > As-is, isn't that still non-graceful? We could be exiting from the loop while the stack still has elements.

Ah didn't see your comment about dropping segments in the stack when termination is signaled when I posted my comment. I was thinking that dropping segments from the stack would be ok if termination is signaled by node shutdown.

> I'm hoping that this whole serialization workaround is temporary anyway, and we can revert as soon as LPMS perf is better.

SGTM 👍

core/transcoder.go
```diff
@@ -55,34 +56,92 @@ func NewLocalTranscoder(workDir string) Transcoder {
...
-	res, err := nv.session.Transcode(in, opts)
-	if err != nil {
-		return nil, err
+func (ss *segStack) push(seg *nvSegData) {
+	ss.mu.Lock()
+	defer ss.mu.Unlock()
+	ss.segs = append(ss.segs, seg)
+	ss.cv.Broadcast()
+	if monitor.Enabled {
+		monitor.GPUBacklog(ss.gpu, len(ss.segs))
```
yondonfu, line 99: Could there be a race condition here?

1. `pop()` locks `ss.mu`
2. `pop()` unlocks `ss.mu` in `ss.cv.Wait()` and pauses
3. `push()` locks `ss.mu` and appends to `ss.segs` s.t. the length is N + 1
4. `push()` invokes `ss.cv.Broadcast()`
5. `pop()` resumes from `ss.cv.Wait()` and locks `ss.mu`
6. `pop()` removes from `ss.segs` s.t. the length is N
7. `push()` calls `monitor.GPUBacklog()` with N instead of N + 1 

Should be resolved if invocation of `monitor.GPUBackLog()` is before `ss.cv.Broadcast()` in `push()`.

j0sh: Yeah, that was actually a bit intentional - wanted to signal the broadcaster ASAP. Not sure if the condvar will release the lock immediately, but I hope it does because there's quite a bit more that happens as the metrics are being recorded. I'd rather give the system an opportunity to execute a pop / transcode if it can (even though it may not always do so).

What you're describing is an example of the [ABA problem](https://en.wikipedia.org/wiki/ABA_problem), where intermediate values are not recorded. In this case I think it's OK; the length is still correct at the moment it's being inserted into the metrics system. At step 7, the backlog is indeed N due to the `pop` in steps 5/6.

Can you think of a case where this might be problematic? Only thing I can think of is if there was a really long delay in re-obtaining the lock after `ss.cv.Broadcast()` for some reason. In the worst case, sustained `push` calls may cause metric to fall behind by at most one segment per `push` - but these pushes can't really make progress either without also obtaining the lock. (And vice versa with recording stack reductions after a `pop`.)

yondonfu: Got it - that seems reasonable to me.

Could we add a comment mentioning that its possible for the metric recording might miss intermediate values just for clarification for a reader?

darkdarkdragon: Maybe I'm missing something here, but I don't see `Broadcast` function releasing lock internally - that means that step 6 will not execute until `push` returned.
@j0sh and if we want for it to work as you've described, we should not lock inside `push` function.

darkdarkdragon (approved):
I think we should add this `Probably needs better buffer management to drop segments that are too old, otherwise they could accumulate indefinitely with busy transcoders.` to separate issue, to not forget it.

LGTM

core/transcoder.go
yondonfu: Ah yeah I think @darkdarkdragon is right.

> Not sure if the condvar will release the lock immediately

When `Wait()` wakes up, it will lock the mutex before returning. But, the mutex is still locked at this point from step 3. So, `Wait()` will be paused until `push()` unlocks the mutex.

darkdarkdragon: and, btw, we could do
```golang
func (ss *segStack) push(seg *nvSegData) {
	ss.segs = append(ss.segs, seg)
        l := len(ss.segs)
	ss.cv.Broadcast()
	if monitor.Enabled {
		monitor.GPUBacklog(ss.gpu, l)
	}
}
```
this should fix it?

j0sh: Yeah, I should have just looked at the [source code](https://golang.org/src/sync/cond.go). So we actually need to hold the lock to protect all accesses to `ss.segs`, including during `len` . We could do something like:

```
lock
append seg
unlock
broadcast
lock
update metric with new length
unlock
```

Anyone have a preference for the above approach? I'm generally in favor of "less code" - keeping the current approach (but if shuffling code around gives us a small optimization win, I'll take it.) Too bad that's not the case here, but open to implementing it.

darkdarkdragon: Why not
```golang
func (ss *segStack) push(seg *nvSegData) {
	ss.mu.Lock()
        ss.segs = append(ss.segs, seg)
        l := len(ss.segs)
        ss.mu.Unlock()
	ss.cv.Broadcast()
	if monitor.Enabled {
		monitor.GPUBacklog(ss.gpu, l)
	}
}
```
?

j0sh: In that case we would be inserting the wrong length if `pop` was called before updating in the GPU backlog. We need to hold the lock during the metrics update.

darkdarkdragon: I thought it would just the behaviour you described in `Yeah, that was actually a bit intentional ` comment, no?

j0sh:
Rebased but left the code surrounding https://github.com/livepeer/go-livepeer/pull/1364#discussion_r382710201 and https://github.com/livepeer/go-livepeer/pull/1364#discussion_r382181531 as-is. Let me know if anyone feels otherwise.

@darkdarkdragon Added https://github.com/livepeer/go-livepeer/issues/1397 to track the issue of backlog management

core/transcoder.go
j0sh: > Yeah, that was actually a bit intentional

So that was referring to putting the metrics update *after* the broadcast, because I thought that might mean the broadcast would release the lock and maybe execute the transcode before updating the metric, then re-acquire the lock prior to returning from the `broadcast` call.

Since that's not actually the case, it doesn't really matter where the metrics update goes, so it's really just a matter of readability in this case. Do you think it's more readable if the metrics are updated before or after the broadcast? I can see how putting the metrics first might prevent some later confusion over the semantics of `broadcast`. But I'm also lazy enough that I'll leave this as-is unless someone asks for it :)

yondonfu (approved):
LGTM

PR #1368, opened 2020-02-10
Move sig verification into verification.SegmentVerifier

mk-livepeer:
This refactors the code related to the sig verification to exist entirely within the verification module. From now on, the sig verification will be invoked as part of the Verify() function.

see #1325 
depends on #1326


verification/verify.go
```diff
@@ -103,7 +112,29 @@ func (sv *SegmentVerifier) Verify(params *Params) (*Params, error) {
 		return nil, nil
 	}
 
-	// TODO sig checking; extract from broadcast.go
+	if params.Results != nil {
```
yondonfu: I think `params.Results` should always be non-nil if B is verifying results from O (otherwise we wouldn't have any results from O) so I don't think the nil check is needed here.

verification/verify.go
```diff
@@ -103,7 +112,29 @@ func (sv *SegmentVerifier) Verify(params *Params) (*Params, error) {
...
+		segHashes := make([][]byte, len(params.Results.Segments))
+		for i, segment := range params.Results.Segments {
+
+			data, err := drivers.GetSegmentData(segment.Url)
```
yondonfu: We'll already have the data in `params.Renditions` if B is using local memory storage so we could skip the download step in that case.

verification/verify.go
```diff
@@ -103,7 +112,29 @@ func (sv *SegmentVerifier) Verify(params *Params) (*Params, error) {
...
+			if err != nil {
+				glog.Errorf("%v error with segment: %v (URL: %v)", monitor.SegmentTranscodeErrorDownload, err, segment.Url)
+				break
+			}
+			segHashes[i] = crypto.Keccak256(data)
+		}
+
+		// Might not have seg hashes if results are directly uploaded to the broadcaster's OS
+		// TODO: Consider downloading the results to generate seg hashes if results are directly uploaded to the broadcaster's OS
```
yondonfu: If we are always downloading data using `drivers.GetSegmentData()` in the above for loop that means that we should always end up with seg hashes needed for sig verification so we can remove this TODO.

verification/verify.go
```diff
@@ -103,7 +112,29 @@ func (sv *SegmentVerifier) Verify(params *Params) (*Params, error) {
...
 	}
 
-	// TODO sig checking; extract from broadcast.go
+	if params.Results != nil {
+		segHashes := make([][]byte, len(params.Results.Segments))
+		for i, segment := range params.Results.Segments {
+
+			data, err := drivers.GetSegmentData(segment.Url)
+			if err != nil {
+				glog.Errorf("%v error with segment: %v (URL: %v)", monitor.SegmentTranscodeErrorDownload, err, segment.Url)
```
yondonfu: Could we just return early with the error here? This would mean that if we can't download the data in order to verify the signature then we return an error.

verification/verify.go
```diff
@@ -103,7 +112,29 @@ func (sv *SegmentVerifier) Verify(params *Params) (*Params, error) {
...
+				break
+			}
+			segHashes[i] = crypto.Keccak256(data)
+		}
+
+		// Might not have seg hashes if results are directly uploaded to the broadcaster's OS
+		// TODO: Consider downloading the results to generate seg hashes if results are directly uploaded to the broadcaster's OS
+		if len(segHashes) != len(params.Results.Segments) &&
```
yondonfu: Let's add tests for this new logic.

yondonfu: If we exit early with an error when we cannot download the segment data above then we don't need this check.

yondonfu (changes requested):

verification/verify.go
mkrufky: I think the `nil` check is indeed needed, or something else may be wrong.

After applying the following:
```diff
diff --git a/verification/verify.go b/verification/verify.go
index a05f39f7..29f872d4 100644
--- a/verification/verify.go
+++ b/verification/verify.go
@@ -112,7 +112,6 @@ func (sv *SegmentVerifier) Verify(params *Params) (*Params, error) {
                return nil, nil
        }
 
-       if params.Results != nil {
        segHashes := make([][]byte, len(params.Results.Segments))
        for i, segment := range params.Results.Segments {
 
@@ -134,7 +133,6 @@ func (sv *SegmentVerifier) Verify(params *Params) (*Params, error) {
                glog.Error("Sig check failed")
                return nil, errPMCheckFailed
        }
-       }
 
        var err error
        res := &Results{}
```
The tests blow up on a nil pointer dereference:
```
go-livepeer/verification$ go test
E0211 15:20:26.387266    1205 epic.go:151] Could not submit request Post : unsupported protocol scheme ""
--- FAIL: TestVerify (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0xcdf2b0]

goroutine 114 [running]:
testing.tRunner.func1(0xc00011a600)
	/home/mk/.gvm/gos/go1.13.7/src/testing/testing.go:874 +0x3a3
panic(0x161e660, 0x232ada0)
	/home/mk/.gvm/gos/go1.13.7/src/runtime/panic.go:679 +0x1b2
github.com/livepeer/go-livepeer/verification.(*SegmentVerifier).Verify(0xc000077e98, 0xc0001f8070, 0x0, 0x0, 0x0)
	/home/mk/git/livepeer/go-livepeer/verification/verify.go:115 +0x50
github.com/livepeer/go-livepeer/verification.TestVerify(0xc00011a600)
	/home/mk/git/livepeer/go-livepeer/verification/verify_test.go:39 +0x2ad
testing.tRunner(0xc00011a600, 0x17bcfe8)
	/home/mk/.gvm/gos/go1.13.7/src/testing/testing.go:909 +0xc9
created by testing.(*T).Run
	/home/mk/.gvm/gos/go1.13.7/src/testing/testing.go:960 +0x350
exit status 2
FAIL	github.com/livepeer/go-livepeer/verification	0.071s
```

verification/verify.go
mkrufky: thanks - done in 4def93a69fdbc03ac0623ab0ece10451070649a9

verification/verify.go
mkrufky: done in 2ba85e1c063328389bc03369ed34fb4adc11ae08

verification/verify.go
mkrufky: I agree.  However, we were able to remove this block of code in 4def93a69fdbc03ac0623ab0ece10451070649a9 , so this particular comment no longer applies.

verification/verify.go
mkrufky: Agreed - we don't need *this* check, but we do still need a check.   After reviewing 4def93a69fdbc03ac0623ab0ece10451070649a9 , I think you'll agree with me. I changed the check from `len(segHashes) != len(params.Results.Segments)` to `len(segHashes) != len(params.Renditions)`

mkrufky: done in 42fc12235cf02880eb2efde14c92fffde1f85174

verification/verify.go
yondonfu: Ah I see we have a few tests that don't include a `Results` in the `Params` passed in. Ok.

verification/verify.go
yondonfu: Not addressing this TODO in this PR is fine with me. This also gives as an opportunity to see if there is a more efficient way to handle the case where B needs to download the segment data twice from an external OS: once to generate a hash for sig verification and once as a part of the call to the ffmpeg lib that accepts a remote filename and will pull the data. Ideally, B would only download the data once. Tracking in #1382

j0sh: It seems like https://github.com/livepeer/go-livepeer/commit/2ba85e1c063328389bc03369ed34fb4adc11ae08 removed a different TODO ?

verification/verify.go
```diff
@@ -103,13 +110,29 @@ func (sv *SegmentVerifier) Verify(params *Params) (*Params, error) {
 		return nil, nil
 	}
 
-	// TODO sig checking; extract from broadcast.go
+	if params.Results != nil {
+		segHashes := make([][]byte, len(params.Results.Segments))
+		if len(segHashes) == len(params.Renditions) {
```
yondonfu: Do we need this if statement if we instead just iterate through `params.Renditions` instead of `params.Results.Segments`? Then, if `len(params.Renditions) != len(params.Results.Segments)` the comparison of `segHashes` and `params.Results.Segments` would fail below.

verification/verify.go
mkrufky: oops!  fixed in 8d66157d12d2b4dc0072908d2cf9bacd00266f4a and f558f4e24e1d2a44a0169d55773f5fc037ca2680

verification/verify_test.go
```diff
@@ -86,17 +86,29 @@ func TestVerify(t *testing.T) {
...
+	// Check pixel count fails when len(params.Results.Segments) != len(params.Renditions)
+	pxls, err = pixels("../server/test.flv")
+	assert.Nil(err)
+	data = &net.TranscodeData{Segments: []*net.TranscodedSegmentData{
+		{Url: "../server/test.flv", Pixels: pxls},
+		{Url: "../server/test.flv", Pixels: pxls},
+	}}
+	renditions = [][]byte{}
+	res, err = sv.Verify(&Params{Results: data, Orchestrator: &net.OrchestratorInfo{TicketParams: &net.TicketParams{}}, Renditions: renditions})
+	assert.NotNil(err)
```
j0sh: Should we check for a specific error here (I think it's `ErrPixelMismatch`)?

mkrufky: done in b4fcacc7dcc69df64ee555c01e0aceb78d4429e9

verification/verify_test.go
```diff
@@ -86,17 +86,29 @@ func TestVerify(t *testing.T) {
...
+		{Url: "../server/test.flv", Pixels: pxls},
+		{Url: "../server/test.flv", Pixels: pxls},
 	}}
 	renditions := [][]byte{{123}, {234}}
-	res, err = sv.Verify(&Params{Results: data, Orchestrator: &net.OrchestratorInfo{}, Renditions: renditions})
+	res, err = sv.Verify(&Params{Results: data, Orchestrator: &net.OrchestratorInfo{TicketParams: &net.TicketParams{}}, Renditions: renditions})
 	assert.Nil(err)
 	assert.NotNil(res)
 
+	// Check pixel count fails when len(params.Results.Segments) != len(params.Renditions)
```
yondonfu: We should have tests that check that sig verification is actually invoked. Structuring those tests is a bit tricky right now since we can't stub the `VerifySig()` implementation. Options include:

1. Generating some sample data and using it in the tests. Ex. Create 2 byte slices, compute the keccak256 hash of each, compute the keccak256 hash of the concatenation of the 2 hashes, sign this hash with an account and then use the signature + byte slices + account in this test
2. Configure the `Policy` with a `sigVerifier` interface that defines the `VerifySig` method and instantiate the `Policy` with an implementation that calls `crypto.VerifySig()`. See the [pm package](https://github.com/livepeer/go-livepeer/blob/75aa225b4667c57644f3289b760f656651e8407a/pm/sigverifier.go#L23) for an example.

verification/verify.go
mkrufky: We still need the `if` statement because after the loop we need to be able to read `params.Results.Sig` without a `nil` pointer dereference

mkrufky: Apologies, I was looking at the wrong `if` statement at first.  I do still believe that we need it, however.  When the following diff is applied:
```diff
diff --git a/verification/verify.go b/verification/verify.go
index 20c3f7ee..a0701cac 100644
--- a/verification/verify.go
+++ b/verification/verify.go
@@ -112,11 +112,9 @@ func (sv *SegmentVerifier) Verify(params *Params) (*Params, error) {
 
        if params.Results != nil {
                segHashes := make([][]byte, len(params.Results.Segments))
-               if len(segHashes) == len(params.Renditions) {
-                       for i := range params.Results.Segments {
+               for i := range params.Renditions {
                        segHashes[i] = crypto.Keccak256(params.Renditions[i])
                }
-               }
 
                // Might not have seg hashes if results are directly uploaded to the broadcaster's OS
                if len(segHashes) != len(params.Results.Segments) &&
```
we get the following runtime error in the line `segHashes[i] = crypto.Keccak256(params.Renditions[i])` on line 116:
```
--- FAIL: TestVerifier_Verify (0.00s)
panic: runtime error: index out of range [0] with length 0 [recovered]
	panic: runtime error: index out of range [0] with length 0

goroutine 788 [running]:
testing.tRunner.func1(0xc0006c6600)
	/home/mk/.gvm/gos/go1.13.7/src/testing/testing.go:874 +0x3a3
panic(0x19989c0, 0xc0005ae300)
	/home/mk/.gvm/gos/go1.13.7/src/runtime/panic.go:679 +0x1b2
github.com/livepeer/go-livepeer/verification.(*SegmentVerifier).Verify(0xc000076e80, 0xc00021a4d0, 0x1, 0xc0003be140, 0x1834100)
	/home/mk/git/livepeer/go-livepeer/verification/verify.go:116 +0x8ec
github.com/livepeer/go-livepeer/server.verify(0xc000076e80, 0xc00068fef8, 0xc0004b7180, 0xc0004bc2c0, 0xc0000d2320, 0xc0006a7560, 0x1, 0x1, 0x1ce1601, 0xc0006a77b0)
	/home/mk/git/livepeer/go-livepeer/server/broadcast.go:533 +0x1ea
github.com/livepeer/go-livepeer/server.TestVerifier_Verify(0xc0006c6600)
	/home/mk/git/livepeer/go-livepeer/server/broadcast_test.go:809 +0x7aa
testing.tRunner(0xc0006c6600, 0x1ab9cc0)
	/home/mk/.gvm/gos/go1.13.7/src/testing/testing.go:909 +0xc9
created by testing.(*T).Run
	/home/mk/.gvm/gos/go1.13.7/src/testing/testing.go:960 +0x350
exit status 2
FAIL	github.com/livepeer/go-livepeer/server	3.115s
```

yondonfu: I think the index out of range issue would be resolved if `segHashes` was initialized with a length of `len(params.Renditions)` rather than `len(params.Results.Segments)`.

verification/verify.go
yondonfu: Actually as-is the TODO:

```
TODO: Consider downloading the results to generate seg hashes if results are directly uploaded to the broadcaster's OS
```

is still relevant because we are not generating seg hashes when B is using an external OS in this PR. See [this comment](https://github.com/livepeer/go-livepeer/pull/1368#discussion_r378497489) for more details. The way the code was originally structured where we called `drivers.GetSegmentData()` for each rendition would have resolved the TODO, but it would have also introduced a double segment data download (as described in [this comment](https://github.com/livepeer/go-livepeer/pull/1368#discussion_r378497489)) when B is using an external OS so since the code has changed to *not* call `drivers.GetSegmentData()` for each rendition I figured it would make sense to address this TODO outside of this PR.

mkrufky: reverted - TODO restored in 87d0539812be8b5bae0fffdd5fec6476d3166988

verification/verify.go
mkrufky: doing so results in another range error, then i changed it to range over `params.Renditions` instead and that causes a nil pointer dereference.  Is there a reason why we can't just keep the check that `params.Results != nil` here?  We'll blow up on a nil pointer otherwise when we try to read `params.Results.Sig`

yondonfu: The nil pointer deference is caused by the call to `lpcrypto.VerifySig()` which references `params.Orchestrator.TicketParams.Recipient` and `params.Orchestrator` is nil in the test that triggers the nil pointer deference. As mentioned in [the pixel count PR](https://github.com/livepeer/go-livepeer/pull/1360/files#r378488954), we'll need an on-chain check before running sig verification now - adding this check should also resolve the nil pointer deference in the tests i.e. `params.Orchestrator != nil && params.Orchestrator.TicketParams != nil `. Ideally we wouldn't need that `params.Orchestrator != nil` check since it would be non-nil in practice, I think a few of the tests have it set to nil so including the check may be easier.

mkrufky: I believe this may be resolved by db98e762579c107cf87067e27c20c3e10bd68bcc and 0e7f71b18090d03153ba8e063db72595d3a0e8b1

verification/verify_test.go
mkrufky: hopefully this will suffice?  f60f3881ac9dc8f8a80a651ad88dc9eba156b740

cmd/livepeer/livepeer.go
```diff
@@ -651,6 +651,27 @@ func main() {
 			}
 
 			n.OrchestratorPool = dbOrchPoolCache
+
+			// Set up verifier
+			if *verifierURL != "" {
```
yondonfu: As mentioned in [this comment](https://github.com/livepeer/go-livepeer/pull/1360#discussion_r378488954) from #1360, there is value in allowing `-verifierURL` to be specified when the node is off-chain mode for testing/experimentation purposes. This change would prevent that - if we revert this change `-verifierURL` could be specified when the node is in on-chain mode.

verification/verify.go
```diff
@@ -98,12 +105,19 @@ func NewSegmentVerifier(p *Policy) *SegmentVerifier {
 }
 
 func (sv *SegmentVerifier) Verify(params *Params) (*Params, error) {
+	return sv.verify(verifySignature, params)
+}
+
+type sigVerifyFunc func(params *Params) error
```
yondonfu: Perhaps it could be easier for testing purposes if the signature of this function was:

```
type sigVerifyFunc func(addr ethcommon.Address, msg, sig []byte) bool
```

Then, the stub implementation of this type could just be configured to return true or false and we wouldn't need to deal with the test data setup that is currently necessary for generating a valid signature.

verification/verify_test.go
```diff
@@ -22,6 +25,10 @@ func (sv *stubVerifier) Verify(params *Params) (*Results, error) {
 	return sv.results, sv.err
 }
 
+func (sv *SegmentVerifier) VerifyNoSig(params *Params) (*Params, error) {
```
yondonfu: Hm feels a bit weird to me to modify the API of `SegmentVerifier` here even if it is only for testing. Another option could be to specify `sigVerifyFunc` as a field on `SegmentVerifier` that can use `lpcrypto.VerifySig()` by default in `NewSegmentVerifier()` and overridden with a stub implementation within the test. WDYT?

cmd/livepeer/livepeer.go
mkrufky: reverted in 99285721f354c0d3ccc32065e1bae94417086ccb

verification/verify.go
mkrufky: resolved by d3e4afb5f4f5f4a90b0982d7e18ad55a5af06db2

verification/verify_test.go
mkrufky: resolved by d3e4afb5f4f5f4a90b0982d7e18ad55a5af06db2

verification/verify_test.go
```diff
@@ -85,27 +89,98 @@ func TestVerify(t *testing.T) {
...
+	assert.Equal(errPMCheckFailed, err)
+	assert.Nil(res)
+
+	// Check sig verifier passes when len(params.Results.Segments) == len(params.Renditions) and sig is valid
+	// We use this stubVerifier to make sure that ffmpeg doesnt try to read from a file
+	sv = NewSegmentVerifier(&Policy{Verifier: &stubVerifier{
+		results: nil,
+		err:     nil,
+	}, Retries: 2})
+	testData0 := []byte{1, 2, 3, 4, 5, 6, 7, 8}
```
yondonfu: I don't think we need this test data setup now that we can stub the `verifySig` field in `SegmentVerifier` with an implementation that returns true/false.

verification/verify.go
```diff
@@ -160,6 +171,32 @@ func IsRetryable(err error) bool {
...
+	}
+
+	segHashes := make([][]byte, len(params.Renditions))
+	for i := range params.Renditions {
+		segHashes[i] = crypto.Keccak256(params.Renditions[i])
+	}
+
+	// Might not have seg hashes if results are directly uploaded to the broadcaster's OS
+	// TODO: Consider downloading the results to generate seg hashes if results are directly uploaded to the broadcaster's OS
+	if !sv.verifySig(
```
yondonfu, line 190: Just noting that this check will always fail when B is using an external OS since `Renditions` would be populated with a bunch of nil slices in that case. We can tackle that problem as a part of #1382.

yondonfu (changes requested):

verification/verify_test.go
mkrufky: We still need the test data setup in order for the code to generate a proper signature & pixel counts.   If if we don't need it, is there any reason to remove it?  Please keep in mind that unit tests also serve as documentation for developers.

verification/verify.go
mkrufky: ok

verification/verify_test.go
mkrufky: understood.

done.

yondonfu (approved):
LGTM - let's clean up the commit history before merging.

PR #1371, opened 2020-02-10
Move func VerifySig() to its own package => crypto.VerifySig()

mk-livepeer:
* Move the `VerifySig()` helper to its own `crypto` package
* Remove the DefaultSigVerifier struct from the pm package - the pm package doesn't need to be exposed to the details of how sig verification is implemented.

blocks #1325

### DIFF
--- a/core/orchestrator.go
+++ b/core/orchestrator.go
@@ -25,6 +25,7 @@ import (
 	"github.com/livepeer/go-livepeer/net"
 	"github.com/livepeer/go-livepeer/pm"
 
+	lpcrypto "github.com/livepeer/go-livepeer/crypto"
 	lpmon "github.com/livepeer/go-livepeer/monitor"
 	ffmpeg "github.com/livepeer/lpms/ffmpeg"
 	"github.com/livepeer/lpms/stream"
@@ -67,7 +68,7 @@ func (orch *orchestrator) VerifySig(addr ethcommon.Address, msg string, sig []by
 	if orch.node == nil || orch.node.Eth == nil {
 		return true
 	}
-	return pm.VerifySig(addr, crypto.Keccak256([]byte(msg)), sig)
+	return lpcrypto.VerifySig(addr, crypto.Keccak256([]byte(msg)), sig)
 }
 
 func (orch *orchestrator) Address() ethcommon.Address {
--- a/crypto/verify.go
+++ b/crypto/verify.go
@@ -1,0 +1,56 @@
+package crypto
+
+import (
+	"errors"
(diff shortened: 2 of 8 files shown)


pm/validator.go
```diff
@@ -31,12 +32,12 @@ type Validator interface {
 
 // validator is an implementation of the Validator interface
 type validator struct {
-	sigVerifier   SigVerifier
+	sigVerifier   lpcrypto.SigVerifier
```
yondonfu: I think we should keep the `SigVerifier` interface in the `pm` package so that the `pm` package defines the interface it requires for signature verification as opposed to relying on an interface described by a different package. The implementation provided by the `crypto` package can just adhere to this interface.

yondonfu (changes requested):

pm/validator.go
mkrufky: done in c813aa883b728eafcb616f266566e29505c8d832

yondonfu (commented):
Changes look good - let's squash the commits into one.

mk-livepeer:
@yondonfu please approve so that I can use the GitHub "squash and merge" button

yondonfu:
My understanding is that the GH "squash and merge" option enforces a particular commit message structure for multi-commit PRs that includes the titles for each of the commits. The team has been following a convention of rebasing to squash fixup commits which gives us better control over the descriptions in the commit history prior to approval + merge so let's stick with that process for this PR.

mk-livepeer:
GitHub's "squash and merge" starts with a suggested message that I always override.

If you allow me to squash first then you might miss the possibility that I snuck a bug or something else into the pr before merge.

If you prefer that I do this in the insecure way, causing an extra step in the flow, then say so.  Otherwise, please approve - I never push pre-filled commit messages

mk-livepeer:
I squashed and pushed.  This is insecure, like I said - you should re-review the code to make sure it's the same change that you reviewed previously.

j0sh:
> If you allow me to squash first then you might miss the possibility that I snuck a bug or something else into the pr before merge.

FWIW, this is still possible regardless. If someone approves, then the committer can force-push and merge arbitrary changes, regardless of which Github button is used. At least this way we can double check the supposedly-final commit structure before giving final approval.

ericxtang:
Thanks for following the current commit procedure for the sake of making progress @mk-livepeer.  I think we should all be open to change our process if we can agree on a better one.  If you want to bring this up in the next eng retro, we can talk about it as a team.

yondonfu (approved):
LGTM

PR #1383, opened 2020-02-14
multi: Remove acceptable payment error mechanism in favour of an OrchestratorInfo expiration mechanism

### LINKED ISSUE #1075: Unacceptable payment error when ticket params change for concurrent streams
**Describe the bug**

If a broadcaster sends > 3 concurrent streams to an orchestrator and the ticket params change (i.e. because of a change in the orchestrator's cached gas price, a winning ticket redemption, etc.), the broadcaster will send > 3 tickets with invalid params to the orchestrator which will cause the broadcaster's error count to exceed the current [maximum of 3](https://github.com/livepeer/go-livepeer/blob/master/cmd/livepeer/livepeer.go#L73). The 4th ticket received by the orchestrator before the broadcaster receives updated ticket params will trigger an unacceptable payment error and the orchestrator will not transcode the segment that the ticket came with.

**To Reproduce**
1. Create a network with 1 B and 1 O in on-chain mode
2. Send 4 streams into B
3. Send a few dummy transactions into the ETH node with a random gas price
4. Wait for O to cache the new gas price
5. Observe an unacceptable payment error in O's logs

**Expected behavior**

Ideally, an orchestrator would not skip transcoding a segment just because the broadcaster is sending multiple streams in and the ticket params changed.

### LINKED ISSUE #1343: TicketParams expiration time
# `TicketParams` Expiration

---
name: `TicketParams` expiration
about: Add an expiration block number to `TicketParams`
author: @kyriediculous (based on an [original idea](https://github.com/livepeer/go-livepeer/issues/1075#issuecomment-576753652) form @yondonfu )

---

## Abstract

`OrchestratorInfo` is the data structure that is communicated over the wire from an Orchestrator to a Broadcaster. It contains two fields important for payments to work: pricing information and ticket parameters for the _probabilistic micropayments_ protocol. 

```
// The orchestrator sends this in response to `GetOrchestrator`, containing
// miscellaneous data related to the job.
message OrchestratorInfo {

  // URI of the transcoder to use for submitting segments.
  string transcoder  = 1;

  // Parameters for probabilistic micropayment tickets
  TicketParams ticket_params = 2;

  // Price Info containing the price per pixel to transcode
  PriceInfo price_info = 3;

  // Orchestrator returns info about own input object storage, if it wants it to be used.
  repeated OSInfo storage = 32;
}
```
[1] [go_livepeer/net/lp_rpc.proto](https://github.com/livepeer/go-livepeer/blob/685a797b92d365772c7e06ab2a68ec6927a2f87c/net/lp_rpc.proto#L89)

Due to the client side implementation of the payment protocol, this information is subject to regular change: to avoid dealing with [gas price](https://ethgasstation.info/FAQpage.php) fluctuations on the Ethereum network for redeeming winning payments a transaction overhead is charged. 

Other reasons for change of `TicketParams` would be a winning ticket redemption on-chain or manual change in the node configuration.

When `OrchestratorInfo` changes mid-stream a Broadcaster is unaware that the old parameters will no longer be honoured by an Orchestrator resulting in a payment error and interruption of transcoding. The Orchestrator currently offers a grace period, but that could prove to be a game-able mechanic allowing a Broadcaster to send _any_ `TicketParams` while in the grace period. 

We can remove the current grace period implementation and instead introduce the concept of an **expiration time** on `TicketParams`. This would:

- Give protocol certainty to the Broadcaster that payments with non-expired parameters are honoured, and thus transcoding not interrupted, if an Orchestrator is not deviating from the protocol. 

- Gives an Orchestrator protocol certainty that it will always receive payments with valid parameters and thus redeemable tickets, as long as the payments parameters are not expired and the Broadcaster is not deviating from the protocol.

- Prevents transcoding workflow from being interrupted and running into payment errors if both nodes are behaving according to the protocol rules. 

## Motivation

The feature _should_ increase transcoding reliability as payment errors should no longer occur during a happy flow as well as removes the issue the number of concurrent streams from a broadcaster exceeding the maximum acceptable error count. The expiration scheme is an improvement over the current implementation of the grace period which introduces several issues of its own:

1. The current grace period is set at **3** acceptable payment errors. When a Broadcaster sends in more concurrent streams than this it will inevitably run over the error count when `OrchestratorInfo` changes even when neither party deviates from the protocol [2].

2. The broadcaster is unaware if a particular set of `OrchestratorInfo` will be honoured by the Orchestrator it is working with while still sending a payment along with it's transcode request. This is can be exploited by Orchestrators by setting large negative balances for _any_ broadcaster and high ticket winning probability, allowing to potentially redeem winning tickets without doing work. 

3. The grace period creates a new attack vector on the Orchestrator whereby a Broadcaster can get free transcoding: 
    - The grace period is tracked by ManifestID, which is picked by the Broadcaster. The Broadcaster could change this value every three segments and never run into the maximum error count. 
    - The Broadcaster can wait for a change in gas price, which resets the error count, to send new transcode requests with invalid payments.

Given this set of known issues there's two options:
1. Look for a better solution
2. Find ways to mitigate the risks so they are no longer statistically relevant

Rather than trying to limit the attack surface of the grace period mechanism which would involve adding more complexity to an already complex mechanism we can spend slightly more effort building a more robust solution. 

The expiration time mechanism is unaffected by the number of concurrent streams a Broadcaster operates, allowing it to immediately scale without additional implementation considerations. 

A simpler solution will also make the `pm` package in `go-livepeer` more maintainable and accessible to new engineers. 

The benefits mentioned above would outweigh the downside of the extra engineering effort created by stripping out the grace period logic completely as well as having to add a few additional bytes to transcoding responses sent over the wire witch each segment by having an Orchestrator attach it's current `TicketParams` with each response. 

[2] https://github.com/livepeer/go-livepeer/issues/1075

## Proposed Solution

From a high-level the proposed solution would look as follows:

A new field is added to the `OrchestratorInfo` message `ExpirationTime` this will contain a timestamp until which the current parameters will be valid. 

The Broadcaster can check whether the `OrchestratorInfo` contains a valid `ExpirationTime` during payment generation when creating its transcode request. At this point in the process we also validate the price and ticket parameters already in the current implementation of `go-livepeer`. In case `OrchestratorInfo` is past expiration when generating a payment, the Broadcaster will request the latest `OrchestratorInfo` from the Orchestrator. If no valid `OrchestratorInfo` is provided upon this refresh, the Orchestrator will be removed from the working set. 

The Orchestrator can keep track of previously advertised sets of `OrchestratorInfo` that have not expired and cross-check received tickets and their expiration time to see whether it should accept or reject an inbound transcoding request. It can do this by binding ticket parameters to the `recipientRand` (HMAC) generation. 

To allow the Broadcaster to always have access to a non-expired set of `OrchestratorInfo`, the data should be sent back by the Orchestrator in each response to a transcode request. 

#### Global clock

We'll need a logical clock to give participants in the system a global synchronised view of time. We can use the block numbers of blocks in the Ethereum network to indicate expiration time. 

If we use **block number** instead of the block timestamps we don't need to alter the current `eth/blockwatch` package (it only saves the block number) or create an additional service that queries a remote Ethereum node for the block timestamp when a new block is found. 

Since the timestamp is ambiguous with the block number (block timestamps are part of the block validation to make sure they are within certain bounds) there should be no difference. 

#### Extending `OrchestratorInfo`

We can simply extend the current `OrchestratorInfo` protobuf message with a new field `ExpirationBlock` with an ascending index. Broadcaster nodes that don't upgrade their node software will ignore the field. 

```
message OrchestratorInfo {
...
bytes expiration_block = 4;

}
````

Analogously we can adjust `pm.TicketParams` to include both `ExpirationBlock *big.Int` and `PricePerPixel *big.Rat` so we have access to these values inside the `sender` and `recipient`. 

#### Default Expiration Time

A set of `TicketParams` T would have to remain valid until the next advertised set of `TicketParams` T+1 is able to be received by the Broadcaster so that the Broadcaster is given ample time to update without disrupting transcoding due to minor latency discrepancies. 

Since we'll be using block numbers the expiration time will be at least the _block time_  of the Ethereum network (current ~12 seconds). 

Given that an Orchestrator needs to respond within 8 seconds anyway that would put the maximum possible time required for a Broadcaster to receive the updated `TicketParams` at `blockTime + 8 seconds`. 

Using block numbers avoids fluctuations in block time and the expiration time would have to be **2 blocks**. As a result `TicketParams` generated at block height `N` will be valid up until (but not included) block height `N+2`.

#### `TicketParams` validation

An orchestrator needs to make sure that when receiving a payment it is with parameters that it has previously advertised. 

To do this we can incorporate the ticket parameters when generating the recipientRand (as an [HMAC](https://en.wikipedia.org/wiki/HMAC)). 

```
recipientRand = HMAC_SHA256(secret || seed || sender || faceValue || winProb || expirationBlock || price)
```

By binding the ticket params to the HMAC we can check that `recipientRandHash` from a received ticket is the same as one previously advertised: the Orchestrator reconstructs the `recipientRand` from the parameters provided with the ticket and checks it against the `recipientRandHash` on the ticket. Since the HMAC is generated with a secret only the Orchestrator knows, it is impossible for a Broadcaster to forge ticket parameters. 

### Reworking `RoundsWatcher` into a more generic `TimeWatcher`

The last seen block numbers needs to be read in three different places: 
- For the `sender` when generating a payment (this is part of the transcoding workflow).
- For the `recipient` when generating ticket parameters and when validating received tickets.
- For the `ticketQueue` to check whether to send tickets created from expired `OrchestratorInfo` can be redeemed.

The easiest path to achieve this is to extend the functionalities of  the `roundsWatcher`, which currently watches for the latest round number and block hash of the block it was initialised in, to keep track of the last seen block number as well. Since the `roundsWatcher` already subscribes to the `blockwatcher` this can be as simple as keeping track of a single additional state variable. 

Alternatively we could use the DB to query this information since it is already stored there, but this would require additional dependency injection as well as an alternative polling mechanism for the ticketQueue rather than relying on a channel (see **Delay Ticket Redemption until `TicketParams` Expiration**) 

#### Delay Ticket Redemption Until `TicketParams` Expiration
Currently winning tickets are redeemed on-chain instantly when possible upon receiving them, revealing the current `recipientRand` value in the _commit-reveal_ scheme. 

This opens up a new attack vector in the case of the `TicketParams` not yet being expired. The Orchestrator would still honour the revealed `recipientRand` but it would be possible for a Broadcaster to grind out `ticket.senderNonces` that would result in signatures over the ticket that result in non-winning tickets only. 

`KECCAK256(sig, recipientRand) < winProb`

Within the current version of the payment protocol which uses [ECDSA signatures](https://gist.github.com/yondonfu/352ae689b0d802cc6e3f5403739b4e6a) and a commit-reveal scheme the easiest solution to implement right now would be to delay winning ticket validation until after expiration of the `TicketParams` that constitute the ticket. 

We could leverage the existing `ticketQueue` for this. Caveat here is that tickets would be stored and retried on a per-sender basis despite ticket expiration being global across senders. This should be ok however.

`ExpirationBlock` can be added to `pm.TicketParams` as well as `pm.Ticket` (when using `pm.NewTicket()` and `batch.Ticket()`), giving access to a ticket's expiration block in the `ticketQueue`.

Instead of trying to redeem tickets in the queue when the `maxFloat` for a sender changes, we can run the loop upon discovery of a new block. To achieve this we can create a subscription on `TimeWatcher` that allows us to pass in a sink channel (`chan<- *big.Int`) on which to receive updates. 

Having tickets not being redeemed immediately could lead to issues at the end of rounds where the Orchestrator will be removed from the active set in the coming round. Given however that active set churn is very low momentarily and there's other end-of round improvements that can be made for the payment scheme it seems reasonable to set that concern aside for the time being. 

The longer term solution anyway would be to remove interactivity from the winning ticket selection protocol by moving from deterministic signatures to a non-interactive protocol such as Verifiable Random Functions. The unknown is whether the cost of a VRF implementation to be executed on the EVM is warranted.  

### Refresh TicketParams when failing over 

It is possible that whenever the Broadcaster needs to fail over and select a new Orchestrator that the `OrchestratorInfo` for the new Orchestrator has expired if we haven't refreshed the session recently. 

Consider the following scenario:

1. B has orchestrators O1, O2 in its active set
2. O1 is performing well, so B continues using it
3. O2 ticket expires
4. O1 has a quick, transient issue that causes transcoding to take longer than realtime for one particular segment.
5. B tries to fall back to O2. Ticket has expired, so B drops O2 from the working set. No other orchestrators available, so B also drops the segment.

The quick solution here would be to just retry submitting the segment once and directly querying the selected Orchestrator for updated `OrchestratorInfo`. The latency added by this request should only be a couple hundred milliseconds at the maximum, otherwise the Orchestrator will likely be slow to respond with transcoded results as well. 

In case the latency is found to be excessive in the first iteration of testing we can move to a background service that updates  `OrchestratorInfo` for each orchestrator in the working set. 

- Assert no payment errors when Orchestrator and Broadcaster behave according to protocol rules 

- Assert that an Orchestrator honours previously advertised, non-expired `TicketParams`

- Assert that the Orchestrator responds with a payment related error when either the `TicketParams` are invalid or have expired 

- Assert that the Broadcaster correctly fails over if an `Orchestrator` doesn't honour previously advertised, non-expired `TicketParams`

- Assert that the Broadcaster fails over when generating a payment when the Orchestrator advertises out of data `TicketParams`

- Assert that `OrchestratorInfo` is returned in every successful transcode response. It doesn't need to be returned if transcoding is not successful because the Broadcaster will fail over anyway.

- Assert that when a Broadcaster deviates from the protocol by using _any_ price, it will run into an insufficient balance error after the first segment with a payment of 1 ticket with valid parameters

- Assert that tickets are only redeemed on-chain after their associated `TicketParams` have expired

## Known Unknowns

- Tracking accounting balances per Broadcaster address instead of Manifest ID is still an open issue and area of discussion. To ensure sufficient safeguarding by the accounting mechanism it _could_ be included in the scope of this project. 

- The usage cost and security of current open-source VRF implementations is still an unknown. Neither do we know the engineering cost associated with writing a cost-efficient, secure low level implementation currently.

- Switching from ECDSA signatures to VRF's would potentially require an additional security audit regardless of using an open-source or home tailored implementation as it is a significant change and is related heavily to the security of the payment protocol. 

- End-of-round optimisations get added to tech debt

## Alternatives

- [Dynamic acceptable error count](https://github.com/livepeer/go-livepeer/issues/1075#issuecomment-572628520)

- [Gracefully Retry on payment error](https://github.com/livepeer/go-livepeer/issues/1075#issuecomment-575548874)

## Additional Context

[Original discussion](https://github.com/livepeer/go-livepeer/issues/1075)

### PR DESCRIPTION
**Pull request by kyriediculous**

This PR completely removes any notion and usage of acceptable payment errors for processing payments while trying to not disrupt transcoding. Instead the orchestrator will now send `OrchestratorInfo` updates with segment served and it will have an `ExpirationBlock` attached. 

The goal is that payment errors should not degrade reliability and in an optimal happy workflow a payment error should now never be encountered anymore as a broadcaster should always have access to up to date `TicketParams` and `PriceInfo` for an orchestrator. 

- adcfc0e - Moves the `ratPriceInfo` helper into the `common/util.go` file and export it there since it is now used across multiple packages. 
- a720c3a - Adds `ExpirationBlock` to the `net.TicketParams` protobuf message and its counterpart types in the `pm` package.
- cb13981 - Orchestrator calculates the `ExpirationBlock` for its `OrchestratorInfo` and attaches it to `TicketParams`
- bc5262c - Attaches `WinProb`, `FaceValue`, `ExpirationBlock` and `PricePerPixel` to the HMAC generation which results in `recipientRand` so that an orchestrator can validate whether it has previously advertised the ticket params attached to a ticket upon receiving it. 
- a669c67 - Refreshes `OrchestratorInfo` for the currently selected orchestrator _after initial selection_ when the first attempt at generating a payment fails due to outdated ticketParams, the broadcaster will then query for updated params and retry submitting the segment.
- 63218d1 - Orchestrator now always send back the latest `OrchestratorInfo` when serving a segment instead of when a payment error is encountered. When a payment error is encountered a `Bad Request` HTTP error is returned instead.
- 5da0080 - Changes the `ticketQueue` that would store and retry tickets when the `maxFloat` for a `sender` is insufficient to a retry queue for all tickets. Each time we a new block is discovered tickets for which the params have expired will be redeemed on-chain , if the max-float isn't sufficient the ticket is added back to the queue. 
- b5d0320 - removes the `acceptablePrice` check from `orchestrator.ProcessPayment` as we no longer deal with acceptable or unacceptable errors. 
- 0f347a4 - remove dealing with acceptable and unacceptable ticket receive errors as well as keeping track of error counting through the `ErrorMonitor`
- e029da3 - extends and renames `roundsWatcher` to be a more generic `timeWatcher` that keeps track of all time related events in the protocol. Currently this would be rounds and blocks. 

Fixes #1075 
Fixes #1343 

**Checklist:**

### CODE REVIEW

**Comment by kyriediculous** [comment]
Resolved merge conflicts and rebased on top of master, squashing the fixup commits,  for easier review

File: monitor/census.go [hunk 1]
```diff
@@ -209,8 +208,7 @@ func InitCensus(nodeType, nodeID, version string) {
 	// Metrics for receiving payments
 	census.mTicketValueRecv = stats.Float64("ticket_value_recv", "TicketValueRecv", "gwei")
 	census.mTicketsRecv = stats.Int64("tickets_recv", "TicketsRecv", "tot")
-	census.mPaymentRecvAcceptableError = stats.Int64("payment_recv_acceptable_errors", "PaymentRecvAcceptableError", "tot")
-	census.mPaymentRecvUnacceptableError = stats.Int64("payment_recv_unacceptable_errors", "PaymentRecvUnacceptableError", "tot")
+	census.mPaymentRecvErr = stats.Int64("payment_recv__errors", "PaymentRecvErr", "tot")
```
yondonfu: Extra underscore? "payment_recv__errors" -> "payment_recv_errors"

pm/queue.go
```diff
@@ -136,19 +125,25 @@ func (q *ticketQueue) startQueueLoop() {
...
-				case q.redeemable <- nextTicket:
-					q.removeHead()
-				case <-q.quit:
-					return
+		case err := <-sub.Err():
+			glog.Error(err)
+		case latestBlock := <-blockNums:
+			numTickets := len(q.queue)
+			for i := 0; i < numTickets; i++ {
+				nextTicket := q.queue[0]
```
yondonfu, line 121: Accessing `q.queue` in this manner is not thread safe since there is no lock protection (previously we were only accessing `q.queue` prior to calling `q.cond.L.Unlock()`) and `Add()` could be called concurrently.

This can be fixed by moving the `nextTicket` declaration above `q.cond.L.Unlock()`. I'm not sure how important it is to to iterate through all tickets in the queue whenever we receive a new block - the only time where all N tickets would actually be submitted for redemption simultaneously is if there are N tickets that all have expired params *and* the sender's max float can cover all N tickets (otherwise most of the tickets would just be re-added to the queue). Seems like N would be quite small in practice and that pulling a single ticket off the queue when we receive a new block might be fine [1].

[1] If we do need to pull multiple tickets off the queue then you'll need to access them under a lock.

pm/queue.go
```diff
@@ -136,19 +125,25 @@ func (q *ticketQueue) startQueueLoop() {
...
+				q.removeHead()
+				if nextTicket.ParamsExpirationBlock.Cmp(latestBlock) < 0 {
```
yondonfu: If we're saying that `ParamsExpirationBlock` is the block number at which the params expire then this should be:

```
if nextTicket.ParamsExpirationBlock.Cmp(latestBlock) <= 0 { ... }
```

pm/queue_test.go
```diff
@@ -64,29 +66,21 @@ func TestTicketQueueLoop(t *testing.T) {
...
 
-	for i := 0; i < numTickets; i++ {
-		q.SignalMaxFloat(big.NewInt(60))
-	}
+	// Test signaling a new blockNum and remove tickets
+	tm.blockNumSink <- big.NewInt(1)
 
 	// Queue should be empty now
-	time.Sleep(time.Millisecond * 20)
+	time.Sleep(time.Second * 2)
```
yondonfu: Does the sleep need to be this long? Millisecond -> second seems like a big jump.

pm/queue_test.go
```diff
@@ -64,29 +66,21 @@ func TestTicketQueueLoop(t *testing.T) {
 
 	assert.Equal(int32(numTickets), q.Length())
 
-	// Test signaling max float without removing tickets
-
-	q.SignalMaxFloat(big.NewInt(40))
```
yondonfu, line 69: Let's include tests for signaling a new block without removing tickets (i.e. the next ticket in the queue does not have expired params).

pm/queue_test.go
```diff
@@ -99,47 +93,3 @@ func TestTicketQueueLoop(t *testing.T) {
 		assert.Equal(uint32(i), redeemable[i].SenderNonce)
 	}
 }
-
-func TestTicketQueueLoopConcurrent(t *testing.T) {
```
yondonfu, line 103: We still want to include tests that ensure that the queue is concurrency safe given the changes in this PR. We could make some tweaks to this test function to reflect signaling to the queue using new blocks instead of new max float values.

pm/recipient.go
```diff
@@ -159,51 +163,73 @@ func (r *recipient) ReceiveTicket(ticket *Ticket, sig []byte, seed *big.Int) (st
...
 	}
 
-	return sessionID, won, r.acceptTicket(ticket, sig, recipientRand)
+	if err := r.updateSenderNonce(recipientRand, ticket.SenderNonce); err != nil {
+		return sessionID, won, err
+	}
+
+	// check advertised params aren't expired
+	latestBlock := r.tm.LastSeenBlock()
+	if ticket.ParamsExpirationBlock.Cmp(latestBlock) < 0 {
```
yondonfu: If we're saying that `ParamsExpirationBlock` is the block at which the params expire then we should use <= instead of < for this comparison.

pm/recipient.go
```diff
@@ -15,10 +16,13 @@ import (
 )
 
 var errInsufficientSenderReserve = errors.New("insufficient sender reserve")
+var ErrTicketParamsExpired = errors.New("TicketParams expired")
 
 // maxWinProb = 2^256 - 1
 var maxWinProb = new(big.Int).Sub(new(big.Int).Lsh(big.NewInt(1), 256), big.NewInt(1))
 
+var paramsExpirationTime = big.NewInt(1)
```
yondonfu: Something like `paramsBlocksToExpiration` that mentions blocks as the units of time might be clearer.

Also, I think this is too short - as discussed in the spec it should be at least 2 blocks since nodes are not guaranteed to always have the same view of the latest block. IMO we can make this even longer to something like 10 blocks to target ~2 minutes to allow for more breathing room - this expands the time period in which B can use slightly outdated payment params, but the impact is based on how much the estimated gas price for tx inclusion can change and how much B's max float can change in that time period. 2 minutes seems ok to me.

server/broadcast.go
```diff
@@ -102,6 +106,8 @@ func (bsm *BroadcastSessionsManager) completeSession(sess *BroadcastSession) {
 	bsm.sessLock.Lock()
 	defer bsm.sessLock.Unlock()
 
+	fmt.Println("EXISTING", bsm.sessMap[sess.OrchestratorInfo.Transcoder])
```
yondonfu: Can remove

cmd/livepeer/livepeer.go
```diff
@@ -488,21 +486,17 @@ func main() {
 			}
 
 			sigVerifier := &pm.DefaultSigVerifier{}
-			validator := pm.NewValidator(sigVerifier, roundsWatcher)
+			validator := pm.NewValidator(sigVerifier, timeWatcher)
 			gpm := eth.NewGasPriceMonitor(backend, blockPollingTime)
 			// Start gas price monitor
-			gasPriceUpdate, err := gpm.Start(ctx)
+			_, err := gpm.Start(ctx)
```
yondonfu, line 494: If we're no longer consuming values on the gas price update channel returned by `Start()` we can remove the logic in the gas price monitor for sending a value on the channel whenever the cached gas price changes. Otherwise, those channel sends will block.

core/orchestrator.go
```diff
@@ -127,36 +127,23 @@ func (orch *orchestrator) ProcessPayment(payment net.Payment, manifestID Manifes
...
 
-		if monitor.Enabled {
-			monitor.PaymentRecvError(sender.String(), string(manifestID), err.Error(), ok && acceptablePriceErr.Acceptable())
-		}
-	}
+	seed := new(big.Int).SetBytes(payment.TicketParams.Seed)
 
-	if err == nil || (ok && acceptablePriceErr.Acceptable()) {
-		acceptablePrice = true
+	priceInfoRat, err := common.RatPriceInfo(priceInfo)
```
yondonfu, line 134: Do we use `priceInfo` anywhere besides here? If not could we just directly pass `payment.GetExpectedPrice()` here?

core/orchestrator.go
```diff
@@ -127,36 +127,23 @@ func (orch *orchestrator) ProcessPayment(payment net.Payment, manifestID Manifes
...
+	if err != nil {
+		return fmt.Errorf("invalid expected price sent with payment - expectedPrice=%v err=%v", priceInfo, err)
```
yondonfu: `priceInfo` is a `net.PriceInfo`. Should we instead use `priceInfoRat.FloatString(3)`?

core/orchestrator.go
```diff
@@ -249,7 +225,16 @@ func (orch *orchestrator) TicketParams(sender ethcommon.Address) (*net.TicketPar
 		return nil, nil
 	}
 
-	params, err := orch.node.Recipient.TicketParams(sender)
+	price, err := orch.PriceInfo(sender)
```
yondonfu: This method internally creates a `big.Rat` and then converts it into a `net.PriceInfo`. As a result, by calling `orch.PriceInfo()` and then `common.RatPriceInfo()` we're doing an unnecessary conversion by converting from `big.Rat` to `net.PriceInfo` and then back to `big.Rat`. Maybe we can extract that logic into a helper `orch.priceInfo()` which can be used in `orch.PriceInfo()` and directly called here.

eth/watchers/timewatcher.go
```diff
@@ -0,0 +1,233 @@
...
+	"github.com/livepeer/go-livepeer/eth/contracts"
+
+	"github.com/golang/glog"
+	"github.com/livepeer/go-livepeer/eth"
+	"github.com/livepeer/go-livepeer/eth/blockwatch"
+)
+
+// TimeWatcher is a type for a thread safe in-memory cache that watches for the following on-chain events:
+// 	* New rounds
+//  * New blocks
```
yondonfu: Looks like this point is misaligned with the rest.

eth/watchers/timewatcher.go
```diff
@@ -0,0 +1,233 @@
...
+}
+
+func (tw *TimeWatcher) setTranscoderPoolSize(size *big.Int) {
+	tw.mu.Lock()
+	defer tw.mu.Unlock()
+	tw.transcoderPoolSize = size
+}
+
+func (tw *TimeWatcher) LastSeenBlock() *big.Int {
+	tw.mu.Lock()
```
yondonfu: We can use `RLock`/`RUnlock` here.

pm/recipient.go
```diff
@@ -159,51 +163,76 @@ func (r *recipient) ReceiveTicket(ticket *Ticket, sig []byte, seed *big.Int) (st
...
+		return sessionID, won, err
+	}
+
+	// check advertised params aren't expired
+	latestBlock := r.tm.LastSeenBlock()
+	if ticket.ParamsExpirationBlock.Cmp(latestBlock) < 0 {
+		return sessionID, won, ErrTicketParamsExpired
+	}
+
+	if !r.validRand(recipientRand) {
```
yondonfu: Previously, this check occurred before the `updateSenderNonce()` call which allows the method to fail early and not bother doing state updates if the recipientRand was already revealed. Can we do that here as well?

pm/recipient.go
```diff
@@ -159,51 +163,76 @@ func (r *recipient) ReceiveTicket(ticket *Ticket, sig []byte, seed *big.Int) (st
...
 	if err != nil {
 		return err
 	}
 
 	for i := 0; i < len(tickets); i++ {
-		if err := r.redeemWinningTicket(tickets[i], sigs[i], recipientRands[i]); err != nil {
-			return err
-		}
+		r.sm.QueueTicket(tickets[i].Sender, &SignedTicket{tickets[i], sigs[i], recipientRands[i]})
+		glog.Infof("Queued ticket sender=%x recipientRandHash=%x senderNonce=%v", tickets[i].Sender.Hex(), tickets[i].RecipientRandHash.Hex(), tickets[i].SenderNonce)
```
yondonfu: Should we move this log statement into `QueueTicket()` so we don't need to remember to include the log statement every time we call `QueueTicket()` in this file?

core/orch_test.go
```diff
@@ -929,89 +946,34 @@ func TestProcessPayment_GivenReceiveTicketError_ReturnsError(t *testing.T) {
...
 	manifestID := ManifestID("some manifest")
 
 	recipient.On("TxCostMultiplier", mock.Anything).Return(big.NewRat(1, 1), nil)
 	recipient.On("ReceiveTicket", mock.Anything, mock.Anything, mock.Anything).Return("", false, errors.New("ReceiveTicket error")).Once()
-	// This should trigger a redemption even though it returns an error because it still returns won = true
-	recipient.On("ReceiveTicket", mock.Anything, mock.Anything, mock.Anything).Return("", true, errors.New("not first error")).Once()
 	recipient.On("ReceiveTicket", mock.Anything, mock.Anything, mock.Anything).Return("", true, nil).Once()
 
-	numTickets := 3
+	numTickets := 2
```
yondonfu, line 962: What's the reason for this change?

pm/sender.go
```diff
@@ -155,12 +154,17 @@ func (s *sender) validateTicketParams(ticketParams *TicketParams, numTickets int
 		return fmt.Errorf("ticket faceValue %v > max faceValue %v", ticketParams.FaceValue, maxFaceValue)
 	}
 
+	latestBlock := s.timeManager.LastSeenBlock()
+	if ticketParams.ExpirationBlock == nil || ticketParams.ExpirationBlock.Cmp(latestBlock) < 0 {
```
yondonfu: `<=` instead of `<` if we're saying that `ExpirationBlock` is the block at which the params expire.

pm/sendermonitor_test.go
```diff
@@ -206,29 +186,32 @@ func TestQueueTicketAndSignalMaxFloat(t *testing.T) {
...
-	require.Nil(err)
+	tm.blockNumSink <- big.NewInt(5)
+	time.Sleep(20 * time.Millisecond)
+	assert.Equal(sm.(*senderMonitor).senders[addr].queue.Length(), int32(0))
 
-	time.Sleep(time.Millisecond * 20)
 	tickets := qc.Redeemable()
 	assert.Equal(1, len(tickets))
-	assert.Equal(uint32(0), tickets[0].SenderNonce)
+	assert.Equal(uint32(0), tickets[0].Ticket.SenderNonce)
```
yondonfu: Why not `.SenderNonce`?

pm/sendermonitor_test.go
```diff
@@ -206,29 +186,32 @@ func TestQueueTicketAndSignalMaxFloat(t *testing.T) {
...
+
+	// check that ticket is now removed from queue
+	assert.Equal(sm.(*senderMonitor).senders[addr].queue.Length(), int32(0))
```
yondonfu: This check doesn't seem to be necessary given the check on L207

server/rpc_test.go
```diff
@@ -270,284 +275,6 @@ func TestRPCSeg(t *testing.T) {
 	o.sessCapErr = nil
 }
 
-func TestEstimateFee(t *testing.T) {
```
yondonfu: I see that these tests have been moved - code shuffling is fine if it leads to a more preferable structure, but can we leave that out of this already big PR? Including it here makes it more difficult to spot logical changes that might have occurred, but that are obscured by code blocks being moved around.

yondonfu (changes requested):

pm/queue.go
kyriediculous: I also expected N to be quite small and therefore assumed just looping through might be the best trade-off to make considering end-of-round optimization and that we really want to try and redeem all current tickets in the queue before the round ends. 

Good point about thread safety.

pm/queue_test.go
kyriediculous: Probably not, I likely changed this for debugging purposes and forgot to change it back.

core/orch_test.go
kyriediculous: IIRC my reasoning here was that we no longer needed to test the difference between acceptable and unacceptable so to test the error paths for this test two tickets were sufficient.

server/broadcast.go
kyriediculous: should be removed in a subsequent fixup but don't remember exactly which one

cmd/livepeer/livepeer.go
kyriediculous: fixed in 97ed5638f35534d5bda4589008caf9a854d59ab6

core/orchestrator.go
kyriediculous: Impossible as this error will only occur when dividing by zero but the caller isn't necessarily aware of that. 

When `priceInfo` is `nil`  the return values are `(nil, nil)`

On another note , should this be `if priceInfoRat == nil || err != nil` instead?

pm/recipient.go
kyriediculous: Good point.

Would you like to add ticket queue monitoring as well? We can display the amount of tickets in the queue in grafana. 

Can always do the latter in a follow-up PR as well.

server/rpc_test.go
kyriediculous: Yeah you're right. Want me to patch mode this out and submit a seperate PR for it? Although I can't retrieve the old commits , those have been rebased.

yondonfu: Yeah let's move these changes out of this PR.

monitor/census.go
kyriediculous: fixed in a23df3276b2e6b3c90c3046951e25f4ad3f04f02

pm/queue.go
kyriediculous: On second thought, I removed this test because we don't actually signal block number updates concurrently like we did with max float signalling. 

I should keep testing concurrent adds though.

pm/recipient.go
kyriediculous: What do you think of 4 blocks? 

This would make the lifetime of a ticket fall within the window of the moving average of the last 5 blocks (iirc) geth uses to suggest the gas price.

eth/watchers/timewatcher.go
kyriediculous: hmm , strange 
image

core/orchestrator.go
kyriediculous: In the error message

core/orchestrator.go
kyriediculous: fixed in 6b93f5b2

pm/recipient.go
kyriediculous: fixed in https://github.com/livepeer/go-livepeer/pull/1383/commits/3e727d0c0ee87fff1c658785375198cb5b40c3e9

pm/recipient.go
kyriediculous: fixed 4860c1fe965df76518e4595d9599f7026730805b 5cbf05933fa8dbfbc5690d013eda905eaf95355e

eth/watchers/timewatcher.go
kyriediculous: fixed in ba4677189d4014db7663b387af817ebb72c4abe7

pm/sender.go
kyriediculous: fixed in 7151a7aa2b1847c9f7f57e4ab991e7a4d7d309c0

pm/recipient.go
kyriediculous: ff653b187cafa1255db5f151f2a8549c38810560

pm/recipient.go
kyriediculous: fixed in 0080b33409cc8e06120cf64808c05fa582a5d76e

pm/queue_test.go
kyriediculous: fixed in 756fe853cb627a508e1f33e21725a8fc7b465805

pm/queue_test.go
kyriediculous: fixed in 756fe853cb627a508e1f33e21725a8fc7b465805

pm/queue_test.go
kyriediculous: fixed in 756fe853cb627a508e1f33e21725a8fc7b465805

pm/queue.go
kyriediculous: fixed in 756fe853cb627a508e1f33e21725a8fc7b465805

pm/queue.go
kyriediculous: fixed in 756fe853cb627a508e1f33e21725a8fc7b465805

pm/sendermonitor_test.go
kyriediculous: fixed in 4860c1fe965df76518e4595d9599f7026730805b

pm/sendermonitor_test.go
kyriediculous: fixed in 4860c1fe965df76518e4595d9599f7026730805b

pm/recipient.go
yondonfu: > This would make the lifetime of a ticket fall within the window of the moving average of the last 5 blocks (iirc) geth uses to suggest the gas price.

I don't think we should try to optimize this value based on geth's behavior since the # of past blocks that the gas price estimator will look back is configurable so we should steer away from making too many assumptions about the frequency of suggested gas price changes. I think we should lean towards starting with more breathing room with the expiration time. What about 5 blocks i.e. ~1 min (only 1 more than the current value, but approximating 1 min feels easier to justify)?

core/orchestrator.go
yondonfu: If `RatPriceInfo()` returns an error then `priceInfo.PricePerUnit = 0` I'm not sure how useful logging the value of `priceInfo` is in that scenario since the error message will contain all the necessary information already. If we don't log `priceInfo` here then we don't need to manage the `priceInfo` and `priceInfoRat` variables since we can allocate `priceInfoRat` with a single call.

> On another note , should this be if priceInfoRat == nil || err != nil instead?

Yeah that would be a good defensive guard. Without it, a B that sends a payment with a nil `ExpectedPrice` would cause O to crash. I think we should separate the checks though: if `priceInfoRat == nil` we would return an error indicating a missing price and if `err != nil` we would return an error indicating why the expected price was invalid.

pm/recipient.go
yondonfu: > Would you like to add ticket queue monitoring as well? We can display the amount of tickets in the queue in grafana.

I can see how that might be useful down the line, but let's discuss that outside of this PR.

core/orchestrator.go
```diff
@@ -286,6 +265,20 @@ func (orch *orchestrator) PriceInfo(sender ethcommon.Address) (*net.PriceInfo, e
 	}, nil
 }
 
+func (orch *orchestrator) priceInfo(sender ethcommon.Address) (*big.Rat, error) {
+	if orch.node == nil || orch.node.Recipient == nil {
```
yondonfu: We can let the caller handle this check - `TicketParams()` already does this. If we make this change, then we don't need the `price == nil` check in `PriceInfo()`.

eth/watchers/timewatcher.go
yondonfu: Maybe a spaces vs. tabs thing? Could try re-indenting all points. This isn't a big deal though.

pm/recipient.go
```diff
@@ -15,10 +16,13 @@ import (
 )
 
 var errInsufficientSenderReserve = errors.New("insufficient sender reserve")
+var ErrTicketParamsExpired = errors.New("TicketParams expired")
```
yondonfu, line 19: Can we group exported errors separately from non-exported errors (i.e. declare the exported errors up top and then declare the non-exported errors)? IMO makes the distinction b/w variables for external vs. internal use clearer.

pm/recipient.go
```diff
@@ -159,51 +163,75 @@ func (r *recipient) ReceiveTicket(ticket *Ticket, sig []byte, seed *big.Int) (st
...
 	return nil
 }
 
 // RedeemWinningTicket redeems a single winning ticket
 func (r *recipient) RedeemWinningTicket(ticket *Ticket, sig []byte, seed *big.Int) error {
-	recipientRand := r.rand(seed, ticket.Sender)
-	return r.redeemWinningTicket(ticket, sig, recipientRand)
+	recipientRand := r.rand(seed, ticket.Sender, ticket.FaceValue, ticket.WinProb, ticket.ParamsExpirationBlock, ticket.PricePerPixel)
+	r.sm.QueueTicket(ticket.Sender, &SignedTicket{ticket, sig, recipientRand})
+	glog.Infof("Queued ticket sender=%x recipientRandHash=%x senderNonce=%v", ticket.Sender, ticket.RecipientRandHash, ticket.SenderNonce)
```
yondonfu: Can be removed since `QueueTicket()` logs a message now.

pm/recipient.go
```diff
@@ -159,51 +163,75 @@ func (r *recipient) ReceiveTicket(ticket *Ticket, sig []byte, seed *big.Int) (st
...
-	recipientRand := r.rand(seed, sender)
-	recipientRandHash := crypto.Keccak256Hash(ethcommon.LeftPadBytes(recipientRand.Bytes(), uint256Size))
 
 	faceValue, err := r.faceValue(sender)
 	if err != nil {
 		return nil, err
 	}
 
+	lastBlock := r.tm.LastSeenBlock()
+	if lastBlock == nil {
```
yondonfu: We don't check for nil when calling `LastSeenBlock()` elsewhere (i.e. in `ReceiveTicket()`) since this should always be populated - can we remove this?

pm/sender.go
```diff
@@ -155,12 +154,17 @@ func (s *sender) validateTicketParams(ticketParams *TicketParams, numTickets int
 		return fmt.Errorf("ticket faceValue %v > max faceValue %v", ticketParams.FaceValue, maxFaceValue)
 	}
 
+	latestBlock := s.timeManager.LastSeenBlock()
+	if ticketParams.ExpirationBlock == nil || ticketParams.ExpirationBlock.Cmp(latestBlock) <= 0 {
```
yondonfu: This is always going to return true when B works with an O that has not upgraded to support param expiration. We could remove the `ticketParams.ExpirationBlock == nil` check and the implication would be:

- B is able to work with Os that have not upgraded to support param expiration
- A malicious O could send params with `ExpirationBlock = nil` and then respond to the segment with an error claiming that the params have expired for the payment. B would just drop O from its working set in this case (B only tries to re-fetch params from O if B is the one that detects expired params and not if O responds with an error claiming that the params have expired)

server/segment_rpc.go
```diff
@@ -66,37 +66,29 @@ func (h *lphttp) ServeSegment(w http.ResponseWriter, r *http.Request) {
...
-		oInfo, err = orchestratorInfo(orch, sender, orch.ServiceURI().String())
-		if err != nil {
-			glog.Errorf("Error updating orchestrator info: %v", err)
-			http.Error(w, "Internal Server Error", http.StatusInternalServerError)
-			return
-		}
-
-		glog.Errorf("Acceptable error occured when processing payment: %v", paymentError)
+	if err := orch.ProcessPayment(payment, segData.ManifestID); err != nil {
+		glog.Errorf("error occured processing payment: %v", err)
```
yondonfu: For brevity, we can just say "error processing payment"

server/broadcast.go
```diff
@@ -206,12 +210,19 @@ func selectOrchestrator(n *core.LivepeerNode, params *streamParameters, cpl core
...
+			ticketParams *pm.TicketParams
+		)
 
 		if n.Sender != nil {
+			ticketParams = pmTicketParams(tinfo.TicketParams)
+			ratPrice, err := common.RatPriceInfo(tinfo.PriceInfo)
+			if err != nil {
+				return sessions, err
+			}
+			ticketParams.PricePerPixel = ratPrice
```
yondonfu: Doesn't seem like we need this - the `PricePerPixel` field isn't actually used by `n.Sender`

server/broadcast.go
```diff
@@ -347,13 +359,30 @@ func transcodeSegment(cxn *rtmpConnection, seg *stream.HLSSegment, name string,
...
+	res, tErr := SubmitSegment(sess, seg, nonce)
+	// If TicketParams have expired, refresh them
+	if tErr != nil && tErr == pm.ErrTicketParamsExpired {
+		glog.V(common.VERBOSE).Infof("ticketparams expired, refreshing for orchestrator=%v", sess.OrchestratorInfo.Transcoder)
+		oURL, err := url.Parse(sess.OrchestratorInfo.Transcoder)
+		if err != nil {
+			cxn.sessManager.removeSession(sess)
+			return nil, fmt.Errorf("invalid orchestrator URL url=%v", sess.OrchestratorInfo.Transcoder)
+		}
+		oInfo, err := getOrchestratorInfoRPC(context.Background(), sess.Broadcaster, oURL)
```
yondonfu: We need to pass in a context with deadline here or else this will never time out.

server/broadcast.go
```diff
@@ -347,13 +359,30 @@ func transcodeSegment(cxn *rtmpConnection, seg *stream.HLSSegment, name string,
...
+		if err != nil {
+			cxn.sessManager.removeSession(sess)
+			return nil, fmt.Errorf("unable to refresh ticketparams for orchestrator=%v, err=%v", sess.OrchestratorInfo.Transcoder, err)
+		}
+		sess.OrchestratorInfo = oInfo
```
yondonfu: Could we wrap all of this into a helper function `refreshSession()` that uses the existing `updateSession()` function to return a copied session with updated fields (if the `OrchestratorInfo` changes we might need to update additional fields in the session). Something like:

```
func refreshSession(sess *BroadcastSession) (*BroadcastSession, error) {
	uri, err := url.Parse(sess.OrchestratorInfo.Transcoder)
	if err != nil {
		return nil, err
	}

	refreshTimeout := 3 * time.Second
	ctx, cancel := context.WithTimeout(context.Background(), refreshTimeout)
	defer cancel()

	oInfo, err := GetOrchestratorInfo(ctx, sess.Broadcaster, uri)
	if err != nil {
		return nil, err
	}

        // Create dummy result
	res := &ReceivedTranscodeResult{
		LatencyScore: sess.LatencyScore,
		Info:         oInfo,
	}

	return updateSession(sess, res), nil
}
```

If we use a helper, we would only need to worry about calling `removeSession()` if an error is returned by the helper.

yondonfu (changes requested):

server/broadcast.go
```diff
@@ -347,13 +359,30 @@ func transcodeSegment(cxn *rtmpConnection, seg *stream.HLSSegment, name string,
 	// send segment to the orchestrator
 	glog.V(common.DEBUG).Infof("Submitting segment nonce=%d manifestID=%s seqNo=%d orch=%s", nonce, cxn.mid, seg.SeqNo, sess.OrchestratorInfo.Transcoder)
 
-	res, err := SubmitSegment(sess, seg, nonce)
-	if err != nil || res == nil {
+	res, tErr := SubmitSegment(sess, seg, nonce)
+	// If TicketParams have expired, refresh them
+	if tErr != nil && tErr == pm.ErrTicketParamsExpired {
```
yondonfu: Could we avoid some of this nesting and additional error checking if we called `sender.ValidateTicketParams()` instead? Something like:

```
err := sess.Sender.ValidateTicketParams(params)
if err != nil {
	if err != pm.ErrTicketParamsExpired {
		return nil, err
	}

	sess, err = refreshSession(sess)
	if err != nil {
		cxn.sessManager.removeSession(sess)
		return nil, err
	}
}

// Call SubmitSegment
```

yondonfu: Could some of this extra nesting + error/nil checking be avoided if we use `sender.ValidateTicketParams()` here instead? Something like:

```
if sess.Sender != nil {
	err := sess.Sender.ValidateTicketParams(params)
	if err != nil {
		if err != pm.ErrTicketParamsExpired {
			return nil, err
		}

		sess, err = refreshSession(sess)
		if err != nil {
			cxn.sessManager.removeSession(sess)
			return nil, err
		}
	}
}

// Call SubmitSegment
```

This might simplify tests a bit as well.

server/rpc_test.go
kyriediculous: fixed in 24f0a5c

core/orchestrator.go
kyriediculous: fixed in dc2227a5

eth/watchers/timewatcher.go
kyriediculous: oh yes, indeed. Fixed in 5fde1f56

pm/recipient.go
kyriediculous: fixed in fa5f61d1

pm/recipient.go
kyriediculous: sounds good, fixed in 0c3596c5

pm/recipient.go
kyriediculous: fixed in 8b60a0fd

pm/recipient.go
kyriediculous: fixed in 2d5dbd07

pm/sender.go
kyriediculous: Actually `ticketParams.ExpirationBlock` can never be nil here now that I think of it. Because when converting `net.TicketParams` to `pm.TicketParams` we essentially set it to be `new(big.Int).SetBytes(nil)` which would be `0` and not `nil` 

Therefore I think what we would need to do to support the features you describe here is to add a check before this block 

```
if ticketParams.ExpirationBlock.Int64() == 0 { return nil }
```

kyriediculous: fixed in cf577f2a6c336ebe5fa9f78d359e5c4b34bb7be9

server/segment_rpc.go
kyriediculous: fixed in 0dbcf97df79a4ff27e77f4cc545ff7e9a11950cb

server/broadcast.go
kyriediculous: fixed in 35ac30958477f13a16d30689c8b7e98be39e3c4d

server/broadcast.go
kyriediculous: What would be a sensible timeout value here? I currently set it at 2.5 sec in afb03034

kyriediculous: oh sorry , I just checked the code snipped below. Nvm

server/broadcast.go
kyriediculous: fixed in c274f67c5296ea35f0c635d0b86ef5780506cf4a & ebb7532edd782df1b2352b1beb5e46159687474b

server/broadcast.go
kyriediculous: fixed in c274f67c5296ea35f0c635d0b86ef5780506cf4a & ebb7532edd782df1b2352b1beb5e46159687474b

core/orchestrator.go
yondonfu: Bump

server/broadcast.go
```diff
@@ -346,12 +356,28 @@ func transcodeSegment(cxn *rtmpConnection, seg *stream.HLSSegment, name string,
...
 	// send segment to the orchestrator
 	glog.V(common.DEBUG).Infof("Submitting segment nonce=%d manifestID=%s seqNo=%d orch=%s", nonce, cxn.mid, seg.SeqNo, sess.OrchestratorInfo.Transcoder)
+	if sess.Sender != nil {
+		err := sess.Sender.ValidateTicketParams(pmTicketParams(sess.OrchestratorInfo.TicketParams))
+		if err != nil {
+			if err != pm.ErrTicketParamsExpired {
+				return nil, err
+			}
 
+			glog.V(common.VERBOSE).Infof("ticketparams expired, refreshing for orchestrator=%v", sess.OrchestratorInfo.Transcoder)
```
yondonfu: We use "ticket params" instead of "ticketparams" elsewhere ([example](https://github.com/livepeer/go-livepeer/blob/d3f536acf007f8a98edf76f007f2e6ea5517987d/discovery/db_discovery.go#L102)) so I think we should stick with the former for consistency.

server/broadcast.go
```diff
@@ -346,12 +356,28 @@ func transcodeSegment(cxn *rtmpConnection, seg *stream.HLSSegment, name string,
 
 	// send segment to the orchestrator
 	glog.V(common.DEBUG).Infof("Submitting segment nonce=%d manifestID=%s seqNo=%d orch=%s", nonce, cxn.mid, seg.SeqNo, sess.OrchestratorInfo.Transcoder)
+	if sess.Sender != nil {
+		err := sess.Sender.ValidateTicketParams(pmTicketParams(sess.OrchestratorInfo.TicketParams))
```
yondonfu: Since `err` isn't used outside of this code block I think we can do:

```golang
if err := sess.Sender.ValidateTicketParams(...); err != nil { ... }
```

for brevity since we only need to keep `err` around in the scope of this code block.

server/broadcast.go
yondonfu: We use `orch` as the key name in log messages elsewhere so I think we should stick with that convention for consistency here.

server/broadcast.go
```diff
@@ -346,12 +356,28 @@ func transcodeSegment(cxn *rtmpConnection, seg *stream.HLSSegment, name string,
...
+		if err != nil {
+			if err != pm.ErrTicketParamsExpired {
+				return nil, err
+			}
 
+			glog.V(common.VERBOSE).Infof("ticketparams expired, refreshing for orchestrator=%v", sess.OrchestratorInfo.Transcoder)
+			newSess, err := refreshSession(sess)
+			if err != nil {
+				cxn.sessManager.removeSession(sess)
+				return nil, fmt.Errorf("unable to refresh ticketparams for orchestrator=%v, err=%v", sess.OrchestratorInfo.Transcoder, err)
```
yondonfu: We don't use commas in between key=value pairs in log messages elsewhere so I think we should stick with that convention for consistency here.

yondonfu: See above comment about using `orch` as the key name.

yondonfu: See above comment about "ticket params"

server/broadcast.go
```diff
@@ -346,12 +356,28 @@ func transcodeSegment(cxn *rtmpConnection, seg *stream.HLSSegment, name string,
...
+			sess = newSess
+		}
+	}
 	res, err := SubmitSegment(sess, seg, nonce)
 	if err != nil || res == nil {
+		glog.Infof("unable to submit segment, removing session for orchestrator=%v err=%v", sess.OrchestratorInfo.Transcoder, err)
 		cxn.sessManager.removeSession(sess)
 		if res == nil && err == nil {
-			return nil, errors.New("Empty response")
+			return nil, fmt.Errorf("unable to submit segment, removing session for orchestrator=%v err=%v", sess.OrchestratorInfo.Transcoder, err)
```
yondonfu: I'm on board with annotating errors. But, as-is this error is not being used/logged anywhere. If we're going to do this I'd prefer if we return annotated errors so that they can be handled by the caller (i.e. for logging) and *not* also log the error in the callee (as is done on L377). In this case, we could log the error in [processSegment()](https://github.com/livepeer/go-livepeer/blob/d3f536acf007f8a98edf76f007f2e6ea5517987d/server/broadcast.go#L301).

Can we do this outside of this PR?

Side note: When we do this, `err` is nil here so it would probably be more informative if the error included indicated that the response was empty like the error that is returned in the current code prior to this PR.

server/broadcast_test.go
```diff
@@ -451,6 +460,115 @@ func TestCleanupSessions(t *testing.T) {
...
+	successOrchInfoUpdate := &net.OrchestratorInfo{
+		Transcoder: ts.URL,
+		PriceInfo: &net.PriceInfo{
+			PricePerUnit:  1,
+			PixelsPerUnit: 1,
+		},
+		TicketParams: &net.TicketParams{},
+	}
+
+	getOrchestratorInfoRPC = func(ctx context.Context, bcast common.Broadcaster, orchestratorServer *url.URL) (*net.OrchestratorInfo, error) {
```
yondonfu, line 565: Let's make sure to reset `getOrchestratorInfoRPC` after the test function returns so that other tests that depend on the variable start with a clean slate. Ex.

```golang
oldGetOrchestratorInfoRPC = getOrchestratorInfoRPC
defer func () { getOrchestratorInfoRPC = oldGetOrchestorInfoRPC }()
...
```

server/broadcast_test.go
```diff
@@ -955,3 +1067,20 @@ func TestVerifier_HLSInsertion(t *testing.T) {
 	assert.Nil(err)
 	assert.Equal(baseURL+"/resp2", pl.uri)
 }
+
+func defaultTicketbatch() *pm.TicketBatch {
```
yondonfu: Nit: "TicketBatch"

server/broadcast.go
```diff
@@ -346,12 +356,28 @@ func transcodeSegment(cxn *rtmpConnection, seg *stream.HLSSegment, name string,
 
 	// send segment to the orchestrator
 	glog.V(common.DEBUG).Infof("Submitting segment nonce=%d manifestID=%s seqNo=%d orch=%s", nonce, cxn.mid, seg.SeqNo, sess.OrchestratorInfo.Transcoder)
+	if sess.Sender != nil {
+		err := sess.Sender.ValidateTicketParams(pmTicketParams(sess.OrchestratorInfo.TicketParams))
+		if err != nil {
+			if err != pm.ErrTicketParamsExpired {
+				return nil, err
```
yondonfu, line 366: I think we'll need to call `cxn.sessManager.removeSession()` - if the error is not an expiry error then we just remove the O.

server/broadcast.go
```diff
@@ -604,3 +630,26 @@ func updateSession(sess *BroadcastSession, res *ReceivedTranscodeResult) *Broadc
...
+		return nil, err
+	}
+
+	// Create dummy result
+	res := &ReceivedTranscodeResult{
+		LatencyScore: sess.LatencyScore,
+		Info:         oInfo,
+	}
+
+	return updateSession(sess, res), nil
```
yondonfu, line 655: Let's add a test that ensures that the returned session actually uses the latest `OrchestratorInfo` returned by the `getOrchestratorInfoRPC`

server/broadcast.go
```diff
@@ -604,3 +630,26 @@ func updateSession(sess *BroadcastSession, res *ReceivedTranscodeResult) *Broadc
...
 	return newSess
 }
+
+func refreshSession(sess *BroadcastSession) (*BroadcastSession, error) {
+	uri, err := url.Parse(sess.OrchestratorInfo.Transcoder)
+	if err != nil {
+		return nil, err
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), refreshTimeout)
```
yondonfu, line 641: Let's add a test for the timeout. A stub value for `refreshTimeout` can be used in the test i.e. a few milliseconds.

server/broadcast.go
```diff
@@ -604,3 +630,26 @@ func updateSession(sess *BroadcastSession, res *ReceivedTranscodeResult) *Broadc
 
 	return newSess
 }
+
+func refreshSession(sess *BroadcastSession) (*BroadcastSession, error) {
```
yondonfu, line 636: Let's make sure to test the error paths in this function. `refreshSession()` can be unit tested on its own to avoid the test setup cruft involved with `transcodeSegment()`.

yondonfu (changes requested):

core/orchestrator.go
kyriediculous: fixed in 70aa54910e33c7c697878520e51a588bf2cd95be

server/broadcast.go
kyriediculous: not sure how to test for a timeout here

server/broadcast.go
kyriediculous: fixed all these small change requests in this file in 774341f6e04e79aade04aaa6b7e98a07fd26618d

server/broadcast_test.go
kyriediculous: fixed in 20340ceabc68452fb4c8914a568e965e1c5f5d30

server/broadcast_test.go
kyriediculous: fixed in 20340ceabc68452fb4c8914a568e965e1c5f5d30

server/broadcast.go
kyriediculous: fixed in 20340ceabc68452fb4c8914a568e965e1c5f5d30

server/broadcast.go
kyriediculous: fixed in 20340ceabc68452fb4c8914a568e965e1c5f5d30

server/broadcast.go
kyriediculous: fixed in 20340ceabc68452fb4c8914a568e965e1c5f5d30

server/broadcast.go
yondonfu: Something like the following might work:

```golang
oldRefreshTimeout = refreshTimeout
defer func() { refreshTimeout = oldRefreshTimeout }()
refreshTimeout = 10 * time.Millisecond

oldGetOrchestratorInfoRPC = getOrchestratorInfoRPC
defer func() { getOrchestratorInfoRPC = oldGetOrchestratorInfoRPC }()
getOrchestratorInfoRPC = func(ctx context.Context, bcast common.Broadcaster, serv *url.URL) (*net.OrchestratorInfo, error) {
    // Wait until the refreshTimeout has elapsed
    select {
        case <-ctx.Done():
        case <-time.After(20 * time.Millisecond):
            return nil, errors.New("wrong error")
    }

    return nil, errors.New("context timeout")
}

_, err := refreshSession(..)
// Test for a timeout related error here
```

kyriediculous: Thanks for the example ! 

Added in 913ef20e

server/broadcast.go
kyriediculous: Creating an issue to track this https://github.com/livepeer/go-livepeer/issues/1410

server/broadcast.go
```diff
@@ -346,12 +356,28 @@ func transcodeSegment(cxn *rtmpConnection, seg *stream.HLSSegment, name string,
...
+			if err != nil {
+				cxn.sessManager.removeSession(sess)
+				return nil, fmt.Errorf("unable to refresh ticket params for orch=%v err=%v", sess.OrchestratorInfo.Transcoder, err)
+			}
+			sess = newSess
+		}
+	}
 	res, err := SubmitSegment(sess, seg, nonce)
 	if err != nil || res == nil {
+		glog.Infof("unable to submit segment, removing session for orchestrator=%v err=%v", sess.OrchestratorInfo.Transcoder, err)
```
yondonfu: Since this is an error case, we should use `glog.Errorf` here.

yondonfu: `orch` instead of `orchestrator`? Same comment for L380.

server/broadcast.go
yondonfu: Thanks for creating the issue.

Looks like this error is annotated, but not the one on L382. Can we either annotate both or leave this error annotation out of this PR?

yondonfu (changes requested):

server/broadcast.go
kyriediculous: fixed in 501dc6ed, removed annotation for L380

kyriediculous: fixed in 501dc6ed

server/broadcast.go
kyriediculous: removed annotation for errors , we can display a more informative logging message in processSegment as you describe.

kyriediculous:
Testing results so far on latest build (100% win prob tickets). 

1 B , 1 OT 
1 concurrent stream (~15 minutes)
No payment errors encountered

1B, 1 OT
3 concurrent streams (~15 minutes)
Payment errors encountered:
- invalid creation round (5) 


1, 2OT
3 concurrent streams 
Test failover 
Successful failover, no payment errors encountered

server/broadcast.go
```diff
@@ -346,12 +356,28 @@ func transcodeSegment(cxn *rtmpConnection, seg *stream.HLSSegment, name string,
...
+			if err != nil {
+				cxn.sessManager.removeSession(sess)
+				return nil, fmt.Errorf("unable to refresh ticket params for orch=%v err=%v", sess.OrchestratorInfo.Transcoder, err)
+			}
+			sess = newSess
+		}
+	}
 	res, err := SubmitSegment(sess, seg, nonce)
 	if err != nil || res == nil {
+		glog.Errorf("unable to submit segment, removing session for orch=%v err=%v", sess.OrchestratorInfo.Transcoder, err)
```
yondonfu: Noticed while testing that this log statement is actually unneeded, because `SubmitSegment()` already logs errors for various error cases - for example: [when an HTTP request cannot be sent](https://github.com/livepeer/go-livepeer/blob/b6e07fee00114427cd8b7638976a53fc49bfcc54/server/segment_rpc.go#L355) and [when segment upload failed](https://github.com/livepeer/go-livepeer/blob/b6e07fee00114427cd8b7638976a53fc49bfcc54/server/segment_rpc.go#L377). As a result, we can end up with duplicate logs:

```
E0305 17:24:41.111521   11006 segment_rpc.go:347] Unable to submit segment orch=https://127.0.0.1:8936 nonce=14008752097281453067 manifestID=foo seqNo=24 orch=https://127.0.0.1:8936 err=Post https://127.0.0.1:8936/segment: dial tcp 127.0.0.1:8936: connect: connection refused
E0305 17:24:41.111548   11006 broadcast.go:377] unable to submit segment, removing session for orch=https://127.0.0.1:8936 err=Post https://127.0.0.1:8936/segment: dial tcp 127.0.0.1:8936: connect: connection refused
```

In the interest of avoiding log clutter, let's remove this log statement.

server/broadcast.go
```diff
@@ -346,12 +356,28 @@ func transcodeSegment(cxn *rtmpConnection, seg *stream.HLSSegment, name string,
...
 	// send segment to the orchestrator
 	glog.V(common.DEBUG).Infof("Submitting segment nonce=%d manifestID=%s seqNo=%d orch=%s", nonce, cxn.mid, seg.SeqNo, sess.OrchestratorInfo.Transcoder)
+	if sess.Sender != nil {
+		if err := sess.Sender.ValidateTicketParams(pmTicketParams(sess.OrchestratorInfo.TicketParams)); err != nil {
+			if err != pm.ErrTicketParamsExpired {
+				cxn.sessManager.removeSession(sess)
+				return nil, err
+			}
 
+			glog.V(common.VERBOSE).Infof("ticket params expired, refreshing for orch=%v", sess.OrchestratorInfo.Transcoder)
```
yondonfu: Nit: The rest of the informational log statements capitalize the first letter so it would good to stay consistent with that convention (i.e. "ticket params" -> "Ticket params")

```
I0305 17:24:41.111745   11006 broadcast.go:358] Submitting segment nonce=14008752097281453067 manifestID=foo seqNo=24 orch=https://127.0.0.1:8937
I0305 17:24:41.111773   11006 broadcast.go:366] ticket params expired, refreshing for orch=https://127.0.0.1:8937
I0305 17:24:41.111787   11006 rpc.go:223] Connecting RPC to https://127.0.0.1:8937
I0305 17:24:41.115996   11006 segment_rpc.go:634] Created new payment - manifestID=foo recipient=0x325100b05e0886c8BDE7D0be6AFD7d401C8055F6 faceValue=1000000000000 WEI winProb=1.0000000000 price=1.000 wei/pixel numTickets=1
```

server/broadcast.go
kyriediculous: Can I then close the issue I made regarding logging this in processsegment ?

yondonfu: Given that the current convention in both `transcodeSegment()` and `SubmitSegment()` is to log error messages in place instead of passing errors to their callers, I think we should stick with that convention for now - I'd be ok with closing the issue for now and if there is a strong desire to change this convention at a later point in time then the case for doing so can be made in a new issue then.

pm/queue.go
```diff
@@ -121,6 +106,10 @@ func (q *ticketQueue) Length() int32 {
 // is sufficient to cover the face value of the ticket at the head of the queue. If the max float is sufficient, we pop
 // the ticket at the head of the queue and send it into q.redeemable which an external listener can use to receive redeemable tickets
 func (q *ticketQueue) startQueueLoop() {
+	blockNums := make(chan *big.Int, 10)
```
yondonfu, line 109: While the queue is empty, this channel will fill up as new blocks are received and when it is full the sender will block when it tries to send another value into the channel.

Consider the following function from `TimeWatcher`:

```
func (tw *TimeWatcher) handleBlockNum(event *blockwatch.Event) {
	last := tw.LastSeenBlock()
	new := event.BlockHeader.Number
	if last == nil || last.Cmp(new) != 0 {
		tw.setLastSeenBlock(new)
		tw.blockSubFeed.Send(new)
	}
}
```

One of the subscribers to `tw.blockSubFeed` is created here and the `blockNums` channel is used by the subscriber to receive values. Once `blockNums` is full, `tw.blockSubFeed.Send()` will block preventing the `TimeWatcher` from returning from `handleBlockNum()`. This means that `TimeWatcher` will stop updating its cached values for the last seen block.

I think we can fix this by removing the use of the condvar `q.cond` to lock and wait until the queue is non-empty. Previously, the condvar allowed the loop to pause (instead of running continuously thereby avoiding a busy loop) until a ticket was actually appended to the queue. Now that we only pop from the queue when a new block is received, we could instead just wait for a block to be received on `blockNums` and skip the use of the condvar to wait for a ticket to be added to the queue.

```golang
for {
        // Condvar logic removed

	select {
	case err := <-sub.Err():
		glog.Error(err)
	case latestBlock := <-blockNums:
		// Same code that we have right now for popping tickets from the queue
		case <-q.quit:
			return
		}
	}
```

yondonfu (commented):
Changes look good - let's rebase and fix the current conflict with master and I'll give the PR a final pass before merging.

kyriediculous:
rebased !

yondonfu (approved):
LGTM pending green CI 🚢

Issue #1332, opened 2020-01-28
go-livepeer: incorrect context handling


mk-livepeer:
There are many calls to `context.Background()` across the `go-livepeer` codebase.  This is wrong.  Instead,
```ctx := context.Background()```
...should be assigned once at the top of the main function, and
```
ctx, cancel := context.WithCancel(ctx)
```
...should be used to fork the context as necessary.

This covers the `main()` function, but all the others will also need to be modified to take a forked context out of main, rather than creating a new `context.Background()`

A forked context from ctx, or ctx itself should be passed from `main()` to the downstream functions:
```
cmd/livepeer/livepeer.go:		chainID, err := backend.ChainID(context.Background())
cmd/livepeer/livepeer.go:			ctx, cancel := context.WithCancel(context.Background())
cmd/livepeer/livepeer.go:			gasPriceUpdate, err := gpm.Start(context.Background())
cmd/livepeer/livepeer.go:			rs.Start(context.Background())
cmd/livepeer/livepeer.go:		blockWatchCtx, cancel := context.WithCancel(context.Background())
cmd/livepeer/livepeer.go:			ctx, cancel := context.WithCancel(context.Background())
cmd/livepeer/livepeer.go:	msCtx, cancel := context.WithCancel(context.Background())
common/testutil.go:	return peer.NewContext(context.Background(), p)
core/orchestrator.go:	return context.WithTimeout(context.Background(), transcodeLoopTimeout)
core/orchestrator.go:			ctx, cancel := context.WithTimeout(context.Background(), transcodeLoopTimeout)
core/orchestrator.go:	ctx, cancel := context.WithTimeout(context.Background(), RemoteTranscoderTimeout)
discovery/db_discovery.go:	ctx, cancel := context.WithTimeout(context.Background(), getOrchestratorsTimeoutLoop)
discovery/db_discovery.go:	ctx, cancel := context.WithTimeout(context.Background(), getOrchestratorsTimeoutLoop)
discovery/discovery.go:	ctx, cancel := context.WithTimeout(context.Background(), getOrchestratorsTimeoutLoop)
eth/blockwatch/client.go:	ctx, cancel := context.WithTimeout(context.Background(), rc.requestTimeout)
eth/blockwatch/client.go:	ctx, cancel := context.WithTimeout(context.Background(), rc.requestTimeout)
eth/blockwatch/client.go:	ctx, cancel := context.WithTimeout(context.Background(), rc.requestTimeout)
[...]

PR #1387, opened 2020-02-18
Reuse existing context.Context where available

mk-livepeer:
Pass context down the function hierarchy rather than abusing `context.Background()`

This is a "best practices / coding style" cleanup.  This compiles down to zero change in the resulting binary, except for the changes to `server/webserver.go`, which is optimized to now use context from the incoming http request rather than the global context retrieved from `context.Background()`.

This specific change has zero impact on any structures or function signatures.  Because of this, I see it as "low hanging fruit" that can be reviewed easily and separately from any other more drastic changes relating to context usage.

see #1332

### DIFF
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -163,6 +163,8 @@ func main() {
 		ethController string
 	}
 
+	ctx := context.Background()
+
 	configOptions := map[string]*NetworkConfig{
 		"rinkeby": {
 			ethUrl:        "https://rinkeby.infura.io/v3/09642b98164d43eb890939eb9a7ec500",
@@ -326,7 +328,7 @@ func main() {
 			return
 		}
 
-		chainID, err := backend.ChainID(context.Background())
+		chainID, err := backend.ChainID(ctx)
 		if err != nil {
 			glog.Errorf("failed to get chain ID from remote ethereum node: %v", err)
 			return
@@ -477,10 +479,10 @@ func main() {
 				return
 			}
 
-			ctx, cancel := context.WithCancel(context.Background())
+			orchSetupCtx, cancel := context.WithCancel(ctx)
 			defer cancel()
 
-			if err := setupOrchestrator(ctx, n, *initializeRound); err != nil {
+			if err := setupOrchestrator(orchSetupCtx, n, *initializeRound); err != nil {
 				glog.Errorf("Error setting up orchestrator: %v", err)
 				return
 			}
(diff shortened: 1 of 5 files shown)


cmd/livepeer/livepeer.go
```diff
@@ -726,7 +728,7 @@ func main() {
 	ec := make(chan error)
 	tc := make(chan struct{})
 	wc := make(chan struct{})
-	msCtx, cancel := context.WithCancel(context.Background())
+	msCtx, cancel := context.WithCancel(ctx)
```
kyriediculous, line 731: Can't we just declare the contexts with cancel once top-level?
Don't we currently redeclare these despite them being the same value?

kyriediculous (changes requested):

cmd/livepeer/livepeer.go
mkrufky: > Can't we just declare the contexts with cancel once top-level?

`msCtx` is used for special handling by the media server and should not be confused with the context it was forked from.

> Don't we currently redeclare these despite them being the same value?

`context.Background()` returns the global context.  It's like using global variables in an application, which we all know is a no-no, especially in cases of concurrency.  Right now it compiles down to the same binary with or without this cleanup, but code readability is a problem.  This patch aims to begin fixing the code readability of the context handling.

context should usually be the same within a function, and may or may not be forked into the functions called from it.  When a function needs specific cancellation handling, that is when we definitely want to fork a given context to a contextWithTimeout, contextWithCancel or contextWithDeadline so as not to affect the context that it was forked from.

For now, this is a step forward.  It begins to improve the readability of context handling, and it has no negative effects.

kyriediculous: > When a function needs specific cancellation handling, that is when we definitely want to fork a given context to a contextWithTimeout, contextWithCancel or contextWithDeadline so as not to affect the context that it was forked from.

Thanks for clarifying !

kyriediculous (approved):
Giving my approval, but perhaps you wanna wait for someone else's as well.

monitor/census.go
```diff
@@ -158,6 +158,7 @@ func InitCensus(nodeType, nodeID, version string) {
 		success:     make(map[uint64]*segmentsAverager),
 	}
 	var err error
+	ctx := context.Background()
```
yondonfu, line 161: This could be passed into `InitCensus`, but I'm guessing you created a background context here to minimize changes with the intent of adjusting context flow later on?

darkdarkdragon (approved):
LGTM 🚢

monitor/census.go
mkrufky: Correct.  This first pass does not alter any function prototypes.   The next one does.

PR #1392, opened 2020-02-21
Cap transcoding attempts.

### LINKED ISSUE #1057: Cap transcode retries.
Currently broadcaster tries to transcode segment indefinitely (it only stops if there is not orchestrators left in its list). This can lead to the situation when segment is transcoded, but window of live playlist is already passed, so that segment is not needed. 
This will make broadcaster pay for the work it is not needed and will skew results of our tests - transcode will be recorded as successful, but for broadcaster it is actually a failure to get result too late.

Probably we could step trying to transcode if there is (LIVE_LIST_LENGTH-1)*SegLen seconds passed since start of this segment transcoding attempts.

Something like this:
```golang
	start := time.Now()
	for {
		// if fails, retry; rudimentary
		if err := transcodeSegment(cxn, seg, name); err == nil {
			return nil
		}
		if time.Since(start) > (core.LIVE_LIST_LENGTH-1)*SegLen {
			return fmt.Errorf("Timed out")
		}
	}
```

j0sh:
Cap transcoding attempts.

Avoids thundering herd problems around repeated retries.

This will lower the success rate a bit, but is a more
realistic interpretation of satisfactory transcoding criteria.

The impact on the success rate is usually a few percent, < 5%.
When we see this type of impact, we're usually already below
99% anyway; for example going from ~95% -> ~92%. That
extra three percent would be for the segments that are retried
sometimes dozens of times. When that happens, it's also likely
that other segments are prone to be dropped, since a this point
it's a race between the O refresh and running out of Os.

The CLI flag added here is called `maxAttempts` . Originally this was `maxRetries` but this naming presented a few problems:

* When counting towards the cap, we count attempts, not just retries. So the [terminating condition](https://github.com/livepeer/go-livepeer/compare/ja/retrycap?expand=1#diff-cecc27bfe2b6b19181ce443733dfba57R294) would have to be `maxRetries+1` which felt a little unintuitive at first glance.

* Specifying attempts allows the user to configure to skip transcoding for whatever reason (eg, segmentation only), apart from specifying an empty set of orchestrators. This can be done with a value <= 0. Not sure how useful this is, but it "works" with this small semantic tweak. (A retry would mean a minimum of one transcode attempt.)

* Should we preserve the ability to do unlimited transcodes? This could probably be signaled with `-maxAttempts 0`, with the terminating condition modified accordingly. It's not here right now, but can add it in if folks feel that's important, since that's the behavior that we currently support.

- Adds a new `maxAttempts` CLI flag. Default value of 3.
- Caps transcode attempts based on `maxAttempts`.
- Adds unit tests
- Applies some automatic changes from gofmt https://github.com/livepeer/go-livepeer/commit/590e02601ea5a02565065fdf9a49edb61c058906

Unit tests, manual tests with multi-O

Fixes https://github.com/livepeer/go-livepeer/issues/1057

This is simpler than the algorithm described in #1057 ; this is only a counter and doesn't take elapsed time into account.

**Checklist:**

### DIFF
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -98,6 +98,7 @@ func main() {
 	broadcaster := flag.Bool("broadcaster", false, "Set to true to be a broadcaster")
 	orchSecret := flag.String("orchSecret", "", "Shared secret with the orchestrator as a standalone transcoder")
 	transcodingOptions := flag.String("transcodingOptions", "P240p30fps16x9,P360p30fps16x9", "Transcoding options for broadcast job")
+	maxAttempts := flag.Int("maxAttempts", 3, "Maximum transcode attempts")
 	maxSessions := flag.Int("maxSessions", 10, "Maximum number of concurrent transcoding sessions for Orchestrator, maximum number or RTMP streams for Broadcaster, or maximum capacity for transcoder")
 	currentManifest := flag.Bool("currentManifest", false, "Expose the currently active ManifestID as \"/stream/current.m3u8\"")
 	nvidia := flag.String("nvidia", "", "Comma-separated list of Nvidia GPU device IDs to use for transcoding")
@@ -700,6 +701,10 @@ func main() {
 		} else if *network != "offchain" {
 			server.Policy = &verification.Policy{Retries: 2}
 		}
+
+		// Set max transcode attempts. <=0 is OK; it just means "don't transcode"
+		server.MaxAttempts = *maxAttempts
+
 	} else if n.NodeType == core.OrchestratorNode {
 		suri, err := getServiceURI(n, *serviceAddr)
 		if err != nil {
--- a/server/broadcast.go
+++ b/server/broadcast.go
@@ -25,6 +25,7 @@ import (
 
 var Policy *verification.Policy
 var BroadcastCfg = &BroadcastConfig{}
+var MaxAttempts = 3
 
 type BroadcastConfig struct {
 	maxPrice *big.Rat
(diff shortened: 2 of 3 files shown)


j0sh:
Here's a snippet to check the number of transcoding attempts per segment. Run this on the B logs (here, `~/lp/bcast.log`)

```
grep Submitting ~/lp/bcast.log | grep segment_rpc | grep -o 'manifestID=.* seqNo=[0-9]\+' | sort | uniq -c | sort -nk 1 | awk '{print $1}' | uniq -c
```

cmd/livepeer/livepeer.go
```diff
@@ -98,6 +98,7 @@ func main() {
 	broadcaster := flag.Bool("broadcaster", false, "Set to true to be a broadcaster")
 	orchSecret := flag.String("orchSecret", "", "Shared secret with the orchestrator as a standalone transcoder")
 	transcodingOptions := flag.String("transcodingOptions", "P240p30fps16x9,P360p30fps16x9", "Transcoding options for broadcast job")
+	maxAttempts := flag.Int("maxAttempts", 3, "How many times to attempt a transcode")
```
mkrufky: Might "Maximum number of transcode attempts" or even just "Maximum transcode attempts" be more descriptive here?

mkrufky (changes requested):

darkdarkdragon (approved):
LGTM

cmd/livepeer/livepeer.go
j0sh: Updated to language to "Maximum transcode attempts" in 8e12347944a0460ee298acca2fbd02b1772915f1

mkrufky (approved):
:+1:

j0sh:
Rebased, merged.

iameli:
Thanks for quick turnaround on this, excited to run another big test next week :)

PR #1393, opened 2020-02-21
Download segment data only once for verification

### LINKED ISSUE #1382: B should download segment data once for both sig verification and pixel count verification
* `broadcast.verify()` should always download rendition data to populate the `Renditions` field of `verification.Params`

* `verification.countPixels()` should only take a `[]byte`.  It should always write a temp file and always use that file rather than the original URL.

mk-livepeer:
Broadcaster nodes should download segment data once for both sig verification and pixel count verification

closes #1382

### DIFF
--- a/server/broadcast.go
+++ b/server/broadcast.go
@@ -13,8 +13,6 @@ import (
 
 	"github.com/golang/glog"
 
-	"github.com/ethereum/go-ethereum/crypto"
-
 	"github.com/livepeer/go-livepeer/common"
 	"github.com/livepeer/go-livepeer/core"
 	"github.com/livepeer/go-livepeer/drivers"
@@ -33,6 +31,7 @@ var BroadcastCfg = &BroadcastConfig{}
 var MaxAttempts = 3
 
 var getOrchestratorInfoRPC = GetOrchestratorInfo
+var downloadSeg = drivers.GetSegmentData
 
 type BroadcastConfig struct {
 	maxPrice *big.Rat
@@ -399,11 +398,11 @@ func transcodeSegment(cxn *rtmpConnection, seg *stream.HLSSegment, name string,
 	}
 
 	var dlErr error
-	segHashes := make([][]byte, len(res.Segments))
+	segData := make([][]byte, len(res.Segments))
 	n := len(res.Segments)
 	segURLs := make([]string, len(res.Segments))
-	segHashLock := &sync.Mutex{}
-	cond := sync.NewCond(segHashLock)
+	segLock := &sync.Mutex{}
+	cond := sync.NewCond(segLock)
 
 	dlFunc := func(url string, pixels int64, i int) {
 		defer func() {
@@ -415,16 +414,27 @@ func transcodeSegment(cxn *rtmpConnection, seg *stream.HLSSegment, name string,
 			cond.L.Unlock()
 		}()
 
-		if bos := sess.BroadcasterOS; bos != nil && !drivers.IsOwnExternal(url) {
-			data, err := drivers.GetSegmentData(url)
+		bos := sess.BroadcasterOS
+
+		var data []byte
+		// Download segment data in the following cases:
+		// - A verification policy is set. The segment data is needed for signature verification and/or pixel count verification
+		// - The segment data needs to be uploaded to the broadcaster's own OS
+		if verifier != nil || (bos != nil && !drivers.IsOwnExternal(url)) {
+			d, err := downloadSeg(url)
 			if err != nil {
 				errFunc(monitor.SegmentTranscodeErrorDownload, url, err)
-				segHashLock.Lock()
+				segLock.Lock()
 				dlErr = err
-				segHashLock.Unlock()
+				segLock.Unlock()
 				cxn.sessManager.removeSession(sess)
 				return
 			}
+
+			data = d
+		}
+
+		if bos != nil && !drivers.IsOwnExternal(url) {
 			name := fmt.Sprintf("%s/%d.ts", sess.Profiles[i].Name, seg.SeqNo)
 			newURL, err := bos.SaveData(name, data)
 			if err != nil {
@@ -437,11 +447,6 @@ func transcodeSegment(cxn *rtmpConnection, seg *stream.HLSSegment, name string,
 				return
 			}
 			url = newURL
-
-			hash := crypto.Keccak256(data)
-			segHashLock.Lock()
-			segHashes[i] = hash
-			segHashLock.Unlock()
 		}
 
 		// Store URLs for the verifier. Be aware that the segment is
@@ -449,9 +454,10 @@ func transcodeSegment(cxn *rtmpConnection, seg *stream.HLSSegment, name string,
 		// external. If a client were to ignore the playlist and
 		// preemptively fetch segments, they could be reading tampered
 		// data. Not an issue if the delivery protocol is being obeyed.
-		segHashLock.Lock()
+		segLock.Lock()
 		segURLs[i] = url
-		segHashLock.Unlock()
+		segData[i] = data
+		segLock.Unlock()
 
 		if monitor.Enabled {
 			monitor.TranscodedSegmentAppeared(nonce, seg.SeqNo, sess.Profiles[i].Name)
--- a/server/broadcast.go
+++ b/server/broadcast.go
@@ -479,7 +479,7 @@ func transcodeSegment(cxn *rtmpConnection, seg *stream.HLSSegment, name string,
 
 	if verifier != nil {
 		// verify potentially can change content of segURLs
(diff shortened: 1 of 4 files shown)


verification/verify_test.go
```diff
@@ -300,9 +300,9 @@ func TestVerifyPixels(t *testing.T) {
 
 	// Test no writing temp file with incorrect pixels
 	err = verifyPixels("../server/test.flv", nil, 50)
-	assert.Equal(ErrPixelMismatch, err)
+	assert.EqualError(err, "Invalid data found when processing input")
 
 	// Test no writing temp file with correct pixels
 	err = verifyPixels("../server/test.flv", nil, p)
-	assert.Nil(err)
+	assert.EqualError(err, "Invalid data found when processing input")
```
yondonfu: Since `countPixels()` will return this error whenever nil data is passed and since the above test already handles this case, I don't think we need this test anymore.

verification/verify_test.go
```diff
@@ -300,9 +300,9 @@ func TestVerifyPixels(t *testing.T) {
 
 	// Test no writing temp file with incorrect pixels
 	err = verifyPixels("../server/test.flv", nil, 50)
-	assert.Equal(ErrPixelMismatch, err)
+	assert.EqualError(err, "Invalid data found when processing input")
```
yondonfu, line 345: We can update the comment for this test since `countPixels()` always writes a temp file. Maybe something like "Test nil data"

yondonfu (dismissed):
We'll need to make sure `params.Renditions` is populated as mentioned in #1382

verification/verify_test.go
mkrufky: done in 98113590b8478473e3b10cf74693555dba91edb0

verification/verify_test.go
mkrufky: done in e2942c57d4fba7fe78fde65be183a53668c38da8

server/broadcast.go
```diff
@@ -28,6 +25,8 @@ var Policy *verification.Policy
 var BroadcastCfg = &BroadcastConfig{}
 var MaxAttempts = 3
 
+var downloadSeg = drivers.GetSegmentData
```
yondonfu, line 34: I don't feel great about declaring this package level variable, but this made writing tests for this PR much easier because I was able to monkey patch the `drivers.GetSegmentData` implementation.

An alternative could be extract the [dlFunc](https://github.com/livepeer/go-livepeer/blob/b6e07fee00114427cd8b7638976a53fc49bfcc54/server/broadcast.go#L382) function which is defined inline in `transcodeSegment()` into a standalone function. Then, `dlFunc` could accept a function with the signature `func(url string) ([]byte, error)` as an argument - in practice, `drivers.GetSegmentData` would be passed in, but a stub function could be used in tests. This would've required a bit of refactoring so I opted not to take this approach right now, but could be worth considering outside of this PR.

server/broadcast.go
```diff
@@ -389,21 +388,30 @@ func transcodeSegment(cxn *rtmpConnection, seg *stream.HLSSegment, name string,
...
 
-		if bos := sess.BroadcasterOS; bos != nil && !drivers.IsOwnExternal(url) {
-			data, err := drivers.GetSegmentData(url)
+		bos := sess.BroadcasterOS
+
+		var data []byte
+		// Download segment data in the following cases:
+		// - A verification policy is set. The segment data is needed for signature verification and/or pixel count verification
+		// - The segment data needs to be uploaded to the broadcaster's own OS
+		if verifier != nil || (bos != nil && !drivers.IsOwnExternal(url)) {
```
yondonfu, line 423: NB. The broadcaster will unnecessarily download segment data in the following scenario:

- Off-chain mode
- Using a verification policy with a verifier that counts pixels on its own
- Using an external OS

Technically, the broadcaster does not need to download segment data here in that case. I chose not to handle that case right now as it requires the broadcaster to know if the policy's verifier will count pixels on its own.

server/broadcast.go
kyriediculous: There's a few other areas where we currently do this to make writing tests much easier. I feel this is okay for now but we should probably revisit if we can think of a pattern that doesn't require a huge refactor or add it to tech debt. 

other places where we monkey patch off the top off my head
- discovery (binding server.GetOrchestratorInfo and redefining it for discovery tests)
- testing payment param refresh by overwriting external `GetOrchestratorInfo` call (kinda similar to the above but within one package)

server/broadcast.go
kyriediculous: Does unnecessarily downloading the segment data impact the latency for the off-chain workflow in scope of the video user team and pilots we are doing? 

If it does it might be worthwhile to address but I have a hard time judging this from my perspective.

yondonfu: > Does unnecessarily downloading the segment data impact the latency for the off-chain workflow in scope of the video user team and pilots we are doing?

The unnecessary segment download should not impact the off-chain workflow because the off-chain workflow typically does not use a verification policy. The only time an off-chain B would use a verification policy is for internal verification specific testing. I'm inclined to leave this as-is for now since the inefficiency is isolated to verification specific testing scenarios and because additional verification policy changes such as livepeer/verification-classifier#64 may affect the choice of solution for resolving this inefficiency.

server/broadcast.go
```diff
@@ -488,35 +492,16 @@ func shouldStopSession(err error) bool {
...
-				return errors.New("Missing Local Data")
-			}
-			renditionData[i] = data
-		} else {
-			return errors.New("Expected local storage but did not have it")
-		}
+	renditionData := make([][]byte, len(segData))
+	for i := range segData {
+		renditionData[i] = make([]byte, len(segData[i]))
+		copy(renditionData[i], segData[i])
```
kyriediculous: Can't we just do `renditionData[i] = segData[i]` ?

server/broadcast_test.go
```diff
@@ -101,7 +101,7 @@ func (s *stubOSSession) SaveData(name string, data []byte) (string, error) {
 func (s *stubOSSession) EndSession() {
 }
 func (s *stubOSSession) GetInfo() *net.OSInfo {
-	return nil
+	return &net.OSInfo{}
```
kyriediculous: It seems like we don't need this change ?

server/broadcast_test.go
```diff
@@ -1033,3 +983,118 @@ func TestVerifier_HLSInsertion(t *testing.T) {
...
+	segData := []*net.TranscodedSegmentData{
+		{Url: url, Pixels: 100},
+	}
+
+	buf, err := proto.Marshal(&net.TranscodeResult{
+		Result: &net.TranscodeResult_Data{
+			Data: &net.TranscodeData{Segments: segData},
+		},
+	})
+	assert.Nil(t, err, fmt.Sprintf("Could not marshal results for %s", url))
```
kyriediculous: nit: shouldn't this be `require.Nil` ?

kyriediculous (changes requested):

server/broadcast_test.go
yondonfu: Yep - the assert statement was copy and pasted. Fixed in 75f9d11

server/broadcast_test.go
yondonfu: Ah yep - I think this was left over from before the introduction of the `downloadSeg` package variable. Now that we can stub `downloadSeg` in the tests, this change should no longer be required. Fixed in 75f9d11

server/broadcast.go
yondonfu: I originally copied the data in `segData[i]` because changes to `segData[i]` would result in changes to `params.Renditions[i]` since they would point to the same slice header. But, `segData` is not used anywhere in `server/broadcast.go` besides in `verify()` so it should be safe to set `params.Renditions` to `segData`. In 75f9d11, I updated a few tests to allocate a new slice for `segData` whenever passing it as an argument to `verify()` (otherwise, some tests would fail because mutations to the same slice for `segData` would mutate the values for `params.Renditions`).

kyriediculous: Ahh I see, I would be okay with not mutating as a good practice here as well. I'll leave it up to you.

kyriediculous:
Changes look good. I'll leave that one decision up to you whether you want to drop 75f9d11 or not. 

Can we rebase and resolve conflicts?

server/broadcast.go
yondonfu: Going to keep this change as avoiding the copy of segment data (which can be largeish) is nice.

kyriediculous: Ah yes didn't think about the size, good point.

yondonfu:
Rebased.

kyriediculous (approved):
LGTM 👍

PR #1394, opened 2020-02-21
cmd, common, eth, server: show price per pixel in CLI

kyriediculous:
Show price per pixel in CLI when picking `3. Show registered orchestrators`.

The price per pixel is queried from the database so it actually includes the transaction cost overhead of 1%. 

Checked that the pricing of my mainnet orch is shown correctly (45+45/100)

**Checklist:**

### DIFF
--- a/cmd/livepeer_cli/wizard_bond.go
+++ b/cmd/livepeer_cli/wizard_bond.go
@@ -34,7 +34,7 @@ func (w *wizard) registeredOrchestratorStats() map[int]common.Address {
 	fmt.Println("+------------------------+")
 
 	table := tablewriter.NewWriter(os.Stdout)
-	table.SetHeader([]string{"ID", "Address", "Active", "Delegated Stake", "Reward Cut (%)", "Fee Share (%)", "Service URI"})
+	table.SetHeader([]string{"ID", "Address", "Active", "Delegated Stake", "Reward Cut (%)", "Fee Share (%)", "Service URI", "Price Per Pixel"})
 
 	for _, t := range orchestrators {
 		table.Append([]string{
@@ -45,6 +45,7 @@ func (w *wizard) registeredOrchestratorStats() map[int]common.Address {
 			eth.FormatPerc(t.RewardCut),
 			eth.FormatPerc(t.FeeShare),
 			t.ServiceURI,
+			t.PricePerPixel.FloatString(3),
 		})
 
 		orchestratorIDs[nextId] = t.Address
--- a/common/util.go
+++ b/common/util.go
@@ -26,6 +26,10 @@ const HTTPTimeout = 8 * time.Second
 
 const maxInt64 = int64(math.MaxInt64)
 
+// using a scaleFactor of 1000 for orchestrator prices
+// resulting in max decimal places of 3
+const priceScalingFactor = int64(1000)
+
 var (
 	ErrParseBigInt = fmt.Errorf("failed to parse big integer")
 	ErrProfile     = fmt.Errorf("failed to parse profile")
@@ -207,7 +211,12 @@ func GenErrRegex(errStrings []string) *regexp.Regexp {
 // PriceToFixed converts a big.Rat into a fixed point number represented as int64
 // using a scaleFactor of 1000 resulting in max decimal places of 3
 func PriceToFixed(price *big.Rat) (int64, error) {
-	return ratToFixed(price, 1000)
+	return ratToFixed(price, priceScalingFactor)
+}
+
+// FixedToPrice converts an fixed point number with 3 decimal places represented as in int64 into a big.Rat
+func FixedToPrice(price int64) *big.Rat {
+	return big.NewRat(price, priceScalingFactor)
 }
 
 // BaseTokenAmountToFixed converts the base amount of a token (i.e. ETH/LPT) represented as a big.Int into a fixed point number represented
--- a/server/cliserver_test.go
+++ b/server/cliserver_test.go
@@ -1,8 +1,11 @@
 package server
 
 import (
+	"encoding/json"
+	"errors"
 	"fmt"
 	"io/ioutil"
+	"math/big"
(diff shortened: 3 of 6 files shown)


cmd/livepeer_cli/wizard_bond.go
```diff
@@ -34,7 +34,7 @@ func (w *wizard) registeredOrchestratorStats() map[int]common.Address {
 	fmt.Println("+------------------------+")
 
 	table := tablewriter.NewWriter(os.Stdout)
-	table.SetHeader([]string{"ID", "Address", "Active", "Delegated Stake", "Reward Cut (%)", "Fee Share (%)", "Service URI"})
+	table.SetHeader([]string{"ID", "Address", "Active", "Delegated Stake", "Reward Cut (%)", "Fee Share (%)", "Service URI", "Price per pixel"})
```
yondonfu: "Price Per Pixel" to match the capitalization convention in the other column names?

common/util.go
```diff
@@ -226,6 +226,10 @@ func BaseTokenAmountToFixed(baseAmount *big.Int) (int64, error) {
 	return ratToFixed(rat, 100000)
 }
 
+func FixedToPrice(price, scalingFactor int64) *big.Rat {
```
yondonfu: Since this function is used specifically for converting fixed point numbers to prices (which have a hardcoded scaling factor of 1000 right now) can we omit the `scalingFactor` param and hardcode the scaling factor value in the function body similar to what we do in `PriceToFixed`?

IMO it would also be nice to place this function next to `PriceToFixed()` since the two functions are the inverse of each other.

server/webserver.go
```diff
@@ -912,6 +912,16 @@ func (s *LivepeerServer) cliWebServerHandlers(bindAddr string) *http.ServeMux {
 				return
 			}
 
+			for _, o := range orchestrators {
+				dbO, err := s.LivepeerNode.Database.SelectOrchs(&lpcommon.DBOrchFilter{
+					Addresses: []common.Address{o.Address},
+				})
+				if err != nil {
+					continue
```
yondonfu, line 951: Let's log the error.

server/webserver.go
```diff
@@ -912,6 +912,16 @@ func (s *LivepeerServer) cliWebServerHandlers(bindAddr string) *http.ServeMux {
 				return
 			}
 
+			for _, o := range orchestrators {
+				dbO, err := s.LivepeerNode.Database.SelectOrchs(&lpcommon.DBOrchFilter{
```
yondonfu, line 942: Might be worth adding a non-zero length check for the returned slice just to be safe. The length should usually be non-zero, but I think there could be an edge case where this endpoint is hit after a new address is added to the on-chain pool, but before the address is added to the DB by the orchestrator watcher.

yondonfu (changes requested):

server/webserver.go
kyriediculous: good point, if `len(dbO) == 0` we should set o.PricePerPixel to `big.NewRat(0, 1)` otherwise the floatstring conversion in the CLI will fail.

server/webserver.go
kyriediculous: fixed in 6702d485

server/webserver.go
kyriediculous: good point, if `len(dbO) == 0` we should set o.PricePerPixel to `big.NewRat(0, 1)` otherwise the floatstring conversion in the CLI will fail. 

Fixed in eac1d3b9

cmd/livepeer_cli/wizard_bond.go
kyriediculous: fixed in d74e79c68ae186fe7c2bcf3c3718296d0a7676b4

kyriediculous: fixed in d74e79c68ae186fe7c2bcf3c3718296d0a7676b4

common/util.go
kyriediculous: fixed in 926eac300b23364c9a1ed48df920837255d75838

moved function in subsequent fixup 80e013ac

kyriediculous: fixed in 926eac300b23364c9a1ed48df920837255d75838

moved function in subsequent fixup 80e013ac

server/webserver.go
```diff
@@ -912,6 +912,21 @@ func (s *LivepeerServer) cliWebServerHandlers(bindAddr string) *http.ServeMux {
 				return
 			}
 
+			for _, o := range orchestrators {
```
yondonfu, line 941: Can we add a test for this update (`cliserver_test.go` might be a good place for the test to go)?

yondonfu (changes requested):

server/cliserver_test.go
```diff
@@ -86,3 +92,63 @@ func TestGetDelegatorInfo(t *testing.T) {
...
+	}
+
+	n, _ := core.NewLivepeerNode(eth, "./tmp", dbh)
+
+	s := NewLivepeerServer("127.0.0.1:1938", n)
+	mux := s.cliWebServerHandlers("addr")
+	srv := httptest.NewServer(mux)
+	defer srv.Close()
+
+	res, err := http.Get(fmt.Sprintf("%s/registeredOrchestrators", srv.URL))
```
yondonfu, line 148: Let's add a test for when `SelectOrchs()` returns an empty slice resulting in the orchestrator's default `PricePerPixel` being set to `big.NewRat(0, 1)`?

yondonfu (changes requested):

yondonfu (commented):
Changes LGTM - let's squash the fixup commits.

kyriediculous:
I rebased but CI still has issues, should be g2g though.

kyriediculous:
CI is green , just need an approval !

yondonfu (approved):
LGTM! 🚢

PR #1400, opened 2020-02-25
discovery: Shuffle orchestrators prior to OrchestratorInfo

j0sh:

Shuffle orchestrators each time GetOrchestratorInfo is invoked, rather than just once at the beginning. This improves success rates by a few percent if the O list is static.

- Randomize orchestrators prior to GetOrchestratorInfo
- Tests around randomization
- Gofmt / lint fixes for tests
- More test cleanup, removing the `perm` helper which is no longer needed. Other tests don't rely on randomization and the uncertainty involved with goroutine scheduling around GetOrchestratorInfo (one goroutine per request) means there isn't a need to make the `perm` reproducible.

Manual testing, unit tests.

**Checklist:**

### DIFF
--- a/discovery/discovery_test.go
+++ b/discovery/discovery_test.go
@@ -431,7 +431,6 @@ func TestNewDBOrchestorPoolCache_PollOrchestratorInfo(t *testing.T) {
 
 func TestNewOrchestratorPoolCache_GivenListOfOrchs_CreatesPoolCacheCorrectly(t *testing.T) {
 	addresses := stringsToURIs([]string{"https://127.0.0.1:8936", "https://127.0.0.1:8937", "https://127.0.0.1:8938"})
-	expected := stringsToURIs([]string{"https://127.0.0.1:8938", "https://127.0.0.1:8937", "https://127.0.0.1:8936"})
 	assert := assert.New(t)
 
 	// creating NewOrchestratorPool with orch addresses
@@ -441,7 +440,7 @@ func TestNewOrchestratorPoolCache_GivenListOfOrchs_CreatesPoolCacheCorrectly(t *
 	offchainOrch := NewOrchestratorPool(nil, addresses)
 
 	for i, uri := range offchainOrch.uris {
-		assert.Equal(uri, expected[i])
+		assert.Equal(uri, addresses[i])
 	}
 }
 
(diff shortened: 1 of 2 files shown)


discovery/discovery_test.go
```diff
@@ -1106,5 +1090,67 @@ func TestEthOrchToDBOrch(t *testing.T) {
...
 	assert.Equal(dbo.ActivationRound, o.ActivationRound.Int64())
-	assert.Equal(dbo.DeactivationRound,  int64(math.MaxInt64))
+	assert.Equal(dbo.DeactivationRound, int64(math.MaxInt64))
+}
+
+func TestOrchestratorPool_ShuffleGetOrchestrators(t *testing.T) {
+	assert := assert.New(t)
+
+	addresses := stringsToURIs([]string{"https://127.0.0.1:8936", "https://127.0.0.1:8937", "https://127.0.0.1:8938"})
+	//expected := stringsToURIs([]string{"https://127.0.0.1:8938", "https://127.0.0.1:8937", "https://127.0.0.1:8936"})
```
yondonfu: Can remove?

discovery/discovery_test.go
```diff
@@ -1106,5 +1090,67 @@ func TestEthOrchToDBOrch(t *testing.T) {
...
+	defer func() { serverGetOrchInfo = oldOrchInfo }()
+	serverGetOrchInfo = func(ctx context.Context, bcast common.Broadcaster, server *url.URL) (*net.OrchestratorInfo, error) {
+		ch <- server
+		return nil, nil
+	}
+
+	pool := NewOrchestratorPool(nil, addresses)
+
+	// Check that randomization happens: check for elements in a different order
+	// Sometimes this could fail due to scheduling randomization
```
yondonfu: This is because even if the original list is shuffled we don't have tight control over the order in which the `GetOrchestrator()` responses are received in the channel? So, we might know the order of the original list given a specific speed, but we do not know the order of the responses with certainty?

yondonfu (changes requested):
Looks good - left one comment about a stray code comment in the tests and one clarifying question.

> More test cleanup, removing the perm helper which is no longer needed.

A nice side effect of this is that we can run `go test -v -failfast -count=N` in the `discovery` package now woohoo (I believe this didn't work previously because the `perm` package variable was not being reset properly after certain tests). Might come in handy if we're ever trying to hunt down flaky tests...

discovery/discovery_test.go
j0sh: Fixed in 09b658c502b09b14f38e6150dd7d2cf35a41cd22

discovery/discovery_test.go
j0sh: Correct - more specifically, it's because we don't control the order of execution of the GetOrchestratorInfo calls:

https://github.com/livepeer/go-livepeer/blob/cc7ab9d78680569de817927e250f7c9140163c01/discovery/discovery.go#L83-L85

Updated the language a bit in 09b658c502b09b14f38e6150dd7d2cf35a41cd22 ; hopefully it's clearer now (but happy to make further adjustments)

darkdarkdragon (approved):
LGTM

yondonfu (commented):
LGTM after a rebase/squash

j0sh:
Rebased

darkdarkdragon (approved):
LGTM

yondonfu (approved):
LGTM

PR #1402, opened 2020-02-25
Measure duration of auth webhook requests

darkdarkdragon:
Measures duration of auth webhook requests
and expose as a metric

Added new metric

Manually

**Checklist:**

### DIFF
--- a/monitor/census.go
+++ b/monitor/census.go
@@ -98,6 +98,7 @@ type (
 		mTranscodeLatency             *stats.Float64Measure
 		mTranscodeOverallLatency      *stats.Float64Measure
 		mUploadTime                   *stats.Float64Measure
+		mAuthWebhookTime              *stats.Float64Measure
 
 		// Metrics for GPUs
 		mGPUBacklog *stats.Int64Measure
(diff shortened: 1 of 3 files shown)


monitor/census.go
```diff
@@ -825,11 +834,17 @@ func SegmentUploaded(nonce, seqNo uint64, uploadDur time.Duration) {
...
-	defer cen.lock.Unlock()
 	stats.Record(cen.ctx, cen.mSegmentUploaded.M(1), cen.mUploadTime.M(float64(uploadDur/time.Second)))
 }
 
+func AuthWebhookFinished(dur time.Duration) {
+	census.authWebhookFinished(dur)
+}
+
+func (cen *censusMetricsCounter) authWebhookFinished(dur time.Duration) {
+	stats.Record(cen.ctx, cen.mAuthWebhookTime.M(dur.Seconds()))
```
j0sh: Is there an advantage to using seconds as opposed to milliseconds? Theoretically *most* of these values should be in the millisecond range, which might be nicer with fewer non-floaty numbers...

server/broadcast.go
```diff
@@ -8,6 +8,7 @@ import (
 	"net/url"
 	"strings"
 	"sync"
+	"time"
```
j0sh, line 11: For future reference, the changes in this file could be put into its own commit. Not mandatory to be fixed up for this PR, but it's generally good Git practice to keep unrelated changes separate. (I'm OK with bundling commits that are only semi-related in the same PR.)

j0sh (approved):
A couple small comments / questions that I'll leave to you whether to fix up. LGTM otherwise.

PR #1403, opened 2020-02-25
discovery: Avoid shared state in GetOrchestratorInfo responses

j0sh:
Follows conventions used elsewhere in the codebase, in addition to
being more idiomatic golang.

See here for a similar pattern:

https://github.com/livepeer/go-livepeer/blob/56a54051e13889fdd8bfb73011598f631fa4c9f1/discovery/db_discovery.go#L174-L207

 Refactors how we collect responses from the GetOrchestratorInfo query to avoid shared state. Use channels instead.

Aside from logging tweaks, no functional changes.

Unit testing, manual testing

**Checklist:**

### DIFF
--- a/discovery/discovery.go
+++ b/discovery/discovery.go
@@ -15,7 +15,7 @@ import (
 	"github.com/golang/glog"
 )
 
-const getOrchestratorsTimeoutLoop = 3 * time.Second
+var getOrchestratorsTimeoutLoop = 3 * time.Second
 
 var serverGetOrchInfo = server.GetOrchestratorInfo
 
--- a/discovery/discovery.go
+++ b/discovery/discovery.go
@@ -5,7 +5,6 @@ import (
 	"math"
 	"math/rand"
 	"net/url"
-	"sync"
 	"time"
 
 	"github.com/livepeer/go-livepeer/common"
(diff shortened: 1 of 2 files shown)


discovery/discovery.go
```diff
@@ -49,27 +49,23 @@ func (o *orchestratorPool) GetOrchestrators(numOrchestrators int) ([]*net.Orches
...
+			infoCh <- info
+			return
 		}
 		if err != nil && monitor.Enabled {
 			monitor.LogDiscoveryError(err.Error())
 		}
-		if numSuccessResp >= numOrchestrators || numResp >= len(o.uris) {
-			orchChan <- struct{}{}
+		if err == nil {
+			err = defaultErr
```
yondonfu: Seems like we might not need `defaultErr` here because:

1. The predicate function [logs an error](https://github.com/livepeer/go-livepeer/blob/56a54051e13889fdd8bfb73011598f631fa4c9f1/discovery/db_discovery.go#L102) when it fails.

2. We are not logging the error received on `errCh`. Logging the error here would add clutter right now since each error would be logged twice since [GetOrchestratorInfo()](https://github.com/livepeer/go-livepeer/blob/56a54051e13889fdd8bfb73011598f631fa4c9f1/server/rpc.go#L215) logs the error before returning it.

Worth considering updating the predicate function and `GetOrchestratorInfo()` to only return the error and let the caller handle logging if needed - out of scope for this PR, but just mentioning it.

discovery/discovery.go
```diff
@@ -82,28 +78,24 @@ func (o *orchestratorPool) GetOrchestrators(numOrchestrators int) ([]*net.Orches
...
-			numOrchestrators = len(orchInfos)
-		}
-		returnOrchs := orchInfos[:numOrchestrators]
-		respLock.Unlock()
-		glog.Info("Done fetching orch info for orchestrators, numResponses fetched: ", len(returnOrchs))
-		cancel()
-		return returnOrchs, nil
 	}
+	cancel()
+	glog.Infof("Done fetching orch info nb=%d responses=%d/%d timeout=%t",
```
yondonfu: `numOrchs` might be a clearer key name than `nb` in the message.

darkdarkdragon (approved):
LGTM

discovery/discovery.go
j0sh: Fixed in https://github.com/livepeer/go-livepeer/pull/1403/commits/dc7828a8a4befdd2a90a6b66498687f404901dec

discovery/discovery.go
j0sh: Good call - this was actually here because I wasn't sure that we could return `nil` from the error channel. I noticed that we didn't actually have tests for this condition, so added some more in https://github.com/livepeer/go-livepeer/pull/1403/commits/dc7828a8a4befdd2a90a6b66498687f404901dec

j0sh:
The timeout behavior was a little tricky to get right so added some tests for it. https://github.com/livepeer/go-livepeer/pull/1403/commits/a6a7eb5ae5f60199f83c39f11c5de896fd83d1b7

Note that how we handle timeouts in [1] has a minor bug: the `break` applies to the innermost `select` (a weird design choice by golang IMO), and what happens is we may actually end up looping and hitting `case <-ctx.Done()`  a few times until breaking out of the for-loop.

[1] https://github.com/livepeer/go-livepeer/blob/56a54051e13889fdd8bfb73011598f631fa4c9f1/discovery/db_discovery.go#L174-L207

yondonfu:
> Note that how we handle timeouts in [1] has a minor bug: the break applies to the innermost select (a weird design choice by golang IMO), and what happens is we may actually end up looping and hitting case <-ctx.Done() a few times until breaking out of the for-loop.

Ah good catch. Tracking in #1407

yondonfu (approved):
LGTM after squashing the fixup commit!

j0sh:
Rebased and squashed

PR #1404, opened 2020-02-26
Nv/roundtrip latencyscore

### LINKED ISSUE #1353: Flaky test: TestSubmitSegment_Success
This test failed once but passed after repeating it multiple times.
```
--- FAIL: TestSubmitSegment_Success (0.02s)
    segment_rpc_test.go:1334: 
        	Error Trace:	segment_rpc_test.go:1334
        	Error:      	"7.667e-06" is not less than "5.7034e-06"
        	Test:       	TestSubmitSegment_Success
```
This can cause CI to become unpredictable.  It isn't an urgent problem right now - just creating an issue so that it can be tracked.

### LINKED ISSUE #1376: Calculate latency score based on round trip latency instead of transcode duration

The latency score used for O selection is currently calculated as [transcode duration / segment duration](https://github.com/livepeer/go-livepeer/blob/70691f5944369d3ad2ee6176ac2a9eacef2eba0e/server/segment_rpc.go#L467). This latency score formulation helps B route work to the Os with the fastest transcoding times (includes the time to transcode as well as the time to respond with the result). However, an O with fast transcoding times may not be the best option for B if upload times are slow. The longer it takes to upload, the longer it takes for B to receive a result that could be used in its playlist. 

Slower upload times may result from B's architecture/config or O's architecture/config. Either way, they increase the time until results become available in the playlist. So, B should optimize for round trip latency for a result instead of only optimizing for transcode duration.

Instead of calculating the latency score based on transcode duration we can calculate the latency based on round trip latency which includes both the upload *and* transcode duration.

kyriediculous:
- Use round trip latency to calculate latency score
- Fix flaky `SubmitSegment` test where latency score tests sometimes failed

- use `tookAllDur` to calculate latency score
- add a small sleep to the stub `/segment` request to simulate network latency, increase the segment

Fixes #1376 
Fixes #1353 

**Checklist:**

### DIFF
--- a/server/segment_rpc.go
+++ b/server/segment_rpc.go
@@ -464,7 +464,7 @@ func SubmitSegment(sess *BroadcastSession, seg *stream.HLSSegment, nonce uint64)
 	return &ReceivedTranscodeResult{
 		TranscodeData: tdata,
 		Info:          tr.Info,
-		LatencyScore:  transcodeDur.Seconds() / seg.Duration,
+		LatencyScore:  tookAllDur.Seconds() / seg.Duration,
 	}, nil
 }
 
--- a/server/segment_rpc_test.go
+++ b/server/segment_rpc_test.go
@@ -13,6 +13,7 @@ import (
 	"net/url"
 	"strings"
 	"testing"
+	"time"
 
 	ethcommon "github.com/ethereum/go-ethereum/common"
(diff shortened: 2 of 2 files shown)


yondonfu (approved):
LGTM! 🚢

PR #1411, opened 2020-03-04
Log length of body of HTTP push request

darkdarkdragon:
Log length of body of HTTP push request

One line log update, no functional changes

**Checklist:**


iameli (approved):
LGTM!

Should we look into rejecting zero-length segments on the ingest side? Minimally, less than 188 bytes, we can be confident is not valid MPEG-TS.

darkdarkdragon:
> Should we look into rejecting zero-length segments on the ingest side? Minimally, less than 188 bytes, we can be confident is not valid MPEG-TS.

I'm all in for that.

PR #1412, opened 2020-03-09
misc fixes: ticket queue error logging and formatting hex values

yondonfu:

This PR includes two fixes:

- Only log non-nil block subscription errors in `pm/queue.go`
- Format hex values properly in log statements in `SenderMonitor.QueueTicket()`.

- d9fb95f: Adds a nil-check for block subscription errors in `pm/queue.go` and also updates the error message to indicate that it is a block subscription error
- 07886b9: Updates the log message in `SenderMonitor.QueueTicket()` to use `%v` instead of `%x` for already encoded hex values

Ran a local on-chain setup using the devtool.

N/A

**Checklist:**

### DIFF
--- a/pm/queue.go
+++ b/pm/queue.go
@@ -113,7 +113,9 @@ func (q *ticketQueue) startQueueLoop() {
 	for {
 		select {
 		case err := <-sub.Err():
-			glog.Error(err)
+			if err != nil {
+				glog.Errorf("Block subscription error err=%v", err)
+			}
 		case latestBlock := <-blockNums:
 			numTickets := q.Length()
 			for i := 0; i < int(numTickets); i++ {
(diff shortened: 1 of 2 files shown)


pm/queue.go
```diff
@@ -113,7 +113,9 @@ func (q *ticketQueue) startQueueLoop() {
 	for {
 		select {
 		case err := <-sub.Err():
-			glog.Error(err)
+			if err != nil {
```
kyriediculous, line 116: Can error ever be nil here though because we are only assigning upon receiving a value ?

yondonfu: A nil value will be received on `sub.Err()` when the subscription is closed (see [this comment](https://github.com/ethereum/go-ethereum/blob/18213281627b310ac68d22bca7c401ab5f67e1da/event/subscription.go#L35)). This is log output without the nil check:

Screen Shot 2020-03-09 at 11 58 11 AM

The subscriptions for each queue are closed and we log the nil value received on the error each channel for each of those subscriptions.

kyriediculous (approved):
LGTM !

PR #1424, opened 2020-03-15
common, discovery, server: exclude orchestrators that return transcode errors or timeout

### LINKED ISSUE #1375: Exclude Os that timeout from session refresh for a period of time

B may encounter a HTTP timeout while waiting for a response from O. However, O may still respond quickly to B's `GetOrchestrator()` call. As a result, O could be consistently re-added to B's working set during broadcast session refreshes despite always timing out.

Furthermore, if O is still transcoding B's segment and B submits another segment to O, then B will immediately encounter an `OrchestratorBusy` error. The following example illustrates this problem:

1. B submits segment 1 to O1
2. B encounters an HTTP timeout
3. B starts a broadcast session refresh
4. O is added back to B's working set because it responds quickly to `GetOrchestrator()`
5. B submits segment 1 to O1
6. O1 is still working on segment 1 from step 1 so it returns `OrchestratorBusy`

If B times out with an O, exclude that O from being re-added to the working set for a period of time in order to allow other Os to be added to the working set.

kyriediculous:

This PR implements an orchestrator suspension list which temporarily suspends orchestrators from being included in a session refresh by _any_ active or newly started broadcast session if they timeout, are not reachable or return a transcode error (`OrchestratorBusy` or  `OrchestratorCapped`). 

- Created new `suspensionList` and `priorityQueue` types
- Expanded `OrchestratorPool` API with a `SuspendOrchestrator(orch string)` method 
- Make use of `shouldStopSession(error)` to determine if we should suspend an orchestrator and add it to `suspensionList`
- When calling `OrchestratorPool.GetOrchestrators()` withhold responses from suspended orchestrators and add them to `priorityQueue` if we don't get sufficient responses from non-suspended orchestrators fill up the list with responses from suspended orchestrators until the list is full or we ran out of responses. The priority is decided by the time the orchestrator was suspended at, earlier is higher priority. 

Added / adjusted unit tests

Fixes #1375 

**Checklist:**

### DIFF
--- a/common/types.go
+++ b/common/types.go
@@ -17,6 +17,7 @@ type OrchestratorPool interface {
 	GetURLs() []*url.URL
 	GetOrchestrators(int) ([]*net.OrchestratorInfo, error)
 	Size() int
+	SuspendOrchestrator(orch string)
 }
 
 type OrchestratorStore interface {
--- a/discovery/db_discovery.go
+++ b/discovery/db_discovery.go
@@ -35,6 +35,7 @@ type DBOrchestratorPoolCache struct {
 	ticketParamsValidator ticketParamsValidator
 	rm                    common.RoundsManager
 	bcast                 common.Broadcaster
+	suspensions           *suspensionList
 }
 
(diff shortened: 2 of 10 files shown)


kyriediculous:
closed in favor of #1435

PR #1428, opened 2020-03-19
core,net,pm,server: O sends expected TicketExpirationParams to B along with TicketParams

### LINKED ISSUE #1187: Invalid ticket creation round when starting a broadcaster after being offline for awhile
**Describe the bug**

When starting a broadcaster up after being offline for awhile, the broadcaster encountered consecutive payment related errors when submitting segments to an orchestrator. The orchestrator logged the following error repeatedly:

```
Error receiving ticket manifestID=f recipientRandHash=5638662d577dbb78d15bf3698feee72a42b3034c1fdc3d26d9e942de88790791 senderNonce=1: invalid ticket creation round
```

It is possible that the broadcaster and orchestrator's views of the current round were different for some reason.

**Screenshots**

Orchestrator logs:
Screen Shot 2019-11-11 at 3 05 05 PM

Broadcaster logs:
Screen Shot 2019-11-11 at 3 05 27 PM

kyriediculous:
The orchestrator sends along TicketExpirationParams as ExpectedExpirationParams with TicketParams. The broadcaster then uses these values for the ticket CreationRound  and AuxData instead of relying on the TimeWatcher (view of the current chain for B). Both CreationRound  and CreationRoundBlockHash are added to the HMAC scheme to validate these values were advertised by the Orchestrator. 

Two considerable benefits:
- It removes the synchronization requirement of the Ethereum network between O and B for payment creation. This was a dependency I was quite uncomfortable with. Instead it is now the responsability of the Orchestrator to provide correct values that result in redeemable tickets.
- We no longer encounter unwanted "invalid ticket creation round" errors during round transitions that might interrupt transcoding

Downsides:
- Increase the size of the OrchestratorInfo message with a couple of bytes 
- Likely due to increase of size of the message for the HMAC generation this will become marginally slower (not sure here, not that familiar with the scheme but seems like a logical assumption). 

- Embedded `TicketExpirationParams` in `TicketParams` protobuf message
- Add `pm.TicketExpirationParams` field to `pm.TicketParams`
- `Recipient` reads last initialized round and its blockhash from memory when `recipient.TicketParams()` is called and the values are used for the `ExpectedExpirationParams` to send to B
- Add `CreationRound` and `CreationRoundBlockHash` to `recipientRand` generation and thus ticket validation
- Remove `validateCreationRound()` and its usage in ticket validation
- B uses the `ExpectedExpirationParams` provided in the `TicketParams` from O to populate the values for `CreationRound` and `CreationRoundBlockHash` when creating tickets

Added unit tests
Ran nodes with devtool

Fixes #1187 

**Checklist:**


pm/recipient.go
```diff
@@ -222,17 +222,23 @@ func (r *recipient) TicketParams(sender ethcommon.Address, price *big.Rat) (*Tic
 
 	winProb := r.winProb(faceValue)
 
-	recipientRand := r.rand(seed, sender, faceValue, winProb, expirationBlock, price)
+	ticketExpirationparams := &TicketExpirationParams{
```
yondonfu: "ticketExpirationparams" -> "ticketExpirationParams"

pm/recipient.go
```diff
@@ -387,14 +393,22 @@ func (r *recipient) redeemWinningTicket(ticket *Ticket, sig []byte, recipientRan
...
 	msg := append(seed.Bytes(), sender.Bytes()...)
 	msg = append(msg, faceValue.Bytes()...)
 	msg = append(msg, winProb.Bytes()...)
 	msg = append(msg, expirationBlock.Bytes()...)
 	msg = append(msg, price.Num().Bytes()...)
 	msg = append(msg, price.Denom().Bytes()...)
+	msg = append(
+		msg,
+		append(
+			ethcommon.LeftPadBytes(big.NewInt(ticketExpirationParams.CreationRound).Bytes(), uint256Size),
```
yondonfu: Since there is already an `AuxData()` method defined for `Ticket`, maybe we can shuffle some code around and reuse that logic here. Something like:

```golang
func (t *TicketExpirationParams) AuxData() []byte {
    // Current implementation of Ticket.AuxData()
}

func (t *Ticket) AuxData() []byte {
    return &TicketExpirationParams{
        CreationRound: t.CreationRound,
        CreationRoundBlockHash: t.CreationRoundBlockHash,
    }.AuxData()
}
```

Then, we can simplify this line to:

```golang
msg = append(msg, ticketExpirationParams.AuxData())
```

pm/sender.go
```diff
@@ -108,10 +108,16 @@ func (s *sender) CreateTicketBatch(sessionID string, size int) (*TicketBatch, er
 		return nil, err
 	}
 
-	expirationParams := s.expirationParams()
+	ticketParams := &session.ticketParams
+
+	expirationParams := ticketParams.ExpectedExpirationParams
+	// ensure backwards compatbility
```
yondonfu: Can we elaborate on this comment and explain why this is backwards compatible?

pm/ticket.go
```diff
@@ -45,6 +45,8 @@ type TicketParams struct {
 	ExpirationBlock *big.Int
 
 	PricePerPixel *big.Rat
+
+	ExpectedExpirationParams *TicketExpirationParams
```
yondonfu: Can the name of this field just be `ExpirationParams`?

yondonfu (changes requested):
While the [pmTicketParams()](https://github.com/livepeer/go-livepeer/blob/ce819a15a616277a7081b5cb91339c4121a22538/discovery/db_discovery.go#L325) helper in `db_discovery.go` doesn't *need* to populate the expiration params field since those fields are not accessed within the file at the moment, it might be a good idea to populate the field anyways for consistency.

yondonfu:
Some thoughts on implications of this change below:

Since B always uses the creation round (and block hash - when I say creation round in this comment I also refer to block hash associated with the creation round) that it receives, O is able to set the creation round in the range `[currentRound - 2, currentRound]` where 2 is the current on-chain ticket validity period.

Why would O set the creation round to anything less than the current round? One possible reason is to control the fee pool that fees are sent to. Recall that an O has a fee pool for each round and fees sent to a fee pool for round N are distributed to delegators that were staked to O in round N. In this case, O can choose to send fees to the fee pool for any round in the range `[currentRound - 2, currentRound]`. O *could* benefit from this ability if it notices that delegators have already claimed their fees for one of rounds in this range - additional fees sent to the pool at this point would be distributed amongst delegators (including the O itself) that have not yet claimed from those pools.

I don't think this is a big problem in practice because:

- Delegators typically do not need to claim earnings unless they have > 100 rounds of earnings to claim and they need to perform a bonding action. So, if delegators claim earnings infrequently and only when absolutely needed they will reduce the likelihood of "leaving fees on the table".
- O controls the creation round, but does not control when tickets will win so it is difficult to opportunistically send fees to fee pools for past rounds based on observed delegator behavior.

kyriediculous:
Good points you raise, I didn't consider those. 

But I think it would also require the O to not call reward itself then, otherwise it would have already called it for that EarningsPool, which would be visible in the explorer and might deter delegates as well. Plus there is always the possibility of an orchestrator trying such shenanigans coming under community scrutiny. 

fixups in cb323b0

pm/recipient_test.go
```diff
@@ -81,6 +82,13 @@ func genRecipientRand(sender ethcommon.Address, secret [32]byte, params *TicketP
 	msg = append(msg, params.ExpirationBlock.Bytes()...)
 	msg = append(msg, params.PricePerPixel.Num().Bytes()...)
 	msg = append(msg, params.PricePerPixel.Denom().Bytes()...)
+	msg = append(
```
yondonfu: I think this can be shortened to:

```golang
msg = append(msg, params.ExpirationParams.AuxData()...)
```

pm/sender.go
```diff
@@ -108,10 +108,18 @@ func (s *sender) CreateTicketBatch(sessionID string, size int) (*TicketBatch, er
 		return nil, err
 	}
 
-	expirationParams := s.expirationParams()
+	ticketParams := &session.ticketParams
+
+	expirationParams := ticketParams.ExpirationParams
+	// ensure backwards compatbility
```
yondonfu: Nit: "ensure" -> "Ensure"

pm/ticket.go
```diff
@@ -174,15 +176,17 @@ func (t *Ticket) Hash() ethcommon.Hash {
...
+		CreationRound:          t.CreationRound,
+		CreationRoundBlockHash: t.CreationRoundBlockHash,
+	}).AuxData()
+}
 
-	return append(
-		ethcommon.LeftPadBytes(big.NewInt(t.CreationRound).Bytes(), uint256Size),
-		t.CreationRoundBlockHash.Bytes()...,
-	)
+func (t *Ticket) ticketExpirationParams() *TicketExpirationParams {
```
yondonfu: Since the method receiver makes it clear that we are referring to a ticket what do you think about shortening this method name to `expirationParams()`?

yondonfu (changes requested):
> While the pmTicketParams() helper in db_discovery.go doesn't need to populate the expiration params field since those fields are not accessed within the file at the moment, it might be a good idea to populate the field anyways for consistency.

Bump on the above comment.

pm/ticket.go
kyriediculous: fixed in 87762826

pm/sender.go
kyriediculous: fixed in 87762826

pm/recipient_test.go
kyriediculous: fixed in 87762826

yondonfu (commented):
Changes look good - let's rebase.

kyriediculous:
rebased !

yondonfu (approved):
LGTM 🚢

PR #1429, opened 2020-03-20
MP4 support for HTTP push.

### LINKED ISSUE #1418: MP4 Support
## Abstract

Currently, go-livepeer works on mpegts segments. This proposal extends go-livepeer to support receiving MP4 segments in HTTP push, and allows the transcoder to output MP4.

## Motivation

Multiple input and output formats has been requested by users. This is also helps lay the groundwork for more robust VOD support, where diverse format support may be a requirement.

## Proposed Solution

* Accept MP4s as an HTTP ingest format
* The default output from the transcoder should match the input format. If a MP4 was ingested, the transcoded result should also be MP4 (and vice versa for mpegts).
* Stretch goal: configure the output format via webhook. This would allow for the ingest and output formats to differ. 

### HTTP Ingest

Currently, we reject any non-ts input for HTTP ingest. The only place this is "checked" is in ensuring the [request suffix](https://github.com/livepeer/go-livepeer/blob/master/server/mediaserver.go#L590-L596) ends with ".ts" [1]. This can be extended to accept `.mp4` or a Content-Type of `video/mp4`.

Once the ingest content type is determined, it needs to be passed on to the output so the output matches, unless specified otherwise. How to do this is TBD; most likely through inspecting the stream parameters after the [stream is authorized, but prior to registering the stream](https://github.com/livepeer/go-livepeer/blob/6c8dabd79c51d441048f4f56e9ef1cf905489183/server/mediaserver.go#L614-L623). Will probably need an "undefined" value for the output format in order to distinguish between an unset value and a default such as mpegts.

[1] In fact, it is possible today to send a MP4 [2] into Livepeer, and receive an output in mpegts. One only has to make the request with a `.ts` extension.

```
curl --data-binary "@video.mp4" http://localhost:8935/live/josh/10.ts
```

[2] A MP4 can be sent... or any arbitrary format that ffmpeg can read, really.  🐰🥚  

### LPMS Transcoding Changes

Currently, LPMS supports arbitrary [muxers and options](https://github.com/livepeer/lpms/blob/a886f6ba53baef55e872922d5df986f028e29af7/ffmpeg/ffmpeg.go#L56). However, the most straightforward way to propagate the specified muxer from the RPC layer would be to add the information to the [VideoProfiles struct](https://github.com/livepeer/lpms/blob/a886f6ba53baef55e872922d5df986f028e29af7/ffmpeg/videoprofile.go#L20-L26), slotting into our existing RPC conversion [1][2] .

The addition to the VideoProfiles will be an enum to facilitate validity checking and to minimize the user-exposed surface area for errors. The tradeoff here is a bit more boilerplate, in exchange for safety.

Adding muxing options to the VideoProfiles may be a bit redundant since the transcoder itself already supports arbitrary muxers, but it appears to be the most straightforward path to propagate the information through go-livepeer.

The API for the LPMS transcoder will not change: specifying a custom muxer outside what's made available within VideoProfile will still require setting the corresponding [options](https://github.com/livepeer/lpms/blob/a886f6ba53baef55e872922d5df986f028e29af7/ffmpeg/ffmpeg.go#L56). Setting muxer within the VideoProfile acts as a 'shortcut' for the given set of options. For example, the MP4 muxer may have the "faststart" option enabled to make the resulting MP4s playable from HTTP.  If both the VideoProfile and the TranscodeOptions both specify different muxers, then only the configuration enabled by the VideoProfile shortcut is used.

[1] [RPC "FullProfiles" to VideoProfile](https://github.com/livepeer/go-livepeer/blob/6c8dabd79c51d441048f4f56e9ef1cf905489183/server/segment_rpc.go#L206-L222)

[2] [Webhook response to VideoProfile](https://github.com/livepeer/go-livepeer/blob/6c8dabd79c51d441048f4f56e9ef1cf905489183/server/mediaserver.go#L203-L217)

### LPMS Playback Changes

The LPMS [segment handler](https://github.com/livepeer/lpms/blob/d03c749890cbbe028ab6daedf4855b711354abfd/vidplayer/player.go#L107-L108) is hard coded to only take m3u8 and ts files. This will have to be extended to also support mp4.

It is unclear whether such path constraints are even necessary.  However, removing those constraints are out of scope here, because further study will be needed to gauge the impact of doing so.

### Protobuf

Modify the [VideoProfile protobuf struct](https://github.com/livepeer/go-livepeer/blob/6c8dabd79c51d441048f4f56e9ef1cf905489183/net/lp_rpc.proto#L131-L146)  to take a muxer. The shape of the protobuf changes  should mirror that of [ffmpeg.VideoProfile](https://github.com/livepeer/lpms/blob/a886f6ba53baef55e872922d5df986f028e29af7/ffmpeg/videoprofile.go#L20-L26), so any decisions from ffmpeg should carry over, such as using an enumeration to restrict the muxers in use. (Enums would also have the advantage of being significantly smaller on the wire.)

### RPC Compatibility: Un-upgraded B working with upgraded O

If muxer is not specified by B within the RPC request, then use the mpegts by default. This will allow matching the current behavior. This way, an un-upgraded B working with an upgraded O can still get the results it expects: a mpegts segment.

### RPC Compatibility: Upgraded B working with non-upgraded O  - Protocol Versioning

Upgraded Bs using non-upgraded Os presents a problem: the non-upgraded Os will ignore the directive to output in a specific format, and return mpegts in all cases.

Ideally, the B could specify some minimum version (or at least, `O.version >= B.version`) Currently, we are missing a version signaling mechanism during orchestrator discovery. This can be implemented by adding a `version` field within [OrchestratorInfo](https://github.com/livepeer/go-livepeer/blob/6c8dabd79c51d441048f4f56e9ef1cf905489183/net/lp_rpc.proto#L87-L102), and filtering out [responses](https://github.com/livepeer/go-livepeer/blob/6c8dabd79c51d441048f4f56e9ef1cf905489183/discovery/discovery.go#L54-L64) within the discovery mechanism.

We previously had such a versioning mechanism earlier as part of an auth token exchange, but it was [removed](https://github.com/livepeer/go-livepeer/commit/fae70f384ab151b2f2c50495f3c406c706ef0412). Also see https://github.com/livepeer/go-livepeer/issues/695

```
curl --data-binary "@video.mp4" http://localhost:8935/live/josh/11.ts
```

### Documentation

This feature and all its implicit behaviors needs to be fully documented prior to merging, with [existing documentation](https://github.com/livepeer/go-livepeer/blob/master/doc/ingest.md#http-push) updated.

### Webhook Configurability

As a stretch goal, we may want to configure the output via webhook. This would involve modifying the expected [webhook response](https://github.com/livepeer/go-livepeer/blob/6c8dabd79c51d441048f4f56e9ef1cf905489183/server/mediaserver.go#L103-L114), then adding corresponding tests and documentation.

Each new or modified behavior should have complete test coverage.

## Additional Context

The MP4 support proposed here does not include fragmented MP4 (fMP4). This would require further work around managing the fMP4 initializer, and is currently not in scope.

The output of the RTMP segmenter would remain as mpegts. There does not seem to be a need to make this configurable (and would likely imply fMP4 anyway). However, the transcoded output can still be configured as MP4.

Taking RTMP input, transcoding that to MP4 poses problems for consumption, because inserting non-fragmented MP4s in HLS playlists violates the [HLS spec](https://tools.ietf.org/html/draft-pantos-http-live-streaming-23#section-3.1). While we could explicitly disable RTMP-to-MP4, it may still be useful to directly pull segments asynchronously.

While this work is being done as part of the groundwork for VOD, there remains much left to do for reasonable VOD support. Some short-term considerations include ingesting longer MP4s which may run up against the current 8-second HTTP timeout. Depending on user feedback, we may also have to make the timeout configurable, or make the timeout a multiplier over the content duration.

j0sh:

This implements most of https://github.com/livepeer/go-livepeer/issues/1418 . The exception is that it does not implement output configuration via webhook. The output can be set to MP4 via the HTTP push file extension. This can be quickly tested on the broadcaster via:

```
ffmpeg -re -i input -c:a copy -c:v copy -f hls -hls_segment_filename "http://localhost:8935/live/mp4/%d.mp4" -
```

and the results inspected via `curl` the resulting playlist(s) at http:localhost:8935/live/mp4.m3u8 . ffplay of individual segments will work as well.

- server: Take MP4s as input for HTTP push …
Most of the changes are in the following areas:

  * Adds a format -> extension lookup function
  * Adds a extension -> format inverse lookup function
  * Adds a format -> mime type lookup function
  * Protocol Buffers changes, and conversions to/from VideoProfile
  * Slotting in the lookups and error handling where appropriate
  * Lots and lots of tests.

The extensive testing and strict error handling is both to codify
intended behavior and to ensure misuse resistance to further changes. https://github.com/livepeer/go-livepeer/pull/1429/commits/949a06320544d326c25766f1da7c162c181a1e5e

- The file extension of temp files mostly doesn't matter for us anymore, so renamed the temp files we generate in order to avoid future confusion about the file type. https://github.com/livepeer/go-livepeer/pull/1429/commit/c23fc6079fd76d9c7564f3345df7695e4274b6bf

- Added documentation https://github.com/livepeer/go-livepeer/pull/1429/commits/39598171b72337b4f85b7aa3b43fcbecada6f153

- Updated LPMS to https://github.com/livepeer/lpms/pull/176 (needs to be merged prior to this ; any changes there will be updated here in go.mod)

- Unit testing. Tests are the bulk of the additions here.
- Manual testing / playback with the following simple scenarios (B -> OT, local storage on both, played back via ffplay):
  * RTMP
  * ffmpeg HLS push (with mpegts)
  * ffmpeg HLS push (with mp4, as described above)
  * Standalone T with MP4 and mpegts
- Visual inspection of multipart hexdump (but did not attempt to parse + extract)
- Checked with GCS to verify file extension and content type

Fixes https://github.com/livepeer/go-livepeer/issues/1418

**Checklist:**


darkdarkdragon (changes requested):
`mp4` muxer/demuxer should be added to `install_ffmpeg.sh`

j0sh:
@darkdarkdragon Ah yes.

Added a unit test that will iterate through all our formats, to catch this type of oversight in the future:  https://github.com/livepeer/go-livepeer/pull/1429/commits/8abe8abe03532820c8d8fe382d83f5220f8ddbc8 . That [failed](https://circleci.com/gh/livepeer/go-livepeer/5120?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) exactly  as expected.

Added mp4 to the muxer/demuxer and tests pass now. https://github.com/livepeer/go-livepeer/pull/1429/commits/06b87835cc5f671d342b60a91d821cc24759575f

j0sh:
Rebased to fix up a merge conflict. Git SHAs have changed but commit structure is otherwise the same in case someone was in the middle of reviewing things (eg, still contains the pending fixups from https://github.com/livepeer/go-livepeer/pull/1429#issuecomment-606856207)

j0sh:
https://github.com/livepeer/go-livepeer/pull/1429/commits/2c075d9ee24fb8f1b8e5f5eb9dabb78022355ec6 Updated to incorporate latest LPMS, including the libavfilter performance fix and MP4 support among other things:

https://github.com/livepeer/lpms/pull/176
https://github.com/livepeer/lpms/pull/177
https://github.com/livepeer/lpms/pull/178
https://github.com/livepeer/lpms/pull/179

darkdarkdragon (approved):
LGTM 🚢

server/segment_rpc.go
```diff
@@ -133,13 +134,20 @@ func (h *lphttp) ServeSegment(w http.ResponseWriter, r *http.Request) {
...
 	for i := 0; err == nil && i < len(res.TranscodeData.Segments); i++ {
-		name := fmt.Sprintf("%s/%d.ts", segData.Profiles[i].Name, segData.Seq) // ANGIE - NEED TO EDIT OUT JOB PROFILES
+		var ext string
+		ext, err = common.ProfileFormatExtension(segData.Profiles[i].Format)
+		if err != nil {
+			glog.Errorf("Unknown format extension manifestID=%s seqNo=%d err=%s", segData.ManifestID, segData.Seq, err)
+			break
+		}
+		name := fmt.Sprintf("%s/%d%s", segData.Profiles[i].Name, segData.Seq, ext)
+		// The use of := here is probably a bug?!?
```
yondonfu, line 150: Is the intent for any error encountered in this loop to be persisted via the `err` declared outside of this loop after we break out of so that the value of `err` can be used to populate the `net.TranscodeResult_Error` on L171 and passed to B?

```golang
var result net.TranscodeResult
if err != nil {
    glog.Errorf(...)
    result = net.TranscodeResult{Result: &net.TranscodeResult_Error{Error: err.Error()}}
}
```

If so, then yeah I think the short declaration on this line is indeed a bug since the `err` variable declared outside of this loop would still be nil when we break out of the loop and as a result the error from `res.OS.SaveData()` would not be passed to B.

Fine with addressing this separately from this PR though.

yondonfu (approved):
LGTM! Left one note about a comment about a possible bug in existing code. I'll leave it up to you whether to address it in this PR or separately (I noted that I'm fine with doing it separately).

server/segment_rpc.go
j0sh: That was the behavior I was expecting, yeah - failure to save should be propagated immediately to the B as an error.

But it looks like this has been here for a long time, and it's actually codified in a [unit test](https://github.com/livepeer/go-livepeer/blob/9659e5de2b63933f20bd34775c748781033b5074/server/segment_rpc_test.go#L337-L401). So I'm uncertain if ignoring the error was meant to address some other case. Best I can think of is something like using the results that were successfully saved, and retrying the rest? But since that's not anything we support right now, I'm also OK correcting the behavior.

j0sh: Bug fixed in https://github.com/livepeer/go-livepeer/pull/1429/commits/9aca289d6e123cdd3541493bc6ee388d36d111a6

Note that we [debit any fees](https://github.com/livepeer/go-livepeer/blob/9aca289d6e123cdd3541493bc6ee388d36d111a6/server/segment_rpc.go#L163-L164) that may have accumulated from an earlier iteration, but we don't return the accompanying pixel counts anymore. Not sure if that'll lead to a problem if the O is used again for subsequent segments (eg, if a refresh occurs soon thereafter). If it seems like this needs more work / thought, I'll drop the commit from this PR and we can address it separately.

yondonfu: > Not sure if that'll lead to a problem if the O is used again for subsequent segments (eg, if a refresh occurs soon thereafter)

With the per-stream suspension list, the O wouldn't be reused for awhile so I think this should only be a problem if there are only a small # of available Os (so B can't cycle through other Os before reusing one that returned an error).

That being said, I lean towards debiting fees only if we're returning segments/pixel counts and not debiting fees if we're returning a transcode error to minimize accounting issues (also feels right to not debit fees from a B's balance if O is not returning segments that could actually be used by B). Since this would require moving the `DebitFees()` call and writing some additional test cases I'm cool with dropping the commit and addressing separately.

j0sh: Reverted in 998619c86ce6ba027a18dddbd1931bc6344294d1 - will remove the commit from the tree when rebasing if that's a go

yondonfu: Sounds good 👍

PR #1430, opened 2020-03-24
server: error logging for transcodeSegment

### LINKED ISSUE #1421: O not receiving content from the B under "normal" operations.
**Describe the bug**

The O isn't receiving content from the B under "normal" operations

**To Reproduce**

1. Start an O with:
```
./livepeer-linux-amd64/livepeer -network rinkeby -orchestrator -cliAddr 127.0.0.1:7936 -httpAddr 127.0.0.1:8936 -serviceAddr 127.0.0.1:8936 -pricePerUnit 2000000000 -pixelsPerUnit 1 -ticketEV 1000000000000000 -v 99 -orchSecret secret
```
2. Enter password to decrypt private key
3. Start a T with:
```
./livepeer-linux-amd64/livepeer -network rinkeby -transcoder -orchAddr 127.0.0.1:8936 -cliAddr 127.0.0.1:7937 -httpAddr 127.0.0.1:8937 -orchSecret secret -maxSessions 1 -v 99
```
4. Start a B with:
```
./livepeer-linux-amd64/livepeer -network rinkeby -broadcaster -cliAddr 127.0.0.1:7935 -rtmpAddr 127.0.0.1:1935 -httpAddr 0.0.0.0:8935 -orchAddr 127.0.0.1:8936 -transcodingOptions P144p30fps16x9 -depositMultiplier 1 -maxPricePerUnit 5000000000 -pixelsPerUnit 1 -maxTicketEV 1000000000000000000 -v 99
```
5. Enter password to decrypt private key
6. Wait for the B to report `LPMS Server listening on rtmp://127.0.0.1:1935`
7. Start publishing into the B with:
```
ffmpeg -re -f lavfi -i testsrc=size=640x360:rate=30,format=yuv420p -f lavfi -i sine -threads 1 -c:v libx264 -b:v 1000k -preset ultrafast -x264-params keyint=30 -strict -2 -c:a aac -f flv rtmp://127.0.0.1:1935/hello_world
```
6. Observe that the B start to receive an process content
7. Run the following command:
```
curl http://127.0.0.1:8935/stream/hello_world.m3u8
```
8. Observe the following results:
```
#EXTM3U
#EXT-X-VERSION:3
#EXT-X-STREAM-INF:PROGRAM-ID=0,BANDWIDTH=4000000,RESOLUTION=640x360
hello_world/source.m3u8
```
9. Observe that the O and T are inactive (aside from the O caching gas prices)
10. Stop the process I started in step 7.

**Expected behavior**

At step 8 above, I expect to see:
```
#EXTM3U
#EXT-X-VERSION:3
#EXT-X-STREAM-INF:PROGRAM-ID=0,BANDWIDTH=4000000,RESOLUTION=640x360
hello_world/source.m3u8
#EXT-X-STREAM-INF:PROGRAM-ID=0,BANDWIDTH=400000,RESOLUTION=256x144
hello_world/P144p30fps16x9.m3u8
```

At step 9 above, I expect that the B will start sending content to the O, which in turn will send content to the T for transcoding, and then the O will serve the transcoded content to the B.

**Server**
 - OS: Ubuntu 18.04
```
Livepeer Node Version: 0.5.5
Architecture: amd64
Operating system: linux
```

**Log Files**

From the O
[Orchestrator Logs.txt](https://github.com/livepeer/go-livepeer/files/4329435/Orchestrator.Logs.txt)

From the T
[Transcoder Logs.txt](https://github.com/livepeer/go-livepeer/files/4329436/Transcoder.Logs.txt)

From the B
[Broadcaster Logs.txt](https://github.com/livepeer/go-livepeer/files/4329437/Broadcaster.Logs.txt)

kyriediculous:
Add error logging when failing ticket params validation (before submitting segment) or any error encountered when submitting the segment 

Fixes #1421 

**Checklist:**

### DIFF
--- a/pm/sender.go
+++ b/pm/sender.go
@@ -85,12 +85,12 @@ func (s *sender) validateSender(info *SenderInfo) error {
 		return ErrSenderValidation{fmt.Errorf("unable to validate sender: deposit and reserve is set to unlock soon")}
 	}
 
-	if info.Reserve.FundsRemaining.Int64() == 0 {
+	if info.Reserve.FundsRemaining.Cmp(big.NewInt(0)) == 0 {
 		return ErrSenderValidation{errors.New("no sender reserve")}
 	}
 
-	if info.Deposit.Int64() == 0 {
-		return errors.New("no sender deposit")
+	if info.Deposit.Cmp(big.NewInt(0)) == 0 {
+		return ErrSenderValidation{errors.New("no sender deposit")}
 	}
 
 	return nil
--- a/pm/sender.go
+++ b/pm/sender.go
@@ -79,15 +79,18 @@ func (s *sender) EV(sessionID string) (*big.Rat, error) {
 	return ticketEV(session.ticketParams.FaceValue, session.ticketParams.WinProb), nil
 }
 
-func (s *sender) validateSender() error {
-	info, err := s.senderManager.GetSenderInfo(s.signer.Account().Address)
-	if err != nil {
-		return ErrSenderValidation{fmt.Errorf("unable to validate sender: could not get sender info: %v", err)}
-	}
-
+func (s *sender) validateSender(info *SenderInfo) error {
 	maxWithdrawRound := new(big.Int).Add(s.timeManager.LastInitializedRound(), big.NewInt(1))
 	if info.WithdrawRound.Int64() != 0 && info.WithdrawRound.Cmp(maxWithdrawRound) != 1 {
 		return ErrSenderValidation{fmt.Errorf("unable to validate sender: deposit and reserve is set to unlock soon")}
+	}
+
+	if info.Reserve.FundsRemaining.Int64() == 0 {
+		return ErrSenderValidation{errors.New("no sender reserve")}
+	}
+
+	if info.Deposit.Int64() == 0 {
+		return errors.New("no sender deposit")
 	}
 
 	return nil
@@ -97,10 +100,6 @@ func (s *sender) validateSender() error {
 func (s *sender) CreateTicketBatch(sessionID string, size int) (*TicketBatch, error) {
 	session, err := s.loadSession(sessionID)
 	if err != nil {
-		return nil, err
-	}
-
-	if err := s.validateSender(); err != nil {
 		return nil, err
 	}
 
@@ -146,15 +145,20 @@ func (s *sender) ValidateTicketParams(ticketParams *TicketParams) error {
 
 // validateTicketParams checks if ticket params are acceptable for a specific number of tickets
 func (s *sender) validateTicketParams(ticketParams *TicketParams, numTickets int) error {
+	info, err := s.senderManager.GetSenderInfo(s.signer.Account().Address)
+	if err != nil {
+		return err
+	}
+
+	// validate sender
+	if err := s.validateSender(info); err != nil {
+		return err
+	}
+
 	ev := ticketEV(ticketParams.FaceValue, ticketParams.WinProb)
 	totalEV := ev.Mul(ev, new(big.Rat).SetInt64(int64(numTickets)))
 	if totalEV.Cmp(s.maxEV) > 0 {
 		return fmt.Errorf("total ticket EV %v for %v tickets > max total ticket EV %v", totalEV.FloatString(5), numTickets, s.maxEV.FloatString(5))
-	}
-
-	info, err := s.senderManager.GetSenderInfo(s.signer.Account().Address)
-	if err != nil {
-		return err
 	}
 
 	maxFaceValue := new(big.Int).Div(info.Deposit, big.NewInt(int64(s.depositMultiplier)))
--- a/server/broadcast.go
+++ b/server/broadcast.go
@@ -361,6 +361,7 @@ func transcodeSegment(cxn *rtmpConnection, seg *stream.HLSSegment, name string,
 	if sess.Sender != nil {
 		if err := sess.Sender.ValidateTicketParams(pmTicketParams(sess.OrchestratorInfo.TicketParams)); err != nil {
 			if err != pm.ErrTicketParamsExpired {
+				glog.Error("Invalid ticket params err=", err)
 				cxn.sessManager.removeSession(sess)
 				return nil, err
 			}
@@ -378,7 +379,7 @@ func transcodeSegment(cxn *rtmpConnection, seg *stream.HLSSegment, name string,
 	if err != nil || res == nil {
 		cxn.sessManager.removeSession(sess)
 		if res == nil && err == nil {
-			return nil, err
+			err = errors.New("empty response")
 		}
 		return nil, err
 	}
--- a/pm/sender_test.go
+++ b/pm/sender_test.go
@@ -164,7 +164,7 @@ func TestSender_ValidateSender(t *testing.T) {
 	err = s.validateSender(info)
 	assert.EqualError(err, "no sender deposit")
 	_, ok = err.(ErrSenderValidation)
-	assert.False(ok)
+	assert.True(ok)
 }
 
 func TestCreateTicketBatch_NonExistantSession_ReturnsError(t *testing.T) {
--- a/pm/sender_test.go
+++ b/pm/sender_test.go
@@ -93,6 +93,7 @@ func TestSenderEV(t *testing.T) {
 
 func TestSender_ValidateSender(t *testing.T) {
 	assert := assert.New(t)
+	require := require.New(t)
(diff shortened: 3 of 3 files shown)


server/broadcast.go
```diff
@@ -361,6 +361,7 @@ func transcodeSegment(cxn *rtmpConnection, seg *stream.HLSSegment, name string,
 	if sess.Sender != nil {
 		if err := sess.Sender.ValidateTicketParams(pmTicketParams(sess.OrchestratorInfo.TicketParams)); err != nil {
 			if err != pm.ErrTicketParamsExpired {
+				glog.Error("unable to submit segment err=", err)
```
yondonfu: Can we update the language in this log statement to better reflect the operation that failed? Maybe something like:

```golang
glog.Error("Invalid ticket params err=", err)
```

server/broadcast.go
```diff
@@ -376,10 +377,12 @@ func transcodeSegment(cxn *rtmpConnection, seg *stream.HLSSegment, name string,
 	}
 	res, err := SubmitSegment(sess, seg, nonce)
 	if err != nil || res == nil {
+		glog.Error("unable to submit segment err=", err)
```
yondonfu: Except for a few cases, most of the errors encountered in `SubmitSegment()` are logged in place so adding additional log statements here would lead to two log statements for most errors (and the second one has less information than the first one). So, I don't think we should add an additional log statement here right now. 

Alternatives include to make sure all errors are logged include:

- Add in place log statements for the remaining cases in `SubmitSegment()` - one example is the error from [newBalanceUpdate()](https://github.com/livepeer/go-livepeer/blob/ce819a15a616277a7081b5cb91339c4121a22538/server/segment_rpc.go#L302)
- Remove the in place log statements and instead return annotated errors that can be logged by the caller of `SubmitSegment()`.

The former is probably the simplest and least intrusive of the two.

server/broadcast.go
```diff
@@ -376,10 +377,12 @@ func transcodeSegment(cxn *rtmpConnection, seg *stream.HLSSegment, name string,
...
 		cxn.sessManager.removeSession(sess)
 		if res == nil && err == nil {
-			return nil, err
+			err = errors.New("empty response")
```
yondonfu, line 382: Ah so we used to return `errors.New("empty response")` here and I think it was accidentally changed in a previous PR - 👍 to switch back to returning `errors.New("empty response")` here.

yondonfu (changes requested):

server/broadcast.go
kyriediculous: Ah my bad this is actually not necessary. We already log the error when sender doesn't have enough balance.

We just never reach that code path if the sender doesn't have deposit & reserve because instead we'll already get an error because the max face value is 0 when reserve is 0. 

`E0325 13:30:42.773116   85004 broadcast.go:364] Invalid ticket params err=ticket faceValue 10000000000000000 > max faceValue 0`

Previously we would only validate the ticket params _after_ self-validating as a sender. 

I don't deem additional error logs in SubmitSegment necessary at this point for those cases that currently don't have error logging. WDYT?

kyriediculous: We could add a validateSender check https://github.com/livepeer/go-livepeer/blob/ce819a15a616277a7081b5cb91339c4121a22538/server/broadcast.go#L362 for a more informative error when the sender doesn't have enough balance.

server/broadcast.go
kyriediculous: yeah exactly. Could have done it in a seperate PR but I felt like it was somewhat related and the diff is rather small.

server/broadcast.go
kyriediculous: fixed in 571ab3d

server/broadcast.go
yondonfu: I'm on board with returning a more informative error from `ValidateTicketParams()`. Maybe something like this?

```golang
fmt.Errorf("faceValue too high faceValue=%v maxFaceValue=%v deposit=%v, ticketParams.FaceValue, maxFaceValue, sender.Deposit)
```

yondonfu: I'm on board with making the error clearer (i.e. note that the sender's deposit is 0) in `validateSender()`.

kyriediculous: Ok sorry I didn't look at the entire context for this.

I thought we check for 0 deposit/reserve in `validateSender` but this is not the case , what we do check is that the broadcasting funds are not set to unlock soon - https://github.com/livepeer/go-livepeer/pull/1318/files

When we run into a `validateSender` error we close the RTMP connection. 

Currently when the sender has 0 deposit/reserve we run into a `ticket faceVal > max faceVal` error emitted by  `ValidateTicketParams` and never reach `validateSender` in `SubmitSegment`  this error does _NOT_ close the RTMP connection and does not necessarily indicate 0 broadcasting funds. 

I think a  case  where the broadcaster doesn't have broadcasting funds  _SHOULD_ close the RTMP connection however. So how about we remove sender validation from `CreateTicketBatch` and instead export it and call `ValidateSender` in `transcodeSegment`.

yondonfu: Can we call `validateSender()` in `ValidateTicketParams()`? Seems like it accomplishes the same thing if `validateSender()` contains a zero deposit check. Would be good to have a zero reserve check in there as well.

kyriediculous: Ah yes, that's true. I would only check the reserve, as long as there is reserve the broadcaster does have "broadcasting funds" in theory and can still transcode potentially even when the deposit is 0 so that shouldn't lead to closing the RTMP connection.

yondonfu: The max faceValue check in `validateTicketParams()` will fail if the deposit is zero so B still would not send out a stream in that case. Might be worth adding the zero deposit check in `validateSender()` just to return a clear and explicit error.

kyriediculous: fixed in 571ab3d

yondonfu: Doesn't look like the check mentioned in this thread is in 571ab3d - is the plan to add that check in this PR?

kyriediculous: I believe with the current changes we don't need to add additional error logging in SubmitSegment, we now cover all critical user-facing error logs.

yondonfu: I'm referring to the deposit check in `validateSender()` that is discussed in the above comments. Wasn't sure if the intent here was to include it in this PR. FWIW I'm fine with not including that in this PR.

kyriediculous: Ah looks I forgot to push up the commit dae2a45

pm/sender.go
```diff
@@ -79,17 +79,20 @@ func (s *sender) EV(sessionID string) (*big.Rat, error) {
...
-		return ErrSenderValidation{fmt.Errorf("unable to validate sender: could not get sender info: %v", err)}
-	}
-
+func (s *sender) validateSender(info *SenderInfo) error {
 	maxWithdrawRound := new(big.Int).Add(s.timeManager.LastInitializedRound(), big.NewInt(1))
 	if info.WithdrawRound.Int64() != 0 && info.WithdrawRound.Cmp(maxWithdrawRound) != 1 {
 		return ErrSenderValidation{fmt.Errorf("unable to validate sender: deposit and reserve is set to unlock soon")}
 	}
 
+	if info.Reserve.FundsRemaining.Int64() == 0 {
```
yondonfu: `Int64()` can result in an underflow so we should use the `Cmp()` method for the comparison instead both here and the check below.

pm/sender.go
```diff
@@ -79,17 +79,20 @@ func (s *sender) EV(sessionID string) (*big.Rat, error) {
...
+		return ErrSenderValidation{errors.New("no sender reserve")}
+	}
+
+	if info.Deposit.Int64() == 0 {
+		return errors.New("no sender deposit")
```
yondonfu: Why not wrap this as a `ErrSenderValidation` error since no transcoding will happen until B has a non-zero deposit anyway?

yondonfu (changes requested):

pm/sender.go
kyriediculous: fixed in facbd57

pm/sender.go
kyriediculous: fixed in facbd57

yondonfu (commented):
Changes look good.

When rebasing can we squash the changes in dae2a45 and facbd57 into a separate commit since they are in a separate package and are not directly related to the original log statement update in this PR?

kyriediculous:
Sure thing, done !

yondonfu (approved):
LGTM 🚢

PR #1433, opened 2020-03-30
discovery: Add versioning to filter out incompatible orchestrators.

j0sh:

discovery: Add versioning to filter out incompatible orchestrators.

This is network protocol version, rather than a node version or
smart contract version. Its primary use is for the broadcaster
to filter out orchestrators that may not support certain features.

Broadcasters will only work with orchestrators such that

    orchestrator.version >= broadcaster.version

Newer orchestrators should remain backwards compatible with older
broadcasters, but we won't assume forward compatibility vis-a-vis
older orchestrators working with newer broadcasters.

This is independent of the node or smart contract versions, both
of which could be upgraded without breaking changes that would
require a bump in network protocol version compatibility.

- Updates OrchestratorInfo with a Version field
- Sets current version to 1 ( pending merge into https://github.com/livepeer/go-livepeer/pull/1429 )
- Updates discovery to filter out orchestrators < current version
- Fixes up unit tests to add a Version field for all OrchestratorInfo responses (this is a bit annoying)
- Add unit tests to ensure O exclusion on version mismatch

Unit tests

**Checklist:**


kyriediculous:
`orchestrator.version >= broadcaster.version`

An orchestrator might implement fields for `net.TicketParams` that are required to generate valid tickets as a B for that O. One such field is `ParamsExpirationBlock` introduced with the Ticket Expiry feature, another one is the expected `CreationRound` and `CreationRoundBlockHash` in #1428.
 
eg. B using 0.5.3 won't be able to send valid tickets to an O running 0.5.5 but a B running 0.5.5 is able to transcode with an O running 0.5.3 

> Newer orchestrators should remain backwards compatible with older
broadcasters

Therefore, I think this should be the other way around, _newer broadcasters should be compatible with older orchestrators_, orchestrators that are newer and not compatible with B or too old to be compatible with B will simply be dropped from the working set. Broadcasters cant build in compatibility for protocol message fields they have no knowledge over that were introduced in newer versions and having orchestrators handle payments differently based on broadcaster version would not be a good decision in my opinion. 

eg. 
- https://github.com/livepeer/go-livepeer/blob/31ab350970a5f10c381785b4fab5150ec445ecbb/pm/sender.go#L113
- https://github.com/livepeer/go-livepeer/blob/31ab350970a5f10c381785b4fab5150ec445ecbb/pm/sender.go#L169

j0sh:
Sure, just to clarify the problem this PR is trying to solve. Incorporating the version number (and bumping it later on) is meant for scenarios where the O would not be able to fail out on an unsupported request by the B.

Specifically,  for MP4 support (https://github.com/livepeer/go-livepeer/pull/1429), an older O would simply ignore the additional `Format` field in `net.VideoProfile` and continue transcoding to an incorrect format.

There might be other ways to achieve early fail-out on older Os outside of discovery, eg by modifying the submitted `net.SegData` to trigger a failure if using the older code path, which would achieve a similar effect, except being deferred from discovery time to segment submission time. An extension of this specifically for MP4 support might be to trigger such failures only if such a MP4 request was made, and keep things working for a mpegts request, allowing B to continue using older orchestrators only for mpegts requests. That allows the B to use blended networks, depending on the job parameters. But that approach is quite situational, requires a lot of careful plumbing and testing, and doesn't really generalize to other scenarios. At some point, we might want a hard cut-off, so I just opted to introduce that here.

Note that the versioning mechanism here is meant to be fairly coarse, I think / hope that we wouldn't be bumping this up too often absent some thought on how to maintain compatibility. For example, I don't expect that we'll need to bump the version each time we add a new format to LPMS, because the orchestrator can now fail out quickly if it receives a request for an unsupported format. That'll allow us to run on blended networks without bending over backwards. Note that this is yet another reason that a per-stream exclusion list makes more sense (https://github.com/livepeer/go-livepeer/issues/1375) : a stream requesting an older supported format may work with a given orchestrator, while another stream on the same B using a newer format may not work with the same orchestrator.

> Therefore, I think this should be the other way around, newer broadcasters should be compatible with older orchestrators

How do you propose to handle this if a broadcaster wants to use a feature that's not supported by the O?

At some point we may want to introduce discovery level broadcaster versioning. But that's not really the goal of this PR, since the problem is on the orchestrator side.

For the example of using ticket params, discovery-level versioning is not *required* as the O would still reject tickets from older Bs immediately, so things still work as intended. Discovery level versioning just gives a way to perform an explicit compatibility check earlier in the process, but bumping versions is basically a compatibility break and fragments the network. So that really should be the last resort, and balanced against the cost / complexity of maintaining compatibility each time the version is bumped.

darkdarkdragon (approved):
LGTM

kyriediculous:
Thanks for clarifying the goals of the PR. A few concerns still remain however.

- For this case using versioning doesn't take into account the actual format requested by the broadcaster. A broadcaster may upgrade its node software but not necessarily request MP4 format. The versioning now reduces the potential number of O's to work with since we have recently ensured backwards compatibility for the payment mechanism to nodes using older versions.
- I'm a little worried this change might be too harsh for what could potentially be a temporary solution which then could result in a one off version bump and as you mentioned on discord, it can be really hard to remove the change without breaking backwards compatibility again. 

I'd be in favor of not introducing a coarse mechanism that might be hard to remove in case it's temporary and spend more time thinking about capability discovery. 

> Note that this is yet another reason that a per-stream exclusion list makes more sense (#1375) : a stream requesting an older supported format may work with a given orchestrator, while another stream on the same B using a newer format may not work with the same orchestrator.

But we don't suspend for format errors _currently_ so I don't see how this would affect suspension. However, I would be in favor of moving to a per-stream suspension list and suspending for format errors if that means we don't have to implement coarse version checking and altering the network protocol. The check would now be at segment submission time but it is more flexible to change/build upon later on.

j0sh:
Superseded by https://github.com/livepeer/go-livepeer/pull/1439

Issue #1370, opened 2020-02-10
Flakey Test: TestOrchWatcher_HandleLog_TranscoderActivated


mk-livepeer:
This test fails sometimes but usually passes after repeating it.
```
--- FAIL: TestOrchWatcher_HandleLog_TranscoderActivated (0.01s)
    orchestratorwatcher_test.go:79: 
        	Error Trace:	orchestratorwatcher_test.go:79
        	Error:      	Not equal: 
        	            	expected: 1
        	            	actual  : 0
        	Test:       	TestOrchWatcher_HandleLog_TranscoderActivated
```
This can cause CI to become unpredictable. It isn't an urgent problem right now - just creating an issue so that it can be tracked.

yondonfu:
This test fails sometimes because the assertion checks for a statement that is logged in a goroutine, but the statement might not be logged yet after the fixed [sleep](https://github.com/livepeer/go-livepeer/blob/cf3e4e59e67e06c76d3efb54e22e2f6e83e5620d/eth/watchers/orchestratorwatcher_test.go#L77) time. This scenario is possible for all event watcher tests that sleep prior to running an assertion that checks state dependent on code execution in a goroutine.

One way to address this issue is to increase the sleep time in these types of tests. However, the downside is that these tests take longer to run. Furthermore, if we do not increase the sleep time by a sufficient amount then the tests could still fail sometimes.

A more robust approach might involve something like the following:

- Defining a `logHandler` interface with a `HandleLog()` function
- Each event watcher type uses a `logHandler` interface and has an associated type that implements this interface. For example, the `OrchestratorWatcher` would use a `orchestratorLogHandler` type
- We can unit test the log handling behavior for the type that implements the `logHandler` interface
- We can unit test the internal loop started by an event watcher and check that it properly invokes the `HandeLog()` function of the `logHandler` by stubbing the `logHandler` in tests.

kyriediculous:
This problem is however not exclusive to this area of the codebase (eg. `TestSelectSession`), the proposed solution would require us to implement additional logic everywhere we have a test depending on a goroutine to finish, despite that logic not being necessary for normal node operations. 

It also involves additional engineering overhead for anywhere we will be using this pattern in the future. 

Perhaps another less coarse solution is to introduce an environment variable for the sleep duration where the default is something like 2 milliseconds, but for CI we can slightly bump this up due to less available resources. 

e.g. `SLEEP=5ms go test ./...`

[...]

PR #1435, opened 2020-04-02
multi: per-stream O suspension list with refresh prioritisation mechanism

### LINKED ISSUE #1375: Exclude Os that timeout from session refresh for a period of time

kyriediculous:

This PR adds a per -stream orchestrator suspension list. Orchestrators will be suspended for Transcoding related errors and timeouts and no longer be included in a session refresh unless there aren't enough orchestrators available otherwise. 

The suspension time is dynamic based on the `OrchestratorPool` size : `OrchestratorPool.Size() / numOrchs`

Suspended orchestrators can still get selected upon a refresh though in case there are not enough non-suspended orchestrators available to fill the working set up to the required size. This happens based on a priority queue implemented as a min-heap, where the value to sort on is the suspension penalty associated with the orchestrator.

The penalty is the refresh count at which the orchestrator is no longer suspended (zero for non-suspended orchestrators). 

* the suspension list is for all streams and is stored in memory 

- Introduced a `suspender` type in `server`
- Introduced `priorityQueue` in `discovery`
- Introduced a `Suspender`  interface in `common`

- Changed the `OrchestratorPool.GetOrchestrators(numOrchs int)` API signature to also take a  type implementing the`Suspender` interface as a second argument 
- Added a `Suspender.Suspended() == 0` check in the `GetOrchestratorInfo` loop to determine if an orchestrator is suspended, suspended orchestrators get added to the `priorityQueue`
- If the length of retrieved O's is smaller than `numOrchs` requested, Pop from the `priorityQueue` until full

- Call `newSuspender` in `NewSessionManager()` constructor to initialise each new `BroadcastSessionsManager`  with a `suspender`
- Added a `BroadcastSessionsManager.suspendOrchestator()` method, this will call `suspender.Suspend()` if `shouldStopSession()` evaluates to true
- Removed the `selectOrchestrators` method and the `BroadcastSessionsManager.createSessions` struct field. Instead turned this into a `BroadcastSessionsManager.createSessions()` method
- Stored the values we pass into `selectOrchestrators` as state on the `BroadcastSessionsManager` instead, this should not change much as the memory reserved for the arguments also exists until BroadcastSessionsManager is GC'd (when the stream ends). 

Added & ran unit tests

Fixes #1375

**Checklist:**

### DIFF
--- a/discovery/discovery.go
+++ b/discovery/discovery.go
@@ -44,13 +44,13 @@ func (o *orchestratorPool) GetURLs() []*url.URL {
 	return o.uris
 }
 
-func (o *orchestratorPool) GetOrchestrators(numOrchestrators int) ([]*net.OrchestratorInfo, error) {
+func (o *orchestratorPool) GetOrchestrators(numOrchestrators int, suspender common.Suspender) ([]*net.OrchestratorInfo, error) {
 	numAvailableOrchs := len(o.uris)
 	numOrchestrators = int(math.Min(float64(numAvailableOrchs), float64(numOrchestrators)))
 	ctx, cancel := context.WithTimeout(context.Background(), getOrchestratorsTimeoutLoop)
 
-	infoCh := make(chan *net.OrchestratorInfo, len(o.uris))
-	errCh := make(chan error, len(o.uris))
+	infoCh := make(chan *net.OrchestratorInfo, numAvailableOrchs)
+	errCh := make(chan error, numAvailableOrchs)
 	getOrchInfo := func(uri *url.URL) {
 		info, err := serverGetOrchInfo(ctx, o.bcast, uri)
 		if err == nil && (o.pred == nil || o.pred(info)) {
@@ -64,8 +64,8 @@ func (o *orchestratorPool) GetOrchestrators(numOrchestrators int) ([]*net.Orches
 	}
 
 	// Shuffle into new slice to avoid mutating underlying data
-	uris := make([]*url.URL, len(o.uris))
-	for i, j := range rand.Perm(len(o.uris)) {
+	uris := make([]*url.URL, numAvailableOrchs)
+	for i, j := range rand.Perm(numAvailableOrchs) {
 		uris[i] = o.uris[j]
 	}
 
@@ -76,11 +76,13 @@ func (o *orchestratorPool) GetOrchestrators(numOrchestrators int) ([]*net.Orches
 	timeout := false
 	infos := []*net.OrchestratorInfo{}
 	nbResp := 0
-	for i := 0; i < len(uris) && len(infos) < numOrchestrators && !timeout; i++ {
+	for i := 0; i < numAvailableOrchs && len(infos) < numOrchestrators && !timeout; i++ {
 		select {
 		case info := <-infoCh:
-			infos = append(infos, info)
-			nbResp++
+			if penalty := suspender.Suspended(info.Transcoder); penalty == 0 {
+				infos = append(infos, info)
+				nbResp++
+			}
 		case <-errCh:
 			nbResp++
 		case <-ctx.Done():
--- a/discovery/discovery.go
+++ b/discovery/discovery.go
@@ -1,6 +1,7 @@
 package discovery
 
 import (
+	"container/heap"
 	"context"
 	"math"
 	"math/rand"
@@ -75,14 +76,17 @@ func (o *orchestratorPool) GetOrchestrators(numOrchestrators int, suspender comm
 
 	timeout := false
 	infos := []*net.OrchestratorInfo{}
+	suspendedInfos := newSuspensionQueue()
 	nbResp := 0
 	for i := 0; i < numAvailableOrchs && len(infos) < numOrchestrators && !timeout; i++ {
 		select {
 		case info := <-infoCh:
 			if penalty := suspender.Suspended(info.Transcoder); penalty == 0 {
 				infos = append(infos, info)
-				nbResp++
+			} else {
+				heap.Push(suspendedInfos, &suspension{info, penalty})
 			}
+			nbResp++
 		case <-errCh:
 			nbResp++
 		case <-ctx.Done():
@@ -90,6 +94,15 @@ func (o *orchestratorPool) GetOrchestrators(numOrchestrators int, suspender comm
 		}
 	}
 	cancel()
+
+	if len(infos) < numOrchestrators {
+		diff := numOrchestrators - len(infos)
+		for i := 0; i < diff && suspendedInfos.Len() > 0; i++ {
+			info := heap.Pop(suspendedInfos).(*suspension).orch
+			infos = append(infos, info)
+		}
+	}
+
 	glog.Infof("Done fetching orch info numOrch=%d responses=%d/%d timeout=%t",
 		len(infos), nbResp, len(uris), timeout)
 	return infos, nil
--- a/discovery/discovery_test.go
+++ b/discovery/discovery_test.go
@@ -72,7 +72,7 @@ func TestDeadLock(t *testing.T) {
 	assert := assert.New(t)
 	wg.Add(len(uris))
 	pool := NewOrchestratorPool(nil, uris)
-	infos, err := pool.GetOrchestrators(1)
+	infos, err := pool.GetOrchestrators(1, newStubSuspender())
 	assert.Nil(err, "Should not be error")
 	assert.Len(infos, 1, "Should return one orchestrator")
 	assert.Equal("transcoderfromtestserver", infos[0].Transcoder)
@@ -119,7 +119,7 @@ func TestDeadLock_NewOrchestratorPoolWithPred(t *testing.T) {
 
 	wg.Add(len(uris))
 	pool := NewOrchestratorPoolWithPred(nil, uris, pred)
-	infos, err := pool.GetOrchestrators(1)
+	infos, err := pool.GetOrchestrators(1, newStubSuspender())
 
 	assert.Nil(err, "Should not be error")
 	assert.Len(infos, 1, "Should return one orchestrator")
@@ -235,7 +235,7 @@ func TestNewDBOrchestratorPoolCache_GivenListOfOrchs_CreatesPoolCacheCorrectly(t
 	pool, err := NewDBOrchestratorPoolCache(ctx, node, &stubRoundsManager{})
 	require.NoError(err)
 	assert.Equal(pool.Size(), 3)
-	orchs, err := pool.GetOrchestrators(pool.Size())
+	orchs, err := pool.GetOrchestrators(pool.Size(), newStubSuspender())
 	for _, o := range orchs {
 		assert.Equal(o.PriceInfo, expPriceInfo)
 		assert.Equal(o.Transcoder, expTranscoder)
@@ -579,7 +579,7 @@ func TestCachedPool_AllOrchestratorsTooExpensive_ReturnsEmptyList(t *testing.T) 
 
 	urls := pool.GetURLs()
 	assert.Len(urls, 0)
-	infos, err := pool.GetOrchestrators(len(addresses))
+	infos, err := pool.GetOrchestrators(len(addresses), newStubSuspender())
 
 	assert.Nil(err, "Should not be error")
 	assert.Len(infos, 0)
@@ -670,7 +670,7 @@ func TestCachedPool_GetOrchestrators_MaxBroadcastPriceNotSet(t *testing.T) {
 	for _, url := range urls {
 		assert.Contains(addresses, url.String())
 	}
-	infos, err := pool.GetOrchestrators(50)
+	infos, err := pool.GetOrchestrators(50, newStubSuspender())
 	for _, info := range infos {
 		assert.Equal(info.PriceInfo, expPriceInfo)
 		assert.Equal(info.Transcoder, expTranscoder)
@@ -782,7 +782,7 @@ func TestCachedPool_N_OrchestratorsGoodPricing_ReturnsNOrchestrators(t *testing.
 		assert.Contains(addresses[25:], url.String())
 	}
 
-	infos, err := pool.GetOrchestrators(len(orchestrators))
+	infos, err := pool.GetOrchestrators(len(orchestrators), newStubSuspender())
 
 	assert.Nil(err, "Should not be error")
 	assert.Len(infos, 25)
@@ -844,7 +844,7 @@ func TestCachedPool_GetOrchestrators_TicketParamsValidation(t *testing.T) {
 	sender.On("ValidateTicketParams", mock.Anything).Return(errors.New("ValidateTicketParams error")).Times(25)
 	sender.On("ValidateTicketParams", mock.Anything).Return(nil).Times(25)
 
-	infos, err := pool.GetOrchestrators(len(addresses))
+	infos, err := pool.GetOrchestrators(len(addresses), newStubSuspender())
 	assert.Nil(err)
 	assert.Len(infos, 25)
 	sender.AssertNumberOfCalls(t, "ValidateTicketParams", 50)
@@ -852,7 +852,7 @@ func TestCachedPool_GetOrchestrators_TicketParamsValidation(t *testing.T) {
 	// Test 0 out of 50 orchs pass ticket params validation
 	sender.On("ValidateTicketParams", mock.Anything).Return(errors.New("ValidateTicketParams error")).Times(50)
 
-	infos, err = pool.GetOrchestrators(len(addresses))
+	infos, err = pool.GetOrchestrators(len(addresses), newStubSuspender())
 	assert.Nil(err)
 	assert.Len(infos, 0)
 	sender.AssertNumberOfCalls(t, "ValidateTicketParams", 100)
@@ -950,7 +950,7 @@ func TestCachedPool_GetOrchestrators_OnlyActiveOrchestrators(t *testing.T) {
 	for _, url := range urls {
 		assert.Contains(addresses[:25], url.String())
 	}
-	infos, err := pool.GetOrchestrators(50)
+	infos, err := pool.GetOrchestrators(50, newStubSuspender())
 	for _, info := range infos {
 		assert.Equal(info.PriceInfo, expPriceInfo)
 		assert.Equal(info.Transcoder, expTranscoder)
@@ -993,7 +993,7 @@ func TestNewWHOrchestratorPoolCache(t *testing.T) {
 	whpool.mu.Lock()
 	lastReq := whpool.lastRequest
 	whpool.mu.Unlock()
-	orchInfo, err := whpool.GetOrchestrators(2)
+	orchInfo, err := whpool.GetOrchestrators(2, newStubSuspender())
 	require.Nil(err)
 	assert.Len(orchInfo, 2)
 	assert.Equal(3, whpool.Size())
@@ -1012,7 +1012,7 @@ func TestNewWHOrchestratorPoolCache(t *testing.T) {
 	whpool.mu.Lock()
 	whpool.lastRequest = lastReq
 	whpool.mu.Unlock()
-	orchInfo, err = whpool.GetOrchestrators(2)
+	orchInfo, err = whpool.GetOrchestrators(2, newStubSuspender())
 	require.Nil(err)
 	assert.Len(orchInfo, 2)
 	assert.Equal(3, whpool.Size())
@@ -1035,7 +1035,7 @@ func TestNewWHOrchestratorPoolCache(t *testing.T) {
 	whpool.mu.Lock()
 	whpool.lastRequest = lastReq
 	whpool.mu.Unlock()
-	orchInfo, err = whpool.GetOrchestrators(2)
+	orchInfo, err = whpool.GetOrchestrators(2, newStubSuspender())
 	require.Nil(err)
 	assert.Len(orchInfo, 2)
 	assert.Equal(3, whpool.Size())
@@ -1055,7 +1055,7 @@ func TestNewWHOrchestratorPoolCache(t *testing.T) {
 	whpool.mu.Lock()
 	whpool.lastRequest = lastReq
 	whpool.mu.Unlock()
-	orchInfo, err = whpool.GetOrchestrators(2)
+	orchInfo, err = whpool.GetOrchestrators(2, newStubSuspender())
 	require.Nil(err)
 	assert.Len(orchInfo, 2)
 	assert.Equal(3, whpool.Size())
@@ -1143,21 +1143,23 @@ func TestOrchestratorPool_GetOrchestrators(t *testing.T) {
 	serverGetOrchInfo = func(ctx context.Context, bcast common.Broadcaster, server *url.URL) (*net.OrchestratorInfo, error) {
 		defer wg.Done()
 		err := orchCb()
-		return &net.OrchestratorInfo{}, err
+		return &net.OrchestratorInfo{
+			Transcoder: server.String(),
+		}, err
 	}
 
 	pool := NewOrchestratorPool(nil, addresses)
 
 	// Check that we receive everything
 	wg.Add(len(addresses))
-	res, err := pool.GetOrchestrators(len(addresses))
+	res, err := pool.GetOrchestrators(len(addresses), newStubSuspender())
 	assert.Nil(err)
 	assert.Len(res, len(addresses))
 
 	// Check that partial results are received if requested
 	wg.Add(len(addresses))
 	assert.Greater(len(addresses), 1) // sanity
-	res, err = pool.GetOrchestrators(1)
+	res, err = pool.GetOrchestrators(1, newStubSuspender())
 	assert.Nil(err)
 	assert.Len(res, 1)
 	wg.Wait() // prevents races on remaining responses
@@ -1165,7 +1167,7 @@ func TestOrchestratorPool_GetOrchestrators(t *testing.T) {
 	// Check error handling: all errors
 	wg.Add(len(addresses))
 	orchCb = func() error { return errors.New("Error") }
-	res, err = pool.GetOrchestrators(len(addresses))
+	res, err = pool.GetOrchestrators(len(addresses), newStubSuspender())
 	assert.Nil(err)
 	assert.Len(res, 0)
 
@@ -1183,7 +1185,7 @@ func TestOrchestratorPool_GetOrchestrators(t *testing.T) {
 	}
 	wg.Add(len(addresses))
 	start := time.Now()
-	res, err = pool.GetOrchestrators(len(addresses))
+	res, err = pool.GetOrchestrators(len(addresses), newStubSuspender())
 	end := time.Now()
 	assert.Nil(err)
 	assert.Len(res, len(addresses)-1)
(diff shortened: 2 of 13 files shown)


kyriediculous:
I will drop 7f58e51 in the rebase as this is fixed in https://github.com/livepeer/go-livepeer/pull/1438

discovery/discovery.go
```diff
@@ -75,19 +76,37 @@ func (o *orchestratorPool) GetOrchestrators(numOrchestrators int) ([]*net.Orches
...
 	nbResp := 0
-	for i := 0; i < len(uris) && len(infos) < numOrchestrators && !timeout; i++ {
+	for i := 0; i < numAvailableOrchs && len(infos) < numOrchestrators && !timeout; i++ {
 		select {
 		case info := <-infoCh:
-			infos = append(infos, info)
-			nbResp++
+			if penalty := suspender.Suspended(info.Transcoder); penalty == 0 {
+				infos = append(infos, info)
+				nbResp++
```
yondonfu: Looks like we can move `nbResp++` out of the if/else blocks since it is executed in both cases.

discovery/discovery.go
```diff
@@ -75,19 +76,37 @@ func (o *orchestratorPool) GetOrchestrators(numOrchestrators int) ([]*net.Orches
...
 		case <-errCh:
 			nbResp++
 		case <-ctx.Done():
 			timeout = true
 		}
 	}
 	cancel()
+
+	if len(infos) < numOrchestrators {
+		diff := int(math.Min(float64(numOrchestrators-nbResp), float64(suspendedInfos.Len())))
```
yondonfu: `nbResp` includes error responses so it could actually be greater than `numOrchestrators`. Also, if we're checking the length of `suspendedInfos` in the loop below, then we don't need to use `math.Min` here.

I think this could just be:

```golang
diff := numOrchestrators - len(infos)
```

discovery/discovery.go
```diff
@@ -75,19 +76,37 @@ func (o *orchestratorPool) GetOrchestrators(numOrchestrators int) ([]*net.Orches
...
+		for i := 0; i <= diff; i++ {
```
yondonfu: I think the check should be `i < diff` or else we'll pop one too many infos from `suspendedInfos`.

discovery/discovery.go
```diff
@@ -75,19 +76,37 @@ func (o *orchestratorPool) GetOrchestrators(numOrchestrators int) ([]*net.Orches
...
+			if suspendedInfos.Len() == 0 {
```
yondonfu: Nit: We can move this into the for loop conditional:

```golang
for i := 0; i < diff && suspendedInfos.Len() > 0; i++ { ... }
```

discovery/priorityqueue.go
```diff
@@ -0,0 +1,46 @@
+package discovery
+
+import (
+	"container/heap"
+
+	"github.com/livepeer/go-livepeer/net"
+)
+
+// A PriorityQueue implements heap.Interface and holds Items.
+type priorityQueue []*suspension
```
yondonfu: I feel like the name of this type can better capture the elements it stores since the type is not a generic priority queue and will only be used to store `suspension` elements anyway. Maybe `suspensionQueue`?

discovery/discovery_test.go
```diff
@@ -1187,18 +1189,54 @@ func TestOrchestratorPool_GetOrchestrators(t *testing.T) {
...
+
+	// don't include suspended orchestrators if enough orchestrators are available
+	res, err := pool.GetOrchestrators(2, sus)
+	assert.Nil(err)
+	assert.Len(res, 2)
+	assert.NotEqual(res[0].GetTranscoder(), "https://127.0.0.1:8938")
+	assert.NotEqual(res[1].GetTranscoder(), "https://127.0.0.1:8938")
+
+	// include suspended O's if not enough non-suspended O's available
+	require.Greater(sus.Suspended("https://127.0.0.1:8938"), int64(0))
```
yondonfu: A few additional cases that I think would be useful to add tests for:

- Insufficient non-suspended Os, but zero suspended Os (i.e. X non-suspended Os, 0 suspended Os, ask for X + 1 Os)
- Insufficient non-suspended Os, but non-zero insufficient suspended Os (i.e. X non-suspended Os, Y suspended Os, ask for X + Y + 1 Os)
- Given insufficient non-suspended Os, a suspended O with a lower penalty is included before a suspended O with a higher penalty

discovery/priorityqueue.go
```diff
@@ -0,0 +1,46 @@
...
+// A suspension is the item we manage in the priority queue.
+type suspension struct {
+	orch    *net.OrchestratorInfo
+	penalty int64
+}
+
+func (pq priorityQueue) Len() int { return len(pq) }
+
+func (pq priorityQueue) Less(i, j int) bool {
+	return pq[i].penalty > pq[j].penalty
```
yondonfu: I think this should be:

```golang
pq[i].penalty < pq[j].penalty
```

since we want `heap.Pop()` to return the O with the lowest penalty.

discovery/discovery.go
```diff
@@ -75,19 +76,37 @@ func (o *orchestratorPool) GetOrchestrators(numOrchestrators int) ([]*net.Orches
...
+				break
+			}
+			info := suspendedInfos.Pop().(*suspension).orch
```
yondonfu: This call won't preserve the heap invariant (i.e. the next popped element is the minimum) for `suspendedInfos`. In order to preserve the heap invariant, we need to call `heap.Pop(suspendedInfos)`.

server/broadcast.go
```diff
@@ -509,7 +518,13 @@ func transcodeSegment(cxn *rtmpConnection, seg *stream.HLSSegment, name string,
 	return segURLs, nil
 }
 
-var sessionErrStrings = []string{"dial tcp", "unexpected EOF", core.ErrOrchBusy.Error(), core.ErrOrchCap.Error()}
+func (bsm *BroadcastSessionsManager) suspendOrchestrator(orch *net.OrchestratorInfo, err error) {
+	if shouldStopSession(err) {
```
yondonfu: There are a number of other errors that we should probably suspend Os for:

- `No transcoders available`
- `Remote transcoder took too long`
- Errors related to unsupported formats (see #1439) 

This list is not exhaustive and there may be more. Maybe we can just suspend Os for any error returned from `SubmitSegment()`. The downside to this is that some of the errors returned from `SubmitSegment()` may not be returned from O. That seems ok to me given that if B triggers an error in `SubmitSegment()` before sending a request to O then it means that either a) something went wrong in the segment/payment setup for the O b) there is a problem with B that may impact its ability to work with any O. In both situations, suspending the O doesn't seem to have major negative consequences. The upside to this is that we do not need to do special error handling - if an error is encountered while trying to use an O it gets suspended.

yondonfu (changes requested):

discovery/discovery_test.go
kyriediculous: fixed in 848c69e79e042d7a97508d80f238406285c85a0a

server/broadcast.go
kyriediculous: fixed in d4b6bf9

discovery/priorityqueue.go
kyriediculous: fixed in 98953f4

discovery/discovery.go
kyriediculous: fixed in 98953f4

discovery/discovery.go
kyriediculous: fixed in 98953f4

discovery/discovery.go
kyriediculous: fixed in 98953f4

discovery/discovery.go
kyriediculous: fixed in 98953f4

kyriediculous:
cd0868c, cd0868c and 3b1d7fe changes the `Suspender` API to use  `int` for both `count` and `penalty` rather than `int64` which stemmed from when this mechanism still involved timestamps.

server/broadcast.go
```diff
@@ -509,7 +518,7 @@ func transcodeSegment(cxn *rtmpConnection, seg *stream.HLSSegment, name string,
 	return segURLs, nil
 }
 
-var sessionErrStrings = []string{"dial tcp", "unexpected EOF", core.ErrOrchBusy.Error(), core.ErrOrchCap.Error()}
+var sessionErrStrings = []string{"Client.Timeout", "dial tcp", "unexpected EOF", core.ErrOrchBusy.Error(), core.ErrOrchCap.Error()}
```
yondonfu: I think we can revert this change since we no longer are using `shouldStopSession()` to determine when to suspend Os.

server/broadcast_test.go
```diff
@@ -164,6 +189,7 @@ func TestStopSessionErrors(t *testing.T) {
 
 	// check error cases
 	errs := []string{
+		"Client.Timeout exceeded",
```
yondonfu: I think we can revert this change since we're not longer using `shouldStopSession()` to determine when to suspend Os.

server/mediaserver_test.go
```diff
@@ -91,6 +88,19 @@ func (d *stubDiscovery) GetOrchestrators(num int) ([]*net.OrchestratorInfo, erro
 	return d.infos, err
 }
 
+func (d *stubDiscovery) SuspendOrchestrator(orch string) {
```
yondonfu: Doesn't look like this method is used.

server/broadcast_test.go
```diff
@@ -260,13 +287,11 @@ func TestSelectSession(t *testing.T) {
...
+	// assert session list gets refreshed if under threshold
 	bsm.numOrchs = 1
-	var wg sync.WaitGroup
-	wg.Add(1)
-	bsm.createSessions = func() ([]*BroadcastSession, error) { wg.Done(); return nil, fmt.Errorf("err") }
-	bsm.selectSession()
-	assert.True(wgWait(&wg), "Session refresh timed out")
+	updateBsmSessList(bsm.BroadcastSessionsManager, []*BroadcastSession{StubBroadcastSession("transcoder3")})
+	sess = bsm.selectSession()
+	assert.Equal(bsm.sessMap["transcoder3"], sess)
```
yondonfu: I don't think this updated test is checking the same thing as the old test.

The old test used a wait group to make sure that `createSessions()` was called asynchronously by `selectSession()`. Looks like this updated test adds transcoder3 to the sessMap using `updateBsmSessList()` and then checks that sessMap contains transcoder3 which would be checking the behavior of `updateBsmSessList()` as opposed to `selectSession()`.

server/broadcast_test.go
```diff
@@ -393,9 +546,10 @@ func TestRefreshSessions(t *testing.T) {
 	sess3 := StubBroadcastSession("transcoder3")
 	sess4 := StubBroadcastSession("transcoder4")
 
-	bsm.createSessions = func() ([]*BroadcastSession, error) {
-		return []*BroadcastSession{sess3, sess4}, nil
-	}
+	bsm.sessMap[sess3.OrchestratorInfo.Transcoder] = sess3
```
yondonfu: This code directly adds new sessions to sessMap/sessList and then we check if these sessions are properly added to sessMap/sessList. This is a different from the old test which was checking if given that `createSessions()` returns sess3 and sess4 whether the sessMap/sessList will be properly updated after a call to `refreshSessions()`.

server/broadcast_test.go
```diff
@@ -405,13 +559,12 @@ func TestRefreshSessions(t *testing.T) {
 	assert.Equal(bsm.sessList()[1], sess4)
 
 	// asserting that refreshes stop while another is in-flight
-	bsm.createSessions = func() ([]*BroadcastSession, error) {
-		return []*BroadcastSession{StubBroadcastSession("5"), StubBroadcastSession("6")}, nil
-	}
+	delete(bsm.sessMap, "transcoder3")
```
yondonfu: Shouldn't this test be checking that the length of sessMap remains unchanged even when refreshing would introduce new sessions (i.e. sessions 5 and 6) because there is already an in-flight session? We would just want to have the refresh introduce new sessions as opposed to changing the current sessions in sessMap.

server/broadcast_test.go
```diff
@@ -435,21 +588,18 @@ func TestRefreshSessions(t *testing.T) {
...
 	}
 	assert.True(wgWait(&wg), "Session refresh timed out")
+	bsm.node.OrchestratorPool = pool
 
 	// check empty returns from createSession. Run this under -race
-	bsm.createSessions = func() ([]*BroadcastSession, error) {
-		time.Sleep(time.Millisecond)
-		return []*BroadcastSession{}, nil
-	}
+	updateBsmSessList(bsm.BroadcastSessionsManager, []*BroadcastSession{})
```
yondonfu: I think this is different from the old test behavior. The old test behavior simulated an empty session list being returned from `createSessions()`. On L598 we assign set the orchestrator pool to an existing instance of `stubDiscovery` which already has a populated `infos` slice. This call to `updateBsmSessList()` accepts an empty session list, but the underlying `stubDiscovery` instance would still have a non-empty session list.

server/broadcast.go
```diff
@@ -185,24 +190,27 @@ func NewSessionManager(node *core.LivepeerNode, params *streamParameters, pl cor
...
+		poolSize: int(poolSize),
+		sus:      newSuspender(),
 	}
 	bsm.refreshSessions()
 	return bsm
 }
 
-func selectOrchestrator(n *core.LivepeerNode, params *streamParameters, cpl core.PlaylistManager, count int) ([]*BroadcastSession, error) {
-	if n.OrchestratorPool == nil {
+func (bsm *BroadcastSessionsManager) createSessions() ([]*BroadcastSession, error) {
```
yondonfu: I understand that defining this method on `BroadcastSessionsManager` feels cleaner, but from looking at the tests it actually feels like this change makes writing tests for the refresh behavior of `BroadcastSessionsManager` more difficult. Previously, we could just assign a stub implementation of `createSessions()` to simulate different session lists returned during a refresh. Now, we have to modify a `stubDiscovery` instance to simulate different session lists, but we also have to execute all the code in this `createSessions()` method creating additional dependencies that we have to take into consideration as we update/add BroadcastSessionsManager tests related to refresh behavior.

At the moment, this doesn't feel like a worthwhile trade, but I'm happy to hear arguments for why it would be worthwhile (maybe if there are tweaks to the proposed refactor).

server/broadcast.go
```diff
@@ -377,6 +385,7 @@ func transcodeSegment(cxn *rtmpConnection, seg *stream.HLSSegment, name string,
 	}
 	res, err := SubmitSegment(sess, seg, nonce)
 	if err != nil || res == nil {
+		cxn.sessManager.sus.suspend(sess.OrchestratorInfo.GetTranscoder(), cxn.sessManager.poolSize/cxn.sessManager.numOrchs)
```
yondonfu: Taking a look at `transcodeSegment()`, there are a few other places where we remove an O from the working set:

- When a direct upload to O fails
- When ticket params validation fails with an error that is not `ErrTicketParamsExpired`
- When an attempt to refresh the ticket params for an O fails
- When an attempt to download from O fails
- When we encounter a retryable verification error (tampering was detected, but could be a false negative)

We don't want to suspend an O for the last case (since it could be a false negative [1]), but we could suspend Os for the other cases by adding the suspension logic into `removeSession()`. In order to avoid suspending an O for the last case we could replace the [removeSession()](https://github.com/livepeer/go-livepeer/blob/ce63d6c5e7ef9da33c05b55fed7b573f0cf4adf4/server/broadcast.go#L546) call with `deleteSession()` (naming?) that does not include suspension logic and *only* removes the session from sessMap.

[1] Side note: We still need to add in logic for re-adding those Os back to the working set in that scenario.

yondonfu (changes requested):

server/broadcast.go
kyriediculous: I agree that it makes writing the tests easier, experienced that first hand, but should we really be monkey patching a method within a single package? I more of consider that to be a last resort. There's a good balance to everything, isn't this one slightly tipping the scales to the wrong side a little bit? I mean looking at it is is essentially just a method for the type. 

Also I bit the bullet on this because my compiler showed an error trying to pass `suspender` as an argument to the function we assign to `createSessions` but now that I fixed my go tooling that seemed to have been a false error.

j0sh: Agreed with Yondon - the changes to the testing flow is a little heavy-handed, so this is actually one instance where it feels like a worthwhile tradeoff. Moreover, there is a lot of additional state being added to the `BroadcastSessionManager` simply for this one function; I believe that was another reason for keeping `selectOrchestrators` outside the BroadcastSessionManager. The less state we have to carry around, the better.

yondonfu: > but should we really be monkey patching a method within a single package?

The fact that we can insert a stub function during tests isn't too different from inserting a stub implementation of an interface during tests (consider if `createSessions` was instead an `SessionCreator` interface with a single `CreateSessions` method). 

I think the reason that the current usage of `createSessions` doesn't feel clean is probably because `NewSessionManager` always uses the same function implementation instead of accepting the function implementation as a parameter and because the current function implementation closes around free variables that are passed to `selectOrchestrator` instead of storing them in a struct.

server/mediaserver_test.go
kyriediculous: fixed in 2c78fb9a784ea81c42a9bdbd0d91c71ab6fc255b

server/broadcast_test.go
kyriediculous: fixed in 2c78fb9a784ea81c42a9bdbd0d91c71ab6fc255b

server/broadcast_test.go
kyriediculous: fixed in 2c78fb9a784ea81c42a9bdbd0d91c71ab6fc255b

server/broadcast_test.go
kyriediculous: fixed in 2c78fb9a784ea81c42a9bdbd0d91c71ab6fc255b

server/broadcast_test.go
kyriediculous: fixed in 2c78fb9a784ea81c42a9bdbd0d91c71ab6fc255b

server/broadcast.go
kyriediculous: dropped those changes , fixed in 2c78fb9a784ea81c42a9bdbd0d91c71ab6fc255b

server/broadcast_test.go
kyriediculous: fixed in f25299b

server/broadcast.go
kyriediculous: fixed in f25299b

server/broadcast.go
kyriediculous: I'd rather stick to duplicating 
```
		cxn.sessManager.sus.suspend(sess.OrchestratorInfo.GetTranscoder(), cxn.sessManager.poolSize/cxn.sessManager.numOrchs)
```

I think suspension of an orchestrator and removal of a session are different things,  and combining this in a method could lead to confusion. For the similar reason I would also think that eg. keeping `removeSession` and introducing `suspendAndRemoveSession` wouldn't necessarily be better. 

So in my opinion duplicate, but perhaps we can introduce a wrapper method
`bsm.suspendOrchestrator(*BroadcastSession)` for brevity

server/broadcast.go
```diff
@@ -185,24 +189,28 @@ func NewSessionManager(node *core.LivepeerNode, params *streamParameters, pl cor
...
-		mid:            params.mid,
-		sel:            sel,
-		sessMap:        make(map[string]*BroadcastSession),
-		createSessions: func() ([]*BroadcastSession, error) { return selectOrchestrator(node, params, pl, numOrchs) },
-		sessLock:       &sync.Mutex{},
-		numOrchs:       numOrchs,
+		mid:     params.mid,
+		sel:     sel,
+		sessMap: make(map[string]*BroadcastSession),
+		createSessions: func(sus *suspender) ([]*BroadcastSession, error) {
```
yondonfu: What do you think about avoiding changing the signature of `createSessions()` by declaring `sus` first and having `selectOrchestrator` close over it? Example:

```golang
sus := newSuspender()
bsm := &BroadcastSessionsManager{
    ...
    createSessions: func() ([]*BroadcastSession, error) {
        return selectOrchestrator(node, params, pl, numOrchs, sus)
    },
    ...
    sus: sus,
}
```

server/broadcast.go
yondonfu: Agreed that suspending an O and removing a session are different things and I see how the method name `removeSession()` may not adequately convey to a reader that it will both suspend an O and remove a session.

I'm fine with a wrapper `suspendOrchestrator` method. Let's also add tests to check that we invoke this method for each of the relevant cases mentioned.

server/broadcast_test.go
```diff
@@ -247,15 +250,18 @@ func TestSelectSession(t *testing.T) {
 
 	// assert last session selected and sessList is correct length
 	sess := bsm.selectSession()
+	time.Sleep(2 * time.Millisecond)
```
yondonfu: Are these time.Sleep calls necessary? Looks like the next few tests aren't checking for refresh behavior.

yondonfu (changes requested):

server/broadcast_test.go
kyriediculous: I added these sleeps because the previous goroutine wasn’t finishing before starting the waitgroup test, the wait group counter got deducted by the goroutine from the previous test and so our actual test would result in a negative count and thus a panic. This happened consistently for me locally.

yondonfu: Ah I see.

I think another way to resolve this without using sleeps is to create a new `bsm` before the waitgroup test. Starting after L270:

```golang
bsm = newSessionsManagerLIFO(bsmWithSessList([]*BroadcastSession{}))
bsm.sessMap["transcoder1"] = StubBroadcastSession("transcoder1")
bsm.sessMap["transcoder2"] = StubBroadcastSession("transcoder2")
```

The last 2 lines aren't strictly necessary. But, previously, sessMap had 2 entries when running the waitgroup test and we can replicate that scenario with the last 2 lines.

As mentioned in [this comment](https://github.com/livepeer/go-livepeer/issues/1370#issuecomment-611250411) I think it is worthwhile to avoid introducing new tests that use sleeps if we can given the flaky behavior we've seen in CI in the past.

server/broadcast.go
kyriediculous: having to reference the suspender twice here is another reason why I don't like the current setup for `createSessions()` but having dropped the previous commits I guess it is what it is at this point.

server/broadcast_test.go
kyriediculous: Since we also alter the `createSessions` implementation for this specific test I don't really see much added value in trying to necessarily replicate the previous scenario as it wasn't really relevant initially either. But I'm okay with creating a new `bsm` , thanks !

kyriediculous: fixed in d313eb53

server/broadcast.go
kyriediculous: fixed in d313eb53

server/broadcast.go
kyriediculous: fixed in d313eb53

Created new tests for the upload and download cases

discovery/discovery_test.go
```diff
@@ -1193,18 +1195,80 @@ func TestOrchestratorPool_GetOrchestrators(t *testing.T) {
...
 }
 
+func TestOrchestratorPool_GetOrchestrators_SuspendedOrchs(t *testing.T) {
+	assert := assert.New(t)
+	require := require.New(t)
+	addresses := stringsToURIs([]string{"https://127.0.0.1:8936", "https://127.0.0.1:8937", "https://127.0.0.1:8938"})
+
+	orchCb := func() error { return nil }
+	oldOrchInfo := serverGetOrchInfo
+	defer func() { serverGetOrchInfo = oldOrchInfo }()
```
yondonfu: Looks like CI is complaining about a data race here:

```
==================
WARNING: DATA RACE
Write at 0x000002cf3c28 by goroutine 191:
  github.com/livepeer/go-livepeer/discovery.TestOrchestratorPool_GetOrchestrators_SuspendedOrchs.func2()
      /build/discovery/discovery_test.go:1205 +0x3a
  github.com/livepeer/go-livepeer/discovery.TestOrchestratorPool_GetOrchestrators_SuspendedOrchs()
      /build/discovery/discovery_test.go:1258 +0xf06
  testing.tRunner()
      /usr/lib/go-1.13/src/testing/testing.go:909 +0x199

Previous read at 0x000002cf3c28 by goroutine 182:
  github.com/livepeer/go-livepeer/discovery.(*orchestratorPool).GetOrchestrators.func1()
      /build/discovery/discovery.go:56 +0xa0

Goroutine 191 (running) created at:
  testing.(*T).Run()
      /usr/lib/go-1.13/src/testing/testing.go:960 +0x651
  testing.runTests.func1()
      /usr/lib/go-1.13/src/testing/testing.go:1202 +0xa6
  testing.tRunner()
      /usr/lib/go-1.13/src/testing/testing.go:909 +0x199
  testing.runTests()
      /usr/lib/go-1.13/src/testing/testing.go:1200 +0x521
  testing.(*M).Run()
      /usr/lib/go-1.13/src/testing/testing.go:1117 +0x2ff
  main.main()
      _testmain.go:90 +0x223

Goroutine 182 (finished) created at:
  github.com/livepeer/go-livepeer/discovery.(*orchestratorPool).GetOrchestrators()
      /build/discovery/discovery.go:74 +0x470
  github.com/livepeer/go-livepeer/discovery.TestOrchestratorPool_GetOrchestrators_SuspendedOrchs()
      /build/discovery/discovery_test.go:1221 +0x3cc
  testing.tRunner()
      /usr/lib/go-1.13/src/testing/testing.go:909 +0x199
==================
--- FAIL: TestOrchestratorPool_GetOrchestrators_SuspendedOrchs (0.00s)
    testing.go:853: race detected during execution of test
```

Should be able to fix with something like what is done in this commit: https://github.com/livepeer/go-livepeer/commit/2752da4a394b5623dce06542d6e64848b121efd8

kyriediculous: fixed in 6d3937b89aac4c754ba97b6ffd844d0035ee8f38

discovery/discovery_test.go
```diff
@@ -1193,18 +1195,88 @@ func TestOrchestratorPool_GetOrchestrators(t *testing.T) {
...
+	// include suspended O's if not enough non-suspended O's available
+	wg.Add(len(addresses))
+	require.Greater(sus.Suspended("https://127.0.0.1:8938"), 0)
+	res, err = pool.GetOrchestrators(3, sus)
+	assert.Nil(err)
+	assert.Len(res, 3)
+	// suspended Os are added last
+	assert.Equal(res[2].Transcoder, "https://127.0.0.1:8938")
+
+	// no suspended O's, insufficient non-suspended O's
```
yondonfu, line 1240: Doesn't seem like this comment is accurate since there are still suspended Os in this test case since the same instance of `sus` is used and we suspended `https://127.0.0.1:8938` on L1220?

discovery/discovery_test.go
```diff
@@ -1193,18 +1195,88 @@ func TestOrchestratorPool_GetOrchestrators(t *testing.T) {
...
+	wg.Add(len(addresses))
+	res, err = pool.GetOrchestrators(4, sus)
+	assert.Nil(err)
+	assert.Len(res, 3)
+
+	// insufficient non-suspended O's, insufficient suspended O's
+	wg.Add(len(addresses))
+	sus.list["https://127.0.0.1:8938"] = 5
```
yondonfu, line 1249: I think `https://127.0.0.1:8938` is already suspended at this point if we're using the same instance of `sus`?

yondonfu (changes requested):
2 minor testing related comments - feel free to rebase after addressing.

kyriediculous:
Addressed those comments and rebased

yondonfu (approved):
LGTM :shipit: pending green CI

PR #1436, opened 2020-04-02
[READY] DroneCI GPU integration

ya7ya:
This adds DroneCI configuration to `go-livepeer`, which is like circleCI but self hosted and we have a modified version that allows us to run the tests on GPU enabled pods.

**How does it work?**
Once a PR is opened or new commits are pushed, DroneCI will run 2 `go-livepeer` deployments, one running the PR build , and one running Master build. Then using `stream-tester` it'll stream a sample file to each deployment and get back the results.

The results can then be compared to check whether the new PR build has a better success rate than the current `master` build. The goal is to always make sure we're improving the success rate and not adding new issues.

**improvements**

It'd be cool to compile a sample file that contains all the edge case segments that we come across to make sure the transcoder can still handle them with the new PR modifications.

**Current issues with this PR**

- The tests will fail only if one of the processes returns an error code, this isn't triggered when the transcoder gets a bad segment and keeps on going. but it'll be reflected in the success rate 

- the current test deployment uses 1 broadcaster 1 orchestrator configuration. but this can be changed to a 3 orchestrator setup if needed.

Let me know what you think, Feedback is welcome :wave:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -47,14 +47,18 @@ jobs:
           # Our Docker tag name should be our branch name with just alphanums
           BRANCH_TAG=$(echo $CIRCLE_BRANCH | sed 's/\//-/g' | tr -cd '[:alnum:]_-')
           VERSION_TAG=$(./print_version.sh)
+          SHA_TAG=$(git describe --always --long --abbrev=8)
           docker build -t current-build -f docker/Dockerfile.release-linux .
-          for TAG in $BRANCH_TAG $VERSION_TAG; do
+          for TAG in $BRANCH_TAG $VERSION_TAG $SHA_TAG; do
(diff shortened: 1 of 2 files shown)


criticaltv:
Hooray, I support all use of Continuous Integration. Bravo!

Here's a proposal for something to test:

**Input:** binaries created by [@darkdarkdragon's automated-build-process](https://github.com/criticaltv/go-livepeer/blob/patch-1/doc/install.md#option-5-automated-build-process---latest-codebase).

**Setup:** [run the binary](https://github.com/criticaltv/go-livepeer/blob/patch-1/doc/install.md#running-the-software) with `./livepeer -broadcaster`

**Test:** perform [basic test of the running software](https://github.com/criticaltv/go-livepeer/blob/patch-1/doc/install.md#basic-test-of-the-installation).

PR #1437, opened 2020-04-02
pm: fix expected ticket expiration params backwards compatability

kyriediculous:
This PR fixes the backwards compatibility requirement for #1428. 

Originally there is a check that retrieves the necessary `ExpirationParams` from the `Timewatcher` if the Orchestrator does not provide these in `TicketParams`
```
	expirationParams := ticketParams.ExpirationParams
	// Ensure backwards compatbility
	// If no expirationParams are included by O
	// B sets the values based upon its last seen round
	if expirationParams == nil {
		expirationParams = s.expirationParams()
	}
```

We however us a helper `pmTicketParams` on the broadcaster to convert the received `net.TicketParams` to `pm.TicketParams`

This helper initialises `ticketParams.ExpirationParams` 

```		
ExpirationParams: &pm.TicketExpirationParams{
			CreationRound:          params.ExpirationParams.GetCreationRound(),
			CreationRoundBlockHash: ethcommon.BytesToHash(params.ExpirationParams.GetCreationRoundBlockHash()),
		},
```

So we actually have to check for zero values on `expirationParams` fields and not the encapsulating struct itself 

Adjusted & ran unit tests
Ran Labrador on rinkeby

**Checklist:**

### DIFF
--- a/pm/sender.go
+++ b/pm/sender.go
@@ -6,6 +6,7 @@ import (
 	"sync"
 	"sync/atomic"
 
+	ethcommon "github.com/ethereum/go-ethereum/common"
 	"github.com/pkg/errors"
 )
 
@@ -113,7 +114,7 @@ func (s *sender) CreateTicketBatch(sessionID string, size int) (*TicketBatch, er
 	// Ensure backwards compatbility
 	// If no expirationParams are included by O
 	// B sets the values based upon its last seen round
-	if expirationParams == nil {
+	if expirationParams == nil || expirationParams.CreationRound == 0 || expirationParams.CreationRoundBlockHash == (ethcommon.Hash{}) {
 		expirationParams = s.expirationParams()
 	}
 
--- a/pm/sender_test.go
+++ b/pm/sender_test.go
@@ -286,9 +286,8 @@ func TestCreateTicketBatch_UsesSessionParamsInBatch(t *testing.T) {
 	assert.Equal(ticketParams.ExpirationBlock, batch.ExpirationBlock)
 	assert.Equal(ticketParams.PricePerPixel, batch.PricePerPixel)
 	assert.Equal(expectedExpParams, batch.ExpirationParams)
-	// No ExpirationParams, get data from TimeManager
 
-	// ExpirationParams sent by orchestrator
+	// No ExpirationParams, get data from TimeManager
 	ticketParams = TicketParams{
 		Recipient:         recipient,
 		FaceValue:         big.NewInt(1111),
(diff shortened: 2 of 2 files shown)


pm/sender.go
```diff
@@ -113,7 +114,7 @@ func (s *sender) CreateTicketBatch(sessionID string, size int) (*TicketBatch, er
 	// Ensure backwards compatbility
 	// If no expirationParams are included by O
 	// B sets the values based upon its last seen round
-	if expirationParams == nil {
+	if expirationParams == nil || (expirationParams.CreationRound == 0 && expirationParams.CreationRoundBlockHash == ethcommon.Hash{}) {
```
yondonfu:
```golang
expirationParams == nil || expirationParams.CreationRound == 0 || expirationParams.CreationRoundBlockHash == ethcommon.Hash{}
```

Should the conditional be the above instead? If either field in `expirationParams` is the zero value the sender would set the expiration params on its own.

pm/sender_test.go
```diff
@@ -297,6 +296,7 @@ func TestCreateTicketBatch_UsesSessionParamsInBatch(t *testing.T) {
 		RecipientRandHash: recipientRandHash,
 		ExpirationBlock:   big.NewInt(1),
 		PricePerPixel:     big.NewRat(1, 1),
+		ExpirationParams:  &TicketExpirationParams{},
```
yondonfu, line 299: Should we have a test that checks for a nil `ExpirationParams` as well as the new conditions?

pm/sender.go
kyriediculous: Yes although both values are always used in tandem so it's never either.

We _could_ even skip the `expirationParams == nil` check since it should never be nil (due to the usage of `pmticketparams`

pm/sender_test.go
kyriediculous: This should actually never be nil in the real case due to the usage of `pmTicketParams()` but I can add a test (see comment above though about the nil check)

pm/sender.go
kyriediculous: fixed in c3bcc39f

pm/sender_test.go
kyriediculous: fixed in c3bcc39f

kyriediculous:
CI should pass once the rebase includes #1438

yondonfu (commented):

kyriediculous:
rebased !

yondonfu (approved):
LGTM 🚢

PR #1439, opened 2020-04-04
server, net: Handle back compat for MP4.

j0sh:
This is another approach to compatibility rather than a hard break via versioning as described in https://github.com/livepeer/go-livepeer/pull/1433 .

* Makes MP4 requests fail out on orchestrators that don't support MP4 (or more generally, the `FullProfile` field for non-preset renditions introduced in https://github.com/livepeer/go-livepeer/pull/1242 )

* Maintains compatibility with orchestrators that can only process mpegts

- https://github.com/livepeer/go-livepeer/pull/1439/commits/5deedb6ae742af5a939214927ddfa65633550e1f Adds a new `FullProfiles2` field to net.SegData
- https://github.com/livepeer/go-livepeer/pull/1439/commits/103c540ae9761e2f15f998e72e8a656e7e7b5334 Use the old `FullProfiles` field if the output format is only mpegts . This maintains compatibility with older orchestrators that only know mpegts.
- https://github.com/livepeer/go-livepeer/pull/1439/commits/2212e9cd00639d0bdfa93d861993a38de0d20d86 Fails out with orchestrators that do not support the above two fields
- Adds unit tests
- https://github.com/livepeer/go-livepeer/pull/1439/commits/9dfcabfb3cc4676410be52a329fc675e759e8ea6 Update to generated protocol buffers, to be fixed up with https://github.com/livepeer/go-livepeer/pull/1429/commits/bf412084c7d5ec464915ad84bdd2149ea47f7e80

* Added unit tests
* Manual testing with:
  * old B -> new O (mpegts only)
  * new B -> old O (one test for mpegts, another for mp4)
  * new B -> new O (one test for mpegts, another for mp4)

**Checklist:**

### DIFF
--- a/net/lp_rpc.proto
+++ b/net/lp_rpc.proto
@@ -124,8 +124,12 @@ message SegData {
   // XXX should we include this in a sig somewhere until certs are authenticated?
   repeated OSInfo storage = 32;
 
-  // Transcoding profiles to use. Supersedes `profiles` field 
+  // Transcoding profiles to use. Supersedes `profiles` field
+  // Deprecated by `fullProfiles2` but may still be used for mpegts formats
   repeated VideoProfile fullProfiles = 33;
+
+  // Transcoding profiles to use. Supersedes `fullProfiles` field
+  repeated VideoProfile fullProfiles2 = 34;
 }
 
 message VideoProfile {
--- a/server/segment_rpc.go
+++ b/server/segment_rpc.go
@@ -252,7 +252,9 @@ func verifySegCreds(orch Orchestrator, segCreds string, broadcaster ethcommon.Ad
 	}
 
 	profiles := []ffmpeg.VideoProfile{}
-	if len(segData.FullProfiles) > 0 {
+	if len(segData.FullProfiles2) > 0 {
+		profiles, err = makeFfmpegVideoProfiles(segData.FullProfiles2)
+	} else if len(segData.FullProfiles) > 0 {
 		profiles, err = makeFfmpegVideoProfiles(segData.FullProfiles)
 	} else if len(segData.Profiles) > 0 {
 		profiles, err = common.BytesToVideoProfile(segData.Profiles)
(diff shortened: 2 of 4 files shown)


common/util.go
```diff
@@ -147,17 +147,24 @@ func DefaultProfileName(width int, height int, bitrate int) string {
 	return fmt.Sprintf("%dx%d_%d", width, height, bitrate)
 }
 
-func FFmpegProfiletoNetProfile(ffmpegProfiles []ffmpeg.VideoProfile) ([]*net.VideoProfile, error) {
+func FFmpegProfiletoNetProfile(ffmpegProfiles []ffmpeg.VideoProfile) ([]*net.VideoProfile, bool, error) {
```
j0sh: I'm not thrilled with this change - it is a bit "leaky" in that it's quite specifically catering to the caller's need to know which `net.SegData.FullProfile/2` field to use within `genSegCreds`, not to mention the sig change to all the test cases - but the alternative is to do another iteration through the set of profiles in `genSegCreds` .

Happy to make changes if someone has a strong preference for another way.

core/orchestrator.go
```diff
@@ -663,7 +663,7 @@ func (rt *RemoteTranscoder) Transcode(job string, fname string, profiles []ffmpe
 		return nil, RemoteTranscoderFatalError{err}
 	}
 
-	fullProfiles, err := common.FFmpegProfiletoNetProfile(profiles)
+	fullProfiles, _, err := common.FFmpegProfiletoNetProfile(profiles)
```
j0sh: Assuming for now that standalone Ts will be upgraded in tandem with Os, so we aren't attempting to handle back compat here.

common/util.go
yondonfu: I would prefer to iterate through the set of profiles in `genSegCreds` because it would isolate the all mpegts checking code making it easier to remove in the future if we wanted to (i.e. at a point that it seems unlikely for a node to be unaware of `net.SegData.FullProfile2`). While iterating through the profiles twice is not ideal, I don't think there will be *that* many profiles in practice so this feels ok.

net/lp_rpc.proto
```diff
@@ -126,6 +126,10 @@ message SegData {
 
   // Transcoding profiles to use. Supersedes `profiles` field 
   repeated VideoProfile fullProfiles = 33;
+
+  // Transcoding profiles to use.
+  //  Supersedes `fullProfiles` field but may still be used for mpegts formats
```
yondonfu: Nit: Looks like these two lines are misaligned.

yondonfu (changes requested):

net/lp_rpc.proto
j0sh: Redid the comments in https://github.com/livepeer/go-livepeer/pull/1439/commits/f27bef80a7de76bf591d737cbda7231c97826eec

common/util.go
j0sh: Fixed in https://github.com/livepeer/go-livepeer/pull/1439/commits/f27bef80a7de76bf591d737cbda7231c97826eec

server/segment_rpc.go
```diff
@@ -506,13 +508,32 @@ func genSegCreds(sess *BroadcastSession, seg *stream.HLSSegment) (string, error)
...
+	// If all outputs are mpegts, use the older SegData.FullProfiles field
+	// for compatibility with older orchestrators
+	allTS := true
+	for _, p := range sess.Profiles {
+		switch p.Format {
+		case ffmpeg.FormatNone: // default output is mpegts for FormatNone
+		case ffmpeg.FormatMPEGTS:
+		default:
+			allTS = false
+			break
```
yondonfu: I believe this will only break out of the switch, but will not break out of the for loop since break statements only terminate the innermost for, switch, select (see https://golang.org/ref/spec#Break_statements). We'd have to label the loop to break out of the loop - something like:

```golang

allTSLoop:
for _, p := range sess.Profiles {
    switch p.Format {
        ...
        default:
            allTS = false
            break allTSLoop
    }
}

j0sh: Ah yeah. Am I the only one who thinks that switch-break is a nonsense feature for a language that doesn't have default switch case fallthrough? (Unlike, say, C/C++ which clearly inspired this behavior.) Fixed in fa819dfee0d05f25b5eee8d6734e42c2848b629f

j0sh: ( further fixed up by 6358ceb403198e59cfd017fc8f48cd28c4de4f21  ; original changes didn't even compile )

yondonfu (approved):
LGTM :shipit: after a squash

PR #1444, opened 2020-04-09
Remove LIFO nvidia stack.

### LINKED ISSUE #1443: Transcoding broken with nvidia and standalone T
**Describe the bug**

When running in OT mode with software transcoding, we use disk storage which works fine. But with nvidia and standalone T, local memory OS is used. Local memory OS is routed through LPMS, and that currently only allows a limited set of file extensions through. LPMS does not like our ".tempfile" extension. Fixed by https://github.com/livepeer/lpms/pull/182

https://github.com/livepeer/go-livepeer/blob/4ba2a732a31b673254008faeb903e28bf7039a74/core/orchestrator.go#L490-L506

We could also fix the behavior in some ways other than https://github.com/livepeer/lpms/pull/182. We could use a proper extension, but the actual type of the source file isn't available at this point; it needs to be propagated up, probably from the content type within ServeSegment. Note the extension doesn't really matter, at least for ffmpeg which determines the file type by probing the contents. So that seems like unnecessary effort to satisfy a rather arbitrary restriction in LPMS.

We could also use a valid extension such as *.ts but that would be incorrect when other formats are used, such as MP4, and it seems that would lead to some confusion later on. So I'd prefer to err on the side of making it clear that the file extension does not matter at all. Also see https://github.com/livepeer/go-livepeer/commit/adfa5beab97db4022d421e43379deec03dd815aa

j0sh:
Removes the LIFO nvidia stack that was introduced in https://github.com/livepeer/go-livepeer/pull/1364 . No longer needed since https://github.com/livepeer/lpms/pull/179 improves perf enough that serialization is no longer necessary.

- https://github.com/livepeer/go-livepeer/commit/22f7fffe111833f2f420f2166c348b2320b9afbb Reverts most of https://github.com/livepeer/go-livepeer/pull/1364 and associated code / tests
- https://github.com/livepeer/go-livepeer/commit/f285743f438696bf26fc3b57bb43847eccde7886 Updates LPMS to incorporate https://github.com/livepeer/lpms/pull/182 to fix https://github.com/livepeer/go-livepeer/issues/1443 (can split out this to a separate PR if that's preferred, but needed for testing at the moment)

Manual testing

**Checklist:**


darkdarkdragon (approved):
LGTM 🚢

j0sh:
e55a15b0f7eb4ace03dd9e24e29339210846535b Updated LPMS to incorporate https://github.com/livepeer/lpms/pull/184 (needs to be merged first, then will finalize the fixup here)

yondonfu (approved):

PR #1445, opened 2020-04-10
On node startup try to transcode 

### LINKED ISSUE #1373: Should check if GPUs accessible at start
If node is configured with wrong GPU number it starts successfully, but can't transcode:
```
Feb 10 20:12:16 orchestrator-1080-9fbd9957d-ndqvp chi [AVHWDeviceContext @ 0x7fdec9240180] cu->cuDeviceGet(&hwctx->internal->cuda_device, device_idx) failed -> CUDA_ERROR_INVALID_DEVICE: invalid device ordinal
Feb 10 20:12:16 orchestrator-1080-9fbd9957d-ndqvp chi Freeing input based on OPEN INPUT error
Feb 10 20:12:16 orchestrator-1080-9fbd9957d-ndqvp chi E0211 04:12:19.078064       6 ffmpeg.go:292] Transcoder Return : Unknown error occurred
Feb 10 20:12:16 orchestrator-1080-9fbd9957d-ndqvp chi I0211 04:12:19.078123       6 lb.go:189] LB: Stopping transcoder due to error for 2378debc-8e2d-4a10-90c5-6738cf096dbb_8
```

Try to transcode some test segment at node startup and refuse to start if can't transcode.

darkdarkdragon:
On node startup try to transcode test
segment on the GPU, do not start node
if transcoding failed
Fixes #1373

Transcoded test segment on startup

- 
- 
- 

manually

Fixes #1373 

**Checklist:**


core/transcoder.go
```diff
@@ -77,6 +152,79 @@ func (nv *NvidiaTranscoder) Transcode(job string, fname string, profiles []ffmpe
...
+		return
+	}
+	if len(td.Segments) == 0 || td.Pixels == 0 {
+		res <- errors.New("Empty transcoded segment")
+		return
+	}
+	res <- nil
+}
+
+// TestNvidiaTranscoder tries to transcoded test segment on all the devices
```
kyriediculous: nit: "transcoded -> transcode"

core/transcoder.go
```diff
@@ -77,6 +152,79 @@ func (nv *NvidiaTranscoder) Transcode(job string, fname string, profiles []ffmpe
...
+	glog.Infof(fname)
+	err = ioutil.WriteFile(fname, mp4testSeg, 0644)
+	if err != nil {
+		return err
+	}
+	start := make(chan interface{})
+	res := make(chan error)
+	const sessionsToTest = 4
+	for _, device := range devices {
+		for i := 0; i < sessionsToTest; i++ {
```
j0sh: If the intent is to detect whether there's a session cap then a more straightforward way might be to transcode multiple renditions (the driver counts each rendition as one session). So, something like this:

```
p := ffmpeg.Profiles{Resolution:"16x16", Bitrate: "1k", Format: ffmpeg.FormatMP4}
_, err := tc.Transcode("", fname, []ffmpeg.Profiles{p, p, p, p})
```

I suspect this would make much of the concurrency around the `testOneNvidiaDevice` unnecessary.

core/transcoder.go
```diff
@@ -16,6 +19,78 @@ import (
 	"github.com/golang/glog"
 )
 
+var testSegment = []byte{0x1f, 0x8b, 0x8, 0x8, 0xfc, 0xdd, 0x90, 0x5e, 0x0, 0x3, 0x7a, 0x7a, 0x7a, 0x7a, 0x2e, 0x6d,
```
j0sh: It might be good to put this in a separate file with a reproducible way to generate the segment. I was able to get a much smaller (182 byte) segment like this:

```sh
#!/usr/bin/env bash

FILE=testSegment.go

echo "package core" > $FILE
echo "// Auto generated by $0 DO NOT EDIT" >> $FILE
echo "var testSegment = []byte{" >> $FILE
ffmpeg -loglevel quiet -f lavfi -i color=white:s=16x16 -vframes 1 \
    -metadata title='!' -metadata service_provider='!' -f mpegts - \
    | gzip | xxd -i | head -c -1 >> $FILE
echo "}" >> $FILE
```

We can check in both the script and the resulting `testSegment.go` and regenerate it only when needed. (So most users would not need any of these utilities installed unless they're actually regenerating the file. They are fairly standard in most distros anyway except for ffmpeg.)

cmd/livepeer/livepeer.go
```diff
@@ -242,6 +242,10 @@ func main() {
 	if *transcoder {
 		core.WorkDir = *datadir
 		if *nvidia != "" {
+			err := core.TestNvidiaTranscoder(*nvidia)
+			if err != nil {
+				panic(fmt.Errorf("Unable to transcode using nividia gpu=%s err=%v", *nvidia, err))
```
j0sh: `glog.Fatalf` is what we use elsewhere in this file

cmd/livepeer/livepeer.go
```diff
@@ -242,6 +242,10 @@ func main() {
 	if *transcoder {
 		core.WorkDir = *datadir
 		if *nvidia != "" {
+			err := core.TestNvidiaTranscoder(*nvidia)
```
j0sh: Can we make this optional via flag? (But default to true.) Something like, `-testTranscoder`

j0sh (commented):
Nice! I know this is still a WIP but some quick comments.

core/transcoder.go
```diff
@@ -77,6 +152,79 @@ func (nv *NvidiaTranscoder) Transcode(job string, fname string, profiles []ffmpe
...
+		return err
+	}
+	tmp, err := ioutil.TempDir("", "")
+	if err != nil {
+		return err
+	}
+	if true {
+		defer os.RemoveAll(tmp)
+	}
+	fname := path.Join(tmp, "testseg.mp4")
```
j0sh: We have a `WorkDir`, might be more consistent to use that. Also, I believe `filepath` is recommended when dealing with files, so:
 
```
fname = filepath.Join(WorkDir, "test.tempfile")
defer os.Remove(fname)
```

core/transcoder.go
darkdarkdragon: But I thought we still can't transcode one-frame segments? That's why I've made segment with 12 frames - just to be sure it will transcode successfully.

j0sh: https://github.com/livepeer/lpms/issues/168 is being worked on by @jailuthra so that will be resolved sooner rather than later.

Do we actually need to complete a transcode though? Wouldn't simply attempting to open the GPU be enough to trigger a failure? This would detect both invalid devices and hitting the session cap; these cases wouldn't actually go all the way to a complete transcode.

core/test_segment.sh
```diff
@@ -0,0 +1,19 @@
...
+GOFILE=$FILE.go
+
+
+echo "package core" > $GOFILE
+echo "// Auto generated by $0 DO NOT EDIT" >> $GOFILE
+echo "var testSegment = []byte{" >> $GOFILE
+# can't pipe .mp4 file - need seekable output
+ffmpeg -loglevel quiet -f lavfi -i color=white:s=48x48 -vframes 5 \
+    -g 2 -pix_fmt yuv420p -profile:v main -preset:v slow \
+    -metadata title='!' -metadata service_provider='!' -f mp4 $VIDEOFILE
```
j0sh: MP4 produces much larger outputs than mpegts because of all the extra metadata that's written out. I'd suggest staying with ts files for the test cases.

darkdarkdragon: .ts file generated as you suggested results in segment violation in LPMS's C code (not found where yet, need to compile on Linux machine with GPU)
trying to generate .ts file using same command as I'm using for .mp4 but with .ts output results in `Undefined constant or missing '(' in 'main'` error from ffmpeg

core/transcoder.go
darkdarkdragon: Thanks, fixed

core/transcoder.go
darkdarkdragon: > Wouldn't simply attempting to open the GPU be enough to trigger a failure? 

Do we have LPMS API to just try 'open the GPU'? 
But anyway, I think it is good to be sure that we really can transcode, not just open GPU.

core/test_segment.sh
j0sh: Heh, dumb mistake there on my part - h264 is not being generated with that script. Nice find though, responded to https://github.com/livepeer/lpms/issues/185 about it.

Curious, are we still able to decode h264 elementary streams in go-livepeer without adding it to the list of demuxers? If so, this would generate an even smaller file - 61 bytes on my system:

```
ffmpeg -f lavfi -i color=white:s=16x16 -vframes 1  -c:v libx264 -f mpegts - | \
ffmpeg -f mpegts -i pipe:0 -c:v copy -bsf:v 'filter_units=pass_types=1-5|7-15' -f h264 segment.h264
```

It's so well compressed that gzipping actually makes this larger. That means no need to unzip within the code.

If go-livepeer can't demux that file, then we can just use mpegts. Would rather minimize the number of formats on the list.

```
ffmpeg -f lavfi -i color=white:s=16x16 -vframes 1  -c:v libx264 -f mpegts - | \
ffmpeg -f mpegts -i pipe:0 -c:v copy -bsf:v 'filter_units=pass_types=1-5|7-15' -f mpegts -metadata title='!' -metadata service_provider='! - | gzip > segment.ts.gz
```

gzip'd, this yields a 186 byte file. This can be further piped into `xxd`  without any intermediate files hanging around.

The rough equivalent for mp4 is around 464 bytes gzip'd, and it's not pipeable, so it also needs an annoying intermediate file:

```
ffmpeg -f lavfi -i color=white:s=16x16 -vframes 1  -c:v libx264 -f mpegts - | \
ffmpeg -f mpegts -i pipe:0 -c:v copy -bsf:v 'filter_units=pass_types=1-5|7-15' segment.mp4
gzip segment.mp4
```

core/transcoder.go
```diff
@@ -77,6 +82,44 @@ func (nv *NvidiaTranscoder) Transcode(job string, fname string, profiles []ffmpe
...
+	err = ioutil.WriteFile(fname, mp4testSeg, 0644)
+	if err != nil {
+		return err
+	}
+	defer os.Remove(fname)
+	for _, device := range devices {
+		t1 := NewNvidiaTranscoder(device)
+		p := ffmpeg.VideoProfile{Resolution: "48x48", Bitrate: "1k", Format: ffmpeg.FormatMP4}
+		profiles := []ffmpeg.VideoProfile{p, p, p, p}
+		profiles[0].Format = ffmpeg.FormatMP4
```
j0sh: stray?

core/transcoder.go
```diff
@@ -1,10 +1,15 @@
 package core
 
+//go:generate ./test_segment.sh
```
j0sh: We could also add this to the Makefile so it's automatically run every time `test_segment.sh` is modified. Same way we generate the protobufs automatically as needed. Don't need to introduce an explicit `go generate` build step. For example:

```make
all: core/testSegment.go <other files here>

core/testSegment.go: core/test_segment.sh
    core/test_segment.sh
```

(don't copy the above; make expects a leading tab and will break with leading spaces as I have. Dunno how to get tabs in here without taking the text box out of focus...)

darkdarkdragon: Fixed

core/transcoder.go
darkdarkdragon: Fixed

core/test_segment.sh
darkdarkdragon: >  decode h264 elementary streams

Tested - didn't work
So I've changed to .ts segment with 5 frames

core/test_segment.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+
+FILE=test_segment.go
+
+pushd core
```
j0sh: This imposes an expectation that the script is run only from the parent directory, which is kind of weird. One way to mitigate this is just to take the output filename as an argument to the script, eg

FILE="$1"

Then in the makefile:

```
core/testSegment.go: core/test_segment.sh
    core/test_segment.sh core/testSegment.go
```

Another way is to use `dirname` within the script to change the working directory to the script's base path, eg:

```
cd $(dirname "$0")
FILE=test_segment.go
```
No need for pushd / popd here; changes to the shell environment typically don't persist outside scripts.

core/test_segment.sh
```diff
@@ -0,0 +1,15 @@
...
+echo "package core" > $FILE
+echo "// Auto generated by $0 DO NOT EDIT" >> $FILE
+echo "var testSegment = []byte{" >> $FILE
+ffmpeg -f lavfi -i color=white:s=48x48 -vframes 5  -c:v libx264 -f mpegts - |  \
+  ffmpeg -f mpegts -i pipe:0 -c:v copy -bsf:v 'filter_units=pass_types=1-5|7-15' \
+  -f mpegts -metadata title='!' -metadata service_provider='!' - | gzip -9  | \
+  xxd -i | awk 'NR > 1 { print prev } { prev=$0 } END { ORS=""; print }'  >> $FILE
```
j0sh, line 11: Was there an issue with the original suggestion of `xxd -i | head -c -1` ? That's a lot shorter.

core/test_segment.sh
darkdarkdragon: Added argument

core/test_segment.sh
darkdarkdragon: `head -c -1` gives me
`head: illegal byte count -- -1` (MacOS)

j0sh: Gotcha - macOS ships with an ancient version of coreutils (usually the first thing I update on a fresh system via nixpkgs...)

j0sh:
LGTM after a rebase / squash

j0sh (approved):
🚢

PR #1446, opened 2020-04-12
Enable/disable HTTP ingest

### LINKED ISSUE #1441: Add -httpIngest flag

HTTP ingest is enabled by default, but a user should be able to choose to enable/disable it.

Add an `-httpIngest <true|false> `flag that is used as follows:

- If `-httpIngest` is false then the push endpoint will reject requests
- If `-httpIngest` is true then the push endpoint will accept requests
- The default value for `-httpIngest` is true
- If the value for `-httpAddr` is a non-localhost address AND `-authWebhookUrl` is not specified:
    - The value for `-httpIngest` is false
    - Log a message to the user saying: "HTTP ingest disabled because `-httpAddr` is publicly accessible. To enable, configure `-authWebhookUrl` or specify `httpIngest true`"

yondonfu:

This PR enables/disables HTTP ingest according to the following rules:

- HTTP ingest is enabled by default
- If the HTTP server is publicly accessible (i.e. `-httpAddr` uses a non-local host) and there is no authentication webhook URL specified then HTTP ingest is disabled
    - In this case, HTTP ingest can be enabled with any of the following:
        - Configure `-authWebhookUrl`
        - Set `-httpIngest`

See commit history for a description of each update.

Updated unit tests and tested manually.

Fixes #1441

**Checklist:**

### DIFF
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"math/big"
+	"net"
 	"net/http"
 	"net/url"
 	"os"
@@ -89,6 +90,7 @@ func main() {
 	verifierURL := flag.String("verifierUrl", "", "URL of the verifier to use")
 
 	verifierPath := flag.String("verifierPath", "", "Path to verifier shared volume")
+	httpIngest := flag.Bool("httpIngest", true, "Set to true to enable HTTP ingest")
 
 	// Transcoding:
 	orchestrator := flag.Bool("orchestrator", false, "Set to true to be an orchestrator")
@@ -141,6 +143,9 @@ func main() {
 
 	flag.Parse()
 	vFlag.Value.Set(*verbosity)
+
+	isFlagSet := make(map[string]bool)
+	flag.Visit(func(f *flag.Flag) { isFlagSet[f.Name] = true })
 
 	blockPollingTime := time.Duration(*blockPollingInterval) * time.Second
 
@@ -676,6 +681,16 @@ func main() {
 			server.AuthWebhookURL = *authWebhookURL
 		}
 
+		isLocalHTTP, err := isLocalURL("https://" + *httpAddr)
+		if err != nil {
+			glog.Errorf("Error checking for local -httpAddr: %v", err)
+			return
+		}
+		if !isFlagSet["httpIngest"] && !isLocalHTTP && server.AuthWebhookURL == "" {
+			glog.Warning("HTTP ingest is disabled because -httpAddr is publicly accessible. To enable, configure -authWebhookUrl or use the -httpIngest flag")
+			*httpIngest = false
+		}
+
 		// Set up verifier
 		if *verifierURL != "" {
 			_, err := validateURL(*verifierURL)
@@ -724,7 +739,7 @@ func main() {
 	//Create Livepeer Node
 
 	//Set up the media server
-	s := server.NewLivepeerServer(*rtmpAddr, n)
+	s := server.NewLivepeerServer(*rtmpAddr, n, *httpIngest)
 	ec := make(chan error)
 	tc := make(chan struct{})
 	wc := make(chan struct{})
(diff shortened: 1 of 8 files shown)


server/mediaserver.go
```diff
@@ -578,6 +585,11 @@ func getRTMPStreamHandler(s *LivepeerServer) func(url *url.URL) (stream.RTMPVide
 
 // HandlePush processes request for HTTP ingest
 func (s *LivepeerServer) HandlePush(w http.ResponseWriter, r *http.Request) {
+	if !s.HTTPIngest {
+		respondWith400(w, "HTTP ingest is disabled")
```
iameli: Arguably an HTTP 405 "Method Not Allowed" but idk if we observe that sort of thing at all in go-livepeer

yondonfu: Fixed in 727d98b

cmd/livepeer/livepeer.go
```diff
@@ -820,6 +834,20 @@ func validateURL(u string) (*url.URL, error) {
...
 }
 
+func isLocalURL(u string) (bool, error) {
+	uri, err := url.ParseRequestURI(u)
+	if err != nil {
+		return false, err
+	}
+
+	hostname := uri.Hostname()
+	if hostname == "127.0.0.1" || hostname == "localhost" {
```
j0sh: Technically, the entire 127.0.0.0/8 range [1] (or 127.0.0.1/32, depending on who you're talking to [2])  is reserved for loopback. In practice, it probably doesn't really matter because our default is `127.0.0.1`.

[1] RFC 6890, Section 2.2.2. Table 4, "Loopback" https://tools.ietf.org/html/rfc6890

[2] RFC 3330, Section 2  "Global and Other Specialized Address Blocks" https://tools.ietf.org/html/rfc3330

server/mediaserver.go
```diff
@@ -578,6 +585,11 @@ func getRTMPStreamHandler(s *LivepeerServer) func(url *url.URL) (stream.RTMPVide
 
 // HandlePush processes request for HTTP ingest
 func (s *LivepeerServer) HandlePush(w http.ResponseWriter, r *http.Request) {
+	if !s.HTTPIngest {
```
j0sh: Rather than checking here, should we just not register the path, and allow the framework to take care of the error handling? Eg,

https://github.com/livepeer/go-livepeer/blob/c6b06679fe4adaa204c6d0a9628e184b2fadc0b6/server/mediaserver.go#L132-L134

Changes to:

```
if lpNode.NodeType == core.BroadcasterNode && HTTPIngest {
 	opts.HttpMux.HandleFunc("/live/", ls.HandlePush) 
 } 
```

Up to you how to propagate in the `HTTPIngest` variable. Could be via a global in `mediaserver.go` or add an extra param to `NewLivepeerServer`. Slight preference for the latter, since that's less state (whether globally or in `LivepeerServer`).

j0sh (changes requested):

cmd/livepeer/livepeer.go
yondonfu: Fixed in c299a2a

server/mediaserver.go
yondonfu: Fixed in 94ee818

cmd/livepeer/livepeer.go
```diff
@@ -676,6 +678,17 @@ func main() {
 			server.AuthWebhookURL = *authWebhookURL
 		}
 
+		isLocalHTTP, err := isLocalURL("https://" + *httpAddr)
+		if err != nil {
+			glog.Errorf("Error checking for local -httpAddr: %v", err)
+			return
+		}
+		if !*httpIngest && !isLocalHTTP && server.AuthWebhookURL == "" {
```
j0sh: This is one place where default values are tricky. How can the user still disable HTTP ingest even with localhost and/or a webhook?

Does golang CLI offer a way to distinguish between "user set the flag" and "user did not set the flag, so using the default value" ?

Option types would work much better here, but golang doesn't really have a notion of that...

yondonfu: Ah yeah meant to look into this further.

> Does golang CLI offer a way to distinguish between "user set the flag" and "user did not set the flag, so using the default value" ?

46db19e: Use [flag.Visit()](https://golang.org/pkg/flag/#Visit) to loop through all set flags to populate a map that tracks set flags.

f51a50b: Add a note in the HTTP ingest docs about always disabling HTTP ingest.

Unfortunately, it doesn't look like the `flag` package exposes a more convenient way to check if a flag is set ([cobra](https://github.com/spf13/cobra), another CLI package, exposes a `Changed()` method on flags that can be used to check if the flag was set).

Also changed the default value of `-httpIngest` to true and now the value is set to false if the flag is not set and a non-local `-httpAddr` is used without an auth webhook URL.

```golang
if !isFlagSet["httpIngest"] && !isLocalHTTP & server.AuthWebhookURL == "" {
    ...
    *httpIngest = false
}
```

j0sh: Fantastic. Thanks!

j0sh (approved):
LGTM 🚢

PR #1448, opened 2020-04-14
cmd: Remove default values for -ethUrl

### LINKED ISSUE #1434: Remove default ethUrl values

Currently, the default value for the `-ethUrl` flag is an Infura URL on both Rinkeby and mainnet. This Infura URL will be sunset soon so the node should update its default values.

Remove the default value for `-ethUrl` flag and require the user to pass in the URL of their preferred ETH RPC provider.

yondonfu:

This PR removes the default values for the `-ethUrl` flag. After this PR, whenever the node connects to an on-chain network, an Ethereum node JSON-RPC URL will need to be explicitly provided using the `-ethUrl` flag.

- Remove the default values for the `-ethUrl` flag
- Only run mainnet tests in `test_args.sh` if the `MAINNET_ETH_URL` env variable is set
- Only run Rinkeby tests in `test_args.sh` if the `RINKEBY_ETH_URL` env variable is set
- Use `-ethUrl` in README commands

Values for `MAINNET_ETH_URL` and `RINKEBY_ETH_URL` are set via CircleCI.

Updated tests in `test_args.sh`.

Fixes #1434 

**Checklist:**

### DIFF
--- a/doc/ethereum.md
+++ b/doc/ethereum.md
@@ -1,0 +1,12 @@
+# Ethereum
+
+## Connecting to an Ethereum network
+
+When connecting to an Ethereum network, an Ethereum RPC provider needs to be specified via the `-ethUrl` flag.
+
+- To connect to mainnet, the node should be started with `-network mainnet` and the URL for `-ethUrl` should be for a mainnet Ethereum node.
+- To connect to Rinkeby, the node should be started with `-network rinkeby` and the URL for `-ethUrl` should be for a Rinkeby Ethereum node.
+- To connect to a private network, the node should be started with `-network <NETWORK_NAME>` (`<NETWORK_NAME>` is the name of the private network), the URL for `-ethUrl` should be for a private network Ethereum node and value for `-ethController` should be the address of the Controller contract deployed on the private network.
+- To connect to an off-chain network, the node should be started without the `-network` flag (the default value is `offchain`). The `-ethUrl` and the `-ethController` flags are unnecessary.
+
+See [this guide](https://livepeer.readthedocs.io/en/latest/quickstart.html#connecting-to-an-ethereum-node) for instructions on obtaining a URL that be used with the `-ethUrl` flag.
--- a/README.md
+++ b/README.md
@@ -45,7 +45,8 @@ everything from scratch.
 ### Quick start
 - Make sure you have successfully gone through the steps in 'Installing Livepeer' and 'Additional Dependencies'.
 
-- Run `./livepeer -broadcaster -network rinkeby`.
+- Run `./livepeer -broadcaster -network rinkeby -ethUrl <ETHEREUM_RPC_URL>`.
+  * `<ETHEREUM_RPC_URL>` is the JSON-RPC URL of an Ethereum node
 
 - Run `./livepeer_cli`.
   * You should see a wizard launch in the command line.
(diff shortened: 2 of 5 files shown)


j0sh:
Curious, will users have a convenient way to discover or acquire an Ethereum JSON-RPC endpoint? Should that be documented somewhere? This is going to require that mainnet users upgrading to the next release make changes to their CLI flags, right?

yondonfu:
Users will still be able to use a hosted node service like Infura or [Alchemy](https://alchemyapi.io/) if they don't want to run their own node they'll just have to sign up for an account themselves.

livepeer/docs#58 updates RTD with a section on connecting either to hosted node service or a self-hosted node. Side note: I was thinking about whether this information should go into this repo or in the RTD and decided to update RTD since some of that information was already there. Would be interested in just rendering go-livepeer specific docs from this repo at some point and having RTD linking to the "go-livepeer guide".

> This is going to require that mainnet users upgrading to the next release make changes to their CLI flags, right?

Yeah. We're removing the default Infura endpoints because we're going to only allow requests from the explorer soon for those endpoints (I believe this is in response to updates to Infura's pricing plan).

yondonfu:
70c2315: Add development docs with notes on running tests and specifying mainnet/Rinkeby Ethereum node JSON-RPC API URLs in order to run certain tests.

j0sh:
> Side note: I was thinking about whether this information should go into this repo or in the RTD and decided to update RTD

Maybe also include a link to the RTD docs for those trying to figure out how to obtain an Ethereum JSON-RPC URL?

I know this has been forewarned a few times on Discord, the forums, etc, but I'd also make sure this change is prominently called out in the release notes and announcement as something that requires user action in order to complete the upgrade. Also doesn't hurt to explain the reason for it, eg Infura pricing, so users feel a little more informed and it seems a little less arbitrary, because registering with a RPC service or setting up a node is going to be a point of friction.

kyriediculous (approved):
LGTM 👍

yondonfu:
> Maybe also include a link to the RTD docs for those trying to figure out how to obtain an Ethereum JSON-RPC URL?

Added in c808bb4. The link to RTD doesn't actually go to the relevant section, but this will be fixed once livepeer/docs#58 is merged.

> I'd also make sure this change is prominently called out in the release notes and announcement as something that requires user action in order to complete the upgrade. Also doesn't hurt to explain the reason for it, eg Infura pricing, so users feel a little more informed and it seems a little less arbitrary, because registering with a RPC service or setting up a node is going to be a point of friction.

👍 Definitely. Planning on calling this out in the release notes + Discord announcement.

doc/ethereum.md
```diff
@@ -0,0 +1,12 @@
+# Ethereum
+
+## Connecting to an Ethereum network
+
+When connecting to an Ethereum network, an ETH RPC provider needs to be specified via the `-ethUrl` flag.
```
kyriediculous: nit: could we replace "ETH" with "Ethereum" since ETH is more commonly used for the cryptocurrency ?

kyriediculous (changes requested):

doc/ethereum.md
yondonfu: Fixed in ef463ef

j0sh (approved):
LGTM after a fixup

kyriediculous (approved):

kyriediculous:
Looks good, let's squash the fixup commit 👍

PR #1461, opened 2020-04-21
cmd,eth,server: add voting to CLI

### LINKED ISSUE #1447: Add voting support in livepeer_cli

Orchestrators should have a way to vote in polls without exporting their keys to a web based wallet.

A possible user flow could be: 

1. Go to proposal in explorer
2. Click "Vote using CLI"
3. Step-by-step modal opens
4. Copy-paste contract address
5. CLI prompts user for Poll contract address
6. CLI shows voting options
7. CLI prompts option ID
8. CLI asks confirmation 
    - yes > send tx
   -  no > step 7

Add support for voting in polls using `livepeer_cli`.

kyriediculous:
This PR adds the option to vote from the CLI (for orchestrators) so that they don't have to export their keys. 

- generate poll contract abi bindings 
- create a `voteHandler` for the new CLI server `/vote` endpoint
- add a `Vote()` method to `LivepeerEthClient`
- Add an option to the CLI to vote 

Added & ran unit tests
Ran workflow in CLI (except for testing with a live poll on rinkeby) 

Fixes #1447 

**Checklist:**

**Example**
```
What would you like to do? (default = stats)
1. Get node status
2. View protocol parameters
3. List registered orchestrators
4. Invoke "initialize round"
5. Invoke "bond"
6. Invoke "unbond"
7. Invoke "rebond"
10. Invoke "claim" (for rewards and fees)
11. Invoke "transfer" (LPT)
12. Invoke "reward"
13. Invoke multi-step "become an orchestrator"
14. Set orchestrator config
15. Set Eth gas price
16. Get test LPT
17. Get test ETH
18. Sign a message
19. Vote on a poll
> 19
Enter the contract address for the poll you want to vote on> 0xf4e8Ef0763BCB2B1aF693F5970a00050a6aC7E1B
Identifier	Voting Options
0		Yes
1		No
2		Abstain
Enter the ID of the option you want to vote for> 0
Are you sure you want to vote "Yes" ? (y/n)> n
Identifier	Voting Options
0		Yes
1		No
2		Abstain
Enter the ID of the option you want to vote for> 2
Are you sure you want to vote "Abstain" ? (y/n)> n
Identifier	Voting Options
0		Yes
1		No
2		Abstain
Enter the ID of the option you want to vote for> 1
Are you sure you want to vote "No" ? (y/n)> y

 unable to submit vote transaction err=no contract code at given address
```

```
Enter the contract address for the poll you want to vote on> 0xbf790e51fa21e1515cece96975b3505350b20083
Identifier	Voting Options
0		Yes
1		No
2		Abstain
Enter the ID of the option you want to vote for> 1
Are you sure you want to vote "No" ? (y/n)> y

 success
```

```
I0422 03:30:44.191809   43457 backend.go:96]

Invoking transaction: "vote". Inputs: "_choiceID: 1"  Hash: "0xf6957c190f1f16fc2ca4a93846903eb435c5e08fa7f6f40b6e159aab6d74905f".

**************************************************************************
```

### DIFF
--- a/cmd/livepeer_cli/livepeer_cli.go
+++ b/cmd/livepeer_cli/livepeer_cli.go
@@ -108,6 +108,7 @@ func (w *wizard) initializeOptions() []wizardOpt {
 			w.read()
 		}, testnet: true},
 		{desc: "Sign a message", invoke: w.signMessage},
+		{desc: "Vote in a poll", invoke: w.vote, orchestrator: true},
 	}
 	return options
 }
--- a/cmd/livepeer_cli/wizard_transcoder.go
+++ b/cmd/livepeer_cli/wizard_transcoder.go
@@ -4,11 +4,15 @@ import (
 	"fmt"
 	"math/big"
 	"net/url"
+	"os"
 	"strconv"
 	"strings"
+	"text/tabwriter"
 
+	ethcommon "github.com/ethereum/go-ethereum/common"
 	"github.com/golang/glog"
 	lpcommon "github.com/livepeer/go-livepeer/common"
+	"github.com/livepeer/go-livepeer/eth/types"
 )
 
 const defaultRPCPort = "8935"
@@ -184,4 +188,59 @@ func (w *wizard) callReward() {
 
 	fmt.Printf("Calling reward for round %v\n", c)
 	httpGet(fmt.Sprintf("http://%v:%v/reward", w.host, w.httpPort))
+}
+
+func (w *wizard) vote() {
+	if w.offchain {
+		glog.Error("Can not vote in 'offchain' mode")
+		return
+	}
+
+	fmt.Print("Enter the contract address for the poll you want to vote in -")
+	poll := w.readStringAndValidate(func(in string) (string, error) {
+		if !ethcommon.IsHexAddress(in) {
+			return "", fmt.Errorf("invalid hex address address=%v", in)
+		}
+		return in, nil
+	})
+
+	var (
+		confirm = "n"
+		choice  = types.VoteChoice(-1)
+	)
+
+	for confirm == "n" {
+		choice = types.VoteChoice(-1)
+		w.showVoteChoices()
+
+		for {
+			fmt.Printf("Enter the ID of the choice you want to vote for -")
+			choice = types.VoteChoice(w.readInt())
+			if choice.IsValid() {
+				break
+			}
+			fmt.Println("Must enter a valid ID")
+		}
+
+		fmt.Printf("Are you sure you want to vote \"%v\"? (y/n) -", choice.String())
+		confirm = w.readStringYesOrNo()
+	}
+
+	data := url.Values{
+		"poll":     {poll},
+		"choiceID": {fmt.Sprintf("%v", int(choice))},
+	}
+
+	result := httpPostWithParams(fmt.Sprintf("http://%v:%v/vote", w.host, w.httpPort), data)
+
+	fmt.Printf("\n%v\n", result)
+}
+
+func (w *wizard) showVoteChoices() {
+	wtr := tabwriter.NewWriter(os.Stdout, 0, 8, 1, '\t', 0)
+	fmt.Fprintln(wtr, "Identifier\tVoting Choices")
+	for _, choice := range types.VoteChoices {
+		fmt.Fprintf(wtr, "%v\t%v\n", int(choice), choice.String())
+	}
+	wtr.Flush()
 }
--- a/eth/client.go
+++ b/eth/client.go
@@ -11,7 +11,7 @@ package eth
 //go:generate abigen --abi protocol/abi/RoundsManager.abi --pkg contracts --type RoundsManager --out contracts/roundsManager.go
 //go:generate abigen --abi protocol/abi/Minter.abi --pkg contracts --type Minter --out contracts/minter.go
 //go:generate abigen --abi protocol/abi/LivepeerTokenFaucet.abi --pkg contracts --type LivepeerTokenFaucet --out contracts/livepeerTokenFaucet.go
-
+//go:generate abigen --abi protocol/abi/Poll.abi --pkg contracts --type Poll --out contracts/poll.go
 import (
 	"context"
 	"errors"
@@ -106,6 +106,9 @@ type LivepeerEthClient interface {
 	InflationChange() (*big.Int, error)
 	TargetBondingRate() (*big.Int, error)
 	Paused() (bool, error)
+
+	// Governance
+	Vote(ethcommon.Address, *big.Int) (*types.Transaction, error)
 
 	// Helpers
 	ContractAddresses() map[string]ethcommon.Address
@@ -718,6 +721,21 @@ func (c *client) TranscoderPool() ([]*lpTypes.Transcoder, error) {
 	}
 
 	return transcoders, nil
+}
+
+func (c *client) Vote(pollAddr ethcommon.Address, choiceID *big.Int) (*types.Transaction, error) {
+	poll, err := contracts.NewPoll(pollAddr, c.backend)
+	if err != nil {
+		return nil, err
+	}
+
+	gl, gp := c.GetGasInfo()
+	opts, err := c.accountManager.CreateTransactOpts(gl, gp)
+	if err != nil {
+		return nil, err
+	}
+
+	return poll.Vote(opts, choiceID)
 }
 
 // Helpers
--- a/eth/types/contracts.go
+++ b/eth/types/contracts.go
@@ -98,3 +98,27 @@ type ProtocolParameters struct {
 	TotalSupply                   *big.Int
 	Paused                        bool
 }
+
+type VoteChoice int
+
+const (
+	Yes VoteChoice = iota
+	No
+)
+
+var VoteChoices = []VoteChoice{Yes, No}
+
+func (v VoteChoice) String() string {
+	switch v {
+	case Yes:
+		return "Yes"
+	case No:
+		return "No"
+	default:
+		return ""
+	}
+}
+
+func (v VoteChoice) IsValid() bool {
+	return v == Yes || v == No
+}
--- a/server/handlers.go
+++ b/server/handlers.go
@@ -6,9 +6,11 @@ import (
 	"math/big"
 	"net/http"
 
+	ethcommon "github.com/ethereum/go-ethereum/common"
 	"github.com/golang/glog"
 	"github.com/livepeer/go-livepeer/common"
 	"github.com/livepeer/go-livepeer/eth"
+	"github.com/livepeer/go-livepeer/eth/types"
 	"github.com/livepeer/go-livepeer/pm"
 )
 
@@ -307,5 +309,58 @@ func signMessageHandler(client eth.LivepeerEthClient) http.Handler {
 
 		w.WriteHeader(http.StatusOK)
 		w.Write(signed)
+	})
+}
+
+func voteHandler(client eth.LivepeerEthClient) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if client == nil {
+			respondWith500(w, "missing ETH client")
+			return
+		}
+
+		poll := r.FormValue("poll")
+		if poll == "" {
+			respondWith500(w, "missing poll contract address")
+			return
+		}
+		if !ethcommon.IsHexAddress(poll) {
+			respondWith500(w, "invalid poll contract address")
+			return
+		}
+
+		choiceStr := r.FormValue("choiceID")
+		if choiceStr == "" {
+			respondWith500(w, "missing choiceID")
+			return
+		}
+
+		choiceID, ok := new(big.Int).SetString(choiceStr, 10)
+		if !ok {
+			respondWith500(w, "choiceID is not a valid integer value")
+			return
+		}
+		if !types.VoteChoice(int(choiceID.Int64())).IsValid() {
+			respondWith500(w, "invalid choiceID")
+			return
+		}
+
+		// submit tx
+		tx, err := client.Vote(
+			ethcommon.HexToAddress(poll),
+			choiceID,
+		)
+		if err != nil {
+			respondWith500(w, fmt.Sprintf("unable to submit vote transaction err=%v", err))
+			return
+		}
+
+		if err := client.CheckTx(tx); err != nil {
+			respondWith500(w, fmt.Sprintf("unable to mine vote transaction err=%v", err))
+			return
+		}
+
+		w.WriteHeader(http.StatusOK)
+		w.Write([]byte("vote success"))
 	})
 }
--- a/eth/backend.go
+++ b/eth/backend.go
@@ -25,6 +25,7 @@ var abis = []string{
 	contracts.RoundsManagerABI,
 	contracts.ServiceRegistryABI,
 	contracts.TicketBrokerABI,
+	contracts.PollABI,
 }
 
 type Backend interface {
--- a/eth/contracts/poll.go
+++ b/eth/contracts/poll.go
@@ -1,0 +1,385 @@
+// Code generated - DO NOT EDIT.
+// This file is a generated binding and any manual changes will be lost.
+
+package contracts
+
+import (
+	"math/big"
+	"strings"
+
+	ethereum "github.com/ethereum/go-ethereum"
+	"github.com/ethereum/go-ethereum/accounts/abi"
+	"github.com/ethereum/go-ethereum/accounts/abi/bind"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/event"
+)
+
+// Reference imports to suppress errors if they are not otherwise used.
+var (
+	_ = big.NewInt
+	_ = strings.NewReader
+	_ = ethereum.NotFound
+	_ = abi.U256
+	_ = bind.Bind
+	_ = common.Big1
+	_ = types.BloomLookup
+	_ = event.NewSubscription
+)
+
+// PollABI is the input ABI used to generate the binding from.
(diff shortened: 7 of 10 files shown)


cmd/livepeer_cli/wizard_transcoder.go
```diff
@@ -185,3 +194,68 @@ func (w *wizard) callReward() {
...
+	result := httpPostWithParams(fmt.Sprintf("http://%v:%v/vote", w.host, w.httpPort), data)
+
+	fmt.Printf("\n %v \n", result)
+}
+
+func (w *wizard) showVoteOptions() map[int]string {
+	opts := make([]string, 3)
+	for i, opt := range voteOptions {
+		opts[i] = opt
+	}
```
kyriediculous: This is kind of hacky but it ensures ordering

cmd/livepeer_cli/livepeer_cli.go
```diff
@@ -108,6 +108,7 @@ func (w *wizard) initializeOptions() []wizardOpt {
 			w.read()
 		}, testnet: true},
 		{desc: "Sign a message", invoke: w.signMessage},
+		{desc: "Vote on a poll", invoke: w.vote, orchestrator: true},
```
yondonfu: Should this be "Vote in a poll" since a user would vote in a poll as opposed to on a poll?

eth/client.go
```diff
@@ -720,6 +723,26 @@ func (c *client) TranscoderPool() ([]*lpTypes.Transcoder, error) {
...
+	if err != nil {
+		return err
+	}
+
+	tx, err := poll.Vote(opts, choiceID)
+	if err != nil {
+		return err
+	}
+
+	return c.CheckTx(tx)
```
yondonfu: Most of the other tx submitting methods pass the responsibility of waiting for a tx receipt using `CheckTx()` to the caller so that the caller can have access to the returned `Transaction` struct. Would be good to do the same here for consistency and so callers can access tx information if needed.

server/handlers.go
```diff
@@ -309,3 +310,27 @@ func signMessageHandler(client eth.LivepeerEthClient) http.Handler {
...
+
+func voteHandler(client eth.LivepeerEthClient) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if client == nil {
+			respondWith500(w, "missing ETH client")
+			return
+		}
+
+		poll := r.FormValue("poll")
+		choiceID, _ := new(big.Int).SetString(r.FormValue("choiceID"), 10)
```
yondonfu: Let's check the boolean returned by `SetString()`.

yondonfu: I think we should check that the `choiceID` value is one of the valid choices. `livepeer_cli` will already do this validation, but this endpoint could be directly used to submit a vote tx so adding validation here would be helpful for avoiding accidentally submitting txs with invalid choices. `livepeer_cli` and this handler could even rely on a shared definition of valid choices so they don't have to be defined in separate places.

server/handlers.go
```diff
@@ -309,3 +310,27 @@ func signMessageHandler(client eth.LivepeerEthClient) http.Handler {
...
+		if err := client.Vote(
+			ethcommon.HexToAddress(poll),
+			choiceID,
+		); err != nil {
+			respondWith500(w, fmt.Sprintf("unable to submit vote transaction err=%v", err))
+			return
+		}
+
+		w.WriteHeader(http.StatusOK)
+		w.Write([]byte("success"))
```
yondonfu: "vote success" to be more specific?

cmd/livepeer_cli/wizard_transcoder.go
```diff
@@ -185,3 +194,68 @@ func (w *wizard) callReward() {
...
 	httpGet(fmt.Sprintf("http://%v:%v/reward", w.host, w.httpPort))
 }
+
+func (w *wizard) vote() {
+	if w.offchain {
+		glog.Error("Can not vote in 'offchain' mode")
+		return
+	}
+
+	fmt.Print("Enter the contract address for the poll you want to vote on")
```
yondonfu: "vote on" -> "vote in"?

cmd/livepeer_cli/wizard_transcoder.go
```diff
@@ -185,3 +194,68 @@ func (w *wizard) callReward() {
...
+	poll := w.readStringAndValidate(func(in string) (string, error) {
+		if !ethcommon.IsHexAddress(in) {
+			return "", fmt.Errorf("invalid hex address address=%v", in)
+		}
+		return in, nil
+	})
+
+	var (
+		confirm = "n"
+		choice  = 99
```
yondonfu: Could use -1 as the default here as it would never be a valid choice ID.

cmd/livepeer_cli/wizard_transcoder.go
```diff
@@ -185,3 +194,68 @@ func (w *wizard) callReward() {
...
+
+	fmt.Printf("\n %v \n", result)
+}
+
+func (w *wizard) showVoteOptions() map[int]string {
+	opts := make([]string, 3)
+	for i, opt := range voteOptions {
+		opts[i] = opt
+	}
+	wtr := tabwriter.NewWriter(os.Stdout, 0, 8, 1, '\t', 0)
```
yondonfu, line 240: We already use the [tablewriter](https://github.com/livepeer/go-livepeer/blob/a55045b290eed1f69780527a6b71e271e61ea31f/cmd/livepeer_cli/wizard_bond.go#L99) package elsewhere to render text tables - could be helpful to use that package here since it should automatically handle spacing adjustments in the table.

cmd/livepeer_cli/wizard_transcoder.go
yondonfu: An alternative could be to use a slice to list the options and represent the option as an enum type:

```golang
const voteOptions = []VoteOption{VoteOption.Yes, ...}

func (v VoteOption) String() string {
    switch v {
        case VoteOption.Yes:
            return "Yes"
        ...
    }
}
```

An invalid vote option would return an empty string for the `String()` call - could event wrap this into `ValidVoteOption()` function for improved readability.

cmd/livepeer_cli/wizard_transcoder.go
```diff
@@ -4,15 +4,24 @@ import (
...
+	"text/tabwriter"
 
+	ethcommon "github.com/ethereum/go-ethereum/common"
 	"github.com/golang/glog"
 	lpcommon "github.com/livepeer/go-livepeer/common"
 )
 
 const defaultRPCPort = "8935"
 
+var voteOptions = map[int]string{
```
yondonfu: Should this be `voteChoices` to match use of "choice" in `choiceID`?

yondonfu (changes requested):

cmd/livepeer_cli/wizard_transcoder.go
kyriediculous: I felt like a table might have been overkill ,  we use `tabwriter` to display video profiles in a similar workflow so I felt that would be sufficient. WDYT?

server/handlers.go
kyriediculous: if we validate the form value first then will always result in `true` because it's guaranteed to be an index of the enum

kyriediculous:
Addressed the change requests in their respective fixup commits 


fixup! cmd: add voting option to the CLI
00b49d6

fixup! server: CLI server voting handler
70c8b1c

fixup! eth: abigen for Poll contract
c385a51

server/handlers.go
kyriediculous: I added the following checks and unit tests

- missing form params check (both poll and choiceID)
- invalid poll address
- invalid choice ID

adamsoffer:
@kyriediculous @yondonfu do we want to include `abstain` as a choice in this first iteration? It's not currently accounted for in the spec defined in [LIP-16](https://github.com/livepeer/LIPs/blob/master/LIPs/LIP-16.md#off-chain-indexer), the subgraph, or the UI.

cmd/livepeer_cli/wizard_transcoder.go
```diff
@@ -185,3 +189,69 @@ func (w *wizard) callReward() {
...
+
+	fmt.Println(data)
+
+	result := httpPostWithParams(fmt.Sprintf("http://%v:%v/vote", w.host, w.httpPort), data)
+
+	fmt.Printf("\n %v \n", result)
+}
+
+func (w *wizard) showVoteChoices() map[int]string {
+	choices := make(map[int]string)
```
yondonfu: Doesn't look like we need to return this map anymore.

cmd/livepeer_cli/wizard_transcoder.go
yondonfu: Sounds good to me 👍

cmd/livepeer_cli/wizard_transcoder.go
```diff
@@ -185,3 +189,69 @@ func (w *wizard) callReward() {
...
+
+	wtr := tabwriter.NewWriter(os.Stdout, 0, 8, 1, '\t', 0)
+	fmt.Fprintln(wtr, "Identifier\tVoting Options")
```
yondonfu: "Voting Options" -> "Voting Choices" to match the "choice" terminology.

cmd/livepeer_cli/wizard_transcoder.go
```diff
@@ -185,3 +189,69 @@ func (w *wizard) callReward() {
...
+	for idx, choice := range types.VoteChoices {
+		fmt.Fprintf(wtr, "%v\t%v\n", idx, choice.String())
```
yondonfu: I suggest using the actual choice value here instead of the index i.e.

```golang
fmt.Fprintf(wtr, "%v\t%v\n", int(choice), choice.String())
```

Currently, the choice value equals the index in `VoteChoice`. But, using the actual choice value here ensures that if the choice value happened to not equal the index in the future (i.e. maybe if a choice in the middle of the slice is removed) that the correct value will be printed.

cmd/livepeer_cli/wizard_transcoder.go
```diff
@@ -185,3 +189,69 @@ func (w *wizard) callReward() {
...
+		for ok := choice.IsValid(); !ok; {
+			fmt.Printf("Enter the ID of the option you want to vote for")
+			choice = types.VoteChoice(w.readInt())
+			if ok := choice.IsValid(); ok {
+				break
+			}
+			fmt.Println("Must enter a valid ID")
+		}
+
+		fmt.Printf("Are you sure you want to vote \"%v\" ? (y/n)", choice.String())
```
yondonfu: Looks like there is an extra space before the question mark here.

cmd/livepeer_cli/wizard_transcoder.go
```diff
@@ -185,3 +189,69 @@ func (w *wizard) callReward() {
...
+	)
+
+	for confirm == "n" {
+		choice = types.VoteChoice(-1)
+		choices := w.showVoteChoices()
+		if choices == nil {
+			return
+		}
+
+		for ok := choice.IsValid(); !ok; {
```
yondonfu: Could this be simplified a bit to avoid doing 2 `IsValid()` checks in the loop? i.e.

```golang
for {
    fmt.Printf("Enter the ID...")
    choice = types.VoteChoice(w.readInt())
    if !choice.IsValid() {
        break
    }
    fmt.Println(...)
}
```

cmd/livepeer_cli/wizard_transcoder.go
```diff
@@ -185,3 +189,69 @@ func (w *wizard) callReward() {
...
 	httpGet(fmt.Sprintf("http://%v:%v/reward", w.host, w.httpPort))
 }
+
+func (w *wizard) vote() {
+	if w.offchain {
+		glog.Error("Can not vote in 'offchain' mode")
+		return
+	}
+
+	fmt.Print("Enter the contract address for the poll you want to vote in")
```
yondonfu: The the other wizard prompts look like this:

```
Enter foo bar - >
```

Perhaps we should follow that format here by adding a dash to the end of user prompts i.e. "... vote in - "? Same comment applies for the other user prompts in this file.

cmd/livepeer_cli/wizard_transcoder.go
```diff
@@ -185,3 +189,69 @@ func (w *wizard) callReward() {
...
+		fmt.Printf("Are you sure you want to vote \"%v\" ? (y/n)", choice.String())
+		confirm = w.readStringYesOrNo()
+	}
+
+	data := url.Values{
+		"poll":     {poll},
+		"choiceID": {fmt.Sprintf("%v", int(choice))},
+	}
+
+	fmt.Println(data)
```
yondonfu: Seems like this is an extraneous print statement?

cmd/livepeer_cli/wizard_transcoder.go
```diff
@@ -185,3 +189,69 @@ func (w *wizard) callReward() {
...
+
+	for confirm == "n" {
+		choice = types.VoteChoice(-1)
+		choices := w.showVoteChoices()
+		if choices == nil {
+			return
+		}
+
+		for ok := choice.IsValid(); !ok; {
+			fmt.Printf("Enter the ID of the option you want to vote for")
```
yondonfu: "option" -> "choice"

eth/types/contracts.go
```diff
@@ -98,3 +98,33 @@ type ProtocolParameters struct {
...
+		return "No"
+	case Abstain:
+		return "Abstain"
+	default:
+		return ""
+	}
+}
+
+func (v VoteChoice) IsValid() bool {
+	if v != Yes && v != No && v != Abstain {
```
yondonfu: Could the conditional be omitted?

```golang
return v == Yes || v == No || v == Abstain
```

server/handlers.go
```diff
@@ -309,3 +311,47 @@ func signMessageHandler(client eth.LivepeerEthClient) http.Handler {
...
+		}
+
+		choiceStr := r.FormValue("choiceID")
+		if choiceStr == "" {
+			respondWith500(w, "missing choiceID")
+			return
+		}
+
+		choiceID, _ := new(big.Int).SetString(choiceStr, 10)
+		fmt.Println(choiceID)
```
yondonfu: Extraneous print statement?

server/handlers.go
```diff
@@ -309,3 +311,47 @@ func signMessageHandler(client eth.LivepeerEthClient) http.Handler {
...
+		if ok := types.VoteChoice(int(choiceID.Int64())).IsValid(); !ok {
```
yondonfu: Could be simplified to:

```golang
if !types.VoteChoice(int(choiceID.Int64())).IsValid() { ... }
```

server/handlers.go
```diff
@@ -309,3 +311,47 @@ func signMessageHandler(client eth.LivepeerEthClient) http.Handler {
...
+			return
+		}
+
+		choiceStr := r.FormValue("choiceID")
+		if choiceStr == "" {
+			respondWith500(w, "missing choiceID")
+			return
+		}
+
+		choiceID, _ := new(big.Int).SetString(choiceStr, 10)
```
yondonfu: We should check the boolean here because if `choiceStr` is not a valid integer then `choiceID` will be nil and L340 will trigger a nil pointer error.

eth/client.go
yondonfu: Bump on this comment.

yondonfu (changes requested):

cmd/livepeer_cli/wizard_transcoder.go
kyriediculous: oh yeah sorry I forgot to remove this

cmd/livepeer_cli/wizard_transcoder.go
kyriediculous: So in my mind you choose an option, only after you have chosen does the option become a choice but I don't wanna discuss semantics here so I'm fine renaming everything to choice. It just felt like weird use of the word here to me.

eth/client.go
kyriediculous: fixed in e5fa5ed51e4917427539616e7cc6f300aa6adc18

cmd/livepeer_cli/wizard_transcoder.go
kyriediculous: fixed in 4ecc00b

cmd/livepeer_cli/wizard_transcoder.go
kyriediculous: fixed in 4ecc00b

cmd/livepeer_cli/wizard_transcoder.go
kyriediculous: fixed in 4ecc00b

cmd/livepeer_cli/wizard_transcoder.go
kyriediculous: fixed in 4ecc00b

cmd/livepeer_cli/wizard_transcoder.go
kyriediculous: fixed in 4ecc00b

cmd/livepeer_cli/wizard_transcoder.go
kyriediculous: fixed in 4ecc00b

cmd/livepeer_cli/wizard_transcoder.go
kyriediculous: fixed in 4ecc00b

cmd/livepeer_cli/wizard_transcoder.go
kyriediculous: fixed in 4ecc00b

server/handlers.go
kyriediculous: fixed in 2857efe

server/handlers.go
kyriediculous: fixed in 2857efe

server/handlers.go
kyriediculous: fixed in 2857efe

cmd/livepeer_cli/wizard_transcoder.go
```diff
@@ -185,3 +189,58 @@ func (w *wizard) callReward() {
...
+	)
+
+	for confirm == "n" {
+		choice = types.VoteChoice(-1)
+		w.showVoteChoices()
+
+		for {
+			fmt.Printf("Enter the ID of the choice you want to vote for -")
+			choice = types.VoteChoice(w.readInt())
+			if ok := choice.IsValid(); ok {
```
yondonfu: Can be shortened to `if choice.IsValid() { ... }`

cmd/livepeer_cli/wizard_transcoder.go
```diff
@@ -185,3 +189,58 @@ func (w *wizard) callReward() {
...
+	}
+
+	data := url.Values{
+		"poll":     {poll},
+		"choiceID": {fmt.Sprintf("%v", int(choice))},
+	}
+
+	result := httpPostWithParams(fmt.Sprintf("http://%v:%v/vote", w.host, w.httpPort), data)
+
+	fmt.Printf("\n %v \n", result)
```
yondonfu: Don't think we need the leading and trailing space as it will show up as:

```
 vote success
```

rather than 

```
vote success
```

yondonfu (commented):
Left 2 minor comments.

Other than that, changes look good so feel free to address the comments when rebasing.

kyriediculous:
Addressed the last comments and rebased

yondonfu:
Looks like CI is failing:

```
eth/watchers/orchestratorwatcher_test.go:25:35: cannot use lpEth (type *eth.StubClient) as type eth.LivepeerEthClient in argument to NewOrchestratorWatcher:
	*eth.StubClient does not implement eth.LivepeerEthClient (wrong type for Vote method)
		have Vote("github.com/ethereum/go-ethereum/common".Address, *big.Int) error
		want Vote("github.com/ethereum/go-ethereum/common".Address, *big.Int) (*"github.com/ethereum/go-ethereum/core/types".Transaction, error)
```

kyriediculous:
Apparently something went wrong in the rebase. Should be fixed.

I also added a test case for a `CheckTx()` error which should bring us to 100% coverage for the handler.

yondonfu (approved):
LGTM pending green CI

Issue #1416, opened 2020-03-11
windows build: lock to a specific golang version


iameli:
Currently, we just use whatever the latest version is that's provided by `pacman`. The Linux and Mac builds both pin to a specific golang version. We should do this in Windows as well, for consistency.

This needs to be adjusted in `./windows-bulid.ps1`.

iameli:
If I remember right, all that's necessary for this is to find the actual file downloaded by Pacman and then point it at that specific archive instead.

PR #1462, opened 2020-04-22
fix missing DLL in Windows build; add self-checks

iameli:
v0.5.6 is broken on Windows. This PR tweaks a library path to fix it. It also adds two checks to ensure the Azure Pipelines build doesn't pass unless the desired libraries are all present and `liveeer.exe -version` runs successfully outside of the MSYS2 environment.

This happened because MSYS2 upgraded some libraries and we're kinda at their mercy right now, packaging whatever libraries that MSYS2 happens to ship with its latest build. https://github.com/livepeer/go-livepeer/issues/1416 would help a little; we could perhaps follow a similar procedure to lock the versions of all the dependencies we use... sounds kinda hard. I kind of think instead we should just get a Windows build environment set up in an archive somewhere so we can just deterministically download something every time.

### DIFF
--- a/upload_build.sh
+++ b/upload_build.sh
@@ -51,9 +51,9 @@ if [[ $ARCH == "windows" ]]; then
   FILE=$BASE.zip
   ls /mingw64/bin/
   # This list was produced by `ldd livepeer.exe`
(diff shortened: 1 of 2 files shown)


kyriediculous (approved):
Thanks for being so quick on the ball !

yondonfu (approved):

iameli:
image
 
Verified working, though I had to run with `-testTranscoder=false`. I'll file a bug.

PR #1465, opened 2020-04-22
Embed mime types for some video formats

darkdarkdragon:
Embed mime types for some video formats
Needed because golang not always loads mime types
from host machine

- 
- 
- 

Manually

**Checklist:**

### DIFF
--- a/common/util.go
+++ b/common/util.go
@@ -39,6 +39,11 @@ var (
 	ErrFormatProto = fmt.Errorf("unknown VideoProfile format for protobufs")
 	ErrFormatMime  = fmt.Errorf("unknown VideoProfile format for mime type")
 	ErrFormatExt   = fmt.Errorf("unknown VideoProfile format for extension")
+
+	ext2mime = map[string]string{
+		".ts":  "video/mp2t",
+		".mp4": "video/mp4",
+	}
 )
 
 func init() {
@@ -226,6 +231,9 @@ func ProfileFormatMimeType(f ffmpeg.Format) (string, error) {
 	ext, err := ProfileFormatExtension(f)
 	if err != nil {
 		return "", err
+	}
+	if m, ok := ext2mime[ext]; ok && m != "" {
+		return m, nil
 	}
 	m := mime.TypeByExtension(ext)
 	if m == "" {


darkdarkdragon:
probably we should move `ProfileFormatMimeType` to `LPMS` because `mime.TypeByExtension` used in `LPMS` too, but we can merge that now and move it as a next step.

common/util.go
```diff
@@ -39,6 +39,24 @@ var (
 	ErrFormatProto = fmt.Errorf("unknown VideoProfile format for protobufs")
 	ErrFormatMime  = fmt.Errorf("unknown VideoProfile format for mime type")
 	ErrFormatExt   = fmt.Errorf("unknown VideoProfile format for extension")
+
+	mimeExts = []string{
```
iameli: Should this just be a `map[string]string` rather than two slices?

common/util.go
```diff
@@ -39,6 +39,24 @@ var (
...
+		".ts",
+		".mp4",
+		".mpg",
+		".ogv",
+		".mov",
+		"m3u8",
```
iameli: This one doesn't have a `.`

iameli (changes requested):

j0sh:
ProfileFormatMimeType should stay in go-livepeer, but I do think that we should add a lookup table to LPMS mapping the extension to the mime type, eg `ExtensionMimeTypes = map[string]string{..}` following the example here:

https://github.com/livepeer/lpms/blob/6e167971ec68573d2cdf47cbb0b163fbb3a981a7/ffmpeg/videoprofile.go#L76-L79

Or the lookup could convert from a `Format` to the corresponding mime type, eg `map[ffmpeg.Format]string`  - not sure which would be more useful.

j0sh (changes requested):
Can this be a lookup table in LPMS following the format of [1] rather than two separate lists here?

[1] https://github.com/livepeer/lpms/blob/master/ffmpeg/videoprofile.go#L76-L79

common/util.go
darkdarkdragon: map will require mutex for syncing, given size of the data - not worth it
actually wanted to add comment about this, but forgot :(
will add one

j0sh: Can this be a lookup table in LPMS following the format of [1] rather than two separate lists here?

[1] https://github.com/livepeer/lpms/blob/master/ffmpeg/videoprofile.go#L76-L79

j0sh: A mutex for what? We're not mutating the table

common/util.go
darkdarkdragon: fixed

common/util.go
darkdarkdragon: it just stuck in my head that every map access should be guarded...
fixed

common/util.go
```diff
@@ -227,6 +236,11 @@ func ProfileFormatMimeType(f ffmpeg.Format) (string, error) {
 	if err != nil {
 		return "", err
 	}
+	for mext, mtype := range ext2mime {
+		if mext == ext {
+			return mtype, nil
+		}
+	}
```
iameli: Can't this just be
```suggestion
	mtype, ok := ext2mime[mext]
	if ok {
		return mtype, nil
	}
```

darkdarkdragon: fixed

iameli (approved):
LGTM!

common/util.go
```diff
@@ -227,7 +236,11 @@ func ProfileFormatMimeType(f ffmpeg.Format) (string, error) {
 	if err != nil {
 		return "", err
 	}
-	m := mime.TypeByExtension(ext)
+	m := ext2mime[ext]
```
j0sh: `if m, ok := ext2mime[ext]; ok { return m, nil }`

common/util.go
```diff
@@ -39,6 +39,15 @@ var (
 	ErrFormatProto = fmt.Errorf("unknown VideoProfile format for protobufs")
 	ErrFormatMime  = fmt.Errorf("unknown VideoProfile format for mime type")
 	ErrFormatExt   = fmt.Errorf("unknown VideoProfile format for extension")
+
+	ext2mime = map[string]string{
```
j0sh, line 43: As mentioned earlier, I'd prefer we put this lookup table in LPMS but I know this is more expedient. Letting this through probably means the LPMS changes will never get done though...

common/util.go
```diff
@@ -39,6 +39,15 @@ var (
...
+		".ts":   "video/mp2t",
+		".mp4":  "video/mp4",
+		".mpg":  "video/mpeg",
+		".ogv":  "video/ogg",
+		".mov":  "video/quicktime",
+		".m3u8": "application/x-mpegurl",
```
j0sh: Can we limit ourselves to formats that are actually defined in LPMS `ffmpeg.Format` ? Meaning, just ts and mp4 for now.

j0sh (dismissed):

common/util.go
darkdarkdragon: fixed

common/util.go
darkdarkdragon: moved condition to be one the same line, but not used `ok` flag - don't see any need for that

j0sh: Zero values can be semantically awkward ("does this key actually exist but is empty?") so rather than having future developers check whether this is always safe for a given map/table, I'd rather we just be consistent throughout the codebase and use the "comma-ok" idiom.

Not that I expect we'll ever actually have an extension mapping to an empty mime type, but the consistency feels important from a code quality / robustness / readability perspective.

darkdarkdragon: > "does this key actually exist but is empty?"

It this particular place it doesn't matter - empty mime type is error here.
But I've also added check for key existence.

common/util.go
```diff
@@ -232,7 +232,7 @@ func ProfileFormatMimeType(f ffmpeg.Format) (string, error) {
 	if err != nil {
 		return "", err
 	}
-	if m := ext2mime[ext]; m != "" {
+	if m, ok := ext2mime[ext]; ok && m != "" {
```
j0sh, line 235: The extra `m != ""` check isn't really needed but harmless I suppose.

This is OK to merge after a rebase / fixup.

PR #1466, opened 2020-04-22
Return 'name' parameter in 'Content-Type' header

### LINKED ISSUE #1463: add `file` field to `content-type` field of multipart renditions
When you create a multipart body with `mpack`, the `content-type` and `content-disposition` headers look like this:

`Content-Type: application/octet-stream; name="example_file"`
`Content-Disposition: inline; filename="example_file"`

And then when you unpack it with `munpack` it uses the `content-type` header to determine the name of the output files. Currently, we add the filename to the `Content-Disposition` header but not the `Content-Type` header. Presumably other multipart parsers use this field in this way as well. Let's support it? Seems harmless.

darkdarkdragon:
Return 'name' parameter in 'Content-Type' header in the HTTP push handler

- 
- 
- 

Added unit test

Fixes #1463

**Checklist:**


j0sh (approved):
LGTM

PR #1475, opened 2020-04-28
server: Propagate last error when max transcodes hit.

j0sh:
Also suppresses a returned error when MaxAttempts == 0.

- Propagates last error when max attempts hit
- Suppresses returned error when MaxAttempts == 0
- Adjusts test cases to match

Makes it easier to diagnose errors. If max attempts is consistently being hit, this is usually an indication something is wrong with the segment.

**Checklist:**


j0sh:
Something unrelated in CircleCI is making builds fail, not sure why. But tests should pass if running locally.

darkdarkdragon (approved):
LGTM

PR #1476, opened 2020-04-29
Allow multi-O setups with different wallets

yondonfu:

This PR enables multiple Os to be run under the same on-chain registered address. Each O:

- Has its own ETH address that is distinct from the on-chain registered address
- Can redeem winning tickets on behalf of the on-chain registered address
- Can sign transcoded results on behalf of the on-chain registered address

The `-ethOrchAddr` flag can be used to specify an on-chain registered address for an O that is different from the node's ETH address. 

As a result of these changes, a B's unknown session list (the session list that B runs stake weighted random selection on) can contain multiple sessions with the same on-chain registered address. If each session's stake weight is equal to the stake weight of the on-chain registered address then an on-chain registered address can increase its probability of receiving segments by increasing its # of Os (which increases the # of its sessions that could be in B's unknown session list). In order to avoid this, B sets a session's stake weight to the stake weight of the on-chain registered address divided by the # of sessions the address is associated with.

This PR only adds support for discovering multiple Os run under the same on-chain registered address via the `-orchAddr` flag (i.e. passing in the service URIs for each O) and via the `-orchWebhookUrl` flag (i.e. passing in a URL that can be used to fetch service URIs for each O). Support for on-chain discovery (i.e. B can discover multiple Os running under the same on-chain registered address without prior knowledge of the service URI of each O) is left for a separate PR.

This PR still requires each O to have ETH funds for redeeming winning tickets. Support for running an O without ETH funds is left for a separate PR.

- 54a98dd a0e15d7 09b3a13: Include an ETH address in the OrchestratorInfo msg that is transcoded results should be signed by
- 6c3d199: Update Os to check if the on-chain registered address is active instead of checking if their own address is active
- 77517c6: Add the `-ethOrchAddr` flag and disable round initialization and reward calls for Os if their node address is *not* the on-chain registered address
- bfac06d: Update stake weight assignment for sessions when multiple sessions are associated with the same on-chain registered address

Added unit tests.

TODO: Manual testing on Rinkeby .

**Checklist:**


yondonfu:
CI failing due to connection issue with a Rinkeby node - will poke CI after the issue is fixed.

EDIT: CI is fixed.

server/selection.go
```diff
@@ -190,8 +195,11 @@ func (s *MinLSSelector) selectUnknownSession() *BroadcastSession {
 	// The greater the stake weight of a session, the more likely that it will be selected because subtracting its stake weight from r
 	// will result in a value <= 0
 	for i, sess := range s.unknownSessions {
+		addr := ethcommon.BytesToAddress(sess.OrchestratorInfo.TicketParams.Recipient)
 		// If we could not fetch the stake weight for addrs[i] then its stake weight defaults to 0
-		r -= stakes[addrs[i]]
+		// The stake weight for a session is the stake weight of the address divided by the # of sessions the
+		// address is associated with
+		r -= (stakes[addr] / int64(addrCount[addr]))
```
j0sh: Am I correct in assuming that rounding might only a problem here for tiny amounts of stake, which isn't typically a problem in practice? (So it's safe to ignore the small bias from rounding down.)

j0sh (approved):
Just a small question, looks pretty good to me otherwise

core/orchestrator.go
```diff
@@ -307,10 +307,10 @@ func (orch *orchestrator) DebitFees(addr ethcommon.Address, manifestID ManifestI
 	orch.node.Balances.Debit(addr, manifestID, priceRat.Mul(priceRat, big.NewRat(pixels, 1)))
 }
 
-func (orch *orchestrator) isActive() (bool, error) {
+func (orch *orchestrator) isActive(addr ethcommon.Address) (bool, error) {
```
kyriediculous, line 310: I feel like we lose the purpose of isActive as a method here when passing in the arbitrary argument address as it no longer pertains to the address saved as state on `orchestrator`. 

Can we either make this into a pure function `isActive(OrchestratorStore, Address)`  if we choose to change the API or alternatively save the "master" address as state as well ?

server/rpc.go
```diff
@@ -243,22 +243,22 @@ func genOrchestratorReq(b common.Broadcaster) (*net.OrchestratorRequest, error)
...
 	}
 
 	// currently, orchestrator == transcoder
-	return orchestratorInfo(orch, addr, orch.ServiceURI().String())
+	return orchestratorInfo(orch, bAddr, orch.Address(), orch.ServiceURI().String())
 }
 
-func orchestratorInfo(orch Orchestrator, addr ethcommon.Address, serviceURI string) (*net.OrchestratorInfo, error) {
-	priceInfo, err := orch.PriceInfo(addr)
+func orchestratorInfo(orch Orchestrator, bAddr ethcommon.Address, oAddr ethcommon.Address, serviceURI string) (*net.OrchestratorInfo, error) {
```
kyriediculous: I feel like `bcast ethcommon.Address` and `orch etchommon.Address` conveys clearly enough that we are talking about an O and B in terms of their address. I don't think we need to convey types in the variable names here as well but that's just personal preference. Leaving it up to you !

kyriediculous (commented):
Just two minor comments. Since they are somewhat opinionated I didn't explicitly mark it as a change request.

server/selection.go
kyriediculous: If  `stakes[addr] < addrCount[addr]` we end  up with a stake weight of 0 , I think it's valid to think about how we should handle that. Should we penalise the node operator for running too many nodes with a very low stake which would not warrant running many nodes because the chances of being selected  are slim or should we default to `stakes[addr]` when it is  smaller than the `addrCount`?

server/rpc.go
yondonfu: I ended up reverting back to the original `orchestratorInfo()` signature in 0d73205 because the orch address can be accessed via the `orch Orchestrator` parameter that is already passed into the function so we don't have to worry about additional parameter naming right now. At some point if the `OrchestratorInfo.Address` field needs to be populated with a value that is not accessible via `orch Orchestrator` then we can consider updates to the `orchestratorInfo()` signature then.

server/selection.go
yondonfu: I updated the stake weight division in 1502b62.

For each session associated with an address, the session is assigned the remaining stake weight for the address (`stakes[addr]`) divided by the current # of sessions for the address (`addrCount[addr]`). After assigning the stake weight of a session, the session's stake weight is subtracted by the remaining stake weight for the address and the current # of sessions for the address is decremented. The result is that the last session for an address during a `selectUnknownSession()` call will be assigned the remaining stake weight of the address such that the entire stake weight of the address is used.

In the case where `stakes[addr] < addrCount[addr]` at the start of selection, each session would be assigned a stake weight of 0 except for one session which will be assigned a stake weight of `stakes[addr]`. This means that the maximum # of nodes that can be eligible for selection for an operator is `stakes[addr]` (the working set is capped at 8 right now so you can divide stake amongst at most 8 sessions right now). That seems like an ok restriction right now.

core/orchestrator.go
yondonfu: `isActive()` still relies on the state of `orch` (calls `orch.rm.LastInitializedRound()` and `orch.node.Database.SelectOrchs()`) and the difference between `isActive(Orchestrator, Address)` and the current implementation is whether Orchestrator is a method receiver or function parameter. Don't think I see the benefit in switching to a function here so I'd prefer to keep the method.

Since the address passed into `isActive()` is always accessible via the recipient field in a ticket I'd prefer not to save the "master" address to avoid saving additional state.

yondonfu:
4753212 fixes a bug that was introduced in a previous commit in this PR where the node was logging that the `ethOrchAddr` address was inactive when it was active.

server/selection.go
j0sh: Nice approach @yondonfu .

If I'm understanding the overall flow correctly, one consequence might be that the stake is weighted towards the last node of `addr` in the `unknownSessions` list. Is the last node also the one that took the longest to respond to `OrchestratorInfo` request? Which may mean the node is under load, far away, network congested, etc.

Not sure how much of an effect this will actually have in practice though; I imagine it's only under specific cases (minuscule amounts of stake + issues with the last O) where we might see issues as the B selects the first node or so from a refreshed list.

yondonfu: Yeah the last node of `addr` in `unknownSessions` would either be the node that took the longest to respond during discovery or a node that was suspended, but included in `unknownSessions` because there were insufficient non-suspended nodes that responded during discovery. In practice, I think this will only matter in the case that you mention with small amounts of stake + issues with the last node. Thought it would be nice to avoid any bias toward specific nodes even in this rare case so I pushed up 4a1cc12 which randomly assigns the "leftover" stake for `addr` to one of its sessions.

While working on that commit, I also discovered a bug that could cause zero stake weight sessions to be selected before non-zero stake weight selections. Fixed in 3d89c2b.

server/selection.go
```diff
@@ -183,10 +183,10 @@ func (s *MinLSSelector) selectUnknownSession() *BroadcastSession {
 		totalStake += stake
 	}
 
-	// Generate a random stake weight between 0 and totalStake
 	r := int64(0)
+	// Generate a random stake weight between 1 and totalStake
 	if totalStake > 0 {
-		r = rand.Int63n(totalStake)
+		r = 1 + rand.Int63n(totalStake)
```
yondonfu, line 189: When `totalStake > 0`, the min for `r` needs to be 1. Otherwise, if `r` is set to 0 and there are zero stake weight sessions before non-zero stake weight sessions in `unknownSessions`, then the zero stake weight sessions would be selected before the non-zero stake weight sessions.

kyriediculous (approved):

PR #1477, opened 2020-04-30
server: Set request timeouts based on segment duration.

### LINKED ISSUE #1480: Server: Set request timeouts based on segment duration
Allow the API user to set timeout.

PR: https://github.com/livepeer/go-livepeer/pull/1477

j0sh:
### TODO
- [x] Some sort of hard cap on durations. 5 minutes? ~~10? 15?~~
- [x] Watchdog to defer session teardown. Maybe not absolutely necessary, since multipart results are still returned even if the session is cleaned out from underneath it, but it could cause problems in the future. For example, the HLS playlist is not populated due to a `SessionEnded` error.
- [ ] ~~Split T support (requires changes to the wire protocol and probably payment processing). Issue writeup forthcoming~~ Split T was more work than expected, so will leave it out of this PR. Payment processing changes here: https://github.com/livepeer/go-livepeer/pull/1491

### What does this pull request do? Explain your changes. (required)

Sets broadcaster request timeouts based on segment duration, and the associated support for that.

### Specific updates (required)

- Set request timeout using a cancellable context. The timeout is padded to 4 times the segment duration.
- Remove default `common.HTTPTimeout` from `http.Client`  which would have triggered if `request timeout > common.HTTPTimeout`
- Uses `common.HTTPTimeout` as a minimum if the (padded) segment duration is shorter. This is so we don't time out on very short segments - there are typically overheads with transport, processing, etc.
- Kicks the session timeout watchdog if the segment is still being transcoded and we're within 90% of session expiration.
- Add unit tests. All HTTP push tests are now run under the race detector.

#### server: Set request timeouts based on segment duration. … 349e2c5
Better accommodates segments that may be longer than 2 seconds.

Uses a duration multiplier of 4, which should roughly line up with
the default timeout of 8 seconds (4 * 2s default segment length).

#### server: Constrain min and max segment durations. … 860dfb8
* Minimizes the risk of a broken connection from long transcodes
* Minimizes value at risk with probabilistic micropayments

#### server: Prefix push tests. … 8579778
Allows running push tests standalone.

#### server: Run HTTP push tests under -race. … 7fc2f08
The change to the BroadcastJobVideoProfiles assignment was the
only thing necessary to make the race detector pass. All other
tests still pass unchanged.

This *may* be a behavioral change if someone was, for some reason,
starting up the node with an empty set of transcoding profiles.

#### server: Kick watchdog periodically. … 9f94d61
Prevents the session from being cleaned up in the middle of a
long-running transcode.

### How did you test each of these updates (required)

- [x] Unit tests
- [x] Manually tested with an entire big buck bunny

### Does this pull request close any open issues?**

- Closes https://github.com/livepeer/go-livepeer/issues/1480
- Partial solution for https://github.com/livepeer/go-livepeer/issues/1478

**Checklist:**

### DIFF
--- a/server/broadcast.go
+++ b/server/broadcast.go
@@ -25,6 +25,8 @@ import (
 )
 
 var refreshTimeout = 2500 * time.Millisecond
+var maxDuration = (5 * time.Minute)
+var maxDurationSec = maxDuration.Seconds()
 
 var Policy *verification.Policy
 var BroadcastCfg = &BroadcastConfig{}
@@ -278,6 +280,11 @@ func processSegment(cxn *rtmpConnection, seg *stream.HLSSegment) ([]string, erro
 	cpl := cxn.pl
 	mid := cxn.mid
 	vProfile := cxn.profile
+
+	if seg.Duration > maxDurationSec || seg.Duration < 0 {
+		glog.Errorf("Invalid duration nonce=%d manifestID=%s seqNo=%d dur=%v", nonce, mid, seg.SeqNo, seg.Duration)
+		return nil, fmt.Errorf("Invalid duration %v", seg.Duration)
+	}
 
 	glog.V(common.DEBUG).Infof("Processing segment nonce=%d manifestID=%s seqNo=%d dur=%v", nonce, mid, seg.SeqNo, seg.Duration)
 	if monitor.Enabled {
--- a/server/mediaserver.go
+++ b/server/mediaserver.go
@@ -137,7 +137,9 @@ func NewLivepeerServer(rtmpAddr string, lpNode *core.LivepeerNode, httpIngest bo
 
 //StartMediaServer starts the LPMS server
 func (s *LivepeerServer) StartMediaServer(ctx context.Context, transcodingOptions string, httpAddr string) error {
-	BroadcastJobVideoProfiles = parsePresets(strings.Split(transcodingOptions, ","))
+	if transcodingOptions != "" { // mostly to mitigate race conditions in tests
+		BroadcastJobVideoProfiles = parsePresets(strings.Split(transcodingOptions, ","))
+	}
 
 	glog.V(common.SHORT).Infof("Transcode Job Type: %v", BroadcastJobVideoProfiles)
 
--- a/server/mediaserver.go
+++ b/server/mediaserver.go
@@ -680,6 +680,7 @@ func (s *LivepeerServer) HandlePush(w http.ResponseWriter, r *http.Request) {
 	// Do the transcoding!
 	urls, err := processSegment(cxn, seg)
 	if err != nil {
+		// TODO distinguish between user errors (400) and server errors (500)
 		http.Error(w, err.Error(), http.StatusInternalServerError)
 		return
 	}
--- a/server/mediaserver.go
+++ b/server/mediaserver.go
@@ -62,6 +62,10 @@ var BroadcastJobVideoProfiles = []ffmpeg.VideoProfile{ffmpeg.P240p30fps4x3, ffmp
 var AuthWebhookURL string
 
 var refreshIntervalHttpPush = 1 * time.Minute
+var httpPushResetTimer = func() (context.Context, context.CancelFunc) {
+	sleepDur := time.Duration(int64(float64(refreshIntervalHttpPush) * 0.9))
+	return context.WithTimeout(context.Background(), sleepDur)
+}
 
 type streamParameters struct {
 	mid        core.ManifestID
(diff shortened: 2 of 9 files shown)


server/segment_rpc.go
```diff
@@ -340,8 +341,21 @@ func SubmitSegment(sess *BroadcastSession, seg *stream.HLSSegment, nonce uint64)
 		return nil, err
 	}
 
+	// set a minimum timeout to accommodate transport / processing overhead
+	dur := common.HTTPTimeout
+	paddedDur := 4.0 * seg.Duration // use a multiplier of 4 for now
+	if paddedDur > dur.Seconds() {
+		dur, err = time.ParseDuration(fmt.Sprintf("%fs", paddedDur))
```
darkdarkdragon: Why ParseDuration here? Why not
`dur = time.Duration(paddedDur * float64(time.Second))` ?

j0sh: Good question. I actually remember spending some time on this particular line, but I can't for the life of me remember the reason for that. It might have to do with invalid value handling, but again don't remember the specific case (float range? seems unlikely / irrelevant).

Anyway, I agree this approach is better, and it doesn't seem to break anything else, so corrected in 856ceaf836dc780c454d476acd68b94c7cb2aab7

server/broadcast.go
```diff
@@ -25,6 +25,8 @@ import (
 )
 
 var refreshTimeout = 2500 * time.Millisecond
+var maxDuration = (5 * time.Minute)
```
yondonfu, line 28: Do we need this var? Could we just do `var maxDurationSec = (5 * time.Minute).Seconds()`?

server/mediaserver.go
```diff
@@ -62,6 +62,10 @@ var BroadcastJobVideoProfiles = []ffmpeg.VideoProfile{ffmpeg.P240p30fps4x3, ffmp
 var AuthWebhookURL string
 
 var refreshIntervalHttpPush = 1 * time.Minute
+var watchdogResetTimer = func() (context.Context, context.CancelFunc) {
```
yondonfu: Would `httpPushResetTimer` be a more descriptive name here?

server/push_test.go
```diff
@@ -592,7 +594,131 @@ func TestShouldNotPanicIfSessionAlreadyRemoved(t *testing.T) {
 	refreshIntervalHttpPush = oldRI
 }
 
-func TestFileExtensionError(t *testing.T) {
+func TestPush_ResetWatchdog(t *testing.T) {
+
+	wgWait(&watchdogWg) // wait for any earlier tests to complete
```
yondonfu: If `wgWait()` returns false that means we timed out for some reason before `watchDogWg.Wait()` completes right? Should we assert that it returns true since if it returns false that would imply there was a problem waiting for earlier tests to complete?

server/push_test.go
```diff
@@ -592,7 +594,131 @@ func TestShouldNotPanicIfSessionAlreadyRemoved(t *testing.T) {
...
+		return ctx, wrappedCancel
+	}
+	defer func() { watchdogResetTimer = oldWatchdogReset }()
+
+	// sanity check : normal flow should result in a single cancel
+	w := httptest.NewRecorder()
+	req := httptest.NewRequest("POST", "/live/name/0.ts", nil)
+	s.HandlePush(w, req)
+	assert.True(waitBarrier(timerCreationBarrier), "timer creation timed out")
+	wgWait(&wg)
```
yondonfu: Should we assert that this returns true to make sure we didn't time out? Same question for the rest of the `wgWait` invocations.

server/broadcast.go
j0sh: Yep, that's originally how it was, then started working on the next PR and ended up with this: https://github.com/livepeer/go-livepeer/pull/1491/commits/c0507c67206b957a2e85327dc2568cfaa7014051#diff-cecc27bfe2b6b19181ce443733dfba57R30

So I just went ahead and preemptively split them up here.

server/mediaserver.go
j0sh: Technically the term of art for this type of mechanism is ["watchdog"](https://en.wikipedia.org/wiki/Watchdog_timer) which I do find more descriptive with regards to its intent - it periodically checks for activity and closes the session otherwise. (And this particular addition "kicks" the watchdog if we're still mid-activity.) At least, that seems more  descriptive to me than the "refresh interval" that it's associated with - which doesn't actually *refresh* anything.

Anyway, fixed in 8c1e19495b48daeb15c6ab5fa9ba313d686472bc .

Another option is we can keep the "watchdog" terminology (or use something else more descriptive, I'm open to whatever), and rename the "refreshIntervalHttpPush" and related code to associate better with the semantic concept, eg "httpPushWatchdogTimeout".

server/push_test.go
j0sh: Yep, fixed in https://github.com/livepeer/go-livepeer/commit/8c1e19495b48daeb15c6ab5fa9ba313d686472bc

For reference, this is needed to make tests race-clean, otherwise it complains when we reset the watchdogResetTimer/httpPushResetTimer function: since the reset timer is run asynchronously, an earlier test may still be in the process of "finishing up" when this test case resets the function.

server/push_test.go
j0sh: Yep, fixed in https://github.com/livepeer/go-livepeer/commit/8c1e19495b48daeb15c6ab5fa9ba313d686472bc

server/mediaserver.go
yondonfu: The watchdog terminology + using var names that associate with the semantic concept sound good to me.

yondonfu (approved):
Left one comment re: var naming, but other than that LGTM!

server/mediaserver.go
j0sh: `httpPushWatchdogTimeout` turned out to be too much of a mouthful (and `watchdogTimeout` is perhaps not specific enough given this code handles a lot of things), so kept your suggestion and renamed `refreshIntervalHttpPush` to `httpPushTimeout`, and added a few comments to clarify its usage. 62be37c4f41e72c6fe0887a725b187c0ae6eda94

j0sh:
Fixed up feedback commits, and did another small renaming commit 62be37c following the discussion in https://github.com/livepeer/go-livepeer/pull/1477#discussion_r425990674

Will merge Monday unless anyone has other comments.

darkdarkdragon (approved):

Issue #1472, opened 2020-04-24
redeemWinningTicket fails if round is not initialized


criticaltv:
**Describe the bug**

If the O calls `redeemWinningTicket`, and the transaction is processed between rounds (i.e. before the next round has been initialised), then the transaction fails.

First, this wastes gas :(

Second, it's not clear whether the O automatically retries to redeem, or whether that ticket must be claimed manually.

**To Reproduce**

1. Receive a winning ticket
2. Attempt to redeem it after one round is finished and before the next round is initialized
3. See that the transaction fails with `Fail with error 'current round is not initialized'`

Example: https://rinkeby.etherscan.io/tx/0x963052482da1bb03de1f918926ee5a273a6c2e5227ab8337096aab8e3c8294e9

**Expected behavior**

First, I expect that a winning ticket ought to be redeemable at all time, independent of whether the round has been initialised :)

Second, I expect that the node should check whether the round has been initialised before attempting to redeem a precious winning ticket. It should perhaps check whether the transaction is always failing?

Third, I expect that redemption should be retried at a later stage (note: I am unable to verify if this happened or not)

kyriediculous:
Thanks for the catch! 

Since v0.5.5 tickets will no longer be redeemed instantly upon receiving them (have to wait until O recipient reveals the seed). 

I just checked and it doesn't seem to be the case that transaction failures are being re-added to the queue.
https://github.com/livepeer/go-livepeer/blob/0ee99d8905d461d734b638b0f5b533caa3e2a016/pm/recipient.go#L361

We should definitely re-add those tickets to the queue.

I think we should also avoid submitting transactions when the round isn't initialised but that might require additional engineering work so perhaps let's split this up into two smaller issues. This would be similar to checking the maxFloat. I think we need to add state to the `roundsWatcher` that not only checks the latest round number but whether it has been initialised. 
https://github.com/livepeer/go-livepeer/blob/0ee99d8905d461d734b638b0f5b533caa3e2a016/pm/recipient.go#L334

1. Fix resubmission on tx submission failure
2. Avoid submitting always failing transactions due to round not being initialised (we already do this if a B's max float is insufficient to cover the ticket facevalue)

kyriediculous:
fixed by #1495

Issue #654, opened 2018-12-26
pm: support ETH call and transaction submission retries in Broker implementation


yondonfu:
**Is your feature request related to a problem? Please describe.**

Errors from ETH calls and/or transaction submission when interacting with the TicketBroker contract can be housed in a single place such that users of the Broker interface implementation i.e. recipient and sender do not each have to handle these types of errors and retry.

EDIT: We could also include tx replacements for existing txs that sit pending for a long time.

**Describe the solution you'd like**

One option is that the Broker interface implementation functions can retry for these types of errors for a certain period of time and then only return an error if there is a fairly high chance that something is definitely wrong with the ETH connection i.e. the Ethereum node is not online or cannot support the request.

Another option is that the RPC client used to send requests to Ethereum nodes can implement the retries for these types of errors.

EDIT: For existing txs that sit pending for a long time, we could automatically submit replacement txs with bumped gas prices (within some acceptable range so we don't bid in a gas price auction endlessly).

**Describe alternatives you've considered**

If we don't handle retries for these types of errors in a single location we could end up with duplicated retry code in a bunch of areas of the codebase. Handling these types of errors in a single location can also relieve callers of the burden of trying to recover from these types of errors because the callers will know that something is actually wrong with the ETH connection when they receive an error.

eladmallel:
@yondonfu perhaps it's a good idea to give some specific examples of errors that should be auto-retried, versus the kind of errors we won't want to retry?

PR #1493, opened 2020-05-14
common,core,cmd,pm: DB Backed ticket retry queue 

### LINKED ISSUE #942: DB backed ticket retry queue

#926 adds an concurrency safe in-memory queue that keeps track of winning tickets that are awaiting redemption because the sender's max float was insufficient to cover the ticket face value when the recipient originally attempted to redeem the ticket.

Since we are already writing winning tickets to the DB upon receiving them, we can actually switch to a DB backed queue implementation which has a number of benefits:

- The state of the queue is persisted across reboots
- Avoids additional reasoning about channel constructs used to handle concurrent additions to the queue since we're just reading tickets that are already written to the DB

Modify the existing `ticketQueue` implementation in `pm/queue.go` to get the head of the queue using a DB query that fetches the oldest winning ticket that has not been confirmed on-chain (and that has not failed for some reason i.e. contract revert) for a particular sender ETH address. 

We should also keep track of the status of a ticket in the DB as one of: `null` (not submitted yet), `pending`, `success` or `failed`. When we fetch the head of a queue, tickets with status `success` or `failed` will be excluded from the query. When a transaction confirms we should update the ticket status in the DB.

We could stick with the existing in-memory queue, but it lacks some of the properties describes as benefits of using a DB backed queue.

**Additional context**

kyriediculous:
This PR introduces a DB backed `ticketqueue` and removes the need for the `recipient` to store or redeem winning tickets. Instead the sendermonitor will now consume winning tickets and the queue has permanent storage. 

- bfcae67: Reworks the DB table and methods for `winningTickets` 
   - The method signature for `StoreWinningTicket` now uses `pm.SignedTicket` to reduce arguments 
  - Added `LoadLatestTicket(sender)` , `RemoveWinningTicket(signedTicket)` and `WinningTicketCount(sender)`  

- d7c0f76: Reworks the `TicketStore` interface to accomodate to the database API changes, fb6cc80 reworks the stub used for testing. 

- Removes `TicketStore` and ticket redemption logic from the `recipient` , instead move that over to the `SenderMonitor` 

- 69ef438  reworks the ticketqueue to use `TicketStore`: 
  - Remove the conditional variable with lock from the queue, implementations of the `TicketStore` have to make sure they are thread safe. 
  - Updated Add() and Length() and added pop() 
  - Accept sender address and ticketstore in the constructor

Added , adjusted & ran unit tests

Fixes #942 

**Checklist:**

### DIFF
--- a/common/db.go
+++ b/common/db.go
@@ -672,6 +672,7 @@ func (db *DB) UnbondingLocks(currentRound *big.Int) ([]*DBUnbondingLock, error) 
 	return unbondingLocks, nil
 }
 
+// StoreWinningTicket stores a signed ticket
 func (db *DB) StoreWinningTicket(ticket *pm.SignedTicket) error {
 	if ticket == nil || ticket.Ticket == nil {
 		return errors.New("cannot store nil ticket")
@@ -703,6 +704,8 @@ func (db *DB) StoreWinningTicket(ticket *pm.SignedTicket) error {
 	return nil
 }
 
+// MarkWinningTicketRedeemed stores the on-chain transaction hash and timestamp of redemption
+// This marks the ticket as being 'redeemed'
 func (db *DB) MarkWinningTicketRedeemed(ticket *pm.SignedTicket, txHash ethcommon.Hash) error {
 	if ticket == nil || ticket.Ticket == nil {
 		return errors.New("cannot update nil ticket")
@@ -725,6 +728,7 @@ func (db *DB) MarkWinningTicketRedeemed(ticket *pm.SignedTicket, txHash ethcommo
 	return nil
 }
 
+// RemoveWinningTicket removes a ticket
 func (db *DB) RemoveWinningTicket(ticket *pm.SignedTicket) error {
 	if ticket == nil || ticket.Ticket == nil {
 		return errors.New("cannot delete nil ticket")
@@ -742,6 +746,8 @@ func (db *DB) RemoveWinningTicket(ticket *pm.SignedTicket) error {
 	return nil
 }
 
+// SelectEarliestWinningTicket selects the earliest stored winning ticket for a 'sender'
+// which is not yet redeemed
 func (db *DB) SelectEarliestWinningTicket(sender ethcommon.Address) (*pm.SignedTicket, error) {
 	row := db.selectEarliestWinningTicket.QueryRow(sender.Hex())
 	var (
@@ -782,6 +788,7 @@ func (db *DB) SelectEarliestWinningTicket(sender ethcommon.Address) (*pm.SignedT
 	}, nil
 }
 
+// WinningTicketCount returns the amount of non-redeemed winning tickets for a 'sender'
 func (db *DB) WinningTicketCount(sender ethcommon.Address) (int, error) {
 	row := db.winningTicketCount.QueryRow(sender.Hex())
 	var count64 int64
--- a/common/db.go
+++ b/common/db.go
@@ -33,6 +33,10 @@ type DB struct {
 	unbondingLocks                   *sql.Stmt
 	withdrawableUnbondingLocks       *sql.Stmt
 	insertWinningTicket              *sql.Stmt
+	selectEarliestWinningTicket      *sql.Stmt
+	winningTicketCount               *sql.Stmt
+	markWinningTicketRedeemed        *sql.Stmt
+	removeWinningTicket              *sql.Stmt
 	insertMiniHeader                 *sql.Stmt
 	findLatestMiniHeader             *sql.Stmt
 	findAllMiniHeadersSortedByNumber *sql.Stmt
@@ -111,8 +115,26 @@ var schema = `
 		sig BLOB,
 		sessionID STRING
 	);
-
 	CREATE INDEX IF NOT EXISTS idx_winningtickets_sessionid ON winningTickets(sessionID);
+
+	CREATE TABLE IF NOT EXISTS ticketQueue (
+		createdAt DATETIME DEFAULT CURRENT_TIMESTAMP,
+		sender STRING,
+		recipient STRING,
+		faceValue BLOB,
+		winProb BLOB,
+		senderNonce INTEGER,
+		recipientRand BLOB,
+		recipientRandHash STRING,
+		sig BLOB PRIMARY KEY,
+		creationRound int64,
+		creationRoundBlockHash STRING,
+		paramsExpirationBlock int64,
+		redeemedAt DATETIME,
+		txHash STRING
+	);
+
+	CREATE INDEX IF NOT EXISTS idx_ticketqueue_sender ON ticketQueue(sender);
 
 	CREATE TABLE IF NOT EXISTS blockheaders (
 		number int64,
@@ -182,7 +204,7 @@ func InitDB(dbPath string) (*DB, error) {
 	// selectKV prepared statement
 	stmt, err := db.Prepare("SELECT value FROM kv WHERE key=?")
 	if err != nil {
-		glog.Error("Unable to prepare selectKV stmt", err)
+		glog.Error("Unable to prepare selectKV stmt ", err)
 		d.Close()
 		return nil, err
 	}
@@ -269,7 +291,10 @@ func InitDB(dbPath string) (*DB, error) {
 	d.withdrawableUnbondingLocks = stmt
 
 	// Winning tickets prepared statements
-	stmt, err = db.Prepare("INSERT INTO winningTickets(sender, recipient, faceValue, winProb, senderNonce, recipientRand, recipientRandHash, sig, sessionID) VALUES(?, ?, ?, ?, ?, ?, ?, ?, ?)")
+	stmt, err = db.Prepare(`
+	INSERT INTO ticketQueue(sender, recipient, faceValue, winProb, senderNonce, recipientRand, recipientRandHash, sig, creationRound, creationRoundBlockHash, paramsExpirationBlock)
+	VALUES(:sender, :recipient, :faceValue, :winProb, :senderNonce, :recipientRand, :recipientRandHash, :sig, :creationRound, :creationRoundBlockHash, :paramsExpirationBlock)
+	`)
 	if err != nil {
 		glog.Error("Unable to prepare insertWinningTicket ", err)
 		d.Close()
@@ -277,10 +302,45 @@ func InitDB(dbPath string) (*DB, error) {
 	}
 	d.insertWinningTicket = stmt
 
+	// Select earliest ticket
+	stmt, err = db.Prepare("SELECT sender, recipient, faceValue, winProb, senderNonce, recipientRand, recipientRandHash, sig, creationRound, creationRoundBlockHash, paramsExpirationBlock FROM ticketQueue WHERE sender=? AND redeemedAt IS NULL AND txHash IS NULL ORDER BY createdAt ASC LIMIT 1")
+	if err != nil {
+		glog.Error("Unable to prepare selectEarliestWinningTicket ", err)
+		d.Close()
+		return nil, err
+	}
+	d.selectEarliestWinningTicket = stmt
+
+	stmt, err = db.Prepare("SELECT count(sig) FROM ticketQueue WHERE sender=? AND redeemedAt IS NULL AND txHash IS NULL")
+	if err != nil {
+		glog.Error("Unable to prepare winningTicketCount ", err)
+		d.Close()
+		return nil, err
+	}
+	d.winningTicketCount = stmt
+
+	// Remove latest ticket
+	stmt, err = db.Prepare("DELETE FROM ticketQueue WHERE sig=?")
+	if err != nil {
+		glog.Error("Unable to prepare removeWinningTicket ", err)
+		d.Close()
+		return nil, err
+	}
+	d.removeWinningTicket = stmt
+
+	// Mark ticket redeemed
+	stmt, err = db.Prepare("UPDATE ticketQueue SET redeemedAt=dateTime('now'), txHash=? WHERE sig=?")
+	if err != nil {
+		glog.Error("Unable to prepare markWinningTicketRedeemed", err)
+		d.Close()
+		return nil, err
+	}
+	d.markWinningTicketRedeemed = stmt
+
 	// Insert block header
 	stmt, err = db.Prepare("INSERT INTO blockheaders(number, parent, hash, logs) VALUES(?, ?, ?, ?)")
 	if err != nil {
-		glog.Error("Unable to prepare insertMiniHeader", err)
+		glog.Error("Unable to prepare insertMiniHeader ", err)
 		d.Close()
 		return nil, err
 	}
@@ -289,7 +349,7 @@ func InitDB(dbPath string) (*DB, error) {
 	// Find the latest block header
 	stmt, err = db.Prepare("SELECT * FROM blockheaders ORDER BY number DESC LIMIT 1")
 	if err != nil {
-		glog.Error("Unable to prepare findLatestMiniHeader", err)
+		glog.Error("Unable to prepare findLatestMiniHeader ", err)
 		d.Close()
 		return nil, err
 	}
@@ -298,7 +358,7 @@ func InitDB(dbPath string) (*DB, error) {
 	// Find all block headers sorted by number
 	stmt, err = db.Prepare("SELECT * FROM blockheaders ORDER BY number DESC")
 	if err != nil {
-		glog.Error("Unable to prepare findAllMiniHeadersSortedByNumber", err)
+		glog.Error("Unable to prepare findAllMiniHeadersSortedByNumber ", err)
 		d.Close()
 		return nil, err
 	}
@@ -307,7 +367,7 @@ func InitDB(dbPath string) (*DB, error) {
 	// Delete block header
 	stmt, err = db.Prepare("DELETE FROM blockheaders WHERE hash=?")
 	if err != nil {
-		glog.Error("Unable to prepare deleteMiniHeader", err)
+		glog.Error("Unable to prepare deleteMiniHeader ", err)
 		d.Close()
 		return nil, err
 	}
@@ -345,6 +405,18 @@ func (db *DB) Close() {
 	}
 	if db.insertWinningTicket != nil {
 		db.insertWinningTicket.Close()
+	}
+	if db.selectEarliestWinningTicket != nil {
+		db.selectEarliestWinningTicket.Close()
+	}
+	if db.winningTicketCount != nil {
+		db.winningTicketCount.Close()
+	}
+	if db.markWinningTicketRedeemed != nil {
+		db.markWinningTicketRedeemed.Close()
+	}
+	if db.removeWinningTicket != nil {
+		db.removeWinningTicket.Close()
 	}
 	if db.insertMiniHeader != nil {
 		db.insertMiniHeader.Close()
@@ -600,74 +672,128 @@ func (db *DB) UnbondingLocks(currentRound *big.Int) ([]*DBUnbondingLock, error) 
 	return unbondingLocks, nil
 }
 
-func (db *DB) StoreWinningTicket(sessionID string, ticket *pm.Ticket, sig []byte, recipientRand *big.Int) error {
-	if ticket == nil {
+func (db *DB) StoreWinningTicket(ticket *pm.SignedTicket) error {
+	if ticket == nil || ticket.Ticket == nil {
 		return errors.New("cannot store nil ticket")
 	}
-	if sig == nil {
+	if ticket.Sig == nil {
 		return errors.New("cannot store nil sig")
 	}
-	if recipientRand == nil {
+	if ticket.RecipientRand == nil {
 		return errors.New("cannot store nil recipientRand")
 	}
-	glog.V(DEBUG).Infof("db: Inserting winning ticket from %v, recipientRand %d, senderNonce %d", ticket.Sender.Hex(), recipientRand, ticket.SenderNonce)
 
-	_, err := db.insertWinningTicket.Exec(ticket.Sender.Hex(), ticket.Recipient.Hex(), ticket.FaceValue.Bytes(), ticket.WinProb.Bytes(), ticket.SenderNonce, recipientRand.Bytes(), ticket.RecipientRandHash.Hex(), sig, sessionID)
+	_, err := db.insertWinningTicket.Exec(
+		sql.Named("sender", ticket.Sender.Hex()),
+		sql.Named("recipient", ticket.Recipient.Hex()),
+		sql.Named("faceValue", ticket.FaceValue.Bytes()),
+		sql.Named("winProb", ticket.WinProb.Bytes()),
+		sql.Named("senderNonce", ticket.SenderNonce),
+		sql.Named("recipientRand", ticket.RecipientRand.Bytes()),
+		sql.Named("recipientRandHash", ticket.RecipientRandHash.Hex()),
+		sql.Named("sig", ticket.Sig),
+		sql.Named("creationRound", ticket.CreationRound),
+		sql.Named("creationRoundBlockHash", ticket.CreationRoundBlockHash.Hex()),
+		sql.Named("paramsExpirationBlock", ticket.ParamsExpirationBlock.Int64()),
+	)
 
 	if err != nil {
-		return errors.Wrapf(err, "failed inserting winning ticket for sessionID: %v, ticket: %v", sessionID, ticket)
+		return errors.Wrapf(err, "failed inserting winning ticket sender=%v", ticket.Sender.Hex())
 	}
 	return nil
 }
 
-func (db *DB) LoadWinningTickets(sessionIDs []string) (tickets []*pm.Ticket, sigs [][]byte, recipientRands []*big.Int, err error) {
-	rows, err := db.dbh.Query(buildWinningTicketsQuery(sessionIDs))
-	defer rows.Close()
+func (db *DB) MarkWinningTicketRedeemed(ticket *pm.SignedTicket, txHash ethcommon.Hash) error {
+	if ticket == nil || ticket.Ticket == nil {
+		return errors.New("cannot update nil ticket")
+	}
+	if ticket.Sig == nil {
+		return errors.New("cannot update nil sig")
+	}
 
+	res, err := db.markWinningTicketRedeemed.Exec(txHash.Hex(), ticket.Sig)
 	if err != nil {
-		err = errors.Wrapf(err, "failed loading winning tickets for sessionIDs %v", sessionIDs)
-		return
+		return errors.Wrapf(err, "failed marking winning ticket as redeemed sender=%v", ticket.Sender.Hex())
 	}
-
-	for rows.Next() {
-		var sender, recipient, recipientRandHash, sessionID string
-		var faceValue, winProb, recipientRandBytes, sig []byte
-		var senderNonce uint32
-
-		err = rows.Scan(&sender, &recipient, &faceValue, &winProb, &senderNonce, &recipientRandBytes, &recipientRandHash, &sig, &sessionID)
-		if err != nil {
-			err = errors.Wrapf(err, "failed scanning a winning ticket row for sessionID %v", sessionID)
-			return
-		}
-
-		ticket := &pm.Ticket{
-			Sender:            ethcommon.HexToAddress(sender),
-			Recipient:         ethcommon.HexToAddress(recipient),
-			FaceValue:         new(big.Int).SetBytes(faceValue),
-			WinProb:           new(big.Int).SetBytes(winProb),
-			SenderNonce:       senderNonce,
-			RecipientRandHash: ethcommon.HexToHash(recipientRandHash),
-		}
-		recipientRand := new(big.Int).SetBytes(recipientRandBytes)
-
-		tickets = append(tickets, ticket)
-		sigs = append(sigs, sig)
-		recipientRands = append(recipientRands, recipientRand)
+	rows, err := res.RowsAffected()
+	if err != nil {
+		return err
 	}
-
-	return
+	if rows == 0 {
+		return fmt.Errorf("no record found for sig=0x%x", ticket.Sig)
+	}
+	return nil
 }
 
-// We are building a query string instead of using a prepared statement because prepared statements don't
-// support IN queries. We want to use IN for the performance benefit, rather than running len(sessionIDs)
-// queries.
-// The known risk with string queries is SQL injection, however in this case all sessionID values are generated
-// by O, so such injections are not likely.
-func buildWinningTicketsQuery(sessionIDs []string) string {
-	for i := 0; i < len(sessionIDs); i++ {
-		sessionIDs[i] = strconv.Quote(sessionIDs[i])
+func (db *DB) RemoveWinningTicket(ticket *pm.SignedTicket) error {
+	if ticket == nil || ticket.Ticket == nil {
+		return errors.New("cannot delete nil ticket")
 	}
-	return "SELECT sender, recipient, faceValue, winProb, senderNonce, recipientRand, recipientRandHash, sig, sessionID FROM winningTickets WHERE sessionID IN (" + strings.Join(sessionIDs, ", ") + ")"
+	if ticket.Sig == nil {
+		return errors.New("cannot delete nil sig")
+	}
+
+	_, err := db.removeWinningTicket.Exec(
+		ticket.Sig,
+	)
+	if err != nil {
+		return errors.Wrapf(err, "failed deleting winning ticket sender=%v", ticket.Sender.Hex())
+	}
+	return nil
+}
+
+func (db *DB) SelectEarliestWinningTicket(sender ethcommon.Address) (*pm.SignedTicket, error) {
+	row := db.selectEarliestWinningTicket.QueryRow(sender.Hex())
+	var (
+		senderString           string
+		recipient              string
+		faceValue              []byte
+		winProb                []byte
+		senderNonce            int
+		recipientRand          []byte
+		recipientRandHash      string
+		sig                    []byte
+		creationRound          int64
+		creationRoundBlockHash string
+		paramsExpirationBlock  int64
+	)
+	if err := row.Scan(&senderString, &recipient, &faceValue, &winProb, &senderNonce, &recipientRand, &recipientRandHash, &sig, &creationRound, &creationRoundBlockHash, &paramsExpirationBlock); err != nil {
+		if err.Error() != "sql: no rows in result set" {
+			return nil, fmt.Errorf("could not retrieve earliest ticket err=%v", err)
+		}
+		// If there is no result return no error, just nil value
+		return nil, nil
+	}
+
+	return &pm.SignedTicket{
+		Ticket: &pm.Ticket{
+			Sender:                 sender,
+			Recipient:              ethcommon.HexToAddress(recipient),
+			FaceValue:              new(big.Int).SetBytes(faceValue),
+			WinProb:                new(big.Int).SetBytes(winProb),
+			SenderNonce:            uint32(senderNonce),
+			RecipientRandHash:      ethcommon.HexToHash(recipientRandHash),
+			CreationRound:          creationRound,
+			CreationRoundBlockHash: ethcommon.HexToHash(creationRoundBlockHash),
+			ParamsExpirationBlock:  big.NewInt(paramsExpirationBlock),
+		},
+		Sig:           sig,
+		RecipientRand: new(big.Int).SetBytes(recipientRand),
+	}, nil
+}
+
+func (db *DB) WinningTicketCount(sender ethcommon.Address) (int, error) {
+	row := db.winningTicketCount.QueryRow(sender.Hex())
+	var count64 int64
+	if err := row.Scan(&count64); err != nil {
+		if err.Error() != "sql: no rows in result set" {
+			return 0, fmt.Errorf("could not retrieve latest header: %v", err)
+		}
+		// If there is no result return no error, just nil value
+		return 0, nil
+	}
+
+	return int(count64), nil
 }
 
 func buildSelectOrchsQuery(filter *DBOrchFilter) (string, error) {
--- a/common/db_test.go
+++ b/common/db_test.go
@@ -7,6 +7,7 @@ import (
 	"math/big"
 	"strconv"
 	"testing"
+	"time"
 
 	"github.com/ethereum/go-ethereum/common"
 	ethcommon "github.com/ethereum/go-ethereum/common"
@@ -737,6 +738,62 @@ func TestDBUnbondingLocks(t *testing.T) {
 	}
 }
 
+func TestWinningTicketCount(t *testing.T) {
+	assert := assert.New(t)
+	dbh, dbraw, err := TempDB(t)
+	defer dbh.Close()
+	defer dbraw.Close()
+	require := require.New(t)
+	require.Nil(err)
+
+	sender := pm.RandAddress()
+
+	_, ticket, sig, recipientRand := defaultWinningTicket(t)
+	ticket.Sender = sender
+	count, err := dbh.WinningTicketCount(sender)
+	assert.Nil(err)
+	assert.Equal(count, 0)
+
+	err = dbh.StoreWinningTicket(&pm.SignedTicket{
+		Ticket:        ticket,
+		Sig:           sig,
+		RecipientRand: recipientRand,
+	})
+	require.Nil(err)
+
+	_, ticket, sig, recipientRand = defaultWinningTicket(t)
+	ticket.Sender = pm.RandAddress()
+	err = dbh.StoreWinningTicket(&pm.SignedTicket{
+		Ticket:        ticket,
+		Sig:           sig,
+		RecipientRand: recipientRand,
+	})
+	require.Nil(err)
+
+	count, err = dbh.WinningTicketCount(sender)
+	assert.Nil(err)
+	assert.Equal(count, 1)
+
+	// add a submitted ticket , should not change count
+	_, ticket, sig, recipientRand = defaultWinningTicket(t)
+	ticket.Sender = sender
+	err = dbh.StoreWinningTicket(&pm.SignedTicket{
+		Ticket:        ticket,
+		Sig:           sig,
+		RecipientRand: recipientRand,
+	})
+	dbh.MarkWinningTicketRedeemed(&pm.SignedTicket{
+		Ticket:        ticket,
+		Sig:           sig,
+		RecipientRand: recipientRand,
+	}, pm.RandHash())
+	require.Nil(err)
+
+	count, err = dbh.WinningTicketCount(sender)
+	assert.Nil(err)
+	assert.Equal(count, 1)
+}
+
 func TestInsertWinningTicket_GivenValidInputs_InsertsOneRowCorrectly(t *testing.T) {
 	dbh, dbraw, err := TempDB(t)
 	defer dbh.Close()
@@ -744,16 +801,22 @@ func TestInsertWinningTicket_GivenValidInputs_InsertsOneRowCorrectly(t *testing.
 	require := require.New(t)
 	require.Nil(err)
 
-	sessionID, ticket, sig, recipientRand := defaultWinningTicket(t)
+	_, ticket, sig, recipientRand := defaultWinningTicket(t)
 
-	err = dbh.StoreWinningTicket(sessionID, ticket, sig, recipientRand)
+	err = dbh.StoreWinningTicket(&pm.SignedTicket{
+		Ticket:        ticket,
+		Sig:           sig,
+		RecipientRand: recipientRand,
+	})
 	require.Nil(err)
 
-	row := dbraw.QueryRow("SELECT sender, recipient, faceValue, winProb, senderNonce, recipientRand, recipientRandHash, sig, sessionID FROM winningTickets")
-	var actualSender, actualRecipient, actualRecipientRandHash, actualSessionID string
+	row := dbraw.QueryRow("SELECT sender, recipient, faceValue, winProb, senderNonce, recipientRand, recipientRandHash, sig, creationRound, creationRoundBlockHash, paramsExpirationBlock FROM ticketQueue")
+	var actualSender, actualRecipient, actualRecipientRandHash, actualCreationRoundBlockHash string
 	var actualFaceValueBytes, actualWinProbBytes, actualRecipientRandBytes, actualSig []byte
 	var actualSenderNonce uint32
-	err = row.Scan(&actualSender, &actualRecipient, &actualFaceValueBytes, &actualWinProbBytes, &actualSenderNonce, &actualRecipientRandBytes, &actualRecipientRandHash, &actualSig, &actualSessionID)
+	var actualCreationRound int64
+	var paramsExpirationBlock int64
+	err = row.Scan(&actualSender, &actualRecipient, &actualFaceValueBytes, &actualWinProbBytes, &actualSenderNonce, &actualRecipientRandBytes, &actualRecipientRandHash, &actualSig, &actualCreationRound, &actualCreationRoundBlockHash, &paramsExpirationBlock)
 
 	assert := assert.New(t)
 	assert.Equal(ticket.Sender.Hex(), actualSender)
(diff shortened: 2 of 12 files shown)


kyriediculous:
I'm not sure what the exact process is to bump the DB version @j0sh could you shed some light on this? 

I also need to rename `LatestWinningTicket` , because it's not a correct name it takes the earliest ticket for a sender. What would a good name here be? `EarliestWinningTicket` ?

j0sh:
This looks great! Will give it a closer review shortly, but to quickly answer the question on bumping the DB version:

Was the `winningTickets` table used before at all? It seems to have been a leftover from when we were doing batch redemptions and never made it into a release, but I'm not quite sure of that. If the `winingTickets` was indeed unused, then that makes things simpler because we don't have to worry about migrating old data to the new structure. There's a couple options I see:

* Rename the table, eg `winningTickets2` and leave everything else as-is. The new table will automatically be created on node startup, and it won't clash with anything else. This seems simplest. (Older nodes will still have the old table sitting around, but that's harmless.)

* Write a migration procedure. Luckily, I don't *think* we have any data that will need to be preserved. So we can just drop the table, then re-run the [schema initialization](https://github.com/livepeer/go-livepeer/blob/fb6cc8098cb7b414613165d6940c2b152167f390/common/db.go#L156-L164) procedure. The [LivepeerDBVersion](https://github.com/livepeer/go-livepeer/blob/fb6cc8098cb7b414613165d6940c2b152167f390/common/db.go#L70) will need to be bumped.

In general, the migration procedure when the [DB has to be upgraded](https://github.com/livepeer/go-livepeer/blob/fb6cc8098cb7b414613165d6940c2b152167f390/common/db.go#L179-L181) is as follows:

```golang
if dbVersion == 1 {
 // Migrate to DB version 2:
 // Run SQL migration script, eg `DROP TABLE winningTickets` + reinitialize schema
 // Update dbVersion in database, eg `UPDATE TABLE kv SET value = 2 WHERE key = 'dbVersion'`
 dbVersion = 2
}
if dbVersion == 2 {
  // Migrate to DB version 3:
}
// ... and so on until dbVersion == LivepeerDBVersion 
```

In general, migration has more "moving parts" so it's nice to avoid that if possible, including considering this during the initial schema design (eg, treating the DB as another public API with its own back and forward-compatibility concerns, and to avoid building ourselves into a situation where we need to migrate stuff later on.)

Once we start migrating, it'd be pretty great to have regression tests of some sort against each version of the DB. Not sure if this is something that can be baked into unit tests, or if it fits better in the regression test tool.

common/db.go
```diff
@@ -269,14 +273,43 @@ func InitDB(dbPath string) (*DB, error) {
 	d.withdrawableUnbondingLocks = stmt
 
 	// Winning tickets prepared statements
-	stmt, err = db.Prepare("INSERT INTO winningTickets(sender, recipient, faceValue, winProb, senderNonce, recipientRand, recipientRandHash, sig, sessionID) VALUES(?, ?, ?, ?, ?, ?, ?, ?, ?)")
+	stmt, err = db.Prepare(`
+	INSERT INTO winningTickets(createdAt, sender, recipient, faceValue, winProb, senderNonce, recipientRand, recipientRandHash, sig, creationRound, creationRoundBlockHash)
+	VALUES(datetime('now'), :sender, :recipient, :faceValue, :winProb, :senderNonce, :recipientRand, :recipientRandHash, :sig, :creationRound, :creationRoundBlockHash)
```
j0sh: Since `createdAt` has the `DEFAULT CURRENT_TIMESTAMP` property, do we need to explicitly insert it as `datetime('now')` ?

common/db.go
```diff
@@ -269,14 +273,43 @@ func InitDB(dbPath string) (*DB, error) {
...
+	stmt, err = db.Prepare("SELECT count(sig) FROM winningTickets WHERE sender=?")
+	if err != nil {
+		glog.Error("Unable to prepare loadWinningTicketCount")
+		d.Close()
+		return nil, err
+	}
+	d.winningTicketCount = stmt
+
+	// Remove latest ticket
+	stmt, err = db.Prepare("DELETE FROM winningTickets WHERE sender=? AND sig=?")
```
j0sh: In addition to persistence, the DB can be really useful as a forensic / exploratory tool - we lose that property if we're removing data from it.

For "deletion" can we set something like a "submittedAt" column?  For the latest tickets, `SELECT ... FROM winningTickets where submittedAt IS NULL` . When the ticket is "deleted", set submittedAt to the current timestamp. If it becomes too annoying to deal with the conditional, we can create a view table of non-submitted tickets.

(Later on we could also store things like transaction ids, confirmation block, etc... lots of interesting things we can put in the DB. This could also come in handy independently, eg if the tx needs to be re-submitted for whatever reason, such as a spike in gas prices.)

kyriediculous: I like storing redeemed ticket transaction hashes (metadata lookup can be done remotely using the hash) but wouldn't it be better to use a separate table for this ? 

While `ticketqueues`  are per-sender this table isn't and having this table not grow too large while still having sender indexing seems like a nice property to have in terms of efficiency.

common/db.go
kyriediculous: Ah yes 👍  this must have snuck back in during a rebase

kyriediculous:
> Was the winningTickets table used before at all?

It was used to store winning tickets from the moment they were received and they are never deleted. Some orchestrator's DBs might contain tickets that have not been redeemed but are still redeemable however the chances of that right now are rather small (considering tickets are only valid for 2 rounds).

In most cases this DB will just contain stale tickets.

common/db.go
j0sh: >  wouldn't it be better to use a separate table for this 

It could be a separate table - depends on what we want to do with it. Eg, track re-submissions? Then maybe an independent table would make sense, with one row per submission. It's kind of nice to have the data alongside other ticket parameters though, or maybe the final tx hash at least. Makes common types of queries easier and we don't have to worry about cross-table consistency.

> having this table not grow too large while still having sender indexing seems like a nice property to have in terms of efficiency.

I wouldn't be too concerned about the table size. The data is valuable in itself, and with effective indexing, lookups will be very fast. If the size really becomes an issue later on, we can always offer tools for pruning the table. Until then, it seems like a loss to be deleting this data.

kyriediculous: I believe the broker handles transaction retries already so tickets should only be removed from the database once they are succesfully redeemed, otherwise they are readded to the DB (I might still have to implement this (#1472). 

I'm wondering if there's other alternatives possible to serve as an archive for forensic tools such as the subgraph. The only data currently we wouldn't have access to through the event indexing is the `pricePerPixel` for specific tickets. It's part of the HMAC but not part of any on-chain data but we could add it to the `auxData` for exactly this purpose down the line.

j0sh:
> It was used to store winning tickets from the moment they were received and they are never deleted. 

Yeah - I can see they're stored, but are they actually used? The only load call I see is in `RedeemWinningTickets`  which is unused. So, strictly speaking, we're only losing the historical data.

Since the `sender` field was removed and `creationRound` / `creationRoundBlockHash` were added, are pending tickets still redeemable with the old schema?

> In most cases this DB will just contain stale tickets.

Sure, that's OK. The question is whether this info may still be useful, eg for analysis?

I'll leave it up to you whether to actually preserve this (the migration will be more involved; alter table) but in general I'm uncomfortable with deleting historical data. So I'd really rather not make it a habit.

kyriediculous:
>So, strictly speaking, we're only losing the historical data.

Yup ! 

> Since the sender field was removed and creationRound / creationRoundBlockHash were added, are pending tickets still redeemable with the old schema?

I don't think the tickets in the DB stored according to the old schema are redeemable anyway in the current implementation regardless because they lack the necessary fields for valid `auxData` 

> Sure, that's OK. The question is whether this info may still be useful, eg for analysis?

Since `pricePerPixel` was  only added to tickets recently and it's not stored in the DB I believe all that data should be available on the subgraph as well. 

> in general I'm uncomfortable with deleting historical data. So I'd really rather not make it a habit.

Yeah, perhaps I can just create a new table for the ticketqueue. I'd like to perhaps think about what the least painful upgrade path is for our users. Wondering what @yondonfu thinks about this.

kyriediculous:
5a86b8f7 creates a new table `ticketQueue` instead of repurposing the existing `winningTickets` table, this should prevent backwards incompatible schema changes. 

I'll reword the original commit message on rebase

common/db.go
```diff
@@ -33,6 +33,9 @@ type DB struct {
 	unbondingLocks                   *sql.Stmt
 	withdrawableUnbondingLocks       *sql.Stmt
 	insertWinningTicket              *sql.Stmt
+	loadLatestTicket                 *sql.Stmt
```
yondonfu: We use "find" and "select" in the variable names for other read-only query statements - let's use one of those terms here for consistency.

Since the other winning ticket related query statements let's also clarify that this statement is for a winning ticket for consistency i.e. `...LatestWinningTicket`.

common/db.go
```diff
@@ -111,9 +114,24 @@ var schema = `
...
+		sender STRING,
+		recipient STRING,
+		faceValue BLOB,
+		winProb BLOB,
+		senderNonce INTEGER,
+		recipientRand BLOB,
+		recipientRandHash STRING,
+		sig BLOB,
+		creationRound int64,
+		creationRoundBlockHash BLOB
```
yondonfu: We'll also need the params expiration block attached to the ticket since the queue will check to make sure that ticket params are expired before passing a ticket to the SenderMonitor to be redeemed.

common/db.go
```diff
@@ -269,14 +287,43 @@ func InitDB(dbPath string) (*DB, error) {
...
 		glog.Error("Unable to prepare insertWinningTicket ", err)
 		d.Close()
 		return nil, err
 	}
 	d.insertWinningTicket = stmt
 
+	// Find latest ticket
+	stmt, err = db.Prepare("SELECT sender, recipient, faceValue, winProb, senderNonce, recipientRand, recipientRandHash, sig, creationRound, creationRoundBlockHash FROM ticketQueue WHERE sender=? ORDER BY createdAt ASC LIMIT 1")
+	if err != nil {
+		glog.Error("Unable to prepare loadLatestTicket")
```
yondonfu: Log error here? Same question for error handling around the other statement prep code.

common/db.go
```diff
@@ -269,14 +287,43 @@ func InitDB(dbPath string) (*DB, error) {
...
+	stmt, err = db.Prepare("SELECT count(sig) FROM ticketQueue WHERE sender=?")
+	if err != nil {
+		glog.Error("Unable to prepare loadWinningTicketCount")
+		d.Close()
+		return nil, err
+	}
+	d.winningTicketCount = stmt
+
+	// Remove latest ticket
+	stmt, err = db.Prepare("DELETE FROM ticketQueue WHERE sender=? AND sig=?")
```
yondonfu: If we make `sig` a primary key can we just delete using that without referencing `sender`? We should have run sender nonce and sig verification (which contains a [check](https://github.com/livepeer/go-livepeer/blob/master/crypto/verify.go#L34) to make sure that we only accept a single canonical signature per hash) prior to storing the ticket to the DB which should guarantee uniqueness of the sig.

Another option is to store the hash of the ticket as its unique identifier - this would match the TicketBroker contract's usage of the ticket hash as a unique ticket identifier.

common/db.go
```diff
@@ -600,74 +656,104 @@ func (db *DB) UnbondingLocks(currentRound *big.Int) ([]*DBUnbondingLock, error)
...
 	}
-	if recipientRand == nil {
+	if ticket.RecipientRand == nil {
 		return errors.New("cannot store nil recipientRand")
 	}
-	glog.V(DEBUG).Infof("db: Inserting winning ticket from %v, recipientRand %d, senderNonce %d", ticket.Sender.Hex(), recipientRand, ticket.SenderNonce)
 
-	_, err := db.insertWinningTicket.Exec(ticket.Sender.Hex(), ticket.Recipient.Hex(), ticket.FaceValue.Bytes(), ticket.WinProb.Bytes(), ticket.SenderNonce, recipientRand.Bytes(), ticket.RecipientRandHash.Hex(), sig, sessionID)
+	_, err := db.insertWinningTicket.Exec(
+		sql.Named("sender", ticket.Ticket.Sender.Hex()),
```
yondonfu: Since `Ticket` is embedded in `SignedTicket` all these fields should be promoted so we should be able to access them directly on `ticket` instead of accessing them on `ticket.Ticket`.

common/db.go
```diff
@@ -600,74 +656,104 @@ func (db *DB) UnbondingLocks(currentRound *big.Int) ([]*DBUnbondingLock, error)
...
+		sql.Named("recipientRand", ticket.RecipientRand.Bytes()),
+		sql.Named("recipientRandHash", ticket.Ticket.RecipientRandHash.Hex()),
+		sql.Named("sig", ticket.Sig),
+		sql.Named("creationRound", ticket.Ticket.CreationRound),
+		sql.Named("creationRoundBlockHash", ticket.Ticket.CreationRoundBlockHash.Bytes()),
+	)
 
 	if err != nil {
-		return errors.Wrapf(err, "failed inserting winning ticket for sessionID: %v, ticket: %v", sessionID, ticket)
+		return errors.Wrapf(err, "failed inserting winning ticket sender=%v recipientRand=%v nonce=%v", ticket.Ticket.Sender.Hex(), ticket.RecipientRand, ticket.Ticket.SenderNonce)
```
yondonfu: I know we log `recipientRand` and `senderNonce` elsewhere in the codebase, but I'm not sure how useful it is to actually include these values in error messages. I'd be in favor of just logging the sender address. Curious what you think though.

common/db.go
```diff
@@ -600,74 +656,104 @@ func (db *DB) UnbondingLocks(currentRound *big.Int) ([]*DBUnbondingLock, error)
...
+		senderNonce            int
+		recipientRand          []byte
+		recipientRandHash      string
+		sig                    []byte
+		creationRound          int64
+		creationRoundBlockHash []byte
+	)
+	if err := row.Scan(&senderString, &recipient, &faceValue, &winProb, &senderNonce, &recipientRand, &recipientRandHash, &sig, &creationRound, &creationRoundBlockHash); err != nil {
+		if err.Error() != "sql: no rows in result set" {
+			return nil, fmt.Errorf("could not retrieve latest header: %v", err)
```
yondonfu: Need to update this error message

common/db.go
yondonfu: Let's use a string here since we use that for hashes elsewhere.

common/db_test.go
```diff
@@ -737,23 +737,54 @@ func TestDBUnbondingLocks(t *testing.T) {
...
+	assert.Equal(count, 0)
+
+	err = dbh.StoreWinningTicket(&pm.SignedTicket{
+		Ticket:        ticket,
+		Sig:           sig,
+		RecipientRand: recipientRand,
+	})
+	require.Nil(err)
+
+	count, err = dbh.WinningTicketCount(ticket.Sender)
```
yondonfu: Let's add a test when there are tickets in the table for other senders as well to make sure the count returned is only for a specific sender.

common/db.go
```diff
@@ -269,14 +287,43 @@ func InitDB(dbPath string) (*DB, error) {
...
+	`)
 	if err != nil {
 		glog.Error("Unable to prepare insertWinningTicket ", err)
 		d.Close()
 		return nil, err
 	}
 	d.insertWinningTicket = stmt
 
+	// Find latest ticket
+	stmt, err = db.Prepare("SELECT sender, recipient, faceValue, winProb, senderNonce, recipientRand, recipientRandHash, sig, creationRound, creationRoundBlockHash FROM ticketQueue WHERE sender=? ORDER BY createdAt ASC LIMIT 1")
```
yondonfu: Since this statement sorts the tickets in ascending order based on the `createdAt` timestamp, it actually returns the oldest ticket so that should be reflected in the variable name used for the statement given that latest would imply the most recent ticket.

common/db_test.go
```diff
@@ -846,141 +888,97 @@ func TestStoreWinningTicket_GivenNilRecipientRand_ReturnsError(t *testing.T) {
...
-	tickets, sigs, recipientRands, err := dbh.LoadWinningTickets([]string{"some sessionID"})
-
-	assert := assert.New(t)
+	latest, err = dbh.LoadLatestTicket(ticket.Sender)
 	assert.Nil(err)
-	assert.Len(tickets, 0)
-	assert.Len(sigs, 0)
-	assert.Len(recipientRands, 0)
-}
+	assert.Equal(signedTicket0.Ticket, latest.Ticket)
```
yondonfu: Could this just be `assert.Equal(latest, signedTicket0)` similar to the check on L944?

common/db_test.go
```diff
@@ -846,141 +888,97 @@ func TestStoreWinningTicket_GivenNilRecipientRand_ReturnsError(t *testing.T) {
...
 
-	tickets, sigs, recipientRands, err := dbh.LoadWinningTickets([]string{""})
-
-	assert := assert.New(t)
+	latest, err = dbh.LoadLatestTicket(ticket.Sender)
 	assert.Nil(err)
-	assert.Len(tickets, 0)
-	assert.Len(sigs, 0)
-	assert.Len(recipientRands, 0)
+	assert.Equal(latest, signedTicket0)
```
yondonfu: Let's also add a test when there is a ticket in the table that was created earlier than `signedTicket0`, but that is for a different sender just to make sure we're always returning tickets for the right sender.

common/db_test.go
```diff
@@ -846,141 +888,97 @@ func TestStoreWinningTicket_GivenNilRecipientRand_ReturnsError(t *testing.T) {
...
-	assert.Equal(sig0, sigs[0])
-	assert.Equal(recipientRand0, recipientRands[0])
-	assert.Equal(ticket1, tickets[1])
-	assert.Equal(sig1, sigs[1])
-	assert.Equal(recipientRand1, recipientRands[1])
+	// removing the wrong ticket should return nil
+	signedTicketDup := *signedTicket
+	signedTicket.Sender = pm.RandAddress()
+	err = dbh.RemoveWinningTicket(&signedTicketDup)
+	assert.NoError(err)
```
yondonfu, line 1100: Should we also check that the ticket count did not change?

pm/queue.go
```diff
@@ -75,15 +66,8 @@ func (q *ticketQueue) Stop() {
...
-func (q *ticketQueue) Add(ticket *SignedTicket) {
-	// Lock conditional variable while adding to the queue
-	q.cond.L.Lock()
-	q.queue = append(q.queue, ticket)
-	atomic.AddInt32(&q.queueLen, 1)
-	q.cond.L.Unlock()
-
-	// Signal that there are tickets in the queue
-	q.cond.Signal()
+func (q *ticketQueue) Add(ticket *SignedTicket) error {
```
yondonfu, line 70: Can we update the comment for this function to reflect that a caller does not only add tickets to the queue when the sender's max float is insufficient to cover the ticket face value?

pm/queue.go
```diff
@@ -141,11 +131,13 @@ func (q *ticketQueue) startQueueLoop() {
...
 
 // removeHead removes the head of the queue
-func (q *ticketQueue) removeHead() {
-	// Lock conditional variable while removing from the queue
-	q.cond.L.Lock()
-	q.queue[0] = nil
-	q.queue = q.queue[1:]
-	atomic.AddInt32(&q.queueLen, -1)
-	q.cond.L.Unlock()
+func (q *ticketQueue) pop() (*SignedTicket, error) {
```
yondonfu: Can we update the comment since the method name changed?

pm/sendermonitor.go
```diff
@@ -79,11 +78,13 @@ type senderMonitor struct {
 	// each of currently active remote senders
 	redeemable chan *SignedTicket
 
+	ticketStore func() TicketStore
```
yondonfu: Why is this a function?

pm/queue_test.go
```diff
@@ -164,3 +184,90 @@ func TestTicketQueueConsumeBlockNums(t *testing.T) {
...
+	assert := assert.New(t)
+
+	sender := RandAddress()
+	ts := newStubTicketStore()
+	tm := &stubTimeManager{}
+
+	q := newTicketQueue(ts, sender, tm.SubscribeBlocks)
+	q.Start()
+	defer q.Stop()
+	time.Sleep(5 * time.Millisecond)
```
yondonfu: Doesn't seem like we need to call Start/Stop or use a sleep here since we're only testing `Add()` in this test and that method does not depend on the ticket queue loop.

pm/queue_test.go
```diff
@@ -164,3 +184,90 @@ func TestTicketQueueConsumeBlockNums(t *testing.T) {
...
+	ticket := defaultSignedTicket(sender, 0)
+
+	ts.storeShouldFail = true
+	err := q.Add(ticket)
+	assert.EqualError(err, "stub ticket store store error")
+	ts.storeShouldFail = false
+
+	err = q.Add(ticket)
+	assert.Nil(err)
+	t.Log(ts.tickets[sender.Hex()])
```
yondonfu: Can remove this

pm/queue_test.go
```diff
@@ -164,3 +184,90 @@ func TestTicketQueueConsumeBlockNums(t *testing.T) {
...
+	require := require.New(t)
+
+	sender := RandAddress()
+	ts := newStubTicketStore()
+	tm := &stubTimeManager{}
+
+	q := newTicketQueue(ts, sender, tm.SubscribeBlocks)
+	q.Start()
+	defer q.Stop()
+	time.Sleep(5 * time.Millisecond)
```
yondonfu: Doesn't seem like we need to call Start/Stop or use a sleep here since we're only testing `pop()` in this test and that method does not depend on the ticket queue loop.

pm/queue_test.go
```diff
@@ -164,3 +184,90 @@ func TestTicketQueueConsumeBlockNums(t *testing.T) {
...
+	assert.Len(ts.tickets[sender.Hex()], 1)
+	popped, err = q.pop()
+	assert.Nil(popped)
+	assert.EqualError(err, "stub TicketStore remove error")
+	assert.Len(ts.tickets[sender.Hex()], 1)
+	ts.removeShouldFail = false
+
+	popped, err = q.pop()
+	assert.Nil(err)
+	assert.Equal(popped, ticket)
```
yondonfu: Should we add an assertion for the ticket store to confirm that `RemoveWinningTicket()` was called on it?

pm/queue_test.go
```diff
@@ -164,3 +184,90 @@ func TestTicketQueueConsumeBlockNums(t *testing.T) {
...
+	assert := assert.New(t)
+
+	sender := RandAddress()
+	ts := newStubTicketStore()
+	tm := &stubTimeManager{}
+
+	q := newTicketQueue(ts, sender, tm.SubscribeBlocks)
+	q.Start()
+	defer q.Stop()
+	time.Sleep(5 * time.Millisecond)
```
yondonfu: Doesn't seem like we need to call Start/Stop or use a sleep here since we're only testing `Length()` in this test and that method does not depend on the ticket queue loop.

yondonfu (changes requested):

common/db.go
yondonfu: > I believe the broker handles transaction retries already

There currently is no redemption tx retry/replace logic for the case where a redemption tx sits pending for awhile and then `CheckTx()` times out. Adding such logic is tracked by #654.

yondonfu:
> I also need to rename LatestWinningTicket , because it's not a correct name it takes the earliest ticket for a sender. What would a good name here be? EarliestWinningTicket ?

EarliestWinningTicket sounds good to me.

> Yeah, perhaps I can just create a new table for the ticketqueue. I'd like to perhaps think about what the least painful upgrade path is for our users. Wondering what @yondonfu thinks about this.

New table and keeping the old table sounds good to me.

common/db.go
kyriediculous: fixed in the following commits 

2e4e7b9
6e7a823
433eb12
df308ae

common/db.go
kyriediculous: fixed in c0c037a

common/db.go
kyriediculous: fixed in 3143abbdff25fb42b93db012bee435b2999e730f

common/db.go
kyriediculous: fixed in b561729

common/db.go
kyriediculous: fixed in 50bb271

common/db.go
kyriediculous: I have no clear preference at this point so just logging the sender address is fine by me

kyriediculous: fixed in 5104db87

common/db.go
kyriediculous: fixed in 5104db87

common/db.go
kyriediculous: fixed in 572e05a8

common/db.go
kyriediculous: this was already renamed in an earlier fixup but this c542b9db commit addresses the comment

common/db_test.go
kyriediculous: fixed in 74a4532e8212d3375c1158b84923d68b8e71bc70

common/db_test.go
kyriediculous: fixed in ffc9005d05bb100aa6a141d8f8b76fa744d3656a

pm/sendermonitor.go
kyriediculous: SenderMonitor only needs to pass this dependency to newly created ticketqueues so rather than storing it as state I felt it would be okay to wrap it in a function

common/db_test.go
kyriediculous: fixed in 0d44843d

common/db_test.go
kyriediculous: fixed in 0d44843d

pm/queue.go
kyriediculous: fixed in df7f47fe

pm/queue.go
kyriediculous: fixed in df7f47fe

pm/queue_test.go
kyriediculous: I guess we can just pop again and check that it returns nil (no more tickets in the queue)

pm/queue_test.go
kyriediculous: fixed in 5e9feb24

pm/queue_test.go
kyriediculous: fixed in 5e9feb24

pm/queue_test.go
kyriediculous: fixed in 5e9feb24

pm/queue_test.go
kyriediculous: fixed in 5e9feb24

pm/queue_test.go
kyriediculous: fixed in 5e9feb24

common/db.go
```diff
@@ -269,14 +288,43 @@ func InitDB(dbPath string) (*DB, error) {
...
 		glog.Error("Unable to prepare insertWinningTicket ", err)
 		d.Close()
 		return nil, err
 	}
 	d.insertWinningTicket = stmt
 
+	// Select earliest ticket
+	stmt, err = db.Prepare("SELECT sender, recipient, faceValue, winProb, senderNonce, recipientRand, recipientRandHash, sig, creationRound, creationRoundBlockHash, paramsExpirationBlock FROM ticketQueue WHERE sender=? ORDER BY createdAt ASC LIMIT 1")
+	if err != nil {
+		glog.Error("Unable to prepare selectEarliestWinningTicket", err)
```
yondonfu: Needs a space at the end of the string or else there will be no spacing between the message and the error. Same comment for the other `glog.Error()` calls in this file.

common/db.go
```diff
@@ -600,74 +657,106 @@ func (db *DB) UnbondingLocks(currentRound *big.Int) ([]*DBUnbondingLock, error)
...
+		sql.Named("recipientRandHash", ticket.RecipientRandHash.Hex()),
+		sql.Named("sig", ticket.Sig),
+		sql.Named("creationRound", ticket.CreationRound),
+		sql.Named("creationRoundBlockHash", ticket.CreationRoundBlockHash.Hex()),
+		sql.Named("paramsExpirationBlock", ticket.ParamsExpirationBlock.Int64()),
+	)
 
 	if err != nil {
-		return errors.Wrapf(err, "failed inserting winning ticket for sessionID: %v, ticket: %v", sessionID, ticket)
+		return errors.Wrapf(err, "failed inserting winning ticket sender=%v recipientRand=%v nonce=%v", ticket.Sender.Hex(), ticket.RecipientRand, ticket.SenderNonce)
```
yondonfu: Logging the sender should be sufficient?

common/db_test.go
```diff
@@ -846,141 +901,112 @@ func TestStoreWinningTicket_GivenNilRecipientRand_ReturnsError(t *testing.T) {
...
+		RecipientRand: recipientRand,
+	}
+	err = dbh.StoreWinningTicket(signedTicket1)
 	require.Nil(err)
 
-	tickets, sigs, recipientRands, err := dbh.LoadWinningTickets([]string{firstSessionID})
-
-	assert := assert.New(t)
+	// no tickets found
+	latest, err := dbh.SelectEarliestWinningTicket(ethcommon.HexToAddress("charizard"))
```
yondonfu: Declarations of `latest` in this file when calling `SelectEarliestWinningTickets()` should be `earliest`

yondonfu (changes requested):

common/db.go
kyriediculous: Should we consider #654 in scope for this epic?

common/db.go
kyriediculous: Oh really? That's ugly.

JS autospaces this eg

```
console.log("hello", "world")
VM186:1 hello world
```

common/db_test.go
kyriediculous: fixed in 852a37e

common/db.go
kyriediculous: fixed in 36472ed

kyriediculous: fixed in f1a5826

common/db.go
j0sh: Can we still consider marking submitted rows as such, rather than deleting them from the table outright? There is really very little reason to delete data here.

kyriediculous: Would the storage growth be a worthwhile tradeoff to make considering once a ticket has been redeemed we also have historical data available through the subgraph or archive ETH nodes ?

common/db.go
kyriediculous: fixed in f1a5826

common/db.go
yondonfu: > Should we consider #654 in scope for this epic?

Let's keep #654 outside of the scope of this epic since the ticket queue can already handle retries for tickets that have not been successfully submitted. A solution for #654 may be useful for handling tx replacements with a gas price bump if the tx is not mined within some time window, but that can be addressed later.

> Would the storage growth be a worthwhile tradeoff to make considering once a ticket has been redeemed we also have historical data available through the subgraph or archive ETH nodes ?

I can see the benefits of keeping ticket data even after redemption based on the discussion thus far i.e. forensics/analysis without dependence on external tools, treating tickets as valuable data that shouldn't be deleted by default, etc.

Another benefit I see is if the node crashes in between removing a ticket from the queue and re-adding the ticket to the queue then the node can still redeem the ticket when it restarts. An example scenario might be if the node is waiting for the redemption tx to be mined when gas prices are high, the node crashes for some reason and the redemption tx is never mined because gas prices keep rising.

The main disadvantage of keeping ticket data even after redemption seems to be storage growth. But, storage growth should be manageable (if each ticket is ~284 bytes then 100000 winning tickets would be 28.4 MB) and as was mentioned earlier in the discussion we can add an opt-in pruning routine to delete really old tickets if needed. 

Given the advantages of keeping ticket data even after redemption and the manageable disadvantage of keeping ticket data, I think keeping ticket data makes sense.

kyriediculous: fixed in d69225c ab7363e

common/db.go
```diff
@@ -111,9 +114,26 @@ var schema = `
...
+		faceValue BLOB,
+		winProb BLOB,
+		senderNonce INTEGER,
+		recipientRand BLOB,
+		recipientRandHash STRING,
+		sig BLOB PRIMARY KEY,
+		creationRound int64,
+		creationRoundBlockHash STRING,
+		paramsExpirationBlock int64,
+		submitted BOOLEAN
```
yondonfu: Could this be a timestamp instead as suggested by Josh i.e. `submittedAt`  since it not only can be used to filter out tickets that have not been submitted, but it also captures additional information about when the ticket was submitted?

pm/queue.go
```diff
@@ -140,12 +126,17 @@ func (q *ticketQueue) startQueueLoop() {
...
+// pop returns the head of the queue
+func (q *ticketQueue) pop() (*SignedTicket, error) {
+	ticket, err := q.store.SelectEarliestWinningTicket(q.sender)
+	if err != nil {
+		return nil, err
+	}
+	if ticket == nil {
+		return nil, nil
+	}
+	if err := q.store.StoreWinningTicket(ticket, true); err != nil {
```
yondonfu: Instead of overloading the usage of `StoreWinningTicket()` here, could we:

- Update `RemoveWinningTIcket()` to mark a ticket as submitted
- Return the earliest winning ticket in `pop()`
- Wait for the ticket to be submitted before selecting the next earliest winning ticket in the queue loop
- If the ticket was submitted without an error, call `RemoveWinningTicket()` for the ticket
- At this point, select the next earliest winning ticket

This would keep the sole responsibility for `StoreWinningTicket()` as writing a new ticket to the store, changes the "delete" operation in `RemoveWinningTicket()` to just update an existing ticket as submitted and avoids updating the store when a ticket needs to be retried.

The implementation might look something like:


```golang
type redeemSignedTicket struct {
    *SignedTicket
    errCh chan error
}

func (q *ticketQueue) startQueueLoop() {
    ...
    if nextTicket.ParamsExpirationBlock.Cmp(latestBlock) <= 0 {
        errCh := make(chan error, 1)
        select {
            case q.redeemable <- &redeemSignedTicket{SignedTicket: nextTicket, errCh: errCh}
                // Make sure that the consumer of q.redeemable uses errCh to pass an error value
                // back (nil indicates success)
                if err := <-errCh; err != nil {
                    // Remove ticket
                } else {
                    // Handle error
                }
        }
    }
}
```

yondonfu (changes requested):

common/db.go
kyriediculous: I don't really see any added value in doing so. Once a ticket has been submitted on-chain all this metadata is already available on more permanent resources including its block timestamp. I don't think there's a clear use case for having submission timestamps for tickets that errored out eventually either. 

The purpose of the db backed ticket queue is to not lose winning tickets making this a forensics tool seems out of scope for this issue given its still questionable usefulness in the future as opposed to external tools we already have available. 

I feel like not removing winning tickets is already a good compromise here without knowing the full usefulness of retaining historical data.

pm/queue.go
kyriediculous: I think there's two separate discussions here 

- Using UPSERT vs Using two queries 
- Using an error channel vs having the consumer be responsible for re-adding tickets upon error

RE UPSERT: 
1. It's an UPSERT SQL query , I don't see how that's overloading. Having to create an additional query for this would be overloading the API, no? 

2. The function will have a totally different functionality than it's description (not insurmountable, but see 1)

3. The RemoveWinningTicket function should be reserved for DB pruning and actually deleting a winning ticket. For the same reason the method has not been removed despite being unused

4. The sole responsibility characteristic of `StoreWinningTicket` as an UPSERT query does not change. It stores winning tickets and their status accordingly.


RE ErrCh 
The behaviour described is not thread-safe, _I think_,  since tickets are not marked submitted at the moment they are sent into the channel. While this is not an issue right now, it is if we were to allow such behaviour in the future. 

I do see the benefit because in case the node shuts down while the current ticket is still in flight it will not be marked submitted and automatically retried when the node comes back online. _But_ this seems like a very rare edge case, especially when using the redemption service (basically when the user shuts down the node or if there's an unexpected panic).

So I'm wondering if this edge case warrants completely changing the behaviour of the existing implementation.

common/db.go
yondonfu: > The purpose of the db backed ticket queue is to not lose winning tickets making this a forensics tool seems out of scope for this issue

Agreed that the forensics is not the top priority here.

However, in this case, using a timestamp captures strictly more information while still accomplishing the same goal as a boolean. Furthermore, checking if a timestamp is not null vs. checking if a boolean is true/false for the purposes of filtering out submitted tickets requires the same amount of code. Given that the amount of code required for using a timestamp is equivalent to amount of code required for using a boolean, why not use a timestamp that captures extra information that could be useful?

pm/queue.go
j0sh: On UPSERT, it's also a matter of being clear about intention. UPSERTs are useful if you aren't sure if the value exists in the table. In this case, we know that a value will will always exist. (Maybe it should error out otherwise, but I'm indifferent.) The `RemoveWinningTicket` function can be renamed to `MarkTicketSubmitted` or something and the query updated accordingly. That makes the effect of "marking a ticket as submitted" more straightforward for the reader.

Another way to consider it is, there is no overlap in the fields that are actually updated between the initial insert and the update. The `submitted` field can be omitted from the initial insert, and updated from a separate query. This preserves the clarity of intent.

> So I'm wondering if this edge case warrants completely changing the behaviour of the existing implementation.

I'm not totally clear on when a ticket is considered "submitted" versus "in-flight" but will #654 mean accommodating this particular edge case? If so, it might be good to make the suggested change anyway, as a first step to avoid re-working code later on once #654 is implemented.

kyriediculous: In the current implementation we mark a ticket as submitted as soon as it's sent to the consumer. 

In [this comment](https://github.com/livepeer/go-livepeer/pull/1493#discussion_r434041781) a ticket would only be marked submitted once the consumer returns an error (or `nil`), so with in-flight I mean the period between sending it to the consumer and the moment we mark it as submitted. This would also put the responsibility of re-adding tickets to the queue for errors back with the queue, while the consumer already does some pre-flight checks so we'd have to start checking the returned error as well. 

Because of this "in-flight" status we wouldn't be able to run the inner loop concurrently if we'd ever choose to (e.g. if we receive new block events faster than we can run through the loop). 

I'm fine with a `WinningTicketSubmitted` function, but depending on the implementation detail mentioned above this has implications for what the API would look like. 

If we don't change the current queue-redeemer flow we'd have to set the `submittedAt` field to `NULL` again in the case of an error so the API signature would change to also include a boolean. 

eg `WinningTicketSubmitted(*pm.SignedTicket,  bool) error` vs `WinningTicketSubmitted(*pm.Signedticket) error`

kyriediculous: > On UPSERT, it's also a matter of being clear about intention. UPSERTs are useful if you aren't sure if the value exists in the table.

I was looking at the implementation details. An UPSERT would actually be more favorable if we don't change the current implementation of the queueloop to also receive errors on a channel because tickets are added back to the queue by the consumer on error. If we don't use an UPSERT `Queue.Add` would have to check if the ticket already exists. 

Thus my compromise proposition is 
- Don't change queue loop behaviour for now
- Use a `submittedAt` timestamp (set it to time.Now if we sent a ticket to the consumer over the channel, set it back to NULL if a ticket gets re-added to the queue but already exists)
- keep the UPSERT to avoid an extra check in `queue.Add`

the caveat is that if the node fails when a ticket is still being submitted and exits, its submittedAt timestamp would have to be changed back manually for it to be auto-retried. If we feel like this edge case is a problem we can consider changing the loop implementation as described by the suggestion.

yondonfu: > Because of this "in-flight" status we wouldn't be able to run the inner loop concurrently if we'd ever choose to (e.g. if we receive new block events faster than we can run through the loop).

Note that the inner loop you describe (the one that iterates through all the tickets in the queue) already cannot progress faster than the block time if the queue is not empty. At the moment, the redeemable channel consumer redeems a single ticket at a time and redemption takes at least a single block. So, by the time the channel consumer is ready for the next ticket, a new block would have been received, but before the ticket loop can handle the new block the inner loop would need to finish.

>  If we don't use an UPSERT Queue.Add would have to check if the ticket already exists.

Wouldn't this be a non-issue if we only update the timestamp after the inner loop receives confirmation that the ticket was submitted (if the confirmation is sent after the `redeemWinningTicket()` method returns then this would mean the tx was also mined)? Then, we wouldn't have to worry about re-adding tickets to the queue. 

From my POV, it seems that waiting for confirmation in the inner loop is a small change that only has benefits i.e. avoiding re-adding logic and also addressing the edge case that you mentioned. Let me know if I'm missing anything though.

I also agree with Josh about the better clarity of intent with a separate method/query (maybe `UpdateWinningTicket()`)  for updating the timestamp instead of the `StoreWinningTicket()` method and an UPSERT query.

kyriediculous: I agree that if we update waiting for confirmation in the inner loop would favor not using an UPSERT. I also acknowledge the benefits it provides. 

But as you mention we redeem tickets one block at a time, so if there's case where a queue receives multiple winning tickets per block then it's playing catch up until it doesn't receive winning tickets from that sender for a while. I do acknowledge this is an edge case, but one solution down the road if this ever becomes a problem could be to run `redeemWinningTicket()` (and possible that inner loop) on a go-routine. In this case if we were only to indicate a ticket to be submitted after it is mined we could en up concurrently trying the same transaction, wouldn't we?

yondonfu: > so if there's case where a queue receives multiple winning tickets per block then it's playing catch up until it doesn't receive winning tickets from that sender for a while.

I'm not sure if the pace of ticket redemption not matching the rate at which blocks are received is a problem - do you see any issues that may arise? Furthermore, one other option we have if we do believe that the pace of ticket redemption needs to be faster is to redeem a batch of tickets in a single tx (using all the tickets from the queue).

j0sh: > If we don't use an UPSERT Queue.Add would have to check if the ticket already exists.

Attempting to INSERT on an already-existing row should fail (primary key collision) so we can just return that error, rather than doing an explicit check?

kyriediculous: fixed in 726da54ac56c1724864ca99e57791dfc936148fc and 5783c60038ba7cf9e2ceb9226c622d5dbba299ab

I didn't decide to bother fixing up the `recipient` as the end consumer of the ticketqueue because that gets removed in the subsequent PR so I just made sure the unit tests pass for now and will fix up the remainder of the implementation in #1495

common/db.go
```diff
@@ -111,9 +115,27 @@ var schema = `
...
+		faceValue BLOB,
+		winProb BLOB,
+		senderNonce INTEGER,
+		recipientRand BLOB,
+		recipientRandHash STRING,
+		sig BLOB PRIMARY KEY,
+		creationRound int64,
+		creationRoundBlockHash STRING,
+		paramsExpirationBlock int64,
+		submittedAt DATETIME,
```
yondonfu: Minor question: perhaps `redeemedAt` would be more descriptive since we're setting this timestamp after the ticket has been redeemed?

pm/queue.go
```diff
@@ -110,22 +95,41 @@ func (q *ticketQueue) startQueueLoop() {
...
+					})
 					select {
-					case q.redeemable <- nextTicket:
+					case q.redeemable <- &redemption{nextTicket, resCh}:
+						res := <-resCh
+						if res.err != nil {
+							glog.Error(err)
+							continue
+						}
+						q.store.MarkWinningTicketRedeemed(nextTicket, res.txHash)
```
yondonfu: Need to handle error

pm/queue.go
```diff
@@ -110,22 +95,41 @@ func (q *ticketQueue) startQueueLoop() {
...
+			if err != nil {
+				glog.Errorf("Error getting queue length err=%v", err)
+				continue
+			}
 			for i := 0; i < int(numTickets); i++ {
-				q.cond.L.Lock()
-				nextTicket := q.queue[0]
-				q.cond.L.Unlock()
-				q.removeHead()
+				nextTicket, err := q.pop()
```
yondonfu: Seems like this could be replaced with:

```golang
nextTicket, err := q.store.SelectEarliestWinningTicket(q.sender)
```

pm/sendermonitor.go
```diff
@@ -92,7 +93,8 @@ func NewSenderMonitor(claimant ethcommon.Address, broker Broker, smgr SenderMana
 		smgr:            smgr,
 		tm:              tm,
 		senders:         make(map[ethcommon.Address]*remoteSender),
-		redeemable:      make(chan *SignedTicket),
+		redeemable:      make(chan *redemption),
+		ticketStore:     func() TicketStore { return store },
```
yondonfu: If we're adding this state to the struct let's just use the TicketStore directly.

pm/sendermonitor_test.go
```diff
@@ -185,32 +185,43 @@ func TestQueueTicketAndSignalNewBlock(t *testing.T) {
...
+	done := make(chan struct{})
+	go qc.Wait(1, sm, done)
 
 	tm.blockNumSink <- big.NewInt(5)
-	time.Sleep(20 * time.Millisecond)
-
+	<-done
 	// check that ticket is now removed from queue
-	assert.Equal(sm.(*senderMonitor).senders[addr].queue.Length(), int32(0))
+	time.Sleep(time.Millisecond)
```
yondonfu: Let's make this longer (i.e. at least 20) since we know CI can be flaky especially with short sleeps.

yondonfu (changes requested):

pm/queue.go
kyriediculous: fixed in d2a979a8cc8d70cfd436c82a5c06da150c8e9b0c

pm/queue.go
kyriediculous: fixed in 985649238628b9e8d824d2c54280bd2e84da0466

pm/sendermonitor.go
kyriediculous: fixed in 985649238628b9e8d824d2c54280bd2e84da0466

pm/sendermonitor_test.go
kyriediculous: fixed in 985649238628b9e8d824d2c54280bd2e84da0466

common/db.go
kyriediculous: fixed in 4df5f0d769d1bb242c14b39c1fb37bd4165765db

pm/queue.go
```diff
@@ -140,12 +148,14 @@ func (q *ticketQueue) startQueueLoop() {
...
-// removeHead removes the head of the queue
-func (q *ticketQueue) removeHead() {
-	// Lock conditional variable while removing from the queue
-	q.cond.L.Lock()
-	q.queue[0] = nil
-	q.queue = q.queue[1:]
-	atomic.AddInt32(&q.queueLen, -1)
-	q.cond.L.Unlock()
+// pop returns the head of the queue
+func (q *ticketQueue) pop() (*SignedTicket, error) {
```
yondonfu: This can be removed.

kyriediculous:
fixed in a9d18c1a

pm/queue.go
yondonfu: Looks like a9d18c1 updates a test, but does not remove this method.

kyriediculous: I didn't save the file 🤦‍♂️ e90d987

yondonfu (commented):
Changes look good. Will give a LGTM for merge after [this comment](https://github.com/livepeer/go-livepeer/pull/1493#discussion_r434937684) is addressed in #1495.

kyriediculous:
Are you okay with rebasing already?

yondonfu:
> Are you okay with rebasing already?

Sure 👍

kyriediculous:
rebased and added comments for the `TicketStore` API in 41f4eae412371dc8d7e94fecba727cb25ea5bebe

yondonfu (approved):
LGTM

common/db.go
```diff
@@ -33,6 +33,10 @@ type DB struct {
 	unbondingLocks                   *sql.Stmt
 	withdrawableUnbondingLocks       *sql.Stmt
 	insertWinningTicket              *sql.Stmt
+	selectEarliestWinningTicket      *sql.Stmt
+	winningTicketCount               *sql.Stmt
+	markWinningTicketRedeemed        *sql.Stmt
+	removeWinningTicket              *sql.Stmt
```
j0sh, line 39: This particular function is never used?

j0sh (approved):
🚢

PR #1495, opened 2020-05-15
cmd, pm: make sendermonitor the ticketqueue consumer

### LINKED ISSUE #1494: Make sendermonitor the ticketqueue consumer
Since the `SenderMonitor` already has access to the `Broker` interface we can make it the `ticketqueue` consumer , this allows us to unexport `AddFloat` and `SubFloat` from the API.

kyriediculous:

This PR reworks the `SenderMonitor` to become the `ticketqueue` consumer instead of `recipient`. This changes allows us to unexport `AddFloat` and `SubFloat` from the API.

**open question:** how to handle sendernonces? Just run a cleanup loop? 

-  removed `Start()` and `Stop()`
- removed `invalidRands` map from recipient along with `validRand()` method
- make `senderMonitor` the `ticketqueue` consumer
- unexport `AddFloat` and `SubFloat`

adjusted & ran unit tests

Fixes #1494

**Checklist:**


pm/recipient.go
```diff
@@ -120,20 +107,9 @@ func NewRecipientWithSecret(addr ethcommon.Address, broker Broker, val Validator
 		secret:       secret,
 		senderNonces: make(map[string]uint32),
 		cfg:          cfg,
-		quit:         make(chan struct{}),
 	}
 }
 
-// Start initiates the helper goroutines for the recipient
-func (r *recipient) Start() {
```
yondonfu, line 128: We may need to initiate helper goroutines to clean up senderNonces for #1498. I think its worthwhile considering whether that will be necessary now so we don't need to re-introduce these methods & their related code if these methods will be used to initiate & tear down helper goroutines.

pm/sendermonitor.go
```diff
@@ -286,3 +270,76 @@ func (sm *senderMonitor) cleanup() {
...
+func (sm *senderMonitor) redeemWinningTicket(ticket *SignedTicket) (err error) {
+	maxFloat, e := sm.MaxFloat(ticket.Ticket.Sender)
+	if e != nil {
+		err = e
+		return
+	}
+
+	// if max float is zero, there is no claimable reserve left or reserve is 0
+	if maxFloat.Cmp(big.NewInt(0)) <= 0 {
+		sm.QueueTicket(ticket)
```
yondonfu: Error needs to be handled

pm/sendermonitor.go
```diff
@@ -286,3 +270,76 @@ func (sm *senderMonitor) cleanup() {
...
+		err = errors.Errorf("max float is zero")
+		return
+	}
+
+	// If max float is insufficient to cover the ticket face value, queue
+	// the ticket to be retried later
+	if maxFloat.Cmp(ticket.Ticket.FaceValue) < 0 {
+		sm.QueueTicket(ticket)
```
yondonfu: Error needs to be handled

pm/sendermonitor.go
```diff
@@ -286,3 +270,76 @@ func (sm *senderMonitor) cleanup() {
...
+		// Add the ticket face value back to the sender's current max float
+		// This amount is no longer considered pending since the ticket
+		// redemption transaction either confirmed on-chain or was not
+		// submitted at all
+		//
+		// TODO(yondonfu): Should ultimately add back only the amount that
+		// was actually successfully redeemed in order to take into account
+		// the case where the ticket was not redeemd for its full face value
+		// because the reserve was insufficient
+		e := sm.addFloat(ticket.Ticket.Sender, ticket.Ticket.FaceValue)
```
yondonfu: IIUC the reason a named return arg is being used here is so we can return the error returned by `addFloat()` instead of just logging it.

The scenario where `addFloat()` returns an error is when the amount passed in is for some reason greater than a sender's pendingAmount. This should never happen, but the check in `addFloat()` is there just to be safe. An alternative is to instead update `addFloat()` to subtract min(pendingAmount, amount) instead of returning an error and then we wouldn't need to used named return args here.

yondonfu (changes requested):

pm/sendermonitor.go
kyriediculous: I decided to leave it as is for now because we now return an error from both addFloat and subFloat regardless of the math operations as those functions are now responsible for sending the max float update. 

We can always revisit this and make the caller of those (eg `redeemWinningTicket`) responsible for sending updates if necessary.

pm/sendermonitor.go
kyriediculous: included in 9e75eecf but can patch them out separately if you'd like

pm/sendermonitor.go
kyriediculous: included in 9e75eecf but can patch them out separately if you'd like

pm/recipient.go
kyriediculous: good point, I'll drop that commit when I do a rebase.

kyriediculous:
9e75eecf introduces a `SenderMonitor.MonitorMaxFloat(sender ethcommon.Address, sink chan<-  *big.Int) event.Subcription` method

pm/sendermonitor.go
```diff
@@ -19,28 +22,24 @@ var unixNow = func() int64 {
...
 
 	// MaxFloat returns a remote sender's max float
 	MaxFloat(addr ethcommon.Address) (*big.Int, error)
 
 	// ValidateSender checks whether a sender's unlock period ends the round after the next round
 	ValidateSender(addr ethcommon.Address) error
+
+	// MonitorMaxFloat listens for max floats updates for a sender
+	// This method only needs to be "implemented" when the SenderMonitor consumer is a fire-and-forget module for winning tickets
+	MonitorMaxFloat(sender ethcommon.Address, sink chan<- *big.Int) event.Subscription
```
j0sh: Does the sink need to be supplied by the caller, or can it be created and returned from the function? Like this:

```golang
MonitorMaxFloat(sender ethcommon.Address) (sink chan<- *big.Int, event.Subscription)
```

pm/sendermonitor.go
```diff
@@ -19,28 +22,24 @@ var unixNow = func() int64 {
...
+	QueueTicket(ticket *SignedTicket) error
 
 	// MaxFloat returns a remote sender's max float
 	MaxFloat(addr ethcommon.Address) (*big.Int, error)
 
 	// ValidateSender checks whether a sender's unlock period ends the round after the next round
 	ValidateSender(addr ethcommon.Address) error
+
+	// MonitorMaxFloat listens for max floats updates for a sender
+	// This method only needs to be "implemented" when the SenderMonitor consumer is a fire-and-forget module for winning tickets
```
j0sh: Isn't all ticket queuing now fire-and-forget from the perspective of the SenderMonitor consumer? Do you mean when the SenderMonitor itself is a remote / RPC redeemer? (If so, isn't float monitoring still necessary for local redeeming?)

kyriediculous: There's a difference here between `senderMonitor` and `SenderMonitor` (I know this gets confusing when talking about it.). 

The `SenderMonitor` is implemented by `senderMonitor` and `redeemerClient` and is consumed by `recipient` or `Redeemer` (server). 

While `recipient` would be fire and forget, the ticket redemption _server_  is not as it needs to know when the `senderMonitor` updates `maxFloat` for a sender, to send an update to the `redeemerClients` accordingly. Since `AddFloat` and `SubFloat` are no longer part of the `SenderMonitor` API , consumers of this package have no other way of knowing when maxFloat gets updated. 

From discord: 
image

pm/sendermonitor.go
kyriediculous: Having the caller supply the channel makes unit testing a little easier and we have the caller supply the channel for all subscriptions in the repo so it made sense to be consistent.

pm/queue.go
```diff
@@ -113,6 +113,9 @@ ticketLoop:
 					glog.Errorf("Unable to pop ticket from queue err=%v", err)
 					continue ticketLoop
 				}
+				if nextTicket == nil {
```
yondonfu, line 117: Shouldn't this never the case since we only enter the for loop if `numTickets > 0`?

pm/sendermonitor.go
```diff
@@ -280,9 +284,106 @@ func (sm *senderMonitor) cleanup() {
...
+		// This amount is no longer considered pending since the ticket
+		// redemption transaction either confirmed on-chain or was not
+		// submitted at all
+		//
+		// TODO(yondonfu): Should ultimately add back only the amount that
+		// was actually successfully redeemed in order to take into account
+		// the case where the ticket was not redeemd for its full face value
+		// because the reserve was insufficient
+		if err := sm.addFloat(ticket.Ticket.Sender, ticket.Ticket.FaceValue); err != nil {
+			returnErr = err
```
yondonfu: If `addFloat()` returns an error then it will overwrite any other error returned prior to the execution of this defer statement. For example, we would log an `addFloat()` error instead of a tx submission failure error - this would be a loss of useful error information. While the previous approach was harder to test due to the `addFloat()` error not being returned by the method, it had the more desirable behavior of exposing all error information - in the provided example the tx submission failure error would be returned and the `addFloat()` error would be logged.

At the moment, I don't have a solution that doesn't involve logging the `addFloat()` error instead of returning it, but I think that would still be preferable to overwriting errors and potentially losing error information.

yondonfu (changes requested):

pm/queue.go
kyriediculous: We need it I think

pop() will return nil if there are no more tickets remaining even after starting the loop.

pm/sendermonitor.go
kyriediculous: I actually think returning the error would be preferable especially since `subFloat` also returns an error now. 

Alternatively have the caller of these two methods (`sm.redeemWinningTicket`) be responsible for pushing these max float updates to the channel.

pm/sendermonitor.go
j0sh: Hmm, this might just be an implementation detail that's going over my head, but is my understanding correct of the three possible "modes" here?

1. Orchestrator doing local redemptions, using DB backed queue
1. Orchestrator doing RPC redemptions, presumably this is what's fire-and-forget
1. RPC redemption server, using DB backed queue (hits the same code path as the O doing local redemptions, after receiving a ticket from RPC)

Don't all of these need to monitor for the max float? (This is where I'm not quite understanding why `MonitorMaxFloat` is only needed for the second mode, fire-and-forget orchestrators doing RPC redemptions [1].)

1. Orchestrator doing local redemptions still needs to monitor the blockchain if the B's float changes elsewhere, eg winning ticket on another node, or withdrawal, plus local bookkeeping for winning-but-not-yet-cleared tickets
1. Orchestrator doing RPC redemptions receives float updates via RPC monitoring (this interface method?)
1. RPC redemption server "receives" float updates via blockchain *and* redemptions from other orchestrators. Notifies other orchestrators when this happens.

I guess the first and third modes "monitor" the float more implicitly? Spitballing here, but would it help to make this more explicit / symmetrical?

[1] My interpretation of "fire and forget" is that'd be a property of whatever's populating the ticket queue, whether via RPC call or DB insert. That transfers responsibility to whatever's responsible for pulling from the queue for submission, retries, etc. Am guessing that the "fire and forget" term has slightly different connotations here, referring only to the RPC submission method. But maybe that helps visualize how I see the symmetry between RPC and local submission.

j0sh: Another point of confusion here is the fact this takes a broadcaster address while the RPC float monitoring doesn't...

kyriediculous: >  is my understanding correct of the three possible "modes" here?

Yep that describes it well 

> Don't all of these need to monitor for the max float? 

No it depends on the consumer package and whether it needs outside knowledge about when a max float value is updated or not. Not to be confused with receiving updates over the wire protocol. 

in the first setup the `recipient` would be the consumer but that package doesn't care about tracking `maxFloat` it just needs to be able to access that value on the `SenderMonitor` through its API.  

In the third setup we already receive updated max float values through the wire protocol and also have access to a remote call if needed. 

The second setup however, the server, needs to know when the `SenderMonitor` abstraction does internal state updates for its max float so it can send updates over the wire to the clients.

I believe @yondonfu posted a suggestion to switch around the interfaces a bit to make this less confusing but I'm struggling to find it at the moment. 

> Another point of confusion here is the fact this takes a broadcaster address while the RPC float monitoring doesn't...

 I renamed this method to `SubscribeMaxFloat` if it helps

yondonfu: > I believe @yondonfu posted a suggestion to switch around the interfaces a bit to make this less confusing but I'm struggling to find it at the moment.

Sorry I posted a comment, but then deleted it because I wanted to make some edits first. Re-posting below:

IIUC, `MonitorMaxFloat()` is only used by the `Redeemer` in #1513, but it is not required by the current consumers of the `SenderMonitor` API (i.e. `recipient` in the `pm` package). So, `MonitorMaxFloat()` doesn't actually need to be a part of the `SenderMonitor` API. However, we want to use `senderMonitor` in `Redeemer` since it contains the required ticket queuing and max float monitoring functionality. At the moment, we can't directly use `senderMonitor` in `Redeemer` since it is not exported and we only have access to the exported `SenderMonitor` interface. 

What if we did the following:

- Export `senderMonitor` and rename it to something like `LocalSenderMonitor`
- Add `MonitorMaxFloat()` as a method for `LocalSenderMonitor` and do not include it in the `SenderMonitor` interface
- Use `LocalSenderMonitor` in `Redeemer`

I see the following benefits from these changes:

- Avoids adding an additional method to the `SenderMonitor` API when it is not needed by any of the API consumers
- `redeemerClient` in #1513 does not need to implement `MonitorMaxFloat()`

If we use the approach described above, we end up with the following:
- The `SenderMonitor` interface is implemented by `LocalSenderMonitor` and `redeemerClient` (in #1513)
- `Redeemer` (in #1513) uses `LocalSenderMonitor` for ticket queue and max float monitoring functionality
- `LocalSenderMonitor` exposes max float monitoring functionality via a `MonitorMaxFloat()` (or `SubscribeMaxFloat()` if we’re using that naming convention) which is relevant for any user that needs access to max float updates as they happen (i.e. `Redeemer`), but that is *not* relevant for a user that uses the `SenderMonitor` interface
- `redeemerClient` receives max float values from the remote `Redeemer` in the background over the wire and returns the latest max float for a broadcaster via the `SenderMonitor` API method `MaxFloat()`
- Max float monitoring is abstracted away from the `SenderMonitor` API consumer since it is a detail of the `SenderMonitor` implementation. All the `SenderMonitor` API consumer is aware of is a `MaxFloat()` method that will return the latest max float for a broadcaster.

pm/queue.go
yondonfu: Hm we're calling `pop()` only as many times as there are tickets in the queue though. 

Having this nil check here doesn't hurt, but I just wanted to clarify my own understanding of when this code path could be triggered.

pm/sendermonitor.go
yondonfu: > I actually think returning the error would be preferable especially since subFloat also returns an error now.

But, the way the code is currently structured, a tx submission failure error would be overwritten by an `addFloat()` error such that the tx submission failure error would not be logged/exposed to the user. How would you address that?

> Alternatively have the caller of these two methods (sm.redeemWinningTicket) be responsible for pushing these max float updates to the channel.

The `maxFloat()` call (used to get the value to send to the channel) could still return an error and that would need to be handled after `addFloat()` is called so I don't think this addresses the problem of overwriting errors in the `defer` statement.

pm/recipient.go
kyriediculous: We can also re-add them in #1562 instead when we merge that PR, they are included again there anyway.

pm/sendermonitor.go
kyriediculous: sounds good to me, I believe 20ba922e covers the necessary changes in scope for this PR.

pm/sendermonitor.go
kyriediculous: fixed in eed4c5e6

I also changed the behaviour of the subscription in 
d2d3fd1 to make the subscriber responsible for calling `LocalSenderMonitor.MaxFloat` to get the latest value when receiving an update notification.

kyriediculous: Hmm I added a change in the `eth/senderWatcher` by accident in that second commit but I don't think I can undo it without rebasing.

yondonfu: > I also changed the behaviour of the subscription in
d2d3fd1 to make the subscriber responsible for calling LocalSenderMonitor.MaxFloat to get the latest value when receiving an update notification.

Why the change?

> Hmm I added a change in the eth/senderWatcher by accident in that second commit but I don't think I can undo it without rebasing.

You mean in eed4c5e? You could push a commit that removes the change.

pm/sendermonitor.go
```diff
@@ -125,15 +112,16 @@ func (sm *senderMonitor) AddFloat(addr ethcommon.Address, amount *big.Int) error
 	// Subtracting from pendingAmount = adding to max float
 	pendingAmount := sm.senders[addr].pendingAmount
 	if pendingAmount.Cmp(amount) < 0 {
-		return errors.New("cannot subtract from insufficient pendingAmount")
+		glog.Error(errors.New("cannot subtract from insufficient pendingAmount"))
```
yondonfu: If we're logging this error then we don't need to return an error type in this method.

Also, we don't need to wrap the string in an error if we're using `glog.Error`

pm/sendermonitor.go
```diff
@@ -234,11 +223,13 @@ func (sm *senderMonitor) cache(addr ethcommon.Address) {
 // startTicketQueueConsumerLoop initiates a loop that runs a consumer
 // that receives redeemable tickets from a ticketQueue and feeds them into
 // a single output channel in a fan-in manner
-func (sm *senderMonitor) startTicketQueueConsumerLoop(queue *ticketQueue, done chan struct{}) {
+func (sm *LocalSenderMonitor) startTicketQueueConsumerLoop(queue *ticketQueue, done chan struct{}) {
 	for {
 		select {
 		case ticket := <-queue.Redeemable():
-			sm.redeemable <- ticket
+			if err := sm.redeemWinningTicket(ticket); err != nil {
```
yondonfu: Leaving this note for myself that we need to address [this comment](https://github.com/livepeer/go-livepeer/pull/1493#discussion_r434937684).

pm/sendermonitor.go
```diff
@@ -272,17 +263,103 @@ func (sm *senderMonitor) startCleanupLoop() {
...
+		// This amount is no longer considered pending since the ticket
+		// redemption transaction either confirmed on-chain or was not
+		// submitted at all
+		//
+		// TODO(yondonfu): Should ultimately add back only the amount that
+		// was actually successfully redeemed in order to take into account
+		// the case where the ticket was not redeemd for its full face value
+		// because the reserve was insufficient
+		if err := sm.addFloat(ticket.Ticket.Sender, ticket.Ticket.FaceValue); err != nil {
+			glog.Error(errors.Wrap(err, "error adding to max float"))
```
yondonfu: Right now we're logging an error in `addFloat()` and here as well.

We should either:
- Log the error in `addFloat()` and not handle an error here

or

- Return an error in `addFloat()` and log the error here

yondonfu (changes requested):

pm/sendermonitor.go
kyriediculous: > Why the change?

I felt it made slightly more sense to make the subscriber responsible for getting the max float value and the subscription to just notify when the state changes. 

While in essence there is little difference it does allow us to revert to the original `subFloat` and `addFloat` behaviour without an additional `maxFloat` call and instead offloading that responsibility to the receiver of the channel. 

this would also be true for the watchers `watchSenderInfoChange` and `watchPoolSizeChange` (coming in a subsequent commit) , they can just notify instead of sending updated values.

pm/sendermonitor.go
kyriediculous: Oh yes, now returning an error again from addFloat and logging it in the caller dc0e6020

kyriediculous:
d96dd74  & 02f03fd  addresses the necessary event watchers to  `LocalSenderMonitor` to be able to notify its subscribers about max float changes as discussed in https://github.com/livepeer/go-livepeer/pull/1513#discussion_r434128832

pm/sendermonitor.go
kyriediculous: Since we'd need to rebase and force push before we can address this please let me know when you have reviewed the remaining open change request so that I don't overwrite the existing review process by a force push.

pm/sendermonitor.go
kyriediculous: Should we make `LocalSenderMonitor` implement `SenderMonitor` implicitly as well (return struct, not the interface from the constructor) ?

eth/watchers/senderwatcher.go
```diff
@@ -128,6 +134,11 @@ func (sw *SenderWatcher) Clear(addr ethcommon.Address) {
 	}
 }
 
+// SubscribeSenderInfoChange notifies subscribers when the sender info for a particular sender changes
+func (sw *SenderWatcher) SubscribeSenderInfoChange(sink chan<- ethcommon.Address) event.Subscription {
```
yondonfu: The subscription returned by this method only sends updates on the sink channel when the sender's reserve is funded. Should the method name be `SubscribeReserveFunded()` instead?

eth/watchers/senderwatcher_test.go
```diff
@@ -708,3 +708,54 @@ func TestNewRoundEvent_LogRemoved(t *testing.T) {
...
+
+	go sm.Watch()
+	defer sm.Stop()
+	time.Sleep(time.Millisecond)
+
+	// set initial values
+	_, err = sm.GetSenderInfo(stubSender)
+	require.Nil(err)
+
+	watcher.sink <- []*blockwatch.Event{blockEvent}
```
yondonfu: I suspect that the time.Sleep calls here can cause this test to be flaky when run in CI. Could we write this test without using time.Sleep by directly calling the `handleLog()` method?

pm/sendermonitor.go
yondonfu: Sure - don't see any reason why the interface should be returned here instead of the struct.

eth/watchers/senderwatcher_test.go
kyriediculous: No because that would be a cyclical import

(`eth/watchers` imports `pm` and the proposed change would require to import `eth/watchers` in `pm`

Furthermore the `handleLog` method is not exported from `SenderWatcher` anyway and we should not contaminate the API if it's not necessary for external callers other than unit tests I think.

yondonfu: Not sure how a cyclical import is an issue here. Just to make sure we're on the same page:

- I am not suggesting to make any functional changes
- I am suggesting to *only* make changes in this particular test
- I am suggesting to call `SenderWatcher.handleLog()` to trigger a send into `SenderWatcher.senderInfoFeed`. This should accomplish the same thing as the current code in the test that sends into `watcher.sink`. The difference is that we would not need to deal with waiting (via time.Sleep) for the `SenderWatcher` watch loop to process the value sent into `watcher.sink`

kyriediculous:
Return struct instead of interface from NewSenderMonitor e191e13

Return TxHash from `LocalSenderMonitor` redemption loop, to `ticketQueue` 0420149

I already rebased to make progress on the PR, apologies for overwriting the previous review process but I believe most issues were resolved.

eth/watchers/senderwatcher_test.go
kyriediculous: Oh sorry I was looking at the relevant tests in `sendermonitor_test` instead.

kyriediculous: While I included the change currently, I feel like this change request might be a little premature since the test has not proven to be flaky and none of the tests in that file call that method directly and have worked just fine.

eth/watchers/senderwatcher.go
```diff
@@ -244,6 +256,7 @@ func (sw *SenderWatcher) handleLog(log types.Log) error {
 			return fmt.Errorf("GetSenderInfo RPC call to remote node failed: %v", err)
 		}
 		sw.senders[sender] = info
+		sw.senderInfoFeed.Send(sender)
```
yondonfu: This will send updates to subscribers when any of the events in the above switch statement are received and re-orged out. Since `SubscribeReserveChange()` is supposed to only subscribe for updates when a reserve is funded, updates should only be sent into the feed when the `ReserveFunded` event is received.

eth/watchers/senderwatcher.go
```diff
@@ -128,6 +134,11 @@ func (sw *SenderWatcher) Clear(addr ethcommon.Address) {
 	}
 }
 
+// SubscribeReserveChange notifies subscribers when the sender info for a particular sender changes
+func (sw *SenderWatcher) SubscribeReserveChange(sink chan<- ethcommon.Address) event.Subscription {
+	return sw.senderInfoScope.Track(sw.senderInfoFeed.Subscribe(sink))
```
yondonfu: `reserveChangeScope`/`reserveChangeFeed` may be a better names than `senderInfoScope`/`senderInfoFeed` given how these fields are used.

yondonfu (changes requested):

eth/watchers/senderwatcher_test.go
yondonfu: Ah after taking a second look at that test I realized that the assertions are placed after a statement that blocks until a value is received on a channel (i.e. `sender := <-sink`). Since the assertion is not run until after a value is received on a channel flaky test behavior from waiting for goroutines shouldn't be a problem. 

Thanks for submitting the commit that addresses the original comment though - it is still useful as it more directly triggers the desired state update (in `handleLog()`) which helps test readability.

eth/watchers/senderwatcher.go
kyriediculous: I added a conditional.  66c5fa90

Do you want to add a unit test that checks we don't send a value on the channel when the event is not a ReserveFunded event or is that trivial ?

eth/watchers/senderwatcher.go
kyriediculous: fixed in 66c5fa90

eth/watchers/senderwatcher.go
yondonfu: > Do you want to add a unit test that checks we don't send a value on the channel when the event is not a ReserveFunded event or is that trivial ?

Yeah let's add a test for this to make sure we only send values to the `reserveChangeFeed` for `ReserveFunded` events.

kyriediculous: Added a test in eee7f2187d591ecf748ebd3331c439085dfd1833 (I decreased the wait time to 1 second in the subsequent commit).

go.sum
```diff
@@ -72,6 +72,7 @@ github.com/golang/glog v0.0.0-20160126235308-23def4e6c14b h1:VKtxabqXZkF25pY9ekf
 github.com/golang/glog v0.0.0-20160126235308-23def4e6c14b/go.mod h1:SBH7ygxi8pfUlaOkMMuAQtPIUF8ecWP5IEl/CR7VP2Q=
 github.com/golang/groupcache v0.0.0-20190702054246-869f871628b6 h1:ZgQEtGgCBiWRM39fZuwSd1LwSqqSW0hOdXCYYDX0R3I=
 github.com/golang/groupcache v0.0.0-20190702054246-869f871628b6/go.mod h1:cIg4eruTrX1D+g88fzRXU5OdNfaM+9IcxsU14FzY7Hc=
+github.com/golang/mock v1.1.1 h1:G5FRp8JnTd7RQH5kemVNlMeyXQAztQ3mOWV95KxsXH8=
```
yondonfu: Is this update necessary for this PR?

kyriediculous: Not sure where this exactly snuck in

kyriediculous: removed in cb035165, ready for rebase ?

yondonfu (commented):
👍  for a rebase

kyriediculous:
rebased

yondonfu (approved):
LGTM 🚢  pending green CI (which I think should be fixed by rebasing with master to pick up the latest fix for `TestSubmitSegment_HttpPostError`).

PR #1499, opened 2020-05-17
./install_ffmpeg.sh needs curl

chrishobcroft:
`curl` is required as part of `./install_ffmpeg.sh` script and the process fails without it.

It is not included in other packages, so must be explicitly installed for the process to work.

- 
- 
- 

**Checklist:**


j0sh (approved):
Thanks!

Issue #1496, opened 2020-05-15
Trailing moov breaks standalone T


j0sh:
**Describe the bug**

Remote (standalone) transcoders can not transcode MP4 files if the file does not have a moov atom at the beginning of the file. This is because the transcoder pulls directly from the orchestrator URL, which is unable to peek towards the end if the file is served over HTTP. Combined OT nodes do not have this problem because segments are written to disk.

**To Reproduce**
1. Set up a B, non-transcoding O and standalone T
1. Generate a mp4 with a trailing moov, eg `ffmpeg -i <input> -c:v libx264 out.mp4`
2. Submit for transcoding: `curl --data-binary "@out.mp4" http://localhost:8935/live/name/0.mp4`
3. Observe errors on T, eg:

```
[mov,mp4,m4a,3gp,3g2,mj2 @ 0xa001800] stream 0, offset 0x30: partial file
[mov,mp4,m4a,3gp,3g2,mj2 @ 0xa001800] Could not find codec parameters for stream 0 (Video: h264 (avc1 / 0x31637661), none, 1280x720, 496 kb/s): unspecified pixel format
Consider increasing the value for the 'analyzeduration' and 'probesize' options
```

Also, for some reason the orchestrator waits for the transcoder to time out here, so that should be looked at too: if there are no additional transcoders then the O should return an error immediately.

**Expected behavior**

Ideally, transcoding should happen regardless. Few ways to do this:

* Always download disk to disk for transcoders (whether combined OT or standalone). Probably the easiest method, but slowest. Tuning by the user may mitigate this, such as use of a tmpfs, but it's not automatic. Related: https://github.com/livepeer/go-livepeer/issues/1405 , https://github.com/livepeer/go-livepeer/issues/846

* Probe the segment on the broadcaster. This increases latency but may help usability. This would also be helpful for other things such as obtaining the segment duration for HTTP push. After probing fails, we could:
  * Return an informative error to the user -  prior to transcoding
  * Attempt to correct the issue on the broadcaster, if enough information is available.

PR #1507, opened 2020-05-19
core: Check for remote transcoding when using disk.

j0sh:
Nvidia and other local transcoding types no longer go through
the network.

Partially mitigates https://github.com/livepeer/go-livepeer/issues/1496 (but does not fully fix it)

### DIFF
--- a/core/orchestrator.go
+++ b/core/orchestrator.go
@@ -487,9 +487,9 @@ func (n *LivepeerNode) transcodeSeg(config transcodeConfig, seg *stream.HLSSegme
 	transcoder := n.Transcoder
 
 	var url string
-	_, isLocal := transcoder.(*LocalTranscoder)
+	_, isRemote := transcoder.(*RemoteTranscoderManager)
 	// Small optimization: serve from disk for local transcoding
-	if isLocal {
+	if !isRemote {
 		url = fname
 	} else if drivers.IsOwnExternal(seg.Name) {
 		// We're using a remote TC and segment is already in our own OS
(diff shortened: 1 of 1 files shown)


darkdarkdragon (approved):
LGTM

jailuthra (approved):
LGTM

kyriediculous:
From a first glance this doesn't really change anything except for semantics and is actually limiting because users might be implementing different types of remote transcoder manager-like types , such as Livepool does. 

If this truly doesn't change anything except for semantics I'd like to revert this.

j0sh:
> From a first glance this doesn't really change anything except for semantics

What do you mean, change semantics? The behavior changes: non-remote transcoders (nvidia, cpu) are both read segments from disk now.

Without this change, only cpu transcoders were reading from disk. Nvidia transcodes were being pulled via O http, which broke with certain inputs because of https://github.com/livepeer/go-livepeer/issues/1496

Now, you might prefer the check to be `LocalTranscoder || NvidiaTranscoder` . The reason that wasn't done is because as soon as vaapi lands (or some other type of local transcoding), this line is going to have to be updated as well, if #1496 isn't fixed by then. Explicitly enumerating the supported formats seems brittle, especially since it fixes a rather subtle bug.

But on the other hand, if we add another remote transcoder type (or change the current behavior), it'll immediately break. That can be more easily diagnosed and corrected. That seems preferable to leaving room for a rather subtle bug to recur.

> users might be implementing different types of remote transcoder manager-like types

Cool - but that also involves changes to the go-livepeer codebase, doesn't it? So that seems a bit outside the scope here, because this particular spot isn't really a public interface boundary for third-party extensions. We don't do "pluggable transcoders" beyond what has already been defined in this codebase.

kyriediculous:
Thanks for clarifying @j0sh , from first glance it just seemed like we were inverting the boolean expression.

Issue #1478, opened 2020-04-30
Transcode long segments


j0sh:
## Abstract

Much of the code base was built around the assumption of processing [2-second segments](https://github.com/livepeer/go-livepeer/blob/0ee99d8905d461d734b638b0f5b533caa3e2a016/server/mediaserver.go#L57). Transcoding longer segments is a useful capability to have. One of the barriers to handling longer segments is our use of a fixed timeout for transcode requests, set at 8 seconds in a few places around the codebase: ([1](https://github.com/livepeer/go-livepeer/blob/0ee99d8905d461d734b638b0f5b533caa3e2a016/server/segment_rpc.go#L45)) ([2](https://github.com/livepeer/go-livepeer/blob/0ee99d8905d461d734b638b0f5b533caa3e2a016/core/orchestrator.go#L682-L692))([3](https://github.com/livepeer/go-livepeer/blob/0ee99d8905d461d734b638b0f5b533caa3e2a016/server/segment_rpc.go#L110-L116)).

Rather than use a fixed timeout, the segment duration should be used as the basis for the timeout.

Accommodating longer segments via expanded timeouts can be accommodated on the broadcaster in a relatively low-risk manner. A PR is open for this: https://github.com/livepeer/go-livepeer/pull/1477

Accommodating longer segments on the orchestrator, however, is more fraught due to the risk of DoS and under-payment.

This spec seeks clarity on a few issues:

* Timeout multiplier
* Hard cap on segment lengths
* Orchestrator processing of longer segments in on-chain mode
* Any other considerations that may have been overlooked.

## Segment Duration

We will depend on the segment duration as supplied by the HLS segmenter (for RTMP ingest) or the `Content-Duration` header for HTTP ingest. The default length for a missing Content-Duration remains at 2 seconds ([1](https://github.com/livepeer/go-livepeer/blob/0ee99d8905d461d734b638b0f5b533caa3e2a016/server/mediaserver.go#L667-L671)).

If the `Content-Duration` is missing, then the received segment could be probed in an attempt to determine its duration. However, implementing this probe is out of scope right now, as users have indicated they already have the information necessary to supply the duration. Note there are some drawbacks to probing: increased CPU processing requirements, increased latency, and a requirement to have the segment available on the broadcaster, which would negate any advantage from use cases such as pre-posting the segment to an object store so the broadcaster only needs to pass through the URL without fully downloading the segment.

[...]

PR #1508, opened 2020-05-20
server: Estimate input framerate in fee estimation for fps passthrough.

j0sh:

There is oversight with fee estimation when doing FPS pass-through. FPS pass-through is signaled via a `Framerate` value of 0, which would result in an estimated fee of zero.

Since we don't yet have the exact source frame rate (or the number of source frames), we just make a conservative guess of 120fps. This should ensure the estimated fee is sufficient in most normal scenarios.

Also added webhook tests because I needed to double check that it actually worked, and didn't see any corresponding unit tests.

- server: Estimate input framerate in fee estimation for fps passthrough. cc390ee
- server: Add webhook tests for fps passthrough. e6d65ce

- Unit tests

Addresses some issues mentioned in https://github.com/livepeer/go-livepeer/issues/1478 around fee estimation.

**Checklist:**

### DIFF
--- a/core/lb.go
+++ b/core/lb.go
@@ -215,7 +215,12 @@ func calculateCost(profiles []ffmpeg.VideoProfile) int {
 		if err != nil {
 			continue
 		}
-		cost += w * h * int(v.Framerate) // TODO incorporate duration
+		framerate := int(v.Framerate)
+		if 0 == framerate {
+			// passthrough; estimate 30fps for load balancing purposes
+			framerate = 30
+		}
+		cost += w * h * framerate // TODO incorporate duration
 	}
 	return cost
 }
--- a/core/lb_test.go
+++ b/core/lb_test.go
@@ -14,6 +14,15 @@ import (
 	"github.com/livepeer/lpms/ffmpeg"
 )
 
+func TestLB_CalculateCost(t *testing.T) {
+	assert := assert.New(t)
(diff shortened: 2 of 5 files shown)


jailuthra (approved):
LGTM

Will it be feasible to use `ffprobe` or something similar to get a framecount/framerate of the source video here?

Another (possibly unrelated) place we should probably use a hardcoded fps value is the load balancing [cost estimation](https://github.com/livepeer/go-livepeer/blob/0ee2557/core/lb.go#L218)

j0sh:
> Will it be feasible to use ffprobe or something similar to get a framecount/framerate of the source video here?

We can incorporate probing at some point - we'll probably need this sooner rather than later for other things such as duration estimation in HTTP push, rather than always expecting the user to supply it. Likely needs to be something in LPMS though, because we can't expect ffmpeg / ffprobe to be available on the user's system.

j0sh:
> Another (possibly unrelated) place we should probably use a hardcoded fps value is the load balancing cost estimation

Good point, added this in 87e3613 . Used a smaller estimate of 30fps for this case, since I'm not quite sure if there's a benefit to substantially over-estimating this for load balancing.

darkdarkdragon (approved):
LGTM

yondonfu (approved):
LGTM

Issue #1474, opened 2020-04-27
Non Integer Frame Rate Support


j0sh:
TODO:
- [x] Spec
- [x] Implementation for LPMS
- [ ] Implementation for go-livepeer

Notes:
- Configurable via auth webhook
- Propagation through codebase via VideoProfile, similar to MP4
- Add `fps_den` field (rational number representation)
- Example: MP4 [spec](https://github.com/livepeer/go-livepeer/issues/1418), MP4 implementation ([1](https://github.com/livepeer/go-livepeer/pull/1429])) ([2](https://github.com/livepeer/go-livepeer/pull/1439)) ([3](https://github.com/livepeer/lpms/pull/176))

iameli:
I'd like to put in a vote on the webhook spec for representing the fraction as a two-element tuple rather than an additional `fps_den` field.

```json
{
  "profiles": [
    {
      "name": "test_stream_29.97fps",
      "width": 1280,
      "height": 720,
      "bitrate": 4000000,
      "fps": [30000, 1001]
    },
    {
      "name": "test_stream_30fps",
      "width": 1280,
      "height": 720,
      "bitrate": 4000000,
      "fps": 30
    }
  ]
}
```

Bit more terse and readable IMO.

j0sh:
The trouble is, this makes it more difficult to statically define the JSON schema to generate de/serializers. In googling around how to do this for golang, the consensus is basically "this is bad practice; don't do it."  Eg: https://medium.com/@nate510/dynamic-json-umarshalling-in-go-88095561d6a0

jailuthra:
## Abstract

Currently go-livepeer supports transcoding videos to integer framerates. This is a proposal to add support for arbitrary non-integer framerates in the output video profiles.

## Motivation

Some broadcasting standards (NTSC/ATSC) work on esoteric non-int framerates like 29.97 etc. As these standards are still widely used, users want the ability to transcode videos to non-int framerates through go-livepeer.

## Proposed Solution

* Accept non-integer framerates as a rational number (numerator/denominator) in go-livepeer.
* Propogate the FPS value through the VideoProfile struct, pass it on as an AVRational to FFmpeg.
* Ability to set non-int FPS through the auth webhook.

## Implementation Tasks and Considerations

### LPMS Changes

#### Cgo

[...]

PR #1511, opened 2020-05-21
Use seg duration to time out remote transcoders.

j0sh:

As described in https://github.com/livepeer/go-livepeer/issues/1478 , orchestrator timeouts for a remote T need to be adjusted to handle longer segment durations.

In addition to transmitting the duration from the B to the O in [`net.SegData`](https://github.com/livepeer/go-livepeer/commit/3358454a3099bdfbf747ca8e0071ed0f3f74713f#diff-afa2cdc3d0285595d3ca53db8b21226fR126) , the duration also needs to be made available to the orchestrator's [remote transcoder manager](https://github.com/livepeer/go-livepeer/commit/21568b10c3761fe80ed47bd03c4d576306feb725#diff-b76fbd82bbf5b7a4d2fdf59ecd8d4ca6R689) to make use of it. The remote transcoder manager follows the [Transcoder interface](https://github.com/livepeer/go-livepeer/commit/b1189c879a9962caac5df9e82b071c269a4d9265#diff-418dc657e12875665c93dffe96a0e464L23), which offers nowhere to put the duration:

```golang
Transcode(job string, fname string, profiles []ffmpeg.VideoProfile) (*TranscodeData, error)
```

Since both the orchestrator and broadcaster make use of `core.SegTranscodingMetadata` (CSTMD), which has similar fields as `net.SegData`, we formalize this relationship [1] and extract mappings ( [1](https://github.com/livepeer/go-livepeer/commit/445797907c0a692cccc1938b351190a31d65df26#diff-d8e92e957c82b6b03540a4be02e23720R255), [2](https://github.com/livepeer/go-livepeer/commit/445797907c0a692cccc1938b351190a31d65df26#diff-d8e92e957c82b6b03540a4be02e23720R494) ) to / from each: the latter for propagation over the wire, and the former for propagation throughout the codebase.

Having the `Transcoder` interface take a `CSTMD` effectively subsumes the `job` and `profiles` parameters, with `CSTMD.ManifestID` and `CSTMD.Profiles` respectively. From there, there is little reason not to also [incorporate the `fname`](https://github.com/livepeer/go-livepeer/commit/b9d39074ea3dd321d792725bc8470b26d755d670#diff-bc0ce8ba055f1f8945691e8ff8a6974dR26) parameter, resulting in the this  interface:

```
Transcode(md *SegTranscodingMetadata) (*TranscodeData, error)
```

The O-T protocol is also [adjusted to take `net.SegData`](https://github.com/livepeer/go-livepeer/commit/445797907c0a692cccc1938b351190a31d65df26#diff-afa2cdc3d0285595d3ca53db8b21226fR214). Although not strictly necessary since the standalone transcoder does not currently require the duration [2], this has the side effect of a small reduction in technical debt for standalone T, due to reuse of the mappings between `net.SegData` and `CSTMD` . Any fields that are added to these structs, such as a denominator for the framerate (https://github.com/livepeer/go-livepeer/issues/1474#issuecomment-630952970) will be available by default to standalone T. Back-compat logic that is developed within the mappings will also carry over.

[1] We could have opted to carry `net.SegData` throughout the codebase rather than making use of `CSTMD`, but the back-compat in `net.SegData` requires some translation before its fields can be used within the codebase, eg around profiles. This gives us the ability to evolve each structure as needed. We've already done so with the incorporation of the `Fname` field to CSTMD which isn't necessary over the wire.

[2] The transcoder will need the duration as soon as it's ready to be incorporated into the load balancer.

- cmd, core: Typo fixes. 286a520
- net: Regenerate proto with segment duration. 5f03718
- core, server: Use net.SegData in the O-T protocol. … 4457979
- server, net: Send segment duration over the wire. 3358454
- server: Refactor test cases to use coreSegMetadata. … fe29742
- core: Add input filename to SegTranscodingMetadata. … b9d3907
- core, server: Pass SegTranscodingMetadata to transcoder. b1189c8
- core: Set remote transcoder timeout based on segment duration. 21568b1

- Unit testing
- Manual testing

**Checklist:**


server/rpc.go
```diff
@@ -305,3 +305,46 @@ func pmTicketParams(params *net.TicketParams) *pm.TicketParams {
...
+		glog.Error("Unable to deserialize profiles ", err)
+		return nil, err
+	}
+
+	var os *net.OSInfo
+	if len(segData.Storage) > 0 {
+		os = segData.Storage[0]
+	}
+
+	dur := time.Duration(int64(segData.Duration) * int64(time.Millisecond))
```
darkdarkdragon:
```suggestion
	dur := time.Duration(segData.Duration) * time.Millisecond
```

net/lp_rpc.proto
```diff
@@ -211,16 +214,21 @@ message NotifySegment {
...
     int64 taskId   = 16;
 
-    // Set of profiles to transcode this segment into.
+// Deprecated fields. May still be populated sometimes for back compat
+
+    // Deprecated by segData. Job the segment belongs to.
+    string job      = 2;
+
+    // Deprecated by fullProfiles. Set of presets to transcode into.
     bytes profiles = 17;
```
darkdarkdragon, line 230: recommended way of deprecating fields
```suggestion
    bytes profiles = 17 [deprecated=true];
```

net/lp_rpc.proto
```diff
@@ -211,16 +214,21 @@ message NotifySegment {
...
 
-    // Transcoding profiles to use. Supersedes `profiles` field

+    // Deprecated by segData. Transcoding configuration to use.
     repeated VideoProfile fullProfiles = 33;
```
darkdarkdragon:
```suggestion
  repeated VideoProfile fullProfiles = 33 [deprecated=true];
```

net/lp_rpc.proto
```diff
@@ -211,16 +214,21 @@ message NotifySegment {
...
+    SegData segData = 3;
 
     // ID for this particular transcoding task.
     int64 taskId   = 16;
 
-    // Set of profiles to transcode this segment into.
+// Deprecated fields. May still be populated sometimes for back compat
+
+    // Deprecated by segData. Job the segment belongs to.
+    string job      = 2;
```
darkdarkdragon: recommended way of deprecating field that's not used anymore
```suggestion
    // string job      = 2;
    reserved 2;
    reserved "job";
```

darkdarkdragon (changes requested):
Looks good, just couple of small suggestions

server/rpc.go
j0sh: Fixed in b7a6479c1ec41cf2a373d629eff8d1a687d40ec7

net/lp_rpc.proto
j0sh: Very nice - didn't know about this annotation! Actually, kept this line in place since we still "use" it via setting an invalid value to induce failures. Added a comment to the protobuf indicating this in 4d957856cdf5136a1e21972169e9f50d64e244c4

According to the protobuf docs, this is a no-op in all languages except Java, where it emits a warning. My concern is that if golang ever picks this up - golang only has errors, no warnings -  this could cause code to stop compiling and the annotation would have to be removed. I'm not entirely sure about the implications of this, so it seems safer to just keep things as-is.

net/lp_rpc.proto
j0sh: Made this `reserved` in 4d957856cdf5136a1e21972169e9f50d64e244c4

net/lp_rpc.proto
j0sh: Good idea, reserved the number in 4d957856cdf5136a1e21972169e9f50d64e244c4 . Didn't reserve the field name since we don't do JSON, so there should not be a conflict. (Can still add that in if you feel it should be there though.)

j0sh:
Addressed comments via fixup, and also made a clean-up pass through the protobuf file to remove trailing whitespace in 98ab2120512f1dbad66da60fa0a784c93524db17. Cosmetic changes; nothing functional.

darkdarkdragon (approved):
LGTM

net/lp_rpc.proto
```diff
@@ -211,17 +214,23 @@ message NotifySegment {
...
-    // Job the segment belongs to.
-    string job      = 2;
+    // Configuration for the transcoding job
+    SegData segData = 3;
 
     // ID for this particular transcoding task.
     int64 taskId   = 16;
 
-    // Set of profiles to transcode this segment into.
+// Deprecated fields. May still be populated if necessary
```
yondonfu: Needs an extra indent?

server/ot_rpc.go
```diff
@@ -131,22 +129,21 @@ func runTranscoder(n *core.LivepeerNode, orchAddr string, capacity int) error {
...
 	glog.Infof("Transcoding taskId=%d url=%s", notify.TaskId, notify.Url)
 	var contentType string
 	var body bytes.Buffer
 
-	tData, err := n.Transcoder.Transcode(notify.Job, notify.Url, profiles)
+	md, err := coreSegMetadata(notify.SegData)
+	if err != nil {
+		glog.Error("Unable to parse segData err=", err)
+		md = &core.SegTranscodingMetadata{} // avoid crash.
+		// TODO short-circuit error handling
```
yondonfu, line 141: Perhaps we can short-circuit error handling by doing:

```golang
if err != nil {
    glog.Error("Unable to parse segData err=", err)
}
var tData *core.TranscodeData
if err == nil {
    tData, err = n.Transcoder.Transcode(md)
    ...
    if err == nil && len(tData.Segments) != len(profiles) {
        ...
    }
}
if err != nil {
    // Write error to body
}
```

j0sh: Yeah it was the large reflow of code I was hoping to avoid, plus additional work to test the error code etc. Error handling is generally broken at this point, though; this isn't an isolated case. One of the larger tech-debty things in the O/T protocol.

net/lp_rpc.proto
j0sh: Was kind of intentional to indicate scope, "everything past this point is deprecated" but can re-indent

yondonfu: Got it. I don't feel strongly about the indentation so I'll leave that up to you.

server/ot_rpc.go
yondonfu: Yeah the suggestion above was the least non-invasive change that came to mind, but definitely hear you that the structure could be cleaner/easier to test.

I'm fine with not going into a rabbit hole for this PR to fix all error handling in the O/T protocol. Created #1518 to track it - the description is pretty brief so feel free to add any relevant details there.

net/lp_rpc.proto
j0sh: Fixed indentation in 89a3acfe2b7155b203e0945682bb2b98f9e3192c and also improved the comment

server/ot_rpc.go
j0sh: Thanks for making the ticket. Added a reference to the ticket in the code - 89a3acfe2b7155b203e0945682bb2b98f9e3192c

yondonfu (approved):
LGTM pending green CI

j0sh:
Rebased, merged

PR #1513, opened 2020-05-24
net, server: Ticket Redemption Server & Client 

### LINKED ISSUE #1487: Create ticket redemption service server implementation
Use the new `senderMonitor` implementation from #1494 for the server implementation for the ticket redemption service. 

- Write the protobuf definitions
- Write the server implementation backed by the `senderMonitor`

### LINKED ISSUE #1515: Create ticket Redemption Service client implementation
The client implementation should implement the `SenderMonitor` interface as well as the necessary RPC calls `MonitorMaxFloat` and `QueueTicket`

kyriediculous:

- Server implementation for the Ticket Redemption gRPC service using the existing `senderMonitor`implementation from #1495 
- Client implementation for the ticket redemption service that implements the `SenderMonitor` interface except for the `MonitorMaxFloat` method. 

Will update once I clean up the commit history 

Fixes #1487 
Fixes #1515 

**Checklist:**


net/redeemer.proto
```diff
@@ -0,0 +1,29 @@
...
+    bytes recipient_rand_hash = 6;
+    int64 creation_round = 7;
+    bytes creation_round_block_hash = 8;
+    bytes sig = 9;
+    bytes recipient_rand = 10;
+    int64 params_expiration_block = 11;
+}
+
+message MaxFloatUpdate {
+        bytes sender = 3;
```
kyriediculous: Need to update this

server/redeemer.go
```diff
@@ -0,0 +1,303 @@
...
+
+const rpcTimeout = 8 * time.Second
+
+type redeemer struct {
+	recipient   ethcommon.Address
+	subs        sync.Map
+	eth         eth.LivepeerEthClient
+	sm          pm.SenderMonitor
+	quit        chan struct{}
+	liveSenders sync.Map
```
kyriediculous, line 34: How to clean this up ? 
No way to feasibly know when the queue for a sender is feasibly empty 

Why do we have this mapping in the first place?
`QueueTicket` fires a watchdog, but these are per sender and not per ticket. We'd want to avoid multiple watchdogs consuming the same channel to perform the same operation. The first to read from the channel will send and update to the subscribed clients but this spins up a lot of go routines until a ticket is redeemed. Therefore if a sender already exists in `liveSenders`  we don't spin up an additional watchdog and clean up the goroutine.

server/redeemer.go
```diff
@@ -0,0 +1,303 @@
...
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/peer"
+	"google.golang.org/grpc/status"
+)
+
+const rpcTimeout = 8 * time.Second
+
+type redeemer struct {
+	recipient   ethcommon.Address
+	subs        sync.Map
```
kyriediculous: I guess we could use `event.Subscription` here , but since this is third party should we really be relying on it too much?

server/redeemer.go
```diff
@@ -0,0 +1,303 @@
...
+			return nil
+		case <-stream.Context().Done():
+			return nil
+		}
+	}
+}
+
+type redeemerClient struct {
+	rpc      net.TicketRedeemerClient
+	maxFloat map[ethcommon.Address]*big.Int
```
kyriediculous: Run a watchdog to clean up this map?

kyriediculous:
c634bcbb adds a cleanup loop to clean up the `liveSenders` map if a ticket hasn't been received from a sender since the loop last ran.

Makefile
```diff
@@ -2,7 +2,7 @@ SHELL=/bin/bash
 
 all: net/lp_rpc.pb.go core/test_segment.go livepeer livepeer_cli
 
-net/lp_rpc.pb.go: net/lp_rpc.proto
+net/lp_rpc.pb.go: net/*.proto
```
j0sh: Make targets look at *outputs*, and running protoc on `redeemer.proto` will generate `redeemer.pb.go`, rather than `lp_rpc.pb.go`. Also, using `net/*.proto` is telling Make that all the proto files are dependent in order to generate `lp_rpc.pb.go`, which isn't correct either; it only depends on that one `lp_rpc.proto` file.

Try something like this (untested)

```make
all: net/lp_rpc.pb.go net/redeemer.pb.go core/test_segment.go ...
net/%.pb.go: net/%.proto
    protoc -I=. --go_out=plugins=grpc:. $^
```

(although tbh I'd be fine with just keeping everything in that `lp_rpc.proto` file as well)

net/redeemer.proto
```diff
@@ -0,0 +1,34 @@
+syntax = "proto3";
+
+import "google/protobuf/empty.proto";
+
+package net;
+
+service TicketRedeemer {
+    rpc QueueTicket(Ticket) returns (google.protobuf.Empty) {}
```
j0sh: Stub out an empty response type here, rather than using `Empty`.  We should be acking the request anyway to ensure the ticket was actually written to disk. This also gives us room to add fields to the response if necessary without breaking compatibility. Something like:

```proto
message QueueAck{
}
rpc QueueTicket(Ticket) returns (QueueAck) {}
```

net/redeemer.proto
```diff
@@ -0,0 +1,34 @@
...
+    rpc MaxFloat(MaxFloatRequest) returns (MaxFloatUpdate) {}
+    rpc MonitorMaxFloat(google.protobuf.Empty) returns (stream MaxFloatUpdate) {}
```
j0sh: Same goes for the use of `google.protobuf.Empty` here - let's stub out an empty request parameter.

(Perhaps I don't quite understand yet how the pieces here come together, but wouldn't we want to parameterize MonitorMaxFloat by the broadcaster address?)

kyriediculous: > (Perhaps I don't quite understand yet how the pieces here come together, but wouldn't we want to parameterize MonitorMaxFloat by the broadcaster address?)

That would mean a listener for each sender rather than a single listener for all updates though.

j0sh: Is it an issue to have a listener for each sender? Having a single listener for all updates means each orchestrator node gets updates from all senders across the entire multi-O fleet, even though a particular orchestrator doesn't need the majority of that data. That's pretty chatty and will have a lower scalability ceiling than a listener-per-sender.

kyriediculous: I believe we can still use a single listener and the service as is. 

The difference between sending a message for a `sender` to all clients instead of to only clients that are working with that sender would simply require additional bookkeeping on the server using a map that can be cleaned up on the same interval as the `liveSenders` map. 

We would have to alter `MonitorMaxFloat` to take in a `message MonitorMaxFlotReq{ bytes sender = 1; }`. 

The functionality would be as it is now , if there is already a stream existing for that client we will just use that stream to push updates. Additionally we will now also add the sender to the map for which that client wants to receive updates.

net/redeemer.proto
```diff
@@ -0,0 +1,34 @@
...
+    bytes recipient_rand_hash = 6;
+    int64 creation_round = 7;
+    bytes creation_round_block_hash = 8;
+    bytes sig = 9;
+    bytes recipient_rand = 10;
+    int64 params_expiration_block = 11;
+}
+
+message MaxFloatUpdate {
+        bytes sender = 1;
```
yondonfu: Let's fix the indentation and spacing here.

net/redeemer.proto
kyriediculous: Looked at this some more

> even though a particular orchestrator doesn't need the majority of that data. That's pretty chatty and will have a lower scalability ceiling than a listener-per-sender.

There's nothing chatty about it in my opinion, it's one way traffic. It's essentially a gossip protocol which scales very well with hundreds of clients and hundreds of messages. We'll never even reach such scale because it pertains to _winning_ tickets and a _multi-O_ setup. An O controls its winning probability and amount of clients in the cluster. If you have high win probability for tickets you should alter your parameters, if you have many O's in your fleet, consider running a multi-T setup instead.  

I also think it's not entirely correct to say they don't need the majority of the data. They _might_ not need it , but they might as well need it due to load balancing: 

- Client side load balancing: All O's supplied with `orchAddr` will need to discover the max float for the broadcaster, not pushing updates to all clients means doing remote calls.

- Server side load balancing:  If the request gets directed to another node during the transcode workflow we now have to do a remote request on that other node to get the maxfloat for B even though the cluster was already aware. 

All in all I think only pushing particular updates to clients is a premature optimisation and leads to possible synchronisation issues that need to be aleviated with remote calls.

net/redeemer.proto
kyriediculous: I think I replied to this somewhere else as well. 

`empty.Empty` is designed exactly for this case and will not break backwards compatibility of the wire protocol. 

https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#empty

gRPC will return an error and internal error status code when we run into an error throughout the operation. If there is no error then an empty object and status OK will be returned.

j0sh: > will not break backwards compatibility of the wire protocol.

Are you certain of that? The documentation doesn't claim anything about back compatibility. Some quick googling indicates this could break, and the use of `Empty` is generally not encouraged:

https://stackoverflow.com/questions/50993815/is-google-protobuf-empty-dangerous-for-backwards-compatibility

By the author of Protocol Buffers v2 and Cap'n Proto: https://stackoverflow.com/questions/31768665/can-i-define-a-grpc-call-with-a-null-request-or-response/31772973#comment51526939_31772973

net/redeemer.proto
j0sh: In general, writing to the network is a bit like writing to the blockchain: avoid if possible.  But I *think* I can see the reasoning for doing that here, especially coupled with this note https://github.com/livepeer/go-livepeer/pull/1521#discussion_r432424537 .

A few more questions, just to clarify my own understanding:

> Client side load balancing: All O's supplied with orchAddr will need to discover the max float for the broadcaster, not pushing updates to all clients means doing remote calls.

* You mean `redeemerAddr` ?
* Wouldn't a remote call still be necessary in most cases, in order to obtain the max float during discovery? How is the load balanced here?

> Server side load balancing: If the request gets directed to another node during the transcode workflow we now have to do a remote request on that other node to get the maxfloat for B even though the cluster was already aware.

* That is true only if a winning ticket has been submitted by the first node? In most cases, a remote call would still be necessary for pricing info?

So altogether, it seems like we're only saving ourselves RPC calls for the (relatively infrequent) case where a winning ticket is received? Anyway, for now, keeping things relatively stateless seems okay. Just ask that we also incorporate https://github.com/livepeer/go-livepeer/pull/1513#discussion_r431337263 so we have room to make modifications later if necessary.

Please let me know if the following flow is correct (it's a bit hard to piece everything together from the various PRs, and I don't see a canonical description / doc anywhere):

1. On startup, O subscribes to `MonitorMaxFloat` on a RPC redeemer for float updates
1. During discovery / GetOrchestratorInfo , an O invokes MaxFloat RPC and caches the resulting value for the given B. Only that one O gets the MaxFloat value. (How long is the value kept in cache?)
1. Anytime a winning ticket is submitted to the redeemer, a MaxFloat update for a given B is broadcast to all O subscribers, whether or not they're working with the B. (Does this also create a local cached value for the max float even if the current orchestrator isn't working with the B?)

BTW it would be great to have this behavior documented somewhere, eg in `/doc`, including the DB schema updates.

net/redeemer.proto
kyriediculous: I think what is meant in the first SO thread is that it can surely break backwards compatibility for the generated stubs but the behaviour between using `empty.Empty` or an empty defined message remains the same so it changes nothing over the wire (both are empty objects in gRPC). 

I'm fine with defining an empty message explicitly though if it aleviates concerns and I can see why that type safety  for generated stubs could be relevant in the future.

net/redeemer.proto
```diff
@@ -0,0 +1,34 @@
...
+    rpc QueueTicket(Ticket) returns (google.protobuf.Empty) {}
+    rpc MaxFloat(MaxFloatRequest) returns (MaxFloatUpdate) {}
+    rpc MonitorMaxFloat(google.protobuf.Empty) returns (stream MaxFloatUpdate) {}
+}
+
+message MaxFloatRequest {
+    bytes sender = 1;
+}
+
+message Ticket {
```
yondonfu, line 11: Since every field in this message besides `recipient_rand` can already be transmitted across the wire in the `Payment` message, could we use the `Payment` message to carry these values and transmit it along with the recipientRand value? Something like:

```
message RedeemablePayment {
    Payment payment = 1;
    repeated bytes recipient_rands = 2;
}
```

Since a `Payment` can contain many winning tickets, the above message allows for multiple recipientRand values to be included as well. In practice, this message will probably only contain a single winning ticket, but structuring the message to allow for multiple winning tickets may be beneficial for future flexibility since we'd like to avoid wire message format changes whenever possible.

server/redeemer.go
```diff
@@ -0,0 +1,398 @@
...
+	"github.com/livepeer/go-livepeer/eth"
+	"github.com/livepeer/go-livepeer/net"
+	"github.com/livepeer/go-livepeer/pm"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/peer"
+	"google.golang.org/grpc/status"
+)
+
+const rpcTimeout = 8 * time.Second
```
yondonfu: We already define a GRPC timeout [here](https://github.com/livepeer/go-livepeer/blob/d9ea3358e0a43f07a0fd78880f37bd48750c1be2/server/rpc.go#L31) - can we reuse that?

server/redeemer.go
```diff
@@ -0,0 +1,398 @@
...
+	close(r.quit)
+}
+
+func (r *redeemerClient) QueueTicket(ticket *pm.SignedTicket) error {
+	ctx, cancel := context.WithTimeout(context.Background(), rpcTimeout)
+	defer cancel()
+	// QueueTicket either returns an error on failure
+	// or an empty object on success so we can ignore the response object.
+	res := make(chan error)
+	go func() {
```
yondonfu: Why can't we just directly call the method? i.e.

```
return r.rpc.QueueTicket(ctx, protoTicket(ticket))
```

server/redeemer.go
```diff
@@ -0,0 +1,398 @@
...
+	}
+
+	// request max float from redeemer if not locally available
+	ctx, cancel := context.WithTimeout(context.Background(), rpcTimeout)
+	defer cancel()
+
+	mfC := make(chan *big.Int)
+	errC := make(chan error)
+	go func() {
+		mf, err := r.rpc.MaxFloat(ctx, &net.MaxFloatRequest{Sender: sender.Bytes()})
```
yondonfu: Why can't we just directly call the method? i.e.

```
maxFloat, err := rp.rpc.MaxFloat(ctx, ...)
```

server/redeemer.go
```diff
@@ -0,0 +1,398 @@
...
+	"google.golang.org/grpc/peer"
+	"google.golang.org/grpc/status"
+)
+
+const rpcTimeout = 8 * time.Second
+
+var cleanupLoopTime = 1 * time.Hour
+
+// Redeemer is the interface for a ticket redemption gRPC service
+type Redeemer interface {
```
yondonfu: Do we need to export this interface? Doesn't look like there are any users of this interface. If this is the case, we can just export the struct.

server/redeemer.go
```diff
@@ -0,0 +1,398 @@
...
+	maxFloat map[ethcommon.Address]*big.Int
+	mu       sync.RWMutex
+	quit     chan struct{}
+	sm       pm.SenderManager
+	tm       pm.TimeManager
+}
+
+// NewRedeemerClient instantiates a new client for the ticket redemption service
+// The client implements the pm.SenderMonitor interface
+func NewRedeemerClient(uri *url.URL, sm pm.SenderManager, tm pm.TimeManager) (pm.SenderMonitor, *grpc.ClientConn, error) {
```
yondonfu: I don't think we need to return the `SenderMonitor` interface here - the fact that `redeemerClient` implements the `SenderMonitor` interface can be implicit.

server/redeemer.go
```diff
@@ -0,0 +1,398 @@
...
+		glog.Errorf("Did not connect to orch=%v err=%v", uri, err)
+		return nil, nil, fmt.Errorf("Did not connect to orch=%v err=%v", uri, err)
+	}
+	return &redeemerClient{
+		rpc:      net.NewTicketRedeemerClient(conn),
+		sm:       sm,
+		tm:       tm,
+		maxFloat: make(map[ethcommon.Address]*big.Int),
+		quit:     make(chan struct{}),
+	}, conn, nil
```
yondonfu: Since we already have `Stop()` method that is used to close the quit channel could we store the ClientConn in the struct and call `Close()` on it instead of having the caller close the ClientConn and call `Stop()`?

server/redeemer.go
yondonfu: We should clean up this map similar to what we do in the current `SenderMonitor` implementation and call `Clear()` on the `pm.SenderManager` instance so that cached sender info can be cleared.

server/redeemer.go
```diff
@@ -0,0 +1,398 @@
...
+	defer listener.Close()
+	if err != nil {
+		return err
+	}
+	// slice of gRPC options
+	// Here we can configure things like TLS
+	opts := []grpc.ServerOption{}
+	// var s *grpc.Server
+	s := grpc.NewServer(opts...)
+	defer s.Stop()
```
yondonfu: By the time this defer statement executes, the server would've stopped already when `s.Serve()` returns. May make more sense to store the server in the struct so that `Stop()` can be called in `redeemer.Stop()` which would allow the server to be stopped when the program exits.

yondonfu (changes requested):

server/redeemer.go
kyriediculous: In theory one could supply a context with cancel to the outer method, when the cancel function is called then instead of the entire call finishing it will exit immediately, at least that's how I understood contexts.

server/redeemer.go
kyriediculous: Yeah sure we can export the struct instead

server/redeemer.go
yondonfu: `rpc.QueueTicket()` should return immediately if its context is cancelled assuming it handles context cancellation correctly. See this [other example](https://github.com/livepeer/go-livepeer/blob/d9ea3358e0a43f07a0fd78880f37bd48750c1be2/server/rpc.go#L185).

server/redeemer.go
```diff
@@ -0,0 +1,398 @@
...
+	if err := r.sm.QueueTicket(t); err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+	glog.Infof("ticket queued sender=0x%x", ticket.Sender)
+
+	go r.monitorMaxFloat(ethcommon.BytesToAddress(ticket.Sender))
+	return &empty.Empty{}, nil
+}
+
+func (r *redeemer) monitorMaxFloat(sender ethcommon.Address) {
```
yondonfu: If this is only called in `QueueTicket()`, then the server will only subscribe to max float updates for a sender via the SenderMonitor when a ticket is queued, but not when a sender's max float changes due to other reserve changes. We should subscribe to max float updates for a sender via the SenderMonitor as soon as a client indicates that it cares about float updates for the sender.

yondonfu (changes requested):

server/redeemer.go
kyriediculous: Ah yes you're right ! 

But changing where this method is called wouldn't change much I think. 
This method subscribes to max float updates from the `SenderMonitor` through a channel to then send it to the RPC clients. We do this upon receiving a ticket if we are not already monitoring this sender. 

The issue is that we only receive updates on the `sink` channel from `addFloat` and `subFloat` but not when we receive ETH events. 

I think the solution would be to send updates into the subscription from `SenderMonitor.SubscribeMaxFloat`. 

Because this state is actually kept and updated on the `SenderManager` I think we would have to create a `SubscribeMaxFloatUpdate` subscription on the `SenderManager` for the `SenderMonitor` to subscribe to.

net/redeemer.proto
kyriediculous: A `Payment` can only be for a single set of `TicketParams` thus only for a single `recipientRand` as well, right?  

I felt like duplicating the message definition makes the code quite a bit cleaner and nicer on the eye, easier type conversions and no loops. 

So the usefulness of reusing the `Payment` message to me seems to be related to whether we expect to receive multiple winning tickets within a single payment often. 

I can see the benefit of not having to send multiple RPC calls when we receive multiple winning tickets in a payment but do we really expect it to occur that often?

server/redeemer.go
kyriediculous: fixed in 463ff651

server/redeemer.go
kyriediculous: fixed in 67f221ee

server/redeemer.go
kyriediculous: fixed in 67f221ee

net/redeemer.proto
kyriediculous: fixed in 7f9ec39f

server/redeemer.go
kyriediculous: fixed in 272bbfe493bfa2df89210b3a8dd193cd9d651bab

server/redeemer.go
kyriediculous: fixed in 25b10e8d19440a55c9489a269e0e3c5e3858b352

server/redeemer.go
yondonfu: Ah I see a subscription created via `SenderMonitor.SubscribeMaxFloat()` only receives max float updates with the max float for a sender changes when a a ticket is redeemed.

Instead of adding a subscription method to `SenderManager` just so `SenderMonitor` can receive max float updates, what do you think about fetching the max float from `SenderMonitor` using the `MaxFloat()` at a regular interval? This may simplify things and would eliminate the need for the `SenderMonitor.SubscribeMaxFloat()` method as well.

A rough sketch of what this might look like:

```golang
// This is pretty arbitrary, maybe there is a better value
var maxFloatPollInterval = 5 * time.Second

...
ticker := time.NewTicker(maxFloatPollInterval)
for {
    select {
    case <-quit:
        return
    case <-ticker.C:
        maxFloat, err := r.sm.MaxFloat(sender)
        if err != nil {
            // Handle error
        }
        // Check if maxFloat is different from last cached maxFloat
        // If it is different, update the cached maxFloat and send the updated value
        // to listeners
    }
}
```

The downside of this approach is that the server needs to cache maxFloat updates to avoid sending updates to clients even when the maxFloat has not changed. The upside of this approach is that we can avoid managing subscription code in both `SenderMonitor` and `SenderWatcher`.

kyriediculous: I think from a workflow POV _push_ (eg subscription) would be more preferable because we'd like to update the clients ASAP instead of having to wait for the next polling interval. 

From a runtime perspective I think it's also more optimal especially in the case where `blockPollingInterval` > `maxFloatPollInterval`. 

So given that and the fact that we'd have to manage a watchdog for either solution what do you think of sticking with subscriptions? It might be a useful method down the line even while the polling interval would be isolated logic.

net/redeemer.proto
yondonfu: > A Payment can only be for a single set of TicketParams thus only for a single recipientRand as well, right?

Ah yeah you're right. Only a single recipientRand would be needed for a payment that could contain multiple tickets.

> I felt like duplicating the message definition makes the code quite a bit cleaner and nicer on the eye, easier type conversions and no loops.

Yeah good point.

It might make sense to structure the Ticket message in a similar manner as the Payment message though i.e.

```golang
message Ticket {
    TicketParams ticket_params = 1;
    bytes sender = 2;
    TicketExpirationParams expiration_params = 3;
    TicketSenderParams sender_params = 4;
    bytes recipient_rand = 5;
}
```

This way if the ticket format ever happens to change both the Payment and Ticket messages can pick up those changes since they share the same structs i.e. TicketParams.

server/redeemer.go
yondonfu: Yeah subscriptions would be more ideal - polling would just accept introducing a chance that updates are not sent exactly when they happen (although this  chance could be minimized by tuning the polling interval) in exchange for reduced code update requirements.

> we'd have to manage a watchdog for either solution

Right just so I make sure I understand your point here - when polling the Redeemer would need to know when to clear the cached max float and when subscribing the Redeemer would need to know when close the subscription. While this particular point doesn't highlight an advantage of either solution, I do acknowledge that the polling solution has additional state requirements which is a disadvantage of the approach.

> It might be a useful method down the line even while the polling interval would be isolated logic.

Yeah fair enough. If you think that code updates required for adding subscribing to SenderInfo updates on the SenderManager from the SenderMonitor aren't particularly large then I'm cool with taking this approach.

server/redeemer.go
kyriediculous: added a cleanup loop in c5609320614c7c842151f2a41b395e1725c76aa5

server/redeemer.go
j0sh: I'm mostly just following along on this series of PRs, and don't really have any concrete suggestions right now, but just wanted to point out an area that is still confusing to me from the outside, from this conversation and some others around float monitoring, such as https://github.com/livepeer/go-livepeer/pull/1495#discussion_r432845661 and https://github.com/livepeer/go-livepeer/pull/1513#discussion_r432720399 

This is my current understanding:

(Using the terms 'broadcaster' rather than 'sender' just because it's easier for me to contextualize mentally who is doing what)

* RPC float monitoring does *not* depend on a per-broadcaster subscription. The RPC redeemer blasts out float updates to all orchestrators upon on ticket redemption, regardless of whether the orchestrator is working with the broadcaster. (MonitorMaxFloat)

* Orchestrators still need to create a per-broadcaster subscription for blockchain event monitoring and other bookkeeping. (SubscribeMaxFloat) This is separate from MonitorMaxFloat.

* If a RPC redeemer is running, it does *not* monitor for blockchain events, only winning tickets from other orchestrators. (Or does it?)

My sense has been that these could have been a single call that's abstracted out into a RPC monitor or a local monitor, depending on the mode. Since I think we're planning to build a blockchain monitoring service at some point anyway, this seemed like a sensible first step?

Anyway, this isn't really a request to change anything - just wanted to express (again) a bit of where my confusion was coming from. It would be great to codify the current design under `/doc` so later folks have an easier time in following along, especially because the terms / concepts used in the implementation are a bit on the abstract side.

kyriediculous: Max float is a composite value of state kept on the `senderMonitor` as well as the `senderWatcher`

`senderWatcher.SubscribeMaxFloatChange` (tentative naming)  would be to tell the `senderMonitor` like "hey the max float for this sender changed , just letting you know in case you wanted to notify your own subscribers" (which would be the `Redeemer`) 

`senderMonitor.SubscribeMaxFloat` would allow consumers of the `senderMonitor` to know when `maxfloat` for a broadcaster changes due to state changes in either `senderWatcher` or `senderMonitor`

kyriediculous: Actually I spent some more time thinking about this and I don't believe we don't need senderWatcher subscriptions at all. 

So which events would impact a max float change ? 
- Winning ticket redeemed
- Fund Reserve
- Withdraw 
- New Round 

In case of the ticket redemption we already would send a max float update due to the deferred `addFloat` call 

In case of a withdrawal we would have already stopped working with the broadcaster before the deposit and reserve unlocked 

Fund reserve , there seems to be no negative impact here by not immediately sending updates when the reserve gets funded unless the segment length would be adequately long to be able to refund a reserve that would have been insufficient by the time the next segment rolls around. 

New round , would have to update max float values for all senders right now if the pool size changes. I believe this to be a an edge case and impact would be minimal unless there's a huge shift in pool size.

j0sh: > In case of a withdrawal we would have already stopped working with the broadcaster before the deposit and reserve unlocked

Can a winning ticket still draw down locked funds? Eg, a B starts working with an O, puts everything under a withdrawal lock, then suddenly generates a winning ticket. Or is that a non-issue? If so, it seems okay.

Am guessing that the next call to MaxFloat will exclude locked funds from the balance? So the O can decline to work with the B at that point.

yondonfu: > Fund reserve , there seems to be no negative impact here by not immediately sending updates when the reserve gets funded unless the segment length would be adequately long to be able to refund a reserve that would have been insufficient by the time the next segment rolls around.

The negative impact is that the last max float that the redeemer sends to O will be inaccurate until a winning ticket from B is redeemed which can affect the ticket face value and price that O advertises to B.

> I believe this to be a an edge case and impact would be minimal unless there's a huge shift in pool size.

While the difference between the perceived max float and the actual max float might not be large when there is a small change in the on-chain pool size, I think we should account for these changes for the reasons mentioned above.

> senderWatcher.SubscribeMaxFloatChange (tentative naming)

`SubscribeReserveUpdate()` might be a more accurate name since the `SenderWatcher` is not aware of the concept of a max float - all it knows is that the on-chain reserve has changed.

net/redeemer.proto
kyriediculous: Fixed in the following commits 

protofile , stubs & mocks: 258e02747160df22b465787adfd245ce9ed02a65
server: 28432f0acd05767f4310277263ef73bf63781e9f
client: 759114fb02f384681b8003e48efb2bb9801bd842

net/redeemer.proto
kyriediculous: Will definitely document the behaviour as I notice it is a source of confusion. 

Fixed the initial comment from this thread in 258e02747160df22b465787adfd245ce9ed02a65

> On startup, O subscribes to MonitorMaxFloat on a RPC redeemer for float updates

Correct. 

> During discovery / GetOrchestratorInfo , an O invokes MaxFloat RPC and caches the resulting value for the given B. Only that one O gets the MaxFloat value. (How long is the value kept in cache?)

We don't keep it in cache until we actually start working with the B

> Anytime a winning ticket is submitted to the redeemer, a MaxFloat update for a given B is broadcast to all O subscribers, whether or not they're working with the B. (Does this also create a local cached value for the max float even if the current orchestrator isn't working with the B?)

yes all subscribers currently, if this appears to be a bottleneck I believe we can address it later on. It will create local cache on the orchestrators that will be cleaned up if it's not used.

net/redeemer.proto
kyriediculous: fixed in 258e02747160df22b465787adfd245ce9ed02a65

server/redeemer.go
kyriediculous: fixed in 717ef7cb128335ca6b690a98a4a6f908f3ad4cce

server/redeemer.go
kyriediculous: fixed in 44044b6da3ffbf0afde1a98be1354eea884017e3

server/redeemer.go
kyriediculous: This has been fixed in #1495

kyriediculous:
2bc7f569 renames the helper `monitorMaxFloat` to `watchMaxFloatUpdate` to avoid confusion with the `MonitorMaxFloat` RPC method.

kyriediculous:
I just forced pushed this branch with quite a few updates , I'll list the major changes. 

I use `sender` instead of `broadcaster` because that is just the terminology used within the `pm` protocol, in the case of Livepeer in its current state these are currently interchangeable. 

**Redeemer Client**

- No longer opens up a single stream to the server on node startup on which to receive `MaxFloatUpdate` for ALL senders/broadcasters. 

- `RedeemerClient.MaxFloat()` will instead open up a stream on a PER-SENDER basis when this method is called and no local cache is available (which also means we don't have an open stream for this sender yet). A listener will listen for sender-specific `maxfloat` updates and update the local cache on the client accordingly. 

_The above two points imply that orchestrators no longer receive updates from all senders, just the ones they care about as-they-go_ 

- Reduced `cleanupLoopTime` from 1 hour to 5 minutes because we now open up a stream (and subscription on the server) for each sender , even those that just require `MaxFloat` during discovery. We want to clean up idle streams (and subscriptions on the server) as soon as is reasonable. 

**Redeemer Server**

- The `LocalSenderMonitor.SubscribeMaxFloatChange` allows the server to receive notifications from the `LocalSenderMonitor` when it updates its state for that sender. Because the streams to clients are now on a per-sender basis, these local subscriptions are now _per-client AND per-sender_ 

- Because streams are now on a per-sender basis we can close these subscriptions as soon as the client closes the stream  (e.g. when it cleans up locally) to reduce the number of active subscriptions opened due to e.g. discovery


**SenderMonitor**
- Added a check in the `LocalSenderMonitor` cleanup loop that prevents `remoteSenders` from being cleaned up when they have active subscribers. Otherwise a stream to a client might still be open but the server would no longer be receiving state updates to send over the stream. 

ticketflow (1)

j0sh:
> The above two points imply that orchestrators no longer receive updates from all senders, just the ones they care about as-they-go

Just curious - what motivated this change? While we discussed doing per-sender subscription earlier, it seemed like we were settled on the old method of a single subscription per O. Wondering what tipped the scales the other way.

kyriediculous:
> what tipped the scales the other way.

There's two things that led to this: 

1. We needed a way for clients to tell the server when they stop working with a sender instead of the server running a cleanup loop to keep track of active clients and senders. Removing the cleanup loop from the `Redeemer` (server) prevents it from cleaning up state prematurely that clients would still be using. 

The first idea was to add a START/STOP enum to the `MaxFloatReq` message that would allow the client to open a stream on node startup and then subscribe/unsubscribe for updates over a bi-di stream by defining the sender and START/STOP. But this quickly ran into issues with concurrent sends over the same stream, which isn't really possible. 

So using per-sender streams instead of a single general funnel allowed us to avoid concurrency issues as well as reduce complexity by a fair amount. A client can now simply tell the server it is unsubscribing for a sender by closing the streams.

2. from a discussion with Yondon it seemed like having the `Redeemer` process as a whole subscribe to `LocalSenderMonitor.SubscribeMaxFloatChange(sender)` for each sender defeated the purpose of using subscriptions in the first place. 

So now we have each separate server side stream thread (which is now per sender due to the above) subscribe to `LocalSenderMonitor.SubscribeMAxFloatChange()` separately. 

While this adds some minor overhead (more channels to loop through when sending a notification) it allowed the `Redeemer` itself to be completely stateless 


tl;dr 

There is a nice synergy between opening and closing a stream between `RedeemerClient` and `Redeemer`  and subscribing/unsubscribing for Max float changes between the `Redeemer`'s active streams and the `LocalSenderMontitor`

kyriediculous:
Accidentally removed the Start method for the client, fixed in e3a0b8af

net/redeemer.proto
```diff
@@ -0,0 +1,27 @@
...
+}
+
+message QueueTicketRes {}
+
+message MaxFloatReq {
+    bytes sender = 1;
+}
+
+message MaxFloatUpdate {
+    bytes max_float= 2;
```
yondonfu: The field number could be 1 here. Also, looks like there is a missing space before the equals sign.

server/redeemer.go
```diff
@@ -0,0 +1,383 @@
...
+		return err
+	}
+	defer listener.Close()
+	if err != nil {
+		return err
+	}
+	// slice of gRPC options
+	// Here we can configure things like TLS
+	opts := []grpc.ServerOption{}
+	// var s *grpc.Server
```
yondonfu: Can remove this.

server/redeemer.go
```diff
@@ -0,0 +1,383 @@
...
+// QueueTicket adds a ticket to the ticket queue
+func (r *Redeemer) QueueTicket(ctx context.Context, ticket *net.Ticket) (*net.QueueTicketRes, error) {
+	t := pmTicket(ticket)
+	if r.recipient != t.Recipient {
+		return nil, status.Error(codes.PermissionDenied, fmt.Sprintf("invalid ticket recipient 0x%x, expected %v", ticket.TicketParams.Recipient, r.recipient.Hex()))
+	}
+	if err := r.sm.QueueTicket(t); err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+	glog.Infof("ticket queued sender=0x%x", ticket.Sender)
```
yondonfu: Let's capitalize "Ticket" for consistency since a lot of our info logs capitalize the first word in the message.

server/redeemer.go
```diff
@@ -0,0 +1,383 @@
...
+	sink := make(chan struct{}, 10)
+	sub := r.sm.SubscribeMaxFloatChange(sender, sink)
+	defer sub.Unsubscribe()
+	for {
+		select {
+		case <-r.quit:
+			return nil
+		case err := <-sub.Err():
+			if err == nil {
+				return status.Error(codes.Canceled, "subscription closed")
```
yondonfu, line 123: Doesn't seem like we need to return an error here if we receive nil on the `sub.Err()` channel.

server/redeemer.go
```diff
@@ -0,0 +1,383 @@
...
+			}
+			return status.Error(codes.Canceled, err.Error())
+		case <-sink:
+			maxFloat, err := r.sm.MaxFloat(sender)
+			if err != nil {
+				glog.Error(err)
```
yondonfu, line 129: If we encounter an error when calling `r.sm.MaxFloat()` we should skip the stream send.

server/redeemer.go
```diff
@@ -0,0 +1,383 @@
...
+		rpc:     net.NewTicketRedeemerClient(conn),
+		sm:      sm,
+		tm:      tm,
+		senders: make(map[ethcommon.Address]*remoteSender),
+		quit:    make(chan struct{}),
+	}, nil
+}
+
+func (r *RedeemerClient) Start() {
+	r.startCleanupLoop()
```
yondonfu: Needs to be in a goroutine or else this method will block.

server/redeemer.go
```diff
@@ -0,0 +1,383 @@
...
+	conn, err := grpc.Dial(
+		uri.String(),
+		grpc.WithBlock(),
+		grpc.WithTimeout(GRPCConnectTimeout),
+		grpc.WithInsecure(),
+	)
+
+	//TODO: PROVIDE KEEPALIVE SETTINGS
+	if err != nil {
+		glog.Errorf("Did not connect to orch=%v err=%v", uri, err)
```
yondonfu: "orch" -> "redeemer"?

server/redeemer.go
```diff
@@ -0,0 +1,383 @@
...
+}
+
+// NewRedeemerClient instantiates a new client for the ticket redemption service
+// The client implements the pm.SenderMonitor interface
+func NewRedeemerClient(uri *url.URL, sm pm.SenderManager, tm pm.TimeManager) (*RedeemerClient, error) {
+	conn, err := grpc.Dial(
+		uri.String(),
+		grpc.WithBlock(),
+		grpc.WithTimeout(GRPCConnectTimeout),
+		grpc.WithInsecure(),
```
yondonfu: Could we use `grpc.WithTransportCredentials(credential.NewTLS(tlsConfig))` like we do [here](https://github.com/livepeer/go-livepeer/blob/d0ea014d68488c1bc26c4e5882949a0d1dc1c86c/server/rpc.go#L225) so its easier to swap in a valid TLS config with `InsecureSkipVerify: false` in the future?

server/redeemer.go
```diff
@@ -0,0 +1,383 @@
...
+	defer cancel()
+	mfu, err := r.rpc.MaxFloat(ctx, &net.MaxFloatReq{Sender: sender.Bytes()})
+	if err != nil {
+		return nil, err
+	}
+	mf := new(big.Int).SetBytes(mfu.MaxFloat)
+
+	// Request updates for sender from Redeemer
+	stream, err := r.rpc.MonitorMaxFloat(ctx, &net.MaxFloatReq{Sender: sender.Bytes()})
+	// By sending a message over the existing stream to the Redeemer
```
yondonfu: Not sure that this comment is accurate - the client isn't sending a message over the stream to the Redeemer here?

server/redeemer.go
```diff
@@ -0,0 +1,383 @@
...
+
+	r.senders[sender] = &remoteSender{
+		maxFloat:   mf,
+		stream:     stream,
+		done:       make(chan struct{}),
+		lastAccess: time.Now(),
+	}
+
+	// release the lock before starting a watchdog on the stream for the 'sender'
+	r.mu.Unlock()
```
yondonfu, line 247: We probably don't want to hold the lock while RPC calls are pending and we don't have to either since `r.senders` is access before and after the RPC calls. We could lock/unlock when checking if a local cache exists for the sender and then lock/unlock when updating the cache.

server/redeemer.go
```diff
@@ -0,0 +1,383 @@
...
+
+	for {
+		select {
+		case <-r.quit:
+			if err := s.stream.CloseSend(); err != nil {
+				glog.Error(err)
+			}
+			return
+		case <-s.done:
+			glog.Infof("closing stream for sender=%v", sender.Hex())
```
yondonfu: Let's capitalize "Closing" for consistency since a lot of our info logs capitalize the first word in the message.

server/redeemer.go
```diff
@@ -0,0 +1,383 @@
...
+				return status.Error(codes.Canceled, "subscription closed")
+			}
+			return status.Error(codes.Canceled, err.Error())
+		case <-sink:
+			maxFloat, err := r.sm.MaxFloat(sender)
+			if err != nil {
+				glog.Error(err)
+			}
+			if err := stream.Send(&net.MaxFloatUpdate{MaxFloat: maxFloat.Bytes()}); err != nil {
+				glog.Error(err)
```
yondonfu: According to the [docs](https://godoc.org/google.golang.org/grpc#ServerStream), any non-EOF error will abort the stream so I'm not sure if it is safe to continue using this stream in the loop in that scenario. May be safer to just return here.

server/redeemer.go
```diff
@@ -0,0 +1,383 @@
...
+			} else {
+				updateC <- update
+			}
+		}
+	}()
+
+	for {
+		select {
+		case <-r.quit:
+			if err := s.stream.CloseSend(); err != nil {
```
yondonfu: It appears that it is possible to leak a goroutine/context even after the client calls `CloseSend()` on the stream unless one of the actions mentioned [here](https://godoc.org/google.golang.org/grpc#ClientConn.NewStream) are performed. Cancelling the context that was provided to `rpc.MonitorMaxFloat()` may be the best bet here (I'm assuming that the context provided to that RPC method is used to create the stream).

yondonfu (changes requested):

server/redeemer.go
kyriediculous: I believe nil on the channel means the channel and thus subscription is closed for an unexpected reason. In this case we should close the stream and inform the client to clear it's local cache so it can start a new monitoring sequence.

server/redeemer.go
kyriediculous: Should we skip , i.e. `continue` or `return` are there cases where this is recoverable? 

I think remote ETH node unavailable if no local cache is present, but that's about it.

server/redeemer.go
kyriediculous: I assume we no longer need `grpc.WithInsecure()` then because we now have valid transport credentials that accomplish the same?

server/redeemer.go
kyriediculous: According to the docs for `RecvMsg()`

> It returns io.EOF when the client has performed a CloseSend.

So don't we already do 

> 3. Call RecvMsg until a non-nil error is returned.

which we then handle as 

```
update, err := s.stream.Recv()
  if err != nil {
    errC <- err
    if err == io.EOF {
    return
    }
}
```

Or do we need to explicitly pass the context into the `r.monitorMaxFloat(sender)` watcher an cancel it when we encounter an error? Isn't that context only used for the request to the server?

net/redeemer.proto
kyriediculous: fixed in https://github.com/livepeer/go-livepeer/pull/1513/commits/36c56e9889a7ea1b378e649b3253217a185cae62

server/redeemer.go
kyriediculous: fixed in https://github.com/livepeer/go-livepeer/pull/1513/commits/a8e3d49938dcf3a62e11bff72accdc054c2d205d

server/redeemer.go
kyriediculous: fixed in https://github.com/livepeer/go-livepeer/pull/1513/commits/a8e3d49938dcf3a62e11bff72accdc054c2d205d

server/redeemer.go
kyriediculous: fixed in https://github.com/livepeer/go-livepeer/pull/1513/commits/a8e3d49938dcf3a62e11bff72accdc054c2d205d

server/redeemer.go
kyriediculous: fixed in https://github.com/livepeer/go-livepeer/pull/1513/commits/a8e3d49938dcf3a62e11bff72accdc054c2d205d

server/redeemer.go
kyriediculous: fixed in https://github.com/livepeer/go-livepeer/pull/1513/commits/a8e3d49938dcf3a62e11bff72accdc054c2d205d

kyriediculous: fixed in https://github.com/livepeer/go-livepeer/pull/1513/commits/6b7ac892cd933ab613b66a1cd5ae30cd990c128e

server/redeemer.go
kyriediculous: fixed in https://github.com/livepeer/go-livepeer/pull/1513/commits/6b7ac892cd933ab613b66a1cd5ae30cd990c128e

server/redeemer.go
kyriediculous: fixed in https://github.com/livepeer/go-livepeer/pull/1513/commits/6b7ac892cd933ab613b66a1cd5ae30cd990c128e

server/redeemer.go
kyriediculous: fixed in https://github.com/livepeer/go-livepeer/pull/1513/commits/6b7ac892cd933ab613b66a1cd5ae30cd990c128e

server/redeemer.go
kyriediculous: fixed in https://github.com/livepeer/go-livepeer/pull/1513/commits/6b7ac892cd933ab613b66a1cd5ae30cd990c128e

Makefile
kyriediculous: I added the redeemer and its mocks separately to the makefile, does it check out ?

server/redeemer.go
yondonfu: Yeah the transport credentials should accomplish the same thing and we should be able to pass in valid credentials with `InsecureSkipVerify: false` in the future by updating the `tlsConfig` passed in.

server/redeemer.go
yondonfu: According to the [docs](https://godoc.org/github.com/ethereum/go-ethereum#Subscription) for `Subscription`, when `Unsubscribe()` is called the error channel will be closed. At that point, the last value sent on the channel due to the close operation will be nil (which in this case is the zero value of the channel data type). 

Technically, we shouldn't ever receive nil on the error channel since the `Unsubscribe()` call is deferred until after the for loop returns. But, *if* there is a nil error returned on the error channel I don't think returning an error is correct since no error actually occurred.

yondonfu: Ah I see I thought that the error channel only received a nil value when it was closed due to `Unsubscribe()` being called, but I see in the docs that the error channel could also be closed due to an issue with the subscription. In that case, returning an error here makes sense 👍

server/redeemer.go
yondonfu: `continue` sounds good to me.

server/redeemer.go
```diff
@@ -0,0 +1,378 @@
...
+	if err != nil {
+		return err
+	}
+
+	s := grpc.NewServer()
+	r.server = s
+
+	net.RegisterTicketRedeemerServer(s, r)
+
+	return s.Serve(listener)
```
yondonfu, line 75: I think `s.Serve()` needs to be run in a goroutine or else this method will block until `s.Serve()` returns?

server/redeemer.go
```diff
@@ -0,0 +1,378 @@
...
+}
+
+// Start starts a Redeemer server
+func (r *Redeemer) Start(host *url.URL) error {
+	listener, err := gonet.Listen("tcp", host.String())
+	if err != nil {
+		return err
+	}
+	defer listener.Close()
+	if err != nil {
```
yondonfu: This check can be removed.

server/redeemer.go
```diff
@@ -0,0 +1,378 @@
...
+	ctx, cancel := context.WithTimeout(context.Background(), GRPCTimeout)
+	defer cancel()
+	mfu, err := r.rpc.MaxFloat(ctx, &net.MaxFloatReq{Sender: sender.Bytes()})
+	if err != nil {
+		return nil, err
+	}
+	mf := new(big.Int).SetBytes(mfu.MaxFloat)
+
+	// Request updates for sender from Redeemer
+	stream, err := r.rpc.MonitorMaxFloat(ctx, &net.MaxFloatReq{Sender: sender.Bytes()})
```
yondonfu, line 229: I think we want to use a separate context here. If we re-use the `ctx` that was passed to `MaxFloat()` then the timeout allowed for this call will end up being shorter than `GRPCTimeout`.

server/redeemer.go
```diff
@@ -0,0 +1,378 @@
...
+			if err := s.stream.CloseSend(); err != nil {
+				glog.Error(err)
+			}
+			return
+		case update := <-updateC:
+			r.mu.Lock()
+			r.senders[sender].maxFloat = new(big.Int).SetBytes(update.MaxFloat)
+			r.senders[sender].lastAccess = time.Now()
+			r.mu.Unlock()
+		case <-errC:
```
yondonfu: Log error here since we're not doing so in the goroutine with `stream.Recv()`?

server/redeemer.go
```diff
@@ -0,0 +1,378 @@
...
+		return
+	}
+
+	updateC := make(chan *net.MaxFloatUpdate)
+	errC := make(chan error)
+	go func() {
+		for {
+			update, err := s.stream.Recv()
+			if err != nil {
+				errC <- err
```
yondonfu, line 283: Is it safe to continue calling `Recv()` after a non-nil, but non-EOF error is returned?

The docs for [ClientStream.RecvMsg()](https://godoc.org/google.golang.org/grpc#ClientStream) (which is called by `stream.Recv()` under the hood) note that for a non-EOF error the stream is aborted. Seems that we should be exiting this loop as soon as we receive a non-nil error.

server/redeemer.go
yondonfu: Ah I missed that. Yeah you're right - calling `stream.Recv()` (which calls `RecvMsg()` under the hood) until a non-nil error should ensure that the stream is cleaned up properly.

> Or do we need to explicitly pass the context into the r.monitorMaxFloat(sender) watcher an cancel it when we encounter an error? Isn't that context only used for the request to the server?

The context passed into `rpc.MonitorMaxFloat()` is used as the context for the client stream. So, another way to clean up the stream would be to cancel that context.

All that being said...

Is the client actually supposed to call `CloseSend()`? The docs note that `CloseSend()` closes the send direction of the stream and the client already calls `CloseSend()` in the auto-generated [MonitorMaxFloat()](https://github.com/livepeer/go-livepeer/pull/1513/files#diff-39b2e318cf89dc828db18678277bcf53R293) method.

server/redeemer_test.go
```diff
@@ -0,0 +1,671 @@
...
+
+	errC := make(chan error)
+	// test stream send
+	stream.EXPECT().Send(&net.MaxFloatUpdate{MaxFloat: maxFloat.Bytes()}).Return(nil).Times(1)
+	go func() {
+		errC <- r.MonitorMaxFloat(req, stream)
+	}()
+	time.Sleep(time.Second)
+	sm.sink <- struct{}{}
+	time.Sleep(time.Second)
```
yondonfu: Missing an assertion that an error was not received?

Ex.

```
timer, cancel := context.WithTimeout(context.Background(), 1 * time.Second)
select {
    case <-errC:
        t.Fail()
    case <-timer.Done():
}
cancel()
```

server/redeemer_test.go
```diff
@@ -0,0 +1,671 @@
...
+
+	stream.EXPECT().Recv().Return(&net.MaxFloatUpdate{MaxFloat: expFloat.Bytes()}, nil).AnyTimes()
+	go rc.monitorMaxFloat(sender)
+
+	time.Sleep(20 * time.Millisecond)
+	rc.mu.Lock()
+	stream.EXPECT().CloseSend().Return(nil)
+	close(rc.senders[sender].done)
+	rc.mu.Unlock()
+	time.Sleep(time.Millisecond)
```
yondonfu: Let's make this at least 20 to avoid possibility of flakiness if CI is slow

yondonfu (changes requested):

server/redeemer.go
kyriediculous: That's right it will block. Currently however the caller of `Start` wraps this in a goroutine. Want to change this to have the callee start the goroutine? 

https://github.com/livepeer/go-livepeer/blob/c63b8c0afc3afca9ec1f2f49a836d6f9ec335eb4/cmd/livepeer/livepeer.go#L619

server/redeemer.go
kyriediculous: I believe the `CloseSend` in the generated stubs is for the _request_ side, i.e. opening the stream. This is the _send_ side of the client to the server

I think this is just how gRPC implements streams, they are implemented as bidi streams but if we want only server or client side we close that side of the stream after the first message. 

Anyway I think you might be right though about the `CloseSend` that we explicitly call from the client. 

But I'm looking for an elegant way to stop the server-side stream from the client, we can close the stream client side but the server won't know the stream is severed until it does another send.

server/redeemer.go
kyriediculous: Ah I see , because the deadline is actually based on `time.Now().Add(timeout)`. Nice to know ! :+1:

yondonfu:
I think the client should be able to terminate the stream by cancelling the
context that was used to create the stream (i.e the one passed into the
MonitorMaxFloat RPC call).

kyriediculous:
> I think the client should be able to terminate the stream by cancelling the
context that was used to create the stream (i.e the one passed into the
MonitorMaxFloat RPC call).

Yep that's right, from what I read in the docs the issue however is that the server won't know that the stream is closed from the client side until it receives an error  when trying a `SendMsg()` after the client has closed the stream.

This means that what is considered "idle" cache on the client, will not be considered idle on the server until after the next `SendMsg`.  So idle cache on the server-side can stick around much longer. After all clients would close the stream for a `sender` that cache and the streams will remain active on the server until the `LocalSenderMonitor` closes the `subscriptionScope` for that `sender` (up to 1 hour after last access).  

Or will cancelling the context on the client result in receiving a message on the `	stream.Context().Done()` channel server side? 

I think so?  from https://grpc.io/docs/what-is-grpc/core-concepts/#cancelling-an-rpc and  https://grpc.io/docs/what-is-grpc/core-concepts/#deadlines

> Either the client or the server can cancel an RPC at any time. A cancellation terminates the RPC immediately so that no further work is done.

> On the server side, the server can query to see if a particular RPC has timed out, or how much time is left to complete the RPC.

server/redeemer.go
yondonfu: Got it. I'm cool with having the caller of `Start()` wrap it in a goroutine. I think it would be helpful to add a comment on this method noting that it will block.

yondonfu:
> Or will cancelling the context on the client result in receiving a message on the stream.Context().Done() channel server side?

I think this should be the case, but could be worth creating a test to confirm.

kyriediculous:
I believe everything is addressed in 7ef5fff

server/redeemer.go
```diff
@@ -0,0 +1,374 @@
...
+
+	sink := make(chan struct{}, 10)
+	sub := r.sm.SubscribeMaxFloatChange(sender, sink)
+	defer sub.Unsubscribe()
+	for {
+		select {
+		case <-r.quit:
+			return nil
+		case <-stream.Context().Done():
+			return status.Error(codes.Canceled, "context cancelled")
```
yondonfu: I don't think context cancellation should trigger an error since the context being cancelled should be pretty normal.

yondonfu (changes requested):

kyriediculous:
fixed in bb4e7ac0

yondonfu (commented):
Changes look good to me. 👍  for a rebase from me as long as comments from other reviewers have been addressed.

We should be able to drop adee003 during the rebase too.

kyriediculous:
Fixed a couple of bugs that popped up during end to end testing 

- `RedeemerClient.MaxFloat` properly releasing lock after returning value from cache 
- `RedeemerClient.MaxFloat` no longer using a context with timeout to start the stream, this will cancel the stream context once the timeout expires regardless of whether the stream is active or not. 
- using `string` instead of `url.URL` to specify the addresses for client and server. 
- `RedeemerClient.startCleanupLoop` closing `done` channel instead of simply sending a value over it when a sender gets cleared.

server/redeemer.go
```diff
@@ -160,10 +158,11 @@ type remoteSender struct {
 
 // NewRedeemerClient instantiates a new client for the ticket redemption service
 // The client implements the pm.SenderMonitor interface
-func NewRedeemerClient(uri *url.URL, sm pm.SenderManager, tm pm.TimeManager) (*RedeemerClient, error) {
+func NewRedeemerClient(uri string, sm pm.SenderManager, tm pm.TimeManager) (*RedeemerClient, error) {
 	conn, err := grpc.Dial(
-		uri.String(),
+		uri,
 		grpc.WithTransportCredentials(credentials.NewTLS(tlsConfig)),
+		grpc.WithInsecure(),
```
yondonfu: Do we need this if the `tlsConfig` in the previous line already specifies `InsecureSkipVerify: true`?

kyriediculous: The client wouldn't connect to the server without it. 

Might also have been due to 3 sec timeout at that point (hadn't reached that one yet) , I'll double check.

You're right that we _shouldn't_ need this.

kyriediculous: fixed in 82eb1280

kyriediculous:
rebased

yondonfu (approved):
LGTM 🚢  pending green CI (which I think should be fixed by rebasing with master to pick up the latest fix for `TestSubmitSegment_HttpPostError`). Let's drop ce65b1e during the rebase too.

Issue #781, opened 2019-03-07
Endpoint to List All Streams


ericxtang:
@f1l1b0x had suggested to create an endpoint to list all active streams.  This is something offered by most transcoding products out there, and we expect customers to demand this feature.  It gives users the ability to manage on-going streams.  

Would love to get feedback from @j0sh and @darkdarkdragon

j0sh:
We already have this in the CLI `/status` endpoint. We really need to document these endpoints because there is a lot of useful information hidden in there. See https://github.com/livepeer/go-livepeer/issues/553

darkdarkdragon:
Yes, I use it all the time. But maybe @f1l1b0x was taking about exposing endpoint to users of our Transcoding API?

j0sh:
Yeah, why not. Although perhaps we should reverse proxy this specifically for the demo network instead of exposing the entirety of the CLI directly or re-implementing this on a separate port.

There is probably a useful conversation to be had on the "surface area" of the node and how much should be exposed. Also related is my earlier comments on exposing the metrics endpoint, https://github.com/livepeer/go-livepeer/issues/760#issuecomment-468833079

ericxtang:
In general, I feel like the "CLI-specific" endpoints should be grouped differently from "user-facing" endpoints.  What do you guys think about having "API grouping", and being able to assign different groups to different ports?  So far we'd have `rtmp`, `stream`(hls, and maybe later dash), `crud`, and `cli`.  The command line would look something like `-rtmpPort`, `-streamPort`, `-crudPort`, `-cliPort`.

Metrics is a little different, in that the data is not logged on the `go-livepeer` node.

darkdarkdragon:
Sounds like a good idea. Probably we should call it something more than `crud`, because we're essentially implementing our transcoding API into node, and probably it will be something more than just `crud`. Maybe just call it `api`?

j0sh:
> I feel like the "CLI-specific" endpoints should be grouped differently from "user-facing" endpoints

It really depends on what what we consider "CLI specific" and what's "user facing". The CLI exposes useful diagnostic and administrative features; surely that can be considered part of the API? Even if not, there are various degrees of being "user facing" ; public or administrative/internal? There are infinite variations to this and we're already starting to see splintering (eg, metrics).

We're now talking about opening five ports, four of which are HTTP-based:

* CLI
* Stream / gRPC
* Metrics
* "CRUD"
* RTMP

[...]

PR #1521, opened 2020-05-27
cmd: redeemer and redeemerAddr flags

### LINKED ISSUE #1516: Expose user flags to start in multi-O mode using a Ticket Redemption Service
- Provide a flag to start an orchestrator that is funded with ETH to also run a Ticket Redemption Service
- Provide a flag to start an unfunded orchestrator as a Ticket Redemption Client 
- Start the necessary services

kyriediculous:
Introduces flags to start an orchestrator with a ticket redemption service by providing the `-redeemer=true` flag. Other ochestrators in a multi-O cluster can connect to it using the `-redeemerAddr` flag. 

If either flag is omitted the node will start normally 

I didn't choose to support operating in standalone mode _yet_ . I feel like that can be done once we rework the main.go file a little bit and extract e.g. the blockwatcher as well. 

Ran a local setup using the devtool 

```
I0527 03:11:52.402810   60759 livepeer.go:534] redeemer started
I0527 03:11:52.403949   60759 redeemer.go:164] new MonitorMaxFloat subscriber: 127.0.0.1:50398
```

Fixes #1516 

**Checklist:**

### DIFF
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -127,6 +127,9 @@ func main() {
 	pixelsPerUnit := flag.Int("pixelsPerUnit", 1, "Amount of pixels per unit. Set to '> 1' to have smaller price granularity than 1 wei / pixel")
 	// Interval to poll for blocks
 	blockPollingInterval := flag.Int("blockPollingInterval", 5, "Interval in seconds at which different blockchain event services poll for blocks")
+	// Redemption service
+	redeemer := flag.Bool("redeemer", false, "Set to true to run a ticket redemption service")
+	redeemerAddr := flag.String("redeemerAddr", "", "URL of the ticket redemption service to use")
 	// Metrics & logging:
 	monitor := flag.Bool("monitor", false, "Set to true to send performance metrics")
 	version := flag.Bool("version", false, "Print out the version")
@@ -265,6 +268,8 @@ func main() {
 		n.NodeType = core.TranscoderNode
 	} else if *broadcaster {
 		n.NodeType = core.BroadcasterNode
+	} else if *redeemer {
+		n.NodeType = core.RedeemerNode
 	} else {
 		glog.Fatalf("Node type not set; must be one of -broadcaster, -transcoder or -orchestrator")
 	}
@@ -282,6 +287,8 @@ func main() {
 			nodeType = "orch"
 		case core.TranscoderNode:
 			nodeType = "trcr"
+		case core.RedeemerNode:
+			nodeType = "rdmr"
 		}
 		lpmon.InitCensus(nodeType, nodeID, core.LivepeerVersion)
 	}
@@ -300,6 +307,7 @@ func main() {
 	}
 
 	watcherErr := make(chan error)
+	redeemerErr := make(chan error)
 	var timeWatcher *watchers.TimeWatcher
 	if *network == "offchain" {
 		glog.Infof("***Livepeer is in off-chain mode***")
@@ -460,8 +468,14 @@ func main() {
 		n.Balances = core.NewAddressBalances(cleanupInterval)
 		defer n.Balances.StopCleanup()
 
-		if *orchestrator {
+		// By default the ticket recipient is the node's address
+		// If the address of an on-chain registered orchestrator is provided, then it should be specified as the ticket recipient
+		recipientAddr := n.Eth.Account().Address
+		if *ethOrchAddr != "" {
+			recipientAddr = ethcommon.HexToAddress(*ethOrchAddr)
+		}
 
+		if *orchestrator {
 			// Set price per pixel base info
 			if *pixelsPerUnit <= 0 {
 				// Can't divide by 0
@@ -488,13 +502,6 @@ func main() {
 			orchSetupCtx, cancel := context.WithCancel(ctx)
 			defer cancel()
 
-			// By default the ticket recipient is the node's address
-			// If the address of an on-chain registered orchestrator is provided, then it should be specified as the ticket recipient
-			recipientAddr := n.Eth.Account().Address
-			if *ethOrchAddr != "" {
-				recipientAddr = ethcommon.HexToAddress(*ethOrchAddr)
-			}
-
 			if err := setupOrchestrator(orchSetupCtx, n, recipientAddr); err != nil {
 				glog.Errorf("Error setting up orchestrator: %v", err)
 				return
@@ -511,7 +518,19 @@ func main() {
 			}
 			defer gpm.Stop()
 
-			sm := pm.NewSenderMonitor(n.Eth.Account().Address, n.Eth, senderWatcher, timeWatcher, n.Database, cleanupInterval, smTTL)
+			var sm pm.SenderMonitor
+			if *redeemerAddr != "" {
+				*redeemerAddr = defaultAddr(*redeemerAddr, "127.0.0.1", RpcPort)
+				rc, err := server.NewRedeemerClient(*redeemerAddr, senderWatcher, timeWatcher)
+				if err != nil {
+					glog.Error("Unable to start redeemer client:", err)
+					return
+				}
+				sm = rc
+			} else {
+				sm = pm.NewSenderMonitor(recipientAddr, n.Eth, senderWatcher, timeWatcher, n.Database, cleanupInterval, smTTL)
+			}
+
 			// Start sender monitor
 			sm.Start()
 			defer sm.Stop()
@@ -579,6 +598,28 @@ func main() {
 				glog.Infof("Maximum transcoding price per pixel is not greater than 0: %v, broadcaster is currently set to accept ANY price.\n", *maxPricePerUnit)
 				glog.Infoln("To update the broadcaster's maximum acceptable transcoding price per pixel, use the CLI or restart the broadcaster with the appropriate 'maxPricePerUnit' and 'pixelsPerUnit' values")
 			}
+		}
+
+		if *redeemer {
+			r, err := server.NewRedeemer(
+				recipientAddr,
+				n.Eth,
+				pm.NewSenderMonitor(recipientAddr, n.Eth, senderWatcher, timeWatcher, n.Database, cleanupInterval, smTTL),
+			)
+			if err != nil {
+				glog.Errorf("Unable to create redeemer: %v", err)
+				return
+			}
+
+			*httpAddr = defaultAddr(*httpAddr, "127.0.0.1", RpcPort)
+			go func() {
+				if err := r.Start(*httpAddr); err != nil {
+					redeemerErr <- err
+					return
+				}
+			}()
+			defer r.Stop()
+			glog.Infof("Redeemer started on %v", *httpAddr)
 		}
 
 		blockWatchCtx, cancel := context.WithCancel(ctx)
@@ -764,9 +805,11 @@ func main() {
 		s.StartCliWebserver(*cliAddr)
 		close(wc)
 	}()
-	go func() {
-		ec <- s.StartMediaServer(msCtx, *transcodingOptions, *httpAddr)
-	}()
+	if n.NodeType != core.RedeemerNode {
+		go func() {
+			ec <- s.StartMediaServer(msCtx, *transcodingOptions, *httpAddr)
+		}()
+	}
 
 	go func() {
 		if core.OrchestratorNode != n.NodeType {
@@ -799,6 +842,8 @@ func main() {
 		glog.Infof("Video Ingest Endpoint - rtmp://%v", *rtmpAddr)
 	case core.TranscoderNode:
 		glog.Infof("**Liveepeer Running in Transcoder Mode***")
+	case core.RedeemerNode:
+		glog.Infof("**Livepeer Running in Redeemer Mode**")
 	}
 
 	c := make(chan os.Signal)
@@ -810,6 +855,10 @@ func main() {
 	case err := <-ec:
 		glog.Infof("Error from media server: %v", err)
 		return
+	case err := <-redeemerErr:
+		if err != nil {
+			glog.Fatalf("Error starting redemption service: %v", err)
+		}
 	case <-msCtx.Done():
 		glog.Infof("MediaServer Done()")
 		return
--- a/core/livepeernode.go
+++ b/core/livepeernode.go
@@ -41,6 +41,7 @@ const (
 	BroadcasterNode NodeType = iota
 	OrchestratorNode
 	TranscoderNode
+	RedeemerNode
 )
 
 //LivepeerNode handles videos going in and coming out of the Livepeer network.


net/redeemer.proto
```diff
@@ -6,9 +6,14 @@ package net;
 
 service TicketRedeemer {
     rpc QueueTicket(Ticket) returns (google.protobuf.Empty) {}
+    rpc MaxFloat(MaxFloatRequest) returns (MaxFloatUpdate) {}
```
j0sh: Curious, when is this `MaxFloat` RPC call invoked? If it's done alongside the `MonitorMaxFloat` subscriptions to initialize state, then can we just send down the max float immediately after subscribing, and avoid the separate RPC call?

cmd/livepeer/livepeer.go
```diff
@@ -66,6 +66,7 @@ var (
 const RtmpPort = "1935"
 const RpcPort = "8935"
 const CliPort = "7935"
+const gRPCPort = "6935"
```
j0sh: Do we need to define a new port for this service? Can we reuse the existing `RpcPort` on 8935? In general, we have tried to minimize the number of ports we have exposed, because that directly leads to additional operational complexity.

j0sh:
> If either flag is omitted the node will start normally

Want to be sure that we maintain existing behavior wherever possible. So for an orchestrator that's started with `-orchestrator` that should also imply starting a local redeemer service if onchain mode is used. That local redeemer service can then be disabled via specifying `-redeemerAddr` or just `-redeemer=false` (but I suppose with the latter it wouldn't be able to submit winning tickets at all). For a standalone redeemer, the node can be started via `-redeemer` (omitting the `-orchestrator` flag).

net/redeemer.proto
kyriediculous: `MonitorMaxFloat` opens up a single stream to the client to send max float updates for all senders. This stream is started when the client initialises. 

`MaxFloat` would be called during e.g. initial discovery when there is no `maxFloat` value initialised yet for the sender on the client because it has not yet redeemed a winning ticket at the redeemer. This method would never be called during the transcoding workflow as the client should always have local cache available. 

I think what you describe could be possible now that I think of it if `client.MaxFloat()` opens up the subscription rather than when queueing the first ticket. We would have to create a `MonitorMaxFloatReq` message that contains the sender field. Will look further into it!

kyriediculous:
> So for an orchestrator that's started with -orchestrator that should also imply starting a local redeemer service if onchain mode is used. 

If neither flag is used the orchestrator will just use the current workflow without a redemption service. The node will be responsible for redeeming winning tickets locally as it does currently (although with a DB backed queue and a reworked sendermonitor implementation)

Users in a single-O setup have the ability to do key separation however if they start with the `-redeemer` flag. This will spin up a local redeemer and single client connected to it. 

In a multi O setup a user would start one orchestrator with the `-redeemer` flag and other O's in its cluster can connect to it by specifying its `-redeemerAddr` , this will spin up a client connected to the orchestrator running the redeemer. 

I have not yet forseen a scenario whereby a redeemer is started in standalone mode without supplying the `-orchestrator` flag.

cmd/livepeer/livepeer.go
kyriediculous: I felt like we could reserve a separate port for multi-purpose services in multi-o setups, eg a blockchain event watching service. 

It thought might be good to separate that traffic from discovery/transcode requests.

net/redeemer.proto
j0sh: Ah, I was thinking the workflow was something like this:

1. Orchestrator starts out monitoring nothing
1. When orchestrator receives a new job, invoke a rpc call to monitor for the max float for a specific broadcaster. (Would also immediately receive the current MaxFloat upon subscribing)
1. Receive new float values from redeemer service as necessary for that particular broadcaster
1. When job expires on orchestrator, halt subscription

j0sh:
> If neither flag is used the orchestrator will just use the current workflow without a redemption service. The node will be responsible for redeeming winning tickets locally

Gotcha.

> Users in a single-O setup have the ability to do key separation however if they start with the -redeemer flag. This will spin up a local redeemer and single client connected to it.
>

Can you clarify this part a bit? How are the keys separate with a "local RPC" redeemer service that's sharing the same process as the orchestrator?

> I have not yet forseen a scenario whereby a redeemer is started in standalone mode without supplying the -orchestrator flag.

Running a RPC redeemer service probably means you don't want that redeemer to also handle the video traffic of an orchestrator node. At least, that's the guidance I'd make. I'd be fine with *not* supporting running a RPC redeemer service in the same process as an orchestrator that also handles video, just because that works against the separation of keys / concerns that's gained by using a standalone redeemer. My thoughts boil down to something like this:

* `-orchestrator` : Local redeeming, video processing - same as now
* `-redeemer` : Standalone RPC redeemer. Does not handle video.
* `-orchestrator -redeemerAddr` : Uses the RPC redeemer at `redeemerAddr`

cmd/livepeer/livepeer.go
j0sh: For standalone redeemers, the discovery / transcode paths would not be enabled. I have a preference to not allow RPC redeemers alongside video stuff, but even if we did mix them up, that seems OK.

More sophisticated operators will either be segregating their redeemer / orchestrator appropriately, rendering the issue moot, or filtering routes via proxy, so they can shape their setups as needed.

Minimizing port usage also minimizes bikeshedding on our part over determining which services are appropriate for a new RPC, or trying to figure out whether a new service justifies a new port. Also see earlier discussion on the topic: https://github.com/livepeer/go-livepeer/issues/781#issuecomment-470718308

kyriediculous:
> Can you clarify this part a bit? How are the keys separate with a "local RPC" redeemer service that's sharing the same process as the orchestrator?

Woops sorry got a little ahead of myself here. That would be possible once we allow the redeemer (and other potential services) to be ran as a standalone node.

kyriediculous:
I'd rather rework the main.go file and make the `blockwatcher` into a blockchain event service first before supporting standalone redeemers just because from an implementation standpoint main.go already is a mess and adding standalone support would turn it into an even greater mess I think. But that's my only argument and it's a temporary one.

net/redeemer.proto
kyriediculous: That does not work because discovery and receiving jobs are two different things. We also need `maxFloat` during discovery. Let's avoid a scenario whereby we start monitoring max float for every sender that discovers our node.

j0sh: Yep, gotcha - it makes sense why we'd need `MaxFloat` during the initial phase, eg discovery. In this case, the additional RPC call for a per-broadcaster subscription is less of a big deal - it can be initiated asynchronously when a job is received, and the cached value used in the meantime. But as mentioned in https://github.com/livepeer/go-livepeer/pull/1513#discussion_r432720399 , keeping RPC monitoring stateless is OK for now.

yondonfu:
> I'd rather rework the main.go file and make the blockwatcher into a blockchain event service first before supporting standalone redeemers 

I agree with Josh that this PR should support standalone redeemers and that transcoding should be turned off by default when the `-redeemer` flag is used. One of the benefits of a ticket redemption service is the ability to separate transcoding operations and redemption operations onto different machines so I think it is reasonable for running a ticket redemption service to imply that you're going to transcode on a different machine.

IIUC your point in the quoted comment is that ideally a single O can propagate *all* blockchain related information to other Os so that each O does not need its own Ethereum connection. I think that makes sense, but I think it is an enhancement as opposed to a blocker for a standalone redeemer since an O fleet with a standalone redeemer can still function fine if each O has its own Ethereum connection.

> just because from an implementation standpoint main.go already is a mess and adding standalone support would turn it into an even greater mess I think.

I don't think adding standalone redeemer support would be much different from the current set of changes in this PR. I think this would require:

- Not starting a redeemer client when the `-redeemer` flag is used
- Turning off the [orchestrator server](https://github.com/livepeer/go-livepeer/blob/d9ea3358e0a43f07a0fd78880f37bd48750c1be2/cmd/livepeer/livepeer.go#L780) when the `-redeemer` flag is used

I think the last requirement can be accomplished with something like:

```
if *redeemer  == "" {
    go func() {
        // Start orchestrator server
    }()
}
```

kyriediculous:
This PR is now updated to support standalone redeemer instances (with other services as well). 

A redeemer can be started using `livepeer -orchestrator -redeemer -httpAddr <host:port> ...` 

An orchestrator connecting to this service can be started using 
`livepeer -orchestrator -ethOrchAddr <on-chain addr> -redeemerAddr <redeemer URL> ...`

j0sh:
> A redeemer can be started using livepeer -orchestrator -redeemer -httpAddr <host:port>

Is the `-orchestrator` necessary for standalone redeemers?

cmd/livepeer/livepeer.go
```diff
@@ -127,6 +127,9 @@ func main() {
 	pixelsPerUnit := flag.Int("pixelsPerUnit", 1, "Amount of pixels per unit. Set to '> 1' to have smaller price granularity than 1 wei / pixel")
 	// Interval to poll for blocks
 	blockPollingInterval := flag.Int("blockPollingInterval", 5, "Interval in seconds at which different blockchain event services poll for blocks")
+	// Redemption service
+	redeemer := flag.Bool("redeemer", false, "Set to true to run a ticket redemption service")
+	redeemerAddr := flag.String("redeemerAddr", "127.0.0.1:"+RpcPort, "URL of the ticket redemption service to use")
```
yondonfu: Since this flag has a default value, it will always be the case that `redeemerAddr != ""`. As a result, a `RedeemerClient` will always be created when the node starts with the `-orchestrator` flag. I think we want the default behavior to be for the node to use `LocalSenderMonitor` instead of `RedeemerClient` in `pm.Recipient` when the node starts with the `-orchestrator` flag.

One way to address this could be to set the default value for `-redeemerAddr` to the empty string and then if `redeemerAddr != ""`, we can use:

```
*redeemerAddr = defaultAddr(*redeemerAddr, "127.0.0.1", RpcPort)
```

to fill in defaults for the IP or port if either is missing from the provided value.

cmd/livepeer/livepeer.go
```diff
@@ -581,6 +600,32 @@ func main() {
...
+				n.Eth.Account().Address,
+				n.Eth,
+				pm.NewSenderMonitor(recipientAddr, n.Eth, senderWatcher, timeWatcher, n.Database, cleanupInterval, smTTL),
+			)
+			if err != nil {
+				glog.Errorf("Unable to create redeemer: %v", err)
+				return
+			}
+
+			uri, err := url.Parse(*httpAddr)
```
yondonfu: We could do something like:

```
*httpAddr = defaultAddr(*httpAddr, "127.0.0.1", RpcPort)
```

so the user can pass in values like `:<PORT>` or `<IP>:` with the defaults used for the missing fields. We do this in other places in this file for `httpAddr`, but I don't think we do it before this code executes. I think `defaultAddr` should also just return the passed in address if all fields are filled in so it should be safe to use multiple times on the same address, but this is worth double checking.

cmd/livepeer/livepeer.go
```diff
@@ -815,6 +864,9 @@ func main() {
 	case err := <-ec:
 		glog.Infof("Error from media server: %v", err)
 		return
+	case err := <-redeemerErr:
+		glog.Errorf("Error starting redemption service: %v", err)
```
yondonfu: Since `grpc.Server.Serve()` (called by `RedeemerClient.Start()`) will return nil if `grpc.Server.Stop()` is called, maybe we should check if the error is not nil here before logging.

yondonfu (changes requested):

cmd/livepeer/livepeer.go
kyriediculous: fixed in c0f617298df0d7c9472e110eb0d33e06f5cc99bf

cmd/livepeer/livepeer.go
kyriediculous: fixed in c0f617298df0d7c9472e110eb0d33e06f5cc99bf

cmd/livepeer/livepeer.go
kyriediculous: fixed in c0f617298df0d7c9472e110eb0d33e06f5cc99bf

yondonfu (commented):
Changes look good to me. 👍  for a rebase from me as long as comments from other reviewers have been addressed.

kyriediculous:
Now starting node type enums from 1 instead of 0 so that we don't run as broadcaster when the nodetype is unset (when running in redeemer mode). 

Don't run mediaserver and cliserver in redeemer mode

cmd/livepeer/livepeer.go
```diff
@@ -765,19 +802,25 @@ func main() {
...
-	go func() {
-		s.StartCliWebserver(*cliAddr)
-		close(wc)
-	}()
-	go func() {
-		ec <- s.StartMediaServer(msCtx, *transcodingOptions, *httpAddr)
-	}()
+	if n.NodeType != 0 {
+		go func() {
+			s.StartCliWebserver(*cliAddr)
```
yondonfu: Is there a downside to starting the CLI server when running a redeemer? I could see benefits in allowing the node operator to still use CLI server endpoints when running the node as a redeemer.

core/livepeernode.go
```diff
@@ -38,7 +38,7 @@ var MaxSessions = 10
 type NodeType int
 
 const (
-	BroadcasterNode NodeType = iota
+	BroadcasterNode NodeType = iota + 1
```
yondonfu: Can we define an actual enum at the end of this list instead of changing the initial numbering here?

yondonfu (changes requested):

cmd/livepeer/livepeer.go
kyriediculous: There could certainly be benefits but I don't think a standalone redeemer has much use for the existing CLI endpoints.

core/livepeernode.go
kyriediculous: I had initially defined something like `ServiceNode` , but in the end it felt cleaner to stick with the current existing enums and have a standalone redeemer deployment not have " node type" for now.

j0sh: Why not a "RedeemerNode" ?

kyriediculous: in the future it could also be a Redeemer + Blockwatcher + RoundInitialiser , so to not get caught up on naming I decided to just make a zero value

cmd/livepeer/livepeer.go
yondonfu: The CLI server would be required to export metrics from the node - in the case of a redeemer, relevant metrics would be be for value redeemed and ticket redemption errors. The CLI server also exposes endpoints such as `/status` which are convenient for health checks when a node is run in a cluster. So, I think it'd be better if the CLI server is still run for a redeemer.

core/livepeernode.go
yondonfu: There is existing code that maps a node type to string - here is an [example](https://github.com/livepeer/go-livepeer/blob/ed7f24c763ad8db4592db5b4d1a2f2ce83aa6ca0/cmd/livepeer/livepeer.go#L280). So, a new node type should have an associated name + string representation. Otherwise, when a redeemer is run, based on the code in the linked example, the node will call `InitCensus()` with nodeType = "bctr".

If the worry is around changing the node type name in the future, then we could just use the following check that doesn't rely on a node type:

```golang
if !*redeemer {
    // Do something if -redeemer=false
}
```

j0sh: > Redeemer + Blockwatcher + RoundInitialiser

That seems okay - a "RedeemerNode" could also implicitly run the blockwatcher + round initializer. If we ever needed to make one of these standalone by default, we can also add those as separate node types, eg "BlockwatcherNode" . But that doesn't preclude a RedeemerNode also being a blockwatcher.

We do similar for OT mode, where a `-transcoder` alone is a standalone TranscoderNode, while an `-orchestrator -transcoder` is an OrchestratorNode that implicitly runs a transcoder.

kyriediculous: Ok let's go with an additional definition then , that seems like the best way forward from here.

kyriediculous: fixed in d5993ce

cmd/livepeer/livepeer.go
kyriediculous: fixed in d5993ce

j0sh (approved):
🚢

yondonfu (approved):
LGTM 🚢  pending green CI (which I think should be fixed by rebasing with master to pick up the latest fix for `TestSubmitSegment_HttpPostError`).

PR #1525, opened 2020-05-30
Add non-integer framerate support

### LINKED ISSUE #1474: Non Integer Frame Rate Support

jailuthra:

This PR adds support for non-integer framerates (#1474) by passing an optional FPS denominator field everywhere (like protobufs, webhooks etc.).

- bce44ea817ec6df1049f30da6e8ccc3517174c07
  - Add FPS Denominator field in VideoProfile protobuf struct
  - Verify correct conversion of VideoProfile structs between LPMS <-> Protobuf
  - Ensure all uses of FPS like pixel calc & cost calc use denominator as well
- b669b5549ff367d01a4211df0bdc6de1652a6fcb Add ability to set fps denominator through the AuthWebhook's JSON response
- 0ef95a5621a41a99d3dad72925c59a19cfbf75ef Add `FullProfiles3` field to maintain back compat with late upgraders who wouldn't support the new Denominator field
- 2958518ecda29f7a138431d7309a1f6e93e4cf26 Serialize `ffmpeg.VideoProfile` to `net.VideoProfile`, also update webhook docs.

- Unit Tests for every change
- Manual Testing
  - [x] Non-int FPS with new B -> new OT Works
  - [x] Non-int FPS with new B -> old OT Fails
  - [x] Int FPS with old B -> new OT Works

Fixes #1474 

**Checklist:**

### DIFF
--- a/core/streamdata.go
+++ b/core/streamdata.go
@@ -61,7 +61,7 @@ func NetSegData(md *SegTranscodingMetadata) (*net.SegData, error) {
 		Hash:       md.Hash.Bytes(),
 		Storage:    storage,
 		Duration:   int32(md.Duration / time.Millisecond),
-		// Triggers failure on Os that don't know how to use FullProfiles/2
+		// Triggers failure on Os that don't know how to use FullProfiles/2/3
 		Profiles: []byte("invalid"),
 	}
 
@@ -76,10 +76,20 @@ func NetSegData(md *SegTranscodingMetadata) (*net.SegData, error) {
 			allTS = false
 		}
 	}
-	if allTS {
-		segData.FullProfiles = fullProfiles
+	allIntFPS := true
+	for i := 0; i < len(md.Profiles) && allIntFPS; i++ {
+		if md.Profiles[i].FramerateDen != 0 {
+			allIntFPS = false
+		}
+	}
+	if allIntFPS {
+		if allTS {
+			segData.FullProfiles = fullProfiles
+		} else {
+			segData.FullProfiles2 = fullProfiles
+		}
 	} else {
-		segData.FullProfiles2 = fullProfiles
+		segData.FullProfiles3 = fullProfiles
 	}
 
 	return segData, nil
--- a/net/lp_rpc.proto
+++ b/net/lp_rpc.proto
@@ -160,6 +160,9 @@ message VideoProfile {
     MP4        = 1;
   }
   Format format = 21;
+
+  // FPS Denominator of VideoProfile
+  uint32 fpsDen = 22;
 }
 
 // Individual transcoded segment data.
--- a/net/lp_rpc.proto
+++ b/net/lp_rpc.proto
@@ -135,7 +135,11 @@ message SegData {
   repeated VideoProfile fullProfiles = 33;
 
   // Transcoding profiles to use. Supersedes `fullProfiles` field
+  // Deprecated by `fullProfiles3` but may still be used for integer FPS
   repeated VideoProfile fullProfiles2 = 34;
+
+  // Transcoding profiles to use. Supersedes `fullProfiles2` field
+  repeated VideoProfile fullProfiles3 = 35;
 }
 
 message VideoProfile {
--- a/server/mediaserver.go
+++ b/server/mediaserver.go
@@ -116,6 +116,7 @@ type authWebhookResponse struct {
 		Height  int    `json:"height"`
 		Bitrate int    `json:"bitrate"`
 		FPS     uint   `json:"fps"`
+		FPSDen  uint   `json:"fpsDen"`
 	} `json:"profiles"`
 }
 
@@ -217,10 +218,11 @@ func createRTMPStreamIDHandler(s *LivepeerServer) func(url *url.URL) (strmID str
 						profile.Bitrate)
 				}
 				prof := ffmpeg.VideoProfile{
-					Name:       name,
-					Bitrate:    fmt.Sprint(profile.Bitrate),
-					Framerate:  profile.FPS,
-					Resolution: fmt.Sprintf("%dx%d", profile.Width, profile.Height),
+					Name:         name,
+					Bitrate:      fmt.Sprint(profile.Bitrate),
+					Framerate:    profile.FPS,
+					FramerateDen: profile.FPSDen,
+					Resolution:   fmt.Sprintf("%dx%d", profile.Width, profile.Height),
 				}
 				profiles = append(profiles, prof)
 			}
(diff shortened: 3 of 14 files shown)


server/mediaserver_test.go
```diff
@@ -475,7 +476,7 @@ func TestCreateRTMPStreamHandlerWebhook(t *testing.T) {
 	// set profiles and presets
 	ts9 := makeServer(`{"manifestID":"a", "presets":["P240p30fps16x9", "P720p30fps16x9"], "profiles": [
 		{"name": "prof1", "bitrate": 432, "fps": 560, "width": 123, "height": 456},
-		{"name": "prof2", "bitrate": 765, "fps": 876, "width": 456, "height": 987},
+		{"name": "prof2", "bitrate": 765, "fps": 876, "fps_den": 12, "width": 456, "height": 987},
 		{"name": "passthru_fps", "bitrate": 890, "width": 789, "height": 654}]}`)
```
j0sh, line 482: Might be good to sanity-check the `FramerateDen` field as zero somewhere, maybe as part of the existing check for `passthru_fps` later on in this test case.

server/segment_rpc_test.go
```diff
@@ -386,31 +386,35 @@ func TestMakeFfmpegVideoProfiles(t *testing.T) {
...
 			Fps:     uint32(912),
+			FpsDen:  uint32(15),
 		},
 		{
 			Name:    "prof2",
 			Width:   int32(987),
 			Height:  int32(654),
 			Bitrate: int32(321),
 			Fps:     uint32(198),
+			FpsDen:  uint32(101),
```
j0sh: Should we omit this, just to sanity check the zero value?

server/rpc.go
```diff
@@ -313,7 +313,9 @@ func coreSegMetadata(segData *net.SegData) (*core.SegTranscodingMetadata, error)
 	}
 	var err error
 	profiles := []ffmpeg.VideoProfile{}
-	if len(segData.FullProfiles2) > 0 {
+	if len(segData.FullProfiles3) > 0 {
+		profiles, err = makeFfmpegVideoProfiles(segData.FullProfiles3)
```
j0sh, line 317: Needs a test case for this, probably in [`TestCoreSegMetadata_FullProfiles`](https://github.com/livepeer/go-livepeer/blob/d9ea3358e0a43f07a0fd78880f37bd48750c1be2/server/segment_rpc_test.go#L330)

core/streamdata.go
```diff
@@ -76,10 +76,20 @@ func NetSegData(md *SegTranscodingMetadata) (*net.SegData, error) {
...
+	}
+	if allIntFPS {
+		if allTS {
+			segData.FullProfiles = fullProfiles
+		} else {
+			segData.FullProfiles2 = fullProfiles
+		}
 	} else {
-		segData.FullProfiles2 = fullProfiles
+		segData.FullProfiles3 = fullProfiles
```
j0sh, line 92: Needs test cases for generating FullProfiles3 with frac FPS. This behavior has been mostly tested within [`TestGenSegCreds_FullProfiles`](https://github.com/livepeer/go-livepeer/blob/d9ea3358e0a43f07a0fd78880f37bd48750c1be2/server/segment_rpc_test.go#L225) and I believe that will also ensure that the back-compat behavior is preserved.

(That particular test case is currently a bit out of place ever since NetSegData was factored out of `genSegCreds` but I forgot to also move out the test cases as part of https://github.com/livepeer/go-livepeer/pull/1511 . Feel free to put the new test wherever you like.)

server/segment_rpc_test.go
j0sh: Should we sanity check the non-populated zero case for `FpsDen` as well?

net/lp_rpc.proto
```diff
@@ -160,6 +164,9 @@ message VideoProfile {
     MP4        = 1;
   }
   Format format = 21;
+
+  // FPS Denominator of VideoProfile
+  uint32 fps_den = 22;
```
j0sh: I think we've used a camelCase convention here so far; `fpsDen` is probably more consistent?

server/mediaserver.go
```diff
@@ -116,6 +116,7 @@ type authWebhookResponse struct {
 		Height  int    `json:"height"`
 		Bitrate int    `json:"bitrate"`
 		FPS     uint   `json:"fps"`
+		FPSDen  uint   `json:"fps_den"`
```
j0sh: I think we've used a camelCase convention for the other JSON fields (`manifestID` and `streamKey`); `fpsDen` is probably more consistent?

j0sh (changes requested):
Looks great, just a few comments around test cases and field naming.

The webhook documentation also needs to be updated to reflect the new field: https://github.com/livepeer/go-livepeer/blob/master/doc/rtmpwebhookauth.md It probably doesn't hurt to mention in the doc that if the field is optional, and if  omitted, a value of `1` is used by default.

net/lp_rpc.proto
jailuthra: I saw snake_style being used in all other protobuf structs except SegData. But as this is part of SegData I'll use camelCase here as well.

server/segment_rpc_test.go
```diff
@@ -386,31 +411,35 @@ func TestMakeFfmpegVideoProfiles(t *testing.T) {
...
 			Fps:     uint32(912),
+			FpsDen:  uint32(15),
 		},
 		{
 			Name:    "prof2",
 			Width:   int32(987),
 			Height:  int32(654),
 			Bitrate: int32(321),
 			Fps:     uint32(198),
+			//FpsDen:  0,
```
j0sh: Stray? It's ok to remove this comment and keep the zero value implicit

jailuthra: I wanted to keep it explicit for the reader but I see this might be confusing a few months down the line, will remove it and update the commit after testing manually :)

j0sh (approved):
🚢

PR #1537, opened 2020-06-11
Miscellaneous test-related changes.

j0sh:

Miscellaneous changes to various aspects of the unit tests.

- e96be1c test: Simplify test script.
There's a couple effects from this change:
  - Reduces the verbosity of the tests, making it easier to spot failing cases
  - Using `set -e` makes the script exit immediately on a nonzero exit code. Mitigates the need to explicitly track and check test results.

- 12cb7eb discovery: Replace actual addresses with stubs. … 
This particular test fails if there happens to be another
process listening at any of the given ports.
For example, an actual Livepeer process.
Using dummy values seems to work and won't collide.

- 3e9d02d watcher: Increase sleep times during tests. … 
Makes tests less flaky on slower CI instances.

- b9cc7ef install_ffmpeg: Decompress and extract in a single step. … 
Not all machines have xd installed.

Unit tests

**Checklist:**

### DIFF
--- a/install_ffmpeg.sh
+++ b/install_ffmpeg.sh
@@ -59,9 +59,7 @@ if [[ $(uname) != *"MSYS"* ]]; then
   if [ ! -e "$HOME/gmp-6.1.2" ]; then
     cd "$HOME"
     curl -LO https://github.com/livepeer/livepeer-builddeps/raw/34900f2b1be4e366c5270e3ee5b0d001f12bd8a4/gmp-6.1.2.tar.xz
-    xz -d gmp-6.1.2.tar.xz
-    tar xf gmp-6.1.2.tar
-    rm gmp-6.1.2.tar
+    tar xf gmp-6.1.2.tar.xz
     cd "$HOME/gmp-6.1.2"
     ./configure --prefix="$HOME/compiled" --disable-shared  --with-pic --enable-fat
     make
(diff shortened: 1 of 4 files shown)


install_ffmpeg.sh
```diff
@@ -84,9 +81,7 @@ if [[ $(uname) != *"MSYS"* ]]; then
   if [ ! -e "$HOME/gnutls-3.5.18" ]; then
     cd $HOME
     curl -LO https://www.gnupg.org/ftp/gcrypt/gnutls/v3.5/gnutls-3.5.18.tar.xz
-    xz -d gnutls-3.5.18.tar.xz
-    tar xf gnutls-3.5.18.tar
-    rm gnutls-3.5.18.tar
```
yondonfu, line 89: I don't have a strong preference, but should we remove the archive after decompressing?

j0sh: So far, we haven't been removing the downloaded files - only the intermediate tar files. Right now, we don't produce those intermediate files anymore, so that's why those lines were removed.

Only a slight preference for keeping the downloaded files - the pack rat in me likes to have the downloaded files, just because you never know if they might be handy some day (suppose the archive disappears from the Internet!) but I don't really have a better argument in favor of it. Happy to remove if others are more inclined that way.

I've been debating whether to move the working / install directory for the `install_ffmpeg.sh` script into `go-livepeer/deps` just so the downloaded / extracted stuff would be out of the way and self-contained within `deps`. Eg, `HOME=$PWD`. That might break CI caching without a bit more work though - to the extent that "CI caching" actually happens anymore (we used to attempt caching of ffmpeg and friends to improve CI run times).

yondonfu (approved):
LGTM! :shipit:

Issue #1519, opened 2020-05-26
Capability Discovery


j0sh:
## Abstract

A mechanism for capability discovery is proposed. Capability discovery eases compatibility and discovery concerns related to operating in a heterogeneous network where nodes support a mix of features. The technical mechanism is as follows:
* A [bitstring](https://en.wikipedia.org/wiki/Bit_array) enumerates the orchestrator's supported and broadcaster's required features
* An additional message to specify non-binary constraints such as a numeric range of acceptable parameters. We do not currently require any constraints, but the basic mechanism is sketched here for completeness.
* The broadcaster filters for orchestrators satisfying its requirements during the discovery phase, matching on the bitstring and comparing against the constraints.

## Motivation

When a new feature is added that requires support on both sides of the B-O network boundary [1] (recent examples: MP4 support, non-integer frame rates, durations), then consideration needs to be given towards compatibility. A robust compatibility mechanism is important to minimize the impact of changes, to maximize the reach of the network, and to provide a framework for achieving such  while minimizing engineering effort. The criteria for a compatibility system are as follows:

1. How to identify nodes that have upgraded to support a new feature
2. How to identify nodes that might not support the new feature, whether due to being non-upgraded or having the feature disabled
3. How to interoperate with non-upgraded nodes when the new feature is not needed

Capability discovery is proposed as a mechanism for this.

[1] The O-T network boundary also has similar issues around compatibility, since transcoders cannot be expected to be upgraded in tandem with orchestrators. The same capability matching mechanism should be re-used during transcoder selection by orchestrators.

### Proposed Solution

Two parts: feature bit-string and capability constraints.

The capabilities of the orchestrator are advertised as a bitstring. Each bit in the string corresponds to a certain capability (or feature). Capabilities using this bitstring mechanism are binary: either they are enabled, or not.

On discovery, the orchestrator sends down its capability bitstring within the `OrchestratorInfo` response. The broadcaster constructs its own (reduced) bitstring corresponding to the transcoding requirements for that particular job. Capability checking of a particular orchestrator corresponds to checking that the AND between the two bitstrings is equivalent to the broadcaster's own capability requirements:

```
isOrchestratorCapable := (broadcaster.bitstring AND orchestrator.bitstring) == broadcaster.bitstring
```

[...]

PR #1538, opened 2020-06-11
Move StreamParameters into core

j0sh:

The upcoming support for capability discovery (https://github.com/livepeer/go-livepeer/issues/1519) means we need a way to:

* Determine the capabilities a job requires
* Use these capabilities to filter out eligible orchestrators during discovery

The most suitable, pre-existing structure for this seems to be `streamParameters`. Currently, `streamParameters` is used to propagate job-specific data between LPMS handlers, and is the primary configuration structure used by the webhook. It is also used during orchestrator selection. Hence, adding capabilities to `streamParameters` seems suitable, since most of the ... stream parameters .... required to generate the capabilities are already present within the structure. However, the current form of `streamParameters` poses a few issues for capability discovery:

* We probably don't want capability discovery to be located within `server` as that is large enough. `core` seems to be a better fit. (Happy to discuss this point further if anyone feels otherwise.)

* Fields within `streamParameters` need to be public to facilitate its use from `server` and `core`.

* `streamParameters` doesn't yet capture all the information that might be needed for capability discovery. One example: indicating which storage system the broadcaster is using. This would be useful as we expand driver support (eg: Azure, IPFS, Sia, etc) and need to select for orchestrators that support these drivers.

* Moves `server.streamParameters` into `core.StreamParameters` 3ccf893
* Makes fields public for use within other modules 9e4998c
* Adds an "OS" field to the stream parameters. Not strictly necessary, but done mostly as an example. One nice side effect is this is we can remove the `PlaylistManager` from the orchestrator selection. f0d9fe1

Unit testing. This is mostly a refactoring; no functional changes.

**Checklist:**

### DIFF
--- a/server/mediaserver.go
+++ b/server/mediaserver.go
@@ -68,26 +68,13 @@ var httpPushResetTimer = func() (context.Context, context.CancelFunc) {
 	return context.WithTimeout(context.Background(), sleepDur)
 }
 
-type streamParameters struct {
-	ManifestID core.ManifestID
-	RtmpKey    string
-	Profiles   []ffmpeg.VideoProfile
-	Resolution string
-	Format     ffmpeg.Format
-	OS         drivers.OSSession
-}
-
-func (s *streamParameters) StreamID() string {
-	return string(s.ManifestID) + "/" + s.RtmpKey
-}
-
 type rtmpConnection struct {
 	mid         core.ManifestID
 	nonce       uint64
 	stream      stream.RTMPVideoStream
 	pl          core.PlaylistManager
 	profile     *ffmpeg.VideoProfile
-	params      *streamParameters
+	params      *core.StreamParameters
 	sessManager *BroadcastSessionsManager
 	lastUsed    time.Time
 }
@@ -260,7 +247,7 @@ func createRTMPStreamIDHandler(s *LivepeerServer) func(url *url.URL) (strmID str
 		if key == "" {
 			key = common.RandomIDGenerator(StreamKeyBytes)
 		}
-		return &streamParameters{
+		return &core.StreamParameters{
 			ManifestID: mid,
 			RtmpKey:    key,
 			// HTTP push mutates `profiles` so make a copy of it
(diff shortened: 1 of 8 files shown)


server/mediaserver.go
```diff
@@ -395,21 +383,25 @@ func (s *LivepeerServer) registerConnection(rtmpStrm stream.RTMPVideoStream) (*r
...
 	}
-	storage := drivers.NodeStorage.NewSession(string(mid))
+	var storage drivers.OSSession
 	// Build the source video profile from the RTMP stream.
-	if params.resolution == "" {
-		params.resolution = fmt.Sprintf("%vx%v", rtmpStrm.Width(), rtmpStrm.Height())
+	if params.Resolution == "" {
+		params.Resolution = fmt.Sprintf("%vx%v", rtmpStrm.Width(), rtmpStrm.Height())
+	}
+	if params.OS == nil {
```
yondonfu, line 395: Wouldn't this mean that if `params.OS` is not nil then a nil `storage` will be passed to `NewBasicPlaylistManager()` on L416 instead of `params.OS`?

yondonfu (changes requested):

server/mediaserver.go
j0sh: Well spotted. Fixed in c204e276def1ea4ed10188cdbec18107c755826e

Technically the storage is not used anymore within the playlist manager. Considered removing it from the playlist manager entirely, but opted to keep the scope of changes in this PR narrow.

j0sh:
Rebased on top of master to fix a merge conflict warning in the github UI - no other functional changes

yondonfu (approved):
LGTM

j0sh:
The fixup https://github.com/livepeer/go-livepeer/pull/1538/commits/15dd055e8020cb53504a738106737502c39039e2 was missing a proper test case, so added one here https://github.com/livepeer/go-livepeer/pull/1538/commits/2ec1cf6c6e9d9163305429e68a4adafec5c98b80 (verified that it fails without the fix.)

j0sh:
Rebased to master, fixed up commits, merged

PR #1414, opened 2020-03-10
bump Linux build to go 1.14

iameli:
It wasn't in the Ubuntu golang backports PPA we were using for some reason. So I downloaded it manually from Google.

Our Darwin and Windows versions are dependent on whatever homebrew and MSYS2 install, so it turns out they were already on Go 1.14. Sweet! That's a pretty good indication that it works okay.

PR #1539, opened 2020-06-11
Capability discovery

### LINKED ISSUE #1519: Capability Discovery

j0sh:

Implements the first stages of capability discovery - https://github.com/livepeer/go-livepeer/issues/1519

This does not implement (and will not implement):

* Additional constraints. Only stubs out the structures.
* Capability storage in broadcaster DB
* Capability discovery for standalone T

- net: Regenerate proto file.  c2aed29
- core: Add support for capability enumeration.  2a3f05a
- discovery: Add capability discovery. cba60a5
- core: Add encoding profiles to capabilities. bef2014 Added to demonstrate how to add a new capability. Note that the list of profiles is explicit, if one ever needs to be disabled or superseded.
- core: Add GOP to capabilities. cb4d6ae Added to demonstrate how to add a new capability.
- core: Add mandatory capability support. efdc188 See https://github.com/livepeer/go-livepeer/issues/1519 for a description of orchestrator mandatory capabilities. Currently, nothing is mandatory, but some things might be soon (eg, transcoding protocol authentication - https://github.com/livepeer/go-livepeer/issues/1569 )
-server: Add orchestrator checking of segment capabilities. … e79ec85 . Acts as a validation check prior to segment processing, and needed for capability matching with standalone T (not currently supported).

* Unit testing
* Regression testing via https://gist.github.com/j0sh/9d884a19bb92e504092e71fd19d812e5
* Manual testing, including  builds with certain capabilities intentionally disabled to induce failures. (There is currently no way to selectively disable capabilities outside of compile time.)
  * Capability Discovery B,  Old O
  * Old B, Capability Discovery O
  * Capability Discovery B, Capability Discovery O, All Features Enabled
  * Capability Discovery B, Capability Discovery O, mpegts disabled, MP4 transcoding request
  * Capability Discovery B, Capability Discovery O, mpegts disabled, mpegts transcoding request
  * [ ] TODO Test a build with object storage disabled, and external OS in use

Fixes https://github.com/livepeer/go-livepeer/issues/1519

**Checklist:**


core/capabilities.go
```diff
@@ -0,0 +1,166 @@
...
+const (
+	Capability_Invalid Capability = iota - 2
+	Capability_Unused
+	Capability_H264
+	Capability_MPEGTS
+	Capability_MP4
+	Capability_FractionalFramerates
+	Capability_StorageDirect
+	Capability_StorageS3
+	Capability_StorageGCS
```
j0sh, line 30: Figured it'd be good to be as specific as possible about all the features that are being used, including the storage driver, in case we ever need to disable support for one of these someday.

jailuthra (approved):
LGTM!

core/capabilities.go
```diff
@@ -0,0 +1,166 @@
...
+	for k, _ := range caps {
+		capList = append(capList, k)
+	}
+
+	return &Capabilities{bitstring: NewCapabilityString(capList)}, nil
+}
+
+func (bcast *Capabilities) CompatibleWith(orch *net.Capabilities) bool {
+	if bcast == nil {
+		// Weird golang behavior: interface value can evaluate to non-nil
```
kyriediculous, line 104: An interface in go will only equal nil if both the type and value are nil , since an interface consist of both a type and a value.

kyriediculous:
Just lurking here 👀  

Uttermost interested in "Capability discovery for standalone T"

Should we consider gitignoring the generated grpc go stubs because they're part of the makefile anyhow (I notice a conflict) ?

j0sh:
> Should we consider gitignoring the generated grpc go stubs

@kyriediculous Do you mean the `net/lp_rpc.pb.go` file ? That would require the protobuf-go compiler be installed - it's a bit easier sometimes without that. Also, different versions of the compiler can cause unwanted changes or even breakage as @jailuthra encountered recently [1]. Checking in the generated code allows us to punt on the issue and ensure everything is synchronized for everyone.

[1] We should definitely still fix these types of issues eventually, maybe at the same time we fix go 1.14 😅  https://github.com/livepeer/go-livepeer/pull/1414

j0sh:
Rebased against master and added a few more things.

Profiles and GOP support (separately) to demonstrate how to add new capabilities.

Mandatory capabilities because this allows for "forward compatibility" - even though we don't have any such mandatory capabilities right now, broadcasters that know to check the field will be able to reject orchestrators that add mandatory capabilities in the future, even if the broadcasters themselves are not aware of the capability. See https://github.com/livepeer/go-livepeer/issues/1519 for more details.

Orchestrator checking of job capabilities - because this is going to be needed at some point, and it's a good sanity check on the orchestrator side.

Still needs a little more manual testing, especially around object storage, but things are ready for review. This PR is as complete as it's going to get, feature-wise.

core/capabilities.go
```diff
@@ -0,0 +1,246 @@
...
+	// set expected storage
+	storageCap, err := storageToCapability(params.OS)
+	if err != nil {
+		return nil, err
+	}
+	caps[storageCap] = true
+
+	// generate bitstring
+	capList := []Capability{}
+	for k, _ := range caps {
```
yondonfu: Should be able to shorten this to `for k := range caps`

core/capabilities.go
```diff
@@ -0,0 +1,246 @@
...
+
+	if bcast == nil {
+		// Weird golang behavior: interface value can evaluate to non-nil
+		// even if the underlying concrete type is nil.
+		// cf. common.CapabilityComparator
+		return false
+	}
+
+	// For now, check this:
+	// ( orch.mandatories  AND orch.bitstring  ) == orch.mandatories &&
```
yondonfu: Is this comment accurate? Based on the checks below I would think that this comment should say:

```
( orch.mandatories AND bcast.bitstring ) == orch.mandatories &&
( bcast.bitstring AND orch.bitstring ) == bcast.bitstring
```

server/rpc.go
```diff
@@ -340,12 +342,20 @@ func coreSegMetadata(segData *net.SegData) (*core.SegTranscodingMetadata, error)
 		dur = 2 * time.Second // assume 2sec default duration
 	}
 
+	caps := core.CapabilitiesFromNetCapabilities(segData.Capabilities)
+	if caps == nil {
+		// For older broadcasters. Note if there are any orchestrator
+		// mandatory capabilities, segment verification will fail.
```
yondonfu: Should "segment verification will fail" say something like "seg creds verification will fail" since I believe it is referring to the compatibility check in `verifySegCreds()`?

yondonfu (commented):
Looks great just a few minor comments.

core/capabilities.go
j0sh: Fixed in 176c77e0d5ec5b7cd30e1cb13746ee16728d9513

core/capabilities.go
j0sh: Yep, typos. Fixed in 192981f3893798f656f0ea973d4b45c48a2e63a6

server/rpc.go
j0sh: Fixed in 2eca7be58e2990dabf716372c7734bee1e7d992f

yondonfu (approved):
LGTM :shipit:

PR #1542, opened 2020-06-19
Add support for H264 profiles

### LINKED ISSUE #1531: H.264 Profiles
## Abstract

This is a proposal to add support for configuring the H.264 profiles (baseline, main, high) of the videos transcoded by go-livepeer.

## Motivation

Some decoders support limited H.264 profiles. For example OpenH264 used for WebRTC only supports the Baseline profile. Similarly [older phones](https://superuser.com/a/489092) can usually playback Main and Baseline but not High H.264 content. Refer to https://github.com/livepeer/lpms/issues/190 for the original discussion on B-frame removal for WebRTC ingest.

Most encoders (like ffmpeg, OBS) allow configuring the H.264 profile - extending similar functionality to go-livepeer users is a good idea in general.

## Proposed Solution

Add a new enum field called `Profile` in the VideoProfile structure, which is used to configure the encoder. The enum will comprise of a limited subset of the profiles that we want to support.

The VideoProfile structure is defined in many places throughout the codebase like - LPMS (Go struct), Protobuf and Auth webhook JSON. We will need to add the new field everywhere and finally pass it off as a `AVDictionary` parameter to FFmpeg, similar to the changes done in [#1474 - non-integer framerate support](https://github.com/livepeer/go-livepeer/issues/1474#issuecomment-630952970).

### LPMS

1. Create an enum for the codec profiles with the values `H264_Baseline, H264_Main, H264_High` for now.
2. Add a new field `Profile` of the enum type in the [ffmpeg.VideoProfile](https://github.com/livepeer/lpms/blob/1613618/ffmpeg/videoprofile.go#L28) struct.
3. Use this field to populate the [AVDictionary](https://github.com/livepeer/lpms/blob/1613618/ffmpeg/lpms_ffmpeg.h#L18) passed to the ffmpeg video encoder in the Cgo module. Both CPU (libx264) and GPU (h264_nvenc) encoders support the ffmpeg AVOption `-profile [main/baseline/high]`. Will need to convert the enum value to a String that is supported by the FFmpeg AVOption.

### Go-Livepeer

1. Add a new field `Profile profile` to the Protobuf [VideoProfile struct](https://github.com/livepeer/go-livepeer/blob/dace5d5/net/lp_rpc.proto#L141-L163), where Profile is a Protobuf enum similar to the one made in Go.
2. Ensure the same field is generated (or add it manually) in the [net.VideoProfile](https://github.com/livepeer/go-livepeer/blob/dace5d5982bafe3bbfdc67b122192f49a23f46e2/net/lp_rpc.pb.go#L553-L568) Go struct mapped to the Protobuf.
3. Serialze & Deserialize between ffmpeg.VideoProfile <-> net.VideoProfile
4. Add a new json field in the [authWebhookResponse.Profiles](https://github.com/livepeer/go-livepeer/blob/dace5d5/server/mediaserver.go#L109-L120) which is used to configure go-livepeer via a webhook.

* Extensive unit tests for the new codepaths.
* Test transcoding videos manually with both go-livepeer and lpms and verify that the output has the correct H.264 profile.
* Dedicated unit tests in the CPU & GPU test suites of LPMS to sanity check that all profiles encoded correctly.

## Alternatives

~~1. We could use an **Enum** instead of a String for the H264 profile field in all the structs (similar to how `Format` is defined). This is useful in case we only want to support the three most popular profiles - baseline, main, high and not all the profiles supported by the underlying encoder. The implementation tasks would look very similar, with the added step of converting between String <-> Enum as FFmpeg expects a String.~~

1. We could have used a String directly instead of an enum for the H264 profile field, but we want to limit the surface area we expose to the end user so we've chosen the enum route for now. Look at the [comment below](https://github.com/livepeer/go-livepeer/issues/1531#issuecomment-640949472) for details.
2. The name `Profile` of the new field is in a struct called `VideoProfile`. This might be confusing to the reader. We may alternatively call the new field `H264Profile`.

jailuthra:
This PR adds support for specifying the H264 encoder profile (#1531) by passing it as a field in the VideoProfile related structs.

- 1a4543e510e3eb154a8f15ea13eea37109b1f779 Add the field in Protobuf `net.VideoProfile` struct - Serialize/Deserialize to LPMS VideoProfile
- 1298f091555145b690dc973c1ecc2825c96661e3 Back Compat with older Os by using `FullProfiles3` field only when the new Profiles field is used
- 1786a39c82c95a97ee9919a572c5857ba0417f71 Add ability to configure the H264 profile from the auth webhook JSON response

- Unit Tests for new codepaths
- Manual testing with off-chain B-OT setup

Fixes #1531 

**Checklist:**


go.sum
```diff
@@ -127,6 +127,8 @@ github.com/livepeer/joy4 v0.1.2-0.20191121080656-b2fea45cbded h1:ZQlvR5RB4nfT+cO
 github.com/livepeer/joy4 v0.1.2-0.20191121080656-b2fea45cbded/go.mod h1:xkDdm+akniYxVT9KW1Y2Y7Hso6aW+rZObz3nrA9yTHw=
 github.com/livepeer/lpms v0.0.0-20200529175251-ec8e1e4e2420 h1:KnCB0ax1Nz1qr7NIL2Rv0ZNn9i38DRJusYS3uehd8RM=
 github.com/livepeer/lpms v0.0.0-20200529175251-ec8e1e4e2420/go.mod h1:Vd2O4BVQI/Yj2hzjnNcyTj+qib8zb0hh7GunKP6KJc8=
+github.com/livepeer/lpms v0.0.0-20200619063758-7ae8da134ed4 h1:/0dAE31cmZEHMrmuL/DrAp6Cys++9F9+8eC8ZaSvYyQ=
```
j0sh: `go mod tidy` will remove the older LPMS lines

common/util.go
```diff
@@ -212,6 +228,14 @@ func ProfilesNames(profiles []ffmpeg.VideoProfile) string {
 	return strings.Join(names, ",")
 }
 
+func EncoderProfileNameToValue(profile string) ffmpeg.Profile {
```
j0sh, line 231: Should we make the matching here case insensitive?

j0sh (changes requested):

jailuthra:
PS: For some reason the segmenter test case is failing - I guess it's unrelated to my changes because it fails on my local `master` build as well.

server/mediaserver_test.go
```diff
@@ -432,9 +432,9 @@ func TestCreateRTMPStreamHandlerWebhook(t *testing.T) {
 
 	// set profiles with valid values, presets empty
 	ts7 := makeServer(`{"manifestID":"a", "profiles": [
-		{"name": "prof1", "bitrate": 432, "fps": 560, "width": 123, "height": 456},
+		{"name": "prof1", "bitrate": 432, "fps": 560, "width": 123, "height": 456, "profile": "H264Baseline"},
 		{"name": "prof2", "bitrate": 765, "fps": 876, "fpsDen": 12, "width": 456, "height": 987},
-		{"name": "passthru_fps", "bitrate": 890, "width": 789, "height": 654}]}`)
+		{"name": "passthru_fps", "bitrate": 890, "width": 789, "height": 654, "profile": "H264ConstrainedHigh"}]}`)
```
j0sh, line 437: TeSt FoR mIxEd CaSe?

j0sh: Actually it already *is* mixed case....!!

j0sh (approved):
LGTM 🚢 

Added some back-compat regression tests here https://gist.github.com/j0sh/9d884a19bb92e504092e71fd19d812e5#file-profiles-standalone-sh and https://gist.github.com/j0sh/9d884a19bb92e504092e71fd19d812e5#file-profiles-sh , things seem to work nicely.

PR #1544, opened 2020-06-23
server: fix TestSubmitSegment_HttpPostError test

kyriediculous:
Fixes a failing test that checks for a "connection refused" error on foo.com but the server for that domain doesn't actually have http2 enabled so the test fails. Sets it to post to `https://localhost/segment` instead.


ya7ya (approved):
LGTM

PR #1547, opened 2020-06-25
cmd: allow orchestrator+redeemer mode 

kyriediculous:
Allows to run a bockchain service node : 
- redeemer
- round initializer
- reward service

without running a transcoder server. 

e.g. `livepeer  -redeemer -intializeRound`


yondonfu (changes requested):
The current changes in this PR result in the following:

- Either the `-transcoder` flag or the `-orchSecret` flag also need to be specified if the `-orchestrator` and `-redeemer` flags are provided
- The node will still be considered as running in orchestrator mode because the `-orchestrator` flag has the highest precedence when setting the [node type](https://github.com/livepeer/go-livepeer/blob/87cab20a1671555088429d10975879e02968d4e2/cmd/livepeer/livepeer.go#L261)

If the node is going to support running with both the `-orchestrator` and `-redeemer` flags so that the node can run a reward service and a ticket redemption service with the on-chain registered orchestrator address then I think the following updates should be made:

- Allow the node to be started with the `-orchestrator` and `-redeemer` flags *without* the `-transcoder` flag or the `-orchSecret` flag
- Set the node type to be `NodeType.Redeemer` if the `-redeemer` flag is specified

I think both of the updates mentioned above can be accomplished by giving the `-redeemer` flag the highest precedence when setting the node type i.e.

```golang
if *redeemer {
    n.NodeType = core.RedeemerNode
} else if *orchestrator {
    ...
} ...
```

This change should also avoid starting an orchestrator server when the `-redeemer` flag is specified without any additional checks because the orchestrator server is only started if the node type is `NodeType.Orchestrator`.

j0sh:
Is there a reason this functionality can't be bundled under `-redeemer` in its entirety? To me, the orchestrator node plays a specific role: it facilitates transcoding, whether local or remote. If the goal is to not transcode, then why does the `-orchestrator` flag also need to be set?

This is from a UX perspective - I acknowledge that the current structure of the code may make it easier to just use this particular combination of flags, but "this is how the code is" really shouldn't dictate the final UX.

> the node can run a reward service and a ticket redemption service with the on-chain registered orchestrator address

If we take the word "orchestrator" out of this sentence, the important part is "on-chain registered address" - can the redeemer itself run with the on-chain address?

kyriediculous:
> I acknowledge that the current structure of the code may make it easier to just use this particular combination of flags, but "this is how the code is" really shouldn't dictate the final UX.

Yeah you're right.

I adjusted the code with @yondonfu suggestion and I also extracted the rewardService and roundInitialiser outside of the `orchestrator` block. 

These last two services will now start when the node address is the `recipientAddr` (on-chain registered)  regardless of Orchestrator or Redeemer mode.

cmd/livepeer/livepeer.go
```diff
@@ -816,6 +817,10 @@ func main() {
 			return
 		}
 
+		if *redeemer {
```
yondonfu: We don't need this check because if the `-redeemer` flag is used the node type will be `NodeType.RedeemerNode` and the check on L816 will evaluate to false.

cmd/livepeer/livepeer.go
```diff
@@ -553,23 +554,23 @@ func main() {
...
-				}
-
-				// Create reward service to claim/distribute inflationary rewards every round
-				rs := eventservices.NewRewardService(n.Eth, blockPollingTime)
-				rs.Start(ctx)
-				defer rs.Stop()
+		// Only initialize rounds and call reward if the node is using an on-chain registered orchestrator address
+		// Assume that the node is using an on-chain registered orchestrator address if -ethOrchAddr is not specified or
+		// if the ticket recipient is the node's address
+		if *ethOrchAddr == "" || n.Eth.Account().Address == recipientAddr {
```
yondonfu: Since this conditional is outside of the `if *orchestrator { ... }` block, the code below will be executed when a node runs with `-broadcaster`.

This could be addressed with an updated conditional such as the following:

```golang
if (*orchestrator && *ethOrchAddr == "") || (n.Eth.Account().Address == *ethOrchAddr && *ethOrchAddr != "") { ... }
```

The first part of the conditional preserves existing reward service behavior for orchestrators - if the orchestrator is run without `-ethOrchAddr` (meaning it is not a part of a multi-O setup) then the reward service is started.

The second part of the conditional ensures that following code is only executed if the `-ethOrchAddr` is specified and the address is equal to the node's address (the problem with the current conditional is that `n.Eth.Account().Address` will be equal to `recipientAddr` whenever `-ethOrchAddr` is not specified which is the case when a user runs a broadcaster).

yondonfu (changes requested):

cmd/livepeer/livepeer.go
yondonfu: Actually I realized a more straightforward way to handle this could be the following:

```golang
// If the node address is an on-chain registered address, start the reward service
t, err := n.Eth.GetTranscoder(n.Eth.Account().Address)
if err != nil {
    ...
}
if t.Status == "Registered" {
    // Start reward service
}

// The node will only initialize rounds if it in the upcoming active set for the round
if *initializeRound {
    // Start round initializer
}
```

These changes would result in the following:
- If `livepeer` is run with an on-chain registered address, the reward service is started (but the node will only call reward if it is active in the current round).
- If `livepeer` is run with `-initializeRound`, the round initializer is started (but the node will only initialize rounds if it is in the upcoming active set for the round).

yondonfu:
Hm so right now the reward service can only be turned on if the node type is set (i.e. via `-broadcaster`, `-redeemer`, `-transcoder` or `-orchestrator`). Since the multi-O docs in #1548 describe a setup where one node is the redeemer and another runs the reward service, could it make sense to allow the reward service to be turned on even if the node type is not explicitly set and use a "default" node type? This would avoid requiring additional node type flags if all you want to do is run the reward service and/or other services like the round initializer.

I think something like the following might work:

```diff
diff --git a/cmd/livepeer/livepeer.go b/cmd/livepeer/livepeer.go
index 5408963b..99d9f918 100644
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -271,7 +271,7 @@ func main() {
        } else if *redeemer {
                n.NodeType = core.RedeemerNode
        } else {
-               glog.Fatalf("Node type not set; must be one of -broadcaster, -transcoder or -orchestrator")
+               glog.Info("Node type not set i.e -broadcaster, -redeemer, -transcoder or -orchestrator")
        }

        if *monitor {
@@ -289,6 +289,8 @@ func main() {
                        nodeType = "trcr"
                case core.RedeemerNode:
                        nodeType = "rdmr"
+               default:
+                       nodeType = "dflt"
                }
                lpmon.InitCensus(nodeType, nodeID, core.LivepeerVersion)
        }
diff --git a/core/livepeernode.go b/core/livepeernode.go
index e907b278..2756388e 100644
--- a/core/livepeernode.go
+++ b/core/livepeernode.go
@@ -38,7 +38,8 @@ var MaxSessions = 10
 type NodeType int

 const (
-       BroadcasterNode NodeType = iota
+       DefaultNode NodeType = iota
+       BroadcasterNode
        OrchestratorNode
        TranscoderNode
        RedeemerNode
```

kyriediculous:
> could it make sense to allow the reward service to be turned on even if the node type is not explicitly set and use a "default" node type?

Yeah I think it does make sense to support that.

j0sh:
FWIW, with this and a couple other changes in the pipeline (eg, https://github.com/livepeer/go-livepeer/compare/ja/pull ) then the distinction between the "node types" is becoming increasingly vague.

At some point we may to omit the notion of a "node type" entirely and simply check whether a certain feature is enabled. This allows for a mix of features to be running, rather than pigeonholing behaviors within a specific "node type".

cmd/livepeer/livepeer.go
```diff
@@ -268,10 +270,8 @@ func main() {
 		n.NodeType = core.TranscoderNode
 	} else if *broadcaster {
 		n.NodeType = core.BroadcasterNode
-	} else if *redeemer {
-		n.NodeType = core.RedeemerNode
 	} else {
-		glog.Fatalf("Node type not set; must be one of -broadcaster, -transcoder or -orchestrator")
+		glog.Info("Node type not set; must be one of -broadcaster, -transcoder, -orchestrator or -redeemer")
```
yondonfu: Let's remove the "must be one of" part because that is no longer accurate.

An example updated message could be:

"Node type not set with -broadcaster, -transcoder, -orchestrator or -redeemer"

yondonfu (changes requested):

kyriediculous:
fixed in 57b8f790

yondonfu (commented):
Changes look good let's rebase

kyriediculous:
rebased

yondonfu (approved):
LGTM pending green CI

cmd/livepeer/livepeer.go
```diff
@@ -268,10 +270,8 @@ func main() {
 		n.NodeType = core.TranscoderNode
 	} else if *broadcaster {
 		n.NodeType = core.BroadcasterNode
-	} else if *redeemer {
-		n.NodeType = core.RedeemerNode
 	} else {
-		glog.Fatalf("Node type not set; must be one of -broadcaster, -transcoder or -orchestrator")
+		glog.Info("Node type not set with -broadcaster, -transcoder, -orchestrator or -redeemer")
```
j0sh, line 274: It does seem like allowing a "default" node type that basically does nothing is a bit of a usability regression. I understand that we didn't want to enumerate a bunch of NodeTypes (and I think we probably should rework that mechanism entirely), but in the past, we've failed fast if the user didn't specify the mode they were in.

Right now, we don't fail fast - we just sit there with services disabled - and I think that's going to confuse some people. This log message might help a bit, but it's not immediately obvious, because it's sandwiched in between a bunch of other messages:

```
/Users/josh/livepeer/code/go-livepeer/livepeer \
    -datadir /Users/josh/livepeer/run/datadir/broadcaster \
    -currentManifest -orchAddr :8936 \
    -rtmpAddr 0.0.0.0 \
    -transcodingOptions datadir/broadcaster/profiles.json \
    -v 99
I0702 12:41:17.288835    6911 livepeer.go:215] ***Livepeer is running on the offchain network***
I0702 12:41:17.290573    6911 db.go:376] Initialized DB node
I0702 12:41:17.290629    6911 livepeer.go:279] Node type not set with -broadcaster, -transcoder, -orchestrator or -redeemer
I0702 12:41:17.290662    6911 livepeer.go:320] ***Livepeer is in off-chain mode***
I0702 12:41:17.291110    6911 livepeer.go:871] Current ManifestID will be available over 
I0702 12:41:17.291175    6911 mediaserver.go:137] Transcode Job Type: [{144p 100000 0 0 256x144  0 0 0s} {240p 150000 0 0 320x240  0 0 0s}]
I0702 12:41:17.291239    6911 webserver.go:70] CLI server listening on 127.0.0.1:7935
```

PR #1548, opened 2020-06-26
doc: multi-o setup instructions

kyriediculous:
Provide instructions for various ways to start a multi-o setup with a redeemer.

### DIFF
--- a/doc/multi-o.md
+++ b/doc/multi-o.md
@@ -1,0 +1,115 @@
+# Multi Orchestrator Setup
+
+This document describes the various ways an operator could run multiple Orchestrator nodes. Each Orchestrator node will still have its own keypair and will still accept payments on behalf of the on-chain registered Ethereum address. This setup allows an operator to separate node operations; e.g. one node responsible for redeeming winning tickets, one node for calling `reward` and multiple nodes solely responsible for handling transcode requests. 
+
+It is also possible to connect multiple Transcoder nodes to an Orchestrator node. The documentation about scaling transcoding by splitting Orchestrator and Transcoder nodes can be found [here](https://livepeer.readthedocs.io/en/latest/transcoding.html#scaling-transcoding).
+
+## Unlocking Account and Connecting to Ethereum
+
+The Livepeer node requires you to unlock your Ethereum account and provide connection information for an Ethereum JSON-RPC provider in order to interact with the [Livepeer Protocol](https://github.com/livepeer/protocol), unless the node is in [standalone transcoder]((https://livepeer.readthedocs.io/en/latest/transcoding.html#scaling-transcoding)) mode.
+
+For brevity we'll exclude these flags in the examples and replace them with `<...ETH SETUP...>`. If no Ethereum account is available the Livepeer node will create one for you.
+
+More detailed instructions for connecting to Ethereum can be found [here](https://livepeer.readthedocs.io/en/latest/quickstart.html#connecting-to-an-ethereum-node). 
+
+```
+livepeer \
+    -network <mainnet|rinkeby> \
+    -ethUrl <JSON_RPC_PROVIDER> \
+    -ethKeystorePath <PATH_TO_KEYSTORE_DIR> \
+    -ethAcctAddr <ETHEREUM_ADDRESS> \
+    -ethPassword <ETHEREUM_ACCOUNT_PASSWORD (string|path)> \
+    -ethController <CONTROLLER_CONTRACT_ADDRESS (required if '-network' not provided)>
+```
+
+## Single Orchestrator with Redeemer
+
+This setup allows an operator to use a seperate Ethereum account for redeeming winning tickets on-chain and paying for that transaction while the ticket recipient is still the operator's on-chain registered Ethereum address.
+
+The Orchestrator node will still be responsible for calling `reward`*.
+
+\* _currently, only the on-chain registered address can call `reward`_
+
+1. Start the Redeemer 
+
+```shell
+livepeer \
+    <...ETH SETUP...> \
+    -redeemer \ 
+    -ethOrchAddr <ORCHESTRATOR_ON_CHAIN_ETH_ADDR (also the recipient address)>
+```
+
+2. Start the Orchestrator node
+
+
+```shell
+livepeer \
+    <...ETH SETUP...> \
+    -orchestrator -transcoder \
+    -redeemerAddr <REDEEMER_HTTP_ADDR> \
+    -pricePerUnit <PRICE (wei/pixel if '-pixelsPerUnit' is not set)>
+```
+
+
+## Blockchain Service Node with (Multiple) Orchestrator node(s)
+
+In this setup a node started with the keys for the on-chain registered address will be responsible for all transactions. 
+
+In order to use multiple Orchestrator nodes with this setup, the cluster of Orchestrator nodes responsible for transcoding would have to be behind a load balancer (e.g. DNS load balancing). The URI of the load balancer will be the on-chain registered Service URI.
+
+1. Start the Blockchain Service Node 
+
+```shell
+livepeer \
+    <...ETH SETUP...> \
+    -redeemer \ 
+    -httpAddr <REDEEMER_HTTP_ADDR (host):port> \
+```
+
+2. Start an Orchestrator node
+```shell
+livepeer \
+    <...ETH SETUP...> \
+    -orchestrator -transcoder \
+    -ethOrchAddr <ORCHESTRATOR_ON_CHAIN_ETH_ADDR> \
+    -redeemerAddr <REDEEMER_HTTP_ADDR> \
+    -pricePerUnit <PRICE (wei/pixel if '-pixelsPerUnit' is not set)>
+```
+
+## Redeemer + RewardService + Multiple Orchestrator Nodes
+
+This is the "coldest" setup possible for on-chain registered addresses. The keys for the on-chain registered address will be responsible for calling reward.
+
+To use this setup with multiple Orchestrator nodes a load balancer is required as described above. 
+
+1. Start the Redeemer
+
+```shell
+livepeer \
+    <...ETH SETUP...> \
+    -redeemer \ 
+    -httpAddr <REDEEMER_HTTP_ADDR (host):port> \ 
+    -ethOrchAddr <ORCHESTRATOR_ON_CHAIN_ETH_ADDR>
+```
+
+2. Start the RewardService
+
+The address specified via `-ethAcctAddr` should be the on-chain registered address. 
+
+```shell
+livepeer \
+    <...ETH SETUP...> \
+```
+
+3. Start an Orchestrator  node
+
+The main difference is to use the `-ethOrchAddr` flag and specify the on-chain registered Ethereum address, in this case that of the RewardService node. 
+
+```shell
+livepeer \
+    <...ETH SETUP...> \
+    -orchestrator -transcoder \
+    -ethOrchAddr <ORCHESTRATOR_ON_CHAIN_ETH_ADDR (also the recipient address)> \
+    -redeemerAddr <REDEEMER_HTTP_ADDR> \
+    -pricePerUnit <PRICE (wei/pixel if '-pixelsPerUnit' is not set)>
+```


doc/multi-o.md
```diff
@@ -0,0 +1,117 @@
...
+This document describes the various ways an orchestrator could run multiple nodes and separate concerns. 
+
+## Unlocking Account and Connecting to Ethereum
+
+Unlocking your Ethereum account and connecting to a JSON-RPC provider are necessary steps to run the Livepeer node unless it's in standalone transcoder mode. 
+
+For brevity we'll exclude these flags in the examples by `<...ETH SETUP...>` but these should always be required. If no Ethereum account is available the Livepeer node will create one for you.
+
+```
+livepeer
```
yondonfu: Let's add a backslash after `livepeer` in all of the commands in this doc since it would be required for a valid multi-line bash command.

doc/multi-o.md
```diff
@@ -0,0 +1,117 @@
...
+In this setup a node started with the keys for the on-chain registered address will be responsible for all transactions. 
+
+In order to use multiple orchestrators with this setup, the cluster of orchestrator nodes responsible for transcoding would have to be behind a load balancer, **the URI of the load balancer will be the on-chain registered Service URI.**
+
+1. Start the Blockchain Service Node 
+
+```shell
+livepeer
+    <...ETH SETUP...\
+    -orchestrator= true \
```
yondonfu: Extra space before true can be removed.

doc/multi-o.md
```diff
@@ -0,0 +1,117 @@
...
+
+Unlocking your Ethereum account and connecting to a JSON-RPC provider are necessary steps to run the Livepeer node unless it's in standalone transcoder mode. 
+
+For brevity we'll exclude these flags in the examples by `<...ETH SETUP...>` but these should always be required. If no Ethereum account is available the Livepeer node will create one for you.
+
+```
+livepeer
+    -network <mainnet|rinkeby> \
+    -ethUrl <JSON_RPC_PROVIDER> \
+    -ethKeystorePath <PATH_TO_KEY> \
```
yondonfu: `-ethKeystorePath` points to a directory and not to the keystore file itself so this should be updated to something like `<PATH_TO_KEYSTORE_DIR>`.

doc/multi-o.md
```diff
@@ -0,0 +1,117 @@
...
+    -ethAcctAddr <ETHEREUM_ADDRESS> \
+    -ethPassword <ETHEREUM_ACCOUNT_PASSWORD (string|path)> \
+    -ethController <CONTROLLER_CONTRACT_ADDRESS (required if '-network' not provided)>
+```
+
+## Single Orchestrator with Redeemer
+
+This allows an orchestrator running a single on-chain node to use a seperate Ethereum account to redeem winning tickets on-chain and pay for that transaction while the recipient is still orchestrator node's Ethereum address.
```
yondonfu: Suggestion to clarify that the recipient address is the *on-chain registered address*:

"recipient is still orchestrator node's Ethereum address" -> "the ticket recipient is still the on-chain registered Ethereum address".

doc/multi-o.md
```diff
@@ -0,0 +1,117 @@
+# Multi Orchestrator Setup
+
+This document describes the various ways an orchestrator could run multiple nodes and separate concerns.
```
yondonfu: I think it could be worth clarifying that the nodes would still accept payments on behalf of the same on-chain registered address while each node uses its own address.

yondonfu: For terminology, I suggest we use "orchestrator operator" to refer to the entity running orchestrator nodes. So, this sentence would be updated to: "an orchestrator operator could run multiple nodes...".

doc/multi-o.md
```diff
@@ -0,0 +1,117 @@
...
+```
+
+2. Start the Orchestrator
+
+
+```shell
+livepeer
+    <...ETH SETUP...> \
+    -orchestrator=true -transcoder=true \
+    -initializeRound=true \
```
yondonfu: Orchestrators aren't required to initialize rounds right now which is why the round initializer service is off by default and we don't communicate that orchestrators should be running with this flag right now. What do you think about excluding the round initialization bit from these docs right now and only including it when we're communicating that the flag should be used by orchestrators regardless of what kind of setup they are using?

doc/multi-o.md
```diff
@@ -0,0 +1,117 @@
...
+
+The orchestrator node will still be responsible for initialising rounds and calling `reward`. 
+
+1. Start the Redeemer 
+
+```shell
+livepeer
+    <...ETH SETUP...\
+    -redeemer=true \ 
+    -httpAddr <REDEEMER_HTTP_ADDR (host):port \
```
yondonfu: Missing > at the end. Same comment applies for a few of the other flag values in this doc so it'd be good to take a pass through all them to add the missing > at the end.

doc/multi-o.md
```diff
@@ -0,0 +1,117 @@
...
+```
+
+3. Start an Orchestrator 
+
+The main difference is to use the `-ethOrchAddr` flag and specify the Ethereum address of the RewardService & RoundInitializer node. 
+
+```shell
+livepeer
+    <...ETH SETUP...> \
+    -orchestrator=true \
```
yondonfu: Needs `-transcoder`.

doc/multi-o.md
```diff
@@ -0,0 +1,117 @@
...
+
+2. Start the Orchestrator
+
+
+```shell
+livepeer
+    <...ETH SETUP...> \
+    -orchestrator=true -transcoder=true \
+    -initializeRound=true \
+    -httpAddr <ORCH_HTTP_ADDR (host):port \
```
yondonfu: Might be ok to omit this as the default should be ok.

yondonfu (changes requested):

doc/multi-o.md
j0sh: Could be specific about the node type, eg "an operator can run multiple orchestrator nodes."

I'd also elaborate a bit on the "separate concerns", probably in a follow-on sentence. What concerns are those? Wallet and key management? Blockchain monitoring? Submitting transactions?

doc/multi-o.md
```diff
@@ -0,0 +1,117 @@
+# Multi Orchestrator Setup
+
+This document describes the various ways an orchestrator could run multiple nodes and separate concerns. 
+
+## Unlocking Account and Connecting to Ethereum
+
+Unlocking your Ethereum account and connecting to a JSON-RPC provider are necessary steps to run the Livepeer node unless it's in standalone transcoder mode.
```
j0sh: Maybe use the active voice? "The Livepeer node needs you to unlock your Ethereum account and provide connection information for a JSON-RPC provider, unless the node is in standalone transcoder mode." [1]

Do we have documentation on what connecting to a JSON-RPC provider entails? Maybe link to that, or add a quick blurb if none exists? Also note the slightly modified language here, since the node is the one connecting to the JSON-RPC provider, not the user; the user just needs to provide the connection information.

Also, maybe briefly describe "standalone transcoder" mode, or link if there's a corresponding doc.

j0sh: Also - this section is seems more broadly applicable to all on-chain use cases, rather than multi-O specifically. Probably don't need to refactor the docs right now, but just noting that this part could be extracted into more general-purpose onchain documentation someday.

doc/multi-o.md
```diff
@@ -0,0 +1,117 @@
...
+    -ethOrchAddr <ORCHESTRATOR_ON_CHAIN_ETH_ADDR (also the recipient address)>
+```
+
+2. Start the Orchestrator
+
+
+```shell
+livepeer
+    <...ETH SETUP...> \
+    -orchestrator=true -transcoder=true \
```
j0sh: The `=true` bits are not necessary, and can be omitted here and in all the following examples. Invoking bool flags automatically sets their value to true.

doc/multi-o.md
```diff
@@ -0,0 +1,117 @@
...
+    -ethOrchAddr <ORCHESTRATOR_ON_CHAIN_ETH_ADDR (also the recipient address)> \
+    -redeemerAddr <REDEEMER_HTTP_ADDR> \
+    -httpAddr <ORCH_HTTP_ADDR (host):port \
+    -pricePerUnit <PRICE (wei/pixel if '-pixelsPerUnit' is not set)
+```
+
+## Redeemer + RewardService & RoundInitializer + Multiple Orchestrators
+
+This is the "coldest" setup possible for on-chain registered addresses. The keys for the on-chain registered address will be responsible for initialising rounds and calling reward, since reward can not be called on behalf of another address in the current version of the protocol.
```
j0sh: Maybe split up that last sentence a bit. "The keys for the on-chain registered address will be responsible for calling reward. In the current version of the protocol, `reward` can not be called on behalf of another address."

doc/multi-o.md
```diff
@@ -0,0 +1,117 @@
...
+    -httpAddr <ORCH_HTTP_ADDR (host):port \
+    -pricePerUnit <PRICE (wei/pixel if '-pixelsPerUnit' is not set)
+```
+
+
+## Blockchain Service Node with Multiple Orchestrators
+
+In this setup a node started with the keys for the on-chain registered address will be responsible for all transactions. 
+
+In order to use multiple orchestrators with this setup, the cluster of orchestrator nodes responsible for transcoding would have to be behind a load balancer, **the URI of the load balancer will be the on-chain registered Service URI.**
```
j0sh: Maybe elaborate on what forms this load balancer could take, since that's not immediately clear from the description here? FWIW, I think DNS is basically the easiest option right now, other than having the broadcaster take an explicit list via webhook or `-orchAddr`. For those interested in building their own LB, it might be worth linking to the networking documentation around GetOrchestrator/OrchestratorInfo (it's unlikely anyone will try to at this stage, but maybe just mention it for completeness).

Also, I think the bolded parts can be split into its own sentence.

doc/multi-o.md
```diff
@@ -0,0 +1,117 @@
...
+livepeer
+    <...ETH SETUP...\
+    -redeemer=true \ 
+    -httpAddr <REDEEMER_HTTP_ADDR (host):port \ 
+    -ethOrchAddr <ORCHESTRATOR_ON_CHAIN_ETH_ADDR (also the recipient address)>
+```
+
+2. Start the RewardService & RoundInitializer 
+
+In the current go-livepeer version (0.5.8) this node type will still run a transcoder server. To avoid this node also transcoding avoid making it accessible through the load balancer of your choice.
```
j0sh: This isn't really a documentation issue per se, but just noting that reading through this makes me realize that we really need to loosen up restriction on who can call `reward`, because that would simplify much of the guidance here.

> To avoid this node also transcoding avoid making it accessible through the load balancer of your choice.

An easier way may be to set the ServiceURI to localhost and ensure the box itself is not available to the public?

"To prevent this node from transcoding, set the service URI to localhost and ensure the system is not behind a publicly available IP: `-serviceAddr 127.0.0.1:8935`"

(BTW I just checked this and it seems that we listen to all ipv6 interfaces regardless of what the ipv4 address is, so that needs to be corrected at some point, but should be a non-issue as long as the system itself isn't open to the public, eg NAT'd.)

kyriediculous: >  we really need to loosen up restriction on who can call reward

Yeah we briefly discussed this on discord yesterday. This would be a requirement indeed before we can have complete cold storage for on-chain registered addresses. 

> An easier way may be to set the ServiceURI to localhost and ensure the box itself is not available to the public?

This only overwrites the serviceURI locally and not on-chain, right?

doc/multi-o.md
kyriediculous: I decided to not include any specifics about that here due to the fact that anything other than simple load balancing between a cluster of nodes within reasonable proximity is still an open area of investigation. We currently only use an explicit list on our broadcasters, which is only an option in a BOT setup. 

I think we can go more into detail once we have a better sense of the specifics here and what would work well for livepeer and what wouldn't ?

j0sh: Yeah, that's reasonable. Should we still mention "load balancing" even though we don't necessarily have concrete guidance around that? Seems like that would invite more questions than it answers. Maybe something like,

"In order to select multiple orchestrator nodes from the same operator, the broadcaster needs to explicitly list each orchestrator, either via the `-orchAddr` or`-orchWebhookUrl` flags. Additional load balancing options are being developed."

DNS should also work reasonably well out-of-the-box, although I don't know if anyone has actually tried it yet. Should we mention that as well?

doc/multi-o.md
j0sh: > This only overwrites the serviceURI locally and not on-chain, right?

Correct. The user should also ensure the on-chain address points elsewhere, and that the current host not publicly available.

doc/multi-o.md
kyriediculous: I think we should avoid a scenario whereby people feel comfortable omitting this flag, so how about the reverse ? 

Keep it in but mention that livepeer currently 'sponsors' round initialisation but that ideally node operators shouldn't rely on single points of failures like that and make sure they can do everything within their means to run robust infra.

doc/multi-o.md
kyriediculous: fixed in 4131a02b

doc/multi-o.md
kyriediculous: fixed in 4131a02b

doc/multi-o.md
kyriediculous: fixed in 4131a02b

doc/multi-o.md
kyriediculous: fixed in 4131a02b

doc/multi-o.md
kyriediculous: fixed in 4131a02b

doc/multi-o.md
kyriediculous: fixed in 4131a02b

doc/multi-o.md
kyriediculous: fixed in 4131a02b

doc/multi-o.md
kyriediculous: fixed in 4131a02

doc/multi-o.md
kyriediculous: removed the second part because we mention it earlier as well, fixed in 4131a02

doc/multi-o.md
kyriediculous: How about we refer to the role as "Orchestrator"  and the nodes as "orchestrator nodes" , wouldn't that be simple yet understandable enough?

doc/multi-o.md
kyriediculous: Yeah you might be right I think this is covered elsewhere in the docs plenty.

doc/multi-o.md
kyriediculous: fixed in 4131a02

doc/multi-o.md
```diff
@@ -0,0 +1,123 @@
...
+    -ethUrl <JSON_RPC_PROVIDER> \
+    -ethKeystorePath <PATH_TO_KEYSTORE_DIR> \
+    -ethAcctAddr <ETHEREUM_ADDRESS> \
+    -ethPassword <ETHEREUM_ACCOUNT_PASSWORD (string|path)> \
+    -ethController <CONTROLLER_CONTRACT_ADDRESS (required if '-network' not provided)>
+```
+
+## Single Orchestrator with Redeemer
+
+This setup allows an Orchestrator to use a seperate Ethereum account for redeeming winning tickets on-chain and paying for that transaction while the ticket recipient is still Orcherstrator's on-chain registered Ethereum address.
```
yondonfu: "still Orchestrator's" -> "still the Orchestrator's..."

doc/multi-o.md
yondonfu: Cool then I think we can remove this (here and in other commands in this doc where it isn't necessary as well) in the interest of focusing on the flags absolute required for these instructions.

doc/multi-o.md
```diff
@@ -0,0 +1,123 @@
...
+## Redeemer + RewardService & RoundInitializer + Multiple Orchestrators
+
+This is the "coldest" setup possible for on-chain registered addresses. The keys for the on-chain registered address will be responsible for initialising rounds and calling reward.
+
+To use this setup with multiple orchestrators a load balancer is required as described above. 
+
+1. Start the Redeemer
+
+```shell
+livepeer
```
yondonfu: Missing backslash for this commands and the ones after it.

doc/multi-o.md
```diff
@@ -0,0 +1,123 @@
...
+```
+
+2. Start the RewardService & RoundInitializer 
+
+In the current go-livepeer version (0.5.8) this node type will still run a transcoder server. To avoid incoming transcode requests on this node, don't make the box available on the public internet. 
+
+```shell
+livepeer
+    <...ETH SETUP...> \
+    -orchestrator -trancsoder \
```
yondonfu: With the recent changes in #1547 this command can instead be:

```golang
livepeer \
    <...ETH SETUP...> \
    -initializeRound \
    -ethOrchAddr <ORCHESTRATOR_ON_CHAIN_ETH_ADDR> \
```

We can add a note that the address specified via `-ethAcctAddr` should be the same as the address specified by `-ethOrchAddr`.

doc/multi-o.md
```diff
@@ -0,0 +1,123 @@
...
+To use this setup with multiple orchestrators a load balancer is required as described above. 
+
+1. Start the Redeemer
+
+```shell
+livepeer
+    <...ETH SETUP...> \
+    -redeemer \ 
+    -httpAddr <REDEEMER_HTTP_ADDR (host):port> \ 
+    -ethOrchAddr <ORCHESTRATOR_ON_CHAIN_ETH_ADDR>
```
yondonfu, line 92: With the recent changes in #1547 we can add a note that the address specified via `-ethAcctAddr` should be different than the address specified by `-ethOrchAddr` (or else the reward service will be started with this command).

yondonfu (changes requested):

doc/multi-o.md
yondonfu: "This document describes the various ways an orchestrator could run multiple orchestrator nodes".

vs.

"This document describes the various ways an operator could run multiple orchestrator nodes".

The second option reads a bit smoother to me - I think it makes a clearer distinction between the entity running the nodes and the node type so I have a slight preference for this option.

doc/multi-o.md
yondonfu: I hear you on wanting orchestrator operators to expect to initialize rounds. But, it is worth noting that we currently do not indicate anywhere in our docs that the `-initializeRound` flag should be used. So, it feels a bit weird to drop in this flag in this specific doc which is focused on multi-O setups when the flag is not mentioned anywhere else. I'd be more inclined to mention the `-initializeRound` flag in a more general purpose doc first along with a description of what it does before including it in this doc.

doc/multi-o.md
kyriediculous: While that would work for internal usage or a BOT setup I think it might be good to have these docs also be useful for Orchestrators.  I would avoid mentioning broadcasters here. 

I also feel like the current non-default selection is just an intermediate step that works for us but not necessarily for your average broadcaster node. 

I tried DNS load balancing with NGINX but only for https, it was only a single O but it should work quite the same for multiple O's. 

I added an annotation "e.g. NGINX" but I can do "e.g. DNS Load Balancing" to not be vendor specific if you want.

doc/multi-o.md
```diff
@@ -0,0 +1,119 @@
...
+```shell
+livepeer \
+    <...ETH SETUP...> \
+    -orchestrator -transcoder \
+    -ethOrchAddr <ORCHESTRATOR_ON_CHAIN_ETH_ADDR> \
+    -redeemerAddr <REDEEMER_HTTP_ADDR> \
+    -pricePerUnit <PRICE (wei/pixel if '-pixelsPerUnit' is not set)>
+```
+
+## Redeemer + RewardService  + Multiple Orchestrators
```
yondonfu: Let's remove the extra space after "RewardService".

doc/multi-o.md
kyriediculous: I don't think this is possible because on must always supply a node type, orchestrator or redeemer in this case. Even with the changes in #1547

doc/multi-o.md
kyriediculous: We already have a note that states the address should be that of the rewardservice, but i'll elaborate on it.

doc/multi-o.md
kyriediculous: fixed in 86cebba

doc/multi-o.md
kyriediculous: fixed in 86cebba

doc/multi-o.md
kyriediculous: fixed in daacf97

doc/multi-o.md
kyriediculous: replaced with "e.g DNS load balancing"

doc/multi-o.md
kyriediculous: This should no longer be necessary with the latest changes in #1547

doc/multi-o.md
```diff
@@ -0,0 +1,115 @@
...
+```shell
+livepeer \
+    <...ETH SETUP...> \
+    -orchestrator -transcoder \
+    -redeemerAddr <REDEEMER_HTTP_ADDR> \
+    -pricePerUnit <PRICE (wei/pixel if '-pixelsPerUnit' is not set)>
+```
+
+
+## Blockchain Service Node with (Multiple) Orchestrator(s) node(s)
```
yondonfu: The (s) after Orchestrator can be removed if it is followed by "node(s)".

yondonfu (commented):
Just one comment - feel free to address it and rebase

kyriediculous:
addressed the comment and rebased

yondonfu (approved):
LGTM

PR #1551, opened 2020-06-26
Set GOP

j0sh:

This enables the user to configure the output GOP via webhook.

The changes here are fairly mechanical; more attention is probably needed for https://github.com/livepeer/lpms/pull/198 .

- Updates LPMS to https://github.com/livepeer/lpms/pull/198 . Will finalize the git revision here prior to merging.
- Adds a `gop` field of type `int32` to net.VideoProfile, encoded as milliseconds - same as the `SegData.Duration` protobuf field.
- Adds a `gop` field of type string to the webhook
- Conversion code to / from the two

This also incorporates a GOP mode to select intra-only frames, signaled via a special negative value (`ffmpeg.GOPIntraOnly`). This was done to lessen the speculative aspect of a few choices:

* Gives a concrete use case a `string` value in the webhook - "intra" is accepted
* Makes use of the signed integer type for the protobuf field - otherwise an unsigned might be justified.
* Used within back-compat regression tests.

* Unit tests
* Manual tests
* Added GOP regression tests in https://gist.github.com/j0sh/9d884a19bb92e504092e71fd19d812e5#file-run-sh

**Checklist:**


jailuthra (approved):
LGTM!

j0sh:
Updated to latest LPMS, and added explicit signaling for intra mode. More information in the PR description.

PR #1558, opened 2020-07-03
Use JSON config file to specify renditions

### LINKED ISSUE #1271: B should refuse to start if invalid value provided via `-transcodingOptions`
**Describe the bug**
If invalid value provided throw `-transcodingOptions` B silently ignores that and successfully starts with empty list of profiles to transcode.

**To Reproduce**
1. Run B providing random string in `-transcodingOptions`
2. B starts with zero profiles to transcode

**Expected behaviour**
Should fail to start with error message.

j0sh:
**TODO**

- [x] Documentation updates.
- [x] Manual testing with existing workflows - have only tested with the pull prototype
- [x] Limited set of integration tests [here](https://gist.github.com/j0sh/9d884a19bb92e504092e71fd19d812e5#file-run-sh)

Currently, only pre-defined presets can be selected via CLI flags. These presets expose a limited subset of the configuration available to the user. In order to to access all the configuration options, the user needs to run a separate webhook service. While webhooks are extremely useful in a distributed, scaled environment, they are unwieldy for local  use of the node.

This PR enables the ability to configure the output presets via JSON file. The existing `-transcodingOptions` flag is overloaded as follows:

* If the `-transcodingOptions` flag is not set, nothing happens (default presets are used). Matches current behavior. 
* Attempts to read a file based on the value of `-transcodingOptions`.
* If reading the file succeeds, parses and processes the JSON.
* If there is an error reading the file, attempts to parse the preset string. Matches current behavior.
* If no profile is successfully read, fails. This is new behavior.

The existing JSON definition from the webhook is reused. This was the fastest way to get things working. The only thing worse than JSON for configuration is YAML. 🙈  We can always add better format support later.

- Moves transcodingOptions processing to `NewLivepeerServer` from `StartMediaServer`. This is a bit neater and makes later work easier ([pull mode](https://github.com/livepeer/go-livepeer/compare/ja/pull) does not start the media server but still needs to initialize a livepeer server)
- Moves JSON-to-VideoProfile parsing into its own helper function.
- Invokes the helper function within `NewLivepeerServer`, in addition to the webhook handler.
- Tightens up error handling around invalid encoding profiles
- Adds test coverage
- Updates documentation, including the addition of a page on transcoding configuration

Tons of unit tests. Most of the changes here are unit tests.

Manually tested as part of the pull mode work.

Regression tests. This also goes through earlier tests which exercises webhook components, which ensures the existing functionality remains unchanged. https://gist.github.com/j0sh/9d884a19bb92e504092e71fd19d812e5/revisions#diff-1b0c2b516b83393edb7200ad5ff12181

Fixes https://github.com/livepeer/go-livepeer/issues/1271

**Checklist:**

### DIFF
--- a/common/util_test.go
+++ b/common/util_test.go
@@ -201,6 +201,24 @@ func TestVideoProfile_FormatExtension(t *testing.T) {
 		t.Error("Did not get expected error")
 	}
 }
+
+func TestVideoProfile_ProfileNameToValue(t *testing.T) {
+	assert := assert.New(t)
+	inp := []string{"", "h264baseline", "h264main", "h264high", "h264constrainedhigh"}
+	assert.Len(inp, len(ffmpeg.ProfileParameters), "Missing a new profile?")
+	outp := []ffmpeg.Profile{ffmpeg.ProfileNone, ffmpeg.ProfileH264Baseline, ffmpeg.ProfileH264Main, ffmpeg.ProfileH264High, ffmpeg.ProfileH264ConstrainedHigh}
+	assert.Len(inp, len(outp))
+	for i, profile := range inp {
+		p, err := EncoderProfileNameToValue(profile)
+		assert.Nil(err)
+		assert.Equal(outp[i], p)
+	}
+	p, _ := EncoderProfileNameToValue("none")
+	assert.Equal(ffmpeg.ProfileNone, p)
+	_, err := EncoderProfileNameToValue("invalid")
+	assert.Equal(ErrProfName, err, "Could not get profile value")
+}
+
 func TestPriceToFixed(t *testing.T) {
 	assert := assert.New(t)
 
--- a/doc/rtmpwebhookauth.md
+++ b/doc/rtmpwebhookauth.md
@@ -1,6 +1,8 @@
-# RTMP Webhook Authentication
+# Webhook Authentication
 
-Incoming RTMP streams can be authenticated using Webhooks. To use these webhooks, node operators must implement their own web service / endpoint to be accessed only by the Livepeer node. As new RTMP streams appear, the Livepeer node will call this endpoint to determine whether the given stream is allowed.
+Incoming streams can be authenticated using webhooks. To use these webhooks, node operators must implement their own web service / endpoint to be accessed only by the Livepeer node. As new streams appear, the Livepeer node will call this endpoint to determine whether the given stream is allowed.
+
+Webhooks can authenticate streams supported by the RTMP and HTTP push ingest protocols. See the [ingest documentation](ingest.md) for details on how to use these protocols.
 
 To enable webhook authentication functionality, the Livepeer node should be started with the `-authWebhookUrl` flag, along with the webhook endpoint URL.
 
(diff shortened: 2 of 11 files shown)


j0sh:
Ready for review.

common/util_test.go
```diff
@@ -201,6 +201,22 @@ func TestVideoProfile_FormatExtension(t *testing.T) {
 		t.Error("Did not get expected error")
 	}
 }
+
+func TestVideoProfile_ProfileNameToValue(t *testing.T) {
+	assert := assert.New(t)
+	inp := []string{"", "h264baseline", "h264main", "h264high", "h264constrainedhigh"}
```
yondonfu, line 207: Missing "none" which is included in the `EncoderProfileLookup` map in `EncoderProfileNameToValue()`?

doc/rtmpwebhookauth.md
```diff
@@ -10,9 +12,9 @@ For example:
 livepeer -authWebhookUrl http://ownserver/auth
 ```
 
-For each incoming RTMP stream, the Livepeer node will make a `POST` request to the `http://ownserver/auth` endpoint, passing the URL of the RTMP request as JSON object.
+For each incoming stream, the Livepeer node will make a `POST` request to the `http://ownserver/auth` endpoint, passing the URL of the RTMP request as JSON object.
```
yondonfu: "passing the URL of the RTMP request" -> "passing the URL of the request"?

doc/transcodingoptions.md
```diff
@@ -0,0 +1,157 @@
...
+
+ If the aspect ratio of the source video does not match the target aspect ratio, the output is fit to the larger dimension, and the smaller dimension is rescaled proportionately to maintain the source aspect ratio.
+
+If a different behavior is needed, please [let us know](https://github.com/livepeer/go-livepeer/issues/new?template=feature_request.md) by filing a feature request.
+
+### Webhook Authentication
+
+See the [webhook documentation](rtmpwebhookauth.md) for details. To configure the transcoding output, either the `profiles` or `presets` fields in the webhook response can be set, or both.
+
+.se  The `profiles` field in the webhook follows the JSON options that are described in this document, The valid `preset` names follows as described.
```
yondonfu: Extraneous ".se"?

server/mediaserver.go
```diff
@@ -312,6 +304,49 @@ func authenticateStream(url string) (*authWebhookResponse, error) {
...
+				profile.Height,
+				profile.Bitrate)
+		}
+		var gop time.Duration
+		if profile.GOP != "" {
+			if profile.GOP == "intra" {
+				gop = ffmpeg.GOPIntraOnly
+			} else {
+				gopFloat, err := strconv.ParseFloat(profile.GOP, 64)
+				if err != nil || gopFloat <= 0.0 {
```
yondonfu: Looks like we can avoid the nested error check here with something like:

```
if err != nil {
    return nil, err
}

if gopFloat <= 0.0 {
    return nil, errors.New("invalid gop value")
}
```

yondonfu (changes requested):
Docs look great! Just a few small comments.

darkdarkdragon:
Agreed with Yondon's comments, otherwise looks good.

Small note - this one fixes #1271 😄

server/mediaserver.go
j0sh: Fixed in 4b17b6ca9b4e86f1e4c5d5648be235e0580fafc6

common/util_test.go
j0sh: Fixed by adding a separate check in 4b17b6ca9b4e86f1e4c5d5648be235e0580fafc6 . Note that we can't add a new element to that list, since it matches exactly the number of profiles we have (it's intended to trigger a failure if we ever add a new profile to LPMS without corresponding support in go-livepeer)

doc/rtmpwebhookauth.md
j0sh: Fixed in bf686d88df7a7bc34da1baca46351a730d83fa37 , and a number of other usages of "RTMP" that were also missed.

doc/transcodingoptions.md
j0sh: Yeah that whole line is nonsense. Fixed in bf686d88df7a7bc34da1baca46351a730d83fa37

yondonfu (approved):
LGTM!

darkdarkdragon (approved):
LGTM

PR #1559, opened 2020-07-05
server: webserver pricing specific endpoints for bcast max price and orchestrator base price

kyriediculous:

Add price specific setter endpoints to the webserver  

- created new handler `/setBroadcastMaxPrice`
- created new handler `/setOrchestratorPrice`

**Checklist:**

### DIFF
--- a/server/webserver.go
+++ b/server/webserver.go
@@ -82,7 +82,36 @@ func (s *LivepeerServer) cliWebServerHandlers(bindAddr string) *http.ServeMux {
 
 	mux.Handle("/vote", mustHaveFormParams(voteHandler(s.LivepeerNode.Eth), "poll", "choiceID"))
 
-	//Set the broadcast config for creating onchain jobs.
+	// Set the broadcaster's max price
+	mux.HandleFunc("/setBroadcastMaxPrice", func(w http.ResponseWriter, r *http.Request) {
+		if err := r.ParseForm(); err != nil {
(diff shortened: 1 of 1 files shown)


yondonfu (commented):
While I understand that having a standalone price setter makes it convenient to *just* programatically set the price,  I wonder if an alternative could be to modify the `/setBroadcastConfig` and `/setOrchestratorConfig` endpoints to allow for certain params to be omitted so that node state is only updated for the provided params? In this case, if the only params provided to a `/setBroadcastConfig` request are `pricePerUnit` and `pixelsPerUnit`, only a price update is executed and other state updates are skipped. And in the case where no params are provided, an error could be returned indicating that no params were provided. I bring this up because we might want to allow other config options to be set during run time such as the ticket EV and I'm a bit wary of adding new endpoints for each new config option - curious what you think.

kyriediculous:
If you think it's worth the engineering  effort to do this properly _right now_  considering it's not on the critical path (the project for which this is a requirement is an untested, unofficial tool for now).

yondonfu:
Thanks for the clarification on what this feature is needed for.

Given that the external tool that depends on this feature is still being iterated on, I don't see a need to rush this feature (it is also missing tests) in if there might be a better way to implement it. Let's think about what we believe is the "proper" way to implement this feature and then we can work on merging it in.

kyriediculous:
closed in favor of https://github.com/livepeer/go-livepeer/pull/1597

PR #1560, opened 2020-07-06
Add reward flag and fail fast if no services are enabled

yondonfu:

This PR adds a `-reward` flag that can be used to explicitly enable/disable the reward service and also updates the node to fail fast if no services are enabled.

- a8d9d4b: Add the `-reward` flag. By default, the flag value is false. If the node detects that its address is registered on-chain *and* the flag is not set, the flag value is set to true.
- e919460: Update the node to fail fast if no services are enabled. A recent PR updated the node not to fail fast if a node type is not set because we want users to be able to run the node with standalone services such as the reward service or round initializer service. Now, the node considers node types services as well and will fail fast if there are no services enabled.
- 25cf7fd: Add docs for the flags to enable/disable the reward service and round initializer service
- feaa81b: Increase sleep in a flaky test

Manual testing and added a test for fail fast behavior when no services are enabled in `test_args.sh`.

**Checklist:**

### DIFF
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -130,6 +130,8 @@ func main() {
 	// Redemption service
 	redeemer := flag.Bool("redeemer", false, "Set to true to run a ticket redemption service")
 	redeemerAddr := flag.String("redeemerAddr", "", "URL of the ticket redemption service to use")
+	// Reward service
+	reward := flag.Bool("reward", false, "Set to true to run a reward service")
 	// Metrics & logging:
 	monitor := flag.Bool("monitor", false, "Set to true to send performance metrics")
 	version := flag.Bool("version", false, "Print out the version")
@@ -622,13 +624,19 @@ func main() {
 			glog.Infof("Redeemer started on %v", *httpAddr)
 		}
 
-		// If the node address is an on-chain registered address, start the reward service
-		t, err := n.Eth.GetTranscoder(n.Eth.Account().Address)
-		if err != nil {
-			glog.Error(err)
-			return
+		if !isFlagSet["reward"] {
+			// If the node address is an on-chain registered address, start the reward service
+			t, err := n.Eth.GetTranscoder(n.Eth.Account().Address)
+			if err != nil {
+				glog.Error(err)
+				return
+			}
+			if t.Status == "Registered" {
+				*reward = true
+			}
 		}
-		if t.Status == "Registered" {
+
+		if *reward {
 			// Start reward service
 			// The node will only call reward if it is active in the current round
 			rs := eventservices.NewRewardService(n.Eth, blockPollingTime)
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -272,8 +272,8 @@ func main() {
 		n.NodeType = core.TranscoderNode
 	} else if *broadcaster {
 		n.NodeType = core.BroadcasterNode
-	} else {
-		glog.Info("Node type not set with -broadcaster, -transcoder, -orchestrator or -redeemer")
+	} else if !*reward && !*initializeRound {
+		glog.Fatalf("No services enabled; must be at least one of -broadcaster, -transcoder, -orchestrator, -redeemer, -reward or -initializeRound")
 	}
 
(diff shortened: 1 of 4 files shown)


cmd/livepeer/livepeer.go
```diff
@@ -270,8 +272,8 @@ func main() {
 		n.NodeType = core.TranscoderNode
 	} else if *broadcaster {
 		n.NodeType = core.BroadcasterNode
-	} else {
-		glog.Info("Node type not set with -broadcaster, -transcoder, -orchestrator or -redeemer")
+	} else if !*reward && !*initializeRound {
+		glog.Fatalf("No services enabled; must be one of -broadcaster, -transcoder, -orchestrator, -redeemer, -reward or -initializeRound")
```
kyriediculous: I don't think "one of" is entirely accurate anymore how about "must be at least one of" , not that the latter is entirely accurate because not all combinations are possible either.

cmd/livepeer/livepeer.go
```diff
@@ -622,13 +624,19 @@ func main() {
 			glog.Infof("Redeemer started on %v", *httpAddr)
 		}
 
-		// If the node address is an on-chain registered address, start the reward service
-		t, err := n.Eth.GetTranscoder(n.Eth.Account().Address)
-		if err != nil {
-			glog.Error(err)
-			return
+		if !isFlagSet["reward"] {
```
kyriediculous, line 627: Can't we just do `if *reward` because it's just a boolean and not an actual value ? Anyway there's little actual difference here so leaving it up to you.

kyriediculous (commented):
Looks good ! Just some minor comments, not going to explicitly request changes.

cmd/livepeer/livepeer.go
yondonfu: If we check the value of `*reward` here, the code in the following block would never be executed because the default value of `*reward` is false even if the user does not explicitly set the `-reward` flag. Instead, we can check `isFlagSet["reward"]` which will be false when the `-reward` flag is not set. The resulting behavior is:

- When the `-reward` flag is not set, if the node's address is registered on-chain, set `*reward` to true and *then* use the value of `*reward` to determine whether to enable/disable the reward service
    - This should preserve the existing behavior that orchestrator operators expect - if the node is run with `-orchestrator` and it uses an on-chain registered address, the node runs a reward service
- When the `-reward` flag is set, use the value of `*reward` to determine whether to enable/disable the reward service

cmd/livepeer/livepeer.go
yondonfu: Fixed in 447190b

cmd/livepeer/livepeer.go
kyriediculous: AFAIK it is still not different from 

```
if !*reward {
  if on-chain addr == node addr {
   *reward = true
}
}

if *reward {
 // start  reward service 
}
```

kyriediculous:
Let's rebase !

cmd/livepeer/livepeer.go
yondonfu: In your example, the reward service would be enabled when the node's address is registered on-chain even if the user specifies `-reward=false`. In the current code,  the user can disable the reward service even when the node's address is registered on-chain by specifying `-reward=false`.

yondonfu:
Rebased

kyriediculous (approved):

cmd/livepeer/livepeer.go
kyriediculous: Right, but that would be completely useless at this stage since the on-chain address is the only one that can call reward 

Which means starting a node with the on-chain addr but `-reward=false` would only be useful in the case of key duplication which we don't advise to do anyway.

yondonfu: You're right that the use cases for running a node with an on-chain registered address and `-reward=false` feel limited right now. However, I think disabling the reward service when the user specifies `-reward=false` even if the node's address is registered on-chain better matches user expectations. As a user, if I know that `-reward` enables/disables the reward service, I expect `-reward=false` to disable the reward service. If we can better match flag usage with user expectations, I think making a small code change is worthwhile.

PR #1566, opened 2020-07-10
eth: reward with hint

### LINKED ISSUE #1546: Have the rewardService use hints
Currently the rewardService calls `BondingManager.Reward()` without using hints. 
This means that for every reward call by the orchestrator it's position in the active set will be recalculated in linear fashion even if the position doesn't change. 

The protocol provides the ability to provide hints to the `reward()` call,  which indicate what the position of the orchestrator in the active set should become after the call. When the hints are correct the operation becomes constant. 

In favor of not changing the `LivepeerEthClient` API signature for `reward()` , which currently calls the generated stubs [directly](https://github.com/livepeer/go-livepeer/blob/3775af2a72ac8cea857ba3246b5558d541414d2e/eth/client.go#L140) , we can create a wrapper method `client.Reward()` which calculates the hints and then calls `client.RewardWithHint`.

**Additional context**
This PR in LivepeerJS adds the same functionality for staking related operations to the explorer: https://github.com/livepeer/livepeerjs/pull/776

kyriediculous:
Adds support for calling reward with hint by the reward service 

- create a helper `getStakingHints()` that is reusable for all staking actions 
- create a wrapper on the EthClient for calling Reward , which calls `getStakingHints()` and then `RewardWithHint()`  with the hints 
- Added a unit test for `getStakingHints` 

Wrote a unit test for the hint helper function with 100% coverage (per `go-code-cover`)

Fixes #1546

**Checklist:**

### DIFF
--- a/eth/client.go
+++ b/eth/client.go
@@ -14,9 +14,9 @@ package eth
 //go:generate abigen --abi protocol/abi/Poll.abi --pkg contracts --type Poll --out contracts/poll.go
 import (
 	"context"
-	"errors"
 	"fmt"
 	"math/big"
+	"sort"
 	"strings"
 	"time"
 
@@ -32,6 +32,7 @@ import (
 	"github.com/livepeer/go-livepeer/eth/contracts"
 	lpTypes "github.com/livepeer/go-livepeer/eth/types"
 	"github.com/livepeer/go-livepeer/pm"
+	"github.com/pkg/errors"
 )
 
 var (
@@ -738,7 +739,107 @@ func (c *client) Vote(pollAddr ethcommon.Address, choiceID *big.Int) (*types.Tra
 	return poll.Vote(opts, choiceID)
 }
 
+func (c *client) Reward() (*types.Transaction, error) {
+	addr := c.accountManager.Account().Address
+
+	currentRound, err := c.CurrentRound()
+	if err != nil {
+		return nil, err
+	}
+
+	ep, err := c.GetTranscoderEarningsPoolForRound(c.accountManager.Account().Address, currentRound)
+	if err != nil {
+		return nil, err
+	}
+
+	mintable, err := c.CurrentMintableTokens()
+	if err != nil {
+		return nil, errors.Wrapf(err, "unable to get current mintable tokens")
+	}
+
+	totalBonded, err := c.GetTotalBonded()
+	if err != nil {
+		return nil, errors.Wrapf(err, "unable to get total bonded")
+	}
+
+	if totalBonded.Cmp(big.NewInt(0)) == 0 {
+		return nil, errors.New("no rewards to be minted")
+	}
+
+	// reward = (current mintable tokens for the round * active transcoder stake) / total active stake
+	reward := new(big.Int).Div(new(big.Int).Mul(mintable, ep.TotalStake), totalBonded)
+
+	// get the transcoder pool
+	transcoders, err := c.TranscoderPool()
+	if err != nil {
+		return nil, errors.Wrapf(err, "unable to get transcoder pool")
+	}
+
+	// get max pool size
+	maxSize, err := c.GetTranscoderPoolMaxSize()
+	if err != nil {
+		return nil, errors.Wrapf(err, "unable to get transcoder pool max size")
+	}
+
+	hints := simulateTranscoderPoolUpdate(addr, reward.Add(reward, ep.TotalStake), transcoders, len(transcoders) == int(maxSize.Int64()))
+
+	return c.RewardWithHint(hints.PosPrev, hints.PosNext)
+}
+
 // Helpers
+
+// simulateTranscoderPoolUpdate simulates an update to the transcoder pool and returns the positional hints for a transcoder accordingly.
+// if the transcoder will not be in the updated set no hints will be returned
+func simulateTranscoderPoolUpdate(del ethcommon.Address, newStake *big.Int, transcoders []*lpTypes.Transcoder, isFull bool) lpTypes.TranscoderPoolHints {
+	for i, t := range transcoders {
+		if t.Address == del {
+			// I don't think an out-of-bounds panic is an issue here when i == len(transcoders) - 1
+			// because transcoders[len(transcoders):] is valid
+			transcoders = append(transcoders[:i], transcoders[i+1:]...)
+			break
+		}
+	}
+
+	// insert 'del' into the pool
+	transcoders = append(transcoders, &lpTypes.Transcoder{
+		Address:        del,
+		DelegatedStake: newStake,
+	})
+
+	// re-sort the list
+	sort.SliceStable(transcoders, func(i, j int) bool {
+		return transcoders[i].DelegatedStake.Cmp(transcoders[j].DelegatedStake) > 0
+	})
+
+	// if the list was full evict the last transcoder
+	if isFull {
+		transcoders = transcoders[:len(transcoders)-1]
+	}
+
+	return findTranscoderHints(del, transcoders)
+}
+
+func findTranscoderHints(del ethcommon.Address, transcoders []*lpTypes.Transcoder) lpTypes.TranscoderPoolHints {
+	hints := lpTypes.TranscoderPoolHints{}
+
+	// do a linear search to get the previous and next transcoder relative to 'del'
+	for i, t := range transcoders {
+		if t.Address == del && len(transcoders) > 1 {
+			if i == 0 {
+				// 'del' is head
+				hints.PosNext = transcoders[i+1].Address
+			} else if i == len(transcoders)-1 {
+				// 'del' is tail
+				hints.PosPrev = transcoders[i-1].Address
+			} else {
+				hints.PosNext = transcoders[i+1].Address
+				hints.PosPrev = transcoders[i-1].Address
+			}
+		}
+	}
+
+	return hints
+}
 
 func (c *client) ContractAddresses() map[string]ethcommon.Address {
 	addrMap := make(map[string]ethcommon.Address)
--- a/eth/client_test.go
+++ b/eth/client_test.go
@@ -1,0 +1,159 @@
+package eth
+
+import (
+	"math/big"
+	"testing"
+
+	ethcommon "github.com/ethereum/go-ethereum/common"
+	lpTypes "github.com/livepeer/go-livepeer/eth/types"
+	"github.com/stretchr/testify/assert"
+)
+
+func copyTranscoders(transcoders []*lpTypes.Transcoder) []*lpTypes.Transcoder {
+	cp := make([]*lpTypes.Transcoder, 0)
+	for _, tr := range transcoders {
+		trCp := new(lpTypes.Transcoder)
+		trCp.Address = tr.Address
+		trCp.DelegatedStake = new(big.Int)
+		*trCp.DelegatedStake = *tr.DelegatedStake
+		cp = append(cp, trCp)
+	}
+	return cp
+}
+
+func TestSimulateTranscoderPool(t *testing.T) {
+	assert := assert.New(t)
+
+	// use copyTranscoders() to avoid mutating the orignal slice
+	transcoders := []*lpTypes.Transcoder{
+		{
+			Address:        ethcommon.HexToAddress("aaa"),
+			DelegatedStake: big.NewInt(5),
+		},
+		{
+			Address:        ethcommon.HexToAddress("bbb"),
+			DelegatedStake: big.NewInt(4),
+		},
+		{
+			Address:        ethcommon.HexToAddress("ccc"),
+			DelegatedStake: big.NewInt(3),
+		},
+		{
+			Address:        ethcommon.HexToAddress("ddd"),
+			DelegatedStake: big.NewInt(2),
+		},
+		{
+			Address:        ethcommon.HexToAddress("eee"),
+			DelegatedStake: big.NewInt(1),
+		},
+	}
+
+	// transcoder is not in pool and doesn't have enough stake to join the pool
+	hints := simulateTranscoderPoolUpdate(ethcommon.HexToAddress("fff"), big.NewInt(1), copyTranscoders(transcoders), true)
+	assert.Equal(hints, lpTypes.TranscoderPoolHints{})
+
+	// transcoder is not in pool and pool is full, transcoder joins the list, current tail is evicted
+	pool := copyTranscoders(transcoders)
+	hints = simulateTranscoderPoolUpdate(ethcommon.HexToAddress("fff"), big.NewInt(2), pool, true)
+	assert.Equal(hints, lpTypes.TranscoderPoolHints{
+		PosPrev: ethcommon.HexToAddress("ddd"),
+	})
+	assert.NotEqual(pool[len(pool)-1].Address, ethcommon.HexToAddress("eee"))
+
+	// transcoder is not in pool and pool is not full
+	// transcoder takes last spot
+	hints = simulateTranscoderPoolUpdate(ethcommon.HexToAddress("fff"), big.NewInt(1), copyTranscoders(transcoders), false)
+	assert.Equal(hints, lpTypes.TranscoderPoolHints{
+		PosPrev: ethcommon.HexToAddress("eee"),
+	})
+
+	// transcoder is not in pool and pool is not full
+	// transcoder takes first spot
+	hints = simulateTranscoderPoolUpdate(ethcommon.HexToAddress("fff"), big.NewInt(100), copyTranscoders(transcoders), false)
+	assert.Equal(hints, lpTypes.TranscoderPoolHints{
+		PosNext: ethcommon.HexToAddress("aaa"),
+	})
+
+	// transcoder is not in pool and pool is not full
+	// transcoder takes second spot
+	hints = simulateTranscoderPoolUpdate(ethcommon.HexToAddress("fff"), big.NewInt(5), copyTranscoders(transcoders), false)
+	assert.Equal(hints, lpTypes.TranscoderPoolHints{
+		PosPrev: ethcommon.HexToAddress("aaa"),
+		PosNext: ethcommon.HexToAddress("bbb"),
+	})
+
+	// last transcoder's stake increases but remains in the last spot
+	hints = simulateTranscoderPoolUpdate(ethcommon.HexToAddress("eee"), big.NewInt(2), copyTranscoders(transcoders), false)
+	assert.Equal(hints, lpTypes.TranscoderPoolHints{
+		PosPrev: ethcommon.HexToAddress("ddd"),
+	})
+
+	// transcoder is in pool and moves up one from the last spot
+	pool = copyTranscoders(transcoders)
+	hints = simulateTranscoderPoolUpdate(ethcommon.HexToAddress("eee"), big.NewInt(3), pool, false)
+	assert.Equal(hints, lpTypes.TranscoderPoolHints{
+		PosPrev: ethcommon.HexToAddress("ccc"),
+		PosNext: ethcommon.HexToAddress("ddd"),
+	})
+	assert.Equal(ethcommon.HexToAddress("ddd"), pool[len(pool)-1].Address)
+
+	// transcoder is in pool and takes the first spot
+	hints = simulateTranscoderPoolUpdate(ethcommon.HexToAddress("ccc"), big.NewInt(6), copyTranscoders(transcoders), false)
+	assert.Equal(hints, lpTypes.TranscoderPoolHints{
+		PosNext: ethcommon.HexToAddress("aaa"),
+	})
+
+	// transcoder decreases stake to be equal of that of the last transcoder, transcoder takes the last spot instead
+	hints = simulateTranscoderPoolUpdate(ethcommon.HexToAddress("aaa"), big.NewInt(1), copyTranscoders(transcoders), false)
+	assert.Equal(hints, lpTypes.TranscoderPoolHints{
+		PosPrev: ethcommon.HexToAddress("eee"),
+	})
+
+	// transcoder decreases stake to be lower than that of the last transcoder, transcoder takes the last spot
+	hints = simulateTranscoderPoolUpdate(ethcommon.HexToAddress("aaa"), big.NewInt(0), copyTranscoders(transcoders), false)
+	assert.Equal(hints, lpTypes.TranscoderPoolHints{
+		PosPrev: ethcommon.HexToAddress("eee"),
+	})
+}
+
+func TestFindTranscoderHints(t *testing.T) {
+	assert := assert.New(t)
+
+	transcoders := []*lpTypes.Transcoder{
+		{
+			Address:        ethcommon.HexToAddress("aaa"),
+			DelegatedStake: big.NewInt(5),
+		},
+		{
+			Address:        ethcommon.HexToAddress("bbb"),
+			DelegatedStake: big.NewInt(4),
+		},
+		{
+			Address:        ethcommon.HexToAddress("ccc"),
+			DelegatedStake: big.NewInt(3),
+		},
+		{
+			Address:        ethcommon.HexToAddress("ddd"),
+			DelegatedStake: big.NewInt(2),
+		},
+		{
+			Address:        ethcommon.HexToAddress("eee"),
+			DelegatedStake: big.NewInt(1),
+		},
+	}
+
+	// del == 'aaa' == head
+	hints := findTranscoderHints(ethcommon.HexToAddress("aaa"), transcoders)
+	assert.Equal(hints.PosPrev, ethcommon.Address{})
+	assert.Equal(hints.PosNext, ethcommon.HexToAddress("bbb"))
+
+	// del == 'eee' == tail
+	hints = findTranscoderHints(ethcommon.HexToAddress("eee"), transcoders)
+	assert.Equal(hints.PosPrev, ethcommon.HexToAddress("ddd"))
+	assert.Equal(hints.PosNext, ethcommon.Address{})
+
+	// del == 'ccc'
+	hints = findTranscoderHints(ethcommon.HexToAddress("ccc"), transcoders)
+	assert.Equal(hints.PosPrev, ethcommon.HexToAddress("bbb"))
+	assert.Equal(hints.PosNext, ethcommon.HexToAddress("ddd"))
+}
--- a/eth/types/contracts.go
+++ b/eth/types/contracts.go
@@ -122,3 +122,8 @@ func (v VoteChoice) String() string {
 func (v VoteChoice) IsValid() bool {
 	return v == Yes || v == No
 }
+
+type TranscoderPoolHints struct {
+	PosNext common.Address
+	PosPrev common.Address
+}


yondonfu (commented):
First, a few comments/observations:

IIUC the order of operations in `getStakingHints()` is:

1. Do a linear search in `transcoders` to find the addresses we care about so we can update their delegated stake
2. Sort `transcoders` based on delegated stake
3. Do a linear search in `transcoders` to find the addresses we care about so we can record the insert hint for each address (i.e the previous address in the sorted slice and the next address in the sorted slice)

Given that this method would typically be called using a slice returned by `TranscoderPool()` which is already sorted, it would be nice if we can avoid the sort operation and an extra linear search operation. 

Additionally, I don't think this helper takes into account the situation where a transcoder is evicted from the pool. This isn't a problem for this particular PR, but just noting it since I think the intent for the helper is for it to be re-usable.

---

~~After doing an initial review, I wonder if a simpler solution might be to use a read-only function (defined in its own standalone contract) that simulates a reward call and then returns the state of the pool after the reward call. Something like~~:

```
address public bondingManager;

function simulateReward() public returns (Transcoder[] memory) {
    // Use delegatecall so the msg.sender for this function call is the same
    // as the msg.sender in BondingManager.reward()
    bondingManager.delegatecall(
        abi.encodeWithSignature("reward()")
    );

    // This wouldn't be a delegatecall since we actually care about the state of the BondingManager
    return getTranscoderPool();
}
```

~~The above code snippet assumes that there is a function to return the entire transcoder pool - see livepeer/protocol#388. The caller could then calculate the hint by doing a linear search through the returned pool to find where the relevant address is located [1]. The benefit would be that we can rely on the existing contract code that already handles finding the correct insertion position after a transaction that updates the pool without replicating it off-chain as well. Another benefit would be that we can avoid the N RPC requests for fetching the pool where N is the # of orchestrators in the pool (max 100).~~

~~EDIT: This might not work...investigating.~~

~~[1] Would be even simpler if we could return the hint from the read-only function, but the BondingManager only exposes a function to fetch the next transcoder in the pool and not the previous transcoder...~~

UPDATE: The solution describe above doesn't work - see [this comment](https://github.com/livepeer/go-livepeer/pull/1566#issuecomment-665835877).

kyriediculous:
Given
- LivepeerJS uses a similar logic 
- livepeer/protocol#388 is not implemented
- urgency
- reward calls only happen once a day so optimising for RPC requests might feel premature

I'd rather push this forward as is and optimise iteratively if really necessary, WDYT? 

edit: And potentially accommodate for evictions

yondonfu:
> After doing an initial review, I wonder if a simpler solution might be to use a read-only function (defined in its own standalone contract) that simulates a reward call and then returns the state of the pool after the reward call.

After thinking about this some more, I realized the solution described doesn't work because:

- I think the only way to access state after updates simulated in an `eth_call` is if a read-only function itself accesses the state after the simulated update and returns relevant values
- The above means that the `reward()` call would need to be wrapped in a read-only function
- The read-only function cannot invoke `reward()` via a delegatecall because while `msg.sender` will be the EOA that invokes the read-only function, `reward()` will be executed using the storage context of the caller contract
- The read-only function cannot invoke `reward()` via a call because `msg.sender` will be the caller contract

I'll update my original comment to reflect this.

> I'd rather push this forward as is and optimise iteratively if really necessary, WDYT?

I was mainly wondering if a simpler solution existed as re-implementing the hint position calculation off-chain is error-prone. Since the alternative described above is not feasible, then calculating the hint position off-chain might be the best that we can do right now. In that light, calculating the hint position off-chain sounds good to me.

eth/client.go
```diff
@@ -738,8 +739,99 @@ func (c *client) Vote(pollAddr ethcommon.Address, choiceID *big.Int) (*types.Tra
...
+	hints := c.getStakingHints(addr, ethcommon.Address{}, reward, transcoders)
+	return c.RewardWithHint(hints.newPosPrev, hints.newPosNext)
+
+}
+
 // Helpers
 
+// getStakingHints returns hints for both the new and old (if applicable) delegate
+// amount provided should be negative when unbonding.
+func (c *client) getStakingHints(newDel ethcommon.Address, oldDel ethcommon.Address, amount *big.Int, transcoders []*lpTypes.Transcoder) *struct {
```
yondonfu: This function also does not seem to be able to handle the case where the new delegate's stake increases by X and the old delegate's stake decreases by Y where X > Y (i.e. if a delegator moves stake *and* bonds additional stake).

Additionally, having to pass an additional empty address argument feels cumbersome since it would be required in most cases except for when a delegator is moving stake from an old delegate to a new delegate.

WDYT about something like the following:

```
func simulateTranscoderPoolUpdate(del ethcommon.Address, newStake *big.Int, transcoders []*lpTypes.Transcoder) ... {
    // Impl of this function is largely the same as the current function except for the below code
    // which only searches for del instead of oldDel and newDel
    for _, t := range transcoders {
        if t.Address == del {
            // Update delegated stake
        }
    }

    // Sort transcoders

    // Calculate and return hint for `del`
}
```

Usage for the case mentioned above would be something like:

```
oldDelStake = oldDel.stake - movedStake
oldDelHint = simulateTranscoderPoolUpdate(oldDel, oldDelStake, transcoders)

newDelStake = newDel.stake + movedStake + additionalStake
// transcoders would already be updated based on oldDel update
newDelHint = simulateTranscoderPoolUpdate(newDel, newDelStake, transcoders)
```

The downside of this approach is that two calls to `simulateTranscoderPoolUpdate` would require two sort operations. This would only be the case when using this function for a delegator that is moving stake. Doesn't feel pressing to address right now, but if this becomes a problem we could always remove the sort operation and instead do a linear search + insertion - the transcoder pool would start off sorted anyway so each linear search + insertion would result in a slice that is still sorted.

yondonfu: Looks like this can just be a function instead of a method on `client` since it doesn't access any state on `client`.

yondonfu: Looks like this function does more than just get a hint since it actually needs to simulate an update (mutate a slice) to the transcoder pool in order to calculate the hint. I think something like `simulateTranscoderPoolUpdate()` would be more descriptive - WDYT?

kyriediculous:
I just realised the current implementation also doesn't account for a new transcoder joining the active set.

Currently thinking about how I can implement that within the same linear search as updating an existing transcoder. Think I might have something. 

Another issue is that we might be better off using this as a client method rather than a pure function because we currently have no idea if the transcoder list passed in as an argument is actually "full" or not, this is important to insert a new transcoder in the list to check whether we need to evict someone or not. 

The main issue with turning this into a method instead of a pure function is the testability because we need to access a bit of on-chain state now such as: 
- Max pool size 
- Delegate's activation and deactivation round (if transcoder) to determine if it is already in the pool (this prevents us from having to do a first linear search to establish this and then a second one to update the pool)
- Current Round to check against the activation and deactivation round 
- And now that we're at it , the transcoder pool itself instead of as a function argument

kyriediculous:
I split up the logic in a pure function and a client method so that purely finding the next and previous positions for an address in a list is tested. The helper assumes the list provided is already sorted to acquire correct hints. 

I think we cover all cases now.

eth/client.go
```diff
@@ -738,8 +739,109 @@ func (c *client) Vote(pollAddr ethcommon.Address, choiceID *big.Int) (*types.Tra
...
+	if err != nil {
+		glog.Errorf("Unable to calculate hints, submitting reward transaction without hints. err=%v", err)
+	}
+	return c.RewardWithHint(hints.PosPrev, hints.PosNext)
+
+}
+
 // Helpers
 
+// simulateTranscoderPoolUpdate simulated an  update to the transcoder pool and returns the positional hints for a delegator accordingly.
```
yondonfu: "simulated" -> "simulates"

yondonfu: "an  update" -> "an update"

^ Looks like there is an extra space between those two words.

yondonfu: "positional hints for a delegator" -> "positional hints for a transcoder" since we're calculating a hint for an address that is going to be in the transcoder pool.

eth/client.go
```diff
@@ -738,8 +739,109 @@ func (c *client) Vote(pollAddr ethcommon.Address, choiceID *big.Int) (*types.Tra
...
+// if the delegator will not be in the updated set 'nil' will be returned.
```
yondonfu: Since we're returning a struct value and not a struct pointer, the return value would be the zero value for the struct instead of nil. Maybe the comment can note that the hint will be empty in this case.

eth/client.go
```diff
@@ -738,8 +739,109 @@ func (c *client) Vote(pollAddr ethcommon.Address, choiceID *big.Int) (*types.Tra
...
+func (c *client) simulateTranscoderPoolUpdate(del ethcommon.Address, newStake *big.Int) (lpTypes.StakingHints, error) {
+
+	// get max pool size
+	maxSize, err := c.GetTranscoderPoolMaxSize()
+	if err != nil {
+		return lpTypes.StakingHints{}, err
+	}
+	maxSizeInt := int(maxSize.Int64())
+
+	// get the transcoder  pool
```
yondonfu: "transcoder  pool" -> "transcoder pool"

^ Looks like there is an extra space.

eth/client.go
```diff
@@ -738,8 +739,109 @@ func (c *client) Vote(pollAddr ethcommon.Address, choiceID *big.Int) (*types.Tra
...
+	transcoders, err := c.TranscoderPool()
+	if err != nil {
+		return lpTypes.StakingHints{}, err
+	}
+
+	// if the transcoder pool is full and the newStake for 'del' is lower  than that of the last transcoders
```
yondonfu: "lower" -> "less than or equal to" since the address will not join the active set when its stake is equal to the stake of the last transcoder in the pool.

eth/client.go
```diff
@@ -738,8 +739,109 @@ func (c *client) Vote(pollAddr ethcommon.Address, choiceID *big.Int) (*types.Tra
...
+		return nil, errors.Wrapf(err, "unable to get curernt mintable tokens")
+	}
+
+	totalBonded, err := c.GetTotalBonded()
+	if err != nil {
+		return nil, errors.Wrapf(err, "unable to get total bonded")
+	}
+
+	// reward = (current mintable tokens for the round * active transcoder stake) / total active stake
+	reward := new(big.Int).Div(new(big.Int).Mul(mintable, t.DelegatedStake), totalBonded)
```
yondonfu: The reward is actually calculated based on the active stake for the round (stored in the `totalStake` field for the earnings pool for the round) and not the delegated stake which could change after the start of a round. Using the delegated stake in this calculation could result in an incorrect value for the reward in the scenario where the delegated stake is updated before reward is called.

eth/client.go
```diff
@@ -738,8 +739,109 @@ func (c *client) Vote(pollAddr ethcommon.Address, choiceID *big.Int) (*types.Tra
...
+	addr := c.accountManager.Account().Address
+
+	t, err := c.GetTranscoder(addr)
+	if err != nil {
+		return nil, errors.Wrapf(err, "unable to get transcoder %v", addr.Hex())
+	}
+
+	mintable, err := c.CurrentMintableTokens()
+	if err != nil {
+		return nil, errors.Wrapf(err, "unable to get curernt mintable tokens")
```
yondonfu: "curernt" -> "current"

eth/client.go
```diff
@@ -738,8 +739,109 @@ func (c *client) Vote(pollAddr ethcommon.Address, choiceID *big.Int) (*types.Tra
...
+	if err != nil {
+		return nil, errors.Wrapf(err, "unable to get total bonded")
+	}
+
+	// reward = (current mintable tokens for the round * active transcoder stake) / total active stake
+	reward := new(big.Int).Div(new(big.Int).Mul(mintable, t.DelegatedStake), totalBonded)
+
+	hints, err := c.simulateTranscoderPoolUpdate(addr, reward.Add(reward, t.DelegatedStake))
+	if err != nil {
+		glog.Errorf("Unable to calculate hints, submitting reward transaction without hints. err=%v", err)
```
yondonfu: It feels a bit weird to me to submit the reward tx without hints when `simulateTranscoderPoolUpdate()` returns an error, but to not submit the reward tx without hints when any of the other calls (i.e. `GetTranscoder()`, `GetTotalBonded()`, etc.) earlier in this method return an error considering that the other calls are also necessary for calculating hints.

I think we should have the same behavior for all calls that are necessary for calculating hints. I also think it would be reasonable to *not* submit a reward tx without hints and just pass the error back to the caller in these scenarios and let the caller handle retrying given that if one of the calls returns an error there is a good chance that there is a problem with the ETH connection which means that the reward tx would likely return an error too.

eth/client.go
```diff
@@ -738,8 +739,109 @@ func (c *client) Vote(pollAddr ethcommon.Address, choiceID *big.Int) (*types.Tra
...
+
+	// get the transcoder  pool
+	transcoders, err := c.TranscoderPool()
+	if err != nil {
+		return lpTypes.StakingHints{}, err
+	}
+
+	// if the transcoder pool is full and the newStake for 'del' is lower  than that of the last transcoders
+	// 'del' will not be in the active set and thus there are no hints to return.
+	if len(transcoders) == maxSizeInt && newStake.Cmp(transcoders[maxSizeInt-1].DelegatedStake) <= 0 {
```
yondonfu: This condition would also evaluate to true when the stake of a transcoder that is already in the pool decreases (i.e. due to unbonding) - we should still generate a hint in that case.

eth/client.go
```diff
@@ -738,8 +739,109 @@ func (c *client) Vote(pollAddr ethcommon.Address, choiceID *big.Int) (*types.Tra
...
+		return lpTypes.StakingHints{}, nil
+	}
+
+	// Find the transcoder and update it's stake
```
yondonfu: Would we be able to test more of this logic if it was moved into the pure function `findTranscoderHints()` and the values returned by the ETH calls in this method passed into the pure function?

yondonfu (changes requested):

eth/client.go
kyriediculous: sorry my spacebar was broken 😅

eth/client.go
kyriediculous: We could but it seemed a little spaghetti-ish 🍝  to me so I opted for the in-between solution.

eth/client.go
kyriediculous: I don't think this conditional returns true if the transcoder remains in the transcoder pool after unbonding ? 

The list is full and the updated stake is lower than that of the last transcoder in the list -> transcoder is evicted so there are no hints to be calculated. 

This statement kind of short-circuits this function to skip sorting and a linear search but removing it would lead to the same result.

eth/client.go
kyriediculous: Okay that sounds reasonable to me 👍

eth/client.go
yondonfu: If the pool is full and the stake of a transcoder in the pool decreases such that it is lower than that of the last transcoder in the pool the transcoder would remain in the pool, but it would just be in the last position in the pool. At the moment, I believe this conditional would return true and no hint would be returned when instead we would want to return a hint that indicates that the transcoder should now be in the last position of the pool.

eth/client.go
yondonfu: What about something like the following:

```golang
func (c *client) Reward() (*types.Transaction, error) {
	...

	transcoders, err := c.TranscoderPool()
	if err != nil {
		return nil, err
	}

	hint := simulateTranscoderPoolUpdate(addr, reward.Add(reward, ep.TotalStake), transcoders)

	return c.RewardWithHint(hint.PosPrev, hint.PosNext)
}

func simulateTranscoderPoolJoin(del ethcommon.Address, newStake *big.Int, transcoders []*lpTypes.Transcoder, maxSize int) lpTypes.StakingHints {
	if len(transcoders) == maxSize {
		if newStake.Cmp(transcoders[len(transcoders)-1].DelegatedStake) <= 0 {
			return lpTypes.StakingHints{}
		}

		// Evict the last transcoder
		transcoders = transcoders[1:]
	}

	return simulateTranscoderPoolUpdate(del, newStake, transcoders)
}

func simulateTranscoderPoolUpdate(del ethcommon.Address, newStake *big.Int, transcoders []*lpTypes.Transcoder) lpTypes.StakingHints {
	// Find the transcoder and update it's stake
	for i, t := range transcoders {
		// if 'del' is in the pool simply update it's stake
		if t.Address == del {
			t.DelegatedStake = newStake
		} else if i == len(transcoders)-1 {
			// 'del' is not in the pool, insert it and sort later
			transcoders = append(transcoders, &lpTypes.Transcoder{
				Address:        del,
				DelegatedStake: newStake,
			})
		}
	}

	// re-sort the list after the stake update
	sort.SliceStable(transcoders, func(i, j int) bool {
		return transcoders[i].DelegatedStake.Cmp(transcoders[j].DelegatedStake) > 0
	})

	hints := lpTypes.StakingHints{}

	// do a linear search to get the previous and next transcoder relative to 'del'
	for i, t := range transcoders {
		if t.Address == del {
			if i == 0 && len(transcoders) > 1 {
				// 'del' is head
				hints.PosNext = transcoders[i+1].Address
			} else if i == len(transcoders)-1 && len(transcoders) > 1 {
				// 'del' is tail
				hints.PosPrev = transcoders[i-1].Address
			} else {
				hints.PosNext = transcoders[i+1].Address
				hints.PosPrev = transcoders[i-1].Address
			}
		}
	}

	return hints
}
```

The pool update simulation logic in `simulateTranscoderPoolUpdate()` could be unit tested - a convenient way to write the tests might be to use the [rapid](https://github.com/flyingmutant/rapid) package to auto-generate test cases that can be executed to not only ensure that the relevant transcoder's stake is updated in the pool, but to also ensure that an invariant (i.e. the transcoder pool slice is sorted after each function call) is preserved. The rapid package is used in a [few other places](https://github.com/livepeer/go-livepeer/search?q=rapid&unscoped_q=rapid) in the codebase.

The eviction logic could be isolated in `simulateTranscoderPoolJoin()` which calls `simulateTranscoderPoolUpdate()` internally after executing the eviction logic. This logic could be unit tested too. This might be helpful since the eviction logic only matters when `bondWithHint()` is invoked in the future.

WDYT?

eth/client.go
```diff
@@ -738,8 +739,113 @@ func (c *client) Vote(pollAddr ethcommon.Address, choiceID *big.Int) (*types.Tra
...
+	}
+	maxSizeInt := int(maxSize.Int64())
+
+	// get the transcoder pool
+	transcoders, err := c.TranscoderPool()
+	if err != nil {
+		return lpTypes.StakingHints{}, err
+	}
+
+	// if the transcoder pool is full and the newStake for 'del' is less than that of the last transcoder
```
yondonfu: "less" should be "less than or equal to" since the transcoder will not be able to join the pool if its stake is equal to the stake of the last transcoder in the pool - see [this code](https://github.com/livepeer/protocol/blob/826af1124f645a819b54eb0500ddc31f0d2af755/contracts/bonding/BondingManager.sol#L1034).

eth/client.go
kyriediculous: Sorry, yup you're right (there is no new transcoder to take it's place right away)

eth/client.go
kyriediculous: I don't really like having a separate method for bonding and it takes curves around the bigger issue at hand which is the testability of our ETH client package, not that I think the latter should be in scope for this PR either. 

I know this reply doesn't get us farther ahead but do you strongly feel that the current testability of this new logic is insufficient? 

To me it feels like being able to correctly calculate the hints given a transcoder and sorted transcoder list is at the core here especially if we get rid of L796.

yondonfu: > I don't really like having a separate method for bonding and it takes curves around the bigger issue at hand which is the testability of our ETH client package

I think there are two points here:

1. Pushing as much logic as possible into a pure function so that it can be tested
2. Using a separate function when calculating the hint for `bondWithHint()`

Regarding 1: Suppose we had the ability to stub the contract bindings right now. Presumably, we'd be writing tests that stub the calls to the contract bindings and then execute the code paths after those calls. If that is the case, why not write the tests for those code paths now? Then, if we are able to stub the contract bindings in the future we would already have the tests for the other code paths and would only need to add tests that involve the stub contract binding calls.

Regarding 2: The main reason I suggested this is to avoid adding code that is never tested either via a test suite or manually while still allowing for the possibility of adding that code easily in the future. At the moment, we're including eviction logic that is not unit testable and not manually testable since it is not triggered anywhere (since it is not required for `rewardWithHint()`). One way to address this would be to add the eviction logic later in a function that invokes `simulateTranscoderPoolUpdate()` under the hood when we actually need to invoke that logic. 

What is the downside of having a separate function for bonding? Seem like a reasonable option given that joining the transcoder pool can be considered an update to the transcoder pool preceded by eviction logic that is only necessary for the join operation. 

> To me it feels like being able to correctly calculate the hints given a transcoder and sorted transcoder list is at the core

Yeah those operations are important, but they only work if the logic for finding the right position for the transcoder is also correct.

kyriediculous: Okay I understand your points.

To me it just felt a little bit excessive considering we don't have testing for the rest of the `Client` methods and considering there's actually no real grave negative side effects possible. 

But if you feel we should tackle full testability through moving to a pure function for now , so be it happy to make the change.

yondonfu: > we don't have testing for the rest of the Client methods

If the logic introduced in this PR was simpler (i.e. just a couple lines of a code that use values from contract calls) then I would be more inclined to not add tests. But, the reasons why I do think it makes sense to add additional tests for this PR are:

- The logic for correctly calculating a hint is more involved
- Most of the logic required to correctly calculate a hint does not depend on contract calls (besides using initial values from contract calls)

Lastly, I think adding a few tests actually makes our lives easier. They can confirm for us that the hint is actually correctly calculated. If we were only manually testing, it would be more difficult to determine whether the hint is correctly calculated (we'd see a gas cost reduction, but we might not be certain that this is the greatest gas cost reduction that is possible).

If we want to minimize some of the testing, I think it is reasonable to omit the eviction/join logic until we actually need it i.e. when we work on #1567.

kyriediculous: I think now we have 

- a single method
- full test coverage
- ejections and joins covered
- changing delegates covered (two calls required, one for the old del and one for the new del)

There's been a quite a few changes so reviewing the final diff might be more helpful.

eth/client.go
kyriediculous: actually removed this line in the latest commit

eth/client.go
kyriediculous: added a test for this case f79e4f96

eth/client.go
```diff
@@ -738,8 +739,108 @@ func (c *client) Vote(pollAddr ethcommon.Address, choiceID *big.Int) (*types.Tra
...
+
+	return findTranscoderHints(del, transcoders)
+}
+
+func findTranscoderHints(del ethcommon.Address, transcoders []*lpTypes.Transcoder) lpTypes.StakingHints {
+	hints := lpTypes.StakingHints{}
+
+	// do a linear search to get the previous and next transcoder relative to 'del'
+	for i, t := range transcoders {
+		// fmt.Println(i, t.Address.Hex(), t.DelegatedStake.Int64())
```
yondonfu: Let's remove this.

eth/types/contracts.go
```diff
@@ -122,3 +122,8 @@ func (v VoteChoice) String() string {
 func (v VoteChoice) IsValid() bool {
 	return v == Yes || v == No
 }
+
+type StakingHints struct {
```
yondonfu: Would a more accurate name be `PoolHints` since this struct would also be used for calling reward and is generally used to represent positional hints within the on-chain transcoder pool?

eth/client_test.go
```diff
@@ -0,0 +1,153 @@
...
+	})
+	assert.Equal(ethcommon.HexToAddress("ddd"), pool[len(pool)-1].Address)
+
+	// transcoder is in pool and takes the first spot
+	hints = simulateTranscoderPoolUpdate(ethcommon.HexToAddress("ccc"), big.NewInt(6), copyTranscoders(transcoders), false)
+	assert.Equal(hints, lpTypes.StakingHints{
+		PosNext: ethcommon.HexToAddress("aaa"),
+	})
+
+	// transcoder decreases stake , takes the last spt
```
yondonfu: "spt" -> "spot"

eth/client_test.go
```diff
@@ -0,0 +1,153 @@
...
+	hints = simulateTranscoderPoolUpdate(ethcommon.HexToAddress("aaa"), big.NewInt(0), copyTranscoders(transcoders), false)
```
yondonfu, line 113: FWIW a transcoder wouldn't be in the pool if its stake is 0 so 1 would be a more realistic value. Not a big deal either way since the value used for `newStake` just needs to move the transcoder to the last position in the pool so I'll leave it up to you whether you want to use a different value.

yondonfu (commented):
Changes look pretty good. Just left a few minor comments.

Have you also tested these changes by running a node connected to a Ethereum network and observing lower gas costs for calling reward relative to when the node calls reward without using hints?

eth/types/contracts.go
kyriediculous: Agreed, I'll name it `TranscoderPoolHints` for clarity

kyriediculous:
> Have you also tested these changes by running a node connected to a Ethereum network and observing lower gas costs for calling reward relative to when the node calls reward without using hints?

I have not but sounds like a good idea to do, I'll set up a manual test by logging the gas cost of the reward transaction and filling up a transcoder pool using the devtool.

eth/client_test.go
kyriediculous: ~~Wouldn't that result in that transcoder taking the one before last spot because it was already in the list, had a higher stake originally and we traverse in descending order? If it were a new transcoder not in the list yet, it would take the last spot.~~

Agreed that using '0'  is not representative of actual on-chain behaviour, I could bump everything up by 1.

kyriediculous: Actually scratch the above,  I think it should indeed take the last spot because it would get removed first and then re-added. This seems to be inconsistent with the way we sort the slice.

kyriediculous: Thanks your comment made me change that test and observed it failing, which uncovered that bug.

We now mimic the on-chain behaviour of `updateKey()` 

1. If the transcoder is in the pool remove it 
2. Append the transcoder 
3. Sort the list 
4. Evict the last if  list is full 
5. Get the hints 

Also added a test that confirms we no longer encounter this bug

 84c2b0c9

kyriediculous: I think this bug might actually exist in LivepeerJS as well https://github.com/livepeer/livepeerjs/blob/0e5edcc097a972f5a4221055617a31b48f912b5d/packages/explorer-2.0/lib/utils.tsx#L355

eth/client.go
```diff
@@ -786,49 +786,49 @@ func (c *client) Reward() (*types.Transaction, error) {
 
 // simulateTranscoderPoolUpdate simulates an update to the transcoder pool and returns the positional hints for a transcoder accordingly.
 // if the transcoder will not be in the updated set no hints will be returned
-func simulateTranscoderPoolUpdate(del ethcommon.Address, newStake *big.Int, transcoders []*lpTypes.Transcoder, isFull bool) lpTypes.StakingHints {
-	// Find the transcoder and update it's stake
-	inPool := false
-	for _, t := range transcoders {
-		// if 'del' is in the pool simply update it's stake
+func simulateTranscoderPoolUpdate(del ethcommon.Address, newStake *big.Int, transcoders []*lpTypes.Transcoder, isFull bool) lpTypes.TranscoderPoolHints {
+	transcodersUpdate := make([]*lpTypes.Transcoder, 0)
```
yondonfu: Are you trying to avoid mutating `transcoders` here? If so, the rest of this function will still mutate `transcoders` because  mutations to `transcodersUpdate` will affect the array that backs `transcoders`.

If you want to avoid mutating `transcoders`, then the best bet here is probably something like:

```golang
for _, t := range transcoders {
    if t.Address == del {
        continue
    }
    transcodersUpdate = append(transcodersUpdate, t)
}
```

While avoiding mutations within the function is helpful for reasoning about code, the one downside I see is that you wouldn't be able to chain calls to this function i.e.

```golang
// Update A
simulateTranscoderPoolUpdate(A, ...)
// Update B
simulateTranscoderPoolUpdate(B, ...)
```

which you might need to do when moving stake and calling `bondWithHint()` later on.

If you aren't trying to avoid mutations in this function, then a simplified version of the below code could be:

```golang
for i, t := range transcoders {
    if t.Address == del {
        // I don't think an out-of-bounds panic is an issue here when i == len(transcoders) - 1
        // because transcoders[len(transcoders):] is valid
        transcoders = append(transcoders[:i], transcoders[i+1:]...)
        break
    }
}
```

kyriediculous: Not trying to avoid mutating. 

the last example would work. Thanks.

Fixed in 933023d3

eth/client_test.go
yondonfu: Just to clarify:

Suppose we have transcoders T1 and T2. In the pool, if transcoders have the same stake, both the ordering T1 -> T2 and the ordering T2 -> T1 can be valid based on the [validInsertPosition()](https://github.com/livepeer/protocol/blob/826af1124f645a819b54eb0500ddc31f0d2af755/contracts/libraries/SortedDoublyLL.sol#L250) logic in the SortedDoublyLL library right now. To illustrate:

**Scenario 1**

T1 is inserted into the pool with a hint where `prev = T2`. T1 would be inserted in the pool after T2.

**Scenario 2**

T1 is inserted into the pool without a hint. We would descend the list of transcoders in the pool. When the next transcoder in the pool is T2 we would see that T1's stake is >= the prev transcoder's stake and is = to T2's stake. This is also a valid insert position for T1.

kyriediculous:
Manual tests results: 

1 transcoder in the list and no hints
```
Invoking transaction: "rewardWithHint". Inputs: "_newPosPrev: 0x0000000000000000000000000000000000000000  _newPosNext: 0x0000000000000000000000000000000000000000"  Hash: "0xed3b58a614011fb18c4926f0e92dc5163b8cdf437cfbe76b6388bf3d75129c48"  Gas Cost: "387528".
```

4 transcoders in the list and using hints: 
```
  Invoking transaction: "rewardWithHint". Inputs: "_newPosPrev: 0x7e4B3D1EDEE2a9506e1334003a46f8eD863D84Cb  _newPosNext: 0x17CE8fe02A26C2F1C0c7CA08707914DAc78e9363"  Hash: "0x52d53f3b4d642c41ae37b49db634e44aefbdfd453996f3716fad24f127f1ed97"  Gas Cost: "331766".
```

Did run into a division by zero error that I'm currently looking in to: 

```
goroutine 31 [running]:
math/big.nat.div(0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0xc011fda440, 0x2, 0x7, 0xc011fda400, ...)
	/usr/local/go/src/math/big/nat.go:648 +0x481
math/big.(*Int).QuoRem(0xc011f7cf80, 0xc01073bdc8, 0xc011f7cf60, 0xc01073bca0, 0x8, 0xc00012a8c0)
	/usr/local/go/src/math/big/int.go:239 +0xbf
math/big.(*Int).Div(0xc011f7cf80, 0xc01073bdc8, 0xc011f7cf60, 0xc01073bdc8)
	/usr/local/go/src/math/big/int.go:250 +0x6d
github.com/livepeer/go-livepeer/eth.(*client).Reward(0xc0001b8000, 0xc011f7c6c0, 0xffffffffffffffff, 0x0)
	/Users/nico/Desktop/livepeer/go-livepeer/eth/client.go:766 +0x2ac
github.com/livepeer/go-livepeer/eth/eventservices.(*RewardService).tryReward(0xc00030eff0, 0xc010856748, 0x2)
	/Users/nico/Desktop/livepeer/go-livepeer/eth/eventservices/rewardservice.go:119 +0x601
github.com/livepeer/go-livepeer/eth/eventservices.(*RewardService).Start.func1(0xc0106abb60, 0xc00030eff0, 0x5224020, 0xc0106beb80)
	/Users/nico/Desktop/livepeer/go-livepeer/eth/eventservices/rewardservice.go:47 +0xd9
created by github.com/livepeer/go-livepeer/eth/eventservices.(*RewardService).Start
	/Users/nico/Desktop/livepeer/go-livepeer/eth/eventservices/rewardservice.go:43 +0xc3
```

edit: `totalBonded` was 0 , I'll add a check but this should really not occur on mainnet.

yondonfu (approved):
LGTM after a rebase

PR #1577, opened 2020-07-15
verification: Update results parsing + remove model from request

yondonfu:

This PR updates the `EpicClassifier` type in the `verification` package to be compatible with the latest version of [verification-classifier](https://github.com/livepeer/verification-classifier). The specific changes in verification-classifier that required updates in `EpicClassifier` are:

- The API response returns an integer for the `tamper` field where 1 indicates Tampered and 0 indicates Correct. 1/0 is returned now instead of a continuous value because the verifier uses a meta model (combines the output of the old OCSVM model with another Catboost Classifier model - the former is better at avoiding misclassifying correct videos as tampered and the latter is better at catching tampered videos) that does not output a continuous value
- The API response returns a float for the `ocsvm_dist` field that *is* a continuous value representing the distance from the decision boundary for the OCSVM model. This value was previously returned for the `tamper` field. This value can still be used to order results in cases where all results are marked as tampered which might indicate a misclassification
- The model URL is no longer needed in the API request because the verifier now loads a specific model on startup

- 82c82c0: Update the results parsing code for the API response and remove the model URL from the API request

Updated unit tests and ran a local broadcaster + orchestrator + verifier.

**Checklist:**

### DIFF
--- a/verification/epic.go
+++ b/verification/epic.go
@@ -44,7 +44,6 @@ type epicRequest struct {
 	Source         string          `json:"source"`
 	Renditions     []epicRendition `json:"renditions"`
 	OrchestratorID string          `json:"orchestratorID"`
-	Model          string          `json:"model"`
 }
 
 type epicResultFields struct {
@@ -52,7 +51,11 @@ type epicResultFields struct {
 	AudioAvailable bool    `json:"audio_available"`
 	AudioDistance  float64 `json:"audio_dist"`
 	Pixels         int64   `json:"pixels"`
-	Tamper         float64 `json:"tamper"`
+	// 1 if the model detects a tamper and 0 otherwise
+	Tamper int `json:"tamper"`
+	// This value is used to order multiple results that are all marked as tampered which might indicate a misclassification
+	// In this case, the result with the highest distance is considered the most preferable
+	OCSVMDist float64 `json:"ocsvm_dist"`
 }
 
 type epicResults struct {
@@ -79,13 +82,13 @@ func epicResultsToVerificationResults(er *epicResults) (*Results, error) {
 		if v.AudioAvailable && v.AudioDistance != 0.0 && err == nil {
 			err = ErrAudioMismatch
 		}
-		if v.VideoAvailable && v.Tamper <= 0 && err == nil {
+		if v.VideoAvailable && v.Tamper > 0 && err == nil {
 			err = ErrTampered
 		}
 		if !v.VideoAvailable && err == nil {
 			err = ErrVideoUnavailable
 		}
-		score += v.Tamper / float64(len(er.Results))
+		score += v.OCSVMDist / float64(len(er.Results))
 		pixels = append(pixels, v.Pixels)
 	}
 	return &Results{Score: score, Pixels: pixels}, err
@@ -135,7 +138,6 @@ func (e *EpicClassifier) Verify(params *Params) (*Results, error) {
 		Source:         sourcePath,
 		Renditions:     renditions,
 		OrchestratorID: oid,
-		Model:          "https://storage.googleapis.com/verification-models/verification.tar.xz",
 	}
 	reqData, err := json.Marshal(req)
 	if err != nil {
--- a/verification/epic_test.go
+++ b/verification/epic_test.go
@@ -35,8 +35,8 @@ func TestEpic_EpicResultsToVerificationResults(t *testing.T) {
 	// Success case
 	r = &epicResults{
 		Results: []epicResultFields{
-			{VideoAvailable: true, Pixels: 123, Tamper: 2.0},
-			{VideoAvailable: true, Pixels: 456, Tamper: 3.0},
+			{VideoAvailable: true, Pixels: 123, Tamper: 0, OCSVMDist: 2.0},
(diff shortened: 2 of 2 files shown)


kyriediculous (approved):
I don't see much wrong with this but my domain knowledge is not exactly up to par here.

verification/epic.go
```diff
@@ -135,7 +139,6 @@ func (e *EpicClassifier) Verify(params *Params) (*Results, error) {
 		Source:         sourcePath,
 		Renditions:     renditions,
 		OrchestratorID: oid,
-		Model:          "https://storage.googleapis.com/verification-models/verification.tar.xz",
```
darkdarkdragon, line 138: Shouldn't we remove `Model` field from `epicRequest` entirely?

verification/epic.go
```diff
@@ -52,7 +52,11 @@ type epicResultFields struct {
 	AudioAvailable bool    `json:"audio_available"`
 	AudioDistance  float64 `json:"audio_dist"`
 	Pixels         int64   `json:"pixels"`
```
darkdarkdragon, line 53: In my test with master branch of https://github.com/livepeer/verification-classifier, it returns pixels field as string:
```json
"pixels": "1034500",
````

**Review by darkdarkdragon — CHANGES_REQUESTED** [review]

File: verification/epic.go (hunk 2 above)
Comment (yondonfu) [reply verification/epic.go:53]: Hm I just tested this and it looks like the pixels field is returned as an integer:

```
Response Body: {"model":"http://storage.googleapis.com/verification-models/verification-metamodel-2020-07-06.tar.xz","orchestrator_id":"https://0.0.0.0:8935","results":[{"audio_available":true,"audio_dist":0.0,"fps":30.0,"frame_rate":true,"height":240,"ocsvm_dist":0.017332596389410038,"pixels":6134400 ...
```

I tested this by doing the following:

- `git clone https://github.com/livepeer/verification-classifier.git`
- `cd verification-classifier`
- `bash launch_api.sh`
- Running a local broadcaster with `-v 99` connected to the verifier and a local orchestrator

File: verification/epic.go (hunk 1 above)
Comment (yondonfu) [reply verification/epic.go:138]: Fixed in 41fbc8c

File: verification/epic.go (hunk 2 above)
Comment (darkdarkdragon) [reply verification/epic.go:53]: OK, I found the problem - `verification-classifier` just sends back what was in `pixels` field.
I did this request, which I've found in readme for `verification-classifier`:
```
POST http://localhost:5000/verify 
content-type: application/json


{
  "source": "https://storage.googleapis.com/livepeer-verifier-renditions/480p/-3MYFnEaYu4.mp4",
  "renditions": [
    {
      "uri": "https://storage.googleapis.com/livepeer-verifier-renditions/144p_black_and_white/-3MYFnEaYu4.mp4"
    },
    {
      "uri": "https://storage.googleapis.com/livepeer-verifier-renditions/144p/-3MYFnEaYu4.mp4",
      "resolution": {
        "height": "144",
        "width": "256"
      },
      "frame_rate": "24",
      "pixels": "1034500"
    }
  ],
  "orchestratorID": "foo"
}
```
and getting back
```
{
  "model": "http://storage.googleapis.com/verification-models/verification-metamodel-2020-07-06.tar.xz",
  "orchestrator_id": "foo",
  "results": [
    {
      "audio_available": false,
      "fps": 29.97002997002997,
      "height": 144,
      "ocsvm_dist": -0.05450179660177856,
      "tamper": 1,
      "tamper_sl": 1,
      "tamper_ul": 1,
      "uri": "https://storage.googleapis.com/livepeer-verifier-renditions/144p_black_and_white/-3MYFnEaYu4.mp4",
      "video_available": true,
      "width": 256
    },
    {
      "audio_available": false,
      "fps": 29.97002997002997,
      "frame_rate": false,
      "height": 144,
      "ocsvm_dist": 0.015780843648907666,
      "pixels": 1034500,
      "pixels_post_verification": 2.8062608506944446,
      "pixels_pre_verification": 0.09354202835648148,
      "resolution": {
        "height": "144",
        "height_post_verification": 1.0,
        "height_pre_verification": 1.0,
        "width": "256",
        "width_post_verification": 1.0,
        "width_pre_verification": 1.0
      },
      "tamper": 0,
      "tamper_sl": 1,
      "tamper_ul": 0,
      "uri": "https://storage.googleapis.com/livepeer-verifier-renditions/144p/-3MYFnEaYu4.mp4",
      "video_available": true,
      "width": 256
    }
  ],
  "source": "https://storage.googleapis.com/livepeer-verifier-renditions/480p/-3MYFnEaYu4.mp4"
}
```

Though I don't know how to interpret that - is that a bug in the `verification-classifier`? Shouldn't it be something that `verification-classifier` calculate by itselves?

yondonfu: Hm you're right about the verifier just setting the pixels field to whatever is included in the request body...

I think the verifier does compute the pixels, but that information is incorporated into `pixels_pre_verification` and `pixels_post_verification` with the former being the ratio b/w the expected pixels (indicated in the request) and the pixels calculated based on metadata and the latter being the ratio b/w the expected pixels and the pixels calculated based on reading the frames. Created livepeer/verification-classifier#128 to track a fix there.

darkdarkdragon (approved):
LGTM

PR #1582, opened 2020-07-20
Record mode

### LINKED ISSUE #1572: proposal: generic URL-based OS syntax
## Abstract
Let's define OS like `s3://whatever`

## Motivation

Currently, OS have to be defined the `-s3bucket`, `-s3creds`, and other such CLI options. If we continue to add more CLI parameters for each OS, we'll have thousands. Additionally, we're going to want to refer to OS for failed segments (#1398), recording (#1565) and webhooks at some point, so it's good to have a generic way to refer to OSs.

## Proposed Solution

TLDR: Change OS to have the form s3://ACCESS_KEY_ID:SECRET_ACCESS_KEY@eu-central-1/testbucket. Those get called by `-objectStore`, and maybe eventually `-failObjectStore`, and `-recordObjectStore`.

This will allow us to refer to object stores generically in multiple contexts using the same syntax. We should probably deprecate and support the previous options for backwards-compatibilty. This proposed syntax borrows heavily from ffmpeg's syntax for customizing various protocols.

### S3 stores

Before:

```shell
livepeer -broadcaster -s3bucket eu-central-1/testbucket -s3creds ACCESS_KEY_ID/SECRET_ACCESS_KEY
```

After:

```
livepeer -broadcaster -objectStore s3://ACCESS_KEY_ID:SECRET_ACCESS_KEY@eu-central-1/testbucket
```

### GCP Stores

This one allows a keyfile, so I'm not really sure how to express that. This is what you'd technically do for URL compatibility with `./path/to/keyfile` but it sure isn't very fun to write on the command line:

```
livepeer -broadcaster -objectStore "gs://ACCESS_KEY_ID:.%2Fpath%2Fto%2Fkeyfile@example-gcp-bucket/path"
```

We should also support providing the key as an inline JSON blob, but that's even harder to type.

### Custom S3 Stores (https://github.com/livepeer/go-livepeer/pull/1206)

Before:

```
livepeer -broadcaster -s3bucket http://localhost:9000/eli-bucket -s3creds minioadmin/minioadmin
```

After:

```shell
livepeer -broadcaster -objectStore s3://minioadmin:minioadmin@localhost:9000/eli-bucket
```

### Saving errored segments (https://github.com/livepeer/go-livepeer/pull/1398)

```
livepeer -orchestrator -transcoder -failObjectStore s3://ACCESS_KEY_ID:SECRET_ACCESS_KEY@eu-central-1/testbucket
```

### Filesystem storage (https://github.com/livepeer/go-livepeer/commit/5b776565debc43256a1ad235896f89faebd79a86)

```
livepeer -broadcaster -objectStore file://./path/to/whatever
```

(Or file could be the default to make it a little cleaner.)

We should probably do the primary OS refactor first, ship that, and then incorporate long-running PRs like https://github.com/livepeer/go-livepeer/pull/1398 and https://github.com/livepeer/go-livepeer/pull/1206.

## Known Unknowns

How do we handle GCP? Their access keys are massive JSON blobs containing a private TLS key. Separate keyfile param as a query string?

## Alternatives

Perhaps instead of having three things called `objectStore`, `failObjectStore`, and `recordObjectStore`, we could add support for multiple object stores per-stream, with each one configurable to handle different cases...

```
livepeer -broadcaster -objectStore s3://ACCESS_KEY_ID:SECRET_ACCESS_KEY@eu-central-1/testbucket -objectStore file://./my-recordings?record=true -objectStore gs://user:password/bucket/failed-segments?failed=true
```

The only use case I can think of for that would be to have multiple records... you can't have more than one primary OS and I'm not sure why you'd want to send failed segments several places. Multiple records, though, is potentially interesting — perhaps you want to record to an S3 bucket but also save a copy on the local filesystem as a backup.

### LINKED ISSUE #1609: Mist timeouts with record mode enabled
I have a stream that's working fine when running against broadcasters on master (f1e26c0a9d8e4638b1c617fe95d0e45784be0a8e) but on `it/record-mode` I get consistent timeouts:

```
 Aug 25 12:49:11 mdw-default-staging-mist-server-0 mist [2020-08-25 19:49:09] MistProcLivepeer:36ynvagtquow1m6w→36ynvagtquow1m6w (1475) FAIL: Timeout while retrieving https://chi.livepeer.monster/live/814b2c00-e570-4bae-b978-e3e69ecbb79a/318.ts
 Aug 25 12:49:57 mdw-default-staging-mist-server-0 mist [2020-08-25 19:49:54] MistProcLivepeer:36ynvagtquow1m6w→36ynvagtquow1m6w (1475) FAIL: Timeout while retrieving https://chi.livepeer.monster/live/814b2c00-e570-4bae-b978-e3e69ecbb79a/340.ts
 Aug 25 12:50:01 mdw-default-staging-mist-server-0 mist [2020-08-25 19:49:59] MistProcLivepeer:36ynvagtquow1m6w→36ynvagtquow1m6w (1475) FAIL: Timeout while retrieving https://chi.livepeer.monster/live/814b2c00-e570-4bae-b978-e3e69ecbb79a/340.ts
 Aug 25 12:50:47 mdw-default-staging-mist-server-0 mist [2020-08-25 19:50:45] MistProcLivepeer:36ynvagtquow1m6w→36ynvagtquow1m6w (1475) FAIL: Timeout while retrieving https://chi.livepeer.monster/live/814b2c00-e570-4bae-b978-e3e69ecbb79a/366.ts
 Aug 25 12:51:07 mdw-default-staging-mist-server-0 mist [2020-08-25 19:51:05] MistProcLivepeer:36ynvagtquow1m6w→36ynvagtquow1m6w (1475) FAIL: Timeout while retrieving https://chi.livepeer.monster/live/814b2c00-e570-4bae-b978-e3e69ecbb79a/374.ts
```

Bug, or just Backblaze not being fast enough?

darkdarkdragon:
Allows to record streams 

- add generic URL-based OS syntax https://github.com/livepeer/go-livepeer/pull/1582/commits/6b7b0c75280a300e906fe6f0cffeca3b906b0c20
- add `-objectStore` command line argument https://github.com/livepeer/go-livepeer/pull/1582/commits/4cfedeae2e46588f117c8ffe4b588c856493b4ec
- allow to specify OS per-stream in webhook https://github.com/livepeer/go-livepeer/pull/1582/commits/f768a952f4fef93d8d19e3ee0ed3966b3932d175

PR allows to specify global OS for saving streams (using `-recordStore` command line option) or to specify OS per-stream using webhook. 
Two fields added to webhook response structure:
`objectStore` and `recordObjectStore`. 
`objectStore` specifies URL of OS to use for stream in transcoding pipeline, and
`recordObjectStore` specifies URL of OS to store stream to.

Recorded streams can be played back using http://broadcaster:8935/recordings/manifestID/index.m3u8 URL.

Units tests

Fixes #1609 
Fixes #1572 

**Checklist:**


iameli (dismissed):
So we're going to run into some trouble with `nodeId`.

1. It's not set unless `-monitor=true` is set, which results in an empty string being used in the S3 upload URLs and playback breaking.
2. In on-chain mode it uses the broadcaster address... except that'll be the same across all of our production broadcasters in a region, so their JSON uploads will all collide.

I dunno what we wanna do about that exactly.

iameli:
I tried to make a copy of this branch and accidentally deleted and recreated it, sorry if that causes problems anywhere 😂

cmd/livepeer/livepeer.go
```diff
@@ -720,16 +722,53 @@ func main() {
...
+		objectstore = &ustr
+	}
+
+	if *gsBucket != "" && *gsKey != "" {
+		u := url.URL{
+			Scheme:   "gs",
+			Host:     *gsBucket,
+			RawQuery: fmt.Sprintf("keyfile=%s", *gsKey),
+		}
+		glog.Warningf("-gsbucket and -gskey are deprecated. Instead, you can use -objectStore '%s'", u.String())
```
yondonfu: Looks like extraneous single quotes around `%s`.

cmd/livepeer/livepeer.go
```diff
@@ -720,16 +722,53 @@ func main() {
...
+		objectstore = &ustr
+	}
+
+	if *objectstore != "" {
+		prepared, err := drivers.PrepareOSURL(*objectstore)
+		if err != nil {
+			glog.Error("Error creating object store driver: ", err)
+			return
+		}
+		drivers.NodeStorage, err = drivers.ParseOSURL(prepared, false)
```
yondonfu, line 753: Missing error handling.

cmd/livepeer/livepeer.go
```diff
@@ -720,16 +722,53 @@ func main() {
...
+	}
+
+	if *recordstore != "" {
+		prepared, err := drivers.PrepareOSURL(*recordstore)
+		if err != nil {
+			glog.Error("Error creating recordings object store driver: ", err)
+			return
+		}
+		drivers.RecordStorage, err = drivers.ParseOSURL(prepared, true)
```
yondonfu, line 766: Missing error handling.

cmd/livepeer/livepeer.go
```diff
@@ -720,16 +722,53 @@ func main() {
...
 	}
 
 	if *gsBucket != "" && *gsKey != "" {
-		drivers.GSBUCKET = *gsBucket
-		drivers.NodeStorage, err = drivers.NewGoogleDriver(*gsBucket, *gsKey)
+		store, err := drivers.NewGoogleDriver(*gsBucket, *gsKey, false)
```
yondonfu: Do we still need the logic in this conditional if we set `drivers.NodeStorage` with the `if *objectstore != ""` conditional above?

drivers/drivers.go
```diff
@@ -46,14 +92,86 @@ func NewSession(info *net.OSInfo) OSSession {
...
 
+// PrepareOSURL used for resolving files when necessary and turning into a URL. Don't use
+// this when the URL comes from untrusted sources e.g. AuthWebhookUrl.
+func PrepareOSURL(input string) (string, error) {
+	u, err := url.Parse(input)
+	if err != nil {
+		return "", err
+	}
+	if u.Scheme == "gs" {
+		m, _ := url.ParseQuery(u.RawQuery)
```
yondonfu: Needs error handling.

core/playlistmanager.go
```diff
@@ -19,35 +25,175 @@ type PlaylistManager interface {
...
+
+type jsonSeg struct {
+	SeqNo      uint64 `json:"seq_no,omitempty"`
+	URI        string `json:"uri,omitempty"`
+	DurationMS uint64 `json:"duration_ms,omitempty"`
+}
+
+type JsonPlaylist struct {
+	name       string
+	DurationMs uint64               `json:"duration_ms,omitempty"` // total duration of the saved sagments
```
yondonfu, line 65: Nit: Match capitalization for this field in `jsonSeg` and here?

core/playlistmanager.go
```diff
@@ -19,35 +25,175 @@ type PlaylistManager interface {
...
+
+// AddSegmentsToMPL adds segments to the MediaPlaylist
+func (jpl *JsonPlaylist) AddSegmentsToMPL(manifestID, trackName string, mpl *m3u8.MediaPlaylist, extURL string) {
+	for _, seg := range jpl.Segments[trackName] {
+		// make relative URL from absolute one
+		uri := seg.URI
+		mindex := strings.Index(uri, manifestID)
+		if mindex != -1 {
+			if extURL != "" {
+				uri = path.Join(extURL, uri[mindex:])
```
yondonfu, line 107: a) `<extURL>/<manifestID>/...`

b) `/...`

So we expect the URI to be a) when there is an external URL and b) when there is not an external URL? Why do we include the manifestID in a) but exclude it in b)?

core/playlistmanager.go
```diff
@@ -19,35 +25,175 @@ type PlaylistManager interface {
...
+		mseg := &m3u8.MediaSegment{
+			URI:      uri,
+			Duration: float64(seg.DurationMS) / 1000.0,
+		}
+		mpl.InsertSegment(seg.SeqNo, mseg)
+	}
+}
+
+func (jpl *JsonPlaylist) hasTrack(trackName string) bool {
+	for _, track := range jpl.Tracks {
```
yondonfu, line 121: Note to self that this linear search should be ok because in practice the # of tracks = # of renditions which will typically be small (i.e. less than 10).

core/playlistmanager.go
```diff
@@ -63,6 +209,31 @@ func (mgr *BasicPlaylistManager) GetOSSession() drivers.OSSession {
...
+			glog.Error("Error encoding playlist: ", err)
+			return
+		}
+		go func(name string, data []byte) {
+			now := time.Now()
+			_, err := mgr.recordSession.SaveData(name, b, nil)
+			glog.V(common.VERBOSE).Infof("Saving json playlist name=%s size=%d bytes took=%s err=%v", name,
+				len(b), time.Since(now), err)
+		}(mgr.jsonList.name, b)
+		if mgr.jsonList.DurationMs > 60*60*1000 { // 1 hour
```
yondonfu: Use a constant here?

install_ffmpeg.sh
```diff
@@ -115,7 +115,7 @@ if [ ! -e "$HOME/ffmpeg/libavcodec/libavcodec.a" ]; then
     --disable-encoders --disable-decoders --disable-filters --disable-bsfs \
     --disable-postproc --disable-lzma \
     --enable-gnutls --enable-libx264 --enable-gpl \
-    --enable-protocol=https,rtmp,file \
+    --enable-protocol=https,http,rtmp,file \
```
yondonfu, line 118: Hm does this mean that http:// URLs couldn't be passed as filenames for transcoding previously? And is that the main reason for this change?

server/broadcast.go
```diff
@@ -299,7 +300,25 @@ func processSegment(cxn *rtmpConnection, seg *stream.HLSSegment) ([]string, erro
 		return nil, err
 	}
 	name := fmt.Sprintf("%s/%d%s", vProfile.Name, seg.SeqNo, ext)
-	uri, err := cpl.GetOSSession().SaveData(name, seg.Data)
+	ros := cpl.GetRecordOSSession()
+	segDurMs := strconv.Itoa(int(seg.Duration * 1000))
+	if ros != nil {
+		go func() {
+			for i := 0; i < 2; i++ {
```
yondonfu: Can this logic be moved into a function since it needs to be executed for both the source as well as for each rendition?

server/broadcast.go
```diff
@@ -299,7 +300,25 @@ func processSegment(cxn *rtmpConnection, seg *stream.HLSSegment) ([]string, erro
...
+				now := time.Now()
+				uri, err := ros.SaveData(name, seg.Data, map[string]string{"duration": segDurMs})
+				if err != nil {
+					glog.Errorf("Error saving nonce=%d manifestID=%s name=%s to record store try=%d err=%v", nonce, mid, name, i, err)
+				} else {
+					cpl.InsertHLSSegmentJSON(vProfile, seg.SeqNo, uri, seg.Duration)
+					glog.Infof("Successfully saved nonce=%d manifestID=%s name=%s size=%d bytes to record store took=%s",
```
yondonfu: `size=` -> `bytes=` since we use that convention elsewhere? Also avoids the need to specify the units afterwords.

core/playlistmanager.go
```diff
@@ -19,35 +25,175 @@ type PlaylistManager interface {
...
+			i := sort.Search(len(curSegs), func(i int) bool {
+				return curSegs[i].SeqNo >= seg.SeqNo
+			})
+			if i < len(curSegs) && curSegs[i].SeqNo == seg.SeqNo {
+				// x is present at data[i]
+			} else {
+				// x is not present in data,
+				// but i is the index where it would be inserted.
+				if i < len(curSegs) {
+					// glog.Errorf("track %s cur segs len %d i %d seq %d", trackName, len(curSegs), i, seg.SeqNo)
```
yondonfu: Let's clean this up.

drivers/drivers.go
```diff
@@ -46,14 +92,86 @@ func NewSession(info *net.OSInfo) OSSession {
...
+		if err != nil {
+			return "", err
+		}
+		u.User = url.User(string(content))
+	}
+	return u.String(), nil
+}
+
+// ParseOSURL returns the correct OS for a given OS url
+func ParseOSURL(input string, useFullAPI bool) (OSDriver, error) {
```
yondonfu, line 127: When is useFullAPI supposed to be true vs. false?

drivers/gs.go
```diff
@@ -95,21 +103,35 @@ func NewGoogleDriver(bucket, keyFileName string) (OSDriver, error) {
 		jsKey:     &gsKey,
 		parsedKey: parsedKey,
 	}
+	if useFullAPI {
+		client, err := storage.NewClient(context.Background(), option.WithCredentialsJSON(os.keyData))
```
yondonfu: Looks like we're just creating a client and then immediately closing it. What is the purpose of this?

drivers/gs.go
```diff
@@ -125,6 +147,184 @@ func newGSSession(info *net.S3OSInfo) OSSession {
...
 }
 
+func (os *gsSession) OS() OSDriver {
+	return os.gos
+}
+
+func (os *gsSession) createClient() error {
+	client, err := storage.NewClient(context.Background(), option.WithCredentialsJSON(os.keyData))
+	if err != nil {
+		glog.Errorf("Error creating GCP client err=%v", err)
```
yondonfu: If we're returning the error let's let the caller handle it to avoid accidentally handling it in two places?

drivers/gs.go
```diff
@@ -125,6 +147,184 @@ func newGSSession(info *net.S3OSInfo) OSSession {
...
+	return nil
+}
+
+func (os *gsSession) ListFiles(ctx context.Context, prefix, delim string) (PageInfo, error) {
+	if !os.useFullAPI {
+		return nil, errors.New("Not implemented")
+	}
+	if os.client == nil {
+		if err := os.createClient(); err != nil {
+			return nil, err
```
yondonfu, line 258: Why don't we just create and reuse a single client and store it when creating the session?

drivers/local.go
```diff
@@ -185,3 +257,21 @@ func (dc *dataCache) GetData(name string) []byte {
 	}
 	return nil
 }
+
+type dumbPageInfo struct {
```
yondonfu: Is a more descriptive name here something like `singlePageInfo` to convey that there is no next page?

server/broadcast.go
```diff
@@ -525,7 +567,7 @@ func transcodeSegment(cxn *rtmpConnection, seg *stream.HLSSegment, name string,
 
 	if verifier != nil {
 		// verify potentially can change content of segURLs
-		err := verify(verifier, cxn, sess, seg, res.TranscodeData, segURLs, segData)
+		err := verify(verifier, cxn, sess, seg, res.TranscodeData, segURLs, segData, cpl.GetOSSession())
```
yondonfu: Passing cpl.GetOSSession() shouldn't be necessary since we're already passing in cxn which has access to the OSSession vis cxn.pl.GetOSSession()

server/rpc.go
```diff
@@ -283,8 +283,12 @@ func orchestratorInfo(orch Orchestrator, addr ethcommon.Address, serviceURI stri
 
 	os := drivers.NodeStorage.NewSession(authToken.SessionId)
 
-	if os != nil && os.IsExternal() {
-		tr.Storage = []*net.OSInfo{os.GetInfo()}
+	if os != nil {
+		if os.IsExternal() {
+			tr.Storage = []*net.OSInfo{os.GetInfo()}
+		} else {
+			os.EndSession()
```
yondonfu, line 290: Why are we ending the session here?

verification/verify.go
```diff
@@ -57,6 +58,9 @@ type Params struct {
 
 	// Cached data when local object storage is used
 	Renditions [][]byte
+
+	// External object storage used by broadcaster (can be defined per-stream)
+	OS drivers.OSSession
```
yondonfu, line 63: Why do we need this? URIs should already contain locations of the renditions when data is in an external OS.

server/mediaserver.go
```diff
@@ -285,7 +326,7 @@ func authenticateStream(url string) (*authWebhookResponse, error) {
 	rbody, err := ioutil.ReadAll(resp.Body)
 	resp.Body.Close()
 	if resp.StatusCode != 200 {
-		return nil, errors.New(resp.Status)
+		return nil, fmt.Errorf("Status=%d error=%s", resp.StatusCode, string(rbody))
```
yondonfu: Nit: "Status" -> "status" to match capitalization of error

server/mediaserver.go
```diff
@@ -648,6 +690,10 @@ func getRTMPStreamHandler(s *LivepeerServer) func(url *url.URL) (stream.RTMPVide
 // HandlePush processes request for HTTP ingest
 func (s *LivepeerServer) HandlePush(w http.ResponseWriter, r *http.Request) {
 	start := time.Now()
+	if r.Method != "POST" && r.Method != "PUT" {
+		w.WriteHeader(http.StatusMethodNotAllowed)
```
yondonfu: Add an error log here?

server/mediaserver.go
```diff
@@ -896,6 +949,254 @@ func (s *LivepeerServer) HandlePush(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+// HandleRecordings handle requests to /recordings/ endpoint
+func (s *LivepeerServer) HandleRecordings(w http.ResponseWriter, r *http.Request) {
+	if r.Method != "GET" {
+		w.WriteHeader(http.StatusMethodNotAllowed)
```
yondonfu, line 968: Add error logs for failure paths?

yondonfu (changes requested):
I see that we're rolling over JSON playlists every hour and each JSON playlist is identified using a timestamp. So, the external OS ends up with a bunch of those JSON playlists and then we fetch all the playlists prefixed with a specific manifestID to combine them together in HandleRecordings?

core/playlistmanager.go
darkdarkdragon: added as comment:
```
			// If extURL was specified we will put absolute URL to the segment into manifest
			// extURL points to the root of object store, so we should take part of the 'uri'
			// which contains manifestID.
			// If extURL is not specified then we're serving relative URL to the segment,
			// and address at which manifest is served already contains manfiestID
			// (broadcaster.com/recordings/manifestID/index.m3u8), so we're taking
			// part of the 'uri' after the manifestID
```

drivers/drivers.go
darkdarkdragon: false for -objectStore - OS used in transcoding pipeline, true for -recordStore.
This is because of objectStore we need to be able to pass OS credentials between broadcaster and orchestrator and we only need to be able to write segment and to read it back by name.
For recordStore we don't need to pass credentials, but we need to be able list files/dirs, write/read metadata etc.

drivers/gs.go
darkdarkdragon: Probably I was trying to check if credentials are valid, but in reality it doesn't do anything useful. Removed.

drivers/gs.go
darkdarkdragon: Because `NewSession` doesn't return error and so we don't have a way to surface error in case client creation fails.

server/rpc.go
darkdarkdragon: Because in this branch session object is not used, but it will remain in the session's map
https://github.com/livepeer/go-livepeer/blob/72eea36e9cca167c5a565b2384fc7cd2e3c1e533/drivers/local.go#L49
so it will be a memory leak.

verification/verify.go
darkdarkdragon: But in verifier we need to distinguish between local OS and external OS

https://github.com/livepeer/go-livepeer/blob/0e3578d322ec397926f0eea9c63444ecf8e2a4e5/verification/epic.go#L191

install_ffmpeg.sh
darkdarkdragon: Honestly I don't remember, but probably we was using local min.io for OS - which is on http, of course.

cmd/livepeer/livepeer.go

cmd/livepeer/livepeer.go
darkdarkdragon: Fixed

cmd/livepeer/livepeer.go
darkdarkdragon: Fixed.

cmd/livepeer/livepeer.go
darkdarkdragon: Removed.

drivers/drivers.go
darkdarkdragon: Fixed.

core/playlistmanager.go
darkdarkdragon: Fixed.

core/playlistmanager.go
darkdarkdragon: Fixed.

server/broadcast.go
darkdarkdragon: Fixed.

server/broadcast.go
darkdarkdragon: Fixed.

core/playlistmanager.go
darkdarkdragon: Fixed.

drivers/gs.go
darkdarkdragon: Fixed.

drivers/local.go
darkdarkdragon: Fixed.

server/broadcast.go
darkdarkdragon: Fixed.

server/mediaserver.go
darkdarkdragon: Fixed.

server/mediaserver.go
darkdarkdragon: Fixed.

server/mediaserver.go
darkdarkdragon: Fixed.

darkdarkdragon:
> I see that we're rolling over JSON playlists every hour and each JSON playlist is identified using a timestamp. So, the external OS ends up with a bunch of those JSON playlists and then we fetch all the playlists prefixed with a specific manifestID to combine them together in HandleRecordings?

Yes, exactly.

server/broadcast.go
yondonfu: I was actually referring to the for loop with the retry code for `SaveData()` that is executed for the source as well as for each rendition.

darkdarkdragon: fixed

drivers/drivers.go
```diff
@@ -175,6 +175,22 @@ func ParseOSURL(input string, useFullAPI bool) (OSDriver, error) {
 	return nil, fmt.Errorf("unrecognized OS scheme: %s", u.Scheme)
 }
 
+// SaveRetried tries to SaveData specified number of times
+func SaveRetried(sess OSSession, name string, data []byte, meta map[string]string, retryCount int) (string, error) {
+	if retryCount < 1 {
+		return "", fmt.Errorf("Invalid retry cound %d", retryCount)
```
yondonfu: "cound" -> "count"

verification/verify.go
yondonfu: Ah so we removed drivers.IsOwnExternal and now we call IsExternal() and IsOwn() on each driver session.

yondonfu (approved):
Left a comment about a typo but feel free to fix during the squash.

Besides that LGTM

drivers/drivers.go
darkdarkdragon: fixed

PR #1586, opened 2020-07-21
drivers, server: support webhook-defined OS, OS URLs, S3-compatible stores

### LINKED ISSUE #1263: Spec: webhook specified object stores
## Abstract
We should allow to provide object store credentials in web hook response.

## Motivation
Right now object store credentials is specified per broadcaster, we want to specify it per stream.

## Proposed Solution
Web hook response should send web hook credentials to the broadcaster.

Eli Edit: This will follow the fields and syntax defined here https://github.com/livepeer/go-livepeer/issues/1572

## Additional Context
We should also add authentication for web hook, but this is out of scope for this spec.

### LINKED ISSUE #1572: proposal: generic URL-based OS syntax

iameli:

* URL-based OS refactor: https://github.com/livepeer/go-livepeer/issues/1572
* Webhook-defined OS https://github.com/livepeer/go-livepeer/issues/1263
* S3-compatible OS support https://github.com/livepeer/go-livepeer/pull/1206

- Added `-objectStore` parameter per #1572 
- `-s3creds`, `-s3bucket`, `-gskey`, and `-gsbucket` now display a deprecated message and prompt with the correct OS URL
- Added `s3+http://` and `s3+https://` syntax for supporting OS outside of the primary S3 store
- Added an `objectStorage` 

Still to do:
- Figure out how to handle the IsOwnExternal workflow, that'll break with hook-defined OS at the moment
- Update webhook docs
- Integrate with record mode after that lands (or merge these PRs together)
- GS support is a mess
  - `-objectStore gs://%2Fpath%2Fto%2Fkey@bucket-name` is ugly and hard to type
  - There's no actual way to define GS credentials through the webhook, you can only provide them through the command line. The good news is I did test embedding an entire JSON blob in the username field of the URL and it parsed totally fine, so we can just add that for webhook support.
  - On the other hand, we should probably _dis_allow the path option from the webhook. I mean, generally we trust whoever is on the opposite end of `-authWebhookUrl`, but that's no reason to give them the ability to probe for the existence of arbitrary files on the local filesystem!

So far I've just tested with Minio locally. (I'm on the streaming laptop, so no actual cloud credentials allowed here.) I'll make sure to test with S3 & Minio & GCP on both the B and the O before this ships

Fixes #1263
Fixes #1206 
Fixes #1572 

**Checklist:**

- [ ] README and other documentation updated

### DIFF
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -729,22 +729,26 @@ func main() {
 
 	if *gsBucket != "" && *gsKey != "" {
 		u := url.URL{
-			Scheme: "gs",
-			Host:   *gsBucket,
-			User:   url.User(*gsKey),
+			Scheme:   "gs",
+			Host:     *gsBucket,
+			RawQuery: fmt.Sprintf("keyfile=%s", *gsKey),
 		}
-		glog.Warningf("-gsbucket and -gskey are deprecated. Instead, you can use -objectStore %s", u.String())
+		glog.Warningf("-gsbucket and -gskey are deprecated. Instead, you can use -objectStore '%s'", u.String())
 		ustr := u.String()
 		objectstore = &ustr
 	}
 
 	if *objectstore != "" {
-		drivers.NodeStorage, err = drivers.ParseOSURL(*objectstore, true)
+		prepared, err := drivers.PrepareOSURL(*objectstore)
 		if err != nil {
-			glog.Error("Error creating object store driver:", err)
+			glog.Error("Error creating object store driver: ", err)
 			return
 		}
-
+		drivers.NodeStorage, err = drivers.ParseOSURL(prepared, true)
+		if err != nil {
+			glog.Error("Error creating object store driver: ", err)
+			return
+		}
 	}
 
 	core.MaxSessions = *maxSessions
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -159,6 +159,9 @@ func main() {
 
 	// Storage:
 	datadir := flag.String("datadir", "", "data directory")
+	objectstore := flag.String("objectStore", "", "url of primary object store")
+
+	// All deprecated
 	s3bucket := flag.String("s3bucket", "", "S3 region/bucket (e.g. eu-central-1/testbucket)")
 	s3creds := flag.String("s3creds", "", "S3 credentials (in form ACCESSKEYID/ACCESSKEY)")
 	gsBucket := flag.String("gsbucket", "", "Google storage bucket")
@@ -704,10 +707,6 @@ func main() {
 		glog.Error("Should specify both s3bucket and s3creds")
 		return
 	}
-	if *s3bucket != "" {
-		s3bp := strings.Split(*s3bucket, "/")
-		drivers.S3BUCKET = s3bp[1]
-	}
 	if *gsBucket != "" && *gsKey == "" || *gsBucket == "" && *gsKey != "" {
 		glog.Error("Should specify both gsbucket and gskey")
 		return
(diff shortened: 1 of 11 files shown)


darkdarkdragon:
Looks good

> On the other hand, we should probably _dis_allow the path option from the webhook. 

Sounds reasonable

> Figure out how to handle the IsOwnExternal workflow, that'll break with hook-defined OS at the moment

Best option I think would be:
to `OSSession` interface add method `IsOwn(uri string) bool`
In `broadcast.go` change
`if verifier != nil || (bos != nil && !drivers.IsOwnExternal(url)) {`
to
`if verifier != nil || (bos != nil && !bos.IsOwn(url)) {`

In `orchestrator.go` `if drivers.IsOwnExternal(seg.Name)` - left as is

In `verify.go` to `type Params struct` add `BroadcasterOS    drivers.OSSession` field and in `epic.go` change
`if drivers.IsOwnExternal(params.Source.Name) {` to
`if  params.BroadcasterOS != nil && params.BroadcasterOS.IsOwn(params.Source.Name) {`

Other option would be to keep global map of all OS drivers created through web hooks and have `IsOwnExternal` to check against that map - but that will require some cleanup logic.

iameli:
@darkdarkdragon Thanks, on this now.

I'm thinking of the GS file syntax like this: `-objectStore gs://bucket-name?keyfile=/path/to/key`. This pattern would also let us eventually do things like `-objectStore s3://bucket-name?keyfile=/home/iameli/.aws/credentials`, which is kinda nice.

Then, in the CLI, we run a "PrepareOSURL" step that transforms `gs://bucket-name?keyfile=/path/to/key` into `gs://< insert ugly URL-encoded JSON blob here>@bucket-name`.

In the webhook, we won't run that function, so the file options will be a no-op.

iameli:
How do I know whether one of these strings IsOwn? Is it just that the server hostnames and paths match, implying they're the same bucket?

iameli:
Ok, I think I have most of the IsOwn stuff in place. I'm not sure how to get `IsOwn` in epic.go, maybe I'm missing something — looks like the local struct doesn't have an OS.

iameli:
Closing in favor of https://github.com/livepeer/go-livepeer/pull/1582

PR #1588, opened 2020-07-21
Toggle local verification with -localVerify

yondonfu:

This PR adds a `-localVerify` flag with the following behavior:

- If the node is in on-chain mode, the default is `localVerify = true`
- If the node is in off-chain mode, the default is `localVerify = false`
- A user can also explicitly set `-localVerify=true|false`
- When `localVerify = true`, the node will enable pixel count and signature verification
- When `-verifierURL != ""`, the node will also enable pixel count and signature verification
- When `localVerify = false`, the node will disable pixel count and signature verification

This flag should be useful when the node is running in on-chain mode, but working with trusted orchestrators. In this case, the node does not need to incur additional overhead for local verification.

This PR also adds docs on verification that provides a high level overview on how to enable/disable local + tamper verification.

See commit history.

Updated `test_args.sh` and manually tested.

**Checklist:**

### DIFF
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -110,6 +110,7 @@ func main() {
 	verifierURL := flag.String("verifierUrl", "", "URL of the verifier to use")
 
 	verifierPath := flag.String("verifierPath", "", "Path to verifier shared volume")
+	localVerify := flag.Bool("localVerify", true, "Set to true to enable local verification i.e. pixel count and signature verification.")
 	httpIngest := flag.Bool("httpIngest", true, "Set to true to enable HTTP ingest")
 
 	// Transcoding:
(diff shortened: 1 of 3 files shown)


j0sh (approved):
LGTM! 🚢

j0sh:
Might be nice to add some checks to `test_args.sh` but up to you.

kyriediculous:
Do we feel comfortable defaulting to true , this is still quite experimental in my opinion and should be optional.

kyriediculous:
Sorry ignore my previous comment.

Why does this need a flag again? I feel like this should be enabled regardless. 

Tamper verification can be an optional addition

darkdarkdragon (approved):
LGTM

yondonfu:
> Might be nice to add some checks to test_args.sh but up to you.

Added tests in `test_args.sh` in 37c1af8

> Why does this need a flag again? I feel like this should be enabled regardless.

@kyriediculous The flag allows a broadcaster operator to disable local verification i.e. pixel count (used for checking if the broadcaster is charged properly based on # of pixels processed) + signature verification (used for gathering signed evidence which can be submitted to a third party if a tamper is detected) when it is interacting with orchestrators that it trusts. 

This PR already has an approval, but I'll let it sit for another 2 hours and then I'll merge it.

PR #1589, opened 2020-07-23
Pull mode

j0sh:

Adds a single-shot pull mode. With `livepeer -pull <input>` , it'll load a video, segment and transcode. After transcoding completes, the node exits. The input has been tested with local VOD files, HTTP [1] and RTMP.

Also adds a `-streamName` parameter to set the output stream name (manifest ID). If omitted, the stream name is randomly generated.

As segments come in, they are added to a queue. Workers pull from the queue until completion. Multiple segments can be processed in parallel,  but the concurrency factor is hard-coded to 1 until there's a better way of modulating the concurrency (more info within code comments).

The internal node type is set to a broadcaster, but listeners for RTMP/HTTP/CLI are not enabled. This should allow for most broadcaster-specific features to be enabled.

The source segments are placed into `<datadir/media/streamName/>` . The `streamName/` directory is removed after transcoding completes.

In all, this is somewhat useless until the node receives a proper filesystem OS and recording support. See the [earlier pull branch](https://github.com/livepeer/go-livepeer/compare/ja/pull) and the [accompanying documentation](https://gist.github.com/j0sh/5d1112843e620530e759d09ca275cfed) for how the pull feature has worked in the past.

Should in theory also work with mainnet orchestrators, but only tested offchain. (DB writes might become an issue under heavy concurrency, so maybe use separate datadirs if doing multiple single-shot pulls while onchain.)

Some form of this should make it into LPMS eventually to replace the RTMP listener; it would resolve issues around segment polling and faster-than-realtime video ingest (eg, ffmpeg without `-re`).

[1] Normal HTTP videos please. Tested with live HTTP MP4s generated by Mist and those don't play well with the ffmpeg segmenter at the moment; not only are proper segments not produced (seems to be missing sps/pps extradata),  the transcoder *does not exit*. Big problem.

- Pull mode via `-pull`
- Stream name via `-streamName`
- Unit tests

Unit testing, manual testing.

**Checklist:**

### DIFF
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -113,6 +113,8 @@ func main() {
 	httpIngest := flag.Bool("httpIngest", true, "Set to true to enable HTTP ingest")
 
 	// Transcoding:
+	pull := flag.String("pull", "", "URL or path to transcode")
+	streamName := flag.String("streamName", "", "Stream name for output (pull only)")
 	orchestrator := flag.Bool("orchestrator", false, "Set to true to be an orchestrator")
 	transcoder := flag.Bool("transcoder", false, "Set to true to be a transcoder")
 	broadcaster := flag.Bool("broadcaster", false, "Set to true to be a broadcaster")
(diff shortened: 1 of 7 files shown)


thomshutt:

PR #1592, opened 2020-07-29
pm: Use minDepositPendingRatio for max float calc

### LINKED ISSUE #1590: Orchestrator might stop accepting work when it is waiting for a ticket redemption to confirm
**Describe the bug**

Note: There are actually a few more requirements for this scenario to occur that I'll add in follow up comments later. 

Currently, an orchestrator (O) stops accepting work in the following scenario:

Let a broadcaster's (B) reserve allocation for the orchestrator be X.

1. O sets face value of tickets to X to match the B's reserve allocation
2. B's max float is currently X
3. O receives a winning ticket and redeems it
4. B's max float is now 0 because there X amount of value considered as pending until the redemption tx confirms on-chain
5. O calculate the new ticket params to send back to B
6. O sees that B's max float is 0 so it returns an `insufficient sender reserve` error to B
7. Step 6 repeats until O sees that the redemption tx confirms on-chain

When the above scenario occurs, the orchestrator logs will look like this:

```
I0729 18:44:46.518036   93856 backend.go:96]

Invoking transaction: "redeemWinningTicket". Inputs: "_ticket: { Recipient: 0x6788F9AAfEACD5e31559EAaD31fd785DE034260e  Sender: 0x8A8d7Bee38783dfe5568a64Dd50dbceAb58F1dBe  FaceValue: 500000000000000000  WinProb: 23158417847463239084714197001737581570653996933128112807891600000000000000000  SenderNonce: 2  RecipientRandHash: 0x5483875c46c23a6f4be472727d2c737cd91076d922ee2171d22e6edc554e263e  AuxData: 0x000000000000000000000000000000000000000000000000000000000000004783d8f1c512151560453e2f6f3e990d8c2725db83b2888ca5321d569770f8cc8d }  _sig: 0xa76731de83e5999f4ad123b9d22c0e7891156b0dec1d0c89351b6d7d61d983790bd0ca9b265c355f6d0228a130fffae18b062c765e8e375ab1fb7193477148091b  _recipientRand: 2655046112346169819941714599222515298333337682031180840906545582045193452504"  Hash: "0xdd096a25664113c75084935056c656c74bc27371e8a4191cb392127159d4be39".

I0729 18:44:47.340410   93856 gaspricemonitor.go:134] Cached gas price: 20000000000
I0729 18:44:47.818960   93856 orchestrator.go:532] Transcoding of segment manifestID=4772b69c seqNo=12 took=2.782608018s
I0729 18:44:47.818994   93856 orchestrator.go:547] Transcoded segment manifestID=4772b69c seqNo=12 profile=P240p30fps16x9 len=201724
I0729 18:44:47.820103   93856 orchestrator.go:547] Transcoded segment manifestID=4772b69c seqNo=12 profile=P360p30fps16x9 len=293844
I0729 18:44:47.822847   93856 player.go:105] LPMS got HTTP request @ /stream/4772b69c/P240p30fps16x9/12.ts
I0729 18:44:47.822862   93856 player.go:105] LPMS got HTTP request @ /stream/4772b69c/P360p30fps16x9/12.ts
I0729 18:44:48.340974   93856 gaspricemonitor.go:134] Cached gas price: 20000000000
E0729 18:44:48.374584   93856 segment_rpc.go:88] Error updating orchestrator info - err=insufficient sender reserve
```

Note the last log statement which indicates that the broadcaster's reserve is insufficient because the broadcaster's max float is 0.

The broadcaster logs will look like this:

```
Trying to transcode segment nonce=3822204207721735489 seqNo=12
I0729 18:44:45.024911   93846 segment_rpc.go:682] Created new payment - manifestID=4772b69c recipient=0x6788F9AAfEACD5e31559EAaD31fd785DE034260e faceValue=0.5 ETH winProb=0.2000000000 price=5050000000.000 wei/pixel numTickets=2
I0729 18:44:45.025031   93846 segment_rpc.go:382] Submitting segment nonce=3822204207721735489 manifestID=4772b69c seqNo=12 bytes=528844 orch=https://127.0.0.1:8936
I0729 18:44:45.035622   93846 segment_rpc.go:416] Uploaded segment nonce=3822204207721735489 manifestID=4772b69c seqNo=12 orch=https://127.0.0.1:8936 dur=10.555521ms
I0729 18:44:46.750934   93846 broadcast.go:288] Processing segment nonce=3822204207721735489 manifestID=4772b69c seqNo=13 dur=1.98
I0729 18:44:46.750971   93846 broadcast.go:360] No sessions available for segment nonce=3822204207721735489 manifestID=4772b69c seqNo=13
I0729 18:44:46.750999   93846 broadcast.go:130] Starting session refresh manifestID=4772b69c
I0729 18:44:46.751325   93846 rpc.go:223] Connecting RPC to https://127.0.0.1:8936
E0729 18:44:46.756965   93846 rpc.go:215] Could not get orchestrator orch=https://127.0.0.1:8936 err=rpc error: code = Unknown desc = insufficient sender reserve
```

Note the last log statement which indicates that the orchestrator is refusing to connect with the broadcaster because the broadcaster's max float is 0.

I think the original rationale for the current behavior was that it was the most conservative/defensive behavior for O - the current strategy ensures that O will never receive a winning ticket with a face value that is not fully covered by the B's reserve allocation. However, the problem with the above is that O stops accepting work (meaning that the B could never come back to the O!) even if B has a large deposit. So, this doesn't seem like the right tradeoff to make. There is probably an alternative design where the O takes on slightly more risk that the ticket face value is not fully covered by the B's reserve allocation, but will continue to accept work in this scenario even if it is waiting for a ticket redemption to confirm on-chain. One possibility is for O to also look at B's deposit and if it is sufficient (maybe some multiple of the reserve allocation or the tx cost of redemption) then O will continue accepting work.

Rough sketch of an idea around a possible approach to get thinking started:

If B's deposit > B's reserve allocation * a constant factor, then use B's reserve allocation as the max float even if there is pending value (i.e. a pending winning ticket redemption tx).

Note that the above is even more of a problem when B is working with multiple Os associated with the same on-chain address because when a winning ticket is sent *all* Os will stop accepting work until that redemption tx confirms on-chain.

yondonfu:

This PR implements [this spec](https://github.com/livepeer/go-livepeer/issues/1590#issuecomment-665305869).

Differences from the spec:
- The implementation uses the term "pendingAmount" instead of the term "float" because the existing code already uses "pendingAmount". Additionally, the `subFloat()` and `addFloat()` methods in `pm.SenderMonitor` are used to increase the pendingAmount (decrease max float) and decrease the pendingAmount (increase max float) respectively so I felt like using the term "float" would cause confusion with these existing methods
- `minDepositFloatRatio` is re-named to `minDepositPendingRatio` since the implementation does not use the term "float"
- The value for `minDepositPendingRatio` is 3.0 instead of 10.0 as suggested in the spec.
    - I felt like 10.0 was too high
    - Suppose the reserve is 2 ETH and there are 50 active orchestrators. This results in a reserve alloc of .04
    - If the pendingAmount is .08 (i.e. 2 ticket redemptions awaiting confirmation), then with `minDepositPendingRatio = 10.0`, the deposit needs to be 0.8. With `minDepositPendingRatio = 3.0`, the deposit needs to be .24
    - This feels more reasonable, but this value is still a bit arbitrary. Open to other suggestions.

See commit history.

Added unit tests.

TODO:
- [x] Manual e2e tests to confirm that the problem in #1590 is resolved with 1 O + 1 B
- [ ] Manual e2e tests to confirm that the problem in #1590 is resolved with N Os (same address) + 1 B

Fixes #1590

**Checklist:**

### DIFF
--- a/pm/sendermonitor.go
+++ b/pm/sendermonitor.go
@@ -14,6 +14,10 @@ import (
 	"github.com/pkg/errors"
 )
 
+// The minimum value for the ratio between a sender's deposit and its pending amount required for the
+// pending amount to be ignored when calculating the sender's max float
+const minDepositPendingRatio = 3.0
+
 // unixNow returns the current unix time
 // This is a wrapper function that can be stubbed in tests
 var unixNow = func() int64 {
@@ -169,15 +173,33 @@ func (sm *LocalSenderMonitor) ValidateSender(addr ethcommon.Address) error {
 	return nil
 }
 
-// maxFloat is a helper that returns the sender's max float as:
-// reserveAlloc - pendingAmount
+// maxFloat is a helper that returns the sender's max float
+// If the sender's deposit pending ratio (deposit / pendingAmount) >= minDepositPendingRatio, then the max float is reserveAlloc
+// Otherwise, the max float is reserveAlloc - pendingAmount
 // Caller should hold the lock for LocalSenderMonitor
 func (sm *LocalSenderMonitor) maxFloat(addr ethcommon.Address) (*big.Int, error) {
 	reserveAlloc, err := sm.reserveAlloc(addr)
 	if err != nil {
 		return nil, err
 	}
-	return new(big.Int).Sub(reserveAlloc, sm.senders[addr].pendingAmount), nil
+
+	info, err := sm.smgr.GetSenderInfo(addr)
+	if err != nil {
+		return nil, err
+	}
+
+	pendingAmount := sm.senders[addr].pendingAmount
+	depositPendingRatio := big.NewFloat(0.0)
+	if pendingAmount.Cmp(big.NewInt(0)) > 0 {
+		depositPendingRatio.Quo(new(big.Float).SetInt(info.Deposit), new(big.Float).SetInt(pendingAmount))
+	}
+	// If the actual deposit pending ratio exceeds the minimum deposit pending ratio we ignore the sender's pendingAmount
+	// and return the sender's reserve allocation as the max float
+	if depositPendingRatio.Cmp(big.NewFloat(minDepositPendingRatio)) >= 0 {
+		return reserveAlloc, nil
+	}
+
+	return new(big.Int).Sub(reserveAlloc, pendingAmount), nil
 }
 
 func (sm *LocalSenderMonitor) reserveAlloc(addr ethcommon.Address) (*big.Int, error) {
--- a/pm/sendermonitor_test.go
+++ b/pm/sendermonitor_test.go
@@ -82,11 +82,64 @@ func TestMaxFloat(t *testing.T) {
 	}
 }
 
+func TestMaxFloat_MinDepositPendingRatio(t *testing.T) {
+	require := require.New(t)
+	assert := assert.New(t)
+
+	claimant, b, smgr, tm := localSenderMonitorFixture()
+	addr := ethcommon.BytesToAddress([]byte("foo"))
+
+	smgr.info[addr] = &SenderInfo{
+		Reserve: &ReserveInfo{
+			FundsRemaining:        big.NewInt(500),
+			ClaimedInCurrentRound: big.NewInt(0),
+		},
+	}
+	smgr.claimedReserve[addr] = big.NewInt(0)
+	tm.transcoderPoolSize = big.NewInt(1)
+
+	sm := NewSenderMonitor(claimant, b, smgr, tm, newStubTicketStore(), 5*time.Minute, 3600)
+
+	reserveAlloc := smgr.info[addr].Reserve.FundsRemaining
+	pendingAmount := big.NewInt(100)
+	// Set pendingAmount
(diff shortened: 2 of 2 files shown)


kyriediculous (approved):
This looks good to me for what the PR tries to accomplish. 

One additional optimisation might be to increase the default for `-depositMultiplier` from 1 to an higher number. This of course offers no guarantees but it does provide honest broadcasters with a more sensible default  for funds management.

yondonfu:
> One additional optimisation might be to increase the default for -depositMultiplier from 1 to an higher number. This of course offers no guarantees but it does provide honest broadcasters with a more sensible default for funds management.

Agreed that `-depositMultiplier` would help a broadcaster manage its deposit by rejecting orchestrators that require a face value that would drain the deposit too quickly. Although I think this utility is slightly related, but separate from the issue that this PR addresses (feel free to elaborate more if I'm missing something here).

PR #1593, opened 2020-07-30
mod: Update to latest LPMS with timestamp handling

### LINKED ISSUE #1591: Update LPMS dep to support custom timestamp handling for transcoding

Needed to support https://github.com/livepeer/lpms/pull/201

jailuthra:
Update LPMS to the latest commit (with custom timestamp handling)

- Update LPMS go module
- Fix breaking unit tests that compared raw transcoded segment size (this changes as our transcoded output frames/timestamps don't match 1-on-1 with ffmpeg anymore)

Fixes #1591 

**Checklist:**

- [x] README and other documentation updated (no updates needed)
- [x] Node runs in OSX and devenv (tested on linux)


kyriediculous:
This looks good to me although I have very little clue as to why the length differs, could you explain in video noob terms?

Additional thought unrelated to this PR: should we consider proper versioning for LPMS ?

jailuthra:
@kyriediculous 
> This looks good to me although I have very little clue as to why the length differs, could you explain in video noob terms?

The encoded segments' binary output is of different length now. This might change due to any process change in the underlying transcoder - like updating x264/fffmpeg version, or in the current case - changing how many frames get encoded and with what PTS values.

_Optional read_
Your question made me curious, as the output lengths were all noticeably smaller now, so I did a little digging around:
The input segment we're using (`core/test.ts`) has a weird jump in the frame-by-frame timestamps:
```
$ ffprobe -select_streams v -show_frames test.ts | grep pkt_pts= | less
...
pkt_pts=54090
pkt_pts=57060
pkt_pts=60030
pkt_pts=153180
pkt_pts=156150
...
```
We encode the [initial ~402KB](https://github.com/livepeer/go-livepeer/blob/1a92afcbf5fe96817bfbce3e6232719bb5897341/core/core_test.go#L28) of that file which is exactly 87 frames.

Earlier (w/ go-livepeer/master) because of the PTS jump we encoded these 87frames (29.97 fps) to 120frames (30fps) because of the jump in the input PTS. (The fps filter duplicated all those frames between the jumped timestamps)

Now (w/ latest LPMS) we transcode the 87-frames to exactly 87-frames, thus the file size is much shorter. We simply ignore the jumps in timestamps within a segment, and recalculate the starting timestamp of the next segment to prevent any drift issues like reported by Eli before.

**tl;dr:** ignoring weird PTS jump now, so we have less output frames - segment length is shorter now.

Also, unrelated but I can't see any video in `core/test.ts` at all while playing through ffmpeg/vlc/mpv. Is it just me? Should we use a different segment for testing? (`core/test2.ts` seems to work)

> Additional thought unrelated to this PR: should we consider proper versioning for LPMS ?

I'm not sure about this, is LPMS used directly by the public or only through go-livepeer? If go-livepeer is the only consumer versioning might not be needed, as LPMS master is usually quite stable and can be branched-off from any commit in the upcoming go-livepeer release.

yondonfu (approved):
LGTM :shipit:

yondonfu:
@jailuthra Thanks for writing up the explanation on why segment sizes are now shorter in the tests!

> Also, unrelated but I can't see any video in core/test.ts at all while playing through ffmpeg/vlc/mpv. Is it just me? Should we use a different segment for testing? (core/test2.ts seems to work)

I also can't see any video for `core/test.ts` when using ffplay - I just see a black screen although there is audio. Not sure if that is intentional or not. I'm cool with using `core/test2.ts` although I don't feel strongly about it either way - mainly because I think the go-livepeer tests primarily serve as sanity checks to make sure the LPMS transcode function is actually invoked properly with the actual checks for transcoding correctness living in LPMS.

> I'm not sure about this, is LPMS used directly by the public or only through go-livepeer? If go-livepeer is the only consumer versioning might not be needed, as LPMS master is usually quite stable and can be branched-off from any commit in the upcoming go-livepeer release.

👍  I'm cool with holding off on semver versioning LPMS until there is a stronger need.

kyriediculous (approved):
Thanks for explaining @jailuthra !

PR #1595, opened 2020-07-30
build: add --disable-nls to gnutls

iameli:
I was having trouble building go-livepeer since upgrading to Catalina, and adding `--disable-nls` to the gnutls build fixed it. I'm not really sure what's up with that but I don't think we're using NLS for anything.

A previous version of this PR also moved the ffmpeg build directory from the user's home directory to a `.gitignore`'d subdirectory. Then I realized that `~/compiled` is hardcoded in like 30 places in the code and decided not to do that today.

Ran B/O/T locally and everything seemed to work. I'll run a quick GPU test but gnutls only manages the ffmpeg-layer HTTPS connections so I think we're fine as long as the B<>O connection works at all.

**Checklist:**


darkdarkdragon:
Just checked this - in addition to `--disable-nls` , I'm also needed to add `--disable-guile` to successfully build ffmpeg.

thomshutt:

PR #1596, opened 2020-07-30
Redeem based on available funds and tx cost

### LINKED ISSUE #1580: Error inserting duplicate winning ticket into DB when trying to redeem
**Describe the bug**

A user reported the following error log for its orchestrator:

```
E0716 19:50:37.537155    8106 queue.go:133] Error redeeming err=failed inserting winning ticket sender=0xc3c7c4C8f7061B7d6A72766Eee5359fE4F36e61E: UNIQUE constraint failed: ticketQueue.sig
```

**To Reproduce**

I believe this is happening because:

1. The orchestrator advertised ticket params with face value = X to a broadcaster
2. The broadcaster's max float for the orchestrator decreased to Y s.t. Y < X
3. The orchestrator received a winning ticket
4. The orchestrator tries to redeem the winning ticket, but the broadcaster's max float cannot cover the ticket face value so the orchestrator [re-adds](https://github.com/livepeer/go-livepeer/blob/0723bac6873f09e08c71ad7bf5f98f0466db97fc/pm/sendermonitor.go#L311) the ticket to the queue which triggers an error because the ticket already exists in the DB

**Expected behavior**

The node should not try to insert a duplicate winning ticket into its DB.

I think this issue can be resolved by not trying to re-add a ticket to the DB during ticket redemption when the broadcaster's [max float is 0](https://github.com/livepeer/go-livepeer/blob/0723bac6873f09e08c71ad7bf5f98f0466db97fc/pm/sendermonitor.go#L300) or when the broadcaster's [max float is insufficient](https://github.com/livepeer/go-livepeer/blob/0723bac6873f09e08c71ad7bf5f98f0466db97fc/pm/sendermonitor.go#L309).

 - OS: [e.g. iOS]: Ubuntu
 - Version [e.g. 22]: 0.5.9

### LINKED ISSUE #1581: Orchestrator should redeem ticket even when face value exceeds broadcaster max float under certain conditions

An orchestrator currently only redeems a winning ticket if:

- The broadcaster's [max float is not 0](https://github.com/livepeer/go-livepeer/blob/master/pm/sendermonitor.go#L300)
- The broadcaster's [max float is insufficient to cover the ticket face value](https://github.com/livepeer/go-livepeer/blob/0723bac6873f09e08c71ad7bf5f98f0466db97fc/pm/sendermonitor.go#L309)

This can result in a situation where an orchestrator will let a winning ticket expire because the broadcaster's max float decreases (i.e. due to a increase in the # of active orchestrators) such that it is no longer sufficient to cover the ticket's face value. This could happen even if the broadcaster has enough funds in its deposit to cover the ticket face value. 

An orchestrator should redeem a winning ticket as long as the broadcaster's max float + the broadcaster's deposit is >= the ticket face value.

Note that there is a risk with this condition in that if the broadcaster's deposit is drained before the orchestrator's redemption tx confirms then the orchestrator may actually lose value with the redemption tx if the tx cost is not covered. The second condition described below would address this, but likely requires some code shuffling because the `LocalSenderMonitor` responsible for redeeming winning tickets does not currently have access to the estimated tx cost of a redemption (only the `Recipient` has access to this right now).

Other conditions worth considering:

- The ticket face value > the estimated tx cost because the estimated tx cost could be different at the time of redemption vs. the time at which the ticket face value was originally calculated
- The broadcaster's max float >= the estimated tx cost because this would mean that orchestrator is at least guaranteed that the the estimated tx cost will be covered by the redemption tx

yondonfu:

This PR implements [this spec](https://github.com/livepeer/go-livepeer/issues/1581#issuecomment-666750376).

2914f6a also adds a missing error check in `LocalSenderMonitor.reserveAlloc()` - updated the unit tests to make sure that they fail without the error check.

See commit history.

TODO:
- [x] Update unit tests
- [x] Manual e2e test to confirm that the problem in #1581 is resolved

Fixes #1581 
Fixes #1580 

**Checklist:**


pm/sendermonitor.go
```diff
@@ -216,6 +230,28 @@ func (sm *LocalSenderMonitor) reserveAlloc(addr ethcommon.Address) (*big.Int, er
 	return new(big.Int).Sub(new(big.Int).Div(reserve, poolSize), claimed), nil
 }
 
+// Returns the current available funds for a sender that could cover redemptions
+func (sm *LocalSenderMonitor) availFunds(addr ethcommon.Address) (*big.Int, error) {
```
kyriediculous: nit: Can we just name this `availableFunds`. Four letters don't seem to matter that much in terms of length.

yondonfu: Fixed 6e568b4

pm/sendermonitor.go
```diff
@@ -313,26 +349,24 @@ func (sm *LocalSenderMonitor) cleanup() {
...
-	// the ticket to be retried later
-	if maxFloat.Cmp(ticket.Ticket.FaceValue) < 0 {
-		if err := sm.QueueTicket(ticket); err != nil {
-			return nil, err
-		}
-		return nil, fmt.Errorf("insufficient max float sender=%v faceValue=%v maxFloat=%v", ticket.Ticket.Sender.Hex(), ticket.Ticket.FaceValue, maxFloat)
+	// We only submit a redemption if availableFunds covers the redemption tx cost
+	// Otherwise, we return an error so we can try the redemption later
+	txCost := new(big.Int).Mul(big.NewInt(int64(sm.cfg.RedeemGas)), gasPrice)
+	if availableFunds.Cmp(txCost) <= 0 {
```
kyriediculous, line 368: This will leave the ticket as the earliest winning ticket for the orchestrator and not mark it as redeemed.

This ticket will then be tried to be redeemed forever, so it seems like we'd need https://github.com/livepeer/go-livepeer/pull/1598 as well for a full robust solution.

yondonfu: > This will leave the ticket as the earliest winning ticket for the orchestrator and not mark it as redeemed.

Yep thanks for pointing that out!

Worth noting that this is behavior is not introduced by this PR and this PR just changes the condition under which the earliest winning ticket would be left at the head of the queue.

That being said, it is worthwhile acknowledging this so we can make sure to address this in other PRs such as #1598. We should probably continue retrying the ticket unless we know for certain that it would never be redeemable i.e. it is now expired.

kyriediculous:
PR looks good to me ! Let's rebase

One last note we currently would redeem a ticket even if the gains over the txcost would be minimal (one or a couple of wei). I Don't see this as a big issue that we should address and as you mentioned in the spec designing for simplicity here is okay.

yondonfu:
Rebased.

kyriediculous (commented):
Were you planning on running the manual tests described in the OP still or has that happened already?

yondonfu:
4cb0c9b fixes a typo in the WinningTicketTransfer event signature. This typo previously resulted in an invalid topic hash and caused the block watcher to skip over WinningTicketTransfer events when aggregating logs for a block. I discovered this bug while running the manual test described below - I observed that O never updated its view of B's available funds after a ticket was redeemed. There might be a way to ensure that the event signatures used to generate topics for the block watcher are correct i.e. pull the event signatures we care about directly from the contract ABIs based on the events we are watching for in the watchers in `eth/watchers`, but I opted not to dive into that for now.

> Were you planning on running the manual tests described in the OP still or has that happened already?

Just completed a manual test. The easiest way I could think of testing these changes in an e2e workflow was to make the following modifications:

```diff
diff --git a/eth/client_ticketbroker.go b/eth/client_ticketbroker.go
index 8bfe7df0..208bbb7b 100644
--- a/eth/client_ticketbroker.go
+++ b/eth/client_ticketbroker.go
@@ -1,6 +1,7 @@
 package eth

 import (
+       "context"
        "math/big"
        "strings"

@@ -17,6 +18,18 @@ import (
 // value to the sum of the provided deposit and penalty escrow amounts
 func (c *client) FundDepositAndReserve(depositAmount, reserveAmount *big.Int) (*types.Transaction, error) {
        opts := c.TicketBrokerSession.TransactOpts
+
+       redeemGas := 250000
+       gasPrice, err := c.backend.SuggestGasPrice(context.Background())
+       if err != nil {
+               return nil, err
+       }
+       txCost := new(big.Int).Mul(big.NewInt(int64(redeemGas)), gasPrice)
+
+       // depositAmount + reserveAmount = txCost + 2
+       depositAmount = new(big.Int).Add(txCost, big.NewInt(1))
+       reserveAmount = big.NewInt(1)
+
        opts.Value = new(big.Int).Add(depositAmount, reserveAmount)

        return c.TicketBrokerSession.Contract.FundDepositAndReserve(&opts, depositAmount, reserveAmount)
diff --git a/pm/recipient.go b/pm/recipient.go
index ec5cae7a..f90c9821 100644
--- a/pm/recipient.go
+++ b/pm/recipient.go
@@ -183,6 +183,7 @@ func (r *recipient) TicketParams(sender ethcommon.Address, price *big.Rat) (*Tic
        if err != nil {
                return nil, err
        }
+       // faceValue := new(big.Int).Add(r.txCost(), big.NewInt(1))

        lastBlock := r.tm.LastSeenBlock()
        expirationBlock := new(big.Int).Add(lastBlock, paramsExpirationBlock)
@@ -217,6 +218,8 @@ func (r *recipient) txCost() *big.Int {
 }

 func (r *recipient) faceValue(sender ethcommon.Address) (*big.Int, error) {
+       return new(big.Int).Add(r.txCost(), big.NewInt(1)), nil
+
        // faceValue = txCost * txCostMultiplier
        faceValue := new(big.Int).Mul(r.txCost(), big.NewInt(int64(r.cfg.TxCostMultiplier)))

diff --git a/pm/sender.go b/pm/sender.go
index 6192f48e..8b2a7a8d 100644
--- a/pm/sender.go
+++ b/pm/sender.go
@@ -81,6 +81,7 @@ func (s *sender) EV(sessionID string) (*big.Rat, error) {
 }

 func (s *sender) validateSender(info *SenderInfo) error {
+       return nil
        maxWithdrawRound := new(big.Int).Add(s.timeManager.LastInitializedRound(), big.NewInt(1))
        if info.WithdrawRound.Int64() != 0 && info.WithdrawRound.Cmp(maxWithdrawRound) != 1 {
                return ErrSenderValidation{fmt.Errorf("unable to validate sender: deposit and reserve is set to unlock soon")}
```

These modifications allowed me to run the following manual test (used the devtool):

- Set B's deposit to txCost + 1
- Set B's reserve to 1
- As a result of the above, B's initial available funds is txCost + 2
- Set O's face value to txCost + 1
- B sends O a winning ticket
- O redeems the ticket because B's available funds > txCost
- B sends O another winning ticket
- B's available funds is now 1
- O does not redeem the ticket because availableFunds < txCost

cmd/livepeer/livepeer.go
```diff
@@ -552,7 +552,15 @@ func main() {
 				}
 				sm = rc
 			} else {
-				sm = pm.NewSenderMonitor(recipientAddr, n.Eth, senderWatcher, timeWatcher, n.Database, cleanupInterval, smTTL)
+				smCfg := &pm.LocalSenderMonitorConfig{
```
kyriediculous: Would it hurt avoiding duplicating this on line L628 and just initialising the config regardless but only use it  in those two cases?

kyriediculous:
> 4cb0c9b fixes a typo in the WinningTicketTransfer event signature

Oh good find indeed... A more robust solution here is definitely to be considered, unit testing didn't catch this because there we use the right signature. 

The manual test scenario looks good to me.

I left one tiny comment but I'll leave it up to you if you want to address it or not, if not lmk and I'll approve this PR as is.

cmd/livepeer/livepeer.go
yondonfu: Fixed in bae3a83

kyriediculous:
Changes look good, let's rebase !

yondonfu:
Rebased

kyriediculous (approved):
LGTM 👍

PR #1597, opened 2020-08-02
server: make broadcaster and orchestrator config endpoint form values optional

kyriediculous:

This pull requests allows the form values passed into the `setOrchestratorConfig` and `setBroadcasterConfig` endpoints to be optional. 

if a field is not provided then its related configuration will not be altered on the node. 

This PR is a requirement for https://github.com/Livepool-io/livepeer-usd-pricer

- [x] manual testing 

**Checklist:**

### DIFF
--- a/server/webserver.go
+++ b/server/webserver.go
@@ -2,7 +2,7 @@ package server
 
 import (
 	"encoding/json"
-	"errors"
+
 	"flag"
 	"fmt"
 	"math/big"
@@ -25,6 +25,7 @@ import (
 	lpTypes "github.com/livepeer/go-livepeer/eth/types"
 	"github.com/livepeer/go-livepeer/monitor"
 	ffmpeg "github.com/livepeer/lpms/ffmpeg"
+	"github.com/pkg/errors"
 )
 
 var vFlag *glog.Level = flag.Lookup("v").Value.(*glog.Level)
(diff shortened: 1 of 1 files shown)


server/webserver.go
```diff
@@ -90,53 +90,51 @@ func (s *LivepeerServer) cliWebServerHandlers(bindAddr string) *http.ServeMux {
...
-		}
 
-		var price *big.Rat
-		if pr > 0 {
-			price = big.NewRat(pr, px)
-		}
+		if pricePerUnit != "" && pixelsPerUnit != "" {
+			pr, err := strconv.ParseInt(pricePerUnit, 10, 64)
+			if err != nil {
+				glog.Errorf("Error converting string to int64: %v\n", err)
```
darkdarkdragon: Could we return error here like this:
```golang
				msg := fmt.Sprintf("Error converting string to int64: %v\n", err)
				glog.Error(msg)
				w.WriteHeader(http.StatusBadRequest)
				w.Write([]byte(msg))
```
?
And in other places?

server/webserver.go
```diff
@@ -375,7 +379,7 @@ func (s *LivepeerServer) cliWebServerHandlers(bindAddr string) *http.ServeMux {
 			return
 		}
 
-		if t.RewardCut.Cmp(eth.FromPerc(blockRewardCut)) != 0 || t.FeeShare.Cmp(eth.FromPerc(feeShare)) != 0 {
+		if feeShareStr != "" && blockRewardCutStr != "" && t.RewardCut.Cmp(eth.FromPerc(blockRewardCut)) != 0 || t.FeeShare.Cmp(eth.FromPerc(feeShare)) != 0 {
```
darkdarkdragon:
```suggestion
		if feeShareStr != "" && blockRewardCutStr != "" && (t.RewardCut.Cmp(eth.FromPerc(blockRewardCut)) != 0 || t.FeeShare.Cmp(eth.FromPerc(feeShare))) != 0 {
```

darkdarkdragon (changes requested):

server/webserver.go
kyriediculous: fixed in 47f9815c

server/webserver.go
kyriediculous: fixed in 47f9815c

server/webserver.go
```diff
@@ -85,58 +86,66 @@ func (s *LivepeerServer) cliWebServerHandlers(bindAddr string) *http.ServeMux {
 	//Set the broadcast config for creating onchain jobs.
 	mux.HandleFunc("/setBroadcastConfig", func(w http.ResponseWriter, r *http.Request) {
 		if err := r.ParseForm(); err != nil {
-			glog.Errorf("Parse Form Error: %v", err)
+			err = errors.Wrapf(err, "Parse form error")
+			glog.Error(err)
+			respondWith500(w, err.Error())
```
darkdarkdragon: `respondWith400` would be more appropriate?  Problem is with user's data, so it is a `BadRequest`, not `InternalServerError`

darkdarkdragon (changes requested):

server/webserver.go
kyriediculous: fixed in 808250b for all user input related errors.

darkdarkdragon (approved):
LGTM

PR #1598, opened 2020-08-02
db,pm: avoid redeeming expired tickets

### LINKED ISSUE #632: pm: recipient redeems old valid winning tickets on start up

### LINKED ISSUE #1594: Orchestrator tries to redeem expired tickets
**Describe the bug**
There is no clause in the `SelectEarliestWinningTicket()` SQL query to prevent expired tickets from being selected. We currently just select the earliest winning ticket but didn't assume stale tickets could still exist in the database.

This will result in an always failing transaction. 

This can occur in case the orchestrator didn't have enough funds available to redeem the ticket within it's validity period.

A potential solution would be to do a check in the SQL query since we have access to the creation round.  

**Expected behavior**
Don't query expired tickets 

**Screenshots**
image

kyriediculous:
This PR avoids the redemption of expired winning tickets.

- Change signatures of `SelectEarliestWinningTicket` and `WinningTicketCount` to take in a `minCreationRound` parameter. The `creationRound` for tickets stored would have to be greater or equal to this value to be considered a non-expired ticket (the expiry is two rounds)
- Adjust the respective SQL queries
- Check whether a ticket is already redeemed before trying to redeem a ticket, mark it as redeemed if that's the case 7b3a95c0

Added unit tests , cherry picked this onto the Livepool orchestrator and observe my stale ticket is no longer being retried endlessly.

Fixes #1594 
Fixes #632

**Checklist:**

### DIFF
--- a/pm/queue.go
+++ b/pm/queue.go
@@ -4,7 +4,6 @@ import (
 	"math/big"
 
 	ethcommon "github.com/ethereum/go-ethereum/common"
-	"github.com/ethereum/go-ethereum/event"
 	"github.com/golang/glog"
 )
 
@@ -32,9 +31,7 @@ type redemption struct {
 //
 // Based off of: https://github.com/lightningnetwork/lnd/blob/master/htlcswitch/queue.go
 type ticketQueue struct {
-	// blockSub returns a subscription to receive the last seen block number
-	blockSub func(chan<- *big.Int) event.Subscription
-
+	tm TimeManager
 	// redeemable is a channel that a queue consumer will receive
 	// redeemable tickets on as a sender's max float becomes
 	// sufficient to cover the face value of tickets
@@ -46,11 +43,11 @@ type ticketQueue struct {
 	quit chan struct{}
 }
 
-func newTicketQueue(store TicketStore, sender ethcommon.Address, blockSub func(chan<- *big.Int) event.Subscription) *ticketQueue {
+func newTicketQueue(sender ethcommon.Address, sm *LocalSenderMonitor) *ticketQueue {
 	return &ticketQueue{
-		blockSub:   blockSub,
+		tm:         sm.tm,
 		redeemable: make(chan *redemption),
-		store:      store,
+		store:      sm.ticketStore,
 		sender:     sender,
 		quit:       make(chan struct{}),
 	}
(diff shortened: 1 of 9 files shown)


kyriediculous:
@jailuthra I just tagged you in case you might be interested in these areas of the codebase, please don't feel obligated to review this !

pm/queue.go
```diff
@@ -80,7 +77,7 @@ func (q *ticketQueue) Redeemable() chan *redemption {
 
 // Length returns the current length of the queue
 func (q *ticketQueue) Length() (int, error) {
-	return q.store.WinningTicketCount(q.sender)
+	return q.store.WinningTicketCount(q.sender, new(big.Int).Sub(q.tm.LastInitializedRound(), big.NewInt(2)).Int64())
```
yondonfu: Let's define 2 as a constant `ticketValidityPeriod` so we can just use that everywhere it is needed?

yondonfu (changes requested):
Looks good just one small comment.

pm/queue.go
kyriediculous: fixed in f0ed1057

yondonfu (approved):
LGTM after squashing fixup commits

PR #1599, opened 2020-08-05
eth: fix reward call failures

### LINKED ISSUE #1568: reward call fails due to incredibly high gas prices when retrying transactions
**Describe the bug**
Multiple users running against self-operated Ethereum nodes (e.g. geth) have reported failing reward transactions and requiring a node restart to fix.

kyriediculous:

Alleviate some issues users have been experiencing with reward calls 

- Avoid nonce gap issues 
- Avoid multiple tx in flight 
- 

Added a unit tests that confirms nonce gaps are not possible

Fixes #1568 

**Checklist:**

### DIFF
--- a/eth/backend_test.go
+++ b/eth/backend_test.go
@@ -1,11 +1,16 @@
 package eth
 
 import (
+	"context"
+	"crypto/ecdsa"
+	"log"
 	"math/big"
 	"testing"
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/ethereum/go-ethereum/ethclient"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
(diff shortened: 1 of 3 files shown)


eth/backend_test.go
```diff
@@ -45,3 +50,46 @@ func TestNewTxLog(t *testing.T) {
 	txLog, err = b.newTxLog(tx)
 	assert.EqualError(err, "no method signature")
 }
+
+func TestSendTransaction_SendErr_DontUpdateNonce(t *testing.T) {
+	// client := &ethclient.Client{}
+	client, err := ethclient.Dial("https://rinkeby.infura.io")
```
yondonfu, line 56: Looks like there might be a way to generate a test chain - see the [ethclient tests](https://github.com/ethereum/go-ethereum/blob/2de37f28e069b059d6a8fcf546ecf307373e2419/ethclient/ethclient_test.go#L187) from go-ethereum. Haven't tried it, but if it works this could make writing/running this test easier since the ethclient instance wouldn't need to interact with an actual ETH RPC provider.

kyriediculous: Sadly they are unexported and they seem to be more "chain/block propagation" related whereas we need a simple RPC error. 

Given the urgency for this fix would you be okay keeping the test as is and exploring a more robust solution for this in the future? 

I can see this end up being flaky in the following cases:
- There is no internet connection 
- The endpoint is down ( == the entire rinkeby cluster for infura is down)

yondonfu: Ah actually I just realized that a requirement for this test is to trigger an error when `ethclient.SendTransaction()` is called and this is already accomplished because Infura will return an error due to a missing project ID. Sounds good to me!

eth/eventservices/rewardservice.go
```diff
@@ -120,22 +120,18 @@ func (s *RewardService) tryReward() error {
 			if err != nil {
 				return err
 			}
+			s.pendingTx = tx
```
yondonfu: If we successfully submit a replacement tx on L108, then that tx will be the latest pending tx so any additional gas price bumps would need to be calculated based off of the gas price for that tx. If we only set the pending tx here then I think the following is possible:

1. Submit original tx with nonce N with gas price X
2. Original tx with nonce N is not confirmed within the timeout
3. Submit replacement tx 1 with nonce N with gas price 1.1X
4. Replacement tx 1 with nonce N is not confirmed within the timeout
5. Submit replacement tx 2 with nonce N with gas price 1.1X
    - Note that the gas price is 1.1X since it is calculated off of the gas price for the original tx instead of the gas price for replacement tx 1 (which would be 1.21X)

In order to avoid the above scenario, we should be able to set the pending tx both after a successful `Reward()` call and after a successful `ReplaceTransaction()` call i.e.

```golang
if s.pendingTx != nil {
    tx, err = s.client.ReplaceTransaction(s.pendingTx, "reward", nil)
    ...
} else {
    tx, err = s.client.Reward()
    ...
}

s.pendingTx = tx
```

yondonfu (changes requested):

eth/eventservices/rewardservice.go
kyriediculous: Yeah acknowledged, although I don't think speeding up more than once is the right strategy here though (cancellations might be cheaper and better in that case).

Given urgency though let's roll with what you propose and consider additional strategies down the line. WDYT?

kyriediculous: fixed in b06500f1

yondonfu (approved):
LGTM after a rebase

kyriediculous:
rebased 

Edit: oh already approved, I'll just wait for green CI then 😃

PR #1606, opened 2020-08-18
Fix incorrect Minter param values after LIP-40 in livepeer_cli

### LINKED ISSUE #1605: livepeer_cli displays incorrect Minter parameter values after LIP-40 deployment on Rinkeby
**Describe the bug**

After [LIP-40](https://github.com/livepeer/LIPs/blob/master/LIPs/LIP-40.md) was deployed on Rinkeby (LPSFT-0), livepeer_cli shows the following when selecting the option "2. View protocol parameters":

```
*--------------------------------*--------------------------------*
|                  Inflation (%) |                           0.05 |
*--------------------------------*--------------------------------*
|            InflationChange (%) |                           0.05 |
*--------------------------------*--------------------------------*
|          TargetBondingRate (%) |                          80000 |
*--------------------------------*--------------------------------*
```

These values are incorrect. Inflation should be .00005, InflationChange should .00005 and TargetBondingRate should be 80%. livepeer_cli shows these incorrect values because the [FormatPerc()](https://github.com/livepeer/go-livepeer/blob/b4b77f915834bca5d938b3125c44176c709dc349/cmd/livepeer_cli/wizard_stats.go#L121) helper is used to convert the on-chain value which is out of a constant PERC_DIVISOR into a percentage. FormatPerc() assumes that PERC_DIVISOR is 1000000, but LIP-40 changes this value to 1000000000 specifically for the Minter contract (and no other contracts). So this assumption causes the calculated percentage to be incorrect for the parameters for the Minter contract.

This will also be the case on mainnet once LIP-40 is deployed there.

livepeer_cli should be updated to show the correct values for the Minter parameters. Note that different values for PERC_DIVISOR will need to be assumed for the Minter (i.e. 1000000000) vs. other contracts (i.e. 1000000).

yondonfu:

See #1605 for background.

This PR adds an additional percentage helper `FormatPercMinter()` in the `eth` package that is used specifically to convert percentage values for the Minter because after LIP-40 the Minter uses a different PERC_DIVISOR value for calculating percentages than the other contracts. This helper is used when displaying the Minter params in livepeer_cli.

See commit history.

Added unit tests for `FormatPercMinter()`.

Also tested manually by running livepeer_cli while connected to Rinkeby - the correct Minter param values are shown below:

```
*--------------------------------*--------------------------------*
|                  Inflation (%) |                         0.0001 |
*--------------------------------*--------------------------------*
|            InflationChange (%) |                          5e-05 |
*--------------------------------*--------------------------------*
|          TargetBondingRate (%) |                             80 |
```

Fixes #1605 

**Checklist:**


kyriediculous (approved):
LGTM 👍

Issue #1569, opened 2020-07-10
Transcoding Protocol Authentication


j0sh:
## Abstract

Currently, there is no authentication for the transcoding network protocol. A method of authentication is proposed for the transcoding network protocol which re-introduces an [authentication token](https://github.com/livepeer/go-livepeer/commit/fae70f384ab151b2f2c50495f3c406c706ef0412) that is acquired during discovery. This authentication scheme will also be useful for broadcasters running [Livepeer ingest](https://gist.github.com/j0sh/df016ea5da7e4e368e4f648faa9d4a32) ("proxy mode") and helps resolve some other [long-standing](https://github.com/livepeer/go-livepeer/issues/695) issues.

## Motivation

There are a few scenarios that might want access to be checked:

* Staging / testing environments that may be otherwise publicly accessible
* Livepeer.com hosted service - for both hosted broadcasters (proxy - authenticating customers) and orchestrators (for usage by internal / trusted Bs)
* Other types of private networks
* Onchain orchestrators that may want to implement an allow-list or deny-list of broadcaster addresses
* External rate limiting of jobs or certain addresses

## Proposed Solution

* Configure offchain broadcasters with an identifier. Use this value in the [`OrchestratorRequest.address`](https://github.com/livepeer/go-livepeer/blob/9fd21617aee0d73ac10e191ab4b6c5b33789c1dc/net/lp_rpc.proto#L29-L32) RPC field. This could be an API key, for example. If omitted, the value can remain empty.
* For onchain broadcasters, this identifier remains the ETH address.
* During [OrchestratorInfo handling](https://github.com/livepeer/go-livepeer/blob/9fd21617aee0d73ac10e191ab4b6c5b33789c1dc/server/rpc.go#L245-L253), make a webhook call if a webhook is configured. The configuration can reuse the `-authWebhookUrl` flag.
* The `url` field of the webhook request can be something like `livepeer://<broadcaster-address>` , where the address is encoded, maybe as hexadecimal (Eth standard), base64 (more compact), base32 (human friendly), etc. Optionally, it might help someday to distinguish the modes in use, eg proxy or on-chain, such as `livepeer+proxy://<broadcaster-address>` (B in proxy mode) or `livepeer+mainnet://<broadcaster-address>` (O on mainnet) or `livepeer+proxy+rinkeby://<broadcaster-address>` (B in proxy mode using Rinkeby).
[...]

PR #1613, opened 2020-09-15
Add auth token

### LINKED ISSUE #695: Transcode loop is indexed by ManifestID

yondonfu:

This PR re-introduces an auth token into the B <> O transcoding protocol. The auth token was previously removed in https://github.com/livepeer/go-livepeer/commit/fae70f384ab151b2f2c50495f3c406c706ef0412. It is re-introduced in this PR because:

- Allows O to replace use a sessionID that it generates on its own instead of a B supplied manifestID for internal bookkeeping. This resolves long standing concerns around using B supplied manifestIDs for internal bookkeeping which are described in #695.
- Avoids manifestID collisions for transcode sessions on O when HTTP push clients retry with different Bs using the same manifestID. Now, retrying with a different B will result in a distinct transcode session even if B uses the same O as a previous B.
- Unifies O's external storage location for source and transcoded segments - they are all stored with the same sessionID prefix now. See [this doc](https://hackmd.io/0C2Zi0VVSyGEgN_4lEu_WA) for previous notes on this topic.
- Lays the groundwork for adding [transcoding protocol authentication](https://github.com/livepeer/go-livepeer/issues/1569#issue-655054118) in the future.

At a high level, the auth token is used in the following manner:

- O generates a fresh auth token during discovery in response to a GetOrchestrator call which includes an HMAC over all of the data in the auth token which can by used by O for validation at segment request time
- The auth token validity period is currently hardcoded to 30 minutes
- In on-chain mode, B tracks stream balances (for payments) using the auth token sessionID instead of its own manifestID
- B includes the auth token with all segment requests sent to O
- If the auth token is expired, B sends a GetOrchestrator request to the O to refresh the auth token
- O validates the auth token and checks that it is not expired
- O uses the auth token sessionID for all internal bookkeeping instead of the B supplied manifestID

A few additional notes:

- B's manifestID is still preserved. It is still included in logs on both B & O so that a single manifestID that is used on one or many Bs can be associated with different sessionIDs generated by Os.
- The auth token is added as a **mandatory capability**. This means that this represents a breaking networking protocol change. Newer Os will *not* work with older Bs because older Bs will not support the auth token capability. Similarly, newer Bs will *not* work with older Os because older Os will not support the auth token capability. This decision was made because a) using B supplied manifestIDs for bookkeeping on O enables undesirable behavior so I don't think we want to support this b) a breaking change is cleaner as it avoids supporting both B supplied manifestIDs and auth token sessionIDs for bookkeeping on B & O.

See commit history.

Added unit tests and did manual testing in on-chain and off-chain mode. Modified the hardcoded auth token validity period to 20 seconds to test the auth token refresh flow.

Fixes #695 

**Checklist:**


core/orchestrator.go
```diff
@@ -320,6 +323,18 @@ func (orch *orchestrator) Capabilities() *net.Capabilities {
 	return orch.node.Capabilities.ToNetCapabilities()
 }
 
+func (orch *orchestrator) AuthToken(sessionID string, expiration int64) *net.AuthToken {
+	h := hmac.New(sha256.New, orch.secret[:])
```
darkdarkdragon: What is the point of `[:]` syntax here?

server/segment_rpc.go
```diff
@@ -296,7 +298,22 @@ func verifySegCreds(orch Orchestrator, segCreds string, broadcaster ethcommon.Ad
...
+	if !bytes.Equal(verifyToken.Token, segData.AuthToken.Token) {
+		return nil, errors.New("invalid auth token")
+	}
+
+	expiration := time.Unix(segData.AuthToken.Expiration, 0)
+	if time.Now().After(expiration) {
+		return nil, errors.New("expired auth token")
+	}
+
+	if err := orch.CheckCapacity(core.ManifestID(md.AuthToken.SessionId)); err != nil {
```
darkdarkdragon: I think using `segData.AuthToken` here would be better for consistency

server/broadcast.go
```diff
@@ -672,3 +677,29 @@ func refreshSession(sess *BroadcastSession) (*BroadcastSession, error) {
 
 	return updateSession(sess, res), nil
 }
+
+func shouldRefreshSession(sess *BroadcastSession) (bool, error) {
+	if sess.OrchestratorInfo.AuthToken == nil {
+		return false, errors.New("missing auth token")
+	}
+
+	if time.Now().After(time.Unix(sess.OrchestratorInfo.AuthToken.Expiration, 0)) {
```
darkdarkdragon: Maybe we should do something like this:
```go
	if time.Now().After(time.Unix(sess.OrchestratorInfo.AuthToken.Expiration-60, 0)) {
```
so we will refresh token slightly before it expires?

server/broadcast.go
```diff
@@ -252,8 +253,7 @@ func selectOrchestrator(n *core.LivepeerNode, params *core.StreamParameters, cou
 		bcastOS := params.OS
 		if bcastOS.IsExternal() {
 			// Give each O its own OS session to prevent front running uploads
-			pfx := fmt.Sprintf("%v/%v", params.ManifestID, core.RandomManifestID())
-			bcastOS = drivers.NodeStorage.NewSession(pfx)
+			bcastOS = drivers.NodeStorage.NewSession(tinfo.AuthToken.SessionId)
```
darkdarkdragon: I think it would be good to do
```go
			pfx := fmt.Sprintf("%v/%v", params.ManifestID, tinfo.AuthToken.SessionId)
			bcastOS = drivers.NodeStorage.NewSession(pfx)
```
that way it would be easier to reason about data stored in the broadcaster's OS

darkdarkdragon (changes requested):
couple of notes, otherwise looks good!

core/orchestrator.go
yondonfu: Ah that is not needed here. It was leftover from when `orch.secret` was a `[32]byte`, but now it is a `[]byte`.

Fixed in f286b0bcaee039954051dcacaa66e259751aad04

server/segment_rpc.go
yondonfu: Fixed in 3b0c5b65c1aeedfddc354f72838bf8a7c970cc4b

server/broadcast.go
yondonfu: Good call. I elected to refresh the token when it is within the last 10% of the token's valid period in 23f0779587a2b60dca95e1d3e0bb7de46236eca2.

server/broadcast.go
yondonfu: Ah yeah agreed that this would make it easier to reason about data stored in the broadcaster's OS. Fixed in 0b6362978236ab43dd5491a7c7f359d5a303b2f3.

darkdarkdragon (approved):
LGTM

jailuthra (approved):
LGTM :ship: 

Slightly confusing to read the codebase now that we seem to have both (?) `ManifestID` (for client <> B) and `SessionID` (for B <> O and bookkeeping) but some functions/types are still called `ManifestID` that are used for both. But we can do it later as it is a non-functional change.

yondonfu:
> Slightly confusing to read the codebase now that we seem to have both (?) ManifestID (for client <> B) and SessionID (for B <> O and bookkeeping) but some functions/types are still called ManifestID that are used for both. But we can do it later as it is a non-functional change.

Yeah definitely agreed that it is a bit confusing. Since this PR was getting a bit big, I elected to not also change a bunch of existing function signatures that reference the ManifestID type, but I do think it is worth cleaning up separately.

j0sh:
Woohoo! 🎉

PR #1616, opened 2020-09-17
Enable free transcoding for on-chain mode when price is 0

### LINKED ISSUE #1615: Allow on-chain Os to transcode for free when price is set to 0

There are a few scenarios where it could be desirable for an on-chain O to transcode for free:

- The user operating the O is also operating the B. The user might not need payments to be exchanged between B & O in this case
- An O operator might want the flexibility to offer transcoding for free if it has some external incentive to do
- In the future, it might be useful for an O operator to be able to selectively charge some Bs for transcoding, but offer transcoding for free to other Bs. See [this comment](https://github.com/livepeer/go-livepeer/issues/1569#issuecomment-657316909) for an example of such a scenario

Changes required to support this:

- Allow O to set its price to 0 via `-pricePerUnit`. There is currently a [check](https://github.com/livepeer/go-livepeer/blob/master/cmd/livepeer/livepeer.go#L518) preventing that.
- When the price is 0, the ticket params advertised by O can have faceValue = 0 and winProb = 0 since the EV of a ticket can be 0 (since no payments will be necessary). This would override the [max float calculation](https://github.com/livepeer/go-livepeer/blob/master/pm/recipient.go#L234) that typically happen when generating ticket params. A side effect of this change is that on-chain transcoding with price = 0 can serve as an alternative to off-chain transcoding which would require both B and O to be restarted in off-chain mode. This should simplify node management for anyone that is running both Bs & Os that talk to each other.
- The [validation step](https://github.com/livepeer/go-livepeer/blob/ce63d6c5e7ef9da33c05b55fed7b573f0cf4adf4/pm/sender.go#L83) run by B should be loosened. It should be fine for the deposit and/or the reserve to be 0 if the ticket params have faceValue = 0 and winProb = 0 because that would imply that no payments need to be sent and that no deposit and/or reserve is required

A node operator can just switch to off-chain mode, but this is tedious and also does not provide the other benefits described earlier in this issue. Furthermore, *not* requiring a switch to off-chain mode just to transcode for free narrows the gap between on-chain mode and off-chain mode which I think is positive.

yondonfu:

This PR enables on-chain Os to transcode for free when their price is set to 0. See #1615 for a few reasons why this can be useful.

Additional implications of this PR include:

- The transcoding workflow in on-chain mode when price = 0 is pretty much equivalent to the transcoding workflow in off-chain mode
- Since an O's price can be updated via the webserver endpoint, it's possible to toggle payments on/off not only while the node is running, but during a stream as well. So, if there is a problem with the payment workflow for whatever reason O's price can be set to 0 to disable payments.

See commit history.

Updated unit tests.

I've done some manual testing using the devtool scripts + a private on-chain network that has consisted of:

- B with deposit & reserve > 0 streaming into O with price = 0
- B with deposit & reserve = 0 streaming into O with price = 0
- B with deposit & reserve = 0 streaming into O with price = 0 and then O sets price > 0 causing transcoding to stop
- B with deposit & reserve = 0 failing to stream into O with price > 0 and then O sets price = 0 causing transcoding to start

Fixes #1615 

**Checklist:**

### DIFF
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -513,11 +513,14 @@ func main() {
 			// Set price per pixel base info
 			if *pixelsPerUnit <= 0 {
 				// Can't divide by 0
-				panic(fmt.Errorf("The amount of pixels per unit must be greater than 0, provided %d instead\n", *pixelsPerUnit))
+				panic(fmt.Errorf("-pixelsPerUnit must be > 0, provided %d", *pixelsPerUnit))
 			}
-			if *pricePerUnit <= 0 {
-				// Prevent orchestrator from unknowingly provide free transcoding
-				panic(fmt.Errorf("Price per unit of pixels must be greater than 0, provided %d instead\n", *pricePerUnit))
+			if !isFlagSet["pricePerUnit"] && *pricePerUnit == 0 {
+				// Prevent orchestrators from unknowingly providing free transcoding
+				panic(fmt.Errorf("-pricePerUnit must be set"))
+			}
+			if *pricePerUnit < 0 {
+				panic(fmt.Errorf("-pricePerUnit must be >= 0, provided %d", *pricePerUnit))
 			}
 			n.SetBasePrice(big.NewRat(int64(*pricePerUnit), int64(*pixelsPerUnit)))
 			glog.Infof("Price: %d wei for %d pixels\n ", *pricePerUnit, *pixelsPerUnit)
--- a/pm/recipient.go
+++ b/pm/recipient.go
@@ -36,7 +36,7 @@ type Recipient interface {
 	// for a provided sender ETH adddress
 	TicketParams(sender ethcommon.Address, price *big.Rat) (*TicketParams, error)
 
-	// TxCostMultiplier returns the multiplier -
+	// TxCostMultiplier returns the tx cost multiplier for an address
 	TxCostMultiplier(sender ethcommon.Address) (*big.Rat, error)
 
 	// EV returns the recipients EV requirement for a ticket as configured on startup
@@ -179,9 +179,14 @@ func (r *recipient) TicketParams(sender ethcommon.Address, price *big.Rat) (*Tic
 
 	seed := new(big.Int).SetBytes(randBytes)
 
-	faceValue, err := r.faceValue(sender)
-	if err != nil {
-		return nil, err
+	faceValue := big.NewInt(0)
+	// If price is 0 face value, win prob and EV are 0 because no payments are required
+	if price.Num().Cmp(big.NewInt(0)) > 0 {
+		var err error
+		faceValue, err = r.faceValue(sender)
+		if err != nil {
+			return nil, err
+		}
 	}
 
 	lastBlock := r.tm.LastSeenBlock()
--- a/pm/sender.go
+++ b/pm/sender.go
@@ -146,17 +146,29 @@ func (s *sender) ValidateTicketParams(ticketParams *TicketParams) error {
 
 // validateTicketParams checks if ticket params are acceptable for a specific number of tickets
 func (s *sender) validateTicketParams(ticketParams *TicketParams, numTickets int) error {
+	if ticketParams.ExpirationBlock.Int64() == 0 {
+		return nil
+	}
+
+	latestBlock := s.timeManager.LastSeenBlock()
+	if ticketParams.ExpirationBlock.Cmp(latestBlock) <= 0 {
+		return ErrTicketParamsExpired
+	}
+
+	ev := ticketEV(ticketParams.FaceValue, ticketParams.WinProb)
+	if ev.Cmp(big.NewRat(0, 1)) <= 0 {
+		return nil
+	}
+
 	info, err := s.senderManager.GetSenderInfo(s.signer.Account().Address)
 	if err != nil {
 		return err
 	}
 
-	// validate sender
 	if err := s.validateSender(info); err != nil {
 		return err
 	}
 
-	ev := ticketEV(ticketParams.FaceValue, ticketParams.WinProb)
 	totalEV := ev.Mul(ev, new(big.Rat).SetInt64(int64(numTickets)))
 	if totalEV.Cmp(s.maxEV) > 0 {
 		return fmt.Errorf("total ticket EV %v for %v tickets > max total ticket EV %v", totalEV.FloatString(5), numTickets, s.maxEV.FloatString(5))
@@ -165,15 +177,6 @@ func (s *sender) validateTicketParams(ticketParams *TicketParams, numTickets int
 	maxFaceValue := new(big.Int).Div(info.Deposit, big.NewInt(int64(s.depositMultiplier)))
 	if ticketParams.FaceValue.Cmp(maxFaceValue) > 0 {
 		return fmt.Errorf("ticket faceValue %v > max faceValue %v", ticketParams.FaceValue, maxFaceValue)
-	}
-
-	if ticketParams.ExpirationBlock.Int64() == 0 {
-		return nil
-	}
-
-	latestBlock := s.timeManager.LastSeenBlock()
-	if ticketParams.ExpirationBlock.Cmp(latestBlock) <= 0 {
-		return ErrTicketParamsExpired
 	}
 
 	return nil
--- a/server/rpc_test.go
+++ b/server/rpc_test.go
@@ -650,6 +650,8 @@ func TestValidatePrice(t *testing.T) {
 	err := validatePrice(s)
 	assert.Nil(err)
 
+	defer BroadcastCfg.SetMaxPrice(nil)
+
 	// B MaxPrice > O Price
 	BroadcastCfg.SetMaxPrice(big.NewRat(5, 1))
 	err = validatePrice(s)
--- a/core/orch_test.go
+++ b/core/orch_test.go
@@ -1408,6 +1408,15 @@ func TestPriceInfo(t *testing.T) {
 	assert.Equal(priceInfo.PricePerUnit, expPrice.Num().Int64())
 	assert.Equal(priceInfo.PixelsPerUnit, expPrice.Denom().Int64())
 
+	// basePrice = 0 => expPricePerPixel = 0
+	n.SetBasePrice(big.NewRat(0, 1))
+	orch = NewOrchestrator(n, nil)
+
+	priceInfo, err = orch.PriceInfo(ethcommon.Address{})
+	assert.Nil(err)
+	assert.Zero(priceInfo.PricePerUnit)
+	assert.Equal(int64(1), priceInfo.PixelsPerUnit)
+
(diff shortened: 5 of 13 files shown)


kyriediculous (changes requested):
I don't think this is ideal.

I'm not against allowing 0 pricing, but let's only allow it when explicitly using `-pricePerPixel 0`

We should keep the error when the flag is not set. 

I guess we can use a string instead of an int and parse  that to an int. An unset flag would then equal `''` instead of `0`

yondonfu:
> I'm not against allowing 0 pricing, but let's only allow it when explicitly using -pricePerPixel 0

Yeah fair point on allowing the price to be set to 0, but making that a choice the user has to make as opposed to making that the silent default.

3640ce2 requires -pricePerUnit to be set and allows the user to explicitly specify -pricePerUnit 0.

server/rpc_test.go
```diff
@@ -650,6 +650,8 @@ func TestValidatePrice(t *testing.T) {
 	err := validatePrice(s)
 	assert.Nil(err)
 
+	defer BroadcastCfg.SetMaxPrice(nil)
```
yondonfu, line 653: This was needed because this package var was not being reset which affected other tests that depend on it.

cmd/livepeer/livepeer.go
```diff
@@ -515,9 +515,12 @@ func main() {
 				// Can't divide by 0
 				panic(fmt.Errorf("The amount of pixels per unit must be greater than 0, provided %d instead\n", *pixelsPerUnit))
 			}
-			if *pricePerUnit <= 0 {
-				// Prevent orchestrator from unknowingly provide free transcoding
-				panic(fmt.Errorf("Price per unit of pixels must be greater than 0, provided %d instead\n", *pricePerUnit))
+			if !isFlagSet["pricePerUnit"] && *pricePerUnit == 0 {
```
yondonfu, line 518: Technically we don't need the `*pricePerUnit == 0` check since the default is 0. But, I figured it would be nice to have in case the default becomes non-zero in the future since we only want to error out if the flag is not set *and* the default value is 0.

cmd/livepeer/livepeer.go
```diff
@@ -513,11 +513,14 @@ func main() {
 			// Set price per pixel base info
 			if *pixelsPerUnit <= 0 {
 				// Can't divide by 0
-				panic(fmt.Errorf("The amount of pixels per unit must be greater than 0, provided %d instead\n", *pixelsPerUnit))
+				panic(fmt.Errorf("-pixelsPerUnit must be > 0, provided %d", *pixelsPerUnit))
```
yondonfu, line 516: Since new error messages were being added for the pricePerUnit checks below I thought this would be a good opportunity to:

- Reference the flag name itself in the error message
- Make the error message a bit more concise which makes it slightly less cumbersome to check for the messages in `test_args.sh` using grep

pm/recipient.go
```diff
@@ -36,8 +36,8 @@ type Recipient interface {
 	// for a provided sender ETH adddress
 	TicketParams(sender ethcommon.Address, price *big.Rat) (*TicketParams, error)
 
-	// TxCostMultiplier returns the multiplier -
-	TxCostMultiplier(sender ethcommon.Address) (*big.Rat, error)
+	// TxCostMultiplier returns the tx cost multiplier based on the provide face value
```
darkdarkdragon:
```suggestion
	// TxCostMultiplier returns the tx cost multiplier based on the provided face value
```

pm/sender.go
```diff
@@ -86,14 +86,6 @@ func (s *sender) validateSender(info *SenderInfo) error {
 		return ErrSenderValidation{fmt.Errorf("unable to validate sender: deposit and reserve is set to unlock soon")}
```
darkdarkdragon: Shouldn't we skip this check if price is zero?

darkdarkdragon (changes requested):
One question, otherwise looks good.

pm/sender.go
```diff
@@ -86,14 +86,6 @@ func (s *sender) validateSender(info *SenderInfo) error {
...
 	}
 
-	if info.Reserve.FundsRemaining.Cmp(big.NewInt(0)) == 0 {
```
kyriediculous: In light of the comment above, shouldn't we keep these checks but skip them if price is zero?

kyriediculous (changes requested):
Overall I find the changes a bit intrusive given the fact that this is a feature that we actually never expect our end-users to use. This change is purely for the Livepeer Inc _private_ orchestrator deployments to toggle functionality for quickly improving reliability by disabling some of the on-chain workflow.

cmd/livepeer/livepeer.go
```diff
@@ -513,11 +513,14 @@ func main() {
...
+			}
+			if !isFlagSet["pricePerUnit"] && *pricePerUnit == 0 {
+				// Prevent orchestrators from unknowingly providing free transcoding
+				panic(fmt.Errorf("-pricePerUnit must be set"))
 			}
-			if *pricePerUnit <= 0 {
-				// Prevent orchestrator from unknowingly provide free transcoding
-				panic(fmt.Errorf("Price per unit of pixels must be greater than 0, provided %d instead\n", *pricePerUnit))
+			if *pricePerUnit < 0 {
+				panic(fmt.Errorf("-pricePerUnit must be > 0, provided %d", *pricePerUnit))
```
jailuthra:
```suggestion
				panic(fmt.Errorf("-pricePerUnit must be >= 0, provided %d", *pricePerUnit))
```
Even though most public Os won't ever set it = 0, I still think we should keep the error msg close to the truth for some future dev reading the error msg or reading the codebase and make it clear that it can be 0?

pm/sender.go
jailuthra: +1

Also if this means sending price everywhere along with sender address like done in the `faceValue` function before, and that leading to unclean code - maybe we can try to see if it makes sense to add `price` as a field in one of the structs (maybe SenderInfo, im not sure?)

yondonfu:
> this is a feature that we actually never expect our end-users to use. This change is purely for the Livepeer Inc private orchestrator deployments to toggle functionality for quickly improving reliability by disabling some of the on-chain workflow.

I disagree that this feature is only useful in the scenario that you describe. See #1615 for a few other reasons why free transcoding can be useful. Furthermore:

- Supporting this feature moves the node toward a world where there does not necessarily need to be as strong of a distinction between "on-chain mode" and "off-chain mode" and instead specific features (i.e. the payment workflow) can be toggled on/off and configured (i.e. different types of discovery - webhook vs. on-chain registry is one example here) based on a user's requirements
- Orchestrators should be able to charge whatever price that they want (given the open price advertisement mechanism used on the network right now) and that includes 0 if it makes sense for them to do so. See #1615 for a continuation of this thought as well as a link to other discussions that have provided examples of when a price of 0 might make sense for an orchestrator

> Overall I find the changes a bit intrusive

This is a fair point. I'm not that thrilled about the structure of the current set of changes given some of the special case logic required on the O side in order to skip the PM related checks when the price is set to 0. An alternative that I'm considering is to disable the PM related checks by setting the EV to 0 (which automatically results in a minimum credit stream balance of 0). The EV could be set to 0 manually either via `-ticketEV` on startup or via a webserver request. The downside here is that it requires a user to set both the price and EV. We could consider by default setting the EV to 0 when the price is set to 0, but that still requires the user to update both the price and EV to non-zero values when it wants to charge for transcoding. Will think about it some more and loop back.

cmd/livepeer/livepeer.go
yondonfu: Fixed in a81e82b and 5f5bf2a

yondonfu:
> Will think about it some more and loop back.

Made a few changes that slightly reduce the scope of this PR:

**Changes to the Sender in the pm package**

- fc93639 reverts 137d3ef to rollback the removal of the deposit/reserve checks for the sender
- Instead, we now simply check if the ticket EV is > 0 in 1964464. This preserves the existing `validateSender()` checks when the ticket EV is > 0 and disables them when the ticket EV = 0. Since the default behavior in this PR for an O when its price is set to 0 is to advertise ticket EV = 0, these `validateSender()` checks will be disabled when working with an O with price = 0

**Changes to the Recipient in the pm package and orchestrator in the core package**

- 02ad742 reverts 676468c to rollback the previous logic for setting faceValue & winProb = 0 when price <= 0
- Instead, in 9d25615 we isolate the updates in `TicketParams()` in the `pm` package and `priceInfo()` in the `core` package

I ended up sticking with the current approach of overriding an O's ticket EV to 0 when its price is set to 0 because I felt like it would be easier to manage than requiring an O operator to also set its ticket EV to 0 on its own.

pm/sender.go
yondonfu: Fixed in 1964464 - see [this comment](https://github.com/livepeer/go-livepeer/pull/1616#issuecomment-698581224) for more details on the overall changes.

pm/sender.go
yondonfu: Fixed in 1964464 - see [this comment](https://github.com/livepeer/go-livepeer/pull/1616#issuecomment-698581224) for more details on the overall changes.

pm/recipient.go
yondonfu: Fixed in 9d25615 - see [this comment](https://github.com/livepeer/go-livepeer/pull/1616#issuecomment-698581224) for more details on the overall changes.

darkdarkdragon (approved):
LGTM

pm/sender.go
```diff
@@ -151,12 +151,14 @@ func (s *sender) validateTicketParams(ticketParams *TicketParams, numTickets int
 		return err
 	}
 
+	ev := ticketEV(ticketParams.FaceValue, ticketParams.WinProb)
 	// validate sender
-	if err := s.validateSender(info); err != nil {
-		return err
+	if ev.Cmp(big.NewRat(0, 1)) > 0 {
```
kyriediculous: Just a minor thing, I don't think executing this entire function is necessary if `ev == 0` because 

- totalEV will also be 0 and will never exceed `maxEV`
- faceValue is 0 and will never exceed `maxFaceValue`

So if I'm not mistake we could move the `TicketParams.ExpirationBlock` checks before this block

i.e.

```
... TicketParams expiration checks ... 

if ev.Cmp(big.newRat(0,1)) <=0 { return nil }

...valitedateSender, maxEv and maxFaceVal checks...
```

kyriediculous (changes requested):

pm/sender.go
yondonfu: Fixed in e067db8

kyriediculous (approved):

kyriediculous (changes requested):
Changes look good , let's rebase !

yondonfu:
Rebased

kyriediculous (approved):
Great ! LGTM Pending green CI (although the windows build is failing due to a timeout to an external IP)

PR #1617, opened 2020-09-18
server: All stake wt. to fastest session of an address

### LINKED ISSUE #1614: proposal: Improve on-chain Orchestrator selection
## Abstract

This is a proposal to improve the on-chain Orchestrator selection performed by the Broadcaster node. The stream can be made more robust against failovers by relying on latency and network conditions to break ties amongst equally-staked Orchestrators.

## Motivation

To transcode an incoming stream the Broadcaster needs to select the best performing Orchestrator amongst the ones that respond during the [initial discovery](https://github.com/livepeer/go-livepeer/blob/f1e26c0a9/discovery/discovery.go#L48). Currently, in the on-chain mode the Broadcaster performs a random selection [weighted by the LPT Stake](https://github.com/livepeer/go-livepeer/blob/f1e26c0a9d8/server/selection.go#L149) to select an Orchestrator amongst the list returned after discovery.

Relying on stake alone is troublesome in case many Os have roughly the same stake, or the worst-case where *all* Os are operated under the same ETH account and thus have *exactly same stake*. It is problematic as not all Orchestrators under the same ETH account or similar stake operate under similar network conditions or global regions - and thus offer wildly different quality from the B's perspective.

This often causes the B to select a relatively poor O at the start of the stream, and then do a *costly failover* later on when it performs poorly. Ideally the B would like to choose the initial O smartly amongst similar-staked Os, by considering how fast they respond to the discovery pings, or how quickly transcode a small initial sample.

## Proposed Solution

The simplest way to get a rough idea of Orchestrator latency is by relying on the order in which the Os are present in the [`unknownSessions`](https://github.com/livepeer/go-livepeer/blob/f1e26c0a9d8/server/selection.go#L113) list returned after discovery, which is filled in order as each [Go routine](https://github.com/livepeer/go-livepeer/blob/f1e26c0a9/discovery/discovery.go#L101) to ping a particular O [returns](https://github.com/livepeer/go-livepeer/blob/f1e26c0a9/discovery/discovery.go#L110).

Currently, for all Os of a particular ETH account that we're considering during selection, we divide up the total account stake equally amongst them. The simplest way to improve this would be to give *all the stake* of a particular account to the first/fastest O in the `unknownSessions` list and zero stake to other Os of the same account, before doing the usual random-weighted selection.

This will ensure that only the fastest O of each account is considered during the initial selection. While this approach does not consider the case of Os across accounts that might have same (or roughly same) stake, it is still a considerable improvement over the current selection mechanism.

## Alternatives

If we want to prioritize selection of a faster O with exactly same or similar stake as other Os *across accounts*, we can consider an alternate approach which does not divide up the stake amongst Os of each account to perform a weighted-random-selection amongst all Os.

Instead we can make *stake-wise buckets* with multiple Os (of exactly same stake, or stake within some range), do weighted random selection between the buckets, and select the fastest O in the selected bucket - which should be the first O if we use an ordered List for the bucket, and maintain the initial ordering from `unknownSessions`.

## Known Unknowns

It is unclear how Go Routines get scheduled in case of limited CPU cores or threads i.e. it might be possible that the go routine to ping `Orch0` returns much earlier than `Orch1` not because it had lower ping, but because it got the unfair advantage of being scheduled earlier in an environment of little resources. This will result in the `unknownSessions` list to not be a fair representation of actual ping times.

For now we can assume that the scheduling is good enough - but if we face issues we can consider adding *timing* code within the Go routine to get the exact network-level latency, and sort the list by that timing instead.

jailuthra:

While selecting an Orchestrator during on-chain mode, give all stake weight to the *fastest* O session if multiple sessions belong to the same ETH address. (See #1614 for details)

Look at the commits

Updated the unit test for Stake weight division amongst sessions of the same address

Resolves #1614 

**Checklist:**

- [x] Node runs on ArchLinux ~~OSX and devenv~~

### DIFF
--- a/server/selection_test.go
+++ b/server/selection_test.go
@@ -322,13 +322,12 @@ func TestMinLSSelector_SelectUnknownSession_UniformWeights(t *testing.T) {
 	}
 }
 
-func TestMinLSSelector_SelectUnknownSession_SameAddressDivideStake(t *testing.T) {
+func TestMinLSSelector_SelectUnknownSession_SameAddress(t *testing.T) {
 	stakeRdr := newStubStakeReader()
 
-	addr := ethcommon.BytesToAddress([]byte("foo"))
 	selections := 100000
 
-	createSessions := func(num int) []*BroadcastSession {
+	createSessions := func(num int, addr ethcommon.Address) []*BroadcastSession {
 		sessions := make([]*BroadcastSession, num)
 		for i := 0; i < num; i++ {
 			sessions[i] = &BroadcastSession{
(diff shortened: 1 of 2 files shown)


server/selection_test.go
```diff
@@ -325,10 +325,9 @@ func TestMinLSSelector_SelectUnknownSession_UniformWeights(t *testing.T) {
 func TestMinLSSelector_SelectUnknownSession_SameAddressDivideStake(t *testing.T) {
```
yondonfu: Could we update the name of this test since we're no longer dividing stake for the same address?

yondonfu (changes requested):
Looks good! Just a minor comment about a test name.

yondonfu (approved):
LGTM after squashing the fixup!

PR #1619, opened 2020-09-22
pm: Fix busy ticket queue loop

### LINKED ISSUE #1618: transactor sits at 100% single-core CPU usage
**Describe the bug**
From our prod infra:

```
ewr-prod-livepeer-transactor-7cbf9c65cc-sr66x        1003m        40Mi
ewr-staging-livepeer-transactor-5765c8854f-n2x6s     2m           34Mi
prod-livepeer-transactor-7854d44cbf-dpkph                  1006m        37Mi
staging-livepeer-secondary-transactor-66cdfd7678-m6dnt     1007m        32Mi
```

We very frequently see single-threaded transactor usage stuck at 100% single core CPU usage. This would be consistent with an infinite loop within a single goroutine. I believe this is adversely affecting the startup time of our on-chain orchestrators.

I have a core dump from one of these nodes ([got it like this](https://pro-tips-dot-com.tumblr.com/post/47677612115/kill-a-hung-go-process-and-print-stack-traces)) but I'm not 100000% sure that it doesn't contain any sensitive key information so ask me for it directly.

yondonfu:

This PR fixes a busy ticket queue loop that was causing CPU usage on any node that creates a ticket queue loop (either an O that is redeeming tickets on its own or a transactor that is running a ticket redeemer service) to spike to 100% as soon as the ticket queue loop is created (i.e. if `SenderMonitor.MaxFloat()` is called). The CPU spike would last until the SenderMonitor stops the ticket queue loop when it [stops monitoring an address due to inactivity](https://github.com/livepeer/go-livepeer/blob/master/pm/sendermonitor.go#L343) for a certain period of time.

See commit history.

Tested manually. Without this change, as soon as you connect an O to a transactor running a redeemer service (i.e. using the `-redeemer` flag), you can observe the CPU usage of the transactor spike to 100%. With this change, the CPU usage of the transactor in this setup does not spike. The same is true if you start an O that does standalone redemption.

Fixes #1618 

**Checklist:**

### DIFF
--- a/pm/queue.go
+++ b/pm/queue.go
@@ -145,7 +145,6 @@ ticketLoop:
 			}
 		case <-q.quit:
 			return
-		default:
 		}
 	}
 }


kyriediculous (approved):
Interesting.

So I guess an empty `default` doesn't mean "do nothing" until a new message is received and there appears to be a difference between that and blocking until a message is received ? 

Change looks good to me  👍

yondonfu:
> So I guess an empty default doesn't mean "do nothing" until a new message is received and there appears to be a difference between that and blocking until a message is received ?

Yep. A `default` case in a switch statement will fire if no other case in the select statement fires first.

kyriediculous:
I’m still not entirely sure how this causes 100% CPU load though as there is node code block to be executed. I imagined it would immediately finish and wait for the next iteration of the loop.

yondonfu:
> I’m still not entirely sure how this causes 100% CPU load though as there is node code block to be executed. I imagined it would immediately finish and wait for the next iteration of the loop.

The `default` statement would keep on firing and the loop would keep on running because at the next iteration the `default` statement would once again fire first. The loop wouldn't be doing anything, but it would take up CPU time that is forced to execute the loop iteration which could otherwise be doing something else.

kyriediculous:
Gotcha ! Thanks for clarifying

Op di 22 sep. 2020 om 20:14 schreef Yondon Fu <notifications@github.com>

PR #1623, opened 2020-09-29
More metrics

darkdarkdragon:
1. Adds 'source_segment_duration_seconds' metric
2. Adds 'transcode_time_avg_seconds' and 'transcode_overall_latency_avg_seconds' metrics - which is average of 'transcode_time_seconds' and 'transcode_overall_latency_seconds' over one minute but calculated on node itself.

Unit tests


monitor/census.go
```diff
@@ -200,12 +230,22 @@ func InitCensus(nodeType, nodeID, version string) {
 	census.mTranscodersLoad = stats.Int64("transcoders_load", "Total load of transcoders currently connected to orchestrator", "tot")
 	census.mSuccessRate = stats.Float64("success_rate", "Success rate", "per")
 	census.mTranscodeTime = stats.Float64("transcode_time_seconds", "Transcoding time", "sec")
+	census.mTranscodeTimeAvg = stats.Float64("transcode_time_avg_seconds", "Transcoding time average over one minute", "sec")
+	census.mTranscodeTimeRealtimeRatio = stats.Float64("transcode_time_realtime_ratio", "Transcoding time divided by segment's duration", "perc")
+	census.mTranscodeTimeRealtimeRatioAvg = stats.Float64("transcode_time_realtime_ratio_avg", "Transcoding time divided by segment's duration average over one minute", "perc")
 	census.mTranscodeLatency = stats.Float64("transcode_latency_seconds",
-		"Transcoding latency, from source segment emered from segmenter till transcoded segment apeeared in manifest", "sec")
+		"Transcoding latency, from source segment emered from segmenter till transcoded segment appeared in manifest", "sec")
```
yondonfu: "emered" -> "emerged". Same comment for the other descriptions below.

core/transcoder.go
```diff
@@ -51,7 +51,7 @@ func (lt *LocalTranscoder) Transcode(md *SegTranscodingMetadata) (*TranscodeData
 		// When orchestrator works as transcoder, `fname` will be relative path to file in local
 		// filesystem and will not contain seqNo in it. For that case `SegmentTranscoded` will
 		// be called in orchestrator.go
-		monitor.SegmentTranscoded(0, seqNo, time.Since(start), common.ProfilesNames(profiles))
+		monitor.SegmentTranscoded(0, seqNo, 0, time.Since(start), common.ProfilesNames(profiles))
```
yondonfu: Can we use `md.Duration` here?

yondonfu (changes requested):
I know that we believe that the [Prometheus query](https://github.com/livepeer/docker-livepeer/blob/master/monitoring/grafana/livepeer_overview.json#L1121) we use to calculate average latency metrics right now is wrong, but do you a) know what was wrong with that Prometheus query and b) also believe that there is no way to correctly calculate the average metrics (i.e. using Prometheus queries) besides calculating them on our own within the node?

core/transcoder.go
darkdarkdragon: Fixed

monitor/census.go
darkdarkdragon: Fixed

darkdarkdragon:
@yondonfu 
> a) know what was wrong with that Prometheus query and

Don't know
> b) also believe that there is no way to correctly calculate the average metrics (i.e. using Prometheus queries) besides calculating them on our own within the node?

This requires investigation. Calculating own averages will give me some data to compare with.

yondonfu:
> This requires investigation. Calculating own averages will give me some data to compare with.

I just filed #1628 which I think may be contributing to metrics that "look wrong" in the Grafana dashboards.

The TLDR is that when we're recording durations in the sub-second range, we're recording 0 values for those measurements. So, what could be happening on the Prometheus side is that for a metric like overall latency or upload time the _count counter increases, but the _sum counter remains unchanged. Not sure how the `rate()` Prometheus query function handles this under the hood since the [docs](https://prometheus.io/docs/prometheus/latest/querying/functions/#rate) note that it handles breaks in monotonicity (is this a increasing requirement or strictly increasing requirement?) and does some stuff to handle missing data. I think its worth fixing #1628 and doing some testing to make sure the metrics are being recorded properly before adding new metrics and a new way of calculating moving averages - what do you think?

darkdarkdragon:
Fixed #1628 in 7281c148e19e799950f46967549ab8e61906f755

`rate()` function in Prometheus does [interpolation](https://github.com/prometheus/prometheus/issues/3746), so I still want to compare own calculation of averages with Prometheus one. If after the fix Prometheus'es averages will start to look normally (and on par with own) - we can remove own averages.

yondonfu:
> rate() function in Prometheus does interpolation, so I still want to compare own calculation of averages with Prometheus one. If after the fix Prometheus'es averages will start to look normally (and on par with own) - we can remove own averages.

In that case, can the fix to #1628 + non-average related metrics be put in one PR and the moving average calculation in another PR? Then, we can work on merging the former fairly quickly. From there, possible steps I see are:

1. See the updated Prometheus metric calculations in staging based on the fix to #1628 
2. If there are still problems, we can also update one cluster in staging with the branch with the moving average calculation to compare

We could even do 1 & 2 in parallel.

darkdarkdragon:
looks like it is not needed anymore

PR #1625, opened 2020-10-05
ci: Use msys2/setup-msys2 for Windows

yondonfu:

This PR fixes the Windows GH action workflow which is currently broken.

- Switched from `numworks/setup-msys2` action to the `msys2/setup-msys2` GH action since the latter looks like latter has [replaced the former](https://github.com/numworks/setup-msys2/issues/37)
- Switch to a Sourceforge mirror for installing go1.13 using pacman. At the time of writing this GH issue, `repo.msys2.org` is down (and I believe has been down for a few days...). pacman fails over to other mirrors when installing packages, but since we're installing go1.13 via a remote URL I think we need to specify a different URL
- Update the Windows dependencies referenced in `upload_build.sh`. Looks like they've changed with the version of msys2 setup with the new GH action

See commit history.

N/A

**Checklist:**

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -7,21 +7,26 @@ jobs:
       run: git config --global core.autocrlf false
     - uses: actions/checkout@v1
     - name: Setup Msys2 environment
-      uses: numworks/setup-msys2@v1
+      uses: msys2/setup-msys2@v2
       with:
         msystem: MSYS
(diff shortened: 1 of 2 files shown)


darkdarkdragon (approved):
LGTM
checked broadcaster on Windows - it works

yondonfu:
Thanks @darkdarkdragon !

PR #1627, opened 2020-10-13
Allow to use ManifestID provided by AuthWebhook in HTTP push flow

### LINKED ISSUE #1528: Webhook manifestID breaks HTTP push
**Describe the bug**

In the [auth webhook](https://github.com/livepeer/go-livepeer/blob/master/doc/rtmpwebhookauth.md), a custom ManfiestID field can be specified. This is intended to offer a degree of protection by distinguishing the playback name from the publishing name.

With RTMP ingest, a single reference to the [rtmpConnection](https://github.com/livepeer/go-livepeer/blob/d9ea3358e0a43f07a0fd78880f37bd48750c1be2/server/mediaserver.go#L83-L92) struct (hereafter, the 'job') is maintained for the lifetime of the RTMP connection, so there is no problem if its `mid` field is (re-)set during the webhook authentication process.

With HTTP push, the manifest ID is extracted from the URL. If the webhook specifies a separate manifest ID for playback, then that one is registered and used for further lookups. Subsequent segments using the original manifest ID extracted from the URL will not be able to look up the existing job that's registered with the webhook-specified ID. This leads to a couple issues:

* Results in an avoidable auth webhook call for each segment. The webhook should only be invoked once for as long as the job is active. [Registering the job](https://github.com/livepeer/go-livepeer/blob/d9ea3358e0a43f07a0fd78880f37bd48750c1be2/server/mediaserver.go#L417-L420) after receiving the webhook response then fails, since the job already exists. This results in an [HTTP error](https://github.com/livepeer/go-livepeer/blob/d9ea3358e0a43f07a0fd78880f37bd48750c1be2/server/mediaserver.go#L643-L647) being returned and the transcode does not run.

* The [watchdog](https://github.com/livepeer/go-livepeer/blob/d9ea3358e0a43f07a0fd78880f37bd48750c1be2/server/mediaserver.go#L649-L665) can not clean up the job since it attempts to look up using the original ManifestID, while the job is actually indexed by the webhook-specified ManifestID. This leaks the job.

**To Reproduce**
1. Start a broadcaster node with `-authWebhookUrl`
2. Start an auth endpoint that returns `{"manifestID":"foo"}`
3. Do a HTTP push with a different manifestID, eg `curl --data-binary "@input.mp4" "http://localhost:8935/live/bar/0.mp4"`
4. Do another HTTP push, eg `curl --data-binary "@input.mp4" "http://localhost:8935/live/bar/1.mp4"`
5. Second push should fail. Curl output should show an error like this, "Could not create stream ID:"
6. Observe a failed cleanup for manifest ID "bar" around a minute after the first push. Broadcaster console should show something like " Attempted to end unknown stream with manifest ID bar"

**Expected behavior**

We probably need two ways of indexing the job: by user submitted id (for HTTP push), or webhook submitted id (for playback).  A two-step lookup for one of these might be okay as well.

The distinction could also be clearer in the field naming, eg maybe "stream name" or "playback name" or something. Bonus points if we can gradually start phasing out the use of the term "manifest ID". Related: https://github.com/livepeer/go-livepeer/issues/871

darkdarkdragon:
Allow to use ManifestID provided by AuthWebhook in HTTP push flow

Added map that keeps mapping between ManifestID sent in HTTP push URL and one provided by webhook

Added unit tests

Fixes #1528

### DIFF
--- a/server/mediaserver.go
+++ b/server/mediaserver.go
@@ -88,10 +88,11 @@ type LivepeerServer struct {
 
 	// Thread sensitive fields. All accesses to the
 	// following fields should be protected by `connectionLock`
-	rtmpConnections map[core.ManifestID]*rtmpConnection
-	lastHLSStreamID core.StreamID
-	lastManifestID  core.ManifestID
-	connectionLock  *sync.RWMutex
+	rtmpConnections   map[core.ManifestID]*rtmpConnection
+	internalManifests map[core.ManifestID]core.ManifestID
+	lastHLSStreamID   core.StreamID
+	lastManifestID    core.ManifestID
+	connectionLock    *sync.RWMutex
 }
 
 type authWebhookResponse struct {
@@ -146,7 +147,8 @@ func NewLivepeerServer(rtmpAddr string, lpNode *core.LivepeerNode, httpIngest bo
 	}
 	server := lpmscore.New(&opts)
 	ls := &LivepeerServer{RTMPSegmenter: server, LPMS: server, LivepeerNode: lpNode, HTTPMux: opts.HttpMux, connectionLock: &sync.RWMutex{},
-		rtmpConnections: make(map[core.ManifestID]*rtmpConnection),
+		rtmpConnections:   make(map[core.ManifestID]*rtmpConnection),
+		internalManifests: make(map[core.ManifestID]core.ManifestID),
 	}
 	if lpNode.NodeType == core.BroadcasterNode && httpIngest {
 		opts.HttpMux.HandleFunc("/live/", ls.HandlePush)
@@ -510,19 +512,26 @@ func (s *LivepeerServer) registerConnection(rtmpStrm stream.RTMPVideoStream) (*r
 	return cxn, nil
 }
 
-func removeRTMPStream(s *LivepeerServer, mid core.ManifestID) error {
+func removeRTMPStream(s *LivepeerServer, extmid core.ManifestID) error {
 	s.connectionLock.Lock()
 	defer s.connectionLock.Unlock()
-	cxn, ok := s.rtmpConnections[mid]
+	intmid := extmid
+	if _intmid, exists := s.internalManifests[extmid]; exists {
+		// Use the internal manifestID that was stored for the provided manifestID
+		// to index into rtmpConnections
+		intmid = _intmid
+	}
+	cxn, ok := s.rtmpConnections[intmid]
 	if !ok || cxn.pl == nil {
-		glog.Error("Attempted to end unknown stream with manifest ID ", mid)
+		glog.Errorf("Attempted to end unknown stream with manifestID=%s", extmid)
 		return errUnknownStream
 	}
 	cxn.stream.Close()
 	cxn.sessManager.cleanup()
 	cxn.pl.Cleanup()
-	glog.Infof("Ended stream with id=%s", mid)
-	delete(s.rtmpConnections, mid)
+	glog.Infof("Ended stream with manifestID=%s external manifestID=%s", intmid, extmid)
+	delete(s.rtmpConnections, intmid)
+	delete(s.internalManifests, extmid)
 
 	if monitor.Enabled {
 		monitor.StreamEnded(cxn.nonce)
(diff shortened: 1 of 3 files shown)


server/mediaserver.go
```diff
@@ -711,11 +719,18 @@ func (s *LivepeerServer) HandlePush(w http.ResponseWriter, r *http.Request) {
...
 				s.connectionLock.RUnlock()
 				if time.Since(lastUsed) > httpPushTimeout {
-					_ = removeRTMPStream(s, mid)
+					_ = removeRTMPStream(s, intmid, extmid)
 					return
 				}
 			}
-		}(s, mid)
+		}(s, cxn.mid, mid)
+		if cxn.mid != mid {
```
yondonfu, line 731: What do you think about moving this `s.internalManifests` update above the watchdog goroutine? If we do that, then `mid` would be set to the value used to index into `s.rtmpConnections` (i.e. either the push URL manifestID or the webhook manifestID if it is different) before the watchdog goroutine. Then, I think we can just pass `mid` into the watchdog goroutine and also avoid the extra argument in `removeRTMPStream()` by doing something like:

```golang
func removeRTMPStream(s *LivepeerServer, mid core.ManifestID) error {
    ...
    // Clean up s.rtmpConnections
    // Delete the internal manifestID if it exists and no-op if it doesn't
    delete(s.internalManifests, mid)
}
```

yondonfu (changes requested):

iameli:
Presently doing an integration test with this in mdw-staging

server/mediaserver.go
darkdarkdragon: `internalManifests` and `rtmpConnections` indexed by different values, so we need both inside `removeRTMPStream`

I should have called it `external2internalManifest` instead of `internalManifests`...

yondonfu: Ah right my mistake.

- If the webhook returns a different manifestID than the push URL manifestID, rtmpConnections is indexed by the webhook manifestID and internalManifests maps push URL manifestID -> webhook manifestID
- Otherwise, rtmpConnections is indexed by the push URL manifest and internalManifests is not updated
- The push URL manifestID is the "external manifestID" and the webhook manifestID is the "internal manifestID"

Could it make sense to always pass the external manifestID to `removeRTMPStream()` and then do a lookup inside the function for the internal manifestID if it exists? Something like:

```golang
func removeRTMPStream(s *LivepeerServer, mid core.ManifestID) error {
    s.connectionLock.Lock()
    defer s.connectionLock.Unlock()
    if intmid, exists := s.internalManifests[mid]; exists {
        // Use the internal manifestID that was stored for the provided manifestID
        // to index into rtmpConnections
        mid = intmid
    }
    cxn, ok := s.rtmpConnections(mid)
    ...
}
```

This would involve an extra lookup in `removeRTMPStream()`, but IMO it feels preferable to having to pass two manifestIDs into `removeRTMPStream()` every time it is used which seems a bit cumbersome.

iameli:
Seems to be working well.

yondonfu:
Hm I see there was a force push with 824bb89, but doesn't look like anything changed? Was that intended @darkdarkdragon ?

iameli:
I think that was a rebase onto master to pick up your auth token branch

server/mediaserver.go
darkdarkdragon: Fixed in b014d7ed82ac4482071248edc58a4dcf6ef9d53c

server/mediaserver.go
```diff
@@ -701,7 +713,7 @@ func (s *LivepeerServer) HandlePush(w http.ResponseWriter, r *http.Request) {
 
 		// Start a watchdog to remove session after a period of inactivity
 		ticker := time.NewTicker(httpPushTimeout)
-		go func(s *LivepeerServer, mid core.ManifestID) {
+		go func(s *LivepeerServer, intmid, extmid core.ManifestID) {
```
yondonfu, line 716: Only need to pass in extmid here since `removeRTMPStream()` just needs the external manifestID now.

darkdarkdragon: fixed in 63d4ab6ebadda6b3ad0f52eda2bc014d31ac5054
Need both arguments because we also need to index `rtmpConnections` inside watchdog function.

yondonfu: Oh I see...Got it. Perhaps there is room for further hiding the notion of an internal manifestID in this case so a user of `rtmpConnections` only ever has to worry about the external manifestID...think that's fine to address separately.

yondonfu (approved):
LGTM after squashing the fixup commits.

PR #1630, opened 2020-10-15
add QoS metrics

darkdarkdragon:
- Add 'source_segment_duration_seconds' metric
- Add metrics in HTTP push handler that calcs realtime ratio of the request

Unit tests

### DIFF
--- a/monitor/census.go
+++ b/monitor/census.go
@@ -99,6 +99,7 @@ type (
 		mTranscodeOverallLatency      *stats.Float64Measure
 		mUploadTime                   *stats.Float64Measure
 		mAuthWebhookTime              *stats.Float64Measure
+		mSourceSegmentDuration        *stats.Float64Measure
 
 		// Metrics for sending payments
 		mTicketValueSent    *stats.Float64Measure
@@ -206,6 +207,7 @@ func InitCensus(nodeType, nodeID, version string) {
 		"Transcoding latency, from source segment emered from segmenter till all transcoded segment apeeared in manifest", "sec")
 	census.mUploadTime = stats.Float64("upload_time_seconds", "Upload (to Orchestrator) time", "sec")
 	census.mAuthWebhookTime = stats.Float64("auth_webhook_time_milliseconds", "Authentication webhook execution time", "ms")
+	census.mSourceSegmentDuration = stats.Float64("source_segment_duration_seconds", "Source segment's duration", "sec")
 
 	// Metrics for sending payments
 	census.mTicketValueSent = stats.Float64("ticket_value_sent", "TicketValueSent", "gwei")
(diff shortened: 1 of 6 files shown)


monitor/census.go
```diff
@@ -176,6 +184,13 @@ func InitCensus(nodeType, nodeID, version string) {
 	if err != nil {
 		glog.Fatal("Error creating context", err)
 	}
+	census.mHTTPClientTimeout1 = stats.Int64("http_client_timeout_1", "Number of times HTTP connection was dropped before transcoding complete", "tot")
+	census.mHTTPClientTimeout2 = stats.Int64("http_client_timeout_2", "Number of times HTTP connection was dropped before transcoded segments was sent back to client", "tot")
+	census.mRealtime3x = stats.Int64("segment_transcoded_realtime_3x", "Number of segment transcoded 3x faster than realtime", "tot")
```
yondonfu: Should these names also have the "http_client_" prefix to clarify that the metrics are measured from the POV of the HTTP client and *not* the POV of the broadcaster since we have other metrics such as `transcode_overall_latency_seconds` that are measured from the POV of the broadcaster?

server/mediaserver.go
```diff
@@ -819,18 +830,47 @@ func (s *LivepeerServer) HandlePush(w http.ResponseWriter, r *http.Request) {
...
+		err = mw.Close()
+	}
+	if err != nil {
+		glog.Errorf("Error sending transcoded response url=%s err=%v", r.URL.String(), err)
+		if monitor.Enabled {
+			monitor.HTTPClientTimedOut2()
+		}
+		return
+	}
+	w.(http.Flusher).Flush()
```
yondonfu: Think its good practice (based on the [http.Flusher](https://golang.org/pkg/net/http/#Flusher) docs), to make sure that the `http.ResponseWriter` implements the `http.Flusher` interface i.e.

```golang
if f, ok := w.(http.Flusher); ok {
    f.Flush()
}
```

yondonfu: Why do we need this flush if the multi-part data was already sent prior to this?

yondonfu (changes requested):

monitor/census.go
darkdarkdragon: fixed in 1d573763b9e6c362b8e1473f7a675194395fd9fe

server/mediaserver.go
darkdarkdragon: fixed in e5d86dda116caadf2c6a7d99a94ce2bb63e4215f

darkdarkdragon: `ResponseWriter` is buffered, so at that point last part of the response wasn't sent yet.

yondonfu (approved):
LGTM :shipit: after squashing the fixup commits.

PR #1634, opened 2020-10-19
eth: Remove gas price logging

### LINKED ISSUE #1483: Verbose setting for transcoding related activities only
While running a transcoder and orchestrator in parallel, -v 99 just spams the log with "Cached gas price" every second. 

A verbose setting that only displays transcoding related activities would be nice and helpful for debugging any transcoder related problems.

yondonfu:

This PR removes logging whenever the GasPriceMonitor fetches & caches the gas price returned by a `eth_gasPrice` RPC call.

I don't feel particularly strongly about this change, but it does seem like this log statement provides little to no utility and just results in a sizable (a new log every 5 seconds with the default -blockPollingInterval value) number of logs that need to be sifted through.

An alternative is to increase the verbosity of the log statement from DEBUG to VERBOSE as suggested [here](https://github.com/livepeer/go-livepeer/issues/1483#issuecomment-628173813). Since there isn't really any actionable information in the current gas price log statements, I'm leaning towards just removing it altogether in this PR.

See commit history.

None.

Closes #1483 

**Checklist:**

### DIFF
--- a/eth/gaspricemonitor.go
+++ b/eth/gaspricemonitor.go
@@ -8,7 +8,6 @@ import (
 	"time"
 
 	"github.com/golang/glog"
-	"github.com/livepeer/go-livepeer/common"
 	"github.com/livepeer/go-livepeer/monitor"
 )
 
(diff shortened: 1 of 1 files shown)


jailuthra (approved):
LGTM!

Don't know enough to comment on removing the log altogether v/s switching it to `VERBOSE`, maybe Nico will have some thoughts there.

kyriediculous:
I don't have a strong opinion here either, it's only a bit verbose that during node idle times this is the only log being emitted every polling interval. 

Removing it seems ok to me, I don't really use that log statement anyway.

PR #1637, opened 2020-10-20
cmd, eth, server: remove claim earnings option

### LINKED ISSUE #1635: Remove manual claim earnings option in livepeer_cli

Now that both LIP-36 and LIP-52 are deployed on mainnet, manual earnings claiming is no longer necessary since earnings will be automatically claimed as a part of all staking transactions regardless of how many rounds of earnings need to be claimed. So, the manual earnings claiming option should be removed from livepeer_cli (as is the case in the explorer) to avoid any confusion around whether one needs to manually claim earnings or not.

- Remove the manual claim earnings option in livepeer_cli (https://github.com/livepeer/go-livepeer/blob/master/cmd/livepeer_cli/livepeer_cli.go#L94)
- Remove the manual claim earnings logic related logic in livepeer_cli (https://github.com/livepeer/go-livepeer/blob/master/cmd/livepeer_cli/wizard_bond.go#L336)
- Remove calls to `autoClaimEarnings()` in the [client](https://github.com/livepeer/go-livepeer/blob/master/eth/client.go) staking methods
- Remove the [ClaimEarnings()](https://github.com/livepeer/go-livepeer/blob/master/eth/client.go#L496) method on the client since we can now just use the promoted `ClaimEarnings()` method from the BondingManager contract binding embedded in the client which will just directly submit a `claimEarnings()` transaction

We can leave the `/claimEarnings` webserver endpoint to maintain backwards compatibility with any third party tools that have been developed that make use of that endpoint.

kyriediculous:
Removes the option from the CLI to manually claim earnings 

- 46e7eee >> Removes the manual claim option from the CLI  
- 105a575 >> Removes autoClaimEarnings function from the ETH client and unwraps most functions that made use of it to now use the `BondingManager` contract bindings directly
- 170f099d65f724fe7b32318150e3bd3bb9b06b67 >> Keep the /claimEarnings endpoint around but no longer require an endRound field , ignore it if it is provided and just claim through currentRound 

Fixes #1635 

**Checklist:**

### DIFF
--- a/server/webserver.go
+++ b/server/webserver.go
@@ -692,22 +692,6 @@ func (s *LivepeerServer) cliWebServerHandlers(bindAddr string) *http.ServeMux {
 
 	mux.HandleFunc("/claimEarnings", func(w http.ResponseWriter, r *http.Request) {
 		if s.LivepeerNode.Eth != nil {
-			if err := r.ParseForm(); err != nil {
-				glog.Errorf("Parse Form Error: %v", err)
-				return
-			}
-
-			endRoundStr := r.FormValue("endRound")
-			if endRoundStr == "" {
-				glog.Errorf("Need to provide amount")
-				return
-			}
-			endRound, err := lpcommon.ParseBigInt(endRoundStr)
-			if err != nil {
-				glog.Error(err)
-				return
-			}
-
 			claim := func() error {
 				init, err := s.LivepeerNode.Eth.CurrentRoundInitialized()
 				if err != nil {
@@ -717,11 +701,15 @@ func (s *LivepeerServer) cliWebServerHandlers(bindAddr string) *http.ServeMux {
 				if !init {
 					return errors.New("Round not initialized")
 				}
-				_, err = s.LivepeerNode.Eth.ClaimEarnings(endRound)
+				currRound, err := s.LivepeerNode.Eth.CurrentRound()
 				if err != nil {
 					return err
 				}
-				return nil
+				tx, err := s.LivepeerNode.Eth.ClaimEarnings(currRound)
+				if err != nil {
+					return err
+				}
+				return s.LivepeerNode.Eth.CheckTx(tx)
 			}
 
 			if err := backoff.Retry(claim, backoff.WithMaxRetries(backoff.NewConstantBackOff(time.Second*15), 5)); err != nil {
--- a/server/webserver.go
+++ b/server/webserver.go
@@ -717,7 +717,7 @@ func (s *LivepeerServer) cliWebServerHandlers(bindAddr string) *http.ServeMux {
 				if !init {
 					return errors.New("Round not initialized")
 				}
-				err = s.LivepeerNode.Eth.ClaimEarnings(endRound)
+				_, err = s.LivepeerNode.Eth.ClaimEarnings(endRound)
 				if err != nil {
 					return err
 				}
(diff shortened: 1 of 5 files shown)


server/webserver.go
```diff
@@ -717,7 +701,11 @@ func (s *LivepeerServer) cliWebServerHandlers(bindAddr string) *http.ServeMux {
 				if !init {
 					return errors.New("Round not initialized")
 				}
-				err = s.LivepeerNode.Eth.ClaimEarnings(endRound)
+				currRound, err := s.LivepeerNode.Eth.CurrentRound()
+				if err != nil {
+					return err
+				}
+				_, err = s.LivepeerNode.Eth.ClaimEarnings(currRound)
```
yondonfu: Since the other webserver endpoints for tx submission return after the tx is mined ([example](https://github.com/livepeer/go-livepeer/blob/master/server/webserver.go#L557)), I think it would be good to have that behavior here as well by calling `s.LivepeerNode.CheckTx()` with the transaction returned by `ClaimEarnings()` as the argument.

server/webserver.go
```diff
@@ -692,22 +692,6 @@ func (s *LivepeerServer) cliWebServerHandlers(bindAddr string) *http.ServeMux {
...
-				glog.Errorf("Parse Form Error: %v", err)
-				return
-			}
-
-			endRoundStr := r.FormValue("endRound")
-			if endRoundStr == "" {
-				glog.Errorf("Need to provide amount")
-				return
-			}
-			endRound, err := lpcommon.ParseBigInt(endRoundStr)
```
yondonfu, line 705: Seems like it might be ok to allow a user to still specify an endRound. Don't imagine users needing to do this, but if they have tools that depend on being able to do this for whatever reason it might be good to not break that expectation.

yondonfu (changes requested):

server/webserver.go
kyriediculous: I don't think there's much tooling using this, however it might be okay to force a user's hand in changing their expectation of how this functionality truly behaves currently. If those expectations are not altered somehow it will just end up incurring additional costs for those users and their expectations misalign with what optimal behaviour would be.

yondonfu: I believe there are a few users that have automated claimEarnings calls via this webserver endpoint in the past. Generally, we should avoid breaking API compatibility and we should avoid making assumptions around how users use an API when we can, but I see your point about a user not realizing that there is basically no benefit in calling claimEarnings with an endRound set to anything besides the current round. Will leave this case up to you.

server/webserver.go
```diff
@@ -717,10 +701,17 @@ func (s *LivepeerServer) cliWebServerHandlers(bindAddr string) *http.ServeMux {
...
-				err = s.LivepeerNode.Eth.ClaimEarnings(endRound)
+				currRound, err := s.LivepeerNode.Eth.CurrentRound()
 				if err != nil {
 					return err
 				}
+				tx, err := s.LivepeerNode.Eth.ClaimEarnings(currRound)
+				if err != nil {
+					return err
+				}
+				if err := s.LivepeerNode.Eth.CheckTx(tx); err != nil {
```
yondonfu: We can shorten this to `return s.LivepeerNode.Eth.CheckTx(tx)`.

server/webserver.go
kyriediculous: I actually think it's fine because the API itself doesn't change. Users can still call the same endpoint with an argument (it will just be ignored).

If there are tools out there that automate claimEarnings they have become rather obsolete anyway even it's for up until the current round. 

If a user needs funds it's more economical to perform a  staking action for example unbond, or withdraw fees directly.
From that perspective it make even make sense to just remove the endpoint entirely, while it might misalign with user expectations (we can clarify this in the release notes) it protects users from operating inefficiently.

server/webserver.go
kyriediculous: Good point, fixed c6cb779

yondonfu (approved):
LGTM after squashing fixup commits

Issue #1640, opened 2020-10-22
broadcasters stuck after bandwidth pressure


iameli:
**Describe the bug**
@ericxtang and I were running a big load test. After about 400ish streams, something broke that we're still investigating. The working theory is that we ran out of bandwidth in the transcoding cluster.

Afterward, lots of the broadcasters in the cluster believed they still had streams. There was no data flowing into the B, but `GET /status` still showed a lot in there, and the logs for the B showed that they were consistently doing a watchdog reset:

```
I1022 05:22:28.632438       1 mediaserver.go:771] watchdog reset mid=[redacted] seq=1 dur=2000 started=2020-10-22 04:42:52.596496644 +0000 UTC m=+5627.615492239
I1022 05:22:28.902201       1 mediaserver.go:771] watchdog reset mid=[redacted] seq=4 dur=1600 started=2020-10-22 04:58:10.896452534 +0000 UTC m=+6545.915448149
I1022 05:22:29.261122       1 mediaserver.go:771] watchdog reset mid=[redacted] seq=158 dur=2000 started=2020-10-22 04:42:53.2199255 +0000 UTC m=+5628.238921105
I1022 05:22:38.652812       1 mediaserver.go:771] watchdog reset mid=[redacted] seq=145 dur=2000 started=2020-10-22 04:42:08.642852375 +0000 UTC m=+5583.661847980
I1022 05:22:44.069157       1 mediaserver.go:771] watchdog reset mid=[redacted] seq=163 dur=2000 started=2020-10-22 04:42:14.058966498 +0000 UTC m=+5589.077962124
I1022 05:22:54.067318       1 mediaserver.go:771] watchdog reset mid=[redacted] seq=213 dur=2000 started=2020-10-22 04:42:24.050308485 +0000 UTC m=+5599.069304070
```

[Full unredacted logs and `/status` output in internal Livepeer Discord here](https://discordapp.com/channels/423160867534929930/646030422618144779/768708036017389568).

Seems like maybe there's a failure condition that we're not considering that's causing the session to remain open?

**To Reproduce**
I don't yet have a repro outside of stressing one of our clusters in a bandwidth-constrained setting

**Expected behavior**
The streams should timeout and return to 0

PR #1644, opened 2020-10-22
Fix memory leak

darkdarkdragon:
Fixes memory leak caused by unbuffered channels

Added buffers to channels

unit tests

**Checklist:**

### DIFF
--- a/discovery/db_discovery.go
+++ b/discovery/db_discovery.go
@@ -240,7 +240,7 @@ func (dbo *DBOrchestratorPoolCache) cacheDBOrchs() error {
 		return fmt.Errorf("could not retrieve orchestrators from DB: %v", err)
 	}
 
-	resc, errc := make(chan *common.DBOrch), make(chan error)
+	resc, errc := make(chan *common.DBOrch, len(orchs)), make(chan error, len(orchs))
 	ctx, cancel := context.WithTimeout(context.Background(), getOrchestratorsTimeoutLoop)
 	defer cancel()
 
--- a/discovery/db_discovery.go
+++ b/discovery/db_discovery.go
@@ -164,7 +164,7 @@ func (dbo *DBOrchestratorPoolCache) cacheOrchestratorStake() error {
 		return fmt.Errorf("could not retrieve orchestrators from DB: %v", err)
 	}
 
-	resc, errc := make(chan *common.DBOrch), make(chan error)
+	resc, errc := make(chan *common.DBOrch, len(orchs)), make(chan error, len(orchs))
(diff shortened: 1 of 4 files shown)


discovery/discovery_test.go
```diff
@@ -25,6 +25,7 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/mock"
 	"github.com/stretchr/testify/require"
+	"go.uber.org/goleak"
```
iameli, line 28: slick name. this makes me want to do `go.livepeer.org/whatever` for purely cosmetic reasons

iameli:
Interesting that we've never noticed this. When would it show up?

discovery/discovery_test.go
darkdarkdragon: I like the idea 😃

iameli: actually we totally should, looks like it's as easy as https://github.com/GoogleCloudPlatform/govanityurls

darkdarkdragon:
> Interesting that we've never noticed this. When would it show up?

Found it in the logs for #1640 
It leaks only when `GetOrchestrator` request takes more than 3 second. And to notice this we should add alert on number of goroutines.

discovery/discovery_test.go
darkdarkdragon: Now we need package to publish 😃

discovery/db_discovery.go
```diff
@@ -240,7 +240,7 @@ func (dbo *DBOrchestratorPoolCache) cacheDBOrchs() error {
 		return fmt.Errorf("could not retrieve orchestrators from DB: %v", err)
 	}
 
-	resc, errc := make(chan *common.DBOrch), make(chan error)
+	resc, errc := make(chan *common.DBOrch, len(orchs)), make(chan error, len(orchs))
```
yondonfu, line 243: So I think there is actually a bug with the timeout behavior because on L284 when a timeout is triggered the break statement will break out of the select statement and not the for loop. As a result, when the timeout is triggered control flow will just move onto the next iteration of the for loop. If that is the case, then I don't think there was a memory leak since the for loop would run N times once for each of the N goroutines created for an O. 

That being said, if we fix the timeout behavior then there would be a memory leak without this change so it might make sense to fix the timeout behavior alongside this change in this PR.

darkdarkdragon: >  loop would run N times

loop will run N times but will read from channels N-1 times -  because one time will spent in `case <-ctx.Done()`
Anyway, fixed it

yondonfu (approved):
LGTM

PR #1649, opened 2020-10-29
server: fix current round locked check in activateOrchestrator

### LINKED ISSUE #1648: CLI: Activate orchestrator always results in "error: current round is locked"
**Describe the bug**
After Invoke multi-step "become an orchestrator" follow with
Error activating orchestrator: current round is locked.

The round is not actually locked 

**Expected behavior**
Should only throw error when the roundi s locked 

**Additional context**
the following code https://github.com/livepeer/go-livepeer/blob/9d47b956559b00d07951dd4ae29f4d03969bf08f/server/webserver.go#L234 results in a false positive for the round being locked 
```golang
		ok, err := s.LivepeerNode.Eth.CurrentRoundLocked()
		if err != nil {
			respondWith500(w, err.Error())
			return
		}
		if !ok {
			respondWith500(w, "current round is locked")
			return
		}

```

`if !ok`  should be `if ok` but we should also rename `ok` to `isLocked` for readability

kyriediculous:
Fix current round locked check in `/activateOrchestrator` endpoint for the CLI server which will always respond that the current round is locked 

Fixes #1648 

**Checklist:**

### DIFF
--- a/server/webserver.go
+++ b/server/webserver.go
@@ -231,12 +231,12 @@ func (s *LivepeerServer) cliWebServerHandlers(bindAddr string) *http.ServeMux {
 			return
 		}
 
-		ok, err := s.LivepeerNode.Eth.CurrentRoundLocked()
+		isLocked, err := s.LivepeerNode.Eth.CurrentRoundLocked()
 		if err != nil {
 			respondWith500(w, err.Error())
 			return
 		}
-		if !ok {
+		if isLocked {
 			respondWith500(w, "current round is locked")
 			return
 		}
--- a/eth/stubclient.go
+++ b/eth/stubclient.go
@@ -181,6 +181,8 @@ type StubClient struct {
 	CheckTxErr                   error
 	TotalStake                   *big.Int
 	TranscoderPoolError          error
+	RoundLocked                  bool
+	RoundLockedErr               error
 }
 
 type stubTranscoder struct {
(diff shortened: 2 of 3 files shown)


server/webserver.go
```diff
@@ -231,12 +231,12 @@ func (s *LivepeerServer) cliWebServerHandlers(bindAddr string) *http.ServeMux {
 			return
 		}
 
-		ok, err := s.LivepeerNode.Eth.CurrentRoundLocked()
+		isLocked, err := s.LivepeerNode.Eth.CurrentRoundLocked()
 		if err != nil {
 			respondWith500(w, err.Error())
 			return
 		}
-		if !ok {
```
yondonfu, line 239: Since this is something a test would've caught what do you think about adding a test case that a) fails without this change and b) passes with this change?

kyriediculous: Added in e8e0bb51 along with some form param tests , decided to omit some other stuff would take quite a bit of time to add 100% coverage due to all the several codepaths and is not relevant for this PR.

kyriediculous: Not sure why the subsequent tests fail all of a sudden still, I made sure to properly close the server and request bodies in 133b459

yondonfu: > Not sure why the subsequent tests fail all of a sudden still

915349c should fix those tests. Previously, e8e0bb5 passed `eth.StubClient{}` as an arg to the LivepeerNode constructor in `newMockServer()` which caused test cases in `TestGetEthChainID` that expected LivepeerNode to be in off-chain mode (the test cases for error cases) to break.

yondonfu (approved):
LGTM pending green CI and squashing fixup commits.

PR #1653, opened 2020-11-05
db: prevent nil pointer error if no rows in sql query result

kyriediculous:
If there are no rows in the SQL result set then `rows` will/could be `nil` and calling `rows.Close()` before handling the error could result in a nil pointer error. 

SQLite seems to handle this well but I ran into it using postgres, might be good riddance regardless. 

**Checklist:**


yondonfu (approved):
LGTM

PR #1670, opened 2020-11-26
Record and log additional metrics for O

### LINKED ISSUE #1661: Add logs for and metrics for real-time ratio and segment duration on O/T
To be filled in.

yondonfu:

Record and log the following metrics on O:
- Source segment duration
- Download time for source segment
- Transcode score for a segment which is defined as source segment duration / transcode time (this is added for standalone T as well)

B downloads the transcoded results from O separately and the associated HTTP handler lives in LPMS so I did not try to add metrics instrumentation there for now. As a result, this PR does not include the ability to measure the ratio between source segment duration and round trip time from O's POV.

A few other minor updates:
- 1702ba5 uses a NodeType type instead of a string in the `monitor` package. I felt that this would be less prone to error when writing conditionals based on the current node type as is done in 7a05697 which skips extra state tracking code in `SegmentEmerged` if the function is used for a node type other than a broadcaster
- 321f5d0 tweaks the view distribution for the source segment duration metric to make the percentile graph of source segment duration easier to understand

See commit history.

Tested manually and observed metrics using livepeer/docker-livepeer#44.

Fixes #1661 

**Checklist:**

### DIFF
--- a/server/segment_rpc.go
+++ b/server/segment_rpc.go
@@ -80,6 +80,12 @@ func (h *lphttp) ServeSegment(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	glog.V(common.VERBOSE).Infof("Received segment manifestID=%s sessionID=%s seqNo=%d dur=%v", segData.ManifestID, segData.AuthToken.SessionId, segData.Seq, segData.Duration)
+
+	if monitor.Enabled {
+		monitor.SegmentEmerged(0, uint64(segData.Seq), len(segData.Profiles), segData.Duration.Seconds())
+	}
+
 	if err := orch.ProcessPayment(payment, core.ManifestID(segData.AuthToken.SessionId)); err != nil {
 		glog.Errorf("error processing payment: %v", err)
 		http.Error(w, err.Error(), http.StatusBadRequest)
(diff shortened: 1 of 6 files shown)


server/segment_rpc.go
```diff
@@ -518,7 +532,7 @@ func SubmitSegment(sess *BroadcastSession, seg *stream.HLSSegment, nonce uint64)
 
 	// transcode succeeded; continue processing response
 	if monitor.Enabled {
-		monitor.SegmentTranscoded(nonce, seg.SeqNo, transcodeDur, common.ProfilesNames(params.Profiles))
+		monitor.SegmentTranscoded(nonce, seg.SeqNo, time.Duration(seg.Duration)*time.Second, transcodeDur, common.ProfilesNames(params.Profiles))
```
darkdarkdragon:
```suggestion
		monitor.SegmentTranscoded(nonce, seg.SeqNo, time.Duration(seg.Duration*float64(time.Second)), transcodeDur, common.ProfilesNames(params.Profiles))
```

darkdarkdragon (changes requested):
One small change
and not for the future: we're using `dur=` in logs both for segment's duration and network transfer duration. I think it would be good to separate them, use `segDur=` for example for segment's duration (not in this PR of course).

server/segment_rpc.go
yondonfu: Fixed in 91b3825

yondonfu:
Fixed a `monitor` package test in 58ffeed that was failing due to the broadcaster only conditional check in `SegmentEmerged`.

darkdarkdragon (approved):
LGTM

Issue #1658, opened 2020-11-16
More flexible segment in flight rule


yondonfu:
Both B & O implement a single segment in flight (SSIF) rule for streams. B will not send another segment to O if it has not responded for a previous segment for the stream yet. O will not accept another segment if it has not finished transcoding a previous segment for a transcode session yet. As a result, for a stream with uniform segment durations, an O will receive all segments in the stream as long as it returns results in real-time. However, there are other considerations to take into account in practice:

- If a stream has variable segment durations (which is not uncommon), segment N + 1 might be shorter than segment N such that even if O can return segment N in real-time, segment N + 1 could be emitted before O returns segment N
- O is slower to transcode the first segment in a transcode session due to GPU session init overhead
- O might return results in real-time most of the time, but slower than real-time sometimes due to factors such as fluctuation in bandwidth capacity
- Combination of all of the above

The points listed above can cause the following problems:
- If B only has access to a single O, it will drop a bunch segments even if those segments could have been transcoded by O in real-time most of the time and an acceptable amount slower than real-time other times
- If B has access to many Os, it will switch Os more often and each switch can potentially trigger a GPU session init which will slow down the transcode for that segment

One way to address these problems might be to update the segment in flight rule to something like:

- If a segment is in flight for O:
    - If the time in flight for the segment is less than `x`:
        - Send the next segment to this O
    - Else:
        - If there are other Os available:
            - Send the next segment to another O
        - Else:
            - If the time in flight for the segment is less than `y`:
                - Send the next segment to this O
            - Else:
                - Drop the segment

The above assumes that the current O would be the one selected for the next segment by B's selection algo. 

Since this might result in multiple segments in flight, we could keep track of the time in flight for the oldest segment in flight.

`x` is the time in flight threshold for continuing to send segments to an O. A default value for this might be the duration of the oldest segment that is in flight. 

`y` is the time in flight threshold for continuing to send segments to an O if the requirement for `x` is not met and if there are other Os available. This value controls how often segments get dropped.

A few implications of this proposed rule if the suggested default value for `x` is used:
- Regardless of the durations of segments in a stream, if an O can return results for each segment in real-time it will receive all the segments of the stream.
- If O fluctuates between real-time and slightly slower than real-time, it can still consistently receive segments of the stream if there are no other Os available which reduces the # of dropped segments
- If there are other Os available, B will still switch if the current Os is not keeping up in real-time

This would require changes to the segment in flight management rule on both B & O.

[...]

PR #1672, opened 2020-11-30
server/pushapi: continue with old RTMP connection

### LINKED ISSUE #1665: HTTP 500 errors when B is overloaded through Push API
**Describe the bug**
The Broadcaster node sometimes emits HTTP internal error responses (500) without any additional logs when overloaded with input segments through the Push API.

**Reproduction**
1. Start up a local B/OT setup.
2. Send two segments *simultaneously* for a new stream via the Push API by calling parallel curl requests:
```
curl -XPOST --data-binary @test.ts -v http://127.0.0.1:8936/live/test/0.ts -H "accept: multipart/mixed" -i --output out.ts &\
curl -XPOST --data-binary @test.ts -v http://127.0.0.1:8936/live/test/1.ts -H "accept: multipart/mixed" -i --output out2.ts
```

3. One of the segments will successfully return transcoded renditions with an HTTP 200
4. The other segment will not be transcoded and curl will output an HTTP 500 due to the StreamAlreadyExists error at https://github.com/livepeer/go-livepeer/blob/7f9534c365123f455c7fd3bfb9ebcbf5a99cdd43/server/mediaserver.go#L517

jailuthra:

If an old connection is found later due to segments arriving too closely, continue with that connection instead of returning a 500 to the client.

- Improve handling for simultaneous segments at the start of a push api stream 

1. Followed the reproduction steps at #1665
2. Got a 500 before this patch.
3. Apply this patch
4. Get a 503 "Service Unavailable" due to "No sessions available" instead

Partly handles #1665 
For actually handling simultaneous segments instead of throwing the 503, need to implement changes mentioned in #1658 

**Checklist:**

### DIFF
--- a/server/mediaserver.go
+++ b/server/mediaserver.go
@@ -291,8 +291,14 @@ func createRTMPStreamIDHandler(s *LivepeerServer) func(url *url.URL) (strmID str
 			return nil
 		}
 		if _, exists := s.rtmpConnections[mid]; exists {
-			glog.Errorf("Manifest manifestID=%v already exists for streamID url=%s", mid, url.String())
-			return nil
+			if extmid != mid {
+				// only error out if we're seeing segments from different extmids but same mid trying to use the same connection
+				// FIXME s.internalManifests might not be updated if segments are too close together
+				if _, extmidSeenBefore := s.internalManifests[extmid]; !extmidSeenBefore {
+					glog.Errorf("RTMP connection for manifest manifestID=%v already exists (streamID url=%s)", mid, url.String())
+					return nil
+				}
+			} // by default we will handle re-using the extmid's older connection in HandlePush
 		}
 
 		// Generate RTMP part of StreamID
@@ -541,6 +547,7 @@ func (s *LivepeerServer) registerConnection(rtmpStrm stream.RTMPVideoStream) (*r
 	if exists {
 		// We can only have one concurrent stream per ManifestID
 		s.connectionLock.Unlock()
+		cxn.sessManager.cleanup()
 		return oldCxn, errAlreadyExists
 	}
 	s.rtmpConnections[mid] = cxn
@@ -765,12 +772,13 @@ func (s *LivepeerServer) HandlePush(w http.ResponseWriter, r *http.Request) {
 
 		cxn, err = s.registerConnection(st)
 		if err != nil {
+			st.Close()
 			if err != errAlreadyExists {
 				httpErr := fmt.Sprintf("http push error url=%s err=%v", r.URL, err)
 				glog.Error(httpErr)
 				http.Error(w, httpErr, http.StatusInternalServerError)
 				return
-			}
+			} // else we continue with the old cxn
 		} else {
 			// Start a watchdog to remove session after a period of inactivity
 			ticker := time.NewTicker(httpPushTimeout)
@@ -790,6 +798,8 @@ func (s *LivepeerServer) HandlePush(w http.ResponseWriter, r *http.Request) {
 				}
 			}(s, cxn.mid, mid)
 		}
+		// Regardless of old/new cxn returned by registerConnection, we make sure
+		// our internalManifests mapping is OK before moving on
 		if cxn.mid != mid {
 			// AuthWebhook provided different ManifestID
 			s.connectionLock.Lock()
--- a/server/mediaserver.go
+++ b/server/mediaserver.go
@@ -511,11 +511,11 @@ func (s *LivepeerServer) registerConnection(rtmpStrm stream.RTMPVideoStream) (*r
 	hlsStrmID := core.MakeStreamID(mid, &vProfile)
 	s.connectionLock.RLock()
 	// Fast path - check early if session exists - creating new session can take time
-	_, exists := s.rtmpConnections[mid]
+	oldCxn, exists := s.rtmpConnections[mid]
 	s.connectionLock.RUnlock()
 	if exists {
 		// We can only have one concurrent stream per ManifestID
-		return nil, errAlreadyExists
+		return oldCxn, errAlreadyExists
 	}
 
 	playlist := core.NewBasicPlaylistManager(mid, storage, recordStorage)
@@ -535,13 +535,13 @@ func (s *LivepeerServer) registerConnection(rtmpStrm stream.RTMPVideoStream) (*r
 	}
 
 	s.connectionLock.Lock()
-	_, exists = s.rtmpConnections[mid]
+	oldCxn, exists = s.rtmpConnections[mid]
 	// Check if session exist again - potentially two sessions can be created simultaneously,
 	// so we don't want to overwrite one that was already created
 	if exists {
 		// We can only have one concurrent stream per ManifestID
 		s.connectionLock.Unlock()
-		return nil, errAlreadyExists
+		return oldCxn, errAlreadyExists
 	}
 	s.rtmpConnections[mid] = cxn
 	s.lastManifestID = mid
@@ -765,29 +765,31 @@ func (s *LivepeerServer) HandlePush(w http.ResponseWriter, r *http.Request) {
 
 		cxn, err = s.registerConnection(st)
 		if err != nil {
-			httpErr := fmt.Sprintf("http push error url=%s err=%v", r.URL, err)
-			glog.Error(httpErr)
-			http.Error(w, httpErr, http.StatusInternalServerError)
-			return
-		}
-
-		// Start a watchdog to remove session after a period of inactivity
-		ticker := time.NewTicker(httpPushTimeout)
-		go func(s *LivepeerServer, intmid, extmid core.ManifestID) {
-			defer ticker.Stop()
-			for range ticker.C {
-				var lastUsed time.Time
-				s.connectionLock.RLock()
-				if cxn, exists := s.rtmpConnections[intmid]; exists {
-					lastUsed = cxn.lastUsed
-				}
-				s.connectionLock.RUnlock()
-				if time.Since(lastUsed) > httpPushTimeout {
-					_ = removeRTMPStream(s, extmid)
-					return
-				}
+			if err != errAlreadyExists {
+				httpErr := fmt.Sprintf("http push error url=%s err=%v", r.URL, err)
+				glog.Error(httpErr)
+				http.Error(w, httpErr, http.StatusInternalServerError)
+				return
 			}
-		}(s, cxn.mid, mid)
+		} else {
+			// Start a watchdog to remove session after a period of inactivity
+			ticker := time.NewTicker(httpPushTimeout)
+			go func(s *LivepeerServer, intmid, extmid core.ManifestID) {
+				defer ticker.Stop()
+				for range ticker.C {
+					var lastUsed time.Time
+					s.connectionLock.RLock()
+					if cxn, exists := s.rtmpConnections[intmid]; exists {
+						lastUsed = cxn.lastUsed
+					}
+					s.connectionLock.RUnlock()
+					if time.Since(lastUsed) > httpPushTimeout {
+						_ = removeRTMPStream(s, extmid)
+						return
+					}
+				}
+			}(s, cxn.mid, mid)
+		}
 		if cxn.mid != mid {
 			// AuthWebhook provided different ManifestID
 			s.connectionLock.Lock()
--- a/go.mod
+++ b/go.mod
@@ -33,7 +33,7 @@ require (
 	github.com/jackpal/go-nat-pmp v1.0.1 // indirect
 	github.com/karalabe/usb v0.0.0-20190919080040-51dc0efba356 // indirect
 	github.com/livepeer/lpms v0.0.0-20201130142735-967c18793f06
-	github.com/livepeer/m3u8 v0.11.0
+	github.com/livepeer/m3u8 v0.11.1
 	github.com/mattn/go-colorable v0.1.2 // indirect
 	github.com/mattn/go-sqlite3 v1.11.0
 	github.com/mohae/deepcopy v0.0.0-20170929034955-c48cc78d4826 // indirect
--- a/go.sum
+++ b/go.sum
@@ -213,8 +213,9 @@ github.com/livepeer/joy4 v0.1.2-0.20191121080656-b2fea45cbded h1:ZQlvR5RB4nfT+cO
 github.com/livepeer/joy4 v0.1.2-0.20191121080656-b2fea45cbded/go.mod h1:xkDdm+akniYxVT9KW1Y2Y7Hso6aW+rZObz3nrA9yTHw=
 github.com/livepeer/lpms v0.0.0-20201130142735-967c18793f06 h1:g8BFFP464d+vlc6hyouO4IaqDy4VAEn/9OZfruBMRwc=
 github.com/livepeer/lpms v0.0.0-20201130142735-967c18793f06/go.mod h1:Vd2O4BVQI/Yj2hzjnNcyTj+qib8zb0hh7GunKP6KJc8=
-github.com/livepeer/m3u8 v0.11.0 h1:aI2hLXV5h5VqxjjmAOs55TpUR35KzNL2XWLkbETql5g=
 github.com/livepeer/m3u8 v0.11.0/go.mod h1:IUqAtwWPAG2CblfQa4SVzTQoDcEMPyfNOaBSxqHMS04=
+github.com/livepeer/m3u8 v0.11.1 h1:VkUJzfNTyjy9mqsgp5JPvouwna8wGZMvd/gAfT5FinU=
+github.com/livepeer/m3u8 v0.11.1/go.mod h1:IUqAtwWPAG2CblfQa4SVzTQoDcEMPyfNOaBSxqHMS04=
 github.com/mattn/go-colorable v0.1.2 h1:/bC9yWikZXAL9uJdulbSfyVNIR3n3trXl+v8+1sx8mU=
 github.com/mattn/go-colorable v0.1.2/go.mod h1:U0ppj6V5qS13XJ6of8GYAs25YV2eR4EVcfRqFIhoBtE=
 github.com/mattn/go-isatty v0.0.8 h1:HLtExJ+uU2HOZ+wI0Tt5DtUDrx8yhUqDcp7fYERX4CE=
--- a/server/push_test.go
+++ b/server/push_test.go
@@ -1018,3 +1018,43 @@ func TestPush_OSPerStream(t *testing.T) {
 	body, _ = ioutil.ReadAll(resp.Body)
 	assert.True(len(body) > 0)
 }
+
+func TestPush_ConcurrentSegments(t *testing.T) {
+	assert := assert.New(t)
+
+	drivers.NodeStorage = drivers.NewMemoryDriver(nil)
+	n, _ := core.NewLivepeerNode(nil, "./tmp", nil)
+	n.NodeType = core.BroadcasterNode
+	s, _ := NewLivepeerServer("127.0.0.1:1938", n, true, "")
+	oldURL := AuthWebhookURL
+	defer func() { AuthWebhookURL = oldURL }()
+	AuthWebhookURL = ""
+
+	var wg sync.WaitGroup
+	start := make(chan struct{})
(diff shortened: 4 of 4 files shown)


server/mediaserver.go
```diff
@@ -764,29 +764,31 @@ func (s *LivepeerServer) HandlePush(w http.ResponseWriter, r *http.Request) {
...
+				httpErr := fmt.Sprintf("http push error url=%s err=%v", r.URL, err)
+				glog.Error(httpErr)
+				http.Error(w, httpErr, http.StatusInternalServerError)
+				return
 			}
-		}(s, cxn.mid, mid)
+		} else {
+			// Start a watchdog to remove session after a period of inactivity
+			ticker := time.NewTicker(httpPushTimeout)
+			go func(s *LivepeerServer, intmid, extmid core.ManifestID) {
```
yondonfu, line 785: If an external manifestID foo is provided with the same internal manifestID bar as another external manifestID baz that was previously provided then a watchdog will call `removeRTMPStream()` with baz, but not foo. As a result, `internalManifests` will contain an entry for foo due to L795, but that entry will never be cleared since the [map deletion](https://github.com/livepeer/go-livepeer/blob/7f9534c365123f455c7fd3bfb9ebcbf5a99cdd43/server/mediaserver.go#L577) in `removeRTMPStream()` will never be called for foo.

yondonfu (changes requested):
> For actually handling simultaneous segments instead of throwing the 503, need to implement changes mentioned in #1658

Until #1658 is addressed, this can be somewhat mitigated if B is working with multiple Os.

Isn't it still possible to trigger the manifestID already exists error described in #1673 in `createRTMPStreamIDHandler()` (which will use the internal manifestID returned by the auth webhook to index into rtmpConnections) since this PR only addresses the StreamAlreadyExists error returned by `registerConnection()`?

Can we also add unit tests for the updated behavior?

server/mediaserver.go
jailuthra: Ah good catch! I also now notice that I'm not cleaning up the `BasicRTMPVideoStream` created for "foo" here as well. Will fix thanks.

jailuthra: Oh wait my bad. I think that case won't happen within this PR's context. Different external streamID but same internal manifestID is a separate case altogether, that I've moved to separate issue #1673 and will handle through a separate PR.

Within this PR a segment with a different externalID but same internal manifestID should never even reach L795. The case will get returned with a 500 during `createRTMPStreamIDHandler` itself (for now)

> I'm not cleaning up the BasicRTMPVideoStream created for "foo" here as well. Will fix thanks.

But this still applies for the RTMP struct created for the 'late' segment in the two simultaneous ones arriving for this case. S still a good catch :smile:

jailuthra:
> Until 1658 is addressed, this can be somewhat mitigated if B is working with multiple Os.

If I understood this correctly, the B would find no open sessions for the same O it sent the first segment too, and thus will get a 503. But if it knows multiple Os, it will send the second segment to another O? If yes then great!! I'll test this out locally as well.

> Isn't it still possible to trigger the manifestID already exists error described in 1673 in createRTMPStreamIDHandler() (which will use the internal manifestID returned by the auth webhook to index into rtmpConnections) since this PR only addresses the StreamAlreadyExists error returned by registerConnection()?

Yeah I never really saw the "manifest already exists" error happening with two segments of the same externalID - neither in staging/prod logs nor during local testing by sending two segments close together for same stream session.

It might be theoretically possible that one segment passes the `exists {` check in createRTMPStreamIDHandler() while the other one does not because by that time the first segment has added itself in the map inside `registerConnection`. So I'll think about how to handle this as well.

But I was hoping to get this merged fast (at least in staging) to see how well it does - as most of the same-session-simultaneous-segments I saw caused the error inside `registerConnection()`. The "manifest already exists" only seemed to happen for different external IDs (and client not respecting 1 minute timeout at B).

server/mediaserver.go
darkdarkdragon: Also potentially new playlist and new session manager could be created:
```go
	playlist := core.NewBasicPlaylistManager(mid, storage, recordStorage)
```
```go
		sessManager: NewSessionManager(s.LivepeerNode, params, NewMinLSSelector(stakeRdr, 1.0)),
```
both should be cleaned up like in `removeRTMPStream`:
```go
	cxn.stream.Close()
	cxn.sessManager.cleanup()
	cxn.pl.Cleanup()
```

yondonfu: IMO it might be easier to read if the function that creates the resource is also responsible for cleaning it up when an error is triggered.

So, in `HandlePush()`

```golang

...

if err != nil {
    st.Close()
    if err != errAlreadyExists {
        ...
    }
}
```

and then in `registerConnection()` we can cleanup the SessionManager and the PlaylistManager if we hit an error condition.

yondonfu (commented):
> If I understood this correctly, the B would find no open sessions for the same O it sent the first segment too, and thus will get a 503. But if it knows multiple Os, it will send the second segment to another O?

Yep.

Also re: unit test - see [push_test.go](https://github.com/livepeer/go-livepeer/blob/master/server/push_test.go) for some examples of programatically testing the push API. Would be nice to have a test case for concurrent pushes since this is what caused the issue in the first place.

jailuthra:
> Also re: unit test - see push_test.go for some examples of programatically testing the push API. Would be nice to have a test case for concurrent pushes since this is what caused the issue in the first place.

Thanks, I missed your comment around the unit tests the first time around. I'll try to replicate my simultaneous `curl` pushes in `push_test.go`. Finally get to write cool goroutine concurrency from scratch woo!!

server/mediaserver.go
```diff
@@ -540,6 +540,8 @@ func (s *LivepeerServer) registerConnection(rtmpStrm stream.RTMPVideoStream) (*r
 	if exists {
 		// We can only have one concurrent stream per ManifestID
 		s.connectionLock.Unlock()
+		cxn.sessManager.cleanup()
```
jailuthra, line 550: @darkdarkdragon After doing the cleanup changes here - I noticed that instead of getting a 503 that I was getting before - I am now again getting 500s due to a new reason.

Closing the session manager created by the *second* segment, closes the B session for the old connection too for some reason. I don't understand why exactly this is happening, but I'll do a quick change without this cleanup call here and see if it works out.

jailuthra: I get
```
E1205 00:32:59.274014 1204668 mediaserver.go:857] http push error processing segment url=http://127.0.0.1:8936/live/test/1.ts manifestID=test err=Session ended
```
And sending any segment again on the same streamID too I keep getting this. So I think cleaning up the newly created session here is somehow closing both the old and new sessions.

jailuthra:
To handle the issue mentioned with broadcaster-session deletion here https://github.com/livepeer/go-livepeer/pull/1672#discussion_r536316994 - I have opted to move the B playlist and sessionmanager creation within the RW lock in https://github.com/livepeer/go-livepeer/pull/1672/commits/4bd6201f2435c65a840c67751f9dccae7baa5443

That solved the "Session ended" errors, albeit adds an extra ~10ms delay for a new connection if it competes for the RW lock (from my local testing)

>> Isn't it still possible to trigger the manifestID already exists error described in #1673 in createRTMPStreamIDHandler()

> It might be theoretically possible that one segment passes the exists { check in createRTMPStreamIDHandler() while the other one does not because by that time the first segment has added itself in the map inside registerConnection. So I'll think about how to handle this as well.

@yondonfu my unit test did end up giving a rare 500 on this though, so for now I've added a `mid != extmid` check in `createRTMPStreamIDHandler`, and pass on for this PRs case of `mid == extmid` as we anyway close the stream later when we find a connection exists.

After all of this I am unable to reproduce the 500s in #1665 by either my unit test or via `curl`.

But my unit test does fail when ran with `-race` due to contention of both segments at InsertHLSSegment()
WDYT my next step should be here?

```
=== RUN   TestPush_ConcurrentSegments
==================
WARNING: DATA RACE
Read at 0x00c000bfc158 by goroutine 77:
  github.com/livepeer/m3u8.(*MediaPlaylist).Count()
      /home/darkapex/go/pkg/mod/github.com/livepeer/m3u8@v0.11.0/writer.go:733 +0x14a
  github.com/livepeer/go-livepeer/core.(*BasicPlaylistManager).InsertHLSSegment()
      /home/darkapex/go/src/github.com/livepeer/go-livepeer/core/playlistmanager.go:287 +0x174
  github.com/livepeer/go-livepeer/server.processSegment()
      /home/darkapex/go/src/github.com/livepeer/go-livepeer/server/broadcast.go:329 +0x981
  github.com/livepeer/go-livepeer/server.(*LivepeerServer).HandlePush()
      /home/darkapex/go/src/github.com/livepeer/go-livepeer/server/mediaserver.go:841 +0x1547
  github.com/livepeer/go-livepeer/server.(*LivepeerServer).HandlePush-fm()
      /home/darkapex/go/src/github.com/livepeer/go-livepeer/server/mediaserver.go:684 +0x64
  net/http.HandlerFunc.ServeHTTP()
      /usr/lib/go/src/net/http/server.go:2042 +0x51
  github.com/livepeer/go-livepeer/server.TestPush_ConcurrentSegments.func2()
      /home/darkapex/go/src/github.com/livepeer/go-livepeer/server/push_test.go:1042 +0x38a

Previous write at 0x00c000bfc158 by goroutine 67:
  github.com/livepeer/m3u8.(*MediaPlaylist).appendSegment()
      /home/darkapex/go/pkg/mod/github.com/livepeer/m3u8@v0.11.0/writer.go:343 +0x67a
  github.com/livepeer/m3u8.(*MediaPlaylist).InsertSegment()
      /home/darkapex/go/pkg/mod/github.com/livepeer/m3u8@v0.11.0/writer.go:395 +0x236
  github.com/livepeer/go-livepeer/core.(*BasicPlaylistManager).InsertHLSSegment()
      /home/darkapex/go/src/github.com/livepeer/go-livepeer/core/playlistmanager.go:294 +0x1b3
  github.com/livepeer/go-livepeer/server.processSegment()
      /home/darkapex/go/src/github.com/livepeer/go-livepeer/server/broadcast.go:329 +0x981
  github.com/livepeer/go-livepeer/server.(*LivepeerServer).HandlePush()
      /home/darkapex/go/src/github.com/livepeer/go-livepeer/server/mediaserver.go:841 +0x1547
  github.com/livepeer/go-livepeer/server.(*LivepeerServer).HandlePush-fm()
      /home/darkapex/go/src/github.com/livepeer/go-livepeer/server/mediaserver.go:684 +0x64
  net/http.HandlerFunc.ServeHTTP()
      /usr/lib/go/src/net/http/server.go:2042 +0x51
  github.com/livepeer/go-livepeer/server.TestPush_ConcurrentSegments.func2()
      /home/darkapex/go/src/github.com/livepeer/go-livepeer/server/push_test.go:1042 +0x38a

Goroutine 77 (running) created at:
  github.com/livepeer/go-livepeer/server.TestPush_ConcurrentSegments()
      /home/darkapex/go/src/github.com/livepeer/go-livepeer/server/push_test.go:1054 +0x4a4
  testing.tRunner()
      /usr/lib/go/src/testing/testing.go:1123 +0x202

Goroutine 67 (running) created at:
  github.com/livepeer/go-livepeer/server.TestPush_ConcurrentSegments()
      /home/darkapex/go/src/github.com/livepeer/go-livepeer/server/push_test.go:1053 +0x472
  testing.tRunner()
      /usr/lib/go/src/testing/testing.go:1123 +0x202
==================
    testing.go:1038: race detected during execution of test
--- FAIL: TestPush_ConcurrentSegments (0.30s)
```

jailuthra:
In case the data race issue is actually bad, the next *'clean'* option I see is to move all of the connection creation code below within the RW mutex `connectionLock`

https://github.com/livepeer/go-livepeer/blob/7138591ec3ee61d49afe1f4fa3720539f30c72ae/server/mediaserver.go#L747-L798

That would certainly cause a block to happen for ~10/20ms for every new stream session, but I don't see a better route as clearly the earlier idea of keeping minimum-viable locks is still leading to race-conditions in production.

EDIT: Or we could return a 500 for one of the segments and let the client retry like we already do :P 

EDIT2: I also see this in go docs https://golang.org/doc/articles/race_detector.html#Excluding_Tests but I don't know if it applies in this case ¯\_ (ツ)_/¯.

server/mediaserver.go
yondonfu: @jailuthra I believe the cause of the "Session ended" errors described in your [comment](https://github.com/livepeer/go-livepeer/pull/1672#issuecomment-739001491) were caused by calling `cxn.pl.Cleanup()` and not by calling `cxn.sessManager.cleanup()` (I'm assuming you were also calling `cxn.pl.Cleanup()` in a previous commit).

When two concurrent segments are received, the second call to `createRTMPStreamIDHandler()` will call [os.NewSession](https://github.com/livepeer/go-livepeer/blob/master/server/mediaserver.go#L276) with the same webhook manifestID as the first call to `createRTMPStreamIDHandler()`. In this case, `os.NewSession()` will return an already [cached storage session](https://github.com/livepeer/go-livepeer/blob/master/drivers/local.go#L44) since the first call to `createRTMPStreamIDHandler()` had created that storage session for the webhook manifestID. As a result when `cxn.pl.Cleanup()` is called in the code block that is executed if an RTMP conn already exists, the storage session for both the first and second segment will be [ended](https://github.com/livepeer/go-livepeer/blob/7138591ec3ee61d49afe1f4fa3720539f30c72ae/core/playlistmanager.go#L234) since the playlist for both RTMP conns reference the same storage session. Then, the next time that storage session is used to save data i.e. during the segment workflow, we get the "Session ended" error.

An easy way to avoid this error is to not call `cxn.pl.Cleanup()` in that RTMP conn already exists code block which is safe to do right now since the cleanup function is not responsible for anything else besides ending the storage session. But, if that cleanup function ever included anything besides ending the storage session then we'd after to reconsider omitting the `cxn.pl.Cleanup()` call...I suppose that might be preferable to increasing the time under lock (as is done in the latest commit) at least until we can figure out additional improvements to the code flow here...

server/mediaserver.go
```diff
@@ -290,8 +290,10 @@ func createRTMPStreamIDHandler(s *LivepeerServer) func(url *url.URL) (strmID str
 			return nil
 		}
 		if _, exists := s.rtmpConnections[mid]; exists {
-			glog.Errorf("Manifest manifestID=%v already exists for streamID url=%s", mid, url.String())
-			return nil
+			if extmid != mid {
```
yondonfu, line 294: So, this is saying if we already have an RTMP conn for the webhook manifestID AND the streamID is different from it then we hit an error condition. Otherwise, if the streamID is the same as the webhook manifestID we do not hit an error condition. This avoids the error condition in the concurrent push test, but I think in practice we could still hit this error condition for concurrent pushes with streamIDs that are different from the webhook manifestID (as would be the case when the playbackId is returned by the webhook).

yondonfu (commented):
> But my unit test does fail when ran with -race due to contention of both segments at InsertHLSSegment()
WDYT my next step should be here?

Hmmm looks like the [MediaPlaylist.Count()](https://github.com/livepeer/go-livepeer/blob/7138591ec3ee61d49afe1f4fa3720539f30c72ae/core/playlistmanager.go#L312) method from the [livepeer/m3u8](https://github.com/livepeer/m3u8) package is not thread safe...so there would be a race when [MediaPlaylist.InsertSegment()](https://github.com/livepeer/go-livepeer/blob/7138591ec3ee61d49afe1f4fa3720539f30c72ae/core/playlistmanager.go#L316) is concurrently called.

I suspect this has always been an issue and we just never triggered it until you wrote your concurrent push test. I think it'd be addressed by changing the mutex in MediaPlyalist to an RWMutex and making sure to call RLock/RUnlock in Count()? While this issue might've existed prior to this PR, if the Push tests with -race enabled fail otherwise then I think we should look into fixing the race.

yondonfu:
I also suspect that these race conditions are only popping up now because RTMP conns were previously *not* ever used concurrently. The RTMP conn code was adapted for the HTTP push use case and in the RTMP live setting there was typically a 1 RTMP conn - 1 stream mapping with segments always coming in one after the other and concurrent segments (and concurrent use of a single RTMP conn) was not a thing in that scenario. I wonder if another approach could be to somehow serialize incoming push requests in the push handler (rough idea - a goroutine for each RTMP conn that receives push segments on a channel so that processSegment() is called for each segment for a RTMP conn serially) such that we can avoid concurrent use of a single RTMP conn as I'm not sure what other issues may arise from that...will think on it a bit more.

server/mediaserver.go
jailuthra: Good catch! I completely missed the case with webhook.

I've added an extra conditional check here https://github.com/livepeer/go-livepeer/pull/1672/commits/5326dbc2564cc73e8eafb45346238ec564b68818 which looks for the `extmid` in the keys of `s.internalManifests` - but it might still be the case that if segments are too close together the older segment hasn't yet updated that mapping - while the newer segment is trying to look for it here.

But imo we can proceed for now and fix it later - as this logic might change again once we improve our extmid/mid split in general in #1673 ?

(Haven't added a test case for concurrent pushes for the webhook-case because it will sometimes fail as noted here)

yondonfu: I believe it would still be possible to trigger the error condition here (for an existing RTMP conn) when `extmid == mid`. I don't have a better suggestion at the moment thought, so I'm fine with this code as-is as long as we come back and address it more thoroughly for #1673.

yondonfu (commented):
Looks good, but there are some CI failures that need to be fixed:

https://app.circleci.com/pipelines/github/livepeer/go-livepeer/1419/workflows/2fd68e2c-8c85-4c56-870b-186aadaf5b54/jobs/7045

```
GO111MODULE=on CGO_LDFLAGS="" go build -tags "mainnet" -ldflags="-X github.com/livepeer/go-livepeer/core.LivepeerVersion=0.5.12-5326dbc2" cmd/livepeer/*.go
go: updates to go.sum needed, disabled by -mod=readonly
Makefile:29: recipe for target 'livepeer' failed
make: *** [livepeer] Error 1
The command '/bin/sh -c make livepeer livepeer_cli' returned a non-zero code: 2
```

server/mediaserver.go
jailuthra: @yondonfu Oh I missed this comment in my notifications while sending the update. Thanks! will fix in a new update rn.

jailuthra:
> Looks good, but there are some CI failures that need to be fixed:

Ah that must be because I was manually cleaning up older `m3u8` line from `go.sum`. It's weird and I don't understand how go.sum works, but somehow it won't let me cleanup old version, despite it using the newer commit from m3u8 package as specified in `go.mod`. Have fixed it now (will also rebase with master and check again once approved, as master has had a `go mod tidy`)

Have also fixed an earlier comment about not cleaning up playlist and keeping the locking minimal.

darkdarkdragon:
@jailuthra 
> despite it using the newer commit from m3u8 package as specified in go.mod

I suspect it is because `lmps` uses m3u8 too and it uses older version

jailuthra:
> I suspect it is because lmps uses m3u8 too and it uses older version

@darkdarkdragon Ah makes sense. Will update it there separate to this PR as well then!

darkdarkdragon:
@jailuthra I think it would be good to also push new tag to the m3u8 repo (`v0.11.1`). This will make links in the go.mod to be more clear 😄

yondonfu (approved):
LGTM after squashing and resolving the merge conflicts.

server/mediaserver.go
```diff
@@ -291,8 +291,14 @@ func createRTMPStreamIDHandler(s *LivepeerServer) func(url *url.URL) (strmID str
...
+				// only error out if we're seeing segments from different extmids but same mid trying to use the same connection
+				// FIXME s.internalManifests might not be updated if segments are too close together
+				if _, extmidSeenBefore := s.internalManifests[extmid]; !extmidSeenBefore {
+					glog.Errorf("RTMP connection for manifest manifestID=%v already exists (streamID url=%s)", mid, url.String())
+					return nil
```
darkdarkdragon, line 299: I think instead of returning error here we should add `ExtrenalManifestID` field to the `core.StreamParameters` struct, and based on that info push handler will be able to remove old session with same playbackID and replace it with the new one

PR #1674, opened 2020-11-30
User session

darkdarkdragon:
Join couple of transcoding sessions to
one 'user' session in recordings

- add 'previousSessions' field to the AuthWebhook response
- in /recordings handler use 'previousSessions' field's content to find all the transcoding sessions belonging to this one
  'user' session and join them to one manifest

Manually 

**Checklist:**

### DIFF
--- a/server/mediaserver.go
+++ b/server/mediaserver.go
@@ -963,26 +963,31 @@ func (s *LivepeerServer) HandlePush(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+// getPlalistsFromStore finds all the json playlist files belonging to the provided manifests
+// returns:
+// - a map of manifestID -> a list of indices pointing to JSON files in the returned list of JSON files
+// - a list of JSON files for all manifestIDs provided
+// - the latest playlist time
 func getPlalistsFromStore(ctx context.Context, sess drivers.OSSession, manifests []string) (map[string][]int, []string, time.Time, error) {
 	var latestPlaylistTime time.Time
 	var jsonFiles []string
 	filesMap := make(map[string][]int)
 	for _, manifestID := range manifests {
 		filesMap[manifestID] = nil
-		now2 := time.Now()
+		start := time.Now()
 		filesPage, err := sess.ListFiles(ctx, manifestID+"/", "/")
 		if err != nil {
 			return nil, nil, latestPlaylistTime, err
 		}
-		glog.V(common.VERBOSE).Infof("Listing directories for manifestID=%s took=%s", manifestID, time.Since(now2))
+		glog.V(common.VERBOSE).Infof("Listing directories for manifestID=%s took=%s", manifestID, time.Since(start))
 		dirs := filesPage.Directories()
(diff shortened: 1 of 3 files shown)


server/mediaserver.go
```diff
@@ -962,6 +963,51 @@ func (s *LivepeerServer) HandlePush(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+func getPlalistsFromStore(ctx context.Context, sess drivers.OSSession, manifests []string) (map[string][]int, []string, time.Time, error) {
```
yondonfu: getPlalistsFromStore -> getPlaylistsFromStore

server/mediaserver.go
```diff
@@ -962,6 +963,51 @@ func (s *LivepeerServer) HandlePush(w http.ResponseWriter, r *http.Request) {
...
+	var latestPlaylistTime time.Time
+	var jsonFiles []string
+	filesMap := make(map[string][]int)
+	for _, manifestID := range manifests {
+		filesMap[manifestID] = nil
+		now2 := time.Now()
```
yondonfu: `now` or `start` instead? Unless there is a specific reason why the 2 is needed?

server/mediaserver.go
```diff
@@ -1057,60 +1107,29 @@ func (s *LivepeerServer) HandleRecordings(w http.ResponseWriter, r *http.Request
...
+	glog.V(common.VERBOSE).Infof("request url=%s found json files: %+v", r.URL, jsonFiles)
+
 	if len(jsonFiles) == 0 {
 		w.WriteHeader(http.StatusNotFound)
 		return
 	}
 	if time.Since(latestPlaylistTime) > 24*time.Hour && !finalizeSet {
 		finalize = true
 	}
+	// finalize = false // REMOVE
```
yondonfu: Can remove?

server/mediaserver.go
yondonfu: I think a short comment describing what this function is supposed to return would be helpful.

Personally, took me a bit to understand that it returns:

- A map of manifestID -> a list of indices pointing to JSON files in the returned list of JSON files
- A list of JSON files for all manifestIDs provided
- The latest playlist time

yondonfu:
Would be nice to have a unit test for checking the session merging functionality.

server/mediaserver.go

server/mediaserver.go

server/mediaserver.go
darkdarkdragon: Fixed

darkdarkdragon: Fixed

darkdarkdragon:
@yondonfu added unit test

yondonfu (approved):
Thanks  for adding the unit test. Would also be nice to have that new behavior in the HandleRecordings handler to be unit tested, but the handler is pretty hefty and creating the necessary stubs may be tricky so if you feel like manual testing did the trick then I think that's ok.

LGTM

PR #1676, opened 2020-12-01
cmd: Add transcoder benchmarking tool

### LINKED ISSUE #1664: Add lpms-bench as buildable binary
It would be useful to have an easy way to test/benchmark segmented transcoding on a machine without having to run an actual node and sending a stream into it. We can add something like [this script](https://gist.github.com/j0sh/640cadfe39b759c38f953baca11e78b0#file-lpms-bench-go) as a buildable binary in this repo so that anyone can clone this repo and build `lpms-bench`. 

Alternatives:
- A user can test/benchmark transcoding using ffmpeg. The benefits of using `lpms-bench` instead are 1) tests the transcoding code that will actually be used by a node 2) tests the *segmented* transcoding workflow.
- A user can clone the script linked above from the gist. The benefits of including `lpms-bench` in this repo are that it can be more easily discovered by users and also maintained as the LPMS development continus. There may be concerns about having to maintain `lpms-bench` as a tool, but I think it would be worthwhile to 1) use a standard benchmarking tool for our transcoding software that is visibile 2) offer a standard benchmarking tool for all transcoder operators.

jailuthra:

Adapt an existing Transcoder benchmarking [script](https://gist.github.com/j0sh/640cadfe39b759c38f953baca11e78b0#file-lpms-bench-go) into a proper go-livepeer tool.

- Use `flag` library like `livepeer` command
- Specify transcoding profile optionally through a JSON file
- Remove the hard requirement of passing in the # of segments (by default read all segments in .m3u8)
- Optionally save the transcoded output segments

WIP/TODO:
- [ ] Add support for live mode
- [x] Calculate total transcode time (sum for all segs, average across concurrent threads)
- [x] Documentation for tool usage similar to https://github.com/livepeer/docs/pull/64/
- [x] Upload a sample m3u8 to a bucket

Manually tested the features on CPU.
Will confirm on GPU as well.

Fixes #1664 

**Checklist:**


cmd/livepeer_bench/livepeer_bench.go
```diff
@@ -0,0 +1,216 @@
...
+	// Override the default flag set since there are dependencies that
+	// incorrectly add their own flags (specifically, due to the 'testing'
+	// package being linked)
+	flag.Set("logtostderr", "true")
+	flag.CommandLine = flag.NewFlagSet(os.Args[0], flag.ExitOnError)
+
+	fname := flag.String("fname", "", "Input m3u8 manifest file")
+	conc := flag.Int("conc", 1, "# of concurrent transcode sessions")
+	segs := flag.Int("segs", 0, "Maximum # of segments to transcode (default all)")
+	profs := flag.String("profs", "P240p30fps16x9,P360p30fps16x9,P720p30fps16x9", "Transcoding options for broadcast job, or path to json config")
```
yondonfu: `transcodingOptions` to match the naming convention used in [livepeer](https://github.com/livepeer/go-livepeer/blob/7f9534c365123f455c7fd3bfb9ebcbf5a99cdd43/cmd/livepeer/livepeer.go#L124)?

cmd/livepeer_bench/livepeer_bench.go
```diff
@@ -0,0 +1,216 @@
...
+	nvidia := flag.String("nvidia", "", "Comma-separated list of Nvidia GPU device IDs to use for transcoding")
+	outPrefix := flag.String("outPrefix", "", "Output segments' prefix (no segments are generated by default)")
+
+	flag.Parse()
+
+	if *fname == "" {
+		panic("Please provide the input manifest as `-fname <input.m3u8>`. See -h or -help for more.")
```
yondonfu: Use `glog.Fatal()` or `glog.Fatalf()` (see [docs](https://godoc.org/github.com/golang/glog#Fatalf)) to match the usage in livepeer.go? Generally, I think it is preferable to use Fatal/Fatalf to handle user input errors and to only use panic for oh-no-something-is-really-wrong errors where the stack trace might be necessary for debugging. This isn't a problem in this file right now, but one downside of using Fatal/Fatalf is that defer statements won't be executed.

cmd/livepeer_bench/livepeer_bench.go
```diff
@@ -0,0 +1,216 @@
...
+			wg.Done()
+		}(i, &wg)
+		time.Sleep(300 * time.Millisecond)
+	}
+	wg.Wait()
+	fmt.Fprintf(os.Stderr, "Took %v to transcode %v segments\n",
+		time.Now().Sub(start).Seconds(), segCount)
+}
+
+func parseVideoProfiles(inp string) []ffmpeg.VideoProfile {
```
yondonfu, line 158: At some point would be nice to use shared code here and in the [video profile parsing code](https://github.com/livepeer/go-livepeer/blob/7f9534c365123f455c7fd3bfb9ebcbf5a99cdd43/server/mediaserver.go#L133) of livepeer.

Think some duplication for now is fine though.

cmd/livepeer_bench/livepeer_bench.go
```diff
@@ -0,0 +1,216 @@
...
+		accel = ffmpeg.Nvidia
+		devices = strings.Split(*nvidia, ",")
+	}
+
+	ffmpeg.InitFFmpeg()
+	segCount := 0
+	var wg sync.WaitGroup
+	dir := path.Dir(*fname)
+	start := time.Now()
+	fmt.Fprintf(os.Stderr, "Program %s Source %s Concurrency %d Profiles %s\n", os.Args[0], *fname, *conc, *profs)
```
kyriediculous: WDYT of printing this to `os.Stdout` instead of `os.Stderr` as it seems informative rather than an error ? 

Alternatively what do you think of displaying this in a table format ? For example

```go
	table := tablewriter.NewWriter(os.Stdout)
	data := [][]string{
		{"Program", os.Args[0},
		{"Source File", *fname},
		{"Concurrent Streams", fmt.Sprintf("%v", *conc)},
		{"Transcoding Options", *transcodingOptions)},
	}

	for _, v := range data {
		table.Append(v)
	}

	table.SetAlignment(tablewriter.ALIGN_RIGHT)
	table.SetCenterSeparator("*")
	table.SetRowLine(true)
	table.SetColumnSeparator("|")
	table.Render()
```

See examples for table formatting within the livepeer CLI here: https://github.com/livepeer/go-livepeer/blob/7f9534c365123f455c7fd3bfb9ebcbf5a99cdd43/cmd/livepeer_cli/wizard_stats.go#L48

cmd/livepeer_bench/livepeer_bench.go
```diff
@@ -0,0 +1,216 @@
...
+				}
+				segCount++
+			}
+			tc.StopTranscoder()
+			wg.Done()
+		}(i, &wg)
+		time.Sleep(300 * time.Millisecond)
+	}
+	wg.Wait()
+	fmt.Fprintf(os.Stderr, "Took %v to transcode %v segments\n",
```
kyriediculous: Would it be better to print to `os.Stdout`? 

We can also do `time.Since(start).Seconds()` it's a little more expressive, although this is fine as well. 

If we are printing to `stdout` we could do `Printf` instead of `FPrintf` as well

cmd/livepeer_bench/livepeer_bench.go
```diff
@@ -0,0 +1,216 @@
...
+		devices = strings.Split(*nvidia, ",")
+	}
+
+	ffmpeg.InitFFmpeg()
+	segCount := 0
+	var wg sync.WaitGroup
+	dir := path.Dir(*fname)
+	start := time.Now()
+	fmt.Fprintf(os.Stderr, "Program %s Source %s Concurrency %d Profiles %s\n", os.Args[0], *fname, *conc, *profs)
+	fmt.Println("time,stream,segment,length")
```
kyriediculous: What do you think of using a table format here as well ? I think we could also omit displaying Unix time or is it a requirement for forensics? 

I also added the transcode error here as per the comment below this one but feel free to omit it.

We can initiate the table on this line 

```go

type struct resultsTableData {
 sync.Mutex
 data [][]string
}
resultsTable := tablewriter.NewWriter(os.Stdout)
resultsTable.SetHeader([]string{"Stream No.", "Segment No.", "Latency", "Error"})
resultsData := &resultsTableData{
 data: [][]string{}{}
}
```

Then on L123 we can append to `data` rather than printing out directly 
```go
resultsData.Lock()
resultsData.data = append(resultsData.data, []string{
  fmt.Sprintf("%v", k),
  fmt.Sprintf("%v", j),
  fmt.Sprintf("%v s", time.Since(t).Seconds()),
  err.Error(), 
})
resultsData.Unlock()
```

When the waitgroup finishes, before appending `data` to `resultsTable` we can sort `data` based on 1. Stream No. and 2. Segment No. to get all in order. 

After that we can append the data to the table and render it

```go
resultsData.Lock()
for _, v := range resultsData.data {
    table.Append(v)
}
resultsData.Unlock()
table.SetAlignment(tablewriter.ALIGN_RIGHT)
table.SetCenterSeparator("*")
table.SetRowLine(true)
table.SetColumnSeparator("|")
table.Render()
```

cmd/livepeer_bench/livepeer_bench.go
```diff
@@ -0,0 +1,216 @@
...
+					}
+					return opts
+				}
+				out := profs2opts(profiles)
+				t := time.Now()
+				_, err := tc.Transcode(in, out)
+				end := time.Now()
+				fmt.Printf("%s,%d,%d,%0.2v\n", end.Format("2006-01-02 15:04:05.999999999"), k, j, end.Sub(t).Seconds())
+				if err != nil {
+					panic(err)
```
kyriediculous: I don't think panicing within this goroutine is desirable, this will stop the execution of the entire program but doesn't clean up running goroutines, neither will we continue with additional segments. 

Should we opt to log the error (and possibly add this to the table per segment mentioned in my earlier comment) ? 

```go
if err != nil {
data = append(data, []string{
  fmt.Sprintf("%v", k),
  fmt.Sprintf("%v", j),
  fmt.Sprintf("%v s", time.Since(t).Seconds()),
  err.Error(), 
})
 glog.Error(err)
 continue
}
data = append(data, []string{
  fmt.Sprintf("%v", k),
  fmt.Sprintf("%v", j),
  fmt.Sprintf("%v s", time.Since(t).Seconds()),
  "n/a", 
})

```

kyriediculous (changes requested):

cmd/livepeer_bench/livepeer_bench.go
jailuthra: Yeah I thought about moving the json->ffmpeg.VideoProfile parsing stuff to the `common` package. But that would involve a lot of renaming/refactoring changes that I wanted to avoid with this PR.

Can do it as a separate PR later tho :+1:

cmd/livepeer_bench/livepeer_bench.go
jailuthra: > WDYT of printing this to os.Stdout instead of os.Stderr as it seems informative rather than an error ?

I liked how the tool printed stats to `stdout` and rest of the information on `stderr` - this allows the user to do a quick export of the stats CSV `./livepeer_bench ... > stats.csv` for more processing - while still seeing the other information on their terminal output.

Ofc this can be handled better by taking in the CSV filename as an optional argument - but for now this seemed good enough and unix-like. WDYT?

> Alternatively what do you think of displaying this in a table format ? For example

Yeah that looks really cool, I'll try it and send an update.

jailuthra: s/json/csv

cmd/livepeer_bench/livepeer_bench.go
jailuthra: Yeah this would be neat, and separate out diff streams, will do.

I'd still prefer keeping the data printing in a csv/tsv like format though - as one of the requirements in the spec #1664 was to let the user export data for more processing with other tools. And almost everyone could export a tsv into excel/sheets but `|` separator might give trouble.

cmd/livepeer_bench/livepeer_bench.go
jailuthra: Ah, still learning the golang quirks. Will change :+1:
And table append sounds good.

cmd/livepeer_bench/livepeer_bench.go
yondonfu: Agree with @jailuthra here on the value of being able to export the stats in a standard format such as CSV so that the data can be easily imported into other tools. I suggest we stick with the current formatting and consider improvements such as different stats visualization formats outside of this PR.

yondonfu:
Think we'll need to specify the `null` muxer [here](https://github.com/livepeer/go-livepeer/blob/7f9534c365123f455c7fd3bfb9ebcbf5a99cdd43/install_ffmpeg.sh#L119) when installing ffmpeg deps for the scenario where `livepeer_bench` does not produce transcoded outputs.

jailuthra:
> Think we'll need to specify the `null` muxer [here](https://github.com/livepeer/go-livepeer/blob/7f9534c365123f455c7fd3bfb9ebcbf5a99cdd43/install_ffmpeg.sh#L119) when installing ffmpeg deps for the scenario where `livepeer_bench` does not produce transcoded outputs.

Ah good catch! I was running with the lpms version of `install_ffmpeg.sh`
TODO for myself - make a combined script for both :)

yondonfu:
From #1664:

> - The tool can be run via a pre-built binary
> - The tool can be built from source easily because all required dependencies (including GPU related dependencies) can be installed by using scripts and docs included in the repo

While I do think having CI upload a pre-built binary that can be included in go-livepeer releases will be helpful, I think we can leave that as a TODO for next week so we can also prepare a new go-livepeer release that includes the pre-built livepeer_bench binary. 

What do you think about just addressing the second point about building from source from now?

I think the one missing piece for that are instructions to build from source with GPU support [here](https://github.com/livepeer/go-livepeer/blob/master/doc/install.md#option-2-build-from-source). We could add a note with the additional dependencies required for GPU support (also indicating that it is Linux only) which I believe are:
 
```
apt-get install -y software-properties-common
apt-key adv --keyserver keyserver.ubuntu.com --recv 15CF4D18AF4F7421
add-apt-repository "deb [arch=amd64] http://apt.llvm.org/xenial/ llvm-toolchain-xenial-8 main"
apt-get update
apt-get -y install clang-8 clang-tools-8
```

The above is based on the list of dependencies installed in this [Dockerfile](https://github.com/livepeer/go-livepeer/blob/master/docker/Dockerfile.build-linux).

jailuthra:
> We could add a note with the additional dependencies required for GPU support (also indicating that it is Linux only) which I believe are:

@yondonfu sounds good will add.

Also looks like don't need all the extra repository-adding commands (which might only be needed for docker or a datacenter with custom subset of ubuntu apt repositories)

Both `clang-8` and `clang-tools-8` seem to be available in default Ubuntu repositories for end-user of 16.04/18.04/20.04:
https://packages.ubuntu.com/xenial/clang-tools-8
https://packages.ubuntu.com/xenial/clang-8

So will add those two in the deps install command in `install.md`

cmd/livepeer_bench/livepeer_bench.go
jailuthra: Changed to `glog.Fatalf` instead of `panic` which should terminate all goroutines.

Decided to not go the *log-error-per-segment* route for now, as the transcoder session might not be in an OK state after throwing an error at us - so better to quit everything (different sessions are processing the same segments anyway)

cmd/livepeer_bench/livepeer_bench.go
jailuthra: Used `time.Since(start)` but kept it `stderr` for now (reasons mentioned in another comment)

cmd/livepeer_bench/livepeer_bench.go
```diff
@@ -0,0 +1,228 @@
...
+				}
+				segCount++
+			}
+			tc.StopTranscoder()
+			wg.Done()
+		}(i, &wg)
+		time.Sleep(300 * time.Millisecond)
+	}
+	wg.Wait()
+	fmt.Fprintf(os.Stderr, "Took %0.4v seconds to transcode total %v segments (%v concurrent sessions)\n",
```
yondonfu: Think the total duration of the segments is useful information since this line would then tell a user that X seconds of video was transcoded in Y seconds which is helpful context since the transcoding speed will depend on the duration of the inputs.

yondonfu (commented):
Looks good! Left one small comment. I'll also try running this from scratch too.

yondonfu:
> I'll also try running this from scratch too.

Got livepeer_bench running smoothly with a GPU (1070 Ti) and also tested with the a custom JSON file for `-transcodingOptions` woo!

cmd/livepeer_bench/livepeer_bench.go
jailuthra: Done, also now printing the segment count in the stream, rather than sum of all sessions.

yondonfu (approved):
LGTM after squashing

cmd/livepeer_bench/livepeer_bench.go
```diff
@@ -0,0 +1,232 @@
...
+							oname = fmt.Sprintf("%s_%s_%d_%d_%d.ts", *outPrefix, p.Name, n, k, j)
+						} else {
+							oname = "-"
+						}
+						o := ffmpeg.TranscodeOptions{
+							Oname:        oname,
+							Profile:      p,
+							Accel:        accel,
+							AudioEncoder: ffmpeg.ComponentOptions{Name: "drop"},
+							Muxer:        ffmpeg.ComponentOptions{Name: "mpegts"},
```
yondonfu: We don't have have to explicitly set the null muxer here? Does the null muxer get used automatically if the Oname is set to "-"?

jailuthra: Yepp exactly. I set it back to "null" here https://github.com/livepeer/go-livepeer/pull/1676/commits/7837b36a126b3057a65963de7c605adf32ad513c but that lead to `transcoder: Unable to open output` error for me on Nvidia

yondonfu: Cool!

> lead to transcoder: Unable to open output error for me on Nvidia

Weird. I remember encountering something like that before, but IIRC the problem was that I did not enable the null muxer in the install_ffmpeg.sh script.

jailuthra: Also looks like `null` muxer doesn't need to be explicitly added in `./configure` script as well, so I will remove that line from `install_ffmpeg.sh` that I added here https://github.com/livepeer/go-livepeer/pull/1676/commits/fce6bcf8b72e47f507a34bbad3305eb13218dc31

jailuthra: >  I remember encountering something like that before, but IIRC the problem was that I did not enable the null muxer in the install_ffmpeg.sh script.

Oh wait you are right, I forgot to run `install_ffmpeg.sh` on my GPU machine.

I'll revert it back to the status-quo of having "null" for "-" and "mpegts" for "prefix/output.ts"

kyriediculous (approved):
Looks good to me as well for the requested changes

PR #1679, opened 2020-12-03
Fix recordings

darkdarkdragon:
Fixes absolute urls to the segments in the VOD playlists

- when `recordObjectStoreUrl` set for object store correctly generate public urls for sessions specified in `previousSessions` field
 

Added unit test

**Checklist:**

### DIFF
--- a/common/util.go
+++ b/common/util.go
@@ -380,3 +380,10 @@ func RatPriceInfo(priceInfo *net.PriceInfo) (*big.Rat, error) {
 
 	return big.NewRat(priceInfo.PricePerUnit, pixelsPerUnit), nil
 }
+
+func JoinPaths(path1, path2 string) string {
+	if !strings.HasSuffix(path1, "/") {
+		return path1 + "/" + path2
+	}
+	return path1 + path2
+}
--- a/common/util.go
+++ b/common/util.go
@@ -381,9 +381,9 @@ func RatPriceInfo(priceInfo *net.PriceInfo) (*big.Rat, error) {
 	return big.NewRat(priceInfo.PricePerUnit, pixelsPerUnit), nil
 }
 
-func JoinPaths(path1, path2 string) string {
-	if !strings.HasSuffix(path1, "/") {
-		return path1 + "/" + path2
+func JoinURL(url, path string) string {
+	if !strings.HasSuffix(url, "/") {
+		return url + "/" + path
 	}
-	return path1 + path2
+	return url + path
 }
--- a/server/recordings_test.go
+++ b/server/recordings_test.go
@@ -16,6 +16,72 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+func TestRecordingHandler(t *testing.T) {
+	drivers.Testing = true
+	lpmon.NodeID = "testNode"
+	assert := assert.New(t)
+	s := setupServer()
+	defer serverCleanup(s)
+	whts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		out, _ := ioutil.ReadAll(r.Body)
+		var req authWebhookReq
+		err := json.Unmarshal(out, &req)
+		if err != nil {
+			glog.Error("Error parsing URL: ", err)
+			w.WriteHeader(http.StatusForbidden)
+			return
+		}
+		w.Write([]byte(`{"manifestID":"playback01", "recordObjectStore": "memory://recstore5",
+		"recordObjectStoreUrl":"https://pub.test/",
+		"previousSessions":["sess1","sess2"]}`))
+	}))
+	defer whts.Close()
+	AuthWebhookURL = whts.URL
+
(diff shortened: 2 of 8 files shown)


server/recordings_test.go
```diff
@@ -16,6 +16,72 @@ import (
...
+			glog.Error("Error parsing URL: ", err)
+			w.WriteHeader(http.StatusForbidden)
+			return
+		}
+		w.Write([]byte(`{"manifestID":"playback01", "recordObjectStore": "memory://recstore5",
+		"recordObjectStoreUrl":"https://pub.test/",
+		"previousSessions":["sess1","sess2"]}`))
+	}))
+	defer whts.Close()
+	AuthWebhookURL = whts.URL
```
yondonfu, line 41: Let's make sure to reset this package level var when this test completes to ensure that other tests that depend on `AuthWebhookURL` start with a clean state.

i.e.

```golang
oldAuthWebhookURL = AuthWebhookURL
AuthWebhookURL = whts.URL
defer func() { AuthWebhookURL = oldAuthWebhookURL }()
```

server/recordings_test.go
```diff
@@ -16,6 +16,72 @@ import (
...
+	jpl = core.NewJSONPlaylist()
+	jpl.InsertHLSSegment(&profile, 1, "sess3/testNode/P144p25fps16x9/3.ts", 2100)
+	bjpl, _ = json.Marshal(jpl)
+	msess3.SaveData("testNode/playlist_3.json", bjpl, nil)
+
+	resp = makeReq("GET", "/recordings/sess3/P144p25fps16x9.m3u8")
+	body, _ := ioutil.ReadAll(resp.Body)
+	resp.Body.Close()
+	assert.Equal(200, resp.StatusCode)
+	assert.Equal("#EXTM3U\n#EXT-X-VERSION:3\n#EXT-X-MEDIA-SEQUENCE:0\n#EXT-X-TARGETDURATION:2100\n#EXTINF:2100.000,\nhttps:/pub.test/sess1/testNode/P144p25fps16x9/1.ts\n#EXT-X-DISCONTINUITY\n#EXTINF:2100.000,\nhttps:/pub.test/sess2/testNode/P144p25fps16x9/2.ts\n#EXT-X-DISCONTINUITY\n#EXTINF:2100.000,\nhttps:/pub.test/sess3/testNode/P144p25fps16x9/3.ts\n#EXT-X-ENDLIST\n", string(body))
```
yondonfu: I see that this assertion passes, but shouldn't the URLs be be prefixed with `https://pub.test` as opposed to `https:/pub.test`?

yondonfu (changes requested):
Is my understanding here correct?

If a user passes any of the manifestIDs associated with a previous session for a `/recordings` request then the response will always be the full playlist for all previous sessions regardless of the manifestID provided?

server/recordings_test.go
darkdarkdragon: Nice catch. I'm again made mistake by using 'path.Join' on URLs 😞 . Fixed

server/recordings_test.go
darkdarkdragon: Fixed

darkdarkdragon:
@yondonfu 
> If a user passes any of the manifestIDs associated with a previous session for a /recordings request then the response will always be the full playlist for all previous sessions regardless of the manifestID provided?

No, only sessions previous for provided one (if user requests 'sess1' then VOD will contain 'sess1' and 'sess2').

common/util.go
```diff
@@ -380,3 +380,10 @@ func RatPriceInfo(priceInfo *net.PriceInfo) (*big.Rat, error) {
 
 	return big.NewRat(priceInfo.PricePerUnit, pixelsPerUnit), nil
 }
+
+func JoinPaths(path1, path2 string) string {
```
yondonfu: Would `JoinURL(url, path string)` be a more descriptive name if this function is only intended to be used as an alternative to `path.Join()` when the first string param is a URL?

yondonfu (commented):
> No, only sessions previous for provided one (if user requests 'sess1' then VOD will contain 'sess1' and 'sess2').

Ah ok I confused myself - the full playlist returned will combine the playlists for all previous sessions (specified in the webook) and the session requested by the user. So, in the test added with this PR the full playlist combines the playlists for sess1, sess2 and sess3 because sess1 and sess2 are specified as previous sessions and sess3 is requested by the user. And for the example in your comment, the full playlist combines the playlists for sess1 and sess2 because they are specified as previous sessions (with sess1 also requested by the user).

common/util.go
darkdarkdragon: Fixed

darkdarkdragon:
@yondonfu 
Yes. If user requests 'sess3' then webhook will contain ```"previousSessions":["sess1","sess2"]``` and VOD will contain all three sessions.
If user requests 'sess2' then `previousSessions` will contain only 'sess1' and VOD will contain only two sessions.

yondonfu (approved):
LGTM after squashing

Issue #1671, opened 2020-11-29
Transactions can exceed manually set gas price


kyriediculous:
**Is your feature request related to a problem? Please describe.**
Users are allowed to specify a gas price of their choosing to have the Livepeer node pay transaction fees on the Ethereum network. There are a couple of UX issues with the current flow of enabling this flag that might lead to unexpected or less than optimal behaviour. 

1. With reward calls the Livepeer node will automatically bump up the gas price and send a replacement transaction if the initial call's gas price is too low for the transaction to confirm within 10 minutes. https://github.com/livepeer/go-livepeer/blob/e69e95cb11a9d9a334c2d12402e570f91b631e21/cmd/livepeer/livepeer.go#L48

    Currently the gas price for the replacement transaction is _always_ determined by the gas price oracle even if a manual gas price is set. This is because the argument is nil here: https://github.com/livepeer/go-livepeer/blob/e69e95cb11a9d9a334c2d12402e570f91b631e21/eth/eventservices/rewardservice.go#L110
 
    If the suggested gas price by the gas price oracle is higher the manual set gas price will be exceeded

    If it is lower the replacement transaction will error out https://github.com/livepeer/go-livepeer/blob/7f9534c365123f455c7fd3bfb9ebcbf5a99cdd43/eth/client.go#L791 

2. An orchestrator when sending `TicketParams` for the broadcaster node to create payment tickets with, will increase its pricing to accommodate for the transaction costs of a ticket redemption. The gas price oracle will always be used with the actual gas price to do this. If the manual set gas price however is much higher then the actual ticket redemption will be sent using the manual set gas price and this could greatly affect profitability of winning tickets.

**Describe the solution you'd like**
1. Transaction gas price should never exceed the one set through the `-gasPrice` flag, not even with a replacement transaction

2. If the actual gas price is lower than the manually set one, use the actual one instead and treat the flag as a `-maxGasPrice` instead. 

From @yondonfu  in #1659 : 

A few thoughts on how the current behavior could be improved upon:

- Allow the user to specify a maximum gas price to use for ETH transactions
The gas price for all transactions even if they are replacement transactions will never exceed the maximum gas price set by a user
- The -gasPrice flag could be aliased to a new -maxGasPrice flag
Could consider keeping the existing -gasPrice flag functionality to set the exact gas price used by transactions, but I'm not sure how useful that would be in practice since gas prices fluctuate so in most cases it feels like setting the maximum gas price is more useful
[...]

PR #1696, opened 2020-12-08
eth: rewardservice to use round subscription

### LINKED ISSUE #1584: Remove polling from reward service

The [reward service](https://github.com/livepeer/go-livepeer/blob/master/eth/eventservices/rewardservice.go) polls to check if the node should call reward once every `blockPollingInterval` seconds. This results in a lot of RPC requests per day since each polling interval involves [multiple RPC requests](https://github.com/livepeer/go-livepeer/blob/96ca7f8aa5ff4ec2d91f949f76292f332fd431bc/eth/eventservices/rewardservice.go#L79). This is bad for users that use rate limited ETH RPC providers (i.e. Infura).

Remove polling from the reward service and instead do something along the lines of the following:

- Wait for a new round event
- When a new round event is received, check if the node should call reward
- If the node should call reward, do so
- Otherwise, do nothing until the next new round event

The above flow would only require a few RPC requests when a new round event is emitted instead of requiring RPC requests every `blockPollingInterval` seconds.

kyriediculous:
Remove polling on a set interval (which executes a ton of RPC requests) to only trying reward when a new round event is received from the `TimeWatcher`. 

- Start reward service with time watcher
- use a go routine to start reward service, remove the go-routine inside of `rewardService.Start` and block instead. This follows the convention of most of our other services
- use an interface for the `TimeWatcher` so that we can unit test this service
- Add basic reward service unit tests
- Add some methods to the `eth.MockClient` to facilitate unit tests
- Use a single channel `serviceErr` to catch errors from startup goroutines for several services (currently rewardService and redeemer)

Manually tested using the devtool and added log events 
Unit tests 

Fixes #1584 

**Checklist:**

### DIFF
--- a/eth/eventservices/rewardservice.go
+++ b/eth/eventservices/rewardservice.go
@@ -3,8 +3,10 @@ package eventservices
 import (
 	"context"
 	"fmt"
+	"math/big"
 
 	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/event"
 	"github.com/golang/glog"
 	"github.com/livepeer/go-livepeer/eth"
 	"github.com/livepeer/go-livepeer/eth/watchers"
@@ -19,7 +21,12 @@ type RewardService struct {
 	client       eth.LivepeerEthClient
 	working      bool
 	cancelWorker context.CancelFunc
-	tw           *watchers.TimeWatcher
+	tw           roundsWatcher
+}
+
+type roundsWatcher interface {
+	SubscribeRounds(sink chan<- types.Log) event.Subscription
+	LastInitializedRound() *big.Int
 }
 
 func NewRewardService(client eth.LivepeerEthClient, tw *watchers.TimeWatcher) *RewardService {
@@ -55,7 +62,7 @@ func (s *RewardService) Start(ctx context.Context) error {
 		case <-rounds:
 			err := s.tryReward()
 			if err != nil {
-				glog.Errorf("Error trying to call reward: %v", err)
+				glog.Errorf("Error trying to call reward err=%v", err)
 			}
 		case <-cancelCtx.Done():
 			glog.V(5).Infof("Reward service done")
@@ -80,27 +87,14 @@ func (s *RewardService) IsWorking() bool {
 }
 
 func (s *RewardService) tryReward() error {
-	currentRound, err := s.client.CurrentRound()
-	if err != nil {
-		return err
-	}
-
-	initialized, err := s.client.CurrentRoundInitialized()
-	if err != nil {
-		return err
-	}
+	currentRound := s.tw.LastInitializedRound()
 
 	t, err := s.client.GetTranscoder(s.client.Account().Address)
 	if err != nil {
 		return err
 	}
 
-	active, err := s.client.IsActiveTranscoder()
-	if err != nil {
-		return err
-	}
-
-	if t.LastRewardRound.Cmp(currentRound) == -1 && initialized && active {
+	if t.LastRewardRound.Cmp(currentRound) == -1 && t.Active {
 		tx, err := s.client.Reward()
 		if err != nil {
 			return err
@@ -116,9 +110,12 @@ func (s *RewardService) tryReward() error {
 				if err != nil {
 					return err
 				}
+				if err := s.client.CheckTx(tx); err != nil {
+					return err
+				}
+			} else {
+				return err
 			}
-
-			return err
 		}
 
 		tp, err := s.client.GetTranscoderEarningsPoolForRound(s.client.Account().Address, currentRound)
--- a/eth/eventservices/rewardservice.go
+++ b/eth/eventservices/rewardservice.go
@@ -41,28 +41,27 @@ func (s *RewardService) Start(ctx context.Context) error {
 	sub := s.tw.SubscribeRounds(rounds)
 	defer sub.Unsubscribe()
 
-	go func(ctx context.Context) {
-		for {
-			select {
-			case err := <-sub.Err():
-				if err != nil {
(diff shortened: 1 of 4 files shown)


eth/eventservices/rewardservice_test.go
```diff
@@ -0,0 +1,156 @@
...
+	time.Sleep(1 * time.Second)
+
+	eth.AssertNumberOfCalls(t, "Reward", 2)
+	eth.AssertNumberOfCalls(t, "CheckTx", 2)
+	eth.AssertNumberOfCalls(t, "ReplaceTransaction", 1)
+
+	errorLogsAfter = glog.Stats.Error.Lines()
+	infoLogsAfter = glog.Stats.Info.Lines()
+	assert.Equal(int64(0), errorLogsAfter-errorLogsBefore)
+	assert.Equal(int64(1), infoLogsAfter-infoLogsBefore)
```
kyriediculous, line 153: Should log 2 info statements here 
1. "replacing transaction" 
2. "called reward" 

But only 1 is picked up by `glog`

eth/eventservices/rewardservice.go
```diff
@@ -40,26 +42,31 @@ func (s *RewardService) Start(ctx context.Context) error {
...
+	for {
+		select {
+		case err := <-sub.Err():
+			if err != nil {
+				glog.Errorf("Block subscription error err=%v", err)
+			}
+		case <-rounds:
+			err := s.tryReward()
+			if err != nil {
+				glog.Errorf("Error trying to call reward: %v", err)
```
yondonfu: "Error trying to call reward err=%v" to match the err field convention used elsewhere.

eth/eventservices/rewardservice.go
```diff
@@ -99,44 +106,26 @@ func (s *RewardService) tryReward() error {
...
 		if err != nil {
 			if err == context.DeadlineExceeded {
-				glog.Infof("Reward tx did not confirm within defined time window - will try to replace pending tx next time")
+				glog.Infof("Reward tx did not confirm within defined time window - will try to replace pending tx once")
+				// Previous attempt to call reward() still pending
+				// Replace pending tx by bumping gas price
+				tx, err = s.client.ReplaceTransaction(tx, "reward", nil)
+				if err != nil {
+					return err
+				}
```
yondonfu, line 112: Since there is no call to `CheckTx()` using the tx returned from `ReplaceTransaction()`, `tryReward()` could return on L126 with nil (when `ReplaceTransaction()` returns a valid tx with a nil error). Then, when the next block event is received, if the replacement tx was not mined, a separate reward tx would be submitted with an incremented nonce. The second reward tx would not be mined until the first replacement tx is mined and when it is mined it would fail.

eth/eventservices/rewardservice_test.go
yondonfu: See my other comment about the current behavior when `ReplaceTransaction()` is called - the log statement for "Called reward..." will never be reached because `tryReward()` will return before then.

yondonfu (changes requested):

eth/eventservices/rewardservice.go
kyriediculous: Calling `checkTx()` for a replacement tx and returning an error if it fails in 2a5a59a099c35208140ab7083ab177af82aa29d9 

However we currently call `tryReward()` when receiving a new round event instead of a new block , ideally the next call will be to call reward for the round after any still pending tx. If replacing the tx fails the user will have to speed it up manually if necessary. We can implement retries when we implement a `MaxGasPrice` #1671

eth/eventservices/rewardservice.go
```diff
@@ -40,26 +44,31 @@ func (s *RewardService) Start(ctx context.Context) error {
...
+	for {
+		select {
+		case err := <-sub.Err():
+			if err != nil {
+				glog.Errorf("Block subscription error err=%v", err)
+			}
+		case <-rounds:
+			err := s.tryReward()
+			if err != nil {
+				glog.Errorf("Error trying to call reward=%v", err)
```
yondonfu: "reward=%v" -> "reward err=%v"

eth/eventservices/rewardservice.go
```diff
@@ -78,65 +87,39 @@ func (s *RewardService) IsWorking() bool {
...
 	t, err := s.client.GetTranscoder(s.client.Account().Address)
 	if err != nil {
 		return err
 	}
 
-	active, err := s.client.IsActiveTranscoder()
-	if err != nil {
-		return err
-	}
+	active := t.DeactivationRound.Cmp(currentRound) >= 0
```
yondonfu: If we're already calling `GetTranscoder()` on L92 we can just use `t.Active` to check if the address is active.

eth/eventservices/rewardservice.go
yondonfu: So, the reason for only replacing a reward tx once and not supporting multiple replacements for the scenario where the first replacement does not confirm within a specific time window is that `tryReward()` is now only called when a NewRound event is emitted instead of at a regular polling interval. Got it. I think that's a reasonable change in behavior for now especially because replacement transactions without a gas price cap historically haven't been particularly helpful for users (if anything they've caused more harm in some cases).

yondonfu (changes requested):

eth/eventservices/rewardservice.go
kyriediculous: Yes, pretty much. We could loop but without a gas cap costs could run up rapidly under rare circumstances. I feel like the gas price oracle does a fairly good job of recommending a gas price that under recent conditions should confirm within a couple of blocks.

yondonfu (approved):
LGTM after squashing

PR #1695, opened 2020-12-08
eth: opt-in subgraph to fetch transcoder pool

kyriediculous:
**What does this pull request do? Explain your changes. (required)**
This PR adds a subgraph client that users can enable by using the `-subgraph` flag to point to the [Livepeer Subgraph](https://thegraph.com/explorer/subgraph/livepeer/livepeer). 

This client is added to the `LivepeerEthClient` as an option way to speed up several requests require fetching on-chain data through multiple JSON-RPC calls. One prominent one is fetching the active transcoder set, which is something the Livepeer Subgraph indexes and we can fetch witch a single HTTP request. 

The `LivepeerSubgraph` interface created in this PR currently supports a single method to fetch the `TranscoderPool`.

If the call the the subgraph is unsuccessful the node will failover to JSON-RPC requests to the provided `-ethUrl`. 

Using the subgraph decreases the time to fetch the on-chain transcoder significantly, [benchmarks](https://github.com/livepeer/protocol/issues/388#issuecomment-739614624) show a decrease of roughly 30 seconds to half a second.

**Specific updates (required)**
- Add a startup flag `-subgraph`
 
- Adds a `eth/subgraph` which exposes a single interface 

```go
type LivepeerSubgraph interface {
	GetActiveTranscoders() ([]*lpTypes.Transcoder, error)
} 
``` 

- A `livepeerSubgraph` type that implements the `LivepeerSubgraph` interface 

```go 
type livepeerSubgraph struct {
	http httpClient
	addr string
}
```

- the `LivepeerEthClient.TranscoderPool()` method will try to use the livepeer subgraph if the `-subgraph` flag was provided 

**How did you test each of these updates (required)**
- Unit tests
- Mainnet testing and benchmarking , see [this comment](https://github.com/livepeer/protocol/issues/388#issuecomment-739614624)

**Does this pull request close any open issues?**
Fixes #1583 

**Checklist:**
- [ ] README and other documentation updated
- [x] Node runs in OSX and devenv
- [x] All tests in `./test.sh` pass

PR #1702, opened 2020-12-09
eth: remove polling from round initialiser

### LINKED ISSUE #1585: Remove polling from round initializer

The [round initializer](https://github.com/livepeer/go-livepeer/blob/master/eth/roundinitializer.go) polls to check if the current round should be initialized every `blockPollingInterval` seconds. This results in a lot of RPC requests per day since each polling interval involves [multiple RPC requests](https://github.com/livepeer/go-livepeer/blob/96ca7f8aa5ff4ec2d91f949f76292f332fd431bc/eth/roundinitializer.go#L72). This is bad for users that use rate limited ETH RPC providers (i.e. Infura).

Remove polling from the round initializer and instead do something along the lines of the following:

- Keep track of a `initialized` boolean that represents whether the current round is initialized
- Wait for new round events - if a new round event is received set `initialized = true`
- Wait for new blocks (via a subscription with the block watcher)
- Keep track of a `currentRound` value
- For each new block received, calculate the round
- If the calculated round > `currentRound`, set `initialized = false`
- If `initialized = true`, then check if the node should initialize the round

The above flow should only require RPC requests when the node checks if it should initialize the round which should occur each block after a new round begins (but is not yet initialized) until the round is initialized. This value should be substantially less than the current RPC request count per round (a few RPC requests every `blockPollingInterval` seconds)

kyriediculous:
Remove polling from the round initialiser as well as reduce RPC requests by utilising the `TimeWatcher`. 

The only remaining RPC calls in the `RoundInitializer` are `TranscoderPool`, `InitializeRound` and `CheckTx`. 
The RPC calls to get the transcoder pool can be avoided through #1695 

- Move RoundInitializer to `eventservices` package, allows them to share stubs 
- Add `TimeWatcher.CurrentRoundStartBlock()` 
- `RoundInitializer` keeps track of a `nextRoundStartBlock *big.int`
- `Roundinitializer` uses `TimeWatcher.SubscribeBlocks()` instead of polling 
- Try Initialize if last seen block > `nextRoundStartBlock` 
- Update unit tests

Unit tests

Fixes #1585 

**Checklist:**

### DIFF
--- a/eth/watchers/timewatcher.go
+++ b/eth/watchers/timewatcher.go
@@ -29,6 +29,7 @@ type TimeWatcher struct {
 	mu                       sync.RWMutex
 	lastInitializedRound     *big.Int
 	lastInitializedBlockHash [32]byte
+	currentRoundStartBlock   *big.Int
 	transcoderPoolSize       *big.Int
 	lastSeenBlock            *big.Int
 
@@ -75,11 +76,18 @@ func (tw *TimeWatcher) LastInitializedBlockHash() [32]byte {
 	return tw.lastInitializedBlockHash
 }
 
-func (tw *TimeWatcher) setLastInitializedRound(round *big.Int, hash [32]byte) {
+func (tw *TimeWatcher) CurrentRoundStartBlock() *big.Int {
+	tw.mu.RLock()
+	defer tw.mu.RUnlock()
+	return tw.currentRoundStartBlock
+}
+
+func (tw *TimeWatcher) setLastInitializedRound(round *big.Int, hash [32]byte, startBlk *big.Int) {
 	tw.mu.Lock()
 	defer tw.mu.Unlock()
 	tw.lastInitializedRound = round
 	tw.lastInitializedBlockHash = hash
+	tw.currentRoundStartBlock = startBlk
 }
 
 func (tw *TimeWatcher) GetTranscoderPoolSize() *big.Int {
@@ -116,7 +124,11 @@ func (tw *TimeWatcher) Watch() error {
 	if err != nil {
 		return fmt.Errorf("error fetching initial lastInitializedBlockHash value err=%v", err)
 	}
-	tw.setLastInitializedRound(lr, bh)
+	num, err := tw.lpEth.CurrentRoundStartBlock()
+	if err != nil {
+		return fmt.Errorf("error fetching current round start block %v=", err)
+	}
+	tw.setLastInitializedRound(lr, bh, num)
 
 	if err := tw.fetchAndSetTranscoderPoolSize(); err != nil {
 		return fmt.Errorf("error fetching initial transcoderPoolSize err=%v", err)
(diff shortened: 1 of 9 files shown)


kyriediculous:
Moving the round initializer into eventservices to avoid code duplication is now in a separate (optional) commit: ec4d66e 

Alternative: move reward service into the eth package.

eth/watchers/timewatcher.go
```diff
@@ -75,11 +76,18 @@ func (tw *TimeWatcher) LastInitializedBlockHash() [32]byte {
...
 }
 
-func (tw *TimeWatcher) setLastInitializedRound(round *big.Int, hash [32]byte) {
+func (tw *TimeWatcher) CurrentRoundStartBlock() *big.Int {
+	tw.mu.RLock()
+	defer tw.mu.RUnlock()
+	return tw.currentRoundStartBlock
+}
+
+func (tw *TimeWatcher) setLastInitializedRound(round *big.Int, hash [32]byte, num *big.Int) {
```
yondonfu: Is there a more descriptive name for the `num` param? `startBlk`?

eth/eventservices/roundinitializer.go
```diff
@@ -1,63 +1,92 @@
...
-			ticker.Stop()
-			return
-		case <-ticker.C:
-			if err := r.tryInitialize(); err != nil {
-				glog.Errorf("error trying to initialize round: %v", err)
+			glog.Infof("Stopping round initializer")
+			return nil
+		case err := <-blockSub.Err():
+			if err != nil {
+				glog.Error(err)
```
yondonfu: Annotate this error (and in the next case) too as is done in the reward service?

eth/eventservices/roundinitializer.go
```diff
@@ -69,26 +98,10 @@ func (r *RoundInitializer) Stop() {
 }
 
 func (r *RoundInitializer) tryInitialize() error {
-	initialized, err := r.client.CurrentRoundInitialized()
```
yondonfu: We should keep this check because we don't want to proceed if the round is already initialized (we can't assume that rounds are initialized using the logic in this service).

yondonfu (commented):
> Alternative: move reward service into the eth package.

I'd be inclined to just move the reward service into the base of the `eth` package. IIRC there was a gradual transition away from the eventservices package which is why it only contains a single service at this point and there doesn't seem to be a particular reason for the package's existence.

eth/eventservices/roundinitializer.go
kyriediculous: If the current round is initialized this would fail on-chain so wouldn't the gas estimation for this also fail when trying to send the transaction? Saving an RPC call and a check, the current round will be marked "initialized" when receiving a new round event by setting `nextRoundStartBlock`

kyriediculous:
Moved reward service into eth package instead 043a16b

eth/watchers/timewatcher.go
kyriediculous: fixed in 70f0e9d

eth/eventservices/roundinitializer.go
kyriediculous: fixed but already caught in the rebase

eth/eventservices/roundinitializer.go
yondonfu: Yeah you're right the gas estimation would fail so we should be able to skip that check. We do end up with a gas estimation error logged for the user in this case and not all ETH clients/providers (AFAIK only geth) support returning the revert reason during gas estimation which might be a bit confusing, but don't feel particularly strongly one way or another about this right now.

yondonfu (approved):
LGTM after squashing

PR #1703, opened 2020-12-10
It/errors rename

darkdarkdragon:
Better error reporting

- rename playlist error to DuplicateSegment error to better convey meaning
- correctly report 'Orchestrator Capped' error that is returned using 403 status code

**Checklist:**

### DIFF
--- a/server/segment_rpc.go
+++ b/server/segment_rpc.go
@@ -453,8 +453,12 @@ func SubmitSegment(sess *BroadcastSession, seg *stream.HLSSegment, nonce uint64)
 		errorString := strings.TrimSpace(string(data))
 		glog.Errorf("Error submitting segment nonce=%d manifestID=%s sessionID=%s seqNo=%d code=%d orch=%s err=%v", nonce, params.ManifestID, sess.OrchestratorInfo.AuthToken.SessionId, seg.SeqNo, resp.StatusCode, ti.Transcoder, string(data))
 		if monitor.Enabled {
-			monitor.SegmentUploadFailed(nonce, seg.SeqNo, monitor.SegmentUploadError(resp.Status),
-				fmt.Sprintf("Code: %d Error: %s", resp.StatusCode, errorString), false)
+			if resp.StatusCode == 403 && strings.Contains(errorString, "OrchestratorCapped") {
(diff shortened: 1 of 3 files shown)


server/segment_rpc.go
```diff
@@ -453,8 +453,12 @@ func SubmitSegment(sess *BroadcastSession, seg *stream.HLSSegment, nonce uint64)
 		errorString := strings.TrimSpace(string(data))
 		glog.Errorf("Error submitting segment nonce=%d manifestID=%s sessionID=%s seqNo=%d code=%d orch=%s err=%v", nonce, params.ManifestID, sess.OrchestratorInfo.AuthToken.SessionId, seg.SeqNo, resp.StatusCode, ti.Transcoder, string(data))
 		if monitor.Enabled {
-			monitor.SegmentUploadFailed(nonce, seg.SeqNo, monitor.SegmentUploadError(resp.Status),
-				fmt.Sprintf("Code: %d Error: %s", resp.StatusCode, errorString), false)
+			if resp.StatusCode == 403 && strings.Contains(errorString, "OrchestratorCapped") {
+				monitor.SegmentTranscodeFailed(monitor.SegmentTranscodeErrorOrchestratorCapped, nonce, seg.SeqNo, errors.New(errorString), false)
```
yondonfu: Technically, based on the current B <> O transcoding protocol, if a non-200 status code is returned then the error here happened during the upload phase since O does not start transcoding until after it returns a 200 status code. Given that, should we use `SegmentUploadFailed()` here instead?

darkdarkdragon: Do you think it would be useful to distinguish between `OrchestratorCapped` error that happened before sending body and `OrchestratorCapped` happened after sending body? I don't have any opinion here - fine for me in either way.

yondonfu: Hmm I can't immediately think of a scenario where it would be particularly important to distinguish between the OrchestratorCapped error occurring during the upload vs. transcode phase, but perhaps for the sake of accurate record keeping it might be good to record an error during the upload phase as an upload error as opposed to a transcode error. If this is as easy as just using `SegmentUploadFailed` instead of `SegmentTranscodeFailed` then I'd say go for it.

darkdarkdragon: Fixed in 26c7c1f0310ddc9b06c3877342568734249f42b7

yondonfu (approved):
LGTM after squashing

PR #1704, opened 2020-12-10
CI: Add livepeer_bench to automated builds

### LINKED ISSUE #1693: Add livepeer_bench to CI build process

livepeer_bench is useful for benchmarking transcoding, but right now users need to build it from source. Automatically building the binary in CI and including it in releases would make it easier for users to get started with benchmarking.

Add livepeer_bench to the CI build process.

Required steps:

1. Build livepeer_bench in the base [Dockerfile](https://github.com/livepeer/go-livepeer/blob/master/docker/Dockerfile.build#L24)
2. Copy the livepeer_bench into the [directory](https://github.com/livepeer/go-livepeer/blob/a37b5b710c6d451559e735ef9ba44c4b29bd5fdb/upload_build.sh#L45) that is used to created an archive that is uploaded to a storage bucket

### LINKED ISSUE #1697: ci: Only generate sha256 hash of archive
Currently, CI generates the [sha256 *and* md5 hashes](https://github.com/livepeer/go-livepeer/blob/a37b5b710c6d451559e735ef9ba44c4b29bd5fdb/upload_build.sh#L47) of the livepeer and livepeer_cli binaries for each platform (darwin, linux, windows). After #1693 we will have 3 binaries which requires generating an additional 6 hashes (for a new binary, 2 hashes for each platform). In reality, we don't need to use 2 hashing algorithms and we can probably just generate a hash for the entire archive as opposed to a hash for each individual binary. We should:

- Remove md5 hashing from [upload_build.sh](https://github.com/livepeer/go-livepeer/blob/master/upload_build.sh)
- Only generate the sha256 hash for the archive created from the directory that contains all binaries (i.e. livepeer, livepeer_cli, livepeer_bench)

jailuthra:

- Don't generate MD5 sums for binaries, instead only generate SHA256
- Add `livepeer_bench` binary to automated build process(es)

- [ ] Let the CI process run on this PR and verify builds on discord

Fixes #1693
Fixes #1697

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -21,7 +21,7 @@ jobs:
       run: ./install_ffmpeg.sh
     - name: Build Livepeer
       shell: msys2 {0}
-      run: ./ci_env.sh make livepeer livepeer_cli
+      run: ./ci_env.sh make livepeer livepeer_cli livepeer_bench
     - name: Upload build
       shell: msys2 {0}
       env: 
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,5 +13,5 @@ script:
   - git fetch --unshallow --tags
   - ./install_ffmpeg.sh
   - go mod download
-  - ./ci_env.sh make livepeer livepeer_cli
+  - ./ci_env.sh make livepeer livepeer_cli livepeer_bench
   - ./upload_build.sh
--- a/docker/Dockerfile.build
+++ b/docker/Dockerfile.build
@@ -21,6 +21,6 @@ COPY go.sum go.sum
 RUN go mod download
 
 COPY . .
-RUN make livepeer livepeer_cli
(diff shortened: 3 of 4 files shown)


yondonfu (commented):
Looks great!

Do you think it could make sense to just generate the SHA256 hash for the archive that contains livepeer, livepeer_cli and livepeer_bench? At the moment, we have a hash for each of the binaries, but I'm wondering how useful that is in practice or if it just adds clutter to the release notes.

jailuthra:
> Do you think it could make sense to just generate the SHA256 hash for the archive that contains livepeer, livepeer_cli and livepeer_bench?

Having a single SHA256 definitely sounds cleaner and solves the same purpose. Will change!

yondonfu (approved):
LGTM! :shipit:

PR #1706, opened 2020-12-13
PushAPI: Handle playbackID clash between different streamIDs

### LINKED ISSUE #1673: HTTP500 errors when creating a new session of a stream too fast
**Describe the bug**

The Push API's handling of segments of different "external manifests" (aka sessionID, unique for a particular broadcast session of a stream) but same "internal manifest" (aka manifestID, unique per stream) is not good enough.

For example, when one stream session is over the client needs to wait for **1 minute** before opening another stream session. If they don't and their new stream session lands up at the same B node - they will get a 500 due to "Manifest already exists" error.

**To Reproduce**
1. Start a local B with `-authWebhookUrl`
2. Start an auth endpoint that always returns `{"manifestID":"SameInternalManifestID"}`
3. Do a HTTP push with a session ID, eg `curl -XPOST --data-binary @input.ts -v http://127.0.0.1:8935/live/session1/0.ts -H "accept: multipart/mixed" -i`
4. Before the `session1` is cleaned up (~1minute), do another HTTP push with a different session ID, eg `curl -XPOST --data-binary @input.ts -v http://127.0.0.1:8935/live/session2/0.ts -H "accept: multipart/mixed" -i`
5. Second push should fail with `Manifest manifestID=SameInternalManifestID already exists for streamID url=http://127.0.0.1:8935/live/session2/0.ts`

**Expected behavior**
We can consider reducing the timeout here, or better communicating the error to the client rather than throwing a generic 500.

**Additional context**
This issue was forked from this comment https://github.com/livepeer/go-livepeer/issues/1665#issuecomment-735103092

jailuthra:

Improve handling of new streams that re-use a playbackID for which an old stream session hasn't ended yet.

- [x] Remove HTTP500 error for playbackID clash
- [x] Close the old stream session for same playbackID and open a new one 
- [x] Unit Tests
- ~~[ ] WIP: Re-use HLS playlist via [EXT-X-DISCONTINUITY](https://tools.ietf.org/html/draft-pantos-http-live-streaming-23#section-4.3.2.3) tag~~ *will move it to a separate PR*

- [x] Manual test via the `curl` reproduction steps
- [x] Unit Test before & after

Fixes #1673 

**Checklist:**

### DIFF
--- a/server/mediaserver.go
+++ b/server/mediaserver.go
@@ -271,6 +271,10 @@ func createRTMPStreamIDHandler(s *LivepeerServer) func(url *url.URL) (strmID str
 		if mid == "" {
 			mid = core.RandomManifestID()
 		}
+		// Generate RTMP part of StreamID
+		if key == "" {
+			key = common.RandomIDGenerator(StreamKeyBytes)
+		}
 
 		if os != nil {
 			oss = os.NewSession(string(mid))
@@ -282,28 +286,12 @@ func createRTMPStreamIDHandler(s *LivepeerServer) func(url *url.URL) (strmID str
 		} else if drivers.RecordStorage != nil {
 			ross = drivers.RecordStorage.NewSession(recordPath)
 		}
-
 		// Ensure there's no concurrent StreamID with the same name
 		s.connectionLock.RLock()
 		defer s.connectionLock.RUnlock()
 		if core.MaxSessions > 0 && len(s.rtmpConnections) >= core.MaxSessions {
 			glog.Errorf("Too many connections for streamID url=%s err=%v", url.String(), err)
 			return nil
-		}
-		if _, exists := s.rtmpConnections[mid]; exists {
-			if extmid != mid {
-				// only error out if we're seeing segments from different extmids but same mid trying to use the same connection
-				// FIXME s.internalManifests might not be updated if segments are too close together
-				if _, extmidSeenBefore := s.internalManifests[extmid]; !extmidSeenBefore {
-					glog.Errorf("RTMP connection for manifest manifestID=%v already exists (streamID url=%s)", mid, url.String())
-					return nil
-				}
-			} // by default we will handle re-using the extmid's older connection in HandlePush
-		}
-
-		// Generate RTMP part of StreamID
-		if key == "" {
-			key = common.RandomIDGenerator(StreamKeyBytes)
 		}
 		return &core.StreamParameters{
 			ManifestID: mid,
@@ -398,8 +386,7 @@ func jsonProfileToVideoProfile(resp *authWebhookResponse) ([]ffmpeg.VideoProfile
 	return profiles, nil
 }
 
-func streamParams(rtmpStrm stream.RTMPVideoStream) *core.StreamParameters {
-	d := rtmpStrm.AppData()
+func streamParams(d stream.AppData) *core.StreamParameters {
 	p, ok := d.(*core.StreamParameters)
 	if !ok {
 		glog.Error("Mismatched type for RTMP app data")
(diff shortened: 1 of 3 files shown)


server/mediaserver.go
```diff
@@ -789,6 +804,10 @@ func (s *LivepeerServer) HandlePush(w http.ResponseWriter, r *http.Request) {
 					s.connectionLock.RLock()
 					if cxn, exists := s.rtmpConnections[intmid]; exists {
 						lastUsed = cxn.lastUsed
+					} else {
+						s.connectionLock.RUnlock()
+						glog.Warningf("Watchdog tried closing session for streamID=%s, which was already closed", extmid)
```
darkdarkdragon, line 803: I think also it would be good to add check for 
```go
if intmid, exists := s.internalManifests[extmid]; exists {
```
that way we will prevent two competing watchdog timers running simultaneously.

server/mediaserver.go
```diff
@@ -759,10 +754,30 @@ func (s *LivepeerServer) HandlePush(w http.ResponseWriter, r *http.Request) {
...
+				}
+			}
+			s.connectionLock.RUnlock()
+			if oldStreamID != "" && params.ExternalID != oldStreamID {
+				// Close the old connection, and open a new one
+				// TODO try to re-use old HLS playlist?
+				glog.Warningf("Ending streamID=%v as new streamID=%s with same manifestID=%s has arrived",
+					oldStreamID, params.ExternalID, params.ManifestID)
+				removeRTMPStream(s, oldStreamID)
+				mid = params.ExternalID
```
darkdarkdragon: At that point `mid` is equal to the `params.ExternalID` anyway. Probably we shouldn't have added `ExternalID` field to the `params` - it's the same value parsed from the URL. We could just check here if 
```go
mid != params.ManifestID && s.rtmpConnections[params.ManifestID] != nil && s.internalManifests[mid] == nil {
```
instead of 
```go
if params.ExternalID != "" {
```

darkdarkdragon (changes requested):
Couple of comments, otherwise looks good.

darkdarkdragon:
@jailuthra 
> Re-use HLS playlist via EXT-X-DISCONTINUITY tag

This will complicate state management - we will need to keep track of the seqNo and rename segments before inserting into the playlist. Honestly I would prefer not to do it now. @yondonfu what do you think?

yondonfu:
@darkdarkdragon 

> Honestly I would prefer not to do it now. @yondonfu what do you think?

I think it could make sense to implement the discontinuity tag in a separate PR as Jai suggested, but would that mean that playback from Bs for streams where there is a streamID switch for the same playbackID would not work until then? I believe the expected behavior after this PR is that there would a duplicate segment error when trying to insert segments into the playlist due to overlapping seqNo values for the two streamIDs?

darkdarkdragon:
>  I believe the expected behavior after this PR is that there would a duplicate segment error when trying to insert segments into the playlist due to overlapping seqNo values for the two streamIDs?

No - it this PR we're completely removing old `rtmpConnection` structure that contains playlists and creating new one - so no  duplicate segment errors.

darkdarkdragon:
> but would that mean that playback from Bs for streams where there is a streamID switch for the same playbackID would not work until then?

It will work, but players will see reset of seqNo and disappearance of the last segments from the old session

yondonfu:
> No - it this PR we're completely removing old rtmpConnection structure that contains playlists and creating new one - so no duplicate segment errors.

Ah got it thanks for the explanation I missed that part.

> It will work, but players will see reset of seqNo and disappearance of the last segments from the old session

Got it. In that case, I agree that it makes sense to implement the discontinuity tag outside of this PR. At the very least, this PR will improve performance for push clients leaving the discontinuity issues with playback on Bs for the separate PR.

darkdarkdragon (approved):
LGTM

server/mediaserver_test.go
```diff
@@ -586,10 +586,6 @@ func TestCreateRTMPStreamHandler(t *testing.T) {
 	if err := handler(u, st); err != nil {
 		t.Error("Handler failed ", err)
 	}
-	// Test collisions via stream reuse
-	if sid := createSid(u); sid != nil {
```
yondonfu, line 590: Instead of removing this test completely should we modify it to check for the absence of a failure when there is a naming collision?

yondonfu (commented):
Looks good! Just one comment about a test modification.

server/mediaserver_test.go
jailuthra: Makes sense, changed it while rebase+autosquash!

yondonfu (approved):
LGTM

PR #1717, opened 2020-12-23
Classify more errors as timeouts

darkdarkdragon:
Check segment upload error string for more variants of timeout error.

- 
- 
- 

Manually

**Checklist:**

### DIFF
--- a/server/segment_rpc.go
+++ b/server/segment_rpc.go
@@ -352,7 +352,7 @@ func SubmitSegment(sess *BroadcastSession, seg *stream.HLSSegment, nonce uint64)
 	segCreds, err := genSegCreds(sess, seg)
 	if err != nil {
 		if monitor.Enabled {
-			monitor.SegmentUploadFailed(nonce, seg.SeqNo, monitor.SegmentUploadErrorGenCreds, err.Error(), false)
+			monitor.SegmentUploadFailed(nonce, seg.SeqNo, monitor.SegmentUploadErrorGenCreds, err, false)
 		}
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


server/segment_rpc.go
```diff
@@ -429,9 +430,10 @@ func SubmitSegment(sess *BroadcastSession, seg *stream.HLSSegment, nonce uint64)
 	resp, err := httpClient.Do(req)
 	uploadDur := time.Since(start)
 	if err != nil {
+		timedout := err.(*url.Error).Timeout()
 		glog.Errorf("Unable to submit segment orch=%v nonce=%d manifestID=%s sessionID=%s seqNo=%d orch=%s err=%v", ti.Transcoder, nonce, params.ManifestID, sess.OrchestratorInfo.AuthToken.SessionId, seg.SeqNo, ti.Transcoder, err)
 		if monitor.Enabled {
-			monitor.SegmentUploadFailed(nonce, seg.SeqNo, monitor.SegmentUploadErrorUnknown, err.Error(), false)
+			monitor.SegmentUploadFailed(nonce, seg.SeqNo, monitor.SegmentUploadErrorUnknown, err.Error(), false, timedout)
```
yondonfu: What error message are we trying to account for with the `err.Timeout()` check that is not accounted by string checking in `monitor.SegmentUploadFailed()`?

If we're just generally worried about timeout related errors that are not covered by string checking, could a cleaner approach be to pass the error as an `error` type to `monitor.SegmentUploadFailed()`? This would avoid adding a boolean to the method signature that at the moment would always be set to false except for this single call. Then, we could add this check within the method to complement the string checking:

```golang
if err, ok := err.(net.Error); ok && err.Timeout() {
    // Check that the error implements the net.Error interface
    // and the Timeout() method just to be safe
}
```

If we're hesitant to change the method signature (by passing an error type instead of an error reason string) we could also add the above check here and set the error code to `monitor.SegmentUploadErrorTimeout`.

darkdarkdragon: > What error message are we trying to account for with the err.Timeout() check that is not accounted by string checking in monitor.SegmentUploadFailed()?

I'm just trying to cover everything I can.

Fixed in 1b5a68c867b580e01f285c133124062fee32352d by passing error object to the `SegmentUploadFailed` function.

yondonfu (approved):
LGTM :shipit: after squashing

PR #1727, opened 2021-01-05
eth: max gas price

### LINKED ISSUE #1671: Transactions can exceed manually set gas price

kyriediculous:
Changes `-gasPrice` flag to `-maxGasPrice` and allows users to set a max gas price to pay for transactions.

- Changes `-gasPrice` flag to be `-maxGasPrice` instead. Removes `-gasLimit` flag (seems superflious). 

- Implements `SuggestGasPrice` that checks for the max gas price when set , `MaxGasPrice` and `SetMaxGasPrice` on `Backend`.

- Changes CLI server API endpoings `/gasPrice` and `/setGasPrice` to `/maxGasPrice` and `/setMaxGasPrice` 

- Since the gasLimit flag is removed and the `gasPrice` flag repurposed `LivepeerEthClient.GasInfo` and `LivepeerEthClient.SetGasInfo` are no longer needed and removed.

- Moved unlocking the Eth account into the `LivepeerEthClient` constructor and removed `LivepeerEthClient.Setup` as it now only unlocks the account after removing `LivepeerEthClient.SetGasInfo`

- Updates the CLI to allow for setting a max gas price 

Fixes #1671 

**Checklist:**


cmd/livepeer/livepeer.go
```diff
@@ -134,8 +134,7 @@ func main() {
 	ethOrchAddr := flag.String("ethOrchAddr", "", "ETH address of an on-chain registered orchestrator")
 	ethUrl := flag.String("ethUrl", "", "Ethereum node JSON-RPC URL")
 	ethController := flag.String("ethController", "", "Protocol smart contract address")
-	gasLimit := flag.Int("gasLimit", 0, "Gas limit for ETH transactions")
-	gasPrice := flag.Int("gasPrice", 0, "Gas price for ETH transactions")
```
yondonfu, line 138: I'm on board with removing the gasLimit flag, but what do you think about keeping the existing gasPrice flag?

Reasons:

- It is already here and I don't think it is harmful to have
- Leaves the possibility for anyone that already is or that wants to use a fixed default gas price
- Leaves another route to control gas price at least until better options are added (i.e. integrating with a better gas price estimation API)

Also was looking at Chainlink's [gas price control flags](https://docs.chain.link/docs/configuration-variables#eth_gas_price_default) for some inspiration as to how this area might be improved down the line.

eth/client.go
```diff
@@ -602,13 +573,7 @@ func (c *client) Vote(pollAddr ethcommon.Address, choiceID *big.Int) (*types.Tra
...
 	}
 
-	gl, gp := c.GetGasInfo()
-	opts, err := c.accountManager.CreateTransactOpts(gl, gp)
-	if err != nil {
-		return nil, err
-	}
-
-	return poll.Vote(opts, choiceID)
+	return poll.Vote(nil, choiceID)
```
yondonfu, line 612: I think this might cause breakage. If the first argument is nil, then the underlying [Transact()](https://github.com/livepeer/go-livepeer/blob/04ef3afbc20797cf102fe55c925c385c74191dea/eth/contracts/poll.go#L226) call will receive a nil `*bind.TransactOpts`. Doesn't look like there are any nil checks in code for the underlying [Transact()](https://github.com/ethereum/go-ethereum/blob/19d7a37abb9f3b9bf1a94baf6bd8c7d5042e54f8/accounts/abi/bind/base.go#L207) call as it expects to receive a non-nil `*bind.TransactOpts`. So, if we pass nil here I think there would be a nil pointer error triggered.

yondonfu (changes requested):

cmd/livepeer/livepeer.go
kyriediculous: I'm wondering if it makes sense to keep it. 

Let's say you want to use a fixed gas price, are there any benefits that could be gained over using a max gas price instead ? 

Assuming with a fixed gas price you don't want to use a gas price higher than what is set, you can achieve the same thing by setting a max gas price instead. 

Then when the gas price is lower than the max gas price you would still have the benefit of using the lower gas price with similar confirmation times.

yondonfu: Sure I also think the maxGasPrice flag is almost always more useful then a gasPrice flag that sets a fixed gas price. I mainly want to go through the exercise of checking if it makes sense to immediately remove a user facing flag if we don't know if anyone uses them.

If it's not harmful to leave these flags for now (now also including -gasLimit in there as well), I think we should consider keeping them and their associated functionality for another release, for now logging a deprecation warning similar to what is done for the old [S3 object store flags](https://github.com/livepeer/go-livepeer/blob/fd07a5673175aa6d10f56dde6eb3ef2a9f5691b9/cmd/livepeer/livepeer.go#L740) (which I like as a good practice in these cases where we want to clean up code, but also give users an early deprecation warning) and then removing them in another release.

eth/client.go
```diff
@@ -162,212 +157,32 @@ func NewClient(accountAddr ethcommon.Address, keystoreDir string, eth *ethclient
...
-		c.gasPrice = gasPrice
-		return nil
-	}
-}
-
-func (c *client) GetGasInfo() (gasLimit uint64, gasPrice *big.Int) {
-	return c.gasLimit, c.gasPrice
-}
-
-func (c *client) setContracts(opts *bind.TransactOpts) error {
```
yondonfu, line 207: Why is this being removed? This will cause breakage because the contract sessions used by the client will all be nil.

eth/client.go
```diff
@@ -602,13 +417,7 @@ func (c *client) Vote(pollAddr ethcommon.Address, choiceID *big.Int) (*types.Tra
...
 	}
 
-	gl, gp := c.GetGasInfo()
-	opts, err := c.accountManager.CreateTransactOpts(gl, gp)
-	if err != nil {
-		return nil, err
-	}
-
-	return poll.Vote(opts, choiceID)
+	return poll.Vote(c.defaultTxOpts, choiceID)
```
yondonfu: What's the benefit of introducing defaultTxOpts instead of the existing code which uses the value returned by CreateTransactOpts in the Vote() method? Seems like that would've worked fine.

yondonfu (changes requested):

cmd/livepeer/livepeer.go
```diff
@@ -133,9 +133,9 @@ func main() {
 	ethKeystorePath := flag.String("ethKeystorePath", "", "Path for the Eth Key")
 	ethOrchAddr := flag.String("ethOrchAddr", "", "ETH address of an on-chain registered orchestrator")
 	ethUrl := flag.String("ethUrl", "", "Ethereum node JSON-RPC URL")
-	ethController := flag.String("ethController", "", "Protocol smart contract address")
-	gasLimit := flag.Int("gasLimit", 0, "Gas limit for ETH transactions")
```
yondonfu, line 137: Let's follow the same deprecation procedure that we're using for `-gasPrice` for `-gasLimit` as well.

cmd/livepeer/livepeer.go
```diff
@@ -378,20 +378,18 @@ func main() {
...
-		if err != nil {
-			glog.Errorf("Failed to create client: %v", err)
-			return
-		}
-
-		var bigGasPrice *big.Int
+		var bigMaxGasPrice *big.Int
 		if *gasPrice > 0 {
-			bigGasPrice = big.NewInt(int64(*gasPrice))
+			glog.Infof("-gasPrice is deprecated, you can use -maxGasPrice instead")
```
yondonfu: Let's use `Warningf()` since this is a deprecation warning.

eth/backend.go
```diff
@@ -40,13 +41,18 @@ type Backend interface {
...
+	SetMaxGasPrice(gp *big.Int)
 }
 
 type backend struct {
 	*ethclient.Client
 	abiMap       map[string]*abi.ABI
 	nonceManager *NonceManager
 	signer       types.Signer
+
+	sync.Mutex
```
yondonfu: Should we use a RWMutex here since the # of reads of the maxGasPrice will likely far exceed the # of writes in practice?

eth/client.go
```diff
@@ -162,46 +156,35 @@ func NewClient(accountAddr ethcommon.Address, keystoreDir string, eth *ethclient
...
-	}, nil
-}
+	if err := am.Unlock(password); err != nil {
+		return nil, err
+	}
 
-func (c *client) Setup(password string, gasLimit uint64, gasPrice *big.Int) error {
-	err := c.accountManager.Unlock(password)
+	// Setting gasLimit and gasPrice to zero, nil allows it to be overwritten by eth.Backend
+	opts, err := am.CreateTransactOpts(0, nil)
```
yondonfu, line 176: Since we're following the deprecation procedure described in the previous round of review, let's keep the existing functionality for -gasPrice and -gasLimit until they are actually removed?

yondonfu (changes requested):
Since we're following the deprecation procedure described in

eth/client.go
kyriediculous: While I think we should be aware that by keeping the gasPrice flag around with existing behaviour it can lead to worse UX for the node however since the new behaviour is different I'm okay with keeping it around.

cmd/livepeer/livepeer.go
kyriediculous: decided to keep around `-gasLimit` and not deprecate it for now even though I'm not sure how useful the flag is.

cmd/devtool/devtool.go
```diff
@@ -159,16 +159,15 @@ func ethSetup(ethAcctAddr, keystoreDir string, isBroadcaster bool) {
...
 	if err != nil {
 		glog.Errorf("Failed to create client: %v", err)
 		return
 	}
 
-	err = client.Setup(passphrase, uint64(0), nil)
-	if err != nil {
-		glog.Fatalf("Failed to setup client: %v", err)
+	if err := client.SetGasInfo(0, nil); err != nil {
+		glog.Errorf("Failed to setup Livepeer Ethereum Client: %v", err)
```
yondonfu: Looks like an old error message is used here. Let's update it to be more descriptive of the error condition.

cmd/livepeer/livepeer.go
```diff
@@ -378,20 +379,25 @@ func main() {
...
 		if *gasPrice > 0 {
+			glog.Warning("-gasPrice is deprecated, you can use -maxGasPrice instead")
 			bigGasPrice = big.NewInt(int64(*gasPrice))
 		}
 
-		err = client.Setup(*ethPassword, uint64(*gasLimit), bigGasPrice)
-		if err != nil {
-			glog.Errorf("Failed to setup client: %v", err)
+		if err := client.SetGasInfo(uint64(*gasLimit), bigGasPrice); err != nil {
+			glog.Errorf("Failed to setup Livepeer Ethereum Client: %v", err)
```
yondonfu: Looks like an old error message is used here. Let's update it to be more descriptive of the error condition.

eth/backend.go
```diff
@@ -99,6 +105,31 @@ func (b *backend) SendTransaction(ctx context.Context, tx *types.Transaction) er
 	return nil
 }
 
+func (b *backend) SuggestGasPrice(ctx context.Context) (*big.Int, error) {
+	gp, err := b.Client.SuggestGasPrice(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	if b.maxGasPrice != nil && gp.Cmp(b.maxGasPrice) >= 0 {
```
yondonfu: Minor: Should the conditional be > if we're maxGasPrice is the maximum allowable value (which is implied by the error message)?

eth/client.go
```diff
@@ -602,8 +593,8 @@ func (c *client) Vote(pollAddr ethcommon.Address, choiceID *big.Int) (*types.Tra
 		return nil, err
 	}
 
-	gl, gp := c.GetGasInfo()
-	opts, err := c.accountManager.CreateTransactOpts(gl, gp)
+	// Setting gasLimit and gasPrice to zero, nil allows it to be overwritten by eth.Backend
```
yondonfu: Why not just leave the existing code as-is since gasLimit will default to 0 and gasPrice will default to nil anyway especially since all the other contract methods will continue to use the stored gasLimit and gasPrice?

yondonfu (changes requested):

cmd/livepeer/livepeer.go
kyriediculous: I felt like this was most descriptive as it's mostly old code repurposed from `client.Setup`. 

It doesn't only set up the gas info but also binds the options to the contract. 

Maybe "Failed to set up contract transactors", but that seemed geth specific lingo.

kyriediculous:
Should have addressed remaining comments, added max gas price to node stats in CLI, improved error logging.

eth/helpers.go
```diff
@@ -245,3 +246,21 @@ func handleEthSlice(val reflect.Value) string {
...
+
+type Denomination *big.Int
+
+var (
+	Wei   Denomination = big.NewInt(params.Wei)
+	GWei               = big.NewInt(params.GWei)
+	Ether              = big.NewInt(params.Ether)
+)
+
+func FromWei(amount *big.Int, to Denomination) *big.Rat {
```
yondonfu: Could `FromBaseAmount()` be used to accomplish the same thing and also handle string formatting if it accepted a maxDecimals param (instead of the hardcoded constant set to 18) to achieve the same thing? Something like:

```golang
func FromWei(amount *big.Int, to int) string {
    maxDecimals := int(math.Log10(float64(to)))
    return FromBaseAmount(amount, maxDecimals)
}

// Usage
FromWei(amount, params.GWei)
FromWei(amount, params.Ether)
```

cmd/livepeer_cli/wizard_stats.go
```diff
@@ -58,6 +60,7 @@ func (w *wizard) stats(showOrchestrator bool) {
 		[]string{"ETH Account", w.getEthAddr()},
 		[]string{"LPT Balance", eth.FormatUnits(lptBal, "LPT")},
 		[]string{"ETH Balance", eth.FormatUnits(ethBal, "ETH")},
+		[]string{"Max Gas Price", fmt.Sprintf("%v GWei", strings.TrimRight(strings.TrimRight(eth.FromWei(maxGasPrice, eth.GWei).FloatString(18), "0"), "."))},
```
yondonfu: Can we avoid the trims by making sure the value returned by whatever function is called here is properly formatted?

eth/helpers.go
kyriediculous: fixed in 1bc1b8b0

cmd/livepeer_cli/wizard_stats.go
kyriediculous: fixed in 1bc1b8b0

yondonfu (approved):
LGTM

kyriediculous:
Rebased and tested on rinkeby, pending green CI

kyriediculous:
Finally got CI to pass , kept running into issues with unrelated tests in the server repo (SubmitSegment).

eth/client.go
```diff
@@ -162,46 +161,35 @@ func NewClient(accountAddr ethcommon.Address, keystoreDir string, eth *ethclient
...
 		return nil, err
 	}
 
-	return &client{
+	if err := am.Unlock(password); err != nil {
+		return nil, err
+	}
+
+	// Setting gasLimit and gasPrice to zero, nil allows it to be overwritten by eth.Backend
+	opts, err := am.CreateTransactOpts(0, nil)
```
yondonfu, line 176: Looks like a change was added here after the squash and before the merge to master. 

- We don't need this if we're calling SetGasInfo
- setContracts is called once in this constructor and then again in SetGasInfo

PR #1731, opened 2021-01-07
server: transcode auth webhook

### LINKED ISSUE #1569: Transcoding Protocol Authentication

kyriediculous:
Allows Orchestrators to specify an `-authWebhookUrl` to authenticate broadcasters upon discovery (`GetOrchestatorInfo`). 

- created an `authenticateBroadcaster()` function in `server/rpc.go` this will no-op if no `-authWebhookUrl` is specified

- Add a call to `authenticateBroadcaster()` when calling `getOrchestrator()` in `server/rpc.go`

- Added unit tests for this new behaviour for `getOrchestrator()` using the `httptest` package which spins up a one-off http server which we specify as  `server.AuthWebhookUrl`

- Updated documentation on auth webhooks in `doc` 

Added unit/integration tests

Fixes #1569 

**Checklist:**

### DIFF
--- a/doc/rtmpwebhookauth.md
+++ b/doc/rtmpwebhookauth.md
@@ -1,6 +1,6 @@
 # Webhook Authentication
 
-Incoming streams can be authenticated using webhooks. To use these webhooks, node operators must implement their own web service / endpoint to be accessed only by the Livepeer node. As new streams appear, the Livepeer node will call this endpoint to determine whether the given stream is allowed.
+Incoming streams can be authenticated using webhooks on both orchestrator and broadcaster nodes. To use these webhooks, node operators must implement their own web service / endpoint to be accessed only by the Livepeer node. As new streams appear, the Livepeer node will call this endpoint to determine whether the given stream is allowed.
 
 Webhooks can authenticate streams supported by the RTMP and HTTP push ingest protocols. See the [ingest documentation](ingest.md) for details on how to use these protocols.
 
--- a/doc/rtmpwebhookauth.md
+++ b/doc/rtmpwebhookauth.md
@@ -54,11 +54,24 @@ There is simple webhook authentication server [example](https://github.com/livep
 
 ## Orchestrators
 
-Webhooks can be used to authenticate discovery requests. When a webhook URL is provided on node startup using the `-authWebhookUrl` flag the Livepeer node will make a `POST` request to the specified URL on each `GetOrchestratorInfo` call. 
+Webhooks can be used to authenticate discovery requests. When a webhook URL is provided on node startup using the `-authWebhookUrl` flag the Livepeer node will make a `POST` request to the specified URL on each `GetOrchestratorInfo` call.
+
+If a valid `priceInfo` object is provided in the response the orchestrator will use it instead of its default price. A valid price requires `pricePerUnit >= 0` and `pixelsPerUnit > 0`.
 
 #### Request Object
 ```json
 {
-    "id": ""
+    "id": string
+}
+```
+
+#### Response Object
+
+```json
+{
+    "priceInfo": {
+        "pricePerUnit": string,
+        "pixelsPerUnit": string
+    }
 }
 ```
--- a/doc/rtmpwebhookauth.md
+++ b/doc/rtmpwebhookauth.md
@@ -2,7 +2,7 @@
 
 Incoming streams can be authenticated using webhooks on both orchestrator and broadcaster nodes. To use these webhooks, node operators must implement their own web service / endpoint to be accessed only by the Livepeer node. As new streams appear, the Livepeer node will call this endpoint to determine whether the given stream is allowed.
 
-Webhooks can authenticate streams supported by the RTMP and HTTP push ingest protocols. See the [ingest documentation](ingest.md) for details on how to use these protocols.
+The webhook server should respond with HTTP status code `200` in order to authenticate / authorize the stream. A response with a HTTP status code other than `200` will cause the Livepeer node to disconnect the stream.
 
 To enable webhook authentication functionality, the Livepeer node should be started with the `-authWebhookUrl` flag, along with the webhook endpoint URL.
 
@@ -11,6 +11,10 @@ For example:
 ```console
 livepeer -authWebhookUrl http://ownserver/auth
 ```
+
+## Broadcasters 
+
+Webhooks can authenticate streams supported by the RTMP and HTTP push ingest protocols. See the [ingest documentation](ingest.md) for details on how to use these protocols.
 
 For each incoming stream, the Livepeer node will make a `POST` request to the `http://ownserver/auth` endpoint, passing the URL of the request as JSON object.
 
@@ -21,8 +25,6 @@ For example, if the incoming request was made to `rtmp://livepeer.node/manifest`
     "url": "rtmp://livepeer.node/manifest"
 }
 ```
-
-The webhook server should respond with HTTP status code `200` in order to authenticate / authorize the stream. A response with a HTTP status code other than `200` will cause the Livepeer node to disconnect the stream.
 
 The webhook may respond with an empty body.  In this case, the `manifestID` property of the stream will be taken from the URL.  If the URL does not specify a manifest id, then it will be generated at random.  Otherwise, the webhook endpoint should respond with a JSON object in the following format:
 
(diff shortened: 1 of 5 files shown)


server/rpc.go
```diff
@@ -302,6 +309,34 @@ func verifyOrchestratorReq(orch Orchestrator, addr ethcommon.Address, sig []byte
 	return orch.CheckCapacity("")
 }
 
+// authenticateBroadcaster returns an error if authentication fails
+func authenticateBroadcaster(id string) error {
+	if AuthWebhookURL == "" {
```
darkdarkdragon, line 334: `AuthWebhookURL` is set only for Broadcaster node

darkdarkdragon (changes requested):

server/rpc.go
kyriediculous: fixed in 9ce9e44f

doc/rtmpwebhookauth.md
```diff
@@ -1,8 +1,8 @@
 # Webhook Authentication
 
-Incoming streams can be authenticated using webhooks. To use these webhooks, node operators must implement their own web service / endpoint to be accessed only by the Livepeer node. As new streams appear, the Livepeer node will call this endpoint to determine whether the given stream is allowed.
+Incoming streams can be authenticated using webhooks on both Orchestrator and Broadcaster nodes. To use these webhooks, node operators must implement their own web service / endpoint to be accessed only by the Livepeer node. As new streams appear, the Livepeer node will call this endpoint to determine whether the given stream is allowed.
 
-Webhooks can authenticate streams supported by the RTMP and HTTP push ingest protocols. See the [ingest documentation](ingest.md) for details on how to use these protocols.
+The webhook server should respond with HTTP status code `200` in order to authenticate / authorize the stream. A response with a HTTP status code other than `200` will cause the Livepeer node to disconnect the stream.
```
darkdarkdragon, line 5: This says that code should be strictly 200 for success, but actual code checks for whole 2XX range. Should we make them match?

darkdarkdragon (changes requested):

doc/rtmpwebhookauth.md
```diff
@@ -1,8 +1,8 @@
 # Webhook Authentication
 
-Incoming streams can be authenticated using webhooks. To use these webhooks, node operators must implement their own web service / endpoint to be accessed only by the Livepeer node. As new streams appear, the Livepeer node will call this endpoint to determine whether the given stream is allowed.
+Incoming streams can be authenticated using webhooks on both Orchestrator and Broadcaster nodes. To use these webhooks, node operators must implement their own web service / endpoint to be accessed only by the Livepeer node. As new streams appear, the Livepeer node will call this endpoint to determine whether the given stream is allowed.
```
yondonfu: I don't think we capitalize orchestrator and broadcaster elsewhere. If that is the case, then I suggest staying consistent with that convention.

server/rpc.go
```diff
@@ -302,6 +309,34 @@ func verifyOrchestratorReq(orch Orchestrator, addr ethcommon.Address, sig []byte
...
+		return err
+	}
+
+	body, err := ioutil.ReadAll(res.Body)
+	defer res.Body.Close()
+	if err != nil {
+		return err
+	}
+
+	if res.StatusCode < 200 || res.StatusCode >= 300 {
```
yondonfu: Second Ivan's earlier comment about matching up the status code check here with the description in the docs. FWIW the webhook authentication on B also strictly checks for a 200 status code.

yondonfu (changes requested):

doc/rtmpwebhookauth.md
kyriediculous: fixed in 4090a12

doc/rtmpwebhookauth.md
kyriediculous: fixed in 4090a12

yondonfu (approved):
LGTM

darkdarkdragon (approved):

Issue #1719, opened 2020-12-24
New Metric: Number of O Swaps on B


ericxtang:

jailuthra:
Closed in #1735

PR #1735, opened 2021-01-12
Support multiple segments in-flight between B & O

### LINKED ISSUE #1658: More flexible segment in flight rule

jailuthra:

Implements flexible segment-in-flight handling between a B and an O as described in https://github.com/livepeer/go-livepeer/issues/1658. Previously only a single segment could be in flight on a particular B<>O session.

- https://github.com/livepeer/go-livepeer/pull/1735/commits/23a7532fc858f23e105a6922b4139fddfaf1dde5 - Add a new metric to count Orchestrator Swaps (needed by https://github.com/livepeer/docker-livepeer/pull/50)
- https://github.com/livepeer/go-livepeer/pull/1735/commits/5dc0b25f1f58043f8bc6a66072b2a6098c08941d 
    - Track the last used BroadcasterSession in a `lastSession` variable and also track all segments-in-flight for a particular session in a queue.
    - Re-use the last session whenever possible - as specified in https://github.com/livepeer/go-livepeer/issues/1658
    - Increment the new metric for "Orchestrator Swaps" whenever new session's O URI is different from last session's O URI

A lot of manual testing -
1. Ensuring session re-use when simultaneous segments arriving at the same session (i.e. get a 200 response for all segments when sent as described in https://github.com/livepeer/go-livepeer/issues/1665 even when B knows only one O)
2. Ensuring Segments-In-Flight queue is cleaned up when all segments return for a particular session
3. Check the new `mOrchestratorSwaps` metric is updated whenever a swap occurs

**TODO**
- [x] Fix existing Unit Tests
- [x] Write unit tests for new code
- [x] Remove Log lines
- [x] Race condition checks

Fixes #1658 and #1719

**Checklist:**

### DIFF
--- a/core/orchestrator.go
+++ b/core/orchestrator.go
@@ -31,6 +31,8 @@ import (
 	"github.com/livepeer/lpms/stream"
 )
 
+const maxSegmentChannels = 4
+
 var transcodeLoopTimeout = 1 * time.Minute
 
 // Gives us more control of "timeout" / cancellation behavior during testing
@@ -447,7 +449,7 @@ func (n *LivepeerNode) getSegmentChan(md *SegTranscodingMetadata) (SegmentChan, 
 	if len(n.SegmentChans) >= MaxSessions {
 		return nil, ErrOrchCap
 	}
-	sc := make(SegmentChan, 1)
+	sc := make(SegmentChan, maxSegmentChannels)
 	glog.V(common.DEBUG).Infof("Creating new segment chan for manifestID=%s sessionID=%s", md.ManifestID, md.AuthToken.SessionId)
 	if err := n.transcodeSegmentLoop(md, sc); err != nil {
 		return nil, err
--- a/monitor/census.go
+++ b/monitor/census.go
@@ -132,6 +132,7 @@ type (
 		mRecordingSaveLatency         *stats.Float64Measure
 		mRecordingSaveErrors          *stats.Int64Measure
 		mRecordingSavedSegments       *stats.Int64Measure
+		mOrchestratorSwaps            *stats.Int64Measure
 
 		// Metrics for sending payments
 		mTicketValueSent    *stats.Float64Measure
@@ -255,6 +256,7 @@ func InitCensus(nodeType NodeType, version string) {
 		"How long it takes to save segment to the OS", "sec")
 	census.mRecordingSaveErrors = stats.Int64("recording_save_errors", "Number of errors during save to the recording OS", "tot")
 	census.mRecordingSavedSegments = stats.Int64("recording_saved_segments", "Number of segments saved to the recording OS", "tot")
+	census.mOrchestratorSwaps = stats.Int64("orchestrator_swaps", "Number of orchestrator swaps mid-stream", "tot")
 
 	// Metrics for sending payments
 	census.mTicketValueSent = stats.Float64("ticket_value_sent", "TicketValueSent", "gwei")
@@ -591,6 +593,13 @@ func InitCensus(nodeType NodeType, version string) {
 			TagKeys:     baseTags,
 			Aggregation: view.LastValue(),
 		},
+		{
+			Name:        "orchestrator_swaps",
+			Measure:     census.mOrchestratorSwaps,
+			Description: "Number of orchestrator swaps mid-stream",
+			TagKeys:     baseTags,
+			Aggregation: view.Count(),
+		},
 
 		// Metrics for sending payments
 		{
@@ -888,6 +897,12 @@ func MaxSessions(maxSessions int) {
 	census.lock.Lock()
 	defer census.lock.Unlock()
 	stats.Record(census.ctx, census.mMaxSessions.M(int64(maxSessions)))
+}
+
+func OrchestratorSwapped() {
+	census.lock.Lock()
+	defer census.lock.Unlock()
+	stats.Record(census.ctx, census.mOrchestratorSwaps.M(1))
 }
 
 func CurrentSessions(currentSessions int) {
--- a/server/broadcast.go
+++ b/server/broadcast.go
@@ -42,6 +42,11 @@ type BroadcastConfig struct {
 	mu       sync.RWMutex
 }
 
+type SegFlightMetadata struct {
+	startTime time.Time
+	segDur    time.Duration
+}
+
 func (cfg *BroadcastConfig) MaxPrice() *big.Rat {
 	cfg.mu.RLock()
 	defer cfg.mu.RUnlock()
@@ -61,6 +66,7 @@ type BroadcastSessionsManager struct {
 	mid      core.ManifestID
 	sel      BroadcastSessionsSelector
 	sessMap  map[string]*BroadcastSession
+	lastSess *BroadcastSession
 	numOrchs int // how many orchs to request at once
 	poolSize int
 
@@ -80,18 +86,35 @@ func (bsm *BroadcastSessionsManager) selectSession() *BroadcastSession {
 		if numSess < int(math.Ceil(float64(m.numOrchs)/2.0)) {
 			go m.refreshSessions()
 		}
-		return numSess > 0
+		return (numSess > 0 || bsm.lastSess != nil)
 	}
 	for checkSessions(bsm) {
-		sess := bsm.sel.Select()
-		// Handle the case where there is an error during selection and Select() returns nil when session list length > 0
+		var sess *BroadcastSession
+
+		if bsm.lastSess != nil && len(bsm.lastSess.SegsInFlight) > 0 &&
+			time.Now().Sub(bsm.lastSess.SegsInFlight[0].startTime) < bsm.lastSess.SegsInFlight[0].segDur {
+			// Re-use last session if oldest segment is in-flight for < segDur
+			sess = bsm.lastSess
+		} else {
+			// Or try a new session from the available ones
+			sess = bsm.sel.Select()
+		}
+
+		// If no new sessions are available, re-use last session when oldest segment is in-flight for < 2 * segDur
+		if sess == nil && bsm.lastSess != nil && len(bsm.lastSess.SegsInFlight) > 0 &&
+			time.Now().Sub(bsm.lastSess.SegsInFlight[0].startTime) < 2*bsm.lastSess.SegsInFlight[0].segDur {
+			sess = bsm.lastSess
+		}
+
+		// No session found, return nil
 		if sess == nil {
+			if bsm.lastSess != nil {
+				bsm.lastSess.SegsInFlight = nil
+				bsm.lastSess = nil
+			}
 			return nil
 		}
 
-		if _, ok := bsm.sessMap[sess.OrchestratorInfo.Transcoder]; ok {
-			return sess
-		}
 		/*
 		   Don't select sessions no longer in the map.
 
@@ -100,6 +123,24 @@ func (bsm *BroadcastSessionsManager) selectSession() *BroadcastSession {
 		   To avoid a runtime search of the session list under lock, simply
 		   fixup the session list at selection time by retrying the selection.
 		*/
+		if _, ok := bsm.sessMap[sess.OrchestratorInfo.Transcoder]; ok {
+			if monitor.Enabled && bsm.lastSess != nil && bsm.lastSess.OrchestratorInfo.Transcoder != sess.OrchestratorInfo.Transcoder {
+				monitor.OrchestratorSwapped()
+			}
+			bsm.lastSess = sess
+			return sess
+		}
+
+		// Last session got removed from map (possibly due to a failure) so stop tracking its in-flight segments
+		if bsm.lastSess != nil && sess.OrchestratorInfo.Transcoder == bsm.lastSess.OrchestratorInfo.Transcoder {
+			bsm.lastSess.SegsInFlight = nil
+			bsm.lastSess = nil
+		}
+	}
+	// No session found, return nil
+	if bsm.lastSess != nil {
+		bsm.lastSess.SegsInFlight = nil
+		bsm.lastSess = nil
 	}
 	return nil
 }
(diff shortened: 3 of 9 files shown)


monitor/census.go
```diff
@@ -890,6 +899,16 @@ func MaxSessions(maxSessions int) {
 	stats.Record(census.ctx, census.mMaxSessions.M(int64(maxSessions)))
 }
 
+func OrchestratorSwapped() {
+	census.orchestratorSwapped()
```
yondonfu: Since `orchestratorSwapped()` is only called here, could we just move the code from that method here similar to what is done in `CurrentSessions()` below?

server/broadcast.go
```diff
@@ -74,24 +80,57 @@ type BroadcastSessionsManager struct {
 func (bsm *BroadcastSessionsManager) selectSession() *BroadcastSession {
 	bsm.sessLock.Lock()
 	defer bsm.sessLock.Unlock()
+	numSess := bsm.sel.Size()
```
yondonfu: Moving this method call outside of `checkSessions()` means that `checkSessions()` will not see any updates to the number of sessions which is updated by the `refreshSessions()`. I think this can result in the following scenario:

- `numSess > 0`
- `checkSessions()` starts a session refresh by calling `refreshSessions()` because `numSess` is below the desired threshold
- `checkSessions()` returns true
- If the loop on L91 runs more than once (i.e. if the first selected session has already been removed from `sessMap`)
    - `checkSessions()` always returns true because `numSess` is never updated
    - `refreshSessions()` is always called in `checkSessions()` because `numSess` is still below the desired threshold

Can we preserve the currently property of `checkSessions()` always seeing updates to `numSess`?

server/broadcast.go
```diff
@@ -74,24 +80,57 @@ type BroadcastSessionsManager struct {
...
-		sess := bsm.sel.Select()
-		// Handle the case where there is an error during selection and Select() returns nil when session list length > 0
+		var sess *BroadcastSession
+
+		if bsm.lastSess != nil && len(bsm.lastSess.SegsInFlight) > 0 &&
+			time.Now().Sub(bsm.lastSess.SegsInFlight[0].startTime) < bsm.lastSess.SegsInFlight[0].segDur {
+			// Re-use last session if oldest segment is in-flight for < segDur
+			glog.Warningf("REUSING OLD SESSION")
+			sess = bsm.lastSess
+		} else if numSess > 0 {
```
yondonfu: Is the `numSess` check here necessary? Would it suffice to just use an else block here since `sel.Select()` would return nil if there are no sessions left and we have a nil check for the returned session on L105?

server/broadcast.go
```diff
@@ -429,6 +486,11 @@ func transcodeSegment(cxn *rtmpConnection, seg *stream.HLSSegment, name string,
 		sess = newSess
 	}
 
+	cxn.sessManager.sessLock.Lock()
+	sess.SegsInFlight = append(sess.SegsInFlight, SegFlightMetadata{startTime: time.Now(), segDur: time.Duration(int64(seg.Duration*1e9)) * time.Nanosecond})
+	cxn.sessManager.sessLock.Unlock()
+	glog.Warningf("SUBMITTING SEG %v Session=%v InFlightStart=%v SegDur=%v", seg.SeqNo, sess.OrchestratorInfo.Transcoder, sess.SegsInFlight[len(sess.SegsInFlight)-1].startTime, sess.SegsInFlight[len(sess.SegsInFlight)-1].segDur)
```
yondonfu: Possible race here since we read SegsInFlight without a lock? (Not a problem if this log is going to be removed anyway).

server/broadcast.go
```diff
@@ -429,6 +486,11 @@ func transcodeSegment(cxn *rtmpConnection, seg *stream.HLSSegment, name string,
 		sess = newSess
 	}
 
+	cxn.sessManager.sessLock.Lock()
+	sess.SegsInFlight = append(sess.SegsInFlight, SegFlightMetadata{startTime: time.Now(), segDur: time.Duration(int64(seg.Duration*1e9)) * time.Nanosecond})
```
yondonfu: Could use `time.Duration(seg.Duration * float64(time.Second))` since I think seg.Duration is already a denominated in seconds to avoid extra unit conversions.

server/broadcast.go
```diff
@@ -116,12 +160,24 @@ func (bsm *BroadcastSessionsManager) completeSession(sess *BroadcastSession) {
 	defer bsm.sessLock.Unlock()
 
 	if existingSess, ok := bsm.sessMap[sess.OrchestratorInfo.Transcoder]; ok {
+		if bsm.lastSess != nil && bsm.lastSess.OrchestratorInfo.Transcoder == sess.OrchestratorInfo.Transcoder {
+			if len(bsm.lastSess.SegsInFlight) > 1 {
+				bsm.lastSess.SegsInFlight = bsm.lastSess.SegsInFlight[1:]
+				// don't place session back into the selector in case a segment is still in flight
+				glog.Warningf("NOT COMPLETING SESSION AS SEGMENT IN FLIGHT")
+				return
+			} else {
```
yondonfu: If we're returning in the if block, it's common to skip the else block in Go i.e.

```golang
if (len(bsm.lastSess.SegsInFlight) > 1 {
    ...
    return
}

...
bsm.lastSess = sess
```

server/broadcast.go
```diff
@@ -74,24 +80,57 @@ type BroadcastSessionsManager struct {
...
 		}
 
 		if _, ok := bsm.sessMap[sess.OrchestratorInfo.Transcoder]; ok {
+			if monitor.Enabled && bsm.lastSess != nil && bsm.lastSess.OrchestratorInfo.Transcoder != sess.OrchestratorInfo.Transcoder {
+				monitor.OrchestratorSwapped()
+			}
+			bsm.lastSess = sess
+			glog.Warningf("RETURNING SESSION %v", sess.OrchestratorInfo.Transcoder)
 			return sess
+		} else if bsm.lastSess != nil && sess.OrchestratorInfo.Transcoder == bsm.lastSess.OrchestratorInfo.Transcoder {
```
yondonfu: If we're returning in the if block, it's common to skip the else if block in Go i.e.

```golang
if ... {
}

if bsm.lastSess != nil && ... {
}
```

yondonfu (changes requested):
Took a first pass.

We should probably also make sure that the orchestrator can support multiple in-flight segments as mentioned [here](https://github.com/livepeer/go-livepeer/issues/1658#issuecomment-737423979). 

I think it'd be a good idea to run some of the unit tests for this new code with the race detector ([example](https://github.com/livepeer/go-livepeer/blob/897d5760c9aef0519f0dae0b4e34be7888a3a0bc/test.sh#L24)) since we had some concerns around concurrent session usage.

server/broadcast.go
jailuthra: Ah I encountered an error when `sel.Select()` was called on an empty list with `LIFOSelector` but I guess a better solution would be to add a check in it's `Select()` method and return nil - rather than adding a check during calling. Will do in the update and revert back the `numSess` changes.

server/broadcast.go
jailuthra: Yep planning to remove the log before push!

server/broadcast.go
jailuthra: Good catch, I completely missed that we'd want to keep that variable updated between loop iterations.

jailuthra:
- Fixed all review comments
- Completed Todo in PR description
- Added a new unit test to reproduce the manual testing done with multiple in-flight segments and verify the state variables are set correctly 
- Added race detection in `test.sh` script

Will squash/cleanup commit history before merge, once approved

core/orchestrator.go
```diff
@@ -447,7 +447,7 @@ func (n *LivepeerNode) getSegmentChan(md *SegTranscodingMetadata) (SegmentChan,
 	if len(n.SegmentChans) >= MaxSessions {
 		return nil, ErrOrchCap
 	}
-	sc := make(SegmentChan, 1)
+	sc := make(SegmentChan, 4)
```
yondonfu: Can we define a constant for this so its clear that its an important parameter that controls the maximum # of segments that can be queued on an orchestrators?

server/broadcast.go
```diff
@@ -100,6 +123,23 @@ func (bsm *BroadcastSessionsManager) selectSession() *BroadcastSession {
...
+			if monitor.Enabled && bsm.lastSess != nil && bsm.lastSess.OrchestratorInfo.Transcoder != sess.OrchestratorInfo.Transcoder {
+				monitor.OrchestratorSwapped()
+			}
+			bsm.lastSess = sess
+			return sess
+		}
+
+		// Last session got removed from map (possibly due to a failure) so stop tracking its in-flight segments
+		if bsm.lastSess != nil && sess.OrchestratorInfo.Transcoder == bsm.lastSess.OrchestratorInfo.Transcoder {
+			bsm.lastSess.SegsInFlight = nil
```
yondonfu, line 136: If the last session was removed from the map should we also set `bsm.lastSess = nil` here since we know that we wouldn't want to use the last session in the next iteration of the loop?

server/broadcast.go
```diff
@@ -116,12 +156,19 @@ func (bsm *BroadcastSessionsManager) completeSession(sess *BroadcastSession) {
 	defer bsm.sessLock.Unlock()
 
 	if existingSess, ok := bsm.sessMap[sess.OrchestratorInfo.Transcoder]; ok {
+		if bsm.lastSess != nil && bsm.lastSess.OrchestratorInfo.Transcoder == sess.OrchestratorInfo.Transcoder {
+			if len(bsm.lastSess.SegsInFlight) > 1 {
+				bsm.lastSess.SegsInFlight = bsm.lastSess.SegsInFlight[1:]
+				// don't place session back into the selector in case a segment is still in flight
```
yondonfu: I think we should still set `bsm.lastSess` and update `sessMap` (L170) because the provided `sess` is going to be a new copy (see [this line](https://github.com/livepeer/go-livepeer/blob/897d5760c9aef0519f0dae0b4e34be7888a3a0bc/server/broadcast.go#L442)) that was updated based on an existing session. If we don't make these updates, we could end up using a session for the same orchestrator later on with out-dated metadata fields. We could write a test to verify this by changing the `OrchestratorInfo` message returned with a segment response.

server/broadcast_test.go
```diff
@@ -323,6 +326,124 @@ func TestSelectSession(t *testing.T) {
 	// XXX check refresh condition more precisely - currently numOrchs / 2
 }
 
+func TestSelectSession_MultipleInFlight(t *testing.T) {
```
yondonfu, line 329: I see that we're running tests prefixed with `TestSelectSession_` using the race detector, but I don't see any tests involving concurrency below. Concurrent access for the last session seems possible with:

- `selectSession()` called in `transcodeSegment()`
- `completeSession()` called in `transcodeSegment()`
- The update of the `SegsInFlight` for the last session in `transcodeSegment()`

Perhaps we can just setup scenarios where those 3 events occur concurrently (i.e. parallel goroutines).

yondonfu (changes requested):

server/broadcast.go
jailuthra: This wasn't a hard requirement as we would also test for `len(bsm.lastSess.SegsInFlight) > 0` in next iteration and then end up using a new session or nil.
But adding this definitely makes the code more explicit/clear so done that in the update!

server/broadcast_test.go
jailuthra: Will send another quick PR for this as this PR blocks other things!

yondonfu (approved):
LGTM after squashing

PR #1736, opened 2021-01-13
server: discovery auth webhook priceInfo setting

### LINKED ISSUE #1723: Auth webhook configurable price
See https://github.com/livepeer/go-livepeer/issues/1569#issuecomment-657316909.

kyriediculous:
Implements a `PriceInfo` field on the discovery auth webhook response to be used for further price settings when sending `TicketParams` instead of the Orchestrator's default set price + overhead. 

- Rename `authBroadcasterRes` to `discoveryAuthWebhookRes`
- Add `PriceInfo` field to `discoveryAuthWebhookRes` 
- Add a `cache.Cache` , `discoveryAuthWebhookCache` to store webhook responses 
- Add a function `getPriceInfo()` in `server/rpc.go` that reads PriceInfo from the `discoveryAuthWebhookCache` for a broadcaster, if none exists returns default orchestrator price + overhead instead.
- use the new `getPriceInfo()` to get price in `orchestratorInfo()`
- Add unit tests 

unit tests

Fixes #1723 

**Checklist:**

### DIFF
--- a/server/rpc.go
+++ b/server/rpc.go
@@ -22,6 +22,7 @@ import (
 	"github.com/livepeer/go-livepeer/pm"
 	ffmpeg "github.com/livepeer/lpms/ffmpeg"
 	"github.com/livepeer/lpms/stream"
+	"github.com/patrickmn/go-cache"
 
 	ethcommon "github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/crypto"
@@ -34,6 +35,9 @@ const GRPCConnectTimeout = 3 * time.Second
 const GRPCTimeout = 8 * time.Second
 
 var authTokenValidPeriod = 30 * time.Minute
+var discoveryAuthWebhookCacheCleanup = 5 * time.Minute
+
+var discoveryAuthWebhookCache = cache.New(authTokenValidPeriod, discoveryAuthWebhookCacheCleanup)
 
 type Orchestrator interface {
 	ServiceURI() *url.URL
@@ -253,26 +257,29 @@ func genOrchestratorReq(b common.Broadcaster) (*net.OrchestratorRequest, error) 
 func getOrchestrator(orch Orchestrator, req *net.OrchestratorRequest) (*net.OrchestratorInfo, error) {
 	addr := ethcommon.BytesToAddress(req.Address)
 	if err := verifyOrchestratorReq(orch, addr, req.Sig); err != nil {
-		return nil, fmt.Errorf("Invalid orchestrator request (%v)", err)
+		return nil, fmt.Errorf("Invalid orchestrator request: %v", err)
 	}
 
-	webhookRes, err := authenticateBroadcaster(addr.Hex())
-	if err != nil {
+	if _, err := authenticateBroadcaster(addr.Hex()); err != nil {
 		return nil, fmt.Errorf("authentication failed: %v", err)
 	}
 
 	// currently, orchestrator == transcoder
-	oInfo, err := orchestratorInfo(orch, addr, orch.ServiceURI().String())
-	if err != nil {
-		return nil, err
-	}
+	return orchestratorInfo(orch, addr, orch.ServiceURI().String())
+}
 
-	// Adjust the necessary orchestrator info values
-	return handleWebhookResponse(webhookRes, oInfo), nil
+func getPriceInfo(orch Orchestrator, addr ethcommon.Address) (*net.PriceInfo, error) {
+	if AuthWebhookURL != "" {
+		webhookRes := getFromDiscoveryAuthWebhookCache(addr.Hex())
+		if webhookRes != nil && webhookRes.PriceInfo != nil {
+			return webhookRes.PriceInfo, nil
+		}
+	}
+	return orch.PriceInfo(addr)
 }
 
 func orchestratorInfo(orch Orchestrator, addr ethcommon.Address, serviceURI string) (*net.OrchestratorInfo, error) {
-	priceInfo, err := orch.PriceInfo(addr)
+	priceInfo, err := getPriceInfo(orch, addr)
 	if err != nil {
 		return nil, err
 	}
@@ -317,10 +324,13 @@ func verifyOrchestratorReq(orch Orchestrator, addr ethcommon.Address, sig []byte
 	return orch.CheckCapacity("")
 }
 
-type authBroadcasterResponse struct{}
+type discoveryAuthWebhookRes struct {
+	PriceInfo *net.PriceInfo `json:"priceInfo,omitempty"`
+}
 
 // authenticateBroadcaster returns an error if authentication fails
-func authenticateBroadcaster(id string) (*authBroadcasterResponse, error) {
+// on success it caches the webhook response
+func authenticateBroadcaster(id string) (*discoveryAuthWebhookRes, error) {
 	if AuthWebhookURL == "" {
 		return nil, nil
 	}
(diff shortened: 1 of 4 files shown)


server/rpc.go
```diff
@@ -316,15 +322,18 @@ func verifyOrchestratorReq(orch Orchestrator, addr ethcommon.Address, sig []byte
...
 
-type authBroadcasterResponse struct{}
+type discoveryAuthWebhookRes struct {
+	PriceInfo *net.PriceInfo `json:"price,omitempty"`
+}
 
 // authenticateBroadcaster returns an error if authentication fails
-func authenticateBroadcaster(id string) (*authBroadcasterResponse, error) {
+// on success it caches the webhook response
+func authenticateBroadcaster(addr ethcommon.Address) (*discoveryAuthWebhookRes, error) {
```
yondonfu: I don't think this function needs to know that the ID is an ETH address - the ID can be opaque here leaving its interpretation to the webhook server implementation. Might also be useful to keep the ID as a string if the format is updated later on.

server/rpc.go
```diff
@@ -35,6 +36,8 @@ const GRPCTimeout = 8 * time.Second
 
 var authTokenValidPeriod = 30 * time.Minute
 
+var discoveryAuthWebhookCache = cache.New(authTokenValidPeriod, authTokenValidPeriod)
```
yondonfu: We might want the cleanup interval to be shorter otherwise in the worst case we could end up with cache items that expire after 30 minutes, but that are not cleaned up for awhile (i.e. something like 29 minutes).

server/rpc.go
```diff
@@ -255,23 +258,26 @@ func getOrchestrator(orch Orchestrator, req *net.OrchestratorRequest) (*net.Orch
...
-	}
+	return orchestratorInfo(orch, addr, orch.ServiceURI().String())
+}
 
-	// Adjust the necessary orchestrator info values
-	return handleWebhookResponse(webhookRes, oInfo), nil
+func getPriceInfo(orch Orchestrator, addr ethcommon.Address) (*net.PriceInfo, error) {
+	if AuthWebhookURL != "" {
+		webhookRes := getFromDiscoveryAuthWebhookCache(addr)
+		if webhookRes != nil && webhookRes.PriceInfo != nil {
```
yondonfu, line 274: Will this result in `orch.PriceInfo()` being called if the webhook does not return a price?

How do we expect the webhook server to tell the orchestrator to use its default price? Should the webhook server omit the price field, set the price field to null, or would any of those options suffice? Would be good to document the requirements here.

server/rpc.go
```diff
@@ -344,17 +353,36 @@ func authenticateBroadcaster(id string) (*authBroadcasterResponse, error) {
 		return nil, errors.New(string(body))
 	}
 
-	webhookRes := &authBroadcasterResponse{}
+	webhookRes := &discoveryAuthWebhookRes{}
 	if err := json.Unmarshal(body, webhookRes); err != nil {
 		return nil, err
 	}
 
+	addToDiscoveryAuthWebhookCache(addr, webhookRes, authTokenValidPeriod)
```
yondonfu: Should we only hit the webhook if there is not already a cached response and return the cached response otherwise?

IIUC the flow will look something like this:

1. B calls GetOrchestrator on O
2. O hits the webhook which returns a price and caches the response which will expire when the auth token associated with this OrchestratorInfo response expires
3. O uses the webhook price instead of calling `orch.PriceInfo()` and it is used as an input into the HMAC construction for the recipientRandHash in the TicketParams sent back to B
4. Whenever O receives a payment, it will be able to confirm that the price included was a price it advertised that has not yet expired (based on the TicketParams expiration period)
5. O also returns an OrchestratorInfo message after each segment. Since we want to avoid hitting the webhook (we want to hit the webhook only during discovery and not during the "hot" transcoding flow) here we want to read from a cache instead.

yondonfu (changes requested):

server/rpc.go
kyriediculous: Hmm good question. 

We will (under happy conditions) only store cache for an O if authentication passes, in other words it is not a blacklisted node. 

This response can be an empty object or contain `priceInfo`, both are totally fine, this will indicate that the broadcaster is authenticated. 

From this point on I don't think we will ever call the webhook in the transcoding path, only during discovery. 

So yes we will hit the webhook on every `getOrchestratorInfo` call , even if there already is cache available. But I'm not sure whether that's harmful.  Otherwise I felt like updating the cache would rely too much on it being cleaned by the cleanup interval. 

Looking forward to your thoughts.

server/rpc.go
kyriediculous: Should definitely document this , we will always call `orch.PriceInfo` if the webhook response in the cache doesn't have a `priceInfo` field defined or its value is undefined.

server/rpc.go
kyriediculous: fixed in 7dd34f5

server/rpc.go
kyriediculous: fixed in 7dd34f5

server/rpc.go
```diff
@@ -252,26 +258,29 @@ func genOrchestratorReq(b common.Broadcaster) (*net.OrchestratorRequest, error)
...
 	if err := verifyOrchestratorReq(orch, addr, req.Sig); err != nil {
-		return nil, fmt.Errorf("Invalid orchestrator request (%v)", err)
+		return nil, fmt.Errorf("Invalid orchestrator request err=%v", err)
 	}
 
-	webhookRes, err := authenticateBroadcaster(addr.Hex())
-	if err != nil {
-		return nil, fmt.Errorf("authentication failed: %v", err)
+	if _, err := authenticateBroadcaster(addr.Hex()); err != nil {
+		return nil, fmt.Errorf("authentication failed err=%v", err)
```
yondonfu: I don't think we need to use the err= convention unless we're logging something. Otherwise if we also use the this convention when these errors are logged we'll end up with log statements that look like `err=...err=` which is a bit noisy.

server/rpc.go
```diff
@@ -35,6 +36,11 @@ const GRPCTimeout = 8 * time.Second
 
 var authTokenValidPeriod = 30 * time.Minute
 
+// using authTokenValidPeriod minus 1 minute as cache duration. With a 30m duration, worst case we could end up with cache items that expire after 30 minutes
```
yondonfu: Why set the default cleanup interval to 29 minutes here? A 29 minute cleanup interval doesn't seem like it would result in much of a difference vs. a 30 minute cleanup interval.

Thinking back to my previous comment, it also seems like the cleanup interval value here may end up not being that important because if a cached response is stale, then that should mean the auth token used by a B has expired which would require the B to refresh the auth token triggering a webhook call that would overwrite the expired cached response anyway.

server/rpc.go
yondonfu: Ah I was concerned about hitting the webhook (generally want to minimize external network calls during the transcoding flow to minimize potential impact of service degradation/unavailability of webhook) whenever a new OrchestratorInfo message is created in ServeSegment, but missed the fact that authenticateBroadcaster is only called for GetOrchestratorInfo calls and not in the orchestratorInfo function which is invoked in ServeSegment for each segment.

The only remaining case where the webhook will be hit during the transcoding flow is if B needs to refresh its auth token and/or ticket params. Hitting the webhook when the auth token is expired makes sense since the B needs to re-authenticate. Hitting the webhook when the ticket params are expired is not necessary, but might be ok to avoid having to distinguish between these two cases on the orchestrator side. The # of webhook calls during the transcoding flow for the expired auth token case should be ok given the 30 minute validity period. The # of webhook calls during the transcoding flow for the expired ticket params case will be higher given the shorter validity period there. Okay for now, but just noting to keep in mind.

server/rpc.go
kyriediculous: My bad , I misunderstood your earlier comment

server/rpc.go
kyriediculous: I'd be okay with checking for existing cache in `authenticateBroadCaster()` and/or lengthening ticket params expiry if that sits better.

server/rpc.go
```diff
@@ -252,26 +259,29 @@ func genOrchestratorReq(b common.Broadcaster) (*net.OrchestratorRequest, error)
 func getOrchestrator(orch Orchestrator, req *net.OrchestratorRequest) (*net.OrchestratorInfo, error) {
 	addr := ethcommon.BytesToAddress(req.Address)
 	if err := verifyOrchestratorReq(orch, addr, req.Sig); err != nil {
-		return nil, fmt.Errorf("Invalid orchestrator request (%v)", err)
+		return nil, fmt.Errorf("Invalid orchestrator reques: %v", err)
```
yondonfu: "reques" typo

server/rpc.go
```diff
@@ -34,6 +35,12 @@ const GRPCConnectTimeout = 3 * time.Second
 const GRPCTimeout = 8 * time.Second
 
 var authTokenValidPeriod = 30 * time.Minute
+var discoveryAuthWebhookCacheCleanup = 5 * time.Minute
+
+// using authTokenValidPeriod minus 1 minute as cache duration. With a 30m duration, worst case we could end up with cache items that expire after 30 minutes
```
yondonfu: Let's remove this comment as it doesn't seem relevant anymore.

yondonfu (changes requested):

yondonfu (approved):
LGTM pending passing tests

PR #1749, opened 2021-01-28
server,core: Improve in-flight segment handling

### LINKED ISSUE #1742: Improve logging around choosing BroadcastSession after a delayed segment
After the changes for flexible segments-in-flight #1658 it is unclear why a particular broadcast session decides to stick with a slightly poorly performing O.

- If a segment takes longer than real-time (i.e. `timeInFlight > segDur`) add log-lines around whatever decision is taken - switch to different O, stick to same O or dropping the segment
- Also ensure via manual testing + unit test if there is a missed edge case anywhere in the current FSIF implementation

**Additional context**

### LINKED ISSUE #1743: Improve the current flexible-segment-in-flight logic
**Bug 1: B does not factor in download time when deciding whether to re-use last broadcast session**

This can cause a B to stick with an O for a stream even if it is not keeping up in real-time.

From @jailuthra 

> Ohhh ok I figured it out. These are the segments with > 2s overall latency https://my.papertrailapp.com/systems/lon-prod-livepeer-broadcaster-4cks-686c4f4f44-md984/events?focus=1287124139278770185&q=nonce%3D17176576702808233726%20latency%3D2.

> And for many of them, I see that O responds to the B that it has completed transcoding at the end of SubmitSegment when timeInFlight < 2s and that's when the current FSIF implementations marks it as done https://gist.github.com/jailuthra/8f8684ac884d086b1b4d351847feb08c#file-rkv_latency-txt-L6

> When ideally it should mark it as done only when all the renditions have been downloaded (or at least whenever we're logging the latency=2.xyz line https://gist.github.com/jailuthra/8f8684ac884d086b1b4d351847feb08c#file-rkv_latency-txt-L12)

**Solution:** Update the segment-in-flight-queue only after downloading all rendition(s) from the O around here https://github.com/livepeer/go-livepeer/blob/6fe41d3f29d6aff640251c1f25bd713ba8f4b8ac/server/broadcast.go#L634

--------------

**Bug 2: First `completeSession` clears the SegmentInFlight queue for lastSess, losing data in case multiple segments were in flight**

This happens because `updateSession` sets the slice variable to nil https://github.com/livepeer/go-livepeer/blob/6fe41d3f29d6aff640251c1f25bd713ba8f4b8ac/server/broadcast.go#L751 and `completeSession` sets `bsm.lastSess` to this updated session https://github.com/livepeer/go-livepeer/blob/6fe41d3f29d6aff640251c1f25bd713ba8f4b8ac/server/broadcast.go#L510

**Solution:** Ensure the segments-in-flight queue is retained in `updateSession` either via a deep copy or a pointer-to-a-slice or switch to using a proper queue data structure to keep things clear.

------------

**Bug 3: Channel buffer size == 1 for the loadbalancing transcoder module, leading to TranscoderBusy errors**

From @yondonfu 

> https://github.com/livepeer/go-livepeer/pull/1735/commits/882ed38fc1383dd8ae25e6abd20fd9e5a4a104c9 increased the # of in-flight segments supported on O, but there is another in-flight segment restriction via the channel buffer size on the transcoder load balancer. Since the channel buffer size on the transcoder load balancer is still 1, if O receives multiple in-flight segments TranscoderBusy would be returned
https://github.com/livepeer/go-livepeer/blob/6fe41d3f29d6aff640251c1f25bd713ba8f4b8ac/core/lb.go#L202.

jailuthra:

Fix the bugs in, and clean up the logic of the recently introduced flexible segment-in-flight feature between the B & O.

- Remove segments from the in-flight queue only after downloading
- Add goroutines in the test to check for race conditions around `selectSession()` and `completeSession()`
- Simplify `completeSession` logic, and always make sure bsm.lastSess is set to the updated session (with the updated SegsInFlight queue)
- Increase channel buffer count of LB transcoder to match that of the O set previously

- Manual testing with simultaneous curl calls
- Unit tests with stubbed sessions

Fixes #1742 and Fixes #1743

**Checklist:**

### DIFF
--- a/server/broadcast.go
+++ b/server/broadcast.go
@@ -103,6 +103,7 @@ func (bsm *BroadcastSessionsManager) selectSession() *BroadcastSession {
 		// If no new sessions are available, re-use last session when oldest segment is in-flight for < 2 * segDur
 		if sess == nil && bsm.lastSess != nil && len(bsm.lastSess.SegsInFlight) > 0 &&
 			time.Now().Sub(bsm.lastSess.SegsInFlight[0].startTime) < 2*bsm.lastSess.SegsInFlight[0].segDur {
+			glog.V(common.DEBUG).Infof("No sessions in the selector for manifestID=%v re-using orch=%v with acceptable in-flight time", bsm.mid, bsm.lastSess.OrchestratorInfo.Transcoder)
 			sess = bsm.lastSess
 		}
 
@@ -124,8 +125,11 @@ func (bsm *BroadcastSessionsManager) selectSession() *BroadcastSession {
 		   fixup the session list at selection time by retrying the selection.
 		*/
 		if _, ok := bsm.sessMap[sess.OrchestratorInfo.Transcoder]; ok {
-			if monitor.Enabled && bsm.lastSess != nil && bsm.lastSess.OrchestratorInfo.Transcoder != sess.OrchestratorInfo.Transcoder {
-				monitor.OrchestratorSwapped()
+			if bsm.lastSess != nil && bsm.lastSess.OrchestratorInfo.Transcoder != sess.OrchestratorInfo.Transcoder {
+				glog.V(common.DEBUG).Infof("Swapping from orch=%v to orch=%v for manifestID=%s", bsm.lastSess.OrchestratorInfo.Transcoder, sess.OrchestratorInfo.Transcoder, bsm.mid)
+				if monitor.Enabled {
+					monitor.OrchestratorSwapped()
+				}
 			}
 			bsm.lastSess = sess
 			return sess
@@ -133,6 +137,10 @@ func (bsm *BroadcastSessionsManager) selectSession() *BroadcastSession {
 
 		// Last session got removed from map (possibly due to a failure) so stop tracking its in-flight segments
 		if bsm.lastSess != nil && sess.OrchestratorInfo.Transcoder == bsm.lastSess.OrchestratorInfo.Transcoder {
+			glog.V(common.DEBUG).Infof("Removing orch=%v from manifestID=%s session list", bsm.lastSess.OrchestratorInfo.Transcoder, bsm.mid)
+			if monitor.Enabled {
+				monitor.OrchestratorSwapped()
+			}
 			bsm.lastSess.SegsInFlight = nil
 			bsm.lastSess = nil
 		}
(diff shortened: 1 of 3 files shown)


server/broadcast.go
```diff
@@ -103,6 +103,7 @@ func (bsm *BroadcastSessionsManager) selectSession() *BroadcastSession {
 		// If no new sessions are available, re-use last session when oldest segment is in-flight for < 2 * segDur
 		if sess == nil && bsm.lastSess != nil && len(bsm.lastSess.SegsInFlight) > 0 &&
 			time.Now().Sub(bsm.lastSess.SegsInFlight[0].startTime) < 2*bsm.lastSess.SegsInFlight[0].segDur {
+			glog.Warningf("No sessions remaining in the selector for manifestID=%v, so re-using older session orch=%v as in-flight time is under bounds", bsm.mid, bsm.lastSess.OrchestratorInfo.Transcoder)
```
yondonfu: Can we set the verbosity to debug here (I think this will require the use of `Infof()` instead)? There are generally a lot of logs already and this one seems like it'll only be needed by advanced users and/or users that are debugging node behavior.

server/broadcast.go
```diff
@@ -124,15 +125,22 @@ func (bsm *BroadcastSessionsManager) selectSession() *BroadcastSession {
 		   fixup the session list at selection time by retrying the selection.
 		*/
 		if _, ok := bsm.sessMap[sess.OrchestratorInfo.Transcoder]; ok {
-			if monitor.Enabled && bsm.lastSess != nil && bsm.lastSess.OrchestratorInfo.Transcoder != sess.OrchestratorInfo.Transcoder {
-				monitor.OrchestratorSwapped()
+			if bsm.lastSess != nil && bsm.lastSess.OrchestratorInfo.Transcoder != sess.OrchestratorInfo.Transcoder {
+				glog.Warningf("Swapping BroadcastSession from orch=%v to orch=%v for manifestID=%s", bsm.lastSess.OrchestratorInfo.Transcoder, sess.OrchestratorInfo.Transcoder, bsm.mid)
```
yondonfu: Can we set the verbosity to debug here (I think this will require the use of `Infof()` instead)? There are generally a lot of logs already and this one seems like it'll only be needed by advanced users and/or users that are debugging node behavior.

server/broadcast.go
yondonfu: Maybe we can shorten the log line a bit as well since we just want to highlight that we're sticking with an existing O.

"No sessions in the selector for manifestID=%v re-using orch=%v with acceptable in-flight time"

server/broadcast.go
yondonfu: Omit "BroadcastSession" for brevity since I think we really just care about the URIs of the old and new orch?

server/broadcast.go
```diff
@@ -124,15 +125,22 @@ func (bsm *BroadcastSessionsManager) selectSession() *BroadcastSession {
...
+					monitor.OrchestratorSwapped()
+				}
 			}
 			bsm.lastSess = sess
 			return sess
 		}
 
 		// Last session got removed from map (possibly due to a failure) so stop tracking its in-flight segments
 		if bsm.lastSess != nil && sess.OrchestratorInfo.Transcoder == bsm.lastSess.OrchestratorInfo.Transcoder {
+			glog.Infof("Removing BroadcastSession of orch=%v, manifestID=%s, from the session list as it was removed from map", bsm.lastSess.OrchestratorInfo.Transcoder, bsm.mid)
```
yondonfu: Can we set the verbosity to debug here (I think this will require the use of Infof() instead)? There are generally a lot of logs already and this one seems like it'll only be needed by advanced users and/or users that are debugging node behavior.

server/broadcast.go
```diff
@@ -631,6 +653,8 @@ func transcodeSegment(cxn *rtmpConnection, seg *stream.HLSSegment, name string,
 		return nil, dlErr
 	}
 
+	cxn.sessManager.popSegInFlight(updatedSess)
```
yondonfu: Could it make sense to just move the `bsm.completeSession()` down here and move the seg in flight pop logic into that function to avoid having 2 call sites for `sel.Complete()` and to have a single "completion" procedure for a session that is run after downloading is done?

server/broadcast.go
```diff
@@ -227,6 +234,27 @@ func (bsm *BroadcastSessionsManager) refreshSessions() {
...
+			startTime: time.Now(),
+			segDur:    time.Duration(seg.Duration * float64(time.Second)),
+		})
+}
+
+func (bsm *BroadcastSessionsManager) popSegInFlight(sess *BroadcastSession) {
+	bsm.sessLock.Lock()
+	defer bsm.sessLock.Unlock()
+	if len(sess.SegsInFlight) == 1 {
+		sess.SegsInFlight = nil
```
yondonfu: It should be safe to pop from the front of the slice using the `SegsInFlight[1:]` syntax as long as the slice is not empty. So, this could be simplified to:

```golang
if len(sess.SegsInFlight) > 0 {
    sess.SegsInFlight = sess.SegsInFlight[1:]
}
if len(sess.SegsInFlight) == 0 {
    bsm.sel.Complete(sess)
}
```

server/broadcast.go
yondonfu: Omit BroadcastSession? Ex.

"Removing orch=%v from manifestID=%s session list"?

yondonfu (dismissed):

server/broadcast.go
jailuthra: Setting the slice explicitly to `nil` here because that would free the underlying array. https://yourbasic.org/golang/clear-slice/

Otherwise the underlying array for SegsInFlight will keep getting extended, and slice pointer moved, but will only get cleared after a stream gets over, and might lead to memory leak for extremely long streams.

server/broadcast.go
jailuthra: Good catch, fixed in the update!

iameli (approved):
LGTM!

Issue #1603, opened 2020-08-13
Orchestrator does not ensure that ticket face value covers estimated tx cost


yondonfu:
**Describe the bug**

At the moment, it is possible for an orchestrator to accept tickets with a face value that is <= the estimated tx cost of redemption ([example](https://etherscan.io/tx/0xd05bf9676006101c15249494d5a1c83a16d3e8751ae589291d64cdd370b36314)). As of 0.5.10, an orchestrator will not redeem these tickets because it will only redeem a winning ticket if its face value is > the estimated tx cost of redemption. However, the orchestrator should not be accepting tickets with a face value that is <= the estimated tx cost of redemption in the first place.

The cause of this bug is the lack of a check ([relevant code](https://github.com/livepeer/go-livepeer/blob/master/pm/recipient.go#L239)) if a broadcaster's max float is <= the estimated tx cost. In this case, when the default ticket face value (which is based on a target tx cost overhead of 1%) exceeds the broadcaster's max float, the ticket face value is set to the max float meaning that the ticket face value is also <= the estimated tx cost. 

**Expected behavior**

The orchestrator should not accept tickets from broadcasters if their max float is <= the estimated tx cost of redemption. Note the <= instead of < in the previous sentence - since the estimated tx cost of redemption does not include other small costs such as withdrawal or claiming, we use <= to indicate that the max float must be strictly greater than the estimated tx cost.

**Desktop (please complete the following information):**
 - Version [e.g. 22]: 0.5.10

chrishobcroft:
Maybe an elegant solution to this could be to give an O a way to set `-minimimFaceValue` for tickets they wish to receive?

yondonfu:
Closed by #1751

PR #1751, opened 2021-02-01
pm: tx cost checks

yondonfu:

This PR adds checks on the orchestrator to:

- Ensure that the estimated ticket redemption tx cost can be covered by the advertised ticket face value. At the moment, when the default ticket face value is higher than the broadcaster's max float (derived from its reserve), the ticket face value is adjusted to match the broadcaster's max float. A check is added to require that the broadcaster's max float >= the estimated ticket redemption tx cost
- Ensure that tickets are only redeemed if their face value would cover the estimated ticket redemption tx cost

This PR focuses on adding the above missing checks and leaves further configuration to be addressed separately.

See #1603 for more details.

See commit history.

Added unit tests.

**Checklist:**

### DIFF
--- a/pm/recipient.go
+++ b/pm/recipient.go
@@ -242,8 +242,8 @@ func (r *recipient) faceValue(sender ethcommon.Address) (*big.Int, error) {
 	}
 
 	if faceValue.Cmp(maxFloat) > 0 {
-		if maxFloat.Cmp(r.cfg.EV) < 0 {
-			// If maxFloat < EV, then there is no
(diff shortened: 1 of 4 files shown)


darkdarkdragon (approved):
LGTM

jailuthra (approved):
LGTM as well!

(although I do not fully understand the details of ticket redemption, but the changes and tests seem to match the solution described in https://github.com/livepeer/go-livepeer/issues/1603)

Issue #1758, opened 2021-02-08
HTTP listen and serve error getting swallowed


BisonAl:
**Describe the bug**
If there's ever an error returned to https://github.com/livepeer/go-livepeer/blob/master/server/webserver.go#L73 it's not reported. If you check out https://golang.org/pkg/net/http/#ListenAndServe, you'll see that `ListenAndServe` could return an error, which we could in turn be logging here.

We've ran into a scenario today where we're seeing the webserver close, however we're having a difficult time debugging because no error gets bubbled up or logged. 

We'll see this webserver start at https://github.com/livepeer/go-livepeer/blob/master/cmd/livepeer/livepeer.go#L914-L917, and know that `wc` channel is eventually getting closed at https://github.com/livepeer/go-livepeer/blob/master/cmd/livepeer/livepeer.go#L988-L990 because the following log gets thrown:

```
livepeer.go:989] CLI webserver shut down
```

**Expected behavior**
If the http webserver crashes, we can can see the relevant error in outputted logs.

**Desktop:**
Running from alpine linux, `v0.5.14`

**Additional context**
Should be a straight forward fix, editing https://github.com/livepeer/go-livepeer/blob/master/server/webserver.go#L73 from:

```
srv.ListenAndServe()
```

to

```
err := srv.ListenAndServe()
glog.Error(err)
```

If you wanted to get even fancier I think you could one-line it as:
```
glogError(srv.ListenAndServe())
```
but I find this hard to read personally^

yondonfu:
Closed by #1759

PR #1759, opened 2021-02-10
Capture and log error on webserver crash

AlexMapley:
I wrote up some more info on this problem in https://github.com/livepeer/go-livepeer/issues/1758.

We're running into some issues in `v0.5.14` with the webserver silently crashing, and figured having the underlying error logged would help a lot with usability.

- Captures the error returned from https://github.com/livepeer/go-livepeer/blob/master/server/webserver.go#L73
- Logs this error using `glog.Error()`

I didn't test these, the change is small enough I assume it should work alright.
Up to the livepeer devs if they think this is reasonable, or if I should do some more testing myself.

Here's an example 20 lines down where we use this same pattern: https://github.com/livepeer/go-livepeer/blob/master/server/webserver.go#L92-L93

https://github.com/livepeer/go-livepeer/issues/1758

**Checklist:**

I skipped over these as well^, am so used to running livepeer in alpine linux I'll admit I'm not too versed manually building for this binary for OSX.
I can follow through with these steps if we'd like, just let me know!

Didn't see any relevant documentation that we'd want to update for this.

### DIFF
--- a/server/webserver.go
+++ b/server/webserver.go
@@ -70,7 +70,8 @@ func (s *LivepeerServer) StartCliWebserver(bindAddr string) {
 	}
 
 	glog.Info("CLI server listening on ", bindAddr)
-	srv.ListenAndServe()
+	err := srv.ListenAndServe()
+	glog.Error(err)
 }
 
 func (s *LivepeerServer) setOnChainConfig() {}


AlexMapley:
cc @yondonfu, following up from discord. 

If you want me to test this branch out a bit more that's okay, just lmk.

yondonfu (approved):
LGTM!

Thanks @AlexMapley!

BisonAl:
Wooooo thanks @yondonfu!

Hoping to see the livepeer team soon once lockdown ends

PR #1761, opened 2021-02-11
Orchestrator router

yondonfu:

This PR adds a new buildable binary `livepeer_router`. The binary runs a gRPC server that accepts GetOrchestrator requests from Bs and forwards OrchestratorInfo responses from Os. A router is configured with a set of Os and is responsible for routing a requesting B to one of the Os by passing the OrchestratorInfo response from that O to the B. The B will then submit segments to the URI indicated in the `Transcoder` field of the OrchestratorInfo response. If any Os go offline, are at capacity or are have network transit issues, the router can route a different O to the B. This router should also help mainnet node operators start running multiple Os with a single on-chain registered ETH address instead of running each O with a different on-chain registered ETH address.

At the moment, the router sends a received GetOrchestrator request to all of its Os and forwards the first OrchestratorInfo response back to the B. As a result, a requesting B will be routed to the O that is closest to the router. A more sophisticated routing algorithm can be introduced separately. For example, an improved routing algorithm might involve doing a geoip lookup on the Os and the requesting B and routing the B to the O with the smallest [distance](https://en.wikipedia.org/wiki/Haversine_formula) based on lat/lng coordinates.

I chose to include this router functionality in a separate binary because the `livepeer` binary already has a lot of configuration options and I was hesitant to introduce another "node type" there especially since this is a relatively early concept. Plus this allows for some experimentation with this functionality without having to incorporate it into the code for the `livepeer` binary just quite yet. If there is a desire to move this functionality into the `livepeer` binary I think we can consider doing so later. 

See commit history.

Tested manually.

**Checklist:**

### DIFF
--- a/cmd/livepeer_router/livepeer_router.go
+++ b/cmd/livepeer_router/livepeer_router.go
@@ -1,0 +1,108 @@
+package main
+
+import (
+	"flag"
+	"net/url"
+	"os"
+	"os/signal"
+	"os/user"
+	"path/filepath"
+	"strings"
+
+	"github.com/golang/glog"
+	"github.com/livepeer/go-livepeer/server"
+)
+
+func defaultAddr(addr, defaultHost, defaultPort string) string {
+	if addr == "" {
+		return defaultHost + ":" + defaultPort
+	}
+	if addr[0] == ':' {
+		return defaultHost + addr
+	}
+	// not IPv6 safe
+	if !strings.Contains(addr, ":") {
+		return addr + ":" + defaultPort
+	}
+	return addr
+}
+
+func main() {
+	flag.Set("logtostderr", "true")
+	flag.CommandLine = flag.NewFlagSet(os.Args[0], flag.ExitOnError)
+
+	datadir := flag.String("datadir", "", "Directory that data is stored in")
+	httpAddr := flag.String("httpAddr", "", "Address (IP:port) to bind to for HTTP")
+	serviceAddr := flag.String("serviceAddr", "", "Publicly accessible URI (IP:port or hostname) to receive requests at")
+	orchAddr := flag.String("orchAddr", "", "Comma delimited list of orchestrator URIs (IP:port or hostname) to use")
+
+	flag.Parse()
+
+	usr, err := user.Current()
+	if err != nil {
+		glog.Fatalf("Cannot find current user: %v", err)
+	}
+
+	if *datadir == "" {
+		homedir := os.Getenv("HOME")
+		if homedir == "" {
+			homedir = usr.HomeDir
+		}
+		*datadir = filepath.Join(homedir, ".lpRouterData")
+	}
+
+	if _, err := os.Stat(*datadir); os.IsNotExist(err) {
+		glog.Infof("Creating datadir: %v", *datadir)
+		if err = os.MkdirAll(*datadir, 0755); err != nil {
+			glog.Fatalf("Error creating datadir: %v", err)
+		}
+	}
+
+	if *serviceAddr == "" {
+		glog.Fatal("Missing -serviceAddr")
+	}
+
+	serviceURI, err := url.ParseRequestURI("https://" + *serviceAddr)
+	if err != nil {
+		glog.Fatalf("Could not parse -serviceAddr: %v", err)
+	}
+
+	*httpAddr = defaultAddr(*httpAddr, "0.0.0.0", serviceURI.Port())
+	uri, err := url.ParseRequestURI("https://" + *httpAddr)
+	if err != nil {
+		glog.Fatalf("Could not parse -httpAddr: %v", err)
+	}
+
+	var uris []*url.URL
+	if len(*orchAddr) > 0 {
+		for _, addr := range strings.Split(*orchAddr, ",") {
+			addr = strings.TrimSpace(addr)
+			if !strings.HasPrefix(addr, "http") {
+				addr = "https://" + addr
+			}
+			uri, err := url.ParseRequestURI(addr)
+			if err != nil {
+				glog.Fatalf("Could not parse orchestrator URI: %v", err)
+			}
+			uris = append(uris, uri)
+		}
+	}
+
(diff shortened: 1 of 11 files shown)


docker/Dockerfile.build-linux
```diff
@@ -18,6 +18,10 @@ RUN apt-get update \
 RUN update-alternatives --install /usr/bin/clang++ clang++ /usr/bin/clang++-8 30 \
   && update-alternatives --install /usr/bin/clang clang /usr/bin/clang-8 30
 
+RUN GRPC_HEALTH_PROBE_VERSION=v0.3.6 && \
```
yondonfu, line 21: This CLI tool can be used in k8s environments for liveness and startup probes instead of vanilla HTTP probes which won't work right out of the box for a gRPC server. See livepeer/livepeer-infra#416.

server/router.go
```diff
@@ -0,0 +1,151 @@
...
+	creds, err := credentials.NewServerTLSFromFile(certFile, keyFile)
+	if err != nil {
+		return err
+	}
+
+	s := grpc.NewServer(grpc.Creds(creds))
+	r.srv = s
+
+	net.RegisterOrchestratorServer(s, r)
+	grpc_health_v1.RegisterHealthServer(s, health.NewServer())
```
yondonfu, line 53: This is required for the gRPC server to implement the [gRPC health checking protocol]()https://github.com/grpc/grpc/blob/master/doc/health-checking.md) which allows us to use tools like grpc_health_probe to check the health of the server.

server/router.go
```diff
@@ -0,0 +1,151 @@
...
+
+	_, err = client.Ping(ctx, &net.PingPong{Value: []byte{}})
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func getOrchestratorInfo(ctx context.Context, uris []*url.URL, req *net.OrchestratorRequest) (*net.OrchestratorInfo, error) {
```
yondonfu, line 102: This function is pretty similar to some of the code in the `discovery` package so an initial thought was to use the code there that implements the `common.OrchestratorPool` interface. But, there are some additional dependencies with that approach such as a broadcaster interface, capabilities interface and suspender interface which are not requirements here (yet). So, for now, I chose to just add this relatively small piece of code for now.

cmd/livepeer_router/livepeer_router.go
```diff
@@ -0,0 +1,108 @@
...
+	"os/signal"
+	"os/user"
+	"path/filepath"
+	"strings"
+
+	"github.com/golang/glog"
+	"github.com/livepeer/go-livepeer/server"
+)
+
+func defaultAddr(addr, defaultHost, defaultPort string) string {
```
kyriediculous, line 16: I'm okay with duplicating this, alternatively could move it to `common.DefaultAddr` as it's also used in `cmd/livepeer.go`

cmd/livepeer_router/livepeer_router.go
```diff
@@ -0,0 +1,108 @@
...
+		return addr + ":" + defaultPort
+	}
+	return addr
+}
+
+func main() {
+	flag.Set("logtostderr", "true")
+	flag.CommandLine = flag.NewFlagSet(os.Args[0], flag.ExitOnError)
+
+	datadir := flag.String("datadir", "", "Data directory")
```
kyriediculous: It seems like the usage of this flag is to be able to use a pre-signed certificate by a CA rather than generate a self-signed one. The naming and description however makes it a little opaque that this is the intended usage. 

Perhaps we could update the description to something like "Directory for SSL certificate and key, will be auto-generated if no certificate is available". 

Alternatively splitting up the flags into `-certificate`, `-certKey` with paths to the files rather than the directory or something along those lines (this might also help prevent overwriting other certificates if the user can bring its own filenames).

kyriediculous (changes requested):

cmd/livepeer_router/livepeer_router.go
yondonfu: Actually the `-datadir` flag here serves the same purpose as in `livepeer`. The flag is used to define a directory where all data for the binary should be stored. In the case of `livepeer_router`, the directory is just used to store the self-signed cert that is generated at startup (as is the case for `livepeer` as well), but it could be used to store other data as well if needed. I tried to make the description slightly clearer for `livepeer_router` in 990363c as well as the description for `livepeer` in ec7a640 to match. Open to other suggestions.

Neither `livepeer` nor `livepeer_router` support user defined certs right now because a self-signed cert is always generated at startup via the [getCert()](https://github.com/livepeer/go-livepeer/blob/master/server/cert.go#L82) function. I did not intent to include that feature in this PR, but it might be a useful feature to consider separately.

kyriediculous (approved):
LGTM after squashing the fixup commits 👍

server/router.go
```diff
@@ -0,0 +1,151 @@
...
+			defer conn.Close()
+
+			info, err := client.GetOrchestrator(cctx, req)
+			if err != nil {
+				errCh <- fmt.Errorf("%v err=%v", uri, err)
+				return
+			}
+
+			select {
+			case infoCh <- info:
```
darkdarkdragon, line 129: `infoCh` is unbuffered, so if two go-routines will arrive here one of them will stuck here forever.

darkdarkdragon (changes requested):
One small comment, otherwise looks good.

server/router.go
darkdarkdragon: Sorry, my mistake 😄

darkdarkdragon (approved):
LGTM

darkdarkdragon:
though CircleCI is failing with `cp: cannot stat './livepeer_router': No such file or directory`

yondonfu:
> though CircleCI is failing with `cp: cannot stat './livepeer_router': No such file or directory`

Ah looks like the intermediate Docker image was overwritten by another PR. Fixed by re-building to make sure the intermediate Docker image for CI contains the livepeer_router binary.

PR #1766, opened 2021-02-18
Flush recording JSON playlist after changing it

darkdarkdragon:
Flushed JSON playlist used in recording after it was modified.

Move around `FlushRecord` call.

Manually.

**Checklist:**


server/broadcast.go
```diff
@@ -584,6 +584,7 @@ func transcodeSegment(cxn *rtmpConnection, seg *stream.HLSSegment, name string,
 				now := time.Now()
 				uri, err := drivers.SaveRetried(bros, name, data, map[string]string{"duration": segDurMs}, 2)
 				took := time.Since(now)
+				recordWG.Done()
```
yondonfu: Is FlushRecord() supposed to be called after all the segments have been inserted into the JSON playlist? If so, should this Done() call be after the InsertHLSSegmentJSON() call below?

darkdarkdragon: Good catch, fixed.

yondonfu (approved):
LGTM

PR #1773, opened 2021-02-22
install_ffmpeg: download pre-configured nasm

iameli:
the autoconf step was failing on windows. this should sidestep that.
Added sha256sum so we're not just downloading some random archive.


yondonfu (approved):
LGTM!

PR #1775, opened 2021-02-24
Fix LB transcoder race condition around session-cleanup

### LINKED ISSUE #1750: LB transcoder race condition for segment that comes in really fast for the same session after a LPMS transcode error
From Discord:

IIUC the unexpected behavior here is the TranscoderStopped error for the second segment right? We'd expect Mist to retry the first segment and for B to continue sending segments after the no keyframe segment to the same O since B treats the no keyframe segment as non-retryable. But, since the next segment triggers a TranscoderStopped error from O which is not non-retryable, when Mist retries that segment B will send it to a different O.

The TranscoderStopped error is being returned after each no keyframe segment because the transcoder load balancer actually stops the underlying transcoder for a session if it returns an error [1][2][3]. I believe this results in the following:

- Whenever a no keyframe error is returned by the transcoder, the current session is automatically cleaned up
- If the next segment comes in really fast before the current session is cleaned up, a TranscoderStopped error is returned

Another realization: https://github.com/livepeer/go-livepeer/pull/1735/commits/882ed38fc1383dd8ae25e6abd20fd9e5a4a104c9 increased the # of in-flight segments supported on O, but there is another in-flight segment restriction via the channel buffer size on the transcoder load balancer. Since the channel buffer size on the transcoder load balancer is still 1, if O receives multiple in-flight segments TranscoderBusy would be returned https://github.com/livepeer/go-livepeer/blob/6fe41d3f29d6aff640251c1f25bd713ba8f4b8ac/core/lb.go#L202.

[1] Error check https://github.com/livepeer/go-livepeer/blob/6fe41d3f29d6aff640251c1f25bd713ba8f4b8ac/core/lb.go#L185
[2] Return of TranscoderStopped error https://github.com/livepeer/go-livepeer/blob/6fe41d3f29d6aff640251c1f25bd713ba8f4b8ac/core/lb.go#L156
[3] Session cleanup after transcoder loop exits https://github.com/livepeer/go-livepeer/blob/6fe41d3f29d6aff640251c1f25bd713ba8f4b8ac/core/lb.go#L107

jailuthra:

Ensure that no segments are left hanging since we increased the LB channel size in https://github.com/livepeer/go-livepeer/commit/882ed38fc1383dd8ae25e6abd20fd9e5a4a104c9

- https://github.com/livepeer/go-livepeer/commit/415c1c02b1f39525ab08d003c119b26fe6f6eede - Add a test case to check the case where Seg 1 fails, causing session loop to return, and Seg 2 comes in before the session channels are cleaned up https://github.com/livepeer/go-livepeer/issues/1750#issuecomment-773651383
- https://github.com/livepeer/go-livepeer/commit/8c6f5b43b11e05822f917b07813262180ef59800 - Fix the race condition by adding a separate `done` channel to signal the sender that the session loop has closed (and return `ErrTranscoderStopped` to the caller)
- https://github.com/livepeer/go-livepeer/pull/1775/commits/ffd4505117f5fdbc7f3a15fa9becfa92d25ec98e - Fix some old unit tests to use `maxSegmentChannels` and initialize the done channel

- Run the test case in https://github.com/livepeer/go-livepeer/commit/415c1c02b1f39525ab08d003c119b26fe6f6eede to reproduce hanging behaviour
- Run the test case again after the fix to verify hanging behaviour does not happen anymore

Fixes https://github.com/livepeer/go-livepeer/issues/1750

**Checklist:**

### DIFF
--- a/core/lb_test.go
+++ b/core/lb_test.go
@@ -69,6 +69,41 @@ func TestLB_Ratchet(t *testing.T) {
 	})
 }
 
+func TestLB_SessionCleanupRace(t *testing.T) {
+	// Reproduce race condition around session cleanup #1750
+
+	assert := assert.New(t)
+	lb := NewLoadBalancingTranscoder("0", newStubTranscoder).(*LoadBalancingTranscoder)
+	sess := "sess"
+	// Force create a new session
+	_, err := lb.Transcode(stubMetadata(sess, ffmpeg.P144p30fps16x9))
+	assert.Nil(err)
+	// Mark transcoder to error out
+	transcoder := lb.sessions[sess].transcoder.(*StubTranscoder)
+	transcoder.FailTranscode = true
+
+	// Send 2 jobs concurrently to trigger the stuck issue
+	wg := newWg(2)
+	errSignal := make(chan struct{})
+	// Error out on Job 1
+	go func() {
+		lb.mu.Lock() // lock the LB to prevent session cleanup
+		_, err = lb.sessions[sess].Transcode(stubMetadata(sess, ffmpeg.P144p30fps16x9))
+		assert.Equal(ErrTranscode, err)
+		errSignal <- struct{}{}
+		wg.Done()
+	}()
+	// Job 2 arrives when the session loop() has closed, but session isn't cleaned up yet
+	go func() {
+		<-errSignal
+		_, err = lb.sessions[sess].Transcode(stubMetadata(sess, ffmpeg.P144p30fps16x9))
+		assert.Equal(ErrTranscoderStopped, err)
+		wg.Done()
+	}()
+	assert.True(wgWait(wg))
+	lb.mu.Unlock() // unlock for cleanup
+}
+
 func TestLB_LoadAssignment(t *testing.T) {
 
 	// Property: Overall load only increases after first segment
--- a/core/lb_test.go
+++ b/core/lb_test.go
@@ -142,7 +142,8 @@ func TestLB_SessionCancel(t *testing.T) {
 
 	sess := &transcoderSession{
 		transcoder:  newStubTranscoder(""),
-		sender:      make(chan *transcoderParams, 1),
+		done:        make(chan struct{}),
+		sender:      make(chan *transcoderParams, maxSegmentChannels),
 		makeContext: ctxFunc,
 	}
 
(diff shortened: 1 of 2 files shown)


core/lb_test.go
```diff
@@ -69,6 +69,43 @@ func TestLB_Ratchet(t *testing.T) {
...
+	// Job 2 arrives when the session loop() has closed, but session isn't cleaned up yet
+	go func() {
+		sig := <-errSignal
+		if sig {
+			_, err = lb.sessions[sess].Transcode(stubMetadata(sess, ffmpeg.P144p30fps16x9))
+			assert.Equal(ErrTranscoderStopped, err)
+		}
+		wg.Done()
+	}()
+	wg.Wait()
```
yondonfu: At the moment, this test will hang upon failure. While this failure behavior will still tell us if the test breaks in the future it will also cause the entire CI process to grind to a halt. We could avoid that by updating the test to return upon a timeout. A few other tests use the [wgWait](https://github.com/livepeer/go-livepeer/blob/fd07a5673175aa6d10f56dde6eb3ef2a9f5691b9/core/orch_test.go#L204) test helper for these types of assertions. For example:

```golang
// Returns true if all WaitGroup tasks completed
// Returns false if not all WaitGroup tasks completed within the 1 second timeout
assert.True(wgWait(wg))
```

core/lb_test.go
```diff
@@ -69,6 +69,43 @@ func TestLB_Ratchet(t *testing.T) {
...
+	go func() {
+		lb.mu.Lock() // lock the LB to prevent session cleanup
+		_, err = lb.sessions[sess].Transcode(stubMetadata(sess, ffmpeg.P144p30fps16x9))
+		assert.Equal(ErrTranscode, err)
+		errSignal <- true
+		wg.Done()
+	}()
+	// Job 2 arrives when the session loop() has closed, but session isn't cleaned up yet
+	go func() {
+		sig := <-errSignal
```
yondonfu: Minor: Doesn't seem like a bool is required for the channel value so you can avoid some extra checks here if the channel used an empty struct instead.

For example:

```golang
sigCh := make(chan struct{})
go func() {
    ...
    sigCh <- struct{}
    wg.Done()
}()

go func() {
    <-sigCh
    ...
    wg.Done()
}()

File: core/lb_test.go [hunk 3]
```diff
@@ -107,7 +144,8 @@ func TestLB_SessionCancel(t *testing.T) {
 
 	sess := &transcoderSession{
 		transcoder:  newStubTranscoder(""),
-		sender:      make(chan *transcoderParams, 1),
+		done:        make(chan struct{}, 1),
```
yondonfu: Not that it matters in the context of this test, but this can be unbuffered as is the case in the actual LoadBalancingTranscoder impl right?

yondonfu (commented):
Looks pretty good. Just a few minor comments relating to the tests.

kyriediculous (commented):
Looks good to me pending Yondon's comments 👍

yondonfu (approved):
LGTM after squashing

PR #1778, opened 2021-03-01
cmd/livepeer_bench: live mode, better stats

### LINKED ISSUE #1699: Add live mode to livepeer_bench
A user can configure the tool to simulate transcoding a live stream by transcoding a segment only after it would appear in a live playlist based on the segment's duration.

For consideration: the option to interleave segments from concurrent sessions i.e. seg1 from sess1 comes in, seg1 from sess2 comes in, seg2 from sess1 comes in as we suspect utilization will be higher in this case.

EDIT: Interleaving segments i.e. by staggering the start of streams is out of scope for this particular issue, but could be considered separately.

jailuthra:

Add option to simulate live-mode in the benchmarking tool. Also, clean up the output statistics to show real-time segment ratio and stream duration ratio.

- Adds an option to enable live-mode, which would wait to transcode segments based on their duration in the input playlist.
- Improve output statistics

Manual testing with `./livepeer_bench -in ~/Videos/bbb/source.m3u8 -outPrefix /tmp/out -nvidia 0 -concurrentSessions 20 -transcodingOptions cmd/livepeer_bench/transcodingOptions.json -segs 20`

Fixes #1699 

**Checklist:**

### DIFF
--- a/cmd/livepeer_bench/livepeer_bench.go
+++ b/cmd/livepeer_bench/livepeer_bench.go
@@ -30,6 +30,7 @@ func main() {
 	flag.CommandLine = flag.NewFlagSet(os.Args[0], flag.ExitOnError)
 
 	in := flag.String("in", "", "Input m3u8 manifest file")
+	live := flag.Bool("live", true, "Simulate live stream")
 	concurrentSessions := flag.Int("concurrentSessions", 1, "# of concurrent transcode sessions")
 	segs := flag.Int("segs", 0, "Maximum # of segments to transcode (default all)")
 	transcodingOptions := flag.String("transcodingOptions", "P240p30fps16x9,P360p30fps16x9,P720p30fps16x9", "Transcoding options for broadcast job, or path to json config")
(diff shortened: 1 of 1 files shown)


cmd/livepeer_bench/livepeer_bench.go
```diff
@@ -30,6 +30,7 @@ func main() {
 	flag.CommandLine = flag.NewFlagSet(os.Args[0], flag.ExitOnError)
 
 	in := flag.String("in", "", "Input m3u8 manifest file")
+	liveMode := flag.Bool("liveMode", true, "Simulate live stream")
```
yondonfu: Minor: Maybe just "live" for brevity?

cmd/livepeer_bench/livepeer_bench.go
```diff
@@ -151,8 +163,22 @@ func main() {
...
 	}
 	wg.Wait()
-	fmt.Fprintf(os.Stderr, "Took %0.4v seconds to transcode %v segments of total duration %vs (%v concurrent sessions)\n",
-		time.Since(start).Seconds(), segCount, segTotalDur, *concurrentSessions)
+	statsTable := tablewriter.NewWriter(os.Stderr)
+	stats := [][]string{
+		{"Concurrent Sessions", fmt.Sprintf("%v", *concurrentSessions)},
+		{"Total Segs Transcoded", fmt.Sprintf("%v", segCount)},
+		{"Real-Time Segs Transcoded", fmt.Sprintf("%v", realTimeSegCount)},
+		{"* Real-Time Segs Ratio *", fmt.Sprintf("%0.4v", float64(realTimeSegCount)/float64(segCount))},
```
yondonfu, line 174: Curious why asterisks are added here? Anything particularly special about this stat relative to the other ones?

yondonfu, line 174: Minor: segCount and srcDur could be 0 if the user passes in a playlist with no segments. This would be a case of user error. Good practice to add a check for that to avoid a divide by zero panic though.

cmd/livepeer_bench/livepeer_bench.go
```diff
@@ -136,13 +138,23 @@ func main() {
...
+				mu.Lock()
+				transcodeDur += segTxDur
+				srcDur += v.Duration
+				segCount++
+				if segTxDur <= v.Duration {
+					realTimeSegCount += 1
+				}
+				mu.Unlock()
+				iterEnd := time.Now()
+				segDur := time.Duration(int64(v.Duration*1000)) * time.Millisecond
```
yondonfu: This can be shortened to:

```golang
segDur := time.Duration(v.Duration * float64(time.Second))
```

`time.Second` is stored in milliseconds so multiplying it by `v.Duration`, which is stored in seconds, will give us `v.Duration` in terms of milliseconds which can then be converted into a `time.Duration`.

cmd/livepeer_bench/livepeer_bench.go
jailuthra: Wanted to highlight this and `Real-Time Duration Ratio` in some manner - as a user I would personally be most interested in quickly eyeballing these two ratios. WDYT?

cmd/livepeer_bench/livepeer_bench.go
jailuthra: Neat thx!

cmd/livepeer_bench/livepeer_bench.go
yondonfu: Got it. Sounds reasonable.

yondonfu (approved):
LGTM after squashing!

PR #1781, opened 2021-03-02
doc: Reformat install doc

yondonfu:

Some formatting updates to make the install doc a bit more readable by:

- Adding a table of contents
- Distinguishing between install guides and build (i.e. Docker and/or from source) guides
- Moving longer code blocks onto their own line

### DIFF
--- a/doc/install.md
+++ b/doc/install.md
@@ -1,137 +1,132 @@
 # Installing Livepeer
 
-## Option 1: Download pre-built executables from Livepeer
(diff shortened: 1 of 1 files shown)


yondonfu:
Decided to put the updated install guide in the new docs https://github.com/livepeer/livepeer-org/pull/71 so this PR is no longer necessary. The current plan is to eventually use the install guide as the single source of truth.

PR #1782, opened 2021-03-02
Fix for multi-seg-in-flight

darkdarkdragon:
Fixes case where data in `SegsInFlight` array gets lost after session refresh

- Add unit test showing bug
- Add fix - after session update new session object should be assigned to `lastSess` field.

Unit test.

**Checklist:**

### DIFF
--- a/server/broadcast.go
+++ b/server/broadcast.go
@@ -249,6 +249,14 @@ func (bsm *BroadcastSessionsManager) pushSegInFlight(sess *BroadcastSession, seg
 		})
 }
 
+func (bsm *BroadcastSessionsManager) updateLastSession(oldSess, newSess *BroadcastSession) {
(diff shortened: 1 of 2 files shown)


darkdarkdragon:
@jailuthra can you take a look? I think in case of call to `refreshSession` data in `SegsInFlight` gets lost...

jailuthra (approved):
Sorry for the delayed response.
Good catch & LGTM!

EDIT: Before merge it would be amazing if we can update the pending Changelog file for v0.5.16 https://github.com/livepeer/go-livepeer/commit/eaf986be5a5214f8f71dcc254f100ebd2c564ec7

darkdarkdragon:
@jailuthra sorry I saw your comment about Changelog right after I've pressed 'merge' button

PR #1784, opened 2021-03-04
core/lb: use SessionID as map index

### LINKED ISSUE #1748: lb: LoadBalancingTranscoder should use AuthToken.SessionId instead of ManifestID to index into sessions
**Describe the bug**

The orchestrator uses [AuthToken.SessionId](https://github.com/livepeer/go-livepeer/blob/master/core/orchestrator.go#L457) to index into its session map, but the LoadBalancingTranscoder still uses the [ManifestID to index into sessions](https://github.com/livepeer/go-livepeer/blob/master/core/lb.go#L70). As a result, two different transcode sessions with the same manifestID can result in the use of the same load balancer session when different load balancer sessions should be used.

jailuthra:

Use SessionID instead of ManifestID to index the loadbalancer sessions.

- https://github.com/livepeer/go-livepeer/commit/a3862f17202b383ec4bb639f0019472a9c1f9b85 Use SessionID as map index

- Run the lb unit tests
- Verify that SessionId is now used within the log lines on a local O
```
I0305 01:40:55.807153   31849 orchestrator.go:474] Submitted segment to transcode loop manifestID=movmovv sessionID=a7a6b2b9 seqNo=1
I0305 01:40:55.807386   31849 lb.go:54] LB: Using existing transcode session for a7a6b2b9_0
```

Fixes #1748

**Checklist:**

### DIFF
--- a/core/lb.go
+++ b/core/lb.go
@@ -48,7 +48,7 @@ func NewLoadBalancingTranscoder(devices string, newTranscoderFn newTranscoderFn)
 func (lb *LoadBalancingTranscoder) Transcode(md *SegTranscodingMetadata) (*TranscodeData, error) {
 
 	lb.mu.RLock()
-	session, exists := lb.sessions[string(md.ManifestID)]
+	session, exists := lb.sessions[string(md.AuthToken.SessionId)]
 	lb.mu.RUnlock()
 	if exists {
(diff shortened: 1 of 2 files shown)


yondonfu (approved):
LGTM! :shipit:

PR #1785, opened 2021-03-04
windows: fully-static build, bump to go1.15

iameli:
* Bump Windows go build to go1.15
* Implement a fully-static livepeer.exe file with no more bundled DLLs. And it only took me two years.

It works!
image

Checked with both GPU and CPU transcoding. I didn't check on-chain mode but presumably anything that's mostly Go code is fine.


yondonfu (approved):
Woot! LGTM :shipit:

PR #1786, opened 2021-03-04
cmd/livepeer_bench: Support delay

yondonfu:

- 
- 
- 

**Checklist:**


jailuthra:
Nice. I see this PR is still in draft - but I think we can combine the pre-existing 300ms sleep timer at the end of the loop https://github.com/livepeer/go-livepeer/blob/8acb472878eecb292c90fb07129121890e6b5319/cmd/livepeer_bench/livepeer_bench.go#L169 with this new delay field (and probably set the default value for the field to 300ms)

yondonfu:
I opened this draft PR before I realized there is already a hardcoded [delay](https://docs.google.com/document/d/1CB3J4-yeHdyI2l2SLXhYPUYIKlKshe1Do4lFVpfR2l0/edit). I don't plan on following up here to parameterize that hardcoded value anytime soon, so I'm gonna close this PR for now. If there is a need to parameterize later on it should be pretty straightforward.

PR #1790, opened 2021-03-11
Add changelog process

yondonfu:

Introduce a changelog process described in `doc/contributing.md`. The motivation behind adding this changelog process is to make it easier to add the list of changes to release notes whenever we cut a release - if we update the changelog for each PR then it'll already be ready come release time and that makes it easier for the release coordinator!

Most of this process is based off of the process used in [tendermint](https://github.com/tendermint/tendermint/blob/master/CONTRIBUTING.md#changelog) with a few simplifications to make the transition to this process a bit easier (i.e. a few less rules).

See commit history.

N/A

N/A

**Checklist:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,1 @@
+# Changelog
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -21,6 +21,7 @@
 - \#1778 Add live mode to `livepeer_bench` and expose additional metrics (@jailuthra)
 - \#1785 Update the Windows build to be fully static and to use go1.15 (@iameli)
 - \#1727 Add a `-maxGasPrice` flag to set the maximum gas price to use for transactions (@kyriediculous)
+- \#1790 Add changelog process
 
 #### Broadcaster
 
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -1,0 +1,49 @@
+# Unreleased Changes
+
+## vX.X
+
+### Upcoming Changes
+
+- The following flags are pending deprecation and will be removed in the next release:
+    - `-gasPrice`
+    - `-s3bucket`
+    - `-s3creds`
+    - `-gsbucket`
+    - `-gskey`
+
+### Features ⚒
+
+#### General
+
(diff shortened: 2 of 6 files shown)


linkify_changelog.py
```diff
@@ -0,0 +1,14 @@
+import fileinput
```
kyriediculous: II don't  really like randomly throwing python into this repo, now it is a dependency required for contributing to this repo while it's a go repo. I don't think that's great developer experience. 

Seems trivial to rewrite this in go copy-pasting the regex.

Can write the results back to file as well of course rather than printing them. 

```go
package main

import (
	"bufio"
	"fmt"
	"log"
	"os"
	"regexp"
)

func main() {
	file, err := os.Open("./changelog.md")
	if err != nil {
		log.Fatal(err)
	}
	defer file.Close()

	m := regexp.MustCompile(`#([0-9]*)`)

	scanner := bufio.NewScanner(file)
	for scanner.Scan() {
		line := m.ReplaceAllString(scanner.Text(), `[#$1](https://github.com/livepeer/go-livepeer/pull/$1)`)
		fmt.Println(line)
	}

	if err := scanner.Err(); err != nil {
		log.Fatal(err)
	}
}
```

Input
```
#1 hello world (@kyriediculous)
#2 hello mars (@yondonfu)
```

Output
```
$ go run changelog-link.go
[#1](https://github.com/livepeer/go-livepeer/pull/1) hello world (@kyriediculous)
[#2](https://github.com/livepeer/go-livepeer/pull/2) hello mars (@yondonfu)
```

kyriediculous (changes requested):

linkify_changelog.py
yondonfu: Good call on sticking with Go for this script to limit the number of language dependencies for the repo. Converted the script to Go in 876d96f and updated the linkify command in `doc/releases.md` 5d41439.

> now it is a dependency required for contributing to this repo while it's a go repo. I don't think that's great developer experience.

Just to add a bit of clarification to this point - most contributors wouldn't be expected to have to run this script and the only person that would be expected to run this script (once when a release is being created) is the release coordinator.

kyriediculous (approved):
LGTM after squashing 👍

PR #1795, opened 2021-03-12
install_ffmpeg.sh: add -O2 to gnutls build

### LINKED ISSUE #1788: ci: Darwin builds broken
Earliest failing build https://travis-ci.org/github/livepeer/go-livepeer/builds/762065326

`ERROR: gnutls not found using pkg-config`

Seems like something from the Windows build PR https://github.com/livepeer/go-livepeer/pull/1785 might be the cause?

Also we probably missed this because the Travis build wasn't showing up in the PR UI...Should fix that too.

iameli:
Apparently `-O2` optimizations change the macros around enough that the Darwin bug doesn't exist anymore

fixes #1788


yondonfu (approved):
LGTM! Thanks for testing this alternative fix. Not sure what the implications are of changing the optimization level, but I guess we'll see...

PR #1809, opened 2021-03-23
db: return Miniheader with zero values rather than nil from FindLatestMiniheader when no headers are found

### LINKED ISSUE #1780: blockwatcher: Remove extraneous log statements when there are no blocks to catch up on
```
Mar 01 15:32:42 beefcake livepeer[17450]: I0301 15:32:42.745240   17450 block_watcher.go:318] Backfilling block events (this can take a while)...
Mar 01 15:32:42 beefcake livepeer[17450]: I0301 15:32:42.745252   17450 block_watcher.go:319] Start block: 11953162                 End block: 11953161                Blocks elapsed: -1
Mar 01 15:32:42 beefcake livepeer[17450]: I0301 15:32:42.747350   17450 block_watcher.go:537] fetching block logs from=11953162 to=11953161
```

If the endBlock is before the startBlock indicating there are no blocks to catch up on, the extra log statement can be avoided to avoid user confusion.

### LINKED ISSUE #1803: nil pointer in block watcher syncToLatestBlock
Observed this error with an O connected to Rinkeby. There is probably a missing nil check somewhere in the block watcher code:

```
[signal SIGSEGV: segmentation violation code=0x1 addr=0x40 pc=0xb30609]

goroutine 47 [running]:
github.com/livepeer/go-livepeer/eth/blockwatch.(*Watcher).syncToLatestBlock(0xc000390360, 0x0, 0x0)
        /build/eth/blockwatch/block_watcher.go:162 +0xe9
github.com/livepeer/go-livepeer/eth/blockwatch.(*Watcher).Watch(0xc000390360, 0x1fec980, 0xc0009e2600, 0x1fcf4c0, 0xc000428300)
        /build/eth/blockwatch/block_watcher.go:123 +0x165
main.main.func6(0xc000390360, 0x1fec980, 0xc0009e2600, 0xc0009dcba0)
        /build/cmd/livepeer/livepeer.go:711 +0x45
created by main.main
        /build/cmd/livepeer/livepeer.go:710 +0x411f
```

kyriediculous:
Instead of returning `nil` from `db.FindLatestMiniHeader` when no entries are found in the database we return a zero-value to avoid unexpected behaviour whereby no error and no value is returned.  

This behaviour is similar to the underlying `ethClient.HeaderByNumber` implementation. 

**note**
Consider changing `blockWatch.Miniheader.Number` to an `int64` or `big.Int` instead of a pointer reference `*big.Int` to avoid explicit initialisation of this field. 

```
&blockwatch.MiniHeader{
  Number: big.NewInt(0)
}
```

- Added a test case
- Return zero value when no results found in FindLatestMiniHeader
- fixed other test cases for functions that use FindLatestMiniHeader
- Changed the check for `blocksElapsed` to be `<=0` to avoid verbose logging when no blocks have elapsed. 

Ran unit tests

Fixes #1803 
Fixes #1780 

**Checklist:**

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -8,6 +8,8 @@
 
 - \#1810 Display "n/a" in CLI when max gas price isn't specified (@kyriediculous)
 - \#1827 Limit the maximum size of a segment read over HTTP (@jailuthra)
+- \#1809 Don't log statement that blocks have been backfilled when no blocks have elapsed
+- \#1809 Avoid nil pointer error in SyncToLatestBlock when no blocks are present in the database (@kyriediculous)
 
 #### Orchestrator
 
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -8,7 +8,7 @@
 
 - \#1810 Display "n/a" in CLI when max gas price isn't specified (@kyriediculous)
 - \#1827 Limit the maximum size of a segment read over HTTP (@jailuthra)
-- \#1809 Don't log statement that blocks have been backfilled when no blocks have elapsed
+- \#1809 Don't log statement that blocks have been backfilled when no blocks have elapsed (@kyriediculous)
 - \#1809 Avoid nil pointer error in SyncToLatestBlock when no blocks are present in the database (@kyriediculous)
 
 #### Orchestrator
--- a/eth/blockwatch/block_watcher.go
+++ b/eth/blockwatch/block_watcher.go
@@ -159,6 +159,10 @@ func (w *Watcher) syncToLatestBlock() error {
 		return err
 	}
 
+	if lastSeenHeader == nil {
+		return w.pollNextBlock()
+	}
+
 	for i := lastSeenHeader.Number; i.Cmp(newestHeader.Number) < 0; i = i.Add(i, big.NewInt(1)) {
 		if err := w.pollNextBlock(); err != nil {
 			return err
@@ -332,7 +336,7 @@ func (w *Watcher) getMissedEventsToBackfill(ctx context.Context) ([]*Event, erro
 		return events, nil
 	}
 
-	if blocksElapsed = latestBlockNum - startBlockNum; blocksElapsed == 0 {
+	if blocksElapsed = latestBlockNum - startBlockNum; blocksElapsed <= 0 {
 		return events, nil
 	}
 


CHANGELOG_PENDING.md
```diff
@@ -1,3 +1,9 @@
 # Unreleased Changes
 
 ## vX.X
+
+### Bug Fixes 🐞
+
+#### General
+
+- [#1809](https://github.com/livepeer/go-livepeer/pull/1809) Avoid nil pointer error in SyncToLatestBlock when no blocks are present in the database (@kyriediculous)
```
yondonfu: Let's update this based on format description in the [contributing](https://github.com/livepeer/go-livepeer/blob/master/doc/contributing.md#changelog) guide.

yondonfu (changes requested):
Looks like there are multiple places in `block_watcher.go` that depend on FindLatestMiniHeader() returning a nil header when there are no headers in the DB. Here is [one](https://github.com/livepeer/go-livepeer/blob/5784f8cea0a6033749a1e35af72ed1539e3fef87/eth/blockwatch/block_watcher.go#L194) and here is [another](https://github.com/livepeer/go-livepeer/blob/5784f8cea0a6033749a1e35af72ed1539e3fef87/eth/blockwatch/block_watcher.go#L158). There may be more. Currently, I think returning the zero value for a header in FindLatestMiniHeader() might cause breakage in those areas.

Could be worth weighing continuing to return a nil header and just adding a check for that in the code path that was causing a nil pointer error.

kyriediculous:
Reverted to old behaviour and added a nil check. Also added a quick one character fix for #1780. 

Tested on rinkeby and checked if there were other cases we should add a nil check but didn't see any immeadiatly. Will double-check.

eth/blockwatch/block_watcher.go
```diff
@@ -159,6 +159,12 @@ func (w *Watcher) syncToLatestBlock() error {
 		return err
 	}
 
+	if lastSeenHeader == nil {
+		lastSeenHeader = &MiniHeader{
```
yondonfu: Looks like this would result in `newestHeader.Number` calls to `pollNextBlock()` when `lastSeenHeader` is nil i.e. the number of blocks since genesis.

If `lastSeenHeader` is nil we would only want to call `pollNextBlock()` up to `newestHeader.Number - w.startBlockDepth` times (`pollNextBlock()` also starts from `w.startBlockDepth` if the latest header from the DB is nil).

yondonfu (changes requested):

eth/blockwatch/block_watcher.go
kyriediculous: Fixed in 0e11f59

eth/blockwatch/block_watcher.go
```diff
@@ -159,7 +159,12 @@ func (w *Watcher) syncToLatestBlock() error {
 		return err
 	}
 
-	for i := lastSeenHeader.Number; i.Cmp(newestHeader.Number) < 0; i = i.Add(i, big.NewInt(1)) {
+	lastSeenHeaderNumber := big.NewInt(w.startBlockDepth.Int64())
```
yondonfu: On second thought given the current configuration of the block watcher I don't think iterating starting from startBlockDepth leads to the correct desired behavior.

First, some acknowledgements. Typically we would not expect to set lastSeenHeaderNumber to startBlockDepth because syncToLatestBlock() is called in the watch loop started by Watch() which is always called after backfilling and backfilling would ensure that there is a block stored in the DB. 

*If* we happen to get to this case for some reason, then the current code would set lastSeenHeaderNumber to  startBlockDepth which is set to [rpc.LatestBlockNumber](https://github.com/livepeer/go-livepeer/blob/4e24711bfabc873f1be7e98cdd8599f68a6cfedd/cmd/livepeer/livepeer.go#L436). Since `rpc.LatestBlockNumber` is an [alias for -1](https://github.com/ethereum/go-ethereum/blob/053ed9cc847647a9b3ef707d0efe7104c4ab2a4c/rpc/types.go#L73) the loop on L167 would start at -1 and end at newestHeader.Number which is not the behavior we want.

A startBlockDepth of `rpc.LatestBlockNumber` means that we just want to fetch the latest block since passing `rpc.LatestBlockNumber` to the `eth_*` RPC calls will return data on the latest block. So, I think the correct behavior here would be something like:

```golang
if lastSeenHeader == nil {
    return w.pollNextBlock() 
}

for i := lastSeenHeader.Number; i.Cmp(newestHeader.Number) < 0; i = i.Add(i, big.NewInt(1)) { ... }
```

`pollNextBlock()` should know to fetch the latest block [when there are no blocks in the DB](https://github.com/livepeer/go-livepeer/blob/b909883c2535a434dfc4cd5078d8c249df9a1f18/eth/blockwatch/block_watcher.go#L181).

yondonfu (changes requested):

yondonfu (approved):
LGTM after squashing

CHANGELOG_PENDING.md
```diff
@@ -6,8 +6,11 @@
 
 #### General
 
+<<<<<<< HEAD
```
yondonfu: Looks like this this conflict needs to be resolved.

yondonfu (changes requested):

CHANGELOG_PENDING.md
```diff
@@ -8,6 +8,8 @@
 
 - \#1810 Display "n/a" in CLI when max gas price isn't specified (@kyriediculous)
 - \#1827 Limit the maximum size of a segment read over HTTP (@jailuthra)
+- \#1809 Don't log statement that blocks have been backfilled when no blocks have elapsed
```
yondonfu: Missing author here.

yondonfu (approved):
One comment on changelog, but LGTM after fixing.

kyriediculous:
This one okay to merge ?

yondonfu:
Yep and already approved.

PR #1811, opened 2021-03-24
changelog: Update for #1782

jailuthra:
- [x] [Pending changelog](./CHANGELOG_PENDING.md) updated for #1782

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -1,3 +1,25 @@
 # Unreleased Changes
 
 ## vX.X
+
+### Upcoming Changes
+
+### Features ⚒
+
+#### General
+
+#### Broadcaster
+
+#### Orchestrator
+
+### Bug Fixes 🐞
+
+#### General
+
+#### Broadcaster
+
+- \#1782 Fix SegsInFlight data-loss on refreshing O sessions (@darkdragon)
+
+#### Transcoder
+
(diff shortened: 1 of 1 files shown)


CHANGELOG_PENDING.md
```diff
@@ -1,3 +1,25 @@
 # Unreleased Changes
 
-## vX.X
+## v0.5.16
```
yondonfu: Fine to leave this as vX.X as we'll add the release version to `CHANGELOG` when the release is cut.

CHANGELOG_PENDING.md
```diff
@@ -1,3 +1,25 @@
...
+
+#### General
+
+#### Broadcaster
+
+#### Orchestrator
+
+### Bug Fixes 🐞
+
+- [#1782](https://github.com/livepeer/go-livepeer/pull/1782) Fix SegsInFlight data-loss on refreshing O sessions (@darkdragon)
```
yondonfu: Fine to not add the PR link here and instead use the [format](https://github.com/livepeer/go-livepeer/blob/master/doc/contributing.md#changelog) described in the changelog portion of the contributing guide because it will be automatically added when we cut a release. So, let's just use the format described there.

yondonfu (commented):
Thanks for submitting this update! Just a few small comments about the new changelog process.

yondonfu (approved):
LGTM

PR #1813, opened 2021-03-25
core,discovery: add checks for PriceInfo.pixelsPerUnit==0

kyriediculous:
Using `common.RatPrinceInfo` for converting `net.PrinceInfo` -> `big.Rat` and handling the error case. 

This function already does a check for the `PriceInfo` denominator being 0 

- Using common.RatPriceInfo 
- Add unit tests

Didn't explicitly test the predicate function, would require us to move it to global scope I believe. Can always do this still. 

**Checklist:**

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -16,6 +16,8 @@
 
 #### General
 
+- \#1813 Check that priceInfo.pixelsPerUnit is not 0 (@kyriediculous)
+
 #### Broadcaster
 
 - \#1782 Fix SegsInFlight data-loss on refreshing O sessions (@darkdragon)
--- a/discovery/db_discovery.go
+++ b/discovery/db_discovery.go
@@ -109,12 +109,12 @@ func (dbo *DBOrchestratorPoolCache) GetOrchestrators(numOrchestrators int, suspe
 		// check if O's price is below B's max price
 		maxPrice := server.BroadcastCfg.MaxPrice()
 		price, err := common.RatPriceInfo(info.PriceInfo)
-		if price == nil {
-			glog.V(common.DEBUG).Infof("no price info received for orch=%v", info.GetTranscoder())
-			return false
-		}
 		if err != nil {
 			glog.V(common.DEBUG).Infof("invalid price info orch=%v err=%v", info.GetTranscoder(), err)
+			return false
+		}
+		if price == nil {
+			glog.V(common.DEBUG).Infof("no price info received for orch=%v", info.GetTranscoder())
 			return false
 		}
 		if maxPrice != nil && price.Cmp(maxPrice) > 0 {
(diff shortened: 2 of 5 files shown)


CHANGELOG_PENDING.md
```diff
@@ -16,10 +16,14 @@
 
 #### General
 
+- [#1813](https://github.com/livepeer/go-livepeer/pull/1813) Check that priceInfo.pixelsPerUnit is not 0
```
yondonfu: We don't need the link here since it will be automatically added when a release is cut so let's use the format described in the [contribution](https://github.com/livepeer/go-livepeer/blob/master/doc/contributing.md#changelog) guide which will allow that link to be automatically added later on.

CHANGELOG_PENDING.md
```diff
@@ -16,10 +16,14 @@
...
+
 #### Broadcaster
 
 - \#1782 Fix SegsInFlight data-loss on refreshing O sessions (@darkdragon)
 
 #### Transcoder
 
 [Full list of changes](https://github.com/livepeer/go-livepeer/compare/v0.5.15...v0.5.16)
+
+https://github.com/livepeer/go-livepeer/pull/1813
```
yondonfu: Can be removed.

discovery/db_discovery.go
```diff
@@ -108,9 +108,13 @@ func (dbo *DBOrchestratorPoolCache) GetOrchestrators(numOrchestrators int, suspe
 
 		// check if O's price is below B's max price
 		maxPrice := server.BroadcastCfg.MaxPrice()
-		price := big.NewRat(info.PriceInfo.PricePerUnit, info.PriceInfo.PixelsPerUnit)
+		price, err := common.RatPriceInfo(info.PriceInfo)
```
yondonfu, line 111: I think problems can still occur if `info.PriceInfo` is nil. In this case, `RatPriceInfo()` will return `nil, nil`. Since the error is nil, a nil `price` will be used in the below conditional which would trigger a nil pointer error. We should make sure that `price` is not nil before using it.

yondonfu (changes requested):

discovery/db_discovery.go
```diff
@@ -108,9 +108,17 @@ func (dbo *DBOrchestratorPoolCache) GetOrchestrators(numOrchestrators int, suspe
...
+		if price == nil {
```
yondonfu: I think this check should be after the error check. Otherwise, if an error is returned from `RatPriceInfo()`, price will be nil and this conditional will trigger first, but we won't log the error returned from `RatPriceInfo()`. If we do the error check first that ensures that if there is an error from `RatPriceInfo()` we'll log it.

yondonfu (changes requested):

kyriediculous:
Sorry , yes you're entirely correct.
Fixed in 1a7be84

yondonfu (approved):
LGTM after squashing!

PR #1818, opened 2021-03-26
discovery,pm,server: Additional checks to avoid nil pointer errors caused by unexpected orchestrator configurations

kyriediculous:

- Return an error from `validateTicketParams()` if the provided `ticketParams` argument is `nil`
- Skip an orchestrator during latency / stake weight selection if it has no ticket params 
- Skip caching an orchestrator to the DB if it has no ETH address 

- Unit tests
- Reproduce crash cases and check they don't happen again. 

**Checklist:**

### DIFF
--- a/server/selection.go
+++ b/server/selection.go
@@ -162,6 +162,9 @@ func (s *MinLSSelector) selectUnknownSession() *BroadcastSession {
 	var addrs []ethcommon.Address
 	addrCount := make(map[ethcommon.Address]int)
 	for _, sess := range s.unknownSessions {
+		if sess.OrchestratorInfo.GetTicketParams() == nil {
+			continue
+		}
 		addr := ethcommon.BytesToAddress(sess.OrchestratorInfo.TicketParams.Recipient)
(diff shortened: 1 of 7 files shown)


server/selection.go
```diff
@@ -162,6 +162,9 @@ func (s *MinLSSelector) selectUnknownSession() *BroadcastSession {
 	var addrs []ethcommon.Address
 	addrCount := make(map[ethcommon.Address]int)
 	for _, sess := range s.unknownSessions {
+		if sess.OrchestratorInfo == nil || sess.OrchestratorInfo.TicketParams == nil {
```
yondonfu: Could this be shortened to `sess.OrchestratorInfo.GetTicketParams() == nil` since `GetTicketParams()` handles the case where `OrchestratorInfo` is nil gracefully?

yondonfu (changes requested):

server/selection.go
kyriediculous: fixed in 334e281b

yondonfu (approved):
LGTM 🚢

PR #1821, opened 2021-03-30
.github/workflows+docker: Docker layer caching for Linux build

yondonfu:

- 
- 
- 

**Checklist:**


yondonfu:
The Linux build in this PR takes 5m and the tests take 6m.

The Linux build in https://github.com/livepeer/go-livepeer/pull/1840 took 3m and the tests took 9m.

I didn't realize this (doh) before working on this PR that the current Linux build already supports a form of caching by pulling `livepeerci/build-platform:latest` and specifying it as the `--cache-from` value when building a fresh builder image. And it looks like that approach is faster than loading the cached Docker layers using buildx + the GH cache action in this PR.

The reason that the tests are faster in this PR than the latest master is because https://github.com/livepeer/go-livepeer/pull/1821/commits/99d200c0196adbca136e9679d074c34fd1291546 removed the localdocker build which shaves off some time. This could be worth doing regardless.

So, at the moment, it doesn't look like the caching approach in this PR improves on the current caching approach substantially...I'll leave this draft PR open for a bit in case any additional improvements come to mind to try, but if not will close it at the end of the week and extract out the removal of the localdocker build in the tests (which I don't think is necessary anymore).

PR #1823, opened 2021-04-01
server: Update NonRetryable errors to latest LPMS

### LINKED ISSUE #1774: Mark all LPMS errors as non-retryable?

There are other LPMS transcode errors not in the exported LPMSErrors map that are not currently marked as non-retryable. As a result, when B encounters these errors from O it'll retry the segment even though the segments cannot be transcoded.

Ex.

- [No video parameters found while initializing stream](https://github.com/livepeer/lpms/blob/967c18793f06efbdb4af3fb6f35d3bc5981ec911/ffmpeg/ffmpeg.go#L198)
- `Invalid argument`

Perhaps it'd make sense to just treat all errors that come from LPMS as non-retryable instead of trying to enumerate all of them. This may be too extreme though - should think about whether there are cases where the errors from LPMS should be retryable.

jailuthra:

Update non-retriable error list to include more errors

*See commit history*

Manually tested desired behaviour occurs for newly added errors in the list like "Audio Only Segment".

```
W0406 05:11:44.990130 2605883 broadcast.go:447] Not retrying current segment nonce=9309343988758959823 seqNo=1 due to non-retryable error err=TranscoderInvalidVideo
E0406 05:11:44.990153 2605883 mediaserver.go:874] http push error processing segment url=http://127.0.0.1:8936/live/test/1.ts manifestID=test err=TranscoderInvalidVideo
```

Fixes #1774 

**Checklist:**


server/broadcast.go
```diff
@@ -872,8 +872,8 @@ func getSegDurMsString(seg *stream.HLSSegment) string {
 
 func isNonRetryableError(e error) bool {
 	foundErr := false
-	for _, v := range ffmpeg.LPMSErrors {
-		if e.Error() == v.Desc {
+	for _, v := range ffmpeg.NonRetryableErrs {
```
yondonfu, line 875: Iterating through the slice is probably ok for now since it is still small, but we'll want to keep an eye on this if the slice gets larger since isNonRetryableError() is called for every single segment.

yondonfu (approved):
LGTM after resolving conflicts.

server/broadcast.go
jailuthra: Did a quick fix for it https://github.com/livepeer/go-livepeer/pull/1823/commits/e30a8087246f88c8f6081ab1b54365a036fa701d and tested that it works.

@yondonfu Still good to merge this PR? (I'm cool with keeping it till https://github.com/livepeer/go-livepeer/pull/1823/commits/1b6a7e3b617459cc43cdefb0fbd82949b82dcc91 and doing the bool map change in LPMS sometime later)

yondonfu (approved):
LGTM

server/broadcast.go
yondonfu: > doing the bool map change in LPMS sometime later

👍

jailuthra: dang it, saw your comment rn (github comment refresh is weird)
but will move the map code into LPMS when I get a chance

PR #1827, opened 2021-04-04
drivers,server: Cap data read for a seg

### LINKED ISSUE #1523: Cap data read in HandlePush handler
Currently, `HandlePush` handler reads incoming request unconditionally, all the way through.
```golang
	// we read this unconditionally, mostly for ffmpeg
	body, err := ioutil.ReadAll(r.Body)
```
That means that bad actor can stream until node will eat up all the memory and will be killed.

We should cap this read at some reasonable amount.

jailuthra:

Cap the data read for a segment to prevent bad segments eating up a lot of memory.

*see the commit history*

Currently the max segment size is set to ~300MB (at 8000kbps using the current max duration of 5minutes)

- Manually tested on a local B/OT setup with a smaller limit set (2MB) and a bigger segment (2.4MB). The segment data read was being capped at the defined size.

With Limit Set:
B: `I0404 22:11:10.044429 2542338 broadcast.go:372] Processing segment nonce=17696729379380681746 manifestID=test seqNo=1 dur=2 bytes=2097152`
OT: `[h264 @ 0x7f810403d640] error while decoding MB 60 59, bytestream -12`
Transcoded Rendition Duration: 1.23 s

Without Limit Set:
B: `I0404 22:12:42.698367 2543964 broadcast.go:372] Processing segment nonce=13032173926343571666 manifestID=test seqNo=1 dur=2 bytes=2405084`
OT: *No error*
Transcoded Rendition Duration: 2.7 s

2. Added unit tests for Push API and B<->O serve segment in https://github.com/livepeer/go-livepeer/pull/1827/commits/9d23156dc5aa076b4e8ec0becb5d777eede42eba

Fixes #1523 

**Checklist:**

### DIFF
--- a/common/util.go
+++ b/common/util.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"encoding/hex"
 	"fmt"
+	"io"
+	"io/ioutil"
 	"math"
 	"math/big"
 	"math/rand"
@@ -26,8 +28,14 @@ import (
 // HTTPTimeout timeout used in HTTP connections between nodes
 var HTTPTimeout = 8 * time.Second
 
-// Max Segment Size (cap reading HTTP response body at this size)
-const MaxSegSize = 20 * 1024 * 1024 // 20 MiB - 20s of 1080p30fps video @ 8000kbps
+// Max Segment Duration
+var MaxDuration = (5 * time.Minute)
+
+// Max Input Bitrate (bits/sec)
+var maxInputBitrate = 8000 * 1000 // 8000kbps
+
+// Max Segment Size in bytes (cap reading HTTP response body at this size)
+var MaxSegSize = int(MaxDuration.Seconds()) * (maxInputBitrate / 8)
 
 const maxInt64 = int64(math.MaxInt64)
 
@@ -389,4 +397,16 @@ func JoinURL(url, path string) string {
 		return url + "/" + path
 	}
 	return url + path
+}
+
+// Read at most n bytes from an io.Reader
+func ReadAtMost(r io.Reader, n int) ([]byte, error) {
+	// Reading one extra byte to check if input Reader
+	// had more than n bytes
+	limitedReader := io.LimitReader(r, int64(n)+1)
+	b, err := ioutil.ReadAll(limitedReader)
+	if err == nil && len(b) > n {
+		return nil, errors.New("input bigger than max buffer size")
+	}
+	return b, err
 }
--- a/common/util.go
+++ b/common/util.go
@@ -26,6 +26,9 @@ import (
 // HTTPTimeout timeout used in HTTP connections between nodes
 var HTTPTimeout = 8 * time.Second
 
+// Max Segment Size (cap reading HTTP response body at this size)
+const MaxSegSize = 20 * 1024 * 1024 // 20 MiB - 20s of 1080p30fps video @ 8000kbps
+
 const maxInt64 = int64(math.MaxInt64)
 
 // using a scaleFactor of 1000 for orchestrator prices
--- a/server/mediaserver.go
+++ b/server/mediaserver.go
@@ -692,9 +692,7 @@ func (s *LivepeerServer) HandlePush(w http.ResponseWriter, r *http.Request) {
 		http.Error(w, httpErr, http.StatusMethodNotAllowed)
 		return
 	}
-	// we read this unconditionally, mostly for ffmpeg
-	bodyReader := io.LimitReader(r.Body, common.MaxSegSize)
-	body, err := ioutil.ReadAll(bodyReader)
+	body, err := common.ReadAtMost(r.Body, common.MaxSegSize)
 
 	if err != nil {
 		httpErr := fmt.Sprintf(`Error reading http request body: %s`, err.Error())
--- a/server/mediaserver.go
+++ b/server/mediaserver.go
@@ -693,7 +693,8 @@ func (s *LivepeerServer) HandlePush(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	// we read this unconditionally, mostly for ffmpeg
-	body, err := ioutil.ReadAll(r.Body)
+	bodyReader := io.LimitReader(r.Body, common.MaxSegSize)
+	body, err := ioutil.ReadAll(bodyReader)
 
 	if err != nil {
 		httpErr := fmt.Sprintf(`Error reading http request body: %s`, err.Error())
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -7,6 +7,7 @@
 #### General
 
 - \#1810 Display "n/a" in CLI when max gas price isn't specified (@kyriediculous)
+- \#1827 Limit the maximum size of a segment read over HTTP (@jailuthra)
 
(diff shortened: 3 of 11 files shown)


common/util.go
```diff
@@ -387,3 +392,8 @@ func JoinURL(url, path string) string {
 	}
 	return url + path
 }
+
+func ReadAtMost(r io.Reader, n int64) ([]byte, error) {
```
jailuthra: Not logging any warnings or returning an error here for the case when the segment size > `MaxSegSize` (and we end up reading it partially)

We could add some errors/warning around this if we write this function from ground up using manual `Read` calls and increasing buffer gradually if needed - like done in [`ReadAll`'s implementation](https://golang.org/src/io/io.go?s=19249:19287#L616)

server/mediaserver.go
```diff
@@ -693,7 +693,7 @@ func (s *LivepeerServer) HandlePush(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	// we read this unconditionally, mostly for ffmpeg
```
darkdarkdragon: I think this comment is obsolete.

common/util.go
darkdarkdragon: I think we should return error from here if not all data was read - because if broadcaster didn't downloaded all segment it doesn't make sense to send it to transcoding - anyway it will be broken segment.

darkdarkdragon (dismissed):

common/util.go
jailuthra: Makes sense on it being a broken segment and needing error log of some sort, will update!

NB: I was able to transcode partial segment on CPU (haven't tested on GPU), so was uncertain if we should still make an attempt at transcoding or not - but for now makes sense to just error out and reconsider later if we see huge segments.

darkdarkdragon: I don't think users expect they segments to be _partially_ transcoded 😄

common/util.go
```diff
@@ -26,6 +28,9 @@ import (
 // HTTPTimeout timeout used in HTTP connections between nodes
 var HTTPTimeout = 8 * time.Second
 
+// Max Segment Size (cap reading HTTP response body at this size)
+const MaxSegSize = 20 * 1024 * 1024 // 20 MiB - 20s of 1080p30fps video @ 8000kbps
```
yondonfu: The max duration of a segment is currently [5 minutes](https://github.com/livepeer/go-livepeer/blob/ffabc358fe3115c81e0fb6fc83f1b6b281194498/server/broadcast.go#L32). Support for segments up to this max duration was added awhile back. I'm not sure how common longer segments as a use case are today (they came up while experimenting with preliminary VOD support + mp4 transcoding), but I think we should set this byte size limit based on the current duration limit.

yondonfu (changes requested):
After the update to start returning an error if the byte limit is exceeded while reading a request body, let's make sure to test for that error. I recommend small unit test cases as those should make it easy to construct request bodies that exceed the byte limit.

iameli:
@yondonfu @jailuthra @darkdarkdragon I feel medium-strongly this should be a configurable parameter with a default value, perhaps `-httpIngestMaxSize=20MiB` or some such. Could use https://github.com/c2h5oh/datasize to make it easy to parse various sizes?

Just feels like the kind of thing I'd like the ability to do in a production setting to unbreak an integration that's ever-so-slightly over the limit.

darkdarkdragon:
@iameli we also using this size on the O side. If we make this configurable then we should put this value in discovery protocol so O will clearly advertise its limits - or else B can get unexpected errors.

jailuthra:
@iameli If we go with the bigger hardcoded limit of `5minutes` as proposed by @yondonfu above, would we still need to increase it anytime soon-ish in a prod environment?

This can avoid changes on discovery side (for now), and I am definitely in for iterating over this in future!

iameli:
Sure, that's reasonable.

common/util.go
```diff
@@ -387,3 +395,15 @@ func JoinURL(url, path string) string {
...
+
+// Read at most n bytes from an io.Reader
+// error out if the io.Reader had bytes remaining
+func ReadAtMost(r io.Reader, n int) ([]byte, error) {
+	limitedReader := io.LimitReader(r, int64(n)+1)
+	b, err := ioutil.ReadAll(limitedReader)
+	if err == nil && len(b) > n {
+		return nil, errors.New("input bigger than max buffer size")
+	} else {
+		return b, err
```
yondonfu: More idiomatic in Go to do:

```golang
if err == nil && len(b) > n {
    return nil, errors.New("input bigger than max buffer size")
}

return b, err
```

common/util.go
```diff
@@ -26,6 +28,12 @@ import (
 // HTTPTimeout timeout used in HTTP connections between nodes
 var HTTPTimeout = 8 * time.Second
 
+// Max Segment Duration
+var MaxDuration = (5 * time.Minute)
+
+// Max Segment Size in bytes (cap reading HTTP response body at this size)
+var MaxSegSize = int(MaxDuration.Seconds()) * 1024 * 1024 // 1080p30fps video @ 8000Kb/s or 1MiB/s
```
yondonfu: This might be obvious, but could you explain this calculation for my own understanding?

common/util.go
```diff
@@ -387,3 +395,15 @@ func JoinURL(url, path string) string {
 	}
 	return url + path
 }
+
+// Read at most n bytes from an io.Reader
+// error out if the io.Reader had bytes remaining
+func ReadAtMost(r io.Reader, n int) ([]byte, error) {
+	limitedReader := io.LimitReader(r, int64(n)+1)
```
yondonfu, line 406: Should the limit be n as opposed to n + 1?

yondonfu (changes requested):
Let's add some small tests for the new limiting behavior?

common/util.go
jailuthra: Assuming the biggest video configuration we expect is 1080p30fps @ 8000kbps (i.e. ~ 1MBytes / s) every second of the video will take up `1024*1024` bytes
Edit: Fixed the comment to make it clearer for future reader

common/util.go
jailuthra: If we keep it as `n`, `len(b)` in the subsequent check will always be `<= n`, and we won't be able to find out if the underlying `io.Reader` had more bytes than we wanted to read (i.e. `n`)

I'll add a comment to make this clearer.

common/util.go
yondonfu: Ah right brain fart on my end heh 👍

jailuthra: Thinking more about this, I'll create a `MaxBitrate` variable here to make it easier on the eyes lol.

yondonfu:
Latest changes look good to go for me pending the tests for the byte reading limiting behavior.

jailuthra:
> Latest changes look good to go for me pending the tests for the byte reading limiting behavior.

Added a couple of tests in https://github.com/livepeer/go-livepeer/pull/1827/commits/9d23156dc5aa076b4e8ec0becb5d777eede42eba, rebased the PR and added a changelog entry.

Also created a fork debt issue for making the max segment size configurable.

yondonfu (approved):
LGTM

PR #1833, opened 2021-04-08
eth,pm: avoid watchPoolSizeChange nil pointer error

### LINKED ISSUE #1828: Nil pointer in LocalSenderMonitor watchPoolSizeChange()
Screen Shot 2021-04-04 at 4 59 36 PM

The line that triggers it is [here](https://github.com/livepeer/go-livepeer/blob/4e24711bfabc873f1be7e98cdd8599f68a6cfedd/pm/sendermonitor.go#L489).

kyriediculous:
This issue might be caused by an unexpected return value from the ETH provider. 

The error is caused by the following expression 

```go
if poolSize := sm.tm.GetTranscoderPoolSize(); poolSize.Cmp(lastPoolSize) != 0
```

Both values are fetched from `TimeWatcher.GetTranscoderPoolSize()`

The value of `TimeWatcher.transcoderPoolSize` is initialised on node startup and refetched from the ETH provider when events are received. 

The contract binding `BondingManagerSession.GetTranscoderPoolSize()` returns the following `return *(new(*big.Int)), err`. 
This means that even though the return value (a pointer) can't be nil because it is initialised , but value the pointer refers to can still be and this is what is returned. If for some reason there is a nil error and an unexpected value from the contract call a nil pointer could still occur. 

- Wrap the contract call in case it returns `nil, nil`

Fixes #1828 

**Checklist:**

### DIFF
--- a/eth/stubclient.go
+++ b/eth/stubclient.go
@@ -201,7 +201,6 @@ type StubClient struct {
 	ProcessHistoricalUnbondError error
 	Orchestrators                []*lpTypes.Transcoder
 	Round                        *big.Int
-	RoundsErr                    error
 	SenderInfo                   *pm.SenderInfo
 	PoolSize                     *big.Int
 	ClaimedAmount                *big.Int
@@ -213,6 +212,7 @@ type StubClient struct {
 	TranscoderPoolError          error
 	RoundLocked                  bool
 	RoundLockedErr               error
+	Errors                       map[string]error
 }
 
 type stubTranscoder struct {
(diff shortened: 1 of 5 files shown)


yondonfu:
> This means that even though the return value (a pointer) can't be nil because it is initialised , but value the pointer refers to can still be and this is what is returned. If for some reason there is a nil error and an unexpected value from the contract call a nil pointer could still occur.

Did you actually confirm that the contract binding is returning a nil value even when the error is nil?

The nil pointer error in #1828 is being trigger in [this line](https://github.com/livepeer/go-livepeer/blob/master/pm/sendermonitor.go#L490). I believe `poolSize` could be nil *or* `lastPoolSize` could be nil. The latter makes more sense to me because the first time that the pool size in the TimeWatcher is set is in the [Watch()](https://github.com/livepeer/go-livepeer/blob/f9ddb6195981ad9e0c360fb29eddf057cb84ca3e/eth/watchers/timewatcher.go#L133) method which is called in a [goroutine](https://github.com/livepeer/go-livepeer/blob/f9ddb6195981ad9e0c360fb29eddf057cb84ca3e/cmd/livepeer/livepeer.go#L454). So, if `lastPoolSize` is initialized in [LocalSenderMonitor](https://github.com/livepeer/go-livepeer/blob/master/pm/sendermonitor.go#L480) before the pool size is initialized in the TimeWatcher then `lastPoolSize` would be nil.

I think this indicates another problem - LocalSenderMonitor depends on state in TimeWatcher that might not be initialized yet. One way to address this particular problem could be to move the state initialization logic (such as [this line](https://github.com/livepeer/go-livepeer/blob/master/eth/watchers/timewatcher.go#L133)) into the TimeWatcher constructor which is called synchronously before `Watch()` is called in a goroutine. Additionally, it may make sense to default the return value of `TimeWatcher.GetTranscoderPoolSize()` to some non-nil value to guard against nil pointer errors just to be safe (i.e. 0).

kyriediculous:
> Did you actually confirm that the contract binding is returning a nil value even when the error is nil?

No but it could be possible if a ethereum client is used with potential non-default behaviour here I guess (or some erroneous code). You're probably right that a properly implemented client likely doesn't return `nil, nil` here. 

So yeah a race condition is more likely to be the cause here.

yondonfu:
> No but it could be possible if a ethereum client is used with potential non-default behaviour here I guess (or some erroneous code). You're probably right that a properly implemented client likely doesn't return nil, nil here.

I think the [contract binding](https://github.com/livepeer/go-livepeer/blob/3610c2e763dc0907d278484518fb085a8bbfa2c2/eth/contracts/bondingManager.go#L596) implementation should guarantee that the `big.Int` value is always non-nil.

eth/stubclient.go
```diff
@@ -213,28 +212,35 @@ type StubClient struct {
...
 func (e *StubClient) BlockHashForRound(round *big.Int) ([32]byte, error) {
-	return e.BlockHashToReturn, nil
+	return e.BlockHashToReturn, e.Errors["BlockHashForRound"]
 }
-func (e *StubClient) CurrentRoundInitialized() (bool, error)    { return false, nil }
-func (e *StubClient) CurrentRoundLocked() (bool, error)         { return e.RoundLocked, e.RoundLockedErr }
-func (e *StubClient) CurrentRoundStartBlock() (*big.Int, error) { return nil, nil }
-func (e *StubClient) Paused() (bool, error)                     { return false, nil }
+func (e *StubClient) CurrentRoundInitialized() (bool, error) { return false, nil }
+func (e *StubClient) CurrentRoundLocked() (bool, error)      { return e.RoundLocked, e.RoundLockedErr }
```
yondonfu, line 239: Should RoundLockedErr also be replaced with Errors[<METHOD_NAME>] as was done for RoundsErr in CurrentRound()?

eth/watchers/timewatcher.go
```diff
@@ -54,12 +54,44 @@ func NewTimeWatcher(roundsManagerAddr ethcommon.Address, watcher BlockWatcher, l
...
+	if err != nil {
+		return nil, fmt.Errorf("error fetching initial lastInitializedRound value err=%v", err)
+	}
+	bh, err := tw.lpEth.BlockHashForRound(lr)
+	if err != nil {
+		return nil, fmt.Errorf("error fetching initial lastInitializedBlockHash value err=%v", err)
+	}
+	num, err := tw.lpEth.CurrentRoundStartBlock()
+	if err != nil {
+		return nil, fmt.Errorf("error fetching current round start block %v=", err)
```
yondonfu: I think this typo was here before, but look likes %v= should be err=%v.

yondonfu (changes requested):
Looks pretty good. Just 2 minor comments.

eth/stubclient.go
kyriediculous: Since we don't call this method in the `TimeWatcher` I decided to keep the diff small for now, but can definitely submit a follow up PR to clean up the rest of the errors in the stub Eth client to follow the same principle.

eth/watchers/timewatcher.go
yondonfu: The related unit test needs to be fixed as well.

kyriediculous: Missed that, should be fixed now

yondonfu (approved):
LGTM after squashing

PR #1837, opened 2021-04-11
server+discovery: Only log errors if request was not cancelled

### LINKED ISSUE #1846: Silence discovery error logs if the request is cancelled

During discovery, a B will call GetOrchestrator() on all available Os. Once B receives a sufficient number of responses, it will cancel the remaining requests using a context. The cancellation of these requests results in a ton of error messages indicating that the GetOrchestrator() call failed due to a cancellation. As a result, the logs of B can become flooded with these error logs even though the logs don't indicate any issues with the Os because it was the B that cancelled the request.

Silence discovery error logs if the GetOrchestrator() call was cancelled.

This [draft PR](https://github.com/livepeer/go-livepeer/pull/1837) outlines one potential approach to silencing those discovery error logs by using the new [errors Is() API](https://golang.org/pkg/errors/#Is). But, the code is untested and it is not clear if the `Is()` API is going to work as expected - I'm not sure if the function that returns the error would need to support error wrapping/unwrapping. An alternative would be to check if the error string contains a context cancellation message. Being able to use the errors `Is()` API could be cleaner though as I think this type of error checking is what it is meant for.

yondonfu:

- Wrap discovery related errors in the `server` package instead of concatenating the error message strings. This allows usage of `errors.Is()` because it can now unwrap error messages.
- Add a check to see if an error from discovery doesn't contain `context.Canceled` or `context.Exceeded` and only log the error if it doesn't

Ran a broadcaster on rinkeby

Fixes #1846 

**Checklist:**

### DIFF
--- a/discovery/discovery.go
+++ b/discovery/discovery.go
@@ -3,6 +3,7 @@ package discovery
 import (
 	"container/heap"
 	"context"
+	"errors"
 	"math"
 	"math/rand"
 	"net/url"
@@ -85,8 +86,11 @@ func (o *orchestratorPool) GetOrchestrators(numOrchestrators int, suspender comm
 			infoCh <- info
 			return
 		}
-		if err != nil && monitor.Enabled {
-			monitor.LogDiscoveryError(err.Error())
+		if err != nil && !errors.Is(err, context.Canceled) && !errors.Is(err, context.DeadlineExceeded) {
+			glog.Error(err)
+			if monitor.Enabled {
+				monitor.LogDiscoveryError(err.Error())
+			}
 		}
 		errCh <- err
 	}
--- a/discovery/discovery.go
+++ b/discovery/discovery.go
@@ -86,7 +86,7 @@ func (o *orchestratorPool) GetOrchestrators(numOrchestrators int, suspender comm
 			infoCh <- info
 			return
 		}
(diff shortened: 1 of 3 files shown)


discovery/discovery.go
```diff
@@ -85,8 +86,11 @@ func (o *orchestratorPool) GetOrchestrators(numOrchestrators int, suspender comm
 			infoCh <- info
 			return
 		}
-		if err != nil && monitor.Enabled {
-			monitor.LogDiscoveryError(err.Error())
+		if err != nil && !errors.Is(err, context.Canceled) && !errors.Is(err, context.DeadlineExceeded) {
```
yondonfu: I think we should still log an error if the error wraps a `context.DeadlineExceeded` as that would indicate a timeout with the O which still feels useful to be aware of as it tells us the O could not respond within the timeout while a `context.Cancelled` (and empirically, the majority of excessive logging on mainnet for Bs seems to be from `context.Cancelled`) is always B initiated and doesn't really tell us anything about O's ability to respond. WDYT?

kyriediculous: Fixed and rebased, will approve when CI passes.

kyriediculous (approved):

kyriediculous:
I just updated the commit message and force pushed (removed "deadline exceeded" bit) , CI already passed before this so I think iit should be okay to merge.

PR #1840, opened 2021-04-14
Enumerate Nvidia GPU IDs

### LINKED ISSUE #1205: Implement -nvidia=all
"Please run this transcoder on all nvidia GPUs available" I think is going to be a pretty common case — it'd be nice if Kubernetes manifests and whatnot could contain a line like`-nvidia=all` so that I can have the same command work across a variety of hardware.

jailuthra:

Automatically utilize all available Nvidia GPUs for transcoding, if the user sets the command line parameter as `-nvidia=all` or `-nvidia all`.

See commit history

Manually tested, and verified correct behaviour on:
- [x] Linux desktop with a single GTX 1060 6gb
- [x] Linux intel laptop with no nvidia GPU
- [x] Berlin staging pod with multiple nvidia GPUs
- [x] [Eli's test](https://github.com/livepeer/go-livepeer/pull/1840#issuecomment-820107695) on a windows machine

Fixes #1205

**Checklist:**


iameli:
Tested on Windows, working great!

image

common/util.go
```diff
@@ -410,3 +411,45 @@ func ReadAtMost(r io.Reader, n int) ([]byte, error) {
...
+
+	for i := 0; i < nvidiaCardCount; i++ {
+		s := strconv.Itoa(i)
+		devices = append(devices, s)
+	}
+
+	return devices
+}
+
+func ParseNvidiaDevicesFlag(nvidia string) []string {
```
kyriediculous: I wonder if there's a better place for this other than `common` as it's not really shared across packages, it's only used in `cmd`.  Don't really have a suggestion atm.

common/util.go
```diff
@@ -410,3 +411,45 @@ func ReadAtMost(r io.Reader, n int) ([]byte, error) {
...
+	var devices []string
+	if nvidia == "all" {
+		devices = enumerateNvidiaDevices()
+		if devices == nil || len(devices) == 0 {
```
kyriediculous: Devices can never be `nil` as it's a value type (all arrays in golang are value types), we can shorten this to `len(devices) == 0`

See https://play.golang.org/p/evGL_j96MHb

```go
package main

import (
	"fmt"
)

func main() {
 	a :=[]string{}
	b := []*string{}
	a = nil
	b = nil
	fmt.Println(a, b)
}
```

Console output: 
```bash
[] []

```

common/util.go
```diff
@@ -410,3 +411,45 @@ func ReadAtMost(r io.Reader, n int) ([]byte, error) {
...
+			glog.Warningf("Unable to query available Nvidia gpus, using default device ID '0'")
+			devices = []string{"0"}
```
kyriediculous: If no devices are found (`devices == 0`) defaulting to a GPU seems likeit could lead to unexpected behaviour or errors (in case there really is no GPU). 

Perhaps a sensible default would be to default to CPU transcoding and clearly log this message. Alternatively, when a user explicitly provides a -nvidia flag the node should always have access to a GPU, in case `devices == 0` the node should exit with an error. 

WDYT? Might be missing something here.

kyriediculous (changes requested):

common/util.go
jailuthra: Ah TIL, will update!

common/util.go
jailuthra: > If no devices are found (devices == 0) defaulting to a GPU seems like it could lead to unexpected behaviour or errors (in case there really is no GPU)

This was done as a fallback for the case when this newly introduced "GPU detection" fails for a user, but a GPU is still present.

If a GPU is not present at all it would anyway fail gracefully during startup with `-testTranscoder` enabled by default. And we currently allow a user to manually set `-nvidia=0` on a CPU-only machine, relying on that test.

Once we're sure of the reliability of this detection on different user's systems, we can use it to decide to fallback on CPU or even to verify if the user's explicitly passed GPU ID is valid. But we'll still probably need `-testTranscoder` to detect session limit caps.

For now I'm leaning on keeping this as-is, WDYT?

common/util.go
jailuthra: The other obvious option is moving this in a subpackage in cmd like `cmd/utils` - but both `livepeer.go` and `livepeer_bench.go` anyway use helper functions from `common/util.go` - eg. [parsing video profiles](https://github.com/livepeer/go-livepeer/blob/3908252d23edd62590c61b02598c5fffc148014b/cmd/livepeer_bench/livepeer_bench.go#L214) for the transcoder. But I don't have a strong opinion either way.

common/util.go
```diff
@@ -410,3 +411,47 @@ func ReadAtMost(r io.Reader, n int) ([]byte, error) {
 	}
 	return b, err
 }
+
+func enumerateNvidiaDevices() []string {
```
yondonfu: Instead of logging errors in these helper functions can we return the errors and let the caller determine how to handle them (i.e. logging the errors in `livepeer.go`)? At the moment, we just end up logging these errors, but this change would give the caller the choice of how to handle the errors and doesn't require errors to always be logged in case the functions are used in other contexts.

common/util.go
```diff
@@ -410,3 +411,47 @@ func ReadAtMost(r io.Reader, n int) ([]byte, error) {
...
+	gpu, err := ghw.GPU()
+	if err != nil {
+		glog.Errorf("Error initializing GPU detection via ghw: %v", err)
+		return nil
+	}
+
+	nvidiaCardCount := 0
+	for _, card := range gpu.GraphicsCards {
+		if strings.EqualFold(card.DeviceInfo.Vendor.Name[:6], "nvidia") {
```
yondonfu: I wonder if a check that doesn't require slicing into the string with a hardcoded value could be to use a regular expression like this:

```golang
re := regexp.MustCompile("(?i)Nvidia") // I think this should be a case insensitive regexp
if re.MatchString(card.DeviceInfo.Vendor.Name) {
     ...
}
```

Just a suggestion to consider. I don't feel that strongly about it.

common/util.go
yondonfu: I'm in favor of being more explicit with the failure case if no devices are found and clearly logging a message indicating that no devices were found instead of allowing startup to proceed and hit the `-testTranscoder` tests. The `-testTranscoder` tests will fail if it is actually the case that no devices are found, but I don't think the error messages are particularly user friendly (someone feel free to question this). So, if there is an opportunity to fail earlier if we know there are no devices available then I think doing so can be more informative for the user.

I also agree with Nico that defaulting to the first device if there is a problem with GPU detection when the user specifies `-nvidia=all` feels like it would be surprising to a user. From my own POV, I would be surprised if I passed in `-nvidia=all` expecting to use all 8 of my GPUs only to realize later on that only 1 GPU is being used. Additionally, if no devices are detected, but there *are* actually devices present due to some issue with the GPU detection feature, then a user is still able to manually specify device IDs if absolutely necessary.

common/util.go
```diff
@@ -410,3 +411,47 @@ func ReadAtMost(r io.Reader, n int) ([]byte, error) {
...
+
+	for i := 0; i < nvidiaCardCount; i++ {
+		s := strconv.Itoa(i)
+		devices = append(devices, s)
+	}
+
+	return devices
+}
+
+func ParseNvidiaDevicesFlag(nvidia string) []string {
```
yondonfu: I suggest dropping the "Flag" as we can treat this as a function that parses a string of Nvidia devices that *could* come from a flag, but does not necessarily need to be from a flag.

yondonfu (changes requested):
Since we're using a package that gives us access to all device info now, an additional UX improvement could be to log the Nvidia device names that the node is configured with on startup so it is clear to the user what devices are being used. Feel free to consider this for a separate PR though.

common/util.go
kyriediculous: let's keep it in `common` then 👍

common/util.go
jailuthra: > I would be surprised if I passed in -nvidia=all expecting to use all 8 of my GPUs only to realize later on that only 1 GPU is being used.

Yep that makes sense. I've updated the log line, and also using `Fatalf` in the fixup commit.

common/util.go
jailuthra: Neat idea to use regexp here; updated!

jailuthra:
Fixed pending comments, and squashed+rebased on top of latest `master`. Also did some sanity testing on my 1060 again.

>  an additional UX improvement could be to log the Nvidia device names that the node is configured with on startup so it is clear to the user what devices are being used. Feel free to consider this for a separate PR though.

That would be a very cool feature. Will open a separate tracking issue for it.

doc/gpu.md
```diff
@@ -22,16 +29,19 @@ interleaved). Anything else will return an error.
...
 * **CUDA Availability** If running the Livepeer binary, the CUDA shared libraries are expected to be installed in `/usr/local/cuda`. If the CUDA location differs on your machine, run the node with `LD_LIBRARY_PATH=</path/to/cuda>` environment variable.
 
-So far, Livepeer has been tested  to work with the following driver versions:
+So far, Livepeer has been tested to work with the following driver versions:
 
 CUDA | Nvidia
 --|--
 10.0.130 |
 10.1 | 418.39 , 430.50
-10.2.89 | 440.33.01
```
yondonfu, line 31: What's the rationale for removing references to these CUDA versions?

cmd/livepeer/livepeer.go
```diff
@@ -269,13 +269,21 @@ func main() {
 	if *transcoder {
 		core.WorkDir = *datadir
 		if *nvidia != "" {
+			// Get a list of device ids
+			devices, err := common.ParseNvidiaDevices(*nvidia)
+			if err != nil {
+				glog.Fatalf("Couldn't parse Nvidia GPUs: %v", err)
```
yondonfu: Suggestion: "Couldn't parse -nvidia" to be more explicit that the problem is with the argument passed to `-nvidia`.

cmd/livepeer_bench/livepeer_bench.go
```diff
@@ -63,8 +63,12 @@ func main() {
 	accel := ffmpeg.Software
 	devices := []string{}
 	if *nvidia != "" {
+		var err error
 		accel = ffmpeg.Nvidia
-		devices = strings.Split(*nvidia, ",")
+		devices, err = common.ParseNvidiaDevices(*nvidia)
+		if err != nil {
+			glog.Fatalf("Couldn't parse Nvidia GPUs: %v", err)
```
yondonfu: Same comment as the one for `livepeer.go`.

doc/gpu.md
jailuthra: Combined this particular one with 440.118.02 (which was also tested with CUDA version 10.2.xx in BER - not sure what xx was). Most users can anyway not control the minor/micro version's of CUDA/Nvidia libs that are available for them to install.

Removed `11.0 | 450.36.06` because 450 occasionally lead to stuck sessions.

cmd/livepeer/livepeer.go
jailuthra: Updated in the fixup https://github.com/livepeer/go-livepeer/pull/1840/commits/bbdb6499813206e5679181195a07701aa8673698

kyriediculous:
Requested changes look good to me, will let Yondon follow up on his change requests seperately.

kyriediculous (approved):

yondonfu (approved):
LGTM

Windows build is failing due to what looks like a flaky gpg issue. Please trigger another build just to check if the Windows build passes and then this should be good to merge.

jailuthra:
Squashed the fixup to trigger windows re-build. Everything works now. Merging.

PR #1845, opened 2021-04-16
eth: hints for staking actions

### LINKED ISSUE #1567: staking actions with hints in the CLI
Provide support for staking actions with hints

Create wrapper methods on the Ethclient instead of calling the contract binding directly. 
See #1556 for an example

kyriediculous:
This PR adds support for hints when performing staking actions on the Livepeer smart contracts. 

- Add wrappers function for staking action to `eth/Client.go`

- [x] test on rinkeby and confirm gas used is lower and hints are correct

 
Fixes #1567 

**Checklist:**

### DIFF
--- a/eth/client.go
+++ b/eth/client.go
@@ -399,7 +399,8 @@ func (c *client) Transcoder(blockRewardCut, feeShare *big.Int) (*types.Transacti
 }
 
 func (c *client) Bond(amount *big.Int, to ethcommon.Address) (*types.Transaction, error) {
-	allowance, err := c.Allowance(c.Account().Address, c.bondingManagerAddr)
+	sender := c.Account().Address
+	allowance, err := c.Allowance(sender, c.bondingManagerAddr)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


kyriediculous:
Wanted to rework the helpers to get the hints first so we could avoid repeating getting the transcoder pool, transcoder pool size, ... . 

Then I realised the reason why we went with the functional programming approach for those is because of testability. 

Using a method on `client` would prevent the method to be testable with mock data with the current state of that implementation.

kyriediculous:
Added a fix in the way we calculate the stake amounts, previously posPrev always ended up being 0x00... 

tested on rinkeby and works now

eth/client.go
```diff
@@ -418,7 +419,186 @@ func (c *client) Bond(amount *big.Int, to ethcommon.Address) (*types.Transaction
...
+	if delegator.DelegateAddress != to {
+		oldHints = findTranscoderHints(delegator.DelegateAddress, transcoders)
+	}
+
+	currentRound, err := c.CurrentRound()
+	if err != nil {
+		return nil, err
+	}
+
+	ep, err := c.GetTranscoderEarningsPoolForRound(to, currentRound)
```
yondonfu: When inserting into the on-chain pool, the transcoder's `delegatedAmount` will be considered which can change during the round, but we're using the transcoder's total stake stored in the earnings pool which is locked in at the beginning of a round. We should be able to use the DelegatedAmount field included in the value returned by `GetDelegator()` called for the delegate here instead of calling `GetTranscoderEarningsPoolForRound()`.

eth/client.go
```diff
@@ -418,7 +419,186 @@ func (c *client) Bond(amount *big.Int, to ethcommon.Address) (*types.Transaction
...
+	if err != nil {
+		return nil, errors.Wrapf(err, "unable to get transcoder pool max size")
+	}
+
+	currentRound, err := c.CurrentRound()
+	if err != nil {
+		return nil, err
+	}
+
+	ep, err := c.GetTranscoderEarningsPoolForRound(sender, currentRound)
```
yondonfu: Same comment as for `Bond()` about using `delegatedAmount` instead

eth/client.go
```diff
@@ -418,7 +419,186 @@ func (c *client) Bond(amount *big.Int, to ethcommon.Address) (*types.Transaction
...
+	if err != nil {
+		return nil, errors.Wrapf(err, "unable to get transcoder pool max size")
+	}
+
+	currentRound, err := c.CurrentRound()
+	if err != nil {
+		return nil, err
+	}
+
+	ep, err := c.GetTranscoderEarningsPoolForRound(to, currentRound)
```
yondonfu: Same comment as for `Bond()` about using `delegatedAmount` instead

eth/client.go
```diff
@@ -418,7 +419,186 @@ func (c *client) Bond(amount *big.Int, to ethcommon.Address) (*types.Transaction
...
+	if err != nil {
+		return nil, err
+	}
+
+	totalBonded := ep.TotalStake
+
+	lock, err := c.GetDelegatorUnbondingLock(sender, unbondingLockID)
+	if err != nil {
+		return nil, errors.Wrapf(err, "Unable to get unbonding lock")
```
yondonfu: Unable -> unable for consistency in error messages

eth/client.go
```diff
@@ -418,7 +419,186 @@ func (c *client) Bond(amount *big.Int, to ethcommon.Address) (*types.Transaction
...
+	ep, err := c.GetTranscoderEarningsPoolForRound(sender, currentRound)
+	if err != nil {
+		return nil, err
+	}
+
+	totalBonded := ep.TotalStake
+
+	lock, err := c.GetDelegatorUnbondingLock(sender, unbondingLockID)
+	if err != nil {
+		return nil, errors.Wrapf(err, "Unable to get unbonding lock")
```
yondonfu: Unable -> unable for consistency in error messages

yondonfu (changes requested):
> tested on rinkeby and works now

Can you confirm that the calculated hints are working as expected and the gas costs are reduced relative to if the transactions were submitted without hints (i.e. by posting gas cost differences)?

kyriediculous:
State before 

image


**Self bond 5 LPT (16.66 + 5  = 21.66 LPT) :** 

```
Invoking transaction: "bondWithHint". Inputs: "_amount: 5000000000000000000  _to: 0x3566eFB780B6e9534399d9F8BeA1ac0F2B179d83  _oldDelegateNewPosPrev: 0x0000000000000000000000000000000000000000  _oldDelegateNewPosNext: 0x0000000000000000000000000000000000000000  _currDelegateNewPosPrev: 0xDAC817294c0c87ca4fA1895eF4b972EAde99f2fd  _currDelegateNewPosNext: 0xE358c19D115f0BCCdB51B5b7a04be24808E89418"  Hash: "0xd403ade95d4b70b1fc9d6dc84a438775ae8529ebe8acfd1e01f733319046abcb".
```

Transcoder list after transaction (hints match the new position) 

image


BondWithHint Transaction: https://rinkeby.etherscan.io/tx/0xd403ade95d4b70b1fc9d6dc84a438775ae8529ebe8acfd1e01f733319046abcb

Bond Transaction https://rinkeby.etherscan.io/tx/0x6b3c51ee116b0209b577c6fb13723f2b3c109265e3554469551e608db4c13ea3

For some reason the regular Bond transaction is ~20K gas cheaper, despite the hints being correct. The state update is the same. 

**Unbond 5 LPT** 

```
Invoking transaction: "unbondWithHint". Inputs: "_amount: 5000000000000000000  _newPosPrev: 0x0b041CcE35E3C08c5E2AFb76A25b6a53421abBc1  _newPosNext: 0xA6ddcF04D8EbF15E66440eB41C2C2075Ff358465"  Hash: "0x0c22c8f3d3936ad9273e6ee2408ff55d723d5652e2629c5db6377db0f8599ca0".
```

Hints should match the prev and nex positions of the original state (first picture) , they do. 

UnbondWithHint transaction https://rinkeby.etherscan.io/tx/0x0c22c8f3d3936ad9273e6ee2408ff55d723d5652e2629c5db6377db0f8599ca0

Unbond transaction
0xb6de7b28f9a25361eea490ec61f4515b1a38705e9368554dd72f2a7724e63def

UnbondWithHint is about 110K gas cheaper

**RebondWithHint**

```
Invoking transaction: "rebondWithHint". Inputs: "_unbondingLockId: 1  _newPosPrev: 0xDAC817294c0c87ca4fA1895eF4b972EAde99f2fd  _newPosNext: 0xE358c19D115f0BCCdB51B5b7a04be24808E89418"  Hash: "0x462a94b8bec172deafd584dc3886a5c2a36a011b525e14f86a8543a14e84a364".
```

Hints should again match the ones from the BondWithHint transaction, they do

RebondWithHint transaction https://rinkeby.etherscan.io/tx/0x462a94b8bec172deafd584dc3886a5c2a36a011b525e14f86a8543a14e84a364

Rebond transaction https://rinkeby.etherscan.io/tx/0xad50716c5883dced9606ea6debed524d089d97c2896c37cab2f32f3d3b20f51f

RebondWithHint is about ~27k gas cheaper

eth/client.go
kyriediculous: fixed in 6c5b9d4d

eth/client.go
kyriediculous: fixed in 6c5b9d4d

eth/client.go
kyriediculous: fixed in 6c90f60

yondonfu:
> For some reason the regular Bond transaction is ~20K gas cheaper, despite the hints being correct. The state update is the same.

Can you investigate why this is the case and confirm if the bondWithHint() tx is more expensive than the bond() tx due to a problem with the hints or some other reason?

From glancing at the Rinkeby txs, it looks like the two txs for bonding were submitted from the same account one after the other around the same time. Could it be the case that the first bondWithHint() tx changed the transcoder pool ordering such that the second bond() tx (without hints) required less iterations in the transcoder pool (or even no iterations at all) and as a result the bondWithHint() tx ended up being more expensive due to the gas costs associated with including/reading the hints on-chain? I bring this up because we should make sure (if we're not doing so already) that when comparing txs with and without hints that the on-chain transcoder pool state is the same before each tx to make it a fair comparison.

kyriediculous:
The `BondWithHint` transaction includes the call for `autoClaimEarnings` 

The regular `Bond` transaction took place in the same round thus skipped `UpdateDelegatorWithEarnings`

I ran these transactions through the tenderly execution tracer and this confirms as much. As you can see in the second image there is no call to `UpdateDelegatorWithEarnings` and `pendingStakeAndFees`. So In reality the BonWithHint function would have been several thousands of gas cheaper if they would have been identical. 

- BondWithHint 

image

- Bond

image

eth/client.go
```diff
@@ -418,7 +419,149 @@ func (c *client) Bond(amount *big.Int, to ethcommon.Address) (*types.Transaction
...
+	// Get current delegate
+	delegator, err := c.GetDelegator(sender)
+	if err != nil {
+		return nil, errors.Wrapf(err, "unable to get delegator")
+	}
+
+	// Switching delegate's calculate old delegate positions
+	var oldHints lpTypes.TranscoderPoolHints
+	if delegator.DelegateAddress != to {
+		oldHints = findTranscoderHints(delegator.DelegateAddress, transcoders)
```
yondonfu: This should be a call to `simulateTranscoderPoolUpdate()` with the new stake (old delegate total stake - pendingStake(delegator)) of the old delegate in order to find the hint for the old delegate's new position in the pool.

yondonfu (changes requested):

eth/client.go
```diff
@@ -431,24 +431,32 @@ func (c *client) Bond(amount *big.Int, to ethcommon.Address) (*types.Transaction
...
 	// Switching delegate's calculate old delegate positions
 	var oldHints lpTypes.TranscoderPoolHints
 	if delegator.DelegateAddress != to {
-		oldHints = findTranscoderHints(delegator.DelegateAddress, transcoders)
+		// Get total bonded
+		totalBonded, err := c.TranscoderTotalStake(delegator.DelegateAddress)
+		if err != nil {
+			return nil, err
+		}
+		oldHints = simulateTranscoderPoolUpdate(delegator.DelegateAddress, new(big.Int).Sub(totalBonded, delegator.BondedAmount), transcoders, isFull)
```
yondonfu: `delegator.BondedAmount` needs to be replaced with a `PendingStake()` call because `delegator.BondedAmount` will not include accumulated rewards until after the those rewards are automatically claimed & credited to `delegator.BondedAmount` at the start of the execution of the bond tx.

eth/client.go
```diff
@@ -418,7 +419,161 @@ func (c *client) Bond(amount *big.Int, to ethcommon.Address) (*types.Transaction
...
+	}
+
+	// Get total bonded
+	totalBonded, err := c.TranscoderTotalStake(to)
+	if err != nil {
+		return nil, err
+	}
+	newStake := totalBonded.Add(totalBonded, amount)
```
yondonfu, line 473: If the delegator is switching delegates we also need to add the delegator's pending stake here since that will be added to the new delegate's stake as well.

eth/client.go
```diff
@@ -418,7 +419,161 @@ func (c *client) Bond(amount *big.Int, to ethcommon.Address) (*types.Transaction
...
+		return nil, errors.Wrapf(err, "unable to get transcoder pool")
+	}
+
+	// Get max pool size
+	maxSize, err := c.GetTranscoderPoolMaxSize()
+	if err != nil {
+		return nil, errors.Wrapf(err, "unable to get transcoder pool max size")
+	}
+
+	totalBonded := delegator.DelegatedAmount
```
yondonfu: We need to use `TranscoderTotalStake(delegator.DelegateAddress)` here to get the delegate's delegated amount (we don't want to use the delegator's delegated amount as is done here).

yondonfu (changes requested):

eth/client.go
yondonfu: Bump. Looks like now we're including the delegator's pending stake, but not including the additional amount if it is non-zero (i.e. if new LPT is staked).

eth/client.go
```diff
@@ -418,7 +419,184 @@ func (c *client) Bond(amount *big.Int, to ethcommon.Address) (*types.Transaction
...
+	}
+
+	transcoderStake, err := c.TranscoderTotalStake(delegator.DelegateAddress)
+	if err != nil {
+		return nil, err
+	}
+
+	isFull := int64(len(transcoders)) == maxSize.Int64()
+
+	newStake := transcoderStake.Add(totalBonded, lock.Amount)
```
yondonfu: We don't need to include the delegator's pending stake here since that is already accounted for in the transcoder's total stake.

yondonfu (changes requested):

eth/client.go
```diff
@@ -418,7 +419,174 @@ func (c *client) Bond(amount *big.Int, to ethcommon.Address) (*types.Transaction
...
+	delegator, err := c.GetDelegator(sender)
+	if err != nil {
+		return nil, errors.Wrapf(err, "unable to get delegator")
+	}
+
+	isFull := int64(len(transcoders)) == maxSize.Int64()
+
+	// Switching delegate's calculate old delegate positions
+	var oldHints lpTypes.TranscoderPoolHints
+	if delegator.DelegateAddress != to {
```
yondonfu: We can add `delegator.DelegateAddress != ethcommon.Address{}` to this conditional to skip the code block below when `Bond()` is called by an account that has not staked yet.

eth/client.go
```diff
@@ -418,7 +419,174 @@ func (c *client) Bond(amount *big.Int, to ethcommon.Address) (*types.Transaction
...
+		if err != nil {
+			return nil, err
+		}
+		amount = new(big.Int).Add(delegatorTotalStake, amount)
+		// Get total bonded
+		totalBonded, err := c.TranscoderTotalStake(delegator.DelegateAddress)
+		if err != nil {
+			return nil, err
+		}
+		oldHints = simulateTranscoderPoolUpdate(delegator.DelegateAddress, new(big.Int).Sub(totalBonded, amount), transcoders, isFull)
```
yondonfu: The old delegate's new stake should be `new(big.Int).Sub(totalBonded, delegatorTotalStake)`.

At the moment, `amount = delegatorTotalStake + amount` where `amount` is additional LPT being staked so `new(big.Int).Sub(totalBonded, amount)` results in subtracting too much from `totalBonded`.

yondonfu (changes requested):

eth/client.go
kyriediculous: fixed in 13e5e37 and added a comment

eth/client.go
kyriediculous: fixed in 13e5e37

yondonfu (approved):
LGTM after squashing

PR #1848, opened 2021-04-20
cmd: use fee cut instead of fee share in the CLI

### LINKED ISSUE #1808: Use fee cut instead of fee share in CLI
The explorer uses fee cut as user facing language even though the protocol uses feeShare (the opposite). 

It would be good that all user facing language is ubiquitous to avoid confusion. 

From discord: 

> AcesHigh — Today at 2:39 AM
I noticed that he explorer shows "Fee Cut" as a column and the command line has a parameter feeShare.  Does feeShare mean the opposite of feeCut - i.e. feeSHare is how much fees the orchestrator will share back to the delegators and feeCut is how much the orchestrator keeps??
In other words, if I set feeShare to 60, Fee Cut will show as 40 on the UI?

**Additional context**

kyriediculous:
Use fee cut instead of fee share for user facing language in the CLI 

Manual test on rinkeby on all affected code paths

Fixes #1808 

**Checklist:**

### DIFF
--- a/cmd/livepeer_cli/wizard_transcoder.go
+++ b/cmd/livepeer_cli/wizard_transcoder.go
@@ -18,6 +18,11 @@ import (
 
 const defaultRPCPort = "8935"
 
+const defaultRewardCut = float64(10)
+const defaultFeeCut = float64(95)
+
+var hundredPercent = eth.FromPerc(100)
+
 func (w *wizard) isOrchestrator() bool {
 	isT := httpGet(fmt.Sprintf("http://%v:%v/IsOrchestrator", w.host, w.httpPort))
 	return isT == "true"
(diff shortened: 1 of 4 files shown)


cmd/livepeer_cli/wizard_transcoder.go
```diff
@@ -41,19 +46,21 @@ func (w *wizard) promptOrchestratorConfig() (float64, float64, int, int, string)
...
-	fmt.Printf("Enter block reward cut percentage (current=%v default=10) - ", blockRewardCut)
+	feeCut := HUNDRED_PERCENT - feeShare
+
+	fmt.Printf("Enter block reward cut percentage (current=%v default=%v) - ", blockRewardCut, defaultRewardCut)
 	blockRewardCut = w.readDefaultFloat(blockRewardCut)
 
-	fmt.Printf("Enter fee share percentage (current=%v default=5) - ", feeShare)
-	feeShare = w.readDefaultFloat(feeShare)
+	fmt.Printf("Enter fee cut percentage (current=%v default=%v) - ", feeCut, defaultFeeCut)
+	feeCut = w.readDefaultFloat(feeCut)
```
yondonfu, line 59: After this function call, feeCut is never used and `promptOrchestratorConfig` will return feeShare which will always be set to the current on-chain value.

cmd/livepeer_cli/wizard_transcoder.go
```diff
@@ -268,3 +275,7 @@ func (w *wizard) showVoteChoices() {
 	}
 	wtr.Flush()
 }
+
+func feeShareToFeeCutString(feeShare *big.Int) string {
```
yondonfu: What do you think about changing this something like `flipPerc()` and returning the `big.Int` representation 100% - X% where X is the argument to the function? Then, this function could be used anywhere where the `big.Int` representation of the fee cut is required and we could still convert that value into a float64 or string if needed as opposed to right now we'd go from big.Int -> big.Int -> string -> big.Int/float64 if we ever needed the value in a different format than a string.

For example:

```golang
...
} else {
    blockRewardCut = eth.ToPerc(orch.RewardCut)
    feeCut = eth.ToPerc(flipPerc(orch.FeeShare))
}

...

feeCut = readDefaultFloat(feeCut)

...

return blockRewardCut, HUNDRED_PERCENT - feeCut, pricePerUnit, pixelsPerUnit, serviceURI
```

And then anywhere we need to print the fee cut we could use use `eth.FormatPerc(flipPerc(orch.FeeShare))`

yondonfu (changes requested):

yondonfu (approved):
LGTM

PR #1849, opened 2021-04-21
core: reuse remote transcoder for stream

### LINKED ISSUE #1273: Make RemoteTranscoderManager sticky for LB
Currently `RemoteTranscoderManager` distributes work evenly between all connected transcoders.
For LB we should change that so `RemoteTranscoderManager` sends segments for the same stream to the same transcoder.

Relates to #1108

### LINKED ISSUE #1712: RemoteTranscoderManager should stick with transcoder as long as conditions are good
This issue pertains to running an O/T split setup with remote GPU transcoders. 

Currently for each segment from an incoming stream, the `RemoteTranscodeManager` will pick a _new_ transcoder each segment. This is not ideal as it will result in GPU session startup overhead each time a transcoder gets re-selected after a few segments. 

Under ideal circumstances the stream is transcoded by the same remote transcoder until there's a negative outcome: error, timeout, invalid result, ...  .

Add a `Sessions` concept to the `RemoteTranscoderManager`. A `Session` keeps track of the current `Transcoder` to be used for each ongoing stream `ManifestID`. 

This can be tracked in a mapping `map[manifestID]*remoteTranscoder`.

When a `remoteTranscoder` is assigned a `ManifestID` its `load` will increase by 1. 
When a `remoteTranscoder` is removed for a `ManifestID` (either due to bad performance, error or the segment loop for the `manifestID` timing out) its load will be reduced by 1. 

If there is an existing session whenever a `Transcode` request for a segment comes in we will use the `remoteTranscoder` currently in the map rather than call `selectTranscoder()` for each incoming request. 

Other than trying to improve O/T performance. This new per-stream rather than per-segment behaviour also opens the door for more suitable selection algorithms such as latency based selection. 

The `selectTranscoder` and `completeTranscoders` functions can be re-used to manage sessions instead. 
The biggest change/addition would be a way for the `RemoteTranscoderManager` to listen for segment loop closures (or run its own cleanup loop).

### LINKED ISSUE #1842: Re-use remote transcoders for a stream
## Abstract

The split orchestrator and transcoder architecture enables the creation of transcoding pools. These pools can either be private or public.

When an orchestrator receives a stream it will currently send each segment for a `sessionID` to a different transcoder. 

Analogous to how a broadcaster sticks with the same orchestrator for all segments of a stream as long as it can transcode in real-time, an orchestrator should stick with the same transcoder for a stream. 

## Motivation

Sending each segment of a stream to a different transcoder could lead to a degradation of performance: 
- Transcoders can differ in performance (speed, quality, ...)
- Harder to re-use GPU sessions
- Fanning out on a per-stream basis can negatively impact work retention

### Current Flow 

1. Segment comes in

2. The list of connected `RemoteTranscoders` is sorted by `load`

    *Each segment in flight for a `RemoteTranscoder` increments `load` by 1*
    
3. Select the head of the sorted `RemoteTranscoder` list 

    *If the `RemoteTranscoder` is no longer connected, discard it and continue*
    
4. Send segment to `RemoteTranscoder`

5. Decrement `load` for `RemoteTranscoder` upon return

Since ordering of the `RemoteTranscoder` list is not guaranteed at any point in time selection is entirely random.

### Proposed Flow 

1. New segment comes in for stream with `sessionID`

2. Check if there is a `RemoteTranscoder` assigned to `sessionID`
    - No assignment - proceed to step 3
    - Transcoder assigned - proceed to step 4

3. The list of connected `RemoteTranscoders` is sorted by `load`, select the head of the list and increment its `load`. 

    *`load` and `capacity` will be defined on a per-stream basis rather than a per-segment basis. Each stream session for a `RemoteTranscoder` increments `load` by 1*
    
4. Check whether the (newly) assigned `RemoteTranscoder` is still connected
    - connected - proceed to step 5
    - not connected - go back to step 3 

    *If the `RemoteTranscoder` is no longer connected, remove it from the list, remove it from the assignment and decrement its `load`*

5. Update `RemoteTranscoderManager` state to reflect the assigned `RemoteTranscoder` for `sessionID` (if necessary)

6. Send segment to `RemoteTranscoder`
    - success: no-op
    - error: remove it from the list, remove it from the assignment and decrement its `load`

One potential way of implementing this could be as follows: 

Add a map to the `RemoteTranscoderManager` to keep track of the assigned `RemoteTranscoder` for an `sessionID`. 

```go
type RemoteTranscoderManager struct {
    ...
    StreamSessions: map[sessionID]*RemoteTranscoder
}
```

`RemoteTranscoderManager.selectTranscoder()` can be updated to take a sessionID as argument. 

```go
func(rtm *RemoteTranscoderManager) selectTranscoder(string sessionID) *RemoteTranscoder {
    rt, ok := rtm.StreamSessions[sessionID]
    if !ok {
        // ... current selection logic
        currentTranscoder.load++
        sort.Sort(byLoadFactor(rtm.remoteTranscoders))
        // Assign currentTranscoder
        rt = currentTranscoder
	}
    }
    
    return rt
}
```

`RemoteTranscoder.completeTranscoder()` would also have to take `sessionID` as an argument and be tied to the transcode session cleanup loop, or a separate cleanup loop. 

- A transcoder should be re-used for a stream with `sessionID` as long as no error or timeout occurs.

reubenr0d:

- Adds map to keep track of transcoders assigned to sessions
- Updated selectTranscoder to return and set transcoders to sessions
- Removed transcoder from map in case of session end/transcoder errors

On local offchain setup:
- Setup 1 O, 2 T, 1 B
- Start stream and observed that only one transcoder is processing stream
- Killed that transcoder and stream moves to other transcoder
- Created Unit Tests

Fixes #1842 
Fixes #1712 
Fixes #1273 

**Checklist:**

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -17,6 +17,7 @@
 
 - \#1845 Staking actions with hints (@kyriediculous)
 - \#1873 Increase TicketParams expiration to 10 blocks (@kyriediculous)
+- \#1849 Re-use remote transcoders for a stream sessions (@reubenr0d)
 
 #### Transcoder
 
--- a/core/orch_test.go
+++ b/core/orch_test.go
@@ -9,6 +9,7 @@ import (
 	"math/big"
 	"math/rand"
 	"os"
+	"strconv"
 	"sync"
 	"testing"
 	"time"
@@ -290,7 +291,7 @@ func TestSelectTranscoder(t *testing.T) {
 
 	// register transcoders, which adds transcoder to liveTranscoders and remoteTranscoders
 	wg := newWg(1)
-	go func() { m.Manage(strm, 2) }()
+	go func() { m.Manage(strm, 1) }()
 	time.Sleep(1 * time.Millisecond) // allow time for first stream to register
 	go func() { m.Manage(strm2, 1); wg.Done() }()
 	time.Sleep(1 * time.Millisecond) // allow time for second stream to register
@@ -299,30 +300,44 @@ func TestSelectTranscoder(t *testing.T) {
 	assert.NotNil(m.liveTranscoders[strm2])
 	assert.Len(m.remoteTranscoders, 2)
 
+	testSessionId := "testID"
+	testSessionId2 := "testID2"
+	testSessionId3 := "testID3"
+
 	// assert transcoder is returned from selectTranscoder
 	t1 := m.liveTranscoders[strm]
 	t2 := m.liveTranscoders[strm2]
-	currentTranscoder := m.selectTranscoder()
+	currentTranscoder, err := m.selectTranscoder(testSessionId)
+	assert.Nil(err)
 	assert.Equal(t2, currentTranscoder)
 	assert.Equal(1, t2.load)
 	assert.NotNil(m.liveTranscoders[strm])
 	assert.Len(m.remoteTranscoders, 2)
 
-	// assert transcoder with less load selected
-	currentTranscoder2 := m.selectTranscoder()
-	assert.Equal(t1, currentTranscoder2)
+	// assert that same transcoder is selected for same sessionId
+	// and that load stays the same
+	currentTranscoder, err = m.selectTranscoder(testSessionId)
+	assert.Nil(err)
+	assert.Equal(t2, currentTranscoder)
+	assert.Equal(1, t2.load)
+	m.completeStreamSession(testSessionId)
+
+	// assert that a new transcoder is selected for a new sessionId
+	currentTranscoder, err = m.selectTranscoder(testSessionId2)
+	assert.Nil(err)
+	assert.Equal(t1, currentTranscoder)
 	assert.Equal(1, t1.load)
 
-	currentTranscoder3 := m.selectTranscoder()
-	assert.Equal(t1, currentTranscoder3)
-	assert.Equal(2, t1.load)
-
-	// assert no transcoder returned if all at they capacity
-	noTrans := m.selectTranscoder()
+	// Add some more load and assert no transcoder returned if all at capacity
+	currentTranscoder, err = m.selectTranscoder(testSessionId)
+	assert.Nil(err)
+	assert.Equal(t2, currentTranscoder)
+	noTrans, err := m.selectTranscoder(testSessionId3)
+	assert.Equal(err, ErrNoTranscodersAvailable)
 	assert.Nil(noTrans)
 
-	m.completeTranscoders(t1)
-	m.completeTranscoders(t1)
+	// assert that load is empty after ending stream sessions
+	m.completeStreamSession(testSessionId2)
 	assert.Equal(0, t1.load)
 
 	// unregister transcoder
@@ -331,25 +346,94 @@ func TestSelectTranscoder(t *testing.T) {
 	assert.Nil(m.liveTranscoders[strm2])
 	assert.NotNil(m.liveTranscoders[strm])
 
-	// assert t1 is selected and t2 drained
-	currentTranscoder = m.selectTranscoder()
+	// assert t1 is selected and t2 drained, but was previously selected
+	currentTranscoder, err = m.selectTranscoder(testSessionId)
+	assert.Nil(err)
 	assert.Equal(t1, currentTranscoder)
 	assert.Equal(1, t1.load)
 	assert.NotNil(m.liveTranscoders[strm])
-	assert.Len(m.remoteTranscoders, 2)
+	assert.Len(m.remoteTranscoders, 1)
 
 	// assert transcoder gets added back to remoteTranscoders if no transcoding error
-	_, err := m.Transcode(&SegTranscodingMetadata{})
+	transcodedData, err := m.Transcode(&SegTranscodingMetadata{AuthToken: &net.AuthToken{SessionId: testSessionId}})
+	assert.NotNil(transcodedData)
 	assert.Nil(err)
-	assert.Len(m.remoteTranscoders, 2)
+	assert.Len(m.remoteTranscoders, 1)
 	assert.Equal(1, t1.load)
-	m.completeTranscoders(t1)
+	m.completeStreamSession(testSessionId)
 	assert.Equal(0, t1.load)
+}
+
+func TestCompleteStreamSession(t *testing.T) {
+	m := NewRemoteTranscoderManager()
+	strm := &StubTranscoderServer{manager: m}
+	testSessionId := "testID"
+	assert := assert.New(t)
+
+	// register transcoders
+	go func() { m.Manage(strm, 1) }()
+	time.Sleep(1 * time.Millisecond) // allow time for first stream to register
+	t1 := m.liveTranscoders[strm]
+
+	// selectTranscoder and assert that session is added
+	m.selectTranscoder(testSessionId)
+	assert.Equal(t1, m.streamSessions[testSessionId])
+	assert.Equal(1, t1.load)
+
+	// complete session and assert that it is cleared
+	m.completeStreamSession(testSessionId)
+	transcoder, ok := m.streamSessions[testSessionId]
+	assert.Nil(transcoder)
+	assert.False(ok)
+	assert.Equal(0, t1.load)
+}
+
+func TestRemoveFromRemoteTranscoders(t *testing.T) {
+	remoteTranscoderList := []*RemoteTranscoder{}
+	assert := assert.New(t)
+
+	// Create 4 tanscoders
+	tr := make([]*RemoteTranscoder, 4)
+	for i := 0; i < 4; i++ {
+		tr[i] = &RemoteTranscoder{addr: "testAddress" + strconv.Itoa(i)}
+	}
+
+	// Add to list
+	remoteTranscoderList = append(remoteTranscoderList, tr...)
+	assert.Len(remoteTranscoderList, 4)
+
+	// Remove transcoder froms head of the list
+	remoteTranscoderList = removeFromRemoteTranscoders(tr[0], remoteTranscoderList)
+	assert.Equal(remoteTranscoderList[0], tr[1])
+	assert.Equal(remoteTranscoderList[1], tr[2])
+	assert.Equal(remoteTranscoderList[2], tr[3])
+	assert.Len(remoteTranscoderList, 3)
+
+	// Remove transcoder from the middle of the list
+	remoteTranscoderList = removeFromRemoteTranscoders(tr[3], remoteTranscoderList)
+	assert.Equal(remoteTranscoderList[0], tr[1])
+	assert.Equal(remoteTranscoderList[1], tr[2])
+	assert.Len(remoteTranscoderList, 2)
+
+	// Remove transcoder from the end of the list
+	remoteTranscoderList = removeFromRemoteTranscoders(tr[2], remoteTranscoderList)
+	assert.Equal(remoteTranscoderList[0], tr[1])
+	assert.Len(remoteTranscoderList, 1)
+
+	// Remove the last transcoder
+	remoteTranscoderList = removeFromRemoteTranscoders(tr[1], remoteTranscoderList)
+	assert.Len(remoteTranscoderList, 0)
+
+	// Remove a transcoder when list is empty
+	remoteTranscoderList = removeFromRemoteTranscoders(tr[1], remoteTranscoderList)
+	emptyTList := []*RemoteTranscoder{}
+	assert.Equal(remoteTranscoderList, emptyTList)
 }
 
 func TestTranscoderManagerTranscoding(t *testing.T) {
 	m := NewRemoteTranscoderManager()
 	s := &StubTranscoderServer{manager: m}
+	testSessionId := "testID"
 
 	// sanity checks
 	assert := assert.New(t)
@@ -357,9 +441,10 @@ func TestTranscoderManagerTranscoding(t *testing.T) {
 	assert.Empty(m.remoteTranscoders)
 
 	// Attempt to transcode when no transcoders in the set
-	_, err := m.Transcode(&SegTranscodingMetadata{})
+	transcodedData, err := m.Transcode(&SegTranscodingMetadata{AuthToken: &net.AuthToken{SessionId: testSessionId}})
+	assert.Nil(transcodedData)
 	assert.NotNil(err)
-	assert.Equal(err.Error(), "No transcoders available")
+	assert.Equal(err, ErrNoTranscodersAvailable)
 
 	wg := newWg(1)
 	go func() { m.Manage(s, 5); wg.Done() }()
@@ -370,14 +455,15 @@ func TestTranscoderManagerTranscoding(t *testing.T) {
 	assert.NotNil(m.liveTranscoders[s])
 
 	// happy path
-	res, err := m.Transcode(&SegTranscodingMetadata{})
+	res, err := m.Transcode(&SegTranscodingMetadata{AuthToken: &net.AuthToken{SessionId: testSessionId}})
 	assert.Nil(err)
 	assert.Len(res.Segments, 1)
 	assert.Equal(string(res.Segments[0].Data), "asdf")
 
 	// non-fatal error should not remove from list
 	s.TranscodeError = fmt.Errorf("TranscodeError")
-	_, err = m.Transcode(&SegTranscodingMetadata{})
+	transcodedData, err = m.Transcode(&SegTranscodingMetadata{AuthToken: &net.AuthToken{SessionId: testSessionId}})
+	assert.NotNil(transcodedData)
 	assert.Equal(s.TranscodeError, err)
 	assert.Len(m.remoteTranscoders, 1)           // sanity
 	assert.Equal(0, m.remoteTranscoders[0].load) // sanity
@@ -387,13 +473,15 @@ func TestTranscoderManagerTranscoding(t *testing.T) {
 
 	// fatal error should retry and remove from list
 	s.SendError = fmt.Errorf("SendError")
-	_, err = m.Transcode(&SegTranscodingMetadata{})
+	transcodedData, err = m.Transcode(&SegTranscodingMetadata{AuthToken: &net.AuthToken{SessionId: testSessionId}})
 	assert.True(wgWait(wg)) // should disconnect manager
+	assert.Nil(transcodedData)
 	assert.NotNil(err)
-	assert.Equal(err.Error(), "No transcoders available")
-	_, err = m.Transcode(&SegTranscodingMetadata{}) // need second try to remove from remoteTranscoders
+	assert.Equal(err, ErrNoTranscodersAvailable)
+	transcodedData, err = m.Transcode(&SegTranscodingMetadata{AuthToken: &net.AuthToken{SessionId: testSessionId}}) // need second try to remove from remoteTranscoders
+	assert.Nil(transcodedData)
 	assert.NotNil(err)
-	assert.Equal(err.Error(), "No transcoders available")
+	assert.Equal(err, ErrNoTranscodersAvailable)
 	assert.Len(m.liveTranscoders, 0)
 	assert.Len(m.remoteTranscoders, 0) // retries drain the list
 	s.SendError = nil
@@ -409,7 +497,8 @@ func TestTranscoderManagerTranscoding(t *testing.T) {
 	oldTimeout := common.HTTPTimeout
 	common.HTTPTimeout = 1 * time.Millisecond
 	defer func() { common.HTTPTimeout = oldTimeout }()
-	_, err = m.Transcode(&SegTranscodingMetadata{})
+	transcodedData, err = m.Transcode(&SegTranscodingMetadata{AuthToken: &net.AuthToken{SessionId: testSessionId}})
+	assert.Nil(transcodedData)
 	_, fatal := err.(RemoteTranscoderFatalError)
 	wg.Wait()
 	assert.True(fatal)
--- a/core/orchestrator.go
+++ b/core/orchestrator.go
@@ -628,6 +628,12 @@ func (n *LivepeerNode) transcodeSegmentLoop(md *SegTranscodingMetadata, segChan 
 				// timeout; clean up goroutine here
 				os.EndSession()
 				los.EndSession()
+				// check to avoid nil pointer caused by garbage collection while this go routine is still running
+				if n.TranscoderManager != nil {
+					n.TranscoderManager.RTmutex.Lock()
+					n.TranscoderManager.completeStreamSession(md.AuthToken.SessionId)
+					n.TranscoderManager.RTmutex.Unlock()
+				}
 				glog.V(common.DEBUG).Infof("Segment loop timed out; closing manifestID=%s sessionID=%s", md.ManifestID, md.AuthToken.SessionId)
 				n.segmentMutex.Lock()
 				mid := ManifestID(md.AuthToken.SessionId)
@@ -684,6 +690,7 @@ func NewRemoteTranscoderFatalError(err error) error {
 }
 
 var ErrRemoteTranscoderTimeout = errors.New("Remote transcoder took too long")
+var ErrNoTranscodersAvailable = errors.New("no transcoders available")
 
 func (rt *RemoteTranscoder) done() {
 	// select so we don't block indefinitely if there's no listener
@@ -760,10 +767,12 @@ func NewRemoteTranscoderManager() *RemoteTranscoderManager {
 	return &RemoteTranscoderManager{
 		remoteTranscoders: []*RemoteTranscoder{},
 		liveTranscoders:   map[net.Transcoder_RegisterTranscoderServer]*RemoteTranscoder{},
-		RTmutex:           &sync.Mutex{},
+		RTmutex:           sync.Mutex{},
 
 		taskMutex: &sync.RWMutex{},
 		taskChans: make(map[int64]TranscoderChan),
+
+		streamSessions: make(map[string]*RemoteTranscoder),
 	}
 }
 
@@ -782,12 +791,15 @@ func (r byLoadFactor) Less(i, j int) bool {
 type RemoteTranscoderManager struct {
 	remoteTranscoders []*RemoteTranscoder
 	liveTranscoders   map[net.Transcoder_RegisterTranscoderServer]*RemoteTranscoder
-	RTmutex           *sync.Mutex
+	RTmutex           sync.Mutex
 
 	// For tracking tasks assigned to remote transcoders
 	taskMutex *sync.RWMutex
 	taskChans map[int64]TranscoderChan
 	taskCount int64
+
+	//Map for keeping track of sessions and their respective transcoders
+	streamSessions map[string]*RemoteTranscoder
 }
 
 // RegisteredTranscodersCount returns number of registered transcoders
@@ -847,7 +859,33 @@ func (rtm *RemoteTranscoderManager) Manage(stream net.Transcoder_RegisterTransco
 	}
 }
 
-func (rtm *RemoteTranscoderManager) selectTranscoder() *RemoteTranscoder {
+func removeFromRemoteTranscoders(rt *RemoteTranscoder, remoteTranscoders []*RemoteTranscoder) []*RemoteTranscoder {
+	if len(remoteTranscoders) == 0 {
+		// No transocerds to remove, return
+		return remoteTranscoders
+	}
+
+	lastIndex := len(remoteTranscoders) - 1
+	last := remoteTranscoders[lastIndex]
+	if rt == last {
+		return remoteTranscoders[:lastIndex]
+	}
+
+	newRemoteTs := make([]*RemoteTranscoder, 0)
+	for i, t := range remoteTranscoders {
+		if t == rt {
+			if i == 0 {
+				return remoteTranscoders[1:]
+			}
+			newRemoteTs = remoteTranscoders[i-1 : i]
+			newRemoteTs = append(newRemoteTs, remoteTranscoders[i+1:]...)
+			break
+		}
+	}
+	return newRemoteTs
+}
+
+func (rtm *RemoteTranscoderManager) selectTranscoder(sessionId string) (*RemoteTranscoder, error) {
 	rtm.RTmutex.Lock()
 	defer rtm.RTmutex.Unlock()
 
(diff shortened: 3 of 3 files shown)


core/orchestrator.go
```diff
@@ -628,6 +628,7 @@ func (n *LivepeerNode) transcodeSegmentLoop(md *SegTranscodingMetadata, segChan
 				// timeout; clean up goroutine here
 				os.EndSession()
 				los.EndSession()
+				delete(n.TranscoderManager.streamSessions, md.AuthToken.SessionId)
```
kyriediculous: There can be multiple segments (video chunks) in flight that access this map for a particular key, ideally we thus want this to be concurrency safe. 

Let's wrap this `delete` statement in a method that also locks and unlocks the resource. 

```go
func (rtm *RemoteTranscoderManager) deleteSession(sesssionId string) {
  rtm.RTmutex.Lock()
  defer rtm.RTmutex.Unlock()
  delete(n.TranscoderManager.streamSessions, md.AuthToken.SessionId)
}
```

We can then do `n.TranscoderManager.deleteSession(md.AuthToken.SessionId)`

core/orchestrator.go
```diff
@@ -856,16 +862,23 @@ func (rtm *RemoteTranscoderManager) selectTranscoder() *RemoteTranscoder {
...
-			// transcoder does not exist in table; remove and retry
-			rtm.remoteTranscoders = rtm.remoteTranscoders[:last]
-			continue
-		}
-		if currentTranscoder.load == currentTranscoder.capacity {
-			// Head of queue is at capacity, so the rest must be too. Exit early
-			return nil
+		currentTranscoder, transcoderExists := rtm.streamSessions[sessionId]
+		// Assign new transcoder if not assigned, or at full load
+		if !transcoderExists || currentTranscoder.load == currentTranscoder.capacity {
```
kyriediculous: While this is logically correct we can avoid this OR statement by splitting the IF statement.

Also WDYT about rewording `transcoderExists` to something like `sessionExists` to avoid confusion with the comment on L8714where we check whether a transcoder has an open gRPC stream with the Orchestrator.

```go
if sessionExists {
  return currentTranscoder
}

last := len(rtm.remoteTranscoders) - 1
// ... rest of the logic
```

core/orchestrator.go
```diff
@@ -899,7 +912,7 @@ func (rtm *RemoteTranscoderManager) totalLoadAndCapacity() (int, int, int) {
 
 // Transcode does actual transcoding using remote transcoder from the pool
 func (rtm *RemoteTranscoderManager) Transcode(md *SegTranscodingMetadata) (*TranscodeData, error) {
-	currentTranscoder := rtm.selectTranscoder()
+	currentTranscoder := rtm.selectTranscoder(md.AuthToken.SessionId)
```
kyriediculous: WDYT about returning `errors.New("No transcoders available")` instead of returning `nil` from `selectTranscoder()` and handling the error case instead of the nil case?  Feels a little bit cleaner and we have the opportunity to do that now. 

Can also turn it into a constant since the error message doesn't take arguments. 

`var errNoTranscodersAvailable = errors.New("No transcoders available")`

kyriediculous:
We also want to remove the currentTranscoder from the `streamSessions` map when it returns an error here https://github.com/livepeer/go-livepeer/blob/7a0fd1f6b958d61177fc01c488972f54b8bad1e3/core/orchestrator.go#L906

kyriediculous:
We also want to remove the currentTranscoder from the `streamSessions` map when the connection closes like we do here https://github.com/livepeer/go-livepeer/blob/7a0fd1f6b958d61177fc01c488972f54b8bad1e3/core/orchestrator.go#L840

core/orchestrator.go
kyriediculous: Missed something in my earlier comment, my bad. 

If we do 

```
if sessionExists { return currentTranscoder }
```

We need to remove the currentTranscoder from the `streamSessions` if its connection terminates for some reason before hitting this code path. Otherwise we will try to transcode against a closed connection which will error out and only then select a new transcoder. 

So we also want to implement https://github.com/livepeer/go-livepeer/pull/1849#issuecomment-824731915 and https://github.com/livepeer/go-livepeer/pull/1849#issuecomment-824737313

core/orchestrator.go
kyriediculous: On second thought let's repurpose https://github.com/livepeer/go-livepeer/blob/7a0fd1f6b958d61177fc01c488972f54b8bad1e3/core/orchestrator.go#L878 

```go
func (rtm *RemoteTranscoderManager) completeStreamSession(sessionId string) {
	rtm.RTmutex.Lock()
	defer rtm.RTmutex.Unlock()
	
	t, ok := rtm.streamSessions[sessionId]
	if !ok {
		return
	}
	t.load--
	sort.Sort(byLoadFactor(rtm.remoteTranscoders))
	delete(rtm.streamSessions, sessionId)
}
```

We want to call this function in the following cases: 

- When connection closes https://github.com/livepeer/go-livepeer/pull/1849#issuecomment-824737313
- When the current transcoder returns an error transcoding https://github.com/livepeer/go-livepeer/pull/1849#issuecomment-824731915
- When the segment loop times out (L631 here)


We also want to remove the call here https://github.com/livepeer/go-livepeer/blob/7a0fd1f6b958d61177fc01c488972f54b8bad1e3/core/orchestrator.go#L916

reubenr0d: Thanks for reviewing, @kyriediculous. Will take care of this, I was almost certain that the cleanup would have to be handled in other places as well.

core/orchestrator.go
reubenr0d: If I understand right, we would need to increment load and re-sort the transcoders even if we are using a transcoder that had been previously assigned. So I guess we would have to repeat those statements (which are currently outside the condition) if we do the early return. Please correct me incase I am wrong here, how do you think we should proceed?

kyriediculous: Currently load is incremented and decremented whenever a segment completes (regardless from which stream it is). 

In this PR we are changing this behaviour to only increment when a transcoder is assigned a stream (tracked by sessionId in the sessionsMap) so we only have to decrement load when it gets un-assigned.

(The added bonus of that is that -maxSessions behaviour is now the same for the O and the T) 

We can do this easily by repurposing `completeTranscoder` in this comment https://github.com/livepeer/go-livepeer/pull/1849#discussion_r618294959

core/orchestrator.go
reubenr0d: @kyriediculous, I have pushed the changes suggested here and above as well, could you please review how the error handling is now done here, is this okay? I'm not quite sure. Leaving everything unresolved so you can verify if this addresses everything and resolve them.

I am yet to look into the test related nil pointer error that occurs on the last test [on this line](https://github.com/livepeer/go-livepeer/pull/1849/files#diff-9734c79b7138aa52372c8c9714be7aba54b219fb23099883e1533f44d9f79385R350).

core/orchestrator.go
```diff
@@ -899,20 +923,20 @@ func (rtm *RemoteTranscoderManager) totalLoadAndCapacity() (int, int, int) {
 
 // Transcode does actual transcoding using remote transcoder from the pool
 func (rtm *RemoteTranscoderManager) Transcode(md *SegTranscodingMetadata) (*TranscodeData, error) {
-	currentTranscoder := rtm.selectTranscoder()
-	if currentTranscoder == nil {
-		return nil, errors.New("No transcoders available")
+	currentTranscoder, err := rtm.selectTranscoder(md.AuthToken.SessionId)
+	if err == ErrNoTranscodersAvailable {
```
kyriediculous: We can shorten this to `if err != nil { return err }`

core/core_test.go
```diff
@@ -140,7 +140,7 @@ func TestTranscodeLoop_GivenNoSegmentsPastTimeout_CleansSegmentChan(t *testing.T
 	ss := StubSegment()
 	md := &SegTranscodingMetadata{Profiles: videoProfiles, AuthToken: stubAuthToken()}
 	n.Transcoder = NewLocalTranscoder(tmp)
-
+	n.TranscoderManager = NewRemoteTranscoderManager()
```
kyriediculous: We probably want to test this separately from using a LocalTranscoder (O and T combined)

core/orch_test.go
```diff
@@ -299,30 +299,37 @@ func TestSelectTranscoder(t *testing.T) {
...
 
+	testSessionId := "testID"
+	testSessionId2 := "testID2"
+	testSessionId3 := "testID3"
+
 	// assert transcoder is returned from selectTranscoder
 	t1 := m.liveTranscoders[strm]
 	t2 := m.liveTranscoders[strm2]
-	currentTranscoder := m.selectTranscoder()
+	currentTranscoder, _ := m.selectTranscoder(testSessionId)
```
kyriediculous: Let's assert the error to be nil, same comment for subsequent lines in this test where we call selectTranscoder

core/orch_test.go
```diff
@@ -387,13 +417,13 @@ func TestTranscoderManagerTranscoding(t *testing.T) {
 
 	// fatal error should retry and remove from list
 	s.SendError = fmt.Errorf("SendError")
-	_, err = m.Transcode(&SegTranscodingMetadata{})
+	_, err = m.Transcode(&SegTranscodingMetadata{AuthToken: &net.AuthToken{SessionId: testSessionId}})
```
kyriediculous: Let's also assert the first return value to be nil

core/orch_test.go
```diff
@@ -387,13 +417,13 @@ func TestTranscoderManagerTranscoding(t *testing.T) {
...
 	assert.True(wgWait(wg)) // should disconnect manager
 	assert.NotNil(err)
-	assert.Equal(err.Error(), "No transcoders available")
-	_, err = m.Transcode(&SegTranscodingMetadata{}) // need second try to remove from remoteTranscoders
+	assert.Equal(err.Error(), "no transcoders available")
```
kyriediculous: Just a suggestion: there is also assert.EqualError(error, expected message) 

We can also use the variable we created for this specific error : ErrNoTranscodersAvailable

core/orchestrator.go
```diff
@@ -628,6 +628,9 @@ func (n *LivepeerNode) transcodeSegmentLoop(md *SegTranscodingMetadata, segChan
 				// timeout; clean up goroutine here
 				os.EndSession()
 				los.EndSession()
+				if n.TranscoderManager != nil {
```
kyriediculous, line 632: Let's add a comment that this check is to avoid a nil pointer error when TranscoderManager is garbage collected while this go routine is still running.

core/orchestrator.go
```diff
@@ -842,12 +851,19 @@ func (rtm *RemoteTranscoderManager) Manage(stream net.Transcoder_RegisterTransco
 		totalLoad, totalCapacity, liveTranscodersNum = rtm.totalLoadAndCapacity()
 	}
 	rtm.RTmutex.Unlock()
+
+	// Remove all streamSessions assigned to this transcoder
+	for sessionId, assignedTranscoder := range rtm.streamSessions {
```
kyriediculous: Good call to loop here ! Forgot that we needed to clear all sessions.

kyriediculous (changes requested):

core/core_test.go
reubenr0d: Oh, I had added this in hopes of avoiding the garbage collection related nil pointer, but since we have the check there don't think this is needed. Removing the statement, the other suggestions were great as well, will add them in. Thanks.

core/orch_test.go
reubenr0d: That would return the transcoded data correct? I think you meant NotNil? Adding it in.

core/orch_test.go
```diff
@@ -290,7 +290,7 @@ func TestSelectTranscoder(t *testing.T) {
 
 	// register transcoders, which adds transcoder to liveTranscoders and remoteTranscoders
 	wg := newWg(1)
-	go func() { m.Manage(strm, 2) }()
+	go func() { m.Manage(strm, 1) }()
```
kyriediculous, line 294: The second argument is the transcoder's capacity. Any particular reason for this change, if I'm not mistaken it doesn't affect this test.

core/orch_test.go
```diff
@@ -299,30 +299,40 @@ func TestSelectTranscoder(t *testing.T) {
...
+	// assert that a new transcoder is selected for a new sessionId
+	currentTranscoder, err = m.selectTranscoder(testSessionId2)
+	assert.Nil(err)
+	assert.Equal(t1, currentTranscoder)
+	assert.Equal(1, t1.load)
 
 	// assert no transcoder returned if all at they capacity
-	noTrans := m.selectTranscoder()
+	noTrans, err := m.selectTranscoder(testSessionId3)
+	assert.Equal(err, errors.New("no transcoders available"))
```
kyriediculous: We can use `ErrNoTranscodersAvailable` here instead of `errors.New("no transcoders available"))`

core/orch_test.go
```diff
@@ -331,35 +341,61 @@ func TestSelectTranscoder(t *testing.T) {
...
 	assert.Len(m.remoteTranscoders, 2)
 	assert.Equal(1, t1.load)
-	m.completeTranscoders(t1)
+	m.completeStreamSession(testSessionId)
+	assert.Equal(0, t1.load)
+}
+
+func TestCompleteStreamSession(t *testing.T) {
+	m := NewRemoteTranscoderManager()
+	strm := &StubTranscoderServer{manager: m, WithholdResults: false}
```
kyriediculous: I don't think we need `WithholdResults` here ?

core/orch_test.go
```diff
@@ -331,35 +341,61 @@ func TestSelectTranscoder(t *testing.T) {
...
+	t1 := m.liveTranscoders[strm]
+
+	// selectTranscoder and assert that session is added
+	m.selectTranscoder(testSessionId)
+	assert.Equal(t1, m.streamSessions[testSessionId])
+	assert.Equal(1, t1.load)
+
+	// complete session and assert that it is cleared
+	m.completeStreamSession(testSessionId)
+	assert.Nil(m.streamSessions[testSessionId])
```
kyriediculous: In golang the idiomatic way of checking whether a key exists in a map is using the "comma ok" idiom

i.e. `transcoder, ok := m.streamSessions[sessionId]` 

Where the second argument is a boolean value indicating whether the _key_ exists in the map. 

A key could explicitly be assigned `nil` (perhaps not in this case) in which case `ok` would evaluate to `true`. 

Let's add an "ok" check to be false here for good practice.

core/orch_test.go
```diff
@@ -387,13 +424,15 @@ func TestTranscoderManagerTranscoding(t *testing.T) {
...
+	assert.Nil(transcodedData)
 	assert.NotNil(err)
-	assert.Equal(err.Error(), "No transcoders available")
-	_, err = m.Transcode(&SegTranscodingMetadata{}) // need second try to remove from remoteTranscoders
+	assert.Equal(err.Error(), "no transcoders available")
+	transcodedData, err = m.Transcode(&SegTranscodingMetadata{AuthToken: &net.AuthToken{SessionId: testSessionId}}) // need second try to remove from remoteTranscoders
+	assert.Nil(transcodedData)
 	assert.NotNil(err)
-	assert.Equal(err.Error(), "No transcoders available")
+	assert.Equal(err.Error(), "no transcoders available")
```
kyriediculous: Can use `ErrNoTranscodersAvailable` here

core/orch_test.go
reubenr0d: We are asserting that no transcoders are returned if all are at capacity [here](https://github.com/livepeer/go-livepeer/pull/1849/files/b248bef1940018c77b7b4e4d0a317865484a6c0c#diff-9734c79b7138aa52372c8c9714be7aba54b219fb23099883e1533f44d9f79385R331), which is why it was changed to capacity 1. 
But I just realized that [this test](https://github.com/livepeer/go-livepeer/pull/1849/files/b248bef1940018c77b7b4e4d0a317865484a6c0c#diff-9734c79b7138aa52372c8c9714be7aba54b219fb23099883e1533f44d9f79385R331) can also pass because the previous transcoder is at full load, and not particularly because it is a new session. Maybe I should change the capacity of the transcoder initialized after this one and then add another selectTranscoder with a new session before asserting for "no transcoders available". What do you think?

kyriediculous (changes requested):

core/orch_test.go
reubenr0d: Still checking what exactly that does (had copied it from the previous test), but let me remove it. The other suggestions look good. Pushing shortly.

kyriediculous: It determines whether it should return the results to the remote transcoder manager or an error I believe, but we're not actually transcoding in this test.

core/orch_test.go
kyriediculous: Oh I see, it mimics the behaviour of the old test but with the updated behaviour of the actual code, gotcha. That's fine as is, however if the tests checks for a _new_ transcoder we could simply delete the current session before we do that part of the test.

reubenr0d: Yup. Just pushed a change to the previous commit, I deleted the current session before that test, however I had to add the load again before asserting for "no transcoders available". Let me know if you recommend any other changes.

core/orch_test.go
```diff
@@ -299,30 +299,44 @@ func TestSelectTranscoder(t *testing.T) {
...
+	assert.Nil(err)
+	assert.Equal(t1, currentTranscoder)
+	assert.Equal(1, t1.load)
 
-	// assert no transcoder returned if all at they capacity
-	noTrans := m.selectTranscoder()
+	// Add some more load and assert no transcoder returned if all at capacity
+	currentTranscoder, err = m.selectTranscoder(testSessionId)
+	assert.Nil(err)
+	assert.Equal(t2, currentTranscoder)
```
kyriediculous, line 334: I don't believe these 3 Lines are required. 

t1 and t2 will already be at capacity (1) here as they both have an active session at this point. 

```suggestion

```

core/orch_test.go
```diff
@@ -299,30 +299,44 @@ func TestSelectTranscoder(t *testing.T) {
...
+	noTrans, err := m.selectTranscoder(testSessionId3)
+	assert.Equal(err, ErrNoTranscodersAvailable)
 	assert.Nil(noTrans)
 
-	m.completeTranscoders(t1)
-	m.completeTranscoders(t1)
+	// assert that load is emtry after ending stream sessions
```
kyriediculous:
```suggestion
	// assert that load is empty after ending stream sessions
```

kyriediculous (changes requested):
Looks good ! Just two very small comments. 

Feel free to rebase and flatten the commit history once you've addressed those. 

Let's use a commit message that indicates which packages were altered and also use that in the title of the PR 

e.g. "core: re-use remote transcoder for stream"

core/orch_test.go
reubenr0d: Oh, I am actually closing the earlier session [here](https://github.com/reubenr0d/go-livepeer/blob/reuse-transcoder-for-stream/core/orch_test.go#L322), so this would add the load again, before asserting for "no transcoders available".

CHANGELOG_PENDING.md
```diff
@@ -10,6 +10,8 @@
 
 #### Orchestrator
 
+- \#1842 Re-use remote transcoders for a stream sessions (@reubenr0d)
```
yondonfu: Mind changing this to reference the PR number instead of the issue number? The current convention in this repo is to use the PR number if available.

core/orchestrator.go
```diff
@@ -856,35 +873,45 @@ func (rtm *RemoteTranscoderManager) selectTranscoder() *RemoteTranscoder {
...
 			rtm.remoteTranscoders = rtm.remoteTranscoders[:last]
 			continue
 		}
 		if currentTranscoder.load == currentTranscoder.capacity {
 			// Head of queue is at capacity, so the rest must be too. Exit early
-			return nil
+			return nil, ErrNoTranscodersAvailable
 		}
+
+		// Assing transcoder to session for future use
```
```suggestion
		// Assigning transcoder to session for future use
```

core/orchestrator.go
```diff
@@ -842,12 +852,19 @@ func (rtm *RemoteTranscoderManager) Manage(stream net.Transcoder_RegisterTransco
 		totalLoad, totalCapacity, liveTranscodersNum = rtm.totalLoadAndCapacity()
 	}
 	rtm.RTmutex.Unlock()
+
+	// Remove all streamSessions assigned to this transcoder
+	for sessionId, assignedTranscoder := range rtm.streamSessions {
```
yondonfu: Instead of looping through `streamSessions` when a transcoder is removed from the live transcoders map, which will get more expensive with the # of streams being concurrently served by the orchestrator, what about checking if the transcoder is still live in `selectTranscoder()` and calling `completeStreamSession()` if it is no longer live? This would defer the completion of the stream session to selection time and the RemoteTranscoderManager could quickly select another transcoder. See my other comment in `selectTranscoder()`.

core/orchestrator.go
```diff
@@ -856,35 +873,45 @@ func (rtm *RemoteTranscoderManager) selectTranscoder() *RemoteTranscoder {
 	}
 
 	for checkTranscoders(rtm) {
+		currentTranscoder, sessionExists := rtm.streamSessions[sessionId]
```
yondonfu, line 897:
```suggestion
		currentTranscoder, sessionExists := rtm.streamSessions[sessionId]
		if !sessionExists {
		    last := len(rtm.remoteTranscoders) - 1
		    currentTranscoder = rtm.remoteTranscoders[last]
		}
```

I think this should allow the RemoteTranscoderManager to quickly switch transcoders if a transcoder for an existing stream session is no longer live.

core/orchestrator.go
```diff
@@ -856,35 +873,45 @@ func (rtm *RemoteTranscoderManager) selectTranscoder() *RemoteTranscoder {
...
+			return currentTranscoder, nil
+		}
+
 		last := len(rtm.remoteTranscoders) - 1
-		currentTranscoder := rtm.remoteTranscoders[last]
+		currentTranscoder = rtm.remoteTranscoders[last]
+
 		if _, ok := rtm.liveTranscoders[currentTranscoder.stream]; !ok {
 			// transcoder does not exist in table; remove and retry
 			rtm.remoteTranscoders = rtm.remoteTranscoders[:last]
```
yondonfu: Doesn't look like we're currently removing a transcoder from this slice if the transcoder was selected for an existing stream session (as opposed to just being the last transcoder in the slice). We'll need to figure out a way to do this given that in this scenario we can't just remove the last transcoder from the slice since the transcoder that was selected may not necessarily be the last in the slice.

cc @kyriediculous to advise here.

yondonfu (changes requested):

core/orchestrator.go
reubenr0d: Yup, this would certainly get more expensive with the increase in the # of streams, although even if we were to handle it in `selectTranscoder()`, we would have to loop over all the sessions, since it's a mapping of streamSession => T to remove all sessions for a particular T. Also, I suppose that would increase the response time of `selectTranscoder()`, which is more critical than the speed at which an EOF is handled. Is that right?

core/orchestrator.go
reubenr0d: Well, it is doing just that in the following lines, after the early return if sessionExists. It would effectively be the same, I suppose.

CHANGELOG_PENDING.md
reubenr0d: Oh, let me change that.

core/orchestrator.go
yondonfu: My thought was that we can avoid the loop all together. Instead of removing all sessions for a T in a single loop when the T disconnects, leave the sessions intact. Then, when `selectTranscoder()` is called, the for loop will make sure that transcoders that are no longer in the live map are skipped and then a session can be removed at that point:

```golang
for checkTranscoders(rtm) {
    ...
    if _, ok := rtm.liveTranscoders[currentTranscoder.stream]; !ok {
        // Remove the stream session because the transcoder is no longer live
        if sessionExists {
            rtm.completeStreamSession(sessionId)
        }
        ...
    }
}
```

core/orchestrator.go
yondonfu: The idea here is always run the live transcoder map check after selecting a transcoder. Previously, if a session already existed, the method was returning the transcoder immediately before running the live transcoder map check.

reubenr0d: Oh right, this and the above comment make sense now, thanks for clarifying, will make the changes.

core/orchestrator.go
kyriediculous: The only possibility I see is looping through the slice if last != currentTranscoder 

We are already resorting this list constantly anyway so more iterative behaviour doesn't seem like it would hurt too much. Not that it's optimal and this entire system of tracking a map of connections alongside an array of transcoders is very weird and should likely be redesigned at some point to be a unified and single data structure tracking remote transcoders.

yondonfu: Sounds reasonable for now.

core/orchestrator.go
kyriediculous: The idea was that we wouldn't have to do this check in `selectTranscoders` but just straight up select a new one. 

But I'm fine with this suggestion.

yondonfu: > The idea was that we wouldn't have to do this check in selectTranscoders but just straight up select a new one.

Not sure I follow? I believe what you're describing is what will happen with the suggestion. If a transcoder for an existing session is selected, but is no longer live it will be discarded and another transcoder will be selected.

kyriediculous: The current implementation would remove the session from `streamSessions` as soon as the connection between O and T closes. 

So we would already select a new transcoder on the first iteration rather than on the second. 

With the suggestion we would first remove the current transcoder that is no longer live in the first iteration, then select a new one in the second iteration. 

But as I mentioned I do see the cleanliness benefit of the suggestion as well so I don't have a strong opinion towards one or the other.

yondonfu: Cool. Let's go with the suggestion then.

core/orchestrator.go
kyriediculous: @reubenr0d perhaps we can create a functional helper for this something like 

```go
func removeFromRemoteTranscoders(rt *RemoteTranscoder, remoteTranscoders []*RemoteTranscoder) []*RemoteTranscoder {
	lastIndex := len(remoteTranscoders) - 1
	last := remoteTranscoders[lastIndex]
	if rt == last {
		return remoteTranscoders[:lastIndex]
	}

	newRemoteTs := make([]*RemoteTranscoder, 0)
	for i, t := range remoteTranscoders {
		if t == rt {
			if i == 0 {
				return remoteTranscoders[1:]
			}
			newRemoteTs = remoteTranscoders[i-1 : i]
			newRemoteTs = append(newRemoteTs, remoteTranscoders[i+1:]...)
			break
		}
	}
	return newRemoteTs
}
```

And then replace L886 with a call to this function

core/orchestrator.go
reubenr0d: So correct me here: after this PR, we would only be incrementing load on new sessions, so if a segment on the same session comes in, it would still increment the load in the lines below. Also it would error out with `ErrNoTranscodersAvailable` if it as max load, even if it gets another segment from an already existing stream session.
Thoughts? How do you'll recommend I proceed? I could  add an `if (!sessionExists)` around the capacity check and subsequent lines until the return, let me do that.

reubenr0d:
I've pushed the changes to the same commit(will use !fixup henceforth), have a look. The Ts are getting removed from the list properly now, but don't seem to adding back, if there is no error. [This](https://github.com/reubenr0d/go-livepeer/blob/reuse-transcoder-for-stream/core/orch_test.go#L360) is the assertion in question. Shouldn't it be 2 after the T succeeds? I'm not quite sure how would it get selected, if the T was not in the list in the first place. Any advice?

kyriediculous:
> I've pushed the changes to the same commit(will use !fixup henceforth), have a look. The Ts are getting removed from the list properly now, but don't seem to adding back, if there is no error. This is the assertion in question. Shouldn't it be 2 after the T succeeds? I'm not quite sure how would it get selected, if the T was not in the list in the first place. Any advice?

@reubenr0d Transcoder 2 gets removed here: https://github.com/reubenr0d/go-livepeer/blob/59b545b57a94f350879307da304a838e99d2c390/core/orch_test.go#L343

reubenr0d:
>Transcoder 2 gets removed here

Right, the comment "assert transcoder gets added back to remoteTranscoders if no transcoding error" threw me off, if it's not supposed to get added back, shall I just remove the last set of assertions after that comment?

core/orchestrator.go
```diff
@@ -856,35 +889,46 @@ func (rtm *RemoteTranscoderManager) selectTranscoder() *RemoteTranscoder {
...
-	return nil
+	return nil, ErrNoTranscodersAvailable
 }
 
-func (rtm *RemoteTranscoderManager) completeTranscoders(trans *RemoteTranscoder) {
-	rtm.RTmutex.Lock()
-	defer rtm.RTmutex.Unlock()
-
-	t, ok := rtm.liveTranscoders[trans.stream]
+func (rtm *RemoteTranscoderManager) completeStreamSession(sessionId string) {
```
kyriediculous, line 931: Let's add a comment here for clarity 

```
// compleStreamSessions end a stream session for a remote transcoder and decrements its laod
// caller should hold the mutex lock
```

core/orchestrator.go
```diff
@@ -847,7 +859,28 @@ func (rtm *RemoteTranscoderManager) Manage(stream net.Transcoder_RegisterTransco
 	}
 }
 
-func (rtm *RemoteTranscoderManager) selectTranscoder() *RemoteTranscoder {
+func removeFromRemoteTranscoders(rt *RemoteTranscoder, remoteTranscoders []*RemoteTranscoder) []*RemoteTranscoder {
```
kyriediculous, line 862: Let's add a unit test for this. 

let's make sure to tackle following cases:
- Array length is 0 
- T to remove is the head
- T to remove is the tail
- T to remove is somewhere in the middle

kyriediculous (changes requested):

core/orchestrator.go
reubenr0d: I did add the tests [here](https://github.com/livepeer/go-livepeer/pull/1849/files#diff-9734c79b7138aa52372c8c9714be7aba54b219fb23099883e1533f44d9f79385R390), although I tested with only two Ts, let me add some more and test the cases you mentioned.

reubenr0d: Added the change [here](https://github.com/livepeer/go-livepeer/pull/1849/commits/e5654d807f84faad8c6e87225d65494f0172608a), let me know if you have any other suggestions.

kyriediculous:
@yondonfu Did you have any other remarks regarding the changes you requested ?

yondonfu:
See https://github.com/livepeer/go-livepeer/pull/1864#issuecomment-832172977

kyriediculous (approved):

yondonfu (approved):
LGTM!

PR #1854, opened 2021-04-28
Allow to pass region in the custom s3 storage url 

darkdarkdragon:
All the previous S3-compatible storages we've tested was ignoring 'region' part of configuration.
So in our code it was hardcoded as value 'ignored'.
But filebase.com does want it to be set to specific value.
This updates allows to pass region as the part of the custom URL, like this:

`s3+http://user:password@example.com:9000/path/to/region-name/bucket-name`

Change parsing of custom s3 url

unit test

**Checklist:**

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -7,6 +7,7 @@
 #### General
 
 - \#1848 Use fee cut instead of fee share for user facing language in the CLI (@kyriediculous)
+- \#1854 Allow to pass region in the custom s3 storage URL (@darkdarkdragon)
 
 #### Broadcaster
 
--- a/drivers/drivers_test.go
+++ b/drivers/drivers_test.go
@@ -26,12 +26,26 @@ func TestS3URL(t *testing.T) {
 
 func TestCustomS3URL(t *testing.T) {
 	assert := assert.New(t)
-	os, err := ParseOSURL("s3+http://user:password@example.com:9000/path/to/bucket-name", true)
+	os, err := ParseOSURL("s3+http://user:password@example.com:9000/bucket-name", true)
 	s3, iss3 := os.(*s3OS)
 	assert.Equal(true, iss3)
 	assert.Equal(nil, err)
 	assert.Equal("http://example.com:9000", s3.host)
 	assert.Equal("bucket-name", s3.bucket)
+	assert.Equal("user", s3.awsAccessKeyID)
+	assert.Equal("password", s3.awsSecretAccessKey)
+	assert.Equal("ignored", s3.region)
+}
+
+func TestCustomS3URLWithRegion(t *testing.T) {
+	assert := assert.New(t)
+	os, err := ParseOSURL("s3+http://user:password@example.com:9000/path/to/region-name/bucket-name", true)
+	s3, iss3 := os.(*s3OS)
+	assert.Equal(true, iss3)
+	assert.Equal(nil, err)
+	assert.Equal("http://example.com:9000", s3.host)
+	assert.Equal("bucket-name", s3.bucket)
+	assert.Equal("region-name", s3.region)
 	assert.Equal("user", s3.awsAccessKeyID)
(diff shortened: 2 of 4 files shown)


yondonfu (commented):
Is the part of the path before the bucket name always the region name for S3 compatible APIs or is this a filebase.com specific convention? And related to that: is it possible for a path to have multiple parts before the bucket name such that none of the parts should be treated as a region name i.e. `not-a-region/also-not-a-region/bucket-name`?

Also, please make sure to update the [pending changelog](https://github.com/livepeer/go-livepeer/pull/CHANGELOG_PENDING.md) according to the [contribution guide](https://github.com/livepeer/go-livepeer/pull/doc/contributing.md).

iameli (commented):
Hmmmm yeah not sure I understand what this does. If I'm running a Minio server at `https://iame.li/storage/minio`, will `s3+https://user:password@iame.li/storage/minio` still work? What's the difference between a region and a path in this context?

drivers/drivers_test.go
```diff
@@ -26,14 +26,28 @@ func TestS3URL(t *testing.T) {
...
 	assert.Equal("http://example.com:9000", s3.host)
 	assert.Equal("bucket-name", s3.bucket)
 	assert.Equal("user", s3.awsAccessKeyID)
 	assert.Equal("password", s3.awsSecretAccessKey)
+	assert.Equal("ignored", s3.region)
+}
+
+func TestCustomS3URLWithRegion(t *testing.T) {
+	assert := assert.New(t)
+	os, err := ParseOSURL("s3+http://user:password@example.com:9000/path/to/region-name/bucket-name", true)
```
iameli, line 42: What happened to `/path/to`? There's no `assert.Equal` that has it.

darkdarkdragon: It is ignored. It was always ignored (`hosturl.Path = ""`)

darkdarkdragon:
> Hmmmm yeah not sure I understand what this does. If I'm running a Minio server at `https://iame.li/storage/minio`, will `s3+https://user:password@iame.li/storage/minio` still work? What's the difference between a region and a path in this context?

Was it working before? We're just throwing out `/storage` part and not using it anywhere.

iameli:
Hmm, it was my intent to have that work when I wrote the spec and implementation. If that's not happening, I think that's a bug. Sometimes stuff is in subdirectories.

darkdarkdragon:
> Hmmmm yeah not sure I understand what this does. If I'm running a Minio server at `https://iame.li/storage/minio`, will `s3+https://user:password@iame.li/storage/minio` still work? What's the difference between a region and a path in this context?

Actually I've looked into minio's docs - it can't be ran on subpath (https://github.com/minio/minio/issues/5947#issuecomment-415342037)

iameli:
Oh, hah. Okay. I'm cool with this, then.

darkdarkdragon:
> Is the part of the path before the bucket name always the region name for S3 compatible APIs or is this a filebase.com specific convention? And related to that: is it possible for a path to have multiple parts before the bucket name such that none of the parts should be treated as a region name i.e. `not-a-region/also-not-a-region/bucket-name`?
> 


There is no convention for such URLs. 
Minio doesn't allow to run it under non-root path, and I'm not aware of any S3-compatible service that does it.
Also, right now we're just ignoring anything in the path name except last part which we treat as bucket name.

darkdarkdragon:
@yondonfu added changelog

CHANGELOG_PENDING.md
```diff
@@ -7,6 +7,7 @@
 #### General
 
 - \#1848 Use fee cut instead of fee share for user facing language in the CLI (@kyriediculous)
+- \#1854 Allow to pass region in the custom s3 storage URL (@kdarkdarkdragon)
```
yondonfu: Looks like there is an extra k in the username

yondonfu (approved):
Just one comment on a typo in the changelog, but LGTM otherwise so feel free to fix and then merge.

PR #1856, opened 2021-04-28
eth: gas price monitor stability improvements

### LINKED ISSUE #1851: Many insufficient faceValue ticket face value for redeem tx cost errors
**Describe the bug**

There has been an uptick in "insufficient faceValue ticket face value..." errors reported by Os.

The current hypothesis is that the Os are receiving gas prices from their JSON-RPC provider that are really low or even 0 in certain cases and as a result, the computed face value end up being [less than the ticket EV](https://github.com/livepeer/go-livepeer/blob/7a0fd1f6b958d61177fc01c488972f54b8bad1e3/pm/recipient.go#L237) resulting in the face value to be set to the ticket EV. Since the ticket EV is 1000 gwei, in pretty much all cases the face value will be less than the tx cost for redeeming a winning ticket.

See this [tweet thread](https://twitter.com/YondonFu/status/1386693693088964609) for some resources to support the hypothesis that Os are receiving really low gas prices from their JSON-RPC provider at least for brief periods of time.

The negative consequences of this issue are:

- Os receive many winning tickets, but none of them are redeemed
- Os end up with a ton of logs with the aforementioned error
- Any tickets that should be redeemed are stuck behind these low face value tickets in the queue

kyriediculous:
Uses the mean of N last seen gas prices to calculate a gas price suggestion. 

If the gas price oracle suggest a gas price that is not within 3 standard deviations of the current N prices in the list, the newly seen price will be discarded.

- [x] unit tests
- [ ] manual tests

Fixes #1851 

**Checklist:**

### DIFF
--- a/eth/gaspricemonitor.go
+++ b/eth/gaspricemonitor.go
@@ -11,6 +11,8 @@ import (
 	"github.com/livepeer/go-livepeer/monitor"
 )
 
+const gasPricesLength = 3
+
 // GasPriceOracle defines methods for fetching a suggested gas price
 // for submitting transactions
 type GasPriceOracle interface {
@@ -30,9 +32,10 @@ type GasPriceMonitor struct {
 
 	// gasPriceMu protects access to gasPrice
 	gasPriceMu sync.RWMutex
-	// gasPrice is the current gas price to be returned to users
+	// gasPrice is the current gas price to be returned to users, mean of the values in gasPrices
 	gasPrice *big.Int
-
+	// gasPrices is a queue of gas prices
+	gasPrices *gasPrices
 	// update is a channel used to send notifications to a listener
 	// when the gas price is updated
 	update chan struct{}
@@ -44,6 +47,7 @@ func NewGasPriceMonitor(gpo GasPriceOracle, pollingInterval time.Duration) *GasP
 		gpo:             gpo,
 		pollingInterval: pollingInterval,
 		gasPrice:        big.NewInt(0),
+		gasPrices:       newGasPrices(gasPricesLength),
 	}
 }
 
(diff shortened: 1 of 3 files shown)


eth/gaspricemonitor.go
```diff
@@ -139,3 +148,109 @@ func (gpm *GasPriceMonitor) updateGasPrice(gasPrice *big.Int) {
...
+		variance   = big.NewInt(0)
+		length     = big.NewInt(int64(len(q.list)))
+		deviations = big.NewInt(3) // The max acceptable std, anything outside will be removed from the set
+	)
+
+	mean := q.mean()
+	// Calculate variance (sum(x - mean)^2 )/ length
+	for _, price := range q.list {
+		// Calculate the summation
+		x := new(big.Int).Sub(price, mean)
```
yondonfu, line 239: If all the prices in the list are the same, then x = 0 and the variance = 0. As a result, the upper/lower bound would both be the mean so all prices would be discarded unless they are exactly equal to the mean.

yondonfu:
A concern that came to mind while reviewing the stddev check is that given the way it is implemented right now, the gas price monitor can get stuck if it fills the gas price list with low values and then then if eth_gasPrice starts returning higher values that are not skewed by super low gas price MEV txs the gas price monitor will treat all those values as outliers and discard them.

Based on a gas price graph from the last 24 hours I think this is not just theoretical and would have actually happened if this change were deployed during this time period:

Screen Shot 2021-04-29 at 10 47 24 AM

In one of the valleys of the graph the gas price is consistently 6 gwei and then it eventually shoots back up to 45 gwei. Given the way the stddev check is implemented, I believe the gas price monitor would get stuck at 6 gwei (see my other review comment).

Given that the main problem with #1851 seems to be gas prices that are waaaaay to low i.e. 0 or any value that causes the calculated face value to be less than the default ticket EV, I'm leaning toward getting a less "elegant", but simpler solution out more quickly. Since the problem is with the unreasonably low gas prices, we could just discard gas prices that are say less than 1 gwei for now.

What do you think?

eth/gaspricemonitor.go
kyriediculous: We don't insert a price into the list if it is the same as the back of the queue to avoid this exact scenario https://github.com/livepeer/go-livepeer/blob/3a9637e0c5513bc64a94d6779acb3116a594490c/eth/gaspricemonitor.go#L131

yondonfu:
Superseded by #1860. Leaving this open for now in case we revisit a stddev based check for gas price filtering.

PR #1860, opened 2021-04-29
eth: discard gas prices below minGasPrice

### LINKED ISSUE #1851: Many insufficient faceValue ticket face value for redeem tx cost errors

kyriediculous:
Adds a minGasPrice to the gas price monitor and discards fetched gas prices below it 

Sets a specific setting for mainnet 

Unit tests

Fixes #1851 

**Checklist:**

### DIFF
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -22,6 +22,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/ethereum/go-ethereum/params"
 	"github.com/ethereum/go-ethereum/rpc"
 	"github.com/livepeer/go-livepeer/build"
 	"github.com/livepeer/go-livepeer/pm"
@@ -192,6 +193,7 @@ func main() {
 
 	type NetworkConfig struct {
 		ethController string
+		minGasPrice   *big.Int
 	}
 
 	ctx := context.Background()
@@ -202,6 +204,7 @@ func main() {
 		},
 		"mainnet": {
 			ethController: "0xf96d54e490317c557a967abfa5d6e33006be69b3",
+			minGasPrice:   big.NewInt(int64(params.GWei)),
 		},
 	}
 
@@ -224,9 +227,11 @@ func main() {
 	}
 
 	// Setting config options based on specified network
+	var minGasPrice = big.NewInt(0)
 	if netw, ok := configOptions[*network]; ok {
 		if *ethController == "" {
 			*ethController = netw.ethController
+			minGasPrice = netw.minGasPrice
 		}
 		glog.Infof("***Livepeer is running on the %v network: %v***", *network, *ethController)
 	} else {
@@ -542,7 +547,7 @@ func main() {
 
 			sigVerifier := &pm.DefaultSigVerifier{}
 			validator := pm.NewValidator(sigVerifier, timeWatcher)
-			gpm := eth.NewGasPriceMonitor(backend, blockPollingTime)
+			gpm := eth.NewGasPriceMonitor(backend, blockPollingTime, minGasPrice)
 			// Start gas price monitor
 			_, err := gpm.Start(ctx)
 			if err != nil {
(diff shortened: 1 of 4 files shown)


cmd/livepeer/livepeer.go
```diff
@@ -202,6 +203,7 @@ func main() {
 		},
 		"mainnet": {
 			ethController: "0xf96d54e490317c557a967abfa5d6e33006be69b3",
+			minGasPrice:   big.NewInt(1e9),
```
yondonfu: Should we use [params.GWei](https://github.com/ethereum/go-ethereum/blob/d8ff53dfb8a516f47db37dbc7fd7ad18a1e8a125/params/denomination.go#L26)?

eth/gaspricemonitor.go
```diff
@@ -32,18 +32,25 @@ type GasPriceMonitor struct {
...
 
 	// update is a channel used to send notifications to a listener
 	// when the gas price is updated
 	update chan struct{}
 }
 
 // NewGasPriceMonitor returns a GasPriceMonitor
-func NewGasPriceMonitor(gpo GasPriceOracle, pollingInterval time.Duration) *GasPriceMonitor {
+func NewGasPriceMonitor(gpo GasPriceOracle, pollingInterval time.Duration, minGasPrice *big.Int) *GasPriceMonitor {
+	minGasP := big.NewInt(0)
```
yondonfu, line 45: Don't think we need this as the minGasPrice param value is not from an untrusted source (unlike in other cases where we should be more defensive since an untrusted source could give us unexpected values) and the caller is responsible for passing in a sane value.

eth/gaspricemonitor_test.go
```diff
@@ -131,14 +132,21 @@ func TestStart_Polling(t *testing.T) {
 	// There should be more queries now
 	assert.Greater(gpo.Queries(), queries)
 	assert.Equal(gasPrice3, gpm.GasPrice())
+
+	go func() {
+		gpo.SetGasPrice(gasPrice4)
+	}()
+	time.Sleep(100 * time.Millisecond)
+	assert.Greater(gpo.Queries(), queries)
```
yondonfu, line 142: I think we want to update `queries` with `queries = gpo.Queries()` before L136 in order to take a snapshot of the # of queries before and after the sleep to ensure that an additional query was ran during the sleep and after the goroutine executed.

yondonfu (changes requested):
Overall LGTM. Just a few minor comments. After those are addressed and a changelog entry is added should be good to go.

eth/gaspricemonitor.go
kyriediculous: I disagree, I don't think it's correct to assume that a caller will pass in sane values, ever.  It's not because we pass in a sane value that a third party potentially using this package will. Passing in a `nil` value will lead to unexpected behaviour under the form of a panic so we do need this.

cmd/livepeer/livepeer.go
kyriediculous: It seems like the compiler looks at `int64(params.GWei)` as a constant still rather than a variable that needs to convert the untyped constant to an int64 at runtime so I'm okay with this.

eth/gaspricemonitor.go
yondonfu: Upon further thought, I agree with your point here 👍 .

yondonfu:
Good to go after a changelog entry.

kyriediculous:
added changelog entry and rebased

yondonfu (approved):
LGTM!

PR #1863, opened 2021-04-30
.github/workflows: Run GH actions on PR

yondonfu:

Running GH actions on `pull_request` should allow GH actions (includes tests) to run even if the PR is from a fork which is useful for contributors that don't have push access to the repo.

See commit history.

**Checklist:**

- [x] ~[Pending changelog](./CHANGELOG_PENDING.md) updated~ -> no changes related to the node

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -1,5 +1,9 @@
 name: Linux Build
-on: push
+on:
+  pull_request:
+  push:
+    branches:
+      - master
 jobs:
   build:
     runs-on: ubuntu-16.04
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -1,5 +1,9 @@
 name: Windows Build
-on: push
+on:
+  pull_request:
+  push:
+    branches:
+      - master
 jobs:
   build:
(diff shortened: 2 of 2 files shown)


kyriediculous:
IIUC this will only run CI on normal pushes if the branch is `master` unless it's a feature branch with an open PR. Is that what we intend?

yondonfu:
> IIUC this will only run CI on normal pushes if the branch is `master` unless it's a feature branch with an open PR. Is that what we intend?

Yep. The motivation for this configuration is because this seemed like the easiest way to prevent multiple GH action triggers with a single push on a branch. For example, if the `push` event was not restricted to master, then each commit on this branch would trigger an action twice. Now, instead, the `push` event is only relevant for `master` and the `pull_request` event is used to trigger actions for any other branch. The disadvantage to this is that we will no longer be able to run CI just by pushing to a branch without also creating a PR for that branch. But, given that its pretty easy to create a draft PR for a WIP change, I think this disadvantage is acceptable. If anyone knows of an easy alternative to trigger actions for either the `push` or `pull_request` event, but not both without restricting the `push` event to master I'm all ears!

kyriediculous (approved):

PR #1864, opened 2021-05-01
server: fix OT error handling

### LINKED ISSUE #1762: Remote Transcoder hangs after EOF
**Describe the bug**
When a _remote_ transcoder node runs into an `io.EOF` error it will briefly close its connection to the server after it finishes all running transcodes and reconnect to the server. Usually in a gRPC streaming context this error indicates an issue with sending or receiving messages on the stream. 

Sometimes however the transcoder will be stuck in a normally transient state, waiting for running transcode sessions to complete. This means the node will _not_ automatically reconnect to the orchestrator and has to be manually restarted. 

https://github.com/livepeer/go-livepeer/blob/f47f99ccd799510238966799742dc2c02c2e42f6/server/ot_rpc.go#L120

While originally the idea was that this issue could be related due to stuck transcode sessions, it still persists using the go 0.5.14 release. However if there are other root causes of stuck sessions then this problem might still persist. Alternatively the connection to the server is never closed after an `io.EOF` error , this should perhaps be done as well. 

**To Reproduce**
1. Run an orchestrator on mainnet
2. Run a transcoder in split mode 
3. Wait for EOF errors and observe the node being stuck, the transcoder itself will be responding with `no transcoders available` when this occurs.

### LINKED ISSUE #1843: Improve RemoteTranscoder error handling
## Abstract

In a split orchestrator and transcoder architecture, whenever a `RemoteTranscoder` errors out (time-out, bad return data, transcoder error, ...) its connection to the Orchestrator will be reset (closed and re-opened) entirely. 

When this happens all segments in flight for that `RemoteTranscoder`, regardless of their `sessionID`, will be short-circuited as well and the `RemoteTranscoder` loses all concurrent work it was doing, even though only a single segment might have been at fault. 

The current error handling also results in a general `EOF` error when the gRPC stream is closed from the Orchestrator side, allowing for little useful error propagation back to the `RemoteTranscoder`

## Motivation

A `RemoteTranscoder` should not lose all the work it is doing in case it errors out on a single stream. 

In the case where there is only a single orchestrator and remote transcoder, the orchestrator will also lose all current work as it will throw a `NoTranscodersAvailable` error.

When operating in private or public pool mode, it would be useful for the transcoder to also receive informative error messages back about what went wrong with a particular segment/stream.

reubenr0d:

- Moved check for transcode errors on O before decodingPixels
- Added nil check for `chanData.TranscodeData`

- Tested with errored out video segments, one without a keyframe, and another without Video
- Observed that proper error messages now show on the O
- Observed that EOF is not triggered in case of these errors

Fixes #1843 
Fixes #1762

**Checklist:**

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -26,14 +26,13 @@
 
 #### General
 
-- \#1864 Fix OT error handling (@reubenr0d)
-
 #### Broadcaster
 
 #### Orchestrator
 
 - \#1860 Discard low gas prices to prevent insufficient ticket faceValue errors (@kyriediculous)
 - \#1859 Handle error for invalid inferred orchestrator public IP on node startup (@reubenr0d)
+- \#1864 Fix OT error handling (@reubenr0d)
 
 #### Transcoder
 
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -26,6 +26,8 @@
 
 #### General
 
+- \#1864 Fix OT error handling (@reubenr0d)
+
 #### Broadcaster
 
 #### Orchestrator
--- a/core/orchestrator.go
+++ b/core/orchestrator.go
@@ -550,7 +550,7 @@ func (n *LivepeerNode) transcodeSeg(config transcodeConfig, seg *stream.HLSSegme
 	start := time.Now()
 	tData, err := transcoder.Transcode(md)
 	if err != nil {
-		glog.Errorf("Error transcoding manifestID=%s sessionID=%s segNo=%d segName=%s - %v", string(md.ManifestID), md.AuthToken.SessionId, seg.SeqNo, seg.Name, err)
+		glog.Errorf("Error transcoding manifestID=%s sessionID=%s segNo=%d segName=%s err=%v", string(md.ManifestID), md.AuthToken.SessionId, seg.SeqNo, seg.Name, err)
 		return terr(err)
 	}
 
@@ -741,8 +741,12 @@ func (rt *RemoteTranscoder) Transcode(md *SegTranscodingMetadata) (*TranscodeDat
 	case <-ctx.Done():
 		return signalEOF(ErrRemoteTranscoderTimeout)
 	case chanData := <-taskChan:
+		segmentLen := 0
+		if chanData.TranscodeData != nil {
+			segmentLen = len(chanData.TranscodeData.Segments)
+		}
 		glog.Infof("Successfully received results from remote transcoder=%s segments=%d taskId=%d fname=%s dur=%v err=%v",
-			rt.addr, len(chanData.TranscodeData.Segments), taskID, fname, time.Since(start), chanData.Err)
+			rt.addr, segmentLen, taskID, fname, time.Since(start), chanData.Err)
 		return chanData.TranscodeData, chanData.Err
 	}
 }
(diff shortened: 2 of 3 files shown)


kyriediculous:
I like the solution, it clearly returns the error that should be returned. 

Let's also make sure the header is always a sane value. Currently when we encounter a transcoding error from the T https://github.com/livepeer/go-livepeer/blob/48409dfcfe988a02f5166d64db2834b4dea4b6f0/server/ot_rpc.go#L149

We will not set the "Pixels" header https://github.com/livepeer/go-livepeer/blob/48409dfcfe988a02f5166d64db2834b4dea4b6f0/server/ot_rpc.go#L190 

We should always be setting the "Pixels" header, even if `tData` is nil and an error is encountered. 

If this header is not set we will run into an error converting the string to an integer. 

At L190 in ot_rpc.go let's make this change

```go 
pixels := int64(0)
    if tData != nil {
        pixels = tData.Pixels
    }
    req.Header.Set("Pixels", strconv.FormatInt(pixels, 10))
 ```
 
 ____________ 
 
 I'm also wondering if we can add unit tests for this, we can create erroneous responses like this:
 
 ```
 		body.Write([]byte(err.Error()))
		contentType = transcodingErrorMimeType
```

See https://github.com/livepeer/go-livepeer/blob/48409dfcfe988a02f5166d64db2834b4dea4b6f0/server/ot_rpc.go#L156

kyriediculous:
Nevermind on the unit tests, this unit tests should already take care of those cases: https://github.com/livepeer/go-livepeer/blob/48212dd5dc3d0f1a977f710cea81dd096f8cd944/server/ot_rpc_test.go#L178

kyriediculous (commented):
LGTM 👍 

@yondonfu any remarks ? We have tested this both seperately and came to the same conclusions. 

No source: 

```
E0504 11:50:45.678492   78423 orchestrator.go:553] Error transcoding manifestID=test sessionID=53e9ccc9 segNo=1 segName=https://127.0.0.1:8935/stream/53e9ccc9/1.tempfile - TranscoderInvalidVideo
```

No keyframe

```
E0504 11:48:54.873341   78423 orchestrator.go:553] Error transcoding manifestID=test sessionID=0e05dfc6 segNo=1 segName=https://127.0.0.1:8935/stream/0e05dfc6/1.tempfile - No keyframes in input
```

kyriediculous:
I know we didn't touch this in this PR but 

https://github.com/livepeer/go-livepeer/blob/master/core/orchestrator.go#L553

This log statement still uses a dash to seperate the error from the rest of the message. We currently use label=.. 

do you think we can change the end of this statement to `err=%v` to match this idiom ?

reubenr0d:
> do you think we can change the end of this statement to `err=%v` to match this idiom ?

Yea, sure. [Done](https://github.com/livepeer/go-livepeer/pull/1864/commits/085ff8e47d729355093c6a8f34f37c34a28b1664).

kyriediculous:
Looks good to me, let's rebase (squash the fixup) 

https://thoughtbot.com/blog/autosquashing-git-commits

yondonfu:
Looks good to me, but I'm still looking into enabling CI runs for PRs based off of forks (thought I got it working, but looks like some additional tweaks are needed). Will approve & merge once we get CI running & passing.

reubenr0d:
> Looks good to me, but I'm still looking into enabling CI runs for PRs based off of forks (thought I got it working, but looks like some additional tweaks are needed). Will approve & merge once we get CI running & passing.

Ahh, seems to me like the dockerhub creds aren't getting picked up from the Github Secrets over [here](https://github.com/livepeer/go-livepeer/runs/2502426922?check_suite_focus=true). Maybe it would be a good idea to not do the `docker push` for MRs, but only on master merge and only do the build/tests locally for PRs? Let me see what I can do tomorrow.

kyriediculous (approved):

PR #1871, opened 2021-05-04
Fix CI runs for forks

yondonfu:

This PR is motivated by the desire to run build & test CI jobs for PRs opened from forks.

Prior to this PR:

- The Linux (Docker) build would fail because it requires credentials that are not accessible to forks
- The Windows build would fail because it requires credentials that are not accessible to forks
- The Darwin build will still fail due to lack of credentials for forks, but that can be addressed later once it is ported to GH actions
- Tests would not run because they only run after a successful Docker build

After this PR:

- The Linux (no Docker) build will run and skip any steps that require credentials for forks
- The Windows build will run and skip any steps that require credentials for forks
- The Docker build will run and skip any steps that require credentials for forks
- Tests (after the Linux build) will always run for forks

Additionally, this PR fixes an issue with the branch name used in all GH action builds that was introduced in https://github.com/livepeer/go-livepeer/pull/1863 which updated the GH actions to be triggered by PRs. Prior to this PR, the branch name used was of the form `refs/pull/1871/merge` instead of `yf/gh-actions-fork`.

- Separate the Docker workflow from the Linux workflow
- Cache go modules and ffmpeg dependencies in the Linux workflow to speed up run times
- Conditionally run certain steps of workflows depending on whether the action was triggered by a fork
- Conditionally set the ref used to generate branch names in workflows based on whether the action was triggered by a pull request or a push
- Use the `GHA_REF` env var to store the ref used to generate branch names instead of `GITHUB_REF` because the latter cannot be overridden

Used a [PR from a fork](https://github.com/livepeer/go-livepeer/pull/1878).

**Checklist:**

- [ ] ~[Pending changelog](./CHANGELOG_PENDING.md) updated~ No changes to the node.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -6,6 +6,7 @@ on:
       - master
 jobs:
   build:
+    if: github.event.pull_request.head.repo.full_name == github.repository
     runs-on: ubuntu-16.04
     steps:
     - name: Checkout
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,0 +1,52 @@
+name: Docker Build
+on:
+  pull_request:
+  push:
+    branches:
+      - master
+jobs:
+  build:
+    runs-on: ubuntu-16.04
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2
+      with:
+        # Needed for commands that depend on git tags
+        fetch-depth: 0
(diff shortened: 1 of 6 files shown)


kyriediculous:
Looks okay to me from a first glance, one additional set of eyes might be good though.

kyriediculous (approved):

PR #1873, opened 2021-05-05
pm: increase TicketParams expiration to 10 blocks

### LINKED ISSUE #1872: pm: Increase TicketParams expiration
Currently the `TicketParams` expiration is set to 5 blocks which forces relatively strict sync between Orchestrators and Broadcasters in terms of their view of the canonical Ethereum chain. 

https://github.com/livepeer/go-livepeer/blob/1e5154a5a04d8361b7033110899b298a360775b3/pm/recipient.go#L24

It's also possible for the `TicketParams` to expire when a segment is in flight, while this is an edge case because [input segments are capped in size](https://github.com/livepeer/go-livepeer/pull/1827) it is still possible to occur, especially since Ethereum blocktime is not fixed. 

The propasal is to double this duration to 10 blocks , which would be 2 minutes on average. This seems to be a reasonable balance between sync requirements between nodes and staleness of values such as ticket faceValue, which depends on the gas price. 

**Additional context**
We also want to refresh ticket params before they expire at e.g. 10% before that in https://github.com/livepeer/go-livepeer/issues/1852. At the current expiration this would involve frequent refreshes, at 10 blocks this seems more reasonable.

kyriediculous:
Increase the ticketparams expiration to 10 blocks 

Fixes #1872 

**Checklist:**

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -16,6 +16,7 @@
 #### Orchestrator
 
 - \#1845 Staking actions with hints (@kyriediculous)
+- \#1873 Increase TicketParams expiration to 10 blocks (@kyriediculous)
 
 #### Transcoder
 
--- a/pm/recipient.go
+++ b/pm/recipient.go
@@ -21,7 +21,7 @@ var errInsufficientSenderReserve = errors.New("insufficient sender reserve")
 // maxWinProb = 2^256 - 1
 var maxWinProb = new(big.Int).Sub(new(big.Int).Lsh(big.NewInt(1), 256), big.NewInt(1))
 
-var paramsExpirationBlock = big.NewInt(5)
+var paramsExpirationBlock = big.NewInt(10)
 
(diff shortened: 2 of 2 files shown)


yondonfu (changes requested):
Looks good. Just missing a changelog entry.

CHANGELOG_PENDING.md
```diff
@@ -8,6 +8,7 @@
 
 - \#1848 Use fee cut instead of fee share for user facing language in the CLI (@kyriediculous)
 - \#1854 Allow to pass region in the custom s3 storage URL (@darkdarkdragon)
+- \#1873 Increase TicketParams expiration to 10 blocks (@kyriediculous)
```
yondonfu: I think this should be under "Orchestrator" since it is an orchestrator implementation detail.

yondonfu (approved):
Minor comment about categorizing the changelog entry, but feel free to address prior to merge.

LGTM

PR #1874, opened 2021-05-05
server,discovery: fetch as many Os as possible during discovery before a shorter timeout

### LINKED ISSUE #1853: Fetch as many Os as possible during discovery before a shorter timeout

The discovery mechanism currently to create a list of Os to select from of size min(poolSize, maxInflight * 2) where poolSize is the total # of Os that B is aware of and maxInflight = 4 which is based on an 8 second HTTP timeout and an assumed segment length of 2 seconds [1]. 

A few implications of the current behavior of discovery:

- B will cancel all in-flight discovery requests once it hits its target list size. If the target list size is hit very quickly, the requests are cancelled very quickly and in some cases can abruptly drop the connection in the middle of a TLS handshake which will a) cause an error to be logged until (although #1646 will help with this) and b) if this happens a lot services can flag this as suspicious DoS like behavior. It would be nice to avoid such quick cancellations if possible
- B refreshes its list once it is at half of its target list size. So, if the target list size is 8, B refreshes the list i.e. sends discovery requests to all known Os after 4 Os have been removed from the list. It would be nice to refresh the list less often
- B prioritizes selecting from Os that are very quick to respond. B does not get an opportunity to select from Os that might be a bit slower to respond, but also able to return transcoded results quickly. It would be nice to give an opportunity to Os that are within an acceptable range of B to be selected

[1] The reader will notice that the maxInflight formula is a bit outdated due to the popularity of non-uniform segment lengths on the network today.

- Use the OrchestratorPool size as the target # of Os for discovery as is done in [this commit](https://github.com/livepeer/go-livepeer/commit/4ecc9c6ee32e078040e39bba6f6b88d1d234dac0)
- Reduce the [discovery timeout](https://github.com/livepeer/go-livepeer/blob/7a0fd1f6b958d61177fc01c488972f54b8bad1e3/discovery/discovery.go#L19) to 1 second
- There may be some other considerations around the `numOrchs` field stored in `BroadcastSessionsManager` which influences how long Os are [temporarily suspended for](https://github.com/livepeer/go-livepeer/blob/7a0fd1f6b958d61177fc01c488972f54b8bad1e3/server/broadcast.go#L269) and also when to [refresh the session list](https://github.com/livepeer/go-livepeer/blob/7a0fd1f6b958d61177fc01c488972f54b8bad1e3/server/broadcast.go#L87)

This [commit](https://github.com/livepeer/go-livepeer/commit/4ecc9c6ee32e078040e39bba6f6b88d1d234dac0) is a good starting point.

Note that a larger list size and less frequent refreshes will also mean that it will be more common for B to refresh an individual sessions (i.e. due to an expired auth token) by calling GetOrchestrator on an O right before it sends a segment to it. Might be ok.

Note that these changes will mean it will either be the case that the 1 second timeout has elapsed or *all* available Os have responded during discovery. The former is more likely. Thus, in most cases we should expect discovery to take 1 second before B's session list is updated. During a stream this is fine since the refresh happens in the background. But, at the start of a stream, the first segment can't be submitted until the refresh completes which means there will always be a 1 second delay at the start of a stream before a segment is submitted. We should observe the behavior of this in practice and if needed we can consider reducing the discovery timeout further as well.

**Additional context**

We should experiment with this and observe the following data points:

- Distribution of work before and after these changes
- Frequency of session list refreshes
- Impact of a fixed 1 second delay of discovery on the first segment submission of a stream

kyriediculous:
This PR changes discovery to take into account all possible orchestrators as long as they respond to a discovery request within 1 second. Previously we would stop gathering responses once we received a small subset of responses. 

- change `numOrchs` to be equal to the Orchestrator pool size 
- Decreased the timeout to 1 second
- Changed the suspension time to poolSize/numSessions (not sure if this is better but at least this avoids poolSize/numOrchs which would always be 1)
- Changed the cut-off point at which we do a refresh to be at less than 20% of the original selected set.

https://github.com/livepeer/internal-project-tracking/issues/124

Fixes #1853 

**Checklist:**

### DIFF
--- a/discovery/discovery.go
+++ b/discovery/discovery.go
@@ -17,7 +17,9 @@ import (
 	"github.com/golang/glog"
 )
 
-var getOrchestratorsTimeoutLoop = 3 * time.Second
+const MinWorkingSetSize = 8
+
+var getOrchestratorCutoffTime = 1 * time.Second
 
 var serverGetOrchInfo = server.GetOrchestratorInfo
 
@@ -49,7 +51,6 @@ func (o *orchestratorPool) GetURLs() []*url.URL {
 func (o *orchestratorPool) GetOrchestrators(numOrchestrators int, suspender common.Suspender, caps common.CapabilityComparator) ([]*net.OrchestratorInfo, error) {
 	numAvailableOrchs := len(o.uris)
 	numOrchestrators = int(math.Min(float64(numAvailableOrchs), float64(numOrchestrators)))
-	ctx, cancel := context.WithTimeout(context.Background(), getOrchestratorsTimeoutLoop)
 
 	infoCh := make(chan *net.OrchestratorInfo, numAvailableOrchs)
 	errCh := make(chan error, numAvailableOrchs)
@@ -80,6 +81,9 @@ func (o *orchestratorPool) GetOrchestrators(numOrchestrators int, suspender comm
 		}
 		return caps.CompatibleWith(info.Capabilities)
 	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), getOrchestratorCutoffTime)
+
 	getOrchInfo := func(uri *url.URL) {
 		info, err := serverGetOrchInfo(ctx, o.bcast, uri)
 		if err == nil && isCompatible(info) {
@@ -126,7 +130,7 @@ func (o *orchestratorPool) GetOrchestrators(numOrchestrators int, suspender comm
 	}
 	cancel()
 
-	if len(infos) < numOrchestrators {
+	if len(infos) <= server.MinWorkingSetSize(numOrchestrators) {
 		diff := numOrchestrators - len(infos)
 		for i := 0; i < diff && suspendedInfos.Len() > 0; i++ {
 			info := heap.Pop(suspendedInfos).(*suspension).orch
--- a/discovery/discovery.go
+++ b/discovery/discovery.go
@@ -19,7 +19,7 @@ import (
 
 const MinWorkingSetSize = 8
 
-var getOrchestratorCutoffTime = 1 * time.Second
+var getOrchestratorCutoffTime = 250 * time.Millisecond
 
 var serverGetOrchInfo = server.GetOrchestratorInfo
 
--- a/server/broadcast.go
+++ b/server/broadcast.go
@@ -34,6 +34,8 @@ import (
 	"github.com/livepeer/lpms/stream"
 )
 
+var minWorkingSetRatio = 0.1
+
 var refreshTimeout = 2500 * time.Millisecond
 var maxDurationSec = common.MaxDuration.Seconds()
 
@@ -46,6 +48,8 @@ var MetadataPublishTimeout = 1 * time.Second
 
 var getOrchestratorInfoRPC = GetOrchestratorInfo
 var downloadSeg = drivers.GetSegmentData
+
+const MIN_SIZE = 1
 
 type BroadcastConfig struct {
 	maxPrice *big.Rat
@@ -97,7 +101,7 @@ func (bsm *BroadcastSessionsManager) selectSession() *BroadcastSession {
 
 	checkSessions := func(m *BroadcastSessionsManager) bool {
 		numSess := m.sel.Size()
-		if numSess < int(math.Ceil(float64(m.numOrchs)/2.0)) {
+		if numSess < int(math.Ceil(float64(m.numOrchs)/5.0)) {
 			go m.refreshSessions()
 		}
 		return (numSess > 0 || bsm.lastSess != nil)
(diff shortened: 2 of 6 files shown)


server/broadcast.go
```diff
@@ -448,6 +454,7 @@ func processSegment(cxn *rtmpConnection, seg *stream.HLSSegment) ([]string, erro
 		}
 
 		// recoverable error, retry
+		glog.V(common.DEBUG).Infof("Retrying current segment mid=%v nonce=%d seqNo=%d", mid, nonce, seg.SeqNo)
```
yondonfu: I would prefer to remove this log since there is already a similar log that is being updated with the manifestID in https://github.com/livepeer/go-livepeer/pull/1875.

server/broadcast.go
```diff
@@ -266,16 +268,20 @@ func (bsm *BroadcastSessionsManager) cleanup() {
 }
 
 func (bsm *BroadcastSessionsManager) suspendOrch(sess *BroadcastSession) {
-	bsm.sus.suspend(sess.OrchestratorInfo.GetTranscoder(), bsm.poolSize/bsm.numOrchs)
+	poolSize := math.Max(MIN_SIZE, float64(bsm.poolSize))
+	selSize := math.Max(MIN_SIZE, float64(bsm.sel.Size()))
+	penalty := int(math.Ceil(poolSize / selSize))
+	fmt.Println(poolSize, selSize, penalty)
```
yondonfu: Can be removed.

yondonfu (changes requested):
A few minor comments, but largely looks good and just missing a changelog entry.

As the next step, let's analyze the following things (the ability to fetch [work distribution data](https://github.com/livepeer/internal-project-tracking/issues/124) should be useful) mentioned in https://github.com/livepeer/go-livepeer/issues/1853 in tests on the network:

> We should experiment with this and observe the following data points:

> - Distribution of work before and after these changes
> - Frequency of session list refreshes
> - Impact of a fixed 1 second delay of discovery on the first segment submission of a stream

discovery/discovery.go
```diff
@@ -80,6 +81,10 @@ func (o *orchestratorPool) GetOrchestrators(numOrchestrators int, suspender comm
 		}
 		return caps.CompatibleWith(info.Capabilities)
 	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), getOrchestratorCutoffTime)
+	defer cancel()
```
yondonfu: I don't think we need this deferred cancel since cancel() will always be called on L132 already before this method returns.

server/broadcast.go
```diff
@@ -315,6 +322,7 @@ func selectOrchestrator(n *core.LivepeerNode, params *core.StreamParameters, cou
 		return nil, err
 	}
 
+	glog.Infof("Found %v orchestrators", len(tinfos))
```
yondonfu: Don't think we need this log statement since there is already one [here](https://github.com/livepeer/go-livepeer/blob/d187acd2caa12cf1fb24b76ceeb3d46027b4a3a9/discovery/discovery.go#L137) that will be logged right before this with the same information.

discovery/discovery.go
```diff
@@ -17,7 +17,9 @@ import (
 	"github.com/golang/glog"
 )
 
-var getOrchestratorsTimeoutLoop = 3 * time.Second
+const MinWorkingSetSize = 8
```
yondonfu, line 20: Looks like we're introducing a min working set size here to determine when to pop Os off the suspension queue. I think an alternative that would avoid explicitly introducing a min working set size variable/function is the following:

---

Make the following update to the discovery loop:

```golang
// This loop runs until either all orchestrators have responded or a timeout
for i := 0; i < numAvailableOrchs && !timeout; i++ {
    // ...
}
```

Then, only the caller of `GetOrchestrators()` needs to be aware of the min working set size (which is the case right now outside of this PR) instead of making the `GetOrchestrators()` function aware of a min working set size variable/function.

The end result should be the same as what is currently implemented in this PR. The caller specifies the min working set size via `numOrchestrators` and `GetOrchestrators()` will fetch as many responses as possible from all Os, but if a bunch of Os are suspended such that the number of non-suspended Os is less than `numOrchestrators`, we'll pop Os off the suspension queue. The caller would be BroadcastSessionsManager which sets its `numOrchs` field to the min working set size which is passes as `numOrchestrators` in the `GetOrchestrators()` call.

WDYT?

yondonfu (changes requested):

discovery/discovery.go
kyriediculous: What you mention makes sense, the main reason I went with this (and I'm open to alternatives) is because if we select as many O's as possible e.g. 50 , and 5 don't respond. We'll still have a working set of 45 which seems sufficient enough to not have to pop on suspended O's until there's 50 O's again. It might make more sense to do that only when the working set would otherwise get too small.

yondonfu: Right. So if `numOrchestrators` reflected a min working set size that is smaller than 50 (i.e. 8), which is the # of available Os in this scenario, defined in BroadcastSessionsManager then we wouldn't pop any Os from the suspension queue in the scenario that you described.

yondonfu:
A variant of this was merged to master previously so closing.

PR #1877, opened 2021-05-06
pm: refresh ticket params before expiry from sender side

### LINKED ISSUE #1852: Refresh session when within the last 10% of ticket params validity period

At the moment, B will refresh a session (i.e. call GetOrchestrator on O to get a fresh OrchestratorInfo message) if the O's [ticket params are expired](https://github.com/livepeer/go-livepeer/blob/7a0fd1f6b958d61177fc01c488972f54b8bad1e3/server/broadcast.go#L854). However, ticket params can expire while a segment is in flight to O. If this happens, the O will return a TicketParams expired error when the segment is received and B will temporarily suspend O for returning an error.

B should refresh a session if O's ticket params are *almost* expired. B refreshes a session if O's auth token is within 10% of the [end of its validity period](https://github.com/livepeer/go-livepeer/blob/7a0fd1f6b958d61177fc01c488972f54b8bad1e3/server/broadcast.go#L845). We can do something similar with ticket params where B refreshes a session if O's ticket params are within 10% of the end of its validity period.

kyriediculous:
Refresh TicketParams for the active orchestrator session on the broadcaster at 10% before they expire. 

- Updated the expiry check in `sender.ValidateTicketParams`

- Added unit tests

Fixes #1852 

**Checklist:**

### DIFF
--- a/pm/recipient.go
+++ b/pm/recipient.go
@@ -22,6 +22,7 @@ var errInsufficientSenderReserve = errors.New("insufficient sender reserve")
 var maxWinProb = new(big.Int).Sub(new(big.Int).Lsh(big.NewInt(1), 256), big.NewInt(1))
 
 var paramsExpirationBlock = big.NewInt(10)
+var paramsExpiryBuffer = int64(1)
 
 // Recipient is an interface which describes an object capable
 // of receiving tickets
--- a/pm/sender.go
+++ b/pm/sender.go
@@ -155,7 +155,9 @@ func (s *sender) validateTicketParams(ticketParams *TicketParams, numTickets int
 	}
 
 	latestBlock := s.timeManager.LastSeenBlock()
-	if ticketParams.ExpirationBlock.Cmp(latestBlock) <= 0 {
+
+	currentBuffer := new(big.Int).Sub(ticketParams.ExpirationBlock, latestBlock).Int64()
+	if currentBuffer <= paramsExpiryBuffer {
 		return ErrTicketParamsExpired
 	}
 
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -8,6 +8,8 @@
 
 #### Broadcaster
 
+- \#1877 Refresh TicketParams for the active session before expiry (@kyriediculous)
+
 #### Orchestrator
 
 #### Transcoder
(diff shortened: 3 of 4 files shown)


pm/recipient.go
```diff
@@ -22,6 +22,7 @@ var errInsufficientSenderReserve = errors.New("insufficient sender reserve")
 var maxWinProb = new(big.Int).Sub(new(big.Int).Lsh(big.NewInt(1), 256), big.NewInt(1))
 
 var paramsExpirationBlock = big.NewInt(5)
+var paramsExpiryBuffer, _ = new(big.Float).Mul(new(big.Float).SetInt(paramsExpirationBlock), big.NewFloat(0.9)).Int64()
```
yondonfu: A problem I see with this method of calculating a buffer before the expiration is that it explicitly depends on `paramsExpirationBlock` which is determined by O. While this is a hardcoded variable, there is no guarantee that B & O will have the same view of the value of this variable since they could be running different versions of go-livepeer [1]. For example, it could be the case that after #1873 `paramsExpirationBlock = 10` on B, but an O is still using an older version of go-livepeer where `paramsExpirationBlock = 5`).

What if we just return `ErrTicketParamsExpired` if we are within 1-2 blocks of the expiration block?

[1] I realize that this is also a problem for the [auth token](https://github.com/livepeer/go-livepeer/blob/master/server/broadcast.go#L846) buffer calculation

kyriediculous: fixed in 2a26c888

kyriediculous: > [1] I realize that this is also a problem for the auth token buffer calculation

Want to address this in a separate issue ?

pm/sender.go
```diff
@@ -155,7 +155,9 @@ func (s *sender) validateTicketParams(ticketParams *TicketParams, numTickets int
 	}
 
 	latestBlock := s.timeManager.LastSeenBlock()
-	if ticketParams.ExpirationBlock.Cmp(latestBlock) <= 0 {
+
+	currentBuffer := new(big.Int).Sub(latestBlock, ticketParams.ExpirationBlock).Int64()
```
yondonfu: Nit: If we're defining a buffer as the # of blocks before the expiration block could the following be slightly more readable?

```golang
currentBuffer := new(big.Int).Sub(ticketParams.ExpirationBlock, latestBlock).Int64()
if currentBuffer <= paramsExpiryBuffer {
    ...
}
```

Otherwise, the current code evaluates to a negative buffer in the normal case and only evaluates to a positive buffer when `ErrTicketParamsExpired` should be returned.

pm/recipient.go
yondonfu: > Want to address this in a separate issue ?

👍

yondonfu (approved):
LGTM after squashing & fixing merge conflict.

PR #1879, opened 2021-05-09
add mp4 download

### LINKED ISSUE #1855: Allow to download recorded video in MP4 format
Right now /recordings endpoint constructs HLS manifest with .ts segments in it. We also want to download recording in MP4 format.
We will add two URL query params: `format` and `rendition`. If `format=mp4` then we will take `rendition` (source by default), will read all the .ts segments one-by-one, transmux them to mp4 on the fly and return that mp4.

darkdarkdragon:
Adds ability to download recorded stream as .mp4 file

`/recordings/` endpoint handler changed to recognise .mp4 extension. For media streams (like `/recordings/stream-id/240p0.mp4`) it will read stored .ts segments one-by-one, transmux them to .mp4 format on the fly and write it as the request output. 

Manually

Fixes #1855 

**Checklist:**


server/mediaserver.go
```diff
@@ -1037,6 +1038,83 @@ func getPlaylistsFromStore(ctx context.Context, sess drivers.OSSession, manifest
...
+			VideoEncoder: ffmpeg.ComponentOptions{
+				Name: "copy",
+			},
+			AudioEncoder: ffmpeg.ComponentOptions{
+				Name: "copy",
+			},
+			Profile: ffmpeg.VideoProfile{Format: ffmpeg.FormatNone},
+			Muxer: ffmpeg.ComponentOptions{
+				Name: "mp4",
+				Opts: map[string]string{"movflags": "frag_keyframe+negative_cts_offsets+omit_tfhd_offset+disable_chpl+default_base_moof"},
```
yondonfu, line 1085: Can we add a comment with a brief description of why we need these options?

server/mediaserver.go
```diff
@@ -1037,6 +1038,83 @@ func getPlaylistsFromStore(ctx context.Context, sess drivers.OSSession, manifest
...
+			},
+		},
+	}
+	for _, seg := range jpl.Segments[track] {
+		if seg.GetDiscontinuity() {
+			tc.Discontinuity()
+		}
+
+		ir, iw, err := os.Pipe()
```
yondonfu, line 1094: Instead of creating a new pipe, writing the downloaded segment data into the pipe, calling `Transcode()` and closing the pipe for each segment, could it make sense to create a single input pipe, write all downloaded segment data into the pipe and call `Transcode()` once? Something like:

```golang
ir, iw, err := os.Pipe()
if err != nil { ... }

defer func() {
    tc.StopTranscoder()
    ow.Close()
    iw.Close()
}

for _, seg := range jpl.Segments[track] {
    ...
    go func() {
        // Download segment data
       _, err = io.Copy(iw, resp.Body)
       if err != nil { ... }
    }
}

fname := fmt.Sprintf("pipe:%d", ir.Fd())
in := &ffmpeg.TranscodeOptions{Fname: fname, Transmuxing: true}
_, err = tc.Transcode(in, out)
if err != nil { ... }
```

If this works, the main benefit I see of this approach is a slightly cleaner separation of logic within the function i.e. for loop responsible for just downloading/writing segment data into a single pipe and then everything else is handled by the input/output pipe + Transcode() call.

server/mediaserver.go
```diff
@@ -1037,6 +1038,83 @@ func getPlaylistsFromStore(ctx context.Context, sess drivers.OSSession, manifest
...
+		if err != nil {
+			glog.Error(err)
+			w.WriteHeader(http.StatusInternalServerError)
+			return
+		}
+		fname := fmt.Sprintf("pipe:%d", ir.Fd())
+
+		in := &ffmpeg.TranscodeOptionsIn{Fname: fname, Transmuxing: true}
+		go func() {
+			glog.V(common.VERBOSE).Infof("Adding uri=%s to mp4", seg.URI)
```
yondonfu: Should we include an identifier for the recording in these logs? Could this be `track`?

server/mediaserver.go
```diff
@@ -1037,6 +1038,83 @@ func getPlaylistsFromStore(ctx context.Context, sess drivers.OSSession, manifest
...
+				return
+			}
+			defer resp.Body.Close()
+			if resp.StatusCode != 200 {
+				glog.Errorf("Non-200 response for status=%v uri=%s request=%s", resp.Status, seg.URI, r.URL.String())
+				return
+			}
+			_, err = io.Copy(iw, resp.Body)
+			if err != nil {
+				glog.Errorf("Error transmuxing to mp4 request=%s seg=%s err=%v", r.URL.String(), seg.URI, err)
```
yondonfu: Should we use the same key `uri=` for `seg.URI` as is used in the other logs?

server/mediaserver.go
```diff
@@ -1037,6 +1038,83 @@ func getPlaylistsFromStore(ctx context.Context, sess drivers.OSSession, manifest
...
+			}
+			iw.Close()
+		}()
+		_, err = tc.Transcode(in, out)
+		if err != nil {
+			glog.Errorf("Error transmuxing to mp4 request=%s seg=%s err=%v", r.URL.String(), seg.URI, err)
```
yondonfu: Same comment as above about the `uri=` key.

server/mediaserver.go
darkdarkdragon: fixed

server/mediaserver.go

server/mediaserver.go

server/mediaserver.go

server/mediaserver.go
darkdarkdragon: No, because we need to signal discontinuity to transcoding loop. 
Say, user streamed for 10 min, stopped, and then immediately started again and streamed for another 10 min.
We will have this as single recording, but timestamps of the second parted will be started from zero.
To make from that one mp4 file we need to shift timestamps of second part to make on continuous stream.
To do that we need a way to signal that discontinuity to the transcoding loop - which is not possible if have one loop for the whole transmuxing process.
That's why I've changed lpms so for transmuxing it closes input after Transcode call, but keeps open output.
That way I can signal discontiniuty to the transmuxer between Transcode calls.
 https://github.com/livepeer/go-livepeer/blob/d87fcc6006af3684c38947a9fd7b831315b703e3/server/mediaserver.go#L1079-L1081

server/mediaserver.go
```diff
@@ -1037,6 +1038,85 @@ func getPlaylistsFromStore(ctx context.Context, sess drivers.OSSession, manifest
...
+			}
+			_, err = io.Copy(iw, resp.Body)
+			if err != nil {
+				glog.Errorf("Error transmuxing to mp4 request=%s uri=%s err=%v", r.URL.String(), seg.URI, err)
+			}
+			iw.Close()
+		}()
+		_, err = tc.Transcode(in, out)
+		if err != nil {
+			glog.Errorf("Error transmuxing to mp4 request=%s uri=%s err=%v", r.URL.String(), seg.URI, err)
```
yondonfu: If we have access to the manifestID in this function now then let's include it in all the log statements.

server/mediaserver.go
```diff
@@ -1037,6 +1038,85 @@ func getPlaylistsFromStore(ctx context.Context, sess drivers.OSSession, manifest
...
+		if err != nil {
+			glog.Error(err)
+			w.WriteHeader(http.StatusInternalServerError)
+			return
+		}
+		fname := fmt.Sprintf("pipe:%d", ir.Fd())
+
+		in := &ffmpeg.TranscodeOptionsIn{Fname: fname, Transmuxing: true}
+		go func() {
+			glog.V(common.VERBOSE).Infof("Adding mid=%s track=%s uri=%s to mp4", manifestID, track, seg.URI)
```
yondonfu: I *think* `manifestID=` is more common at least in the `server` package. For the sake of consistency let's use that?

server/mediaserver.go
```diff
@@ -1037,6 +1038,85 @@ func getPlaylistsFromStore(ctx context.Context, sess drivers.OSSession, manifest
...
+			defer resp.Body.Close()
+			if resp.StatusCode != 200 {
+				glog.Errorf("Non-200 response for status=%v uri=%s request=%s", resp.Status, seg.URI, r.URL.String())
+				return
+			}
+			_, err = io.Copy(iw, resp.Body)
+			if err != nil {
+				glog.Errorf("Error transmuxing to mp4 request=%s uri=%s err=%v", r.URL.String(), seg.URI, err)
+			}
+			iw.Close()
```
yondonfu: Do we need to make sure Close() is called if we hit an error condition earlier on in this goroutine?

server/mediaserver.go
```diff
@@ -1037,6 +1038,85 @@ func getPlaylistsFromStore(ctx context.Context, sess drivers.OSSession, manifest
...
+		}()
+		_, err = tc.Transcode(in, out)
+		if err != nil {
+			glog.Errorf("Error transmuxing to mp4 request=%s uri=%s err=%v", r.URL.String(), seg.URI, err)
+			w.WriteHeader(http.StatusInternalServerError)
+			return
+		}
+		ir.Close()
```
yondonfu: Do we need to make sure Close() is called if we hit an error condition prior to this?

server/mediaserver.go
```diff
@@ -1037,6 +1038,85 @@ func getPlaylistsFromStore(ctx context.Context, sess drivers.OSSession, manifest
...
 }
 
+func (s *LivepeerServer) streamMP4(w http.ResponseWriter, r *http.Request, jpl *core.JsonPlaylist, manifestID, track string) {
+	w.Header().Set("Access-Control-Allow-Origin", "*")
+	contentType, _ := common.TypeByExtension(".mp4")
+	w.Header().Set("Content-Type", contentType)
+	w.Header().Set("Connection", "keep-alive")
+
+	tc := ffmpeg.NewTranscoder()
+	or, ow, err := os.Pipe()
```
yondonfu, line 1048: Do we need to eventually Close() `or` too?

yondonfu (changes requested):

server/mediaserver.go
darkdarkdragon: fixed in 50daabe9c6c91455a8f63f9913df6aca58e86494

server/mediaserver.go
darkdarkdragon: fixed in 50daabe9c6c91455a8f63f9913df6aca58e86494

server/mediaserver.go
darkdarkdragon: fixed in 50daabe9c6c91455a8f63f9913df6aca58e86494

server/mediaserver.go
darkdarkdragon: fixed in 50daabe9c6c91455a8f63f9913df6aca58e86494

server/mediaserver.go
darkdarkdragon: Nice catch!
I've also changed how this function's exit handled - I think it was causing errors I've saw sometimes - when last couple of kb of resulting mp4 was not written to the network.
Now, before exit, it waits for Copy function to complete.

server/mediaserver.go
```diff
@@ -1037,6 +1038,97 @@ func getPlaylistsFromStore(ctx context.Context, sess drivers.OSSession, manifest
...
+	w.Header().Set("Access-Control-Allow-Origin", "*")
+	contentType, _ := common.TypeByExtension(".mp4")
+	w.Header().Set("Content-Type", contentType)
+	w.Header().Set("Connection", "keep-alive")
+	var sourceBytesSent, resultBytesSent int64
+
+	tc := ffmpeg.NewTranscoder()
+	or, ow, err := os.Pipe()
+	if err != nil {
+		glog.Errorf("manifestID=%s error creating pipe err=%v", manifestID, err)
```
yondonfu:
```suggestion
		glog.Errorf("Error creating pipe manifestID=%s err=%v", manifestID, err)
```

server/mediaserver.go
```diff
@@ -1037,6 +1038,97 @@ func getPlaylistsFromStore(ctx context.Context, sess drivers.OSSession, manifest
...
 }
 
+func (s *LivepeerServer) streamMP4(w http.ResponseWriter, r *http.Request, jpl *core.JsonPlaylist, manifestID, track string) {
+	w.Header().Set("Access-Control-Allow-Origin", "*")
+	contentType, _ := common.TypeByExtension(".mp4")
+	w.Header().Set("Content-Type", contentType)
+	w.Header().Set("Connection", "keep-alive")
+	var sourceBytesSent, resultBytesSent int64
+
+	tc := ffmpeg.NewTranscoder()
```
yondonfu: I think we should move this after the pipe creation. Otherwise, if there is an error during pipe creation, while unlikely, the internal handle of the transcoder will never be cleared by a call to `StopTranscoder()`.

server/mediaserver.go
```diff
@@ -1037,6 +1038,97 @@ func getPlaylistsFromStore(ctx context.Context, sess drivers.OSSession, manifest
...
+	or, ow, err := os.Pipe()
+	if err != nil {
+		glog.Errorf("manifestID=%s error creating pipe err=%v", manifestID, err)
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+	done := make(chan struct{})
+	go func() {
+		resultBytesSent, _ = io.Copy(w, or)
```
yondonfu: Handle the error returned by `io.Copy`?

server/mediaserver.go
```diff
@@ -1037,6 +1038,97 @@ func getPlaylistsFromStore(ctx context.Context, sess drivers.OSSession, manifest
...
+		_, err = tc.Transcode(in, out)
+		ir.Close()
+		if err != nil {
+			glog.Errorf("Error transmuxing to mp4 request=%s uri=%s manifestID=%s err=%v", r.URL.String(), seg.URI, manifestID, err)
+			stop()
+			return
+		}
+	}
+	stop()
+	glog.Errorf("For mp4 request=%s manifestID=%s sent source bytes=%d destBytes=%d", r.URL.String(),
```
yondonfu:
```suggestion
	glog.Infof("Completed mp4 request=%s manifestID=%s sourceBytes=%d destBytes=%d", r.URL.String(),
```

server/mediaserver.go
```diff
@@ -1037,6 +1038,97 @@ func getPlaylistsFromStore(ctx context.Context, sess drivers.OSSession, manifest
...
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+	done := make(chan struct{})
+	go func() {
+		resultBytesSent, _ = io.Copy(w, or)
+		or.Close()
+		done <- struct{}{}
+	}()
+	stop := func() {
```
yondonfu: Can we just defer this?

server/mediaserver.go
darkdarkdragon: fixed in e3a1a3f3925366193b21ad0cfa51d7f3e4cb626c

server/mediaserver.go
darkdarkdragon: fixed in e3a1a3f3925366193b21ad0cfa51d7f3e4cb626c

server/mediaserver.go
darkdarkdragon: changed to defer in e3a1a3f3925366193b21ad0cfa51d7f3e4cb626c

server/mediaserver.go
```diff
@@ -1037,6 +1038,98 @@ func getPlaylistsFromStore(ctx context.Context, sess drivers.OSSession, manifest
...
+		},
+	}
+	for _, seg := range jpl.Segments[track] {
+		if seg.GetDiscontinuity() {
+			tc.Discontinuity()
+		}
+
+		ir, iw, err := os.Pipe()
+		if err != nil {
+			glog.Errorf("manifestID=%s error creating pipe err=%v", manifestID, err)
```
yondonfu:
```suggestion
			glog.Errorf("Error creating pipe manifestID=%s err=%v", manifestID, err)
```

server/mediaserver.go
```diff
@@ -1037,6 +1038,98 @@ func getPlaylistsFromStore(ctx context.Context, sess drivers.OSSession, manifest
...
+			defer resp.Body.Close()
+			if resp.StatusCode != 200 {
+				glog.Errorf("Non-200 response for status=%v uri=%s manifestID=%s request=%s", resp.Status, seg.URI, manifestID, r.URL.String())
+				return
+			}
+			wn, err := io.Copy(iw, resp.Body)
+			if err != nil {
+				glog.Errorf("Error transmuxing to mp4 request=%s uri=%s manifestID=%s err=%v", r.URL.String(), seg.URI, manifestID, err)
+			}
+			sourceBytesSent += wn
```
yondonfu, line 1119: Don't think this is concurrency safe since multiple goroutines can concurrently try to update `sourceBytesSent`?

If this is the case, could do something like this instead:

```golang
sourceBytesSentCh := make(chan int64, len(jpl.Segments[track]))

for _, seg := range jpl.Segments[track] {
    var wn int64
    var err error

    defer func() {
        iw.Close()
        sourceBytesSentCh <- wn
    }()
    
    // Other stuff
    ...

    wn, err = io.Copy(iw, resp.Body)
    ...
    // When defer statement executes, wn should be populated
    // If we error out before calling io.Copy, wn will be 0
}

sourceBytesSent := 0
for sourceBytes := range <-sourceBytesSentCh {
    sourceBytesSent += sourceBytes
}
```

darkdarkdragon: > Don't think this is concurrency safe since multiple goroutines can concurrently try to update sourceBytesSent?
It can't be update in parallel - it is only changed in one place, and it called strictly sequentially.

yondonfu: Ah right and `sourceBytesSent` is only read once all goroutines finish since the `Transcode()` call blocks until the input for the pipe is closed which only happens when the goroutine finishes which means the `sourceBytesSent` update is done.

yondonfu (approved):
LGTM after squashing

PR #1880, opened 2021-05-12
pm: check winning ticket transactions until confirm or error

### LINKED ISSUE #1806: pm: Do not submit a new tx for a winning ticket if another tx is already pending

Currently, if gas prices are rising, then its possible for a tx to be submitted for a ticket at the head of the queue. But, that tx will sit pending for awhile and internally the node will get a context timeout from waiting for the tx to confirm. Then, the node will try to submit the ticket again because it has not been removed from the head of the queue. This second tx will confirm after the first pending tx confirms and will revert because the ticket will already be used on-chain by the first tx.

The most basic thing to ensure first is that there are never multiple in-flight txs for a single tx.

An improvement might be to track in-flight txs for tickets and then replace them with a higher gas price similar to what is done for reward transactions.

kyriediculous:
Don't return from `senderMonitor.redeemWinningTicket` when context times out, loop instead until a concrete response is returned (other error or non-nil tx)

Fixes #1806 

**Checklist:**

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -32,6 +32,7 @@
 
 - \#1860 Discard low gas prices to prevent insufficient ticket faceValue errors (@kyriediculous)
 - \#1859 Handle error for invalid inferred orchestrator public IP on node startup (@reubenr0d)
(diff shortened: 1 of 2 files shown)


yondonfu:
@kyriediculous I don't think this PR is necessary any more since #1923 has been merged right?

kyriediculous:
That's right !

PR #1893, opened 2021-05-18
multi: Remove high card keys for payment metrics

### LINKED ISSUE #1763: Monitoring: Too many data points for "tickets received"
If livepeer is run with the `-monitor` flag, it reports every single ticket received - a different series for every manifest. This causes problems in grafana since there are just too many data points for it to handle. E.g. selecting a longer timeframe (12h+) just crashes grafana. 

I don't know how the metric collection/push happens "under the hood", but I suppose only reporting the total tickets received as a single metric would work (ignoring different manifestIDs).

I've already tried limiting the number of data points and trying calculations/aggregations like rate/deriv etc. Doesn't seem to help (or I didn't apply it correctly...)

**Additional context**
image

yondonfu:

Remove high cardinality keys i.e. recipient/sender and manifestID from payment/ticket metrics. Now, only aggregate payment/ticket metrics are reported and there is no metrics reporting on a per recipient/sender or manifestID basis.

Prometheus doesn't like high cardinality keys/labels so when using Grafana to load Prometheus data a dashboard that uses payment/ticket metrics the dashboard might be really slow to render or may even cause the Grafana app to crash as reported in #1763.

- Remove high cardinality keys from payment/metrics reporting functions in `monitor` package
- Update arguments passed into payment/metrics reporting function calls in other packages

TBD.

Fixes #1763 

**Checklist:**

### DIFF
--- a/monitor/census.go
+++ b/monitor/census.go
@@ -606,21 +606,21 @@ func InitCensus(nodeType NodeType, version string) {
 			Name:        "ticket_value_sent",
 			Measure:     census.mTicketValueSent,
 			Description: "Ticket value sent",
-			TagKeys:     append([]tag.Key{census.kRecipient, census.kManifestID}, baseTags...),
+			TagKeys:     baseTags,
 			Aggregation: view.Sum(),
 		},
 		{
 			Name:        "tickets_sent",
 			Measure:     census.mTicketsSent,
 			Description: "Tickets sent",
-			TagKeys:     append([]tag.Key{census.kRecipient, census.kManifestID}, baseTags...),
+			TagKeys:     baseTags,
 			Aggregation: view.Sum(),
 		},
 		{
 			Name:        "payment_create_errors",
 			Measure:     census.mPaymentCreateError,
 			Description: "Errors when creating payments",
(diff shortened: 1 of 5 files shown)


monitor/census.go
```diff
@@ -643,21 +643,21 @@ func InitCensus(nodeType NodeType, version string) {
...
 			Description: "Tickets received",
-			TagKeys:     append([]tag.Key{census.kSender, census.kManifestID}, baseTags...),
+			TagKeys:     baseTags,
 			Aggregation: view.Sum(),
 		},
 		{
 			Name:        "payment_recv_errors",
 			Measure:     census.mPaymentRecvErr,
 			Description: "Errors when receiving payments",
-			TagKeys:     append([]tag.Key{census.kSender, census.kManifestID, census.kErrorCode}, baseTags...),
```
darkdarkdragon, line 660: `kErrorCode` is still used, so it shouldn't be removed

darkdarkdragon (changes requested):
small change

monitor/census.go
yondonfu: Good catch! Fixed in 9889725

darkdarkdragon (approved):
LGTM

PR #1899, opened 2021-05-21
monitor+server: Record million pixels processed metric

yondonfu:

Adds a million pixels processed metric. For now, this is only recorded on B by taking the total pixel count for a set of transcoded results, dividing it by 1m and then recording the result. This metric is helpful for monitoring pixel throughput on a B while also monitoring payment/ticket throughput since the latter should be based on the former.

The risk of overflow while using a float64 to record the metric should be minimal. Consider the following example:

- 179729280000 pixels per hour for a stream with 240p30fps, 360p30fps, 480p30fps, 720p30fps outputs (does not factor in decoded pixels from the input)
- 179729280000 / 1000000 = 179729.28 million pixels per hour
- 179729.28 million pixels per hour * 24 hours * 365 days = 1574428492.8 million pixels per year
- 1574428492.8 million pixels per year * 1000000 streams = 1574428492800000.0

So, with 1000000 streams running 24/7 for an entire year the million pixels value should still be below the max float64 value.

- Add `mil_pixels_processed` metric in the `monitor` package
- Record `mil_pixels_processed` metric on B after counting the pixels in the transcoded results

Manually.

N/A

**Checklist:**

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -10,6 +10,7 @@
 
 - \#1877 Refresh TicketParams for the active session before expiry (@kyriediculous)
 - \#1879 Add mp4 download of recorded stream (@darkdarkdragon)
+- \#1899 Record million pixels processed metric (@yondonfu)
 
 #### Orchestrator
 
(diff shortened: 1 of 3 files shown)


darkdarkdragon (approved):
LGTM

PR #1900, opened 2021-05-24
B/O/T segment processing with content-detection enabled

### LINKED ISSUE #1885: B/O/T content detection request processing
See:

- "Broadcaster Segment Submission"
- "Orchestrator Processing Detection Requests"
- "Transcoder Processing Detection Requests"

in https://github.com/livepeer/go-livepeer/issues/1881.

jailuthra:

See title

See commit history

Fixes #1885 

**Checklist:**

### DIFF
--- a/server/segment_rpc.go
+++ b/server/segment_rpc.go
@@ -557,18 +557,24 @@ func genSegCreds(sess *BroadcastSession, seg *stream.HLSSegment) (string, error)
 		storage = bos.GetInfo()
 	}
 
+	detectorProfiles := []ffmpeg.DetectorProfile{}
+	if sess.Params.Detection.Freq != 0 && seg.SeqNo%uint64(sess.Params.Detection.Freq) == 0 {
+		detectorProfiles = sess.Params.Detection.Profiles
+	}
+
 	// Generate signature for relevant parts of segment
 	params := sess.Params
 	hash := crypto.Keccak256(seg.Data)
 	md := &core.SegTranscodingMetadata{
-		ManifestID: params.ManifestID,
-		Seq:        int64(seg.SeqNo),
-		Hash:       ethcommon.BytesToHash(hash),
-		Profiles:   params.Profiles,
-		OS:         storage,
-		Duration:   time.Duration(seg.Duration * float64(time.Second)),
-		Caps:       params.Capabilities,
-		AuthToken:  sess.OrchestratorInfo.GetAuthToken(),
+		ManifestID:       params.ManifestID,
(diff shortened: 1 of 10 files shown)


server/segment_rpc.go
```diff
@@ -295,6 +296,31 @@ func makeFfmpegVideoProfiles(protoProfiles []*net.VideoProfile) ([]ffmpeg.VideoP
...
+func makeNetDetectData(ffmpegDetectData []ffmpeg.DetectData) []*net.DetectData {
+	netDataList := []*net.DetectData{}
+	for _, data := range ffmpegDetectData {
+		var netData *net.DetectData
+		switch data.Type() {
+		case ffmpeg.SceneClassification:
+			d := data.(ffmpeg.SceneClassificationData)
+			netClasses := make(map[uint32]*net.Fraction)
+			for classID, prob := range d {
+				netClasses[uint32(classID)] = &net.Fraction{
```
yondonfu, line 309:
```suggestion
				netClasses[uint32(classID)] = &net.Fraction{Num: 0, Denom: 1}
                                  rat := new(big.Rat).SetFloat64(prob)
                                  if rat != nil {
                                      netClasses[uint32(classID)].Num = rat.Num()
                                      netClasses[uint32(classID)].Denom = rat.Denom()
                                  }
```

^ Suggestion for converting the float64 probability into a fraction that doesn't require hardcoded conversation constants.

yondonfu:
I don't think split O/T is supported by this PR as there is some additional network message handling that will be necessary for that. That's okay for now and can be handled in a separate PR, but just noting for awareness.

server/segment_rpc.go
jailuthra: Hmm `rat.Num()` returns a `big.Int` which might not necessarily fit our `net.Fraction.Num` which is a `uint32`. https://play.golang.org/p/XLYKIR31Lvt

Even if we convert `big.Int` using the `Uint64` method, there is no guarantee that an arbitrary float we get from the C code will not overflow it -> the docs suggest runtime checks if it overflows via https://golang.org/pkg/math/big/#Int.IsUint64

Maybe we can stick with a hardcoded constant for now? At least till we figure out how to handle floats/fractions consistently across C->Go->Protobuf->JSON.

I will increase the constant from `1000` to `1e6` to retain more precision.

yondonfu: Good catch.

I suspect there might be easier ways to pass a float across the wire...Let's revisit this.

yondonfu (approved):
LGTM after squashing.

Looks like the failing test is a flaky one.

PR #1903, opened 2021-05-26
Generate valid MP4 files

### LINKED ISSUE #1897: Generate MP4 that is playable in QuickTime
MP4 we're generating now is not playable in QuickTime. 
There is official Apple's example of fragmented MP4 which is playable in QuickTime - https://developer.apple.com/streaming/examples/advanced-stream-fmp4.html
But it is not clear if it is possible to generate such file using ffmpeg.

darkdarkdragon:
Fixes MP4 generation to generate MP4 playable in QuickTime/default Windows player.

Enables `extract_extradata` filter in ffmpeg build. MPEG-TS format doesn't store SPS/PPS (information needed to initialize h264 decoder) in container itself - so this information is only passed inside h264 bitsream. But MP4 container also store this information in the separate container field. `extract_extradata` filter used by ffmpeg to extract that data from h264 bitstream when parsing MPEG-TS streams. This is not needed during transcoding, but without that information generated MP4 had empty field that should've contained SPS/PPS - and that what confused QuickTime/Windows player.

Manually

Fixes #1897 

**Checklist:**

### DIFF
--- a/install_ffmpeg.sh
+++ b/install_ffmpeg.sh
@@ -126,7 +126,7 @@ if [ ! -e "$HOME/ffmpeg/libavcodec/libavcodec.a" ]; then
     --enable-gnutls --enable-libx264 --enable-gpl \
     --enable-protocol=https,http,rtmp,file,pipe \
     --enable-muxer=mpegts,hls,segment,mp4,null --enable-demuxer=flv,mpegts,mp4,mov \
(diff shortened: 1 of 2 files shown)


yondonfu (approved):
LGTM. I think its fine to omit a pending changelog entry here since the mp4 download feature was not included in a published release yet.

PR #1905, opened 2021-06-01
Add 'Content-Disposition' header to the /recordings  mp4 response. Us…

darkdarkdragon:
`Content-Disposition` header forces browsers to download file instead of trying to play it.
Usage of last part of request URL as filename allows to form URLs like this:
`livepeer.com/recordings/sessionId/source/someFileName.mp4`
That way we can form URL (on frontend) that will contain user defined stream name and session creation timestamp without need to pass this information to go-livepeer node by other means.

Add 'Content-Disposition' header to the /recordings  mp4 response. Use last part
of the request URL as file name in 'Content-Disposition' response.

Manually

**Checklist:**

### DIFF
--- a/server/mediaserver.go
+++ b/server/mediaserver.go
@@ -1038,11 +1038,12 @@ func getPlaylistsFromStore(ctx context.Context, sess drivers.OSSession, manifest
 	return filesMap, jsonFiles, latestPlaylistTime, nil
 }
 
-func (s *LivepeerServer) streamMP4(w http.ResponseWriter, r *http.Request, jpl *core.JsonPlaylist, manifestID, track string) {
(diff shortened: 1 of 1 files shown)


server/mediaserver.go
```diff
@@ -1308,7 +1309,7 @@ func (s *LivepeerServer) HandleRecordings(w http.ResponseWriter, r *http.Request
 			w.WriteHeader(http.StatusNotFound)
 			return
 		}
-		s.streamMP4(w, r, mainJspl, manifestID, track)
+		s.streamMP4(w, r, mainJspl, manifestID, track, pp[len(pp)-1])
```
iameli, line 1312: Cool - this means that `source.mp4` as well as `source/streamName.mp4` will both work, yeah? Sounds great.

iameli (approved):

server/mediaserver.go
darkdarkdragon: Yes

PR #1910, opened 2021-06-08
[TESTING] e2e scene classification

yondonfu:


yondonfu:
**Notes**

9a5bf65 can be dropped after [this comment](https://github.com/livepeer/FFmpeg/pull/1#issuecomment-857321554) is addressed.

45b5c88 adds support for `-sceneClassificationModelPath` flag which can be used to specify the path to the model which replaces the `-sceneClassification` boolean flag which was introduced in a previous commit. However, the value for this flag currently needs to match the [ModelPath field](https://github.com/oscar-davids/lpms/blob/a784401db320470de804cb1591cdac6438e815b2/ffmpeg/detector.go#L35) for the hardcoded scene classification profiles (unless the model exists at two different paths - one that is used for loading at startup and one that is used when running the lvpdnn filter). We should improve the usability here later.

Linux builds are passing right now, but I had to make sure to install libtensorflow whenever a binary is run because libtensorflow is a run time dependency regardless of how the binary is used (broadcaster, orchestrator, etc.).

Windows & Mac builds are currently failing due to problems with the libtensorflow dependency.

PR #1901, opened 2021-05-24
cmd,server: Post detection results to webhook

jailuthra:
**What does this pull request do? Explain your changes. (required)**

See title

**Specific updates (required)**

See commit history

**How did you test each of these updates (required)**

**Does this pull request close any open issues?**

Fixes #1886 

**Checklist:**

- [x] Read the [contribution guide](./doc/contributing.md)
- [x] `make` runs successfully
- [x] All tests in `./test.sh` pass
- [x] README and other documentation updated
- [ ] [Pending changelog](./CHANGELOG_PENDING.md) updated

PR #1895, opened 2021-05-19
core: enable experimental capabilities

jailuthra:
**What does this pull request do? Explain your changes. (required)**

Enable extra capabilities if the go build tag `experimental` is set.

**Specific updates (required)**

*see commit history*

**How did you test each of these updates (required)**

**Does this pull request close any open issues?**

Fixes #1882 

**Checklist:**

- [x] Read the [contribution guide](./doc/contributing.md)
- [x] `make` runs successfully
- [x] All tests in `./test.sh` pass
- [x] README and other documentation updated
- [ ] [Pending changelog](./CHANGELOG_PENDING.md) updated

PR #1911, opened 2021-06-09
Enable content detection

### LINKED ISSUE #1881: Content Detection Integration (Experimental)
## Abstract

An experimental content detection integration will allow transcoders to run content detection as a part of the transcoding pipeline.

## Motivation

Content detection will run as a part of the filtergraph after input video is decoded and before passing video frames to the encoder. This [PR](https://github.com/livepeer/lpms/pull/228) adds support for content detection in LPMS. Since this feature is still early and there are still a number of things that need to be done before it can be used on the public network in a trustless/trust minimized manner, we want to treat it as experimental and this spec describes steps that can be taken to make the experimental nature of the feature clear.

## Proposed Solution

The scope of this spec includes:

- Experimental capability support
- Protobuf message updates
- Broadcaster stream auth webhook to configure detection [1]
- Broadcaster segment submission with detection requests
- Orchestrator processing detection requests
- Transcoder processing detection requests
- Broadcaster detection webhook [2]

[1] While detection could be configured using other options, similar to how video profiles can be configured using a variety of options, for now we'll just focus on the stream auth webhook.

[2] We'll eventually want to expose the detection results in the multi-part response returned by the broadcaster's HTTP push API which is currently used by the Livepeer MistServer plugin, but this feature is not in scope for this spec right now.

### Experimental Capability Support

Currently, a [hardcoded list of default capabilities](https://github.com/livepeer/go-livepeer/blob/b933f4bf977a529814dd3bf89dfbc89efd524380/cmd/livepeer/livepeer.go#L71) is defined in `cmd/livepeer/livepeer.go` which is used to [set the capabilities of an orchestrator](https://github.com/livepeer/go-livepeer/blob/b933f4bf977a529814dd3bf89dfbc89efd524380/cmd/livepeer/livepeer.go#L854).

In order to incorporate experimental capabilities into the capability configuration process, we can define a package level function in `core` that returns a combination of default and experimental capabilities:

```golang=
func DefaultCapabilities() []Capability {
    return append(defaultCapabiltiies, experimentalCapabilities)
}
```

`experimentalCapabilities` can be conditionally set depending on whether a build uses the `experimental` tag (i.e. `go build ... -tags=experimental`).

In `core/capabilities_experimental_on.go`, we can include any capabilities that are currently experimental:

```golang=
+build experimental

var experimentalCapabilities = []Capability{
    Capability_SceneClassification,
}
```

In `core/capabilities_experimental_off.go`, `experimentalCapabilities` will just be an empty slice:

```golang=
+build !experimental

var experimentalCapabilities = []Capability{}
```

### Protobuf Messages

The following messages will need to be added/updated in [net/lp_rpc.proto](https://github.com/livepeer/go-livepeer/blob/master/net/lp_rpc.proto).

```proto=
// [EXPERIMENTAL]
// Describes a class defined by a trained model to detect
message DetectorClass {
    // ID of the class to detect
    uint32 class_id = 0;
    // Name of the class in the trained model to detect
    string class_name = 1;
}

// [EXPERIMENTAL]
// Describes content detection configuration
message DetectorProfile {    
    oneof value {
        SceneClassificationProfile scene_classification = 0;
        // Example for future:
        // ObjectDetectionProfile object_detection = 1; 
    }
}

// [EXPERIMENTAL]
// Describes scene classification configuration
message SceneClassificationProfile {
    uint32 sample_rate = 0;
    repeated DetectorClass classes = 1;
}

message SegData {
    ...
    
    repeated VideoProfile fullProfiles3 = 35;
    
    // [EXPERIMENTAL]
    repeated DetectorProfile detector_profiles = 36;
}
```

https://github.com/livepeer/lpms/pull/228 includes a few additional fields for the `DetectorProfile` Go struct, but we may be able to exclude them from the `DetectorProfile` protobuf message. I don't think the broadcaster needs to be exposed to the `ModelPath`, `Input` and `Output` fields which are specific to the model being executed by the orchestrator because in this initial implementation we are not including support for broadcasters to specify specific models - they will use the models available to the orchestrators.

```proto=
message Fraction {
    uint32 num = 0;
    uint32 denom = 1;
}

// [EXPERIMENTAL]
// Describes detection results
message DetectData {
    oneof value {
        SceneClassificationData scene_classification = 0;
        // Example for future:
        // ObjectDetectionData object_detection = 1;
    }
}

// [EXPERIMENTAL]
// Describes scene classification results
message SceneClassificationData {
    map<uint32, Fraction> class_probs = 0;
}

message TranscodeData {    
    repeated TranscodedSegmentData segments = 1;
    
    bytes sig = 2;
    
    // [EXPERIMENTAL]
    // Detection result data in the order of the detector profiles for the job
    repeated DetectData detections = 3;
}
```

All protobuf messages related to content detection are labeled with the `[EXPERIMENTAL]` comment to indicate that they are associated with an experimental feature/capability that is going through rapid iteration and to signal that backwards incompatible changes are to be expected.

### Broadcaster Stream Auth Wehbook

The [authWebhookResponse struct](https://github.com/livepeer/go-livepeer/blob/3610c2e763dc0907d278484518fb085a8bbfa2c2/server/mediaserver.go#L101) used to store data from the stream auth webhook, which is executed at the beginning of a stream, will need to be updated to support detector profile configurations:

```golang=
type authWebhookResponse struct {
    ...
    Profiles []struct {
        ...
    } `json:"profiles"`
    Detection struct {
        // Run detection on 1 / freq segments
        Freq int `json:"freq"`
        
        SceneClassificationProfile struct {
            SampleRate int `json:"sampleRate"`
            Classes []struct {
                ID int `json:"id"`
                Name string `json:"name"`
            } `json:"classes"`
        } `json:"sceneClassificationProfile"`
        
    } `json:"detection"`
}
```

The [StreamParams](https://github.com/livepeer/go-livepeer/blob/b933f4bf977a529814dd3bf89dfbc89efd524380/core/streamdata.go#L25) struct will need to be updated to store the detection configuration:

```golang=
type DetectionConfig struct {
    Freq int
    DetectionProfiles []ffmpeg.DetectorProfile
}

type StreamParams struct {
    ...
    Detection DetectionConfig
}
```

The `DetectorProfile` message can nest one of many message types although the default will be the `SceneClassificationProfile` message. We can have the equivalent in Go mirror the message structure. We could accomplish this by defining a `DetectorProfile` interface that the `SceneClassificationProfile` type implements:

```golang=
type DetectorType int

const (
	DetectorType SceneClassification = iota
         // Example for future:
	// ObjectDetection
)

type DetectorProfile interface {
	Type() DetectorType
}

type SceneClassificationProfile struct {
	...
}

func (p *SceneClassificationProfile) Type() DetectorType {
	return SceneClassification
}
```

I think the upside of this approach is that we won't need to add a new field for each potential type of detection if there are multiple types of detection in the future. I imagine the alternative would end up looking something like:

```golang=
type SegTranscodingMetadata struct {
    ...
    SceneClassificationProfile
    ObjectDetectionProfile
    AudioTranscriptionProfile
    ...
}
```

Personally, I find the following to be a bit cleaner:

```golang=
type SegTranscodingMetadata struct {
    ...
    DetectionProfiles []ffmpeg.DetectionProfile
}

func ToSceneClassificationProfile(detectorProfile DetectorProfile) (ffmpeg.SceneClassificationProfile, error) {
    if detectorProfile.Type() != SceneClassification {
        return ffmpeg.SceneClassificationProfile{}, errors.New("not a SceneClassificationProfile")
    }
    return detectorProfile.(ffmpeg.SceneClassificationProfile)
}
```

When the auth webhook response is parsed, the detection config can be stored in the [StreamParams](https://github.com/livepeer/go-livepeer/blob/3610c2e763dc0907d278484518fb085a8bbfa2c2/server/mediaserver.go#L296) returned by `createRTMPStreamIDHandler`.

The broadcaster then uses the stream parameters (which are based on the auth webhook response) to [generate the job capabilities](https://github.com/livepeer/go-livepeer/blob/3610c2e763dc0907d278484518fb085a8bbfa2c2/server/mediaserver.go#L491). The [JobCapabilities()](https://github.com/livepeer/go-livepeer/blob/b933f4bf977a529814dd3bf89dfbc89efd524380/core/capabilities.go#L82) function will need to be updated to set the scene classification or object detection capability based on the detector profiles configured which could look like this:

```golang=
func JobCapabilities(params *StreamParameters) (*Capabilities, error) {
    ...
    
    for _, p := range params.Detection.DetectorProfiles {
        switch p.Type() {
            case SceneClassificationProfile:
                caps[Capability_SceneClassification] = true
        }
    }
}
```

### Broadcaster Segment Submission

A broadcaster should request detection for 1 / `freq` segments. Since the detection profiles will be stored in a `StreamParams` struct that is cached for the entire stream, we will need to only send the detection profiles to the orchestrator when detection is requested. We can do this by checking if detection should be requested in [genSegCreds()](https://github.com/livepeer/go-livepeer/blob/3610c2e763dc0907d278484518fb085a8bbfa2c2/server/mediaserver.go#L296), which is used to generate the metadata (video profiles, detection profiles, capabilities, etc.) to send with a segment, and only including detector profiles if needed which could look like this:

```golang=
func genSegCreds(sess *BroadcastSession, seg *stream.HLSSegment) (string, error) {
    ...
    
    detectorProfiles := []ffmpeg.DetectorProfile{}
    if seg.SeqNo % sess.Params.Detection.Freq == 0 {
        detectorProfiles = sess.Params.Detection.DetectorProfiles
    }
    
    md := &core.SegTranscodingMetadata{
        ...
        DetectorProfiles: detectorProfiles,
    }
}
```

We will need to include the `DetectorProfiles` field in both `core.SegTranscodingMetaData` and update the [helper](https://github.com/livepeer/go-livepeer/blob/b933f4bf977a529814dd3bf89dfbc89efd524380/core/streamdata.go#L64) to convert `core.SegTranscodingMetaData` to `net.SegData` which will be passed to the orchestrator.

### Orchestrator Processing Detection Requests

The orchestrator will need to convert the `net.SegData` back into `core.SegTranscodingData` in order to access the detector profiles (if there are any).

The Go struct that [stores result data](https://github.com/livepeer/go-livepeer/blob/b933f4bf977a529814dd3bf89dfbc89efd524380/core/orchestrator.go#L387) will need to be updated to mirror the `TranscodeData` protobuf message. 

### Transcoder Processing Detection Requests

We'll need to update [Transcoder interface implementations](https://github.com/livepeer/go-livepeer/blob/master/core/transcoder.go) to read detector profiles from `core.SegTranscodingMetaData` and properly pass them into the LPMS `Transcode()` API calls. The result of the LPMS calls will then need to be stored in the `TranscodeData` - these updates can be made in the [resToTranscodeData()](https://github.com/livepeer/go-livepeer/blob/b933f4bf977a529814dd3bf89dfbc89efd524380/core/transcoder.go#L159) helper.

### Broadcaster Detection Webhook

The broadcaster can submit a request to an external detection webhook with detection results whenever they are received. The webhook request structure will be:

```json=
{
    "manifestID": "foo",
    "seqNo": 9,
    "sceneClassification": {
         "classProbs": {
            1: {
                "probNum": 99,
                "probDenom": 100
            },
            2: {
                "probNum": 97,
                "probDenom": 100
            }
         }
     }
}
```

The webhook should return a 200 response if the request was successful.

The [transcodeSegment()](https://github.com/livepeer/go-livepeer/blob/b933f4bf977a529814dd3bf89dfbc89efd524380/server/broadcast.go#L458) function is responsible for downloading renditions from the orchestrator and inserting them into the playlist so we could also submit a request to the webhook here as well which could look like this:

```golang=

func transcodeSegment(...) {
    ...
    
    if len(res.Detections) > 0 {
        go callDetectionWebhook(cxn.mid, seg.SeqNo, res.Detections)
    }
    
        go dlFunc(v.Url, v.Pixels, i)
    }
}

func callDetectionWebhook(manifestID string, seqNo int64, detections []ffmpeg.Detection) {
    // Send a HTTP request to the content detection webhook
}
```

The webhook implementation would be responsible for interpreting the detection results i.e. performing some action based on the probabilities of the classes requested.

- Unit tests!
- Make sure that transcoding works without detection enabled
- Make sure that transcoding works when detection is enabled
- Make sure that detection is requested only at the specified frequency
- Make sure that detection uses the configuration specified by the stream auth webhook
- Make sure that the detection webhook is called whenever a detection result is received

## Known Unknowns

- The latency for segments that are transcoded when detection is run

jailuthra:

Enable content detection (experimental) in go-livepeer.

- Update to latest https://github.com/livepeer/FFmpeg/pull/1 and https://github.com/livepeer/lpms/pull/228
- https://github.com/livepeer/go-livepeer/pull/1911/commits/1c1070649370bbeca4b6a7ffc50b3916b123b49b - Add debug logs for content detection
- https://github.com/livepeer/go-livepeer/pull/1911/commits/f08ec26a19d6d40f88ec0431e6016f4249ae3773 - Update the `simple_auth_server` tool to configure detection on the B via the auth-webhook
- API changes from https://github.com/livepeer/go-livepeer/pull/1922
- https://github.com/livepeer/go-livepeer/pull/1911/commits/e953cafd13eb340ba15faf50638588d78fe0ac1f, https://github.com/livepeer/go-livepeer/pull/1911/commits/fb688598e946da17654cec820627badf7917636d, https://github.com/livepeer/go-livepeer/pull/1911/commits/403b456447d444d824a149a12a997c739065d23d - Update the transcoder, `livepeer` and `livepeer_bench` tools to load the "AdultSoccer" scene classification model.
- CI changes from https://github.com/livepeer/go-livepeer/pull/1926

End-to-end testing both locally and in staging.

Fixes #1881 

**Checklist:**


jailuthra:
Cleaned up the commit history, and updated the description with the changes.

Keeping the target branch as `jai/scene-classification` to simplify review process, but we can merge this directly into `master` (as the changes in jai/scene-classification are already reviewed in https://github.com/livepeer/go-livepeer/pull/1895 - https://github.com/livepeer/go-livepeer/pull/1901)

Build is currently failing due to some unrelated certificate issue with upstream x264 - will look into it.

CHANGELOG_PENDING.md
```diff
@@ -6,6 +6,8 @@
 
 #### General
 
+- \#1911 [Experimental] Enable content detection (@jailuthra,@yondonfu)
```
yondonfu: Let's add a note that only adult/soccer is supported at the moment since we are only loading the model for the adult/soccer profile on node startup.

yondonfu (approved):
Just one comment about the pending changelog entry. Feel free to address and then merge after looking into the Windows build.

LGTM 🚢

jailuthra:
>  Feel free to address and then merge after looking into the Windows build.

The windows build has been failing for some other issue (GPG keyring) even for `master`, since https://github.com/livepeer/go-livepeer/runs/2887772138
Not taking any action for it on this PR and moving ahead for the merge.

>> Build is currently failing due to some unrelated certificate issue with upstream x264 - will look into it.

This was a networking blip in the Linux builds, and it got resolved after re-triggering the build.

PR #1912, opened 2021-06-10
Should not save (when recording) segments with zero video frames

### LINKED ISSUE #1888: Should not save (when recording) segments with zero video frames
## Problem

Mist sometimes emit segments with zero video frames. We have special check for such frames in lpms, so they not transcoded. But if recording is enabled - these segments saved in the source stream. 
And these segments cause problems when playing such streams - our test player and bitmovin player stop at such a segment.

## Proposed solution

We do have function in lpms that check for such segments 
https://github.com/livepeer/lpms/blob/master/ffmpeg/extras.h#L5
but we don't expose this function publicly.
We should expose this function and use to check source segments before saving.

## Other potential solutions

We could delay saving source segment until transcoder response and check for specific error.
That solution have potential problem that if transcoder is unavailable then broadcaster can't tell if segment is good or bad.

darkdarkdragon:
Do not save segments with zero video frame to record store.

Check segment before saving using `lpmg.HasZeroVideoFrameBytes` function.

Unit test

Fixes #1888 

**Checklist:**


server/broadcast.go
```diff
@@ -386,6 +386,19 @@ func processSegment(cxn *rtmpConnection, seg *stream.HLSSegment) ([]string, erro
 	if ros != nil {
 		go func() {
 			now := time.Now()
+			hasZeroVideoFrame, err := ffmpeg.HasZeroVideoFrameBytes(seg.Data)
+			if err != nil {
+				glog.Warningf("Error checking for zero video frame nonce=%d manifestID=%s name=%s bytes=%d took=%s err=%v",
+					nonce, mid, name, len(seg.Data), time.Since(now), err)
+			} else {
+				glog.Infof("Checking for zero video frame nonce=%d manifestID=%s name=%s bytes=%d took=%s returned res=%v",
```
iameli: This'll log every segment, yeah? Seems like it could be a `glog.Debugf` at least.

darkdarkdragon: Good point. Moved it around so it will show only for 0-frame segments.

server/broadcast.go
```diff
@@ -386,6 +386,19 @@ func processSegment(cxn *rtmpConnection, seg *stream.HLSSegment) ([]string, erro
 	if ros != nil {
 		go func() {
 			now := time.Now()
+			hasZeroVideoFrame, err := ffmpeg.HasZeroVideoFrameBytes(seg.Data)
```
yondonfu, line 389: I see that this function was added in https://github.com/livepeer/lpms/pull/237 and it returns an integer. Fine for now since that PR was already merged, but I think returning a boolean would make this function more straightforward to use and suggest making that change separately.

yondonfu (approved):
LGTM

PR #1915, opened 2021-06-11
cmd,eth: use gas price monitor for all transactions instead of 'eth_gasPrice' from the ethereum JSON-RPC provider

### LINKED ISSUE #1906: Failed gas estimation with revert for txs when another tx affecting the same state is pending
# Background

A few orchestrators have reported gas estimation failing for reward txs with the confusing error message of:

```
failed to estimate gas needed: execution reverted: caller has already called reward for the current round
```

This error message is confusing because in these cases reward txs for these orchestrators had not yet been mined in the current round. This error message was observed when the node tried to automatically submit reward txs and when `livepeer_cli` was used to manually submit a reward tx. 

In one case, the operator was able to solve this problem by restarting the orchestrator with a different ETH provider (i.e. switched from a cluster of geth nodes to Infura). After the reward tx was mined, the operator observed a sequence of failed txs:

1. [Successful reward tx](https://etherscan.io/tx/0xdd9ea3c395af5cc174f2cbf124a66ef7bb00e742f0bdfe5cc638ad2b0193710c)
2. [Reverted reward tx](https://etherscan.io/tx/0xcfbc7664378b5726402944f49283dedabbd6798bbf5ec0fe7d782cbc93e0d83f)

The orchestrator had also previously attempted to submit transcoder txs (omitted from the above list, but can be found on the account's Etherscan page), but the node timed out waiting for each of the txs to be mined. The failed tx sequence and the node timing out waiting for some of the txs in the sequence leads me to the believe that the problem was that a bunch of txs were stuck pending which resulted in a few additional problems one of them being failed gas estimation with a revert. I'll focus on this problem in this issue.

The reward tx submitted by the orchestrator while another reward tx was pending resulted in a failed gas estimation with a revert because the node currently uses [EstimateGas](https://github.com/ethereum/go-ethereum/blob/0e9c7d564d6000a372cd3df914ec6c969aea29f2/ethclient/ethclient.go#L505) which calls the `eth_estimateGas` RPC method under the hood without a specified block number. When `eth_estimateGas` is called without a specific block number, it will by default by run against the current ["pending" block](https://github.com/ethereum/go-ethereum/blob/5cff9754d795971451f2f4e8a2cc0c6f51ce9802/internal/ethapi/api.go#L1040). I believe this means that a pending reward tx from the orchestrator would be applied to state such that when gas estimation is invoked for the new reward tx it will execute against state that was already updated by the pending reward tx. As a result, the new reward tx would revert during gas estimation since the updated state reflects that reward was already called in the current round.

This problem will typically arise if an initial reward tx (or any tx for that matter) is submitted with a low gas price such it gets stuck in the pending state. Previously, this would only happen if the required gas price for a tx to be mined in a timely manner spiked up right after a tx was submitted. However, lately, the `eth_gasPrice` RPC call can return extremely low gas prices (lower than what is actually required) depending on the ETH node/provider as a result of an uptick of 0/1 wei gas price MEV txs being mined. geth updated its gas price oracle implementation to stabilize values returned by `eth_gasPrice` in [v1.10.3](https://github.com/ethereum/go-ethereum/releases/tag/v1.10.3), but orchestrator operators may be slow to upgrade and ETH providers running geth might still be using an older version - for example, [Infura is still on v.1.10.2](https://forkmon.ethdevops.io/). While #1860 improved the stability of the gas price monitor in the livepeer node, it is still possible for an [extremely low gas price to be attached to a tx at submission time](https://github.com/ethereum/go-ethereum/blob/0e9c7d564d6000a372cd3df914ec6c969aea29f2/accounts/abi/bind/base.go#L229).

# Possible Solutions

#1841 could help by giving a livepeer node access to a more robust gas price service.

I think support for tx retries with gas price bumping for all txs (we currently have a more limited version implemented just for reward txs) would also help here assuming that the implementation includes:

- A max # of retries
- A time interval between retries
- A strategy to bump the previous gas price by X% OR use the current fetched gas price if it is higher than the result of the former

The time interval between retries would give the livepeer node an opportunity to get a fresh gas price after a period of time - even if the `eth_gasPrice` RPC call returns extremely low gas prices sometimes it is less likely to consistently return extremely low gas prices. The gas price setting strategy ensures that if the fresh gas price is more reasonable than the previous extremely low gas price then it will be used.

Given that recent versions of geth help with this problem when still using the `eth_gasPrice` RPC call to fetch gas prices (so its a matter of orchestrator operators upgrading their geth or ETH providers upgrading their geth), I'm a bit more interested in exploring the tx retries solution first as that will also be generally useful outside of addressing this problem.

kyriediculous:
Improve transaction management with better UX 

- [x] use gas price monitor for all transactions instead of 'eth_gasPrice' from the ethereum JSON-RPC provider

- Use `GasPriceMonitor.GasPrice()` instead of `ethClient.SuggestGasPrice()` in `Backend.SuggestGasPrice()`. Since `Backend` overwrites the `eth.Transactor` interface all transactions, also contract calls through the contract bindings, will use the gas price stored by the `GasPriceMonitor`. 
- Use the `GasPriceMonitor` regardless of the node being in O or B mode

- Unit tests

Fixes #1906 

**Checklist:**

### DIFF
--- a/cmd/devtool/devtool.go
+++ b/cmd/devtool/devtool.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"bytes"
+	"context"
 	"flag"
 	"fmt"
 	"io"
@@ -159,7 +160,16 @@ func ethSetup(ethAcctAddr, keystoreDir string, isBroadcaster bool) {
 	}
 	glog.Infof("Using controller address %s", ethController)
 
-	client, err := eth.NewClient(ethcommon.HexToAddress(ethAcctAddr), keystoreDir, passphrase, backend,
+	gpm := eth.NewGasPriceMonitor(backend, 5*time.Second, big.NewInt(0))
+	// Start gas price monitor
+	_, err = gpm.Start(context.Background())
+	if err != nil {
+		glog.Errorf("error starting gas price monitor: %v", err)
+		return
+	}
+	defer gpm.Stop()
+
+	client, err := eth.NewClient(ethcommon.HexToAddress(ethAcctAddr), keystoreDir, passphrase, backend, gpm,
 		ethcommon.HexToAddress(ethController), ethTxTimeout, nil)
 	if err != nil {
 		glog.Errorf("Failed to create client: %v", err)
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -380,7 +380,16 @@ func main() {
 			bigMaxGasPrice = big.NewInt(int64(*maxGasPrice))
 		}
 
-		client, err := eth.NewClient(ethcommon.HexToAddress(*ethAcctAddr), keystoreDir, *ethPassword, backend, ethcommon.HexToAddress(*ethController), EthTxTimeout, bigMaxGasPrice)
+		gpm := eth.NewGasPriceMonitor(backend, blockPollingTime, minGasPrice)
+		// Start gas price monitor
+		_, err = gpm.Start(ctx)
+		if err != nil {
+			glog.Errorf("Error starting gas price monitor: %v", err)
+			return
+		}
+		defer gpm.Stop()
+
+		client, err := eth.NewClient(ethcommon.HexToAddress(*ethAcctAddr), keystoreDir, *ethPassword, backend, gpm, ethcommon.HexToAddress(*ethController), EthTxTimeout, bigMaxGasPrice)
 		if err != nil {
 			glog.Errorf("Failed to create Livepeer Ethereum client: %v", err)
 			return
@@ -537,14 +546,6 @@ func main() {
 
 			sigVerifier := &pm.DefaultSigVerifier{}
 			validator := pm.NewValidator(sigVerifier, timeWatcher)
-			gpm := eth.NewGasPriceMonitor(backend, blockPollingTime, minGasPrice)
-			// Start gas price monitor
-			_, err := gpm.Start(ctx)
-			if err != nil {
-				glog.Errorf("error starting gas price monitor: %v", err)
-				return
-			}
-			defer gpm.Stop()
 
 			var sm pm.SenderMonitor
 			if *redeemerAddr != "" {
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -7,6 +7,7 @@
 #### General
 
 - \#1911 [Experimental] Enable scene classification for Adult/Soccer (@jailuthra, @yondonfu)
+- \#1915 Use gas price monitor for gas price suggestions for all Ethereum transactions (@kyriediculous)
 
 #### Broadcaster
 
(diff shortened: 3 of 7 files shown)


cmd/livepeer/livepeer.go
```diff
@@ -390,7 +390,16 @@ func main() {
 			bigMaxGasPrice = big.NewInt(int64(*maxGasPrice))
 		}
 
-		client, err := eth.NewClient(ethcommon.HexToAddress(*ethAcctAddr), keystoreDir, *ethPassword, backend, ethcommon.HexToAddress(*ethController), EthTxTimeout, bigMaxGasPrice)
+		gpm := eth.NewGasPriceMonitor(backend, blockPollingTime, minGasPrice)
+		// Start gas price monitor
+		_, err = gpm.Start(ctx)
+		if err != nil {
+			glog.Errorf("error starting gas price monitor: %v", err)
```
```suggestion
			glog.Errorf("Error starting gas price monitor: %v", err)
```

For consistency with other error messages.

cmd/devtool/devtool.go
```diff
@@ -159,7 +160,16 @@ func ethSetup(ethAcctAddr, keystoreDir string, isBroadcaster bool) {
 	}
 	glog.Infof("Using controller address %s", ethController)
 
-	client, err := eth.NewClient(ethcommon.HexToAddress(ethAcctAddr), keystoreDir, passphrase, backend,
+	gpm := eth.NewGasPriceMonitor(backend, 5*time.Second, big.NewInt(0))
+	// Start gas price monitor
+	_, err = gpm.Start(context.TODO())
```
yondonfu: Is context.TODO() here actually meant to be a placeholder or would it be fine to just use context.Background()?

I believe both are technically equivalent, but typically TODO() is used to serve as a placeholder with the expectation that it will be replaced while Background() is used when an empty context is explicitly desired.

yondonfu (changes requested):
Let's add a unit test checking that SuggestGasPrice on backend returns the value of GasPrice() on GasPriceMonitor?

cmd/devtool/devtool.go
kyriediculous: fixed in 0b64e17

cmd/livepeer/livepeer.go
kyriediculous: fixed in 0b64e17

yondonfu (approved):
LGTM after squashing and fixing merge conflict.

Looks like GH actions is stuck for some reason...Let's see if it runs properly after a rebase against master...

kyriediculous:
The windows builds seem to be failing consistently for all PRs , not sure what's wrong there.

yondonfu:
> The windows builds seem to be failing consistently for all PRs , not sure what's wrong there.

Yeah it has been failing consistently for reasons unrelated to recent code changes. We'll need to fix that separately. Let's go ahead with the merge for now.

PR #1922, opened 2021-06-17
server: Simplify detection webhook shapes

### LINKED ISSUE #1917: finalize detection webhook JSON shapes
In an effort to maximize the usability of the scene classification API, as well as keep things as consistent as possible between go-livepeer and livepeer-com, the following is proposed as the shape for the scene detection webhook workflow, based on discussion and consensus with @yondonfu @ericxtang @victorges. Some basic objectives:

* Replace opaque integer IDs with human-readable strings
* Replace fraction types with floating-point numbers; exact precision isn't as necessary when we're talking about being 96.1391296% certain that there is soccer on the screen
* Keep the API shape as consistent as possible; when possible the same scene classification concepts should be referenced in the same way

## Incoming auth webhook response from API server:
```json5
{
    // ... other fields here, "profiles" and "manifestID" and whatnot ...

    detection: {
        frequency: 1,
        sampleRate: 100,
        sceneClassification: [
            {name: "soccer"},
            {name: "adult"}
        ],
        
        // someday, we'll have other detection-related fields here, like:
        objectDetection: [
            {name: "soccer ball"}
        ]
    }
}
```

## Outgoing classification message to `-detectionWebhookURL`
```json5
{
    manifestID: "ABC123",
    seqNo: 100,
    sceneClassification: [
        {name: "soccer", probability: 0.5},
        {name: "adult", probability: 0.01}
    ]
    
    // and some
    objectDetection: [
        {name: "soccer ball", coords: [1,2,3,4]}
    ]
}
```

jailuthra:

Simplify detection webhook shapes - for both stream-wide configuration via auth webhook and segment-wise results via detection webhook.

*see commit history*

Updated unit test for authwebhook -> detector configuration.

Manual testing for detection results webhook:
```json
{"manifestID":"test","seqNo":0,"sceneClassification":[{"name":"soccer","probability":0.557692289352417},{"name":"adult","probability":0}]}
{"manifestID":"test","seqNo":4,"sceneClassification":[{"name":"soccer","probability":1},{"name":"adult","probability":0}]}
{"manifestID":"test","seqNo":8,"sceneClassification":[{"name":"soccer","probability":1},{"name":"adult","probability":0}]}
{"manifestID":"test","seqNo":12,"sceneClassification":[{"name":"soccer","probability":1},{"name":"adult","probability":0}]}
{"manifestID":"test","seqNo":16,"sceneClassification":[{"name":"soccer","probability":0.9193333387374878},{"name":"adult","probability":0}]}
```

Fixes #1917 

**Depends upon**
Some changes in LPMS https://github.com/livepeer/lpms/commit/b196bdaaea2a670816cfdbde4f8e35d6aeb51b3a which are currently a part of a working branch (`jai/adddnn`).
But we can ignore the dependency for now, as the target for this PR is also a working branch (`jai/e2e-detection`)

**Checklist:**


server/broadcast.go
```diff
@@ -550,18 +550,33 @@ func transcodeSegment(cxn *rtmpConnection, seg *stream.HLSSegment, name string,
 	// [EXPERIMENTAL] send content detection results to callback webhook
 	if DetectionWebhookURL != "" && len(res.Detections) > 0 {
 		glog.V(common.DEBUG).Infof("Got detection result %v", res.Detections)
-		go func(mid core.ManifestID, seqNo uint64, detections []*net.DetectData) {
+		go func(mid core.ManifestID, config core.DetectionConfig, seqNo uint64, detections []*net.DetectData) {
+			type SceneClassificationResult struct {
+				ClassName   string  `json:"name"`
```
```suggestion
				Name   string  `json:"name"`
```

^ To match the JSON field name?

server/broadcast.go
```diff
@@ -550,18 +550,33 @@ func transcodeSegment(cxn *rtmpConnection, seg *stream.HLSSegment, name string,
...
+				Probability float32 `json:"probability"`
```
```suggestion
				Probability float64 `json:"probability"`
```

^ If we're using a double for the protobuf field (which maps to a Go float64 type) might as well use float64 in Go?

server/mediaserver.go
```diff
@@ -121,14 +121,11 @@ type authWebhookResponse struct {
...
-			SampleRate uint `json:"sampleRate"`
-			Classes    []struct {
-				ID   int    `json:"id"`
-				Name string `json:"name"`
-			} `json:"classes"`
-		} `json:"sceneClassificationProfile"`
+		Freq                uint `json:"freq"`
+		SampleRate          uint `json:"sampleRate"`
+		SceneClassification []struct {
+			ClassName string `json:"name"`
```
yondonfu:
```suggestion
			Name string `json:"name"`
```

^ To match the JSON field?

server/mediaserver.go
```diff
@@ -277,7 +274,11 @@ func createRTMPStreamIDHandler(s *LivepeerServer) func(url *url.URL) (strmID str
 
 			// set Detection profile if provided
 			if resp.Detection.Freq != 0 {
-				detectionConfig = jsonDetectionToDetectionConfig(resp)
+				detectionConfig, err = jsonDetectionToDetectionConfig(resp)
+				if err != nil {
+					glog.Errorf("Failed to parse detector profile from JSON for streamID url=%s err=%v", url.String(), err)
```
yondonfu:
```suggestion
					glog.Errorf("Failed to parse detection config from JSON for streamID url=%s err=%v", url.String(), err)
```

core/streamdata.go
```diff
@@ -23,8 +23,9 @@ const (
 )
 
 type DetectionConfig struct {
-	Freq     uint
-	Profiles []ffmpeg.DetectorProfile
+	Freq               uint
+	SelectedClassNames []string
```
yondonfu, line 27: Does this field need to be at the top level of this struct if an individual DetectorProfile that is implemented by a SceneClassificationProfile stores the classes there already in the `Classes` field?

server/mediaserver.go
```diff
@@ -407,25 +408,36 @@ func jsonProfileToVideoProfile(resp *authWebhookResponse) ([]ffmpeg.VideoProfile
...
+			if !skipProfile {
+				c.SampleRate = resp.Detection.SampleRate
+				detection.Profiles = append(detection.Profiles, &c)
+			}
+		} else {
+			return detection, errors.New("No detector found for class name: " + class.ClassName)
+		}
 	}
-	return detection
+	glog.V(common.DEBUG).Infof("Parsed detection config from webhook %v",
```
yondonfu: Prefer to either:

- Remove this debug statement since dumping the struct to the logs just adds noise
- Log the relevant struct fields in a more structured way so it is more readable

server/broadcast.go
```diff
@@ -550,18 +550,33 @@ func transcodeSegment(cxn *rtmpConnection, seg *stream.HLSSegment, name string,
...
 			}
 			req := DetectionWebhookRequest{ManifestID: mid, SeqNo: seqNo}
 			for _, detection := range detections {
 				switch x := detection.Value.(type) {
 				case *net.DetectData_SceneClassification:
-					data := x.SceneClassification
-					req.SceneClassification = append(req.SceneClassification, *data)
+					probs := x.SceneClassification.ClassProbs
+					// match returned result (classID map) with one of the selected class names
+					for _, className := range config.SelectedClassNames {
```
yondonfu: If we no longer have SelectedClassNames we could still do:

```golang
for i, detection := range detections {
    ...
    probs := x.SceneClassification.ClassProbs
    prof, ok := config.Profiles[i].(SceneClassificationProfile)
    if !ok {
        ...
    }
    
    for _, class := range prof.Classes {
         classID, ok := ffmpeg.DetectorClassIDLookup[class.Name]
         if !ok {
             continue
         }

         prob, ok := probs[uint32(classID)]
         if !ok {
             continue
         }

         req.SceneClassification = append(req.SceneClassification, SceneClassificationResult{Name: class.Name, Probabiliy: prob})
     }
}
```

server/mediaserver.go
```diff
@@ -407,25 +408,36 @@ func jsonProfileToVideoProfile(resp *authWebhookResponse) ([]ffmpeg.VideoProfile
...
+			// Check if a previous selected class had the same model, if yes we can skip the profile
+			skipProfile := false
+			for _, selectedClass := range detection.SelectedClassNames {
+				sc, ok := ffmpeg.SceneClassificationProfileLookup[selectedClass]
+				if ok && sc.ModelPath == c.ModelPath {
+					skipProfile = true
+					break
+				}
+			}
+			detection.SelectedClassNames = append(detection.SelectedClassNames, class.ClassName)
```
yondonfu: If we didn't have SelectedClassNames we could still define a local map to track already parsed profiles to determine whether we should skip a profile or not.

Ex.

```golang
modelPaths := make(map[string]bool)
for _, class := range resp.Detection.SceneClassification {
    c, ok := ffmpeg.SceneClassificationProfileLookup[class.Name]
    if !ok {
        ...
    }

    if _, ok := modelPaths[c.ModelPath]; ok {
        // Skip this profile because we already have a profile with a model that covers this class
        continue
    }

    modelPaths[c.ModelPath] = true
    
    c.SampleRate = resp.Detection.SampleRate
    detection.Profiles = append(detection.Profiles, &c)
}
```

I think the nice thing about using a local map rather than SelectedClassNames is that we can avoid carrying additional state in the DetectionConfig struct and we keep everything local to this function since it is not required elsewhere.

yondonfu (changes requested):

core/streamdata.go
jailuthra: The main motive to keep this here is for API users who (let's say) only ask for "soccer" detection and not "adult" detection. Internal details like those two classes being part of the same trained model (which could ofc change in future) should be opaque to the end-user, by filtering the detection results for classes the user didn't ask for.

An alternative is to keep only the selected classes inside `ffmpeg.DetectorProfile` before sending it to the O - but ultimately the model would return results for all trained classes and we'd need to keep the filtering logic on the O/LPMS side if not here. Right now O/LPMS don't need to care about user-selected classes - which is cleaner.

server/broadcast.go
jailuthra: Dropping `SelectedClassNames` will miss out on a usecase mentioned in https://github.com/livepeer/go-livepeer/pull/1922#discussion_r654218562 but happy to change this if we decide to drop it.

server/mediaserver.go
jailuthra: This is much cleaner even if we decide to stick with `SelectedClassNames` - will fix.

core/streamdata.go
yondonfu: Got it!

yondonfu (approved):
LGTM after squashing

Issue #1822, opened 2021-03-31
Log from reward transaction always shows 0 LPTU minted upon success.


kyriediculous:
**Describe the bug**

The log statement that logs how many LPT rewards were minted after a successful reward call always indicates 0 LPTU were minted

```
I0330 18:08:15.313752    7227 rewardservice.go:117] Called reward for round 2108 - 0 LPTU rewards minted
```

The log statement relies on the earnings pool to fetch rewards for a round, but this is no longer in use since [LIP-36](https://medium.com/livepeer-blog/earnings-claiming-gas-cost-reductions-lip-36-and-lip-52-recap-f136416fdb1) 

We could add an event listener in the `eth` package to watch for `Reward` events. 

The `rewardService` could then use this to log the amount of rewards minted each round.

PR #1923, opened 2021-06-17
eth,misc: Use transaction manager with confirmation and optional replacement queue

### LINKED ISSUE #1906: Failed gas estimation with revert for txs when another tx affecting the same state is pending

kyriediculous:
Use a `TransactionManager` to handle transactions. This service will forward transactions to the ethereum JSON-RPC, when a transaction is broadcasted to the ethereum network it will be added to the queue. Lowest non-confirmed transactions are popped off the queue to wait until they are mined and retrieve their transaction receipt. 

Once the transaction receipt is found it will be sent over a subscription of transaction receipts that services that want feedback on tx confirmation can subscribe to. This subscription service now powers `LivepeerEthClient.CheckTx()` under the hood. 

- 75eed97c84e9d40194403996fc75443f7d375d9e  eth: transaction manager with confirmation queue and optional replacement transactions

-  60d33a0c763505098488050999b220f51ed740a3  pm: handle ticket revert error

- a7b7c5e46e7d406a0f6330f53751a53e20e217ac  eth: remove replacement transactions from reward service

- d4886f688b6edc8f25243d846872a8821ca9b16f  eth,server: move max gas price logic to gas price monitor

- b12a24dc6888ff94d6a68fe3f6628cc828a1e015 eth: make abiMap and txLog helpers in the eth package rather than Backend methods

- 1b89b5c37b1a57da231df387bb8f3d163ed412a0  cmd: replace transaction flag

- efa84b466cebc138501eb8dcb380abe7fbbfd539  cmd,eth: use struct for LivepeerEthClient constructor parameters that aren't other services

- Unit tests

- [ ] TODO add `TransactionManager` unit tests

Fixes #1906

**Checklist:**

### DIFF
--- a/cmd/devtool/devtool.go
+++ b/cmd/devtool/devtool.go
@@ -161,6 +161,7 @@ func ethSetup(ethAcctAddr, keystoreDir string, isBroadcaster bool) {
 	glog.Infof("Using controller address %s", ethController)
 
 	gpm := eth.NewGasPriceMonitor(backend, 5*time.Second, big.NewInt(0))
+
 	// Start gas price monitor
 	_, err = gpm.Start(context.Background())
 	if err != nil {
@@ -169,8 +170,17 @@ func ethSetup(ethAcctAddr, keystoreDir string, isBroadcaster bool) {
 	}
 	defer gpm.Stop()
 
-	client, err := eth.NewClient(ethcommon.HexToAddress(ethAcctAddr), keystoreDir, passphrase, backend, gpm,
-		ethcommon.HexToAddress(ethController), ethTxTimeout, nil)
+	clientCfg := eth.LivepeerEthClientConfig{
+		AccountAddr:         ethcommon.HexToAddress(ethAcctAddr),
+		KeystoreDir:         keystoreDir,
+		Password:            passphrase,
+		ControllerAddr:      ethcommon.HexToAddress(ethController),
+		TxTimeout:           ethTxTimeout,
+		MaxGasPrice:         nil,
+		ReplaceTransactions: false,
+	}
+
+	client, err := eth.NewClient(backend, gpm, clientCfg)
 	if err != nil {
 		glog.Errorf("Failed to create client: %v", err)
 		return
--- a/cmd/devtool/devtool.go
+++ b/cmd/devtool/devtool.go
@@ -17,6 +17,7 @@ import (
 	"github.com/ethereum/go-ethereum/accounts/keystore"
 	ethcommon "github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/console"
+	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/ethclient"
 	"github.com/ethereum/go-ethereum/rpc"
 	"github.com/livepeer/go-livepeer/eth"
@@ -160,7 +161,7 @@ func ethSetup(ethAcctAddr, keystoreDir string, isBroadcaster bool) {
 	}
 	glog.Infof("Using controller address %s", ethController)
 
-	gpm := eth.NewGasPriceMonitor(backend, 5*time.Second, big.NewInt(0))
+	gpm := eth.NewGasPriceMonitor(backend, 5*time.Second, big.NewInt(0), nil)
 
 	// Start gas price monitor
 	_, err = gpm.Start(context.Background())
@@ -170,19 +171,40 @@ func ethSetup(ethAcctAddr, keystoreDir string, isBroadcaster bool) {
 	}
 	defer gpm.Stop()
 
-	clientCfg := eth.LivepeerEthClientConfig{
-		AccountAddr:         ethcommon.HexToAddress(ethAcctAddr),
-		KeystoreDir:         keystoreDir,
-		Password:            passphrase,
-		ControllerAddr:      ethcommon.HexToAddress(ethController),
-		TxTimeout:           ethTxTimeout,
-		MaxGasPrice:         nil,
-		ReplaceTransactions: false,
+	chainID, err := backend.ChainID(context.Background())
+	if err != nil {
+		glog.Errorf("Failed to get chain ID from remote ethereum node: %v", err)
+		return
 	}
 
-	client, err := eth.NewClient(backend, gpm, clientCfg)
+	signer := types.NewEIP155Signer(chainID)
+
+	am, err := eth.NewAccountManager(ethcommon.HexToAddress(ethAcctAddr), keystoreDir, signer)
 	if err != nil {
-		glog.Errorf("Failed to create client: %v", err)
+		glog.Errorf("Error creating Ethereum account manager: %v", err)
+		return
+	}
+
+	if err := am.Unlock(passphrase); err != nil {
+		glog.Errorf("Error unlocking Ethereum account: %v", err)
+		return
+	}
+
+	tm := eth.NewTransactionManager(backend, gpm, am, 5*time.Minute, 0)
+	go tm.Start()
+	defer tm.Stop()
+
+	ethCfg := eth.LivepeerEthClientConfig{
+		AccountManager:     am,
+		ControllerAddr:     ethcommon.HexToAddress(ethController),
+		EthClient:          backend,
+		GasPriceMonitor:    gpm,
+		TransactionManager: tm,
+	}
+
+	client, err := eth.NewClient(ethCfg)
+	if err != nil {
+		glog.Errorf("Failed to create Livepeer Ethereum client: %v", err)
 		return
 	}
 
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -393,7 +393,17 @@ func main() {
 		}
 		defer gpm.Stop()
 
-		client, err := eth.NewClient(ethcommon.HexToAddress(*ethAcctAddr), keystoreDir, *ethPassword, backend, gpm, ethcommon.HexToAddress(*ethController), EthTxTimeout, bigMaxGasPrice)
+		ethCfg := eth.LivepeerEthClientConfig{
+			AccountAddr:         ethcommon.HexToAddress(*ethAcctAddr),
+			KeystoreDir:         keystoreDir,
+			Password:            *ethPassword,
+			ControllerAddr:      ethcommon.HexToAddress(*ethController),
+			TxTimeout:           EthTxTimeout,
+			MaxGasPrice:         bigMaxGasPrice,
+			ReplaceTransactions: *replaceTx,
+		}
+
+		client, err := eth.NewClient(backend, gpm, ethCfg)
 		if err != nil {
 			glog.Errorf("Failed to create Livepeer Ethereum client: %v", err)
 			return
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -22,6 +22,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/params"
 	"github.com/ethereum/go-ethereum/rpc"
 	"github.com/livepeer/go-livepeer/build"
@@ -45,8 +46,7 @@ import (
 )
 
 var (
-	ErrKeygen    = errors.New("ErrKeygen")
-	EthTxTimeout = 600 * time.Second
+	ErrKeygen = errors.New("ErrKeygen")
 
 	// The timeout for ETH RPC calls
 	ethRPCTimeout = 20 * time.Second
@@ -113,6 +113,8 @@ func main() {
 	ethKeystorePath := flag.String("ethKeystorePath", "", "Path for the Eth Key")
 	ethOrchAddr := flag.String("ethOrchAddr", "", "ETH address of an on-chain registered orchestrator")
 	ethUrl := flag.String("ethUrl", "", "Ethereum node JSON-RPC URL")
+	txTimeout := flag.Duration("transactionTimeout", 5*time.Minute, "Amount of time to wait for an Ethereum transaction to confirm before timing out")
+	maxTxReplacements := flag.Int("maxTransactionReplacements", 1, "Number of times to automatically replace pending Ethereum transactions")
 	gasLimit := flag.Int("gasLimit", 0, "Gas limit for ETH transactions")
 	minGasPrice := flag.Int64("minGasPrice", 0, "Minimum gas price for ETH transactions")
 	maxGasPrice := flag.Int("maxGasPrice", 0, "Maximum gas price for ETH transactions")
@@ -384,7 +386,7 @@ func main() {
 			bigMaxGasPrice = big.NewInt(int64(*maxGasPrice))
 		}
 
-		gpm := eth.NewGasPriceMonitor(backend, blockPollingTime, big.NewInt(*minGasPrice))
+		gpm := eth.NewGasPriceMonitor(backend, blockPollingTime, big.NewInt(*minGasPrice), bigMaxGasPrice)
 		// Start gas price monitor
 		_, err = gpm.Start(ctx)
 		if err != nil {
@@ -393,17 +395,32 @@ func main() {
 		}
 		defer gpm.Stop()
 
-		ethCfg := eth.LivepeerEthClientConfig{
-			AccountAddr:         ethcommon.HexToAddress(*ethAcctAddr),
-			KeystoreDir:         keystoreDir,
-			Password:            *ethPassword,
-			ControllerAddr:      ethcommon.HexToAddress(*ethController),
-			TxTimeout:           EthTxTimeout,
-			MaxGasPrice:         bigMaxGasPrice,
-			ReplaceTransactions: *replaceTx,
+		signer := types.NewEIP155Signer(chainID)
+
+		am, err := eth.NewAccountManager(ethcommon.HexToAddress(*ethAcctAddr), keystoreDir, signer)
+		if err != nil {
+			glog.Errorf("Error creating Ethereum account manager: %v", err)
+			return
 		}
 
-		client, err := eth.NewClient(backend, gpm, ethCfg)
+		if err := am.Unlock(*ethPassword); err != nil {
+			glog.Errorf("Error unlocking Ethereum account: %v", err)
+			return
+		}
+
+		tm := eth.NewTransactionManager(backend, gpm, am, *txTimeout, *maxTxReplacements)
+		go tm.Start()
+		defer tm.Stop()
+
+		ethCfg := eth.LivepeerEthClientConfig{
+			AccountManager:     am,
+			ControllerAddr:     ethcommon.HexToAddress(*ethController),
+			EthClient:          backend,
+			GasPriceMonitor:    gpm,
+			TransactionManager: tm,
+		}
+
+		client, err := eth.NewClient(ethCfg)
 		if err != nil {
 			glog.Errorf("Failed to create Livepeer Ethereum client: %v", err)
 			return
--- a/eth/backend.go
+++ b/eth/backend.go
@@ -16,7 +16,6 @@ import (
 	"github.com/ethereum/go-ethereum/params"
 	"github.com/golang/glog"
 	"github.com/livepeer/go-livepeer/eth/contracts"
-	"github.com/livepeer/go-livepeer/monitor"
 )
 
 var abis = []string{
@@ -31,6 +30,8 @@ var abis = []string{
 	contracts.PollABI,
 }
 
+var abiMap = makeABIMap()
+
 type Backend interface {
 	ethereum.ChainStateReader
 	ethereum.TransactionReader
@@ -43,36 +44,27 @@ type Backend interface {
 	ethereum.LogFilterer
 	ethereum.ChainReader
 	ChainID(ctx context.Context) (*big.Int, error)
-	MaxGasPrice() *big.Int
-	SetMaxGasPrice(gp *big.Int)
-	MinGasPrice() *big.Int
-	SetMinGasPrice(gp *big.Int)
+	GasPriceMonitor() *GasPriceMonitor
 }
 
 type backend struct {
 	*ethclient.Client
-	abiMap       map[string]*abi.ABI
 	nonceManager *NonceManager
 	signer       types.Signer
 	gpm          *GasPriceMonitor
+	tm           *TransactionManager
 
 	sync.RWMutex
-	maxGasPrice *big.Int
 }
 
-func NewBackend(client *ethclient.Client, signer types.Signer, gpm *GasPriceMonitor) (Backend, error) {
-	abiMap, err := makeABIMap()
-	if err != nil {
-		return nil, err
-	}
-
+func NewBackend(client *ethclient.Client, signer types.Signer, gpm *GasPriceMonitor, tm *TransactionManager) Backend {
 	return &backend{
 		Client:       client,
-		abiMap:       abiMap,
 		nonceManager: NewNonceManager(client),
 		signer:       signer,
 		gpm:          gpm,
-	}, nil
+		tm:           tm,
+	}
 }
 
 func (b *backend) PendingNonceAt(ctx context.Context, account common.Address) (uint64, error) {
@@ -83,7 +75,10 @@ func (b *backend) PendingNonceAt(ctx context.Context, account common.Address) (u
 }
 
 func (b *backend) SendTransaction(ctx context.Context, tx *types.Transaction) error {
-	sendErr := b.Client.SendTransaction(ctx, tx)
+	// Use the transaction manager instead of the ethereum client
+	if err := b.tm.SendTransaction(ctx, tx); err != nil {
+		return err
+	}
 
 	msg, err := tx.AsMessage(b.signer)
 	if err != nil {
@@ -91,22 +86,10 @@ func (b *backend) SendTransaction(ctx context.Context, tx *types.Transaction) er
 	}
 	sender := msg.From()
 
-	txLog, err := b.newTxLog(tx)
-	if err != nil {
-		txLog.method = "unknown"
-	}
-
-	if sendErr != nil {
-		glog.Infof("\n%vEth Transaction%v\n\nInvoking transaction: \"%v\". Inputs: \"%v\"   \nTransaction Failed: %v\n\n%v\n", strings.Repeat("*", 30), strings.Repeat("*", 30), txLog.method, txLog.inputs, sendErr, strings.Repeat("*", 75))
-		return sendErr
-	}
-
 	// update local nonce
 	b.nonceManager.Lock(sender)
 	b.nonceManager.Update(sender, tx.Nonce())
 	b.nonceManager.Unlock(sender)
-
-	glog.Infof("\n%vEth Transaction%v\n\nInvoking transaction: \"%v\". Inputs: \"%v\"  Hash: \"%v\". \n\n%v\n", strings.Repeat("*", 30), strings.Repeat("*", 30), txLog.method, txLog.inputs, tx.Hash().String(), strings.Repeat("*", 75))
 
 	return nil
 }
@@ -115,10 +98,11 @@ func (b *backend) SuggestGasPrice(ctx context.Context) (*big.Int, error) {
 
 	// Use the gas price monitor instead of the ethereum client to suggest a gas price
 	gp := b.gpm.GasPrice()
+	maxGp := b.gpm.MaxGasPrice()
 
-	if b.maxGasPrice != nil && gp.Cmp(b.maxGasPrice) > 0 {
+	if maxGp != nil && gp.Cmp(maxGp) > 0 {
 		return nil, fmt.Errorf("current gas price exceeds maximum gas price max=%v GWei current=%v GWei",
-			FromWei(b.maxGasPrice, params.GWei),
+			FromWei(maxGp, params.GWei),
 			FromWei(gp, params.GWei),
 		)
 	}
@@ -126,62 +110,13 @@ func (b *backend) SuggestGasPrice(ctx context.Context) (*big.Int, error) {
 	return gp, nil
 }
 
-func (b *backend) SetMaxGasPrice(gp *big.Int) {
-	b.Lock()
-	defer b.Unlock()
-	b.maxGasPrice = gp
-
-	if gp != nil && monitor.Enabled {
-		monitor.MaxGasPrice(gp)
-	}
-}
-
-func (b *backend) MaxGasPrice() *big.Int {
-	b.RLock()
-	defer b.RUnlock()
-	return b.maxGasPrice
-}
-
-func (b *backend) SetMinGasPrice(gp *big.Int) {
-	b.gpm.SetMinGasPrice(gp)
-}
-
-func (b *backend) MinGasPrice() *big.Int {
-	return b.gpm.MinGasPrice()
+func (b *backend) GasPriceMonitor() *GasPriceMonitor {
+	return b.gpm
 }
 
 type txLog struct {
 	method string
 	inputs string
-}
-
-func (b *backend) newTxLog(tx *types.Transaction) (txLog, error) {
-	var txParamsString string
-	data := tx.Data()
-	if len(data) < 4 {
-		return txLog{}, errors.New("no method signature")
-	}
-	methodSig := data[:4]
-	abi, ok := b.abiMap[string(methodSig)]
-	if !ok {
-		return txLog{}, errors.New("unknown ABI")
-	}
-	method, err := abi.MethodById(methodSig)
-	if err != nil {
-		return txLog{}, err
-	}
-	txParams := make(map[string]interface{})
-	if err := decodeTxParams(b.abiMap[string(methodSig)], txParams, data); err != nil {
-		return txLog{}, err
-	}
-
-	for _, arg := range method.Inputs {
-		txParamsString += fmt.Sprintf("%v: %v  ", arg.Name, txParams[arg.Name])
-	}
-	return txLog{
-		method: method.Name,
-		inputs: strings.TrimSpace(txParamsString),
-	}, nil
 }
 
 func (b *backend) CallContract(ctx context.Context, msg ethereum.CallMsg, blockNumber *big.Int) ([]byte, error) {
@@ -212,18 +147,48 @@ func (b *backend) retryRemoteCall(remoteCall func() ([]byte, error)) (out []byte
 	return out, err
 }
 
-func makeABIMap() (map[string]*abi.ABI, error) {
+func makeABIMap() map[string]*abi.ABI {
 	abiMap := make(map[string]*abi.ABI)
 
 	for _, ABI := range abis {
 		parsedAbi, err := abi.JSON(strings.NewReader(ABI))
 		if err != nil {
-			return map[string]*abi.ABI{}, err
+			glog.Errorf("Error creating ABI map err=%v", err)
+			return map[string]*abi.ABI{}
 		}
 		for _, m := range parsedAbi.Methods {
 			abiMap[string(m.ID())] = &parsedAbi
 		}
 	}
 
-	return abiMap, nil
+	return abiMap
+}
+
+func newTxLog(tx *types.Transaction) (txLog, error) {
+	var txParamsString string
+	data := tx.Data()
+	if len(data) < 4 {
+		return txLog{}, errors.New("no method signature")
+	}
+	methodSig := data[:4]
+	abi, ok := abiMap[string(methodSig)]
+	if !ok {
+		return txLog{}, errors.New("unknown ABI")
+	}
+	method, err := abi.MethodById(methodSig)
+	if err != nil {
+		return txLog{}, err
+	}
+	txParams := make(map[string]interface{})
+	if err := decodeTxParams(abiMap[string(methodSig)], txParams, data); err != nil {
+		return txLog{}, err
+	}
+
+	for _, arg := range method.Inputs {
+		txParamsString += fmt.Sprintf("%v: %v  ", arg.Name, txParams[arg.Name])
+	}
+	return txLog{
+		method: method.Name,
+		inputs: strings.TrimSpace(txParamsString),
+	}, nil
 }
--- a/eth/client.go
+++ b/eth/client.go
@@ -17,10 +17,8 @@ import (
 	"fmt"
 	"math/big"
 	"sort"
-	"strings"
 	"time"
 
-	ethereum "github.com/ethereum/go-ethereum"
 	"github.com/ethereum/go-ethereum/accounts"
 	"github.com/ethereum/go-ethereum/accounts/abi/bind"
 	ethcommon "github.com/ethereum/go-ethereum/common"
@@ -113,7 +111,6 @@ type LivepeerEthClient interface {
 	// Helpers
 	ContractAddresses() map[string]ethcommon.Address
 	CheckTx(*types.Transaction) error
-	ReplaceTransaction(*types.Transaction, string, *big.Int) (*types.Transaction, error)
 	Sign([]byte) ([]byte, error)
 	SetGasInfo(uint64) error
 }
@@ -148,7 +145,17 @@ type client struct {
 	txTimeout time.Duration
 }
 
-func NewClient(accountAddr ethcommon.Address, keystoreDir, password string, eth *ethclient.Client, gpm *GasPriceMonitor, controllerAddr ethcommon.Address, txTimeout time.Duration, maxGasPrice *big.Int) (LivepeerEthClient, error) {
+type LivepeerEthClientConfig struct {
+	AccountAddr         ethcommon.Address
+	KeystoreDir         string
+	Password            string
+	ControllerAddr      ethcommon.Address
+	TxTimeout           time.Duration
+	MaxGasPrice         *big.Int
+	ReplaceTransactions bool
+}
+
+func NewClient(eth *ethclient.Client, gpm *GasPriceMonitor, cfg LivepeerEthClientConfig) (LivepeerEthClient, error) {
 	chainID, err := eth.ChainID(context.Background())
 	if err != nil {
 		return nil, err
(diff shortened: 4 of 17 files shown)


eth/backend.go
```diff
@@ -42,32 +44,27 @@ type Backend interface {
...
+	GasPriceMonitor() *GasPriceMonitor
 }
 
 type backend struct {
 	*ethclient.Client
-	abiMap       map[string]*abi.ABI
 	nonceManager *NonceManager
 	signer       types.Signer
+	gpm          *GasPriceMonitor
+	tm           ethereum.TransactionSender
```
kyriediculous: prolly better to just refer to `*TransactionManager`

cmd/livepeer/livepeer.go
```diff
@@ -134,6 +134,7 @@ func main() {
 	ethKeystorePath := flag.String("ethKeystorePath", "", "Path for the Eth Key")
 	ethOrchAddr := flag.String("ethOrchAddr", "", "ETH address of an on-chain registered orchestrator")
 	ethUrl := flag.String("ethUrl", "", "Ethereum node JSON-RPC URL")
+	replaceTx := flag.Bool("replaceTransactions", true, "Set to true to replace ethereum transactions with a higher gas price if they time out")
```
yondonfu: What about a `-maxTxReplacements <INTEGER>` flag instead? The value for this flag is the # of times a tx can be replaced (also the # of times the gas price of a tx can be increased). The default value could be something that feels reasonable like 3. The user could set the value to 0 to disable tx replacements.

While an excessively high # of tx replacements may be unlikely in practice if things are implemented correctly, setting a ceiling will be safer to account for pathological cases where the node might just keep endlessly replacing txs. And for users that want to be more careful, they can set a lower ceiling. With this in mind, I think it makes sense to use the same flag to express both enabling/disabling tx replacements and the max # of tx replacements.

pm/queue.go
```diff
@@ -131,7 +131,7 @@ ticketLoop:
 						// If the ticket is used, we can mark it as redeemed
 						if res.err != nil {
 							glog.Errorf("Error redeeming err=%v", res.err)
-							if res.err != errIsUsedTicket {
+							if res.err != errIsUsedTicket && err != errTicketRevert {
```
yondonfu:
```suggestion
							if res.err != errIsUsedTicket && res.err != errTicketRevert {
```

eth/backend.go
```diff
@@ -197,18 +147,48 @@ func (b *backend) retryRemoteCall(remoteCall func() ([]byte, error)) (out []byte
...
 
-func makeABIMap() (map[string]*abi.ABI, error) {
+func makeABIMap() map[string]*abi.ABI {
 	abiMap := make(map[string]*abi.ABI)
 
 	for _, ABI := range abis {
 		parsedAbi, err := abi.JSON(strings.NewReader(ABI))
 		if err != nil {
-			return map[string]*abi.ABI{}, err
+			glog.Errorf("Error creating ABI mappings, can not parse transaction logs err=%v", err)
```
yondonfu:
```suggestion
			glog.Errorf("Error creating ABI map err=%v", err)
```

since there is no transaction log parsing in this function.

eth/client.go
```diff
@@ -883,84 +888,31 @@ func (c *client) ContractAddresses() map[string]ethcommon.Address {
...
+	for {
+		select {
+		case err := <-txSub.Err():
+			return err
+		case event := <-txs:
+			if tx.Hash() == event.TxHash {
+				if event.Status == uint64(0) {
+					return fmt.Errorf("transaction failed txHash=%v", event.TxHash)
+				}
+				break txLoop
```
yondonfu: Can we just return nil here?

eth/transactionManager.go
```diff
@@ -0,0 +1,233 @@
...
+		for {
+			if err := tm.wait(tx); err == context.DeadlineExceeded && tm.replaceTx {
+				newTx, replaceErr := tm.replace(tx)
+				if replaceErr != nil {
+					glog.Error(replaceErr)
+					break
+				}
+				tx = newTx
+			} else {
+				glog.Error(err)
```
yondonfu: There should be an error check here. Otherwise this will log err even if it is nil.

eth/transactionManager.go
```diff
@@ -0,0 +1,233 @@
...
+	defer cancel()
+
+	receipt, err := bind.WaitMined(ctx, tm.eth, tx)
+	if err != nil {
+		return err
+	}
+
+	tm.feed.Send(receipt)
+
+	if receipt.Status == uint64(0) {
```
yondonfu: What about just letting the caller i.e. CheckTx() handle this check? Otherwise, we currently have this check here and in CheckTx().

eth/client.go
```diff
@@ -883,84 +888,31 @@ func (c *client) ContractAddresses() map[string]ethcommon.Address {
...
 
 func (c *client) CheckTx(tx *types.Transaction) error {
-	ctx, cancel := context.WithTimeout(context.Background(), c.txTimeout)
-	defer cancel()
-
-	receipt, err := bind.WaitMined(ctx, c.backend, tx)
-	if err != nil {
-		return err
+	txs := make(chan *types.Receipt, 10)
+	b, ok := c.backend.(*backend)
```
yondonfu: Could we avoid this type assertion if the TransactionManager was just a field on `client`? Feels a bit weird to do a type assertion just so we can fetch the TransactionManager field on the underlying backend type.

eth/client.go
```diff
@@ -883,84 +888,31 @@ func (c *client) ContractAddresses() map[string]ethcommon.Address {
...
+	txSub := b.tm.Subscribe(txs)
+	defer txSub.Unsubscribe()
+
+txLoop:
+	for {
+		select {
+		case err := <-txSub.Err():
+			return err
+		case event := <-txs:
+			if tx.Hash() == event.TxHash {
```
yondonfu: Looks like this TransactionManager API currently implies that a tx will always be confirmed which is signaled by receiving a receipt in the subscription channel. Do we really want the API to make this guarantee to the caller? If we end up adding a max # of replacements there seems like there could be edge cases where the max # of replacements is hit without the tx being confirmed. In that scenario, a receipt would never be received in the subscription channel and the caller would wait forever. Perhaps the API should give the caller a receipt *or* an error?

yondonfu (changes requested):

cmd/livepeer/livepeer.go
kyriediculous: I think this can be a neat addition yeah. One thing that I don't really like here is the hardcoded `EThTxTimeout`, while a user can set the number of max replacements to do it is not immeadiatly known how often the replacement will be tried. 

Allowing a `ethTxTimeout` flag could help here but also again increases the amount of config params required. 

What are your thoughts on that?

kyriediculous:
@yondonfu Do we want persistent storage for the transaction confirmation queue to avoid unexpected behaviour between node restarts when transactions are still pending ?

eth/client.go
kyriediculous: The contract bindings use `Backend` to send transactions which uses `TransactionManager` under the hood. 

The `Backend` does not have access to the `client`.

eth/client.go
kyriediculous: Yeah in retrospect I like your idea better than the current implementation, this will also give us access to revert reason error messages being returned back through `checkTx` 

Updated this API and its implementation in 41d57f11cc1d93e63a2345f082783d00eb600e24

eth/client.go
yondonfu: I was thinking something like this:

```golang
type client struct {
    // Other fields
    ...
    transactionManager *TransactionManager // This is the same TransactionManager used by the backend
}

func (c *client) CheckTx(tx *types.Transaction) error {
    receipts := make(chan *transactionReceipt, 10)
    txSub := c.transactionManager.Subscribe(receipts)
    defer txSub.Unsubscribe()
    ...
}
```

The main reason for this suggestion is that it feels a bit weird to break the Backend interface abstraction with a type assertion just to fetch the TransactionManager field on the underlying backend type. If we're breaking the Backend interface abstraction within the client then it raises the question of why the Backend interface is being used if we're exposing the implementation details of the underlying type to the client.

eth/gaspricemonitor.go
```diff
@@ -125,6 +127,18 @@ func (gpm *GasPriceMonitor) Stop() error {
 	return nil
 }
 
+func (gpm *GasPriceMonitor) SetMaxGasPrice(gp *big.Int) {
+	gpm.gasPriceMu.Lock()
+	defer gpm.gasPriceMu.Unlock()
+	gpm.maxGasPrice = gp
+}
+
+func (gpm *GasPriceMonitor) MaxGasPrice() *big.Int {
```
yondonfu, line 168: I think the maxGasPrice field should also be initialized in the constructor.

eth/client.go
```diff
@@ -148,34 +145,42 @@ type client struct {
...
 }
 
-func NewClient(accountAddr ethcommon.Address, keystoreDir, password string, eth *ethclient.Client, controllerAddr ethcommon.Address, txTimeout time.Duration, maxGasPrice *big.Int) (LivepeerEthClient, error) {
+type LivepeerEthClientConfig struct {
+	AccountAddr       ethcommon.Address
+	KeystoreDir       string
+	Password          string
+	ControllerAddr    ethcommon.Address
+	TxTimeout         time.Duration
+	MaxGasPrice       *big.Int
```
yondonfu: Doesn't look like this is used anywhere. If we add support for initializing the maxGasPrice field in the GasPriceMonitor constructor we could remove this field since the client is initialized with an existing GasPriceMonitor instance which should already have an initialized maxGasPrice.

eth/transactionManager.go
```diff
@@ -0,0 +1,241 @@
...
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/ethclient"
+	"github.com/ethereum/go-ethereum/event"
+	"github.com/golang/glog"
+	"github.com/livepeer/go-livepeer/common"
+)
+
+type TransactionManager struct {
+	txTimeout       time.Duration
+	maxReplaceMents int
```
yondonfu:
```suggestion
	maxTxReplacements int
```

eth/transactionManager.go
```diff
@@ -0,0 +1,241 @@
...
+	am  AccountManager
+
+	cond sync.Cond
+
+	quit chan struct{}
+}
+
+type transactionQueue []*types.Transaction
+
+type transactionReceipt struct {
```
yondonfu, line 52: Would `transactionResult` be a more clearer name given that this struct already holds a transaction receipt?

eth/client.go
```diff
@@ -148,34 +145,42 @@ type client struct {
...
 	if err != nil {
 		return nil, err
 	}
-	backend.SetMaxGasPrice(maxGasPrice)
 
-	am, err := NewAccountManager(accountAddr, keystoreDir, signer)
-	if err != nil {
-		return nil, err
-	}
+	tm := NewTransactionManager(eth, gpm, am, cfg.TxTimeout, cfg.MaxTxReplacements)
```
yondonfu: Doesn't look like `Start()` and `Stop()` are ever called for the TransactionManager which would be required to start the checkTxLoop.

eth/transactionManager.go
```diff
@@ -0,0 +1,241 @@
...
+		var (
+			receipt *types.Receipt
+			err     error
+		)
+
+		for i := 0; i <= tm.maxReplaceMents; i++ {
+			if receipt, err = tm.wait(tx); err == context.DeadlineExceeded && tm.maxReplaceMents > 0 {
+				newTx, replaceErr := tm.replace(tx)
+				if replaceErr != nil {
+					glog.Error(replaceErr)
```
yondonfu: I think `replaceError` should be propagated back to the feed so that a consumer of the feed can get error information about the failed replacement. At the moment, after we break here, the value sent to the feed will just include the `err` variable which would just be `context.DeadlineExceeded`.

eth/transactionManager.go
```diff
@@ -0,0 +1,241 @@
...
+					break
+				}
+				tx = newTx
+			} else {
+				if err != nil {
+					glog.Error(err)
```
yondonfu: I think this error should be propagated back to the feed so that a consumer of the feed can get error information about the failed replacement. At the moment, after we break here, the value sent to the feed will just include the original `err` variable which would be `context.DeadlineExceeded`.

pm/sendermonitor.go
```diff
@@ -422,7 +424,7 @@ func (sm *LocalSenderMonitor) redeemWinningTicket(ticket *SignedTicket) (*types.
 		if monitor.Enabled {
 			monitor.TicketRedemptionError()
 		}
-		return nil, err
+		return nil, errTicketRevert
```
yondonfu: Returning `errTicketRevert` here obfuscates the error returned by `CheckTx` which also includes the hash of the reverted tx. As a result, we lose access to that information. Could we continue passing the error from `CheckTx` here and having a different check for a reverted tx error?

yondonfu (changes requested):
How were these changes tested?

Given that the logic of the transaction manager for handling replacing pending transactions is important for funds management and node operation could we add unit tests particularly around the automated transaction replacement flow?

eth/transactionManager.go
kyriediculous: This error is already propagated back to the feed by default and can only ever result in it being `context.DeadlineExceeded` when the node isn't set to replace transactions in which case we likely want to notify the user to manually check confirmation through this error.

cmd/devtool/devtool.go
```diff
@@ -159,10 +161,42 @@ func ethSetup(ethAcctAddr, keystoreDir string, isBroadcaster bool) {
 	}
 	glog.Infof("Using controller address %s", ethController)
 
-	client, err := eth.NewClient(ethcommon.HexToAddress(ethAcctAddr), keystoreDir, passphrase, backend,
-		ethcommon.HexToAddress(ethController), ethTxTimeout, nil)
+	gpm := eth.NewGasPriceMonitor(backend, 5*time.Second, big.NewInt(0), nil)
+
+	chainID, err := backend.ChainID(context.Background())
+	if err != nil {
+		glog.Errorf("failed to get chain ID from remote ethereum node: %v", err)
```
yondonfu:
```suggestion
		glog.Errorf("Failed to get chain ID from remote ethereum node: %v", err)
```

cmd/devtool/devtool.go
```diff
@@ -159,10 +161,42 @@ func ethSetup(ethAcctAddr, keystoreDir string, isBroadcaster bool) {
...
+		return
+	}
+
+	signer := types.NewEIP155Signer(chainID)
+
+	am, err := eth.NewAccountManager(ethcommon.HexToAddress(ethAcctAddr), keystoreDir, signer)
+	if err != nil {
+		glog.Errorf("error creating Ethereum account manager: %v", err)
```
yondonfu:
```suggestion
		glog.Errorf("Error creating Ethereum account manager: %v", err)
```

cmd/devtool/devtool.go
```diff
@@ -159,10 +161,42 @@ func ethSetup(ethAcctAddr, keystoreDir string, isBroadcaster bool) {
...
+		return
+	}
+
+	if err := am.Unlock(passphrase); err != nil {
+		glog.Errorf("error unlocking Ethereum account: %v", err)
+		return
+	}
+
+	tm := eth.NewTransactionManager(backend, gpm, am, 5*time.Minute, 0)
+	tm.Start()
```
yondonfu: This will block so it should be called in a goroutine.

cmd/livepeer/livepeer.go
```diff
@@ -390,7 +392,41 @@ func main() {
 			bigMaxGasPrice = big.NewInt(int64(*maxGasPrice))
 		}
 
-		client, err := eth.NewClient(ethcommon.HexToAddress(*ethAcctAddr), keystoreDir, *ethPassword, backend, ethcommon.HexToAddress(*ethController), EthTxTimeout, bigMaxGasPrice)
+		gpm := eth.NewGasPriceMonitor(backend, blockPollingTime, minGasPrice, bigMaxGasPrice)
+		// Start gas price monitor
+		_, err = gpm.Start(ctx)
+		if err != nil {
+			glog.Errorf("error starting gas price monitor: %v", err)
```
```suggestion
			glog.Errorf("Error starting gas price monitor: %v", err)
```

cmd/livepeer/livepeer.go
```diff
@@ -390,7 +392,41 @@ func main() {
...
+			return
+		}
+		defer gpm.Stop()
+
+		signer := types.NewEIP155Signer(chainID)
+
+		am, err := eth.NewAccountManager(ethcommon.HexToAddress(*ethAcctAddr), keystoreDir, signer)
+		if err != nil {
+			glog.Errorf("error creating Ethereum account manager: %v", err)
```
```suggestion
			glog.Errorf("Error creating Ethereum account manager: %v", err)
```

cmd/livepeer/livepeer.go
```diff
@@ -390,7 +392,41 @@ func main() {
...
+			return
+		}
+
+		if err := am.Unlock(*ethPassword); err != nil {
+			glog.Errorf("error unlocking Ethereum account: %v", err)
```
```suggestion
			glog.Errorf("Error unlocking Ethereum account: %v", err)
```

cmd/devtool/devtool.go
```diff
@@ -159,10 +161,42 @@ func ethSetup(ethAcctAddr, keystoreDir string, isBroadcaster bool) {
...
+	signer := types.NewEIP155Signer(chainID)
+
+	am, err := eth.NewAccountManager(ethcommon.HexToAddress(ethAcctAddr), keystoreDir, signer)
+	if err != nil {
+		glog.Errorf("error creating Ethereum account manager: %v", err)
+		return
+	}
+
+	if err := am.Unlock(passphrase); err != nil {
+		glog.Errorf("error unlocking Ethereum account: %v", err)
```
yondonfu:
```suggestion
		glog.Errorf("Error unlocking Ethereum account: %v", err)
```

cmd/livepeer/livepeer.go
```diff
@@ -390,7 +392,41 @@ func main() {
...
+			return
+		}
+
+		tm := eth.NewTransactionManager(backend, gpm, am, EthTxTimeout, *maxTxReplacements)
+		tm.Start()
```
yondonfu: This will block so it needs to be called in a goroutine.

eth/client.go
```diff
@@ -883,84 +870,28 @@ func (c *client) ContractAddresses() map[string]ethcommon.Address {
...
+		select {
+		case err := <-txSub.Err():
+			return err
+		case receipt := <-receipts:
+			if tx.Hash() == receipt.TxHash {
+				if receipt.err != nil {
+					return receipt.err
+				}
+				if receipt.Status == uint64(0) {
+					return fmt.Errorf("transaction failed txHash=%v", receipt.TxHash)
```
yondonfu:
```suggestion
					return fmt.Errorf("transaction failed txHash=%v", receipt.TxHash.Hex())
```

I think we need to call Hex() in order to log a hex string here?

eth/transactionManager.go
```diff
@@ -0,0 +1,242 @@
...
+	txTimeout       time.Duration
+	maxReplacements int
+
+	queue transactionQueue
+
+	// subscriptions
+	feed  event.Feed
+	scope event.SubscriptionScope
+
+	eth *ethclient.Client
```
yondonfu: A thought on how we could test the tx replacement behavior:

Replacing this `ethclient.Client` with an interface might make writing tests for the replacement logic easier. 

Something like:

```golang
type TransactionSenderReader interface {
	ethereum.TransactionSender
	ethereum.TransactionReader
	// Required for bind.DeployBackend argument in bind.WaitMined
	CodeAt(context.Context, ethcommon.Address, *big.Int) ([]byte, error)
}
```

Then, a stub implementation of the interface could be used in tests allowing us to control how long it takes for a transaction receipt to be returned in `bind.WaitMined()` since the stub implementation could wait X ms before returning a receipt for a call to `TransactionReceipt()`. If the `txTimeout` is set to Y ms such that Y < X, then we should be able to trigger a timeout which should prompt the TransactionManager to submit a replacement tx. The stub implementation could keep track of the txs sent via `SendTransaction()` and we could check the gas price and nonce for txs to assert that replacement txs are actually being sent. Then, we could also write assertions that the TransactionManager does not submit more than `maxReplacements` replacement txs.

eth/transactionManager.go
yondonfu: My mistake - I misread the code. You're right that this error is propagated to the feed.

eth/transactionManager.go
```diff
@@ -0,0 +1,242 @@
...
+				tx = newTx
+			} else {
+				if err != nil {
+					glog.Error(err)
+				}
+				break
+			}
+		}
+
+		tm.feed.Send(&transactionReceipt{*receipt, err})
```
yondonfu: This will trigger a nil pointer error if `receipt` is nil which will happen if we encounter an error with the tx such that we get a nil receipt back. If the field on the struct here was a receipt pointer instead of a receipt value we could avoid this and then it would be the responsibility of the feed consumer to only rely on the receipt field if there is a non-nil error.

eth/transactionManager.go
```diff
@@ -0,0 +1,242 @@
...
+
+		tx := tm.queue.pop()
+		tm.cond.L.Unlock()
+
+		var (
+			receipt *types.Receipt
+			err     error
+		)
+
+		for i := 0; i <= tm.maxReplacements; i++ {
```
yondonfu: Suggestion for a refactor here that I think flattens the code structure to improve readability:

```golang
receipt, err = tm.wait(tx)
// context.DeadlineExceeded indicates that we hit the txTimeout
// If we hit the txTimeout, replace the tx up to maxReplacements times
for i := 0; err == context.DeadlineExceeded && i < tm.maxReplacements; i++ {
	tx, err = tm.replace(tx)
	// Do not attempt additional replacements if there was an error submitting this
	// replacement tx
	if err != nil {
		break
	}
	receipt, err = tm.wait(tx)
}
```

pm/sendermonitor.go
```diff
@@ -25,6 +25,8 @@ var unixNow = func() int64 {
 	return time.Now().Unix()
 }
 
+type checkTxError error
```
yondonfu:
```suggestion
type errCheckTx error
```

To match the error naming convention used elsewhere in this package.

yondonfu (changes requested):

eth/transactionManager.go
kyriediculous: Yeah probably better to use a pointer here but copying the actual values felt much cleaner and it does feel like we sometimes might overuse pointers a bit which might lead to memory safety issues that can be hard to debug (started thinking about this after the discussing about golang vs rust in discord). However that's unlikely to be much of an issue here so yeah; pointer.

eth/transactionManager.go
kyriediculous: I understand that it's a little bit cleaner and shorter but what I generally dislike is having the same function call before the loop (to 'initialize') as inside of it. 

However this suggestion made me realise a small error in the conditional of the loop in my code which also shortens that up.

kyriediculous: Ended up going with the suggestion because otherwise we don't wait for the last iteration of the replacement loop and thus the last replacement tx

cmd/livepeer/livepeer.go
```diff
@@ -134,6 +135,7 @@ func main() {
 	ethKeystorePath := flag.String("ethKeystorePath", "", "Path for the Eth Key")
 	ethOrchAddr := flag.String("ethOrchAddr", "", "ETH address of an on-chain registered orchestrator")
 	ethUrl := flag.String("ethUrl", "", "Ethereum node JSON-RPC URL")
+	maxTxReplacements := flag.Int("maxTransactionReplacements", 0, "Number of times to automatically replace pending Ethereum transactions")
```
yondonfu: What do you think about setting the default value for this flag to 2? This would enable tx replacements by default which helps address scenarios like super low gas price txs without requiring manual intervention from the node operator. A node operator could always disable the feature by manually passing `-maxTransactionReplacements 0` on startup. The suggestion of 2 is fairly arbitrary although it feels like a reasonable number of times to replace a tx.

eth/transactionManager.go
yondonfu: Bump.

eth/transactionManager.go
```diff
@@ -0,0 +1,259 @@
...
+func (tm *TransactionManager) Stop() {
+	tm.scope.Close()
+	close(tm.quit)
+}
+
+func (tm *TransactionManager) wait(tx *types.Transaction) (*types.Receipt, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), tm.txTimeout)
+	defer cancel()
+
+	receipt, err := bind.WaitMined(ctx, tm.eth, tx)
```
yondonfu: Can shorten this to just:

```golang
return bind.WaitMined(ctx, tm.eth, tx)
```

eth/transactionManager.go
```diff
@@ -0,0 +1,259 @@
...
+	if err != nil && err != ethereum.NotFound {
+		return nil, err
+	}
+	// If tx was found
+	// If `pending` is false, the tx was mined and included in a block
+	if err == nil && !pending {
+		return nil, ErrReplacingMinedTx
+	}
+
+	// Updated gas price must be at least 10% greater than the gas price used for the original transaction in order
```
yondonfu: Let's move this comment into `calcReplacementGasPrice()`

eth/transactionManager.go
```diff
@@ -0,0 +1,259 @@
...
+	// since the time of the original tx submission we cannot use the lower suggested gas price and we still need to use
+	// the bumped gas price in order to properly replace a still pending tx
+	if suggestedGasPrice.Cmp(gasPrice) == 1 {
+		gasPrice = suggestedGasPrice
+	}
+
+	// Bump gas price exceeds max gas price, return early
+	max := tm.gpm.MaxGasPrice()
+	if gasPrice.Cmp(max) > 0 {
+		return nil, fmt.Errorf("Replacement gas price exceeds max gas price suggested=%v max=%v", gasPrice, max)
```
yondonfu:
```suggestion
		return nil, fmt.Errorf("replacement gas price exceeds max gas price suggested=%v max=%v", gasPrice, max)
```

To follow convention of error messages in lowercase (except if its extra annotation text in an error log statement)

eth/transactionManager_test.go
```diff
@@ -0,0 +1,320 @@
...
+	assert.EqualError(err, expErr.Error())
+	eth.err["TransactionByHash"] = nil
+
+	// Test no error - ErrReplacingMinedTx
+	eth.pending = false
+	tx, err = tm.replace(stubTx)
+	assert.Nil(tx)
+	assert.EqualError(err, ErrReplacingMinedTx.Error())
+
+	// Test error is eth.TransactionNotFound - fail at next step
```
yondonfu:
```suggestion
	// Test error is ethereum.NotFound - fail at next step
```

cmd/devtool/devtool.go
kyriediculous: Sorry missed this one, fixed in the latest commit

eth/transactionManager_test.go
```diff
@@ -0,0 +1,320 @@
...
+	tx, err = tm.replace(stubTx)
+	logsAfter = glog.Stats.Info.Lines()
+	assert.Nil(err)
+	expTx := types.NewTransaction(1, *stubTx.To(), stubTx.Value(), 100000, calcReplacementGasPrice(stubTx), stubTx.Data())
+	assert.Equal(tx, expTx)
+	assert.Equal(logsAfter-logsBefore, int64(1))
+
+	// Replacement gas price lower than suggest gas price
+	// Use market gas price
+	gpm.gasPrice = big.NewInt(999)
```
yondonfu, line 245: Could be worthwhile to add an assertion here making sure that the GasPriceMonitor gas price is greater than the value returned by `calcReplacementGasPrice(stubTx)`.

eth/transactionManager_test.go
```diff
@@ -0,0 +1,320 @@
...
+	sink := make(chan *transactionReceipt, 10)
+	sub := tm.Subscribe(sink)
+	defer sub.Unsubscribe()
+	tm.SendTransaction(context.Background(), stubTx)
+
+	event := <-sink
+	assert.NotNil(event)
+	assert.Nil(event.err)
+
+	// Wait error no replacements
```
yondonfu, line 302: Isn't maxReplacement in the TransactionManager still set to 1 here? And in that case, we get a context.DeadlineExceeded error after 2 calls to wait() with the second being for a replacement tx.

eth/transactionManager_test.go
```diff
@@ -0,0 +1,320 @@
...
+	eth.err["TransactionReceipt"] = nil
+
+	// Wait error, replacements
+	// Replace tx error
+	tm.maxReplacements = 1
+	eth.err["TransactionReceipt"] = context.DeadlineExceeded
+	eth.err["TransactionByHash"] = errors.New("TransactionByHash error")
+	sink = make(chan *transactionReceipt)
+	sub = tm.Subscribe(sink)
+	time.Sleep(5 * time.Millisecond)
```
yondonfu: Why is there a sleep here after the Subscribe() call, but not after any of the other previous Subscribe() calls?

eth/transactionManager_test.go
```diff
@@ -0,0 +1,320 @@
...
+	logsBefore = glog.Stats.Info.Lines()
+	tx, err = tm.replace(stubTx)
+	logsAfter = glog.Stats.Info.Lines()
+	assert.Nil(err)
+	expTx = types.NewTransaction(1, *stubTx.To(), stubTx.Value(), 100000, gpm.gasPrice, stubTx.Data())
+	assert.Equal(tx, expTx)
+	assert.Equal(logsAfter-logsBefore, int64(1))
+}
+
+func TestTransactionManager_CheckTxLoop(t *testing.T) {
```
yondonfu, line 256: I think we should add a test to ensure that in the worst case we only submit maxReplacements number of replacement txs. This test would assert that maxReplacements number of replacement txs were submitted, but no more than that.

eth/transactionManager.go
```diff
@@ -0,0 +1,259 @@
...
+
+		tm.feed.Send(&transactionReceipt{
+			Receipt: receipt,
+			err:     err,
+		})
+
+	}
+}
+
+func calcReplacementGasPrice(tx *types.Transaction) *big.Int {
```
yondonfu, line 243: We don't need to address this in this PR, but just making a note that we will probably want to eventually make the gas price bump a parameter given that its a value that influences how soon a replacement tx will be mined and given that the gas price bump required by an ETH node technically doesn't have to be 10% (that just happens to be the default on geth, but there is an option to customize it).

eth/rewardservice.go
```diff
@@ -91,30 +91,11 @@ func (s *RewardService) tryReward() error {
...
-		}
-
-		tp, err := s.client.GetTranscoderEarningsPoolForRound(s.client.Account().Address, currentRound)
-		if err != nil {
+		if err := s.client.CheckTx(tx); err != nil {
 			return err
 		}
 
-		glog.Infof("Called reward for round %v - %v rewards minted", currentRound, FormatUnits(tp.RewardPool, "LPTU"))
+		glog.Infof("Called reward for round %v", currentRound)
```
yondonfu, line 98: Just noting that this is related to https://github.com/livepeer/go-livepeer/issues/1822

yondonfu (changes requested):

cmd/livepeer/livepeer.go
yondonfu: Oops missed this comment in the first couple of reviews.

I think it makes sense to allow the timeout to be configurable. What about `-transactionTimeout`?

cmd/livepeer/livepeer.go
kyriediculous: A feature that changes the existing behaviour should in my opinion not be enabled by default.

kyriediculous: But I'm open to discuss this, I feel like if replacements should be the default then I think the default should be one replacement, which at least for the reward calls retains the same behaviour.

eth/transactionManager.go
kyriediculous: Wouldn't that only be supportable by self-hosted geth nodes because if the configuration on the livepeer node is set to a lower replacement bump than i.e. infura you would still see `replacement tx underpriced` no ?

eth/rewardservice.go
kyriediculous: Yeah sorry for including it here, it doesn't fully address #1822 in that rather than fixing the value shown I decided to remove it for now to avoid confusion.

eth/transactionManager.go
kyriediculous: I went with `transactionReceipt` since the receipt values are embedded in this type (through albeit through a pointer) with the addition of an error (which can be the revert message in some cases).

eth/transactionManager_test.go
kyriediculous: Would something like this work 840c9f4 ?

eth/transactionManager_test.go
```diff
@@ -296,6 +298,7 @@ func TestTransactionManager_CheckTxLoop(t *testing.T) {
 	event := <-sink
 	assert.NotNil(event)
 	assert.Nil(event.err)
+	sub.Unsubscribe()
```
yondonfu, line 300: Should we remove the deferred Unsubscribe() call on L295 if we're calling Unsubscribe() for each new subscription at the end of each test case?

eth/transactionManager_test.go
yondonfu: Looks good.

eth/rewardservice.go
yondonfu: Yep that's fine I was just making this note for myself as a reminder.

eth/transactionManager.go
yondonfu: Yeah if your gas price bump is lower than the gas price bump required by the ETH provider you would receive a `replacement tx underpriced` error.

Aside from speeding up the confirmation of replacement txs, the motivation for supporting custom gas price bumps is to give the livepeer node operator to adjust the gas price bump (i.e. increase it) used if the ETH provider requires a gas price bump different (i.e. greater) than the current default of 10%. If the node operator does not control the ETH provider (i.e. self-hosted node(s)) and the ETH provider is configured to use a non-default gas price bump (or if the default ever changes) then the current hardcoded gas price bump in the livepeer node may not work.

Anyways this isn't that crucial at the moment, but just wanted to point this out for future reference.

cmd/livepeer/livepeer.go
yondonfu: > A feature that changes the existing behaviour should in my opinion not be enabled by default.

Well that depends on whether the existing default behavior is a problem. Based on #1906, I think this is a case where the existing default behavior can actually be a problem. Also, if the default is 0, then we'd be changing existing behavior for reward calls except this is actually hampering the node's ability to deal with getting transactions confirmed in a timely manner for reward calls instead of improving it.

> I feel like if replacements should be the default then I think the default should be one replacement

I'd be onboard with a default of 1 to be a bit more conservative with the default.

yondonfu (commented):
Just two comments to address:

- [Default flag values](https://github.com/livepeer/go-livepeer/pull/1923#discussion_r669943483)
- [Small test related comment](https://github.com/livepeer/go-livepeer/pull/1923#discussion_r669905869)

Let's also rebase on top of master to resolve the conflicts. Feel free to do this in conjunction with addressing the remaining comments.

kyriediculous:
Rebased, resolved merge conflicts and addresses the remaining changes. 

Also moved min and max gas price completely into `GasPriceMonitor` as discussed previously.

eth/backend.go
```diff
@@ -83,63 +75,43 @@ func (b *backend) PendingNonceAt(ctx context.Context, account common.Address) (u
...
-	if gp != nil && monitor.Enabled {
-		monitor.MaxGasPrice(gp)
-	}
-}
-
-func (b *backend) MaxGasPrice() *big.Int {
-	b.RLock()
-	defer b.RUnlock()
-	return b.maxGasPrice
+func (b *backend) GasPriceMonitor() *GasPriceMonitor {
```
yondonfu, line 113: Since we're removing the max gas price setters/getters from backend should we also remove the min gas price setters/getters with the expectation that the appropriate setters/getters can be called on the GasPriceMonitor returned here?

eth/gaspricemonitor.go
```diff
@@ -148,6 +151,18 @@ func (gpm *GasPriceMonitor) Stop() error {
 	return nil
 }
 
+func (gpm *GasPriceMonitor) SetMaxGasPrice(gp *big.Int) {
```
yondonfu, line 158: This is currently missing the metrics recording call via the monitor package that was included in the [method definition](https://github.com/livepeer/go-livepeer/blob/0a72c0ac52b97b37bd2fb51c865ac3f2a3592bc3/eth/backend.go#L134) in backend.

eth/gaspricemonitor.go
```diff
@@ -54,6 +56,7 @@ func NewGasPriceMonitor(gpo GasPriceOracle, pollingInterval time.Duration, minGa
 		pollingInterval: pollingInterval,
 		gasPrice:        big.NewInt(0),
 		minGasPrice:     minGasP,
+		maxGasPrice:     maxGasPrice,
```
yondonfu, line 63: Let's add a metrics recording call for maxGasPrice similar to the one we have for minGasPrice on L52 (will probably need a nil check as well in case maxGasPrice is nil).

yondonfu (changes requested):

eth/backend.go
kyriediculous: Oh yeah forgot to remove these, the calls were already updated to call them on the GPM instead

eth/gaspricemonitor.go
kyriediculous: fixed in 91ceac9

eth/gaspricemonitor.go
kyriediculous: fixed in 91ceac9

eth/backend.go
kyriediculous: fixed in 91ceac9

yondonfu (approved):
LGTM after squashing 🚢

PR #1926, opened 2021-06-22
ci: Separate pipeline for tensorflow-enabled linux builds

### LINKED ISSUE #1919: Fix run-time `libtensorflow.so` dependency for non-scene-classification node env

jailuthra:

Separate out the build process when `libtensorflow` is enabled.

*see commit history*

The CI builds + uploads work as expected.
https://build.livepeer.live/0.5.18-3aa0d8ca/livepeer-linux-tensorflow-amd64.tar.gz <-- Depends on libtensorflow.so - `ldd livepeer | grep tensorflow` 
https://build.livepeer.live/0.5.18-3aa0d8ca/livepeer-linux-amd64.tar.gz <-- Standard build with no runtime dependency

Fixes #1919 

**Checklist:**

### DIFF
--- a/.github/workflows/linux-tensorflow.yml
+++ b/.github/workflows/linux-tensorflow.yml
@@ -1,4 +1,4 @@
-name: Linux Build
+name: Linux Tensorflow Build
 on:
   pull_request:
   push:
@@ -31,14 +31,14 @@ jobs:
       uses: actions/cache@v2.1.5
       with:
         path: ~/compiled
-        key: ${{ runner.os }}-ffmpeg-${{ hashFiles('**/install_ffmpeg.sh') }}
+        key: ${{ runner.os }}-ffmpeg-tensorflow-${{ hashFiles('**/install_ffmpeg.sh') }}
         restore-keys: |
-          ${{ runner.os }}-ffmpeg
+          ${{ runner.os }}-ffmpeg-tensorflow
     - name: Cache binaries
       uses: actions/cache@v2.1.5
       with:
         path: ~/build
-        key: ${{ runner.os }}-binaries-${{ github.sha }}
+        key: ${{ runner.os }}-binaries-tensorflow-${{ github.sha }}
     - name: Install dependencies
       run: |
         sudo apt-get update \
@@ -50,12 +50,17 @@ jobs:
 
         sudo update-alternatives --install /usr/bin/clang++ clang++ /usr/bin/clang++-8 30 \
           && sudo update-alternatives --install /usr/bin/clang clang /usr/bin/clang-8 30
+
+        LIBTENSORFLOW_VERSION=2.3.0 \
+          && curl -LO https://storage.googleapis.com/tensorflow/libtensorflow/libtensorflow-gpu-linux-x86_64-${LIBTENSORFLOW_VERSION}.tar.gz \
+          && sudo tar -C /usr/local -xzf libtensorflow-gpu-linux-x86_64-${LIBTENSORFLOW_VERSION}.tar.gz \
+          && sudo ldconfig
     - name: Install go modules
       if: steps.cache-go-mod.outputs.cache-hit != 'true'
       run: go mod download
     - name: Install ffmpeg
       if: steps.cache-ffmpeg.outputs.cache-hit != 'true'
-      run: ./install_ffmpeg.sh
+      run: BUILD_TAGS=experimental ./install_ffmpeg.sh
     - name: Build binaries
       run: |
         export PKG_CONFIG_PATH=~/compiled/lib/pkgconfig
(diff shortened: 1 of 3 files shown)


install_ffmpeg.sh
```diff
@@ -111,7 +111,12 @@ else
   if which clang > /dev/null; then
     echo "clang detected, building with GPU support"
 
-    EXTRA_FFMPEG_FLAGS="--enable-cuda --enable-cuda-llvm --enable-cuvid --enable-nvenc --enable-decoder=h264_cuvid --enable-filter=scale_cuda --enable-encoder=h264_nvenc --enable-libtensorflow --enable-filter=lvpdnn"
+    EXTRA_FFMPEG_FLAGS="--enable-cuda --enable-cuda-llvm --enable-cuvid --enable-nvenc --enable-decoder=h264_cuvid --enable-filter=scale_cuda --enable-encoder=h264_nvenc"
+
+    if [[ $BUILD_TAGS == *"experimental"* ]]; then
+        echo "experimental tag detected, building with Tensorflow support"
+        EXTRA_FFMPEG_FLAGS="--enable-tensorflow"
```
yondonfu: Think we want to add `--enable-tensorflow` to `EXTRA_FFMPEG_FLAGS` and not set the var which would overwrite its existing value yeah?

install_ffmpeg.sh
```diff
@@ -111,7 +111,12 @@ else
   if which clang > /dev/null; then
     echo "clang detected, building with GPU support"
 
-    EXTRA_FFMPEG_FLAGS="--enable-cuda --enable-cuda-llvm --enable-cuvid --enable-nvenc --enable-decoder=h264_cuvid --enable-filter=scale_cuda --enable-encoder=h264_nvenc --enable-libtensorflow --enable-filter=lvpdnn"
+    EXTRA_FFMPEG_FLAGS="--enable-cuda --enable-cuda-llvm --enable-cuvid --enable-nvenc --enable-decoder=h264_cuvid --enable-filter=scale_cuda --enable-encoder=h264_nvenc"
+
+    if [[ $BUILD_TAGS == *"experimental"* ]]; then
```
yondonfu, line 116: Oh cool does the `*...*` syntax result in a substring check for "experimental" in `BUILD_TAGS`?

install_ffmpeg.sh
```diff
@@ -130,7 +135,7 @@ if [ ! -e "$HOME/ffmpeg/libavcodec/libavcodec.a" ]; then
     --enable-bsf=h264_mp4toannexb,aac_adtstoasc,h264_metadata,h264_redundant_pps,extract_extradata \
     --enable-parser=aac,aac_latm,h264 \
     --enable-filter=abuffer,buffer,abuffersink,buffersink,afifo,fifo,aformat \
-    --enable-filter=aresample,asetnsamples,fps,scale \
+    --enable-filter=aresample,asetnsamples,fps,scale,lvpdnn \
```
yondonfu, line 138: Should we conditionally enable `lvpdnn` based on the `experimental` tag as well?

jailuthra: The `vf_lvpdnn.c` file currently has the definition of the model loading functions, whose declaration is in `avfilter.h` that we always include.
Not enabling this filter lead to compiler errors for the undefined methods.

install_ffmpeg.sh
jailuthra: Yep!

install_ffmpeg.sh
jailuthra: Ah good catch, my bad.

install_ffmpeg.sh
jailuthra: But agreed, once we land up on a better solution for model loading on the ffmpeg side - we should conditionally include that filter.

yondonfu: Gotcha thanks for the clarification.

install_ffmpeg.sh
jailuthra: Fixed in https://github.com/livepeer/go-livepeer/pull/1926/commits/19d2c35db875faab4f0e058afeb86bd94685dca6

.github/workflows/linux-tensorflow.yml
```diff
@@ -0,0 +1,102 @@
...
+        LIBTENSORFLOW_VERSION=2.3.0 \
+          && curl -LO https://storage.googleapis.com/tensorflow/libtensorflow/libtensorflow-gpu-linux-x86_64-${LIBTENSORFLOW_VERSION}.tar.gz \
+          && sudo tar -C /usr/local -xzf libtensorflow-gpu-linux-x86_64-${LIBTENSORFLOW_VERSION}.tar.gz \
+          && sudo ldconfig
+    - name: Install go modules
+      if: steps.cache-go-mod.outputs.cache-hit != 'true'
+      run: go mod download
+    - name: Install ffmpeg
+      if: steps.cache-ffmpeg.outputs.cache-hit != 'true'
+      run: BUILD_TAGS=experimental ./install_ffmpeg.sh
```
jailuthra, line 63: Decided to explicitly enable the experimental tag rather than using the `ci_env.sh` script as we'd *always* want a tensorflow build here, even for a tagged release.

yondonfu: So, this GH action will always build Tensorflow enabled binaries. But, the scene classification capability will only be enabled in the binary for non-semver tagged releases (i.e. an orchestrator would not advertise the capability to a broadcaster even though the binary itself has Tensorflow support). Seems reasonable.

yondonfu (commented):
Summary of my current understanding of the CI process:

- Linux
    - Tensorflow Build
        - Always enables Tensorflow. User will need to download libtensorflow themselves to use with the binary
        - Only enables scene classification capability for non-semver tagged releases
    - Regular Build
        - Never enables Tensorflow
        - Only enables scene classification capability for non-semver tagged releases
        - Since Tensorflow is never enabled, scene classification can be enabled for a non-semver tagged release, but will not actually work
 - Docker
     - Always enables Tensorflow and libtensorflow is included in the image
     - Only enables scene classification capability for non-semver tagged releases
 - Windows
     - Never enables Tensorflow
     - Only enables scene classification capability for non-semver tagged releases
     - Since Tensorflow is never enabled, scene classification can be enabled for a non-semver tagged release, but will not actually work
 - Darwin
     - Never enables Tensorflow
     - Only enables scene classification capability for non-semver tagged releases
     - Since Tensorflow is never enabled, scene classification can be enabled for a non-semver tagged release, but will not actually work

yondonfu (approved):
Just want to make sure we're on the same page about [this comment](https://github.com/livepeer/go-livepeer/pull/1926#pullrequestreview-689817272).

Other than that LGTM after squashing.

jailuthra:
> Summary of my current understanding of the CI process:
> ..

Thanks for the summary, and yeah that sounds about right!
Keeping it functional only for linux + non-semver because the usecase (for now) is limited to our deployments and not to the wider community.

PR #1930, opened 2021-06-24
Support custom minimum gas price

### LINKED ISSUE #1914: Enable customization of `minGasPrice` value
## Problem
The `minGasPrice` value is leveraged in the new `GasPriceMonitor` module. It's used to filter out garbage gas price values returned from the ETH client's `eth_gasPrice` RPC call. 

Read [this tweet thread ](https://twitter.com/YondonFu/status/1386693693088964609)to understand how the evolving MEV environment on Ethereum plays a part in negatively affecting this functionality recently (and will continue too).

Currently, the value of `minGasPrice` is not customizable per node operator. It is hardcoded to 1 Gwei as seen below:
1. [`minGasPrice` is set in the `GasPriceMonitor` constructor, passed in as a parameter](https://github.com/livepeer/go-livepeer/blob/48409dfcfe988a02f5166d64db2834b4dea4b6f0/eth/gaspricemonitor.go#L44))
2. [Where the constructor of `GasPriceMonitor` is called here, passing in `minGasPrice`](https://github.com/livepeer/go-livepeer/blob/52c797a09ce5a51dce49d1b933b54745713bec10/cmd/livepeer/livepeer.go#L550)
3. [With the value for `minGasPrice` being defined here](https://github.com/livepeer/go-livepeer/blob/52c797a09ce5a51dce49d1b933b54745713bec10/cmd/livepeer/livepeer.go#L207)
4. [Which imports its value from the Geth namespace here
](https://github.com/livepeer/go-livepeer/blob/52c797a09ce5a51dce49d1b933b54745713bec10/cmd/livepeer/livepeer.go#L25)
5. [Where the value imported from the Geth namespace is defined here, as 1 Gwei](https://github.com/ethereum/go-ethereum/blob/master/params/denomination.go#L26)

Setting `minGasPrice` to 1 Gwei may be substantial enough to solve the original issue, since it weeds out 0 and 1 wei gas prices. However, I see the role of `minGasPrice` expanding within the Livepeer client, as the `GasPriceMonitor` module expands. Hardcoding the value to 1 Gwei isn't nearly as effective.

See this comment thread #1906 to understand the new suggested role of `GasPriceMonitor`, where it will act as an appropriate filter for `gasPrice` values, which includes screening for too low gas prices.

Some Orchestrators/node operators would prefer more customization on there Livepeer client txns. Currently, txn gas prices default to the same across operators, based on what the ETH client (ie. Geth) suggests. This is susceptible to quick rises in gas prices which will cause pending/hanging txns, etc.

**Example:** Setting a custom higher `minGasPrice` to put priority on their broadcast txns such as `rewardWithHint()` which has a deadline, and must complete within each appropriate round in order to be valid and claim rewards.

Giving this optionality to node operators leads to more predictable business operations as well. If the Livepeer client screens gas prices it's using to broadcast txns with `minGasPrice`, an operator can evaluate "safe" gas prices to use where they're willing to spend the ETH required.

**Example:** Node operator sets their `minGasPrice` at 40 Gwei. Now they have peace of mind they're client will never try to broadcast a txn below 40 Gwei, and they have more control over how much risk and optimization they wish to partake in via txn costs/hanging txns, etc.

Also, note there's a `maxGasPrice` CLI argument, which sets precedent for such txn arguments to be customizable per node operator.

# Solution
A potential solution is to add a `minGasPrice` CLI argument. It can follow the existing simple pattern of other valid CLI arguments such as `maxGasPrice`.

yondonfu:

Adds support for configuring a custom minimum gas price using the `-minGasPrice` flag, requests to the `/setMinGasPrice` webserver endpoint and `livepeer_cli`.

This PR is based off of https://github.com/livepeer/go-livepeer/pull/1915 which injects GasPriceMonitor into the backend type defined in `eth` so in this PR we define setters/getters for minGasPrice for GasPriceMonitor which are used by the backend.

See commit history.

Added unit tests and tested manually using devtool to confirm that if the last polled gas price < minGasPrice, txs will be submitted with minGasPrice.

Fixes #1914

**Checklist:**

### DIFF
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -114,6 +114,7 @@ func main() {
 	ethOrchAddr := flag.String("ethOrchAddr", "", "ETH address of an on-chain registered orchestrator")
 	ethUrl := flag.String("ethUrl", "", "Ethereum node JSON-RPC URL")
 	gasLimit := flag.Int("gasLimit", 0, "Gas limit for ETH transactions")
+	minGasPrice := flag.Int64("minGasPrice", 0, "Minimum gas price for ETH transactions")
 	maxGasPrice := flag.Int("maxGasPrice", 0, "Maximum gas price for ETH transactions")
 	ethController := flag.String("ethController", "", "Protocol smart contract address")
 	initializeRound := flag.Bool("initializeRound", false, "Set to true if running as a transcoder and the node should automatically initialize new rounds")
@@ -173,7 +174,7 @@ func main() {
 
 	type NetworkConfig struct {
 		ethController string
-		minGasPrice   *big.Int
+		minGasPrice   int64
 	}
 
 	ctx := context.Background()
@@ -184,7 +185,7 @@ func main() {
 		},
 		"mainnet": {
 			ethController: "0xf96d54e490317c557a967abfa5d6e33006be69b3",
-			minGasPrice:   big.NewInt(int64(params.GWei)),
+			minGasPrice:   int64(params.GWei),
 		},
 	}
 
@@ -207,12 +208,15 @@ func main() {
 	}
 
 	// Setting config options based on specified network
-	var minGasPrice = big.NewInt(0)
 	if netw, ok := configOptions[*network]; ok {
 		if *ethController == "" {
 			*ethController = netw.ethController
-			minGasPrice = netw.minGasPrice
 		}
+
+		if !isFlagSet["minGasPrice"] {
+			*minGasPrice = netw.minGasPrice
+		}
+
 		glog.Infof("***Livepeer is running on the %v network: %v***", *network, *ethController)
 	} else {
 		glog.Infof("***Livepeer is running on the %v network***", *network)
@@ -380,7 +384,7 @@ func main() {
 			bigMaxGasPrice = big.NewInt(int64(*maxGasPrice))
 		}
 
-		gpm := eth.NewGasPriceMonitor(backend, blockPollingTime, minGasPrice)
+		gpm := eth.NewGasPriceMonitor(backend, blockPollingTime, big.NewInt(*minGasPrice))
 		// Start gas price monitor
 		_, err = gpm.Start(ctx)
 		if err != nil {
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -509,7 +509,7 @@ func main() {
 			CleanupInterval: cleanupInterval,
 			TTL:             smTTL,
 			RedeemGas:       redeemGas,
-			SuggestGasPrice: backend.SuggestGasPrice,
+			SuggestGasPrice: client.Backend().SuggestGasPrice,
 			RPCTimeout:      ethRPCTimeout,
 		}
 
--- a/eth/backend.go
+++ b/eth/backend.go
@@ -44,6 +44,8 @@ type Backend interface {
 	ChainID(ctx context.Context) (*big.Int, error)
 	MaxGasPrice() *big.Int
 	SetMaxGasPrice(gp *big.Int)
+	MinGasPrice() *big.Int
+	SetMinGasPrice(gp *big.Int)
 }
 
 type backend struct {
@@ -133,6 +135,14 @@ func (b *backend) MaxGasPrice() *big.Int {
 	b.RLock()
 	defer b.RUnlock()
 	return b.maxGasPrice
+}
+
+func (b *backend) SetMinGasPrice(gp *big.Int) {
+	b.gpm.SetMinGasPrice(gp)
+}
+
+func (b *backend) MinGasPrice() *big.Int {
+	return b.gpm.MinGasPrice()
 }
 
 type txLog struct {
--- a/eth/client.go
+++ b/eth/client.go
@@ -43,7 +43,7 @@ var (
 
 type LivepeerEthClient interface {
 	Account() accounts.Account
-	Backend() (Backend, error)
+	Backend() Backend
 
 	// Rounds
 	InitializeRound() (*types.Transaction, error)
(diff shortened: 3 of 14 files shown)


cmd/livepeer/livepeer.go
```diff
@@ -227,12 +228,15 @@ func main() {
...
 	// Setting config options based on specified network
-	var minGasPrice = big.NewInt(0)
 	if netw, ok := configOptions[*network]; ok {
 		if *ethController == "" {
 			*ethController = netw.ethController
-			minGasPrice = netw.minGasPrice
 		}
+
+		if !isFlagSet["minGasPrice"] {
+			*minGasPrice = netw.minGasPrice
```
yondonfu, line 217: Moved this line out of the `*ethController` conditional block because we'd want to set the min gas price using the network config regardless of whether `-ethController` was specified or not.

cmd/livepeer/livepeer.go
```diff
@@ -515,7 +519,7 @@ func main() {
 			CleanupInterval: cleanupInterval,
 			TTL:             smTTL,
 			RedeemGas:       redeemGas,
-			SuggestGasPrice: backend.SuggestGasPrice,
+			SuggestGasPrice: client.Backend().SuggestGasPrice,
```
yondonfu, line 512: This ensures that [tx cost estimation in LocalSenderMonitor](https://github.com/livepeer/go-livepeer/blob/887fb1c4e4b2aa84f1d890a23d9378d22bfe8d70/pm/sendermonitor.go#L382) uses the gas price returned by GasPriceMonitor.

eth/client.go
```diff
@@ -361,12 +361,8 @@ func (c *client) Account() accounts.Account {
...
 }
 
-func (c *client) Backend() (Backend, error) {
-	if c.backend == nil {
-		return nil, ErrMissingBackend
-	} else {
-		return c.backend, nil
-	}
+func (c *client) Backend() Backend {
+	return c.backend
```
yondonfu, line 365: Dropped the error in the return type for this method because there should always be a backend configured when a client is instantiated and doing so makes it easier to call backend methods when the caller only has access to the client i.e. we can do `client.Backend().SomeMethod()` without additional error handling.

eth/gaspricemonitor.go
```diff
@@ -59,9 +59,25 @@ func (gpm *GasPriceMonitor) GasPrice() *big.Int {
 	gpm.gasPriceMu.RLock()
 	defer gpm.gasPriceMu.RUnlock()
 
+	if gpm.gasPrice.Cmp(gpm.minGasPrice) < 0 {
```
yondonfu, line 62: We were already dropping polled gas prices that were below the min gas price, but this ensures that the lowest value returned by GasPrice() is min gas price.

server/webserver.go
```diff
@@ -1142,11 +1133,12 @@ func (s *LivepeerServer) cliWebServerHandlers(bindAddr string) *http.ServeMux {
 	})
 
 	mux.HandleFunc("/maxGasPrice", func(w http.ResponseWriter, r *http.Request) {
-		b, err := s.LivepeerNode.Eth.Backend()
-		if err != nil {
-			respondWith400(w, err.Error())
+		if s.LivepeerNode.Eth == nil {
```
yondonfu, line 1136: Added this check so `livepeer_cli` doesn't immediately crash the node when running in off-chain mode. Though other things can probably crash the node when running in off-chain mode. I haven't take a thorough look - this just happened to be the one crash scenario that stood out to me.

eth/backend.go
```diff
@@ -44,6 +44,8 @@ type Backend interface {
 	ChainID(ctx context.Context) (*big.Int, error)
 	MaxGasPrice() *big.Int
 	SetMaxGasPrice(gp *big.Int)
+	MinGasPrice() *big.Int
+	SetMinGasPrice(gp *big.Int)
```
kyriediculous, line 48: In https://github.com/livepeer/go-livepeer/pull/1923/files `MaxGasPrice` and `SetMaxGasPrice` are moved to the `GasPriceMonitor` 

Instead backend now exposes `GasPriceMonitor() *GasPriceMonitor`  to give the CLI server access to the `GasPriceMonitor` by calling `node.Eth.Backend().GasPriceMonitor()` for setting and getting the max gas price through the CLI. 

I'm not sure if that makes sense to do but I found it a bit cleaner than having the setters/getters for a particular service (i.e. GasPriceMonitor) live on a different API than the actual service. WDYT of doing that here as well so that here as well ?

kyriediculous (changes requested):
Overall looks good to me 👍  the comments you added in the review were really helpful. I just wanted to point out that https://github.com/livepeer/go-livepeer/pull/1923/files does some restructuring in the `eth` package (see comment) so it might be worth to base off of that PR instead (which in itself using `nv/tx-fixes` as a base branch).

yondonfu:
Rebased off of master and switched the target branch to master as well now that https://github.com/livepeer/go-livepeer/pull/1915 has been merged.

eth/backend.go
yondonfu: As discussed offline, the plan is to merge this PR as-is and address the GasPriceMonitor related changes mentioned in this comment in https://github.com/livepeer/go-livepeer/pull/1923 since that PR refactors the `backend` type in the `eth` package to expose a `GasPriceMonitor()` method.

kyriediculous:
Looks like there is a failing test, I assume because minGasPrice isn't set for the test.

yondonfu:
> Looks like there is a failing test, I assume because minGasPrice isn't set for the test.

Fixed in 5eb1ee1.

kyriediculous:
LGTM let's squash the fixup 👍

yondonfu:
> LGTM let's squash the fixup 👍

Done.

kyriediculous (approved):
🚢

PR #1932, opened 2021-06-26
Serialize writes of JSON playlist

### LINKED ISSUE #1924: Investigate "Error saving json playlist" errors.
**Describe the bug**
In our logs there is error `Error saving json playlist name=playlist_1624048218739776430.json bytes=103516 took=10.000206608s err=context deadline exceeded` constantly appears. Need to find out why.

**To Reproduce**
Check logs.

**Expected behavior**
Shouldn't be any error.

darkdarkdragon:
Fixes `Error saving json playlist` error.
We're making a lot of writes (to object store) to same json playlist file during stream.
Each write starts in own goroutine, so often new writes starts before previous completed.
In that case request to Google Cloud Storage often takes more than 10s, which is our timeout.
This PR fixes problem by serializing write requests.

- Create new `OvewriteQueue` class that serializes writes to the same file in OS.
- Use `OvewriteQueue` for writing JSON playlists.

Unit tests

Fixes #1924 

**Checklist:**


drivers/overwrite_queue.go
```diff
@@ -0,0 +1,119 @@
...
+
+// StopAfter stops reading loop after some time
+func (oq *OvewriteQueue) StopAfter(pause time.Duration) {
+	go func(p time.Duration) {
+		time.Sleep(p)
+		close(oq.quit)
+	}(pause)
+}
+
+// waitForQueueToClear used in tests
```
jailuthra: wdyt about moving this to `drivers/overwrite_queue_test.go`?

drivers/overwrite_queue_test.go
```diff
@@ -0,0 +1,70 @@
...
+
+	mos.waitForCh = true
+	oq.Save(dataw1)
+	<-mos.back
+	oq.Save(data2)
+	oq.Save(data3)
+	mos.waitCh <- nil
+
+	oq.waitForQueueToClear(5 * time.Second)
+	mos.AssertExpectations(t)
```
jailuthra, line 82: for completeness, we could also add assertion for SaveData not being called for "data02" here using https://pkg.go.dev/github.com/stretchr/testify/mock#Mock.AssertNotCalled

currently it may get called along with dataw01, and data03 but we'll miss that in the test.

jailuthra (commented):
Before merge, would it be feasible to test it out in a local dev environment and/or staging, aside from the unit tests?

LGTM otherwise. Will let somebody else mark it as approved, as this is my first dive into this part of the codebase.

drivers/overwrite_queue.go
```diff
@@ -0,0 +1,119 @@
...
+
+func NewOverwriteQueue(session OSSession, name, desc string, maxRetries int, initialTimeout, maxTimeout time.Duration) *OvewriteQueue {
+	oq := &OvewriteQueue{
+		desc:           desc,
+		name:           name,
+		maxRetries:     maxRetries,
+		session:        session,
+		initialTimeout: initialTimeout,
+		maxTimeout:     maxTimeout,
+		queue:          make(chan []byte, 32),
```
yondonfu: Let's use a constant for 32.

drivers/overwrite_queue.go
```diff
@@ -0,0 +1,119 @@
...
+	"github.com/golang/glog"
+	"github.com/livepeer/go-livepeer/common"
+)
+
+const (
+	timeoutMultiplier = 1.5
+)
+
+type (
+	OvewriteQueue struct {
```
yondonfu:
```suggestion
	OverwriteQueue struct {
```

drivers/overwrite_queue.go
```diff
@@ -0,0 +1,119 @@
...
+			return
+		}
+	}
+}
+
+func (oq *OvewriteQueue) getLastMessage(current []byte) []byte {
+	res := current
+	for {
+		select {
+		case data := <-oq.queue:
```
yondonfu, line 98: Just to clarify my own understanding here:

If the queue has multiple items in it I believe this for loop will run until the queue is empty and then return `res` which will be set to the last item in the queue. And this is the desired behavior?

server/broadcast_test.go
```diff
@@ -1380,6 +1380,7 @@ func TestVerifier_Verify(t *testing.T) {
 	assert := assert.New(t)
 
 	os := drivers.NewMemoryDriver(nil).NewSession("")
+	core.JsonPlaylistQuitTimeout = 0 * time.Second
```
yondonfu, line 1386: Let's reset this package var to its original value after this test (i.e. using a defer statement) so that package level state isn't mutated between tests.

server/push_test.go
```diff
@@ -169,6 +169,7 @@ func TestPush_MultipartReturn(t *testing.T) {
 	osd := drivers.NewMemoryDriver(url)
 	osSession := osd.NewSession("testPath")
 
+	core.JsonPlaylistQuitTimeout = 0 * time.Second
```
yondonfu, line 176: Let's reset this package var to its original value after this test (i.e. using a defer statement) so that package level state isn't mutated between tests.

server/push_test.go
```diff
@@ -1221,6 +1223,7 @@ func TestPush_ConcurrentSegments(t *testing.T) {
 
 func TestPush_ReuseIntmidWithDiffExtmid(t *testing.T) {
 	defer goleak.VerifyNone(t, common.IgnoreRoutines()...)
+	core.JsonPlaylistQuitTimeout = 0 * time.Second
```
yondonfu, line 1240: Let's reset this package var to its original value after this test (i.e. using a defer statement) so that package level state isn't mutated between tests.

drivers/overwrite_queue.go
```diff
@@ -0,0 +1,119 @@
...
+
+func (oq *OvewriteQueue) workerLoop() {
+	var err error
+	var took time.Duration
+	for {
+		select {
+		case data := <-oq.queue:
+			timeout := oq.initialTimeout
+			for try := 0; try < oq.maxRetries; try++ {
+				// we only care about last data
```
yondonfu, line 67: So if data is added to the queue really quickly we'll always ignore all data except for the last one in the queue? Why is that?

drivers/overwrite_queue.go
```diff
@@ -0,0 +1,119 @@
...
+						len(data), took, try)
+					break
+				}
+				timeout = time.Duration(float64(timeout) * timeoutMultiplier)
+				if timeout > oq.maxTimeout {
+					timeout = oq.maxTimeout
+				}
+			}
+			if err != nil {
+				glog.Errorf("Error saving %s name=%s bytes=%d took=%s try=%d err=%v", oq.desc, oq.name,
```
yondonfu, line 84: The value of `try=` is always going to be the max # of retries. Should it be the # of the try that succeeded?

drivers/session_mock.go
```diff
@@ -2,17 +2,33 @@ package drivers
...
 
-func (s *MockOSSession) SaveData(name string, data []byte, meta map[string]string) (string, error) {
-	args := s.Called()
+func (s *MockOSSession) init() {
+	s.waitCh = make(chan interface{})
+	s.back = make(chan interface{})
+}
+
+func (s *MockOSSession) SaveData(name string, data []byte, meta map[string]string, timeout time.Duration) (string, error) {
+	fmt.Printf("mock SaveData %s %s %s\n", name, data, timeout)
```
yondonfu: Can remove.

drivers/session_mock.go
```diff
@@ -2,17 +2,33 @@ package drivers
...
+	"time"
 
 	"github.com/livepeer/go-livepeer/net"
 	"github.com/stretchr/testify/mock"
 )
 
 type MockOSSession struct {
 	mock.Mock
+	waitForCh bool
+	waitCh    chan interface{}
```
yondonfu: Can these channels use the struct{} type since it looks like they're just used for signaling and the actual values are never actually used?

drivers/session_mock.go
```diff
@@ -2,17 +2,33 @@ package drivers
...
+	back      chan interface{}
 }
 
-func (s *MockOSSession) SaveData(name string, data []byte, meta map[string]string) (string, error) {
-	args := s.Called()
+func (s *MockOSSession) init() {
```
yondonfu:
```suggestion
func (s *MockOSSession) NewMockOSSession() *MockOSSession {
```

We can just return a properly initialized MockOSSession.

core/playlistmanager.go
```diff
@@ -226,16 +233,30 @@ func NewBasicPlaylistManager(manifestID ManifestID,
...
+		bplm.makeNewOverwriteQueue()
 	}
 	return bplm
 }
 
+func (mgr *BasicPlaylistManager) makeNewOverwriteQueue() {
+	if mgr.jsonListWriteQueue != nil {
+		mgr.jsonListWriteQueue.StopAfter(JsonPlaylistQuitTimeout)
+	}
+	mgr.jsonListWriteQueue = drivers.NewOverwriteQueue(mgr.recordSession, mgr.jsonList.name, "json playlist",
```
yondonfu: Should the manifestID be used instead of "json playlist"?

drivers/overwrite_queue.go
```diff
@@ -0,0 +1,119 @@
...
+	"github.com/livepeer/go-livepeer/common"
+)
+
+const (
+	timeoutMultiplier = 1.5
+)
+
+type (
+	OvewriteQueue struct {
+		desc           string
```
yondonfu, line 17: Should this be a manifestID instead so that the manifestID can always be included in the save data logs?

yondonfu (changes requested):

drivers/overwrite_queue.go
darkdarkdragon: Because previous data will be overwritten with latest data anyway, so there is no point in writing something if we already know that we will overwrite it in the next call (this is why it is called __*Overwrite*__ Queue 😄 )

drivers/overwrite_queue.go
darkdarkdragon: > 
> 
> The value of `try=` is always going to be the max # of retries. Should it be the # of the try that succeeded?

What do you mean by `# of the try that succeeded`? This is message for the error case.

drivers/overwrite_queue.go
darkdarkdragon: Yes. We care only about last data - any data before last one anyway will be overwritten with the last one - this is the whole point of this queue - not to write data that anyway will be overwritten moment later.

drivers/overwrite_queue.go
darkdarkdragon: moved

drivers/overwrite_queue_test.go
darkdarkdragon: added

drivers/overwrite_queue.go
darkdarkdragon: fixed

drivers/overwrite_queue.go
darkdarkdragon: fixed

server/broadcast_test.go
darkdarkdragon: done

server/push_test.go
darkdarkdragon: done

server/push_test.go
darkdarkdragon: done

drivers/session_mock.go
darkdarkdragon: done

drivers/session_mock.go
darkdarkdragon: Changed, but what's the reason for this change? I'm usually using `interface` because I don't need to create value to send into it - I can just send `nil`.

drivers/session_mock.go
darkdarkdragon: Fixed

core/playlistmanager.go
darkdarkdragon: Added manifesID to the description

drivers/overwrite_queue.go
darkdarkdragon: Added manifestID to the description

drivers/overwrite_queue.go
yondonfu: Got it. Thanks for the explanation!

drivers/overwrite_queue.go
yondonfu: Ah my mistake. I now see that we'll only get to this line when we've hit the max # of retries and the latest value for `err` is not nil.

drivers/overwrite_queue.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+// Save queues data to be saved
+func (oq *OverwriteQueue) Save(data []byte) {
+	oq.queue <- data
+}
+
+// StopAfter stops reading loop after some time
+func (oq *OverwriteQueue) StopAfter(pause time.Duration) {
+	glog.Infof("=======> stop after %s", pause)
```
yondonfu: If we're including this I think it should be only for the VERBOSE verbosity level.

drivers/overwrite_queue.go
```diff
@@ -0,0 +1,106 @@
...
+	// panic(nil)
```
yondonfu: Can remove.

drivers/session_mock.go
```diff
@@ -2,17 +2,33 @@ package drivers
...
+	waitForCh bool
+	waitCh    chan struct{}
+	back      chan struct{}
 }
 
-func (s *MockOSSession) SaveData(name string, data []byte, meta map[string]string) (string, error) {
-	args := s.Called()
+func NewMockOSSession() *MockOSSession {
+	return &MockOSSession{
+		waitCh: make(chan struct{}),
```
yondonfu, line 20: Re: the previous comment on using struct{} vs. interface{} for the channel.

I believe struct{} is typically used for signal channels like these because its the smallest available data type (at least according to https://stackoverflow.com/questions/52035390/why-using-chan-struct-when-wait-something-done-not-chan-interface/52035455). I feel like its probably not a big deal to use one or the other in most cases, but since we mainly use struct{} for signal channels elsewhere in the codebase already that's typically been my default.

darkdarkdragon: Thanks! Didn't know that `The empty struct struct {} requires no memory`

drivers/overwrite_queue.go
darkdarkdragon: removed - this one was for debugging

drivers/overwrite_queue.go
darkdarkdragon: removed

yondonfu (approved):
LGTM 🚢

PR #1937, opened 2021-06-28
cmd+server: Redact webhook basic auth info from log

### LINKED ISSUE #1936: omit auth portion of URL from webhook logs
**Describe the bug**
We're presently logging the `-authWebhookUrl` and `-detectionWebhookUrl` parameters in their entirety. Since https://github.com/livepeer/livepeer-com/commit/1f55250cdbf7c23e48849498e793701d625c87d5, we're using the basic auth functionality to authenticate webhook requests for the server. When we log those, we should omit the auth portion of the URL, or at least the password bit.

Before:
```
livepeer.go:762] Using detection webhook URL http://username:very-secret-password@staging-livepeer-api/api/stream/hook/detection
```

After
```
livepeer.go:762] Using detection webhook URL http://username:REDACTED@staging-livepeer-api/api/stream/hook/detection
```
or
```
livepeer.go:762] Using detection webhook URL http://REDACTED@staging-livepeer-api/api/stream/hook/detection
```

victorges:
Redacts basic auth credentials from webhook URLs before logging them.

This gave me second thoughts about using basic auth in the first place, but I
think it's still better than before especially after we ship the access rules for
API tokens. In the future we may want to switch to a separate flag with the
token so we can then use regular `Bearer` authorization header.

- Remove parsed basic auth user info before logging cli flag

https://play.golang.org/p/fqGr9eufH5A

Closes https://github.com/livepeer/go-livepeer/issues/1936

**Checklist:**

- [-] README and other documentation updated

### DIFF
--- a/server/mediaserver_test.go
+++ b/server/mediaserver_test.go
@@ -319,7 +319,7 @@ func TestCreateRTMPStreamHandlerCap(t *testing.T) {
 		rtmpConnections: make(map[core.ManifestID]*rtmpConnection),
 	}
 	createSid := createRTMPStreamIDHandler(s)
-	u, _ := url.Parse("http://hot/id1/secret")
+	u := mustParseUrl("http://hot/id1/secret")
 	oldMaxSessions := core.MaxSessions
 	core.MaxSessions = 1
 	// happy case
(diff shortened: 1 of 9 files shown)


yondonfu:
Seems like [url.Redacted](https://golang.org/pkg/net/url/#URL.Redacted) could be used instead of manually redacting the entire User field of the url? According to the docs, `url.Redacted` will return a string for the url with any password replaced with "xxxxx". The user portion will still be preserved though.

There is at least one other place in the codebase where DetectionWebhookUrl is logged: https://github.com/livepeer/go-livepeer/blob/887fb1c4e4b2aa84f1d890a23d9378d22bfe8d70/server/broadcast.go#L603

I have not done an exhaustive search to see if there are additional places where these URLs are logged.

If we end up using `url.Redacted`, it might make sense to store AuthWebhookUrl and DetectionWebhookUrl as `url.URL` types instead of strings so that if they need to be logged in other places to avoid an additional string to `url.URL` conversion step.

victorges:
Cool, even though redacting the user ID (which is what we're using for the user) would be nice as well, I think it will be easier to just use the existing helper on the system lib. Especially if we'll be redacting it in other places as well, to avoid repeating this `.User` replacing logic. I'll make the switch and look into the other places this may be logged.

server/mediaserver_test.go
```diff
@@ -1161,3 +1161,11 @@ func TestJsonProfileToVideoProfiles(t *testing.T) {
 	assert.Nil(p)
 	assert.Equal(common.ErrProfName, err)
 }
+
+func mustParseUrl(str string) *url.URL {
+	url, err := url.Parse(str)
+	if err != nil {
+		panic(err)
```
yondonfu: Nit: I think it would be slightly better not to panic during tests, even if this particular case should never be reached, because a panic will cause all tests, not just the current one being run, to stop. If we use [T.Fatalf](https://golang.org/pkg/testing/#T.Fatalf) from the `testing` package here instead then we can immediately mark the test that this function is run within as a failure and exit the test but allow the next tests to run.

Ex.

```golang
func mustParseUrl(t *testing.T, str string) *url.URL {
    ...
    if err != nil {
        t.Fatalf(err)
    }
}
```

yondonfu (approved):
LGTM!

Regarding Git commit history:

We typically structure commit histories in the following format in this repo (inspired by [this](https://github.com/lightningnetwork/lnd/blob/master/docs/code_contribution_guidelines.md#ideal-git-commit-structure)):

`<PACKAGE_NAMES>: <DESCRIPTION>`

So, a commit message following this format for this PR would be something like:

cmd+server: Redact auth info in webhook URL logs

Feel free to adjust prior to merge.

victorges:
Cool, thanks. Is it ok to only rename the PR and merge with squash, or should I reword every commit?

yondonfu:
> Cool, thanks. Is it ok to only rename the PR and merge with squash, or should I reword every commit?

In this case, it seems appropriate to have a single commit (as opposed to other cases where it might make sense to break things up into multiple commits and to retain that history) so updating the description in the squash commit should be fine.

PR #1938, opened 2021-06-28
cmd/livepeer: Sticky scene classification on GPU0 & better O capability

jailuthra:

#### 1. Single GPU
Scene classification is currently buggy and inconsistent when being loaded on multiple GPUs.
This is a temporary fix to make it sticky to the first available GPU (id: `0`) on the machine.

#### 2. Better O capability
Some Os, while built with the `experimental` tag, might not load the model at startup.

Ideally such Os just error out and the B picks another O - but the error returned is "Invalid argument" which we currently classify as NonRetryable, thus the B will not switch to a different O and the stream will get stuck.

To prevent this, an O should broadcast `Capability_SceneClassification` capability *only if* it loads the model during startup.

*See commit history*

Tested with 3x GTX 1070 in a testing environment.

**Checklist:**

- [] README and other documentation updated


yondonfu:
Note that the lvpdnn filter will still be [configured with whatever device ID](https://github.com/livepeer/lpms/blob/4e4a3a87d259d88ce9ddbb06f92bf9fc53ddb047/ffmpeg/ffmpeg.go#L317) the transcoding load balancer chooses to use. Could try hardcoding the device used by the filter to GPU 0 [1].

[1] https://github.com/livepeer/FFmpeg/commit/b344789a31a2d806be702239d7614f7a9f51f941#diff-765b152e21f0f875b4c8cd98268e7993e53872ca2a7f241a12acc56328055bdeR242

jailuthra:
> Note that the lvpdnn filter will still be [configured with whatever device ID](https://github.com/livepeer/lpms/blob/4e4a3a87d259d88ce9ddbb06f92bf9fc53ddb047/ffmpeg/ffmpeg.go#L317) the transcoding load balancer chooses to use. Could try hardcoding the device used by the filter to GPU 0 [1].
> 
> [1] [livepeer/FFmpeg@b344789#diff-765b152e21f0f875b4c8cd98268e7993e53872ca2a7f241a12acc56328055bdeR242](https://github.com/livepeer/FFmpeg/commit/b344789a31a2d806be702239d7614f7a9f51f941#diff-765b152e21f0f875b4c8cd98268e7993e53872ca2a7f241a12acc56328055bdeR242)

Did that in the LPMS update https://github.com/livepeer/lpms/commit/0f5e97f62f9b11aeed4ff973afa721bec51e8372 fetched in https://github.com/livepeer/go-livepeer/pull/1938/commits/43964323d098cb131af21bbf2a7dd11c5e570e2c

yondonfu (approved):
LGTM

PR #1942, opened 2021-07-01
eth,monitor: min/max gas price logs

kyriediculous:
Adds monitor statements for min and max gas price

- Added necessary logic to `monitor` package
- Emit logs from min and max gas price setters if monitoring is enabled

Partially fixes https://github.com/livepeer/livepeer-infra/issues/578

**Checklist:**

### DIFF
--- a/eth/backend.go
+++ b/eth/backend.go
@@ -16,6 +16,7 @@ import (
 	"github.com/ethereum/go-ethereum/params"
 	"github.com/golang/glog"
 	"github.com/livepeer/go-livepeer/eth/contracts"
+	"github.com/livepeer/go-livepeer/monitor"
 )
 
 var abis = []string{
@@ -129,6 +130,10 @@ func (b *backend) SetMaxGasPrice(gp *big.Int) {
 	b.Lock()
 	defer b.Unlock()
 	b.maxGasPrice = gp
+
+	if monitor.Enabled {
+		monitor.MaxGasPrice(gp)
+	}
 }
 
 func (b *backend) MaxGasPrice() *big.Int {
--- a/eth/gaspricemonitor.go
+++ b/eth/gaspricemonitor.go
@@ -46,6 +46,9 @@ func NewGasPriceMonitor(gpo GasPriceOracle, pollingInterval time.Duration, minGa
 	if minGasPrice != nil {
 		minGasP = minGasPrice
 	}
+	if monitor.Enabled {
+		monitor.MinGasPrice(minGasP)
+	}
 	return &GasPriceMonitor{
 		gpo:             gpo,
 		pollingInterval: pollingInterval,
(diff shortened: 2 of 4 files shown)


yondonfu:
@kyriediculous Looks like this PR records metrics for the min/max gas price, but livepeer/livepeer-infra#578 is related to the max *transcoding price* on the broadcaster. So, separate PR(s) will be needed to address that issue.

kyriediculous:
> @kyriediculous Looks like this PR records metrics for the min/max gas price, but [livepeer/livepeer-infra#578](https://github.com/livepeer/livepeer-infra/issues/578) is related to the max _transcoding price_ on the broadcaster. So, separate PR(s) will be needed to address that issue.

Right sorry, got things mixed up again there. Let me add that momentarily.

eth/gaspricemonitor.go
```diff
@@ -70,6 +70,10 @@ func (gpm *GasPriceMonitor) SetMinGasPrice(minGasPrice *big.Int) {
 	gpm.gasPriceMu.Lock()
 	defer gpm.gasPriceMu.Unlock()
 	gpm.minGasPrice = minGasPrice
+
+	if monitor.Enabled {
```
yondonfu, line 77: We'll need to record the min gas price in the GasPriceMonitor constructor as well since the min gas price can be initialized there. Otherwise, if the user never updates the min gas price while the node is running, there will be no min gas price recorded even if the node was started with the `-minGasPrice` flag.

eth/backend.go
```diff
@@ -129,6 +130,10 @@ func (b *backend) SetMaxGasPrice(gp *big.Int) {
 	b.Lock()
 	defer b.Unlock()
 	b.maxGasPrice = gp
+
+	if monitor.Enabled {
```
yondonfu, line 134: We'll need to record the max gas price in the backend constructor as well since the max gas price can be initialized there. Otherwise, if the user never updates the max gas price while the node is running, there will be no max gas price recorded even if the node was started with the `-maxGasPrice` flag.

yondonfu (changes requested):

eth/client.go
```diff
@@ -161,6 +162,9 @@ func NewClient(accountAddr ethcommon.Address, keystoreDir, password string, eth
 		return nil, err
 	}
 	backend.SetMaxGasPrice(maxGasPrice)
+	if monitor.Enabled && maxGasPrice != nil {
```
yondonfu: Oh I see now that the only way to set the maxGasPrice field on `backend` is via the `SetMaxGasPrice()` method right now. And we're calling that method on L164 which would record the maxGasPrice. So, it actually doesn't seem like we need this piece of code - we'd only need to record the maxGasPrice in other areas besides within the `SetMaxGasPrice()` method if the maxGasPrice field can be set in other areas.

eth/gaspricemonitor.go
```diff
@@ -45,6 +45,9 @@ func NewGasPriceMonitor(gpo GasPriceOracle, pollingInterval time.Duration, minGa
 	minGasP := big.NewInt(0)
 	if minGasPrice != nil {
 		minGasP = minGasPrice
+		if monitor.Enabled {
+			monitor.MinGasPrice(minGasPrice)
```
yondonfu: I think this should be called unconditionally regardless of the value of the `minGasPrice` argument passed in the constructor. This way if the `minGasPrice` argument is nil we would still record the current minGasPrice as 0 which is the default value.

yondonfu (changes requested):

yondonfu (approved):
LGTM

Issue #1939, opened 2021-06-29
Implement metadata queue for stream health


iameli:
### What is it?

We're going to teach go-livepeer to write segment metadata into a queue, one message (including all retries) per incoming segment. This will facilitate a "stream health" dashboard elsewhere.

We'll want a pluggable queue architecture, similar to objectStores, perhaps:
```
-metadataQueue=aqmp(s)://rabbitmqserver.example.com:5672
```

We'll start with AQMP support for RabbitMQ but later could support other queues — Kafka has been brought up once or twice.

### Queue messages
* One message per segment, including all retries. Fatal (not-retryable) errors get reported right away. Otherwise, include all retries in the message, including eventual success or failure. Here's a success case after two retries:

```json5
{
   manifestId: "ABC123",
   seqNo: 0,
   segmentDuration: 5,
   success: true,
   transcodeAttempts: [
      { orchestrator: "orchestrator.example.com", duration: 4.251, error: "OrchestratorBusy" },
      { orchestrator: "orchestrator2.example.com", duration: 4.251, error: "timeout" },
      { orchestrator: "orchestrator3.example.com", duration: 4.251, error: null }
   ]
}
```

Non-retryable error:

```json5
{
   manifestId: "ABC123",
   seqNo: 1,
   segmentDuration: 2,
   success: false,
   transcodeAttempts: [
      { orchestrator: "orchestrator.example.com", duration: 4.251, error: "no keyframes in input" }
   ]
}
```

PR #1946, opened 2021-07-06
cmd,server: Send stream health events to a metadata queue

victorges:
This is the first step for implementing a stream health dashboard. The idea is
to send events about stream transcoding to an AMQP queue (exchange). This
queue will later be processed by another (go) service which will aggregate the
information so it can be displayed to the final user.

In the order of the commits:
- Add some code on `broadcaster.go` to gather transcode attempts information
- Create a higher level RabbitMQ (producer) client `[1]` under `server/event` package `[2]`
- Configure an instance of a producer depending on CLI flags
- Actually send stream healh transcode events

`[1]` In the ideal world, we would just use an existing library that handled
all the protocol complexity, bur I couldn't find any other library with enough
developer activity to be safe enough to use.
`[2]` Is it ok to create sub-packages like that, or should it all be under `server`? Or
maybe somewhere else and have `server` depend on it?

 - Currently I've only tested that the existing tests don't break with `./test.sh`
 - Still pending to actually add some tests to this new queue producer/stream
health code, as well as actually test the whole thing running.
 - (Is it runnable locally?)

Implements https://github.com/livepeer/go-livepeer/issues/1939

**Checklist:**


cmd/livepeer/livepeer.go
```diff
@@ -140,6 +141,8 @@ func main() {
 	monitor := flag.Bool("monitor", false, "Set to true to send performance metrics")
 	version := flag.Bool("version", false, "Print out the version")
 	verbosity := flag.String("v", "", "Log verbosity.  {4|5|6}")
+	metadataAmqpUri := flag.String("metadataAmqpUri", "", "URI for AMQP host to send operation metadata")
+	metadataExchange := flag.String("metadataExchange", "lp_golivepeer_metadata", "Name of AMQP exchange to send operation metadata")
```
iameli: `lp_broadcaster_metadata` might be a little more apt?

iameli (commented):
The more I think about this the more I think that we should probably be pushing an event for every transcode attempt and failure. It'll be good for the Stream Health service to be able to evaluate how things are performing in real-time, not only after multiple retries are happening. (Though there oughta be some sort of correlation ID so it can figure out which attempts go with which successes, I suppose.)

I'll leave review of the actual code structure here to @yondonfu and @jailuthra

iameli:
It occurs to me that we'll probably want to develop this in conjunction with the service that consumes it — inevitably we'll want to make tweaks as everything comes together. So maybe we'll keep this branch open for a little bit if folks don't object.

cmd/livepeer/livepeer.go
victorges: So I'll share my full rationale here for completeness, since I forgot to include it in the PR description. My thinking was to use AMQP `topic` exchanges on our favor here. In the current format, this is what we have:
 - Exchange name is `lp_golivepeer_metadata`: 
   - `lp_` would be a prefix for us to add to every exchange populated by livepeer services, not to conflict with users' own exchanges when pointing our services to their rabbits. 
   - The next segment would be the name of the system that generates the events, in this case it felt like `golivepeer` was the more appropriate I guess
   - And then the last segment, which feels optional, specifies the kind of data published in that exchange. 
   - In this format the webhooks exchange we have would look like: `lp_api_webhooks` (but I don't really intend to change it, just for comparison)
 - The events go under a topic `${nodeType}.${whatever}`. In the case of stream health events, they will be `broadcaster.stream_health.transcode.${manifestId}`

This means that a consumer can bind their queue to the `lp_golivepeer_metadata` exchange and specify in their "subscription" the exact topics of events that they want:
 - Only stream health events from broadcasters: `broadcaster.stream_health.#`
 - All broadcasters events: `broadcaster.#`
 - All stream health events from any node type, if we ever send those from other node types: `*.stream_health.#`
 - Only the stream health events from a specific stream: `*.stream_health.*.${manifestId}`
 - Every event sent by `go-livepeer` code: `#`

That way we can easily start sending events from other node types as well without having to change any of the logic here. Some alternatives would be:
 - Have a default value of `""`  for the exchange name here, and build the default exchange name using the `nodeType` instead, and then probably remove the `nodeType` prefix from the event topics. One would still be able to subscribe to "all node types" events, but they would need to create multiple bindings for their queues, one for each node type.
 - Have a default value of `lp_broadcaster_metadata` as you suggested, and fail the command if an AMQP URI is specified but the node is not a broadcaster. Same `nodeType=>exchange` relationship as above. Would probably make sense to change flag names here as well? `broadcasterMetadata*` maybe

WDYT? Do you prefer any of those?

iameli: Hmmmm. I guess we might want to use it with Os someday too — perhaps that'd be useful to someone building a scaled public network platform atop orchestrators. Your reasoning makes sense to me, good enough for now :+1:

cmd/livepeer/livepeer.go
```diff
@@ -901,6 +904,14 @@ func main() {
 		drivers.NodeStorage = drivers.NewMemoryDriver(n.GetServiceURI())
 	}
 
+	if *metadataAmqpUri != "" {
+		uri, exchange, keyNs := *metadataAmqpUri, *metadataExchange, n.NodeType.String()
```
yondonfu: `n.NodeType.String()` is going to be something like "broadcaster", "orchestrator", etc. (since `n.NodeType.String()` is a [enum](https://github.com/livepeer/go-livepeer/blob/4e24711bfabc873f1be7e98cdd8599f68a6cfedd/core/livepeernode.go#L38)). IIUC this ends up being the `keyNs` arg when creating the AMQP producer which is a prefix for all keys used in messages. Do we want all keys to be prefixed with strings in this manner and if so, what is the motivation for this?

server/broadcast.go
```diff
@@ -473,34 +475,55 @@ func processSegment(cxn *rtmpConnection, seg *stream.HLSSegment) ([]string, erro
...
-			return nil, err
+			break
 		}
-
 		// recoverable error, retry
 	}
-	if err != nil {
+	if MetadataQueue != nil {
+		var (
+			key      = fmt.Sprintf("stream_health.transcode.%c.%s", mid[0], mid)
```
yondonfu: Why are using the first character of the manifestID followed by the full manifestID?

server/broadcast.go
```diff
@@ -473,34 +475,55 @@ func processSegment(cxn *rtmpConnection, seg *stream.HLSSegment) ([]string, erro
...
 }
 
 func transcodeSegment(cxn *rtmpConnection, seg *stream.HLSSegment, name string,
-	verifier *verification.SegmentVerifier) ([]string, error) {
+	verifier *verification.SegmentVerifier) (urls []string, info data.TranscodeAttemptInfo, err error) {
+	defer func(startTime time.Time) {
+		info.LatencyMs = time.Since(startTime).Milliseconds()
+		if err != nil {
+			errStr := err.Error()
+			info.Error = &errStr
```
yondonfu, line 530: Could we just store the error itself to preserve that information for as long as possible in case its needed later on as opposed to only storing the string representation here?

server/broadcast.go
```diff
@@ -978,6 +1005,11 @@ func nonRetryableErrMapInit() map[string]bool {
 
 var NonRetryableErrMap = nonRetryableErrMapInit()
 
-func isNonRetryableError(e error) bool {
-	return NonRetryableErrMap[e.Error()]
+func isNonRetryableError(err error) bool {
+	for e := err; e != nil; e = errors.Unwrap(e) {
```
yondonfu, line 1025: Neat - so this keeps unwrapping an error and if at any point the error is non-retryable we return true. Curious if this was needed specifically to handle a case where we actually needed to detect a non-retryable error that was wrapped?

server/metadata_queue.go
```diff
@@ -0,0 +1,23 @@
...
+
+	"github.com/golang/glog"
+	"github.com/livepeer/livepeer-data/pkg/event"
+)
+
+const queuePublishTimeout = 1 * time.Second
+
+var MetadataQueue event.Producer
+
+func BackgroundPublish(queue event.Producer, key string, body interface{}) {
```
yondonfu: Can this be a method on the producer?

server/metadata_queue.go
```diff
@@ -0,0 +1,23 @@
...
+
+import (
+	"context"
+	"time"
+
+	"github.com/golang/glog"
+	"github.com/livepeer/livepeer-data/pkg/event"
+)
+
+const queuePublishTimeout = 1 * time.Second
```
yondonfu: What do you think about accepting this as an arg for the producer and then defining the timeout in `livepeer.go`? That way the producer has a configurable publish timeout and we can just hardcode a timeout we use specifically for go-livepeer in `livepeer.go`.

cmd/livepeer/livepeer.go
```diff
@@ -142,6 +143,8 @@ func main() {
 	monitor := flag.Bool("monitor", false, "Set to true to send performance metrics")
 	version := flag.Bool("version", false, "Print out the version")
 	verbosity := flag.String("v", "", "Log verbosity.  {4|5|6}")
+	metadataAmqpUri := flag.String("metadataAmqpUri", "", "URI for AMQP host to send operation metadata")
```
yondonfu: Looks like we took a slightly different approach here vs. the description in https://github.com/livepeer/go-livepeer/issues/1939. So, instead of having a generic metadata queue flag with the prefix of the URI indicating the protocol (i.e. AMQP) to be used, we're just using a AMQP specific metadata queue flag. This feels reasonable for now given that we're just supporting a single queue implementation, but it does seem like it would be nice to follow the same convention as the `-objectStore` flag where we can parse the prefix for a URI to determine which driver/implementation to use. Can consider for the future.

yondonfu (changes requested):
Pair review from @jailuthra and I.

Would be great to add some tests especially for the bits in `server/broadcast.go` for the metadata stuff.

cmd/livepeer/livepeer.go
victorges: Yeah! I explicitly added that `String()` implementation to provide a stable namespacing for the keys. The whole logic for how I designed the AMQP topics is described in full here: https://github.com/livepeer/go-livepeer/pull/1946#discussion_r668326074

The TL;DR for this specific section is that AMQP allows for subscribing (binding) to specific exchanges under specific topics, and there are some handy "wildcards" that can be used for one or many segments in the topic (aka routing key). 

So for example, all events from the broadcaster would be like `broadcater.stream_health.transcode.<shardKey>.<streamId>`. An orchestrator could send the same event, but it'd have an `orchestrator.` prefix instead.

A consumer, on the other side, could independently subscribe to only `broadcaster` or `orchestrator` events, or if they don't care about the node type just subscribe to `*.stream_health.transcode...`. The segments in the topic serve to _allow_ consumers to be more specific in the kinds of events that they want to receive, but don't oblige them to.

So I thought the node type could be useful at some point, but if you think it's unlikely, we can also use something else or just use nothing at all as well. I wanted to get the region in there somehow but I think it's not available to the code here, and thinking further about it maybe it doesn't make much sense anyway.

server/broadcast.go
victorges: I'll make this clearer by creating a temp var before this one, specifying what it means.

Context:
We use the first character in the manifest ID as a sharding key between some of the infrastructure components, like the broadcaster specifically. So this first character is like a "shard key" that could be used by consumers to only get the events from a specific shard. That would be necessary if we also want to shard the stream health (`data/analyzer`) nodes that consume these events.

By default, which is how we are going to start, the `analyzer` nodes just subscribe to `*` on that routing key segment so there's no sharding. We already have all the paraphernalia in place to turn it on though, in which case an `analyzer` from shard like `abcd` (all streams starting with a, b, c or d) will subscribe (bind their rabbitmq stream) to the topics having each of those letters (low level: they actually create N bindings, one for each letter). This is the code snippet that does that: [analyzer/reducers/transcode.go](https://github.com/livepeer/livepeer-data/blob/007786cad78c4de5dcd0e1f2c98cff47805acfcf/health/reducers/transcode.go#L33-L40)

server/broadcast.go
victorges: The annoying thing is that the `error` is not json serializable, so I'd need to create a custom type, either a struct with all the same fields but `error`s instead of `string`s, and convert them to `string`s in the last minute, or like an `error` wrapper that holds an `error` inside but provides custom `MarshalJSON` functions to serialize them nicely (like this [`UnixNanoTime`](https://github.com/livepeer/livepeer-data/blob/9b508371591e5679d54e01d8289dd78916b12ad6/pkg/data/time.go#L8))

Do you think either of those are worth it?

server/broadcast.go
victorges: Yeah! Basically the same logic as `errors.Is` and `errors.As` from the stdlib :)

It was exactly fixing a test that broke, because of the logic change for retrying the transcode attempts. Previously, if we got a non-retriable error on the last attempt, we would just return immediately from inside the `for` loop. Now, we always finish the `for` loop so the transcode event can be sent. That cause the non-retriable events to be wrapped on a `"hit max transcode attempts: %w"` error and then this function stopped detecting them correctly.

server/metadata_queue.go
victorges: I guess that makes sense. I think I will make it a helper in the `event` lib instead which takes the producer itslef as an argument. Otherwise every `Producer interface` implementation would have to rewrite it, when it only needs the public interface to work.

server/metadata_queue.go
victorges: Actually, the producer already has a configurable timeout which can even be sent differently on every publish, which is the `context.Context` argument. This function/timeout definition here was mostly some sugar to avoid needing a go-routine+new ctx with timeout+log embedded in the business logic code. 

So I'm pro making this timeout a configurable flag, will do! But I think we don't need to move it to the producer logic since it is already "more granularly"-configurable for that. Regarding the `BackgroundPublish` sugar, I'm actually considering if I should move it to the `event` lib, or maybe just get rid of it, and do that boilerplate logic inline on the code that sends the event. I'll do one of those, but feel free to lmk any hard preference you might have.

cmd/livepeer/livepeer.go
victorges: So the main issue is that it is not possible to refer to a specific AMQP exchange or queue with a single URI. The URI can only go as far as to specify what's the server/cluster to connect and the `vhost` inside that server. Then since we needed a separate flag to specify a "queue" [1] already, I thought I should just make this amqp-specific as well. But I think we can do an intermediate solution here, have the URI be just `metadataUri` and we check if the protocol is `amqp(s)` before creating the AMQP producer. If not, we err, if yes, we create it with the other configurations provided in other flags. Same for any other protocol we give support to, which we can create specific flags for additional config if they need so, but this main URI can be kept protocol-agnostic 👌 

[1] instead of publishing events directly to a queue, I opted for an `exchange` instead which decouples `go-livepeer` from how the consumers want to consume the events. So the stream health node for example creates a stream, but some other service could create a regular queue if they want and process the events at-least-once, whatever.

server/metadata_queue.go
victorges: Decided to get rid of this helper (and file) altogether.

server/broadcast.go
```diff
@@ -38,6 +41,9 @@ var Policy *verification.Policy
 var BroadcastCfg = &BroadcastConfig{}
 var MaxAttempts = 3
 
+var MetadataQueue event.Producer
+var MetadatPublishTimeout = 1 * time.Second
```
jailuthra:
```suggestion
var MetadataPublishTimeout = 1 * time.Second
```
Will also need to change wherever it's being used.

server/broadcast.go
```diff
@@ -473,34 +479,62 @@ func processSegment(cxn *rtmpConnection, seg *stream.HLSSegment) ([]string, erro
...
 		if isNonRetryableError(err) {
 			glog.Warningf("Not retrying current segment nonce=%d seqNo=%d due to non-retryable error err=%v", nonce, seg.SeqNo, err)
-			return nil, err
+			break
 		}
-
 		// recoverable error, retry
 	}
-	if err != nil {
+	if MetadataQueue != nil {
```
jailuthra, line 507: Would be great to have a small unit test to verify the `MetadataQueue.Publish` call is being called with the data that we expect.

jailuthra (commented):
Thanks for keeping this up-to-date and explaining the queries from last review! The core changes LGTM barring a small typo.

I'm assuming the plan is to merge this after https://github.com/livepeer/livepeer-data/pull/3 is ready, right?

Also, any clue why the CI pipeline didn't run for the last update?

server/broadcast.go
victorges: Yeah, will do! Sorry haven't stop to implement the tests yet, but will do before merging (probably today)

victorges:
> I'm assuming the plan is to merge this after livepeer/livepeer-data#3 is ready, right?

Actually, we already have a RabbitMQ running in our clusters which can receive the events sent by the code here.
Then there is also a working `livepeer-data/analyzer` service running processing those in staging, but if they are
not there it will be just a no-op on RabbitMQ when publishing the event, it won't fail. There is a flag `mandatory`
which I believe asks RabbitMQ to give an error if no one is "subscribed" to the event, but we don't set that.

That #3 PR actually adds a new feature in the API of the `analyzer` service, to allow consuming the events via SSE.
But it is not required to merge or even deploy this 😁 

> Also, any clue why the CI pipeline didn't run for the last update?

No idea actually. It seems to have run after I pushed some fixes now. (tho failing for some reason, will check later)

server/broadcast.go
yondonfu: Gotcha thanks for the explanation.

I'm not sure if go-livepeer should be aware of the fact that it is a part of sharded infrastructure let alone a particular method of sharding based on a manifestID as would be the case here be exposing the notion of a manifestID based shard key. But, I don't have a better suggestion right now so okay with me since this the first pass!

server/broadcast.go
yondonfu: Got it. Keeping it as a string seems reasonable then. Although curious why that field is a [string pointer](https://github.com/livepeer/livepeer-data/blob/d24bb6e4e42d5407041f6839e6ce6d1edc0dae42/health/events.go#L54) as opposed to string given that the other string fields in structs from that package are not pointers?

cmd/livepeer/livepeer.go
```diff
@@ -142,6 +143,9 @@ func main() {
 	monitor := flag.Bool("monitor", false, "Set to true to send performance metrics")
 	version := flag.Bool("version", false, "Print out the version")
 	verbosity := flag.String("v", "", "Log verbosity.  {4|5|6}")
+	metadataQueueUri := flag.String("metadataQueueUri", "", "URI for message broker to send operation metadata")
+	metadataAmqpExchange := flag.String("metadataAmqpExchange", "lp_golivepeer_metadata", "Name of AMQP exchange to send operation metadata")
+	metadataPublishTimeout := flag.Duration("metadataPublishtimeout", 1*time.Second, "Max time to wait in background for publishing operation metadata events")
```
```suggestion
	metadataPublishTimeout := flag.Duration("metadataPublishTimeout", 1*time.Second, "Max time to wait in background for publishing operation metadata events")
```

cmd/livepeer/livepeer.go
```diff
@@ -901,6 +905,26 @@ func main() {
...
+		uri, err := url.ParseRequestURI(*metadataQueueUri)
+		if err != nil {
+			glog.Fatalf("Error parsing -metadataQueueUri: %q", err)
+		}
+		switch uri.Scheme {
+		case "amqp", "amqps":
+			uriStr, exchange, keyNs := *metadataQueueUri, *metadataAmqpExchange, n.NodeType.String()
+			server.MetadataQueue, err = event.NewAMQPExchangeProducer(context.Background(), uriStr, exchange, keyNs)
+			if err != nil {
+				glog.Fatalf("Error estabilishing AMQP connection: err=%q", err)
```
```suggestion
				glog.Fatalf("Error establishing AMQP connection: err=%q", err)
```

server/broadcast.go
victorges: So the `*string` was more of a "stylistic" choice for the JSON output object, so that the event object for a successful attempt could explicitly have an `"error": null` field. That seemed better than the `"error": ""` or no `"error"` field at all (if we used `omitempty`) in the JSON. That especially because, in the end, this object is a "user-facing" JSON that will be parsed from any language, not only `go`, like for example on the browser via SSE. So then:
 - `"error": ""`: This would fail a simple check like `error != null` on a language where strings are nullable
 - `omitempty`: I thought it could be confusing to not have any `"error"` field in most events (assuming most are successful), like a bit harder to validate the schema when building something on top of it without "trusting" that the field will be there in case an error happened. (although just `null` doesn't 100% guarantee it will be a string either, but at least the field name is there)
 
Anyway, as on the first comment, if you dislike the `*string` I can change it as well, not a strong preference here. If I create that "error wrapper" that I proposed, I could have it serialize as `null` when no `error` is there, or just make it a pointer as well which would still be less weird since it'd not be a `string`. I've kind of accustomed to using a pointer to a literal when the value is optional so I find it ok, but it might not be coherent with the `go-livepeer` style.

victorges:
@yondonfu @jailuthra Finally added the tests! Ready for another round 😁 

I see the conflicts on `go.mod/sum`, haven't rebased just so GitHub doesn't f' up with the review/commit ordering.

server/broadcast.go
victorges: Done!

server/broadcast.go
victorges: @yondonfu When you get back to this PR, just a reminder of this thread. Let me know if you'd prefer to have the `ErrorString` type on `livepeer-data/pkg/data` or if you have any other suggestions.

yondonfu: I don't dislike the use of `*string` - I mainly wanted to understand why we were using a pointer instead of a value so thanks for explaining. I'm cool with keeping the `*string` type for the error.

cmd/livepeer/livepeer.go
```diff
@@ -901,6 +905,26 @@ func main() {
 		drivers.NodeStorage = drivers.NewMemoryDriver(n.GetServiceURI())
 	}
 
+	if *metadataPublishTimeout > 0 {
+		server.MetadataPublishTimeout = *metadataPublishTimeout
+	}
+	if *metadataQueueUri != "" {
+		uri, err := url.ParseRequestURI(*metadataQueueUri)
+		if err != nil {
+			glog.Fatalf("Error parsing -metadataQueueUri: %q", err)
```
yondonfu: I see we log err= for the AMQP error, but not here. Let's do so here as well then.

We started doing the key=value logging format in a few places in the codebase, but not yet in `livepeer.go`, but I'm cool with making the transition in this file as well.

server/broadcast.go
```diff
@@ -473,34 +479,57 @@ func processSegment(cxn *rtmpConnection, seg *stream.HLSSegment) ([]string, erro
...
+		success := err == nil && len(urls) > 0
+		key, evt := newTranscodeEvent(mid, seg, startTime, success, attempts)
+		go func() {
+			ctx, cancel := context.WithTimeout(context.Background(), MetadataPublishTimeout)
+			defer cancel()
+			if err := MetadataQueue.Publish(ctx, key, evt, false); err != nil {
+				glog.Errorf("Error publishing stream transcode event: err=%q, manifestId=%q, seqNo=%d, key=%q, event=%+v", err, mid, seg.SeqNo, key, evt)
```
```suggestion
				glog.Errorf("Error publishing stream transcode event: err=%q, manifestID=%q, seqNo=%d, key=%q, event=%+v", err, mid, seg.SeqNo, key, evt)
```

Since `manifestID` is used as the key elsewhere in logging.

Also, minor note: I don't think we use a comma + space as the delimiter for the key=value pairs in logging elsewhere so for consistency we could just stick with a space as the delimiter i.e. `manifestID=%q seqNo=%d ...`.

server/broadcast_test.go
```diff
@@ -1114,6 +1127,160 @@ func TestProcessSegment_MaxAttempts(t *testing.T) {
 	assert.Len(bsm.sessMap, 0)
 }
 
+type queueEvent struct {
+	key  string
+	data interface{}
+}
+
+type ProducerChan struct {
```
yondonfu: Unless we want to make this part of the `server` package API, I suggest we make this private i.e. `producerChan`. We can always export it later if the need arises later, but I personally like the default of keeping it private unless we know we want to expose it as a part of the public API.

server/broadcast_test.go
```diff
@@ -1818,26 +1993,21 @@ func genBcastSess(t *testing.T, url string, os drivers.OSSession, mid core.Manif
...
 		Params:           &core.StreamParameters{ManifestID: mid, Profiles: []ffmpeg.VideoProfile{ffmpeg.P144p30fps16x9}, OS: os},
 		BroadcasterOS:    os,
-		OrchestratorInfo: &net.OrchestratorInfo{Transcoder: ts.URL, AuthToken: stubAuthToken},
+		OrchestratorInfo: &net.OrchestratorInfo{Transcoder: transcoderURL, AuthToken: stubAuthToken},
 	}
 }
+
+func stubTestTranscoder(ctx context.Context, handler http.HandlerFunc) string {
+	ts, mux := stubTLSServer()
+	go func() { <-ctx.Done(); ts.Close() }()
```
yondonfu, line 2010: Does waiting for ctx.Done() here fix the timing issue mentioned in the now deleted defer code block that we previously worked around by adding a hardcoded sleep?

yondonfu (commented):
Overalls looks good and thanks for adding some tests! Just a few minor comments and one clarifying question.

server/broadcast_test.go
victorges: Oh true, not sure why I made (and kept) it private

server/broadcast_test.go
victorges: Yeah! Since this function will be called with a `context.Context` that is only (and immediately) cancelled when the respective test is done, this will hold the server alive as long as the test is running i.e. while the server might be in use. 

It might take a little longer to close in case of multiple test iterations on the same function that creates separate test servers, which is the case for the metadata queue test, but I don't think we need to optimize for those (we could create and cancel separate contexts for each iteration). Closing them all after the test function is finished seems good enough to me.

yondonfu (approved):
LGTM! But dunno why GH actions are not running on this branch...hmmmm

victorges:
Yeah weird, let me try rebasing to see if it works.

Edit: Apparently it did 🤔

PR #1948, opened 2021-07-08
install_ffmpeg: Allow specifying ffmpeg installation root dir

victorges:
My local build only really worked after I ran the `./install_ffmpeg.sh`
script, even though I already had _some_ `ffmpeg` version installed
via `brew`.

The issue I had with it though was that it wrote a bunch of directories
in my home path, which felt intrusive. 

The point of this change was to allow specifying some different path
to the script, to be used as the root dir for installation instead. Simply
pass an argument to it like `./install_ffmpeg.sh ~/workspace/pkg/ffmeg`
and it should install everything to that path instead.

 - Create `$ROOT` var to use instead of `$HOME`
 - Set `$ROOT`: from cli arg, or; `$HOME`, or; `/build` (windows*)
 - `s/HOME/ROOT/g`
 - `s/~\//`$ROOT\//`

* I'll see if this works for the windows build or if we need to force a
different `$HOME` even if it exists.

Ran script both with custom and no argument and made sure it installed everything.

No.

**Checklist:**

- [-] [Pending changelog](./CHANGELOG_PENDING.md) updated

### DIFF
--- a/install_ffmpeg.sh
+++ b/install_ffmpeg.sh
@@ -101,7 +101,7 @@ if [ ! -e "$ROOT/gnutls-3.7.0" ]; then
   make install
   # gnutls doesn't properly set up its pkg-config or something? without this line ffmpeg and go
   # don't know that they need gmp, nettle, and hogweed
-  sed -i'' -e "s/-lgnutls/-lgnutls -lhogweed -lnettle -lgmp $EXTRA_GNUTLS_LIBS/g" ~/compiled/lib/pkgconfig/gnutls.pc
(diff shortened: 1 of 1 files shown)


victorges:
It later came to my mind that I could have used the script to install fmpeg on a different location by calling it like 
```
HOME=~/workspace/pkg/ffmpeg ./install_ffmpeg.sh
```

Still feels a bit weird to override the `$HOME` built-in env like that though. So I think this change would be a nice to have, but not necessary. Just let me know if it's better to keep it as is for any reason and I can drop it.

jailuthra (approved):
Really neat! LGTM for a squash & merge.

Issue #1964, opened 2021-07-27
SIGSEGV Encountered after Reward Call


GistKen:
**Describe the bug**
SIGSEGV Encountered after Reward Call

**To Reproduce**
With the automatic reward call enabled, send a reward call. Unclear what is specific about this configuration that resulted in  SIGSEGV

**Expected behavior**
Livepeer encountered a invalid memory address or nil pointer dereference during the timeout period after attempting to call a reward. The reward call occurred at 6:48:13 with the SIGSEGV at 06:53:13.
Max Gas price was set to 2500000 GWei. Using the alchemy service

**Screenshots**
```
Jul 27 06:48:13 Cosmos livepeer[20146]: ***************************************************************************
Jul 27 06:53:13 Cosmos livepeer[20146]: I0727 06:53:13.466748   20146 transactionManager.go:186]
Jul 27 06:53:13 Cosmos livepeer[20146]: ******************************Eth Transaction******************************
Jul 27 06:53:13 Cosmos livepeer[20146]: Replacement transaction: "rewardWithHint".  Gas Price: 28600000010
Jul 27 06:53:13 Cosmos livepeer[20146]: Transaction Failed: <nil>
Jul 27 06:53:13 Cosmos livepeer[20146]: ***************************************************************************
Jul 27 06:58:13 Cosmos livepeer[20146]: panic: runtime error: invalid memory address or nil pointer dereference
Jul 27 06:58:13 Cosmos livepeer[20146]: [signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x1054eed]
Jul 27 06:58:13 Cosmos livepeer[20146]: goroutine 135 [running]:
Jul 27 06:58:13 Cosmos livepeer[20146]: github.com/livepeer/go-livepeer/eth.(*client).CheckTx(0xc000d5c000, 0xc000909200, 0x0, 0x0)
Jul 27 06:58:13 Cosmos livepeer[20146]: #011/home/runner/work/go-livepeer/go-livepeer/eth/client.go:878 +0x1ed
Jul 27 06:58:13 Cosmos livepeer[20146]: github.com/livepeer/go-livepeer/eth.(*RewardService).tryReward(0xc000279770, 0xc000086e74, 0x3)
Jul 27 06:58:13 Cosmos livepeer[20146]: #011/home/runner/work/go-livepeer/go-livepeer/eth/rewardservice.go:94 +0x171
[...]

PR #1977, opened 2021-08-03
eth: Fix error logging for failed replacement tx

yondonfu:

This PR logs the correct error when a replacement tx fails at submission time.

See commit history.

N/A

Related to the logs provided in https://github.com/livepeer/go-livepeer/issues/1964.

**Checklist:**

### DIFF
--- a/eth/transactionManager_test.go
+++ b/eth/transactionManager_test.go
@@ -237,6 +237,7 @@ func TestTransactionManager_Replace(t *testing.T) {
 	logsAfter = glog.Stats.Info.Lines()
 	assert.Nil(err)
 	expTx := types.NewTransaction(1, *stubTx.To(), stubTx.Value(), 100000, calcReplacementGasPrice(stubTx), stubTx.Data())
+	assert.Equal(tx.Hash(), expTx.Hash())
 	assert.Equal(tx, expTx)
 	assert.Equal(logsAfter-logsBefore, int64(1))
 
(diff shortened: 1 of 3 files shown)


eth/transactionManager_test.go
```diff
@@ -237,6 +237,7 @@ func TestTransactionManager_Replace(t *testing.T) {
 	logsAfter = glog.Stats.Info.Lines()
 	assert.Nil(err)
 	expTx := types.NewTransaction(1, *stubTx.To(), stubTx.Value(), 100000, calcReplacementGasPrice(stubTx), stubTx.Data())
+	assert.Equal(tx.Hash(), expTx.Hash())
```
yondonfu, line 240: Fun discovery that led to the addition of this assertion as well as the assertion on L253:

The go-ethereum `Transaction` type maintains a cache for the hash of the tx and the `Hash()` method will only compute the hash of the tx if the [cache is not populated](https://github.com/ethereum/go-ethereum/blob/28ba686cbfa6a46a9b4f03b620460867d502f34e/core/types/transaction.go#L369). The internal [hash field](https://github.com/ethereum/go-ethereum/blob/28ba686cbfa6a46a9b4f03b620460867d502f34e/core/types/transaction.go#L56) of the struct is the zero value until `Hash()` is called for the first time. 

Prior to this PR, `newSignedTx.Hash()` was not called in `replace()` when an error was not returned from `SubmitTransaction()`. But, now `newSignedTx.Hash()` is called in `replace()` when an error is not returned from `SubmitTransaction()` because we call `newSignedTx.Hash()` in the log statement on L188. As a result, without these assertions that call the `Hash()` method on both `tx` and `expTx`, one of the structs would have the zero value for the internal hash field and the assertion comparing `tx` and `expTx` would fail. By first calling `Hash()` on both `tx` and `expTx` we ensure that the internal hash field of both structs is populated.

kyriediculous: Oh nice find, interesting behaviour.

kyriediculous (approved):
LGTM 👍

PR #1978, opened 2021-08-03
Replace Travis with GH action for macOS build

yondonfu:

This PR replaces the Travis macOS build with a GH action macOS build. As is the case with the Linux GH action, the macOS GH action supports caching of Go modules and ffmpeg deps.

See commit history.

Ran `livepeer -version` using the binary produced by the macOS GH action.

N/A

**Checklist:**


.github/workflows/mac.yml
```diff
@@ -0,0 +1,58 @@
+name: MacOS Build
+on:
+  pull_request:
+  push:
+    branches:
+      - master
```
victorges, line 6: Should we run this on other branches as well, to get possible errors on PRs?

.github/workflows/mac.yml
```diff
@@ -0,0 +1,58 @@
...
+    - name: Install go modules
+      if: steps.cache-go-mod.outputs.cache-hit != 'true'
+      run: go mod download
+    - name: Install ffmpeg
+      if: steps.cache-ffmpeg.outputs.cache-hit != 'true'
+      run: ./install_ffmpeg.sh
+    - name: Build binaries
+      run: |
+        export PKG_CONFIG_PATH=~/compiled/lib/pkgconfig
+        ./ci_env.sh make livepeer livepeer_cli livepeer_bench livepeer_router
```
victorges: Is there like a `make all` (or perhaps just `make`)?

victorges (approved):
LGTM!

Haven't really tested the output binary tho, do you want me to try that? Got even an M1 Mac around here if that should work too.

.github/workflows/mac.yml
yondonfu: The `pull_request` option right before `push` will allow this action to run on all PRs as well. But, the only branch where direct pushes without an associated PR will trigger this action is master.

I *think* the reason we used this configuration for the Linux action as well is because if we support other branches for `push` then a PR would trigger the action twice - once for the push and once for the PR. And the reason we wanted to specifically support triggering the action with PRs is so parts of the CI pipeline (i.e. the tests) could run if someone opens a PR from a fork of the repo.

.github/workflows/mac.yml
yondonfu: Just `make` would indeed be simpler!

There is, but previously it wouldn't build the livepeer_bench and livepeer_router binaries mentioned here.

bade226 updates the Makefile so `make` will also build livepeer_bench and livepeer_router
b337997 simplifies the GH actions to all just run `make`

yondonfu:
> Haven't really tested the output binary tho, do you want me to try that? Got even an M1 Mac around here if that should work too.

I've just verified that the binary can run via `livepeer -version`. If you have time to try doing other things with the binary that'd be great though. If anything comes up can address separately too.

yondonfu:
Oops look like the Windows build is failing due to a missing protoc dependency after b337997. Will fix...

.github/workflows/mac.yml
victorges: Didn't note the previous `pull_request` directive. This setup makes a lot of sense! Will try to do sth like that on the livepeer-data repo as well.

victorges:
> I've just verified that the binary can run via livepeer -version. If you have time to try doing other things with the binary that'd be great though. If anything comes up can address separately too.

Downloaded the file from GCS [1] and managed to run a simple `broadcaster` command which is what I'm using for testing stream health. Seems to be working fine!

[1] I went manually to the cloud console, is there an easy cli for that? Maybe a `download_build.sh` would be useful as well since the files are apparently public?

yondonfu:
> I went manually to the cloud console, is there an easy cli for that? Maybe a download_build.sh would be useful as well since the files are apparently public?

Currently, if you go to the #builds channel in Discord you can find the download links for the archives and then use curl/wget.

PR #1979, opened 2021-08-04
Upgrade to ffmpeg v4.4 and use new detector pre-loading API

### LINKED ISSUE #1962: Upgrade to ffmpeg 4.4

jailuthra:

Upgrade ffmpeg and LPMS to use v4.4, a better detector API and misc. transcoding fixes.

https://github.com/livepeer/go-livepeer/pull/1979/commits/81a4e06be0c2f7ec30702b5753c205a0bbc082c7
In this commit we switch to the new session pre-loading API introduced in https://github.com/livepeer/lpms/pull/246#issuecomment-891131495

Earlier we used to initialize FFmpeg at the time of node startup using `InitFFmpegWithDetectorProfile` - which would internally load a tensorflow model and share it across instances of our scene classification filter. We did that as it was an expensive (~3s) operation that couldn't be done when a new stream came in.

The earlier approach was problematic because we shared (a non-thread-safe) Tensorflow graph across concurrent go-routines, and it was inflexible as we could only initialize one detection model per node.

In this commit, I've mostly sticked close to the legacy idea of "pre-loading a single model" at node startup, but using the newer transcode-session API. Whenever a new stream comes in, we now initialize a new detector session, which has an impact of (~500ms) on the transcode time of the first segment.

This meets the real-time threshold on my local testing, but the plan is to improve the load balancer to have a constant `MinPreinitializedDetectorSessions` always ready at hand before a stream comes in. Doing that will reduce the additional overhead introduced by the new API, while also preventing any thread-safety issues in the Tensorflow models.

*See the commit history for rest*

- [x] Local benchmarks and B-O-T Testing
- [x] Test in staging
- [x] Test in a single production region

Fixes #1962 

**Checklist:**


core/lb.go
```diff
@@ -14,17 +14,21 @@ import (
 
 var ErrTranscoderBusy = errors.New("TranscoderBusy")
 var ErrTranscoderStopped = errors.New("TranscoderStopped")
+var DetectorProfile ffmpeg.DetectorProfile
```
yondonfu, line 20: Can we add a comment here as a reminder that this is just for convenience for now since we only support pre-loading the model for a single detector profile?

cmd/livepeer_bench/livepeer_bench.go
```diff
@@ -104,17 +104,24 @@ func main() {
 			Input:      ffmpeg.DSceneAdultSoccer.Input,
 			Output:     ffmpeg.DSceneAdultSoccer.Output,
 			Classes:    ffmpeg.DSceneAdultSoccer.Classes,
+			ModelPath:  ffmpeg.DSceneAdultSoccer.ModelPath,
 		},
 	}
+
+	ffmpeg.InitFFmpegWithLogLevel(ffmpeg.FFLogWarning)
+	var detector_tc *ffmpeg.Transcoder
```
yondonfu: Nit: Camel case is more conventional.

cmd/livepeer_bench/livepeer_bench.go
```diff
@@ -126,7 +133,23 @@ func main() {
...
+				if detector_tc != nil {
+					tc = detector_tc
+					detector_tc = nil
+				} else {
+					t := time.Now()
+					tc, err = ffmpeg.NewTranscoderWithDetector(detectionOpts.Detector, *nvidia)
+					end := time.Now()
+					fmt.Printf("InitDetectorSession time %0.4v for session %v\n", end.Sub(t).Seconds(), i)
+					if err != nil {
+						glog.Fatalf("Could not initialize detector profiles")
```
yondonfu:
```suggestion
						glog.Fatalf("Could not initialize detector")
```

cmd/livepeer_bench/livepeer_bench.go
```diff
@@ -126,7 +133,23 @@ func main() {
 	for i := 0; i < *concurrentSessions; i++ {
 		wg.Add(1)
 		go func(k int, wg *sync.WaitGroup) {
-			tc := ffmpeg.NewTranscoder()
+			var tc *ffmpeg.Transcoder
+			if *detectionFreq > 0 {
+				if detector_tc != nil {
```
yondonfu: Isn't this set earlier on L115?

yondonfu (changes requested):

cmd/livepeer_bench/livepeer_bench.go
jailuthra: We set it to `nil` a couple lines below here - so that from the 2nd session onwards we create a new session on-the-fly.

Alternatively, I think we should match it to what we do on the livepeer node - i.e. ignore the pre-initialized session altogether and init a new one each time.

Will fix!

core/lb.go
jailuthra: Will do!

cmd/livepeer_bench/livepeer_bench.go
```diff
@@ -104,17 +104,25 @@ func main() {
...
 		},
 	}
+
+	ffmpeg.InitFFmpegWithLogLevel(ffmpeg.FFLogWarning)
+	var detectorTc *ffmpeg.Transcoder
 	if *detectionFreq > 0 {
-		err = ffmpeg.InitFFmpegWithDetectorProfile(detectionOpts.Detector, *nvidia)
+		t := time.Now()
+		// We don't actually use this transcoder session, but initialize it to save time for model loading
+		detectorTc, err = ffmpeg.NewTranscoderWithDetector(detectionOpts.Detector, *nvidia)
```
jailuthra, line 116: Hardcode this to "0"
Thanks @yondonfu for catching this!

core/lb.go
```diff
@@ -80,8 +89,18 @@ func (lb *LoadBalancingTranscoder) createSession(md *SegTranscodingMetadata) (*t
 	// Acquire transcode session. Map to job id + assigned transcoder
 	key := job + "_" + transcoder
 	costEstimate := calculateCost(md.Profiles)
+	var lpmsSession TranscoderSession
+	if md.DetectorEnabled {
+		var err error
+		lpmsSession, err = lb.newDetectorT(DetectorProfile, transcoder)
```
jailuthra, line 95: Hardcode detection device to "0"
Thanks @yondonfu for catching this!

jailuthra: Made this change in https://github.com/livepeer/go-livepeer/pull/1979/commits/c17d16e113775f0957bcaefb41a65b66dd756ea6 in `core/transcoder.go` as we'd still want to schedule the transcoding on the GPU that load balancer selected - but only the detection filter hardcoded to device "0".

core/transcoder.go
```diff
@@ -145,6 +145,16 @@ func NewNvidiaTranscoder(gpu string) TranscoderSession {
 	}
 }
 
+func NewNvidiaTranscoderWithDetector(detector ffmpeg.DetectorProfile, gpu string) (TranscoderSession, error) {
+	// Hardcode detection to device 0 for now
+	// Transcoding can still run on a separate GPU as we copy frames to CPU before detection
+	session, err := ffmpeg.NewTranscoderWithDetector(detector, "0")
```
yondonfu, line 151: Just a note that can be addressed outside of this PR:

I think this API is a bit confusing. While I know that the device ID arg is only relevant for initializing the DNN filtergraph, when I first read this function my first thought is that the device ID would be used for both the transcoder and the DNN filtergraph.

install_ffmpeg.sh
```diff
@@ -137,8 +137,8 @@ if [ ! -e "$ROOT/ffmpeg/libavcodec/libavcodec.a" ]; then
     --enable-muxer=mpegts,hls,segment,mp4,null --enable-demuxer=flv,mpegts,mp4,mov \
     --enable-bsf=h264_mp4toannexb,aac_adtstoasc,h264_metadata,h264_redundant_pps,extract_extradata \
     --enable-parser=aac,aac_latm,h264 \
-    --enable-filter=abuffer,buffer,abuffersink,buffersink,afifo,fifo,aformat \
-    --enable-filter=aresample,asetnsamples,fps,scale,lvpdnn \
+    --enable-filter=abuffer,buffer,abuffersink,buffersink,afifo,fifo,aformat,format \
```
yondonfu, line 140: I might've missed this in the LPMS PR review - what is the `format` filter used for?

yondonfu (approved):
Just a few non-blocking questions, but LGTM

yondonfu (commented):
Mind also adding an entry to PENDING_CHANGELOG?

install_ffmpeg.sh
jailuthra: The `format` filter is needed after `hwdownload` filter to signal that we expect downloaded CUDA frames are in `NV12` pixel format 
https://ffmpeg.org/ffmpeg-filters.html#hwdownload-1 used [here in lpms](https://github.com/livepeer/lpms/commit/e85ff00f8df274bacc9bca2a154493800979b275#diff-4ab68a4de67768f1acb5be217d4ec78bab27b99381025beedf82c11a5fa9c7aeR310)

yondonfu: Ah right! Thanks for the explanation.

jailuthra:
Tested the last fix in a multi-GPU setup and it worked as expected. Rebased and added a PENDING_CHANGELOG entry.

Holding on a merge until the new go-livepeer release is officially out.

yondonfu:
> Tested the last fix in a multi-GPU setup and it worked as expected.

🥳 🎉 

> Holding on a merge until the new go-livepeer release is officially out.

The release was just published so good to merge after resolving the conflict 👍 .

yondonfu:
Btw I forgot to clear CHANGELOG_PENDING in https://github.com/livepeer/go-livepeer/pull/1986, but looks like the commit with the CHANGELOG_PENDING entry will clear all the old entries anyway 👍 .

PR #1988, opened 2021-08-10
Fix Windows build

yondonfu:

This PR fixed the Windows build that was previously failing because the GH action could not find the `protoc` dependency when running `make` to build the binaries. For some reason, the GH action ran successfully on master in a previous build, but just recently starting failing due to this dependency issue. So, as a temporary fix, this PR specifies the individual binaries when running `make` which side steps the `protoc` requirement which is only required when running `make` on its own.

This PR also updates `go.sum` by running `go mod download` locally with go1.16. Prior to this change, upload step would fail because the Git state would be dirty after installing the Go modules. I suspect this happened because by default `prepare_mingw64.sh` uses the current Go version supported by the mingw64 package system which is now go1.16 and using go1.16 vs. go1.15 results in some small differences in `go.sum`.

Note: The Windows build is currently using go1.16 while the Linux/Max builds are using go1.15. Separately, we should look into pinning a Go version for the Windows build so we use the same Go version across the board.

See commit history.

Ran CI.

N/A

**Checklist:**


darkdarkdragon (approved):
LGTM

PR #1989, opened 2021-08-10
monitor: Create realtime ratio histogram metric

victorges:
This will allow for simpler queries to be performed to evaluate our transcoding
performance. Using Prometheus' [histogram_quantile()](https://prometheus.io/docs/prometheus/latest/querying/functions/#histogram_quantile) function, we can actually
plot and alert on a concrete p99 value for the transcoding realtime ratio.

e.g.: "p99" (actually p01 since higher is better) of transcode real-time ratio as
seen by `ewr` orchs: [[query]](https://ewr-monitoring.livepeer.live/prometheus/graph?g0.expr=histogram_quantile(0.01%2C%20sum%20by%20(le)%20(rate(livepeer_transcode_score_bucket%5B1m%5D)))&g0.tab=0&g0.stacked=0&g0.range_input=1w)

Ended up coding this when digging deeper on the real-time alert, to make it less noisy. 

- Add `http_client_segment_transcoded_realtime_ratio` metric to `monitor/census.go`

Not really sure how to test the Prometheus logic locally, but the change is pretty simple.
Also cross-checked the implementation with the `transcode_score` metric which is almost 
the same idea (it's only measured on a different step in the code, on orchs and transcoders afaict)

Closes https://github.com/livepeer/livepeer-infra/issues/627 (still need to update the actual alert definitions though)

**Checklist:**
- [-] README and other documentation updated
- [-] [Pending changelog](./CHANGELOG_PENDING.md) updated

### DIFF
--- a/monitor/census.go
+++ b/monitor/census.go
@@ -383,7 +383,7 @@ func InitCensus(nodeType NodeType, version string) {
 		{
 			Name:        "http_client_segment_transcoded_realtime_ratio",
 			Measure:     census.mRealtimeRatio,
-			Description: "Ratio of source segment duration / transcode time as measured on broadcaster API",
+			Description: "Ratio of source segment duration / transcode time as measured on HTTP client",
 			TagKeys:     baseTags,
 			Aggregation: view.Distribution(0.5, 1, 2, 3, 5, 10, 50, 100),
 		},
--- a/monitor/census.go
+++ b/monitor/census.go
@@ -230,7 +230,7 @@ func InitCensus(nodeType NodeType, version string) {
 	}
 	census.mHTTPClientTimeout1 = stats.Int64("http_client_timeout_1", "Number of times HTTP connection was dropped before transcoding complete", "tot")
 	census.mHTTPClientTimeout2 = stats.Int64("http_client_timeout_2", "Number of times HTTP connection was dropped before transcoded segments was sent back to client", "tot")
-	census.mRealtimeRatio = stats.Float64("http_client_segment_transcoded_realtime_ratio", "Ratio of source segment duration / transcode time as measured on broadcaster API", "rat")
+	census.mRealtimeRatio = stats.Float64("http_client_segment_transcoded_realtime_ratio", "Ratio of source segment duration / transcode time as measured on HTTP client", "rat")
 	census.mRealtime3x = stats.Int64("http_client_segment_transcoded_realtime_3x", "Number of segment transcoded 3x faster than realtime", "tot")
 	census.mRealtime2x = stats.Int64("http_client_segment_transcoded_realtime_2x", "Number of segment transcoded 2x faster than realtime", "tot")
 	census.mRealtime1x = stats.Int64("http_client_segment_transcoded_realtime_1x", "Number of segment transcoded 1x faster than realtime", "tot")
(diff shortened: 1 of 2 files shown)


monitor/census.go
```diff
@@ -359,6 +361,13 @@ func InitCensus(nodeType NodeType, version string) {
 			TagKeys:     baseTags,
 			Aggregation: view.Count(),
 		},
+		{
+			Name:        "http_client_segment_transcoded_realtime_ratio",
+			Measure:     census.mRealtimeRatio,
+			Description: "Ratio of source segment duration / transcode time as measured on broadcaster API",
```
yondonfu:
```suggestion
			Description: "Ratio of source segment duration / transcode time as measured on HTTP client",
```

monitor/census.go
```diff
@@ -218,6 +219,7 @@ func InitCensus(nodeType NodeType, version string) {
 	}
 	census.mHTTPClientTimeout1 = stats.Int64("http_client_timeout_1", "Number of times HTTP connection was dropped before transcoding complete", "tot")
 	census.mHTTPClientTimeout2 = stats.Int64("http_client_timeout_2", "Number of times HTTP connection was dropped before transcoded segments was sent back to client", "tot")
+	census.mRealtimeRatio = stats.Float64("http_client_segment_transcoded_realtime_ratio", "Ratio of source segment duration / transcode time as measured on broadcaster API", "rat")
```
yondonfu:
```suggestion
	census.mRealtimeRatio = stats.Float64("http_client_segment_transcoded_realtime_ratio", "Ratio of source segment duration / transcode time as measured on HTTP client", "rat")
```

yondonfu (commented):
Looks great! Just two minor suggestions.

Mind adding an entry to CHANGELOG_PENDING since this is a user facing update? See [the docs](https://github.com/livepeer/go-livepeer/blob/master/doc/contributing.md#changelog) for more info on the entry format. I think we should add an entry under "Breaking changes" as well since I think any Prometheus/Grafana tools that depend on the individual tier metrics won't work any more since the individual tier metrics are no longer recorded?

Also, related to the last point, we'll need to update the Grafana templates in docker-livepeer for the same reason right?

victorges:
@yondonfu  Picking this back up!

> Mind adding an entry to CHANGELOG_PENDING since this is a user facing update?

Done!

>  I think we should add an entry under "Breaking changes" as well since I think any Prometheus/Grafana tools that depend on the individual tier metrics won't work any more since the individual tier metrics are no longer recorded?

Actually, we haven't stopped measuring the individual tier metrics, we've only started measuring the new histogram metric as well. Notice that [here](https://github.com/livepeer/go-livepeer/blob/e2d297538dd8302d3c7a4924da286f35c8c5100f/monitor/census.go#L1144) we send both of the `stats.Measuremenet` to the `stats.Record` function, one from the individual tier metric and one from the histogram. So there is actually no breaking change here which I think is a safer and more seamless update 🙏 

I plan on updating our realtime alert query to use the histogram, but it will continue to work the same way as before until then, no need to worry about breaking them with this change. (although they are currently disabled due to the Grafana/VictoriaMetrics upgrade, we've got to look into that)

yondonfu (approved):
@victorges Thanks for the explanation re: no breaking change!

LGTM! Looks like some of the commits can be squashed before merge.

victorges:
Thanks! Merged with squash to avoid the low-level history in `master`

PR #2001, opened 2021-08-22
eth: fix max gas price nil pointer error in replace transaction

kyriediculous:
Check that `gpm.MaxGasPrice` isn't nil in `TransctionManager.replace()`

```
[signal 0xc0000005 code=0x0 addr=0x0 pc=0x7ff626d592eb]

goroutine 44 [running]:
math/big.(*Int).Cmp(0xc00107b5a0, 0x0, 0x1)
        D:/a/_temp/msys/msys64/mingw64/lib/go/src/math/big/int.go:328 +0x4b
github.com/livepeer/go-livepeer/eth.(*TransactionManager).replace(0xc0004340f0, 0xc001760750, 0x7ff6292630c0, 0x7ff628912a01, 0x7ff6292630c0)
        D:/a/go-livepeer/go-livepeer/eth/transactionManager.go:169 +0x1e8
github.com/livepeer/go-livepeer/eth.(*TransactionManager).checkTxLoop(0xc0004340f0)
        D:/a/go-livepeer/go-livepeer/eth/transactionManager.go:222 +0x2a9
github.com/livepeer/go-livepeer/eth.(*TransactionManager).Start(0xc0004340f0)
        D:/a/go-livepeer/go-livepeer/eth/transactionManager.go:124 +0x32
created by main.main
        D:/a/go-livepeer/go-livepeer/cmd/livepeer/livepeer.go:412 +0x2a9a
```

added nil check

Removed max gas price initial value from unit tests

**Checklist:**

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -22,7 +22,7 @@
 
 #### General
 
-- \#2001  fix max gas price nil pointer error in replace transaction (@kyriediculous)
+- \#2001 Fix max gas price nil pointer error in replace transaction (@kyriediculous)
 
 #### Broadcaster
 
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -22,6 +22,8 @@
 
 #### General
 
+- \#2001  fix max gas price nil pointer error in replace transaction (@kyriediculous)
+
 #### Broadcaster
 
 #### Orchestrator
--- a/eth/transactionManager_test.go
+++ b/eth/transactionManager_test.go
@@ -219,6 +219,12 @@ func TestTransactionManager_Replace(t *testing.T) {
 	tx, err = tm.replace(stubTx)
 	assert.Nil(tx)
 	assert.EqualError(err, expErr.Error())
+
+	// Test when max gas price is nil - should still return signing replacement tx error
+	gpm.maxGasPrice = nil
+	tx, err = tm.replace(stubTx)
+	assert.Nil(tx)
+	assert.EqualError(err, expErr.Error())
 	sig.err = nil
 
 	// Error sending replacement tx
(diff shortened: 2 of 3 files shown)


CHANGELOG_PENDING.md
```diff
@@ -21,6 +21,8 @@
 
 #### General
 
+- \#2001  fix max gas price nil pointer error in replace transaction (@kyriediculous)
```
yondonfu:
```suggestion
- \#2001 Fix max gas price nil pointer error in replace transaction (@kyriediculous)
```

eth/transactionManager_test.go
```diff
@@ -209,6 +209,7 @@ func TestTransactionManager_Replace(t *testing.T) {
 	)
 
 	// Error signing replacement tx
+	gpm.maxGasPrice = nil
```
yondonfu: This line doesn't actual change the behavior of this test because the maxGasPrice is set again on L219 before the `replace()` call. So, if you revert the update in `transactionManager.go`, all tests would still pass.

I also recommend creating a separate test case specifically for this nil pointer error as opposed to modifying the state that an existing test case depends on to make it easier to identify the state that is relevant for the test case. For example:

```diff
diff --git a/eth/transactionManager_test.go b/eth/transactionManager_test.go
index aee3f34c..37c20d6c 100644
--- a/eth/transactionManager_test.go
+++ b/eth/transactionManager_test.go
@@ -209,7 +209,6 @@ func TestTransactionManager_Replace(t *testing.T) {
        )

        // Error signing replacement tx
-       gpm.maxGasPrice = nil
        expErr = errors.New("SignTx error")
        sig := &stubTransactionSigner{
                err: nil,
@@ -220,6 +219,13 @@ func TestTransactionManager_Replace(t *testing.T) {
        tx, err = tm.replace(stubTx)
        assert.Nil(tx)
        assert.EqualError(err, expErr.Error())
+
+       // Test when max gas price is nil - should still return signing replacement tx error
+       gpm.maxGasPrice = nil
+       tx, err = tm.replace(stubTx)
+       assert.Nil(tx)
+       assert.EqualError(err, expErr.Error())
+
        sig.err = nil

        // Error sending replacement tx
```

Without the update in `transactionManager.go` this new test case should cause tests to fail due to a nil pointer error and with the update all tests should pass.

yondonfu (changes requested):

yondonfu (approved):
LGTM

PR #2002, opened 2021-08-23
server: Updates to stream health events for multistream launch

victorges:
This is for the latest updates to the metadata queue logic for the production deploy this Friday.

 - Updates livepeer-data lib version to send transcode event timestamps in milliserconds.
 - Reads StreamID & SessionID fields from auth webhook to include as IDs in the events.

- `go get ... && go mod tidy`
- Add new fields to `authWebhookResponse`
- Add new fields to `StreamParameters` (pending real name to `RealStreamID` field)
- Use new `StreamID` on the transcode event 

`make`

Needed for multistream launch. 

Discussion regarding new Stream ID field: https://github.com/livepeer/livepeer-com/pull/569#discussion_r690186405

**Checklist:**


core/streamdata.go
```diff
@@ -30,6 +30,8 @@ type DetectionConfig struct {
 
 type StreamParameters struct {
 	ManifestID   ManifestID
+	RealStreamID string // lol
```
victorges: Need suggestions about how to properly name this field. We already have a `StreamID` method below for implementing the `AppData` interface, so it can't be called that. I didn't want to make just an aesthetic difference like `StreamID_` either, since they actually do not refer to the same values. Unless any of you can assure me I can simply return the `StreamID_` from the auth webhook from the `StreamID` function below.

server/broadcast.go
```diff
@@ -997,13 +1001,13 @@ func shouldRefreshSession(sess *BroadcastSession) (bool, error) {
 	return false, nil
 }
 
-func newTranscodeEvent(mid core.ManifestID, seg *stream.HLSSegment, startTime time.Time, success bool, attempts []data.TranscodeAttemptInfo) (key string, event *data.TranscodeEvent) {
+func newTranscodeEvent(mid core.ManifestID, streamID string, seg *stream.HLSSegment, startTime time.Time, success bool, attempts []data.TranscodeAttemptInfo) (key string, event *data.TranscodeEvent) {
 	var (
 		shardKey = string(mid[0])
 		segMeta  = data.SegmentMetadata{seg.Name, seg.SeqNo, seg.Duration, len(seg.Data)}
```
darkdarkdragon: linter gives warning for this, I think it's better not to use such pattern.

core/streamdata.go
```diff
@@ -29,15 +29,17 @@ type DetectionConfig struct {
...
-	Profiles     []ffmpeg.VideoProfile
-	Resolution   string
-	Format       ffmpeg.Format
-	OS           drivers.OSSession
-	RecordOS     drivers.OSSession
-	Capabilities *Capabilities
-	Detection    DetectionConfig
+	ManifestID       ManifestID
+	ExternalStreamID string
+	SessionID        string
```
darkdarkdragon, line 34: It is not being sent to RabbitMQ?

server/broadcast.go
```diff
@@ -997,13 +1001,13 @@ func shouldRefreshSession(sess *BroadcastSession) (bool, error) {
 	return false, nil
 }
 
-func newTranscodeEvent(mid core.ManifestID, seg *stream.HLSSegment, startTime time.Time, success bool, attempts []data.TranscodeAttemptInfo) (key string, event *data.TranscodeEvent) {
+func newTranscodeEvent(mid core.ManifestID, streamID string, seg *stream.HLSSegment, startTime time.Time, success bool, attempts []data.TranscodeAttemptInfo) (key string, event *data.TranscodeEvent) {
```
darkdarkdragon: @jailuthra @yondonfu @kyriediculous what do you think about using named return values? We haven't used them in before, and, for me, they make code harder to read. Personally I would prefer not to use them.

darkdarkdragon (changes requested):

server/broadcast.go
jailuthra: Makes sense to stick to a single convention throughout the repo

victorges: I particularly like them when the return type itself doesn't say much about what the returned value means, like a `string` in this case doesn't say that it's the message (event) routing key (and I don't think it's worth creating a type alias only for this).

But no strong preference here and I wasn't aware this is not used across `go-livepeer`, so I think it makes sense to avoid for consistency. Fixed.

core/streamdata.go
victorges: It's actually not. We don't need it for the `analyzer` logic and we don't have it from a couple places where we send events (`mist-api-connector` and some events on the `api`), so I didn't want to include it in the "base event fields" for now. When we decide to include it in the future, we can work on those places to have the consistent session ID all around. 

I was planning on leaving it in the `authWebhook` anyway since it doesn't hurt and will facilitate that change in the future, but let me know if you rather remove it.

jailuthra: If we're relatively sure that this field will see use within a few months, then makes sense to keep it in. Else I'm inclined to remove it right now (maybe keeping it in a branch) and re-add later.

jailuthra (approved):
LGTM :shipit:

server/broadcast.go
yondonfu: Comment on named return values (not a specific request for a change right now):

I do think that named return values can be useful in certain scenarios where the named return value can clarify what is being returned, but...I also think in many (not all!) of those scenarios the need to use a named return value may be a code smell indicating there may be an opportunity to simplify the return values for a function. For example, in this case, I wonder if a separate `newEventKey(core.ManifestID, string) string` function that returns a key would have helped since it feels like the key definition need not be tied with the creation of the `data.TranscodeEvent`.

victorges: Good point! I didn't want to create too many helpers here since that doesn't seem to be the practice in `go-livepeer` either, but if you're OK with that I do prefer that version :)
Done: https://github.com/livepeer/go-livepeer/pull/2002/commits/a7d582126121ae3ec4960a52b0f6f2d0831c6826 :feelsgood:

core/streamdata.go
victorges: I think we'll very soon want to differentiate sessions in the stream health dashboard, so I think we will be using it within a few months indeed. Keeping it then 🙆‍♂️

Issue #1993, opened 2021-08-17
Configure verification-enabled streams


jailuthra:
- Read `VerificationFreq` from auth-webhook response for a particular stream
- Read trust score for each orchestrator from orch-webhook response
- Use the capability signalling mechanism between B <> O to make sure we send verification enabled streams to Os that support generating mpeg7 signatures

Refer to steps 1, 2 and 8 in the spec #1991 for details

yondonfu:
Re: step 8 in #1991: 

> Update the list of orchestrator capabilities to include a new Capability_PerceptualSignature capability.

I suggest we name this capability `Capability_MPEG7VideoSignature` to make it clear that this capability is not just any perceptual hash/signature, but specifically the MPEG-7 video signature. I think this is important because there are different perceptual hash algorithms so we should treat the MPEG-7 video signature as just one specific perceptual hash algorithm that can be applied to videos. In the future, we may find that we want to use a different perceptual hash algorithm and then we can add a separate capability for that specific algorithm that can be distinguished from the MPEG-7 video signature capability.

> Use the capability signalling mechanism between B <> O to make sure we send verification enabled streams to Os that support generating mpeg7 signatures

I think this feature can be broken up into the following tasks (could all be worked on by the same person or in parallel by separate people):

- Update the network protobuf messages
- Update O to use the new protobuf message definitions in order to configure MPEG-7 video signature generation during transcoding
- Update B to use the new protobuf message definitions when it is requesting MPEG-7 video signatures

jailuthra:
> In the future, we may find that we want to use a different perceptual hash algorithm and then we can add a separate capability for that specific algorithm that can be distinguished from the MPEG-7 video signature capability

Ah good foresight! I'll make a quick edit to the original spec with this less-generic capability name.

> Update O to use the new protobuf message definitions in order to configure MPEG-7 video signature generation during transcoding

My original intent was to cover the details of B<>O protobuf communication in steps 4, 5 of the original spec, which would be implemented in the PR #1994 

But I don't see anything wrong with implementing it within a PR for this issue. Leaving the final decision on whoever starts working on these two issues (which could also be me).

yondonfu:
Re: step 2 in #1991

> For now this trustScore is either 0 or 1 for an untrusted or trusted O respectively. Parse this trustScore for each O into the orchestratorPool struct:

[...]

PR #2006, opened 2021-08-24
net,core: add fields for signatures in B<>O messages

### LINKED ISSUE #2003: Update protobuf messages for fast verification with mpeg7 signatures

jailuthra:

Add fields necessary for perceptual signature generation in local structs and B<>O protobuf messages.

*See commit history*

No testing

Fixes #2003 

**Checklist:**


core/orchestrator.go
```diff
@@ -390,8 +390,9 @@ type TranscodeData struct {
 
 // TranscodedSegmentData contains encoded data for a profile
 type TranscodedSegmentData struct {
-	Data   []byte
-	Pixels int64 // Encoded pixels
+	Data    []byte
+	SigData []byte // Perceptual signature data (maybe nil)
```
```suggestion
	PHash []byte // Perceptual hash data (maybe nil)
```

AFAIK "perceptual signature" isn't a standard term so if the goal here is to use general terms that capture the category of algorithms that MPEG-7 video signatures fall into then I suggest using the term "perceptual hash" as that is a standard term in industry/research.

core/streamdata.go
```diff
@@ -56,6 +56,7 @@ type SegTranscodingMetadata struct {
 	AuthToken        *net.AuthToken
 	DetectorEnabled  bool
 	DetectorProfiles []ffmpeg.DetectorProfile
+	CalcSign         bool
```
yondonfu: WDYT about enabling MPEG-7 in `ffmpeg.VideoProfile` since it would be run for each output? If we do that, I suspect that it will make it easier to construct a LPMS API call that can just calculate MPEG-7 on the input by omitting any other filters and using a null muxer which might be useful (for example, if we wanted to *just* calculate MPEG-7 for a segment - this would allow us to easily compare the video sig calculated on a rendition segment with the video sig returned).

yondonfu (changes requested):

core/streamdata.go
jailuthra: > It will make it easier to construct a LPMS API call that can just calculate MPEG-7 on the input by omitting any other filters and using a null muxer

I think we can still achieve that without moving it to `ffmpeg.VideoProfile`. The `CalcSign` field still is set per-rendition in the LPMS API via the [`ffmpeg.TranscodeOptions` struct](https://github.com/livepeer/lpms/pull/253/files#diff-4ab68a4de67768f1acb5be217d4ec78bab27b99381025beedf82c11a5fa9c7aeR74)

`Profiles []ffmpeg.VideoProfile` here gets converted into `[]ffmpeg.TranscodeOptions` in [`profilesToTranscodeOptions`](https://github.com/livepeer/go-livepeer/pull/2007/files#diff-ce19a36ebc858e8df7be76b484991146af940cc72ad2e93f9e10db553c8c7d67R217) where we'll get the choice to pick a muxer and set `CalcSign`.

yondonfu: Ah I see. At first, I was going to suggest setting the field in `ffmpeg.VideoProfile` since wherever `ffmpeg.TranscodeOptions` is used, the field can be read from the `ffmpeg.VideoProfile` field set on the `ffmpeg.TranscodeOptions`. But...a few reasons I can see for why setting the field in `ffmpeg.VideoProfile` might not be desirable:

- While we're calculating the phash per output, I don't currently see a scenario where we'd calculate the phash for one output, but not the others. So, if we had to set the field per video profile, that would be annoying when the value of the field should be the same for all video profiles
- Given that I think its preferable to avoid storing a per video profile field if it is always going to be same value for each profile, for https://github.com/livepeer/go-livepeer/issues/1993, I also think it'd be annoying having to set the field on JSON profiles just so we can convert them into `ffmpeg.VideoProfile` structs with that field set in the [helper](https://github.com/livepeer/go-livepeer/blob/92d274d3aae43c09f4106ca5f9621e3b80b101b4/server/mediaserver.go#L374)

Given the above reasoning, I think keeping the field as-is set on `SegTranscodingMetadata` and its equivalent structs (i.e. protobuf message) is reasonable.

yondonfu (approved):
LGTM

PR #2011, opened 2021-09-01
cmd/livepeer_bench: configurable delay between sessions

### LINKED ISSUE #2010: livepeer_bench: Configure sleep time between concurrent sessions
**Describe the bug**
The sleep time between two concurrent sessions launched during our benchmarking used to be `300ms` but was changed to `2300ms` while testing detection.

https://github.com/livepeer/go-livepeer/blob/92d274d3aae43c09f4106ca5f9621e3b80b101b4/cmd/livepeer_bench/livepeer_bench.go#L222

This timeout is quite large for general transcode benchmarking, as the recommended stream for benchmarking is only `60s` long. Thus with the default stream, `60/2.3 = ~26` is the max number of concurrent sessions that will be transcoding at any given time, and benchmarking with any more sessions will lead to spurious results.

**Expected behavior**
Roll back to a saner default sleep time of `300ms` between sessions, and make it configurable via a CLI parameter `-concurrentSessionDelay` for other use-cases like testing detection.

jailuthra:

See title

*See commits*

Manually tested that the default delay is `300ms` and new delay cli parameter works

Fixes #2010 

**Checklist:**


cmd/livepeer_bench/livepeer_bench.go
```diff
@@ -38,6 +38,7 @@ func main() {
 	outPrefix := flag.String("outPrefix", "", "Output segments' prefix (no segments are generated by default)")
 	detectionFreq := flag.Int("detectionFreq", 0, "Run content-detection on every nth segment. No detection occurs for default frequency of 0.")
 	detectionSampleRate := flag.Uint("detectionSampleRate", 1, "Run content-detection on every nth frame of a particular segment, if detectionFreq > 0.")
+	concurrentSessionDelay := flag.Uint("concurrentSessionDelay", 300, "Delay before starting a new concurrent session (in ms)")
```
yondonfu: What do you think about using [flag.Duration](https://pkg.go.dev/flag#Duration) here which is typically used for time based arguments which would also allow us to skip the extra conversion step on L223? The value provided would be of the form 300ms, 1.5h, 2h45m, etc. (basically anything accepted by [time.ParseDuration](https://pkg.go.dev/time#ParseDuration)).

jailuthra: Neat, fixed!

yondonfu (approved):
LGTM

Issue #1972, opened 2021-08-03
EIP-1559 Transactions


yondonfu:
# EIP-1559 Transactions

## Resources

I suggest reviewing the following resources for more context:

https://hackmd.io/@timbeiko/1559-resources
https://hackmd.io/@q8X_WM2nTfu6nuvAzqXiTQ/1559-wallets
https://hackmd.io/@timbeiko/1559-json-rpc
https://blog.alchemy.com/blog/eip-1559

## Background

Since EIP-1559 is backwards compatible, all transaction related functionality in go-livepeer will continue to work after the Ethereum London hardfork. However, it will be advantageous to add support for EIP-1559 transaction because these transactions can be more cost efficient than legacy transactions in certain scenarios since there is an opportunity to get a refund with EIP-1559 transactions equal to `maxFeePerGas - (baseFeePerGas + maxPriorityFeePerGas` (`baseFeePerGas` will depend on the block that a transaction is included in).

This post includes the following:

- Gas price estimation behavior before and after EIP-1599 for informational purposes
- Gas estimation behavior before and after EIP-1559 for informational purposes
- Replacement transaction behavior before and after EIP-1559 for informational purposes
- Updates that should be made to go-livepeer to support EIP-1559 transactions

## Gas Price Estimation

Before EIP-1559, the `eth_gasPrice` RPC method could be used to query an Ethereum node for a gas price estimation based on the node's gas price oracle implementation.

After EIP-1559, the `eth_maxPriorityFeePerGas` RPC method can be used to query an Ethereum node for a max priority fee estimation based on the node's gas price oracle implementation and the header of the latest mined block contains the base fee for the next block. 

The default behavior for Go contract bindings generated via `abigen` will be:

- If the priority fee is not explicity specified, [set the max priority fee to the result returned by `eth_maxPriorityFeePerGas`](https://github.com/ethereum/go-ethereum/blob/95bbd46eabc5d95d9fb2108ec232dd62df2f44ab/accounts/abi/bind/base.go#L263) [1].
- If the max fee is not explicitly specified, [set the max fee to the max priority fee + 2x the base fee](https://github.com/ethereum/go-ethereum/blob/95bbd46eabc5d95d9fb2108ec232dd62df2f44ab/accounts/abi/bind/base.go#L270) [2].

[...]

PR #2013, opened 2021-09-03
EIP-1559 transactions support

### LINKED ISSUE #1147: Re-generate TicketBroker binding to include named structs

At the moment, the TicketBroker contract binding [does not support named structs](https://github.com/livepeer/go-livepeer/blob/master/eth/contracts/ticketBroker.go#L176) because the version of `abigen` used to generate the binding did not support using `internalType` information from a contract's ABI and as a result could not use the name of a Solidity struct as the name for the corresponding Go struct. 

An [abigen PR](https://github.com/ethereum/go-ethereum/pull/20179) was just merged that adds support for using the `internalType` information from a contract's ABI to generate a binding that includes named structs. 

Re-generate the TicketBroker binding in `eth/contracts/ticketBroker.go` using the latest version of `abigen`. This should also allow us to remove the custom code in [client.GetSenderInfo()](https://github.com/livepeer/go-livepeer/blob/master/eth/client_ticketbroker.go#L67) - instead, we can directly call `TicketBrokerSession.GetSenderInfo()`.

### LINKED ISSUE #1973: Upgrade go-ethereum dep to v1.10.8

### LINKED ISSUE #1974: Re-generate contract bindings to support EIP-1559 transactions

### LINKED ISSUE #1975: Consider updating gas price monitor to use EIP-1559 fee fields for gas price estimation instead of eth_gasPrice

yondonfu:

This PR adds support for EIP-1559 transactions.

Under the hood, all transactions submitted by the node will use the new EIP-1559 gas fee fields. Refer to https://github.com/livepeer/go-livepeer/issues/1972 and the linked resources there for an overview of the new gas fee fields.

At the moment, we still use gas prices in the following manner:

- The node interprets the gas price as priority fee + base fee which serves as an estimation of what the user will pay to be included in the next block
- The min/max gas price flags set min/max for the priority fee + base fee the user will pay
- Later on, we could set the max gas price as the maxFeePerGas field on a transaction to control the maximum fee a user will pay since max gas price and maxFeePerGas should be semantically equivalent, but I left that out for now since we already have a max gas price check mechanism. This change would be simplification we could make later

I recommend reviewing each commit individually and sequentially.

See commit history.

Updated unit tests.

~TODO: Manual testing.~

- [x] Ran some manual tests using devtool to confirm that transactions are sent and mined properly
- [x] Currently using this branch for a node on mainnet - the [latest reward transaction](https://etherscan.io/tx/0xef0da3cee4ca925be8532568e542767d2852976a40345479abc7914d479cd3ac) was a type 2 (i.e. 1559 compliant) transaction
- [x] Test replacement tx
- [x] Test connecting to an ETH provider that doesn't support the `eth_maxPriorityFeePerGas` RPC call

Fixes #1147
Fixes #1973 
Fixes #1974 
Fixes #1975

**Checklist:**


eth/accountmanager.go
```diff
@@ -124,27 +124,24 @@ func (am *accountManager) CreateTransactOpts(gasLimit uint64) (*bind.TransactOpt
...
 	}
 
-	return &bind.TransactOpts{
-		From:     am.account.Address,
-		GasLimit: gasLimit,
-		Signer: func(signer types.Signer, address ethcommon.Address, tx *types.Transaction) (*types.Transaction, error) {
-			if address != am.account.Address {
-				return nil, errors.New("not authorized to sign this account")
-			}
+	opts, err := bind.NewKeyStoreTransactorWithChainID(am.keyStore, am.account, am.chainID)
```
yondonfu, line 127: [bind.NewKeyStoreTransactorWithChainID](https://github.com/ethereum/go-ethereum/blob/3aea432b35f0344d9f7c1f1843a4eca496229ffc/accounts/abi/bind/auth.go#L122) basically does the same thing as the removed code snippet for creating a `bind.TransactOpts`.

However, note that before we were configuring a EIP-155 signer, but now this method configures a signer using [types.LatestSignerForChainID](https://github.com/ethereum/go-ethereum/blob/c503f98f6d5e80e079c1d8a3601d188af2a899da/core/types/transaction_signing.go#L86). The signer returned by this method will still support EIP-155 replay protection, but will also support additional transaction types - see [this comment](https://github.com/ethereum/go-ethereum/blob/c503f98f6d5e80e079c1d8a3601d188af2a899da/core/types/transaction_signing.go#L175).

eth/accountmanager.go
```diff
@@ -124,27 +124,24 @@ func (am *accountManager) CreateTransactOpts(gasLimit uint64) (*bind.TransactOpt
...
-			return am.SignTx(tx)
-		},
-	}, nil
+	return opts, nil
 }
 
 // Sign a transaction. Account must be unlocked
 func (am *accountManager) SignTx(tx *types.Transaction) (*types.Transaction, error) {
-	signature, err := am.keyStore.SignHash(am.account, am.signer.Hash(tx).Bytes())
+	signer := types.LatestSignerForChainID(am.chainID)
```
yondonfu, line 138: Note that we use `types.LatestSignerForChainID` to configure a signer to match the behavior of `bind.NewKeyStoreTransactorWithChainID` which we are now using to create the `bind.TransactOpts` used for contract binding transactions.

eth/backend.go
```diff
@@ -80,11 +81,10 @@ func (b *backend) SendTransaction(ctx context.Context, tx *types.Transaction) er
 		return err
 	}
 
-	msg, err := tx.AsMessage(b.signer)
+	sender, err := types.Sender(b.signer, tx)
```
yondonfu, line 84: [types.Sender](https://github.com/ethereum/go-ethereum/blob/a541fbea182edcb0313d3b578f939ce2c7045aed/core/types/transaction_signing.go#L131) returns the address derived from the signature over the provided transaction.

eth/stubclient.go
```diff
@@ -383,5 +383,5 @@ func (c *StubClient) NextValidRequest(common.Address) (*big.Int, error) { return
 
 // Governance
 func (c *StubClient) Vote(pollAddr ethcommon.Address, choiceID *big.Int) (*types.Transaction, error) {
-	return &types.Transaction{}, c.Err
+	return types.NewTx(&types.DynamicFeeTx{}), c.Err
```
yondonfu, line 386: This is the new way of creating transaction objects - we have to specify whether we are creating a dynamic fee tx (i.e. supports EIP-1559) or a legacy tx (i.e. does not support EIP-1559).

eth/transactionManager_test.go
```diff
@@ -244,7 +244,6 @@ func TestTransactionManager_Replace(t *testing.T) {
 	assert.Nil(err)
 	expTx := types.NewTransaction(1, *stubTx.To(), stubTx.Value(), 100000, calcReplacementGasPrice(stubTx), stubTx.Data())
 	assert.Equal(tx.Hash(), expTx.Hash())
-	assert.Equal(tx, expTx)
```
yondonfu, line 247: I think these assertions for the actual transactions themselves needed to be removed because the transaction type contains a time field (used for spam avoidance use cases in the mem pool) that is [set when a transaction is created](https://github.com/ethereum/go-ethereum/blob/a541fbea182edcb0313d3b578f939ce2c7045aed/core/types/transaction.go#L197) which means that this internal field will be different for transactions created at different times.

eth/backend.go
```diff
@@ -110,6 +110,36 @@ func (b *backend) SuggestGasPrice(ctx context.Context) (*big.Int, error) {
...
 
+func (b *backend) SuggestGasTipCap(ctx context.Context) (*big.Int, error) {
+	tip, err := b.Client.SuggestGasTipCap(ctx)
+	if err != nil {
+		// SuggestGasTipCap() uses the eth_maxPriorityFeePerGas RPC call under the hood which
+		// is not a part of the ETH JSON-RPC spec.
+		// In the future, eth_maxPriorityFeePerGas can be replaced with an eth_feeHistory based algorithm (see https://github.com/ethereum/go-ethereum/issues/23479).
+		// For now, if the provider does not support eth_maxPriorityFeePerGas (i.e. not geth), then we calculate the priority fee as
+		// eth_gasPrice - baseFee.
+		if err.Error() == "Method not found" {
```
yondonfu: This logic should only be executed if the node is a) not using Infura/Alchemy [1] and b) not using geth.

The use of the fee history API would be better. This is just a temporary patch so that the node doesn't just completely fail if a) and b) mentioned above do not hold.

[1] I'm not sure if there are other hosted services to consider here, but at least these two support the eth_maxPriorityFeePerGas RPC method.

eth/transactionManager.go
```diff
@@ -150,29 +155,15 @@ func (tm *TransactionManager) replace(tx *types.Transaction) (*types.Transaction
...
-
-	suggestedGasPrice := tm.gpm.GasPrice()
-
-	// If the suggested gas price is higher than the bumped gas price, use the suggested gas price
-	// This is to account for any wild market gas price increases between the time of the original tx submission and time
-	// of replacement tx submission
-	// Note: If the suggested gas price is lower than the bumped gas price because market gas prices have dropped
-	// since the time of the original tx submission we cannot use the lower suggested gas price and we still need to use
-	// the bumped gas price in order to properly replace a still pending tx
-	if suggestedGasPrice.Cmp(gasPrice) == 1 {
```
yondonfu, line 163: Note that the node will no longer check if the bumped gas price is less than the suggested gas price. Removed to simplify a few things (avoid breaking up suggested gas price into priority fee, base fee in order to set the gas fee fields on the transaction) and because seemed like the situation is pretty rare and the current replacement behavior should largely cover it by allowing for another bump if the first is insufficient.

yondonfu:
Successful transaction replacements on Rinkeby:

```

Invoking transaction: "transfer". Inputs: "recipient: 0x81A876960E71B9d07C33338822b8BBa3a1D56Bf2  amount: 1"  Hash: "0xddfe157bfe390e8276fd3c9d4fb44a9fbfe20176fbce9bb8aa8df6473f9d1c1a".

I0929 09:53:13.279458   12574 transactionManager.go:180]

Replacement transaction: "transfer".  Hash: "0xc312ffea34fb2d5804b5348b8c3742a8efd1dc19ac52915a318de49af7417ba6".  Priority Fee: 1110000000 Max Fee: 1110000022

I0929 09:53:15.350021   12574 transactionManager.go:180]

Replacement transaction: "transfer".  Hash: "0xd6a62eb95d52caa27246b5a3aac467a23f3cbbe2d446b28f9ae0bb9514392f92".  Priority Fee: 1232100000 Max Fee: 1232100024

I0929 09:53:17.418175   12574 transactionManager.go:180]

Replacement transaction: "transfer".  Hash: "0xf38df1ab995d1d90bbad2634d406306fa8d57262c3595fc1582e7bd239d51648".  Priority Fee: 1367631000 Max Fee: 1367631026

```

Ran this test by setting `-transactionTimeout 2s -maxTransactionReplacements 3` on the node.

kyriediculous (approved):

yondonfu:
Holding off on a merge here because I discovered a few remaining issues with this PR:

- The max gas price check is broken because SuggestGasPrice will not be called for the new EIP-1559 transactions
- The conditional for executing the fallback logic in SuggestGasTipCap needs to be updated because ETH providers do not necessarily return "Method not found" if the RPC method is not supported - we can probably just execute the fallback logic if the `eth_maxPriorityFeePerGas` RPC method fails the first time
- I believe there is a bug in the go-ethereum base contract binding code that causes gas fee fields to be cached in a TransactOpts object after the first transaction is sent. Should be fixed in this [commit](https://github.com/livepeer/go-ethereum/commit/dfbe99f6f898ac5e6f18b74b1c04627f984747a1) in livepeer/go-ethereum, but still need to update this branch to use the fork

yondonfu:
> The max gas price check is broken because SuggestGasPrice will not be called for the new EIP-1559 transactions

For now, this is fixed in e73a5a8. The one downside of the change in that commit is that if the value returned by eth_gasPrice is lower than the sum of eth_maxPriorityFeePerGas + the base fee then a transaction could go through with a priority fee + base fee that exceeds the max gas price. This is possible because the max gas price check is run via `backend.SuggestGasPrice()`, but the actual priority fee to use for the transaction is fetched immediately after via `SuggestGasTipCap()`. I think the probability of that happening is fairly low since the gas fee fields for the transaction are set immediately after the max gas price check so I opted to stick with this change for simplicity sake, but this possibility is worth being aware of.

> The conditional for executing the fallback logic in SuggestGasTipCap needs to be updated because ETH providers do not necessarily return "Method not found" if the RPC method is not supported - we can probably just execute the fallback logic if the eth_maxPriorityFeePerGas RPC method fails the first time

This is fixed by 31eb840 (squashed a fixup commit in).

> I believe there is a bug in the go-ethereum base contract binding code that causes gas fee fields to be cached in a TransactOpts object after the first transaction is sent. Should be fixed in this commit in livepeer/go-ethereum, but still need to update this branch to use the fork

I opened a PR to fix this in go-ethereum [here](https://github.com/ethereum/go-ethereum/pull/23663). For now, while that PR is unmerged, I've switched the go-ethereum dependency to a [fork with the fix](https://github.com/livepeer/go-ethereum/tree/livepeer) in 63a8a2d.

PR #2018, opened 2021-09-07
pm: Only mark failed tx as redeemed for check tx err

### LINKED ISSUE #2016: Winning tickets incorrectly marked as redeemed (txHash 0x0000...)
**Describe the bug**
When receiving a winning ticket, it gets marked as redeemed and entered into the DB with txHash "0x0000000000000000000000000000000000000000000000000000000000000000"
Also reported by "_ptr | 0x10e" here: https://discord.com/channels/423160867534929930/429350403272998913/883002818234056724

**Additional context**
We both have set a `-maxGasPrice` flag on our O (both 100 gwei). Maybe this somehow messes with the automatic ticket redemption?

yondonfu:

At the moment, if an error is encountered while calling `eth.client.CheckTx()` for a ticket redemption tx, the ticket is marked as redeemed with the tx hash set to the null hash (i.e. 0x000...). This behavior is a problem because `eth.client.CheckTx()` could return an error even if the tx did not fail. For example, an error is returned in the scenario where the tx needs to be replaced, but the required gas price for a replacement exceeds the node's configured max gas price. The consequence of this behavior is that the node incorrectly marks the ticket as redeemed in the node's DB which prevents the node from retrying the ticket and also makes it more difficult to query the DB for redeemable tickets in other tools such as the script in https://github.com/livepeer/go-livepeer/pull/2014.

This PR introduces logic to only mark tickets for failed redeemed (i.e. revert, OOG) txs as redeemed.

See commit history and in-line code comments.

Ran unit tests.

Fixes #2016 

**Checklist:**

### DIFF
--- a/pm/queue.go
+++ b/pm/queue.go
@@ -2,6 +2,7 @@ package pm
 
 import (
 	"math/big"
+	"strings"
 
 	ethcommon "github.com/ethereum/go-ethereum/common"
 	"github.com/golang/glog"
@@ -128,16 +129,14 @@ ticketLoop:
 					case res := <-resCh:
 						// after receiving the response we can close the channel so it can be GC'd
 						close(resCh)
-						// If the ticket is used, we can mark it as redeemed
 						if res.err != nil {
 							glog.Errorf("Error redeeming err=%v", res.err)
-							_, checkTxErr := res.err.(errCheckTx)
-							if res.err != errIsUsedTicket && !checkTxErr {
+							// If the error is non-retryable then we mark the ticket as redeemed
+							if !isNonRetryableTicketErr(res.err) {
 								continue
 							}
 						}
-						err := q.store.MarkWinningTicketRedeemed(nextTicket, res.txHash)
-						if err != nil {
+						if err := q.store.MarkWinningTicketRedeemed(nextTicket, res.txHash); err != nil {
 							glog.Error(err)
 							continue
 						}
@@ -150,4 +149,9 @@ ticketLoop:
 			return
 		}
 	}
+}
+
+func isNonRetryableTicketErr(err error) bool {
+	// The latter check depends on logic in eth.client.CheckTx()
+	return err == errIsUsedTicket || strings.Contains(err.Error(), "transaction failed")
 }
--- a/pm/sendermonitor.go
+++ b/pm/sendermonitor.go
@@ -25,8 +25,6 @@ var unixNow = func() int64 {
 	return time.Now().Unix()
 }
 
-type errCheckTx error
-
 // SenderMonitor is an interface that describes methods used to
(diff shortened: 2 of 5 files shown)


pm/queue.go
```diff
@@ -151,3 +150,8 @@ ticketLoop:
 		}
 	}
 }
+
+func isNonRetryableTicketErr(err error) bool {
+	// The latter check depends on logic in eth.client.CheckTx()
+	return err == errIsUsedTicket || strings.Contains(err.Error(), "transaction failed")
```
yondonfu, line 156: I don't feel great about this string check, but I ran into issues implementing an error type check since the tx failure check occurs in the `eth` package and importing an error type from the `eth` package in the `pm` package would result in an import cycle. There may be a better way to structure this check, but I went with the simple solution of a string check right now accepting the downside of exposing the `pm` package to the implementation details of `eth.client.CheckTx()`.

pm/sendermonitor.go
```diff
@@ -25,8 +25,6 @@ var unixNow = func() int64 {
 	return time.Now().Unix()
 }
 
-type errCheckTx error
```
yondonfu, line 28: Removed this error type because we really care about whether the error when checking the tx is the result of a tx failure and this is accomplished by the string check in this PR making this error type unnecessary.

pm/sendermonitor.go
```diff
@@ -291,17 +289,20 @@ func (sm *LocalSenderMonitor) startTicketQueueConsumerLoop(queue *ticketQueue, d
...
-					err    error
-				}{tx.Hash(), nil}
+			res := struct {
+				txHash ethcommon.Hash
+				err    error
+			}{
+				ethcommon.Hash{},
+				err,
+			}
+			// FIXME: If there are replacement txs then tx.Hash() could be different
```
yondonfu, line 299: Note: Create an issue to track this.

This was a problem before this PR and more substantial changes will be required to address it (i.e. return the tx that was mined from `CheckTx()`). The negative impact is relatively minimal though - the DB will just end up tracking the wrong tx hash for a redeemed ticket if a replacement tx was mined.

pm/sendermonitor.go
```diff
@@ -424,7 +426,8 @@ func (sm *LocalSenderMonitor) redeemWinningTicket(ticket *SignedTicket) (*types.
 		if monitor.Enabled {
 			monitor.TicketRedemptionError()
 		}
-		return nil, errCheckTx(err)
+		// Return tx so caller can utilize the tx if it fails
+		return tx, err
```
yondonfu, line 430: This gives the caller access to the hash of a failed tx.

darkdarkdragon (approved):
LGTM

PR #2022, opened 2021-09-09
Randomization for unknown sessions at a random frequency

yondonfu:

This PR adds a randFreq field to MinLSSelector which defines a random frequency at which to randomize the selection of unknown sessions instead of running stake weighted selection when in on-chain mode.

See commit history.

Added unit tests, but only for simple scenarios as unit testing randomization is a bit tricky so my plan is to also do some manual testing with non-trivial values of randFreq i.e. not 0.0 or 1.0.

**Checklist:**


jailuthra (approved):
LGTM

PR #2025, opened 2021-09-09
-autoAdjustPrice flag to enable/disable automatic price adjustments

### LINKED ISSUE #1966: A sudden peak in gas price for few seconds result in all jobs are canceled.
when the gas price peaks for a few seconds, we lose all the jobs in progress because the price per pixel briefly exceeds the maximum allowed (1200).

This affects the continuity of service for the broadcaster.

yondonfu:

At the moment, an orchestrator can be configured with a base price (per pixel) which is then automatically adjusted based on the estimated overhead (the % representing tx cost / ticket face value) for redeeming winning tickets on-chain. As a result, the price advertised by the orchestrator is typically higher than the configured base price and it will fluctuate depending on the current estimated overhead for redeeming tickets. For example, given a base price of 1000:

- If the overhead is 1%, the advertised price is 1010
- If the overhead is 20%, the advertised price is 1200
- If the overhead is 50%, the advertised price is 1500

The motivation for the auto price adjustment mechanism is to allow orchestrators to dynamically update their price to compensate for higher overheads for ticket redemptions if gas prices on Ethereum spike. However, this mechanism can also cause a few problems:

- The orchestrator is unable to establish certainty around what price it is advertising
- If gas prices spike causing the prices to spike, the orchestrator may lose jobs because its price exceeds the max price configured by broadcasters (see #1966). An orchestrator may prefer to have the option to retain the jobs in this scenario instead and take on the risk of waiting for lower gas price periods to redeem tickets

This PR introduces a `-autoAdjustPrice` flag that can be used to enable/disable automatic price adjustments. By default, the flag is set to true which matches the current behavior of the node. If the flag is set to false, the orchestrator will always return its base price as its advertised price regardless of the current estimated overhead of redeeming tickets. For example, if the base price is 1000, regardless of the estimated overhead for ticket redemption, the advertised price would be 1000.

See commit history.

Added unit tests.

Fixes #1966.

**Checklist:**

    - [ ] Update livepeer.org/docs before merging!

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -11,6 +11,8 @@
 
 #### Orchestrator
 
+- \#2025 Add -autoAdjustPrice flag to enable/disable automatic price adjustments (@yondonfu)
+
 #### Transcoder
 
 - \#1979 Upgrade to ffmpeg v4.4 and improved API for (experimental) AI tasks (@jailuthra)
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -132,6 +132,7 @@ func main() {
 	maxPricePerUnit := flag.Int("maxPricePerUnit", 0, "The maximum transcoding price (in wei) per 'pixelsPerUnit' a broadcaster is willing to accept. If not set explicitly, broadcaster is willing to accept ANY price")
 	// Unit of pixels for both O's basePriceInfo and B's MaxBroadcastPrice
 	pixelsPerUnit := flag.Int("pixelsPerUnit", 1, "Amount of pixels per unit. Set to '> 1' to have smaller price granularity than 1 wei / pixel")
+	autoAdjustPrice := flag.Bool("autoAdjustPrice", true, "Enable/disable automatic price adjustments based on the overhead for redeeming tickets")
 	// Interval to poll for blocks
 	blockPollingInterval := flag.Int("blockPollingInterval", 5, "Interval in seconds at which different blockchain event services poll for blocks")
 	// Redemption service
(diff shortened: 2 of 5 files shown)


jailuthra (approved):
LGTM :ship: 

To clarify for my own understanding: following the example from discord https://github.com/livepeer/go-livepeer/issues/1966#issuecomment-915561648
> But I'd actually prefer for the pixel price NOT to be automatically adjusted and instead just set a maxGasPrice flag. So I could set the pixel price to 1000 wei (and a maxGasPrice of 150gwei) and it would stay like this no matter the current gas price. If the gas price is too high during the time I get the ticket, I would have to manually redeem it at a later time. That way, I could also make sure that I stay profitable when transcoding but would not lose all the sessions during gas spikes.

After this change the user can peg the price to a constant of `1000 (pixelPrice) + 150 (maxGasPrice) = 1150 wei` by setting `-pricePerUnit 1150 -autoAdjustPrice false` and redeem their tickets later when gas prices are under 150 wei; right?

yondonfu:
> After this change the user can peg the price to a constant of 1000 (pixelPrice) + 150 (maxGasPrice) = 1150 wei by setting -pricePerUnit 1150 -autoAdjustPrice false and redeem their tickets later when gas prices are under 150 wei; right?

Instead of O automatically adjusting the price to accommodate for higher fees to redeem a ticket (due to high gas prices), O can advertise a constant price and just not redeem a ticket when the fees are too high (due to high gas prices) and wait until fees are reasonable for the user to redeem.

After this change, a user could do the following:

- Set their base price to 1000 wei via `-pricePerUnit 1000`
- Set their max gas price to 150 gwei via `-maxGasPrice 150000000000`
- Disable auto price adjustments via `-autoAdjustPrice=false`
- The O would always advertise a constant price of 1000 wei instead of automatically adjusting the price up based on the current ETH gas price
- If O receives a winning ticket and the current ETH gas price exceeds 150 gwei, O will not redeem the winning ticket and instead it will regularly check if the ETH gas price falls below 150 gwei and only redeem the ticket then

jailuthra:
> Set their max gas price to 150 gwei via -maxGasPrice 150000000000

> If O receives a winning ticket and the current ETH gas price exceeds 150 gwei, O will not redeem the winning ticket and instead it will regularly check if the ETH gas price falls below 150 gwei and only redeem the ticket then

Ah I completely missed this - thanks for clarifying!

PR #2029, opened 2021-09-12
eth: Fix active total stake calc in RewardWithHint

### LINKED ISSUE #2015: Increased Gas cost when calling reward via CLI with Master build 0.5.19-e9b81b25
**Describe the bug**
When i call reward with CLI, the max gas price is way over the normal transaction

**To Reproduce**
Call reward with CLI with last master build 0.5.19-e9b81b25/livepeer-windows-amd64

https://etherscan.io/tx/0x4b3b8530564976fe5a6c20aa916a23648d970a0e3431458e7eb6d0972fbe7851

**Expected behavior**
it seems like the fields for previous and next O were not populated correctly (filled with 0 instead of the Orch addresses)
(this is most likely the reason for the increased gas cost - it needs to do additional calculations if you don't specify the position of you O. It's basically the same as calling reward without hint)

For the moment, i have reinstalled the last official Master build, i have no more fund for the moment but i ll try to call reward asap to see if the problem is fixed this way.

Also reported by Papa_Bear

yondonfu:

This PR fixes a bug in the total stake calculations used for hint generation in `eth.client.Reward()`. I've added in-line PR comments to explain the bug and the solutions used.

See commit history.

Tested manually using devtool.

Fixes #2015

**Checklist:**

### DIFF
--- a/eth/client.go
+++ b/eth/client.go
@@ -613,16 +613,17 @@ func (c *client) GetTranscoder(addr ethcommon.Address) (*lpTypes.Transcoder, err
 	}
 
 	return &lpTypes.Transcoder{
-		Address:           addr,
-		ServiceURI:        serviceURI,
-		LastRewardRound:   tInfo.LastRewardRound,
-		RewardCut:         tInfo.RewardCut,
-		FeeShare:          tInfo.FeeShare,
-		DelegatedStake:    delegatedStake,
-		ActivationRound:   tInfo.ActivationRound,
-		DeactivationRound: tInfo.DeactivationRound,
-		Active:            active,
-		Status:            status,
+		Address:                    addr,
+		ServiceURI:                 serviceURI,
+		LastRewardRound:            tInfo.LastRewardRound,
+		RewardCut:                  tInfo.RewardCut,
+		FeeShare:                   tInfo.FeeShare,
+		DelegatedStake:             delegatedStake,
+		ActivationRound:            tInfo.ActivationRound,
+		DeactivationRound:          tInfo.DeactivationRound,
+		LastActiveStakeUpdateRound: tInfo.LastActiveStakeUpdateRound,
+		Active:                     active,
+		Status:                     status,
 	}, nil
 }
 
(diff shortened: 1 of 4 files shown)


Franck-UltimaRatio:
Thanks Yondon :+1:

eth/client.go
```diff
@@ -792,7 +794,7 @@ func (c *client) Reward() (*types.Transaction, error) {
 		return nil, errors.Wrapf(err, "unable to get transcoder pool max size")
 	}
 
-	hints := simulateTranscoderPoolUpdate(addr, reward.Add(reward, ep.TotalStake), transcoders, len(transcoders) == int(maxSize.Int64()))
+	hints := simulateTranscoderPoolUpdate(addr, reward.Add(reward, tr.DelegatedStake), transcoders, len(transcoders) == int(maxSize.Int64()))
```
yondonfu, line 797: Previously, we used the `TotalStake` field on the earnings pool here. There were two problems with this:

1. If the orchestratrator's `LastActiveStakeUpdateRound` is before the current round, then the `TotalStake` field would be 0 and empty hints would be returned which results in the increased gas usage described in #2029 
2. Even if the correct earnings pool was fetched, the `TotalStake` field on the earnings pool would reflect the orchestrator's active total stake which remains fixed throughout the current round. If the orchestrator receives delegation during the round, its active total stake would not change, but its total stake *would* change and its total stake is the value that should be used for `simulateTranscoderPoolUpdate`

Both 1 and 2 are resolved by using the `DelegatedStake` field on the `Transcoder` struct which accurately reflects the orchestrator's total stake at all points in the round.

eth/client.go
```diff
@@ -753,15 +754,16 @@ func (c *client) Vote(pollAddr ethcommon.Address, choiceID *big.Int) (*types.Tra
...
 	addr := c.accountManager.Account().Address
 
-	currentRound, err := c.CurrentRound()
+	tr, err := c.GetTranscoder(addr)
 	if err != nil {
 		return nil, err
 	}
 
-	ep, err := c.GetTranscoderEarningsPoolForRound(c.accountManager.Account().Address, currentRound)
+	ep, err := c.GetTranscoderEarningsPoolForRound(addr, tr.LastActiveStakeUpdateRound)
```
yondonfu, line 762: Previously, we always fetched the earnings pool for the current round. When `LastActiveStakeUpdateRound` is in the past, the `TotalStake` field of the earnings pool for the current round would be 0 and the total stake of the orchestrator would be stored in the earnings pool for `LastActiveStakeUpdateRound`. This is resolved by always fetching the earnings pool for `LastActiveStakeUpdateRound` here.

jailuthra (approved):
LGTM

PR #2007, opened 2021-08-25
orchestrator/transcoder: calculate perceptual signatures and upload to OS

jailuthra:
**What does this pull request do? Explain your changes. (required)**

*See title*

**Specific updates (required)**

*See commit history* 

**How did you test each of these updates (required)**

Update unit tests

**Does this pull request close any open issues?**

Fixes #1995 

**Checklist:**

- [x] Read the [contribution guide](./doc/contributing.md)
- [x] `make` runs successfully
- [x] All tests in `./test.sh` pass
- [x] README and other documentation updated
- [ ] [Pending changelog](./CHANGELOG_PENDING.md) updated

PR #2036, opened 2021-09-21
orchestrator: Generate MPEG7 perceptual hashes for fast verification

### LINKED ISSUE #2035: MPEG-7 video signature capability on orchestrators
Implementation is already complete and the relevant commits just need to be cherry picked into a PR that can be merged into master.

jailuthra:

Add capability to generate mpeg7 perceptual hashes on the transcoder node, and send them back to the broadcaster with the renditions.

- https://github.com/livepeer/go-livepeer/pull/2036/commits/9c4cc8bcde7bbc52b5a1225341b5c68f3de5d71f - https://github.com/livepeer/go-livepeer/pull/2036/commits/a34b3cdb0af9b6c9e1d3323e7a4705d48669ff73 Already reviewed in #2006 and #2007
- https://github.com/livepeer/go-livepeer/pull/2036/commits/d3a79a4ebbe2c929cb3086a999b1c9dffe35ab38 Os returns `Capability_MPEG7VideoSignature` in the capability bitstring, so that Bs on the network know which O supports the new capability
- https://github.com/livepeer/go-livepeer/pull/2036/commits/ca7fc31b8b4a27f309f58450fecd7b5f3ad964c8 For split O-T setups, send the generated perceptual hashes from T to O in the multipart return, interleaved with the segments.

Locally tested that perceptual hashes are being generated in both combined and split O-T setups, if the B asks for it by setting `md.CalcPerceptualHash = true`. Also tested that the URI communicated from O to B for downloading the pHash works.

Fixes #2035 

**Checklist:**


server/ot_rpc.go
```diff
@@ -304,14 +317,20 @@ func (h *lphttp) TranscodeResults(w http.ResponseWriter, r *http.Request) {
...
+				encodedPixels, err := strconv.ParseInt(p.Header.Get("Pixels"), 10, 64)
+				if err != nil {
+					glog.Error("Error getting pixels in header:", err)
+					res.Err = err
+					break
+				}
+				segments = append(segments, &core.TranscodedSegmentData{Data: body, Pixels: encodedPixels})
+			} else if p.Header.Get("Content-Type") == "application/octet-stream" {
+				// Perceptual hash data for last segment
+				seg := *segments[len(segments)-1]
```
darkdarkdragon: Why not just
`segments[len(segments)-1].PHash = body` ? Array contains pointer, so it should work?
Also, if somehow transcoder will send wrong data, then array `segments` could be empty and thus it will crash.
Shouldn't we check array's length?

jailuthra: Fixed

darkdarkdragon (approved):
LGTM

core/orchestrator.go
```diff
@@ -581,6 +582,10 @@ func (n *LivepeerNode) transcodeSeg(config transcodeConfig, seg *stream.HLSSegme
 				string(md.ManifestID), md.AuthToken.SessionId, seg.SeqNo, len(tSegments[i].Data))
 			return terr(fmt.Errorf("ZeroSegments"))
 		}
+		if md.CalcPerceptualHash && tSegments[i].PHash == nil {
+			glog.Errorf("Could not find perceptual hash for manifestID=%s sesionID=%s seqNo=%d profile=%v",
```
```suggestion
			glog.Errorf("Could not find perceptual hash for manifestID=%s sessionID=%s seqNo=%d profile=%v",
```

core/orchestrator.go
```diff
@@ -581,6 +582,10 @@ func (n *LivepeerNode) transcodeSeg(config transcodeConfig, seg *stream.HLSSegme
 				string(md.ManifestID), md.AuthToken.SessionId, seg.SeqNo, len(tSegments[i].Data))
 			return terr(fmt.Errorf("ZeroSegments"))
 		}
+		if md.CalcPerceptualHash && tSegments[i].PHash == nil {
```
yondonfu, line 585: I think we should return an error here because if O doesn't receive a phash from T that means that a phash will be missing in the response that O sends to B and B will reject O's response. WDYT?

core/transcoder.go
```diff
@@ -189,7 +189,18 @@ func resToTranscodeData(res *ffmpeg.TranscodeResults, opts []ffmpeg.TranscodeOpt
...
+			// Extract perceptual hash if calculated
+			var s []byte = nil
+			if opts[i].CalcSign {
+				sigfile := oname + ".bin"
+				s, err = ioutil.ReadFile(sigfile)
+				if err != nil {
+					glog.Error("Cannot read perceptual hash at ", sigfile)
+					return nil, err
+				}
+				os.Remove(sigfile)
```
yondonfu: [os.Remove](https://pkg.go.dev/os#Remove) returns an error. Let's handle it? Could just log the error and continue.

core/transcoder.go
```diff
@@ -203,14 +214,15 @@ func resToTranscodeData(res *ffmpeg.TranscodeResults, opts []ffmpeg.TranscodeOpt
 	}, nil
 }
 
-func profilesToTranscodeOptions(workDir string, accel ffmpeg.Acceleration, profiles []ffmpeg.VideoProfile) []ffmpeg.TranscodeOptions {
+func profilesToTranscodeOptions(workDir string, accel ffmpeg.Acceleration, profiles []ffmpeg.VideoProfile, calcPhash bool) []ffmpeg.TranscodeOptions {
```
yondonfu:
```suggestion
func profilesToTranscodeOptions(workDir string, accel ffmpeg.Acceleration, profiles []ffmpeg.VideoProfile, calcPHash bool) []ffmpeg.TranscodeOptions {
```

to conform with the "PHash" naming convention used elsewhere?

server/ot_rpc.go
```diff
@@ -197,8 +210,15 @@ func runTranscode(n *core.LivepeerNode, orchAddr string, httpc *http.Client, not
 	if err != nil {
 		glog.Error("Error submitting results ", err)
 	} else {
-		ioutil.ReadAll(resp.Body)
+		rbody, rerr := ioutil.ReadAll(resp.Body)
 		resp.Body.Close()
+		if resp.StatusCode < 200 || resp.StatusCode >= 300 {
+			if rerr != nil {
+				glog.Errorf("Orchestrator returned HTTP %v", resp.StatusCode)
```
yondonfu: Include `rerr` in the logged message since it is non-nil here? Seems like we can do the error handling here similar to how it is done [here](https://github.com/livepeer/go-livepeer/blob/fce8a401fbd3cc24dd6341a95921bd5e63b31001/server/broadcast.go#L674) for the detection webhook response.

server/ot_rpc.go
```diff
@@ -304,14 +324,25 @@ func (h *lphttp) TranscodeResults(w http.ResponseWriter, r *http.Request) {
 				break
 			}
 
-			encodedPixels, err := strconv.ParseInt(p.Header.Get("Pixels"), 10, 64)
-			if err != nil {
-				glog.Error("Error getting pixels in header:", err)
-				res.Err = err
-				break
+			if len(p.Header.Values("Pixels")) > 0 {
```
yondonfu, line 327: I think this if/else if block will result in either the pixel count being read from the response or the phash being read from the response. Instead, we should read the pixel count from the response if it is available and the phash if it is available.

server/segment_rpc.go
```diff
@@ -192,6 +192,16 @@ func (h *lphttp) ServeSegment(w http.ResponseWriter, r *http.Request) {
 			Url:    uri,
 			Pixels: res.TranscodeData.Segments[i].Pixels,
 		}
+		// Save perceptual hash if generated
+		if res.TranscodeData.Segments[i].PHash != nil {
+			pHashFile := name + ".phash"
```
yondonfu, line 197: I'm guessing the extension doesn't matter here right? We're using `.phash` here, but we could just as well use `.bin` which ffmpeg uses?

yondonfu (changes requested):

server/segment_rpc.go
jailuthra: Yep file extensions are inconsequential for non-standard binary data. Chose `.phash` for the object store because it's more specific than the generic `.bin`.

yondonfu (approved):
LGTM after squashing :shipit:

core/orchestrator.go
```diff
@@ -581,6 +582,11 @@ func (n *LivepeerNode) transcodeSeg(config transcodeConfig, seg *stream.HLSSegme
...
+			glog.Errorf("Could not find perceptual hash for manifestID=%s sessionID=%s seqNo=%d profile=%v",
+				string(md.ManifestID), md.AuthToken.SessionId, seg.SeqNo, md.Profiles[i].Name)
+			return terr(fmt.Errorf("MissingPerceptualHash"))
```
jailuthra: Temporarily remove this, as this might break O-T split case during grace period. Will add this after 14 days grace period is over.

PR #2051, opened 2021-10-04
It/no http

### LINKED ISSUE #2040: remove http and https protocols from compiled ffmpeg
Need to check if we still passing http/https urls directly to ffmpeg C code
(need to check how data passed between O-T)
This will allow us to remove tls dependencies from cgo and will make build process easier.

darkdarkdragon:
Removes HTTP and HTTPS protocols from FFmpeg build.

Fixes #2040 

**Checklist:**

### DIFF
--- a/docker/Dockerfile.build-linux
+++ b/docker/Dockerfile.build-linux
@@ -13,7 +13,7 @@ RUN apt-get update \
   && apt-key adv --keyserver keyserver.ubuntu.com --recv 15CF4D18AF4F7421 \
   && add-apt-repository "deb [arch=amd64] http://apt.llvm.org/xenial/ llvm-toolchain-xenial-8 main" \
   && apt-get update \
-  && apt-get -y install clang-8 clang-tools-8 build-essential pkg-config autoconf gnutls-dev sudo git python docker-ce-cli
+  && apt-get -y install clang-8 clang-tools-8 build-essential pkg-config autoconf sudo git python docker-ce-cli
 
 RUN update-alternatives --install /usr/bin/clang++ clang++ /usr/bin/clang++-8 30 \
   && update-alternatives --install /usr/bin/clang clang /usr/bin/clang-8 30
--- a/install_ffmpeg.sh
+++ b/install_ffmpeg.sh
@@ -3,6 +3,13 @@
 set -ex
 
 ROOT="${1:-$HOME}"
+ARCH="$(uname -m)"
+
+if [[ $ARCH == "arm64" ]] && [[ $(uname) == "Darwin" ]]; then
+  # Detect Apple Silicon
+  IS_M1=1
+fi
+echo "Arch $ARCH ${IS_M1:+(Apple Silicon)}"
 
 # Windows (MSYS2) needs a few tweaks
 if [[ $(uname) == *"MSYS"* ]]; then
@@ -37,8 +44,7 @@ if [ $(uname) != "Darwin" ]; then
   fi
 fi
 
-# Static linking of gnutls on Linux/Mac
-if [[ $(uname) != *"MSYS"* ]]; then
+if [[ $(uname) != *"MSYS"* ]] && [[ ! $IS_M1 ]]; then
   if [ ! -e "$ROOT/nasm-2.14.02" ]; then
     # sudo apt-get -y install asciidoc xmlto # this fails :(
     cd "$ROOT"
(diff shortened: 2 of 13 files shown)


prepare_mingw64.sh
```diff
@@ -1,42 +1,16 @@
...
-    exit $EXIT
-  fi
-  pacman -U --noconfirm *.zst
-  echo "prepare_mingw64: deleting shared libraries"
-  # https://narkive.com/Fjlrbrjg:20.646.48
-  find /mingw64/lib -name "*.dll.a" -exec rm -v {} \;
-done
-# And one more dynamic library to remove, this one in the system itself...
-rm -rfv /mingw64/x86_64-w64-mingw32/lib/libwinpthread.dll.a
+exit 0
```
iameli: Was this needed for some reason?

darkdarkdragon: Yep, this was used during experiments.
Removed.

prepare_mingw64.sh
```diff
@@ -1,42 +1,14 @@
...
-    exit $EXIT
-  fi
-  pacman -U --noconfirm *.zst
-  echo "prepare_mingw64: deleting shared libraries"
-  # https://narkive.com/Fjlrbrjg:20.646.48
-  find /mingw64/lib -name "*.dll.a" -exec rm -v {} \;
-done
-# And one more dynamic library to remove, this one in the system itself...
-rm -rfv /mingw64/x86_64-w64-mingw32/lib/libwinpthread.dll.a
+echo "removing all dlls"
```
iameli, line 12: Nice. I guess I just assumed that something would break when you did this, but I guess `.dll.a` files are only needed for making new dynamic binaries, which we're not doing? Cool.

install_ffmpeg.sh
```diff
@@ -37,8 +44,7 @@ if [ $(uname) != "Darwin" ]; then
   fi
 fi
 
-# Static linking of gnutls on Linux/Mac
-if [[ $(uname) != *"MSYS"* ]]; then
+if [[ $(uname) != *"MSYS"* ]] && [[ ! $IS_M1 ]]; then
   if [ ! -e "$ROOT/nasm-2.14.02" ]; then
```
iameli, line 48: So how do we handle NASM on the M1? Homebrew? Is it only a build-time dependency?

iameli (commented):
No opinion on the Go changes, one comment on the build

prepare_mingw64.sh
darkdarkdragon: Looks like it 😄 Without that it was linking some dynamic libs, and it was not clear where it finds those `.dll.a` so I've just removed everything. And it works 😃

install_ffmpeg.sh
darkdarkdragon: > Is it only a build-time dependency?

If I understand it correctly, `nasm` only needed to build `x264`. And newest `x264` builds correctly on M1 without `nasm` installed (I suspect it uses MacOs native assembler - at least `configure` of `x264` showed that it found assembler).

install_ffmpeg.sh
```diff
@@ -3,6 +3,13 @@
 set -ex
 
 ROOT="${1:-$HOME}"
+ARCH="$(uname -m)"
+
+if [[ $ARCH == "arm64" ]] && [[ $(uname) == "Darwin" ]]; then
+  # Detect Apple Silicon
+  IS_M1=1
+fi
+echo "Arch $ARCH Apple Silicon $IS_M1"
```
victorges: WDYT of only printing the apple silicon bit when the variable is actually present? Otherwise it could get confusing when it is not an apple silicon.

```suggestion
echo "Arch $ARCH ${IS_M1:+(Apple Silicon)}"
```

victorges (approved):
No opinion about the go changes either as I don't think I have context on the code. Wasn't able to tell if it was just a refactor or a functional change, or if it is related to the HTTP(S) removal from ffmpeg.

The ffmpeg part LGTM with a small suggestion

server/ot_rpc.go
```diff
@@ -132,72 +134,112 @@ func runTranscoder(n *core.LivepeerNode, orchAddr string, capacity int) error {
...
+	if _, err := os.Stat(n.WorkDir); os.IsNotExist(err) {
+		err = os.Mkdir(n.WorkDir, 0700)
+		if err != nil {
+			glog.Errorf("Transcoder cannot create workdir err=%v", err)
+			sendTranscodeResult(n, orchAddr, httpc, notify, contentType, &body, tData, err)
+			return
+		}
+	}
+	// Create input file from segment. Removed after transcoding done
+	fname = path.Join(n.WorkDir, common.RandName()+".tempfile")
```
victorges, line 167: Should we use `ioutil.Tempfile` like stream-tester? Maybe even `Tempdir` as well

darkdarkdragon: I've just followed a way it is done in other part of the orchestrator code. 
If we think there is need to change it to `os.Tempdir` I think we should track it is in another issue.

victorges: Makes sense!

install_ffmpeg.sh
darkdarkdragon: Thanks!
I'm bad at `bash`, so I've put a lot of prints to check that it works as I've intended.

docker/Dockerfile.build-linux
```diff
@@ -13,7 +13,7 @@ RUN apt-get update \
   && apt-key adv --keyserver keyserver.ubuntu.com --recv 15CF4D18AF4F7421 \
   && add-apt-repository "deb [arch=amd64] http://apt.llvm.org/xenial/ llvm-toolchain-xenial-8 main" \
   && apt-get update \
-  && apt-get -y install clang-8 clang-tools-8 build-essential pkg-config autoconf gnutls-dev sudo git python docker-ce-cli
```
leszko, line 16: Don't you need to also update the official Livepeer doc that `gnutls` is not needed anymore?

https://livepeer.org/docs/installation/install-livepeer

server/ot_rpc.go
```diff
@@ -132,72 +134,112 @@ func runTranscoder(n *core.LivepeerNode, orchAddr string, capacity int) error {
...
 
 	glog.Infof("Transcoding taskId=%d url=%s", notify.TaskId, notify.Url)
-	var contentType string
+	var contentType, fname string
 	var body bytes.Buffer
+	var tData *core.TranscodeData
 
 	md, err := coreSegMetadata(notify.SegData)
 	if err != nil {
-		glog.Error("Unable to parse segData err=", err)
```
leszko, line 140: I see you changed some error handling. E.g. here, before the `md` struct was sent to `n.Transcoder.Transcode(md)` and now, in case of this error, we don't even call `n.Transcoder.Transcode(md)`. I guess the change is good. Just wanted to point it out.

server/ot_rpc.go
```diff
@@ -132,72 +134,112 @@ func runTranscoder(n *core.LivepeerNode, orchAddr string, capacity int) error {
...
+		err = os.Mkdir(n.WorkDir, 0700)
+		if err != nil {
+			glog.Errorf("Transcoder cannot create workdir err=%v", err)
+			sendTranscodeResult(n, orchAddr, httpc, notify, contentType, &body, tData, err)
+			return
+		}
+	}
+	// Create input file from segment. Removed after transcoding done
+	fname = path.Join(n.WorkDir, common.RandName()+".tempfile")
+	if err = ioutil.WriteFile(fname, data, 0644); err != nil {
```
leszko: Wouldn't `0600` be enough, do we need other users to access this file?

leszko (approved):
LGTM 👍 

Added some minor comments, other than that, it looks good.

prepare_mingw64.sh
```diff
@@ -1,42 +1,14 @@
 #!/bin/bash
 
 set -eo pipefail
+set -x
 
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 
 export MINGW_INSTALLS=mingw64
-pacman -S --noconfirm --noprogressbar --ask=20 --needed mingw-w64-x86_64-binutils mingw-w64-x86_64-gcc mingw-w64-x86_64-pkg-config mingw-w64-x86_64-go mingw-w64-x86_64-nasm mingw-w64-x86_64-clang git make autoconf automake patch libtool texinfo gtk-doc zip
+pacman -S --noconfirm --noprogressbar --ask=20 --needed mingw-w64-x86_64-binutils mingw-w64-x86_64-gcc mingw-w64-x86_64-pkg-config mingw-w64-x86_64-go mingw-w64-x86_64-nasm mingw-w64-x86_64-clang git make autoconf automake patch libtool texinfo gtk-doc zip zlib
```
leszko: Why do we need `zlib` now?

iameli:
Tested on Windows 10 with split O/T and NVENC enabled, looking good! No zlib required apparently.

image

server/ot_rpc.go

prepare_mingw64.sh
darkdarkdragon: Don't remember 😄 Removed.

PR #2054, opened 2021-10-12
docker: libtensorflow 2.6.0 + CUDA 11.2.2

yondonfu:

- 
- 
- 

**Checklist:**

### DIFF
--- a/docker/Dockerfile.build-linux
+++ b/docker/Dockerfile.build-linux
@@ -1,4 +1,4 @@
-FROM ubuntu:16.04
+FROM ubuntu:18.04
 
 ENV PATH "/usr/local/go/bin:/go/bin:${PATH}"
 ENV PKG_CONFIG_PATH "/root/compiled/lib/pkgconfig"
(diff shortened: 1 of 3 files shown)


docker/Dockerfile.build-linux
```diff
@@ -1,4 +1,4 @@
-FROM ubuntu:16.04
+FROM ubuntu:18.04
```
iameli, line 1: Note that this will likely cause go-livepeer to stop working on Ubuntu versions earlier than 18.04, as well as other distributions around that era, due to incompatibilities in the glibc version. I don't have strong feelings about that, just pointing out that's why this was set to 16.04 in the first place.

iameli (approved):
LGTM with a caveat

docker/Dockerfile.build-linux
yondonfu: Ah yeah thanks for flagging. We already build release binaries on 18.04 in the [Linux GH action](https://github.com/livepeer/go-livepeer/blob/e1518db8fb0350c520c6b304a1fb7046380dc3da/.github/workflows/linux.yml#L9) so I think matching with that should be ok.

iameli (changes requested):
Using this branch for servers that require it, please do not merge until we've completed upgrades 😁

thomshutt:
@iameli @yondonfu Safe to fix up the conflicts here and merge now?

PR #2056, opened 2021-10-14
broadcaster: fast verification frequency

### LINKED ISSUE #2033: Run fast verification for only certain segments and use verified orchestrator for subsequent segments

jailuthra:

- https://github.com/livepeer/go-livepeer/pull/2056/commits/59f43283b68dfd2734b9c578ea09b7026ecc9146 If the verificationFreq > 0 we should ask the Os to generate perceptual hashes, and download those hashes, regardless of how many Os are available or if we can even compare those hashes.
- https://github.com/livepeer/go-livepeer/pull/2056/commits/f2c156fee1ef59ed507a09cfc53cd644e0b98512 Once a session is verified, we use it for subsequent segments, as long as it is selected in the 3 sessions returned from the pools.
- https://github.com/livepeer/go-livepeer/pull/2056/commits/25373bda1fcd603b77892ba8ba2727a2b99ab684 Instead of sticking to a verified orchestrator throughout the stream, we run verification probabilistically on 1/VerificationFreq segments.
- https://github.com/livepeer/go-livepeer/pull/2056/commits/a73a85ec5c3da1ef95bb501554bb645113c83d97 If fast verification is disabled, default to selecting from untrusted pool first, only switching to trusted pool if no untrusted Os are available.

Local end-to-end testing as described in https://github.com/livepeer/internal-project-tracking/issues/183

Fixes #2033 

**Checklist:**

### DIFF
--- a/common/util.go
+++ b/common/util.go
@@ -372,6 +372,10 @@ func RandName() string {
 	return RandomIDGenerator(10)
 }
 
+var RandomUintUnder = func(max uint) uint {
+	return uint(rand.Uint32()) % max
+}
+
 func ToInt64(val *big.Int) int64 {
 	if val.Cmp(big.NewInt(maxInt64)) > 0 {
 		return maxInt64
--- a/server/broadcast.go
+++ b/server/broadcast.go
@@ -419,7 +419,9 @@ func (bsm *BroadcastSessionsManager) selectSessions() ([]*BroadcastSession, bool
 
 		// Only return the last verified session if:
 		// - It is present in the 3 sessions returned by the selector
-		if bsm.verifiedSession != nil && includesSession(sessions, bsm.verifiedSession) {
+		// - With probability 1 - 1/VerificationFrequency
+		if bsm.verifiedSession != nil && includesSession(sessions, bsm.verifiedSession) &&
+			common.RandomUintUnder(bsm.VerificationFreq) > 0 {
 			glog.V(common.DEBUG).Infof("Reusing verified orch=%v", bsm.verifiedSession.OrchestratorInfo.Transcoder)
 			// Mark remaining unused sessions returned by selector as complete
 			remaining := removeSessionFromList(sessions, bsm.verifiedSession)
--- a/server/broadcast.go
+++ b/server/broadcast.go
@@ -435,9 +435,10 @@ func (bsm *BroadcastSessionsManager) selectSessions() ([]*BroadcastSession, bool
 		return sessions, true
 	}
 
-	sessions := bsm.trustedPool.selectSessions(1)
+	// Default to selecting from untrusted pool
+	sessions := bsm.untrustedPool.selectSessions(1)
 	if len(sessions) == 0 {
-		sessions = bsm.untrustedPool.selectSessions(1)
+		sessions = bsm.trustedPool.selectSessions(1)
 	}
 
 	return sessions, false
--- a/server/broadcast.go
+++ b/server/broadcast.go
@@ -343,6 +343,8 @@ type BroadcastSessionsManager struct {
 
 	trustedPool   *SessionPool
 	untrustedPool *SessionPool
+
+	verifiedSession *BroadcastSession
 }
 
 func NewSessionManager(node *core.LivepeerNode, params *core.StreamParameters, sel BroadcastSessionsSelectorFactory) *BroadcastSessionsManager {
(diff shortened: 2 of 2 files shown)


server/broadcast.go
```diff
@@ -367,6 +369,7 @@ func NewSessionManager(node *core.LivepeerNode, params *core.StreamParameters, s
 		VerificationFreq: params.VerificationFreq,
 		trustedPool:      NewSessionPool(params.ManifestID, int(trustedPoolSize), trustedNumOrchs, susTrusted, createSessionsTrusted, sel()),
 		untrustedPool:    NewSessionPool(params.ManifestID, int(untrustedPoolSize), untrustedNumOrchs, susUntrusted, createSessionsUntrusted, sel()),
+		verifiedSession:  nil,
```
yondonfu: Can be omitted because `verifiedSession` will be default to nil.

server/broadcast.go
```diff
@@ -410,11 +413,27 @@ func (bsm *BroadcastSessionsManager) selectSessions() ([]*BroadcastSession, bool
...
+		// Select 1 trusted O and 2 untrusted Os
 		sessions := bsm.trustedPool.selectSessions(1)
-		untrustedSesions := bsm.untrustedPool.selectSessions(2)
-		calcPerceptualHash := true
-		return append(sessions, untrustedSesions...), calcPerceptualHash
+		untrustedSessions := bsm.untrustedPool.selectSessions(2)
+		sessions = append(sessions, untrustedSessions...)
+
+		// Only return the last verified session if:
+		// - It is present in the 3 sessions returned by the selector
```
yondonfu, line 421: What's the rationale for not just always returning the verified session if one exists?

Could the following be a simplification to the logic here?

```
// Always return multiple sessions if fast verification should be run
if common.RandomUintUnder(bsm.VerificationFreq) < 0 {
    sessions := bsm.trustedPool.selectSessions(1)
    untrustedSessions := bsm.untrustedPool.selectSessions(2)
    return append(sessions, untrustedSessions...), true
}

// If fast verification is enabled and fast verification does not need to be run return the verified session
if bsm.VerificationFreq > 0 && bsm.verifiedSession != nil {
    return bsm.verifiedSession, true
}

// Default to selecting from untrusted pool
sessions := bsm.untrustedPool.selectSession(1)
if len(sessions) <= 0 {
    sessions = bsm.trustedPool.selectSessions(1)
}

return sessions, false
```

server/broadcast.go
```diff
@@ -410,11 +413,29 @@ func (bsm *BroadcastSessionsManager) selectSessions() ([]*BroadcastSession, bool
...
+		// - With probability 1 - 1/VerificationFrequency
+		if bsm.verifiedSession != nil && includesSession(sessions, bsm.verifiedSession) &&
+			common.RandomUintUnder(bsm.VerificationFreq) > 0 {
+			glog.V(common.DEBUG).Infof("Reusing verified orchestrator sess=%v", bsm.verifiedSession.OrchestratorInfo.Transcoder)
```
```suggestion
			glog.V(common.DEBUG).Infof("Reusing verified orch=%v", bsm.verifiedSession.OrchestratorInfo.Transcoder)
```

common/util.go
```diff
@@ -372,6 +372,10 @@ func RandName() string {
 	return RandomIDGenerator(10)
 }
 
+var RandomUintUnder = func(max uint) uint {
```
yondonfu, line 375: What's the rationale for defining this as a package var? Is the intent to stub this out in unit tests?

yondonfu (changes requested):

server/broadcast.go
jailuthra: I did something similar initially, but if the `verifiedSession` went offline while we were solely relying on it, it lead to transcoding requests for multiple segments timing out. Things only recovered after a periodic refresh or the probabilistic draw for verification.

A possible way to fix it was to manually set `verifiedSession = nil` if we remove and suspend an orchestrator `sess == verifiedSession` due to a failed or timed out transcode. This would lead to a fresh set of `selectSession` calls immediately for the next segment.

But ultimately I ended up with the current solution, to not scatter the logic of setting/unsetting `verifiedSession` at multiple places.

By calling `pool.selectSessions(int)` for every segment, and verifying that our `verifiedSession` is in there, we ensure that it's alive and wasn't suspended for last segment. And the pool selectors anyway try to stick to `lastSessions` so this doesn't lead to a lot of switches.

```
// Default to selecting from untrusted pool
sessions := bsm.untrustedPool.selectSession(1)
if len(sessions) <= 0 {
    sessions = bsm.trustedPool.selectSessions(1)
}
```

^ Will fix this though, thanks!

common/util.go
jailuthra: Yes that was the initial rationale, although still a TODO to update the unit test for the selection logic. Will fix.

server/broadcast.go
```diff
@@ -410,16 +412,35 @@ func (bsm *BroadcastSessionsManager) selectSessions() ([]*BroadcastSession, bool
...
+		// - It is present in the 3 sessions returned by the selector
+		// - With probability 1 - 1/VerificationFrequency
+		if bsm.verifiedSession != nil && includesSession(sessions, bsm.verifiedSession) &&
+			common.RandomUintUnder(bsm.VerificationFreq) > 0 {
+			glog.V(common.DEBUG).Infof("Reusing verified orch=%v", bsm.verifiedSession.OrchestratorInfo.Transcoder)
+			// Mark remaining unused sessions returned by selector as complete
+			remaining := removeSessionFromList(sessions, bsm.verifiedSession)
+			for _, sess := range remaining {
+				bsm.sessLock.Unlock()
+				bsm.completeSession(sess)
```
darkdarkdragon: Can we create `completeSessionLocked` function which doesn't do locking inside?

darkdarkdragon (changes requested):
Small comment, otherwise looks good.

server/broadcast.go
```diff
@@ -410,16 +412,35 @@ func (bsm *BroadcastSessionsManager) selectSessions() ([]*BroadcastSession, bool
...
+				bsm.sessLock.Lock()
```
yondonfu: Is this a problem if we lock here after the loop completes? Could there be another call to a function that needs to call `bsm.sessLock.Lock()` after this loop completes that will block since `sessLock` is left in a locked state?

jailuthra: `sessLock` should get unlocked a few lines below when we return from this function, due to the `defer bsm.sessLock.Unlock()` at the start of the function.

Regardless, I like Ivan's idea above for having a `completeSession` method that does not do locking inside, to prevent this weird looking call. Will fix.

server/broadcast.go
jailuthra: Fixed

yondonfu (approved):
LGTM

PR #2057, opened 2021-10-15
Update LPMS with `panic` for `CUDA_ERROR_ILLEGAL_ADDRESS` stuck sessions

### LINKED ISSUE #1921: Crash when we get CUDA_ERROR_ILLEGAL_ADDRESS
Related: https://github.com/livepeer/lpms/issues/239

My understanding of the latest `CUDA_ERROR_ILLEGAL_ADDRESS` errors we're seeing: CUDA itself has gotten into some sort of bad state and any further CUDA calls will fail. From [Nvidia docs](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__TYPES.html) (thanks @jailuthra):

> CUDA_ERROR_ILLEGAL_ADDRESS = 700
>
> While executing a kernel, the device encountered a load or store instruction on an invalid memory address. This leaves the process in an inconsistent state and any further CUDA work will return the same error. To continue using CUDA, the process must be terminated and relaunched. 

So in the event we hit `CUDA_ERROR_ILLEGAL_ADDRESS`, we should just `panic()` and let the local environment restart the node. No point in continuing to run if we can't transcode anymore.

jailuthra:

See https://github.com/livepeer/lpms/pull/267

*See commits* 
 
 

Fixes #1921

**Checklist:**

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -25,3 +25,5 @@
 #### Orchestrator
 
 #### Transcoder
+
+- \#2057 Prevent stuck sessions by crashing on unrecoverable CUDA errors (@jailuthra)
--- a/core/transcoder_test.go
+++ b/core/transcoder_test.go
@@ -40,43 +40,44 @@ func TestLocalTranscoder(t *testing.T) {
 }
 
 func TestNvidia_Transcoder(t *testing.T) {
-	tmp, _ := ioutil.TempDir("", "")
-	defer os.RemoveAll(tmp)
-	WorkDir = tmp
-	defer func() { WorkDir = "" }()
-	tc := NewNvidiaTranscoder("123")
-	ffmpeg.InitFFmpeg()
-
-	// test.ts sample isn't in a supported pixel format, so use this instead
-	fname := "test2.ts"
-
-	// transcoding should fail due to invalid devices
-	profiles := []ffmpeg.VideoProfile{ffmpeg.P144p30fps16x9, ffmpeg.P240p30fps16x9}
-	md := stubMetadata("", profiles...)
-	md.Fname = fname
-	_, err := tc.Transcode(md)
-	if err == nil ||
-		(err.Error() != "Unknown error occurred" &&
-			err.Error() != "Cannot allocate memory") {
-		t.Error(err)
-	}
-
 	dev := os.Getenv("NV_DEVICE")
 	if dev == "" {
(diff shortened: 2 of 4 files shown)


yondonfu (commented):
A few questions:

- How will B behave if an O/T crashes during a transcode as a result of this change?
- Does the HTTP conn get closed cleanly with an error causing B to immediately switch to another O/T?
- Was there any progress with distinguishing CUDA illegal addresses so Bs could mark those errors as non-retryable as mentioned in https://github.com/livepeer/go-livepeer/issues/1921#issuecomment-942893848?

yondonfu:
I haven't tested this, but a few thoughts below...

> Does the HTTP conn get closed cleanly with an error causing B to immediately switch to another O/T?

I don't think the HTTP conn will be closed cleanly if the O/T panics in the middle of a transcode.

> How will B behave if an O/T crashes during a transcode as a result of this change?

I think the B will wait for a timeout and then switch to a different O/T.

It would be preferable if the O/T could return an error immediately to B so B can switch before the O/T panics so that B doesn't have to wait for a timeout. Maybe panic recovery as described in [this post](https://medium.com/@masnun/panic-recovery-middleware-for-go-http-handlers-51147c941f9) could help here? The O/T could recover from the panic, return the error and then trigger another panic to crash. An alternative would be for go-livepeer to check for the unrecoverable LPMS error introduced in https://github.com/livepeer/lpms/pull/267/files and trigger the panic only in go-livepeer instead of triggering it in LPMS.

> Was there any progress with distinguishing CUDA illegal addresses so Bs could mark those errors as non-retryable as mentioned in #1921 (comment)?

I'm not sure how confident we are that the "Unknown error" that is currently classified as unrecoverable in LPMS is always attributable as a fault of the source segment as opposed to the O/T (as already mentioned in [this comment](https://github.com/livepeer/go-livepeer/issues/1921#issuecomment-941178946)). There is some discussion around propagating an HTTP status code to the client of the broadcaster to tell them to stop retrying if the broadcaster hits the max # of retries - if that goes through then distinguishing CUDA illegal address errors on Bs would be less important.

yondonfu:
Since this is blocking https://github.com/livepeer/go-livepeer/pull/2066 I'm going to approve/merge this for now. Created https://github.com/livepeer/go-livepeer/issues/2070 to track the improvement to O/T behavior.

yondonfu (approved):
LGTM

PR #2059, opened 2021-10-20
Add metrics for Fast Verification

### LINKED ISSUE #2052: Add metrics for fast verification
### Metrics to add

Counters:
- Verifications done
- Verifications failed
In Grafana that will be shown as
- Verifications Success = fail / done

Counters:
- for `segment_transcoded_total` counter add tags trusted/untrusted

- add `verify_enabled_current_sessions_total`. It will be number of current sessions that have verification enabled.
In Grafana it will be
`verify_enabled_current_sessions_total`/`current_sessions_total` = percentage of current sessions that have verification enabled

Counters:
- add `using_untrusted_verified_orchestrator_current_sessions_total`. It will be number of current sessions that have verification enabled and actually using untrusted (and verified) orchestartor.
In Grafana it will be
`using_untrusted_verified_orchestrator_current_sessions_total`/`current_sessions_total` = percentage of current sessions that actually use untrusted orchestrator.

darkdarkdragon:
Adds new metrics for tracking fast verification

Manually

Fixes #2052 

**Checklist:**

### DIFF
--- a/core/orchestrator.go
+++ b/core/orchestrator.go
@@ -569,7 +569,7 @@ func (n *LivepeerNode) transcodeSeg(config transcodeConfig, seg *stream.HLSSegme
 	took := time.Since(start)
 	glog.V(common.DEBUG).Infof("Transcoding of segment manifestID=%s sessionID=%s seqNo=%d took=%v", string(md.ManifestID), md.AuthToken.SessionId, seg.SeqNo, took)
 	if monitor.Enabled {
-		monitor.SegmentTranscoded(0, seg.SeqNo, md.Duration, took, common.ProfilesNames(md.Profiles))
+		monitor.SegmentTranscoded(0, seg.SeqNo, md.Duration, took, common.ProfilesNames(md.Profiles), true, true)
 	}
 
 	// Prepare the result object
--- a/core/transcoder.go
+++ b/core/transcoder.go
@@ -54,7 +54,7 @@ func (lt *LocalTranscoder) Transcode(md *SegTranscodingMetadata) (*TranscodeData
 		// When orchestrator works as transcoder, `fname` will be relative path to file in local
 		// filesystem and will not contain seqNo in it. For that case `SegmentTranscoded` will
 		// be called in orchestrator.go
-		monitor.SegmentTranscoded(0, seqNo, md.Duration, time.Since(start), common.ProfilesNames(profiles))
+		monitor.SegmentTranscoded(0, seqNo, md.Duration, time.Since(start), common.ProfilesNames(profiles), true, true)
 	}
 
 	return resToTranscodeData(res, opts)
@@ -95,7 +95,7 @@ func (nv *NvidiaTranscoder) Transcode(md *SegTranscodingMetadata) (*TranscodeDat
 		// When orchestrator works as transcoder, `fname` will be relative path to file in local
 		// filesystem and will not contain seqNo in it. For that case `SegmentTranscoded` will
 		// be called in orchestrator.go
-		monitor.SegmentTranscoded(0, seqNo, md.Duration, time.Since(start), common.ProfilesNames(profiles))
+		monitor.SegmentTranscoded(0, seqNo, md.Duration, time.Since(start), common.ProfilesNames(profiles), true, true)
 	}
 
 	return resToTranscodeData(res, out)
--- a/monitor/census.go
+++ b/monitor/census.go
@@ -90,6 +90,8 @@ type (
 		kRecipient                    tag.Key
 		kManifestID                   tag.Key
 		kSegmentType                  tag.Key
+		kTrusted                      tag.Key
+		kVerified                     tag.Key
 		mSegmentSourceAppeared        *stats.Int64Measure
 		mSegmentEmerged               *stats.Int64Measure
 		mSegmentEmergedUnprocessed    *stats.Int64Measure
@@ -156,6 +158,12 @@ type (
 		// Metrics for pixel accounting
 		mMilPixelsProcessed *stats.Float64Measure
 
+		// Metrics for fast verification
+		mFastVerificationDone                   *stats.Int64Measure
+		mFastVerificationFailed                 *stats.Int64Measure
+		mFastVerificationEnabledCurrentSessions *stats.Int64Measure
+		mFastVerificationUsingCurrentSessions   *stats.Int64Measure
+
 		lock        sync.Mutex
 		emergeTimes map[uint64]map[uint64]time.Time // nonce:seqNo
 		success     map[uint64]*segmentsAverager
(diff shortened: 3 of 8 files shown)


server/mediaserver.go
```diff
@@ -602,15 +602,30 @@ func (s *LivepeerServer) registerConnection(rtmpStrm stream.RTMPVideoStream) (*r
...
 
 	return cxn, nil
 }
 
+func countStreamsWithFastVerificationEnabled(rtmpConnections map[core.ManifestID]*rtmpConnection) (int, int) {
+	var enabled, using int
+	for _, cxn := range rtmpConnections {
+		if cxn.params.VerificationFreq > 0 {
+			enabled++
+			if cxn.sessManager.usingVerified() {
```
jailuthra, line 621: From what I understand `cxn.sessManager.verifiedSession` is always `nil` at the start of a stream. Only after the first segment is processed, and one of the untrusted Os pass verification, will this evaluate to true.

And this metric is only updated when a new stream is started or an old stream ends, so it will be lagging behind the actual value right? Not a problem if we have > 1 streams, but want to clarify for my understanding.

jailuthra (approved):
Had a question around the verified session count metric, but other than that LGTM!

And sorry about the force-push, I realized later that this PR had target as my branch. So I re-applied your patch via `git format-patch` and `git am` and updated this PR.

monitor/census.go
```diff
@@ -286,6 +296,14 @@ func InitCensus(nodeType NodeType, version string) {
 	// Metrics for pixel accounting
 	census.mMilPixelsProcessed = stats.Float64("mil_pixels_processed", "MilPixelsProcessed", "mil pixels")
 
+	// Metrics for fast verification
+	census.mFastVerificationDone = stats.Int64("fast_verification_done", "FastVerificationDone", "tot")
+	census.mFastVerificationFailed = stats.Int64("fast_verification_failed", "FastVerificationFailed", "tot")
+	census.mFastVerificationEnabledCurrentSessions = stats.Int64("fast_verification_enabled_current_sessions_total",
+		"Number of currently transcoded streams that have fast verificatoin enabled", "tot")
```
yondonfu:
```suggestion
		"Number of currently transcoded streams that have fast verification enabled", "tot")
```

monitor/census.go
```diff
@@ -286,6 +296,14 @@ func InitCensus(nodeType NodeType, version string) {
...
+	census.mFastVerificationUsingCurrentSessions = stats.Int64("fast_verification_using_current_sessions_total",
+		"Number of currently transcoded streams that have fast verificatoin enabled", "tot")
```
yondonfu:
```suggestion
		"Number of currently transcoded streams that have fast verification enabled and that are using an untrusted orchestrator", "tot")
```

monitor/census.go
```diff
@@ -736,6 +754,36 @@ func InitCensus(nodeType NodeType, version string) {
...
+			Name:        "fast_verification_failed",
+			Measure:     census.mFastVerificationFailed,
+			Description: "Number of fast verifications failed",
+			TagKeys:     baseTags,
+			Aggregation: view.Count(),
+		},
+		{
+			Name:        "fast_verification_enabled_current_sessions_total",
+			Measure:     census.mFastVerificationEnabledCurrentSessions,
+			Description: "Number of currently transcoded streams that have fast verificatoin enabled",
```
yondonfu:
```suggestion
			Description: "Number of currently transcoded streams that have fast verification enabled",
```

monitor/census.go
```diff
@@ -736,6 +754,36 @@ func InitCensus(nodeType NodeType, version string) {
...
+			TagKeys:     baseTags,
+			Aggregation: view.Count(),
+		},
+		{
+			Name:        "fast_verification_using_current_sessions_total",
+			Measure:     census.mFastVerificationUsingCurrentSessions,
+			Description: "Number of currently transcoded streams that have fast verificatoin enabled and using untrusted orchestrator",
```
yondonfu:
```suggestion
			Description: "Number of currently transcoded streams that have fast verification enabled and that are using an untrusted orchestrator",
```

server/mediaserver.go
yondonfu: Good point by Jai.

Seems like we should:

- Update `fast_verification_enabled_current_sessions_total` when a new stream starts or when an existing stream ends
- Update `fast_verification_using_current_sessions_total` when a stream starts using a verified untrusted orchestrator and when a stream stops using a verified untrusted orchestrator (i.e. in `broadcast.go`)

core/orchestrator.go
```diff
@@ -569,7 +569,7 @@ func (n *LivepeerNode) transcodeSeg(config transcodeConfig, seg *stream.HLSSegme
 	took := time.Since(start)
 	glog.V(common.DEBUG).Infof("Transcoding of segment manifestID=%s sessionID=%s seqNo=%d took=%v", string(md.ManifestID), md.AuthToken.SessionId, seg.SeqNo, took)
 	if monitor.Enabled {
-		monitor.SegmentTranscoded(0, seg.SeqNo, md.Duration, took, common.ProfilesNames(md.Profiles))
+		monitor.SegmentTranscoded(0, seg.SeqNo, md.Duration, took, common.ProfilesNames(md.Profiles), true, true)
```
yondonfu, line 572: Should we have a separate function to record this metric when using it on an orchestrator to make it clear that the `trusted` and `verified` tags are not relevant when recording the metric for an orchestrator? Otherwise, this call reads as `trusted = true` and `verified = true` which is a bit weird since that information isn't relevant for the orchestrator.

core/transcoder.go
```diff
@@ -95,7 +95,7 @@ func (nv *NvidiaTranscoder) Transcode(md *SegTranscodingMetadata) (*TranscodeDat
 		// When orchestrator works as transcoder, `fname` will be relative path to file in local
 		// filesystem and will not contain seqNo in it. For that case `SegmentTranscoded` will
 		// be called in orchestrator.go
-		monitor.SegmentTranscoded(0, seqNo, md.Duration, time.Since(start), common.ProfilesNames(profiles))
+		monitor.SegmentTranscoded(0, seqNo, md.Duration, time.Since(start), common.ProfilesNames(profiles), true, true)
```
yondonfu, line 98: Same comment for these calls to `SegmentTranscoded` as my comment in `orchestrator.go`.

yondonfu (changes requested):

core/orchestrator.go
darkdarkdragon: Problem here is that Prometheus metrics tied to tags - we can't have `segment_transcoded_total` metric with trusted/untrusted tags and `segment_transcoded_total` metrics completely without tags.
We can have separate `broadcaster_segment_transcoded_total` metric and `orchestrator_segment_transcoded_total` metric, or we can use one `segment_transcoded_total` metric and use special value `orch` (or something) for value for these tags.
But for me it seems unnecessary . I don't mind seeing 'trusted/verified' on orchestrator's metrics - in the end, orchestrator is trusted for itself 😄

server/mediaserver.go
darkdarkdragon: Fixed in dad2396e442539c8572d9f3e9e492584db002724

yondonfu: I think there is still a small delay with the changes in dad2396 because if B switches to an untrusted or trusted O, the metric won't be updated until the next segment comes in because this metric update happens before an O is actually selected. In practice, this is probably okay for now, but an improvement would be to actually update the `fast_verification_using_current_sessions_total` metric upon selecting an untrusted/trusted O in `broadcast.go`.

core/orchestrator.go
yondonfu: I see. Okay for now.

yondonfu (approved):
LGTM. Let's squash the additional commits before a merge.

yondonfu (changes requested):
Actually looks like there is a race detected in tests. Let's look into that first.

yondonfu (approved):

PR #2068, opened 2021-10-24
server: Generic error non-retryable

yondonfu:

- 
- 
- 

**Checklist:**


yondonfu:
This is probably not a good idea generally because if a "Generic error..." is returned due to a OOM error then it could either be the case that the source segment was bad and somehow triggered an OOM error or it could be the case that all the available memory was actually used up. For the latter case, we would actually want to retry with another O because the problem was with the O as opposed to with the source segment.

PR #2071, opened 2021-10-26
pm: Update tx cost check for faceValue()

### LINKED ISSUE #2049: Remove tx cost check when calculating ticket face value for O
Right now, when an O calculates the face value for a ticket, if the face value is less than the estimated tx cost, the O returns an [insufficient sender reserve error](https://github.com/livepeer/go-livepeer/blob/e1518db8fb0350c520c6b304a1fb7046380dc3da/pm/recipient.go#L245). The idea behind this behavior was that if the estimated tx cost can't be covered by the face value then the O can just return an error early. But, the problem with this behavior is that the estimated tx cost could change in between the time that the ticket face value is calculated and advertised and the time that a winning ticket is to be redeemed. It is possible that the face value can't cover the estimated tx cost right now, but it *could* cover it later on when gas prices have come down. I think it would be preferable behavior for the O to return the ticket face value even if it can't cover the estimated tx cost right now because it could cover the estimated tx cost later. Additionally, the O is configured not to redeem the ticket if the face value at the time of redemption can't cover the tx cost so there is still protection from unprofitable redemptions there.

yondonfu:

This PR addresses the problem described in #2049 which can occur when there are really big gas price spikes (think 500+ gwei) on mainnet that last for a relatively short period of time.

The solution in this PR contains an update to the solution presented in #2049. The tx cost check should not be completely removed because then it is possible for an O to do a bunch of transcoding without ever being able to redeem any tickets profitably. Instead, the tx cost check is updated such that O calculates the tx cost based on the current estimated gas price and the average estimated gas price (which could be calculated over a time period such as a few hours, a day, etc.). If there is a gas price spike and the desired ticket face value is less than the tx cost based on the current gas price then O also checks the face value against the tx cost based on the average gas price. If this fallback check passes then O proceeds to generate ticket params. The idea is that typically the current gas price will be less than or equal to the average gas price so the fallback check never runs, but if the current gas price exceeds the average gas price then as long as the average gas price check passes there is still a reasonable chance that the current gas price falls at some point such that the ticket face value can cover the tx cost based on the current gas price.

In this first pass, the average gas price value is hardcoded for now and the intent is to replace the hardcoded value with a dynamic way to determine the average gas price over a period of time later (i.e. using the ETH JSON-RPC feeHistory API or in a GasPriceMonitor implementation).

See commit history.

Updated unit tests.

Fixes #2049 

**Checklist:**

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -24,6 +24,8 @@
 
 #### Orchestrator
 
+- \#2071 Update tx cost check when generating ticket params to be more resistant to gas price spikes (@yondonfu)
+
 #### Transcoder
 
 - \#2057 Prevent stuck sessions by crashing on unrecoverable CUDA errors (@jailuthra)
--- a/pm/recipient.go
+++ b/pm/recipient.go
@@ -26,6 +26,10 @@ var paramsExpiryBuffer = int64(1)
 
 var evMultiplier = big.NewInt(100)
 
+// Hardcode to 200 gwei
+// TODO: Replace this hardcoded value by dynamically determining the average gas price during a period of time
+var avgGasPrice = new(big.Int).Mul(big.NewInt(200), new(big.Int).Exp(big.NewInt(10), big.NewInt(9), nil))
+
 // Recipient is an interface which describes an object capable
 // of receiving tickets
 type Recipient interface {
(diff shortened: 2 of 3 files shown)


darkdarkdragon (approved):
LGTM

PR #2073, opened 2021-10-26
Make filtering orchestrators that haven't been updated in last day optional

### LINKED ISSUE #1482: "error processing payment: orchestrator is inactive" -> but orchestrator is active
**Describe the bug**
Getting the following error messages:
`error processing payment: orchestrator is inactive, cannot process payments`
But my orchestrator is active and successfully calling rewards and the round is initialized.

**Screenshots**
image

yondonfu:

This PR addresses https://github.com/livepeer/go-livepeer/issues/1482.

Based on the OP in the above issue and [this comment](https://github.com/livepeer/go-livepeer/issues/1482#issuecomment-943323704) I suspect the problem is that all orchestrator table queries will filter out any orchestrator entries that have not been updated in the last day. I had previously thought that existing orchestrator entries would be updated at least once a day when a new round is initialized as mentioned [here](https://github.com/livepeer/go-livepeer/issues/1482#issuecomment-943323704). I'm still not sure why this wouldn't be happening. But, the screenshot provided in the OP of https://github.com/livepeer/go-livepeer/issues/1482 does indicate a `updatedAt` timestamp that is more than 1 day in the past which would explain the failed active check - the `updatedAt` timestamp was from 4/30/20 and the active check started failing on 5/5/20. Note: I'm assuming that the DB dump in that post was created on or after 5/5/20.

So, the `updatedAt` check is my best guess of the problem right now and this PR disables the `updatedAt` check unless it is explicitly enabled using `DBOrchFilter.UpdatedLastDay`. The check is explicitly enabled in `DBOrchestratorPoolCache` when deciding which orchestrators to query in order to preserve existing behavior for now - changes to this behavior can be considered separately. 

Also included some additional logging around the active check for informational purposes in case there is an additional problem in addition to the `updatedAt` check.

See commit history.

Added unit tests.

Fixes #1482

**Checklist:**

### DIFF
--- a/common/db.go
+++ b/common/db.go
@@ -63,9 +63,10 @@ type DBUnbondingLock struct {
 
 // DBOrchFilter is an object used to attach a filter to a selectOrch query
 type DBOrchFilter struct {
-	MaxPrice     *big.Rat
-	CurrentRound *big.Int
-	Addresses    []ethcommon.Address
+	MaxPrice       *big.Rat
+	CurrentRound   *big.Int
+	Addresses      []ethcommon.Address
+	UpdatedLastDay bool
 }
 
 var LivepeerDBVersion = 1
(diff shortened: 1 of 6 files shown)


common/db.go
```diff
@@ -822,29 +823,40 @@ func buildOrchCountQuery(filter *DBOrchFilter) (string, error) {
 }
 
 func buildFilterOrchsQuery(filter *DBOrchFilter) (string, error) {
-	qry := "WHERE updatedAt >= datetime('now','-1 day')"
+	qry := ""
+	var filters []string
```
yondonfu, line 827: Switched to joining a slice of strings with the " AND " delimiter because it is possible that we end up with a query without a WHERE clause or any ANDs.

darkdarkdragon (approved):
LGTM

PR #2082, opened 2021-11-01
wip: arm64 Docker builds

iameli:
For M1 macs and neato 128-core ARM servers.

Looks like CUDA drivers are available for ARM64, so the release isn't a problem, but there's no TensorFlow support quite yet. That's the only change between the AMD and ARM builds.

Downstream of https://github.com/livepeer/go-livepeer/pull/2054 and https://github.com/livepeer/go-livepeer/pull/2051, those should get merged first.


iameli:
Closed in favor of https://github.com/livepeer/go-livepeer/pull/2499

PR #2083, opened 2021-11-02
server/broadcast: don't retry segments after max attempts

### LINKED ISSUE #2069: Return 422 to HTTP push client when max # of retries is hit
We already return a 422 if the transcoding error is a [non-retryable error](https://github.com/livepeer/go-livepeer/blob/e1518db8fb0350c520c6b304a1fb7046380dc3da/server/mediaserver.go#L931). There are cases where a B cannot currently attribute a transcoding error to the source segment or the O. As a result, B cannot just mark these errors as non-retryable and should retry. However, if the B hits the max # of retries then it is probably safe (given a reasonable cap on the # of retries - the default is 3) to consider the segment as not transcodeable. In these cases, B can return a 422 as well to tell the HTTP push client to not send the segment again.

jailuthra:

 Return 422 to push-api clients once the B has retried a segment `MaxAttempts` times.
 
See commit history 

1. Unit test to ensure 422 is returned when retried a segment `MaxAttempts` times
2. Manual testing of Push API after returning an error for transcoding:
```diff
diff --git a/server/broadcast.go b/server/broadcast.go
index ff17a0d5..34c89e37 100644
--- a/server/broadcast.go
+++ b/server/broadcast.go
@@ -612,6 +612,9 @@ func transcodeSegment(cxn *rtmpConnection, seg *stream.HLSSegment, name string,
                        err = errors.New("empty response")
                }
                return nil, info, err
+       } else if sess != nil {
+               cxn.sessManager.completeSession(sess)
+               return nil, info, errors.New("Unknown error")
        }

        // download transcoded segments from the transcoder
```

Fixes #2069 

**Checklist:**

### DIFF
--- a/server/push_test.go
+++ b/server/push_test.go
@@ -55,12 +55,6 @@ func TestPush_ShouldReturn422ForNonRetryable(t *testing.T) {
 	w := httptest.NewRecorder()
 	req := httptest.NewRequest("POST", "/live/mani/18.ts", reader)
 
-	oldAttempts := MaxAttempts
-	defer func() {
-		MaxAttempts = oldAttempts
-	}()
-	MaxAttempts = 1
-
 	dummyRes := func(err string) *net.TranscodeResult {
 		return &net.TranscodeResult{
 			Result: &net.TranscodeResult_Error{Error: err},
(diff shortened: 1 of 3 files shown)


server/push_test.go
```diff
@@ -95,15 +89,40 @@ func TestPush_ShouldReturn422ForNonRetryable(t *testing.T) {
 
 	s.rtmpConnections["mani"] = cxn
 
-	req.Header.Set("Accept", "multipart/mixed")
+	// Should return 503 if got a retryable error with no sessions to use
```
yondonfu, line 97: I was thinking about whether there could be a case where B runs out of sessions before it hits the max # of attempts. Dumping some thoughts below just for visibility.

Based on https://github.com/livepeer/go-livepeer/blob/fbcdccb6ca545ec68f8a6bd533f3d843ec6ad9b1/server/broadcast.go#L775 I believe that if B runs out of sessions, it will still hit the max # of attempts - each of the remaining attempts will just result in an error caused by B being out of sessions. And since we check for a non-retryable error [here](https://github.com/livepeer/go-livepeer/blob/fbcdccb6ca545ec68f8a6bd533f3d843ec6ad9b1/server/mediaserver.go#L959) and return before the [no sessions available error check](https://github.com/livepeer/go-livepeer/blob/fbcdccb6ca545ec68f8a6bd533f3d843ec6ad9b1/server/mediaserver.go#L974) in the HTTP push handler we will always return a 422 in this scenario before getting to the code that triggers a 503.

yondonfu (approved):
LGTM

PR #2085, opened 2021-11-02
server: Set max refresh sessions threshold to 8

### LINKED ISSUE #2080: Fix cyclic CPU usage spikes on broadcasters due to discovery requests
https://github.com/livepeer/go-livepeer/pull/2022/commits/de71460cc2fc2c4fa33873f8af9f622b7c56be15

yondonfu:

This PR sets the max refresh sessions threshold to 8.

Prior to this PR, a B would refresh its session list (resulting in running orchestrator discovery to create a new session list) whenever the size of the list falls below `numOrchs / 2` where `numOrchs` is the # of Os that the B is aware of. The problem with this check is illustrated by the following example:

- numOrchs = 100
- Only 50 Os out of 100 respond during discovery
- As a result, the typical size of the session list is 50
- If a single O in that session list returns an error and is suspended from the list the size of the list falls to 49
- B refreshes its session list
- Repeat the above process

The consequence is that B can end up running discovery/refreshing its session list really frequently.

This PR changes B's behavior so that if `numOrchs / 2 > 8`, then B will only refresh its session list if the size falls below 8. However, if `numOrchs / 2 < 8` then B will refresh its session list if the size falls below `numOrchs / 2`. This means that if B is aware of a lot of Os (i.e. on the public network) it will wait until the size of the session list falls to a lower level before refreshing. But, if B is not aware of many Os (i.e. in a smaller local private network) it will refresh when the session list is at a higher level.

See commit history.

Tested manually. Would like to add unit tests for this area of the codebase but haven't gotten to it yet...

Fixes #2080 

**Checklist:**

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -10,6 +10,8 @@
 
 #### Broadcaster
 
+- \#2085 Set max refresh sessions threshold to 8 (@yondonfu)
+
 #### Orchestrator
 
 #### Transcoder
--- a/server/broadcast.go
+++ b/server/broadcast.go
@@ -37,6 +37,9 @@ import (
 var refreshTimeout = 2500 * time.Millisecond
 var maxDurationSec = common.MaxDuration.Seconds()
 
+// Max threshold for # of broadcast sessions under which we will refresh the session list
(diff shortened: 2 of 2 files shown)


jailuthra (approved):
LGTM

Issue #2070, opened 2021-10-25
Return error to B from O before panic due to unrecoverable LPMS error


yondonfu:
Repost from https://github.com/livepeer/go-livepeer/pull/2057#issuecomment-950984324

> I don't think the HTTP conn will be closed cleanly if the O/T panics in the middle of a transcode.

> I think the B will wait for a timeout and then switch to a different O/T.

> It would be preferable if the O/T could return an error immediately to B so B can switch before the O/T panics so that B doesn't have to wait for a timeout. Maybe panic recovery as described in this post could help here? The O/T could recover from the panic, return the error and then trigger another panic to crash. An alternative would be for go-livepeer to check for the unrecoverable LPMS error introduced in https://github.com/livepeer/lpms/pull/267/files and trigger the panic only in go-livepeer instead of triggering it in LPMS.

yondonfu:
Relevant areas of the codebase:

The transcoding call for a combined OT happens here:

https://github.com/livepeer/go-livepeer/blob/d3a869665318cc69126427260e7ac0b4f106bf85/core/orchestrator.go#L557

Note that we use a single `Transcoder` interface that can be implemented by `RemoteTranscoderManager` when running as a standalone O and that can be implemented by `LoadBalancingTranscoder` (which uses `NvidiaTranscoder` under the hood for each individual GPU device available) or `LocalTranscoder`. When we run a combined OT with GPU access we default to using `LoadBalancingTranscoder` as the implementation for the `Transcoder` interface - see [here](https://github.com/livepeer/go-livepeer/blob/d3a869665318cc69126427260e7ac0b4f106bf85/cmd/livepeer/livepeer.go#L296).

[...]

PR #2094, opened 2021-11-05
Gracefully notify orchestrator in case of a panic in transcoder

### LINKED ISSUE #2088: Return error from T to O before panic due to unrecoverable LPMS error
Related to #2070 

If T returns an error to O before a panic in this scenario then O can immediately switch to another T instead of waiting for a timeout.

Relevant areas of codebase:

The transcoding call happens in T https://github.com/livepeer/go-livepeer/blob/fbcdccb6ca545ec68f8a6bd533f3d843ec6ad9b1/server/ot_rpc.go#L149

Logic for writing an error from T to O https://github.com/livepeer/go-livepeer/blob/fbcdccb6ca545ec68f8a6bd533f3d843ec6ad9b1/server/ot_rpc.go#L156

Logic on O for handling results from T https://github.com/livepeer/go-livepeer/blob/fbcdccb6ca545ec68f8a6bd533f3d843ec6ad9b1/core/orchestrator.go#L716

leszko:

Fix transcoder to send an error to orchestrator before crashing. Before, when the transcoder crashed after `panic()`, the orchestrator had to time out before trying another transcoder. The change is related only to the split O+T topology.

- Add `recover()` in to the `transcoder` functions (both for Nvidia and the standard CPU transcoding)
- Add handling `UnrecoverableError` in `ot_rpc.go` to send the error message to orchestrator and only then to panic and crash

1. Add an artificial `panic(errors.New("Some error"))` inside `transcoder.go#Transcode()` function
2. Build the project
3. Start orchestrator + 2 transcoders
4. Start broadcaster and a start streaming a video
5. The first transcoder crashes, orchestrator is notified and selects the second transcoder without the timeout

The same test without this PR, makes orchestrator wait for the timeout before selecting the second transcoder.

fix #2088 

**Checklist:**

### DIFF
--- a/core/transcoder.go
+++ b/core/transcoder.go
@@ -27,12 +27,18 @@ type LocalTranscoder struct {
 	workDir string
 }
 
-type UnrecoverableError error
+type UnrecoverableError struct {
+	error
+}
+
+func NewUnrecoverableError(err error) UnrecoverableError {
+	return UnrecoverableError{err}
+}
 
 var WorkDir string
 
 func (lt *LocalTranscoder) Transcode(md *SegTranscodingMetadata) (td *TranscodeData, retErr error) {
-	// Returns UnrecoverableError instead of panic to gracefully notify orchestrator about transcoder's failure
+	// Returns UnrecoverableError instead of panicking to gracefully notify orchestrator about transcoder's failure
 	defer recoverFromPanic(&retErr)
 
 	// Set up in / out config
@@ -75,7 +81,7 @@ type NvidiaTranscoder struct {
 }
 
 func (nv *NvidiaTranscoder) Transcode(md *SegTranscodingMetadata) (td *TranscodeData, retErr error) {
-	// Returns UnrecoverableError instead of panic to gracefully notify orchestrator about transcoder's failure
+	// Returns UnrecoverableError instead of panicking to gracefully notify orchestrator about transcoder's failure
 	defer recoverFromPanic(&retErr)
 
 	in := &ffmpeg.TranscodeOptionsIn{
@@ -265,6 +271,6 @@ func recoverFromPanic(retErr *error) {
 		if !ok {
 			err = errors.New("unrecoverable transcoding failure")
 		}
-		*retErr = UnrecoverableError(err)
+		*retErr = NewUnrecoverableError(err)
 	}
 }
--- a/core/transcoder.go
+++ b/core/transcoder.go
@@ -27,9 +27,14 @@ type LocalTranscoder struct {
 	workDir string
 }
 
+type UnrecoverableError error
+
 var WorkDir string
 
-func (lt *LocalTranscoder) Transcode(md *SegTranscodingMetadata) (*TranscodeData, error) {
+func (lt *LocalTranscoder) Transcode(md *SegTranscodingMetadata) (td *TranscodeData, retErr error) {
+	// Returns UnrecoverableError instead of panic to gracefully notify orchestrator about transcoder's failure
+	defer recoverFromPanic(&retErr)
+
 	// Set up in / out config
 	in := &ffmpeg.TranscodeOptionsIn{
 		Fname: md.Fname,
@@ -69,7 +74,9 @@ type NvidiaTranscoder struct {
 	session *ffmpeg.Transcoder
 }
 
-func (nv *NvidiaTranscoder) Transcode(md *SegTranscodingMetadata) (*TranscodeData, error) {
+func (nv *NvidiaTranscoder) Transcode(md *SegTranscodingMetadata) (td *TranscodeData, retErr error) {
+	// Returns UnrecoverableError instead of panic to gracefully notify orchestrator about transcoder's failure
+	defer recoverFromPanic(&retErr)
 
 	in := &ffmpeg.TranscodeOptionsIn{
 		Fname:  md.Fname,
@@ -250,4 +257,14 @@ func detectorsToTranscodeOptions(workDir string, accel ffmpeg.Acceleration, prof
 		opts[i] = o
 	}
 	return opts
+}
+
+func recoverFromPanic(retErr *error) {
+	if r := recover(); r != nil {
+		err, ok := r.(error)
+		if !ok {
+			err = errors.New("unrecoverable transcoding failure")
+		}
+		*retErr = UnrecoverableError(err)
+	}
 }
--- a/core/transcoder_test.go
+++ b/core/transcoder_test.go
@@ -300,6 +300,7 @@ func TestRecoverFromPanic(t *testing.T) {
 
 	assert.NotNil(err)
 	assert.Equal("unrecoverable transcoding failure", err.Error())
+	assert.IsType(UnrecoverableError{}, err)
 }
 
 func TestRecoverFromPanic_WithError(t *testing.T) {
--- a/core/transcoder_test.go
+++ b/core/transcoder_test.go
@@ -1,6 +1,7 @@
 package core
 
 import (
+	"errors"
 	"io/ioutil"
 	"os"
 	"testing"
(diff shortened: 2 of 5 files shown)


core/transcoder.go
```diff
@@ -27,9 +27,20 @@ type LocalTranscoder struct {
 	workDir string
 }
 
+type UnrecoverableError struct {
```
yondonfu, line 30: Since all this struct does is wrap an `error` could we simplify by defining `UnrecoverableError` as an `error`?

```golang
type UnrecoverableError error
```

And then if we need to create an `UnrecoverableError` we could just cast an existing error i.e. `UnrecoverableError(err)`.

server/ot_rpc.go
```diff
@@ -152,6 +152,9 @@ func runTranscode(n *core.LivepeerNode, orchAddr string, httpc *http.Client, not
 		err = errors.New("segment / profile mismatch")
 	}
 	if err != nil {
+		if _, ok := err.(core.UnrecoverableError); ok {
+			defer panic(err)
```
yondonfu, line 156: I recall that when we were pairing we saw an IDE warning about deferred panics. Curious what that warning was about and if you think it is fine to use a deferred panic here?

core/transcoder_test.go
```diff
@@ -287,3 +287,31 @@ func TestTranscoder_Formats(t *testing.T) {
...
+
+	f := func() (err error) {
+		defer recoverFromPanic(&err)
+		panic(struct{}{})
+	}
+
+	err := f()
+
+	assert.NotNil(err)
+	assert.Equal("unrecoverable transcoding failure", err.Error())
```
yondonfu, line 302: Should we also add an assertion to make sure that the "unrecoverable transcoding failure" error is also a `UnrecoverableError`?

yondonfu (commented):
Looks great! Just a few minor nit comments.

Also, I believe the changes in `LocalTranscoder` and `NvidiaTranscoder` should be sufficient, but I wanted to just note that when the node does GPU transcoding, it actually uses the [LoadBalancingTranscoder](https://github.com/livepeer/go-livepeer/blob/d3a869665318cc69126427260e7ac0b4f106bf85/core/lb.go#L30) type which also implements the `Transcoder` interface. Under the hood, the `LoadBalancingTranscoder` uses `NvidiaTranscoder` [1]. When `LoadBalancingTranscoder`'s [Transcode() method](https://github.com/livepeer/go-livepeer/blob/d3a869665318cc69126427260e7ac0b4f106bf85/core/lb.go#L55) is used, there is actually a [nested NvidiaTranscoder.Transcode() call in a goroutine](https://github.com/livepeer/go-livepeer/blob/d3a869665318cc69126427260e7ac0b4f106bf85/core/lb.go#L185). I believe a panic in `NvidiaTranscoder`'s `Transcode` call would be caught and returned as an error to `LoadBalancingTranscoder` and then it would be handled in the `ot_rpc.go` code the same way that it would be handled if the error was returned directly by `NvidiaTranscoder` (but feel free to correct me here if there is any disagreement on the expected behavior!).

[1] See this comment for more details https://github.com/livepeer/go-livepeer/issues/2070#issuecomment-962163428

leszko:
> I believe a panic in `NvidiaTranscoder`'s `Transcode` call would be caught and returned as an error to `LoadBalancingTranscoder` and then it would be handled in the `ot_rpc.go` code the same way that it would be handled if the error was returned directly by `NvidiaTranscoder` (but feel free to correct me here if there is any disagreement on the expected behavior!).

Yes, you're right, it'll work exactly as you expected. I just double-checked it. We could consider adding a unit test in `lb_test.go`, however, I think we already test returning a custom error in [lb_test.go#TestLB_ConcurrentSessionErrors()](https://github.com/livepeer/go-livepeer/blob/master/core/lb_test.go#L211). So, IMO the scenario you described is already covered.

core/transcoder.go
leszko: Since `error` is an interface and not a struct, I think that if we do it this way, then we won't be able to distinguish.

In other words, If we use it this way.

```golang
type UnrecoverableError error
...
*retErr = UnrecoverableError(err)
```

Then, this condition is true for any error (not only for `UnrecoverableError`).

```golang
if _, ok := err.(core.UnrecoverableError); ok { }
```

Do you have any ideas for this?

core/transcoder_test.go
leszko: Good idea, added :+1:

server/ot_rpc.go
leszko: Here's the whole GoLand's warning.

>defer should not call panic() directly 
> Inspection info: Reports defer and go statements that call panic() or recover() directly.
> Such statements are rarely useful and might indicate a misuse of the panic() and recover() mechanism. In particular:
> - go panic(): a newly-started goroutine will panic immediately.
> - defer panic(): a function with this statement will always panic on exit.
> - go recover(): has no effect as newly-started goroutine cannot panic.
> - defer recover(): function with this statement will silently stop panicking. This could be a valid usage, but an idiomatic way is to inspect the value returned by recover():
> ```golang 
> defer func() {
>  if r := recover(); r != nil {
>    fmt.Println("Recovered from: ", r)
>  }
> }()
> ```
> For more information about go statements and panics handling, see Handling panics and Go statements in the Go Language Specification.

So, in our case, the warning is `a function with this statement will always panic on exit`. But I think it's exactly what we want to achieve (!). Wdyt?

leszko (commented):
Thanks for the review @yondonfu . I addressed your comments. PTAL.

core/transcoder.go
yondonfu: Ah I see. I always forget that `error` is an interface and not a struct. In that case, defining `UnrecoverableError` as a concrete type that implements the `error` interface so we can use a type assertion to check if an `error` is an `UnrecoverableError` seems reasonable to me.

server/ot_rpc.go
yondonfu: Hah yeah you're right. In this case, we do want to panic, but only when the function returns so that it can execute all the logic for returning the HTTP response before the panic. Guess this is a case where we can go against the warning!

yondonfu (commented):
Looks great! Last thing before merging...

We typically like to prefix commit messages with the package that the commit updates (see [this doc](https://github.com/lightningnetwork/lnd/blob/master/docs/code_contribution_guidelines.md#ideal-git-commit-structure) - we loosely follow the guidelines described) as its helpful when sifting through commit history later on. Would be great to update the commits accordingly and to also squash the commits updating CHANGELOG_PENDING into a single one [1]. Then we're good to go!

[1] We've typically saved this process for the end of the PR review process and at that time rebasing/modifying commit history is considered fair game (generally we try to avoid doing so in middle of the review process just for the reviewer's convenience although there are exceptions).

leszko:
> We typically like to prefix commit messages with the package that the commit updates (see [this doc](https://github.com/lightningnetwork/lnd/blob/master/docs/code_contribution_guidelines.md#ideal-git-commit-structure) - we loosely follow the guidelines described) as its helpful when sifting through commit history later on. Would be great to update the commits accordingly and to also squash the commits updating CHANGELOG_PENDING into a single one [1]. Then we're good to go!
> 
> [1] We've typically saved this process for the end of the PR review process and at that time rebasing/modifying commit history is considered fair game (generally we try to avoid doing so in middle of the review process just for the reviewer's convenience although there are exceptions).

Ok, good to know. Then, I think that this PR should be a single commit with the message `[core+server] Gracefully notify orchestrator in case of panic in transcoder`. When the PR is approved (by both @yondonfu  and @jailuthra), I'll just click "Squash and merge" and update the commit message there. Is that ok?

jailuthra (approved):
LGTM! :ship: :partying_face: 

> When the PR is approved [..], I'll just click "Squash and merge" and update the commit message there.

Yeah :+1: I think it's okay to squash and merge using github UI and edit the commit message there, given the changes in this PR don't touch many things.

For more complicated changes it's preferred to run a local rebase+squash to split independent things in separate commits, and force-push on the PR branch before merging.

>  I think that this PR should be a single commit with the message `[core+server] Gracefully notify orchestrator in case of panic in transcoder`

The message can probably be formatted as `core,server: Gracefully notify..` or `core+server: Gracefully notify..` to be closer to rest of the commit history. I don't think we've used the square brackets `[]` before.

yondonfu (approved):
LGTM! 🥳

PR #2103, opened 2021-11-12
server: Add suspending sessions that did not pass the pHash verification

leszko:
Suspend untrusted session when the following conditions are correct:
- The untrusted session returned incorrect p-hash
- Another untrusted session returned correct p-hash

It would be nice to add some automated tests for this change, however, it's hard to do it with the current design. I think we should tackle it separately while refactoring `broadcast.go`.
  
- Suspend the untrusted session that returned incorrect p-hash
  
1. Create a fake orchestrator that returns incorrect p-hash
2. Run the setup with 1 trusted orchestrator, 2 correct untrusted orchestrators, and 1 incorrect untrusted orchestrator
3. Check that from time to time the incorrect untrusted orchestrator is suspended for one session selection
  
  **Checklist:**
  
  
### DIFF
--- a/server/broadcast.go
+++ b/server/broadcast.go
@@ -519,6 +519,8 @@ func (bsm *BroadcastSessionsManager) chooseResults(submitResultsCh chan *SubmitR
 			trustedResults.TranscodeResult.Segments[segmToCheckIndex].PerceptualHashUrl, err)
 		return nil, nil, err
 	}
+
+	var sessionsToSuspend []*BroadcastSession
 	for _, untrustedResult := range untrustedResults {
 		untrustedHash, err := drivers.GetSegmentData(untrustedResult.TranscodeResult.Segments[segmToCheckIndex].PerceptualHashUrl)
 		if err != nil {
@@ -542,9 +544,16 @@ func (bsm *BroadcastSessionsManager) chooseResults(submitResultsCh chan *SubmitR
 			if untrustedResult.Err == nil {
 				bsm.sessionVerified(untrustedResult.Session)
 			}
+			// suspend sessions which returned incorrect results
+			for _, s := range sessionsToSuspend {
+				bsm.suspendAndRemoveOrch(s)
+			}
 			return untrustedResult.Session, untrustedResult.TranscodeResult, untrustedResult.Err
-		} else if monitor.Enabled {
-			monitor.FastVerificationFailed()
+		} else {
+			sessionsToSuspend = append(sessionsToSuspend, untrustedResult.Session)
+			if monitor.Enabled {
+				monitor.FastVerificationFailed()
+			}
 		}
 	}
 
--- a/server/push_test.go
+++ b/server/push_test.go
@@ -1630,7 +1630,6 @@ func TestPush_MultipartReturnMultiSession(t *testing.T) {
 	sess2.OrchestratorScore = common.Score_Untrusted
 
 	sess3 := StubBroadcastSession(ts3.URL)
-	glog.Errorf("====> ts 3 url %s ", ts3.URL)
 	sess3.Params.Profiles = []ffmpeg.VideoProfile{ffmpeg.P144p30fps16x9}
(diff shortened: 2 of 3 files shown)


darkdarkdragon (approved):
LGTM

jailuthra (approved):
LGTM. Made an unrelated sidecomment

server/broadcast.go
```diff
@@ -542,9 +544,16 @@ func (bsm *BroadcastSessionsManager) chooseResults(submitResultsCh chan *SubmitR
...
 				bsm.sessionVerified(untrustedResult.Session)
 			}
+			// suspend sessions which returned incorrect results
+			for _, s := range sessionsToSuspend {
+				bsm.suspendAndRemoveOrch(s)
+			}
 			return untrustedResult.Session, untrustedResult.TranscodeResult, untrustedResult.Err
-		} else if monitor.Enabled {
-			monitor.FastVerificationFailed()
+		} else {
```
jailuthra, line 527: Not related to this exact issue or very high priority, so maybe we should open a new issue/PR for this:

Maybe we should reset our `verifiedSession` to `nil` if the hash it sends doesn't match?

```suggestion
		} else {
			if (bsm.verifiedSession == untrustedResult.Session) {
				bsm.verifiedSession = nil
			}
```

leszko: I see one issue with adding this condition. If we add it, then a false-negative will cause `bsm.verifiedSession = nil` and in a result, we'll start looking for a new session. That can in turn cause bouncing between sessions if we encounter many false-negative scenarios.

We decided that we'll suspend sessions only if it failed p-hash verification but at the same time the other untrusted sessions passed the verification. We could do the same and maybe set `verifiedSession` to `nil` when we suspend the session, but technically it does not matter, because [this condition](https://github.com/livepeer/go-livepeer/blob/e2f94ca538a5bfe6b9ac9fe49848fb4b9536e135/server/broadcast.go#L434) takes into consideration both `bsm.verifiedSession = nil` and excluding suspended sessions.

jailuthra: Good point about false negatives causing swaps, and yeah we'd anyway suspend a verified session and look for a new one if the other untrusted session passes. LGTM feel free to merge :+1:

server/push_test.go
```diff
@@ -1604,14 +1629,24 @@ func TestPush_MultipartReturnMultiSession(t *testing.T) {
 	sess2.Params.ManifestID = "mani"
 	sess2.OrchestratorScore = common.Score_Untrusted
 
-	bsm := bsmWithSessListExt([]*BroadcastSession{sess1}, []*BroadcastSession{sess2}, false)
+	sess3 := StubBroadcastSession(ts3.URL)
+	glog.Errorf("====> ts 3 url %s ", ts3.URL)
```
yondonfu: Can remove?

server/push_test.go
```diff
@@ -1665,4 +1700,48 @@ func TestPush_MultipartReturnMultiSession(t *testing.T) {
 	assert.Equal(1, i)
 	assert.Equal(uint64(12), cxn.sourceBytes)
 	assert.Equal(uint64(32), cxn.transcodedBytes)
+
+	// now make unverified to respond with bash hash
```
yondonfu:
```suggestion
	// now make unverified to respond with bad hash
```

yondonfu (commented):
Just a few nits, but looks good!

server/push_test.go
leszko: done

leszko (commented):
Thanks for the comments @yondonfu . I addressed them. PTAL.

yondonfu (approved):
LGTM! Let's just cleanup the last few commits either via squashing and/or a rebase + push prior to merging.

leszko:
> LGTM! Let's just cleanup the last few commits either via squashing and/or a rebase + push prior to merging.

Yeah, sure I'll clean it up

PR #2110, opened 2021-11-18
server: Use http.Transport for HTTP client

yondonfu:

Replace http2.Transport to have option to downgrade from http2 if needed.

We've observed what appears to be dead locks with multiple concurrent HTTP requests during segment submission from a B to O. According to https://github.com/golang/go/issues/32388 there was a dead lock bug fixed in the http2 package that was backported into the latest go1.16 and go1.17. go-livepeer is currently built using go1.15. While we could upgrade to one of the go versions with this fix, we should probably do more testing when bumping to those go versions. So, this PR instead allows http2 to be disabled which should in theory bypass the dead lock http2 bug.

See commit history.

- Ran a B with `GODEBUG=http2debug=2` and observed http2 logs when B submits a segment to O.
- Ran a B with `GODEBUG=http2debug=2,http2client=0` and did not observe http2 logs when B submits a segment to O.

Note: We've yet to test whether running with `GODEBUG=http2client=0` with these changes actually resolves the issue we observed with dead locking concurrent HTTP requests and it is difficult to repro that.

N/A

**Checklist:**

### DIFF
--- a/server/segment_rpc.go
+++ b/server/segment_rpc.go
@@ -22,7 +22,6 @@ import (
 	"github.com/livepeer/go-livepeer/net"
 	"github.com/livepeer/lpms/ffmpeg"
 	"github.com/livepeer/lpms/stream"
-	"golang.org/x/net/http2"
 
 	ethcommon "github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/crypto"
@@ -44,15 +43,18 @@ var errProfile = errors.New("unrecognized encoder profile")
 var errDuration = errors.New("invalid duration")
 var errCapCompat = errors.New("incompatible capabilities")
 
+var dialTimeout = 2 * time.Second
+
 var tlsConfig = &tls.Config{InsecureSkipVerify: true}
 var httpClient = &http.Client{
-	Transport: &http2.Transport{
+	Transport: &http.Transport{
 		TLSClientConfig: tlsConfig,
-		DialTLS: func(network, addr string, cfg *tls.Config) (gonet.Conn, error) {
-			netDialer := &gonet.Dialer{
-				Timeout: 2 * time.Second,
-			}
-			return tls.DialWithDialer(netDialer, network, addr, cfg)
+		DialTLSContext: func(ctx context.Context, network, addr string) (gonet.Conn, error) {
+			cctx, cancel := context.WithTimeout(ctx, dialTimeout)
+			defer cancel()
+
+			tlsDialer := &tls.Dialer{Config: tlsConfig}
+			return tlsDialer.DialContext(cctx, network, addr)
 		},
 	},
 	// Don't set a timeout here; pass a context to the request
--- a/server/segment_rpc.go
+++ b/server/segment_rpc.go
@@ -56,6 +56,11 @@ var httpClient = &http.Client{
 			tlsDialer := &tls.Dialer{Config: tlsConfig}
 			return tlsDialer.DialContext(cctx, network, addr)
 		},
+		// Required for the transport to try to upgrade to HTTP/2 if TLSClientConfig is non-nil or
+		// if custom dialers (i.e. via DialTLSContext) are used. This allows us to by default
+		// transparently support HTTP/2 while maintaining the flexibility to use HTTP/1 by running
+		// with GODEBUG=http2client=0
+		ForceAttemptHTTP2: true,
 	},
 	// Don't set a timeout here; pass a context to the request
 }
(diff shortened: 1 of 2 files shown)


server/segment_rpc.go
```diff
@@ -46,14 +45,15 @@ var errCapCompat = errors.New("incompatible capabilities")
...
 		TLSClientConfig: tlsConfig,
-		DialTLS: func(network, addr string, cfg *tls.Config) (gonet.Conn, error) {
+		DialTLSContext: func(ctx context.Context, network, addr string) (gonet.Conn, error) {
 			netDialer := &gonet.Dialer{
 				Timeout: 2 * time.Second,
 			}
-			return tls.DialWithDialer(netDialer, network, addr, cfg)
+			return tls.DialWithDialer(netDialer, network, addr, tlsConfig)
 		},
+		ForceAttemptHTTP2: true,
```
yondonfu, line 63: According to the [http.Transport docs](https://pkg.go.dev/net/http#Transport) and [the http package code](https://cs.opensource.google/go/go/+/refs/tags/go1.17.3:src/net/http/transport.go;l=387), `ForceAttemptHTTP2: true` is needed for the transport to try to upgrade to HTTP/2 if `TLSClientConfig` is non-nil or if custom dialers (i.e. `DialTLSContext`) are used. By including this, we can default to transparently supporting HTTP/2 while maintaining the flexibility to use HTTP/1 by running with `GODEBUG=http2client=0`.

leszko: Nit: You can add a comment into the code for the explanation.

leszko (commented):
The change looks fine. I just have one question.

Since if I understand correctly, we encountered the bug you mentioned on prod in livepeer.com. And `x/net/http2` is used internally in `net/http`. Does it mean that we'll need to use `http2client=0` in livepeer.com prod to avoid this bug?

server/segment_rpc.go
```diff
@@ -46,14 +45,15 @@ var errCapCompat = errors.New("incompatible capabilities")
 
 var tlsConfig = &tls.Config{InsecureSkipVerify: true}
 var httpClient = &http.Client{
-	Transport: &http2.Transport{
+	Transport: &http.Transport{
 		TLSClientConfig: tlsConfig,
-		DialTLS: func(network, addr string, cfg *tls.Config) (gonet.Conn, error) {
+		DialTLSContext: func(ctx context.Context, network, addr string) (gonet.Conn, error) {
```
victorges, line 52: Any way to use this context here? Otherwise we may not respect the timeout during dial

server/segment_rpc.go
yondonfu: Fixed in ca94081

server/segment_rpc.go
yondonfu: Fixed in 5079f86

Now we set a dial timeout, but also allow the caller of `DialTLSContext` (i.e. the `http.Transport` internally) to cancel the context if it needs to by passing the context param to a `tlsDialer.DialContext()` call.

victorges: Great addition on an explicit dial timeout as well!

victorges (approved):
LGTM

server/segment_rpc.go
```diff
@@ -44,16 +43,24 @@ var errProfile = errors.New("unrecognized encoder profile")
...
-		DialTLS: func(network, addr string, cfg *tls.Config) (gonet.Conn, error) {
-			netDialer := &gonet.Dialer{
-				Timeout: 2 * time.Second,
-			}
-			return tls.DialWithDialer(netDialer, network, addr, cfg)
+		DialTLSContext: func(ctx context.Context, network, addr string) (gonet.Conn, error) {
+			cctx, cancel := context.WithTimeout(ctx, dialTimeout)
+			defer cancel()
+
+			tlsDialer := &tls.Dialer{Config: tlsConfig}
```
victorges, line 56: **nit:** Should this be shared among multiple connections (declared on the same lvl as `tlsConfig` and `httpClient` or is it ok/advised to instantiate a new one for every dial?

yondonfu: I'm not 100% sure, but I haven't seen any information recommending re-using a single dialer or indicating that instantiating a new dialer for each dial could result in inefficiencies. One data point here is that the [code for tls.Dial](https://cs.opensource.google/go/go/+/refs/tags/go1.17.3:src/crypto/tls/tls.go;l=171) actually instantiates a new `net.Dialer` (not `tls.Dialer`, but I imagine both types would have the same considerations) under the hood whenever it is called so that makes me feel like instantiating a new `tls.Dialer` in our case is probably fine.

darkdarkdragon (approved):
LGTM

yondonfu:
@leszko

> Since if I understand correctly, we encountered the bug you mentioned on prod in livepeer.com. And x/net/http2 is used internally in net/http. Does it mean that we'll need to use http2client=0 in livepeer.com prod to avoid this bug?

Yeah that's the idea for now. And once we upgrade to the latest Go version (along with completing the necessary testing and review of the changelog) with the bug fix in the http2 package we could stop using GODEBUG=http2client altogether.

leszko (approved):

PR #2111, opened 2021-11-19
eth: Ensure `maxFeePerGas` in all transactions never exceed `-masGasPrice` defined by the user

### LINKED ISSUE #2084: Set -maxGasPrice flag equal to max fee per gas, current implementation makes spending unpredictable
I really would suggest to implement the `-maxGasPrice` as setting the max fee per gas. The current implementation makes it really unpredictable how much gas you'll end up spending. As an example: I just had a ticket redeemed at 155 gwei despite having set the maxGasPrice to 110: https://etherscan.io/tx/0x3b07dbf395c69fb028a63d69596c50d1b4c268aa5f4f417850089b9f69c5baea
Here's what happened: 
- In block 13536342, the base fee was at 109, which triggered the transaction. 
- This block and the next 2 were full, which caused the base fee to increase 12.5% per block. 
- Now since the tx was submitted with a priority fee of only 1 gwei, it is unlikely to be included during high demand/when blocks are full. I guess I was lucky since block 13536345 was almost empty, so my 1 gwei priority tx was included. 
- However, by then the base fee was at 155 gwei which means I ended up paying almost 50% more than expected.

With the current implementation, you could easily end up paying twice as much as your `-maxGasPrice` setting, which seems very counterintuitive to me.

leszko:

Limit the maxFeePerGas in the eth transaction to the `-maxGasPrice` value defined by the user. Before it was possible that the user paid more than they specified as `-maxGasPrice`.

Additionally, I removed the `calcGasPrice()` logic from `transactionManager.go`, because I believe it was incorrect. It assumed that `maxFeePerGas` will be calculated in the [go-ethereum code](https://github.com/ethereum/go-ethereum/blob/fa9671851272129c9cba90ed4385d5b0558d34a1/accounts/abi/bind/base.go#L254),  but I think we override it in our code [here](https://github.com/livepeer/go-livepeer/blob/master/eth/transactionManager.go#L254).

- Adjust `GasFeeCap` before sending the transaction
- Remove `calcGasPrice` function in `replace()`

I did test the change both locally and using Rinkeby.

**Local testing**

1. [Set up devtool environment](https://github.com/livepeer/go-livepeer/tree/master/cmd/devtool)
2. Start orchestrator and wait for the first transaction to happen

```

Invoking transaction: "rewardWithHint". Inputs: "_newPosPrev: 0x0000000000000000000000000000000000000000  _newPosNext: 0x0000000000000000000000000000000000000000"  Hash: "0x175ff3b3eaebbabc0014526a8dcac53d9505129cbec0f33122da3c9ee70eb509". 

```

3. Check the transaction, note the default value of `maxFeePerGas`: `1000000014` (`2 * base fee + priority fee`)

```
$ seth tx 0x175ff3b3eaebbabc0014526a8dcac53d9505129cbec0f33122da3c9ee70eb509
accessList              []
blockHash               0xc880bb7bb980bc6e5de23976a20b12d02ce8431c3af25cc4e35da0478bbf020e
...
maxFeePerGas            1000000014
...
```

4. Set `max gas price` (via livepeer CLI) to `1000000013`
5. Wait for the next transaction
6. Check the transaction, note the value `maxFeePerGas` is set to `1000000013`

```
$ seth tx 0x2222d3d7a1026fe0712352e482a55cac167d491b3f4b17b05467bd5a92de08b1
accessList              []
blockHash               0xf8729bb0ba9b6925ca77bf03817fb648985e304cde08bf6ab16814146ba4e12a
...
maxFeePerGas            1000000013
...
```

7. Set `max gas price` to a value which is too low, e.g. `5`
8. Note that the transaction is not sent

```
E1125 12:57:56.999314    8399 roundinitializer.go:87] current gas price exceeds maximum gas price max=0.000000005 GWei current=1.000000007 GWei
```

**Rinkeby testing**

1. Start livepeer connected to `rinkeby`
2. Invoke any transaction (e.g. in case of broadcaster, deposit ETH)

```

Invoking transaction: "fundDepositAndReserve". Inputs: "_depositAmount: 100000000000000000  _reserveAmount: 300000000000000000"  Hash: "0x382e9a5b7635862fd8a5ac64ddf7e1b1d6111c4b67cb1318c13d7fede3fa31f0". 

```

3. Check the transaction in Etherscan

https://rinkeby.etherscan.io/tx/0x382e9a5b7635862fd8a5ac64ddf7e1b1d6111c4b67cb1318c13d7fede3fa31f0

```
Base: 0.000000024 Gwei |Max: 1.000000044 Gwei |Max Priority: 1 Gwei
```

4. Set max gas price to `1000000055` and run the transaction again

```

Invoking transaction: "fundDepositAndReserve". Inputs: "_depositAmount: 1000000000000000  _reserveAmount: 0"  Hash: "0xc6cfc72bb7aef2e4f026332ebcf9eec00ba10b4a0a94290d7c59a3bc0d1f8e9f". 

```

5. Check the transaction in Etherscan

https://rinkeby.etherscan.io/tx/0xc6cfc72bb7aef2e4f026332ebcf9eec00ba10b4a0a94290d7c59a3bc0d1f8e9f

```
Base: 0.00000002 Gwei |Max: 1.000000055 Gwei |Max Priority: 1 Gwei
```

fix #2084 

**Checklist:**

### DIFF
--- a/eth/transactionManager.go
+++ b/eth/transactionManager.go
@@ -101,7 +101,6 @@ func (tm *TransactionManager) SendTransaction(ctx context.Context, tx *types.Tra
 	adjustedRawTx := tm.newAdjustedTx(tx)
 	adjustedTx, err := tm.sig.SignTx(adjustedRawTx)
 	if err != nil {
-		glog.Infof("%s\n", err.Error())
 		return err
 	}
 
--- a/eth/transactionManager.go
+++ b/eth/transactionManager.go
@@ -98,9 +98,16 @@ func NewTransactionManager(eth transactionSenderReader, gpm *GasPriceMonitor, si
 }
 
 func (tm *TransactionManager) SendTransaction(ctx context.Context, tx *types.Transaction) error {
-	sendErr := tm.eth.SendTransaction(ctx, tx)
+	adjustedRawTx := tm.newAdjustedTx(tx)
+	adjustedTx, err := tm.sig.SignTx(adjustedRawTx)
+	if err != nil {
+		glog.Infof("%s\n", err.Error())
+		return err
+	}
 
-	txLog, err := newTxLog(tx)
+	sendErr := tm.eth.SendTransaction(ctx, adjustedTx)
+
+	txLog, err := newTxLog(adjustedTx)
 	if err != nil {
 		txLog.method = "unknown"
 	}
@@ -112,11 +119,11 @@ func (tm *TransactionManager) SendTransaction(ctx context.Context, tx *types.Tra
 
 	// Add transaction to queue
 	tm.cond.L.Lock()
-	tm.queue.add(tx)
+	tm.queue.add(adjustedTx)
 	tm.cond.L.Unlock()
 	tm.cond.Signal()
 
-	glog.Infof("\n%vEth Transaction%v\n\nInvoking transaction: \"%v\". Inputs: \"%v\"  Hash: \"%v\". \n\n%v\n", strings.Repeat("*", 30), strings.Repeat("*", 30), txLog.method, txLog.inputs, tx.Hash().String(), strings.Repeat("*", 75))
+	glog.Infof("\n%vEth Transaction%v\n\nInvoking transaction: \"%v\". Inputs: \"%v\"  Hash: \"%v\". \n\n%v\n", strings.Repeat("*", 30), strings.Repeat("*", 30), txLog.method, txLog.inputs, adjustedTx.Hash().String(), strings.Repeat("*", 75))
 
 	return nil
 }
@@ -159,7 +166,7 @@ func (tm *TransactionManager) replace(tx *types.Transaction) (*types.Transaction
 
 	// Bump gas price exceeds max gas price, return early
 	max := tm.gpm.MaxGasPrice()
-	newGasPrice := calcGasPrice(newRawTx)
+	newGasPrice := newRawTx.GasFeeCap()
 	if max != nil && newGasPrice.Cmp(max) > 0 {
 		return nil, fmt.Errorf("replacement gas price exceeds max gas price suggested=%v max=%v", newGasPrice, max)
 	}
@@ -234,30 +241,37 @@ func (tm *TransactionManager) checkTxLoop() {
 	}
 }
 
+func (tm *TransactionManager) newAdjustedTx(tx *types.Transaction) *types.Transaction {
+	baseTx := newDynamicFeeTx(tx)
+	if tm.gpm.MaxGasPrice() != nil {
+		baseTx.GasFeeCap = tm.gpm.MaxGasPrice()
+	}
+
+	return types.NewTx(baseTx)
+}
+
 func applyPriceBump(val *big.Int, priceBump uint64) *big.Int {
 	a := big.NewInt(100 + int64(priceBump))
 	b := new(big.Int).Mul(a, val)
 	return b.Div(b, big.NewInt(100))
 }
 
-// Calculate the gas price as gas tip cap + base fee
-func calcGasPrice(tx *types.Transaction) *big.Int {
-	// Assume that the gas fee cap is calculated as gas tip cap + (baseFee * 2)
-	baseFee := new(big.Int).Div(new(big.Int).Sub(tx.GasFeeCap(), tx.GasTipCap()), big.NewInt(2))
-	return new(big.Int).Add(baseFee, tx.GasTipCap())
+func newReplacementTx(tx *types.Transaction) *types.Transaction {
+	baseTx := newDynamicFeeTx(tx)
+	baseTx.GasFeeCap = applyPriceBump(tx.GasFeeCap(), priceBump)
+	baseTx.GasTipCap = applyPriceBump(tx.GasTipCap(), priceBump)
+
+	return types.NewTx(baseTx)
 }
 
-func newReplacementTx(tx *types.Transaction) *types.Transaction {
-	baseTx := &types.DynamicFeeTx{
-		Nonce: tx.Nonce(),
-		// geth requires the price bump to be applied to both the gas tip cap and gas fee cap
-		GasFeeCap: applyPriceBump(tx.GasFeeCap(), priceBump),
-		GasTipCap: applyPriceBump(tx.GasTipCap(), priceBump),
+func newDynamicFeeTx(tx *types.Transaction) *types.DynamicFeeTx {
+	return &types.DynamicFeeTx{
+		To:        tx.To(),
+		Nonce:     tx.Nonce(),
+		GasFeeCap: tx.GasFeeCap(),
+		GasTipCap: tx.GasTipCap(),
 		Gas:       tx.Gas(),
 		Value:     tx.Value(),
 		Data:      tx.Data(),
-		To:        tx.To(),
 	}
-
-	return types.NewTx(baseTx)
 }
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -9,6 +9,7 @@
 #### General
 
 - \#2114 Add option to repeat the benchmarking process (@jailuthra)
+- \#2111 Ensure `maxFeePerGas` in all transactions never exceed `-masGasPrice` defined by the user (@leszko)
 
 #### Broadcaster
 
--- a/doc/ethereum.md
+++ b/doc/ethereum.md
@@ -14,12 +14,13 @@ The round initialization service is disabled by default and can be enabled by st
 
 ## Gas Prices
 
-After the EIP-1559 upgrade on Ethereum, the node uses `maxFeePerGas` parameter to limit the maximum fee paid for the transaction execution:
-
-- If max gas price is specified, then use its value as `maxFeePerGas`
-- If max gas price is not specified, then use the default `maxFeePerGas` and gradually increase it with the replacement transactions
+After the EIP-1559 upgrade on Ethereum, the node treats the gas price as priority fee + base fee.
 
 ### Max gas price
+
+The `maxGasPrice` parameter makes sure the transaction fee never exceeds the specified limit.
+- If the current network gas price is higher than `maxGasPrice`, the transaction is not sent
+- The transaction parameter `maxFeePerGas` is set to `maxGasPrice`
 
 The following options can be used to get the max gas price:
 
--- a/doc/ethereum.md
+++ b/doc/ethereum.md
@@ -14,7 +14,10 @@ The round initialization service is disabled by default and can be enabled by st
 
 ## Gas Prices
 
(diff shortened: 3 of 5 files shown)


yondonfu:
Reaction based on a quick skim:

I think this draft PR is going in the right direction. My understanding of the transaction flow after these changes:

- Contract binding triggers a call to [transact()](https://github.com/livepeer/go-ethereum/blob/80e3965c7b18f1ae8f065c9bf570b36f5e5c4cb7/accounts/abi/bind/base.go#L236)
- `transact()` sets the [GasFeeCap on the tx to 2x base fee + GasTipCap](https://github.com/livepeer/go-ethereum/blob/80e3965c7b18f1ae8f065c9bf570b36f5e5c4cb7/accounts/abi/bind/base.go#L274)
- `transact()` passes the created tx to [SendTransaction()](https://github.com/livepeer/go-ethereum/blob/80e3965c7b18f1ae8f065c9bf570b36f5e5c4cb7/accounts/abi/bind/base.go#L349) which is actually implemented by the TransactionManager
- The `SendTransaction()` method on TransactionManager uses `newAdjustedTx()` to adjust the GasFeeCap on the tx if needed - if it is less than than `GasPriceMonitor.MaxGasPrice()` then we keep it, otherwise we set the GasFeeCap to `GasPriceMonitor.MaxGasPrice()`. As a result the GasFeeCap never exceeds the value of `GasPriceMonitor.MaxGasPrice()`

leszko:
Thanks for the quick feedback @yondonfu . I understand it exactly the same as you do. So, I'll perform the testing and if everything works fine, I'll update this draft PR.

eth/transactionManager.go
```diff
@@ -98,9 +98,16 @@ func NewTransactionManager(eth transactionSenderReader, gpm *GasPriceMonitor, si
 }
 
 func (tm *TransactionManager) SendTransaction(ctx context.Context, tx *types.Transaction) error {
-	sendErr := tm.eth.SendTransaction(ctx, tx)
+	adjustedRawTx := tm.newAdjustedTx(tx)
+	adjustedTx, err := tm.sig.SignTx(adjustedRawTx)
+	if err != nil {
+		glog.Infof("%s\n", err.Error())
```
yondonfu: Do we need this log if we're just returning the error to the caller here?

eth/transactionManager.go
```diff
@@ -234,30 +241,37 @@ func (tm *TransactionManager) checkTxLoop() {
 	}
 }
 
+func (tm *TransactionManager) newAdjustedTx(tx *types.Transaction) *types.Transaction {
+	baseTx := newDynamicFeeTx(tx)
+	if tm.gpm.MaxGasPrice() != nil {
+		baseTx.GasFeeCap = tm.gpm.MaxGasPrice()
```
yondonfu, line 246: I forget if we talked about this, but is there a reason for always setting this to the max gas price as opposed to only setting it to the max gas price if the max gas price exceeds the current value?

eth/transactionManager.go
leszko: You're right. Updated!

eth/transactionManager.go
leszko: Both approaches will work. However, IMO the always setting max gas price is slightly better because:
1. It's simpler for the user what happens.
2. Since the replacement transaction needs to be at least 10% higher, we may end up never reaching exactly `maxGasPrice`.

What do you think?

leszko (commented):
@yondonfu I addressed your comemnts. PTAL

eth/transactionManager.go
yondonfu: Sounds reasonable. But, this also means that a replacement tx would never be sent if the user sets a `maxGasPrice` because `GasFeeCap` is always set to `maxGasPrice` meaning that we are never able to increase `GasFeeCap` for a replacement tx. And the only scenario where a replacement tx could be sent is if the user does not set a `maxGasPrice`. I think this is okay, but just want to make sure we're on the same page about the expected behavior.

yondonfu (commented):
As of this PR, I think the `-maxGasPrice` flag is basically equivalent to `maxFeePerGas` since its value is always set as `GasTipCap` for transactions. The [gas price check](https://github.com/livepeer/go-livepeer/pull/2013/commits/e26b20fa7ef85c78f5933632fcac9ebbfd5da16e#diff-9cc7046826ce92d9cc5432cf41ed4fc14bb6907da44f771e3932302c40e8dbc2L103) in `backend.SuggestGasPrice()` can be interpreted as:

- The gas price cached by the GasPriceMonitor could be interpreted as "a suggested maxFeePerGas" i.e. the current base fee + a suggested priority fee
- We compare the user provided `maxFeePerGas` (via `-maxGasPrice`) against this suggested maxFeePerGas

If this is the case, I think we should make some small updates to docs in a few places:

- https://github.com/livepeer/go-livepeer/pull/2013/commits/df3bc3b5b2e622bc4beba930a72440261db81ba8
- https://github.com/livepeer/go-livepeer/pull/2013/commits/57e985c0143447785c06b7c1987fa01a3306a78a

eth/transactionManager.go
leszko: You're right. Actually, there is one more scenario for the replacement if the user manually increases `maxGasPrice` while the transaction is pending. Then the replacement can also happen.

leszko:
> As of this PR, I think the `-maxGasPrice` flag is basically equivalent to `maxFeePerGas` since its value is always set as `GasTipCap` for transactions. 

That is correct.

> The [gas price check](https://github.com/livepeer/go-livepeer/pull/2013/commits/e26b20fa7ef85c78f5933632fcac9ebbfd5da16e#diff-9cc7046826ce92d9cc5432cf41ed4fc14bb6907da44f771e3932302c40e8dbc2L103) in `backend.SuggestGasPrice()` can be interpreted as:
> 
> * The gas price cached by the GasPriceMonitor could be interpreted as "a suggested maxFeePerGas" i.e. the current base fee + a suggested priority fee
> * We compare the user provided `maxFeePerGas` (via `-maxGasPrice`) against this suggested maxFeePerGas

That is not correct. "suggested gas price" from GasPriceMonitor is not interpreted as `maxFeePerGas`. If `maxGasPrice` is not set by the user, then `maxFeePerGas` is set to the default value `2 * base fee + tip cap`. I believe that `backend.SuggestGasPrice()` is not used at all. It was used before EIP-1559.

You made me think we should make some cleanup, but I believe it should be a separate PR, not to mix to many things here.

> If this is the case, I think we should make some small updates to docs in a few places:
- I updated the doc here: https://github.com/livepeer/go-livepeer/pull/2111/commits/ff84decd1473c415271cdb341e693676dbb51b4c
- I believe the param description in `livepeer.go` is correct, it says `"Maximum gas price (priority fee + base fee) for ETH transactions in wei, 40 Gwei = 40000000000"`

leszko (commented):
@yondonfu Added my comments. PTAL

yondonfu:
> I believe that backend.SuggestGasPrice() is not used at all.

That method is actually called in [backend.SuggestGasTipCap()](https://github.com/livepeer/go-livepeer/blob/f3a299bb95af04004fd957cdc3b2b4a69c72c90b/eth/backend.go#L116) right now. And `backend.SuggestGasTipCap()` is called when a transaction is created for a contract binding if the `GasTipCap` is [not explicitly specified in the transact opts](https://github.com/ethereum/go-ethereum/blob/721c5723c0e31c4d764c61ee99c505d18466b363/accounts/abi/bind/base.go#L243) used. This means `backend.SuggestGasPrice()` is called every time a transaction is created since we never explicitly set the `GasTipCap` in the transact opts in order to always use the suggested `GasTipCap` based on current ETH network transaction activity.

> "suggested gas price" from GasPriceMonitor is not interpreted as maxFeePerGas. If maxGasPrice is not set by the user, then maxFeePerGas is set to the default value 2 * base fee + tip cap

Yeah I was a bit off there.

So, the `GasPriceMonitor.GasPrice()` method will return the cached value for the latest gas price returned from an ETH node which is really base fee + suggested priority fee (i.e. `GasTipCap`). If `-maxGasPrice` is set, `maxGasPrice = maxFeePerGas` so `backend.SuggestGasPrice()` is comparing the latest base fee + suggested priority fee against maxFeePerGas. 

I was just thinking about how to describe what "gas price" means. I think the way to describe this is:

- Gas price = base fee + priority fee
- Max gas price = maximum value of base fee + priority fee aka maxFeePerGas
- Min gas price = minimum value of base fee + priority fee

How does that sound to you?

> I believe the param description in livepeer.go is correct, it says

Okay yeah so if `-maxGasPrice` is maxFeePerGas then that description should still be correct.

leszko:
> That method is actually called in [backend.SuggestGasTipCap()](https://github.com/livepeer/go-livepeer/blob/f3a299bb95af04004fd957cdc3b2b4a69c72c90b/eth/backend.go#L116) right now. And `backend.SuggestGasTipCap()` is called when a transaction is created for a contract binding if the `GasTipCap` is [not explicitly specified in the transact opts](https://github.com/ethereum/go-ethereum/blob/721c5723c0e31c4d764c61ee99c505d18466b363/accounts/abi/bind/base.go#L243) used. This means `backend.SuggestGasPrice()` is called every time a transaction is created since we never explicitly set the `GasTipCap` in the transact opts in order to always use the suggested `GasTipCap` based on current ETH network transaction activity.

This is correct with one comment. The function `backend.SuggestGasPrice()` is called inside in every transaction, but its result is not usually used. Its result is used only if calling to Eth endpoint `eth_maxPriorityFeePerGas` [results in an error](https://github.com/livepeer/go-livepeer/blob/10c8f1340b1704eaef4a2efd3fd906c2e2610908/eth/backend.go#L122). That means that in most cases (happy path), we don't use `backend.SuggestGasPrice()`.

> So, the `GasPriceMonitor.GasPrice()` method will return the cached value for the latest gas price returned from an ETH node which is really base fee + suggested priority fee (i.e. `GasTipCap`). If `-maxGasPrice` is set, `maxGasPrice = maxFeePerGas` so `backend.SuggestGasPrice()` is comparing the latest base fee + suggested priority fee against maxFeePerGas.

Yes, I think it's right. `gas price` is the value from the Eth endpoint `eth_gasPrice`, which might be `base fee + suggested priority` as you wrote.

> I was just thinking about how to describe what "gas price" means. I think the way to describe this is:
> 
> * Gas price = base fee + priority fee
> * Max gas price = maximum value of base fee + priority fee aka maxFeePerGas
> * Min gas price = minimum value of base fee + priority fee

I'm not sure what's the meaning of `gas price` for the user if we don't set it anywhere in the transaction. And it's not the Livepeer parameter, maybe we should not describe it at all 🤔

leszko:
@yondonfu I updated the doc again after our conversation. PTAL.

yondonfu (approved):
LGTM!

PR #2122, opened 2021-12-01
1656 upload timeouts

### LINKED ISSUE #1656: B <> O upload timeouts

At the moment, we use a single HTTP timeout for B <> O segment submission. The timer for this timeout starts when B [submits a request to /segment](https://github.com/livepeer/go-livepeer/blob/0b8f3696640e28f0c74d3e08088df900af9ed63f/server/segment_rpc.go#L394) and ends when B finishes reading the response body returned by O. 

This timeout covers:
- Segment upload time to O
- Transcode time on O
- Response time on O [1]

[1] Note rendition download time is not included here since the response from O to segment submission includes the URLs for the renditions which are downloaded separately.

If there are networking/bandwidth issues with O causing slow uploads, B could end up waiting the full HTTP timeout (defaults to 4x the segment duration) before it fails over to a different O. There is an opportunity to detect the networking/bandwidth issue with O earlier if B defined more granular timeouts in addition to the HTTP timeout for segment submission.

B can define an upload timeout which is strictly shorter than the segment submission HTTP timeout. The upload starts when B submits a request to /segment and ends when B receives a response status from O [2]. If this upload exceeds the defined timeout, then B can fail over to a different O instead of waiting the full segment submission HTTP timeout. The upload timeout might be defined as `segment duration * uploadTimeoutMultiplier` where `uploadTimeoutMultiplier` could default to something like 0.5 such that a 2 second segment has a 1 second upload timeout.

[Code sketch](https://github.com/livepeer/go-livepeer/compare/yf/upload-time-out)

[2] The transcoding protocol has O send back a response status to acknowledge that the upload completed *before* it starts sending back data in the response body which only begins when it completes transcoding.

leszko:
While making an HTTP request from B -> O, introduce a separate timeout for the segment data upload. Before there was one timeout for the whole request.

- Add timeout for segment upload + unit tests

Check locally if transcoding (still) works. Then, introduced an artificial sleep in O and tried again. Note the timeout error message in B.

fix #1656 

**Checklist:**

### DIFF
--- a/server/segment_rpc.go
+++ b/server/segment_rpc.go
@@ -830,21 +830,18 @@ func validatePrice(sess *BroadcastSession) error {
 }
 
 func sendReqWithTimeout(req *http.Request, timeout time.Duration) (*http.Response, error) {
-	type result struct {
-		resp *http.Response
-		err  error
-	}
-	resCh := make(chan result)
+	ctx, cancel := context.WithCancel(req.Context())
+	timeouter := time.AfterFunc(timeout, cancel)
 
-	go func() {
-		resp, err := httpClient.Do(req)
-		resCh <- result{resp, err}
-	}()
-
-	select {
-	case res := <-resCh:
-		return res.resp, res.err
-	case <-time.After(timeout):
-		return nil, errTimeout
+	req = req.WithContext(ctx)
+	resp, err := httpClient.Do(req)
+	if timeouter.Stop() {
+		return resp, err
 	}
+	// timeout has already fired and cancelled the request
+	if err != nil {
+		return nil, err
+	}
+	resp.Body.Close()
+	return nil, context.DeadlineExceeded
 }
--- a/server/segment_rpc.go
+++ b/server/segment_rpc.go
@@ -42,6 +42,7 @@ var errFormat = errors.New("unrecognized profile output format")
 var errProfile = errors.New("unrecognized encoder profile")
 var errDuration = errors.New("invalid duration")
 var errCapCompat = errors.New("incompatible capabilities")
+var errTimeout = errors.New("timeout")
 
 var dialTimeout = 2 * time.Second
 
@@ -444,6 +445,7 @@ func SubmitSegment(sess *BroadcastSession, seg *stream.HLSSegment, nonce uint64,
 	defer cancel()
 
 	ti := sess.OrchestratorInfo
+
 	req, err := http.NewRequestWithContext(ctx, "POST", ti.Transcoder+"/segment", bytes.NewBuffer(data))
 	if err != nil {
 		glog.Errorf("Could not generate transcode request to orch=%s", ti.Transcoder)
@@ -465,7 +467,8 @@ func SubmitSegment(sess *BroadcastSession, seg *stream.HLSSegment, nonce uint64,
 
 	glog.Infof("Submitting segment nonce=%d manifestID=%s sessionID=%s seqNo=%d bytes=%v orch=%s timeout=%s", nonce, params.ManifestID, sess.OrchestratorInfo.AuthToken.SessionId, seg.SeqNo, len(data), ti.Transcoder, dur)
 	start := time.Now()
-	resp, err := httpClient.Do(req)
+	uploadTimeout := time.Duration(0.5 * seg.Duration * float64(time.Second))
+	resp, err := sendReqWithTimeout(req, uploadTimeout)
 	uploadDur := time.Since(start)
 	if err != nil {
 		glog.Errorf("Unable to submit segment orch=%v nonce=%d manifestID=%s sessionID=%s seqNo=%d orch=%s err=%v", ti.Transcoder, nonce, params.ManifestID, sess.OrchestratorInfo.AuthToken.SessionId, seg.SeqNo, ti.Transcoder, err)
@@ -815,4 +818,24 @@ func validatePrice(sess *BroadcastSession) error {
 		return fmt.Errorf("Orchestrator price higher than the set maximum price of %v wei per %v pixels", maxPrice.Num().Int64(), maxPrice.Denom().Int64())
 	}
 	return nil
+}
+
+func sendReqWithTimeout(req *http.Request, timeout time.Duration) (*http.Response, error) {
+	type result struct {
+		resp *http.Response
+		err  error
+	}
+	resCh := make(chan result)
+
+	go func() {
+		resp, err := httpClient.Do(req)
+		resCh <- result{resp, err}
+	}()
+
+	select {
+	case res := <-resCh:
+		return res.resp, res.err
+	case <-time.After(timeout):
+		return nil, errTimeout
+	}
 }
--- a/server/segment_rpc.go
+++ b/server/segment_rpc.go
@@ -447,7 +447,7 @@ func SubmitSegment(sess *BroadcastSession, seg *stream.HLSSegment, nonce uint64,
 	// timeout for the segment upload, until HTTP returns OK 200
 	uploadTimeout := time.Duration(segUploadTimeoutMultiplier * seg.Duration * float64(time.Second))
 	if uploadTimeout <= 0 {
-		uploadTimeout = httpTimeout
+		uploadTimeout = common.SegmentUploadTimeout
 	}
 
 	ctx, cancel := context.WithTimeout(context.Background(), httpTimeout)
--- a/server/segment_rpc.go
+++ b/server/segment_rpc.go
@@ -45,7 +45,6 @@ var errFormat = errors.New("unrecognized profile output format")
 var errProfile = errors.New("unrecognized encoder profile")
 var errDuration = errors.New("invalid duration")
 var errCapCompat = errors.New("incompatible capabilities")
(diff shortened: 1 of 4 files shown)


darkdarkdragon (approved):
LGTM

server/segment_rpc.go
```diff
@@ -434,16 +435,24 @@ func SubmitSegment(sess *BroadcastSession, seg *stream.HLSSegment, nonce uint64,
...
-	dur := common.HTTPTimeout
 	paddedDur := 4.0 * seg.Duration // use a multiplier of 4 for now
-	if paddedDur > dur.Seconds() {
-		dur = time.Duration(paddedDur * float64(time.Second))
+	if paddedDur > httpTimeout.Seconds() {
+		httpTimeout = time.Duration(paddedDur * float64(time.Second))
 	}
-	ctx, cancel := context.WithTimeout(context.Background(), dur)
+	// timeout fot the segment upload, until HTTP returns OK 200
+	uploadTimeout := time.Duration(0.5 * seg.Duration * float64(time.Second)) // use 0.5 multiplier
```
victorges: I think this comment is unnecessary, since the code already states that we are multiplying by `0.5`. It's similar to the "multipler of 4" above which I also think could be removed IMO.
```suggestion
	uploadTimeout := time.Duration(0.5 * seg.Duration * float64(time.Second))
```

We can create `consts` for it to give them a name and make it more explicit, which I think would be better than these comments in the end. Maybe `segHttpPushTimeoutMultipler` and `segUploadTimeoutMultiplier`?

server/segment_rpc.go
```diff
@@ -434,16 +435,24 @@ func SubmitSegment(sess *BroadcastSession, seg *stream.HLSSegment, nonce uint64,
...
+	if uploadTimeout <= 0 {
```
victorges, line 449: How is this possible? I think `seg.Duration` is always positive.

Did you mean to have a minimum timeout here, say `1*second` perhaps? (then the line below is timeout = `min`, not `httpTimeout`)

server/segment_rpc.go
```diff
@@ -816,3 +825,23 @@ func validatePrice(sess *BroadcastSession) error {
...
+	go func() {
+		resp, err := httpClient.Do(req)
+		resCh <- result{resp, err}
+	}()
+
+	select {
+	case res := <-resCh:
+		return res.resp, res.err
+	case <-time.After(timeout):
+		return nil, errTimeout
```
victorges: This does not stop the request to continue to be sent, and may also create a resource leak in case the request finishes after we return from here. In that case, we would have:
 - 1 stuck goroutine trying to send on the `resCh` with no receiver on the other side. This can be fixed by adding a little buffer (`1`) to the `resCh` above so it doesn't block.
 - 1 `http.Response` object that won't be processed nor have it's body closed for disposal/connection re-use. We could try to fix this in some ways, like for example by not having the channel buffer at all, but instead having `select` case in the goroutine above, which `defaults` to just ignoring disposing the response in case no one is receiving on the other side.

I think both of these issues can be fixed in a different way tho, by inverting the logic of how we handle the cancellation goroutine here: Instead of doing the request in a separate go-routine and only waiting for the result in the current one, create a separate goroutine for handling the timeout/cancellation and do the main work on the main routine executing this function. For the cancellation, we use a custom context that we replace in the request in order to cancel it early (upload timeout) _only if the response headers haven't been received_. Sounds like a fun logic so let me give it a `go`:
```go
func sendReqWithTimeout(req *http.Request, timeout time.Duration) (*http.Response, error) {
	ctx, cancel := context.WithCancel(req.Context())
	timeouter := time.AfterFunc(timeout, cancel)

	req = req.WithContext(ctx)
	resp, err := httpClient.Do(req)
	if timeouter.Stop() {
		return resp, err
	}
	// timeout has already fired and cancelled the request
	if err != nil {
		return nil, err
	}
	resp.Body.Close()
	return nil, context.DeadlineExceeded
}
```

server/segment_rpc_test.go
```diff
@@ -2261,6 +2261,30 @@ func TestSubmitSegment_Success(t *testing.T) {
...
+	resp, err := sendReqWithTimeout(req, 5*time.Second)
+	assert.NoError(err)
+	assert.Equal(200, resp.StatusCode)
+
+	// timeout
+	wg.Add(1)
+	resp, err = sendReqWithTimeout(req, time.Millisecond)
+	wg.Done()
+	assert.Nil(resp)
+	assert.ErrorIs(err, errTimeout)
```
victorges: Another check to add here is whether the request continued executing on the server after the client cancellled the request. You could check that by, in the server handler, writing to some channel that you can read here. For example, write `r.Context().Err()` which would be **not nil** in case the request was really cancelled for the server or `nil` if it kept running in the end (which also means it kept running for the client of `sendReqWithTimeout`).

victorges (commented):
LGTM overall, but I think the core `sendReqWithTimeout` has some issues that could be improved.

server/segment_rpc.go
leszko: Yes, good idea. Updated in https://github.com/livepeer/go-livepeer/pull/2122/commits/b582d2531f254390197abe680f3af39857dadcd8

server/segment_rpc.go
leszko: I think `seg.Duration` can be `0` in some cases. Then, without this condition, in such scenarios, the code will panic.

server/segment_rpc.go
leszko: @victorges You're right. I think your solution is better. I applied it.

One question, you wrote:

> create a separate goroutine for handling the timeout/cancellation and do the main work on the main routine executing this function.

What do you mean? Did you mean that calling `sendReqWithTimeout()` should be in a separate goroutine? I guess there is no need for it.

server/segment_rpc_test.go
leszko: I think it's not so easy to test it in a reliable manner, because `cancel()` is called by `timeouter` asynchronously (in a separate routine). That means that we can't rely on when the `r.Context().Err()` is actually set to some value.

We could do the following, but I don't like that we use this `time.sleep()` everywhere. Or maybe you have some other idea how to test it?

```golang
func TestSendReqWithTimeout(t *testing.T) {
	assert := assert.New(t)

	var wg sync.WaitGroup
	ctxErrCh := make(chan error, 1)
	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		wg.Wait()
		w.WriteHeader(http.StatusOK)
		time.Sleep(100 * time.Millisecond)  // to reduce the flakiness of the test
		ctxErrCh <- r.Context().Err()
	}))
	defer server.Close()
	req, _ := http.NewRequestWithContext(context.Background(), "POST", server.URL, nil)

	// no timeout
	resp, err := sendReqWithTimeout(req, 5*time.Second)
	assert.NoError(err)
	assert.Equal(200, resp.StatusCode)
	assert.NoError(<-ctxErrCh)

	// timeout
	wg.Add(1)
	resp, err = sendReqWithTimeout(req, time.Millisecond)
	wg.Done()
	assert.Nil(resp)
	assert.ErrorIs(err, context.Canceled)
	assert.ErrorIs(<-ctxErrCh, context.Canceled)
}
```

leszko (commented):
Thanks for the review @victorges . Added my comments.

server/segment_rpc.go
victorges: > What do you mean? Did you mean that calling sendReqWithTimeout() should be in a separate goroutine? I guess there is no need for it.

Ah, no! I just meant that it would be easier to deal with the cancellation, resource disposal and errors if we call `httpClient.Do` directly from `sendReqWithTimeout`, instead of from a goroutine started from it that sent the result over a channel. I was trying to explain the changes that I ended up proposing with the implementation.

server/segment_rpc_test.go
victorges: Ah I see! I tried thinking of some way of guaranteeing the test consistency, but it does feel kind of hard given the concurrent nature of the problem here. Nevermind this, I think we can leave it as is.

server/segment_rpc.go
victorges: Ah I see... 

**nit:** Another possibility is to not have any upload timeout logic when the segment duration is `<= 0` then. e.g. start `sendReqWithTimeout` with something like `if timeout <= 0 { return httpClient.Do(req) }`

But I think we might always want to have a default upload timeout here in case the segment duration is undefined. WDYT of defining a `common.SegmentUploadTimeout` just like the `common.HTTPTimeout` above, which we use by default in case the segment duration is `<= 0` (instead of `== httpTimeout` which is a noop)

victorges (approved):
LGTM!

server/segment_rpc.go
leszko: Great, thanks 👍

server/segment_rpc.go
leszko: Good idea, added `common.SegmentUploadTimeout` and set it to `2s`

PR #2124, opened 2021-12-02
server: B should not retry transcoding if client closed/canceled the connection

### LINKED ISSUE #2113: B should cancel segment processing if the HTTP push client request ends

A HTTP push client may close its connection or cancel a request before B finishes processing a segment. At the moment, we [check if the request context is cancelled](https://github.com/livepeer/go-livepeer/blob/b96a0f5fc690740b8e2b8cc79d67992c7d90d8f3/server/mediaserver.go#L966), but only after B finishes processing a segment. As a result, B could continue processing a segment even though the client has already given up on the request and is no longer listening for the response.

Consider, the following example with Mist server as the HTTP push client:

- Mist sets the upload timeout for a segment to segment duration + 2s resulting in a 4s timeout for a 2s segment
- Mist sends the segment to B
- B starts processing the segment and gets stuck waiting for an upload to O
- Mist times out after 4s
- B times out waiting for an upload to O after 8s (the timeout is currently set to 4x the segment duration) and then retries 3 times meaning that it could spend 8s * 3 = 24s processing the segment if there are networking issues with each O
- B completes segment processing and is ready to return an error to Mist, but Mist already stopped listening for a response

If the HTTP push client cancels its request, B should immediately cancel segment processing because the client is no longer listening for a response.

Here is a sketch of one way we could implement this:

```golang
func processSegmentWithContext(ctx context.Context, cxn *rtmpConnection, seg *stream.HLSSegment) ([]string, error) {
    resCh := make(chan struct {
        urls []string
        err error
    })

    go func() {
        urls, err := processSegment(cxn, seg)
        resCh <- struct{urls []string; err error}{urls, err}
    }()

    select {
        case <-ctx.Done():
            return nil, ctx.Err()
        case res := <-resCh:
            return res.urls, res.err
    }
}
```

We may want to pass a context to `transcodeSegment()` as well?

N/A

**Additional context**

N/A

leszko:
B will take any more attempts of sending the segment to O if the client closed/canceled the request connection.

Before, even if the client was not interested in transcoding anymore, B tried transcoding until the `maxAttemps` number is reached.

- Use the request context to determine if B should retry the transcoding

Introduced `println` and sent requests from the script for two options:
- request is suddenly closed => observed no retries
- request is kept open => retries occurred

fix #2113

**Checklist:**

### DIFF
--- a/server/broadcast.go
+++ b/server/broadcast.go
@@ -811,7 +811,8 @@ func processSegment(ctx context.Context, cxn *rtmpConnection, seg *stream.HLSSeg
 			glog.Warningf("Not retrying current segment nonce=%d seqNo=%d due to non-retryable error err=%v", nonce, seg.SeqNo, err)
 			break
 		}
-		if err = ctx.Err(); err != nil {
+		if ctxErr := ctx.Err(); ctxErr != nil {
+			err = ctxErr
 			glog.Warningf("Not retrying current segment nonce=%d seqNo=%d due to error err=%v", nonce, seg.SeqNo, err)
 			break
 		}
--- a/server/broadcast.go
+++ b/server/broadcast.go
@@ -813,7 +813,7 @@ func processSegment(ctx context.Context, cxn *rtmpConnection, seg *stream.HLSSeg
 		}
 		if ctxErr := ctx.Err(); ctxErr != nil {
 			err = ctxErr
-			glog.Warningf("Not retrying current segment nonce=%d seqNo=%d due to error err=%v", nonce, seg.SeqNo, err)
+			glog.Warningf("Not retrying current segment nonce=%d seqNo=%d due to context cancellation err=%v", nonce, seg.SeqNo, err)
 			break
 		}
 		// recoverable error, retry
--- a/server/broadcast.go
+++ b/server/broadcast.go
@@ -677,7 +677,7 @@ func selectOrchestrator(n *core.LivepeerNode, params *core.StreamParameters, cou
 	return sessions, nil
 }
 
-func processSegment(cxn *rtmpConnection, seg *stream.HLSSegment) ([]string, error) {
+func processSegment(ctx context.Context, cxn *rtmpConnection, seg *stream.HLSSegment) ([]string, error) {
 
 	rtmpStrm := cxn.stream
 	nonce := cxn.nonce
@@ -809,6 +809,10 @@ func processSegment(cxn *rtmpConnection, seg *stream.HLSSegment) ([]string, erro
 		}
 		if isNonRetryableError(err) {
 			glog.Warningf("Not retrying current segment nonce=%d seqNo=%d due to non-retryable error err=%v", nonce, seg.SeqNo, err)
+			break
+		}
+		if err = ctx.Err(); err != nil {
+			glog.Warningf("Not retrying current segment nonce=%d seqNo=%d due to error err=%v", nonce, seg.SeqNo, err)
 			break
 		}
 		// recoverable error, retry
--- a/server/broadcast_test.go
+++ b/server/broadcast_test.go
@@ -771,21 +771,21 @@ func TestProcessSegment_MaxAttempts(t *testing.T) {
 
 	// Sanity check: zero attempts should not transcode
 	MaxAttempts = 0
-	_, err := processSegment(cxn, seg)
+	_, err := processSegment(context.Background(), cxn, seg)
 	assert.Nil(err)
 	assert.Equal(0, transcodeCalls, "Unexpectedly submitted segment")
 	assert.Len(bsm.trustedPool.sessMap, 2)
 
 	// One failed transcode attempt. Should leave another in the map
 	MaxAttempts = 1
-	_, err = processSegment(cxn, seg)
+	_, err = processSegment(context.Background(), cxn, seg)
 	assert.NotNil(err)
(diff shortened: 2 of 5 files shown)


server/broadcast.go
```diff
@@ -811,6 +811,11 @@ func processSegment(cxn *rtmpConnection, seg *stream.HLSSegment) ([]string, erro
 			glog.Warningf("Not retrying current segment nonce=%d seqNo=%d due to non-retryable error err=%v", nonce, seg.SeqNo, err)
 			break
 		}
+		if ctxErr := ctx.Err(); ctxErr != nil {
+			err = ctxErr
+			glog.Warningf("Not retrying current segment nonce=%d seqNo=%d due to error err=%v", nonce, seg.SeqNo, err)
```
victorges: **nit:** just for debugging swiftness, we can be a little more explicit here. WDYT?
```suggestion
			glog.Warningf("Not retrying current segment nonce=%d seqNo=%d due to context cancellation err=%v", nonce, seg.SeqNo, err)
```

server/broadcast_test.go
```diff
@@ -771,29 +771,32 @@ func TestProcessSegment_MaxAttempts(t *testing.T) {
...
-	_, err = processSegment(cxn, seg)
+	// Context canceled. Execute once and not retry
+	MaxAttempts = 10
+	ctx, cancel := context.WithCancel(context.Background())
+	cancel()
+	_, err = processSegment(ctx, cxn, seg)
 	assert.NotNil(err)
-	assert.Equal("Hit max transcode attempts: UnknownResponse", err.Error())
+	assert.Contains("context canceled", err.Error())
 	assert.Equal(2, transcodeCalls, "Segment submission calls did not match")
```
victorges: Shouldn't this be 1?

Better yet, should we check if the context is cancelled **before** we make the attempt, so that we don't try even once in case the context is cancelled before we start trying?

victorges (approved):
LGTM

server/broadcast.go
leszko: Good idea, applied.

server/broadcast_test.go
leszko: No, it shouldn't be one. This value is aggregated across the whole test. Meaning:
1. `+1` in line 781
2. `+1` in line 791 (because we try one time to transcode)

Is it good or bad that we have one unit test for basically one function, not a separate function for one test scenario, it's a good question. I'd personally change it to have one unit test function per each scenario, but here I want to be consistent with what we do in the whole project.

> Better yet, should we check if the context is cancelled before we make the attempt, so that we don't try even once in case the context is cancelled before we start trying?

It boils down to what we discussed https://github.com/livepeer/go-livepeer/pull/2124, especially these comments:

https://github.com/livepeer/go-livepeer/issues/2113#issuecomment-981834244
> It would make sense to me to always make a "reasonable attempt" to complete the transcode if a segment was (fully?) received, regardless of client connection status. 

And

https://github.com/livepeer/go-livepeer/issues/2113#issuecomment-984847401
>  it doesn't completely break the ability to use ffmpeg to push segments and then query for a playlist after

The thing is that if we don't try to transcode even once, then If the connection is suddenly closed, we don't have any results.

leszko (commented):
Thanks for the review @victorges

darkdarkdragon (approved):
LGTM

server/broadcast_test.go
victorges: > The thing is that if we don't try to transcode even once, then If the connection is suddenly closed, we don't have any results.

Makes total sense, agreed!

> No, it shouldn't be one. This value is aggregated across the whole test. Meaning:

I see... Yeah, this running of multiple test cases on the same test is quite confusing right now, so IMO it's bad. We could theorically create some setup functions and call them before each case, but there are some nice test libraries out there that we could use which make this much easier. 

One of those that I really like (but also one of the few that I have XP with) is [goconvey](https://github.com/smartystreets/goconvey) which has some nice sugar for setting things up so that you write the setup code on the same function where you're going to use it, but it's also re-executed for each "leaf" test function (parent functions basically work as "before each" steps). An example of where I've used it: https://github.com/victorges/nuledger/blob/main/authorizer/handler_test.go#L68

As for our current case, WDYT of resetting the `transcodeCalls` on the beginning of each test case, so that we can deal with a more reasonable number in each of these cases? Could be done in the same line where we set the `MaxAttempts` like `MaxAttempts, transcodeCalls = 10, 0`

victorges: Either way, total optional/nitpick change now that I understood this is correct. Feel free to just resolve the comment and leave this as is.

leszko: > One of those that I really like (but also one of the few that I have XP with) is goconvey which has some nice sugar for setting things up so that you write the setup code on the same function where you're going to use it, but it's also re-executed for each "leaf" test function (parent functions basically work as "before each" steps). An example of where I've used it: https://github.com/victorges/nuledger/blob/main/authorizer/handler_test.go#L68

Yeah, In my opinion, it'd also be better. Now these tests are messy.

> As for our current case, WDYT of resetting the transcodeCalls on the beginning of each test case, so that we can deal with a more reasonable number in each of these cases? Could be done in the same line where we set the MaxAttempts like MaxAttempts, transcodeCalls = 10, 0

Yeah, good idea. Updated.

Issue #2134, opened 2021-12-11
Support additional container formats for VP8/9


cyberj0g:
As LPMS support for VP8 and VP9 is added by [this PR](https://github.com/livepeer/lpms/pull/279), we need to support containers compatible with these codecs. I propose enabling WEBM and MKV support in production `libavcodec` bulid. It will require adding something like that to ffmpeg build arguments:
```
--enable-demuxer=matroska,webm,opus
```
We support hardware accelerated VP8/9 decoding, but not encoding, therefore, only viable route is `.mkv` or `.webm` input and `.mp4` or `.ts` output. We need to enable Ffmpeg's internal `opus` demuxer (and maybe `vorbis`) to support WEBM, which won't come with AAC audio. MKV support is feasible as it's a very common container format too. I'm not sure if we are interested in supporting WEBM DASH, given streaming is handled by Mist server. The question is,  how `.webm` and `.mkv` would be split into segments for transcoding?

darkdarkdragon:
Shouldn't we also enable `opus` codec? I suspect main use case for VP8 will be to take in WebRTC stream (with VP8/opus) and transcode it into h264/aac.

> The question is, how .webm and .mkv would be split into segments for transcoding?

It is not the job of go-livepeer node to split files (.web, .mkv) into segments.

Thulinma:
Isn't the Opus demuxer only for `.opus` files? As far as I know, the matroska demuxer can handle Opus inside `.mkv` just fine without further requirements (but I could be wrong about this...).
Oh, perhaps good to know: webm is a strict subset of matroska. So the matroska demuxer _can_ handle that by itself, too.

Not supporting DASH seems reasonable (and sensible) to me. 👍
If you're doing MP4/TS outputs, plain HLS of course works just fine. It _would_ however be nice to (optionally?) support MKV format outputs as well, as the overhead is much smaller than with MP4 or TS, and parsing that format uses less RAM/CPU. It would give a decent performance boost 🙂. Of course that is something that could be added at a later date as well (perhaps when we move to a fully-streaming workflow?).

darkdarkdragon:
Didn't know that webm is a subset of matroska. Then it makes sense to enable them both for input and output.

cyberj0g:
Thanks for the input. I like the idea of including MKV (and WEBM) support. WEBM demuxer and muxer still needs to be explicitly enabled. Currently, relevant Ffmpeg options for production build look like this (HW codecs are included separately):
```
[...]

PR #2135, opened 2021-12-13
Unified install_ffmpeg.sh for go-livepeer and lpms with HEVC and VP8/9 support

cyberj0g:
- support for [HEVC, VP8/9 PR](https://github.com/livepeer/lpms/pull/279) on `go-livepeer` side
- use a single `install_ffmpeg.sh` for both `lpms` and `go-livepeer`, which will reside in `go-livepeer` repo
- #2134 

All changes are in `install_ffmpeg.sh`:
- add `libx265` and `libvpx` 
- add `debug-video` mode for development and testing of `lpms`
- add `libtensorflow` installation

#2134 

**Checklist:**


install_ffmpeg.sh
```diff
@@ -75,47 +75,89 @@ if [ ! -e "$ROOT/x264" ]; then
...
 if [ $(uname) == "Darwin" ]; then
   EXTRA_LDFLAGS="-framework CoreFoundation -framework Security"
 else
   # If we have clang, we can compile with CUDA support!
-  if which clang > /dev/null; then
-    echo "clang detected, building with GPU support"
-
-    EXTRA_FFMPEG_FLAGS="--enable-cuda --enable-cuda-llvm --enable-cuvid --enable-nvenc --enable-decoder=h264_cuvid --enable-filter=scale_cuda,signature_cuda --enable-encoder=h264_nvenc"
-
+  if which nvidia-smi > /dev/null; then
```
iameli, line 110: This isn't quite right - the magic of `clang` here is that you can build ffmpeg without any requirements to have CUDA, nvcc, or anything like that installed.

iameli: This isn't quite right — the magic of `clang` here is that you can build with CUDA support without having CUDA, nvcc, or any other nvidia dependencies involved.

install_ffmpeg.sh
```diff
@@ -75,47 +75,89 @@ if [ ! -e "$ROOT/x264" ]; then
...
+    echo "CUDA detected, building with GPU support"
+    sudo apt install -y clang
```
iameli: The only reason this didn't break the Windows build is because of my previous comment. We have a handful of orchestrators using Windows CUDA support, we don't want to break it.

install_ffmpeg.sh
```diff
@@ -75,47 +75,89 @@ if [ ! -e "$ROOT/x264" ]; then
   make install-lib-static
 fi
 
+if [ $(uname) == "Linux" && $BUILD_TAGS == *"debug-video"*]; then
```
iameli: No support for VP8/VP9/H265 on Windows or Mac, then? That's a shame.

iameli (dismissed):
Concerned about multi-platform support - this script should work in as many environments as possible. I can live with no Tensorflow on Windows if we don't think it's feasible, but we should be able to get x265 and libvpx working on Mac and Windows.

install_ffmpeg.sh
cyberj0g: Yes, no software support for these codecs on Windows and Mac, I think it's not first priority getting these dependencies working there. It's not feasible to use SW codecs on production, we are using them mostly for tests in lpms. For dev purposes, users are welcome to use Linux, or follow these codecs install guides for Mac - or even use latest WSL with CUDA support, I think?

install_ffmpeg.sh
cyberj0g: Makes sense, I didn't know Ffmpeg doesn't require nvcc.

cyberj0g:
> Concerned about multi-platform support - this script should work in as many environments as possible. I can live with no Tensorflow on Windows if we don't think it's feasible, but we should be able to get x265 and libvpx working on Mac and Windows.

Why support any software codec on any platform in production, and why spend efforts to support unusual codec-platform combinations for development build, which only required for running some tests in `lpms`?

darkdarkdragon:
> > Concerned about multi-platform support - this script should work in as many environments as possible. I can live with no Tensorflow on Windows if we don't think it's feasible, but we should be able to get x265 and libvpx working on Mac and Windows.
> 
> Why support any software codec on any platform in production, and why spend efforts to support unusual codec-platform combinations for development build, which only required for running some tests in `lpms`?

Not just tests in `lpms` - I'm running go-livepeer node on Mac in software mode for development and would like to be able to continue to do so 😄 
Also, many people use go-livepeer node in production on Windows, so them should have software versions of codecs too.

cyberj0g:
> I'm running go-livepeer node on Mac in software mode for development and would like to be able to continue to do so

In this case I invite you to add support for MacOS, because I don't have access to one and won't be able to test it 😁

> many people use go-livepeer node in production on Windows

Does someone really running a transcoder on Windows with software codecs? If it's remotely possible with H.264, HEVC and Vp8/9 unlikely to work real-time. Nvidia codecs are working just fine without libx265 and libvpx.

iameli:
> Does someone really running a transcoder on Windows with software codecs? If it's remotely possible with H.264, HEVC and Vp8/9 unlikely to work real-time.

`.\ffmpeg.exe -i "2021-12-15 07-17-32.mkv" -c:v libx265 -c:a copy -y "2021-12-15 07-17-32.h265.mkv"` on my mid-range Windows PC, no tuning of that command whatsoever, [seems to produce just under 3x realtime transcoding](https://gist.github.com/iameli/39fc7557e82325cb780d589ee9f1f7b0). Why would non-Linux platforms be slower? It's all just C.

> Why support any software codec on any platform in production

Because we're a weird, decentralized transcoding platform. It's very hard for us to predict ahead of time the unusual environments where our software will be deployed. If odd platform combinations end up causing bugs, that's one thing, and we could have a conversation about the effort required to fix those bugs. But so far, once we've gotten things compiling successfully, the actual transcoding has worked flawlessly on all three platforms. Let's not give up on that just because it makes the build script a bit more complex!

> and why spend efforts to support unusual codec-platform combinations for development build, which only required for running some tests in lpms?

I don't follow — this script isn't just for the development build, it's also what we use to produce binaries for distribution. Surely those tests exist in lpms because we want to eventually support these codecs in go-livepeer?

> In this case I invite you to add support for MacOS, because I don't have access to one and won't be able to test it 😁

That's fair! I'd ask that you start out by writing a version of this that works the same on all platforms and let the CI process attempt to build it. It might just work out of the box. If you run into platform-specific compile problems, or you want @darkdarkdragon or I to test the resulting binaries on Mac/Windows, give us a ping.

cyberj0g:
> seems to produce just under 3x realtime transcoding.

That's faster than I expected, however, I suggest running the test using our test segment:
```
wget https://storage.googleapis.com/lp_testharness_assets/bbb_1080p_30fps_1min_2sec_hls.tar.gz
```
And ffmpeg command closer resembling transcoding logic:
```
ffmpeg -y -i bbb/source.m3u8 -filter_complex '[0:v]split=3[out1][out2][out3]' \
                  -map '[out1]' -s 1280x720 -acodec copy -vcodec libx265 output720p.mp4 \
                  -map '[out2]' -s 640x360  -acodec copy -vcodec libx265 output360p.mp4 \
                  -map '[out3]' -s 320x240  -acodec copy -vcodec libx265 output240p.mp4
```
Also try `libvpx` and `libvpx-vp9` with `.mkv` outputs. According to last GPU benchmarks (with H.264, though), GTX 1080 was able to run about 20 such sessions real-time in parallel with mining or CUDA computations. You can calculate LPT payout per watt for CPU, and I believe it will be much lower than electricity cost. Livepeer was never intended to run on CPU and existence of Windows software transcoding nodes is news to me. If we must support that - I'll fix the script, no problem.

> I don't follow — this script isn't just for the development build

As I see it:
- all production deployments (where users supposed to be incentivized by payouts) are using hardware acceleration and Nvidia codecs (hevc_cuvid, h264_cuvid, vp8_cuvid, vp9_cuvid, hevc_nvenc, h264_nvenc), which don't require software codecs `libx265` or `libvpx` - so they are not built by this version of script for any platform
- developers may want to use software codecs if they don't have Nvidia GPU in their environment
- tests in `lpms` are using software codecs only for two things: 1. to run on CircleCI without GPU 2. to validate encoding quality of hardware codecs compared to reference software decoded segments.

For me, it makes perfect sense to not support `libvpx` and `libx265` at all in this codec-related PR. We can add support of these for developers separately, or even just add install instructions to contributing guide.

iameli:
> all production deployments (where users supposed to be incentivized by payouts) are using hardware acceleration and Nvidia codecs

I'm not confident of that at all - I think people run all sorts of different hardware on the network, though I hear you that these software encoders are much slower than others.

Started hacking on a cross-platform approach here, but it'll probably take some iteration to get right: https://github.com/livepeer/go-livepeer/pull/2136. I'm cool with this landing in the meantime if it unblocks our support for hardware HEVC and whatnot.

cyberj0g:
Added comment on static build of new codecs: https://github.com/livepeer/lpms/pull/279#issuecomment-995700995

yondonfu:
Chiming in on the discussion of supporting software based transcoding for the new codecs:

I see the following prioritization right now for new codec support:

1. HW transcoding on Linux
    - Relevant for production usage, since its the most common [1] setup on the network
2. HW transcoding on Windows
    - Relevant for production usage, since its the second most common [2] setup on the network
3. SW transcoding on Linux
4. SW transcoding on Mac
    - I don't expect this to ever be used in production, but since Mac/Linux are common development environments I rank this above SW transcoding on Windows
5. SW transcoding on Windows

Given the above prioritization, I think it is reasonable to focus on 1/2 and to look into 3/4/5 separately if a) it is challenging to support everything at once and b) we believe there is a reasonable way for someone without access to a GPU to do manual testing with go-livepeer for new codecs (not to test the actual decoding/encoding, but for the code paths around selecting the codec). If we do this, we can just make it clear that initially there is limited support for new codecs.

At the same time, while for *new* codecs we can opt to support HW transcoding first and look into supporting SW transcoding separately, I do think it is important not to take away support for SW transcoding for *existing* codecs (i.e. H264) just because I think we should aim to avoid removing features unless we absolutely need to (i.e. security reasons, we are confident that the pain of supporting it outweighs any benefit certain users might get from it, etc.).

[1] This is anecdotal, but I feel like its right...
[2] This is also anecdotal, but I also feel like its right...

install_ffmpeg.sh
```diff
@@ -75,47 +75,88 @@ if [ ! -e "$ROOT/x264" ]; then
...
 else
   # If we have clang, we can compile with CUDA support!
   if which clang > /dev/null; then
     echo "clang detected, building with GPU support"
-
-    EXTRA_FFMPEG_FLAGS="--enable-cuda --enable-cuda-llvm --enable-cuvid --enable-nvenc --enable-decoder=h264_cuvid --enable-filter=scale_cuda,signature_cuda --enable-encoder=h264_nvenc"
-
+    EXTRA_FFMPEG_FLAGS="--enable-cuda --enable-cuda-llvm --enable-cuvid --enable-nvenc --enable-decoder=h264_cuvid,hevc_cuvid,vp8_cuvid,vp9_cuvid --enable-filter=scale_cuda,signature_cuda,hwupload_cuda --enable-encoder=h264_nvenc,hevc_nvenc"
     if [[ $BUILD_TAGS == *"experimental"* ]]; then
+        if [ ! -e "$ROOT/compiled/lib/libtensorflow_framework.so" ]; then
```
yondonfu: If we're adding this here, mind removing the installation step [here](https://github.com/livepeer/go-livepeer/blob/a8a3ede471a67162bc2246c388ad93c82a109be9/docker/Dockerfile.build-linux#L25)?

install_ffmpeg.sh
```diff
@@ -75,47 +75,88 @@ if [ ! -e "$ROOT/x264" ]; then
...
     --prefix="$ROOT/compiled" \
-    $EXTRA_FFMPEG_FLAGS
-  make
+    $EXTRA_FFMPEG_FLAGS \
+    $DEV_FFMPEG_FLAGS
+fi
+
+if [[ ! -e "$ROOT/ffmpeg/libavcodec/libavcodec.a" || $BUILD_TAGS == *"debug-video"* ]]; then
+  cd "$ROOT/ffmpeg"
+  make $FFMPEG_MAKE_EXTRA_ARGS
```
yondonfu, line 159: Shouldn't we run make unconditionally after running the ffmpeg configure script?

yondonfu (commented):
Hm looks like the Linux Tensorflow Build is failing. Any idea why?

install_ffmpeg.sh
cyberj0g: Make will run with same condition as configure (static lib is missing), plus it will run always in debug mode for convenience, in case source files were edited.

install_ffmpeg.sh
cyberj0g: Done, also switched to libtensorflow version with both CPU and GPU support.

cyberj0g:
> Hm looks like the Linux Tensorflow Build is failing

Fixed, I initially thought it's CircleCI caching issue, but it was permission-related.

install_ffmpeg.sh
yondonfu: Looks like we're still using libtensorflow 2.3.0. Does that have CPU & GPU support?

install_ffmpeg.sh
```diff
@@ -75,47 +75,88 @@ if [ ! -e "$ROOT/x264" ]; then
...
     echo "clang detected, building with GPU support"
-
-    EXTRA_FFMPEG_FLAGS="--enable-cuda --enable-cuda-llvm --enable-cuvid --enable-nvenc --enable-decoder=h264_cuvid --enable-filter=scale_cuda,signature_cuda --enable-encoder=h264_nvenc"
-
+    EXTRA_FFMPEG_FLAGS="--enable-cuda --enable-cuda-llvm --enable-cuvid --enable-nvenc --enable-decoder=h264_cuvid,hevc_cuvid,vp8_cuvid,vp9_cuvid --enable-filter=scale_cuda,signature_cuda,hwupload_cuda --enable-encoder=h264_nvenc,hevc_nvenc"
     if [[ $BUILD_TAGS == *"experimental"* ]]; then
+        if [ ! -e "$ROOT/compiled/lib/libtensorflow_framework.so" ]; then
+          LIBTENSORFLOW_VERSION=2.3.0 \
+          && curl -LO https://storage.googleapis.com/tensorflow/libtensorflow/libtensorflow-gpu-linux-x86_64-${LIBTENSORFLOW_VERSION}.tar.gz \
+          && tar -C $ROOT/compiled -xzf libtensorflow-gpu-linux-x86_64-${LIBTENSORFLOW_VERSION}.tar.gz \
```
yondonfu: Why do we have two tar commands here with `-C $ROOT/compiled` and `-C /usr/local`? Note that [Dockerfile.release-linux](https://github.com/livepeer/go-livepeer/blob/3649803de21ef36e1fe814c51ca6b94bef4daed6/docker/Dockerfile.release-linux#L17), which creates the final image with the go-livepeer binary, assumes that libtensorflow is installed in `/usr/local`.

install_ffmpeg.sh
yondonfu: Ah! Right because that static lib wouldn't exist after the configure step and would only exist after make is run. Got it!

install_ffmpeg.sh
cyberj0g: Yes, GPU version supports CPU.

install_ffmpeg.sh
cyberj0g: The short answer is because CircleCI build fails (Ffmpeg's configure cannot find libtensorflow) without copying the lib to system location, as it was implemented previously. I couldn't reproduce this issue by running docker build locally and don't think there's much harm in installing libtensorflow both system-wide and into $ROOT/compiled/lib.

yondonfu: If there is no reason to install libtensorflow in $ROOT/compiled/lib and we're already installing libtensorflow system wide in /usr/local which is what ffmpeg's configure requires can we change the libtensorflow_framework.so conditional check to be based on /usr/local instead of $ROOT/compiled/lib and remove the installation in $ROOT/compiled/lib?

I hear you that installing it in two different places doesn't do much harm, but I think it'd be preferable for a user (i.e. dev) running the script to just have a single install location as opposed to have libtensorflow dropped in multiple locations. Additionally, if ffmpeg's configure requires the /usr/local installation anyway I'm a bit wary of the scenario where somehow libtensorflow exists in $ROOT/compiled/lib, but not in /usr/local so this step never runs.

cyberj0g: Ok, switched to `/usr/local` only installation, we can postpone figuring out why `configure` can't pick it up from $ROOT on CircleCI until we need a cross-platform build with Tensorflow.

yondonfu (approved):
LGTM! Let's squash all the commits prior to merging.

PR #2137, opened 2021-12-16
cmd: Add config file support

### LINKED ISSUE #101: Support Config File

leszko:
Add a possibility to configure Livepeer using a config file instead of flags.

Sample usage:
```
$ livepeer -config livepeer.conf
```

Sample `livepeer.conf`

```
broadcaster
cliAddr :7937
httpAddr :8938
orchAddr 127.0.0.1:8935,127.0.0.1:8936
v 6
```

Used the library https://github.com/peterbourgon/ff, which allows using config file for the flags.

1. Created a sample config file and checked if it works
2. Checked if both config files and flags are defined if flags take precedence over config file

fix #101

**Checklist:**

### DIFF
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -29,6 +29,7 @@ import (
 	"github.com/livepeer/go-livepeer/pm"
 	"github.com/livepeer/go-livepeer/server"
 	"github.com/livepeer/livepeer-data/pkg/event"
+	"github.com/peterbourgon/ff/v3"
 
 	ethcommon "github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/ethclient"
@@ -159,7 +160,13 @@ func main() {
 	orchWebhookURL := flag.String("orchWebhookUrl", "", "Orchestrator discovery callback URL")
 	detectionWebhookURL := flag.String("detectionWebhookUrl", "", "(Experimental) Detection results callback URL")
 
-	flag.Parse()
+	// Config file (used instead of flags)
+	_ = flag.String("config", "", "Config file in the format 'key value', flags take precedence over the config file")
+	ff.Parse(flag.CommandLine, os.Args[1:],
+		ff.WithConfigFileFlag("config"),
+		ff.WithConfigFileParser(ff.PlainParser),
+	)
+
 	vFlag.Value.Set(*verbosity)
 
 	isFlagSet := make(map[string]bool)
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -164,6 +164,7 @@ func main() {
 	_ = flag.String("config", "", "Config file in the format 'key value', flags take precedence over the config file")
 	ff.Parse(flag.CommandLine, os.Args[1:],
 		ff.WithConfigFileFlag("config"),
+		ff.WithEnvVarPrefix("LP"),
 		ff.WithConfigFileParser(ff.PlainParser),
 	)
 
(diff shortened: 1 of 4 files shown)


cmd/livepeer/livepeer.go
```diff
@@ -159,7 +160,13 @@ func main() {
 	orchWebhookURL := flag.String("orchWebhookUrl", "", "Orchestrator discovery callback URL")
 	detectionWebhookURL := flag.String("detectionWebhookUrl", "", "(Experimental) Detection results callback URL")
 
-	flag.Parse()
+	// Config file (used instead of flags)
```
yondonfu: Is this comment accurate? IIUC the flags are still used if they are specified and they take precedence over the config file if that is specified as well.

hthillman:
> Checked if both config files and flags are defined if flags take precedence over config file

Just to be super clear here - if flags are defined, they'll override the config?

yondonfu (commented):
Looks good! Just one minor question related to a comment.

I also think it would be great to add a short guide for configuring `livepeer` using a config file in https://github.com/livepeer/livepeer-org/tree/master (the repo for https://livepeer.org/docs). Maybe a section called "Configuring Livepeer" under the "Installation" header...

Also, just want to confirm my understanding of the `ff` package:

Since the package will parse flags, env vars *and* the config file the following should be true right?

- If `-maxGasPrice` is specified in all three, the value used in the flag takes precedence
- If `-maxGasPrice` is not specified in flags, but other flags are specified and the config file specifies `-maxGasPrice` then the value from the config file will be used

EDIT: Regarding the above question - just read your testing steps! 😓

leszko:
> > Checked if both config files and flags are defined if flags take precedence over config file
> 
> Just to be super clear here - if flags are defined, they'll override the config?

Yes, they'll override the config.

cmd/livepeer/livepeer.go
```diff
@@ -159,7 +160,13 @@ func main() {
...
+	_ = flag.String("config", "", "Config file in the format 'key value', flags take precedence over the config file")
+	ff.Parse(flag.CommandLine, os.Args[1:],
+		ff.WithConfigFileFlag("config"),
```
iameli, line 166: I think we'll need something like this to enable environment variables?
```suggestion
		ff.WithConfigFileFlag("config"),
		ff.WithEnvVarPrefix("LP"),
```

leszko: I didn't know we also want to use env variables. But if yes, then right.

cmd/livepeer/livepeer.go
leszko: Right, updated.

leszko:
> I also think it would be great to add a short guide for configuring `livepeer` using a config file in https://github.com/livepeer/livepeer-org/tree/master (the repo for https://livepeer.org/docs). Maybe a section called "Configuring Livepeer" under the "Installation" header...

Yes, working on this together with @hthillman in the branch `docs-config-updates` of the https://github.com/livepeer/livepeer-org repo. I think we can merge this PR and later merge the doc-related PR.

> Also, just want to confirm my understanding of the `ff` package:
> 
> Since the package will parse flags, env vars _and_ the config file the following should be true right?
> 
> * If `-maxGasPrice` is specified in all three, the value used in the flag takes precedence
> * If `-maxGasPrice` is not specified in flags, but other flags are specified and the config file specifies `-maxGasPrice` then the value from the config file will be used
> 
> EDIT: Regarding the above question - just read your testing steps! 😓

Yeah, it works exactly as you described. One more thing is that after the comment from @iameli , we added the env variables. So the priority here is (from the highest):
- flags
- env variables
- config file

yondonfu (approved):
LGTM!

The ability to use env vars for config is a bonus - nice!

iameli (approved):
LGTM!

PR #2146, opened 2021-12-21
core: Allows Os receive payments while pending activation

### LINKED ISSUE #2140: Allow Os to receive tickets when they are pending activation
Allow Os to receive tickets if they are pending activation as opposed the current requirement where they need to be active to receive tickets. The reason for this change is to allow Os that are scheduled to be active in the next round after they migrate to L2 to immediately be able to accept work on L2 in order to minimize the period where an O cannot receive work at all. Since the O is pending activation and since tickets are valid for at least 2 rounds (the current ticket validity period), the O will be able to redeem any winning tickets as soon as the next round.

- O calls isActive() [1] in ProcessPayment() [2] which is called whenever O receives a segment [3]
- isActive() queries the DB with a filter in order to see if the O's address is active [4]
- At the moment, `CurrentRound` is set to the last initialized round which means the query should only return addresses that are active as of the last initialized round
- The only address passed to `Addresses` is the O's own address because it just cares if its own address is active
- The actual DB filtering logic is implemented via a dynamically constructed query [5]

One way to implement this feature could be set the `CurrentRound` in the `DBOrchFilter` struct created in `isActive()` to `orch.rm.LastInitializedRound + 1` which would mean an address that is active in the next round would be included in the result. 

If we do this we may want to consider changing the name of the function from `isActive()` to something `isPaymentEligible()` to indicate that an O can be eligible to receive payments even before it is active on-chain - in this case, it is eligible to receive payments since it is pending activation in the next round.

[1] https://github.com/livepeer/go-livepeer/blob/832166f4dbfe7c631a0f6d6ad3dc4916a148d58a/core/orchestrator.go#L127
[2] https://github.com/livepeer/go-livepeer/blob/832166f4dbfe7c631a0f6d6ad3dc4916a148d58a/core/orchestrator.go#L111
[3] https://github.com/livepeer/go-livepeer/blob/832166f4dbfe7c631a0f6d6ad3dc4916a148d58a/server/segment_rpc.go#L101
[4] https://github.com/livepeer/go-livepeer/blob/832166f4dbfe7c631a0f6d6ad3dc4916a148d58a/core/orchestrator.go#L353
[5] https://github.com/livepeer/go-livepeer/blob/832166f4dbfe7c631a0f6d6ad3dc4916a148d58a/common/db.go#L524

leszko:
Allow receiving payments by orchestrator which is not yet activated (but will be activated in the next round).

N/A

1. Set up local geth
2. Set up Broadcaster and Orchestrator
3. Manually updated SQLite DB for Orchester to modify acivationRound
4. Check that the segments are still transcoded when the current round is `activationRound - 1`, but are not transcoded when the current round is `activationRound - 2`.

fix #2140

**Checklist:**

### DIFF
--- a/core/orch_test.go
+++ b/core/orch_test.go
@@ -1195,19 +1195,32 @@ func TestIsPaymentEligible(t *testing.T) {
 	addr := defaultRecipient
 	dbh, dbraw := tempDBWithOrch(t, &common.DBOrch{
 		EthereumAddr:      addr.Hex(),
-		ActivationRound:   1,
+		ActivationRound:   2,
 		DeactivationRound: 999,
 	})
 	defer dbh.Close()
 	defer dbraw.Close()
 
+	// not active yet
 	n, _ := NewLivepeerNode(nil, "", dbh)
 	rm := &stubRoundsManager{
-		round: big.NewInt(10),
+		round: big.NewInt(0),
 	}
 	orch := NewOrchestrator(n, rm)
 
 	ok, err := orch.isPaymentEligible(addr)
+	assert.False(ok)
+	assert.NoError(err)
+
+	// pending activation
+	rm.round = big.NewInt(1)
+	ok, err = orch.isPaymentEligible(addr)
+	assert.True(ok)
+	assert.NoError(err)
+
+	// active
+	rm.round = big.NewInt(10)
+	ok, err = orch.isPaymentEligible(addr)
 	assert.True(ok)
 	assert.NoError(err)
 
--- a/core/orch_test.go
+++ b/core/orch_test.go
@@ -799,7 +799,7 @@ func TestProcessPayment_ActiveOrchestrator(t *testing.T) {
 
 	// orchestrator inactive -> error
 	err := orch.ProcessPayment(defaultPayment(t), ManifestID("some manifest"))
-	expErr := fmt.Sprintf("orchestrator %v is inactive in round %v, cannot process payments", addr.Hex(), 10)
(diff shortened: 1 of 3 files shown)


core/orch_test.go
```diff
@@ -1207,13 +1207,13 @@ func TestIsActive(t *testing.T) {
 	}
 	orch := NewOrchestrator(n, rm)
 
-	ok, err := orch.isActive(addr)
+	ok, err := orch.isPaymentEligible(addr)
```
yondonfu, line 1211: WDYT about adding an additional test for when `addr` is pending activation?

yondonfu, line 1211: WDYT about re-naming `TestIsActive()` to `TestIsPaymentEligible`?

yondonfu (commented):
Looks good! Just two minor comments.

core/orch_test.go
leszko: Good idea, added in https://github.com/livepeer/go-livepeer/pull/2146/commits/9ae8d38313c948746d08d25d9e04113010a77d48

leszko: Updated in https://github.com/livepeer/go-livepeer/pull/2146/commits/07a2560b3df9680245cbe83ecfc905f727449d19

leszko (commented):
Thanks for the review @yondonfu I addressed your comments. PTAL.

yondonfu (approved):
LGTM!

darkdarkdragon (approved):
LGTM

Issue #2123, opened 2021-12-02
Support H265/HEVC, VP8, VP9


cyberj0g:
## Abstract
Currently, we support only H.264 codec in the transcoding pipeline, while H.265 (HEVC) and VP8/VP9 are very popular and widely supported alternatives. 

## Proposed Solution

As Livepeer focuses on using Nvidia GPUs capabilities, it makes sense to first add support for codecs which has hardware implementations on these devices. As per [support matrix](https://developer.nvidia.com/video-encode-and-decode-gpu-support-matrix-new):
* VP8/9: **encoding** is not currently supported
* VP8/9: **decoding** is supported on newer devices
* H.265: both encoding and decoding is supported on most devices starting Maxwell (2nd gen), except some entry-level mobile GPUs

This asks for a way in the protocol to announce which encoders, decoders and capabilities in general T supports hardware-wise or wiling to provide. I'm not very knowledgeable on the protocol, so feel free to add your ideas on how to best implement that, or maybe it's already implemented? Adding H.265 support to transcoder is fairly trivial, but require data structures refactoring and corner case testing.

## Implementation Tasks and Considerations

1. Extend `VideoProfile` structure [here](https://github.com/livepeer/lpms/blob/fe330766146dba62f3e1fccd07a4b96fa1abcf4d/ffmpeg/videoprofile.go#L48) to contain both input and output codec. Treat it as input parameters, and `TranscodeOptions` as runtime parameters with specific codec names, which are created from `VideoProfile` [here](https://github.com/livepeer/go-livepeer/blob/f3a299bb95af04004fd957cdc3b2b4a69c72c90b/core/transcoder.go#L233). The main reason for such separation is a requirement to use different encoder\decoder names for same codec with and without HW acceleration.
2. Add corresponding constants for default types of renditions [here](https://github.com/livepeer/lpms/blob/fe330766146dba62f3e1fccd07a4b96fa1abcf4d/ffmpeg/videoprofile.go#L62), when transcoding H265>H264, H264>H265, H265>H265. 
2. Support configuring specific codec with additional parameters for HW processing [here](https://github.com/livepeer/lpms/blob/fe330766146dba62f3e1fccd07a4b96fa1abcf4d/ffmpeg/ffmpeg.go#L331).
[...]

PR #2150, opened 2021-12-23
Transcoder capabilities discovery and validation

cyberj0g:
This PR is the last part of https://github.com/livepeer/go-livepeer/issues/2123 implementation. It will allow Orchestrator to send tasks to correct Transcoder according to requested capabilities (codecs). Also it adds capability test at startup for Nvidia GPUs (which may support only a subset of capabilities).

This PR includes changes of `install_ffmpeg.sh` discussed [here](https://github.com/livepeer/go-livepeer/pull/2135). Corresponding [LPMS PR](https://github.com/livepeer/lpms/pull/279) needs to be merged first.

* add capability test function which is run on node startup
* add logic to select transcoder according to capabilities in O-T mode
* add logic to validate capabilities in OT mode 

Ran tests

https://github.com/livepeer/go-livepeer/issues/2123

**Checklist:**

**Additional notes:**
One missing piece is automatic capability (codec) detection from segment. Currently, if segment is encoded with codec, for which T doesn't support decoding - and decoding capability is not explicitly requested, it will bypass capability-based T selection logic and may fail, even if O has Ts which could decode it. Should `go-livepeer` has such segment codec check (at which point?), or it's best to handle on MistServer side?


oscar-davids (approved):
LGTM!

cyberj0g:
Non-obvious quirk on multi-GPU system I just realized: there's no GPU-level segment scheduling logic, and capability discovery will provide a list of capabilities from less capable GPU, so if user has a rig with 5 x RTX 3090 and 1 GTX 745 to hook up the monitor, no codecs except H.264 would show up in capabilities, unless `-nvidia` argument excludes latter GPU. It worth documenting.

core/transcoder.go
```diff
@@ -114,41 +115,66 @@ func (nv *NvidiaTranscoder) Transcode(ctx context.Context, md *SegTranscodingMet
...
+				}
+			}
+		}
+		caps = append(caps, c)
+		test_fail:
 	}
-	defer os.Remove(fname)
-	for _, device := range devices {
-		t1 := NewNvidiaTranscoder(device)
-		// "145x1" is the minimal resolution that succeeds on Windows, so use "145x145"
```
darkdarkdragon, line 137: Can we put this comment back?

core/transcoder.go
```diff
@@ -114,41 +115,66 @@ func (nv *NvidiaTranscoder) Transcode(ctx context.Context, md *SegTranscodingMet
...
+			err = ioutil.WriteFile(fname, mp4testSeg, 0644)
+			if err != nil {
+				glog.Errorf("Error writing test segment for capability %d: %s", c, err)
+				continue
+			}
+			defer os.Remove(fname)
+			// check that capability is supported on all devices
+			for _, device := range devices {
+				t1 := NewNvidiaTranscoder(device)
+				// transcode in 3 sessions which is a maximum on consumer Nvidia GPU with official driver
```
darkdarkdragon: Our test intentionally used four output renditions - to let use know about problem at the startup time, not when user started to get real work on production node.

server/ot_rpc.go
```diff
@@ -155,7 +156,10 @@ func runTranscode(n *core.LivepeerNode, orchAddr string, httpc *http.Client, not
 	}
 	ctx = clog.AddSeqNo(ctx, uint64(md.Seq))
 	ctx = clog.AddVal(ctx, "taskId", strconv.FormatInt(notify.TaskId, 10))
-
+	if n.Capabilities!=nil && !md.Caps.CompatibleWith(n.Capabilities.ToNetCapabilities()) {
```
darkdarkdragon: Can we `gofmt` this?

core/orchestrator.go
```diff
@@ -902,19 +905,32 @@ func removeFromRemoteTranscoders(rt *RemoteTranscoder, remoteTranscoders []*Remo
...
 	defer rtm.RTmutex.Unlock()
 
 	checkTranscoders := func(rtm *RemoteTranscoderManager) bool {
 		return len(rtm.remoteTranscoders) > 0
 	}
 
+	findCompatibleTranscoder := func(rtm *RemoteTranscoderManager) int {
+		for i := len(rtm.remoteTranscoders) - 1; i >= 0; i-- {
+			// no capabilities = default capabilities, all transcoders must support them
+			if caps == nil || caps.bitstring.CompatibleWith(rtm.remoteTranscoders[i].capabilities.bitstring) {
```
darkdarkdragon, line 919: What will happen if old transcoder which does not present capabilities will be connected to that orchestrator?

cmd/livepeer/livepeer.go
```diff
@@ -901,12 +904,11 @@ func main() {
...
 		*httpAddr = defaultAddr(*httpAddr, "", n.GetServiceURI().Port())
 
-		caps := core.DefaultCapabilities()
 		if *sceneClassificationModelPath != "" {
 			// Only enable experimental capabilities if scene classification model is actually loaded
-			caps = append(caps, core.ExperimentalCapabilities()...)
+			transcoderCaps = append(transcoderCaps, core.ExperimentalCapabilities()...)
 		}
-		n.Capabilities = core.NewCapabilities(caps, core.MandatoryCapabilities())
+		n.Capabilities = core.NewCapabilities(transcoderCaps, core.MandatoryOCapabilities())
```
darkdarkdragon, line 917: That means that standalone O (which does not do transcoding) will advertise only default capabilities?

darkdarkdragon (changes requested):

cmd/livepeer/livepeer.go
cyberj0g: Yes. Updating of O capabilities when new T connects is not addressed in this PR, as well as B support for such case. Added [comment](https://github.com/livepeer/go-livepeer/issues/2123#issuecomment-1027782218) to issue.

core/transcoder.go
cyberj0g: Ok, got it, changed it back to 4 sessions. IIRC the error message without patch won't be verbose, something like `Out of memory`. Is there section in docs which covers possible solutions to this on consumer GPUs?

core/transcoder.go
cyberj0g: Moved it to `capabilities.go` where test segment structure is.

server/ot_rpc.go
cyberj0g: Done.

cyberj0g (commented):
Thanks for review, all comments addressed.

core/orchestrator.go
cyberj0g: Good catch, addressed it in `ot_rpc.go`.

core/transcoder.go
darkdarkdragon: I know, error message is confusing.
> Is there section in docs which covers possible solutions to this on consumer GPUs?

I'm not sure we want to cover this in our docs - patching drivers is in grey zone from legal viewpoint, so we don't want to encourage it.

server/ot_rpc.go
```diff
@@ -155,7 +156,10 @@ func runTranscode(n *core.LivepeerNode, orchAddr string, httpc *http.Client, not
 	}
 	ctx = clog.AddSeqNo(ctx, uint64(md.Seq))
 	ctx = clog.AddVal(ctx, "taskId", strconv.FormatInt(notify.TaskId, 10))
-
+	if n.Capabilities != nil && !md.Caps.CompatibleWith(n.Capabilities.ToNetCapabilities()) {
+		clog.Errorf(ctx, "Requested capabilities for segment are not compatible with this node taskId=%d url=%s err=%q", notify.TaskId, notify.Url, err)
+		return
```
darkdarkdragon, line 163: you need to call `sendTranscodeResult` before return.
Also `err` is nil here, so I think we should define new error to return here.

darkdarkdragon (changes requested):

server/ot_rpc.go
cyberj0g: Fixed.

darkdarkdragon (approved):
LGTM after squash/rebase

PR #2155, opened 2021-12-27
Update docs and install_ffmpeg.sh for dev on Apple Silicon (M1)

yondonfu:

This PR contains updates to the docs and the `install_ffmpeg.sh` script to support dev on Apple Silicon (M1).

See commit history.

Compiled `livepeer_*` binaries locally and ran the test suite.

N/A

**Checklist:**

### DIFF
--- a/install_ffmpeg.sh
+++ b/install_ffmpeg.sh
@@ -63,11 +63,11 @@ fi
 if [ ! -e "$ROOT/x264" ]; then
   git clone http://git.videolan.org/git/x264.git "$ROOT/x264"
   cd "$ROOT/x264"
-  if [[ ! $IS_M1 ]]; then
+  if [[ $IS_M1 ]]; then
     # newer git master, compiles on Apple Silicon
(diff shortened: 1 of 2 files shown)


install_ffmpeg.sh
```diff
@@ -63,11 +63,11 @@ fi
 if [ ! -e "$ROOT/x264" ]; then
   git clone http://git.videolan.org/git/x264.git "$ROOT/x264"
   cd "$ROOT/x264"
-  if [[ ! $IS_M1 ]]; then
+  if [[ $IS_M1 ]]; then
```
yondonfu, line 66: IIUC prior to this change we would checkout the newer version of x264 if the user is *not* running on M1 . So, I changed this conditional so we checkout the newer version of x264 if the user *is* running on M1.

Without this change, when I ran this script it would fail with `No C compiler found` when trying to install x264.

darkdarkdragon (approved):
LGTM

PR #2157, opened 2021-12-27
Add support for EIP-712 typed data signing in livepeer_cli

### LINKED ISSUE #2141: Support EIP-712 signing
The current Confluence L1Migrator contract supports authorization via a EIP-712 signature.

At the moment, the `livepeer_cli` does not support EIP-712 signing. However, there is logic in clef

There should be a "Sign typed message" option in `livepeer_cli`. When this option is selected:

- The user can paste in the typed data to sign in JSON form. The typed data should follow the format described [here](https://github.com/ethereum/go-ethereum/tree/master/cmd/clef#account_signtypeddata) for the data arg
- `livepeer_cli` will send the data to a `/signTypedMessage` endpoint on the `livepeer` server similar to the `/signMessage` endpoint
- The server will convert the data to an instance of [apitypes.TypedData](https://github.com/ethereum/go-ethereum/blob/619a3e70858e60240ce1df75bdf65ba748387e57/signer/core/apitypes/types.go#L201)
    - See these [tests](https://github.com/ethereum/go-ethereum/blob/619a3e70858e60240ce1df75bdf65ba748387e57/signer/core/signed_data_test.go) for reference
- There can be a `SignTypedMessage` method defined in the eth.AccountManager similar to the [Sign](https://github.com/livepeer/go-livepeer/blob/885079883fb86adffc6fabc8f8d1b1395a2c2856/eth/accountmanager.go#L148) method except it is responsible for producing EIP-712 signatures
- The implementation of this method can be based off of this [method](https://github.com/ethereum/go-ethereum/blob/619a3e70858e60240ce1df75bdf65ba748387e57/signer/core/signed_data.go#L236)
    - At first glance, does not look like we can import this method directly so next easiest thing may be to pull that logic into go-livepeer

**Alternative**

We could integrate with clef, geth's external signer. I like the idea of separating the signer codebase from the rest of the go-livepeer codebase. But, this is a bigger change to architecture that I think is better done separately. For this issue, I think we should find the shortest path to supporting EIP-712 signing with minimal change to UX.

yondonfu:

This PR adds support for EIP-712 typed data signing in `livepeer_cli`.

Prior to this PR, the "Sign message" option in `livepeer_cli` sent a request to the `/signMessage` endpoint on the node's server in order to sign an arbitrary user provided message. The L1Migrator contract for the Confluence LIP allows signature based authorization, but uses the more recent [EIP-712 signing standard](https://eips.ethereum.org/EIPS/eip-712) which describes a standard way to sign the hash of structured data. 

In order to support EIP-712 signing in `livepeer_cli` I've made the following updates:

- Upgraded the go-ethereum dep in order to use the latest package APIs for EIP-712 signing
- Added a `SignTypedData()` method to complement the existing `Sign()` method in the `eth.AccountManager` implementation
- Added support for requesting EIP-712 signing in the `/signMessage` handler by setting a `SigFormat` header
- Added a "Sign typed data" option in `livepeer_cli` to complement the existing "Sign message" option

Updated unit tests and manually tested EIP-712 signing with `livepeer_cli`.

Fixes #2141 

**Checklist:**


server/handlers.go
```diff
@@ -300,11 +301,42 @@ func signMessageHandler(client eth.LivepeerEthClient) http.Handler {
...
+			var d apitypes.TypedData
+			err = json.Unmarshal([]byte(message), &d)
+			if err != nil {
+				respondWith400(w, fmt.Sprintf("could not unmarshal typed data - err=%q", err))
+				return
+			}
+
+			signed, err = client.SignTypedData(d)
+			if err != nil {
+				respondWith500(w, fmt.Sprintf("could not signed typed data - err=%q", err))
```
darkdarkdragon:
```suggestion
				respondWith500(w, fmt.Sprintf("could not sign typed data - err=%q", err))
```

darkdarkdragon (approved):
small typo, otherwise LGTM

server/handlers.go
yondonfu: Ah good catch! Fixed in 1c82132

PR #2158, opened 2021-12-28
[wip] cmd: allow `livepeer` to be embedded elsewhere

iameli:
Just food-for-thought for now. Potentially looking at doing something like https://github.com/livepeer/livepeer-data/pull/31 in all of our projects to create one mega-binary for Livepeer in a Box 2.0.

If we move forward with this we'd probably do the same thing in all the other binaries.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,7 +26,7 @@ endif
 
 .PHONY: livepeer
 livepeer:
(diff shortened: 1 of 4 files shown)


iameli:
nah

Issue #2145, opened 2021-12-20
Fix flaky tests (compilation)


yondonfu:
This is a thread compiling all known flaky tests that should be fixed.

yondonfu:
Data race in HandlePush()

```
==================
WARNING: DATA RACE
Read at 0x000003084930 by goroutine 132:
  github.com/livepeer/go-livepeer/server.(*LivepeerServer).HandlePush.func4()
      /home/runner/work/go-livepeer/go-livepeer/server/mediaserver.go:954 +0x7c

Previous write at 0x000003084930 by goroutine 84:
  github.com/livepeer/go-livepeer/server.setupServerWithCancel()
      /home/runner/work/go-livepeer/go-livepeer/server/mediaserver_test.go:83 +0x175
  github.com/livepeer/go-livepeer/server.TestPush_ReuseIntmidWithDiffExtmid()
      /home/runner/work/go-livepeer/go-livepeer/server/push_test.go:1469 +0x2fc
  testing.tRunner()
      /opt/hostedtoolcache/go/1.15.5/x64/src/testing/testing.go:1123 +0x202

Goroutine 132 (running) created at:
  github.com/livepeer/go-livepeer/server.(*LivepeerServer).HandlePush()
      /home/runner/work/go-livepeer/go-livepeer/server/mediaserver.go:952 +0x1b68
  github.com/livepeer/go-livepeer/server.(*LivepeerServer).HandlePush-fm()
      /home/runner/work/go-livepeer/go-livepeer/server/mediaserver.go:774 +0x64
  net/http.HandlerFunc.ServeHTTP()
      /opt/hostedtoolcache/go/1.15.5/x64/src/net/http/server.go:2042 +0x51
  github.com/livepeer/go-livepeer/server.TestPush_ConcurrentSegments.func2()
      /home/runner/work/go-livepeer/go-livepeer/server/push_test.go:1432 +0x38a

Goroutine 84 (running) created at:
  testing.(*T).Run()
      /opt/hostedtoolcache/go/1.15.5/x64/src/testing/testing.go:1168 +0x5bb
  testing.runTests.func1()
      /opt/hostedtoolcache/go/1.15.5/x64/src/testing/testing.go:1439 +0xa6
  testing.tRunner()
      /opt/hostedtoolcache/go/1.15.5/x64/src/testing/testing.go:1123 +0x202
  testing.runTests()
[...]

PR #2161, opened 2021-12-30
Ip/flaky tests

cyberj0g:
Fixes flaky tests

- TestSubmitSegment_Timeout: Increase segment length and timeouts to prevent seg_length/2 timeout under high CPU load  
- TestMinLSSelector_SelectUnknownSession_RandFreq: run segment selection 10 times and require B to select O with not top stake just once

Multiple runs of `go test ./...`

#2145

**Checklist:**

### DIFF
--- a/server/selection_test.go
+++ b/server/selection_test.go
@@ -4,6 +4,7 @@ import (
 	"container/heap"
 	"context"
 	"errors"
+	"github.com/google/go-cmp/cmp"
 	"math"
 	"math/big"
 	"sort"
@@ -496,7 +497,13 @@ func TestMinLSSelector_SelectUnknownSession_RandFreq(t *testing.T) {
 	sel.Add(sessions)
 
 	// When randFreq = 1.0 we should select randomly instead of selecting the session with the most stake
-	sess := sel.selectUnknownSession(context.TODO())
+	var sess *BroadcastSession
+	for i:=0; i<10; i++ {
+		sess = sel.selectUnknownSession(context.TODO())
+		if !cmp.Equal(sess.OrchestratorInfo.TicketParams.Recipient, addr.Bytes()) {
+			break
+		}
+	}
 	assert.NotEqual(sess.OrchestratorInfo.TicketParams.Recipient, addr.Bytes())
 
 	// When randFreq = 0.0 we should select the session with the most stake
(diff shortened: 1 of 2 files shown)


server/selection_test.go
```diff
@@ -496,7 +497,13 @@ func TestMinLSSelector_SelectUnknownSession_RandFreq(t *testing.T) {
 	sel.Add(sessions)
 
 	// When randFreq = 1.0 we should select randomly instead of selecting the session with the most stake
-	sess := sel.selectUnknownSession(context.TODO())
+	var sess *BroadcastSession
+	for i:=0; i<10; i++ {
+		sess = sel.selectUnknownSession(context.TODO())
+		if !cmp.Equal(sess.OrchestratorInfo.TicketParams.Recipient, addr.Bytes()) {
```
yondonfu: We should be able to use [bytes.Equal](https://pkg.go.dev/bytes#Equal) from the Go stdlib instead of using another package here (note: if we used `github.com/google/go-cmp/cmp` then we would also need to update the `go.mod` and `go.sum` files to track the package version used).

server/selection_test.go
```diff
@@ -496,7 +497,13 @@ func TestMinLSSelector_SelectUnknownSession_RandFreq(t *testing.T) {
 	sel.Add(sessions)
 
 	// When randFreq = 1.0 we should select randomly instead of selecting the session with the most stake
-	sess := sel.selectUnknownSession(context.TODO())
+	var sess *BroadcastSession
+	for i := 0; i < 10; i++ {
+		sess = sel.selectUnknownSession(context.TODO())
+		if !bytes.Equal(sess.OrchestratorInfo.TicketParams.Recipient, addr.Bytes()) {
```
yondonfu, line 503: f16de03ba49817a7f49b9313a7f52ef8bb1e0a4c swaps `bytes.Equal` in for the `cmp.Equal` call using `go-cmp`.

yondonfu (approved):
LGTM!

In the interest of fixing some of these tests to make the release process a bit easier (consistently getting blocked by failing tests right now) I addressed my own comment and I'm going to merge this.

server/selection_test.go
yondonfu: f16de03ba49817a7f49b9313a7f52ef8bb1e0a4c swaps `bytes.Equal` in for the `cmp.Equal` call using `go-cmp`.

PR #2163, opened 2022-01-05
Revert eth: Ensure maxFeePerGas in all transactions never exceed -masGasPrice

leszko:
The following PR: https://github.com/livepeer/go-livepeer/pull/2111 caused the following bug: https://github.com/livepeer/go-livepeer/issues/2152

We revert it for now and will implement maxGasPrice in a different manner.

**What does this pull request do? Explain your changes. (required)**
N/A

**Specific updates (required)**
N/A

**How did you test each of these updates (required)**
Tested that it fixes the related bug.

**Does this pull request close any open issues?**
fix #2152

**Checklist:**

- [x] Read the [contribution guide](./doc/contributing.md)
- [x] `make` runs successfully
- [x] All tests in `./test.sh` pass
- [x] README and other documentation updated
- [x] [Pending changelog](./CHANGELOG_PENDING.md) updated

Issue #2152, opened 2021-12-27
Session crashes (without errors in log) right after the O sends a tx


0xVires:
Seeing some strange behavior with the latest livepeer version:
All my sessions crash without anything showing up in the logs. All the transcoding just seems to stop:
```
Dec 26 23:53:24 livepeer[129136]: I1226 23:53:24.710594  129136 orchestrator.go:766] manifestID=19782171-0479-4561-be44-2c3417c1c94d seqNo=370 orchSessionID=304da1aa Successfully received results from remote transcoder=x:59400 segments=2 taskId=737779 fname=https://x:8935/stream/304da1aa/370.tempfile dur=463.42726ms
Dec 26 23:53:26 livepeer[129136]: I1226 23:53:26.322271  129136 orchestrator.go:766] manifestID=09dcsceke8mmnotk seqNo=135 orchSessionID=66d7a6b4 Successfully received results from remote transcoder=x:44522 segments=4 taskId=737781 fname=https://x:8935/stream/66d7a6b4/135.tempfile dur=109.798038ms
Dec 27 00:29:26 livepeer[129136]: 2021/12/27 00:29:26 http: TLS handshake error from 143.244.33.78:59807: EOF
```
Only in the T logs I see a few `Transcode loop timed out for key=ebf7dc72_0`  and `Deleted transcode session for key=ebf7dc72_0`, but also no errors. Some more info here: https://discord.com/channels/423160867534929930/426114749370204170/924479708605861920

The weird thing is that both of those crashes happened just a minute or two after the O sent (or was supposed to send) a tx. The first time right after the start of a new round, the second time right after the O claimed a winning ticket.

Crash 1 (new round started 4min before 19:00):
![image](https://user-images.githubusercontent.com/25792899/147423827-5f57bc45-2dca-4370-9f03-0c4051e82bef.png)

Crash 2 (winning ticket tx was mined at 23:50):
![image](https://user-images.githubusercontent.com/25792899/147423976-0288bfdf-e35c-4838-a30e-eadf6aac3bbc.png)

Any guesses why my sessions crash without anything showing up in the logs?

Strykar:
I see similar behavior approximately 6 minutes after a winning ticket was claimed.
`Dec 26 16:32:21 i5 livepeer[3604]: Invoking transaction: "redeemWinningTicket". Inputs: "_ticket: { Recipient: 0x1a...`
[...]

PR #2171, opened 2022-01-10
Make transactions compatible with Arbitrum and fix setting max gas price

### LINKED ISSUE #2084: Set -maxGasPrice flag equal to max fee per gas, current implementation makes spending unpredictable

### LINKED ISSUE #2243: Update transaction format for Arbitrum
Arbitrum treats the gas price differently than Ethereum. It uses `gasPrice` as the maximum gas price that a user may pay. Also Aribtrum probably does not accept the post-london upgrade transaction format.

We need address the following points:
- Gas Price Monitor (oracle) does not make sense anymore
- Transaction Replacement does not make sense anymore
- All places where we use baseFee do not make sense anymore (there is no baseFee in Arbitrum)
- Setting max gas price is different

leszko:

Fix setting max gas price for both legacy tx and dynamic fee tx. Additionally, fix the transaction replacement for the legacy tx.

This PR uses `TransactOpts` to set `GasFeeCap`. There was already another approach merged to solve this issue by modifying `TransactionManager` https://github.com/livepeer/go-livepeer/pull/2111, but it caused to stop processing streams https://github.com/livepeer/go-livepeer/issues/2152, and therefore had to be reverted https://github.com/livepeer/go-livepeer/pull/2163.

- Add `TransactOpts` to the `Client` struct
- Make all bonding contracts use `TransactOpts` from `Client` instead of the default one
- Add lock to guard any changes to `TransactOpts` in `Client`
- Update replace logic in `transactionManager` to cover legacy tx

- Tested with [devtool](https://github.com/livepeer/go-livepeer/tree/master/cmd/devtool)
  - Check that transactions have `maxFeePerGas` set to the value defined by user `maxGasPrice`
  - Check that streams do not stop to be processed when `maxGasPrice` is set
  - Checl that if `maxGasPrice` is reset (set to `0`), then `maxFeePerGas` has the default value again
- Tested with Rinkeby
  - Sample transaction: https://rinkeby.etherscan.io/tx/0xe5f5e326e843e487acfb4a2ea1c6c93b9a70e08fb8c7e17104510b1588fa9356
- Tested max gas price with Arbitrum Rinkeby Testnet
  - Sample transaction: https://testnet.arbiscan.io/tx/0x5a5cbd9d807de5735ded121a6c8f92cfb403f9ff0d6729cfaf9bfac79da88858

I didn't test the transaction replacement in Arbitrum Rinkeby Testnet

fix #2084
fix #2243

**Checklist:**

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -10,6 +10,7 @@
 - \#2208 Improve Feed PubSub: execute subscribers' blocking operations in separate goroutines (@leszko)
 - \#2222 Use L1 block number for Ticket Parameters and Round Initialization (@leszko)
 - \#2240 Backfill always from the back last seen in DB (instead of the last round block) (@leszko)
+- \#2171 Make transactions compatible with Arbitrum and fix setting max gas price (@leszko)
 
 #### Broadcaster
 
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -469,6 +469,7 @@ func main() {
 			glog.Errorf("Failed to set gas info on Livepeer Ethereum Client: %v", err)
 			return
 		}
+		client.SetMaxGasPrice(bigMaxGasPrice)
 
 		n.Eth = client
 
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -469,7 +469,10 @@ func main() {
 			glog.Errorf("Failed to set gas info on Livepeer Ethereum Client: %v", err)
 			return
 		}
-		client.SetMaxGasPrice(bigMaxGasPrice)
+		if err := client.SetMaxGasPrice(bigMaxGasPrice); err != nil {
+			glog.Errorf("Failed to set max gas price: %v", err)
+			return
+		}
 
 		n.Eth = client
 
--- a/eth/client.go
+++ b/eth/client.go
@@ -13,6 +13,7 @@ package eth
 //go:generate abigen --abi protocol/abi/LivepeerTokenFaucet.json --pkg contracts --type LivepeerTokenFaucet --out contracts/livepeerTokenFaucet.go
 //go:generate abigen --abi protocol/abi/Poll.json --pkg contracts --type Poll --out contracts/poll.go
 import (
+	"context"
 	"fmt"
 	"math/big"
 	"sort"
@@ -117,7 +118,7 @@ type LivepeerEthClient interface {
 	Sign([]byte) ([]byte, error)
 	SignTypedData(apitypes.TypedData) ([]byte, error)
 	SetGasInfo(uint64) error
-	SetMaxGasPrice(*big.Int)
+	SetMaxGasPrice(*big.Int) error
 }
 
 type client struct {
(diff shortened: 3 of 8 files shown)


yondonfu (changes requested):
I think there may be a problem with this approach since the `client` struct currently [embeds each of the binding sessions](https://github.com/livepeer/go-livepeer/blob/c50c0ccf3cec22b4e10eb70d1103348ff2f5a751/eth/client.go#L135). As a result, if the caller invokes a method from a binding session that isn't explicitly defined on `client`, the method will be invoked on the underlying binding session that defines it [1].

For example, consider the method [LivepeerTokenSession.Transfer()](https://github.com/livepeer/go-livepeer/blob/c50c0ccf3cec22b4e10eb70d1103348ff2f5a751/eth/contracts/livepeerToken.go#L597). At the moment, `Transfer()` is not defined on `client`, so a call to `client.Transfer()` (such as the one [here](https://github.com/livepeer/go-livepeer/blob/c50c0ccf3cec22b4e10eb70d1103348ff2f5a751/server/webserver.go#L1049)) is equivalent to `client.LivepeerTokenSession.Transfer()`. With the current changes in this PR, that call should result in a nil pointer error because the `TransactOpts` filed in `LivepeerTokenSession` is not set.

I suggest thinking about a way to handle this case.

[1] https://go.dev/doc/effective_go#embedding

leszko:
@yondonfu Good point! You're right that it can be very confusing because you'd always need to remember that you can only use functions that were overridden in the client.

I checked and you can't configure `abigen` to generate only `get` functions in the `session` objects. Another approach could be to not store `TransactOpts` in `client`, but to update each session with `TransactOpts` while calling `setTransactOpts()`. Something like this.

```golang
func (c *client) setTransactOpts(opts bind.TransactOpts) {
	c.transOptsMu.Lock()
	c.LivepeerTokenSession.TransactOpts = opts
	c.LivepeerTokenFaucetSession.TransactOpts = opts
	c.ServiceRegistrySession.TransactOpts = opts
	c.BondingManagerSession.TransactOpts = opts
	// ...
	c.transOptsMu.Unlock()
}
```

But I think it won't work because assignment operation is not thread-safe in Go and `TransactOpts` is used in other places in the auto-generated Go code.

Since I don't think you can embed structs partly in Go, then I think there is no better option than getting rid of embedding the sessions at all. It results in some boilerplate delegation functions, but at least it's pretty straightforward. I added this commit: https://github.com/livepeer/go-livepeer/pull/2171/commits/6cdba5ee52b445308b632338b722e23e07033d3c . Let me know if you see any better solution to it.

yondonfu (approved):
LGTM after resolving conflicts!

yondonfu:
> Let me know if you see any better solution to it.

Hm don't see an immediate better solution. The extra boilerplate is not great, but can live with it for now!

leszko:
I tried to test it in Arbitrum Testnet and got the following message, which I guess is a blocker for this PR.

```
maxFeePerGas or maxPriorityFeePerGas specified but london is not active yet
```

yondonfu (approved):
LGTM. Let's update the changelog entry to reflect that this PR also make updates to be compatible with Arbitrum.

PR #2174, opened 2022-01-11
Support lpms tests in install_ffmpeg.sh

cyberj0g:
A quick follow-up to #2135 
Motivation is to make all lpms tests work with new `install_ffmpeg.sh` by building for dev/test with default options (without explicit `--disable-...`). Production build will still have only necessary components enabled.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -18,7 +18,7 @@ require (
 	github.com/golang/protobuf v1.5.2
 	github.com/jaypipes/ghw v0.7.0
 	github.com/livepeer/livepeer-data v0.2.0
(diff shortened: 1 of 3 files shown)


yondonfu (approved):
LGTM!

Let's aim to follow the Git commit message convention where we prefix the message either with the package that is being changed or the general functionality that is being changed. This structure has made it helpful to understand previous history and in the event of a bug tracking down the commit that introduced the cause.

For example, for this PR we could do something like:

```
mod: Bump to livepeer/lpms@0eb91f2
install_ffmpeg: Enable components for LPMS tests
```

PR #2180, opened 2022-01-13
Generate contract bindings for Arbitrum + related fixes

leszko:
Regenerate contract bindings related to the confluence protocol upgrade.

- Regenerate contract bindings
- Update `withdrawFees` with additional parameters (related to https://github.com/livepeer/protocol/pull/511)
- Add `ArbitrumTestnetChainID` to livepeer cli test networks to enable the options "Get test LPT"

- Orchestrator registration + streaming workflow (as described [here](https://github.com/livepeer/internal-project-tracking/issues/287#issuecomment-1009922804))
- Calling `withdrawFees` from Livepeer CLI created [this transaction](https://testnet.arbiscan.io/tx/0x792ea8cdf312aa1959efdfae57ceb582cfe463332a1a903c4ba2b5f1b74501ce) 

fix https://github.com/livepeer/internal-project-tracking/issues/287

**Checklist:**


cmd/livepeer_cli/wizard_bond.go
```diff
@@ -329,5 +329,15 @@ func (w *wizard) withdrawStake() {
...
 func (w *wizard) withdrawFees() {
-	httpPost(fmt.Sprintf("http://%v:%v/withdrawFees", w.host, w.httpPort))
+	dInfo, err := w.getDelegatorInfo()
+	if err != nil {
+		glog.Errorf("Error getting delegator info: %v", err)
+		return
+	}
+
+	val := url.Values{
+		"amount": {fmt.Sprintf("%v", dInfo.Fees.String())},
```
yondonfu:
```suggestion
		"amount": {fmt.Sprintf("%v", dInfo.PendingFees.String())},
```

If an account has earned fees in recent round, then the `PendingFees` field [1] on the `lpTypes.Delegator` struct can be greater than the `Fees` field. This is because within the BondingManager contract each address has a `fees` state that is only updated when the address submits a transaction to the contract and when that happens the `fees` state is incremented by any additional fees that have been earned recently. Prior to this, you can calculate the fees for an address, which includes its current `fees` state plus any additional earned, by calling the `pendingFees()` function on the BondingManager contract [2]. 

So, we should use the `PendingFees` field here because it will represent the address' true total fees that can be withdrawn while the `Fees` field *might* be missing some fees if the address' `fees` state in the contract hasn't been updated in awhile.

Note: The same relationship exists between the `BondedAmount` and `PendingStake` fields with the BondingManager contract function `pendingStake()` calculating the address' current stake.

[1] https://github.com/livepeer/go-livepeer/blob/81a42873663bd34cbba007cb11a5554f84404f92/eth/client.go#L699
[2] https://github.com/livepeer/go-livepeer/blob/81a42873663bd34cbba007cb11a5554f84404f92/eth/client.go#L679

server/webserver.go
```diff
@@ -677,7 +677,23 @@ func (s *LivepeerServer) cliWebServerHandlers(bindAddr string) *http.ServeMux {
 
 	mux.HandleFunc("/withdrawFees", func(w http.ResponseWriter, r *http.Request) {
 		if s.LivepeerNode.Eth != nil {
-			tx, err := s.LivepeerNode.Eth.WithdrawFees()
+			if err := r.ParseForm(); err != nil {
```
yondonfu: Since we're updating this handler it could be a good opportunity for a small refactor which also improves testability of handlers now that we have additional helpers in the codebase for handling form values.

See this example [hander definition](https://github.com/livepeer/go-livepeer/blob/81a42873663bd34cbba007cb11a5554f84404f92/server/handlers.go#L92), [attaching it to the mux](https://github.com/livepeer/go-livepeer/blob/81a42873663bd34cbba007cb11a5554f84404f92/server/webserver.go#L1175) and its associated [tests](https://github.com/livepeer/go-livepeer/blob/81a42873663bd34cbba007cb11a5554f84404f92/server/handlers_test.go#L188).

An example of what attaching this handler to the mux could look like:

```
mux.Handle("/withdrawFees", mustHaveFormParams(withdrawFeesHandler(s.LivepeerNode.Eth), "amount")
```

WDYT?

yondonfu (changes requested):
Just two comments. Also, let's also sync `dev` with `master` to pull in some of the testing fixes so that CI won't consistently fail for this PR.

cmd/livepeer_cli/wizard_bond.go
leszko: Ha, good catch! Thanks for this comment.

server/webserver.go
leszko: Good idea. I'll refactor all the handlers.

CHANGELOG_PENDING.md
```diff
@@ -7,6 +7,8 @@
 ### Features ⚒
 
 #### General
+- \#2163 Fix Session crashes right after the O sends a tx, revert #2111 Ensure `maxFeePerGas` in all transactions never exceed `-masGasPrice` (@leszko)
```
yondonfu: I think this entry was inadvertently added and can be removed.

server/handlers.go
```diff
@@ -72,30 +89,20 @@ func currentBlockHandler(getter BlockGetter) http.Handler {
...
 }
 
 func fundDepositAndReserveHandler(client eth.LivepeerEthClient) http.Handler {
-	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		if client == nil {
-			respondWith500(w, "missing ETH client")
-			return
-		}
-
+	return mustHaveClient(client, http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
yondonfu, line 105: 👏  for extracting this check. I've been getting tired of doing that nil check heh.

yondonfu (approved):
LGTM! Just one comment about removing an extraneous CHANGELOG_PENDING entry - feel free to remove prior to merging.

PR #2189, opened 2022-01-18
Point ffmpeg to with fix for fast verification

darkdarkdragon:
**What does this pull request do? Explain your changes. (required)**
Uses latest FFMpeg

**Specific updates (required)**
Changed install_ffmpeg.sh

**How did you test each of these updates (required)**
Unit test in LPMS repo

**Does this pull request close any open issues?**
Fixes #2175 

**Checklist:**

- [ ] Read the [contribution guide](./doc/contributing.md)
- [ ] `make` runs successfully
- [ ] All tests in `./test.sh` pass
- [ ] README and other documentation updated
- [ ] [Pending changelog](./CHANGELOG_PENDING.md) updated

PR #2190, opened 2022-01-18
Update transcoded segments length to fix tests

cyberj0g:
Cherry picked some changes from #2150 to fix tests.

### DIFF
--- a/core/core_test.go
+++ b/core/core_test.go
@@ -79,10 +79,10 @@ func TestTranscode(t *testing.T) {
 	}
 
 	// 	Check transcode result
-	if Over1Pct(len(tr.TranscodeData.Segments[0].Data), 64484) { // 144p
+	if Over1Pct(len(tr.TranscodeData.Segments[0].Data), 218268) { // 144p
 		t.Error("Unexpected transcode result ", len(tr.TranscodeData.Segments[0].Data))
 	}
-	if Over1Pct(len(tr.TranscodeData.Segments[1].Data), 88172) { // 240p
+	if Over1Pct(len(tr.TranscodeData.Segments[1].Data), 302868) { // 240p
 		t.Error("Unexpected transcode result ", len(tr.TranscodeData.Segments[1].Data))
 	}
 
--- a/core/transcoder_test.go
+++ b/core/transcoder_test.go
@@ -33,10 +33,10 @@ func TestLocalTranscoder(t *testing.T) {
 	if len(res.Segments) != len(videoProfiles) {
 		t.Error("Mismatched results")
 	}
-	if Over1Pct(len(res.Segments[0].Data), 152280) {
+	if Over1Pct(len(res.Segments[0].Data), 522264) {
 		t.Errorf("Wrong data %v", len(res.Segments[0].Data))
 	}
-	if Over1Pct(len(res.Segments[1].Data), 208116) {
+	if Over1Pct(len(res.Segments[1].Data), 715528) {
 		t.Errorf("Wrong data %v", len(res.Segments[1].Data))
 	}
 }


darkdarkdragon (approved):
LGTM

iameli:
Merged to unblock merging of https://github.com/livepeer/go-livepeer/pull/2189

core/core_test.go
```diff
@@ -79,10 +79,10 @@ func TestTranscode(t *testing.T) {
 	}
 
 	// 	Check transcode result
-	if Over1Pct(len(tr.TranscodeData.Segments[0].Data), 64484) { // 144p
+	if Over1Pct(len(tr.TranscodeData.Segments[0].Data), 218268) { // 144p
```
yondonfu, line 82: @cyberj0g Out of curiosity - any idea why the size of the transcoded data is so much bigger now? I know you suspected that this is related to the recent LPMS change that explicitly set the bitrate for h264. Does the size increase match expectations?

cyberj0g: The size increase is because libx264 is in [ABR](https://github.com/livepeer/FFmpeg/blob/7d09579190def3ef7562399489e628f3b65714ce/libavcodec/libx264.c#L644) (average bitrate) mode now. Before, we didn't set `bitrate` parameter, only [`rc_max_rate` and `rc_buffer_size`](https://github.com/livepeer/lpms/blob/0eb91f2facdb40c1bd06401015b274315263436b/ffmpeg/encoder.c#L238). With this configuration libx264 used CQP mode to control quality, but we don't set QP-related parameters explicitly. I believe, current settings are closer to 'controlling quality through bitrate' which probably what we want to achieve. Plus, it seem to work well with h264_nvenc, libx265, hevc_nvenc. It may be worth further research, though, because I'm not sure what's the best practice for online streaming. The intuition is 'black screen' test segment we are using should have much lower than average bitrate after encoding.

Thulinma: This is a bit of a... tricky area of encoding parameters. Constant bit rate encoding implies that a very complex scene with lots of movement should take the same amount of bits per second as a black screen. That is great if you want a very predictable network load for your stream, but absolutely horrible if you want to guarantee a certain level of quality. On the other hand, constant quality encoding will change the bit rate depending on content (black frame -> very low, complex scene / lots of "noise" -> very high).

In general, the most common/effective approach is to use constant quality with a maximum bit rate it will not go above. But for streaming specifically - especially over UDP - a _truly_ constant bit rate is better, as UDP behaves better with a predictable load. Most broadcast applications also mandate a truly constant bit rate.

In short... there's no "this is always best" answer, and if possible we should allow users to select what mode they want.

yondonfu: @cyberj0g Thanks for the explanation! I just wanted to make sure we understand the reason why the size of the transcoded data increased.

So, my understanding from a glance:

- The [source segment in the test](https://github.com/livepeer/go-livepeer/blob/9ab7a8842f2f4c12a0ca9b88cd8ef9b6a1bc1b73/core/core_test.go#L29) is 402696 bytes
- The [video profiles in the test](https://github.com/livepeer/go-livepeer/blob/9ab7a8842f2f4c12a0ca9b88cd8ef9b6a1bc1b73/core/core_test.go#L37) set the bitrates of the outputs to [400k for the 144p profile](https://github.com/livepeer/lpms/blob/0eb91f2facdb40c1bd06401015b274315263436b/ffmpeg/videoprofile.go#L91) and [600k for the 240p profile](https://github.com/livepeer/lpms/blob/0eb91f2facdb40c1bd06401015b274315263436b/ffmpeg/videoprofile.go#L88)
- Previously, the output transcoded data for both profiles was significantly smaller. And I believe that is because libx264 was not in ABR mode so there was no attempt to try to control the bitrate based on the value set in the profile resulting in a much smaller video?
- Now, the output transcoded data for both profiles is larger than before, but still significantly smaller than the original segment. And I believe that is because with libx264 in ABR mode there is an attempt to control the bitrate based on the profile?

@Thulinma Just to clarify - is it correct to say the ABR mode that Ivan mentioned for libx264 is equivalent to the constant bitrate encoding that you describe since it focuses on controlling bitrate while giving up more in terms of controlling quality?

For now, unless we expect/observe problems with this change in bitrate mode, I think we can leave this as-is, but consider exposing the option to set the bitrate/quality tradeoff separately.

Thulinma: Sorry for the late reply on this, my inbox was a bit flooded. 😲
Yes, the method described by Ivan is roughly equivalent to constant bit rate encoding.

Do note: your understanding above is more or less inverted from reality. Before, ABR mode was active (controllable through a maximum bit rate ceiling and a target sliding average), causing bit rate to drop below the intended bit rate if the visual content did not need a higher rate (thus adapting the bit rate down for simpler visuals: ABR). Now, we're effectively in CBR mode, and the same bit rate is used regardless of visual content.
Technically speaking both are ABR, but the new config makes it _behave_ more akin to CBR. Hence the increased size.

yondonfu: Got it - thanks for the explanation!

PR #2197, opened 2022-01-20
install_ffmpeg.sh: add -j`nproc` to speed up compilation

iameli:
Hoping this makes the build quite a bit faster.

### DIFF
--- a/install_ffmpeg.sh
+++ b/install_ffmpeg.sh
@@ -4,6 +4,7 @@ set -ex
 
 ROOT="${1:-$HOME}"
 ARCH="$(uname -m)"
+NPROC=${NPROC:-$(nproc)}
 
 if [[ $ARCH == "arm64" ]] && [[ $(uname) == "Darwin" ]]; then
   # Detect Apple Silicon
@@ -55,8 +56,8 @@ if [[ $(uname) != *"MSYS"* ]] && [[ ! $IS_M1 ]]; then
     rm nasm-2.14.02.tar.gz nasm-2.14.02.tar.gz.sha256
     cd "$ROOT/nasm-2.14.02"
     ./configure --prefix="$ROOT/compiled"
-    make
-    make install || echo "Installing docs fails but should be OK otherwise"
+    make -j$NPROC
+    make -j$NPROC install || echo "Installing docs fails but should be OK otherwise"
   fi
 fi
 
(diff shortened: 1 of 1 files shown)


install_ffmpeg.sh
```diff
@@ -159,6 +159,6 @@ fi
 
 if [[ ! -e "$ROOT/ffmpeg/libavcodec/libavcodec.a" || $BUILD_TAGS == *"debug-video"* ]]; then
   cd "$ROOT/ffmpeg"
-  make $FFMPEG_MAKE_EXTRA_ARGS
-  make install
+  make -j$(nproc) $FFMPEG_MAKE_EXTRA_ARGS
```
cyberj0g: If we adding this, `$FFMPEG_MAKE_EXTRA_ARGS` could be deleted completely, it does the same

cyberj0g (commented):
Parallel compilation could be quite memory hungry, but I see it works fine on CircleCI, should be ok, then.

cyberj0g (approved):

install_ffmpeg.sh
yondonfu: Hm I actually think that it is a good thing that the current usage of `$FFMPEG_MAKE_EXTRA_ARGS` allows the `-j` flag to be toggled on/off (although this is only for running make for `ffmpeg` and it is set specifically to `-j4`). IIUC the change in this PR would result in this script always maxing out processing units when running make in this file. Since this flag is run locally as a part of the development process as well, I could see scenarios where someone might not necessarily want to dedicate all available resources to compilation especially if they are doing other things at the same time. Could an alternative be to make it easy to toggle `-j$(nproc)` on/off?

yondonfu (changes requested):

install_ffmpeg.sh
iameli: How about we do this at the top of the file?
```
NPROC=${NPROC:-$(nproc)}
```
And the make commands can be
```
make -j$NPROC
```
That way the default will be to use all CPU cores, but in development you could run
```
NPROC=1 ./install_ffmpeg.sh
```
and only use a single CPU core.

yondonfu: That sounds reasonable to me!

iameli:
$NPROC change made, ready for additional review.

install_ffmpeg.sh
```diff
@@ -159,6 +160,6 @@ fi
 
 if [[ ! -e "$ROOT/ffmpeg/libavcodec/libavcodec.a" || $BUILD_TAGS == *"debug-video"* ]]; then
   cd "$ROOT/ffmpeg"
-  make $FFMPEG_MAKE_EXTRA_ARGS
-  make install
+  make -j$NPROC $FFMPEG_MAKE_EXTRA_ARGS
```
yondonfu: Seems that we can remove `FFMPEG_MAKE_EXTRA_ARGS` here and in the definition earlier in the file as mentioned by @cyberj0g since all it does right now is set `-j4` which is no longer needed with this change.

iameli: Done!

yondonfu (approved):
LGTM!

PR #2199, opened 2022-01-20
Fix flaky session selection test

cyberj0g:
Refactor test to fix test case affecting subsequent test case.


yondonfu (commented):
Hm its not immediately clear to me why these changes fix the flaky test. I see that we extracted a bunch of logic from the `_SelectUnknownSession_RandFreq` test into its own function and then call it at the beginning of the test, but it doesn't seem like anything functionally changed in the test. Could you elaborate on how this change fixes the flaky test?

cyberj0g:
@yondonfu I moved test setup, where broadcast sessions are created, to separate function, which is now called twice, for each test case. Before that, if for loop in first test case randomly picked a session with max stake (with about 0.1 chance), next test case, which expects this session to be picked, failed.

yondonfu (approved):
> I moved test setup, where broadcast sessions are created, to separate function, which is now called twice, for each test case. Before that, if for loop in first test case randomly picked a session with max stake (with about 0.1 chance), next test case, which expects this session to be picked, failed.

Ah I missed that second call to the function to create a new selector.

LGTM after squashing the commits!

Issue #2207, opened 2022-01-25
Improve Feed PubSub mechanism


leszko:
## Abstract

In `go-livepeer` we use the in-process PubSub mechanism, based on the [event.Feed structure](https://pkg.go.dev/github.com/ethereum/go-ethereum/event#Feed), which distribute events between different goroutines. We use it with an assumption that it's async, but in reality, it's a blocking mechanism. This spec describes a set of activities to fix this issue.

## Motivation

There were a number of bugs reported, which most probably are related to this issue:
- https://github.com/livepeer/go-livepeer/issues/2152
- https://github.com/livepeer/go-livepeer/issues/2079

A sample scenario to experience the issue is to start an orchestrator and send a transaction which is not been minted for a long time. Then one of the PubSub subscribers waits on [client.CheckTx](https://github.com/livepeer/go-livepeer/blob/c2c73fa6f24473311766baee83e2762c87ef708d/eth/client.go#L870), which blocks the PubSub sender, which in result makes the orchestrator unable to do anything, even transcoding a segment.

## Proposed Solution

This diagram presents how we use the Feed PubSub mechanism in which each feed is a PubSub topic.

![image](https://user-images.githubusercontent.com/2834997/150986190-78420b43-0ca7-46fb-959a-3bb4e28ee676.png)

If anything blocks in any subscriber, it can block the whole graph. Why is it so?

The [Feed subscription mechanism](https://pkg.go.dev/github.com/ethereum/go-ethereum/event#Feed.Subscribe) works with Go channels and when the channel is full, the sender is blocked. The recommendation from the doc is:

> The channel should have ample buffer space to avoid blocking other subscribers. Slow subscribers are not dropped.

However, we have these 2 issues:

1. Subscribers execute blocking operations
2. Blocking operations don't use any deadline/timeout

As an example, we can see [client.CheckTx() in rewardservice](https://github.com/livepeer/go-livepeer/blob/v0.5.26/eth/rewardservice.go#L94). It's blocking and does not use any timeout.

We should address both of these issues, so make sure that each subscriber does not block and add timeouts for all blocking operations.

## Implementation Tasks and Considerations

1. Execute subscribers' blocking operations in separate goroutines
   - PR: #2208
2. Add timeouts for all blocking operations

## Testing Tasks and Considerations

Testing scenarios:
1. Insert `time.Sleep(1 * time.Hour)` into `CheckTx()` and other operations and check that the transcoding is not blocked
[...]

PR #2208, opened 2022-01-25
Improve Feed PubSub: execute subscribers' blocking operations in separate goroutines

### LINKED ISSUE #2168: Improve TimeWatcher PubSub mechanism
Currently, blocking any of the [watchers](https://github.com/livepeer/go-livepeer/tree/master/eth/watchers) may result in completely blocking the whole orchestrator. E.g. not processing a transaction may result in the orchestrator's inability to transcode streams.

Here is the scenario:
1. Any of watchers blocks
    - e.g. [roundinitializer in client.CheckTx()](https://github.com/livepeer/go-livepeer/blob/c2c73fa6f24473311766baee83e2762c87ef708d/eth/roundinitializer.go#L124)
2. It causes the channel used for the feed subscriptions to fill up
    - e.g. [blockSink in roundinitializer](https://github.com/livepeer/go-livepeer/blob/c2c73fa6f24473311766baee83e2762c87ef708d/eth/roundinitializer.go#L49)
3. In result, any send to the given feed is blocked
    - e.g. [blockwatcher sending the current block update](https://github.com/livepeer/go-livepeer/blob/c2c73fa6f24473311766baee83e2762c87ef708d/eth/blockwatch/block_watcher.go#L205)
4. The whole feed mechanism stops working
    - e.g. block number is not updated every `blockPollingTime` which implies the wrong block number in `TicketParams` which implies not accepting segments to transcode

We need to address it because it makes the whole feed mechanism very unreliable. One solution to this would be to make sure that watchers never block in their code. Another solution would be to change the feed pubsub mechanism to not block if the target channel is full.

N/A

**Additional context**
This issue was first discovered during analyzing the bug https://github.com/livepeer/go-livepeer/issues/2152

leszko:
Avoid blocking Feed mechanism by executing every blocking operation in a separate goroutine. As a result, if any operation blocks, then we have a blocked/leaked goroutine instead of stopping the whole livepeer.

See #2207 for a detailed description.

- Serve each feed event in a separate goroutine
- Add locks to avoid multiple goroutines performing the same operations more than once
- Renaming the feed sink variables to keep the same naming convention in all files

Checked the same scenario using the code from `master` and the code from this PR.

1. Insert `time.Sleep(1 * time.Hour)` into `roundinitializer.go#TryInitialize()` to simulate that this operation is blocked
2. Start local Geth, Orchestrator, Broadcaster, and a video stream.
3. From Livepeer CLI, execute a few times `Initialize Round` and `Reward`.
4. After 5 min check the logs from orchestrator

In the case of `master`, orchestrator was blocked and stopped transcoding. Broadcaster failed transcoding with the following logs.

```
I0125 11:23:11.586426    6783 mediaserver.go:925] manifestID=movie nonce=17482076527360272128 seqNo=0 Finished push request at url=http://localhost:8935/live/movie/test197.ts ua=Go-http-client/1.1 addr=127.0.0.1:57864 bytes=995648 dur= resolution= took=26.432ms
I0125 11:23:11.586430    6783 broadcast.go:122] manifestID=movie nonce=17482076527360272128 seqNo=0 Starting session refresh
I0125 11:23:11.586697    6783 rpc.go:265] manifestID=movie nonce=17482076527360272128 Connecting RPC to uri=https://127.0.0.1:8936
I0125 11:23:11.588929    6783 db_discovery.go:129] manifestID=movie nonce=17482076527360272128 invalid ticket params orch=https://127.0.0.1:8936 err="TicketParams expired"
```

In the case of code from this PR, everything worked, even though the roundinitializer was blocked.

fix #2168

**Checklist:**

### DIFF
--- a/eth/watchers/orchestratorwatcher.go
+++ b/eth/watchers/orchestratorwatcher.go
@@ -42,12 +42,12 @@ func NewOrchestratorWatcher(bondingManagerAddr ethcommon.Address, watcher BlockW
 
 // Watch starts the event watching loop
 func (ow *OrchestratorWatcher) Watch() {
-	roundEvents := make(chan types.Log, 10)
-	roundSub := ow.tw.SubscribeRounds(roundEvents)
+	roundSink := make(chan types.Log, 10)
+	roundSub := ow.tw.SubscribeRounds(roundSink)
 	defer roundSub.Unsubscribe()
 
-	events := make(chan []*blockwatch.Event, 10)
-	sub := ow.watcher.Subscribe(events)
+	blockSink := make(chan []*blockwatch.Event, 10)
+	sub := ow.watcher.Subscribe(blockSink)
 	defer sub.Unsubscribe()
 
 	for {
@@ -56,13 +56,13 @@ func (ow *OrchestratorWatcher) Watch() {
 			return
 		case err := <-sub.Err():
 			glog.Error(err)
-		case events := <-events:
+		case block := <-blockSink:
 			go func() {
-				ow.handleBlockEvents(events)
+				ow.handleBlockEvents(block)
 			}()
-		case roundEvent := <-roundEvents:
+		case round := <-roundSink:
 			go func() {
-				if err := ow.handleRoundEvent(roundEvent); err != nil {
+				if err := ow.handleRoundEvent(round); err != nil {
 					glog.Errorf("error handling new round event: %v", err)
 				}
 			}()
--- a/eth/watchers/orchestratorwatcher.go
+++ b/eth/watchers/orchestratorwatcher.go
@@ -57,11 +57,15 @@ func (ow *OrchestratorWatcher) Watch() {
 		case err := <-sub.Err():
 			glog.Error(err)
 		case events := <-events:
-			ow.handleBlockEvents(events)
+			go func() {
+				ow.handleBlockEvents(events)
+			}()
 		case roundEvent := <-roundEvents:
-			if err := ow.handleRoundEvent(roundEvent); err != nil {
-				glog.Errorf("error handling new round event: %v", err)
-			}
+			go func() {
+				if err := ow.handleRoundEvent(roundEvent); err != nil {
+					glog.Errorf("error handling new round event: %v", err)
+				}
+			}()
 		}
 	}
 }
--- a/eth/watchers/orchestratorwatcher.go
+++ b/eth/watchers/orchestratorwatcher.go
@@ -60,9 +60,7 @@ func (ow *OrchestratorWatcher) Watch() {
 		case err := <-sub.Err():
 			glog.Error(err)
 		case block := <-blockSink:
-			go func() {
-				ow.handleBlockEvents(block)
-			}()
+			go ow.handleBlockEvents(block)
 		case round := <-roundSink:
 			go func() {
 				if err := ow.handleRoundEvent(round); err != nil {
--- a/eth/watchers/orchestratorwatcher.go
+++ b/eth/watchers/orchestratorwatcher.go
@@ -3,6 +3,7 @@ package watchers
 import (
 	"math"
 	"math/big"
+	"sync"
 
 	ethcommon "github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/types"
@@ -22,6 +23,8 @@ type OrchestratorWatcher struct {
 	lpEth   eth.LivepeerEthClient
 	tw      timeWatcher
 	quit    chan struct{}
+	blockMu sync.Mutex
+	roundMu sync.Mutex
 }
 
 func NewOrchestratorWatcher(bondingManagerAddr ethcommon.Address, watcher BlockWatcher, store common.OrchestratorStore, lpEth eth.LivepeerEthClient, tw timeWatcher) (*OrchestratorWatcher, error) {
@@ -94,6 +97,9 @@ func (ow *OrchestratorWatcher) handleLog(log types.Log) error {
 		// Noop if we cannot find the event name
 		return nil
 	}
+
+	ow.blockMu.Lock()
+	defer ow.blockMu.Unlock()
 
 	switch eventName {
 	case "TranscoderActivated":
@@ -168,6 +174,9 @@ func (ow *OrchestratorWatcher) handleTranscoderDeactivated(log types.Log) error 
 }
 
 func (ow *OrchestratorWatcher) handleRoundEvent(log types.Log) error {
+	ow.roundMu.Lock()
+	defer ow.roundMu.Unlock()
+
 	round, err := ow.lpEth.CurrentRound()
 	if err != nil {
 		return err
--- a/pm/queue.go
+++ b/pm/queue.go
@@ -91,9 +91,11 @@ func (q *ticketQueue) Length() (int, error) {
 // is sufficient to cover the face value of the ticket at the head of the queue. If the max float is sufficient, we pop
 // the ticket at the head of the queue and send it into q.redeemable which an external listener can use to receive redeemable tickets
 func (q *ticketQueue) startQueueLoop() {
-	blockNums := make(chan *big.Int, 10)
-	sub := q.tm.SubscribeBlocks(blockNums)
+	blockSink := make(chan *big.Int, 10)
+	sub := q.tm.SubscribeBlocks(blockSink)
 	defer sub.Unsubscribe()
+	blockNumCh := make(chan *big.Int)
+	defer close(blockNumCh)
 
 ticketLoop:
 	for {
@@ -102,7 +104,11 @@ ticketLoop:
 			if err != nil {
 				glog.Errorf("Block subscription error err=%q", err)
 			}
-		case latestBlock := <-blockNums:
+		case blockNum := <-blockSink:
+			go func() {
+				blockNumCh <- blockNum
+			}()
+		case latestBlock := <-blockNumCh:
 			numTickets, err := q.Length()
 			if err != nil {
 				glog.Errorf("Error getting queue length err=%q", err)
--- a/pm/queue.go
+++ b/pm/queue.go
@@ -94,65 +94,64 @@ func (q *ticketQueue) startQueueLoop() {
 	blockSink := make(chan *big.Int, 10)
 	sub := q.tm.SubscribeBlocks(blockSink)
 	defer sub.Unsubscribe()
-	blockNumCh := make(chan *big.Int)
-	defer close(blockNumCh)
 
-ticketLoop:
 	for {
 		select {
 		case err := <-sub.Err():
 			if err != nil {
 				glog.Errorf("Block subscription error err=%q", err)
 			}
-		case blockNum := <-blockSink:
+		case block := <-blockSink:
 			go func() {
-				blockNumCh <- blockNum
+				q.handleBlockEvent(block)
 			}()
-		case latestBlock := <-blockNumCh:
-			numTickets, err := q.Length()
-			if err != nil {
-				glog.Errorf("Error getting queue length err=%q", err)
-				continue
-			}
-			for i := 0; i < int(numTickets); i++ {
-				nextTicket, err := q.store.SelectEarliestWinningTicket(q.sender, new(big.Int).Sub(q.tm.LastInitializedRound(), big.NewInt(ticketValidityPeriod)).Int64())
-				if err != nil {
-					glog.Errorf("Unable select earliest winning ticket err=%q", err)
-					continue ticketLoop
-				}
-				if nextTicket == nil {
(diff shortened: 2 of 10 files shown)


server/redeemer.go
```diff
@@ -120,31 +121,46 @@ func (r *Redeemer) MaxFloat(ctx context.Context, req *net.MaxFloatReq) (*net.Max
...
+					if err != nil {
+						errResMu.Lock()
+						if errRes == nil {
+							errRes = status.Error(codes.Internal, err.Error())
+						}
+						errResMu.Unlock()
+						cancel()
+					}
+				}
+			}()
```
victorges: Are all these operations thread-safe or might we need a lock here as well?

server/redeemer.go
```diff
@@ -120,31 +121,46 @@ func (r *Redeemer) MaxFloat(ctx context.Context, req *net.MaxFloatReq) (*net.Max
 func (r *Redeemer) MonitorMaxFloat(req *net.MaxFloatReq, stream net.TicketRedeemer_MonitorMaxFloatServer) error {
 	sender := ethcommon.BytesToAddress(req.Sender)
 
+	ctx, cancel := context.WithCancel(stream.Context())
+	// first error noticed is returned
+	var errRes error
+	var errResMu sync.Mutex
```
victorges: **nit:** You could consider having an error channel here instead of the ctx+error+mutex. 

Then from the `MaxFloat` event goroutine you can simply `errCh <- err` and from your main select case you can have a `case err := errCh: return err`. WDYT?

To avoid deadlock sending to the channel you can give it a buffer of 1, since we only return 1 error anyway, and surround the sending code with a select like
```
select {
  case errCh <- err:
  default:
    // someone already sent an error first, ignore
}
```

pm/queue.go
```diff
@@ -91,62 +94,70 @@ func (q *ticketQueue) Length() (int, error) {
...
-					glog.Errorf("Unable select earliest winning ticket err=%q", err)
-					continue ticketLoop
-				}
-				if nextTicket == nil {
-					continue ticketLoop
-				}
+		case block := <-blockSink:
+			go func() {
+				q.handleBlockEvent(block)
+			}()
```
victorges: **nit:** No need for the anonymous function in this case
```suggestion
			go q.handleBlockEvent(block)
```

(thanks for creating a helper btw, much needed)

This applies to all other places where the anonymous` func() { }` only does 1 call to another function and ignores the result.

victorges (approved):
Reviewing only the go logic for processing events in a separate go-routine. My understanding is that it works just because we don't stop receiving from the src channel and allow it to fill up, blocking the sender. That looks good to me, just make sure that the order of the events really do not matter, as mutexes are not FIFO.

It would be good to avoid holding so many (locked) goroutines in background. At the same time as you could use a channel for the locking instead, with which you could provide rate-limiting and timeouts, it could also be a premature optimization right now that we don't really need. So if we do implement timeouts for all the internal operations of those routines (which I understood is the next step here), we should be good already.

victorges:
Thinking further about this, WDYT instead of providing a helper function somewhat like this? It basically works as a channel with a dynamically-sized, growing buffer.
```go
func pipeBuffered(ctx context.Context, in <-chan interface{}) <-chan interface{} {
	out := make(chan interface{}, 1)
	go func() {
		defer close(out)
		buf := []interface{}{}
	pipe:
		for {
			if len(buf) == 0 {
				select {
				case <-ctx.Done():
					return
				case val, ok := <-in:
					if !ok {
						break pipe
					}
					buf = append(buf, val)
				}
				continue
			}
			select {
			case <-ctx.Done():
				return
			case val, ok := <-in:
				if !ok {
					break pipe
				}
				buf = append(buf, val)
			case out <- buf[0]:
				buf = buf[1:]
			}
		}
		for val := range buf {
			select {
			case <-ctx.Done():
				return
			case out <- val:
			}
		}
	}()
	return out
}
```

With it, you would pass the subscription channel and it would return a channel which only has a "direct" buffer of 1, but actually has a resizable buffer kept as an off-channel slice. It would keep the events ordering while still holding more control over the parallelism, and a simpler usage in the end (just the way it was before). For example, later we could define a hard limit to the slice buffer and have events be dropped if we reach it. Or maybe add a log in case the buffer gets too big. 

Implementation can be optimized on how it grows/shrinks the slice, for example with a circular buffer, but I guess it could be enough for now. Performance-wise, I think it would even be a little cheaper than maintaining N locked goroutines, given that each goroutine will also need to hold at least 1 event in the stack, which is everything that we hold in the slice buffer.

leszko:
> Thinking further about this, WDYT instead of providing a helper function somewhat like this? It basically works as a channel with a dynamically-sized, growing buffer.
> 
> ```go
> func pipeBuffered(ctx context.Context, in <-chan interface{}) <-chan interface{} {
> 	out := make(chan interface{}, 1)
> 	go func() {
> 		defer close(out)
> 		buf := []interface{}{}
> 	pipe:
> 		for {
> 			if len(buf) == 0 {
> 				select {
> 				case <-ctx.Done():
> 					return
> 				case val, ok := <-in:
> 					if !ok {
> 						break pipe
> 					}
> 					buf = append(buf, val)
> 				}
> 				continue
> 			}
> 			select {
> 			case <-ctx.Done():
> 				return
> 			case val, ok := <-in:
> 				if !ok {
> 					break pipe
> 				}
> 				buf = append(buf, val)
> 			case out <- buf[0]:
> 				buf = buf[1:]
> 			}
> 		}
> 		for val := range buf {
> 			select {
> 			case <-ctx.Done():
> 				return
> 			case out <- val:
> 			}
> 		}
> 	}()
> 	return out
> }
> ```
> 
> With it, you would pass the subscription channel and it would return a channel which only has a "direct" buffer of 1, but actually has a resizable buffer kept as an off-channel slice. It would keep the events ordering while still holding more control over the parallelism, and a simpler usage in the end (just the way it was before). For example, later we could define a hard limit to the slice buffer and have events be dropped if we reach it. Or maybe add a log in case the buffer gets too big.
> 
> Implementation can be optimized on how it grows/shrinks the slice, for example with a circular buffer, but I guess it could be enough for now. Performance-wise, I think it would even be a little cheaper than maintaining N locked goroutines, given that each goroutine will also need to hold at least 1 event in the stack, which is everything that we hold in the slice buffer.

This looks very nice at first, but... Golang (at least the version we use in the project) does not have generics and therefore it'll not bake in nicely into our watchers. We would need to create a similar function for each chan parameter that we use (which is probably overkill).

At first, I thought we could change all the Subscribe functions to use `sink chan<- interface{}` instead of `sink chan<- types.Log` (or similar typed channels), but it will not work as well, because you actually need to pass a typed channel into `feed.Subscribe()`. Otherwise `feed.Send()` does not work.

leszko:
> Reviewing only the go logic for processing events in a separate go-routine. My understanding is that it works just because we don't stop receiving from the src channel and allow it to fill up, blocking the sender. That looks good to me, just make sure that the order of the events really do not matter, as mutexes are not FIFO.

Yes. I think that the order of events does not matter. Also, these locks are actually kind-of optimization to not execute the same transaction at the same time. It's enough to execute it once, so trying to send it twice at the same time may result in some gas burnt unnecessarily.

> 
> It would be good to avoid holding so many (locked) goroutines in background. At the same time as you could use a channel for the locking instead, with which you could provide rate-limiting and timeouts, it could also be a premature optimization right now that we don't really need. So if we do implement timeouts for all the internal operations of those routines (which I understood is the next step here), we should be good already.

Yes, I think it's technically possible to convert it into a channel-based solution, I just find it more difficult. Check my comment into your solution with `pipeBuffered()`.

server/redeemer.go
leszko: Yes, good idea. Updated in https://github.com/livepeer/go-livepeer/pull/2208/commits/88832228bbf1c51d21db9a95d5f98ca85b2e47c8

server/redeemer.go
leszko: To me, they all look thread-safe. It's a bunch of RPC calls, and if there are some fields accessed, they are already protected with a mutex.

pm/queue.go
leszko: Good point, updated in https://github.com/livepeer/go-livepeer/pull/2208/commits/01585d35d7570ac937e248c2bf954033f49166d6

victorges:
> This looks very nice at first, but... Golang (at least the version we use in the project) does not have generics and therefore it'll not bake in nicely into our watchers. We would need to create a similar function for each chan parameter that we use (which is probably overkill).

Yeah makes sense! Can't wait for generics to reach a stable version 🤩

victorges (approved):
LGTM!

eth/watchers/orchestratorwatcher.go
```diff
@@ -164,6 +171,9 @@ func (ow *OrchestratorWatcher) handleTranscoderDeactivated(log types.Log) error
 }
 
 func (ow *OrchestratorWatcher) handleRoundEvent(log types.Log) error {
+	ow.mu.Lock()
+	ow.mu.Unlock()
```
yondonfu: Need a defer here or else this will just unlock immediately after locking in the previous line.

eth/watchers/orchestratorwatcher.go
```diff
@@ -91,6 +95,9 @@ func (ow *OrchestratorWatcher) handleLog(log types.Log) error {
 		return nil
 	}
 
+	ow.mu.Lock()
```
yondonfu: I wonder if it would make sense to have separate mutexes for the handleBlockEvents vs. handleRoundEvent goroutines so they could be executed concurrently if needed. Not sure if this would present any problems. Using a single mutex works, but just throwing this out there for awareness + separate consideration.

server/redeemer.go
```diff
@@ -129,22 +132,28 @@ func (r *Redeemer) MonitorMaxFloat(req *net.MaxFloatReq, stream net.TicketRedeem
 			return nil
 		case <-stream.Context().Done():
 			return nil
+		case err := <-errCh:
```
yondonfu, line 135: Hm this will cause this loop to exit if `stream.Send()` returns an error. Seems like we should log the error and allow the loop to continue running in case the `stream.Send()` error was ephemeral (i.e. temp problem with the remote service). WDYT?

yondonfu (changes requested):

eth/watchers/orchestratorwatcher.go
leszko: Wow, good catch! Updated in https://github.com/livepeer/go-livepeer/pull/2208/commits/34cb4770ab34960766ffdd3f54fb51fb3585b0ae

eth/watchers/orchestratorwatcher.go
leszko: Yeah, I like this idea. Changed to two mutexes. The code does not share any variables, so IMO it's safe. Updated in https://github.com/livepeer/go-livepeer/pull/2208/commits/34cb4770ab34960766ffdd3f54fb51fb3585b0ae

server/redeemer.go
leszko: Hmm, good question. This is actually the same behavior as it was before https://github.com/livepeer/go-livepeer/blob/master/server/redeemer.go#L145

I think quitting the loop may actually make sense here because this function is a handler for an RPC request and it streams the data to the caller. So, kind-of of opening a connection and polling the data. If not possible to send some data to the sender, we may actually close the connection and the sender needs to retry. Wdyt? 

But I definitely don't have the whole picture of who is calling this RPC call and why.

leszko (commented):
@yondonfu I addressed your comments, PTAL

server/redeemer.go
yondonfu: Ah I see. In that case, keeping the current behavior for now seems fine.

> But I definitely don't have the whole picture of who is calling this RPC call and why.

The caller for this RPC method would be an O that is outsourcing ticket redemption to a node that is running a standalone Redeemer service. Normally, an O would monitor the max float [1] of Bs that are sending it tickets (i.e. payments) with a stream. If an O is outsourcing ticket redemption to a Redeemer then an O will subscribe to max float updates for Bs that are published by the Redeemer which becomes the one responsible for monitoring the max float of Bs. 

[1] Think of the "max float" as the maximum sum of value across winning tickets from B that O is willing to hold because that value is guaranteed to be available in B's on-chain reserve funds. See https://github.com/livepeer/wiki/blob/master/spec/streamflow/pm.md#reserve for more details.

yondonfu (approved):
LGTM! Looks like the commit structure could use a bit of clean up before merging.

server/redeemer.go
leszko: Ahh, I see. I didn't even know that we can run Livepeer as a standalone Redeemer service! Thanks for the explanation :)

PR #2218, opened 2022-01-28
Display http status codes in livepeer_cli

### LINKED ISSUE #2217: Return http response codes in livepeer_cli responses
I was using `livepeer_cli` and i got error:
```ERROR: logging before flag.Parse: E0128 02:49:31.223383      72 wizard_stats.go:98] Error getting current round: http response status not ok```

I have tried to understand what's the problem for a moment and checked the code to check why would it occure:
```
	if resp.StatusCode != http.StatusOK {
		return nil, errors.New("http response status not ok")
	}
```
As you can see from a very wide information about the error, we only return one: was connection successful or not.

In my opinion it would be nice for users and also livepeer contributors to just return status codes in response.
e.g: `Error getting current round: http response status 418`

noisersup:
Adds to error messages in livepeer_cli  their status codes for better error understanding.

- changed error message in `wizard.currentRound` and `wizard.currentBlock`

standard `./test.sh` check

Fixes #2217 

**Checklist:**

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -9,6 +9,7 @@
 #### General
 
 - \#2216 Fix Accept Multiline message on Windows (@leszko)
+- \#2218 Display http status codes in livepeer_cli (@noisersup)
 
 #### Broadcaster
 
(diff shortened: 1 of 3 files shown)


leszko (approved):
LGTM

PR #2219, opened 2022-01-28
drivers: Make Size an optional property of files

victorges:
The Filebase S3-compatible API does not return a
file size in the response for files greater than 255 bytes.

That caused a panic on s3.go:267 when reading recordings.

With this change, the problem won't happen anymore cause
we don't even try to dereference the ContentLength, but
hold it as a nilable pointer that should be checked wherever
it is used (right now: nowhere).

- Make `FileInfo.Size` a pointer
- Change every place that sets it to set a pointer instead of raw int value

 - `make` 
 - Check all usages of `Size` property (there are none apart from the writes)

Related to https://github.com/livepeer/livepeer-com/issues/879

**Checklist:**

### DIFF
--- a/drivers/drivers.go
+++ b/drivers/drivers.go
@@ -43,7 +43,7 @@ type FileInfo struct {
 	Name         string
 	ETag         string
 	LastModified time.Time
-	Size         int64
+	Size         *int64
 }
 
 type FileInfoReader struct {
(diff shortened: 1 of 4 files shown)


victorges:
Filebase fixed the issue on their side. Was caused by undesired GZIPing of text-based responses larger than 255 bytes. 

Assuming that the `Content-Length` property is always present in the response instead, I think we can close this safely.

PR #2222, opened 2022-02-01
Use L1 block number for Ticket Parameters and Round Initialization

### LINKED ISSUE #2220: Track last seen L1 block in addition to last seen L2 block

There are a few places where `LastSeenBlock()` is either called either on the [TimeWatcher](https://github.com/livepeer/go-livepeer/blob/970d0aa7892cd9c30aaf069a7ee8e72d8fa64111/eth/watchers/timewatcher.go#L140) or the [DB](https://github.com/livepeer/go-livepeer/blob/970d0aa7892cd9c30aaf069a7ee8e72d8fa64111/common/db.go#L440). In the former case, the value returned is read from memory. In the latter case, the value returned is read from a sqlite3 DB [1]. When the node is connected to Arbitrum, in both cases, the method will return the *L2 block number*. This is because the node receives the latest block number using the [blockwatch package](https://github.com/livepeer/go-livepeer/blob/master/eth/blockwatch/block_watcher.go) which will poll for new block headers from the RPC provider. When connected to an Arbitrum RPC provider, the block headers will be for the Arbitrum L2 chain so the number included in each header will be the L2 block number. While it makes sense for the node to be monitoring L2 blocks for events that it cares about (since the events would be emitted from transactions included in those L2 blocks), the use of L2 block numbers in `LastSeenBlock()` can be problematic for a few reasons:

1. Os use the last seen block to set expiration params for tickets

- [The expiration block](https://github.com/livepeer/go-livepeer/blob/7836d1ce43ee4aabb2fc71ac29572018aceec564/pm/recipient.go#L200) for ticket params is set based on the last seen block
- During ticket validation, the expiration block is compared against [the last seen block](https://github.com/livepeer/go-livepeer/blob/7836d1ce43ee4aabb2fc71ac29572018aceec564/pm/recipient.go#L170)
- Bs will check the expiration block of ticket params against [the last seen block](https://github.com/livepeer/go-livepeer/blob/c218058f8d03200c8ad19f196bdacffae5cd09f6/pm/sender.go#L157) to determine whether they need to refresh a set of ticket params

Using a L2 block number can be problematic because ticket params expire after [a fixed # of blocks](https://github.com/livepeer/go-livepeer/blob/7836d1ce43ee4aabb2fc71ac29572018aceec564/pm/recipient.go#L24) (this value should probably be higher as well) which was originally set based on average L1 block times of ~13s. The current 10 blocks would be roughly (10 * 13) ~= 2m. L2 blocks are created much faster and can also be created in bursts so at the moment they do not seem as a reliable "clock" that nodes can coordinate around.

2. The round initializer service uses the last seen block to determine whether to submit a round initialization tx

- The [last seen block](https://github.com/livepeer/go-livepeer/blob/970d0aa7892cd9c30aaf069a7ee8e72d8fa64111/eth/roundinitializer.go#L100) is used to calculate a seed value to determine if the address for the node is responsible for submitting the tx

Using a L2 block number can be problematic because the start and end of rounds are still set in terms of L1 block numbers. So, in order for the round initialize to determine where we are in the length of a round a L2 block number wouldn't be helpful.

Note: With all this being said, there *are* cases where using the L2 block number from `LastSeenBlock()` is useful. The main case I see right now is to [get the last L2 block that the node has seen to determine which block headers that it should fetch on startup](https://github.com/livepeer/go-livepeer/blob/6c575bcc14fc48727ed48d9f7ec74edfc87d18e8/cmd/livepeer/livepeer.go#L469).

[1] The original motivation for having an in-memory *and* an on-disk cache was to read from the former when the value is needed on a latency sensitive code path and to read from the latter when not on a latency sensitive code path.

Since it seems useful for the node to be aware of its last seen L1 block number *and* last seen L2 block number, we could consider adding a `LastSeenL1Block()` method to complement the existing `LastSeenBlock()` method. I think we just need to implement this on `TimeWatcher` as that impl is used for the above cases, but I would double check if it should be implemented on `DB` as well in case the method on `DB` would be useful anywhere.

`LastSeenBlock()` should continue returning the last seen block number from the connected RPC provider.

`LastSeenL1Block()` should return the last seen L1 block number. If the node is connected to a L1 RPC provider this will be the same value as `LastSeenBlock()`. If the node is connected to a L2 RPC provider this will be a different value from `LastSeenBlock()`.

How will the node get the L1 block number if it is connected to a L2 RPC provider? Luckily, it looks like Arbitrum L2 RPC providers will include the L1 block number in a `l1BlockNumber` field in the response for the `eth_getBlockByNumber` RPC call. Use this command to confirm:

```
curl -X POST --data '{
    "id": 1337,
    "jsonrpc": "2.0",
    "method": "eth_getBlockByNumber",
    "params":["latest", true]
}' https://rinkeby.arbitrum.io/rpc | jq .
```

We could update the `blockwatch` package to also parse this `l1BlockNumber` field when fetching block headers.

The blockwatcher currently uses `blockwatch.Client` to send RPC requests to the RPC provider for block headers. Luckily, the [HeaderByNumber](https://github.com/livepeer/go-livepeer/blob/c6563c780f4ce09f0fa76db6f6d0cca636653eb1/eth/blockwatch/client.go#L55) method uses a raw RPC call for `eth_getBlockByNumber`. A good place to start could be to update the [struct](https://github.com/livepeer/go-livepeer/blob/c6563c780f4ce09f0fa76db6f6d0cca636653eb1/eth/blockwatch/client.go#L47) that the response is unmarshalled into to parse `l1BlockNumber` and store it as a field in the struct. Then, we'd have access to this value in the rest of the package and we can access it elsewhere in the codebase as is needed i.e. storing in-memory in `TimeWatcher`.

N/A

**Additional context**

N/A

leszko:
Use L1 block numbers during Round Initialization and in Ticket Params.

Important comments:
- The L1 block numbers are not stored in DB and therefore need to be enriched with L1 Block Number [using Arbitrum RPC API](https://github.com/leszko/go-livepeer/blob/2dbae16237ff82706324975e96050e28aea884a3/eth/blockwatch/block_watcher.go#L625); an alternative approach would be to store L1 Block Numbers in DB:
  - it would mean fewer calls to Aribtrum RPC API
  - it would be a backward-incompatible change
  - I think enriching blocks on-the-fly is good enough
- `TimeWatcher` now has a function `LastSeenL1Block()`, but I found it does not need to have `LastSeenBlock()` anymore
- Having now both `blocks` and `l1Blocks` in the code can be super-confusing, I tried to rename all the necessary parts of the code, so that `block` always refers to L2 Block, however, there are still a places for which we may need an additional renaming in the future:
   - [TicketParams](https://github.com/leszko/go-livepeer/blob/7125e3c9778622f216441baa7b951cbd009943db/pm/ticket.go#L34)
   - [lp_rpc.proto](https://github.com/leszko/go-livepeer/blob/4fceaff0fce6bed65dae12428006408dc5064e15/net/lp_rpc.proto#L1)
   - Eth contracts or/and contract bindings
- There is one place where we could think about optimizing more, which is to try to fetch L1 Number while fetching logs in [this part](https://github.com/leszko/go-livepeer/blob/079f4ebe0801327fec6664cd5756100cf19e4d82/eth/blockwatch/block_watcher.go#L398)), but it doesn't seem to be an issue at the moment.

- Add L1 Block Number while fetching blocks by Block Watcher from ETH API
- Use L1 Block Number in Time Watcher (and propagate it to Ticket Sender and Recipient)
- Rename all `block` into `l1Block` when it refers to L1 Blocks
- Refactor Block Watcher Client (reuse ETH API code)

Tests on Arbitrum:
1. Added debug logs to Round Initializer to check that it uses L1 block numbers
2. Add debug logs to ticket params to check that that sender and receiver use  L1 block numbers for ticket params

fix #2220

**Checklist:**

### DIFF
--- a/eth/blockwatch/block_watcher.go
+++ b/eth/blockwatch/block_watcher.go
@@ -97,7 +97,7 @@ func (w *Watcher) BackfillEventsIfNeeded(ctx context.Context) error {
 		return err
 	}
 	if len(events) > 0 {
-		w.blockFeed.Send(events)
+		w.blockFeed.Send(w.enrichWithL1(events))
 	}
 	return nil
 }
@@ -137,7 +137,11 @@ func (w *Watcher) Subscribe(sink chan<- []*Event) event.Subscription {
 
 // GetLatestBlock returns the latest block processed
 func (w *Watcher) GetLatestBlock() (*MiniHeader, error) {
-	return w.stack.Peek()
+	h, err := w.stack.Peek()
+	if err != nil {
+		return nil, err
+	}
+	return w.enrichWithL1BlockNumber(h)
 }
 
 // InspectRetainedBlocks returns the blocks retained in-memory by the Watcher instance. It is not
@@ -202,7 +206,7 @@ func (w *Watcher) pollNextBlock() error {
 	// Even if an error occurred, we still want to emit the events gathered since we might have
 	// popped blocks off the Stack and they won't be re-added
 	if len(events) != 0 {
-		w.blockFeed.Send(events)
+		w.blockFeed.Send(w.enrichWithL1(events))
 	}
 	if err != nil {
 		return err
@@ -607,6 +611,38 @@ func (w *Watcher) filterLogsRecursively(from, to int, allLogs []types.Log) ([]ty
 	}
 	allLogs = append(allLogs, logs...)
 	return allLogs, nil
+}
+
+// enrichWithL1 adds L1 block number to the event with the highest block number.
+//
+// Adding L1 block number to an event requires an RPC call and the only code using L1 block number is TimeWatcher that
+// is interested only in the highest L1 block number, not in all of them.
+func (w *Watcher) enrichWithL1(events []*Event) []*Event {
+	if len(events) == 0 {
+		return events
+	}
+
+	max := 0
+	for i, e := range events {
+		if events[max].BlockHeader.Number.Cmp(e.BlockHeader.Number) <= 0 {
+			max = i
+		}
+	}
+
+	block, err := w.enrichWithL1BlockNumber(events[max].BlockHeader)
+	if err != nil {
+		glog.Errorf("Cannot fetch L1 block number for block number %d err=%q", events[max].BlockHeader.Number, err)
+	}
+
+	events[max].BlockHeader = block
+	return events
+}
+
+func (w *Watcher) enrichWithL1BlockNumber(header *MiniHeader) (*MiniHeader, error) {
+	if header == nil || header.L1BlockNumber != nil {
+		return header, nil
+	}
+	return w.client.HeaderByHash(header.Hash)
 }
 
 func isUnknownBlockErr(err error) bool {
--- a/eth/blockwatch/client.go
+++ b/eth/blockwatch/client.go
@@ -3,6 +3,7 @@ package blockwatch
 import (
 	"context"
 	"errors"
+	"fmt"
 	"math/big"
 	"time"
 
@@ -44,33 +45,39 @@ func NewRPCClient(rpcURL string, requestTimeout time.Duration) (*RPCClient, erro
 	return &RPCClient{rpcClient: rpcClient, client: ethClient, requestTimeout: requestTimeout}, nil
 }
 
-type getBlockByNumberResponse struct {
-	Hash       common.Hash `json:"hash"`
-	ParentHash common.Hash `json:"parentHash"`
-	Number     string      `json:"number"`
+type getHeaderResponse struct {
+	Hash          common.Hash `json:"hash"`
+	ParentHash    common.Hash `json:"parentHash"`
+	Number        string      `json:"number"`
+	L1BlockNumber string      `json:"l1BlockNumber"`
 }
 
 // HeaderByNumber fetches a block header by its number. If no `number` is supplied, it will return the latest
 // block header. If no block exists with this number it will return a `ethereum.NotFound` error.
 func (rc *RPCClient) HeaderByNumber(number *big.Int) (*MiniHeader, error) {
-	ctx, cancel := context.WithTimeout(context.Background(), rc.requestTimeout)
-	defer cancel()
-
 	var blockParam string
 	if number == nil {
 		blockParam = "latest"
 	} else {
 		blockParam = hexutil.EncodeBig(number)
 	}
-	shouldIncludeTransactions := false
 
-	// Note(fabio): We use a raw RPC call here instead of `EthClient`'s `BlockByNumber()` method because block
-	// hashes are computed differently on Kovan vs. mainnet, resulting in the wrong block hash being returned by
-	// `BlockByNumber` when using Kovan. By doing a raw RPC call, we can simply use the blockHash returned in the
-	// RPC response rather than re-compute it from the block header.
-	// Source: https://github.com/ethereum/go-ethereum/pull/18166
-	var header getBlockByNumberResponse
-	err := rc.rpcClient.CallContext(ctx, &header, "eth_getBlockByNumber", blockParam, shouldIncludeTransactions)
+	return rc.callEth("eth_getBlockByNumber", blockParam)
+}
+
+// HeaderByHash fetches a block header by its block hash. If no block exists with this number it will return
+// a `ethereum.NotFound` error.
+func (rc *RPCClient) HeaderByHash(hash common.Hash) (*MiniHeader, error) {
+	return rc.callEth("eth_getBlockByHash", hash)
+}
+
+func (rc *RPCClient) callEth(method string, arg interface{}) (*MiniHeader, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), rc.requestTimeout)
+	defer cancel()
+
+	var header getHeaderResponse
+	err := rc.rpcClient.CallContext(ctx, &header, method, arg, false)
+
 	if err != nil {
 		return nil, err
 	}
@@ -79,31 +86,24 @@ func (rc *RPCClient) HeaderByNumber(number *big.Int) (*MiniHeader, error) {
 		return nil, ethereum.NotFound
 	}
 
+	// If no L1BlockNumber, then Livepeer is running on L1, so L1BlockNumber is the same as BlockNumber
+	if header.L1BlockNumber == "" {
+		header.L1BlockNumber = header.Number
+	}
+
 	blockNum, ok := math.ParseBig256(header.Number)
 	if !ok {
 		return nil, errors.New("Failed to parse big.Int value from hex-encoded block number returned from eth_getBlockByNumber")
 	}
-	miniHeader := &MiniHeader{
-		Hash:   header.Hash,
-		Parent: header.ParentHash,
-		Number: blockNum,
-	}
-	return miniHeader, nil
-}
-
-// HeaderByHash fetches a block header by its block hash. If no block exists with this number it will return
-// a `ethereum.NotFound` error.
-func (rc *RPCClient) HeaderByHash(hash common.Hash) (*MiniHeader, error) {
-	ctx, cancel := context.WithTimeout(context.Background(), rc.requestTimeout)
-	defer cancel()
-	header, err := rc.client.HeaderByHash(ctx, hash)
-	if err != nil {
-		return nil, err
+	l1BlockNum, ok := math.ParseBig256(header.L1BlockNumber)
+	if !ok {
+		return nil, fmt.Errorf("Failed to parse big.Int value from hex-encoded L1 block number returned from %v", method)
 	}
 	miniHeader := &MiniHeader{
-		Hash:   header.Hash(),
-		Parent: header.ParentHash,
-		Number: header.Number,
+		Hash:          header.Hash,
+		Parent:        header.ParentHash,
+		Number:        blockNum,
+		L1BlockNumber: l1BlockNum,
 	}
 	return miniHeader, nil
 }
--- a/eth/blockwatch/testdata/fake_client_block_poller_fixtures.json
+++ b/eth/blockwatch/testdata/fake_client_block_poller_fixtures.json
@@ -801,6 +801,48 @@
           "parent": "0x7ac62ad1dc23753eb35ccbe193fdf1ad33828a68ed1c497a17ae66f7baa98144",
           "number": 11
         }
+      },
+      "getBlockByHash": {
+        "0x22a63eebe1b8f4034c3552502c0ba4ffbd613b6ee3cd5b530f75202ddc6d6220": {
+          "hash": "0x22a63eebe1b8f4034c3552502c0ba4ffbd613b6ee3cd5b530f75202ddc6d6220",
+          "parent": "0x2804206056bff166dbeef754799a248c48dd9521b21ee5898b7632837069200c",
+          "number": 5
+        },
+        "0x85fe65e36f4cc457a774cfe500ba7f3efaa6535dd6cb80373b3e91978ecaa01a": {
+          "hash": "0x85fe65e36f4cc457a774cfe500ba7f3efaa6535dd6cb80373b3e91978ecaa01a",
+          "parent": "0x22a63eebe1b8f4034c3552502c0ba4ffbd613b6ee3cd5b530f75202ddc6d6220",
+          "number": 6
+        },
+        "0xc308739706b1cf523663cbd1d8ce01208f7811cc6ff2005d23a13898f2f5adff": {
+          "hash": "0xc308739706b1cf523663cbd1d8ce01208f7811cc6ff2005d23a13898f2f5adff",
+          "parent": "0x85fe65e36f4cc457a774cfe500ba7f3efaa6535dd6cb80373b3e91978ecaa01a",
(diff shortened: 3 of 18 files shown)


eth/blockwatch/client.go
```diff
@@ -79,31 +85,24 @@ func (rc *RPCClient) HeaderByNumber(number *big.Int) (*MiniHeader, error) {
...
-// a `ethereum.NotFound` error.
-func (rc *RPCClient) HeaderByHash(hash common.Hash) (*MiniHeader, error) {
-	ctx, cancel := context.WithTimeout(context.Background(), rc.requestTimeout)
-	defer cancel()
-	header, err := rc.client.HeaderByHash(ctx, hash)
-	if err != nil {
-		return nil, err
+	l1BlockNum, ok := math.ParseBig256(header.L1BlockNumber)
+	if !ok {
+		return nil, errors.New("Failed to parse big.Int value from hex-encoded L1 block number returned from eth_getBlockByNumber")
```
yondonfu:
```suggestion
		return nil, fmt.Errorf("Failed to parse big.Int value from hex-encoded L1 block number returned from %v", method)
```

since the method name is a variable.

eth/blockwatch/testdata/fake_client_block_poller_fixtures.json
```diff
@@ -801,6 +801,48 @@
           "parent": "0x7ac62ad1dc23753eb35ccbe193fdf1ad33828a68ed1c497a17ae66f7baa98144",
           "number": 11
         }
+      },
+      "getBlockByHash": {
```
yondonfu, line 805: Out of curiosity how was this data generated?

eth/blockwatch/block_watcher.go
```diff
@@ -387,6 +395,11 @@ func (w *Watcher) getMissedEventsToBackfill(ctx context.Context) ([]*Event, erro
 					Number: big.NewInt(0).SetUint64(log.BlockNumber),
 					Logs:   []types.Log{},
 				}
+				// TODO: Consider extracting L1 block number directly from logs if possible
```
yondonfu: I don't think this is possible right now. If you call `eth_getTransactionReceipt` on Arbitrum Rinkeby with the hash 0xe22f5ac3db57a421e6c00853c206440193fd5de5cba61c3876b88b8707b7c43e you'll see that the nested log objects in the logs array only contain the L2 block number and not the L1 block number :(

yondonfu (changes requested):
Just a few minor requests.

> The L1 block numbers are not stored in DB and therefore need to be enriched with L1 Block Number [using Arbitrum RPC API](https://github.com/leszko/go-livepeer/blob/2dbae16237ff82706324975e96050e28aea884a3/eth/blockwatch/block_watcher.go#L625); an alternative approach would be to store L1 Block Numbers in DB:

I was slightly concerned about this because the # of RPC requests from the node is already going to be higher when connected to Arbitrum Rinkeby/mainnet due to the higher frequency of blocks. If the request count increases significantly that can affect the costs of using certain ETH JSON-RPC providers - for example, Infura's pricing tiers are based on the # of RPC requests sent per day. However, I see that extra RPC requests are only added by enriching the MiniHeaders with the L1 block number when:

- We batch request logs when backfilling and there would be an additional RPC request per log. However, since the # of logs will be determined by the # of relevant logs found with the topics that the node cares about (this is what FilterLogs does by setting specific topics as parameters), if there are very few logs that the node cares about there would be very few additional RPC requests. I suspect this will be common right now
- In `buildCanonicalChain()`. I don't completely grok the # of additional RPC requests that would be required there. Would be good to get a sense of this

Generally, I think your point that this may be okay for now is reasonable, but I do think we should get a sense of the additional RPC requests for that last point that result from this enrichment process.

eth/blockwatch/block_watcher.go
leszko: Ok, let me remove this comment since we already know we can't address it.

eth/blockwatch/testdata/fake_client_block_poller_fixtures.json
leszko: I just copied it from the other part of this file.

leszko:
Good points @yondonfu!

I made one improvement in terms of the number of RPC calls here https://github.com/livepeer/go-livepeer/pull/2222/commits/1382f59ad78aa7f819ba84207d1d1ad4c90e5e6b I realized that we actually don't need to add L1 Block Number to every block we process, it's enough to add it into the block with the highest number. It makes the code a little more tricky, but on the other hand, we minimize the number of RPC calls we make. Let me know your thoughts.

With regards to your concerns.

> * We batch request logs when backfilling and there would be an additional RPC request per log. However, since the # of logs will be determined by the # of relevant logs found with the topics that the node cares about (this is what FilterLogs does by setting specific topics as parameters), if there are very few logs that the node cares about there would be very few additional RPC requests. I suspect this will be common right now

Right, especially after this change https://github.com/livepeer/go-livepeer/pull/2222/commits/1382f59ad78aa7f819ba84207d1d1ad4c90e5e6b, I think this part is not an issue.

> * In `buildCanonicalChain()`. I don't completely grok the # of additional RPC requests that would be required there. Would be good to get a sense of this

My understanding is that when the next block number is the child of the current block, then `buildCanonicalChain()` does not execute recursively and finishes [here](https://github.com/leszko/go-livepeer/blob/a0675ee52e57d8fbfbba35f77f6de1fadf19627c/eth/blockwatch/block_watcher.go#L244). Maybe I miss something, but I actually never encountered a case when it would reach the point of [enrichWithL1BlockNumber()](https://github.com/leszko/go-livepeer/blob/a0675ee52e57d8fbfbba35f77f6de1fadf19627c/eth/blockwatch/block_watcher.go#L251).

eth/blockwatch/block_watcher.go
```diff
@@ -395,8 +416,6 @@ func (w *Watcher) getMissedEventsToBackfill(ctx context.Context) ([]*Event, erro
 					Number: big.NewInt(0).SetUint64(log.BlockNumber),
 					Logs:   []types.Log{},
 				}
-				// Logs don't contain L1 block number, so we need to fetch it separately
-				blockHeader, err = w.enrichWithL1BlockNumber(blockHeader)
```
yondonfu: If we are backfilling don't we still need this so that the events generated from backfilling have L1 block numbers attached to the headers?

eth/blockwatch/block_watcher.go
```diff
@@ -206,14 +206,36 @@ func (w *Watcher) pollNextBlock() error {
...
+		w.blockFeed.Send(w.enrichWithL1(events))
 	}
 	if err != nil {
 		return err
 	}
 	return nil
 }
 
+// Add L1 Block Number to the event with the highest block number
+func (w *Watcher) enrichWithL1(events []*Event) []*Event {
```
yondonfu: Could be good to have some test cases for this.

yondonfu:
> it's enough to add it into the block with the highest number. It makes the code a little more tricky, but on the other hand, we minimize the number of RPC calls we make.

Because the block with the highest number is the latest block that we've received from polling and all other blocks we should've seen already? So, I think this should actually generally be true when using Arbitrum since re-orgs [1] shouldn't be a thing while Arbitrum is using a sequencer so we wouldn't expect `buildCanonicalChain()` to return a slice of events of length greater than 1 (i.e. should just contain an event for the latest block). Thus, I think your solution works for Arbitrum and since that is the focus and reason for fetching L1 block numbers in the first place. A easy way to still maintain compatibility with L1 may be to set the L1 block number on all the events in `enrichL1()` using the L1 block number fetched for the newest block since if the node is connected to L1 each event should have BlockNumber == L1BlockNumber.

[1] i.e. When certain blocks that have already occurred are discarded and new blocks take their place - transactions in the previous blocks generally end up eventually being included in the new blocks. This happens from time to time on L1 Ethereum.

> Maybe I miss something, but I actually never encountered a case when it would reach the point of [enrichWithL1BlockNumber()](https://github.com/leszko/go-livepeer/blob/a0675ee52e57d8fbfbba35f77f6de1fadf19627c/eth/blockwatch/block_watcher.go#L251).

See above point about re-orgs being a case to handle (which is why `buildCanonicalChain()` is implemented in the first place) on L1 Ethereum, but not being a problem with Arbitrum (at least that I am aware of).

leszko:
> [1] i.e. When certain blocks that have already occurred are discarded and new blocks take their place - transactions in the previous blocks generally end up eventually being included in the new blocks. This happens from time to time on L1 Ethereum.

Ahh, I missed this information. So, I start to understand the idea behind `buildCanonicalChain()` and the reason why I never actually encountered its recursive call. @yondonfu thanks for the explanation!

> > it's enough to add it into the block with the highest number. It makes the code a little more tricky, but on the other hand, we minimize the number of RPC calls we make.
> 
> Because the block with the highest number is the latest block that we've received from polling and all other blocks we should've seen already?

Because we're not interested in the L1 block number of all other blocks. Timewatcher is the only entity that actually reads the L1 block number and Timewatcher is [only interested in the highest L1 block number ](https://github.com/leszko/go-livepeer/blob/1382f59ad78aa7f819ba84207d1d1ad4c90e5e6b/eth/watchers/timewatcher.go#L209) as it checks:

```
last := tw.LastSeenL1Block()
new := event.BlockHeader.L1BlockNumber
if new != nil && (last == nil || last.Cmp(new) != 0) {
	tw.setLastSeenL1Block(new)
	tw.l1BlockSubFeed.Send(new)
}
```
It's the only place in the whole codebase where we read the L1 Block Number from the event. So, I think it's safe to enrich L1 only the event with the highest block number. Obviously assuming that The highest L2 Block number contains the highest L1 Block Number. But I understand it's the case, because of this sequencer thingy.

> A easy way to still maintain compatibility with L1 may be to set the L1 block number on all the events in `enrichL1()` using the L1 block number fetched for the newest block since if the node is connected to L1 each event should have BlockNumber == L1BlockNumber.

I think we don't need to set L1 Block Number on all events. It's already compatible with L1, since the block with the highest number is enriched with L1 Block Number (which is equal to Block Number in case of running on L1).

eth/blockwatch/block_watcher.go
leszko: This is no need to do it here anymore since we moved it to `enrichWithL1()` and enrich only the highest block number [here](https://github.com/leszko/go-livepeer/blob/1382f59ad78aa7f819ba84207d1d1ad4c90e5e6b/eth/blockwatch/block_watcher.go#L100)

eth/blockwatch/block_watcher.go
leszko: Added.

yondonfu:
> Obviously assuming that The highest L2 Block number contains the highest L1 Block Number. But I understand it's the case, because of this sequencer thingy.

Should be true.

> I think we don't need to set L1 Block Number on all events. It's already compatible with L1, since the block with the highest number is enriched with L1 Block Number (which is equal to Block Number in case of running on L1).

Hm let me try to walk through a scenario that I think might happen when connected to a L1 provider:

- The TimeWatcher receives a slice of events of length > 1 in `handleBlockEvents()` and calls [handleBlockNum()](https://github.com/livepeer/go-livepeer/blob/754c1ed203539275332363eb2cd2940a7fa6c72a/eth/watchers/timewatcher.go#L195) on each event
- The slice of events might have been published by the BlockWatcher because of an internal call to `buildCanonicalChain()` which needed to handle a re-org. For example, the highest block number might have been 10, but due to re-org, instead of the next block polling returning a slice with an event for block number 11 it returns a slice with events for block numbers 8, 9, 10 and 11 because blocks 8, 9 and 10 were removed and re-added in the re-org
- `handleBlockNum()` would be called for each of these blocks including block 11 and publish the block number to its own [feed](https://github.com/livepeer/go-livepeer/blob/754c1ed203539275332363eb2cd2940a7fa6c72a/eth/watchers/timewatcher.go#L215)
- As a result, subscribers to the TimeWatcher block number feed would receive blocks 8, 9, 10 and 11 and not just block 11
- If the other blocks don't have their `L1BlockNumber` fields set then TimeWatcher would return nil for `LastSeenL1BlockNum()` if that method is called before block 11 is processed by the TimeWatcher.

I'm not sure if existing synchronization prevents the last case from happening - thoughts?

yondonfu:
> I'm not sure if existing synchronization prevents the last case from happening - thoughts?

Discussed offline. The check in https://github.com/leszko/go-livepeer/blob/1382f59ad78aa7f819ba84207d1d1ad4c90e5e6b/eth/watchers/timewatcher.go#L209 will ensure that `handleBlockNum()` will only publish a block number to the TimeWatcher block number feed if the header has a non-nil `L1BlockNumber` field. So, in the example from the previous comment, blocks 8, 9 and 10 would not publish anything to the feed and only block 11 would publish something to feed because block 11 has a non-nil `L1BlockNumber`. Until block 11 is processed, `LastSeenL1BlockNum()` would return the last known L1 block number which would be whatever the L1 block number was for block 10 before it was re-orged and after it is processed it would return the next L1 block number for block 11. So, the concern I mentioned shouldn't be an issue.

yondonfu (approved):
LGTM after cleaning up commit history & resolving conflicts.

PR #2223, opened 2022-02-01
drivers: Refactor package as a more generic and performant library

victorges:
This is to make some improvements to the `drivers` package so that it can also be used
as a library from other projects, specifically the `task-runner` for VOD tasks. It is currently
using the drivers library for a platform-agnostic way of accessing Object Stores for reading
and writing VOD assets.

The greatest advantage of that is that we can just use the same Object Store API from
`livepeer-com` to declare where assets should be stored as well as stream recordings,
keeping the whole interface leaner.

- Make the `Size` property optional in our `File` representation. It is optional in the AWS SDK and is not returned in all implementations of it, like previous versions of Filebase, so it was required at some point (not anymore) but I still think it makes sense to keep it.
- Make the drivers receive an `io.Reader` instead of the whole buffered file. The reader is compatible with clients that have the whole bytes in memory, but requiring all clients to buffer the entire contents in memory is not ideal. This would have made the VOD tasks much more inefficient, needing to read the entire files in memory before sending to the OS.
- Use an S3 Uploader for uploading files to S3. This avoids the need of buffering the files before uploading, since the simpler upload SDK requires the file to be an `io.ReadSeeker` due to hashing purposes. The uploader is the suggested solution in [the open issue in their SDK repo](https://github.com/aws/aws-sdk-go/issues/915).

 - Fixed and then ran/passed all tests in `go-livepeer`
 - Used the `drivers` lib in the `task-runner` for VOD and already deployed it in production, it's working great
 - Did a bunch of performance tests with Storj folks to come up with the optimal configurations for the S3 uploader (which end up being good both for Storj and for GCP)

Part of https://github.com/livepeer/task-runner/issues/14

**Checklist:**

- [-] README and other documentation updated (not necessary IMO)

(does it need a changelog update?)

### DIFF
--- a/drivers/s3.go
+++ b/drivers/s3.go
@@ -10,7 +10,6 @@ import (
 	"encoding/hex"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"mime/multipart"
 	"net/http"
 	"path"
@@ -27,6 +26,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws/request"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/s3"
+	"github.com/aws/aws-sdk-go/service/s3/s3manager"
 )
 
 // S3_POLICY_EXPIRE_IN_HOURS how long access rights given to other node will be valid
@@ -45,6 +45,7 @@ type s3OS struct {
 	awsAccessKeyID     string
 	awsSecretAccessKey string
 	s3svc              *s3.S3
+	s3sess             *session.Session
 	useFullAPI         bool
 }
 
@@ -60,6 +61,7 @@ type s3Session struct {
 	storageType net.OSInfo_StorageType
 	fields      map[string]string
 	s3svc       *s3.S3
+	s3sess      *session.Session
 }
 
 func s3Host(bucket string) string {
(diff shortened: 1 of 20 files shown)


drivers/s3.go
```diff
@@ -112,10 +117,13 @@ func NewCustomS3Driver(host, bucket, region, accessKey, accessKeySecret string,
...
-		cfg := aws.NewConfig().WithRegion(os.region).WithCredentials(creds)
-		cfg = cfg.WithEndpoint(host)
-		cfg = cfg.WithS3ForcePathStyle(true)
-		os.s3svc = s3.New(session.New(), cfg)
+		cfg := aws.NewConfig().
+			WithRegion(os.region).
+			WithCredentials(creds).
+			WithEndpoint(host).
+			WithS3ForcePathStyle(true)
+		os.s3sess = session.New(cfg)
```
leszko: nit: I know it's not strictly related to your PR, but `session.New(cfg)` is deprecated, so consider changing to `session.NewSession(cfg)`.

drivers/s3.go
```diff
@@ -285,32 +294,40 @@ func (os *s3Session) saveDataPut(ctx context.Context, name string, data []byte,
...
 
-	params := &s3.PutObjectInput{
-		Bucket:        bucket,  // Required
-		Key:           keyname, // Required
-		Metadata:      metadata,
-		Body:          bytes.NewReader(data),
-		ContentType:   contentType,
-		ContentLength: aws.Int64(int64(len(data))),
+	uploader := s3manager.NewUploader(os.s3sess, func(u *s3manager.Uploader) {
+		u.Concurrency = 8
```
leszko: Consider extracting this magic number to some constants with the name. Or at least give here some comments about how you came to these numbers.

leszko (approved):
Added 2 comments, other than that LGTM 👍 

Also remember to:
- Update CHANGELOG_PENDING.md
- Consider squashing some commits before merging

drivers/s3.go
victorges: Cool! I had thought this would be a big change since it requires changing the output variables and left it as before, but in the end the main caller of this function (`ParseOSURL`) already returned an error itself. So only needed to change that and a couple of tests. Nice improvement.

drivers/s3.go
victorges: Sure!

PR #2229, opened 2022-02-03
add darwin arm64 build, bump to Go 1.17.6

iameli:
This PR adds Darwin cross-compilation support for supporting both legacy amd64 and M1 arm64 macs.

All of the pieces here — install_ffmpeg.sh, Makefile, and upload_build.sh, are controlled through the $GOARCH environment variable. Should make it easy. If we ever support full cross-compilation, controlling it through $GOOS makes sense to me too.

**This also includes a version bump to Go 1.17.6**. There was a specific [compilation bug in 1.15.5](https://github.com/golang/go/issues/42565) and Go didn't support darwin-arm64 targets until Go 1.16. But if we're upgrading anyway I figure it makes sense to go to the latest and greatest.

### DIFF
--- a/install_ffmpeg.sh
+++ b/install_ffmpeg.sh
@@ -5,9 +5,20 @@ set -ex
 ROOT="${1:-$HOME}"
 ARCH="$(uname -m)"
 NPROC=${NPROC:-$(nproc)}
+EXTRA_CFLAGS=""
+EXTRA_LDFLAGS=""
+EXTRA_FFMPEG_FLAGS=""
 
 if [[ $ARCH == "arm64" ]] && [[ $(uname) == "Darwin" ]]; then
   # Detect Apple Silicon
+  IS_M1=1
+fi
+if [[ $ARCH == "x86_64" ]] && [[ $(uname) == "Darwin" ]] && [[ "${GOARCH:-}" == "arm64" ]]; then
+  echo "cross-compiling darwin-arm64"
+  EXTRA_CFLAGS="$EXTRA_CFLAGS --target=arm64-apple-macos11"
+  EXTRA_LDFLAGS="$EXTRA_LDFLAGS --target=arm64-apple-macos11"
+  HOST_OS="--host=aarch64-darwin"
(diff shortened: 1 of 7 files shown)


install_ffmpeg.sh
```diff
@@ -5,11 +5,22 @@ set -ex
...
 NPROC=${NPROC:-$(nproc)}
+EXTRA_CFLAGS=""
+EXTRA_LDFLAGS=""
+EXTRA_FFMPEG_FLAGS=""
 
 if [[ $ARCH == "arm64" ]] && [[ $(uname) == "Darwin" ]]; then
   # Detect Apple Silicon
   IS_M1=1
 fi
+if [[ $ARCH == "x86_64" ]] && [[ $(uname) == "Darwin" ]] && [[ "${GOARCH:-}" == "arm64" ]]; then
```
hjpotter92, line 16: is the `$ARCH` check needed here?

iameli: I think the only consequence would be that this block would run during native M1 ARM64 builds, resulting in ffmpeg having `--enable-cross-compile` enabled even though it doesn't need it. I'm not 100% sure what that flag does.

hjpotter92 (approved):
lgtm!

hjpotter92:
> **This also includes a version bump to Go 1.17.6**. There was a specific [compilation bug in 1.15.5](https://github.com/golang/go/issues/42565) and Go didn't support darwin-arm64 targets until Go 1.16. But if we're upgrading anyway I figure it makes sense to go to the latest and greatest.

Should this also bump version in `go.mod`?

iameli:
> Should this also bump version in go.mod?

My understanding is that field should reflect the most recent version of Go we'd expect go-livepeer to work with — our CI after this PR is using 1.17 but we'd expect things to work since the last time there was a breaking change. Helps with people pulling us in as a library and whatnot.

iameli:
Held off on this until Confluence landed, merging now so we can get a darwin-arm64 build in the next semver release 😃

PR #2268, opened 2022-02-15
Add pixel-format in capability logic

AlexKordic:

Extending `GetCodecInfo()` in lpms [PR](https://github.com/livepeer/lpms/pull/303)

PixelFormat of input video is split into:
- decode capability: Chroma Subsampling
- decode capability: Color Depth in bits: `8`bit or `10`bit

VideoProfile struct is extended with ChromaSubsampling and ColorDepthBits . Json decoding and protobuf message is updated accordingly. 

`TestTranscoderCapabilities()` now check sample bitstreams for optional capabilites:
- `Capability_H264_Decode_444_8bit`
- `Capability_H264_Decode_422_8bit`
- `Capability_H264_Decode_444_10bit`
- `Capability_H264_Decode_422_10bit`
- `Capability_H264_Decode_420_10bit`

The default H264 420 8bit is defined for sample 

**What is missing**:
- ☑ Protobuf compile not done properly
- ☑ Test missing

- 
- 
- 

**Checklist:**


oscar-davids:
I have some questions and comments.

I guess that the purpose to add `GetCodecInfo `function is to support decoding when B is fed variable types of video, that's B inspects codec type and sends this video to the proper Orchestrator.

we have to be concerned about the decoding profiles [here](https://developer.nvidia.com/video-encode-and-decode-gpu-support-matrix-new) decoding matrix.
Every O registers decoding capacity into B, according to capacities,  B sends a stream to O which can decode this stream.
we can classify four types of decoding profiles such as H264-420(only 8bit) vp9, H265-420, H265-444, and each type is also classified 8,10,12 bit.  I think we should consider these in this implementation.
Capability_H264_Decode_444_8bit~Capability_H264_Decode_420_10bit capacities seems meaningless.

if my thought is wrong, would love any feedback.

AlexKordic:
Good question @oscar-davids, exactly the same reasoning we were applying in tackling this task. 

This PR is about H264 decode capabilities. It is true that nvidia hardware supports 420 8bit decode which is represented by [Capability_H264](https://github.com/livepeer/go-livepeer/blob/master/core%2Fcapabilities.go#L28). Makes sense to add other capability flags for `B` to fail early or choose `O` that provides software decoder.

The [TestCapability_TranscoderCapabilities](https://github.com/livepeer/go-livepeer/blob/ak/pixelformat_capability/core/capabilities_test.go#L112) validates nvidia decoder capabilities and software decoder capabilities.

You are right, other codecs are not yet addressed and some questions still unanswered. For example there is maximum resolution supported by decoder.

So far existing capability mechanism have shown to be good enough. 

We can expect to continually add more capabilities with new codecs and vendors.

server/mediaserver.go
```diff
@@ -848,11 +811,10 @@ func (s *LivepeerServer) HandlePush(w http.ResponseWriter, r *http.Request) {
 		ctx = clog.AddNonce(ctx, cxn.nonce)
 	}
```
oscar-davids, line 769: Maybe here we need redefine errorOut function.

core/capabilities.go
```diff
@@ -78,7 +83,7 @@ var CapabilityTestLookup = map[Capability]CapabilityTest{
 	// 145x145 is the lowest resolution supported by NVENC on Windows
 	Capability_H264: {
 		inVideoData: testSegment_H264,
-		outProfile:  ffmpeg.VideoProfile{Resolution: "145x145", Bitrate: "1000k", Format: ffmpeg.FormatMPEGTS},
+		outProfile:  ffmpeg.VideoProfile{Resolution: "146x146", Bitrate: "1000k", Format: ffmpeg.FormatMPEGTS},
```
oscar-davids, line 91: any special reason for this change?

server/mediaserver.go
AlexKordic: Sure, please give me some direction what could be improved?

core/capabilities_test.go
```diff
@@ -76,6 +79,64 @@ func TestCapability_CompatibleBitstring(t *testing.T) {
 	})
 }
 
+func decode_capabilities(caps []Capability) []string {
+	var found []string = make([]string, 0, 5)
```
oscar-davids: this function hasn't been used yet.

core/capabilities.go
AlexKordic: `ffmpeg` complains `width must be multiple of 2`. This being arbitrary small resolution it made sense to increment 1

core/capabilities_test.go
AlexKordic: I used it in debugging sessions and later decided to remove log output in tests. Do you think we should remove it ?

oscar-davids: I think it is right to remove the code that is not used in real production because it can be used for hacking.

server/mediaserver.go
oscar-davids: at line 829, used glog but now clog. so I think we should move function redefine here.
```suggestion
errorOut = func(status int, s string, params ...interface{}) {
		httpErr := fmt.Sprintf(s, params...)
		clog.Errorf(ctx, httpErr)
		http.Error(w, httpErr, status)
	}
```

AlexKordic: fixed

core/capabilities_test.go
AlexKordic: fixed

core/capabilities_test.go
```diff
@@ -76,6 +79,45 @@ func TestCapability_CompatibleBitstring(t *testing.T) {
...
+	return tmp, cleanup
+}
+
+func TestCapability_TranscoderCapabilities(t *testing.T) {
+	tmpdir, cleanup := setupWorkDir()
+	defer cleanup()
+
+	// nvidia test
+	devices, err := common.ParseNvidiaDevices("all")
+	devicesAvailable := err != nil && len(devices) > 0
```
oscar-davids: maybe missed an equal symbol
```suggestion
	devicesAvailable := err == nil && len(devices) > 0
```

core/transcoder.go
```diff
@@ -115,23 +115,44 @@ func (nv *NvidiaTranscoder) Transcode(ctx context.Context, md *SegTranscodingMet
...
+	if err == nil {
+		return name
+	}
+	return "unknown"
+}
+
+type continueLoop bool
+
+func forEachTranscoderSample(handler func(*transcodeTestParams) continueLoop) {
 	// default capabilities
```
oscar-davids, line 147: Nit: Could it be changed to a more explicit function name?
```suggestion
	func transcodeWithSample(handler func(*transcodeTestParams) continueLoop) 
```

core/capabilities_test.go
AlexKordic: good catch!

core/transcoder.go
AlexKordic: fixed

core/capabilities.go
```diff
@@ -185,6 +226,26 @@ func JobCapabilities(params *StreamParameters) (*Capabilities, error) {
...
+				caps[Capability_H264_Decode_444_8bit] = true
+			case cap_422_8bit:
+				caps[Capability_H264_Decode_422_8bit] = true
+			case cap_444_10bit:
+				caps[Capability_H264_Decode_444_10bit] = true
+			case cap_422_10bit:
+				caps[Capability_H264_Decode_422_10bit] = true
+			case cap_420_10bit:
+				caps[Capability_H264_Decode_420_10bit] = true
+			}
```
cyberj0g, line 249: `Properties()` seem to return much larger number of combinations, should we return error on unknown\unsupported combination here?

core/capabilities.go
cyberj0g: This makes sense, can we add it to the comment on line 83?

cyberj0g (changes requested):
Nice work @AlexKordic! Please address few small comments I added. Special thanks for refactoring. Side note: I think it's best to keep PRs smaller, to make reviews easier and rebasing\merging less painful.

oscar-davids (approved):
LGTM!

common/util.go
```diff
@@ -224,17 +225,30 @@ func FFmpegProfiletoNetProfile(ffmpegProfiles []ffmpeg.VideoProfile) ([]*net.Vid
 		} else {
 			gop = int32(profile.GOP.Milliseconds())
 		}
+		chromaFormat := net.VideoProfile_CHROMA_420
+		switch profile.ChromaFormat {
+		case ffmpeg.ChromaSubsampling420:
+			// chromaFormat := net.VideoProfile_CHROMA_420
```
thomshutt: Why is this commented?

core/capabilities_test.go
```diff
@@ -100,6 +142,42 @@ func TestCapability_JobCapabilities(t *testing.T) {
...
+	// assert.True(checkPixelFormat(ffmpeg.PixelFormatYUV422P16BE, []Capability{Capability_AuthToken, Capability_H264, }))
+	// assert.True(checkPixelFormat(ffmpeg.PixelFormatYUV444P16LE, []Capability{Capability_AuthToken, Capability_H264, }))
+	// assert.True(checkPixelFormat(ffmpeg.PixelFormatYUV444P16BE, []Capability{Capability_AuthToken, Capability_H264, }))
+	// assert.True(checkPixelFormat(ffmpeg.PixelFormatYUV420P12BE, []Capability{Capability_AuthToken, Capability_H264, }))
+	// assert.True(checkPixelFormat(ffmpeg.PixelFormatYUV420P12LE, []Capability{Capability_AuthToken, Capability_H264, }))
+	// assert.True(checkPixelFormat(ffmpeg.PixelFormatYUV422P12BE, []Capability{Capability_AuthToken, Capability_H264, }))
+	// assert.True(checkPixelFormat(ffmpeg.PixelFormatYUV422P12LE, []Capability{Capability_AuthToken, Capability_H264, }))
+	// assert.True(checkPixelFormat(ffmpeg.PixelFormatYUV444P12BE, []Capability{Capability_AuthToken, Capability_H264, }))
+	// assert.True(checkPixelFormat(ffmpeg.PixelFormatYUV444P12LE, []Capability{Capability_AuthToken, Capability_H264, }))
```
thomshutt: Can we uncomment / get rid of this?

thomshutt (approved):

common/util.go
AlexKordic: fixed

core/capabilities_test.go
AlexKordic: fixed

cyberj0g:
LGTM! Please make sure pinned lpms commit is compatible, and squash+rebase/merge.

cyberj0g (approved):

PR #2280, opened 2022-02-20
core: Add segment clipping support

### LINKED ISSUE #2072: feature: segment slicing wire protocol
To make clipping work well, we're going to need to make frame-accurate cuts of video clips. When cutting from the end of a video clip (the "from point"), any muxer may simply discard video frames. However, when cutting from the front of a video clip (the "to point"), we need to ensure a keyframe gets created at the appropriate time; we must decode the entire video clip but only start encoding from the given timestamp.

My attempt at accounting for all of the necessary changes to implement such a thing:

### HTTP Ingest

* We already have a `Content-Duration` parameter in milliseconds. Proposal: add two additional HTTP headers to allow for in points and out points, like so:

```
# I'd like to cut one second out of the middle of this eight-second clip, so:
Content-Duration: 8000
Content-Slice-From: 4000
Content-Slice-To: 5000
```

Timestamps are in milliseconds and are calculated from the first timestamp present in the video segment.

### B <> O Wire Protocol

**Discovery**: Only Os that have updated to the latest version will support this feature, so we'll need a new capability for discovery, perhaps `Capability_SegmentSlicing`

**Protobuf**: We'll need two new optional fields on `VideoProfile` to make this happen, perhaps `from` and `to`? I've never named a protobuf field before.

### LPMS

This far down, it should be as simple as using the correct frame-accurate ffmpeg parameters to make the cut, something like: `-ss 00:01:00 -to 00:02:00`.

### Open questions that I don't know the answer to

* Does this muck with our pixel-counting logic at all? We're decoding N frames of video but only encoding N-K frames of video. Is this a DDOS vector — sending in huge video segments and asking for one frame in the middle? Personally, as a user, I'd be 100% fine with paying for the full segment either way even if some of the output frames are discareded.

* Ditto, but does this cause problems with fast verification at all? Is it a problem with the sampling?

darkdarkdragon:
Adds ability to clip segments

- Adds Capability_SegmentSlicing capability
- Adds `Content-Slice-From` and `Content-Slice-To` headers to the HTTP push handler
- Adds `From/To` fields to the `SegData` message
- 

Fixes #2072 

**Checklist:**


core/streamdata.go
```diff
@@ -49,6 +49,12 @@ func (s *StreamParameters) StreamID() string {
 	return string(s.ManifestID) + "/" + s.RtmpKey
 }
 
+type SegmentParameters struct {
+	// From clip segment from this timestamp
```
leszko: I think this comment is not clear, could you try to make it more comprehensible or just remove?

server/mediaserver.go
```diff
@@ -820,8 +822,32 @@ func (s *LivepeerServer) HandlePush(w http.ResponseWriter, r *http.Request) {
...
+	if valMs, err := strconv.ParseUint(sliceFromStr, 10, 64); err == nil {
+		sliceFromDur = time.Duration(valMs) * time.Millisecond
+	}
+	var sliceToDur time.Duration
+	if valMs, err := strconv.ParseUint(sliceToStr, 10, 64); err == nil {
+		sliceToDur = time.Duration(valMs) * time.Millisecond
+	}
+	var segPar *core.SegmentParameters
+	if sliceFromDur > 0 || sliceToDur > 0 {
+		if sliceFromDur > 0 && sliceToDur > 0 && sliceFromDur > sliceToDur {
```
leszko, line 785: Shouldn't this condition be just like this?

```
if sliceFromDur > sliceToDur
```

As it is right now, means that if `sliceFromDur = 5` and `sliceToDur = 0`, then it won't result in an error. What is then the meaning of `sliceToDur == 0`?

server/mediaserver.go
```diff
@@ -820,8 +822,32 @@ func (s *LivepeerServer) HandlePush(w http.ResponseWriter, r *http.Request) {
 	remoteAddr := getRemoteAddr(r)
 	ctx = clog.AddVal(ctx, clog.ClientIP, remoteAddr)
 
-	clog.Infof(ctx, "Got push request at url=%s ua=%s addr=%s bytes=%d dur=%s resolution=%s", r.URL.String(), r.UserAgent(), remoteAddr, len(body),
-		r.Header.Get("Content-Duration"), r.Header.Get("Content-Resolution"))
+	sliceFromStr := r.Header.Get("Content-Slice-From")
```
leszko, line 770: Is it maybe possible to add some unit test for this part? Maybe in `push_test,go`?

server/mediaserver.go
darkdarkdragon: If `sliceToDur` is zero that means we don't slice at the end.

server/mediaserver.go
darkdarkdragon: Added

leszko (approved):
LGTM 👍

darkdarkdragon:
blocked by https://github.com/livepeer/lpms/pull/309

cyberj0g:
https://github.com/livepeer/lpms/pull/309 is merged

thomshutt:
@cyberj0g Do you think this is worth making mergeable and keeping, or should we close and revisit clipping in the future?

cyberj0g:
@thomshutt this piece is important, or, at least, it was. I thought it already merged. Not sure if current aim is to make Mist server handle such things as clipping? @yondonfu

victorges:
We actually still intend to merge this, but it became ownerless for a while. I plan on picking this back up now, so will be rebasing and addressing any PR feedback to be able to merge

thomshutt:
Awesome, thanks @victorges!

net/lp_rpc.pb.go
```diff
@@ -1938,6 +1999,7 @@ func init() {
 	proto.RegisterFile("net/lp_rpc.proto", fileDescriptor_034e29c79f9ba827)
 }
 
+// TODO: How do I regenerate this?
```
victorges: Could anyone help me to regenerate this bit? Is this whole file actually auto-generated? How can I trigger it myself?

victorges:
OK rebased and fixed the build (locally) here! Could anyone take another shot on this one?
(also pending re-generating the [net/lp_rpc.pb.go](https://github.com/livepeer/go-livepeer/pull/2280/files/a4ef489b92a1d74dbea3da1a872b62c5857d79fb#diff-50552fac4b8fd4b949f57cf7641160ecfb1b6b8a86d4e1951c89f91f4c9a8807) file, need help!)

net/lp_rpc.pb.go
leszko: I regenerated the file in 1dd38ec8cb3176cfa5368f67a2a4195bf2797504

FYI: to regenerate the file I did execute the following command:
```
protoc --go_out=plugins=grpc:. net/lp_rpc.proto
```

I think it should be executed in `Makefile`, but for some reason it does not work 🤔 I think it's a quick fix, but... we also need to fix the `protoc` version. Currently it only works with the specific version. Please check this [Discord discussion](https://discord.com/channels/423160867534929930/943138984698200137/943521382564769862).

Anyway, at some point, it's worth spending some time to update protoc to the latest version + fix Makefile.

leszko (approved):
@victorges thanks for taking over this PR!

I pushed the commit with the generated `lp_rpc.pb.go`. The rest LGTM. 

Note that I didn't test the change, so I leave it up to you.

net/lp_rpc.pb.go
victorges: Great, thanks!

Yesterday I did actually find that makefile command, tried installing protoc but then it generated some files very different from the current ones, so I didn't know which version to install instead. If protoc is installable via go, I think there's a way to install a "tool" and then invoke a tool in the specific version installed in the project. We could try doing that later.

Regarding makefile, the issue was that the build step is not phony. By default, if a file exists with the same name of a step, it assumes that it has already been made and skips the build step. Since the make step has exactly the name of the go file that is generated it doesn't regenerate it.

To skip that check we can make the build step phony, like some others that we have already in the makefile with something like:
`.PHONY: net/lp_rpc.pb.go`

Then it will run the step even if the file already exists.

leszko: Yes, but TBH I'd do it in a separate PR. And make sure everyone who runs `make` (everyone!) uses the same protoc golang plugin version. Otherwise, it'll be a burden, because you run make and the generated file is suddenly completely different (the same what you experienced).

victorges:
Just finished testing this in staging, no regressions on the record-tester running there continuously, and was just able to clip a couple of segments and videos to my liking! Merging now

PR #1891, opened 2021-05-17
.github/workflows: checkout action use head commit

yondonfu:
**What does this pull request do? Explain your changes. (required)**

Always checkout the head commit for GH actions.

When a `pull_request` event triggers a GH action, the `@actions/checkout` action will checkout from the repo using the value of `github.ref`. The value of `github.ref` in this scenario is the SHA hash of the merge commit that is created by merging the base to the head. The problem with this is that the merge commit hash will be referenced in the go-livepeer version string, but the merge commit might not be present in the commit history (see #1887) which makes it difficult to debug builds.

**Specific updates (required)**

Set `github.ref` to the SHA of the head commit when checking out.

**How did you test each of these updates (required)**

Ran CI.

**Does this pull request close any open issues?**

Fixes #1887 

**Checklist:**

- [x] Read the [contribution guide](./doc/contributing.md)
- [x] `make` runs successfully
- [x] All tests in `./test.sh` pass
- [x] README and other documentation updated
- [ ] ~[Pending changelog](./CHANGELOG_PENDING.md) updated~

PR #2282, opened 2022-02-22
workflow: Setup checksum generation and gpg signing for releases.

### LINKED ISSUE #2147: Add a checksums.txt file or equivalent with binary releases
Currently, the checksum needs to be copy pasted by hand to verify it in a terminal.
Adding a checksum file enables packagers to script the download and verification and integrate it with common distro packaging tools.

Create a checksums.txt file with every release that places the SHA256 checksums in it.

Short of parsing all the HTTP on the page (which breaks every distro's packaging guideline) there are no alternatives.

**Additional context**
Even better, would be for for Livepeer to publish a public PGP key and use that to sign and ship a `.asc` file for each `tar.gz`.

hjpotter92:

- Adds a checksum file alongside release archives
- Creates detached gpg signature for each archive
- Automate release process

Fixes #2147


hjpotter92:
The actual gpg/checksum generation workflow needs it to be merged to master branch first... bootstrap paradox

upload_build.sh
```diff
@@ -33,7 +42,7 @@ fi
 BASE="livepeer-$PLATFORM-$ARCH"
 BRANCH="${TRAVIS_BRANCH:-unknown}"
 if [[ "${GHA_REF:-}" != "" ]]; then
-  BRANCH="$(echo $GHA_REF | sed 's/refs\/heads\///')"
+  BRANCH="$(echo $GHA_REF | sed 's:refs/heads/::')"
```
yondonfu, line 45: Oh is this pattern equivalent to the previous pattern?

install.sh
```diff
@@ -1,11 +1,78 @@
...
-  | grep browser_download_url \
-  | grep "${os}.tar" \
-  | cut -d '"' -f 4 \
-  | xargs -n 1 curl -L \
-  | tar xz \
-  && cp -R "livepeer_${os}/." /usr/local/bin \
-  && rm -rf "livepeer_${os}";
+set -e
+
+PLATFORM="$([ $(uname | grep -i "Darwin") ] && echo 'darwin' || echo 'linux')"
```
yondonfu, line 5: 🥳  for automating GPG signature verification + checksum verification in this script.

If this is going to be the recommended script to run to install the latest version of livepeer we should update the user docs for installing a binary release as well https://docs.livepeer.org/installation/install-livepeer/binary-release - a PR can be opened separately at https://github.com/livepeer/docs. We'll likely want to add instructions for installing the required dependencies as well (i.e. curl, gpg) if they are not installed by default. My thought is that we could suggest using the install script as the first option on the page since it will automatically detect your platform and just leave the other platform specific instructions in case anyone needs them for some reason.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,49 @@
...
+  workflow_run:
+    workflows:
+      - Build binaries
+    types:
+      - "completed"
+
+jobs:
+  release:
+    runs-on: ubuntu-20.04
+    if: ${{ github.event.workflow_run.conclusion == 'success' }}
```
yondonfu, line 13: This event means that all other workflows succeeded (i.e. build, test, etc.)?

yondonfu, line 13: This event means that all other workflows succeeded (i.e. build, test, etc.)?

.github/workflows/release.yaml
```diff
@@ -0,0 +1,49 @@
...
+        id: match-tag
+        with:
+          text: ${{ github.event.workflow_run.head_branch }}
+          regex: "^v([0-9]+\.\d+\.\d+)$"
+
+      - name: Release to github
+        uses: softprops/action-gh-release@v1
+        if: ${{ steps.match-tag.outputs.match != '' }}
+        with:
+          generate_release_notes: true
```
yondonfu, line 46: What will this generate? Will it just generate something based on merged PRs since the last release?

yondonfu, line 46: What will this generate? Will it just generate something based on merged PRs since the last release?

.github/workflows/test.yaml
```diff
@@ -0,0 +1,92 @@
...
+            && sudo apt-get update \
+            && sudo apt-get -y install clang-8 clang-tools-8 build-essential pkg-config autoconf git python
+
+          sudo update-alternatives --install /usr/bin/clang++ clang++ /usr/bin/clang++-8 30 \
+            && sudo update-alternatives --install /usr/bin/clang clang /usr/bin/clang-8 30
+
+      - name: Install go modules
+        run: go mod download
+
+      - name: Install ffmpeg
```
yondonfu, line 67: Previously we always ran this test workflow after the Linux build step so that we could use cached deps for a faster test workflow. Possible to continue doing this instead of repeating the dep installation process here that was already done for the Linux build step?

yondonfu, line 70: Previously we always ran this test workflow after the Linux build step so that we could use cached deps for a faster test workflow. Possible to continue doing this instead of repeating the dep installation process here that was already done for the Linux build step?

yondonfu (commented):
Great PR!

A few questions:

- What benefits are there is moving all platform build steps into a single workflow file? Or is this just an organizational change?
- IIUC these changes will result in a new release being published after each merge to master. That's cool. Maybe we should mark all of those as pre-releases to distinguish them from formal semver releases. WDYT? The pre-releases could then just use pre-generated release notes with a changeset and the formal semver releases could include a little more extra description since those would be for bigger milestones

.github/workflows/release.yaml
hjpotter92: The current behaviour of github actions is that if any of the `workflow_run` workflows defined above (line 4-8) succeed, this action gets triggered. The check here validates that no jobs from all workflows had a failure result.

.github/workflows/release.yaml
hjpotter92: yes, I couldn't find a way to modify the changelog_pending file in place (yet!). the generated release notes are derived in a similar fashion as below:

image

.github/workflows/test.yaml
hjpotter92: I think I can skip the installation step if the cache restore step succeeded.

install.sh
hjpotter92: sure! I'll do that :)

upload_build.sh
hjpotter92: `sed` scripts take first character as the delimiter. using anything other than `/` reduces the escape sequences needed later :)

hjpotter92:
Thanks!

> * What benefits are there is moving all platform build steps into a single workflow file? Or is this just an organizational change?

The release action was getting triggered after _any_ of the dependent workflows completed. This is currently a limitation of github actions.

> * IIUC these changes will result in a new release being published after each merge to master. That's cool. Maybe we should mark all of those as pre-releases to distinguish them from formal semver releases. WDYT? The pre-releases could then just use pre-generated release notes with a changeset and the formal semver releases could include a little more extra description since those would be for bigger milestones

The release workflow runs for merges to master branch, but the release would be created only for semver releases (based on the action check below):

```yaml

      - uses: actions-ecosystem/action-regex-match@v2
        id: match-tag
        with:
          text: ${{ github.event.workflow_run.head_branch }}
          regex: "^v([0-9]+\.\d+\.\d+)$"
```

install.sh
hjpotter92: @yondonfu created the docs PR: livepeer/docs#184 and added the deps installation stage check in test workflow.

.github/workflows/build.yaml
```diff
@@ -16,7 +16,6 @@ jobs:
         with:
           # Needed for commands that depend on git tags
           fetch-depth: 0
-          ref: ${{ github.event.pull_request.head.sha }}
```
yondonfu: Why is this removed here but not for the other platform builds?

hjpotter92: Good catch. It isn't a big issue to keep/remove it, but it is just an unnecessary value,so I've now removed it from all other platforms.

yondonfu: Hm I dug up this old PR https://github.com/livepeer/go-livepeer/pull/1891 and it seems that specifying `github.ref` is actually important because otherwise the checkout action will use the SHA of the merge commit that is created by merging the base to the head, but that merge commit might not be present in commit history (i.e. if a rebase/squash + merge strategy is used). Since the go-livepeer version string uses that SHA, if the merge commit does not exist in the commit history it could make it harder to debug builds.

hjpotter92: thanks for bringing that up. i think I'll need to update this across other projects using sha for ephemeral builds too.

added a comment with the pr link for future reference

yondonfu (approved):
LGTM!

PR #2284, opened 2022-02-22
Fix issue with not redeeming tickets by Redeemer

### LINKED ISSUE #2283: Not possible to redeem tickets with Redeemer on Arbitrum
It's not possible to redeem tickets with a separate Redeemer, because it returns the following error [reported in Discord](https://discord.com/channels/423160867534929930/943004634262413332/945578692023836712):

```
I0221 23:10:46.788973   20060 queue.go:134] Ticket recipient is not active in this round, cannot redeem ticket recipient=0xBe8770603dAf200b1Fa136aD354BA854928e602B
```

**Problem**

1. Redeemer checks if the ticket recipient is an active orchestrator [in the local SQLite](https://github.com/livepeer/go-livepeer/blob/4a8b8bdf08fbe696885b02d1609bdc6c868700d2/pm/queue.go#L172) (in `orchestrator` table).
2. The `orchestrator` table is a cache of the current on-chain orchestrators, which is refreshed:
    - When on-chain block logs are propagated with [orchestrator registration ](https://github.com/livepeer/go-livepeer/blob/754c1ed203539275332363eb2cd2940a7fa6c72a/eth/watchers/orchestratorwatcher.go#L137)
    - (Broadcaster-only) during the [orchestrator discovery](https://github.com/livepeer/go-livepeer/blob/9ab7a8842f2f4c12a0ca9b88cd8ef9b6a1bc1b73/discovery/db_discovery.go#L192)
    - (Orchestrator-only) during the [orchestrator setup](https://github.com/livepeer/go-livepeer/blob/4fb87465094588bfb75ae52a3933cb9a2f2b7940/cmd/livepeer/livepeer.go#L1198)
3. Redeemer may not have the correct `orchestrator` cache in the following cases
    - It was first started after orchestrators were already registered (we backfill blocks only until last block found in SQLite DB)
    - Logs has crashed (probably because of this bug: https://github.com/livepeer/go-livepeer/issues/2265

**Solution**

The solution is to either just allow redeeming tickets if the given orchestrator is not found in the local SQLite DB or refresh `orchestrator` if the given orchestrator is not found.

leszko:
Fix issue when Redeemer can't redeem a ticket.

- Add `ethOrchAddr` to DB `orchestrator` while starting Livepeer Node in the Redeemer mode (before it was added only when started in the Orchestrator mode)
- Ease the condition for redeeming the ticket Before it was that the orchestrator exists in SQLite DB and had the activation round before the current round. Now, we also redeem a ticket when no orchestrator is found.

unit tests

fix #2283

**Checklist:**

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -21,5 +21,6 @@
 #### Broadcaster
 
 #### Orchestrator
+- \#2284 Fix issue with not redeeming tickets by Redeemer (@leszko)
 
 #### Transcoder
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -635,10 +635,7 @@ func main() {
 				return
 			}
 
-			orchSetupCtx, cancel := context.WithCancel(ctx)
-			defer cancel()
-
-			if err := setupOrchestrator(orchSetupCtx, n, recipientAddr); err != nil {
+			if err := setupOrchestrator(n, recipientAddr); err != nil {
 				glog.Errorf("Error setting up orchestrator: %v", err)
 				return
 			}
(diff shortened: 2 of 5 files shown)


yondonfu (approved):
LGTM

PR #2289, opened 2022-02-24
Add timeouts to eth client

### LINKED ISSUE #2207: Improve Feed PubSub mechanism

leszko:

Add timeouts for `CheckTx()` function and all ETH RPC calls.

Note that the timeout values are quite conservative, by default:
- 5 min for `CheckTx()`
- 10 min for every ETH RPC call

I don't think we need smaller numbers, because in the correct flow of action, these timeouts should never actually be reached.

- Add timeout for the `CheckTx()` execution
- Add timeout for all ETH RPC (contracts) calls
- Refactor client to not include contract sessions anymore

- Checked that using eth client and sending transactions still work fine
- Change `CheckTxTimeout` to 1 ns and check that it always times out
- Change `ethRpcTimeout` to 1 ns and check that all ETH RPC requests fail

fix #2207

**Checklist:**

### DIFF
--- a/cmd/devtool/devtool.go
+++ b/cmd/devtool/devtool.go
@@ -31,7 +31,7 @@ const (
 )
 
 var (
-	ethTxTimeout              = 600 * time.Second
+	ethTxTimeout              = 5 * time.Minute
 	endpoint                  = "http://localhost:8545/"
 	gethMiningAccount         = "87da6a8c6e9eff15d703fc2773e32f6af8dbe301"
 	gethMiningAccountOverride = false
@@ -188,7 +188,8 @@ func ethSetup(ethAcctAddr, keystoreDir string, isBroadcaster bool) {
 		return
 	}
 
-	tm := eth.NewTransactionManager(backend, gpm, am, 5*time.Minute, 0)
+	maxTxReplacements := 0
+	tm := eth.NewTransactionManager(backend, gpm, am, ethTxTimeout, maxTxReplacements)
 	go tm.Start()
 	defer tm.Stop()
 
@@ -199,6 +200,7 @@ func ethSetup(ethAcctAddr, keystoreDir string, isBroadcaster bool) {
 		GasPriceMonitor:    gpm,
 		TransactionManager: tm,
 		Signer:             types.LatestSignerForChainID(chainID),
+		CheckTxTimeout:     time.Duration(int64(ethTxTimeout) * int64(maxTxReplacements+1)),
 	}
 
 	client, err := eth.NewClient(ethCfg)
--- a/eth/client.go
+++ b/eth/client.go
@@ -39,7 +39,7 @@ var (
 	ErrReplacingMinedTx   = fmt.Errorf("trying to replace already mined tx")
 	ErrCurrentRoundLocked = fmt.Errorf("current round locked")
 	ErrMissingBackend     = fmt.Errorf("missing Ethereum client backend")
-	ethRpcTimeout         = 5 * time.Minute
+	ethRpcTimeout         = 2 * time.Minute
 )
 
 type LivepeerEthClient interface {
--- a/eth/client.go
+++ b/eth/client.go
@@ -39,6 +39,7 @@ var (
 	ErrReplacingMinedTx   = fmt.Errorf("trying to replace already mined tx")
 	ErrCurrentRoundLocked = fmt.Errorf("current round locked")
 	ErrMissingBackend     = fmt.Errorf("missing Ethereum client backend")
+	ethRpcTimeout         = 5 * time.Minute
 )
 
 type LivepeerEthClient interface {
@@ -395,7 +396,20 @@ func (c *client) transactOpts() *bind.TransactOpts {
 	opts := c.transOpts
 	c.transOptsMu.RUnlock()
 
+	opts.Context = newEthRpcContext()
+
 	return &opts
+}
+
+func (c *client) callOpts() *bind.CallOpts {
+	return &bind.CallOpts{
+		Context: newEthRpcContext(),
+	}
+}
+
+func newEthRpcContext() context.Context {
+	ctx, _ := context.WithTimeout(context.Background(), ethRpcTimeout)
+	return ctx
 }
 
 func (c *client) Account() accounts.Account {
@@ -408,16 +422,16 @@ func (c *client) Backend() Backend {
 
 // Controller
 func (c *client) GetContract(hash ethcommon.Hash) (ethcommon.Address, error) {
-	return c.controllerSess.GetContract(hash)
+	return c.controllerSess.Contract.GetContract(c.callOpts(), hash)
 }
 
 func (c *client) Paused() (bool, error) {
-	return c.controllerSess.Paused()
+	return c.controllerSess.Contract.Paused(c.callOpts())
 }
 
 // Rounds
 func (c *client) InitializeRound() (*types.Transaction, error) {
-	i, err := c.roundsManagerSess.CurrentRoundInitialized()
+	i, err := c.roundsManagerSess.Contract.CurrentRoundInitialized(c.callOpts())
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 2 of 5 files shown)


eth/client.go
```diff
@@ -139,23 +140,23 @@ type client struct {
...
-	controllerSess          *contracts.ControllerSession
-	livepeerTokenSess       *contracts.LivepeerTokenSession
-	serviceRegistrySess     *contracts.ServiceRegistrySession
-	bondingManagerSess      *contracts.BondingManagerSession
-	ticketBrokerSess        *contracts.TicketBrokerSession
-	roundsManagerSess       *contracts.RoundsManagerSession
-	minterSess              *contracts.MinterSession
-	livepeerTokenFaucetSess *contracts.LivepeerTokenFaucetSession
+	// Contracts
+	controller          *contracts.Controller
```
victorges, line 144: Like the refactor, but is this related to the timeouts change? Seems to be the bulk of the PR here.

eth/client.go
```diff
@@ -1101,6 +1089,8 @@ func (c *client) CheckTx(tx *types.Transaction) error {
 
 	for {
 		select {
+		case <-time.After(c.checkTxTimeout):
```
victorges: I believe you actually need to create the `time.After` timer only once, outside the `for` loop, otherwise you will reset the timer for every receipt that you get even if not related to the tx you're checking.

leszko: Updated in a354f7148d50b275912ac379f405d2f7dd9124ca

eth/client.go
leszko: It is related because before adding the timeout, we actually needed to store sessions here (`CallOpts` was encapsulated in the session structs). Now, we don't need sessions, because we always provide our `CallOpts`.

victorges: Cool! Thanks for clarification

victorges (approved):
LGTM!

cmd/devtool/devtool.go
```diff
@@ -199,6 +199,7 @@ func ethSetup(ethAcctAddr, keystoreDir string, isBroadcaster bool) {
 		GasPriceMonitor:    gpm,
 		TransactionManager: tm,
 		Signer:             types.LatestSignerForChainID(chainID),
+		CheckTxTimeout:     ethTxTimeout,
```
yondonfu: WDYT about setting this in the same way as in `livepeer.go` for consistency? i.e. multiply the timeout by the max replacements?

eth/client.go
```diff
@@ -39,6 +39,7 @@ var (
 	ErrReplacingMinedTx   = fmt.Errorf("trying to replace already mined tx")
 	ErrCurrentRoundLocked = fmt.Errorf("current round locked")
 	ErrMissingBackend     = fmt.Errorf("missing Ethereum client backend")
+	ethRpcTimeout         = 10 * time.Minute
```
yondonfu: I know this was the original timeout value, but seems like this should be lower than 10 minutes for receiving a response from a RPC call...maybe 2 minutes?

eth/client_ticketbroker.go
```diff
@@ -26,7 +26,7 @@ func (c *client) FundDeposit(amount *big.Int) (*types.Transaction, error) {
 	opts := c.transactOpts()
 	opts.Value = amount
```
yondonfu, line 27: This was already here before this PR, but since `transactOpts` is stored on `client` and `c.transactOpts()` returns a pointer to that field does the update to the `Value` field in the next line get reflected in that field such that the next call to `c.transactOpts()` would return a pointer to a `bind.TransactOpts` with the `Value` field set (which we wouldn't want)?

cmd/devtool/devtool.go
leszko: Yes, updated in ee6a8c16ee301b9d3eb1d6decd4f5280a9dca799.

You also made me realize that in `livepeer.go` it should actually be `txTimeout * (maxTxReplacements + 1)`, not `txTimeout * maxTxReplacements`. Updated as well.

eth/client.go
leszko: Yeah, 2 min sounds good to me. Updated in cdfb8d15dc15300483eaed64a2e8f4d9bf7b796e

eth/client_ticketbroker.go
leszko: Actually, that's not the case, because `c.transactOpts()` returns a copy of `TransactOpts`. The copying happens [here](https://github.com/leszko/go-livepeer/blob/cdfb8d15dc15300483eaed64a2e8f4d9bf7b796e/eth/client.go#L368).

leszko (commented):
@yondonfu Addressed your comments. PTAL

eth/client_ticketbroker.go
yondonfu: Aha I see.

The field `c.transactOpts` is a value so `opts := c.transactOpts` will copy the value and assign it to `opts`. Then, `c.transactOpts()` will return a pointer to the *copy* of `c.transactOpts` and not to `c.transactOpts`. As a result, any update to the pointer returned by `c.transactOpts()` will only impact the copy of `c.transactOpts` and not `c.transactOpts`. As a result, the update to the `Value` field in the referenced line will only affect a copy of `c.transactOpts` and since we call `c.transactOpts()` to return a pointer to a fresh copy of `c.transactOpts` for each method we don't need to worry about that update affecting other method calls.

yondonfu (approved):
LGTM after squashing!

eth/client_ticketbroker.go
leszko: yes, exactly. Also, the reason for copying `TransactOpts` makes the code thread-safe.

PR #2290, opened 2022-02-24
allow orchestrator to set ticket maxfacevalue

ad-astra-video:
Allows Orchestrator nodes to specify ticket max face value while node is running.  Default is no max face value.

- Add new fields to pm/recipient.go for maxfacevalue to adjust TicketParams faceValue.  Update also included adding new methods to Recipient interface to provide means to update the values.
- Add launch flag to specify value at startup
- Add functions to core/livepeernode.go to update orchestrator node settings for above pm/recipient.go updates.
- Add webserver endpoint /setMaxFaceValue to enable updating new maxfacevalue setting.

After updates completed, ran orchestrator node on mainnet and monitored logs after updating settings with webserver endpoints. 

Ticket was redeemed for 0.01 ETH by remote redeemer node for orchestrator. https://discord.com/channels/423160867534929930/808142181679235083/945862545527439370.

How to use webserver endpoints:
```
curl -F maxfacevalue=[face value in wei] http://127.0.0.1:7935/setMaxFaceValue
```

No

**Checklist:**

- [x] `make` runs successfully - with these changes applied to v0.5.29 source code


leszko (commented):
@0xB79 Thanks a lot for the PR. Great to see the community contribution!

Before we dig into the code review, could you please explain the use cases for the changes you made?
1. I understand that one of the changes is to make `-ticketEV` flag configurable at runtime. In which scenarios would the Broadcaster need it?
2. You added two configuration parameters `faceValueLimit` and `faceValueFixed`. When would you configure each of them? Maybe we will need to also at them to the flags, but first I need to understand the use cases.

Thanks!

ad-astra-video:
Hello @leszko

These updates are to allow the Orchestrator more control when setting the TicketParams sent to the Broadcasters.

1. Updating `ticketEV` while Orchestrator node is running just updates the frequency the orchestrator gets tickets.  With each update to `ticketEV` the `winProb` changes proportionally to the `faceValue`.  To note, updating `ticketEV` to amount higher than current value while node is running may drop streams because the Orchestrator node, I believe, determines that the Broadcaster has not sent enough payment for transcoding.  This happened when running the tests below on an Orchestrator node running on mainnet but streams picked back up within a few minutes.
2. These allow Orchestrator ability to override the default `faceValue` calculated in `pm/recipient.go`.  `faceValueLimit` would be to set an upper limit on the `faceValue` calculated and `fixedFaceValue` would be to override to any `faceValue`.

These could be added as command line flags and be set at run time.  Another addition that would be helpful is adding ability to set these options with `livepeer_cli`.

Some examples of tickets received by orchestrator after updating running Orchestrator node:
Default TicketParams
```Feb 25 11:37:49 s1155 livepeer[2190226]: I0225 11:37:49.080371 2190226 orchestrator.go:170] manifestID=7e1ef526-de34-4a00-8cb9-ea4e1a00f346 seqNo=6361 orchSessionID=6e38d210 clientIP=84.17.50.99 sender=0xc3c7c4C8f7061B7d6A72766Eee5359fE4F36e61E Receiving ticket sessionID=6e38d210 faceValue=0.09300341136 ETH winProb=0.0000107523 ev=1000000000000.00```

Set faceValueLimit to 0.02ETH: `curl -F facevaluelimit=20000000000000000 http://127.0.0.1:7935/setFaceValueLimit`
```Feb 25 11:41:17 s1155 livepeer[2190226]: I0225 11:41:17.673767 2190226 orchestrator.go:170] manifestID=747fa4e9-53e6-415d-9311-9a850433071b seqNo=14709 orchSessionID=3146c688 clientIP=89.187.178.130 sender=0xc3c7c4C8f7061B7d6A72766Eee5359fE4F36e61E Receiving ticket sessionID=3146c688 faceValue=0.02 ETH winProb=0.0000500000 ev=1000000000000.00```

Set fixedFaceValue to 0.11 ETH, just above calculated faceValue: `curl -F fixedfacevalue=110000000000000000 http://127.0.0.1:7935/setFixedFaceValue`
```Feb 25 11:43:39 s1155 livepeer[2190226]: I0225 11:43:39.443904 2190226 orchestrator.go:170] manifestID=747fa4e9-53e6-415d-9311-9a850433071b seqNo=14780 orchSessionID=3146c688 clientIP=89.187.178.130 sender=0xc3c7c4C8f7061B7d6A72766Eee5359fE4F36e61E Receiving ticket sessionID=3146c688 faceValue=0.11 ETH winProb=0.0000090909 ev=1000000000000.00```

Set fixedFaceValue to 0.50 ETH: `curl -F fixedfacevalue=500000000000000000 http://127.0.0.1:7935/setFixedFaceValue`
```Feb 25 12:07:08 s1155 livepeer[2190226]: I0225 12:07:08.629615 2190226 orchestrator.go:170] manifestID=5db2e08f-6e4c-4740-abbc-4a6f421feb46 seqNo=413 orchSessionID=e3a42ac7 clientIP=195.181.174.39 sender=0xc3c7c4C8f7061B7d6A72766Eee5359fE4F36e61E Receiving ticket sessionID=e3a42ac7 faceValue=0.5 ETH winProb=0.0000020000 ev=1000000000000.00```

Set ticketEV to double default ticketEV to get less tickets with higher winProb: `curl -F ticketev=2000000000000 http://127.0.0.1:7935/setTicketEV`
```Feb 25 11:49:25 s1155 livepeer[2190226]: I0225 11:49:25.614930 2190226 orchestrator.go:170] manifestID=d8b0370ycw1jkeop seqNo=26 orchSessionID=d4cfc904 clientIP=143.244.61.193 sender=0xc3c7c4C8f7061B7d6A72766Eee5359fE4F36e61E Receiving ticket sessionID=d4cfc904 faceValue=0.11 ETH winProb=0.0000181818 ev=2000000000000.00```

**Comment by leszko** [comment]
@0xB79 Thanks for the explanation!

I add @yondonfu and @hthillman to decide if we want to add these `faceValueLimit` and `fixedFaceValue` parameters. I suggest that if we add the endpoints, we should also add `livepeer` flags to be consistent here.

Concerning `ticketEV`, I think it's ok to change it at runtime, so adding this endpoints is a good idea in my opinion. 

I'll add some comments wrt the code itself after the comments from @yondonfu and/or @hthillman .

**Comment by yondonfu** [comment]
I think giving Os more control over the faceValue used for tickets makes sense. However, I'm in favor of a configurable flag to adjust the maximum faceValue to use for a ticket, but not a configurable flag to set a fixed faceValue to use for a ticket right now because the security of the PM protocol right now depends on an O using a faceValue that does *not* exceed the reserve allocation that a B commits to the O (i.e. the B's reserve divided by the # of active Os). If an O uses a faceValue that exceeds this reserve allocation then the difference between the faceValue and the reserve allocation is the value that O risks losing out on in the even that a B's deposit is depleted before the O redeems a winning ticket. So, I'd advocate for the node to behave as safely as possible as a default to avoid scenarios where an O sets a fixed faceValue that causes it to lose out on value if a B's deposit is depleted.

If O can set the maximum faceValue to use then in the scenario where the B's reserve allocation is less than the maximum faceValue the node will still automatically set the faceValue to the reserve allocation as the "safe" value to use. In the scenario where the B's reserve allocation is greater than the maximum faceValue the node will just use the maximum faceValue as the faceValue. I believe this is more or less the behavior of the `-faceValueLimit` flag included in the current PR. In order to be consistent with naming, since we already have a `-maxTicketEV` flag, I suggest using the name `-maxFaceValue`.

**Comment by ad-astra-video** [comment]
I have updated to add command line flag and remove fixedFaceValue option.  Tested by adding these commits to v0.5.29 code and building. 

Also, added livepeer_cli option to update while orchestrator node is running.

File: cmd/livepeer/livepeer.go [hunk 1]
```diff
@@ -633,10 +634,19 @@ func main() {
...
 			if err != nil {
 				glog.Errorf("Error setting up PM recipient: %v", err)
 				return
 			}
+			mfv, _ := new(big.Int).SetString(*maxFaceValue, 10)
+			if mfv == nil {
+				glog.Errorf("-maxFaceValue must be a valid integer, but %v provided. Restart the node with a different valid value for -maxFaceValue", *maxFaceValue)
+				return
+			} else {
+				n.SetFaceValueLimit(mfv)
```
leszko: For the sake of consistency, maybe let's name it `n.SetMaxFaceValue()` instead of `n.SetFaceValueLimit()`?

cmd/livepeer/livepeer.go
```diff
@@ -633,10 +634,19 @@ func main() {
 				timeWatcher,
 				cfg,
 			)
+			n.SetTicketEV(ev)
```
leszko: Isn't that line setting the fixed face value? I think we wanted to allow only setting max face value, why do we need this line then?

leszko (changes requested):
Thanks for the changes @0xB79 !

Added two comments. In short, the change about setting `maxFaceValue` is fine, however I'm not sure about the changes related to setting EV. I thought we only want to add the ability to set max face value in this PR.

Also, please rebase you branch to `master` since there are some conflicts.

ad-astra-video:
Hello,

I applied the changes to dev branch using master branch code base.  I am not sure I did this correctly...if I did not, commit [`https://github.com/0xB79/go-livepeer/commit/280f67ff398b62e935f5a01e689df2056222cea1`](this commit) should be able to applied to master branch. Would you prefer I do a pull request for that commit on master branch?

I incorporated the changes requested where it only allows setting a max face value.  I tested by building and running on working orchestrator node.

There was one bug fix in starter.go where the `cfg` variable was changed to a TicketParamsConfig variable.  It seemed to me this was not intentional so I updated the variable name to `tcfg`.  I needed the `cfg` variable that had `MaxFaceValue` set from flag/config file after the lines below to set the value in the pm recipient.

```
@@ -705,7 +708,7 @@ func StartLivepeer(ctx context.Context, cfg LivepeerConfig) {
			sm.Start()
			defer sm.Stop()

			cfg := pm.TicketParamsConfig{
			tcfg := pm.TicketParamsConfig{
				EV:               ev,
				RedeemGas:        redeemGas,
				TxCostMultiplier: txCostMultiplier,
@@ -717,12 +720,20 @@ func StartLivepeer(ctx context.Context, cfg LivepeerConfig) {
				gpm,
				sm,
				timeWatcher,
				cfg,
				tcfg,
			)
			if err != nil {
				glog.Errorf("Error setting up PM recipient: %v", err)
				return
			}
```

leszko (changes requested):
Thanks for the update @0xB79 

I think there is something wrong with the PR:
- it's sent against `livepeer:dev`, not `livepeer:master`
- It shows way too many changes in GH (140 files changed).

Could you please clean it up?

ad-astra-video:
@leszko I changed it to the livepeer:master branch and it shows 9 files changed now.

cmd/livepeer/starter/starter.go
```diff
@@ -717,12 +720,20 @@ func StartLivepeer(ctx context.Context, cfg LivepeerConfig) {
...
+				tcfg,
 			)
 			if err != nil {
 				glog.Errorf("Error setting up PM recipient: %v", err)
 				return
 			}
+			mfv, _ := new(big.Int).SetString(*cfg.MaxFaceValue, 10)
+			if mfv == nil {
+				glog.Errorf("-maxFaceValue must be a valid integer, but %v provided. Restart the node with a different valid value for -maxFaceValue", *cfg.MaxFaceValue)
+				return
```
```suggestion
				panic(fmt.Errorf("-maxFaceValue must be a valid integer, but %v provided. Restart the node with a different valid value for -maxFaceValue", *cfg.MaxFaceValue))
```

Please use `panic()` instead.

pm/recipient.go
```diff
@@ -269,7 +278,13 @@ func (r *recipient) faceValue(sender ethcommon.Address) (*big.Int, error) {
 		// Set faceValue = maxFloat
 		faceValue = maxFloat
 	}
-
+	
+	if r.maxfacevalue.Cmp(big.NewInt(0)) > 0 {
+		//if faceValue > maxfacevalue, set faceValue=maxfacevalue
+		if r.maxfacevalue.Cmp(faceValue) < 0 {
```
leszko:
```suggestion
		if r.maxfacevalue.Cmp(faceValue) < 0 {
```

Nit: There is no need for this comment.

server/handlers.go
```diff
@@ -517,6 +517,25 @@ func (s *LivepeerServer) setServiceURI(client eth.LivepeerEthClient, serviceURI
 	return nil
 }
 
+func (s *LivepeerServer) setMaxFaceValueHandler() http.Handler {
```
leszko, line 520: Would be nice if you could add a unit test in `handlers_test.go`.

leszko (changes requested):
Added some minor comments. Other than that, please rebase to master. Thanks!

pm/recipient.go
ad-astra-video: Agree, not used to Go with using .Cmp.  Will remove comment.

server/handlers.go
ad-astra-video: Unit tests added for maxfacevalue handler.  Took some time for me to get up to speed on how to make a mock recipient in unit tests for the stub livepeer server.

cmd/livepeer/starter/starter.go
ad-astra-video: Updated

server/handlers.go
ad-astra-video: Unit tests added.

leszko (approved):
Thanks for all the changes @0xB79 LGTM 👍

cmd/livepeer/starter/starter.go
```diff
@@ -717,12 +720,20 @@ func StartLivepeer(ctx context.Context, cfg LivepeerConfig) {
...
 				sm,
 				timeWatcher,
-				cfg,
+				tcfg,
 			)
 			if err != nil {
 				glog.Errorf("Error setting up PM recipient: %v", err)
 				return
 			}
+			mfv, _ := new(big.Int).SetString(*cfg.MaxFaceValue, 10)
```
thomshutt, line 729: Should we also check the returned bool that indicates success here?

thomshutt (approved):
LGTM - Could you update the PR description to reflect what the changes now look like please and then we'll merge

leszko:
@0xB79 Last comments from Thom and we can merge your PR 🥳 

Also, please run `go fmt`, because the GH Action test failed on the formatting.

cmd/livepeer/starter/starter.go
ad-astra-video: My understanding was `SetString` returns `nil` if it fails.  Is this correct?  This is same pattern used on lines 675 and 740. I can update if prefer to check bool rather than value or prefer to check both.

ad-astra-video:
Updated title and pull request description to remove the items dropped from review comments.

cmd/livepeer/starter/starter.go
thomshutt: From looking at the `SetString` code, it seems like the way you've done it probably is safe:

> If SetString fails, the value of z is undefined but the returned value is nil

but I think to be safe and not depend on that behavior never changing, let's also check the `success` bool isn't false

leszko: I think that for this one, for the sake of consistency, we can keep it as it is. And later convert all calls `mfv, _ := new(big.Int).SetString(*cfg.MaxFaceValue, 10)` to what Thom suggested.

leszko:
Thanks again for the PR @0xB79

Issue #315, opened 2018-03-01
Use filter with a start block from the end of the previous subscription when resubscribing


yondonfu:
There is an an edge case with subscription retries where a transcoder's new job subscription could end due to a closed websocket connection with Infura, a new job is created that is assigned to the transcoder and when the transcoder resubscribes to new jobs it will have missed the new job that was just created. 

I think we should be able to add some logic to track the block at which a subscription ends and then update the filter used for the subsequent resubscribe so that it will filter for events starting from the block at which the previous subscription ended so that the transcoder would still receive the new job event that was fired during the brief period of time that it wasn't subscribed. Similar logic could be used for other events as well.

Additionally, if subscriptions keep failing we should probably just stop resubscribing after a number of retries in the case where there is an issue with Infura such that a websocket connection just cannot be opened successfully. We could inform the user of such a case

ericxtang:
From @yondonfu - 

> Perhaps an extension to this proposal that is related to #315 is that a transcoder keeps track of the last block that it was assigned a job and we introduce an option in the CLI, something like "Sync Jobs". When that option is triggered, the transcoder can filter for job events from the last block that it was assigned a job until the current block and check if it was assigned to any of them and start doing work for the job.

j0sh:
> if subscriptions keep failing we should probably just stop resubscribing after a number of retries

For long-lived nodes such as a transcoder, it is probably better (and harmless) to keep re-trying for as long as needed, maybe with some backoff up to a few minutes. We certainly should still inform the user though.

dob:
Related, but perhaps not exactly the same, is "subscribing" to the actual streams on the network. When a transcoder restarts does it need to resubscribe to any stream that has not been explicitly ended for which it was assigned? This is also realted to the concept of long running reserved jobs, where a broadcaster gets assigned a transcoder, and may start/stop streaming intermittently for hours or days until the endBlock of the job.

Just because it sends the subscribe request back towards the broadcaster doesn't mean that it necessarily will continue to get segments, if the broadcaster is in fact, not broadcasting. But it won't continue to get any segments it should get if it doesn't resubscribe. (I also may be off on the updated architecture)

j0sh:
Posted some related thoughts on re-subscription here  https://github.com/livepeer/go-livepeer-basicnet/issues/21#issuecomment-372748851 , related to the work for https://github.com/livepeer/go-livepeer-basicnet/issues/29

yondonfu:
[...]

PR #2291, opened 2022-02-25
Calling video comparison to improve the security strength

oscar-davids:
**Calling video comparison functions on results received from different Os**

-  Add downloading function [here](https://github.com/livepeer/go-livepeer/blob/cddc3fffe93fe01e1a361bf7a7e0dea5a96dc00a/server/broadcast.go#L488) to download segmToCheckIndex-th trusted segment.
- Also [here ](https://github.com/livepeer/go-livepeer/blob/cddc3fffe93fe01e1a361bf7a7e0dea5a96dc00a/server/broadcast.go#L500)call GetSegmentData function to download untrusted segment.
- Call [CompareVideoByBuffer ](https://github.com/livepeer/lpms/blob/44fe3cdd279054f6ff399b223d01ed19f10c9e17/ffmpeg/ffmpeg.go#L233)to compare two segments before calling [CompareSignatureByBuffer](https://github.com/livepeer/go-livepeer/blob/cddc3fffe93fe01e1a361bf7a7e0dea5a96dc00a/server/broadcast.go#L506).

- Call video comparison function in `chooseResults()`.
- Passed Unit Test.

Fix [#315](https://github.com/livepeer/internal-project-tracking/issues/315)

**Checklist:**

### DIFF
--- a/server/push_test.go
+++ b/server/push_test.go
@@ -1533,7 +1533,7 @@ func TestPush_ReuseIntmidWithDiffExtmid(t *testing.T) {
 
 func TestPush_MultipartReturnMultiSession(t *testing.T) {
 	assert := assert.New(t)
-	//need real video data for fast verification
+	// need real video data for fast verification
 	transcodeddata, err := ioutil.ReadFile("../core/test.ts")
 	assert.NoError(err)
 	goodHash, err := ioutil.ReadFile("../core/test.phash")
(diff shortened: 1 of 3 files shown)


server/push_test.go
```diff
@@ -1533,7 +1533,9 @@ func TestPush_ReuseIntmidWithDiffExtmid(t *testing.T) {
 
 func TestPush_MultipartReturnMultiSession(t *testing.T) {
 	assert := assert.New(t)
-
+	//need real video data for fast verification
```
leszko:
```suggestion
	// need real video data for fast verification
```

server/push_test.go
```diff
@@ -1533,7 +1533,9 @@ func TestPush_ReuseIntmidWithDiffExtmid(t *testing.T) {
...
+	transcodeddata, err := ioutil.ReadFile("../core/test.ts")
```
leszko, line 1537: Does it test the fast verification? If yes, then which check `pHash` or `CompareVideoByBuffer`? I think we should have somewhere tests for both.

leszko (commented):
Looks, good! Good work @oscar-davids 

Added 2 minor comments. Other than that, here are some comments things to consider:
1. Traffic-wise, we add downloading the whole segment. I guess it's ok, but just want to double-check it.
2. One optimization would be to check first `pHash` and only if `pHash` matches, then download the segment. But not sure we need it because we actually expect `pHash` to match in most cases.
3. One other optimization could be that we actually download the segment from untrusted O twice. One for the fast verification and later to download the result [here](https://github.com/livepeer/go-livepeer/blob/985d0ec65659c0586cf1abdce1ac5e639e6d9b8f/server/broadcast.go#L1112). Again, maybe we don't need to change it right now.
4. I didn't check if `ffmpeg.CompareVideoByBuffer()` works fine and the implication on having some false-negative results.

server/push_test.go
oscar-davids: Will add unit test  for `CompareVideoByBuffer`  separately.

oscar-davids:
> 1. Traffic-wise, we add downloading the whole segment. I guess it's ok, but just want to double-check it.

Yeah, there is additional traffic for downloading the random n-th segment.

> 2. One optimization would be to check first `pHash` and only if `pHash` matches, then download the segment. But not sure we need it because we actually expect `pHash` to match in most cases.
> 3. One other optimization could be that we actually download the segment from untrusted O twice. One for the fast verification and later to download the result here. Again, maybe we don't need to change it right now.

I prefer the second optimization path because I think it is a little late to check validity at that time [here](https://github.com/livepeer/go-livepeer/blob/985d0ec65659c0586cf1abdce1ac5e639e6d9b8f/server/broadcast.go#L1112) with the third way.

> 4. I didn't check if `ffmpeg.CompareVideoByBuffer()` works fine and the implication on having some false-negative results.

For H264 encoding, it showed 100% accuracy in GPU vs CPU cross-comparison. I think we need to re-evaluate the H265 codec.

AlexKordic (approved):
Well Done!

leszko (approved):
LGTM 👍 

Thanks for adding the unit test!

PR #2299, opened 2022-03-07
Split devtool Orchestrator scripts with/without external transcoder. Prevent run scripts from using same CLI port.

thomshutt:
It's currently not possible to run the `run_broadcaster` and `run_transcoder` scripts generated by devtool simultaneously, because `run_broadcaster` specifies a CLI Port of `7935` and run_transcoder doesn't specify one (and so go-livepeer uses `7935` by default).

* Specify <orchestrator port + 1> as the CLI port for the transcoder script
* Split out Orchestrator script into two separate ones (with/without transcoder running)
* Update docs to reflect script changes

* Generated the scripts and streamed with both B/O and B/O/T setup

No

**Checklist:**

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -17,7 +17,7 @@
 ### Bug Fixes 🐞
 
 #### General
-- \#2299 Prevent Transcoder/Broadcaster run scripts from using same CLI port
+- \#2299 Split devtool Orchestrator run scripts into versions with/without external transcoder and prevent Transcoder/Broadcaster run scripts from using same CLI port (@thomshutt)
 
 #### Broadcaster
 - \#2296 Increase orchestrator discovery timeout from `500ms` to `1` (@leszko)
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -17,6 +17,7 @@
 ### Bug Fixes 🐞
 
 #### General
+- \#2299 Prevent Transcoder/Broadcaster run scripts from using same CLI port
 
 #### Broadcaster
 - \#2296 Increase orchestrator discovery timeout from `500ms` to `1` (@leszko)
--- a/cmd/devtool/README.md
+++ b/cmd/devtool/README.md
@@ -31,5 +31,7 @@ This command will submit the setup transactions for a broadcaster and generate t
 
 `go run cmd/devtool/devtool.go setup transcoder`
 
-This command will submit the setup transactions for an orchestrator/transcoder and generate the Bash scripts 
-`run_orchestrator_<ETH_ACCOUNT>.sh` which can be used to start an orchestrator node and `run_transcoder_<ETH_ACCOUNT>.sh` which can be used to start a transcoder node.
+This command will submit the setup transactions for an orchestrator/transcoder and generate the Bash scripts:
+
+* `run_orchestrator_with_transcoder_<ETH_ACCOUNT>.sh` which can be used to start an orchestrator node that contains a transcoder (combined OT)
+* `run_orchestrator_standalone_<ETH_ACCOUNT>.sh`  and `run_transcoder_<ETH_ACCOUNT>.sh` which can be used to start separate orchestrator and transcoder nodes (split O/T)
--- a/cmd/devtool/devtool.go
+++ b/cmd/devtool/devtool.go
@@ -102,7 +102,7 @@ func main() {
 	}
 	serviceURI += strconv.Itoa(mediaPort + nodeIndex)
 	mediaPort += nodeIndex
-	cliPort += nodeIndex
+	cliPort += nodeIndex * 2 // Because we need different CLI ports for the Orchestrator and Transcoder scripts
 	rtmpPort += nodeIndex
 
 	t := getNodeType(isBroadcaster)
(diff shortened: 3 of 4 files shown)


leszko (commented):
@thomshutt Thanks for the PR.

I wonder if this change actually solves the issue of executing `run_transcoder.sh`. Here are my thoughts:
1. The O/T topology (when the orchestrator is run together with transcoder in one process) works anyway, just execute:
```
./run_broadcaster*.sh
./run_orchestrator*.sh
```
There is no port conflict, everything works fine.

2. The split OT topology (which your PR is supposed to fix) didn't work for me after your change. Maybe I did something wrong...

I think to make the split OT topology it'd require some more changes. Even if we fix the CLI port, the following execution does not work.

```
./run_broadcaster*.sh
./run_orchestrator*.sh
./run_transcoder*.sh
```

The reason for this is that the Orchestrator is anyway run with `-transcoder` parameter, so it does not accept other transcoders.

My suggestion is to:
- either fix scripts for the split O/T topology
- either just accept the split O/T does not work and maybe create a GH issue to fix it

thomshutt:
@leszko Thanks for reviewing! 

That makes sense, I'll remove the `-transcoder` parameter, since all of the documentation (and in particular https://github.com/livepeer/go-livepeer/blob/master/cmd/devtool/README.md) talks about B/O/T as separate entities and having a Transcoder automatically started by the Orchestrator confused me when I was getting this set up.

leszko:
TBH I'd keep the possibility for the O/T topology and maybe add an option to generate the split O/T (separate transcoder)? I think O/T is good, because it's the simplest way to test something on-chain. So I use it very often.

thomshutt:
👍 Fair enough, I'll go with that then

CHANGELOG_PENDING.md
```diff
@@ -17,6 +17,7 @@
 ### Bug Fixes 🐞
 
 #### General
+- \#2299 Prevent Transcoder/Broadcaster run scripts from using same CLI port
```
leszko: I would rephrase it, because:
1. It's not longer a bug fix, it's actually a feature (to enable both split O/T and combined OT setup in devtool)
2. I'd mentioned that it's about devtool

cmd/devtool/README.md
```diff
@@ -31,5 +31,7 @@ This command will submit the setup transactions for a broadcaster and generate t
 
 `go run cmd/devtool/devtool.go setup transcoder`
 
-This command will submit the setup transactions for an orchestrator/transcoder and generate the Bash scripts 
-`run_orchestrator_<ETH_ACCOUNT>.sh` which can be used to start an orchestrator node and `run_transcoder_<ETH_ACCOUNT>.sh` which can be used to start a transcoder node.
+This command will submit the setup transactions for an orchestrator/transcoder and generate the Bash scripts:
+
+* `run_orchestrator_with_transcoder_<ETH_ACCOUNT>.sh` which can be used to start an orchestrator node that contains a transcoder
+* `run_orchestrator_standalone_<ETH_ACCOUNT>.sh`  and `run_transcoder_<ETH_ACCOUNT>.sh` which can be used to start separate orchestrator and transcoder nodes.
```
leszko:
```suggestion
* `run_orchestrator_with_transcoder_<ETH_ACCOUNT>.sh` which can be used to start an orchestrator node that contains a transcoder (combined OT)
* `run_orchestrator_standalone_<ETH_ACCOUNT>.sh`  and `run_transcoder_<ETH_ACCOUNT>.sh` which can be used to start separate orchestrator and transcoder nodes (split O/T)
```

cmd/devtool/devtool.go
```diff
@@ -133,6 +132,7 @@ func main() {
 	ethSetup(acc, keystoreDir, isBroadcaster)
 	createRunScript(acc, dataDir, serviceHost, isBroadcaster)
 	if !isBroadcaster {
+		cliPort += 1
```
leszko, line 135: How about just hardocoding the cli ports rather than incrementing it here? We can just use:
- `7935` for broadcaster
- `7936` for orchestrator (both standalone and with transcoder)
- `7937` for transcoder

cmd/devtool/devtool.go
```diff
@@ -352,18 +349,29 @@ func createRunScript(ethAcctAddr, dataDir, serviceHost string, isBroadcaster boo
 
 	if !isBroadcaster {
 		script += fmt.Sprintf(` -initializeRound=true \
-    -serviceAddr %s:%d  -transcoder=true -orchestrator=true \
+    -serviceAddr %s:%d -orchestrator=true \
     -orchSecret secre -pricePerUnit 1
     `, serviceHost, mediaPort)
+		fName := fmt.Sprintf("run_%s_standalone_%s.sh", getNodeType(isBroadcaster), ethAcctAddr)
```
leszko, line 371: The `standalone` script in your PR does not work, because it misses `\` before the last line.

You have the following:
```
    -orchSecret secre -pricePerUnit 1
     -transcoder=true
```

But it should be
```
    -orchSecret secre -pricePerUnit 1 \
     -transcoder=true
```

cmd/devtool/devtool.go
```diff
@@ -330,13 +330,10 @@ func ethSetup(ethAcctAddr, keystoreDir string, isBroadcaster bool) {
 func createTranscoderRunScript(ethAcctAddr, dataDir, serviceHost string) {
```
leszko, line 330: Not related to your PR, but we could clean up a little bit the generated scripts, now some of the commands are in new lines, some of them are in the same, I'd make it consistent. Wdyt?

Generated `run_broadcaster.sh` (some parameters are in new lines, some are in the same):
```
#!/bin/bash
./livepeer -v 99 -ethController 0x77a0865438f2efd65667362d4a8937537ca7a5ef -datadir ./.lpdev2/broadcaster_0x06AAaf48472D6a64f33C7e281b5dB0fE419878F6 \
    -ethAcctAddr 0x06AAaf48472D6a64f33C7e281b5dB0fE419878F6 \
    -ethUrl http://localhost:8545/ \
    -ethPassword "" \
    -network=devenv \
    -blockPollingInterval 1 \
    -monitor=false -currentManifest=true -cliAddr 127.0.0.1:7935 -httpAddr 127.0.0.1:8935  -broadcaster=true -rtmpAddr 127.0.0.1:1935
```

Generated `run_trascoder.sh` (everything is in one line):
```
#!/bin/bash
./livepeer -v 99 -datadir ./.lpdev2/transcoder_0x065195849B39Bf5D99b4EbD28F61C75C6EB8771a -orchSecret secre -orchAddr 127.0.0.1:8936 -cliAddr 127.0.0.1:7937 -transcoder
```

leszko (commented):
Thanks for the change. I like it. And also great that you added `run*.sh` to `.gitignore`. Added some comments.

cmd/devtool/devtool.go
thomshutt: I'd thought about hardcoding initially, but it looks like this stuff is intended to be more dynamic:

https://github.com/livepeer/go-livepeer/blob/9c3046299b6a850cc32dcc5f732dd970a7cb2b3e/cmd/devtool/devtool.go#L83-L98

Not sure how widely used it is, but it means that we can't just hardcode ports.

cmd/devtool/devtool.go
thomshutt: Good spot, fixed

cmd/devtool/devtool.go
leszko: 👍

cmd/devtool/devtool.go
leszko: @thomshutt Could you check this as well?

leszko (approved):
LGTM

Before merging please:

1. Look also into [this comment](https://github.com/livepeer/go-livepeer/pull/2299/files#r821540212).

2. Squash, rename, and rebase your commits according to what's described in the [Contribution Guide: Commits](https://github.com/livepeer/go-livepeer/blob/master/CONTRIBUTING.md#commits)

thomshutt:
Thanks for all your help here @leszko 🎉

PR #2309, opened 2022-03-14
discovery: Add dynamic timeout for the orchestrator discovery

### LINKED ISSUE #2306: Dynamic timeout for Orchestrator discovery in Broadcaster

Broadcast has a timeout to discover orchestrator, currently `1s`. It causes a problem with teststreams, because if a given Orchestrator does not respond in `1s` it has not stats at all.

Broadcaster should gradually increase the timeout, starting from `500ms` and the maximum `6s`.

```
1. Start from tm = 500ms
2. Try to discover Orchestrators in tm
3. If no Orchestratos are found, then change tm = 2*tm and go to point 2
```

Other alternatives:
- Make timeout configurable with the flag (then we could set a higher number for teststreams)
- Increase the status timeout to some higher number like `4s`.

**Additional context**
- [GH Issue with failing tesstreams](https://github.com/livepeer/go-livepeer/issues/2270)
- [Discord discussion](https://discord.com/channels/423160867534929930/950538877028925480/951190060173238283)

leszko:

Add dynamic timeout for the orchestrator discovery process.

- 
- 
- 

Tested with local geth. Introduced artificial delay in O and checked the logs in B.

fix #2306

**Checklist:**

### DIFF
--- a/discovery/discovery.go
+++ b/discovery/discovery.go
@@ -19,7 +19,8 @@ import (
 )
 
 var getOrchestratorsTimeoutLoop = 3 * time.Second
-var getOrchestratorsCutoffTimeout = 1 * time.Second
+var getOrchestratorsCutoffTimeout = 500 * time.Millisecond
+var maxGetOrchestratorCutoffTimeout = 6 * time.Second
 
 var serverGetOrchInfo = server.GetOrchestratorInfo
 
@@ -77,10 +78,6 @@ func (o *orchestratorPool) GetOrchestrators(ctx context.Context, numOrchestrator
 
 	numAvailableOrchs := len(linfos)
 	numOrchestrators = int(math.Min(float64(numAvailableOrchs), float64(numOrchestrators)))
-	ctx, cancel := context.WithTimeout(clog.Clone(context.Background(), ctx), getOrchestratorsCutoffTimeout)
-
-	infoCh := make(chan *net.OrchestratorInfo, numAvailableOrchs)
-	errCh := make(chan error, numAvailableOrchs)
 
 	// The following allows us to avoid capability check for jobs that only
 	// depend on "legacy" features, since older orchestrators support these
@@ -108,7 +105,7 @@ func (o *orchestratorPool) GetOrchestrators(ctx context.Context, numOrchestrator
 		}
 		return caps.CompatibleWith(info.Capabilities)
 	}
-	getOrchInfo := func(uri *url.URL) {
+	getOrchInfo := func(ctx context.Context, uri *url.URL, infoCh chan *net.OrchestratorInfo, errCh chan error) {
 		info, err := serverGetOrchInfo(ctx, o.bcast, uri)
 		if err == nil && isCompatible(info) {
 			infoCh <- info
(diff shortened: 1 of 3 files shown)


yondonfu (commented):
IIUC the current implementation will re-send requests to all Os after each timeout until B reaches the max timeout. I'm not sure that B should re-send requests to all Os after each timeout because if the O is taking awhile to respond to the first request (and has not returned an error) re-sending the request doesn't seem like it would help. I actually was thinking that B would just extend the timeout for in-flight requests until the max timeout such that there is only ever a single request per O during a discovery process. WDYT? Curious how you'd compare the benefits of re-sending the requests against the approach I mentioned.

If we implement this, I think the top level context could be used and a timer + `timer.Reset()` may come in handy. Something like this:

```golang
        currTimeout := getOrchestratorsCutoffTimeout
	timer := time.NewTimer(currTimeout)
	for i := 0; i < numAvailableOrchs && len(infos) < numOrchestrators && !timeout; i++ {
		select {
		case info := <-infoCh:
			if penalty := suspender.Suspended(info.Transcoder); penalty == 0 {
				infos = append(infos, info)
			} else {
				heap.Push(suspendedInfos, &suspension{info, penalty})
			}
			nbResp++
		case <-errCh:
			nbResp++
		case <-timer.C:
			// The top level context also timed out after maxGetOrchestratorCutoffTimeout
			// Set timeout to true and continue so we exit on next loop iteration
			select {
			case <-ctx.Done():
				timeout = true
			default:
				continue
			}

			// The current timer timed out and we received at least 1 response
			// Set timeout to true and continue so we exit on next loop iteration
			if nbResp > 0 {
				timeout = true
				continue
			}

			// The top level context did not time out yet
			// The current timer timed out, but we did not receive at least response
			// Reset the timer to extend the timeout
                         currTimeout = currTimeout * 2
			timer.Reset(currTimeout)
		}
	}
	cancel()
```

EDIT: I think the benefit of your approach is that *if* a lot of the Os return errors during discovery, re-sending the requests could give those Os an opportunity to return a valid response if they were encountering ephemeral issues which can be helpful in the scenario where there are no other Os to work with. However, as-is, the approach would also involve re-sending requests to Os that B already received valid responses from previously.

leszko:
> the approach would also involve re-sending requests to Os that B already received valid responses from previously.

That is actually not the case, because we only re-send requests if no orchestrator replied with a valid response.

leszko:
> IIUC the current implementation will re-send requests to all Os after each timeout until B reaches the max timeout. I'm not sure that B should re-send requests to all Os after each timeout because if the O is taking awhile to respond to the first request (and has not returned an error) re-sending the request doesn't seem like it would help. I actually was thinking that B would just extend the timeout for in-flight requests until the max timeout such that there is only ever a single request per O during a discovery process. WDYT? Curious how you'd compare the benefits of re-sending the requests against the approach I mentioned.

That is a very good point! And right, your proposed solution has the benefits that we don't start again with the requests from scratch, but just extend the waiting time.

I see the following benefits of each solution.

| Benefits of re-sending requests | Benefits of extending the waiting time |
| -------------------------------  | -------------------------------------- |
| gives a chance to orchestrators which returned errors | less requests received by the orchestrators, won't overload with requests |
| gives a chance to orchestrators which didn't even receive the request (e.g. request was lost in the network) | lower overall latency, because e.g. if O replies in `3s`, then we wait `4s`, don't need to wait `7.5s = 0.5s + 1s + 2s + 4s` |
| gives a chance to orchestrators which were temporary down | |

Each solution will work for us. And each solution has its benefits, the first one is more tolerant, the second optimizes for the load and latency. Our context is that we won't actually see much retries in the real prod env, because the starting `500ms` is already a high number for at least 1 O to return its response. So, I see this feature for two use-cases:
- teststreams
- some network anomaly (probably temporary and limited to a region)

Looking into these 2 cases, I'll probably not optimize for load or latency, but for a more "bullet proof" solution, so keeping it as it is. WDYT? @yondonfu

discovery/discovery.go
```diff
@@ -129,30 +126,52 @@ func (o *orchestratorPool) GetOrchestrators(ctx context.Context, numOrchestrator
...
-	nbResp := 0
-	for i := 0; i < numAvailableOrchs && len(infos) < numOrchestrators && !timeout; i++ {
-		select {
-		case info := <-infoCh:
-			if penalty := suspender.Suspended(info.Transcoder); penalty == 0 {
-				infos = append(infos, info)
-			} else {
-				heap.Push(suspendedInfos, &suspension{info, penalty})
+		timedOut = false
+		suspendedInfos = newSuspensionQueue()
```
yondonfu, line 146: Just want to confirm that the following is the desired behavior:

- If an O is suspended and responds to the discovery request it will be pushed to the suspension queue and will not be counted as a valid response
- If we timeout, then we reset the suspension queue during the next discovery round
- If we hit the max timeout, *then* we will consider using the suspended O in the code block starting on L176
- We end up sending discovery requests to suspended Os more than once if we do multiple rounds of discovery
- The impact is that B priorities using Os that are slower in discovery over Os that are suspended

This seems reasonable, but just want to double check that it is the expected behavior.

yondonfu (changes requested):
Would it be possible to add a unit test for the new discovery behavior by stubbing `serverGetOrchInfo()` as is done [here](https://github.com/livepeer/go-livepeer/blob/master/discovery/discovery_test.go#L58)?

yondonfu:
@leszko 

> Looking into these 2 cases, I'll probably not optimize for load or latency, but for a more "bullet proof" solution, so keeping it as it is. WDYT?

Sounds reasonable to me.

discovery/discovery.go
leszko: Yes, I think it's correct. And believe it's the right thing to do, to prioritize slower Os over suspended Os.

leszko:
> Would it be possible to add a unit test for the new discovery behavior by stubbing `serverGetOrchInfo()` as is done [here](https://github.com/livepeer/go-livepeer/blob/master/discovery/discovery_test.go#L58)?

Technically it's possible, but the approach with calling `time.sleep()` in tests comes with drawbacks and we need to weigh if they don't exceed the benefits. The drawbacks are:
- Every call to `time.sleep()` effectively increases the test suite execution time
- Every call to `time.sleep()` makes the tests (more) flaky

So, personally, I'd use reserve using sleeping in tests for the cases where we really need it (e.g. `push_test.go`) and not test the retries. In this case, a simple retry tests add `500ms`, but if we want to test the whole exp. backoff spectrum, we add > `10s`.

leszko:
@yondonfu Addressed your comments. PTAL.

yondonfu:
@leszko 

Agreed that your second point on making tests more flaky should be weighed as a cost relative to the benefit of introducing the suggested unit test. FWIW I think the first point could be addressed by setting a lower custom value for `getOrchestratorsCutoffTimeout` (that can be reset to its original value at the end of the test). With that being said, setting a lower custom value probably will exacerbate second point as the test could become flakier with lower timeouts and values passed to `time.Sleep()`.

I'm comfortable moving forward without the unit test. Do you think the retry behavior could be a good candidate for an e2e test scenario once we're ready to add e2e tests?

yondonfu (approved):
LGTM

leszko:
> FWIW I think the first point could be addressed by setting a lower custom value for `getOrchestratorsCutoffTimeout`
Yeah, agree, we could mitigate it with overriding getOrchestratorsCutoffTimeout in tests.

> I'm comfortable moving forward without the unit test. Do you think the retry behavior could be a good candidate for an e2e test scenario once we're ready to add e2e tests?
Yes, I think it's maybe not exactly he scenario for this timeout, but some scenarios for reliability in terms of discovery.

Issue #2231, opened 2022-02-03
Advertise changes in O capabilities to B when new T connects or disconnects


cyberj0g:
Separate issue to track [this](https://github.com/livepeer/go-livepeer/issues/2123#issuecomment-1027782218) part of new codecs support implementation.

yondonfu:
Closed by #2312

PR #2312, opened 2022-03-15
Update O capabilities when T connects and disconnects, basic capacity support

cyberj0g:
Part of #2123 implementation, specifically #2231
It adds capacity map to the capabilities structure, which is currently used only for standalone T accounting on O. With this change, B will know if O has Ts with capabilities required for job upon next [refreshSessions()](https://github.com/livepeer/go-livepeer/blob/10705d2467decefce6f1bea030d005adcfebfc23/server/broadcast.go#L120) call.

**What it doesn't do**
It doesn't affect B-O discovery in any way. Theoretically, capability discovery should be decoupled with node discovery and, if standalone Ts would be common, should happen more frequently, maybe continuously.

- add capacity map to capabilities structure
- do not add any capabilities on start for standalone O
- add and remove capabilities when T connects and disconnects

New tests, existing tests.

#2231

**Checklist:**

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -76,12 +76,6 @@ jobs:
           go fmt ./...
           git diff --exit-code
 
-      - name: Compile proto files
-        run: |
-          protoc --version
-          touch net/lp_rpc.proto net/redeemer.proto
-          make net/lp_rpc.pb.go net/redeemer.pb.go
-
       - name: Lint
         run: |
           export PKG_CONFIG_PATH=~/compiled/lib/pkgconfig
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -81,7 +81,6 @@ jobs:
           protoc --version
           touch net/lp_rpc.proto net/redeemer.proto
           make net/lp_rpc.pb.go net/redeemer.pb.go
-          git diff --exit-code
 
       - name: Lint
         run: |
--- a/core/capabilities.go
+++ b/core/capabilities.go
@@ -2,10 +2,10 @@ package core
 
 import (
 	"errors"
-
 	"github.com/livepeer/go-livepeer/drivers"
 	"github.com/livepeer/go-livepeer/net"
 	"github.com/livepeer/lpms/ffmpeg"
+	"sync"
 )
 
 type Capability int
@@ -15,6 +15,8 @@ type Capabilities struct {
 	bitstring   CapabilityString
 	mandatories CapabilityString
 	constraints Constraints
+	capacities  map[Capability]int
+	mutex       sync.Mutex
 }
 type CapabilityTest struct {
 	inVideoData []byte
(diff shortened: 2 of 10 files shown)


go.mod
```diff
@@ -43,3 +43,5 @@ require (
 	google.golang.org/grpc v1.38.0
 	pgregory.net/rapid v0.4.0
 )
+
+replace github.com/livepeer/lpms => ../lpms
```
thomshutt: Was this intentional? I think it's what's causing the build to break

cyberj0g: Yes, you are right. We usually pin LPMS version right before merge once PR is approved, so I didn't pay attention to `go.mod` for now.

core/capabilities_test.go
```diff
@@ -14,6 +14,56 @@ import (
...
+	// convert to net.Capabilities and back
+	caps = CapabilitiesFromNetCapabilities(caps.ToNetCapabilities())
+	// check all capacities present and set to 1
+	assert.Equal(len(caps.capacities), len(capabilities))
+	for _, c := range capabilities {
+		v, exist := caps.capacities[c]
+		assert.True(exist)
+		assert.Equal(1, v)
+	}
+	// test increase capacity
```
thomshutt, line 32: Could we split this up into multiple tests? i.e converting to/from net.Capabilities, increasing capacity, decreasing capacity separately

core/capabilities.go
```diff
@@ -281,30 +283,97 @@ func (c *Capabilities) ToNetCapabilities() *net.Capabilities {
...
+		newCapacity := curCapacity - capacity
+		if newCapacity <= 0 {
+			delete(cap.capacities, capability)
+			arrIdx := int(capability) / 64
+			bitIdx := int(capability) % 64
+			if arrIdx >= len(cap.bitstring) {
+				// shouldn't reach this point
+				continue
+			}
+			cap.bitstring[arrIdx] &= ^uint64(1 << bitIdx)
```
thomshutt: Since we're doing this bit manipulation in a couple of places, might be worth splitting it out into its own method to reduce the size of these one and allow it to be testable on its own

cyberj0g: Refactored a bit.

core/capabilities_test.go
cyberj0g: Yes we could, but I'm not sure if it's necessary, this test is focused on pretty small piece of code.

core/capabilities.go
```diff
@@ -281,30 +276,111 @@ func (c *Capabilities) ToNetCapabilities() *net.Capabilities {
 	if c == nil {
 		return nil
 	}
-	return &net.Capabilities{Bitstring: c.bitstring, Mandatories: c.mandatories}
+	netCaps := &net.Capabilities{Bitstring: c.bitstring, Mandatories: c.mandatories, Capacities: make(map[uint32]uint32)}
```
AlexKordic, line 281: Makes sense to use mutex here as this looks like place of consumption of `.capacities` by upstream code.
```suggestion
	netCaps := &net.Capabilities{Bitstring: c.bitstring, Mandatories: c.mandatories, Capacities: make(map[uint32]uint32)}
	c.mutex.Lock()
	defer c.mutex.Unlock()
```

core/capabilities.go
```diff
@@ -281,30 +276,111 @@ func (c *Capabilities) ToNetCapabilities() *net.Capabilities {
...
 
+func (cap *Capabilities) AddCapacity(newCaps *Capabilities) {
+	cap.mutex.Lock()
+	defer cap.mutex.Unlock()
+	for capability, capacity := range newCaps.capacities {
+		curCapacity, e := cap.capacities[capability]
+		if !e {
+			cap.capacities[capability] = 0
+		}
+		cap.capacities[capability] = curCapacity + capacity
```
AlexKordic, line 338: Do we have to worry about int overflow here?

.github/workflows/test.yaml
```diff
@@ -81,7 +81,6 @@ jobs:
           protoc --version
           touch net/lp_rpc.proto net/redeemer.proto
           make net/lp_rpc.pb.go net/redeemer.pb.go
-          git diff --exit-code
```
AlexKordic, line 84: Isn't this good check to have?

AlexKordic (approved):
Well done! Posted few questions.

core/capabilities.go
cyberj0g: It's good to note that it potentially could overflow in 'giga-orchestrator' setting, but as it's not a user supplied value, and we don't ever expect any single O to serve billion of transcoders, I think we can leave it as is for clarity.

core/capabilities.go
cyberj0g: Fixed

.github/workflows/test.yaml
cyberj0g: I'm not sure this is a good check to have, and the step should probably be removed altogether, because:
* different *minor* versions of protoc generate different files, I think it's not a good idea to force developers to install specific protoc version, which may not be available from OS package manager
* specific protoc version is not pinned above in this job, when it's installed through apt, so the check may break - as already happened for me locally
* no relevant info in CONTRIBUTING.md
* the check of auto generated file is neither semantic nor syntactic
* all protocol functions should be covered by unit tests explicitly

Let me know what you think. Maybe @darkdarkdragon can provide any additional context on this too.

oscar-davids (approved):
LGTM!

core/capabilities.go
AlexKordic: Got it. 
In the future, do we want to have encoding capacities in pixels per second? If yes that changes the scale.

Issue #2285, opened 2022-02-22
Reduce number of stream-specific metrics


iameli:
Right now it clogs everything up, oh no

hthillman:
@iameli can you specify what will be removed? is this a matter of reducing frequency or taking out some stats?

darkdarkdragon:
Example /metrics output would really help here

figintern:
https://github.com/livepeer/go-livepeer/pull/2313

figintern:
https://www.notion.so/livepeer/Single-Stream-Health-OpenCensus-b402a38f5cf54479a7ad9845c57e5604

PR #2313, opened 2022-03-16
Reduce dimensionality of per-stream metrics

figintern:
https://www.notion.so/livepeer/Single-Stream-Health-OpenCensus-b402a38f5cf54479a7ad9845c57e5604
go-livepeer exports too many metrics when `metricsPerStream` and `metricsExposeClientIP` flags are turned
 Analyzing the /metrics endpoint reveals the high cardinality of metrics is caused by in part by the increase in dimensionality from the new `manifestId` and `clientIp` tags along with (> 5 bucket count) Distribution views forming too high of a `dimensionality`
This is supposedly overwhelming the Prometheus scraper and causing telemetry drop-off
Requesting /metrics on a go-livepeer pod with per-stream metrics enabled at the end of a day results in a ~6500 line 1.2M file (per Eli's Gist: https://gist.github.com/iameli/def99f35d1b649337d863d5b349a66ad#file-metrics-oneday-txt) 
Requesting /metrics on a go-livepeer (bctr) pod without the flags enabled results in a ~850 line 132kb file

- metricsPerStream flag no longer required to enable metrics with EthAddr
- changed `http_client_segment_transcoded_realtime_ratio` back to baseTags `(~1000 lines / 6500, ~15%)`
- added `http_client_segment_transcoded_realtime_ratio_per_stream` using single bucket Distribution view
- added `http_client_segment_transcoded_realtime_ratio_per_client_ip` using single bucket Distribution view
- changed `transcode_overall_latency_seconds` back to baseTags `(~1360 lines / 6500, ~21%)`
- added `transcode_overall_latency_seconds_per_stream` using single bucket Distribution view
- added `upload_time_seconds_per_stream` using single bucket Distribution view
- changed `download_time_seconds` back to baseTags `(~730 lines / 6500, ~11%)`
- added `download_time_seconds_per_stream` using single bucket Distribution view
- changed `source_segment_duration_seconds` back to baseTags `(~1550 lines / 6500, ~24%)`
- added `source_segment_duration_seconds_per_stream` using single bucket Distribution view
- changed `segment_source_uploaded_total` back to baseTags `(~220 lines / 6500, ~3%)`
- added `segment_source_uploaded_total_per_stream` which drops kOrchestratorURI dimension
- changed `segment_source_upload_failed_total` back to baseTags `(~80 lines / 6500, ~1%)`
- added `segment_source_upload_failed_total_per_stream` which drops kOrchestratorURI dimension

The above updates should reduce cardinality/size of metrics export by ~50%

https://github.com/livepeer/go-livepeer/issues/2285

**Checklist:**

### DIFF
--- a/monitor/census.go
+++ b/monitor/census.go
@@ -323,17 +323,18 @@ func InitCensus(nodeType NodeType, version string) {
 		glog.Fatal("Error creating tagged context", err)
 	}
 	baseTags := []tag.Key{census.kNodeID, census.kNodeType}
+	baseTagsWithEthAddr := append([]tag.Key{census.kSender}, baseTags...)
 	baseTagsWithManifestID := baseTags
-	baseTagsWithEthAddr := baseTags
-	baseTagsWithManifestIDAndEthAddr := baseTags
+	baseTagsWithClientIP := baseTags
+	baseTagsWithEthAddrAndManifestID := baseTagsWithEthAddr
+	baseTagsWithManifestIDAndClientIP := baseTagsWithEthAddr
 	if PerStreamMetrics {
-		baseTagsWithManifestID = []tag.Key{census.kNodeID, census.kNodeType, census.kManifestID}
-		baseTagsWithEthAddr = []tag.Key{census.kNodeID, census.kNodeType, census.kSender}
-		baseTagsWithManifestIDAndEthAddr = []tag.Key{census.kNodeID, census.kNodeType, census.kManifestID, census.kSender}
+		baseTagsWithManifestID = append([]tag.Key{census.kManifestID}, baseTags...)
+		baseTagsWithEthAddrAndManifestID = append([]tag.Key{census.kManifestID}, baseTagsWithEthAddr...)
 	}
-	baseTagsWithManifestIDAndIP := baseTagsWithManifestID
 	if ExposeClientIP {
-		baseTagsWithManifestIDAndIP = append([]tag.Key{census.kClientIP}, baseTagsWithManifestID...)
+		baseTagsWithClientIP = append([]tag.Key{census.kClientIP}, baseTags...)
+		baseTagsWithManifestIDAndClientIP = append([]tag.Key{census.kClientIP}, baseTagsWithManifestID...)
 	}
 
 	views := []*view.View{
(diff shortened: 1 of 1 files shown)


monitor/census.go
```diff
@@ -369,56 +370,70 @@ func InitCensus(nodeType NodeType, version string) {
...
+			TagKeys:     baseTagsWithManifestID,
+			Aggregation: view.Distribution(), // single bucket with boundaries (-infinity, +infinity)
+		},
+		{
+			Name:        "http_client_segment_transcoded_realtime_ratio_per_client_ip",
+			Measure:     census.mRealtimeRatio,
+			Description: "Ratio of source segment duration / transcode time as measured on HTTP client",
+			TagKeys:     baseTagsWithClientIP,
+			Aggregation: view.Distribution(), // single bucket with boundaries (-infinity, +infinity)
+		},
```
victorges, line 403: What data exactly will these metrics provide? It it just gonna be a count o the number of segments transcoded for each manfiest ID and each IP? Or will the realtime ratio still be present in some way in the exported data, even it having a single bucket?

If that's the case might as well make it an explicit "count" metric instead 🤔 
Like brekaing up `segment_transcoded_total` instead.

monitor/census.go
```diff
@@ -369,56 +370,70 @@ func InitCensus(nodeType NodeType, version string) {
...
+		{
+			Name:        "http_client_segment_transcoded_realtime_ratio_per_stream",
+			Measure:     census.mRealtimeRatio,
+			Description: "Ratio of source segment duration / transcode time as measured on HTTP client",
+			TagKeys:     baseTagsWithManifestID,
+			Aggregation: view.Distribution(), // single bucket with boundaries (-infinity, +infinity)
+		},
+		{
+			Name:        "http_client_segment_transcoded_realtime_ratio_per_client_ip",
+			Measure:     census.mRealtimeRatio,
```
victorges, line 399: Is it ok to reuse this `mRealtimeRatio` on multiple the views? I don't really understand the opencensus abstractions 😬

monitor/census.go
```diff
@@ -369,56 +370,70 @@ func InitCensus(nodeType NodeType, version string) {
...
+			TagKeys:     baseTagsWithManifestIDAndClientIP,
 			Aggregation: view.Count(),
 		},
 		{
 			Name:        "http_client_segment_transcoded_realtime_slow",
 			Measure:     census.mRealtimeSlow,
 			Description: "Number of segment transcoded more than two times slower than realtime",
-			TagKeys:     baseTagsWithManifestIDAndIP,
+			TagKeys:     baseTagsWithManifestIDAndClientIP,
 			Aggregation: view.Count(),
```
victorges, line 437: These `http_client_segment_transcoded_realtime_*` are actually just the old versions of the `http_client_segment_transcoded_realtime_ratio` histogram/distribution. The latter is actually a super set of these, since it contains the 0.5, 1, 2 and 3 values in the distribution.

That said, I think these legacy metrics can be made to use only the `baseTags`, and the per-stream and per-IP metrics can already be checked from that newer histogram metric instead. WDYT?

monitor/census.go
```diff
@@ -439,21 +454,35 @@ func InitCensus(nodeType NodeType, version string) {
...
 		},
 		{
 			Name:        "segment_source_uploaded_total",
 			Measure:     census.mSegmentUploaded,
 			Description: "SegmentUploaded",
-			TagKeys:     append([]tag.Key{census.kOrchestratorURI}, baseTagsWithManifestID...),
+			TagKeys:     append([]tag.Key{census.kOrchestratorURI}, baseTags...),
```
victorges, line 464: Should this be a `baseTagsWithOrchestrator` as well, for consistency?

monitor/census.go
```diff
@@ -439,21 +454,35 @@ func InitCensus(nodeType NodeType, version string) {
...
+			Description: "SegmentUploaded",
+			TagKeys:     baseTagsWithManifestID,
 			Aggregation: view.Count(),
 		},
 		{
 			Name:        "segment_source_upload_failed_total",
 			Measure:     census.mSegmentUploadFailed,
 			Description: "SegmentUploadedFailed",
-			TagKeys:     append([]tag.Key{census.kErrorCode, census.kOrchestratorURI}, baseTagsWithManifestID...),
+			TagKeys:     append([]tag.Key{census.kErrorCode, census.kOrchestratorURI}, baseTags...),
```
victorges, line 478: Here we can use the `WithOrchestrator` as well (maybe too much to create one with error code as well

monitor/census.go
```diff
@@ -516,9 +545,16 @@ func InitCensus(nodeType NodeType, version string) {
...
 			Description: "Transcoding latency, from source segment emerged from segmenter till all transcoded segment apeeared in manifest",
-			TagKeys:     append([]tag.Key{census.kProfiles}, baseTagsWithManifestID...),
+			TagKeys:     append([]tag.Key{census.kProfiles}, baseTags...),
 			Aggregation: view.Distribution(0, .500, .75, 1.000, 1.500, 2.000, 2.500, 3.000, 3.500, 4.000, 4.500, 5.000, 10.000),
 		},
+		{
+			Name:        "transcode_overall_latency_seconds_per_stream",
+			Measure:     census.mTranscodeOverallLatency,
+			Description: "Transcoding latency, from source segment emerged from segmenter till all transcoded segment apeeared in manifest",
+			TagKeys:     append([]tag.Key{census.kProfiles}, baseTagsWithManifestID...),
```
victorges, line 555: I'm not sure if this separation makes sense. A given `manifestId` will always have the same profiles, so this `transcode_overall_latency_seconds_per_stream` will have exactly the same cardinality/values as `transcode_overall_latency_seconds`. 

Or is this `census.kProfiles` not what I think it is?

monitor/census.go
```diff
@@ -554,13 +590,27 @@ func InitCensus(nodeType NodeType, version string) {
 			TagKeys:     append([]tag.Key{census.kOrchestratorURI}, baseTags...),
 			Aggregation: view.Distribution(0, .10, .20, .50, .100, .150, .200, .500, .1000, .5000, 10.000),
 		},
+		{
+			Name:        "upload_time_seconds_per_stream",
+			Measure:     census.mUploadTime,
+			Description: "UploadTime, seconds",
+			TagKeys:     append([]tag.Key{census.kOrchestratorURI}, baseTagsWithManifestID...),
```
victorges, line 597: Did you mean only `baseTagsWithManifestID` here?

If we also have `kOrchestratorURI` this means that we can derive `upload_time_seconds` from this `upload_time_seconds_per_stream`, which means that we don't need to create the separate metric here.

Edit: Just saw now that this is using a single bucket distribution instead. This is the main thing I am confused about this PR which is repeated on many metrics, I'll stop commenting on the reoccurencies now. Is this the OpenCensus way of creating a [summary](https://prometheus.io/docs/practices/histograms/#count-and-sum-of-observations)?

victorges (approved):
LGTM but I'm really confused about the empty `view.Distribution()`. Does it just create a prometheus summary instead, with a `sum` and a `count` of the observations?

monitor/census.go
figintern: Yup from examining the output metrics I found that the best way to create a summary with count+sum and a single bucket distribution is to use an empty view.Distribution(). There didn't seem to be a convention available to do it otherwise without explicitly creating more metrics specifically for `sum` or `count`. There isn't any documentation on (https://opencensus.io/stats/view/#aggregations) covering this specifically but when I used it we get a single bucket distribution + sum + count metric exported (summary).

figintern:
> LGTM but I'm really confused about the empty `view.Distribution()`. Does it just create a prometheus summary instead, with a `sum` and a `count` of the observations?

Yup this is essentially what it accomplishes. More details in comment above.

> Edit: Just saw now that this is using a single bucket distribution instead. This is the main thing I am confused about this PR which is repeated on many metrics, I'll stop commenting on the reoccurencies now. Is this the OpenCensus way of creating a [summary](https://prometheus.io/docs/practices/histograms/#count-and-sum-of-observations)?

monitor/census.go
figintern: Yup a `view` combines a measure such as `census.mRealtimeRatio` with an `Aggregation` such as `Distribution` (which includes a bucket distribution, count, and a sum) to export metrics through the exporter.

You can reuse different `measures` in different views, same as you can choose different Aggregation types.

monitor/census.go
victorges: This was just a confusion about `Distribution()` as well. Explained in other replies.

monitor/census.go
victorges: Ah I see! And what is exactly this single bucket distribution? Is it, in practice, another `count`?

Nice trick to avoid creating separate `sum` and` count` metrics btw :)

PR #2337, opened 2022-03-24
Fix dynamic discovery timeout to not retry sending requests, but wait for the same request to complete

leszko:
The last change with the dynamic timeout caused an issue in teststreams described in [Discord](https://discord.com/channels/423160867534929930/932724294230900776/956293748340428821).

The [dynamic timeout PR](https://github.com/livepeer/go-livepeer/pull/2309) changed the discovery between B<>O to work as follows:
1. B tries to discover O in 500 ms
2. If no Os found, then increase the timeout to 1s and send the requests again
3. In no Os found, then increase the timeout to 2s and send the requests again

The problem is that if O responds in, let's say, `1.5s`. Then, now the discovery will now take: 0.5s + 1s + 1.5s = 3s
So even though the response is in 1.5s, the overall time from the black-box perspective is 3s. That pollutes the orch teststream data.

This PR changes the dynamic timeout to not send new request, but to wait for the initial requests to complete (as [proposed initially](https://github.com/livepeer/go-livepeer/pull/2309#pullrequestreview-909438232) by @yondonfu)

- 
- 
- 

Tested with local geth. Introduced artificial delay in O and checked the logs in B.

**Checklist:**

### DIFF
--- a/discovery/discovery.go
+++ b/discovery/discovery.go
@@ -127,51 +127,49 @@ func (o *orchestratorPool) GetOrchestrators(ctx context.Context, numOrchestrator
 	}
 
 	var infos []*net.OrchestratorInfo
-	var suspendedInfos *suspensionQueue
-	var timedOut bool
-	var nbResp int
+	suspendedInfos := newSuspensionQueue()
+	timedOut := false
+	nbResp := 0
+	infoCh := make(chan *net.OrchestratorInfo, numAvailableOrchs)
+	errCh := make(chan error, numAvailableOrchs)
 
-	// try to discover orchestrators until at least 1 is found (with the exponential backoff timout)
+	ctx, cancel := context.WithTimeout(clog.Clone(context.Background(), ctx), maxGetOrchestratorCutoffTimeout)
+	for _, uri := range uris {
+		go getOrchInfo(ctx, uri, infoCh, errCh)
+	}
+
+	// try to wait for orchestrators until at least 1 is found (with the exponential backoff timout)
 	timeout := getOrchestratorsCutoffTimeout
-	for {
-		ctx, cancel := context.WithTimeout(clog.Clone(context.Background(), ctx), timeout)
-		infoCh := make(chan *net.OrchestratorInfo, numAvailableOrchs)
-		errCh := make(chan error, numAvailableOrchs)
+	timer := time.NewTimer(timeout)
 
-		for _, uri := range uris {
-			go getOrchInfo(ctx, uri, infoCh, errCh)
-		}
-
-		timedOut = false
-		suspendedInfos = newSuspensionQueue()
-		nbResp = 0
-		for i := 0; i < numAvailableOrchs && len(infos) < numOrchestrators && !timedOut; i++ {
-			select {
-			case info := <-infoCh:
-				if penalty := suspender.Suspended(info.Transcoder); penalty == 0 {
-					infos = append(infos, info)
-				} else {
-					heap.Push(suspendedInfos, &suspension{info, penalty})
-				}
-				nbResp++
-			case <-errCh:
-				nbResp++
-			case <-ctx.Done():
+	for nbResp < numAvailableOrchs && len(infos) < numOrchestrators && !timedOut {
+		select {
+		case info := <-infoCh:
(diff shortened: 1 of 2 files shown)


yondonfu:
> The problem is that if O responds in, let's say, 1.5s. Then, now the discovery will now take: 0.5s + 1s + 1.5s = 3s
So even though the response is in 1.5s, the overall time from the black-box perspective is 3s. That pollutes the orch teststream data.

Hm yeah I think this makes sense.

orch-tester relies on [metrics scraped](https://github.com/livepeer/stream-tester/blob/e2f50535d1afa551d90cb102ed69905b1cea8a56/cmd/orch-tester/orch_tester.go#L419) from the B for the avg upload, download and round trip times. When B receives a segment, it will record the start time [here](https://github.com/livepeer/go-livepeer/blob/master/server/broadcast.go#L718) and then run discovery [here](https://github.com/livepeer/go-livepeer/blob/master/server/broadcast.go#L877). At the beginning of a stream, discovery will block until the working O set is populated (during the stream discovery would happen in the background to re-populate the working set) so if discovery takes longer then that would be reflected in the times recorded for the first segment of the stream. Additionally, discovery will block again if the working O set is empty for any subsequent segment which I think might happen in the scenario where there is only 1 O available and it doesn't return a response fast enough before the next segment arrives [1] - in this scenario discovery would add to the times recorded as well.

[1] I think the threshold is 2x the segment duration so if the segment duration is 2s and the previous segment has taken longer than 4s then we have to re-run discovery, but if it it took less than 4s we just send the next segment to the same O. https://github.com/livepeer/go-livepeer/blob/fce545fa16b6a2292394fe5134a755cd0f830e35/server/broadcast.go#L244

discovery/discovery.go
```diff
@@ -127,51 +127,48 @@ func (o *orchestratorPool) GetOrchestrators(ctx context.Context, numOrchestrator
...
-		cancel()
-
-		if len(infos) > 0 {
-			break
-		}
 
-		timeout *= 2
-		if timeout >= maxGetOrchestratorCutoffTimeout {
-			break
+			timer.Reset(timeout)
```
yondonfu, line 162: Should we move L162-166 first? Otherwise, I think the timer will be set to `getOrchestratorsCutoffTimeout` twice - once when it is initialized on L143 and then again the first time this case is triggered.

discovery/discovery.go
```diff
@@ -127,51 +127,48 @@ func (o *orchestratorPool) GetOrchestrators(ctx context.Context, numOrchestrator
...
-				} else {
-					heap.Push(suspendedInfos, &suspension{info, penalty})
-				}
-				nbResp++
-			case <-errCh:
-				nbResp++
-			case <-ctx.Done():
+	// try to wait for orchestrators until at least 1 is found (with the exponential backoff timout)
+	timeout := getOrchestratorsCutoffTimeout
+	timer := time.NewTimer(timeout)
```
yondonfu, line 143: Do timers need to be cleaned up at all i.e. with Stop() if there is a chance that the `timer.C` channel is not drained before moving on? I'm thinking of the case where right now in the `<-timer.C` case we extend the timer no matter what and then move on to the next iteration of the loop, but if `timedOut` is set to true we exit at the next iteration. At this point, the timer was extended and its `timer.C` channel was not drained. Not sure if that is a problem? If it is, perhaps calling Stop() when we exit the loop or not extending the timer if we're going to exit the loop on the next iteration would work?

discovery/discovery.go
leszko: I think it's correct, because it works like this:
1. We first wait `0.5s`, then we reach this like line
2. We call `timer.Reset(0.5s)`, so now until we reach this line again, we waited in total `0.5s+0.5s=1s`
3. And so on

But it's tricky, so added a comment in 81a6a8a7cb71ca830df7c3025290512123a908e6

discovery/discovery.go
leszko: Good question. I think it's ok to not drain (effectively close) the timer channel. At least according to [this thread](https://stackoverflow.com/questions/8593645/is-it-ok-to-leave-a-channel-open), this is common and the channel will just be garbage collected.

yondonfu (approved):
LGTM after squashing

PR #2345, opened 2022-03-29
livepeer_cli: Improve user feedback when specifying numeric values for some wizard options

### LINKED ISSUE #1602: livepeer_cli gas price validation
**Describe the bug**
Seems like livepeer_cli accepts a new line as input, leading to unexpected behaviour 

**To Reproduce**
choosing `15. Set Eth gas price` leads prompt `Enter new gas price in Wei (enter "0" for automatic)>` that expects a number. 

If for some reason the input is new line, the result is as follows : 

```
What would you like to do? (default = stats)
1. Get node status
2. View protocol parameters
3. List registered orchestrators
4. Invoke "initialize round"
5. Invoke "bond"
6. Invoke "unbond"
7. Invoke "rebond"
10. Invoke "claim" (for rewards and fees)
11. Invoke "transfer" (LPT)
12. Invoke "reward"
13. Invoke multi-step "become an orchestrator"
14. Set orchestrator config
15. Set Eth gas price
16. Sign a message
17. Vote in a poll
> 15
Current gas price: 100
Enter new gas price in Wei (enter "0" for automatic)>
>
```

That is some how misleading. 

if at that point user decides to choose other option ( 1-17 ), livepeer_cli accepts that value as gas price for `15.`

**Screenshots**
Screenshot 2020-08-13 at 16 06 52

kparkins:
This PR addresses issues outlined in Issue #1602 relating to poor user experience in the livepeer_cli when specifying numeric values for some wizard options (gas price, bonding, unbonding, and transfer amount). In particular, it attempts to provide better feedback to the user when invalid values are supplied.

- Add prompt parameter to _readBigInt_ in wizard.go and change to log an error when parsing an int fails.
- Update calls to _readBigInt_ to pass a relevant prompt. 

- Ran all unit tests (PASSED)
- Tested the CLI manually for each of the affected options.

- Fixes #1602 
- Closing 1602 may mean a new issue should be created to follow-up on other wizard options that were not addressed.

**Checklist:**

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -21,7 +21,7 @@
 ### Bug Fixes 🐞
 
 #### CLI
-- \#1602 Improve user feedback when specifying numeric values for some wizard options (@kparkins)
+- \#2345 Improve user feedback when specifying numeric values for some wizard options (@kparkins)
 
 #### General
 - \#2299 Split devtool Orchestrator run scripts into versions with/without external transcoder and prevent Transcoder/Broadcaster run scripts from using same CLI port (@thomshutt)
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -20,6 +20,9 @@
 
 ### Bug Fixes 🐞
 
+#### CLI
+- \#1602 Improve user feedback when specifying numeric values for some wizard options (@kparkins)
+
 #### General
 - \#2299 Split devtool Orchestrator run scripts into versions with/without external transcoder and prevent Transcoder/Broadcaster run scripts from using same CLI port (@thomshutt)
 
--- a/cmd/livepeer_cli/wizard.go
+++ b/cmd/livepeer_cli/wizard.go
@@ -5,7 +5,6 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
-	"github.com/livepeer/go-livepeer/build"
 	"io/ioutil"
 	"math/big"
 	"net/http"
@@ -13,6 +12,8 @@ import (
 	"os"
 	"strconv"
 	"strings"
+
+	"github.com/livepeer/go-livepeer/build"
 
 	"github.com/ethereum/go-ethereum/log"
 	lpcommon "github.com/livepeer/go-livepeer/common"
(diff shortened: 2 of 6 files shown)


CHANGELOG_PENDING.md
```diff
@@ -20,6 +20,9 @@
 
 ### Bug Fixes 🐞
 
+#### CLI
+- \#1602 Improve user feedback when specifying numeric values for some wizard options (@kparkins)
```
```suggestion
- \#2345 Improve user feedback when specifying numeric values for some wizard options (@kparkins)
```

This should be the PR number (not the GH Issue number).

cmd/livepeer_cli/wizard.go
```diff
@@ -161,16 +162,16 @@ func (w *wizard) readDefaultInt(def int) int {
 	return def
 }
 
-func (w *wizard) readBigInt() *big.Int {
+func (w *wizard) readBigInt(prompt string) *big.Int {
+	if prompt == "" {
```
leszko: Do we need this check? I don't think we ever use `readBigInt("")`.

cmd/livepeer_cli/wizard_bond.go
```diff
@@ -165,8 +165,7 @@ func (w *wizard) bond() {
 
 	amount := big.NewInt(0)
 	for amount.Cmp(big.NewInt(0)) == 0 || balBigInt.Cmp(amount) < 0 {
-		fmt.Printf("Enter bond amount - ")
-		amount = w.readBigInt()
+		amount = w.readBigInt("Enter bond amount - > ")
```
leszko: How about passing just `"Enter bond amount"` to the `readBigInt` function? Then we can have something like this in the function (to make the UX more consistent).

```go
func (w *wizard) readBigInt(msg string) *big.Int {
    promt := fmt.Sptinf("%s - > ", msg)
}
```

leszko (commented):
Thanks the PR @kparkins . I think it'll be a good improment. Added a very minor comments, please take a look.

cmd/livepeer_cli/wizard.go
kparkins: I think you're right about this. Probably will never call it with empty string.

cmd/livepeer_cli/wizard_bond.go
kparkins: I debated this with myself as well. The problem, like you said, is the CLI is not consistent atm. I.e. the bonding amount has `-  > ` as the cursor. But, some other options use only `> `.  E.g. `"Enter new maximum gas price in Wei (enter \"0\" for no maximum gas price) > "`

So, pushing this down into the function would change all of them for this function but then there would be inconsistency between values that use _readBigInt_ and values that use, for instance, _readInt_ or _readString_ and these options are often part of the same sub-menu. I decided to bubble it up so it could be controlled without changing the _readBigInt_ function for this reason (no implicit behavior). The changes preserve existing behavior but if instead, it would be preferable to change the 'standard' prompt to `- >` this will create new inconsistencies. If we want to bubble down the `- > ` I would suggest using only `> `. This would preserve behavior for the current options (they seem to have consistency within a specific sub-menu) at the expense of the global consistency that I think you are alluding to.

CHANGELOG_PENDING.md
kparkins: Ohh. Thanks.

leszko (approved):
LGTM 👍 

Thanks for the PR @kparkins

PR #2347, opened 2022-03-31
server: Fix "Too many connections" bug

leszko:

There is a problem with the session cleanup if there is any orchestrator that is transcoding for more than `1 min`.

This is what I believe happens.

1. Session cleanup happens in the `runCheck()` function [1] 
2. We call this session cleanup after `ticker.C`, so `1 min`  after the session started [2]
3. We never call `runCheck()` again (but we should! And we did before the "Add pixel-format in capability logic" PR)

This actually explains the flakiness, because I believe test streams mostly finish the session within 1 min. And I think that the only reason we don't see this issue in the prod broadcasters is that we didn't deploy `master` to prod for a long time.

[1] https://github.com/livepeer/go-livepeer/blob/master/server/mediaserver.go#L840
[2] https://github.com/livepeer/go-livepeer/blob/master/server/mediaserver.go#L867

**Checklist:**

- [ ] ~~[Pending changelog](./CHANGELOG_PENDING.md) updated~~

### DIFF
--- a/server/mediaserver.go
+++ b/server/mediaserver.go
@@ -863,13 +863,15 @@ func (s *LivepeerServer) HandlePush(w http.ResponseWriter, r *http.Request) {
 						}
 					}
 				}
-				select {
-				case <-ticker.C:
-					if runCheck() {
+				for {
+					select {
+					case <-ticker.C:
+						if runCheck() {
+							return
+						}
+					case <-s.context.Done():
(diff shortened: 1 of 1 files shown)


AlexKordic (approved):
Correct logic! Good find

thomshutt (approved):

Issue #2317, opened 2022-03-16
E2E Test Scenario: Register Orchestrator


leszko:
Create an [E2E Test](https://github.com/livepeer/go-livepeer/issues/2297) which covers the following scenario:
1. Start Orchestrator
2. Register Orchestrator
3. Check that Orchestrator is registered and activated

**In this work, we should focus on a few aspects:**
- Consider refactoring `livepeer.go`. Currently it's difficult to start a Livepeer node from the code, because `livepeer.go` is designed as a command-line tool, not a library to start Livepeer nodes. On the other hand we want to start a Livepeer node from the code. That's why it'd be good to extract the code responsible for starting the node to a separate struct `LivepeerOrchestrator` filled using flags or tests parameters.
- Decide if we execute operations on Orchestrator using HTTP calls or from the code
- For the chain we can use the `livepeer/geth-with-livepeer-protocol` Docker image started with the [testcontainer](https://github.com/testcontainers/testcontainers-go) library.

**Resources:**
- [E2E tests Feature Spec](https://github.com/livepeer/go-livepeer/issues/2297)
- [PoC for this work](https://github.com/leszko/go-livepeer/pull/4)

**Test structure**

```golang
func TestRegisterOrchestrator(t *testing.T) {
	// given
	gethC := setupGeth(context.TODO(), t)
	defer gethC.Terminate(context.TODO())
	ethClient := newEthClient(gethC.URI, t)

	// when
	orch := startOrchestrator(ethClient)
	registerOrchestrator(orch)
	
	// then
	assertOrchestratorRegisteredAndActivated(orch)
}
```

Issue #2297, opened 2022-03-04
E2E tests


leszko:
## Abstract
Add automated E2E/Integration tests that cover on-chain interactions.

## Motivation
Currently, we don't test Livepeer node software with the chain in any automated way. That approach results in:
- A lot of time spent on manually testing each chain-related change
- Bugs in the released software (e.g. #2283 or #2152)

Our current test suite covers unit tests and [video-related integration tests](https://github.com/livepeer/go-livepeer/blob/master/server/push_test.go), but we use [mocks for all on-chain interactions](https://github.com/livepeer/go-livepeer/blob/master/eth/stubclient.go).

## Proposed Solution
I suggest creating a **separate suite of tests** in Golang, which uses the **testcontainer library**. Then, implement **white-box tests** covering **user-facing scenarios**.

[Here](https://github.com/leszko/go-livepeer/pull/4)'s a PoC code skeleton how it would look like in practice.

### 1. Separate suite of tests
The E2E/Integration tests need to have a connection to the chain, so they'll have specific environment requirements, that's why I suggest to create:
- separate `test_e2e.sh` script to run them
- separate directory in `go-livepeer` called `tests_e2e` (or `e2e`)
- separate GH Action

### 2. Testcontainer library
As for the chain I think we should use [Livepeer Geth docker image](https://github.com/livepeer/docker-livepeer/tree/master/geth-with-protocol), the same what we currently use for the local dev testing. To integrate it well with Golang tests, we can make use of the [testcontainer](https://github.com/testcontainers/testcontainers-go) library, which enables running integration tests in the exactly same way we run unit tests (using `go test` or directly from IDE). The only requirement for the tests is to have Docker running.

### 3. White-box testing
There are two ways we can approach creating tests: **black-box testing** and **white-box testing**.

In our case, we can imagine a sample **black-box** test as the following script:
```bash
# 1. Start local Geth
docker run livepeer/geth-with-livepeer-protocol:streamflow
# 2. Start Orchestrator
livepeer -network=devenv -ethUrl http://localhost:8545/ -orchestrator
# 3. Register Orchestrator
curl http://localhost:7936/activateOrchestrator
# 4. Start Broadcaster
livepeer -network=devenv -ethUrl http://localhost:8545/ -broadcaster
# 5. Fund Deposit/Reserve
curl http://localhost:7937/fundDepositAndReserve
# ...
```

[...]

PR #2351, opened 2022-04-05
Refactor livepeer.go to enable running Livepeer node from the code

leszko:

Refactor `livepeer.go` to allow running Livepeer node from the code. After this change, you can start Livepeer node from the code (without passing any flags) in the following manner.

```golang
// Start Livepeer node
go func() {
        // Sample configuration for combined O/T topology
	datadir := "~/temp/livepeer/TEST_OT"
	orchestrator := true
	transcoder := true
	serviceAddr := "127.0.0.1:8935"
	cfg := DefaultLivepeerConfig()
	cfg.datadir, cfg.orchestrator, cfg.transcoder, cfg.serviceAddr = &datadir, &orchestrator, &transcoder, &serviceAddr

        // Starting the Livepeer node
	StartLivepeer(ctx, cfg)
}()
```

The context for this change is the work on implementing [E2E Testing](https://github.com/livepeer/go-livepeer/issues/2297) which would cover interactions with the chain. Additionally, I believe it's a way better code structure when we split the configuration and the flags parsing.

Some ideas for future improvements (not included in this PR):
- Allow running Livepeer as library for end-users (improve LivepeerConfig structure)
- Refactor `main()` to extract it into functions like [`parseOrchAddrs()`](https://github.com/leszko/go-livepeer/blob/0a176ef1fefe1ec7e1a76644c80c6fe1602504a6/cmd/livepeer/livepeer.go#L1313)
- Make sure it's possible to run multiple Livepeer instances in one process (in general it works, but may need some tweaks)

- Create the `LivepeerConfig` type and the `DefaultLivepeerConfig()` function
- Change parsing flags to fill `LivepeerConfig`
- Set nils in `LivepeerConfig` instead of using the `isFlagSet` map
- Extract `parseOrchAddrs()` function

Started Livepeer on-chain and off-chain and played with transcoding and CLI interactions.

No, but it's related to https://github.com/livepeer/go-livepeer/issues/2317

**Checklist:**

### DIFF
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -73,103 +73,85 @@ const RtmpPort = "1935"
 const RpcPort = "8935"
 const CliPort = "7935"
 
+type LivepeerConfig struct {
+	network                      *string
+	rtmpAddr                     *string
+	cliAddr                      *string
+	httpAddr                     *string
+	serviceAddr                  *string
+	orchAddr                     *string
+	verifierURL                  *string
+	ethController                *string
+	verifierPath                 *string
+	localVerify                  *bool
+	httpIngest                   *bool
+	orchestrator                 *bool
+	transcoder                   *bool
+	broadcaster                  *bool
+	orchSecret                   *string
+	transcodingOptions           *string
+	maxAttempts                  *int
+	selectRandFreq               *float64
+	maxSessions                  *int
+	currentManifest              *bool
+	nvidia                       *string
+	testTranscoder               *bool
+	sceneClassificationModelPath *string
+	ethAcctAddr                  *string
+	ethPassword                  *string
+	ethKeystorePath              *string
+	ethOrchAddr                  *string
+	ethUrl                       *string
+	txTimeout                    *time.Duration
+	maxTxReplacements            *int
+	gasLimit                     *int
+	minGasPrice                  *int64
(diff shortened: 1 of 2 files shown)


cmd/livepeer/livepeer.go
```diff
@@ -204,19 +183,277 @@ func main() {
...
+	cfg.authWebhookURL = flag.String("authWebhookUrl", *cfg.authWebhookURL, "RTMP authentication webhook URL")
+	cfg.orchWebhookURL = flag.String("orchWebhookUrl", *cfg.orchWebhookURL, "Orchestrator discovery callback URL")
+	cfg.detectionWebhookURL = flag.String("detectionWebhookUrl", *cfg.detectionWebhookURL, "(Experimental) Detection results callback URL")
+
+	return cfg
+}
+
+// updateNilsForUnsetFlags changes some cfg fields to nil if they were not explicitly set with flags.
+// For some flags, the behavior is different whether the value is default or not set by the user at all.
+func updateNilsForUnsetFlags(cfg LivepeerConfig) LivepeerConfig {
```
thomshutt, line 418: 👍 I've always found it annoying that the Go flags package doesn't give a way to distinguish between zero value / not set, but good to at least do it all in one place

thomshutt (approved):
Looks good, I like the ideas for future steps too

PR #2352, opened 2022-04-05
core: Fix standalone orchestrator not crashing under UnrecoverableError

leszko:

Fix the issue that O does not fail after the unrecoverable error from lpms (ffmpeg).

The issue is described in this [Discord thread](https://discord.com/channels/423160867534929930/960520369465327647/960520767710306304). In short, when GPU crashes, NVIDIA recommends restarting the whole process, since it's not possible to recover. It worked that way until the [Gracefully notify orchestrator in case of a panic in transcoder PR](https://github.com/livepeer/go-livepeer/pull/2094). In the mentioned PR, the following logic was implemented:
1. Recover from LPMS (ffmpeg) panic
2. Send a message T->O
3. Panic

The logic is fine for the split O/T topology, however for the combined OT, there is a separate code which also needs to implement Point 3. Panic.

- 
- 
- 

Artificially introduce `panic()` and check that a standalone O crashes.

**Checklist:**

### DIFF
--- a/core/orchestrator.go
+++ b/core/orchestrator.go
@@ -552,6 +552,9 @@ func (n *LivepeerNode) transcodeSeg(ctx context.Context, config transcodeConfig,
 	start := time.Now()
 	tData, err := transcoder.Transcode(ctx, md)
 	if err != nil {
+		if _, ok := err.(UnrecoverableError); ok {
+			panic(err)
+		}
 		clog.Errorf(ctx, "Error transcoding segName=%s err=%q", seg.Name, err)
 		return terr(err)
 	}
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -34,5 +34,6 @@
 
 #### Orchestrator
 - \#2284 Fix issue with not redeeming tickets by Redeemer (@leszko)
+- \#2352 Fix standalone orchestrator not crashing under UnrecoverableError (@leszko)
 
 #### Transcoder


core/orchestrator.go
```diff
@@ -552,6 +552,9 @@ func (n *LivepeerNode) transcodeSeg(ctx context.Context, config transcodeConfig,
 	start := time.Now()
 	tData, err := transcoder.Transcode(ctx, md)
 	if err != nil {
+		if _, ok := err.(UnrecoverableError); ok {
```
victorges, line 555: So would we need some similar logic from B->O to let it know about the failure, or is it ok to just panic, crash the process, and the connection getting closed abruptly?

If we do close the connection before replying to the broadcaster, I am guessing that the B will treat the error as retryable since it will only see a network error.

I think we didn't have a consensus whether this error should be retryable or not from the B perspective, so raising the issue here so we can make an official decision. WDYT? Should we have a clean response here like in the O->T and only then crash the process instead?

core/orchestrator.go
```diff
@@ -552,6 +552,9 @@ func (n *LivepeerNode) transcodeSeg(ctx context.Context, config transcodeConfig,
...
+			panic(err)
```
victorges, line 556: So is this run from an HTTP handler goroutine? Or maybe a background one?

Cause if it's in the first, l had understood that server would recover from that panic and not crash the process. Is that not the case? Or do we perhaps have or own code to make sure we crash on a panic?

victorges (approved):
Nice! Just a few questions/discussions on the crashing logic

core/orchestrator.go
leszko: There is no problem with B->O, when the connection is closed, B will retry the segment with a different O.

And yes, B treats this error as retryable, but my understanding is that it's a good thing. Because the problem is not with the segment, but with O, so we want to retry the segment with a different O.

core/orchestrator.go
leszko: It's run in a separate (background) goroutine, so there is no issue with HTTP handling the panic.

core/orchestrator.go
victorges: OK cool, let's see how this performs then. Yeah, on the case we saw the error was returned for a `record-tester` stream which certainly has well-behaved segments. Hopefully the problem hadn't been caused by a previous segment from another stream that left the orchestrator stuck on that error, tho I think the retries won't be a big problem anyway. 

If we treated the error as unretriable, it also meant that the stream would be stuck on that orchestrator anyway so we would require further changes in the O balancing logic of the Bs, so let's go with the simpler solution for now.

victorges (approved):
LGTM

PR #2353, opened 2022-04-05
workflow: codesign macOS builds

hjpotter92:

The macOS binaries are signed using apple developer certificates

- adds a `GO_BUILD_DIR` variable, to output built binaries to custom path
- codesign and notarize binaries when building on macos runner

gh-action runs

**Checklist:**

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -5,6 +5,8 @@ on:
   push:
     branches:
       - master
+    tags:
+      - "v*"
 
 jobs:
   linux:
@@ -162,7 +164,7 @@ jobs:
         id: match-tag
         with:
           text: ${{ github.event.workflow_run.head_branch }}
-          regex: '^v([0-9]+\.\d+\.\d+)$'
+          regex: '^(master|main|v[0-9]+\.\d+\.\d+)$'
 
       - name: Codesign and notarize binaries
         if: ${{ steps.match-tag.outputs.match != '' }}
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -165,7 +165,7 @@ jobs:
           regex: '^v([0-9]+\.\d+\.\d+)$'
 
       - name: Codesign and notarize binaries
-        # if: ${{ steps.match-tag.outputs.match != '' }}
+        if: ${{ steps.match-tag.outputs.match != '' }}
         uses: livepeer/action-gh-codesign-apple@latest
         with:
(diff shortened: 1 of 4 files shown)


.github/workflows/build.yaml
```diff
@@ -104,25 +104,27 @@ jobs:
     runs-on: macos-latest
     steps:
       - name: Set build environment
-        run: echo "GOARCH=${{ matrix.arch }}" >> $GITHUB_ENV
+        run: |
+          echo "GOARCH=${{ matrix.arch }}" >> $GITHUB_ENV
+          echo "GO_BUILD_DIR=lp-builds/" >> $GITHUB_ENV
```
hjpotter92, line 111: could not find a way to generate a random build directory name and still use it later for codesigning stage. hardcoded as `lp-builds` here until then.

.github/workflows/build.yaml
```diff
@@ -156,6 +158,24 @@ jobs:
...
           GHA_REF: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.ref || github.ref }}
 
+      - uses: actions-ecosystem/action-regex-match@v2
+        id: match-tag
+        with:
+          text: ${{ github.event.workflow_run.head_branch }}
+          regex: '^v([0-9]+\.\d+\.\d+)$'
+
+      - name: Codesign and notarize binaries
+        if: ${{ steps.match-tag.outputs.match != '' }}
```
leszko, line 170: Is this tag matching only to make sure that this is not executed for `master` builds and pull requests? If yes, then we don't we want to execute this step for `master` builds and pull requests?

leszko (commented):
Added one comment. Other than that, it looks good. Thanks for the PR @hjpotter92

.github/workflows/build.yaml
hjpotter92: oh, completely missed that. will update the pattern to work with master branch as well run the workflow for tag push.

thanks for pointing this out!

leszko: If we want to run it also on `master`, then do we need the `actions-ecosystem/action-regex-match` step at all?

hjpotter92: we don't want it to sign binaries which are generated for all the commits made on an open PR. The jobs would currently run for any open PR as well as newly pushed tags and any push to the master branch:

```yaml
on:
  pull_request:
  push:
    branches:
      - master
    tags:
      - "v*"
```
The regex-match action limits codesigning to only work with `main|master` branch or a semver tag

leszko (approved):

PR #2357, opened 2022-04-08
server: Begin accepting auth header and have it override callback URL values

thomshutt:
- Begin accepting a new `Livepeer-Transcode-Configuration` header in the HTTP Push segment endpoint. 
- If present, this header should have the same format as the response from the "HTTP Auth Callback URL".
- If present and the Manifest ID from the path is **not** referring to an existing Manifest ID, then a call is still made to the callback URL as before (and fails if this call returns an error) but the fields it returns are overridden by the ones in the header
- If present and the Manifest ID in the URL already exists, this field is ignored

- Create a new method for parsing the header and associated unit tests
- Add logic to override response from callback URL in the case outlined above. Unit test for this override behavior.

✅ Wrote new unit tests
✅ TODO: Manual testing

- https://github.com/livepeer/go-livepeer/issues/2332

**Checklist:**

### DIFF
--- a/server/auth.go
+++ b/server/auth.go
@@ -14,6 +14,8 @@ import (
 	"github.com/livepeer/go-livepeer/monitor"
 )
 
+const LIVERPEER_TRANSCODE_CONFIG_HEADER = "Livepeer-Transcode-Configuration"
+
 // Call a webhook URL, passing the request URL we received
 // Based on the response, we can authenticate and confirm whether to accept an incoming stream
 func authenticateStream(authURL *url.URL, incomingRequestURL string) (*authWebhookResponse, error) {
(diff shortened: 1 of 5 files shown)


oscar-davids (approved):
LGTM!

server/auth.go
```diff
@@ -57,3 +59,39 @@ func authenticateStream(authURL *url.URL, incomingRequestURL string) (*authWebho
...
+func (a authWebhookResponse) areProfilesEqual(b authWebhookResponse) bool {
+	// Return quickly in simple cases without trying to marshal JSON
+	if len(a.Profiles) != len(b.Profiles) {
+		return false
+	}
+	if len(a.Profiles) == 0 {
+		return true
+	}
+
+	profilesA, err := json.Marshal(a.Profiles)
```
cyberj0g, line 86: Nitpick: is it ok to suppress error here?

cyberj0g (approved):
LGTM!

server/auth.go
thomshutt: Yeah, I can't see any actual situation where this wouldn't marshal but probably good practice to do something with the error anyway

PR #2362, opened 2022-04-12
eth, pm: Use previous round L1 block for ticket expiration if the current block hash is zero

### LINKED ISSUE #2360: Backdate tickets by one round if the block hash for the current round is 0

Related to https://github.com/livepeer/go-livepeer/issues/2358

O sets the creationRound and creationRoundBlockHash to use for tickets here [1]. At the moment, the values used here are the last initialized round and the block hash for the last initialized round. 

[1] https://github.com/livepeer/go-livepeer/blob/786769b7dc8b2b3a0426090ee68f369df1fda473/pm/recipient.go#L205

O could check if the block hash stored for the last initialized round is 0. If so, then O could "backdate" the ticket by using last initialized round - 1 for the creationRound and the block hash for the last initialized round - 1 for the creationRoundBlockHash. 

The downside of this approach is that the validity period for a ticket becomes 1 round (basically the ticket is only valid until the start of the next round) instead of 2 since the ticket would be set to expire at last initialized round + 1 (last initialized round - 1 + 2 where 2 is the validity period for a ticket) and this also only helps if the block hash of the last initialized round - 1 is not zero - if you had two rounds in a row with zero block hashes this also breaks. The former case would only apply for the scenario where the block hash for the current round is 0 which we expect to be rare going forward and this change would primarily be to provide better behavior in the edge case where the block hash for the current round is 0 for some reason. Also worth noting that typically winning tickets are redeemed in the round that they are received.

**Additional context**

leszko:

Arbitrum sometimes returns `0x00` as the L1 Block Hash. To prevent a situation when all Os cannot receive tickets for the whole round, we can use the previous round as the ticket expiration parameter.

- 
- 
- 

- unit tests
- local geth with `streamflow` contracts

fix #2360

**Checklist:**

### DIFF
--- a/eth/watchers/timewatcher.go
+++ b/eth/watchers/timewatcher.go
@@ -26,12 +26,13 @@ import (
 //	* Last Seen Block Number
 type TimeWatcher struct {
 	// state
-	mu                         sync.RWMutex
-	lastInitializedRound       *big.Int
-	lastInitializedL1BlockHash [32]byte
-	currentRoundStartL1Block   *big.Int
-	transcoderPoolSize         *big.Int
-	lastSeenL1Block            *big.Int
+	mu                            sync.RWMutex
+	lastInitializedRound          *big.Int
+	lastInitializedL1BlockHash    [32]byte
+	preLastInitializedL1BlockHash [32]byte
+	currentRoundStartL1Block      *big.Int
+	transcoderPoolSize            *big.Int
+	lastSeenL1Block               *big.Int
 
 	// last initialized round number subscription feeds
 	roundSubFeed  event.Feed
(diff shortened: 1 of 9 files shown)


eth/watchers/timewatcher.go
```diff
@@ -70,11 +71,15 @@ func NewTimeWatcher(roundsManagerAddr ethcommon.Address, watcher BlockWatcher, l
 	if err != nil {
 		return nil, fmt.Errorf("error fetching initial lastInitializedL1BlockHash value err=%q", err)
 	}
+	pbh, err := tw.lpEth.BlockHashForRound(prevRound(lr))
+	if err != nil {
+		return nil, fmt.Errorf("error fetching initial prelastInitializedL1BlockHash value err=%q", err)
```
yondonfu:
```suggestion
		return nil, fmt.Errorf("error fetching initial preLastInitializedL1BlockHash value err=%q", err)
```

eth/watchers/timewatcher.go
```diff
@@ -102,24 +107,32 @@ func (tw *TimeWatcher) LastInitializedRound() *big.Int {
...
 
-// LastInitializedBlockHash returns the blockhash of the block the last round was initiated in
+// LastInitializedL1BlockHash returns the blockhash of the L1 block the last round was initiated in
 func (tw *TimeWatcher) LastInitializedL1BlockHash() [32]byte {
 	tw.mu.RLock()
 	defer tw.mu.RUnlock()
 	return tw.lastInitializedL1BlockHash
 }
 
+// PreLastInitializedL1BlockHash returns the blockhash of the L1 block the previous to the last round was initiated in
```
yondonfu:
```suggestion
// PreLastInitializedL1BlockHash returns the blockhash of the L1 block for the round proceeding the last initialized round
```

yondonfu (commented):
Looks good after fixing the linter error and addressing a few minor tweaks in comments/error messages.

leszko:
@yondonfu I've incorporated your suggestions, added unit tests, and fixed the linter. PTAL.

yondonfu (approved):
LGTM after squashing!

PR #2369, opened 2022-04-25
orch watcher parallel fetching

### LINKED ISSUE #2327: Parallelize handling events in Orchestrator Watcher
After each new round is started, for each active orchestrator, we cache its stake in the local DB. This operation is performed sequentially for each orchestrator. In the loop we call [cacheOrchestratorStake()](https://github.com/livepeer/go-livepeer/blob/754c1ed203539275332363eb2cd2940a7fa6c72a/eth/watchers/orchestratorwatcher.go#L189).

Currently, for 90 active orchestrator, it takes around 3 minutes to execute, but the time will grow linearly together with the number of active orchestrators. Therefore we need to parallelize the `cacheOrchestratorStake()` execution.

It may be as simple as adding the `go` keyword and removing the [`roundMu`](https://github.com/livepeer/go-livepeer/blob/754c1ed203539275332363eb2cd2940a7fa6c72a/eth/watchers/orchestratorwatcher.go#L175) lock, but it's better to double-check if all operations are thread-safe.

red-0ne:
Fetching orchestrator's stake by `OrchestratorWatcher` is now done in parallel instead of fetching them sequencially.

- Call  `ow.cacheOrchestratorStake()` in a goroutine 

`go test -run ^TestOrchWatcher`

Also running
```
./livepeer -broadcaster -v 6 -network arbitrum-one-rinkeby -ethUrl https://rinkeby.arbitrum.io/rpc
```
and observe `orchestrators` table with sqlite.

Fixes #2327 

**Checklist:**

### DIFF
--- a/eth/watchers/orchestratorwatcher.go
+++ b/eth/watchers/orchestratorwatcher.go
@@ -12,7 +12,6 @@ import (
 	"github.com/livepeer/go-livepeer/eth"
 	"github.com/livepeer/go-livepeer/eth/blockwatch"
 	"github.com/livepeer/go-livepeer/eth/contracts"
-	lpTypes "github.com/livepeer/go-livepeer/eth/types"
 )
 
 const maxFutureRound = int64(math.MaxInt64)
@@ -191,32 +190,25 @@ func (ow *OrchestratorWatcher) handleRoundEvent(log types.Log) error {
 }
 
 func (ow *OrchestratorWatcher) cacheOrchestratorStake(addr ethcommon.Address, round *big.Int) error {
-	var err error
-
-	defer func() {
-		if err != nil {
-			glog.Errorf("could not cache stake update for orchestrator %v and round %v", addr, round)
-		}
-	}()
-
-	var ep *lpTypes.TokenPools
-	ep, err = ow.lpEth.GetTranscoderEarningsPoolForRound(addr, round)
+	ep, err := ow.lpEth.GetTranscoderEarningsPoolForRound(addr, round)
 	if err != nil {
+		glog.Errorf("could not cache stake update for orchestrator %v and round %v", addr, round)
 		return err
 	}
 
-	var stakeFp int64
-	stakeFp, err = common.BaseTokenAmountToFixed(ep.TotalStake)
+	stakeFp, err := common.BaseTokenAmountToFixed(ep.TotalStake)
 	if err != nil {
+		glog.Errorf("could not cache stake update for orchestrator %v and round %v", addr, round)
 		return err
 	}
 
-	if err = ow.store.UpdateOrch(
+	if err := ow.store.UpdateOrch(
 		&common.DBOrch{
 			EthereumAddr: addr.Hex(),
 			Stake:        stakeFp,
 		},
 	); err != nil {
+		glog.Errorf("could not cache stake update for orchestrator %v and round %v", addr, round)
 		return err
 	}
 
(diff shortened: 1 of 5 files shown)


eth/watchers/orchestratorwatcher.go
```diff
@@ -185,27 +186,41 @@ func (ow *OrchestratorWatcher) handleRoundEvent(log types.Log) error {
...
+
 	for _, o := range orchs {
-		if err := ow.cacheOrchestratorStake(ethcommon.HexToAddress(o.EthereumAddr), round); err != nil {
-			glog.Errorf("could not cache stake update for orchestrator %v and round %v", o.EthereumAddr, round)
-		}
+		wg.Add(1)
+		go ow.cacheOrchestratorStake(ethcommon.HexToAddress(o.EthereumAddr), round, wg)
 	}
 
+	wg.Wait()
```
leszko: I wonder if we need this `WaitGroup` at all.

I think we shouldn't need to wait for all `cacheOrchestratorStake()` to complete before returning from `handleRoundEvent`. The function `handleRoundEvent()` is anyway executed asynchronously and calling `cacheOrchestratorStake()` is the last thing it does.

leszko (changes requested):
@red-0ne Thanks for the PR. Added a comment. PTAL.

eth/watchers/orchestratorwatcher.go
```diff
@@ -186,26 +184,34 @@ func (ow *OrchestratorWatcher) handleRoundEvent(log types.Log) error {
...
 	}
 
 	return nil
 }
 
 func (ow *OrchestratorWatcher) cacheOrchestratorStake(addr ethcommon.Address, round *big.Int) error {
-	ep, err := ow.lpEth.GetTranscoderEarningsPoolForRound(addr, round)
+	var err error
+
+	defer func() {
```
leszko: I think you can improve on this error handling:
1. You don't need to return an error from this function if it's not used.

2. I'd avoid using `defer func() {}` just for logging the error. Especially that depending on the reason for an error you could log something different. For example,

```
	ep, err := ow.lpEth.GetTranscoderEarningsPoolForRound(addr, round)
	if err != nil {
		glog.Errorf("could not cache stake update for orchestrator %v and round %v, getting transcoder earnings pool for round failed err=%q", o.EthereumAddr, round, err)
        return
	}
```

leszko (changes requested):
Added one comment.

Other than that, I think that running `go test` is not enough to test your change. Currently automated test suite doesn't actually test any on-chain interactions (we'll introduce it in [E2E Tests](https://github.com/livepeer/go-livepeer/issues/2297)). So, for now you need to really start the Orchestrator on-chain (local or Arbitrum Testnet) and see if it works (e.g. check the SQLite DB if the entries are there).

eth/watchers/orchestratorwatcher.go
red-0ne: OK, will fully run the node and observe the output. The issue body will be updated once done.

leszko (approved):
The PR looks good to me. 

Before merging, please do the following:
1. Clean up commit history (your PR should contain just one commit)
2. Follow the [commit message naming convention](https://github.com/livepeer/go-livepeer/blob/master/CONTRIBUTING.md#commits) (remember about the `watchers:` prefix)
3. For merging use `Rebase` or `Squash and merge`.
4. Update your PR description:
   - `#2327` => `fix #2327` (then merging the PR will automatically close the related GH Issue)
   - Mark or cross out all the points from the checklist

leszko (approved):

Issue #2323, opened 2022-03-16
Research and create proposal for addressing the session limit problem when fast verification is used


yondonfu:
Fast verification is run at random intervals based on configurable frequency (i.e. 1 out of every 5 segments). As a result, whenever it is run, B will send a source segment to multiple Os, but afterwards will continue sending segments of the stream to just one of those Os. The other Os would have created sessions in order to handle the segment, but afterwords these sessions would be idle and waiting to be cleaned up after a timeout. If this happens consistently, it is possible for session usage on Os to increase and to diverge a lot from the actual resource utilization (i.e. hardware). Many Os would see spikes in sessions that do not reflect actual resource utilization and eventually many Os could hit session limits preventing them from accepting additional jobs.

Some starter ideas:

- Perhaps we could have B signal to O to tear down session instead of waiting for timeout
    - Additional consideration - initialization sessions with Nvidia can be expensive so what is the implication of initializing + tearing down really often if we have Bs signal to Os to tear down sessions
- Reserve Os specifically for fast verification so transcoding capacity doesn’t get impacted by verification capacity
    - Or reserve certain sessions on O for one type of task vs. the other
    - If we do this, prob want clear delineations/boundaries
        - Maybe just default to split O + T, some Ts dedicated to one and other Ts dedicated to the other
- Have some way of differentiating these sessions so that they don’t count / count less towards to limit
- Algorithm for how frequently verification runs takes into account number of Orchestrators and their session capacity (in addition to target frequency interval)

cyberj0g:
**Findings**

From looking through the code and running some tests:
1. *Transcoding* session accounting is entirely on O side. Transcoder capacity is 10 by default, controlled by `maxSessions` parameter and is reported when T connects to O.
2. T initializes HW context (if acceleration is enabled) once transcoding of the first segment starts on `Transcoder` instance (instance meaning the object, not the node). Separate transcoder instance is associated with each HW accelerator. It will not ever destroy HW context, thus, we don't have to worry about `initialization sessions with Nvidia can be expensive`.
3. [This](https://github.com/livepeer/go-livepeer/blob/5d9bcaa325d8caa18151b42f50f4f78bf76834d7/core/orchestrator.go#L938) piece in `RemoteTranscoderManager.selectTranscoder()` (runs on O) will prevent O from sending more sessions to the T, which is overbooked by idle sessions.
[...]

PR #2381, opened 2022-04-27
Session tear down logic for fast verification

cyberj0g:

Implementation of session tear down logic as part of verification process #2323 

- 
- 
- 

**Checklist:**


cyberj0g:
That's still work in progress, adding the tests next. @yondonfu could you please have a quick look and maybe suggest some improvements, as you are the only person available who originally worked on this logic?
CC: @thomshutt @AlexKordic @oscar-davids

core/orchestrator.go
```diff
@@ -958,7 +958,24 @@ func (rtm *RemoteTranscoderManager) selectTranscoder(sessionId string, caps *Cap
 	return nil, ErrNoTranscodersAvailable
 }
 
-// compleStreamSessions end a stream session for a remote transcoder and decrements its laod
+// ends transcoding session and releases resources
+func (rtm *RemoteTranscoderManager) EndSession(sessionId string) {
+	// session may be already terminated by timeout
+	if sess, exists := rtm.streamSessions[sessionId]; exists {
+		// send NotifySegment with empty URL to signal session teardown
+		// xxx get valid auth token?
```
yondonfu: Worth double checking, but...I think the AuthToken.Token field is only validated in [verifySegCreds()](https://github.com/livepeer/go-livepeer/blob/5d9bcaa325d8caa18151b42f50f4f78bf76834d7/server/segment_rpc.go#L399) which is only called by an O for segment requests from B. And in all other cases I think the AuthToken struct is just used to pass around the SessionId field.

server/rpc.go
```diff
@@ -297,6 +331,21 @@ func getOrchestrator(orch Orchestrator, req *net.OrchestratorRequest) (*net.Orch
...
+	addr := ethcommon.BytesToAddress(req.Address)
+	if err := verifyOrchestratorReq(orch, addr, req.Sig); err != nil {
+		return nil, fmt.Errorf("Invalid orchestrator request: %v", err)
+	}
+
+	if _, err := authenticateBroadcaster(addr.Hex()); err != nil {
+		return nil, fmt.Errorf("authentication failed: %v", err)
+	}
+
+	node.TranscoderManager.EndSession(req.SessionId)
```
yondonfu: Seems like we would want to execute all the cleanup logic that we would otherwise run when we hit a timeout in the O's [transcode loop](https://github.com/livepeer/go-livepeer/blob/e3f9286f80044dc1e55a4687eaeccb689a2ebf84/core/orchestrator.go#L638) here as well - perhaps we could extract all of that logic into a shared function that is triggered in the following circumstances:

- If the O's transcode loop times out due to inactivity (we already do this)
- If the O receives an EndSessionRequest from B

server/rpc.go
```diff
@@ -297,6 +331,21 @@ func getOrchestrator(orch Orchestrator, req *net.OrchestratorRequest) (*net.Orch
 	return orchestratorInfo(orch, addr, orch.ServiceURI().String())
 }
 
+func endSession(node *core.LivepeerNode, orch Orchestrator, req *net.EndSessionRequest) (*net.EndSessionResponse, error) {
+	addr := ethcommon.BytesToAddress(req.Address)
```
yondonfu: Instead of validating the B signature here perhaps we can just attach an AuthToken field as is done for the [SegData message](https://github.com/livepeer/go-livepeer/blob/e3f9286f80044dc1e55a4687eaeccb689a2ebf84/net/lp_rpc.proto#L197).

The AuthToken used in the EndSessionRequest should be the same AuthToken that the B was embedding in the SegData messages that was sending to O in segment requests. Then, when O receives the EndSessionRequest it can validate the Token field of the AuthToken message to make sure the request to end the session is authenticated using the AuthToken associated with that session - see [this example](https://github.com/livepeer/go-livepeer/blob/e3f9286f80044dc1e55a4687eaeccb689a2ebf84/server/segment_rpc.go#L399) for verifying the token. And we'd still be able to access the sessionId since it is included in the AuthToken message.

server/broadcast.go
```diff
@@ -908,7 +911,7 @@ func transcodeSegment(ctx context.Context, cxn *rtmpConnection, seg *stream.HLSS
 		res, err = SubmitSegment(ctx, sess.Clone(), seg, nonce, calcPerceptualHash, verified)
 		if err != nil || res == nil {
 			if isNonRetryableError(err) {
-				cxn.sessManager.completeSession(sess)
+				cxn.sessManager.completeSession(ctx, sess, false)
 				return nil, info, err
 			}
 			cxn.sessManager.suspendAndRemoveOrch(sess)
```
yondonfu, line 969: I was considering whether we should send a EndSessionRequest if we encounter an error such that we're suspending and removing the O from selection. On one hand, if we encounter an error here the O might not be in a state where it can process the EndSessionRequest (i.e. if we hit a timeout because it is not responding to requests). On another hand, if there is a chance that the O can handle the EndSessionRequest then that would be "cleaner". But, then there is also the possibility that the EndSessionRequest hangs and we need to make sure to set a proper timeout there...

Just laying out some thinking above...I suspect that it is fine to not send an EndSessionRequest here for now.

server/router.go
```diff
@@ -26,6 +26,10 @@ type Router struct {
 	srv  *grpc.Server
 }
 
+func (r *Router) EndSession(ctx context.Context, request *net.EndSessionRequest) (*net.EndSessionResponse, error) {
+	panic("implement me")
```
yondonfu: I think it could fine for this to be a no-op i.e. just return `&net.EndSessionRequest, nil` because the way the router is supposed to work is:

- B sends GetOrchestrator to router
- Router forwards GetOrchestrator to all available Os and returns the first OrchestratorInfo response to B
- The OrchestratorInfo.Transcoder will be the URL used by the B for subsequent segment requests
- So, a EndSessionRequest from B will go to OrchestratorInfo.Transcoder and not the router

If this is the case, then the router wouldn't need to handle the EndSessionRequest and could likely just no-op.

TBH I don't know if anyone actually uses this as it was experimental, but if the above addresses how to handle this then can punt what we want to do with this router concept...

yondonfu (commented):
For my own understanding:

Is the reason why we don't send an EndSessionRequest to O when we cleanup the session on B due to inactivity [1] because we expect that O would have timed out the session on its end at that point as well?

[1] I think here https://github.com/livepeer/go-livepeer/blob/e3f9286f80044dc1e55a4687eaeccb689a2ebf84/server/mediaserver.go#L879 at least for HTTP push?

cyberj0g:
@yondonfu thanks for comments and insights, all addressed.
> Is the reason why we don't send an EndSessionRequest to O when we cleanup the session on B due to inactivity [1] because we expect that O would have timed out the session on its end at that point as well?

I think it's a good idea to signal tear down there as well.

server/ot_rpc_test.go
```diff
@@ -50,6 +50,10 @@ func (st *stubTranscoder) Transcode(ctx context.Context, md *core.SegTranscoding
 	return testRemoteTranscoderResults, nil
 }
 
+func (st *stubTranscoder) EndTranscodingSession(sessionId string) {
+
+}
+
```
oscar-davids, line 57: Is it empty function?

core/livepeernode_test.go
```diff
@@ -35,6 +35,10 @@ func stubTranscoderWithProfiles(profiles []ffmpeg.VideoProfile) *StubTranscoder
 	return &StubTranscoder{Profiles: profiles}
 }
 
+func (t *StubTranscoder) EndTranscodingSession(sessionId string) {
+
+}
+
```
oscar-davids, line 42: Is it empty function?

cyberj0g: Yes, it's a stub for testing to support the interface.

server/ot_rpc_test.go
cyberj0g: It's a stub for testing - nothing to tear down here.

oscar-davids (approved):
LGTM!

PR #2392, opened 2022-05-10
cmd,common,server: Add LP_EXTEND_TIMEOUTS environment variable

### LINKED ISSUE #144: Disable timeouts in Broadcaster for test streams
We recently had a lot of issues in teststreams, because B timed out either during the discovery process or during the segment upload. While the discovery timeout issue is already addressed, the segment upload timeout issue is still in place. Check related [Discord message](https://discord.com/channels/423160867534929930/956293748340428821/968780640726171658) for details.

We need to tweak Broadcaster to disable such strict timeouts for test streams. We may need to add an additional flag (best if not accessible for users) or some env variable, like `LIVEPEER_DISABLE_TIMEOUTS`.

leszko:

Add env variable `LP_EXTEND_TIMEOUTS` which extends all broadcaster's timeouts to `8s`.

This is a hack to fix Test Streams, so the parameter is not exposed as a flag (or in config file). It's planned to be temporary until the poper solution for Test Streams is implemented: https://github.com/livepeer/stream-tester/issues/145

- 
- 
- 

Added printlns with timeouts and tested on local devnet

fix https://github.com/livepeer/stream-tester/issues/144

**Checklist:**

### DIFF
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -8,9 +8,11 @@ import (
 	"flag"
 	"fmt"
 	"github.com/livepeer/go-livepeer/cmd/livepeer/starter"
+	"github.com/livepeer/go-livepeer/common"
 	"os"
 	"os/signal"
 	"runtime"
+	"strconv"
 	"time"
 
 	"github.com/livepeer/livepeer-data/pkg/mistconnector"
@@ -48,6 +50,7 @@ func main() {
 	}
 
 	vFlag.Value.Set(*verbosity)
+	extendTimeouts()
 
 	cfg = updateNilsForUnsetFlags(cfg)
 
@@ -197,4 +200,16 @@ func updateNilsForUnsetFlags(cfg starter.LivepeerConfig) starter.LivepeerConfig 
 	}
 
 	return res
+}
+
+// extendTimeouts extends transcoding timeouts for the testing purpose.
+// This functionality is intended for Stream Tester to avoid timing out while measuring orchestrator performance.
+func extendTimeouts() {
+	if boolVal, _ := strconv.ParseBool(os.Getenv("LP_EXTEND_TIMEOUTS")); boolVal {
+		// Make all timeouts 8s for the common segment durations
+		common.SegUploadTimeoutMultiplier = 4.0
+		common.SegmentUploadTimeout = 8 * time.Second
+		common.HTTPDialTimeout = 8 * time.Second
+		common.SegHttpPushTimeoutMultiplier = 4.0
+	}
 }
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -10,6 +10,7 @@
 - \#2383 Add E2E Tests for checking Livepeer on-chain interactions (@leszko)
 
 #### Broadcaster
+- \#2392 Add LP_EXTEND_TIMEOUTS env variable to extend timeouts for Stream Tester (@leszko)
 
 #### Orchestrator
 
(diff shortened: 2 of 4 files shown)


yondonfu (approved):
This looks like a reasonable change to me, but going to tag some network video folks as well for a second look and for awareness.

AlexKordic (approved):
LGTM!

AlexKordic:
@leszko Why is longer timeout needed?

leszko:
> @leszko Why is longer timeout needed?

It's only for the stream tester. Here's the context: https://github.com/livepeer/stream-tester/issues/144

thomshutt (approved):

cmd/livepeer/livepeer.go
```diff
@@ -1499,3 +1500,15 @@ func checkOrStoreChainID(dbh *common.DB, chainID *big.Int) error {
 
 	return nil
 }
+
+// extendTimeouts extends transcoding timeouts for the testing purpose.
+// This functionality is intended for Stream Tester to avoid timing out while measuring orchestrator performance.
+func extendTimeouts() {
+	if os.Getenv("LP_EXTEND_TIMEOUTS") == "true" {
```
victorges: **nit:** Can also do this to accept more formats
```suggestion
	if boolVal, _ := strconv.ParseBool(os.Getenv("LP_EXTEND_TIMEOUTS")); boolVal {
```

cmd/livepeer/livepeer.go
```diff
@@ -1499,3 +1500,15 @@ func checkOrStoreChainID(dbh *common.DB, chainID *big.Int) error {
...
+		// Make all timeouts 8s for the common segment durations
+		common.SegUploadTimeoutMultiplier = 4.0
+		common.SegmentUploadTimeout = 8 * time.Second
+		common.HTTPDialTimeout = 8 * time.Second
+		common.SegHttpPushTimeoutMultiplier = 4.0
```
victorges, line 213: Should this value be any different than the default?

victorges (approved):
LGTM, but I'm not sure if I understand the high-level need for this.

Specifically, is there any reason for these test streams to hit the broadcaster timeouts but that not happening for production streams? Does the test streams have anything different than what we see in prod?

The link on https://github.com/livepeer/stream-tester/issues/144 to the discord message seems to be broken (or at least I didn't manage to open it) so I couldn't find the original discussion about this.

cmd/livepeer/livepeer.go
leszko: It's the same, but anyway I think it's good to set it up here, to make sure that it's still ok when the default changes.

cmd/livepeer/livepeer.go
leszko: It makes sense. Thanks!

leszko:
> LGTM, but I'm not sure if I understand the high-level need for this.
> 
> Specifically, is there any reason for these test streams to hit the broadcaster timeouts but that not happening for production streams? Does the test streams have anything different than what we see in prod?

Test Streams are different than the production. In a sense, that even if you're not "good enough" to transcode the data in production, you still want to see your test stream data.

For example, we have Os which have seg transcoding latency > 2s. Then, all Bs requests times out. And now:
- For production => this is a good behavior, we don't want to use this O, because it's too slow; then B needs to fail over to another O
- For test streams => we still do want to see the data for O, just the data will show that the latency is not that good; currently we don't see ANY data for such O, which makes it hard to reason (both for us and for O operator) if there is some problem with test streams or connectivity or it's just that O is slow

victorges:
Makes a lot of sense! Thanks for explaining

Issue #2336, opened 2022-03-23
Spec more granular metrics + data collection for fast verification per orchestrator


yondonfu:
- We should be able to determine the fast verification success/failure rate for each O
- We should be able to determine which step of fast verification that an O failed at
- We should be able to collect data from failure cases for Os

yondonfu:
For the first two - I know we've had trouble with per orchestrator metrics with Prometheus in the past due to Prometheus not liking high cardinality data labels so we likely will need to take that into account.

@victorges @iameli Any thoughts here on the best way to collect per orchestrator metrics these days? Could https://github.com/livepeer/livepeer-data help?

victorges:
So now we are actually using Victoria Metrics instead of Prometheus, which does have support for higher cardinality metrics!

@tqian1 is adding support for per-stream metrics in the go-livepeer code, the only problem with that AFAIK (and correct me if I'm wrong) are the actual metrics exporting libraries themselves like OpenCensus or Prometheus, since once we measure one stream they stay in memory forever and there is no clear way to delete them, so the list of metrics only grows forever. Mist fixes this by having their own implementation of a metrics library which only keeps a metric while a stream is currently active.

As for exporting metrics per orchestrator tho, not necessarily per stream, we should be really fine! Maybe even in Prometheus, since metrics are always broken down by instance anyway with all the pod labels.

It would also be interesting to integrate that data into `livepeer-data` (stream health/analyzer) somehow, but if the purpose is only having visibility of these metrics in internal dashboards it is not _necessary_ to do so, everything should hopefully 'just work' in our Prometheus+Victoria Metrics pipeline already.

figintern:
Thanks for the mention @victorges 

Victor summarizes the problem well - the main limitation for per-stream metrics is the library (OpenCensus) that we use in go-livepeer. It does not have a clean way to register and deregister stream-specific tags like `manifestId` and so the list of exported metrics grows forever - very quickly. However this only affects metrics that we are tagging with stream specific tags because manifestId is a tag that increases indefinitely in cardinality.

I was about to create a new issue in the repo to go over an RFC doc I wrote around this, but I can perhaps I can piggyback on this one if that's alright:

https://www.notion.so/livepeer/Single-Stream-Health-OpenCensus-b402a38f5cf54479a7ad9845c57e5604

[...]

Issue #2371, opened 2022-04-25
Per Orchestrator metrics in Broadcaster


leszko:
**Is your feature request related to a problem? Please describe.**
Currently, Livepeer Broadcaster doesn't export any metrics tagged per Orchestrator. This would be useful, for example, in test streams monitoring.

**Describe the solution you'd like**
Add per Orchestrator metrics recorded by Broadcaster:
- `transcode_overall_latency_seconds_per_orchestrator`
- `segment_transcoded_all_appeared_total_per_orchestrator`
- `segment_transcoded_unprocessed_total_per_orchestrator`

**Describe alternatives you've considered**

**Additional context**

yondonfu:
Would it make sense to add labels (i.e. an orchestrator label) for the existing metrics instead of having separate metrics with the `_per_orchestrator` suffix?

leszko:
> Would it make sense to add labels (i.e. an orchestrator label) for the existing metrics instead of having separate metrics with the `_per_orchestrator` suffix?

I think we could do it that way. I actually tried to followed the pattern we have for `per_stream` metrics. We do have both:
- `success_rate`
- `success_rate_per_stream`

@tqian1 do you see any drawbacks of just adding a label to existing metrics instead of creating new ones? Also do you maybe know why we added new ones for `per_stream`?

red-0ne:
I've added the requested metrics in census.

The approach was to create the metric and insert a tag with key `kOrchestratorURI`. The issue now is that we need to get `uri` from somewhere since it's not passed by `SegmentFullyTranscoded()`. I was not sure if I can get that from `ctx` or need to pass the value all along the function calls.

red-0ne:
According to the implementation of `manifestWithTag` it seems that it's possible to get keys from `ctx`. @tqian1 Do you think it's possible to get orchestrator's uri from there? Also, If I tag that uri in the `transcode_overall_latency_seconds` instead of `..._per_orchestrators` Can I have both functionalities (metrics per and overall orchestrators) by looking up by `kOrchestratorURI` key?

figintern:
> @tqian1 do you see any drawbacks of just adding a label to existing metrics instead of creating new ones? Also do you maybe know why we added new ones for `per_stream`?

Sorry for delay here folks:
Adding a label to existing metrics is preferable to creating new ones

[...]

PR #2397, opened 2022-05-11
get matrics for fast verification per orchestrator

oscar-davids:

Issue #2336
We should be able to determine the fast verification success/failure rate for each O

- 
- 
- 

**Checklist:**

### DIFF
--- a/server/broadcast.go
+++ b/server/broadcast.go
@@ -525,7 +525,7 @@ func (bsm *BroadcastSessionsManager) chooseResults(ctx context.Context, submitRe
 		}
 		equal, err := ffmpeg.CompareSignatureByBuffer(trustedHash, untrustedHash)
 		if monitor.Enabled {
-			monitor.FastVerificationDone(ctx)
+			monitor.FastVerificationDone(ctx, untrustedResult.Session.Address())
 		}
 		if err != nil {
 			clog.Errorf(ctx, "error comparing perceptual hashes from url=%s err=%q",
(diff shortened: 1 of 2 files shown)


server/broadcast.go
```diff
@@ -563,7 +564,7 @@ func (bsm *BroadcastSessionsManager) chooseResults(ctx context.Context, submitRe
 		} else {
 			sessionsToSuspend = append(sessionsToSuspend, untrustedResult.Session)
 			if monitor.Enabled {
-				monitor.FastVerificationFailed(ctx)
+				monitor.FastVerificationFailed(ctx, ouri)
```
thomshutt, line 567: How does this URI look? Do you have an example of the metric that's emitted?

Just thinking about if we want to normalize it into something a little less verbose than a full URI

oscar-davids: > How does this URI look? Do you have an example of the metric that's emitted?

I referred the following [comment ](https://github.com/livepeer/go-livepeer/issues/2371#issuecomment-1121715955)and [here](https://github.com/livepeer/go-livepeer/blob/4d80d57efbb99e57317d9ce5722101def44ef48e/monitor/census.go#L1135).

> Just thinking about if we want to normalize it into something a little less verbose than a full URI

I think we have to consider @leszko's opinion [here](https://github.com/livepeer/go-livepeer/issues/2371#issuecomment-1124671741).

thomshutt (approved):

PR #2401, opened 2022-05-12
Fix crash on GPU detection

cyberj0g:

A fix for *crash* on GPU detection on startup. 

**Important:** the root cause is missing [pci.ids](https://man7.org/linux/man-pages/man5/pci.ids.5.html) file on the docker image. Normally, it present on Linux OS, but it doesn't seem to be the case for some docker images. Before, we either used other parent image, or it was [downloaded automatically](https://github.com/jaypipes/pcidb/issues/28), which is not a good approach, in my opinion. I think, we need to add `apt install pciutils` explicitly when creating the images, but I'm not sure where to put it @iameli @yondonfu. Without that, `-nvidia all` command line option won't work on new docker images.

Error message:
```
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0xafbd16]

goroutine 27 [running]:
github.com/livepeer/go-livepeer/common.detectNvidiaDevices()
    /build/common/util.go:440 +0xb6
github.com/livepeer/go-livepeer/common.ParseAccelDevices({0x7ffe42096df4, 0x1d}, 0x1)
    /build/common/util.go:460 +0x38
main.StartLivepeer({, }, {0xc00028dfa0, 0xc00028dfb0, 0xc00028dfc0, 0xc00028dfd0, 0xc00028dfe0, 0xc00028dff0, 0xc00029c000, 0xc00029c0c0, ...})
    /build/cmd/livepeer/livepeer.go:569 +0xc65
main.main.func1()
    /build/cmd/livepeer/livepeer.go:191 +0x58
created by main.main
    /build/cmd/livepeer/livepeer.go:190 +0x607 
```

- 
- 
- 

**Checklist:**

### DIFF
--- a/docker/Dockerfile.build-linux
+++ b/docker/Dockerfile.build-linux
@@ -13,7 +13,7 @@ RUN apt-get update \
   && apt-key adv --keyserver keyserver.ubuntu.com --recv 15CF4D18AF4F7421 \
   && add-apt-repository "deb [arch=amd64] http://apt.llvm.org/xenial/ llvm-toolchain-xenial-8 main" \
   && apt-get update \
-  && apt-get -y install clang-8 clang-tools-8 build-essential pkg-config autoconf sudo git python docker-ce-cli
+  && apt-get -y install clang-8 clang-tools-8 build-essential pkg-config autoconf sudo git python docker-ce-cli pciutils
 
 RUN update-alternatives --install /usr/bin/clang++ clang++ /usr/bin/clang++-8 30 \
   && update-alternatives --install /usr/bin/clang clang /usr/bin/clang-8 30
--- a/common/util.go
+++ b/common/util.go
@@ -437,7 +437,7 @@ func detectNvidiaDevices() ([]string, error) {
 	nvidiaCardCount := 0
 	re := regexp.MustCompile("(?i)nvidia") // case insensitive match
 	for _, card := range gpu.GraphicsCards {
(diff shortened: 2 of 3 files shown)


docker/Dockerfile.build-linux
```diff
@@ -13,7 +13,8 @@ RUN apt-get update \
   && apt-key adv --keyserver keyserver.ubuntu.com --recv 15CF4D18AF4F7421 \
   && add-apt-repository "deb [arch=amd64] http://apt.llvm.org/xenial/ llvm-toolchain-xenial-8 main" \
   && apt-get update \
-  && apt-get -y install clang-8 clang-tools-8 build-essential pkg-config autoconf sudo git python docker-ce-cli
+  && apt-get -y install clang-8 clang-tools-8 build-essential pkg-config autoconf sudo git python docker-ce-cli \
+  && apt-get -y install pciutils
```
hjpotter92: you can just add `pciutils` at the end of previous line :) also, seems like `sudo` is unnecessary here. everything is running as root.

thomshutt: 👍 Condensed onto one line, will leave the `sudo` thing just to keep this patch simple

yondonfu (approved):
LGTM after testing that the new install in the Docker image addresses the missing pci-ids file problem.

Also, recommend during the merge to fix up the commit messages to:

`common: ...` for the first one.

`docker: ...` for the following ones.

PR #2409, opened 2022-05-18
Add E2E test to GH Action

### LINKED ISSUE #2318: Operationalize E2E Tests (test script, GH action)
We need to integrate E2E tests in our CI, so that they run with every PR and push to master.

For this purpose:
- create `test_e2e.sh` script similar to [test.sh](https://github.com/livepeer/go-livepeer/blob/master/test.sh) which executes only E2E tests
- exclude E2E tests from `test.sh`
- Add `test_e2e.sh` to [test GH action](https://github.com/livepeer/go-livepeer/blob/master/.github/workflows/test.yaml) or create a separate `test_e2e` GH Action

Note that we need `test_e2e.sh` as a separate script, not run in `test.sh`, because the requirement for running `test_e2e.sh` is to have a running Docker daemon available.

leszko:
fix #2318

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -87,6 +87,7 @@ jobs:
         run: |
           export PKG_CONFIG_PATH=~/compiled/lib/pkgconfig
           ./test.sh
+          ./test_e2e.sh
 
       - name: Upload coverage reports
         uses: codecov/codecov-action@v3


hjpotter92:
wouldn't the `./...` testing defined in `test.sh` cover this?

```sh
go test -coverprofile cover.out ./...
```

leszko:
> wouldn't the `./...` testing defined in `test.sh` cover this?
> 
> ```shell
> go test -coverprofile cover.out ./...
> ```

E2E tests are excluded from `test.sh` with 
```
$(go list ./... | grep -v 'test/e2e')
```

I did it on purpose. E2E test have a different requirements (you need to have (amd64) Docker daemon running).

thomshutt (approved):

thomshutt:
Is it possible to run them in parallel with the regular tests?

leszko:
> Is it possible to run them in parallel with the regular tests?

Do you mean a separate GH Action workflow? Yes, we could do it. But for now I think it's not needed. E2E tests are super short comparing to the usual tests. So, let's wait, when they grow in terms of execution time, we can move them to a separate workflow. Maybe we could be even more granular with tests.

hjpotter92:
> I did it on purpose. E2E test have a different requirements (you need to have (amd64) Docker daemon running).

but the current change doesn't actually setup the docker daemon (not sure if github runners have that by default); and only removes the e2e tests from coverage reports.

To add docker daemon as a dependency here, you might need to declare a service:

```yaml
service:
  docker:
    image: docker:latest
```
the image might be `docker:dind` depending on how the e2e tests function.

leszko:
GH Runners have Docker by default. So no need to add anything. You can check, I actually merged this PR and everything works fine.

PR #2416, opened 2022-05-19
livepeer_cli: use the O's currently registered Service URI as default address

### LINKED ISSUE #2293: When configuring an O with option 13 in the CLI offer the current registered Service URI as the default for "the public host:port of node"
When changing options using the CLI option 13 the last configuration "the public host:port of node" defaults to using the IP address of the node, which requires typing out a domain every time a change is made.  Many orchestrators use a Domain Name instead of their IP address so it would make sense for the default option to use the currently registered Service URI.

Have the default be the currently registered Service URI

emranemran:
When changing config options for the Orchestrator via livepeer_cli, the
public host:port of the node is requested for the configuration to be
applied successfully. This host:port address defaults to using the
ip-address of the node instead of using the global domain name that is
registered when setting up an O node. Use the registered Service URI as
the default option such that the entry can be left blank when
configuring options via the livepeer-cli.

(see above commit desc)

(see above commit desc)

* testing locally using both off-chain and on-chain OT nodes. 

fix \#2293

**Checklist:**

- ~~[ ] All tests in `./test.sh` pass~~
- ~~[ ] README and other documentation updated~~

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -21,6 +21,7 @@
 ### Bug Fixes 🐞
 
 #### CLI
+- \#2416 Use the O's currently registered Service URI as default address (@emranemran)
 
 #### General
 
--- a/cmd/livepeer_cli/wizard_transcoder.go
+++ b/cmd/livepeer_cli/wizard_transcoder.go
@@ -35,13 +35,14 @@ func myHostPort() string {
 	// 	http://whatismyipaddress.com/api
 	// 	http://ipinfo.io/ip
 	ip := strings.TrimSpace(httpGet("https://api.ipify.org/?format=text"))
-	return ip + ":" + defaultRPCPort
+	return "https://" + ip + ":" + defaultRPCPort
 }
 
 func (w *wizard) promptOrchestratorConfig() (float64, float64, int, int, string) {
 	var (
 		blockRewardCut float64
 		feeCut         float64
+		addr           string
 	)
 
 	orch, _, err := w.getOrchestratorInfo()
@@ -66,13 +67,16 @@ func (w *wizard) promptOrchestratorConfig() (float64, float64, int, int, string)
 	fmt.Printf("Enter the price for %d pixels in Wei (required) ", pixelsPerUnit)
 	pricePerUnit := w.readDefaultInt(0)
 
-	addr := myHostPort()
+	if orch.ServiceURI == "" {
+		addr = myHostPort()
+	} else {
+		addr = orch.ServiceURI
+	}
 	fmt.Printf("Enter the public host:port of node (default: %v)", addr)
 	serviceURI := w.readStringAndValidate(func(in string) (string, error) {
 		if "" == in {
 			in = addr
 		}
-		in = "https://" + in
 		uri, err := url.ParseRequestURI(in)
 		if err != nil {
 			return "", err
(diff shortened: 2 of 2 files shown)


emranemran:
Tested locally with both off-chain and on-chain BOT nodes - it's working as expected and addresses the ticket's need.

cmd/livepeer_cli/wizard_transcoder.go
```diff
@@ -66,13 +67,16 @@ func (w *wizard) promptOrchestratorConfig() (float64, float64, int, int, string)
 	fmt.Printf("Enter the price for %d pixels in Wei (required) ", pixelsPerUnit)
 	pricePerUnit := w.readDefaultInt(0)
 
-	addr := myHostPort()
+	if w.offchain {
```
leszko: I think that this `if` does not make sense, because you can never run this function in the offchain mode. Orchestrator activation and setting orchestrator config is only for on-chain.

However, I think you need another condition here.
1. If `orch.ServiceURI` is already set (orchestrator is already registered on-chain), then use `orch.ServiceURI`. 
2. Otherwise, use `myHostPort()`

leszko (changes requested):
Thanks for the PR @emranemran Added one comment. 

Other than that, could you improve the PR description with the following points?
1. Change `#2293` to `fix #2293`. Then, merging this PR will automatically close the related GH Issue
2. Update `PENDING_CHANGELOG.md`
3. Fill the checklist:
  - Run `./test` locally and check it if it works
  - Cross out `README and other documentation updated` if there is README update related to this PR
  - Update `PENDING_CHANGELOG.md` and check `Pending changelog updated`

cmd/livepeer_cli/wizard_transcoder.go
emranemran: Got it, that makes more sense. I was thinking there might be some use cases for off-chain where `myHostPort` needs  is used as-is. Will update PR.

emranemran: Addressed all comments. Also added a condition to not allow orchestrator config settings to be changed if off-chain is detected based on your comments. 

Only outstanding issue was that test.sh fails locally for unrelated reasons - likely related to clang compiler that gets used by default on my Mac M1. The tests executed in this PR all pass so this should be OK to merge. If it's alright, I can follow up with a fix for test.sh failures on Macs as part of a separate ticket. 
```
/opt/homebrew/Cellar/ffmpeg/5.0.1/include/libavformat/avformat.h:2161:41: note: passing argument to parameter 'decoder_ret' here
extras.c:193:13: error: implicit declaration of function 'avfilter_compare_sign_bypath' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
extras.c:193:13: note: did you mean 'lpms_compare_sign_bypath'?
extras.c:191:5: note: 'lpms_compare_sign_bypath' declared here
extras.c:204:13: error: implicit declaration of function 'avfilter_compare_sign_bybuff' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
extras.c:204:13: note: did you mean 'lpms_compare_sign_bybuffer'?
extras.c:202:5: note: 'lpms_compare_sign_bybuffer' declared here
```

CHANGELOG_PENDING.md
```diff
@@ -20,6 +20,7 @@
 ### Bug Fixes 🐞
 
 #### CLI
+- \#2416 Use the O's currently registered Service URI as default address
```
```suggestion
- \#2416 Use the O's currently registered Service URI as default address (@emranemran)
```

cmd/livepeer_cli/wizard_transcoder.go
leszko: If tests pass on GH Actions, that's fine. M1 may fail for some unrelated reasons. I also experience some test failures on M1.

leszko (approved):
Added one more minor suggestion, other than that, LGTM 👍

PR #2418, opened 2022-05-23
added 2 type errors matrices for fast verification 

oscar-davids:

Related to #2336

- We should be able to determine which step of fast verification that an O failed at

**Checklist:**

### DIFF
--- a/monitor/census.go
+++ b/monitor/census.go
@@ -70,12 +70,10 @@ const (
 )
 
 const (
-	//general type fail of fast verification
-	FVGeneralError = iota
 	//mpeg7-sign comparison fail of fast verification
-	FVType1Error
+	FVType1Error = 1
 	//video comparison fail of fast verification
-	FVType2Error
+	FVType2Error = 2
 )
 
 // Enabled true if metrics was enabled in command line
--- a/monitor/census.go
+++ b/monitor/census.go
@@ -69,6 +69,15 @@ const (
 	segTypeRec     = "recorded" // segment in the stream for which recording is enabled
 )
 
+const (
+	//general type fail of fast verification
+	FVGeneralError = iota
+	//mpeg7-sign comparison fail of fast verification
+	FVType1Error
+	//video comparison fail of fast verification
+	FVType2Error
+)
+
 // Enabled true if metrics was enabled in command line
 var Enabled bool
 var PerStreamMetrics bool
@@ -102,6 +111,7 @@ type (
 		kClientIP                     tag.Key
 		kOrchestratorURI              tag.Key
 		kOrchestratorAddress          tag.Key
+		kFVErrorType                  tag.Key
 		mSegmentSourceAppeared        *stats.Int64Measure
 		mSegmentEmerged               *stats.Int64Measure
 		mSegmentEmergedUnprocessed    *stats.Int64Measure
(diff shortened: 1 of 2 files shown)


monitor/census.go
```diff
@@ -787,6 +793,27 @@ func InitCensus(nodeType NodeType, version string) {
 			TagKeys:     baseTags,
 			Aggregation: view.LastValue(),
 		},
+		{
+			Name:        "fast_verification_succeed",
+			Measure:     census.mFastVerificationSucceed,
+			Description: "Number of fast verifications succeed",
+			TagKeys:     append([]tag.Key{census.kOrchestratorURI}, baseTagsWithManifestID...),
```
red-0ne: If we use `baseTagsWithOrchInfo...` here and below, we'll be able to get both Os URI and eth address. Any reason to avoid it?

monitor/census.go
```diff
@@ -1651,3 +1678,24 @@ func FastVerificationFailed(ctx context.Context, uri string) {
 		clog.Errorf(ctx, "Error recording metrics err=%q", err)
 	}
 }
+func FastVerificationSucceed(ctx context.Context, uri string) {
```
red-0ne: Is it possible to get O's address here?

red-0ne (commented):
We should get O's info whenever possible. Other than that, LGTM

monitor/census.go
oscar-davids: Yeah, we can get O's address before call this function.

monitor/census.go
oscar-davids: Good point, Just now, I think we don't need O's eth address for fast verification.

thomshutt:
We already have `FastVerificationDone` and `FastVerificationFailed` - if we have `failure_type` as a tag on that second metric then I think we don't need any of these three new metrics

server/broadcast.go
```diff
@@ -544,26 +544,28 @@ func (bsm *BroadcastSessionsManager) chooseResults(ctx context.Context, submitRe
...
 			vequal, err = ffmpeg.CompareVideoByBuffer(trustedSegm, untrustedSegm)
-			if monitor.Enabled && (!vequal || err != nil) {
-				monitor.FastVerificationType2Failed(ctx, ouri)
-			}
 			if err != nil {
 				clog.Errorf(ctx, "error uri=%s comparing video from url=%s err=%q", ouri,
 					untrustedResult.TranscodeResult.Segments[segmToCheckIndex].Url, err)
+				if monitor.Enabled {
+					monitor.FastVerificationFailed(ctx, ouri, monitor.FVType2Error)
+					monitor.FastVerificationFailed(ctx, ouri, monitor.FVGeneralError)
```
thomshutt: This looks good now, can we just delete the `monitor.FastVerificationFailed(ctx, ouri, monitor.FVGeneralError)` line though? I think we should only ever fire this metric once and it should be tagged with one error type

oscar-davids: I think the number of FVGeneralError equals # of FVType1Error + # of FVType2Error.
Below this line the function returns. like `return nil, nil, err`

thomshutt: Right, so if we want all errors then we just look for `FastVerificationFailed` metrics (any tags) and if we want to get more specific in our queries then we use the tags. Does having the `FVGeneralError` give us anything additional?

oscar-davids: FVGeneralError doesn't give us additional anything so I removed this.

thomshutt (approved):
LGTM!

PR #2430, opened 2022-05-25
Bump up go-ethereum dependency to v1.10.17

leszko:
Bump up the `go-ethereum` version to keep up-to-date.

A small fix in tests was needed, because the implementation of `util.WaitMined()` has changes in terms of the error handling.


eth/transactionManager_test.go
```diff
@@ -35,6 +35,9 @@ func (stm *stubTransactionSenderReader) TransactionByHash(ctx context.Context, t
 }
 
 func (stm *stubTransactionSenderReader) TransactionReceipt(ctx context.Context, txHash common.Hash) (*types.Receipt, error) {
+	if stm.receipt == nil {
+		return nil, ethereum.NotFound
+	}
```
thomshutt, line 40: Could we have this as a separate commit / mentioned in the PR name?

thomshutt:
Also, it'd be good to just add a description to say why we're bumping the dep (i.e is it just to keep up to date, or is there specific functionality we want to pull in?)

leszko:
> Also, it'd be good to just add a description to say why we're bumping the dep (i.e is it just to keep up to date, or is there specific functionality we want to pull in?)

The initial trigger was this: https://github.com/livepeer/go-livepeer/pull/2428

But in general we should keep up to date with all dependencies, because of:
- security fixes
- bug fixes
- to avoid having a big gap between latest versions and what we use (at some point it may just be too difficult to update the dep version)

Do you see any downsides of keeping up-to-date with all dependencies?

eth/transactionManager_test.go
leszko: I'm not sure it should be a separate commit. It's related to the dep version update. Without that, the tests fail. Why do you think it should be a separate commit?

Btw. added a description about it.

thomshutt: Yep that's fine, I was just saying that if it wasn't related to the bump then it should go in a separate commit

thomshutt:
No downsides, it's just useful to have the context if we ever need to revisit this PR and e.g check what we'll lose if we have to back it out

thomshutt (approved):

eth/transactionManager_test.go
leszko: yeah, makes sense. If unrelated, always better to put in a separate commit (or even PR).

Issue #2394, opened 2022-05-10
Productionize Soccer Stream Detection


thomshutt:

cyberj0g:
Done once metrics [PR](https://github.com/livepeer/go-livepeer/pull/2663) is merged.
To have scene classification on production and staging:
1. Run the Bs with `-metricsPerStream` and `-detectContent` flags
2. Pass stream configuration from Studio or API 
      ```
      Livepeer-Transcode-Configuration: {"detection": {"freq": 2, "sampleRate": 10, "sceneClassification": [{"name": "soccer"},{"name":"adult"}]}}'
      ```
3. Update [sample dashboard](https://eu-metrics-monitoring.livepeer.monster/grafana/d/oQZqfBd4k/scene-classification-demo?orgId=1) with query:
```
sort_desc(avg by(manifest_id) (livepeer_segment_scene_class_prob{seg_class_name="soccer"} > 0.5))
```
4. Configure alerts based on above (to be addressed)

cyberj0g:
Scene classification is deployed to RKV region.
PR which adds missing dependencies to docker images: https://github.com/livepeer/go-livepeer/pull/2695
Infra PR: https://github.com/livepeer/livepeer-infra/pull/1134

cyberj0g:
When planning for enabling content detection for all streams processed on GPUs, we must consider increase in video memory consumption. We already researched that, but worth briefly summarizing again here.

#### Without content detection
One transcoding session **without content detection** consumes about **222 Mb** of VRAM, and that's uniform. Therefore, to get the video memory dictated max number of transcoding sessions, we can just divide VRAM amount on per-session consumption. For the 8 Gb card, it gives roughly 36 sessions, which is usually above the transcoding performance bottleneck.

#### With content detection
With content detection enabled, **2400 Mb** of VRAM is allocated for CUDA and CuDNN runtime libraries and then shared among all CUDA sessions. There's no known way of reducing that amount. Each transcoding session additionally include the content detection model itself, and adds **350 Mb** of VRAM. Thus, the 8 Gb card will be able to run **just 16** transcoding sessions, so we may be at risk of getting OOM errors more frequently, if `-maxSessions` parameter doesn't account for that with regards to pod hardware.

cyberj0g:
Ready to go. Requires `-detectContent` on OTs and `-metricsPerStream` on Bs to identify streams in the monitoring dashboard.

PR #2431, opened 2022-05-25
Productionize Soccer Stream Detection

cyberj0g:
Make content detection functionality production-ready.

**Limitations**
Current implementation de facto uses a separate ad-hoc data stream to output results - the webhook. Detection results are not included with the renditions, because the DNN filter is run in a separate output, which doesn't produce any renditions. Model output is included into AVFrame metadata only for the purpose of transferring data back to LPMS. This would work for internal process of unwanted content detection, but the process needs to be re-worked to produce 'output stream' to end user, if required in the future for ML capabilities.

* Add `-detectContent` command line flag to enable content detection for transcoder
* Fix list of T capabilities to include SceneDetection capability, if enabled
* Create correct list of T capabilities in case capabilities testing is disabled

- Stdout echo web server for content detection webhook
```
while true; do {   echo -ne "HTTP/1.0 200 OK\r\nContent-Length: 0"; } | nc -l -p 8081 -q 1 ; done
```
- B
```
livepeer -broadcaster -rtmpAddr :1936 -cliAddr 127.0.0.1:7937 -nvidia all -httpAddr 127.0.0.1:8936 -orchAddr 127.0.0.1:8935 -orchSecret 11111111111 -v 6 -detectionWebhookUrl http://localhost:8081/
```
- OT
```
livepeer -transcoder -orchestrator -serviceAddr 127.0.0.1:8935 -nvidia all -v 7 -detectContent
```
- Stream
```
ffmpeg -re -i bbb.mp4 -c:a copy -c:v copy -f hls -hls_time 2 -headers 'Livepeer-Transcode-Configuration:
                                                                    {"detection": {"freq": 2, "sampleRate": 10, "sceneClassification": [{"name": "soccer"}]}}' http://localhost:8936/live/movie3/
```
- Webhook output
```
{"manifestID":"movie3","seqNo":10,"sceneClassification":[{"name":"soccer","probability":0.00676799938082695}]}
```

#2394 

**Checklist:**

**TBD: more tests for standalone T**


cyberj0g:
Converting this to draft, because discovered the following while testing:
1. Doesn't work in a separate O-T configuration, because detection results are not transmitted between T and O by HTTP
2. Detection webhook, which I expect to be the primary way of detecting unwanted content, [not get called](https://github.com/livepeer/go-livepeer/blob/eb7e829b79b7dd50b6bbd0dcf83fa06f0fd845f3/server/broadcast.go#L992) when verification is enabled.

thomshutt (approved):
Looks good, do we have any metrics around this at the moment? It'd be good when we begin testing to be able to see how many detections we're getting

cyberj0g:
> Looks good, do we have any metrics around this at the moment? It'd be good when we begin testing to be able to see how many detections we're getting

No, there are no metrics there. I expect that classification results will be reported through webhook, and metrics probably should be logged on that end.

thomshutt (approved):

common/util_test.go
```diff
@@ -458,15 +458,6 @@ func TestParseAccelDevices_WrongDriver(t *testing.T) {
 	getPCI = originGetPCI
 }
 
-func TestParseAccelDevices_FailedDetection(t *testing.T) {
```
cyberj0g, line 461: @red-0ne what is the purpose of this test? It always fails, if Nvidia GPU is present.

red-0ne: You are right. I think it should setup its `GetGPU`/`GetCPU`
The idea was to fail when no GPUs or appropriate PCI devices are detected (it fails now if you have a real one)
A minimal resolution would be
```go
	getGPU = func() ([]*gpu.GraphicsCard, error) {
		return []*gpu.GraphicsCard{}, nil
	}
	getPCI = func() ([]*pci.Device, error) {
		return []*pci.Device{}, nil
	}
```

PR #2443, opened 2022-06-01
rl/e2e configure orchestrator

### LINKED ISSUE #2421: E2E Test scenario: Configure Orchestrator
Create an [E2E Test](https://github.com/livepeer/go-livepeer/issues/2297) which covers the following scenario:
1. Start and register Orchestrator
2. Set new Orchestrator config (change some parameters, e.g. reward cut or/and service URI)
3. Check that the config is updated on-chain
4. Unregister Orchestrator
5. Check that the orchestrator is unregistered and that stake is back in the ETH account

**Additional comments**
- This scenario can be implemented as extending the [existing scenario for orchestrator activation](https://github.com/livepeer/go-livepeer/blob/master/test/e2e/register_orchestrator_test.go)
- Some parts (registering orchestrator, performing transcoding) may be shared across multiple E2E Test scenarios
   - refactor them into reusable parts
   - check related issues:
     - #2319 
     - #2420 
     - #2422

red-0ne:
E2E test orchestrator configuration

- Start a geth node
- Start go-livepeer node in orchestrator mode
- Register orchestrator
- Wait for the next round
- Reconfigure orchestrator
- Check updated values

Tests with `./test.sh` and `./test_e2e.sh` pass

Fixes #2421 

**Checklist:**


server/handlers.go
```diff
@@ -593,7 +594,7 @@ func unbondHandler(client eth.LivepeerEthClient) http.Handler {
 			respond500(w, err.Error())
 			return
 		}
-		respondOk(w, nil)
+		respondOk(w, tx.Hash().Bytes())
```
leszko: Why do you need to return tx hash from here? I think that the unbound handler should not need to return anything about the transaction.

server/handlers_test.go
```diff
@@ -17,6 +13,11 @@ import (
 	"strings"
 	"testing"
 
+	ethtypes "github.com/ethereum/go-ethereum/core/types"
```
leszko: This change does not fit this PR I guess.

server/webserver.go
```diff
@@ -3,6 +3,7 @@ package server
 import (
 	"flag"
 	"net/http"
+
```
leszko: No need to add this new line in this PR. It's unrelated.

test/e2e/configure_orchestrator_test.go
```diff
@@ -0,0 +1,129 @@
...
+
+	waitForNextRound(t, o.dev.Client)
+	o.dev.InitializeRound()
+
+	// then
+	assertOrchestratorConfigured(t, o, cfg)
+}
+
+func startAndRegisterOrchestrator(t *testing.T, geth *gethContainer, ctx context.Context) *livepeer {
+	const (
```
leszko: This code looks like a duplication of `register_orchestrator_test.go:startOrchestrator()`. How about doing some refactor and moving the common code to `e2e.go` and reuse?

test/e2e/configure_orchestrator_test.go
```diff
@@ -0,0 +1,129 @@
...
+	// when
+	o := startAndRegisterOrchestrator(t, geth, ctx)
+	defer o.stop()
+
+	waitForNextRound(t, o.dev.Client)
+	o.dev.InitializeRound()
+
+	requireOrchestratorRegisteredAndActivated(t, o.dev.Client)
+
+	claimRewards(o)
```
leszko: I think we should not claim rewards in this test. The scope of this E2E Test should be limited to registering and configuring the Orchestrator.

test/e2e/configure_orchestrator_test.go
```diff
@@ -0,0 +1,129 @@
...
+
+	geth := setupGeth(t)
+	defer terminateGeth(t, geth)
+
+	// when
+	o := startAndRegisterOrchestrator(t, geth, ctx)
+	defer o.stop()
+
+	waitForNextRound(t, o.dev.Client)
+	o.dev.InitializeRound()
```
leszko: I don't think you need to initialize the round here. The round should be automatically initialized by the orchestrator.

test/e2e/configure_orchestrator_test.go
```diff
@@ -0,0 +1,129 @@
...
+		PixelsPerUnit:  12,
+		BlockRewardCut: 25.0,
+		FeeShare:       55.0,
+		ServiceURI:     "127.0.0.1:18545",
+	}
+
+	configureOrchestrator(o, cfg)
+
+	waitForNextRound(t, o.dev.Client)
+	o.dev.InitializeRound()
```
leszko: The same here, no need to initialize the round. It's enough to wait for the next round.

test/e2e/configure_orchestrator_test.go
```diff
@@ -0,0 +1,129 @@
...
+
+type OrchestratorConfig struct {
+	PricePerUnit   int
+	PixelsPerUnit  int
+	BlockRewardCut float64
+	FeeShare       float64
+	ServiceURI     string
+}
+
+func TestConfigureOrchestrator(t *testing.T) {
```
leszko, line 15: I think that this tests includes all the logic from `register_orchestrator_test.go`. Then, I think your test could effectively replace Register Orchestrator Test. How about removing `register_orchestrator_test.go` in your PR? I think it does not bring any value to have both.

test/e2e/remove_orchestrator_test.go
```diff
@@ -0,0 +1,119 @@
...
+	"testing"
+	"time"
+
+	ethcommon "github.com/ethereum/go-ethereum/common"
+
+	"github.com/livepeer/go-livepeer/eth/contracts"
+	"github.com/stretchr/testify/require"
+)
+
+func TestRemoveMOrchestrator(t *testing.T) {
```
leszko:
```suggestion
func TestRemoveOrchestrator(t *testing.T) {
```

typo

test/e2e/remove_orchestrator_test.go
```diff
@@ -0,0 +1,119 @@
...
+	o.dev.InitializeRound()
+
+	lockId := deactivateOrchestrator(o)
+
+	waitForNextRound(t, o.dev.Client)
+	o.dev.InitializeRound()
+
+	requireOrchestratorDeactivated(t, o)
+
+	lock, _ := o.dev.Client.GetDelegatorUnbondingLock(o.dev.Client.Account().Address, lockId)
```
leszko: I think it would be better to use more black-box approach. I think that if you can do this operation from `livepeer_cli`, then you shouldn't need to use `o.dev.Client` here.

test/e2e/remove_orchestrator_test.go
```diff
@@ -0,0 +1,119 @@
...
+					return event.UnbondingLockId
+				}
+			}
+		}
+		time.Sleep(200 * time.Millisecond)
+	}
+}
+
+func assertOrchestratorRemoved(t *testing.T, o *livepeer, oldBalance *big.Int) {
+	require := require.New(t)
```
leszko, line 56: I wonder about this check, shouldn't we just check that the orchestrator is not longer in the trancoder pool? That would mean that the orchestrator is removed. However, you check the balance here, I'm not sure I understand how it's related.

leszko (changes requested):
Thanks for the PR Redouane. I added a few comments. Other than that:

1. Could you update the Checklist in the PR description (either check the box or cross out the whole point if it's not related)
2. There is something weird about why your test takes so much time to execute and why you had to split it into 2 tests.

I tried to do all the required steps manually using devtool and livepeer_cli. And doing them manually took only 1.5 min. So, I guess your test should take no more than 1.5 min. I recored a movie of the steps. Could you please check and try it yourself and then try to see what's wrong with the test you wrote?

https://user-images.githubusercontent.com/2834997/173810080-91c5ec64-04ca-4742-9ea3-0f9dbf72a6a0.mp4

server/handlers.go
red-0ne: I needed the unbonding lock to perform withdrawals. I think there is an endpoint to do that that I may use instead of relying on tx

server/webserver.go
red-0ne: This is probably my linter. Will fix it.

test/e2e/configure_orchestrator_test.go
red-0ne: Sure

red-0ne:
> Could you please check and try it yourself and then try to see what's wrong with the test you wrote?

I think I know what's happening. It's actually when changing `blockRewardCut` and `feeShare` that are on-chain operations that require waiting for rounds. I guess if you try to trigger option `13. Set orchestrator config` after being registered and activated, this will cause the error I faced. Will double check it and update here.

test/e2e/configure_orchestrator_test.go
red-0ne: O registration will be used for most of the e2e tests. I think it makes sense to factor the flow and use it elsewhere (Maybe just stripping the final assertion from it). As for having its own test, I think we should keep it at least with something like
```go
func TestRegisterOrchestrator(t *testing.T) {
  // given
  oConfig := OrchestratorConfig{...}

  // when
  startAndRegisterOrchestrator(params)

  // then
  requireOrchestratorRegisteredAndActivated(oConfig)
```

The `startAndRegisterOrchestrator` will be used on other tests. WDYT?

test/e2e/remove_orchestrator_test.go
red-0ne: There are two stages of o removal. By just unbonding (without withdrawal), the orchestrator is indeed no longer active and removed from the transcoder pool, but is still linked to the LP protocol by having his tokens still locked. For complete resignation, I think O needs to get its stake back to consider full removal.

test/e2e/configure_orchestrator_test.go
red-0ne: That's the thing. In order to reconfigure an orchestrator, we need to either claim rewards or be registered **and** not active. I had issues having O in that state (registered but inactive).
See https://github.com/livepeer/protocol/blob/confluence/contracts/bonding/BondingManager.sol#L423, it has
```go
...
require(isRegisteredTranscoder(msg.sender), "transcoder must be registered");
...
require(
    !isActiveTranscoder(msg.sender) || t.lastRewardRound == currentRound,
    "caller can't be active or must have already called reward for the current round"
);
```

test/e2e/remove_orchestrator_test.go
red-0ne: It's true that the function body does not reflect its name. Beside the requirement of the PR

> Check that the orchestrator is unregistered and that stake is back in the ETH account

If we stick to `func assertOrchestratorRemoved` and want to assert O complete removal, then we should check that O is not present in the delegator's pool and that the unbonding lock entry is processed.

Checking balance should still be there but go to a separate assertion.

WDYT?

test/e2e/configure_orchestrator_test.go
leszko: Interesting, could you check how it's possible that it works correctly via Livepeer CLI? As I tried in the screencast attached, I tried manually and it worked fine. I suggest to try the same steps I did manually.

test/e2e/configure_orchestrator_test.go
leszko: Yeah, that's fine. We can keep `register_orchestrator_test.go`, but please refactor the common code and reuse.

test/e2e/remove_orchestrator_test.go
leszko: > There are two stages of o removal. By just unbonding (without withdrawal), the orchestrator is indeed no longer active and removed from the transcoder pool, but is still linked to the LP protocol by having his tokens still locked. For complete resignation, I think O needs to get its stake back to consider full removal.

Ok, I didn't know about it. I thought that removal just means that O disappears from the transcoder pool.

I'm not sure if we should test the "full removal" or not. I leave it up to you.



> It's true that the function body does not reflect its name. Beside the requirement of the PR
> 
> > Check that the orchestrator is unregistered and that stake is back in the ETH account
> 
> If we stick to `func assertOrchestratorRemoved` and want to assert O complete removal, then we should check that O is not present in the delegator's pool and that the unbonding lock entry is processed.
> 
> Checking balance should still be there but go to a separate assertion.
> 
> WDYT?

Yeah, in that case, I'd do one of the following:
- Rename it to `assertOrchestratorStakeIsZero()` or something similar OR
- Keep the name as it is, but explain in the comment why you check the balance to check if orchestrator is removed and also the meaning of the "full removal"

cmd/livepeer/starter/starter.go
```diff
@@ -4,12 +4,21 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"github.com/golang/glog"
-	"github.com/livepeer/go-livepeer/core"
 	"os"
 	"os/user"
 	"time"
 
+	"github.com/golang/glog"
```
leszko: Nit: if you reorder imports, how about keeping them in some logical groups, like maybe first the standard lib imports, and then all `github.com...` imports. Wdyt?

cmd/livepeer/starter/starter.go
```diff
@@ -1064,8 +1066,9 @@ func StartLivepeer(ctx context.Context, cfg LivepeerConfig) {
 		s.ExposeCurrentManifest = *cfg.CurrentManifest
 	}
 
+	srv := &http.Server{Addr: *cfg.CliAddr}
```
leszko, line 1085: That's a very good change! I think we should shutdown the http server while exiting the livepeer node.

server/handlers.go
```diff
@@ -3,6 +3,14 @@ package server
 import (
 	"encoding/json"
 	"fmt"
+	"math/big"
```
leszko: Nit: I think it does not make sense to change this file in your PR just to reorder the imports.

server/handlers_test.go
```diff
@@ -4,10 +4,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	ethtypes "github.com/ethereum/go-ethereum/core/types"
```
leszko, line 7: nit: the same here.

server/webserver.go
```diff
@@ -14,14 +15,10 @@ var vFlag *glog.Level = flag.Lookup("v").Value.(*glog.Level)
...
-	mux := s.cliWebServerHandlers(bindAddr)
-	srv := &http.Server{
-		Addr:    bindAddr,
-		Handler: mux,
-	}
-
-	glog.Info("CLI server listening on ", bindAddr)
+func (s *LivepeerServer) StartCliWebserver(srv *http.Server) {
+	mux := s.cliWebServerHandlers(srv.Addr)
+	srv.Handler = mux
```
leszko:
```suggestion
	srv.Handler =  s.cliWebServerHandlers(srv.Addr)
```

Nit.

test/e2e/configure_orchestrator_test.go
```diff
@@ -0,0 +1,119 @@
...
+	// given
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	geth := setupGeth(t, ctx)
+	defer terminateGeth(t, geth, ctx)
+
+	firstRunCtx, firstRunCancel := context.WithCancel(context.Background())
```
leszko: The test is good. I would just restructure it slightly for the sake of readability.

I think in the `// given` section we should have something like this:
```go
	// given
	ctx, cancel := context.WithCancel(context.Background())
	defer cancel()

	geth := setupGeth(t, ctx)
	defer terminateGeth(t, geth, ctx)

	oCtx, oCancel := context.WithCancel(ctx)
	o := startOrchestrator(t, geth, oCtx)
	registerOrchestrator(o)
	waitUntilRound(t, o.dev.Client)

	// orchestrator needs to get restarted to become active
	o.stop()
	oCancel()
	cfg := starter.LivepeerConfig{EthAcctAddr: o.cfg.EthAcctAddr, Datadir: o.cfg.Datadir}
	o = startOrchestrator(t, geth, ctx, &cfg)
	defer o.stop()

	waitUntilOrchestratorIsConfigurable(t)
```

The `// when` section should describe what we're testing

```go
	// when
	configureOrchestrator(o)
```

And the assertion.

```go
	// then
	assertOrchestratorConfigured(t, o)
```

test/e2e/configure_orchestrator_test.go
```diff
@@ -0,0 +1,119 @@
...
+	o := startOrchestrator(t, geth, firstRunCtx, nil)
+
+	lpEth := o.dev.Client
+	<-o.ready
+
+	initialCfg := OrchestratorConfig{
```
leszko: Nit: I'd probably move this constant for the sake of readability.

test/e2e/configure_orchestrator_test.go
```diff
@@ -0,0 +1,119 @@
...
+	sameAcct := starter.LivepeerConfig{EthAcctAddr: o.cfg.EthAcctAddr, Datadir: o.cfg.Datadir}
+	o = startOrchestrator(t, geth, ctx, &sameAcct)
+	defer o.stop()
+
+	lpEth = o.dev.Client
+	<-o.ready
+
+	waitUntilOrchestratorIsConfigurable(t, lpEth)
+
+	newCfg := &OrchestratorConfig{
```
leszko: Nit: The same with this part, I'd either move it to constants or to a function.

test/e2e/configure_orchestrator_test.go
```diff
@@ -0,0 +1,119 @@
...
+
+	for {
+		if _, ok := httpPostWithParams(fmt.Sprintf("http://%s/setOrchestratorConfig", *o.cfg.CliAddr), val); ok {
+			return
+		}
+		time.Sleep(200 * time.Millisecond)
+	}
+}
+
+func waitUntilOrchestratorIsConfigurable(t *testing.T, lpEth eth.LivepeerEthClient) {
```
leszko, line 46: Nit: Reorder functions, this one is first used,  so should be before `configureOrchestrator()`

test/e2e/configure_orchestrator_test.go
```diff
@@ -0,0 +1,119 @@
...
+
+		initialized, err := lpEth.CurrentRoundInitialized()
+		require.NoError(err)
+
+		t, err := lpEth.GetTranscoder(lpEth.Account().Address)
+		require.NoError(err)
+		rewardCalled := t.LastRewardRound.Cmp(big.NewInt(0)) > 0
+
+		if active && initialized && rewardCalled {
+			time.Sleep(2 * time.Second)
```
leszko: If the orchestrator is already ok, do we still need to wait `2s`?

test/e2e/e2e.go
```diff
@@ -71,6 +72,15 @@ type livepeer struct {
 	ready chan struct{}
 }
 
+type OrchestratorConfig struct {
```
leszko: Nit: I think it can be package-private, so `orchestratorConfig` instead of `OrchestratorConfig`.

test/e2e/e2e.go
```diff
@@ -144,6 +163,50 @@ func startLivepeer(t *testing.T, lpCfg starter.LivepeerConfig, geth *gethContain
...
+	require.Len(transPool, 1)
+	trans := transPool[0]
+	require.True(trans.Active)
+	require.Equal("Registered", trans.Status)
+	require.Equal(big.NewInt(params.LptStake), trans.DelegatedStake)
+	require.Equal(eth.FromPerc(params.FeeShare), trans.FeeShare)
+	require.Equal(eth.FromPerc(params.BlockRewardCut), trans.RewardCut)
+}
+
+func startOrchestrator(t *testing.T, geth *gethContainer, ctx context.Context, cfg *starter.LivepeerConfig) *livepeer {
```
leszko: I think this last parameter you added (`cfg`) is confusing, because someone may think you use this config for the Livepeer node, but actually you just use it to override two fields `EthAcctAddr` and `Datadir`. I suggest to create a separate function `restartOrchestrator()` which will take `EthAcctAddr` and `Datadir` as params.

test/e2e/e2e.go
```diff
@@ -100,31 +110,40 @@ func lpCfg() starter.LivepeerConfig {
 	return cfg
 }
 
-func startLivepeer(t *testing.T, lpCfg starter.LivepeerConfig, geth *gethContainer) *livepeer {
-	datadir := t.TempDir()
-	keystoreDir := filepath.Join(datadir, "keystore")
-	acc := devtool.CreateKey(keystoreDir)
+func startLivepeer(t *testing.T, lpCfg starter.LivepeerConfig, geth *gethContainer, ctx context.Context) *livepeer {
```
leszko, line 129: `ctx` should be usually be the first param in the function. In this case, `t` also should be first, so I suggest the following parameter order `startLivepeer(t *testing.T, ctx context.Context, lpCfg starter.LivepeerConfig, geth *gethContainer)`

test/e2e/remove_orchestrator_test.go
```diff
@@ -0,0 +1,126 @@
...
+	waitUntilRound(lock.WithdrawRound, lpEth, t)
+	// round is not initialized when o is not in the next active set, so we do it here
+	lpEth.InitializeRound()
+	waitUntilRoundInitialized(t, lpEth)
+
+	withdrawStake(o, big.NewInt(lock.ID))
+
+	// then
+	assertOrchestratorRemoved(t, o, big.NewInt(lock.ID))
+	assertStakeWithdrawn(t, o, balance)
```
leszko: I think we can do what you suggested in the Discord comment.

> Regarding full withdrawal of  funds, I kept the command and the assertion for review and tests to pass now, but think we should disable/comment it. I takes a dozen of minutes to finish and impacts any CI push.

I think that if it impacts significantly the test execution time, then we can remove the withdraw part at all.

test/e2e/remove_orchestrator_test.go
```diff
@@ -0,0 +1,126 @@
...
+		PixelsPerUnit:  10,
+		BlockRewardCut: 30.0,
+		FeeShare:       50.0,
+		LptStake:       50,
+	}
+
+	balance, _ := o.dev.Client.BalanceOf(o.dev.Client.Account().Address)
+
+	// when
+	registerOrchestrator(o, &initialCfg)
```
leszko: You can try to restructure a little bit the test to make it more readable. Some suggestions to the points I made in the other test. In particular, the `// when` section should contain what we really want to test in this particular test.

leszko (changes requested):
Thanks @red-0ne for all the work. Added some more comments. They are mainly about:

1. Removing the time-consuming withdraw stake part
2. Some restructure of the tests for the sake of readability

Other than that, the tests are fine. I'm happy we have them :)

cmd/livepeer/starter/starter.go
red-0ne: It's the linter that seems to reorder them when saving. I should disable that rule

CHANGELOG_PENDING.md
```diff
@@ -7,7 +7,9 @@
 ### Features ⚒
 
 #### General
+<<<<<<< HEAD
 - \#1333 Display git-sha in startup logging (@emranemran)
```
```suggestion
- \#1333 Display git-sha in startup logging (@emranemran)
```

Leftover

leszko (approved):
Added one comment. Other than that LGTM 👍 

Thanks for the good work @red-0ne

PR #2445, opened 2022-06-03
livepeer_cli: Fix the "ERROR: logging before flag.Parse" warning when any option is selected

### LINKED ISSUE #2186: livepeer_cli for Broadcaster: "Error getting status" on startup and when calling "Get node status"
## Summary

When using the `livepeer_cli` to configure a Broadcaster, the following message is displayed in the cli UI on a) startup, and b) calling `1. Get node status`:

```
ERROR: logging before flag.Parse: E0116 11:52:21.005601    6268 wizard_stats.go:28] Error getting status:json: cannot unmarshal string into Go struct field OrchestratorLocalInfo.OrchestratorPoolInfos.Url of type url.URL
```

## To Reproduce

1. Run livepeer in `-broadcaster` mode with the following command (for completeness):
```
./livepeer -broadcaster \
        -network rinkeby \
        -ethUrl https://rinkeby.infura.io/v3/888a...243d \
        -pixelsPerUnit 1
```
2. Observe the aforementioned error message being displayed (see screenshot)
3. Run option `1. Get node status`
4. Observe the aforementioned error message being displayed (see screenshot)

## Expected behavior

I expect the `livepeer_cli` to not display unnecessary error message, in order to give new Broadcasters a feeling that the software is robust and mature for their usage.

## Environment

- Ubuntu 20.04.3 LTS
- livepeer 0.5.25

## Screenshots:

On starting up the cli:

image

On calling `1. Get node status`

image

emranemran:
As part of commit 78789b192 (Issue #928), the call to flag.Parse() was commented out perhaps unintentionally. This call is required prior to using glog calls as noted here:
https://github.com/golang/glog/blob/master/glog.go#L40

- (see above commit desc)

- (see above commit desc)

- Tested locally by invoking livepeer_cli and verifying the erroneous warning is suppressed. 

Fix #2186

**Checklist:**

### DIFF
--- a/cmd/livepeer_cli/livepeer_cli.go
+++ b/cmd/livepeer_cli/livepeer_cli.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"bufio"
+	"flag"
 	"fmt"
 	"math/rand"
 	"net/http"
@@ -58,7 +59,7 @@ func main() {
 		return nil
 	}
 	app.Version = core.LivepeerVersion
-	// flag.Parse()
+	flag.Parse()
 	app.Run(os.Args)
 }
 
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -23,6 +23,7 @@
 
 #### CLI
 - \#2438 Add new menu option to gracefully exit livepeer_cli (@emranemran)
+- \#2186 Fix the "ERROR: logging before flag.Parse" warning when any option is selected (@emranemran)
 
 #### General
 


cmd/livepeer_cli/livepeer_cli.go
```diff
@@ -58,7 +59,7 @@ func main() {
 		return nil
 	}
 	app.Version = core.LivepeerVersion
-	// flag.Parse()
+	flag.Parse()
```
leszko, line 62: I think that we cannot parse flags here, because they need to be parsed in [Urfave CLI framework](https://github.com/urfave/cli).

I think after your change, the flags won't work. Please check if the following works for you.

```
./livepeer_cli -http 7937
```

leszko (changes requested):
I think that the flags stop to work after you change. Please check. If that's the case, then I suggest removing this `// flag.Parse()` and just not use anything flag-raleted before calling `app.Run(os.Args)`.

cmd/livepeer_cli/livepeer_cli.go
emranemran: Yep, you're right. Silly of me to not realize this before opening up for review - will revisit the CLI framework and fix this PR.

Quintendos:
Ask @emranemran to merge this item.

PR #2448, opened 2022-06-07
Websocket streaming proof-of-concept

AlexKordic:

Proof of concept for streaming workflow protocol using websocket as transport mechanism.

This is fully streaming workflow using virtual segment boundaries signaled between frames.

# Exploring frame signing

Workflow:
- For each frame `B` (`WebsocketPublish` struct) signs media bytes and sends this signature to `O/T` node.
- `O` (`TranscodingConnection` struct) receives a MPEG-TS frame with `B` signature
- If signature is bad we fail early.

The frame data is signed and later checked

# Exploring virtual segment boundaries

`MistMockup` decides when to place virtual segment boundary.

New message `VirtualSegmentBoundary` struct is sent when old segment ends and new segment is about to begin.

`EndOfInput` message is sent to signal no more input data.

# Exploring segment signing

Replicating segment signing coupled with virtual segment boundaries.

Workflow:
- For each frame `B` (`WebsocketPublish` struct) calculates `Keccak256` hash, each frame adding to final hash
- `O` (`TranscodingConnection` struct) does same hash calculation on received data
- `VirtualSegmentBoundary` signals end of segment carrying Hash and Signature from `B`
- `O` checks received hash to match the hash of received data
- `O` checks received signature of this hash
- Restart Hashing and signature process for next virtual segment

`StreamingHash` struct is used to calculate hash of transferring frame data.

I am not sure if segment signing is done same as in our production code.

- No changes to existing sources, separate POC test and code

```
cd server
go test . -v -count 1 -tags nvidia -run TestWebsocket_Publish
```

Output:
```
=== RUN   TestWebsocket_Publish
serving HTTP on 0.0.0.0:9090
Mist dial ws://127.0.0.1:9090/wslive/testing/1.ts
ws connected, doing handshake
> Mist sent first chunk 
Transcoder started
> Mist sent 1 chunk size=49752
ffmpeg push 49752
> Mist sent 2 chunk size=49752
ffmpeg push 49752
> Mist sent 3 chunk size=49752
ffmpeg push 49752
> Mist sent 4 chunk size=49752
ffmpeg push 49752
decoder.c:185] No audio stream found in input
< ffmpeg 0 output got 1316
< Mist receiving w569 0
< ffmpeg 1 output got 1880
Mist created output file ./encoded-w569.ts
< Mist storing 1316 bytes
< ffmpeg 2 output got 1128
< Mist receiving w474 1
Mist created output file ./encoded-w474.ts
< Mist storing 1880 bytes
< Mist receiving w356 2
Mist created output file ./encoded-w356.ts
< Mist storing 1128 bytes
< ffmpeg 0 output got 752
< Mist receiving w569 0
< Mist storing 752 bytes

...

--- PASS: TestWebsocket_Publish (2.30s)
PASS
```

**Checklist:**


cyberj0g:
I like it, seem to tick most of the boxes for streaming protocol we discussed.
Few questions:
* does WebSocket guarantee message order? Looks like we rely on connection having correct state, if we are are sending chunk 'header' as a separate message
* can we send multiple rendition streams in parallel?
* we may have requirement to annotate video data (not through frame metadata) at some point e.g. detected faces, how much complexity it would add to control protocol? Does it make sense to use gRPC-like messages with text headers and binary data encapsulated?

AlexKordic:
> does WebSocket guarantee message order? Looks like we rely on connection having correct state, if we are are sending chunk 'header' as a separate message

Websocket guarantees message order. Much like TCP with added message size and kind.

On top of websocket transport, we create streaming protocol with different messages intended for `T`-node and different messages produced by `T`-node.

In code, we ensure message order by encapsulating sending and receiving. 

[Here](https://github.com/livepeer/go-livepeer/blob/ak/websocket_poc/server/poc_websocket_client.go#L51-L90) is `M>>B` send code. WebsocketPublish struct handles all protocol specifics including message reorder, hashing, signing. 

[The B counterpart](https://github.com/livepeer/go-livepeer/blob/ak/websocket_poc/server/poc_websocket_ingest.go#L104) handles transport protocol message ordering and [Connection](https://github.com/livepeer/go-livepeer/blob/ak/websocket_poc/server/poc_websocket_connection.go#L117) takes action on received messages without interfering with transport layer.

All abstraction layers have same concept of broken connection.

Websocket transport also has ping-pong mechanism to detect broken channel before TCP timeout.

----

> can we send multiple rendition streams in parallel?

Yes we do it right now. The test `MistMockup` produces `encoded-w356.ts`, `encoded-w474.ts`, `encoded-w569.ts` files from received frames.

If we have `x` number of renditions, we spawn `x` number of [output goroutines](https://github.com/livepeer/go-livepeer/blob/ak/websocket_poc/server/poc_websocket_connection.go#L195-L199). Each blocks reading from ffmpeg pipe. Then [sends output frame](https://github.com/livepeer/go-livepeer/blob/ak/websocket_poc/server/poc_websocket_connection.go#L109) back upstream.

The [sendTranscodedChunk](https://github.com/livepeer/go-livepeer/blob/ak/websocket_poc/server/poc_websocket_connection.go#L65) method makes sure messages are sent sequentially.

If the pipe splits output in several smaller chunks, it would be minimal overhead on the wire.

----

> we may have requirement to annotate video data (not through frame metadata) at some point e.g. detected faces, how much complexity it would add to control protocol? 

Adding another output stream can happen [here](https://github.com/livepeer/go-livepeer/blob/ak/websocket_poc/server/poc_websocket_connection.go#L200):

```
	jpegOutput := EncodedOutput{ffmpeg.TranscodeOptions{}, 4, "thumbnail_%d_ms.jpg"}
	// start your processing method:
	go c.produceThumbnailsAndSendBackJpegs(jpegOutput, spawned.Signal())
```

Right now all outputs read from the ffmpeg pipe. Our new jpeg stream that is not tied to ffmpeg pipe needs to be handled differently in `EncodedOutputList` struct that operates on all outputs.

> Does it make sense to use gRPC-like messages with text headers and binary data encapsulated?

Yes when target of the message is not same M/B/O node we already have connection to.

If its same Node upstream it is best to introduce new message to send back. Metadata gets JSON coded and binary data is transferred without additional encoding overhead.

PR #2449, opened 2022-06-07
mediaserver: add useful debug info to http response headers

### LINKED ISSUE #1333: Display git tip rev hash in startup logging
It is difficult to determine exactly what version of the code is running from within a docker container:
```
# livepeer --version
Livepeer Node Version: 0.5.3
Architecture: amd64
Operating system: linux
```
It would make life easier for everybody if we include the tip rev hash in both the startup message and the `--version` output.

### LINKED ISSUE #2415: Return HTTP headers with Orchestrator information
This is for debugging purposes - it should be possible to stream an HLS manifest and segments and then view the HTTP response headers for a given segment and see some information about how it was generated.

Suggestions for initial headers:

1. LP-SHA - Git SHA the binary was built with (https://blog.alexellis.io/inject-build-time-vars-golang/)
2. LP-ORCH - Some sort of identified for the Orchestrator

emranemran:
When viewing HLS manifests/segments, it is often useful to view information on how those segments were generated. This commit adds fields such as git-sha to the HTTP response headers.

- Update `LPMSOpts` struct to include a new `VidPlayerDebug` struct (defined in lpms repo) which holds additional parameters. These opts are used to pass useful debugging fields when instantiating a `VidPlayer` (to playback HLS segments). 
- Relevant LPMS repo changes: https://github.com/livepeer/lpms/pull/332

- localhost testing using e2e setup in both off-chain and on-chain (using devtool)
- Chrome devtool inspect tab to verify headers were tagged appropriately in both off-chain and on-chain modes

- Fix \#2415
- Fix \#1333

**Checklist:**
~~[ ] All tests in `./test.sh` pass~~

### DIFF
--- a/server/mediaserver.go
+++ b/server/mediaserver.go
@@ -142,6 +142,10 @@ func NewLivepeerServer(rtmpAddr string, lpNode *core.LivepeerNode, httpIngest bo
 		RtmpDisabled: true,
 		WorkDir:      lpNode.WorkDir,
 		HttpMux:      http.NewServeMux(),
+		HttpDebug:    vidplayer.VidPlayerDebug {
+			DebugEnabled:    true,
+			LivepeerVersion: core.LivepeerVersion,
+		},
 	}
 	switch lpNode.NodeType {
 	case core.BroadcasterNode:
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -7,6 +7,8 @@
 ### Features ⚒
 
 #### General
+- \#2449 Add useful debug info to http response headers (@emranemran)
+- \#1333 Display git-sha in startup logging (@emranemran)
 
 #### Broadcaster
 
--- a/Makefile
+++ b/Makefile
@@ -16,8 +16,10 @@ core/test_segment.go:
 	core/test_segment.sh core/test_segment.go
 
 version=$(shell cat VERSION)
+PRODUCT_RELEASE=$(shell ./print_version.sh)
 
-ldflags := -X github.com/livepeer/go-livepeer/core.LivepeerVersion=$(shell ./print_version.sh)
+ldflags := -X github.com/livepeer/go-livepeer/core.LivepeerVersion=$(PRODUCT_RELEASE)
+
 cgo_cflags :=
 cgo_ldflags :=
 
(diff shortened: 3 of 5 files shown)


server/mediaserver.go
```diff
@@ -137,11 +137,21 @@ type authWebhookResponse struct {
...
 		RtmpDisabled: true,
 		WorkDir:      lpNode.WorkDir,
 		HttpMux:      http.NewServeMux(),
+		HttpDebug:    vidplayer.VidPlayerDebug {
+			DebugEnabled:    true,
+			LivepeerVersion: core.LivepeerVersion,
+			NodeType:        lpNode.NodeType.String(), 
+			NodeEthAddress:  nodeEthAddr, 
+		},
 	}
```
emranemran, line 155: Here's an example of what the headers look like in each config: 

**off-chain:**
Screen Shot 2022-06-07 at 1 05 47 PM

**on-chain (using devtool):**
Screen Shot 2022-06-07 at 1 07 14 PM

Is it OK to list the Node's Eth Address in the response headers? It looks like public info to me but please confirm.

leszko:
@emranemran Could you maybe limit the number of reviewers of this PR to 2 pers?

server/mediaserver.go
```diff
@@ -137,11 +137,21 @@ type authWebhookResponse struct {
 }
 
 func NewLivepeerServer(rtmpAddr string, lpNode *core.LivepeerNode, httpIngest bool, transcodingOptions string) (*LivepeerServer, error) {
+	var nodeEthAddr string = "n/a"
+	if lpNode.Eth != nil {
+		nodeEthAddr = lpNode.Eth.Account().Address.String()
```
emranemran: @RiccardoBiosas: can you help confirm if it's OK to reveal a node's Eth Address in HTTP responses? It's marked as public in these structs so my assumption is yet but would like to confirm.

leszko: ETH Address is public, so there are no security concerns when it comes to exposing it.

emranemran:
In its current state, this change has limited uses and the streaming workflow changes will invalidate the changes here -- we'll follow up after streaming workflow lands to achieve the requested change.

PR #2456, opened 2022-06-10
cli: Show O rather than B options when -redeemer flag set

### LINKED ISSUE #2257: Livepeer_CLI With Redeemer flag shows broadcaster options when it should be orchestrator flags
**Describe the bug**
If you launch livepeer with the ```redeemer``` flag, the options from the livepeer_cli go from the orchestrator flags (shown below)

```
10. Invoke "transfer" (LPT)
11. Invoke "reward"
12. Invoke multi-step "become an orchestrator"
13. Set orchestrator config
14. Set maximum Ethereum gas price
15. Set minimum Ethereum gas price

```

to when you have the flag turned on it is more around broadcaster settings:

```
11. Invoke "deposit broadcasting funds" (ETH)
12. Invoke "unlock broadcasting funds"
13. Invoke "cancel unlock of broadcasting funds"
14. Invoke "withdraw broadcasting funds"
15. Set broadcast config
```

**To Reproduce**
1. modify your config file to include ```redeemer``` or launch with ```-redeemer``` and run livepeer
2. run livepeer_cli and you can see the settings change

**Expected behavior**
Seeing as the redeemer would be redeeming rewards, redeeming tickets etc. You would expect the livepeer_cli to have options angled at an orchestrator.

**Screenshots**
below you can see the full differences with the two (top options are with redeemer flag, bottom is without)
image

 - OS: Linux (ubuntu 20)
 - Version: 0.5.28-31e63092

**Additional context**
The redeemer flag can become crucial to the scaling strategy for orchestrators without opening them up to extended risk to their wallets/keys. Having the settings properly denoted will allow orchestrators to execute commands when needed from a single trusted source.

thomshutt:
The CLI now shows Orchestrator rather than Broadcaster options when go-livepeer is running with the `-redeemer` flag set.

- Add a new /IsRedeemer HTTP endpoint to avoid overloading /IsOrchestrator
- Make CLI option generation check this endpoint as well as /IsOrchestrator

Ran the app with the `-redeemer` flag and the CLI before and after my change

fix #2257

**Checklist:**

### DIFF
--- a/cmd/livepeer_cli/livepeer_cli.go
+++ b/cmd/livepeer_cli/livepeer_cli.go
@@ -120,12 +120,14 @@ func (w *wizard) initializeOptions() []wizardOpt {
 }
 
 func (w *wizard) filterOptions(options []wizardOpt) []wizardOpt {
+	isOrchestratorOrRedeemer := w.orchestrator || w.isRedeemer()
+	println("isOrchestratorOrRedeemer", isOrchestratorOrRedeemer)
 	filtered := make([]wizardOpt, 0, len(options))
 	for _, opt := range options {
 		if opt.testnet && !w.testnet {
 			continue
 		}
-		if !opt.orchestrator && !opt.notOrchestrator || w.orchestrator && opt.orchestrator || !w.orchestrator && opt.notOrchestrator {
+		if !opt.orchestrator && !opt.notOrchestrator || isOrchestratorOrRedeemer && opt.orchestrator || !isOrchestratorOrRedeemer && opt.notOrchestrator {
 			filtered = append(filtered, opt)
 		}
 	}
--- a/cmd/livepeer_cli/livepeer_cli.go
+++ b/cmd/livepeer_cli/livepeer_cli.go
@@ -52,6 +52,7 @@ func main() {
 			in:       bufio.NewReader(os.Stdin),
 		}
 		w.orchestrator = w.isOrchestrator()
+		w.redeemer = w.isRedeemer()
 		w.checkNet()
 		w.run()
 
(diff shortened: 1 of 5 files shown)


cmd/livepeer_cli/livepeer_cli.go
```diff
@@ -120,12 +120,14 @@ func (w *wizard) initializeOptions() []wizardOpt {
 }
 
 func (w *wizard) filterOptions(options []wizardOpt) []wizardOpt {
+	isOrchestratorOrRedeemer := w.orchestrator || w.isRedeemer()
```
leszko: I think that for the sake of consistency, I'd not execute `w.isRedeemer()` (and therefore an HTTP call) here, but rather in the place where `wizard` is initialized.

https://github.com/livepeer/go-livepeer/blob/783bfdb392b114f8e5ddfcc95624ededaea5788d/cmd/livepeer_cli/livepeer_cli.go#L54

Also, an idea, maybe we should just rename `w.orchestrator` => `w.orchestratorOrRedeemer`?

leszko (approved):
Added one comment, other than that LGTM.

Also, just a note on the process. You already closed the related GH Issue #2257. However I think you should:
- Keep the GH Issue open until this PR is merged
- Comment in the PR Description `fix #2257` instead of `#2257` (then the GH Issue will be automatically closed when the PR is merged)

cmd/livepeer_cli/livepeer_cli.go
thomshutt: Thanks @leszko, I hadn't realised about the `fix #commit` flow, will do that next time.

I've moved the `isRedeemer` call as suggested, but for the rename there are some places that `w.orchestrator` is used where I wasn't sure if it made sense to also do it for a redeemer node: https://github.com/livepeer/go-livepeer/blob/master/cmd/livepeer_cli/livepeer_cli.go#L85

If you think it makes sense to include redeemers there, then I'll update to `w.orchestratorOrRedeemer` as suggested

leszko: That's fine what you did. Let's keep it as you coded it! 👍

Issue #2319, opened 2022-03-16
E2E Test scenario: B<>O Streaming workflow


leszko:
Create an [E2E Test](https://github.com/livepeer/go-livepeer/issues/2297) which covers the following scenario:

0. Start and register Orchestrator
1. Start Broadcaster
2. Fund Deposit/Reserve
3. Start a stream
4. Check if the stream was transcoded
5. Check that a ticket was received by Orchestrator
6. Redeem the ticket
7. Check that ticket was redeemed and ETH was paid
8. Withdraw Fees and check that ETH is in the O's acount

**In this work, we should focus on a few aspects:**
- We may either send the real video data or mock the video part. I suggest to send one real video segment via HTTP ingest
- It may be hard (or impossible) to tweak `maxTicketEV` and `ticketEV`, so that we're certain to receive a winning ticket, therefore we may need to set `winProb` directly in the code (white-box testing approach)
- If this scenario becomes too long, we can consider splitting it into a few scenarios (e.g. one for sending stream, one for redeeming ticket, one for withdrawing fees)

yondonfu:
> It may be hard (or impossible) to tweak maxTicketEV and ticketEV, so that we're certain to receive a winning ticket, therefore we may need to set winProb directly in the code (white-box testing approach).

If support for a `-maxFaceValue` flag is added in https://github.com/livepeer/go-livepeer/pull/2290 then we should be able to simulate a winning ticket just using CLI flags on B & O by doing the following:

- Let X be some value less than or equal to B's reserve allocation for O
     - Ex. If B's reserve = 100 and there is 1 O then the reserve allocation for that O is 100 so we'd want X <= 100
- Set `-maxFaceValue X` on O
- Set `-ticketEV X` on O
- Set `-maxTicketEV 0` on B
- Set `-depositMultiplier 1` on B

The first two CLI flags should result in a winProb = 100% because the face value of the ticket would be equal to the EV and the only winProb value that achieves that is 100%.

The last two CLI flags relax the restrictions on the faceValue and EV acceptable by B (typically those flags allow the B to restrict the values if needed).

leszko:
@yondonfu Since we're not currently making any changes to that logic, I think it's less of a priority right now. What do you think about moving it to Icebox?

yondonfu:
@leszko Sounds good.

Heads up that many of the items in the OP were covered in the e2e tests added in https://github.com/livepeer/go-livepeer/pull/2655. Still missing 5-8 though.

PR #2463, opened 2022-06-14
e2e tests: added tests to validate e2e streaming workflow between B and O

RiccardoBiosas:
Adds the test cases to validate the e2e streaming workflow between a broadcaster and an orchestrator - from setting up the orchestrator and depositing the required on-chain broadcaster's funds to the probabilistic micropayment workflow.

Updated the files under `tests/e2e`

Run the `test_e2e.sh` script

[](https://github.com/livepeer/go-livepeer/issues/2319)

**Checklist:**

### DIFF
--- a/test/e2e/e2e.go
+++ b/test/e2e/e2e.go
@@ -73,11 +73,11 @@ type livepeer struct {
 
 func lpCfg() starter.LivepeerConfig {
 	mu.Lock()
-	serviceAddr := fmt.Sprintf("127.0.0.1:%d", httpPort)
+	serviceAddr := fmt.Sprintf("localhost:%d", httpPort)
 	httpPort++
-	cliAddr := fmt.Sprintf("127.0.0.1:%d", cliPort)
+	cliAddr := fmt.Sprintf("localhost:%d", cliPort)
 	cliPort++
-	rtmpAddr := fmt.Sprintf("127.0.0.1:%d", rtmpPort)
+	rtmpAddr := fmt.Sprintf("localhost:%d", rtmpPort)
 	rtmpPort++
 	mu.Unlock()
 
@@ -201,4 +201,16 @@ func httpPostWithParamsHeaders(url string, val url.Values, headers map[string]st
 	}
 
 	return string(result), resp.StatusCode >= 200 && resp.StatusCode < 300
+}
+
+func httpGet(url string) string {
+	resp, _ := http.Get(url)
+
+	defer resp.Body.Close()
+	result, err := ioutil.ReadAll(resp.Body)
+	if err != nil || string(result) == "" {
+		return ""
+	}
+	return string(result)
+
 }
--- a/test/e2e/register_orchestrator_test.go
+++ b/test/e2e/register_orchestrator_test.go
@@ -30,8 +30,13 @@ func TestRegisterOrchestrator(t *testing.T) {
 
 func startOrchestrator(t *testing.T, geth *gethContainer) *livepeer {
 	lpCfg := lpCfg()
+	MaxFaceValue := big.NewInt(1000000).String()
+	TicketEV := big.NewInt(1000000).String()
 	lpCfg.Orchestrator = boolPointer(true)
 	lpCfg.Transcoder = boolPointer(true)
+	lpCfg.Redeemer = boolPointer(true)
+	lpCfg.MaxFaceValue = &MaxFaceValue
+	lpCfg.TicketEV = &TicketEV
 	return startLivepeer(t, lpCfg, geth)
 }
 
@@ -49,7 +54,7 @@ func registerOrchestrator(o *livepeer) {
 		"pixelsPerUnit":  {fmt.Sprintf("%d", pixelsPerUnit)},
 		"blockRewardCut": {fmt.Sprintf("%v", rewardCut)},
 		"feeShare":       {fmt.Sprintf("%v", feeShare)},
-		"serviceURI":     {fmt.Sprintf("http://%v", o.cfg.HttpAddr)},
+		"serviceURI":     {fmt.Sprintf("http://%s", *o.cfg.HttpAddr)},
 		"amount":         {fmt.Sprintf("%d", lptStake)},
 	}
 
--- a/test/e2e/streaming_e2e_test.go
+++ b/test/e2e/streaming_e2e_test.go
@@ -1,0 +1,184 @@
+package e2e
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"math/big"
+	"net/http"
+	"net/url"
+	"testing"
+	"time"
+
+	"github.com/livepeer/go-livepeer/eth"
+	lpTypes "github.com/livepeer/go-livepeer/eth/types"
+	"github.com/livepeer/go-livepeer/pm"
+	"github.com/stretchr/testify/require"
+)
+
+func TestE2EStreamingWorkflow(t *testing.T) {
+	geth := setupGeth(t)
+	defer terminateGeth(t, geth)
+
+	// 1. Start and register Orchestrator
+	o := startOrchestrator(t, geth)
+	// 2. Start Broadcaster
+	b := startBroadcaster(t, geth)
+	oEth := o.dev.Client
+	bEth := b.dev.Client
+	defer o.stop()
+	defer b.stop()
+	<-o.ready
+	<-b.ready
+
+	// when
+	// 1. Start and register Orchestrator
+	registerOrchestrator(o)
+	waitForNextRound(t, oEth)
+
+	// then
+	fmt.Println("b.Account() orch ", oEth.Account())
+	fmt.Println("b.Account() broadcaster ", bEth.Account())
+	assertOrchestratorRegisteredAndActivated(t, oEth)
+
+	// when
+	// 3. Fund Deposit/Reserve
+	fundBroadcasterDepositAndReserve(b)
+
+	// then
+	// 4. Check if broadcaster's funds have been correctly deposited
+	assertBroadcasterFundsDeposited(t, b, bEth)
+	// // waitForNextRound(t, bEth)
+
+	// when
+	// 5. Start a stream
+	startStream(b)
+
+	// then
+	// 6. Check if the stream was transcoded
+	assertSuccessfulStreamTranscoding(t, b)
+	waitForNextRound(t, oEth)
+
+	// 6. Check that a ticket was received by Orchestrator
+	// 7. Redeem the ticket
+	// 8. Check that ticket was redeemed and ETH was paid
+	assertEarnings(t, o)
+	// 9. Withdraw Fees and check that ETH is in the O's acount
+	assertWithdrawal(t, o)
+
+}
+
+func startBroadcaster(t *testing.T, geth *gethContainer) *livepeer {
+	lpCfg := lpCfg()
+	lpCfg.Broadcaster = boolPointer(true)
+	CurrentManifest := true
+	DepositMultiplier := 1
+	MaxTicketEV := "0"
+	lpCfg.CurrentManifest = &CurrentManifest
+	lpCfg.DepositMultiplier = &DepositMultiplier
+	lpCfg.MaxTicketEV = &MaxTicketEV
+	return startLivepeer(t, lpCfg, geth)
+}
+
+func fundBroadcasterDepositAndReserve(b *livepeer) {
+
+	val := url.Values{
+		"depositAmount": {fmt.Sprintf("%d", big.NewInt(1000000))},
+		"reserveAmount": {fmt.Sprintf("%d", big.NewInt(1000000))},
+	}
+
+	if _, ok := httpPostWithParams(fmt.Sprintf("http://%s/fundDepositAndReserve", *b.cfg.CliAddr), val); ok {
+		return
+	}
+	time.Sleep(200 * time.Millisecond)
+
+}
+
+func getBroadcasterFunds(urlAddr string, b eth.LivepeerEthClient) pm.SenderInfo {
+	response, _ := http.Get(fmt.Sprintf("http://%s/senderInfo", urlAddr))
+	result, _ := ioutil.ReadAll(response.Body)
+	var broadcasterFunds pm.SenderInfo
+	json.Unmarshal(result, &broadcasterFunds)
+	return broadcasterFunds
+
+}
+
+func assertBroadcasterFundsDeposited(t *testing.T, bNode *livepeer, b eth.LivepeerEthClient) {
+	require := require.New(t)
+
+	broadcasterFunds := getBroadcasterFunds(*bNode.cfg.CliAddr, b)
+
+	require.Equal(broadcasterFunds.Deposit, big.NewInt(1000000))
+	require.Equal(broadcasterFunds.Reserve.FundsRemaining, big.NewInt(1000000))
+}
+
+func startStream(b *livepeer) {
+	fakeSegment, _ := ioutil.ReadFile("../../server/test.ts")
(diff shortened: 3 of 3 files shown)


test/e2e/e2e.go
```diff
@@ -73,11 +73,11 @@ type livepeer struct {
 
 func lpCfg() starter.LivepeerConfig {
 	mu.Lock()
-	serviceAddr := fmt.Sprintf("127.0.0.1:%d", httpPort)
+	serviceAddr := fmt.Sprintf("localhost:%d", httpPort)
```
leszko, line 76: Just out of curiosity, why did you change it? Did `127.0.0.1` cause any issues?

test/e2e/e2e.go
```diff
@@ -202,3 +202,15 @@ func httpPostWithParamsHeaders(url string, val url.Values, headers map[string]st
 
 	return string(result), resp.StatusCode >= 200 && resp.StatusCode < 300
 }
+
+func httpGet(url string) string {
```
leszko, line 206: Why do you use this function? If unused, please remove.

test/e2e/register_orchestrator_test.go
```diff
@@ -30,8 +30,13 @@ func TestRegisterOrchestrator(t *testing.T) {
 
 func startOrchestrator(t *testing.T, geth *gethContainer) *livepeer {
```
leszko, line 31: If this function is used both in `register_orchestrator_test.go` and `streaming_e2e_test.go`, then how about moving it to `e2e.go`?

test/e2e/register_orchestrator_test.go
```diff
@@ -30,8 +30,13 @@ func TestRegisterOrchestrator(t *testing.T) {
...
 	lpCfg := lpCfg()
+	MaxFaceValue := big.NewInt(1000000).String()
+	TicketEV := big.NewInt(1000000).String()
 	lpCfg.Orchestrator = boolPointer(true)
 	lpCfg.Transcoder = boolPointer(true)
+	lpCfg.Redeemer = boolPointer(true)
```
leszko, line 37: I think you don't need to set it to `true`. By default O should redeem the received tickets. AFAIK `redeemer=true` is only used when you run a separate Redeemer service.

test/e2e/streaming_e2e_test.go
```diff
@@ -0,0 +1,184 @@
...
+
+	"github.com/livepeer/go-livepeer/eth"
+	lpTypes "github.com/livepeer/go-livepeer/eth/types"
+	"github.com/livepeer/go-livepeer/pm"
+	"github.com/stretchr/testify/require"
+)
+
+func TestE2EStreamingWorkflow(t *testing.T) {
+	geth := setupGeth(t)
+	defer terminateGeth(t, geth)
```
leszko, line 22:
```suggestion
    // given
	geth := setupGeth(t)
	defer terminateGeth(t, geth)
```

test/e2e/streaming_e2e_test.go
```diff
@@ -0,0 +1,184 @@
...
+
+	// 1. Start and register Orchestrator
+	o := startOrchestrator(t, geth)
+	// 2. Start Broadcaster
+	b := startBroadcaster(t, geth)
```
leszko, line 27:
```suggestion
	o := startOrchestrator(t, geth)
	b := startBroadcaster(t, geth)
```

nit: I think you don't need these comments, because the code actually says the same. It's clear that you start both O and B in these lines.

test/e2e/streaming_e2e_test.go
```diff
@@ -0,0 +1,184 @@
...
+	oEth := o.dev.Client
+	bEth := b.dev.Client
+	defer o.stop()
+	defer b.stop()
+	<-o.ready
+	<-b.ready
+
+	// when
+	// 1. Start and register Orchestrator
+	registerOrchestrator(o)
```
leszko, line 37:
```suggestion
	registerOrchestrator(o)
```

nit: the same here, this comment actually is even misleading because it says `Start`, but you're not starting anything here.

test/e2e/streaming_e2e_test.go
```diff
@@ -0,0 +1,184 @@
...
+	waitForNextRound(t, oEth)
+
+	// then
+	fmt.Println("b.Account() orch ", oEth.Account())
+	fmt.Println("b.Account() broadcaster ", bEth.Account())
+	assertOrchestratorRegisteredAndActivated(t, oEth)
```
leszko, line 43:
```suggestion
	assertOrchestratorRegisteredAndActivated(t, oEth)
```

nit: I'd avoid printlns in the test.

test/e2e/streaming_e2e_test.go
```diff
@@ -0,0 +1,184 @@
...
+
+	// when
+	// 3. Fund Deposit/Reserve
+	fundBroadcasterDepositAndReserve(b)
+
+	// then
+	// 4. Check if broadcaster's funds have been correctly deposited
+	assertBroadcasterFundsDeposited(t, b, bEth)
+	// // waitForNextRound(t, bEth)
```
leszko, line 52: Please clean up the comments.

test/e2e/streaming_e2e_test.go
```diff
@@ -0,0 +1,184 @@
...
+	}
+
+	if _, ok := httpPostWithParams(fmt.Sprintf("http://%s/fundDepositAndReserve", *b.cfg.CliAddr), val); ok {
+		return
+	}
+	time.Sleep(200 * time.Millisecond)
+
+}
+
+func getBroadcasterFunds(urlAddr string, b eth.LivepeerEthClient) pm.SenderInfo {
```
leszko, line 98: 1. I think the `b` parameter is not used. If that's the case, how about removing it?
2. Nit: Consider moving this function under `assertBroadcasterFundsDeposited`. In general, it's easier to read code, when the used function is just under the using one.

test/e2e/streaming_e2e_test.go
```diff
@@ -0,0 +1,184 @@
...
+	require := require.New(t)
+
+	broadcasterFunds := getBroadcasterFunds(*bNode.cfg.CliAddr, b)
+
+	require.Equal(broadcasterFunds.Deposit, big.NewInt(1000000))
+	require.Equal(broadcasterFunds.Reserve.FundsRemaining, big.NewInt(1000000))
+}
+
+func startStream(b *livepeer) {
+	fakeSegment, _ := ioutil.ReadFile("../../server/test.ts")
```
leszko, line 117: Nit: Consider copying this `test.ts` into the `test/e2e` directory. I think E2E Test can have their own testing segments.

test/e2e/streaming_e2e_test.go
```diff
@@ -0,0 +1,184 @@
...
+	reader := bytes.NewReader(fakeSegment)
+
+	response, _ := http.Post(fmt.Sprintf("http://%s/live/name/12.ts", *b.cfg.ServiceAddr), "application/x-www-form-urlencoded", reader)
+
+	defer response.Body.Close()
+
+}
```
leszko, line 124:
```suggestion
	defer response.Body.Close()
}
```

test/e2e/streaming_e2e_test.go
```diff
@@ -0,0 +1,184 @@
...
+
+func assertSuccessfulStreamTranscoding(t *testing.T, b *livepeer) {
+	require := require.New(t)
+	response, _ := http.Get(fmt.Sprintf("http://%s/stream/current.m3u8", *b.cfg.ServiceAddr))
+	defer response.Body.Close()
+	body, _ := ioutil.ReadAll(response.Body)
+
+	require.Equal(200, response.StatusCode)
+	require.Equal("#EXTM3U\n#EXT-X-VERSION:3\n#EXT-X-STREAM-INF:PROGRAM-ID=0,BANDWIDTH=4000000,RESOLUTION=0x0\nname/source.m3u8\n", string(body))
```
leszko, line 133: I wonder, won't this header string never change? If it's possible that this header changes in the future, maybe we could be a little less strict here and check something like this?

```
require.Contains(string(body), "source.m3u8")
```

test/e2e/streaming_e2e_test.go
```diff
@@ -0,0 +1,184 @@
...
+	delegateInfoResponseBefore, _ := http.Get(fmt.Sprintf("http://%s/delegatorInfo", *o.cfg.CliAddr))
+	delegateInfoResponseBody, _ := ioutil.ReadAll(delegateInfoResponseBefore.Body)
+	json.Unmarshal(delegateInfoResponseBody, &delegatorInfoBefore)
+
+	httpPostWithParams(fmt.Sprintf("http://%s/claimEarnings", *o.cfg.CliAddr), nil)
+
+	delegateInfoResponseAfter, _ := http.Get(fmt.Sprintf("http://%s/delegatorInfo", *o.cfg.CliAddr))
+	delegateInfoBodyAfter, _ := ioutil.ReadAll(delegateInfoResponseAfter.Body)
+	json.Unmarshal(delegateInfoBodyAfter, &delegatorInfoAfter)
+	require.GreaterOrEqual(delegatorInfoAfter.Fees.Uint64(), delegatorInfoBefore.Fees.Uint64())
```
leszko, line 151: I think that `GreaterOrEqual` does not really check what we need, because even if O didn't receive any payment, it will pass the test. Wdyt?

test/e2e/streaming_e2e_test.go
```diff
@@ -0,0 +1,184 @@
...
+	delegateInfoResponseAfter, _ := http.Get(fmt.Sprintf("http://%s/delegatorInfo", *o.cfg.CliAddr))
+	delegateInfoBodyAfter, _ := ioutil.ReadAll(delegateInfoResponseAfter.Body)
+	json.Unmarshal(delegateInfoBodyAfter, &delegatorInfoAfter)
+
+	oEthBalanceResponseAfter, _ := http.Get(fmt.Sprintf("http://%s/ethBalance", *o.cfg.CliAddr))
+	oEthBalanceResponseBodyAfter, _ := ioutil.ReadAll(oEthBalanceResponseAfter.Body)
+	json.Unmarshal(oEthBalanceResponseBodyAfter, &oEthBalanceAfter)
+
+	require.Equal(delegatorInfoAfter.Fees, big.NewInt(0))
+	require.Greater(oEthBalanceAfter.Uint64(), oEthBalanceBefore.Uint64())
```
leszko, line 183: How do we ensure that the winning ticket is received? Don't we need to set `maxFaceValue` or `winProb` somewhere? Otherwise, if the ticket is probabilistic, how can we ensure that the winning ticket is received?

leszko (changes requested):
Thanks for the PR @RiccardoBiosas . Added some comments.

Other than that, could you also check the following points:
- Update Description => Add `fix #2319` under `Does this pull request close any open issues?` (then merging your PR will automatically close the related GH Issue)
- Update Description => Mark the checklist (either check the box or cross out the whole point if it's not related)
- Please contact with @red-0ne  how he uses the `context`. Otherwise you're test will fail on `error getting gas price: Post "http://localhost:49960": dial tcp [::1]:49960: connect: connection refused`
- GH Action failed ([link](https://github.com/livepeer/go-livepeer/runs/6880494560)), could you check it?
- `./test_e2e.sh` fails for me locally as well, could you check? Here's the error.

```
--- FAIL: TestE2EStreamingWorkflow (112.98s)
    assertion_compare.go:313: 
                Error Trace:    streaming_e2e_test.go:183
                                                        streaming_e2e_test.go:68
                Error:          "3895020945059707114" is not greater than "3895108342060318893"
                Test:           TestE2EStreamingWorkflow
                Messages:       []
FAIL
FAIL    github.com/livepeer/go-livepeer/test/e2e        158.956s
FAIL
```

test/e2e/register_orchestrator_test.go
red-0ne: It would be interesting in general to move all the partial flows to `e2e.go` / dedicated directory. IMO, this will help reuse these instead of heading to write test-specific ones. Our test-writing-flow should involve looking for and using appropriate partial-flow from the ones already there, **then** if none is suitable, try to add params to close/similar existing ones or; if it's not possible write a new one.

test/e2e/streaming_e2e_test.go
```diff
@@ -0,0 +1,184 @@
...
+	defer b.stop()
+	<-o.ready
+	<-b.ready
+
+	// when
+	// 1. Start and register Orchestrator
+	registerOrchestrator(o)
+	waitForNextRound(t, oEth)
+
+	// then
```
red-0ne, line 40: This seems like a "nested" then, I started using the same pattern but it's kinda tricky to express/signify these nested thens.
I think `//then` should be for the final outcome and intermediary assertions like `assertOrchestratorRegisteredAndActivated` are just intermediary checks used to fail tests early.

```
// given
setupTest() // everything needed to setup the test

// when
doSomeAction1()
requireSomeConditionToBeMet() // prefer requireXXX over assertXXX for non final checks
doSomeAction2()

// then
assertFinalState() // this is the final outcome we are looking for
```

This is only an esthetic/dev experience change that may help better reason and maintain e2e tests. WDYT @RiccardoBiosas, @leszko ?

test/e2e/streaming_e2e_test.go
```diff
@@ -0,0 +1,184 @@
...
+	json.Unmarshal(delegateInfoResponseBody, &delegatorInfoBefore)
+
+	oEthBalanceResponseBefore, _ := http.Get(fmt.Sprintf("http://%s/ethBalance", *o.cfg.CliAddr))
+	oEthBalanceResponseBodyBefore, _ := ioutil.ReadAll(oEthBalanceResponseBefore.Body)
+	json.Unmarshal(oEthBalanceResponseBodyBefore, &oEthBalanceBefore)
+	val := url.Values{
+		"amount": {fmt.Sprintf("%v", delegatorInfoBefore.PendingFees.String())},
+	}
+
+	httpPostWithParams(fmt.Sprintf("http://%s/withdrawFees", *o.cfg.CliAddr), val)
```
red-0ne, line 172: I don't think we should alter state (do post requests) inside assertion functions.
What about
```
withdrawFunds(...)
assertWithdrawalSuccessful(...)
```
I know it's tricky because you want to compare old an new values but we can do something like
```
oldData := getOrchestratorBalanceAndInfo(o)
withdrawFunds(o)
assertWithdrawalSuccessful(o, oldData)
```

Does this makes sense?

test/e2e/streaming_e2e_test.go
red-0ne: > ... Don't we need to set `maxFaceValue` or `winProb` somewhere?

isn't this enough?
`test/e2e/register_orchestrator_test.go#L33`
```go
...
MaxFaceValue := big.NewInt(1000000).String()
...
```

test/e2e/streaming_e2e_test.go
```diff
@@ -0,0 +1,184 @@
...
+func assertEarnings(t *testing.T, o *livepeer) {
+	require := require.New(t)
+	var delegatorInfoBefore lpTypes.Delegator
+	var delegatorInfoAfter lpTypes.Delegator
+
+	delegateInfoResponseBefore, _ := http.Get(fmt.Sprintf("http://%s/delegatorInfo", *o.cfg.CliAddr))
+	delegateInfoResponseBody, _ := ioutil.ReadAll(delegateInfoResponseBefore.Body)
+	json.Unmarshal(delegateInfoResponseBody, &delegatorInfoBefore)
+
+	httpPostWithParams(fmt.Sprintf("http://%s/claimEarnings", *o.cfg.CliAddr), nil)
```
red-0ne, line 146: Same as for withdrawal below, try to avoid changing state inside assertions.

PR #2470, opened 2022-06-21
Use OffchainLabs/go-ethereum fork

leszko:
Arbitrum Nitro requires to use their fork of `go-ethereum` to run Arbitrum Nitro. Please check the related [Discord discussion](https://discord.com/channels/585084330037084172/961349868612386856/988455606841139242).

- Update `go-ethereum` dependency to use `OffchainLabs` fork instead of `ethereum/go-ethereum`.
- 
- 

Run `go-livepeer` in Arbitrum Nitro Devnet.

N/A

**Checklist:**

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -7,6 +7,7 @@
 ### Features ⚒
 
 #### General
+- \#2470 Use OffchainLabs/go-ethereum fork (@leszko)
 
 #### Broadcaster
 - \#2462 cmd: Delete temporary env variable LP_IS_ORCH_TESTER (@leszko)
--- a/go.mod
+++ b/go.mod
@@ -41,3 +41,5 @@ require (
 	google.golang.org/grpc v1.38.0
 	pgregory.net/rapid v0.4.0
 )
+
+replace github.com/ethereum/go-ethereum v1.10.17 => github.com/OffchainLabs/go-ethereum v0.0.0-20220618201649-b37353cfcd4e
--- a/go.sum
+++ b/go.sum
@@ -85,6 +85,8 @@ github.com/Microsoft/hcsshim v0.8.16/go.mod h1:o5/SZqmR7x9JNKsW3pu+nqHm0MF8vbA+V
 github.com/Microsoft/hcsshim/test v0.0.0-20201218223536-d3e5debf77da/go.mod h1:5hlzMzRKMLyo42nCZ9oml8AdTlq/0cvIaBv6tK1RehU=
 github.com/Microsoft/hcsshim/test v0.0.0-20210227013316-43a75bb4edd3/go.mod h1:mw7qgWloBUl75W/gVH3cQszUg1+gUITj7D6NY7ywVnY=
 github.com/NYTimes/gziphandler v0.0.0-20170623195520-56545f4a5d46/go.mod h1:3wb06e3pkSAbeQ52E9H9iFoQsEEwGN64994WTCIhntQ=
+github.com/OffchainLabs/go-ethereum v0.0.0-20220618201649-b37353cfcd4e h1:yWh1bA7XrymN0rJq51Wlo4fQN+VzJK3VPR7RrHyCIMg=
+github.com/OffchainLabs/go-ethereum v0.0.0-20220618201649-b37353cfcd4e/go.mod h1:r2L9GE4f8HY+vHfAJymogfZwkQO+MwRUtf9z9gRQdEM=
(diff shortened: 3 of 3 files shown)


yondonfu:
Looks like this PR is against `master` which means that the go-ethereum dependency update would be reflected in the next release. 

@leszko Could this dependency update have any negative side effects for go-livepeer releases connecting to `arbitrum-one-mainnet` prior to the Nitro upgrade? If we are not sure then I think it would be preferable to not merge the dependency update into `master` and instead test the dependency update on a separate branch.

thomshutt:
Similarly to @yondonfu's comment, do you know if they have any plans to merge their fork back into master? If not, then we'll be reliant on them rebasing off anything that comes into mainline `go-ethereum`

leszko:
If they merge it upstream, that would be the best for us. However, they currently don't know if it's going to be merged upstream. Here's the comment from the [related Discord conversation](https://discord.com/channels/585084330037084172/961349868612386856/988758431857016862):

> here is the relevant PR https://github.com/OffchainLabs/go-ethereum/pull/92 I am not sure it will be merged upstream

Now, what we can do about it?
1. Manually test if it works and merge this PR to `master`
2. Create the `nitro` branch and wait for switching Arbitrum Mainnet to Nitro
    - if they merge to upstream, we don't need to do anything
    - if they don't merge to upstream, we'll merge `nitro` branch into `master` during the Mainnet switch

I'm leaning towards Point 1, because of the following reasons:
- During Mainnet=>Arbitrum, Os will still use the old version of Livepeer and suddenly it will start crashing, so it's better if they have the right version upront
- During Mainnnet=>Arbitrum, we'll have two things changing at the same time 1) Chain switching to Nitro 2) Us changing the `go-ethereum` dep; this may result in hard times debugging what's wrong

Apart from that, Arbitrum uses their fork, so I believe it may be actually even safer to use for the current Mainnet.

Wdyt? @yondonfu @thomshutt

thomshutt:
👍 Makes sense and looks like they are keeping it reasonably up to date vs. the original go-ethereum. Happy for you two to make a call on this

yondonfu:
From glancing at the OffchainLabs/go-ethereum commit history, it looks like the last time it was synced with upstream ethereum/go-ethereum was for the v1.10.17 release https://github.com/OffchainLabs/go-ethereum/commit/7dc8af1913e719ffa94fc1ae9d3a26495ecfbf5f.

From glancing at the ethereum/go-ethereum commit history since the v1.10.17 there are at least a few commits that are associated with the packages we use i.e. `accounts/abi`, `ethclient`:

- https://github.com/ethereum/go-ethereum/commit/af02e97929d47238fa0dbd233846270802b99e1a
- https://github.com/ethereum/go-ethereum/commit/f5ff022dbca2b14af59974154874537b5ed4cc5e
- https://github.com/ethereum/go-ethereum/commit/84041e8f31d310470ada70d86b95458ad1fa46df
- https://github.com/ethereum/go-ethereum/commit/eb69f490ed6758efdd20f995b8d3ead5002911e9
- https://github.com/ethereum/go-ethereum/commit/bf5cacfb8fb28d5894d11c236830007f1bfc04b6

I have not gone through the details of any of the commits to confirm that they include fixes or features that are important for us. However, I think it is worth noting the following:

- There are already commits *not* in the OffchainLabs fork that are relevant to packages we depend on
- There could be future commits relevant to the packages we depend on upstream that will also *not* be in the OffchainLabs fork

Given the above considerations, I could see a scenario where we switch to the OffchainLabs fork, but then realize we want to pull in some of the upstream commits relevant to the packages we depend on. At that point, the best option would likely be to merge upstream/cherry-pick commits into our own fork based off of the OffchainLabs fork.

Since Nitro is still under development and the timeline for Nitro mainnet is still TBD, I am inclined to *not* use the OffchainLabs fork right now to avoid introducing headaches around the above in the short term and instead letting development of Nitro proceed a bit and we can always switch to the OffchainLabs fork at a later point in time (we don't necessarily need to wait until the Nitro upgrade on mainnet - we could just wait until we get a little closer to then) and figure out the right strategy for combining more recent relevant commits from upstream with the required Arbitrum commits in the OffchainLabs fork. I believe this lines up with option 2 suggested by @leszko.

leszko:
Ok, then let's keep this PR (and the branch) open for now and see how it goes with the Arbitrum fork. Maybe they'll merge upstream. Or maybe they'll pull the current `go-ethereum` in.

leszko:
I checked again the current master of go-livepeer and it seems to work correctly on Arbitrum Testnet Nitro. That means that Arbitrum fixed the issue with the incompatible transaction format, and therefore we do not need to merge the OffchainLabs/go-ethereum PR.

Closing this PR.

PR #2472, opened 2022-06-21
livepeer/starter: remove testTranscoder flag

### LINKED ISSUE #2450: running `-nvidia=0` with `-testTranscoder=false` results in an O with no capabilities
Once upon a time you had to run `-testTranscoder` false if you were going to use Nvidia transcoding on Windows. That appears to have been fixed, hooray!

But I did it anyway:

```bash
./livepeer.exe -orchestrator true -transcoder true \
  -network offchain -serviceAddr 192.168.0.49:8935 \
  -nvidia 0 -v 9 -testTranscoder false
```

and it results in a completely useless O: it reports that it has zero capabilities to do anything whatsoever; Bs will always pass it over during the discovery process. No obvious reason for such a thing to be running at all — I guess it could make reward calls or something.

Minimally we should print a log line for this or crash entirely. But perhaps it's time to remove the `testTranscoder` flag entirely if it's not useful anymore?

emranemran:
The -testTranscoder flag defaults to true which requires all hardware
transcoders to call core.TestTranscoderCapabilities() to discover
transcoding caps. Setting this flag to false is not a valid use case
anymore so this change deprecates that option.

- Remove the use of  testTranscoder flag and allow code path to always call TestTranscoderCapabilities() which it already does as the option defaults to false. 

- WIP: to be tested on GCP w/ nvidia hardware...

Fix \#2450

**Checklist:**
~~[ ] All tests in `./test.sh` pass~~

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -20,6 +20,7 @@
 
 - \#2466 bugfix: rendition resolution fix for portrait input videos; Min resolution applied for Nvidia hardware (@AlexKordic)
 - \#338 lpms: Add exception handling code for importing a binary signature (@oscar_davids)
+- \#2472 livepeer/starter: deprecate testTranscoder flag to prevent starting an O with no capabilities (@emahbub)
 
 #### CLI
 - \#2456 cli: Show O rather than B options when -redeemer flag set (@thomshutt)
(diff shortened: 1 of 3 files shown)


thomshutt:
Searching across all repos, it looks like we set it to false in some of our livepeer-in-a-box configs: https://github.com/search?q=org%3Alivepeer+testTranscoder&type=code

@iameli would removing this break things?

AlexKordic:
Tested on nvidia card with:
```
go run cmd/livepeer/livepeer.go --datadir=/home/alex/temp/livepeer/TEST_T1 --transcoder=true --orchSecret=foo --orchAddr="127.0.0.1:18935" --cliAddr=":17936" --nvidia=0 --v=6
```

emranemran:
Perhaps this flag in livepeer-in-a-box is alright for now since it's a fork. I'm guessing we'll have to setup automerges from `master` in go-livepeer to livepeer-in-a-box at some point or just rebase it frequently to pick up changes. @iameli what do you think?

cyberj0g:
I think we can merge this PR, since capability test runs only with hardware acceleration enabled. LGTM after rebasing.

cyberj0g (approved):

Quintendos:
@emranemran could you please close up this ticket?

iameli (approved):
just noticed this in my list lol :+1: LGTM 😆

PR #2481, opened 2022-06-29
pm: lower avgGasPrice

### LINKED ISSUE #2473: Streams drop using low face value
**Describe the bug**
Streams can drop due to face value of ticket being low using the `maxFaceValue` flag.  This only happens when gas prices spike.

This has to do with the code in `recipient.go` below where it uses 200 gwei as the `avgGasPrice` to check if ticket `faceValue` is less than current txCost or txCost with `avgGasPrice`.

https://github.com/livepeer/go-livepeer/blob/00a3e6d81edaf76bca19264d65269e4aa5f8c1c5/pm/recipient.go#L29-L31

This 200 gwei `avgGasPrice` is then used to compare `faceValue` to `txCost` with `avgGasPrice` of 200 gwei with the much higher gas used on Arbitrum (1,200,000 hardcoded, was 350,000 before switch to Arbitrum).  This test will always fail because it calculates a 0.24ETH value to compare the `faceValue` to.  

https://github.com/livepeer/go-livepeer/blob/00a3e6d81edaf76bca19264d65269e4aa5f8c1c5/pm/recipient.go#L304-L306

**To Reproduce**
1. Set `maxFaceValue` to really low value of 0.0012 when gas price is above 1 gwei
2. Observe `insufficient sender reserve` error in logs

**Expected behavior**
`faceValue` check to `txCost` should use an `avgGasPrice` that is a reasonable amount for Arbitrum.

**Additional context**
Ticket face value determined in `recipient.go` can return `insufficient sender reserve` error when ticket face value is set really low and gas prices start to spike.  See table below:

image

leszko:

Lower the hardcoded avg gas price to prevent Os from dropping streams in case of the gas price spikes.

The value is changed from `200` => `140`. Here's the rationale why this value was chosen. In our problematic scenario, in the case when `maxFaceValue` is **not** set, then:

1. In order to not drop the stream during gas spike this condition `faceValue.Cmp(r.txCostWithGasPrice(avgGasPrice)) < 0` must be `false`
2. In the worse case scenario, face value [equals to max float](https://github.com/livepeer/go-livepeer/blob/8b3fcd674859f184407323266b6d1aae0ab42534/pm/recipient.go#L279) (which corresponds to the [B's reserve](https://github.com/livepeer/go-livepeer/blob/5d9bcaa325d8caa18151b42f50f4f78bf76834d7/pm/sendermonitor.go#L191), currently `180000000000000000`)
3. `txCostWithGasPrice()` is calculated as `redeemGas * avgGasPrice`, where Arbitrum `redeemGas` is [hardcoded](url) as `1200000`
4. That means that `180000000000000000 < 1200000 * avgGasPrice` must be `false`
5. That means that `180000000000000000 >= 1200000 * avgGasPrice` must be `true`
6. That means that `avgGasPrice <= 180000000000000000 / 1200000` => `avgGasPrice <= 150000000000`

So, with the current B's reserve and the current hardcoded `redeemGas`, the `avgGasPrice` must be smaller than `150 Gwei`. To make a buffer, I changed it to `140`.

- Change `avgGasPrice`
- Simplify the condition check

Debugged with Orchestrator.

fix #2473

**Checklist:**

### DIFF
--- a/pm/recipient.go
+++ b/pm/recipient.go
@@ -28,7 +28,7 @@ var evMultiplier = big.NewInt(100)
 
 // Hardcode to 200 gwei
 // TODO: Replace this hardcoded value by dynamically determining the average gas price during a period of time
-var avgGasPrice = new(big.Int).Mul(big.NewInt(200), new(big.Int).Exp(big.NewInt(10), big.NewInt(9), nil))
+var avgGasPrice = new(big.Int).Mul(big.NewInt(140), new(big.Int).Exp(big.NewInt(10), big.NewInt(9), nil))
 
 // Recipient is an interface which describes an object capable
 // of receiving tickets
@@ -300,8 +300,7 @@ func (r *recipient) faceValue(sender ethcommon.Address) (*big.Int, error) {
 	// because there is a good chance that the current gasPrice will come back down by the time a winning ticket is received
 	// and needs to be redeemed.
 	// For now, avgGasPrice is hardcoded. See the comment for avgGasPrice for TODO information.
-	// Use || here for lazy evaluation. If the first clause is true we ignore the second clause.
-	if !(faceValue.Cmp(txCost) >= 0 || faceValue.Cmp(r.txCostWithGasPrice(avgGasPrice)) >= 0) {
+	if faceValue.Cmp(txCost) < 0 && faceValue.Cmp(r.txCostWithGasPrice(avgGasPrice)) < 0 {
 		return nil, errInsufficientSenderReserve
 	}
 
--- a/pm/recipient.go
+++ b/pm/recipient.go
@@ -26,9 +26,9 @@ var paramsExpiryBuffer = int64(1)
 
 var evMultiplier = big.NewInt(100)
 
-// Hardcode to 200 gwei
+// Hardcode to 3 gwei
 // TODO: Replace this hardcoded value by dynamically determining the average gas price during a period of time
-var avgGasPrice = new(big.Int).Mul(big.NewInt(140), new(big.Int).Exp(big.NewInt(10), big.NewInt(9), nil))
+var avgGasPrice = new(big.Int).Mul(big.NewInt(3), new(big.Int).Exp(big.NewInt(10), big.NewInt(9), nil))
 
 // Recipient is an interface which describes an object capable
 // of receiving tickets
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -33,6 +33,7 @@
 #### Broadcaster
 
 #### Orchestrator
+- \#2493 cmd: Fix reward flag (@leszko)
 - \#2481 Lower `avgGasPrice` to prevent dropping streams during the gas price spikes (@leszko)
 
 #### Transcoder
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -33,6 +33,6 @@
 #### Broadcaster
 
 #### Orchestrator
-- \#2493 cmd: Fix reward flag (@leszko)
+- \#2481 Lower `avgGasPrice` to prevent dropping streams during the gas price spikes (@leszko)
 
 #### Transcoder


yondonfu:
Just to clarify my own understanding (I wrote this avgGasPrice check awhile ago, but I need to jog my memory around how it works...):

`avgGasPrice` is supposed to represent the average gas price during a period of time. For simplicity, it was originally hardcoded to 200 gwei which indicates that the average gas price is 200 gwei (which was probably accurate at some point on L1). 

The check `faceValue.Cmp(r.txCostWithGasPrice(avgGasPrice)) < 0` is used as a secondary check when `faceValue < txCost` where `txCost` is calculated based on the current estimated gas price. The reason for having this secondary check is that when `faceValue < txCost` it might be the case that we are in the middle of a gas price spike which is why `faceValue < txCost`. If we are in the middle of a gas price spike it would be reasonable to expect the gas price to fall after a period of time. As long as the `faceValue` can cover the expected tx cost based on `avgGasPrice` (recall that this is meant to represent the average gas price) then O advertises ticket params assuming that the gas price will fall to `avgGasPrice` or lower.

With this in mind, it makes sense that a lower `avgGasPrice` makes sense here because the average gas price on Arbitrum is substantially lower than 200 gwei. In reality, the average gas price is even lower than 140 gwei (I know that the OP describes deriving this value based on an assumption of the value of B's reserve, but I generally don't think we should hardcoded values based on an assumption of the value of B's reserve).

I think this PR is fine as-is though as the more ideal solution is to not hardcode a value for `avgGasPrice` at all which can be followed up on.

pm/recipient.go
```diff
@@ -28,7 +28,7 @@ var evMultiplier = big.NewInt(100)
 
 // Hardcode to 200 gwei
```
yondonfu: Let's update this comment

yondonfu (approved):
Minor comment about fixing an outdated comment.

Other than that LGTM.

leszko:
> Just to clarify my own understanding (I wrote this avgGasPrice check awhile ago, but I need to jog my memory around how it works...):
> 
> `avgGasPrice` is supposed to represent the average gas price during a period of time. For simplicity, it was originally hardcoded to 200 gwei which indicates that the average gas price is 200 gwei (which was probably accurate at some point on L1).
> 
> The check `faceValue.Cmp(r.txCostWithGasPrice(avgGasPrice)) < 0` is used as a secondary check when `faceValue < txCost` where `txCost` is calculated based on the current estimated gas price. The reason for having this secondary check is that when `faceValue < txCost` it might be the case that we are in the middle of a gas price spike which is why `faceValue < txCost`. If we are in the middle of a gas price spike it would be reasonable to expect the gas price to fall after a period of time. As long as the `faceValue` can cover the expected tx cost based on `avgGasPrice` (recall that this is meant to represent the average gas price) then O advertises ticket params assuming that the gas price will fall to `avgGasPrice` or lower.
> 
> With this in mind, it makes sense that a lower `avgGasPrice` makes sense here because the average gas price on Arbitrum is substantially lower than 200 gwei. In reality, the average gas price is even lower than 140 gwei (I know that the OP describes deriving this value based on an assumption of the value of B's reserve, but I generally don't think we should hardcoded values based on an assumption of the value of B's reserve).
> 
> I think this PR is fine as-is though as the more ideal solution is to not hardcode a value for `avgGasPrice` at all which can be followed up on.

Yes, everything you described is correct. You may be right that we should maybe not make any assumptions about our B reserve. So how, about setting average gas price to some more realistic number in Arbitrum, like `2` or `3`. Wdyt?

yondonfu:
> So how, about setting average gas price to some more realistic number in Arbitrum, like 2 or 3. Wdyt?

👍

pm/recipient.go
leszko: Updated in 98cad6dc510f138751a9ac9cc889588372ef9ceb

leszko:
> > So how, about setting average gas price to some more realistic number in Arbitrum, like 2 or 3. Wdyt?
> 
> 👍

Updated in 98cad6dc510f138751a9ac9cc889588372ef9ceb

PR #2488, opened 2022-07-01
save segment when FV failed

oscar-davids:
For #[352](https://github.com/livepeer/internal-project-tracking/issues/352)

- 
- 
- 

**Checklist:**

### DIFF
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -167,6 +167,9 @@ func parseLivepeerConfig() starter.LivepeerConfig {
 	cfg.Objectstore = flag.String("objectStore", *cfg.Objectstore, "url of primary object store")
 	cfg.Recordstore = flag.String("recordStore", *cfg.Recordstore, "url of object store for recordings")
 
+	// Fast Verification GS bucket:
+	cfg.FVfailGsBucket = flag.String("FVfailGsbucket", *cfg.FVfailGsBucket, "Google storage bucket for storing segments for fast verification")
+	cfg.FVfailGsKey = flag.String("FVfailGskey", *cfg.FVfailGsKey, "Google Storage private key file name or key itself (in json format)")
 	// API
 	cfg.AuthWebhookURL = flag.String("authWebhookUrl", *cfg.AuthWebhookURL, "RTMP authentication webhook URL")
 	cfg.OrchWebhookURL = flag.String("orchWebhookUrl", *cfg.OrchWebhookURL, "Orchestrator discovery callback URL")
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -168,8 +168,8 @@ func parseLivepeerConfig() starter.LivepeerConfig {
 	cfg.Recordstore = flag.String("recordStore", *cfg.Recordstore, "url of object store for recordings")
 
 	// Fast Verification GS bucket:
-	cfg.FVfailGsBucket = flag.String("FVfailGsbucket", *cfg.FVfailGsBucket, "Google storage bucket for storing segments for fast verification")
-	cfg.FVfailGsKey = flag.String("FVfailGskey", *cfg.FVfailGsKey, "Google Storage private key file name or key itself (in json format)")
+	cfg.FVfailGsBucket = flag.String("FVfailGsbucket", *cfg.FVfailGsBucket, "Google Cloud Storage bucket for storing segments, which failed fast verification")
+	cfg.FVfailGsKey = flag.String("FVfailGskey", *cfg.FVfailGsKey, "Google Cloud Storage private key file name or key in JSON format for accessing FVfailGsBucket")
 	// API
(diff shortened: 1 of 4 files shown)


server/broadcast.go
```diff
@@ -559,12 +559,26 @@ func (bsm *BroadcastSessionsManager) chooseResults(ctx context.Context, submitRe
...
+
 			}
-			clog.Infof(ctx, "Video comparison from url=%s and url=%s are equal=%v",
+			clog.Infof(ctx, "Video comparison from url=%s and url=%s are equal=%v saveenable=%v",
 				trustedResult.TranscodeResult.Segments[segmToCheckIndex].Url,
-				untrustedResult.TranscodeResult.Segments[segmToCheckIndex].Url, vequal)
+				untrustedResult.TranscodeResult.Segments[segmToCheckIndex].Url, vequal, drivers.FailSaveEnabled())
+		} else {
+			if drivers.FailSaveEnabled() {
+				if drivers.FailSaveEnabled() {
```
cyberj0g: Duplicate `if`?

server/broadcast.go
```diff
@@ -559,12 +559,26 @@ func (bsm *BroadcastSessionsManager) chooseResults(ctx context.Context, submitRe
...
 				return nil, nil, err
 			}
-			if monitor.Enabled && !vequal {
-				monitor.FastVerificationFailed(ctx, ouri, monitor.FVType2Error)
+			if !vequal {
+				if monitor.Enabled {
+					monitor.FastVerificationFailed(ctx, ouri, monitor.FVType2Error)
+				}
+				if drivers.FailSaveEnabled() {
+					drivers.SavePairData2GS(trustedResult.TranscodeResult.Segments[segmToCheckIndex].Url, trustedSegm,
```
cyberj0g: Should we upload async, since it's on the latency-critical path?

cmd/livepeer/livepeer.go
```diff
@@ -167,6 +167,9 @@ func parseLivepeerConfig() starter.LivepeerConfig {
 	cfg.Objectstore = flag.String("objectStore", *cfg.Objectstore, "url of primary object store")
 	cfg.Recordstore = flag.String("recordStore", *cfg.Recordstore, "url of object store for recordings")
 
+	// Fast Verification GS bucket:
+	cfg.FVfailGsBucket = flag.String("FVfailGsbucket", *cfg.FVfailGsBucket, "Google storage bucket for storing segments for fast verification")
```
cyberj0g: Sorry for nitpicking, but let's make user-facing messages more concise, like:
```
Google Cloud Storage bucket for storing segments, which failed fast verification

Google Cloud Storage private key file name or key in JSON format for accessing FVfailGsBucket
```

cyberj0g (changes requested):
The code is fine after addressing few comments. I have a slight concern regarding whether we should add this function to master branch, because users will never really need it, but it may be useful for an attacker to analyze verification behavior. Maybe we can just use this branch for testing? WDYT @thomshutt @yondonfu @oscar-davids

server/broadcast.go
oscar-davids: fixed.

cmd/livepeer/livepeer.go
oscar-davids: fixed.

oscar-davids:
> The code is fine after addressing few comments. I have a slight concern regarding whether we should add this function to master branch, because users will never really need it, but it may be useful for an attacker to analyze verification behavior. Maybe we can just use this branch for testing? 

It is currently used for analyzing Fast Verification fail cases but in the future, it would be used for Full Verification, I think.

server/broadcast.go
thomshutt: 👍 Definitely

thomshutt:
> The code is fine after addressing few comments. I have a slight concern regarding whether we should add this function to master branch, because users will never really need it, but it may be useful for an attacker to analyze verification behavior. Maybe we can just use this branch for testing? WDYT @thomshutt @yondonfu @oscar-davids

Interesting point, would someone looking to attack it in that way not already have what they need from this code being open source though?

server/broadcast.go
oscar-davids: Fixed

cyberj0g:
Right, the code is open source, but program arguments are lower hanging fruit for someone willing to do the research. Also, by adding parameters not directly required for performing intended function, we increasing potential attack surface in a broad sense (e.g. someone at some point could commit the code or post the comment with organization's GSC access credentials in the arguments, or there may be a critical vulnerability in GC library). However, if it's not a single-use addition, then I'm totally fine with merging it.

thomshutt:
@cyberj0g fair points, I had the same initial reaction to merging this in too but I think it's going to be too difficult to diagnose any false FV failure in Production without it. I'd be open to removing it in the future once we're comfortable this is working as intended though.

@oscar-davids could you confirm that this won't currently break Production (i.e where we don't have the GC config values set) if we tried to deploy it?

oscar-davids:
I am sure that this won't currently break the Production.

cyberj0g (approved):

PR #2489, opened 2022-07-01
eth: Backfill blocks in batches

### LINKED ISSUE #2480: Make it easier to keep nodes synced on block info
Lately I have been having issues with nodes not updating block info (usually Singapore, sometimes other locations). The orchestrator service stays available, so automatic health checks won't pick up on these issues. This means that I have to randomly restart my nodes in order to backfill block events which is a PITA

Assuming that having fairly recent block info is important for being able to receive streams, I would expect the node to:
- Keep retrying the defined RPC endpoint to get recent info. It seems that for some situations it just stops trying to pull new block info. Sometimes after an RPC error, but sometimes without any error at all.
- Throw an error if it hasn't been able to update block info for a while

An additional solution would be adding the option to define multiple RPC endpoints, so it can rotate between them. This way I can safely use the community node and have Alchemy and offchain RPC endpoints as backup. Since they all have their stability issues it would be nice if we could just use them all

leszko:
Optimize the number of ETH RPC request by first backfilling blocks in batches during the block sync. In result, only the latest block is polled separately instead of polling all missing blocks one-by-one.

This also fixes the Livepeer downtime issue in the case when Livepeer cannot catch up with the blocks and therefore can't transcode.

- In `syncToLatestBlock()`, execute first block backfilling and then poll block only for the latest one
- Update unit tests

- Deployed an orchestrator and checked that it can transcode the video
- Check that block backfilling catches all interesting information (e.g. DepositFund, etc.)

fix #2480

**Checklist:**

### DIFF
--- a/eth/blockwatch/block_watcher.go
+++ b/eth/blockwatch/block_watcher.go
@@ -85,10 +85,9 @@ func New(config Config) *Watcher {
 	return bs
 }
 
-// BackfillEventsIfNeeded finds missed events that might have occured while the
-// node was offline and sends them to event subscribers. It blocks until
-// it is done backfilling or the given context is canceled.
-func (w *Watcher) BackfillEventsIfNeeded(ctx context.Context) error {
+// BackfillEvents finds missed events and sends them to event subscribers.
+// It blocks until it is done backfilling or the given context is canceled.
+func (w *Watcher) BackfillEvents(ctx context.Context) error {
 	events, err := w.getMissedEventsToBackfill(ctx)
 	if err != nil {
 		return err
@@ -117,7 +116,7 @@ func (w *Watcher) Watch(ctx context.Context) error {
 			ticker.Stop()
 			return nil
 		case <-ticker.C:
-			if err := w.syncToLatestBlock(); err != nil {
+			if err := w.syncToLatestBlock(ctx); err != nil {
 				glog.Errorf("blockwatch.Watcher error encountered - trying again on next polling interval err=%q", err)
 			}
 		}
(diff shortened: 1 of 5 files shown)


red-0ne:
> This also fixes the Livepeer downtime issue in the case when Livepeer cannot catch up with the blocks and therefore can't transcode.

Is this solved by the fact that sync is faster because of fewer RPC calls?

eth/blockwatch/block_watcher.go
```diff
@@ -325,7 +333,9 @@ func (w *Watcher) getMissedEventsToBackfill(ctx context.Context) ([]*Event, erro
 	if err != nil {
 		return events, err
 	}
-	latestBlockNum := int(latestBlock.Number.Int64())
+
+	// Latest block will be polled separately in syncToLatestBlock(), so it's not backfilled.
+	preLatestBlockNum := int(latestBlock.Number.Int64()) - 1
```
red-0ne, line 338: Is there any reason behind omitting the latest block? Is it to always treat the latest block as a new one?

red-0ne (commented):
Have a couple of questions. Other than that, it looks fine for me.

leszko:
> > This also fixes the Livepeer downtime issue in the case when Livepeer cannot catch up with the blocks and therefore can't transcode.
> 
> Is this solved by the fact that sync is faster because of fewer RPC calls?

Right, it's solved, because in many cases Livepeer will never catch up with blocks polling them one by one, so it's effectively down. Using backfilling, it will most likely catch with the missing blocks.

eth/blockwatch/block_watcher.go
leszko: It's tricky, I tried to explain it here: https://github.com/livepeer/go-livepeer/pull/2489/files#diff-2f2752a740f61bfa5688f822d8850e6d751507520b6c28198afc60f35b8f90b3R89.

In short, we watch blocks for 2 reasons:
- Updating local cache => backfilling in batches is good enough
- Timing => Passing the block to the TimeWatcher, then we need to poll the block separately, because even if the batch does not contain the interesting event, we need to pass it to TimeWatcher

red-0ne (approved):
Looks good to me.

thomshutt (approved):

PR #2499, opened 2022-07-13
linux-arm64 build

iameli:
Three big changes here:

1. Implemented cross-compilation for linux-arm64 using `clang`; this is triggered with `GOARCH=arm64` just like we do for darwin-arm64.
2. Removed the multiple-Dockerfile mess and replaced them with a single, unified multi-stage Dockerfile
3. Switched to `docker build buildx` for the Docker build process, which uses the above cross-compilation to build and publish an arm64 go-livepeer build alongside our existing amd64 build. [Info on how that works here](https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/).

~~One note: I had to upgrade from Ubuntu 18.04 to Ubuntu 20.04 because I needed clang 10 - some bug related to `fastmath` stuff when we were on clang 8. This will bump our glibc dependency by two years. Possible we should either figure out how to downpatch or look into `musl`.~~ Fixed, build now works with Ubuntu 18.04.


docker/Dockerfile
```diff
@@ -0,0 +1,71 @@
...
+
+COPY go.mod go.mod
+COPY go.sum go.sum
+RUN go mod download
+
+COPY . .
+ENV CPATH /usr/aarch64-linux-gnu/include
+RUN make livepeer livepeer_cli livepeer_bench livepeer_router
+
+CMD ./upload_build.sh
```
hjpotter92: is upload needed from inside the container?

install_ffmpeg.sh
```diff
@@ -77,7 +87,7 @@ fi
 if [[ ! -e "$ROOT/x264" ]]; then
   git clone http://git.videolan.org/git/x264.git "$ROOT/x264"
   cd "$ROOT/x264"
-  if [[ $IS_M1 ]]; then
+  if [[ $IS_ARM64 ]]; then
     # newer git master, compiles on Apple Silicon
```
hjpotter92, line 91: and `arm64` ;)

.github/workflows/docker.yaml
```diff
@@ -53,7 +41,12 @@ jobs:
           # Our Docker tag name should be our branch name with just alphanums
           BRANCH_TAG="$(echo $GHA_REF | sed -e 's:refs/heads/::; s:/:-:g' | tr -cd '[:alnum:]_-')"
           VERSION_TAG=$(./print_version.sh)
-          docker build -t current-build -f docker/Dockerfile.release-linux .
+          docker buildx \
```
hjpotter92: would it support multi-platform builds for docker images as well? I can update this flow accordingly then.

iameli: I did that once, and it worked, but it took like 75 minutes, which would be even slower than the Windows build. We might be better off using the cross-compiler for the arm64 Docker image too... can different stages of a multistage Docker build use different architectures?

docker/Dockerfile
iameli: Not anymore, good point

.github/workflows/docker.yaml
iameli: They can, and I implemented it, woo

leszko (commented):
@iameli how about keeping the reviewers list shorter? I think 1-2 reviewers should be enough for this PR.

hjpotter92 (approved):

iameli:
@leszko will do!

iameli:
Attempting to get a build with 18.04 working for maximum compatibility, think this maybe will do it...

Issue #2485, opened 2022-06-30
Broadcast fails after a few segments when using -orchAddr domain.com


papabear99:
I was running a test stream this morning with a potential client to use my O for 24/7 streaming and the Broadcast was failings after a few segments leaving the B in an unusable stuck state.  I had him try using the IP address of one of my nodes and then it worked.

Start Broadcaster with the following:
.\livepeer -broadcaster -network arbitrum-one-mainnet -ethUrl https://arbitrum-mainnet.infura.io/v3/xxx -ethKeystorePath C:\Users\xxxx.lpData\arbitrum-one-mainnet\keystore\UTC--2022-06-xxx -rtmpAddr 127.0.0.1:1935 -httpAddr 0.0.0.0:8935 -transcodingOptions P240p25fps16x9,P360p25fps16x9,P576p25fps16x9,P720p25fps16x9 -orchAddr solarfarm.multiorch.com

**Expected behavior**
The stream should work with the domain name specified in the -orchAddr

**Screenshots**
This is what the Broadcaster showed
![BroadcasterScreenshot](https://user-images.githubusercontent.com/89408276/176712730-1ce846a4-30c4-4a2f-84b7-8d7f00436ed5.png)

**Additional context**
Please see this discord discussion from June 18th for additional context where a few Os were testing a similar setup and saw the same behavior. https://discord.com/channels/423160867534929930/431166747085897740/987811196348559411

Log from earlier test below
```
I think they are referring to this part where it transcoded a few segments before losing connection to the orchestrator.
broadcaster  | I0618 20:07:00.855896       1 segment_rpc.go:626] orchSessionID=3bc969d4 orchestrator=https://orch-chicago.xodeapp.xyz:8935 Successfully transcoded segment segName= seqNo=3 orch=https://orch-chicago.xodeapp.xyz:8935 dur=146.352305ms
broadcaster  | I0618 20:07:00.855946       1 drivers.go:210] Downloading uri=https://orch-chicago.xodeapp.xyz:8935/stream/3bc969d4/P360p30fps16x9/3.ts
broadcaster  | I0618 20:07:00.856062       1 drivers.go:210] Downloading uri=https://orch-chicago.xodeapp.xyz:8935/stream/3bc969d4/P240p30fps16x9/3.ts
[...]

PR #2500, opened 2022-07-13
net: fix no sessions error if O hostname doesn't match

cyberj0g:
Issue #2485 
Particularly nasty bug with observable issue occurring in unrelated session management logic, but straightforward resolution. The root cause is OrchestratorPool returning empty OrchestratorLocalInfo, if O's `-serviceAddr` doesn't match B's `-orchAddr` precisely (e.g. **localhost** won't match **127.0.0.1**). It might have caused silent O selection issues in any case when `-orchAddr` or `-serviceAddr` were specified.

- Add host name resolution function, with support for `localhost`

Issue #2485 

**Checklist:**

### DIFF
--- a/discovery/discovery.go
+++ b/discovery/discovery.go
@@ -105,7 +105,6 @@ func (o *orchestratorPool) GetInfo(uri string) common.OrchestratorLocalInfo {
 			}
 		}
 	}
-	// TODO: indicate error
 	return res
 }
 
--- a/discovery/discovery.go
+++ b/discovery/discovery.go
@@ -6,7 +6,9 @@ import (
 	"errors"
 	"math"
 	"math/rand"
+	net2 "net"
 	"net/url"
+	"strings"
 	"time"
 
 	"github.com/livepeer/go-livepeer/clog"
@@ -23,6 +25,7 @@ var getOrchestratorsCutoffTimeout = 500 * time.Millisecond
 var maxGetOrchestratorCutoffTimeout = 6 * time.Second
 
 var serverGetOrchInfo = server.GetOrchestratorInfo
+var hostLookupFn = net2.LookupHost
 
 type orchestratorPool struct {
 	infos []common.OrchestratorLocalInfo
@@ -55,14 +58,54 @@ func (o *orchestratorPool) GetInfos() []common.OrchestratorLocalInfo {
 	return o.infos
 }
 
+func ResolveURL(urlToResolve *url.URL) []*url.URL {
+	resolvedUrls := make([]*url.URL, 0, 1)
+	hostPort := strings.Split(urlToResolve.Host, ":")
+	if net2.ParseIP(hostPort[0]) == nil {
+		// handle localhost - won't work for custom hosts entry!
+		var addrs []string
+		var err error
+		if strings.ToLower(hostPort[0]) == "localhost" {
+			addrs = []string{"127.0.0.1"}
+		} else {
+			addrs, err = hostLookupFn(hostPort[0])
+		}
+		if err != nil {
+			clog.Errorf(context.TODO(), "Couldn't resolve domain for URL %s, check local DNS settings. Discovery may not work properly.", urlToResolve.String())
+			return append(resolvedUrls, urlToResolve)
+		}
+		for _, addr := range addrs {
+			newUrl, _ := url.Parse(urlToResolve.String())
+			if len(hostPort) == 2 {
+				newUrl.Host = addr + ":" + hostPort[1]
+			} else {
+				newUrl.Host = addr
+			}
+			resolvedUrls = append(resolvedUrls, newUrl)
+		}
+	} else {
+		// already ip, do nothing
+		return append(resolvedUrls, urlToResolve)
+	}
+	return resolvedUrls
+}
+
 func (o *orchestratorPool) GetInfo(uri string) common.OrchestratorLocalInfo {
 	var res common.OrchestratorLocalInfo
+	queryUrl, _ := url.Parse(uri)
+	resolvedUrls := ResolveURL(queryUrl)
 	for _, info := range o.infos {
-		if info.URL.String() == uri {
-			res = info
-			break
+		oResolvedUrls := ResolveURL(info.URL)
+		for _, oUrl := range oResolvedUrls {
+			for _, queryUrl := range resolvedUrls {
+				if oUrl.String() == queryUrl.String() {
+					res = info
+					break
+				}
+			}
 		}
 	}
+	// TODO: indicate error
 	return res
 }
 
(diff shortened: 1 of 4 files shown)


discovery/discovery.go
```diff
@@ -55,12 +59,65 @@ func (o *orchestratorPool) GetInfos() []common.OrchestratorLocalInfo {
...
+	queryUrl, _ := url.Parse(uri)
+	resolvedUrls := ResolveURL(queryUrl)
 	for _, info := range o.infos {
-		if info.URL.String() == uri {
-			res = info
-			break
+		oResolvedUrls := ResolveURL(info.URL)
+		for _, oUrl := range oResolvedUrls {
+			for _, queryUrl := range resolvedUrls {
+				if oUrl.String() == queryUrl.String() {
```
leszko:
```suggestion
			for _, qUrl := range resolvedUrls {
				if oUrl.String() == qUrl.String() {
```

Nit: you already have the `queryUrl` defined.

discovery/discovery.go
```diff
@@ -55,12 +59,65 @@ func (o *orchestratorPool) GetInfos() []common.OrchestratorLocalInfo {
...
 func (o *orchestratorPool) GetInfo(uri string) common.OrchestratorLocalInfo {
 	var res common.OrchestratorLocalInfo
+	queryUrl, _ := url.Parse(uri)
+	resolvedUrls := ResolveURL(queryUrl)
 	for _, info := range o.infos {
-		if info.URL.String() == uri {
-			res = info
-			break
+		oResolvedUrls := ResolveURL(info.URL)
+		for _, oUrl := range oResolvedUrls {
```
leszko:
```suggestion
		for _, oUrl := range ResolveURL(info.URL) {
```

Nit: inline

discovery/discovery.go
```diff
@@ -55,12 +59,65 @@ func (o *orchestratorPool) GetInfos() []common.OrchestratorLocalInfo {
 	return o.infos
 }
 
+func ResolveURL(urlToResolve *url.URL) []*url.URL {
+	resolvedUrls := make([]*url.URL, 0, 0)
+	// check if there's a port in host, for ipv6 it's after ]
```
leszko: Isn't `urlToResolve` already parsed to `host` and `port`. I think that `urlToResolve.Host` should contain host (without port) and `urlToResolve.Port()` should contain the port.

If that's not the case, I guess we should just use `url.ParseRequestURI()` instead of this custom logic below.

discovery/discovery.go
```diff
@@ -55,12 +59,65 @@ func (o *orchestratorPool) GetInfos() []common.OrchestratorLocalInfo {
 	return o.infos
 }
 
+func ResolveURL(urlToResolve *url.URL) []*url.URL {
+	resolvedUrls := make([]*url.URL, 0, 0)
```
leszko:
```suggestion
	var resolvedUrls []*url.URL
```

Nit.

discovery/discovery.go
```diff
@@ -55,12 +59,65 @@ func (o *orchestratorPool) GetInfos() []common.OrchestratorLocalInfo {
...
+		var addrs []string
+		var err error
+		if strings.ToLower(host) == "localhost" {
+			// ipv6 addr may be abbreviated or not
+			addrs = []string{"127.0.0.1", "::1", "0000:0000:0000:0000:0000:0000:0000:0001"}
+		} else {
+			addrs, err = hostLookupFn(host)
+		}
+		if err != nil {
+			clog.Errorf(context.TODO(), "Couldn't resolve domain for URL %s, check local DNS settings. Discovery may not work properly.", urlToResolve.String())
```
leszko:
```suggestion
			glog.Errorf("Couldn't resolve domain for URL %s, check local DNS settings", urlToResolve.String())
```
Nit

discovery/discovery.go
```diff
@@ -55,12 +59,65 @@ func (o *orchestratorPool) GetInfos() []common.OrchestratorLocalInfo {
 	return o.infos
 }
 
+func ResolveURL(urlToResolve *url.URL) []*url.URL {
```
leszko, line 62: Should we always add the input `urlToResolve` to `resolvedUrls`, just in case we missed something in the logic below, but the addresses are already equal? That would be safer in terms of backward-compatibility.

server/broadcast.go
```diff
@@ -722,6 +722,10 @@ func selectOrchestrator(ctx context.Context, n *core.LivepeerNode, params *core.
 			bcastOS = bcastOS.OS().NewSession(pfx)
 		}
 
+		oinfo := n.OrchestratorPool.GetInfo(tinfo.Transcoder)
```
leszko, line 725: Is this change related to your PR? This seems like a candidate for a separate PR.

leszko (changes requested):
@cyberj0g Thanks for the PR. Added some comments. But in general it's a very necessary change!

discovery/discovery.go
cyberj0g: Fixed.

discovery/discovery.go
cyberj0g: Thanks for suggestion, replaced with library functions.

discovery/discovery.go
cyberj0g: Fixed.

server/broadcast.go
cyberj0g: Yes, that what was making this bug hard to track down. `GetInfo` silently returns uninitialized structure in case there's no URL match. Ideally, we need to refactor it to return error, but it will require changes to the interface, probably should be a separate PR.

discovery/discovery.go
cyberj0g: Fixed.

discovery/discovery.go
cyberj0g: Fixed.

discovery/discovery.go
cyberj0g: Added backward compatibility shortcut to `GetInfo()`.

leszko (approved):

server/broadcast.go
RiccardoBiosas, line 725: is it the reason why the bug was occurring only when establishing a session for the transcoding?
Because when a broadcaster first submits an orchestrator's domain, it is parsed in a way that is supposed to match the format of the `serviceAddr` - except apparently for the localhost/127.0.0.1 edge-case. In fact in the case of https://github.com/livepeer/go-livepeer/issues/2485 not prefixing the O domain `solarfarm.multiorch.com` with `https` is not a problem right away because the [`parseOrchAddrs` function](https://github.com/livepeer/go-livepeer/blob/master/cmd/livepeer/starter/starter.go#L1191-L1209) takes care of reformatting it - if so, rather than creating a new function (`ResolveURL`) can't we reuse the same util used in `starter.go` (with the appropriate changes to handle the ipv6/localhost edge-cases)?

PR #2510, opened 2022-07-16
docker.yaml: Provide `BUILD_TAGS` for docker build steps.

hjpotter92:

Adds missing `BUILD_TAGS` arg when building docker image

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -13,6 +13,19 @@ jobs:
     if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name == github.repository
     runs-on: ubuntu-20.04
     steps:
+      - name: Check out code
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+          # Check https://github.com/livepeer/go-livepeer/pull/1891
+          # for ref value discussion
+          ref: ${{ github.event.pull_request.head.sha }}
+
+      - name: Get build tags
+        id: build-tag
+        run: |
+          ./ci_env.sh
+
       - name: Set up QEMU
         uses: docker/setup-qemu-action@v2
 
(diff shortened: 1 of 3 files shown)


victorges (approved):
LGTM

.github/workflows/docker.yaml
```diff
@@ -38,13 +54,14 @@ jobs:
...
             type=semver,pattern={{major}}.{{minor}}
-            type=raw,value={{date 'YYYYMMDDHHmmss'}}
             type=raw,value=latest,enable={{is_default_branch}}
             type=raw,value=${{ github.event.pull_request.head.ref }}
 
       - name: Build minimal livepeer distributable
         uses: docker/build-push-action@v3
         with:
+          build-args: |
+            BUILD_TAGS=${{ steps.build-tag.outputs.build-tags }}
```
victorges, line 61: Isn't the output name just `tags`?

.github/workflows/docker.yaml
```diff
@@ -13,6 +13,22 @@ jobs:
...
+          # Check https://github.com/livepeer/go-livepeer/pull/1891
+          # for ref value discussion
+          ref: ${{ github.event.pull_request.head.sha }}
+
+      - name: Get build tags
+        id: build-tag
+        run: |
+          ./ci_env.sh echo '::set-output name=build-tags::$BUILD_TAGS'
+          set -x
+          echo "::set-output name=tags::$BUILD_TAGS"
```
victorges: Actually, they are inconsistent, the first one [1] has `build-tags` while the second one has `tags`.

[1] I actually don't know what that first echo does. What is it for?

ci_env.sh
```diff
@@ -55,4 +55,6 @@ if [[ $generatedVersion != $definedVersion ]]; then
   export BUILD_TAGS="${BUILD_TAGS},experimental"
 fi
 
+echo "::set-output name=build-tags::${BUILD_TAGS}"
+
 exec "$@"
```
victorges, line 60: Do we still need this?

victorges (approved):
LGTM

ci_env.sh
hjpotter92: the build script in `build.yaml` still uses this as `./ci_env.sh make` (and it works o_0)

docker/Dockerfile
```diff
@@ -1,68 +1,66 @@
-FROM --platform=$BUILDPLATFORM ubuntu:18.04 as build
```
iameli, line 1: @hjpotter92 This some kind of autoformatter?

hjpotter92: yeah, dockfmt: <https://github.com/jessfraz/dockfmt>

PR #2541, opened 2022-07-27
server: Enforce a minimum timeout for segment upload

victorges:
This is to fix the segment upload timeout for really short segments, which I 
found to be an issue when investigating frequent timeouts in VOD tests.

In this case ([logs](https://eu-metrics-monitoring.livepeer.live/grafana/explore?orgId=1&left=%5B%221658945061444%22,%221658945187292%22,%22Loki%22,%7B%22exemplar%22:true,%22expr%22:%22%7Bapp%3D%5C%22prod-task-runner%5C%22%7D%20%7C~%20%5C%229c771156-978a-404a-b4f1-66245328fec1%5C%22%22,%22refId%22:%22A%22%7D%5D)), there is a frequent `header timeout` happening in the tests,
and even by increasing the retries and adding some backoff logic, it seems to keep
happening pretty often.

I then noticed that it always happened with the same segment of the test video, and 
when segmenting it locally with the same code I found that this segment has a duration
of `533ms` (and only so short cause it's the last seg in the file).

This means that the upload timeout given to it would be a mere `250ms`, which
can easily be hit for longer round-trip times to orchestrators even if they would be able to
respond healthily.

- Make `MinSegmentUploadTimeout` be a minimum instead of just a default

Ran automatic tests. 

No.

**Checklist:**

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -22,6 +22,9 @@
 
 #### Broadcaster
 
+-   [#2541](https://github.com/livepeer/go-livepeer/pull/2541) Enforce a minimum
+    timeout for segment upload (@victorges)
+
 #### Orchestrator
 
 #### Transcoder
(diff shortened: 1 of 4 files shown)


leszko (approved):
Good fix, thanks @victorges 👍

PR #2544, opened 2022-07-28
docker: Fix amd64 build

victorges:
Newer CUDA version apparently breaks things,
but armr64 cannot be built without that.

Build only amd64 docker images for now with older CUDA.

Just wait for docker images build.

Fixes go-livepeer docker image.

**Checklist:**


iameli (approved):
✅ to unblock deploys, let's get the arm64 builds back soon though

PR #2550, opened 2022-08-01
server: Use HeaderByNumber() instead of BlockByNumber() in the function requesting LPT from faucet

### LINKED ISSUE #2547: Unable to get latest block: transaction type not supported
env:

image
image

bug:

image
image

please help me       then  I want to become a node for make money

leszko:

After the Arbitrum Nitro migration, the function `BlockByNumber()` from `ethereum/go-ethereum` fails with the following message:
```
Unable to get latest block: transaction type not supported
```
It's related to the fact that Arbitrum uses its own transaction format, and therefore requires its own `go-ethereum` fork [OffchainLabs/go-ethereum](https://github.com/OffchainLabs/go-ethereum).

One option is to use [OffchainLabs/go-ethereum](https://github.com/OffchainLabs/go-ethereum), what we considered in https://github.com/livepeer/go-livepeer/pull/2470. Another option, implemented in this PR, is to find all the chain interactions that fail and fix them. I checked all the transaction executions and actually this one is the one that fails. Therefore, I suggest to continue using `ethereum/go-ethereum` and merge this PR.

- 
- 
- 

Tested in Arbitrum Rinkeby Testnet

fix https://github.com/livepeer/go-livepeer/issues/2547

**Checklist:**

### DIFF
--- a/server/handlers.go
+++ b/server/handlers.go
@@ -1029,13 +1029,13 @@ func requestTokensHandler(client eth.LivepeerEthClient) http.Handler {
 		}
 
 		backend := client.Backend()
-		blk, err := backend.BlockByNumber(r.Context(), nil)
+		h, err := backend.HeaderByNumber(r.Context(), nil)
 		if err != nil {
 			respond500(w, fmt.Sprintf("Unable to get latest block: %v", err))
 			return
 		}
 
-		now := int64(blk.Time())
+		now := int64(h.Time)
 		if nextValidRequest.Int64() != 0 && nextValidRequest.Int64() > now {
 			respond500(w, fmt.Sprintf("Error requesting tokens from faucet: can only request tokens once every hour, please wait %v more minutes", (nextValidRequest.Int64()-now)/60))
 			return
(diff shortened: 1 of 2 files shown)


thomshutt (approved):
LGTM - This is safe to merge now right? (i.e we don't need to align deploying it with the migration)

leszko:
> LGTM - This is safe to merge now right? (i.e we don't need to align deploying it with the migration)

Yes, it's safe.

server/handlers.go
```diff
@@ -1029,13 +1029,13 @@ func requestTokensHandler(client eth.LivepeerEthClient) http.Handler {
 		}
 
 		backend := client.Backend()
-		blk, err := backend.BlockByNumber(r.Context(), nil)
+		h, err := backend.HeaderByNumber(r.Context(), nil)
```
yondonfu, line 1032: Probably better to only fetch the block header as opposed to the entire block here anyway because we don't care about most of the data in the block anyway and we can just get the information we need from the slimmer header.

yondonfu (approved):
LGTM

Issue #2560, opened 2022-08-04
go-livepeer should crash if it has a nil orchestrator pool


iameli:
Encountered this problem with a mainnet broadcaster today:

```
INFO: E0804 16:06:14.773260     194 starter.go:975] Could not create orchestrator pool with DB cache: Could not refresh DB list of orchestrators: unsupported block number 19384378
```

I'm not sure of the root cause. We saw problems very similar to this [when Infura was omitting their block number](https://community.infura.io/t/arbitrum-eth-getblockbynumber-omits-l1blocknumber-field/5462/2). So root cause still under investigation.

That said, **we should crash if we find ourselves in this state**. If you try and transcode without an orchestrator pool, we just segfault over and over again:
```
2022/08/04 15:53:15 http: panic serving 127.0.0.1:42912: runtime error: invalid memory address or nil pointer dereference
goroutine 1423 [running]:
net/http.(*conn).serve.func1()
	/opt/hostedtoolcache/go/1.17.12/x64/src/net/http/server.go:1802 +0xb9
panic({0x367c80, 0x23c7570})
	/opt/hostedtoolcache/go/1.17.12/x64/src/runtime/panic.go:1047 +0x266
github.com/livepeer/go-livepeer/discovery.(*DBOrchestratorPoolCache).nextRound(0x0)
	/home/runner/work/go-livepeer/go-livepeer/discovery/db_discovery.go:342 +0x1d
github.com/livepeer/go-livepeer/discovery.(*DBOrchestratorPoolCache).Size(0x0)
	/home/runner/work/go-livepeer/go-livepeer/discovery/db_discovery.go:156 +0x3b
github.com/livepeer/go-livepeer/discovery.(*DBOrchestratorPoolCache).SizeWith(0x10, 0xc0007aade8)
	/home/runner/work/go-livepeer/go-livepeer/discovery/db_discovery.go:165 +0x32
github.com/livepeer/go-livepeer/server.NewSessionManager({0x7066c0, 0xc013d14120}, 0xc0002bd4a0, 0xc0005ad950, 0x0)
	/home/runner/work/go-livepeer/go-livepeer/server/broadcast.go:374 +0xe9
github.com/livepeer/go-livepeer/server.(*LivepeerServer).registerConnection(0xc010dbcfc0, {0x7066c0, 0xc010dd41b0}, {0x71edd8, 0xc00031c540}, 0xc0007ab0f0, {0x0}, 0xc0001b52c0)
[...]

PR #2563, opened 2022-08-09
Retry remote contract calls for unsupported block number errors

yondonfu:

We've occasionally observed an `unsupported block number` error for RPC requests which can cause problems for node operation (see #2560).

According to an [Offchain Labs team member](https://discord.com/channels/585084330037084172/765654444825641001/968910226583584798), an Arb node can return this error if the node is not synced. This situation could occur if:

- A single Arb node has fallen behind the tip of the chain
- If requests are load balanced across a set of nodes and request 1 fetches block number N from node 1, but request 2 that references block number N is routed to node 2 which has only seen block number N - 1

The `unsupported block number` error appears to originate from the [getSnapshot()](https://github.com/OffchainLabs/arbitrum/blob/93340983cc28b9689c67f5117c1a6e42ba3f5332/packages/arb-rpc-node/web3/eth.go#L689) and [getSnapshotForNumberForHash()](https://github.com/OffchainLabs/arbitrum/blob/93340983cc28b9689c67f5117c1a6e42ba3f5332/packages/arb-rpc-node/web3/eth.go#L732) functions in `offchainlabs/arbitrum/packages/arb-rpc-node/web3` which are called by many of the functions implementing the ETH JSON-RPC API [1].

From reviewing Discord conversations, it appears that this error is more likely to happen when sending concurrent RPC requests.

The solution in this PR is to add the `unsupported block number` error to the list of errors that will be retried when sending requests for remote contract calls. Note: This change only applies to contract calls and not other RPC requests which could be affected by this error as well. Addressing this issue with *all* RPC requests requires additional code changes, but since this error is primarily affecting contract calls right now, I've opted to just address this for contract calls.

See commit history.

- Added unit tests for helper functions
- Rolled this change to a broadcaster and [observed retries](https://eu-metrics-monitoring.livepeer.live/grafana/explore?orgId=1&left=%5B%22now-30m%22,%22now%22,%22Loki%22,%7B%22expr%22:%22%7Bapp%20%3D~%20%5C%22prod-livepeer-secondary-broadcaster-.%2B%5C%22%7D%20%7C~%20%5C%22unsupported%5C%22%22,%22refId%22:%22A%22,%22range%22:true%7D%5D) for `unsupported block number` errors

Addresses the root cause of #2560 although it could still make sense to update the behavior of the node following the suggestion in the issue.

**Checklist:**

### DIFF
--- a/eth/backend.go
+++ b/eth/backend.go
@@ -7,6 +7,7 @@ import (
 	"math/big"
 	"strings"
 	"sync"
+	"time"
 
 	ethereum "github.com/ethereum/go-ethereum"
 	"github.com/ethereum/go-ethereum/accounts/abi"
@@ -169,6 +170,11 @@ func (b *backend) retryRemoteCall(remoteCall func() ([]byte, error)) (out []byte
 		if err != nil && isRetryableRemoteCallError(err) {
 			glog.Error(err)
 			glog.V(4).Infof("Retrying call to remote ethereum node")
+			// We could use the backoff package to configure an exponential backoff here, but it makes it
+			// more difficult to propagate the result from the call if the remote call is successful because
+			// the backoff functions can only return a single error type.
+			// So, to keep things simple just sleep here for 1 second before trying again.
+			time.Sleep(1 * time.Second)
 		} else {
 			retry = false
 		}
--- a/eth/backend.go
+++ b/eth/backend.go
@@ -168,8 +168,7 @@ func (b *backend) retryRemoteCall(remoteCall func() ([]byte, error)) (out []byte
 	for i := 0; i < count && retry; i++ {
 		out, err = remoteCall()
 		if err != nil && isRetryableRemoteCallError(err) {
-			glog.Error(err)
-			glog.V(4).Infof("Retrying call to remote ethereum node")
+			glog.V(4).Infof("Retrying call to remote ethereum node err=%v", err)
 			// We could use the backoff package to configure an exponential backoff here, but it makes it
 			// more difficult to propagate the result from the call if the remote call is successful because
 			// the backoff functions can only return a single error type.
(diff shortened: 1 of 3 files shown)


eth/backend.go
```diff
@@ -161,13 +165,17 @@ func (b *backend) PendingCallContract(ctx context.Context, msg ethereum.CallMsg)
...
 		out, err = remoteCall()
-		if err != nil && (err.Error() == "EOF" || err.Error() == "tls: use of closed connection") {
-			glog.V(4).Infof("Retrying call to remote ethereum node")
+		if err != nil && isRetryableRemoteCallError(err) {
+			glog.V(4).Infof("Retrying call to remote ethereum node err=%v", err)
+			// We could use the backoff package to configure an exponential backoff here, but it makes it
+			// more difficult to propagate the result from the call if the remote call is successful because
+			// the backoff functions can only return a single error type.
+			// So, to keep things simple just sleep here for 1 second before trying again.
+			time.Sleep(remoteCallRetrySleep)
```
leszko: I think you could implement exponential backoff just by multiplying it with `i`.

```
time.Sleep(i * remoteCallRetrySleep)
```

Then, you could make `remoteCallRetrySleep` slightly smaller, like `500ms`.

leszko (approved):
Added one comment. Other than that LGTM 👍

eth/backend.go
yondonfu: Good point. I ended up implementing a linear backoff by multiplying `remoteCallRetrySleep` by `i` in [c04c2ee](https://github.com/livepeer/go-livepeer/pull/2563/commits/c04c2ee7eff5f86b74adad238fac101e474f926b) based on your suggestion which feels good enough.

yondonfu:
Rebased to re-order commits.

thomshutt (commented):
LGTM

PR #2577, opened 2022-09-02
cmd+pm: Replace depositMultipler with maxFaceValue for sender

yondonfu:

Remove the `-depositMultiplier` flag in favor of re-using the existing `-maxFaceValue` flag to configure the maximum faceValue that will be accepted in ticket params received by a broadcaster. If `-maxFaceValue 100` is set, then a broadcaster will reject ticket params sent from an orchestrator if faceValue > 100`. By default, `-maxFaceValue` is 0 and there will be maximum imposed by the broadcaster.

Previously, the value of the `-depositMultiplier` flag would be used to derive a maxFaceValue as `current deposit / depositMultiplier`. In practice, this feels like a pretty confusing way to configure the maxFaceValue and explicitly setting it for a broadcaster via the existing `-maxFaceValue` flag feels much simpler. I considered keeping the `-depositMultiplier` flag for backwards compatibility, but 1) I suspect that it is not used 2) maintaining support for that flag as well as the use of the `-maxFaceValue` would result in some tech debt code paths that felt inconvenient. Welcome feedback on this point though.

See commit history.

Updated unit tests. 

N/A

**Checklist:**

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -4,6 +4,8 @@
 
 ### Breaking Changes 🚨🚨
 
+- The `-depositMultiplier` flag has been removed. This flag was previously used by a broadcaster to configure a maximum face value for ticket params sent by orchestrator by dividing the broadcaster's current deposit by the flag's value. The `-maxFaceValue` flag can now be used to explicitly set the maximum face value for ticket params instead.
+
 ### Features ⚒
 
 #### General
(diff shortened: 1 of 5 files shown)


leszko (approved):
The PR looks good 👍 

My only concern is if we want to release the non-backward-compatible change in the patch version release (`v0.5.34` => `v0.5.35`). Anyway, maybe it's ok, I leave it up to you to decide.

yondonfu:
Hm I thought a bit more about the breaking change of removing the `-depositMultiplier` flag here and I'll see if there is a low effort way to add a deprecation warning for the flag for v0.5.35 and then remove the flag in v0.5.36.

PR #2581, opened 2022-09-05
Enable metrics by default

### LINKED ISSUE #2580: change -monitor default to true
I'm unclear on why we don't default to including the monitoring metrics. There's not much overhead associated with them and they're useful. I've never heard of another service where Prometheus metrics are off by default. They're only exposed on the already-needs-to-be-secure 7935 port.

Personally, the only place I've ever encountered this flag is helping people understand why they can't query for `localhost:7935/metrics` and it's because they don't have the monitoring flag enabled.

Any objections?

thomshutt:
Makes metrics enabled by default

- Remove `-monitor` CLI flag 
- Remove all checks for this flag
- Fix test race condition that's introduced by now calling monitor during test

- Ran all tests

Fixes #2580 

**Checklist:**


cyberj0g (commented):
I'm OK with enabling monitoring by default, but have mixed feelings on removing the option to disable it. It seem to create goroutines on each call, which try to connect to local grafana, even if it's not there. Not sure whether this, along with network stack timeout, could theoretically create some bottlenecks.

thomshutt:
Closing this, will revisit if / when we can do it without constantly trying to connect to a local Grafana

PR #2583, opened 2022-09-07
eth: Set tx GasFeeCap to min(gasPriceEstimate, current GasFeeCap)

### LINKED ISSUE #2574: 'Failed: insufficient funds for gas * price + value' - Cannot process transactions after Nitro upgrade
**Describe the bug**
After upgrading to Nitro, some orchestrators including myself cannot execute transactions like calling reward or redeeming a winning ticket. 

Initially, the error looked like this:
```
HTTP Response Error 500: Error calling reward: gas required exceeds allowance (152271)
```
Upon node restart, the error evolved into this:
```
Aug 31 17:21:42 lpr-Not-Specified livepeer[554821]: Invoking transaction: "rewardWithHint". Inputs: "_newPosPrev: 0xE3a5793d7c1D2a04A903FA1695b3E3555d6084CA  _newPosNext: 0x4a43B1D7e62                   27C8b0512e413F406555647ff7bdB"
Aug 31 17:21:42 lpr-Not-Specified livepeer[554821]: Transaction Failed: insufficient funds for gas * price + value: address 0x9D61ae5875E89036FBf6059f3116d01a22ACe3C8 have 604543848649                   53520 want 266627000000000000
```
Or, on the node trying to redeem a winning ticket, this:

```
Invoking transaction: "redeemWinningTicket". Inputs: "_ticket: { Recipient: 0x9D61ae5875E89036FBf6059f3116d01a22ACe3C8  Sender: 0xc3c7c4C8f7061B7d6A72766Eee5359fE4F36e61E  FaceValue: 12000000000000000  WinProb: 9649340769776349618630915417390658987772498722136713669954799000000000000  SenderNonce: 1  RecipientRandHash: 0x83fe58b22811435b18d0c2b90eba17206f612f1072f2805fa792c808d06a0170  AuxData: 0x0000000000000000000000000000000000000000000000000000000000000a7a613ebf57b80834b99a2ec150ea7b2b0cb97419d0c2d31dfeca4659fc6a92688e }  _sig: 0x1f956188526c89394e467aecd529669c65b5c5a5913079721585486dee654ccc1dd79c631f18385e1aeb079634202328ce3d85345ccc9e22e0b51d8f7694da5c1b  _recipientRand: 55259894694139197534321694537779125253224093903070031163731600211776028401548"
Transaction Failed: insufficient funds for gas * price + value: address 0xcdF93B13EFCf63f84f2afFB01BeC48A5c096C69E have 25795810339399299 want 328761600000000000

E0831 22:08:41.746572       1 queue.go:149] Error redeeming err="insufficient funds for gas * price + value: address 0xcdF93B13EFCf63f84f2afFB01BeC48A5c096C69E have 25795810339399299 want 328761600000000000"
```
It looks like it's requesting an absurd amount to redeem a winning ticket or call reward, and the cost seems to get higher over time.
We've tried multiple RPC providers (Alchemy, offchain, full arb node) and restarting our nodes but have not had any success.

We CAN, however, call reward normally via https://arbiscan.io/address/0x35Bcf3c30594191d53231E4FF333E8A770453e40#writeProxyContract

My node setup:
Independent local Ubuntu node for calling reward
7 Dockerized Geolocated O/T's 

Upon talking to other O's using similar setups, this does not seem to be an issue that affects them. 
We also had no issues prior to the upgrade.

**To Reproduce**
1. Call reward or redeem a winning ticket
2. Observe logs
3. If the transaction completes normally, you're not affected by this
4. If you see the error listed above, you're in the same boat as me and some others... sorry

**Expected behavior**
Transactions should process normally after the Nitro upgrade.

 - OS: Ubuntu
 - Version: 22.04.1

**Additional context**
It's hard to see how this could be a configuration error on the orchestrators side, as this started immediately after the Nitro upgrade and several RPC providers have been tested, none of them resolving the issue, while other orchestrators don't seem to have a problem.

yondonfu:

At the moment, if the `-maxGasPrice` flag is used the `GasFeeCap` for all txs is also set to the `maxGasPrice` value. The `GasFeeCap` serves as the maximum fee/price that will be charged per unit of gas consumed by a tx. The minimum balance of an account is the `gas used by the tx * GasFeeCap`. If `GasFeeCap` exceeds the actual gas price that is required for txs then a refund will be provided to the account based on the difference between `GasFeeCap` and the actual gas price. But, if the account does not have enough balance cover `gas used by the tx * GasFeeCap` then an `Insufficient funds...` error will be encountered.

In https://github.com/livepeer/go-livepeer/issues/2574, some O operators set a value for `-maxGasPrice` that is quite a bit higher than the actual average gas price paid on Arbitrum (i.e. ~0.1 gwei). As a result, even though their account balances could cover a tx based on the actual gas price paid, their account balances could not cover a tx based on `GasFeeCap = maxGasPrice`. The current available workarounds are to either top off account balances to ensure the balance can cover `gas * GasFeeCap` (with the excess fee refunded to the balance at the end of the successful tx), set a lower value for `-maxGasPrice` or stop using the `-maxGasPrice` flag.

I felt that while there are available workarounds, the way the `-maxGasPrice` flag currently works is a bit weird since it can easily lead to a node operator thinking they are just being safe by setting a high value for `-maxGasPrice`, but then having that high value cause tx issues if they do not have super high account balances. So, I opted to modify the `GasFeeCap` of txs to `min(gasPriceEstimate, current GasFeeCap` where `gasPriceEstimate = baseFee * 2` so that in these cases a node operator doesn't run into tx issues.

See in-line code comments for more details.

See commit history.

Tested on Arbitrum Rinkeby:

1. Ran a node off the master branch

2. Used `livepeer_cli` to set maxGasPrice = 300 gwei

3. Invoked a LPT transfer and observed an `insufficient funds...` error

```

Invoking transaction: "transfer". Inputs: "recipient: 0x3F7bcA8c726Bc34a903ed3e838997d7d3A29B99d  amount: 0"
Transaction Failed: insufficient funds for gas * price + value: address 0x3F7bcA8c726Bc34a903ed3e838997d7d3A29B99d have 6233423641163 want 8282700000000000

E0907 22:23:51.538450       1 handlers.go:1421] HTTP Response Error 500: insufficient funds for gas * price + value: address 0x3F7bcA8c726Bc34a903ed3e838997d7d3A29B99d have 6233423641163 want 8282700000000000
```

4. Restarted the node running off of the branch for this PR

5. Used `livepeer_cli` to set maxGasPrice = 300 gwei

6. Invoked a LPT transfer and observed a successful tx.

```

Invoking transaction: "transfer". Inputs: "recipient: 0x3F7bcA8c726Bc34a903ed3e838997d7d3A29B99d  amount: 0"  Hash: "0x5db565e84ba009689fd63792f80ecdb3cb16500ce2c23d4249c17dfb26245528".

I0907 22:21:12.384937       1 handlers.go:1018] Transferred 0 LPTU to 0x3F7bcA8c726Bc34a903ed3e838997d7d3A29B99d
```

Fixes #2574 

**Checklist:**

### DIFF
--- a/eth/client.go
+++ b/eth/client.go
@@ -370,6 +370,30 @@ func (c *client) transactOpts() *bind.TransactOpts {
 
 	opts.Context = newEthRpcContext()
 
+	// If GasFeeCap is nil then one of the following will be true:
+	// - A dynamic tx will be created by BoundContract and GasFeeCap will automatically be set.
+	// - A legacy tx will be created by BoundContract in which case the GasFeeCap field is not used.
+	if opts.GasFeeCap == nil {
+		return &opts
+	}
+
+	// If GasFeeCap is non-nil ensure that we adjust it to be min(gasPriceEstimate, current GasFeeCap).
+	gasPriceEstimate := opts.GasFeeCap
+	head, err := c.backend.HeaderByNumber(context.Background(), nil)
+	if err != nil {
+		glog.Errorf("failed to calculate gas price estimate - defaulting to using GasFeeCap = maxGasPrice")
+	} else {
+		gasPriceEstimate = new(big.Int).Mul(head.BaseFee, big.NewInt(2))
+	}
+	// Setting GasFeeCap > gasPriceEstimate is detrimental to the user because the user account will need at least gas * GasFeeCap in their balance
+	// to pay for the tx when they only need gas * gasPriceEstimate. GasFeeCap is initially going to be set to the maxGasPrice specified by the user which
+	// they expect to be the maximum gas price they will ever pay. So, a user might set maxGasPrice to 100 gwei as a safety precaution even if the average gas price
+	// is 0.1 gwei. In this case, the gasPriceEstimate would be 0.1 * 2 = 0.2 gwei (multiplied by 2 to add a buffer). So, the user account would need 500x more funds
+	// in its balance if its GasFeeCap is set to the maxGasPrice.
+	if opts.GasFeeCap.Cmp(gasPriceEstimate) > 0 {
+		opts.GasFeeCap = gasPriceEstimate
+	}
+
 	return &opts
 }
 
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -19,9 +19,10 @@
 #### CLI
 
 #### General
+- [#2583](https://github.com/livepeer/go-livepeer/pull/2583) eth: Set tx GasFeeCap to min(gasPriceEstimate, current GasFeeCap) (@yondonfu)
 
 #### Broadcaster
--   [#2573](https://github.com/livepeer/go-livepeer/pull/2573) server: Fix timeout for stream recording background jobs (@victorges)
+- [#2573](https://github.com/livepeer/go-livepeer/pull/2573) server: Fix timeout for stream recording background jobs (@victorges)
 
 #### Orchestrator
 


eth/client.go
```diff
@@ -370,6 +370,26 @@ func (c *client) transactOpts() *bind.TransactOpts {
...
+	if err != nil {
+		glog.Errorf("failed to calculate gas price estimate - defaulting to using GasFeeCap = maxGasPrice")
+	} else {
+		gasPriceEstimate = new(big.Int).Mul(head.BaseFee, big.NewInt(2))
+	}
+	// If GasFeeCap is non-nil ensure that we adjust it to be min(gasPriceEstimate, current GasFeeCap).
+	// Setting GasFeeCap > gasPriceEstimate is detrimental to the user because the user account will need at least gas * GasFeeCap in their balance
+	// to pay for the tx when they only need gas * gasPriceEstimate. GasFeeCap is initially going to be set to the maxGasPrice specified by the user which
+	// they expect to be the maximum gas price they will ever pay. So, a user might set maxGasPrice to 100 gwei as a safety precaution even if the average gas price
+	// is 0.1 gwei. In this case, the gasPriceEstimate would be 0.1 * 2 = 0.2 gwei (multipled by 2 to add a buffer). So, the user account would need 500x more funds
```
leszko:
```suggestion
	// is 0.1 gwei. In this case, the gasPriceEstimate would be 0.1 * 2 = 0.2 gwei (multiplied by 2 to add a buffer). So, the user account would need 500x more funds
```

eth/client.go
```diff
@@ -370,6 +370,26 @@ func (c *client) transactOpts() *bind.TransactOpts {
...
+	// in its balance if its GasFeeCap is set to the maxGasPrice.
+	// If GasFeeCap is nil then one of the following will be true:
+	// - A dynamic tx will be created by BoundContract and GasFeeCap will automatically be set.
+	// - A legacy tx will be created by BoundContract in which case the GasFeeCap field is not used.
+	if opts.GasFeeCap != nil && opts.GasFeeCap.Cmp(gasPriceEstimate) > 0 {
```
leszko: I suggest to move the condition `opts.GasFeeCap != nil` upper, because if it's `nil`, then we don't even need to make an RPC call to get the current base fee. So something like this:

```
// Only for London-ready txs
	if opts.GasFeeCap != nil {
		gasPriceEstimate := opts.GasFeeCap
		head, err := c.backend.HeaderByNumber(context.Background(), nil)
		if err != nil {
			glog.Errorf("failed to calculate gas price estimate - defaulting to using GasFeeCap = maxGasPrice")
		} else {
			gasPriceEstimate = new(big.Int).Mul(head.BaseFee, big.NewInt(2))
		}
		// If GasFeeCap is non-nil ensure that we adjust it to be min(gasPriceEstimate, current GasFeeCap).
		// Setting GasFeeCap > gasPriceEstimate is detrimental to the user because the user account will need at least gas * GasFeeCap in their balance
		// to pay for the tx when they only need gas * gasPriceEstimate. GasFeeCap is initially going to be set to the maxGasPrice specified by the user which
		// they expect to be the maximum gas price they will ever pay. So, a user might set maxGasPrice to 100 gwei as a safety precaution even if the average gas price
		// is 0.1 gwei. In this case, the gasPriceEstimate would be 0.1 * 2 = 0.2 gwei (multipled by 2 to add a buffer). So, the user account would need 500x more funds
		// in its balance if its GasFeeCap is set to the maxGasPrice.
		// If GasFeeCap is nil then one of the following will be true:
		// - A dynamic tx will be created by BoundContract and GasFeeCap will automatically be set.
		// - A legacy tx will be created by BoundContract in which case the GasFeeCap field is not used.
		if opts.GasFeeCap.Cmp(gasPriceEstimate) > 0 {
			opts.GasFeeCap = gasPriceEstimate
		}
	}
```

leszko (approved):
Added 2 comments. Other than that LGTM. Good investigation @yondonfu !

eth/client.go
yondonfu: Good call!

yondonfu: Fixed in https://github.com/livepeer/go-livepeer/pull/2583/commits/f8f6e908f2f0f5a85bf81162cc487401b8c88fbd

eth/client.go
yondonfu: Fixed in https://github.com/livepeer/go-livepeer/pull/2583/commits/e2a15eefb391b932c71bf0b56ddad71886b19224

yondonfu:
I'll see if there are any reports in https://github.com/livepeer/go-livepeer/issues/2574 and then rebase + merge.

PR #2587, opened 2022-09-15
build.yaml: Add workflow to generate build manifests

hjpotter92:
- Add job to generate branch manifest file alongside uploading archives to GCP storage
 - Merge macos and linux builders to single job

---

Built binaries, are uploaded to gcp bucket for each commit. Additionally, merge multiple build jobs into a single one.

Part of work wrt: livepeer/livepeer-infra#917


cyberj0g (approved):
LGTM!

Issue #1801, opened 2021-03-18
Run golangci-lint on all packages


yondonfu:
We currently only run golangci-lint on [certain packages](https://github.com/livepeer/go-livepeer/blob/6f084a3d2d089249818546e85ef9e285fb64292a/.github/workflows/linux.yml#L67) in CI. We should run golangci-lint locally, fix all the warnings/errors and then start running golangci-lint on all packages in CI. The one thing to consider here is which linters should be enabled by default (or just use the default linters used by golangci-lint).

cyberj0g:
Might want to prioritize because of bugs like #2589 (PR #2591). 
`golangci-lint` can catch that:
```
58:core/orchestrator.go:634:4: lostcancel: the cancel function is not used on all paths (possible context leak) (govet)
59-			ctx, cancel := context.WithTimeout(context.Background(), transcodeLoopTimeout)
60-			^
61:core/orchestrator.go:639:5: lostcancel: this return statement may be reached without using the cancel var defined on line 634 (govet)
62-				return
63-				^
```
There are about 200 linting warnings currently, I see a few options how to minimize efforts of enabling it:
1. disable linters which check for less severe errors, like `errcheck`, `govet`, `unused`, `gosimple`
2. only lint files changed in the commit, which still may be painful and could be combined with option 1
3. do not lint tests

yondonfu:
> disable linters which check for less severe errors

In the case of https://github.com/livepeer/go-livepeer/issues/2589 we would have wanted `govet` to be enabled to flag the context issues right?

> only lint files changed in the commit, which still may be painful and could be combined with option 1

A few alternatives I am wondering about:

1. Lint changed lines only
2. Configure the linter to emit certain less severe errors as just warnings so that CI doesn't fail for them and only fails for the more "interesting" errors

Then, a gradual approach might look like:

- Lint files changed in the commit
- If someone opens a PR that fails linting, first create a separate PR that fixes the linting issues in the relevant files
- Merge the separate PR and then rebase the original PR which should pass linting

PR #2591, opened 2022-09-16
core: Return from transcode loop if seg chan closed

### LINKED ISSUE #2589: O transcode loop never returns if the transcode session is ended via B
This O wasn't doing much, just transcoding the occasional segment. However its CPU usage (as well as the other staging orchestrators in the cluster) were stuck at 1500%. [Here's a dump of all the available goroutines.](https://gist.github.com/iameli/714b7a9c7882ee522b771c3bbeec25c5). Looks like there's lots and lots of routines with traces something like this...?
```
goroutine 79303 [runnable]:
time.Time.Add({0xc0c127f998d0c387, 0x11b13e8736cf, 0x27c6860}, 0xdf8475800)
	/usr/local/go/src/time/time.go:820 +0x146
context.WithTimeout({0x87d330, 0xc00004a0d0}, 0xdf8475800)
	/usr/local/go/src/context/context.go:507 +0x32
github.com/livepeer/go-livepeer/core.(*LivepeerNode).transcodeSegmentLoop.func1()
	/src/core/orchestrator.go:634 +0x7c
created by github.com/livepeer/go-livepeer/core.(*LivepeerNode).transcodeSegmentLoop
	/src/core/orchestrator.go:631 +0x436
```

yondonfu:

This PR updates the transcode loop to exit the loop if the segment channel is closed.

https://github.com/livepeer/go-livepeer/commit/a1fb761474685ae0679e26349e27f19012d95a75 introduced the possibility of the segment channel being closed via a EndTranscodingSession RPC call. When that happens, the bug described in [this comment](https://github.com/livepeer/go-livepeer/issues/2589#issuecomment-1249591252) can occur.

See commit history.

[45bf1e1](https://github.com/livepeer/go-livepeer/pull/2591/commits/45bf1e1faf9a42d50aba551ea6787282d96c22a2) contains a small update to avoid logging a nil error if there was no error when B calls EndTranscodingSession.

- Added a unit test that checks of the goroutine for the transcode loop is cleaned up after the transcode session is ended
- Ran manual tests with an offchain B & O [mediaserver.httpTimeout](https://github.com/livepeer/go-livepeer/blob/5d4e145b4d2f6efbbcca0f5cd6f8481163ac8e81/server/mediaserver.go#L74) set to 30 seconds so B always sends an EndTranscodingSession RPC call before O's transcode loop times out
     - Test 1: Build B & O off `master`
     - Test 2: Build B & O off this branch
     - After both tests, I ran `kill -ABRT <PID>` where `<PID>` was the process ID for O to dump a stack trace of goroutines
     - The stack trace for test 1 showed 1 `core.transcodeSegmentLoop()` goroutine which indicates that the goroutine was not cleaned up properly after the transcoding session was ended
     - The stack trace for test 2 showed 0 `core.transcodeSegmentLoop()` goroutines which indiciates that the goroutine was cleaned up properly after the transcoding session was ended

Fixes #2589 

**Checklist:**

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -15,12 +15,12 @@
 #### Transcoder
 
 ### Bug Fixes 🐞
-- [#2586](https://github.com/livepeer/go-livepeer/pull/2586) Broadcaster: Don't pass a nil context into grpc call or it panics
 
 #### CLI
 
 #### General
 - [#2583](https://github.com/livepeer/go-livepeer/pull/2583) eth: Set tx GasFeeCap to min(gasPriceEstimate, current GasFeeCap) (@yondonfu)
+- [#2586](https://github.com/livepeer/go-livepeer/pull/2586) Broadcaster: Don't pass a nil context into grpc call or it panics (@thomshutt, @cyberj0g)
 
 #### Broadcaster
 - [#2573](https://github.com/livepeer/go-livepeer/pull/2573) server: Fix timeout for stream recording background jobs (@victorges)
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -27,5 +27,6 @@
 - [#2586](https://github.com/livepeer/go-livepeer/pull/2586) Refactor RTMP connection object management to prevent race conditions (@cyberj0g)
 
 #### Orchestrator
+- [#2591](https://github.com/livepeer/go-livepeer/pull/2591) Return from transcode loop if transcode session is ended by B (@yondonfu)
 
 #### Transcoder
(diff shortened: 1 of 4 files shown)


cyberj0g (approved):
:scream: how did I miss that. Wonder if some static analysis tool would catch such errors. We definitely need to review PRs with concurrency more thoroughly as well.

Edit: found an existing issue, related to linting #1801

yondonfu:
I think looking into starting to use some static analysis tooling to catch bugs such as missing context cancellations is a good idea. Adding test cases for resource cleanup for changes that could affect goroutine management should help as well (i.e. the one added in this PR would've caught the bug).

PR #2606, opened 2022-09-23
Possible fix for dependabot issues and add CodeQL workflow

hjpotter92:

- The dependabot generated PRs don't have access to secrets. This PR tries to resolve that.
- Adds CodeQL workflow to be run parallel with tests.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -1,12 +1,17 @@
 name: Build binaries
 
 on:
+  pull_request_target:
   pull_request:
   push:
     branches:
       - master
     tags:
       - "v*"
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
 
 jobs:
   build:
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -160,15 +160,9 @@ jobs:
         uses: actions/setup-go@v3
         with:
           go-version: 1.17
-
-      - name: Cache go modules
-        id: cache-go-mod
-        uses: actions/cache@v2.1.5
-        with:
-          path: ~/go/pkg/mod
-          key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
-          restore-keys: |
-            ${{ runner.os }}-go-
+          go-version-file: go.mod
+          cache: true
+          cache-dependency-path: go.sum
 
       - name: Cache ffmpeg
         id: cache-ffmpeg
@@ -203,7 +197,7 @@ jobs:
             && sudo ldconfig
 
       - name: Install go modules
-        if: steps.cache-go-mod.outputs.cache-hit != 'true'
+        if: steps.go.outputs.cache-hit != 'true'
         run: go mod download
 
       - name: Install ffmpeg
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -1,12 +1,19 @@
 name: Trigger test suite
 
 on:
+  pull_request_target:
+    branches:
+      - master
   pull_request:
     branches:
       - master
   push:
     branches:
       - master
+
+concurrency:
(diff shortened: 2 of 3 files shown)


.github/workflows/build.yaml
```diff
@@ -1,13 +1,18 @@
 name: Build binaries
 
 on:
+  pull_request_target:
```
leszko, line 4: Isn't this a security hole? E.g., a malicious user could write a code that steals our secrets and send a PR.

.github/workflows/build.yaml
```diff
@@ -1,13 +1,18 @@
...
   pull_request:
   push:
     branches:
       - master
     tags:
       - "v*"
 
+concurrency:
```
leszko, line 12: Is it related to `The dependabot generated PRs don't have access to secrets`? I can't connect the dots how it's related.

.github/workflows/test.yaml
```diff
@@ -93,6 +94,42 @@ jobs:
         uses: codecov/codecov-action@v3
         with:
           files: ./cover.out
-          name: go-livepeer
+          name: ${{ github.event.repository.name }}
           verbose: true
           gcov_ignore: ./eth/stubclient.go
+
+  codeql:
```
leszko, line 101: Is this related to `The dependabot generated PRs don't have access to secrets`?

leszko (changes requested):

.github/workflows/build.yaml
hjpotter92: dependabot secrets are stored separate from the default ones: 

image

but any PR (from external users) would still need to be approved by repo owner before the CI picks up the job. The same is true for `pull_request` based trigger.

.github/workflows/build.yaml
hjpotter92: concurrency avoids running the same pipeline multiple times; and with dependabot generating/rebasing the PRs, it is just a good practice to use concurrency.

.github/workflows/test.yaml
hjpotter92: codeql itself isn't related to dependabot.

.github/workflows/build.yaml
leszko: Ok, then it makes sense.

.github/workflows/build.yaml
leszko: 👍

.github/workflows/test.yaml
leszko: ok, so could you please send it as a separate PR or update the description of this PR?

hjpotter92: updated the description here.

.github/workflows/test.yaml
```diff
@@ -93,6 +94,42 @@ jobs:
...
+        uses: github/codeql-action/autobuild@v1
+
+      # ℹ️ Command-line programs to run using the OS shell.
+      # 📚 https://git.io/JvXDl
+
+      # ✏️ If the Autobuild fails above, remove it and uncomment the following three lines
+      #    and modify them (or add more) to build your code if your project
+      #    uses a compiled language
+
+      #- run: |
```
leszko: Could you maybe clean up this commented-out code?

.github/workflows/test.yaml
```diff
@@ -93,6 +94,42 @@ jobs:
...
+          ref: ${{ github.event.pull_request.head.sha }}
+
+      # Initializes the CodeQL tools for scanning.
+      - name: Initialize CodeQL
+        uses: github/codeql-action/init@v1
+        with:
+          languages: go
+
+      # Autobuild attempts to build any compiled languages  (C/C++, C#, or Java).
+      # If this step fails, then you should remove it and run the build manually (see below)
```
leszko: I'm not sure I understand this command. When should we "remove it"?

leszko (approved):
LGTM after cleaning up comments.

Issue #2582, opened 2022-09-07
Segments fail to upload on commit 73aaca8c3da1f5269fe059dd03b0cbf4d6796a1c


stronk-dev:
All Orchestrator nodes were operating fine on 0.5.34, except for Boston, which was running on a [special version version of 0.5.34](https://github.com/livepeer/go-livepeer/commits/yf/arb-tip-0)

Almost all segments fail to upload, causing test stream scores to plummet. Weirdly enough, it did not seem to affect my actual transcoding work. One other Orchestrator also confirmed this issue while he was on above version and also rolled back to 0.5.34 to fix the issue

I don't much info to give you, as the code does not print the actual error message when it fails...
```
E0907 04:44:32.830831 2603746 segment_rpc.go:199] manifestID=1a8d689d-9357-4d12-9aa8-0286f8c42876 seqNo=3934 orchSessionID=5084a2e3 clientIP=195.181.169.69 sender=0xc3c7c4C8f7061B7d6A72766Eee5359fE4F36e61E Could not upload segment
E0907 04:44:31.397356 2603746 segment_rpc.go:199] manifestID=12fdaa0c-ea69-44cf-8fe3-dbbe427a66d2 seqNo=1327 orchSessionID=8210a3e9 clientIP=212.102.58.242 sender=0xc3c7c4C8f7061B7d6A72766Eee5359fE4F36e61E Could not upload segment
E0907 04:44:30.675425 2603746 segment_rpc.go:199] manifestID=1a8d689d-9357-4d12-9aa8-0286f8c42876 seqNo=3933 orchSessionID=5084a2e3 clientIP=195.181.169.69 sender=0xc3c7c4C8f7061B7d6A72766Eee5359fE4F36e61E Could not upload segment
E0907 04:44:29.525661 2603746 segment_rpc.go:199] manifestID=12fdaa0c-ea69-44cf-8fe3-dbbe427a66d2 seqNo=1326 orchSessionID=8210a3e9 clientIP=212.102.58.242 sender=0xc3c7c4C8f7061B7d6A72766Eee5359fE4F36e61E Could not upload segment
E0907 04:44:28.652747 2603746 segment_rpc.go:199] manifestID=1a8d689d-9357-4d12-9aa8-0286f8c42876 seqNo=3932 orchSessionID=5084a2e3 clientIP=195.181.169.69 sender=0xc3c7c4C8f7061B7d6A72766Eee5359fE4F36e61E Could not upload segment
[...]

PR #2607, opened 2022-09-26
protect storage session per transcoding one

oscar-davids:

for #2582

- 
- 
- 

**Checklist:**

### DIFF
--- a/core/livepeernode.go
+++ b/core/livepeernode.go
@@ -85,13 +85,13 @@ type LivepeerNode struct {
 	Sender pm.Sender
 
 	// Thread safety for config fields
-	mu sync.RWMutex
-	// Transcoder private fields
-	priceInfo      *big.Rat
-	serviceURI     url.URL
-	segmentMutex   *sync.RWMutex
+	mu             sync.RWMutex
 	StorageConfigs map[string]*transcodeConfig
 	storageMutex   *sync.RWMutex
+	// Transcoder private fields
+	priceInfo    *big.Rat
+	serviceURI   url.URL
+	segmentMutex *sync.RWMutex
 }
 
 //NewLivepeerNode creates a new Livepeer Node. Eth can be nil.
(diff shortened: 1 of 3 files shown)


core/livepeernode.go
```diff
@@ -87,10 +87,11 @@ type LivepeerNode struct {
...
 	mu sync.RWMutex
 	// Transcoder private fields
-	priceInfo     *big.Rat
-	serviceURI    url.URL
-	segmentMutex  *sync.RWMutex
-	StorageConfig *transcodeConfig
+	priceInfo      *big.Rat
+	serviceURI     url.URL
+	segmentMutex   *sync.RWMutex
+	StorageConfigs map[string]*transcodeConfig
```
leszko, line 89: Nit: can't it be package-private, like `storageConfigs`? Do you use it outside of this package?

core/orchestrator.go
```diff
@@ -653,10 +656,13 @@ func (n *LivepeerNode) transcodeSegmentLoop(logCtx context.Context, md *SegTrans
 
 func (n *LivepeerNode) endTranscodingSession(sessionId string, logCtx context.Context) {
 	// timeout; clean up goroutine here
-	if n.StorageConfig != nil {
-		n.StorageConfig.OS.EndSession()
-		n.StorageConfig.LocalOS.EndSession()
+	n.storageMutex.Lock()
+	if storage, exists := n.StorageConfigs[sessionId]; exists {
+		storage.OS.EndSession()
```
leszko, line 661: This seems like an OS operation which may take a moment, then would it maybe be possible to execute outside of the lock? Otherwise, you may experience other goroutines waiting on this lock.

leszko (changes requested):

core/orchestrator.go
oscar-davids: StorageConfigs need thread safety because called by go routine [here](https://github.com/livepeer/go-livepeer/blob/80e70628ed1cace7a88784a37dc5312bb86663bf/core/orchestrator.go#L638) and grpc protocol.

core/livepeernode.go
oscar-davids: moved StorageConfigs into Thread safety fields

cyberj0g (approved):
LGTM!

leszko (approved):

Issue #2613, opened 2022-10-05
Config variables not working properlly


stronk-dev:
Config parameters often break, where even though they are defined in the launch script they do not get parsed correctly and use their default value instead. Changing the order of the parameters or using a `=` instead of a ` ` often fixes the issue, which causes confusion for new O's coming in:

- EG, [John Obelish had to move their `monitor` flag in order to get it to work](https://discord.com/channels/423160867534929930/932724294230900776/1022512517530005575)
- Organic-node has `-autoAdjustPrice=false` and `-pricePerUnit=1199` but didn't receive any work since it was adjusting above 1200 (pricePerUnit is now set to 1099 and he is receiving work again)
- [Another instance of the config variables not being parsed correctly](https://discord.com/channels/423160867534929930/429350403272998913/1013836772834283611)

There have been plenty of other instances where over the past few months, maybe we can add some more verbosity over what config variables it has read or what values it is using?

cyberj0g:
I think, it's a good idea to echo customized (because full list is too large) config values after `livepeer` start.  WDYT?

stronk-dev:
Yeah that sounds like a good first step, maybe we can also echo the full launch command on boot? So far I've only seen these kind of issues pop up with command line flags and not with with environment variables or config files

thomshutt:
Not sure how easy this would be given we're using the golang `flags` package, but maybe printing all non-default flags could be a good subset?

thomshutt:
Just realised that's what you meant by `customized` @cyberj0g 🤦 In that case, sounds good to me!

PR #2616, opened 2022-10-07
Print set config values on node start

cyberj0g:
As discussed in #2613, echo non-nil fields of config object, which changed with CLI arguments, config file or environment variables.

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -19,6 +19,7 @@
 #### CLI
 
 #### General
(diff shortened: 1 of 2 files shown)


thomshutt (approved):

iameli:
Reverting — this breaks `livepeer -j` by returning a bunch of non-JSON that MistController can't parse when it's booting up

iameli:
Reverting — this breaks `livepeer -j` by returning a bunch of non-JSON that MistController can't parse when it's booting up

PR #2624, opened 2022-10-17
cmd: don't print a table if we want Mist JSON

iameli:
Fixes `livepeer -j` JSON-based service discovery utilized by MistServer.


cyberj0g (approved):
LGTM!

PR #2650, opened 2022-11-11
Add a new header field in the HTTP PUSH endpoint that increases upload timeout

thomshutt:
Allows non-live workflows to specify that they're more tolerant of slow connections to transcoders.

- Accept new field `transcodeTimeoutMultiplier` in `Livepeer-Transcode-Configuration` header
- Use value from this (if set) to increase upload timeout

- Ran unit tests
- Tested with local T and confirmed timeout in logs

No

**Checklist:**

### DIFF
--- a/server/mediaserver.go
+++ b/server/mediaserver.go
@@ -146,7 +146,8 @@ type authWebhookResponse struct {
 			Name string `json:"name"`
 		} `json:"sceneClassification"`
 	} `json:"detection"`
-	VerificationFreq uint `json:"verificationFreq"`
+	VerificationFreq           uint `json:"verificationFreq"`
+	TranscodeTimeoutMultiplier int  `json:"transcodeTimeoutMultiplier"`
 }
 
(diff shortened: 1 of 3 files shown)


emranemran (approved):

server/mediaserver.go
```diff
@@ -865,6 +865,9 @@ func (s *LivepeerServer) HandlePush(w http.ResponseWriter, r *http.Request) {
 			return
 		}
 		params := streamParams(appData)
+		if authHeaderConfig != nil {
+			params.SegUploadTimeoutMultiplier = authHeaderConfig.TranscodeTimeoutMultiplier
```
yondonfu: Any reason why this is called TranscodeTimeoutMultiplier instead of SegUploadTimeoutMultiplier? Seems like it should be the latter given that this only affects the upload timeout and not the transcode timeout.

thomshutt: Good point, have updated to be consistent now

PR #2652, opened 2022-11-14
Fast verification multi-session test

cyberj0g:

A set of more comprehensive fast verification tests, including:
- selecting and using 'good' untrusted session in case of mismatched hash
- selecting and using 'good' untrusted session in case of mismatched segment data
- selecting and using trusted session if none of untrusted sessions successfully verified
- working with more than 3 sessions

List of changes:
- rename `shouldRunVerification()` to `shouldSkipVerification()` to match semantics
- an option to make multi-session transcoding sequential for testing
- tests and refactoring of the tests

### DIFF
--- a/discovery/discovery.go
+++ b/discovery/discovery.go
@@ -36,8 +36,13 @@ func NewOrchestratorPool(bcast common.Broadcaster, uris []*url.URL, score float3
 		glog.Error("Orchestrator pool does not have any URIs")
 	}
 	infos := make([]common.OrchestratorLocalInfo, 0, len(uris))
-	for _, uri := range uris {
-		infos = append(infos, common.OrchestratorLocalInfo{URL: uri, Score: score})
+	if len(uris) == 2 {
+		infos = append(infos, common.OrchestratorLocalInfo{URL: uris[0], Score: common.Score_Trusted})
+		infos = append(infos, common.OrchestratorLocalInfo{URL: uris[1], Score: common.Score_Untrusted})
+	} else {
+		for _, uri := range uris {
+			infos = append(infos, common.OrchestratorLocalInfo{URL: uri, Score: score})
+		}
 	}
 
 	return &orchestratorPool{infos: infos, bcast: bcast}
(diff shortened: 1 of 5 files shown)


server/mediaserver.go
```diff
@@ -336,6 +336,7 @@ func createRTMPStreamIDHandler(_ctx context.Context, s *LivepeerServer, webhookR
 		} else {
 			profiles = BroadcastJobVideoProfiles
 		}
+		//VerificationFreq = 1
```
thomshutt: Looks like we can drop this

server/broadcast.go
```diff
@@ -359,7 +363,7 @@ func (bsm *BroadcastSessionsManager) isVerificationEnabled() bool {
 	return bsm.VerificationFreq > 0
 }
 
-func (bsm *BroadcastSessionsManager) shouldRunVerification(sessions []*BroadcastSession) bool {
+func (bsm *BroadcastSessionsManager) shouldSkipVerification(sessions []*BroadcastSession) bool {
 	if bsm.verifiedSession == nil {
```
thomshutt, line 367: So it's now saying we should skip verification if there isn't a verified session?

discovery/discovery.go
```diff
@@ -36,8 +36,13 @@ func NewOrchestratorPool(bcast common.Broadcaster, uris []*url.URL, score float3
 		glog.Error("Orchestrator pool does not have any URIs")
 	}
 	infos := make([]common.OrchestratorLocalInfo, 0, len(uris))
-	for _, uri := range uris {
-		infos = append(infos, common.OrchestratorLocalInfo{URL: uri, Score: score})
+	if len(uris) == 2 {
```
thomshutt: Could you add a comment here about why the behaviour is different with 2 URIs and how we know the first is trusted and the second untrusted?

thomshutt (approved):
Looks like you've found some pretty major issues and great to add some testing around this. The only things I've commented on are bits I don't understand, so just adding some clarifying comments should be enough.

server/broadcast.go
cyberj0g: It says we should *not* skip verification (return false), if there's no verified session. Previous method name was confusing.

server/mediaserver.go
cyberj0g: Removed.

discovery/discovery.go
cyberj0g: Good catch, it's a debugging piece I forgot.

Issue #2614, opened 2022-10-05
Tons of `Invalid Ticket senderNonce` for the past few months


stronk-dev:
There seems to be tons of `Invalid Ticket senderNonce` coming in, especially in North America:
[1] https://discord.com/channels/423160867534929930/932724294230900776/1024598556784463902
[2] https://discord.com/channels/423160867534929930/932724294230900776/1023253242697764924

It seems to be affecting all Orchestrators, did something change in the Broadcaster config? Maybe related to changes to use secondary broadcasters rather than the primary ones

Franck-UltimaRatio:
I can confirm that, it greatly increases the error rate on tickets received, especially in North America.
![image](https://user-images.githubusercontent.com/87469831/194094837-4376d8e2-0c69-4bbe-9648-7403406b80b9.png)

Franck-UltimaRatio:
I can confirm that, it greatly increases the error rate on tickets received, especially in North America.
![image](https://user-images.githubusercontent.com/87469831/194095193-04f5f5af-394d-4d10-a291-956f737f1069.png)

stronk-dev:
Have to add that I have also had very few winning tickets lately, even though there have been plenty of work/tickets coming in (averaging 0.0153 ETH daily in tickets in the last 7d, but only one winning ticket of 0.012 Eth). Could this be related?

leszko:
@Franck-UltimaRatio @stronk-dev Can you confirm that is still an issue?

Franck-UltimaRatio:
Yes it s still an issue, this is the number of error during the last hour on our NYC node (linux ubuntu)
![image](https://user-images.githubusercontent.com/87469831/196125481-e8b43e7f-67b3-4540-9c41-1cdf16453e10.png)

stronk-dev:
![image](https://user-images.githubusercontent.com/9502378/196228695-48e661f6-ee90-4700-ab48-a39a7e4f21e0.png)
Seem to be happening less than before, but still some of them in US only

leszko:
Is there any way you can check the Brodcaster's ETH addresses these request are coming from?

Franck-UltimaRatio:
[...]

PR #2655, opened 2022-11-17
test/e2e: Add deposit & sequential segment push tests

yondonfu:

Adds e2e tests for the following workflows:

- A B depositing funds (i.e. funding its deposit and reserve on-chain)
- A B pushing segments sequentially to an O

Also, 8daa31f96e84ae940c9103149fe50f410646d649 fixes a bug in the O registration + activation test where the service URI registered on-chain was incorrect. Previously, the pointer address of a service URI var was being registered on-chain instead of the string value of the service URI var.

This PR is a prelude for adding a e2e test for B pushing segments in parallel to an O that reproduces https://github.com/livepeer/go-livepeer/issues/2614.

See commit history.

Ran the test suite.

N/A

**Checklist:**


cyberj0g (approved):
LGTM!

PR #2666, opened 2022-11-24
B re-use last session if it passes latency score threshold check

### LINKED ISSUE #2660: B eagerly switches to another O if it has used that O before and thinks it will be faster even if the current O has a real-time latency score

**Background**

After B runs discovery in order to create a pool of Os for selection, the following rules apply for selection that are relevant for this issue:

- A "known session" is defined as a session that uses an O that B has sent a segment to before and that has a last known latency score
- An "unknown session" is defined as a session that uses an O that B has not sent a segment to before and that does not have a last known latency score
- A latency score is stored for each [BroadcastSession](https://github.com/livepeer/go-livepeer/blob/a5045a223d43f80884ab2013045a942b35c2c17e/server/rpc.go#L116) and is [calculated as the ratio](https://github.com/livepeer/go-livepeer/blob/a5045a223d43f80884ab2013045a942b35c2c17e/server/segment_rpc.go#L634) of the round trip transcode time for a segment (upload + transcode + download) to the segment duration
- If there are zero known sessions, B selects from the list of unknown sessions [1]
- If there are existing known sessions, but the one with the best latency score is > the threshold (currently set to 1.0), B selects from the list of unknown sessions [2]
- If there are existing known sessions and the one with the best latency score is <= the threshold, B selects the known session with the best latency score [3]
- B keeps track of "segments in flight" for a session. If there is a segment in flight for a session then, B will only re-use the session for the next segment if the criteria [here](https://github.com/livepeer/go-livepeer/blob/a5045a223d43f80884ab2013045a942b35c2c17e/server/broadcast.go#L209) are fulfilled. Otherwise, B will not re-use the session

[1] https://github.com/livepeer/go-livepeer/blob/a5045a223d43f80884ab2013045a942b35c2c17e/server/selection.go#L137
[2] https://github.com/livepeer/go-livepeer/blob/a5045a223d43f80884ab2013045a942b35c2c17e/server/selection.go#L142
[3] https://github.com/livepeer/go-livepeer/blob/a5045a223d43f80884ab2013045a942b35c2c17e/server/selection.go#L145

**Problem**

Suppose the known session list for B is [O1.LatencyScore = 0.7. O2.LatencyScore = 0.8] and the following scenario:

1. B submits segment N to O1
2. The known session list = [O2.LatencyScore = 0.8]
3. B receives the results for segment N
4. [completeSession](https://github.com/livepeer/go-livepeer/blob/a5045a223d43f80884ab2013045a942b35c2c17e/server/broadcast.go#L324) is called for O1 and the SegsInFlight field is cleared for the session since there is no longer a segment in flight.
5. The latency score of O1 for segment N was 0.9 so now the known session list = [O2.LatencyScore = 0.8, O1.LatencyScore = 0.9]
6. B submits segment N + 1 to O2 because it has the best latency score even though O1 had a latency score that met the acceptable threshold which causes a swap from O1 to O2 for the stream

This type of swapping behavior can be problematic because:

1. If the switch is to an O that was previously used, but no longer has an active transcoding session there will be overhead for initializing a new session which increases latency

2. Video players may not handle data coming in at very different rates i.e. the last segment came in in Xs, this segment came in in Ys where X and Y are quite different

3. There are likely going to be diminishing returns once you're faster than real-time i.e. if writing transcoded data (so that it can be delivered) is slower than transcoding speed

B should stick with the last O that it used for transcoding a stream as long as it met the acceptable threshold.

The use of a last known latency score for each session also contributes to this problem because if an O is slightly slower than real-time for a single segment then a swap to another O could be triggered which may not be ideal because:

- The new O may not do better esp given the overhead of session initialization for the first segment it receives
- Being slightly slower than real-time might not be that big of a deal for a single segment

A better scoring mechanism for the session would incorporate latency scores over time - something like a moving average or a EWMA. This type of latency score formulation would dampen the impact of a single slightly slower than real-time segment on selection which can help stabilize selection for streams. I suspect this would be a bigger change though and I think it can be tackled separately/in parallel to this issue in https://github.com/livepeer/go-livepeer/issues/1232.

**Additional context**

yondonfu:

This PR implements the solution described in https://github.com/livepeer/go-livepeer/issues/2660.

Outside of this PR I think it would be a good idea to do a refactor to move as much selection logic (i.e. last session tracking, segment in flight tracking, etc.) into MinLSSelector in selection.go to keep that logic encapsulated in a single place. This felt like a hefty change though so I opted not to do that refactor in this PR.

We may also want to separately consider tweaking the value of the constant `SELECTOR_LATENCY_SCORE_THRESHOLD` to be slightly higher i.e. 1.1, 1.2 to be more lenient with the latency score threshold check, at least until we get to implementing https://github.com/livepeer/go-livepeer/issues/1232, if we still end up seeing a lot of swapping between Os during selection - I think the probability of this happening will be higher in scenarios with short segments (i.e. 1s) and lots of rendition profiles which would increase transcode + download time.

See commit history.

[8dcfaf4](https://github.com/livepeer/go-livepeer/pull/2666/commits/8dcfaf480648686e2698e7eeaf5cf2eadc613196) and [5be5fe8](https://github.com/livepeer/go-livepeer/pull/2666/commits/5be5fe8fd93bf9b71dc40d0bc46c909be3bd6ac8) contain small code refactors.

[3fc43f3](https://github.com/livepeer/go-livepeer/pull/2666/commits/3fc43f3ed4a230b19a738cb1ba8cfb0b618985ba) contains the main update.

Updated unit tests.

Fixes https://github.com/livepeer/go-livepeer/issues/2660

**Checklist:**

### DIFF
--- a/server/broadcast.go
+++ b/server/broadcast.go
@@ -207,11 +207,23 @@ func removeSessionFromList(sessions []*BroadcastSession, sess *BroadcastSession)
 
 func selectSession(sessions []*BroadcastSession, exclude []*BroadcastSession, durMult int) *BroadcastSession {
 	for _, session := range sessions {
-		if len(session.SegsInFlight) > 0 &&
-			time.Since(session.SegsInFlight[0].startTime) < time.Duration(durMult)*session.SegsInFlight[0].segDur &&
-			!includesSession(exclude, session) {
-			// Re-use last session if oldest segment is in-flight for < durMult * segDur
-			return session
+		// A session in the exclusion list is not selectable
+		if includesSession(exclude, session) {
+			continue
+		}
+
+		// A session with segments in flight might be selectable under certain conditions
+		if len(session.SegsInFlight) > 0 {
+			// The 0th segment in the slice is the oldest segment in flight since segments are appended
+			// to the slice as they are sent out
+			oldestSegInFlight := session.SegsInFlight[0]
+			timeInFlight := time.Since(oldestSegInFlight.startTime)
+			// durMult can be tuned by the caller to tighten/relax the maximum in flight time for the oldest segment
+			maxTimeInFlight := time.Duration(durMult) * oldestSegInFlight.segDur
+
+			if timeInFlight < maxTimeInFlight {
+				return session
+			}
 		}
 	}
 	return nil
--- a/server/broadcast.go
+++ b/server/broadcast.go
@@ -212,6 +212,14 @@ func selectSession(sessions []*BroadcastSession, exclude []*BroadcastSession, du
 			continue
 		}
 
+		// A session without any segments in flight and that has a latency score that meets the selector
+		// threshold is selectable
+		if len(session.SegsInFlight) == 0 {
+			if session.LatencyScore > 0 && session.LatencyScore <= SELECTOR_LATENCY_SCORE_THRESHOLD {
+				return session
+			}
+		}
+
 		// A session with segments in flight might be selectable under certain conditions
 		if len(session.SegsInFlight) > 0 {
 			// The 0th segment in the slice is the oldest segment in flight since segments are appended
@@ -352,6 +360,13 @@ func (sp *SessionPool) completeSession(sess *BroadcastSession) {
 			// we will return it later in transcodeSegment() once all in-flight segs downloaded
 			return
 		}
+
+		// If the latency score meets the selector threshold, we skip giving the session back to the selector
+		// because we consider it for re-use in selectSession()
+		if sess.LatencyScore > 0 && sess.LatencyScore <= SELECTOR_LATENCY_SCORE_THRESHOLD {
+			return
+		}
+
 		sp.sel.Complete(sess)
 	}
 }
--- a/server/broadcast_test.go
+++ b/server/broadcast_test.go
@@ -446,6 +446,35 @@ func TestSelectSession_MultipleInFlight2(t *testing.T) {
 	assert.Equal(1, orchInfoCalled)
 }
 
+func TestSelectSession_NoSegsInFlight(t *testing.T) {
+	assert := assert.New(t)
+
+	sess := &BroadcastSession{}
+	sessList := []*BroadcastSession{sess}
+
+	// Session has segs in flight
+	sess.SegsInFlight = []SegFlightMetadata{
+		{startTime: time.Now().Add(time.Duration(-1) * time.Second), segDur: 1 * time.Second},
+	}
+	s := selectSession(sessList, nil, 1)
+	assert.Nil(s)
+
+	// Session has no segs in flight, latency score = 0
+	sess.SegsInFlight = nil
+	s = selectSession(sessList, nil, 1)
+	assert.Nil(s)
+
+	// Session has no segs in flight, latency score > SELECTOR_LATENCY_SCORE_THRESHOLD
+	sess.LatencyScore = SELECTOR_LATENCY_SCORE_THRESHOLD + 0.001
+	s = selectSession(sessList, nil, 1)
+	assert.Nil(s)
+
+	// Session has no segs in flight, latency score > 0 and < SELECTOR_LATENCY_SCORE_THRESHOLD
+	sess.LatencyScore = SELECTOR_LATENCY_SCORE_THRESHOLD - 0.001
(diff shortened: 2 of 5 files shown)


server/broadcast_test.go
```diff
@@ -704,7 +733,9 @@ func TestTranscodeSegment_CompleteSession(t *testing.T) {
 	// Create stub server
 	ts, mux := stubTLSServer()
 	defer ts.Close()
+	transcodeDelay := 100 * time.Millisecond
```
leszko, line 736: I'd try to make unit tests time-dependent, because the test is inherently flaky. Especially that 100ms is very low time, so this test may fail in many environments.

leszko (approved):
Added one comment, other than that LGTM.

And I'm also in favor of doing some refactoring, but as a separate PR.

server/broadcast.go
```diff
@@ -207,11 +207,31 @@ func removeSessionFromList(sessions []*BroadcastSession, sess *BroadcastSession)
 
 func selectSession(sessions []*BroadcastSession, exclude []*BroadcastSession, durMult int) *BroadcastSession {
 	for _, session := range sessions {
-		if len(session.SegsInFlight) > 0 &&
-			time.Since(session.SegsInFlight[0].startTime) < time.Duration(durMult)*session.SegsInFlight[0].segDur &&
-			!includesSession(exclude, session) {
-			// Re-use last session if oldest segment is in-flight for < durMult * segDur
-			return session
+		// A session in the exclusion list is not selectable
+		if includesSession(exclude, session) {
```
thomshutt, line 211: Even though it's a side effect of your change, I really like the readability of having these broken out into distinct checks

thomshutt (approved):

server/broadcast.go
yondonfu: Glad you agree that this is more readable! TBH I was having trouble parsing the one-liner with multiple logical operators so felt compelled to do the small refactor.

server/broadcast_test.go
yondonfu: Agreed that throwing in these types of sleeps should be avoided, but in this case...

1. I needed to manipulate the value of the `LatencyScore` field calculated [here](https://github.com/livepeer/go-livepeer/blob/67c582922a398746f56aaa4868e143628064b56f/server/segment_rpc.go#L634). I can directly control the value of `seg.Duration`, but if `tookAllDur` is determined by the amount of time that elapses during the request so if `tookAllDur` is 0 or a value really close to 0 regardless of what I set as `seg.Duration` I could end up with a 0 value for `LatencyScore`. So, AFAIK adding a sleep in the `/segment` handler that is triggered during the request was the easiest way to also manipulate the value of `tookAllDur`.
2. I actually don't think the test will be flaky with the updates (see the snippet below) to the tests in [8cbd6a2](https://github.com/livepeer/go-livepeer/pull/2666/commits/8cbd6a202ea4ea2df060f97810b6d874b9c7e03b) (I rebased + force-pushed) because the values of `tookAllDur` and `seg.Duration` are set so that we can always guarantee that the calculated `LatencyScore` is what we want for the assertions we are making (i.e. a score < `SELECTOR_LATENCY_SCORE_THRESHOLD` and a score > `SELECTOR_LATENCY_SCORE_THRESHOLD`).

```
segDurMultiplier := 100.0
// The seg duration that meets the latency score threshold
// A larger seg duration gives a smaller/better latency score
// A smaller seg duration gives a larger/worse latency score
segDurLatencyScoreThreshold := transcodeDelay.Seconds() / SELECTOR_LATENCY_SCORE_THRESHOLD
// The seg duration that is better than the threshold (i.e. "fast")
segDurFastLatencyScore := segDurLatencyScoreThreshold * segDurMultiplier
// The seg duration that is worse than the threshold (i.e "slow")
segDurSlowLatencyScore := segDurLatencyScoreThreshold / segDurMultiplier
```

yondonfu:
> Outside of this PR I think it would be a good idea to do a refactor to move as much selection logic (i.e. last session tracking, segment in flight tracking, etc.) into MinLSSelector in selection.go to keep that logic encapsulated in a single place. This felt like a hefty change though so I opted not to do that refactor in this PR.

Tracking in https://github.com/livepeer/go-livepeer/issues/2669

PR #2686, opened 2022-12-09
Control scene classification with T command line args

cyberj0g:
1. Add `-detectionSampleRate` parameter to control non-stream specific scene classification on T
2. Log scene classification results on B, even if not requested for the stream
3. Minor refactoring and cleanup

### DIFF
--- a/core/transcoder.go
+++ b/core/transcoder.go
@@ -8,6 +8,7 @@ import (
 	"flag"
 	"fmt"
 	"io/ioutil"
+	"math"
 	"os"
 	"path/filepath"
 	"strconv"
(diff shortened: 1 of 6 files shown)


server/broadcast.go
```diff
@@ -1013,9 +1013,30 @@ func transcodeSegment(ctx context.Context, cxn *rtmpConnection, seg *stream.HLSS
 			}
 			return nil, info, err
 		}
+		// log all received detection results
+		if monitor.Enabled {
+			if len(res.Detections) > 0 {
+				for _, detection := range res.Detections {
+					switch x := detection.Value.(type) {
+					case *net.DetectData_SceneClassification:
```
thomshutt, line 1021: Worth logging in a `default` case to catch other cases?

core/transcoder.go
```diff
@@ -41,6 +42,36 @@ func NewUnrecoverableError(err error) UnrecoverableError {
...
+			actualProfile.(*ffmpeg.SceneClassificationProfile).SampleRate = uint(math.Min(float64(presetSampleRate),
+				float64(requestedSampleRate)))
+			// copy other fields from default AI capability, as we don't yet support custom ones
+			actualProfile.(*ffmpeg.SceneClassificationProfile).ModelPath = DetectorProfile.(*ffmpeg.SceneClassificationProfile).ModelPath
+			actualProfile.(*ffmpeg.SceneClassificationProfile).Input = DetectorProfile.(*ffmpeg.SceneClassificationProfile).Input
+			actualProfile.(*ffmpeg.SceneClassificationProfile).Output = DetectorProfile.(*ffmpeg.SceneClassificationProfile).Output
+			actualProfile.(*ffmpeg.SceneClassificationProfile).Classes = DetectorProfile.(*ffmpeg.SceneClassificationProfile).Classes
+		}
+	}
+	if actualProfile != nil && actualProfile.(*ffmpeg.SceneClassificationProfile).SampleRate < math.MaxUint32 {
```
thomshutt, line 66: What's the reason for comparison to `MaxUint32` rather than having it be zero?

core/transcoder.go
```diff
@@ -41,6 +42,36 @@ func NewUnrecoverableError(err error) UnrecoverableError {
 
 var WorkDir string
 
+func setEffectiveDetectorConfig(md *SegTranscodingMetadata) {
+	aiEnabled := DetectorProfile != nil
+	actualProfile := DetectorProfile
+	if aiEnabled {
+		presetSampleRate := DetectorProfile.(*ffmpeg.SceneClassificationProfile).SampleRate
+		if md.DetectorEnabled && len(md.DetectorProfiles) == 1 {
```
thomshutt, line 50: Is there a case where there can be more than one DetectorProfiles?

thomshutt (approved):
This all looks good, left a few questions that were more for my understanding than problems with the PR

core/transcoder.go
cyberj0g: Yes, these structures were designed with an aim to have multiple AI capabilities some day.

core/transcoder.go
cyberj0g: That value ends up in something like `select=not(mod(Nframe, SampleRate))` in Ffmpeg filter, so we don't want to have 0 here. Hence, MAX as default value.

core/transcoder.go
thomshutt: Yep for sure, I was more checking that there was no chance we'd unintentionally hit `md.DetectorEnabled && len(md.DetectorProfiles) > 1` at some point

core/transcoder.go
thomshutt: 👍  Gotcha

PR #2695, opened 2022-12-19
Enable tensorflow for content detection in a production build

cyberj0g:
Changes:
- `install_ffmpeg.sh` now installs Tensorflow library and enables it for Ffmpeg without `experimental` BUILD_TAGS (no side effects when running in non-Nvidia environment)
- `Dockerfile` copies Tensorflow distribution files to final container, because libtensorflow.so is loaded at runtime by Ffmpeg DNN filter and is not linked statically inside `livepeer` binary, like the rest of dependencies

Kudos to @hjpotter92 for helping to figure out Docker stuff.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -72,8 +72,8 @@ RUN apt update && apt install -yqq software-properties-common curl apt-transport
 
 RUN LIBTENSORFLOW_VERSION=2.6.3 \
             && curl -LO https://storage.googleapis.com/tensorflow/libtensorflow/libtensorflow-gpu-linux-x86_64-${LIBTENSORFLOW_VERSION}.tar.gz \
-            && sudo tar -C /usr/local -xzf libtensorflow-gpu-linux-x86_64-${LIBTENSORFLOW_VERSION}.tar.gz \
-            && sudo ldconfig
+            && tar -C /usr/local -xzf libtensorflow-gpu-linux-x86_64-${LIBTENSORFLOW_VERSION}.tar.gz \
+            && ldconfig
 
 RUN	ldconfig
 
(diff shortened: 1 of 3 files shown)


docker/Dockerfile
```diff
@@ -51,24 +57,17 @@ RUN	go mod download
 
 COPY	.	.
 
-ARG	BUILD_TAGS
```
hjpotter92, line 54: i remember why i had it put so low in `Dockerfile`. if we move this to start of the file, each run of `docker build` invalidates the existing cache, and takes quite a long time to build.

i think moving it to just before `RUN ./install_ffmpeg.sh` line should do the trick. (instead of the very beginning, where we're just installing curl/apt/grpc etc)

docker/Dockerfile
```diff
@@ -51,24 +57,17 @@ RUN	go mod download
...
 ENV	NVIDIA_DRIVER_CAPABILITIES=all
 
 COPY --from=build	/build/	/usr/local/bin/
 COPY --from=build	/usr/bin/grpc_health_probe	/usr/local/bin/grpc_health_probe
 COPY --from=build	/src/tasmodel.pb	/tasmodel.pb
 COPY --from=build	/usr/share/misc/pci.ids	/usr/share/misc/pci.ids
-
-RUN	ldconfig
+# libtensorflow.so is required at runtime, because Ffmpeg DNN filter loads it dynamically
+COPY --from=build	/tf/	/usr/local/
```
hjpotter92, line 72: do we not need a run of `ldconfig`?

docker/Dockerfile
cyberj0g: Got it, moved it just before `install_ffmpeg.sh`.

docker/Dockerfile
cyberj0g: I don't think we need to run it, when copying libs to system location. It should be run when we modify search paths in `/etc/ld.so.conf`.

thomshutt (approved):

PR #2707, opened 2023-01-03
Fix: remote transcoders quietly getting dropped from selection

### LINKED ISSUE #2706: Remote T's can get removed from the `remoteTranscoders` list if any T reconnects
## Describe the bug
Whenever a T reconnects, they get a new entry in the `remoteTranscoders` list of transcoder objects held by the `RemoteTranscoderManager`
Their 'expired' entry in the `remoteTranscoders` list only gets removed when they later get selected for a session:
https://github.com/livepeer/go-livepeer/blob/e2c46a1e30ca3c5aafcc1a4f11f55523235da871/core/orchestrator.go#L968-L976

Up to this point everything looks alright (although I wonder why does the Transcoder get removed from `liveTranscoders` immediately, but only gets removed from `remoteTranscoders` on selection)

Anyway, the error happens in the `removeFromRemoteTranscoders` function, where it splices the expired entry out of the list
https://github.com/livepeer/go-livepeer/blob/e2c46a1e30ca3c5aafcc1a4f11f55523235da871/core/orchestrator.go#L914-L938

It creates the new list as 
```
newRemoteTs = remoteTranscoders[i-1 : i] 
newRemoteTs = append(newRemoteTs, remoteTranscoders[i+1:]...) 
```
So if `i` is higher than 1, any transcoder in this list with an index lower than `i-1` is going to get dropped from the list of remote transcoders!

In all regards the T is still considered a live transcoder connected to the Orchestrator, so no error pops up on the O or T side. They just won't receive any more streams

## To Reproduce
Have multiple remote T's connected in a split O/T setup. Have one (or more) T's reconnect and wait for a few streams to come in. At some point you will see one T not receiving any more streams

stronk-dev:
### What does this pull request do?

Fixes a bug in split O/T setups, where a Transcoder can get dropped from selection alltogether, without any indication on the O or T side

See https://github.com/livepeer/go-livepeer/issues/2706 for more details

### Specific updates

Seems to have been introduced in https://github.com/livepeer/go-livepeer/commit/bf5397d210624b19c643bfbc506f40d8ae1cb325 so a looong time ago

### How did you test each of these updates

Currently running the fix in production on the `video miner` pool. So far no connected transcoder has been dropped from the `remoteTranscoders` set as of yet

### Does this pull request close any open issues?

Fixes #2706

### DIFF
--- a/core/orch_test.go
+++ b/core/orch_test.go
@@ -373,31 +373,39 @@ func TestRemoveFromRemoteTranscoders(t *testing.T) {
 	remoteTranscoderList := []*RemoteTranscoder{}
 	assert := assert.New(t)
 
-	// Create 4 tanscoders
-	tr := make([]*RemoteTranscoder, 4)
-	for i := 0; i < 4; i++ {
+	// Create 6 transcoders
+	tr := make([]*RemoteTranscoder, 5)
+	for i := 0; i < 5; i++ {
 		tr[i] = &RemoteTranscoder{addr: "testAddress" + strconv.Itoa(i)}
 	}
 
 	// Add to list
 	remoteTranscoderList = append(remoteTranscoderList, tr...)
-	assert.Len(remoteTranscoderList, 4)
+	assert.Len(remoteTranscoderList, 5)
 
 	// Remove transcoder froms head of the list
 	remoteTranscoderList = removeFromRemoteTranscoders(tr[0], remoteTranscoderList)
 	assert.Equal(remoteTranscoderList[0], tr[1])
 	assert.Equal(remoteTranscoderList[1], tr[2])
 	assert.Equal(remoteTranscoderList[2], tr[3])
-	assert.Len(remoteTranscoderList, 3)
+	assert.Equal(remoteTranscoderList[3], tr[4])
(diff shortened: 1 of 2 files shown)


cyberj0g (changes requested):
That's a nasty one, great finding! Before merging, could you please:
- check why [test](https://github.com/livepeer/go-livepeer/blob/e58fa6d474fec44abcd5ce6757e849644d7431ec/core/orch_test.go#L372) of `removeFromRemoteTranscoders` doesn't catch that and update it
- rebase cleanly on top of `master`

stronk-dev:
Two issues with the `TestRemoveFromRemoteTranscoders`:
- At the point where it says it is removing the middle element, it is actually removing the final element, for which exists a separate code path
- The bug only drops Transcoders with an index _lower than_ `i-1` and the size of the list when testing removal of a middle Transcoder is `3`. Thus, when removing the Transcoder at index `1`, the Transcoder at index `0` is still safe. The test would have to be done by removing a Transcoder at index `2` or higher where the index is not the last Transcoder in the list

I'll modify the test to reflect these changes

cyberj0g (approved):
LGTM!

Issue #2661, opened 2022-11-21
senderNonce replay prevention on O that allows for out-of-order senderNonce values in tickets


yondonfu:
**Is your feature request related to a problem? Please describe.**

This issue occurs whenever multiple segments are pushed in parallel or in quick succession with minimal time delay to B via the `/live` endpoint (i.e. HTTP push ingest). This can occur for a livestream if the client that is pushing to B (i.e. MistProcLivepeer in Catalyst) pushes really short segments really quickly or in parallel by simultaneously retrying segment N while pushing segment N + 1. This can occur for VOD file if the client is configured to push multiple segments in parallel to B (i.e. catalyst-api in Catalyst).

The issue can be repro'd by checking out the following branch and running a parallel segment push e2e test:

```
git checkout yf/fix-sender-nonce
go test -count=1 -v -run TestHTTPPushBroadcaster ./test/e2e
```

The test should fail with output that looks like this:

```
E1121 14:43:59.238132   29209 orchestrator.go:179] manifestID=559e2858f20f7ec9c6f7 seqNo=5 orchSessionID=c37fb7db sender=0xFd2f241e6b8Ca65be71f3A23d2badB4ED8BaA5a9 clientIP=127.0.0.1 Error receiving ticket sessionID=c37fb7db recipientRandHash=0c4c1e4ba5bb8d49e4354313636ac7d598810ceb20f15521309c3886bbb2aa49 senderNonce=4: invalid ticket senderNonce sender=0xFd2f241e6b8Ca65be71f3A23d2badB4ED8BaA5a9 nonce=4 highest=5
E1121 14:43:59.238138   29209 segment_rpc.go:101] manifestID=559e2858f20f7ec9c6f7 seqNo=5 orchSessionID=c37fb7db sender=0xFd2f241e6b8Ca65be71f3A23d2badB4ED8BaA5a9 clientIP=127.0.0.1 error processing payment: invalid ticket senderNonce sender=0xFd2f241e6b8Ca65be71f3A23d2badB4ED8BaA5a9 nonce=4 highest=5
[...]

PR #2710, opened 2023-01-05
Allow for out of order nonce

### LINKED ISSUE #2614: Tons of `Invalid Ticket senderNonce` for the past few months

cyberj0g:
Implementation of https://github.com/livepeer/go-livepeer/issues/2661 to fix https://github.com/livepeer/go-livepeer/issues/2614 based on introducing sorted hash map to store seen nonce values. Also added @yondonfu's test from corresponding branch, which now passes.

### DIFF
--- a/pm/recipient.go
+++ b/pm/recipient.go
@@ -4,6 +4,7 @@ import (
 	"crypto/hmac"
 	"crypto/rand"
 	"crypto/sha256"
+	"github.com/elliotchance/orderedmap"
 	"math/big"
 	"sync"
 
@@ -20,6 +21,9 @@ var errInsufficientSenderReserve = errors.New("insufficient sender reserve")
 
 // maxWinProb = 2^256 - 1
 var maxWinProb = new(big.Int).Sub(new(big.Int).Lsh(big.NewInt(1), 256), big.NewInt(1))
+
+// max number of sender nonces for a given recipient random hash
+var maxSenderNonces = 50
 
 var paramsExpirationBlock = big.NewInt(10)
 var paramsExpiryBuffer = int64(1)
@@ -85,10 +89,7 @@ type recipient struct {
 	secret       [32]byte
 	maxfacevalue *big.Int
 
-	senderNonces map[string]*struct {
-		nonce           uint32
-		expirationBlock *big.Int
-	}
+	senderNonces     map[string]*orderedmap.OrderedMap
 	senderNoncesLock sync.Mutex
 
 	cfg TicketParamsConfig
@@ -123,12 +124,9 @@ func NewRecipientWithSecret(addr ethcommon.Address, broker Broker, val Validator
 		addr:         addr,
 		secret:       secret,
 		maxfacevalue: big.NewInt(0),
-		senderNonces: make(map[string]*struct {
-			nonce           uint32
-			expirationBlock *big.Int
-		}),
-		cfg:  cfg,
-		quit: make(chan struct{}),
+		senderNonces: make(map[string]*orderedmap.OrderedMap),
+		cfg:          cfg,
+		quit:         make(chan struct{}),
 	}
 }
 
@@ -363,16 +361,21 @@ func (r *recipient) updateSenderNonce(rand *big.Int, ticket *Ticket) error {
 	defer r.senderNoncesLock.Unlock()
 
 	randStr := rand.String()
-	sn, ok := r.senderNonces[randStr]
-	if ok && ticket.SenderNonce <= sn.nonce {
-		return errors.Errorf("invalid ticket senderNonce sender=%v nonce=%v highest=%v", ticket.Sender.Hex(), ticket.SenderNonce, sn.nonce)
+	nonceToExpBlock, randKeySeen := r.senderNonces[randStr]
+	if randKeySeen {
+		_, nonceSeen := nonceToExpBlock.Get(ticket.SenderNonce)
+		if nonceSeen {
+			return errors.Errorf("invalid ticket senderNonce: already seen sender=%v nonce=%v", ticket.Sender.Hex(), ticket.SenderNonce)
+		}
+	} else {
+		r.senderNonces[randStr] = orderedmap.NewOrderedMap()
 	}
-
-	r.senderNonces[randStr] = &struct {
-		nonce           uint32
-		expirationBlock *big.Int
-	}{ticket.SenderNonce, ticket.ParamsExpirationBlock}
-
+	// add new nonce
+	r.senderNonces[randStr].Set(ticket.SenderNonce, ticket.ParamsExpirationBlock)
+	// delete first seen nonce, if ordered map exceeds max size
+	if r.senderNonces[randStr].Len() > maxSenderNonces {
+		r.senderNonces[randStr].Delete(r.senderNonces[randStr].Front().Key)
+	}
 	return nil
 }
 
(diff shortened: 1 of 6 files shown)


pm/recipient_test.go
```diff
@@ -224,8 +225,8 @@ func TestReceiveTicket_ValidNonWinningTicket(t *testing.T) {
 	recipientRand := genRecipientRand(sender, secret, params)
 	senderNonce := r.(*recipient).senderNonces[recipientRand.String()]
 
-	if senderNonce.nonce != newSenderNonce {
-		t.Errorf("expected senderNonce to be %d, got %d", newSenderNonce, senderNonce.nonce)
+	if senderNonce.Front().Key != newSenderNonce {
```
thomshutt: This is probably me misunderstanding, but I'd expect this be `.Back()` to look at the most recent one

cyberj0g: Right, `Back()` would be most recent nonce, but in this test we test only one ticket anyway.

thomshutt (approved):
LGTM, probably good to wait for Yondon's approval too since he originally documented this issue

pm/recipient.go
```diff
@@ -363,16 +361,21 @@ func (r *recipient) updateSenderNonce(rand *big.Int, ticket *Ticket) error {
...
+			return errors.Errorf("invalid ticket senderNonce: already seen sender=%v nonce=%v", ticket.Sender.Hex(), ticket.SenderNonce)
+		}
+	} else {
+		r.senderNonces[randStr] = orderedmap.NewOrderedMap()
+	}
+	// add new nonce
+	r.senderNonces[randStr].Set(ticket.SenderNonce, ticket.ParamsExpirationBlock)
+	// delete first seen nonce, if ordered map exceeds max size
+	if r.senderNonces[randStr].Len() > maxSenderNonces {
+		r.senderNonces[randStr].Delete(r.senderNonces[randStr].Front().Key)
```
yondonfu: Seems simpler to just return an error if the max size is hit. If we delete the first seen nonce, then that means that in theory B could re-use that deleted nonce after the max size is hit and if we return an error if the max size is hit we at least wouldn't need to reason about this scenario. This would mean that once the max size is hit, B would *need* a new set of ticket params with a fresh `recipientRandHash`, but as mentioned in the original issue at the moment this should happen after B receives a response for each transcoded segment.

If we don't need to delete the first seen senderNonce value I think that would also simplify the data struct impl s.t. we might not need to depend on the custom ordered map package.

WDYT?

pm/recipient.go
```diff
@@ -363,16 +361,21 @@ func (r *recipient) updateSenderNonce(rand *big.Int, ticket *Ticket) error {
...
+	if randKeySeen {
+		_, nonceSeen := nonceToExpBlock.Get(ticket.SenderNonce)
+		if nonceSeen {
+			return errors.Errorf("invalid ticket senderNonce: already seen sender=%v nonce=%v", ticket.Sender.Hex(), ticket.SenderNonce)
+		}
+	} else {
+		r.senderNonces[randStr] = orderedmap.NewOrderedMap()
+	}
+	// add new nonce
+	r.senderNonces[randStr].Set(ticket.SenderNonce, ticket.ParamsExpirationBlock)
```
yondonfu: I don't think we need to track an expiration block per senderNonce - we only need to track an expiration block per recipientRandHash (aka `randStr` in this function). And we only need to track seen senderNonce values associated with a particular recipientRandHash until that recipientRandHash expires.

I think this would simplify the state tracking since we can track an expiration block per recipientRandHash instead of an expiration block for every senderNonce value. With the current impl I think the map ends up looking like `{ key: 1 value: X, key: 2 value: X, key 3 value: X }`.

pm/recipient.go
```diff
@@ -393,9 +396,15 @@ func (r *recipient) senderNoncesCleanupLoop() {
 			glog.Error(err)
 		case latestL1Block := <-sink:
 			r.senderNoncesLock.Lock()
-			for recipientRand, sn := range r.senderNonces {
-				if sn.expirationBlock.Cmp(latestL1Block) <= 0 {
-					delete(r.senderNonces, recipientRand)
+			for recipientRand, nonceMap := range r.senderNonces {
+				for _, key := range nonceMap.Keys() {
```
yondonfu: If we track a single expiration block per recipientRandHash then we wouldn't need the nested loop here and instead could just clear all senderNonce values for the recipientRandHash once the expirationBlock is reached.

Quintendos:
Ivan to fix up Yondon's comments and then ready to merge.

cyberj0g:
Thanks for reviews @yondonfu @thomshutt, if we don't want FIFO and don't need to track expiration block number per nonce, that makes the logic simpler and removes sorted hash map dependency. I still used map instead of slice to have O(1) retrieval.

pm/recipient.go
```diff
@@ -363,16 +366,24 @@ func (r *recipient) updateSenderNonce(rand *big.Int, ticket *Ticket) error {
 	defer r.senderNoncesLock.Unlock()
 
 	randStr := rand.String()
-	sn, ok := r.senderNonces[randStr]
-	if ok && ticket.SenderNonce <= sn.nonce {
-		return errors.Errorf("invalid ticket senderNonce sender=%v nonce=%v highest=%v", ticket.Sender.Hex(), ticket.SenderNonce, sn.nonce)
+	senderNonce, randKeySeen := r.senderNonces[randStr]
```
yondonfu:
```suggestion
	senderNonces, randKeySeen := r.senderNonces[randStr]
```

Nit: The plural feels a bit clearer here since the map is tracking a collection of nonces as opposed to a single one.

pm/recipient.go
```diff
@@ -86,7 +89,7 @@ type recipient struct {
 	maxfacevalue *big.Int
 
 	senderNonces map[string]*struct {
-		nonce           uint32
+		nonceSeen       map[uint32]byte
```
yondonfu: Nit: Is there an advantage of using `byte` here instead of a `bool`?

IIUC a `bool` should be 1 byte which would be the same size as a `byte` [1] so I don't think there are any storage benefits of using `byte`. If this is the case, I suggest using a `bool` as I think using true/false values will be clearer for code readability than using 0/1.

[1] Can verify using `unsafe.Sizeof()` on a `bool` var and a `byte` var.

pm/recipient.go
```diff
@@ -363,16 +366,24 @@ func (r *recipient) updateSenderNonce(rand *big.Int, ticket *Ticket) error {
...
+		}{make(map[uint32]byte), ticket.ParamsExpirationBlock}
 	}
-
-	r.senderNonces[randStr] = &struct {
-		nonce           uint32
-		expirationBlock *big.Int
-	}{ticket.SenderNonce, ticket.ParamsExpirationBlock}
-
+	// check nonce map size
+	if len(r.senderNonces[randStr].nonceSeen) > maxSenderNonces-1 {
```
yondonfu:
```suggestion
	if len(r.senderNonces[randStr].nonceSeen) >= maxSenderNonces {
```

Nit: IMO feels a bit clearer (no arithmetic needed) and conveys "if we are at the max size or for some reason exceeded it, return an error".

yondonfu (approved):
Left a few nits. Feel free to address prior to merging.

Other than that LGTM.

PR #2713, opened 2023-01-06
Add support for keyfiles with -ethKeystorePath, update flag descriptions, flagset output to stdout

### LINKED ISSUE #2689: cli help is printed to stderr instead of stdout
**Describe the bug**

cli help should print to stdout, so I can pipe output to grep, to find the option info I am interested in.
However, currently it's printed to stderr, and grep doesn't work by default.

**To Reproduce**

grep help msg like this:
```
livepeer -h | grep ethKeystorePath
```

**Expected behavior**

I should only see the option I grepped, but the whole help msg is still printed to screen.

### LINKED ISSUE #2690: cli help didn't mention `-ethPassword` and `-orchSecret` can be file path
**Describe the bug**

In cli help, help text for `-ethPassword` and `-orchSecret` are asking users to provide the secrets in plain text in cli, which is not secure.
After digging into code, I noticed that both of them can actually be a file.
This is good feature, but it's important to mention it in cli,  many potential secrets leaks can be avoid.

**To Reproduce**
1. cli tells you to provide them as string:
```
livepeer -h
...
  -ethPassword string
    	Password for existing Eth account address
...
  -orchSecret string
    	Shared secret with the orchestrator as a standalone transcoder
...
```

2. but they can both be a file, with secret in it, e.g.:

```
-ethPassword  /etc/livepeer/ethPassword.txt -orchSecret /etc/livepeer/orchSecret.txt
```
This is much more secure.

Please mention this feature in cli help.

**Additional context**

The magic happens in this `ReadFromFile`  func:

https://github.com/livepeer/go-livepeer/blob/7be4c3b518544f6cf29f79872e545b1a34cf72d9/common/readfromfile.go#L12

It will first read the string as file, if failed, return string itself. Otherwise return content of the file.

### LINKED ISSUE #2691: Usage for `-ethKeystorePath` option is confusing
**Describe the bug**

The name and cli help for this option is unclear and misleading.

**To Reproduce**

1. In cli help:

```
  -ethKeystorePath string
    	Path for the Eth Key
```

2. When I read this, I am not sure what I should provide.

From the name `ethKeystorePath`,  it looks like I should provide a dir path like:

```
/root/.lpData/keystore
```

But, from the help msg, it looks like I should provide a file path, like:
```
/root/.lpData/keystore/UTC--2022-12-03T10-31-30.841732443Z--6710ff8ab1d776097a6892c181964683442507b2`
```
This is too lengthy, I believe most users would rather not to provide it.

3.  actual behavior

If I provide `/root/.lpData/keystore`,   the actually dir code will look into will be `/root/.lpData/`, which is unexpcted.
To get this right,  I must provide the path with trailing `/`:  `/root/.lpData/keystore/`.

If I provide the lengthy key file path,  code will look into the parent dir, the actual key file name is ignored.
If you have multiple keys in that dir, you will need to specify it again, with `-ethAcctAddr`.  Also unexpected.

related source code:

https://github.com/livepeer/go-livepeer/blob/7be4c3b518544f6cf29f79872e545b1a34cf72d9/cmd/livepeer/starter/starter.go#L525-L530

If the path exists, either it's dir or file, the path is split, which is not reasonable.

**Expected behavior**

This option should be a path to a `keystore` dir,  no matter it has trailing `/` or not.

eliteprox:
- Updates various flag descriptions that a file can be used in place of a string.
- Changes the default flagset output to `os.StdOut`. This has the effect of enabling users to grep --help output, 
- Add support for keystore file with `-ethKeystorePath`. This can be used alternatively to `ethAcctAddr`
  - ETH address is parsed from keyfile. `-ethAcctAddr` is optional, but if provided the address must match with the one found in the keyfile.

- Updated flag descriptions for `OrchSecret`, `EthAcctAddr`, `EthPassword`, `EthKeystorePath` and `PricePerBroadcaster`
- Changed command line output for flagset to os.StdOut
- Parse ETH address from keyfile when a file is provided to `-ethKeystorePath`
- Validates ethAcctAddr and keystore address match or that ethAcctAddr is not provided, before using address from keystore
- Added test coverage

I tested the `ethKeystorePath` and `ethAcctAddr` flags in the following ways:
- With a directory and with a valid file. Eth account is properly found from keystore
- When set to a directory: `keystoreDir` was set to same path and `ethAcctAddr` flag still works for selecting eth account.
- When set to a valid key file: `keystoreDir` was set to parent directory of the file and ETH address is parsed. If `-ethAcctAddr` flag is also provided, the address parsed from keyfile must match the ethAcctAddr, otherwise error: `-ethKeystorePath and -ethAcctAddr were both provided, but ethAcctAddr does not match the address found in keystore`
- With a missing file or directory: `provided -ethKeystorePath was not found`
- With invalid json in file: `error opening keystore`

fix #2689 cli help is printed to stderr instead of stdout
fix #2690 cli help didn't mention -ethPassword and -orchSecret can be file path
fix #2691 Usage for -ethKeystorePath option is confusing

**Checklist:**

### DIFF
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -144,7 +144,7 @@ func parseLivepeerConfig() starter.LivepeerConfig {
 	// Onchain:
 	cfg.EthAcctAddr = flag.String("ethAcctAddr", *cfg.EthAcctAddr, "Existing Eth account address. For use when multiple ETH accounts exist in keystore directory")
 	cfg.EthPassword = flag.String("ethPassword", *cfg.EthPassword, "Password for existing Eth account address or path to file")
-	cfg.EthKeystorePath = flag.String("ethKeystorePath", *cfg.EthKeystorePath, "Path to ETH keystore directory or keyfile. Defaults to [dataDir]\keystore. If keyfile, overrides -ethAcctAddr and uses parent directory")
+	cfg.EthKeystorePath = flag.String("ethKeystorePath", *cfg.EthKeystorePath, "Path to ETH keystore directory or keyfile. If keyfile, overrides -ethAcctAddr and uses parent directory")
 	cfg.EthOrchAddr = flag.String("ethOrchAddr", *cfg.EthOrchAddr, "ETH address of an on-chain registered orchestrator")
 	cfg.EthUrl = flag.String("ethUrl", *cfg.EthUrl, "Ethereum node JSON-RPC URL")
 	cfg.TxTimeout = flag.Duration("transactionTimeout", *cfg.TxTimeout, "Amount of time to wait for an Ethereum transaction to confirm before timing out")
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -142,9 +142,9 @@ func parseLivepeerConfig() starter.LivepeerConfig {
 	cfg.DetectionSampleRate = flag.Uint("detectionSampleRate", *cfg.DetectionSampleRate, "Run content detection automatically on every nth frame of each segment, independently of requested stream transcoding configuration.")
 
 	// Onchain:
-	cfg.EthAcctAddr = flag.String("ethAcctAddr", *cfg.EthAcctAddr, "Existing Eth account address. For use when multiple ETH accounts exist in keystore directory. Overridden when by providing a valid key file using -ethKeystorePath")
+	cfg.EthAcctAddr = flag.String("ethAcctAddr", *cfg.EthAcctAddr, "Existing Eth account address. For use when multiple ETH accounts exist in keystore directory. Overridden when key file provided with -ethKeystorePath")
 	cfg.EthPassword = flag.String("ethPassword", *cfg.EthPassword, "Password for existing Eth account address or path to file")
-	cfg.EthKeystorePath = flag.String("ethKeystorePath", *cfg.EthKeystorePath, "Path to keystore directory or path to ETH key file. If file provided, ethAcctAddr will be set and parent directory will be used for new accounts")
+	cfg.EthKeystorePath = flag.String("ethKeystorePath", *cfg.EthKeystorePath, "Path to ETH keystore directory or keyfile. If valid keyfile, overrides -ethAcctAddr and uses parent directory for keystore path")
 	cfg.EthOrchAddr = flag.String("ethOrchAddr", *cfg.EthOrchAddr, "ETH address of an on-chain registered orchestrator")
 	cfg.EthUrl = flag.String("ethUrl", *cfg.EthUrl, "Ethereum node JSON-RPC URL")
 	cfg.TxTimeout = flag.Duration("transactionTimeout", *cfg.TxTimeout, "Amount of time to wait for an Ethereum transaction to confirm before timing out")
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -144,7 +144,7 @@ func parseLivepeerConfig() starter.LivepeerConfig {
 	// Onchain:
 	cfg.EthAcctAddr = flag.String("ethAcctAddr", *cfg.EthAcctAddr, "Existing Eth account address. For use when multiple ETH accounts exist in keystore directory")
 	cfg.EthPassword = flag.String("ethPassword", *cfg.EthPassword, "Password for existing Eth account address or path to file")
-	cfg.EthKeystorePath = flag.String("ethKeystorePath", *cfg.EthKeystorePath, "Path to ETH keystore directory or keyfile. If valid keyfile, overrides -ethAcctAddr and uses parent directory for keystore path")
+	cfg.EthKeystorePath = flag.String("ethKeystorePath", *cfg.EthKeystorePath, "Path to ETH keystore directory or keyfile. Defaults to [dataDir]\keystore. If keyfile, overrides -ethAcctAddr and uses parent directory")
 	cfg.EthOrchAddr = flag.String("ethOrchAddr", *cfg.EthOrchAddr, "ETH address of an on-chain registered orchestrator")
 	cfg.EthUrl = flag.String("ethUrl", *cfg.EthUrl, "Ethereum node JSON-RPC URL")
 	cfg.TxTimeout = flag.Duration("transactionTimeout", *cfg.TxTimeout, "Amount of time to wait for an Ethereum transaction to confirm before timing out")
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -142,7 +142,7 @@ func parseLivepeerConfig() starter.LivepeerConfig {
 	cfg.DetectionSampleRate = flag.Uint("detectionSampleRate", *cfg.DetectionSampleRate, "Run content detection automatically on every nth frame of each segment, independently of requested stream transcoding configuration.")
 
 	// Onchain:
-	cfg.EthAcctAddr = flag.String("ethAcctAddr", *cfg.EthAcctAddr, "Existing Eth account address. For use when multiple ETH accounts exist in keystore directory. Overridden when key file provided with -ethKeystorePath")
+	cfg.EthAcctAddr = flag.String("ethAcctAddr", *cfg.EthAcctAddr, "Existing Eth account address. For use when multiple ETH accounts exist in keystore directory")
 	cfg.EthPassword = flag.String("ethPassword", *cfg.EthPassword, "Password for existing Eth account address or path to file")
 	cfg.EthKeystorePath = flag.String("ethKeystorePath", *cfg.EthKeystorePath, "Path to ETH keystore directory or keyfile. If valid keyfile, overrides -ethAcctAddr and uses parent directory for keystore path")
 	cfg.EthOrchAddr = flag.String("ethOrchAddr", *cfg.EthOrchAddr, "ETH address of an on-chain registered orchestrator")
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -142,9 +142,9 @@ func parseLivepeerConfig() starter.LivepeerConfig {
 	cfg.DetectionSampleRate = flag.Uint("detectionSampleRate", *cfg.DetectionSampleRate, "Run content detection automatically on every nth frame of each segment, independently of requested stream transcoding configuration.")
 
 	// Onchain:
-	cfg.EthAcctAddr = flag.String("ethAcctAddr", *cfg.EthAcctAddr, "Existing Eth account address")
+	cfg.EthAcctAddr = flag.String("ethAcctAddr", *cfg.EthAcctAddr, "Existing Eth account address. For use when multiple ETH accounts exist in keystore directory. Overridden when by providing a valid key file using -ethKeystorePath")
 	cfg.EthPassword = flag.String("ethPassword", *cfg.EthPassword, "Password for existing Eth account address or path to file")
-	cfg.EthKeystorePath = flag.String("ethKeystorePath", *cfg.EthKeystorePath, "Path for the Eth Key")
+	cfg.EthKeystorePath = flag.String("ethKeystorePath", *cfg.EthKeystorePath, "Path to keystore directory or path to ETH key file. If file provided, ethAcctAddr will be set and parent directory will be used for new accounts")
 	cfg.EthOrchAddr = flag.String("ethOrchAddr", *cfg.EthOrchAddr, "ETH address of an on-chain registered orchestrator")
 	cfg.EthUrl = flag.String("ethUrl", *cfg.EthUrl, "Ethereum node JSON-RPC URL")
 	cfg.TxTimeout = flag.Duration("transactionTimeout", *cfg.TxTimeout, "Amount of time to wait for an Ethereum transaction to confirm before timing out")
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -167,7 +167,7 @@ func parseLivepeerConfig() starter.LivepeerConfig {
 	// Unit of pixels for both O's basePriceInfo and B's MaxBroadcastPrice
 	cfg.PixelsPerUnit = flag.Int("pixelsPerUnit", *cfg.PixelsPerUnit, "Amount of pixels per unit. Set to '> 1' to have smaller price granularity than 1 wei / pixel")
 	cfg.AutoAdjustPrice = flag.Bool("autoAdjustPrice", *cfg.AutoAdjustPrice, "Enable/disable automatic price adjustments based on the overhead for redeeming tickets")
-	cfg.PricePerBroadcaster = flag.String("pricePerBroadcaster", *cfg.PricePerBroadcaster, `json list of price per broadcaster. or path to json config file. Example: {"broadcasters":[{"ethaddress":"address1","priceperunit":1000,"pixelsperunit":1},{"ethaddress":"address2","priceperunit":1200,"pixelsperunit":1}]}`)
+	cfg.PricePerBroadcaster = flag.String("pricePerBroadcaster", *cfg.PricePerBroadcaster, `json list of price per broadcaster or path to json config file. Example: {"broadcasters":[{"ethaddress":"address1","priceperunit":1000,"pixelsperunit":1},{"ethaddress":"address2","priceperunit":1200,"pixelsperunit":1}]}`)
 	// Interval to poll for blocks
 	cfg.BlockPollingInterval = flag.Int("blockPollingInterval", *cfg.BlockPollingInterval, "Interval in seconds at which different blockchain event services poll for blocks")
 	// Redemption service
--- a/cmd/livepeer/livepeer.go
+++ b/cmd/livepeer/livepeer.go
@@ -31,6 +31,7 @@ func main() {
 	vFlag := flag.Lookup("v")
 	//We preserve this flag before resetting all the flags.  Not a scalable approach, but it'll do for now.  More discussions here - https://github.com/livepeer/go-livepeer/pull/617
 	flag.CommandLine = flag.NewFlagSet(os.Args[0], flag.ExitOnError)
+	flag.CommandLine.SetOutput(os.Stdout)
 
 	// Help & Log
 	mistJson := flag.Bool("j", false, "Print application info as json")
@@ -127,7 +128,7 @@ func parseLivepeerConfig() starter.LivepeerConfig {
 	cfg.Orchestrator = flag.Bool("orchestrator", *cfg.Orchestrator, "Set to true to be an orchestrator")
 	cfg.Transcoder = flag.Bool("transcoder", *cfg.Transcoder, "Set to true to be a transcoder")
 	cfg.Broadcaster = flag.Bool("broadcaster", *cfg.Broadcaster, "Set to true to be a broadcaster")
-	cfg.OrchSecret = flag.String("orchSecret", *cfg.OrchSecret, "Shared secret with the orchestrator as a standalone transcoder")
+	cfg.OrchSecret = flag.String("orchSecret", *cfg.OrchSecret, "Shared secret with the orchestrator as a standalone transcoder or path to file")
 	cfg.TranscodingOptions = flag.String("transcodingOptions", *cfg.TranscodingOptions, "Transcoding options for broadcast job, or path to json config")
 	cfg.MaxAttempts = flag.Int("maxAttempts", *cfg.MaxAttempts, "Maximum transcode attempts")
 	cfg.SelectRandFreq = flag.Float64("selectRandFreq", *cfg.SelectRandFreq, "Frequency to randomly select unknown orchestrators (on-chain mode only)")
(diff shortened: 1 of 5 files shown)


joegraviton:
Thanks for fixing my 3 issues in batch :+1:

cmd/livepeer/starter/starter.go
```diff
@@ -5,8 +5,8 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"github.com/ethereum/go-ethereum/common/math"
 	"io/ioutil"
+	"math"
```
eliteprox, line 9: This is an intentional correction. Math should be from regular class, not ethereum package. It is used only here

https://github.com/livepeer/go-livepeer/blob/484fbe4091af9666b344ce2b824393099237de57/cmd/livepeer/starter/starter.go#L158

leszko:
@eliteprox Thanks for the PR. Mark as ready when you think you finalized it. Then, we'll review and comment on this.

eliteprox:
@leszko This PR is ready for review.

leszko:
> @leszko This PR is ready for review.

Perfect. Thank you. I'll review it.

cmd/livepeer/livepeer.go
```diff
@@ -141,9 +142,9 @@ func parseLivepeerConfig() starter.LivepeerConfig {
 	cfg.DetectionSampleRate = flag.Uint("detectionSampleRate", *cfg.DetectionSampleRate, "Run content detection automatically on every nth frame of each segment, independently of requested stream transcoding configuration.")
 
 	// Onchain:
-	cfg.EthAcctAddr = flag.String("ethAcctAddr", *cfg.EthAcctAddr, "Existing Eth account address")
-	cfg.EthPassword = flag.String("ethPassword", *cfg.EthPassword, "Password for existing Eth account address")
-	cfg.EthKeystorePath = flag.String("ethKeystorePath", *cfg.EthKeystorePath, "Path for the Eth Key")
+	cfg.EthAcctAddr = flag.String("ethAcctAddr", *cfg.EthAcctAddr, "Existing Eth account address. For use when multiple ETH accounts exist in keystore directory")
```
leszko:
```suggestion
	cfg.EthAcctAddr = flag.String("ethAcctAddr", *cfg.EthAcctAddr, "Existing Eth account address. For use when multiple ETH accounts exist in the keystore directory")
```

cmd/livepeer/starter/starter.go
```diff
@@ -1412,3 +1417,31 @@ func getBroadcasterPrices(broadcasterPrices string) []BroadcasterPrice {
 
 	return pricesSet.Prices
 }
+
+type KeystorePath struct {
+	path, address interface{}
```
leszko: Can we use `string` here? Then there would be no need for casting.

```suggestion
	path, address string
```

cmd/livepeer/starter/starter.go
```diff
@@ -1412,3 +1417,31 @@ func getBroadcasterPrices(broadcasterPrices string) []BroadcasterPrice {
...
+}
+
+func ParseEthKeystorePath(ethKeystorePath string) (KeystorePath, error) {
+	var keystore KeystorePath = KeystorePath{"", ""}
```
```suggestion
	var keystore = KeystorePath{"", ""}
```

cmd/livepeer/starter/starter.go
```diff
@@ -526,16 +526,21 @@ func StartLivepeer(ctx context.Context, cfg LivepeerConfig) {
...
+				}
+			} else {
+				keystoreDir = filepath.Join(*cfg.Datadir, "keystore")
+			}
 
-		if keystoreDir == "" {
-			glog.Errorf("Cannot find keystore directory")
-			return
+			if err != nil {
+				glog.Warning(fmt.Errorf(err.Error()+" Defaulting to %s", keystoreDir))
```
leszko: I think we should fail fast here instead of just logging the warning. This is the case in which a user gave the incorrect `EthKeystorePath`, so the configuration is definitely incorrect.

cmd/livepeer/starter/starter.go
```diff
@@ -526,16 +526,21 @@ func StartLivepeer(ctx context.Context, cfg LivepeerConfig) {
...
-		var keystoreDir string
-		if _, err := os.Stat(*cfg.EthKeystorePath); !os.IsNotExist(err) {
-			keystoreDir, _ = filepath.Split(*cfg.EthKeystorePath)
-		} else {
-			keystoreDir = filepath.Join(*cfg.Datadir, "keystore")
-		}
+		var keystoreDir = filepath.Join(*cfg.Datadir, "keystore")
+		if *cfg.EthKeystorePath != "" {
+			keystoreInfo, err := ParseEthKeystorePath(*cfg.EthKeystorePath)
+			if keystoreInfo.path.(string) != "" {
```
leszko: Shouldn't we actually check for an error here and fail if there is any `err` coming from `ParseEthKeystorePath`?

cmd/livepeer/starter/starter.go
```diff
@@ -526,16 +526,21 @@ func StartLivepeer(ctx context.Context, cfg LivepeerConfig) {
...
+				keystoreDir = keystoreInfo.path.(string)
+				if keystoreInfo.address.(string) != "" && err == nil {
+					*cfg.EthAcctAddr = keystoreInfo.address.(string)
```
leszko: I think this is a tricky part, because from what I understand now, if you have both `EthKeystorePath` and `EthAcctAddr` defined then, livepeer node will use the address from keystore.

The best would be to fail fast if you have `EthKeystorePath` and `EthAcctAddr` defined with different ETH addresses. But this can be difficult to check, then in that case, I think we should maybe not allow specifying `EthAcctAddr` if your `EthKeystorePath` points directly to a file. What do you think?

cmd/livepeer/starter/starter.go
```diff
@@ -1412,3 +1417,31 @@ func getBroadcasterPrices(broadcasterPrices string) []BroadcasterPrice {
...
+
+type KeystorePath struct {
+	path, address interface{}
+}
+
+func ParseEthKeystorePath(ethKeystorePath string) (KeystorePath, error) {
+	var keystore KeystorePath = KeystorePath{"", ""}
+	ethKeystorePath = strings.TrimSuffix(ethKeystorePath, "/")
+
+	if fileInfo, err := os.Stat(ethKeystorePath); !os.IsNotExist(err) && fileInfo != nil {
```
leszko: Why `!os.IsNotExist(err)` and not just `err != nil`? What should we do in case of other errors?

cmd/livepeer/starter/starter.go
```diff
@@ -1412,3 +1417,31 @@ func getBroadcasterPrices(broadcasterPrices string) []BroadcasterPrice {
...
+		if fileInfo.IsDir() {
+			keystore.path = ethKeystorePath
+		} else {
+			keystore.path, _ = filepath.Split(ethKeystorePath)
+			if keyText, err := common.ReadFromFile(ethKeystorePath); err == nil {
+				if address, err := common.ParseEthAddr(keyText); err == nil {
+					keystore.address = address
+				} else {
+					return keystore, errors.New("Error parsing address from keyfile.")
```
```suggestion
					return keystore, errors.New("error parsing address from keyfile")
```

cmd/livepeer/starter/starter.go
```diff
@@ -1412,3 +1417,31 @@ func getBroadcasterPrices(broadcasterPrices string) []BroadcasterPrice {
...
+				}
+			}
+		}
+	} else {
+		return keystore, errors.New("Provided -ethKeystorePath was not found.")
```
```suggestion
		return keystore, errors.New("provided -ethKeystorePath was not found")
```

cmd/livepeer/starter/starter.go
```diff
@@ -1412,3 +1417,31 @@ func getBroadcasterPrices(broadcasterPrices string) []BroadcasterPrice {
...
+func ParseEthKeystorePath(ethKeystorePath string) (KeystorePath, error) {
+	var keystore KeystorePath = KeystorePath{"", ""}
+	ethKeystorePath = strings.TrimSuffix(ethKeystorePath, "/")
+
+	if fileInfo, err := os.Stat(ethKeystorePath); !os.IsNotExist(err) && fileInfo != nil {
+		if fileInfo.IsDir() {
+			keystore.path = ethKeystorePath
+		} else {
+			keystore.path, _ = filepath.Split(ethKeystorePath)
+			if keyText, err := common.ReadFromFile(ethKeystorePath); err == nil {
```
leszko: I think that if `err != nil`, then we should just return an error and fail-fast. Do we want to proceed in case of a wrong path?

common/util.go
```diff
@@ -518,3 +519,13 @@ func ParseAccelDevices(devices string, acceleration ffmpeg.Acceleration) ([]stri
 	}
 	return strings.Split(devices, ","), nil
 }
+
+func ParseEthAddr(strJsonKey string) (string, error) {
+	var keyJson map[string]interface{}
```
leszko, line 524:
```suggestion
	var keyJson map[string]string
```

Then, you'll not need casting later.

leszko (changes requested):
I added a few comments. Most of them are minor. 

From the more important ones, I think we should fail-fast in the following scenarios:
- `EthKeystorePath` points to a non-existing file
- `EthAcctAddr` and `EthKeystorePath` are both defined but they point different ETH addresses

common/util.go
eliteprox: Thanks, that was much easier :)

cmd/livepeer/starter/starter.go
eliteprox: Updated logic flow to return "error opening keystore" when path or file not found

cmd/livepeer/starter/starter.go
eliteprox: There wasn't a good reason, I have now simplified this error check to `err != nil`

cmd/livepeer/starter/starter.go
eliteprox: I created standard formatting for the ETH addresses by using the `ethcommon.HexToAddress` function to santize input, which returns a '0x0' eth address when invalid. EthAcctAddr is only updated if it is empty (flag not provided) or `EthAcctAddr` and `keystoreInfo.address` do not match. Added log message for mismatched addresses.

cmd/livepeer/starter/starter.go
eliteprox: Updated logic flow to fail with err message from`ParseEthKeystorePath` returns any error.

cmd/livepeer/starter/starter.go
eliteprox: It will now fail if EthKeystorePath does not match a path or file.

cmd/livepeer/starter/starter.go
```diff
@@ -1412,3 +1420,37 @@ func getBroadcasterPrices(broadcasterPrices string) []BroadcasterPrice {
 
 	return pricesSet.Prices
 }
+
+type KeystorePath struct {
```
```suggestion
type keystorePath struct {
```

I think we don't need it to be accessible outside this package.

cmd/livepeer/starter/starter.go
```diff
@@ -1412,3 +1420,37 @@ func getBroadcasterPrices(broadcasterPrices string) []BroadcasterPrice {
...
+	path    string
+	address ethcommon.Address
+}
+
+func ParseEthKeystorePath(ethKeystorePath string) (KeystorePath, error) {
```
```suggestion
func parseEthKeystorePath(ethKeystorePath string) (KeystorePath, error) {
```

I think we don't need to have it accessible outside this package.

cmd/livepeer/starter/starter.go
```diff
@@ -526,16 +526,24 @@ func StartLivepeer(ctx context.Context, cfg LivepeerConfig) {
...
+				}
+			}
 		} else {
-			keystoreDir = filepath.Join(*cfg.Datadir, "keystore")
-		}
-
-		if keystoreDir == "" {
-			glog.Errorf("Cannot find keystore directory")
-			return
+			glog.Fatal(fmt.Errorf(err.Error()))
```
leszko, line 547:
```suggestion
			glog.Fatal(fmt.Errorf(err.Error()))
			return
```

We can return after this error.

cmd/livepeer/starter/starter.go
```diff
@@ -1412,3 +1420,37 @@ func getBroadcasterPrices(broadcasterPrices string) []BroadcasterPrice {
...
+	ethKeystorePath = strings.TrimSuffix(ethKeystorePath, "/")
+	fileInfo, err := os.Stat(ethKeystorePath)
+	if err != nil {
+		return keystore, errors.New("provided -ethKeystorePath was not found")
+	}
+
+	if fileInfo.IsDir() {
+		keystore.path = ethKeystorePath
+	} else {
+		keystore.path, _ = filepath.Split(ethKeystorePath)
```
```suggestion
```

I think it makes more sense to return either `path` or `address`. Then, it's simpler to reason about it.

cmd/livepeer/starter/starter.go
```diff
@@ -526,16 +526,24 @@ func StartLivepeer(ctx context.Context, cfg LivepeerConfig) {
...
 	} else {
-		var keystoreDir string
-		if _, err := os.Stat(*cfg.EthKeystorePath); !os.IsNotExist(err) {
-			keystoreDir, _ = filepath.Split(*cfg.EthKeystorePath)
+		var keystoreDir = filepath.Join(*cfg.Datadir, "keystore")
+		if keystoreInfo, err := ParseEthKeystorePath(*cfg.EthKeystorePath); err == nil {
+			if keystoreInfo.path != "" {
+				keystoreDir = keystoreInfo.path
+
+				if (keystoreInfo.address != ethcommon.Address{}) {
```
```suggestion
			if keystoreInfo.path != "" {
				keystoreDir = keystoreInfo.path
			} else if (keystoreInfo.address != ethcommon.Address{}) {
```

It corresponds to my other comment that it's better to return either `path` or `address`.

cmd/livepeer/starter/starter_test.go
```diff
@@ -99,3 +102,66 @@ func TestParseGetBroadcasterPrices(t *testing.T) {
 	assert.Equal(big.NewRat(1000, 1), price1)
 	assert.Equal(big.NewRat(2000, 3), price2)
 }
+
+func TestFlag_KeystorePath_File(t *testing.T) {
```
leszko:
```suggestion
func TestParseEthKeystorePath(t *testing.T) {
```

Follow the naming convention for other tests.

cmd/livepeer/starter/starter_test.go
```diff
@@ -99,3 +102,66 @@ func TestParseGetBroadcasterPrices(t *testing.T) {
...
+
+	var keystoreInfo KeystorePath
+	keystoreInfo, _ = ParseEthKeystorePath(file1.Name())
+
+	assert.NotEmpty(keystoreInfo.path)
+	assert.NotEmpty(keystoreInfo.address)
+	assert.True(addr == keystoreInfo.address.Hex())
+}
+
+func TestFlag_KeystorePath_FileBadAddress(t *testing.T) {
```
leszko:
```suggestion
func TestParseEthKeystorePathIncorrectAddress(t *testing.T) {
```

cmd/livepeer/starter/starter_test.go
```diff
@@ -99,3 +102,66 @@ func TestParseGetBroadcasterPrices(t *testing.T) {
...
+	badJsonfile.WriteString("{{\"address_broken_json\":\"" + addr + "\"}")
+
+	var keystoreInfo KeystorePath
+	keystoreInfo, err = ParseEthKeystorePath(badJsonfile.Name())
+	assert.NotEmpty(keystoreInfo.path)
+	assert.Empty(keystoreInfo.address)
+	assert.True(err.Error() == "error parsing address from keyfile")
+}
+
+func TestFlag_KeystorePath_Directory(t *testing.T) {
```
leszko:
```suggestion
func TestParseEthKeystorePathIncorrectAddressDirectory(t *testing.T) {
```

cmd/livepeer/starter/starter_test.go
```diff
@@ -99,3 +102,66 @@ func TestParseGetBroadcasterPrices(t *testing.T) {
...
+	assert := assert.New(t)
+	tempDir := t.TempDir()
+
+	var keystoreInfo KeystorePath
+	keystoreInfo, err := ParseEthKeystorePath(tempDir)
+	assert.NotEmpty(keystoreInfo.path)
+	assert.True(err == nil)
+}
+
+func TestFlag_KeystorePath_PathNotFound(t *testing.T) {
```
leszko:
```suggestion
func TestParseEthKeystorePathNotFound(t *testing.T) {
```

cmd/livepeer/starter/starter_test.go
```diff
@@ -99,3 +102,66 @@ func TestParseGetBroadcasterPrices(t *testing.T) {
...
+	assert := assert.New(t)
+
+	var addr = "0x0000000000000000000000000000000000000001"
+	var fname = "UTC--2023-01-05T00-46-15.503776013Z--" + addr
+	tempDir := t.TempDir()
+	file1, err := ioutil.TempFile(tempDir, fname)
+	if err != nil {
+		panic(err)
+	}
+	defer syscall.Unlink(fname)
```
leszko:
```suggestion
	defer os.Remove(fname)
```

Use `os.Remove()` instead

cmd/livepeer/starter/starter_test.go
```diff
@@ -99,3 +102,66 @@ func TestParseGetBroadcasterPrices(t *testing.T) {
...
+
+	//Create test file
+	var addr = "0x0000000000000000000000000000000000000001"
+	var fname = "UTC--2023-01-05T00-46-15.503776013Z--" + addr
+	badJsonfile, err := ioutil.TempFile(tempDir, fname)
+	if err != nil {
+		panic(err)
+	}
+
+	defer syscall.Unlink(fname)
```
leszko: Use `os.Remove()` instead

cmd/livepeer/starter/starter_test.go
```diff
@@ -99,3 +102,66 @@ func TestParseGetBroadcasterPrices(t *testing.T) {
...
 	assert.Equal(big.NewRat(2000, 3), price2)
 }
+
+func TestFlag_KeystorePath_File(t *testing.T) {
+	assert := assert.New(t)
+
+	var addr = "0x0000000000000000000000000000000000000001"
+	var fname = "UTC--2023-01-05T00-46-15.503776013Z--" + addr
+	tempDir := t.TempDir()
+	file1, err := ioutil.TempFile(tempDir, fname)
```
leszko:
```suggestion
	file1, err := os.CreateTemp("", "")
```

`ioutil.TempFile()` is deprecated

cmd/livepeer/starter/starter_test.go
```diff
@@ -99,3 +102,66 @@ func TestParseGetBroadcasterPrices(t *testing.T) {
...
+}
+
+func TestFlag_KeystorePath_FileBadAddress(t *testing.T) {
+	assert := assert.New(t)
+	tempDir := t.TempDir()
+
+	//Create test file
+	var addr = "0x0000000000000000000000000000000000000001"
+	var fname = "UTC--2023-01-05T00-46-15.503776013Z--" + addr
+	badJsonfile, err := ioutil.TempFile(tempDir, fname)
```
leszko: `ioutil.TempFile()` is deprecated

common/util.go
leszko, line 524:
```suggestion
	var keyJson map[string]string
```

common/util.go
```diff
@@ -518,3 +519,13 @@ func ParseAccelDevices(devices string, acceleration ffmpeg.Acceleration) ([]stri
...
+	if err := json.Unmarshal([]byte(strJsonKey), &keyJson); err == nil {
+		if address, ok := keyJson["address"].(string); ok {
```
leszko, line 526:
```suggestion
		if address, ok := keyJson["address"]; ok {
```

leszko (changes requested):
Added some super minor comments. Please take a look and address and think it's ready for the merge. Good work 👍

cmd/livepeer/starter/starter_test.go
eliteprox: Fixed

cmd/livepeer/starter/starter_test.go
eliteprox: Fixed

leszko:
@eliteprox when you address all comments, please click "Re-request review"

common/util.go
eliteprox: I had to revert this commit as it caused some runtime issues.

cmd/livepeer/starter/starter.go
eliteprox: Completed

cmd/livepeer/starter/starter_test.go
eliteprox: Renamed to `TestParse_ParseEthKeystorePathFileNotFound`, added tests on both missing directory and file

cmd/livepeer/starter/starter.go
eliteprox: Refactored

cmd/livepeer/starter/starter.go
eliteprox: Added return

common/util.go
eliteprox: After changing to what is recommended, the JSON parsing failed at runtime. I reverted it back and it is working fine: https://github.com/eliteprox/go-livepeer/commit/81e876bc860c9a7fb6735f99caa6524235404040

Please recommend if there is a better way to handle this

leszko: Could you write how it failed for you?

I tried changing the code to the following:
```
     var keyJson map[string]string
        if err := json.Unmarshal([]byte(strJsonKey), &keyJson); err == nil {
            if address, ok := keyJson["address"]; ok {
                    return address, nil
            }
    }
```

And it worked for me. Could you write the command (or test) that fails in your case?

leszko (commented):
Added one more comment, could you check?

common/util.go
eliteprox: I tried this and the test passed, but my functional tests failed. I determined the difference was the test `TestParse_ParseEthKeystorePathValidFile` did not include anything more than a single level JSON string with single address field. I've updated the test to include the entire JSON structure from keystore. You will find that the suggested code will fail on test. 

Thanks for the help, any further suggestions to simplify it?

leszko: Ahh, I didn't know that this file can have nested objects. In that case, let's leave it as it is. Thanks for double-checking and for adding the test case for it.

leszko (approved):
LGTM, thanks for all the changes @eliteprox 

I'll merge as soon as the CI build is completed.

PR #2729, opened 2023-01-17
implement a custom error type for max transcode attempts

mjh1:

Created a custom error and used multi-error wrapping to avoid needing to do any string matching in the error handling.

I modified the existing unit test.

**Checklist:**

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -11,6 +11,7 @@
 #### Broadcaster
 - \#2837 Set a floor of 1.5s for accepted Orchestrator response times, regardless of segment length (@thomshutt)
 - \#2839 Add Orchestrator blocklist CLI parameter (@mjh1)
+- \#2729 Switch to a custom error type for max transcode attempts
 
 #### Orchestrator
 - \#2781 Add automatic session limit and set max sessions with livepeer_cli
--- a/server/broadcast.go
+++ b/server/broadcast.go
@@ -55,6 +55,7 @@ var submitMultiSession = func(ctx context.Context, sess *BroadcastSession, seg *
 	nonce uint64, calcPerceptualHash bool, resc chan *SubmitResult) {
 	go submitSegment(ctx, sess, seg, segPar, nonce, calcPerceptualHash, resc)
 }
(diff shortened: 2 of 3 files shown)


cyberj0g:
@mjh1 thanks for looking into that.
I had a slightly more generic solution in mind, like, a wrapper to add more detailed error message along with custom error type, without masking the root error:
```
ErrFancyCustomError  = errors.New("custom message")
...
func faulty() error {
    err = ...
    return fmt.Errorf("even more detailed message: %w", lperrors.Error(ErrFancyCustomError, err))
    // with a pre-defined message and error type, but root cause error is also included
    // return lperrors.Error(ErrFancyCustomError, err)
}
...
anErr = faulty()
if errors.Is(anErr, ErrFancyCustomError) {
// true
}
// anErr printed like this:
// custom message: even more detailed message: context cancelled
```
Also, I think it's not a good idea to use `errors` as a package name, because there's a builtin package with the same name. Maybe I'm inventing the wheel here, let me know what you think.

PS: it's not a high priority feature

mjh1:
@cyberj0g yeah I was trying to figure out a way to make it generic, it's a little fiddly but I think my latest change should be good enough. It'll get a lot easier in go 1.20 we'll be able wrap multiple errors: https://tip.golang.org/doc/go1.20#errors

cyberj0g:
> https://tip.golang.org/doc/go1.20#errors

Cool! Maybe we could just backport some functions we are interested in, such as: `errrors.Is`, `fmt.Errorf` from 1.20 to a `lperrors` package, and then delete that, once we move go-livepeer to Go 1.20? These functions are few lines usually, it shouldn't be too hard.

mjh1:
ah good plan 👍

mjh1:
@cyberj0g annoyingly i'm not sure it's possible, the `Is` function makes use of this internal reflection package: https://github.com/golang/go/blob/go1.20rc3/src/errors/wrap.go#L8 
what do you think?

cyberj0g:
Ok, let's draft that until Go 1.20 is released. Thanks for researching!

thomshutt (approved):
Beautiful

PR #2740, opened 2023-01-27
Fix incorrect processing of VerificationFreq parameter

cyberj0g:
**About** 
`VerificationFreq` parameter of the transcoding job controls how often fast verification is re-run on an already verified O session. Fast verification involves transcoding segment with 1 trusted and 2 untrusted Os, instead of using a single O. This PR fixes incorrect processing of this parameter.  Valid values are 0 and positive integers. If `VerificationFreq = 0`, fast verification logic is fully disabled. If `VerificationFreq=N>0`, then, after initial verification run for a new O session, it will run again for a segment with a probability of `1 / VerificationFreq`.

**Issue**
The `shouldSkipVerification` function used `VerificationFreq` parameter in an inverted way. `VerificationFreq=1` caused fast verification to run only once, as long as verified session is available, and higher values caused fast verification to run more and more often, with a value of e.g. 40 giving a **97.5%** chance of running fast verification for each new segment of the stream.

### DIFF
--- a/server/broadcast.go
+++ b/server/broadcast.go
@@ -397,7 +397,7 @@ func (bsm *BroadcastSessionsManager) shouldSkipVerification(sessions []*Broadcas
 	if !includesSession(sessions, bsm.verifiedSession) {
 		return false
 	}
-	return common.RandomUintUnder(bsm.VerificationFreq) == 0
+	return common.RandomUintUnder(bsm.VerificationFreq) != 0
 }
 
 func NewSessionManager(ctx context.Context, node *core.LivepeerNode, params *core.StreamParameters, sel BroadcastSessionsSelectorFactory) *BroadcastSessionsManager {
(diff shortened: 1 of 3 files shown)


server/broadcast.go
```diff
@@ -397,7 +397,7 @@ func (bsm *BroadcastSessionsManager) shouldSkipVerification(sessions []*Broadcas
 	if !includesSession(sessions, bsm.verifiedSession) {
 		return false
 	}
-	return common.RandomUintUnder(bsm.VerificationFreq) == 0
+	return common.RandomUintUnder(bsm.VerificationFreq) != 0
```
yondonfu, line 400: Ooof good catch!

IIUC this method generates a random integer `x` s.t. `0 <= x < VerificationFreq` and there would be a `1 / VerificationFreq` probability for the random integer to be 0. Since this method is checking if verification should be *skipped*, if the check on this line is whether the integer is equal to 0 then this method would return true `1 / VerificationFreq` of the time which is the opposite of what we want i.e. this method should return true `(VericationFreq - 1) / VerificationFreq)` of the time.

thomshutt (approved):

Issue #2605, opened 2022-09-23
Disconnecting a remote T causes O to lose all transcoders


eliteprox:
**Describe the bug**

When operating an orchestrator with multiple remote transcoders, disconnecting one remote T can sometimes cause orchestrator to lose all transcoders and go into a `no transcoders available` state.

**To Reproduce**
1. Run standalone orchestrator
2. Run two transcoder processes on localhost with orchestrator
3. Run a third transcoder process on a separate IP address connected to the same orchestrator.
4. Stop the third transcoder process while orchestrator has streams running on all three transcoders.
5. Orchestrator logs are attached.
[orchestratorlogs.txt](https://github.com/livepeer/go-livepeer/files/9630302/orchestratorlogs.txt)
[local-transcoder2.txt](https://github.com/livepeer/go-livepeer/files/9630297/local-transcoder2.txt)
[remote-transcoder3.txt](https://github.com/livepeer/go-livepeer/files/9630303/remote-transcoder3.txt)

**Expected behavior**
The stopped transcoder should disconnect without affecting workflow to the other transcoders

**Desktop (please complete the following information):**
 - OS: Ubunutu Linux 22.04.01 on both machines
 - Version 0.5.34

**Additional context**
See attached logs.
**orchestratorlogs.txt** shows the orchestrator logs during this time period.  
   - We stopped remote transcoder 70.132.135.145:34772 on purpose using `systemctl stop`
```
Sep 21 14:39:39 whitebox systemd[1]: Stopping Livepeer Trancoder 1 Service...
Sep 21 14:39:39 whitebox livepeer[3963]: I0921 14:39:39.183235    3963 ot_rpc.go:116] Exiting Livepeer Transcoder: terminated
Sep 21 14:39:39 whitebox livepeer[3963]: I0921 14:39:39.183300    3963 ot_rpc.go:128] End of stream receive cycle because of err="execution interrupted", waiting for running transcode jobs to complete
Sep 21 14:39:39 whitebox livepeer[3963]: I0921 14:39:39.183372    3963 ot_rpc.go:60] Unregistering transcoder: execution interrupted
Sep 21 14:39:39 whitebox livepeer[3963]: I0921 14:39:39.183377    3963 ot_rpc.go:62] Terminating transcoder because of execution interrupted
Sep 21 14:41:09 whitebox systemd[1]: transcoder1.service: State 'stop-sigterm' timed out. Killing.
[...]

PR #2747, opened 2023-02-07
Fixed a transcoding bug that occurred when remote transcoder was removed

eliteprox:
This change fixes a bug that caused an orchestrator to lose all transcoders when one of it's transcoders disconnects unexpectedly. 

- Simplified logic in `removeFromRemoteTranscoders` so that correct transcoder is removed from RemoteTranscoderManager

We tested up to 4 streams going to an orchestrator with up to 4 remote transcoders. We stopped one of the transcoders in the following scenarios and observed streams move to another transcoder without disconnecting any other transcoders.
- 1 stream on each of the 4 remote transcoders
- 2 streams on each of 2 remote transcoders
- 4 streams on 1 transcoder with 3 additional transcoders having 0 load. 

In every case we observed the streams safely move to another available transcoder without causing other transcoders to break.

#2605
#2706

**Checklist:**

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -15,13 +15,13 @@
 #### Transcoder
 
 ### Bug Fixes 🐞
+- #2746 Fixed a transcoding bug that occured when remote transcoder was removed
 
 #### CLI
 
 #### General
 
 #### Broadcaster
-- \#2709 Add logging for high keyframe interval, reduce log level for discovery loop
 - \#2684 Fix transcode success rate metric
 
 #### Orchestrator
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -15,7 +15,7 @@
 #### Transcoder
 
 ### Bug Fixes 🐞
-- #2746 Fixed a transcoding bug that occurred when remote transcoder was removed
+- #2747 Fixed a transcoding bug that occurred when remote transcoder was removed
 
 #### CLI
 
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -15,7 +15,7 @@
 #### Transcoder
 
 ### Bug Fixes 🐞
-- #2746 Fixed a transcoding bug that occured when remote transcoder was removed
+- #2746 Fixed a transcoding bug that occurred when remote transcoder was removed
 
 #### CLI
 
--- a/core/orch_test.go
+++ b/core/orch_test.go
@@ -429,13 +429,14 @@ func TestRemoveFromRemoteTranscoders_StreamFailover(t *testing.T) {
 	var transcoder_2 = &StubTranscoderServer{manager: rtm}
 
 	assert.Nil(rtm.liveTranscoders[transcoder_1])
+	assert.Nil(rtm.liveTranscoders[transcoder_2])
 	assert.Empty(rtm.remoteTranscoders)
 
 	// register 2 transcoders, 8 stream capacity each
 	go func() { rtm.Manage(transcoder_1, 8, capabilities.ToNetCapabilities()) }()
-	time.Sleep(1 * time.Millisecond) // allow time for first stream to register
+	time.Sleep(1 * time.Millisecond) // allow time for remote T to connect
 	go func() { rtm.Manage(transcoder_2, 8, capabilities.ToNetCapabilities()) }()
-	time.Sleep(1 * time.Millisecond) // allow time for second stream to register
+	time.Sleep(1 * time.Millisecond) // allow time for remote T to connect
 
 	//Validate two transcoders exist
 	assert.NotNil(rtm.liveTranscoders[transcoder_1])
@@ -446,7 +447,6 @@ func TestRemoveFromRemoteTranscoders_StreamFailover(t *testing.T) {
 	for i := 0; i < 8; i++ {
 		testSessionId := fmt.Sprintf("%s%d", "testID", i)
 		rtm.selectTranscoder(testSessionId, nil)
-		time.Sleep(1 * time.Millisecond) // allow time for second stream to register
 	}
 
 	t1 := rtm.liveTranscoders[transcoder_1]
@@ -465,7 +465,6 @@ func TestRemoveFromRemoteTranscoders_StreamFailover(t *testing.T) {
 	for i := 0; i < 8; i++ {
 		testSessionId := fmt.Sprintf("%s%d", "testID", i)
 		rtm.selectTranscoder(testSessionId, nil)
-		time.Sleep(1 * time.Millisecond) // allow time for second stream to register
 	}
 
 	t1 = rtm.liveTranscoders[transcoder_1]
--- a/core/orch_test.go
+++ b/core/orch_test.go
@@ -419,6 +419,63 @@ func TestRemoveFromRemoteTranscoders(t *testing.T) {
 	assert.Equal(remoteTranscoderList, emptyTList)
 }
 
+//Validates that selectTranscoder will shift all streams to another valid transcoder if transcoder fails.
+func TestRemoveFromRemoteTranscoders_StreamFailover(t *testing.T) {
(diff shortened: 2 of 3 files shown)


Titan-Node:
Works great, did many tests with removing and adding multiple Ts without any dropped streams.

thomshutt:
This looks like a great simplification and as far as I can tell, the original code is just convoluted because of micro-optimisations that we don't need. Could you also please add a unit test so that we can avoid breaking this again in the future?

CHANGELOG_PENDING.md
```diff
@@ -15,13 +15,13 @@
...
 
 ### Bug Fixes 🐞
+- #2747 Fixed a transcoding bug that occurred when remote transcoder was removed
 
 #### CLI
 
 #### General
 
 #### Broadcaster
-- \#2709 Add logging for high keyframe interval, reduce log level for discovery loop
```
leszko, line 25: Why is this line removed?

eliteprox:
> This looks like a great simplification and as far as I can tell, the original code is just convoluted because of micro-optimisations that we don't need. Could you also please add a unit test so that we can avoid breaking this again in the future?

Yep! I will write some tests, was thinking the same. This is a critical functionality. Thanks for looking it over 😎

leszko (commented):
Nice, great simplification.

+1 for adding unit tests.

Also wondering, what's the functional difference between the old and the new code? Isn't it working exactly the same, how does it fix the issue you encountered?

CHANGELOG_PENDING.md
eliteprox: This was a note from prior release that somehow was left in the pending changelog.

stronk-dev:
Already has a unit test BTW:
https://github.com/livepeer/go-livepeer/blob/ab553b4c58589e1ce3bb41905dd1f56f804c3900/core/orch_test.go#L372-L420

thomshutt:
Good point @stronk-dev - I'd still want to update that test such that it would've caught the case that was breaking though

eliteprox:
> Also wondering, what's the functional difference between the old and the new code? Isn't it working exactly the same, how does it fix the issue you encountered?

It appears some funky logic or race condition led to it. I couldn't follow the logic there. I discovered this after monitoring my own node. We then tested with real test streams until it was a confirmed bug, and then worked through the solution. The issue occurred only when there was 2 or more streams on the transcoder that was disconnected.

I wrote a new test which reproduced the issue on the old version of `removeRemoteTranscoder` and passed on the new code.
https://github.com/livepeer/go-livepeer/blob/707115e481c487d806cba93769c6b3fe8b73a92f/core/orch_test.go#L423-L477

In the test which uses 2 remote transcoders and 8 streams:
Old code = Only 5 streams made it to the 2nd transcoder
New code = All 8 streams went to the 2nd transcoder

eliteprox:
Added new test `TestRemoveFromRemoteTranscoders_StreamFailover`. I left the previous test because the new one is more focused on the high-level result of shifting streams to the opposite transcoder and uses the `selectTranscoder` function to validate this.

eliteprox:
@leszko @thomshutt Any chance this can make it into 0.5.38? #2753

core/orch_test.go
```diff
@@ -419,6 +419,63 @@ func TestRemoveFromRemoteTranscoders(t *testing.T) {
...
+
+	var transcoder_1 = &StubTranscoderServer{manager: rtm}
+	var transcoder_2 = &StubTranscoderServer{manager: rtm}
+
+	assert.Nil(rtm.liveTranscoders[transcoder_1])
+	assert.Empty(rtm.remoteTranscoders)
+
+	// register 2 transcoders, 8 stream capacity each
+	go func() { rtm.Manage(transcoder_1, 8, capabilities.ToNetCapabilities()) }()
+	time.Sleep(1 * time.Millisecond) // allow time for first stream to register
```
leszko: Won't this test be flaky? I think depending on timing (especially in ms) can make this test sometime pass, sometimes not. I'd avoid having time-dependent tests.

eliteprox: I copied the wait commands from another test.  I think they are not needed. I didn't see anything in my development that indicates they are required, so I can try removing them.

leszko: Yeah, then I suggest removing them if not needed. Then, Do you need to run these functions above in separate goroutines?

eliteprox: Yes, apparently the separate go routine is required in this case, if not done with the go routine, the test will timeout. I tried removing the time.sleep between them, but it caused test to move on without transcoder connected at all.

I pushed a commit to cleanup a few related lines.

leszko: Hmm, THB, in my opinion, it's better to not have tested with `time.Sleep()`, because they will sometimes fail sometimes pass in our CI. So, if it's not possible (or easy) to write a test without sleep, then I suggest removing this test and just merging without it. @thomshutt wdyt?

The improvement to the code is already good, so should be safe to merge.

thomshutt:
@leszko @eliteprox I don't understand - if the problem is this method being called by multiple goroutines in an unsafe way then shouldn't the problem be fixed further up the stack? 

If the problem is inside `removeFromRemoteTranscoders` then we should be able to write a test that just calls that method, without any concurrency code.

leszko:
> @leszko @eliteprox I don't understand - if the problem is this method being called by multiple goroutines in an unsafe way then shouldn't the problem be fixed further up the stack?
> 
> If the problem is inside `removeFromRemoteTranscoders` then we should be able to write a test that just calls that method, without any concurrency code.

From my current understanding, there are 2 issues:
- this function is unnecessarily overcomplicated => this PR solves this issue
- this function is used in an unsafe manner by multiple goroutines => this PR does not solve this issue (though might make it less visible)

Then, I suggest doing the following:
1. For this PR, write a very simple sequential unit test just to test the function `removeFromRemoteTranscoders()`, nothing more
2. Do not include the test `TestRemoveFromRemoteTranscoders_StreamFailover`, since I believe it's flaky
3. Merge this PR
4. Create a GH Issue to solve the concurrency issue (unless it's simple, then @eliteprox feel free to fix it in this PR)

eliteprox:
> * For this PR, write a very simple sequential unit test just to test the function `removeFromRemoteTranscoders()`, nothing more
> 
> * Do not include the test `TestRemoveFromRemoteTranscoders_StreamFailover`, since I believe it's flaky
> 
> * Merge this PR
> 
> * Create a GH Issue to solve the concurrency issue (unless it's simple, then @eliteprox feel free to fix it in this PR)

Can we simply continue using the existing `TestRemoveFromRemoteTranscoders` test for now until the tests can be refactored, and merge this PR? This test properly covers removing transcoders within it's own scope, imho.

I've removed `TestRemoveFromRemoteTranscoders_StreamFailover`. I created GH issue #2756 for the refactoring of tests.

leszko:
> > * For this PR, write a very simple sequential unit test just to test the function `removeFromRemoteTranscoders()`, nothing more
> > * Do not include the test `TestRemoveFromRemoteTranscoders_StreamFailover`, since I believe it's flaky
> > * Merge this PR
> > * Create a GH Issue to solve the concurrency issue (unless it's simple, then @eliteprox feel free to fix it in this PR)
> 
> Can we simply continue using the existing `TestRemoveFromRemoteTranscoders` test for now until the tests can be refactored, and merge this PR? This test properly covers removing transcoders within it's own scope, imho.
> 
> I've removed `TestRemoveFromRemoteTranscoders_StreamFailover`. I created GH issue #2756 for the refactoring of tests.

I'm ok with this. I think this change is safe as it is and it simplifies the code a lot. @thomshutt wdyt?

leszko (approved):
LGTM

Let's wait for @thomshutt to comment/approve and then we can merge.

Titan-Node:
Been running this upgrade on prod for 4 days with no errors to report

thomshutt (approved):

thomshutt:
LGTM, thanks for the fix and for responding to all our questions @eliteprox!

PR #2759, opened 2023-02-17
Parse keystore address without 0x prefix, fix parse error logging

### LINKED ISSUE #2757: EthKeystorePath fails if pointed to a file instead of a directory in 0.5.38
**Describe the bug**
`EthKeystorePath` fails with misleading `Error unlocking Ethereum account: could not decrypt key with given password` if pointed to a file instead of a directory in 0.5.38.

**To Reproduce**
1. Set `ethKeystorePath /var/lib/livepeer/UTC--2022-10-29T19-52-34.103Z--5caaab7626edc7123cf8484edbc66a875dd32cc9.json`
2. Start livepeer, it will fail with `Error unlocking Ethereum account: could not decrypt key with given password`
3. Now set both `ethAcctAddr` and `ethKeystorePath /var/lib/livepeer` with just the directory and no path and it works.

**Expected behavior**
It should not fail even if pointed to the file

 - OS: Arch linux kernel 6.1.11-arch1-1

**Additional context**
@eliteprox FYI
Bug likely introduced via https://github.com/livepeer/go-livepeer/pull/2713

eliteprox:
This resolves a bug from #2713 when parsing keystore addresses without the 0x prefix. Errors from `parseEthKeystorePath` were also not being logged, because `err` was only available within the scope of `if err == nil`. 

- Updated logic flow in `starter.go`.
- Changed `ethcommon.HexToAddress` to `ethcommon.BytesToAddress(ethcommon.FromHex(address))` in parseEthKeystorePath().
  - `ethcommon.FromHex` considers the 0x prefix optional.

- Validated error logging by performing a functional test with valid/invalid JSON, file not found, and bad JSON format
- Validated address parsing with/without 0x prefix. 
- Updated unit test `TestParse_ParseEthKeystorePathValidFile`

Fixes #2757 

**Checklist:**

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -15,6 +15,7 @@
 #### Transcoder
 
 ### Bug Fixes 🐞
+- \# 2759 Parse keystore address without 0x prefix, fix parse error logging
 
 #### CLI
 
--- a/cmd/livepeer/starter/starter.go
+++ b/cmd/livepeer/starter/starter.go
@@ -527,7 +527,8 @@ func StartLivepeer(ctx context.Context, cfg LivepeerConfig) {
 
 	} else {
 		var keystoreDir = filepath.Join(*cfg.Datadir, "keystore")
-		if keystoreInfo, err := parseEthKeystorePath(*cfg.EthKeystorePath); err == nil {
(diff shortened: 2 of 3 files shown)


leszko (approved):
Thanks for the fix @eliteprox

Issue #2762, opened 2023-03-01
Fix segment data locking on O "Session ended" notification 


iameli:
I think it's like
1. Some stream ends via timeout. Any stream on the B, doesn't have to be the one in question. <https://github.com/livepeer/go-livepeer/blob/e7e1183c05f80c30d52a0b9db874effb3b297b13/server/mediaserver.go#L930>
2. `removeRTMPStream` grabs the global connectionLock <https://github.com/livepeer/go-livepeer/blob/e7e1183c05f80c30d52a0b9db874effb3b297b13/server/mediaserver.go#L609>
3. A few layers deeper within that call, we notify the O that the session has ended. This is an async call involving a round-trip to an O that can fail for any number of reasons, ephemeral network errors and whatnot <https://github.com/livepeer/go-livepeer/blob/e7e1183c05f80c30d52a0b9db874effb3b297b13/server/rpc.go#L269>
4. Meanwhile, more segments for _other_ streams are coming in, and they all queue up behind the locked `s.connectionLock` on this line <https://github.com/livepeer/go-livepeer/blob/e7e1183c05f80c30d52a0b9db874effb3b297b13/server/mediaserver.go#L837>
5. Finally the O end-session request either succeeds or times out, unlocking the lock, resulting in the big burst of transcodes that we see back up here <https://github.com/livepeer/go-livepeer/blob/e7e1183c05f80c30d52a0b9db874effb3b297b13/server/mediaserver.go#L837>

And then we get the big burst of segment transcodes that we've been seeing.

Thulinma:
Knowing nothing of Go: would it make sense to just add that O end-session notification to a queue and handle that queue on a separate thread? I assume that notification need not happen with any active lock, anyway. 🤔

iameli:
@Thulinma Starting a new goroutine to do the O updating will have essentially that effect, yeah.

cyberj0g:
Interesting, does the timeout of 3 sec set [here](https://github.com/livepeer/go-livepeer/blob/d8ad5c7c9683bdf0569c1c3601416af360967840/server/rpc.go#L293) cause the issue? It should apply to all gRPC calls.

iameli:
@cyberj0g I don't think that would _cause_ the issue, in fact I suspect that timeout is causing the issue to be not as bad as it could be by capping the "everything is frozen" state to <= 3sec. The real problem is a cascading effect like

* Session closes
* 3 seconds of blocking for a stuck O
[...]

PR #2763, opened 2023-03-02
don't merge - disable session ending

iameli:
Works around https://github.com/livepeer/go-livepeer/issues/2762 while we get a proper fix in place. Which is probably just this with a proper async loop and testing.


iameli:
Replaced with https://github.com/livepeer/go-livepeer/pull/2764

PR #2764, opened 2023-03-02
Call session end async to avoid unnecessary blocking

mjh1:

- We are now calling session end asynchronously. Previously the blocking was causing delays in processing segments and ultimately transcode timeouts

Ran local B/O/T and verified the session end notifications are working asynchronously.

https://github.com/livepeer/go-livepeer/issues/2762

**Checklist:**


server/broadcast.go
```diff
@@ -682,9 +682,11 @@ func (bsm *BroadcastSessionsManager) collectResults(submitResultsCh chan *Submit
 // the caller needs to ensure bsm.sessLock is acquired before calling this.
 func (bsm *BroadcastSessionsManager) completeSessionUnsafe(ctx context.Context, sess *BroadcastSession, tearDown bool) {
 	if tearDown {
-		if err := EndTranscodingSession(ctx, sess); err != nil {
-			clog.Errorf(ctx, "Error completing transcoding session: %q", err)
-		}
+		go func() {
```
mjh1, line 685: @yondonfu @iameli Are you guys happy with this? I've been looking at the B code downstream of here as well as the function that ends up being called on the O side and can't see a need for blocking.

mjh1:
Setting to draft until I fix the unit tests

server/mediaserver_test.go
```diff
@@ -1226,6 +1223,10 @@ func TestRegisterConnection(t *testing.T) {
 }
 
 func TestBroadcastSessionManagerWithStreamStartStop(t *testing.T) {
+	goleakOptions := common.IgnoreRoutines()
+	// allow enough time for the transcode end goroutines to finish
+	goleakOptions = append(goleakOptions, goleak.MaxSleepInterval(5*time.Second), goleak.MaxRetryAttempts(1000))
+	defer goleak.VerifyNone(t, goleakOptions...)
```
cyberj0g, line 1229: Why 1000 retry attempts and 5 seconds delay (I believe it's a delay between attempts, not overall)? I think, the `EndTranscodingSession` goroutine will finish instantly here. Also maybe pass options to `VerifyNone` call directly?

cyberj0g (commented):
LGTM, except the unit test question, if you want my opinion 👍

server/broadcast.go
yondonfu: AFAICT you seem to be right that this call doesn't need to block for B.

.github/workflows/test.yaml
```diff
@@ -48,7 +48,7 @@ jobs:
         uses: actions/cache@v3
         with:
           path: /github/home/compiled
-          key: ${{ runner.os }}-ffmpeg-${{ hashFiles('**/install_ffmpeg.sh') }}
+          key: ${{ runner.os }}-ffmpeg-${{ hashFiles('install_ffmpeg.sh') }}
```
yondonfu, line 51: What was the reason for this change?

yondonfu (approved):

server/mediaserver_test.go
mjh1: Thanks @cyberj0g. Unfortunately for the tests the communication for the `EndTranscodingSession` call isn't set up so it simply times out after the 3 second [timeout](https://github.com/livepeer/go-livepeer/blob/d8ad5c7/server/rpc.go#L37), so that's the reason for waiting. And the other annoying thing is the goleak library starts off with very small microsecond pauses between retries so that's the reason for the high number. 
I've looked into a lot of different ways of abstracting this away for the tests without success, I think it'll have to wait for a later PR.

.github/workflows/test.yaml
mjh1: Builds were failing on master like this: https://github.com/livepeer/go-livepeer/actions/runs/4309780285/jobs/7517555365
It seemed as though the wildcard matching was not working. @hjpotter92 made some changes in this area recently so I'll check with him when he's back but the change I've made has fixed the issue in any case.

PR #2773, opened 2023-03-16
Verification attack proof of concept

cyberj0g:
### About
This is the first 'production' implementation of verification attack on Livepeer network (AFAIK). I believe, such efforts help highlight potential vulnerabilities, before they are exploited in the wild, thus, beneficial for open software. This PR showcases how trivial it is for Transcoder node to bypass local verification on the Broadcaster and compromise the network by:
- uploading fake transcoded segments
- receiving full reward without providing any computational resources

### Status
Production Livepeer network is currently vulnerable.

### Mitigation
- enable fast verification on all Broadcasters (it uses MPEG-7 signatures to validate segments and can't be bypassed trivially)
- make local verification more sophisticated e.g. add file size checks based on bitrate

#### Part 1: improving efficiency and maximizing rewards
- modify libx264 to work closer to a complete noop instead of encoding black frames
- as current local verification doesn't validate output resolution, we can maximize rewards by fitting total rendition pixel count to ticket value just below [MaxTicketEV](https://github.com/livepeer/livepeer-infra/blob/84968775efc0ddb4601423fb63b0beb7e756dd71/values/catalyst.values.yaml#LL103C2-L103C2)
#### Part 2: bypassing fast verification
**1. Exploiting the implementation**
With fast verification, rendition resolution and MPEG-7 signature are compared against results from a trusted node. However, MPEG-7 is *robust* against resolution changes (because it's a copyright violation detector, not the opposite). If we can trick basic video width and height check (with `avformat_find_stream_info`), we can produce a rendition at significantly lower resolution, which will still pass the signature check. One way to achieve that may be to inject a second SPS NAL unit after the first one, which would signal resolution decrease for the rest of the segment.

**2. Exploiting core algorithm**
I strongly suspect that MPEG-7 is not pre-image resistant. It's also known to be robust against resolution changes. It means that we may be able to a) generate desired signature from the source segment (or first downscale it, whichever is easier), b) knowing the signature, generate irrelevant video V' of desired resolution and with the same signature, optionally, computationally easier compared to actually downscaling the source video

### Installation and usage
Same as local installation of `go-livepeer` without Docker. No need for GPU or `-nvidia` flag.

### DIFF
--- a/core/capabilities_test.go
+++ b/core/capabilities_test.go
@@ -7,7 +7,6 @@ import (
 	"testing"
 	"time"
 
-	"github.com/livepeer/go-livepeer/common"
 	"github.com/livepeer/go-livepeer/net"
 	"github.com/livepeer/go-tools/drivers"
 	"github.com/livepeer/lpms/ffmpeg"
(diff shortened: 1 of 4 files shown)


AuthorityNull:
This is fascinating and I can't wait to see this hit production! Being rewarded while offering no computational resources has been an issue for a long time, and seeing it be addressed is exciting.

Issue #2235, opened 2022-02-04
Consider removing block backfilling


leszko:
**Is your feature request related to a problem? Please describe.**

Currently, at the Livepeer startup, we backfill all the blocks starting from the last block stored in the SQLite DB. If a node is offline for a long time, this process can take some time and execute a number of RPC calls to Ethereum API.

**Describe the solution you'd like**

After the move to Arbitrum, we can consider removing this backfilling mechanism, since it may not be necessary anymore. We can also consider removing polling missing blocks in `BlockWatcher.syncToLatestBlock()`.

**Describe alternatives you've considered**
N/A

**Additional context**

Comment from @yondonfu:

> I think we can even get away with not backfilling from the last seen block at all to avoid the situation where backfilling takes a long time if a node has been offline for a long time - if we're not backfilling from the start of the current round, this could take substantially longer. I don't see any immediate problems with doing that right now...

chrishobcroft:
I like this concept, and to reduce reliance on the sqlite DB. I would recommend understanding how the backfilled blocks are used locally, especially in the instances where the livepeer process is retrieving chain data from a remote `ethUrl` e.g. infura / alchemy / https://rinkeby.arbitrum.io/rpc, where data retrieval times are higher than running a `ethUrl` on localhost e.g. geth.

leszko:
Closing this one as technically it's the same work that needs to be done in #2298

leszko:
Design doc: https://www.notion.so/livepeer/Decrease-ETH-RPC-requests-16bda86cdd504dddaf5a46eb36937e86

PR #2775, opened 2023-03-21
Reduce number of ETH RPC calls during block polling

### LINKED ISSUE #2298: Decrease number of ETH RPC requests
A lot of orchestrator operators report that, since we moved to Arbitrum, they experience the rate limit error, caused by too many requests sent to the ETH RPC endpoints.

```
E0222 14:13:26.443007 1704401 block_watcher.go:121] blockwatch.Watcher error encountered - trying again on next polling interval err="429 Too Many Requests: {\"jsonrpc\":\"2.0\",\"error\":{\"code\":429,\"message\":\"Your app has exceeded its compute units per second capacity. If you have retries enabled, you can safely ignore this message. If not, check out https://docs.alchemyapi.io/guides/rate-limits\"},\"id\":3811}"
^CI0222 14:43:23.106921 1704401 livepeer.go:1088] Exiting Livepeer: interrupt
```

We should first consider removing (or limiting) the block backfilling as described in https://github.com/livepeer/go-livepeer/issues/2235. That may have the greatest impact on the number of RPC endpoints.

The ETH RPC requests are mainly caused by polling chain blocks by `BlockWatcher`. We need to investigate if it's possible to reduce the blocks we poll.

N/A

**Additional context**

Here are some Discord links:
- [Link 1](https://discord.com/channels/423160867534929930/426114749370204170/945680415996379159)
- [Link 2](https://discord.com/channels/423160867534929930/429350403272998913/944059535415472240)
- [Link 3](https://discord.com/channels/423160867534929930/429350403272998913/942985575298895904)

leszko:

Reduce the number of ETH RPC calls during the block polling operations.

From my measurements, this change (together with changing `blockPollingInterval` from `10` to `20`) should bring the RPC req reduction by the factor of **7-8x**.

The main improvement is that I removed the `pollNextBlock()` and `buildCanonicalChain()` parts, which I believe are not important after the migration to Arbitrum. The rationale is that Ethereum had a different approach to the block finality than Arbitrum:
- Ethereum - blocks could be reorganized, so at some point in time we could have blocks: `123`, `234`, `345`, but after some time they could be reorganized to `123`, `234`, `456`. 
- Arbitrum - it uses sequencer, so the blocks produced by each validator are always the same, therefore there is no reorganization happening, so if we have blocks `123`, `234`, `345`, they will always stay this way

I believe this whole code part with `pollNextBlock()` was so complex because the block reorg needed to be supported.

- Remove `pollNextBlock()` and `buidCanonicalChain()`
- Increase `maxBlocksInGetLogsQuery` from `60` to `1000` (this limit was too low which resulted in multiple `eth_getLogs` calls instead of one call
- Remove redundant calls to get the most recent blocks (in one operation we made the same ETH RPC call multiple times)

Tested with local runs (running for a long time).

**Further improvements**
The code could be refactored even further, but I don't want to do it in this PR, since I'd like to limit the number of changes done here to the minimum that brings good results. Additional possible improvements:
- Remove all the `stack` and its operations ([stack.go](https://github.com/livepeer/go-livepeer/blob/4b6ede31f040a084ff9e55bd798a0d6fffee1e1b/eth/blockwatch/stack.go#L29))
- Remove block backfilling at the startup (https://github.com/livepeer/go-livepeer/issues/2235)
- Reduce the number of req after a new round is initialized (we're currently making around 700 req, which seems too many)

fix #2298

**Checklist:**


thomshutt (approved):
This looks great, thanks for the detailed explanation of why we've done each piece. If this works then it's also a big win in terms of reducing our code complexity!

mjh1 (approved):

PR #2781, opened 2023-04-04
Add automatic session limit and set max sessions with livepeer_cli

### LINKED ISSUE #1935: Enable `-maxSessions` to be changed without requiring restart
## Is your feature request related to a problem? Please describe.

As the Livepeer network grows, active Orchestrators will likely wish to learn how to use `-maxSessions` to their own benefit, and to the benefit of the network.

Currently, it is not possible to change `-maxSessions` without restarting the software with a different value for the `-maxSessions` parameter in the startup command.

This presents a problem that a successful Orchestrator who is, for example, orchestrating 8 concurrent streams, will need to restarts their node in order to increase to a higher number.

This will result in a poor user experience for anyone watching the streams being transcoded by that Orchestrator when the process is restarted.

## Describe the solution you'd like

First, to show the `-maxSessions` parameter value in the `NODE STATS` for an Orchestrator in `livepeer_cli`.

Secondly, to provide an option in the `livepeer_cli` for an Orchestrator to be able to change this without restarting their node.

## Describe alternatives you've considered

To continue with the process as is, and recommend Orchestrators wait until they are processing 0 streams before restarting their node to change this parameter. This doesn't feel like a very elegant solution though.

eliteprox:

Defines a new "auto session limit" for orchestrators based on the combined remote transcoder capacity that can be set by using `-maxSessions auto` or the livepeer_cli option 21.
 
- Added `-maxSessions auto` as an option for orchestators.
- Added "Set max sessions" option to livepeer_cli as an option for orchestrators to change the session limit without restarting

- Observed `livepeer_max_sessions_total` while adding and removing transcoders with `-maxSessions auto`
- Observed `livepeer_max_sessions_total` while using Option 21 Set Max Sessions in livepeer_cli to various numbers and 'auto'
- Verified other roles like `-broadcaster` and `-transcoder` are not affected by the session limit changes

Fixes #1935 

**Checklist:**


cyberj0g:
Great work! A small note on `auto` value for `maxSessions`. I think it's a bit misleading. For O, this parameter adds an explicit restriction on top of already automatic capacity management. Maybe we should just have value of 0 (no restrictions) by default, instead of adding `auto` value? On the other hand, it would be really useful, if T could automatically detect max number of sessions, as part of startup sequence, where we already testing for available capabilities. The code could be reused from `livepeer_bench` utility. Maybe track is as a separate issue?

eliteprox:
> Maybe we should just have value of 0 (no restrictions) by default, instead of adding auto value?

I agree we should do this, I considered it when coding because the type conversions from string to int are a bit of overhead. 

I think we can apply a "default session limit" of 0 sessions for orchestrators, and make "0" the indicator for `auto`. For broadcasters and transcoders the default limit would remain at 10. I can make this change, would you agree? 

> On the other hand, it would be really useful, if T could automatically detect max number of sessions, as part of startup sequence, where we already testing for available capabilities.

I agree, this is something we could build on a separate issue. I thought of making it so that transcoders can change their session limits while running, but it seemed problematic because the `auto` session limit on the orchestrator would not update accordingly until the transcoder re-connected.

chrishobcroft:
Please can you link the issue that this responds to?

eliteprox:
> Please can you link the issue that this responds to?

https://github.com/livepeer/go-livepeer/issues/1935 Thanks for your review and feedback :)

leszko:
@eliteprox I'm catching up with PRs, is this PR ready for review?

If yes, then please fix the CI builds failures, update the branch, and assign me as reviewer. Thanks!

cmd/livepeer/livepeer.go
```diff
@@ -227,3 +228,16 @@ func updateNilsForUnsetFlags(cfg starter.LivepeerConfig) starter.LivepeerConfig
 
 	return res
 }
+
+// Defaults session limit to "0" (auto) for orchestrators when -maxSessions flag not set
+func defaultOrchestorLimit(cfg starter.LivepeerConfig) starter.LivepeerConfig {
```
leszko:
```suggestion
func defaultOrchestratorMaxSessions(cfg starter.LivepeerConfig) starter.LivepeerConfig {
```

cmd/livepeer/livepeer.go
```diff
@@ -227,3 +228,16 @@ func updateNilsForUnsetFlags(cfg starter.LivepeerConfig) starter.LivepeerConfig
...
+	res := cfg
+
+	isFlagSet := make(map[string]bool)
+	flag.Visit(func(f *flag.Flag) { isFlagSet[f.Name] = true })
+	if isFlagSet["orchestrator"] && !isFlagSet["maxSessions"] {
+		*cfg.MaxSessions = 0
```
leszko: What happens in the case of a standalone O? Won't it now stop accepting any transcoding by default (since its max sessions is set to 0)?

cmd/livepeer/livepeer.go
```diff
@@ -227,3 +228,16 @@ func updateNilsForUnsetFlags(cfg starter.LivepeerConfig) starter.LivepeerConfig
 
 	return res
 }
+
+// Defaults session limit to "0" (auto) for orchestrators when -maxSessions flag not set
```
leszko: How about being more explicit and having a value `--maxSessions auto`? Otherwise, I'm afraid of not keeping the backwards-compatibility for all cases.

core/orchestrator.go
```diff
@@ -708,9 +708,18 @@ func (n *LivepeerNode) serveTranscoder(stream net.Transcoder_RegisterTranscoderS
 	coreCaps := CapabilitiesFromNetCapabilities(capabilities)
 	n.Capabilities.AddCapacity(coreCaps)
 	defer n.Capabilities.RemoveCapacity(coreCaps)
+
+	if n.AutoSessionLimit {
+		n.SetMaxSessions(n.GetCurrentCapacity() + capacity)
```
leszko, line 713: How does it work when you first start the O and then attach T? Won't this value be then incorrect?

leszko (changes requested):

eliteprox:
> How about being more explicit and having a value `--maxSessions auto`? Otherwise, I'm afraid of not keeping the backwards-compatibility for all cases. 

I agree, I will revert this back to `-maxSessions auto`  and make it so that maxSessions defaults to 10

cmd/livepeer/livepeer.go
eliteprox: Good catch, I tested and confirmed this was an issue. I have removed the default auto session limit.

cmd/livepeer/livepeer.go
eliteprox: I have reverted back to -maxSessions auto instead of using `0`

core/orchestrator.go
eliteprox: When we first start the orchestrator with `-maxSessions auto`, the session limit will be 0. By default, without `-maxSessions` it will default to 10.

eliteprox:
Tests are passing on local test suite

cmd/livepeer/livepeer.go
eliteprox: Removed this code, decided not to make auto session limit the default

cmd/livepeer/starter/starter.go
```diff
@@ -290,9 +291,20 @@ func DefaultLivepeerConfig() LivepeerConfig {
...
+			glog.Fatal("-maxSessions 'auto' cannot be used when both -orchestrator and -transcoder are specified")
+			return
+		}
+		core.MaxSessions = 0
+	} else {
+		intMaxSessions, _ := strconv.Atoi(*cfg.MaxSessions)
+		if intMaxSessions <= 0 {
+			glog.Fatal("-maxSessions must be 'auto' or greater than zero")
+			return
+		}
```
```suggestion
		intMaxSessions, err := strconv.Atoi(*cfg.MaxSessions)
		if err != nil || intMaxSessions <= 0 {
			glog.Fatal("-maxSessions must be 'auto' or greater than zero")
			return
		}
```
Please handle the err here.

cmd/livepeer/starter/starter.go
```diff
@@ -746,6 +760,7 @@ func StartLivepeer(ctx context.Context, cfg LivepeerConfig) {
 				}
 			}
 
+			n.AutoSessionLimit = *cfg.MaxSessions == "auto"
```
leszko, line 761: Do we need it here if it's already set in line 416?

cmd/livepeer_cli/wizard_transcoder.go
```diff
@@ -338,3 +338,28 @@ func (w *wizard) setPriceForBroadcaster() {
 	}
 
 }
+
+func (w *wizard) setMaxSessions() {
+	fmt.Println("Enter the maximum # of sessions")
+	// maxSessions := w.readStringAndValidate(10)
```
leszko: Please remove the commented-out code.

core/orchestrator.go
leszko: 👍

leszko (commented):
Added some final nit comments. Other than that it looks good 👍 

Thanks for the PR.

cmd/livepeer/starter/starter.go
eliteprox: Removed from 416, it seems more readable to keep it here

leszko (approved):

PR #2800, opened 2023-05-30
CI: Add editorconfig-checker workflow when testing the project

hjpotter92:
Adds a job to check against `editorconfig-checker` for file not following editorconfig guide

- Runs editorconfig-checker against the project

locally + this PR

nope


.github/workflows/test.yaml
```diff
@@ -140,3 +140,21 @@ jobs:
...
         uses: github/codeql-action/analyze@v2
+
+  editorconfig:
+    name: Run editorconfig checker
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v3.5.2
+        with:
+          fetch-depth: 0
```
pwilczynskiclearcode: `fetch-depth: 0` is for fetching entire git history/branches. It's not needed here, right?

.github/workflows/test.yaml
```diff
@@ -140,3 +140,21 @@ jobs:
 
       - name: Perform CodeQL Analysis
         uses: github/codeql-action/analyze@v2
+
+  editorconfig:
```
pwilczynskiclearcode, line 144: Maybe we can just extend `codeql` with `editorconfig-checker`?

docker/Dockerfile.debian
```diff
@@ -1,48 +1,53 @@
...
+
+COPY --from=builder2	/root/livepeer	/usr/bin/livepeer
 
 COPY docker/start.sh .
-RUN chmod +x start.sh
 
-EXPOSE 7935/tcp
-EXPOSE 8935/tcp
-EXPOSE 1935/tcp
+RUN	chmod +x start.sh
```
pwilczynskiclearcode: `COPY --chmod=777 docker/start.sh .`

pwilczynskiclearcode (approved):

.github/workflows/test.yaml
hjpotter92: codeql takes ~7m to run; editorconfig can be run parallel to that job, and finishes in less than a minute!

.github/workflows/test.yaml
hjpotter92: yeah, will update

docker/Dockerfile.debian
hjpotter92: my original instinct was to just delete the whole `Dockerfile.debian`. not sure if it is useful anyway!

cc: @thomshutt / @leszko is the `Dockerfile.debian` used anywhere? it looks to be something made for local development but any reason that the official `livepeer/go-livepeer` image (and its associated `Dockerfile`) is not used for same?

leszko: I believe it's used only in the `localdocker`: https://github.com/livepeer/go-livepeer/blob/master/Makefile#L71

PR #2815, opened 2023-06-22
Add public logs for Broadcaster instrospection

emranemran:

**clog: add new logging methods to publish a set of public logs**
The O community has shown interest in B introspection where some logging
can be published to a public facing Loki instance that can be used to
debug the selection algorithm. To that end, this change adds a few
methods to the clog package to publish logs with a specific tag
([PublicLogs]) and removes any contextual info that should not be
published.

To log specific lines to the public facing Loki instance, one can do the
following as an example:
 - clog.PublicInfof(ctx, "Updating O ID=%v, reason=%v", "fooID", "bar")

The resulting log will have specific contextual info like manifestID,
orchSessionID, etc pre-pended along with a [PublicLogs] tag.

- Update existing clog package to add new methods:  `PublicInfof`, `PublicCloneCtx`
- Update definition of existing `infof`, `formatMessage` methods to accept a boolean that indicates if the log is supposed to be public or internal facing.
- Updated/added tests

- Unit tests added and all tests pass. Specific tests were added to ensure only non-approved contextual info is not being leaked.

Fixes VID-248

**Checklist:**

- [X] All tests in `./test.sh` pass --> (**NOTE:** this never passed locally on my machine but the CI tests pass)

### DIFF
--- a/server/broadcast.go
+++ b/server/broadcast.go
@@ -205,7 +205,7 @@ func removeSessionFromList(sessions []*BroadcastSession, sess *BroadcastSession)
 	return res
 }
 
-func selectSession(sessions []*BroadcastSession, exclude []*BroadcastSession, durMult int) *BroadcastSession {
+func selectSession(ctx context.Context, sessions []*BroadcastSession, exclude []*BroadcastSession, durMult int) *BroadcastSession {
 	for _, session := range sessions {
 		// A session in the exclusion list is not selectable
 		if includesSession(exclude, session) {
@@ -216,6 +216,15 @@ func selectSession(sessions []*BroadcastSession, exclude []*BroadcastSession, du
 		// threshold is selectable
 		if len(session.SegsInFlight) == 0 {
 			if session.LatencyScore > 0 && session.LatencyScore <= SELECTOR_LATENCY_SCORE_THRESHOLD {
+				clog.PublicInfof(ctx,
+					"Selecting new orchestrator, reason=%v",
+					fmt.Sprintf(
+						"performance: no segments in flight, latency score of %v < %v",
+						session.LatencyScore,
+						durMult,
+					),
+				)
+
 				return session
 			}
 		}
@@ -230,6 +239,15 @@ func selectSession(sessions []*BroadcastSession, exclude []*BroadcastSession, du
 			maxTimeInFlight := time.Duration(durMult) * oldestSegInFlight.segDur
 
 			if timeInFlight < maxTimeInFlight {
+				clog.PublicInfof(ctx,
+					"Selected orchestrator reason=%v",
+					fmt.Sprintf(
+						"performance: segments in flight, latency score of %v < %v",
+						session.LatencyScore,
+						durMult,
+					),
+				)
+
 				return session
 			}
 		}
@@ -258,7 +276,7 @@ func (sp *SessionPool) selectSessions(ctx context.Context, sessionsNum int) []*B
 
 		// Re-use last session if oldest segment is in-flight for < segDur
 		gotFromLast := false
-		sess = selectSession(sp.lastSess, selectedSessions, 1)
+		sess = selectSession(ctx, sp.lastSess, selectedSessions, 1)
 		if sess == nil {
 			// Or try a new session from the available ones
 			sess = sp.sel.Select(ctx)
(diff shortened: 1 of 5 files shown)


server/broadcast.go
```diff
@@ -56,6 +56,8 @@ var submitMultiSession = func(ctx context.Context, sess *BroadcastSession, seg *
 	go submitSegment(ctx, sess, seg, segPar, nonce, calcPerceptualHash, resc)
 }
 
+var PublicLogging bool = false
```
thomshutt: I don't think we need this now that we're going down the promtail route

server/broadcast.go
```diff
@@ -216,6 +216,16 @@ func selectSession(sessions []*BroadcastSession, exclude []*BroadcastSession, du
 		// threshold is selectable
 		if len(session.SegsInFlight) == 0 {
 			if session.LatencyScore > 0 && session.LatencyScore <= SELECTOR_LATENCY_SCORE_THRESHOLD {
+				clog.PublicInfof(ctx,
+					"Selected orchestrator reason=%v, eth-address=0x%v, ip-address=%v",
```
pwilczynskiclearcode: It might be better to put `reason=` at the end of log as it contains commas and spaces. It might be easier for reading and parsing.

server/broadcast.go
```diff
@@ -303,6 +323,13 @@ func (sp *SessionPool) selectSessions(ctx context.Context, sessionsNum int) []*B
 				sess.SegsInFlight = nil
 				sp.lastSess = removeSessionFromList(sp.lastSess, sess)
 				clog.V(common.DEBUG).Infof(ctx, "Removing orch=%v from manifestID=%s session list", sess.Transcoder(), sp.mid)
+				clog.Infof(ctx,
+					"Removing orchestrator=%v from session list, ethaddress=0x%v, ip-address=%v",
+					sess.Transcoder(),
+					ethcommon.Bytes2Hex(sess.OrchestratorInfo.TicketParams.Recipient),
+					sess.OrchestratorInfo.Transcoder,
+				)
```
pwilczynskiclearcode: So this won't be a public log? `clog.Infof`?

pwilczynskiclearcode (approved):

server/broadcast.go
thomshutt: Even better, can we have these as key/value fields rather than baked into the log line?

pwilczynskiclearcode: @thomshutt you mean to extract `eth-address` and `ip-address` as promtail/loki labels? We could log them as string in broadcaster and then add regex logic in `promtail` to extract `eth-address` and `ip-address` as labels.

emranemran: @thomshutt: yes, the plan is to include some of these fields as key/value fields in clog but we should do that only for ones that are repeatedly logged. I'm planning to use a follow-up PR to add more logs based on this [1] and we can determine which fields should be part of clog's key/value pairs. The other goal is to get this merged so that we can verify the full pipeline from logs to loki and then follow on with optimizations. 

@pwilczynskiclearcode: yep, will move reason to end. Let's hold off on labeling anything in promtail/loki until we have the initial pipeline working and verified. We can optimize based on feedback from the O community. 

lmk if this sounds ok to you guys. 

[1] https://github.com/ad-astra-video/go-livepeer/commit/ac3be68af55ad9772e65a6736e1e5086ee5b28e6#diff-6a8752e5207bc4096237c9a19e0520c8d238fab098e942920f95d4045f886f09

server/broadcast.go
emranemran: Oops good catch - this should be public log. Will fix.

emranemran:
Updates: 
* Addressed comments
* Planning to merge tomorrow. Any further changes can be addressed in a follow-up PR once we verify loki logging is working as expected.

server/broadcast.go
ad-astra-video: For what its worth, I updated to same approach.  I did some conformity updates in commits after this one that better followed the clog presentation.  Also added some updates to return information to level where clog context was active (e.g. selection) so could have better conformity.

https://github.com/ad-astra-video/go-livepeer/commit/af65b34728eec99189c4e301c8f7be6f2469253f

https://github.com/ad-astra-video/go-livepeer/commit/9d8d89b10a6ba81b4761d3bd32192d5bce7a7353

emranemran: @ad-astra-video: for this initial PR, I opted for simpler logs. Several unit tests were failing due to code races when attempting to log certain values (e.g. info related to payment tickets) where it was causing certain http responses to time out. 

Also unit tests caught a few panics which were getting triggered when trying to log values like `ethcommon.Bytes2Hex(session.OrchestratorInfo.TicketParams.Recipient)` when no segments were in-flight -- this was being caused by a null `session.OrchestratorInfo` object under certain conditions. I'm not fully ramped up on this code base yet so didn't quite understand all the conditions that can lead to this. I'll be spending some more time on this to get a better understanding. 

Once we verify logging to the public loki instance is working, we can follow up with further changes that will log all the values that will be most beneficial.

PR #2820, opened 2023-07-01
Fix segfault in setting a pricePerBroadcaster for 3 or more B's

### LINKED ISSUE #2819: Setting Broadcaster specific price segfault
It works well in general, but when specifying 3 or more B's using CMD flags the program segfaults. This is caused due to an error being thrown in the `ReadFromFile` function, where `os.Stat(s)` returns an error which was not being covered by `os.IsNotExist(err)`. After that `info.IsDir()` is called but `info` is undefined at that point

https://github.com/livepeer/go-livepeer/blob/8bd65fb984ab611e1343df713bae231b8119090a/cmd/livepeer/starter/starter.go#L1413

https://github.com/livepeer/go-livepeer/blob/8bd65fb984ab611e1343df713bae231b8119090a/common/readfromfile.go#L12-L19

Entire segfault:
```
I0701 09:42:34.148159       1 db.go:458] Closing DB
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x1a087b2]
goroutine 26 [running]:
github.com/livepeer/go-livepeer/common.ReadFromFile({0x7ffc84ab8acc, 0x131})
	/common/readfromfile.go:19 +0x72
github.com/livepeer/go-livepeer/cmd/livepeer/starter.getBroadcasterPrices({0x7ffc84ab8acc, 0x131})
	/cmd/livepeer/starter/starter.go:1444 +0x49
github.com/livepeer/go-livepeer/cmd/livepeer/starter.StartLivepeer({_, _}, {0xc00052e2f0, 0xc00052e300, 0xc00052e310, 0xc00052e320, 0xc00052e330, 0xc00052e340, 0xc00052e350, 0xc00052e410, ...})
	/cmd/livepeer/starter/starter.go:766 +0x3fb3
main.main.func1()
	/cmd/livepeer/livepeer.go:97 +0x58
created by main.main
	/cmd/livepeer/livepeer.go:96 +0xca5
```

Proposed solution is to return the string as-is if the `ReadFromFile` errors

Edit: this issue probably only occurs when using CMD flags to

stronk-dev:
Fixes https://github.com/livepeer/go-livepeer/issues/2819
Maybe we want to add some more checks or logging of what the error is
cc @eliteprox


leszko:
@stronk-dev could you fix `go fmt`, so that CI build would not fail?

leszko (approved):

common/readfromfile.go
```diff
@@ -11,6 +11,9 @@ import (
 // A valid string will always be returned, regardless of whether an error occurred.
 func ReadFromFile(s string) (string, error) {
 	info, err := os.Stat(s)
+	if err != nil {
+		return s, err
+	}
 	if os.IsNotExist(err) {
```
thomshutt: Doesn't this check now become redundant?

stronk-dev: true, removed it

stronk-dev:
Seems like one of the unit tests is failing now, as the `ReadFromFile` function used to return a `nil` error for non-existing files/directories. It now returns a `PathError` instead.
Should I modify the unit test to expect a `PathError`?

thomshutt:
> Should I modify the unit test to expect a `PathError`?

Yeah, I think that makes sense

thomshutt:
@stronk-dev I don't think it's caused by this PR but if you can figure out what's going on with the tests then that'd be great

stronk-dev:
It seems to fail in the `TestHTTPPushBroadcaster` where it tries to send segments to the broadcaster, but it returns a `503` error code. There are some errors before it fails the test:
```
E0704 13:23:40.096942   17647 roundinitializer.go:90] ErrRoundInitialized
E0704 13:23:53.983949   17647 db_discovery.go:331] Could not get orchestrator orch=[https://127.0.0.1:8938:](https://127.0.0.1:8938/) rpc error: code = Unknown desc = insufficient sender reserve
E0704 13:23:57.133644   17647 mediaserver.go:1033] manifestID=71abd7357f1325b54da0 nonce=17646280690537945479 seqNo=0 clientIP=127.0.0.1 No sessions available name=0.ts url=http://127.0.0.1:8939/live/71abd7357f1325b54da0/0.ts
E0704 13:23:57.148748   17647 gaspricemonitor.go:122] error getting gas price: Post "http://172.17.0.1:32774/": EOF
E0704 13:23:57.366899   17647 webserver.go:22] http: Server closed
E0704 13:23:57.367317   17647 webserver.go:22] http: Server closed
```

For the rest I'm not too familiar with the code or how these tests are ran to be able to fix it. From where the modified `ReadFromFile` function gets called, the returned `error` value isn't even used so not sure how these tests suddenly fail now

leszko:
@thomshutt I think the CI passes now and I approved the PR, do you want to re-review or should I merge?

thomshutt:
@leszko let's merge!

PR #2821, opened 2023-07-01
Bump nvidia cuda docker image version

stronk-dev:
Old version was removed, docker builds require a version bump. Not sure if this breaks anything

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -25,6 +25,7 @@ insert_final_newline = unset
 # Makefiles/Dockerfile/golang files
 [{Makefile,Dockerfile{,.debian},*.go}]
 indent_style = tab
+indent_size = 8
 
 # YAML/JSON Files
 [{.ecrc,*.{yml,yaml,sh,json}}]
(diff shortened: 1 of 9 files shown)


leszko:
I see that it increased the Docker image size from `1.29GB` to `2.38GB`. Isn't that an issue? CC: @thomshutt @hjpotter92

hjpotter92:
> I see that it increased the Docker image size from `1.29GB` to `2.38GB`. Isn't that an issue? CC: @thomshutt @hjpotter92

image sizes aren't a problem now. we are provisioning our machines to use the entire SSD storage; instead of earlier 50GB as `/` partition and rest as `/data`. each server now has atleast 500GB of disk space to store docker images in cache and the persitant storage is handled separately (if needed).

thomshutt:
@hjpotter92 that's true, but it still makes it more unwieldy to work with, slower to build etc. and so if there is a way to keep it down then that's probably worth exploring

leszko:
I see the version `12.x` is small, so we may upgrade to it. Just need to think how to test it.

hjpotter92:
I believe the last time this was discussed, @iameli raised an issue with `ubuntu18.04` also getting bumped. if we're planning to keep ubuntu18.04 as the supported runtime version, we should go with 11.6.2 as that has [the smallest image size](https://hub.docker.com/r/nvidia/cuda/tags?page=1&ordering=last_updated&name=8-runtime-ubuntu18.04).

however, if we're also planning to ditch the now EOL-ed ubuntu18.04, should we also consider ubuntu20? I believe when we reprovisioned our infra, the machines were installed with ubuntu20 (and that had been the concern last time this image was bumped).

leszko (approved):

PR #2837, opened 2023-08-09
Set a floor of 1.5s for threshold to switch Os

thomshutt:
We currently have a threshold of `1*<segment duration>` for decided when to swap to another Orchestrator, to ensure we're transcoding in realtime. 

This has been observed to be overly aggressive for 1s segments however, where even high-performing Orchestrators will sit very close to the limit.

Rather than incurring the overheads of swapping for Os that are otherwise keeping up, we're setting a floor of 1.5s for transcoding duration.

- Set a floor of 1.5s on transcode duration before swapping

- I'll deploy to Staging, push through a stream with 1s segments and observe swapping

No

**Checklist:**

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -9,6 +9,7 @@
 #### General
 
 #### Broadcaster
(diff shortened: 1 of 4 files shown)


leszko (approved):
I think this is ok as a quick fix. 👍

PR #2838, opened 2023-08-10
docker.yaml: Split `builder` image generation to another workflow job

hjpotter92:
- Cleanup hosted runner further as suggested on discord [1]

[1]: https://discord.com/channels/423160867534929930/750796877762527262/1138806765648105544

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -31,7 +31,7 @@ jobs:
 
       - name: Cleanup hosted runner
         run: |
-          sudo apt purge -yqq dotnet-* mono-* llvm-* adoptopenjdk-* libllvm* powershell* openjdk-* \
+          sudo apt purge -yqq dotnet-* mono-* llvm-* libllvm* powershell* openjdk-* \
           temurin-* mongodb-* firefox mysql-* \
           hhvm google-chrome-stable \
           libgl1-mesa-dri microsoft-edge-stable \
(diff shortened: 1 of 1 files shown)


.github/workflows/docker.yaml
```diff
@@ -89,6 +84,79 @@ jobs:
             type=raw,value=${{ github.event.pull_request.head.ref }}
             type=raw,value=stable,enable=${{ startsWith(github.event.ref, 'refs/tags/v') }}
 
+      - name: Build and push livepeer docker image
+        uses: docker/build-push-action@v4
+        with:
+          build-args: |
+            BUILD_TAGS=${{ steps.build-tag.outputs.build-tags }}
+          # linux/arm64 is disabled until
```
leszko: The related ticket is resolved, should we remove this comment?

.github/workflows/docker.yaml
```diff
@@ -89,6 +84,79 @@ jobs:
...
+
+      - name: Cleanup hosted runner
+        run: |
+          sudo apt purge -yqq dotnet-* mono-* llvm-* adoptopenjdk-* libllvm* powershell* openjdk-* \
+          temurin-* mongodb-* firefox mysql-* \
+          hhvm google-chrome-stable \
+          libgl1-mesa-dri microsoft-edge-stable \
+          google-cloud-sdk azure-cli
+          sudo apt autoremove -y
+          sudo rm -rf /usr/share/dotnet /usr/local/lib/android
```
leszko, line 125: Do we need to remove it both here and in line 40?

leszko (approved):

.github/workflows/docker.yaml
hjpotter92: not really! but just keeping it here for future-proofing against having similar issues (out of disk space on github runner). go-livepeer won't be needing android libraries (and other packages being removed).

.github/workflows/docker.yaml
hjpotter92: you're right. updated.

PR #2841, opened 2023-08-17
update test workflow to ubuntu 20.04

ad-astra-video:
Updates test workflow to use ubuntu 20.04

- bumped ubuntu version to 20.04
- used Dockerfile to update installation steps
- removed folders not needed to create space on runner

Ran on my fork of go-livepeer https://github.com/ad-astra-video/go-livepeer/actions/runs/5887219287

No

**Checklist:**

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -19,11 +19,14 @@ jobs:
     name: Run tests defined for the project
     runs-on: ubuntu-latest
     container:
-      image: ubuntu:18.04
+      image: ubuntu:20.04
     steps:
       - name: Setup ubuntu container
         run: |
           apt update
+          export TZ=Etc/Utc
+          ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
+          export DEBIAN_FRONTEND=noninteractive
           apt install -yqq build-essential make software-properties-common
           add-apt-repository -y ppa:git-core/ppa
           apt update && apt install -yqq git zip unzip zlib1g-dev zlib1g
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yaml
```diff
@@ -19,11 +19,14 @@ jobs:
...
     container:
-      image: ubuntu:18.04
+      image: ubuntu:20.04
     steps:
       - name: Setup ubuntu container
         run: |
           apt update
+          export TZ=Etc/Utc
+          ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
+          export DEBIAN_FRONTEND=noninteractive
```
hjpotter92: in my experience, github workflows work better if instead of `export`, we do one of:
```sh
echo TZ=Etc/UTC >> $GITHUB_ENV
```
or specify it in the job itself:
```yaml
env:
  TZ: "Etc/UTC"
```

hjpotter92 (commented):
perhaps we should also consider bumping our build pipeline to ubuntu 20 (or 22?)! wdyt @leszko / @thomshutt 
https://github.com/livepeer/go-livepeer/blob/master/.github/workflows/build.yaml#L19-L20

.github/workflows/test.yaml
ad-astra-video: I have "fixed" this by removing it as part of a larger change to moving from `ubuntu-latest` with `ubuntu:18.04` container specified to using `runs-on: ubuntu-20.04`.

thomshutt:
> 



> perhaps we should also consider bumping our build pipeline to ubuntu 20 (or 22?)! wdyt @leszko / @thomshutt
> https://github.com/livepeer/go-livepeer/blob/master/.github/workflows/build.yaml#L19-L20

If it's not too much work then definitely!

hjpotter92:
hey @ad-astra-video ; i've created a new PR which bumps our ubuntu versions for build workflow as well: #2846

i've forked off from your branch! closing this PR in favour of that one.

thanks for your contribution

PR #2843, opened 2023-08-18
segment_rpc: increase log level of payment creation. it's useful!

iameli:
Increases the log level of payment creation from 6 to 3 - a request from @ecmulli for payment tracking


ecmulli (approved):

ad-astra-video:
I agree! Makes it easy to pick up payments sent on the B from the logs.

PR #2846, opened 2023-08-24
update test and build workflows to ubuntu 20.04

hjpotter92:
Updates test and build workflows to use ubuntu 20.04

- bumped ubuntu version to 20.04
- used Dockerfile to update installation steps
- removed folders not needed to create space on runner

this PR workflow runs

old PR this one is based off on: #2841 

**Checklist:**

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -16,8 +16,6 @@ jobs:
   linux-build:
     name: Build binaries for ${{ matrix.platform.name }}-${{ matrix.arch }}
     runs-on: ${{ matrix.platform.runner }}
-    container:
-      image: ubuntu:18.04
     strategy:
       fail-fast: false
       matrix:
@@ -26,9 +24,9 @@ jobs:
           - amd64
         platform:
           - name: linux
-            runner: ubuntu-latest
+            runner: ubuntu-20.04
           - name: windows
-            runner: ubuntu-latest
+            runner: ubuntu-20.04
         exclude:
           - platform:
               name: windows
@@ -37,10 +35,10 @@ jobs:
     steps:
       - name: Setup ubuntu container
         run: |
-          apt update
-          apt install -yqq build-essential make software-properties-common
-          add-apt-repository -y ppa:git-core/ppa
-          apt update && apt install -yqq git zip unzip zlib1g-dev zlib1g
+          sudo apt update
+          sudo apt install -yqq build-essential make software-properties-common
+          sudo add-apt-repository -y ppa:git-core/ppa
+          sudo apt update && sudo apt install -yqq git zip unzip zlib1g-dev zlib1g
 
       - name: Check out code
         uses: actions/checkout@v3.5.3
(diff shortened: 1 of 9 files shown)


.github/workflows/build.yaml
```diff
@@ -81,16 +83,16 @@ jobs:
 
       - name: Install dependencies
```
pwilczynskiclearcode, line 78: Would it make sense to merge it with `Setup ubuntu container` step?

thomshutt:
Massively in favour of this! Looks like the build is failing at the moment though

.github/workflows/build.yaml
```diff
@@ -265,7 +270,7 @@ jobs:
         id: cache-ffmpeg
```
pwilczynskiclearcode, line 260: Should we cache binaries `~/build` like in `linux-build:`

.github/workflows/test.yaml
```diff
@@ -48,22 +51,29 @@ jobs:
...
+          path: /home/runner/compiled
           key: ${{ runner.os }}-ffmpeg-${{ hashFiles('install_ffmpeg.sh') }}
           restore-keys: |
             ${{ runner.os }}-ffmpeg
 
+      - name: Install protobuf
+        run: |
+          go install github.com/golang/protobuf/protoc-gen-go@v1.3.5
+          sudo apt update
+          sudo apt install -yqq protobuf-compiler
```
pwilczynskiclearcode, line 63: can be merged with `setup ubuntu container`?

.github/workflows/build.yaml
hjpotter92: the 'setup ubuntu container' was (and probably is) a temporary step i had to add since github had removed support for ubuntu-18 runners. i do not think we need that step, but just keeping it for future when ubuntu 20 support might go away and we might need to readd the
```yaml
    container:
      image: ubuntu:18.04
```
directive for ubuntu-20 :|

.github/workflows/build.yaml
hjpotter92: oh, that step can be removed entirely. i don't think it is needed at all. it was just some experiment with artifact path resolution i had to do and forgot to remove it.

.github/workflows/build.yaml
```diff
@@ -277,21 +284,23 @@ jobs:
           key: ${{ runner.os }}-binaries-tf-${{ github.sha }}
 
       - name: Install dependencies
+        env:
+          LIBTENSORFLOW_VERSION: "2.12.1"
```
leszko, line 270: Do we still use tensorflow? Don't want to get rid of it everywhere?

leszko (approved):
Added one comment. It's fine with me, but please with merging to resolve all discussions and comments from @pwilczynskiclearcode

.github/workflows/build.yaml
thomshutt: Yeah, if we can tear it out without breaking builds then let's do it. I started trying in https://github.com/livepeer/go-livepeer/pull/2836 but haven't had time to finish it

hjpotter92: oh! that requires quite a lot of changes in go code... which i'd rather not touch in this small-ish PR. would rather bump 2836 with changes from this PR and keep tensorflow removal over in that alone.

PR #2848, opened 2023-08-29
Improve Transcode Quality

leszko:
Related to https://github.com/livepeer/lpms/pull/371


net/lp_rpc.proto
```diff
@@ -301,6 +301,7 @@ message VideoProfile {
     CHROMA_444 = 2;
   }
   ChromaSubsampling chromaFormat = 27;
+  uint32 crf = 28;
```
emranemran: Is this the default? I'd suggest setting this lower since it translates to a pretty high CQ.

emranemran: OK i see now after reading your comment in the other PR. But should this be 27 like the other PR?

leszko: This is not a value, just the field number in the probuf. So, in short, in just needs to be the next number after the line above.

leszko:
@emranemran as discussed, I renamed `crf` to `quality` everywhere. PTAL.

thomshutt (approved):

PR #2849, opened 2023-08-30
`GetOrchestrator` slowness when PM Sender data is not cached

stronk-dev:
**What does this pull request do?**

Titan node had some curious insights into the relationship between your Eth RPC provider and the response times to GetOrchestrator requests, and started investigating. @ad-astra-video assisted and they found out that during a GetOrchestrator request, it was possible that a blocking RPC call would be made to check the float of the Broadcaster

This commit is pulled from @ad-astra-video [fork of go-livepeer](https://github.com/ad-astra-video/go-livepeer/commit/fcbe7ddb641a371a318809592f18ed0ad64cd864) and it increases the timeout for how long B info gets cached and does some explicit calls to refresh data in the background every 10 minutes

My questions for the person reviewing this:
- Are these explicit update calls on the `cleanupInterval` required, or are the blockwatchers already keeping all relevant B data up-to-date as long as the PM sender is cached?
https://github.com/livepeer/go-livepeer/blob/951f6e637fb9323dde3b3add93464fbc26b14e4f/pm/sendermonitor.go#L329-L332
- If there is data we can keep cached this way, what would be a sensible `cleanupInterval` and are there any other calls we might want to do?


ad-astra-video:
Want to add log lines pulled from Titan Node orchestrator when testing this.  These log lines have been removed but are relevant to where in the code the OrchestratorInfo response generation slowed down.  You can also see every other step of generating the response takes 4ms of the 484ms.

When `autoAdjustPrice` is not set to `false` the same SenderInfo and ClaimedReserve calls are the slowdown point.

[log (2).txt](https://github.com/livepeer/go-livepeer/files/12480681/log.2.txt)

Titan-Node:
Been running this update for 48 hours, all streams and node stats look fine.

thomshutt:
Thanks @stronk-dev, I'm going to get someone to start looking at this today

thomshutt:
Looks like we have genuine test failures here, could you take a look please @stronk-dev?

ad-astra-video:
> Looks like we have genuine test failures here, could you take a look please @stronk-dev?

The tests were not updated yet.  We were looking for some feedback on moving the eth rpc calls in the OrchestratorInfo response generation to the background and change in cache time from 60 seconds to 1 week.  Also on the cleanup interval changed from 1 minute to 10 minutes.  There appears to be a blockWatcher that accounts for broadcaster reserve changes but the cache was so short at 60 seconds that it did not operate very long.  With Titan's testing at the 1-2 minute mark the OrchestratorInfo response would slow down because of the eth rpc calls.

The gist is this change can significantly speed up OrchestratorInfo response times that are required at each start of stream and periodically during streams.  The goal is to provide more reliable response times, and possibly, shorten the initial session pool building to speed up first segment transcoding.

leszko:
I'm trying to catch up on this issue. 

So, if I understand correctly, there is a bug that when we have GetOrchestrator call and an RPC request at the same time, then the GetOrchestrator response is delayed for some reason? Is that correct?

I'm actually not sure how changing the timeouts solves this issue. Could you provide more detailed description on that?

About your questions, I don't know, would need to dig into the code, but first I'm trying to understand the actual issue.

stronk-dev:
There's no bug, but due to the timeout being only 1 minute, it happens often that an Orchestrator has to do a blocking RPC call in order to respond to a GetOrchestrator request. In this case the Orchestrator needs to have up-to-date reserve info on the B making the GetOrchestrator request in order to respond

Depending on the speed, distance and availability of the L2 RPC provider (which can be expecially problematic for regions like Singapore) this causes long fluctuations in response times to GetOrchestrator requests. Sometimes it's high enough for the Orchestrator to go above the threshold a B has set to populate their working set of Orchestrators. [You can see here](https://grafana.stronk.tech/d/b8FvMmmVk/orchestrator-tracker?orgId=1&refresh=5s&from=now-30d&to=now&var-orchestrators=titan-node.eth) when Titan applied this patch and his response times immediately became excellent
 
Doing blocking RPC calls _during_ a request like that is inefficient anyway. It's a good idea to make sure any data an O needs to respond to B requests is always up-to-date and available. We're trying to figure out here:
- is increasing the timeout enough: will the block watchers keep any data we need up-to-date already? Or do we require the explicit calls added in this PR to the cleanup-check-loop? If so, is there other data we want to pre-fetch?

leszko:
Ok, I understand, thanks for the explanation Marco.

One thing I don't understand is why `GetOrchestrator` needs to know anything about the Broadcaster / Sender? Do we use any of this info in [getOrchestrator()](https://github.com/livepeer/go-livepeer/blob/d8ad5c7c9683bdf0569c1c3601416af360967840/server/rpc.go#L315)?

Other than that, my understanding is that, in theory, you should never need to refresh the cache because it's updated with the block feed [here](https://github.com/livepeer/go-livepeer/blob/12358f5082b18a3e200dd99534ea92f91baea258/eth/watchers/senderwatcher.go#L163). Not sure actually why we're cleaning up this cache at all 🤔 Saying that, I this PR is ok, to be on the safe side, we can refresh the cache from time to time.

I'll put some comments inline, other than that, I'd love to understand my question above ☝️

cmd/livepeer/starter/starter.go
```diff
@@ -54,9 +54,9 @@ var (
 	txCostMultiplier = 100
 
 	// The interval at which to clean up cached max float values for PM senders and balances per stream
-	cleanupInterval = 1 * time.Minute
+	cleanupInterval = 10 * time.Minute
 	// The time to live for cached max float values for PM senders (else they will be cleaned up) in seconds
-	smTTL = 60 // 1 minute
+	smTTL = 604800 // 1 week
```
leszko: How about setting it to something like 2 days and refreshing the values every round or every 1 day?

pm/sendermonitor.go
```diff
@@ -327,6 +327,8 @@ func (sm *LocalSenderMonitor) startCleanupLoop() {
 	for {
 		select {
 		case <-ticker.C:
+			sm.smgr.UpdateSenderInfos()
```
leszko: We can refresh the cache less frequently, like every day or at the beginning of each round. In theory, block feed should keep the cache always up-to-date. So you could move these 2 lines to a separate loop and have a different `ticker.C` for cleanup and a different one for updating the cache.

leszko (commented):
Added the comments. In general, I understand the issue and this PR, so I think we should get it merged.

stronk-dev:
> One thing I don't understand is why `GetOrchestrator` needs to know anything about the Broadcaster / Sender? Do we use any of this info in [getOrchestrator()](https://github.com/livepeer/go-livepeer/blob/d8ad5c7c9683bdf0569c1c3601416af360967840/server/rpc.go#L315)?

It generates B specific ticket parameters and has a special reject case if the Broadcaster has insufficient reserves to do a payout:

https://github.com/livepeer/go-livepeer/blob/d8ad5c7c9683bdf0569c1c3601416af360967840/server/rpc.go#L315-L326
https://github.com/livepeer/go-livepeer/blob/d8ad5c7c9683bdf0569c1c3601416af360967840/server/rpc.go#L348-L354
https://github.com/livepeer/go-livepeer/blob/f350fba28b717c6ff272ecc106e90c87a39a04d7/core/orchestrator.go#L226-L236
https://github.com/livepeer/go-livepeer/blob/f350fba28b717c6ff272ecc106e90c87a39a04d7/pm/recipient.go#L192-L201
https://github.com/livepeer/go-livepeer/blob/f350fba28b717c6ff272ecc106e90c87a39a04d7/pm/recipient.go#L264-L274
https://github.com/livepeer/go-livepeer/blob/f350fba28b717c6ff272ecc106e90c87a39a04d7/pm/sendermonitor.go#L155-L162

Not sure how important these ticket parameters are for the B. But if the SenderMonitor keeps this data up-to-date anyway, increasing the timeout would be a nice first step

cmd/livepeer/starter/starter.go
stronk-dev: The SenderMonitor sets a `lastAccess` which is used for clearing the cache, so I agree that the timeout doesn't have to be too high. Retaining B info for as long as they've sent a GetOrchestrator request in the past 2 days makes sense to me

ad-astra-video: I agree, the B sends GetOrchestrator requests at startup and every hour so as long as the B is up the cache should not clear on the Os.

https://github.com/livepeer/go-livepeer/blob/86b248dc6cb445132aa84ecf7440ecf71137dafe/discovery/db_discovery.go#L24

https://github.com/livepeer/go-livepeer/blob/86b248dc6cb445132aa84ecf7440ecf71137dafe/discovery/db_discovery.go#L42-L69

https://github.com/livepeer/go-livepeer/blob/86b248dc6cb445132aa84ecf7440ecf71137dafe/discovery/db_discovery.go#L241-L261

pm/sendermonitor.go
ad-astra-video: Refresh at new round start makes sense to me.  Looks like it is already looking for updated information at the start of each round. Did not see that my first time through it!

https://github.com/livepeer/go-livepeer/blob/86b248dc6cb445132aa84ecf7440ecf71137dafe/eth/watchers/senderwatcher.go#L280-L309

stronk-dev: Since `handleRoundEvent` already calls `GetSenderInfo` and `ClaimedReserve` on a new round, should these additions simply be dropped from `startCleanupLoop`?

stronk-dev:
In order to get this PR ready for merge, I think it's just 1 minor question ATM:

Since `handleRoundEvent` already calls `GetSenderInfo` and `ClaimedReserve` on a new round, should the newly added code in this PR which calls these functions simply be dropped from `startCleanupLoop`? Seems like everything should be fine, otherwise we could still at the extra timed loop to refresh that data specificially
cc @leszko

leszko:
> startCleanupLoop

Yes, in my opinion, we can remove it from `startCleanupLoop`.

stronk-dev:
Modified. I guess the added print for when a sender info gets cleared could be nice to keep in there

pm/sendermonitor.go
```diff
@@ -347,6 +347,7 @@ func (sm *LocalSenderMonitor) cleanup() {
 			v.subScope.Close() // close the maxfloat subscriptions
 			delete(sm.senders, k)
 			sm.smgr.Clear(k)
+			glog.Infof("sender cleared from cache addr: %v last access: %v ttl: %vs", hexutil.Encode(k.Bytes()), v.lastAccess, sm.cfg.TTL)
```
leszko: Nit: Maybe Debug logging level?

```suggestion
			glog.V(6).Infof("sender cleared from cache addr: %v last access: %v ttl: %vs", hexutil.Encode(k.Bytes()), v.lastAccess, sm.cfg.TTL)
```

leszko (approved):

leszko:
Waiting for the CI build and then I'll merge! Thanks for the PR @stronk-dev

PR #2851, opened 2023-08-31
Change scale_cuda to scale_npp

leszko:
Change ffmpeg installation to use CUDA SDK (needed to use the `scale_npp` filter).

It also removes tensorflow build, I believe we don't need it anymore.

Related work:
- LPMS PR: https://github.com/livepeer/lpms/pull/373
- [Linear issue](https://linear.app/livepeer/issue/VID-380/lpms-apply-proposed-ffmpeg-parameter-changes)
- [Design Doc](https://www.notion.so/livepeer/Transcode-Quality-investigation-improvements-11314327aa484fb6927a48b102bd54bf#b821f216259b40d78ebbff164d67c062)


docker/Dockerfile.cuda-base
```diff
@@ -0,0 +1,18 @@
+# livepeer/cuda-base:11.7.1-cudnn8-devel-ubuntu20.04
+#
+# Base CUDA Develop image which contains CUDA SDK libs for the following architectures: linux amd64, linux arm64, windows amd64
+#
+# To build this image you need the following steps:
+# 1. Download NVIDIA CUDA SDK for ARM64, extract, and copy into cuda/arm64/usr/local/cuda/
```
hjpotter92, line 6: you can use the `ADD` directive in `Dockerfile`. https://stackoverflow.com/a/52069435/1190388

docker/Dockerfile.cuda-base
```diff
@@ -0,0 +1,18 @@
...
+FROM nvidia/cuda:11.7.1-cudnn8-devel-ubuntu20.04
+
+COPY npp_win/lib/x64/* /usr/local/cuda/lib64/
+COPY npp_win/include/* /usr/local/cuda/include/
+
+RUN mkdir -p /usr/local/cuda_arm64/lib64/
+COPY cuda/arm64/usr/local/cuda/lib64/libnp* /usr/local/cuda_arm64/lib64/
+COPY cuda/arm64/usr/local/cuda/include /usr/local/cuda_arm64/include
+
+RUN ln -s /usr/local/cuda /usr/local/cuda_amd64
```
hjpotter92, line 18: should we be building/pushing this regularly as well? or does it need to be manual? (could add a comment in the Dockerfile if it needs to be manual always

leszko: I'd keep this step manual because :
1. We never change the cuda version, so we don't need to rebuild this image
2. It'd be relatively difficult to automate, because of exe files

docker/Dockerfile.cuda-base
leszko: Ohhh nice. I didn't know about it! I will leave it as it is right know and if at some point we decide to automate building cuda image, then I'll use this ADD + do cleanup of not needed files.

hjpotter92 (approved):

thomshutt (approved):

PR #2852, opened 2023-09-01
Remove race test

leszko:
These tests are flaky and currently are failing in 80-90% or time, which makes our CI unusable. Probably it means that we have some race conditions, but it has existed for such a long time that we should just accept this fact. We may revisit the race condition issues/tests when they start to become an issue in prod.

### DIFF
--- a/server/push_test.go
+++ b/server/push_test.go
@@ -1215,44 +1215,6 @@ func TestPush_ResolutionWithContentResolutionHeader(t *testing.T) {
 	server.rtmpConnections = map[core.ManifestID]*rtmpConnection{}
 }
 
-func TestPush_WebhookRequestURL(t *testing.T) {
-	assert := assert.New(t)
-
-	// wait for any earlier tests to complete
-	assert.True(wgWait(&pushResetWg), "timed out waiting for earlier tests")
-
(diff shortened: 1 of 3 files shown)


mjh1:
I assume all of these tests are also run without `-race` still? 😅 
Also would it not be possible to find out which ones are failing most of the time and only remove those?

thomshutt:
Yeah, I think this is a good move (to stop us getting used to tests being broken) but I'd agree with Max - we should look at either removing specific tests that are failing or removing some of the `-race` flags. From whenever I've looked into this before, it is pretty consistently the same test(s) that's failing

emranemran:
@leszko I think it's this is the flakiest test in the pipeline that we can silence: 
```
: TestPush_WebhookRequestURL (0.02s)
    testing.go:1446: race detected during execution of test
```

leszko:
Ok, changed to the more conservative approach and silenced only `TestPush_WebhookRequestURL` from `-race`.

@thomshutt @mjh1 @emranemran PTAL.

thomshutt (approved):

mjh1 (approved):

PR #2872, opened 2023-09-25
New Selection Algorithm

leszko:
**Changes:**
- New Selection Algorithm as discussed in [Design Doc](https://www.notion.so/livepeer/New-Selection-Algorithm-Design-Doc-62f08c68225f4bf9a4cea152fef65543) and [Discord](https://discord.com/channels/423160867534929930/1149467899346550826/1149467902991409192)
- Periodically fetching Orchestrator Performance Scores
- Probability Selection Algorithm, which calculates probability taking into account weighted stake, price, and randomness
- Added flags:
  - `-selectStakeWeight`, `-selectPriceWeight`: weights of selection algorithms params
  - `-selectPriceExpFactor`: expresses how significant a small change of price is for the selection algorithm; default 100
  - `-orchPerfStatsUrl`: URL to the test stream service, Livepeer Inc service is available at: `https://leaderboard-serverless.vercel.app/api/aggregated_stats`
  - `-minPerfScore`: filter for the minimal O's performance score to be accepted for the selection
  - `-region`: region where B is deployed; used to get the right performance scores

**Some notes about the approach**
- In this version, the selection algorithm is **not fully configurable**, all you can configure right now are: 1) weights of price, stake, randomness, 2) filter for performance score 3) factor for exp price factor; in the future we may want to make the selection algorithm fully configurable using webhook service (aka plugin) or by allowing to specify an equation as a flag like `selection=0.3*exp(price, 100)+0.4*rand+0.3*stake`, but I think we should wait for some more use-cases / feature request to decide on the best approach
- For the performance scores, they are not part of the weighted probability calculation, but a filter, so only Os that have a regional performance score higher than `-minPerfScore` are taken into consideration for the selection. Note that the performance score is calculated as defined in https://github.com/livepeer/leaderboard-serverless, so `perfScore = (RTT / segment_size) * success_rate`. At first, I thought we should use the RTT expressed in `ms`, but that's not a good metric, because RTT depends on the segment length, so the performance score sounds like a better metric for the performance
- Note that if **no orchestrator passes the performance score metric, then all orchestrators are allowed for the selection**. This may result in using Orchestrator with lower performance than defined, however I chose this approach, because I think in practice we'll only see no Orchestrators passing this filter, when Test Stream service is down or gives wrong results (which happened a few times in the past)

### DIFF
--- a/server/selection_algorithm.go
+++ b/server/selection_algorithm.go
@@ -6,28 +6,24 @@ import (
 	"math/rand"
 )
 
-type SelectionAlgorithm interface {
-	Select(addrs []ethcommon.Address, stakes map[ethcommon.Address]int64, prices map[ethcommon.Address]int64) ethcommon.Address
+type ProbabilitySelectionAlgorithm struct {
+	StakeWeight float64
+	PriceWeight float64
+	RandWeight  float64
+
+	PriceExpFactor float64
 }
 
-type probabilitySelectionAlgorithm struct {
-	stakeWeight float64
-	priceWeight float64
-	randWeight  float64
-
-	priceExpFactor float64
-}
-
-func (sa probabilitySelectionAlgorithm) Select(addrs []ethcommon.Address, stakes map[ethcommon.Address]int64, prices map[ethcommon.Address]int64) ethcommon.Address {
+func (sa ProbabilitySelectionAlgorithm) Select(addrs []ethcommon.Address, stakes map[ethcommon.Address]int64, prices map[ethcommon.Address]int64) ethcommon.Address {
 	probabilities := sa.calculateProbabilities(addrs, stakes, prices)
 	return selectBy(probabilities)
 }
 
-func (sa probabilitySelectionAlgorithm) calculateProbabilities(addrs []ethcommon.Address, stakes map[ethcommon.Address]int64, prices map[ethcommon.Address]int64) map[ethcommon.Address]float64 {
+func (sa ProbabilitySelectionAlgorithm) calculateProbabilities(addrs []ethcommon.Address, stakes map[ethcommon.Address]int64, prices map[ethcommon.Address]int64) map[ethcommon.Address]float64 {
 	pricesNorm := map[ethcommon.Address]float64{}
 	for _, addr := range addrs {
 		p := prices[addr]
-		pricesNorm[addr] = math.Exp(-1 * float64(p) / sa.priceExpFactor)
+		pricesNorm[addr] = math.Exp(-1 * float64(p) / sa.PriceExpFactor)
 	}
 
 	var priceSum, stakeSum float64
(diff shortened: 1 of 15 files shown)


server/selection_algorithm.go
```diff
@@ -0,0 +1,104 @@
...
+	MinPerfScore float64
+
+	StakeWeight float64
+	PriceWeight float64
+	RandWeight  float64
+
+	PriceExpFactor float64
+}
+
+func (sa ProbabilitySelectionAlgorithm) Select(addrs []ethcommon.Address, stakes map[ethcommon.Address]int64, prices map[ethcommon.Address]float64, perfScores map[ethcommon.Address]float64) ethcommon.Address {
```
thomshutt, line 23: 🙌  Really nice that you were able to split this out - it should make it a lot easier for people to tweak it in the future without understanding the entirety of go-livepeer

server/selection_algorithm.go
```diff
@@ -0,0 +1,104 @@
...
+	filtered := sa.filter(addrs, perfScores)
+	probabilities := sa.calculateProbabilities(filtered, stakes, prices)
+	return selectBy(probabilities)
+}
+
+func (sa ProbabilitySelectionAlgorithm) filter(addrs []ethcommon.Address, scores map[ethcommon.Address]float64) []ethcommon.Address {
```
thomshutt, line 29: Has the price filtering already happened by this stage? It might be nice to bring that in here too

thomshutt (approved):
Really really neat change!

server/selection_algorithm.go
```diff
@@ -0,0 +1,104 @@
...
+
+func selectBy(probabilities map[ethcommon.Address]float64) ethcommon.Address {
+	if len(probabilities) == 0 {
+		return ethcommon.Address{}
+	}
+
+	var addrs []ethcommon.Address
+	var cumProbs []float64
+	var cumProb float64
+	for addr, prob := range probabilities {
```
ecmulli, line 90: is `probabilities` sorted by value? it should be sorted in decreasing values.

leszko: Why does it need to be sorted? 🤔

ecmulli: actually, youre right. we'd still be sampling from a uniform distribution

server/selection_algorithm.go
leszko: Yeah, that's the plan. Or we can even decide in the future to have another implementation of `SelectionAlgorithm` which will use webook (aka. plugin approach).

server/selection_algorithm.go
leszko: No, filtering by price happens here: 
https://github.com/livepeer/go-livepeer/blob/f069a66779c238f64247f2cb2f666257c77fb4a4/discovery/db_discovery.go#L74

Hmm, actually it would be nice to move it into `filter()`. Let me try moving it.

leszko: Ok, I gave it a second thought and it's not so easy to move it... so I'll not move it in this PR.

The problem is that it currently works like this:
- Select all Os that have price < max price
- From these Os, send `GetInfo()` to each
- Wait `500ms` for the replies (if no O replied, then wait `1s`, then wait `2s` and so on)
- Then from these Os, use the Selection Algorithm

Now if move this price check to the Selection Algorithm, then what may happen is that if we select the max price low, then all Os that replied in `500ms` may have price too high. That would result in a situation in which we don't select any Os, just because some of them have good price, but didn't reply in `500ms`, but others reply in `500ms` but have price too high.

Technically, it's the same issue with filtering by performance score, but performance score is highly correlated with replying fast to `GetOrch()`, so I'm not afraid about this case.

thomshutt: Okay yeah, let's leave it for now then

PR #2876, opened 2023-09-29
Fix build for Windows

leszko:
Windows binaries with CUDA SDK are not working. It's probably fixable, but I think it would require at least 1 week of work. I guess it's not worth spending time on that, so for Windows AMD64 GPU, let's use the old `scale_cuda`. That means that we'll use `scale_npp` only in Linux.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -40,6 +40,10 @@ jobs:
           - platform:
               name: windows
             arch: arm64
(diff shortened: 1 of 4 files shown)


thomshutt:
Do we have a sense of the distribution of what Os are running on? My understanding was that there's a lot of Windows usage out there

thomshutt (approved):

PR #2877, opened 2023-10-02
Remove detailed scores logging

leszko:
It was a leftover. I think we should not print it, because it's too detailed.


thomshutt (approved):
LGTM, could you just throw a line in the PR description about why you're doing this please

PR #2881, opened 2023-10-05
Delete install.sh script

leszko:
This script does not work. We could fix and maintain it, but it's not worth doing it. Nobody complained for a long time that it's not working. Therefore, I'd get rid of it. 

Related docs PR: https://github.com/livepeer/docs/pull/444


thomshutt (approved):
Always in favour of deleting stuff!

PR #2884, opened 2023-10-09
Detect resolution change and reinit hw session

leszko:
It reinitializes the hw session if there is a mismatch in media info between segments.

Note that it does not fix the issue completely, because if the resolution is changed inside a segment, then the given segment is not transcoded correctly. But it makes the issue way less problematic, becuase all consecutive segments are ok.

fix https://linear.app/livepeer/issue/VID-438/analyze-pink-and-green-lines-in-transcoded-video

### DIFF
--- a/server/broadcast.go
+++ b/server/broadcast.go
@@ -840,6 +840,9 @@ func processSegment(ctx context.Context, cxn *rtmpConnection, seg *stream.HLSSeg
 	}
 
 	clog.V(common.DEBUG).Infof(ctx, "Processing segment dur=%v bytes=%v", seg.Duration, len(seg.Data))
+	if segPar != nil && segPar.ForceSessionReinit {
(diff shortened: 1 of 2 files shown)


thomshutt (approved):
LGTM, is there any easy way we can track how often the session reinits are happening?

leszko:
> LGTM, is there any easy way we can track how often the session reinits are happening?

Good point, added logging when HW Session Reinit is requested. PTAL: https://github.com/livepeer/go-livepeer/pull/2884/commits/81b6ffc3b1f106f157702ba750daeb9add16c3a2

thomshutt:
> > LGTM, is there any easy way we can track how often the session reinits are happening?
> 
> Good point, added logging when HW Session Reinit is requested. PTAL: [81b6ffc](https://github.com/livepeer/go-livepeer/commit/81b6ffc3b1f106f157702ba750daeb9add16c3a2)

Looks good! 🚢 🇮🇹

Issue #2387, opened 2022-05-09
Define "WebSocket" Orchestrator Capability


thomshutt:
Let's use `Capability_WSTransport` for the name.

Capabilities are defined in https://github.com/livepeer/go-livepeer/blob/249d9ebd80380dddd594e32bc9f44288d8b6fc33/core/capabilities.go

This property should initially always return "false".

Once @AlexKordic's work on websocket streaming is comeplete, we'll conditionally invoke it based on this property and cut a release with this capability always set to true.

PR #2885, opened 2023-10-09
Fix Orchestrator Swaps

leszko:
Related to two Linear Tickets:
fix https://linear.app/livepeer/issue/VID-430/investigate-why-stream-had-highly-variable-transcode-times
fix https://linear.app/livepeer/issue/VID-429/understand-orchestrator-swap

**Explanation for https://linear.app/livepeer/issue/VID-430/investigate-why-stream-had-highly-variable-transcode-times**
The segment is very short `1s`. And we calculate the in-memory latency score as `RTT transcoding time / segment duration`. If this value is greater than `1`, then we swap the Orchestrators. For such short segments, the RTT is almost always higher than 1. The quick fix is to have a minimal segment duration. I've set it to `1.5s`.

**Explanation for https://linear.app/livepeer/issue/VID-429/understand-orchestrator-swap**
It's hard to tell why the Os were swapped, but I believe that it's because there was a segment in flight for longer than 1.5s, which causes the Orchestrator Swap. This PR does not fix anything wrt to that, but it adds additional logs which will help to analyze further cases like that.

### DIFF
--- a/server/broadcast.go
+++ b/server/broadcast.go
@@ -218,7 +218,7 @@ func selectSession(ctx context.Context, sessions []*BroadcastSession, exclude []
 		if len(session.SegsInFlight) == 0 {
 			if session.LatencyScore > 0 && session.LatencyScore <= SELECTOR_LATENCY_SCORE_THRESHOLD {
 				clog.PublicInfof(ctx,
-					"Selecting new orchestrator, reason=%v",
+					"Reusing Orchestrator, reason=%v",
 					fmt.Sprintf(
 						"performance: no segments in flight, latency score of %v < %v",
 						session.LatencyScore,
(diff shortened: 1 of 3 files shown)


thomshutt:
@leszko we should already have the 1.5s floor from https://github.com/livepeer/go-livepeer/pull/2837, so it'd be good to understand why if that isn't working

mjh1 (approved):

leszko:
> @leszko we should already have the 1.5s floor from #2837, so it'd be good to understand why if that isn't working

This PR and #2837 are two separate things. I'll try to explain it here.
- #2387 is about "how long you wait when there is segment in flight". So, let's say you've sent segment 1 to O, then you want to send segment 2, but O didn't return the renditions for segment 1 yet. So, **there is 1 segment in flight**. Should you send segment 2 to the same O or should you swap to a new one; #2387 changes that we allow at least 1.5s before we swap to another O
- This PR is about in-memory latency score. So, let's say you've sent segment 1 to O, then O returned rendition for segment 1. So, there are **no segments in flight**. You want to send segment 2, should you send it to the same O or should you swap? In-memory latency score decides about it. And we calculate it as `RTT / seg dur`.

thomshutt:
@leszko gotcha, thanks for the explanation!

thomshutt (approved):

PR #2886, opened 2023-10-09
build: refactor install_ffmpeg.sh, add macos --> linux support

iameli:
This implements macOS --> Linux cross-compilation support for hacking on go-livepeer in livepeer-in-a-box.

Just M1/arm64 support for the moment. I assume the process for adding amd64 would be very similar, but I don't have one handy to test.

Specific changes:
* Refactored install_ffmpeg.sh a bit. It will now have four variables defined in all cases: `BUILDOS`, `BUILDARCH`, `GOOS`, and `GOARCH`. This makes it much easier to implement conditional checks later on.
* The Makefile has those variables now also.
* Added a new conditional block for `darwin-arm64 --> linux-arm64` compilation.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -14,43 +14,52 @@ concurrency:
 
 jobs:
   linux-build:
-    name: Build binaries for ${{ matrix.platform.name }}-${{ matrix.hardware.type }}-${{ matrix.arch }}
-    runs-on: ${{ matrix.platform.runner }}
+    name: Build binaries for ${{ matrix.target.GOOS }}-${{ matrix.target.type }}-${{ matrix.target.GOARCH }}
+    runs-on: ${{ matrix.target.runner }}
     container:
-      image: ${{ matrix.hardware.container }}
+      image: ${{ matrix.target.container }}
       env:
         DEBIAN_FRONTEND: noninteractive
     strategy:
       fail-fast: false
       matrix:
-        arch:
-          - arm64
-          - amd64
-        platform:
-          - name: linux
+        target:
+          - GOOS: linux
+            GOARCH: amd64
             runner: ubuntu-20.04
-          - name: windows
-            runner: ubuntu-20.04
-        hardware:
-          - type: cpu
             container: ubuntu:20.04
-          - type: gpu
+            type: cpu
+
+          - GOOS: linux
+            GOARCH: arm64
+            runner: ubuntu-20.04
+            container: ubuntu:20.04
+            type: cpu
+
+          - GOOS: linux
+            GOARCH: amd64
+            runner: ubuntu-20.04
             container: livepeerci/cuda:11.7.1-cudnn8-devel-ubuntu20.04
-        exclude:
-          - platform:
-              name: windows
-            arch: arm64
-          - platform:
-              name: windows
-            hardware:
-              type: gpu
+            type: gpu
+
+          - GOOS: linux
+            GOARCH: arm64
+            runner: ubuntu-20.04
+            container: livepeerci/cuda:11.7.1-cudnn8-devel-ubuntu20.04
+            type: gpu
+
+          - GOOS: windows
+            GOARCH: amd64
+            runner: ubuntu-20.04
+            container: ubuntu:22.04
+            type: cpu
 
     steps:
       - name: Setup ubuntu container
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -22,30 +22,66 @@ cgo_cflags :=
...
-ifeq ($(uname_s),Linux)
-	ifeq ($(GOARCH),arm64)
-		cgo_cflags += --target=aarch64-linux-gnu
-		cgo_ldflags += --target=aarch64-linux-gnu
-		cc = clang --sysroot=/usr/aarch64-linux-gnu
+	ifeq ($(GOOS),linux)
+		ifeq ($(GOARCH),arm64)
+			LLVM_PATH ?= /opt/homebrew/opt/llvm/bin
+			SYSROOT ?= /tmp/sysroot-aarch64-linux-gnu
+			cgo_cflags += --target=aarch64-linux-gnu -Wno-error=unused-command-line-argument
```
iameli: I swear I tried 100 different methods and nothing I tried was able to get `-fuse-ld=$(LLVM_PATH)/ld.lld` passed to `clang` when it was running in linker mode consistently. So now we pass it to _every_ `clang` command, and `-Wno-error=unused-command-line-argument` just ignores it when we're compiling rather than linking. I don't know. There's probably a better way, but no combination of `-ldflags`, `-extld`, `-extldflags`, `LDFLAGS`, and a few others was making it happen for me.

.github/workflows/build.yaml
```diff
@@ -14,43 +14,52 @@ concurrency:
...
+          - GOOS: linux
+            GOARCH: amd64
+            runner: ubuntu-latest-m
             container: ubuntu:20.04
-          - type: gpu
+            type: cpu
+
+          - GOOS: linux
+            GOARCH: arm64
+            runner: ubuntu-latest-m
```
leszko: Do you need `runner` to be part of `matrix` if it's always the same `ubuntu-latest-m`?

.github/workflows/build.yaml
```diff
@@ -69,44 +78,44 @@ jobs:
...
 
-      - name: Cache ffmpeg
-        id: cache-ffmpeg
-        uses: actions/cache@v3
-        with:
-          path: /home/runner/compiled
-          key: ${{ runner.os }}-${{ matrix.platform.name }}-${{ matrix.hardware.type }}-${{ matrix.arch }}-ffmpeg-${{ hashFiles('**/install_ffmpeg.sh') }}
-          restore-keys: |
-            ${{ runner.os }}-${{ matrix.platform.name }}-${{ matrix.hardware.type }}-${{ matrix.arch }}-ffmpeg-
+      # - name: Cache ffmpeg
```
leszko: Could you clean-up the commented-out code?

install_ffmpeg.sh
```diff
@@ -187,7 +227,7 @@ if [[ ! -e "$ROOT/ffmpeg/libavcodec/libavcodec.a" ]]; then
     --extra-ldflags="${EXTRA_FFMPEG_LDFLAGS} -L${ROOT}/compiled/lib -L/usr/local/cuda/lib64" \
     --prefix="$ROOT/compiled" \
     $EXTRA_FFMPEG_FLAGS \
-    $DEV_FFMPEG_FLAGS
+    $DEV_FFMPEG_FLAGS || (tail -100 ${ROOT}/ffmpeg/ffbuild/config.log && exit 1)
```
leszko, line 230: What is this `tail` here? Looks weird, could you at least explain it with a comment?

emranemran: I think the intention here is that if `configure` fails, then print the last 100 log lines for debugging and exit.

iameli: Yeah that!
```suggestion
    $DEV_FFMPEG_FLAGS || (tail -100 ${ROOT}/ffmpeg/ffbuild/config.log && exit 1)
    # If configure fails, then print the last 100 log lines for debugging and exit.
```

.github/workflows/build.yaml
iameli: I'd prefer to leave it for now - in the process of doing this I needed to adjust the runners per-job quite a bit; it's entirely possible the next build we do will have to do the same thing

leszko (approved):

PR #2892, opened 2023-10-17
Fix price per session

### LINKED ISSUE #2891: Selection algo vulnerability + graceful shutdown feature
**Describe the bug**
While trying to find ways to gracefully shutdown my Os I may have found a price weighted selection algorithm vulnerability.

**Describe the bug**
First, I was experimenting with increasing my Price Per Pixel on an active O (ie increase to 100,000 wei) in hopes that it would keep all the current streams until they gracefully finished and all new streams would be rejected as my price was way above any Bs max price.

Unfortunately, changing the Price Per Pixel on the fly affects all current stream which led to the streams crashing and my O going to 0. Not very graceful :(

But that had me thinking of the new selection algo. 

Couldn't I just advertise a 0 Price Per Pixel on discovery and then once the stream started, increase it to 900 and keep all the streams? This would definitely give me the best selection opportunity based on price weight selection and give me the most profit by being able to adjust prices on the fly.

**Two for one fix?**
My proposal is to make the B require a static Price Per Pixel for the entirety of the stream. This would fix the possible manipulation of prices of an active stream and make the price selection weight more honest.

Also... We have a way to gracefully shutdown our nodes! :D

leszko:
fix https://github.com/livepeer/go-livepeer/issues/2891

### DIFF
--- a/core/accounting.go
+++ b/core/accounting.go
@@ -124,6 +124,7 @@ type Balances struct {
 type balance struct {
 	lastUpdate time.Time // Unix time since last update
 	amount     *big.Rat  // Balance represented as a big.Rat
+	fixedPrice *big.Rat  // Fixed price for the session
 }
 
 // NewBalances creates a Balances instance with the given ttl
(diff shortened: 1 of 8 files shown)


thomshutt (approved):
Looks good but this will mean every time an O increases their prices, they lose sessions right? I think this is good as a patch for the vulnerability, but does somewhat limit the ability of the Os to do dynamic pricing changes without losing out on work

0xspeedybird:
This approach is not comprehensive and prevents the market dynamics promoted during the recent selection algo updates.  Please review the comments on #2891.

Titan-Node:
I would rather see a comprehensive fix than just a patch to fix the "vulnerability" today. Technically it's not actually costing the B any more funds than they would have prior to the algo change, Os can still only increase it to less than the maxPrice the B sets. 

If we implement this as it stands all Os started by default that do not include the autoPriceAdjust flag will fail to work.

eliteprox:
I agree, we should keep the same price during the entire stream. Dropping session when the price changes isn't an acceptable solution imho.

Titan-Node:
The Orch would need to cache the price per orchSessionID and maintain that price until that session ends

leszko:
Thanks for all the comments. I think you're right. Let's wait with merging this PR until I send a fix for O to keep the price the same during the whole session.

leszko:
Updated the PR. Now:
- Orchestrator fixes the price per session
- Broadcaster checks if the price was not increased

@Titan-Node @eliteprox @0xspeedybird @thomshutt PTAL.

thomshutt (approved):
LGTM!

PR #2897, opened 2023-10-23
Add '-maxTotalEV' flag

leszko:
This allows B to set separately:
- Max single ticket EV
- Max total EV for one payment


papabear99:
This was brought up in the watercooler chat.  What's the purpose of this pr?

leszko:
> This was brought up in the watercooler chat. What's the purpose of this pr?

The context is that the current `-maxTicketEV` flag set in B has 2 purposes:
- Sets the maximum `-ticketEV` accepted from O
- Sets the maximum total ticket EV sent with one segment

So, we split it into 2 flags. The broader context is that the current ticket EV is usually set to a too-high value, which means overpaying by B for short videos.

thomshutt (approved):

PR #2901, opened 2023-10-30
Decrease the default O's Ticket EV to 1000000000.

leszko:
The rationale behind is that the ticket EV should be smaller than the expected payment for the transcoding of 1 segment. Otherwise, B needs to always pay in advance. The new default value is 1000000000 wei, which should cover transcoding of 1 s segment, 30 FPS, 1280x720 with a price 50ppp.

### DIFF
--- a/cmd/livepeer/starter/starter.go
+++ b/cmd/livepeer/starter/starter.go
@@ -193,7 +193,7 @@ func DefaultLivepeerConfig() LivepeerConfig {
 	defaultMaxGasPrice := 0
 	defaultEthController := ""
 	defaultInitializeRound := false
-	defaultTicketEV := "1000000000000"
+	defaultTicketEV := "1000000000"
 	defaultMaxFaceValue := "0"
 	defaultMaxTicketEV := "3000000000000"
 	defaultMaxTotalEV := "20000000000000"
--- a/cmd/livepeer/starter/starter.go
+++ b/cmd/livepeer/starter/starter.go
@@ -193,7 +193,7 @@ func DefaultLivepeerConfig() LivepeerConfig {
 	defaultMaxGasPrice := 0
 	defaultEthController := ""
 	defaultInitializeRound := false
-	defaultTicketEV := "1000000000"
+	defaultTicketEV := "8000000000"
 	defaultMaxFaceValue := "0"
 	defaultMaxTicketEV := "3000000000000"
 	defaultMaxTotalEV := "20000000000000"


thomshutt (approved):

FranckUltima:
Attention to the processor overload when processing a larger number of tickets. Currently, with the default ticketEV, there is a 1 in 12,000 chance of winning a ticket. If you divide by 1000, then a ticket will have a 1 in 12,000,000 chance of winning. This will generate a significant number of tickets, and it seems to me that a few years ago, tests with lower ticketEV led to saturation and overload of processors and crashes of some nodes (depending on their power). To be verified.

leszko:
> Attention to the processor overload when processing a larger number of tickets. Currently, with the default ticketEV, there is a 1 in 12,000 chance of winning a ticket. If you divide by 1000, then a ticket will have a 1 in 12,000,000 chance of winning. This will generate a significant number of tickets, and it seems to me that a few years ago, tests with lower ticketEV led to saturation and overload of processors and crashes of some nodes (depending on their power). To be verified.

Thanks @FranckUltima 

Did it cause crashes in O or in B? Do you maybe remember?

FranckUltima:
> > Attention to the processor overload when processing a larger number of tickets. Currently, with the default ticketEV, there is a 1 in 12,000 chance of winning a ticket. If you divide by 1000, then a ticket will have a 1 in 12,000,000 chance of winning. This will generate a significant number of tickets, and it seems to me that a few years ago, tests with lower ticketEV led to saturation and overload of processors and crashes of some nodes (depending on their power). To be verified.
> 
> Thanks @FranckUltima
> 
> Did it cause crashes in O or in B? Do you maybe remember?
I remember that the crash occurred on the node of an orchestrator who had lowered the ticketEV to see what effect it would have. I cannot say what value he had entered or what hardware he was using. I am not sure, but maybe it was Papa Bear. You should ask him if he remembers

leszko:
@papabear99 do you maybe remember? 🙃

papabear99:
> @papabear99 do you maybe remember? 🙃

Oh it was a long time ago.  I don't remember what value I used but I think I got a million tickets on a single node in a day.  It didn't cause the node to crash but it caused my test scores to drop a fair amount.  I'll see if I have any notes from then...probably not but I'll look.

papabear99:
Found it on discord.  https://discord.com/channels/423160867534929930/914983525424443452/964316147317084160

-ticketEV=1000000000 

Node went from an avg of 3500 tickets/day to over 3 million

See the chart in the post to view test stream scores.

leszko:
I did some quick testing on a single GPU machine.

| ticket EV | price | tickets / segment | result |
| -- | -- | -- | -- |
| 10^12 (current default) | 1 | 1 / 1000 | everything good |
| 10^9 (planned new default) | 1000 | 20 | small spike in a GPU usage (5% / session), nothing to crazy, the transcoding time is still the same | 
| 10^7 | 1000 | 2000 | error `invalid ticket senderNonce: too many values` |

So, from my measurements, it looks like we should not experience a transcoding slowdown or any big CPU spikes. But let me ask in Discord other Orchestrators.

papabear99:
> I did some quick testing on a single GPU machine.
> 
> ticket EV	price	tickets / segment	result
> 10^12 (current default)	1	1 / 1000	everything good
> 10^9 (planned new default)	1000	20	small spike in a GPU usage (5% / session), nothing to crazy, the transcoding time is still the same
> 10^7	1000	2000	error `invalid ticket senderNonce: too many values`
> So, from my measurements, it looks like we should not experience a transcoding slowdown or any big CPU spikes. But let me ask in Discord other Orchestrators.

My results showed an 18% increase in roundtrip time for Livepeer test streams.  

Do you mean CPU usage went up 5% or GPU as written?

Was the test done online (mainnet) or was it an isolated environment (offline). i.e. Was it an online B sending tickets to an O or was it a simulated test?

What hardware did you test with, CPU model, GPU model, RAM etc?

Maybe an online office hours/working group chat for Os to meet would be a more effective way to discuss our findings?

mikezupper:
What is the impact of the Broadcaster sending 1000x more tickets?

This topic should be discussed in a working group meeting in the livepeer discord. We need to ensure more Orchs can weight in on this topic.

Tibia2000:
If B is overpaying 1000x for the first segment is it not getting countered by the streams which has like 1999 segments where first 1000 segmetns are covered with ticket that comes with 1 segment and next 999 segments are not paid for? That should balance out if average streams are long enough, but I see it can be a problem for a short  streams which idk how much  of total% are.

leszko:
@papabear99 Unfortunately, I was testing it only locally (on some cloud GPU machine), because I don't have an O running on Mainnet. That is why I need your (and other Os' input here).

@mikezupper TBH I didn't measure the impact on B, that's a good question. I'll check it out.

leszko:
I makes me think that we should maybe have a flag like `minTicketEV` in O, because setting `ticketEV` in O to a very low number could kill B 🤔

leszko:
Ok, after all the discussions here and in [Discord](https://discord.com/channels/423160867534929930/1169568988783726673/1169569456771571764), I'm bumping the default ticket EV to 8 gwei.

PR #2907, opened 2023-11-13
update ticket received log line to be summary of tickets

ad-astra-video:
Moves `Receiving ticket` log line to be summary log line of tickets processed.  Was a debug log line that printed for each ticket processed.

- one file updated

@papabear99 ran similar version on mainnet 

No

**Checklist:**

### DIFF
--- a/core/orchestrator.go
+++ b/core/orchestrator.go
@@ -162,6 +162,8 @@ func (orch *orchestrator) ProcessPayment(ctx context.Context, payment net.Paymen
 	totalEV := big.NewRat(0, 1)
 	totalTickets := 0
 	totalWinningTickets := 0
+	totalFaceValue := big.NewInt(0)
+	totalWinProb := big.NewRat(0, 1)
 
 	var receiveErr error
 
@@ -173,8 +175,6 @@ func (orch *orchestrator) ProcessPayment(ctx context.Context, payment net.Paymen
 			sender,
 			tsp.SenderNonce,
 		)
-
-		clog.V(common.DEBUG).Infof(ctx, "Receiving ticket sessionID=%v faceValue=%v winProb=%v ev=%v", manifestID, eth.FormatUnits(ticket.FaceValue, "ETH"), ticket.WinProbRat().FloatString(10), ticket.EV().FloatString(2))
 
 		_, won, err := orch.node.Recipient.ReceiveTicket(
 			ticket,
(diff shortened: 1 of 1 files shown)


thomshutt (approved):

core/orchestrator.go
```diff
@@ -198,6 +198,8 @@ func (orch *orchestrator) ProcessPayment(ctx context.Context, payment net.Paymen
 			ev := ticket.EV()
 			orch.node.Balances.Credit(sender, manifestID, ev)
 			totalEV.Add(totalEV, ev)
+			totalFaceValue.Add(totalFaceValue, ticket.FaceValue)
+			totalWinProb.Add(totalWinProb, ticket.WinProbRat())
```
leszko, line 202: 1. So, we never set those to `0` again and the log will always show the cumulative tickets?
2. I wonder how much sense it takes to add the probabilities and face values together? What this information actually may give you?

leszko (commented):
@ad-astra-video Added a comment. Thanks for the PR.

core/orchestrator.go
ad-astra-video: 1. These variables are initialized in the ProcessPayment function so start at 0 every time. The log line is intended to show the cumulative amounts for the tickets (face value/win prob/ev).   
2. To me, this is a debug log line so more info is likely better right?

leszko: Ahh, you're right. I missed that part that they are initialized for each payment processing. Thansk for the explanation!

leszko (approved):

stronk-dev:
Shouldn't `winProb` be the average win prob rather than the sum?

ad-astra-video:
> Shouldn't `winProb` be the average win prob rather than the sum?

I don't think the average gets you to the `faceValue` mathematically.  Also, wouldn't the average be the `winProb` of one ticket in the batch? The way I looked at it was ten tickets at 10 gwei gets 100 gwei in tickets with a cumulative `winProb` of .000008333.  100 gwei / .000008333 / 1,000,000 = .012 ETH.

stronk-dev:
What does the sum of `winProb` or `faceValue` represent? These are per ticket params
If you earn 10 tickets of 0.12FV and 10% win percentage, that's not equivalent to 1 ticket of 1.2FV and 100% win percentage. Only the total EV makes sense to sum up for a summary, ie:
```
Payment tickets processed 10 tickets with average faceValue=1.2, winProb=10%, ev=0.12 for a total EV of 1.2
```

ad-astra-video:
OK, see now. I was approaching from the reverse starting with EV / total win prob.  I have no issue removing the winProb and faceValue totals. Was just trying to reduce the spam in logs from ticket processing on debug log level. Will get PR in at some point.

stronk-dev:
i mean it's good to reduce logging, i just brought it up cause the change breaks this panel of mine (since there's no way to translate the sums back to a per-ticket basis with the new log format):
image

PR #2915, opened 2023-12-08
Fix ValueRedeemed stat

stronk-dev:
There is a unused prometheus metric to track the value of winning tickets redeemed. This PR modifies the naming to be inline with other prometheus metrics and makes sure it gets updated

not yet, any takers?


thomshutt:
@stronk-dev any thoughts on making this backwards compatible if people have built tooling off the back of the old naming? It seems like the previous method _was_ being called in the code

stronk-dev:
Ah you're right, it _does_ get called, so then this whole PR is moot. I thought it was broken (due to an error in my query)

PR #2888, opened 2023-10-10
Initialize default metrics to 0

papabear99:
**What does this pull request do? Explain your changes. (required)**
Initialize Winning Tickets and Current Sessions to 0

**Specific updates (required)**

**How did you test each of these updates (required)**
Built code locally and deployed to production orchestrator

**Does this pull request close any open issues?**
No

**Checklist:**

- [X ] Read the [contribution guide](./doc/contributing.md)
- [X ] `make` runs successfully
- [X ] All tests in `./test.sh` pass
- [NA] README and other documentation updated
- [ NA] [Pending changelog](./CHANGELOG_PENDING.md) updated

PR #2916, opened 2023-12-08
Init winning ticket value redeemed stat on boot

stronk-dev:

So before #2888 got merged, the winning ticket metric wasn't working properly. At least, when graphing the metric using `livepeer_winning_tickets_recv` you could see that it was counting them tickets, but when querying the increase over time (ie using `sum by (datacenter)(increase(livepeer_winning_tickets_recv[7d]))`) it would be short one or more tickets.

On every restart of the node the prometheus counters and gauges get reset. So using `increase` is the way to correctly display an increasing counter of tickets across restarts of the node

I am currently seeing the same behaviour with `livepeer_value_redeemed`
image
where on the first winning ticket, it counts it as an increase of 0, but then the next ticket gets counted correctly

Here is my theory: when the node boots, not all parameters have their values initialised to 0. It seemd that `increase` does not count an increase from null -> 1, but it does count an increase from 0 -> 1

This PR initialised the `value_redeemed` stat to 0 on boot, which should fix the issue i am seeing

Not yet tested, deploying now...


rickstaa:
This will fix #2917.

FranckUltima:
Thank you! So happy to see this point finally resolved! I don’t know how long I spent on Grafana trying to find the query that would give me the exact number of winning tickets, even after a server restart, without ever succeeding.

stronk-dev:
> Thank you! So happy to see this point finally resolved! I don’t know how long I spent on Grafana trying to find the query that would give me the exact number of winning tickets, even after a server restart, without ever succeeding.

Lol I know how you feel, when @papabear99 PR was merged it fixed the `livepeer_winning_tickets_recv` as an unintended side effect, so just applying the same logic to `livepeer_value_redeemed` now. we can finally accurately track work performed vs payouts redeemed

thomshutt (approved):
Looks good, thanks @stronk-dev!

PR #2928, opened 2023-12-14
Orchestrator: add flag to skip the availability check on startup

stronk-dev:
When booting an Orchestrator, an availability check is performed by sending a Ping GRPC request to the configured service address
On certain setups local loopback might not be available, causing the test to fail even though the Orchestrator is available from remote connections

This change adds a `testOrchAvail` flag to explicitly disable the startup check

Going to use the binaries built by this PR to test this on my Singapore node, which exhibits this issue

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -15,6 +15,7 @@
 #### Orchestrator
 
 - #2911 Set default price with livepeer_cli option 20 (@eliteprox)
+- #2928 Added `startupAvailabilityCheck` param to skip the availability check on startup (@stronk-dev)
 
 #### Transcoder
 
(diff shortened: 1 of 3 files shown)


stronk-dev:
Currently testing on Singapore. Flag seems to have worked, let's see if work still comes in...

thomshutt:
Awesome, thanks @stronk-dev! Let me know once you're happy with the testing and we can merge

stronk-dev:
Seems fine so far, any issues with the naming or comments? Wasn't sure what to group the new flag as

thomshutt:
Naming looks good, the only bit I'd maybe change would be the flag itself to be the full version - `startupAvailabilityCheck` or `testOrchAvailability`? Internal variables etc. can stay as they are though

leszko (approved):
Thanks for the PR @stronk-dev Looks good to me. 

Agree with @thomshutt that we could rename the flag. Other than that LGTM. @thomshutt Feel free to merge this PR.

thomshutt:
LGTM @stronk-dev, could you update the pending changelog please and then I'll merge

PR #2946, opened 2024-01-09
Make non-random selection curve dependent on both stake and price

stronk-dev:

Currently, stake-weighted selection is independent from price-weighted selection. This means that you can never have a situation where:
- All Orchestrators, regardless of their stake, are incentivized to compete on price
- Stake has a meaningful impact for Orchestrators at a similar price

It allows Orchestrators to 'optimise' their utilisation by running a single high-staked orch (to maximise gains from the stake-weighted curve) and run one or more side-orchs to compete on pricing without cannibalising streams from their main orchestrator.

It also reduces network security, as any node who just barely makes the top 100 can capture the majority of streams being distributed on the pricing curve

This PR proposes two changes to the selection algorithm:

Firstly it removes dividing by the sum, as the `priceProb` was already normalised and the `stakeProb` can be normalised by simply using the highest stake found in the set.
This way the price and stake probalities are neatly normalised between [1.0, 0.0] but the code is slightly leaner

Most importantly the percentages for price-weighted selection and stake-weighted selection are compounded, preventing:
- An Orchestrator with high stake running way above market pricing
- An Orchestrator with low stake able to capture majority chunks of traffic (especially in less competitive regions)

TBA, will test changes via a dashboard which simulates the selection algorithm to compare the effects

**Checklist:**

### DIFF
--- a/server/selection_algorithm.go
+++ b/server/selection_algorithm.go
@@ -49,31 +49,23 @@ func (sa ProbabilitySelectionAlgorithm) filter(addrs []ethcommon.Address, scores
 }
 
 func (sa ProbabilitySelectionAlgorithm) calculateProbabilities(addrs []ethcommon.Address, stakes map[ethcommon.Address]int64, prices map[ethcommon.Address]float64) map[ethcommon.Address]float64 {
-	pricesNorm := map[ethcommon.Address]float64{}
+	// Find the biggest stake in the current working set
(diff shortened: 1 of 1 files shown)


stronk-dev:
(Will reopen with a better solution, as obviously the weights do not work in this solution and simply modifying the two percentages without weights is not adjustable enough)

PR #2947, opened 2024-01-09
Modifying price-weighted curve

stronk-dev:

With the current pricing curve, orchestrators running at 'unsustainable' prices can capture a high amount of streams. Due to the exponential curve, an orchestrator at 2 PPP simply has a much higher chance than someone at 200 PPP.

The current selection algorithm has an `expFactor` which could make this less OP, but it would also be much more likely to choose orchs at a price which is too high

This PR changes the selection curve to an inverted sigmoid. This allows you to:
- Center the curve around a point where diminishing returns should kick in to make nodes at extremely low values less OP
- Due to the S curve it still prevents Orchestrators to sit at the high end of the spectrum
- Still allows for adjusting the slope, which is a bit more forgiving for orchestrators trying to dial in their prices

[This link](https://www.desmos.com/calculator/kdo1skpzqt) allows you to play around with the params of the proposed curve and the current curve to compare them

TBA, will test first in a dashboard which simulates the selection algorithm

**Checklist:**


stronk-dev:
Note on this draft PR is that it's just to open discussion and allow people to check & test the changes.

The pricing curve is only an issue current due to the large difference in the lowest PPP and market pricing. If market pricing drops to sub-100 levels, then the price curve likely doesn't even need adjustment

PR #2899, opened 2023-10-24
ci: add protoc on build

iameli-streams:
So... this shouldn't be necessary. `protoc` is needed to build the files in `net` like `net/lp_rpc.pb.go`. Those files are committed to the repo; `make` shouldn't be rebuilding them. And yet, every so often, we get an error on build that [looks like this](https://github.com/livepeer/go-livepeer/actions/runs/6630250258/job/18011282601?pr=2886):

```
protoc -I=. --go_out=plugins=grpc:. net/lp_rpc.proto
/bin/bash: line 1: protoc: command not found
make: *** [Makefile:7: net/lp_rpc.pb.go] Error 127
Error: Process completed with exit code 2.
```

I don't know why that happens. It doesn't happen once you rebuild. But... regenerating those files is harmless, I guess, so this will stop that for everybody forever. IDK.

PR #2954, opened 2024-01-21
feat: protoc-gen-go >=1.20.0 breaking change support

### LINKED ISSUE #2953: Support protoc v1.20.0 and above
## Description

Go-livepeer is currently not compatible with production releases of `protoc-gen-go`.  As a result the generated `.go` files in the `net` package are outdated and should not be used. 

Instead go-livepeer should be upgraded to be compatible with `protoc-gen-go` version 1.20 and above which introduces breaking changes and a full re-architecture. 

Furthermore, users building the repository from source with `make all` wanting to generate the protocol buffers will run into issues. The code generated by the go gRPC bindings is no longer compatible with the code of the `go-livepeer` packages that implement it. 

`go-livepeer` currently uses `protoc-gen-go@v1.3.5`, which has no support. Releases prior to v1.20 are not accessible anymore in the official repository.

### Breaking changes: 

- generating gRPC service code was removed from `protoc-gen-go `and moved to a new plugin, `protoc-gen-go-grpc.`
- XXX_ fields, used internally for protocol buffer functionality extensions from `proto2`, were made unexported. (go-livepeer uses proto3 regardless).
-  overall structure and content of the generated Go code were altered

## Alternative Solution

Vendor (outdated) Go gRPC/Protobuf libraries.

## More Context.

Refer to this discussion in the [Livepeer Discord](https://discord.com/channels/423160867534929930/1051963444598943784/1198632549690388641)

kyriediculous:
Upgrade go-livepeer to support gRPC go bindings generated by `protoc-gen-go >= 1.20.0` and `protoc-gen-go-grpc >= 1.0.0`

This is necessary since the breaking changes and major release of `protoc-gen-go@1.20.0`

- Remove usage of `XXX_...` fields which were lingering from proto2 support
- Add UnimplementedServer types to our gRPC services
- Update install for tests using new protoc-gen-go

ran tests, build new source.

Fixes #2953 

**Checklist:**


kyriediculous:
Have some trouble running tests locally on M2 mac 

```
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

kyriediculous:
Been running this on a linux box, the failing test is regarding the removed `XXX_...` fields and an expected difference in mocked calls.

I'll try to fix it tomorrow. We'll need a custom matcher for `gomock` that ignores the unexported fields from protobufs. It seems `atomicMessage` is a new field which is being autofilled while we set it to `nil`

I'm wondering would this break wire protocol compatibility ? It shouldn't right , the messages are the same I believe those fields were just used for serialization/deserialization.

leszko:
> Been running this on a linux box, the failing test is regarding the removed `XXX_...` fields and an expected difference in mocked calls.
> 
> I'll try to fix it tomorrow. We'll need a custom matcher for `gomock` that ignores the unexported fields from protobufs. It seems `atomicMessage` is a new field which is being autofilled while we set it to `nil`
> 
> I'm wondering would this break wire protocol compatibility ? It shouldn't right , the messages are the same I believe those fields were just used for serialization/deserialization.

I believe that if you don't change any fields in `*.proto` files, then it's backward compatible. 

@kyriediculous Let me know when the PR is ready for review, I'll take a look into it.

kyriediculous:
Having some trouble with some of the tests (further expedited by the fact that I'm having some issue running the tests locally so I always have to push up to a Linux VM haha).

kyriediculous:
Ready for review @leszko ! 

As I understand the failing CI tasks are due to https://github.com/livepeer/go-livepeer/pull/2899

But I don't have permissions to re-run

leszko (approved):
Thanks @kyriediculous

PR #2972, opened 2024-03-06
eth,eth/watcher: Create Chainlink price feed watcher

victorges:
This creates a new client in the eth package to fetch and watch a [Chainlink PriceFeed](https://docs.chain.link/data-feeds/price-feeds).

This will be used by go-livepeer to be able to peg the transcoding price in an alternate currency,
mostly the American dollar, for a more predictable cost and revenue for node operators.

The logic for using this pricefeed info into tuning the price dynamically will be done in a separate PR.

- Create a pricefeed client in `eth` pkg, for basic fetching of price from a price feed
- Create a pricefeed watcher in `eth/watchers` pkg, to keep up to date on pricing changes
- Add a couple tests to those

- `go test`
- Also ran it manually on top of the ETH/USD pricefeed

Implements ENG-1725

**Checklist:**


leszko:
@victorges would it maybe be possible to use Chainlink library instead of creating go bindings on our own?

https://pkg.go.dev/github.com/smartcontractkit/chainlink@v1.13.3/core/gethwrappers/generated/aggregator_v3_interface

eth/pricefeed.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+func NewPriceFeedEthClient(ctx context.Context, rpcUrl, priceFeedAddr string) (PriceFeedEthClient, error) {
+	client, err := ethclient.DialContext(ctx, rpcUrl)
+	if err != nil {
+		return nil, fmt.Errorf("failed to initialize client: %w", err)
+	}
+
+	ok := isContractAddress(priceFeedAddr, client)
+	if !ok {
```
leszko:
```suggestion
	if !isContractAddress(priceFeedAddr, client) {
```
Super nit

eth/pricefeed.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+}
+
+type ethClient interface {
+	CodeAt(ctx context.Context, contract common.Address, blockNumber *big.Int) ([]byte, error)
+}
+
+// isContractAddress checks if the given address is an address of a contract
+// deployed on the corresponding blockchain.
+func isContractAddress(addr string, client ethClient) bool {
```
leszko: Do we need to explicitely check it? Won't it just fail later if the given address is not a smart contract? 🙃

eth/pricefeed.go
```diff
@@ -0,0 +1,117 @@
...
+
+	ok := isContractAddress(priceFeedAddr, client)
+	if !ok {
+		return nil, fmt.Errorf("not a contract address: %s", priceFeedAddr)
+	}
+
+	addr := common.HexToAddress(priceFeedAddr)
+	priceFeed, err := chainlink.NewAggregatorV3Interface(addr, client)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create mock aggregator proxy: %w", err)
```
leszko: Why "mock"? Isn't it just an aggregator proxy?

eth/watchers/pricefeedwatcher.go
```diff
@@ -0,0 +1,176 @@
...
+	if len(currencies) != 2 {
+		return "", "", fmt.Errorf("aggregator description must be in the format 'FROM / TO' but got: %s", description)
+	}
+
+	currencyBase = strings.TrimSpace(currencies[0])
+	currencyQuote = strings.TrimSpace(currencies[1])
+	return
+}
+
+// newTruncatedTicker creates a ticker that ticks at the next time that is a
```
leszko, line 155: 1. Could you explain why the normal ticker couldn't be used for pricefeedwatcher? It's hard to understand why the standard ticker (or timer) would not work.

2. I'm also not sure we're not complicating things too much with this. I'm starting to understand what you want to achieve - that everyone B and O will update the price at the exactly same time. But I think it may be hard to achieve in practice because 1) the clocks are not in sync 2) There are network delays 3) Someone may start the in the middle of an hour. 

Maybe it'd be enough to update the price at the beginning of each round?

eth/watchers/pricefeedwatcher.go
```diff
@@ -0,0 +1,176 @@
...
+// multiple of d, starting from the current time.
+func newTruncatedTicker(ctx context.Context, d time.Duration) <-chan time.Time {
+	return newTruncatedTickerMockable(ctx, d, time.Now, time.After)
+}
+
+// Test helper so we can mock the time functions in tests.
```
leszko: I'd try to avoid making the logic (so much more) complex just because of unit testing.

eth/watchers/pricefeedwatcher.go
```diff
@@ -0,0 +1,176 @@
...
+}
+
+// Current returns the latest fetched price data.
+func (w *PriceFeedWatcher) Current() eth.PriceData {
+	return w.current
+}
+
+// PriceUpdated returns a channel that will receive the updated price data as
+// soon as it changes.
+func (w *PriceFeedWatcher) PriceUpdated() <-chan eth.PriceData {
```
leszko: I wonder if we shouldn't keep the same convention as in other watchers and do the following:
1. Start watching in `starter.go`, not in `New...`
2. Expose `Subscribe()` function instead of exposing the channel (which can be consumer by only one consumer)

eth/watchers/pricefeedwatcher.go
```diff
@@ -0,0 +1,176 @@
...
+	newPrice, err := w.priceFeed.FetchPriceData()
+	if err != nil {
+		return fmt.Errorf("failed to fetch price data: %w", err)
+	}
+
+	if newPrice.UpdatedAt.After(w.current.UpdatedAt) {
+		w.current = newPrice
+		select {
+		case w.priceUpdated <- newPrice:
+		default:
```
leszko: This is kind-of tricky, because it means that if nobody is consuming the price then we stop sending it to the channel. So, if nobody is consuming and suddenly starts consuming, then it receives the old price.

I think to solve it we should do one of the following:
1. Send to the channel just the information that the price was updated (but without any value)
2. Use the `Subscribe()` mechanism we use in other watchers (as I commented in some other comment)

eth/watchers/pricefeedwatcher.go
```diff
@@ -0,0 +1,176 @@
...
+				err := w.updatePrice()
+				if err == nil {
+					break
+				} else if attempt >= priceUpdateMaxRetries {
+					clog.Errorf(ctx, "Failed to fetch updated price from PriceFeed attempts=%d err=%q", attempt, err)
+					break
+				}
+
+				clog.Warningf(ctx, "Failed to fetch updated price from PriceFeed, retrying after retryDelay=%d attempt=%d err=%q", retryDelay, attempt, err)
+				time.Sleep(retryDelay)
```
leszko: I don't love the sleep here, because it effectively blocks the whole for loop. E.g. if someone closes the context, it will stop only after this sleep. How about changing it to another ticker/timer?

Anyway, if you decide to keep this sleep, I'm ok with this. Shouldn't be too harmful :)

eth/watchers/pricefeedwatcher_test.go
```diff
@@ -0,0 +1,319 @@
...
+	require.Equal(t, priceData, w.current)
+	select {
+	case updatedPrice := <-w.priceUpdated:
+		require.Equal(t, priceData, updatedPrice)
+	case <-time.After(2 * time.Second):
+		t.Error("Updated price hasn't been received on channel")
+	}
+}
+
+func TestPriceFeedWatcher_Watch(t *testing.T) {
```
leszko, line 61: I wonder if we should do so many tests which are time-based + fake tickers. The drawback of them is that the tests can be flaky. Plus, with this fake ticker, I guess you had to add some complexity to the prod logic to make it testable....

leszko (commented):
Thanks for the PR @victorges . Added some comments. Other than that it looks good.

eth/pricefeed.go
victorges: Probably not. Honestly I just copied it from that price feed golang example that they have, but I can remove it to reduce entropy here.

eth/pricefeed.go
victorges: likely a typo from when i was trying to write tests. fixed

eth/watchers/pricefeedwatcher.go
victorges: 1. It is in order to have _some_ synchronization across nodes, so it depends only the clock difference and network delays, but not the time when the node started. I believe this would render a variance of a couple seconds most of the time, not up to an hour difference (if we use the same refresh period). Even though we can never solve it perfectly, trying to reduce this desynchronization is worth it IMO.
2. As mentioned above, clock differences and network delays render a much lower desynchronization than if we did nothing at all. So my approach here was to propose something not too complex (the tests themselves were the only complex part IMO) that could get nodes a bit more synchronized.

> 3) Someone may start the in the middle of an hour.

We could build some logic to go around that, but that would seem like more complexity here (or for any other `watch` solution, like round-based) for a problem that I don't think is _that_ relevant. The node could be off-sync for at most an hour, which is already the worst-case scenario with this hourly-based approach anyway. So I'd rather not add further logic (like checking previous "price feed rounds" on startup) to go around that case, coming from the same mindset of not going too much out of our way to fix this "impossible" problem (the pricing desynchronization across nodes).

> Maybe it'd be enough to update the price at the beginning of each round?

IMO this would be more complex in the code (probably more LoC and requiring an integration with other bits of the codebase) and also increase the pricing risk by a lot. We could spend a whole day paying way more than we should, given ETH price volatility, while by updating every hour we get almost all data available from Chainlink oracles (I've seen them update roughly every hour), while also not being expensive in RPC calls etc. We could also increase this 1h period if the frequency seems like an issue tho.

---

In conclusion, IMO this is an approach with a good trade-off between code complexity and accuracy. You've probably given more thought about this problem than I have tho, so let me know if you have a strong opinion towards reimplementing this to be round-based, just dropping the truncated ticker and using a regular one, etc

eth/watchers/pricefeedwatcher.go
victorges: Yeah this was quite a pain to test TBH. Do you think I should just drop those tests?

eth/watchers/pricefeedwatcher.go
victorges: Was relatively simple to fix by switching to a `time.After` instead. Done :)

eth/watchers/pricefeedwatcher_test.go
victorges: Yeah these were quite not the breeze to implement haha

> The drawback of them is that the tests can be flaky. 

The fake tickers make sure they are not flaky (nor take too long)!

> Plus, with this fake ticker, I guess you had to add some complexity to the prod logic to make it testable....

Yeah, you can check the changes I had to made in the prod code here:
- [for testing truncated ticker](https://github.com/livepeer/go-livepeer/pull/2972/commits/4db815345c4c0d45d8f671dc9d6efabd3b391c11#diff-91f8f3d3888f4535a8a9b736fbb8b471b29155db3611c1c520b5105d4e6a7dd3L137)
- [for testing `watch()`](https://github.com/livepeer/go-livepeer/pull/2972/commits/59450bf156a3a1a3234c426c2d3251c837ab9599#diff-91f8f3d3888f4535a8a9b736fbb8b471b29155db3611c1c520b5105d4e6a7dd3L98)

In the end I don't think it was too bad, but let me know if you have a better idea (like dropping the tests, I'd be happy to lol).

eth/watchers/pricefeedwatcher.go
victorges: Yeah good point. My bad not checking the other existing watchers. I didn't want to re-implementa logic to track multiple subscriptions, but that's already abstracted for me! Could just use an `event.Feed` here as well. 

Implemented both suggestions.

eth/watchers/pricefeedwatcher.go
victorges: Yeah good points. I switched to an `event.Feed` and explicit `Subscribe()` now, so we wont' have this automatic dropping of events (OTOH we might block if there's no receivers, but the explicit Subscribe/Unsubscribe is enough protection from that IMO).

eth/watchers/pricefeedwatcher.go
leszko: TBH I'd probably drop the tests.

leszko (approved):
Added one more comment. Other than that LGTM.

eth/watchers/pricefeedwatcher.go
victorges: dropped truncated ticker tests! the watch ones seemed worth to keep, mocking the ticker didnt seem that weird

PR #2981, opened 2024-03-21
cmd/livepeer: Use price feed watcher for dynamic pricePerPixel

victorges:
This uses the Chainlink price feed watcher component created on #2972 to
support dynamic transcode pricing for Broadcasters and Orchestrators. It works
by simply updating the `LivepeerNode.SetBasePrice` or `server.BroadcastCfg.SetMaxPrice`
when there are price updates on the price feed.

- Make the price flags strings to support currency specification
- Create new flag for the price feed address
- Start price update loop for Bs and Os

- `go test` 
- Ran B/O and checked price update happens as expected

Implements ENG-1726

**Checklist:**


cmd/livepeer/starter/starter.go
```diff
@@ -1494,6 +1524,73 @@ func parseEthKeystorePath(ethKeystorePath string) (keystorePath, error) {
 	return keystore, nil
 }
 
+func parsePricePerUnit(pricePerUnitStr string) (*big.Rat, string, error) {
+	match := pricePerUnitRex.FindStringSubmatch(pricePerUnitStr)
```
leszko, line 1576: nit: I'd probably inline this regexp and add a comment. Otherwise, it's very hard to reason about what's going on.
```suggestion
        pricePerUnitRex := regexp.MustCompile(`^(\d+(\.\d+)?)([A-z][A-z0-9]*)?$`)
	match := pricePerUnitRex.FindStringSubmatch(pricePerUnitStr)
```

cmd/livepeer/starter/starter.go
```diff
@@ -1494,6 +1524,73 @@ func parseEthKeystorePath(ethKeystorePath string) (keystorePath, error) {
...
+	if match == nil {
+		return nil, "", fmt.Errorf("price must be in the format of <price><currency>, provided %v", pricePerUnitStr)
+	}
+	price, currency := match[1], match[3]
+
+	pricePerUnit, ok := new(big.Rat).SetString(price)
+	if !ok {
+		return nil, "", fmt.Errorf("price must be a valid number, provided %v", match[1])
+	} else if pricePerUnit.Sign() < 0 {
```
leszko: Why do you check it both here and in `starter.go`?

cmd/livepeer/starter/starter.go
```diff
@@ -1494,6 +1524,73 @@ func parseEthKeystorePath(ethKeystorePath string) (keystorePath, error) {
...
+	}
+
+	base, quote := watcher.Currencies()
+	if base != "ETH" && quote != "ETH" {
+		return fmt.Errorf("price feed does not have ETH as a currency (%v/%v)", base, quote)
+	}
+	if base != currency && quote != currency {
+		return fmt.Errorf("price feed does not have %v as a currency (%v/%v)", currency, base, quote)
+	}
+	updatePrice(priceDataToWei(pricePerUnit, watcher.Current(), base))
```
leszko: What if price update fails here? Shouldn't we return an error (or at least some warning), so that someone knows that something can be misconfigured?

cmd/livepeer/starter/starter.go
```diff
@@ -737,16 +746,26 @@ func StartLivepeer(ctx context.Context, cfg LivepeerConfig) {
...
 				panic(fmt.Errorf("-pixelsPerUnit must be > 0, provided %d", *cfg.PixelsPerUnit))
 			}
+			pixelsPerUnit := new(big.Rat).SetInt64(int64(*cfg.PixelsPerUnit))
 			if cfg.PricePerUnit == nil {
 				// Prevent orchestrators from unknowingly providing free transcoding
 				panic(fmt.Errorf("-pricePerUnit must be set"))
 			}
-			if *cfg.PricePerUnit < 0 {
-				panic(fmt.Errorf("-pricePerUnit must be >= 0, provided %d", *cfg.PricePerUnit))
+			pricePerUnit, currency, err := parsePricePerUnit(*cfg.PricePerUnit)
```
leszko, line 760: I think you'll also need to add parsing here: https://github.com/livepeer/go-livepeer/blob/a3e725c956ad271dc6b0aeda6025053bd4a6f72f/server/handlers.go#L124

That makes me think that maybe we should store the "unparsed" price in Broadcaster in `BroadcastConfg.maxPrice` and only later convert. Think about the following scenario:
1. Someone starts Broadcaster with `-maxPricePerUnit=500USD`.
2. The `startPriceUpdateLoop` is started
3. Someone changes the Broadcaster maxPricePerUnit to `300`
4. The loop `startPriceUpdateLoop` should not be running anymore

The same in the opposite situation.

cmd/livepeer/starter/starter.go
```diff
@@ -737,16 +746,26 @@ func StartLivepeer(ctx context.Context, cfg LivepeerConfig) {
...
+			if err != nil {
+				panic(fmt.Errorf("-pricePerUnit must be a valid integer with an optional currency, provided %v", *cfg.PricePerUnit))
+			} else if pricePerUnit.Sign() < 0 {
+				panic(fmt.Errorf("-pricePerUnit must be >= 0, provided %s", pricePerUnit))
+			}
+			err = startPriceUpdateLoop(ctx, *cfg.EthUrl, *cfg.PriceFeedAddr, pricePerUnit, currency, func(price *big.Rat) {
+				n.SetBasePrice("default", new(big.Rat).Quo(pricePerUnit, pixelsPerUnit))
```
leszko: The same comment as above. Unfortunately, we do allow dynamic price changes for both B and O. So, we need to covert this as well.

cmd/livepeer/starter/starter.go
victorges: Oh good catch. I had implemented it here for the broadcaster then realized the orchestrator check was slightly different, but didn't remove it here.

cmd/livepeer/starter/starter.go
victorges: There is no possibility of failure here! This config is just updated in memory as it used to be before. Do you think we are missing any safety checks here? One I had thought was protecting against 0/negative prices on the price feed, but that seemed like too much of a corner case.

cmd/livepeer/starter/starter.go
victorges: Ok, that was a big miss in this PR. Spent quite some time thinking how I could abstract this on top of the current architecture and had to refactor a lot of stuff. I've proposed it in the commit https://github.com/livepeer/go-livepeer/pull/2981/commits/87907f27410f8bdcfaf2b0ff04f6ad58e06682cb.

None of the ideas I could think of for this seemed ideal. I didn't want to add a lot of logic on the "pure config" components like `BroadcasterCfg`, while at the same time didn't want to do a bigger refactor and remove the pricing info from them. In this solution I went with, I don't like that `onUpdate` part, which also causes repetition of the logs and monitor code. At the same time it felt like an auxiliary/non-complex thing, so not worth abstracting further. 

One thing to consider is if it'd be better to just remove the prices from the "dumb" config components and just move them to a `core.TranscodingPriceProvider` or sth like that which gets used instead. This might incur a bigger refactor, but could be simpler to reason about and maintain. Let me know if you'd prefer that or any other idea you have.

cmd/livepeer/starter/starter.go
```diff
@@ -712,6 +717,13 @@ func StartLivepeer(ctx context.Context, cfg LivepeerConfig) {
 		go serviceRegistryWatcher.Watch()
 		defer serviceRegistryWatcher.Stop()
 
+		core.PriceFeedWatcher, err = watchers.NewPriceFeedWatcher(backend, *cfg.PriceFeedAddr)
+		if err != nil {
+			glog.Errorf("Failed to set up price feed watcher: %v", err)
+			return
+		}
+		// The price feed watch loop is started on demand on first subscribe.
```
leszko: Nit: maybe move this comment above.

cmd/livepeer/starter/starter.go
```diff
@@ -712,6 +717,13 @@ func StartLivepeer(ctx context.Context, cfg LivepeerConfig) {
 		go serviceRegistryWatcher.Watch()
 		defer serviceRegistryWatcher.Stop()
 
+		core.PriceFeedWatcher, err = watchers.NewPriceFeedWatcher(backend, *cfg.PriceFeedAddr)
```
leszko, line 720: Won't this fail if someone does not set `priceFeedAddr` flag (or sets the addr for the wrong chain)? I think we should not require setting (correct) `priceFeedAddr` if someone does not want to use price conversion.

cmd/livepeer/starter/starter.go
```diff
@@ -737,23 +749,37 @@ func StartLivepeer(ctx context.Context, cfg LivepeerConfig) {
...
 			}
-			if *cfg.PricePerUnit < 0 {
-				panic(fmt.Errorf("-pricePerUnit must be >= 0, provided %d", *cfg.PricePerUnit))
+			pricePerUnit, currency, err := parsePricePerUnit(*cfg.PricePerUnit)
+			if err != nil {
+				panic(fmt.Errorf("-pricePerUnit must be a valid integer with an optional currency, provided %v", *cfg.PricePerUnit))
+			} else if pricePerUnit.Sign() < 0 {
+				panic(fmt.Errorf("-pricePerUnit must be >= 0, provided %s", pricePerUnit))
+			}
+			pricePerPixel := new(big.Rat).Quo(pricePerUnit, pixelsPerUnit)
```
leszko, line 766: Why do we use `new(big.Rat).Quo(pricePerUnit, pixelsPerUnit)` and in handlers `big.NewRat()`? 🤔

cmd/livepeer/starter/starter.go
victorges: It won't! I removed the preemptive calls from the `NewPriceFeedWatcher` constructor (`Description()` and `updatePrice()`) to make sure that we wouldn't fail here. Just checked the implementation of the bindings and they also don't make any checks on the provided address before actually using it on a call. I'll double check running these with some invalid address configs, but I believe it should be safe already. Let me know if you think it deserves some explicit logic (or comment?) here though.

cmd/livepeer/starter/starter.go
victorges: `big.NewRat` takes 2 `int64` while `Quo` takes 2 `*big.Rat`, so depending on the context we need one or the other

core/livepeernode.go
```diff
@@ -142,14 +147,18 @@ func (n *LivepeerNode) GetBasePrice(b_eth_addr string) *big.Rat {
 	n.mu.RLock()
 	defer n.mu.RUnlock()
 
-	return n.priceInfo[addr]
+	return n.priceInfo[addr].Value()
```
leszko:
```suggestion
	price := n.priceInfo[addr]
	if price != nil {
		return price.Value()
	}
	return nil
```

I think you need to check if n.priceInfo[addr] is not nil.

cmd/livepeer/starter/starter.go
leszko: 🆗

cmd/livepeer/starter/starter.go
leszko: Perfect, thanks for fixing it!

cmd/livepeer/starter/starter.go
```diff
@@ -737,23 +749,37 @@ func StartLivepeer(ctx context.Context, cfg LivepeerConfig) {
...
+			autoPrice, err := core.NewAutoConvertedPrice(currency, pricePerPixel, func(price *big.Rat) {
+				glog.Infof("Price: %v wei per pixel\n ", price.RatString())
```
leszko: How about converting the format to decimal while printing? Otherwise you get a message like this:
```
Price: 74000000000000000000/73215931 wei per pixel
```
Which is very hard to understand 🙃

cmd/livepeer/livepeer.go
```diff
@@ -133,7 +133,7 @@ func parseLivepeerConfig() starter.LivepeerConfig {
 	cfg.SelectPriceExpFactor = flag.Float64("selectPriceExpFactor", *cfg.SelectPriceExpFactor, "Expresses how significant a small change of price is for the selection algorithm; default 100")
 	cfg.OrchPerfStatsURL = flag.String("orchPerfStatsUrl", *cfg.OrchPerfStatsURL, "URL of Orchestrator Performance Stream Tester")
 	cfg.Region = flag.String("region", *cfg.Region, "Region in which a broadcaster is deployed; used to select the region while using the orchestrator's performance stats")
-	cfg.MaxPricePerUnit = flag.Int("maxPricePerUnit", *cfg.MaxPricePerUnit, "The maximum transcoding price (in wei) per 'pixelsPerUnit' a broadcaster is willing to accept. If not set explicitly, broadcaster is willing to accept ANY price")
+	cfg.MaxPricePerUnit = flag.String("maxPricePerUnit", *cfg.MaxPricePerUnit, "The maximum transcoding price per 'pixelsPerUnit' a broadcaster is willing to accept. If not set explicitly, broadcaster is willing to accept ANY price. Can be specified in wei or a custom currency in the format <price><currency> (e.g. 0.50USD). When using a custom currency, a corresponding price feed must be configured with -priceFeedAddr")
```
leszko, line 136: I wonder what should be the format here, `USD`, pico USD? Tagging also @hthillman here to think about UX :)

If we stick to USD, then with the current max price per pixel (`200`) and the current ETH / USD rate (`3,667.86`), Os would need to set this price to `200 * 3667.86 * 0.000000000000000001 = 0.0000000000007USD`.

So, the param will be as follows:
```
-maxPricePerUnit=0.0000000000007USD
```

Not sure if it's convenient to have so many zeros :)

cmd/livepeer/starter/starter.go
```diff
@@ -1420,29 +1461,52 @@ func checkOrStoreChainID(dbh *common.DB, chainID *big.Int) error {
...
+
+	// Format of broadcasterPrices json
+	// {"broadcasters":[{"ethaddress":"address1","priceperunit":0.5,"currency":"USD","pixelsperunit":1}, {"ethaddress":"address2","priceperunit":0.3,"currency":"USD","pixelsperunit":3}]}
+	var pricesSet struct {
+		Broadcasters []struct {
+			EthAddress    string          `json:"ethaddress"`
+			PricePerUnit  json.RawMessage `json:"priceperunit"`
+			Currency      string          `json:"currency"`
+			PixelsPerUnit int64           `json:"pixelsperunit"`
+		} `json:"broadcasters"`
```
leszko, line 1492: 1. Is this `json` used only while reading from file or do we also send it over the network? 
2. I wonder if it won't break the backwards-compatibility, because the json field `priceperunit` will be now string and not int. If this is only about the configuration read from the file, then maybe we don't need to be 100% backward-compatible, but it's something to keep in mind and announce to Os (and mention in the changelog).
3. Have you tested this part (at least manually)? I think there are many corner cases here, so that's definitely something that needs to be checked manually. What I mean by the manual check is if the values are read correctly from the file.

core/autoconvertedprice.go
```diff
@@ -0,0 +1,133 @@
...
+	cancelSubscription func()
+	onUpdate           func(*big.Rat)
+	basePrice          *big.Rat
+
+	mu      sync.RWMutex
+	current *big.Rat
+}
+
+// NewFixedPrice creates a new AutoConvertedPrice with a fixed price in wei.
+func NewFixedPrice(price *big.Rat) *AutoConvertedPrice {
```
leszko, line 34: Super nit: Is this used anywhere outside the package? If not, then consider making it package-private.

server/handlers.go
```diff
@@ -458,7 +470,7 @@ func (s *LivepeerServer) setOrchestratorConfigHandler(client eth.LivepeerEthClie
 	}))
 }
 
-func (s *LivepeerServer) setOrchestratorPriceInfo(broadcasterEthAddr, pricePerUnitStr, pixelsPerUnitStr string) error {
+func (s *LivepeerServer) setOrchestratorPriceInfo(broadcasterEthAddr, pricePerUnitStr, pixelsPerUnitStr, currency string) error {
```
leszko, line 475: I tried and for me it didn't work. Here's what I tried (you can try it with the local dev setup, with some mocking :) ).

1. I started livepeer with `-pricePerUnit 0.0000000000007USD`
2. Tried with `livepeer_cli`
```
 *-----------------------------*--------------------------------*
|        Base price per pixel |           70000000000000 wei / |
|                             |            360458402809 pixels |
*-----------------------------*--------------------------------*
| Base price for broadcasters |                                |
*-----------------------------*--------------------------------*
```
All good.

3. Then, I executed from `livepeer_cli`, `20 Set price for broadcaster`, and I've set the price per broadcaster and also the default one
4. Tried with `livepeer_cli`
*-----------------------------*---------------------------------------------*
|        Base price per pixel |                            0 wei / 1 pixels |
*-----------------------------*---------------------------------------------*
| Base price for broadcasters | 0xc3c7c4c8f7061b7d6a72766eee5359fe4f36e61e: |
|                             |                                0 per pixel  |
*-----------------------------*---------------------------------------------*

This looks wrong, not sure it's related to your change, but it can be. Could you please play with it more?

eth/watchers/pricefeedwatcher_test.go
```diff
@@ -58,20 +57,59 @@ func TestPriceFeedWatcher_UpdatePrice(t *testing.T) {
...
+
+	require.NotNil(w.cancelWatch)
+	for i := range observedCancelWatch {
+		require.Equal(reflect.ValueOf(w.cancelWatch).Pointer(), reflect.ValueOf(observedCancelWatch[i]).Pointer())
 	}
 
+	// Stop all subscriptions and ensure watch loop stops running
+	for i := range cancelSub {
+		cancelSub[i]()
+		if i < len(cancelSub)-1 {
```
leszko: 1. Why only one subscription should be `nil`? 🤔 
2. Won't this `time.Sleep()` make this test flaky?

leszko (commented):
Added some comments. It looks good, but I think it requires some more manual testing. Unfortunately go-livepeer is a beast and we don't have too many automated tests. Therefore, we need to check manually at least the following scenarios:
1. Set up local dev (or ask some public O to test it on Mainnet):
  - O: Set different prices per Broadcasters read from a file
  - O: Set dynamically prices per Broadcasters
  - O: Set dynamically default price from wei to USD and back to wei and back to USD
  - B: Set max price from commandline
  - B: Set max price dynamically
2. Check if automated tests work in Catalyst (enough to send a draft PR with the update [go-livepeer version](https://github.com/livepeer/catalyst/blob/e0105ba89985f0722be97579e7384ebc2b2c2a82/manifest.yaml#L50) to your commit hash)

cmd/livepeer/livepeer.go
victorges: I see a couple of options:
- Allow using the `e` notation for the price, so that it could be at least `-maxPricePerUnit=7e-13USD` 
  - Notice that we need to adapt the parsing function regex and the dynamic update APIs to support this
- Make use of the `pixelsPerUnit` configuration, which acts basically as a denominator for the price here. I'd say we should align on some value that will keep the `maxPricePerUnit` on a reasonable range.
  - For example, if we set the `-pixelsPerUnit` value to `1000000000000` (or `1e12`), we could set the `-maxPricePerUnit=0.70USD` here which feels more natural.

WDYT?

cmd/livepeer/starter/starter.go
victorges: 1. The type was only used here, so I understood it's only a local file being used as a config. It doesn't seem to get past the cli parsing logic tho
2. It will actually not be parsed as a string, but still a number! I've only made it a `json.RawMessage` so we wouldn't have to parse it as a `float64` here and potentially lose precision over the "string" that was present in the JSON. The parsing with `new(big.Rat).SetString` would fail if this value is a JSON string, since it would contain surrounding `"`. I've added a comment here to make that part mor explicit.
3. I've tested it on the unit tests only, but will do manually with some different configs.

core/autoconvertedprice.go
victorges: Yeah it needs to be used from tests in other packages like this https://github.com/livepeer/go-livepeer/blob/87907f27410f8bdcfaf2b0ff04f6ad58e06682cb/server/handlers_test.go#L262

In the end it's just a helper version that doesn't return an error, so I could avoid verborraging error handling in all tests that required a price.

server/handlers.go
victorges: Will do. I believe I'll also need to update `livepeer_cli` to support getting input on currency that it'd like to update to

eth/watchers/pricefeedwatcher_test.go
victorges: 1. The `cancelWatch` variable is used as an introspection here to know whether the watch loop is running or not. After unsubscribing the last subscription, the watch loop should stop and that variable is cleared as well, which is why the check changes for the last subscription. I rewrote this with some comments to make it clearer.
2. So this sleep is only waiting for the message from `cancelWatch()` to [end this loop](https://github.com/livepeer/go-livepeer/blob/cb39078434fa6600aed3131573ea90a71ae30e0a/eth/watchers/pricefeedwatcher.go#L145) and clear the `cancelWatch` variable (after calling it). It is an entirely "in memory" thing and given there is no I/O in these tests (the pricefeed client is mocked too), I find it extremely unlikely that that "message flow" is not going to complete in `1s` (the whole process would have to freeze I guess). I could try to make this sophisticated, in the tests (like a loop, checking multiple times) or in prod (have some other mechanism to propagate the change, like a `sync.Cond`), but had opted to keep it simple as I didn't think this could flake much. WDYT?

cmd/livepeer/livepeer.go
hthillman: @leszko @victorges great callout, I ran into this exact problem when working with Evan yesterday. My personal reaction is that the pixelsPerUnit configuration is a nicer way of handling this, BUT it will need to be well documented

hthillman: Specifically we'd need to recommend to Os how they should set it. 

Some Os may even keep setting prices in wei (which should still work?)

victorges: Yeah there's ample support for the "pixels per unit" config, it already existed before this change. It will also be possible to keep setting prices in wei :)

server/handlers.go
victorges: Ok I've implemented (and tested e2e) the `livepeer_cli` support here: https://github.com/livepeer/go-livepeer/pull/2981/commits/13bc0a6b50ed85fe56e48284b67ef31465f5c27d

I've tested setting configs for B, O (base and per B). Also tested with some invalid values as well as switching back and forth between wei/usd and checked we don't keep the price feed loop running. 

I was not able to test the catalyst build with a pending go-livepeer version though. I get errors like this [1] on the steps that download go-livepeer. Do you know if I have to trigger something in this repo for those on catalyst to work?

[1]
```
0327 00:19:58.855627 2263829 utils.go:91] Downloading https://github.com/livepeer/playbacklog_uploader/releases/download/v1.3.2/livepeer-mist-bigquery-uploader-linux-arm64.tar.gz.sig
I0327 00:19:58.893342 2263829 utils.go:100] Response statusCode=404
F0327 00:19:58.893403 2263829 downloader.go:244] failed to download livepeer: HTTP 404 while downloading https://build.livepeer.live/go-livepeer/13bc0a6b50ed85fe56e482[84](https://github.com/livepeer/catalyst/actions/runs/8444445609/job/23129999424#step:5:85)b67ef31465f5c27d/livepeer-linux-arm64.tar.gz
```

cmd/livepeer/livepeer.go
leszko: Ok, fine with me 👍

cmd/livepeer/starter/starter.go
leszko: 👍

eth/watchers/pricefeedwatcher_test.go
leszko: Ok 👌

leszko (approved):
LGTM 👍

cmd/livepeer/livepeer.go
victorges: OK, I've added this commit https://github.com/livepeer/go-livepeer/pull/2981/commits/ed7dc9b7285e4f82ba795636c7dbc97aaaa405e7 which makes sure we can always specify the `pixelsPerUnit` with an exponential notation (livepeer args, livepeer_cli or json file). I think this is enough, with the idea being to set a nice number for `pixelsPerUnit` with an easy syntax like `1e12`, and then the `pricePerPixel` can be a small enough number that should be fine to specify on the regular syntax (we won't support `e` notation for that in some flows (args)).

One other thing I could add would be forcing a different default value for the `pixelsPerUnit` when the currency being used is not wei (e.g. on the livepeer_cli). We can add that in a follow-up PR if you think it's useful.

PR #2985, opened 2024-03-28
server,discovery: Fallback to using any O in case none match maxPrice 

victorges:
This is to allow broadcasters to use any Orchestrator, in case their maxPrice configuration
is set to a too low number such that no Orchestrators offer that service.

This is a fallback strategy after implementing the feature that makes the price per pixel
dynamic based on an external currency. It mitigates the risk of the price quote changing 
drastically and the maxPrice becoming automatically too low. When that happens, the B
is going to fallback to any O for a while until the Os update their own prices.

We should also have an alert whenever this happens.

- Make `DBOrchestratorPoolCache` fetch all Os in case no Os match `maxPrice`
- Make `segment_rpc` allow O price to be higher than maxPrice, but only if initial price was already higher
- Make `segment_rpc` allow O price to increase up to 2x before dropping the session

 - `./test.sh`
 - TODO: Run a couple B<->Os with changing prices and see what happens

Implements ENG-1855

**Checklist:**

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -12,6 +12,8 @@
 
 #### Broadcaster
 
+- [#2985](https://github.com/livepeer/go-livepeer/pull/2985) Fallback to using any O in case none match maxPrice (@victorges)
+
 #### Orchestrator
 
 #### Transcoder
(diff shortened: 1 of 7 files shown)


victorges:
Replaced by #2994 and #2995, while the actual fallback logic will be implemented in a different way on the selection algorithm.

PR #2995, opened 2024-04-02
server: Allow Os price to increase up to 2x mid-session

victorges:
This is to allow O prices to increase up to a given threshold (currently 2x)
in the middle of a session.

With #2892, we intended to make the session prices fixed, but since deploying
#2981 we noticed that some sessions are getting dropped due to these prices
changing anyway. So there is some other flow that is updating those prices and
we need to investigate and fix them.

To avoid rolling back on the USD-based price improvement, I'm proposing this
change so we can roll the behavior forward while we find the source of changing
prices in the B<->O logic. There is a risk of still dropping sessions but it should
not be significant over the few days until we make the protocol fix.

- Allow O pries to increase up to 2x since session start
- Fix/update tests

Ran automated tests.

Related to ENG-1855 but a bit parallel to that

**Checklist:**

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -10,6 +10,8 @@
 
 #### Broadcaster
 
+- [#2995](https://github.com/livepeer/go-livepeer/pull/2995) server: Allow Os price to increase up to 2x mid-session (@victorges)
+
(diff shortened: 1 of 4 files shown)


leszko (approved):
LGTM 👍

PR #2999, opened 2024-04-04
server,discovery: Allow B to use any O in case none match maxPrice

victorges:
This is to allow broadcasters to use any Orchestrator in case their maxPrice configuration
is set to a too low number such that no Orchestrators offer that service.

This is a fallback strategy after implementing the feature that makes the price per pixel
dynamic based on an external currency. It mitigates the risk of the price quote changing
drastically and the maxPrice becoming automatically too low. When that happens, the B
is going to fallback to any O for a while until the Os update their own prices.

This is a re-implementation of #2985, doing it inside the selection algorithm itself so that
we still fallback to other Os in case there are a few Os with `price < maxPrice` but that still
can't serve the segments (e.g. perf issues).

- Remove logic on `discovery` that filtered Os by maxPrice
- Add logic on selection algorithm to filter by maxPrice 
- At the same time, also include the logic to fallback to ignoring maxPrice, in case none match

- `./test.sh`
- Ran a B and 3 Os with different prices locally. Made sure that:
  - B still selects an O matching the maxPrice, if one is available
  - B selects an O higher than maxPrice, if no lower ones are available
  - B switches mid-stream between a lower and higher priced O, in case the lower ones become unavailable

Implements ENG-1855

**Checklist:**

### DIFF
--- a/server/selection_algorithm.go
+++ b/server/selection_algorithm.go
@@ -1,11 +1,13 @@
 package server
 
 import (
-	ethcommon "github.com/ethereum/go-ethereum/common"
-	"github.com/golang/glog"
 	"math"
+	"math/big"
 	"math/rand"
 	"time"
+
+	ethcommon "github.com/ethereum/go-ethereum/common"
+	"github.com/golang/glog"
 )
 
 var random = rand.New(rand.NewSource(time.Now().UnixNano()))
@@ -20,7 +22,7 @@ type ProbabilitySelectionAlgorithm struct {
 	PriceExpFactor float64
 }
 
-func (sa ProbabilitySelectionAlgorithm) Select(addrs []ethcommon.Address, stakes map[ethcommon.Address]int64, prices map[ethcommon.Address]float64, perfScores map[ethcommon.Address]float64) ethcommon.Address {
+func (sa ProbabilitySelectionAlgorithm) Select(addrs []ethcommon.Address, stakes map[ethcommon.Address]int64, prices map[ethcommon.Address]*big.Rat, perfScores map[ethcommon.Address]float64) ethcommon.Address {
 	filtered := sa.filter(addrs, perfScores)
 	probabilities := sa.calculateProbabilities(filtered, stakes, prices)
 	return selectBy(probabilities)
@@ -48,10 +50,10 @@ func (sa ProbabilitySelectionAlgorithm) filter(addrs []ethcommon.Address, scores
 	return res
 }
 
-func (sa ProbabilitySelectionAlgorithm) calculateProbabilities(addrs []ethcommon.Address, stakes map[ethcommon.Address]int64, prices map[ethcommon.Address]float64) map[ethcommon.Address]float64 {
+func (sa ProbabilitySelectionAlgorithm) calculateProbabilities(addrs []ethcommon.Address, stakes map[ethcommon.Address]int64, prices map[ethcommon.Address]*big.Rat) map[ethcommon.Address]float64 {
 	pricesNorm := map[ethcommon.Address]float64{}
 	for _, addr := range addrs {
-		p := prices[addr]
+		p, _ := prices[addr].Float64()
 		pricesNorm[addr] = math.Exp(-1 * p / sa.PriceExpFactor)
 	}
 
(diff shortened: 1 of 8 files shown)


server/selection_algorithm.go
```diff
@@ -20,38 +22,60 @@ type ProbabilitySelectionAlgorithm struct {
...
+func (sa ProbabilitySelectionAlgorithm) Select(addrs []ethcommon.Address, stakes map[ethcommon.Address]int64, maxPrice *big.Rat, prices map[ethcommon.Address]*big.Rat, perfScores map[ethcommon.Address]float64) ethcommon.Address {
+	filtered := sa.filter(addrs, maxPrice, prices, perfScores)
 	probabilities := sa.calculateProbabilities(filtered, stakes, prices)
 	return selectBy(probabilities)
 }
 
-func (sa ProbabilitySelectionAlgorithm) filter(addrs []ethcommon.Address, scores map[ethcommon.Address]float64) []ethcommon.Address {
-	if sa.MinPerfScore <= 0 || scores == nil || len(scores) == 0 {
-		// Performance Score filter not defined, return all Orchestrators
+func (sa ProbabilitySelectionAlgorithm) filter(addrs []ethcommon.Address, maxPrice *big.Rat, prices map[ethcommon.Address]*big.Rat, perfScores map[ethcommon.Address]float64) []ethcommon.Address {
```
leszko, line 31: Won't it work the same if we just use one filter, then another, like this? I think it's easier to reason about the code. 

```go
func (sa ProbabilitySelectionAlgorithm) filter(addrs []ethcommon.Address, maxPrice *big.Rat, prices map[ethcommon.Address]*big.Rat, perfScores map[ethcommon.Address]float64) []ethcommon.Address {
	filteredByPerfScore := sa.filterByPerfScore(addrs, perfScores)
	return sa.filterByMaxPrice(filteredByPerfScore, maxPrice, prices)
}

func (sa ProbabilitySelectionAlgorithm) filterByPerfScore(addrs []ethcommon.Address, scores map[ethcommon.Address]float64) []ethcommon.Address {
	if sa.MinPerfScore <= 0 || scores == nil || len(scores) == 0 {
		// Performance Score filter not defined, return all Orchestrators
		return addrs
	}

	var res []ethcommon.Address
	for _, addr := range addrs {
		if scores[addr] >= sa.MinPerfScore {
			res = append(res, addr)
		}
	}

	if len(res) == 0 {
		// If no orchestrators pass the filter, then returns all Orchestrators
		// That may mean some issues with the PerfScore service.
		glog.Warning("No Orchestrators passed min performance score filter, not using the filter")
		return addrs
	}
	return res
}

func (sa ProbabilitySelectionAlgorithm) filterByMaxPrice(addrs []ethcommon.Address, maxPrice *big.Rat, prices map[ethcommon.Address]*big.Rat) []ethcommon.Address {
	if maxPrice == nil || len(prices) == 0 {
		// Performance Score filter not defined, return all Orchestrators
		return addrs
	}

	var res []ethcommon.Address
	for _, addr := range addrs {
		if maxPrice.Cmp(prices[addr]) < 1 {
			res = append(res, addr)
		}
	}

	if len(res) == 0 {
		// If no orchestrators pass the filter, then returns all Orchestrators
		// It means that no orchestrators are below the max price
		glog.Warning("No Orchestrators passed max price filter, not using the filter")
		return addrs
	}
	return res
}
```

leszko (approved):
LGTM. Added one comment, other than that, it looks good.

One thing to keep in mind is that there is one corner case that will work "worse". I think it should not prevent us from merging and deploying it, but something to mention for the sake of awareness.

So, when we send `GetOrch` requests to all Os, we wait only 500ms for the reply. If no Os replies, then we increase this timeout to 1s. Then, to 2s, and so on.

So, in theory, it's possible that all Os that replies in 500ms will have price higher than max price, but the Os with price lower than max price will reply later. In result, we'll select an O with price higher than max price, while an O with price lower than max price exists.

server/selection_algorithm.go
victorges: Sure that works too. No strong opinion, I'll adopt your approach.

PR #3004, opened 2024-04-12
feat(ai): enable AI orchestrator discovery

rickstaa:
This commit incorporates the AIServiceRegistry contract address, superseding the conventional ServiceRegistry contract address. This strategic alteration streamlines the discovery process of AI Orchestrators within the AI Subnet, thereby bolstering network accessibility and interaction. While this approach serves as a swift workaround to enable the feature without extensive code modification, it's important to note that it may disrupt the existing transcoding discovery mechanism. We have to fix this if we want to merge the two networks in the future.

> [!WARNING]
> Do not merge into the main branch this pull request breaks the discovery of the regular transcoding network.

This pull request implements https://linear.app/livepeer-ai-spe/issue/LIV-90/design-and-implement-solution-for-advertising-separate-service-uri-for so that AI subnet orchestrators can be found by the AI gateway.

-  `serviceRegistryAddr` was changed to a newly deployed [AIServiceRegistroyProxy](https://arbiscan.io/address/0x73f1755f34c2e769209e0f91a0c385722ed81b9c) contract. 

I checked if the service URIs of the orchestrators were to the values stored in the [AIServiceRegistroyProxy](https://arbiscan.io/address/0x73f1755f34c2e769209e0f91a0c385722ed81b9c) contract.  I have yet to test if an on-chain broadcaster can still send jobs.

https://linear.app/livepeer-ai-spe/issue/LIV-90/design-and-implement-solution-for-advertising-separate-service-uri-for

**Checklist:**

### DIFF
--- a/doc/ai-subnet.md
+++ b/doc/ai-subnet.md
@@ -381,16 +381,10 @@ To ensure that _AI Gateway_ nodes can discover your _AI Subnet_ Orchestrator, yo
     foundryup
     ```
 
-2. Connect your _Mainnet Transcoding Network_ Orchestrator wallet to the Foundry cast using the following command:
+2. Invoke the `SetServiceURI` function on the [AIServiceRegistry](https://arbiscan.io/address/0x04C0b249740175999E5BF5c9ac1dA92431EF34C5) contract. Use the KeyStore file and password from your _Mainnet Transcoding Network_ Orchestrator wallet to execute this command:
 
     ```bash
-    cast wallet import <WALLET_NAME> --private-key <YOUR_PRIVATE_WALLET_KEY>
-    ```
-
-3. Call the `SetServiceURI` function on the [AIServiceRegistry](https://arbiscan.io/address/0x73f1755F34C2e769209E0F91a0c385722Ed81B9C) contract using the following command:
-
-    ```bash
-    cast send --account <WALLET_NAME> --rpc-url <ARBITRUM_RPC_URI> 0x73f1755F34C2e769209E0F91a0c385722Ed81B9C "setServiceURI(string)" https://<PUBLIC_ORCH_IP_OR_URL>:<PUBLIC_ORCH_PORT>
+    cast send --keystore '<PATH_TO_KEYSTORE_FILE>' --password '<PASSWORD>' --rpc-url <ARBITRUM_RPC_URI> 0x04C0b249740175999E5BF5c9ac1dA92431EF34C5 "setServiceURI(string)" https://<PUBLIC_ORCH_IP_OR_URL>:<PUBLIC_ORCH_PORT>
     ```
 
     If everything was successful, you should see the following output:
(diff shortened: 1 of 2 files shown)


rickstaa:
Related to https://github.com/livepeer/protocol/pull/642. See my comments in https://github.com/livepeer/protocol/pull/642#issuecomment-2052699883.

eth/client.go
```diff
@@ -211,7 +211,9 @@ func (c *client) setContracts(opts *bind.TransactOpts) error {
 
 	glog.V(common.SHORT).Infof("LivepeerToken: %v", c.tokenAddr.Hex())
 
-	serviceRegistryAddr, err := c.GetContract(crypto.Keccak256Hash([]byte("ServiceRegistry")))
+	// TODO: Replace with call to controller once `AIServiceRegistry` is added to the controller.
+	// serviceRegistryAddr, err := c.GetContract(crypto.Keccak256Hash([]byte("ServiceRegistry")))
+	serviceRegistryAddr := ethcommon.HexToAddress("0x73f1755F34C2e769209E0F91a0c385722Ed81B9C")
```
yondonfu: I think hardcoding like this will also break usage of the node with local devnets or testnets since the node will try to use this hardcoded address on every single chain. One way to fix this is by only hardcoding the address if connected to Arbitrum One - something like:

```
chainID, err := c.backend.ChainID()
if err != nil { ... }

arbitrumOneChainID := big.NewInt(42161)
if chainID.Cmp(arbitrumOneChainID) == 0 {
    // Hardcode
} else {
    serviceRegistryAddr, err := c.GetContract(...)
}

...
```

See [this reference](https://docs.arbitrum.io/build-decentralized-apps/reference/node-providers) for Arbitrum chain IDs.

doc/ai-subnet.md
```diff
@@ -369,6 +369,54 @@ The first and **recommended method** is to use the `ethOrchAddr` flag to set the
...
+
+1. Install the [Foundry](https://book.getfoundry.sh/getting-started/installation) smart contract development toolchain:
+
+    ```bash
+    curl -L https://foundry.paradigm.xyz | bash
+    source /home/<USERNAME>/.bashrc
+    foundryup
+    ```
+
+2. Connect your _Mainnet Transcoding Network_ Orchestrator wallet to the Foundry cast using the following command:
```
yondonfu: I think an easier flow could be to use the `--keystore` and `--password` flags supported by [cast send](https://book.getfoundry.sh/reference/cli/cast/send?highlight=keystore#cast-send) since Os use keystore files to run go-livepeer already. 

Ex. to call `setServiceURI()`

```
cast send --keystore <PATH_TO_KEYSTORE_FILE> --password <PASSWORD> ... "setServiceURI(string)" ...
```

This would also avoid the need to directly deal with private keys since the user only needs to deal with keystore files.

eth/client.go
rickstaa: Great point! I've made the change. I now have the non-upgradable `AIServiceRegistry`, which is hardcoded to the mainnet [controller](https://arbiscan.io/address/0xD8E8328501E9645d16Cf49539efC04f734606ee4). Additionally, I deployed a proxy contract that also links to the mainnet [controller](https://arbiscan.io/address/0xD8E8328501E9645d16Cf49539efC04f734606ee4) for our use. I believe both configurations are suitable; see my detailed comment in [this GitHub discussion](https://github.com/livepeer/protocol/pull/642#issuecomment-2062580312).

PR #3013, opened 2024-04-15
feat(ai): add pipelines optimization flags

rickstaa:
> [!WARNING]
> Do not merge before https://github.com/livepeer/ai-worker/pull/61.

This commit adds a new `OptimizationFlags` field to the `aiModels` config so that users can forward optimization environment variables to the [ai-worker](git@github.com:livepeer/ai-worker.git) for more information see https://github.com/livepeer/ai-worker/pull/61.

- New `OptimizationFlags` field added to `aiModels` config.
- Flags forwaded to the [ai-worker](git@github.com:livepeer/ai-worker.git) Warm function.
- A warning is thrown when users try to use the `optimizationFlags` with non warm containers.

1. Ensured the package builds.
2. Ensured tests.sh were succesfull.
3. Spung up off-chain broadcaster.
4. Spung up off-chain orchestrator with `SFAST` optimization flag in the models configuration file.

```bash
[
    {
        "pipeline": "image-to-video",
        "model_id": "stabilityai/stable-video-diffusion-img2vid-xt-1-1",
        "price_per_unit": 3390842,
        "warm": true,
        "optimization_flags": {
            "SFAST": "true",
            "SOME": true
        }
    }
]

```

6. Let the ImageToVideo container warmup and check the container logs to see that SFAST is used and the model is pre-traced.
7. Send a ImageToVideo request to the broadcaster and checked that it stilled worked.

No.

**Checklist:**

- [ ] All tests in `./test.sh` pass - @rickstaa we still need to fix the tests.

### DIFF
--- a/cmd/livepeer/starter/starter.go
+++ b/cmd/livepeer/starter/starter.go
@@ -18,7 +18,6 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
-	"sync"
 	"time"
 
 	ethcommon "github.com/ethereum/go-ethereum/common"
@@ -539,7 +538,6 @@ func StartLivepeer(ctx context.Context, cfg LivepeerConfig) {
 				return
 			}
 
-			var once sync.Once
 			for _, config := range configs {
 				modelConstraint := &core.ModelConstraint{Warm: config.Warm}
 
@@ -555,9 +553,7 @@ func StartLivepeer(ctx context.Context, cfg LivepeerConfig) {
 
 				// Show warning if people set OptimizationFlags but not Warm.
 				if len(config.OptimizationFlags) > 0 && !config.Warm {
-					once.Do(func() {
-						glog.Warningf("OptimizationFlags set for model %v but Warm is not set. OptimizationFlags are currently only used for warm containers.", config.ModelID)
-					})
+					glog.Warningf("Model %v has 'optimization_flags' set without 'warm'. Optimization flags are currently only used for warm containers.", config.ModelID)
 				}
 
 				switch config.Pipeline {
--- a/cmd/livepeer/starter/starter.go
+++ b/cmd/livepeer/starter/starter.go
@@ -18,6 +18,7 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	ethcommon "github.com/ethereum/go-ethereum/common"
@@ -538,6 +539,7 @@ func StartLivepeer(ctx context.Context, cfg LivepeerConfig) {
 				return
 			}
 
+			var once sync.Once
 			for _, config := range configs {
 				modelConstraint := &core.ModelConstraint{Warm: config.Warm}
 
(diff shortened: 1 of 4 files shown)


rickstaa:
@yondonfu, is there a specific reason why the binary execution proceeds even when an AIModelConfig is not specified? I encountered this issue while trying to activate flags for containers that don't start warm, as I wanted to pass Optimization flags directly to the AIWorker constructor.

https://github.com/livepeer/go-livepeer/blob/1019d429eca743218fe133602b5f4344e4766308/cmd/livepeer/starter/starter.go#L534-L605

rickstaa:
Requires https://github.com/livepeer/ai-worker/pull/61.

yondonfu:
@rickstaa 

> is there a specific reason why the binary execution proceeds even when an AIModelConfig is not specified?

I think it would sense to require a config file for `-aiModels` now because the config file is not only used to warm models, but also for specifying prices and determining which models to advertise during capability discovery. Prior to using the config file for configuring prices and capability discovery, the config file was left as optional because the O would just try to execute any request given a model ID. But now, since models have prices and have to be explicitly advertised, the config file is necessary.

cmd/livepeer/starter/starter.go
```diff
@@ -538,19 +539,27 @@ func StartLivepeer(ctx context.Context, cfg LivepeerConfig) {
...
+					if err := n.AIWorker.Warm(ctx, config.Pipeline, config.ModelID, endpoint, config.OptimizationFlags); err != nil {
 						glog.Errorf("Error AI worker warming %v container: %v", config.Pipeline, err)
 						return
 					}
 				}
 
+				// Show warning if people set OptimizationFlags but not Warm.
+				if len(config.OptimizationFlags) > 0 && !config.Warm {
+					once.Do(func() {
+						glog.Warningf("OptimizationFlags set for model %v but Warm is not set. OptimizationFlags are currently only used for warm containers.", config.ModelID)
```
yondonfu: Wouldn't you want to print this warning for each model that does not have `warm: true` and is also configured to use `optimization_flags`?

yondonfu: Since the JSON fields actually set by users use snake case IMO it makes sense to use `optimization_flags` and `warm` here instead of `OptimizationFlags` and `Warm` since those are the fields that the user would be familiar with.

yondonfu (commented):
Noting for the thread that this will need a go mod update to use the latest version of `ai-worker` once it is merged.

rickstaa:
> @rickstaa
> 
> > is there a specific reason why the binary execution proceeds even when an AIModelConfig is not specified?
> 
> I think it would sense to require a config file for `-aiModels` now because the config file is not only used to warm models, but also for specifying prices and determining which models to advertise during capability discovery. Prior to using the config file for configuring prices and capability discovery, the config file was left as optional because the O would just try to execute any request given a model ID. But now, since models have prices and have to be explicitly advertised, the config file is necessary.

Ah thanks for the explanation. Makes sense. I will change the behavoir!

PR #3020, opened 2024-04-18
Fix truncated PNG'

stronk-dev:
Fixes the issue that PNG's returned by AI-workers would be cut off at the end, causing issues with most image viewers


rickstaa:
Thanks @stronk-dev will review it 👍🏻. Fixes https://github.com/livepeer/ai-worker/issues/50.

PR #3029, opened 2024-04-22
Initialize round by any B/O who has the initializeRound flag set to true

leszko:
AFAIU the current logic for the `-initializeRound` flag has the assumption that every O has this flag set to true. Then, it uses some calculation to decide which O sends the tx to initialize the round.

The assumption is not true and with most O having `-initializeRound false` we end up with the round not being initialized. This PR removes all this logic for checking who should intialize the round and just any B/O that has this flag enabled will try to initialize the round. The worst that can happen is that a lot of B/Os will try to initialize the round, but the gas price is low, so that's not too bad.

After this PR gets merged I'll set it to true in all Bs in staging for Livepeer Inc.


eth/roundinitializer.go
```diff
@@ -104,23 +103,7 @@ func (r *RoundInitializer) tryInitialize() error {
 	r.mu.Lock()
 	defer r.mu.Unlock()
 
-	currentL1Blk := r.tw.LastSeenL1Block()
-	lastInitializedL1BlkHash := r.tw.LastInitializedL1BlockHash()
-
-	epochSeed := r.currentEpochSeed(currentL1Blk, r.nextRoundStartL1Block, lastInitializedL1BlkHash)
-
-	ok, err := r.shouldInitialize(epochSeed)
```
victorges, line 112: While the previous logic seemed too complex while at the same time didn't work in practice, I think not having any logic at all to prevent concurrent (or just repeated) round initialization is a big waste here.

While a single failed round initialization transaction is cheap ([$0.002](https://arbiscan.io/tx/0xe51fbb2b082020bb606b37ad849b1de82ed938b4174719a989d22232e5d2749c)), IMO it doesn't sound as cheap if we think about the global cost of having _several_ failed transactions for every round. With 100 orchestrators globally, plus 27 catalysts and 120 broadcasters, we'd basically be wasting $0.50 every day from our protocol, as well as polluting the onchain history for the round initializer contract.

I tried giving some thought on this and I think a relatively simple but potentially good enough solution could be:
- Add a randomized delay here, so we spread nodes across different times
  - This could be deterministic from the address instead, problem is all our Bs would end in the same bucket
  - I thought of a short delay from 0-10s or 0-20s (avg 10s). This should be enough for most nodes to realize the round is already initialized and avoid repeating it.
- Before actually initializing the round, check if it's already initialized first by calling `r.client.CurrentRoundInitialized()`. This is a read transaction that won't incur gas costs

WDYT?

victorges (approved):
LGTM

I have a suggestion for a logic to avoid so many failed transactions onchain every day, but if you prefer to continue with the current version it's fine with me. I'd just request for us not to enable the `-initializeRound` flag on all of our Bs, otherwise it's gonna be a lot of pollution onchain.

eth/roundinitializer.go
leszko: Yeah, good, let's introduce a random delay. I think it does not need to be deterministic, simple rand should work.

leszko: @victorges Added the delay logic. PTAL.

eth/roundinitializer.go
```diff
@@ -103,6 +107,22 @@ func (r *RoundInitializer) tryInitialize() error {
...
 
+	if r.tw.LastSeenL1Block().Cmp(r.nextRoundStartL1Block) < 0 {
+		// Round already initialized
+		return nil
+	}
+
+	if r.maxDelay > 0 {
+		randDelay := time.Duration(rand.Int63n(int64(r.maxDelay)))
+		glog.Infof("Waiting %v before attempting to initialize round", randDelay)
+		time.Sleep(randDelay)
```
victorges, line 118: Is there a risk that L89 would be triggered multiple times as we're sleeping here? IIUC, `l1BlockSink` will stream every new block seen on the chain, so if we wait here up to 30s we could have multiple blocks showing up and multiple `tryInitialize` functions executing (likely sleeping) on the same node. 

If that's the case, wdyt of adding some kind of mutexing here? One built-in solution could be a [singleflight](https://pkg.go.dev/golang.org/x/sync/singleflight) but maybe a manual mutex might be simpler.

victorges (approved):
LGTM!

If you're merging this without squash keep in mind that there's a duplicate commit (msg) in the history (I just prefer squash myself)

eth/roundinitializer.go
victorges: Hmm, or maybe simpler could also be to have the sleep logic outside the go-routine, blocking the main watch loop. Not sure, will leave it up to you!

leszko: There is a mutex in L107, so the worst that can happen is that while waiting we'll create a few unnecessary goroutines. But that should be ok.

PR #3030, opened 2024-04-23
fix(ai): improve AI selection algorithm

rickstaa:

This pull request modifies the selection algorithm to continue retrying for a duration of one second instead of stopping after four attempts. This change addresses issues encountered with the current algorithm's performance in environments with 15 nodes on the network, ensuring more robust and reliable operation until further optimizations can be implemented.

- The range based AI select logic was replaced by a timeout to ensure more Orchstrators get selected.

I tested it offchain/onchain on my development Gateway.

**Checklist:**

### DIFF
--- a/server/ai_process.go
+++ b/server/ai_process.go
@@ -12,6 +12,7 @@ import (
 	"net/http"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"github.com/livepeer/ai-worker/worker"
 	"github.com/livepeer/go-livepeer/clog"
@@ -21,7 +22,7 @@ import (
 	"github.com/livepeer/lpms/stream"
 )
 
-const maxProcessingRetries = 4
+const maxProcessingRetryDuration = 500 * time.Millisecond
 const defaultTextToImageModelID = "stabilityai/sdxl-turbo"
 const defaultImageToImageModelID = "stabilityai/sdxl-turbo"
 const defaultImageToVideoModelID = "stabilityai/stable-video-diffusion-img2vid-xt"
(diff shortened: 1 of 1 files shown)


rickstaa:
Here's @yondonfu; this straightforward solution might help ease the situation. I have several other PRs lined up to fix the double selection, but this one is ready for merging if there are no objections. I also am finishing a tech spec explaining all the improvements that can be made. The only downside I see is the increased cost because we pay orch's for returning an `insufficient capacity` error. To address this, I've created a new pull request, which you can review here: [PR #3035](https://github.com/livepeer/go-livepeer/pull/3035).

eliteprox (approved):
Changes look good to me. This will make the selection process much more efficient

server/ai_process.go
```diff
@@ -311,8 +312,11 @@ func processAIRequest(ctx context.Context, params aiRequestParams, req interface
 
 	var resp *worker.ImageResponse
 
+	cctx, cancel := context.WithTimeout(context.Background(), processingRetryTimeout)
```
yondonfu, line 315: Nit: You'll want to pass `ctx` (param value for `processAIRequest()`) as the parent context here to ensure that if there is a cancellation further up the call stack that it properly propagates - no point in continuing here if no one is waiting for the result.

server/ai_process.go
```diff
@@ -334,6 +338,10 @@ func processAIRequest(ctx context.Context, params aiRequestParams, req interface
 		clog.Infof(ctx, "Error submitting request cap=%v modelID=%v try=%v orch=%v err=%v", cap, modelID, tries, sess.Transcoder(), err)
 
 		params.sessManager.Remove(ctx, sess)
+
+		if cctx.Err() == context.DeadlineExceeded {
```
yondonfu, line 342: More idiomatic Go would be:

```
switch {
    case <-cctx.Done():
        break
    default:
}
```

This should check if the context done signal was fired (i.e. we hit `processingRetryTimeout`) otherwise we'll fall to the default block and continue into another loop iteration.

rickstaa: Yondon. Thanks for your high quality feedback! Not sure why I implemented it like that 🤔. Your solution is a lot cleaner. Merged in https://github.com/livepeer/go-livepeer/commit/72dced7b710e407248f019117cee5e1777d00285e.

PR #3033, opened 2024-04-28
fix(ai): update ai-video selection suspension

ad-astra-video:

Suspension was not working because the `penalty` was always `3`.  This logic was a carryover from transcoding where the suspender always started at a refresh count of 0 because a new session manager was created with each stream.  For AI, we are reusing the session manager and the suspender so the refresh count does not reset between requests.  The fix to suspension is to consider the current refresh count when calculating the penalty so it is 3 more than the current refresh count in the suspender.

There was also an issue where the `discoveryPoolSize` was always 100 and with limited orchestrators providing models a refresh of sessions was being done with every request.  I added an `initialPoolSize` field to track the last refresh pool size to use with the `shouldRefreshSessions` logic rather than 100.   This stabilizes the suspender to allow more orchestrators to be tried with each `Select` call.

Last update was moving the `signalRefresh()` for the suspender that increments the refresh counter in the suspender to the `Refresh` function makes it more stable that every time we refresh sessions we add to the suspender refresh count

Happy to segregate some of these changes to separate PRs.  The suspension fixes can be added separately without dependency on ai-worker PR.

- Updates suspender to use the current refresh count of the suspender in the selector.
- Moves penalty to the AISessionSelector to make it easier to update and available for calculations on the suspension needed
- releases all Os when there are none in the warm and cold pool
- Adds option to not use managed containers.
- 

I have been running these updates on my gateway.  Tested 1-200 requests with 5-10 workers sending to gateway.  All completed with 1-2 orchestrators providing Bytedance model.

**Checklist:**

### DIFF
--- a/server/ai_http.go
+++ b/server/ai_http.go
@@ -119,7 +119,7 @@ func (h *lphttp) StartLiveVideoToVideo() http.Handler {
 
 		// Check if there is capacity for the request
 		if !orch.CheckAICapacity(pipeline, modelID) {
-			respondWithError(w, fmt.Sprintf("insufficient capacity for pipeline=%v modelID=%v", pipeline, modelID), http.StatusServiceUnavailable)
+			respondWithError(w, fmt.Sprintf("Insufficient capacity for pipeline=%v modelID=%v", pipeline, modelID), http.StatusServiceUnavailable)
 			return
 		}
 
@@ -480,7 +480,7 @@ func handleAIRequest(ctx context.Context, w http.ResponseWriter, r *http.Request
 
 	// Check if there is capacity for the request.
 	if !orch.CheckAICapacity(pipeline, modelID) {
-		respondWithError(w, fmt.Sprintf("insufficient capacity for pipeline=%v modelID=%v", pipeline, modelID), http.StatusServiceUnavailable)
+		respondWithError(w, fmt.Sprintf("Insufficient capacity for pipeline=%v modelID=%v", pipeline, modelID), http.StatusServiceUnavailable)
 		return
 	}
 
--- a/server/ai_http.go
+++ b/server/ai_http.go
@@ -119,7 +119,7 @@ func (h *lphttp) StartLiveVideoToVideo() http.Handler {
 
 		// Check if there is capacity for the request
 		if !orch.CheckAICapacity(pipeline, modelID) {
-			respondWithError(w, fmt.Sprintf("Insufficient capacity for pipeline=%v modelID=%v", pipeline, modelID), http.StatusServiceUnavailable)
+			respondWithError(w, fmt.Sprintf("insufficient capacity for pipeline=%v modelID=%v", pipeline, modelID), http.StatusServiceUnavailable)
 			return
 		}
 
@@ -480,7 +480,7 @@ func handleAIRequest(ctx context.Context, w http.ResponseWriter, r *http.Request
 
 	// Check if there is capacity for the request.
 	if !orch.CheckAICapacity(pipeline, modelID) {
-		respondWithError(w, fmt.Sprintf("Insufficient capacity for pipeline=%v modelID=%v", pipeline, modelID), http.StatusServiceUnavailable)
+		respondWithError(w, fmt.Sprintf("insufficient capacity for pipeline=%v modelID=%v", pipeline, modelID), http.StatusServiceUnavailable)
 		return
 	}
 
--- a/server/ai_session.go
+++ b/server/ai_session.go
@@ -128,9 +128,9 @@ func (pool *AISessionPool) Remove(sess *BroadcastSession) {
 	// If this method is called assume that the orch should be suspended
 	// as well.  Since AISessionManager re-uses the pools the suspension
 	// penalty needs to consider the current suspender count to set the penalty
-	last_count, ok := pool.suspender.list[sess.Transcoder()]
+	lastCount, ok := pool.suspender.list[sess.Transcoder()]
 	if ok {
-		penalty = pool.suspender.count - last_count + pool.penalty
+		penalty = pool.suspender.count - lastCount + pool.penalty
 	} else {
 		penalty = pool.suspender.count + pool.penalty
 	}
@@ -230,8 +230,7 @@ func newAICapabilities(cap core.Capability, modelID string, warm bool, minVersio
 
 func (sel *AISessionSelector) Select(ctx context.Context) *AISession {
 	shouldRefreshSelector := func() bool {
-		// Refresh if the # of sessions across warm and cold pools falls below the smaller of the maxRefreshSessionsThreshold and
-		// 1/2 the total # of orchs that can be queried during discovery
+
 		discoveryPoolSize := int(math.Min(float64(sel.node.OrchestratorPool.Size()), float64(sel.initialPoolSize)))
 
 		if (sel.warmPool.Size() + sel.coldPool.Size()) == 0 {
@@ -242,7 +241,8 @@ func (sel *AISessionSelector) Select(ctx context.Context) *AISession {
 				sel.suspender.signalRefresh()
 			}
 		}
-
+		// Refresh if the # of sessions across warm and cold pools falls below the smaller of the maxRefreshSessionsThreshold and
+		// 1/2 the total # of orchs that can be queried during discovery
 		if sel.warmPool.Size()+sel.coldPool.Size() < int(math.Min(maxRefreshSessionsThreshold, math.Ceil(float64(discoveryPoolSize)/2.0))) {
 			return true
 		}
--- a/server/ai_session.go
+++ b/server/ai_session.go
@@ -101,10 +101,6 @@ func (pool *AISessionPool) Add(sessions []*BroadcastSession) {
 	pool.mu.Lock()
 	defer pool.mu.Unlock()
 
-	// If we try to add new sessions to the pool the suspender
-	// should treat this as a refresh
-	pool.suspender.signalRefresh()
-
 	var uniqueSessions []*BroadcastSession
 	for _, sess := range sessions {
 		if _, ok := pool.sessMap[sess.Transcoder()]; ok {
@@ -272,6 +268,10 @@ func (sel *AISessionSelector) Remove(sess *AISession) {
 }
 
 func (sel *AISessionSelector) Refresh(ctx context.Context) error {
+	// If we try to add new sessions to the pool the suspender
+	// should treat this as a refresh
+	sel.suspender.signalRefresh()
+
 	sessions, err := sel.getSessions(ctx)
 	if err != nil {
 		return err
--- a/server/ai_session.go
+++ b/server/ai_session.go
@@ -29,14 +29,16 @@ type AISessionPool struct {
 	sessMap   map[string]*BroadcastSession
 	inUseSess []*BroadcastSession
 	suspender *suspender
+	penalty   int
 	mu        sync.RWMutex
 }
 
-func NewAISessionPool(selector BroadcastSessionsSelector, suspender *suspender) *AISessionPool {
+func NewAISessionPool(selector BroadcastSessionsSelector, suspender *suspender, penalty int) *AISessionPool {
 	return &AISessionPool{
 		selector:  selector,
 		sessMap:   make(map[string]*BroadcastSession),
 		suspender: suspender,
+		penalty:   penalty,
 		mu:        sync.RWMutex{},
 	}
 }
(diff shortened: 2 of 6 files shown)


server/ai_session.go
```diff
@@ -126,10 +124,17 @@ func (pool *AISessionPool) Remove(sess *BroadcastSession) {
...
 	pool.inUseSess = removeSessionFromList(pool.inUseSess, sess)
 
-	// Magic number for now
-	penalty := 3
+	penalty := 0
 	// If this method is called assume that the orch should be suspended
-	// as well
+	// as well.  Since AISessionManager re-uses the pools the suspension
+	// penalty needs to consider the current suspender count to set the penalty
+	last_count, ok := pool.suspender.list[sess.Transcoder()]
```
victorges: **nit/lint:** Vars in go should be `camelCase`

server/ai_session.go
```diff
@@ -222,7 +233,17 @@ func (sel *AISessionSelector) Select(ctx context.Context) *AISession {
 	shouldRefreshSelector := func() bool {
 		// Refresh if the # of sessions across warm and cold pools falls below the smaller of the maxRefreshSessionsThreshold and
 		// 1/2 the total # of orchs that can be queried during discovery
```
victorges: This comment seems out of place now, can you move it closer to L247?

server/ai_session.go
```diff
@@ -222,7 +233,17 @@ func (sel *AISessionSelector) Select(ctx context.Context) *AISession {
...
-		discoveryPoolSize := sel.node.OrchestratorPool.Size()
+		discoveryPoolSize := int(math.Min(float64(sel.node.OrchestratorPool.Size()), float64(sel.initialPoolSize)))
```
victorges, line 235: Why do we need this `Min` now? Can the pool grow from its initial size?

victorges (commented):
Feel like I don't have context to officially approve this, but left some comments. Only nits tho, the implementation makes sense for the PR description.

server/ai_session.go
```diff
@@ -126,10 +124,17 @@ func (pool *AISessionPool) Remove(sess *BroadcastSession) {
...
-	// Magic number for now
-	penalty := 3
+	penalty := 0
 	// If this method is called assume that the orch should be suspended
-	// as well
+	// as well.  Since AISessionManager re-uses the pools the suspension
+	// penalty needs to consider the current suspender count to set the penalty
+	last_count, ok := pool.suspender.list[sess.Transcoder()]
+	if ok {
+		penalty = pool.suspender.count - last_count + pool.penalty
```
leszko: I'm a little lost with this suspension logic. So, I see that:
- the `pool.suspender.count` is increased every time `signalRefresh()` is called
- `pool.penalty` is always set to `3`
- `last_count` is always set to `suspender.count + 3`

So, that logic would mean that we're not taking the suspended orchestrator until 3 times the `signalRefresh()` is called. Is this the idea of this suspension mechanism? That we don't allow the given O to get selected in the 3 refresh sessions?

server/ai_session.go
```diff
@@ -222,7 +233,17 @@ func (sel *AISessionSelector) Select(ctx context.Context) *AISession {
...
 		// 1/2 the total # of orchs that can be queried during discovery
-		discoveryPoolSize := sel.node.OrchestratorPool.Size()
+		discoveryPoolSize := int(math.Min(float64(sel.node.OrchestratorPool.Size()), float64(sel.initialPoolSize)))
+
+		if (sel.warmPool.Size() + sel.coldPool.Size()) == 0 {
+			// release all orchestrators from suspension and try refresh
+			// if there are no orchestrators in the pools
+			clog.Infof(ctx, "refreshing sessions, no orchestrators in pools")
+			for i := 0; i < sel.penalty; i++ {
+				sel.suspender.signalRefresh()
```
leszko, line 242: > release all orchestrators

shouldn't we then just remove them from the `suspender.list()` rather than calling `signalRefresh()`? My understanding is that if `penalty = 3`, then we would need to call `signalRefresh()` 3 times in order to "release all orchestrators from suspension".

rickstaa:
@leszko Are we still planning to merge these fixes? They help alleviate some of the orchestrator stickiness issues we've observed over the past few months.

leszko:
> @leszko Are we still planning to merge these fixes? They help alleviate some of the orchestrator stickiness issues we've observed over the past few months.

Well...if it helps Orchestrators, then yes. Why not? 🙃 Please address the PR review comments and re-request review :)

server/ai_session.go
ad-astra-video: Agree, moved down a few lines

server/ai_session.go
ad-astra-video: A session pool can grow with the orchestrator refreshes every hour if more start serving the model.  Was trying to set this so that if there is less than 8 orchestrators in an initial selector (e.g. 1-2 orchestrators serving a model) it wont try to refresh all sessions with each request.

server/ai_session.go
ad-astra-video: Yes thats right.  A small penalty to take the orchestrator out of the working set for short period.  With transcoding the refresh count starts at 0 for each new session.  For AI the session pools are used across requests and are not reset.  If we don't add the current suspender refresh count it wont suspend any orchestrators.

server/ai_session.go
ad-astra-video: fixed

server/ai_session.go
ad-astra-video: I did it this way thinking that there could be more than 3 orchestrators suspended so it would be less loops to just `signalRefresh()` 3 times. An alternative would be to just create a new suspender for the selector to clear it or kick all the orchs out of the suspended list (will require new function to do second option).

ad-astra-video:
@rickstaa i have updated this PR.

I also think adding something like this would be beneficial for Gateway operators to see the current AI session pool state.  WYDT? https://github.com/ad-astra-video/go-livepeer/commit/9d42251631b48e79710547842fceeb10af508686

EDIT: we included the /getAISessionsPoolInfo endpoint on the gateway cli webserver to provide data about AI pools.

ad-astra-video:
@rickstaa this is ready for review.  Titan did some testing and the suspension was working as expected.

[pool_info_1000.json](https://github.com/user-attachments/files/18653587/pool_info_1000.json)

[pool_info_1_orch.json](https://github.com/user-attachments/files/18653611/pool_info_1_orch.json)

core/ai_orchestrator.go
```diff
@@ -393,18 +393,43 @@ type AIJobRequestData struct {
 }
 
 // CheckAICapacity verifies if the orchestrator can process a request for a specific pipeline and modelID.
-func (orch *orchestrator) CheckAICapacity(pipeline, modelID string) bool {
+func (orch *orchestrator) CheckAICapacity(pipeline, modelID string) (bool, chan<- bool) {
```
leszko: If it's actually reserving the capacity, then shouldn't this actually be called `ReserveAICapacity()`?

core/ai_orchestrator.go
```diff
@@ -393,18 +393,43 @@ type AIJobRequestData struct {
...
+	var hasCapacity bool
 	if orch.node.AIWorker != nil {
 		// confirm local worker has capacity
-		return orch.node.AIWorker.HasCapacity(pipeline, modelID)
+		err := orch.node.ReserveAICapability(pipeline, modelID)
+		if err == nil {
```
leszko: What if err is not nil, but we reserved the capability? Won't the whole node stop working?

leszko (commented):
Added two comments. My suggestion is also to maybe split this PR to make it smaller, because I think it includes some unrelated changes.

1. Wrt reserving capacity inside `CheckAICapacity()`, it looks counter-intuitive, but I may not have the full context. It raises a lot of questions in my head, like what if we reserved the capacity, but the request failed at some point. Won't O stop working? Maybe we need to have some timeout to release it? 
2. Wrt: penalty mechanism, any chance it's possible to send this change in its own separate PR (or is it related to reserve capacity change?).

rickstaa:
> Added two comments. My suggestion is also to maybe split this PR to make it smaller, because I think it includes some unrelated changes.
> 
> 1. Wrt reserving capacity inside `CheckAICapacity()`, it looks counter-intuitive, but I may not have the full context. It raises a lot of questions in my head, like what if we reserved the capacity, but the request failed at some point. Won't O stop working? Maybe we need to have some timeout to release it?
> 2. Wrt: penalty mechanism, any chance it's possible to send this change in its own separate PR (or is it related to reserve capacity change?).

@ad-astra-video I agree with @leszko if we can have the suspense fix in a seperate pull request we can move quicker since that part is more battle tested and reviewd 🙏🏻.

core/ai_orchestrator.go
ad-astra-video: Removed this part to be in a separate PR. 

To answer the question tho, an error only comes back from this function if capability capacity is not reserved. Errors can be from pipeline is not compatible.  This function is also used in `runAIJob` in server/ai_worker.go.

core/ai_orchestrator.go
ad-astra-video: Removed this part to be in a separate PR.

I have no argument on what the name should be. Just tried to minize changes and test changes that would be needed with a function renaming.

ad-astra-video:
I have split the PR. If want split another way please tell me specifically which parts want in separate PRs.  To me, all the remaining changes are needed to stablize orchestrator pools for batch jobs so I have left in this PR.

I retested all batch pipelines after splitting PR.

leszko:
> I have split the PR. If want split another way please tell me specifically which parts want in separate PRs. To me, all the remaining changes are needed to stablize orchestrator pools for batch jobs so I have left in this PR.
> 
> I retested all batch pipelines after splitting PR.

Where is the other PR? Could you share the link?

leszko (approved):
LGTM

leszko:
@rickstaa do you want to review or should I merge?

server/ai_http.go
```diff
@@ -480,7 +480,7 @@ func handleAIRequest(ctx context.Context, w http.ResponseWriter, r *http.Request
 
 	// Check if there is capacity for the request.
 	if !orch.CheckAICapacity(pipeline, modelID) {
-		respondWithError(w, fmt.Sprintf("Insufficient capacity for pipeline=%v modelID=%v", pipeline, modelID), http.StatusServiceUnavailable)
+		respondWithError(w, fmt.Sprintf("insufficient capacity for pipeline=%v modelID=%v", pipeline, modelID), http.StatusServiceUnavailable)
```
rickstaa: @ad-astra-video Just a minor point—why was the capitalization removed here? I can see why it looks nicer in the log line, but all other log lines use the capitalized form. If you prefer the lowercase version (e.g., `err=insufficient` vs. `err=Insufficient`), we should make it consistent across the codebase. If needed, that could be addressed in a separate pull request.

server/ai_session.go
rickstaa: @ad-astra-video I'm not sure why we need this logic. In most cases, `sel.node.OrchestratorPool.Size()` will be around 100, meaning the `initialPoolSize` will typically be used. As a result, if we lose 50% of the discovered orchestrators, a refresh will be triggered. 

I assume you kept this logic to maintain Yondon's original approach, which refreshed when fewer than 50% of the orchestrators were found during discovery. However, since a failed discovery request would trigger a refresh anyway, is this check still necessary? 

I'm fine with keeping this in for now since it was already in there and we need to do a bigger overhaul anyway.

rickstaa (approved):
@ad-astra-video, I made some small code improvements in https://github.com/ad-astra-video/go-livepeer/pull/32. Once those are merged, we're good to go. I thoroughly battle-tested it, trying to break it, but couldn't. 🚀

server/ai_session.go
```diff
@@ -272,20 +296,31 @@ func (sel *AISessionSelector) Remove(sess *AISession) {
...
 	var coldSessions []*BroadcastSession
+
 	for _, sess := range sessions {
 		// If the constraints are missing for this capability skip this session
 		constraints, ok := sess.OrchestratorInfo.Capabilities.Constraints.PerCapability[uint32(sel.cap)]
 		if !ok {
 			continue
 		}
 
+		// this should not be needed, the GetOrchestrators checks for suspension but was seeing orchestrators get back into
```
rickstaa: @ad-astra-video before merging can you add the reason you found concisely to the Todo so developers know why it it's here 🙏.

server/ai_session.go
ad-astra-video: This check is to prevent the refresh of orchestrator pools when the pool is less than 50% of the `initialPoolSize`.  Without this check the refresh would happen for most models all the time because many selectors would not be served by at least 50 Orchestrators.

We discussed on call and agreed to leave it in for now to limit refreshes.  The selection algo will be re-written for AI to also include live-video-to-video pipeline in next few months.

server/ai_http.go
ad-astra-video: This was reverted and the test for retryable errors was improved to not be case sensitive.

server/ai_session.go
ad-astra-video: Updated

PR #3050, opened 2024-05-09
Create option to filter Os by min livepeer version used

leszko:
Add the ability to specify by Broadcaster the minimum acceptable go-livepeer version from Os. It works in the following manner:
1. Broadcaster specifies a flag `-orchMinLivepeerVersion 4.3.1`
2. During the discovery/selection process, B excludes all Os with the version below `4.3.1`
3. In the case of O/T topology, the version of O is taken into consideration, then while serving a segment, O will reject Ts that didn't upgrade.


mjh1 (approved):

server/broadcast.go
```diff
@@ -448,6 +448,7 @@ func (bsm *BroadcastSessionsManager) shouldSkipVerification(sessions []*Broadcas
 }
 
 func NewSessionManager(ctx context.Context, node *core.LivepeerNode, params *core.StreamParameters, sel BroadcastSessionsSelectorFactory) *BroadcastSessionsManager {
+	params.Capabilities.SetMinVersionConstraint(node.Capabilities.MinVersionConstraint())
```
thomshutt: Can `node.Capabilities` ever be null?

thomshutt (approved):

server/broadcast.go
leszko: I _think_ that no. But I'll add the IF just in case.

Issue #3057, opened 2024-05-16
Orchestrator crash, possibly related to pricing


eliteprox:
### Describe the bug

Runtime error occurred on production orchestrator:

```
I0516 16:39:30.673213       1 block_watcher.go:454] Polling blocks from=211907963 to=211907982
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x2729f64]

goroutine 723 [running]:
math/big.(*Rat).Num(...)
        /usr/local/go/src/math/big/rat.go:409
github.com/livepeer/go-livepeer/core.(*orchestrator).priceInfo(0xc000ea9090, {0x68, 0xd6, 0xff, 0x39, 0x38, 0xff, 0x63, 0xd2, 0xdf, ...}, ...)
        /src/core/orchestrator.go:374 +0x204
github.com/livepeer/go-livepeer/core.(*orchestrator).PriceInfo(0xc0009937b8?, {0x68, 0xd6, 0xff, 0x39, 0x38, 0xff, 0x63, 0xd2, 0xdf, ...}, ...)
        /src/core/orchestrator.go:287 +0x5a
github.com/livepeer/go-livepeer/server.getPriceInfo({0xb22b40?, 0xc000ea9090?}, {0x68, 0xd6, 0xff, 0x39, 0x38, 0xff, 0x63, 0xd2, ...}, ...)
        /src/server/rpc.go:363 +0xe4
github.com/livepeer/go-livepeer/server.orchestratorInfoWithCaps({0xb22b40, 0xc000ea9090}, {0x68, 0xd6, 0xff, 0x39, 0x38, 0xff, 0x63, 0xd2, ...}, ...)
        /src/server/rpc.go:374 +0xaa
github.com/livepeer/go-livepeer/server.orchestratorInfo(...)
        /src/server/rpc.go:367
github.com/livepeer/go-livepeer/server.getOrchestrator({0xb22b40, 0xc000ea9090}, 0xc000877440)
        /src/server/rpc.go:341 +0x2b2
github.com/livepeer/go-livepeer/server.(*lphttp).GetOrchestrator(0x5138c0?, {0xc000877440?, 0x18d03c6?}, 0x0?)
        /src/server/rpc.go:185 +0x25
[...]

PR #3047, opened 2024-05-08
Remove -pricePerUnit requirement for -aiWorker flag

eliteprox:
**What does this pull request do? Explain your changes. (required)**

This change removes the requirement for `-pricePerUnit` flag when both `-orchestrator` and `-aiWorker` flag are used. 

**Specific updates (required)**

- Updates `starter.go` to check for `-aiWorker` flag before requiring `-pricePerUnit`

**How did you test each of these updates (required)**

- Verified go-livepeer starts without error when`-orchestrator`, `-aiWorker` flag are used without `-pricePerUnit`

- Verified go-livepeer still requires `-pricePerUnit` when `-orchestrator` flag is used without `-aiWorker`

- Verified `-pricePerUnit` still sets the default price for transcoding when `-orchestrator` is used without `-aiWorker`

**Does this pull request close any open issues?**

LIV-181 Remove the need for the pricePerUnit flag

**Checklist:**

- [x] Read the [contribution guide](./CONTRIBUTING.md)
- [x] `make` runs successfully
- [x] All tests in `./test.sh` pass
- [] README and other documentation updated
- [] [Pending changelog](./CHANGELOG_PENDING.md) updated

PR #3059, opened 2024-05-17
Fix runtime error in aiWorker when pricePerUnit is unset

eliteprox:

Fixes a fatal runtime error when `-pricePerUnit` is not set, which is allowed with `-aiWorker` flag in #3047

- Checks if basePrice is nil before comparing if it is equal to zero. 
- I also added a check if transcodePrice is nil before using it, just in case that condition occurs. Though it shouldn't happen because of startup checks that require `-pricePerUnit` when `-aiWorker` is not set

- Verified that ai jobs process at the correct price from aiModels.json
- Verified `-pricePerBroadcaster` prices are set correctly for transcoding work
- Verified `-pricePerUnit` price is used for transcode jobs when `-pricePerBroadcaster` is unset for the broadcasting address
- Verified that `-autoAdjustPrice` still works correctly in all cases

#3057

**Checklist:**

### DIFF
--- a/core/orchestrator.go
+++ b/core/orchestrator.go
@@ -334,7 +334,7 @@ func (orch *orchestrator) priceInfo(sender ethcommon.Address, manifestID Manifes
 
 	basePrice := big.NewRat(0, 1)
 	if caps == nil {
-		if transcodePrice == nil {
+		if transcodePrice != nil {
 			basePrice = transcodePrice
 		}
 	} else {
--- a/core/orchestrator.go
+++ b/core/orchestrator.go
@@ -332,12 +332,13 @@ func (orch *orchestrator) priceInfo(sender ethcommon.Address, manifestID Manifes
 		transcodePrice = orch.node.GetBasePrice("default")
 	}
 
-	var basePrice *big.Rat
+	basePrice := big.NewRat(0, 1)
 	if caps == nil {
-		basePrice = transcodePrice
+		if transcodePrice == nil {
+			basePrice = transcodePrice
+		}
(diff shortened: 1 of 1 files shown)


core/orchestrator.go
```diff
@@ -359,7 +359,7 @@ func (orch *orchestrator) priceInfo(sender ethcommon.Address, manifestID Manifes
 
 		// If no priced capabilities were signaled by the broadcaster assume that they are requesting
 		// transcoding and set the base price to the transcode price
-		if basePrice.Cmp(big.NewRat(0, 1)) == 0 {
+		if transcodePrice != nil && (basePrice == nil || basePrice.Cmp(big.NewRat(0, 1)) == 0) {
```
rickstaa: @eliteprox, as you mentioned earlier, `basePrice` will never be `nil` because of the following line in the `orchestrator.go` file:

[Orchestrator code](https://github.com/livepeer/go-livepeer/blob/a1b9f33d7261b798d0d33685992a4f93ec4d3473/core/orchestrator.go#L340).

Given this, I believe the condition `basePrice == nil` should be removed. This will ensure that the `basePrice` is correctly set to the `transcodePrice` when `basePrice` is `0`.

Here is the updated code snippet:

```suggestion
		if transcodePrice != nil && basePrice.Cmp(big.NewRat(0, 1)) == 0 {
```

This change ensures that the `basePrice` will be set to `transcodePrice` when it is zero, reflecting the intended behavior more accurately.

rickstaa (changes requested):

PR #3063, opened 2024-05-20
fix(ai): fix cli prices nil error

rickstaa:

This commit ensures that the livepeer_cli does not throw a `nil` error when it tries to retrieve the orchestrator base price.

- Ensures that the CLI can handle `nil` price being returned.

- Ran the client.

**Checklist:**


eliteprox:
Tested and this is working

eliteprox (approved):
Tested and this is working. Later on we can add a way to list the model prices on livepeer_cli

PR #3065, opened 2024-05-28
Return appropriate errors when auth fails, not just 5xx

thomshutt:
Have the Broadcaster return appropriate HTTP status codes on auth failure paths, to make monitoring and diagnosis of issues easier.

- Bring in a new version of LPMS that includes an error return value in the auth method
- Handle this value and convert to HTTP statuses where appropriate

- Ran unit tests

- No

**Checklist:**

### DIFF
--- a/server/mediaserver.go
+++ b/server/mediaserver.go
@@ -259,8 +259,7 @@ func createRTMPStreamIDHandler(_ctx context.Context, s *LivepeerServer, webhookR
 		// do not replace captured _ctx variable
 		ctx := clog.AddNonce(_ctx, nonce)
 		if resp, err = authenticateStream(AuthWebhookURL, url.String()); err != nil {
-			errMsg := fmt.Sprintf("Forbidden: Authentication denied for streamID url=%s err=%q", url.String(), err)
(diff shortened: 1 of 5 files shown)


server/mediaserver.go
```diff
@@ -820,10 +826,15 @@ func (s *LivepeerServer) HandlePush(w http.ResponseWriter, r *http.Request) {
 
 	// Check for presence and register if a fresh cxn
 	if !exists {
-		appData := (createRTMPStreamIDHandler(ctx, s, authHeaderConfig))(r.URL)
-		if appData == nil {
-			errorOut(http.StatusInternalServerError, "Could not create stream ID: url=%s", r.URL)
-			return
+		appData, err := (createRTMPStreamIDHandler(ctx, s, authHeaderConfig))(r.URL)
+		if err != nil {
+			if strings.HasPrefix(err.Error(), "Forbidden") {
```
leszko: Consider returning a const error like below instead of distinguishing errors by checking string message.

```golang
var (
    errForbidden = errors.New("Forbidden")
)
```

leszko (approved):

PR #3076, opened 2024-06-10
Add logging to selection_algorithm.go

leszko:
Related to https://linear.app/livepeer/issue/ENG-2046/investigate-suspicious-orchestrator-prices

### DIFF
--- a/server/selection_algorithm.go
+++ b/server/selection_algorithm.go
@@ -49,7 +49,7 @@ func (sa ProbabilitySelectionAlgorithm) filterByPerfScore(addrs []ethcommon.Addr
 	if len(res) == 0 {
 		// If no orchestrators pass the perf filter, return all Orchestrators.
 		// That may mean some issues with the PerfScore service.
-		glog.Warning("No Orchestrators passed min performance score filter, not using the filter, minPerfScore=%v, scores=%v", sa.MinPerfScore, scores)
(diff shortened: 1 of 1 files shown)


victorges (approved):
LGTM

server/selection_algorithm.go
```diff
@@ -49,7 +49,7 @@ func (sa ProbabilitySelectionAlgorithm) filterByPerfScore(addrs []ethcommon.Addr
 	if len(res) == 0 {
 		// If no orchestrators pass the perf filter, return all Orchestrators.
 		// That may mean some issues with the PerfScore service.
-		glog.Warning("No Orchestrators passed min performance score filter, not using the filter")
+		glog.Warning("No Orchestrators passed min performance score filter, not using the filter, minPerfScore=%v, scores=%v", sa.MinPerfScore, scores)
```
victorges: WDYT of adding `addrs` or probably just `len(addrs)` to these logs? Should help debug the filtering pipeline. Or even we could always have this log, including `len(addrs)` and `len(res)`? (might wanna omit the `scores` if we do that tho as that's big)

leszko: Yes, good idea, I'll add both `len(addrs)` and `addrs`. This should help us investigate the issue.

We don't need `len(res)`, it's always `0` at this point.

PR #3078, opened 2024-06-12
Add audio-to-text pipeline

eliteprox:

Adds the new `/audio-to-text` pipeline to go-livepeer, supporting the `openai/whisper-large-v3` model.

File formats supported are mp3, m4a, mp4, webm, and flac

This change requires https://github.com/livepeer/ai-worker/pull/103 and https://github.com/livepeer/lpms/pull/407

- Refactors `handleAIRequest` and `processAIRequest` to support new response types like TextResponse
- Adds `/audio-to-text` endpoint to `ai_mediaserver.go`
- Pricing fixed to one pixel per millisecond
- Uses `ffmpeg.GetCodecInfo` to calculate duration and requires the lpms pull request above

- Tested with rich vocal audio up to 4 hours long. Regression tested other pipelines to ensure refactoring cause any issues.
- Tested with all supported file formats and unsupported ones

curl request example:
```
curl --request POST   --url http://dev.eliteencoder.net:8937/audio-to-text --header 'Content-Type: multipart/form-data'   --form 'audio=@Recording.mp3'   --form 'model_id=openai/whisper-large-v3'   --form seed=123
```

LIV-429
LIV-289

**Checklist:**


eliteprox:
Added error handlers to respond with "400 bad request" when duration cannot be calculated due to unsupported file format or file corruption. This prevents invalid jobs from being sent to the network.

common/util.go
```diff
@@ -530,3 +538,130 @@ func ParseEthAddr(strJsonKey string) (string, error) {
 	}
 	return "", errors.New("Error parsing address from keyfile")
 }
+
+// determines the duration of an mp3 audio file by reading the frames
+var ErrUnsupportedFormat = errors.New("Unsupported audio file format. Supported formats: mp3, wav, mp4, m4a, webm, flac")
+var ErrorCalculatingDuration = errors.New("Error calculating duration")
+
+func CalculateAudioDuration(audio types.File) (int64, error) {
```
emranemran, line 538: Would it make sense to just use ffmpeg to get the audio track's duration instead of using multiple methods based on fileExt? We do similar things for the video track in catalyst-api. You wouldn't need to do everything we do in catalyst-api [1] but just a call to the the go ffmpeg package might suffice. It would be less code to maintain. 

[1] https://github.com/livepeer/catalyst-api/blob/main/video/probe.go#L93

server/ai_process.go
```diff
@@ -421,14 +507,25 @@ func processAIRequest(ctx context.Context, params aiRequestParams, req interface
...
+			return submitSpeechToText(ctx, params, sess, v)
+		}
+	// Add more cases as needed...
 	default:
-		return nil, errors.New("unknown AI request type")
+		return nil, fmt.Errorf("unsupported request type %T", req)
 	}
 
-	var resp *worker.ImageResponse
+	//var resp *worker.ImageResponse
```
emranemran: minor: remove comment?

emranemran:
Overall LGTM. I would recommend using ffmpeg for audio track processing (like calculating durations) and *only* accepting audio input instead of videos -- ffmpeg can help with that as well and we could potentially commonize the `Probe` package I linked in catalyst-api in the comments. Also, we need to understand what the file length limits are like.

common/util.go
ad-astra-video: This requires ffprobe installed right? I like this idea but it does add a dependency to using go-livepeer binaries.

The LPMS way would be to use this update to [GetCodecInfo](https://github.com/livepeer/lpms/pull/376) to get the duration.  With the ffmpeg upgrade to 7.0, should be covered on codec support as well.

common/util.go
```diff
@@ -11,21 +12,28 @@ import (
...
 	"sort"
 	"strconv"
 	"strings"
 	"testing"
 	"time"
 
+	"github.com/abema/go-mp4"
+	"github.com/ebml-go/webm"
 	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/go-audio/wav"
```
leszko: Nit: Have you considered using https://github.com/gopxl/beep? Seems to convert more formats + have more GitHub stars.

common/util.go
```diff
@@ -530,3 +538,130 @@ func ParseEthAddr(strJsonKey string) (string, error) {
...
+		return GetDuration_FLAC(read)
+	case ".mp3":
+		return GetDuration_MP3(read)
+
+	default:
+		return 0, ErrUnsupportedFormat
+	}
+}
+
+func GetDuration_WAV(audio io.ReadCloser) (int64, error) {
```
leszko: I think that all these functions would benefit from unit tests.

server/ai_process.go
```diff
@@ -382,10 +400,77 @@ func submitUpscale(ctx context.Context, params aiRequestParams, sess *AISession,
...
+		balUpdate.Status = ReceivedChange
+	}
+
+	var res worker.TextResponse
+	if err := json.Unmarshal(data, &res); err != nil {
+		return nil, err
+	}
+
+	// TODO: Refine this rough estimate in future iterations
+	sess.LatencyScore = took.Seconds() / float64(outPixels)
```
leszko: I'd consider extracting all these "temp" conversions hacks (latency score, outPixels) to one place. Otherwise it may be hard later to reason about them.

leszko (approved):
Added some comments, other than that LGTM

server/ai_process.go
eliteprox: Yes, cleaned up

eliteprox:
> Overall LGTM. I would recommend using ffmpeg for audio track processing (like calculating durations) and _only_ accepting audio input instead of videos -- ffmpeg can help with that as well and we could potentially commonize the `Probe` package I linked in catalyst-api in the comments. Also, we need to understand what the file length limits are like.

@emranemran Based I have updated the duration calculation to use ffmpeg. It required a change to [install_ffmpeg.sh](https://github.com/livepeer/go-livepeer/pull/3078/files#diff-4ae778054809274731b9da0c6a5a869c0bd214e92f954a5c9c39181748c2f175) and lpms https://github.com/livepeer/lpms/pull/407

common/util.go
eliteprox: @emranemran We have updated the duration calculation to use ffmpeg. It required a change to [install_ffmpeg.sh](https://github.com/livepeer/go-livepeer/pull/3078/files#diff-4ae778054809274731b9da0c6a5a869c0bd214e92f954a5c9c39181748c2f175) and lpms https://github.com/livepeer/lpms/pull/407

common/util.go
eliteprox: We're going to drop support for WAV right now as we now have all other formats working through lpms/ffmpeg.

common/util.go
eliteprox: Removed functions, writing tests is in the backlog

server/ai_process.go
eliteprox: Added a linear issue to track this https://linear.app/livepeer-ai/issue/AI-505/refactor-latency-scoring-into-common-function

go.sum
```diff
@@ -182,6 +182,8 @@ github.com/dop251/goja_nodejs v0.0.0-20210225215109-d91c329300e7/go.mod h1:hn7BA
 github.com/dop251/goja_nodejs v0.0.0-20211022123610-8dd9abb0616d/go.mod h1:DngW8aVqWbuLRMHItjPUyqdj+HWPvnQe8V8y1nDpIbM=
 github.com/dustin/go-humanize v1.0.1 h1:GzkhY7T5VNhEkwH0PVJgjz+fX1rhBrR7pRT3mDkpeCY=
 github.com/dustin/go-humanize v1.0.1/go.mod h1:Mu1zIs6XwVuF/gI1OepvI0qD18qycQx+mFykh5fBlto=
+github.com/eliteprox/ai-worker v0.0.0-20240705062703-0908b518eb12 h1:LCnOJCD97i2jmR7gY5gJieYLS+01XHnbWkCpmHC1t28=
```
rickstaa: Quick note to replace when worker code is merged.

go.mod
```diff
@@ -205,3 +206,5 @@ require (
 	lukechampine.com/blake3 v1.2.1 // indirect
 	rsc.io/tmplfunc v0.0.3 // indirect
 )
+
+replace github.com/livepeer/ai-worker => github.com/eliteprox/ai-worker v0.0.0-20240705062703-0908b518eb12
```
rickstaa: Remove when releasing.

go.sum
```diff
@@ -531,16 +533,14 @@ github.com/libp2p/go-netroute v0.2.0 h1:0FpsbsvuSnAhXFnCY0VLFbJOzaK0VnP0r1QT/o4n
...
-github.com/livepeer/ai-worker v0.0.8/go.mod h1:Xlnb0nFG2VsGeMG9hZmReVQXeFt0Dv28ODiUT2ooyLE=
 github.com/livepeer/go-tools v0.3.6-0.20240130205227-92479de8531b h1:VQcnrqtCA2UROp7q8ljkh2XA/u0KRgVv0S1xoUvOweE=
 github.com/livepeer/go-tools v0.3.6-0.20240130205227-92479de8531b/go.mod h1:hwJ5DKhl+pTanFWl+EUpw1H7ukPO/H+MFpgA7jjshzw=
 github.com/livepeer/joy4 v0.1.2-0.20191121080656-b2fea45cbded h1:ZQlvR5RB4nfT+cOQee+WqmaDOgGtP2oDMhcVvR4L0yA=
 github.com/livepeer/joy4 v0.1.2-0.20191121080656-b2fea45cbded/go.mod h1:xkDdm+akniYxVT9KW1Y2Y7Hso6aW+rZObz3nrA9yTHw=
 github.com/livepeer/livepeer-data v0.7.5-0.20231004073737-06f1f383fb18 h1:4oH3NqV0NvcdS44Ld3zK2tO8IUiNozIggm74yobQeZg=
 github.com/livepeer/livepeer-data v0.7.5-0.20231004073737-06f1f383fb18/go.mod h1:Jpf4jHK+fbWioBHRDRM1WadNT1qmY27g2YicTdO0Rtc=
-github.com/livepeer/lpms v0.0.0-20240120150405-de94555cdc69 h1:4A6geMb+HfxBBfaS24t8R3ddpEDfWbpx7NTQZMt5Fp4=
-github.com/livepeer/lpms v0.0.0-20240120150405-de94555cdc69/go.mod h1:Hr/JhxxPDipOVd4ZrGYWrdJfpVF8/SEI0nNr2ctAlkM=
+github.com/livepeer/lpms v0.0.0-20240711175220-227325841434 h1:E7PKN6q/jMLapEV+eEwlwv87Xe5zacaVhvZ8T6AJR3c=
```
rickstaa, line 542: Ensure versioning is correct.

core/orchestrator.go
```diff
@@ -945,6 +949,16 @@ func (n *LivepeerNode) upscale(ctx context.Context, req worker.UpscaleMultipartR
 	return n.AIWorker.Upscale(ctx, req)
 }
 
+func (n *LivepeerNode) AudioToText(ctx context.Context, req worker.AudioToTextMultipartRequestBody) (*worker.TextResponse, error) {
```
rickstaa, line 952: NOTE TO SELF: Check if this function needs to be exported.

common/util.go
rickstaa: Great! thanks @eliteprox.

core/orchestrator.go
```diff
@@ -945,6 +949,16 @@ func (n *LivepeerNode) upscale(ctx context.Context, req worker.UpscaleMultipartR
...
+
+	resp, err := n.AIWorker.AudioToText(ctx, req)
```
rickstaa: @eliteprox based on the code in https://github.com/livepeer/ai-worker/blob/0908b518eb1203617a3a68625d3326679a636439/worker/worker.go#L253-L305 we could simply forward the return values without performing the extra error check right 🤔?

common/util.go
```diff
@@ -530,3 +531,28 @@ func ParseEthAddr(strJsonKey string) (string, error) {
 	}
 	return "", errors.New("Error parsing address from keyfile")
 }
+
+// determines the duration of an mp3 audio file by reading the frames
+var ErrUnsupportedFormat = errors.New("Unsupported audio file format")
```
rickstaa: @eliteprox. As also stated in https://github.com/eliteprox/go-livepeer/pull/4/files#r1677148579 this is a very useful error which should be implemented but currently is not used. I tried inputting the following unsupported audio OGG filetype:

[25eb6f27.zip](https://github.com/user-attachments/files/16227789/25eb6f27.zip)

But it threw a audio duration calculation error. Can we implement a audio file type unsupported error?

rickstaa:
@eliteprox looks like our pipeline fails to detect the duration of the following file:

[speech.zip](https://github.com/user-attachments/files/16227820/speech.zip)

Do you maybe know why 🤔?

common/util.go
eliteprox: WAV support re-added via ffmpeg

core/orchestrator.go
eliteprox: Removed unnecessary logic

eliteprox:
> @eliteprox looks like our pipeline fails to detect the duration of the following file:
> 
> [speech.zip](https://github.com/user-attachments/files/16227820/speech.zip)
> 
> Do you maybe know why 🤔?

This one appears to be a concatenated file and ffmpeg has issues calculating the duration in this case. The recommended solution is to re-encode the input to a consistent output format like flac. This can be combined with the effort to send audio-only to the ai-worker to optimize the pipeline.

rickstaa (approved):
LGTM 🚀.

PR #3086, opened 2024-06-26
selection: Clear known sessions if none of them has good enough latency score

leszko:
fix https://linear.app/livepeer/issue/ENG-2051/investigate-warnings-no-orchestrators-passed-max-price-filter

Problem: We never cleaned the known sessions, even if all of them had too low latency score to use; example of what sometimes happens:

1) Broadcaster accumulated 8 known sessions, but all had too low latency score
2) Session refresh was not triggered because we had a lot of known session in the pool
3) At the same time, we didn't have sessions in unknown session pool
4) Selection checked that it cannot use the known session and tried to select from unknown sessions (but the pool was empty or there were no sessions fulfilling the condition of max price or perf score)


server/selection.go
```diff
@@ -135,12 +135,15 @@ func (s *MinLSSelector) Select(ctx context.Context) *BroadcastSession {
...
+	lowestLatencyScoreKnownSession := heap.Pop(s.knownSessions).(*BroadcastSession)
+	if lowestLatencyScoreKnownSession.LatencyScore <= s.minLS {
+		// known session has good enough latency score, use it
+		return lowestLatencyScoreKnownSession
 	}
 
-	return heap.Pop(s.knownSessions).(*BroadcastSession)
+	// known session does not have good enough latency score, clear the heap and use unknown session
+	s.knownSessions = &sessHeap{}
+	return s.selectUnknownSession(ctx)
```
victorges, line 146: Will this repopulate the known sessions heap as well?

victorges (approved):
Makes sense to me on the code snippet that was changed, but FTR I'm not super familiar with that entire session management logic.

LGTM

server/selection.go
```diff
@@ -135,12 +135,15 @@ func (s *MinLSSelector) Select(ctx context.Context) *BroadcastSession {
 		return s.selectUnknownSession(ctx)
 	}
 
-	minSess := sess.(*BroadcastSession)
-	if minSess.LatencyScore > s.minLS && len(s.unknownSessions) > 0 {
-		return s.selectUnknownSession(ctx)
+	lowestLatencyScoreKnownSession := heap.Pop(s.knownSessions).(*BroadcastSession)
```
thomshutt, line 138: What if there's another session on the heap? I know this is how it was before, so just trying to understand the logic here

leszko: I think it works like this: 
- take the first session from the heap
- if no more sessions, then refresh the sessions

So, it's no a problem if there are more than 1 sessions on the heap, we just pop the fist one.

thomshutt: Ah right, I understand now - either the lowest latency known session is good enough or not, which would imply that none of the others are either

thomshutt (approved):

server/selection.go
leszko: No, the `knownSession` is populated when a session is selected from unknown, then used, then completed. Then, we store it as kind-of "hi, I already used that session, it was good, so why not to just re-use it instead of doing the whole discovery process".

PR #3087, opened 2024-06-26
feat(ai): add AI gateway metrics

rickstaa:

This pull request introduces new Gateway AI metrics to the `ai-video` branch:
- **Ticket_value_sent**: The value of the AI tickets that are send to the Orchestrators.
- **Tickets_send**: The total amount of tickets send to the Orchestrators.
- **ai_models_requested**: Tracks the number of requests per capability and model.
- **ai_request_latency_score**: Measures latency scores per model request per orchestrator.
- **ai_request_price**: Records the price paid per unit for each model request.
- **ai_request_errors**: Logs AI request errors per orchestrator. If no orchestrator is specified, this indicates a general request scenario where no orchestrators were found.

- Updates `census.go` to include the new Gateway metrics.
- Updates `ai_process` to log these metrics.

I set up both an on-chain and off-chain gateway to validate the metrics. I verified their visibility at `http://localhost:5935/metrics` and ensured they were correctly visualized in Grafana.

This implements the functionality outlined in https://livepeer-ai.productlane.com/roadmap?id=58c9cd5d-1c64-4fb3-b8d0-a7e20b7865a2.

**Checklist:**

#### How to test

1. Check out this pull request.
2. Spin up an on-chain gateway with attached orchestrators.
3. Clone the repository https://github.com/rickstaa/livepeer-monitor-test.
4. Execute the Dockerfile in that repository to launch Prometheus and Grafana servers.
5. Navigate to `http://localhost:5935/metrics` to view the new AI Gateway metrics.
6. Visit `http://localhost:3000` to inspect these metrics in Grafana.

### DIFF
--- a/monitor/census.go
+++ b/monitor/census.go
@@ -357,8 +357,8 @@ func InitCensus(nodeType NodeType, version string) {
 	// Metrics for AI jobs
 	census.mAIModelsRequested = stats.Int64("ai_models_requested", "Number of AI models requested over time", "tot")
 	census.mAIRequestLatencyScore = stats.Float64("ai_request_latency_score", "AI request latency score, based on smallest pipeline unit", "")
-	census.mAIRequestPrice = stats.Float64("ai_request_price", "Price paid per AI request unit", "")
-	census.mAIRequestError = stats.Int64("ai_request_errors", "Errors when processing AI requests", "tot")
+	census.mAIRequestPrice = stats.Float64("ai_request_price", "AI request price per unit, based on smallest pipeline unit", "")
+	census.mAIRequestError = stats.Int64("ai_request_errors", "Errors during AI request processing", "tot")
 
 	glog.Infof("Compiler: %s Arch %s OS %s Go version %s", runtime.Compiler, runtime.GOARCH, runtime.GOOS, runtime.Version())
 	glog.Infof("Livepeer version: %s", version)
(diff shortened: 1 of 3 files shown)


server/ai_process.go
```diff
@@ -443,7 +538,11 @@ func processAIRequest(ctx context.Context, params aiRequestParams, req interface
 	for {
 		select {
 		case <-cctx.Done():
-			return nil, &ServiceUnavailableError{err: fmt.Errorf("no orchestrators available within %v timeout", processingRetryTimeout)}
+			err := fmt.Errorf("no orchestrators available within %v timeout", processingRetryTimeout)
```
rickstaa, line 666: @mikezupper, @eliteprox We could also replace this error and the one on line 572 with a more general No capacity message if that makes more sense. My goal is to ensure we can identify when pipelines fail due to insufficient Orchestrators.

monitor/census.go
```diff
@@ -855,6 +876,36 @@ func InitCensus(nodeType NodeType, version string) {
...
+			Measure:     census.mAIRequestLatencyScore,
+			Description: "AI orchestrator request latency score",
+			TagKeys:     append([]tag.Key{census.kPipeline, census.kModelName}, baseTagsWithOrchInfo...),
+			Aggregation: view.LastValue(),
+		},
+		{
+			Name:        "ai_request_price",
+			Measure:     census.mAIRequestPrice,
+			Description: "AI request price per unit",
+			TagKeys:     append([]tag.Key{census.kPipeline, census.kModelName}, baseTags...),
```
eliteprox, line 899: It may help to track the price by orchestrator. What do you think?
```suggestion
			TagKeys:     append([]tag.Key{census.kPipeline, census.kModelName}, baseTagsWithOrchInfo...),
```

server/ai_process.go
rickstaa: @mikezupper, @eliteprox, @ad-astra-video we could also split the capacity error in a separate metric but I didn't want to  create unnecessary metrics.

monitor/census.go
rickstaa: @eliteprox I first had that however from the Ecosystem side I would simply want to know the average price per pipeline which I can calculate based on this value. @ad-astra-video, @mikezupper what do you think. Would you use the info per orchestrator?

rickstaa: @eliteprox let's put it right back -> https://github.com/livepeer/go-livepeer/pull/3087/commits/bd5b047694a8b25643f8865a436761a4f0a8c7f6. @ad-astra-video feel free to let me know if you think this will cause to much metrics to be generated 👍🏻.

server/ai_process.go
```diff
@@ -124,6 +135,14 @@ func submitTextToImage(ctx context.Context, params aiRequestParams, sess *AISess
 	}
 	sess.LatencyScore = took.Seconds() / float64(outPixels) / (numImages * numInferenceSteps)
 
+	if monitor.Enabled {
+		pricePerUnit := 0.0
```
rickstaa: This calculation is not correct. Will push a fix soon.

rickstaa: Solved.

monitor/census.go
eliteprox: On second thought, I don't think there is a "normal" use case for having the orchestrator specific pricing. I think it would be helpful as part of a network reporting tool similar to stronk.rocks. 

I agree that it could generate too much data. One option would be to implement something similar to the `-PerStreamMetrics` flag, setting the baseflags conditionally, for example:

https://github.com/livepeer/go-livepeer/blob/b7181d425c64095da04a5c7e277173885afc0ab5/monitor/census.go#L383-L387

rickstaa: I did add the orchestrator label to the pricing as you suggested (see https://github.com/livepeer/go-livepeer/blob/a92e9df2a6f32c2eb6a72ef3d162daf9944887f5/monitor/census.go#L899). However wasn't sure if this will lead to to much data. Curious to hear what @ad-astra-video thinks 🤔. We could add a `-perOrchAIMetrics` flag.

pwilczynskiclearcode:
https://github.com/livepeer/go-livepeer/issues/3185

PR #3095, opened 2024-07-10
Add `/healthz` endpoint

leszko:
Add a simple healthcheck endpoint with the intention to use it for the kubernetes `livenessProbe` / `readinessProbe`.

We already have `/status` endpoint, but it:
- returns a ton of data (we don't need it for k8 health checks)
- does some calculations (we don't want to do a lot of calculations with each healthcheck)

### DIFF
--- a/server/handlers.go
+++ b/server/handlers.go
@@ -29,6 +29,12 @@ import (
 const MainnetChainId = 1
 const RinkebyChainId = 4
 
+func (s *LivepeerServer) healthzHandler() http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		respondOk(w, nil)
+	})
+}
+
 // Status
 func (s *LivepeerServer) statusHandler() http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
--- a/server/mediaserver.go
+++ b/server/mediaserver.go
@@ -199,6 +199,9 @@ func (s *LivepeerServer) StartMediaServer(ctx context.Context, httpAddr string) 
 	// Store ctx to later use as cancel signal for watchdog goroutine
 	s.context = ctx
 
+	// health endpoint
+	s.HTTPMux.Handle("/healthz", s.healthzHandler())
+
 	//LPMS handlers for handling RTMP video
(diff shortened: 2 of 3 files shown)


pwilczynskiclearcode (approved):

server/handlers.go
```diff
@@ -29,6 +29,12 @@ import (
 const MainnetChainId = 1
 const RinkebyChainId = 4
 
+func (s *LivepeerServer) healthzHandler() http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
victorges, line 33: I think we shouldn't use this for the `startupProbe` though, otherwise we'll start sending traffic to the pod during that state where it's just 5xx'ing.

Unless there's no correlation between that time the existing healthcheck fails and the service being ready to serve, and the 5xx are just nginx failing to route to it simply cause k8s _thinks_ it's not ready.

IOW do you know why that existing healthcheck doesn't work for around a minute before the B does get ready? Cause from the outside it looks like it's actually checking something useful haha

victorges (approved):
LGTM

server/handlers.go
leszko: Well... `/status` does not check anything useful. It also does not return any status other than `200`. The reason that it's not available right away is because there are a number of things happening sequentially _before_ starting the http server (exposing endpoints). In that sense, `/healthz` is not a regression. It's just the same.

Saying that, I'm not sure why we're seeing 5xx during deployments. My guess is that we have two different web services starting in parallel (http video: serving videos, http cli: service CLI commands) and maybe one is available and the other not?! This made me think that `/healthz` endpoint should be using the same port as where we serve the video. I moved it in these 2 commits: https://github.com/livepeer/go-livepeer/pull/3095/commits/ab8e557dbf87d6f654dbb4225af4009c51cfcd79 and https://github.com/livepeer/livepeer-infra/pull/2245/commits/2828eedbd376d2e4debf4db8bf8e047cdbf7c678

PTAL: @victorges @pwilczynskiclearcode

victorges: Got it! Thanks for looking this up and explaining. Having this healthcheck on the "video port" also makes sense to me.

victorges (approved):

PR #3096, opened 2024-07-11
Update LPMS to ffmpeg 7

j0sh:
Updates LPMS to ffmpeg 7

- Update LPMS in go.mod
- Swap install_ffmpeg.sh with LPMS version

Unit tests

**Checklist:**


install_ffmpeg.sh
```diff
@@ -1,231 +1,3 @@
...
-    # If configure fails, then print the last 100 log lines for debugging and exit.
-fi
-
-if [[ ! -e "$ROOT/ffmpeg/libavcodec/libavcodec.a" || $BUILD_TAGS == *"debug-video"* ]]; then
-  cd "$ROOT/ffmpeg"
-  make -j$NPROC
-  make -j$NPROC install
-fi
+echo 'WARNING: downloading and executing lpms/install_ffmpeg.sh, use it directly in case of issues'
+curl https://raw.githubusercontent.com/livepeer/lpms/d9c78b62effdb4f5c8fc438b6033da1090d04a03/install_ffmpeg.sh | bash -s $1
```
leszko, line 3: This seems to be related to the specific commit, why not to use the `master`?

leszko (commented):
Thanks for the PR @j0sh 

My understanding is that we want to test it first with 1 O, before merging into `master`, is that correct? If that's the case, then let's wait with merging this one.

One additional note, wdyt about merging the repos: go-livepeer and lpms? I guess that would make it simpler to handle.

install_ffmpeg.sh
Tibia2000: FYI I ve been using this version on my Orchestrator for 5 days now and had no issues.

j0sh:
> My understanding is that we want to test it first with 1 O, before merging into master, is that correct? If that's the case, then let's wait with merging this one.

Sure, whatever you are comfortable with. Note that this update does not change the ffmpeg output (modulo bug fixes), so the network will remain compatible with older orchestrators. It will not be necessary to do a hard switch-over to newer orchestrators.

> One additional note, wdyt about merging the repos: go-livepeer and lpms? I guess that would make it simpler to handle.

I don't mind the separation. It's been that way since day 1 and there is a lot of git / github history baked into the LPMS repo which really helps with "git archaeology" to figure out why things are the way they are. Also it makes it easier to use LPMS outside of go-livepeer and having a separate repo helps enforce that boundary. Personally, it is kind of nice to work on LPMS by itself without having to pull in go-livepeer.

install_ffmpeg.sh
j0sh: > This seems to be related to the specific commit, why not to use the master? 

Reproducibility - the same reason tools like go mod are pinned to specific versions. Ideally we would be able to check out any go-livepeer commit and get the exact version of ffmpeg that commit depends on.

Note that we only have to update this if the LPMS install_ffmpeg.sh script changes. If LPMS is updated without any changes to `install_ffmpeg.sh`, then we don't have to actually change anything.

We could have a script or some tooling to detect mismatches (eg, must match the latest commit to `lpms/install_ffmpeg.sh` at the specific LPMS checkout which this version of go-livepeer depends on) but that can be done later if mismatches become an issue. Even better would be a more suitable build tool that is able to handle third-party dependencies, eg bazel, but that is a completely different issue.

Tibia2000:
@stronk-dev I am really curious about netint performance. Their servers are very dense when it comes to stream capacity. I think that would enable easy data-center deployments in the future.

leszko:
> > My understanding is that we want to test it first with 1 O, before merging into master, is that correct? If that's the case, then let's wait with merging this one.
> 
> Sure, whatever you are comfortable with. Note that this update does not change the ffmpeg output (modulo bug fixes), so the network will remain compatible with older orchestrators. It will not be necessary to do a hard switch-over to newer orchestrators.

Ok, then, please merge it when you do the testing in staging with @emranemran 

> 
> > One additional note, wdyt about merging the repos: go-livepeer and lpms? I guess that would make it simpler to handle.
> 
> I don't mind the separation. It's been that way since day 1 and there is a lot of git / github history baked into the LPMS repo which really helps with "git archaeology" to figure out why things are the way they are. Also it makes it easier to use LPMS outside of go-livepeer and having a separate repo helps enforce that boundary. Personally, it is kind of nice to work on LPMS by itself without having to pull in go-livepeer.

I'll open the discussion in Discord.

leszko (approved):

j0sh:
This upgrade passes the LPMS test suite without significant changes to the output, including timestamp handling, so we should be pretty confident this is backwards-compatible with older orchestrators. 

Several orchestrators seem to have been testing and have not reported issues directly related to this upgrade.

That being said, there are a couple things that might be worth testing:

- Test with blended orchestrators, eg having the broadcaster alternate between upgraded and non-upgraded orchestrators
- Visually confirm the output, preferably with a variety of inputs (RTMP, WebRTC, VOD, clips). Preferably use something like ffplay that will log any obvious issues with the transcoded segments
- Or use a separate tool to validate the stream, eg Apple mediastreamvalidator
- There are also fixes for intra-only mode (GOP=1) and FPS pass through (FPS=0) so it would be good to double check these with real streams

From Emran in Discord:

> Off the top of my head, I think testing the following workflows using the Studio website would give us coverage:
> * Livestreaming
> * VOD
> * Recordings
> * Clipping
> * Customer mobile and desktop streaming (Shih-Yu can help with this)

thomshutt:
@j0sh @leszko I think at this point, I'm confident enough that we should just go ahead and merge + release this. Any objections?

PR #3088, opened 2024-07-01
Implementation of AI Remote Worker (AI-323)

rickstaa:
This pull request introduces the AI remote worker (AI-323) to the AI stack, addressing a component intentionally omitted from the subnet POC to expedite development. This addition aims to accelerate AI job processing on the network, which is crucial for scaling the AI subnet with pools and larger orchestrators. Initial prototyping began in late April, resulting in this [tech spec](https://www.notion.so/livepeer/Engineering-Dashboard-ad42cb26278943f583d8c9a2ac7668ba?p=d8a76b01ced6469c936e7cd8d180dc92&pm=s), with core development starting in June. This implementation in this pull erquest was carried out by @ad-astra-video in collaboration with @mikezupper and the [Tenderize](https://www.tenderize.me/) team for optimal network solutions. We are grateful for this collaboration, and each party will be acknowledged for their contribution as co-authors during the merge ❤️‍🔥.

The implementation builds on the existing [remote Transcoder](https://github.com/livepeer/go-livepeer/blob/c34fce2c3432370d427059fb8e118dda6c9fae1a/core/orchestrator.go#L910) framework. While minor modifications were necessary for AI functionality, the structure remains similar to ensure maintainability and ease of future merging into the main codebase.

> [!NOTE]  
> This branch's commits were squashed and force-pushed multiple times to streamline the review process. Although some commit history was lost, additional history can be found in this pull request or at [https://github.com/ad-astra-video/go-livepeer/tree/ai-video-remoteaiworker-pr](https://github.com/ad-astra-video/go-livepeer/tree/ai-video-remoteaiworker-pr).

**Key points of this implementation**:
- Establishes a new AI remote worker architecture based on the remote Transcoder framework.
- Includes modifications specific to AI functionality while maintaining a similar structure to the remote transcoder.
- Future improvements and enhancements will be addressed in subsequent pull requests.

### Architecture

The flowchart below outlines the new AI remote worker setup, based on the **remote transcoder** setup in the regular transcoding codebase. More details are available in the [Tech Spec](https://www.notion.so/livepeer/Engineering-Dashboard-ad42cb26278943f583d8c9a2ac7668ba?p=d8a76b01ced6469c936e7cd8d180dc92&pm=s).

image

image

### How to Test the AI Remote Worker

ezgif-7-74d51104e8

To test the AI remote worker, follow these steps:

1. **Clone the Repositories:**
[...]

PR #3103, opened 2024-07-25
chore: merge master into ai-video

rickstaa:

This pull request updates the `ai-video` branch with the latest changes from the `master` branch. Initially, I attempted a rebase, but due to significant divergence between the two branches, it became too complex to resolve. Instead, I performed a simple merge and addressed all the resulting merge conflicts.

- Merged master into ai-video branch.
- Fixed merge conflicts which were mostly around new capability constraints.
- Applied the min version constraint (i.e. https://github.com/livepeer/go-livepeer/pull/3050) to the AI codebase.
- Applied the dollar pegging feature (i.e. https://github.com/livepeer/go-livepeer/pull/2981) to the AI codebase.

- Ensured that `make` worked.
- Tested off-chain and on-chain AI gateways and orchestrators.

**Checklist:**

- [ ] All tests in `./test.sh` pass - @victorges I still need to fix the tests on the `ai-video` branch.


CHANGELOG.md
```diff
@@ -1,5 +1,133 @@
 # Changelog
 
+## v0.7.6
+
+-   [#3055](https://github.com/livepeer/go-livepeer/pull/3055) census: Rename broadcaster metrics to gateway metrics
```
rickstaa, line 5: @victorges Since the AI stuff is still in alpha I currently don't use the Changelog files. If you want I can create a seperate `AI_CHANGELOG.md` file but I mostly post the bigest changes on https://livepeer-ai.productlane.com/changelog.

rickstaa:
@victorges, @leszko this requires https://github.com/livepeer/go-livepeer/pull/3111/files to be merged.

go.mod
```diff
@@ -31,27 +32,34 @@ require (
 	go.opencensus.io v0.24.0
 	go.uber.org/goleak v1.3.0
 	golang.org/x/net v0.25.0
-	google.golang.org/grpc v1.57.1
-	google.golang.org/protobuf v1.33.0
+	google.golang.org/grpc v1.65.0
```
rickstaa, line 35: @victorges I updated these two packages due to issues I encountered on my machine (Ubuntu 24). To save time, I opted for the updates instead of troubleshooting. However, we can try downgrading if the lower versions are necessary for the transcoding branch.

rickstaa:
@victorges, @leszko We still need to fix our tests and add tests for some AI-related components. I have tried to include all the necessary AI tests related to the new minVersionConstraint and dollar pegging. If I missed any cases, please let me know.

net/lp_rpc.proto
```diff
@@ -108,16 +108,25 @@ message Capabilities {
     // Capacity corresponding to each capability
     map<uint32, uint32> capacities = 3;
 
-    // Non-binary capability constraints, such as supported ranges.
+    string version = 4;
+
+    Constraints constraints = 5;
+
+    // Non-binary general constraints.
     message Constraints {
```
rickstaa, line 116: @victorges To ensure a non-breaking merge, I created a new `CapabilityConstraints` message within the `Capabilities` message. I reused the `Constraints` message you added for general constraints that apply to all capabilities, while `CapabilityConstraints` are specific to individual capabilities. If you have any suggestions for improving the naming, please let me know.

core/ai.go
```diff
@@ -23,14 +24,40 @@ type AI interface {
 	HasCapacity(pipeline, modelID string) bool
 }
 
+// Custom type to handle both string and int but store as string.
+type StringInt string
```
rickstaa: @victorges I'd love to hear your thoughts on this. I implemented this change to allow both integers and strings in the AIModel config for backward compatibility. However, this approach removes static typing and is a bit of a go anti-patern. If you think switching entirely to strings in the config file is a better solution, please let me know 👍🏻.

core/livepeernode_test.go
```diff
@@ -172,10 +172,36 @@ func TestSetAndGetBasePrice(t *testing.T) {
...
-	assert.Zero(n.priceInfo[addr2].Cmp(price2))
+	n.SetBasePrice(addr1, NewFixedPrice(price1))
+	n.SetBasePrice(addr2, NewFixedPrice(price2))
+	assert.Zero(n.priceInfo[addr1].Value().Cmp(price1))
+	assert.Zero(n.priceInfo[addr2].Value().Cmp(price2))
 	assert.Zero(n.GetBasePrices()[addr1].Cmp(price1))
 	assert.Zero(n.GetBasePrices()[addr2].Cmp(price2))
 }
+
+func TestSetAndGetCapabilityPrices(t *testing.T) {
```
rickstaa, line 183: @victorges I've added a new test for the Auto pricing per capability.

cmd/livepeer/starter/starter.go
```diff
@@ -335,6 +345,9 @@ func DefaultLivepeerConfig() LivepeerConfig {
 		AuthWebhookURL: &defaultAuthWebhookURL,
 		OrchWebhookURL: &defaultOrchWebhookURL,
 
+		// Versioning constraints
```
rickstaa, line 348: I changed this since I think this does not belong under the // API category.

cmd/livepeer/starter/starter.go
```diff
@@ -908,16 +781,36 @@ func StartLivepeer(ctx context.Context, cfg LivepeerConfig) {
...
+				// Can't divide by 0
+				panic(fmt.Errorf("-pixelsPerUnit must be > 0, provided %v", *cfg.PixelsPerUnit))
+			}
 			if cfg.PricePerUnit == nil && !*cfg.AIWorker {
 				// Prevent orchestrators from unknowingly providing free transcoding
 				panic(fmt.Errorf("-pricePerUnit must be set"))
 			} else if cfg.PricePerUnit != nil {
-				if *cfg.PricePerUnit < 0 {
-					panic(fmt.Errorf("-pricePerUnit must be >= 0, provided %d", *cfg.PricePerUnit))
+				pricePerUnit, currency, err := parsePricePerUnit(*cfg.PricePerUnit)
```
rickstaa, line 796: @victorges please duble check if I handled this conflict correctly so that it also works as expected on the transcoding side. Thanks in advance 🙏🏻.

cmd/livepeer/starter/starter.go
```diff
@@ -1143,6 +1060,179 @@ func StartLivepeer(ctx context.Context, cfg LivepeerConfig) {
 		}()
 	}
 
+	var aiCaps []core.Capability
```
rickstaa, line 1063: @victorges I had to move this over to make the PriceFeed available. In the future, we might want to clean up this script to separate the pricing addition from the model loading. For now, I aimed to minimize conflicts with our existing open commits.

cmd/livepeer/starter/starter.go
```diff
@@ -76,81 +78,84 @@ const (
...
+	MaxGasPrice             *int
+	InitializeRound         *bool
+	InitializeRoundMaxDelay *time.Duration
+	TicketEV                *string
+	MaxFaceValue            *string
+	MaxTicketEV             *string
+	MaxTotalEV              *string
+	DepositMultiplier       *int
+	PricePerUnit            *string
+	PixelsPerUnit           *string
```
rickstaa, line 132: @victorges please double check if the right types were used for the transcoding branch.

core/capabilities.go
```diff
@@ -20,16 +21,21 @@ type ModelConstraint struct {
...
 type CapabilityString []uint64
 type Constraints struct {
+	minVersion string
+}
+type PerCapabilityConstraints struct {
 	// Models contains a *ModelConstraint for each supported model ID
 	Models ModelConstraints
 }
-type CapabilityConstraints map[Capability]*Constraints
+type CapabilityConstraints map[Capability]*PerCapabilityConstraints
```
rickstaa, line 30: @victorges feel free to suggest naming improvements if needed 👍🏻.

core/capabilities.go
```diff
@@ -20,16 +21,21 @@ type ModelConstraint struct {
...
-	bitstring   CapabilityString
-	mandatories CapabilityString
-	constraints CapabilityConstraints
-	capacities  map[Capability]int
-	mutex       sync.Mutex
+	bitstring             CapabilityString
+	mandatories           CapabilityString
+	version               string
+	constraints           Constraints
+	capabilityConstraints CapabilityConstraints
```
rickstaa, line 36: @victorges could you check if this extra field was added at the right positions at the transcoding side? Thanks 🙏🏻.

cmd/livepeer/starter/starter.go
```diff
@@ -1293,7 +1383,10 @@ func StartLivepeer(ctx context.Context, cfg LivepeerConfig) {
 		*cfg.CliAddr = defaultAddr(*cfg.CliAddr, "127.0.0.1", TranscoderCliPort)
 	}
 
-	n.Capabilities = core.NewCapabilitiesWithConstraints(append(transcoderCaps, aiCaps...), core.MandatoryOCapabilities(), constraints)
+	n.Capabilities = core.NewCapabilitiesWithConstraints(append(transcoderCaps, aiCaps...), core.MandatoryOCapabilities(), core.Constraints{}, capabilityConstraints)
```
rickstaa, line 1408: @victorges I have extended the `core.NewCapabilitiesWithConstraints` method created by Yondon to accept general constraints (`core.Constraints{}`) in addition to capability constraints. Although the general constraints are not currently used, I included this field for consistency. However, this is not entirely correct as the `core.Constraints` type has no public members, making it unusable in its current form.

I would like to hear your thoughts on how we should handle this. I noticed you defined a setter and getter for the MinVersion constraint, likely to enhance encapsulation, validation, consistency, and immutability, and to prevent runtime nil errors. We could consider removing the `NewCapabilitiesWithConstraints` method and instead create a setter for the capability constraints. If we want to keep the current `core.NewCapabilitiesWithConstraints` implementation, we should make its fields public so users can utilize it. However, this approach goes against the setter-getter pattern you currently use. What do you think?

go.mod
```diff
@@ -4,18 +4,19 @@ go 1.21.5
...
+	github.com/golang/glog v1.2.1
 	github.com/golang/mock v1.6.0
 	github.com/golang/protobuf v1.5.4
 	github.com/jaypipes/ghw v0.10.0
 	github.com/jaypipes/pcidb v1.0.0
 	github.com/livepeer/ai-worker v0.1.0
 	github.com/livepeer/go-tools v0.3.6-0.20240130205227-92479de8531b
 	github.com/livepeer/livepeer-data v0.7.5-0.20231004073737-06f1f383fb18
-	github.com/livepeer/lpms v0.0.0-20240711175220-227325841434
+	github.com/livepeer/lpms v0.0.0-20240726132931-5b7b9f5e831f
```
rickstaa: @rickstaa don't forget to change this to the `ai-video` branch when https://github.com/livepeer/lpms/tree/ai-video-rebase-main is merged.

server/ai_session.go
```diff
@@ -297,7 +297,8 @@ func (sel *AISessionSelector) getSessions(ctx context.Context) ([]*BroadcastSess
 			},
 		},
 	}
-	caps := core.NewCapabilitiesWithConstraints(append(core.DefaultCapabilities(), sel.cap), nil, constraints)
+	caps := core.NewCapabilitiesWithConstraints(append(core.DefaultCapabilities(), sel.cap), nil, core.Constraints{}, capabilityConstraints)
```
victorges, line 300: This`core.Constraints{}` wasn't in either of the `master` nor` ai-video` branches. What does it mean?

install_ffmpeg.sh
```diff
@@ -1,230 +1,3 @@
 #!/usr/bin/env bash
-
-set -exuo pipefail
```
victorges, line 3: It seems like there were conflicts in this file. Did we have any changes on the `ai-video` branch and if so, it it fine to drop them completely in place of the lpms script?

install_ffmpeg.sh
```diff
@@ -1,3 +1,3 @@
 #!/usr/bin/env bash
 echo 'WARNING: downloading and executing lpms/install_ffmpeg.sh, use it directly in case of issues'
-curl https://raw.githubusercontent.com/livepeer/lpms/d9c78b62effdb4f5c8fc438b6033da1090d04a03/install_ffmpeg.sh | bash -s $1
+curl https://raw.githubusercontent.com/livepeer/lpms/5b7b9f5e831f041c6cf707bbaad7b5503c2f138d/install_ffmpeg.sh | bash -s $1
```
victorges, line 3: Ahh nevermind my other comment, it was based on the merge commit without this change.

cmd/livepeer/starter/starter.go
victorges: Got it. It got a little harder to review but I focused only on the pricing stuff and assumed the rest was copied.

core/ai.go
victorges: Hm, a different approach here would be to just add another field for the `Currency`, like it was done for the per-gateway prices config on the original PR that implemented USD pricing: https://github.com/livepeer/go-livepeer/pull/2981/files#diff-8e2167038b87d117b916131f62552da8dee7f52d8112e490330349597b393617R1473

I think this renders a clearer interface for the JSON, with no multi-type fields, and is also fully retro-compatible. It would also be good to keep a consistent schema between different configs for similar things.

core/ai.go
```diff
@@ -23,14 +24,40 @@ type AI interface {
...
 type AIModelConfig struct {
 	Pipeline          string                   `json:"pipeline"`
 	ModelID           string                   `json:"model_id"`
 	URL               string                   `json:"url,omitempty"`
 	Token             string                   `json:"token,omitempty"`
 	Warm              bool                     `json:"warm,omitempty"`
-	PricePerUnit      int64                    `json:"price_per_unit,omitempty"`
-	PixelsPerUnit     int64                    `json:"pixels_per_unit,omitempty"`
+	PricePerUnit      StringInt                `json:"price_per_unit,omitempty"`
+	PixelsPerUnit     StringInt                `json:"pixels_per_unit,omitempty"`
```
victorges: Instead of making these strings, I think you should implement a similar logic to this: https://github.com/livepeer/go-livepeer/pull/2981/files#diff-8e2167038b87d117b916131f62552da8dee7f52d8112e490330349597b393617R1489-R1490

We can't parse these values to an `int` as sometimes they can be a `float` (like `0.20` dollars), but we also can't parse it to a `float` cause we lose precision. So the idea there, which I believe should be replicated here, is to keep it as the raw characters that were in the JSON (`json.RawMessage`) and then parsing it as `big.Rat` 🐀 

One thing you could do to make this less repetitive would be creating a custom type to do exactly that. Not to support a `string` field (the currency should go on a separate field IMO, as per the other comment), but to parse it raw into a `big.Rat` (you don't even need a `json.RawMessage` confusion in that case, it should be pretty straightforward). With that custom type we could use it both here and in the per-gateway pricing config.

WDYT?

victorges: I took the liberty of implementing that and pushing to the branch here, since it was something I already wanted to do and you have a lot on your plate with this merge rn.

Here's the commit: https://github.com/livepeer/go-livepeer/pull/3103/commits/8036856cbfd6724f79b393118e49fd2451282797

Lmk what you think and feel free to revert it if you'd rather go a different direction. 

⚠️ Keep in mind that I haven't fixed the currency field on that commit, need to add that as well.

cmd/livepeer/starter/starter.go
```diff
@@ -1201,6 +1058,174 @@ func StartLivepeer(ctx context.Context, cfg LivepeerConfig) {
...
+				// Set price per unit base info.
+				pixelsPerUnit, ok := new(big.Rat).SetString(*cfg.PixelsPerUnit)
+				if !ok || !pixelsPerUnit.IsInt() {
+					panic(fmt.Errorf("-pixelsPerUnit must be a valid integer, provided %v", *cfg.PixelsPerUnit))
+				}
+				if pixelsPerUnit.Sign() <= 0 {
+					// Can't divide by 0
+					panic(fmt.Errorf("-pixelsPerUnit must be > 0, provided %v", *cfg.PixelsPerUnit))
+				}
+				pricePerUnit, currency, err := parsePricePerUnit(config.PricePerUnit.String())
```
victorges: The price should be specified as a separate field in the `config`, not as a string. See the logic for per-gateway prices: https://github.com/livepeer/go-livepeer/blob/8036856cbfd6724f79b393118e49fd2451282797/cmd/livepeer/starter/starter.go#L820-L826

This is also related to this discussion: https://github.com/livepeer/go-livepeer/pull/3103/commits/6292df161f1c719e42eab42af152b0b748b32a1a#r1698466441

victorges (commented):
Sending a separate review for the [custom-currency AI pricing commit](https://github.com/livepeer/go-livepeer/pull/3103/commits/6292df161f1c719e42eab42af152b0b748b32a1a#), since there a couple comments.

cmd/livepeer/starter/starter.go
victorges: I don't think this is a big issue and I don't think we should do a big refactor for it. The only suggestion I have is that, if the caller can never provide a non-empty `core.Constraints` anyway, we could just remove that argument from the constructor and always initialize the field with the empty value (i.e. don't even need to do nothing. go already does that by default).

Then the process is calling the constructor and then the version setter, instead of creating a `Constraints` object with the right version and then calling the constructor. WDYT?

cmd/livepeer/starter/starter.go
```diff
@@ -250,6 +251,7 @@ func DefaultLivepeerConfig() LivepeerConfig {
 	defaultAuthWebhookURL := ""
 	defaultOrchWebhookURL := ""
 	defaultMinLivepeerVersion := ""
+	defaultIgnorePreReleaseVersions := true
```
victorges: Shouldn't this default to `false` in the AI branch?

---

Also, as a general tip: It can lead to confusing CLI usage to have a boolean flag that defaults to `true`. Normally these boolean flags can be toggled by simply specifying them like `--ignore-pre-release-versions`. But in this case, to change the default one would need to do `--ignore-pre-release-versions=false`, and in some systems like Mist that can't even be done (can only set booleans to true)

So one suggestion is to invert this flag as a `--check-pre-release-versions` instead which defaults to `false` instead. This is irrelevant if you change the default value to `false` anyway on the current flag.

cmd/livepeer/starter/starter.go
```diff
@@ -1124,7 +1124,12 @@ func StartLivepeer(ctx context.Context, cfg LivepeerConfig) {
 					panic(fmt.Errorf("'pricePerUnit' value specified for model '%v' in pipeline '%v' must be >= 0, provided %v", config.ModelID, config.Pipeline, config.PricePerUnit))
 				}
 				pricePerPixel := new(big.Rat).Quo(pricePerUnit, pixelsPerUnit)
-				autoPrice, err := core.NewAutoConvertedPrice(currency, pricePerPixel, nil)
+				var autoPrice *core.AutoConvertedPrice
+				if *cfg.Network == "offchain" {
+					autoPrice = core.NewFixedPrice(pricePerPixel)
+				} else {
+					autoPrice, err = core.NewAutoConvertedPrice(currency, pricePerPixel, nil)
+				}
```
victorges: I believe this change should be reverted. The auto price converter will not panic if you call it with a base currency like `wei` or `ETH`. If you start the service in offchain mode but provide a custom currency price tho, that's a configuration error. We cannot convert the custom-currency price config into a wei price here, that could lead to even worse problems (e.g. charging 1 wei per pixel when the operator wanted 1 USD).

By reverting this, we will keep the behavior of failing startup in case that wrong configuration is made. The operator should then either fix the price to not be in a custom currency, or configure the node to be onchain.

cmd/livepeer/starter/starter.go
victorges: nevermind as this was reverted

install_ffmpeg.sh
victorges: nevermind. the commit hash for the script changed later so I understood what happened

server/ai_session.go
victorges: also understood it later from the other discussions

victorges (approved):
LGTM!! There are a couple required changes but they shouldn't be complex. I can still take another look after you make the updates if you want.

go.mod
victorges: This should be fine, especially if all (manual&auto) tests are working

cmd/livepeer/starter/starter.go
victorges: Made a couple comments about this on the specific commit. TLDR I'd say the currency should go in a separate field in the config, like on the per gateway prices, and not on the same field as a string.

let's move the discussion to one of those comments, ref: https://github.com/livepeer/go-livepeer/pull/3103#discussion_r1700906838

core/capabilities.go
victorges: Let's keep as is and possibly refactor later/on a separate PR merged here.

core/capabilities.go
victorges: Not sure I understand what you mean with position

core/ai.go
rickstaa: @victorges I like the method you suggested in [this commit](https://github.com/livepeer/go-livepeer/pull/3103/commits/8036856cbfd6724f79b393118e49fd2451282797). The only issue is that it no longer works with the old configuration format. If I understand correctly, you'd like users to specify the config in this way instead of:

```json
[
  {
    "pipeline": "text-to-image",
    "model_id": "SG161222/RealVisXL_V4.0_Lightning",
    "price_per_unit": "16USD",
    "warm": true
  }
]
```

changing it to:

```json
[
  {
    "pipeline": "text-to-image",
    "model_id": "SG161222/RealVisXL_V4.0_Lightning",
    "price_per_unit": 16,
    "currency": "USD",
    "warm": true
  }
]
```

Does that capture your intention correctly? I think that approach is clean. The reason I didn't initially go with it was to keep it similar to how the `pricePerUnit` flag worked on the transcoding side. However, since we use JSON files or JSON strings, this makes sense. 👍🏻

rickstaa: One side effect of having non-pixel-related pipelines now is that `PixelsPerUnit     JSONRat` is no longer representative of its function (see https://vscode.dev/github/livepeer/go-livepeer/blob/ai-video-rebase/core/ai.go#L51).. It acts more as a scaling factor than a true `pixelsPerUnit`. Perhaps we should rename it to `ScalingFactor`? Tagging @ad-astra-video, @eliteprox, @pschroedl for visibility.

rickstaa: @victorges I just saw https://github.com/livepeer/go-livepeer/pull/3103#discussion_r1700935708 and https://github.com/livepeer/go-livepeer/pull/3103#discussion_r1700891940👍🏻. Will create a commit.

core/capabilities.go
rickstaa: Ah sorry for that since the same exact structure is used in this version there is no change for the transcoding side code and it should work. Let's re-discuss when I have your improved message implemented 👍🏻.

cmd/livepeer/starter/starter.go
rickstaa: Yea the rest was kept the same 👍🏻.

core/ai.go
rickstaa: This is a great change. Thanks for implementing it. I will implement the `Currency` field 👍🏻.

cmd/livepeer/starter/starter.go
rickstaa: This was added because the NewAutoConvertedPrice does not work without a priceFeed being loaded and in `offChain` mode it will not be loaded. I will try to ensure that pricing in the AI case is only used in on-chain mode 👍🏻.

cmd/livepeer/starter/starter.go
```diff
@@ -1095,6 +1095,16 @@ func StartLivepeer(ctx context.Context, cfg LivepeerConfig) {
 			return
 		}
 
+		// Get base pixels per unit.
```
rickstaa, line 1098: Moved to ensure that the command line flag is used when the `PixelsPerUnit` is missing in the configuration file.

cmd/livepeer/starter/starter.go
```diff
@@ -1105,30 +1115,28 @@ func StartLivepeer(ctx context.Context, cfg LivepeerConfig) {
...
-				pricePerPixel := new(big.Rat).Quo(pricePerUnit, pixelsPerUnit)
 				var autoPrice *core.AutoConvertedPrice
-				if *cfg.Network == "offchain" {
-					autoPrice = core.NewFixedPrice(pricePerPixel)
-				} else {
-					autoPrice, err = core.NewAutoConvertedPrice(currency, pricePerPixel, nil)
-				}
-				if err != nil {
-					panic(fmt.Errorf("error converting price: %v", err))
+				if *cfg.Network != "offchain" {
```
rickstaa, line 1119: @victorges I now added simple if conditions since I didn't want to cause even more conflicts with the already excisting pull requests that are on the `ai-video` branch. I however would like to improve this logic in the future so I created https://linear.app/livepeer-ai/issue/AI-546/improve-ai-configuration-parsing-logic-in-startergo.

core/ai.go
```diff
@@ -49,26 +49,10 @@ type AIModelConfig struct {
 	Warm              bool                     `json:"warm,omitempty"`
 	PricePerUnit      JSONRat                  `json:"price_per_unit,omitempty"`
 	PixelsPerUnit     JSONRat                  `json:"pixels_per_unit,omitempty"`
+	Currency          string                   `json:"currency,omitempty"`
 	OptimizationFlags worker.OptimizationFlags `json:"optimization_flags,omitempty"`
 }
 
-func (config *AIModelConfig) UnmarshalJSON(data []byte) error {
```
rickstaa, line 37: @victorges I removed the default behavoir since I figured we would like to use the default specified by the command line when `PixelsPerUnit` is not supplied.

cmd/livepeer/starter/starter.go
rickstaa: @ad-astra-video already did some improvements in the AI remote worker implementation https://github.com/livepeer/go-livepeer/pull/3088. Maybe we can merge this ugly variant and handle it later?

core/ai.go
rickstaa: @victorges I added the currency json field. See https://github.com/livepeer/go-livepeer/pull/3103#discussion_r1701635888 and https://github.com/livepeer/go-livepeer/pull/3103/commits/8e654d7ad60b68b609be069ebcdfd045978102f6.

cmd/livepeer/starter/starter.go
rickstaa: Fixed in https://github.com/livepeer/go-livepeer/pull/3103/commits/8e654d7ad60b68b609be069ebcdfd045978102f6.

rickstaa:
@victorges I'm currently testing if the pricing still works correctly.

rickstaa:
> @victorges I'm currently testing if the pricing still works correctly.

@victorges if you agree with my last commit I think we can merge this and handle the following things in seperate pull requests:
- [ ]  **Improve the AI starter logic:** handle when merging the AI remote worker.
- [x]  **Improve the Constraint message:** I will create another pull request after this one -> https://github.com/livepeer/go-livepeer/pull/3115.
- [x]  **Improve NewCapabilitiesWithConstraints:** I will create another pull request after this one -> https://github.com/livepeer/go-livepeer/pull/3115.

cmd/livepeer/starter/starter.go
rickstaa: Implemented in https://github.com/livepeer/go-livepeer/pull/3115. I now took the setter getter route and removed the NewCapabilitiesWithConstraints we can also take the route of making the fields public but I wanted to keep it in line with the rest of the codebase. Please let me know what you prefer 👍🏻.

rickstaa:
@victorges looks like the docker container build ci fails most likely due to incompatible system requirements. The new ffmpeg requires Cuda 12.4 and ubuntu 22.04 or 24.04. I can take a look later 👍🏻.

rickstaa:
@victorges, unfortunately, moving the AI config logic introduced a racing condition https://github.com/livepeer/go-livepeer/pull/3115#issuecomment-2265908066. This has to be fixed before we merge it. The reason it was moved is that the [NewAutoConvertedPrice](https://github.com/livepeer/go-livepeer/blob/92f549c5b3c133b8e7ceb91ee0e99ca7f943245c/cmd/livepeer/starter/starter.go#L824) method required the [PriceFeedWatcher](https://github.com/livepeer/go-livepeer/blob/92f549c5b3c133b8e7ceb91ee0e99ca7f943245c/cmd/livepeer/starter/starter.go#L756) to be running.

rickstaa:
@victorges this pull request could possibly contain a racing condition https://github.com/livepeer/go-livepeer/pull/3115#issuecomment-2265908066. Will look at it next week.

cmd/livepeer/starter/starter.go
victorges: I'm not sure if it's allowed to be missing, but makes sense. Is this how the per gateway configs work as well? (dont need to be the same, just to know if it's a new thing or a fix)

cmd/livepeer/starter/starter.go
victorges: So the idea is we ignore any pricing config when it's offchain, is that it? I guess it makes sense.

core/ai.go
victorges: Yeah I think that's valid. These configs could be treated as "overrides" for the specific capability right.

core/ai.go
victorges: new version is perfect!

victorges: > One side effect of having non-pixel-related pipelines now is that PixelsPerUnit     JSONRat is no longer representative of its function (see https://vscode.dev/github/livepeer/go-livepeer/blob/ai-video-rebase/core/ai.go#L51).. It acts more as a scaling factor than a true pixelsPerUnit. Perhaps we should rename it to ScalingFactor?

Yeah I think that could make sense. If we rename this I'd say we should go all the way and think of consistent naming to use for both parameters. I don't think "price per unit" would make that much sense if there's no "unit" mentioned anywhere else for example. We could also use terms like "numerator" and "denominator" to be super clear how they are used, it was pretty confusing for operators when they had to start using this `pixelsPerUnit` flag (and I always have to think twice about it).

TLDR Open to renaming but I think we should have a small discussion about what it should be.

And FTR, I don't think it's a big issue to keep it with a weird name for now. Just the conflicts we're going to have by renaming it here might be enough to avoid it.

cmd/livepeer/starter/starter.go
victorges: I'm ok with the current solution. Just to clarify this:

> This was added because the NewAutoConvertedPrice does not work without a priceFeed being loaded and in offChain mode it will not be loaded.

It doesn't fail if no custom currency is specified (wei or eth): https://github.com/livepeer/go-livepeer/blob/706ec33286344613ec37c4f0ac4d4f3d17657049/core/autoconvertedprice.go#L48

So what is actually wrong is configuring `network=offchain` and a custom currency, since there's no way to use a custom currency offchain. My argument was that that is a config error, not a problem with this code. I'm fine with the current behavior you've implemented tho, just not setting any price anywhere when offchain.

victorges (approved):
LGTM! Fix the build and it's good to go

cmd/livepeer/starter/starter.go
rickstaa: The reason I allow the `pixels_per_unit` to be missing in the AI config is that it can fall back to the default value of 1, which is set [here](https://github.com/livepeer/go-livepeer/blob/9cdeeb5a22674e78bd5593f22604858ed6c2feb2/cmd/livepeer/starter/starter.go#L219).

In the current behavior, if `pixels_per_unit` is not set in the config, it defaults to the value specified in the `-pixelsPerUnit` flag, or to 1 if that flag is also missing.

For pricing, the system will first check the `price_per_unit` in the config. If that is not set, it will use the value specified in the `-pricePerUnit` flag for all pipelines.

We could also opt to throw an error if these prices are not set. This could be implemented in a separate pull request. For the `pricePerUnit` this might make sense. I don't have a real preference.

PR #3109, opened 2024-07-29
change default for -autoAdjustPrice to false

eliteprox:

- 
- 
- 

**Checklist:**

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -3,6 +3,7 @@
 ## vX.X
 
 ### Breaking Changes 🚨🚨
+- #3109 Change default for `-autoAdjustPrice` to **false**
 
 ### Features ⚒
 
--- a/cmd/livepeer/starter/starter.go
+++ b/cmd/livepeer/starter/starter.go
@@ -204,7 +204,7 @@ func DefaultLivepeerConfig() LivepeerConfig {
 	defaultMaxPricePerUnit := "0"
 	defaultPixelsPerUnit := "1"
 	defaultPriceFeedAddr := "0x639Fe6ab55C921f74e7fac1ee960C0B6293ba612" // ETH / USD price feed address on Arbitrum Mainnet
-	defaultAutoAdjustPrice := true
+	defaultAutoAdjustPrice := false
 	defaultPricePerGateway := ""
 	defaultPricePerBroadcaster := ""
 	defaultBlockPollingInterval := 5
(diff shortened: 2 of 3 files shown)


rickstaa:
@eliteprox thanks for creating this pull request. What do you think about applying @Titan-Node's other suggestion of setting a maximum gas price the orchestrator accepts. This will prevent people from paying infinite gas in the worst case scenario 🤔.

eliteprox:
> @eliteprox thanks for creating this pull request. What do you think about applying @Titan-Node's other suggestion of setting a maximum gas price the orchestrator accepts. This will prevent people from paying infinite gas in the worst case scenario 🤔.

I believe that is already implemented with the `-maxGasPrice` flag. This works the same for both orchestrator and gateway. https://github.com/livepeer/go-livepeer/blob/e9d4c393ebf940888e037326c58d9b17c3c83264/cmd/livepeer/livepeer.go#L164

Titan-Node:
I think the conversation around the max has price had something to do with the orchestrator rejecting work based on a gas spike regardless of the -maxGasPrice flag. I think the max gas flag is only for redeeming tickets but someone mentioned an Orch would still reject work if gas spiked. Maybe it was around the ticket EV changing or some other parameter that is affected with gas fluctuation? Is there a way we could test spiking the gas and making sure the Orch still accepts the work?
Would this be done with a unit test?

ad-astra-video:
I think the `avgGasPrice` values need to be updated to help Orchestrators not reject work during gas price spikes.  

When we switched to arbitrum the the gas price was closer to 1 gwei and 3 gwei was used for average gas price to be conservative.  Now gas price is 0.01 usually.

The other factor in the average transaction cost gas used is a hard coded number at 1,200,000 that was appropriate at time of switch to arbitrum.  This may be able to be reduced. I also see the gas used varying wildly on some ticket redemptions driving cost well above the value of the ticket. Maybe there is more going on here that should be looked at to confirm we are still calculating correctly after the changes to arbitrum over the last couple years

$262    0x14a0c5171b9451ddd83073971f43d4f417476e956daf31ad1c9c6f58e3247a4d
$259    0x1dbf66240b878997b1ee4ce23bd033bbd767c765c1351653b7c0ea1136c2b1ed
$16      0x030927eb41dc0030b373c9b9d289a26829e929378f38e7bae1f75a7b534adf38
$10      0x1483f1b3ec7ecdbb0772cc2b696161c2f09bf879933e6b32db15e145c44e011d

https://github.com/livepeer/go-livepeer/blob/e9d4c393ebf940888e037326c58d9b17c3c83264/cmd/livepeer/starter/starter.go#L55

https://github.com/livepeer/go-livepeer/blob/e9d4c393ebf940888e037326c58d9b17c3c83264/pm/recipient.go#L34

https://github.com/livepeer/go-livepeer/blob/e9d4c393ebf940888e037326c58d9b17c3c83264/pm/recipient.go#L306

PR #3111, opened 2024-07-31
feat: allow Gateways to filter by pre-release

rickstaa:

This pull request gives Gateways the ability to filter by pre-release suffix. When a pre-release suffix is specified in the `OrchMinLivepeerVersion` command line argument the software now also checks the pre-release version suffix on the orchestrator.

- Changed the logic in https://github.com/livepeer/go-livepeer/pull/3111/files#diff-fdba37d9bd7a580de9aaa45bf86524a4caa83516a73af118cb116e93d0220c73 to also check the pre-release suffic when present in the `OrchMinLivepeerVersion` CLI argument.
- Updated the tests.

Ensured tests worked. Did not run an off-chain or on-chain transcoding gateway yet but did do this on the AI side.

No but this is required to be able to filter versions on the https://github.com/livepeer/go-livepeer/tree/ai-video branch since we use pre-release suffixes to seperate our versions from the main branhc.

**Checklist:**

### DIFF
--- a/core/capabilities.go
+++ b/core/capabilities.go
@@ -341,11 +341,22 @@ func (bcast *Capabilities) LivepeerVersionCompatibleWith(orch *net.Capabilities)
 		return false
 	}
 
-	// Ignore prerelease versions as in go-livepeer we actually define post-release suffixes
-	minVerNoSuffix, _ := minVer.SetPrerelease("")
-	verNoSuffix, _ := ver.SetPrerelease("")
+	// If minVersion has no pre-release component, ignore pre-release versions.
+	// If versions match without pre-release, but only minVersion has a suffix, return false.
+	// Needed because we use post-release suffixes in go-livepeer.
+	minVerHasSuffix := minVer.Prerelease() != ""
+	verHasSuffix := ver.Prerelease() != ""
+	if minVer.Prerelease() == "" || ver.Prerelease() == "" {
+		minVerNoSuffix, _ := minVer.SetPrerelease("")
+		verNoSuffix, _ := ver.SetPrerelease("")
+		minVer = &minVerNoSuffix
+		ver = &verNoSuffix
+	}
+	if minVer.Equal(ver) && minVerHasSuffix && !verHasSuffix {
+		return false
+	}
 
-	return !verNoSuffix.LessThan(&minVerNoSuffix)
+	return !ver.LessThan(minVer)
 }
 
 func (bcast *Capabilities) CompatibleWith(orch *net.Capabilities) bool {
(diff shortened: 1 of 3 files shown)


core/capabilities.go
```diff
@@ -341,11 +341,16 @@ func (bcast *Capabilities) LivepeerVersionCompatibleWith(orch *net.Capabilities)
 		return false
 	}
 
-	// Ignore prerelease versions as in go-livepeer we actually define post-release suffixes
-	minVerNoSuffix, _ := minVer.SetPrerelease("")
-	verNoSuffix, _ := ver.SetPrerelease("")
+	// If minVersion has no pre-release component, ignore pre-release versions by
+	// default as in go-livepeer, we define post-release suffixes.
```
rickstaa: @victorges I noticed that I typically use punctuation in my comments, whereas it's not common practice here. I'm happy to adjust and remove it if preferred.

core/capabilities_test.go
```diff
@@ -339,12 +339,40 @@ func TestCapability_CompatibleWithNetCap(t *testing.T) {
...
 	bcast = NewCapabilities(nil, nil)
 	bcast.constraints.minVersion = "0.4.1"
 	orch.version = "0.4.1"
 	assert.True(bcast.CompatibleWith(orch.ToNetCapabilities()))
+
+	// broadcaster is compatible with orchestrator - no pre-release min version suffix
+	orch = NewCapabilities(nil, nil)
+	bcast = NewCapabilities(nil, nil)
+	bcast.constraints.minVersion = "0.4.1"
+	orch.version = "0.4.1-0.21000000000000-06f1f383fb18"
```
victorges, line 353: Oh if our versions are like this and not like `ai-video` or `some-specific-feature` then the standard semver comparison from `Masterminds/semver` might be great

core/capabilities.go
```diff
@@ -341,11 +341,22 @@ func (bcast *Capabilities) LivepeerVersionCompatibleWith(orch *net.Capabilities)
...
+	verHasSuffix := ver.Prerelease() != ""
+	if minVer.Prerelease() == "" || ver.Prerelease() == "" {
+		minVerNoSuffix, _ := minVer.SetPrerelease("")
+		verNoSuffix, _ := ver.SetPrerelease("")
+		minVer = &minVerNoSuffix
+		ver = &verNoSuffix
+	}
+	if minVer.Equal(ver) && minVerHasSuffix && !verHasSuffix {
+		return false
+	}
```
victorges, line 357: I don't think we need this logic since `Masterminds/semver` already implements proper pre-release comparison, considering pre-releases lower than the release version. Their code is pretty straightforwad, can dig into `LessThan` implementation below to understand what they do

victorges (approved):
LGTM!

rickstaa:
Closed as this doesn't make sense for the regular transcoding network since we don't use time-based pre-releases. We use commit suffixes (see https://github.com/livepeer/go-livepeer/blob/master/print_version.sh).

PR #3115, opened 2024-08-02
Ai video rebase improve rpc message

rickstaa:
> [!IMPORTANT]
> Should be merged after https://github.com/livepeer/go-livepeer/pull/3103.

This pull request unifies the capabilities constraints message structure by combining the minVersion constraint and CapabilityConstraints into a single Constraints message field. The `NewCapabilitiesWithConstraints` method has been removed for consistency with how the minVersion constraint is handled, allowing the constraints fields to be private.

- Changed the structure of the Capabilities RPC message. 
- Updated all code references to the fields in that message.
- Removed the `NewCapabilitiesWithConstraints` for consistency.

I ensured the binaries could be build successfully and then tested the code in a E2E fashion with off-chain and on-chain AI Gateways and Orchestrators.

It applies suggestions made by @victorges in https://github.com/livepeer/go-livepeer/pull/3103.

**Checklist:**


core/capabilities.go
```diff
@@ -543,13 +542,6 @@ func NewCapabilities(caps []Capability, m []Capability) *Capabilities {
 	return c
 }
 
-func NewCapabilitiesWithConstraints(caps []Capability, m []Capability, constraints Constraints, capabilityConstraints CapabilityConstraints) *Capabilities {
```
rickstaa, line 546: @Yondonfu, just double-checking if there was a specific reason for initializing the constraints in the constructor?

rickstaa:
It looks like https://github.com/livepeer/go-livepeer/pull/3115/commits/1f7ea785e40d378f93a4e3c9e48ae10353deb575 will cause a `nil` pointer error when the following happens:

<details>
<summary>Old Reproduction steps!</summary>

**Reproduction steps**
1. Start AI orchestrator.
2. Start AI Gateway.
3. Send requests.
4. See if it works.
5. Stop Gateway.
6. See orchestrator crash and throw `nil` pointer error.

**Reproduction steps 2**
1. Start AI Gateway.
2. Start AI Orchestrator.
3. Send requests (successfully).
4. Send another request.
5. Stop orchestrator during request.
6. See Gateway throw `nil` pointer error[here](https://github.com/livepeer/go-livepeer/blob/92f549c5b3c133b8e7ceb91ee0e99ca7f943245c/server/ai_process.go#L89) but not crash.

</details>

**Reproduce steps 3**

A race condition causes the issue. You can reproduce it as follows:

1. Start AI Orchestrator.
2. Put a breakpoint on [this line](https://github.com/livepeer/go-livepeer/blob/92f549c5b3c133b8e7ceb91ee0e99ca7f943245c/cmd/livepeer/starter/starter.go#L1408)
3. Wait for the breakpoint to hit.
3. Start an AI Gateway.
4. Send a request to the Gateway.
5. See Orchestrator throw `nil` error on [this line](https://github.com/livepeer/go-livepeer/blob/92f549c5b3c133b8e7ceb91ee0e99ca7f943245c/core/orchestrator.go#L353) since [this logic](https://github.com/livepeer/go-livepeer/blob/92f549c5b3c133b8e7ceb91ee0e99ca7f943245c/cmd/livepeer/starter/starter.go#L869) in the `/GetOrchestrator` endpoint is called before the `caps.Constraints` from the orchestrator side is initialised.

This bug happens now that we moved the [AI startup Logic](https://github.com/livepeer/go-livepeer/blob/1f7ea785e40d378f93a4e3c9e48ae10353deb575/cmd/livepeer/starter/starter.go#L1063-L1256) under the logic that starts the Orchestrator server routes.

core/capabilities.go
rickstaa: Maybe because of `nil` errors 🤔. I will do some debugging to see if I can get this to work.

yondonfu: > just double-checking if there was a specific reason for initializing the constraints in the constructor?

IIRC didn't have much of a reason outside of the fact at the time constraints were only set once upon initializing a Capabilities struct (i.e. based on the models configuration file on startup for an orchestrator or based on the model requested by a gateway while discovering orchestrators for selection) so I think I opted to just set the constraints once in the constructor leaving changes to support modification of the constraints to be made later on when there was a better sense of how constraints should be updated over time (i.e. to reflect changes in supported models).

net/lp_rpc.proto
```diff
@@ -112,9 +112,10 @@ message Capabilities {
 
     Constraints constraints = 5;
 
-    // Non-binary general constraints.
+    // Non-binary constraints.
     message Constraints {
       string minVersion = 1;
+      map<uint32, CapabilityConstraints> PerCapability = 2;
```
victorges, line 118: Yeaah I think this looks much better :)

Is the breaking change gonna be painful?

core/capabilities.go
```diff
@@ -723,6 +715,19 @@ func (bcast *Capabilities) LegacyOnly() bool {
 	return bcast.bitstring.CompatibleWith(legacyCapabilityString)
 }
 
+func (bcast *Capabilities) SetPerCapabilityConstraints(constraints PerCapabilityConstraints) {
+	if bcast != nil {
+		bcast.constraints.perCapability = constraints
```
victorges, line 720: I think your nil reference might've been because the `perCapability` map is not being initialized? You can't write to a map in go without first constructing one (differently from slices).

So perhaps the simplest here would be to add a logic to check if the `perCapability` is `nil` and initializing it so, wdyt?

```suggestion
		if bcast.constraints.perCapability == nil {
			bcast.constraints.perCapability = PerCapabilityConstraints{}
		}
		bcast.constraints.perCapability = constraints
```

rickstaa: @victorges, thanks for your comment. I investigated the issue, and while this problem wasn't the cause of the `nil` error I experienced, I agree that this `nil` error can occur. To address it, I now initialize the field in the `NewCapabilities` constructor.

The cause of the `nil` error I encountered was due to Gateways with old software, which didn't have the RPC fields, occasionally calling my orchestrator during debugging. This explained the infrequency of the crash. I've now handled it by checking if the new fields are present (see [commit f442382](https://github.com/livepeer/go-livepeer/pull/3115/commits/f44238233b09424bb1fde2e1133d5bd08a23b925)).

net/lp_rpc.proto
rickstaa: What happens is that new gateways cannot send requests to old orchestrators, and apparently, old gateways can crash orchestrators running the newer software (see [this discussion](https://github.com/livepeer/go-livepeer/pull/3115#discussion_r1709719310)). However, I have fixed the latter issue. The only thing left to investigate is whether the new gateway, when the version filter is not set, will crash the old orchestrators. I think the RPC check should handle that case but I will double check. After this I think we can merge 👍🏻 .

rickstaa: One thing that surprises me is the lack of RPC validation for calling the GetOrchestrator endpoint 🤔.

rickstaa: Okay I found the behavoir:

**Old orch new Gateway**

This situation is protected since the Gateway throws a `grpc: error unmarshalling request: proto:\u00a0cannot parse invalid wire-format data` error during discovery so will not be considered 👍🏻. This behavoir happens [here](https://github.com/livepeer/go-livepeer/blob/f44238233b09424bb1fde2e1133d5bd08a23b925/net/lp_rpc_grpc.pb.go#L50).

**New Orch old Gateway**

rickstaa:
### Pull Request Report: Understanding Unidirectional Compatibility in gRPC Message Changes

@victorges and @ad-astra-video, I researched why an `error unmarshalling request: proto:\u00a0cannot parse invalid wire-format data` warning is thrown when a **new** Gateway tries to interact with an **old** Orchestrator but not when an **old** Gateway interacts with a **new** Orchestrator.

**TL;DR:** This is because gRPC and Protobuf are designed to be backward-compatible, meaning newer versions can handle older message formats, but not vice versa if there are incompatible changes.

#### Background

In our RPC communication between the Gateway and Orchestrator using gRPC, we have updated the `Capabilities` message format. This update introduced a new field (`version`) and modified the structure of the `constraints` field. The changes were intended to add the AI and min version constraints. However, during testing, we observed that while communication from an old Gateway to a new Orchestrator passed through successfully, the reverse direction (new Gateway to old Orchestrator) failed with an unmarshalling error.

#### Changes Overview

**Old `Capabilities` Message:**
```protobuf
message Capabilities {
    repeated uint64 bitstring = 1;
    repeated uint64 mandatories = 2;
    map<uint32, uint32> capacities = 3;

    message Constraints {
        message ModelConstraint {
            bool warm = 1;
        }
        map<string, ModelConstraint> models = 1;
    }
    map<uint32, Constraints> constraints = 4;
}
```

**New `Capabilities` Message:**
```protobuf
message Capabilities {
    repeated uint64 bitstring = 1;
    repeated uint64 mandatories = 2;
    map<uint32, uint32> capacities = 3;
    string version = 4;

    message Constraints {
        string minVersion = 1;
        map<uint32, CapabilityConstraints> PerCapability = 2;
    }
    Constraints constraints = 5;

    message CapabilityConstraints {
        message ModelConstraint {
            bool warm = 1;
        }
        map<string, ModelConstraint> models = 1;
    }
}
```

#### Issue Explanation

**Old Gateway to New Orchestrator (Passes Through):**
- The old `Capabilities` message format does not include the new `version` field and uses a different structure for the `constraints` field.
- When the new Orchestrator receives the old `Capabilities` message, it can still process the message because Protobuf is designed to ignore unknown fields and handle backward compatibility. This means the new Orchestrator can handle the old format without error, skipping over the fields it does not recognize or expect.

  What happens under the hood is that since the old message format does not contain the `constraints` field (at index 5 in the new format), the new Orchestrator does not find any data at this index in the message and thus sets its internal `constraints` field to `nil`. This is the default behavior when a required field is missing or not present in the incoming message.

  Additionally, it tries to unmarshal the `version` field, which is now at index 4. The binary data in the old message that contains the `constraints` map is misinterpreted as a string, leading to nonsensical values (e.g., `Version = "\b\x1b\x12'\n%\n!SG161222/RealVisXL_V4.0_Lightning\x12\x00"`).

**New Gateway to Old Orchestrator (Fails to Unmarshal):**
- The new `Capabilities` message format includes additional fields and changes to the `constraints` field structure that the old Orchestrator does not recognize.
- When the old Orchestrator receives the new `Capabilities` message, it encounters fields it cannot parse:
  - **Constraints Field Index Change**: The new message places the `constraints` field at index 5, which does not exist in the old message. Instead, the old message expects a `map<uint32, Constraints>` at index 4. This results in the old Orchestrator receiving unexpected data types and structures, causing the unmarshalling to fail.
  - **Version Field Addition**: The new message introduces a `version` field at index 4. The old message does not expect this field and tries to interpret it as the wrong type (a string instead of a map), leading to parsing errors.

#### How Protobuf Backward Compatibility Works

1. **Field Addition**: New fields can be added to the end of a message. Older versions will ignore these fields, allowing backward compatibility. This is why the new Orchestrator can process messages from the old Gateway.
2. **Field Indexing and Type Changes**: Changing the index or type of existing fields breaks compatibility. The old Orchestrator cannot recognize or correctly parse the new field structure and types, resulting in unmarshalling errors.

#### Solution and Recommendation

To ensure smooth backward and forward compatibility in gRPC message formats, adhere to the following guidelines:
1. **Field Addition**: Adding new fields is generally safe as older versions ignore unknown fields. Ensure new fields are added with unique field numbers.
2. **Field Modification**: Avoid changing the type or index of existing fields. If necessary, deprecate the old field and add a new field with a different number and the new structure.

For this particular case, to maintain compatibility, we should:
- Keep the old `constraints` field as deprecated.
- Introduce the new `constraints` field with a new field number.

**Revised New `Capabilities` Message:**
```protobuf
message Capabilities {
    repeated uint64 bitstring = 1;
    repeated uint64 mandatories = 2;
    map<uint32, uint32> capacities = 3;
    string version = 4;
    Constraints constraints = 5;

    // Deprecated constraints field
    map<uint32, Constraints> old_constraints = 6 [deprecated = true];

    message Constraints {
        string minVersion = 1;
        map<uint32, CapabilityConstraints> PerCapability = 2;
    }

    message CapabilityConstraints {
        message ModelConstraint {
            bool warm = 1;
        }
        map<string, ModelConstraint> models = 1;
    }
}
```

This approach ensures that:
- The new Orchestrator can still process messages from the old Gateway correctly by ignoring the unknown `constraints` field.
- The old Orchestrator can handle both old and new messages, ensuring full compatibility across versions.

#### Conclusion

The observed issue highlights the importance of careful management of protocol changes to maintain compatibility. By following the recommended approach, we can ensure robust communication between different versions of our nodes without encountering unmarshalling errors. However, my recommendation now is to **move forward with the merge and deployment as it is since the AI subnet is new and we can still introduce breaking changes**. The fix I did in [this commit](https://github.com/livepeer/go-livepeer/pull/3115/commits/f44238233b09424bb1fde2e1133d5bd08a23b925) prevents new Orchestrators from throwing a `nil` error when an old Gateway sends them a request. Due to how gRPC is implemented, new Gateways will throw an `error unmarshalling request: proto:\u00a0cannot parse invalid wire-format data` warning when they try to discover old Orchestrators. Further, we can use the new `minVersion` constraint to filter out old Orchestrators. We, however, need some kind of deployment strategy. I would like to advise the following:
- We update our 10 base capacity Orchestrators to the new software and add 15 temporary new ones.
- We update our own Gateways to the new software (coordination with Livepeer.cloud needed).
- We then release the new version and urge people to move their Orchestrator to the new version.

PR #3116, opened 2024-08-03
feat(pricing): expand gateway max price to set per capability/model id

ad-astra-video:
This PR allows setting prices per capability and model id.

Current Implementation

Gateways currently use one MaxPrice set at launch with `maxPricePerUnit` flag and can be adjusted with the `setBroadcastConfig` endpoint of the cli webserver while the gateway is running. 

Enhancement Requested

Gateways on the AI subnet need the ability to filter orchestrators on price for different pipelines/model ids.  Transcoding would also benefit from this when different encoding codecs are enabled on the network.  The single MaxPrice needs to be updated to enable setting prices per capability (e.g. text-to-image and image-to-video) separately as well as allow the ability to price models within the capabilities separately.  The single MaxPrice will function as the overall default if no prices are set for specific pipelines and/or models.  The separate pricing needs to also allow setting a price per capability for all models in a capability.

Proposed implementation Changes

A `maxPricePerCapability` launch flag is added to accept json configuration of maximum prices for a capability/model.  The json configuration uses the `maxPricePerCapability` flag configuration to provide consistency in how prices are set on the network.  The `server.BroadcastCfg` is updated on the Gateway to store prices per capability/model id that can then be queried to provide the appropriate MaxPrice to use when selecting orchestrators in the selection algorithm.  The `server.BroadcastCfg` is also extended to add new functions for `GetCapabilitiesMaxPrice` and `GetCapabilityMaxPrice` to provide ability to get correct price with provided capabilities of the request. The `GetCapabilitiesMaxPrice` returns a total MaxPrice for the capabilities included in the request.  If no specific max prices is set for the capabilities requested, the MaxPrice is used.  `GetCapabilityMaxPrice` returns the max price set for that capability/model id or zero if not set.

The base `MaxPrice` and `SetMaxPrice` functions are updated to set pricing of H264 encoding capability to allow extending this to provide different prices for different codecs in the future.  ModelID is not used for transcoding so expected to always be "default".


cmd/livepeer/starter/starter_test.go
```diff
@@ -109,6 +109,31 @@ func TestParseGetGatewayPrices(t *testing.T) {
...
+	assert := assert.New(t)
+
+	jsonTemplate := `{"capabilities_prices": [ {"pipeline": "text-to-image", "model_id": "stabilityai/sd-turbo", "priceperunit": 1000, "pixelsperunit": 1}, {"pipeline": "image-to-video", "model_id": "default", "priceperunit": 2000, "pixelsperunit": 3} ] }`
+
+	prices := getCapabilityPrices(jsonTemplate)
+	assert.NotNil(prices)
+	assert.Equal(2, len(prices))
+
+	//confirm gateway field is initialized to empty string
+	assert.Equal(prices[0].Gateway, "")
```
eliteprox:
```suggestion
	assert.Equal(prices[0].Gateway, "default")
```
Update value for test

server/handlers.go
```diff
@@ -199,6 +199,66 @@ func setBroadcastConfigHandler() http.Handler {
...
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if s.LivepeerNode.NodeType == core.BroadcasterNode {
+			maxPricePerUnit := r.FormValue("maxPricePerUnit")
+			pixelsPerUnit := r.FormValue("pixelsPerUnit")
+			currency := r.FormValue("currency")
+			pipeline := r.FormValue("pipeline")
+			modelID := r.FormValue("modelID")
+
+			if pipeline == "" || modelID == "" {
+				respond400(w, "pipeline and modelID must be set")
```
eliteprox, line 212: Add return statements to prevent runtime error after failed request validation
```suggestion
				respond400(w, "pipeline and modelID must be set")
				return
```

server/handlers.go
```diff
@@ -199,6 +199,66 @@ func setBroadcastConfigHandler() http.Handler {
...
+			}
+
+			cap := core.PipelineToCapability(pipeline)
+			if cap == core.Capability_Unused {
+				respond400(w, "pipeline not supported")
```
eliteprox, line 218:
```suggestion
				respond400(w, "pipeline not supported")
				return
```

server/handlers.go
```diff
@@ -199,6 +199,66 @@ func setBroadcastConfigHandler() http.Handler {
...
+			}
+
+			// set max price
+			if maxPricePerUnit != "" && pixelsPerUnit != "" {
+				pr, ok := new(big.Rat).SetString(maxPricePerUnit)
+				if !ok {
+					respond400(w, fmt.Sprintf("Error parsing pricePerUnit value: %s", maxPricePerUnit))
```
eliteprox, line 226:
```suggestion
					respond400(w, fmt.Sprintf("Error parsing pricePerUnit value: %s", maxPricePerUnit))
					return
```

server/handlers.go
```diff
@@ -199,6 +199,66 @@ func setBroadcastConfigHandler() http.Handler {
...
+				}
+				px, ok := new(big.Rat).SetString(pixelsPerUnit)
+				if !ok {
+					respond400(w, fmt.Sprintf("Error parsing pixelsPerUnit value: %s", pixelsPerUnit))
```
eliteprox, line 231:
```suggestion
					respond400(w, fmt.Sprintf("Error parsing pixelsPerUnit value: %s", pixelsPerUnit))
					return
```

server/handlers.go
```diff
@@ -199,6 +199,66 @@ func setBroadcastConfigHandler() http.Handler {
...
+				}
+				if px.Sign() <= 0 {
+					respond400(w, fmt.Sprintf("pixels per unit must be greater than 0, provided %v", pixelsPerUnit))
```
eliteprox, line 235:
```suggestion
					respond400(w, fmt.Sprintf("pixels per unit must be greater than 0, provided %v", pixelsPerUnit))
					return
```

server/handlers.go
```diff
@@ -199,6 +199,66 @@ func setBroadcastConfigHandler() http.Handler {
...
+				if pricePerPixel.Sign() > 0 {
+					var err error
+					autoPrice, err = core.NewAutoConvertedPrice(currency, pricePerPixel, func(price *big.Rat) {
+						if monitor.Enabled {
+							monitor.MaxPriceForCapability(core.CapabilityNameLookup[cap], modelID, price)
+						}
+						glog.Infof("Maximum price per unit set to %v wei for capability=%v model_id=%v", price.FloatString(3), pipeline, modelID)
+					})
+					if err != nil {
+						respond400(w, errors.Wrap(err, "error converting price").Error())
```
eliteprox, line 250:
```suggestion
						respond400(w, errors.Wrap(err, "error converting price").Error())
						return
```

server/handlers.go
```diff
@@ -199,6 +199,66 @@ func setBroadcastConfigHandler() http.Handler {
...
+					}
+
+					BroadcastCfg.SetCapabilityMaxPrice(cap, modelID, autoPrice)
+					respondOk(w, nil)
+				} else {
+					respond400(w, fmt.Sprintf("pricePerPixel needs to be > 0: %v", pricePerPixel.FloatString(3)))
+				}
+			} else {
+				respond400(w, "maxPricePerUnit and pixelsPerUnit need to be set")
```
eliteprox, line 260:
```suggestion
				respond400(w, "maxPricePerUnit and pixelsPerUnit need to be set")
				return
```

server/handlers.go
```diff
@@ -199,6 +199,66 @@ func setBroadcastConfigHandler() http.Handler {
...
+			}
+		} else {
+			respond400(w, "Node must be gateway node to set max price per capability")
```
eliteprox, line 264:
```suggestion
			respond400(w, "Node must be gateway node to set max price per capability")
			return
```

eliteprox (changes requested):
Great work @ad-astra-video! I tested `-maxPricePerCapability` with AI model pricing and it works correctly. 

I also regression tested `-maxPricePerUnit` to ensure price controls still work with transcoding. Using the H264 capability as the default price index is a creative approach that should allow us to set prices per codec one day in the future. 

I tested `livepeer_cli` with both options 16, 15 and found the price changes also work correctly there. I made a few edits to `setMaxPriceForCapability` to prevent a runtime error and updated the test `TestMaxPricePerCapability`, see my suggestions below.

Overall, changes LGTM!

server/handlers.go
ad-astra-video: Agree

cmd/livepeer/starter/starter_test.go
ad-astra-video: Agree

server/handlers.go
ad-astra-video: Agree

server/handlers.go
ad-astra-video: Agree

server/handlers.go
ad-astra-video: Agree

server/handlers.go
ad-astra-video: Agree

ad-astra-video:
@eliteprox @rickstaa I agree with all suggestions and committed the suggested changes.

eliteprox (approved):
Changes LGTM. Approved pending review from @rickstaa

cmd/livepeer/starter/starter.go
```diff
@@ -1768,6 +1790,61 @@ func getGatewayPrices(gatewayPrices string) []GatewayPrice {
...
+	if err != nil {
+		glog.Errorf("model prices could not be parsed: %s", err)
+		return nil
+	}
+
+	prices := make([]ModelPrice, len(pricesSet.CapabilitiesPrices))
+	for i, p := range pricesSet.CapabilitiesPrices {
+		if p.Gateway == "" {
+			p.Gateway = "default"
+		}
```
victorges: What is this field used for if, from what I understood, this is a config on the gateway itself? I thought this would make sense on an orchestrator config to set prices for specific gateways instead.

cmd/livepeer/starter/starter.go
```diff
@@ -1768,6 +1790,61 @@ func getGatewayPrices(gatewayPrices string) []GatewayPrice {
...
+	err := json.Unmarshal([]byte(pricesFileContent), &pricesSet)
+	if err != nil {
+		glog.Errorf("model prices could not be parsed: %s", err)
+		return nil
+	}
+
+	prices := make([]ModelPrice, len(pricesSet.CapabilitiesPrices))
+	for i, p := range pricesSet.CapabilitiesPrices {
+		if p.Gateway == "" {
+			p.Gateway = "default"
```
victorges: **nit:** Could be nice to support this defaulting for the `model_id` field as well. Then one can specify only the pipeline for it to be the default price for the pipeline

core/ai.go
```diff
@@ -41,6 +42,28 @@ func (s JSONRat) String() string {
...
+	}
+	pipelineName := string(runes)
+	for cap, desc := range CapabilityNameLookup {
+		if pipelineName == desc {
+			return cap
+		}
+	}
+
+	//no capability description matches name
+	return Capability_Unused
```
victorges: I wonder if this should be an explicit error instead (or the caller check this `Unused` value) to avoid silent bad configurations. Otherwise one might be missing a configuration that they swear they are making on the ir JSON config file which could have bad implications (e.g. paying more than they were willing to). In general I think it's better to fail explicitly for bad configs than to ignore them.

victorges (approved):

cmd/livepeer/starter/starter.go
ad-astra-video: Removed Gateway field for now. There was a time that this logic was going to be re-used for the Orchestrator as well to set price per gateway/pipeline/model_id but I think we found a way to re-use the aiModels.json config parsing to set price per gateway at the Orchestrator. Can always add it back in if needed.  

Confirmed test still passed.

core/ai.go
ad-astra-video: I agree that it would be better to update this to return `(Capability, error)` to make it more obvious it should be checked by the caller.

Updated and added a test for this function in ai_test.go

cmd/livepeer/starter/starter.go
ad-astra-video: This is a good idea. I removed the p.Gateway check and changed it to setting p.ModelID to 'default' if not set specifically.

ad-astra-video:
@victorges @rickstaa updates made for review comments.  Small merge conflict on a log line that looked the exact same resolved as well.

rickstaa:
@ad-astra-video the code looks good 🚀. However I was unable to build this on my system 🤔. Could you check if you pushed the latest changes?

```bash
GO111MODULE=on CGO_ENABLED=1 CC="" CGO_CFLAGS="" CGO_LDFLAGS=" " go build -o "./" -tags "mainnet" -ldflags="-X github.com/livepeer/go-livepeer/core.LivepeerVersion=0.7.6-cdd74620" cmd/livepeer/*.go
# github.com/livepeer/go-livepeer/server
server/ai_session.go:206:3: undefined: Models
server/ai_session.go:208:29: capabilityConstraints[cap].Models undefined (type *"github.com/livepeer/go-livepeer/core".PerCapabilityConstraints has no field or method Models)
server/ai_session.go:209:14: undefined: core.NewCapabilitiesWithConstraints
server/broadcast.go:102:50: netCaps.CapabilityConstraints undefined (type *"github.com/livepeer/go-livepeer/net".Capabilities has no field or method CapabilityConstraints)
make: *** [Makefile:100: livepeer] Error 1
```

ad-astra-video:
@rickstaa updated and re-tested this.

docker image build available at `docker pull adastravideo/go-livepeer:ai-video-set-max-price-per-capability-rebase`

cmd/livepeer/livepeer.go
```diff
@@ -135,6 +135,7 @@ func parseLivepeerConfig() starter.LivepeerConfig {
 	cfg.OrchPerfStatsURL = flag.String("orchPerfStatsUrl", *cfg.OrchPerfStatsURL, "URL of Orchestrator Performance Stream Tester")
 	cfg.Region = flag.String("region", *cfg.Region, "Region in which a broadcaster is deployed; used to select the region while using the orchestrator's performance stats")
 	cfg.MaxPricePerUnit = flag.String("maxPricePerUnit", *cfg.MaxPricePerUnit, "The maximum transcoding price per 'pixelsPerUnit' a broadcaster is willing to accept. If not set explicitly, broadcaster is willing to accept ANY price. Can be specified in wei or a custom currency in the format <price><currency> (e.g. 0.50USD). When using a custom currency, a corresponding price feed must be configured with -priceFeedAddr")
+	cfg.MaxPricePerCapability = flag.String("maxPricePerCapability", *cfg.MaxPricePerCapability, `json list of prices per capability/model or path to json config file. Use "model_id": "default" to price all models in a pipeline the same. Example: {"capabilities_prices": [{"pipeline": "text-to-image", "model_id": "stabilityai/sd-turbo", "priceperunit": 1000, "pixelsperunit": 1}, {"pipeline": "upscale", "model_id": "default", priceperunit": 1200, "pixelsperunit": 1}]}`)
```
rickstaa: Like the format! Do we want to add the `currency` field in there as well as an example?

rickstaa: Can we change this to snake case to keep the JSON convention. For this we can also change the `PricePerGateway` in this format (see https://linear.app/livepeer-ai/issue/AI-574/change-priceperbroadcaster-json-input-keys-to-recommended-conventions).

monitor/census.go
```diff
@@ -164,17 +164,17 @@ type (
...
+		mTicketsRecv                   *stats.Int64Measure
+		mPaymentRecvErr                *stats.Int64Measure
+		mWinningTicketsRecv            *stats.Int64Measure
+		mValueRedeemed                 *stats.Float64Measure
+		mTicketRedemptionError         *stats.Int64Measure
+		mSuggestedGasPrice             *stats.Float64Measure
+		mMinGasPrice                   *stats.Float64Measure
+		mMaxGasPrice                   *stats.Float64Measure
+		mTranscodingPrice              *stats.Float64Measure
+		mTranscodingPricePerCapability *stats.Float64Measure
```
rickstaa: Can we name it `PricePerCapability`?

server/broadcast.go
```diff
@@ -56,34 +56,98 @@ var submitMultiSession = func(ctx context.Context, sess *BroadcastSession, seg *
...
+
+	//if no prices set per model, return maxPrice
+	if price.Cmp(big.NewRat(0, 1)) == 0 {
+		price = cfg.MaxPrice()
+	}
+
+	return price
+}
+
+func (cfg *BroadcastConfig) GetCapabilityMaxPrice(cap core.Capability, modelID string) *big.Rat {
```
rickstaa: This one can be lowercase since it is not used outside this file :).

cmd/livepeer/livepeer.go
ad-astra-video: Updated. Will update pricePerGateway to same in separate PR.

monitor/census.go
ad-astra-video: Updated

server/broadcast.go
ad-astra-video: Updated

server/broadcast.go
```diff
@@ -56,8 +56,19 @@ var submitMultiSession = func(ctx context.Context, sess *BroadcastSession, seg *
...
+	mu                    sync.RWMutex
+}
+
+func newBroadcastConfig() *BroadcastConfig {
+	models := make(map[string]*core.AutoConvertedPrice)
+	models["default"] = core.NewFixedPrice(big.NewRat(0, 1))
+
+	return &BroadcastConfig{
+		maxPricePerCapability: map[core.Capability]map[string]*core.AutoConvertedPrice{
+			core.Capability_H264: models,
```
rickstaa: As discussed in our chat this can be core.Capability_Unused.

ad-astra-video: Updated

cmd/livepeer/starter/starter.go
```diff
@@ -1810,6 +1837,58 @@ func getGatewayPrices(gatewayPrices string) []GatewayPrice {
 	return prices
 }
 
+type ModelPrice struct {
+	Gateway       string
```
rickstaa: I think this field is a left over of another feature 🤔? Couldn't see it used.

rickstaa (approved):
LGTM thanks!

PR #3127, opened 2024-08-12
Add flag 'ignoreMaxPriceIfNeeded'

leszko:
This flag decides what to do when all Os have a price above the `maxPricePerPixel` defined by B:
- `true` => Use an O with the price above `maxPricePerPixel`
- `false` => Do not use any O (do not work)


cmd/livepeer/starter/starter.go
```diff
@@ -204,6 +205,7 @@ func DefaultLivepeerConfig() LivepeerConfig {
 	defaultMaxTotalEV := "20000000000000"
 	defaultDepositMultiplier := 1
 	defaultMaxPricePerUnit := "0"
+	defaultEnforceMaxPrice := false
```
victorges: I think conceptually it would make more sense for the max price to be enforced by default (equivalent to having `true` here [1]). That because the non-enforcing behavior can be very misleading for independent node operators, being a behavior that only really makes sense for Studio right now that prefers to overpay than to deny service. On the Studio side, we can toggle the flag in our deployment easily enough.

[1] Even though we could just change the default to `true`, it is always cumbersome to have boolean flags that default to true since you can't disable them as easily as enabling (e.g. just `--enforce-max-price`), and sometimes you can't disable them at all (mist haha). So my suggestion is to have an "inverse" flag here, which actually means enabling the feature of ignoring the max price if no Os match. WDYT of `--ignore-max-price-if-needed`, defaulting to `false`, or sth like that?

victorges (approved):
LGTM either way but I have a soft preference for making the default behavior to enforce the max price

cmd/livepeer/starter/starter.go
leszko: > I think conceptually it would make more sense for the max price to be enforced by default (equivalent to having `true` here [1]). That because the non-enforcing behavior can be very misleading for independent node operators, being a behavior that only really makes sense for Studio right now that prefers to overpay than to deny service. On the Studio side, we can toggle the flag in our deployment easily enough.
> 
> [1] Even though we could just change the default to `true`, it is always cumbersome to have boolean flags that default to true since you can't disable them as easily as enabling (e.g. just `--enforce-max-price`), and sometimes you can't disable them at all (mist haha). So my suggestion is to have an "inverse" flag here, which actually means enabling the feature of ignoring the max price if no Os match. WDYT of `--ignore-max-price-if-needed`, defaulting to `false`, or sth like that?

I'm ok with this. @rickstaa wdyt?

thomshutt: I agree with Victor's reasoning here

rickstaa: @thomshutt I already discussed this with @victorges async and I am onboard with the suggested changes. Feel free to merge with the suggested changes 👍. Thanks for creating this pull request.

leszko: Updated. @thomshutt please approve the PR and I'll merge

thomshutt (approved):

cmd/livepeer/starter/starter.go
```diff
@@ -294,6 +296,7 @@ func DefaultLivepeerConfig() LivepeerConfig {
 		MaxTotalEV:              &defaultMaxTotalEV,
 		DepositMultiplier:       &defaultDepositMultiplier,
 		MaxPricePerUnit:         &defaultMaxPricePerUnit,
+		IgnoreMaxPriceIfNeeded:  &defaultEnforceMaxPrice,
```
rickstaa: @leszko looks good to me thanks! However is there a reason the default parameter name was not refactored while the test was?

rickstaa (changes requested):

cmd/livepeer/starter/starter.go
leszko: Ohh, sorry. Updated now. Also updated the PR description. PTAL.

6b3f95de0bc8a6af0376b5610e775b486e2c7c91

rickstaa (approved):

cmd/livepeer/starter/starter.go
rickstaa: No problem. Approved!

PR #3128, opened 2024-08-13
workflows: Fix issue when github context is empty for merged commits

hjpotter92:
It seems that the github context `github.event.pull_request` is null for merged PRs. This led to build failures for `master` branch (and would subsequently also fail for releases etc.)

From [github docs](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request):

> The `pull_request` webhook event payload is empty for merged pull requests and pull requests that come from forked repositories.

---

I've updated the workflow to make use of the full `github` context and split based on the type of event that triggers it. :crossed_fingers:

### DIFF
--- a/.github/discord-embed-webhook.py
+++ b/.github/discord-embed-webhook.py
@@ -2,6 +2,7 @@ import argparse
 import os
 import json
 import pathlib
+from typing import Any
 
 from requests import Response
 from discord_webhook import DiscordEmbed, DiscordWebhook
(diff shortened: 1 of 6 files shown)


leszko (approved):

pwilczynskiclearcode (approved):

.github/discord-embed-webhook.py
```diff
@@ -13,6 +14,27 @@
...
+    context_vars = {}
+    event: dict[str, Any] = context.get("event")
+    if context.get("event_name") == "pull_request":
+        pull_request_context: dict[str, Any] = event.get("pull_request", {})
+        head = pull_request_context.get("head", {})
+        repo = head.get("repo", {})
+        sha = head.get("sha")
+        context_vars["title"] = head.get("ref")
+        context_vars["sha"] = sha
+        context_vars["commit_url"] = f'{repo.get("html_url")}/commit/{sha}'
```
pwilczynskiclearcode, line 27: I think that `.get()` is overused in many places instead of using `[]`.  If `html_url` would not exist then this would evaluate to sth like `None/commit/324jdsan`
So I think we want explicit failure if it doesn't exist:
```python
context_vars["commit_url"] = f'{repo["html_url"]}/commit/{sha}'
```

or few lines above. Instead of "silently" passing:
```python
sha = head.get("sha")
```
it's better to write:
```
sha = head["sha"]
```
which would fail if `sha` was missing. And I guess we always expect some `sha`

hjpotter92: hmm, since we're parsing github context data; we should be sure that for specific `event_name`, the payload would be consistent with github's docs https://docs.github.com/en/webhooks/webhook-events-and-payloads#pull_request

PR #3131, opened 2024-08-15
Segment anything pipeline image

pschroedl:

Adds support for the ai-worker pipeline which implements Segment Anything 2 from facebook/meta
 (https://github.com/facebookresearch/segment-anything-2)

Corresponding AI-Worker PR:
https://github.com/livepeer/ai-worker/pull/185

~~* Adds a mapping of pipelines to a tag corresponding to the pipeline-specific docker image~~
Mapping of pipelines to a tag corresponding to the pipeline-specific docker image has been pushed down to the ai-worker
* Launches a container with the correct image for pipelines specified warm == True in aiModels.json
* Pricing for semgent-anything-2 requests

E2E testing gateway + orchestrator via curl and jupyter notebook

Example usage: 
https://colab.research.google.com/drive/12TC3NW7BpO77WfKHQyok18ebOtbT9F4-?usp=sharing

**Checklist:**


core/capabilities.go
```diff
@@ -114,6 +115,7 @@ var CapabilityNameLookup = map[Capability]string{
 	Capability_ImageToVideo:               "Image to video",
 	Capability_Upscale:                    "Upscale",
 	Capability_AudioToText:                "Audio to text",
```
rickstaa, line 117: @pschroedl For now, I think it's perfectly fine to add this as a hardcoded capability 👍🏻. In the future, we might want to explore making this mechanism more generic, but let's save that challenge for another time 😅.

cmd/livepeer/starter/starter.go
```diff
@@ -1227,10 +1227,27 @@ func StartLivepeer(ctx context.Context, cfg LivepeerConfig) {
...
+					// If config.Warm is true, setup the AI worker with the selected Docker image
+					if config.Warm && config.URL == "" {
+						// Determine which image to use based on the pipeline
+						imageToUse := *cfg.AIRunnerImage // Default to the image from config
+						if specificImage, ok := core.PipelineToImage[config.Pipeline]; ok {
+							imageToUse = specificImage
+						}
+
+						// Setup the AI worker with the selected Docker image
+						n.AIWorker, err = worker.NewWorker(imageToUse, gpus, modelsDir)
```
rickstaa: @pschroedll Unfortunately, this logic won't work when starting multiple containers. Here's what I suggest we do instead:

- Rename the [containerImageID](https://github.com/pschroedl/ai-worker/blob/50fcc14304dc4b6d7ea8a7cf055b32a35d9d5768/worker/docker.go#L43) to `containerImageIDDefault`.
- Add an optional `containerImageID` parameter to the [function](https://vscode.dev/github/pschroedl/ai-worker/blob/segment_anything_2_pipeline_image/worker/docker.go#L157) and set it in the container config when present.
- Apply the same logic to [docker.go/DockerManager/Borrow](https://github.com/pschroedl/ai-worker/blob/50fcc14304dc4b6d7ea8a7cf055b32a35d9d5768/worker/docker.go#L110).
- Apply the same logic to [docker.go/DockerManager/Warm](https://github.com/pschroedl/ai-worker/blob/50fcc14304dc4b6d7ea8a7cf055b32a35d9d5768/worker/docker.go#L79).
- This will allow us to simplify the logic in the starter as well.

Let me know your thoughts!

cmd/livepeer/starter/starter.go
```diff
@@ -1227,10 +1227,27 @@ func StartLivepeer(ctx context.Context, cfg LivepeerConfig) {
...
-				// If the config contains a URL we call Warm() anyway because AIWorker will just register
-				// the endpoint for an external container
 				if config.Warm || config.URL != "" {
+					// Register the external endpoint if URL is provided
 					endpoint := worker.RunnerEndpoint{URL: config.URL, Token: config.Token}
+
+					// If config.Warm is true, setup the AI worker with the selected Docker image
+					if config.Warm && config.URL == "" {
+						// Determine which image to use based on the pipeline
+						imageToUse := *cfg.AIRunnerImage // Default to the image from config
```
rickstaa: @pschroedl can we replace this with a helperfunction like `getContainerImageID`?

rickstaa (changes requested):
@pschroedll Please check out my requested changed. If those are dealt with it is ready for release.

cmd/livepeer/starter/starter.go
pschroedl: Sounds like a good approach to me - thanks @rickstaa !

cmd/livepeer/starter/starter.go
pschroedl: OK, SGTM.

server/ai_process.go
```diff
@@ -586,6 +587,102 @@ func submitUpscale(ctx context.Context, params aiRequestParams, sess *AISession,
...
+		if monitor.Enabled {
+			monitor.AIRequestError(err.Error(), "segment anything 2", *req.ModelId, sess.OrchestratorInfo)
+		}
+		return nil, err
+	}
+	defer completeBalanceUpdate(sess.BroadcastSession, balUpdate)
+
+	// start := time.Now()
+	resp, err := client.SegmentAnything2WithBodyWithResponse(ctx, mw.FormDataContentType(), &buf, setHeaders)
+	// took := time.Since(start)
```
rickstaa: @pschroedl I enabled this again in https://github.com/livepeer/go-livepeer/pull/3131/commits/6d3ce7315f5485852f233bc8d34719d45cb25c16. Was something wrong with the calculation?

pschroedl: I had only intended to disable the timer until the calculation method had been updated - good catch!

cmd/livepeer/starter/starter.go
pschroedl: After testing this approach, I realized that a better approach might be to delegate custom image mapping down to the AI-Worker. 

One motivation is that Borrow is also called directly from ai-worker itself, leading to a state where the the custom containerImageID name:tag was unavailable.

rickstaa (approved):
LGTM 🚀 !

rickstaa:
@pschroedl, only thing left is updating the worker version before merge.

rickstaa:
@pschroedl although some tests pass these are the same as the ones on the `ai-video` main branch. We still need to fix those and write new tests for the AI stuff. Let's prioritize this after the public launch.

PR #3137, opened 2024-08-16
workflows: Revert to ubuntu-20 to stop selecting self-hosted runners

hjpotter92:
Hopefully it'll stop selecting self-hosted runners for jobs that don't really need them. Ideally, it should be fixed in github's runner selection, but we've no clear route on how to achieve it yet.

### DIFF
--- a/.github/requirements.txt
+++ b/.github/requirements.txt
@@ -1,0 +1,1 @@
+discord-webhook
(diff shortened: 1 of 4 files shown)


rickstaa (approved):
Thanks 🙏🏻!

PR #3156, opened 2024-08-28
validate `-serviceAddr` on startup and cli handler

eliteprox:

This change aims to bring awareness to orchestrators that setting `0.0.0.0` or any other non-public IP as the `-serviceAddr` flag will cause issues with receiving and processing work. This will also help improve network quality.

- Logs a warning on startup if `-serviceAddr` contains `0.0.0.0`
- Returns a validation error in `livepeer_cli` option 13: Set orchestrator config

- Tested orchestrator on chain with https://0.0.0.0:8935 as `-serviceAddr`, confirm warning shows on startup
```
W0828 19:31:05.124127  733748 starter.go:1204] **Warning -serviceAddr is a not a public address or hostname; this is not recommended for onchain networks**
```

- Tested setting orchestrator config with livepeer_cli, option 13, confirm validation error is returned:
```
Enter the public host:port of node (default: https://127.0.0.1:8935)> http://0.0.0.0:8935
Error applying configuration: service address must be a public IP address or hostname
```

AI-586

**Checklist:**

### DIFF
--- a/common/util.go
+++ b/common/util.go
@@ -11,6 +11,7 @@ import (
 	"math/big"
 	"math/rand"
 	"mime"
+	"net/url"
 	"regexp"
 	"sort"
 	"strconv"
@@ -529,4 +530,12 @@ func ParseEthAddr(strJsonKey string) (string, error) {
 		}
 	}
 	return "", errors.New("Error parsing address from keyfile")
+}
+
+func ValidateServiceURI(serviceURI *url.URL) bool {
+	if strings.Contains(serviceURI.Host, "0.0.0.0") {
+		return false
+	} else {
+		return true
+	}
 }
--- a/common/util.go
+++ b/common/util.go
@@ -533,9 +533,5 @@ func ParseEthAddr(strJsonKey string) (string, error) {
 }
 
 func ValidateServiceURI(serviceURI *url.URL) bool {
-	if strings.Contains(serviceURI.Host, "0.0.0.0") {
-		return false
-	} else {
-		return true
-	}
+	return !strings.Contains(serviceURI.Host, "0.0.0.0")
 }
--- a/cmd/livepeer/starter/starter.go
+++ b/cmd/livepeer/starter/starter.go
@@ -1200,8 +1200,8 @@ func StartLivepeer(ctx context.Context, cfg LivepeerConfig) {
 			glog.Exit("Error getting service URI: ", err)
 		}
 
-		if *cfg.Network == "arbitrum-one-mainnet" && strings.Contains(suri.Host, "0.0.0.0") {
-			glog.Exit("Service address must be a public IP address or hostname when operating on mainnet")
+		if *cfg.Network != "offchain" && common.ValidateServiceURI(suri) {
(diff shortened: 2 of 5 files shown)


common/util.go
```diff
@@ -530,3 +531,11 @@ func ParseEthAddr(strJsonKey string) (string, error) {
 	}
 	return "", errors.New("Error parsing address from keyfile")
 }
+
+func ValidateServiceURI(serviceURI *url.URL) bool {
+	if strings.Contains(serviceURI.Host, "0.0.0.0") {
```
leszko: Doesn't `0.0.0.0` mean "expose to ALL network interfaces"? In other words, I think that `0.0.0.0` is a valid public address configuration.

eliteprox: It is a valid in the context that the Orchestrator can bind to it and receive it's own ping request. However, it appears that GetOrchestrator requests will cache this IP and it can result in streams not being processed correctly when an orchestrator is selected. 

I appreciate the feedback, perhaps we can look closer at the GetOrchestrator requests to return a public IP instead of the IP defined here. This change only adds a warning to startup and prevents setting the `-serviceAddr` to `0.0.0.0` in livepeer_cli so it shouldn't cause any breaking changes, and I do not see a case where an orchestrator would use this value onchain due to the current behavior of the gateway's routing

leszko: Ok, I think you're right

common/util.go
```diff
@@ -530,3 +531,11 @@ func ParseEthAddr(strJsonKey string) (string, error) {
...
+		return false
+	} else {
+		return true
+	}
+}
```
leszko:
```suggestion
func ValidateServiceURI(serviceURI *url.URL) bool {
	return !strings.Contains(serviceURI.Host, "0.0.0.0")
}
```

leszko (approved):
Added one minor comment, other than that LGTM

leszko:
I'll wait for the CI build and if it's fine, then I'll merge

Strykar:
@eliteprox @leszko Sorry I'm late to this, I'm traveling.
This and https://github.com/livepeer/go-livepeer/pull/3165 are sweet!

Why accept private IP ranges at all with `onchain`, allow them only for `offchain` instead?
That flag is never expected to work with a private IP, the networking is left up to the Orchestrator to manage right?

Warning the user to use a public IP immediately followed by an example of  `(default: https://127.0.0.1:8935)` seems contradictory.

instead something like:
```
Enter your FQDN hostname:port of the node (example: https://my.orch.net:8935)> http://0.0.0.0:8935
Error service address configuration: in onchain network mode this must be a hostname that resolves to a public IP address.
```

Would've shipped a PR for the error messages, but wanted to know how you felt about using Go `net` to filter by private and public ranges depending on which `-network` was set?

eliteprox:
@Strykar Good suggestion! I agree the log message should be improved. I'd also like to change this from a warning to an error, 

The main concern I have heard is it still common for developers to test on-chain scenarios locally with arbitrum-one-mainnet network instead of a testnet, so for now we decided to go with a warning only

PR #3163, opened 2024-09-05
Clean up inactive sessions

leszko:
Not cleaning up the sessions causes quite a big memory leak which in effect causes OOMKilled

fix https://linear.app/livepeer/issue/PS-726/5xx-broadcasters-alerts-prod-livepeer-broadcaster-08gow


j0sh:
The broadcaster has its own media session timeout right? Could that be tied to the PM session? Just wondering, no real opinion on the PR ... as long as the sessions timeout for media/pm are somewhat synced, the approach seems fine 👍

PR #3165, opened 2024-09-06
Add node version and orch addr to transcoded metadata

j0sh:
Adds metadata that incorporates the node version and orchestrator address. This makes it easier to determine whether a given video came from the Livepeer network, and if so, who processed it.

For mpegts, the `service_provider` metadata field is used; for mp4 it is the `comment` field. Both are added unconditionally before passing on to the transcoder, but the muxer will actually only pick one or the other.

The actual transcoder version is used, rather than the orchestrator version to accommodate split O/T. If the network is offchain, the orchestrator address will say "offchain"

This metadata is taken as-is and not verified; transcoders _could_ modify this with a custom build if they really wanted to. We could just insert the orch addr metadata from the broadcaster itself, but we would not have the transcoder version.

```
josh@josh-gpu:~/bench$ ffprobe -hide_banner http://localhost:8935/stream/stream/P240p30fps16x9/10.ts
Input #0, mpegts, from 'http://localhost:8935/stream/stream/P240p30fps16x9/10.ts':
  Duration: N/A, start: 101.400000, bitrate: N/A
  Program 1
    Metadata:
      service_name    : Service01
      service_provider: Livepeer Transcoder 0.7.8-48a92a17-dirty (offchain)
  Stream #0:0[0x100]: Video: h264 (Main) ([27][0][0][0] / 0x001B), yuv420p(progressive), 426x240 [SAR 1:1 DAR 71:40], 30 fps, 30 tbr, 90k tbn
  Stream #0:1[0x101]: Audio: aac (LC) ([15][0][0][0] / 0x000F), 48000 Hz, stereo, fltp, 163 kb/s
```

- Store the PM recipient address (orch's onchain address) in the LivepeerNode
- Adds a new `orchId` field to the O -> T protobuf definitions to carry the RecipientAddr to the transcoder
- Set the orchId to "offchain" for off-chain transcoding. Use RecipientAddr otherwise.
- Add a new helper, `MakeMetadata` for use both with local O transcoding and split O+T
- Update LPMS for metadata support - https://github.com/livepeer/lpms/pull/421
- Regenerate the protobuf definitions - I could not get an older version of protoc installed, so ended up with this mess of a commit : 3fa2382fbd55adfdd5abc67d63afea1e129c327b

Manual testing with combined OT and split O+T

**Checklist:**


net/lp_rpc.proto
```diff
@@ -340,6 +340,9 @@ message NotifySegment {
     // ID for this particular transcoding task.
     int64 taskId   = 16;
 
+    // Orchestrator identifier for segment metadata
+    string orchId  = 18;
```
leszko, line 344: nit: shouldn't you move it under profiles, so that the indexing is in the order?

leszko (approved):
The PR looks good. Very nice change @j0sh 

I have just one product question (CC: @thomshutt @hthillman ), I wonder if there is any case in which we would **not** want to expose that the video is coming from Livepeer Network. I remember that we had some customers with the special DNS domain to not expose it's coming from Livepeer. Not sure it's valid anymore or/and if there would be any issue exposing it in the video segment metadata.

thomshutt:
> I remember that we had some customers with the special DNS domain to not expose it's coming from Livepeer. 

Good point Rafal, but I think it's buried enough that we don't mind Livepeer's name being attached here - even for the other stuff it was more of a "best effort" approach to anonymity.

thomshutt (approved):
@j0sh looks good! Is there any other metadata (timestamp?) to hand that would be useful to add while we're in here?

net/lp_rpc.proto
j0sh: Isn't `profiles` deprecated?

j0sh:
>  Is there any other metadata (timestamp?) to hand that would be useful to add while we're in here?

Yep, there is a _lot_ we could collect (someone in Discord suggested the GPU type, etc) but I am not sure how much we really want to stuff into this one field. It was mostly meant to scratch my own itch around not being able to figure out where segments came from.

Can definitely add the timestamp if it seems like it would be near-term useful, though. UTC or Unix? I also wonder if there is a better 'place' in mpegts we can put things like the creation date, but google and some quick tests are not turning up anything...

PR #3166, opened 2024-09-09
Clean up inactive sessions

leszko:
Not cleaning up the sessions causes quite a big memory leak which in effect causes OOMKilled

It's an alternative approach to https://github.com/livepeer/go-livepeer/pull/3163 in which, instead of keeping the timestamp of the last usage, we add cleanup into the flow.

Note that I'm 100% we won't have some corner cases, like segment was returned with a delay and now we cannot find the session. Saying that, it should not crash anything. Also, I did quite a lot of local dev chain tests, so I think it should work fine. 

My plan is to get it merged and then deploy to one region for some time to see if we see any regression.


mjh1 (approved):

mjh1:
Simple enough change at least 👍 but yep sounds good on monitoring one region first

PR #3106, opened 2024-07-27
 Remote AI Worker (Livepool version)

kyriediculous:
**What does this pull request do? Explain your changes. (required)**
This PR adds a Remote AI worker to the go-livepeer repo, integrated on the _transcoder_. Its control flow and networking layer are akin to that of the remote O+T setup. 

It allows an Orchestrator to run in standalone mode, and dispatch AI jobs to connected standalone _transcoders_ in which the AI worker is integrated. Remote AI workers register to one or more orchestrators over _gRPC_ and receive a one-directional stream in return, on which AI job notifications are received. The transcoder runs an [AI worker and runner](https://github.com/livepeer/ai-worker) to perform the tasks. Once a job is completed, remote AI workers will _POST_ the results back to the orchestrator.

# Remote Orchestrator+AI Worker Specification

## Abstract
This document outlines the specification for integrating a remote AI worker setup with an Orchestrator, akin to the existing remote Orchestrator + Transcoder configuration. The goal is to enable dynamic orchestrator capabilities while maintaining flexibility for other potential use cases, such as public pools.

image

## Architecture

The architecture mirrors that of the remote transcoder setup for an orchestrator. Instead of hosting AI workers locally, the orchestrator will function as a message broker and session manager, delegating tasks to remote workers.

image

### Design Goals
- Minimize changes/additions to the codebase
- Adhere to go-livepeer conventions and design principles
- Ensure maximum flexibility for various use cases

### O<>W Wire Protocol
The communication between the orchestrator (O) and remote AI workers, integrated in the transcoder node (W), will use gRPC to register workers and notify them of new AI jobs over an open stream. The orchestrator will also run an HTTP server to receive results from remote workers. This approach aligns with the existing remote transcoding setup, facilitating ease of understanding and maintenance. Using the same networking stack eliminates the need for additional servers or microservices.

One huge benefit of this design is that the remote AI worker doesn't have to run a server and purely acts as a gRPC and HTTP RPC client. This means that the host wouldn't need to worry about port forwarding, which is of significant importance for public pools.

#### gRPC 
**`RegisterAIWorker`**
The existing `Transcoder` service will be extended with a new RPC method `RegisterAIWorker` to register remote AI workers, specifying their capabilities, and maintaining an open connection to receive AI tasks.

```go 
service Transcoder {
    ...
  // Called by the transcoder to register a `RemoteAIWorker` worker to an orchestrator
  // notifies the registered `RemoteAIWorker` of AI jobs as they come in.
[...]

PR #3168, opened 2024-09-10
Implementation of AI Remote Worker (AI-323) (rebased)

ad-astra-video:
See PR 3088 rebased to current `ai-video` branch and adding segment anything 2 pipeline.  This was a community effort in the same way PR 3088 was with contributions from multiple contributors in the ecosystem which I am very grateful for.  Detail from the commits of PR 3088 were squashed in the rebase to make it cleaner and easier to complete.  Credit for the contributions to implementing the remote ai worker will be included on the squashed commit on PR approval 

- See PR 3088 and tech spec that was updated after rebase completed
- Tests were added for `core` and `server` parts of the ai-worker additions.
- Implementation is backwards compatible with current gateways while laying the foundation to upgrade to passing json job status/results back separate from binary outputs (e.g. images or videos). 

Current ai-worker has been used on subnet by myself and Pon.  I serve I2I, T2I, I2V and upscale models using 5-6 separate ai-workers on mainnet and have completed over 5,000 requests.

**Checklist:**

- [>>] All tests in `./test.sh` pass    >>`tests in core/ai_test.go and server/ai_worker_test.go pass`


cmd/livepeer/starter/starter.go
```diff
@@ -1505,13 +1465,24 @@ func StartLivepeer(ctx context.Context, cfg LivepeerConfig) {
 		}
 	} else if n.NodeType == core.TranscoderNode {
 		*cfg.CliAddr = defaultAddr(*cfg.CliAddr, "127.0.0.1", TranscoderCliPort)
+	} else if n.NodeType == core.AIWorkerNode {
+		*cfg.CliAddr = defaultAddr(*cfg.CliAddr, "127.0.0.1", AIWorkerCliPort)
+		//need to have default Capabilities if not running transcoder
+		if !*cfg.Transcoder {
+			aiCaps = append(aiCaps, core.DefaultCapabilities()...)
```
leszko, line 1475: I think the default capabilities include transcoding, so in this scenario, we add transcoding capabilities when the node is not a transcoder. Shouldn't it be the opposite?

common/util.go
```diff
@@ -565,3 +566,12 @@ func CalculateAudioDuration(audio types.File) (int64, error) {
 
 	return duration, nil
 }
+
+func ExtensionByType(contentType string) (string, error) {
```
leszko, line 576: Could you add unit tests for this?

core/ai_test.go
```diff
@@ -18,3 +28,665 @@ func TestPipelineToCapability(t *testing.T) {
...
+	time.Sleep(1 * time.Millisecond) // allow the workers to activate
+
+	hasCapacity = o.CheckAICapacity("text-to-image", "livepeer/model1")
+	assert.True(t, hasCapacity)
+
+	// Test when remote AI worker does not have capacity
+	hasCapacity = o.CheckAICapacity("text-to-image", "livepeer/model2")
+	assert.False(t, hasCapacity)
+}
+func TestRemoteAIWorkerProcessPipelines(t *testing.T) {
```
leszko, line 502: This test fails for me, could you check?

core/ai_worker.go
```diff
@@ -0,0 +1,991 @@
...
+		if monitor.Enabled {
+			monitor.AIResultSaveError(ctx, "audio-to-text", *req.ModelId, string(monitor.SegmentUploadErrorUnknown))
+		}
+		return nil, err
+	}
+
+	return res.Results, nil
+}
+
+func (orch *orchestrator) SegmentAnything2(ctx context.Context, requestID string, req worker.SegmentAnything2MultipartRequestBody) (interface{}, error) {
```
leszko: I see that for every new pipeline we'll need to add a new function here (and also in `livepeer/ai-runner`. Would it maybe be possible to do it in an abstract way that adding new pipeline would not need any update here?

I see it's similar in Nico's code: https://github.com/livepeer/go-livepeer/pull/3106/files?diff=unified&w=1#diff-85b4611f2c22c73db96c87db4253da7c72df9107f7b4c545b740f88ce2fda2e5R256
But at least it's a little shorter.

core/capabilities.go
```diff
@@ -757,3 +759,38 @@ func (bcast *Capabilities) MinVersionConstraint() string {
 	}
 	return ""
 }
+
+func (c *Constraints) AddCapabilityConstraints(cap Capability, constraint CapabilityConstraints) {
```
leszko: Where is this function used?

core/capabilities.go
```diff
@@ -757,3 +759,38 @@ func (bcast *Capabilities) MinVersionConstraint() string {
...
+			} else {
+				c.perCapability[cap].Models[modelID] = modelConstraint
+			}
+		} else {
+			c.perCapability[cap] = &CapabilityConstraints{Models: make(ModelConstraints)}
+		}
+	}
+}
+
+func (c *Constraints) RemoveCapabilityConstraints(cap Capability, constraint CapabilityConstraints) {
```
leszko: Where is this function used?

net/lp_rpc.proto
```diff
@@ -368,6 +376,37 @@ message NotifySegment {
...
 }
 
+// Sent by the aiworker to register itself to the orchestrator.
+message RegisterAIWorkerRequest {
+
+    // Shared secret for auth
+    string secret = 1;
+
+    // AIWorker capacity
+    int64 capacity = 2;
```
leszko: Do we need this if we already express the capacity in the model (line 132)?

net/lp_rpc.proto
```diff
@@ -368,6 +376,37 @@ message NotifySegment {
...
+
+    // AIWorker capabilities
+    Capabilities capabilities = 3;
+}
+
+// Sent by the orchestrator to the aiworker
+message NotifyAIJob {
+    // URL of the input to use.
+    string url      = 1;
```
leszko: Do we need this? I see that we don't have any URL in Nico's PR: https://github.com/livepeer/go-livepeer/pull/3106/files?diff=unified&w=1#diff-034e29c79f9ba82705c2ecd90a418ec22feb053f18f84244244a54b995964adaR382

net/lp_rpc.proto
```diff
@@ -368,6 +376,37 @@ message NotifySegment {
...
+
+    // pipeline to use for the job
+    string pipeline = 2;
+
+    // model to process job with
+    string modelID = 3;
```
leszko: Do we need this I see that we don't have modelID in Nico's PR: https://github.com/livepeer/go-livepeer/pull/3106/files?diff=unified&w=1#diff-034e29c79f9ba82705c2ecd90a418ec22feb053f18f84244244a54b995964adaR382

server/ai_http.go
```diff
@@ -371,6 +381,29 @@ func handleAIRequest(ctx context.Context, w http.ResponseWriter, r *http.Request
 		return
 	}
 
+	//backwards compatibility to old gateway api
```
leszko, line 427: Do we actually need it? 

If yes, could you add a longer comment to describe more about it?

server/ai_http.go
```diff
@@ -411,3 +444,127 @@ func handleAIRequest(ctx context.Context, w http.ResponseWriter, r *http.Request
 	w.WriteHeader(http.StatusOK)
 	_ = json.NewEncoder(w).Encode(resp)
 }
+
+//
+// Orchestrator receiving results from the remote AI worker
+//
+
+func (h *lphttp) AIResults() http.Handler {
```
leszko, line 531: Any chance we could have unit tests for this one?

server/ai_process.go
```diff
@@ -33,6 +33,10 @@ const defaultUpscaleModelID = "stabilityai/stable-diffusion-x4-upscaler"
 const defaultAudioToTextModelID = "openai/whisper-large-v3"
 const defaultSegmentAnything2ModelID = "facebook/sam2-hiera-large"
 
+var errWrongFormat = fmt.Errorf("result not in correct format")
+
+var downloadResult = core.GetSegmentData
```
leszko: Nit: why does it need to be a var? 

I think there is no reason right now, we may need it for some unit tests.

server/ai_process.go
```diff
@@ -81,19 +85,34 @@ func processTextToImage(ctx context.Context, params aiRequestParams, req worker.
...
 	newMedia := make([]worker.Media, len(imgResp.Images))
 	for i, media := range imgResp.Images {
+		var result []byte
 		var data bytes.Buffer
+		var name string
 		writer := bufio.NewWriter(&data)
-		if err := worker.ReadImageB64DataUrl(media.Url, writer); err != nil {
-			return nil, err
+		err := worker.ReadImageB64DataUrl(media.Url, writer)
+		if err == nil {
```
leszko, line 119: How is this change related to Remote Worker? I can't connect the dots.
Have we started sending now the image in the URL instead of Base64encoded?

leszko (commented):
Thanks a lot for the PR @ad-astra-video 

1. I created a PR with some super minor improvements, please check and merge https://github.com/ad-astra-video/go-livepeer/pull/31
2. I added some inline comments, PTAL
3. There are a few other questions/comments, but I'll start a thread in Discord to discuss them (you don't need to reply here), mainly about:
  - Transcoder Worker vs AI Worker => Would it maybe make sense to merge them?
  - Code duplication between Transcoder/AI Workers => Do we want to refactor it or maybe we don't want to touch the transcoding part at all, just copy the code and focus on AI?
  - Adding a new pipeline will require now (even) more code changes => Our strategy to address it
  - Co-authoring this PR from Nico => Since you both created similar PR, I think I'd be good to merge one of them, but assign you both as authors

cmd/livepeer/starter/starter.go
ad-astra-video: We need the DefaultCapabilities to be discoverable by Gateways and only using CPU H264 for AI workflows at the moment so the DefaultCapabilities cover this requirement.  I don't think the default capabilities are included elsewhere for an ai-worker node.

common/util.go
ad-astra-video: Yes will add.

server/ai_process.go
ad-astra-video: This is copied implementation from transcoding side, https://github.com/livepeer/go-livepeer/blob/7ae8e2ce41eb2e46223b10aab47aa6bfab9eed73/server/broadcast.go#L51

Reusing `downloadSeg` seemed odd for the result download naming. Happy to update to any suggested naming.

server/ai_process.go
ad-astra-video: The ai-worker runner returns base64 encoded results in the `.Url`field.  In the implementation now the orchestrator recieved the base64 encoded results and sends it up to the gateway to decode and save in NodeStorage session.

I thought passing base64 encoded results from remote ai worker > orchestrator > gateway was not ideal so I updated to using similar implementation as transcoding that uses node storage and passes urls.
 
This is part of the backwards compatibility so gateways can process both results from orchestrators that have not updated and orchestrators using the updated implementation.  Some of this can be removed after a period of time to let orchestrators update.  Or could just force the update on to orchestrators to enable not needing some of this.

net/lp_rpc.proto
ad-astra-video: Nico's version uses a enum for pipeline that will need to be updated with each new pipeline and a data field that i think encodes the modelID into it. The data field also carries all the binary data for the request I believe which puts a cap on the input file size at 4mb (gRPC server size limit)

net/lp_rpc.proto
ad-astra-video: Correct, it is not used. I was hesitent to remove it until we got farther through review in case there can be a use for it in addition to managing capacity per model.

net/lp_rpc.proto
ad-astra-video: `url` is the node storage url passed down to the worker to download the input file. I think Nico's version passes all the data including input files in the `data` field.

server/ai_http.go
ad-astra-video: Will work on adding today.

core/ai_worker.go
ad-astra-video: That is correct and is from the initial architecture of the AI pipelines.  My understanding is that the main driver of this is the different return types and request data structs (including request input file names).  This could pivot to something like `handleAIRequest` but that just trades separate functions for a longer function.

If ai-worker was updated to have a higher level input request struct and results return type I think the go-livepeer side of things could be generalized more.   

Nico's version is shorter because it passes the base64 encoded result up to the gateway which is then decoded by the gateway and saved to storage to provide a url to download to the client.

core/capabilities.go
ad-astra-video: Only in tests right now to enable testing the capability constraints.  Should I move it to a non exported function in the capabilities_test.go file?

I think it was used at one point in implementing this but pivoted to a different way of adding capabilities from the worker.

core/capabilities.go
ad-astra-video: Same as AddCapabilityConstraints above.

core/ai_test.go
ad-astra-video: fixed, needed to create node storage for the request id and update expected return url

leszko: Thanks, it works! 🙏

core/ai_worker.go
leszko: ok, fine. We can leave it for now and revisit if we start having many more AI Worker endpoints (as discussed in Discord).

core/capabilities.go
leszko: If it's only for tests, then please move it to `capabilitiers_test.go`

core/capabilities.go
leszko: The same, if only for tests, then could you move it to `capabilities_test.go` (and make it package-private as well).

net/lp_rpc.proto
leszko: If it's not used, then please remove it. It's always simple to add a new value to proto def if needed.

net/lp_rpc.proto
leszko: Do I understand correctly that it's used only for some pipelines (for the pipelines that requires input URL). For example, image-to-image may need it while text-to-image does not need it?

If that's the case, then I think I like the more generic approach and having everything in the `requestData` (or you can just call it `data`). Otherwise, we may have many different fields that are related only to specific pipelines.

net/lp_rpc.proto
leszko: Isn't Nico's version enum => pipeline in your PR? Then, Nico probably passes the modelID in the `data` field. Is that correct?

For me, the comment is similar to `url`, I think the field is fine if it's related to every pipeline. If only to some of them, then I'd hide it inside `requestData`.

server/ai_http.go
leszko: Thanks 🙏

server/ai_process.go
leszko: Could you just rename `GetSegmentData` to `DownloadResult` and get rid of this `var downloadResult` at all?

server/ai_process.go
leszko: Ok, I see. It makes sense. 

Tagging @thomshutt , because it's somehow related to the issue with returning large data from SAM2 (I know your team is working on that).

thomshutt: >I thought passing base64 encoded results from remote ai worker > orchestrator > gateway was not ideal

I might well be missing something, but is the new flow `ai worker > orchestrator` and then the gateway pulls asynchronously from the orchestrator?

core/capabilities.go
ad-astra-video: moved

server/ai_process.go
ad-astra-video: The ai-worker decodes the base64 result from the ai-runner and then posts a multipart result to the orchestrator `/aiResults` endpoint.  For binary results (image and video), the orchestrator then saves the result in node storage and provides a url for the gateway to download the result from.  This is the workflow that transcoding uses and I thought would be good to start to separate out the result when it is binary.  JSON responses from `audio-to-text` and `segment-anything-2` pass through without this intermediate step to split out the result since they are not base64 encoded.  With the experience from SAM2 and the large responses, wondering if all pipelines should separate the result from the response to the request to generate.  Thoughts?

leszko: IMO it would nice to have the same flow for all models. And I believe this does not add much to the overall latency. @thomshutt Do you have any opinion on that?

thomshutt: Agree, I think that splitting out the response from the reponse data to aid in dealing with large responses should probably come in its own PR and should apply to all models for consistency.

common/util.go
ad-astra-video: I added a test for `image/png` which is the only used extension for this.  Future will look at runner returning multipart/form-data response so can have actual file included.

Note `video/mp4 returns` as .f4v.

net/lp_rpc.proto
ad-astra-video: I pull out modelID so the ai-worker can reserve the GPU before type checking the `requestData`.  modelID is included in all requests though. Can I get it without type checking? Maybe just convert it to json?

Link to request structs created with codegen https://github.com/livepeer/ai-worker/blob/ddda616fe0e266a60ab92f6263cefb1d7598e6c4/worker/runner.gen.go#L45

server/ai_http.go
ad-astra-video: Added ai_http_test.go file. Some of the unit tests were in ai_worker_test.go.  Also added a RunAIJob to test parsing the multipart/mixed result sent to the `AIResults` handler.  I still need to add a test for the SSE response on the LLM pipeline but not sure how to do that at the moment.  Have any suggestions on implementing that?

server/ai_process.go
ad-astra-video: I renamed the GetSegmentData function.  It created a number of changes in the transcoding side as well.  This is in a separate commit.

server/ai_process.go
ad-astra-video: Ok, I agree. To confirm, don't undo what I already have. Just punt on adding things like SSE to other pipelines to a new PR right? Or saying put the text based responses in URLs to download as well?  I have reservations doing this on the LLM pipeline to be honest.

ad-astra-video:
@leszko i merged my working branch into this one.  Attaching first set of test logs.  Need to rebuild docker container and test more tomorrow. Want to test gateway and orchestrator versions 0.7.8-ai.2 to confirm backwards compatibility.


[a2t-new-gateway-new-orch-sep-ai-worker.txt](https://github.com/user-attachments/files/17240514/a2t-new-gateway-new-orch-sep-ai-worker.txt)
[i2i-new-gateway-new-orch-sep-ai-worker.txt](https://github.com/user-attachments/files/17240515/i2i-new-gateway-new-orch-sep-ai-worker.txt)
[llm-new-gateway-new-orch-sep-ai-worker.txt](https://github.com/user-attachments/files/17240517/llm-new-gateway-new-orch-sep-ai-worker.txt)
[sam2-new-gateway-new-orch-sep-ai-worker.txt](https://github.com/user-attachments/files/17240518/sam2-new-gateway-new-orch-sep-ai-worker.txt)
[t2i-new-gateway-new-orch-sep-ai-worker.txt](https://github.com/user-attachments/files/17240519/t2i-new-gateway-new-orch-sep-ai-worker.txt)
[upscale-new-gateway-new-orch-sep-ai-worker.txt](https://github.com/user-attachments/files/17240520/upscale-new-gateway-new-orch-sep-ai-worker.txt)
[i2v-new-gateway-new-orch-sep-ai-worker.txt](https://github.com/user-attachments/files/17240541/i2v-new-gateway-new-orch-sep-ai-worker.txt)

net/lp_rpc.proto
leszko: I see, so you're using this `modelID` here:
https://github.com/livepeer/go-livepeer/blob/139f0347983febbf3c522e14f79498273255a006/server/ai_worker.go#L161

Before parsing `requestData` here:
https://github.com/livepeer/go-livepeer/blob/139f0347983febbf3c522e14f79498273255a006/server/ai_worker.go#L188

IMO parsing request data should be fast, so not really an optimization. Actually it may even work worse, because if the requestData is not parsed correctly, we don't want to allocate GPU.

All in all, if you prefer to have `modelID` in the proto, I'm ok with that as long as we're sure that `modelID` makes sense for each request (which I think is true). However, if `url` is not always present then I'd definitely include it in `requestData`.

server/ai_process.go
leszko: I think it's ok to do what you suggested, meaning "don't undo what you have".

server/ai_http.go
leszko: Not sure, I think if it's difficult to add, we can live without it for now. We can always add tests for LLM SSE part later.

core/ai_worker.go
```diff
@@ -0,0 +1,1019 @@
...
+}
+
+func (n *LivepeerNode) createStorageForRequest(requestID string) error {
+	n.storageMutex.Lock()
+	defer n.storageMutex.Unlock()
+	_, exists := n.StorageConfigs[requestID]
+	if !exists {
+		os := drivers.NodeStorage.NewSession(requestID)
+		n.StorageConfigs[requestID] = &transcodeConfig{OS: os, LocalOS: os}
+		// TODO: Figure out a better way to end the OS session after a timeout than creating a new goroutine per request?
```
leszko, line 808: @ad-astra-video What about these TODOs, do you plan to solve them in this PR?

leszko (commented):
@ad-astra-video Thanks for all the changes and rebasing to `ai-video`. I think it's mostly good to go.

1. Please check if there are no more open comments in this PR
2. I recommend removing this `modelID` and `url` from proto (please check my inline comment for more details)
3. Please do your tests with your O (I only played with this locally)
4. I see some tests failing in CI after you changes (I merged your PR with my "fix tests" PR and [here](https://github.com/livepeer/go-livepeer/actions/runs/11161973065/job/31025626042?pr=3198) is the results). It's mainly about:
    - Your changes broke [`broadcast_test.go` ](https://github.com/livepeer/go-livepeer/actions/runs/11161973065/job/31025626042?pr=3198#step:17:182)
    - There is some hanging goroutine (probably related to `// TODO: ` comments in `ai_worker.go`. [Here](https://github.com/livepeer/go-livepeer/actions/runs/11161973065/job/31025626042?pr=3198#step:17:2356) are the failing tests; for this if we don't have some better solution, we can add [an exception to testutil.go](https://github.com/livepeer/go-livepeer/blob/1b9da7f25fa8c6261e8c465ea7e5c75166d220b6/common/testutil.go#L91)

core/ai_worker.go
ad-astra-video: This was a carryover from implementation with the image-to-video pipeline.  We could be more explicit with this and end the session when the file is pulled but I think would be better to consider in a separate PR.

leszko: ok, it makes sense. Let's leave it for now

net/lp_rpc.proto
ad-astra-video: I have updated to add `AIJobData` to proto file similar to `SegData` so that the specifics of the request are grouped together in the `NotifyAIJob` message sent to the ai-worker.  I had to update a number of the tests to account for this and confirmed all are passing.

The Orchestrator verifies the request received with the open api validation so it should be good once it gets sent to the ai-worker (see route registrations in server/ai_http.go). The type checking at the ai-worker level is mostly to correctly parse the request to the auto generated types from ai-worker repo and confirm nothing went weird in transit.

server/ai_process.go
ad-astra-video: I made some updates to transcoding side to add back in the `downloadSeg` var where needed for tests to pass.

server/ai_http.go
ad-astra-video: This is now line 427. Expanded comment.  The main update made here is the gateway also adds the `protoVerAIWorker` to the request headers sent to the Orchestrator.  We are iterating on communication protocols with the AI jobs so I thought it would be good to establish a compatibility check and adjust the response as needed for a short while.  

This can be removed after a short update window (1-2 week?) after this PR is merged if would prefer to not have a backwards compatibility check in Gateway > Orchestrator requests.

net/lp_rpc.proto
ad-astra-video: This has been updated and moved to AIJobData in proto file

server/ai_http.go
ad-astra-video: Agree, will add AI SPE linear issue to follow up adding SSE test for LLM.

ad-astra-video:
> @ad-astra-video Thanks for all the changes and rebasing to `ai-video`. I think it's mostly good to go.
> 
> 1. Please check if there are no more open comments in this PR

I think all are responded to now.

> 2. I recommend removing this `modelID` and `url` from proto (please check my inline comment for more details)

I repsonded to inline comment. Moved to AIJobData in proto file so the NotifyAIJob only has TaskId and AIJobData.

> 3. Please do your tests with your O (I only played with this locally)

I re-tested locally all pipelines and confirmed working after latest updates.
> 4. I see some tests failing in CI after you changes (I merged your PR with my "fix tests" PR and [here](https://github.com/livepeer/go-livepeer/actions/runs/11161973065/job/31025626042?pr=3198) is the results). It's mainly about:
>    
>    * Your changes broke [`broadcast_test.go` ](https://github.com/livepeer/go-livepeer/actions/runs/11161973065/job/31025626042?pr=3198#step:17:182)

This was from removing _all_ of the `downloadSeg` references and moving that var to broadcast_test.go.  The tests override this function to provide a mock of the segment download.  I moved this back to broadcast.go and tested the failing tests in link above to confirm they passed.  Can you re-run in runner to confirm?

>    * There is some hanging goroutine (probably related to `// TODO: ` comments in `ai_worker.go`. [Here](https://github.com/livepeer/go-livepeer/actions/runs/11161973065/job/31025626042?pr=3198#step:17:2356) are the failing tests; for this if we don't have some better solution, we can add [an exception to testutil.go](https://github.com/livepeer/go-livepeer/blob/1b9da7f25fa8c6261e8c465ea7e5c75166d220b6/common/testutil.go#L91)

I confirmed this is from the image-to-video test in `TestRunAIJob`.  If I update the `aiWorkerResultsTimeout` to an exported variable from core package I can set the timeout to something really short and then get the `logRoutine` lingering go routines that you added to `testutil.go`.  Wasn't really sure it was appropriate to export the `aiWorkerResultsTimeout` variable right now and I also don't have a better solution at the moment.  I added as an item to follow up on in the AI SPE linear.

ad-astra-video:
I worked through more tests over the weekend and fixed a couple things. I tested on chain and offchain for combined Orch/ai-worker and separate ai-worker.  

Fixed some passing of the channels in the separate ai-worker to match the combined where a receive only channels is passed from the ai-worker on the SSE response from the llm pipeline.  Also fixed the image-to-video pipeline on combined Orch/ai-worker.

I will deploy to my Orchestrator in the morning in central US time.  @rickstaa would you be able to run some tests against the latest remote ai worker?

net/lp_rpc.proto
leszko: I still see this `url` field, shouldn't we remote it?

leszko (commented):
Added one more comment, other than that it's good to go. 

@rickstaa Do you want to review or/and test it? Or should we proceed with merging this PR?

net/lp_rpc.proto
ad-astra-video: That is in AIJobData now. https://github.com/livepeer/go-livepeer/blob/dabc0e8db9f95cedbfd4a15de444e44566ba481e/net/lp_rpc.proto#L394

I guess this comment history followed the field since it is the first field?

The requestData is an exact copy of the request received that matches to the ai-worker types. I wanted to keep that field like that so a added the AIJobData shell to hold the url and pipeline fields.

leszko (approved):

common/util.go
ad-astra-video: Closing out this conversation. The ExtensionByType function was updated to set extensions directly using a switch block instead of the `mime` package because the `mime` package returns multiple options for `video/mp4`.

net/lp_rpc.proto
```diff
@@ -371,6 +379,38 @@ message NotifySegment {
...
 }
 
+// Sent by the aiworker to register itself to the orchestrator.
+message RegisterAIWorkerRequest {
+
+    // Shared secret for auth
+    string secret = 1;
+
+    // AIWorker capabilities
+    Capabilities capabilities = 3;
```
rickstaa: @ad-astra-video shouldn't this be:

```suggestion
    Capabilities capabilities = 2;
```

net/lp_rpc.proto
rickstaa: @leszko is this resolved?

core/livepeernode.go
```diff
@@ -116,7 +118,10 @@ type LivepeerNode struct {
 	Database *common.DB
 
 	// AI worker public fields
-	AIWorker AI
+	AIWorker        AI
+	AIWorkerManager *RemoteAIWorkerManager
+	// AI private fields
+	aiJobMutex *sync.Mutex
```
rickstaa: @ad-astra-video this seems unused?

net/lp_rpc.proto
leszko: TBH I'd prefer to put this `url` into `requestData`, because AFAIU not every request has url

net/lp_rpc.proto
ad-astra-video: updated

core/livepeernode.go
ad-astra-video: removed

net/lp_rpc.proto
ad-astra-video: I have moved the `url` to the requestData field.

server/ai_worker.go
```diff
@@ -0,0 +1,532 @@
...
+			return core.NewRemoteAIWorkerFatalError(errZeroCapacity)
+		}
+		if status.Code(err) == codes.Canceled {
+			return core.NewRemoteAIWorkerFatalError(errInterrupted)
+		}
+	}
+	return err
+}
+
+func runAIWorker(n *core.LivepeerNode, orchAddr string, capacity int, caps *net.Capabilities) error {
```
rickstaa, line 95: @ad-astra-video capacity seems to be unused.

server/ai_worker.go
```diff
@@ -0,0 +1,532 @@
...
+	}
+	uploadDur := time.Since(uploadStart)
+	clog.V(common.VERBOSE).InfofErr(ctx, "AI job processing done results sent for taskId=%d uploadDur=%v", taskId, pipeline, modelID, uploadDur, err)
+
+	if monitor.Enabled {
+		monitor.AIResultUploaded(ctx, uploadDur, pipeline, modelID, orchAddr)
+	}
+}
+
+func sendStreamingAIResult(ctx context.Context, n *core.LivepeerNode, orchAddr string, pipeline string, modelID string, httpc *http.Client,
```
rickstaa, line 482: @ad-astra-video Unused modelID and addlData.

server/ai_worker.go
```diff
@@ -0,0 +1,532 @@
...
+			taskId, resultUrl, err)
+		return
+	}
+	req.Header.Set("Authorization", protoVerAIWorker)
+	req.Header.Set("Credentials", n.OrchSecret)
+	req.Header.Set("Content-Type", contentType)
+	req.Header.Set("TaskId", taskId)
+	req.Header.Set("Pipeline", pipeline)
+
+	//TODO consider adding additional information in response header from the addlData field (e.g. transcoding includes Pixels)
```
rickstaa: @ad-astra-video related to the unsued arguments. If we need this information in the future we can add it. Please remove the TODO for now.

server/ai_worker.go
```diff
@@ -0,0 +1,532 @@
...
+		io.Copy(fw, bytes.NewBuffer(jsonResp))
+	}
+
+	w.Close()
+	contentType = "multipart/mixed; boundary=" + boundary
+	sendAIResult(ctx, n, orchAddr, notify.AIJobData.Pipeline, modelID, httpc, contentType, &body, addlResultData, nil)
+}
+
+func sendAIResult(ctx context.Context, n *core.LivepeerNode, orchAddr string, pipeline string, modelID string, httpc *http.Client,
+	contentType string, body *bytes.Buffer, addlData interface{}, err error,
```
rickstaa, line 435: @ad-astra-video Another unusedlData.

rickstaa (approved):
LGTM. We can remove some redundant code in subsequent pull requests 👍🏻.

core/ai.go
```diff
@@ -127,3 +132,102 @@ func ParseStepsFromModelID(modelID *string, defaultSteps float64) float64 {
 
 	return numInferenceSteps
 }
+
+// AddAICapabilities adds AI capabilities to the node.
+func (n *LivepeerNode) AddAICapabilities(caps *Capabilities) {
```
kyriediculous, line 137: This function should not exist

core/ai.go
```diff
@@ -127,3 +132,102 @@ func ParseStepsFromModelID(modelID *string, defaultSteps float64) float64 {
...
+				n.Capabilities.constraints.perCapability[aiCapability].Models[modelId].Capacity += modelConstraint.Capacity
+			} else {
+				n.Capabilities.constraints.perCapability[aiCapability].Models[modelId] = &ModelConstraint{Warm: modelConstraint.Warm, Capacity: modelConstraint.Capacity}
+			}
+		}
+	}
+}
+
+// RemoveAICapabilities removes AI capabilities from the node.
+func (n *LivepeerNode) RemoveAICapabilities(caps *Capabilities) {
```
kyriediculous, line 165: This function should not exist

core/ai.go
```diff
@@ -127,3 +132,102 @@ func ParseStepsFromModelID(modelID *string, defaultSteps float64) float64 {
...
+				return fmt.Errorf("failed to reserve AI capability capacity, model capacity is 0 pipeline=%v modelID=%v", pipeline, modelID)
+			}
+			return nil
+		}
+		return fmt.Errorf("failed to reserve AI capability capacity, model does not exist pipeline=%v modelID=%v", pipeline, modelID)
+	}
+	return fmt.Errorf("failed to reserve AI capability capacity, pipeline does not exist pipeline=%v modelID=%v", pipeline, modelID)
+}
+
+func (n *LivepeerNode) ReleaseAICapability(pipeline string, modelID string) error {
```
kyriediculous, line 215: This function should not exist

core/ai.go
```diff
@@ -127,3 +132,102 @@ func ParseStepsFromModelID(modelID *string, defaultSteps float64) float64 {
...
+					}
+				} else {
+					glog.Errorf("failed to remove AI capability capacity, model does not exist pipeline=%v modelID=%v", capability, modelId)
+				}
+			}
+		}
+	}
+}
+
+func (n *LivepeerNode) ReserveAICapability(pipeline string, modelID string) error {
```
kyriediculous, line 191: This function should not exist

core/ai_worker.go
```diff
@@ -0,0 +1,1054 @@
...
+
+func (rw *RemoteAIWorker) done() {
+	// select so we don't block indefinitely if there's no listener
+	select {
+	case rw.eof <- struct{}{}:
+	default:
+	}
+}
+
+type RemoteAIWorkerManager struct {
```
kyriediculous, line 50: Wrong file structure

net/lp_rpc.proto
```diff
@@ -12,6 +12,13 @@ service Orchestrator {
   rpc Ping(PingPong) returns (PingPong);
 }
 
+service AIWorker {
+
+  // Called by the aiworker to register to an orchestrator. The orchestrator
+  // notifies registered aiworkers of jobs as they come in.
+  rpc RegisterAIWorker(RegisterAIWorkerRequest) returns (stream NotifyAIJob);
```
kyriediculous, line 19: RegisterAIWorker request doesn't need to be a newly defined type

core/ai_worker.go
```diff
@@ -0,0 +1,1054 @@
...
+	"github.com/livepeer/ai-worker/worker"
+	"github.com/livepeer/go-livepeer/clog"
+	"github.com/livepeer/go-livepeer/common"
+	"github.com/livepeer/go-livepeer/monitor"
+	"github.com/livepeer/go-livepeer/net"
+	"github.com/livepeer/go-tools/drivers"
+	"github.com/livepeer/lpms/ffmpeg"
+)
+
+var ErrRemoteWorkerTimeout = errors.New("Remote worker took too long")
```
kyriediculous, line 26: Why is this exported ?

core/ai_worker.go
```diff
@@ -0,0 +1,1054 @@
...
+var ErrNoCompatibleWorkersAvailable = errors.New("no workers can process job requested")
```
kyriediculous, line 27: Why is this exported ?

core/ai_worker.go
```diff
@@ -0,0 +1,1054 @@
...
+var ErrNoWorkersAvailable = errors.New("no workers available")
```
kyriediculous, line 28: Why is this exported ?

ad-astra-video: Followed how other similar errors are implemented in core package. Linked errors and errors above could probably be changed to not be exported.

https://github.com/livepeer/go-livepeer/blob/4a66b2232188d765a7d04bcc1f5ed21a47ef67ef/core/orchestrator.go#L852

core/ai_worker.go
ad-astra-video: Followed how other similar errors are implemented in core package. Linked errors and errors above could probably be changed to not be exported.

https://github.com/livepeer/go-livepeer/blob/4a66b2232188d765a7d04bcc1f5ed21a47ef67ef/core/orchestrator.go#L852

core/ai_worker.go
ad-astra-video: Followed how other similar errors are implemented in core package. Linked errors and errors above could probably be changed to not be exported.

https://github.com/livepeer/go-livepeer/blob/4a66b2232188d765a7d04bcc1f5ed21a47ef67ef/core/orchestrator.go#L852

net/lp_rpc.proto
ad-astra-video: A new `RegisterAIWorkerRequest` was added because the generic `capacity` field is not helpful when trying to manage GPUs for AI jobs in my opinion.  For transcoding one GPU can do multiple requests at a time and there was one job type. For AI, my experience is most models slow down significantly when more than one request is fed to it concurrently. 

Do you think a generic capacity field set at launch of the ai-worker would let the orchestrator appropriately manage the ai workers?

Do you think that AI workers and remote transcoders would always have the same requirements when connecting to the orchestrator?

core/ai_worker.go
ad-astra-video: Most AI related things are in separate files right?

Can you help me understand why using `ai_worker.go` files does not help keep development on AI and transcoding from causing issues for anyone developing on one or the other?

core/ai.go
ad-astra-video: Not following, please explain. Intent here was to start managing capacity for each pipeline/modelID and also work with workers having multiple GPUs serving the same pipeline/modelID behind one ai worker.

core/ai.go
ad-astra-video: Not following, please explain. Intent here was to start managing capacity for each pipeline/modelID and also work with workers having multiple GPUs serving the same pipeline/modelID behind one ai worker.

core/ai.go
ad-astra-video: Not following. Is this not similar to `RemoveCapacity`?

You are saying it would be better to squash this into `RemoveCapacity`? Without updating tests to cover this addition into `RemoveCapacity` function?

New function was added because transcoding does not use capabilities in this way right now and tried to avoid adding complication to a function used with remote transcoder connection that didn't need to be there.

core/ai.go
ad-astra-video: Not following. Is this not similar to `AddCapacity`?

You are saying it would be better to squash this into `AddCapacity`? Without updating tests to cover this addition into `AddCapacity` function?

New function was added because transcoding does not use capabilities in this way right now and tried to avoid adding complication to a function used with remote transcoder connection that didn't need to be there.

PR #3179, opened 2024-09-21
fix: selection algo filtering on max price fixed

ad-astra-video:

Fixes behavior of gateway when no max price is set. `0` maxPrice will return to being no max price filter enabled.

The addition of maxPricePerCapability included initializing a zero max price by default so the map is fully initialized.  This broke the `filterByMaxPrice` function that checked for an unset `maxPrice` of `nil`.

- updates `filterByMaxPrice` function in `selection_algorithm.go` to check for 0 rather than checking for nil
- added test to cover this so will fail in future if behavior of max price filtering changes
- updated TestFilter test to set a maxPrice of 0 if none set

Added new test and ran all tests in selection_algorithm_test.go

**Checklist:**

- [ ] All tests in `./test.sh` pass    `X  selection_algorithm_test.go pass. `


server/selection_algorithm.go
```diff
@@ -58,7 +58,7 @@ func (sa ProbabilitySelectionAlgorithm) filterByPerfScore(ctx context.Context, a
 }
 
 func (sa ProbabilitySelectionAlgorithm) filterByMaxPrice(ctx context.Context, addrs []ethcommon.Address, maxPrice *big.Rat, prices map[ethcommon.Address]*big.Rat) []ethcommon.Address {
-	if maxPrice == nil || len(prices) == 0 {
+	if maxPrice.Num().Int64() == 0 || len(prices) == 0 {
```
thomshutt: We should probably still have a nil check here for safety

victorges: A more idiomatic way of doing this would also be
```suggestion
	if maxPrice == nil || maxPrice.Sign() == 0 || len(prices) == 0 {
```
(still need to check for `nil` tho yeah)

server/selection_algorithm_test.go
victorges, line 1: thanks for adding tests 🙏

victorges (approved):
lgtm

server/selection_algorithm.go
ad-astra-video: Agree

ad-astra-video:
I updated for suggested change. Learned something new, did not know `Sign()` did that!

PR #3183, opened 2024-09-25
Better messages for objectStore errors

mjh1:

In the case of objectStore errors it wasn't clear before this was the reason, so I've added context to the error message by wrapping the `err` objects. I also improved the logging a bit by adding capability to the logging context and added a `String` function to make it easier.

Error message now looks like:
Screenshot 2024-09-25 at 17 43 14

Tested locally simulating a broken objectStore

https://linear.app/livepeer/issue/ENG-2318/dont-throw-500s-if-objectstore-stops-working

**Checklist:**

### DIFF
--- a/core/capabilities.go
+++ b/core/capabilities.go
@@ -623,6 +623,14 @@ func CapabilityToName(capability Capability) (string, error) {
 	return capName, nil
 }
 
+func (c Capability) String() string {
+	name, err := CapabilityToName(c)
+	if err != nil {
+		return fmt.Sprintf("%d", int(c))
+	}
+	return name
+}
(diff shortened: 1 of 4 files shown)


mjh1:
@thomshutt what do you think about the generic 500? I kind of feel a 500 is fine when you can then locate the issue in the logs. I'm not sure what else we could make it, maybe a 503..

thomshutt:
Yeah, I think error code is less important vs. propagating the error up the chain and into the logs

leszko (approved):
LGTM

thomshutt (approved):

PR #3114, opened 2024-07-31
LLM pipeline with stream support

kyriediculous:
**What does this pull request do? Explain your changes. (required)**
Adds support for an LLM pipeline (see https://github.com/livepeer/ai-worker/pull/137).

The LLM pipeline returns either a stream or a final response. Both are handled over HTTP on the go-livepeer side using SSE (server sent events). 

TODO: revert ai-worker version after merge

**How did you test each of these updates (required)**
Ran manuel tests 

**Checklist:**

- [x] Read the [contribution guide](./CONTRIBUTING.md)
- [x] `make` runs successfully
- [ ] All tests in `./test.sh` pass
- [ ] README and other documentation updated
- [ ] [Pending changelog](./CHANGELOG_PENDING.md) updated

PR #3189, opened 2024-10-02
fix: correct the order of the capability list

rickstaa:
This pull request addresses an issue where the capability list was ordered incorrectly, which was introduced in https://github.com/livepeer/go-livepeer/pull/3114/files. The incorrect ordering caused a break in the SAM2 pipeline, which is now resolved.

### DIFF
--- a/core/capabilities.go
+++ b/core/capabilities.go
@@ -58,7 +58,7 @@ const (
 	Capability_ProfileH264ConstrainedHigh Capability = 10
 	Capability_GOP                        Capability = 11
 	Capability_AuthToken                  Capability = 12
-	Capability_SceneClassification        Capability = 13
+	Capability_SceneClassification        Capability = 13 // Deprecated, but can't remove because of Capability ordering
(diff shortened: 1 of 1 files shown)


core/capabilities.go
```diff
@@ -78,8 +78,8 @@ const (
 	Capability_ImageToVideo
 	Capability_Upscale
 	Capability_AudioToText
-	Capability_LLM
 	Capability_SegmentAnything2
+	Capability_LLM
```
leszko: Personally, I'd be more explicit and give each value an explicit int value. Likt it's done in protobufs. For example:
```
Capability_H264 = 0
Capability_MPEGTS = 1
....
```

leszko (approved):

core/capabilities.go
leszko: Discussed in Discord. I've pushed a commit with this change.

thomshutt (approved):

PR #3190, opened 2024-10-02
Fix unit tests and enable tests in CI for ai-video

leszko:
Changes:
- Remove all tests related to deprecated `segData.Profiles` @yondonfu Please double check we don't need it for video anymore (I've seen your commit here: https://github.com/livepeer/go-livepeer/commit/a2b194edf19377212d5831f71d6770de155ac495#diff-7157f6a8ee36dc4ada0f69ca25c9b79cbbae048dba0b4dde96270ad400d900eeL499
- Fix unit tests
- Fix lint errors
- Update golangci-lint
- Enable running tests in CI for `ai-video`


yondonfu:
Re: removing tests for the deprecated `net.SegData.Profiles` field.

I think it is fine to remove the tests and to keep the [commit](https://github.com/livepeer/go-livepeer/commit/a2b194edf19377212d5831f71d6770de155ac495#diff-7157f6a8ee36dc4ada0f69ca25c9b79cbbae048dba0b4dde96270ad400d900eeL499).

IIRC the problem I was trying to solve with that commit was that the AI requests would trigger a failure on the O for the call `common.BytesToVideoProfile(segData.Profiles)` because the B sets the [net.SegData.Profiles field](https://github.com/livepeer/go-livepeer/blob/f2e18325d9d5ad28f3835d1f8638935a09c9ac6b/core/streamdata.go#L106) to `[]byte("invalid")` (the `NetSegData()` function is used in `server.genSegCreds()` in `server.prepareAIPayment()`).

And IIRC the reason for keeping the logic for parsing the `Profiles` field was to maintain compatibility with older versions of go-livepeer - a G on an older version of go-livepeer prior to the introduction of the new profile related fields in `net.SegData` would still be able to speak with an O on a newer version of go-livepeer after the introduction of those fields. But, its been awhile and IMO at some point it makes sense to cleanup this really old deprecated field if no one is running such an old version of go-livepeer. 

I would just recommend announcing with plenty of lead time that this will be a breaking change unlikely to affect anyone, but *if* you are running a old version of go-livepeer for a G you should upgrade to a more recent one that uses the other profile fields of `net.SegData`.

.github/workflows/test.yaml
```diff
@@ -94,9 +95,9 @@ jobs:
       - name: Lint
         uses: golangci/golangci-lint-action@v4
         with:
-          version: v1.52.2
+          version: v1.61.0
           skip-pkg-cache: true
-          args: '--disable-all --enable=gofmt --enable=vet --enable=golint --deadline=4m pm verification'
+          args: '--out-format=colored-line-number --disable-all --enable=gofmt --enable=govet --enable=revive pm verification'
```
rickstaa: @leszko why do we only check the pm and verification folders?

server/segment_rpc_test.go
```diff
@@ -220,27 +220,6 @@ func TestVerifySegCreds_Duration(t *testing.T) {
 	assert.Nil(md)
 }
 
-func TestCoreSegMetadata_Profiles(t *testing.T) {
```
rickstaa, line 223: @leszko Just out of curiosity, do we no longer need to test with incorrect profiles? Was this change due to your handling of this logic inside `coreSegMetadata` in [this commit](https://github.com/livepeer/go-livepeer/commit/aa79cf5eb1888cd0ec2e7f66b0e553a2700128c6#diff-7157f6a8ee36dc4ada0f69ca25c9b79cbbae048dba0b4dde96270ad400d900eeL538)? I noticed the removal in `rpc_test.go`, but I didn’t catch it right away. Is that related?

.github/workflows/test.yaml
leszko: No idea 🙃 

Hehe, I'm just trying to make it work the same as on `master`. We can try to improve this later.

server/segment_rpc_test.go
leszko: It's related to this part: 

https://github.com/livepeer/go-livepeer/commit/a2b194edf19377212d5831f71d6770de155ac495#diff-7157f6a8ee36dc4ada0f69ca25c9b79cbbae048dba0b4dde96270ad400d900eeL499

And you can read the comment from Yondon: https://github.com/livepeer/go-livepeer/pull/3190#issuecomment-2391510209

My understanding is that this is a deprecated code with profiles. And I believe it should be safe to delete, I don't think there are many orgs running go-livepeer as Gateway. So doing an announcement should be fine.

server/selection_test.go
```diff
@@ -1,379 +0,0 @@
-package server
```
rickstaa: @leszko why were these tests removed. Although the tests done in here are quite trivial they still are related to the main logic in the `selection.go` file and not the new `selection_algorithm.go`.

rickstaa (approved):
Had some questions but apart from that looks good to me.

.github/workflows/test.yaml
rickstaa: Fine with me 👍🏻 . Created https://linear.app/livepeer/issue/ECO-1/enable-linting-on-all-go-livepeer-folders might we want to enable it for all folders. I already checked and the amount of lint errors is low.

server/selection_test.go
leszko: There should not be removed. I'll take them back after merging this one: https://github.com/livepeer/go-livepeer/pull/3188

Otherwise, they fail 🙃

server/broadcast.go
```diff
@@ -63,7 +63,6 @@ type BroadcastConfig struct {
 func newBroadcastConfig() *BroadcastConfig {
 	maxPrices := make(map[core.Capability]map[string]*core.AutoConvertedPrice)
 	models := make(map[string]*core.AutoConvertedPrice)
-	models["default"] = core.NewFixedPrice(big.NewRat(0, 1))
```
rickstaa: @leszko I'm not entirely sure about this removal. At first glance, it seems more intuative to have an empty map and return `nil` if the value has not been set yet for a given model. @ad-astra-video was there a specific reason you had to initialise models["default"] with a given price?

server/selection_test.go
rickstaa: No problem thanks!

server/segment_rpc_test.go
rickstaa: @leszko Good catch! I overlooked that—so while it's deprecated for backward compatibility, the field isn't marked as deprecated in the protobuf definition (see [this line](https://github.com/livepeer/go-livepeer/blob/cb4f93b1a3b86ce7ad39eeea031be342f93ebb3e/net/lp_rpc.proto#L185)). Given how long it's been, I agree we can safely remove the tests for it. 👍🏻

PR #3196, opened 2024-10-03
Allow running Orchestrator in External Container mode

leszko:


cmd/livepeer/starter/starter.go
```diff
@@ -1525,8 +1525,8 @@ func StartLivepeer(ctx context.Context, cfg LivepeerConfig) {
 		// if http addr is not provided, listen to all ifaces
 		// take the port to listen to from the service URI
 		*cfg.HttpAddr = defaultAddr(*cfg.HttpAddr, "", n.GetServiceURI().Port())
-		if !*cfg.Transcoder && n.OrchSecret == "" {
-			glog.Exit("Running an orchestrator requires an -orchSecret for standalone mode or -transcoder for orchestrator+transcoder mode")
+		if !*cfg.Transcoder && !*cfg.AIWorker && n.OrchSecret == "" {
+			glog.Info("Running an orchestrator in AI External Container mode")
```
pwilczynskiclearcode: I might be wrong, but shouldn't it be sth like:
```golang
	if !*cfg.Transcoder && !*cfg.AIWorker {
             if  *cfg.AIModels && n.OrchSecret == "" {
	        glog.Info("Running an orchestrator in AI External Container mode")
              } else {
	        glog.Exit("Running an orchestrator requires an -orchSecret for standalone mode or -transcoder for orchestrator+transcoder mode")
            }
        }
```

pwilczynskiclearcode (approved):

cmd/livepeer/starter/starter.go
leszko: Yes, that's better. Updated.

PR #3197, opened 2024-10-03
Update install_ffmpeg.sh

leszko:


j0sh (approved):
Technically, I think the latest commit for install_ffmpeg is actually at https://raw.githubusercontent.com/livepeer/lpms/bd7f8b07600446f018eab6075c22601aa3301bf6/install_ffmpeg.sh but this should be fine too

j0sh (approved):
LGTM - thanks for this fix! Might want to merge / squash https://github.com/livepeer/lpms/pull/425 first and use the updated sha for that

rickstaa (approved):

rickstaa (approved):
LGTM but looks like binary builds are still failing even with action retry.

PR #3199, opened 2024-10-06
[AI] Parler Text to Speech

pschroedl:

Adds support for the ai-worker pipeline which implements the Parler Text-to-Speech model (https://huggingface.co/parler-tts/parler-tts-large-v1)

Corresponding AI-Worker PR:
https://github.com/livepeer/ai-worker/pull/219

- Adds the new /text-to-speech endpoint and handlers
- binding methods to support ai worker
- Pricing for text-to-speech requests by characters

E2E testing gateway + orchestrator binaries via curl and jupyter notebook

Example usage:
https://colab.research.google.com/drive/12UN65fV6q44IyMRL2tyKFal8OMCyyRMN?usp=sharing

```
curl -X POST "http://0.0.0.0:47907/text-to-speech" \
    -H "Content-Type: application/json" \
    -d '{
        "model_id":"parler-tts/parler-tts-large-v1",
        "text_input":"A cool cat on the beach",
        "description":"Jons voice is monotone yet slightly fast in delivery, with a very close recording that almost has no background noise."
    }'
```

**Checklist:**

- [x]  Read the [[contribution guide](https://www.notion.so/CONTRIBUTING.md)](https://www.notion.so/CONTRIBUTING.md)
- [x]  `make` runs successfully
- [ ]  All tests in `./test.sh` pass
- [ ]  README and other documentation updated
- [ ]  [[Pending changelog](https://www.notion.so/CHANGELOG_PENDING.md)](https://www.notion.so/CHANGELOG_PENDING.md) updated


core/capabilities.go
```diff
@@ -117,6 +118,7 @@ var CapabilityNameLookup = map[Capability]string{
 	Capability_AudioToText:                "Audio to text",
 	Capability_SegmentAnything2:           "Segment anything 2",
 	Capability_LLM:                        "Large language model",
+	Capability_TextToSpeech:               "Text to speech",
```
thomshutt, line 126: Should we call this TextToAudio to make it more generatic and match up with the other naming?

server/ai_process.go
```diff
@@ -706,6 +707,84 @@ func submitSegmentAnything2(ctx context.Context, params aiRequestParams, sess *A
 	return resp.JSON200, nil
 }
 
+// TODO: CalculateTextToSpeechLatencyScore should be updated to reflect length of input tokens/output file size
```
thomshutt: Are we going to do this as part of this PR or as a followup?

core/capabilities.go
pschroedl: @rickstaa and I had a discussion about this previously and ended up on Text to Speech for two primary reasons:

1) The functionality implemented has been referred to as text-to-speech for a long time and is immediately clear to most people what the expected output will be
2) TextToAudio would better naming fit for a pipeline for [stable audio](https://huggingface.co/stabilityai/stable-audio-open-1.0) and other music-gen models

From a more architectural standpoint while at the moment it requires more boilerplate in go-bindings and in go-livepeer, it seems to me that it is more ideal to implement isolated and functionality specific pipelines - That is, more endpoints being preferable to switching implementations in the routes/pipelines based on matching one or more model_ids.

Sometimes it is hard to know where to draw the line, but I'm leaning towards the qualifications for an endpoint to share use of multiple models to have at least one of the following:
* nearly identical input schema
* ability to interchange models without significant implementation differences and dependencies
* the differences in output is more subjective and or qualitative, and less use-case specific.

Sorry to be long winded here, but happy to consider alternative perspectives.

server/ai_process.go
pschroedl: The equivalent to output pixels for the TextToSpeech pipeline is input characters ( pricing per character ), so I think we can remove this comment as it is outdated.

In lieu of a comment to that effect, I think I will change the parameter name: `outPixels` -> `inCharacters`.

go.mod
```diff
@@ -238,3 +238,5 @@ require (
 	lukechampine.com/blake3 v1.2.1 // indirect
 	rsc.io/tmplfunc v0.0.3 // indirect
 )
+
+replace github.com/livepeer/ai-worker => /home/administrator/ai-worker
```
rickstaa: Remove before merge.

go.sum
```diff
@@ -623,8 +623,6 @@ github.com/libp2p/go-netroute v0.2.0 h1:0FpsbsvuSnAhXFnCY0VLFbJOzaK0VnP0r1QT/o4n
 github.com/libp2p/go-netroute v0.2.0/go.mod h1:Vio7LTzZ+6hoT4CMZi5/6CpY3Snzh2vgZhWgxMNwlQI=
 github.com/libp2p/go-openssl v0.1.0 h1:LBkKEcUv6vtZIQLVTegAil8jbNpJErQ9AnT+bWV+Ooo=
 github.com/libp2p/go-openssl v0.1.0/go.mod h1:OiOxwPpL3n4xlenjx2h7AwSGaFSC/KZvf6gNdOBQMtc=
```
rickstaa, line 606: Update after local overwrite removal.

rickstaa (approved):
I made some small changes but looks ready for a merge 🚀.

PR #3200, opened 2024-10-09
Go livepeer liveportrait

JJassonn69:

> This pull request implements the go-livepeer side code of the LivePortrait runner. 

- 
- 
- 

> The updates were tested using the local orchestrator and broadcaster nodes running in devnet. 

The request is very similar to other pipelines in the network. Only difference is the source_image and driving_video for the pipeline. Since the models used are not very specific "default" is used instead. 

> curl -X POST "http://127.0.0.1:8935/live-portrait" -F model_id="default" -F source_image=@/Users/s0.jpg -F driving_video=@/Users/d6_trim.mp4 

This PR closes the bounty proposal for [LivePortrait](https://github.com/livepeer/bounties/issues/58/).

**Checklist:**


core/orchestrator.go
```diff
@@ -630,19 +634,24 @@ func (n *LivepeerNode) transcodeFrames(ctx context.Context, sessionID string, ur
 
 	// We only support base64 png data urls right now
 	// We will want to support HTTP and file urls later on as well
-	dirPath := path.Join(n.WorkDir, "input", sessionID+"_"+string(RandomManifestID()))
+	dirPath := path.Join(WorkDir, "input", sessionID+"_"+string(RandomManifestID()))
```
pschroedl: Was this change intentional?  I don't see WorkDir defined in this scope.

JJassonn69: it looks like a typo i missed while testing locally changing the WorkDir to local directory and reverting it back to WorkDir without the prefix n. didnt throw an error while final testing, so i didnt catch it. Just tested it again with live-portrait and image-to-video just be sure and yeah it didnt show any issue.

JJassonn69: changed it back to the original after testing it again with the pipelines `live-portrait and image-to-video`

PR #3206, opened 2024-10-15
feat(rewardService): add reward call retries

rickstaa:

This pull request refactors the RewardService to create a retry worker with exponential backoff for retrying Reward calls. Retries are limited to a specified number and only attempted for the intended round, ensuring efficiency and avoiding redundant calls.

- Update the retry logic to retry failed rewards.
- Updates tests to check the new code.
- Add a new `RewardRetryTimes` parameter for people to set the number of retries.

I wrote tests. I still have to test them on a real orchestrator.

No.

**Checklist:**

### DIFF
--- a/eth/rewardservice.go
+++ b/eth/rewardservice.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"sync"
+	"time"
 
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/golang/glog"
@@ -16,17 +17,20 @@ var (
 )
 
 type RewardService struct {
-	client       LivepeerEthClient
-	working      bool
-	cancelWorker context.CancelFunc
-	tw           timeWatcher
-	mu           sync.Mutex
+	client              LivepeerEthClient
+	working             bool
+	cancelWorker        context.CancelFunc
+	tw                  timeWatcher
+	mu                  sync.Mutex
+	rewardRetryInterval time.Duration
+	cancelRetryWorker   context.CancelFunc
 }
 
-func NewRewardService(client LivepeerEthClient, tw timeWatcher) *RewardService {
+func NewRewardService(client LivepeerEthClient, tw timeWatcher, rewardRetryInterval time.Duration) *RewardService {
 	return &RewardService{
-		client: client,
-		tw:     tw,
+		client:              client,
+		tw:                  tw,
+		rewardRetryInterval: rewardRetryInterval,
 	}
 }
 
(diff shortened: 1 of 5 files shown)


rickstaa:
Just me doing some experimentation.

ad-astra-video:
Couple quick suggestions:
- I would change `rewardRetryInterval` to something that limits the amount of retries (maybe `rewardRetryTimes`).  I would not want an unbounded amount of retries possible.  L2 gas fees are cheap but could drain the ETH from the wallet slowly.
- Use a single interval like 13-26 seconds to retry reward call in a new ethereum L1 block to get past gas fees on the current block
- I can see a case to have both flags for retry interval and number of retries to do to give more control how often and how many times retries are tried.

rickstaa:
Some concerns given by @thomshutt:
> Looks really good, my only concern would be that the retries are unbounded. I think I'd actually prefer to just have it retry X (5?) times by default with a reasonable interval rather than exposing it as yet another parameter, since I don't think people gain much from tweaking the time between retries

eth/rewardservice.go
```diff
@@ -85,6 +94,25 @@ func (s *RewardService) IsWorking() bool {
 	return s.working
 }
 
+func (s *RewardService) startNewWorker(ctx context.Context, round int64) {
```
leszko, line 97: You could consider having one loop and one goroutine instead of starting a new goroutine each time you receive a reward request. I think this would be slightly more go-style and readable as well. Anyway, also not pretty, so up to you.

```golang
s.rewardCh = make(chan struct{}, 1)
var backoffCancel context.CancelFunc

go func() {
	for {
		select {
		case <-s.rewardCh:
			// Cancel any ongoing backoff/retry loop
			if backoffCancel != nil {
				backoffCancel()
			}
			var ctx context.Context
			ctx, backoffCancel = context.WithCancel(context.Background())

			go func(ctx context.Context) {
				expBackoff := backoff.NewExponentialBackOff()
				expBackoff.InitialInterval = s.retryInterval
				expBackoff.MaxInterval = 1 * time.Hour
				expBackoff.MaxElapsedTime = s.maxElapsedTime
				backoffCtx := backoff.WithContext(expBackoff, ctx)

				for {
					err := s.tryReward()
					if err == nil {
						return
					}
					select {
					case <-backoff.After(backoffCtx):
						// continue retrying
					case <-ctx.Done():
						return
					}
				}
			}(ctx)
		}
	}
}()
```

leszko (approved):
Added one inline comment, other than that it looks good. 

Could please also?
1. Update the description given the reason why the reward actually fails from time to time? In other words, to explain why we even need a retry 🙃 
2. Test if it works 😄

PR #3209, opened 2024-10-16
ai: add models webhook url param

gioelecerati:
```
In order to avoid restarting orchestrators whenever we change the runners list we'd like to have a dynamic way of setting up the -aiModels. Similarly to [-orchWebhookUrl](https://github.com/livepeer/go-livepeer/blob/master/doc/orchwebhook.md?plain=1#L4) that can be used for providing dynamic list or orchestrators on the broadcaster side.
```

- Add `aiModelsWebhookUrl` param, accepting a string which can be either a webhook url or a file path
- Webhook gets called every 1 minute, updating the model configs
- `Webhook.GetConfigs()` returns the most recent config
- Added tests for both url and file updates

unit tests

**Checklist:**

- [X] Read the [contribution guide](./CONTRIBUTING.md)

### DIFF
--- a/cmd/livepeer/starter/starter.go
+++ b/cmd/livepeer/starter/starter.go
@@ -97,6 +97,7 @@ type LivepeerConfig struct {
 	OrchSecret              *string
 	TranscodingOptions      *string
 	AIModels                *string
+	AIModelsWebhookUrl      *string
 	MaxAttempts             *int
 	SelectRandWeight        *float64
 	SelectStakeWeight       *float64
@@ -202,6 +203,7 @@ func DefaultLivepeerConfig() LivepeerConfig {
 	defaultAIModels := ""
 	defaultAIModelsDir := ""
 	defaultAIRunnerImage := "livepeer/ai-runner:latest"
+	defaultAIModelsWebhookUrl := ""
 
 	// Onchain:
 	defaultEthAcctAddr := ""
@@ -297,10 +299,11 @@ func DefaultLivepeerConfig() LivepeerConfig {
 		TestTranscoder:       &defaultTestTranscoder,
 
 		// AI:
-		AIWorker:      &defaultAIWorker,
-		AIModels:      &defaultAIModels,
-		AIModelsDir:   &defaultAIModelsDir,
-		AIRunnerImage: &defaultAIRunnerImage,
+		AIWorker:           &defaultAIWorker,
+		AIModels:           &defaultAIModels,
+		AIModelsDir:        &defaultAIModelsDir,
+		AIRunnerImage:      &defaultAIRunnerImage,
+		AIModelsWebhookUrl: &defaultAIModelsWebhookUrl,
 
 		// Onchain:
 		EthAcctAddr:             &defaultEthAcctAddr,
@@ -1191,11 +1194,29 @@ func StartLivepeer(ctx context.Context, cfg LivepeerConfig) {
 			currencyBase = currency
 		}
 
-		if *cfg.AIModels != "" {
-			configs, err := core.ParseAIModelConfigs(*cfg.AIModels)
-			if err != nil {
-				glog.Errorf("Error parsing -aiModels: %v", err)
-				return
+		if *cfg.AIModels != "" && *cfg.AIModelsWebhookUrl != "" {
+			glog.Error("Both '-aiModels' and '-aiModelsWebhookUrl' flags are set. Please specify only one of them.")
+			return
+		} else {
+			var configs []core.AIModelConfig
+			var err error
+
+			if *cfg.AIModelsWebhookUrl != "" {
+				webhook, err := core.NewAIModelWebhook(*cfg.AIModelsWebhookUrl, 1*time.Minute)
+				if err != nil {
+					glog.Errorf("Error creating AI model webhook: %v", err)
+					return
+				}
+				configs = webhook.GetConfigs()
+				n.ModelsWebhook = webhook
+			}
+
+			if *cfg.AIModels != "" {
+				configs, err = core.ParseAIModelConfigs(*cfg.AIModels)
+				if err != nil {
+					glog.Errorf("Error parsing -aiModels: %v", err)
+					return
+				}
 			}
 
 			for _, config := range configs {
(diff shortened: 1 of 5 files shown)


core/ai.go
```diff
@@ -127,3 +133,122 @@ func ParseStepsFromModelID(modelID *string, defaultSteps float64) float64 {
...
+	return fmt.Sprintf("%x", sha256.Sum256([]byte(content)))
+}
+
+func (w *ModelsWebhook) GetConfigs() []AIModelConfig {
+	w.mu.RLock()
+	defer w.mu.RUnlock()
+	return w.configs
+}
+
+func (w *ModelsWebhook) Stop() {
```
mjh1, line 228: Do we definitely need this when it's only called by the tests?

core/ai.go
```diff
@@ -127,3 +133,122 @@ func ParseStepsFromModelID(modelID *string, defaultSteps float64) float64 {
...
+	if err != nil {
+		return nil, err
+	}
+	go webhook.startRefreshing()
+	return webhook, nil
+}
+
+func FetchAIModelConfigs(source string) ([]AIModelConfig, error) {
+	var content string
+	if strings.HasPrefix(source, "file://") {
```
mjh1: couldn't we use the `fetchContent` func here? Actually is this func `FetchAIModelConfigs` used by anything?

core/livepeernode.go
```diff
@@ -116,7 +116,8 @@ type LivepeerNode struct {
 	Database *common.DB
 
 	// AI worker public fields
-	AIWorker AI
+	AIWorker      AI
+	ModelsWebhook *ModelsWebhook
```
mjh1, line 120: I guess we don't need this field for now as it's not used right?

core/ai.go
```diff
@@ -127,3 +133,122 @@ func ParseStepsFromModelID(modelID *string, defaultSteps float64) float64 {
...
+	}
+	err := webhook.refreshConfigs()
+	if err != nil {
+		return nil, err
+	}
+	go webhook.startRefreshing()
+	return webhook, nil
+}
+
+func FetchAIModelConfigs(source string) ([]AIModelConfig, error) {
```
pwilczynskiclearcode: `FetchAIModelConfigs()` looks almost the same like `fetchContent()`. Could we reuse it somehow?

cmd/livepeer/starter/starter.go
```diff
@@ -1191,11 +1194,29 @@ func StartLivepeer(ctx context.Context, cfg LivepeerConfig) {
 			currencyBase = currency
 		}
 
-		if *cfg.AIModels != "" {
-			configs, err := core.ParseAIModelConfigs(*cfg.AIModels)
-			if err != nil {
-				glog.Errorf("Error parsing -aiModels: %v", err)
-				return
+		if *cfg.AIModels != "" && *cfg.AIModelsWebhookUrl != "" {
```
pwilczynskiclearcode, line 1197: You're missing the changes to https://github.com/livepeer/go-livepeer/blob/ai-video/cmd/livepeer/livepeer.go#L160-L161

mjh1 (commented):
Looking good, just a couple of small points 👍

core/ai.go
gioelecerati: I added it because the interval was going on after the test ended, printing several error lines. We can remove, but I think it's better to have a 3 lines method here than a messy output in the tests, wdyt?

core/ai.go
gioelecerati: Right, yeah! It's a leftover, while coding I probably remade the method, removing it

core/livepeernode.go
gioelecerati: Isn't the point of this to fetch updated model configs? I added it to be used where needed doing `updatedModels = ModelsWebhook.GetConfigs()` - but maybe @pwilczynskiclearcode knows better where this is needed to be used

core/ai.go
gioelecerati: We can remove it https://github.com/livepeer/go-livepeer/pull/3209#discussion_r1803425990

core/livepeernode.go
pwilczynskiclearcode: It's assigned [in here](https://github.com/livepeer/go-livepeer/pull/3209/files#diff-8e2167038b87d117b916131f62552da8dee7f52d8112e490330349597b393617R1211)... and must live for the entire orchestrator lifetime and not be garbage collected or stopped. 
I think it's needed.

pwilczynskiclearcode (approved):

core/livepeernode.go
mjh1: Gotcha yep, it'll be integrated in a further PR

mjh1 (approved):

cmd/livepeer/starter/starter.go
```diff
@@ -1191,11 +1194,29 @@ func StartLivepeer(ctx context.Context, cfg LivepeerConfig) {
...
+
+			if *cfg.AIModels != "" {
+				configs, err = core.ParseAIModelConfigs(*cfg.AIModels)
+				if err != nil {
+					glog.Errorf("Error parsing -aiModels: %v", err)
+					return
+				}
 			}
 
 			for _, config := range configs {
```
pwilczynskiclearcode, line 1222: @gioelecerati This entire loop needs to be run each time [there are some changes to the ModelsWebhook.configs](https://github.com/livepeer/go-livepeer/pull/3209/files#diff-85b4611f2c22c73db96c87db4253da7c72df9107f7b4c545b740f88ce2fda2e5R191)

pwilczynskiclearcode:
@leszko Do you think this PR feature makes sense in context of remote worker, realtime AI video and future pricing policy plan

leszko:
> @leszko Do you think this PR feature makes sense in context of remote worker, realtime AI video and future pricing policy plan

@gioelecerati @thomshutt @mjh1 My thinking is that we still need this PR, but with the current setup is not the top priority. As per the [latest discussion in Discord](https://discord.com/channels/423160867534929930/1290221842451398730/1298969873619353660), we won't use neither External Container nor Remote Worker for Phase 1 of Realtime Video (Dec 12th deadline). So, we'll deploy a separate Orchestartor in each TensorDock machine. That means that restarting an Orchestartor is slightly less of a problem, because we'll have another Orchestrators deployed in different TensorDock machines.

Still with Realtime Video, restarting Orchestrator is an issue, because it means a downtime for livestream, but that's a separate conversation.

Long story short, I think that if it's not too much effort, then let's just take it to the finish line. However, if there are some other tasks, I think it's ok to park it for now and revisit in Dec/Jan.

PR #3210, opened 2024-10-17
Scaffolding for realtime-to-realtime

j0sh:

Initial bits of the "live-video-to-video" pipeline (better name welcome)

Depends on https://github.com/livepeer/ai-worker/pull/229 for the realtime schema

Adds a MediaMTX handler for the gateway

Adds realtime-to-realtime capability to the orchestrator

Adds selection support between G/O

Orchestrator will respond to the job with trickle endpoints (publish, subscribe)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,6 +2,8 @@ module github.com/livepeer/go-livepeer
 
 go 1.23.2
 
+replace github.com/livepeer/ai-worker => github.com/j0sh/livepeer-ai-worker v0.0.0-20241017214425-b4f177ea52a7
+
 require (
 	contrib.go.opencensus.io/exporter/prometheus v0.4.2
 	github.com/Masterminds/semver/v3 v3.2.1
--- a/go.mod
+++ b/go.mod
@@ -2,8 +2,6 @@ module github.com/livepeer/go-livepeer
 
 go 1.23.2
 
-replace github.com/livepeer/ai-worker => github.com/j0sh/livepeer-ai-worker v0.0.0-20241017214425-b4f177ea52a7
-
 require (
 	contrib.go.opencensus.io/exporter/prometheus v0.4.2
 	github.com/Masterminds/semver/v3 v3.2.1
--- a/server/ai_http.go
+++ b/server/ai_http.go
@@ -56,6 +56,7 @@ func startAIServer(lp lphttp) error {
 	lp.transRPC.Handle("/llm", oapiReqValidator(lp.LLM()))
 	lp.transRPC.Handle("/segment-anything-2", oapiReqValidator(lp.SegmentAnything2()))
 	lp.transRPC.Handle("/image-to-text", oapiReqValidator(lp.ImageToText()))
+	lp.transRPC.Handle("/live-video-to-video", oapiReqValidator(lp.StartLiveVideoToVideo()))
 	// Additionally, there is the '/aiResults' endpoint registered in server/rpc.go
 
 	return nil
@@ -236,6 +237,34 @@ func (h *lphttp) ImageToText() http.Handler {
 		}
 
 		handleAIRequest(ctx, w, r, orch, req)
+
+	})
+}
+
+func (h *lphttp) StartLiveVideoToVideo() http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+
+		// skipping handleAIRequest for now until we have payments
+
+		var (
+			mid    = string(core.RandomManifestID())
+			pubUrl = "/ai/live-video/" + mid
+			subUrl = pubUrl + "/out"
+		)
+		jsonData, err := json.Marshal(&worker.StartLiveVideoToVideoResponse{
+			JSON200: &worker.ResponseStartVideoToVideo{
+				PublishUrl:   &pubUrl,
+				SubscribeUrl: &subUrl,
+			},
+		})
+		if err != nil {
+			respondWithError(w, err.Error(), http.StatusInternalServerError)
+			return
+		}
+
+		w.Header().Set("Content-Type", "application/json")
+		w.WriteHeader(http.StatusOK)
+		w.Write(jsonData)
 	})
 }
 
--- a/server/ai_http.go
+++ b/server/ai_http.go
@@ -251,11 +251,12 @@ func (h *lphttp) StartLiveVideoToVideo() http.Handler {
 			pubUrl = "/ai/live-video/" + mid
 			subUrl = pubUrl + "/out"
 		)
-		jsonData, err := json.Marshal(&worker.StartLiveVideoToVideoResponse{
-			JSON200: &worker.ResponseStartVideoToVideo{
-				PublishUrl:   &pubUrl,
-				SubscribeUrl: &subUrl,
-			},
+		jsonData, err := json.Marshal(struct {
+			PublishUrl   string
+			SubscribeUrl string
+		}{
+			PublishUrl:   pubUrl,
+			SubscribeUrl: subUrl,
 		})
 		if err != nil {
 			respondWithError(w, err.Error(), http.StatusInternalServerError)
--- a/server/ai_mediaserver.go
+++ b/server/ai_mediaserver.go
@@ -74,6 +74,9 @@ func startAIMediaServer(ls *LivepeerServer) error {
 	ls.HTTPMux.Handle("/segment-anything-2", oapiReqValidator(handle(ls, multipartDecoder[worker.GenSegmentAnything2MultipartRequestBody], processSegmentAnything2)))
 	ls.HTTPMux.Handle("/image-to-text", oapiReqValidator(handle(ls, multipartDecoder[worker.GenImageToTextMultipartRequestBody], processImageToText)))
 
+	// This is called by the media server when the stream is ready
+	ls.HTTPMux.Handle("/live/video-to-video/start", ls.StartLiveVideo())
+
 	return nil
 }
 
@@ -359,5 +362,27 @@ func (ls *LivepeerServer) ImageToVideoResult() http.Handler {
 
 		w.WriteHeader(http.StatusOK)
 		_ = json.NewEncoder(w).Encode(resp)
+	})
+}
+
+func (ls *LivepeerServer) StartLiveVideo() http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		streamName := r.FormValue("stream")
+		if streamName == "" {
+			http.Error(w, "Missing stream name", http.StatusBadRequest)
+			return
+		}
+		requestID := string(core.RandomManifestID())
+		params := aiRequestParams{
+			node:        ls.LivepeerNode,
+			os:          drivers.NodeStorage.NewSession(requestID),
+			sessManager: ls.AISessionManager,
+		}
+		ctx := clog.AddVal(r.Context(), "request_id", requestID)
+		req := worker.StartLiveVideoToVideoFormdataRequestBody{
+			// TODO set model and initial parameters here if necessary (eg, prompt)
+		}
+		resp, err := processAIRequest(ctx, params, req)
+		clog.Infof(ctx, "Received live video AI request stream=%s resp=%v err=%v", streamName, resp, err)
 	})
 }
--- a/server/ai_mediaserver.go
+++ b/server/ai_mediaserver.go
@@ -379,9 +379,8 @@ func (ls *LivepeerServer) StartLiveVideo() http.Handler {
 			sessManager: ls.AISessionManager,
 		}
 		ctx := clog.AddVal(r.Context(), "request_id", requestID)
-		req := worker.StartLiveVideoToVideoFormdataRequestBody{
-			// TODO set model and initial parameters here if necessary (eg, prompt)
-		}
+		// TODO set model and initial parameters here if necessary (eg, prompt)
+		req := struct{}{}
 		resp, err := processAIRequest(ctx, params, req)
 		clog.Infof(ctx, "Received live video AI request stream=%s resp=%v err=%v", streamName, resp, err)
 	})
--- a/server/ai_process.go
+++ b/server/ai_process.go
@@ -34,6 +34,7 @@ const defaultAudioToTextModelID = "openai/whisper-large-v3"
 const defaultLLMModelID = "meta-llama/llama-3.1-8B-Instruct"
 const defaultSegmentAnything2ModelID = "facebook/sam2-hiera-large"
 const defaultImageToTextModelID = "Salesforce/blip-image-captioning-large"
+const defaultLiveVideoToVideoModelID = "cumulo-autumn/stream-diffusion"
 
 var errWrongFormat = fmt.Errorf("result not in correct format")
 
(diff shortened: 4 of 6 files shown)


cmd/livepeer/starter/starter.go
```diff
@@ -1348,6 +1348,15 @@ func StartLivepeer(ctx context.Context, cfg LivepeerConfig) {
 					if *cfg.Network != "offchain" {
 						n.SetBasePriceForCap("default", core.Capability_SegmentAnything2, config.ModelID, autoPrice)
 					}
+				case "realtime-to-realtime":
```
leszko: How about calling it `realtime-ai` or `realtime-inference` or `realtime-video-ai`?

server/ai_http.go
```diff
@@ -46,6 +46,7 @@ func startAIServer(lp lphttp) error {
 	lp.transRPC.Handle("/audio-to-text", oapiReqValidator(lp.AudioToText()))
 	lp.transRPC.Handle("/llm", oapiReqValidator(lp.LLM()))
 	lp.transRPC.Handle("/segment-anything-2", oapiReqValidator(lp.SegmentAnything2()))
+	lp.transRPC.Handle("/realtime-to-realtime", oapiReqValidator(lp.StartRealtimeToRealtime()))
```
leszko, line 49: Will this endpoint be used by MediaMTX to send segments? And then receive synchronously segments?

server/ai_mediaserver.go
```diff
@@ -72,6 +72,7 @@ func startAIMediaServer(ls *LivepeerServer) error {
 	ls.HTTPMux.Handle("/audio-to-text", oapiReqValidator(handle(ls, multipartDecoder[worker.GenAudioToTextMultipartRequestBody], processAudioToText)))
 	ls.HTTPMux.Handle("/llm", oapiReqValidator(ls.LLM()))
 	ls.HTTPMux.Handle("/segment-anything-2", oapiReqValidator(handle(ls, multipartDecoder[worker.GenSegmentAnything2MultipartRequestBody], processSegmentAnything2)))
+	ls.HTTPMux.Handle("/realtime-start", ls.RealtimeStart())
```
leszko: What will this endpoint be used for? Don't we need just one endpoint like `/realtime-ai` where MediaMTX will send segments? Something similar to what you shared here: https://github.com/livepeer/ai-worker/pull/209#issue-2537592592

server/ai_mediaserver.go
```diff
@@ -360,3 +361,26 @@ func (ls *LivepeerServer) ImageToVideoResult() http.Handler {
...
+			node:        ls.LivepeerNode,
+			os:          drivers.NodeStorage.NewSession(requestID),
+			sessManager: ls.AISessionManager,
+		}
+		ctx := clog.AddVal(r.Context(), "request_id", requestID)
+		clog.Infof(ctx, "Received realtime AI request")
+		req := worker.StartRealtimeToRealtimeFormdataRequestBody{
+			// TODO set model and initial parameters here if necessary (eg, prompt)
+		}
+		resp, err := processAIRequest(ctx, params, req)
```
leszko, line 384: I don't think we should follow this AI flow here. I guess what we're trying to do is more similar to transcoding.

leszko (commented):
Added some comments.

cmd/livepeer/starter/starter.go
j0sh: Naming is the hardest thing in computer science :)

For reference here are the names of the other pipelines, which generally follow `<input-type>To<output-type>`

https://github.com/livepeer/go-livepeer/blob/439057979b22c6e6c4d6d388c2a811ee13fbcda3/core/capabilities.go#L75-L81

> How about calling it realtime-ai or realtime-inference or realtime-video-ai?

To me, the `-ai` or `-inference` suffixes don't fit any of the existing patterns and don't convey information in this context since all other pipelines are AI inference. We may also have other types of 'live' pipelines, so I think we want to specify the output type here too.

Another thing I have been thinking - we have been using the `realtime` terminology but I think `live` fits a little better (and is more on-brand for Livepeer). Plus it is slightly shorter.

So one option might be to use a prefix, eg `live-` for `live-video-to-video`

This makes the naming easily extendable to other areas too, eg `live-audio-to-text` for live transcription, `live-segment-anything-2` etc

server/ai_mediaserver.go
j0sh: This is the endpoint that MediaMTX used to signal when a stream starts. Then we can kick off the RTMP pull from there (but that is not in this PR)

server/ai_mediaserver.go
j0sh: Why not? This selects an orchestrator and allows us to receive and process a response from it (so we know where to send and receive media from)

This method works for now - if you want to change it later as part of the selection work that is fine but this gets us going with the existing RPC mechanisms

cmd/livepeer/starter/starter.go
rickstaa: @j0sh I like the live prefixed input output naming scheme.

j0sh: Renamed to `live-video-to-video` in https://github.com/livepeer/go-livepeer/pull/3210/commits/a92dacb663aa2f12a8446c457f8342877c3aa42e

j0sh:
Rebased + force pushed to resolve merge conflicts with the latest ai-video. Opening this PR up for real so we can hopefully merge and build on top of it as a team to minimize further conflicts

server/ai_mediaserver.go
mjh1: I think I see what @leszko is saying, up until now we have just been sending in synchronous AI jobs, the pipeline runs and returns the result. Whereas here the request is to start up the realtime to realtime stream basically? That said I don't think it's a huge issue when there's lots of common code that we're able to reuse by following the same flow.

mjh1 (approved):

go.mod
```diff
@@ -2,6 +2,8 @@ module github.com/livepeer/go-livepeer
 
 go 1.21.5
 
+replace github.com/livepeer/ai-worker => github.com/j0sh/livepeer-ai-worker v0.0.0-20241017214425-b4f177ea52a7
```
leszko: I think we need to remove it before merging.

server/ai_mediaserver.go
```diff
@@ -72,6 +72,7 @@ func startAIMediaServer(ls *LivepeerServer) error {
 	ls.HTTPMux.Handle("/audio-to-text", oapiReqValidator(handle(ls, multipartDecoder[worker.GenAudioToTextMultipartRequestBody], processAudioToText)))
 	ls.HTTPMux.Handle("/llm", oapiReqValidator(ls.LLM()))
 	ls.HTTPMux.Handle("/segment-anything-2", oapiReqValidator(handle(ls, multipartDecoder[worker.GenSegmentAnything2MultipartRequestBody], processSegmentAnything2)))
+	ls.HTTPMux.Handle("/live-video-start", ls.StartLiveVideo())
```
leszko:
```suggestion
         // This is called by Media Server when stream is ready
	ls.HTTPMux.Handle("/live-video-start", ls.StartLiveVideo())
```

1. I'd add the comment that this endpoint is actually called by Media Server
2. Maybe rename to match `/live-video-to-video

server/ai_mediaserver.go
leszko: Yes, I think we can "re-use" it for this starting of AI runner. But then, the next question would be, when does the runner container stops? Or who stops it? We don't need to solve it in this PR, but something to consider.

server/ai_mediaserver.go
```diff
@@ -360,3 +361,26 @@ func (ls *LivepeerServer) ImageToVideoResult() http.Handler {
...
+			os:          drivers.NodeStorage.NewSession(requestID),
+			sessManager: ls.AISessionManager,
+		}
+		ctx := clog.AddVal(r.Context(), "request_id", requestID)
+		clog.Infof(ctx, "Received live video AI request")
+		req := worker.StartLiveVideoToVideoFormdataRequestBody{
+			// TODO set model and initial parameters here if necessary (eg, prompt)
+		}
+		resp, err := processAIRequest(ctx, params, req)
+		fmt.Println("stream", streamName, "resp", resp, "error", err)
```
leszko: If we want to merge it into `ai-video`, then please remove or change to `glog` / `clog`.

server/ai_http.go
```diff
@@ -215,6 +216,34 @@ func (h *lphttp) LLM() http.Handler {
 	})
 }
 
+func (h *lphttp) StartVideoToVideo() http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+
+		// skipping handleAIRequest for now until we have payments
+
+		// whats the point of openapi if we have to do this manually
```
leszko: If you want to merge this PR, then please remove this comment.

server/ai_process.go
```diff
@@ -864,6 +865,32 @@ func submitAudioToText(ctx context.Context, params aiRequestParams, sess *AISess
...
+	if resp.JSON200 != nil {
+		fmt.Println("publish", resp.JSON200.PublishUrl, "subscribe", resp.JSON200.SubscribeUrl)
+	}
+	if resp.JSON400 != nil {
+		fmt.Println("json400", resp.JSON400)
+	}
+	if resp.JSON500 != nil {
+		fmt.Println("json500", resp.JSON500)
+	}
+	fmt.Println("body is", string(resp.Body))
```
leszko: If you plan to merge, then please change to `clog` or remove.

server/ai_process.go
```diff
@@ -33,6 +33,7 @@ const defaultUpscaleModelID = "stabilityai/stable-diffusion-x4-upscaler"
 const defaultAudioToTextModelID = "openai/whisper-large-v3"
 const defaultLLMModelID = "meta-llama/llama-3.1-8B-Instruct"
 const defaultSegmentAnything2ModelID = "facebook/sam2-hiera-large"
+const defaultLiveVideoToVideoModelID = "stream-diffusion" // TODO what should this be?
```
leszko: I guess stream diffusion is not on huggingface, so you can leave it as `stream-diffusion` or change to `cumulo-autumn/stream-diffusion`

leszko (approved):
Added some comments. When resolved and CI passes, then I think it's good enough to merge as a scaffold. Thanks @j0sh !

server/ai_mediaserver.go
j0sh: Yep as @mjh1 says this basically makes a synchronous request using the normal AI inference path. This is pretty convenient, because it gives us an entry point to kick off the rest of the process - see https://github.com/livepeer/go-livepeer/pull/3211 for the beginnings of that.

I expect that we'd be building things like our job monitoring / re-selection, payments, etc somewhere near this '[main loop](https://github.com/livepeer/go-livepeer/pull/3211/files#diff-5dd7b71c3614e76641eba97465018e0d92a2bf6be8918853ac85763e4c2010c6)'

> But then, the next question would be, when does the runner container stops? Or who stops it? We don't need to solve it in this PR, but something to consider.

This is would be a basic RPC request which can be done via whatever call-path makes sense. However, as far as job tear-down goes, I am not sure if we can rely on signaling for that (anything can disappear without notice), so we should ensure things still work well in the absence of explicit tear-down.

We'll need to solve a similar issue with calls to the model control API. The gateway will need some handle to the orchestrator session, the orchestrator to the (remote) worker, and possibly the worker to the runner (if runners have persistent job state). @victorges for visibility since this is his area

leszko: > This is would be a basic RPC request which can be done via whatever call-path makes sense. However, as far as job tear-down goes, I am not sure if we can rely on signaling for that (anything can disappear without notice), so we should ensure things still work well in the absence of explicit tear-down.

Yeah, the best approach from the Distributed (and Decentralized) system perspective is what we do for Transcoding:
1. The params are included in the stream (segments)
2. The AI (Transcoding) pipelines is initialized when the first segment comes in
3. The tear down is down on some timeout (so when there are no segments coming for x min, then it tears down)

Any signaling of warm up / tear down is an "optimization" on top of that. This approach is good for reliability, because a lot may happen, O may go down, we may swap the orchestrator, G may go down (before signaling something), etc. 

CC: @victorges 

Again, I think we can move this discussion outside this PR. The PR can be merged and we can discuss it separately.

rickstaa:
Hey @j0sh, just a heads up—I had to update the dependencies to ensure the network is compatible with the new Go version and other updated libraries. I found a combination that works, but if you need to use lower versions that are also compatible, feel free to adjust them. I'm flexible with it. I'm also improving the code through small pull requests so that it doesn't cause to many conflicts. If you run into problems or need me to rebase just ping me.

go.mod
j0sh: Yep, actually first we need https://github.com/livepeer/ai-worker/pull/229 or maybe https://github.com/livepeer/ai-worker/pull/233 first in order to define the pipeline schema

server/ai_mediaserver.go
j0sh: Comment added in 9e461a7d2ee59b1cd5603382d403c082b71469ad

@victorges [suggested](https://discord.com/channels/423160867534929930/1296851581240610818) using a `/live/` prefix for these paths which sounds good to me so I did that here. Named it `/live/video-to-video/start` since I suspect we will want to append other things to the route, like maybe `/live/video-to-video/api` or thereabouts

server/ai_mediaserver.go
j0sh: changed to clog in https://github.com/livepeer/go-livepeer/commit/9e461a7d2ee59b1cd5603382d403c082b71469ad

btw I might send a PR in the near future to add slog-style parameter lists to clog since they are super convenient for ad-hoc KV pairs (in addition to the pre-defined `clog.Add` method we have)

server/ai_http.go
j0sh: A bit of late-night frustration that crept through :) These handlers are largely boilerplate that should be done for us by codegen. Removed in https://github.com/livepeer/go-livepeer/commit/9e461a7d2ee59b1cd5603382d403c082b71469ad

server/ai_process.go
j0sh: Removed in https://github.com/livepeer/go-livepeer/commit/9e461a7d2ee59b1cd5603382d403c082b71469ad

server/ai_process.go
j0sh: Changed to `cumulo-autumn/streamdiffusion` in https://github.com/livepeer/go-livepeer/commit/9e461a7d2ee59b1cd5603382d403c082b71469ad

j0sh:
Force pushed to fix up a bunch of merge conflicts with the latest ai-video branch

Also squashed everything down to 1 commit to avoid a cascading series of merge conflicts

Builds for this PR are broken for now until the pipeline on the worker / runner side is in, either in https://github.com/livepeer/ai-worker/pull/229 or https://github.com/livepeer/ai-worker/pull/233

j0sh:
Updated to remove the dependency on AI Worker changes. The scaffolding that is there now is now even less useful, but there is still the capability definition and pipeline path, both of which have caused merge conflicts recently as we've added new stuff

If CI is green here, I'll go ahead and merge so it's easier to stay on the leading edge of the tree.

PR #3211, opened 2024-10-18
Gateway publish to orchestrator.

j0sh:
- Requires #3210 
- Set up a trickle HTTP endpoints on the orchestrator (requires golang 1.22 for the new routes; will send a separate patch to bump the go.mod and do any CI adjustments)
- Pulls a RTMP stream from MediaMTX on the gateway when a new stream comes in 
- Converts the RTMP into mpegts segments
- Publishes segments to the orchestrator via trickle HTTP

### DIFF
--- a/server/ai_http.go
+++ b/server/ai_http.go
@@ -25,6 +25,7 @@ import (
 	"github.com/livepeer/go-livepeer/common"
 	"github.com/livepeer/go-livepeer/core"
 	"github.com/livepeer/go-livepeer/monitor"
+	"github.com/livepeer/go-livepeer/trickle"
 	middleware "github.com/oapi-codegen/nethttp-middleware"
 )
 
@@ -50,6 +51,8 @@ func startAIServer(lp lphttp) error {
 
 	openapi3filter.RegisterBodyDecoder("image/png", openapi3filter.FileBodyDecoder)
 
+	trickle.ConfigureServerWithMux(lp.transRPC)
+
 	lp.transRPC.Handle("/text-to-image", oapiReqValidator(aiHttpHandle(&lp, jsonDecoder[worker.GenTextToImageJSONRequestBody])))
 	lp.transRPC.Handle("/image-to-image", oapiReqValidator(aiHttpHandle(&lp, multipartDecoder[worker.GenImageToImageMultipartRequestBody])))
 	lp.transRPC.Handle("/image-to-video", oapiReqValidator(aiHttpHandle(&lp, multipartDecoder[worker.GenImageToVideoMultipartRequestBody])))
@@ -58,8 +61,9 @@ func startAIServer(lp lphttp) error {
 	lp.transRPC.Handle("/llm", oapiReqValidator(aiHttpHandle(&lp, jsonDecoder[worker.GenLLMFormdataRequestBody])))
 	lp.transRPC.Handle("/segment-anything-2", oapiReqValidator(aiHttpHandle(&lp, multipartDecoder[worker.GenSegmentAnything2MultipartRequestBody])))
 	lp.transRPC.Handle("/image-to-text", oapiReqValidator(aiHttpHandle(&lp, multipartDecoder[worker.GenImageToTextMultipartRequestBody])))
-	lp.transRPC.Handle("/live-video-to-video", oapiReqValidator(lp.StartLiveVideoToVideo()))
 	lp.transRPC.Handle("/text-to-speech", oapiReqValidator(aiHttpHandle(&lp, multipartDecoder[worker.GenTextToSpeechJSONRequestBody])))
+
+	lp.transRPC.Handle("/live-video-to-video", oapiReqValidator(lp.StartLiveVideoToVideo()))
 	// Additionally, there is the '/aiResults' endpoint registered in server/rpc.go
 
 	return nil
@@ -89,24 +93,20 @@ func (h *lphttp) StartLiveVideoToVideo() http.Handler {
 
 		var (
 			mid    = string(core.RandomManifestID())
-			pubUrl = "/ai/live-video/" + mid
-			subUrl = pubUrl + "/out"
+			pubUrl = trickle.BaseServerPath + mid
+			subUrl = pubUrl + "-out"
 		)
-		jsonData, err := json.Marshal(struct {
-			PublishUrl   string
-			SubscribeUrl string
-		}{
-			PublishUrl:   pubUrl,
-			SubscribeUrl: subUrl,
-		})
+		jsonData, err := json.Marshal(
+			&worker.LiveVideoToVideoResponse{
+				PublishUrl:   pubUrl,
+				SubscribeUrl: subUrl,
+			})
 		if err != nil {
 			respondWithError(w, err.Error(), http.StatusInternalServerError)
 			return
 		}
 
-		w.Header().Set("Content-Type", "application/json")
-		w.WriteHeader(http.StatusOK)
-		w.Write(jsonData)
+		respondJsonOk(w, jsonData)
 	})
 }
 
--- a/server/ai_http.go
+++ b/server/ai_http.go
@@ -31,6 +31,8 @@ import (
 
 var MaxAIRequestSize = 3000000000 // 3GB
 
+var TrickleHTTPPath = "/ai/trickle/"
+
 func startAIServer(lp lphttp) error {
 	swagger, err := worker.GetSwagger()
 	if err != nil {
@@ -51,7 +53,10 @@ func startAIServer(lp lphttp) error {
 
 	openapi3filter.RegisterBodyDecoder("image/png", openapi3filter.FileBodyDecoder)
 
-	trickle.ConfigureServerWithMux(lp.transRPC)
+	lp.trickleSrv = trickle.ConfigureServer(trickle.TrickleServerConfig{
+		Mux:      lp.transRPC,
+		BasePath: TrickleHTTPPath,
+	})
 
 	lp.transRPC.Handle("/text-to-image", oapiReqValidator(aiHttpHandle(&lp, jsonDecoder[worker.GenTextToImageJSONRequestBody])))
 	lp.transRPC.Handle("/image-to-image", oapiReqValidator(aiHttpHandle(&lp, multipartDecoder[worker.GenImageToImageMultipartRequestBody])))
@@ -93,7 +98,7 @@ func (h *lphttp) StartLiveVideoToVideo() http.Handler {
 
 		var (
 			mid    = string(core.RandomManifestID())
-			pubUrl = trickle.BaseServerPath + mid
+			pubUrl = TrickleHTTPPath + mid
 			subUrl = pubUrl + "-out"
 		)
 		jsonData, err := json.Marshal(
--- a/server/ai_mediaserver.go
+++ b/server/ai_mediaserver.go
@@ -367,7 +367,8 @@ func (ls *LivepeerServer) StartLiveVideo() http.Handler {
 		// Kick off the RTMP pull and segmentation as soon as possible
 		ssr := media.NewSwitchableSegmentReader()
 		go func() {
-			media.RunSegmentation("rtmp://localhost/"+streamName, ssr.Read)
+			ms := media.MediaSegmenter{Workdir: ls.LivepeerNode.WorkDir}
+			ms.RunSegmentation("rtmp://localhost/"+streamName, ssr.Read)
 			// TODO handle stream stops
 		}()
 
--- a/server/ai_mediaserver.go
+++ b/server/ai_mediaserver.go
@@ -14,6 +14,7 @@ import (
 	"github.com/livepeer/go-livepeer/clog"
 	"github.com/livepeer/go-livepeer/common"
 	"github.com/livepeer/go-livepeer/core"
+	"github.com/livepeer/go-livepeer/media"
 	"github.com/livepeer/go-tools/drivers"
 	middleware "github.com/oapi-codegen/nethttp-middleware"
 	"github.com/oapi-codegen/runtime"
@@ -360,15 +361,26 @@ func (ls *LivepeerServer) StartLiveVideo() http.Handler {
 			return
 		}
 		requestID := string(core.RandomManifestID())
-		params := aiRequestParams{
-			node:        ls.LivepeerNode,
-			os:          drivers.NodeStorage.NewSession(requestID),
-			sessManager: ls.AISessionManager,
-		}
 		ctx := clog.AddVal(r.Context(), "request_id", requestID)
-		// TODO set model and initial parameters here if necessary (eg, prompt)
-		req := struct{}{}
-		resp, err := processAIRequest(ctx, params, req)
-		clog.Infof(ctx, "Received live video AI request stream=%s resp=%v err=%v", streamName, resp, err)
+		clog.Infof(ctx, "Received live video AI request for %s", streamName)
+
+		// Kick off the RTMP pull and segmentation as soon as possible
+		ssr := media.NewSwitchableSegmentReader()
+		go func() {
+			media.RunSegmentation("rtmp://localhost/"+streamName, ssr.Read)
+			// TODO handle stream stops
+		}()
+
+		params := aiRequestParams{
+			node:          ls.LivepeerNode,
+			os:            drivers.NodeStorage.NewSession(requestID),
+			sessManager:   ls.AISessionManager,
+			segmentReader: ssr,
+		}
+
+		req := worker.GenLiveVideoToVideoJSONRequestBody{
+			// TODO set model and initial parameters here if necessary (eg, prompt)
+		}
+		processAIRequest(ctx, params, req)
 	})
 }
--- a/server/ai_mediaserver.go
+++ b/server/ai_mediaserver.go
@@ -369,7 +369,7 @@ func (ls *LivepeerServer) StartLiveVideo() http.Handler {
 		go func() {
 			ms := media.MediaSegmenter{Workdir: ls.LivepeerNode.WorkDir}
 			ms.RunSegmentation("rtmp://localhost/"+streamName, ssr.Read)
-			// TODO handle stream stops
+			ssr.Close()
 		}()
 
 		params := aiRequestParams{
--- a/server/ai_process.go
+++ b/server/ai_process.go
@@ -1039,13 +1039,7 @@ func submitLiveVideoToVideo(ctx context.Context, params aiRequestParams, sess *A
 		}
 		clog.V(common.VERBOSE).Infof(ctx, "pub %s sub %s", pub, sub)
 		startTricklePublish(pub, params)
-		startTrickleSubscribe(sub)
-	}
-	if resp.JSON400 != nil {
-		fmt.Println("json400", resp.JSON400)
-	}
-	if resp.JSON500 != nil {
-		fmt.Println("json500", resp.JSON500)
+		startTrickleSubscribe(sub, params)
 	}
 	return resp, nil
 }
--- a/server/ai_process.go
+++ b/server/ai_process.go
@@ -12,6 +12,7 @@ import (
 	"math"
 	"math/big"
 	"net/http"
+	"net/url"
 	"path/filepath"
 	"strconv"
 	"strings"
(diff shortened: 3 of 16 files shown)


server/ai_mediaserver.go
```diff
@@ -370,17 +371,26 @@ func (ls *LivepeerServer) StartLiveVideo() http.Handler {
 			return
 		}
 		requestID := string(core.RandomManifestID())
+		ctx := clog.AddVal(r.Context(), "request_id", requestID)
+		clog.Infof(ctx, "Received live video AI request for %s", streamName)
+
+		// Kick off the RTMP pull and segmentation as soon as possible
+		ssr := media.NewSwitchableSegmentReader()
+		go func() {
+			media.RunSegmentation("rtmp://localhost/"+streamName, ssr.Read)
```
leszko: Does it mean, that Gateway will pull the stream from MediaMTX? Isn't it opposite to what you designed on the diagram?

image

server/ai_process.go
```diff
@@ -827,16 +832,42 @@ func submitLiveVideoToVideo(ctx context.Context, params aiRequestParams, sess *A
...
+			}
+			if pu.Hostname() != "" {
+				// url has a hostname already so use it
+				return pu, nil
+			}
+			// no hostname, so append one
+			u = sess.Transcoder() + u
+			return url.Parse(u)
+		}
+		pub, err := appendHostname(resp.JSON200.PublishUrl)
```
leszko, line 1032: Why do you need PublishUrl in Orchestrator? Isn't it Gateway Publishing to Orchestrator? 

I think in general it's hard to grasp how this trickle server works. Could you describe it somewhere? I understand it should correspond to diagram from your doc, but where is the publish/subscribe part?

image


image

leszko (commented):
Good work, I think this is a very important PR, because it describes the protocol between G<>O. I added 2 comments, because I think I still don't understand how we plan it all to work. 

Other than that, @j0sh is it possible to actually run the code from this PR? If yes, could you describe how to run it locally? I think if I plan with it, I can have some more comments and understand if better.

server/ai_mediaserver.go
j0sh: Yes, gateway initiates the RTMP pull, but the flow of media still goes from mediamtx -> gateway

I guess there could be a better distinction between "who initiates the pull" vis-a-vis the actual flow of media, but RTMP isn't really a request-response protocol in the same way the other HTTP flows are in this diagram. Open to suggestions.

server/ai_process.go
j0sh: > Why do you need PublishUrl in Orchestrator? Isn't it Gateway Publishing to Orchestrator?

Yeah this was a bit of a spur of the moment addition when I was looking at the overall flow, where we are doing this request -> response call anyway. This gives the exact endpoint where a publish should happen. Likewise for the subscribe URL - it tells the gateway where to pull the results.

We could skip this entirely and hard-code the URLs via well-known paths, distinguish jobs via ids in HTTP headers, etc but this is an easy way for us to add a bit of topological flexibility without breaking the protocol later (eg, routing the stream to a different machine).

In fact from some of the conversations on Discord right now, there is probably another way to make this even more robust: to get a _list_ of subscribe URLs (think multiple renditions of low latency video transcoding)

j0sh:
Force-pushed to fix merge conflicts from the ai-video rebase

> is it possible to actually run the code from this PR? If yes, could you describe how to run it locally?

Absolutely! The big thing is making sure you have MediaMTX running. This is pretty simple - it is a single executable + config file. Download a [pre-built release](https://github.com/bluenviron/mediamtx/releases/tag/v1.9.3) and use this [config file](https://gist.githubusercontent.com/j0sh/e6d7132bf9c7656c86e3b400d2b8ace9/raw/7800b2cb4a970529f8298027fe5b5fd11aa5ea32/mediamtx.yml) [1]. Stick the executable in the same directory as the config and run `./mediamtx`

Then for the gateway:
```
./livepeer -gateway -rtmpAddr :1936 -httpAddr :5936 -orchAddr localhost:8935
```

For the orchestrator + worker:
```
./livepeer -orchestrator -aiWorker -aiModels 'live-video-to-video:stream-diffusion:false' -serviceAddr localhost:8935 -transcoder
```

Publish to `http://<mediamtx-host>:8889/streamname/publish` ; see more details in https://github.com/livepeer/ai-worker/pull/209

> I still don't understand how we plan it all to work

That's fair; as mentioned earlier I think that is partially a result of recent pressure to deliver things without enough time to fully design an end-to-end flow within go-livepeer.

BTW, I am afraid that this PR is probably not _quite_ enough for you to base your work on just yet, if the plan is still to carry payments within media stream. The trickle server on the orchestrator only behaves as a simple pipe between publisher and subscriber, and does not have any mechanisms (yet) to execute additional code based on incoming segments, eg for us to process PM tickets or record metrics for selection, and we need to adjust the publisher API to also include custom headers. I'll have those in within the next day or so.

NB: One elephant in the room here is "how do we integrate mediamtx into our infrastructure" and I will spin up a separate thread to discuss that.

[1] The only difference between this config file and the MediaMTX sample config is in the addition of a runOnReady curl hook, and enabling STUN support for WebRTC (because my cloud dev box doesn't work without it; prod might vary). Diff below:

```diff
diff --git b/mediamtx.yml a/mediamtx.yml
index c3aed76..cf7c60c 100644
--- b/mediamtx.yml
+++ a/mediamtx.yml
@@ -376,8 +376,8 @@ webrtcAdditionalHosts: []
 # ICE servers. Needed only when local listeners can't be reached by clients.
 # STUN servers allows to obtain and share the public IP of the server.
 # TURN/TURNS servers forces all traffic through them.
-webrtcICEServers2: []
-  # - url: stun:stun.l.google.com:19302
+webrtcICEServers2:
+  - url: stun:stun.l.google.com:19302
   # if user is "AUTH_SECRET", then authentication is secret based.
   # the secret must be inserted into the password field.
   # username: ''
@@ -643,7 +643,7 @@ pathDefaults:
   #   a regular expression.
   # * MTX_SOURCE_TYPE: source type
   # * MTX_SOURCE_ID: source ID
-  runOnReady:
+  runOnReady: curl http://localhost:5936/live-video-start -F stream=$MTX_PATH
   # Restart the command if it exits.
   runOnReadyRestart: no
   # Command to run when the stream is not available anymore.
```

server/ai_mediaserver.go
leszko: Maybe you can add one more arrow with "initiate RTMP" 🙃 For me it is/was pretty confusing.

leszko:
> Absolutely! The big thing is making sure you have MediaMTX running. This is pretty simple - it is a single executable + 
Thanks. I think I have it working 🙏

trickle/trickle_server.go
```diff
@@ -0,0 +1,423 @@
...
+
+	stream.clear()
+	sm.mutex.Lock()
+	defer sm.mutex.Unlock()
+	delete(sm.streams, streamName)
+	slog.Info("Deleted stream", "streamName", streamName)
+}
+
+func (sm *StreamManager) handlePost(w http.ResponseWriter, r *http.Request) {
+	stream := sm.getOrCreateStream(r.PathValue("streamName"))
```
leszko: IIUC here we'll need to handle payment (or at least check have some check if the segment is covered). Is it correct @j0sh ?

j0sh: Mostly yes, but not in this exact spot because not all streams will have payments, such as the  result stream.

I can add some middleware for the O to handle incoming segments and you can add in payments from there, eg by modifying outgoing headers. But if you are thinking about an out-of-band payments mechanism then the middleware can probably be skipped for now.

leszko: Yeah, even if we do the payment outside of this flow, then we'll need to have somewhere the check "if it's paid" in order to stop sending segments if no payment.

j0sh: Yep - payments outside the flow can be a little simpler, eg a local (in-memory) subscriber rather than middleware that modifies the stream. Then you can gather any needed metrics within the subscribe callback and use those to check payments.

BTW if we do payments by "time" then it might not even be needed!

trickle/trickle_publisher.go
```diff
@@ -0,0 +1,167 @@
...
+	req.Header.Set("Content-Type", c.contentType)
+
+	// Start the POST request in a background goroutine
+	// TODO error handling for these
+	go func() {
+		slog.Info("Initiailzing http client", "idx", index)
+		// Createa new client to prevent connection reuse
+		client := http.Client{Transport: &http.Transport{
+			// ignore orch certs for now
+			TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
```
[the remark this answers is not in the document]
j0sh: bad bot

j0sh:
Marking as ready for review so we can get this in and start building other things on top of it. It it also causing merge conflicts with new pipelines so best to get those out of the way

server/ai_http.go
```diff
@@ -49,6 +50,8 @@ func startAIServer(lp lphttp) error {
 
 	openapi3filter.RegisterBodyDecoder("image/png", openapi3filter.FileBodyDecoder)
 
+	trickle.ConfigureServerWithMux(lp.transRPC)
```
leszko: Do we maybe want to have a feature flag for this? 

For the context, we'll merge `ai-video` into `master` most probably tomorrow. Then, this change may go to out transcoding broadcaster.

trickle/trickle_server.go
```diff
@@ -0,0 +1,423 @@
...
+}
+
+const maxSegmentsPerStream = 5
+
+const BaseServerPath = "/ai/live-video/"
+
+var FirstByteTimeout = errors.New("pending read timeout")
+
+func ConfigureServerWithMux(mux *http.ServeMux) {
+	/* TODO we probably want to configure the below
```
leszko: Why not to configure it right in this PR?

trickle/trickle_server.go
```diff
@@ -0,0 +1,423 @@
...
+
+	streamManager := &StreamManager{
+		streams: make(map[string]*Stream),
+	}
+	mux.HandleFunc("GET "+BaseServerPath+"{streamName}/{idx}", streamManager.handleGet)
+	mux.HandleFunc("POST "+BaseServerPath+"{streamName}/{idx}", streamManager.handlePost)
+	mux.HandleFunc("DELETE "+BaseServerPath+"{streamName}", streamManager.handleDelete)
+}
+
+func (sm *StreamManager) getStream(streamName string) (*Stream, bool) {
```
leszko: Nit: Don't you want to reorder functions to follow the [stepdown rule](https://medium.com/@alankarrocks/enhancing-code-readability-with-the-step-down-rule-3185ff78650b). I guess it's simpler to read the code.

trickle/trickle_server.go
```diff
@@ -0,0 +1,423 @@
...
+}
+
+func (sm *StreamManager) getOrCreateStream(streamName string) *Stream {
+	sm.mutex.Lock()
+	defer sm.mutex.Unlock()
+
+	stream, exists := sm.streams[streamName]
+	if !exists {
+		stream = &Stream{
+			segments: make([]*Segment, 5),
```
leszko:
```suggestion
			segments: make([]*Segment, maxSegmentsPerStream),
```

Shouldn't you use `maxSegmentsPerStream` instead of `5` here?

trickle/trickle_server.go
```diff
@@ -0,0 +1,423 @@
...
+		}
+		sm.streams[streamName] = stream
+		slog.Info("Creating stream", "stream", streamName)
+	}
+	return stream
+}
+
+func (sm *StreamManager) clearAllStreams() {
```
leszko: Were is it (or will it be) used?

trickle/trickle_server.go
```diff
@@ -0,0 +1,423 @@
...
+
+func (sm *StreamManager) handleDelete(w http.ResponseWriter, r *http.Request) {
+	streamName := r.PathValue("streamName")
+	stream, exists := sm.getStream(streamName)
+	if !exists {
+		http.Error(w, "Invalid stream name", http.StatusBadRequest)
+		return
+	}
+
+	// TODO properly clear sessions once we have a good solution
```
leszko: What is the proper solution for clearing the stream?

trickle/trickle_server.go
```diff
@@ -0,0 +1,423 @@
...
+func (s *Stream) clear() {
+	s.mutex.Lock()
+	defer s.mutex.Unlock()
+	for _, segment := range s.segments {
+		segment.close()
+	}
+	s.segments = make([]*Segment, maxSegmentsPerStream)
+}
+
+func (sm *StreamManager) handleDelete(w http.ResponseWriter, r *http.Request) {
```
leszko: IIUC Gateway will send a request to delete the stream. All good, but what happens if Gateway never sends that request? Shouldn't we have some automatic delete/cleanup timeout?

trickle/trickle_server.go
```diff
@@ -0,0 +1,423 @@
...
+// Also the server should close any pending POST and GET requests for X < segment < Y
+// even if these segment have NO data. This would mean marking any pending POST
+// segments as 'closed' and returning empty data (but still a 200 OK) to any
+// clients that GET these segments.
+// (in practice this would just be a single pre-connection for X + 1)
+// Server discards bytes for Segment Y up to its last read byte of Segment X
+// After reaching the last read byte, start buffering up the bytes of Segment Y
+// as they come in. Process requests normally.
+
+// Handle post requests for a given index
```
leszko, line 229: I'm trying to understand the logic of this function. Could you maybe describe it in the docs above? 

My understanding is that it's an HTTP POST request which is kept forever open  and the body is streamed all the time. That is why you have this `timeoutReader`. Is that correct?

leszko (commented):
@j0sh I added some comments, but in general, it's hard for me to understand the details of what happens in the Tickle/Transport protocol. I think I understand how it works with G and O, but I don't understand the details of the protocol itself.

Some things that could help:
1. Add a README or comments in `trickle_server.go` with how it works from the overall perspective (I understanding is that Publisher sends a forever lasting HTTP POST req and subscriber sends a forever lasting GET req, but I'm not sure about it 🙃 )
2. Add better comments to the parts that are not obvious, some examples:
    - Why do we have `firstByteRead`, why there is a difference with failing to read 1st byte and 3rd byte?
    - Why do we have `timeoutReader`?
    - What is `idx` in the stream? If it's an index of the segment in the stream and it's set as POST param, then I guess my understanding that we have one long-lasting HTTP POST connection is wrong
    - ...
3. Add unit tests

Another option is, if we don't have time to make it right ☝️ is to just merge it. I just don't feel comfortable if go live and you're the only person understanding it. But we could merge it and improve on that later if it helps with the Realtime Video work distribution and unblocks some other work. Then, I'm ok with merging it behind some feature flag.

trickle/trickle_server.go
j0sh: Old habits from C, where you have to define (or at least declare) functions before using them. I am somewhat used to looking _up_ for a given definition, rather than down but I suppose it does not really matter in golang. If the convention is step down style, we can do that but there will always be some subjectivity about what is "important" enough to come first (declaration-first at least gives something of a rule to follow)

trickle/trickle_server.go
j0sh: Yeah we will need to sweep the server periodically, but that can come later

trickle/trickle_server.go
j0sh: Shutdown originally but I think it is not used at the moment

trickle/trickle_server.go
j0sh: Clearing the stream is not the problem, but session reuse was (eg, trying to start a new session with the same name). We can work around the issue for now by loosening up some of the constraints around sequence numbering. In practice I think this will be less of an issue as long as the O hands out a fresh ID for each session which it does [here](https://github.com/livepeer/go-livepeer/pull/3211/files/4db8c949d4edb8da4e1906dad4f1f36d3e82cbc3#diff-2d409e07b090aec3653ceddfa542b8809eea772f3f24920e8c285d1e22edd15bR256-R257).

server/ai_http.go
j0sh: This would only be triggered on orchestrators with AI enabled, and in order to do anything, they would also need a live pipeline loaded - otherwise, selection should fail on the gateway side.

We can still add a (temporary?) flag if there is something else that would protect against.

trickle/trickle_server.go
j0sh: Just hadn't gotten around to it yet. Updated the PR with some of this.

trickle/trickle_server.go
j0sh: Yep, fixed in cccdde875b2702c502e5a379accf61dcc0e91488

j0sh:
@leszko Thanks for the review!

Added a README with some details, updated the code to address some of the TODOs and PR feedback, plus bug fixes

trickle/trickle_server.go
j0sh: Thought I answered this earleir - but I added some explanation in the trickle readme that should hopefully clear things up. (Also removed the long comment right above this.)

To sum up, POST requests are _not_ meant to be kept open forever. For video, they are a GOP / segment in length.

Clients will pre-connect the next POST in the sequence to minimize set-up time, and we use the `timeoutReader` to send down periodic 100 Continue keepalives on that preconnect while waiting for content to come in. As soon as content comes in (eg, the first byte), we stop sending keepalives and proceed as normal. Added some [comments](https://github.com/livepeer/go-livepeer/pull/3211/commits/b7c234c84c926ee42c80d4d1fe9dfabf50036754#diff-dacf4cf9a63e993246994800272783c2d061f86392e1dbe007af427a836886ae) to clarify that

server/ai_http.go
leszko: Yeah, maybe you're right. We can go with this part without feature flag.

trickle/local_subscriber.go
```diff
@@ -0,0 +1,77 @@
...
+
+type TrickleLocalSubscriber struct {
+	channelName string
+	server      *Server
+
+	mu  *sync.Mutex
+	seq int
+}
+
+func NewLocalSubscriber(sm *Server, channelName string) *TrickleLocalSubscriber {
```
leszko, line 26: Why do we need it if it's not used?

trickle/README.md
```diff
@@ -0,0 +1,57 @@
+# Trickle Protocol
```
leszko, line 1: Thanks for the README. Super helpful!

Could you also add info about what's `changefeed`?

trickle/trickle_server.go
```diff
@@ -0,0 +1,459 @@
...
+	segment.close()
+}
+
+func (s *Stream) getForWrite(idx int) (*Segment, bool) {
+	s.mutex.Lock()
+	defer s.mutex.Unlock()
+	if idx == -1 {
+		idx = s.latestWrite
+		// TODO figure out how to better handle restarts while maintaining ordering
+		/* } else if idx > s.latestWrite { */
```
leszko: Leftover?

leszko (approved):
@j0sh Thanks for adding README.

I think it's ok to merge it. We'll need to work on some better documentation and unit tests, but we can do it in a separate PR. I can say I understand 50% of how trickle works internally, but I guess it's good enough for now 🙃 

After second thought, I think we don't need any feature flag.

trickle/trickle_server.go
j0sh: Removed in https://github.com/livepeer/go-livepeer/pull/3211/commits/576011677d57c3852141552b74a38957bea861e2

trickle/README.md
j0sh: Added a brief description in https://github.com/livepeer/go-livepeer/pull/3211/commits/576011677d57c3852141552b74a38957bea861e2 but we don't use it here - its still a little iffy, but I have been testing it out with a [pub-sub](https://github.com/j0sh/http-trickle/blob/main/cmd/publisher-out/publisher-out.go#L147-L170) example that has been running on the demo machine

trickle/local_subscriber.go
j0sh: It is used in https://github.com/livepeer/go-livepeer/pull/3232

PR #3213, opened 2024-10-18
Bump golang version to 1.23.2

j0sh:

Bump golang version to 1.23.2 from 1.21.4

This is needed for #3211 - technically only 1.22 is required but golang is already on 1.23 and releases are very stable so we might as well ... _go_ ... straight to the latest version.

- Updates go.mod
- Updates CI downloads

Technically we do not need to update the CI-installed version of golang because apparently go mod will [download updated toolchains](https://go.dev/doc/toolchain) on its own (!) since version 1.21 but it seems better to stay somewhat consistent. This behavior also eases the migration path for fellow developers: if you could already compile go-livepeer before this, you can still compile go-livepeer after  - this without doing anything - since your golang will automatically update itself (!)

Everything compiles and tests pass

**Checklist:**


thomshutt (approved):

rickstaa (approved):

leszko:
@j0sh The PR looks fine, but please check if CI is passing. I just enabled CI with tests and lint, so let's try to keep it green. 

Sorry if it broke your build, but I'm trying to put `ai-video` into good shape (and then merge it into `master`). Thanks for the understanding.

j0sh:
@leszko Seems that CI passed (see image below), unless you are referring to something else?

image

PR #3225, opened 2024-10-28
Live Video Payment Mechanism

leszko:
It splits the payment mechanism from media transport by creating a new endpoint '/payment' dedicated just for payments.

### How will it work with Live Video Transport Protocol?
1. While sending segment(s), Gateway will call `LivePaymentSender.SendPayment()`
2. While receiving segments(s), Orchestrators will call `LivePaymentReceiver.AccountSegment()`
   - If it returns error, we should stop Realtime Video Stream
3. We need to pay upfront for a few segments to avoid race condition (if segment data comes before payment)

### Notes to reviewers
- This code is currently not used anywhere, still, I suggest reviewing it and merging to keep PRs small
- it does not change any existing logic, so for transcoding and AI jobs, the payment is still sent together with the job params


server/segment_rpc.go
```diff
@@ -254,6 +219,78 @@ func (h *lphttp) ServeSegment(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+	clog.V(common.DEBUG).Infof(ctx, "Payment processed, current balance = %s", currentBalance(h, payment, segData))
+
+	w.Write(buf)
+}
+
+func currentBalance(h *lphttp, payment net.Payment, segData *core.SegTranscodingMetadata) string {
+	if h == nil || h.node == nil || h.node.Balances == nil || segData == nil || segData.AuthToken == nil {
+		return ""
```
thomshutt: If this is only being used in logging, should we do something more descriptive than an empty string?

server/broadcast.go
```diff
@@ -1580,6 +1571,19 @@ func updateSession(sess *BroadcastSession, res *ReceivedTranscodeResult) {
...
 
+func refreshSessionIfNeeded(ctx context.Context, sess *BroadcastSession) error {
+	shouldRefresh, err := shouldRefreshSession(ctx, sess)
+	if err != nil {
+		return err
+	}
+	if shouldRefresh {
+		if err := refreshSession(ctx, sess); err != nil {
+			return err
+		}
```
thomshutt:
```suggestion
		return refreshSession(ctx, sess)
```

server/segment_rpc.go
leszko: Updated

thomshutt (approved):

server/rpc.go
```diff
@@ -405,6 +407,8 @@ func orchestratorInfoWithCaps(orch Orchestrator, addr ethcommon.Address, service
 	expiration := time.Now().Add(authTokenValidPeriod).Unix()
 	authToken := orch.AuthToken(sessionID, expiration)
 
+	addr2 := orch.Address()
+	fmt.Println(addr2)
```
j0sh: stray?

net/lp_rpc.pb.go
```diff
@@ -1,7 +1,7 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
-// 	protoc-gen-go v1.33.0
-// 	protoc        v3.12.4
+// 	protoc-gen-go v1.28.1
```
j0sh: Should we be generating with at least the same version of protoc-gen-go?

net/lp_rpc_grpc.pb.go
```diff
@@ -1,7 +1,7 @@
 // Code generated by protoc-gen-go-grpc. DO NOT EDIT.
 // versions:
-// - protoc-gen-go-grpc v1.5.1
-// - protoc             v3.12.4
+// - protoc-gen-go-grpc v1.2.0
```
j0sh: same re: protoc-gen-go-grpc

server/segment_rpc.go
```diff
@@ -254,6 +219,78 @@ func (h *lphttp) ServeSegment(w http.ResponseWriter, r *http.Request) {
 	w.Write(buf)
 }
 
+// Payment receives payment from Gateway and adds it into the orchestrator's balance
+func (h *lphttp) Payment(w http.ResponseWriter, r *http.Request) {
+	payment, segData, oInfo, ctx, err := h.processPaymentAndSegmentHeaders(w, r)
+	if err != nil {
+		return
+	}
+
```
j0sh, line 228: Double checking whether its OK to leave the body un-read here? If so, maybe leave a comment mentioning that is intentional?

j0sh (approved):
Double checking I understand the intent here -

The plan is to replicate the video workflow that sends the payments in as headers?

So in the currently proposed `/payments` endpoint, the payment data comes in as headers, with an empty body?

One more question - we have been calling this a "live" type of pipeline but here this is called "realtime" - do we want to align those terms?

Left a few small questions / comments through the PR but otherwise LGTM.

leszko:
> Double checking I understand the intent here -
> 
> The plan is to replicate the video workflow that sends the payments in as headers?
> 
> So in the currently proposed `/payments` endpoint, the payment data comes in as headers, with an empty body?
> 

Yes. The payment comes in headers just to reuse what we're currently using for payments. So, you'll send Payment in Headers and an empty body. To integrate it with the Transport protol, we'll need to call:
- `LivePaymentSender.SendPayment()` in Gateway anytime a segment (or a couple of segments) is sent
- `LivePaymentReceiver.AccountSegment()` in Orchestrator anytime a segment (or a couple of segments) is received

> One more question - we have been calling this a "live" type of pipeline but here this is called "realtime" - do we want to align those terms?

Yes, I'll update it to use "Live" instead.

> Left a few small questions / comments through the PR but otherwise LGTM.

Thank you 🙏

server/rpc.go
leszko: Thanks.

net/lp_rpc.pb.go
leszko: fair point, updated.

net/lp_rpc_grpc.pb.go
leszko: fair point, updated.

server/segment_rpc.go
leszko: I think you don't need to explicitly close the body on  the server side.

j0sh: Yeah, just that it is unusual so a comment explaining that would be nice.

BTW while I agree this is ok from a golang resource-management perspective, I have seen HTTP clients get blocked if the server ignores the body. Probably not an issue here since there is no body to begin with, but again just highlighting it is kind of unusual to not drain the body.

leszko: ok, I can add a comment (probably in some follow-up PR, since this one is merged).

PR #3231, opened 2024-11-05
monitor: gateway event queue

gioelecerati:
Adding an event queue for the gateway to publish events on Kafka.
This first PR introduces the producer and:
- New payment event

- Added parameters to specify Kafka bootstrap servers and credentials
- Added a kafka producer under monitor
- Create new payment event

**Checklist:**


monitor/census.go
```diff
@@ -1846,6 +1847,10 @@ func (cen *censusMetricsCounter) recordModelRequested(pipeline, modelName string
 		[]tag.Mutator{tag.Insert(cen.kPipeline, pipeline), tag.Insert(cen.kModelName, modelName)}, cen.mAIModelsRequested.M(1)); err != nil {
 		glog.Errorf("Failed to record metrics with tags: %v", err)
 	}
+	SendQueueEvent("ai_model_requested", map[string]string{
```
ecmulli: what is happening to trigger this event?

core/ai_worker.go
```diff
@@ -375,6 +381,14 @@ func (rw *RemoteAIWorker) Process(logCtx context.Context, pipeline string, model
 			monitor.AIResultDownloaded(logCtx, pipeline, modelID, chanData.DownloadTime)
 		}
 
+		monitor.SendQueueEvent("ai_worker_job_sent", map[string]string{
```
ecmulli: "job_sent" is this the worker sending a job to the runner or the gateway sending the job to the worker?

ecmulli (commented):
I tihnk there are too many event types. I'd like to understand them better and hopefully bucket them together.

core/ai_worker.go
```diff
@@ -358,6 +358,12 @@ func (rw *RemoteAIWorker) Process(logCtx context.Context, pipeline string, model
 	}
 
 	clog.V(common.DEBUG).Infof(logCtx, "Job sent to AI worker worker=%s taskId=%d pipeline=%s model_id=%s", rw.addr, taskID, pipeline, modelID)
+	monitor.SendQueueEvent("ai_worker_job_sent", map[string]string{
```
leszko: Let's make `SendQueueEvent()` call async and name it `SendQueueEventAsync()` to be specific that this call never blocks. And I believe it should never block.

monitor/census.go
```diff
@@ -1869,6 +1874,14 @@ func (cen *censusMetricsCounter) recordAIRequestLatencyScore(pipeline string, Mo
 	if err := stats.RecordWithTags(cen.ctx, tags, cen.mAIRequestLatencyScore.M(latencyScore)); err != nil {
 		glog.Errorf("Error recording metrics err=%q", err)
 	}
+	SendQueueEvent("ai_request_started", map[string]string{
```
leszko: Don't send events from `census.go`. Census should be related only to Prometheus metrics. If you want to do both (metrics and event) at the same time, please extract it to some helper function in a separate go file.

monitor/census.go
```diff
@@ -1930,6 +1961,11 @@ func (cen *censusMetricsCounter) recordAIJobPricePerUnit(pipeline string, Model
 		cen.mAIRequestPrice.M(pricePerUnit)); err != nil {
 		glog.Errorf("Error recording metrics err=%q", err)
 	}
+	SendQueueEvent("ai_request_price_per_unit", map[string]string{
```
leszko: I somehow feel that we have too many event types. It's up to @ecmulli , but I believe that we should not send "every event we can think of", but rather limit to what we need and what makes sense for the analytics. Maybe even you can both discuss and create a table with the event types that would make sense in the context of Gateway.

monitor/kafka.go
```diff
@@ -0,0 +1,126 @@
...
+		Dialer:   dialer,
+	})
+
+	return &KafkaProducer{
+		writer: writer,
+		topic:  topic,
+	}, nil
+}
+
+func (p *KafkaProducer) SendEvent(event GatewayEvent) {
```
leszko: Is it possible to make it package-private?

monitor/kafka.go
```diff
@@ -0,0 +1,126 @@
...
+	p.SendEvents([]GatewayEvent{event})
+}
+
+func (p *KafkaProducer) SendEvents(events []GatewayEvent) {
```
leszko: Is it possible to make it package-private?

monitor/kafka.go
```diff
@@ -0,0 +1,126 @@
...
+		if err == nil {
+			return
+		} else {
+			glog.Warningf("error while sending gateway log to Kafka, retrying, topic=%s, try=%d, err=%v", p.topic, i, err)
+		}
+	}
+	glog.Errorf("error while sending gateway log to Kafka, the gateway logs are lost, err=%d", err)
+}
+
+func SendQueueEvent(eventType string, data interface{}) {
```
leszko: I would make sure that this call is async and name it accordingly `SendQueueEventAsync()`. The thing is that I believe we should never want to block here. 

I think this would benefit from the [channel approach I took in `livepeer/catalyst-api`](https://github.com/livepeer/catalyst-api/blob/9264ab09dfcfcdfb2a379809076443784230a9d9/handlers/analytics/log_processor.go#L127). If you don't expect to have too many events, then you don't need to have timer and batch events, maybe it's ok to send them as they come. But I'd do the channel to make sending Kafka events async (without having to start a separate goroutine for each `SendQueueEvent()`)

leszko (commented):
@gioelecerati I general looks good. Added some comments, most minor. Two most important ones are:
1. Please think what event types we need and want to send (up to you and Evan)
2. Make the kafka calls async (they should never block normal execution)

monitor/census.go
gioelecerati: @ecmulli let's remove entirely the events from the census from this PR, I added them before the payments during the local testing phase, but now we want the payments data to get out in staging, so this addressed both the comments on the events and we remove the logic from census. Just made the commit, so that we're sending data exclusively on payment gen

monitor/kafka.go
gioelecerati: Makes total sense! Done, so now we have a similar implementation and the call is async

monitor/kafka.go
gioelecerati: Done

monitor/census.go
gioelecerati: As rafal pointed out, it doesn't make sense to publish events here where the metrics are produced. Removed them from this pr and left only the payment events

monitor/kafka.go
```diff
@@ -0,0 +1,129 @@
...
+	"time"
+
+	"github.com/golang/glog"
+	"github.com/google/uuid"
+	"github.com/segmentio/kafka-go"
+	"github.com/segmentio/kafka-go/sasl/plain"
+)
+
+const (
+	KafkaBatchInterval  = 1 * time.Second
```
leszko, line 17: Is this used anywhere?

monitor/kafka.go
```diff
@@ -0,0 +1,129 @@
...
+			}
+			glog.Warningf("error while sending gateway log to Kafka, retrying, topic=%s, try=%d, err=%v", p.topic, i, writeErr)
+		}
+		if writeErr != nil {
+			glog.Errorf("error while sending gateway log to Kafka, the gateway log is lost, err=%v", writeErr)
+		}
+	}
+}
+
+func (p *KafkaProducer) sendEvent(event GatewayEvent) {
```
leszko: Nit: Consider removing this function and inlining `p.events <- event`

monitor/kafka.go
```diff
@@ -0,0 +1,129 @@
...
+		Brokers:  []string{bootstrapServers},
+		Topic:    topic,
+		Balancer: kafka.CRC32Balancer{},
+		Dialer:   dialer,
+	})
+
+	return &KafkaProducer{
+		writer: writer,
+		topic:  topic,
+		events: make(chan GatewayEvent, 100),
```
leszko: Could you extract `100` to a const ?

monitor/kafka.go
```diff
@@ -0,0 +1,129 @@
...
+
+	event := GatewayEvent{
+		ID:        stringPtr(randomID),
+		Gateway:   stringPtr(""),
+		Type:      &eventType,
+		Timestamp: stringPtr(fmt.Sprint(timestampMs)),
+		Data:      data,
+	}
+
+	kafkaProducer.sendEvent(event)
```
leszko:
```suggestion
	select {
	case kafkaProducer.events <- event:
		// process event
	default:
		glog.Warningf("kafka producer event queue is full, dropping event")
	}
```

You need to change it, otherwise `kafkaProducer.sendEvent(event)` will block if the channel is full.

monitor/kafka.go
gioelecerati: Right! Wasn't managing batches, added them. Also addressed the 2 minor comments and handling the case where the queue is full

leszko (approved):
LGTM

PR #3232, opened 2024-11-06
Trickle subscribe on O and G

j0sh:
* Orchestrator monitors incoming stream
* Gateway subscribes to outgoing stream

Orchestrator also pre-creates the incoming and outgoing streams in order to avoid race conditions.

There is currently no content for the outgoing stream unless something actually creates it, eg by manually starting an inference process. Because of that, the HTTP subscriber on the gateway side will eventually time out. 

Depends on #3211

### DIFF
--- a/server/ai_http.go
+++ b/server/ai_http.go
@@ -94,6 +94,9 @@ func aiHttpHandle[I any](h *lphttp, decoderFunc func(*I, *http.Request) error) h
 func (h *lphttp) StartLiveVideoToVideo() http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 
+		remoteAddr := getRemoteAddr(r)
+		ctx := clog.AddVal(r.Context(), clog.ClientIP, remoteAddr)
+
 		// skipping handleAIRequest for now until we have payments
 
 		var (
@@ -110,6 +113,29 @@ func (h *lphttp) StartLiveVideoToVideo() http.Handler {
 			respondWithError(w, err.Error(), http.StatusInternalServerError)
 			return
 		}
+
+		// Precreate the channels to avoid race conditions
+		// TODO get the expected mime type from the request
+		pubCh := trickle.NewLocalPublisher(h.trickleSrv, mid, "video/MP2T")
+		pubCh.CreateChannel()
+		subCh := trickle.NewLocalPublisher(h.trickleSrv, mid+"-out", "video/MP2T")
+		subCh.CreateChannel()
+
+		// Subscribe to the publishUrl for payments monitoring
+		go func() {
+			sub := trickle.NewLocalSubscriber(h.trickleSrv, mid)
+			for {
+				segment, err := sub.Read()
+				if err != nil {
+					clog.Infof(ctx, "Error getting local trickle segment err=%v", err)
+					return
+				}
+				// We can do something with the segment data here
+				io.Copy(io.Discard, segment.Reader)
+			}
+		}()
+
+		// TODO subscribe to the subscribeUrl for output monitoring
 
 		respondJsonOk(w, jsonData)
 	})
--- a/server/ai_live_video.go
+++ b/server/ai_live_video.go
@@ -43,7 +43,7 @@ func startTrickleSubscribe(url *url.URL, params aiRequestParams) {
 		slog.Info("error getting pipe for trickle-ffmpeg", "url", url, "err", err)
 	}
 
-	// read segments from trickle subscrption
+	// read segments from trickle subscription
 	go func() {
 		defer w.Close()
 		for {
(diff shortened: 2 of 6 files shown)


server/ai_http.go
```diff
@@ -271,6 +274,29 @@ func (h *lphttp) StartLiveVideoToVideo() http.Handler {
 			return
 		}
 
+		// Precreate the channels to avoid race conditions
+		// TODO get the expected mime type from the request
+		pubCh := trickle.NewLocalPublisher(h.trickleSrv, mid, "video/MP2T")
```
leszko, line 119: Do I understand correctly that these 4 lines are not really needed, because the channel will be anyway created when requested from Gateway. However, we create them here, just because we want to subscribe them below and we don't want to have some race condition when subscribing to the non-existing channel? Is that correct?

server/ai_http.go
```diff
@@ -271,6 +274,29 @@ func (h *lphttp) StartLiveVideoToVideo() http.Handler {
...
+		go func() {
+			sub := trickle.NewLocalSubscriber(h.trickleSrv, mid)
+			for {
+				segment, err := sub.Read()
+				if err != nil {
+					clog.Infof(ctx, "Error getting local trickle segment err=%v", err)
+					return
+				}
+				// We can do something with the segment data here
+				io.Copy(io.Discard, segment.Reader)
```
leszko, line 134: We can check the payment here, but let's say the payment check fails (so Gateway hasn't paid for the stream). How do we stop all the processing from here?

server/ai_live_video.go
```diff
@@ -33,4 +36,39 @@ func startTricklePublish(url *url.URL, params aiRequestParams) {
...
 
 func startTrickleSubscribe(url *url.URL, params aiRequestParams) {
+	// subscribe to the outputs and send them into LPMS
+	subscriber := trickle.NewTrickleSubscriber(url.String())
+	r, w, err := os.Pipe()
+	if err != nil {
+		slog.Info("error getting pipe for trickle-ffmpeg", "url", url, "err", err)
+	}
+
+	// read segments from trickle subscrption
```
leszko:
```suggestion
	// read segments from trickle subscription
```

server/ai_http.go
j0sh: They are needed, because of race conditions:

* gateway may subscribe before the job sends anything into the output stream
* job may start subscribing before the gateway starts sending the input (as well as the O wanting to subscribe to the input itself)

server/ai_http.go
j0sh: there's a few things (but I think we can still get basic payments in without waiting on these)

- channels needs to be closed on the orch side - we can use the local publisher which already has a `Close()` method or add an API directly on the server for that
- runner needs to be notified - either via end-of-stream which is sent down the subscribe leg when a channel is closed, or explicit job RPC
- we need to prevent the same channels from being recreated on the O when the next publish comes in, so need to add either a local registry + auth callback to the trickle server, or disallow implicit channel creation via POST and keep pre-creating everything on the O locally.
- and probably some end-of-stream indicator on the publish leg, to send back to the gateway

server/ai_live_video.go
j0sh: fixed in https://github.com/livepeer/go-livepeer/pull/3232/commits/91282eec9d4b5eda97cfb16c1996b9a53ba10e73

leszko (approved):

server/ai_live_video.go
```diff
@@ -33,4 +36,39 @@ func startTricklePublish(url *url.URL, params aiRequestParams) {
...
+		}
+	}()
+
+	// lpms
+	go func() {
+		ffmpeg.Transcode3(&ffmpeg.TranscodeOptionsIn{
+			Fname: fmt.Sprintf("pipe:%d", r.Fd()),
+		}, []ffmpeg.TranscodeOptions{{
+			// TODO take from params
+			Oname:        "rtmp://localhost/out-stream",
```
mjh1, line 68: would this be going into mediaMTX?

j0sh: Yes this is hardcoded into mediamtx for now but we need to take a user configurable output destination. That is where you come in @mjh1 :)

j0sh:
Rebased onto master now that the ai-video merge is complete 🎉 

Also updated to latest trickle in https://github.com/livepeer/go-livepeer/pull/3232/commits/04d2b5facfa8f84a43e3ab893707b62e8d1f4933 ( + https://github.com/livepeer/go-livepeer/pull/3232/commits/1bae5551bc2184a7d206470984eed4cefbc7f841 ) to reflect changes in the [http-trickle](https://github.com/j0sh/http-trickle) repo

Plus a small bug fix https://github.com/livepeer/go-livepeer/pull/3232/commits/6e2b995640b7e532bdc90331fa735ef8daf9328f

PR #3239, opened 2024-11-07
Skip rtmp2segment logic for Windows.

leszko:
It broke the Windows builds and is not used anywhere other than Realtime Video AI.

CC: @j0sh I think you'll need to implement that part for Windows too at some point.


rickstaa (approved):
LGTM!

PR #3240, opened 2024-11-07
Implement mediaMTX RTMP auth

mjh1:
For the mediaMTX config I enabled the control api with `api: yes` and updated runOnReady: `runOnReady: curl http://localhost:5936/live/video-to-video/start -F stream=$MTX_PATH -F source_id=$MTX_SOURCE_ID`

Adding auth via a webhook call for the rtmp input stream.

We do also want to retrieve some config params as the response to the auth webhook but afaik those aren't defined yet and I figured it's probably nicer in a separate PR anyway. @j0sh if you'd rather I do it in this PR let me know or go a bit further with adding in placeholders.

- 
- 
- 

Tested locally running mediaMTX and the gateway manually

**Checklist:**


media/rtmp2segment.go
```diff
@@ -219,6 +223,7 @@ func processSegments(segmentHandler SegmentHandler, outFilePattern string, compl
 }
 
 func readSegment(segmentHandler SegmentHandler, file *os.File, pipeName string) {
+	log.Println("readSegment")
```
j0sh: stray? or at least `slog.Debug` ? (I think this is the only use of the `log` import in this file)

server/ai_mediaserver.go
```diff
@@ -360,10 +360,30 @@ func (ls *LivepeerServer) StartLiveVideo() http.Handler {
...
 		if streamName == "out-stream" {
 			// skip for now; we don't want to re-publish our own outputs
 			return
 		}
+
+		err := authenticateAIStream(AuthWebhookURL, AIAuthRequest{
+			Stream: streamName,
+		})
+		if err != nil {
+			kickErr := kickRTMPConnection(sourceID)
```
j0sh: If the source is WHIP will this still work?

media/rtmp2segment.go
```diff
@@ -36,15 +36,19 @@ func (ms *MediaSegmenter) RunSegmentation(in string, segmentHandler SegmentHandl
...
+	_, err := ffmpeg.Transcode3(&ffmpeg.TranscodeOptionsIn{
 		Fname: in,
 	}, []ffmpeg.TranscodeOptions{{
 		Oname:        outFilePattern,
 		AudioEncoder: ffmpeg.ComponentOptions{Name: "copy"},
 		VideoEncoder: ffmpeg.ComponentOptions{Name: "copy"},
 		Muxer:        ffmpeg.ComponentOptions{Name: "segment"},
 	}})
+	if err != nil {
+		slog.Error("Failed to run segmentation:", "err", err)
```
j0sh: Might be worth including the input here so it is easier to track which job it belongs to

```suggestion
		slog.Error("Failed to run segmentation", "in", in, "err", err)
```

server/auth.go
```diff
@@ -95,3 +95,39 @@ func (a authWebhookResponse) areProfilesEqual(b authWebhookResponse) bool {
 
 	return string(profilesA) == string(profilesB)
 }
+
+type AIAuthRequest struct {
+	Stream string `json:"stream"`
+	// TODO not sure what params we need yet
```
j0sh, line 101: later on, query params

j0sh (approved):
Looks good, just double checking it also works to terminate a WHIP stream.

media/rtmp2segment.go
mjh1: thanks yeah meant to remove 👍

server/ai_mediaserver.go
mjh1: ah 🤦 stupidly forgot, fixed that now 👍

Issue #273, opened 2018-01-18
CLI Improvement for Protocol Actions


ericxtang:
Currently the CLI allows you to do any transaction any time you want.  We should be able to do better than this - the protocol disallows certain actions depending on the state.

For example, you cannot bond more tokens than the amount you have, you cannot immediately bond after unbonding, etc.

ericxtang:
This is too broad.  Needs to be broken down into smaller issues.  It's a big project to implement all of the protocol constraints in the CLI.

ericxtang:
Closing.  Badly specified issue.

PR #3246, opened 2024-11-12
feat(ai): add hardware info from Orchestrators and expand network information available.

ad-astra-video:

This PR enables enhanced reporting on Livepeer network state for a Gateway to use in monitoring or analysis processes.  This could also be the start of exporting all relevant information from a Gateway to implement an external selection process using the orch webhook.

Orchestrator is updated in a couple ways to provide more information to the Gateway during polling process.  AI-worker and ai-runner are updated to retrieve hardware information at startup of the runner ([PR #273](https://github.com/livepeer/ai-worker/pull/273)).  The hardware information is stored in the ai worker list in memory and is removed when the ai-worker disconnects.  Orchestrators are also updated to enable providing additional information to Gateways with the `OrchestratorInfo` gRPC message to add hardware information and capabilities prices to provide Gateways all prices the Orchestrator has set at the time of the `GetOrchestrator` request.

Gateway is updated to save additional information from Orchestrators in the polling process.  This enables the Gateway to export useful information on the state of the network to external monitoring or analysis processes with the added `/getNetworkCapabilities` endpoint. 

[example data for an orchestrator](https://github.com/user-attachments/files/17716930/ad-astra-orch-info.json)

Open to all feedback and design change suggestions. I went with using the local db because this is not intended to serve a real time process and the polling process already saved some data to the db.

cc @ecmulli @gioelecerati 

__Gateway Updates__
- Shorten polling interval to 25 minutes to keep information fresh.
- Local db is updated to save `OrchestratorInfo` json received from each Orchestrator in the polling process in new `remoteInfo` column in the db `orchestrators` table.
- `/getNetworkCapabilities` endpoint is added to local cli webserver to enable exporting lightly formatted network information including `capabilities_prices` and `hardware` information deployed by Orchestrators for each pipeline/model id.
- `/getOrchestratorInfo` endpoint is added to local cli webserver to get raw `OrchestratorInfo` info from an Orchestrator. EDIT: this was removed

__Orchestrator Updates__

- Orchestrator is updated to provide additional information in the `OrchestratorInfo` response only when capabilities included in the `GetOrchestrator` request are `nil`.  All transcoding and AI jobs requests include capabilities which will prevent the additional information being included in the response to not further increase the size of the response and information processing during actual work.
- Hardware information is obtained by AI-Worker polling the runner containers for hardware information at startup and save the information in memory to provide to the Orchestrator at time of connection.
- Orchestrators store the hardware information in memory for each AI-Worker when the AI-Worker connects and is removed when the AI-Worker disconnects.
- `PriceInfo` gRPC message is updated to add two optional fields for capability and constraint to support sending `capabilities_prices` to Gateway

__AI Runner Updates__ [PR #273](https://github.com/livepeer/ai-worker/pull/273)

- AI-Runner container is updated to include new endpoints to provide the hardware information. `/hardware/info` provides the basic information on the hardware at startup and continues to be available after startup. `/hardware/stats` provides more focused information on utilization of the GPU to assist Orchestrators getting current information for monitoring.
    - Stats reporting needs additional build out in go-livepeer to support tracking what worker and gpu is assigned to a request.  I believe this would be a separate gRPC message and workflow so propose the go-livepeer build out of this is in a separate PR.
    - In the AI-Runner, the routes and pipelines need to be updated so the `/hardware/stats` endpoint is not blocked while the pipeline is running.  This can be achieved by adding  `outputs = await asyncio.to_thread([pipeline object], **kwargs)` and the routes updated to `return await pipeline([...])` as well as updating the route and pipeline call functions to be `async` functions where needed.  I tested this on the text-to-image and audio-to-text pipeline and confirmed the `/hardware/stats` endpoint is blocked until the pipeline call returns.

Built docker image and put on my mainnet Orchestrator

No

**Checklist:**


rickstaa:
Thanks! For future reference this was a follow up on https://github.com/livepeer/go-livepeer/pull/3052.

thomshutt:
@ad-astra-video is this ready for review?

ad-astra-video:
Yep! I will fix the editor config issue real quick.

common/db_test.go
```diff
@@ -334,6 +335,34 @@ func TestSelectUpdateOrchs_AddingUpdatingRow_NoError(t *testing.T) {
...
+	assert.Equal(updatedOrch[0].ServiceURI, serviceURIUpdate.ServiceURI)
+	assert.Equal(updatedOrch[0].ActivationRound, activationRoundUpdate.ActivationRound)
+	assert.Equal(updatedOrch[0].DeactivationRound, deactivationRoundUpdate.DeactivationRound)
+	assert.Equal(updatedOrch[0].PricePerPixel, priceUpdate.PricePerPixel)
+	assert.Equal(updatedOrch[0].Stake, stakeUpdate.Stake)
+	assert.Equal(updatedOrch[0].RemoteInfo, remoteInfoUpdate.RemoteInfo)
+}
+
+func TestUpdateOrch_AddRemoteInfo(t *testing.T) {
+
```
rickstaa: @ad-astra-video I think this test was lost in commits?

ad-astra-video: It was moved to test above, will delete this empty test.

core/livepeernode.go
```diff
@@ -261,6 +261,19 @@ func (n *LivepeerNode) GetBasePriceForCap(b_eth_addr string, cap Capability, mod
 	return nil
 }
 
+func (n *LivepeerNode) GetCapsPrices(b_eth_addr string) *CapabilityPrices {
```
leszko, line 272: Is this change related to this PR? I cannot connect the dots why we change anything about prices here.

net/lp_rpc.proto
```diff
@@ -160,8 +164,14 @@ message OrchestratorInfo {
   // Data for transcoding authentication
   AuthToken auth_token = 6;
 
+  // Information on Orchestrator hardware
+  bytes hardware = 7;
```
leszko: Why is it byte, shouldn't we give it some structure?

leszko (commented):
I started reviewing the PR, but I think I have some very high-level questions for now. Sorry if you already discussed it somewhere.

**1. Data transfer**
Why do we want to expose the Orchestrators data via the `/getNetworkCapabilities` endpoint in Gateway? I'm not sure what's the use case of this. Who will call this endpoint?

I think that if we want to collect and analyze the data, the better approach would be to:
- not store anything in the internal DB
- do a discovery from time to time, get the info, push to Kafka

**2. Data structure**
I think we should define what data we actually need and in what format. Now I see you're sending everything. So, the format between G<>O is undefinte (just `bytes`) and the JSON output format is just dumping everything here:  [example data for an orchestrator](https://github.com/user-attachments/files/17716930/ad-astra-orch-info.json). I think this will a pain while maintaining the backward-compatibility, because suddenly, you may have more data, etc. My suggestion is to keep the structure to the minimum, define with @ecmulli what we need and stick this this.

net/lp_rpc.proto
ad-astra-video: I used bytes in the grpc part to try and limit updates needed to the proto files.  go-livepeer has structs to parse the data when it is received by the Orchestrator

core/livepeernode.go
ad-astra-video: This is added to enable getting prices for all capabilities to be able to send all prices in the /getNetworkCapabilities endpoint on the local Gateway webserver.

ad-astra-video:
> 1. Data transfer
> Why do we want to expose the Orchestrators data via the /getNetworkCapabilities endpoint in Gateway? I'm not sure what's the use case of this. Who will call this endpoint?

The /getNetworkCapabilities endpoint is a local only data endpoint for the Gateway to get the current state of the network for capabilities and pricing to enable periodic polling every 30 minutes to get the network state to use in any way a Gateway sees fit.  I used discovery to make this information available because did not want to spam Orchestrators with GetOrchestrator requests if a Gateway sets up a fast refresh to polling the data. 

Some things this data endpoint can support:
- Gateway can use this as a dashboarding datasource for prices, hardware, capacity for all capabilities/models
- Gateway can see capacity/hardware available at different price levels
- This data can be used by a Gateway to adjust the Orchestrators included in the webhook endpoint to manage the Orchestrators it wants to work with in selection.

I don't like making Kafka a requirement to see this data as a Gateway unless we are adding documentation on setting up and using that by a Gateway.  That said, I do see the benefit of adding this to Kafka events to use if the Gateway chooses to though. I think Evan and Gioele can add that after this PR merges or can help add it to this one if want to.

On storing in the DB, can switch to storing in memory but since it does not support real time usage in selection or other parts of the code I thought the DB made more sense.  Also the Gateway was already setup to store updated information in the DB on the periodic refresh for discovery, just added more info being saved with it.

Note that the hardware and capabilities prices additional information is only sent back on discovery polling of orchestrators.  The Orchestrator looks for `nil` capabilities to include this extra data in this PR.  If want to also include the hardware information in session pools for selection we can update to only keep the capabilities prices out of the OrchestratorInfo if capabilities are sent in the GetOrchestrator request by the Gateway.  

> 2. Data structure
I think we should define what data we actually need and in what format. Now I see you're sending everything. So, the format between G<>O is undefinte (just bytes) and the JSON output format is just dumping everything here: [example data for an orchestrator](https://github.com/user-attachments/files/17716930/ad-astra-orch-info.json). I think this will a pain while maintaining the backward-compatibility, because suddenly, you may have more data, etc. My suggestion is to keep the structure to the minimum, define with @ecmulli what we need and stick this this.

The Hardware field is structured when it is in go-livepeer but is sent in OrchestratorInfo as bytes to avoid redefining the data in grpc proto files.  The Hardware field struct is defined by the data structure provided by the ai-runner in `runner.gen.go` and is pretty basic right now.  I used bytes in the grpc proto files to try to avoid maintaining two definitions of the data.  Since the information is not likely going to be exceptionally complicated we could redefine it in grpc proto files with comments added to indicate to match the runner.gen.go struct and add a couple tests to verify the definitions stay the same.  

Is there a way to pull the runner.gen.go definition into the grpc definitions?

The data path for the hardware information right now is:
1) ai-runner responds to /hardware/info at startup which the Orchestrator stores in HardwareInformation struct
2) Orchestrator provides the HardwareInformation up to the Gateway via OrchestratorInfo response on discovery requests only
3) Gateway uses the HardwareInformation provided for external data purposes only.

net/lp_rpc.proto
leszko: Not sure I understand it. I think this is already handled by proto itself, which is designed to be backwards (and forwards) compatible.

Could you please share an example of what upates in go-livepeer structs is a problem? Asking because we should favor structured data over unstructured data, so if we do an exception here, I'm trying to make sure we have a good reason to do it.

leszko:
@ad-astra-video 

Re 1: Hmm, it's very hard for me to comment on this without more context. Do we have any PRD with requirements or the bounty description, in general, the reason why we're implementing it? 

Asking because this is definitely not the right format for Inc Analytics, where we would prefer to push the data every 30 min. Now, here we're making it more complex, like store in internal DB (so making it persistent between go-livepeer restarts), exposing it with the HTTP endpoint, etc. So, we're making it more complex and not a fit for Inc. That's why I'm drilling this topic.

Re 2: Added a comment inline.

net/lp_rpc.proto
ad-astra-video: The `hardware` field is json bytes of the structs defined in `ai/worker/runner.gen.go`.  

Was just trying to avoid having two auto generated go structs so I converted it to bytes before sending through grpc.

I can also add these as new message definitions in proto file if prefer that.  

https://github.com/livepeer/go-livepeer/blob/9a2743592d7c8f299367a03dfb3bdfbb1c351fd8/ai/worker/runner.gen.go#L208-L215

https://github.com/livepeer/go-livepeer/blob/9a2743592d7c8f299367a03dfb3bdfbb1c351fd8/ai/worker/runner.gen.go#L239-L243

ad-astra-video:
@leszko I have updated this PR as suggested to:
- moved storing the network capabilities data to in memory and removed all changes related to DB
- added message definition for HardwareInformation and GPUComputeInfo to proto file
- added the information to Kafka events

The cli webserver endpoint is kept to support non-enterprise type Gateway setups access to this network information.  The cli webserver endpoint is very simple data export now with the changes made above.

@ecmulli Should the kafka event be one push with all Orchestrators info or a push for each orchestrator network info?  Right now I have it pushing each orchestrator network info separately under `network_capabilities`.

common/types.go
```diff
@@ -162,3 +162,16 @@ type OrchestratorStore interface {
 type RoundsManager interface {
 	LastInitializedRound() *big.Int
 }
+
+type NetworkCapabilities struct {
+	Orchestrators []*OrchNetworkCapabilities `json:"orchestrators"`
+}
+type OrchNetworkCapabilities struct {
+	Address            string                     `json:"address"`
```
leszko, line 170: @ecmulli Is this JSON structure ok for your analytics?

server/rpc.go
```diff
@@ -615,3 +629,18 @@ func coreSegMetadata(segData *net.SegData) (*core.SegTranscodingMetadata, error)
 		SegmentParameters:  &segPar,
 	}, nil
 }
+
+func workerHardwareToNetWorkerHardware(orchHdw []worker.HardwareInformation) []*net.HardwareInformation {
+	var workerHardware []byte
+	workerHardware, err := json.Marshal(orchHdw)
+	if err != nil {
```
leszko, line 642: Don't we want to print here some warning/error? I guess this should not happen, so it would be good to log something.

server/rpc.go
```diff
@@ -615,3 +629,18 @@ func coreSegMetadata(segData *net.SegData) (*core.SegTranscodingMetadata, error)
...
+		return []*net.HardwareInformation{}
+	}
+	var netWorkerHardware []*net.HardwareInformation
+	err = json.Unmarshal(workerHardware, &netWorkerHardware)
+	if err != nil {
```
leszko, line 648: Similar here, I think it's good that you return an empty slice, but I'd log something.

core/ai_orchestrator.go
```diff
@@ -1174,3 +1192,18 @@ func (n *LivepeerNode) transcodeFrames(ctx context.Context, sessionID string, ur
 	}
 	return &tr
 }
+
+func hardwareInformationFromNetHardware(hdw []*net.HardwareInformation) []worker.HardwareInformation {
+	var netWorkerHardware []byte
+	netWorkerHardware, err := json.Marshal(hdw)
+	if err != nil {
```
leszko, line 1199: Don't we want to log something here in case of an error?

core/orchestrator.go
```diff
@@ -254,6 +254,63 @@ func (orch *orchestrator) TicketParams(sender ethcommon.Address, priceInfo *net.
...
+				if err != nil {
+					glog.Errorf("error converting %v price for capability %v to int64 err=%w", ethAddr, CapabilityNameLookup[cap], err)
+					continue
+				}
+
+				//update default price to gateway price if exists, add price if default price is not set
+				capPriceUpdated := false
+				for idx, price := range capPrices {
+					if price.Capability == uint32(cap) && price.Constraint == modelID {
+						capPrices[idx] = &net.PriceInfo{PricePerUnit: priceInt64.Num().Int64(), PixelsPerUnit: 1, Capability: uint32(cap), Constraint: modelID}
```
leszko: If we set the `PixelsPerUnit` to `1`, we effectively loose the precision. E.g. Imagine that the `priceInfo` is `1/3`, then you'd set the price to `0` and `PixelsPerUnit` to `1`. 

Maybe we should change it and use something like this?
```golang
priceRat := priceInfo.Value()
...
&net.PriceInfo{PricePerUnit: priceRat.Num().Int64(), PixelsPerUnit: priceRat.Denom().Int64(), ... }
```

core/orchestrator.go
```diff
@@ -254,6 +254,63 @@ func (orch *orchestrator) TicketParams(sender ethcommon.Address, priceInfo *net.
...
+				if priceInfo == nil {
+					continue
+				}
+				priceInt64, err := common.PriceToInt64(priceInfo.Value())
+				if err != nil {
+					glog.Errorf("error converting %v price for capability %v to int64 err=%w", ethAddr, CapabilityNameLookup[cap], err)
+					continue
+				}
+
+				//update default price to gateway price if exists, add price if default price is not set
```
leszko: I'm not sure I understand what's going on here.
1. You check the "default" price and add it to `capPrices`. OK
2. Not you check the "perGateway" prices, but why don't you just add it to to output, but check this `capPriceUpdated`? Could you please describe it somewhere in the comment or refactor to some functions so that it's clear to the read what's going on?

discovery/db_discovery.go
```diff
@@ -20,7 +20,7 @@ import (
 	"github.com/golang/glog"
 )
 
-var cacheRefreshInterval = 1 * time.Hour
+var cacheRefreshInterval = 25 * time.Minute
```
leszko, line 24: Is there any reasoning why it's `25 min` and not `1 Hour`?

discovery/db_discovery.go
```diff
@@ -256,6 +258,10 @@ func (dbo *DBOrchestratorPoolCache) pollOrchestratorInfo(ctx context.Context) er
 }
 
 func (dbo *DBOrchestratorPoolCache) cacheDBOrchs() error {
+	//reset network capabilities
+	dbo.node.ResetNetworkCapabilities()
```
leszko: Why do we need to reset network capabilities here? And won't it cause some race condition that while updating the data it won't be available?

discovery/db_discovery.go
```diff
@@ -311,6 +317,13 @@ func (dbo *DBOrchestratorPoolCache) cacheDBOrchs() error {
...
 			return
 		}
+
+		//add response to network capabilities
+		err = dbo.node.AddNetworkCapabilities(orchInfoToOrchNetworkCapabilities(info, dbOrch))
+		if err != nil {
+			glog.Error("Error adding Orchestrator to network capabilities: ", err)
+		}
+
 		resc <- dbOrch
```
leszko: Are these `dbOrch` taken from the on-chain Registry or from the arg passed with `-orchAddr`? 

Asking, because I always thought that this comes from the registry and AFAIK we're not currently using the registry for AI Jobs, so we may actually not get the network data that we need.

net/lp_rpc.proto
```diff
@@ -477,3 +490,36 @@ message Payment {
   // O's last known price
   PriceInfo expected_price = 5;
 }
+
+// GPU information used by worker on network
+message GPUComputeInfo {
+  // unique id
```
leszko, line 499: What is this ID, is it the unique hardware id assigned the the specific GPU (like a serial number)?

net/lp_rpc.proto
```diff
@@ -477,3 +490,36 @@ message Payment {
...
+  string id = 1;
+
+  // gpu name
+  string name = 2;
+
+  // compute version supported - major
```
leszko, line 505: What are these versions? Nvidia driver versions?

leszko (commented):
Added some comments. I haven't tested it yet. btw. @ad-astra-video how have you tested it?

net/lp_rpc.proto
ad-astra-video: These are compute versions supported by the card.  ADA cards will send 8.9 and Ampere cards will send 8.6.  Older cards will send 7.x and 5.x depending on age of the card.  

The information is from here: https://github.com/livepeer/ai-runner/blob/24ec589c142c14f63374d2d9c1f771672c4f86f1/runner/app/utils/hardware.py#L89

net/lp_rpc.proto
ad-astra-video: Correct, a unique ID for the GPU

The hardware ID is provided by nvml in livepeer/ai-runner https://github.com/livepeer/ai-runner/blob/24ec589c142c14f63374d2d9c1f771672c4f86f1/runner/app/utils/hardware.py#L86

discovery/db_discovery.go
ad-astra-video: The `dbOrch` are pulled from the on-chain Registry with the polling Orchestrators started before the webhook pool is created. 
https://github.com/livepeer/go-livepeer/blob/be7b299da536f7d48dc13beb4facb64343fd6e8e/cmd/livepeer/starter/starter.go#L1459-L1461

If the -orchAddr and -orchWebhookUrl is using on-chain registered Orchestrators the information should be available for the set of Orchestrators wanted to work with.

In the scenario of using a private testnet or mixing in Orchestrators not registered on-chain, there could be a couple options:
1) refactor the info polling to be in the `orchestratorPool` rather than the `DBOrchestratorPoolCache`.  
2) add another info polling function to do the orchestrator info request to the urls in the `orchestratorPool` `GetInfos` urls if they are not in the network capabilities data after the on-chain orchestrator polling completes.
    - maybe relatively light lift to add a call to the n.OrchestratorPool to kick off updating info if not updated in the on-chain Orchestrator polling
3) or is there another way I am not seeing?

discovery/db_discovery.go
ad-astra-video: This is a good point.  Right now the data is not used for any decisions on selection or anything else so it won't be a operational issue I think.

The network capabilities are just a list right now but maybe better to switch to a map with the on-chain eth address as the key to avoid this.  Then can update the data in the map for each orchestrator specifically rather than delete and re-create the list each time.  Only issue I see with this is could get some stale data for Orchestrators that stop responding.

WDYT?  Could add a date/time in a `last_updated` field in the data to enable filtering stale data when using the data.

discovery/db_discovery.go
ad-astra-video: When we initially talked about this was thinking to shorten it to be every at least every thirty minutes to get fresh information.  

I would be fine switching back to 1 hr if preferred to leave it unchanged.

core/ai_orchestrator.go
ad-astra-video: added error log

server/rpc.go
ad-astra-video: added error log

server/rpc.go
ad-astra-video: added error log

core/orchestrator.go
ad-astra-video: I think your suggestion makes more sense for readibility.  To note, the price is converted to an Int64 on line 288 which uses a denominator of 1.  That function returns a rational though so I think is better to not assume it uses 1 as the denominator here.

The `PriceToInt64` function is used because I found in other testing that the numerator/denominator can overflow int64 in some of the AI pricing.  This would cause incorrect pricing to be returned by the orchestrator because the num/denom that is larger than int64 can hold would be truncated when Int64() was called on it.

core/orchestrator.go
ad-astra-video: The gateway prices override the default prices so I was trying to send the one price the gateway will see from the Orchestrator.

I did a light refactor by using an intermediary map to allow setting and overriding the prices with less if/then logic.  Will this work?

core/orchestrator.go
```diff
@@ -254,6 +255,60 @@ func (orch *orchestrator) TicketParams(sender ethcommon.Address, priceInfo *net.
...
+				if priceInfo == nil {
+					continue
+				}
+				priceInt64, err := common.PriceToInt64(priceInfo.Value())
+				if err != nil {
+					glog.Errorf("error converting %v price for capability %v to int64 err=%w", "default", CapabilityNameLookup[cap], err)
+					continue
+				}
+				capPriceName := strconv.Itoa(int(cap)) + "_" + modelID
+				capPricesMap[capPriceName] = &net.PriceInfo{PricePerUnit: priceInt64.Num().Int64(), PixelsPerUnit: 1, Capability: uint32(cap), Constraint: modelID}
```
leszko: I see you're still using `1` as denominator. I understand from your other comment that it's because you're afraid of overflowing int64. However, are we ok losing the precision here?

ad-astra-video: ah dang, that was a copy error on the refactor. Will fix.

discovery/db_discovery.go
leszko: No strong opinion on that.

core/orchestrator.go
leszko: Ok, looks better

discovery/db_discovery.go
leszko: What if you just revert the order and:
1. Fetch new data
2. Override capabilities

?

discovery/db_discovery.go
leszko: Hmm, I think we should change `DBOrchestratorPoolCache` to use whatever is specified in `-orchAddr` (or `-orchWebhookUrl`). I don't think it makes sense that it fetched the Os from the registry if you specified the mentioned flags.

net/lp_rpc.proto
leszko: 👌

ad-astra-video:
> Added some comments. I haven't tested it yet. btw. @ad-astra-video how have you tested it?

I tested it locally, wrote tests to cover the new data passed up on connection and run it on my test gateway. You can see the data on test gateway at https://livepeer-test-gateway.ad-astra.video/ai/network/capabilities.

I did another round of local testing and confirmed backwards compatible to v0.8.3.  Requests did not work to Orchs running v0.7.9 and v0.8.0.  I don't think we need to investigate versions that far back though.

I asked a couple Orchestrators to run this PR as well and confirmed that it will return data when runners are not updated.  It just returns either empty list or data for a pipeline/model without the hardware information.

Example of model reported from runner that is not updated to provide /hardware/info endpoint:
```
{
    "pipeline": "segment-anything-2",
    "model_id": "facebook/sam2-hiera-large"
}, {
    "pipeline": "audio-to-text",
    "model_id": "openai/whisper-large-v3"
}
```

discovery/db_discovery.go
ad-astra-video: I have updated to doing this way.

discovery/db_discovery.go
ad-astra-video: I have updated to use the OrchestratorPool urls so that `orchAddr` and `orchWebhookUrl` orchestrators specified will only be used for `pollOrchestratorInfo`.  Had to also change order in starter.go so the OrchestratorPool is setup before the polling starts.

leszko:
@ad-astra-video Could you check the failing CI Tests?

I'll do the final round of review and testing soon (today/tomorrow) and hopefully we can get it merged soon.

discovery/db_discovery.go
ad-astra-video: This change required some updates to tests that covered the DB Orchestrator Pool and the PollingOrchestratorInfo function. The test updates were not intended to change what the test was doing.  Fixes were to add TicketParams to the stub OrchestratorInfo and adjust the tests to fix race conditions I experienced running the tests locally. 

There is one other key update in this change to `cacheOrchInfos` where we cannot assume we get the on chain ethereum address from the DB in the polling.  I updated to get the on chain address from the ticket params and only update the DB with price information when the ticket params are not nil.  The only place the on chain ethereum address is is the Recipient field in the TicketParams.  I cannot think of a scenario when price is returned with no ticket params.

leszko (approved):
Great work @ad-astra-video 👏 

I'm approving and I'll merge it next week. Sorry that it took so long to review. Plus sorry not to merge it today, but it's mainly because of two reasons:
1. We have some other changes to get deployed soon into prod, so I'd prefer not to mix it
2. I need to double-check why I haven't received the Kafka event when testing your change in our staging; nothing to do from your side, because it may be some configuration on our side

All in all, I'm approving now and plan to merge your PR on Tue, Mon. Thanks again Brad!

leszko:
Ok, I'm merging this PR. I know why the Kafka events are not sent, but I think it's better if I send a follow-up PR. Really great work Brad! 🏅

PR #3251, opened 2024-11-14
trickle-server: Don't auto-create channels on first publish.

j0sh:
Channels can still be created locally on the server.

This is needed in case a payment fails then we need to close the channel but we don't want the publisher to re-open them.

We introduce a flag instead of just hard-coding the behavior because we depend on auto-creating in some [other tools](https://github.com/j0sh/http-trickle/commit/1b5b9e42eeb4a66e1384c4a48f8ec9112c953237) including the server on the demo box.

### DIFF
--- a/trickle/local_publisher.go
+++ b/trickle/local_publisher.go
@@ -28,11 +28,11 @@ func NewLocalPublisher(sm *Server, channelName string, mimeType string) *Trickle
 }
 
 func (c *TrickleLocalPublisher) CreateChannel() {
-	c.server.getOrCreateStream(c.channelName, c.mimeType)
+	c.server.getOrCreateStream(c.channelName, c.mimeType, true)
 }
 
 func (c *TrickleLocalPublisher) Write(data io.Reader) error {
-	stream := c.server.getOrCreateStream(c.channelName, c.mimeType)
+	stream := c.server.getOrCreateStream(c.channelName, c.mimeType, true)
 	c.mu.Lock()
 	seq := c.seq
 	segment, exists := stream.getForWrite(seq)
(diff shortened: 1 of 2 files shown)


leszko (approved):
Thanks a lot for this PR @j0sh Looks good, but I have two questions.

1. I believe we should delay the first payment check and allow all to start and check the first payment after 10, maybe 30s. In other words we take the optimistic approach that everything is fine not to delay the rest. What do you think? The alternative would be to wait with setting up everything until the first payment comes.

2. When the payment check fails, then I close the stream and it prevents the streaming from happening. All good. But I wonder about cleaning up other parts.
- Tear down runner => I guess @victorges implements some timeout, so it should happen automatically
- Any other objects to clean up on either Orchestrator or Gateway?

j0sh:
Thanks for the review @leszko 

> I believe we should delay the first payment check and allow all to start and check the first payment after 10, maybe 30s.

Possibly. We can send the first payment in along with the [initial request](https://github.com/livepeer/go-livepeer/blob/master/server/ai_process.go#L1011) like the other pipelines do it (eg via `prepareAIPayment`). Since the initial request is going to be reserving (scarce) capacity for the pipeline, it might be worth having a "down payment" of sorts, to ensure the gateway has the ability to pay for the job. But if this risk seems manageable then we can start checking after 10-30s.

> Tear down runner => I guess @victorges implements some timeout, so it should happen automatically

The runner will receive an EOS (end of stream) flag and stop subscribing ([PR](https://github.com/livepeer/ai-worker/pull/271)). I think the rest of the pipeline is also terminated after that, but I haven't tested recently.

> Any other objects to clean up on either Orchestrator or Gateway?

Right now, the bookkeeping on the orchestrator is minimal (just publish and subscribe streams IIRC) but I am sure we will have more to clean up later. The gateway also cleans itself up after a moment or so, but I do have some improvements to make on the publisher side so tear-down is a little more controllable, rather than the current backpressure-led error chain.

PR #3253, opened 2024-11-15
allow scientific price notation

rickstaa:

This pull request ensures that orchestrators can specify pricing using the scientific format.

- 
- 
- 

**Checklist:**

- [x] README and other documentation updated -> https://github.com/livepeer/docs/pull/689.


CHANGELOG_PENDING.md
```diff
@@ -1,5 +1,8 @@
 # Unreleased Changes
 
+- [#3253](https://github.com/livepeer/go-livepeer/pull/3253) - Allow orchestrators to specify pricing using scientific notation.
+- [#3248](https://github.com/livepeer/go-livepeer/pull/3248) - Provide AI orchestrators with a way to specify `pricePerUnit` as a float.
```
rickstaa, line 4: @leszko, @ad-astra-video I noticed I forgot this in the last pull request.

rickstaa:
The docs also have been updated in https://github.com/livepeer/docs/pull/689.

common/util.go
```diff
@@ -497,3 +497,20 @@ func MimeTypeToExtension(mimeType string) (string, error) {
 	}
 	return "", ErrNoExtensionsForType
 }
+
+// ParsePricePerUnit parses a price string in the format <price><exponent><currency> and returns the price as a big.Rat and the currency.
+func ParsePricePerUnit(pricePerUnitStr string) (*big.Rat, string, error) {
+	pricePerUnitRex := regexp.MustCompile(`^(\d+(\.\d+)?([eE][+-]?\d+)?)([A-Za-z][A-Za-z0-9]*)?$`)
```
leszko, line 503: Is this change backward-compatible? I see you use a different regex here, so I wonder, won't it break anything for the existing config?

core/ai.go
```diff
@@ -82,9 +83,51 @@ type AIModelConfig struct {
 	Currency      string  `json:"currency,omitempty"`
 }
 
+// UnmarshalJSON allows `PricePerUnit` to be specified as a string.
+func (s *AIModelConfig) UnmarshalJSON(data []byte) error {
```
leszko, line 87: I see you cover different scenarios here, like `price_per_unit` can be a string or an int (or float?). Could you add some units for that to make sure it works for all these cases?

common/util.go
victorges: From a quick glimpse, it seems like the incompatibility would be if we had a currency being specified as `[eE][0-9]*` before, which will be detected as a scientific notation now. I think that case is pretty unlikely tho, but one solution could be to add a required space after the number in scientific notation (after `\d+`), which could be a good practice anyway for clearer `123e-12 USD` instead of a single `123e-12USD`. WDYT?

common/util_test.go
```diff
@@ -476,3 +477,119 @@ func TestMimeTypeToExtension(t *testing.T) {
...
+		{
+			name:             "Valid price with scientific notation and currency",
+			pricePerUnitStr:  "1.23e2USD",
+			expectedPrice:    big.NewRat(123, 1),
+			expectedCurrency: "USD",
+			expectError:      false,
+		},
+		{
+			name:             "Valid price with capital scientific notation and currency",
+			pricePerUnitStr:  "1.23E2USD",
```
victorges, line 547: Yeah I think having a space after the exponential notation would be helpful

core/ai.go
```diff
@@ -82,9 +83,51 @@ type AIModelConfig struct {
...
+	switch v := aux.PricePerUnit.(type) {
+	case string:
+		pricePerUnit, currency, err := common.ParsePricePerUnit(v)
+		if err != nil {
+			return fmt.Errorf("error parsing price_per_unit: %v", err)
+		}
+		price = JSONRat{pricePerUnit}
+		if s.Currency == "" {
+			s.Currency = currency
+		}
```
victorges, line 111: Hmm I see what you are doing there. Is it worth having this logic compared to just keeping the currency in a separate field in the JSON? This provides multiple ways of doing the same thing and leads to ambiguous behavior (e.g. what happens if I specify both the `"currency"` and add a currency to the `pricePerUnit`?)

AFAIK the go JSON lib (and the `JSONRat` impl) already supports scientific notation, so nothing we need to do add that support there.

If we do go with the new way of specifying the price, I believe we should implement it for the other price config JSON's as well, like the video gateway for price per O.

victorges: Thinking further about this, maybe we remove this change on the JSON config, which is redundant, and make it only for the CLI flags, wdyt?

victorges (approved):
LGTM with a couple suggestions

common/util.go
leszko: Yeah, I'm fine even with some "slight" non-backward-compatibility as long as it's ok with Orchestrators.

PR #3255, opened 2024-11-15
Add Control API

leszko:
Add Control API for AI Live Inference. Related AI Worker PR: https://github.com/livepeer/ai-worker/pull/279

**Comments**
1. I've updated how we the path for starting the pipeline to `/live/video-to-video/{stream}/start`(so the stream name is passed as a path param instead of the form param). @victorges @mjh1 @thomshutt let me know if you prefer the form param.
2. There is a slight difference between the [Control API design doc](https://www.notion.so/livepeer/Realtime-AI-Video-Control-API-1300a3485687800197ecc87109260c1b) (the reason for that is that during the implementation I realized it's actually simpler to use Trickle directly from Gateway)
    - Design Doc: `User --HTTP--> Gateway --HTTP--> Orchestrator --Trickle--> Runner`
    - Implementation `User --HTTP--> Gateway --Trickle--> Orchestrator --Trickle--> Runner`
3. There is no retry logic and no unit tests, I think that it's ok for this first version, but something to keep in mind that we'll need to address this tech debt after the 12/13 deadline. It's related not only to this PR, but to all AI Inference code we added.

**How to test it?**

Start MediaMTX with the following config
```yaml
runOnReady: curl http://localhost:5936/live/video-to-video/$MTX_PATH/start -F source_id=$MTX_SOURCE_ID -F source_type=$MTX_SOURCE_TYPE -F query=$MTX_QUERY
```

Start Gateway
```
./livepeer -gateway -rtmpAddr :1936 -httpAddr :5936 -orchAddr localhost:8935 -v 6
```

Start Orchestrator
```
./livepeer -orchestrator -aiWorker -aiModels 'live-video-to-video:cumulo-autumn/stream-diffusion:false' -serviceAddr localhost:8935 -transcoder -v 6
```

Start streaming by accessing the following URL from the browser: http://localhost:8889/stream/publish?video-codec=h264%2F90000

Start Runner and then execute the following HTTP request to the runner to start inference
```
curl 'http://<runner-ip-and-port>/live-video-to-video/' \
--header 'Content-Type: application/json' \
--data '{
  "subscribe_url": "https://host.docker.internal:8935/ai/trickle/<mid>",
  "publish_url": "https://host.docker.internal:8935/ai/trickle/<mid>-out",
  "control_url": "https://host.docker.internal:8935/ai/trickle/<mid>-control"
}'
```

Now, you can send Control API request, e.g.
```
curl --location 'http://localhost:5936/live/video-to-video/stream/update' \
--header 'Content-Type: application/json' \
--data '{
  "some-param": "some-param-value"
}'
```


server/ai_mediaserver.go
```diff
@@ -79,6 +79,7 @@ func startAIMediaServer(ls *LivepeerServer) error {
 
 	// This is called by the media server when the stream is ready
 	ls.HTTPMux.Handle("/live/video-to-video/start", ls.StartLiveVideo())
+	ls.HTTPMux.Handle("/live/video-to-video/{stream}/update", ls.StartLiveVideo())
```
j0sh: We could also make the `/start` path above this take the stream  name too, for consistency. (If we change this we should do it soon, before the official release. No real preference though.)

server/ai_process.go
```diff
@@ -1037,13 +1039,35 @@ func submitLiveVideoToVideo(ctx context.Context, params aiRequestParams, sess *A
...
+		clog.V(common.VERBOSE).Infof(ctx, "pub %s sub %s control %s", pub, sub, control)
 		startTricklePublish(pub, params)
 		startTrickleSubscribe(sub, params)
+		startControlPublish(control)
 	}
 	return resp, nil
 }
 
+// TODO: Remove it and get requests from POST /stream/update
+func startControlPublish(control *url.URL) {
```
j0sh: This file is getting  large so I have been putting the live video stuff into the [ai_live_video.go](https://github.com/livepeer/go-livepeer/blob/master/server/ai_live_video.go)  file, that is where the `startTricklePublish` and `startTrickleSubscribe` stuff lives

j0sh (commented):
Looks like a good start!

server/ai_mediaserver.go
leszko: I updated to use path everywhere. I _slightly_ prefer it. Let's see what @mjh1 @thomshutt and @victorges say. I'm also ok to keep everywhere the form param.

server/ai_process.go
leszko: Moved.

server/ai_live_video.go
```diff
@@ -83,3 +84,13 @@ func mediamtxSourceTypeToString(s string) (string, error) {
 		return "", errors.New("unknown media source")
 	}
 }
+
+func startControlPublish(control *url.URL, params aiRequestParams) {
+	controlPub, err := trickle.NewTricklePublisher(control.String())
+	if err != nil {
+		slog.Info("error starting control publisher stream=%s err=%v", "err", params.stream, err)
```
victorges: Should we `return` after this err? Or we can create the pipeline with an empty `ControlPub` here and have a check on [the handler](https://github.com/livepeer/go-livepeer/pull/3255/files#diff-88f3e5816c6c29e92773629b388dfc58ed59a3682ab624c7f3448cfda3106475R487) first and return some 422 or sth

server/ai_mediaserver.go
```diff
@@ -452,6 +455,48 @@ func (ls *LivepeerServer) StartLiveVideo() http.Handler {
...
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+			return
+		}
+	})
+}
+
+func (ls *LivepeerServer) cleanupLive(stream string) {
+	ls.LivepeerNode.LiveMu.Lock()
+	defer ls.LivepeerNode.LiveMu.Unlock()
+	delete(ls.LivepeerNode.LivePipelines, stream)
```
victorges, line 498: Do we need to close the trickle publisher or sth like that?

victorges (approved):
Looks good!

server/ai_live_video.go
leszko: I think we should return. Updated.

server/ai_mediaserver.go
leszko: I think that no, IMO it should be deleted on the Orchestrator's side. We should not depend on closing the trickle channel from Gateway

j0sh: I agree that we might not want to depend on closing the control channel from the gateway. But we also might want to give the orchestrator a heads-up that the session is complete, so it can release resources attached to the job. One easy way to do this would be to close the _publish_ channel which we do here

https://github.com/livepeer/go-livepeer/blob/6a8fa52c87c4a1fab11d6dce1db3f73b7d8410f2/server/ai_live_video.go#L23-L28

However, the general tear-down process between GOWR is something that still needs some work. Especially transitions when switching Os ... but we can figure that out after the launch when we are ready to tackle selection and related work.

go.sum
```diff
@@ -607,6 +607,14 @@ github.com/libp2p/go-openssl v0.1.0 h1:LBkKEcUv6vtZIQLVTegAil8jbNpJErQ9AnT+bWV+O
...
 github.com/livepeer/ai-worker v0.12.1 h1:V6XGxnRmq02GuJP/PYRKTXIZAd2F1jTuErweO5boWxU=
 github.com/livepeer/ai-worker v0.12.1/go.mod h1:/Deme7XXRP4BiYXt/j694Ygw+dh8rWJdikJsKY64sjE=
+github.com/livepeer/ai-worker v0.12.2-0.20241115144717-cffea497e22e h1:/rGRpiHAhaCzk8fV+RXjkVILea8wSoqpuDQY/CdpilY=
+github.com/livepeer/ai-worker v0.12.2-0.20241115144717-cffea497e22e/go.mod h1:/Deme7XXRP4BiYXt/j694Ygw+dh8rWJdikJsKY64sjE=
+github.com/livepeer/ai-worker v0.12.2-0.20241115145337-b8477a4881ae h1:suFeJeJqm+PKrtJJO7ubkb9m+6CZPOHbO0xQmxycUEI=
+github.com/livepeer/ai-worker v0.12.2-0.20241115145337-b8477a4881ae/go.mod h1:/Deme7XXRP4BiYXt/j694Ygw+dh8rWJdikJsKY64sjE=
+github.com/livepeer/ai-worker v0.12.2-0.20241115145705-4cefa2dc7463 h1:k81OkHOR4EBPEBcM8DhFUmkdiL4BdMyZMgwmJ2N/0+E=
+github.com/livepeer/ai-worker v0.12.2-0.20241115145705-4cefa2dc7463/go.mod h1:/Deme7XXRP4BiYXt/j694Ygw+dh8rWJdikJsKY64sjE=
+github.com/livepeer/ai-worker v0.12.2-0.20241118111050-0a9cd07bca51 h1:+mKaLa8DvxxBJd9KTpbfN48fse+34HteqavOpCklDlI=
+github.com/livepeer/ai-worker v0.12.2-0.20241118111050-0a9cd07bca51/go.mod h1:/Deme7XXRP4BiYXt/j694Ygw+dh8rWJdikJsKY64sjE=
```
j0sh: go mod tidy?

server/ai_mediaserver.go
```diff
@@ -452,6 +455,48 @@ func (ls *LivepeerServer) StartLiveVideo() http.Handler {
...
+		}
+		defer r.Body.Close()
+		params, err := io.ReadAll(r.Body)
+		if err != nil {
+			http.Error(w, err.Error(), http.StatusBadRequest)
+			return
+		}
+
+		clog.V(6).Infof(ctx, "Sending Live Video Update Control API stream=%s, params=%s", stream, string(params))
+		if err := p.ControlPub.Write(strings.NewReader(string(params))); err != nil {
```
j0sh, line 488: Are we also planning on returning a response from the inference process?

If so, can be a separate PR. As the doc mentions, that would be a more involved process with a separate control response subscription. The params would also need to be wrapped in some sort of transaction id to reconcile request/response.

However, if we _don't_ need a response then we can keep things as-is since that is much simpler.

server/ai_process.go
```diff
@@ -89,6 +89,7 @@ type aiRequestParams struct {
 	// For live video pipelines
 	segmentReader *media.SwitchableSegmentReader
 	outputRTMPURL string
+	stream        string
```
j0sh, line 92: I think we will be adding a bunch more stuff to this struct and part of me wonders if we should split the live pipeline fields into its own sub-struct to keep `aiRequestParams` tidy. Doesn't need to be done in this PR though.

j0sh (approved):
Looks great!

server/ai_mediaserver.go
leszko: Yeah, good point. I think we can leave it without a response for now. 

But in the future, we'll probably need a way to signal if the update failed for some reason.

go.sum
leszko: Update in ac9b327ce52369c1e89ba7a01e319e6d471f57a0

server/ai_process.go
leszko: Yeah, you can refactor it in the next PR. I think it will be better as you described it.

server/ai_mediaserver.go
leszko: Ok, added closing the channel. But yeah, I share the same feelings as you that we should work on the tear-down process. Plus we need some retries process for resilience.

PR #3256, opened 2024-11-18
Use RemoteAddr for calls back to mediamtx

mjh1:

To support running mediaMTX on a different pod/machine to the gateways I am now using the remote address from the request from mediaMTX for any requests that the gateway has to make back to mediaMTX. I've also added auth on the mediaMTX API calls.

Tested with a temporary deploy to staging.

**Checklist:**

### DIFF
--- a/docker/Dockerfile.mediamtx
+++ b/docker/Dockerfile.mediamtx
@@ -1,0 +1,9 @@
+FROM ubuntu:20.04
+
+RUN apt update
+RUN apt install -y curl
+
+RUN curl -L https://github.com/bluenviron/mediamtx/releases/download/v1.9.3/mediamtx_v1.9.3_linux_amd64.tar.gz -o /mediamtx.tar.gz
+RUN tar xzvf /mediamtx.tar.gz
+
+ENTRYPOINT ["/mediamtx"]
--- a/docker/Dockerfile.mediamtx
+++ b/docker/Dockerfile.mediamtx
@@ -5,5 +5,8 @@ RUN apt install -y curl
 
 RUN curl -L https://github.com/bluenviron/mediamtx/releases/download/v1.9.3/mediamtx_v1.9.3_linux_amd64.tar.gz -o /mediamtx.tar.gz
 RUN tar xzvf /mediamtx.tar.gz
+RUN rm /mediamtx.tar.gz
+
+COPY docker/mediamtx.yml /mediamtx.yml
 
 ENTRYPOINT ["/mediamtx"]
--- a/server/ai_mediaserver.go
+++ b/server/ai_mediaserver.go
@@ -387,6 +387,14 @@ func (ls *LivepeerServer) StartLiveVideo() http.Handler {
 		}
 		ctx = clog.AddVal(ctx, "source_type", sourceType)
 
+		remoteAddr := r.RemoteAddr
+		if remoteAddr == "" {
+			clog.Errorf(ctx, "remoteAddr is empty")
+			http.Error(w, "Could not find callback address", http.StatusBadRequest)
+			return
+		}
+		ctx = clog.AddVal(ctx, "remote_addr", remoteAddr)
+
 		queryParams := r.FormValue("query")
 		qp, err := url.ParseQuery(queryParams)
 		if err != nil {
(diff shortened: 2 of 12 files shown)


server/ai_mediaserver.go
```diff
@@ -389,6 +389,14 @@ func (ls *LivepeerServer) StartLiveVideo() http.Handler {
 		}
 		ctx = clog.AddVal(ctx, "source_type", sourceType)
 
+		remoteHost, err := getRemoteHost(r.RemoteAddr)
```
j0sh, line 392: This is OK if it gets us going for now. 

I can see more sophisticated setups where this might not really work, eg proxies or if RTMP is listening on a nonstandard port. For those cases I would expect another form parameter in the curl (including possibly a separate field for the control URL if its also on a nonstandard port)

docker/Dockerfile.mediamtx
```diff
@@ -0,0 +1,9 @@
+FROM ubuntu:20.04
+
+RUN apt update
+RUN apt install -y curl
+
+RUN curl -L https://github.com/bluenviron/mediamtx/releases/download/v1.9.3/mediamtx_v1.9.3_linux_amd64.tar.gz -o /mediamtx.tar.gz
+RUN tar xzvf /mediamtx.tar.gz
```
j0sh, line 7: Would it be useful to also include our own default mediamtx.yml in here?

server/ai_mediaserver.go
```diff
@@ -518,9 +540,14 @@ func kickInputConnection(sourceID string, sourceType string) error {
 		return fmt.Errorf("invalid sourceType: %s", sourceType)
 	}
 
-	resp, err := http.Post(fmt.Sprintf("http://localhost:%s/v3/%s/kick/%s", mediaMTXControlPort, apiPath, sourceID), "", nil)
+	req, err := http.NewRequest(http.MethodPost, fmt.Sprintf("http://%s:%s/v3/%s/kick/%s", mediaMTXHost, mediaMTXControlPort, apiPath, sourceID), nil)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to create kick request: %w", err)
+	}
+	req.SetBasicAuth(mediaMTXControlUser, ls.mediaMTXApiPassword)
```
j0sh: Can we make so that this also works without mediamtx auth configured? The default binding for the control port is localhost anyway.

server/ai_mediaserver.go
```diff
@@ -505,9 +524,12 @@ func (ls *LivepeerServer) cleanupLive(stream string) {
...
 }
 
-const mediaMTXControlPort = "9997"
+const (
+	mediaMTXControlPort = "9997"
+	mediaMTXControlUser = "admin"
+)
 
-func kickInputConnection(sourceID string, sourceType string) error {
+func (ls *LivepeerServer) kickInputConnection(mediaMTXHost, sourceID, sourceType string) error {
```
j0sh: I have been putting live AI specific stuff into [`ai_live_video.go`](https://github.com/livepeer/go-livepeer/blob/master/server/ai_live_video.go) to minimize the sprawl in this file - might be good to do it for this function too, but doesn't have to be done in this PR. I suspect we'll need to call this function in other areas too when we need to force terminate the stream.

j0sh (approved):
Looks good overall! Just one request about making mediamtx kicks work without auth configured, and a couple other small notes but otherwise LGTM

server/ai_mediaserver.go
mjh1: Oh yep that still works, in the mediamtx config I've left in the default localhost user with no password

server/ai_mediaserver.go
mjh1: Yep I guess we can configure an environment variable in the mediamtx pod for the callback hostname, then that would be available to the curl command.

docker/Dockerfile.mediamtx
mjh1: Ah ok yeah just for local development with localhost configured for the gateway call? 👍

server/ai_mediaserver.go
mjh1: will do

PR #3259, opened 2024-11-18
ai/live: Improved error handling for trickle publishers

j0sh:
This allows us to propagate normal HTTP errors down to the caller
of `publisher.Write()`

Currently we don't distinguish between 400/500 status codes
(except for 404s which we do explicitly handle) but there is room 
to handle this in the future via separate error types.

[trickle: Publisher error handling](https://github.com/livepeer/go-livepeer/pull/3259/commits/69d643bdb72c62664a0834340da1aa64604d3a92)

Also adds the following optimization to minimize large-ish allocations:
[media: Reuse buffer when reading segments from ffmpeg.](https://github.com/livepeer/go-livepeer/pull/3259/commits/c82b4e5b8468307f2c65bd0f9634619f4d8df02b)

### DIFF
--- a/trickle/trickle_publisher.go
+++ b/trickle/trickle_publisher.go
@@ -2,12 +2,15 @@ package trickle
 
 import (
 	"crypto/tls"
+	"errors"
 	"fmt"
 	"io"
 	"log/slog"
 	"net/http"
 	"sync"
 )
+
+var StreamNotFoundErr = errors.New("stream not found")
 
 // TricklePublisher represents a trickle streaming client
 type TricklePublisher struct {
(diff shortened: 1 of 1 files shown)


leszko (approved):

mjh1 (approved):

media/rtmp2segment.go
```diff
@@ -186,6 +186,9 @@ func processSegments(segmentHandler SegmentHandler, outFilePattern string, compl
 	pipeNum := 0
 	createNamedPipe(fmt.Sprintf(outFilePattern, pipeNum))
 
+	// shared buffer; set to empty and segment reader will initialize
+	buf := []byte{}
```
victorges: **nit:** Could consider using a `bytes.NewBuffer(nil)` as well. With some existing impl like [`ReadFrom`](https://cs.opensource.google/go/go/+/go1.23.4:src/bytes/buffer.go;l=206) we can avoid manually handling buffer sizes/increases etc. 

That would involve a refactor tho so def not a must. Just leaving the idea mostly for the future.

victorges (approved):
LGTM

media/rtmp2segment.go
j0sh: Yep we need to switch to a buffered reader soon to decouple reads from writes, since the right now this can lead to back-pressure if RTMP segment reads out-pace trickle writes.

With that being said, this specific change might end up being redundant after all, so I think I'll just discard the commit entirely (theres a merge conflict with it right now anyway)

victorges: Yeah the `bytes.Buffer` wouldn't solve back pressure as it would return an EOF after each segment. I'm always susprised how there's no buffered pipe built-in in Go. Shouldn't be hard to make one tho

PR #3261, opened 2024-11-20
Live Video Payments

leszko:
Introduce payments into Realtime Video AI.

**How does it work?**
1. Gateway calls Orchestrator to start AI Pipeline (including the [initial payment](https://github.com/livepeer/go-livepeer/pull/3261/files#diff-22d416256190d3c477afb48f77471d17c116dd4c3aa87489be5a286c7611ca89R1009)
2. Orchestrator accepts the payment, starts the pipeline (trickle channels, AI runner, etc.)
3. Orchestrator starts payment processor to periodically probe incoming segments and account payment
4. Gateway starts the payment processor to periodically probe incoming segments and send payments

**Additional comments**
- I removed "segmentsToPayUpfront", because the [initial payment](https://github.com/livepeer/go-livepeer/pull/3261/files#diff-22d416256190d3c477afb48f77471d17c116dd4c3aa87489be5a286c7611ca89R1009) acts as a payment buffer to make sure we won't have race condition between sending the payment (in Gateway) and accounting the payment (in Orchestrator)
- I've hardcoded the [initial payment](https://github.com/livepeer/go-livepeer/pull/3261/files#diff-22d416256190d3c477afb48f77471d17c116dd4c3aa87489be5a286c7611ca89R1009) to `30 seconds, 30fps, 720p`, we could make it a flag or change it, but I think that this const seems good enough
- Sorry for not adding unit tests, but it's quite difficult right now and we're in rush for the Dec deadline


server/ai_http.go
```diff
@@ -125,7 +143,31 @@ func (h *lphttp) StartLiveVideoToVideo() http.Handler {
...
+			err := paymentReceiver.AccountPayment(context.Background(), &SegmentInfoReceiver{
+				sender:    sender,
+				inPixels:  inPixels,
+				priceInfo: priceInfo,
+				sessionID: mid,
+			})
+			if err != nil {
+				slog.Warn("Error accounting payment, stopping stream processing", "err", err)
+				// TODO: Check if we do not need to any other cleanup
+				pubCh.Close()
```
leszko: @j0sh is it how we plan to stop the stream (by closing the streams) and also tear down the container (work related to what @victorges is doing)?

server/ai_live_video.go
```diff
@@ -1,33 +1,52 @@
...
 		// check for end of stream
 		if _, eos := reader.(*media.EOSReader); eos {
 			if err := publisher.Close(); err != nil {
 				slog.Info("Error closing trickle publisher", "err", err)
 			}
+			cancel()
 			return
 		}
 		go func() {
+			reader := paymentProcessor.process(reader)
```
leszko: @j0sh Is it how you imagined tweaking into the segment processing logic?

server/ai_http.go
```diff
@@ -93,18 +94,35 @@ func aiHttpHandle[I any](h *lphttp, decoderFunc func(*I, *http.Request) error) h
...
+		ctx := clog.AddVal(r.Context(), clog.ClientIP, remoteAddr)
+
+		// handle initial payment, the rest of the payments are done separately from the stream processing
+		payment, err := getPayment(r.Header.Get(paymentHeader))
+		if err != nil {
+			respondWithError(w, err.Error(), http.StatusPaymentRequired)
+			return
+		}
+		sender := getPaymentSender(payment)
+		_, ctx, err = verifySegCreds(ctx, h.orchestrator, r.Header.Get(segmentHeader), sender)
```
leszko, line 143: Somewhere inside `verifySegCreds()` there is capacity check, something that will be related to the selection and session management (CC: @mjh1 ).

server/live_payment_processor.go
```diff
@@ -0,0 +1,155 @@
...
+	pipeReader, pipeWriter, err := os.Pipe()
+	if err != nil {
+		slog.Error("Error creating pipe", "err", err)
+		return reader
+	}
+
+	resReader := io.TeeReader(reader, pipeWriter)
+	go func() {
+		// read the segment into the buffer, because the direct use of the reader causes Broken pipe
+		// it's probably related to different pace of reading by trickle and ffmpeg.GetCodecInfo()
```
j0sh, line 98: This might be because probing isn't guaranteed to read the entire pipe, so if you tee the input but only partially read one side, the other will block until both ends have read. Try discarding the remainder of the reader after probing - does that help?

server/ai_http.go
j0sh: That looks good to me, yep

server/ai_live_video.go
j0sh: That seems fine to me, yes. 1s cadence is probably too frequent though - segments are usually longer than that, eg 2-3s keyframe interval.

Now that I am connecting the bits here mentally while reviewing this ... returning a reader via the API is an interesting choice because it does force us to tee the input reader and handle the returned reader on the API side even if there is nothing else to do with it, eg on the orchestrator where we only discard it.

I don't really have a better idea right now for how to do this, other than making the "should skip" function as part of the API and calling `process` only periodically, doing tee outside the API as necessary. Kind of boilerplate-y though.

j0sh (approved):
Looks pretty good to me - I think I see where some of the oddities may have been coming from re: pipes being closed and wrote some comments on those.

The live payment processor API is also a little interesting; having to internally `tee` and return another reader feels a little suboptimal (pipes are pretty fraught) but I don't have a better idea at the moment that doesn't involve making things even more boilerplate-y.

server/live_payment_processor.go
leszko: It does not help 🤷 But I think that buffering here should be ok.

j0sh (approved):
:shipit:

thomshutt (approved):

PR #3263, opened 2024-11-20
ai: Handle live-video-to-video requests and start runner

victorges:

This change updates the `/live-video-to-video` handler on the orchestrator to send trickle stream urls to the ai-runner to begin live video AI inference when a live video-to-video AI job is processed

- Updates the handler method `StartLiveVideoToVideo()` to send a request to the ai-runner to begin AI inference
- Uses `orch.CheckAICapacity()` to ensure ai-runner is ready for the request
- Only creates trickle channels once job accepted by the ai-runner

Sending requests to gateway with active test stream:
```
curl --location 'http://192.168.10.61:8937/live/video-to-video/test/start' \
--form 'stream="test"' \
--form 'source_id="testout"' \
--form 'source_type="webrtcSession"' \
--form 'query="pipeline=streamdiffusion"'
```

**Checklist:**


server/ai_http.go
```diff
@@ -140,9 +138,86 @@ func (h *lphttp) StartLiveVideoToVideo() http.Handler {
...
+			pu, err := url.Parse(urlPath)
+			if err != nil {
+				return nil, err
+			}
+			if pu.Hostname() != "" {
+				// url has a hostname already so use it
+				return pu, nil
+			}
+			// no hostname, so append one
+			_, port, err := net.SplitHostPort(r.Host)
```
j0sh: this looks sus - the request `r` comes from the gateway. This `appendHostname` dance probably needs to be done on the worker before sending the request down to the runner

server/ai_process.go
```diff
@@ -1043,10 +1044,30 @@ func submitLiveVideoToVideo(ctx context.Context, params aiRequestParams, sess *A
...
+		controlStr := control.String()
+
 		clog.V(common.VERBOSE).Infof(ctx, "pub %s sub %s control %s", pub, sub, control)
 		startTricklePublish(pub, params)
 		startTrickleSubscribe(sub, params)
 		startControlPublish(control, params)
+
+		//Switch subscribe and publish urls
+		req.SubscribeUrl = pub.String() // publish url
+		req.PublishUrl = sub.String() // subscriber url
```
victorges: Are these purposefully inverted? Sub = pub and Pub = sub. Maybe it's necessary since the runner args have a different "perspective"

victorges (commented):
LGTM!

server/ai_http.go
eliteprox: Totally agree, I've moved this to `ai_process.go` where it uses `sess.transcoder()`. I will do tests to ensure this works with all different IP types

server/ai_process.go
eliteprox: Correct, the initial values have `-out` as the subscriber url on the orchestrator so it is necessary to swap these so the ai-runner publishes to `-out`

server/ai_process.go
```diff
@@ -1043,10 +1044,30 @@ func submitLiveVideoToVideo(ctx context.Context, params aiRequestParams, sess *A
...
+		req.ControlUrl = &controlStr
+		
+		// Send another request to the orchestrator to send the video to the live pipeline
```
j0sh: I am a little confused why this is needed, can you describe what is happening at a high level maybe in the PR description?

server/ai_http.go
```diff
@@ -86,36 +89,66 @@ func aiHttpHandle[I any](h *lphttp, decoderFunc func(*I, *http.Request) error) h
...
-			})
-		if err != nil {
-			respondWithError(w, err.Error(), http.StatusInternalServerError)
+		var req worker.GenLiveVideoToVideoJSONRequestBody
+		if err := jsonDecoder[worker.GenLiveVideoToVideoJSONRequestBody](&req, r); err != nil {
+			respondWithError(w, err.Error(), http.StatusBadRequest)
 			return
 		}
 
+		// `startStream` is sent on the 2nd request from to the gateway to hand-off the running stream to the ai-runner
```
j0sh: I am still not sure why a second request is necessary; the O already has all the parameters  and the G is just sending those back. Why not  submit those params to the worker as part of processing the first request?

server/ai_process.go
victorges: @eliteprox added on PR desc now!

server/ai_http.go
j0sh: If you do not want to modify the worker to return params back  to the G (in order to use `handleAIRequest`) then you can copy out the relevant parts from `handleAIRequest` and invoke the required calls directly; AFAICS you're left with only a capacity check and `orch.LiveVideoToVideo`

Most of `handleAIRequest` is boilerplate that can be omitted for now (specialized pipeline handling, payments, metrics)

eliteprox: Makes sense, I will modify the flow to exclude `handleAIRequest` and try to do this in a single request. It first seemed there was a context issue with starting the trickle stream before the end of the http request which led to the handoff to fail, but I think `handleAIRequest` was just interfering with the response

eliteprox: @j0sh I've updated `StartLiveVideoToVideo` to include the relevant parts from `handleAIRequest`

server/ai_http.go
```diff
@@ -96,26 +97,95 @@ func (h *lphttp) StartLiveVideoToVideo() http.Handler {
...
+		workerReq := worker.LiveVideoToVideoParams{
+			ModelId:      req.ModelId,
+			PublishUrl:   req.SubscribeUrl, // SubscribeUrl is the publish url for the worker
+			SubscribeUrl: req.PublishUrl,   // PublishUrl is the subscribe url for the worker
+			ControlUrl:   req.ControlUrl,
+			Params:       req.Params,
+		}
+
+		// append params to the request
+		queryParams := r.FormValue("query")
```
j0sh: is this correct? don't the params come in through the request json?

server/ai_http.go
```diff
@@ -96,26 +97,95 @@ func (h *lphttp) StartLiveVideoToVideo() http.Handler {
...
+		if params != "" {
+			if err := json.Unmarshal([]byte(params), &paramsMap); err != nil {
+				respondWithError(w, err.Error(), http.StatusBadRequest)
+				return
+			}
+			workerReq.Params = &paramsMap
+		}
+
+		// Send request to the worker
+		_, err = orch.LiveVideoToVideo(ctx, requestID, workerReq)
```
j0sh: there is a race condition here if the worker starts subscribing before the channel has been created. the channel needs to be created first. (and then if any of the following requests fail, then the channels need to be closed.)

server/ai_http.go
```diff
@@ -96,26 +97,95 @@ func (h *lphttp) StartLiveVideoToVideo() http.Handler {
...
+		}
+
+		orch := h.orchestrator
+		pipeline := "live-video-to-video"
+		cap := core.Capability_LiveVideoToVideo
+		modelID := *req.ModelId
+		clog.V(common.VERBOSE).Infof(ctx, "Received request id=%v cap=%v modelID=%v", requestID, cap, modelID)
+
+		// Create storage for the request (for AI Workers, must run before CheckAICapacity)
+		err = orch.CreateStorageForRequest(requestID)
```
j0sh: what is this used for in this pipeline? if its not used, lets remove it.

server/ai_http.go
```diff
@@ -96,26 +97,95 @@ func (h *lphttp) StartLiveVideoToVideo() http.Handler {
...
+		}
+
+		// Check if there is capacity for the request
+		if !orch.CheckAICapacity(pipeline, modelID) {
+			respondWithError(w, fmt.Sprintf("Insufficient capacity for pipeline=%v modelID=%v", pipeline, modelID), http.StatusServiceUnavailable)
+			return
+		}
+
+		// Start trickle server for live-video
+		mid := string(core.RandomManifestID())
```
j0sh: don't we already have a `requestID` that is created in the exact same way - can that be reused?

server/ai_http.go
```diff
@@ -96,26 +97,95 @@ func (h *lphttp) StartLiveVideoToVideo() http.Handler {
...
+		pubUrl := TrickleHTTPPath + mid
```
j0sh: nit, but the previous code had a neater style with `var ( ... )`, preference for keeping that as-is.

in general - when doing a self review, look at the diff and try to minimize unnecessary changes

server/ai_http.go
```diff
@@ -96,26 +97,95 @@ func (h *lphttp) StartLiveVideoToVideo() http.Handler {
...
+		subUrl := pubUrl + "-out"
+
+		//Append orchestrator service url host to the urls
+		host := h.node.GetServiceURI().String()
+		publishUrl, _ := common.AppendHostname(pubUrl, host)
```
j0sh: preference for handling errors instead of ignoring them.

Also, preference for keeping the old behavior of using a relative URL, not appending a host here. Keep that on the G and add on the W instead. I am not sure exactly how much we should rely on orchestrator-local Service URI as opposed to a concrete address that the G/W is already interacting with. for example, W might be local / nearby and we don't want to reach outside the network for that.

server/ai_http.go
```diff
@@ -96,26 +97,95 @@ func (h *lphttp) StartLiveVideoToVideo() http.Handler {
...
+		req.PublishUrl = publishUrl.String()
+
+		subscribeUrl, _ := common.AppendHostname(subUrl, host)
+		req.SubscribeUrl = subscribeUrl.String()
+
+		controlUrl, _ := common.AppendHostname(pubUrl+"-control", host)
+		controlUrlStr := controlUrl.String()
+		req.ControlUrl = &controlUrlStr
```
j0sh: this is pretty annoying - can we make the `ControlUrl` an optional field that is not [treated as a pointer](https://github.com/livepeer/ai-worker/blob/ca5d650f9ebb60ee4ff8525b4076d19ba93fd169/worker/runner.gen.go#L248-L254)? Eg, an empty / default value would be just fine. Preference for fixing that in the AI worker if possible.

Same goes for the ModelId and Params fields in the LiveVideoToVideoParams

server/ai_process.go
```diff
@@ -1013,41 +1013,25 @@ func submitLiveVideoToVideo(ctx context.Context, params aiRequestParams, sess *A
...
-			if pu.Hostname() != "" {
-				// url has a hostname already so use it
-				return pu, nil
-			}
-			// no hostname, so append one
-			u := sess.Transcoder() + urlPath
-			return url.Parse(u)
-		}
-		pub, err := appendHostname(resp.JSON200.PublishUrl)
+		pub, err := url.Parse(resp.JSON200.PublishUrl)
```
j0sh: Should preserve the old behavior here of appending the hostname if necessary, because I am not sure if orchestrators are guaranteed to have a ServiceURI set

server/ai_process.go
```diff
@@ -1013,41 +1013,25 @@ func submitLiveVideoToVideo(ctx context.Context, params aiRequestParams, sess *A
...
 		if err != nil {
-			return nil, fmt.Errorf("pub url - %w", err)
+			return nil, fmt.Errorf("invalid publish URL: %w", err)
 		}
-		sub, err := appendHostname(resp.JSON200.SubscribeUrl)
+		sub, err := url.Parse(resp.JSON200.SubscribeUrl)
```
j0sh: as mentioned earlier - preference for not changing this. this will still work even if we have a hostname.

server/ai_http.go
```diff
@@ -96,26 +97,95 @@ func (h *lphttp) StartLiveVideoToVideo() http.Handler {
...
-			subUrl     = pubUrl + "-out"
-			controlUrl = pubUrl + "-control"
-		)
-		jsonData, err := json.Marshal(
-			&worker.LiveVideoToVideoResponse{
-				PublishUrl:   pubUrl,
-				SubscribeUrl: subUrl,
-				ControlUrl:   controlUrl,
-			})
+		payment, err := getPayment(r.Header.Get(paymentHeader))
```
j0sh: we aren't dealing with payments at this point so lets omit this and @leszko can handle that in https://github.com/livepeer/go-livepeer/pull/3261

server/ai_process.go
j0sh: correcting myself here - looks like it _is_ set but still not sure we should rely on it instead of a relative path from  [this comment](https://github.com/livepeer/go-livepeer/pull/3263#discussion_r1853282905)

server/ai_http.go
eliteprox: This is being used for logging the eth address of the gateway downstream, is there another way we can get it?

server/ai_http.go
eliteprox: Reordered to start in the correct order and cleanup publishers channels if job request fails

server/ai_process.go
eliteprox: Moved the hostname logic so that it uses the original value from `sess.Transcoder()` on the gateway, accurately identifying the remote IP of the orchestrator. It's appended on the orchestrator side of the request processing before being sent to the ai-runner

server/ai_process.go
eliteprox: Agree, I reverted back to `sess.Transcoder()` from the gateway, see above comment

server/ai_http.go
eliteprox: This is required for the AI Worker downstream to process the request (I tested without it), It is boilerplate on all the pipelines and might need some refactoring later to make it optional since we are not expecting a response back necessarily.

server/ai_http.go
eliteprox: I placed this back to how you had it, apologies for that

server/ai_http.go
eliteprox: Added error handling for url parsing

For the trickle host, I reverted back to `sess.Transcoder()` to get the remote address accurately and then apply it in the ai-runner request. It passes the host through PublishUrl for now

server/ai_http.go
eliteprox: Re-used it

server/ai_http.go
eliteprox: Used an empty ETH address for now

leszko: yes, I'll handle this in #3261

server/ai_http.go
```diff
@@ -139,8 +155,85 @@ func (h *lphttp) StartLiveVideoToVideo() http.Handler {
 			}
 		}()
 
-		// TODO subscribe to the subscribeUrl for output monitoring
+		// TODO: Use params temporarily to pass the host from gateway
+		host := req.PublishUrl
```
leszko: Will the `host` be the same for both runner and gateway?

I think the setup will be like the runner is running inside docker and orchestrator is running directly on the host. Then the publish URL should contain `host.docker.internal`? Or will they both run in the host network?

core/ai_worker.go
```diff
@@ -549,6 +549,39 @@ func (orch *orchestrator) TextToImage(ctx context.Context, requestID string, req
 	return res.Results, nil
 }
 
+func (orch *orchestrator) LiveVideoToVideo(ctx context.Context, requestID string, req worker.GenLiveVideoToVideoJSONRequestBody) (interface{}, error) {
```
leszko, line 552: I undestand that this is the implementation for the Remote Worker? Have you tested if it works in that setup?

server/ai_http.go
```diff
@@ -139,8 +155,85 @@ func (h *lphttp) StartLiveVideoToVideo() http.Handler {
...
+		control, err := common.AppendHostname(controlUrl+"-control", host)
+		if err != nil {
+			respondWithError(w, fmt.Errorf("invalid control URL: %w", err).Error(), http.StatusBadRequest)
+			return
+		}
+
+		// Prepare request to worker
+		workerReq := worker.LiveVideoToVideoParams{
+			ModelId:      req.ModelId,
+			PublishUrl:   sub, // SubscribeUrl is the publish url for the worker
```
leszko: `// SubscribeUrl is the publish url for the worker` => I'm not sure I understand this comment. Could you rephrase it?

server/ai_http.go
leszko: I also don't understand, do we pass all `query` and `params` from Gateway into Orchestrator?

server/ai_http.go
```diff
@@ -139,8 +155,85 @@ func (h *lphttp) StartLiveVideoToVideo() http.Handler {
...
+		}
+
+		// Prepare the response
+		resp := &worker.LiveVideoToVideoResponse{
+			PublishUrl:   workerReq.PublishUrl,
+			SubscribeUrl: workerReq.SubscribeUrl,
+		}
+
+		// TODO: Fix the nullable type in the worker response
+		if workerReq.ControlUrl != nil {
```
leszko: What is this? Why ControlUrl is not passed from the worker the same as in the case of `PublishUrl`?

leszko (commented):
Added some comments. The main thing is that it didn't work for me, because the address of trickle server is outside Docker and then runner (running inside docker) cannot access the trickle server. Could you please comment on this @victorges @eliteprox ? 

Maybe we need some way of specifying out to access Orchestrator from runner?

server/ai_http.go
eliteprox: ControlUrl was nullable on the ai-runner, I have solved this in https://github.com/livepeer/ai-worker/pull/286

server/ai_http.go
eliteprox: I'm having trouble getting this to work correctly. This is why I initially used the ServiceURI of the orchestrator, otherwise we might need to set a startup flag

j0sh: Referring back to [this comment](https://github.com/livepeer/go-livepeer/pull/3263#discussion_r1853282905) - this  is why we should stay close to the original behavior of this code and keep the paths generated by the O relative. At the G, append the host name from sess.Transcoder. At the worker, just before sending down the request to the runner (or within the runner), append host.docker.internal as @leszko mentioned (or the O address in the case of remote workers; presumably the worker already has that info in order to connect to the O).

core/ai_worker.go
eliteprox: I have not yet tested if it works with the remote ai-worker configuration

server/ai_http.go
eliteprox: I've fixed this now so that it uses the `Params` field from ai_mediaserver.go https://github.com/livepeer/go-livepeer/blob/527bd90fd0e053f9c49741fe124ff3f58c11fca6/server/ai_mediaserver.go#L489-L493

https://github.com/livepeer/go-livepeer/blob/527bd90fd0e053f9c49741fe124ff3f58c11fca6/server/ai_http.go#L163C1-L170C4

server/ai_http.go
eliteprox: I have updated it to not optional in https://github.com/livepeer/ai-worker/pull/286. I'll re-visit this to address the pointer, gen_openapi.py is requiring a pointer even when default is used

server/ai_http.go
eliteprox: I have updated this logic from your suggestion and it tested well. On the orchestrator, it will send only the port number to the ai-runner, which uses the client IP from the request to determine the trickle host. https://github.com/livepeer/ai-worker/pull/286/commits/2661dfa31d91caaf0caa900e7b08782d5e9afedf

Passing the trickle port using a temporary param value https://github.com/livepeer/go-livepeer/pull/3263/files#diff-2d409e07b090aec3653ceddfa542b8809eea772f3f24920e8c285d1e22edd15bR157-R161

This tested well across docker containers and with local/remote IPs.

server/ai_http.go
eliteprox: Updated comment to `Subscribe and Publish urls are swapped for the worker`

leszko (commented):
I'm trying to test this PR, but got the following issue, @eliteprox did you maybe happen to encounter it?

```
E1125 15:17:31.015799       1 ai_worker.go:559] clientIP=127.0.0.1 Error processing with local ai worker err="Error response from daemon: invalid mount config for type \"bind\": bind source path does not exist: /models"
2024/11/25 15:17:31 INFO Deleted stream streamName=6392c04a
2024/11/25 15:17:31 INFO Deleted stream streamName=6392c04a-out
2024/11/25 15:17:31 INFO Deleted stream streamName=6392c04a-control
E1125 15:17:31.016013       1 handlers.go:1596] HTTP Response Error statusCode=500 err=Error response from daemon: invalid mount config for type "bind": bind source path does not exist: /models
```

My setup is mostly what we're running in our environment, so dockerized go-livepeer with host network.

eliteprox:
> I'm trying to test this PR, but got the following issue, @eliteprox did you maybe happen to encounter it?
> 
> ```
> E1125 15:17:31.015799       1 ai_worker.go:559] clientIP=127.0.0.1 Error processing with local ai worker err="Error response from daemon: invalid mount config for type \"bind\": bind source path does not exist: /models"
> 2024/11/25 15:17:31 INFO Deleted stream streamName=6392c04a
> 2024/11/25 15:17:31 INFO Deleted stream streamName=6392c04a-out
> 2024/11/25 15:17:31 INFO Deleted stream streamName=6392c04a-control
> E1125 15:17:31.016013       1 handlers.go:1596] HTTP Response Error statusCode=500 err=Error response from daemon: invalid mount config for type "bind": bind source path does not exist: /models
> ```
> 
> My setup is mostly what we're running in our environment, so dockerized go-livepeer with host network.

If you're running go-livepeer in docker, make sure the volume config is set for `/models`. 
` -v ~/.lpData/:/root/.lpData/`

Also make sure the `-aiModelsDir` is set, otherwise it will default to `/models`     
`-aiModelsDir ~/.lpData/models`

The internal volume mount path should match up with the directory used by `-aiModelsDir`

victorges (commented):
LGTM (most of the code is not mine ok :P)

server/ai_http.go
eliteprox: We've reverted this back to an optional field

PR #3264, opened 2024-11-21
ai/live: Configure parameters from auth callback

j0sh:
Adds a JSON schema for the auth callback, wires it up to the rest of the inference params and incorporates a couple more of the query strings. Take note of the field naming - `pipeline` and `pipeline_params` , happy to adjust those if preferred - eg, to match the existing `model_id` fields

Also some minor refactor to move the AI fields for `aiRequestParams` into its own struct to keep things tidy since I think we will be adding a bunch more things there later.

### DIFF
--- a/server/ai_live_video.go
+++ b/server/ai_live_video.go
@@ -20,7 +20,7 @@ func startTricklePublish(url *url.URL, params aiRequestParams) {
 	if err != nil {
 		slog.Info("error publishing trickle", "err", err)
 	}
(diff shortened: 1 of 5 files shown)


eliteprox (approved):
LGTM! Tested with local authwebhook values and works great

j0sh:
MacOS builds are broken because of some [brew thing](https://github.com/Homebrew/brew/issues/18788), merging for now to unblock https://github.com/livepeer/go-livepeer/pull/3263

PR #3265, opened 2024-11-21
Push Kafka initialisation to later in startup

thomshutt:
We were doing it before all the config was set up, which meant that the Gateway value wasn't populated. Took the opportunity to split out the code too, to try and reduce the craziness of starter.go

### DIFF
--- a/cmd/livepeer/starter/kafka.go
+++ b/cmd/livepeer/starter/kafka.go
@@ -8,6 +8,7 @@ import (
 func startKafkaProducer(cfg LivepeerConfig) error {
 	if *cfg.KafkaBootstrapServers == "" || *cfg.KafkaUsername == "" || *cfg.KafkaPassword == "" || *cfg.KafkaGatewayTopic == "" {
 		glog.Warning("not starting Kafka producer as producer config values aren't present")
+		return nil
 	}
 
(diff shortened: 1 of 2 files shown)


cmd/livepeer/starter/kafka.go
```diff
@@ -0,0 +1,25 @@
+package starter
+
+import (
+	"github.com/golang/glog"
+	lpmon "github.com/livepeer/go-livepeer/monitor"
+)
+
+func startKafkaProducer(cfg LivepeerConfig) error {
+	if *cfg.KafkaBootstrapServers == "" || *cfg.KafkaUsername == "" || *cfg.KafkaPassword == "" || *cfg.KafkaGatewayTopic == "" {
+		glog.Warning("not starting Kafka producer as producer config values aren't present")
```
mjh1, line 10: do you need to return here?

thomshutt: No, starting without pushing to Kafka is fine if it's not configured

mjh1 (approved):

cmd/livepeer/starter/kafka.go
thomshutt: Good catch, we do!

PR #3266, opened 2024-11-22
workflows: Rename package to `pkgconf` for macos/brew install

hjpotter92:
https://github.com/actions/runner-images/issues/10984

most likely to be resolved within a week, considering the scale of affected repos/workflows. but we still need a bypass/fix to unblock our build pipelines


j0sh (approved):
🤞

PR #3271, opened 2024-11-22
Introduce new config param for AI webhook auth

mjh1:


server/auth.go
```diff
@@ -133,7 +136,15 @@ func authenticateAIStream(authURL *url.URL, req AIAuthRequest) (*AIAuthResponse,
...
 	}
 
-	resp, err := http.Post(authURL.String(), "application/json", bytes.NewBuffer(jsonValue))
+	request, err := http.NewRequest("POST", authURL.String(), bytes.NewBuffer(jsonValue))
+	if err != nil {
+		return nil, err
+	}
+
+	request.Header.Set("Content-Type", "application/json")
+	request.Header.Set("x-api-key", os.Getenv("SHOWCASE_API_KEY"))
```
j0sh, line 145: If we need random headers then it really should be specified via CLI args rather than hard-coded here

mjh1: Yeah sorry, it was a quick hack, will fix it now.

PR #3275, opened 2024-11-25
feat: add proposal vote cli arg

rickstaa:

This pull request gives orchestrators the ability to vote on active treasury proposal directly from the go-livepeer CLI.

- Adds a new CLI option people can use to vote on active proposals.
- Adds client bindings to interact with the [LivepeerGovernor](https://arbiscan.io/address/0xcFE4E2879B786C3aa075813F0E364bb5acCb6aa0) contract.
- Adds a new `voteOnProposal` handler on the server.
- Adds new tests.
- Updates outdated contracts.
    - @victorges, @leszko these other contracts were changed when I ran `go generate client.go`. Can you check if these really need to be updated or that my setup is incorrect.

I wrote tests and tested the command line argument with a non-orchestrator wallet since we currently don't have a testnet. @victorges, @leszko if you reviewed this I can test with my actual wallet.

**Checklist:**


eth/client.go
```diff
@@ -325,6 +330,23 @@ func (c *client) setContracts(opts *bind.TransactOpts) error {
 
 	glog.V(common.SHORT).Infof("LivepeerTokenFaucet: %v", c.faucetAddr.Hex())
 
+	livepeerGovernorAddr, err := c.GetContract(crypto.Keccak256Hash([]byte("LivepeerGovernor")))
```
leszko, line 333: This will break the local on-chain setup with the local chain `livepeer/geth-with-livepeer-protocol:confluence`. It's not the end of the world, but a regression in the dev process. 

Two options here:
1. Quick => Make this resolution here optional (so don't fail if it's not resolved)
2. Longer => Update `livepeer/geth-with-livepeer-protocol:confluence` Docker image to include the Governor contract

I think that actually Point 1 may be a better option because we don't risk introducing any regression anywhere.

leszko (commented):
Added one comment. Other than that looks good. 

If we don't have testnet, then my suggestion is if @rickstaa, you try it with some real proposal with your local Orchestrator and then merge.

eth/client.go
victorges: @leszko Do you know how that docker image is created? When we implemented the treasury I believe I updated the bootstrap scripts from `livepeer/protocol` to deploy the governor as well. If the docker image is created by simply running that script on a fresh geth, maybe it will just work ®️

cmd/livepeer_cli/livepeer_cli.go
```diff
@@ -113,6 +113,7 @@ func (w *wizard) initializeOptions() []wizardOpt {
 		{desc: "Sign a message", invoke: w.signMessage},
 		{desc: "Sign typed data", invoke: w.signTypedData},
 		{desc: "Vote in a poll", invoke: w.vote, orchestrator: true},
+		{desc: "Vote on a proposal", invoke: w.voteOnProposal, orchestrator: true},
```
victorges: **doc nit:** WDYT of being more specific to avoid confusion? 
```suggestion
		{desc: "Vote in a governance poll", invoke: w.vote, orchestrator: true},
		{desc: "Vote on a treasury proposal", invoke: w.voteOnProposal, orchestrator: true},
```

cmd/livepeer_cli/wizard_transcoder.go
```diff
@@ -278,6 +278,64 @@ func (w *wizard) vote() {
...
+				break
+			}
+			fmt.Println("Must enter a valid ID")
+		}
+
+		fmt.Printf("Are you sure you want to vote \"%v\"? (y/n) -", choice.String())
+		confirm = w.readStringYesOrNo()
+	}
+
+	fmt.Printf("Do you want to provide a reason for your vote? (y/n) -")
```
victorges, line 316: Would be awesome to display these in the dashboard at some point!

eth/client.go
victorges: Agreed that making this optional would be an acceptable solution tho. Feel free to just do that to unblock this PR @rickstaa!

victorges (approved):
Neat!

cmd/livepeer_cli/wizard_transcoder.go
stronk-dev: Mike will be adding it to his [vote tracker](https://tools.livepeer.cloud/vote/history). At the moment it is just me and Titan setting the reason, but with this PR we might see a lot more Orchestrators add a reason to their vote.

eth/client.go
rickstaa: @leszko, @victorges I tried to reproduce the issue but couldn't.  

In Solidity, a mapping is [defined as](https://docs.soliditylang.org/en/v0.8.9/types.html#mapping-types):  

```bash
You can think of mappings as [hash tables](https://en.wikipedia.org/wiki/Hash_table), which are virtually initialized such that every possible key exists and is mapped to a value whose byte representation is all zeros, a type’s [default value](https://docs.soliditylang.org/en/v0.8.9/control-structures.html#default-value). The similarity ends there—the key data is not stored in a mapping; only its keccak256 hash is used to look up the value.
```

This means that if a key doesn't exist, the **default address value (0x0000000000000000000000000000000000000000)** is returned. You can see this in action in the https://github.com/livepeer/protocol/blob/e8b6243c48d9db33852310d2aefedd5b1c77b8b6/contracts/Controller.sol#L27.

The issue would to my knowledge only arise when calling the contract through the client, rather than from within Solidity itself.  When I start and create a devnet I will get:

```bash
I0213 10:13:00.597277 1932855 client.go:350] LivepeerGovernor: 0x0000000000000000000000000000000000000000
```

The error is only thrown when the contract parameter is not a Keccak256 hash.

Let me know if I'm missing something!

leszko: @rickstaa what's the exact issue you encounter?

I'd probably just add this IF (make it optional) and merge this PR, no need to dig deeper into this. Unless, you think it can break something 🙃

rickstaa: @leszko I tested the following on-chain while casting my vote:  
- **Vote with reason (abstain):** [[Transaction](https://arbiscan.io/tx/0x8fee65a00de8d11aea50b9a757f6511a578654649ed18a952f7b704fc019f02a)](https://arbiscan.io/tx/0x8fee65a00de8d11aea50b9a757f6511a578654649ed18a952f7b704fc019f02a)  
- **Vote with reason (for):** [[Transaction](https://arbiscan.io/tx/0xad0d61a65e8380486a5d9bd9cec99f9c82e5e5502b8dc0b24f9504eb586890bb)](https://arbiscan.io/tx/0xad0d61a65e8380486a5d9bd9cec99f9c82e5e5502b8dc0b24f9504eb586890bb)  
I can run additional tests on devnet, but I need time to resolve a deployment issue and won’t be able to check before next week.  Since the other options only modify the calling function and support value—both of which we’ve already verified against the protocol code—I believe we’re good to go.  

As a final check, I can vote **yes** on [this proposal](https://explorer.livepeer.org/treasury/39549981365616941647238744664851111817074961843443607299795325625311288410936) once it goes live.

leszko (approved):

leszko:
LGTM @rickstaa feel free to merge.

rickstaa:
> LGTM @rickstaa feel free to merge.

@leszko thanks. I now also tested the code without a reason.

- **Vote yes without reason**: https://arbiscan.io/tx/0x881434d1bbfdeb1c6d0eb0ceebba9cd9ddb3ef02904ea7e86b386a61ab32150d

PR #3276, opened 2024-11-25
Use system ffmpeg for streaming trickle into RTMP

victorges:


leszko (approved):
I think that if LPMS does not work, we could just merge this one. Asked @rickstaa to test it and it works fine with his Transcoding Orchestrator.

thomshutt (approved):

leszko:
This is how quickly you get from a super temp hack into the prod code 🙃

PR #3281, opened 2024-11-27
Add -liveAITrickleHostForRunner flag

leszko:
Runner can be deployed inside Docker container (usual setup), in a different network, or on a separate machine. This flag adds support for configuring the trickle host for the runner.


cmd/livepeer/livepeer.go
```diff
@@ -164,6 +164,7 @@ func parseLivepeerConfig() starter.LivepeerConfig {
 
 	// Live AI Media Server:
 	cfg.MediaMTXApiPassword = flag.String("mediaMTXApiPassword", "", "HTTP basic auth password for MediaMTX API requests")
+	cfg.LiveAITrickleHostForRunner = flag.String("liveAITrickleHostForRunner", "", "Trickle Host used by AI Runner; It's used to overwrite the publicly available Trickle Host")
```
pwilczynskiclearcode, line 167: If docker is a standard configuration maybe we should place `172.17.0.1:8888` as a default in here?

leszko: I think there are many different setups, so I think the default should be what you expose as a public IP. I believe that if we make `172.17.0.1:8888` default, then it becomes more confusing.

thomshutt (approved):

PR #3283, opened 2024-11-28
Dockerfile.mediamtx remove custom mediamtx.yml config

pwilczynskiclearcode:


hjpotter92 (approved):

docker/Dockerfile.mediamtx
```diff
@@ -14,6 +14,4 @@ RUN curl -L https://github.com/bluenviron/mediamtx/releases/download/${MEDIAMTX_
     && tar xzvf /mediamtx.tar.gz \
```
pwilczynskiclearcode, line 14: `mediamtx.tar.gz` already upacks the config file:
```
$ tar xzvf mediamtx.tar.gz
x mediamtx
x mediamtx.yml
x LICENSE
```

mjh1 (approved):

mjh1:
Thanks @pwilczynskiclearcode we were going to keep this config file around for local development since it has our changes for runOnReady etc but since we do this with env vars for the real environment we can just do it that way for local dev too.
cc @j0sh

PR #3285, opened 2024-11-29
mediamtx: Publish deployed version if runtime args are present

hjpotter92:
preparing the image so that container can optionally inject version metrics to prometheus/victoriametrics at runtime.

### DIFF
--- a/docker/Dockerfile.mediamtx
+++ b/docker/Dockerfile.mediamtx
@@ -1,9 +1,5 @@
 FROM	ubuntu:24.04
 
-ENV	MEDIAMTX_VERSION="1.9.3"
-
-ADD	"https://github.com/bluenviron/mediamtx/releases/download/v${MEDIAMTX_VERSION}/mediamtx_v${MEDIAMTX_VERSION}_linux_amd64.tar.gz"	/opt/mediamtx.tar.gz
-
 # we need curl in the image as it's later used in the runOnReady command
 RUN	apt-get update \
 	&& apt-get install -y \
@@ -11,6 +7,10 @@ RUN	apt-get update \
 		curl \
 	&& apt-get clean \
 	&& rm -rf /var/lib/apt/lists/*
+
+ENV	MEDIAMTX_VERSION="1.9.3"
+
+ADD	"https://github.com/bluenviron/mediamtx/releases/download/v${MEDIAMTX_VERSION}/mediamtx_v${MEDIAMTX_VERSION}_linux_amd64.tar.gz"	/opt/mediamtx.tar.gz
 
 RUN	tar xzf /opt/mediamtx.tar.gz -C /opt/ \
 	&& mkdir -p /usr/local/bin /etc/mediamtx/ \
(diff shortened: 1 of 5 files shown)


docker/mediamtx-entry.bash
```diff
@@ -0,0 +1,19 @@
...
+  fi
+  echo <<EOF |
+# HELP version Current software version as a tag, always 1
+# TYPE version gauge
+version{app="MediaMTX",version="$MEDIAMTX_VERSION"} 1
+EOF
+    curl -X POST --data-binary @- "$LP_PUBLISH_MEDIAMTX_METRICS_ENDPOINT"
+fi
+
+/usr/local/bin/mediamtx
```
pwilczynskiclearcode: It could be:
```bash
exec ${1:-/usr/local/bin/mediamtx}
```

so it would work with `ENTRYPOINT` and `CMD` correctly. And `exec` to get rid of this intermediate bash script running forever.

docker/mediamtx-entry.bash
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+[ -v DEBUG ] && set -x
+
+set -euo pipefail
+
+if [ -v LP_PUBLISH_MEDIAMTX_METRICS ]; then
+  if [ -z "$LP_PUBLISH_MEDIAMTX_METRICS_ENDPOINT" ]; then
```
pwilczynskiclearcode, line 8: Should we exit if the endpoint is not specified?
```bash
if [ -z "$LP_PUBLISH_MEDIAMTX_METRICS_ENDPOINT" ]; then
    echo >&2 "No endpoint specified for publishing mediamtx metrics."
    exit 1
fi
```

docker/Dockerfile.mediamtx
```diff
@@ -1,17 +1,23 @@
-FROM ubuntu:24.04
+FROM	ubuntu:24.04
 
-ENV MEDIAMTX_VERSION="v1.9.3"
+ENV	MEDIAMTX_VERSION="1.9.3"
+
+ADD	"https://github.com/bluenviron/mediamtx/releases/download/v${MEDIAMTX_VERSION}/mediamtx_v${MEDIAMTX_VERSION}_linux_amd64.tar.gz"	/opt/mediamtx.tar.gz
```
pwilczynskiclearcode: Nice!
I would move it down below `apt-get install` as it may change more often than system packages.
And being lower on the dockerfile would make cached layers being used more optimally

pwilczynskiclearcode (approved):

docker/mediamtx-entry.bash
hjpotter92: i was thinking it more as a warning than a fail condition.

docker/mediamtx-entry.bash
hjpotter92: changed it to `${@:-...` so that complex commands would also work

docker/mediamtx-entry.bash
pwilczynskiclearcode: if `LP_PUBLISH_MEDIAMTX_METRICS_ENDPOINT` is missing then `curl` would fail few lines below.

But yeah warning is fine... I think we'd catch it missing anyway as not having mediamtx version on grafana

PR #3290, opened 2024-12-03
Better retry logic for mediaserver pull and small refactor

mjh1:

In order to have better retry logic we need to check if the stream still exists, if the input stream is still coming in then we should continue retrying on errors, if the user stops streaming in then we can bail out. At the moment I've implemented this for the segmentation call, we can implement it in other places where we need retries like the output stream once we're happy with it.

- Refactor mediamtx api calls into separate package
- Stop retrying segmentation if stream does not exist

**Checklist:**

### DIFF
--- a/media/rtmp2segment.go
+++ b/media/rtmp2segment.go
@@ -35,7 +35,7 @@ func (ms *MediaSegmenter) RunSegmentation(ctx context.Context, in string, segmen
 	wg.Add(1)
 	go func() {
 		defer wg.Done()
-		processSegments(segmentHandler, outFilePattern, completionSignal)
+		processSegments(ctx, segmentHandler, outFilePattern, completionSignal)
 	}()
 
 	retryCount := 0
@@ -161,7 +161,7 @@ func openNonBlockingWithRetry(name string, timeout time.Duration, completed <-ch
 	}
 }
 
-func processSegments(segmentHandler SegmentHandler, outFilePattern string, completionSignal <-chan bool) {
+func processSegments(ctx context.Context, segmentHandler SegmentHandler, outFilePattern string, completionSignal <-chan bool) {
 
 	// things protected by the mutex mu
 	mu := &sync.Mutex{}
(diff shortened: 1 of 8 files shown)


j0sh:
Nice cleanup, thanks. Maybe we could put it into the `media` package instead of introducing another top-level `mediaserver` package? (Which also feels a little overloaded with the existing `server/mediaserver.go` and `server/ai_mediaserver.go` files)

media/rtmp2segment.go
```diff
@@ -18,30 +18,41 @@ import (
...
 	go func() {
 		defer wg.Done()
-		processSegments(segmentHandler, outFilePattern, completionSignal)
+		processSegments(ctx, segmentHandler, outFilePattern, completionSignal)
 	}()
 
 	retryCount := 0
-	for retryCount < 5 {
+	for {
+		streamExists, err := ms.MediaMTXClient.StreamExists(ms.MediaMTXHost, id, sourceType)
```
mjh1, line 45: @j0sh not sure how you feel about this code having knowledge of mediamtx itself, maybe it could just be an interface instead `StreamChecker` or something with the `StreamExists` function.

j0sh: An interface is probably cleaner but I wouldn't spend too much time on it since the intent is to bring media handling in-house anyway and the rest of this code is becoming very livepeer-specific anyway with the logging stuff.

This bit is always going to be somewhat racy, eg it's possible the stream will disconnect between the StreamExists call and the reconnect. Do you know if the LPMS return is different if the stream exists vs not, or if the termination was clean vs not? If so, maybe it's worth just reconnecting on a non-nil error and checking the return rather than doing a separate stream exists check.

At some point we may also want to consider a "push" approach, eg having the runOnReady invoke ffmpeg to stream RTMP into go-livepeer, rather than having go-livepeer pull here. This might be a stepping stone to bringing more of the media handling in-house.

mjh1:
> Nice cleanup, thanks. Maybe we could put it into the `media` package instead of introducing another top-level `mediaserver` package? (Which also feels a little overloaded with the existing `server/mediaserver.go` and `server/ai_mediaserver.go` files)

Ah yep meant to call that out too, yeah I didn't like the mediaserver package name when we already have those files. Putting it in `media` sounds good

media/rtmp2segment.go
mjh1: Unfortunately I tried sending an unsupported codec as well as cancelling the input stream and the error returned from LPMS is just `Input/output error` in both cases so can't distinguish currently :(

thomshutt (approved):

PR #3293, opened 2024-12-03
Re-apply "Live Video Payments"

leszko:
Reapply #3261 

**Most notable changes to the original PR**
- Fixed `priceInfo` to get from capabilities (therefore from the `aiModels.json` config instead of the `pricePerUnit` flag)
- Fixed FPS when probing segment to `30` (because probing sometimes gave `9000` FPSs which broke the payment mechanism)
- Detached to separate goroutines/channels/loops the logic for payment processing vs probing segments (sending trickle segments sometimes took > 40s, which influenced payments)


j0sh (approved):
> Fix FPS to always 30 when probing segment

Maybe if FPS is excessively large (> 150) then we can hard-code to 30?

There is also some risk in having _too high_ of a framerate, which is taxing downstream. We can tackle that later though.

🚢 🇮🇹

PR #3296, opened 2024-12-05
trickle: Sweep idle channels on the server.

j0sh:
Eventually cleans up anything on the trickle server side in case we miss a cleanup somewhere.

This does imply that channels need to be active regularly, so for naturally intermittent channels like control or events we will have to send a keepalive via clients periodically. (keepalives will also be useful for keeping client connections alive.)

Default 5-minute idle timeout, with a 1-minute sweep interval (both configurable). This basically requires us to explicitly start the trickle server (in order to start sweeping), and the `Start()` function returns a cancel-type function in order to stop things, a'la context cancel.

Also adds an explicit "channel create" endpoint to the trickle server (empty POST to `/channel-name` without an associated sequence number). Not used right now but it has come in handy while testing elsewhere, eg in [python](https://github.com/j0sh/http-trickle/blob/bf10c47c3c40b24ebc6c7a60ce6fa2173a2d3b16/python/image.py#L18-L19).

Also includes this [change](https://github.com/livepeer/go-livepeer/pull/3296/commits/7c977cf2d946418f3f07068428ed57ae39497b16) to make the `startAIServer` take a `*lphttp` so we aren't copying the entire struct every time we pass this in, and can make changes to its fields (eg, assigning the trickle server)


leszko (approved):

j0sh:
Holding off on merging this until we have keepalives to ensure control API liveness since messages there are intermittent and we don't want those to be swept up

j0sh:
Once https://github.com/livepeer/go-livepeer/pull/3299 and  https://github.com/livepeer/ai-worker/pull/313 are in, I'll merge this one

mjh1 (approved):

PR #3299, opened 2024-12-05
ai/live: Keepalive for control channel.

j0sh:
The control channel only sees messages sporadically, so send down a keepalive periodically to prevent clients from timing out. This also prevents the server from sweeping idle channels.

Needed for https://github.com/livepeer/go-livepeer/pull/3296

Also see https://github.com/livepeer/ai-worker/pull/313 for the ai-worker counterpart


server/ai_live_video.go
```diff
@@ -138,5 +139,35 @@ func startControlPublish(control *url.URL, params aiRequestParams) {
...
+		ticker.Stop()
+		done <- true
+	}
+
+	params.node.LivePipelines[stream] = &core.LivePipeline{
+		ControlPub:  controlPub,
+		StopControl: stop,
+	}
+
+	// send a keepalive periodically to keep both ends of the connection alive
```
victorges, line 155: Slightly confused by this. I thought trickle didn't hold any long running connections? It just individually POST'ed or GET'ed segments from the server. Or is it that it works with a "long get" in case there are no messages to be received? If that's the case, the problem could be solved on a protocol level too, making the long GET stop with a "try again" after those 10s, so we don't need to reimplement this on every channel that doesn't get messages all the time (e.g. I'm creating the events channel now. it naturally has some recurring events, but it could potentially not have too).

server/ai_live_video.go
```diff
@@ -138,5 +139,35 @@ func startControlPublish(control *url.URL, params aiRequestParams) {
...
+	go func() {
+		for {
+			select {
+			case <-ticker.C:
+				msg := `{"keep":"alive"}`
```
victorges: **nit:** Would consider a static const

server/ai_live_video.go
```diff
@@ -138,5 +139,35 @@ func startControlPublish(control *url.URL, params aiRequestParams) {
 	}
 	params.node.LiveMu.Lock()
 	defer params.node.LiveMu.Unlock()
-	params.node.LivePipelines[stream] = &core.LivePipeline{ControlPub: controlPub}
+
+	ticker := time.NewTicker(10 * time.Second)
+	done := make(chan bool, 1)
+	stop := func() {
```
victorges, line 145: Shouldn't this close the publisher too?

victorges (approved):
LGTM with some questions

server/ai_live_video.go
j0sh: > Or is it that it works with a "long get" in case there are no messages to be received?

Yep that is exactly what it does.

> If that's the case, the problem could be solved on a protocol level too, making the long GET stop with a "try again" after those 10s

This is actually how the golang subscriber does it already. I agree handling it on the protocol level is better but it's also a bit tricky to do correctly, and we would have to re-implement similar logic both in pytrickle and on the publisher side too (to avoid the server from sweeping idle streams).

That behavior will be good to implement across all the clients in the medium term, but for now this is a decent 80/20 solution.

>  I'm creating the events channel now. it naturally has some recurring events, but it could potentially not have too

There might be some other informative telemetry we can send back periodically for events, eg in/out frame rates ... but if not, then yeah a keepalive would help for now.

Golang subscriber code:

https://github.com/livepeer/go-livepeer/blob/0e1c676707c57d1022f4003b4a8aae25ac1d5116/trickle/trickle_subscriber.go#L86-L120

server/ai_live_video.go
j0sh: The publisher is closed just before stop is invoked [here](https://github.com/livepeer/go-livepeer/pull/3299/files/9f4a9793c757f1c4945beb538b48d5772d6a6e9b#diff-88f3e5816c6c29e92773629b388dfc58ed59a3682ab624c7f3448cfda3106475R562) ... we could try to combine them but I am a little hesitant to change the logic outside the timer stuff.

How it works right now is a little lazy:
If _gateway_ closes the publish channel, then it immediately stops the timer 
If _orchestrator_ closes the publish channel, then the timer stops on the next tick when the write errors out (no need to close the publisher at this point since it's already closed)

server/ai_live_video.go
j0sh: const'ified in [533e8a64](https://github.com/livepeer/go-livepeer/pull/3299/commits/533e8a64e8520e42f942dfa6b695675d0aedde13) (kept the declaration local since it's not used anywhere else; presumably the compiler is smart enough to hoist it out of the loop)

leszko (approved):

mjh1 (approved):

PR #3300, opened 2024-12-06
monitoring: add events trickle api

gioelecerati:
A trickle api for worker related events on the live-video-to-video pipeline, to monitor errors and stream_status

- Added events trickle api on O, with the G subscribing and the W publishing G -> O <- W 
- The G receives the status of the stream including errors and real time FPS for both input and output 
- Added a publish on Kafka of these events

dev env

**Checklist:**

### DIFF
--- a/core/livepeernode.go
+++ b/core/livepeernode.go
@@ -162,6 +162,7 @@ type LivepeerNode struct {
 type LivePipeline struct {
 	ControlPub  *trickle.TricklePublisher
 	StopControl func()
+	EventsPub   *trickle.TricklePublisher
 }
 
 // NewLivepeerNode creates a new Livepeer Node. Eth can be nil.
--- a/monitor/kafka.go
+++ b/monitor/kafka.go
@@ -149,23 +149,12 @@ func SendQueueEventAsync(eventType string, data interface{}) {
 	randomID := uuid.New().String()
 	timestampMs := time.Now().UnixMilli()
 
-	// Add debug logging
-	if status, ok := data.(PipelineStatus); ok {
-		glog.Infof("Pipeline Status before sending to Kafka - StreamID: %v, Pipeline: %s",
-			status.StreamID, status.Pipeline)
-	}
-
 	event := GatewayEvent{
 		ID:        stringPtr(randomID),
 		Gateway:   stringPtr(kafkaProducer.gatewayAddress),
 		Type:      &eventType,
 		Timestamp: stringPtr(fmt.Sprint(timestampMs)),
 		Data:      data,
-	}
-
-	// Add debug logging for the full event
-	if eventBytes, err := json.Marshal(event); err == nil {
-		glog.Infof("Full event being sent to Kafka: %s", string(eventBytes))
 	}
 
 	select {
--- a/monitor/kafka.go
+++ b/monitor/kafka.go
@@ -35,6 +35,20 @@ type GatewayEvent struct {
 	Data      interface{} `json:"data"`
 }
 
+type PipelineStatus struct {
+	Pipeline        string   `json:"pipeline"`
+	StartTime       float64  `json:"start_time"`
+	InputFPS        float64  `json:"input_fps"`
+	OutputFPS       float64  `json:"output_fps"`
+	LastInputTime   float64  `json:"last_input_time"`
+	LastOutputTime  float64  `json:"last_output_time"`
+	RestartCount    int      `json:"restart_count"`
+	LastRestartTime float64  `json:"last_restart_time"`
+	LastRestartLogs []string `json:"last_restart_logs"`
+	LastError       *string  `json:"last_error"`
+	StreamID        *string  `json:"stream_id"`
+}
+
 var kafkaProducer *KafkaProducer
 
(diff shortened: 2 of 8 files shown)


core/livepeernode.go
```diff
@@ -162,6 +162,7 @@ type LivepeerNode struct {
 type LivePipeline struct {
 	ControlPub  *trickle.TricklePublisher
 	StopControl func()
+	EventsPub   *trickle.TricklePublisher
```
victorges, line 165: Shouldn't this be a sub? 🤔

monitor/kafka.go
```diff
@@ -35,6 +35,23 @@ type GatewayEvent struct {
 	Data      interface{} `json:"data"`
 }
 
+type PipelineStatus struct {
```
victorges, line 38: Hmm, maybe I should auto-generate this type for you. Let me see if I can figure that out and will let you know, otherwise let's just keep this

monitor/pipeline_status.go
```diff
@@ -0,0 +1,30 @@
...
+
+import (
+	"sync"
+)
+
+var (
+	// pipelineStatusMap stores the latest pipeline status for each stream
+	pipelineStatusMap = make(map[string]PipelineStatus)
+	pipelineStatusMu  sync.RWMutex
+)
```
victorges, line 11: WDYT of creating a struct to hold this "contiguous state"? So it's not N loose static vars, only 1 that holds everything and has the methods to interact with it safely (same functions below, just in the struct)

server/ai_http.go
```diff
@@ -163,6 +164,8 @@ func (h *lphttp) StartLiveVideoToVideo() http.Handler {
 		subCh.CreateChannel()
 		controlPubCh := trickle.NewLocalPublisher(h.trickleSrv, mid+"-control", "application/json")
 		controlPubCh.CreateChannel()
+		eventsCh := trickle.NewLocalPublisher(h.trickleSrv, mid+"-events", "application/json")
```
victorges, line 167: Same, thought this was supposed to be a subscriber. Or is this orchestrator code?

server/ai_http.go
```diff
@@ -210,10 +214,12 @@ func (h *lphttp) StartLiveVideoToVideo() http.Handler {
 
 		// Prepare request to worker
 		controlUrlOverwrite := overwriteHost(h.node.LiveAITrickleHostForRunner, controlUrl)
+		eventsUrlOverwrite := overwriteHost(h.node.LiveAITrickleHostForRunner, eventsUrl)
 		workerReq := worker.LiveVideoToVideoParams{
 			ModelId:      req.ModelId,
 			PublishUrl:   overwriteHost(h.node.LiveAITrickleHostForRunner, subUrl),
 			SubscribeUrl: overwriteHost(h.node.LiveAITrickleHostForRunner, pubUrl),
```
victorges: Unrelated but would you mind creating helper vars for these too? I don't get why they are done differently 🤔 kinda bothers me lol

server/ai_live_video.go
```diff
@@ -170,3 +172,54 @@ func startControlPublish(control *url.URL, params aiRequestParams) {
...
+
+	clog.Infof(ctx, "Starting event subscription for URL: %s", url.String())
+
+	go func() {
+		for {
+			clog.Infof(ctx, "Attempting to read from event subscription for URL: %s", url.String())
+			segment, err := subscriber.Read()
+			if err != nil {
+				clog.Infof(ctx, "Error reading events subscription: %s", err)
+				monitor.DeletePipelineStatus(params.liveParams.stream)
```
victorges: I think we should hold on to the last known status of the stream. Only delete it when the stream is done, which is not "decided" here AFAICT. WDYT?

server/ai_live_video.go
```diff
@@ -170,3 +172,54 @@ func startControlPublish(control *url.URL, params aiRequestParams) {
...
+			body, err := io.ReadAll(segment.Body)
+			segment.Body.Close()
+
+			if err != nil {
+				clog.Infof(ctx, "Error reading events subscription body: %s", err)
+				continue
+			}
+
+			stream := params.liveParams.stream
+
```
victorges, line 203: WDYT of keeping that log of every event received? Could be useful for debugging through loki if there's any issues with kafka. We send 1 event every 10s per stream which seems a reasonable log amount.

I'd also add the log exactly here, before any of the checks below that might skip processing of the event. Add the `stream` as a log field, if it's empty it's empty in the log.

core/livepeernode.go
victorges: Hm I'm guessing this is orchestrator code? Or maybe it's unused? The O should start a server not a publisher :thonking:

victorges (approved):
LGTM

monitor/pipeline_status.go
gioelecerati: Since this in memory store is not used yet, we can do it in a subsequent pr if you agree, since we want to merge this asap for evan

server/ai_live_video.go
gioelecerati: As mentioned in the previous comment, commented out and added a todo

PR #3302, opened 2024-12-06
Improve trickle subscribe error handling

mjh1:

Improving trickle subscribe error handling.

- Change the response handling in `ai_process.go` to replicate what we do for the other pipelines, if there isn't a 200 we should return an error back up, for example when there's a 503 capacity error.
- Add some retries into the trickle subscribe logic and return errors via a callback function. Implements this: https://www.notion.so/livepeer/Live-AI-Pipeline-Integration-Points-1510a348568780379b40f42d7f2f2332?pvs=4#1520a34856878036bc84d23f13b6528a

- Tested shutting down the orch mid stream in local dev environment.
- Tested for regressions etc in staging.

**Checklist:**


server/ai_live_video.go
```diff
@@ -81,14 +82,32 @@ func startTrickleSubscribe(ctx context.Context, url *url.URL, params aiRequestPa
...
+		defer func() { params.liveParams.errorCallback(err) }() // wrapper function to be able to reference the latest err
+		retries := 0
 		for {
-			segment, err := subscriber.Read()
+			if !params.inputStreamExists() {
+				clog.Errorf(ctx, "Stopping trickle subscribe, input stream does not exist.")
+				break
+			}
+			var segment *http.Response
+			segment, err = subscriber.Read()
```
leszko, line 100: Will this work fine if called multiple times after retries? @j0sh

server/ai_live_video.go
```diff
@@ -81,14 +82,32 @@ func startTrickleSubscribe(ctx context.Context, url *url.URL, params aiRequestPa
...
+					return
+				}
 				// TODO if not EOS then signal a new orchestrator is needed
 				clog.Infof(ctx, "Error reading trickle subscription: %s", err)
-				return
+				if retries > 2 {
+					return
+				}
+				retries++
+				time.Sleep(5 * time.Second)
```
leszko: Isn't 5s a lot, would it maybe make sense to have it as `1s` and have more retries? Just raising some points, but up to you and @j0sh 🙃

server/ai_mediaserver.go
```diff
@@ -482,6 +482,14 @@ func (ls *LivepeerServer) StartLiveVideo() http.Handler {
 			ls.cleanupLive(streamName)
 		}()
 
+		errorCallback := func(err error) {
+			clog.Errorf(ctx, "Failed to process live video: %s", err)
```
leszko: 1. What if error is `nil`, do we still want to log an error?
2. So, no matter if error is nil or not, we kick of the connection? if that's the case, then could you add a comment explaining it? It sounds counter-intutivie.

leszko (approved):
Added some comments. Approving, but I'd love @j0sh to review it as well if we're good with the retries on the trickle,

server/ai_live_video.go
j0sh: Depending on where it fails - it might try to request the _next_ segment instead of the current one. For our purposes here that is probably fine. Not ideal, but fine for now.

If you really wanted to fix that, we can probably add a Retry() helper to the subscriber that takes a sequence number (or a http.Response and call [GetSeq](https://github.com/livepeer/go-livepeer/blob/master/trickle/trickle_subscriber.go#L40) on it), then temporarily sets the subscriber `idx` , runs (pre-)connect, then resets `idx`, and returns the response. I am doing something similar in an upcoming change set for the publisher side.

server/ai_mediaserver.go
mjh1: You're right, it can be nil in some cases, I'll improve it

mjh1: I've changed things around so that this should never be called with a nil error but added a nil check anyway. We kick the input connection to allow the client to reconnect and restart the pipeline

server/ai_live_video.go
j0sh: Yeah for live video, 5 seconds is a long time ... output GOPs are [currently ~3 seconds](https://github.com/livepeer/ai-worker/blob/bb402bf9b1b55e14b7f619abab804b2ca13b1b41/runner/app/live/trickle/segmenter.py#L19), so we'd be almost 2 segments behind the live edge by the time the sleep completes. I'd do a short sleep, maybe 500ms max, and retry ~5 times

would also be helpful to add a comment explaining why the numbers are the way they are (to fall within a output GOP)

j0sh (approved):

PR #3303, opened 2024-12-06
server/ai: Fix memory leak on trickle subscribe

victorges:
Fix the closing of each segment body on the trickle subscribe function.

We used a defer which would keep the body open until the end of the whole stream.
This makes sure that doesn't happen.

- Remove the defer and close segment body explicitly

Run a stream and make sure it still works. Didn't check change in memory but it's a pretty simple scenario.

N/A

**Checklist:**


j0sh (approved):
Good catch! 🙌

PR #3308, opened 2024-12-11
ai/live: Slow orchestrator detection

j0sh:
Detect 'slow' orchs by keeping track of in-flight segments.
Count the difference between segments produced and segments completed.
There should only be ~1 segment in-flight concurrently.

Sometimes the beginning of the current segment may briefly overlap
with the end of the previous segment, so accommodate that by checking
for the second-to-last segment.


leszko:
I'm not sure I understand it correctly, but isn't 1 segment in flight a very strict requirement? I'm a little concern about the case like this:
- Gateway has some temp network delays
- It results in detecting each Orchestartor as slow
- We swap from O to O, effectively taking the whole network capacity

mjh1:
> I'm not sure I understand it correctly, but isn't 1 segment in flight a very strict requirement? I'm a little concern about the case like this:
> 
> * Gateway has some temp network delays
> * It results in detecting each Orchestartor as slow
> * We swap from O to O, effectively taking the whole network capacity

@j0sh Yeah maybe we can give a couple of chances, if it's detected as slow 2 or 3 times in a row then terminate. What's the effect on the output stream if there are more than 1 segment in flight? Would it just skip a segment which wasn't processed quickly enough and the stream pauses then jumps forward to live?

j0sh:
> What's the effect on the output stream if there are more than 1 segment in flight?

Assuming things are just "slow", the subscriptions would still run in order. If segment A is slow, but segments B completes at a normal pace and C is in-flight, then downstream would continue reading A until its done, then fetch B (which would hopefully download more quickly), then C (which the server will 'catch up' then trickle out)

The server currently retains the last [5 segments](https://github.com/livepeer/go-livepeer/blob/master/trickle/trickle_server.go#L78) so if the subscriber falls more than 5 behind, it will 404 (which is not the best behavior right now; I will update that separately to indicate "this stream exists but this segment doesn't" so the client can handle that better, eg try the next segment or jump up to the leading edge)

j0sh:
Relaxed the check to allow up to 3 segments inflight, and reworked a few things in preparation for the next round of changes after this PR: https://github.com/livepeer/go-livepeer/pull/3308/commits/7c00baed0de3dffb0c710448576322b1b092386b

leszko (approved):
Ok, as discussed in Discord, let's merge this one (without O's swap) and we'll think later what to do with O's swaps.

PR #3309, opened 2024-12-11
Fix race condition with stream existence check

mjh1:

In `startTrickleSubscribe()` we check the stream exists in `LivePipelines` but this is only set up inside `startControlPublish()` so I'm moving the initialise step above both these functions so there's no longer a race condition.

**Checklist:**

### DIFF
--- a/server/ai_process.go
+++ b/server/ai_process.go
@@ -1048,7 +1048,9 @@ func submitLiveVideoToVideo(ctx context.Context, params aiRequestParams, sess *A
 		if err != nil {
 			return nil, fmt.Errorf("invalid control URL: %w", err)
 		}
-		// TODO any errors from these funcs should we kill the input stream?
+
+		params.node.LivePipelines[params.liveParams.stream] = &core.LivePipeline{}
+
 		startTricklePublish(ctx, pub, params, sess)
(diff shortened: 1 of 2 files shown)


server/ai_process.go
```diff
@@ -1048,13 +1048,14 @@ func submitLiveVideoToVideo(ctx context.Context, params aiRequestParams, sess *A
...
 		if err != nil {
 			return nil, fmt.Errorf("invalid events URL: %w", err)
 		}
 		clog.V(common.VERBOSE).Infof(ctx, "pub %s sub %s control %s events %s", pub, sub, control, events)
+
+		params.node.LivePipelines[params.liveParams.stream] = &core.LivePipeline{}
+
 		startTricklePublish(ctx, pub, params, sess)
 		startTrickleSubscribe(ctx, sub, params)
 		startControlPublish(control, params)
```
j0sh: Nice catch! I wonder if we can achieve the same effect by calling `startControlPublish` before `startTricklePublish` without having to change anything else?

mjh1: Yeah you're right that will work too, done 👍

j0sh (approved):

leszko (approved):

PR #3310, opened 2024-12-11
gateway/events: Create status API with stream/pipeline ID info

gioelecerati:

Draft

**Checklist:**


monitor/pipeline_status.go
```diff
@@ -5,26 +5,28 @@ import (
 )
 
 var (
-	// pipelineStatusMap stores the latest pipeline status for each stream
-	pipelineStatusMap = make(map[string]PipelineStatus)
-	pipelineStatusMu  sync.RWMutex
+	streamStatusStore = make(map[string]map[string]interface{})
+	streamStatusMu    sync.RWMutex
```
victorges: Same re wrapping this in a struct! WDYT?
```go
type streamStatusStore struct {
	streamStatusStore map[string]map[string]interface{}
	streamStatusMu    sync.RWMutex
}

var StreamStatusStore = streamStatusStore{streamStatusStore: make(map[string]map[string]interface{})}
```

Then:
- Update methods below to be on `*streamStatusStore`
- Method names can be simplifified, like just `func (s *streamStatusStore) Get(streamID string)`
- Usage looks better too IMO e.g. `StreamStatusStore.Get(streamId)`

server/ai_http.go
```diff
@@ -801,3 +804,29 @@ func parseMultiPartResult(body io.Reader, boundary string, pipeline string) core
...
+
+// handleStreamStatus returns the latest available status of a live-video-to-video stream
+func (h *lphttp) handleStreamStatus() http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		if r.Method != http.MethodGet {
+			http.Error(w, "Method not allowed", http.StatusMethodNotAllowed)
+			return
+		}
+
+		streamID := r.URL.Query().Get("stream")
```
victorges: Could this be in the path? I'd say it's more idiomatic

monitor/pipeline_status.go
victorges, line 1: I'm not sure if this makes sense in the `monitor` package, as that's used more for the "non functional" metrics of `go-livepeer` itself being exported to other systems (like prometheus, kafka, etc). This storage is a functional storage instead, which is even served from a public facing API.

So... WDYT of moving this to the `ai` package? Or even `server` if that doesn't make sense, as that seems like a wildcard place now.

server/ai_http.go
```diff
@@ -801,3 +804,29 @@ func parseMultiPartResult(body io.Reader, boundary string, pipeline string) core
...
+		}
+
+		// Get status for specific stream
+		status, exists := monitor.GetStreamStatus(streamID)
+		if !exists {
+			respondWithError(w, "Stream status not found", http.StatusNotFound)
+			return
+		}
+
+		jsonData, _ := json.Marshal(status)
```
victorges: I know it's very rare, but preferrably do not ignore this error, return a 500 if you get sth

server/ai_live_video.go
```diff
@@ -175,22 +175,21 @@ func startControlPublish(control *url.URL, params aiRequestParams) {
...
 	subscriber := trickle.NewTrickleSubscriber(url.String())
+	stream := params.liveParams.stream
 
 	clog.Infof(ctx, "Starting event subscription for URL: %s", url.String())
 
 	go func() {
+		defer monitor.ClearStreamStatus(stream)
 		for {
-			clog.Infof(ctx, "Attempting to read from event subscription for URL: %s", url.String())
+			clog.Infof(ctx, "Reading from event subscription for URL: %s", url.String())
```
victorges, line 189: Such confidence

server/ai_live_video.go
```diff
@@ -175,22 +175,21 @@ func startControlPublish(control *url.URL, params aiRequestParams) {
...
 			segment, err := subscriber.Read()
 			if err != nil {
 				clog.Infof(ctx, "Error reading events subscription: %s", err)
-				// TODO
-				// monitor.DeletePipelineStatus(params.liveParams.stream)
+				monitor.ClearStreamStatus(stream)
```
victorges: This is redundant with the `defer` I think?

server/ai_live_video.go
```diff
@@ -199,30 +198,47 @@ func startEventsSubscribe(ctx context.Context, url *url.URL, params aiRequestPar
...
+			}
 
-			// TODO: update the in-memory pipeline status
-			// monitor.UpdatePipelineStatus(stream, status)
+			monitor.SendQueueEventAsync("stream_status", status)
 
-			clog.Infof(ctx, "Received event for stream=%s status=%+v", stream, status)
+			streamStatus := map[string]interface{}{
+				"updatedAt": time.Now(),
+			}
```
victorges: The timestamp comes from the runner. This should not be here.

server/ai_live_video.go
```diff
@@ -199,30 +198,47 @@ func startEventsSubscribe(ctx context.Context, url *url.URL, params aiRequestPar
...
 
-			stream := params.liveParams.stream
-
-			if stream == "" {
-				clog.Infof(ctx, "Stream ID is missing")
-				continue
-			}
-
-			var status monitor.PipelineStatus
+			var status map[string]interface{}
```
victorges: FTR: These events are not status anymore. They are just events. So IMO you should rename this var to just `event` and only treat it as a `status` when you check `if typeVal, ok := status["type"]; ok && typeVal == "status" {`

This is just naming, but we will avoid more conceptual confusion in the future as we iterate on this.

server/ai_live_video.go
```diff
@@ -199,30 +198,47 @@ func startEventsSubscribe(ctx context.Context, url *url.URL, params aiRequestPar
...
-			status.StreamID = &stream
+			if _, ok := status["stream_id"]; !ok {
+				clog.Infof(ctx, "Received event for stream=%s status=%+v", stream, status)
+				status["stream_id"] = stream
+				status["request_id"] = params.liveParams.requestID
+			}
 
-			// TODO: update the in-memory pipeline status
-			// monitor.UpdatePipelineStatus(stream, status)
+			monitor.SendQueueEventAsync("stream_status", status)
```
victorges: Similarly, this should also not be called `stream_status`. Maybe `stream_events`? Or even `ai_stream_events`? (or in the singular form, not sure what's the kafka pattern rn)

server/ai_live_video.go
```diff
@@ -199,30 +198,47 @@ func startEventsSubscribe(ctx context.Context, url *url.URL, params aiRequestPar
...
+			}
 
-			monitor.SendQueueEventAsync(
-				"stream_status",
-				status,
-			)
+			lastStreamStatus, ok := monitor.GetStreamStatus(stream)
+			if !ok {
+				lastStreamStatus = make(map[string]interface{})
+			}
```
victorges: **nit:** A `nil` map behaves **exactly the same** as an empty map when you are reading from it. It only explodes when writing. Given this `lastStreamStatus` is read-only (or it should be), this is redundant and can be removed.

server/ai_live_video.go
```diff
@@ -199,30 +198,47 @@ func startEventsSubscribe(ctx context.Context, url *url.URL, params aiRequestPar
...
+
+			if typeVal, ok := status["type"]; ok && typeVal == "status" {
+				if logs, ok := status["last_restart_logs"]; !ok || logs == nil {
+					if lastLogs, exists := lastStreamStatus["lastRestartLogs"]; exists && lastLogs != nil {
+						streamStatus["lastRestartLogs"] = lastLogs
+					}
+				} else {
+					streamStatus["lastRestartLogs"] = logs
+				}
```
victorges: I think this is a bit overcomplicated. This will work almost [1] the same:
```suggestion
				if logs, ok := status["last_restart_logs"]; !ok || logs == nil {
				    streamStatus["last_restart_logs"] = lastStreamStatus["last_restart_logs"]
				}
```

Notice that: 
- you had the wrong casing for the field name. I miss typing to catch stuff like this but let's stay with this untyped map 🥲
- you don't need to set `streamStatus["lastRestartLogs"] = logs`, that's a noop

[1] it will always add a `null` `last_restart_logs`, instead of omitting it when it was never sent. But I think it's fine?

server/ai_live_video.go
```diff
@@ -199,30 +198,47 @@ func startEventsSubscribe(ctx context.Context, url *url.URL, params aiRequestPar
...
+
+				if params, ok := status["last_params"]; !ok || params == nil {
+					if lastParams, exists := lastStreamStatus["lastParams"]; exists && lastParams != nil {
+						streamStatus["lastParams"] = lastParams
+					}
+				} else {
+					streamStatus["lastParams"] = params
+				}
```
victorges: Same re simplifying this

server/ai_live_video.go
```diff
@@ -199,30 +198,47 @@ func startEventsSubscribe(ctx context.Context, url *url.URL, params aiRequestPar
...
+			}
+			monitor.StoreStreamStatus(stream, streamStatus)
```
victorges: This should be inside the above if otherwise you're saving any event as if it were a status?

server/ai_live_video.go
victorges: In general, seems like you're using `camelCase`. You should not since the events come from python which is `snake_case`. This might be inconsistent with other `go-livepeer` APIs (not sure, but studio for sure), but let's keep it as snake for now. Let's say all new AI APIs should be snake.

server/ai_live_video.go
gioelecerati: This is a requirement from data structure. It's a specific event type for this kind of data - other events may have different naming.
cc @ecmulli for input

victorges: WDYM, we should send status on one stream and events on another? I don't see this logic here tho

ecmulli: I'm assuming this is related to our discord [conversation](https://discord.com/channels/423160867534929930/1316106387666894888/1316419426286829580).

`stream_status` refers to the interval checks that outputs fps, restarts, etc. I'd also accept `ai_stream_status` if its important to distinguish between ai and regular video.

lets use `stream_events` (or `ai_stream_events`) to capture the actual events, i.e. errors/restarts/params_update.

ecmulli: The higher level context is that these names are used in clickhouse to create specific tables. for example `type=network_payment` gets directed into a payments table. but they all get ingested via the same kafka topic
image

victorges (approved):
LGTM

PR #3312, opened 2024-12-11
Record current number of live pipelines running

mjh1:

Introduce a new metric to record the current number of inflight live AI pipelines.

**Checklist:**

### DIFF
--- a/server/ai_live_video.go
+++ b/server/ai_live_video.go
@@ -152,6 +152,9 @@ func startControlPublish(control *url.URL, params aiRequestParams) {
 		ControlPub:  controlPub,
 		StopControl: stop,
 	}
+	if monitor.Enabled {
+		monitor.AICurrentLiveSessions(len(params.node.LivePipelines))
+	}
 
 	// send a keepalive periodically to keep both ends of the connection alive
 	go func() {
--- a/server/ai_mediaserver.go
+++ b/server/ai_mediaserver.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"github.com/livepeer/go-livepeer/monitor"
 	"io"
 	"log/slog"
 	"net/http"
@@ -555,6 +556,9 @@ func (ls *LivepeerServer) cleanupLive(stream string) {
 	pub, ok := ls.LivepeerNode.LivePipelines[stream]
 	delete(ls.LivepeerNode.LivePipelines, stream)
 	ls.LivepeerNode.LiveMu.Unlock()
+	if monitor.Enabled {
+		monitor.AICurrentLiveSessions(len(ls.LivepeerNode.LivePipelines))
+	}
 
 	if ok && pub != nil && pub.ControlPub != nil {
 		if err := pub.ControlPub.Close(); err != nil {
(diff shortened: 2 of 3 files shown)


server/ai_live_video.go
```diff
@@ -152,6 +152,9 @@ func startControlPublish(control *url.URL, params aiRequestParams) {
 		ControlPub:  controlPub,
 		StopControl: stop,
 	}
+	if monitor.Enabled {
```
mjh1, line 155: I think we should probably move this initialisation of the LivePipeline somewhere more central now that we're using it to check stream existence as well as just storing these control related fields. Also do we need a mutex around it like we do in `cleanupLive()` @j0sh ?

server/ai_live_video.go
```diff
@@ -152,6 +152,9 @@ func startControlPublish(control *url.URL, params aiRequestParams) {
 		ControlPub:  controlPub,
```
mjh1, line 152: I think we should probably move this initialisation of the LivePipeline somewhere more central now that we're using it to check stream existence as well as just storing these control related fields. Also do we need a mutex around it like we do in `cleanupLive()` @j0sh ?

j0sh: Up to you... I don't see a big issue right now. We might outgrow it at some point but if you want to be preemptive about refactoring, go ahead. Just want to avoid spreading initialization logic around too much, while it is kinda self contained right now.

>  Also do we need a mutex around it like we do in cleanupLive()

Yep we need the mutex but I believe we already hold it at this point. Only slight issue is that we also hold the mutex while calling `stats.Record` but that should be OK for the most part. Recording prom metrics is supposed to be fairly quick but its still kind of non-trivial IIRC

j0sh (approved):

server/ai_live_video.go
mjh1: Oh yeah missed the mutex 👍

server/ai_mediaserver.go
```diff
@@ -555,6 +556,9 @@ func (ls *LivepeerServer) cleanupLive(stream string) {
 	pub, ok := ls.LivepeerNode.LivePipelines[stream]
 	delete(ls.LivepeerNode.LivePipelines, stream)
 	ls.LivepeerNode.LiveMu.Unlock()
+	if monitor.Enabled {
+		monitor.AICurrentLiveSessions(len(ls.LivepeerNode.LivePipelines))
+	}
```
pwilczynskiclearcode, line 561: @mjh1 @j0sh Are we sure that `cleanupLive()` is called always, even in case of errors, panics, etc?

The number of used in-use orchestrators seems high https://eu-metrics-monitoring.livepeer.live/grafana/d/be6llteqebk00b/ai-overview-livestream?orgId=1&from=1734220800000&to=now&viewPanel=6

mjh1: I'm wondering if we might be double counting somehow, the current number is 12 which is double Eric's number of 24/7 streams

thomshutt: @mjh1 I think the versions with / without music are separate streams

mjh1: Yeah I was wondering that, I think the added music streams are just regular studio streams only though https://github.com/ericxtang/random-ai-stream

PR #3315, opened 2024-12-13
ai/live: Add limited retries for segment publish

j0sh:
Only retry if the error occurs before sending any data, and if the next segment hasn't arrived yet.

Depends on #3308

### DIFF
--- a/server/ai_live_video.go
+++ b/server/ai_live_video.go
@@ -54,31 +54,56 @@ func startTricklePublish(ctx context.Context, url *url.URL, params aiRequestPara
 		// check for end of stream
 		if _, eos := reader.(*media.EOSReader); eos {
 			if err := publisher.Close(); err != nil {
-				clog.Infof(ctx, "Error closing trickle publisher. err=%s", err)
+				clog.Infof(ctx, "Error closing trickle publisher. err=%v", err)
 			}
 			cancel()
 			return
 		}
-		if _, atMax := slowOrchChecker.BeginSegment(); atMax {
+		thisSeq, atMax := slowOrchChecker.BeginSegment()
+		if atMax {
 			clog.Infof(ctx, "Orchestrator is slow - terminating")
 			cancel()
 			return
 			// TODO kill the rest of the processing, including ingest
 			// TODO switch orchestrators
 		}
-		go func() {
+		go func(seq int) {
 			defer slowOrchChecker.EndSegment()
 			var r io.Reader = reader
 			if paymentProcessor != nil {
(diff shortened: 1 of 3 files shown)


server/ai_live_video.go
```diff
@@ -54,31 +54,56 @@ func startTricklePublish(ctx context.Context, url *url.URL, params aiRequestPara
...
+				// Retry segment only if nothing has been sent yet
+				// and the next segment has not yet started
+				// otherwise drop
+				if n > 0 {
+					clog.Infof(ctx, "Error publishing segment; dropping remainder wrote=%d err=%v", n, err)
+					return
+				}
+				clog.Infof(ctx, "Error publishing segment before writing; retrying err=%v", err)
+				// Clone in case read head was incremented somewhere, which cloning ressets
+				r = reader.Clone()
```
victorges, line 103: Hmm this seems maybe not 100% guaranteed, since it could happen that something _was_ read from the reader and then failed to be sent to the writer (and we got `0` on L89).

I think there's no really easy option here apart from maybe an additional layer to check if the reader was used at some point, like a wrapped reader that flags if it was read. A built-in approach could be using a `io.TeeReader` to read into a buffer as we read from the input (which would also allow us to retry when _some_ bytes have been read).

Anyway, if you think this is too much of a corner case, this LGTM as it's much simpler. But I think that conceptually it could still have this issue 🤔

victorges (approved):
lgtm

server/ai_live_video.go
j0sh: > since it could happen that something was read from the reader and then failed to be sent to the writer (and we got 0 on L89).

That is basically why we call [`Clone()`](https://github.com/livepeer/go-livepeer/blob/master/media/rw.go#L109-L113) (which resets the read head for the clone) so when we try to read it during the next iteration, it is starting from the beginning, regardless of the actual read head. Note that this isn't a "standard" IO reader but rather something [specialized](https://github.com/livepeer/go-livepeer/blob/7fe7949af9f75d5f74a4d3b6bb6e5ad1f758acbe/media/rw.go#L88-L107) that allows for clean resets, nonblocking reads/writes, multiple readers, etc, without having to go through hoops like TeeReader.

mjh1 (approved):

server/ai_live_video.go
victorges: Ohh that's cool! Didn't know we already had that custom impl. Pretty neat :)

leszko (approved):

PR #3316, opened 2024-12-13
gateway: fix stream-status route

gioelecerati:

- Move the stream-status api on the correct handler

**Checklist:**

### DIFF
--- a/server/ai_mediaserver.go
+++ b/server/ai_mediaserver.go
@@ -584,8 +584,7 @@ func (ls *LivepeerServer) UpdateLiveVideo() http.Handler {
 	})
 }
 
-// GetLiveVideoToVideoStatus returns the latest available status of a live-video-to-video stream
-func (ls *LivepeerServer) GetLiveVideoToVideoStatus() http.HandlerFunc {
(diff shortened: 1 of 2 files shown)


server/ai_mediaserver.go
```diff
@@ -87,6 +87,9 @@ func startAIMediaServer(ls *LivepeerServer) error {
 	ls.HTTPMux.Handle("/live/video-to-video/{prefix}/{stream}/start", ls.StartLiveVideo())
 	ls.HTTPMux.Handle("/live/video-to-video/{stream}/update", ls.UpdateLiveVideo())
 
+	// Stream status
+	ls.HTTPMux.Handle("/live/video-to-video/{stream}/status", ls.GetLiveVideoToVideoStatus())
```
victorges: Should we keep the `{stream-id}` param not to be confused with the `{stream}` above, which is NOT the stream ID?

victorges (approved):

server/ai_mediaserver.go
gioelecerati: Done!

PR #3331, opened 2024-12-19
ai/live: Send event when auth fails

j0sh:
- Send an event when auth fails. If auth is broken, we don't get any events, so it can be hard to determine from events what the state of things is.
- Set the streamID = streamKey by default. This is helpful in dev mode without a webhook setup. If a webhook specifies streamID then that is still used.
- Update some logging

### DIFF
--- a/server/ai_mediaserver.go
+++ b/server/ai_mediaserver.go
@@ -430,7 +430,8 @@ func (ls *LivepeerServer) StartLiveVideo() http.Handler {
 		// if auth webhook returns pipeline config these will be replaced
 		pipeline := qp.Get("pipeline")
 		rawParams := qp.Get("params")
-		var streamID, pipelineID string
+		streamID := streamName
+		var pipelineID string
 		var pipelineParams map[string]interface{}
 		if rawParams != "" {
 			if err := json.Unmarshal([]byte(rawParams), &pipelineParams); err != nil {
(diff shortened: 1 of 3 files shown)


server/ai_mediaserver.go
```diff
@@ -454,6 +455,10 @@ func (ls *LivepeerServer) StartLiveVideo() http.Handler {
 					clog.Errorf(ctx, "failed to kick input connection: %s", kickErr.Error())
 				}
 				clog.Errorf(ctx, "Live AI auth failed: %s", err.Error())
+				monitor.SendQueueEventAsync("ai_stream_events", map[string]string{
+					"type":    "error",
+					"message": "Live AI auth failed: " + err.Error(),
```
j0sh, line 460: @ecmulli @gioelecerati wanted to check this would not break anything downstream since we only have a limited set of fields here. Most of the rest are unknown at this point - they come as part of the auth webhook, but the auth webhook has failed here.

leszko (approved):
Looks good, but let's wait for @ecmulli @gioelecerati before merging to confirm/answer your inline question

mjh1 (approved):

server/ai_mediaserver.go
j0sh: Ping on this @ecmulli @gioelecerati

gioelecerati: @j0sh this shouldn't break anything, since the messages are sent to Kafka without any deserialization - worst case scenario they won't be collected anywhere

ecmulli: @j0sh, @gioelecerati is correct here. the missing fields should be ignored. 

is streamId, pipelineID, etc automatically included in this?

j0sh: > is streamId, pipelineID, etc automatically included in this?

They are not included, because they don't exist at this point. We typically receive those via the auth webhook but this event is reporting that the auth webhook has failed.

If we wanted some form of identifier we could include the stream key as `stream_key` or something, although that is not part of other events IIRC. The key itself might also be a little strange from brute force connection attempts which has some risk to downstream consumers through eg, injection.

PR #3341, opened 2025-01-07
[ENG-2342] ai_first_segment_delay metric added

pwilczynskiclearcode:
Adding `ai_first_segment_delay` prometheus metric.
It measures the time between AI Gateway receiving a GenLiveVideoToVideo request and successfully receiving the first processed segment from O.

### DIFF
--- a/monitor/census.go
+++ b/monitor/census.go
@@ -376,7 +376,7 @@ func InitCensus(nodeType NodeType, version string) {
 	census.mAIResultUploadTime = stats.Float64("ai_result_upload_time_seconds", "Upload (to Orchestrator) time", "sec")
 	census.mAIResultSaveFailed = stats.Int64("ai_result_upload_failed_total", "AIResultUploadFailed", "tot")
 	census.mAICurrentLivePipelines = stats.Int64("ai_current_live_pipelines", "Number of live AI pipelines currently running", "tot")
-	census.mAIFirstSegmentDelay = stats.Int64("ai_first_segment_delay", "Delay of the first live AI segment being processed", "ms")
+	census.mAIFirstSegmentDelay = stats.Int64("ai_first_segment_delay_ms", "Delay of the first live AI segment being processed", "ms")
 
 	glog.Infof("Compiler: %s Arch %s OS %s Go version %s", runtime.Compiler, runtime.GOARCH, runtime.GOOS, runtime.Version())
 	glog.Infof("Livepeer version: %s", version)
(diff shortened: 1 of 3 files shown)


monitor/census.go
```diff
@@ -375,6 +376,7 @@ func InitCensus(nodeType NodeType, version string) {
 	census.mAIResultUploadTime = stats.Float64("ai_result_upload_time_seconds", "Upload (to Orchestrator) time", "sec")
 	census.mAIResultSaveFailed = stats.Int64("ai_result_upload_failed_total", "AIResultUploadFailed", "tot")
 	census.mAICurrentLivePipelines = stats.Int64("ai_current_live_pipelines", "Number of live AI pipelines currently running", "tot")
+	census.mAIFirstSegmentDelay = stats.Int64("ai_first_segment_delay", "Delay of the first live AI segment being processed", "ms")
```
leszko: Nit: maybe `ai_first_segment_delay__milliseconds`?

server/ai_live_video.go
```diff
@@ -137,6 +137,8 @@ func startTrickleSubscribe(ctx context.Context, url *url.URL, params aiRequestPa
 	// read segments from trickle subscription
 	go func() {
 		var err error
+		firstSegment := false
```
leszko:
```suggestion
		firstSegment := true
```

I think it needs to be `true` first.

monitor/census.go
```diff
@@ -1987,6 +1996,10 @@ func AIResultDownloaded(ctx context.Context, pipeline string, model string, down
 	}
 }
 
+func AIFirstSegmentDelay(delay int64) {
+	stats.Record(census.ctx, census.mAIFirstSegmentDelay.M(delay))
```
leszko: I wonder if we may want to tag it with Orchestrator's address? 

I can imagine that for the given Gateway, only some Os may be slow.

monitor/census.go
pwilczynskiclearcode: We pass `ms` as the last parameter so I believe that prometheus and then grafana has the unit information already in.

server/ai_live_video.go
pwilczynskiclearcode: 🤦🏻‍♂️ right

monitor/census.go
pwilczynskiclearcode: Yes. That would be interesting information.

The question is how to get Orchestrator info in `startTrickleSubscribe()` function.
Where of if does it live on the Gateway code? We'd need `OrchestratorInfo` struct, I think.

leszko: You can pass [host](https://github.com/livepeer/go-livepeer/blob/c8cbe797a08ab91bf5b116e04dd55731231e91e2/server/ai_process.go#L1060) to `startTrickleSubscribe()`.

This `host` is good as a tag. It's the IP/DNS of the Orchestrator.

leszko (commented):
Added one more comment on how to tag with `host`

monitor/census.go
pwilczynskiclearcode: OK. I took entire orchInfo and unified it a bit on the census-side

leszko (approved):

